PR #33315, opened 2019-10-22
refactor(core): better error message for undecorated-classes-with-di migration

devversion:
Currently if one of the project targets could not be analyzed
due to AOT compiler program failures, we gracefully proceed
with the migration. This is expected, but we should not
print a message at the end of the migration that the migration
was _successful_. The migration was only done partially, hence
the migration is potentially incomplete and we should make it clear that once
the failures are resolved, the migration should be re-run.


rbuzzibarrios (approved):

kara (approved):
LGTM

kara:
@devversion Can you rebase?

PR #33318, opened 2019-10-22
fix(i18n): handle legacy message ids in ICUs

petebacondarwin:
All but the last commit are just refactorings that made it easier for me to understand what is going on. The bug fix is in the last commit.

See FW-1637

### DIFF
--- a/packages/compiler-cli/test/ngtsc/ngtsc_spec.ts
+++ b/packages/compiler-cli/test/ngtsc/ngtsc_spec.ts
@@ -2179,6 +2179,23 @@ runInEachFileSystem(os => {
       expect(jsContents).not.toContain(':Some text');
     });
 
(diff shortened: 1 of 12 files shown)


AndrewKushnir (approved):
LGTM, thanks for additional refactoring @petebacondarwin  👍 
One comment (we can fix in the next PRs) - we now have 2 `I18nMeta` types: one represents i18n message (including AST), the second one represents metadata collected from `i18n` attributes. We probably need to find different names to disambiguate that in the code.

AndrewKushnir:
- [VE Presubmit](http://test/OCL:276074212:BASE:276074273:1571759941075:168a0b41)
- [Ivy Presubmit](http://test/OCL:276074212:BASE:276074399:1571759996021:6ae69234)

PR #33321, opened 2019-10-22
ci: publish tarballs for all Angular packages as build artifacts on PR builds

gkalpak:
Previously, when one wanted to try out the changes from a PR before it was merged, they had to check out the PR locally and build the Angular packages themselves (which is time-consuming and wasteful given that the packages have already been built on CI).

This commit persists all Angular packages on each build as `.tgz` files, which can be used to install dependencies on an project (supported by both [npm][1] and [yarn][2]).

CircleCI [build artifacts][3] an be used for longer-term storage of the outputs of a build and are designed to be useful around the time of the build, which suits our needs.

[1]: https://docs.npmjs.com/cli/install.html
[2]: https://yarnpkg.com/lang/en/docs/cli/add
[3]: https://circleci.com/docs/2.0/artifacts

### DIFF
--- a/docs/DEVELOPER.md
+++ b/docs/DEVELOPER.md
@@ -8,6 +8,10 @@ It also explains the basic mechanics of using `git`, `node`, and `yarn`.
 * [Installing NPM Modules](#installing-npm-modules)
 * [Building](#building)
 * [Running Tests Locally](#running-tests-locally)
+* [Formatting your Source Code](#formatting-your-source-code)
+* [Linting/verifying your Source Code](#lintingverifying-your-source-code)
+* [Publishing Snapshot Builds](#publishing-snapshot-builds)
+* [Bazel Support](#bazel-support)
 
 See the [contribution guidelines](https://github.com/angular/angular/blob/master/CONTRIBUTING.md)
 if you'd like to contribute to Angular.
@@ -50,6 +54,7 @@ cd angular
 # Add the main Angular repository as an upstream remote to your repository:
 git remote add upstream https://github.com/angular/angular.git
 ```
+
 ## Installing NPM Modules
 
 Next, install the JavaScript modules needed to build and test Angular:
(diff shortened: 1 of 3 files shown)


docs/DEVELOPER.md
```diff
@@ -118,33 +124,79 @@ let g:clang_format#command = '$ANGULAR_PATH/node_modules/.bin/clang-format'
...
-http://github.com/angular/core-builds.
+When a build of any branch on the upstream fork angular/angular is green on CircleCI, it
+automatically publishes build artifacts to repositories in the Angular org, eg. the `@angular/core`
+package is published to https://github.com/angular/core-builds.
 
 You may find that your un-merged change needs some validation from external participants.
-Rather than requiring them to pull your Pull Request and build Angular locally, you can
-publish the `*-builds` snapshots just like our CircleCI build does.
+Rather than requiring them to pull your Pull Request and build Angular locally, you can publish
+snapshots of the Angular packages.
```
josephperrott: In a few locations in this doc, the phrase "you can publish" is used.  I am worried that it sounds like an action that the user needs to take rather than something that will happen for them.

Should we say something like:

>Rather than requiring them to pull your Pull Request and build Angular locally, they can depend on snapshots of the Angular packages created for the Pull Request by each CI run.

scripts/ci/create-package-archives.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
```
josephperrott: Can you add a bit more logging, expressing when the all packages tarball is created and when things start and complete?

josephperrott (approved):
Overall LGTM, we will need to have @IgorMinar review the `docs/DEVELOPER.md` change as well.

docs/DEVELOPER.md
gkalpak: Good point. (This is actually the only occurrence where the phrase is inaccurate (in other places they do indeed need to take an action) :grin:)
Given that the snapshots are not published by CI in all methods described below, I changed it to:

> Rather than requiring them to pull your Pull Request and build Angular locally, they can depend on snapshots of the Angular packages created based on the code in the Pull Request.

I also updated the **Publishing to GitHub repos** section to make it clear that they need to have built the packages locally first.

scripts/ci/create-package-archives.sh
gkalpak: Done.

IgorMinar (approved):
wow. this is awesome! thanks

AndrewKushnir:
@gkalpak there was a conflict merging this PR to patch branch, so I merged it to master. Could you please create a followup PR targeting patch branch with the necessary changes? Thank you.

PR #33329, opened 2019-10-22
build: update zone.js to use the new rollup_bundle

gregmagolan:
This switches the zone.js build to use the new nodejs `rollup_bundle` and `terser_minifed` rules and downlevel with the auto-generated `tsc` rule.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
     "@bazel/jasmine": "0.40.0",
     "@bazel/karma": "0.40.0",
     "@bazel/protractor": "0.40.0",
+    "@bazel/rollup": "0.40.0",
     "@bazel/terser": "0.40.0",
     "@bazel/typescript": "0.40.0",
     "@microsoft/api-extractor": "^7.3.9",
--- a/packages/zone.js/BUILD.bazel
+++ b/packages/zone.js/BUILD.bazel
@@ -3,6 +3,8 @@ load("//packages/zone.js:bundles.bzl", "ES2015_BUNDLES", "ES5_BUNDLES", "ES5_GLO
 
 exports_files([
     "tsconfig.json",
+    "rollup-es5.config.js",
+    "rollup-es5_global-es2015.config.js",
 ])
 
 genrule(
--- a/packages/zone.js/dist/BUILD.bazel
+++ b/packages/zone.js/dist/BUILD.bazel
@@ -21,13 +21,12 @@ genrule(
 [
     rollup_bundle(
         name = b[0].replace("-", "_") + "_rollup",
+        config_file = "//packages/zone.js:rollup-es5.config.js",
         entry_point = b[1] + ".ts",
-        globals = {
-            "electron": "electron",
-        },
-        license_banner = "//packages:license-banner.txt",
         deps = [
             "//packages/zone.js/lib",
+            "@npm//rollup-plugin-commonjs",
+            "@npm//rollup-plugin-node-resolve",
         ],
     )
     for b in ES5_BUNDLES.items()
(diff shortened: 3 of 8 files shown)


AndrewKushnir:
@gregmagolan, there is a new patch branch created (`9.0.x`), could you please verify whether it's a master-only change or should also be included in `9.0.x`? Thank you.

gregmagolan:
@AndrewKushnir This PR only changes the build and has no external effect so master is sufficient.

gregmagolan:
On 2nd thought we should probably land this on patch as well if we are going to update patch to nodejs rules 1.0.0 at some point then its needed.

JiaLiPassion:
Hi, @gregmagolan , I have 2 questions.
in the  new built zone.js `npm_package`.
the version in the license banner looks like this.

```
* @license Angular v0.0.0
```

should it be 

```
* @license Angular v10.0.2
```
2. And the generated function looks like having an additional set of ()
the original bundle is :

```
(function (factory) {
    typeof define === 'function' && define.amd ? define(factory) :
        factory();
}(function () {
}));
```

The new generated looks like

```
(function (factory) {
    typeof define === 'function' && define.amd ? define(factory) :
        factory();
}((function () {
})));
```

Please confirm, thanks!

josephperrott:
@JiaLiPassion The version is left as `0.0.0` unless the `--config=release` is run on the command.  This provides the stamping information.

Though it does expose another issue.  The version numbering is different for zone.js so the stamping system we have in place does not properly stamp the zone.js package.  Which is something we will need to sort out for us to do the release properly.

gregmagolan:
Actually that's a good catch. The legacy rollup_bundle handle the replacement automatically in its templated rollup config but with the new rollup bundle the v0.0.0-PLACEHOLDER replacement needs to be added to the rollup.config.js.

@JiaLiPassion @josephperrott I can add that in to this PR. The license banner used is in `packages/zone.js/rollup-es5_global-es2015.config.js` and `packages/zone.js/rollup-es5.config.js`:

```
const banner = `/**
* @license Angular v0.0.0-PLACEHOLDER
* (c) 2010-2019 Google LLC. https://angular.io/
* License: MIT
*/`;
```

Is this what we want for zone.js?what version should the zone.js dist be stamped with?

JiaLiPassion:
@gregmagolan , @josephperrott , the `version` part is ok, there is document about how to release `zone.js` in https://github.com/angular/angular/blob/master/packages/zone.js/DEVELOPER.md, so the `placeholder` will be replaced with `0.10.2` when release.

Please check the additional `()` code is correct or not, thanks!

gregmagolan:
@JiaLiPassion The extra parentheses is correct and all good.

It was was added to the rollup UMD wrapper was added in rollup 1.25.2 which is now being used for the bundling: https://github.com/rollup/rollup/blob/master/CHANGELOG.md#1252.

It was a performance improvement Here the PR for reference: https://github.com/rollup/rollup/pull/3183

gregmagolan:
Version stamping fixed as well. Can you take another look at the generated package @JiaLiPassion?

JiaLiPassion:
@gregmagolan , LGTM, thanks!

JiaLiPassion (approved):

gregmagolan:
0.40 release is out which includes the changes in the patch here. After the https://github.com/angular/angular/pull/33802 update lands this PR can be updated and land without patches.

gregmagolan:
Patches to `@bazel/terser` removed now that it has an `args` to be able to pass `--comments` to it. This is unblocked and ready

gregmagolan:
Patches to `@bazel/terser` removed now that it has an `args` to be able to pass `--comments` to it. This is unblocked and ready for merging.

tools/BUILD.bazel
```diff
@@ -5,6 +5,7 @@ load("@npm_bazel_typescript//:index.bzl", "ts_config")
 exports_files([
     "tsconfig.json",
     "jasmine-seed-generator.js",
+    "terser_config.json",
```
josephperrott: I am not able to find the `terser_config.json` file?  Am I missing it?  Or is generated somehow?

gregmagolan: Nice catch. That was there earlier but was not needed and missed the reverence. Removing this lint.

josephperrott (approved):
LGTM, global approval

PR #33332, opened 2019-10-22
refactor(bazel): add missing comment about JS providers to ng_module.bzl

gregmagolan:


alexeagle (approved):

gregmagolan:
@caretaker Just a comments PR but looks like still requires a g3sync. Assistance needed. Thanks!

PR #33337, opened 2019-10-22
fix(ivy): move setClassMetadata calls into a pure iife

IgorMinar:
This commit transforms the setClassMetadata calls generated by ngtsc from:

```typescript
/*@__PURE__*/ setClassMetadata(...);
```

to:

```typescript
/*@__PURE__*/ (function() {
  setClassMetadata(...);
})();
```

Without the IIFE, terser won't remove these function calls because the
function calls have arguments that themselves are function calls or other
impure expressions. In order to make the whole block be DCE-ed by terser,
we wrap it into IIFE and mark the IIFE as pure.

It should be noted that this change doesn't have any impact on CLI* with
build-optimizer, which removes the whole setClassMetadata block within
the webpack loader, so terser or webpack itself don't get to see it at
all. This is done to prevent cross-chunk retention issues caused by
webpack's internal module registry.

\* actually we do expect a short-term size regression while angular/angular-cli#16228
is merged and released in the next rc of the CLI. But long term this
change does nothing to CLI + build-optimizer configuration and is done
primarly to correct the seeminly correct but non-function PURE annotation
that builds not using build-optimizer could rely on.


IgorMinar:
merge-assistance: this is a cherry-pick of a commit from #31939 that has been already reviewed and approved. I'm splitting up that PR to land stuff that is good to go.

IgorMinar:
I don't quite get why this is causing size regressions. The generated output before terser pass looks legit: https://hackmd.io/3mGNjS7nThmdq6qcAZQ_uw

And oddly enough the original PR from @alxhub is not failing with this issue: https://github.com/angular/angular/pull/31939

petebacondarwin (approved):

gkalpak:
> And oddly enough the original PR from @alxhub is not failing with this issue: #31939

Even more oddly, the original PR #31939 _is_ failing the [size test for aio][1] but not for integration projects 😕 

[1]: https://circleci.com/gh/angular/angular/490579

filipesilva:
I've looked into this and think that I have identified the problem. It's related to how this PR changes the pattern that allows Build Optimizer (BO) to identify static properties for classes.

In what BO calls "class fold", it tries to "fold" all static properties accesses after the class into a class declaration. We know these are static properties because it is how TS emits them.
```
// input
export class TemplateRef { }
TemplateRef.__NG_ELEMENT_ID__ = () => SWITCH_TEMPLATE_REF_FACTORY(TemplateRef, ElementRef);

// output
export const TemplateRef = /*@__PURE__*/ function () {
  class TemplateRef { }
  TemplateRef.__NG_ELEMENT_ID__ = () => SWITCH_TEMPLATE_REF_FACTORY(TemplateRef, ElementRef);
  return TemplateRef;
}();
```

These property assignments can reference variables, causing them to be retained even if the base class isn't used. Having these property assignment inside the class allows everything to be removed if the class is removed.

Inside `@angular/core` there's a very important class called `ApplicationRef` that has many references. It looks like this after being processed by ngcc:
```
class ApplicationRef {
  // lots of code
}
ApplicationRef.ɵfac = function ApplicationRef_Factory(t) { return new (t || ApplicationRef)(ɵɵinject(NgZone), ɵɵinject(Console), ɵɵinject(Injector), ɵɵinject(ErrorHandler), ɵɵinject(ComponentFactoryResolver), ɵɵinject(ApplicationInitStatus)); };
ApplicationRef.ɵprov = ɵɵdefineInjectable({ token: ApplicationRef, factory: function (t) { return ApplicationRef.ɵfac(t); }, providedIn: null });
/*@__PURE__*/ setClassMetadata(ApplicationRef, [{
        type: Injectable
    }], function () { return [{ type: NgZone }, { type: Console }, { type: Injector }, { type: ErrorHandler }, { type: ComponentFactoryResolver }, { type: ApplicationInitStatus }]; }, { constructor: [], _zone: [], _console: [], _injector: [], _exceptionHandler: [], _componentFactoryResolver: [], _initStatus: [], _bootstrapListeners: [], _views: [], _runningTick: [], _enforceNoNewChanges: [], _stable: [], componentTypes: [], components: [], bootstrap: [], tick: [], attachView: [], detachView: [], _loadComponent: [], _unloadComponent: [], ngOnDestroy: [], viewCount: [] });
/**
 * \@internal
 */
ApplicationRef._tickScope = wtfCreateScope('ApplicationRef#tick()');
/** @nocollapse */
ApplicationRef.ctorParameters = () => [
    { type: NgZone },
    { type: Console },
    { type: Injector },
    { type: ErrorHandler },
    { type: ComponentFactoryResolver },
    { type: ApplicationInitStatus }
];
```

This pattern breaks BOs identification of class static properties because `/*@__PURE__*/ setClassMetadata(...` is in the middle of property accesses. This in turn caused the property accesses after `setClassMetadata` to stay outside the class and retain code.

In https://github.com/angular/angular-cli/pull/15664 we got around this problem by identifying these calls and removing them before the "class fold" transform is ran. This way all properties went inside the class again.

With this PR the shape changes again:
```
ApplicationRef.ɵfac = function ApplicationRef_Factory(t) { return new (t || ApplicationRef)(ɵɵinject(NgZone), ɵɵinject(Console), ɵɵinject(Injector), ɵɵinject(ErrorHandler), ɵɵinject(ComponentFactoryResolver), ɵɵinject(ApplicationInitStatus)); };
ApplicationRef.ɵprov = ɵɵdefineInjectable({ token: ApplicationRef, factory: function (t) { return ApplicationRef.ɵfac(t); }, providedIn: null });
/*@__PURE__*/ (function () { setClassMetadata(ApplicationRef, [{
        type: Injectable
    }], function () { return [{ type: NgZone }, { type: Console }, { type: Injector }, { type: ErrorHandler }, { type: ComponentFactoryResolver }, { type: ApplicationInitStatus }]; }, { constructor: [], _zone: [], _console: [], _injector: [], _exceptionHandler: [], _componentFactoryResolver: [], _initStatus: [], _bootstrapListeners: [], _views: [], _runningTick: [], _enforceNoNewChanges: [], _stable: [], componentTypes: [], components: [], bootstrap: [], tick: [], attachView: [], detachView: [], _loadComponent: [], _unloadComponent: [], ngOnDestroy: [], viewCount: [] }); })();
/**
 * \@internal
 */
ApplicationRef._tickScope = wtfCreateScope('ApplicationRef#tick()');
/** @nocollapse */
ApplicationRef.ctorParameters = () => [
    { type: NgZone },
    { type: Console },
    { type: Injector },
    { type: ErrorHandler },
    { type: ComponentFactoryResolver },
    { type: ApplicationInitStatus }
];
```

`/*@__PURE__*/ (function () { setClassMetadata(ApplicationRef, [{` is identified by BO and is not removed, which causes the properties afterwards to be retained by "class fold".

We could change CLI to also identify the new shape. But if we instead move the `setClassMetadata` call after all static properties, there is no confusion in BO's side. I think that's a better approach because trying to remove certain bits of code in the middle of static props was what caused the problem in the first place.

IgorMinar:
@filipesilva thanks for the info! the whole point of this change is to take away responsibility of the BO to do extra work. It would make no sense to make this change in the emit format and then keep on correcting the emit in BO.

Let's iterate more on this change and get the code emitted correctly. This change can wait and is not RC blocking.

petebacondarwin:
This should help https://github.com/angular/angular/pull/33630

alxhub:
@IgorMinar I rebased this after @petebacondarwin's #33630.

filipesilva:
Looking into the size failure in https://circleci.com/gh/angular/angular/518038#tests/containers/2.
```
Commit a4ac6cb1818cec4bbf9db89b40be94d2cc46f384 uncompressed main-es2015 exceeded expected size by >1% (expected: 268331, actual: 313143).
```

To figure out what changed, I'll compare the file content of the main-es2015 file as built in this pr with the same from the branch point from master (`build: update .nvmrc to correct node version (#33636)`). 

First I got the artifacts for these two builds from CLI (thanks @gkalpak!):
- this PR: https://circleci.com/gh/angular/angular/518037#artifacts/containers/0
- branch point from master: https://circleci.com/gh/angular/angular/517986#artifacts/containers/0

I input the artifacts from this PR to `dist/packages-dist`, and the artifacts from master to `dist/packages-dist-master`. Then I copy `integration/cli-hello-world-lazy` as `integration/cli-hello-world-lazy-master`,edit the latters' package.json deps to point at `dist/packages-dist-master`, and install the dependencies. Unfortunately I run into https://github.com/angular/angular/issues/33686, but just use latest zone.js from npm instead.

At this point I can make comparative builds between the two setups. First we should look at a production build without any optimizations (`ng build --prod --optimization false`) to see how the `setClassMetadata` calls changed. 

A good starting point to compare is `ApplicationRef` where we saw a problem before in https://github.com/angular/angular/pull/33337#issuecomment-545487737.

- master
```
let ApplicationRef = /*@__PURE__*/ (() => {
    class ApplicationRef {
       // lots of code
    }
    ApplicationRef.ɵfac = function ApplicationRef_Factory(t) { return new (t || ApplicationRef)(ɵɵinject(NgZone), ɵɵinject(Console), ɵɵinject(Injector), ɵɵinject(ErrorHandler), ɵɵinject(ComponentFactoryResolver), ɵɵinject(ApplicationInitStatus)); };
    ApplicationRef.ɵprov = /*@__PURE__*/ ɵɵdefineInjectable({ token: ApplicationRef, factory: function (t) { return ApplicationRef.ɵfac(t); }, providedIn: null });
    /**
     * \@internal
     */
    ApplicationRef._tickScope = /*@__PURE__*/ wtfCreateScope('ApplicationRef#tick()');
    /*@__PURE__*/ /*@__PURE__*/ setClassMetadata(ApplicationRef, [{
            type: Injectable
        }], function () { return [{ type: NgZone }, { type: Console }, { type: Injector }, { type: ErrorHandler }, { type: ComponentFactoryResolver }, { type: ApplicationInitStatus }]; }, { constructor: [], _zone: [], _console: [], _injector: [], _exceptionHandler: [], _componentFactoryResolver: [], _initStatus: [], _bootstrapListeners: [], _views: [], _runningTick: [], _enforceNoNewChanges: [], _stable: [], componentTypes: [], components: [], bootstrap: [], tick: [], attachView: [], detachView: [], _loadComponent: [], _unloadComponent: [], ngOnDestroy: [], viewCount: [] });
    return ApplicationRef;
})();
```
- this PR
```
let ApplicationRef = /*@__PURE__*/ (() => {
    class ApplicationRef {
      // lots of code
    }
    ApplicationRef.ɵfac = function ApplicationRef_Factory(t) { return new (t || ApplicationRef)(ɵɵinject(NgZone), ɵɵinject(Console), ɵɵinject(Injector), ɵɵinject(ErrorHandler), ɵɵinject(ComponentFactoryResolver), ɵɵinject(ApplicationInitStatus)); };
    ApplicationRef.ɵprov = /*@__PURE__*/ ɵɵdefineInjectable({ token: ApplicationRef, factory: function (t) { return ApplicationRef.ɵfac(t); }, providedIn: null });
    /**
     * \@internal
     */
    ApplicationRef._tickScope = /*@__PURE__*/ wtfCreateScope('ApplicationRef#tick()');
    return ApplicationRef;
})();
/*@__PURE__*/ /*@__PURE__*/ (function () {
    setClassMetadata(ApplicationRef, [{
            type: Injectable
        }], function () { return [{ type: NgZone }, { type: Console }, { type: Injector }, { type: ErrorHandler }, { type: ComponentFactoryResolver }, { type: ApplicationInitStatus }]; }, { constructor: [], _zone: [], _console: [], _injector: [], _exceptionHandler: [], _componentFactoryResolver: [], _initStatus: [], _bootstrapListeners: [], _views: [], _runningTick: [], _enforceNoNewChanges: [], _stable: [], componentTypes: [], components: [], bootstrap: [], tick: [], attachView: [], detachView: [], _loadComponent: [], _unloadComponent: [], ngOnDestroy: [], viewCount: [] });
})();
```

We can see two interesting things: `setClassMetadata` is wrapped in an IIFE (as the PR name suggests), and it's moved to outside the class (which I didn't assume from the PR name or tests).

Still, from a tree shaking standpoint this looks reasonable enough and should work.

Stepping back from the code change in `ApplicationRef` gives more insight though. Whereas the bundle for master contains 14 references to `setClassMetadata`, the bundle for this PR contains 77 references. This is very surprising.

Trying to narrow it down a bit further shows that this PR actually contains the same number of calls as master in the `setClassMetadata(ApplicationModule, [{` format (as above), but also contains new calls that are not present in master:

- master
```
let HammerModule = /*@__PURE__*/ (() => {
    class HammerModule {
    }
    HammerModule.ɵmod = /*@__PURE__*/ _angular_core__WEBPACK_IMPORTED_MODULE_1__["ɵɵdefineNgModule"]({ type: HammerModule });
    HammerModule.ɵinj = /*@__PURE__*/ _angular_core__WEBPACK_IMPORTED_MODULE_1__["ɵɵdefineInjector"]({ factory: function HammerModule_Factory(t) { return new (t || HammerModule)(); }, providers: HAMMER_PROVIDERS__PRE_R3__ });
    return HammerModule;
})();
```
- this PR
```
let HammerModule = /*@__PURE__*/ (() => {
    class HammerModule {
    }
    HammerModule.ɵmod = /*@__PURE__*/ _angular_core__WEBPACK_IMPORTED_MODULE_1__["ɵɵdefineNgModule"]({ type: HammerModule });
    HammerModule.ɵinj = /*@__PURE__*/ _angular_core__WEBPACK_IMPORTED_MODULE_1__["ɵɵdefineInjector"]({ factory: function HammerModule_Factory(t) { return new (t || HammerModule)(); }, providers: HAMMER_PROVIDERS__PRE_R3__ });
    return HammerModule;
})();
/*@__PURE__*/ /*@__PURE__*/ (function () {
    _angular_core__WEBPACK_IMPORTED_MODULE_1__["ɵsetClassMetadata"](HammerModule, [{
            type: _angular_core__WEBPACK_IMPORTED_MODULE_1__["NgModule"],
            args: [{ providers: HAMMER_PROVIDERS__PRE_R3__ }]
        }], null, null);
})();
```

The `_angular_core__WEBPACK_IMPORTED_MODULE_1__["ɵsetClassMetadata"]` format might seem a bit foreign but it just means that webpack had to get the `ɵsetClassMetadata` export from another module using the webpack module loader. Bear in mind that in webpack a single bundle/chunk (a file on disk) can contain many modules isolated via closures.

This indirection explains why the size increase manifested itself only on the lazy load scenario: when there is no lazy loading and the code is mostly ES2015, webpack can do a single big module. But when lazy loading is present, code that is common to both eager and lazy loaded modules must be separated into another module, resulting in several modules total. This extra code was also present in non-lazy builds but it was stripped away by tree shaking because there was no module loading indirection.

As to why master does not have these calls, the answer is that it actually does have them but Build Optimizer removes them (see [here](https://github.com/angular/angular-cli/blob/9cd09e6523c562c26502186929098bea6a8f9c07/packages/angular_devkit/build_optimizer/src/transforms/scrub-file_spec.ts#L718-L737). Making a build on master without Build Optimizer (`ng build --prod --optimization false --build-optimizer false`) shows the following output:
```
class HammerModule {
}
HammerModule.ɵmod = _angular_core__WEBPACK_IMPORTED_MODULE_1__["ɵɵdefineNgModule"]({ type: HammerModule });
HammerModule.ɵinj = _angular_core__WEBPACK_IMPORTED_MODULE_1__["ɵɵdefineInjector"]({ factory: function HammerModule_Factory(t) { return new (t || HammerModule)(); }, providers: HAMMER_PROVIDERS__PRE_R3__ });
/*@__PURE__*/ _angular_core__WEBPACK_IMPORTED_MODULE_1__["ɵsetClassMetadata"](HammerModule, [{
        type: _angular_core__WEBPACK_IMPORTED_MODULE_1__["NgModule"],
        args: [{ providers: HAMMER_PROVIDERS__PRE_R3__ }]
    }], null, null);
```

The resolution here is that Build Optimizer must still, unfortunately, identify and remove these calls to prevent cross-module references that are problematic for the lazy load case.

Since the point of the change in this PR was to allow Build Optimizer to not do this processing, and now we have ascertained that this isn't possible anyway, the importance of this change is reduced and probably even eliminated. For the sake of correctness though, this really should be an IIFE since the `/*@__PURE__*/` comment does regarding the argument function calls otherwise.

petebacondarwin:
I have another PR that moves these statements even further away from the class - https://github.com/angular/angular/pull/33689 - this ensures that these statements are after calls like `__decorate`.

IgorMinar:
@filipesilva can you please clarify what causes the following code to be retained?

```
/*@__PURE__*/ /*@__PURE__*/ (function () {
    _angular_core__WEBPACK_IMPORTED_MODULE_1__["ɵsetClassMetadata"](HammerModule, [{
            type: _angular_core__WEBPACK_IMPORTED_MODULE_1__["NgModule"],
            args: [{ providers: HAMMER_PROVIDERS__PRE_R3__ }]
        }], null, null);
})();
```

I thought terser would remove any top level function invocations that were prefixed with `__PURE__`.

filipesilva:
@IgorMinar that code will indeed be removed by Terser. All the samples I gave were with optimization turned off. 

But since the failing test is an app with lazy loaded chunks, by the time Terser runs, the "damage" has been done: sections of angular/core have been de-optimized into their own module and hooked up to the webpack module loader. 

Terser is able to remove the `__PURE__` calls, but will not cross module boundaries and remove the code the `__PURE__` calls caused to be retained and exported in the other modules, name angular/core.

filipesilva:
Build Optimizer update in https://github.com/angular/angular-cli/pull/16228

IgorMinar:
@filipesilva you wrote:
> But since the failing test is an app with lazy loaded chunks, by the time Terser runs, the "damage" has been done: sections of angular/core have been de-optimized into their own module and hooked up to the webpack module loader.

that's correct, and this PR doesn't change anything about it. One way or another `setClassMetadata` will be retained by webpack in the main bundle (and with it anything that this function depends on). The only way to avoid that would be to not generate the iifes at all when we don't need the AOT compiled code to interact with JIT compiled code (this interop is the main purpose of having `setClassMetadata` calls emitted).

have we considered an option to guard this iife emit via a `angularCompilerOptions` flag which would be off by default and set to on by the CLI when we are compiling the tsconfig.app.json program for unit tests?

filipesilva:
@IgorMinar this PR changes the format of the `setClassMetadata` call. In the version of the CLI used in the integration tests, Build Optimizer recognizes the previous format of `setClassMetadata` and removed it when loading the files into webpack. That way it never got in the main bundle.

Since the format is now different, Build Optimizer does not recognize it, and it is left in place to later affect bundling.

IgorMinar:
@filipesilva I think we keep on talking cross each other - I'm sorry if this is confusing/frustrating.

the confusing part is that we are making a change that actually has no impact on CLI users because BO already handles this well.

The bottom line is that in order to prevent the retention of the symbol in the main bundle due to cross-chunk retention, we can't rely on terser and PURE annotations to strip the setClassMetadata callsites from generated code. BO *must* do it *in the webpack loader* in order to prevent the retention down the build pipeline.

What threw me off is that we are talking about IIFEs and PURE and that's completely irrelevant in the case of CLI with BO.

So the last question I have is - why are we making this change at all? it doesn't buy us anything. The initial motivation was our hope that we would be able remove the pass from BO, but that's can't happen because of cross-chunk retention which must be dealt with in the loader.

Should we drop close this PR as useless? Am I missing some hidden benefit that is not obvious?

If we do believe that we want to keep this PR, then we should update the commit message and state the benefit, because the currently stated reason is invalid (due to the cross-chunk issue).

filipesilva:
I think the answer to that is this bit:

> Since the point of the change in this PR was to allow Build Optimizer to not do this processing, and now we have ascertained that this isn't possible anyway, the importance of this change is reduced and probably even eliminated.
> 
> For the sake of correctness though, this really should be an IIFE since the /*@__PURE__*/ comment does nothing regarding the argument function calls otherwise. Rollup is able to do away with the setClassMetadata wrapped in the IIFE without Build Optimizers help (as can be seen by cli-hello-world-lazy-rollup not failing), but Webpack cannot.

IgorMinar:
@filipesilva ok. that's fair. let's go ahead with this then.

IgorMinar:
I've just rebased this so that we get the latest numbers. @alxhub or I will update this PR tomorrow to get it in.


we should update the commit message by stripping the last sentence about BO which is wrong and replace it with a clarification along the lines:

>Without the IIFE, terser won't remove these function calls because the function calls have arguments that themselves are function calls or other impure expressions. In order to make the whole block away, we wrap it into IIFE and mark the IIFE as pure.
>
>It should be noted that this change doesn't have any impact on CLI with build-optimizer, which removes the whole `setClassMetadata` block within the webpack loader, so terser or webpack itself don't get to see it at all. This is done to prevent cross-chunk retention issues caused by webpack's internal module registry.*

IgorMinar:
I updated the commit message

alxhub:
[Presubmit](http://test/OCL:281566246:BASE:281566306:1574278810611:86c9bd6c)
[Ivy Presubmit](http://test/OCL:281566246:BASE:281566484:1574278851590:57c7a54b)

aio/scripts/_payload-limits.json
```diff
@@ -12,7 +12,7 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 2987,
-        "main-es2015": 461159,
+        "main-es2015": 506857,
```
gkalpak: I know this is temporary, but 😱 😱 😱

gkalpak (approved):

PR #33339, opened 2019-10-23
docs: add migrating to version 9 guide

kara:
This commit adds a guide to AIO navigation for
"Migrating to Version 9" and moves the schematics
section into the guide that previously lived in
the deprecations page. It also pastes a snippet
of the deprecations page in the new guide so users
don't have to filter out deprecations they've seen
before.

Note: Ivy compatibility section is coming up in a
follow-up PR.

### DIFF
--- a/aio/content/navigation.json
+++ b/aio/content/navigation.json
@@ -704,6 +704,11 @@
           "tooltip": "Angular versioning, release, support, and deprecation policies and practices."
         },
         {
+          "url": "guide/migrating-to-version-9",
+          "title": "Migrating to Version 9",
+          "tooltip": "Support for migrating your application from version 8 to 9."
+        },
+        {
           "url": "guide/deprecations",
           "title": "Deprecations",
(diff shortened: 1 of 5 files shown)


aio/content/navigation.json
```diff
@@ -703,6 +703,11 @@
           "title": "Release Practices",
           "tooltip": "Angular versioning, release, support, and deprecation policies and practices."
         },
+        {
+          "url": "guide/migrating-to-version-9",
+          "title": "Migrating to Version 9",
```
IgorMinar:
```suggestion
          "title": "Updating to Version 9",
```

IgorMinar (approved):
overall lgtm, but to be consistent with `ng update` terminology, can you please update the guide url/filename/title to be called "Updating" instead of "migrating"

aio/content/navigation.json
kara: Done

PR #33340, opened 2019-10-23
test(core): support running performance benchmarks in browser

mhevery:


pkozlowski-opensource (changes requested):
@mhevery there are still other places where we need to change Noop renderer for a DOM-based one - I've created a fixup commit here: https://github.com/pkozlowski-opensource/angular/commit/1c676033247f3a18c896aea74bd536f93256309f - feel free to pick it up or do alternative changes.

mhevery:
@pkozlowski-opensource thanks for the fixup, but I think we are still missing something. The sub `LView`s  don't seem to get added to the host `<div>` I think we are missing a DOM manipulation someplace?

pkozlowski-opensource (approved):
I think that we should merge it as-is to unblock Matias - I know that the nodes are not connected (parent -> child) properly but at least they are created so we should have relevant numbers for styling.

I will follow up with the nodes connecting PR.

PR #33341, opened 2019-10-23
test(ivy): fined tune micro benchmark parameters

mhevery:
Decrease `MIN_SAMPLE_DURATION` to make it more likely that we cane fit into single time slice.
Increase `MIN_SAMPLE_COUNT_NO_IMPROVEMENT` to make it more likely to find the best


xiangdewei:
How do you prove that your number change has performance improvement? what's the difference?

mhevery:
> How do you prove that your number change has performance improvement? what's the difference?

this is what we used for perf testing. We run the same benchmark over and see if the variation in results improves.

AndrewKushnir (approved):

PR #33347, opened 2019-10-23
fix(compiler): do not throw when using abstract directive from other compilation unit

devversion:
Libraries can expose directive/component base classes that will be
used by consumer applications. Using such a base class from another
compilation unit works fine with "ngtsc", but when using "ngc", the
compiler will thrown an error saying that the base class is not
part of a NgModule. e.g.

```
Cannot determine the module for class X in Y! Add X to the NgModule to fix it.
```

This seems to be because the logic for distinguishing directives from
abstract directives is scoped to the current compilation unit within
ngc. This causes abstract directives from other compilation units to
be considered as actual directives (causing the exception).

### DIFF
--- a/packages/compiler-cli/test/ngc_spec.ts
+++ b/packages/compiler-cli/test/ngc_spec.ts
@@ -2308,5 +2308,32 @@ describe('ngc transformer command-line', () => {
       let exitCode = main(['-p', path.join(basePath, 'tsconfig.json')], errorSpy);
       expect(exitCode).toEqual(0);
     });
+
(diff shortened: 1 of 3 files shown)


alxhub (approved):

devversion:
Caretaker note: Can you please run presubmit for this PR to get it green? Thanks!

AndrewKushnir:
- [VE Presubmit](http://test/OCL:276308591:BASE:276308644:1571853370897:be404c05)
- [Ivy Presubmit](http://test/OCL:276308591:BASE:276308792:1571853411724:aaa24b62)

PR #33366, opened 2019-10-23
fix(compiler-cli): package.json writing graceful failure on readonly …

alexeagle:
…sources

This is throwing when the compiler runs under Bazel.

### DIFF
--- a/packages/compiler-cli/ngcc/src/writing/package_json_updater.ts
+++ b/packages/compiler-cli/ngcc/src/writing/package_json_updater.ts
@@ -136,7 +136,14 @@ export class DirectPackageJsonUpdater implements PackageJsonUpdater {
     // Ensure the containing directory exists (in case this is a synthesized `package.json` due to a
     // custom configuration) and write the updated content to disk.
     this.fs.ensureDir(dirname(packageJsonPath));
-    this.fs.writeFile(packageJsonPath, `${JSON.stringify(parsedJson, null, 2)}\n`);
+    try {
+      this.fs.writeFile(packageJsonPath, JSON.stringify(parsedJson, null, 2) + '\n');
+    } catch (e) {
+      // Under Bazel, the input filesystem is readonly
+      if (e.code === 'EROFS') {
+        console.error('[ivy-ngcc] Readonly filesystem: Cannot write changes to', packageJsonPath);
+      }
+    }
   }
 }
 


packages/compiler-cli/ngcc/src/writing/package_json_updater.ts
```diff
@@ -136,7 +136,14 @@ export class DirectPackageJsonUpdater implements PackageJsonUpdater {
     // Ensure the containing directory exists (in case this is a synthesized `package.json` due to a
     // custom configuration) and write the updated content to disk.
     this.fs.ensureDir(dirname(packageJsonPath));
-    this.fs.writeFile(packageJsonPath, `${JSON.stringify(parsedJson, null, 2)}\n`);
+    try {
+      this.fs.writeFile(packageJsonPath, JSON.stringify(parsedJson, null, 2) + '\n');
+    } catch (e) {
+      // Under Bazel, the input filesystem is readonly
+      if (e.code === 'EROFS') {
+        console.error('[ivy-ngcc] Readonly filesystem: Cannot write changes to', packageJsonPath);
```
gkalpak: `ivy-ngcc` has been renamed to `ngcc`.

gkalpak: - `ivy-ngcc` has been renamed to `ngcc`.
- This should be using the [logger][1].

[1]: https://github.com/angular/angular/blob/a1d7b6bb8/packages/compiler-cli/ngcc/src/main.ts#L128

gkalpak (changes requested):
The commit message scope should be `ngcc` (i.e. `fix(ngcc): ...`).

And I agree with Pete :grin:
`ngcc` can fail in more ways if run on a read-only file-system, so we should have a more "universal" solution.

alexeagle:
I'm running `@angular-devkit/build-angular` which has the ngcc integration.
@alan-agius4 maybe you could point me to how to run build-angular on a read-only filesystem?

I think it's `@ngtools/webpack/src/ngcc_processor` that does
```
    mainNgcc({
      basePath: this._nodeModulesDirectory,
      targetEntryPointPath: path.dirname(packageJsonPath),
      propertiesToConsider: this.propertiesToConsider,
      compileAllFormats: false,
      createNewEntryPointFormats: true,
      logger: this._logger,
      pathMappings: this._pathMappings,
    });
```
should this have an option? I do want to build Ivy artifacts but am on a read-only FS under Bazel.

alan-agius4:
@alexeagle, At the moment that is not possible. Currently NGCC only works with the actual FS and not a virtual one. And hence it’s cannot be read only as it needs to write the transpiled VE -> IVY node_modules to disk.

If under Bazel the FS is read only, will you be transpilling the node_modules using the ngcc binary via a post install hook?

alan-agius4:
If you run the ngcc binary prior to triggering the build-angular build the ngcc in ngtools shouldn’t write any files. That is because ngcc will find that the module is already processed and it will bail out.

alexeagle:
Ah well that is not working - we did run ngcc as a postinstall hook

alan-agius4:
Do you have a repo that I can look at? 

My thoughts around this without seeing any reproduction so I am really speculating right now but it can be that during ngcc postinstall hook a an entrypoint is transformed ex: main, but during the build-angular build another format is requested ex: module.

alexeagle:
Yeah simple: https://github.com/alexeagle/prototype_bazel_architect/blob/master/BUILD.bazel

note this patch is included in that repo so you should delete it first to repro.

alan-agius4:
When not under Bazel such behaviour is not desired, because ngtools will use ngcc and transform the correct module.

This is because;
1) Ngtools integration with ngcc is designed to work by itself, and without the need of the post install hook.
2) A package module might be transformed with ngcc via a the postinstall hook to FESM2015 which will be used for build but later another format might be required ex UMD for universal.

I will try to debug your repo on Sun/Mon. But at first glace by looking at the repo nothing stands out why ngcc in ngtools webpack is being trigger and causes a file write. The post install hook that you have configure, is without any argument which means that it will transform all module formats of all packages.

alan-agius4:
@alexeagle I just looked into this and I think I know what is happening.

When calling NGCC binary via the post install hook it running on certain packages for a reason or another not all. Ex: `rxjs`, `tslib` are not being processed nor marked that they were visited by NGCC.

However, when calling NGCC via the API, ngtools will call for each and every module that TSC resolves, hence it will call it for `rxjs`, `tslib` and so on. When a module is visited by NGCC for the first time it will try to create a marker which under Bazel fails because it is read only.

Question to @petebacondarwin / @gkalpak, are the above mentioned modules skipped from processing when calling NGCC bin for some reason?

From ngtools perspective probably makes sense that when running under Bazel we don't call NGCC, since this won't work.

That being said I didn't manage to get past this error even after I disable NGCC:

```bash
ERROR in /workspace/prototype_bazel_architect/src/main.ts
Module build failed (from ./node_modules/@ngtools/webpack/src/index.js):
Error: /workspace/prototype_bazel_architect/src/main.ts is missing from the TypeScript compilation. Please make sure it is in your tsconfig via the 'files' or 'include' property.
```

alan-agius4:
Bdw in my cases, ngcc under ngtools with bazel did seem to transform the files without any exception

```
gitpod /workspace/prototype_bazel_architect $ cat node_modules/rxjs/package.json | grep ivy
gitpod /workspace/prototype_bazel_architect $ yarn build
yarn run v1.17.3
$ bazel build :build
INFO: Writing tracer profile to '/home/gitpod/.cache/bazel/_bazel_gitpod/9ee99402aa38cb79ec57f31ff560e386/command.profile.gz'
INFO: Analyzed target //:build (1 packages loaded, 4 targets configured).
INFO: Found 1 target...
ERROR: /workspace/prototype_bazel_architect/BUILD.bazel:3:1: Action build failed (Exit 1) architect.sh failed: error executing command bazel-out/host/bin/external/npm/@angular-devkit/architect-cli/bin/architect.s
h frontend:build '--outputPath=bazel-out/k8-fastbuild/bin/build' ... (remaining 2 argument(s) skipped)

Use --sandbox_debug to see verbose messages from the sandbox
(node:38960) ExperimentalWarning: The fs.promises API is experimental

frontend:build: 
ERROR in /workspace/prototype_bazel_architect/src/main.ts
Module build failed (from ./node_modules/@ngtools/webpack/src/index.js):
Error: /workspace/prototype_bazel_architect/src/main.ts is missing from the TypeScript compilation. Please make sure it is in your tsconfig via the 'files' or 'include' property.
    at AngularCompilerPlugin.getCompiledFile (/home/gitpod/.cache/bazel/_bazel_gitpod/9ee99402aa38cb79ec57f31ff560e386/sandbox/processwrapper-sandbox/7/execroot/angular_cli_demo/node_modules/@ngtools/webpack/src/
angular_compiler_plugin.js:918:23)
    at plugin.done.then (/home/gitpod/.cache/bazel/_bazel_gitpod/9ee99402aa38cb79ec57f31ff560e386/sandbox/processwrapper-sandbox/7/execroot/angular_cli_demo/node_modules/@ngtools/webpack/src/loader.js:41:31)
    at process._tickCallback (internal/process/next_tick.js:68:7)
 @ multi ./src/main.ts main[0]
ERROR in /workspace/prototype_bazel_architect/src/polyfills.ts
Module build failed (from ./node_modules/@ngtools/webpack/src/index.js):
Error: /workspace/prototype_bazel_architect/src/polyfills.ts is missing from the TypeScript compilation. Please make sure it is in your tsconfig via the 'files' or 'include' property.
    at AngularCompilerPlugin.getCompiledFile (/home/gitpod/.cache/bazel/_bazel_gitpod/9ee99402aa38cb79ec57f31ff560e386/sandbox/processwrapper-sandbox/7/execroot/angular_cli_demo/node_modules/@ngtools/webpack/src/angular_compiler_plugin.js:918:23)
    at plugin.done.then (/home/gitpod/.cache/bazel/_bazel_gitpod/9ee99402aa38cb79ec57f31ff560e386/sandbox/processwrapper-sandbox/7/execroot/angular_cli_demo/node_modules/@ngtools/webpack/src/loader.js:41:31)
    at process._tickCallback (internal/process/next_tick.js:68:7)
 @ multi ./node_modules/@angular-devkit/build-angular/src/angular-cli-files/models/es5-polyfills.js zone.js/dist/zone-legacy ./node_modules/@angular-devkit/build-angular/src/angular-cli-files/models/jit-polyfills.js ./node_modules/@angular-devkit/build-angular/src/angular-cli-files/models/es5-jit-polyfills.js ./src/polyfills.ts polyfills-es5[4]
Exiting with code  1 Error
    at process.exit (/home/gitpod/.cache/bazel/_bazel_gitpod/9ee99402aa38cb79ec57f31ff560e386/sandbox/processwrapper-sandbox/7/execroot/angular_cli_demo/node_modules/@angular-devkit/architect-cli/bin/architect.js:75:51)
    at main.then.code (/home/gitpod/.cache/bazel/_bazel_gitpod/9ee99402aa38cb79ec57f31ff560e386/sandbox/processwrapper-sandbox/7/execroot/angular_cli_demo/node_modules/@angular-devkit/architect-cli/bin/architect.js:165:13)
    at process._tickCallback (internal/process/next_tick.js:68:7)
FAILURE
Result: {
    "success": false,
    "outputPath": "/home/gitpod/.cache/bazel/_bazel_gitpod/9ee99402aa38cb79ec57f31ff560e386/sandbox/processwrapper-sandbox/7/execroot/angular_cli_demo/bazel-out/k8-fastbuild/bin/build",
    "target": {
        "project": "frontend",
        "target": "build"
    }
}

Logs:
Target //:build failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 60.637s, Critical Path: 40.92s
INFO: 0 processes.
FAILED: Build did NOT complete successfully
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
gitpod /workspace/prototype_bazel_architect $ cat node_modules/rxjs/package.json | grep ivy
  "__processed_by_ivy_ngcc__": {
```

petebacondarwin:
In standalone mode ngcc only processes packages that have been compiled via ViewEngine - the primary market being a metadata.json file next to the typings file.

alan-agius4:
Umm, wouldn’t it make sense to keep the same logic for the API? or it would be slower then the marker in package.json?, I am guessing there shouldn’t be any performance hit since instead of basing the logic solely on the ngcc marker, check also for the presence of the metadata field. If no metadata field is present just noop. Unless you checking for the actual existence of the file instead of using the field which will cause additional IO operations.

That being said, if under bazel the FS is always read-only, there is no point in calling ngcc in the first place. Thus, ngtools should not invoke it if maybe we find a BAZEL_TARGET env variable.

petebacondarwin:
I definitely think we should not be invoking ngcc on a read only fs

gkalpak:
FWIW, I think the difference in ngcc behavior is not related to standalone vs API, but to whether is it called with `targetEntryPointPath` or not (which is the case when calling it from ngtools but not for the recommended postinstall script).

When called with `targetEntryPointPath`, if that target happens to have no processable entry-points, ngcc will mark is as processed in `package.json` (I assume for performance reasons).

kylecordes:
I don't know if there is a Node community consensus around it, but I propose the following guideline:

* If there is executable code necessary to finish preparing `node_modules` for downstream use, it **is appropriate** to execute this code from a post-install hook.
* When a Node process is running other than during a post install hook, it **is not appropriate** to attempt to write to `node_modules`, nor to care whether it is a real file system.

To follow that with Angular though, would mean choosing somewhere new to store `ngcc` state and output - maybe a change larger than anyone wants to contemplate.

Yet without something in this direction, is build hermeticity is possible? Every concurrent compilation will have `node_modules` as a pool of shared mutable storage with race conditions and all other manner of excitement possible.

petebacondarwin:
I believe that CLI no longer runs ngcc on read-only filesystems.
For command line use of ngcc, it is up to the process calling it to decide whether the system is readonly or not.

In which case, I think we can close this PR...

---

@kylecordes - regarding your recommendation, I agree but it is not something we can do right now since we need to make life easier for the majority of CLI users. The plan, I believe, for the future is to leverage webpack 5 file caching to avoid writing to the node_modules folder. But this is some way off.

alexeagle:
@petebacondarwin I think that's right, closing

PR #33367, opened 2019-10-24
build: update to @bazel/bazel 1.0.0

gregmagolan:
Also removes `build:remote --spawn_strategy=remote` from .bazelrc. It seems that with Bazel 1.0.0 setting `--incompatible_list_based_execution_strategy_selection=false` no longer works around the issue with npm_package that it did when it was added. The error that was originally observed has returned after updating to Bazel 1.0.0:

```
ERROR: /home/circleci/ng/packages/angular_devkit/build_optimizer/BUILD:66:1: Assembling npm package packages/angular_devkit/build_optimizer/npm_package failed: No usable spawn strategy found for spawn with mnemonic Action. Your --spawn_strategy, --genrule_strategy or --strategy flags are probably too strict. Visit bazelbuild/bazel#7480 for migration advice
```

This commit removes both `—incompatible_list_based_execution_strategy_selection=false` as well as `build:remote --spawn_strategy=remote` which means that Bazel will do the default behavior of picking the first available strategy from the default list, which is `remote,worker,sandboxed,local`. See bazelbuild/bazel#7480 for more details.


alexeagle (approved):

AndrewKushnir:
@gregmagolan could you please rebase this PR and resolve a conflict? Thank you.

gregmagolan:
@AndrewKushnir Done

.bazelrc
```diff
@@ -92,8 +92,23 @@ build:remote-http-caching --google_default_credentials
 #  --config=remote               #
 ##################################
 
-# Load default settings for Remote Build Execution.
-import %workspace%/third_party/github.com/bazelbuild/bazel-toolchains/bazelrc/.bazelrc.notoolchain
+# Starting with Bazel 0.27.0 strategies do not need to be explicitly
```
josephperrott: I think we should be fine to not comment on why we no longer have something in the code.  As this is now the standard, it will become more obsolete as a comment as time goes on.  Unless I am misunderstanding the situation and issue that you reference in this comment.

.bazelrc
```diff
@@ -92,8 +92,23 @@ build:remote-http-caching --google_default_credentials
...
+# defined. See https://github.com/bazelbuild/bazel/issues/7480
+build:remote --define=EXECUTOR=remote
```
josephperrott: We should note that this `--define=EXECUTOR=remote` will be able to be removed once `https://github.com/bazelbuild/bazel/issues/7254` is fixed.

.bazelrc
```diff
@@ -92,8 +92,23 @@ build:remote-http-caching --google_default_credentials
...
+
+# Enable remote execution so actions are performed on the remote systems.
+build:remote --remote_executor=remotebuildexecution.googleapis.com
+
+# Enable encryption.
+build:remote --tls_enabled=true
```
josephperrott: We should be able to remove this as `--tls_enabled` is a noop: https://docs.bazel.build/versions/1.0.0/command-line-reference.html#flag--tls_enabled

packages/bazel/package.bzl
```diff
@@ -29,9 +29,9 @@ def rules_angular_dev_dependencies():
     _maybe(
         http_archive,
         name = "bazel_toolchains",
-        sha256 = "dcb58e7e5f0b4da54c6c5f8ebc65e63fcfb37414466010cf82ceff912162296e",
-        strip_prefix = "bazel-toolchains-0.28.2",
-        url = "https://github.com/bazelbuild/bazel-toolchains/archive/0.28.2.tar.gz",
+        sha256 = "0b36eef8a66f39c8dbae88e522d5bbbef49d5e66e834a982402c79962281be10",
+        strip_prefix = "bazel-toolchains-1.0.1",
```
josephperrott: Should this be toolchains `1.0.0` since we are going to bazel `1.0.0`?

.bazelrc
```diff
@@ -92,8 +92,23 @@ build:remote-http-caching --google_default_credentials
...
 ##################################
 
-# Load default settings for Remote Build Execution.
-import %workspace%/third_party/github.com/bazelbuild/bazel-toolchains/bazelrc/.bazelrc.notoolchain
+# Starting with Bazel 0.27.0 strategies do not need to be explicitly
+# defined. See https://github.com/bazelbuild/bazel/issues/7480
+build:remote --define=EXECUTOR=remote
+
+# Enable remote execution so actions are performed on the remote systems.
+build:remote --remote_executor=remotebuildexecution.googleapis.com
```
josephperrott: Can we move this remote_executor flag to line 134 to be located with the `--remote_cache` and `--remote_instance_name` flags?

.bazelrc
```diff
@@ -92,8 +92,23 @@ build:remote-http-caching --google_default_credentials
...
+
+# Enable encryption.
+build:remote --tls_enabled=true
+
+# Set a higher timeout value, just in case.
+build:remote --remote_timeout=3600
```
josephperrott: I know this is what we had previously coming from the our `/third_party/..` code, but 3600 seconds seems excessive.  If our remote builds either for execution and/or cache hits are taking an hour, I would rather fail earlier.

It seems to me that this makes more sense to be set as something like 600 (10 minutes) which is probably still longer than we are truly okay with it taking.

.bazelrc
```diff
@@ -92,8 +92,23 @@ build:remote-http-caching --google_default_credentials
...
+
+# Enable authentication. This will pick up application default credentials by
+# default. You can use --auth_credentials=some_file.json to use a service
+# account credential instead.
+build:remote --auth_enabled=true
```
josephperrott: This can be removed as it is actually already enabled with `--google_default_credentials`.  `auth_enabled` is the old name of the flag: https://github.com/bazelbuild/bazel/blob/d2de8ee925710b757a637bc8ad1036546a030756/src/main/java/com/google/devtools/build/lib/authandtls/AuthAndTLSOptions.java#L30

josephperrott (commented):
Thanks for doing all this update!  A few comments, mostly about making sure that we are applying active flags since the `.bazelrc.notoolchain` file alot of these came from was pretty dated.

josephperrott (changes requested):
Thanks for doing all this update! A few comments, mostly about making sure that we are applying active flags since the .bazelrc.notoolchain file alot of these came from was pretty dated.

gregmagolan:
Thanks for the detailed review @josephperrott! Updating...

packages/bazel/package.bzl
gregmagolan: Its sometimes off by a patch version. https://github.com/bazelbuild/bazel-toolchains/releases/tag/1.0.1 is still for Bazel 1.0.0

.bazelrc
gregmagolan: 👍 You have it correct and it will become obsolete. The comment comes from https://github.com/bazelbuild/bazel-toolchains/blob/master/bazelrc/bazel-1.0.0.bazelrc#L45 but in this context (which is not versioned) I think you're right that it is better to leave it out as it will become dated.

.bazelrc
gregmagolan: 👍

.bazelrc
gregmagolan: 👍

.bazelrc
gregmagolan: 👍

.bazelrc
gregmagolan: 👍

.bazelrc
gregmagolan: 👍

josephperrott (approved):
LGTM

gregmagolan:
Note: this was reverted because of a Windows CI failure with Bazel 1.0.0

https://app.circleci.com/jobs/github/angular/angular/501770:

```
	at java.base/java.lang.Thread.run(Unknown Source)
---8<---8<--- End of exception details ---8<---8<---
---8<---8<--- Start of log, file at C:/users/circleci/_bazel_circleci/u4uoan2j/bazel-workers/worker-2-TypeScriptCompile.log ---8<---8<---
Error: Cannot find module 'npm/node_modules/@bazel/typescript/internal/tsc_wrapped/tsc_wrapped.js'. Please verify that the package.json has a valid "main" entry
    at Function.module.constructor._resolveFilename (C:\users\circleci\_bazel_circleci\u4uoan2j\execroot\angular\bazel-out\host\bin\external\npm\@bazel\typescript\bin\tsc_wrapped_loader.js:485:17)
    at Function.Module._load (internal/modules/cjs/loader.js:562:25)
    at Object.<anonymous> (C:\users\circleci\_bazel_circleci\u4uoan2j\execroot\angular\bazel-out\host\bin\external\npm\@bazel\typescript\bin\tsc_wrapped_loader.js:524:24)
    at Module._compile (internal/modules/cjs/loader.js:776:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:787:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
    at Function.Module._load (internal/modules/cjs/loader.js:585:3)
    at Function.Module.runMain (internal/modules/cjs/loader.js:829:12)
    at startup (internal/bootstrap/node.js:283:19)
---8<---8<--- End of log ---8<---8<---
```

Investigating...

gregmagolan:
NOTE: I attempted to update to @bazel/bazel 1.1.0 in this PR but it created another issue in CI with docker access permissions:

ERROR: An error occurred during the fetch of repository 'rbe_ubuntu1604_angular':
   Traceback (most recent call last):
    File "/home/circleci/.cache/bazel/_bazel_circleci/9ce5c2144ecf75d11717c0aa41e45a8d/external/bazel_toolchains/rules/rbe_repo.bzl", line 449
        validate_host(ctx)
    File "/home/circleci/.cache/bazel/_bazel_circleci/9ce5c2144ecf75d11717c0aa41e45a8d/external/bazel_toolchains/rules/rbe_repo/util.bzl", line 161, in validate_host
        fail(<2 more arguments>)
attribute Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
: Cannot run rbe_autoconfig as running '%s ps' returned a non 0 exit code, please check you have permissions to run docker. Error message: /usr/bin/docker
ERROR: /home/circleci/ng/tools/BUILD.bazel:16:1: //tools:rbe_ubuntu1604-angular depends on @rbe_ubuntu1604_angular//config:platform in repository @rbe_ubuntu1604_angular which failed to fetch. no such package '@rbe_ubuntu1604_angular//config': Traceback (most recent call last):
    File "/home/circleci/.cache/bazel/_bazel_circleci/9ce5c2144ecf75d11717c0aa41e45a8d/external/bazel_toolchains/rules/rbe_repo.bzl", line 449
        validate_host(ctx)
    File "/home/circleci/.cache/bazel/_bazel_circleci/9ce5c2144ecf75d11717c0aa41e45a8d/external/bazel_toolchains/rules/rbe_repo/util.bzl", line 161, in validate_host
        fail(<2 more arguments>)
attribute Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
: Cannot run rbe_autoconfig as running '%s ps' returned a non 0 exit code, please check you have permissions to run docker. Error message: /usr/bin/docker
ERROR: While resolving toolchains for target //packages/platform-webworker:npm_package: Target //tools:rbe_ubuntu1604-angular was referenced as a platform, but does not provide PlatformInfo
ERROR: Analysis of target '//packages/platform-webworker:npm_package' failed; build aborted: Analysis failed
Looks like due to a change in rbe_autoconfig with bazel_tools 1.1.0. Will leave that battle for another day.

PR #19710, opened 2017-10-13
fix(common): allow URL parameters to be encoded

tiangolo:
Back ends for several languages (including Python, PHP and Node.js) expect to receive encoded parameters with some standard conventions, when they receive them as raw decoded strings they can't parse those parameters. This removes the custom decoding of parameters to raw strings and leaves the default standard encoding for HttpClient.

## PR Checklist
Please check if your PR fulfills the following requirements:

- [ :heavy_check_mark:  ] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [ N/A ] Tests for the changes have been added (for bug fixes / features)
> I only removed code, and the code I removed was not directly tested.
- [ N/A ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

```
[ x ] Bugfix
[ ] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Documentation content changes
[ ] angular.io application / infrastructure changes
[ ] Other... Please describe:
```

## What is the current behavior?

Currently, when creating URL parameters, some characters are decoded and converted to raw strings after the `encodeURIComponent()` had encoded them. The characters are: `@`, `:`, `$`, `,`, `;`, `+`, `=`, `?`, `/`.

This creates communication problems with back ends in at least Python, PHP and Node.js.

Check the referenced issues for more details on all the possible problems this creates.

Issue Number: #11058 #18884 #18274 #14531 #13077 #18261

## What is the new behavior?

With this PR, when using `HttpParams`, the parameters will be encoded as normally, without being decoded by Angular before sending them in the URL to the server.

I will create a parallel PR that will solve the problem for cases using `URLSearchParams`.

## Does this PR introduce a breaking change?
```
[ ] Yes
[ x ] No
```

...or would you consider this a breaking change?

PR #33370, opened 2019-10-24
refactor(common): change encoding function in httpparams

edodusi:
HttpParams uses custom encoding function "standardEncoding" to encode query string preserving specific charachters. This refactoring aims to improve performances and code quality of that function by using a RegExp and a Map object instead of a chain of replace functions.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Parameters string passes through `encodeURIComponent` and then into a chain of `replace` that un-encodes some characters.
Although it has been pointed out that this behaviour is non-standard and hard to understand for some developers (cfr. #19710) this commit doesn't change the function's output but only improves its internals.

## What is the new behavior?

-

- [ ] Yes
- [x] No


packages/common/http/src/params.ts
```diff
@@ -78,17 +78,22 @@ function paramParser(rawParams: string, codec: HttpParameterCodec): Map<string,
...
+    ['40', '@'],
+    ['3A', ':'],
+    ['24', '$'],
+    ['2C', ','],
+    ['3B', ';'],
+    ['2B', '+'],
+    ['3D', '='],
+    ['3F', '?'],
+    ['2F', '/'],
+  ]);
```
petebacondarwin, line 95: There is no need to use `Map` here you can just use a normal JS object:

```ts
const map = {
  '40': '@',
  '3A': ':',
  '24': '$',
  '2C': ',',
  '3B': ';',
  '2B': '+',
  '3D': '=',
  '3F': '?',
  '2F': '/',
};
```

Also this could be moved to the top level of the file to avoid initializing it on every call to `standardEncoding()`.

packages/common/http/src/params.ts
```diff
@@ -78,17 +78,22 @@ function paramParser(rawParams: string, codec: HttpParameterCodec): Map<string,
...
+  return encodeURIComponent(v).replace(/%(\d[a-z0-9])/gi, (_, t) => tmap.get(t) || `%${t}`);
```
petebacondarwin, line 96: No need to recreate the string since it is passed as the first parameter.

```suggestion
  return encodeURIComponent(v).replace(/%(\d[a-z0-9])/gi, (s, t) => tmap.get(t) ?? s);
```

petebacondarwin (changes requested):
Thanks for this PR - I like the idea to avoid searching the whole string for each of the possible replacements. I think we can tighten it up even more.

edodusi:
Thanks @petebacondarwin, tried yout suggested refactoring and it's fine (also added a bit of typings).
Unfortunately I deleted the fork repo from which I opened this PR (because 2 years 😅), so I don't think I can add any more commits to this PR now. What should I do?

petebacondarwin:
The simplest solution will be to reproduce the PR on a new branch and close this PR.

edodusi:
Yep. Ok, closing this one, will open a new PR and tag you in a comment, since I don't think I can assign or add a reviewer.
Thanks again!

PR #33376, opened 2019-10-24
fix(ivy): fix broken typechecking test on Windows

alxhub:
One of the template type-checking tests relies on the newline character,
which is different on Windows. This commit fixes the issue.


AndrewKushnir (approved):

packages/compiler-cli/src/ngtsc/typecheck/test/type_constructor_spec.ts
```diff
@@ -159,7 +159,7 @@ TestClass.ngTypeCtor({value: 'test'});
         const TestClassWithCtor =
             getDeclaration(res.program, _('/main.ts'), 'TestClass', isNamedClassDeclaration);
         const typeCtor = TestClassWithCtor.members.find(isTypeCtor) !;
-        const ctorText = typeCtor.getText().replace(/[ \n]+/g, ' ');
+        const ctorText = typeCtor.getText().replace(/[ \r\n]+/g, ' ');
```
filipesilva: You might want `\r?\n` instead, so it doesn't break on linux now.

PR #33381, opened 2019-10-24
fix(ivy): i18n - support setting locales for each translation file

### LINKED ISSUE #33323: target-language was not needed in VE but seems to be required on Ivy i18n
# 🐞 bug report

### Affected Package

 The issue is caused by package @angular/localize

### Is this a regression?

 No

### Description

`target-language` was not needed in VE but seems to be required on Ivy.

@clydin looked into this and it seems like the xliff 1.2 parser lists it as optional in http://docs.oasis-open.org/xliff/v1.2/os/xliff-core.html.

But `@angular/localize` will throw if it's not there: https://github.com/angular/angular/blob/ec482dadb19eea5cd247eb7428a5d88640bd38f8/packages/localize/src/tools/src/translate/translation_files/translation_parsers/xliff1/xliff1_translation_parser.ts#L56

@petebacondarwin can you shed some light on this? Is this intended, and if so, is there documentation telling users they must do it?

## 🔬 Minimal Reproduction
The e2e test in https://github.com/angular/angular-cli/pull/15913

## 🔥 Exception or Error
<pre><code>

</code></pre>

## 🌍  Your Environment

**Angular Version:**
<pre><code>
@angular/localize@9.0.0-next.12

</code></pre>

**Anything else relevant?**

petebacondarwin:
Previously the target locale of a translation file had to be extracted
from the contents of the translation file.

Now an array of locales can be provided via the `translationFileLocales`
option that overrides any target locale extracted from the file.
This allows us to support translation files that do not have a target
locale specified in their contents.

// FW-1644
Fixes #33323


packages/localize/src/tools/src/translate/translation_files/translation_file_loader.ts
```diff
@@ -20,12 +20,13 @@ export class TranslationLoader {
...
-  loadBundles(translationFilePaths: string[]): TranslationBundle[] {
-    return translationFilePaths.map(filePath => {
+  loadBundles(translationFilePaths: string[], translationFileLocales: (string|undefined)[]):
+      TranslationBundle[] {
+    return translationFilePaths.map((filePath, index) => {
       const fileContents = FileUtils.readFile(filePath);
       for (const translationParser of this.translationParsers) {
         if (translationParser.canParse(filePath, fileContents)) {
-          return translationParser.parse(filePath, fileContents);
+          return translationParser.parse(filePath, fileContents, translationFileLocales[index]);
```
clydin: Can the logic to set the locale if not present in the parsed file be moved up here after it is parsed?
The CLI already has the locale information so it would have to pass it in and then get it right back to avoid the missing locale exception.

This also prevents the CLI from performing a validity check on the parsed file locale since it will now be wiped by the passed in locale.

petebacondarwin: Ok

packages/localize/src/tools/src/translate/translation_files/translation_parsers/translation_parser.ts
```diff
@@ -25,5 +25,5 @@ export interface TranslationParser {
    * @param filePath The absolute path to the translation file.
    * @param contents The contents of the translation file.
    */
-  parse(filePath: string, contents: string): TranslationBundle;
+  parse(filePath: string, contents: string, locale: string|undefined): TranslationBundle;
```
IgorMinar: why not this? 
```suggestion
  parse(filePath: string, contents: string, locale?: string): TranslationBundle;
```

packages/localize/src/tools/src/translate/translation_files/translation_parsers/xliff1/xliff1_translation_parser.ts
```diff
@@ -30,30 +30,34 @@ export class Xliff1TranslationParser implements TranslationParser {
     return (extname(filePath) === '.xlf') && XLIFF_1_2_NS_REGEX.test(contents);
   }
 
-  parse(filePath: string, contents: string): TranslationBundle {
+  parse(filePath: string, contents: string, locale: string|undefined): TranslationBundle {
```
IgorMinar: same as above

packages/localize/src/tools/test/translate/translation_files/translation_parsers/simple_json/simple_json_spec.ts
```diff
@@ -20,18 +20,27 @@ describe('SimpleJsonTranslationParser', () => {
   describe('parse()', () => {
     it('should extract the locale from the JSON contents', () => {
       const parser = new SimpleJsonTranslationParser();
-      const result = parser.parse('/some/file.json', '{"locale": "en", "translations": {}}');
+      const result =
+          parser.parse('/some/file.json', '{"locale": "en", "translations": {}}', undefined);
```
IgorMinar: you wouldn't need this if you made it optional

IgorMinar (approved):
the rest lgtm

packages/localize/src/tools/src/translate/translation_files/translation_parsers/translation_parser.ts
petebacondarwin: They look them same but have slightly different semantics. 
In the case you are suggesting the argument is optional.
In the case in the PR the argument is required but can be undefined.
I like making arguments required if possible because t helps identify places that are forgetting to make a decision about the argument. Also in this case it really is the case that you might pass undefined rather than not specifying the argument.

As it turns out the request from @clydin will probably mean that a lot of this will go away anyway.

IgorMinar: I see your point, and because we are still falling back onto the locale defined in the translation file, it would be helpful to be explicit. In that case I suggest that you document this behavior in the jsdoc above.

petebacondarwin:
After @clydin's feedback I completely redesigned this PR - so all the files have changed since the last reviews (cc @IgorMinar).

@clydin please note that the return type from `TranslationParser.parse()` has changed. I think you are using this directly in the CLI integration?? So this will be a breaking change for this integration.

PTAL!!

packages/localize/src/tools/src/translate/translation_files/translation_loader.ts
```diff
@@ -0,0 +1,57 @@
...
+              translationParser.parse(filePath, fileContents);
+          const locale = providedLocale || parsedLocale;
+          if (locale === undefined) {
+            throw new Error(
+                `No target locale provided nor found in the translation file "${filePath}"`);
+          }
+          if (parsedLocale !== undefined && providedLocale !== undefined &&
+              parsedLocale !== providedLocale) {
+            this.diagnostics.warn(
+                `The provided locale "${providedLocale}" and the target locale "${parsedLocale}" do not match for translation file "${filePath}".`);
```
IgorMinar: I don't think that this is easy to understand. How about "specified locale XX doesn't match locale YY found in the translation file"

packages/localize/src/tools/test/translate/translation_files/translation_loader_spec.ts
```diff
@@ -49,30 +52,71 @@ describe('TranslationLoader', () => {
...
+    });
+
+    it('should throw an error if there is no provided nor parsed target locale', () => {
+      const translations = {};
+      const diagnostics = new Diagnostics();
+      const parser = new MockTranslationParser(true, undefined, translations);
+      const loader = new TranslationLoader([parser], diagnostics);
+      expect(() => loader.loadBundles(['/src/locale/messages.en.xlf'], []))
+          .toThrowError(
+              'No target locale provided nor found in the translation file "/src/locale/messages.en.xlf"');
```
IgorMinar: Maybe add "please provide --target-locale argument"?

IgorMinar (commented):
A few more suggestions.

packages/localize/src/tools/test/translate/translation_files/translation_loader_spec.ts
petebacondarwin: The only problem with this is that it might be called via the Angular CLI which does not have this option...

petebacondarwin: How about: "consider explicitly providing a target-locale for this file"?

packages/localize/src/tools/src/translate/translation_files/translation_parsers/translation_parser.ts
```diff
@@ -5,7 +5,15 @@
...
  */
-import {TranslationBundle} from '../../translator';
+import {ɵMessageId, ɵParsedTranslation} from '@angular/localize/private';
+
+/**
+* An object that holds translations that have been parsed from a translation file.
+*/
+export interface ParsedTranslationBundle {
+  parsedLocale: string|undefined;
+  parsedTranslations: Record<ɵMessageId, ɵParsedTranslation>;
```
clydin: If these names are kept unchanged, no code changes would be needed on the CLI side.

petebacondarwin: This is where it is used in the CLI (for context): https://github.com/angular/angular-cli/blob/395c97912800af435021432e708cbdfab28783fe/packages/angular_devkit/build_angular/src/utils/load-translations.ts

Only the `translations` property appears to be important...

petebacondarwin: I'll switch them back...

petebacondarwin:
As I am a code-owner of this, and having approval from @IgorMinar, I think that this can be merged, right?

packages/localize/src/tools/test/translate/translation_files/translation_parsers/simple_json/simple_json_spec.ts
```diff
@@ -20,7 +20,7 @@ describe('SimpleJsonTranslationParser', () => {
   describe('parse()', () => {
     it('should extract the locale from the JSON contents', () => {
       const parser = new SimpleJsonTranslationParser();
-      const result = parser.parse('/some/file.json', '{"locale": "en", "translations": {}}');
+      const result = parser.parse('/some/file.json', '{"locale": "en", "translations": {}}', );
```
AndrewKushnir: Extra comma?

packages/localize/src/tools/test/translate/translation_files/translation_parsers/simple_json/simple_json_spec.ts
```diff
@@ -31,7 +31,7 @@ describe('SimpleJsonTranslationParser', () => {
         "translations": {
           "Hello, {$ph_1}!": "Bonjour, {$ph_1}!"
         }
-      }`);
+      }`, );
```
AndrewKushnir: Extra comma?

packages/localize/src/tools/src/translate/main.ts
```diff
@@ -47,58 +47,103 @@ if (require.main === module) {
...
-                  missingTranslation, sourceLocale});
+  translateFiles({sourceRootPath, sourceFilePaths, translationFilePaths, translationFileLocales,
+                  outputPathFn, diagnostics, missingTranslation, sourceLocale});
 
   diagnostics.messages.forEach(m => console.warn(`${m.type}: ${m.message}`));
   process.exit(diagnostics.hasErrors ? 1 : 0);
 }
 
 export interface TranslateFilesOptions {
+  /**
```
AndrewKushnir: Thanks for adding docs 👍

AndrewKushnir (approved):
LGTM, just left a couple comments. Thanks Pete.

clydin (approved):

PR #33389, opened 2019-10-25
build: add ng_benchmark macro to run perf benchmarks

IgorMinar:
this makes running and profiling tests much easier. Example usage:

```
yarn bazel run --define=compile=aot //packages/core/test/render3/perf:noop_change_detection
```

See README.md update for more info.

PS: I considered moving the ng_rollup bundle into the macro but I didn't want to make
  too many changes in this PR. If we find running benchmarks in this way useful, we
  should refactor the build file more, and move the ng_rollup_bundle targets into the
  macro.

### DIFF
--- a/tools/ng_benchmark.bzl
+++ b/tools/ng_benchmark.bzl
@@ -1,0 +1,32 @@
+# Copyright Google Inc. All Rights Reserved.
+#
+# Use of this source code is governed by an MIT-style license that can be
+# found in the LICENSE file at https://angular.io/license
+"""Bazel macro for running Angular benchmarks"""
+
+load("@build_bazel_rules_nodejs//:defs.bzl", "nodejs_binary")
+
+def ng_benchmark(name, bundle):
+    """
+
+    This macro creates two targets, one that is runnable and prints results and one that can be used for profiling via chrome://inspect.
+
+    Args:
+      name: name of the runnable rule to create
+      bundle: label of the ng_rollup_bundle to run
+    """
+
+    nodejs_binary(
+        name = name,
+        data = [bundle],
+        entry_point = bundle + ".min_debug.es2015.js",
+        tags = ["local"],
+    )
+
+    nodejs_binary(
+        name = name + "_profile",
+        data = [bundle],
+        entry_point = bundle + ".min_debug.es2015.js",
+        args = ["--node_options=--no-turbo-inlining --node_options=--inspect-brk"],
+        tags = ["local"],
+    )
--- a/packages/core/test/render3/perf/BUILD.bazel
+++ b/packages/core/test/render3/perf/BUILD.bazel
@@ -1,6 +1,6 @@
 package(default_visibility = ["//visibility:private"])
 
-load("//tools:defaults.bzl", "jasmine_node_test", "ng_rollup_bundle", "ts_library")
+load("//tools:defaults.bzl", "jasmine_node_test", "ng_benchmark", "ng_rollup_bundle", "ts_library")
 
 ts_library(
     name = "perf_lib",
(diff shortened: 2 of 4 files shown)


IgorMinar:
@pkozlowski-opensource @matsko @mhevery  - fyi

josephperrott (approved):

tools/ng_benchmark.bzl
```diff
@@ -0,0 +1,31 @@
...
+    """
+
+    This macro creates two targets, one that is runnable and prints results and one that can be used for profiling via chrome://inspect.
+
+    Args:
+      name: name of the runnable rule to create
+      bundle: label of the ng_rollup_bundle to run
+    """
+
+    nodejs_binary(
```
josephperrott: Instead of using `nodejs_binary` here we should use `nodejs_test`, this will allow us to execute using either `bazel run` or `bazel test`.

If we run through `bazel test` we can cache the results so that if no changes occur we just get back the same result since no inputs to the action have actually changed.

josephperrott: I believe we are talking about the same caching.

I am not sure I understand why its desirable to see the fluctuations when no changes to the sources have occurred.  From my understanding/estimation these fluctuations point to inaccuracy/inconsistency in the data that lessens its usefulness.  I definitely might be misunderstanding what we are looking for, my thought comes from viewing these benchmarkings as needing our best effort of full isolation.

We already have logic in place to attempt to run benchmark in the test many times, until we feel that we have a stable value we can rely on.  Once that stable value exists, it seems to me that we should rely on the value until the underlying source changes.  So we would only really want to find the next stable value once any of the sources change.

We can also tag the rule `no-remote` so that it never remotely executes or remotely caches, which would mean the only time we see a cached value it will come from a local execution so we can be confident that the values it has are from the exact same hardware.

pkozlowski-opensource: > I am not sure I understand why its desirable to see the fluctuations when no changes to the sources have occurred. From my understanding/estimation these fluctuations point to inaccuracy/inconsistency in the data that lessens its usefulness. I definitely might be misunderstanding what we are looking for, my thought comes from viewing these benchmarkings as needing our best effort of full isolation.

In practice it is very hard to have results that not fluctuate - there are so many variables and we don't understand all of them. And we continuously work on improving our understanding of fluctuations, but we are not there yet. To get to the "nirvana state" we need to acknowledge that those fluctuations exists, see how much results fluctuate and work towards eliminating fluctuations. But again, we are not there yet... 

> We already have logic in place to attempt to run benchmark in the test many times, until we feel that we have a stable value we can rely on. Once that stable value exists, it seems to me that we should rely on the value until the underlying source changes. So we would only really want to find the next stable value once any of the sources change.

Again, agree on the principle, but practice is different.

> We can also tag the rule no-remote so that it never remotely executes or remotely caches, which would mean the only time we see a cached value it will come from a local execution so we can be confident that the values it has are from the exact same hardware.

Oh, yeh, most definitively we should _not_ remotely-cache it!

pkozlowski-opensource: To put it simply:
- today we know that fluctuations exists
- to get rid of fluctuations we need to see and understand them.

Removing this information will make us content with "one number" that will _not_ reflect reality of things.

josephperrott: Using the `nodejs_test` rule allows for it to be targeted as via `bazel test`, but `bazel run` still also works and will always results in a new execution of the performance test script.  The additional context that you have laid out definitely explains why its important that have the ability to rerun with the same sources.

I am fine with whatever you and Igor decide in terms of how to you want to get to the state. I believe that should definitely add the `no-remote` tag regardless though.

pkozlowski-opensource: Yeah, for the workflows I'm using today introducing caching of results would limit the usefulness of the tool and would remove important information for me. I _hope_ that one day we will get to the point were results are super-stable but we are not there yet..

IgorMinar: I'll keep it as nodejs_binary for now so that we don't execute the tests on CI unintentionally just yet.

IgorMinar:
I did tag all the targets with `local` tag to prevent rbe from picking it up

IgorMinar:
merge-assistance: approval for core

PR #33392, opened 2019-10-25
test(ivy): support `className` in micro benchmarks

mhevery:
The styling algorithm requires that the `RNode` has a `className`
property in order to execute the fast-path. This changes adds the
emulation of this property.

### DIFF
--- a/packages/core/test/render3/perf/noop_renderer.ts
+++ b/packages/core/test/render3/perf/noop_renderer.ts
@@ -7,7 +7,7 @@
  */
 import {ProceduralRenderer3, RComment, RElement, RNode, RText, Renderer3, RendererFactory3, RendererStyleFlags3} from '../../../src/render3/interfaces/renderer';
 
-export class WebWorkerRenderNode implements RNode, RComment, RText {
+export class MicroBenchmarkRenderNode implements RNode, RComment, RText {
   textContent: string|null = null;
   parentNode: RNode|null = null;
   parentElement: RElement|null = null;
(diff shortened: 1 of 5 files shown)


packages/core/test/render3/perf/noop_renderer.ts
```diff
@@ -47,11 +59,39 @@ export class NoopRenderer implements ProceduralRenderer3 {
...
     if (typeof global !== 'undefined') {
-      (global as any).Node = WebWorkerRenderNode;
+      (global as any).Node = MicroBenchmarkRenderNode;
     }
-    return new NoopRenderer();
+    return new MicroBenchmarkRenderer();
   }
 }
+
+function isOurNode(node: any): node is MicroBenchmarkRenderNode {
```
pkozlowski-opensource: I'm not clear why do we need the `isOurNode` check - wouldn't we always run with the `MicroBenchmarkRender` that produces the same types of nodes.

pkozlowski-opensource (approved):
LGTM but marking it with cleanup since:
* I left one comment / question
* my fixup commit needs to be reviewed 
* CLA verification needs to be forced

packages/core/test/render3/perf/noop_renderer.ts
mhevery: is was there for type narrowing. Otherwise I need to do casting in TS to make it happy

pkozlowski-opensource: Oh, I see. Could you leave a comment on this function so a next person doesn't remove it (I as about to attempt to do just this :-) )

mhevery:
GLOBAL APPROVAL

PR #33414, opened 2019-10-25
[wip] Rollup es2015 & then downlevel to es5 for language-service bundle

gregmagolan:
***Branches off of nodejs rules 0.39.0 PR #33426 which should land first.***

Testing replacing ls_rollup_bundle with the new @bazel/rollup rollup_bunde which can only handle es2015 inputs at the moment (the esm5 re-rooting does not work with it). Instead of rolling up esm5, this PR changes the language-service bundle to be a rollup of es2015 + a babel downlevel step.

The new bundle differs from the original. I'll attach the two for comparison.


gregmagolan:
[es2015+babel.txt](https://github.com/angular/angular/files/3774180/es2015%2Bbabel.txt)

gregmagolan:
[esm5.txt](https://github.com/angular/angular/files/3774181/esm5.txt)

kyliau:
The language service bundle could technically be es2015 - there are no other consumers besides the extension. If we don't care about APF then we could just omit the downlevelling.

AndrewKushnir:
It looks like this PR was updated over a year ago. @gregmagolan should we keep it open or we can close for now?

PR #33415, opened 2019-10-26
fix(ivy): handle elements with local refs in i18n blocks

AndrewKushnir:
Prior to this commit, i18n logic which ensures that elements removed in a translation are also removed in DOM, didn't take into account the fact that elements may have local refs. As a result, remove operation failed, since there is no corresponding tNode found. This commit updates the logic to skip all local refs while going though the list of nodes to ensure that DOM matches elements present in translation.

## PR Type

- [x] Bugfix

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/render3/i18n.ts
+++ b/packages/core/src/render3/i18n.ts
@@ -695,7 +695,7 @@ function i18nEndFirstPass(lView: LView, tView: TView) {
       // Divide by 2 to get the number of local refs,
       // since they are stored as an array that also includes directive indexes,
       // i.e. ["localRef", directiveIndex, ...]
-      index += tNode.localNames.length / 2;
+      index += tNode.localNames.length >> 1;
     }
     index++;
   }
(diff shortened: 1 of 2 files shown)


packages/core/src/render3/i18n.ts
```diff
@@ -683,10 +683,21 @@ function i18nEndFirstPass(lView: LView, tView: TView) {
...
+      removeNode(index, lView, /* markAsDetached */ true);
     }
+    // Check if an element has any local refs and skip them
+    const tNode = getTNode(index, lView);
+    if (tNode && (tNode.type === TNodeType.Element || tNode.type === TNodeType.ElementContainer) &&
+        tNode.localNames !== null) {
+      // Divide by 2 to get the number of local refs,
+      // since they are stored as an array that also includes directive indexes,
+      // i.e. ["localRef", directiveIndex, ...]
+      index += tNode.localNames.length / 2;
```
mhevery:
```suggestion
      index += tNode.localNames.length << 1;
```
Division is expensive since the VM needs to do non-integer arithmetic and then convert to integer. Shift is faster.

mhevery (approved):

packages/core/src/render3/i18n.ts
AndrewKushnir: Thanks for the review @mhevery!

AndrewKushnir:
- [VE Presubmit](http://test/OCL:277176385:BASE:277176435:1572308271209:a7833932)
- [Ivy Presubmit](http://test/OCL:277176385:BASE:277176604:1572308331146:ebb22d04)

PR #33419, opened 2019-10-26
feat(ivy): implement unknown element detection in jit mode

crisbeto:
In ViewEngine we used to throw an error if we encountered an unknown element while rendering. We have this already for Ivy in AoT, but we didn't in JiT. These changes implement the error for JiT mode.

### DIFF
--- a/packages/core/src/render3/instructions/element.ts
+++ b/packages/core/src/render3/instructions/element.ts
@@ -10,7 +10,7 @@ import {assertDataInRange, assertDefined, assertEqual} from '../../util/assert';
 import {assertHasParent} from '../assert';
 import {attachPatchData} from '../context_discovery';
 import {registerPostOrderHooks} from '../hooks';
-import {TAttributes, TNodeFlags, TNodeType} from '../interfaces/node';
+import {TAttributes, TNode, TNodeFlags, TNodeType} from '../interfaces/node';
 import {RElement} from '../interfaces/renderer';
 import {StylingMapArray, TStylingContext} from '../interfaces/styling';
 import {isContentQueryHost, isDirectiveHost} from '../interfaces/type_checks';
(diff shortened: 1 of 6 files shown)


packages/core/src/render3/instructions/element.ts
```diff
@@ -241,3 +242,33 @@ function setDirectiveStylingInput(
...
+      errorMessage +=
+          `1. If '${tagName}' is an Angular component, then verify that it is part of this module.\n`;
+      if (tagName && tagName.indexOf('-') > -1) {
+        errorMessage +=
+            `2. If '${tagName}' is a Web Component then add 'CUSTOM_ELEMENTS_SCHEMA' to the '@NgModule.schemas' of this component to suppress this message.`;
+      } else {
+        errorMessage +=
+            `2. To allow any element add 'NO_ERRORS_SCHEMA' to the '@NgModule.schemas' of this component.`;
+      }
+      throw new Error(errorMessage);
```
crisbeto: Question for reviewer: should we be throwing here or logging a warning? For the property validation we decided to warn.

mhevery (approved):

AndrewKushnir:
- [VE Presubmit](http://test/OCL:277127900:BASE:277127929:1572293526916:3a569cd6)
- [Ivy Presubmit](http://test/OCL:277127900:BASE:277128042:1572293566356:ed098541)

AndrewKushnir:
FYI, Ivy presubmit indicated a problem where the "unknown element" error (added in this PR) is thrown in the code where it was working before. I will investigate the problem on g3 side and will come up with a TestBed repro. Thank you.

AndrewKushnir:
@crisbeto investigation in g3 revealed that failing tests are misconfigured (where some modules are imported via ngsummaries only and not present in imports list). The error message introduced in this PR and thrown in those tests, is actually a correct Ivy behavior. I will adjust the tests tomorrow to make sure they are passing with upcoming changes. Meanwhile, could you please rebase this PR on top of the latest master (so we run CI again) and may be change commit message to be `fix(ivy)` vs `feat(ivy)` (and keep `master&patch` as a target)? We can treat this change as a fix of incorrect behavior in Ivy where error was not thrown (vs being thrown in VE). Thank you.

AndrewKushnir:
New set of presubmits:
- [VE Presubmit](http://test/OCL:278032527:BASE:278032545:1572654931297:e6f7e818)
- [Ivy Presubmit](http://test/OCL:278032527:BASE:278032733:1572655012666:8b635c72)

AndrewKushnir:
FYI, VE and Ivy presubmits (including global for Ivy) are successful. Added "merged" label. Thank you.

packages/core/src/render3/instructions/element.ts
kara: We should be warning here. Unfortunately seeing this pretty late :(

crisbeto: I'll submit another PR to switch it to a warning.

PR #33425, opened 2019-10-26
docs(common): fix condition in transformed *ngIf

maiermic:
`<div class="hero-list" *ngIf="heroes else loading">` defines `heroes` as condition, but (`class` attribute value) `hero-list` is used as condition in the transformed code.


StephenFluin (approved):

AndrewKushnir (approved):

AndrewKushnir:
Merge assistance: should not require g3 presubmit, since it's a doc-only change.

PR #33431, opened 2019-10-27
perf(ivy): avoid generating selectors array for directives without a selector

crisbeto:
Now that we've replaced `ngBaseDef` with an abstract directive definition, there are a lot more cases where we generate a directive definition without a selector. These changes make it so that we don't generate the `selectors` array if it's going to be empty.

### DIFF
--- a/packages/compiler/src/render3/view/compiler.ts
+++ b/packages/compiler/src/render3/view/compiler.ts
@@ -38,20 +38,19 @@ const EMPTY_ARRAY: any[] = [];
 // If there is a match, the first matching group will contain the attribute name to bind.
 const ATTR_REGEX = /attr\.([^\]]+)/;
 
-function getStylingPrefix(name: string): string {
-  return name.substring(0, 5);  // style or class
-}
-
 function baseDirectiveFields(
     meta: R3DirectiveMetadata, constantPool: ConstantPool,
     bindingParser: BindingParser): DefinitionMap {
   const definitionMap = new DefinitionMap();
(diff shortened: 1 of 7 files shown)


packages/compiler/src/render3/view/compiler.ts
```diff
@@ -38,20 +38,19 @@ const EMPTY_ARRAY: any[] = [];
 // If there is a match, the first matching group will contain the attribute name to bind.
 const ATTR_REGEX = /attr\.([^\]]+)/;
 
-function getStylingPrefix(name: string): string {
-  return name.substring(0, 5);  // style or class
-}
-
```
petebacondarwin: Was this just dead-code?

petebacondarwin (approved):

packages/compiler/src/render3/view/compiler.ts
crisbeto: Yes, it's been there for a while so I decided to clean it up.

mhevery (approved):

AndrewKushnir:
- [VE Presubmit](http://test/OCL:277127551:BASE:277127625:1572293422903:26961c92)
- [Ivy Presubmit](http://test/OCL:277127551:BASE:277127766:1572293465853:8185d06e)

AndrewKushnir:
@crisbeto FYI, VE and Ivy presubmits are successful for this PR. Feel free to add "merge" label if it's ready to be merged. Thank you.

PR #33435, opened 2019-10-27
feat(bazel): update bazel-schematics to use Ivy and new rollup_bundle

gregmagolan:
Note: the @angular/bazel schematic now appends the package.json "script" field with 'ngcc --properties es2015 browser module main'. If there is an existing script field with ngcc then the schematic modifies it in place removing `--first-only` and `--create-ivy-entry-points`.

ViewEngine sources under node_modules need to be updated in-place for Bazel as it does not know how to use the `__ivy__` entry points that are created by the non-bazel `ngcc` command that is added to "scripts" :`ngcc --properties es2015 browser module main --first-only --create-ivy-entry-points`.

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

@angular/bazel schematics was depending on the legacy rollup_bundle rule which will be removed for the nodejs rules 1.0 release and it was building ViewEngine

## What is the new behavior?

@angular/bazel schematics updated to Ivy & to the new rollup_bundle rule

- [ ] Yes
- [ ] No


packages/bazel/src/schematics/ng-add/index.ts
```diff
@@ -333,19 +335,23 @@ function addPostinstallToGenerateNgSummaries() {
...
-    const ngcCommand = 'ngc -p ./angular-metadata.tsconfig.json';
+    // For bazel we need to compile the all files in place so we
+    // don't use `--properties`, `--first-only` or `--create-ivy-entry-points`
+    const ngccCommand = 'ngcc';
     if (scripts) {
       const postInstall = findPropertyInAstObject(scripts, 'postinstall');
       if (postInstall) {
-        const command = `${postInstall.value}; ${ngcCommand}`;
+        // Our `ngcc` command should go first so it is executed before before the CLI
+        // generated ngcc command as if it goes after it does not work as expected
```
alexeagle: that's strange that you could end up with two ngcc commands?

alexeagle (approved):

packages/bazel/src/schematics/ng-add/index.ts
gregmagolan: ng new with ivy adds ```  "scripts": {
    "postinstall": "ngcc --properties es2015 browser module main --first-only --create-ivy-entry-points"
  }
``` which doesn't work with bazel because of `--first-only` and `--create-ivy-entry-points`

chatted with @filipesilva about this on slack and I think the better solution is to mutate the `ngcc <options>` to just `ngcc` for bazel instead of adding a second

**Review by josephperrott — COMMENTED** [review]
This also moves to bazel `1.1.0`?  We should definitely be noting that in the commits/PR message.

**Comment by gregmagolan** [comment]
> This also moves to bazel `1.1.0`? We should definitely be noting that in the commits/PR message.

👍 will add that note and cleanup the `ngcc` postinstall

**Comment by gregmagolan** [comment]
Actually, given https://github.com/bazelbuild/rules_nodejs/issues/1307 I'll remove the Bazel update on this PR as it will break Windows users.

File: packages/bazel/src/schematics/ng-add/index.ts (hunk 1 above)
Comment (kyliau) [reply packages/bazel/src/schematics/ng-add/index.ts]: Even with just `ngcc`, there are still two `ngcc` commands aren't there?
I think Alex's point was, why not just one ngcc command?

Comment (gregmagolan) [reply packages/bazel/src/schematics/ng-add/index.ts]: Cleanup up. If there is an existing ngcc command then the schematic modified the one in-place removing `--first-only` and `--create-ivy-entry-points`:

```
        if (/\bngcc\b/.test(value)) {
          // `ngcc` is already in the postinstall script
          value = value
              .replace(/\s*--first-only\b/, '')
              .replace(/\s*--create-ivy-entry-points\b/, '');
          replacePropertyInAstObject(recorder, scripts, 'postinstall', value);
        } else {
          const command = `${postInstall.value}; ${ngccCommand}`;
          replacePropertyInAstObject(recorder, scripts, 'postinstall', command);
        }
```

kyliau (approved):
LGTM, thank you!

josephperrott (approved):
LGTM

alexeagle:
Caretaker: assistance because there is a "cla/google" status that isn't turning green

josephperrott:
Removing `merge-assistance` as `cla/google` status is corrected

gregmagolan:
@caretaker Noticed the comment on the commit was outdated so updated it as it's a `feat` commit. No other changes. CI running again.

PR #33444, opened 2019-10-28
i18n: support xmb formatted translation files in compile time inlining

petebacondarwin:


AndrewKushnir (approved):
LGTM, thanks for additional refactoring as well 👍 

Pete, I'm wondering if we can add an e2e test that would:
- generate XMB file based on an app (using `ng xi18n` CLI or APIs)
- update XMB file (simulating translation changes)
- run i18n inlining process (that would also involve XMB parsing)
- verify that the page contains what we expect
That might be helpful to make sure that extraction and inlining (consuming part) are in sync.
Thank you.

AndrewKushnir:
@petebacondarwin this PR seems to have merge conflicts now, since the base PR is now merged. Could you plz rebase and resolve conflicts? Thank you.

petebacondarwin:
@AndrewKushnir - there is a test here that does exactly what you suggest - except that it uses the XLIFF 1.2 format rather than XMB. See https://github.com/angular/angular/blob/master/integration/cli-hello-world-ivy-i18n/package.json#L23

Was your suggestion that we should duplicate this test for XMB?

AndrewKushnir:
> Was your suggestion that we should duplicate this test for XMB?

@petebacondarwin  yeah, I was wondering if we can extend existing test to cover other formats as well.

PR #33450, opened 2019-10-28
fix(compiler): i18n - ignore `alt-trans` tags in XLIFF 1.2

### LINKED ISSUE #33161: i18n xlf alt-trans tag is used instead of target
# 🐞 bug report

### Affected Package
@angular/i18n

### Description

After extracting the translations as xlf files and editing them in my translation tool angular cannot use the translations correctly anymore. The translation tool adds <alt-trans> tags inside the document which angular then reads instead of the actual translation.

In the following example angular displays `Content` (trans-unit > alt-trans > target) instead of `Weiter` (trans-unit > target)

## 🔬 Minimal Reproduction
```xml
<!-- messages.de.xml -->
<trans-unit datatype="html" approved="no" id="registration.submit">
   <source>
      Continue
   </source>
   <target state="translated" xml:lang="de">Weiter</target>
   <context-group purpose="location">
      <context context-type="sourcefile">src/app/auth/registration-form/registration-form.component.html</context>
      <context context-type="linenumber">69</context>
   </context-group>
   <?sid 1110954287-0?>
   <alt-trans origin="autoFuzzy" tool="Swordfish" match-quality="71" ts="63">
      <source xml:lang="en">Content</source>
      <target state="translated" xml:lang="de">Content</target>
   </alt-trans>
</trans-unit>
```

```html
<button type="submit" class="btn btn-submit" i18n="@@registration.submit">
   Continue
</button>
```

## 🌍  Your Environment

**Angular Version:**
<pre>
@angular-devkit/architect         0.803.4
@angular-devkit/build-angular     0.803.4
@angular-devkit/build-optimizer   0.803.4
@angular-devkit/build-webpack     0.803.4
@angular-devkit/core              8.3.4
@angular-devkit/schematics        8.3.6
@angular/cli                      8.3.6
@ngtools/webpack                  8.3.4
@schematics/angular               8.3.6
@schematics/update                0.803.6
<code>

</code></pre>

**Anything else relevant?**
[Swordfish III](https://www.maxprograms.com/products/swordfish.html) used for translations.

petebacondarwin:
The parser was accidentally reading the `target` tag
below the `alt-trans` target and overriding the correct
`target` tag.

(This already worked in `$localize` but a test has been
added to confirm.)

Fixes #33161

---

This needs to be merged to patch as well, but I don't think will go cleanly because of the `$localize` test.
So I will create a new PR for patch once this has been approved.

### DIFF
--- a/packages/compiler/src/i18n/serializers/xliff.ts
+++ b/packages/compiler/src/i18n/serializers/xliff.ts
@@ -25,6 +25,7 @@ const _MARKER_TAG = 'mrk';
 const _FILE_TAG = 'file';
 const _SOURCE_TAG = 'source';
 const _SEGMENT_SOURCE_TAG = 'seg-source';
+const _ALT_TRANS_TAG = 'alt-trans';
 const _TARGET_TAG = 'target';
 const _UNIT_TAG = 'trans-unit';
 const _CONTEXT_GROUP_TAG = 'context-group';
@@ -222,6 +223,7 @@ class XliffParser implements ml.Visitor {
       // ignore those tags
       case _SOURCE_TAG:
       case _SEGMENT_SOURCE_TAG:
+      case _ALT_TRANS_TAG:
         break;
 
       case _TARGET_TAG:
--- a/packages/compiler/test/i18n/serializers/xliff_spec.ts
+++ b/packages/compiler/test/i18n/serializers/xliff_spec.ts
@@ -296,6 +296,30 @@ lignes`,
       it('should return the target locale',
          () => { expect(serializer.load(LOAD_XLIFF, 'url').locale).toEqual('fr'); });
 
+      it('should ignore alt-trans targets', () => {
+        const XLIFF = `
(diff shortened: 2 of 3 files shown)


AndrewKushnir (approved):
LGTM 👍

ocombe:
I don't think the right approach is to ignore alt-trans, it should be to ignore any target tag that isn't a direct descendant of "trans-unit" (if it is in fact always a direct descendant? that's to confirm)

petebacondarwin:
@ocombe - I agree that the best way is to only consider `trans-unit` tags. This is how the new `$localize` XLIFF 1.2 translation loader does it,  and is why the test added here works already.

But for the old `Xliff` parser, there was already an approach of ignoring certain tags, e.g. `source` and `seg-source`. Given that these parsers will be deprecated at some point I don't see the benefit in a big rewrite of those to follow a different approach when this simple fix suffices.

It is the case that the `target` tags that we are interested in must always be direct child (not even an indirect descendant) of `trans-unit`.

PR #33453, opened 2019-10-28
feat(ivy): graceful evaluation of unknown or invalid expressions

JoostK:
During static evaluation of expressions within ngtsc, it may occur that
certain expressions or just parts thereof cannot be statically
interpreted for some reason. The static interpreter keeps track of the
failure reason and the code path that was evaluated by means of
`DynamicValue`, which will allow descriptive errors. In some situations
however, the static interpreter would throw an exception instead,
resulting in a crash of the compilation. Not only does this cause
non-descriptive errors, more importantly does it prevent the evaluated
result from being partial, i.e. parts of the result can be dynamic if
their value does not have to be statically available to the compiler.

This commit refactors the static interpreter to never throw errors for
certain expressions that it cannot evaluate.

Resolves FW-1582

### DIFF
--- a/packages/compiler-cli/src/ngtsc/partial_evaluator/src/dynamic.ts
+++ b/packages/compiler-cli/src/ngtsc/partial_evaluator/src/dynamic.ts
@@ -38,9 +38,10 @@ export const enum DynamicValueReason {
   EXTERNAL_REFERENCE,
 
   /**
-   * A type of `ts.Expression` that `StaticInterpreter` doesn't know how to evaluate.
+   * Syntax that `StaticInterpreter` doesn't know how to evaluate, for example a type of
+   * `ts.Expression` that is not supported.
    */
-  UNKNOWN_EXPRESSION_TYPE,
+  UNSUPPORTED_SYNTAX,
 
   /**
    * A declaration of a `ts.Identifier` could not be found.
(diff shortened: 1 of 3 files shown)


packages/compiler-cli/src/ngtsc/partial_evaluator/src/interpreter.ts
```diff
@@ -360,10 +357,10 @@ export class StaticInterpreter {
         return new ArrayConcatBuiltinFn(node, lhs);
       }
       if (typeof rhs !== 'number' || !Number.isInteger(rhs)) {
-        return DynamicValue.fromUnknown(node);
+        return DynamicValue.fromInvalidExpressionType(node, rhs);
       }
       if (rhs < 0 || rhs >= lhs.length) {
-        throw new Error(`Index out of bounds: ${rhs} vs ${lhs.length}`);
+        return undefined;
```
gkalpak: Maybe just remove this `if` block?

packages/compiler-cli/src/ngtsc/partial_evaluator/src/dynamic.ts
```diff
@@ -42,6 +42,11 @@ export const enum DynamicValueReason {
    */
   UNKNOWN_EXPRESSION_TYPE,
 
+  /**
+   * An operator was used that could not be statically evaluated.
+   */
+  UNKNOWN_OPERATOR,
```
alxhub: I'd rather the `DynamicValue` types not be quite so granular. How about we introduce `UNSUPPORTED_SYNTAX` instead?

packages/compiler-cli/src/ngtsc/partial_evaluator/src/dynamic.ts
```diff
@@ -54,6 +59,12 @@ export const enum DynamicValueReason {
    */
   INVALID_EXPRESSION_TYPE,
 
+  /**
+   * A spread element did not evaluate to an array or object, so could therefore not be statically
+   * evaluated.
+   */
+  INVALID_SPREAD_ELEMENT,
```
alxhub: `INVALID_EXPRESSION_TYPE` covers this case.

alxhub (changes requested):

alxhub (approved):

atscott:
[presubmit](http://test/OCL:277821537:BASE:277821614:1572563273935:a801aa46)
[ivy presubmit](http://test/OCL:277821537:BASE:277821898:1572563364353:a171f98d)

PR #33454, opened 2019-10-28
Unknown pipe & local ref diagnostics

alxhub:


packages/compiler-cli/src/ngtsc/typecheck/src/oob.ts
```diff
@@ -0,0 +1,45 @@
...
+
+import {AbsoluteSourceSpan, BindingPipe, TmplAstReference} from '@angular/compiler';
+import * as ts from 'typescript';
+
+import {ErrorCode, ngErrorCode} from '../../diagnostics';
+
+import {TcbSourceResolver, absoluteSourceSpanToSourceLocation, makeTemplateDiagnostic} from './diagnostics';
+
+/**
+ * Collects reports on template diagnostics which are
```
JoostK: Incomplete comment

packages/compiler-cli/src/ngtsc/typecheck/src/oob.ts
```diff
@@ -0,0 +1,45 @@
...
+export class OutOfBandDiagnosticRecorderImpl implements OutOfBandDiagnosticRecorder {
+  readonly diagnostics: ts.Diagnostic[] = [];
+
+  constructor(private resolver: TcbSourceResolver) {}
+
+  missingReferenceTarget(id: string, ref: TmplAstReference): void {
+    const mapping = this.resolver.getSourceMapping(id);
+    const value = ref.value.trim();
+
+    const errorMsg = `No directive found with exportAs '${value}'.`;
```
JoostK: For future consideration: have an intuitive way of visualising the module scope/available directive exports to better indicate what needs to happen to resolve this diagnostic. Because it's a non-local issue, how to fix it is not straightforward (and one of the areas I would like to see Angular become better is in error reporting!)

packages/compiler-cli/src/ngtsc/typecheck/src/type_check_block.ts
```diff
@@ -797,6 +798,7 @@ class Scope {
       for (const child of node.children) {
         this.appendNode(child);
       }
+      this.checkReferencesOfNode(node);
```
JoostK: I am surprised you took this approach, intertwining the reporting of missing refs with the generation of TCBs, instead of adding some diagnostic reporting facility to `R3TargetBinder` itself.

packages/compiler-cli/src/ngtsc/typecheck/src/type_check_block.ts
```diff
@@ -1025,7 +1036,11 @@ class TcbExpressionTranslator {
     } else if (binding instanceof TmplAstReference) {
       const target = this.tcb.boundTarget.getReferenceTarget(binding);
       if (target === null) {
-        throw new Error(`Unbound reference? ${binding.name}`);
+        // This reference is unbound.
+        this.tcb.oobRecorder.missingReferenceTarget(this.tcb.id, binding);
```
JoostK: I am under the impression that reporting this both here and in `checkReferencesOfNode` is not ideal and may introduce the same diagnostic twice. 

For instance:

```html
<div dir #ref="nonexistent">{{ ref.value }}</div>
```

This will create two diagnostics with identical message but different spans, if I'm not mistaken:
```html
<div dir #ref="nonexistent">{{ ref.value }}</div>
         ^~~~~~~~~~~~~~~~~~    ^~~
```
Both have the message `No directive found with exportAs 'nonexistent'.`. For the first diagnostics this message makes sense, for the second it does not.

packages/compiler/src/render3/view/t2_binder.ts
```diff
@@ -265,11 +265,11 @@ class DirectiveBinder<DirectiveT extends DirectiveMeta> implements Visitor {
             directives.find(
                 dir => dir.exportAs !== null && dir.exportAs.some(value => value === ref.value)) ||
             null;
-
-        // Check if a matching directive was found, and error if it wasn't.
+        // Check if a matching directive was found.
         if (dirTarget === null) {
-          // TODO(alxhub): Return an error value here that can be used for template validation.
-          throw new Error(`Assertion error: failed to find directive with exportAs: ${ref.value}`);
+          // No matching directive was found - this reference points to an unknown target.
```
JoostK: There's a comment on L263 that mentions "One should exist.", this is now outdated.

Also, this is the place where I expected some error reporting mechanism to be used, instead of in the TCB generation code.

packages/compiler-cli/src/ngtsc/typecheck/src/diagnostics.ts
```diff
@@ -55,6 +55,11 @@ export function toAbsoluteSpan(span: ParseSpan, sourceSpan: ParseSourceSpan): Ab
   return <AbsoluteSpan>{start: span.start + offset, end: span.end + offset};
 }
 
+export function absoluteSourceSpanToSourceLocation(
```
JoostK: Hmm, maybe we should just decompose `TcbSourceResolver.sourceLocationToSpan` to receive two arguments instead of the `SourceLocation` bundle.

packages/compiler-cli/src/ngtsc/typecheck/src/oob.ts
```diff
@@ -42,4 +46,18 @@ export class OutOfBandDiagnosticRecorderImpl implements OutOfBandDiagnosticRecor
...
   }
+
+  missingPipe(id: string, ast: BindingPipe, absSpan: AbsoluteSourceSpan): void {
+    const mapping = this.resolver.getSourceMapping(id);
+    const errorMsg = `No pipe found with name '${ast.name}'.`;
+
+    const location = absoluteSourceSpanToSourceLocation(id, absSpan);
+    const sourceSpan = this.resolver.sourceLocationToSpan(location);
+    if (sourceSpan === null) {
+      throw new Error(`Could not map location for error`);
```
JoostK: Pretty cryptic error message ;)

packages/compiler-cli/src/ngtsc/typecheck/src/oob.ts
```diff
@@ -20,12 +20,16 @@ export interface OutOfBandDiagnosticRecorder {
   readonly diagnostics: ts.Diagnostic[];
 
   missingReferenceTarget(id: string, ref: TmplAstReference): void;
+
+  missingPipe(id: string, ast: BindingPipe, sourceSpan: AbsoluteSourceSpan): void;
```
JoostK: I now realize that `id` corresponds with the TCB id, not some pipe id which the function name suggests. These methods could use docs.

packages/compiler-cli/src/ngtsc/typecheck/src/type_check_block.ts
```diff
@@ -975,9 +977,14 @@ class TcbExpressionTranslator {
       return ts.createIdentifier('ctx');
     } else if (ast instanceof BindingPipe) {
       const expr = this.translate(ast.exp);
-      let pipe: ts.Expression;
+      let pipe: ts.Expression|null;
+      const nameAbsoluteSpan = toAbsoluteSpan(ast.nameSpan, this.sourceSpan);
```
JoostK: Move `nameAbsoluteSpan` into the `if (pipe === null)` below

packages/compiler-cli/src/ngtsc/typecheck/src/type_check_block.ts
```diff
@@ -34,9 +35,10 @@ import {checkIfClassIsExported, checkIfGenericTypesAreUnbound, tsCallMethod, tsC
  */
 export function generateTypeCheckBlock(
```
JoostK: I think we should make clear that this function has a side-effect when it comes to producing diagnostics. Having the _out of bands_ parameter does indicate this somewhat, but it could be made more explicit. Come to think of it, it would be a good idea for future reference to document `OutOfBandDiagnosticRecorder` on why it exists and how it's used.

JoostK (changes requested):

packages/compiler-cli/src/ngtsc/typecheck/src/oob.ts
alxhub: I completely agree. One of the goals of having the centralized `ErrorCode` for Angular errors was to keep track of the different classes of errors and make it easier for us to set goals of improving them.

packages/compiler-cli/src/ngtsc/typecheck/src/type_check_block.ts
alxhub: I considered this, actually. Part of the problem is that the way we map expressions back into the original template source depends a lot on the way we traverse the template. This doesn't matter for `Reference`s which are a top level concept and have proper source spans, but for `BindingPipe`s which are an `AST` node, the source mapping depends on the context and the `R3TargetBinder` doesn't currently deal with those concerns.

Ultimately we might refactor in this direction. Especially because the HTML-level template parsing errors are not currently surfaced via our "nice" diagnostics mechanism.

packages/compiler-cli/src/ngtsc/typecheck/src/type_check_block.ts
alxhub: Very good point! Fixed.

packages/compiler/src/render3/view/t2_binder.ts
alxhub: Good catch, fixed.

packages/compiler-cli/src/ngtsc/typecheck/src/diagnostics.ts
alxhub: I'm all for refactoring this to clean it up :)

packages/compiler-cli/src/ngtsc/typecheck/src/type_check_block.ts
alxhub: Done. I updated and expanded the comment to talk about the verifications which happen during generation.

As mentioned above, I think the "right way" to do this is to refactor the template traversal to be independent of everything, and hang various operations off of this central traversal, including both the verification diagnostics and TCB generation.

packages/compiler-cli/src/ngtsc/typecheck/src/type_check_block.ts
alxhub: Done.

packages/compiler-cli/src/ngtsc/typecheck/src/oob.ts
alxhub: Renamed to `templateId`, and added docs.

packages/compiler-cli/src/ngtsc/typecheck/src/oob.ts
alxhub: Updated :)

packages/compiler-cli/src/ngtsc/typecheck/src/oob.ts
alxhub: Completed comment.

alxhub:
[Presubmit](http://test/OCL:277592991:BASE:277593080:1572473092181:613f68f7)
Ivy Presubmit not needed: does not affect generated code, no new errors

packages/compiler-cli/src/ngtsc/typecheck/src/oob.ts
```diff
@@ -34,6 +36,19 @@ export interface OutOfBandDiagnosticRecorder {
    * @param ref the `TmplAstReference` which could not be matched to a directive.
    */
   missingReferenceTarget(templateId: string, ref: TmplAstReference): void;
+
+  /**
+   * Reports usage of a `| pipe` expression in the template for which the named pipe could not be
+   * found.
+   *
+   * @param templateId the template type-cehcking ID of the template which contains the unknown
```
JoostK:
```suggestion
   * @param templateId the template type-checking ID of the template which contains the unknown
```

packages/compiler-cli/src/ngtsc/typecheck/src/oob.ts
```diff
@@ -34,6 +36,19 @@ export interface OutOfBandDiagnosticRecorder {
...
+   * pipe.
+   * @param ast the `BindingPipe` invocation of the pipe which could not be found.
+   * @param sourceSpan the `AbsoluteSourceSpan` of the pipe invocation (ideally, this should be the
+   * source span of the pipe's name). This depends on the source span of the `BindingPipe` itself
+   * plus span of the larger expression context.
+   */
+  missingPipe(id: string, ast: BindingPipe, sourceSpan: AbsoluteSourceSpan): void;
```
JoostK: You seem to have missed the actual param name in the interface here:

```suggestion
  missingPipe(templateId: string, ast: BindingPipe, sourceSpan: AbsoluteSourceSpan): void;
```

packages/compiler-cli/src/ngtsc/typecheck/src/type_check_block.ts
```diff
@@ -988,9 +988,17 @@ class TcbExpressionTranslator {
...
-      let pipe: ts.Expression;
+      let pipe: ts.Expression|null;
       if (this.tcb.env.config.checkTypeOfPipes) {
         pipe = this.tcb.getPipeByName(ast.name);
+        if (pipe === null) {
+          // No pipe by that name exists in scope. Instead,
+          const nameAbsoluteSpan = toAbsoluteSpan(ast.nameSpan, this.sourceSpan);
+          this.tcb.oobRecorder.missingPipe(this.tcb.id, ast, nameAbsoluteSpan);
+
+          //
```
JoostK: Remove empty comment (or make it count :))

JoostK (approved):
Just some more nits, otherwise LGTM

danielehrhardt:
When do you plan to release it?

PR #33457, opened 2019-10-28
fix(ivy): `ViewRef.rootNodes` returns rootNodes from child views as well

mhevery:
The idea behind `rootNodes` is to return all of the nodes which need
to be added to the DOM in order for the `ViewRef` to be rendered.
The new implementation correctly takes into account the case where
one of the rootNodes is a container containing other `ViewRef`s.

### DIFF
--- a/packages/core/src/render3/view_ref.ts
+++ b/packages/core/src/render3/view_ref.ts
@@ -11,10 +11,12 @@ import {ChangeDetectorRef as viewEngine_ChangeDetectorRef} from '../change_detec
 import {ViewContainerRef as viewEngine_ViewContainerRef} from '../linker/view_container_ref';
 import {EmbeddedViewRef as viewEngine_EmbeddedViewRef, InternalViewRef as viewEngine_InternalViewRef} from '../linker/view_ref';
 import {assertDefined} from '../util/assert';
+
 import {checkNoChangesInRootView, checkNoChangesInternal, detectChangesInRootView, detectChangesInternal, markViewDirty, storeCleanupFn} from './instructions/shared';
 import {CONTAINER_HEADER_OFFSET} from './interfaces/container';
 import {TElementNode, TNode, TNodeType, TViewNode} from './interfaces/node';
 import {CONTEXT, FLAGS, HOST, LView, LViewFlags, TView, T_HOST} from './interfaces/view';
+import {assertNodeOfPossibleTypes} from './node_assert';
 import {destroyLView, renderDetachView} from './node_manipulation';
 import {findComponentView, getLViewParent} from './util/view_traversal_utils';
 import {getNativeByTNode, getNativeByTNodeOrNull} from './util/view_utils';
@@ -302,8 +304,12 @@ export class RootViewRef<T> extends ViewRef<T> {
 
 function collectNativeNodes(lView: LView, tNode: TNode | null, result: any[]): any[] {
   while (tNode) {
+    ngDevMode && assertNodeOfPossibleTypes(
+                     tNode, TNodeType.Element, TNodeType.Container, TNodeType.Projection,
+                     TNodeType.ElementContainer);
     const nativeNode = getNativeByTNodeOrNull(tNode, lView);
     nativeNode && result.push(nativeNode);
+
     if (tNode.type === TNodeType.ElementContainer) {
       collectNativeNodes(lView, tNode.child, result);
     } else if (tNode.type === TNodeType.Container) {
--- a/packages/core/src/render3/view_ref.ts
+++ b/packages/core/src/render3/view_ref.ts
@@ -10,11 +10,13 @@ import {ApplicationRef} from '../application_ref';
 import {ChangeDetectorRef as viewEngine_ChangeDetectorRef} from '../change_detection/change_detector_ref';
 import {ViewContainerRef as viewEngine_ViewContainerRef} from '../linker/view_container_ref';
 import {EmbeddedViewRef as viewEngine_EmbeddedViewRef, InternalViewRef as viewEngine_InternalViewRef} from '../linker/view_ref';
+import {assertDefined} from '../util/assert';
 
 import {checkNoChangesInRootView, checkNoChangesInternal, detectChangesInRootView, detectChangesInternal, markViewDirty, storeCleanupFn} from './instructions/shared';
+import {CONTAINER_HEADER_OFFSET} from './interfaces/container';
 import {TElementNode, TNode, TNodeType, TViewNode} from './interfaces/node';
-import {CONTEXT, FLAGS, HOST, LView, LViewFlags, T_HOST} from './interfaces/view';
-import {destroyLView, renderDetachView} from './node_manipulation';
+import {CONTEXT, FLAGS, HOST, LView, LViewFlags, TView, T_HOST} from './interfaces/view';
+import {destroyLView, getLContainer, renderDetachView} from './node_manipulation';
 import {findComponentView, getLViewParent} from './util/view_traversal_utils';
 import {getNativeByTNode, getNativeByTNodeOrNull} from './util/view_utils';
 
@@ -38,7 +40,7 @@ export class ViewRef<T> implements viewEngine_EmbeddedViewRef<T>, viewEngine_Int
   get rootNodes(): any[] {
     if (this._lView[HOST] == null) {
       const tView = this._lView[T_HOST] as TViewNode;
-      return collectNativeNodes(this._lView, tView, []);
+      return collectNativeNodes(this._lView, tView.child, []);
     }
     return [];
   }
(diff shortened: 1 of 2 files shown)


packages/core/src/render3/view_ref.ts
```diff
@@ -299,27 +301,35 @@ export class RootViewRef<T> extends ViewRef<T> {
...
-      collectNativeNodes(lView, tNodeChild, result);
-    } else if (tNodeChild.type === TNodeType.Projection) {
+    if (tNode.type === TNodeType.ElementContainer) {
+      collectNativeNodes(lView, tNode.child, result);
+    } else if (tNode.type === TNodeType.Container) {
+      const lContainer = lView[tNode.index];
+      const containerTView = tNode.tViews !as TView;
+      ngDevMode && assertDefined(containerTView, 'TView expected');
+      const containerFirstChildTNode = containerTView.firstChild !;
+      if (containerFirstChildTNode) {
```
pkozlowski-opensource: Please use a stricter check here (`===` for `null` or whatever the falsy value might be).

packages/core/src/render3/view_ref.ts
```diff
@@ -299,27 +301,35 @@ export class RootViewRef<T> extends ViewRef<T> {
...
-  while (tNodeChild) {
-    const nativeNode = getNativeByTNodeOrNull(tNodeChild, lView);
+function collectNativeNodes(lView: LView, tNode: TNode | null, result: any[]): any[] {
+  while (tNode) {
+    const nativeNode = getNativeByTNodeOrNull(tNode, lView);
     nativeNode && result.push(nativeNode);
-    if (tNodeChild.type === TNodeType.ElementContainer) {
-      collectNativeNodes(lView, tNodeChild, result);
-    } else if (tNodeChild.type === TNodeType.Projection) {
+    if (tNode.type === TNodeType.ElementContainer) {
```
pkozlowski-opensource: I don't see a test for the `TNodeType.IcuContainer` in the series of node type checks. If I recall properly ICU containers "behave" similarly to `<ng-container>`s. I'm not sure if we can bump into ICU containers here, but:
* if no, we should add an assert;
* if yes, we should add a test + handle this case.

packages/core/src/render3/view_ref.ts
```diff
@@ -10,11 +10,13 @@ import {ApplicationRef} from '../application_ref';
...
 import {EmbeddedViewRef as viewEngine_EmbeddedViewRef, InternalViewRef as viewEngine_InternalViewRef} from '../linker/view_ref';
+import {assertDefined} from '../util/assert';
 
 import {checkNoChangesInRootView, checkNoChangesInternal, detectChangesInRootView, detectChangesInternal, markViewDirty, storeCleanupFn} from './instructions/shared';
+import {CONTAINER_HEADER_OFFSET} from './interfaces/container';
 import {TElementNode, TNode, TNodeType, TViewNode} from './interfaces/node';
-import {CONTEXT, FLAGS, HOST, LView, LViewFlags, T_HOST} from './interfaces/view';
-import {destroyLView, renderDetachView} from './node_manipulation';
+import {CONTEXT, FLAGS, HOST, LView, LViewFlags, TView, T_HOST} from './interfaces/view';
+import {destroyLView, getLContainer, renderDetachView} from './node_manipulation';
```
pkozlowski-opensource: I don't see the new `getLContainer ` import used anywhere in the changed code. Remove?

packages/core/src/render3/view_ref.ts
pkozlowski-opensource: Also, I would add an assert to verify that we don't bump into `tNode` being `TNodeType.View` (just for completeness, to make sure that we are not missing handling of any TNode type).

packages/core/test/acceptance/view_container_ref_spec.ts
```diff
@@ -872,6 +872,60 @@ describe('ViewContainerRef', () => {
...
+              this.viewContainerRef.createEmbeddedView(this.template, new MyDirEmbedded());
+          this.viewRef.detectChanges();
+        }
+      }
+
+      @Component({
+        selector: 'edit-form',
+        template: `
+          <ng-template [dir]="true">
+            <div *ngIf="true">Text</div>
```
pkozlowski-opensource: Could we add few more test cases:
- nested `<ng-container>`s at the root of an embedded view;
- ICU inside `<ng-container>` - I'm not sure if we can have them at the root of a view;
- a directive that asks for a `ViewContainerRef` at the root of a view (you've got a case where a VCRef is asked on a container);

pkozlowski-opensource (changes requested):
Did the first pass. The direction LGTM but I'm not sure if we are handling all the cases of nodes that can be at the root of a view (left comments enumerating those cases). As the very minimum we need more tests.

packages/core/src/render3/view_ref.ts
mhevery: fixed

packages/core/src/render3/view_ref.ts
mhevery: removed

packages/core/src/render3/view_ref.ts
mhevery: - I have added asserts
- Added tests and was not able to get `Icu` node to show up here.

packages/core/test/acceptance/view_container_ref_spec.ts
mhevery: added test cases

mhevery (commented):
Updated, PTAL

packages/core/src/render3/view_ref.ts
```diff
@@ -299,27 +302,39 @@ export class RootViewRef<T> extends ViewRef<T> {
...
-    if (tNodeChild.type === TNodeType.ElementContainer) {
-      collectNativeNodes(lView, tNodeChild, result);
-    } else if (tNodeChild.type === TNodeType.Projection) {
+
+    if (tNode.type === TNodeType.ElementContainer) {
+      collectNativeNodes(lView, tNode.child, result);
+    } else if (tNode.type === TNodeType.Container) {
+      const lContainer = lView[tNode.index];
+      const containerTView = tNode.tViews !as TView;
+      ngDevMode && assertDefined(containerTView, 'TView expected');
```
pkozlowski-opensource: I don't think that this assert is correct for 2 reasons:
* it refers to a `TView` _defined_ by this `<ng-template>` and not _inserted_ into a `ViewContainerRef` asked on this `<ng-template>`;
* many different types of TViews can be inserted into one container so we can't assume that all _inserted_ views share the same `TView`
* it can be `null` when the definition is empty `<ng-template></ng-template>`, AFAIK

There is a confusion between an embedded view _defintion_ and the the _inserted instance_ here.

packages/core/test/acceptance/view_container_ref_spec.ts
```diff
@@ -872,6 +872,65 @@ describe('ViewContainerRef', () => {
...
+            <ng-container>container-content</ng-container>
+            <div i18n>assert TNodeType.IcuContainer is never a root node</div>
```
pkozlowski-opensource: This is not enough, you need to have an actual ICU expression inside an element with the `i18n` attribute. Also, having `div` at the root is not enough as a `<div>` is a root node so no need to descend any further.  An example way of testing it would be `<ng-container i18n>Updated {minutes, plural, =0 {just now} =1 {one minute ago} other {several minutes ago}}</ng-container>` and here we need to descent into the ICU container.

tl;dr; I still believe that we need to handle the case of a ICU container.

packages/core/src/render3/view_ref.ts
```diff
@@ -299,27 +302,39 @@ export class RootViewRef<T> extends ViewRef<T> {
...
-function collectNativeNodes(lView: LView, parentTNode: TNode, result: any[]): any[] {
-  let tNodeChild = parentTNode.child;
-
-  while (tNodeChild) {
-    const nativeNode = getNativeByTNodeOrNull(tNodeChild, lView);
+function collectNativeNodes(lView: LView, tNode: TNode | null, result: any[]): any[] {
+  while (tNode) {
+    ngDevMode && assertNodeOfPossibleTypes(
+                     tNode, TNodeType.Element, TNodeType.Container, TNodeType.Projection,
+                     TNodeType.ElementContainer);
```
pkozlowski-opensource: I believe that we need to handle a ICU container, see: https://github.com/angular/angular/pull/33457/files#r340598750

packages/core/test/acceptance/view_container_ref_spec.ts
pkozlowski-opensource: I haven't seen a test for the case where a `ViewContainerRef ` is injected on a regular element.

pkozlowski-opensource (changes requested):
@mhevery I still believe that there are tests / use-cases missing. How about we review / merge https://github.com/angular/angular/pull/33493 instead that has all the test I could think of (+ fixes for those)?

mhevery:
Closing in as @pkozlowski-opensource took over https://github.com/angular/angular/pull/33493

PR #32155, opened 2019-08-15
perf: don't create holey arrays

mhevery:
Don't use `Array` constructor with the size value (ex. `new Array(5)`) - this will create a `HOLEY_ELEMENTS` array (even if this array is filled in later on!);

https://v8.dev/blog/elements-kinds
https://stackoverflow.com/questions/32054170/how-to-resize-an-array

PR #33470, opened 2019-10-29
perf(ivy): class bindings benchmark

pkozlowski-opensource:
This PR adds a new benchmark focused on `class` manipulation + benchpress scripts.

### DIFF
--- a/modules/benchmarks/src/class_bindings/app.component.ts
+++ b/modules/benchmarks/src/class_bindings/app.component.ts
@@ -1,0 +1,38 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import {Component} from '@angular/core';
+
+@Component({
+  selector: 'app-component',
+  template: `
+    <button id="create" (click)="create()">Create</button>
+    <button id="update" (click)="update()">Update</button>
+    <button id="destroy" (click)="destroy()">Destroy</button>
+    <class-bindings *ngIf="show" [msg]="msg" [list]="list"><class-bindings>
+  `
+})
+export class AppComponent {
+  show = false;
+  msg = 'hello';
+  list: {i: number, text: string}[] = [];
+
+  constructor() {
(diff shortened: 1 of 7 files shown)


modules/benchmarks/src/class_bindings/app.component.ts
```diff
@@ -0,0 +1,32 @@
...
+    <button id="create" (click)="create()">Create</button>
+    <button id="update" (click)="update()">Update</button>
+    <button id="destroy" (click)="destroy()">Destroy</button>
+    <class-bindings *ngIf="show" [msg]="msg" [list]="list"><class-bindings>
+  `
+})
+export class AppComponent {
+  show = false;
+  msg = 'hello';
+  list = new Array(1000).fill(0).map((_, i) => ({i, text: 'foobar' + i}));
```
mhevery: I believe this creates holley array. See https://github.com/angular/angular/pull/32155

mhevery (approved):

modules/benchmarks/src/class_bindings/app.component.ts
pkozlowski-opensource: fixed

modules/benchmarks/src/class_bindings/class_bindings.component.ts
```diff
@@ -0,0 +1,36 @@
...
+    <span class="baz">one</span>
+    <span class="qux">two</span>
+    <div>
+      <span class="qux">three</span>
+      <span class="qux">four</span>
+      <span class="baz">five</span>
+      <div>
+        <span class="qux">six</span>
+        <span class="baz">seven</span>
+        <span [class]="msg">eight</span>
```
IgorMinar: It's a bit surprising that a variable called "msg" is bound to class attr. If we change the msg value to contain a space the benchmark results could suddenly be very different.

IgorMinar (approved):
One nit, but I'm not sure if it's worth changing. Lgtm

modules/benchmarks/src/class_bindings/class_bindings.component.ts
pkozlowski-opensource: yeh, I've kept the names from the original benchmark, not sure it worth changing...

PR #33471, opened 2019-10-29
build: remove --nowatchfs flag as --watchfs is a noop on windows now

josephperrott:
Removing `--nowatchfs` to allow for linux/os x to get back, albeit minimal, performance gains.

`--watchfs` is now a noop on bazel for windows as seen [here](https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/skyframe/WatchServiceDiffAwareness.java#L90)


gkalpak (approved):
LGTM

PR #33458, opened 2019-10-29
build: update to nodejs rules 0.39.1

gregmagolan:
Pre-req for updating repo, integration & @bazel/schematics to Bazel 1.0.0 as this release brings in a fix for ts_library on Windows with Bazel 1.0.0+. See https://github.com/bazelbuild/rules_nodejs/issues/1307.

## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [x] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] angular.io application / infrastructure changes
- [ ] Other... Please describe:

## What is the current behavior?

Issue Number: N/A

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

PR #33476, opened 2019-10-29
build: update to @bazel/bazel 1.0.0

gregmagolan:
+ feat(bazel): update @bazel/schematics to Bazel 1.0.0

Also removes `build:remote --spawn_strategy=remote` from .bazelrc. It seems that with Bazel 1.0.0 setting `--incompatible_list_based_execution_strategy_selection=false` no longer works around the issue with npm_package that it did when it was added. The error that was originally observed has returned after updating to Bazel 1.0.0:

```
ERROR: /home/circleci/ng/packages/angular_devkit/build_optimizer/BUILD:66:1: Assembling npm package packages/angular_devkit/build_optimizer/npm_package failed: No usable spawn strategy found for spawn with mnemonic Action. Your --spawn_strategy, --genrule_strategy or --strategy flags are probably too strict. Visit https://github.com/bazelbuild/bazel/issues/7480 for migration advice
```

This commit removes both `—incompatible_list_based_execution_strategy_selection=false` as well as `build:remote --spawn_strategy=remote` which means that Bazel will do the default behavior of picking the first available strategy from the default list, which is `remote,worker,sandboxed,local`. See https://github.com/bazelbuild/bazel/issues/7480 for more details.

Not updating to Bazel 1.1.0 yet due to a docker permissions CI issue that was observed on the angular repo that is unresolved. See https://github.com/angular/angular/pull/33367#issuecomment-547643246.


gregmagolan:
Redo of https://github.com/angular/angular/pull/33367 which was reverted to to Windows CI failure that was fixed with the update to rules_nodejs 0.39.1 which just landed (#33458)

alexeagle (approved):

josephperrott (approved):
LGTM, did you verify that this works in our windows CI environment?

You can do so, by doing a commit that changes the circleci config to run windows, and then after we see it succeed, we can push the correct one again.

gregmagolan:
All good @josephperrott. Thanks for checking. I reproduced the exact issue on my local Windows machine got a fix in in rule_nodejs 0.39.1. Safe to merge.

Issue #4946, opened 2015-10-27
ng-content projection for root component


RomanGotsiy:
`ng-content` projection doesn't work for root component. E.g. for the following code:

```
<app>
   <p> Some content that needs to be projected </p>
</app>

@Component({
  selector: 'app',
})
@View({
  template: `<p>Hello world</p>
              <ng-content></ng-content>`
})
class App {}

bootstrap(App);
```

only `<p> Hello world <p>` persists in the DOM after `App` is bootstrapped.

I've reviewed the [design doc](https://docs.google.com/document/d/1BXUFVpImyWN7ynyPKehAesctP0UJTaDYbDTNVkarSeQ/edit) and haven't found such restriction.

[Plunker](http://plnkr.co/edit/10aPDcz0EUML1rjK7S3V?p=preview)

vicb:
see https://github.com/angular/angular/issues/1858

mhevery:
This should be an easy fix. We just have to pass in the content into the `componentFactory.create` [here](https://github.com/angular/angular/blob/master/modules/@angular/core/src/application_ref.ts#L456).

/cc @tbosch Do you think this is reasonable?

vicb:
note: we do not compile the index (only cmp templates)

mhevery:
@vicb true, I don't think this feature is about compiling, it is only about projecting the nodes, which should be easy.

DzmitryShylovich:
@mhevery 
> We just have to pass in the content into the componentFactory.create here.

The question is how to access that content? Something like this would work
```
const rootNode = document.querySelector(componentFactory.selector);
    if (!rootNode) {
      throw new Error(`The selector "${componentFactory.selector}" did not match any elements`);
    }
    const childNodes = Array.prototype.slice.call(rootNode.childNodes);
    const compRef = componentFactory.create(this._injector, [childNodes], rootNode);
```
but we don't have access to the DOM at this point (Renderer is not available yet).

tbosch:
@mhevery What about `<ng-content select="...">`? We don't have the css selector any more during runtime, not even the attribute `select`...

So we would need to somehow restrict this feature to only work for `<ng-content>` elements that have no `select` attribute, and throw for others. This requires somehow marking these `ng-content` elements during compilation.

tbosch:
Also a component can reproject nodes, so we really need to have the list of nodes.

What could work is the following:

- store the NgContent selectors in the view definition
- add a new method `queryDirectChildren` to the renderer
[...]

PR #33487, opened 2019-10-30
fix(ivy): ComponentFactory.create should clear host element content

AndrewKushnir:
Prior to this change, ComponentFactory.create function invocation in Ivy retained the content of the host element (in case host element reference or CSS selector is provided as an argument). This behavior is different in View Engine, where the content of the host element was cleared, except for the case when ShadowDom encapsulation is used (to make sure native slot projection works). This commit aligns Ivy and View Engine and makes sure the host element is cleared before component content insertion.

This PR resolves FW-1649.

## PR Type

- [x] Bugfix

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/render3/instructions/shared.ts
+++ b/packages/core/src/render3/instructions/shared.ts
@@ -717,9 +717,9 @@ export function locateHostElement(
   ngDevMode && assertHostNodeExists(rElement, elementOrSelector);
 
   // Always clear host element's content when Renderer3 is in use. For procedural renderer case we
-  // make it conditionally and depend on whether ShadowDom encapsulation is used (in which case the
-  // content should be preserved to allow native slot projection). ShadowDom encapsulation requires
-  // procedural renderer, and procedural renderer case is handled above.
+  // make it depend on whether ShadowDom encapsulation is used (in which case the content should be
+  // preserved to allow native slot projection). ShadowDom encapsulation requires procedural
+  // renderer, and procedural renderer case is handled above.
   rElement.textContent = '';
 
   return rElement;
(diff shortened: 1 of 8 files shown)


AndrewKushnir:
Initial presubmit runs to verify the fix:
- [VE Presubmit](http://test/OCL:277548909:BASE:277549011:1572460362639:dc20e01d)
- [Ivy Presubmit](http://test/OCL:277548909:BASE:277549286:1572460428302:38540648)

AndrewKushnir:
- [VE Presubmit](http://test/OCL:278034353:BASE:278034370:1572655880407:f406dd34)
- [Ivy Presubmit](http://test/OCL:278034353:BASE:278034434:1572655917648:eb64b35a)

AndrewKushnir:
- [VE Presubmit](http://test/OCL:279517026:BASE:279517048:1573329137503:2e181072)
- [Ivy Presubmit](https://test.corp.google.com/ui#id=OCL:279517026:BASE:279517117:1573329177328:4ad45f95)

atscott (commented):
LGTM. As discussed, please add a test case covering host elements with more than just a text node.

atscott (approved):

mhevery (approved):

pkozlowski-opensource:
Initially I thought that this change limits our ability to fix #4946 but after digging into it more, I don't think it blocks anything. WE can go ahead with this PR from the existing GH issues point of view.

packages/core/src/render3/instructions/shared.ts
```diff
@@ -678,32 +679,50 @@ function createViewBlueprint(bindingStartIndex: number, initialViewLength: numbe
...
   }
-  return rNode;
+
+  let rElement = typeof elementOrSelector === 'string' ?
+      renderer.querySelector(elementOrSelector) ! :
+      elementOrSelector;
+  ngDevMode && assertHostNodeExists(rElement, elementOrSelector);
+
+  // Always clear host element's content when Renderer3 is in use. For procedural renderer case we
+  // make it conditionally and depend on whether ShadowDom encapsulation is used (in which case the
```
kara:
```suggestion
  // make it conditional and depend on whether ShadowDom encapsulation is used (in which case the
```

kara (approved):
LGTM, once typo is fixed

PR #33493, opened 2019-10-30
Fw 1559 - more cases

pkozlowski-opensource:
This is a version of #33457 that has more tests and handles more cases.

### DIFF
--- a/packages/core/test/acceptance/template_ref_spec.ts
+++ b/packages/core/test/acceptance/template_ref_spec.ts
@@ -13,47 +13,6 @@ import {onlyInIvy} from '@angular/private/testing';
 
 describe('TemplateRef', () => {
   describe('rootNodes', () => {
-    it('should include projected nodes in rootNodes', () => {
-      @Component({
-        selector: 'menu-content',
-        template: `
-          <ng-template>
-            Header
-            <ng-content></ng-content>
-          </ng-template>
-        `,
-        exportAs: 'menuContent'
-      })
-      class MenuContent {
-        @ViewChild(TemplateRef, {static: true}) template !: TemplateRef<any>;
-      }
-
-      @Component({
-        template: `
-          <menu-content #menu="menuContent">
-            <button>Item one</button>
-            <button>Item two</button>
-          </menu-content>
-        `
-      })
-      class App {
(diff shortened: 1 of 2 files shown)


pkozlowski-opensource:
> core/todo/bundle increased by 126 bytes.

This is mostly due to the fact that we are starting to have code duplication with https://github.com/angular/angular/blob/5666d116336e3e0d5426c5d8f6a91c178d797523/packages/core/src/render3/node_manipulation.ts#L75-L116

In the essence both code-paths are solving the same problem - give me a set of root native nodes. We just happen to do it for different purposes, but the returned value is the same. 

We need to remove this code duplication in a follow-up PR(s)

packages/core/test/acceptance/template_ref_spec.ts
```diff
@@ -54,116 +96,89 @@ describe('TemplateRef', () => {
...
+      if (ivyEnabled) {
+        expect(rootNodes.length).toBe(4);
+        expect(rootNodes[0].nodeType).toBe(Node.COMMENT_NODE);  // ng-container
+        expect(rootNodes[1].nodeType).toBe(Node.TEXT_NODE);     // "Updated " text
+        expect(rootNodes[2].nodeType).toBe(Node.COMMENT_NODE);  // ICU container
+        expect(rootNodes[3].nodeType).toBe(Node.TEXT_NODE);     // "one minute ago" text
+      } else {
+        // ViewEngine seems to produce very different DOM structure as compared to ivy
+        // when it comes to ICU containers - this needs more investigation / fix.
+        expect(rootNodes.length).toBe(7);
```
pkozlowski-opensource: @AndrewKushnir it seems like VE and ivy are generating very different DOM when it comes to ICU expressions. Not sure if this is intended and if not, are we tracking it anywhere?

mhevery (approved):

mhevery (approved):

mhevery:
[presubmit](http://test/OCL:277516975:BASE:277517028:1572451712739:e84ad8db)
[ivy presubmit](http://test/OCL:277516975:BASE:277517370:1572451808830:cc108375)

AndrewKushnir (approved):

PR #33494, opened 2019-10-30
build: restore functionality of the micro benchmarks `profile_all.js` script

matsko:
The 4b81bb5c979382d5f698e0cd729763dd675c226c patch seemingly broke the
`profile_all.js` file due to the file renaming. This patch restores the
functionality of said script.

### DIFF
--- a/packages/core/test/render3/perf/profile_all.js
+++ b/packages/core/test/render3/perf/profile_all.js
@@ -30,7 +30,7 @@ const profileTests =
 // build tests
 shell.exec(
     `yarn bazel build --define=compile=aot ` +
-    profileTests.map((name) => `//packages/core/test/render3/perf:${name}.min_debug.es2015.js`)
+    profileTests.map((name) => `//packages/core/test/render3/perf:${name}_lib.min_debug.es2015.js`)
         .join(' '));
 
 // profile tests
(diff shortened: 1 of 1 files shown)


mhevery (approved):

packages/core/test/render3/perf/profile_all.js
```diff
@@ -30,7 +30,7 @@ const profileTests =
 // build tests
 shell.exec(
     `yarn bazel build --define=compile=aot ` +
-    profileTests.map((name) => `//packages/core/test/render3/perf:${name}.min_debug.es2015.js`)
+    profileTests.map((name) => `//packages/core/test/render3/perf:${name}_lib.min_debug.es2015.js`)
```
IgorMinar:
```suggestion
```

packages/core/test/render3/perf/profile_all.js
```diff
@@ -30,7 +30,7 @@ const profileTests =
 // build tests
 shell.exec(
     `yarn bazel build --define=compile=aot ` +
```
IgorMinar:
```suggestion
    `yarn bazel build --define=compile=aot //packages/core/test/render3/perf/...`
```

IgorMinar (changes requested):
those benchmarks can be built by bazel using a `...` target expansion, no need for extra code here.

packages/core/test/render3/perf/profile_all.js
matsko: Unfortunately this doesn't work since `min_debug` isn't the default target to be built. We will need to loop over and build each target using the loop.

packages/core/test/render3/perf/profile_all.js
matsko: see my comment below

IgorMinar (approved):

PR #33500, opened 2019-10-31
build: add --nolegacy_external_runfiles flag for build

josephperrott:


gregmagolan (approved):
LGTM.

This is the correct setting for legacy external runfiles as the default for the flag will be flipped at some point and legacy external runfiles will be removed altogether at some later point.

The breakage for the release script was that the `$BAZEL run -- ${packageLabel}.pack --access public --tag next` calls forced a rebuild of all packages that were built with `$BAZEL build --config=release $NPM_PACKAGE_LABELS`. However because `--config=release` is not set on that line then the packages were published with `0.0.0`. This seems to be because run was settings `run --nolegacy_external_runfiles` which was not set on `build` and that caused the rebuild.

I tried analyzing when the issue was introduced last night but it seems flaky for me locally and didn't seem related to the Bazel version or the nodejs rules version. It seems to happen every time with Bazel 1.0.0 & nodejs rules 0.39.1 and less with Bazel 0.28.1 and nodejs rules 0.38.3 but I think I observed it with the latter as well. Need to make a minimal reproduction today to understand where it is going wrong.

As a safety for this in the future (incase there are other differences between build & run or maybe some output affecting difference in `--config=release`), it may be good to update the release script to `npm_pakcage.pack` first and then check the result before calling `npm_package.publish`.

josephperrott:
Note: In continued investigation it appears that previously with bazel `0.28.1` (and presumably earlier as well) we still were informed that our cache was being discarded (`Build option --legacy_external_runfiles has changed, discarding analysis cache`), but the cache ends up being used anyway.

So while it seems this change has the effect we desire going forward, it is unclear why it worked in the past.

PR #33508, opened 2019-10-31
feat(common): support loading locales from a global

petebacondarwin:
To support compile time localization, we need to be
able to provide the locales via a well known global property
`ng.common.locale`.

This commit changes `findLocaleData()` so that it will
attempt to read the local from the global if the locale
has not already been registered.

---

I had to touch a number of tests because they were registering locales
and leaving them in place, which affected subsequent tests.


petebacondarwin:
Added to https://github.com/angular/angular/pull/33504

PR #33514, opened 2019-10-31
fix(ngcc): emit correct ES5 code

### LINKED ISSUE #30569: ngcc generates invalid property metadata
# 🐞 bug report

### Affected Package

`@angular/compiler-cli` / `ngcc`

### Description

Currently when compiling `@angular/cdk/a11y` with `ngcc`, the generated property metadata call (`setClassMetadata`) incorrectly sets metadata for the `autoCapture` property within the `FocusTrap` twice.

```ts
ɵngcc0.ɵsetClassMetadata(..., {
  autoCapture: [{
      type: Input,
      args: ['cdkTrapFocusAutoCapture']
    }], 
  autoCapture: [], 
});
```

This means that the first occurrence will be accidentally overwritten and the metadata is discarded. The duplication seems to happen because `autoCapture` declares a `getter` and a `setter`.

As far as I could tell the issue is that `ngcc` returns the getter and setter for `reflection.getMembersOfClass(FocusTrap)` but sets decorators that are either defined on getter or setter as `decorators` for both members. This then results in two properties with the same name in the `setClassMetadata` call.

 https://github.com/angular/angular/blob/master/packages/compiler-cli/src/ngtsc/annotations/src/metadata.ts#L59-L62

## 🔬 Minimal Reproduction

https://github.com/devversion/ivy-ngcc-generates-duplicate-prop-metadata

## 🔥 Exception or Error

See: https://github.com/angular/components/issues/16066

Defining the same property multiple times seems to cause an exception in IE11 with `use strict` mode. This is actually reasonable as the first declaration is accidentally overwritten in an object literal.

## 🌍  Your Environment

**Angular Version:**
<pre><code>
@angular/*: 8.0.0-rc.4
@angular/cdk: 8.0.0-rc.1
</code></pre>

### LINKED ISSUE #32665: Ngcc doesn't produce valid es5 code
# 🐞 bug report

### Affected Package

@angular/forms, @angular/common

### Is this a regression?

 Yes, it works without ivy

### Description

Code produced by ngcc is not valid es5.
Examples:
Fragment of node_modules\@angular\forms\__ivy_ngcc__\fesm5\forms.js 
```
const _c0 = ["ng-untouched", "ng-touched", "ng-pristine", "ng-dirty", "ng-valid", "ng-invalid", "ng-pending"];
const _c1 = ["novalidate", ""];
```
const!
Fragment of node_modules\@angular\common\__ivy_ngcc__\fesm5\common.js 
```
NgForOf.ngDirectiveDef = ɵngcc0.ɵɵdefineDirective({ type: NgForOf, selectors: [["", "ngFor", "", "ngForOf", ""]], factory: function NgForOf_Factory(t) { return new (t || NgForOf)(ɵngcc0.ɵɵdirectiveInject(ViewContainerRef), ɵngcc0.ɵɵdirectiveInject(TemplateRef), ɵngcc0.ɵɵdirectiveInject(IterableDiffers)); }, inputs: { ngForOf: "ngForOf", ngForTrackBy: "ngForTrackBy", ngForTemplate: "ngForTemplate" } });
/*@__PURE__*/ ɵngcc0.ɵsetClassMetadata(NgForOf, [{
        type: Directive,
        args: [{ selector: '[ngFor][ngForOf]' }]
    }], function () { return [{ type: ViewContainerRef }, { type: TemplateRef }, { type: IterableDiffers }]; }, { _viewContainer: [], _template: [], _differs: [], _ngForOfDirty: [], _differ: [], ngForOf: [{
            type: Input
        }], ngForTrackBy: [{
            type: Input
        }], ngForTrackBy: [], ngForTemplate: [{
            type: Input
        }], ngDoCheck: [], _applyChanges: [], _perViewChange: [] });
    return NgForOf;
}());
```
Notice double ngForTrackBy. PhantomJS thinks it's Syntax Error.

## 🔬 Minimal Reproduction
Compile any project which used ngFor or NG_VALUE_ACCESSOR 

## 🌍  Your Environment

**Angular Version:**
```json
{
  "dependencies": {
    "@angular/animations": "8.1.1",
    "@angular/common": "8.1.1",
    "@angular/compiler": "8.1.1",
    "@angular/core": "8.1.1",
    "@angular/forms": "8.1.1",
    "@angular/platform-browser": "8.1.1",
    "@angular/platform-browser-dynamic": "8.1.1",
    "@angular/router": "8.1.1",
    "core-js": "2.4.1",
    "rxjs": "6.4.0",
    "rxjs-compat": "6.4.0",
    "tslib": "1.9.1",
    "zone.js": "0.9.1"
  },
  "devDependencies": {
    "@angular/cli": "8.1.1",
    "@angular/compiler-cli": "8.1.1",
    "@angular/language-service": "8.1.1",
    "@ngtools/webpack": "^8.1.1",
    "codelyzer": "^5.0.0",
    "gulp": "^4.0.2",
    "typescript": "3.4.5",
    "webpack": "^4.35.2"
  }
}
```

**Anything else relevant?**

gkalpak:
The PR includes the following commits (see individual commit messages for more info):

- **refactor(compiler-cli): avoid superfluous parenthesis around statements**
- **fix(ngcc): generate correct metadata for classes with getter/setter properties**
- **fix(ngcc): do not emit ES2015 code in ES5 files**

Fixes #30569, fixes #32665.

### DIFF
--- a/packages/compiler-cli/src/ngtsc/translator/src/translator.ts
+++ b/packages/compiler-cli/src/ngtsc/translator/src/translator.ts
@@ -18,7 +18,7 @@ export class Context {
 
   get withExpressionMode(): Context { return this.isStatement ? new Context(false) : this; }
 
-  get withStatementMode(): Context { return this.isStatement ? new Context(true) : this; }
+  get withStatementMode(): Context { return !this.isStatement ? new Context(true) : this; }
 }
 
 const BINARY_OPERATORS = new Map<BinaryOperator, ts.BinaryOperator>([
--- a/packages/compiler-cli/src/ngtsc/translator/src/translator.ts
+++ b/packages/compiler-cli/src/ngtsc/translator/src/translator.ts
@@ -100,17 +100,19 @@ export class ImportManager {
 }
 
 export function translateExpression(
-    expression: Expression, imports: ImportManager,
-    defaultImportRecorder: DefaultImportRecorder): ts.Expression {
+    expression: Expression, imports: ImportManager, defaultImportRecorder: DefaultImportRecorder,
+    scriptTarget: Exclude<ts.ScriptTarget, ts.ScriptTarget.JSON>): ts.Expression {
   return expression.visitExpression(
-      new ExpressionTranslatorVisitor(imports, defaultImportRecorder), new Context(false));
+      new ExpressionTranslatorVisitor(imports, defaultImportRecorder, scriptTarget),
+      new Context(false));
 }
 
 export function translateStatement(
-    statement: Statement, imports: ImportManager,
-    defaultImportRecorder: DefaultImportRecorder): ts.Statement {
+    statement: Statement, imports: ImportManager, defaultImportRecorder: DefaultImportRecorder,
+    scriptTarget: Exclude<ts.ScriptTarget, ts.ScriptTarget.JSON>): ts.Statement {
   return statement.visitStatement(
-      new ExpressionTranslatorVisitor(imports, defaultImportRecorder), new Context(true));
+      new ExpressionTranslatorVisitor(imports, defaultImportRecorder, scriptTarget),
+      new Context(true));
 }
 
 export function translateType(type: Type, imports: ImportManager): ts.TypeNode {
@@ -120,10 +122,14 @@ export function translateType(type: Type, imports: ImportManager): ts.TypeNode {
 class ExpressionTranslatorVisitor implements ExpressionVisitor, StatementVisitor {
   private externalSourceFiles = new Map<string, ts.SourceMapSource>();
   constructor(
-      private imports: ImportManager, private defaultImportRecorder: DefaultImportRecorder) {}
+      private imports: ImportManager, private defaultImportRecorder: DefaultImportRecorder,
+      private scriptTarget: Exclude<ts.ScriptTarget, ts.ScriptTarget.JSON>) {}
 
   visitDeclareVarStmt(stmt: DeclareVarStmt, context: Context): ts.VariableStatement {
-    const nodeFlags = stmt.hasModifier(StmtModifier.Final) ? ts.NodeFlags.Const : ts.NodeFlags.None;
+    const nodeFlags =
+        ((this.scriptTarget >= ts.ScriptTarget.ES2015) && stmt.hasModifier(StmtModifier.Final)) ?
+        ts.NodeFlags.Const :
+        ts.NodeFlags.None;
     return ts.createVariableStatement(
         undefined, ts.createVariableDeclarationList(
                        [ts.createVariableDeclaration(
@@ -149,6 +155,11 @@ class ExpressionTranslatorVisitor implements ExpressionVisitor, StatementVisitor
   }
 
   visitDeclareClassStmt(stmt: ClassStmt, context: Context) {
+    if (this.scriptTarget < ts.ScriptTarget.ES2015) {
+      throw new Error(
+          `Unsupported mode: Visiting a "declare class" statement (class ${stmt.name}) while ` +
+          `targeting ${ts.ScriptTarget[this.scriptTarget]}.`);
+    }
     throw new Error('Method not implemented.');
   }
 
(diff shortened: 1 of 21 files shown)


gkalpak:
FWIW, I also tried it on the [ngcc-validation][1] repo and [nothing broke][2].

[1]: https://github.com/angular/ngcc-validation
[2]: https://circleci.com/gh/angular/ngcc-validation/1379

packages/compiler-cli/src/ngtsc/translator/src/translator.ts
```diff
@@ -100,17 +100,19 @@ export class ImportManager {
...
-      new ExpressionTranslatorVisitor(imports, defaultImportRecorder), new Context(false));
+      new ExpressionTranslatorVisitor(imports, defaultImportRecorder, targetEs5),
+      new Context(false));
 }
 
 export function translateStatement(
-    statement: Statement, imports: ImportManager,
-    defaultImportRecorder: DefaultImportRecorder): ts.Statement {
+    statement: Statement, imports: ImportManager, defaultImportRecorder: DefaultImportRecorder,
+    targetEs5 = false): ts.Statement {
```
alxhub: Instead of an optional `boolean` parameter, let's pass a mandatory `ts.ScriptTarget` here.

alxhub (changes requested):

packages/compiler-cli/src/ngtsc/translator/src/translator.ts
gkalpak: That was my initial intention, but things would get a little confusing imo. `ScriptTarget` can have many values (some of which don't even make sense) and more will be added over time:

```ts
enum ScriptTarget {
  ES3 = 0,
  ES5 = 1,
  ES2015 = 2,
  ES2016 = 3,
  ES2017 = 4,
  ES2018 = 5,
  ES2019 = 6,
  ES2020 = 7,
  ESNext = 99,
  JSON = 100,
  Latest = 99
}
```

I don't see any benefit in passing a `ScriptTarget` and then basically trying to infer whether we should be targeting ES5 or note. (Theoretically, we could use a `< ts.ScriptTarget.ES2015` comparison, but this feels like a TypeScript implementation detail.)

Or am I missing something and we can use `ts.ScriptTarget` in `translateStatement`/`translateExpression` directly?

petebacondarwin: `target === ScriptTarget.ES5`?
Why need `< ScriptTarget.ES2015`?

gkalpak: What is the benefit of passing `ts.ScriptTarget` if we only expect (and handle) `ts.ScriptTarget.ES5`?
It seems confusing to me. Do I miss something?

petebacondarwin: I could imagine in the future making further customisations to the AST based on the `ScriptTarget` being say `ES2015`. By passing in the whole target you are letting the `ExpressionTranslationVisitor` be responsible for what it does with it, rather than exposing the fact that, at the moment, it only differs if it is `ES5`.

alxhub: > Theoretically, we could use a `< ts.ScriptTarget.ES2015` comparison, but this feels like a TypeScript implementation detail.

This is actually what TS does internally, so I think it is an intentional feature of `ts.ScriptTarget` and one that's okay for us to rely on.

petebacondarwin: An alternative approach to this would be to have a new class `Es5ExpressionTranslatorVisitor` which is specifically tailored to the needs of translating to `ES5`.  It would inherit from `ExpressionTranslatorVisitor ` and override `visitDeclareVarStmt` and `visitLocalizedString`.

gkalpak: I switched to mandatory `ts.ScriptTarget` and a `< ts.ScriptTarget.ES2015` check.
Happy to implement an `Es5ExpressionTranslatorVisitor` instead.

petebacondarwin (approved):

alxhub (approved):

kara:
@gkalpak Looks like this needs a rebase

gkalpak:
@kara, rebased.

kara:
@gkalpak Needs one more rebase, sorry!

kara:
[VE presubmit](http://test/OCL:280245223:BASE:280245266:1573674438319:3508fe20)

[Ivy presubmit](http://test/OCL:280245223:BASE:280245457:1573674492617:bccd28a3)

PR #33519, opened 2019-10-31
ci: always run the publish_packages_as_artifacts jobs

josephperrott:
We should always run the `publish_packages_as_artifacts` job since even on `master` there is value in being able to access the artifacts from committed code as well.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -844,8 +844,6 @@ workflows:
           requires:
             - build-npm-packages
       - publish_packages_as_artifacts:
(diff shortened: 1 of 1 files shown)


gkalpak (approved):

devversion (approved):

AndrewKushnir:
@josephperrott could you plz confirm whether this PR should go to patch branch as well? Thank you.

josephperrott:
Yes, we would like for this to go in both `master` and `patch`

PR #33520, opened 2019-10-31
docs: Adds platform-specific dependencies.

dgp1130:
I was going through the process of getting setup on my Mac laptop and found that builds were giving strange errors which I ultimately narrowed down to an outdated XCode install. XCode is necessary for Bazel on Mac, so this PR adds it as a prerequisite in the documentation.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

XCode is not listed as a prerequisite.

Issue Number: N/A

## What is the new behavior?

XCode is now a prerequisite.

- [ ] Yes
- [X] No

### DIFF
--- a/docs/DEVELOPER.md
+++ b/docs/DEVELOPER.md
@@ -32,6 +32,7 @@ following products on your development machine:
 
 * [Java Development Kit](https://www.oracle.com/technetwork/java/javase/downloads/index.html) which is used
   to execute the selenium standalone server for e2e testing.
+* [XCode](https://developer.apple.com/xcode/) (if developing on Mac) which provides compilers/toochains/SDKs for the build system.
 
 ## Getting the Sources
 
--- a/docs/DEVELOPER.md
+++ b/docs/DEVELOPER.md
@@ -32,7 +32,26 @@ following products on your development machine:
 
 * [Java Development Kit](https://www.oracle.com/technetwork/java/javase/downloads/index.html) which is used
   to execute the selenium standalone server for e2e testing.
-* [XCode](https://developer.apple.com/xcode/) (if developing on Mac) which provides compilers/toochains/SDKs for the build system.
+
+### Platform Specific Prerequisites
+
+Each platform has some specific prerequisites. Make sure to set up everything necessary for your particular environment.
+
+#### Ubuntu
+
+Install `pkg-config`, `zip`, `g++`, `zlib1g-dev`, `unzip` and `python3`.
+
+```shell
+sudo apt-get install pkg-config zip g++ zlib1g-dev unzip python3
+```
+
+#### Mac
+
+Install [XCode](https://developer.apple.com/xcode/), which provides compilers/toochains/SDKs for the build system.
+
+#### Windows
+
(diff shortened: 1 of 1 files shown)


docs/DEVELOPER.md
```diff
@@ -32,6 +32,7 @@ following products on your development machine:
 
 * [Java Development Kit](https://www.oracle.com/technetwork/java/javase/downloads/index.html) which is used
   to execute the selenium standalone server for e2e testing.
+* [XCode](https://developer.apple.com/xcode/) (if developing on Mac) which provides compilers/toochains/SDKs for the build system.
```
devversion: Would it make more sense to refer to the Bazel install requirements instead? e.g. https://docs.bazel.build/versions/master/install-os-x.html mentions xcode being a requirement.

dgp1130: I thought of that too. From a technical perspective it makes sense to tell users to "install all Bazel prerequisites", that way we don't need to keep this line updated if Bazel's requirements change. However I would anticipate that many readers might be confused by this and end up installing Bazel (and/or Homebrew and Ruby, which is one means of installation) which is not actually necessary.

IMHO, it's just simpler to tell readers to install XCode, rather than trying to articulate that only Bazel prerequisites are required. This is easier for the reader, but has the cost of possibly falling out of date and needing to be updated occasionally in the future. Thoughts?

kapunahelewong (approved):

docs/DEVELOPER.md
devversion: That is a fair point. I totally agree that it could lead people to installing Bazel completely. 

I'm slightly unsure if just adding `XCode` is helpful. e.g. on Windows, people also need to install Msys2 etc. These things would be covered by referring to the general Bazel prerequisites (which also exist for MacOS, Windows, Linux).

Though this might be something we can do in the future. Adding `xcode` definitely already helps 😄

dgp1130: Hmm, looking through the Bazel docs, they don't really have a "prerequisites" section. Each platform just includes the necessary dependencies. We don't really have a good page or section to link to here. We could maybe ask the Bazel team to refactor this to have prerequisites on a specific page/section which is more linkable, but I'm not sure how hard it would be to make that happen, especially since Bazel docs are targeted at developers using Bazel, while this is just internal Angular docs.

We could make a separate section for platform-specific dependencies and then just copy the ones from Bazel, which would apply this a little bit beyond just XCode for Macs. Thoughts?

devversion: Yeah, looks like there is no page we can link to without risking that people install Bazel globally. A separate section like you proposed would sound like a good solution.

dgp1130: Added a separate section for platform-specific prerequisites, and then just copied the requirements from Bazel.

I skipped the Fedora/CentOS instructions because they seemed specific to setting up the package manager, which I don't think applies in this case.

devversion (approved):

PR #33521, opened 2019-10-31
refactor(ivy): remove TStylingContext locking in favor of firstUpdatePass flag

matsko:
This patch removes the need to lock the style and class context
instances to track when bindings can be added. What happens now is
that the `tNode.firstUpdatePass` is used to track when bindings are
registered on the context instances.

### DIFF
--- a/packages/core/src/render3/instructions/styling.ts
+++ b/packages/core/src/render3/instructions/styling.ts
@@ -256,7 +256,8 @@ export function ɵɵstyleMap(styles: {[styleName: string]: any} | NO_CHANGE | nu
     styles = NO_CHANGE;
   }
 
-  stylingMap(context, tNode, tView, lView, bindingIndex, styles, false, hasDirectiveInput);
+  stylingMap(
+      context, tNode, tView.firstUpdatePass, lView, bindingIndex, styles, false, hasDirectiveInput);
 }
 
 /**
@@ -310,7 +311,8 @@ export function classMapInternal(
     classes = NO_CHANGE;
   }
 
-  stylingMap(context, tNode, tView, lView, bindingIndex, classes, true, hasDirectiveInput);
+  stylingMap(
+      context, tNode, tView.firstUpdatePass, lView, bindingIndex, classes, true, hasDirectiveInput);
 }
 
 /**
@@ -320,15 +322,14 @@ export function classMapInternal(
  * `[class]` bindings in Angular.
  */
 function stylingMap(
-    context: TStylingContext, tNode: TNode, tView: TView, lView: LView, bindingIndex: number,
-    value: {[key: string]: any} | string | null, isClassBased: boolean,
+    context: TStylingContext, tNode: TNode, firstUpdatePass: boolean, lView: LView,
+    bindingIndex: number, value: {[key: string]: any} | string | null, isClassBased: boolean,
     hasDirectiveInput: boolean): void {
   const directiveIndex = getActiveDirectiveId();
   const native = getNativeByTNode(tNode, lView) as RElement;
   const oldValue = getValue(lView, bindingIndex);
   const sanitizer = getCurrentStyleSanitizer();
   const valueHasChanged = hasValueChanged(oldValue, value);
-  const firstUpdatePass = tView.firstUpdatePass;
 
   // [style] and [class] bindings do not use `bind()` and will therefore
   // manage accessing and updating the new value in the lView directly.
--- a/packages/core/src/render3/instructions/styling.ts
+++ b/packages/core/src/render3/instructions/styling.ts
@@ -154,10 +154,9 @@ function stylingProp(
   let updated = false;
 
   const lView = getLView();
+  const firstUpdatePass = lView[TVIEW].firstUpdatePass;
   const tNode = getTNode(elementIndex, lView);
-  const tView = lView[TVIEW];
   const native = getNativeByTNode(tNode, lView) as RElement;
-  const firstUpdatePass = tView.firstUpdatePass;
   const context = isClassBased ? getClassesContext(tNode) : getStylesContext(tNode);
   const sanitizer = isClassBased ? null : getCurrentStyleSanitizer();
 
@@ -237,7 +236,7 @@ export function ɵɵstyleMap(styles: {[styleName: string]: any} | NO_CHANGE | nu
   const index = getSelectedIndex();
   const lView = getLView();
   const tNode = getTNode(index, lView);
-  const tView = lView[TVIEW];
+  const firstUpdatePass = lView[TVIEW].firstUpdatePass;
   const context = getStylesContext(tNode);
   const hasDirectiveInput = hasStyleInput(tNode);
 
@@ -251,13 +250,12 @@ export function ɵɵstyleMap(styles: {[styleName: string]: any} | NO_CHANGE | nu
   // there should not be a situation where a directive host bindings function
   // evaluates the inputs (this should only happen in the template function)
   if (!isHostStyling() && hasDirectiveInput && styles !== NO_CHANGE) {
-    updateDirectiveInputValue(
-        context, lView, tNode, bindingIndex, styles, false, tView.firstUpdatePass);
+    updateDirectiveInputValue(context, lView, tNode, bindingIndex, styles, false, firstUpdatePass);
     styles = NO_CHANGE;
   }
 
   stylingMap(
-      context, tNode, tView.firstUpdatePass, lView, bindingIndex, styles, false, hasDirectiveInput);
+      context, tNode, firstUpdatePass, lView, bindingIndex, styles, false, hasDirectiveInput);
 }
 
 /**
@@ -292,7 +290,7 @@ export function classMapInternal(
     elementIndex: number, classes: {[className: string]: any} | NO_CHANGE | string | null): void {
   const lView = getLView();
   const tNode = getTNode(elementIndex, lView);
-  const tView = lView[TVIEW];
+  const firstUpdatePass = lView[TVIEW].firstUpdatePass;
   const context = getClassesContext(tNode);
   const hasDirectiveInput = hasClassInput(tNode);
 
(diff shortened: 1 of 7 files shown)


matsko:
Presubmit = http://test/OCL:278486043:BASE:278486088:1572912430975:d781cb2

Ivy Presubmit = https://test.corp.google.com/ui#id=OCL:278486043:BASE:278486218:1572912468593:40ac370f

packages/core/src/render3/util/styling_utils.ts
```diff
@@ -93,8 +92,8 @@ export function allowDirectStyling(context: TStylingContext, hostBindingsMode: b
     // `ngDevMode` is required to be checked here because tests/debugging rely on the context being
     // populated. If things are in production mode then there is no need to build a context
     // therefore the direct apply can be allowed (even on the first update).
-    allow = ngDevMode ? contextIsLocked : true;
-  } else if (contextIsLocked) {
+    allow = ngDevMode ? !firstUpdatePass : true;
```
mhevery: This line and comment looks suspicious. It seems to me that we are changing runtime behavior depending on the `ngDevMode`? That does not seem right to me. Can you better explain to me what is going on?

packages/core/src/render3/styling/bindings.ts
```diff
@@ -441,35 +438,21 @@ export function flushStyling(
...
- * and another for host bindings. Either one of these locks will be set when styling is applied
- * during the template binding flush and/or during the host bindings flush.
+ * Directives and component host bindings may include static class/style values which are
+ * bound to the host element. When this happens, the styling context will need to be informed
+ * so it can use these static styling values as defaults when a matching binding is falsy.
+ * These initial styling values are read from the initial styling values slot within the
+ * provided `TStylingContext` (which is an instance of a `StylingMapArray`). This inner map will
+ * be updated each time a host binding applies its static styling values (via `elementHostAttrs`)
+ * so these values are only read at this point because this is the very last point before the
+ * first style/class values are flushed to the element.
```
mhevery: The above is confusing. Could we have a small code snippet / Concrete example showing why this is needed.

mhevery (changes requested):

packages/core/src/render3/util/styling_utils.ts
matsko: We do not need to populate a context whilst in prod mode. The only only reason we do this is in devMode is so that `debugNode.styles.debug` and `debugNode.classes.debug` can properly list out all classes/styles on the element (otherwise it doesn't know because the direct styling apply mode doesn't keep track of anything).

packages/core/src/render3/styling/bindings.ts
matsko: updated

packages/core/src/render3/util/styling_utils.ts
```diff
@@ -79,10 +79,9 @@ export function hasConfig(context: TStylingContext, flag: TStylingConfig) {
  * 3. There are no collisions (i.e. properties with more than one binding) across multiple
  *    sources (i.e. template + directive, directive + directive, directive + component)
  */
-export function allowDirectStyling(context: TStylingContext, hostBindingsMode: boolean): boolean {
+export function allowDirectStyling(context: TStylingContext, firstUpdatePass: boolean): boolean {
```
kara: We should update the comment above to remove references to locking and `hostBindingsMode`

packages/core/src/render3/styling/styling_debug.ts
```diff
@@ -496,19 +494,13 @@ function buildConfig(context: TStylingContext) {
   const hasCollisions = hasConfig(context, TStylingConfig.HasCollisions);
   const hasTemplateBindings = hasConfig(context, TStylingConfig.HasTemplateBindings);
   const hasHostBindings = hasConfig(context, TStylingConfig.HasHostBindings);
-  const templateBindingsLocked = hasConfig(context, TStylingConfig.TemplateBindingsLocked);
-  const hostBindingsLocked = hasConfig(context, TStylingConfig.HostBindingsLocked);
-  const allowDirectStyling =
-      _allowDirectStyling(context, false) || _allowDirectStyling(context, true);
-
+  const allowDirectStyling = _allowDirectStyling(context, false);
```
kara: Why is `firstUpdatePass` always false here?  Comment?

packages/core/src/render3/styling/bindings.ts
```diff
@@ -441,35 +438,62 @@ export function flushStyling(
...
+ *
+ * ```html
+ * <div style="color:red"
+ *     [style.color]="myColor"
+ *     dir-that-has-static-height>
+ *   ...
+ * </div>
+ * ```
+ *
+ * When this gets compiled the `tNode.styles` value will look like so:
```
kara: Nit: This almost sounds like `tNode.styles` is code generated by the compiler. 

```suggestion
 * When the initial styles are processed, the `tNode.styles` value will look like so:
```

packages/core/src/render3/styling/bindings.ts
```diff
@@ -441,35 +438,62 @@ export function flushStyling(
...
+ * `InitialStylingValuePosition` slot.
+ *
+ * Because all static styles/classes are collected and registered on
+ * the initial styling array each time a directive is instantiated,
+ * the context may not yet know about the static values. When this
+ * function is called it will copy over all the static style/class
+ * values from the initial styling array into the context as default
+ * values for each of the matching entries in the context.
+ *
+ * Let's imagine the following example:
```
kara: Examples are super helpful 👍

packages/core/src/render3/instructions/styling.ts
```diff
@@ -302,11 +305,12 @@ export function classMapInternal(
...
   // evaluates the inputs (this should only happen in the template function)
   if (!isHostStyling() && hasDirectiveInput && classes !== NO_CHANGE) {
-    updateDirectiveInputValue(context, lView, tNode, bindingIndex, classes, true);
+    updateDirectiveInputValue(
+        context, lView, tNode, bindingIndex, classes, true, tView.firstUpdatePass);
     classes = NO_CHANGE;
   }
 
-  stylingMap(context, tNode, lView, bindingIndex, classes, true, hasDirectiveInput);
+  stylingMap(context, tNode, tView, lView, bindingIndex, classes, true, hasDirectiveInput);
```
kara: Why not pass through `tView.firstUpdatePass`  at this level? That way, we can avoid reading the property twice (once in line 309 and once in `stylingMap`).
```suggestion
  stylingMap(context, tNode, firstUpdatePass, lView, bindingIndex, classes, true, hasDirectiveInput);
```

kara (changes requested):
Minor comments

packages/core/src/render3/util/styling_utils.ts
mhevery: Does this not imply that we could accidentally rely on that information in `ngDevMode`??? I think it would be better if we did not have such code.

packages/core/src/render3/instructions/styling.ts
```diff
@@ -250,11 +251,13 @@ export function ɵɵstyleMap(styles: {[styleName: string]: any} | NO_CHANGE | nu
   // there should not be a situation where a directive host bindings function
   // evaluates the inputs (this should only happen in the template function)
   if (!isHostStyling() && hasDirectiveInput && styles !== NO_CHANGE) {
-    updateDirectiveInputValue(context, lView, tNode, bindingIndex, styles, false);
+    updateDirectiveInputValue(
+        context, lView, tNode, bindingIndex, styles, false, tView.firstUpdatePass);
```
kara: We should move the property access to line 240 so we don't have to read it twice (once here and once on line 257)
```suggestion
        context, lView, tNode, bindingIndex, styles, false, firstUpdatePass);
```

packages/core/src/render3/instructions/styling.ts
```diff
@@ -237,6 +237,7 @@ export function ɵɵstyleMap(styles: {[styleName: string]: any} | NO_CHANGE | nu
   const index = getSelectedIndex();
   const lView = getLView();
   const tNode = getTNode(index, lView);
+  const tView = lView[TVIEW];
```
kara:
```suggestion
  const firstUpdatePass = lView[TVIEW].firstUpdatePass;
```

packages/core/src/render3/instructions/styling.ts
```diff
@@ -250,11 +251,13 @@ export function ɵɵstyleMap(styles: {[styleName: string]: any} | NO_CHANGE | nu
...
   if (!isHostStyling() && hasDirectiveInput && styles !== NO_CHANGE) {
-    updateDirectiveInputValue(context, lView, tNode, bindingIndex, styles, false);
+    updateDirectiveInputValue(
+        context, lView, tNode, bindingIndex, styles, false, tView.firstUpdatePass);
     styles = NO_CHANGE;
   }
 
-  stylingMap(context, tNode, lView, bindingIndex, styles, false, hasDirectiveInput);
+  stylingMap(
+      context, tNode, tView.firstUpdatePass, lView, bindingIndex, styles, false, hasDirectiveInput);
```
kara:
```suggestion
      context, tNode, firstUpdatePass, lView, bindingIndex, styles, false, hasDirectiveInput);
```

packages/core/src/render3/instructions/styling.ts
```diff
@@ -302,11 +306,13 @@ export function classMapInternal(
   // there should not be a situation where a directive host bindings function
   // evaluates the inputs (this should only happen in the template function)
   if (!isHostStyling() && hasDirectiveInput && classes !== NO_CHANGE) {
-    updateDirectiveInputValue(context, lView, tNode, bindingIndex, classes, true);
+    updateDirectiveInputValue(
+        context, lView, tNode, bindingIndex, classes, true, tView.firstUpdatePass);
```
kara: Same here. We can avoid reading the property twice

```suggestion
        context, lView, tNode, bindingIndex, classes, true, firstUpdatePass);
```

mhevery (approved):
Please take care of @kara comments

packages/core/src/render3/instructions/styling.ts
matsko: fixed

packages/core/src/render3/instructions/styling.ts
matsko: fixed

packages/core/src/render3/instructions/styling.ts
matsko: fixed

packages/core/src/render3/instructions/styling.ts
matsko: fixed

packages/core/src/render3/instructions/styling.ts
matsko: fixed

packages/core/src/render3/util/styling_utils.ts
matsko: The fix for this is to design a debugging layer that works independent of the underling style/class application algorithms.

kara (approved):
LGTM

matsko:
presubmit = http://test/OCL:278486043:BASE:278720559:1572996273861:979d4f59

ivy presubmit = https://test.corp.google.com/ui#id=OCL:278486043:BASE:278720755:1572996333664:1410ab39

PR #23265, opened 2018-04-09
Fix for i18n locale data undefined values (+update CLDR data)

ocombe:
## PR Type
What kind of change does this PR introduce?

```
[x] Bugfix
```

## What is the current behavior?
We remove `undefined` values in CLDR data

Issue Number: #22988

## What is the new behavior?
We replace `undefined` values by a constant instead

## Does this PR introduce a breaking change?
```
[x] No
```

## Other information
I've also updated the CLDR data to v33 that was just released last week.

PR #33523, opened 2019-10-31
i18n - support global locale data

petebacondarwin:
This is my preferred solution to generating (and using) locales via the global object.

See https://hackmd.io/BFc4P-JYT7-VO0id7-afhg

**It is best to review commit by commit, since one simply adds over 500 newly generated locale files.**

The idea is that at translation time (after the build) the required locale data is included in what will be executed by the browser.

* In the CLI
  * The appropriate locale data file will be pre-prended as-is to the `main.js` bundle at translation time (after the main build).
* Outside of CLI:
  * The developer should place a `<script src="locale.js">` tag in the `index.html` before the `main.js` script tag
  * The translation tool should be configured to replace that file with the appropriate locale data file.

Executing this script will attach the locale data to the global scope as `ng.common.locale['xx-xx']` making it available to the application.

At runtime, when Angular requires locale data, the `findLocaleData()` function is called. This function returns locales that have been explicitly registered, the same as before v9, by a call to `registerLocaleData()` but now it will also fallback to looking for the locale data on the global scope, `ng.common.locale[..]`.

---

The previous PR #33504 simply updated the current locale files, which are UMD, to support being attached to global scope. This is not optimal for a number of reasons:

* The locale data lookup is based on a normalized form of the locale name, which is not included in the UMD files. This meant that the global lookup had to do an O(n) search of the global locales to match the normalized locale name.
* The UMD locale files contain a large amount of boilerplate to handle the different usage scenarios. Also the current locale files are split into "basic" and "extra" bundles, each containing the boilerplate. Since the locale scripts must be applied during translation, after the build has completed, there is little chance to compact these files and so the resulting bundle sizes (e.g. of `main.js`) are increased significantly.

---

This PR takes a different approach. The locale data files are generated from CLDR data on a periodic basis. Therefore it is straightforward to generate additional locale files that are optimised for the global attachment use case that is required here. These new files have the following properties:

* Each "global" locale data file contains both the "basic" and "extra" data in a single file, pre-arranged in the format that `findLocaleData()` requires.
* The files do not contain any general UMD boilerplate. They contain a simple IIFE that attaches the locale data to the appropriate global scope.
* The name of the property on the global scope is the "normalized" locale name, making it trivial to retrieve the locale data in the `findLocaleData()` function.


packages/common/locales/global/af-NA.js
```diff
@@ -0,0 +1,75 @@
...
+// THIS CODE IS GENERATED - DO NOT MODIFY
+// See angular/tools/gulp-tasks/cldr/extract.js
+
+(function() {
+  const root = typeof globalThis !== 'undefined' && globalThis ||
+      typeof global !== 'undefined' && global || typeof window !== 'undefined' && window;
+  root.ng = root.ng || {};
+  root.ng.common = root.ng.common || {};
+  root.ng.common.locale = root.ng.common.locale || {};
+  const u = undefined;
```
alfaproject: I'm fairly certain that the compression is better if you don't capture `undefined`

petebacondarwin: Yeah, this is a bit of strange one. It was originally added here https://github.com/angular/angular/pull/23265.
Not 100% sure behind the reasoning for using a constant `u`. 
Perhaps @ocombe can explain?

ocombe: So initially it was because we removed the undefined values in the arrays (because they are not necessary): `[1, , ,3]` is valid, and the files were smaller
But then some people complained that it was causing errors in their linting or something like that, so we were forced to re-add the undefined values
We did some comparisons and since all the files are independent (they are not always bundled, some people lazy load them), it was smaller to just defined a constant equal to undefined.
Since them the compilers might have gotten smarter, maybe it isn't necessary anymore, you can try and benchmark that :)

petebacondarwin: Thanks @ocombe for the explanation. Data is king! We should indeed measure it.

As it turns out this PR is generating JS files that will not be further processed, so I am thinking that the generation of those files should go through terser to get them as compact as possible.

tools/gulp-tasks/cldr/extract.js
```diff
@@ -510,30 +547,27 @@ function toRegExp(s) {
...
-    .replace(
-      toRegExp('function anonymous(n\n/*``*/) {\n'),
-      'function plural(n: number): number {\n  ')
-    .replace(toRegExp('var'), 'let')
-    .replace(toRegExp('if(typeof n==="string")n=parseInt(n,10);'), '')
-    .replace(toRegExp('\n}'), ';\n}');
+  const numberType = withTypes ? ': number' : '';
+  fn = fn.replace(/function anonymous\(n[^}]+{/g, `function plural(n${numberType})${numberType} {`)
+           .replace(toRegExp('var'), 'let')
+           .replace(toRegExp('if(typeof n==="string")n=parseInt(n,10);'), '')
```
trotyl: Any reason for checking `string` given the only allowed type is `number`?

ocombe:
Yes, but it's also nice to be able to open the files and see the content,
which is not possible if it's all minified...

>

tools/gulp-tasks/cldr/extract.js
petebacondarwin: Note that this replacement is "removing" that check. It is in the original CLDR function and we are removing it here.

petebacondarwin:
@ocombe these files are pretty obscure even when they are not minified :-)

clydin:
This appears to address all the concerns I had.  I’ll give it a more thorough review today but nice job putting this together.

tools/gulp-tasks/cldr/extract.js
```diff
@@ -72,62 +78,104 @@ module.exports = (gulp, done) => {
...
+            `${I18N_CORE_FOLDER}/locale_en.ts`
+          ],
+          {base: '.'})
+      .pipe(format.format('file', clangFormat))
+      .pipe(gulp.dest('.'));
 };
 
 /**
- * Generate file that contains basic locale data
+ * Generate file that contains all locale data stored in a global
```
petebacondarwin: Oops this jsdoc is on the wrong function.

tools/gulp-tasks/cldr/extract.js
```diff
@@ -72,62 +78,104 @@ module.exports = (gulp, done) => {
...
+`;
+}
+
+function generateGlobalLocale(locale, localeData, baseCurrencies) {
+  const basicLocaleData = generateBasicLocaleString(locale, localeData, baseCurrencies);
+  const extraLocaleData = generateDayPeriodsSupplementalString(locale, localeData);
+  const data = basicLocaleData.replace(/\]$/, `, ${extraLocaleData}]`);
+  return `${HEADER}
+(function() {
+  const root = typeof globalThis !== 'undefined' && globalThis || typeof global !== 'undefined' && global || typeof window !== 'undefined' && window;
```
mhevery: Could we re-arrange this code slightly?
```
(function (global) {
  global.ng = global.ng || {};
 ...
})(typeof globalThis !== 'undefined' && globalThis || typeof global !== 'undefined' && global || typeof window !== 'undefined' && window);
```

tools/gulp-tasks/cldr/extract.js
```diff
@@ -72,62 +78,104 @@ module.exports = (gulp, done) => {
...
-    ], {base: '.'})
-    .pipe(format.format('file', clangFormat))
-    .pipe(gulp.dest('.'));
+      .src(
+          [
+            `${I18N_DATA_FOLDER}/**/*.ts`, `${I18N_FOLDER}/currencies.ts`,
+            `${I18N_CORE_FOLDER}/locale_en.ts`
+          ],
+          {base: '.'})
+      .pipe(format.format('file', clangFormat))
```
mhevery: I see that you are doing `clang-format` here, but I don't see the output file actually being formatted. Something is going on here.

tools/gulp-tasks/cldr/extract.js
```diff
@@ -136,8 +184,7 @@ function generateLocaleExtra(locale, localeData) {
     }
 
     const dayPeriodsFormat = removeDuplicates([
-      objectValues(dayPeriods.format.narrow),
-      objectValues(dayPeriods.format.abbreviated),
+      objectValues(dayPeriods.format.narrow), objectValues(dayPeriods.format.abbreviated),
```
mhevery: Why did the formatting changed here?

tools/gulp-tasks/cldr/extract.js
```diff
@@ -291,8 +336,7 @@ function getDateTimeTranslations(localeData) {
   const dayPeriods = getDayPeriodsAmPm(localeData);
 
   const dayPeriodsFormat = removeDuplicates([
-    objectValues(dayPeriods.format.narrow),
-    objectValues(dayPeriods.format.abbreviated),
+    objectValues(dayPeriods.format.narrow), objectValues(dayPeriods.format.abbreviated),
```
mhevery: same here

tools/gulp-tasks/cldr/extract.js
```diff
@@ -303,22 +347,17 @@ function getDateTimeTranslations(localeData) {
   ]);
 
   const daysFormat = removeDuplicates([
-    objectValues(dayNames.format.narrow),
-    objectValues(dayNames.format.abbreviated),
-    objectValues(dayNames.format.wide),
-    objectValues(dayNames.format.short)
+    objectValues(dayNames.format.narrow), objectValues(dayNames.format.abbreviated),
```
mhevery: all of these formattings have changed..

tools/gulp-tasks/cldr/extract.js
```diff
@@ -404,11 +437,11 @@ function getFirstDayOfWeek(localeData) {
  */
 function getWeekendRange(localeData) {
   const startDay =
-    localeData.get(`supplemental/weekData/weekendStart/${localeData.attributes.territory}`) ||
-    localeData.get('supplemental/weekData/weekendStart/001');
+      localeData.get(`supplemental/weekData/weekendStart/${localeData.attributes.territory}`) ||
```
mhevery: Maybe move all of the formating changes in separate SHA? Its hard to know what is just re-format and what is actual change.

packages/common/locales/global/af-NA.js
mhevery: 👍 for just inlining `undefined` and not trying to be compressor.
👍 for going through `terser`

packages/common/locales/global/af-NA.js
```diff
@@ -0,0 +1,23 @@
...
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+// THIS CODE IS GENERATED - DO NOT MODIFY
+// See angular/tools/gulp-tasks/cldr/extract.js
+
+(function() {
```
mhevery: I am a bit confused, from your other code it seems this has gone through `clang-format` but the output does not seem to be formatted. What is going on?

packages/common/src/i18n/locale_data_api.ts
```diff
@@ -6,9 +6,8 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {ɵLocaleDataIndex as LocaleDataIndex, ɵfindLocaleData as findLocaleData, ɵgetLocalePluralCase} from '@angular/core';
+import {ɵCurrencyIndex, ɵExtraLocaleDataIndex, ɵLocaleDataIndex, ɵfindLocaleData, ɵgetLocalePluralCase} from '@angular/core';
```
mhevery: This file contains a lot of changes which just prefix `ɵ` so it looks like many things even though it is really noop. Could we change the import to say `ɵfindLocaleData as findLocaleData` that way most of these changes could go away?

packages/common/src/i18n/locale_data_api.ts
```diff
@@ -493,9 +492,9 @@ export const getLocalePluralCase: (locale: string) => ((value: number) => Plural
     ɵgetLocalePluralCase;
 
 function checkFullData(data: any) {
-  if (!data[LocaleDataIndex.ExtraData]) {
+  if (!data[ɵLocaleDataIndex.ExtraData]) {
     throw new Error(
-        `Missing extra locale data for the locale "${data[LocaleDataIndex.LocaleId]}". Use "registerLocaleData" to load new data. See the "I18n guide" on angular.io to know more.`);
+        `Missing extra locale data for the locale "${data[ɵLocaleDataIndex.LocaleId]}". Use "registerLocaleData" to load new data. See the "I18n guide" on angular.io to know more.`);
```
mhevery: Did you mean to change the error text as well???

packages/common/test/i18n/format_date_spec.ts
```diff
@@ -15,7 +13,7 @@ import localeHu from '@angular/common/locales/hu';
 import localeSr from '@angular/common/locales/sr';
 import localeTh from '@angular/common/locales/th';
 import {isDate, toDate, formatDate} from '@angular/common/src/i18n/format_date';
-import {ɵDEFAULT_LOCALE_ID as DEFAULT_LOCALE_ID} from '@angular/core';
+import {ɵDEFAULT_LOCALE_ID, ɵunregisterLocaleData, ɵregisterLocaleData} from '@angular/core';
```
mhevery: same here, import as to have less changes in file.

packages/common/test/i18n/format_number_spec.ts
```diff
@@ -10,33 +10,35 @@ import localeEn from '@angular/common/locales/en';
 import localeEsUS from '@angular/common/locales/es-US';
 import localeFr from '@angular/common/locales/fr';
 import localeAr from '@angular/common/locales/ar';
-import {formatCurrency, formatNumber, formatPercent, registerLocaleData} from '@angular/common';
+import {formatCurrency, formatNumber, formatPercent} from '@angular/common';
 import {describe, expect, it} from '@angular/core/testing/src/testing_internal';
-import {ɵDEFAULT_LOCALE_ID as DEFAULT_LOCALE_ID} from '@angular/core';
+import {ɵDEFAULT_LOCALE_ID, ɵunregisterLocaleData, ɵregisterLocaleData} from '@angular/core';
```
mhevery: import as to have fewer changes in file.

packages/common/test/i18n/locale_data_api_spec.ts
```diff
@@ -6,56 +6,24 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {ɵfindLocaleData as findLocaleData} from '@angular/core';
-import localeCaESVALENCIA from '@angular/common/locales/ca-ES-VALENCIA';
+import {ɵregisterLocaleData, ɵunregisterLocaleData} from '@angular/core';
```
mhevery: same here

packages/common/test/i18n/localization_spec.ts
```diff
@@ -10,20 +10,21 @@ import localeRo from '@angular/common/locales/ro';
 import localeSr from '@angular/common/locales/sr';
 import localeZgh from '@angular/common/locales/zgh';
 import localeFr from '@angular/common/locales/fr';
-import {LOCALE_ID} from '@angular/core';
+import {LOCALE_ID, ɵunregisterLocaleData, ɵregisterLocaleData} from '@angular/core';
```
mhevery: and here

packages/common/test/pipes/date_pipe_spec.ts
```diff
@@ -6,24 +6,21 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {DatePipe, registerLocaleData} from '@angular/common';
+import {DatePipe} from '@angular/common';
 import localeEn from '@angular/common/locales/en';
 import localeEnExtra from '@angular/common/locales/extra/en';
 import {PipeResolver} from '@angular/compiler/src/pipe_resolver';
 import {JitReflector} from '@angular/platform-browser-dynamic/src/compiler_reflector';
+import {ɵunregisterLocaleData, ɵregisterLocaleData} from '@angular/core';
```
mhevery: here

packages/common/test/pipes/number_pipe_spec.ts
```diff
@@ -11,19 +11,22 @@ import localeEsUS from '@angular/common/locales/es-US';
 import localeFr from '@angular/common/locales/fr';
 import localeAr from '@angular/common/locales/ar';
 import localeDeAt from '@angular/common/locales/de-AT';
-import {registerLocaleData, CurrencyPipe, DecimalPipe, PercentPipe, formatNumber} from '@angular/common';
+import {ɵunregisterLocaleData, ɵregisterLocaleData} from '@angular/core';
```
mhevery: here as well

packages/core/src/i18n/locale_data_api.ts
```diff
@@ -51,3 +58,39 @@ export function findLocaleData(locale: string): any {
...
+ */
+export function getLocaleData(normalizedLocale: string): any {
+  return LOCALE_DATA[normalizedLocale];
+}
+
+/**
+ * Helper function to remove all the locale data from `LOCALE_DATA`.
+ */
+export function unregisterLocaleData() {
+  Object.keys(LOCALE_DATA).forEach(key => delete LOCALE_DATA[key]);
```
mhevery: can't we just do `LOCALE_DATA = {}` to reset it?
(you may need to move this function to `locale_data.ts`)

packages/core/test/i18n/locale_data_api_spec.ts
```diff
@@ -25,9 +29,18 @@ import {findLocaleData, registerLocaleData, unregisterLocaleData} from '../../sr
...
+      global.ng.common.locale = global.ng.common.locale || {};
+      global.ng.common.locale['fr'] = fakeGlobalFr;
+      global.ng.common.locale['de'] = localeDe;
+      global.ng.common.locale['de-ch'] = localeDeCH;
     });
 
-    afterAll(() => unregisterLocaleData());
+    afterAll(() => {
+      unregisterLocaleData();
+      delete global.ng.common.locale;
```
mhevery: instead of `delete` could we just set `locale = undefined`? I know it is test, but `delete` has all kinds of bad perf implications for the VM, and I would just prefer we just don't use it anywhere so that we have good example for other code which may end up being in production.

mhevery (changes requested):
Minor comments, LGTM otherwise.

tools/gulp-tasks/cldr/extract.js
```diff
@@ -20,12 +20,15 @@ const I18N_FOLDER = `${COMMON_PACKAGE}/src/i18n`;
...
 const RELATIVE_I18N_FOLDER = path.resolve(__dirname, `../../../${I18N_FOLDER}`);
 const RELATIVE_I18N_CORE_FOLDER = path.resolve(__dirname, `../../../${I18N_CORE_FOLDER}`);
 const RELATIVE_I18N_DATA_FOLDER = path.resolve(__dirname, `../../../${I18N_DATA_FOLDER}`);
-const RELATIVE_I18N_DATA_EXTRA_FOLDER = path.resolve(__dirname, `../../../${I18N_DATA_EXTRA_FOLDER}`);
-const DEFAULT_RULE = 'function anonymous(n\n/*``*/) {\nreturn"other"\n}';
-const EMPTY_RULE = 'function anonymous(n\n/*``*/) {\n\n}';
+const RELATIVE_I18N_DATA_EXTRA_FOLDER =
+    path.resolve(__dirname, `../../../${I18N_DATA_EXTRA_FOLDER}`);
+const RELATIVE_I18N_GLOBAL_FOLDER = path.resolve(__dirname, `../../../${I18N_GLOBAL_FOLDER}`);
+const DEFAULT_RULE = 'function anonymous(n) {\nreturn"other"\n}';
```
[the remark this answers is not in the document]
petebacondarwin: This is because the function string that is generated from the CLDR data uses the name `anonymous` which we then replace.

tools/gulp-tasks/cldr/extract.js
petebacondarwin: OK

tools/gulp-tasks/cldr/extract.js
petebacondarwin: I made a conscious choice not to clang the JS files since it just makes them larger. Since they are already JS they will receive no further processing and so it seemed sensible to keep the size of these files down.

tools/gulp-tasks/cldr/extract.js
petebacondarwin: Clang being run  on save by VS Code.

packages/common/src/i18n/locale_data_api.ts
petebacondarwin: The current situation is inconsistent with some variables using the barred-o prefix and some being aliased to remove them. I chose to change all the use of such private imports with out aliasing:

* it improves consistency
* it makes it clear that these are private APIs that we are using

So this would be a one-off change as part of this commit's refactor.

packages/common/src/i18n/locale_data_api.ts
petebacondarwin: The change here is inside a back-ticked substitution so the barred-O will not be seen in the error message.

packages/core/src/i18n/locale_data_api.ts
petebacondarwin: Oh yes! I just copied the deletion over from when it was in the tests and we didn't have access to just rewrite the variable.

packages/core/test/i18n/locale_data_api_spec.ts
petebacondarwin: I used delete because it also has the effect of removing the key from the container.
But I realise that we only check `global.ng.common.locale !== undefined` anyway so happy to change :-)

tools/gulp-tasks/cldr/extract.js
```diff
@@ -72,62 +78,104 @@ module.exports = (gulp, done) => {
...
+function generateGlobalLocale(locale, localeData, baseCurrencies) {
+  const basicLocaleData = generateBasicLocaleString(locale, localeData, baseCurrencies);
+  const extraLocaleData = generateDayPeriodsSupplementalString(locale, localeData);
+  const data = basicLocaleData.replace(/\]$/, `, ${extraLocaleData}]`);
+  return `${HEADER}
+(function() {
+  const root = typeof globalThis !== 'undefined' && globalThis || typeof global !== 'undefined' && global || typeof window !== 'undefined' && window;
+  root.ng = root.ng || {};
+  root.ng.common = root.ng.common || {};
+  root.ng.common.locale = root.ng.common.locale || {};
```
IgorMinar: the namespace should be "locales" not "locale". Can you please fix?
```suggestion
  root.ng.common.locale = root.ng.common.locales || {};
```

IgorMinar: this is because the ESM entry point is `@angular/common/locales`

petebacondarwin: 👍

packages/common/locales/BUILD.bazel
```diff
@@ -12,7 +12,7 @@ ts_library(
 
 npm_package(
     name = "package",
-    srcs = ["package.json"],
+    srcs = glob(["global/*.js"]) + ["package.json"],
```
clydin: Does the package.json with the side effects flag work as intended?  I remember there were some issues with it in the localize package.

packages/common/locales/global/af-NA.js
petebacondarwin: I'll add a commit.

tools/gulp-tasks/cldr/extract.js
petebacondarwin: Since I am going to add a commit to pass these files through terser before adding to the npm package, I will actually format them with clang after generation.

tools/gulp-tasks/cldr/extract.js
petebacondarwin: I have moved the clang formatting here to a separate commit

packages/common/locales/global/af-NA.js
petebacondarwin: The JS was not being formatted because it was planned to send it straight to the distributable - but now it will be passed through terser - so we will now format the generated files that are checked in.

packages/common/src/i18n/locale_data_api.ts
petebacondarwin: In case you were referring to `registerLocaleData()` - this is still a public function from `@angular/common` and so it is correct not to rename this one.

packages/core/src/i18n/locale_data_api.ts
```diff
@@ -51,3 +59,44 @@ export function findLocaleData(locale: string): any {
...
+    LOCALE_DATA[normalizedLocale] = global.ng && global.ng.common && global.ng.common.locale &&
+        global.ng.common.locale[normalizedLocale];
+  }
+  return LOCALE_DATA[normalizedLocale];
+}
+
+/**
+ * Helper function to remove all the locale data from `LOCALE_DATA`.
+ */
+export function unregisterLocaleData() {
```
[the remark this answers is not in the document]
petebacondarwin: Done

packages/common/locales/BUILD.bazel
petebacondarwin: `"sideEffects": false` indicates to webpack that all files in the package are side-effect free.
Obviously this is not true for the `global` files (`closure-locale.js`) but since they will not be bundled via webpack it makes no difference.

clydin: Technically these files could be imported directly in application code (polyfills file, for instance) and function as intended. Assuming the bundler didn't remove them due to a perceived lack of side effects.

petebacondarwin: Could they be imported? They are not modules... Does webpack support importing plain JS files?

clydin: They can.  Importing a module for side-effects only is part of the spec.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import
https://www.ecma-international.org/ecma-262/6.0/#sec-imports

petebacondarwin: OK, I will tweak the `side-effects` attribute then. Thanks @clydin

petebacondarwin:
I had problems getting Bazel to run terser over the files before adding them to the npm package.
So I will not add a commit to do that in this PR.
Once I work out how to do that I can create another PR, which will reduce the global locale file sizes.

petebacondarwin:
@mhevery - PTAL.

tools/gulp-tasks/cldr/extract.js
petebacondarwin: Done

packages/common/src/i18n/locale_data_api.ts
petebacondarwin: I have moved the renaming of the imports to their private names to the commit before this one.

mhevery (approved):

mhevery:
[presubmit](http://test/OCL:278471457:BASE:278471847:1572908354483:b48b1498)

PR #33530, opened 2019-11-01
ci: re-enable material unit tests job

devversion:
The Material units tests job has been disabled with the typescript upgrade
PR since the components repository was still on an old TypeScript version
(due to cyclic dependencies between the framework and components repository).

Since the components repository has been updated to `9.0.0-next.15` and now
uses the compatible TypeScript version, we can re-enable the job.

Related to this change, we need to ignore the `package.json` engines when installing
the dependencies of the components repo. This is because the components repo already
updated to NodeJS v12 and Yarn v1.19.1. This is not the case for the CI setup of
framework. For now, since we don't want to change the dev setup (as it slows down development
if everyone needs to update NodeJS and Yarn), we temporarily disable the engines.

Additionally, the material unit tests job now depends on the actual release packages
(not on the ngtsc compiled ones). This is because the components repo setup relies
on NGCC being run. This is also helpful for validating ngcc against the framework packages.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,8 +29,8 @@ var_4_win: &cache_key_win_fallback v5-angular-win-node-12.0-
 
 # Cache key for the Material unit tests job. **Note** when updating the SHA in the cache keys,
 # also update the SHA for the "MATERIAL_REPO_COMMIT" environment variable.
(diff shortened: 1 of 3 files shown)


filipesilva (approved):

josephperrott (approved):
LGTM

PR #33536, opened 2019-11-01
fix(core): make QueryList implement Iterable in the type system

### LINKED ISSUE #29842: QueryList is not iterable
Docs say

> Implements an iterable interface, therefore it can be used in both ES6 javascript for (var i of items) loops as well as in Angular templates with *ngFor="let i of myList".

This is plainly untrue:

https://github.com/angular/angular/blob/master/packages/core/src/linker/query_list.ts

In TS, I get:

> TS2488: Type 'QueryList<RecordsListItem>' must have a '[Symbol.iterator]()' method that returns an iterator.

even though we use --downlevelIteration

alxhub:
Originally, QueryList implemented Iterable and provided a Symbol.iterator
on its prototype. This caused issues with tree-shaking, so QueryList was
refactored and the Symbol.iterator added in its constructor instead. As
part of this change, QueryList no longer implemented Iterable directly.

Unfortunately, this meant that QueryList was no longer assignable to
Iterable or, consequently, NgIterable. NgIterable is used for NgFor's input,
so this meant that QueryList was not usable (in a type sense) for NgFor
iteration. View Engine's template type checking would not catch this, but
Ivy's did.

As a fix, this commit adds the declaration (but not the implementation) of
the Symbol.iterator function back to QueryList. This has no runtime effect,
so it doesn't affect tree-shaking of QueryList, but it ensures that
QueryList is assignable to NgIterable and thus usable with NgFor.

Fixes #29842

### DIFF
--- a/packages/core/src/linker/query_list.ts
+++ b/packages/core/src/linker/query_list.ts
@@ -42,7 +42,7 @@ function symbolIterator<T>(this: QueryList<T>): Iterator<T> {
  *
  * @publicApi
  */
-export class QueryList<T>/* implements Iterable<T> */ {
+export class QueryList<T> implements Iterable<T> {
   public readonly dirty = true;
   private _results: Array<T> = [];
   public readonly changes: Observable<any> = new EventEmitter();
@@ -142,4 +142,11 @@ export class QueryList<T>/* implements Iterable<T> */ {
     (this.changes as EventEmitter<any>).complete();
     (this.changes as EventEmitter<any>).unsubscribe();
   }
+
+  // The implementation of `Symbol.iterator` should be declared here, but this would cause
+  // tree-shaking issues with `QueryList. So instead, it's added in the constructor (see comments
+  // there) and this declaration is left here to ensure that TypeScript considers QueryList to
+  // implement the Iterable interface. This is required for template type-checking of NgFor loops
+  // over QueryLists to work correctly, since QueryList must be assignable to NgIterable.
+  [Symbol.iterator] !: () => Iterator<T>;
 }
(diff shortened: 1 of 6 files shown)


mhevery (approved):

AndrewKushnir:
@alxhub should this change be included in patch branch as well? Thank you.

packages/core/src/linker/query_list.ts
```diff
@@ -142,4 +142,11 @@ export class QueryList<T>/* implements Iterable<T> */ {
     (this.changes as EventEmitter<any>).complete();
     (this.changes as EventEmitter<any>).unsubscribe();
   }
+
+  // The implementation of `Symbol.iterator` should be declared here, but this would cause
+  // tree-shaking issues with `QueryList. So instead, it's added in the constructor (see comments
+  // there) and this declaration is left here to ensure that TypeScript considers QueryList to
+  // implement the Iterable interface. This is required for template type-checking of NgFor loops
+  // over QueryLists to work correctly, since QueryList must be assignable to NgIterable.
+  [Symbol.iterator] !: () => Iterator<T>;
```
Splaktar: Very interesting! For anyone else not completely familiar with using TypeScript Symbols to define class members, here's the docs: https://www.typescriptlang.org/docs/handbook/symbols.html

Splaktar (approved):
LGTM!

This would allow me to remove quite a number of `someQueryList.toArray()` calls in templates with Ivy.

alxhub:
[Presubmit](http://test/OCL:281339653:BASE:281339685:1574191829131:36cb98a8)

PR #33537, opened 2019-11-01
fix(ivy): don't infer template context types when in full mode

### LINKED ISSUE #33527: Ivy: fullTemplateTypeCheck errors with NgIterable
# 🐞 bug report

### Affected Package

### Is this a regression?

This is introduced by Ivy.

### Description

A component with a template looking like:

```html
<div *ngFor="let user of users | slice: 0:2 as total">
  {{ total.length }}: {{ user.name }}
</div>
```

`ng build --prod` with `fullTemplateTypeCheck: true` throws, as it thinks `length` is not a property of `total`.

## 🔬 Minimal Reproduction

https://github.com/cexbrayat/ng-iterable

## 🔥 Exception or Error
<pre><code>
ERROR in src/app/app.component.html:2:6 - error TS2339: Property 'length' does not exist on type 'NgIterable<any>'.
  Property 'length' does not exist on type 'Iterable<any>'.

2   {{ total.length }}: {{ user.name }}
       ~~~~~~~~~~~~~

  src/app/app.component.ts:5:16
                     ~~~~~~~~~~~~~~~~~~~~~~
</code></pre>

## 🌍  Your Environment

**Angular Version:**
<pre><code>
Angular CLI: 9.0.0-rc.0
Node: 12.12.0
OS: darwin x64
Angular: 9.0.0-rc.0
... router

</code></pre>

**Anything else relevant?**

This is using ng and CLI `9.0.0-rc.0` and `fTTC`, but *not* using `strictTemplates`.

alxhub:
fix(ivy): don't infer template context types when in full mode

The Ivy template type-checker is capable of inferring the type of a
structural directive (such as NgForOf<T>). Previously, this was done with
fullTemplateTypeCheck: true, even if strictTemplates was false. View Engine
previously did not do this inference, and so this causes breakages if the
type of the template context is not what the user expected.

In particular, consider the template:

```html
<div *ngFor="let user of users as all">
  {{user.index}} out of {{all.length}}
</div>
```

As long as `users` is an array, this seems reasonable, because it appears
that `all` is an alias for the `users` array. However, this is misleading.

In reality, `NgForOf` is rendered with a template context that contains
both a `$implicit` value (for the loop variable `user`) as well as a
`ngForOf` value, which is the actual value assigned to `all`. The type of
`NgForOf`'s template context is `NgForContext<T>`, which declares `ngForOf`'s
type to be `NgIterable<T>`, which does not have a `length` property (due to
its incorporation of the `Iterable` type).

This commit stops the template type-checker from inferring template context
types unless strictTemplates is set (and strictInputTypes is not disabled).

Fixes #33527.

### DIFF
--- a/packages/compiler-cli/src/ngtsc/program.ts
+++ b/packages/compiler-cli/src/ngtsc/program.ts
@@ -426,7 +426,7 @@ export class NgtscProgram implements api.Program {
     if (this.options.fullTemplateTypeCheck) {
       const strictTemplates = !!this.options.strictTemplates;
       typeCheckingConfig = {
-        applyTemplateContextGuards: true,
+        applyTemplateContextGuards: strictTemplates,
         checkQueries: false,
         checkTemplateBodies: true,
         checkTypeOfInputBindings: strictTemplates,
@@ -471,6 +471,7 @@ export class NgtscProgram implements api.Program {
     // based on "fullTemplateTypeCheck".
     if (this.options.strictInputTypes !== undefined) {
       typeCheckingConfig.checkTypeOfInputBindings = this.options.strictInputTypes;
+      typeCheckingConfig.applyTemplateContextGuards = this.options.strictInputTypes;
     }
     if (this.options.strictNullInputTypes !== undefined) {
       typeCheckingConfig.strictNullInputBindings = this.options.strictNullInputTypes;
--- a/packages/compiler-cli/test/ngtsc/template_typecheck_spec.ts
+++ b/packages/compiler-cli/test/ngtsc/template_typecheck_spec.ts
@@ -43,6 +43,16 @@ export declare class IndexPipe {
   static ɵpipe: i0.ɵPipeDefWithMeta<IndexPipe, 'index'>;
 }
 
+export declare class SlicePipe {
+  transform<T>(value: ReadonlyArray<T>, start: number, end?: number): Array<T>;
+  transform(value: string, start: number, end?: number): string;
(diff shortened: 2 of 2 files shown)


packages/compiler-cli/src/ngtsc/typecheck/src/type_check_file.ts
```diff
@@ -71,6 +71,7 @@ export class TypeCheckFile extends Environment {
     // kind of global namespace in TS, which forces a full re-typecheck of the user's program that
     // is somehow more expensive than the initial parse.
     source += '\nexport const IS_A_MODULE = true;\n';
+    console.error(source);
```
JoostK: Oops

packages/compiler-cli/src/transformers/compiler_host.ts
```diff
@@ -367,6 +367,7 @@ export class TsCompilerAotCompilerTypeCheckHostAdapter implements ts.CompilerHos
     const {sourceText, context} = this.emitter.emitStatementsAndContext(
         genFile.genFileUrl, genFile.stmts, /* preamble */ '',
         /* emitSourceMaps */ false);
+    console.error(sourceText);
```
JoostK: Oops 2.

alxhub:
[Presubmit](http://test/OCL:281370373:BASE:281370425:1574200029334:c12c77ae)
No Ivy presubmit needed; only affects template type-checking in fTTC mode.

PR #33540, opened 2019-11-01
refactor(ivy): move all styling configurations into `TNodeFlags`

matsko:
This patch gets rid of the configuration settings present in the
`TStylingContext` array that is used within the styling algorithm
for `[style]`, `[style.prop]`, `[class]` and `[class.name]` bindings.
These configurations now all live inside of the `TNodeFlags`.

### DIFF
--- a/integration/_payload-limits.json
+++ b/integration/_payload-limits.json
@@ -39,7 +39,7 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 2289,
-        "main-es2015": 265613,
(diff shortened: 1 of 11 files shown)


matsko:
Presubmit = http://test/OCL:278484480:BASE:278484519:1572912056813:3247f2f2

Ivy Presubmit = https://test.corp.google.com/ui#id=OCL:278484480:BASE:278484656:1572912099064:320b5624

mhevery (approved):

PR #33542, opened 2019-11-02
Fw 1559

mhevery:
## PR Checklist

## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

https://angular-team.atlassian.net/browse/FW-1559

- [ ] Yes
- [X] No

### DIFF
--- a/packages/core/src/render3/node_manipulation.ts
+++ b/packages/core/src/render3/node_manipulation.ts
@@ -628,7 +628,7 @@ function getNativeAnchorNode(parentTNode: TNode, lView: LView): RNode|null {
     const lContainer = getLContainer(parentTNode as TViewNode, lView);
     if (lContainer === null) return null;
     const index = lContainer.indexOf(lView, CONTAINER_HEADER_OFFSET) - CONTAINER_HEADER_OFFSET;
-    return getBeforeNodeForView(index, lContainer);
+    return getFirstRNodeFromLViewInLContainer(lContainer, index);
   } else if (
       parentTNode.type === TNodeType.ElementContainer ||
       parentTNode.type === TNodeType.IcuContainer) {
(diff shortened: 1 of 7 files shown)


packages/core/src/render3/node_manipulation.ts
```diff
@@ -663,27 +663,54 @@ export function appendChild(childEl: RNode | RNode[], childTNode: TNode, current
...
+    } else {
+      // If we have `ElementContainer` or `IcuContainer` than we need to descend into them and
+      // find the actual first `Element`
+      while (tNodeType === TNodeType.ElementContainer || tNodeType === TNodeType.IcuContainer) {
+        tNode = tNode.child !;
+        ngDevMode && assertDefined(tNode, 'ElementContainer and IcuContainer must have children');
+        tNodeType = tNode.type;
       }
-      return getNativeByTNodeOrNull(tViewNodeChild, lView);
+      return getNativeByTNode(tNode, lView);
```
pkozlowski-opensource: I _think_ that it should be recursive in nature - what might happen is that a child of ElementContainer / IcuContainer is a container which might have views etc.

packages/core/src/render3/node_manipulation.ts
```diff
@@ -663,27 +663,54 @@ export function appendChild(childEl: RNode | RNode[], childTNode: TNode, current
...
+    if (tNode !== null) {
+    let tNodeType = tNode.type;
+    if (tNodeType === TNodeType.Container) {
+      // In this case the first `TNode` is another `LContainer` this means we now need to
+      // recurse
+      // into it the `LView` and retrieve its `RNode`.
+      const childLContainer = lView[tNode.index] as LContainer;
+      ngDevMode && assertLContainer(childLContainer);
+      return getFirstRNodeFromLViewInLContainer(childLContainer, 0);
+    } else {
```
pkozlowski-opensource: I don't see the projection handled here, but it might show up at the root of a view, right?

pkozlowski-opensource (changes requested):
@mhevery I did the first, quick pass on this PR and I think that there are more cases to cover here. Essentially the function you are modifying is _almost_ the same as collecting root nodes (https://github.com/angular/angular/blob/ac9d044cad4fee47ed6bfd44f905b546b41952cc/packages/core/src/render3/view_ref.ts#L305-L348) - the way I understand it is that _only_ difference is that we can stop at the _first_ result.

More generally, I think that we can fold all those root node traversal routines we've got into one - we've got already `applyView` - we would have to "just" add a new tree walk action (collect).

I think that, in this PR, we should:
- add more tests / asserts for all the node types (need to cover the same cases as in `collectNativeNodes` linked above);
- add / modify implementation that covers all the cases.

Then, with tests in place and in a separate PR, refactor code to have the root nodes tree walking in one place only (should drop quite a bit of code duplication!). I've spent quite a bit of time poking around the code so happy to discuss this in more details!

mhevery:
@pkozlowski-opensource I have refactored it to use existing code in 9465a3862d, but as you can see the aesthetics of it are not great. :-( My current thinking is that we should refactor `applyView` and friends into a cursor pattern. We could than just iterate over the cursor as many times as we need to to either add/remove or to retrieve the first item from the list. What do you think?

packages/core/test/acceptance/view_insertion_spec.ts
```diff
@@ -249,4 +277,150 @@ describe('view insertion', () => {
       expect(fixture.debugElement.queryAll(By.css('div.dynamic')).length).toBe(4);
     });
   });
+
+  describe('regression', () => {
+    it('should allow inserting of all different types', () => {
```
pkozlowski-opensource: Could we split this test into several ones (one test for each case, similarly to what I did in https://github.com/angular/angular/pull/33493? )? It is super-hard to follow-as is.

packages/core/test/acceptance/view_insertion_spec.ts
```diff
@@ -249,4 +277,150 @@ describe('view insertion', () => {
       expect(fixture.debugElement.queryAll(By.css('div.dynamic')).length).toBe(4);
     });
   });
+
+  describe('regression', () => {
```
pkozlowski-opensource: nits regarding naming:
* regression -> non-regression? but more generally, I would have sth more descriptive, like "should allow inserting views before views with different node types";
* "should allow inserting of all different types" - the essence of those tests is that we want to insert _before_ a view with specific nodes layout and it should be reflected in the name.

Generally speaking, I would love to have this test split into multiple ones (see below)

mhevery:
@pkozlowski-opensource took over this one here https://github.com/angular/angular/pull/33627

PR #33544, opened 2019-11-02
style: Add VSCode recommended launch and task configurations

mhevery:
This adds a common configurations used when developing code in VSCode.
Specifically it adds support for launching these targets as tasks and
ind debugger.

- `packages/core/test`
- `packages/core/test/render3`
- `packages/core/test/acceptance`

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [X] Code style update (formatting, local variables)
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [X] No

### DIFF
--- a/.vscode/recommended-launch.json
+++ b/.vscode/recommended-launch.json
@@ -1,0 +1,85 @@
+{
+  // Use IntelliSense to learn about possible attributes.
+  // Hover to view descriptions of existing attributes.
+  // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "name": "Attach (inspect)",
+      "type": "node",
+      "request": "attach",
+      "port": 9229,
+      "address": "localhost",
+      "restart": false,
+      "sourceMaps": true,
+      "localRoot": "${workspaceRoot}",
+      "remoteRoot": "${workspaceRoot}",
+      "stopOnEntry": false,
+      "timeout": 60000,
+    },
+    {
+      "name": "Attach (no-sm,inspect)",
+      "type": "node",
+      "request": "attach",
+      "port": 9229,
+      "address": "localhost",
+      "restart": false,
+      "sourceMaps": false,
+      "localRoot": "${workspaceRoot}",
+      "stopOnEntry": false,
+      "remoteRoot": "${workspaceRoot}",
+      "timeout": 60000,
+    },
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "IVY:packages/core/test/acceptance",
+      "program": "${workspaceFolder}/node_modules/.bin/bazel",
+      "args": [
+        "test",
+        "--define=compile=aot",
+        "packages/core/test/acceptance",
+        "--config=debug"
+      ],
+      "port": 9229,
+      "address": "localhost",
+      "restart": true,
+      "timeout": 60000,
+      "sourceMaps": true,
+    },
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "IVY:packages/core/test/render3",
+      "program": "${workspaceFolder}/node_modules/.bin/bazel",
+      "args": [
+        "test",
+        "--define=compile=aot",
+        "packages/core/test/render3",
+        "--config=debug"
+      ],
+      "port": 9229,
+      "address": "localhost",
+      "restart": true,
+      "timeout": 60000,
+      "sourceMaps": true,
+    },
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "IVY:packages/core/test",
+      "program": "${workspaceFolder}/node_modules/.bin/bazel",
+      "args": [
+        "test",
+        "--define=compile=aot",
+        "packages/core/test",
+        "--config=debug"
+      ],
+      "port": 9229,
+      "address": "localhost",
+      "restart": true,
+      "timeout": 60000,
+      "sourceMaps": true,
+    },
+  ]
+}
--- a/.vscode/recommended-launch.json
+++ b/.vscode/recommended-launch.json
@@ -5,7 +5,7 @@
   "version": "0.2.0",
   "configurations": [
     {
-      "name": "Attach (inspect)",
+      "name": "Attach to bazel test ... --config=debug",
       "type": "node",
       "request": "attach",
       "port": 9229,
@@ -18,7 +18,7 @@
       "timeout": 60000,
     },
     {
-      "name": "Attach (no-sm,inspect)",
+      "name": "Attach to bazel test ... --config=debug (no source maps)",
       "type": "node",
       "request": "attach",
       "port": 9229,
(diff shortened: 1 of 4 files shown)


.vscode/recommended-launch.json
```diff
@@ -0,0 +1,85 @@
...
+      "name": "Attach (no-sm,inspect)",
+      "type": "node",
+      "request": "attach",
+      "port": 9229,
+      "address": "localhost",
+      "restart": false,
+      "sourceMaps": false,
+      "localRoot": "${workspaceRoot}",
+      "stopOnEntry": false,
+      "remoteRoot": "${workspaceRoot}",
```
gkalpak: Inconsistent order of properties wrt to the above launch config 😠

.vscode/recommended-launch.json
```diff
@@ -0,0 +1,85 @@
...
+      "address": "localhost",
+      "restart": false,
+      "sourceMaps": true,
+      "localRoot": "${workspaceRoot}",
+      "remoteRoot": "${workspaceRoot}",
+      "stopOnEntry": false,
+      "timeout": 60000,
+    },
+    {
+      "name": "Attach (no-sm,inspect)",
```
gkalpak: `no-sm` is a little cryptic (I would never have guessed what it is without looking at the code below).

.vscode/recommended-launch.json
```diff
@@ -0,0 +1,85 @@
+{
+  // Use IntelliSense to learn about possible attributes.
+  // Hover to view descriptions of existing attributes.
+  // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "name": "Attach (inspect)",
```
gkalpak: Could we have a little more descriptive names ("Attach" is too generic).
Alternatively, consider adding a comment in this file explaining what each config is about/how to use it.

.vscode/recommended-launch.json
```diff
@@ -0,0 +1,85 @@
...
+      "sourceMaps": false,
+      "localRoot": "${workspaceRoot}",
+      "stopOnEntry": false,
+      "remoteRoot": "${workspaceRoot}",
+      "timeout": 60000,
+    },
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "IVY:packages/core/test/acceptance",
```
gkalpak: Please keep the order of properties consistent in across launch configs. It makes it easier to compare them and decide which one is more suitable.

.vscode/recommended-tasks.json
```diff
@@ -0,0 +1,113 @@
...
+    {
+      "label": "IVY:packages/core/test/...",
+      "type": "shell",
+      "command": "${workspaceFolder}/node_modules/.bin/bazel",
+      "args": [
+        "test",
+        "--define=compile=aot",
+        "packages/core/test",
+        "packages/core/test/acceptance",
+        "packages/core/test/render3",
```
gkalpak: Why not `"packages/core/test/..."` instead of individual entries?

.vscode/recommended-tasks.json
```diff
@@ -0,0 +1,113 @@
...
+        "packages/core/test/render3"
+      ],
+      "group": "test",
+      "presentation": {
+        "reveal": "always",
+        "panel": "dedicated"
+      },
+    },
+  ]
+}
```
gkalpak: I feel like the tasks added here are only recommended for a limited group of people working on specific parts of the framework. Depending on which part you are working on (e.g. `aio/`, `ngcc`, other non-core packages) different tasks would make sense.

How about we give this file a more descriptive name (and adding a comment at the top) to indicate this is recommended if you are working on X (e.g. `recommended-tasks-ivy.json` or `recommended-tasks-core.json`)? In the future, we might decide to have more dedicated recommendation (e.g. `recommended-tasks-docs.json` for people working on the docs).

gkalpak (changes requested):
I like the direction 👍 
Left a couple of suggestions.

BTW, it would be a good idea to mention these files in [.vscode/README.md][1] as well (similar to `recommended-settings.json`).

[1]: https://github.com/angular/angular/blob/65fb30513/.vscode/README.md

.vscode/recommended-launch.json
mhevery: fixed

.vscode/recommended-launch.json
mhevery: fixed

.vscode/recommended-launch.json
mhevery: fixed

.vscode/recommended-launch.json
mhevery: fixed

.vscode/recommended-tasks.json
mhevery: `...` will include a lot more tests than just those three.

.vscode/recommended-tasks.json
mhevery: 1) Yes this is not an exclusive list
2) Yes this is heavily biased towards ivy work. 
3) I am concerned that if we have too many recommendations we will create islands and people will have to chose. For this reason I think we should just stick with a single life which over time will grow to more of these. If it becomes too much over time we can re-discuss breaking it up, but as of right now I think it is pre-mature.

mhevery (commented):
Fixup 3b8eb45292 PTAL

.vscode/recommended-tasks.json
gkalpak: The `packages/core/test/...` in the label is a little misleading, but I can live with that :grin:

.vscode/recommended-launch.json
gkalpak: Some properties were still inconsistently ordered. I pushed a fixup commit.

.vscode/recommended-tasks.json
gkalpak: I am still concerned that many community contributors (which are mostly (a) focused on docs and (b) not familiar with our build process and framework internals) might be confused by this (i.e. trying to figure out whether/how these tasks can be useful to them, when they aren't).

All, I am suggesting is to add a comment above the launch configurations/tasks mentioning they are geared towards ivy work. (If we later decide to add more configurations/tasks, we can just comment over the new blocks mentioning what those are for.)

But definitely not a hill I'm willing to die on :grin:
Leaving it up to you 👉

gkalpak (approved):
I pushed a fixup commit with some minor fixes. Feel free to leave it or revert it.
Otherwise LGTM :+1:

gkalpak:
(Nit: The change feels more like a `docs` than a `style` change (although neither is ideal).)

mhevery:
> I pushed a fixup commit with some minor fixes. Feel free to leave it or revert it.
> Otherwise LGTM 👍

thank you!

PR #33555, opened 2019-11-03
fix(ivy): match directives on namespaced elements

### LINKED ISSUE #32061: Issue with directive selector inside SVG when using IVY
# 🐞 bug report

### Affected Package
IVY Compiler

### Is this a regression?

Yes, works fine without IVY

### Description

A directive or component with a selector of type 'element[attribute]' and Inputs will not compile under IVY if used inside an SVG template.

An attribute selector works as expected.

## 🔬 Minimal Reproduction

https://github.com/mitch8/ivy-issue

The issue is only present when building with IVY.
<pre><code>
"angularCompilerOptions": {
    "enableIvy": true,
  }
</code></pre>
**ng build** works fine
**ng build --prod** fail with IVY enabled

## 🔥 Exception or Error

If the directive is part of my app I get an error like this.
<pre><code>
ERROR in src/app/app.component.html(4,9): Property 'test' does not exist on type 'HTMLElement'.

</code></pre>

Interestingly, if the directive is part of a library from node_modules, it will compile without errors under IVY, however you will then get a runtime error. 
<pre><code>
Can't bind to 'test' since it isn't a known property of 'text'
</code></pre>

## 🌍  Your Environment

**Angular Version:**
Angular CLI: 8.2.1
Node: 10.13.0
OS: win32 x64
Angular: 8.2.1
... router

@angular-devkit/architect         0.802.1
@angular-devkit/build-angular     0.802.1
@angular-devkit/build-optimizer   0.802.1
@angular-devkit/build-webpack     0.802.1
@angular-devkit/core              8.2.1
@angular-devkit/schematics        8.2.1
@ngtools/webpack                  8.2.1
@schematics/angular               8.2.1
@schematics/update                0.802.1
webpack                           4.38.0

</code></pre>

JoostK:
Prior to this change, namespaced elements such as SVG elements would not
participate correctly in directive matching as their namespace was not
ignored, which was the case with the View Engine compiler. This led to
incorrect behavior at runtime and template type checking.

This commit resolved the issue by ignoring the namespace of elements and
attributes like they were in View Engine.

Fixes #32061

### DIFF
--- a/packages/core/test/acceptance/directive_spec.ts
+++ b/packages/core/test/acceptance/directive_spec.ts
@@ -260,6 +260,28 @@ describe('directives', () => {
       expect(calls).toEqual(['MyDir.ngOnInit']);
     });
 
+    it('should match namespaced elements', () => {
+      const calls: string[] = [];
+
+      @Directive({selector: 'text[dir]'})
+      class MyDir {
+        ngOnInit() { calls.push('MyDir.ngOnInit'); }
+      }
+
+      @Component({
+        selector: `my-comp`,
+        template: `<svg><text dir></text></svg>`,
+      })
+      class MyComp {
+      }
+
+      TestBed.configureTestingModule({declarations: [MyDir, MyComp]});
+      const fixture = TestBed.createComponent(MyComp);
+      fixture.detectChanges();
+
+      expect(calls).toEqual(['MyDir.ngOnInit']);
+    });
+
     it('should match directives when the node has "class", "style" and a binding', () => {
       const logs: string[] = [];
 
--- a/packages/core/test/acceptance/directive_spec.ts
+++ b/packages/core/test/acceptance/directive_spec.ts
@@ -7,7 +7,7 @@
  */
 
 import {CommonModule} from '@angular/common';
-import {Component, Directive, EventEmitter, Output, TemplateRef, ViewChild, ViewContainerRef} from '@angular/core';
+import {Component, Directive, ElementRef, EventEmitter, Output, TemplateRef, ViewChild, ViewContainerRef} from '@angular/core';
 import {Input} from '@angular/core/src/metadata';
 import {TestBed} from '@angular/core/testing';
 import {By} from '@angular/platform-browser';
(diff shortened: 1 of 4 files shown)


packages/core/test/acceptance/directive_spec.ts
```diff
@@ -260,6 +260,28 @@ describe('directives', () => {
       expect(calls).toEqual(['MyDir.ngOnInit']);
     });
 
+    it('should match namespaced elements', () => {
```
pkozlowski-opensource: "should match directives on elements with namespace"?  We should say what we are trying to match

packages/core/test/acceptance/directive_spec.ts
```diff
@@ -260,6 +260,28 @@ describe('directives', () => {
...
+      const calls: string[] = [];
+
+      @Directive({selector: 'text[dir]'})
```
pkozlowski-opensource: could you please add another test where we try to match a directive on the `<svg>` directly? A case like `<svg dir>` where directive's selector is `svg[dir]`? This case is currently broken on ng-bootstrap and we had to work-around it, see https://github.com/ng-bootstrap/ng-bootstrap/commit/0353fc90600710c648afecf33272c7347cc4746e

packages/core/test/acceptance/directive_spec.ts
JoostK: Thanks, that's what I meant it to say ;)

packages/core/test/acceptance/directive_spec.ts
JoostK: Excellent idea!

pkozlowski-opensource (approved):
LGTM for runtime tests

petebacondarwin (approved):

alxhub (approved):

alxhub:
[Presubmit](http://test/OCL:278906490:BASE:278906607:1573069219955:c6dffba3)
[Ivy Presubmit](http://test/OCL:278906490:BASE:278906834:1573069279929:60686601)

PR #33560, opened 2019-11-03
fix(ivy): do not invoke change detection for destroyed views

AndrewKushnir:
Prior to this commit, calling change detection for destroyed views resulted in errors being thrown in some cases. This commit adds a check to make sure change detection is invoked for non-destroyed views only.

This PR resolves FW-1636.

## PR Type

- [x] Bugfix

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/render3/instructions/shared.ts
+++ b/packages/core/src/render3/instructions/shared.ts
@@ -37,7 +37,7 @@ import {isAnimationProp} from '../util/attrs_utils';
 import {INTERPOLATION_DELIMITER, renderStringify, stringifyForError} from '../util/misc_utils';
 import {getInitialStylingValue} from '../util/styling_utils';
 import {getLViewParent} from '../util/view_traversal_utils';
-import {getComponentLViewByIndex, getNativeByIndex, getNativeByTNode, getTNode, isCreationMode, readPatchedLView, resetPreOrderHookFlags, unwrapRNode, viewAttachedToChangeDetector} from '../util/view_utils';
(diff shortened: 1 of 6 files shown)


AndrewKushnir:
Initial g3 presubmits:
- [VE Presubmit](http://test/OCL:278267693:BASE:278267706:1572828060349:1f02f075)
- [Ivy Presubmit](http://test/OCL:278267693:BASE:278267752:1572828102517:4dccf6df)

packages/core/src/render3/instructions/shared.ts
```diff
@@ -1642,6 +1642,7 @@ export function tickRootContext(rootContext: RootContext) {
 }
 
 export function detectChangesInternal<T>(view: LView, context: T) {
+  if (isViewDestroyed(view)) return;
```
kara: In View Engine, if you try to call `detectChanges()` on a destroyed embedded view, it throws. e.g.

```ts
const viewRef = this.vcr.createEmbeddedView(this.template);
viewRef.destroy();
viewRef.detectChanges();   // should throw
```

We should continue to do the same, but this fix would cause the destroyed view to be ignored rather than throwing the expected error.

Can you tell me more about the G3 case? I'm not sure I understand yet what is triggering the change detection run that causes the `object unsubscribed` error. You mentioned that zoneJS is being triggered, but that runs change detection from the top down, so `detectChangesInternal` would be called with the root view. Is that the one being destroyed in G3?

AndrewKushnir:
Thanks for review @kara!

Closing this PR for now, since more investigation needs to happen on g3 side.

PR #33567, opened 2019-11-04
fix(core): undecorated-classes-with-di migration should report config errors

devversion:
Currently TypeScript projects with an invalid tsconfig configuration,
cause the undecorated-classes-with-di migration to throw. Instead we
should gracefully exit the migration (like we do for syntactical
diagnostics), but report that there are configuration issues.

This issue surfaced when testing this migration in combination
with the Angular CLI migrations. One of the CLI migrations currently
causes invalid tsconfig files which then cause this issue in the
undecorated-classes-with-di migration. https://github.com/angular/angular-cli/issues/16035


kara (approved):
LGTM

kara:
[presubmit](http://test/OCL:278670294:BASE:278670344:1572981785438:7014dd9c)

PR #33571, opened 2019-11-04
fix(core): renderer-to-renderer2 migration not migrating methods

devversion:
The `renderer-to-renderer2` migration currently does not work
properly in v9 because the migration relies on the type checker
for detecting references to `Renderer` from `@angular/core`.

This is contradictory since the `Renderer` is no longer
exported in v9 `@angular/core`. In order to make sure that
the migration still works in v9, and that we can rely on the
type checker for the best possible detection, we take advantage
of module augmentation and in-memory add the `Renderer` export
to the `@angular/core` module.

---

Before coming up with the augmentation solution, I spent quite some time re-working the
detection to be not based on the `ts.TypeChecker`. Though, I went with this solution as it gives us
the best detection of the `Renderer` and doesn't involve reworking the whole detection.

### DIFF
--- a/packages/core/schematics/migrations/renderer-to-renderer2/index.ts
+++ b/packages/core/schematics/migrations/renderer-to-renderer2/index.ts
@@ -18,18 +18,17 @@ import {HelperFunction, getHelper} from './helpers';
 import {migrateExpression, replaceImport} from './migration';
 import {findCoreImport, findRendererReferences} from './util';
 
-
+const MODULE_AUGMENTATION_FILENAME = 'ɵɵRENDERER_MIGRATION_CORE_AUGMENTATION.d.ts';
 
 /**
  * Migration that switches from `Renderer` to `Renderer2`. More information on how it works:
  * https://hackmd.angular.io/UTzUZTnPRA-cSa_4mHyfYw
  */
 export default function(): Rule {
-  return (tree: Tree, context: SchematicContext) => {
+  return (tree: Tree) => {
     const {buildPaths, testPaths} = getProjectTsConfigPaths(tree);
     const basePath = process.cwd();
     const allPaths = [...buildPaths, ...testPaths];
-    const logger = context.logger;
 
     if (!allPaths.length) {
       throw new SchematicsException(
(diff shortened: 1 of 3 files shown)


packages/core/schematics/migrations/renderer-to-renderer2/index.ts
```diff
@@ -44,8 +43,23 @@ export default function(): Rule {
...
-  const program = ts.createProgram(parsed.fileNames, parsed.options, host);
+  const host = createMigrationCompilerHost(tree, parsed.options, basePath, fileName => {
+    // In case the module augmentation file has been requested, we return a source file that
+    // augments "@angular/core" to include a named export called "Renderer". This ensures that
+    // we can rely on the type checker for this migration in v9 where "Renderer" has been removed.
+    if (fileName === MODULE_AUGMENTATION_FILENAME) {
+      return `
+        import '@angular/core';
+        declare module "@angular/core" {
+          class Renderer {}
```
crisbeto: We aren't compiling the code so maybe it doesn't matter, but wouldn't this technically cause compilation errors for something like `someInstance.setElementClass` since the `Renderer` class is empty? Could we do `declare const Renderer: any` instead?

devversion: Yeah it would cause errors, but since we don't collect diagnostics, the error won't be reported. It shouldn't have an effect on the migration either.

 I think switching to `any` will break since the type checker will then return `any` if we call `getTypeAtLocation`.

crisbeto (approved):
LGTM

kara (approved):
LGTM

kara:
[presubmit](http://test/OCL:278669830:BASE:278669896:1572981669913:9613126f)

packages/core/schematics/utils/typescript/compiler_host.ts
```diff
@@ -10,15 +10,18 @@ import {relative} from 'path';
...
 
   // We need to overwrite the host "readFile" method, as we want the TypeScript
   // program to be based on the file contents in the virtual file tree. Otherwise
   // if we run multiple migrations we might have intersecting changes and
   // source files.
   host.readFile = fileName => {
-    const buffer = tree.read(relative(basePath, fileName));
+    const treeRelativePath = relative(basePath, fileName);
+    const fakeOutput = fakeRead ? fakeRead(treeRelativePath) : null;
+    const buffer = fakeOutput !== null ? fakeOutput : tree.read(treeRelativePath);
```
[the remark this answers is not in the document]
devversion: Updated.

devversion:
Caretaker: Failure for `integration_test` seems unrelated to this PR. I rebased on top of 44623a116158c8048ce61a873ed63a290c039202 (latest commit in master).

atscott:
Presubmit failures are also unrelated. Still need to investigate the integration_test failures

PR #33583, opened 2019-11-04
fix(ivy): more descriptive errors for nested i18n sections

AndrewKushnir:
This commit moves nested i18n section detection to an earlier stage where we convert HTML AST to Ivy AST. This also gives a chance to produce better diagnostic message for nested i18n sections, that also includes a file name and location.

## PR Type

- [x] Bugfix

- [ ] Yes
- [x] No

### DIFF
--- a/packages/compiler/src/render3/r3_template_transform.ts
+++ b/packages/compiler/src/render3/r3_template_transform.ts
@@ -90,7 +90,7 @@ class HtmlAstToIvyAst implements html.Visitor {
     if (isI18nRootElement) {
       if (this.inI18nBlock) {
         this.reportError(
-            'Could not mark an element as translatable inside of a translatable section',
+            'Cannot mark an element as translatable inside of a translatable section. Please remove the nested i18n marker.',
             element.sourceSpan);
       }
       this.inI18nBlock = true;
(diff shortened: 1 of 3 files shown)


packages/compiler/src/render3/r3_template_transform.ts
```diff
@@ -80,11 +80,21 @@ class HtmlAstToIvyAst implements html.Visitor {
...
 
   constructor(private bindingParser: BindingParser) {}
 
   // HTML visitor
   visitElement(element: html.Element): t.Node|null {
+    const isI18nRootElement = isI18nRootNode(element.i18n);
+    if (isI18nRootElement) {
+      if (this.inI18nBlock) {
+        this.reportError(
+            'Could not mark an element as translatable inside of a translatable section',
```
kara: I think the message might be even clearer if we explicitly say that the "i18n" marker is the problem and add a call to action.

```suggestion
            'Cannot mark an element as translatable inside of a translatable section. Please remove the nested i18n marker.`
```

kara (changes requested):
Thanks for doing this! Some thoughts on the error message

AndrewKushnir:
- [VE Presubmit](http://test/OCL:278485837:BASE:278485914:1572912378406:5a4a6859)
- [Ivy Presubmit](http://test/OCL:278485837:BASE:278486031:1572912414020:58f9388e)

packages/compiler/src/render3/r3_template_transform.ts
AndrewKushnir: Thanks for the review @kara, the message text is now updated.

kara (approved):
LGTM

PR #33584, opened 2019-11-04
fix(common): update CLDR generated files

josephperrott:


.circleci/config.yml
```diff
@@ -843,7 +843,7 @@ workflows:
             - setup
       - test_aio_preview:
           requires:
-            - aio_preview
+           - aio_preview
```
gkalpak: 😱

tools/gulp-tasks/cldr/extract.js
```diff
@@ -519,7 +519,7 @@ function getPluralFunction(locale) {
 
   fn = fn
     .replace(
-      toRegExp('function anonymous(n\n/*``*/) {\n'),
+      toRegExp('function anonymous\(n\n\) {\n'),
```
gkalpak: Escaping `(` and `)` is handled by `isRegExp()`, so it shouldn't be necessary.

clydin:
Can you add the CLDR version into the commit message?  The large volume of changes makes it hard to find what version is being used.

petebacondarwin (approved):
@josephperrott - can the first commit message be

```
build: update CLDR version to v33
```

Also can you explain what you did in the final fixup commit that changes `\u202f` strings, and why?

josephperrott:
Updated the commit messages to include the CLDR version.

@petebacondarwin I don't know what I did that ended up removing them, but that was the wrong thing since those strings should be there.

petebacondarwin (approved):

doom777:
forgot to remove the old cldr-data directory from .gitignore.
I'll remove it in my PR

PR #33599, opened 2019-11-05
fix(ivy): better support for i18n attributes on <ng-container>s

AndrewKushnir:
Prior to this commit, i18n runtime logic used `elementAttributeInternal` function (that uses `setAttribute` function under the hood) for all elements where i18n attributes are present. However the `<ng-container>` elements in a template may also have i18n attributes and calling `setAttribute` fails, since they are represented as comment nodes in DOM. This commit ensures that we call `setAttribute` on nodes with TNodeType.Element type (that support that operation) only.

This PR resolves FW-1663.

## PR Type

- [x] Bugfix

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/test/acceptance/i18n_spec.ts
+++ b/packages/core/test/acceptance/i18n_spec.ts
@@ -1268,6 +1268,33 @@ onlyInIvy('Ivy i18n logic').describe('runtime i18n', () => {
       expect(comp.attributes['messagetext']).toBe('Bonjour');
       expect(comp.attributes['ng-reflect-message-text']).toBe('Bonjour');
     });
+
+    it('with <ng-container> elements', () => {
+      loadTranslations({[computeMsgId('Hello', 'meaning')]: 'Bonjour'});
+
+      @Directive({selector: '[mydir]'})
+      class Dir {
+        @Input() mydir: string = '';
+      }
+
+      @Component({
+        selector: 'my-cmp',
+        template: `
+          <ng-container i18n-mydir="meaning|description" mydir="Hello"></ng-container>
+        `,
+      })
+      class Cmp {
+      }
+
+      TestBed.configureTestingModule({
+        declarations: [Cmp, Dir],
+      });
+      const fixture = TestBed.createComponent(Cmp);
+      fixture.detectChanges();
+
+      const dir = fixture.debugElement.childNodes[0].injector.get(Dir);
+      expect(dir.mydir).toEqual('Bonjour');
+    });
   });
 
   it('should work with directives and host bindings', () => {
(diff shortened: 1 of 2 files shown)


packages/core/test/acceptance/i18n_spec.ts
```diff
@@ -1265,9 +1265,37 @@ onlyInIvy('Ivy i18n logic').describe('runtime i18n', () => {
       fixture.detectChanges();
 
       const comp = fixture.debugElement.query(By.css('welcome'));
-      expect(comp.attributes['messagetext']).toBe('Bonjour');
+      expect(fixture.nativeElement.textContent).toBe('Bonjour');
+      expect(comp.componentInstance.messageText).toBe('Bonjour');
```
ocombe: you're no longer testing if the attribute has the correct value?

ocombe (changes requested):
I'm pretty sure that `setAttribute` and `setProperty` are not the same, some properties have different names compared to the original attributes. If you try to set the property `class` for example, it'll not work as expected, since the property for the attribute `class` is `className` instead...
Instead you should add a check to see if the element is a container, and use `elementPropertyInternal` on it, otherwise you should still use `elementAttributeInternal`

packages/core/test/acceptance/i18n_spec.ts
AndrewKushnir: I reverted it, since I switched back to `elementAttributeInternal`.

AndrewKushnir:
@ocombe thanks for the review, I've updated the logic to avoid calling `elementAttributeInternal` for non-elements. For non-elements (like `<ng-container>`) we just set the necessary inputs. Could you please have a look again? Thank you.

ocombe (approved):

AndrewKushnir:
- [VE Presubmit](http://test/OCL:278726107:BASE:278726165:1572997941017:7c914c22)
- [Ivy Presubmit](http://test/OCL:278726107:BASE:278726340:1572997995185:66eadf26)

packages/core/test/acceptance/i18n_spec.ts
```diff
@@ -1268,6 +1268,33 @@ onlyInIvy('Ivy i18n logic').describe('runtime i18n', () => {
       expect(comp.attributes['messagetext']).toBe('Bonjour');
       expect(comp.attributes['ng-reflect-message-text']).toBe('Bonjour');
     });
+
+    it('with <ng-container> elements', () => {
```
petebacondarwin: NIT - I appreciate that this is outside the remit of this PR but I find that inverting the `describe` and  `it` clauses - so that the "should" expression is on the `describe`) makes it much more difficult to understand what this is testing, especially when the `describe` is so much further back up the file.

petebacondarwin (approved):

packages/core/test/acceptance/i18n_spec.ts
AndrewKushnir: @petebacondarwin thanks for the review! I've updated the description of the test to be more clear.

mhevery (approved):

PR #33600, opened 2019-11-05
test(ivy): introduce a benchmark for duplicate style/class bindings

matsko:
Prior to this patch all the styling benchmarks only tested for
template-based style/class bindings. Because of each of the bindings
being only present in the template, there was no possibility of
there being any duplicate bindings. This benchmark introduces
style/class bindings being evaluated from both a template and from
various directives.

This benchmark can be executed by calling:

```
bazel build //packages/core/test/render3/perf:duplicate_style_and_class_bindings_lib.min_debug.es2015.js

node dist/bin/packages/core/test/render3/perf/duplicate_style_and_class_bindings_lib.min_debug.es2015.js
```

The benchmark is also run via the `profile_all.js` script (found in
`packages/core/test/render3/perf/`)

### DIFF
--- a/packages/core/test/render3/perf/shared.ts
+++ b/packages/core/test/render3/perf/shared.ts
@@ -1,0 +1,36 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import {DirectiveDef, HostBindingsFunction} from '../../../src/render3/interfaces/definition';
+
+export function createDirectiveDef(selector: string, hostBindings: HostBindingsFunction<any>) {
+  const def: DirectiveDef<any> = {
+    inputs: {},
+    declaredInputs: {},
+    outputs: {},
+    type: ((() => {}) as any),
+    factory: () => ({}),
+    providersResolver: null,
+    exportAs: null,
+    contentQueries: null,
+    template: null,
+    viewQuery: null,
+    hostBindings: hostBindings,
+    selectors: [['', selector, '']],
+    onChanges: null,
+    onInit: null,
+    doCheck: null,
+    afterContentInit: null,
+    afterContentChecked: null,
+    afterViewInit: null,
+    afterViewChecked: null,
+    onDestroy: null,
+    features: null,
+    setInput: null,
+  } as DirectiveDef<any>;
+  return def;
+}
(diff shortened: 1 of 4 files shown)


matsko:
```
┌────────────────────────────────────┬─────────┬──────┬───────────┬───────────┬───┐
│              (index)               │  time   │ unit │ base_time │ base_unit │ % │
├────────────────────────────────────┼─────────┼──────┼───────────┼───────────┼───┤
│           class_binding            │ 167.192 │ 'ns' │           │           │   │
│       directive_instantiate        │  1.744  │ 'us' │           │           │   │
│ duplicate_style_and_class_bindings │  7.854  │ 'ms' │           │           │   │
│        element_text_create         │ 987.821 │ 'ns' │           │           │   │
│           interpolation            │ 151.613 │ 'us' │           │           │   │
│             listeners              │  1.474  │ 'us' │           │           │   │
│ map_based_style_and_class_bindings │ 14.366  │ 'ms' │           │           │   │
│            ng_template             │ 486.683 │ 'ns' │           │           │   │
│       noop_change_detection        │ 19.508  │ 'us' │           │           │   │
│          property_binding          │ 167.807 │ 'us' │           │           │   │
│      property_binding_update       │ 333.315 │ 'us' │           │           │   │
│      style_and_class_bindings      │ 914.12  │ 'us' │           │           │   │
│           style_binding            │ 351.994 │ 'us' │           │           │   │
└────────────────────────────────────┴─────────┴──────┴───────────┴───────────┴───┘
```

packages/core/test/render3/perf/shared.ts
```diff
@@ -0,0 +1,36 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import {DirectiveDef, HostBindingsFunction} from '../../../src/render3/interfaces/definition';
+
+export function createDirectiveDef(selector: string, hostBindings: HostBindingsFunction<any>) {
```
mhevery: why copy this function? Why not just use the one which we already have?

mhevery (changes requested):

mhevery (approved):

pkozlowski-opensource (approved):
LGTM generally speaking. I'm just slightly worried about the fact that we start to have multiple ways of defining test directives, ex. compare your approach with the already existing one: https://github.com/angular/angular/blob/master/packages/core/test/render3/perf/directive_instantiate/index.ts

Could you think of mutualising those directive creation utils (not now, but something to keep in mind).

PR #33605, opened 2019-11-05
refactor(ivy): use hex flags instead of binary for TNodeFlags

matsko:


kara (approved):
LGTM

matsko:
presubmit = http://test/OCL:278721008:BASE:278721055:1572996422145:6ced19f5

ivy presubmit = http://test/OCL:278721008:BASE:278721108:1572996440447:1ac8f151

mhevery (approved):
Why is this better than binary?

matsko:
@mhevery the next PR adds in 10+ more flags. So managing the binary notation is trick and cumbersome. It's hard to track where each bit is.

Given that most flags are independent bits, it makes more sense to just use increment of hex values since there are only 4 to memorize: `0x1`, `0x2`, `0x4` and `0x8`. If and when flags are combined then we can do so using the enum labels (e.g. `hasTemplateAndHostBindings = hasTemplateBindings | hasHostBindings`).

PR #33606, opened 2019-11-05
fix(ivy): Handle overrides for {providedIn: AModule} in R3TestBed

atscott:
This issue was found when debugging a test failure that was using lazy
loaded modules with the router. When doing this, the router calls
`NgModuleFactory.create` for the loaded module. This module gets a new
injector so the overrides provided in TestBed are not applied unless the
Injectable is in the providers list (which is not the case for
{providedIn...} Injectables).

### DIFF
--- a/packages/core/test/test_bed_spec.ts
+++ b/packages/core/test/test_bed_spec.ts
@@ -6,7 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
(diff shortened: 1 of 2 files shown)


AndrewKushnir (approved):
LGTM 👍 Thanks for fixing this @atscott.

atscott:
[Ivy presubmit](http://test/OCL:278715099:BASE:278715361:1572994776489:514e9b23)
Should also run global presubmit

atscott:
[global presubmit](https://test.corp.google.com/ui#id=OCL:279011817:BASE:279011954:1573107463046:ee7a9911)

only regressions are e2e flakes that wouldn't be using TestBed

PR #33608, opened 2019-11-05
test(ivy): introduce a benchmark for duplicate map-based style/class bindings

matsko:
Prior to this patch all the styling benchmarks only tested for
template map-based style/class bindings. Because of each of the bindings
being only present in the template, there was no possibility of
there being any duplicate map-based styling bindings.

This benchmark introduces benchmarking for map-based style/class bindings
that are evaluated from both template bindings as well as directives.

This benchmark can be executed by calling:

```
bazel build //packages/core/test/render3/perf:duplicate_map_based_style_and_class_bindings_lib.min_debug.es2015.js

node dist/bin/packages/core/test/render3/perf/duplicate_map_based_style_and_class_bindings_lib.min_debug.es2015.js
```

The benchmark is also run via the `profile_all.js` script (found in
`packages/core/test/render3/perf/`)

### DIFF
--- a/packages/core/test/render3/perf/BUILD.bazel
+++ b/packages/core/test/render3/perf/BUILD.bazel
@@ -189,3 +189,16 @@ ng_benchmark(
     name = "duplicate_style_and_class_bindings",
     bundle = ":duplicate_style_and_class_bindings_lib",
 )
+
+ng_rollup_bundle(
+    name = "duplicate_map_based_style_and_class_bindings_lib",
+    entry_point = ":duplicate_map_based_style_and_class_bindings/index.ts",
+    deps = [
+        ":perf_lib",
+    ],
+)
+
+ng_benchmark(
+    name = "duplicate_map_based_style_and_class_bindings",
+    bundle = ":duplicate_map_based_style_and_class_bindings_lib",
+)
--- a/packages/core/test/render3/perf/duplicate_map_based_style_and_class_bindings/index.ts
+++ b/packages/core/test/render3/perf/duplicate_map_based_style_and_class_bindings/index.ts
@@ -1,0 +1,175 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import {ɵɵadvance} from '../../../../src/render3/instructions/advance';
+import {ɵɵelement, ɵɵelementEnd, ɵɵelementStart} from '../../../../src/render3/instructions/element';
+import {refreshView} from '../../../../src/render3/instructions/shared';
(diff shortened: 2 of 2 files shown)


packages/core/test/render3/perf/shared.ts
```diff
@@ -0,0 +1,36 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import {DirectiveDef, HostBindingsFunction} from '../../../src/render3/interfaces/definition';
+
+export function createDirectiveDef(selector: string, hostBindings: HostBindingsFunction<any>) {
```
mhevery: i don't see why we need to have this function. Can we just call the one in the `render3` folder?

mhevery (changes requested):

matsko:
```
┌──────────────────────────────────────────────┬─────────┬──────┬───────────┬───────────┬───┐
│                   (index)                    │  time   │ unit │ base_time │ base_unit │ % │
├──────────────────────────────────────────────┼─────────┼──────┼───────────┼───────────┼───┤
│                class_binding                 │ 170.362 │ 'ns' │           │           │   │
│            directive_instantiate             │  1.742  │ 'us' │           │           │   │
│ duplicate_map_based_style_and_class_bindings │ 14.666  │ 'ms' │           │           │   │
│      duplicate_style_and_class_bindings      │  7.884  │ 'ms' │           │           │   │
│             element_text_create              │ 989.532 │ 'ns' │           │           │   │
│                interpolation                 │  157.3  │ 'us' │           │           │   │
│                  listeners                   │  1.48   │ 'us' │           │           │   │
│      map_based_style_and_class_bindings      │ 14.396  │ 'ms' │           │           │   │
│                 ng_template                  │ 481.983 │ 'ns' │           │           │   │
│            noop_change_detection             │ 19.515  │ 'us' │           │           │   │
│               property_binding               │ 170.588 │ 'us' │           │           │   │
│           property_binding_update            │ 324.69  │ 'us' │           │           │   │
│           style_and_class_bindings           │ 903.191 │ 'us' │           │           │   │
│                style_binding                 │ 383.216 │ 'us' │           │           │   │
└──────────────────────────────────────────────┴─────────┴──────┴───────────┴───────────┴───┘
```

mhevery (approved):

pkozlowski-opensource (approved):
LGTM generally speaking. I'm just slightly worried about the fact that we start to have multiple ways of defining test directives, ex. compare your approach with the already existing one: https://github.com/angular/angular/blob/master/packages/core/test/render3/perf/directive_instantiate/index.ts

Could you think of mutualising those directive creation utils (not now, but something to keep in mind).

PR #33609, opened 2019-11-06
fix(ivy): wrap functions from "providers" in parentheses in Closure mode

AndrewKushnir:
Due to the fact that Tsickle runs between analyze and transform phases in Angular, Tsickle may transform nodes (add comments with type annotations for Closure) that we captured during the analyze phase. As a result, some patterns where a function is returned from another function may trigger automatic semicolon insertion, which breaks the code (makes functions return `undefined` instead of a function). In order to avoid the problem, this commit updates the code to wrap all functions in some expression ("privders" and "viewProviders") in parentheses. More info can be found in Tsickle source code here: https://github.com/angular/tsickle/blob/d7974262571c8a17d684e5ba07680e1b1993afdd/src/jsdoc_transformer.ts#L1021

## PR Type

- [x] Bugfix

- [ ] Yes
- [x] No

### DIFF
--- a/packages/compiler-cli/ngcc/src/analysis/decoration_analyzer.ts
+++ b/packages/compiler-cli/ngcc/src/analysis/decoration_analyzer.ts
@@ -89,17 +89,18 @@ export class DecorationAnalyzer {
         this.scopeRegistry, this.scopeRegistry, this.isCore, this.resourceManager, this.rootDirs,
         /* defaultPreserveWhitespaces */ false,
         /* i18nUseExternalIds */ true, /* i18nLegacyMessageIdFormat */ '', this.moduleResolver,
-        this.cycleAnalyzer, this.refEmitter, NOOP_DEFAULT_IMPORT_RECORDER),
+        this.cycleAnalyzer, this.refEmitter, NOOP_DEFAULT_IMPORT_RECORDER,
+        /* annotateForClosureCompiler */ false),
     new DirectiveDecoratorHandler(
         this.reflectionHost, this.evaluator, this.fullRegistry, NOOP_DEFAULT_IMPORT_RECORDER,
-        this.isCore),
+        this.isCore, /* annotateForClosureCompiler */ false),
     new InjectableDecoratorHandler(
         this.reflectionHost, NOOP_DEFAULT_IMPORT_RECORDER, this.isCore,
         /* strictCtorDeps */ false),
     new NgModuleDecoratorHandler(
         this.reflectionHost, this.evaluator, this.fullMetaReader, this.fullRegistry,
         this.scopeRegistry, this.referencesRegistry, this.isCore, /* routeAnalyzer */ null,
-        this.refEmitter, NOOP_DEFAULT_IMPORT_RECORDER),
+        this.refEmitter, NOOP_DEFAULT_IMPORT_RECORDER, /* annotateForClosureCompiler */ false),
     new PipeDecoratorHandler(
         this.reflectionHost, this.evaluator, this.metaRegistry, NOOP_DEFAULT_IMPORT_RECORDER,
         this.isCore),
--- a/packages/compiler-cli/src/ngtsc/annotations/src/util.ts
+++ b/packages/compiler-cli/src/ngtsc/annotations/src/util.ts
@@ -354,7 +354,7 @@ export function readBaseClass(
 
 const parensWrapperTransformerFactory: ts.TransformerFactory<ts.Expression> =
     (context: ts.TransformationContext) => {
-      const visitor = (node: ts.Node): ts.Node => {
+      const visitor: ts.Visitor = (node: ts.Node): ts.Node => {
         const visited = ts.visitEachChild(node, visitor, context);
         if (ts.isArrowFunction(visited) || ts.isFunctionExpression(visited)) {
           return ts.createParen(visited);
--- a/packages/compiler-cli/src/ngtsc/annotations/src/util.ts
+++ b/packages/compiler-cli/src/ngtsc/annotations/src/util.ts
@@ -351,3 +351,28 @@ export function readBaseClass(
 
   return null;
 }
+
+const parensWrapperTransformerFactory: ts.TransformerFactory<ts.Expression> =
+    (context: ts.TransformationContext) => {
+      const visitor = (node: ts.Node): ts.Node => {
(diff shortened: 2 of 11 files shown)


packages/compiler-cli/test/ngtsc/ngtsc_spec.ts
```diff
@@ -357,6 +357,147 @@ runInEachFileSystem(os => {
...
       });
+
+      /**
+       * The following set of tests verify that after Tsickle run we do not have cases which trigger
+       * automatic semicolon insertion, which breaks the code. In order to avoid the problem, we
+       * wrap all function expressions in certain fields ("providers" and "viewProviders") in
+       * parentheses. More info on Tsickle processing related to this case can be found here:
+       * https://github.com/angular/tsickle/blob/d7974262571c8a17d684e5ba07680e1b1993afdd/src/jsdoc_transformer.ts#L1021
+       */
+      describe(
```
petebacondarwin: I would move this `describe` block to include the previous `it` clause and simply rename it:

```
describe('when closure annotations are requested', () => { ...
```

Then you can drop that phase from each of the `it` clauses.

packages/compiler-cli/test/ngtsc/ngtsc_spec.ts
```diff
@@ -357,6 +357,147 @@ runInEachFileSystem(os => {
...
+            `;
+
+            const service = `
+              export class Service {
+                id: string = 'service-id';
+              }
+            `;
+
+            const verifyOutput = (jsContents: string) => {
+              // verify that there is no pattern that triggers automatic semicolon insertion
```
petebacondarwin: Perhaps add a bit more explanation?

```suggestion
              // verify that there is no pattern that triggers automatic semicolon insertion
              // by checking that there are no return statements not wrapped in parentheses.
```

packages/compiler-cli/src/ngtsc/annotations/src/util.ts
```diff
@@ -351,3 +351,28 @@ export function readBaseClass(
 
   return null;
 }
+
+const parensWrapperTransformerFactory: ts.TransformerFactory<ts.Expression> =
+    (context: ts.TransformationContext) => {
```
petebacondarwin: NIT: the type is not needed on this parameter because it is implied by the type of `parensWrapperTransformerFactory`.

```suggestion
    context => {
```

packages/compiler-cli/src/ngtsc/annotations/src/util.ts
```diff
@@ -351,3 +351,28 @@ export function readBaseClass(
...
+      const visitor = (node: ts.Node): ts.Node => {
```
petebacondarwin:
```suggestion
      const visitor: ts.Visitor = node => {
```

packages/compiler-cli/src/ngtsc/annotations/src/util.ts
```diff
@@ -351,3 +351,28 @@ export function readBaseClass(
...
+        const visited = ts.visitEachChild(node, visitor, context);
+        if (ts.isArrowFunction(visited) || ts.isFunctionExpression(visited)) {
+          return ts.createParen(visited);
+        }
+        return visited;
+      };
+      return (node: ts.Expression) => ts.visitEachChild(node, visitor, context);
```
petebacondarwin:
```suggestion
      return node => ts.visitEachChild(node, visitor, context);
```

packages/compiler-cli/ngcc/src/analysis/decoration_analyzer.ts
```diff
@@ -89,17 +89,18 @@ export class DecorationAnalyzer {
         this.scopeRegistry, this.scopeRegistry, this.isCore, this.resourceManager, this.rootDirs,
         /* defaultPreserveWhitespaces */ false,
         /* i18nUseExternalIds */ true, /* i18nLegacyMessageIdFormat */ '', this.moduleResolver,
-        this.cycleAnalyzer, this.refEmitter, NOOP_DEFAULT_IMPORT_RECORDER),
+        this.cycleAnalyzer, this.refEmitter, NOOP_DEFAULT_IMPORT_RECORDER,
+        /* annotateForClosureCompiler */ false),
```
petebacondarwin: Perhaps make this argument optional, here and on the other annotations, which would save having to pass `false` in quite a number of tests

petebacondarwin (approved):
LTGM

* typo in the commit message: `privders` => `providers`
* a handful of nits - nothing blocking

packages/compiler-cli/ngcc/src/analysis/decoration_analyzer.ts
alxhub: No optional arguments in the compiler as a whole. We can make wrapper functions for test that pass default values, but so many issues in the previous compiler were caused by having too many ways to configure things / too many assumptions about what the "default"s were all over the place.

packages/compiler-cli/src/ngtsc/annotations/src/util.ts
alxhub: I prefer explicit return types regardless.

AndrewKushnir:
@petebacondarwin @alxhub thanks for the review! I've updated this PR based on the comments:

- kept more explicit types and parameters as Alex proposed
- updates description and restructured tests as Pete suggested

Could you please have a quick look again and let me know if you have additional feedback?

Thank you.

AndrewKushnir:
- [VE Presubmit](http://test/OCL:280313053:BASE:280313091:1573693210707:926b9cb5)
- [Ivy Presubmit](http://test/OCL:280313053:BASE:280313232:1573693263618:b0fa24df)

packages/compiler-cli/src/ngtsc/annotations/src/util.ts
petebacondarwin: @alxhub did you mean argument type?

petebacondarwin (approved):

AndrewKushnir:
- [VE Presubmit](http://test/OCL:281192219:BASE:281192269:1574127423043:f55f030c)
- [Ivy Presubmit](http://test/OCL:281192219:BASE:281192379:1574127466847:d32f5e06)

alxhub (approved):

PR #33621, opened 2019-11-06
fix(compiler-cli): resource inlining should remove moduleId property

devversion:
When resource inlining has been enabled through the `enableResourceInlining`
compiler option, properties for `moduleId` are left in the component
decorator. This is not ideal because the `moduleId` is no longer needed,
but the dependency on the `moduleId` expression remains. e.g.

```ts
        MatPseudoCheckbox.decorators = [
            { type: i0.Component, args: [{
                        moduleId: module.id,
			template: '...' ,
```

This means that the component will not work if `module.id` is not
present in the environment where the component is loaded. One could
argue that the component author should simply remove `moduleId` from
the component, but the component could be used in JIT in development
mode where the `moduleId` property is needed.

This issue came up in `@angular/bazel` because when building libraries
with Bazel, resources should be inlined and the `moduleId` removed. This
is not the case right now, so when used in combination with
`ng_package`, the UMD bundles do not work in the browser due to
`module.id` being undefined.

**Note**: also fixes that a few test assertions were always passing regardless of the
decorator output.

### DIFF
--- a/packages/compiler-cli/src/transformers/api.ts
+++ b/packages/compiler-cli/src/transformers/api.ts
@@ -290,7 +290,7 @@ export interface CompilerOptions extends ts.CompilerOptions {
   /**
    * Whether to replace the `templateUrl` and `styleUrls` property in all
    * @Component decorators with inlined contents in `template` and `styles`
-   * properties.
+   * properties. Also removes the `moduleId` property if defined.
    * When enabled, the .js output of ngc will have no lazy-loaded `templateUrl`
    * or `styleUrl`s. Note that this requires that resources be available to
    * load statically at compile-time.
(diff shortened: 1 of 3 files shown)


IgorMinar (commented):
Just a drive-by-comment as I randomly came across this PR: `moduleId` in google3 has a different meaning than externally (where it's primarily needed for SystemJS). Before we can merge this in, someone with access to the internal module loader implementation will have to review if we can remove moduleId internally as well, or if stripping `moduleId`s can be performed only externally.

devversion:
Closing this as compilation with ngtsc does not respect that option anyway, and partial compilation inlines the template always. Unfortunately `moduleId` would still be preserved for the JIT `setClassMetadata` invocations, but I guess `moduleId` should just be avoided in general 😄

PR #33627, opened 2019-11-06
Before node for view

pkozlowski-opensource:
This is the fix for FW-1559 split into 2 commits:
* tests for (most) cases I could think of (I think that there are more, but couldn't easily reproduce in tests, need more time on those);
* the actual fix - the fix makes the tests pass but results in some code / logic duplication.

This PR is split into 2 commits so we can decide to only take tests or test + fix.

I believe that we should merge it as-is and do the refactoring in the follow-up PR(s).

### DIFF
--- a/packages/core/src/render3/node_manipulation.ts
+++ b/packages/core/src/render3/node_manipulation.ts
@@ -20,8 +20,8 @@ import {ProceduralRenderer3, RElement, RNode, RText, Renderer3, isProceduralRend
 import {isLContainer, isLView, isRootView} from './interfaces/type_checks';
 import {CHILD_HEAD, CLEANUP, DECLARATION_LCONTAINER, FLAGS, HOST, HookData, LView, LViewFlags, NEXT, PARENT, QUERIES, RENDERER, TVIEW, T_HOST, unusedValueExportToPlacateAjd as unused5} from './interfaces/view';
 import {assertNodeOfPossibleTypes, assertNodeType} from './node_assert';
-import {findComponentView} from './util/view_traversal_utils';
-import {getNativeByTNode, getNativeByTNodeOrNull, unwrapRNode} from './util/view_utils';
+import {findComponentView, getLViewParent} from './util/view_traversal_utils';
+import {getNativeByTNode, unwrapRNode} from './util/view_utils';
 
 const unusedValueToPlacateAjd = unused1 + unused2 + unused3 + unused4 + unused5;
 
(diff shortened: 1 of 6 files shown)


packages/core/src/render3/node_manipulation.ts
```diff
@@ -652,25 +652,55 @@ export function appendChild(childEl: RNode | RNode[], childTNode: TNode, current
...
+  if (tNode !== null) {
+    ngDevMode && assertNodeOfPossibleTypes(
+                     tNode, TNodeType.Element, TNodeType.Container, TNodeType.ElementContainer,
+                     TNodeType.IcuContainer, TNodeType.Projection);
+
+    const tNodeType = tNode.type;
+
+    if (tNodeType === TNodeType.Element) {
+      return getNativeByTNode(tNode, lView);
+    } else if (tNodeType === TNodeType.Container) {
```
mhevery: There is one more use case. You can have an LContainer but the TNode type is a element. This is because any element can be upgraded into a container. I think we need a test case for that.

pkozlowski-opensource: Sure, I can add a test but since views inserted on a VCRef asked on an element are inserted _after_ the element, it shouldn't matter for finding the first node of a view.

In fact I wanted to have a deeper look at the case of a directive asking for a `ViwContainerRef` on a `<ng-container>`

pkozlowski-opensource: Done. Turned out to be super-easy to add those tests so just did it: https://github.com/angular/angular/pull/33627/files#diff-6148157ee1f259b9c8461f23356f0b85R350-R367

AndrewKushnir:
Started g3 presubmits to see which targets might be fixed by this PR:
- [VE Presubmit](http://test/OCL:278900202:BASE:278900245:1573067518574:4b6e8160)
- [Ivy Presubmit](http://test/OCL:278900202:BASE:278900437:1573067575020:bbd379f4)

Will run a global TAP tonight as well.

// cc @atscott

packages/core/src/render3/node_manipulation.ts
mhevery: I see, Thank you.

mhevery (approved):

AndrewKushnir:
FYI, VE and Ivy Presubmits (including global one) are successful 👍 

@pkozlowski-opensource could you plz rebase this PR and resolve a conflict, so this PR is ready to merge? Also quick question: should it be merged to master and patch (it's marked for master only atm)?

Thank you.

pkozlowski-opensource:
@AndrewKushnir rebased

> Also quick question: should it be merged to master and patch (it's marked for master only atm)?

We should get it for both 9.x and 10.x so I guess master & patch?

atscott:
@AndrewKushnir is investigating some weird things we're seeing in presubmits. This should still be good to merge - he'll remove the blocked label when the investigation is finished.

AndrewKushnir:
Just confirmed that the numbers are fine. The "blocked" label is removed. Thank you.

PR #33634, opened 2019-11-06
build: Move CLDR dependency to npm

josephperrott:
Moves CLDR dependency to npm rather than manually downloading the data in our script.

This caused a change in the formatting on the scripts output, but the actual content is the same.


petebacondarwin (approved):
Did you also update the `closure` locale file?

josephperrott:
@petebacondarwin The `closure-locale.ts` file did not result in any changes.

gkalpak (approved):

PR #33636, opened 2019-11-06
build: update .nvmrc to correct node version

josephperrott:


atscott (approved):

.nvmrc
```diff
@@ -1 +1 @@
-10.9.0
+10.13.0
```
gkalpak: Why is this version the correct one? 😕

josephperrott: This was the version the error message said we should be using :smile: 

The error message was however informed by the updated requirements in [`@angular-devkit/architect@0.900.0-rc.0`](https://unpkg.com/browse/@angular-devkit/architect@0.900.0-rc.0/package.json).  When we attempted to release we were using node `10.9.0`, it failed.

PR #33653, opened 2019-11-07
perf: report cold and minimal times when profiling in a browser

pkozlowski-opensource:
This PR extends the reporting from the existing benchmarks, when run in a browser. 
Before: only the average time spent in script + GC was reported;
After: cold (first run), average and fastest times are reported.


kara (approved):
LGTM

kara:
merge-assistance: global

PR #33551, opened 2019-11-02
fix(ivy): recompile component when template changes in ngc watch mode

JoostK:
When the Angular compiler is operated through the ngc binary in watch
mode, changing a template in an external file would not cause the
component to be recompiled if Ivy is enabled.

There was a problem with how a cached compiler host was present that was
unaware of the changed resources, therefore failing to trigger a
recompilation of a component whenever its template changes. This commit
fixes the issue by ensuring that information about modified resources is
correctly available to the cached compiler host.

Fixes #32869

PR #33661, opened 2019-11-07
Revert "fix(ivy): recompile component when template changes in ngc wa…

atscott:
…tch mode (#33551)"

This reverts commit 8912b11f56b9d1ca2a67317ca05005e5b8a6dae3.
CI failure https://app.circleci.com/jobs/github/angular/angular/519284. Reran two additional times and it still failed so it doesn't appear to be a flake


alxhub (approved):

JoostK:
I am guessing this has something to do with path formats and how the cache is invalidated, there's probably a mismatch. I think I know how to resolve that, will try to enable Windows CI on the PR to be able to verify.

PR #33663, opened 2019-11-07
fix(ivy): auto register NgModules with ID

atscott:
For projects that use lazy loaded modules, we've observed that tree shaking in tests sometimes results in the ngfactory file for the lazy loaded module being removed. This results in the `ng_module_factory_registration` map not registering the module's id so when code later attempts to retrieve the factory, it fails. This is because ngfactory files have a side effect of creating an `NgModuleFactory` (which calls `registerNgModuleType` in the constructor). This PR keeps track of modules at the point of creation, providing a backup for retrieving it if the ngfactory was somehow tree-shaken away.

This commit also reverts the changes to TestBed that dealt with restoring the registered modules in `ng_module_factory_registration.ts`. These reverts are necessary because this change would make it so that even if a module is not registered in the map there, it is still retrieved from the map from `defineNgModule`. The original fixes in `TestBed` were not necessary: test setup in google3 has since been modified to not need the state restore because it now recompiles the lazy loaded modules with `TestBed` before each test, ensuring new provider overrides take effect.

We should run global presubmit with this PR to ensure it doesn't break anything.

### DIFF
--- a/packages/core/src/render3/definition.ts
+++ b/packages/core/src/render3/definition.ts
@@ -22,6 +22,7 @@ import {TConstants} from './interfaces/node';
 // while SelectorFlags is unused here, it's required so that types don't get resolved lazily
 // see: https://github.com/Microsoft/web-build-tools/issues/1050
 import {CssSelectorList, SelectorFlags} from './interfaces/projection';
+import {NgModuleType} from './ng_module_ref';
 
 let _renderCompCount = 0;
 
@@ -338,6 +339,8 @@ export function extractPipeDef(type: Type<any>): PipeDef<any> {
   return def !;
 }
 
+export const autoRegisterModuleById: {[id: string]: NgModuleType} = {};
+
 /**
  * @codeGenApi
  */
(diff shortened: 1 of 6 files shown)


packages/core/src/render3/definition.ts
```diff
@@ -376,6 +380,9 @@ export function ɵɵdefineNgModule<T>(def: {
     schemas: def.schemas || null,
     id: def.id || null,
   };
+  if (def.id != null) {
+    autoRegisterModuleById[def.id] = def.type as unknown as NgModuleType;
```
alxhub: I believe this will break tree-shaking, because Closure/Terser will treat this as a side effect and not be able to remove unused `NgModule`s as a result.

atscott:
[ivy presubmit](http://test/OCL:279202116:BASE:279202356:1573175999676:babd8c09)

atscott:
Compare these presubmits (after revert to TestBed) to the ones before the revert.
[ivy presubmit](http://test/OCL:279343374:BASE:279343576:1573238082554:82ae0494)
[presubmit for project with custom lazy loading](https://test.corp.google.com/OCL:279343374:BASE:279345634:1573238611045:3ac37352)

mhevery (approved):

atscott:
[global](https://test.corp.google.com/OCL:279343374:BASE:279506602:1573321093498:f109704a)

AndrewKushnir (approved):

alxhub (approved):

PR #33664, opened 2019-11-07
Missing-injectable migration should not migrate object literal providers without useClass

devversion:
* See individual commits

### DIFF
--- a/packages/compiler-cli/ngcc/src/migrations/missing_injectable_migration.ts
+++ b/packages/compiler-cli/ngcc/src/migrations/missing_injectable_migration.ts
@@ -118,19 +118,11 @@ function migrateProviders(metadata: ResolvedValueMap, field: string, host: Migra
  */
 function migrateProvider(provider: ResolvedValue, host: MigrationHost): void {
   if (provider instanceof Map) {
-    if (!provider.has('provide') || provider.has('useValue') || provider.has('useFactory') ||
-        provider.has('useExisting')) {
-      return;
-    }
-    if (provider.has('useClass')) {
-      // {provide: ..., useClass: SomeClass, deps: [...]} does not require a decorator on SomeClass,
-      // as the provider itself configures 'deps'. Only if 'deps' is missing will this require a
-      // factory to exist on SomeClass.
-      if (!provider.has('deps')) {
-        migrateProviderClass(provider.get('useClass') !, host);
-      }
-    } else {
-      migrateProviderClass(provider.get('provide') !, host);
+    // {provide: ..., useClass: SomeClass, deps: [...]} does not require a decorator on SomeClass,
+    // as the provider itself configures 'deps'. Only if 'deps' is missing will this require a
+    // factory to exist on SomeClass.
+    if (provider.has('provide') && provider.has('useClass') && !provider.has('deps')) {
+      migrateProviderClass(provider.get('useClass') !, host);
     }
   } else if (Array.isArray(provider)) {
     for (const v of provider) {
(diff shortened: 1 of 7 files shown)


packages/compiler-cli/ngcc/src/migrations/missing_injectable_migration.ts
```diff
@@ -118,19 +118,11 @@ function migrateProviders(metadata: ResolvedValueMap, field: string, host: Migra
...
-      // factory to exist on SomeClass.
-      if (!provider.has('deps')) {
-        migrateProviderClass(provider.get('useClass') !, host);
-      }
-    } else {
-      migrateProviderClass(provider.get('provide') !, host);
+    // {provide: ..., useClass: SomeClass, deps: [...]} does not require a decorator on SomeClass,
+    // as the provider itself configures 'deps'. Only if 'deps' is missing will this require a
+    // factory to exist on SomeClass.
+    if (provider.has('provide') && provider.has('useClass') && !provider.has('deps')) {
```
devversion: @JoostK I just noticed that the ngcc migration slightly varies from the ng-update migration here. Does this actually handle cases where someone sets `deps: undefined´? Should we improve this somehow and also check for the same in the ng-update migration?

packages/compiler-cli/ngcc/test/migrations/missing_injectable_migration_spec.ts
```diff
@@ -340,7 +340,7 @@ runInEachFileSystem(() => {
             TestClass.decorators = [
               { type: ${type}, args: [{${args}${propName}: [
                   ServiceA,
-                  {provide: ServiceB},
+                  {provide: ServiceB, useClass: ServiceB},
```
petebacondarwin: Would it not be better to leave this alone here and change the expectation below to be `false`?
As it stands here, this line and the next are basically testing the same thing, right?

packages/compiler-cli/ngcc/test/migrations/missing_injectable_migration_spec.ts
```diff
@@ -121,7 +121,7 @@ runInEachFileSystem(() => {
 
             export class TestClass {}
             TestClass.decorators = [
-              { type: ${type}, args: [{${args}${propName}: [{provide: forwardRef(() => MyService) }]}] }
+              { type: ${type}, args: [{${args}${propName}: [forwardRef(() => MyService)]}] }
```
petebacondarwin: So are you saying here that:

```ts
@Component({ providers: [{ provide: xxx }] })
```

is never migrated even if `xxx` is a call to `forwardRef()`?

So in order to trigger the migration you have removed the `provide:` so that this is more like a "type" provider?
 
```ts
@Component({ providers: [forwardRef(() => SomeType)] })
```

petebacondarwin (approved):

packages/compiler-cli/ngcc/test/migrations/missing_injectable_migration_spec.ts
devversion: Yes. It would be incorrect to add `@Injectable()` to the token that `xxx` is resolving to. This is because in View Engine, `{provide: xxx}` behaves like `{provide: xxx, useValue: undefined}`, hence `xxx` will never be created through DI.

The test just has been updated so that we ensure that `forwardRef` will be evaluated for `Provider` definitions that need to be migrated.

packages/compiler-cli/ngcc/test/migrations/missing_injectable_migration_spec.ts
devversion: agreed. will update 👍

devversion:
Chatted with Kara about this. We'll try to actually take care of that Ivy compatibility breaking change in the `missing-injectable` migration. We don't want it to be a new migration since that will involve more work and involves more risk. It also avoids writing a new migration guide for something that is very related to the `missing-injectable` migration anyway.

See: https://github.com/angular/angular/pull/33709. Closing in favor of that PR for now.

PR #33666, opened 2019-11-07
refactor(ivy): move all style/class DOM operations into `StylingRenderer`

matsko:


matsko:
presubmit = http://test/OCL:279159090:BASE:279159152:1573163260059:8141e664

ivy presubmit = https://test.corp.google.com/ui#id=OCL:279159090:BASE:279159295:1573163299165:6a07886c

packages/core/src/render3/styling/dom_styling_renderer.ts
```diff
@@ -0,0 +1,95 @@
...
+* found in the LICENSE file at https://angular.io/license
+*/
+import {RendererStyleFlags3, isProceduralRenderer} from '../interfaces/renderer';
+import {StylingRenderer} from '../interfaces/styling';
+import {isStylingValueDefined} from '../util/styling_utils';
+
+/**
+ * Wrapper used to apply style/class values to the DOM.
+ */
+export class DomStylingRenderer implements StylingRenderer {
```
mhevery: I am not a fan of using `class` as the rest of Ivy does not use it. Classes are tricky when it comes to minification etc...

packages/core/src/render3/styling/dom_styling_renderer.ts
```diff
@@ -0,0 +1,95 @@
...
+          elementStyle.removeProperty(prop);
+        }
+      }
+    }
+  }
+
+  private _hasRenderer() { return this._renderer && isProceduralRenderer(this._renderer); }
+}
+
+let _renderer: DomStylingRenderer|null = null;
```
mhevery: It seems that the `_renderer` is global. Which makes it non-reenterant, and prone to leaks. I don't quite understand why it should be global, can we discuss this?

packages/core/src/render3/styling/dom_styling_renderer.ts
```diff
@@ -0,0 +1,95 @@
...
+export function getStylingRenderer(): StylingRenderer {
+  if (!_renderer) {
+    _renderer = new DomStylingRenderer();
```
mhevery: The fact that we only need one instance tells me that these should just be free floating functions rather than a class.

packages/core/src/render3/styling/dom_styling_renderer.ts
```diff
@@ -0,0 +1,95 @@
...
+          const classList = element.classList;
+          if (classList != null) {
+            classList.remove(className);
+          }
+        }
+      }
+    }
+  }
+
+  setStyle(element: HTMLElement, prop: string, value: string|null, bindingIndex: number|null):
```
mhevery: We are passing `bindingIndex` but it is not used.

packages/core/src/render3/styling/dom_styling_renderer.ts
```diff
@@ -0,0 +1,95 @@
...
+      void {
+    // Use `isStylingValueDefined` to account for falsy values that should be bound like 0.
+    if (isStylingValueDefined(value)) {
+      // opacity, z-index and flexbox all have number values
+      // and these need to be converted into strings so that
+      // they can be assigned properly.
+      value = value.toString();
```
mhevery: since `value` is `string` why is this needed?

mhevery (changes requested):
I don't quite understand the motivation behind this change. I am concerned about the global state and class. Could we discuss this?

PR #33670, opened 2019-11-07
feat(localize): i18n - add MappedFileTranslationHandler` to translation in…

petebacondarwin:
…lining process

This handler allows files to be replaced per locale when doing compile
time translation inlining.

FW-1658


AndrewKushnir:
Hi @petebacondarwin, I just came across this PR and I'm wondering if this is something we should review and try to land in `10.1.0` or is it not relevant/important? Thank you.

petebacondarwin:
@AndrewKushnir I think this could land in 11.1.x

packages/localize/src/tools/src/translate/main.ts
```diff
@@ -93,6 +96,18 @@ if (require.main === module) {
...
 
+          .option('file-mapping', {
+            array: true,
+            describe:
+                'A string that configures the translator to replace a matched file with a translated file.\n' +
+                'Source files that match are replaced rather than being translated via the source code translation.\n' +
+                'The format of string is "glob => regex => replacer", where:\n' +
+                ' - `glob` is the pattern that is used to identify files to map.\n' +
+                ' - `regex` and `replacer` are used to map the source path to the path where the translated file will be.\n' +
+                '   This can include the expression "{{LOCALE}}" which will be replaced with the locale being translated.',
```
AndrewKushnir, line 107: May be add a quick example on how a string may look like?

packages/localize/src/tools/src/translate/mapped_files/mapped_file_translation_handler.ts
```diff
@@ -0,0 +1,118 @@
...
+
+/**
+ * Translate a file by replacing it with a mapped copy, writing it to the appropriate output
+ * path.
+ */
+export class MappedFileTranslationHandler implements TranslationHandler {
+  private missingTranslation: DiagnosticHandlingStrategy;
+  private mappings: FileMapping[];
+  constructor(private fs: FileSystem, options: MappedFileOptions) {
+    this.missingTranslation = options.missingTranslation || 'error';
```
AndrewKushnir, line 21: I was thinking if it makes sense to set the default value for the `missingTranslation` in a common place (like `packages/localize/src/tools/src/translate/main.ts` file)? It looks like this code should not decide what the default values of config options are, but rather operate with the value passed from the outside.

AndrewKushnir (approved):
The change looks great, thanks @petebacondarwin 👍 (just left a couple minor comment)

petebacondarwin:
Closing as there doesn't appear to be a need to do this right now in the standalone binary. The CLI would have its own implementation if necessary.

PR #29895, opened 2019-04-14
fix(platform-browser): add unique input event plugin

DerSizeS:
This plugin checks if value of element is changed before firing input event  

Internet Explorer 10+ implements 'input' event but it erroneously fires under various situations, e.g. when placeholder changes, when non english placeholder is used or a control is focused

Fixed only from Edge 15.15002
https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/101220/
This bug will not be fixed in Internet Explorer 
https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/11405058/

Closes #14440, #17951, #15299, #16151

## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [x] Bugfix
- [x] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] angular.io application / infrastructure changes
- [ ] Other... Please describe:

## What is the current behavior?
Internet Explorer 10+ implements 'input' event but it erroneously fires under various situations, e.g. when placeholder changes, when non english placeholder is used or a control is focused

Issue Number: #14440, #17951, #15299, #16151

## What is the new behavior?
If plugin config is provided – input event is fired when value of element is changed.

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

## Other information
Here is the problem reproduction (to fix IE behavior- uncomment the provider): https://stackblitz.com/edit/angular-feu5wd?embed=1&file=src/app.module.ts

Issue #17951, opened 2017-07-06
input element with ngModel initialized as null, fires "input" event on focus and blur in IE 


himajasuman:
## I'm submitting a ...

<pre><code>
[ ] Regression (behavior that used to work and stopped working in a new release)
[x] Bug report 
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead see https://github.com/angular/angular/blob/master/CONTRIBUTING.md#question
</code></pre>

## Current behavior

input element with ngModel initialized as null, fires "input" event on focus and blur for empty(null) value

## Expected behavior

Tested it with plain input element with "input" eventListener in IE, it isn't firing "input" events for above scenario. So, I assume it isn't an IE bug. "input" event (and consecutive onChange/ngModelChange) should not be fired on focus and blur.

## Minimal reproduction of the problem with instructions

https://plnkr.co/edit/q9TgF9btZKquDjJuKDVz?p=preview

https://jsfiddle.net/xry2Ltbn/4/ is an example with simple input element with input eventListener

## Please tell us about your environment

<pre><code>
Angular version: 4.2.4

Browser:
- [ ] Chrome (desktop) version XX
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [x] IE version 11
- [ ] Edge version XX
 
For Tooling issues:
- Node version: XX 
- Platform: 

Others:

</code></pre>

kara:
@himajasuman If this is still an issue, can you provide a **minimal** repro case? The one you linked to has a number of components, and it's not clear which one you are looking at.

forever14k:
@kara here is minimal reproduction: http://plnkr.co/edit/4cIdOVToP1J5HWmIvAf9?p=preview
Expected behavior: both forms pristine
Current behavior: first form is dirty and second is pristine

This bug affects only Internet Explorer 10+, more detailed comment here: https://github.com/angular/angular/issues/14440#issuecomment-350812348

uvv001:
Catch this behaviour in search field on my project.

vlad:
This might be considered a duplicate of https://github.com/angular/angular/issues/15299.  It has a link to https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/101220/ which points to a plunkr example that shows other events (like focus, click) are fired on input when the placeholder is set.

[...]

Issue #16151, opened 2017-04-19
Swedish letters in input placeholder makes form dirty on init


Drol:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => search github for a similar issue or PR before submitting
[ ] feature request
[ ] support request => Please do not submit support request here, instead see https://github.com/angular/angular/blob/master/CONTRIBUTING.md#question
```

**Current behavior**
Using reactive forms. When swedish letters (åäö) is used in a input placeholder, the form and input is initialized as dirty. <input placeholder="åäö">

This only affects IE11, or perhaps also earlier versions of IE.

**Expected behavior**
The form and input should not be dirty because of swedish letters in placeholder.

**Minimal reproduction of the problem with instructions**
http://plnkr.co/edit/u3iEtQ?p=preview
~~http://plnkr.co/edit/LpZNI9m7j8m8tgXRFdVG?p=preview~~ (old plunk with typo and not working)

**What is the motivation / use case for changing the behavior?**
bug

**Please tell us about your environment:**
Windows 10

* **Angular version:** 4.0.2, also affects 2.x

* **Browser:** Internet Explorer 11
 
* **Language:** all

Teofrey:
Sadly I noticed the same issue with CCJK characters.

kneefer:
Any feedback? Will anyone take a look on this problem? It's really annoying and forces developers to use ugly tricks in order to support IE properly.

tytskyi:
@Drol please check your own plunker - both inputs are marked as dirty in IE 11. Therefore bug is not related to swedish letters. 

This is known bug and duplicate of https://github.com/angular/angular/issues/14440

Please close the issue in favour of original.

@Teofrey, @kneefer please check the original issue.

Drol:
@tytskyi Sorry I made a typo in the plunk i provided. I have updated the original post with a working plunk.

Yes a saw that bug when i reported this one. And it may be related but it is not exactly the same type of bug.

Again sorry for the not working plunk.

apurvjain:
I am also facing the same issue with Chinese Characters as a placeholder on mdInput, for English characters looks fine but other characters the from becomes dirty instead of pristine on init. waiting for the solution?

flensrocker:
Same here with german umlauts.

AnkHansen:
And after six hours of debugging - I can vote in that Danish characters has same flaw. In my case the letter 'ø'. 

[...]

Issue #15299, opened 2017-03-20
How to prevent IE11/10/9 from triggering automatically input validation?


finalxcode:
**I'm submitting a ...**  (check one with "x")
```
[ x] bug report => search github for a similar issue or PR before submitting
[ ] feature request
[ ] support request => Please do not submit support request here, instead see https://github.com/angular/angular/blob/master/CONTRIBUTING.md#question
```

**Current behavior**

`<md-dialog-content>`
    `<form [formGroup]="privilegeForm">`
        `<md-input-container class="account-login-material" style="width: 100%">`
            `<input mdInput required placeholder="Name" formControlName="name" [(ngModel)]="name" maxlength="16">`
            `<md-hint align="start" *ngIf="privilegeForm.name.valid || privilegeForm.name.pristine">Name is required</md-hint>`
            `<md-hint align="end">{{name?.length}} / 16</md-hint>`
        `</md-input-container>`
    `</form>`
`</md-dialog-content>`

when i use `placeholder="Name"` in IE11/10/9, I get the message "Name is required" is diplayed as soon as the field gets focus.

**Expected behavior**

i expect that there is a method to prevent IE11/10/9 from triggering automatically input validation?

**Minimal reproduction of the problem with instructions**

none

**What is the motivation / use case for changing the behavior?**

when this input field is pristine and valid, the message "required" should not be displayed

**Please tell us about your environment:**

@angular/cli: 1.0.0-beta.31
node: 6.9.4
os: win32 x64

* **Angular version:** 2.0.X

@angular/common: 2.4.8
@angular/compiler: 2.4.8
@angular/core: 2.4.8
@angular/forms: 2.4.8
@angular/http: 2.4.8
@angular/material: 2.0.0-beta.2
@angular/platform-browser: 2.4.8
@angular/platform-browser-dynamic: 2.4.8
@angular/router: 3.4.8
@angular/cli: 1.0.0-beta.31
@angular/compiler-cli: 2.4.8

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

 IE11/10/9

* **Language:** [all | TypeScript X.X | ES6/7 | ES5]
 ES5

* **Node (for AoT issues):** `node --version` =   
node: 6.9.4

DzmitryShylovich:
[...]

PR #33688, opened 2019-11-08
fix(platform-browser): add unique input event plugin

### LINKED ISSUE #14440: Textarea element in form when binding model has initial state as dirty in IE
```
[ ] feature request
```

**Current behavior**
Textarea element in form when binding model has initial state as dirty in IE

**Expected behavior**
Textarea element in form when binding model has initial state as pristine in IE

Using ng-form with textarea element and bind model on that textarea

Plunker example: http://plnkr.co/edit/JEgKqSEXGpUDSRDuMXcD?p=preview
Inspect textarea in browser development tool. `textarea` and `form` element have `ng-pristine` CSS class in Chrome/Firefox/Edge while they have `ng-dirty` CSS class in IE.

* **Angular version:** 2.4.7

* **Browser:** IE 11

DerSizeS:
This plugin checks if value of element is changed before firing input event  

Internet Explorer 10+ implements 'input' event but it erroneously fires under various situations, e.g. when placeholder changes, when non english placeholder is used or a control is focused

Fixed only from Edge 15.15002
https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/101220/
This bug will not be fixed in Internet Explorer 
https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/11405058/

Closes #14440, #17951, #15299, #16151

## PR Checklist

## PR Type

- [x] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Internet Explorer 10+ implements 'input' event but it erroneously fires under various situations, e.g. when placeholder changes, when non english placeholder is used or a control is focused

Issue Number: #14440, #17951, #15299, #16151

## What is the new behavior?
If plugin config is provided – input event is fired when value of element is changed.

- [ ] Yes
- [x] No

## Other information
Here is the problem reproduction (to fix IE behavior- uncomment the provider): https://stackblitz.com/edit/angular-feu5wd?embed=1&file=src/app.module.ts

I had to recreate the [PR](https://github.com/angular/angular/pull/29895) due to the accidental deletion of the associated fork.

### DIFF
--- a/packages/platform-browser/src/dom/events/unique_input_event.ts
+++ b/packages/platform-browser/src/dom/events/unique_input_event.ts
@@ -1,0 +1,137 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {DOCUMENT} from '@angular/common';
+import {Inject, Injectable, InjectionToken, Optional} from '@angular/core';
+
+import {EventManagerPlugin} from './event_manager';
+
+/**
+ * The injection token for enabling unique input event plugin.
+ *
+ * @publicApi
+ */
+export const UNIQUE_INPUT_EVENT_PLUGIN_CONFIG =
+    new InjectionToken<UniqueInputEventPluginConfig>('UniqueInputEventPluginConfig');
+
+/**
+ * @publicApi
+ */
+export interface UniqueInputEventPluginConfig {
+  /**
+   * Function that determines that plugin should be applied to element
+   * Default is HTMLInputElement or HTMLTextAreaElement
+   */
+  shouldApplyToElement?(element: Element): boolean;
+  /**
+   * Function that determines that event is not require check for uniqueness
+   * By default, all non user agent events is trusted
+   */
+  shouldTrustEvent?(event: Event): boolean;
+}
+
+
+const DEFAULT_UNIQUE_INPUT_EVENT_PLUGIN_CONFIG: UniqueInputEventPluginConfig = {
+  shouldApplyToElement: (element: HTMLElement) => {
+    return element instanceof HTMLInputElement || element instanceof HTMLTextAreaElement;
+  },
+  /** Trust all non user agent events */
+  shouldTrustEvent: (event: Event) => !event.isTrusted,
+};
+
+
+/**
+ * @description
+ *
+ * This plugin checks if value of element is changed before firing input event
+ *
+ * Internet Explorer 10+ implements 'input' event but it erroneously fires under various situations,
+ * e.g. when placeholder changes, when non english placeholder is used or a control is focused.
+ * Fixed only from Edge 15.15002
+ * https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/101220
+ * Will not be fixed in Internet Explorer
+ * https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/11405058/
+ *
+ * @usageNotes
+ * ### Example
+ *
+ * ```
+ * @NgModule({
+ *   providers: [{
+ *      provide: EVENT_MANAGER_PLUGINS, multi: true,
+ *      useClass: UniqueInputEventPlugin, deps: [DOCUMENT],
+ *   }]
+ * })
+ * class MyModule {}
+ * ```
+ *
+ * If there is a need to control when the plugin will be enabled (for example, IE 10+),
+ * use `FactoryProvider` instead of `ClassProvider`
+ *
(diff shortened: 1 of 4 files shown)


petebacondarwin (changes requested):
This is a really nice piece of work @DerSizeS - thanks for submitting it.

I am torn though.

Since this is only fixing a bug in a small set of browsers (IE10 and IE11, right?) it feels like it is too much code to add to the standard framework. The event manager architecture is already in place in the framework to allow this to be added on a per project basis without having to modify the core.

I feel that a more effective solution, which would not impact projects that do not require IE support, would be to have this as some kind of extra module that a developer could add to their app as they need it - a bit like the polyfills approach we have already.  But I do feel like this should be easily available to developers without having to go look for some 3rd party library to do this...

@IgorMinar and @kara do you have a view on this?

DerSizeS:
How about exposing `UniqueInputEventPlugin` itself? That way developers could use `FactoryProvider` for conditional enabling plugin. 

Example:
```
export function setupUniqueInputEventManagerPlugin(document: Document) {
  if (isIE()) {
    return new UniqueInputEventPlugin(document);
  }
  return NOOP_EVENT_MANAGER_PLUGIN;
}
```
```
providers: [
  {
    provide: EVENT_MANAGER_PLUGINS,
    useFactory: setupUniqueInputEventManagerPlugin,
    deps: [DOCUMENT],
    multi: true,
  },
],
```

petebacondarwin:
That would still leave the code in framework, which is what I am unsure about. We need a call from @IgorMinar or @kara, I think.

IgorMinar:
this is a great PR, but it also increases the payload size for all apps.

We can merge it only If we figure out how to make this provider optional or IE specific.

I don't believe that there is a way in the CLI to include some code only into the legacy es5 bundle or is there, @filipesilva @clydin?

packages/platform-browser/src/browser.ts
```diff
@@ -70,6 +71,12 @@ export const BROWSER_MODULE_PROVIDERS: StaticProvider[] = [
     multi: true,
     deps: [DOCUMENT, NgZone, PLATFORM_ID]
   },
+  {
+    provide: EVENT_MANAGER_PLUGINS,
+    useClass: UniqueInputEventPlugin,
+    multi: true,
+    deps: [DOCUMENT, [new Optional(), UNIQUE_INPUT_EVENT_PLUGIN_CONFIG]]
+  },
```
petebacondarwin: So if we excluded this block of code from the library (i.e from this PR) then I believe it would get tree shaken by default and not add to the bundle size. But then the onus is on the app developer to find out about this and add it back in for their IE scenarios...

petebacondarwin:
@DerSizeS - I think we are agreed that we could land this if we only implement and export the `UniqueInputEventPlugin` but do not wire it up. We should also add a section to the API docs for it, explaining when it should be used and how it can be provided. By the way, I think it could just use the `useClass` provider, why did you suggest that the developer would need to use the `useFactory` approach?

DerSizeS:
> By the way, I think it could just use the `useClass` provider, why did you suggest that the developer would need to use the `useFactory` approach?

With `useClass` this plugin will be enabled for all browsers, in the case of `useFactory` developer will be able to control when plugin will be enabled (e.g. IE 10+).

For this scenario, I think it's worth adding `NOOP_EVENT_MANAGER_PLUGIN`.

filipesilva:
Exporting it but not wiring would not increase all users payload size, but would still increase all payload size for users that make use of it. Making it conditional in user code still includes it in bundles.

I don't think we have a way to add things to the es5-only polyfills though. @clydin is there a way?

vladimirverbitski:
This fix from @DerSizeS also help with dynamics form in IE, placeholder also trigger after page loading.
Thx!

petebacondarwin:
@IgorMinar how about we just export the `UniqueInputEventPlugin` and document it?

IgorMinar:
We just talked to @atscott yesterday about the unrelying bug and Andrew is doing the research to determine if we can drop IE9&10 support and then remove the problematic code that is causing the event duplication in modern browsers.

Does that sound like an ok solution to this?

ai977313677:
"when a non-English placeholder is used or control is focused"

I cannot find the issue on websites. I think we can use the Unicode or HTML Entity code like this.

```html
 <input nz-input formControlName="appName" placeholder="&#35831;&#36755;&#20837;&#21517;&#31216;">
```

I checked. HTML Entity works well.

petebacondarwin:
Angular v11.0.0+ has dropped support for IE9 & 10, so we ought to be able to just remove the problematic code now, yes?

mhevery:
Closing as this PR seems abandoned, and we don't want to add more API surface to the event plugin API.

PR #33689, opened 2019-11-08
fix(ngcc): move adjacent statements after helpers

petebacondarwin:
// FW-1668


gkalpak (approved):
LGTM :+1:

Could you add some more information in the commit message on why _"they must also put these statements after any helper calls"_?

petebacondarwin:
> Could you add some more information in the commit message on why "they must also put these statements after any helper calls"?

I would if I knew why :-)

PR #33694, opened 2019-11-08
fix(zone.js): define tasks array as enumerable: false on element

### LINKED ISSUE #33692: zone.js attaches some enumerable properties elements
zone.js 0.10.2

zone.js attaches some `enumerable: true` properties to elements with event listeners attached like `__zone_symbol__clickfalse` in case of a `click` listener.

## 🔬 Minimal Reproduction

Go to https://angular.io, select the search field with the element inspector, then `JSON.stringify($0, null, 2)`

you see:

image

There are a lot more keys, but the other ones are properly attached as `enumerable: false`
image

sod:
- [x] Bugfix

fixes https://github.com/angular/angular/issues/33692

### DIFF
--- a/packages/zone.js/lib/common/events.ts
+++ b/packages/zone.js/lib/common/events.ts
@@ -418,7 +418,12 @@ export function patchEventTarget(
             }
           }
         } else {
-          existingTasks = target[symbolEventName] = [];
+          existingTasks = [];
+          Object.defineProperty(target, symbolEventName, {
+            value: existingTasks,
+            configurable: true,
+            writable: true,
+          });
         }
         let source;
         const constructorName = target.constructor['name'];


JiaLiPassion:
@sod , thanks for the PR, but currently `@angular/core` depends on the `eventListeners` method (https://github.com/angular/angular/blob/1b8b04cf26401508cd171374e044e9510842e4f2/packages/core/src/debug/debug_node.ts), which is added by `zone.js`, and inside `eventListeners` implementaiton, it require those property to be `enumerable`. 
The code of `zone.js` is here
https://github.com/angular/angular/blob/c7542a1d09b27b027a0367a1f3732a908cdd617a/packages/zone.js/lib/common/events.ts#L598.

sod:
Thanks for clarification.

Isn't there a security risk to expose event handlers globally accessible on elements? And why are the handlers spreaded onto the element? Doesn't this make the DOM megamorphic and thus code that works with the DOM slower? Why not use a weakmap to internally connect handlers & elements.

JiaLiPassion:
@sod, yeah, the purpose to attach handlers in such data structure is to increase the performance (especially memory GC), just like you said, maybe we can use some better design, we may discuss it here and make some refactor. Thank you!

sod:
ok :)

I'll close this PR then.

PR #33682, opened 2019-11-08
fix: full cldr

ghost:
Blocked on https://github.com/angular/angular/pull/33699

## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [x] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] angular.io application / infrastructure changes
- [ ] Other... Please describe:

## What is the current behavior?

Fixes #33681

## What is the new behavior?
Enables all the previous locales, plus new ones as per cldr-36

## Does this PR introduce a breaking change?
No, but the PR whose behavior it fixes, does

Some locales were removed, some were renamed.

PR #33699, opened 2019-11-08
fix(common): update locales data to remove � characters

josephperrott:
Running `clang-format` via `gulp-clang-format` was causing intermittent instances of Unicode's replacement character (`\uFFFD`/�).  By running `clang-format` via `shelljs` in the `gulp` script, this intermittent difference is no longer occurring.

Unblocks #33682


IgorMinar (approved):
sanity restored! thanks for figuring it out

doom777:
glad that the outcome is good, but overall why is running clang from yarn from shell from shelljs from gulp, good and produces good outcome, but running clang from gulp, bad? I think that the problem is that clang from gulp was not pulling in the correct config

doom777:
Also, it seems that `gulp-clang-format` is an angular library. It's a shame to abandon it

petebacondarwin (approved):

gkalpak (approved):

devversion (approved):

PR #33700, opened 2019-11-08
refactor(ivy): use renderer as a source of truth for style/class debugging

matsko:
Prior to this fix, the styling merge algorithm would intercept all
style/class writes and collect them into to map using a built-in
`StyleApplyFn` function. This setup does works for the styling
merge algorithm, but not for the direct write algorithm. For this
reason the renderer needs to be used instead because both algorithms
use that when applying style/class values.


packages/core/src/render3/styling/styling_debug.ts
```diff
@@ -478,21 +472,88 @@ export class NodeStylingDebug implements DebugNodeStyling {
...
+class MockRendererForStyling {
+  private _classAttr: string = '';
+  private _styleAttr: string = '';
+  private _styles: {[key: string]: any} = {};
+  private _classes: {[key: string]: any} = {};
+
+  listen(target: any, eventName: string, callback: (event: any) => boolean | void): () => void {
+    // this method is only used to trigger the check for the renderer
+    // as a procedural renderer to pass...
+    return () => {};
```
mhevery: Could you add `throw new Error('DebugStyle Mock does not implement')` to all of the methods which are not needed?

packages/core/src/render3/styling/styling_debug.ts
```diff
@@ -478,21 +472,88 @@ export class NodeStylingDebug implements DebugNodeStyling {
...
+    applyStylingViaContext(this.context.context, this._tNode, renderer as {
+    } as ProceduralRenderer3, mockElement, data, true, mapFn, sanitizer, true, this._isClassBased);
+
+    const values =
+        this._isClassBased ? renderer.computeClassValues() : renderer.computeStyleValues();
+    Object.keys(values).forEach(prop => { fn(prop, values[prop]); });
+  }
+}
+
+class MockRendererForStyling {
```
mhevery: As of right now you don't have type safety. I think `class MocRenderForStyling implements Renderer3` and than add `throw new Error('DebugStyle Mock does not implement')` to all of the methods which are not needed?

packages/core/src/render3/styling/style_string_parser.ts
```diff
@@ -0,0 +1,117 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import {StylingMapArray} from '../interfaces/styling';
+
+const enum Char {
```
mhevery: I think this file needs some unit tests (in additional to the integration tests which you have)

mhevery (changes requested):

PR #33644, opened 2019-11-07
fix(ivy): Run ChangeDetection on moved views

mhevery:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [X] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [X] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [X] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] angular.io application / infrastructure changes
- [ ] Other... Please describe:

## What is the current behavior?

Issue Number: N/A

## Does this PR introduce a breaking change?

- [ ] Yes
- [X] No

PR #33702, opened 2019-11-08
Transplant views refactor

mhevery:
Based on https://github.com/angular/angular/pull/33644 (please ignore during review)

## PR Checklist

## PR Type

- [X] Bugfix/Performance
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [X] No


packages/core/test/acceptance/change_detection_spec.ts
```diff
@@ -1098,4 +1102,140 @@ describe('change detection', () => {
...
+          {{greeting}} {{nameWithLog}}!
+        </ng-template>
+      `
+    })
+    class DeclareComp {
+      @ViewChild('myTmpl')
+      myTmpl !: TemplateRef<any>;
+      name: string = 'world';
+      get nameWithLog() {
+        if (ivyEnabled && !getCheckNoChangesMode()) {
```
pkozlowski-opensource: I would _personally_ prefer that we write those tests without being `ivy` specific and without digging into ivy internals. Such tests are brittle and we are not sure that we've got 1:1 behaviour with VE.

Couldn't we assert that the both interpolations are updated in the moved embedded view?

packages/core/src/render3/instructions/shared.ts
```diff
@@ -1481,6 +1481,32 @@ function refreshDynamicEmbeddedViews(lView: LView) {
...
+      if (movedViews !== null) {
+        // We should only CD moved views if the component where they were inserted does not match
+        // the component where they were declared. Moved views also contains intra component moves,
+        // which we don't care about.
+        // TODO(misko): this is not the most efficient way to do this as we have to do a lot of
+        // searches. Will refactor for performance later.
+        const declaredComponentLView = findComponentView(lView);
+        for (let i = 0; i < movedViews.length; i++) {
+          const movedLView = movedViews[i] !;
+          let parentLView = movedLView[PARENT];
```
pkozlowski-opensource: Not sure why we've got the logic of getting a parent and `while (isLContainer(parentLView))` here. Shouldn't `findComponentView` go up the embedded views? 

As the very minimum it would be great to add more comments here as the logic here starts to be non-obvious.

packages/core/src/render3/instructions/shared.ts
```diff
@@ -1481,6 +1481,32 @@ function refreshDynamicEmbeddedViews(lView: LView) {
...
+          while (isLContainer(parentLView)) {
+            parentLView = parentLView[PARENT];
+          }
+          const insertedComponentLView = findComponentView(parentLView !);
+          const insertionIsOnPush =
+              (insertedComponentLView[FLAGS] & LViewFlags.CheckAlways) !== LViewFlags.CheckAlways;
```
pkozlowski-opensource: We are not really determining `insertionIsOnPush` here but rather than it is _not_ CheckAlways, right?

It is not clear to me if this is intentional but I think that we should add comments + tests to explain what happens depending on the insertion view flags (OnPush, OnPush + marked for check, detatched etc.)

packages/core/src/render3/instructions/shared.ts
mhevery: I have refactored it to make it more clear. But the short of it is that `findComponentView` does not cross `LContainer` boundaries.

mhevery:
[presubmit](http://test/OCL:279994760:BASE:279994876:1573581901027:431c4dff)
[ivy presubmit](http://test/OCL:279994760:BASE:279995086:1573581949154:b5bf93b8)

packages/core/src/render3/interfaces/view.ts
```diff
@@ -194,6 +195,75 @@ export interface LView extends Array<any> {
...
+   * (it has been transplanted across component.)
+   *
+   * see also:
+   *   - https://hackmd.io/@mhevery/rJUJsvv9H write up of the problem
+   *   - `LContainer[ACTIVE_INDEX]` for flag which marks which `LContainer` has transplanted views.
+   *   - `LContainer[TRANSPLANT_HEAD]` and `LContainer[TRANSPLANT_TAIL]` storage for transplanted
+   *   - `LView[DECLARATION_LCONTAINER]` similar problem for queries
+   *   - `LContainer[MOVED_VIEWS]` similar problem for queries
+   */
+  [DECLARATION_COMPONENT_VIEW]: LView|null;
```
kara: We may want to note somewhere that the DECLARATION_COMPONENT_VIEW is the view itself if the LView is a component view (the simplest / most common case).

packages/core/test/acceptance/change_detection_spec.ts
```diff
@@ -1179,47 +1178,62 @@ describe('change detection', () => {
...
     it('should CD with declaration', () => {
-      fixture.detectChanges();
-      ivyEnabled && expect(log).toEqual(['Insert']);
+      // NOTE: The CD of VE and Ivy is different and is captured in the assertions:
+      // `expect(log).toEqual(ivyEnabled ? [...] : [...])`
+      //
+      // The reason for this difference is in the algorithm which VE and Ivy use to deal with
+      // transplanted views:
+      // - VE: always runs CD at insertion point. If the insertion component is `OnPush` and the
+      //   transplanted view is `CheckAlways` than the insertion component will be changed to
```
kara:
```suggestion
      //   transplanted view is `CheckAlways` then the insertion component will be changed to
```

packages/core/src/render3/interfaces/container.ts
```diff
@@ -46,6 +46,41 @@ export const VIEW_REFS = 8;
...
+
+  /**
+   * Flag to signify that this `LContainer` may have transplanted views which need to be CD.
+   * (see: `LView[DECLARATION_COMPONENT_VIEW])`.
+   *
+   * This flag once set is never unset for the `LContainer`. This means that when unset we can skip
+   * a lot of work in `refreshDynamicEmbeddedViews`. But when set we still need to verify
+   * that the `MOVED_VIEWS` are transplanted and on-push.
+   */
+  HAS_TRANSPLANTED_VIEWS = 1,
```
kara: I'm guessing that the reason this is not on the TView is because you could have a component that has a default template in the same view (so it can be sometimes intra-component, sometimes inter-component)? Seems like we should add a test for that case to make sure we don't break it

packages/core/src/render3/instructions/shared.ts
```diff
@@ -10,24 +10,24 @@ import {ErrorHandler} from '../../error_handler';
...
+import {CHILD_HEAD, CHILD_TAIL, CLEANUP, CONTEXT, DECLARATION_COMPONENT_VIEW, DECLARATION_VIEW, ExpandoInstructions, FLAGS, HEADER_OFFSET, HOST, INJECTOR, InitPhaseState, LView, LViewFlags, NEXT, PARENT, RENDERER, RENDERER_FACTORY, RootContext, RootContextFlags, SANITIZER, TData, TVIEW, TView, TViewType, T_HOST} from '../interfaces/view';
 import {assertNodeOfPossibleTypes} from '../node_assert';
 import {isNodeMatchingSelectorList} from '../node_selector_matcher';
 import {ActiveElementFlags, enterView, executeElementExitFn, getBindingsEnabled, getCheckNoChangesMode, getIsParent, getPreviousOrParentTNode, getSelectedIndex, hasActiveElementFlag, incrementActiveDirectiveId, leaveView, leaveViewProcessExit, setActiveHostElement, setBindingIndex, setBindingRoot, setCheckNoChangesMode, setCurrentDirectiveDef, setCurrentQueryIndex, setPreviousOrParentTNode, setSelectedIndex} from '../state';
```
kara: Won't let me comment on the right line, but we should remove the unused `findComponentView` export

packages/core/src/render3/instructions/shared.ts
```diff
@@ -172,6 +172,9 @@ export function createLView<T>(
   lView[SANITIZER] = sanitizer || parentLView && parentLView[SANITIZER] || null !;
   lView[INJECTOR as any] = injector || parentLView && parentLView[INJECTOR] || null;
   lView[T_HOST] = tHostNode;
+  lView[DECLARATION_COMPONENT_VIEW] = tView.type == TViewType.Embedded ?
```
kara: This seems to imply that we don't need the `findComponentView` function anymore because we save the declared component view upon creation now (and thus don't need to walk up the declaration tree all the time). Seems like we should clean up all the remaining `findComponentView` usages. Can we do that in a follow-up PR and maybe add a TODO above the function for now?

kara (changes requested):
Like the idea of only searching moved views if the flag is set! Some minor things below

packages/core/src/render3/interfaces/view.ts
```diff
@@ -194,6 +195,75 @@ export interface LView extends Array<any> {
...
+   * ```
+   * The `thenBlock` and `elseBlock` is moved but not transplanted.
+   *
+   * Example showing inter component `LView` movement (transplanted view).
+   * ```
+   * <#VIEW #myComp>
+   *   <ng-template #myTmpl>...</ng-template>
+   *   <insertion-component [template]="myTmpl"></insertion-component>
+   * </#VIEW>
+   * ```
```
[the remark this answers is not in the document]
mhevery: 👍 (thanks for appreciating/noticing)

packages/core/src/render3/interfaces/view.ts
mhevery: added

packages/core/src/render3/instructions/shared.ts
mhevery: Good point. I had a quick look and there are many uses of this function so I don't feel comfortable doing it in this PR. I will start a new one.

packages/core/src/render3/instructions/shared.ts
mhevery: Good point. I had a quick look and there are many uses of this function so I don't feel comfortable doing it in this PR. I will start a new one.

packages/core/src/render3/instructions/shared.ts
mhevery: removed

packages/core/src/render3/instructions/shared.ts
```diff
@@ -1474,45 +1478,67 @@ function refreshDynamicEmbeddedViews(lView: LView) {
   while (viewOrContainer !== null) {
     // Note: viewOrContainer can be an LView or an LContainer instance, but here we are only
     // interested in LContainer
-    if (isLContainer(viewOrContainer) && viewOrContainer[ACTIVE_INDEX] === -1) {
+    let activeIndexFlag: ActiveIndexFlag;
+    if (isLContainer(viewOrContainer) &&
+        (activeIndexFlag = viewOrContainer[ACTIVE_INDEX]) >> ActiveIndexFlag.SHIFT ===
```
kara:
```suggestion
        getLContainerActiveIndex(activeIndexFlag = viewOrContainer[ACTIVE_INDEX]) ===
```

packages/core/src/render3/interfaces/view.ts
```diff
@@ -194,6 +195,78 @@ export interface LView extends Array<any> {
...
+   *
+   * Example:
+   * ```
+   * <#VIEW #myComp>
+   *  <div *ngIf="true">
+   *   <ng-template #myTmpl>...</ng-template>
+   *  </div>
+   * </#VIEW>
+   * ```
+   * In the above case `DECLARATION_VIEW` for `myTmpl` points to the `LView` of `ngIf` where as
```
kara:
```suggestion
   * In the above case `DECLARATION_VIEW` for `myTmpl` points to the `LView` of `ngIf` whereas
```

packages/core/src/render3/interfaces/view.ts
```diff
@@ -194,6 +195,78 @@ export interface LView extends Array<any> {
...
+   * `DECLARATION_COMPONENT_VIEW` points to `LView` of the `myComp` which owns the template.
+   *
+   * The reason for this is that all embedded views are always check-always where as the component
```
kara:
```suggestion
   * The reason for this is that all embedded views are always check-always whereas the component
```

packages/core/src/render3/instructions/shared.ts
```diff
@@ -1474,45 +1478,67 @@ function refreshDynamicEmbeddedViews(lView: LView) {
...
+      // Yes the `LView` is transplanted.
+      // Here we would like to know if the component is `OnPush`. We don't have
+      // explicit `OnPush` flag instead we set `CheckAlways` to false (which is `OnPush`)
+      // Not to be confused with `ManualOnPush` which is used with wether a DOM event
+      // should automatically mark a view as dirty.
+      const insertionComponentIsOnPush =
+          (insertedComponentLView[FLAGS] & LViewFlags.CheckAlways) === 0;
+      if (insertionComponentIsOnPush) {
+        // Here we know that the template has been transplanted across components and is
+        // on-push (not just moved within a component). If the insertion is marked dirty, than
```
kara:
```suggestion
        // on-push (not just moved within a component). If the insertion is marked dirty, then
```

packages/core/src/render3/instructions/shared.ts
kara: Sounds good to me

packages/core/src/render3/instructions/shared.ts
mhevery: I don't think I can do this one. It is written specifically this way so that I don't have to read the `[ACTIVE_INDEX]` more than once.

kara: Ohh I see what you mean. OK resolving.

kara (approved):
LGTM

PR #33705, opened 2019-11-09
fix(ivy): constant object literals shared across element and component instances

crisbeto:
Currently if a consumer does something like the following, the object literal will be shared across the two elements and any instances of the component template. The same applies to array literals:

```
<div [someDirective]="{}"></div>
<div [someDirective]="{}"></div>
```

These changes make it so that we generate a pure function even if an object is constant so that each instance gets its own object.

Note that the original design for this fix included moving the pure function factories into the `consts` array. In the process of doing so I realized that pure function are also used inside of directive host bindings which means that we don't have access to the `consts`.

These changes also:
* Fix an issue that meant that the `pureFunction0` instruction could only be run during creation mode.
* Make the `getConstant` utility slightly more convenient to use. This isn't strictly required for these changes to work, but I had made it as a part of a larger refactor that I ended up reverting.

### DIFF
--- a/packages/compiler/src/render3/view/template.ts
+++ b/packages/compiler/src/render3/view/template.ts
@@ -1420,11 +1420,9 @@ export class ValueConverter extends AstMemoryEfficientTransformer {
         array.span, array.sourceSpan, this.visitAll(array.expressions), values => {
           // If the literal has calculated (non-literal) elements transform it into
           // calls to literal factories that compose the literal and will cache intermediate
-          // values. Otherwise, just return an literal array that contains the values.
+          // values.
           const literal = o.literalArr(values);
-          return values.every(a => a.isConstant()) ?
-              this.constantPool.getConstLiteral(literal, true) :
-              getLiteralFactory(this.constantPool, literal, this.allocatePureFunctionSlots);
+          return getLiteralFactory(this.constantPool, literal, this.allocatePureFunctionSlots);
         });
   }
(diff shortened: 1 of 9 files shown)


mhevery (approved):

mhevery:
[presubmit](http://test/OCL:279842915:BASE:279842961:1573518503205:54f41e2e)
[ivy-presubmit](http://test/OCL:279842915:BASE:279843113:1573518555242:71aa6611)

packages/compiler/src/render3/view/template.ts
```diff
@@ -1406,9 +1406,7 @@ export class ValueConverter extends AstMemoryEfficientTransformer {
           // calls to literal factories that compose the literal and will cache intermediate
           // values. Otherwise, just return an literal array that contains the values.
```
kara: Can we update the comment to match the code? We no longer return literal arrays.

packages/compiler/src/render3/view/template.ts
```diff
@@ -1419,9 +1417,7 @@ export class ValueConverter extends AstMemoryEfficientTransformer {
       // values. Otherwise, just return an literal array that contains the values.
```
kara: Same here

kara (approved):
LGTM once comments are updated

crisbeto:
Rebased and updated the comments.

PR #33706, opened 2019-11-09
fix(ivy): provider override via TestBed should remove old providers from the list

AndrewKushnir:
While overriding providers in Ivy TestBed (via `TestBed.overrideProvider` call), the old providers were retained in the list, since the override takes precedence. However, presence of providers in the list might have side-effect: if a provider has the `ngOnDestroy` lifecycle hook, this hook will be registered and invoked later (when component is destroyed). This commit updates TestBed logic to clear provider list by removing the ones which have overrides.

This PR resolves FW-1670.

## PR Type

- [x] Bugfix

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/test/test_bed_spec.ts
+++ b/packages/core/test/test_bed_spec.ts
@@ -16,6 +16,13 @@ import {onlyInIvy} from '@angular/private/testing';
 
 const NAME = new InjectionToken<string>('name');
 
+@Injectable({providedIn: 'root'})
+class SimpleService {
+  static ngOnDestroyCalls: number = 0;
+  id: number = 1;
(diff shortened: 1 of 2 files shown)


AndrewKushnir:
Initial set of g3 presubmits:
- [VE Presubmit](http://test/OCL:279428001:BASE:279428027:1573264797948:e8f36495)
- [Ivy Presubmit](http://test/OCL:279428001:BASE:279428175:1573264874750:cc3cf60b)

pkozlowski-opensource (approved):
LGTM

atscott (approved):

AndrewKushnir:
FYI, presubmits (including global for Ivy) look good. Thank you.

PR #33709, opened 2019-11-09
feat(core): missing-injectable migration should migrate empty object literal providers

devversion:
In View Engine, providers which neither used `useValue`, `useClass`,
`useFactory` or `useExisting`, were interpreted differently.

e.g.

```ts
{provide: X} -> {provide: X, useValue: undefined}, // this is how it works in View Engine
{provide: X} -> {provide: X, useClass: X}, // this is how it works in Ivy
```

The missing-injectable migration should migrate such providers to the
explicit `useValue` provider. This ensures that there is no unexpected
behavioral change when updating to v9.

See: https://next.angular.io/guide/ivy-compatibility#less-common-changes

### DIFF
--- a/packages/core/schematics/migrations/missing-injectable/transform.ts
+++ b/packages/core/schematics/migrations/missing-injectable/transform.ts
@@ -6,9 +6,8 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {forwardRefResolver} from '@angular/compiler-cli/src/ngtsc/annotations/src/util';
 import {Reference} from '@angular/compiler-cli/src/ngtsc/imports';
-import {DynamicValue, PartialEvaluator, ResolvedValue} from '@angular/compiler-cli/src/ngtsc/partial_evaluator';
+import {DynamicValue, ResolvedValue} from '@angular/compiler-cli/src/ngtsc/partial_evaluator';
 import {TypeScriptReflectionHost} from '@angular/compiler-cli/src/ngtsc/reflection';
 import * as ts from 'typescript';
 
@@ -16,6 +15,7 @@ import {getAngularDecorators} from '../../utils/ng_decorators';
 
 import {ResolvedDirective, ResolvedNgModule} from './definition_collector';
 import {ImportManager} from './import_manager';
+import {ProviderLiteral, ProvidersEvaluator} from './providers_evaluator';
 import {UpdateRecorder} from './update_recorder';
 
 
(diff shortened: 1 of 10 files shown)


devversion:
The schematic ng-update migration (this PR) should be actually good to go. The question whether we should add such a migration to ngcc is still open. I'd need some feedback on that.

This PR also makes the `visitObjectLiteralExpression` method of `StaticIntepreter` protected, so that it can be overwritten by derived classes. Not sure if that is desired, but I think the interpreter is *very* powerful and it would be useful to do that for all methods. Happy to send a PR. cc. @alxhub 

<sup>**Also** it currently won't be possible to do something in ngcc migrations without refactoring the ngcc migration infrastructure. Though, here is a proof that we could get it working: [See here](https://github.com/devversion/angular/commit/35bb1348bf995437c4b0a1774655a1b6781a3b9e)</sup>

packages/core/schematics/migrations/missing-injectable/transform.ts
```diff
@@ -160,11 +167,39 @@ export class MissingInjectableTransform {
   }
 
   /**
-   * Evaluates the given TypeScript expression using the partial evaluator with
-   * the foreign function resolver for handling "forwardRef" calls.
+   * Migrates object literal providers which do neither use useValue, useClass,
```
kara:
```suggestion
   * Migrates object literal providers which do not use useValue, useClass,
```

kara (approved):
LGTM, once typo is fixed

packages/core/schematics/migrations/missing-injectable/transform.ts
devversion: Fixed.

IgorMinar:
do we need to add this to the list of migrations in https://next.angular.io/guide/updating-to-version-9#automated-migrations-for-version-9 ?

alxhub (approved):
👍 

I don't think we need to address this on the ngcc side. Every example of this kind of provider that I've seen has been from people trying to mock things out in tests. I doubt there are many libraries out there with unexpected `undefined` providers from this, as that would cause runtime failures.

devversion:
@alxhub Sounds reasonable to me. Thanks for the feedback!

@IgorMinar Yes. We should add this to the list of migrations. When I chatted with @kara about handling this pattern in an migration, we talked about putting it into the existing `missing-injectable` migration since we do not want to create a separate migration w/ separate migration guide. 

I think we should add it to the `missing-injectable` migration guide, but that somehow feels confusing since the pattern this PR handles is not necessarily related to a missing `@Injectable()`.

Issue #33687, opened 2019-11-08
Renderer2 method listen does not work if wrapped with runOutsideAngular method of injected NgZone


sharikovvladislav:
# 🐞 bug report

### Affected Package

 The issue is caused by package @angular/core

### Is this a regression?

 Yes, the previous version in which this bug was not present was: 9.0.0-rc.1 ViewEngine (without Ivy)

### Description

 A clear and concise description of the problem...

The problem is around Renderer2 `listen` method and `NgZone` `runOutsideAngular` method.

About repro application: I have simple page with div at the center of it. In `ngAfterViewInit` lifecycle hook I add listeners using injected `Renderer2` and it's `listen` method.

I also have 3 unit tests:
1. general test about component is created successfully
2. unit test for testing `renderer.listen` property to listen events
3. unit test same as previous but wrapped into `ngZone.runOutsideAngular` method.

##### Actual behavior: 
In ViewEngine any of those tests are green and success. In Ivy `listen` wrapped with `runOutsideAngular` test fails. 

##### I expect: 
It does not matter is `listen` wrapped in `runOutsideAngular` or not it must work both in ViewEngine and Ivy.

#### The code

The component:

```
export class AppComponent {
  @ViewChild('ref', { static: true }) ref: ElementRef;
  title = 'latest-ng-ivy-boilerplate';

  constructor(private renderer: Renderer2, private ngZone: NgZone) {}

  ngAfterViewInit() {
    this.ngZone.runOutsideAngular(() => {
      this.renderer.listen(this.ref.nativeElement, 'click', this.handlerFromOutsideNgZone);
    });

    this.renderer.listen(this.ref.nativeElement, 'click', this.handlerFromNgZone);
  }

  handlerFromNgZone(evt: MouseEvent) {
    console.log('>> clicked from NgZone', evt);
  }

  handlerFromOutsideNgZone(evt: MouseEvent) {
    console.log('>> clicked from outside NgZone', evt);
  }
}
```

The tests:

```

  it(`should listen click without runOutsideAngular wrapper`, () => {
    const fixture = TestBed.createComponent(AppComponent);
    const debugElement = fixture.debugElement;
    const component = fixture.componentInstance;

    spyOn(component, 'handlerFromNgZone').and.callThrough();
    fixture.detectChanges();

    const elementWithListenerDebugElement = debugElement.query(By.css('.click-me'));

    const event = {};

    elementWithListenerDebugElement.triggerEventHandler('click', event);

[...]

PR #20672, opened 2017-11-28
fix(core): should use native addEventListener in ngZone

JiaLiPassion:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

```
[x] Bugfix
[ ] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Documentation content changes
[ ] angular.io application / infrastructure changes
[ ] Other... Please describe:
```

## What is the current behavior?

Issue Number: 20673
fix #20673

in #18993, there is an optimized version of `addEventListener`, but I used the wrong `addEventListener` when we run in `ngZone`, we should use native `addEventListener` to get better performance when the handler was added in `ngZone`.

And the other issue is in #18993, we support `BLACK_LISTED_EVENTS`, and if an event is blacklisted, it should not run in `angular` zone, but current version it will not work.

## What is the new behavior?
When event handler was added in `angular zone`, use `non patched` addEventListener to get better performance.

And `BLACK_LISTED_EVENTS` should work and run outside of angular zone.

## Does this PR introduce a breaking change?
```
[ ] Yes
[x] No
```

## Other information
@mhevery , please review, thank you!

PR #33711, opened 2019-11-09
fix: resolve event listeners not correct when register outside of ngZone

JiaLiPassion:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #33687

When register event listener outside of `ngZone`, call `triggerEventListener` will not work because `element.eventListeners` will not have the registered event listener.

## What is the new behavior?
`element.eventListeners` will have the registered event listener.

- [ ] Yes
- [x] No

## Other information
In this PR, I removed the logic from `DomEventsPlugin`, because those logic are duplicate with `zone.js` implementation, those logic are for performance improvement at several months ago,  but there is a bug, so the performance improvement in fact not working, so I created another PR here #20672, and it seems have several issues in G3, so now I would like to resolve this issue first and then continue to re-create the PR for performance improvement. 

In this PR, there is no performance down because the logic `deleted` are the same in `zone.js`.

### DIFF
--- a/integration/_payload-limits.json
+++ b/integration/_payload-limits.json
@@ -3,7 +3,7 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 1485,
-        "main-es2015": 143181,
+        "main-es2015": 141575,
         "polyfills-es2015": 36808
       }
     }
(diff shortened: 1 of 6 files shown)


mhevery:
[presubimt](http://test/OCL:279767274:BASE:279767320:1573497421623:9ac92edf)

mhevery (approved):
❤️ it!

Issue #32514, opened 2019-09-06
Angular IVY innerHTML directive broken, throws error in SSR


naveedahmed1:
# 🐞 bug report

### Affected Package
**innerHTML** directive

### Is this a regression?

Yes, the previous version in which this bug was not present was: Angular 8.x

### Description
When we have IVY enabled for an SSR/Universal app, and use innerHTML in component. It throw below error

`ERROR ReferenceError: document is not defined`

## 🔬 Minimal Reproduction

Just add innerHTML to any component and it should throw error. For example:

`<div [innerHTML]="'hello'"></div>`

## 🔥 Exception or Error
<pre><code>
`ERROR ReferenceError: document is not defined`
</code></pre>

## 🌍  Your Environment

**Angular Version:**
<pre><code>
Angular 9.0.0-next.5
Angular CLI: 9.0.0-next.3
</code></pre>

tuurbo:
Still an issue as of

```
Angular CLI: 9.0.0-next.13
Node: 13.0.1
OS: win32 x64
Angular: 9.0.0-next.13
... animations, cli, common, compiler, compiler-cli, core, forms
... language-service, platform-browser, platform-browser-dynamic
... platform-server, router, service-worker

Package                           Version
-----------------------------------------------------------
@angular-devkit/architect         0.900.0-next.15
@angular-devkit/build-angular     0.900.0-next.15
@angular-devkit/build-optimizer   0.900.0-next.15
@angular-devkit/build-webpack     0.900.0-next.15
@angular-devkit/core              9.0.0-next.15
@angular-devkit/schematics        9.0.0-next.13
@angular/cdk                      9.0.0-next.0
@angular/material                 9.0.0-next.0
@angular/pwa                      0.900.0-next.15
@ngtools/webpack                  9.0.0-next.15
@nguniversal/common               9.0.0-next.5
@nguniversal/express-engine       9.0.0-next.5
@schematics/angular               9.0.0-next.13
@schematics/update                0.900.0-next.13
rxjs                              6.5.3
typescript                        3.6.4
webpack                           4.41.2
```

naveedahmed1:
Any update on this?

IgorMinar:
this should be fixed by https://github.com/angular/angular/pull/33712

IgorMinar:
I'm going to close this as dupe of #33651

PR #33712, opened 2019-11-09
fix(ivy): platform agnostic document

### LINKED ISSUE #33651: i18n ICU expressions in Universal/App-Shell - ERROR ReferenceError: document is not defined
# 🐞 bug report

### Affected Package
@angular/core

### Is this a regression?

Yes, works with VE.

### Description
In the CLI @clydin enabled some further tests in the CLI and i18n. Server tests are now failing when using anything which is dependent on ICU expressions such as plural. The culprit seems to be that Ivy i18n relies on the document global rather than the standard way to get the document via the DI token. This will break in universal and app-shell because document as a global doesn't exist, and it will fail with `ERROR ReferenceError: document is not defined`

## 🔬 Minimal Reproduction
Use i18n ICU expressions in universal or app-shell scenario.

## 🔥 Exception or Error
ERROR ReferenceError: document is not defined
</code></pre>

## 🌍  Your Environment

**Angular Version:**
9.0.0-rc.0

**Anything else relevant?**
Exception is thrown from:

https://github.com/angular/angular/blob/10583f951d57bbc74cb70c49e58c6659a1b4662c/packages/core/src/render3/i18n.ts#L1183

Posted also in #fw-i18n

petebacondarwin:
Most of the use of `document` in the framework is within
the DI so they just inject the `DOCUMENT` token and are done.

Ivy is special because it does not rely upon the DI and must
get hold of the document some other way. There are a limited
number of places relevant to ivy that currently consume a global
document object.

The solution is modelled on the `LOCALE_ID` approach, which has
`getLocaleId()` and `setLocaleId()` top-level functions for ivy (see
`core/src/render3/i18n.ts`).  In the rest of Angular (i.e. using DI) the
`LOCALE_ID` token has a provider that also calls setLocaleId() to
ensure that ivy has the same value.

This commit defines `getDocument()` and `setDocument() `top-level
functions for ivy. Wherever ivy needs the global `document`, it calls
`getDocument()` instead.  Each of the platforms (e.g. Browser, Server,
WebWorker) have providers for `DOCUMENT`. In each of those providers
they also call `setDocument()` accordingly.

Fixes #33651 and #32514


packages/core/src/render3/interfaces/renderer.ts
```diff
@@ -18,6 +18,24 @@
...
+}
+
+export function getDocument(): Document {
+  if (DOCUMENT !== undefined) {
+    return DOCUMENT;
+  } else if (typeof document !== 'undefined') {
+    return document;
+  } else {
+    throw new Error(
+        'No "document" can be found. If you are running outside a browser then you must call `setDocument()` before using the renderer.');
```
AndrewKushnir: It looks like `getDocument` is not a part of public api, so the error message might be a bit misleading.

AndrewKushnir (approved):
LGTM (just one comment) 👍Thanks Pete!

packages/core/src/render3/interfaces/renderer.ts
petebacondarwin: I thought about this a bit more.

This situation should never occur in an Angular application, since we call `setDocument()` internally in the `DOCUMENT` providers.

So the only way this would happen is if someone is using ivy outside a proper Angular app, which is not supported right now. Therefore there is little point in providing an error message that suggests a solution, which is using a private API.

So I changed the code to allow an `undefined` value to fall out, which would cause an error somewhere later on in the execution. Anyone who comes across this will be able to trace the cause back to the `undefined` being returned here and so I have left an explanatory comment.

Once we do support the above scenario we will need to add `setDocument()` to the public API and maybe add a more helpful error message here.

**BONUS:** By following this approach, we do not incur the extra size increase of the core bundle by adding the error message to the code.

IgorMinar (approved):
lgtm. thanks!

packages/core/src/render3/interfaces/renderer.ts
AndrewKushnir: Thanks Pete, that makes sense!

AndrewKushnir:
- [VE Presubmit](http://test/OCL:279767157:BASE:279767208:1573497393693:fee30999)
- [Ivy Presubmit](http://test/OCL:279767157:BASE:279767350:1573497432614:ca8eaf3)

IgorMinar:
http://test/OCL:279758427:BASE:279758470:1573495060363:41680269
http://test/OCL:279758427:BASE:279758671:1573495116340:daebdb35

AndrewKushnir:
FYI, VE and Ivy presubmits are successful.

PR #33714, opened 2019-11-09
refactor(ivy): remove calls into other instructions from pipe instructions

crisbeto:
A while ago we made a pass through all instructions to make sure that none of them call directly into other instructions, however it seems like missed the `pipeBind*` since they still call into the pure functions. The result is that we have some unnecessary duplicated accesses of global state like `getLView` which are called twice in a row with nothing changing.

These changes move the common functionality into a shared file and make the pipe instructions call into them with the global state instead.

### DIFF
--- a/packages/core/src/render3/pipe.ts
+++ b/packages/core/src/render3/pipe.ts
@@ -13,8 +13,8 @@ import {getFactoryDef} from './definition';
 import {store} from './instructions/all';
 import {PipeDef, PipeDefList} from './interfaces/definition';
 import {HEADER_OFFSET, LView, TVIEW} from './interfaces/view';
-import {ɵɵpureFunction1, ɵɵpureFunction2, ɵɵpureFunction3, ɵɵpureFunction4, ɵɵpureFunctionV} from './pure_function';
-import {getBindingIndex, getLView} from './state';
+import {pureFunction1Internal, pureFunction2Internal, pureFunction3Internal, pureFunction4Internal, pureFunctionVInternal} from './pure_function';
+import {getBindingIndex, getBindingRoot, getLView} from './state';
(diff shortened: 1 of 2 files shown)


pkozlowski-opensource (approved):
LGTM

packages/core/src/render3/instructions/shared.ts
```diff
@@ -1850,3 +1851,138 @@ export function renderInitialStyling(
...
+ *
+ * @param lView LView in which the function is being executed.
+ * @param bindingRoot Binding root index.
+ * @param slotOffset the offset from binding root to the reserved slot
+ * @param pureFn Function that returns an updated value
+ * @param exp Updated expression value
+ * @param thisArg Optional calling context of pureFn
+ * @returns Updated or cached value
+ */
+export function pureFunction1Internal(
```
kara: The `shared` file is already pretty huge. Any reason we can't put these functions into `pure_function.ts`?

kara (changes requested):

packages/core/src/render3/instructions/shared.ts
crisbeto: There's no reason, I thought that the convention for functions used by multiple instructions was to put them in `shared.ts`.

crisbeto:
@kara what would be the next step here?

kara:
@crisbeto My understanding was that my feedback hadn't been addressed, so I was waiting on that

crisbeto:
The feedback has been addressed @kara.

kara (approved):
LGTM

kara:
[presubmit](http://test/OCL:288738131:BASE:288738217:1578511413943:8773335f)

Issue #33596, opened 2019-11-05
Support TypeScript 3.7


fr0:
# 🚀 feature request

### Relevant Package

 This feature request is for @angular/core

### Description

TypeScript 3.7 was released today. It includes plenty of useful new features, like the optional chaining and nullish coalescing operators. Angular should support it.

gkalpak:
Related PR: #32962

filipesilva:
If you're using the framework/CLI RC and Ivy, you might be able to just use TypeScript 3.7 directly. You'll get a peer dependency warning but in all likelihood it'll just work. Ivy is far less sensitive to TypeScript version changed than View Engine. Let me know if you try this and run into problems.

gligoran:
Any possibility this will be ready for the release of Angular 9?

filipesilva:
Current thinking at the moment is within version 9, but after 9.0 proper.

xiaoxiangmoe:
@filipesilva Will Angular 8 support ts 3.7 in the future?

filipesilva:
@xiaoxiangmoe I'm not aware of plans to backport the changes needed to version 8. Generally only security fixes and very important bugs are fixed in older versions.

xiaoxiangmoe:
Can we open `useDefineForClassFields` in Angular 9? @filipesilva

filipesilva:
You can try, but that looks like it might cause some trouble... if you run into problems please let us know.

alan-agius4:
Closing as Angular ` v9.0.0-rc.9` supports TypeScript version 3.7.

filipesilva:
Support was added in https://github.com/angular/angular/pull/33717

filipesilva:
Hey all,

Starting with `@angular/cli@9.0.0-rc.10` and `@angular/core@9.0.0-rc.9` you should be able to use TypeScript 3.7 with no problems, while 3.6 is also supported. Please let us know if something if you encounter any problems with the TS version..

michaeljota:
@xiaoxiangmoe Did `useDefineForClassFields` work for you? I'm seeing a weird issue, and I think it's because of that.

michaeljota:
I'll try to open another ticket with a reproducible repo but it seems like if you use `useDefineForClassFields` option will throw somehow if you have a service with a property defined but not initialized.

xiaoxiangmoe:
@michaeljota Sorry, I haven't tried to turn on this option because we are busy with other things.

PR #32962, opened 2019-10-02
Ts 3.7-beta support

filipesilva:
Just testing to see what breaks if https://github.com/angular/angular/pull/32946 is followed up by a 3.7-beta update.

## PR Checklist
Please check if your PR fulfills the following requirements:

- [ ] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] angular.io application / infrastructure changes
- [ ] Other... Please describe:

## What is the current behavior?

Issue Number: N/A

## Does this PR introduce a breaking change?

- [ ] Yes
- [ ] No

PR #33717, opened 2019-11-10
build: typescript 3.7 support

andrius-pra:
This PR updates TypeScript version to 3.7 while retaining compatibility with TS3.6.

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [x] Build related changes
- [ ] CI related changes

Issue Number: #33596

- [ ] Yes
- [x] No


packages/zone.js/test/common/fetch.spec.ts
```diff
@@ -150,7 +150,7 @@ describe(
 
         it('cancel fetch should invoke onCancelTask',
            ifEnvSupportsWithDone('AbortController', (done: DoneFn) => {
-             if (isSafari) {
+             if (isSafari()) {
```
alfaproject: <3

packages/compiler-cli/ngcc/src/packages/configuration.ts
```diff
@@ -266,5 +266,6 @@ function findSatisfactoryVersion(
     // So just return the first config that matches the package name.
     return configs[0];
   }
-  return configs.find(config => satisfies(version, config.versionRange)) || null;
+  const options = {includePrerelease: true};
```
andrius-pra: fix for this [test](https://github.com/angular/angular/blob/master/packages/compiler-cli/ngcc/test/integration/ngcc_spec.ts#L659-L692). 
`satisfies('9.0.0-rc.1+34.sha-0e36204.with-local-changes', '*')` returns false without includePrerelease option.

packages/core/src/render3/instructions/shared.ts
```diff
@@ -1277,7 +1277,8 @@ function saveNameToExportMap(
         exportsMap[def.exportAs[i]] = index;
       }
     }
-    if ((def as ComponentDef<any>).template) exportsMap[''] = index;
+
+    if (isComponentDef(def)) exportsMap[''] = index;
```
andrius-pra: Fix for:
```
error TS2774: This condition will always return true since the function is always defined. Did 
you mean to call it instead?
```

packages/platform-server/src/tokens.ts
```diff
@@ -32,4 +32,4 @@ export const INITIAL_CONFIG = new InjectionToken<PlatformConfig>('Server.INITIAL
  * @publicApi
  */
 export const BEFORE_APP_SERIALIZED =
-    new InjectionToken<Array<() => void>>('Server.RENDER_MODULE_HOOK');
+    new InjectionToken<Array<() => void | Promise<void>>>('Server.RENDER_MODULE_HOOK');
```
andrius-pra: Fixes error in this line [line](https://github.com/angular/angular/blob/1d429b216556911edc4b0675ece4cb9081967155/packages/platform-server/src/utils.ts#L57)
```
asyncPromises.push(callbackResult);
                   ~~~~~~~~~~~~~~
Argument of type 'void' is not assignable to parameter of type 'Promise<any>'.
```

andrius-pra:
@filipesilva, I have added you as co-author because of this PR contains changes from #32962.

packages/compiler-cli/src/diagnostics/typescript_symbols.ts
```diff
@@ -823,3 +821,12 @@ function getFromSymbolTable(symbolTable: ts.SymbolTable, key: string): ts.Symbol
 
   return symbol;
 }
+
+function getTypeArguments(
```
andrius-pra: The `typeArguments` property has been removed from the `TypeReference` interface in Typescipt 3.7. we should use the `getTypeArguments` function on `TypeChecker` instances. See [#32962](https://github.com/angular/angular/pull/32962#issuecomment-537524408) (comment)

packages/compiler-cli/src/diagnostics/typescript_symbols.ts
```diff
@@ -673,26 +675,20 @@ function getBuiltinTypeFromTs(kind: BuiltinType, context: TypeContext): ts.Type
           checker.getTypeAtLocation(setParents(<ts.Node>{kind: ts.SyntaxKind.NullKeyword}, node));
       break;
     case BuiltinType.Number:
-      const numeric = <ts.LiteralLikeNode>{
-        kind: ts.SyntaxKind.NumericLiteral,
-        text: node.getText(),
-      };
-      setParents(<any>{kind: ts.SyntaxKind.ExpressionStatement, expression: numeric}, node);
+      const numeric = <ts.NumericLiteral>{kind: ts.SyntaxKind.NumericLiteral, text: '1'};
```
gkalpak: 1. `text: '1'`? Shouldn't this be `text: node.getText()`? Or is it that the value doesn't make a difference, because we only care about the type?
2. Why not `ts.createNumericLiteral(node.getText())`?

(Same below.)

gkalpak (commented):
💯 :tada: ❤️

packages/compiler-cli/src/diagnostics/typescript_symbols.ts
andrius-pra: 1. We only need a type and this type will cached [here](https://github.com/angular/angular/blob/master/packages/compiler-cli/src/diagnostics/typescript_symbols.ts#L96-L102). I can't use `text: node.getText()` because of this error:
```
Error: Debug Failure. False expression: This failure could trigger https://github.com/Microsoft/TypeScript/issues/20809
    ...
        at checkNumericLiteralValueSize (node_modules/typescript/lib/typescript.js:65912:56)
    ...
        at Object.getTypeAtLocation (node_modules/typescript/lib/typescript.js:33327:31)
        at getBuiltinTypeFromTs (packages/compiler-cli/src/diagnostics/typescript_symbols.ts:680:22)
    ...
        at UserContext.<anonymous> (packages/language-service/test/diagnostics_spec.ts:338:26)
```

andrius-pra: 2. `checker.getTypeAtLocation(typeNode)` will always returns `errorType` when parameter `typeNode` has `NodeFlags.Synthesized` flag. All `ts.create*` methods adds this flag. So we can use `ts.create*` method only for child nodes.

aio/package.json
```diff
@@ -165,7 +165,7 @@
     "tree-kill": "^1.1.0",
     "ts-node": "^8.4.1",
     "tslint": "^5.20.0",
-    "typescript": "~3.6.4",
+    "typescript": "~3.7.2",
```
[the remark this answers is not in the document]
andrius-pra: I reverted this change.

packages/compiler-cli/src/diagnostics/typescript_symbols.ts
```diff
@@ -673,26 +675,20 @@ function getBuiltinTypeFromTs(kind: BuiltinType, context: TypeContext): ts.Type
...
+      const stringLiteral = <ts.StringLiteralLike>{kind: ts.SyntaxKind.StringLiteral, text: '1'};
+      setParents(ts.createExpressionStatement(stringLiteral), node);
+      type = checker.getTypeAtLocation(stringLiteral);
       break;
     case BuiltinType.Undefined:
       type = checker.getTypeAtLocation(setParents(
           <ts.Node><any>{
             kind: ts.SyntaxKind.VoidExpression,
-            expression: <ts.Node>{kind: ts.SyntaxKind.NumericLiteral}
+            expression: ts.createNumericLiteral('0')
```
andrius-pra: Fix for : 
```
 Message:
    TypeError: Cannot read property 'length' of undefined
  Stack:
        at checkNumericLiteralValueSize (node_modules/typescript/lib/typescript.js:65901:77)
    ...
        at Object.getTypeAtLocation (node_modules/typescript/lib/typescript.js:33327:31)
        at getBuiltinTypeFromTs (packages/compiler-cli/src/diagnostics/typescript_symbols.ts:688:22)
    ...
        at UserContext.<anonymous> (packages/language-service/test/diagnostics_spec.ts:622:32)
```

gkalpak (commented):
BTW, I can see several more `.typeArguments` accesses through-out the codebase.
Shouldn't all these occurrences be converted to use `checker.getTypeArguments()` if available?

andrius-pra:
`.typeArguments` removed only from ts.TypeReference node (microsoft/TypeScript#33693). All these occurrences are converted to use `checker.getTypeArguments()`.

IgorMinar:
hi @andrius-pra, thanks for working on this PR - I just want to set the expectations right that we won't be merging this PR until we work on 9.1. I estimate that we will come back to this PR and be able to merge it in December, but unlikely sooner than that.

I'm grateful for your help, and we could use this PR in the future to get typescript 3.7 support into 9.1, but that will require cooperation with the CLI team and extensive testing that we no longer have time in 9.0 release cycle.

tony:
@IgorMinar I wanted to chime in and say I think that message you wrote to @andrius-pra was very kind and professional! :blush: In both how it mentioned release timings / priorities / logistics but also gratitude for the contribution and time put into Andrius work.  This is a big project with a lot of moving parts, and being graceful like that was a nice thing to see.

Look forward to seeing TypeScript 3.7 support in Angular!

Svetomechc:
@IgorMinar no 3.7 in 9.0 then?

filipesilva:
Heya @andrius-pra, just wanted to let you know that I'm looking at getting this PR green and ready to be merged. I'll do that by adding some commits to your branch directly, which I can already do because of the [allow edits from maintainers checkbox](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork) on PR creation.

filipesilva:
@googlebot I consent.

filipesilva:
It's noteworthy that protractor itself doesn't seem to support TS 3.7 (https://github.com/angular/protractor/issues/5348) because of the ["Local and Imported Type Declarations Now Conflict" breaking change](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#local-and-imported-type-declarations-now-conflict), but that doesn't seem to affect anything but a few example projects that ran `tsc --project ./`. Not really sure why. 

I changed that command to compile the app and not the e2e since that was the original intention anyway. But we should still push for a Protractor release that supports TS 3.7.

tortuvshin (approved):

.circleci/config.yml
```diff
@@ -764,6 +764,16 @@ jobs:
...
           # older versions.
           command: yarn --ignore-engines --cwd ${MATERIAL_REPO_TMP_DIR} install --frozen-lockfile --non-interactive
+      - run:
+          # Install TypeScript 3.7
+          name: Installing TypeScript 3.7 and TSickle 0.38 on Material dependencies.
+          # TODO: remove this once the repo has been updated to use TypeScript 3.7.
+          # While built Angular packages support both TS 3.6 and 3.7, @angular/bazel itself only
+          # supports TS 3.7 because TSickle 0.38.0 only supports TS 3.7.
+          # It shouldn't be necessary to install tsickle, but it's referenced in material and
+          # doesn't seem to be in the dependencies. Unsure how this worked before.
```
filipesilva: https://github.com/angular/components/pull/18116 will reference it in the dependencies. It was missing.

filipesilva:
Currently blocked on making the `ci/circleci: material-unit-tests` job green.

devversion:
@googlebot I consent.

.circleci/config.yml
```diff
@@ -766,14 +766,14 @@ jobs:
...
-          name: Installing TypeScript 3.7 and TSickle 0.38 on Material dependencies.
+          name: Installing TypeScript 3.7 on Material dependencies.
           # TODO: remove this once the repo has been updated to use TypeScript 3.7.
           # While built Angular packages support both TS 3.6 and 3.7, @angular/bazel itself only
           # supports TS 3.7 because TSickle 0.38.0 only supports TS 3.7.
           # It shouldn't be necessary to install tsickle, but it's referenced in material and
           # doesn't seem to be in the dependencies. Unsure how this worked before.
           # https://github.com/angular/components/blob/636ae5c8754008bfcf4b2f3953c6b2c5a7e27a1a/tools/bazel/BUILD.bazel#L12
-          command: yarn --ignore-engines --cwd ${MATERIAL_REPO_TMP_DIR} add typescript@3.7.4 tsickle@0.38.0 --dev
+          command: yarn --ignore-engines --cwd ${MATERIAL_REPO_TMP_DIR} add typescript@3.7.4 --dev
```
filipesilva: Why is Tsickle removed here? Tsickle 0.38 only works with TS 3.7 and should be updated.

devversion: The comment above sounded like it was only needed because we didn't specify the dependency in `angular/components`. This has been fixed, so I removed it. Though, based on what you said, it sounds like we still need to manually update `tsickle`.

packages/language-service/src/typescript_symbols.ts
```diff
@@ -313,8 +314,13 @@ class TypeWrapper implements Symbol {
...
   typeArguments(): Symbol[]|undefined {
-    // TODO: use checker.getTypeArguments when TS 3.7 lands in the monorepo.
-    const typeArguments: ReadonlyArray<ts.Type> = (this.tsType as any).typeArguments;
+    if (!isReferenceType(this.tsType)) return;
+
+    const typeReference = (this.tsType as ts.TypeReference);
+    let typeArguments: ReadonlyArray<ts.Type>|undefined;
+    if (this.context.checker.getTypeArguments)
+      typeArguments = this.context.checker.getTypeArguments(typeReference);
+    typeArguments = typeReference.typeArguments;
```
filipesilva: @andrius-pra should this be in a else clause?

filipesilva: Added a commit with the else clause.

filipesilva (approved):

packages/language-service/src/typescript_symbols.ts
andrius-pra: yes, it should be in else clause.

aio/package.json
```diff
@@ -165,7 +165,7 @@
     "tree-kill": "^1.1.0",
     "ts-node": "^8.4.1",
     "tslint": "^5.20.0",
-    "typescript": "~3.6.4",
+    "typescript": "~3.7.4",
```
gkalpak: A little surprised this works (e.g. I would expect ngcc from the currently installed @angular/compiler-cli@9.0.0-rc.3 to fail), but if CI is happy, so am I :grin:

gkalpak (dismissed):
:tada:

Unless I am missing something, in order for everything to work correctly we need to account for the `typeArguments` breaking change in more places (e.g. [here][1] and [there][2] just to give a couple of examples - there are more places like this).
It is quite worrying that no tests are failing (unless, again, I am missing something) 😕 

[1]: https://github.com/angular/angular/blob/280aae6fc/packages/compiler-cli/ngcc/src/analysis/module_with_providers_analyzer.ts#L52
[2]: https://github.com/angular/angular/blob/280aae6fc/packages/compiler-cli/src/metadata/evaluator.ts#L469

aio/package.json
filipesilva: Oh jeez, I was also expecting tests to fail and didn't look this up. It's really weird that the TS compilation doesn't show errors on these....

filipesilva:
Oh geez, I also expected compilation to fail and didn't look these up actually.

Hmm I looked these up with TS 3.7 installed and they are of type `ts.NodeWithTypeArguments.typeArguments?`. Which sounds like it should have type arguments. I guess that's why, it's a different type.

andrius-pra:
The `typeArguments` property has been removed only from the [ts.TypeReference](https://github.com/microsoft/TypeScript/commit/250d5a8229e17342f36fe52545bb68140db96a2e#diff-2a68596a5960538c5c8733bc458de56eL2389-L2392) interface in typescript v3.7.0-beta. This type is only used in `typescript_symbols.ts` file.

aio/package.json
andrius-pra: I reverted this change.

IgorMinar:
I kicked off a presubmit for g3 on this PR: http://test/OCL:289739402:BASE:289739456:1579043225284:3be9296c

integration/typings_test_ts37/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "@angular/platform-browser-dynamic": "file:../../dist/packages-dist/platform-browser-dynamic",
+    "@angular/platform-server": "file:../../dist/packages-dist/platform-server",
+    "@angular/platform-webworker": "file:../../dist/packages-dist/platform-webworker",
+    "@angular/platform-webworker-dynamic": "file:../../dist/packages-dist/platform-webworker-dynamic",
+    "@angular/router": "file:../../dist/packages-dist/router",
+    "@angular/service-worker": "file:../../dist/packages-dist/service-worker",
+    "@angular/upgrade": "file:../../dist/packages-dist/upgrade",
+    "@types/jasmine": "2.5.41",
+    "rxjs": "file:../../node_modules/rxjs",
+    "typescript": "3.7.3",
```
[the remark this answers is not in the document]
IgorMinar: I'll push a fixup commit to this PR to fix this.

IgorMinar (approved):
lgtm! thanks @andrius-pra @filipesilva and @gkalpak !!!

I'll push the one fixup commit here and if the g3 presubmit is good we can land this.

IgorMinar:

tools/public_api_guard/common/common.d.ts
```diff
@@ -228,19 +229,19 @@ export declare class NgForOf<T, U extends NgIterable<T> = NgIterable<T>> impleme
 export declare class NgForOfContext<T, U extends NgIterable<T> = NgIterable<T>> {
     $implicit: T;
     count: number;
-    readonly even: boolean;
-    readonly first: boolean;
+    get even(): boolean;
```
StefanKern: Why did you switch from `readonly` to `get`?
- curious angular developer

filipesilva: We didn't, at least intentionally. 

These `.d.ts` files in `tools/public_api_guard/common` are snapshots of the types for the Angular API created by `ts-api-guardian`. The purpose of that tools is to tell us when the public TS api changed. This is useful to detect cases where we did some code change that accidentally changed the API, in which case we revert the change.

In this case however, TS started emitting different types so we had to update the snapshots. TS does not follow semver and can be expected to emit different type output between minors, so it is ok to update the snapshots.

PR #33720, opened 2019-11-10
perf(ivy): chain listener instructions

crisbeto:
Chains multiple listener instructions on a particular element into a single call which results in less generated code. Also handles listeners on templates, host listeners and synthetic host listeners.


pkozlowski-opensource (approved):
Why not, LGTM for runtime.

> results in less generated code

Did you happen to have example numbers from the apps you are working on?

packages/compiler-cli/test/compliance/r3_view_compiler_binding_spec.ts
```diff
@@ -1134,6 +1134,114 @@ describe('compiler compliance: bindings', () => {
...
+            if (rf & 1) {
+              $r3$.ɵɵcomponentHostSyntheticListener("@animation.done", function MyComponent_animation_animation_done_HostBindingHandler($event) { return ctx.done(); })("@animation.start", function MyComponent_animation_animation_start_HostBindingHandler($event) { return ctx.start(); });
+            }
+          }
+        `;
+
+      expectEmit(result.source, template, 'Incorrect template');
+    });
+
+    it('should chain multiple regular and synthetic host listeners into a single instruction',
```
petebacondarwin: Is this test description accurate? It looks to me like we get two instructions.

petebacondarwin (approved):

packages/compiler-cli/test/compliance/r3_view_compiler_binding_spec.ts
crisbeto: Good point, that's what I was trying to say.

crisbeto:
@pkozlowski-opensource I don't have numbers for these changes because I don't have access to my app this week, but the approach is similar to the other chained instructions which showed improvements in bundle size.

packages/core/src/render3/instructions/listener.ts
```diff
@@ -68,11 +69,12 @@ export function ɵɵlistener(
 */
 export function ɵɵcomponentHostSyntheticListener(
     eventName: string, listenerFn: (e?: any) => any, useCapture = false,
-    eventTargetResolver?: GlobalTargetResolver): void {
+    eventTargetResolver?: GlobalTargetResolver): TsickleIssue1009 {
```
AndrewKushnir: Quick comment: it looks like the [mentioned Tsickle issue is resolved](https://github.com/angular/tsickle/issues/1009) :)
We can probably create a followup "cleanup" PR to use real types here and in other places.

AndrewKushnir:
- [VE Presubmit](http://test/OCL:279773962:BASE:279774032:1573499056168:efa58a0f)
- [Ivy Presubmit](http://test/OCL:279773962:BASE:279774291:1573499116069:a7e79f1a)

mhevery (approved):

packages/core/src/render3/instructions/listener.ts
crisbeto: Ah nice, I can do it once I'm done adding chaining to some of the other instructions.

PR #33723, opened 2019-11-11
refactor(elements): ng-add schematics

alan-agius4:
This PR brings a couple of changes;
- Removes undeed dependencies in bazel targets such as `//packages/common` & `//packages/core`
- Removes RxJs usage
- Adds `document-register-element` to architect test targets
- Use @schematics/angular helpers
- Uses the standard `$source": "projectName"` to get the projectName, which is defined in the `schema.json`
- Use workspace writer to update the workspace config

### DIFF
--- a/packages/elements/schematics/ng-add/index.ts
+++ b/packages/elements/schematics/ng-add/index.ts
@@ -36,7 +36,7 @@ function addPolyfillDependency(): Rule {
 
 /** Adds the document-register-element.js to the polyfills file. */
 function addPolyfill(options: Schema): Rule {
-  return async(host, context) => {
+  return async(host: Tree, context: SchematicContext) => {
     const projectName = options.project;
 
     if (!projectName) {
@@ -47,11 +47,12 @@ function addPolyfill(options: Schema): Rule {
     const project = workspace.projects.get(projectName);
 
     if (!project) {
-      throw new SchematicsException(`Project is not defined in this workspace.`);
+      throw new SchematicsException(`Project ${projectName} is not defined in this workspace.`);
     }
 
     if (project.extensions['projectType'] !== 'application') {
-      throw new SchematicsException(`@angular/elements requires a project type of "application".`);
+      throw new SchematicsException(
+          `@angular/elements requires a project type of "application" but ${projectName} isn't.`);
     }
 
     const buildTarget = project.targets.get('build');
@@ -66,7 +67,7 @@ function addPolyfill(options: Schema): Rule {
 
     const content = host.read(polyfills).toString();
     if (!content.includes('document-register-element')) {
-      // Add string at the start of the file.
+      // Add string at the end of the file.
       const recorder = host.beginUpdate(polyfills);
       recorder.insertRight(content.length, `import 'document-register-element';\n`);
       host.commitUpdate(recorder);
--- a/packages/elements/schematics/ng-add/index.ts
+++ b/packages/elements/schematics/ng-add/index.ts
@@ -5,72 +5,73 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
-import {Rule, SchematicContext, Tree, chain, noop} from '@angular-devkit/schematics';
+import {Rule, SchematicContext, SchematicsException, Tree, chain, noop} from '@angular-devkit/schematics';
 import {NodePackageInstallTask} from '@angular-devkit/schematics/tasks';
+import {NodeDependencyType, addPackageJsonDependency} from '@schematics/angular/utility/dependencies';
+import {getWorkspace} from '@schematics/angular/utility/workspace';
+
 import {Schema} from './schema';
 
 export default function(options: Schema): Rule {
   return chain([
-    options && options.skipPackageJson ? noop() : addPackageJsonDependency(), addScript(options)
+    options && options.skipPackageJson ? noop() : addPolyfillDependency(),
+    addPolyfill(options),
   ]);
 }
 
 /** Adds a package.json dependency for document-register-element */
-function addPackageJsonDependency() {
+function addPolyfillDependency(): Rule {
   return (host: Tree, context: SchematicContext) => {
+    addPackageJsonDependency(host, {
+      type: NodeDependencyType.Default,
+      name: 'document-register-element',
+      version: '^1.7.2',
+    });
+    context.logger.info('Added "document-register-element" as a dependency.');
 
-    if (host.exists('package.json')) {
-      const jsonStr = host.read('package.json') !.toString('utf-8');
-      const json = JSON.parse(jsonStr);
-
-      // If there are no dependencies, create an entry for dependencies.
-      const type = 'dependencies';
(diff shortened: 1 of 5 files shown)


packages/elements/schematics/ng-add/index.ts
```diff
@@ -5,72 +5,83 @@
...
+    // Install the dependency
+    context.addTask(new NodePackageInstallTask());
   };
 }
 
 /** Adds the document-register-element.js script to the angular CLI json. */
-function addScript(options: Schema) {
-  return (host: Tree, context: SchematicContext) => {
-    const script = 'node_modules/document-register-element/build/document-register-element.js';
+function addScript(options: Schema): Rule {
```
filipesilva: `document-register-element` is a polyfill though, why is it added to the scripts array instead of the polyfills file?

filipesilva (approved):

packages/elements/schematics/ng-add/index.ts
alan-agius4: That's a very good catch.

//cc @robwormald, do you know what's the reason behind this?

packages/elements/schematics/ng-add/index.ts
```diff
@@ -5,72 +5,73 @@
...
+    options && options.skipPackageJson ? noop() : addPolyfillDependency(),
+    addPolyfill(options),
   ]);
 }
 
 /** Adds a package.json dependency for document-register-element */
-function addPackageJsonDependency() {
+function addPolyfillDependency(): Rule {
   return (host: Tree, context: SchematicContext) => {
+    addPackageJsonDependency(host, {
```
gkalpak: Does this check whether the dependency exists already?

packages/elements/schematics/ng-add/index.ts
```diff
@@ -5,72 +5,73 @@
...
+}
 
-      // If there are no dependencies, create an entry for dependencies.
-      const type = 'dependencies';
-      if (!json[type]) {
-        json[type] = {};
-      }
+/** Adds the document-register-element.js to the polyfills file. */
+function addPolyfill(options: Schema): Rule {
+  return async(host, context) => {
```
gkalpak: Why no parameter types? Can TypeScript infer them? (If so, how?)

packages/elements/schematics/ng-add/index.ts
```diff
@@ -5,72 +5,73 @@
...
-      // Write the JSON back to package.json
-      host.overwrite('package.json', JSON.stringify(json, null, 2));
-      context.logger.log('info', 'Added `document-register-element` as a dependency.');
+    const workspace = await getWorkspace(host);
+    const project = workspace.projects.get(projectName);
 
-      // Install the dependency
-      context.addTask(new NodePackageInstallTask());
+    if (!project) {
+      throw new SchematicsException(`Project is not defined in this workspace.`);
```
gkalpak: Would it be worth including the name of the project that was not found?

packages/elements/schematics/ng-add/index.ts
```diff
@@ -5,72 +5,73 @@
...
-    return host;
-  };
-}
-
-/** Adds the document-register-element.js script to the angular CLI json. */
-function addScript(options: Schema) {
-  return (host: Tree, context: SchematicContext) => {
-    const script = 'node_modules/document-register-element/build/document-register-element.js';
+    if (project.extensions['projectType'] !== 'application') {
+      throw new SchematicsException(`@angular/elements requires a project type of "application".`);
```
gkalpak: Again, I think it might be helpful to include the project name.

packages/elements/schematics/ng-add/index.ts
```diff
@@ -5,72 +5,73 @@
...
-          'warn', 'Failed to add the polyfill document-register-element.js to scripts');
+    const {polyfills} = buildTarget.options;
+    if (typeof polyfills !== 'string') {
+      throw new SchematicsException(`polyfills for ${projectName} build target is not a string.`);
     }
 
-    context.logger.log('info', 'Added document-register-element.js polyfill to scripts');
+    const content = host.read(polyfills).toString();
+    if (!content.includes('document-register-element')) {
+      // Add string at the start of the file.
```
gkalpak: Why at the start?

packages/elements/schematics/ng-add/index_spec.ts
```diff
@@ -40,20 +35,29 @@ describe('Elements Schematics', () => {
...
+  });
+
+  it('should add polyfill as a dependency in package.json', async() => {
+    const tree =
+        await schematicRunner.runSchematicAsync('ng-add', defaultOptions, appTree).toPromise();
+    const pkgJsonText = tree.readContent('/package.json');
+    const pkgJson = JSON.parse(pkgJsonText);
+    const {dependencies} = pkgJson;
+    expect(dependencies['document-register-element']).toBeDefined();
   });
```
gkalpak: It would be cool to have tests that verify what happens when `document-register-element` is included in `package.json` and/or polyfills.

packages/elements/schematics/ng-add/index.ts
alan-agius4: Yes

packages/elements/schematics/ng-add/index.ts
alan-agius4: Yeah from the return type.

packages/elements/schematics/ng-add/index.ts
alan-agius4: typo

start should be end

packages/elements/schematics/ng-add/index_spec.ts
alan-agius4: Like an E2E test you mean?

packages/elements/schematics/ng-add/index.ts
gkalpak: Doh! 🤦‍♂

packages/elements/schematics/ng-add/index_spec.ts
gkalpak: No, I mean adding a couple of tests here. For example, adding the dependency in `package.json` (using a different version), then running the schematic and finally verifying that it was not overwritten.

alan-agius4: Why should be add tests which is provided by another package? 

I mean if for instance you use a method from `@angular/core` you don’t add unit tests that it’s working as “excepted” during its consumption.

With the same argument I should add unit tests for other schematic utils. So why is this any different?

gkalpak: Fair enough. I wrote this before I was sure whether `addPackageJsonDependency()` checks for that or not.
What about not adding it is polyfills if it's already present?

alan-agius4: That can be done, though better the safe than sorry :)

alan-agius4: Following the offline convo, will keep as is

gkalpak (approved):

PR #33726, opened 2019-11-11
docs: update update guide to use `ng update` instead of `npm i`

### LINKED ISSUE #16021: Not saving update to latest CLI version 8 makes repeat attempts error-prone
This issue impacts 9.0.0-rc.0 and happened while updating https://github.com/vmware/clarity.

In https://next.angular.io/guide/updating-to-version-9 the suggested update procedure is to follow these commands:
```
npm install --no-save @angular/cli@^8.3.15
ng update @angular/cli @angular/core --next
```

When the update procedure fails, a user will need to manually revert their repository to before the update (as described in https://github.com/angular/angular-cli/issues/16020) and attempt to update again.

If the user forgets to run `npm install --no-save @angular/cli@^8.3.15`, they will end up an incorrect update.

It would be better to instruct users to save and commit their update to `@angular/cli@^8.3.15` before proceeding in order to reduce mistakes in subsequent update attempts.

cc @StephenFluin @mgechev

alan-agius4:
For more context see: https://github.com/angular/angular-cli/issues/16021

Closes: https://github.com/angular/angular-cli/issues/16021

### DIFF
--- a/aio/content/guide/updating-to-version-9.md
+++ b/aio/content/guide/updating-to-version-9.md
@@ -7,12 +7,14 @@ This guide contains everything you need to know about updating to the next Angul
 If your application uses the CLI, you can update to version 9 automatically with the help of the `ng update` script:
 
 ```
-npm install --no-save @angular/cli@^8.3.15
+ng update @angular/core@8 @angular/cli@8
+git add .
+git commit --all -m "build: update Angular packages to latest 8.x version"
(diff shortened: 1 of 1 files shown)


filipesilva (approved):
Note: the update site itself also needs to reflect these changes.

alan-agius4:
Update site changes: https://github.com/StephenFluin/angular-update-guide/pull/40

PR #33732, opened 2019-11-11
fix(core): support `ngInjectableDef` on types with inherited `ɵprov`

petebacondarwin:
The `ngInjectableDef` property was renamed to `ɵprov`, but core must
still support both because there are published libraries that use the
older term.

We are only interested in such properties that are defined directly on
the type being injected, not on base classes. So there is a check that
the defintion is specifically for the given type.

Previously if you tried to inject a class that had `ngInjectableDef` but
also inherited `ɵprov` then the check would fail on the `ɵprov` property
and never even try the `ngInjectableDef` property resulting in a failed
injection.

This commit fixes this by attempting to find each of the properties
independently.

Fixes https://github.com/angular/ngcc-validation/pull/526

### DIFF
--- a/packages/compiler-cli/integrationtest/bazel/injectable_def/app/test/app_spec.ts
+++ b/packages/compiler-cli/integrationtest/bazel/injectable_def/app/test/app_spec.ts
@@ -160,6 +160,32 @@ describe('ngInjectableDef Bazel Integration', () => {
     expect(() => TestBed.inject(ChildService).value).toThrowError(/ChildService/);
   });
 
+  it('uses legacy `ngInjectable` property even if it inherits from a class that has `ɵprov` property',
+     () => {
+       @Injectable({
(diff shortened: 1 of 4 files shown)


alxhub (approved):

alxhub:
[Presubmit](http://test/OCL:279808849:BASE:279808885:1573509019398:60e24b71)
[Ivy presubmit](http://test/OCL:279808849:BASE:279809067:1573509071672:96704ee9)

kara (approved):
LGTM

kara:
[VE presubmit](https://test.corp.google.com/ui#id=OCL:279996230:BASE:279996311:1573582278006:ddd15f8f)

[Ivy presubmit](http://test/OCL:279996230:BASE:279997089:1573582478440:f213e0a7)

PR #33739, opened 2019-11-11
fix(ivy): throw a better error when DI can't inject a ctor param

alxhub:
Occasionally a factory function needs to be generated for an "invalid"
constructor (one with parameters types which aren't injectable). Typically
this happens in JIT mode where understanding of parameters cannot be done in
the same "up-front" way that the AOT compiler can.

This commit changes the JIT compiler to generate a new `invalidFactoryDep`
call for each invalid parameter. This instruction will error at runtime if
called, indicating both the index of the invalid parameter as well as (via
the stack trace) the factory function which was generated for the type being
constructed.


packages/core/test/render3/ivy/jit_spec.ts
```diff
@@ -342,6 +342,32 @@ ivyEnabled && describe('render3 jit', () => {
...
+      class Service extends Base {
+      }
+
+      const ServiceAny = Service as any;
+
+      expect(ServiceAny.ɵprov).toBeDefined();
+      expect(ServiceAny.ɵprov.providedIn).toBe('root');
+      expect(() => ɵɵinject(Service))
+          .toThrowError(
+              /constructor is not compatible with Angular Dependency Injection because its dependency for its parameter 2 is invalid/);
```
kara: It seems unexpected to me that we're not using zero-based indices here. Shouldn't it be "parameter at index 1"?

```suggestion
              /constructor is not compatible with Angular Dependency Injection because its dependency  at index 1 is invalid/);
```

packages/core/test/render3/ivy/jit_spec.ts
```diff
@@ -342,6 +342,32 @@ ivyEnabled && describe('render3 jit', () => {
...
+    });
+  });
```
kara: Can we add a test for the missing directive decorator case?

kara (changes requested):
Thanks for doing this! Some nits

packages/core/src/di/injector_compatibility.ts
```diff
@@ -114,6 +114,28 @@ export function ɵɵinject<T>(token: Type<T>| InjectionToken<T>, flags = InjectF
...
+ * off, saving bytes of generated code while still providing a good experience in dev mode.
+ *
+ * The name of the class is not mentioned here, but will be in the generated factory function name
+ * and thus in the stack trace.
+ *
+ * @codeGenApi
+ */
+export function ɵɵinvalidFactoryDep(index: number): never {
+  const msg = ngDevMode ?
+      `This constructor is not compatible with Angular Dependency Injection because its dependency for its parameter at index ${index} is invalid.
```
kara: Nit: I'm still having a bit of trouble parsing `its dependency for its parameter`. How about:

```suggestion
      `This constructor is not compatible with Angular Dependency Injection because its dependency at index ${index} of the parameter list is invalid. 
```

alxhub: 👍

kara (approved):
LGTM

kara:

alxhub:
[Presubmit](http://test/OCL:281589058:BASE:281589109:1574285519868:e72dbf99)
[Ivy Presubmit](http://test/OCL:281589058:BASE:281589357:1574285592986:98b992f8)

matsko:
presubmit = https://tap.corp.google.com/ui#id=OCL:281828042:BASE:281828085:1574374000808:cacdfdd

ivy-presubmit = https://test.corp.google.com/ui#id=OCL:281828042:BASE:281828217:1574374035451:c620fa44

mhevery:
Needs rebase and fails g3

```
packages/core/src/di/injector_compatibility.closure.js:145: ERROR - [JSC_UNDEFINED_VARIABLE] variable ngDevMode is undeclared
    const msg = ngDevMode ?
                ^^^^^^^^^

1 error(s), 0 warning(s)
```

alxhub:
[Presubmit](http://test/OCL:284282047:BASE:284282095:1575677274521:5de79d48)
[Ivy Presubmit](http://test/OCL:284282047:BASE:284282095:1575677274493:2f644fe1)

packages/core/test/render3/ivy/jit_spec.ts
alxhub: Done!

PR #33747, opened 2019-11-11
build: set up all packages to publish via wombot proxy

josephperrott:
Set `publishConfig` attribute for all `package.json`s used for release.  Sets the registry to use Google's wombot proxy for publishing.


gkalpak (approved):

IgorMinar (commented):
doesn't this also require an update to the release instructions?

I believe you need to authenticate against the proxy.

IgorMinar:
nevermind, I hear that the instructions have been updated already and @kara is in the loop on this change.

PR #33762, opened 2019-11-12
docs: initial Ivy i18n guide updates

clydin:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

### DIFF
--- a/aio/content/guide/i18n.md
+++ b/aio/content/guide/i18n.md
@@ -140,14 +140,8 @@ The i18n template translation process has four phases:
 4. Merge the completed translation file into the app. To do this, use the Angular CLI `build` command to compile the app, choosing a [locale-specific configuration](#merge), or specifying the following command options.
 
     * `--i18nFile`=*path to the translation file*
-    * `--i18nFormat`=*format of the translation file*
+    * `--i18nFormat`=*format of the translation file (only required if using an earlier version of Angular, or if you [disable Ivy](guide/ivy#opting-out-of-ivy-in-version-9))*
     * `--i18nLocale`= *locale id*
-
-<div class="alert is-helpful">
-
-  The `--i18nFormat` option is only required if you are using an earlier version of Angular, or if you [disable Ivy](guide/ivy#opting-out-of-ivy-in-version-9).
-
-</div>
 
 The command replaces the original messages with translated text, and generates a new version of the app in the target language.
 
@@ -654,14 +648,8 @@ detection or URL parameters.
 To instruct the AOT compiler to use your translation configuration, set the three "i18n" build configuration options in your CLI configuration file, `angular.json`.
 
 * `i18nFile`: the path to the translation file.
-* `i18nFormat`: the format of the translation file.
+* `i18nFormat`: the format of the translation file (only required if using an earlier version of Angular, or if you [disable Ivy](guide/ivy#opting-out-of-ivy-in-version-9)).
 * `i18nLocale`: the locale id.
-
-<div class="alert is-helpful">
-
-  The `--i18nFormat` option is only required if you are using an earlier version of Angular, or if you [disable Ivy](guide/ivy#opting-out-of-ivy-in-version-9).
-
-</div>
 
 You should also direct the output to a locale-specific folder to keep it separate from other locale versions of your app, by setting the `outputPath` configuration option.
 
--- a/aio/content/guide/i18n.md
+++ b/aio/content/guide/i18n.md
@@ -76,12 +76,12 @@ time of writing:
 ## i18n pipes
 
 Angular pipes can help you with internationalization: the `DatePipe`, `CurrencyPipe`, `DecimalPipe`
-and `PercentPipe` use locale data to format data based on the `LOCALE_ID`.
+and `PercentPipe` use locale data to format data based on the `LOCALE_ID` dependency injection token.
 
-By default, Angular only contains locale data for `en-US`. If you set the value of
-`LOCALE_ID` to another locale, you must import locale data for that new locale.
-The CLI imports the locale data for you when you use the parameter `--configuration` with `ng serve` and
-`ng build`.
+By default, Angular only contains locale data for `en-US`.
+The CLI automatically includes the locale data and sets the `LOCALE_ID` value for you when you use the
+parameter `--i18nLocale` or option `i18nLocale` with `ng serve` and `ng build`.
+However, if you manually set the value of `LOCALE_ID` to another locale, you must import locale data for that new locale.
 
 If you want to import locale data for other languages, you can do it manually:
 
@@ -98,6 +98,7 @@ locale id "fr-FR" instead of "fr".
 The files in `@angular/common/locales` contain most of the locale data that you
 need, but some advanced formatting options might only be available in the extra dataset that you can
 import from `@angular/common/locales/extra`. An error message informs you when this is the case.
+Ivy will automatically include the extra locale data if the locale was configured via `i18nLocale`.
 
 <code-example path="i18n/doc-files/app.locale_data_extra.ts" region="import-locale-extra" header="src/app/app.module.ts"></code-example>
 
@@ -131,6 +132,13 @@ The i18n template translation process has four phases:
     * `--i18nFile`=*path to the translation file*
     * `--i18nFormat`=*format of the translation file*
     * `--i18nLocale`= *locale id*
+
+<div class="alert is-helpful">
+
+  The `--i18nFormat` option is only required when using View Engine.
+  Ivy automatically determines the format based on the file specified by `--i18nFile`.
+
+</div>
 
 The command replaces the original messages with translated text, and generates a new version of the app in the target language.
 
@@ -383,12 +391,8 @@ By default, the command creates a file named `messages.xlf` in your project's ro
 
 <div class="alert is-helpful">
 
-If you don't use the CLI, you have two options:
-* You can use the `ng-xi18n` tool directly from the `@angular/compiler-cli` package.
+If you don't use the CLI, you can use the `ng-xi18n` tool directly from the `@angular/compiler-cli` package.
 For more information, see the [`ng xi18n` command documentation](cli/xi18n).
-* You can use the CLI Webpack plugin `AngularCompilerPlugin` from the `@ngtools/webpack` package.
-Set the parameters `i18nOutFile` and `i18nOutFormat` to trigger the extraction.
-For more information, see the [Angular Ahead-of-Time Webpack Plugin documentation](https://github.com/angular/angular-cli/tree/master/packages/ngtools/webpack).
 
 </div>
 
@@ -412,13 +416,13 @@ The command can read and write files in three translation formats:
 * <a href="http://cldr.unicode.org/development/development-process/design-proposals/xmb" >XML Message
 Bundle (XMB)</a>
 
-You can specify the translation format explicitly with the `--i18nFormat` command option, as illustrated in
+You can specify the translation format explicitly with the `--format` command option, as illustrated in
 these example commands:
 
 <code-example language="sh" class="code-shell">
-ng xi18n  --i18n-format=xlf
-ng xi18n  --i18n-format=xlf2
-ng xi18n  --i18n-format=xmb
+ng xi18n  --format=xlf
+ng xi18n  --format=xlf2
+ng xi18n  --format=xmb
 </code-example>
 
 The sample in this guide uses the default XLIFF 1.2 format.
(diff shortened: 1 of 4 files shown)


aio/content/guide/i18n.md
```diff
@@ -736,6 +749,13 @@ The same configuration options can also be provided through the CLI with your ex
 {@a merge-jit}
 ### Merge with the JIT compiler
 
+<div class="alert is-important">
+
+Ivy does not support i18n translation with JIT mode.
+AOT is the default within Ivy for both production and development configurations.
```
jbogarthyde: I'm not sure what you mean by "Ivy does not support translation with JIT mode" -- do you just mean that AOT is the default? 

perhaps change the order? 

In [Ivy](guide/ivy), AOT is the default compilation mode for both production and development configurations. If you choose to use [JIT compilation](guide/glossary#jit) you must take additional steps to support translation.

jbogarthyde (changes requested):
Some suggestions, and a little confusion

aio/content/guide/i18n.md
```diff
@@ -132,6 +133,12 @@ The i18n template translation process has four phases:
     * `--i18nFormat`=*format of the translation file*
     * `--i18nLocale`= *locale id*
 
+<div class="alert is-helpful">
+
+  The `--i18nFormat` option is only required if you are using an earlier version of Angular, or if you [disable Ivy](guide/ivy#opting-out-of-ivy-in-version-9)
+
```
jbogarthyde: add period

aio/content/guide/i18n.md
```diff
@@ -653,6 +655,12 @@ To instruct the AOT compiler to use your translation configuration, set the thre
 * `i18nFormat`: the format of the translation file.
 * `i18nLocale`: the locale id.
 
+<div class="alert is-helpful">
+
+  The `--i18nFormat` option is only required if you are using an earlier version of Angular, or if you [disable Ivy](guide/ivy#opting-out-of-ivy-in-version-9)
+
```
jbogarthyde: add period

aio/content/guide/i18n.md
jbogarthyde: I still don't understand -- what exactly is not supported in Ivy w/JIT? 
What is it that you have to do by hand, as instructed below?

jbogarthyde (changes requested):
minor edits, but still need clarification about exactly what Ivy does not do in JIT mode, that you need to do by hand.

aio/content/guide/i18n.md
clydin: This section describes a process that is unsupported when using Ivy.  If using Ivy and you would like to your application translated, you currently must use AOT.

clydin: This section could also possibly be removed.  It relies on practices that are not recommended and potentially error prone.  The use of a webpack loader directly within the code being the main concern (the `raw-loader` part of `const translations = require('raw-loader!./locale/messages.fr.xlf').default;`).

jbogarthyde: I think you should remove it -- and move this note to line 647, after the first paragraph in "Merge with AOT compiler"
Also, put the link on  [Ivy](guide/ivy).

aio/content/guide/ivy.md
```diff
@@ -15,6 +15,7 @@ Learn more about the [Compiler](https://www.youtube.com/watch?v=anphffaCZrQ) and
 
 AOT compilation with Ivy is faster and should be used by default.
 In the `angular.json` workspace configuration file, set the default build options for your project to always use AOT compilation.
+When using i18n with Ivy, [translation merging](guide/i18n#merge) also requires the use of AOT compilation.
```
jbogarthyde: Perhaps, instead of "i18n", use *localization* as a less jargon-y term  --
 "Localizing an application with Ivy requires AOT compilation. The JIT compiler cannot merge translations."

aio/content/guide/i18n.md
```diff
@@ -630,18 +642,14 @@ translation file.
...
 
-How you provide this information depends upon whether you compile with
-the JIT compiler or the AOT compiler.
-
-  * With [AOT](guide/i18n#merge-aot), you pass the information as configuration settings.
-  * With [JIT](guide/i18n#merge-jit), you provide the information at bootstrap time.
+The [AOT compiler](guide/glossary#aot) is part of a build process that produces a small, fast,
+ready-to-run application package, with Ivy it is now used for both production and development.
 
+<div class="alert is-important">
```
jbogarthyde: There are two reference to `i18n#merge-aot` - you need to restore the anchor tag.

aio/content/guide/ivy.md
clydin: Went with `Application internationalization` as it appears to be mentioned in other guides in this form.

jbogarthyde: That works.

jbogarthyde (approved):
Thanks. LGTM.

aio/content/guide/i18n.md
```diff
@@ -733,34 +750,6 @@ The same configuration options can also be provided through the CLI with your ex
   ng build --prod --i18n-file src/locale/messages.fr.xlf --i18n-format xlf --i18n-locale fr
 </code-example>
 
-{@a merge-jit}
-### Merge with the JIT compiler
```
filipesilva: I imagine VE users still need to do this, right?

filipesilva (approved):
Just a couple of questions regarding VE usage, which we still support.

Maybe we need a big warning on top saying "if you're using VE check out the version 8 site instead"?

filipesilva:
Also, is there any documentation about what changed from VE to Ivy?

aio/content/guide/i18n.md
```diff
@@ -412,13 +425,13 @@ The command can read and write files in three translation formats:
...
 
-You can specify the translation format explicitly with the `--i18nFormat` command option, as illustrated in
+You can specify the translation format explicitly with the `--format` command option, as illustrated in
 these example commands:
 
 <code-example language="sh" class="code-shell">
-ng xi18n  --i18n-format=xlf
-ng xi18n  --i18n-format=xlf2
-ng xi18n  --i18n-format=xmb
+ng xi18n  --format=xlf
```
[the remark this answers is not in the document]
clydin: Yes.  As it is deprecated, any user attempting to use it will receive a warning but it will still work.

aio/content/guide/i18n.md
jbogarthyde: We can refer them to the v8 doc in the note about about Ivy not supporting JIT localization.

clydin: They don't need to but yes they potentially could.

I was thinking the same thing about adding it to the note.  I'll update it.

aio/content/guide/i18n.md
```diff
@@ -25,6 +25,16 @@ After you have set up your app to use i18n, the CLI can help you with the follow
...
 * Creating multiple language versions of your app.
 
+{@a setting-up-cli}
+## Setting up localization with the Angular CLI
+
+The first step to setting up localization when using the Angular CLI is to add the `@angular/localize` package to the project.  This will install the package within your project as well as
+initialize the project to take advantage of Angular's localization features.
+
+<code-example language="sh" class="code-shell">
+  ng add @angular/localize
```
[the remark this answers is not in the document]
clydin: VE will continue to function regardless of the presence of the localize runtime _polyfill_ within `polyfills.ts`.

I added some instructions to the Ivy opt out guide in case a user wishes to remove the runtime element.

jbogarthyde (changes requested):
One more change to address Filipe's concerns.

aio/content/guide/i18n.md
clydin: Link added

jbogarthyde (approved):
LGTM

filipesilva (approved):

aio/content/guide/i18n.md
```diff
@@ -454,8 +467,7 @@ file. This information is not used by Angular, but external translation tools ma
 {@a translate}
 ## Translate the source text
 
-The `ng xi18n` command generates a translation source file named `messages.xlf` in the project `src`
-folder.
+By default, the `ng xi18n` command generates a translation source file named `messages.xlf` in the project `src` folder.
```
petebacondarwin: Outside of the remit of this PR but I thought I would raise it here anyway.

I would really really like to change the default output of `ng xi18n` to  XLIFF 2 (currently it is XLIFF 1.2). This is because there are two digest functions for computing message ids, XMB and XLIFF 2 use the one that we are using internally going forward for all `$localize` messages, while XLIFF 1.2 uses the outdated one that we want to get rid of.

At the moment we have to hack the compiler to insert these legacy message ids every time the developer uses XLIFF 1.2 format. In the future we will switch this off and we will need to migrate everyone over to the new digest for XLIFF 1.2. If they already using XMB or XLIFF 2, then they will not need to be migrated and we will have fewer migrations to support.

aio/content/guide/i18n.md
```diff
@@ -630,18 +642,17 @@ translation file.
...
 format that Angular understands, such as `.xtb`.
 
-How you provide this information depends upon whether you compile with
-the JIT compiler or the AOT compiler.
-
-  * With [AOT](guide/i18n#merge-aot), you pass the information as configuration settings.
-  * With [JIT](guide/i18n#merge-jit), you provide the information at bootstrap time.
+{@a merge-aot}
+The [AOT compiler](guide/glossary#aot) is part of a build process that produces a small, fast,
+ready-to-run application package, with Ivy it is now used for both production and development.
```
petebacondarwin:
```suggestion
ready-to-run application package. With Ivy it is now used for both production and development.
```

petebacondarwin (commented):
Thanks for working on this @clydin and @jbogarthyde .

A few minor comments. It would be nice if the inline code blocks could actually reference real regions from our example code.

A significant area that appears to be missing though is the new configuration of the `angular.json` for i18n.
For example:

```
      "i18n": {
        "sourceLocale": "en-US",
        "locales": {
          "fr": "src/locale/messages.fr.xlf",
          "de": "src/locale/messages.de.xlf"
        }
      },
```

Is this not publicly part of the v9 release? Or is it an oversight?

aio/content/guide/i18n.md
```diff
@@ -383,12 +400,8 @@ By default, the command creates a file named `messages.xlf` in your project's ro
 
 <div class="alert is-helpful">
 
-If you don't use the CLI, you have two options:
-* You can use the `ng-xi18n` tool directly from the `@angular/compiler-cli` package.
+If you don't use the CLI, you can use the `ng-xi18n` tool directly from the `@angular/compiler-cli` package.
```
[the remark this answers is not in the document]
clydin: Good question.  I’ll check but from a usability perspective for 9.x, I think it should.  It’s essentially broken by default otherwise.

aio/content/guide/i18n.md
clydin: I’ll add an item to track changing the default so it can be discussed with the team.  I don’t think any changes related to it should be done in this PR though.

aio/content/guide/i18n.md
```diff
@@ -126,12 +137,18 @@ The i18n template translation process has four phases:
 
 3. Edit the generated translation file: Translate the extracted text into the target language.
 
-4. Merge the completed translation file into the app. To do this, use the Angular CLI `build` command to compile the app, choosing a [locale-specific configuration](#merge-aot), or specifying the following command options.
+4. Merge the completed translation file into the app. To do this, use the Angular CLI `build` command to compile the app, choosing a [locale-specific configuration](#merge), or specifying the following command options.
 
     * `--i18nFile`=*path to the translation file*
     * `--i18nFormat`=*format of the translation file*
```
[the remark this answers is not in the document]
clydin: Personally, I prefer alphabetic order when possible.  Is the re-ordering a common practice in other areas of the documentation?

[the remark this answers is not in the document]
jbogarthyde: An alternative to changing the order is to include the note in the description: 

 * `--i18nFormat`=*format of the translation file (only required if using an earlier version of Angular, or if you [disable Ivy](guide/ivy#opting-out-of-ivy-in-version-9).*

aio/content/guide/i18n.md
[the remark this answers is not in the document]
clydin: Removed this section as I was unable to make the tool work as described. The ivy compilation issue is indeed present.  `--outFile` in particular was not functional.  We can add this back in the future though.

aio/content/guide/i18n.md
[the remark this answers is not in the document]
clydin: Updated to use the inline description suggestion.

petebacondarwin (approved):
LGTM

Issue #33650, opened 2019-11-07
[ngStyle] doesn't remove previous attribute (ivy)


novathore:
# 🐞 bug report

### Is this a regression?
yes, without ivy engine work as expected

### Description
Ivy doesn't delete previous style attribute at dom via [ngStyle]

## 🔬 Minimal Reproduction
https://github.com/novathore/ivy-ng-style
https://github.com/novathore/ivy-ng-style/blob/master/src/app/app.component.ts

## 🌍  Your Environment

**Angular Version:**
<pre><code>

     _                      _                 ____ _     ___
    / \   _ __   __ _ _   _| | __ _ _ __     / ___| |   |_ _|
   / △ \ | '_ \ / _` | | | | |/ _` | '__|   | |   | |    | |
  / ___ \| | | | (_| | |_| | | (_| | |      | |___| |___ | |
 /_/   \_\_| |_|\__, |\__,_|_|\__,_|_|       \____|_____|___|
                |___/
    

Angular CLI: 9.0.0-rc.0
Node: 11.14.0
OS: darwin x64
Angular: 9.0.0-rc.1
... animations, common, compiler, compiler-cli, core, forms
... language-service, platform-browser, platform-browser-dynamic
... router

Package                           Version
-----------------------------------------------------------
@angular-devkit/architect         0.900.0-rc.0
@angular-devkit/build-angular     0.900.0-rc.0
@angular-devkit/build-optimizer   0.900.0-rc.0
@angular-devkit/build-webpack     0.900.0-rc.0
@angular-devkit/core              9.0.0-rc.0
@angular-devkit/schematics        9.0.0-rc.0
@angular/cli                      9.0.0-rc.0
@ngtools/webpack                  9.0.0-rc.0
@schematics/angular               9.0.0-rc.0
@schematics/update                0.900.0-rc.0
rxjs                              6.5.3
typescript                        3.6.4
webpack                           4.41.2
</code></pre>

Ralny:
> #  bug report
> ### Is this a regression?
> yes, without ivy engine work as expected
> 
> ### Description
> Ivy doesn't delete previous style attribute at dom via [ngStyle]
> 
> ##  Minimal Reproduction
> https://github.com/novathore/ivy-ng-style
> https://github.com/novathore/ivy-ng-style/blob/master/src/app/app.component.ts
> 
> ##  Your Environment
> **Angular Version:**
> 
> ```
> 
> 
>      _                      _                 ____ _     ___
[...]

PR #33780, opened 2019-11-13
fix(ivy): reset style property value defined using [style.prop.px]

AndrewKushnir:
Prior to this change, setting style prop value to undefined or empty string would not result in resetting prop value in case the style prop is defined using [style.prop.px] syntax. The problem is that the check for empty value (and thus reseting the value) considered successful only in case of `null` value. This commit updates the check to use `isStylingValueDefined` function that also checks for undefined and empty string.

This PR resolves FW-1676.

## PR Type

- [x] Bugfix

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/test/acceptance/styling_spec.ts
+++ b/packages/core/test/acceptance/styling_spec.ts
@@ -2312,32 +2312,29 @@ describe('styling', () => {
 
   it('should allow to reset style property value defined using [style.prop.px] binding', () => {
     @Component({
-      template: `
-        <div [style.left.px]="left"></div>
-        <button (click)="setLeftTo(undefined)"></button>
-        <button (click)="setLeftTo(null)"></button>
-        <button (click)="setLeftTo('')"></button>
-        <button (click)="setLeftTo('20')"></button>
-      `,
+      template: '<div [style.left.px]="left"></div>',
     })
     class MyComp {
-      left = '10';  // initial value
-      setLeftTo(value: any) { this.left = value; }
+      left = '';
     }
 
     TestBed.configureTestingModule({declarations: [MyComp]});
     const fixture = TestBed.createComponent(MyComp);
     fixture.detectChanges();
 
+    const checks = [
+      ['15', '15px'],
+      [undefined, ''],
+      [null, ''],
+      ['', ''],
+      ['0', '0px'],
+    ];
     const div = fixture.nativeElement.querySelector('div');
-
-    const expected = ['', '', '', '20px'];
-    const buttons = fixture.nativeElement.querySelectorAll('button') !;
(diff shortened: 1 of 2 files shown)


packages/core/test/acceptance/styling_spec.ts
```diff
@@ -2310,6 +2310,37 @@ describe('styling', () => {
...
+    TestBed.configureTestingModule({declarations: [MyComp]});
+    const fixture = TestBed.createComponent(MyComp);
+    fixture.detectChanges();
+
+    const div = fixture.nativeElement.querySelector('div');
+
+    const expected = ['', '', '', '20px'];
+    const buttons = fixture.nativeElement.querySelectorAll('button') !;
+    expect(div.style.left).toBe('10px');  // initial value
+    Array.from(buttons).forEach((button: any, index: number) => {
```
pkozlowski-opensource: do we need all those `<button>` elements in this test? Given that we call `fixture.detectChanges()` on each and every `click`, I would rather:
* remove all the `<button>` elements from a template
* set the `left` on the fixture.ComponentInstance directly.

pkozlowski-opensource (changes requested):
Thnx for the fix @AndrewKushnir - the fix itself LGTM but I think that we can simplify a test (see in-line comment)

pkozlowski-opensource:
@AndrewKushnir wouldn't your change also solve https://github.com/angular/angular/issues/33650 ? 
Maybe we should add a test for the `[ngStyle]` case as well?

packages/core/test/acceptance/styling_spec.ts
matsko: Yes I agree. Just get the component instance and change the class members.

matsko (changes requested):

AndrewKushnir:
Thanks for review @matsko and @pkozlowski-opensource!

I was modeling the test after the use-case where the problem was originally found, but I do agree that it's a bit too complicated and we can make it simpler :) I've updated the test as you proposed, could you please have a look?

Thank you.

pkozlowski-opensource (approved):
LGTM

AndrewKushnir:
- [VE Presubmit](http://test/OCL:280233293:BASE:280233347:1573671368527:83d9075)
- [Ivy Presubmit](http://test/OCL:280233293:BASE:280233496:1573671403352:bed7d951)

AndrewKushnir:
@pkozlowski-opensource sorry I missed your question related to #33650. I just tested the use-case with `[ngStyle]` and the changes in this PR and the problem still exists (so it requires additional investigation and fixed in a separate PR). Thank you.

matsko (approved):

PR #33786, opened 2019-11-13
build: delete rxjs d.ts files referencing rxjs-compat

IgorMinar:
In order to speed up bazel build performance delete all rxjs d.ts files
that reference rxjs-compat.

For all ts_library and ng_module rules Bazel generates tsconfig.json file
that explicitly lists all d.ts files found in required npm package.

In case of rxjs, this means that tsconfig contains all d.ts files that
reference rxjs-compat package, which is an interop/backwards compatibility
package not installed in angular/angular repo.

But because tsconfig contains these d.ts files, tsc will try to resolve
them and silently fail. All these lookups are quite expensive and not
cached. This causes significant slowdown of the build under bazel.

This change removes all of these problematic rxjs d.ts files via an npm
postinstall hook. This is not ideal because it solves the problem only
for our repo, but it's a good start.

Build perf improvements per target:
//packages/core/src/reflect:reflect    5sec =>  3 sec
//packages/core:core                  17sec => 12 sec
//packages/router:router              30sec =>  8 sec


IgorMinar:
hmm.. CI fails in some weird ways

josephperrott (approved):
LGTM

PR #33788, opened 2019-11-13
fix(bazel): update to tsickle 0.37.1 to fix peerDep warnings

IgorMinar:
tsickle 0.37.1 is compatible with typescript 3.6, so we should use it and fix peerDep warnings from npm/yarn.

I had to update golden files along the way because the output has changed slightly.

And I also had to deal with with a breaking change in tsickle introduced by https://github.com/angular/tsickle/commit/cec4b484aa288256720d4c310dfcda0dcdc3da81

### DIFF
--- a/packages/compiler-cli/BUILD.bazel
+++ b/packages/compiler-cli/BUILD.bazel
@@ -45,6 +45,7 @@ ts_library(
         "@npm//@bazel/typescript",
         "@npm//@types/chokidar",
         "@npm//@types/node",
+        "@npm//minimist",
         "@npm//reflect-metadata",
         "@npm//tsickle",
         "@npm//typescript",
--- a/package.json
+++ b/package.json
@@ -115,7 +115,7 @@
     "source-map-support": "0.5.9",
     "systemjs": "0.18.10",
     "terser": "^4.4.0",
-    "tsickle": "0.37.0",
+    "tsickle": "0.37.1",
     "tslib": "^1.10.0",
     "tslint": "5.7.0",
     "typescript": "~3.6.4",
--- a/packages/bazel/package.json
+++ b/packages/bazel/package.json
@@ -28,7 +28,7 @@
   "dependencies": {
     "@microsoft/api-extractor": "^7.3.9",
     "shelljs": "0.8.2",
-    "tsickle": "^0.37.0"
+    "tsickle": "^0.37.1"
   },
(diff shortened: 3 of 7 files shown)


IgorMinar:
it seems that a bunch of stuff has changed in tsickle in .1 release :-( 

https://github.com/angular/tsickle/compare/5662f7266a8a392901984ba4421144bea1aa4755...afb02d51ae2978fa326b448f1095bf7cc5da10dc

the decorator downleveling change is likely due to https://github.com/angular/tsickle/commit/cec4b484aa288256720d4c310dfcda0dcdc3da81

packages/compiler-cli/BUILD.bazel
```diff
@@ -45,6 +45,7 @@ ts_library(
         "@npm//@bazel/typescript",
         "@npm//@types/chokidar",
         "@npm//@types/node",
+        "@npm//minimist",
```
IgorMinar: @gregmagolan @alexeagle fyi: this is likely one of those "that's how npm works" issue, but after a recent change in tsickle (which being updated in this PR) in which [tsickle dropped dependency on minimist](https://github.com/angular/tsickle/compare/5662f7266a8a392901984ba4421144bea1aa4755...afb02d51ae2978fa326b448f1095bf7cc5da10dc#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L15), one of our integration tests started failing because minimist in could not be found by compiler-cli.

It turns out that we were missing an explicit dependency on it and were silently depending on tsickle bringing minimist with it. Now that tsickle no longer brings minimist, the oversign in our build graph became visible.

I'm not sure if this is actionable on your end, just wanted to point it out in case this is not an expected behavior.

filipesilva (approved):

packages/compiler-cli/BUILD.bazel
gregmagolan: Thanks for the heads up. The failure showed up downstream of ts_library? Which integration test?
I tried looking at the CircleCI history but can't figure out how to see the old builds for a particular PR in the new UI 🙄

IgorMinar:

packages/compiler-cli/BUILD.bazel
IgorMinar: the target was //packages/compiler-cli/integrationtest:integrationtest

PR #33790, opened 2019-11-13
docs: add info on setting up VSCode remote development using docker containers

gkalpak:
In f78bda9ff, recommended configuration files were added to be used as basis for setting up [Remote Development using docker containers][1] in VSCode. Apparently, I had forgotten to commit the corresponding `README.md` file.

This commit adds `.devcontainer/README.md` with more info on Remote Development in VSCode.

[1]: https://code.visualstudio.com/docs/remote/containers

### DIFF
--- a/.devcontainer/README.md
+++ b/.devcontainer/README.md
@@ -1,0 +1,31 @@
+# VSCode Remote Development - Developing inside a Containers
+
+This folder contains configuration files that can be used to opt into working on this repository in a [Docker container](https://www.docker.com/resources/what-container) via [VSCode](https://code.visualstudio.com/)'s Remote Development feature (see below).
+
+Info on remote development and developing inside a container with VSCode:
+- [VSCode: Remote Development](https://code.visualstudio.com/docs/remote/remote-overview)
+- [VSCode: Developing inside a Container](https://code.visualstudio.com/docs/remote/containers)
+- [VSCode: Remote Development FAQ](https://code.visualstudio.com/docs/remote/faq)
+
+
+## Usage
+
+_Prerequisite: [Install Docker](https://docs.docker.com/install) on your local environment._
+
+To get started, read and follow the instuctions in [Developing inside a Container](https://code.visualstudio.com/docs/remote/containers). The [.devcontainer/](.) directory contains pre-configured `devcontainer.json` and `Dockerfile` files, which you can use to set up remote development with a docker container.
+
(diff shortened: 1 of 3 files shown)


filipesilva (approved):

.devcontainer/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Usage
+
+_Prerequisite: [Install Docker](https://docs.docker.com/install) on your local environment._
+
+To get started, read and follow the instuctions in [Developing inside a Container](https://code.visualstudio.com/docs/remote/containers). The [.devcontainer/](.) directory contains pre-configured `devcontainer.json` and `Dockerfile` files, which you can use to set up remote development with a docker container.
+
+In a nutshell, you need to:
+- Install the [Remote - Containers](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers) extension.
+- Copy [recommended-Dockerfile](./recommended-Dockerfile) to `Dockerfile` (and optionally tweak to suit your needs).
```
filipesilva: Nit: if these need to be copied/renamed, why not leave them with the correct name already?

filipesilva (approved):

.devcontainer/README.md
gkalpak: Sorry, I don't follow. We are `.gitignore`ing `.devcontainer/devcontainer.json` and `.devcontainer/Dockerfile` to allow people to have their own versions (without accidentally checking them in). We check in the `recommended-`-prefixed versions for people to copy and use as basis for their local configs.

This is similar to what we do with [.vscode/*][1] configs.

[1]: https://github.com/angular/angular/blob/2be0a1d35/.vscode

filipesilva: Sure, the pattern might be the same, but I'm not sure it needs to be. Unlike those settings, I don't think you'd even load these by accident or unknowingly. I guess it'd still be bad to have them checked in though.

gkalpak: This is not about not loading them by accident (although having them with their regular name does cause VSCode to prompt you to load the workspace in a container - you can disable that, but it is still anooying).

The main reason for not using their proper names is so that people can have their own configs (for example, I like to install certain packages in my container that others night not find useful) without them constantly showing up as local changes.

Does that make sense?

filipesilva: Yes, I think I understand it better now. Thank you.

PR #33794, opened 2019-11-13
refactor(ivy): minor improvements / cleanup in the DI code

pkozlowski-opensource:


mhevery (approved):

mhevery:
[presubmit](http://test/OCL:280215623:BASE:280215671:1573666606311:6997b999)
[ivy presubmit](http://test/OCL:280215623:BASE:280215872:1573666671263:a9335334)

packages/core/src/render3/instructions/di.ts
```diff
@@ -40,15 +41,15 @@ export function ɵɵdirectiveInject<T>(token: Type<T>| InjectionToken<T>): T;
 export function ɵɵdirectiveInject<T>(token: Type<T>| InjectionToken<T>, flags: InjectFlags): T;
 export function ɵɵdirectiveInject<T>(
     token: Type<T>| InjectionToken<T>, flags = InjectFlags.Default): T|null {
-  token = resolveForwardRef(token);
```
kara: Wouldn't this mean that we aren't resolving forward refs for tokens in the fallback case on line 47? Can we write a test to make sure we're not breaking that case?

kara (changes requested):

packages/core/src/render3/instructions/di.ts
pkozlowski-opensource: `ɵɵinject` is doing forwardRef resolving so we should be good. But it is always to have a test verifying this! I _was assuming_  that we've got one but I will check if removing forward ref resolution from  `ɵɵinject` breaks any of the existing tests and if not - I will add one.

kara: Ah OK, looks like we are good then!

kara (approved):
LGTM

PR #33801, opened 2019-11-13
Add API doc to secondary package entry points

jbogarthyde:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

 API doc for package pages missing from sub-packages

Issue Number: N/A

## What is the new behavior?
Doc added to all package pages 
https://angular.io/api?type=package

- [ ] Yes
- [x] No

### DIFF
--- a/packages/animations/test/PACKAGE.md
+++ b/packages/animations/test/PACKAGE.md
@@ -1,0 +1,1 @@
+Supplies a testing module for the Angular Animations driver.
--- a/packages/animations/test/PACKAGE.md
+++ b/packages/animations/test/PACKAGE.md
@@ -1,1 +1,0 @@
-Supplies a testing module for the Angular Animations driver.
--- a/packages/common/PACKAGE.md
+++ b/packages/common/PACKAGE.md
@@ -1,3 +1,3 @@
-Implements basic Angular directives and pipes, such as NgIf, NgForOf, DecimalPipe, and so on. 
+Implements basic Angular directives and pipes, such as `NgIf`, `NgForOf`, `DecimalPipe`, and so on.
 
 The `CommonModule` exports are re-exported by `BrowserModule`, which is included automatically in the root `AppModule` when you create a new app with the CLI `new` command.
--- a/packages/common/PACKAGE.md
+++ b/packages/common/PACKAGE.md
@@ -1,3 +1,3 @@
-Implements basic Angular directives and pipes, such as `NgIf`, `NgForOf`, `DecimalPipe`, and so on.
+Implements fundamental Angular framework functionality, including directives and pipes, location services used in routing, HTTP services, localization support, and so on.
 
 The `CommonModule` exports are re-exported by `BrowserModule`, which is included automatically in the root `AppModule` when you create a new app with the CLI `new` command.
--- a/packages/common/testing/PACKAGE.md
+++ b/packages/common/testing/PACKAGE.md
@@ -1,0 +1,1 @@
+Supplies a testing module for basic Angular directives.
--- a/packages/common/testing/PACKAGE.md
+++ b/packages/common/testing/PACKAGE.md
@@ -1,1 +1,1 @@
-Supplies a testing module for basic Angular directives.
+Supplies infrastructure for testing functionality supplied by `@angular/common`.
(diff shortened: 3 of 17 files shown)


jbogarthyde:
CARETAKER  - This touches API in all the code-owner areas, but the sub-package descriptions are short and straightforward - it wouldn't take long to check it.

packages/upgrade/PACKAGE.md
```diff
@@ -1,2 +1,3 @@
-Supports the upgrade path from AngularJS to Angular, allowing
-components from both systems to be used in the same application.
\ No newline at end of file
+**Deprecated** Provided support for upgrading applications from Angular JS to Angular.
```
gkalpak: This file pertains to the `@angular/upgrade` package. The package is not deprecated. (The primary `upgrade` entry-point is deprecated, which is different than deprecating the package.)

gkalpak (changes requested):
~~I don't think your changes have the effect you expect. AFAICT, the `PACKAGE.md` files pertain to a package (as the name suggests). Packages are only the top-level directories under `packages/` (e.g. `packages/animations/`, `packages/core/`, etc.).~~

~~The sub-directories do not represent sub-packages; they are secondary entry-points (which is a different thing). Currently, I don't think we support a separate docs page for entry-points.~~

~~So, only the top-level packages should have `PACKAGE.md` files (but you may want to add info about the secondary entry-points there).~~

_EDIT: It turns out I was wrong. We do take `PACKAGE.md` into account for secondary entry-points._
_(My commend about `@angular/upgrade` not being deprecated is still valid though.)_

packages/common/PACKAGE.md
```diff
@@ -1,3 +1,3 @@
-Implements basic Angular directives and pipes, such as NgIf, NgForOf, DecimalPipe, and so on. 
+Implements basic Angular directives and pipes, such as `NgIf`, `NgForOf`, `DecimalPipe`, and so on.
```
gkalpak: `@angular/common` does not only include directives and pipes, but some other things, such as `Location`, `LocationStrategy`, `ViewportScroller`, etc.

packages/animations/test/PACKAGE.md
```diff
@@ -0,0 +1 @@
+Supplies a testing module for the Angular Animations driver.
```
gkalpak: This directory is not a package or entry-point. (It just contains our automated tests.)
It should not have a `PACKAGE.md` file.

packages/common/testing/PACKAGE.md
```diff
@@ -0,0 +1 @@
+Supplies a testing module for basic Angular directives.
```
gkalpak: As mentioned in a previous comment, `@angular/common` does not only include directives and pipes. Also, this modules doesn't offer much in terms of testing directives/pipes (afaict). It mostly contains mock implementations for its services (e.g. `MockLocationStrategy`, `MockPlatformLocation, `SpyLocation`).

How about:
> Provides infrastructure for testing Angular common functionality.

or

> Provides infrastructure for testing functionality provided by the \`common\` package.

packages/upgrade/static/PACKAGE.md
```diff
@@ -0,0 +1,2 @@
+Supports the upgrade path from AngularJS to Angular, allowing
+components from both systems to be used in the same application.
```
gkalpak: Not only components, but also services.

gkalpak (changes requested):

gkalpak (approved):

IgorMinar:
@jbogarthyde @gkalpak  I just noticed this change, and realized that we are using the terminology wrong here.

most of the items in this PR are entry-points (https://angular.io/guide/glossary#entry-point) and not npm packages (https://angular.io/guide/glossary#npm-package).

e.g. @angular/common is package that contains several entry-points, including the primary entry-point @angular/common (the same name as the package), but also other entry-points like: @angular/common/testing, @angular/common/http, etc. If you go to https://angular.io/api/common you can see all the primary and secondary entrypoints and if you follow a link to one of the secondary entry point docs, you'll find it documented as a package (violet badge "PACKAGE") and not as an entry-point.

Is this a limitation of our docs-infra?

Ideally the docs-infra should enable us to document these entry points correctly, and not accept invalid package names like @angular/common/http.

@gkalpak or @jbogarthyde can you please add this to the docs-infra backlog?

gkalpak:
@IgorMinar: Yes, it seems to be a docs-infra limitation. I guess we only supported packages at first and later we piggy-backed on that feature to also document secondary entry-points.

I created #34081 and added it to the backlog of the docs-infra GitHub project.

PR #32562, opened 2019-09-09
perf(language-service): keep analyzedModules cache when source files don't change

andrius-pra:
This change will improve performance of language service when working in external templates.

## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] angular.io application / infrastructure changes
- [ ] Other... Please describe:

## What is the current behavior?

Issue Number: N/A

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

PR #32543, opened 2019-09-08
fix(language-service): create StaticReflector once only

kyliau:
The creation of `StaticReflector` in `createMetadataResolver()` is a very expensive operation because it involves numerous module resolutions (on the order of hundreds) to initialize core Angular symbols.
Since the API of the Reflector does not provide the ability to invalidate its internal caches, `StaticReflector` has to be destroyed and recreated on *every* program change.
This has a HUGE impact on performance.
This PR fixes this problem by carefully invalidating all StaticSymbols in a file that has changed, thereby greatly reducing the overhead of recomputation on program change.

## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] angular.io application / infrastructure changes
- [ ] Other... Please describe:

## What is the current behavior?

Issue Number: N/A

## Does this PR introduce a breaking change?

- [ ] Yes
- [ ] No

PR #33806, opened 2019-11-13
fix(language-service): Recompute analyzed modules only when source files change

kyliau:
This commit fixes a bug brought up by @andrius-pra whereby the language
service host would recompute the analyzed modules even when none of the
source files changes. This is due to a bug in our unit test that
precludes non-TS files from incrementing the project version.
Consequently, when the external template is updated, the program remains
the same.

With the bug fixed, the next step is to figure out if any source files
have been added / changed / removed since the last computation. The
previously analyzed could be safely retained only when none of these
operations happen.

Note: This restores the behavior @andrius-pra fixed in https://github.com/angular/angular/pull/32562 but regressed in https://github.com/angular/angular/pull/32543

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/packages/language-service/src/typescript_host.ts
+++ b/packages/language-service/src/typescript_host.ts
@@ -192,7 +192,8 @@ export class TypeScriptServiceHost implements LanguageServiceHost {
   }
 
   /**
-   * Checks whether the program has changed, and invalidate caches if it has.
+   * Checks whether the program has changed, and invalidate static symbols in
+   * the source files that have changed.
    * Returns true if modules are up-to-date, false otherwise.
    * This should only be called by getAnalyzedModules().
    */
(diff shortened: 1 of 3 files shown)


packages/language-service/src/typescript_host.ts
```diff
@@ -203,30 +204,45 @@ export class TypeScriptServiceHost implements LanguageServiceHost {
...
+      }
+    }
+
+    // Check if any source files have been removed since last computation.
+    for (const [fileName] of this.fileVersions) {
+      if (!seen.has(fileName)) {
+        filesChangedOrRemoved.push(fileName);  // removed
+        // Because Maps are iterated in insertion order, it is safe to delete
+        // entries from the same map while iterating.
+        // https://stackoverflow.com/questions/35940216
```
ayazhafiz: nit: maybe link to MDN (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) or the spec (https://www.ecma-international.org/ecma-262/10.0/index.html#sec-map.prototype.foreach) instead

ayazhafiz (approved):

packages/language-service/src/typescript_host.ts
kyliau: done, I added the link to the spec, I think the explanation there is better.

kyliau:
caretaker: I am a member of tools-language-service

PR #33813, opened 2019-11-14
build: update bazel to v1.1.0

devversion:
Updates Bazel to the latest stable version. Bazel 1.1.0
supposedly fixes a permission bug in Windows. Hence we
should try to update and see if that fixes the bug.

It's generally good to be up to date. See potential bug
fix commit: https://github.com/bazelbuild/bazel/commit/618e5a28f7f735c37724377b15775a4975349c74.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -143,14 +143,14 @@ rbe_autoconfig(
     # Need to specify a base container digest in order to ensure that we can use the checked-in
     # platform configurations for the "ubuntu16_04" image. Otherwise the autoconfig rule would
     # need to pull the image and run it in order determine the toolchain configuration. See:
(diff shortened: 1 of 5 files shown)


gkalpak (approved):

kyliau:
@devversion Could you also please update https://github.com/angular/angular/blob/b30bb8dd91c61684f16664e6ae6589113737a76d/packages/bazel/src/schematics/ng-add/index.ts#L51 ?

devversion:
@kyliau Done. Pushed a separate commit since this separate from `build`-only changes. Please have another look.

kyliau (approved):
thank you!

josephperrott (approved):
LGTM

PR #33816, opened 2019-11-14
fix(ivy): properly insert views into ViewContainerRef injected by querying <ng-container>

pkozlowski-opensource:
When asking for a ViewContainerRef on <ng-container> we do reuse <ng-container> comment
node as a LContainer's anchor. Before this fix the act of re-using a <ng-container>'s
comment node would result in this comment node being re-appended to the DOM in the wrong
place. With the fix in this PR we make sure that re-using <ng-container>'s comment node
doesn't result in unwanted DOM manipulation (ng-container's comment node is already part
of the DOM and doesn't have to be re-created / re-appended).

### DIFF
--- a/packages/core/src/render3/view_engine_compatibility.ts
+++ b/packages/core/src/render3/view_engine_compatibility.ts
@@ -331,25 +331,28 @@ export function createContainerRef(
     let commentNode: RComment;
     // If the host is an element container, the native host element is guaranteed to be a
     // comment and we can reuse that comment as anchor element for the new LContainer.
+    // The comment node in question is already part of the DOM structure so we don't need to append
+    // it again.
     if (hostTNode.type === TNodeType.ElementContainer) {
       commentNode = unwrapRNode(slotValue) as RComment;
     } else {
       ngDevMode && ngDevMode.rendererCreateComment++;
       commentNode = hostView[RENDERER].createComment(ngDevMode ? 'container' : '');
-    }
 
-    // A container can be created on the root (topmost / bootstrapped) component and in this case we
-    // can't use LTree to insert container's marker node (both parent of a comment node and the
-    // commend node itself is located outside of elements hold by LTree). In this specific case we
-    // use low-level DOM manipulation to insert container's marker (comment) node.
-    if (isRootView(hostView)) {
-      const renderer = hostView[RENDERER];
-      const hostNative = getNativeByTNode(hostTNode, hostView) !;
-      const parentOfHostNative = nativeParentNode(renderer, hostNative);
-      nativeInsertBefore(
-          renderer, parentOfHostNative !, commentNode, nativeNextSibling(renderer, hostNative));
-    } else {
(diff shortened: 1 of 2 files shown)


AndrewKushnir:
@pkozlowski-opensource thanks for the fix, the changes LGTM and I've started g3 presubmits:

- [VE Presubmit](http://test/OCL:280449317:BASE:280449353:1573753577310:84796940)
- [Ivy Presubmit](http://test/OCL:280449317:BASE:280449521:1573753623440:a9b5add2)

I think it'd be great if @mhevery can have a look as well (so I left the "review" label).

Thank you.

packages/core/src/render3/view_engine_compatibility.ts
```diff
@@ -331,25 +331,27 @@ export function createContainerRef(
...
-    if (isRootView(hostView)) {
-      const renderer = hostView[RENDERER];
-      const hostNative = getNativeByTNode(hostTNode, hostView) !;
-      const parentOfHostNative = nativeParentNode(renderer, hostNative);
-      nativeInsertBefore(
-          renderer, parentOfHostNative !, commentNode, nativeNextSibling(renderer, hostNative));
-    } else {
-      appendChild(commentNode, hostTNode, hostView);
+      // A container can be created on the root (topmost / bootstrapped) component and in this case
+      // we can't use LTree to insert container's marker node (both parent of a comment node and the
```
mhevery: `LTree` is not a type we have in system, neither is `LNode` What did you have in mind here. `RNode`???

I am a bit confused, so maybe an example would make this more clear?

packages/core/src/render3/view_engine_compatibility.ts
```diff
@@ -331,25 +331,27 @@ export function createContainerRef(
...
+      // commend node itself is located outside of elements hold by LTree). In this specific case we
```
mhevery: same here.

mhevery (changes requested):

packages/core/src/render3/view_engine_compatibility.ts
pkozlowski-opensource: Oh, this sounds like an early, pre-existing comment, I should have checked it when moving the code... Thnx, will clean-up!

AndrewKushnir:
New set of g3 presubmits:

- [VE Presubmit](http://test/OCL:280554718:BASE:280554762:1573782425089:4b103236)
- [Ivy Presubmit](http://test/OCL:280554718:BASE:280554869:1573782471086:859a6232)

AndrewKushnir:
FYI, VE and Ivy g3 presubmits look good 👍

mhevery (approved):

pkozlowski-opensource:
merge-assistance due to " Status "google3" is pending"

AndrewKushnir:
@pkozlowski-opensource I've updated the status of g3. I did ran presubmits a few days ago and the result looks good. Thank you.

PR #33820, opened 2019-11-14
fix(ivy): i18n - ensure that colons in i18n metadata are not rendered

petebacondarwin:
The `:` char is used as a metadata marker in `$localize` messages.
If this char appears in the metadata it must be escaped, as `\:`.
Previously, although the `:` char was being escaped, the TS AST
being generated was not correct and so it was being output double
escaped, which meant that it appeared in the rendered message.

As of TS 3.6.2 the "raw" string can be specified when creating tagged
template AST nodes, so it is possible to correct this.


ocombe (changes requested):
Looks good, except for `debug-test.sh` and `yarn.lock` that shouldn't be part of the PR! :)

petebacondarwin:
Thanks @ocombe - I removed the `yarn.lock` but I moved the `debug-test.sh` to its own commit since it is an important fix and might as well go in with this PR.

ocombe (approved):

AndrewKushnir (approved):
LGTM 👍

AndrewKushnir:
- [VE Presubmit](http://test/OCL:280446835:BASE:280446929:1573752941439:edbd4c48)
- [Ivy Presubmit](http://test/OCL:280446835:BASE:280447117:1573752987821:b1853a11)

alxhub (approved):

josephperrott (approved):
LGTM

AndrewKushnir:
New set of g3 presubmits:

- [VE Presubmit](http://test/OCL:280554244:BASE:280554306:1573782209900:750a5e8f)
- [Ivy Presubmit](http://test/OCL:280554244:BASE:280554434:1573782269042:2001ba81)

AndrewKushnir:
@petebacondarwin g3 presubmit indicated the following problem:
```
Error: Debug Failure. False expression: Expected argument 'text' to be the normalized (i.e. 'cooked') version of argument 'rawText'.
    at createTemplateLiteralLikeNode (javascript/node_modules/typescript/stable/lib/typescript.js:64087:22)
    at Object.createNoSubstitutionTemplateLiteral (javascript/node_modules/typescript/stable/lib/typescript.js:64111:20)
    at visitLocalizedString (packages/compiler-cli/src/ngtsc/translator/src/translator.ts:552:19)
...
```
Even though TS version requirement is satisfied. I sent you a message in Slack with additional info.

petebacondarwin:
From looking at the assertion - TS is re-parsing the raw text as though it was in the original source.

For example if we are creating a simple no-substitution template literal with the text abc they put this back through their normal tokenizer (scanner) wrapped in backticks `abc` and check that they then get the original `abc` out.

It is likely that the failing messages contain unescaped backticks or `${` sequences.

For example

```
ab`c
```

In this case they would put it through the scanner as
```
`ab`c`
```
which would obviously return only `ab` as the cooked string.

I have added a commit that escapes these sequences.

AndrewKushnir:
Thanks for the fix @petebacondarwin.

I've restarted g3 presubmits with the most up-to-date version of this PR:

- [VE Presubmit](http://test/OCL:280677394:BASE:280677445:1573839814961:95a6f490)
- [Ivy Presubmit](http://test/OCL:280677394:BASE:280677713:1573839903785:4b6d038d)

IgorMinar:
Still fails with "Debug Failure. False expression: Expected argument 'text' to be the normalized (i.e. 'cooked') version of argument 'rawText'."

AndrewKushnir:
@petebacondarwin I performed additional investigation and here is the pattern that is causing failures in g3:

```
<div i18n="Some text \' [BACKUP_MESSAGE_ID: xxx]">Test</div>
```

petebacondarwin:
Thanks @AndrewKushnir - is the `\'` supposed to be rendered as a `'` or as `\'` in the browser?
I can "fix" this by double escaping any `\`s that are found in the description, but I am wondering what the actual intent is.

petebacondarwin:
I have amended the last commit to also escape `\` in the "raw" form.
@AndrewKushnir - please can you try the presubmit one more time?

AndrewKushnir:
Thanks for the fix @petebacondarwin. I've started new presubmits:

- [VE Presubmit](https://test.corp.google.com/ui#id=OCL:280863115:BASE:280863123:1573944232930:123a7a1a)
- [Ivy Presubmit](http://test/OCL:280863115:BASE:280863209:1573944299794:6d4f52bc)

AndrewKushnir:
Hi @petebacondarwin, VE and Ivy presubmits look good with the most recent changes, so I've set the g3 status to "green" and removed the "blocked" label. Thanks for the fixes 👍

packages/compiler-cli/test/compliance/mock_compile.ts
```diff
@@ -14,15 +14,15 @@ import {NgtscProgram} from '../../src/ngtsc/program';
 
 const IDENTIFIER = /[A-Za-z_$ɵ][A-Za-z0-9_$]*/;
 const OPERATOR =
-    /!|\?|%|\*|\/|\^|&&?|\|\|?|\(|\)|\{|\}|\[|\]|:|;|<=?|>=?|={1,3}|!==?|=>|\+\+?|--?|@|,|\.|\.\.\./;
+    /!|\?|%|\*|\/|\^|&&?|\|\|?|\(|\)|\{|\}|\[|\]|:|;|<=?|>=?|={1,3}|!==?|=>|\+\+?|--?|@|,|\.|\.\.\.|`|\\'/;
 const STRING = /'(\\'|[^'])*'|"(\\"|[^"])*"/;
-const BACKTICK_STRING = /\\`(([\s\S]*?)(\$\{[^}]*?\})?)*?\\`/;
+const BACKTICK_STRING = /\\`(([\s\S]*?)(\$\{[^}]*?\})?)*?[^\\]\\`/;
```
gkalpak: OOC, what does the `[^\\]` do?

petebacondarwin: :-)

This is so that we can pick up escaped single quotes (i.e. `\'`) that exist outside of single quoted strings.

gkalpak: I see. I guess you don't need the capturing groups (because they will be messed up anyway :grin:), right?

Nit: BTW, `\$\{[^}]*?\}` doesn't need to be non-greedy: `\$\{[^}]*\}`

PR #33824, opened 2019-11-14
docs(platform-webworker): mark all public APIs as deprecated

gkalpak:
In ccc76f749, `@angular/platform-webworker` and `@angular/platform-webworker-dynamic` were deprecated (i.e. by deprecating all their public APIs). However, some public `@angular/platform-webworker` APIs were accidentally missed, additionally resulting in the package itself not showing up as deprecated on angular.io.

This commit fixes it by marking those remaining public APIs as deprecated.

### DIFF
--- a/packages/platform-webworker/src/web_workers/shared/client_message_broker.ts
+++ b/packages/platform-webworker/src/web_workers/shared/client_message_broker.ts
@@ -144,6 +144,7 @@ interface ResponseMessageData {
 
 /**
  * @publicApi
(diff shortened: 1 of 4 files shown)


gkalpak:
**Before:**
deprecate-webworker before

**After:**
deprecate-webworker after

IgorMinar (approved):
LGTM! thanks for the catching this oversight!

IgorMinar:
merge-assistance: no g3 impact

PR #33826, opened 2019-11-14
docs(ivy): improve the missing `$localize` error message

petebacondarwin:
If the application is not running directly in the browser, e.g.
universal or app-shell, then the `$localize` import must be
adding to a different file than for normal browser applications.

This commit adds more information about this to avoid any
confusion.

// FW-1557


IgorMinar (approved):

IgorMinar:
merge-assistance: no g3 impact

PR #33827, opened 2019-11-14
refactor(ivy): use styling direct mode for all non-duplicate entries

matsko:
This patch ensures that all template-level style and class bindings
are processed using the direct-mode when their values are applied to
an element.


matsko:
presubmit = https://tap.corp.google.com/ui#id=OCL:281647248:BASE:281647298:1574302010861:db78c133

ivy presubmit = https://test.corp.google.com/ui#id=OCL:281647248:BASE:281648237:1574302398351:ea091053

packages/compiler/src/render3/view/styling_builder.ts
```diff
@@ -19,6 +19,7 @@ import {ValueConverter} from './template';
 import {getInterpolationArgsLength} from './util';
 
 const IMPORTANT_FLAG = '!important';
+export const MIN_STYLING_BINDING_SLOTS_REQUIRED = 2;
```
mhevery: Please add documentation explaining why this is needed, and how it is used.

packages/compiler/src/render3/view/styling_builder.ts
```diff
@@ -375,7 +376,12 @@ export class StylingBuilder {
           instructions[instructions.length - 1];
       const value = input.value.visit(valueConverter);
       let referenceForCall = reference;
-      let totalBindingSlotsRequired = 1;  // each styling binding value is stored in the LView
+
+      // each styling binding value is stored in the LView
+      // but there are two values stored for each binding:
+      //   1) the value itself
+      //   2) any intermediate values used for the direct-write algorithm
```
mhevery:
```suggestion
      //   2) an intermediate value (concatenation of style up to this point).
      //      We need to store the intermediate value so that we don't allocate the strings on each CD.
```

packages/core/src/render3/instructions/element.ts
```diff
@@ -64,7 +64,7 @@ export function ɵɵelementStart(
     }
   }
 
-  if ((tNode.flags & TNodeFlags.hasInitialStyling) === TNodeFlags.hasInitialStyling) {
+  if (tNode.flags & TNodeFlags.hasInitialStyling) {
```
mhevery: Why this change? I think the original was correct?

packages/core/src/render3/interfaces/node.ts
```diff
@@ -78,29 +78,37 @@ export const enum TNodeFlags {
    */
   hasHostBindings = 0x80,
 
-  /** Bit #9 - This bit is set if the node has initial styling */
-  hasInitialStyling = 0x100,
+  /** Bit #9 - This bit is set if the node has any initial classes */
```
mhevery: Please add comment which provides value over just repeating what the code already says. A comment should answer:
- Why do we need this.
- How does this work on high level

And ideally the comment should contain an example which demonstrates why this is needed.

packages/core/src/render3/interfaces/node.ts
```diff
@@ -78,29 +78,37 @@ export const enum TNodeFlags {
...
+  hasInitialClasses = 0x100,
+
+  /** Bit #10 - This bit is set if the node has any initial styles */
```
mhevery: same here

packages/core/src/render3/interfaces/node.ts
```diff
@@ -78,29 +78,37 @@ export const enum TNodeFlags {
...
+  hasInitialStyles = 0x200,
+
+  /**
+   * whether or not there are any initial styles or initial classes present
```
mhevery: same here

packages/core/src/render3/interfaces/node.ts
```diff
@@ -148,40 +156,50 @@ export const enum TNodeFlags {
...
    * Examples include:
    * 1. prop + prop: `<div [style.width]="x" dir-that-sets-width>`
    * 2. map + prop: `<div [style]="x" [style.prop]>`
    * 3. map + map: `<div [style]="x" dir-that-sets-style>`
    */
-  hasDuplicateStyleBindings = 0x40000,
+  hasDuplicateStyleBindings = 0x80000,
+
+  /**
+   * whether or not a `TStylingContext` instance is used for classes (merge mode algorithm active)
```
mhevery: improve comment

packages/core/src/render3/interfaces/node.ts
```diff
@@ -148,40 +156,50 @@ export const enum TNodeFlags {
...
+   */
+  hasClassContextInUse = 0x100000,
+
+  /**
+   * whether or not a `TStylingContext` instance is used for styles (merge mode algorithm active)
```
mhevery: improve comment

packages/core/src/render3/interfaces/node.ts
```diff
@@ -580,6 +598,11 @@ export interface TNode {
    */
   styles: StylingMapArray|TStylingContext|null;
 
+  /**
+   * Last binding index for any `[style]` or `[style.prop]` bindings on this node.
+   */
+  stylesBindingIndex: number;
```
mhevery: - Why is this needed (what would not be possible or difficult to do otherwise)
- Add an example which demonstrates how this is used.

packages/core/src/render3/interfaces/node.ts
```diff
@@ -599,6 +622,11 @@ export interface TNode {
    * will be placed into the initial styling slot in the newly created `TStylingContext`.
    */
   classes: StylingMapArray|TStylingContext|null;
+
+  /**
+   * Last binding index for any `[class]` or `[class.name]` bindings on this node.
```
mhevery: here too. OK to reference style comment and just say same but for classes.

packages/core/src/render3/interfaces/styling.ts
```diff
@@ -470,4 +470,18 @@ export const enum StylingMapsSyncMode {
  *
  * The styling algorithm code only needs access to `flags`.
  */
-export interface TStylingNode { flags: TNodeFlags; }
+export interface TStylingNode {
```
mhevery: This seems to be just a subset of `TNode`. If that is the case:
- what is the value in having it? (If it is really needed, can we beef up comments?)
- The type system should ensure that the two don't go out of sync. (So probably a check which asserts that one can be assigned to other?)

packages/core/src/render3/interfaces/styling.ts
```diff
@@ -470,4 +470,18 @@ export const enum StylingMapsSyncMode {
...
+  classesBindingIndex: number;
+  stylesBindingIndex: number;
+  flags: TNodeFlags;
+}
+
+/**
+ * Various flags used for each style/class binding entry stored inside of `TData`.
```
mhevery: - Please add comment which would explain how this is used, why needed etc...
- Each flag should be documented likewise. Under what circumstances is the flag used, why we can't do without it.

mhevery (changes requested):
I was not able to finish the review tonight, will look at it more tomorrow.

packages/core/src/render3/instructions/element.ts
matsko: The flag is now a combination of two other flags. Therefore, when either bit is set, then it should be true. This means the statement above fails unless it just checks for the presence of one of the true bits.

mhevery: I see

packages/core/src/render3/interfaces/node.ts
matsko: done

packages/core/src/render3/interfaces/styling.ts
matsko: done

packages/core/src/render3/interfaces/styling.ts
matsko: done

packages/core/src/render3/interfaces/node.ts
matsko: done

packages/core/src/render3/interfaces/node.ts
matsko: done

packages/core/src/render3/interfaces/node.ts
matsko: done

packages/core/src/render3/interfaces/node.ts
matsko: done

packages/core/src/render3/interfaces/node.ts
matsko: done

packages/core/src/render3/interfaces/node.ts
matsko: done

matsko: done

packages/compiler/src/render3/view/styling_builder.ts
matsko: fixed

packages/compiler/src/render3/view/styling_builder.ts
matsko: done

packages/core/src/render3/instructions/styling.ts
```diff
@@ -618,3 +643,66 @@ function patchHostStylingFlag(tNode: TNode, hostBindingsMode: boolean, isClassBa
...
+      i += StylingMapArrayIndex.TupleSize;
+    }
+  }
+
+  setPreviousBindingIndex(
+      tData, bindingIndex, previousBindingIndex, sanitizationRequired, hasPotentialOverlap);
+}
+
+function hyphenate(value: string): string {
+  return value.replace(/[a-z][A-Z]/g, v => v.charAt(0) + '-' + v.charAt(1)).toLowerCase();
```
mhevery: Extra regexp into `const` so that we don't allocate new regexp on each execution.

packages/core/src/render3/instructions/styling.ts
```diff
@@ -618,3 +643,66 @@ function patchHostStylingFlag(tNode: TNode, hostBindingsMode: boolean, isClassBa
...
+ */
+export function registerBindingIntoTNode(
+    tNode: TNode, tData: TData, prop: string | null, bindingIndex: number, isClassBased: boolean,
+    sanitizationRequired: boolean): void {
+  tData[bindingIndex] = (isClassBased && prop !== null) ? hyphenate(prop) : prop;
+
+  let previousBindingIndex: number;
+  if (isClassBased) {
+    previousBindingIndex = tNode.classesBindingIndex;
+    tNode.classesBindingIndex = bindingIndex;
```
mhevery: It seems like `classesBindingIndex` points to tail. This is not obvious from the name of property. I also think that updating `TNode` like this is an issue. It means that if we reading a binding would have side-effects we could re-enter this code, and we could have two places where such update would happen. (In other words this code is not re-entry safe)

I think `classesBindingIndex` should always point to the beginning of the chain. Let's discuss this.

packages/core/src/render3/interfaces/node.ts
```diff
@@ -78,29 +78,63 @@ export const enum TNodeFlags {
...
-  /** Bit #9 - This bit is set if the node has initial styling */
-  hasInitialStyling = 0x100,
+  /**
+   * Bit #9 - This bit is set if the node has any initial classes
+   *
+   * Tracking initial classes using this flag is important for performance
+   * reasons because it enables the class binding code to avoid certain
+   * operations (i.e. if a class value doesn't exist both in binding code
+   * and in the initial classes code then there is no need to search/replace
+   * the class out of the className string).
```
mhevery: code example would make this even better.

packages/core/src/render3/interfaces/node.ts
```diff
@@ -78,29 +78,63 @@ export const enum TNodeFlags {
...
+   * and in the initial styles code then there is no need to search/replace
+   * the class out of the style attribute string).
+   */
+  hasInitialStyles = 0x200,
+
+  /**
+   * Whether or not there are any initial styles or initial classes present
+   *
+   * This flag is a combination of both flags for tracking initial styles
+   * and classes. There are various cases in element creation that the
```
mhevery: `There are various cases ` List the cases under which this happens as bullet list. The point of documentation is to explain exactly why it is needed.

mhevery (changes requested):
More comments, will continue after lunch

packages/compiler/src/render3/view/compiler.ts
```diff
@@ -547,10 +546,36 @@ function createHostBindingsFunction(
...
+
+  // figure out how many min host vars are required before
+  // interpolations and values are converted.
+  let totalHostVarsCount = 0;
+  bindings && bindings.forEach((binding: ParsedProperty) => {
+    const name = binding.name;
+    const stylingInputWasSet =
+        styleBuilder.registerInputBasedOnName(name, binding.expression, binding.sourceSpan);
+    if (stylingInputWasSet) {
+      totalHostVarsCount += MIN_STYLING_BINDING_SLOTS_REQUIRED;
```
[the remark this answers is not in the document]
matsko: We use `MIN` because interpolation adds additional slot requirements on top of the minimum entries.

packages/compiler/src/render3/view/compiler.ts
```diff
@@ -671,7 +680,8 @@ function createHostBindingsFunction(
         instruction.calls.forEach(call => {
           // we subtract a value of `1` here because the binding slot was already allocated
           // at the top of this method when all the input bindings were counted.
-          totalHostVarsCount += Math.max(call.allocateBindingSlots - 1, 0);
+          totalHostVarsCount +=
```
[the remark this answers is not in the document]
matsko: We first increment the total binding amounts necessary before pipes and interpolation values are added. This is the nature of host bindings code, because all other bindings need to know exactly how many slots are required before they are evaluated. This is why we increment the style/class bindings first (using that constant) and then subtract them out later one (when the interpolations are evaluated).

packages/core/src/render3/instructions/styling.ts
```diff
@@ -175,14 +182,11 @@ export function ɵɵclassProp(className: string, value: boolean | null): Tsickle
...
+    tNode: TNode, firstUpdatePass: boolean, tData: TData, lView: LView, bindingIndex: number,
+    prop: string, value: boolean | number | SafeValue | string | null | undefined | NO_CHANGE,
     isClassBased: boolean): boolean {
   let updated = false;
-
   const native = getNativeByTNode(tNode, lView) as RElement;
-  const context = isClassBased ? getClassesContext(tNode) : getStylesContext(tNode);
-  const sanitizer = isClassBased ? null : getCurrentStyleSanitizer();
 
   // [style.prop] and [class.name] bindings do not use `bind()` and will
```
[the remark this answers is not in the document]
matsko: We do this outside of the instruction with our own binding logic (the `styleProp`, `classProp`, `styleMap` and `classMap` instructions all do their own reading/writing to `LView`).

packages/core/src/render3/instructions/styling.ts
matsko: fixed

packages/core/src/render3/instructions/styling.ts
matsko: The problem is that host bindings have no way of connecting back to template bindings if we do a head-only approach. 

The `tNode.classBindingIndex` and `tNode.stylesBindingIndex` values only get updated during the `firstUpdatePass` for a single instance of the first element. Then each member stays the same for future updates. We do the same thing thing with `TStlyingContext` as new bindings are registered.

packages/core/src/render3/interfaces/node.ts
matsko: done

packages/core/src/render3/interfaces/node.ts
matsko: fixed

packages/compiler-cli/test/compliance/r3_view_compiler_styling_spec.ts
```diff
@@ -504,7 +504,7 @@ describe('compiler compliance: styling', () => {
               type: MyComponent,
               selectors:[["my-component"]],
               decls: 1,
-              vars: 5,
+              vars: 7,
```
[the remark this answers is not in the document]
matsko: That is just a regular binding. So nothing happens. We add `+2` here because of the two style.prop bindings (`[style.width]` and `[style.height]`). We don't add `+3` because `[style]` already uses two binding slots prior to this PR.

packages/compiler/src/render3/view/compiler.ts
```diff
@@ -547,10 +546,36 @@ function createHostBindingsFunction(
...
+  if (eventBindings && eventBindings.length) {
+    const listeners = createHostListeners(eventBindings, name);
+    createStatements.push(...listeners);
+  }
+
+  // Calculate the host property bindings
+  const bindings = bindingParser.createBoundHostProperties(directiveSummary, hostBindingSourceSpan);
+  const allOtherBindings: ParsedProperty[] = [];
+
+  // figure out how many min host vars are required before
```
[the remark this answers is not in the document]
matsko: fixed.

packages/compiler/src/render3/view/compiler.ts
```diff
@@ -531,7 +531,6 @@ function createHostBindingsFunction(
     bindingParser: BindingParser, constantPool: ConstantPool, selector: string,
     name?: string): o.Expression|null {
   // Initialize hostVarsCount to number of bound host properties (interpolations illegal)
```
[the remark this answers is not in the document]
matsko: fixed. thank you.

packages/core/src/render3/styling/direct_write_algorithm.ts
```diff
@@ -0,0 +1,327 @@
...
+ * @returns `true` when the binding has updated (due to a value change).
+ */
+export function setStylingValue(
+    lView: LView, tNode: TNode, tData: TData, state: StylingState, value: any, bindingIndex: number,
+    sanitizer: StyleSanitizeFn | null, isClassBased: boolean): boolean {
+  const lastUpdatedBindingIndex =
+      isClassBased ? state.lastClassBindingIndex : state.lastStyleBindingIndex;
+  const previousValue = getValue(lView, bindingIndex);
+  const updated = hasValueChanged(previousValue, value);
+  if (lastUpdatedBindingIndex < 0 || updated) {
```
mhevery: Why is the check for value changed so deep in the algorithm? I would assume that it would simpler if it would be close to the `styleProp` (and friends) function, so that in the case when there is no change we could exit quickly.

packages/core/src/render3/styling/state.ts
```diff
@@ -59,25 +58,11 @@ export interface StylingState {
...
-  /**
-   * The last class map that was applied (i.e. `[class]="x"`).
-   *
-   * Note that this property is only populated when direct class values are applied
-   * (i.e. context resolution is not used).
-   *
-   * See `allowDirectStyling` for more info.
-  */
-  lastDirectClassMap: StylingMapArray|null;
+  lastClassBindingIndex: number;
```
mhevery: Could we have docs explaining how each of these properties are used.

packages/core/src/render3/util/styling_utils.ts
```diff
@@ -455,3 +449,192 @@ export function splitOnWhitespace(text: string): string[]|null {
 export function selectClassBasedInputName(inputs: PropertyAliases): string {
   return inputs.hasOwnProperty('class') ? 'class' : 'className';
 }
+
+/**
+ * Returns a style entry with the `prop` and `value` concatenated together
+ */
+export function concatStyle(prop: string, value: string) {
+  return `${prop}${STYLE_PROP_VALUE_SEPARATOR}${value}`;
```
mhevery:
```suggestion
  return prop +STYLE_PROP_VALUE_SEPARATOR + value;
```
Is a lot more efficient

mhevery (changes requested):

PR #33828, opened 2019-11-14
Monorepository import support

alxhub:
❤️


packages/compiler-cli/src/ngtsc/file_system/testing/src/mock_file_system.ts
```diff
@@ -132,7 +132,37 @@ export abstract class MockFileSystem implements FileSystem {
...
+    // E.g. if the current directory is /foo/bar, look in /foo/bar/node_modules, then
+    // /foo/node_modules, then /node_modules.
+
+    let path = 'node_modules/typescript/lib';
+    let resolvedPath = this.resolve(path);
+
+    // Construct a path for the top-level node_modules to identify the stopping point.
+    const topLevelNodeModules = this.resolve('/' + path);
+
+    // do-while ensures the loop runs once for the top-level node_modules as well.
```
JoostK: Outdated comment

packages/compiler-cli/src/ngtsc/file_system/testing/src/mock_file_system.ts
```diff
@@ -132,7 +132,37 @@ export abstract class MockFileSystem implements FileSystem {
...
+    while (resolvedPath !== topLevelNodeModules) {
+      if (this.exists(resolvedPath)) {
+        return resolvedPath;
+      }
+
+      // Not here, look one level highier.
```
JoostK:
```suggestion
      // Not here, look one level higher.
```

packages/compiler-cli/src/ngtsc/imports/src/emitter.ts
```diff
@@ -233,6 +238,28 @@ export class LogicalProjectStrategy implements ReferenceEmitStrategy {
...
+ * This strategy can be used if there is no `rootDir`/`rootDirs` structure for the project which
+ * necessitates the stronger logic of `LogicalProjectStrategy`.
+ */
+export class RelativePathStrategy implements ReferenceEmitStrategy {
+  constructor(private reflector: ReflectionHost) {}
+
+  emit(ref: Reference<ts.Node>, context: ts.SourceFile): Expression|null {
+    const destSf = getSourceFile(ref.node);
+    let moduleName =
+        relative(dirname(resolve(context.fileName)), resolve(stripExtension(destSf.fileName)));
```
JoostK: nit: I would move the `stripExtension` to be last, and maybe use `absoluteFromSourceFile`:
```suggestion
        stripExtension(relative(dirname(absoluteFromSourceFile(context)), absoluteFromSourceFile(destSf)));
```

packages/compiler-cli/src/ngtsc/program.ts
```diff
@@ -520,6 +520,24 @@ export class NgtscProgram implements api.Program {
...
+      // sufficient.
+      if (this.options.rootDir !== undefined ||
+          (this.options.rootDirs !== undefined && this.options.rootDirs.length > 0)) {
+        // rootDirs logic is in effect - use the `LogicalProjectStrategy` for in-project relative
+        // imports.
+        localImportStrategy =
+            new LogicalProjectStrategy(this.reflector, new LogicalFileSystem(this.rootDirs));
+      } else {
+        // Plain relative imports are all that's needed.
+        localImportStrategy = new RelativePathStrategy(this.reflector);
```
JoostK: I think you should be able to replace the `LocigalProjectStrategy` with `RelativePath` strategy within ngcc, as I recall there being a TODO there and this new strategy looks to implement exactly that.

packages/compiler-cli/test/ngtsc/monorepo_spec.ts
```diff
@@ -0,0 +1,73 @@
...
+          selector: 'lib-cmp',
+          template: '...',
+        })
+        export class LibCmp {}
+      `);
+
+      env.driveMain();
+
+      const jsContents = env.getContents('app/index.js');
+      expect(jsContents).toContain(`import * as i1 from "../lib/cmp";`);
```
JoostK: We are missing a test for how a reference into the same directory tree is created (i.e. it should be prefixed with `./`)

JoostK (approved):
LGTM with a couple of requests.

packages/compiler-cli/src/ngtsc/imports/src/emitter.ts
alxhub: `stripExtension` is `(string) => string`, and everything else deals in `AbsoluteFsPath`, so it makes sense as the inner operation.

But I do like the idea of `absoluteFromSourceFile` - doing that.

alxhub: It took me a moment to understand this, but 👍

packages/compiler-cli/src/ngtsc/program.ts
alxhub: You're right that there is. But looking at this makes me wonder.

Right now `LogicalProjectStrategy` protects ngcc from emitting a relative import that points outside of the current entrypoint (and thus outside of the current package). This seems like it might be a _reasonable_ assertion to make. I'll leave it as-is for now, and we can examine the tradeoffs here in another PR. I don't wanna couple this important fix to that decision.

alxhub:
[Presubmit](http://test/OCL:281341126:BASE:281341178:1574192200975:9dd901e6)
No Ivy Presubmit needed; these changes don't actually affect g3.

PR #33837, opened 2019-11-15
perf(ivy): chain styling instructions

crisbeto:
Adds support for chaining of `styleProp`, `classProp` and `stylePropInterpolateX` instructions whenever possible which should help generate less code. Note that one complication here is for `stylePropInterpolateX` instructions where we have to break into multiple chains if there are other styling instructions inbetween the interpolations which helps maintain the execution order.

### DIFF
--- a/packages/compiler-cli/test/compliance/r3_view_compiler_styling_spec.ts
+++ b/packages/compiler-cli/test/compliance/r3_view_compiler_styling_spec.ts
@@ -513,8 +513,7 @@ describe('compiler compliance: styling', () => {
                 if (rf & 2) {
                   $r3$.ɵɵstyleSanitizer($r3$.ɵɵdefaultStyleSanitizer);
                   $r3$.ɵɵstyleMap($ctx$.myStyleExp);
-                  $r3$.ɵɵstyleProp("width", $ctx$.myWidth);
-                  $r3$.ɵɵstyleProp("height", $ctx$.myHeight);
+                  $r3$.ɵɵstyleProp("width", $ctx$.myWidth)("height", $ctx$.myHeight);
                   $r3$.ɵɵattribute("style", "border-width: 10px", $r3$.ɵɵsanitizeStyle);
                 }
               },
@@ -706,8 +705,7 @@ describe('compiler compliance: styling', () => {
                 }
                 if (rf & 2) {
                   $r3$.ɵɵclassMap($ctx$.myClassExp);
-                  $r3$.ɵɵclassProp("apple", $ctx$.yesToApple);
-                  $r3$.ɵɵclassProp("orange", $ctx$.yesToOrange);
+                  $r3$.ɵɵclassProp("apple", $ctx$.yesToApple)("orange", $ctx$.yesToOrange);
                   $r3$.ɵɵattribute("class", "banana");
                 }
               },
@@ -916,8 +914,7 @@ describe('compiler compliance: styling', () => {
               $r3$.ɵɵstyleSanitizer($r3$.ɵɵdefaultStyleSanitizer);
               $r3$.ɵɵstyleMap($r3$.ɵɵpipeBind2(1, 8, $ctx$.myStyleExp, 1000));
               $r3$.ɵɵclassMap($r3$.ɵɵpureFunction0(20, _c0));
-              $r3$.ɵɵstyleProp("bar", $r3$.ɵɵpipeBind2(2, 11, $ctx$.barExp, 3000));
-              $r3$.ɵɵstyleProp("baz", $r3$.ɵɵpipeBind2(3, 14, $ctx$.bazExp, 4000));
+              $r3$.ɵɵstyleProp("bar", $r3$.ɵɵpipeBind2(2, 11, $ctx$.barExp, 3000))("baz", $r3$.ɵɵpipeBind2(3, 14, $ctx$.bazExp, 4000));
               $r3$.ɵɵclassProp("foo", $r3$.ɵɵpipeBind2(4, 17, $ctx$.fooExp, 2000));
               $r3$.ɵɵadvance(5);
              $r3$.ɵɵtextInterpolate1(" ", $ctx$.item, "");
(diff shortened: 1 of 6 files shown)


mhevery (approved):

packages/compiler/src/render3/view/compiler.ts
```diff
@@ -664,8 +668,16 @@ function createHostBindingsFunction(
       // we subtract a value of `1` here because the binding slot was already
       // allocated at the top of this method when all the input bindings were
       // counted.
-      totalHostVarsCount += Math.max(instruction.allocateBindingSlots - 1, 0);
-      updateStatements.push(createStylingStmt(instruction, bindingContext, bindingFn));
+      if (instruction.calls.length > 0) {
+        const calls: o.Expression[][] = [];
+
+        instruction.calls.forEach(call => {
+          totalHostVarsCount += Math.max(call.allocateBindingSlots - 1, 0);
```
matsko: Add a comment explaining why there's a `-1`

packages/compiler/src/render3/view/styling_builder.ts
```diff
@@ -25,10 +24,14 @@ const IMPORTANT_FLAG = '!important';
  * A styling expression summary that is to be processed by the compiler
  */
 export interface StylingInstruction {
-  sourceSpan: ParseSourceSpan|null;
   reference: o.ExternalReference;
-  allocateBindingSlots: number;
+  calls: StylingInstructionCall[];
```
matsko: Add a comment explaining what this new member is.

packages/compiler/src/render3/view/styling_builder.ts
```diff
@@ -25,10 +24,14 @@ const IMPORTANT_FLAG = '!important';
  * A styling expression summary that is to be processed by the compiler
  */
 export interface StylingInstruction {
```
matsko: Since we're only using one property on this interface, why not just return a array of `StylingInstructionCall` instances?

packages/core/src/render3/instructions/styling.ts
```diff
@@ -133,7 +135,7 @@ export function stylePropInternal(
  *
  * @codeGenApi
  */
-export function ɵɵclassProp(className: string, value: boolean | null): void {
+export function ɵɵclassProp(className: string, value: boolean | null): TsickleIssue1009 {
```
matsko: Add a comment explaining the return value and why it's an instance of `TsickleIssue1009`

packages/compiler/src/render3/view/template.ts
```diff
@@ -1127,7 +1131,11 @@ export class TemplateDefinitionBuilder implements t.Visitor<void>, LocalResolver
...
     this._updateCodeFns.push(() => {
       const calls = bindings.map(property => {
-        const fnParams = [property.value(), ...(property.params || [])];
+        const value = property.value();
+        const fnParams = Array.isArray(value) ? value : [value];
+        if (property.params) {
+          fnParams.push(...property.params);
+        }
         if (property.name) {
           fnParams.unshift(o.literal(property.name));
```
matsko: Add a comment explaining this `unshift` operation.

matsko (changes requested):

packages/compiler/src/render3/view/styling_builder.ts
crisbeto: It's set up like this so that it's easier to group calls to the same instruction together. If we wanted to change it, we'd have to do something like `StylingInstructionCall[][]` instead, but that would mean looping through all the calls when we're trying to figure out whether to chain them.

packages/core/src/render3/instructions/styling.ts
crisbeto: Isn't this the point of the `TsickleIssue1009` type in the first place? Also Andrew mentioned that this isn't an issue anymore so as soon as these changes land I'll clean out all the usages of the `TsickleIssue1009` type.

crisbeto:
The feedback has been addressed @matsko.

packages/compiler-cli/test/compliance/r3_view_compiler_styling_spec.ts
```diff
@@ -1461,6 +1456,351 @@ describe('compiler compliance: styling', () => {
...
+        },
+        encapsulation: 2
+      });
+     `;
+
+      const result = compile(files, angularFiles);
+      expectEmit(result.source, template, 'Incorrect template');
+    });
+
+    it('should break into multiple chains if there are other styling instructions inbetween',
```
kara:
```suggestion
    it('should break into multiple chains if there are other styling instructions in between',
```

packages/compiler-cli/test/compliance/r3_view_compiler_styling_spec.ts
```diff
@@ -1461,6 +1456,351 @@ describe('compiler compliance: styling', () => {
...
+              },
+              encapsulation: 2
+            });
+          `;
+
+         const result = compile(files, angularFiles);
+         expectEmit(result.source, template, 'Incorrect template');
+       });
+
+    it('should break into multiple chains if there are other styling interpolation instructions inbetween',
```
kara:
```suggestion
    it('should break into multiple chains if there are other styling interpolation instructions in between',
```

kara (approved):
LGTM aside from typos. Note that we should be careful with the G3 presubmit for this because we are changing binding order in some cases, which could break some G3 targets.

packages/compiler/src/render3/view/styling_builder.ts
matsko: ah I see. I missed the `reference` member. This is fine.

matsko (approved):

matsko:
presubmit = http://test/OCL:281116837:BASE:281116883:1574105974370:89ff78f9

ivy presubmit = https://test.corp.google.com/ui#id=OCL:281116837:BASE:281117274:1574106065978:3e19703d

PR #33838, opened 2019-11-15
fix: should not tick nested timer in fakeAsync test

### LINKED ISSUE #33799: FakeAsync's tick flushs nested timeouts
# 🐞 bug report

### Affected Package

### Is this a regression?

dunno

### Description

`tick` call flushs even nested timeouts that have not been installed at the call moment

## 🔬 Minimal Reproduction

```ts
function waitTwice(callback: () => any): void {
  window.setTimeout(() =>
    window.setTimeout(() => callback())
  );
}

describe('fakeAsync', () => {
  it('nested timers', fakeAsync(() => {
    const callback = jasmine.createSpy('callback');

    waitTwice(callback);

    expect(callback).not.toHaveBeenCalled();

    tick();

    expect(callback).not.toHaveBeenCalled();
  }));
});
```

repo with this test

- `git clone git@github.com:artaommahe/fakeasync-nested-timeout.git`
- `yarn install`
- `yarn test`

[test file](https://github.com/artaommahe/fakeasync-nested-timeout/blob/master/src/app/fakeasync.spec.ts)

## 🔥 Exception or Error
```
Chrome 78.0.3904 (Windows 10.0.0) fakeAsync nested timers FAILED
 Error: Expected spy callback not to have been called.
```

## 🌍  Your Environment

**Angular Version:**
<pre><code>

Angular CLI: 8.3.18
Node: 10.16.1
OS: win32 x64
Angular: 8.2.13
... router

@angular-devkit/architect         0.803.18
@angular-devkit/build-angular     0.803.18
@angular-devkit/build-optimizer   0.803.18
@angular-devkit/build-webpack     0.803.18
@angular-devkit/core              8.3.18
@angular-devkit/schematics        8.3.18
@angular/cli                      8.3.18
@ngtools/webpack                  8.3.18
@schematics/angular               8.3.18
@schematics/update                0.803.18

</code></pre>

**Anything else relevant?**

Same test works well with jest's timer functions

```ts
jest.useFakeTimers();
const callback = jest.fn();
waitTwice(callback);

jest.runOnlyPendingTimers();

expect(callback).not.toHaveBeenCalled();

jest.runOnlyPendingTimers();

expect(callback).toHaveBeenCalled();
```

It is a real world case (function with nested zero-timeouts) - we are forced to use this due to parent-child init hooks order for custom elements and for manual elements creation (creating components tree from string markup) when parent's ngAfterViewInit is called sooner than child's one.

JiaLiPassion:
Close #33799

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #33799
```
function waitTwice(callback: () => any): void {
  setTimeout(() =>
    setTimeout(() => callback())
  );
}

describe('fakeAsync', () => {
  it('nested timers', fakeAsync(() => {
    const callback = jasmine.createSpy('callback');
    waitTwice(callback);
    expect(callback).not.toHaveBeenCalled();
    tick();
    expect(callback).toHaveBeenCalled();
    tick();
    expect(callback).toHaveBeenCalled();
  }));
});
```

By default, 
The `callback` will be called after the `first tick`.

```
describe('fakeAsync', () => {
  it('nested timers', fakeAsync(() => {
    const callback = jasmine.createSpy('callback');
    waitTwice(callback);
    expect(callback).not.toHaveBeenCalled();
    tick(0, true);
    expect(callback).not.toHaveBeenCalled();
    tick();
    expect(callback).toHaveBeenCalled();
  }));
});
```
````
## What is the new behavior?
with `tick(0, true)`, the second parameter is `ignoreNestedTimeout`, after the `first tick`, the `nested timeout` will not be invoked.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/testing/src/fake_async.ts
+++ b/packages/core/testing/src/fake_async.ts
@@ -62,13 +62,54 @@ export function fakeAsync(fn: Function): (...args: any[]) => any {
  *
  * {@example core/testing/ts/fake_async.ts region='basic'}
  *
+ * @param millis, the number of millisecond to advance the virtual timer
+ * @param tickOptions, the options of tick with a flag called
+ * processNewMacroTasksSynchronously, whether to invoke the new macroTasks, by default is
+ * false, means the new macroTasks will be invoked
+ *
+ * For example,
+ *
+ * it ('test with nested setTimeout', fakeAsync(() => {
+ *   let nestedTimeoutInvoked = false;
+ *   function funcWithNestedTimeout() {
+ *     setTimeout(() => {
+ *       nestedTimeoutInvoked = true;
+ *     });
+ *   };
+ *   setTimeout(funcWithNestedTimeout);
+ *   tick();
+ *   expect(nestedTimeoutInvoked).toBe(true);
+ * }));
+ *
+ * in this case, we have a nested timeout (new macroTask), when we tick, both the
+ * funcWithNestedTimeout and the nested timeout both will be invoked.
+ *
+ * it ('test with nested setTimeout', fakeAsync(() => {
+ *   let nestedTimeoutInvoked = false;
+ *   function funcWithNestedTimeout() {
+ *     setTimeout(() => {
+ *       nestedTimeoutInvoked = true;
+ *     });
+ *   };
+ *   setTimeout(funcWithNestedTimeout);
+ *   tick(0, {processNewMacroTasksSynchronously: false});
+ *   expect(nestedTimeoutInvoked).toBe(false);
+ * }));
+ *
+ * if we pass the tickOptions with processNewMacroTasksSynchronously to be false, the nested timeout
+ * will not be invoked.
+ *
+ *
  * @publicApi
  */
-export function tick(millis: number = 0): void {
+export function tick(
+    millis: number = 0, tickOptions: {processNewMacroTasksSynchronously: boolean} = {
+      processNewMacroTasksSynchronously: true
+    }): void {
   if (fakeAsyncTestModule) {
-    return fakeAsyncTestModule.tick(millis);
+    return fakeAsyncTestModule.tick(millis, tickOptions);
   } else {
-    return tickFallback(millis);
+    return tickFallback(millis, tickOptions);
   }
 }
 
--- a/packages/zone.js/lib/zone-spec/fake-async-test.ts
+++ b/packages/zone.js/lib/zone-spec/fake-async-test.ts
@@ -72,6 +72,8 @@
     private _currentTime: number = 0;
     // Current real time in millis.
     private _currentRealTime: number = OriginalDate.now();
+    // track requeuePeriodicTimer
+    private _currentTickRequeuePeriodicEntries: any[] = [];
 
     constructor() {}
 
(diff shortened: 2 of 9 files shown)

### CODE REVIEW

**Comment by mhevery** [comment]
[presubmit](http://test/OCL:281123086:BASE:281123154:1574107580499:df5f311f)

File: packages/zone.js/lib/zone-spec/fake-async-test.ts [hunk 1]
```diff
@@ -83,7 +85,8 @@
 
     scheduleFunction(
         cb: Function, delay: number, args: any[] = [], isPeriodic: boolean = false,
-        isRequestAnimationFrame: boolean = false, id: number = -1): number {
+        isRequestAnimationFrame: boolean = false, id: number = -1,
```
mhevery: these function args are getting too long and `id` is inside of the `isRequestAnimationFrame` and `isRequeuePeriodic`. What do you think about refactoring it to have optional named arguments instead?

mhevery:
So this change is breaky in `g3` where 30+ targets fail just in blueprint.

mhevery (commented):
could we add some API docs to the change?

petebacondarwin (commented):
The last "`fix: format`" commit should be a fixup commit for the previous commit

JiaLiPassion:
@mhevery , @petebacondarwin , thank you for the review, I have updated the PR, use new API instead of breaking changes. Please review again, thanks!

petebacondarwin (approved):
LGTM

packages/zone.js/lib/zone-spec/fake-async-test.ts
JiaLiPassion: @mhevery , yeah, you are right, in the updated PR, I refactored the parameters to optional named arguments object.

packages/core/testing/src/fake_async.ts
```diff
@@ -62,13 +62,17 @@ export function fakeAsync(fn: Function): (...args: any[]) => any {
...
  * {@example core/testing/ts/fake_async.ts region='basic'}
  *
+ * @param millis, the number of millisecond to advance the virtual timer
+ * @param ignoreNestedTimeout, whether to invoke the nested timeout, by default is false,
+ *                             means the nested timeout will be triggered
+ *
  * @publicApi
  */
-export function tick(millis: number = 0): void {
+export function tick(millis: number = 0, ignoreNestedTimeout = false): void {
```
mhevery: I fear that over time we will have more of these arguments. Could we turn the options into object literal? That way adding new options in the future will not be an issue.?

packages/core/testing/src/fake_async.ts
```diff
@@ -62,13 +62,17 @@ export function fakeAsync(fn: Function): (...args: any[]) => any {
  *
  * {@example core/testing/ts/fake_async.ts region='basic'}
  *
+ * @param millis, the number of millisecond to advance the virtual timer
+ * @param ignoreNestedTimeout, whether to invoke the nested timeout, by default is false,
```
mhevery: Could you add more text here. Do you mean if the nested timeout will be flushed in the same tick or separately?

mhevery:
[presubmit](http://test/OCL:292587640:BASE:292587675:1580500248141:56894f3f)

packages/core/testing/src/fake_async.ts
mhevery: Also I think the name should be `ignoreNestedTimeout` => `processNewMacroTasksSynchronously` (Open to other names)  What do you think. If I understand correctly, this is not specific to `setTimeout` but rather to the fact that if a macro task is scheduled in the current flush cycle, wether or not it should be flushed synchronously.

mhevery (commented):
Minor suggestions. Running presubmit now.

packages/core/testing/src/fake_async.ts
JiaLiPassion: yes, you are right, it is not only `setTimeout`, the `processNewMacroTasksSynchronously` is a better name, thanks.

packages/core/testing/src/fake_async.ts
JiaLiPassion: Yes, I have added more text and sample codes. Thanks.

mhevery (approved):

mhevery:
[presubmit](http://test/OCL:292587640:BASE:294463027:1581444275652:c9c782ac)

mhevery:
[presubmit](http://test/OCL:292587640:BASE:296256842:1582226671611:88bd667)

PR #33840, opened 2019-11-15
build: fix build scripts on macOS

gkalpak:
In #33823, `scripts/package-builds.sh` (which is used by both `build-packages-dist.sh` and `build-ivy-npm-packages.sh`) was updated to use `realpath`. It turns out that `realpath` does not exist on macOS, so the build scripts do not work there.

In order to fix this (and also reduce the likelihood of introducing similar issues in the future), this commit changes these bash scripts to Node.js scripts (using [ShellJS](https://github.com/shelljs/shelljs) for a cross-platform implementation of Unix shell commands where necessary).


gkalpak:
Superceded by #33854.

PR #33854, opened 2019-11-15
build: fix build scripts on macOS

gkalpak:
In #33823, `scripts/package-builds.sh` (which is used by both `build-packages-dist.sh` and `build-ivy-npm-packages.sh`) was updated to use `realpath`. It turns out that `realpath` does not exist on macOS, so the build scripts do not work there.

In order to fix this (and also reduce the likelihood of introducing similar issues in the future), this commit changes these bash scripts to Node.js scripts (using [ShellJS](https://github.com/shelljs/shelljs) for a cross-platform implementation of Unix shell commands where necessary).

##
<sub>
This is the same as #33840. Opened a new PR, because #33840 was cluttered with irrelevant preview comments and force-pushes, while I was investigating.
</sub>


petebacondarwin (approved):

scripts/package-builder.js
```diff
@@ -0,0 +1,118 @@
...
+  if (!existsSync(absDestPath)) mkdir('-p', absDestPath);
+
+  targets.forEach(target => {
+    const pkg = target.replace(/\/\/packages\/(.*):npm_package/, '$1');
+
+    // Skip any that don't have an "npm_package" target.
+    const srcDir = `${bazelBin}/packages/${pkg}/npm_package`;
+    const destDir = `${absDestPath}/${pkg}`;
+
+    if (existsSync(srcDir) && statSync(srcDir).isDirectory()) {
```
devversion: Optional idea: You could also use `shelljs.test('-d', srcDir)` (same for other calls to `existsSync`).

devversion (approved):
Nice LGTM!

IgorMinar:
@gkalpak did you say that this change came with a significant performance penalty for running bazel? can you please clarify / share details?

IgorMinar:
ah.. nevermind. I see that the issue is now resolved.

IgorMinar:
@gkalpak can you please rebase?

IgorMinar:
@gkalpak nvm, I rebased it for you

IgorMinar (approved):
Awesome! thank you @gkalpak

scripts/package-builder.js
gkalpak: Good catch, I had forgitten about `test()`.
(Too bad I saw this too late 😞)

gkalpak: I ended up making the change in #33895 (because it keeps the code closer to the original bash script and might be easier for people to follow) :smiley:

PR #33856, opened 2019-11-15
ɵɵtemplate instruction refactor

pkozlowski-opensource:
This refactoring clearly separates the first and subsequent creation execution
of the `template` instruction. This approach has the following benefits:
- it is clear what happens during the first vs. subsequent executions;
- we can avoid several memory reads and checks after the first creation pass
(there is measurable performance improvement on various benchmarks);
- the template instructions becomes smaller and should become a candidate
for optimisations / inlining faster;

### DIFF
--- a/packages/core/src/render3/instructions/container.ts
+++ b/packages/core/src/render3/instructions/container.ts
@@ -6,14 +6,14 @@
  * found in the LICENSE file at https://angular.io/license
  */
 import {assertDataInRange, assertEqual} from '../../util/assert';
-import {assertHasParent} from '../assert';
+import {assertFirstCreatePass, assertHasParent} from '../assert';
 import {attachPatchData} from '../context_discovery';
 import {executeCheckHooks, executeInitAndCheckHooks, incrementInitPhaseFlags, registerPostOrderHooks} from '../hooks';
 import {ACTIVE_INDEX, CONTAINER_HEADER_OFFSET, LContainer} from '../interfaces/container';
 import {ComponentTemplate} from '../interfaces/definition';
 import {LocalRefExtractor, TAttributes, TContainerNode, TNode, TNodeType, TViewNode} from '../interfaces/node';
 import {isDirectiveHost} from '../interfaces/type_checks';
-import {FLAGS, HEADER_OFFSET, InitPhaseState, LView, LViewFlags, RENDERER, TVIEW, TViewType, T_HOST} from '../interfaces/view';
+import {FLAGS, HEADER_OFFSET, InitPhaseState, LView, LViewFlags, RENDERER, TVIEW, TView, TViewType, T_HOST} from '../interfaces/view';
 import {assertNodeType} from '../node_assert';
 import {appendChild, removeView} from '../node_manipulation';
 import {getBindingIndex, getCheckNoChangesMode, getIsParent, getLView, getPreviousOrParentTNode, setIsNotParent, setPreviousOrParentTNode} from '../state';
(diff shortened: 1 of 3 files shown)


packages/core/src/render3/instructions/container.ts
```diff
@@ -69,38 +99,29 @@ export function ɵɵtemplate(
...
-    const embeddedTViewNode = createTNode(tView, null, TNodeType.View, -1, null, null) as TViewNode;
-    embeddedTViewNode.injectorIndex = tContainerNode.injectorIndex;
-    embeddedTView.node = embeddedTViewNode;
-
-    if (tView.queries !== null) {
-      tView.queries.template(tView, tContainerNode);
-      embeddedTView.queries = tView.queries.embeddedTView(tContainerNode);
-    }
-  }
+  const tNode = tView.firstCreatePass ?
```
mhevery: I ❤️ this, it is so clear!

packages/core/src/render3/instructions/container.ts
```diff
@@ -69,38 +99,29 @@ export function ɵɵtemplate(
...
-    createDirectivesInstances(tView, lView, tContainerNode);
-  }
-  if (localRefs != null) {
-    saveResolvedLocalsInData(lView, tContainerNode, localRefExtractor);
+  const comment = lView[RENDERER].createComment(ngDevMode ? 'container' : '');
+  appendChild(comment, tNode, lView);
+  attachPatchData(comment, lView);
+
+  // Containers are added to the current view tree instead of their embedded views
+  // because views can be removed and re-inserted.
```
mhevery: This comment is not clear to me. Could you add an example of why this is needed.

packages/core/src/render3/instructions/container.ts
```diff
@@ -44,6 +44,36 @@ export function ɵɵcontainer(index: number): void {
...
 }
 
+function templateFirstCreatePass(
+    index: number, tView: TView, lView: LView, templateFn: ComponentTemplate<any>| null,
+    decls: number, vars: number, tagName?: string | null, attrsIndex?: number | null,
+    localRefsIndex?: number | null): TContainerNode {
+  ngDevMode && assertFirstCreatePass(tView);
+  ngDevMode && ngDevMode.firstCreatePass++;
+  const tViewConsts = tView.consts;
+  // TODO(pk): refactor getOrCreateTNode to have the "create" only version
```
mhevery: 👍

mhevery (commented):
I ❤️ this change. Yes we should do more of this. My only nit is to make the text of the comment more clear as it is not clear too me what you are trying to say. I find that examples tend to make things more clear.

packages/core/src/render3/instructions/container.ts
pkozlowski-opensource: Ha ha, I _really_ need to be more careful with coping existing comments :-)
Actually I've read through it only now and I don't think this comment is relevant / useful any more - I will delete it.

mhevery (approved):

mhevery:
[presubmit](http://test/OCL:281115811:BASE:281115869:1574105710500:88db5951)
[ivy presubmit](http://test/OCL:281115811:BASE:281116108:1574105778589:7bbbc378)

kara (approved):
LGTM

pkozlowski-opensource:
merge-assistance due to "Pending — Status "google3" is pending"

AndrewKushnir:
@pkozlowski-opensource VE and Ivy presubmits are successful, I updated the status of the g3 check. Thank you.

PR #33857, opened 2019-11-15
ngcc: handle localized messages in ES5

petebacondarwin:
Recently the ngtsc translator was modified to be more `ScriptTarget`
aware, which basically means that it will not generate non-ES5 code
when the output format is ES5 or similar.

This commit enhances that change by also "downleveling" localized
messages. In ES2015 the messages use tagged template literals, which
are not available in ES5.

### DIFF
--- a/packages/compiler-cli/src/ngtsc/translator/src/translator.ts
+++ b/packages/compiler-cli/src/ngtsc/translator/src/translator.ts
@@ -262,13 +262,9 @@ class ExpressionTranslatorVisitor implements ExpressionVisitor, StatementVisitor
   }
 
   visitLocalizedString(ast: LocalizedString, context: Context): ts.Expression {
-    if (this.scriptTarget < ts.ScriptTarget.ES2015) {
-      // This should never happen.
-      throw new Error(
-          'Unsupported mode: Visiting a localized string (which produces a tagged template ' +
-          `literal) ' while targeting ${ts.ScriptTarget[this.scriptTarget]}.`);
-    }
-    return visitLocalizedString(ast, context, this);
+    return this.scriptTarget >= ts.ScriptTarget.ES2015 ?
+        createLocalizedStringTaggedTemplate(ast, context, this) :
+        createLocalizedStringFunctionCall(ast, context, this, this.imports);
   }
 
   visitExternalExpr(ast: ExternalExpr, context: Context): ts.PropertyAccessExpression
@@ -467,7 +463,7 @@ export class TypeTranslatorVisitor implements ExpressionVisitor, TypeVisitor {
   }
 
   visitLocalizedString(ast: LocalizedString, context: Context): ts.Expression {
-    return visitLocalizedString(ast, context, this);
+    throw new Error('Method not implemented.');
   }
 
   visitExternalExpr(ast: ExternalExpr, context: Context): ts.TypeNode {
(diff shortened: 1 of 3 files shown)


packages/compiler-cli/src/ngtsc/translator/src/translator.ts
```diff
@@ -458,7 +454,7 @@ export class TypeTranslatorVisitor implements ExpressionVisitor, TypeVisitor {
   }
 
   visitLocalizedString(ast: LocalizedString, context: Context): ts.Expression {
-    return visitLocalizedString(ast, context, this);
+    throw new Error('Method not implemented.');
```
petebacondarwin: I realised that we never actually call this because it is in a visitor that only cares about types.

gkalpak (approved):

packages/compiler-cli/src/ngtsc/translator/src/translator.ts
```diff
@@ -566,3 +563,32 @@ function visitLocalizedString(ast: LocalizedString, context: Context, visitor: E
...
+
+/**
+ * Translate the `LocalizedString` node into a `$localize` call using the imported
+ * `__makeTemplateObject` helper for ES5 formatted output.
+ */
+function createLocalizedStringFunctionCall(
+    ast: LocalizedString, context: Context, visitor: ExpressionVisitor, imports: ImportManager) {
+  const messageParts = [ast.serializeI18nHead()];
+  const expressions: any[] = [];
+  for (let i = 1; i < ast.messageParts.length; i++) {
```
alxhub: Can you add a comment explaining this loop? I'm not quite getting it. Iteration starts at `1` but we read from `ast.expressions` starting at `0`, and then increment by 1 so each index takes its turn as `expressions` or `messageParts`?

It feels like this loop was meant to count by `2`s instead?

packages/compiler-cli/src/ngtsc/translator/src/translator.ts
```diff
@@ -566,3 +563,32 @@ function visitLocalizedString(ast: LocalizedString, context: Context, visitor: E
...
+    expressions.push(ast.expressions[i - 1].visitExpression(visitor, context));
+    messageParts.push(ast.serializeI18nTemplatePart(i));
+  }
+  const {moduleImport, symbol} = imports.generateNamedImport('tslib', '__makeTemplateObject');
+  const __makeTemplateObjectHelper = (moduleImport === null) ?
+      ts.createIdentifier(symbol) :
+      ts.createPropertyAccess(ts.createIdentifier(moduleImport), ts.createIdentifier(symbol));
+  return ts.createCall(ts.createIdentifier('$localize'), undefined, [
```
alxhub: For the `ts.create*` methods, I've taken to adding a comment with each parameter name:
```
ts.createCall(
  /* expression */ ts.createIdentifier('$localize'),
  /* typeArguments */ undefined,
  /* argumentsArray */ [...],
);
```

alxhub (approved):
LGTM with two comments.

alxhub:
[Presubmit](https://test.corp.google.com/ui#id=OCL:281767361:BASE:281767409:1574357747941:e59beffc)

packages/compiler-cli/src/ngtsc/translator/src/translator.ts
petebacondarwin: Added comments to explain the algorithm

alxhub: 👍 thanks, that makes a lot more sense to me!

PR #33860, opened 2019-11-15
fix(language-service): Insert parentheses for method completion

### LINKED ISSUE #15: Completion of method should add parenthesis
When using completion of the `(click)` event for example to complete a method. No parenthesis are added by default to the method.

For example:

image
image

After pressing `TAB`:

image

It would be nice if we could have parenthesis added by default and the cursor inside of these parenthesis like:

image

kyliau:
This commit leverages the `insertText` field in `ts.CompletionEntry` to
return a completion text for class methods that includes parentheses.

PR closes https://github.com/angular/vscode-ng-language-service/issues/15

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/language-service/test/completions_spec.ts
+++ b/packages/language-service/test/completions_spec.ts
@@ -213,6 +213,18 @@ describe('completions', () => {
     expectContain(completions, CompletionKind.METHOD, ['myClick']);
   });
 
+  it('for methods should include parentheses', () => {
+    mockHost.override(TEST_TEMPLATE, `<div (click)="~{cursor}"></div>`);
+    const marker = mockHost.getLocationMarkerFor(TEST_TEMPLATE, 'cursor');
+    const completions = ngLS.getCompletionsAt(TEST_TEMPLATE, marker.start);
+    expect(completions).toBeDefined();
+    expect(completions !.entries).toContain(jasmine.objectContaining({
+      name: 'myClick',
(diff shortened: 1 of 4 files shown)


packages/language-service/test/completions_spec.ts
```diff
@@ -161,6 +161,18 @@ describe('completions', () => {
     expectContain(completions, CompletionKind.METHOD, ['$any']);
   });
 
+  it('for methods should include parentheses', () => {
```
ayazhafiz:
```suggestion
  it('for methods that should include parentheses', () => {
```

ayazhafiz (changes requested):
Is it possible to specify that next insertion index for the user should be between the parentheses?

packages/language-service/test/completions_spec.ts
kyliau: rephrased to "should include parentheses for methods"

kyliau:
> Is it possible to specify that next insertion index for the user should be between the parentheses?

I don't think it is possible to change the user's cursor. That would be bad user experience, as it would catch users by surprise.

ayazhafiz (approved):

josephperrott (approved):
LGTM

PR #31932, opened 2019-07-31
fix(ivy): reuse compilation scope for incremental template changes.

petebacondarwin:
Previously if only a component template changed then we would know to
rebuild its component source file. But the compilation was incorrect if the
component was part of an NgModule, since we were not capturing the
compilation scope information that had a been acquired from the NgModule
and was not being regenerated since we were not needing to recompile
the NgModule.

Now we register compilation scope information for each component, via the
`ComponentScopeRegistry` interface, so that it is available for incremental
compilation.

The `ComponentDecoratorHandler` now reads the compilation scope from a
`ComponentScopeReader` interface which is implemented as a compound
reader composed of the original `LocalModuleScopeRegistry` and the
`IncrementalState`.

Fixes #31654

PR #33862, opened 2019-11-15
fix(ivy): always re-analyze the program during incremental rebuilds

alxhub:
Previously, the ngtsc compiler attempted to reuse analysis work from the
previous program during an incremental build. To do this, it had to prove
that the work was safe to reuse - that no changes made to the new program
would invalidate the previous analysis.

The implementation of this had a significant design flaw: if the previous
program had errors, the previous analysis would be missing significant
information, and the dependency graph extracted from it would not be
sufficient to determine which files should be re-analyzed to fill in the
gaps. This often meant that the build output after an error was resolved
would be wholly incorrect.

This commit switches ngtsc to take a simpler approach to incremental
rebuilds. Instead of attempting to reuse prior analysis work, the entire
program is re-analyzed with each compilation. This is actually not as
expensive as one might imagine - analysis is a fairly small part of overall
compilation time.

Based on the dependency graph extracted during this analysis, the compiler
then can make accurate decisions on whether to emit specific files. A new
suite of tests is added to validate behavior in the presence of source code
level errors.

This new approach is dramatically simpler than the previous algorithm, and
should always produce correct results for a semantically correct program.

### DIFF
--- a/packages/compiler-cli/src/main.ts
+++ b/packages/compiler-cli/src/main.ts
@@ -56,14 +56,33 @@ export function main(
 }
 
 export function mainDiagnosticsForTest(
-    args: string[], config?: NgcParsedConfiguration): ReadonlyArray<ts.Diagnostic|api.Diagnostic> {
+    args: string[], config?: NgcParsedConfiguration,
+    programReuse?: {program: api.Program | undefined},
+    modifiedResourceFiles?: Set<string>| null): ReadonlyArray<ts.Diagnostic|api.Diagnostic> {
   let {project, rootNames, options, errors: configErrors, watch, emitFlags} =
       config || readNgcCommandLineAndConfiguration(args);
   if (configErrors.length) {
     return configErrors;
   }
-  const {diagnostics: compileDiags} = performCompilation(
-      {rootNames, options, emitFlags, emitCallback: createEmitCallback(options)});
+
+  let oldProgram: api.Program|undefined;
+  if (programReuse !== undefined) {
+    oldProgram = programReuse.program;
+  }
+
+  const {diagnostics: compileDiags, program} = performCompilation({
+    rootNames,
+    options,
+    emitFlags,
+    oldProgram,
+    modifiedResourceFiles,
+    emitCallback: createEmitCallback(options),
+  });
+
+  if (programReuse !== undefined) {
+    programReuse.program = program;
+  }
+
   return compileDiags;
 }
 
--- a/packages/compiler-cli/src/main.ts
+++ b/packages/compiler-cli/src/main.ts
@@ -223,26 +223,34 @@ function reportErrorsAndExit(
     allDiagnostics: Diagnostics, options?: api.CompilerOptions,
     consoleError: (s: string) => void = console.error): number {
   const errorsAndWarnings = filterErrorsAndWarnings(allDiagnostics);
-  if (errorsAndWarnings.length) {
-    const formatHost = getFormatDiagnosticsHost(options);
-    if (options && options.enableIvy !== false) {
-      const ngDiagnostics = errorsAndWarnings.filter(api.isNgDiagnostic);
-      const tsDiagnostics = errorsAndWarnings.filter(api.isTsDiagnostic);
-      consoleError(replaceTsWithNgInErrors(
-          ts.formatDiagnosticsWithColorAndContext(tsDiagnostics, formatHost)));
-      consoleError(formatDiagnostics(ngDiagnostics, formatHost));
-    } else {
-      consoleError(formatDiagnostics(errorsAndWarnings, formatHost));
-    }
-  }
+  printDiagnostics(errorsAndWarnings, options, consoleError);
   return exitCodeFromResult(allDiagnostics);
 }
 
 export function watchMode(
     project: string, options: api.CompilerOptions, consoleError: (s: string) => void) {
   return performWatchCompilation(createPerformWatchHost(project, diagnostics => {
-    consoleError(formatDiagnostics(diagnostics, getFormatDiagnosticsHost(options)));
+    printDiagnostics(diagnostics, options, consoleError);
   }, options, options => createEmitCallback(options)));
+}
+
+function printDiagnostics(
+    diagnostics: ReadonlyArray<ts.Diagnostic|api.Diagnostic>,
+    options: api.CompilerOptions | undefined, consoleError: (s: string) => void): void {
+  if (diagnostics.length === 0) {
+    return;
+  }
+
+  const formatHost = getFormatDiagnosticsHost(options);
+  if (options && options.enableIvy !== false) {
+    const ngDiagnostics = diagnostics.filter(api.isNgDiagnostic);
+    const tsDiagnostics = diagnostics.filter(api.isTsDiagnostic);
+    consoleError(replaceTsWithNgInErrors(
+        ts.formatDiagnosticsWithColorAndContext(tsDiagnostics, formatHost)));
+    consoleError(formatDiagnostics(ngDiagnostics, formatHost));
+  } else {
+    consoleError(formatDiagnostics(diagnostics, formatHost));
+  }
 }
 
 // CLI entry point
--- a/packages/compiler-cli/src/ngtsc/incremental/src/README.md
+++ b/packages/compiler-cli/src/ngtsc/incremental/src/README.md
@@ -2,7 +2,7 @@
 
 This package contains logic related to incremental compilation in ngtsc.
 
-In particular, it tracks metadata for `ts.SourceFile`s in between compilations, so the compiler can make intelligent decisions about when to skip certain operations and rely on previously analyzed data.
+In particular, it tracks dependencies between `ts.SourceFile`s, so the compiler can make intelligent decisions about when it's safe to skip certain operations.
 
 # How does incremental compilation work?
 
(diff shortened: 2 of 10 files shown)


packages/compiler-cli/src/ngtsc/incremental/src/README.md
```diff
@@ -2,7 +2,7 @@
 
 This package contains logic related to incremental compilation in ngtsc.
 
-In particular, it tracks metadata for `ts.SourceFile`s in between compilations, so the compiler can make intelligent decisions about when to skip certain operations and rely on previously analyzed data.
+In particular, it tracks dependencies between `ts.SourceFile`s, so the compiler can make intelligent decisions about when it's safe to skip certain operations.
```
mhevery: Thanks for having the doc as well as keeping it updated.

mhevery (approved):

packages/compiler-cli/src/main.ts
```diff
@@ -56,14 +56,33 @@ export function main(
 }
 
 export function mainDiagnosticsForTest(
-    args: string[], config?: NgcParsedConfiguration): ReadonlyArray<ts.Diagnostic|api.Diagnostic> {
+    args: string[], config?: NgcParsedConfiguration,
+    programReuse?: {program: api.Program | undefined},
```
JoostK: Why is this wrapped into an object literal? It seems this can just be:

```suggestion
    programReuse?: api.Program,
```

packages/compiler-cli/src/main.ts
```diff
@@ -204,28 +223,36 @@ function reportErrorsAndExit(
...
-  }
+  printDiagnostics(errorsAndWarnings, options, consoleError);
   return exitCodeFromResult(allDiagnostics);
 }
 
 export function watchMode(
     project: string, options: api.CompilerOptions, consoleError: (s: string) => void) {
   return performWatchCompilation(createPerformWatchHost(project, diagnostics => {
-    consoleError(formatDiagnostics(diagnostics, getFormatDiagnosticsHost(options)));
+    printDiagnostics(diagnostics, options, consoleError);
```
JoostK: Nice!

packages/compiler-cli/src/ngtsc/incremental/src/state.ts
```diff
@@ -8,47 +8,33 @@
...
-import {ClassDeclaration} from '../../reflection';
-import {ComponentScopeReader, ComponentScopeRegistry, LocalModuleScope} from '../../scope';
 import {ResourceDependencyRecorder} from '../../util/src/resource_recorder';
 
 /**
  * Accumulates state between compilations.
  */
-export class IncrementalState implements DependencyTracker, MetadataReader, MetadataRegistry,
-    ResourceDependencyRecorder, ComponentScopeRegistry, ComponentScopeReader {
+export class IncrementalState implements DependencyTracker, ResourceDependencyRecorder {
```
JoostK: The `ComponentScopeRegistry` interface is no longer really used now.
- The `registerComponentScope` method was only for `IncrementalState`, so should be removed.
- The `setComponentAsRequiringRemoteScoping` method it describes is not formally implemented; it's only `LocalModuleScopeRegistry` that implements this method without explicitly implementing `ComponentScopeRegistry`, as the caller of this method knows that it's dealing with `LocalModuleScopeRegistry`.

By looking at the history of `ComponentScopeRegistry`, it was introduced to be able to reuse compilation scopes during incremental compilations in #31932, so we can probably just remove it.

packages/compiler-cli/src/ngtsc/program.ts
```diff
@@ -558,13 +555,12 @@ export class NgtscProgram implements api.Program {
     const evaluator = new PartialEvaluator(this.reflector, checker, this.incrementalState);
     const dtsReader = new DtsMetadataReader(checker, this.reflector);
     const localMetaRegistry = new LocalMetadataRegistry();
-    const localMetaReader = new CompoundMetadataReader([localMetaRegistry, this.incrementalState]);
+    const localMetaReader = new CompoundMetadataReader([localMetaRegistry]);
```
JoostK: The `CompoundMetadataReader` is no longer necessary for a single delegate.

packages/compiler-cli/src/ngtsc/program.ts
```diff
@@ -558,13 +555,12 @@ export class NgtscProgram implements api.Program {
...
     const depScopeReader = new MetadataDtsModuleScopeResolver(dtsReader, this.aliasingHost);
     const scopeRegistry = new LocalModuleScopeRegistry(
-        localMetaReader, depScopeReader, this.refEmitter, this.aliasingHost, this.incrementalState);
-    const scopeReader = new CompoundComponentScopeReader([scopeRegistry, this.incrementalState]);
-    const metaRegistry =
-        new CompoundMetadataRegistry([localMetaRegistry, scopeRegistry, this.incrementalState]);
+        localMetaReader, depScopeReader, this.refEmitter, this.aliasingHost);
+    const scopeReader = new CompoundComponentScopeReader([scopeRegistry]);
```
JoostK: The `CompoundComponentScopeReader` can be skipped, as there's just a single delegate.

packages/compiler-cli/test/ngtsc/env.ts
```diff
@@ -198,7 +199,19 @@ export class NgtscTestEnvironment {
...
   driveDiagnostics(): ReadonlyArray<ts.Diagnostic> {
     // ngtsc only produces ts.Diagnostic messages.
-    return mainDiagnosticsForTest(['-p', this.basePath]) as ts.Diagnostic[];
+    let reuseProgram: {program: Program | undefined}|undefined = undefined;
+    if (this.multiCompileHostExt !== null) {
+      reuseProgram = {
+        program: this.oldProgram || undefined,
+      };
+    }
+    const diags: ts.Diagnostic[] = (mainDiagnosticsForTest(
```
JoostK: The parens and `as ts.Diagnostic[]` cast (or the explicit type for `diags`) looks superfluous to me.

packages/compiler-cli/test/ngtsc/incremental_error_spec.ts
```diff
@@ -0,0 +1,297 @@
...
+      `);
+    });
+
+    afterEach(() => {
+      // Assert that the unrelated.ts file above, which is not changed by any of the test, is never
+      // re-emitted in the last build for each test.
+      expect(env.getFilesWrittenSinceLastFlush()).not.toContain('/unrelated.js');
+    });
+
+    function expectToHaveWritten(files: string[]): void {
```
JoostK: +1 for this helper!

JoostK: With one request: I would make this function also verify that `set \ files  = ∅`.

packages/compiler-cli/test/ngtsc/incremental_error_spec.ts
```diff
@@ -0,0 +1,297 @@
...
+      env.write('test.ts', `
+        import {Component} from '@angular/core';
+
+        @Component({selector: 'test-cmp', template: ...})
+        export class TestCmp {}
+      `);
+      env.flushWrittenFileTracking();
+      const diags = env.driveDiagnostics();
+      expect(diags.length).toBeGreaterThan(0);
+      expect(env.getFilesWrittenSinceLastFlush()).not.toContain(_('/test.js'));
```
JoostK: nit: If we make `expectToHaveWritten` an exact set match, this can become `expectToHaveWritten([])`;

...although this may be more explicit.

packages/compiler-cli/test/ngtsc/incremental_error_spec.ts
```diff
@@ -0,0 +1,297 @@
...
+      `);
+
+      env.flushWrittenFileTracking();
+      env.driveMain();
+
+      expectToHaveWritten(['/test.js']);
+    });
+
+    it('should recover from an error in a component that is part of a module', () => {
+      // In this test, there are two components, TestCmp and TargetCmp, that are part of the same
```
JoostK: These comments are really helpful!

packages/compiler-cli/test/ngtsc/incremental_error_spec.ts
```diff
@@ -0,0 +1,297 @@
...
+
+        @Component({selector: ..., template: '...'})
+        export class TestCmp {}
+      `);
+      env.flushWrittenFileTracking();
+      const diags = env.driveDiagnostics();
+      expect(diags.length).toBeGreaterThan(0);
+
+      // No files should have been written.
+      expect(env.getFilesWrittenSinceLastFlush().size).toBe(0);
```
JoostK: Again, this can be cleaned up if `expectToHaveWritten` becomes an exact match.

There's more occurrences of this; I'd argue only `expectToHaveWritten` should call into `env.getFilesWrittenSinceLastFlush`.

packages/compiler-cli/test/ngtsc/incremental_error_spec.ts
```diff
@@ -0,0 +1,297 @@
...
+        import {Component} from '@angular/core';
+
+        @Component({selector: 'test-cmp', template: '...'})
+        export class TestCmp {}
+      `);
+
+      env.flushWrittenFileTracking();
+      env.driveMain();
+
+      const written = env.getFilesWrittenSinceLastFlush();
```
JoostK: Unused.

packages/compiler-cli/test/ngtsc/incremental_error_spec.ts
```diff
@@ -0,0 +1,297 @@
...
+        selector: 'lib-cmp',
+        template: '...',
+      })
+      export class LibCmp {}
+
+      @NgModule({
+        declarations: [LibCmp],
+        exports: [LibCmp],
+      })
+      export class LibModule
```
JoostK: I was looking for the error for a while, so I'd add a comment:

```suggestion
      export class LibModule // missing braces
```

packages/compiler-cli/test/ngtsc/incremental_error_spec.ts
```diff
@@ -0,0 +1,297 @@
...
+      env.tsconfig();
+
+      // This file is part of the program, but not referenced by anything else. It can be used by
+      // each test to verify that it isn't re-emitted after incremental builds.
+      env.write('unrelated.ts', `
+        export class Unrelated {}
+      `);
+    });
+
+    afterEach(() => {
```
JoostK: If we change `expectToHaveWritten` to be exact, this can be dropped.

JoostK (approved):
LGTM once comments have been addressed.

packages/compiler-cli/src/main.ts
alxhub: It's an input/output parameter. We're returning the new program back to the caller via the reference in this object. Not the best API; we'll overhaul it once VE no longer goes through it too.

packages/compiler-cli/test/ngtsc/incremental_error_spec.ts
alxhub: I thought about this, but then we'd have to list every `.d.ts` file as well, which I don't like.

packages/compiler-cli/test/ngtsc/env.ts
alxhub: The parens were to appease `clang-format` which otherwise breaks apart the cast in a way that causes a syntax error. I refactored the code to work around it.

alxhub:
[Presubmit](http://test/OCL:281554012:BASE:281554053:1574275835798:e86d52fe)
[Ivy Presubmit](http://test/OCL:281554012:BASE:281554185:1574275870664:d3ab5596)

PR #33863, opened 2019-11-15
perf(ivy): R3TestBed - Do not process NgModuleDefs that have already …

atscott:
…been processed

I was observing 10x slower test times in Ivy vs VE and this change made Ivy perform as fast as VE. Perf analysis showed that 96% of the time was spent in `queueTypesFromModulesArray`. This can happen with deep trees of imports that have duplicated NgModule imports throughout.

### DIFF
--- a/packages/core/testing/src/r3_test_bed_compiler.ts
+++ b/packages/core/testing/src/r3_test_bed_compiler.ts
@@ -486,6 +486,9 @@ export class R3TestBedCompiler {
   }
 
   private queueTypesFromModulesArray(arr: any[]): void {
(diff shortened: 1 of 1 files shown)


AndrewKushnir:
Thanks for the fix @atscott 👍

I know PR is still in Draft mode, but I'd propose adding some comments on why we do caching (since we may encounter the same NgModules while going though imports and exports, so we want to avoid re-processing). Otherwise - looks great!

AndrewKushnir (approved):

atscott:
[Ivy presubmit](https://test.corp.google.com/ui#id=OCL:281065803:BASE:281334847:1574190675099:57f86f7a)

PR #33875, opened 2019-11-16
fix(ngcc): always add exports for `ModuleWithProviders` references

### LINKED ISSUE #33701: RC.1/IVY - export 'EffectsFeatureModule' (imported as 'i2') was not found in @ngrx/effects
# 🐞 bug report

### Affected Package

  

After upgrading to RC1 and turning on IVY I get tons of errors from  the @ngrx/effects module.

```
"export 'EffectsFeatureModule' (imported as 'i2') was not found in '@ngrx/effects'
"export 'EffectsFeatureModule' (imported as 'i2') was not found in '@ngrx/effects'
```
... and so on.. 

Turning off IVY everything works without an issue.

This is from a production app running NG8 currently in production.
It uses @ngrx/store and effects module.

I've never had these issues with the .next-x and rc.0 versions.

### Is this a regression?

I think so

I believe so.
 I've never seen this issue before and I've tested all .next versions and rc.0

### Description

ng serve yields lots off errors like:

```
ERROR in ./src/app/store/bizz-store/bizz-store.module.ts 19:79-102
"export 'EffectsFeatureModule' (imported as 'i1') was not found in '@ngrx/effects'
ERROR in ./src/app/modules/lazy-shared/lazy-shared.module.ts 136:26-50
"export 'EffectsFeatureModule' (imported as 'i16') was not found in '@ngrx/effects'
ERROR in ./src/app/feature-modules/dashboard/dashboard.module.ts 58:520-544
"export 'EffectsFeatureModule' (imported as 'i16') was not found in '@ngrx/effects'
ERROR in ./src/app/store/iticket-store/iticket-store.module.ts 22:105-128
"export 'EffectsFeatureModule' (imported as 'i2') was not found in '@ngrx/effects'
ERROR in ./src/app/feature-modules/login/store/login-store.module.ts 24:103-126
"export 'EffectsFeatureModule' (imported as 'i2') was not found in '@ngrx/effects'
ERROR in ./src/app/feature-modules/gdpr/gdpr.module.ts 40:149-172
"export 'EffectsFeatureModule' (imported as 'i3') was not found in '@ngrx/effects'
ERROR in ./src/app/modules/licenseplate-checker/licenseplate-checker-module.ts 25:162-185
"export 'EffectsFeatureModule' (imported as 'i3') was not found in '@ngrx/effects'
ERROR in ./src/app/feature-modules/feedback/feedback.module.ts 38:185-208
"export 'EffectsFeatureModule' (imported as 'i3') was not found in '@ngrx/effects'
ERROR in ./src/app/feature-modules/weather/weather.module.ts 38:182-205
"export 'EffectsFeatureModule' (imported as 'i3') was not found in '@ngrx/effects'
ERROR in ./src/app/feature-modules/receipt/receipt.module.ts 39:170-193
"export 'EffectsFeatureModule' (imported as 'i4') was not found in '@ngrx/effects'
```

## 🔬 Minimal Reproduction

Unfortunately I cannot share my repo but if nobody has seen this issue before I'll try and do a minimal version. 

 https://stackblitz.com/...

## 🔥 Exception or Error
<pre><code>

</code></pre>

## 🌍  Your Environment

**Angular Version:**
<pre><code>

Angular CLI: 9.0.0-rc.1
Node: 12.2.0
OS: darwin x64
Angular: 9.0.0-rc.1
... router, service-worker

@angular-devkit/architect          0.900.0-rc.1
@angular-devkit/build-angular      0.900.0-rc.1
@angular-devkit/build-optimizer    0.900.0-rc.1
@angular-devkit/build-webpack      0.900.0-rc.1
@angular-devkit/core               9.0.0-rc.1
@angular-devkit/schematics         9.0.0-rc.1
@angular/cdk                       9.0.0-rc.0
@angular/material                  9.0.0-rc.0
@angular/material-moment-adapter   9.0.0-rc.0
@ngtools/webpack                   9.0.0-rc.1
@schematics/angular                9.0.0-rc.1
@schematics/update                 0.900.0-rc.1
rxjs                               7.0.0-alpha.0
typescript                         3.6.4

</code></pre>

**Anything else relevant?**

JoostK:
Please see individual commits for detailed descriptions.

Fixes #33701

### DIFF
--- a/packages/compiler-cli/ngcc/src/packages/entry_point_bundle.ts
+++ b/packages/compiler-cli/ngcc/src/packages/entry_point_bundle.ts
@@ -73,16 +73,20 @@ export function makeEntryPointBundle(
 function computePotentialDtsFilesFromJsFiles(
     fs: FileSystem, srcProgram: ts.Program, formatPath: AbsoluteFsPath,
     typingsPath: AbsoluteFsPath) {
-  const relativePath = fs.relative(fs.dirname(formatPath), fs.dirname(typingsPath));
+  const formatRoot = fs.dirname(formatPath);
+  const typingsRoot = fs.dirname(typingsPath);
   const additionalFiles: AbsoluteFsPath[] = [];
   for (const sf of srcProgram.getSourceFiles()) {
     if (!sf.fileName.endsWith('.js')) {
       continue;
     }
-    const dtsPath = fs.resolve(
-        fs.dirname(sf.fileName), relativePath, fs.basename(sf.fileName, '.js') + '.d.ts');
-    if (fs.exists(dtsPath)) {
-      additionalFiles.push(dtsPath);
+
+    // Given a source file at e.g. `esm2015/src/some/nested/index.js`, try to resolve the
+    // declaration file under the typings root in `src/some/nested/index.d.ts`.
+    const mirroredDtsPath =
+        fs.resolve(typingsRoot, fs.relative(formatRoot, sf.fileName.replace(/\.js$/, '.d.ts')));
+    if (fs.exists(mirroredDtsPath)) {
+      additionalFiles.push(mirroredDtsPath);
     }
   }
   return additionalFiles;
--- a/packages/compiler-cli/ngcc/test/analysis/module_with_providers_analyzer_spec.ts
+++ b/packages/compiler-cli/ngcc/test/analysis/module_with_providers_analyzer_spec.ts
@@ -338,7 +338,8 @@ runInEachFileSystem(() => {
             new MockLogger(), false, program.getTypeChecker(), dtsProgram);
         referencesRegistry = new NgccReferencesRegistry(host);
 
-        const analyzer = new ModuleWithProvidersAnalyzer(host, referencesRegistry);
+        const processDts = true;
+        const analyzer = new ModuleWithProvidersAnalyzer(host, referencesRegistry, processDts);
         analyses = analyzer.analyzeProgram(program);
       });
 
@@ -427,21 +428,19 @@ runInEachFileSystem(() => {
 
   describe('tracking references when generic types already present', () => {
     let _: typeof absoluteFrom;
-    let analyses: ModuleWithProvidersAnalyses;
-    let program: ts.Program;
-    let dtsProgram: BundleProgram;
-    let referencesRegistry: NgccReferencesRegistry;
+    let TEST_DTS_PROGRAM: TestFile[];
+    let TEST_PROGRAM: TestFile[];
 
     beforeEach(() => {
       _ = absoluteFrom;
 
-      const TEST_PROGRAM: TestFile[] = [
+      TEST_PROGRAM = [
         {
           name: _('/node_modules/test-package/src/entry-point.js'),
           contents: `
           export * from './explicit';
           export * from './module';
-        `
+        `,
         },
         {
           name: _('/node_modules/test-package/src/explicit.js'),
@@ -469,26 +468,26 @@ runInEachFileSystem(() => {
               };
             }
           }
-          `
+          `,
         },
         {
           name: _('/node_modules/test-package/src/module.js'),
           contents: `
           export class ExternalModule {}
-          `
+          `,
         },
         {
           name: _('/node_modules/some-library/index.d.ts'),
-          contents: 'export declare class LibraryModule {}'
+          contents: 'export declare class LibraryModule {}',
         },
       ];
-      const TEST_DTS_PROGRAM: TestFile[] = [
+      TEST_DTS_PROGRAM = [
         {
           name: _('/node_modules/test-package/typings/entry-point.d.ts'),
           contents: `
           export * from './explicit';
           export * from './module';
-        `
+        `,
         },
         {
           name: _('/node_modules/test-package/typings/explicit.d.ts'),
@@ -502,13 +501,13 @@ runInEachFileSystem(() => {
             static explicitExternalMethod(): ModuleWithProviders<ExternalModule>;
             static explicitLibraryMethod(): ModuleWithProviders<LibraryModule>;
           }
-          `
+          `,
         },
         {
           name: _('/node_modules/test-package/typings/module.d.ts'),
           contents: `
           export declare class ExternalModule {}
-          `
+          `,
         },
         {
           name: _('/node_modules/test-package/typings/core.d.ts'),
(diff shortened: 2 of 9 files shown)


packages/compiler-cli/ngcc/src/packages/entry_point_bundle.ts
```diff
@@ -73,16 +73,28 @@ export function makeEntryPointBundle(
...
     if (!sf.fileName.endsWith('.js')) {
       continue;
     }
-    const dtsPath = fs.resolve(
+
+    const relativeDtsPath = fs.resolve(
         fs.dirname(sf.fileName), relativePath, fs.basename(sf.fileName, '.js') + '.d.ts');
-    if (fs.exists(dtsPath)) {
-      additionalFiles.push(dtsPath);
+    if (fs.exists(relativeDtsPath)) {
```
JoostK: @petebacondarwin Do you think this way of computing the .d.ts path is still necessary? I am not sure whether this way of guessing the .d.ts path is required for some packages, in favor of the new computation I added below. If I remove this now no tests fail, so I'm inclined to just delete this logic ;) (Or we should add a test for cases where this computation is correct).

petebacondarwin: I guess this logic is wrong. It would only work if the src code is one level deep, right?
In that case we should delete this, yes.

petebacondarwin (commented):
First commit message typo:

> as it may declare types _**that**_ are only used internally 

Second commit typo:

> layouy

packages/compiler-cli/ngcc/test/integration/util.ts
```diff
@@ -18,9 +18,152 @@ export type NodeModulesDef = {
...
+  sources: {[path: string]: string;};
 };
 
+/**
+ * A package layout generator function implements the logic on how a set of test files is compiled
+ * into a package layout. The test files are available in `compiledFs` in which the compilation
+ * results can be emitted, then copied over into `fs` according to the desired layout.
+ */
+export type PackageLayoutGenerator =
+    (pkgName: string, rootNames: string[], fs: FileSystem, compileFs: FileSystem) => void;
```
petebacondarwin: These are not really layout generators because they are also responsible for doing compilation.
I think it is a bit of a false abstraction, since the only bit outside the generator is the loading up of the `compileFs` with files, which itself feels like an implementation detail of these functions.

How about you bring the loading inside the generator (you can implement it as a shared function)? You already pass in all the information you need since the `rootNames` provides the list of files to load.

petebacondarwin (commented):
Third commit message: can you use `format` rather than `bundle`?

packages/compiler-cli/ngcc/src/packages/entry_point_bundle.ts
JoostK: Yeah, I figured it's limited to one level as well. I found this while working on the APF integration test, where the relative distance was more than one level and the path was guessed incorrectly.

I removed this now.

packages/compiler-cli/ngcc/test/analysis/module_with_providers_analyzer_spec.ts
```diff
@@ -563,5 +562,33 @@ runInEachFileSystem(() => {
       expect(declarations.has(externalModuleDeclaration.name !)).toBe(true);
       expect(declarations.has(libraryModuleDeclaration.name !)).toBe(false);
     });
+
+    it('should track references even when no typings are processed', () => {
```
petebacondarwin: To match the bug being fixed, and to make it easier to understand this test...

I would change the description to:

```
should track references even when no typings have already been processed
```

Then further down when you call `new ModuleWithProvidersAnalyzer(host, referencesRegistry, false);` add a comment to explain that you are passing `false` to simulate the typings files having been processed already.

petebacondarwin (approved):
Great work (as always) @JoostK 

A few nits with tests and commit messages but generally looks marvellous.

packages/compiler-cli/ngcc/test/integration/util.ts
JoostK: I had quite some attempts at a reasonable API here. It started out as _generator_ that would build some datastructure containing just layout information and compilation options, then the `compileNodeModuleToFs` could actually perform the compilation based on this info. That didn't work out well due to the vast differences in potential layouts, so most logic was moved into the generator, as well as all stuff like `pkgName`, `rootNames` and `compileFs` (I tried hiding this in callback functions but in the end it felt like adding no value whatsoever so I just passed it all as params).

Looking at it again, I think that `PackageLayoutGenerator` should just replace `compileNodeModuleToFs` in its entirety as you suggest, and be called something different.

petebacondarwin: Yeah, I can sympathise - when looking at this I also ran through various different options for abstracting similar to what you did.

JoostK: Even better would be to remove `genNodeModules`. The layout should just be the top level call, `compileIntoApf` or similar. It's much better configurable while still allowing for multiple packages, they'd just be separate calls.

packages/compiler-cli/ngcc/test/integration/util.ts
```diff
@@ -13,81 +13,194 @@ import {MockFileSystemPosix} from '../../../src/ngtsc/file_system/testing';
...
-    compileNodeModuleToFs(fs, pkgName, def[pkgName]);
+  const {rootNames, compileFs} = setupCompileFs(sources);
+
+  const emit = (options: ts.CompilerOptions) => {
+    const host = new MockCompilerHost(compileFs);
+    const program = ts.createProgram({host, rootNames, options});
+    program.emit();
+  };
+
+  emit({declaration: true, module: options.module, target: options.target, lib: []});
```
petebacondarwin:
```suggestion
  emit({declaration: true, lib: [], ...options});
```

packages/compiler-cli/ngcc/test/integration/util.ts
```diff
@@ -13,81 +13,194 @@ import {MockFileSystemPosix} from '../../../src/ngtsc/file_system/testing';
...
+ * @param pkgName The name of the package to compile.
+ * @param sources The TypeScript sources to compile.
  */
-export function genNodeModules(def: NodeModulesDef): void {
+export function compileIntoFlatEs5Package(pkgName: string, sources: PackageSources): void {
+  compileIntoFlatPackage(pkgName, sources, {
+    target: ts.ScriptTarget.ES5,
+    module: ts.ModuleKind.ESNext,
+  });
+}
```
petebacondarwin: NIT: Personally I wouldn't bother with this helper, would just use the `compilerIntoFlatPackage()` directly - but perhaps move the `options` to the second argument, and perhaps only have a `target` argument rather than `options` since at the moment we are not interested in non-ES modules.

packages/compiler-cli/ngcc/test/integration/util.ts
```diff
@@ -13,81 +13,194 @@ import {MockFileSystemPosix} from '../../../src/ngtsc/file_system/testing';
...
 
   fs.writeFile(
       fs.resolve(`/node_modules/${pkgName}/package.json`), JSON.stringify(pkgJson, null, 2));
 }
 
+/**
+ * Prepares a mock filesystem that contains all provided source files, which can be used to emit
+ * compiled code into.
+ */
+function setupCompileFs(sources: PackageSources): {rootNames: string[], compileFs: FileSystem} {
```
petebacondarwin: :+1:

petebacondarwin (approved):

packages/compiler-cli/ngcc/test/integration/util.ts
JoostK: The module kind should allow for generating UMD/CJS bundles. Regarding the position of `options` (or individual params) I put it last because it achieves a consistent pattern at call sites, i.e. `compileIntoApf` and `compileIntoFlatEs5Package`  both have package name and sources in the same position. That's also why `compileIntoFlatEs5Package` delegates to the non-exported `compileIntoFlatPackage`, to avoid having to pass default configuration at call sites.

packages/compiler-cli/ngcc/test/integration/util.ts
JoostK: I prefer the current way, as I'd like to avoid spreading a non-related interface into another one.

petebacondarwin:
@JoostK please remove the clean-up and add the merge labels when you are happy.

packages/compiler-cli/ngcc/src/packages/entry_point_bundle.ts
```diff
@@ -73,16 +73,28 @@ export function makeEntryPointBundle(
...
+  const typingsRoot = fs.dirname(typingsPath);
+  const relativePath = fs.relative(formatRoot, typingsRoot);
   const additionalFiles: AbsoluteFsPath[] = [];
   for (const sf of srcProgram.getSourceFiles()) {
     if (!sf.fileName.endsWith('.js')) {
       continue;
     }
-    const dtsPath = fs.resolve(
+
+    const relativeDtsPath = fs.resolve(
```
gkalpak: This name is a little confusing given that the path with absolute 😕

packages/compiler-cli/ngcc/test/analysis/module_with_providers_analyzer_spec.ts
```diff
@@ -563,5 +564,34 @@ runInEachFileSystem(() => {
       expect(declarations.has(externalModuleDeclaration.name !)).toBe(true);
       expect(declarations.has(libraryModuleDeclaration.name !)).toBe(false);
     });
+
+    it('should track references even when no typings have already been processed', () => {
```
gkalpak: Typo: `even when no typings have ...` --> `even when typings have`

PR #33878, opened 2019-11-16
fix(ngcc): correctly associate decorators with aliased classes

JoostK:
In flat bundle formats, multiple classes that have the same name can be
suffixed to become unique. In ES5-like bundles this results in the outer
declaration from having a different name from the "implementation"
declaration within the class' IIFE, as the implementation declaration
may not have been suffixed.

As an example, the following code would fail to have a `Directive`
decorator as ngcc would search for `__decorate` calls that refer to
`AliasedDirective$1` by name, whereas the `__decorate` call actually
uses the `AliasedDirective` name.

```javascript
var AliasedDirective$1 = /** @Class */ (function () {
    function AliasedDirective() {}
    AliasedDirective = tslib_1.__decorate([
        Directive({ selector: '[someDirective]' }),
    ], AliasedDirective);
    return AliasedDirective;
}());
```

This commit fixes the problem by not relying on comparing names, but
instead finding the declaration and matching it with both the outer
and inner declaration.

### DIFF
--- a/packages/compiler-cli/ngcc/src/host/esm2015_host.ts
+++ b/packages/compiler-cli/ngcc/src/host/esm2015_host.ts
@@ -931,8 +931,21 @@ export class Esm2015ReflectionHost extends TypeScriptReflectionHost implements N
     // has to be checked to actually correspond with the class symbol.
     const helperCalls = this.getHelperCallsForClass(classSymbol, ['__decorate']);
 
+    const outerDeclaration = classSymbol.declaration.valueDeclaration;
+    const innerDeclaration = classSymbol.implementation.valueDeclaration;
+    const matchesClass = (identifier: ts.Identifier) => {
+      const decl = this.getDeclarationOfIdentifier(identifier);
+      if (decl === null) {
+        return false;
+      }
+
+      // The identifier corresponds with the class if its declaration is either the outer or inner
+      // declaration.
+      return decl.node === outerDeclaration || decl.node === innerDeclaration;
+    };
+
     for (const helperCall of helperCalls) {
-      if (isClassDecorateCall(helperCall, classSymbol.name)) {
+      if (isClassDecorateCall(helperCall, matchesClass)) {
         // This `__decorate` call is targeting the class itself.
         const helperArgs = helperCall.arguments[0];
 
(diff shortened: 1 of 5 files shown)


packages/compiler-cli/ngcc/src/host/esm2015_host.ts
```diff
@@ -931,8 +931,21 @@ export class Esm2015ReflectionHost extends TypeScriptReflectionHost implements N
...
+    const innerDeclaration = classSymbol.implementation.valueDeclaration;
+    const matchesClass = (identifier: ts.Identifier) => {
+      const decl = this.getDeclarationOfIdentifier(identifier);
+      if (decl === null) {
+        return false;
+      }
+
+      // The identifier corresponds with the class if its declaration is either the outer or inner
+      // declaration.
+      return decl.node === outerDeclaration || decl.node === innerDeclaration;
```
petebacondarwin: This can be simplified to 

```ts
return decl !== null && (decl.node === outerDeclaration || decl.node === innerDeclaration);
```

Or if we are sure that `innerDeclaration` and `outerDeclaration` are always defined then simply:

```ts
return decl.node === outerDeclaration || decl.node === innerDeclaration;
```

Right?

packages/compiler-cli/ngcc/test/host/esm5_host_import_helper_spec.ts
```diff
@@ -174,12 +174,27 @@ import { Directive } from '@angular/core';
 // Note that the IIFE is not in parentheses
 var SomeDirective = function () {
     function SomeDirective() {}
-    // Note that the decorator is combined with the return statment
+    // Note that the decorator is combined with the return statement
     return SomeDirective = tslib_1.__decorate([
         Directive({ selector: '[someDirective]' }),
     ], SomeDirective);
-}());
+}();
```
petebacondarwin: LOL! Thanks.

petebacondarwin (approved):

packages/compiler-cli/ngcc/src/host/esm2015_host.ts
JoostK: Your first suggestion is possible, but I deliberately choose the current logic. I like how it separates handling of the edge case where `identifier` could not be resolved to its declaration, vs the actual logic of comparing the declarations.

Your second suggestion is essentially what I have now, but the null guard is required because that is targeting `decl`, not `outerDeclaration`/`innerDeclaration`.

tl;dr; I'll keep this as is.

PR #29811, opened 2019-04-10
fix(language-service): infer the type of array-like element.

ivanwonder:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] angular.io application / infrastructure changes
- [ ] Other... Please describe:

## What is the current behavior?

```js
const test: { {[key: string]: number};
```
current behavior of `test.test` will report a error;

## What is the new behavior?
new behaviour will infer the type `number`;

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

PR #33884, opened 2019-11-17
fix(language-service): determine index types accessed using dot notation

### LINKED ISSUE #126: Support indexed types
When using reactive forms, in the HTML template I get the error in the title. 

<ion-item *ngIf="formRegister.controls.company.invalid && formRegister.controls.company.touched">
                <p>Please enter a valid name.</p>
</ion-item>

Both "formRegister.controls.company.invalid && formRegister.controls.company.touched" are underlined as an error but the app works just fine.

### LINKED ISSUE #149: Support optional chaining in template
in 0.1.4 on Windows 10

`errors?.required` is highlighted as an error

<div class="container">

    <form (submit)="onSubmit()" novalidate #f="ngForm" (ngSubmit)="onSubmit(f)">
        <div class="form-group">
            <label for="name">Name</label>
            <input type="text" class="form-control" [(ngModel)]="user.name" name="name" #userName="ngModel" minLength="2" required>
            <div *ngIf="userName.touched && userName.errors?.required" class="alert alert-danger">Name is required</div>
        </div>
        <div class="form-group">
            <label for="name">Email</label>
            <input type="text" class="form-control" [(ngModel)]="user.email" name="email" #userEmail="ngModel" required>
        </div>
        <div class="form-group">
            <label for="age">Phone</label>
            <input type="number" class="form-control" [(ngModel)]="user.phone" #userPhone="ngModel" name="phone" minlength="10">
        </div>
        <input type="submit" value="Submit" class="btn btn-success">
    </form>
</div>

image

ayazhafiz:
Commit 53fc2ed8bf345222e0c3d53ce7f13a4f27f3052e added support for
determining index types accessed using index signatures, but did not
include support for index types accessed using dot notation:

```typescript
const obj<T>: { [key: string]: T };
obj['stringKey']. // gets `T.` completions
obj.stringKey.    // did not peviously get `T.` completions
```

This adds support for determining an index type accessed via dot
notation by rigging an object's symbol table to return the string index
signature type a property access refers to, if that property does not
explicitly exist on the object. This is very similar to @ivanwonder's
work in #29811.

`SymbolWrapper` now takes an additional parameter to explicitly set the
type of the symbol wrapped. This is done because
`SymbolTableWrapper#get` only has access to the symbol of the index
type, _not_ the index signature symbol itself. An attempt to get the
type of the index type will give an error.

Closes #29811
Closes https://github.com/angular/vscode-ng-language-service/issues/126
Closes https://github.com/angular/vscode-ng-language-service/issues/149

## PR Checklist

- [ ] Yes
- [x] No

### DIFF
--- a/packages/language-service/src/typescript_symbols.ts
+++ b/packages/language-service/src/typescript_symbols.ts
@@ -275,7 +275,7 @@ class TypeWrapper implements Symbol {
     // the former includes properties on the base class whereas the latter does
     // not. This provides properties like .bind(), .call(), .apply(), etc for
     // functions.
-    return new SymbolTableWrapper(this.tsType.getApparentProperties(), this.context);
+    return new SymbolTableWrapper(this.tsType.getApparentProperties(), this.context, this.tsType);
   }
 
   signatures(): Signature[] { return signaturesOf(this.tsType, this.context); }
@@ -302,15 +302,12 @@ class TypeWrapper implements Symbol {
 
 class SymbolWrapper implements Symbol {
   private symbol: ts.Symbol;
-  // TODO(issue/24571): remove '!'.
-  private _tsType !: ts.Type;
-  // TODO(issue/24571): remove '!'.
-  private _members !: SymbolTable;
+  private _members?: SymbolTable;
 
   public readonly nullable: boolean = false;
   public readonly language: string = 'typescript';
 
-  constructor(symbol: ts.Symbol, private context: TypeContext) {
+  constructor(symbol: ts.Symbol, private context: TypeContext, private _tsType?: ts.Type) {
     this.symbol = symbol && context && (symbol.flags & ts.SymbolFlags.Alias) ?
         context.checker.getAliasedSymbol(symbol) :
         symbol;
@@ -340,7 +337,7 @@ class SymbolWrapper implements Symbol {
         const typeWrapper = new TypeWrapper(declaredType, this.context);
         this._members = typeWrapper.members();
       } else {
-        this._members = new SymbolTableWrapper(this.symbol.members !, this.context);
+        this._members = new SymbolTableWrapper(this.symbol.members !, this.context, this.tsType);
       }
     }
     return this._members;
@@ -454,8 +451,16 @@ function toSymbols(symbolTable: ts.SymbolTable | undefined): ts.Symbol[] {
 class SymbolTableWrapper implements SymbolTable {
   private symbols: ts.Symbol[];
   private symbolTable: ts.SymbolTable;
+  private stringIndexType?: ts.Type;
 
-  constructor(symbols: ts.SymbolTable|ts.Symbol[]|undefined, private context: TypeContext) {
+  /**
+   * Creates a queryable table of symbols belonging to a TypeScript entity.
+   * @param symbols symbols to query belonging to the entity
+   * @param context program context
+   * @param type original TypeScript type of entity owning the symbols, if known
+   */
+  constructor(
+      symbols: ts.SymbolTable|ts.Symbol[], private context: TypeContext, private type?: ts.Type) {
     symbols = symbols || [];
 
     if (Array.isArray(symbols)) {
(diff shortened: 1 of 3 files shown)


packages/language-service/src/typescript_symbols.ts
```diff
@@ -298,15 +298,12 @@ class TypeWrapper implements Symbol {
...
-  private _tsType !: ts.Type;
-  // TODO(issue/24571): remove '!'.
-  private _members !: SymbolTable;
+  private _members?: SymbolTable;
 
   public readonly nullable: boolean = false;
   public readonly language: string = 'typescript';
 
-  constructor(symbol: ts.Symbol, private context: TypeContext) {
+  constructor(symbol: ts.Symbol, private context: TypeContext, private _tsType?: ts.Type) {
```
kyliau: Could you please add docs to the constructor? It is not immediately what `_tsType` is used for.
Maybe an example will help.
It's a little confusing because `symbol` itself has a `type` member, and there's this additional parameter`_tsType`. Are they different? 
If they are the same, is there a way to safely access `tsType` inside `symbol.type`?

packages/language-service/src/typescript_symbols.ts
```diff
@@ -461,18 +466,41 @@ class SymbolTableWrapper implements SymbolTable {
...
+      }
+    }
+
+    return undefined;
   }
 
   has(key: string): boolean {
     const table: any = this.symbolTable;
-    return (typeof table.has === 'function') ? table.has(key) : table[key] != null;
+    return ((typeof table.has === 'function') ? table.has(key) : table[key] != null) ||
```
kyliau: Since we are changing this function, it's better to clean it up as well.
table will always be a Map, and thus will always have the `get()` function. We could remove the check.

packages/language-service/test/completions_spec.ts
```diff
@@ -117,18 +117,29 @@ describe('completions', () => {
...
+    describe('with string index signatures', () => {
+      it('should work with index notation', () => {
+        mockHost.override(TEST_TEMPLATE, `{{ heroesByName['Jacky'].~{heroes-string-index}}}`);
+        const marker = mockHost.getLocationMarkerFor(TEST_TEMPLATE, 'heroes-string-index');
+        const completions = ngLS.getCompletionsAt(TEST_TEMPLATE, marker.start);
+        expectContain(completions, CompletionKind.PROPERTY, ['id', 'name']);
+      });
+
+      it('should work with dot notation', () => {
+        mockHost.override(TEST_TEMPLATE, `{{ heroesByName.jacky.~{heroes-string-index}}}`);
```
kyliau: hehe, I see what you did here. 😛

kyliau (changes requested):

packages/language-service/src/typescript_symbols.ts
ayazhafiz: They're not the same. Only an Angular symbol has a `type` field, but here `_tsType` refers to the type of the TypeScript symbol. If it is provided, that type will be used; otherwise, the type will be dynamically determined using a typechecker. Added docs for this.

packages/language-service/src/typescript_symbols.ts
ayazhafiz: Unfortunately, this is not true if we want to support pre-2.2 TypeScript:
https://github.com/angular/angular/blob/25dcc7631faac0e3b80ab33becaaad161bdffb3b/packages/language-service/src/typescript_symbols.ts#L787-L793.

packages/language-service/test/completions_spec.ts
ayazhafiz: :)

packages/language-service/src/typescript_symbols.ts
kyliau: I removed the code to support pre 2.2 when I moved this file from compiler-cli to language-service.
Let's leave this clean up to another PR then.

kyliau (approved):

andrius-pra:
@kyliau @ayazhafiz , would it be possible to change label to `master & patch` and also cherry-pick 53fc2ed into 9.0.x branch?

kyliau:
> @kyliau @ayazhafiz , would it be possible to change label to `master & patch` and also cherry-pick [53fc2ed](https://github.com/angular/angular/commit/53fc2ed8bf345222e0c3d53ce7f13a4f27f3052e) into 9.0.x branch?

Ah yes, this should target master & patch. Thank you for pointing that out!

@ayazhafiz Could you please submit a separate PR for https://github.com/angular/angular/commit/53fc2ed8bf345222e0c3d53ce7f13a4f27f3052e that targets patch-only?

ayazhafiz:
sure!

JounQin:
Finally!

PR #33895, opened 2019-11-18
build: replace `fs` methods with ShellJS' `test('-d')`

gkalpak:
This makes the JS script more similar to the original bash script it replaced (see #33854) and will be easier to follow for people that are less familiar with Node.js scripting.

Discussed in: https://github.com/angular/angular/pull/33854#discussion_r346949765


josephperrott (approved):
LGTM

devversion (approved):

PR #33901, opened 2019-11-18
fix(ngcc): properly detect origin of constructor param types

### LINKED ISSUE #33677: ivy - Async pipe gets wrong ChangeDetectorRef instance when using pipe inside a component input
# 🐞 bug report

### Affected Package

 The issue is caused by package @angular/compiler-cli/ngcc

### Is this a regression?

 Yes, it works in ViewEngine and should also work with Ivy after #31438 but it doesn't work for ngcc processed pipes.

### Description

When using Async pipe inside a component input Angular should inject a ChangeDetectorRef of the containing view but it injects the ChangeDetectorRef that refers to component view.

```
@Component({
  selector: 'my-app',
  template: `
   <child [prop]="obs$ | async"></child>
  `
})
class AppComponent {}

// @angular/common
class AsyncPipe {
 constructor(cdRef: ChangeDetectorRef) {}
}

// expected  cdRef.context -> AppComponent
// actual    cdRef.context -> ChildComponent
```

Not too long time ago there was a PR #31438 that was intended to fix that problem with pipes.  It introduced `ɵɵinjectPipeChangeDetectorRef` function to inject `ChangeDetectorRef` so that the factory for pipe that injects `ChangeDetectorRef` should look like:

```
AsyncPipe.ɵfac = function AsyncPipe_Factory(t) {
  return new (t || AsyncPipe)(ɵngcc0.ɵɵinjectPipeChangeDetectorRef(};
}
```
but we get the following:

```
AsyncPipe.ɵfac = function AsyncPipe_Factory(t) { 
  return new (t || AsyncPipe)(ɵngcc0.ɵɵdirectiveInject(ChangeDetectorRef));
};
```
**It should be noted that ngtsc generates correct factories for user-defined pipes** since in this case ngtsc uses TypeScriptReflectionHost to read metadata. The TypeScriptReflectionHost  treats ChangeDetectorRef reference as ExternalExpr while Esm2015ReflectionHost treats it as a local reference https://github.com/angular/angular/blob/b381497126c58321290bd7df5b90ddbb903bed5a/packages/compiler-cli/ngcc/src/host/esm2015_host.ts#L1318

## 🌍  Your Environment

**Angular Version:**
<pre>@angular/core@9.0.0-rc.0
@angular/common@9.0.0-rc.0
@angular/compiler-cli@9.0.0-rc.0
</pre>

@alxhub @petebacondarwin @JoostK @crisbeto

alxhub:
The ReflectionHost supports enumeration of constructor parameters, and one
piece of information it returns describes the origin of the parameter's
type. Parameter types come in two flavors: local (the type is not imported
from anywhere) or non-local (the type comes via an import).

ngcc incorrectly classified all type parameters as 'local', because in the
source files that ngcc processes the type parameter is a real ts.Identifer.
However, that identifier may still have come from an import and thus might
be non-local.

This commit changes ngcc's ReflectionHost(s) to properly recognize and
report these non-local type references.

Fixes #33677

### DIFF
--- a/packages/compiler-cli/ngcc/src/host/esm2015_host.ts
+++ b/packages/compiler-cli/ngcc/src/host/esm2015_host.ts
@@ -9,7 +9,7 @@
 import * as ts from 'typescript';
 
 import {AbsoluteFsPath} from '../../../src/ngtsc/file_system';
-import {ClassDeclaration, ClassMember, ClassMemberKind, ConcreteDeclaration, CtorParameter, Declaration, Decorator, TypeScriptReflectionHost, isDecoratorIdentifier, reflectObjectLiteral} from '../../../src/ngtsc/reflection';
+import {ClassDeclaration, ClassMember, ClassMemberKind, ConcreteDeclaration, CtorParameter, Declaration, Decorator, TypeScriptReflectionHost, TypeValueReference, isDecoratorIdentifier, reflectObjectLiteral} from '../../../src/ngtsc/reflection';
 import {isWithinPackage} from '../analysis/util';
 import {Logger} from '../logging/logger';
 import {BundleProgram} from '../packages/bundle_program';
@@ -1357,12 +1357,34 @@ export class Esm2015ReflectionHost extends TypeScriptReflectionHost implements N
           constructorParamInfo[index] :
           {decorators: null, typeExpression: null};
       const nameNode = node.name;
+
+      let typeValueReference: TypeValueReference|null = null;
+      if (typeExpression !== null) {
+        // `typeExpression` is an expression in a "type" context. Resolve it to a declared value.
+        // Either it's a reference to an imported type, or a type declared locally. Distinguish the
+        // two cases with `getDeclarationOfExpression`.
+        const decl = this.getDeclarationOfExpression(typeExpression);
+        if (decl !== null && decl.node !== null && decl.viaModule !== null &&
+            isNamedDeclaration(decl.node)) {
+          typeValueReference = {
+            local: false,
+            valueDeclaration: decl.node,
+            moduleName: decl.viaModule,
+            name: decl.node.name.text,
+          };
+        } else {
+          typeValueReference = {
+            local: true,
+            expression: typeExpression,
+            defaultImportStatement: null,
+          };
+        }
+      }
+
       return {
         name: getNameText(nameNode),
         nameNode,
-        typeValueReference: typeExpression !== null ?
-            {local: true as true, expression: typeExpression, defaultImportStatement: null} :
-            null,
+        typeValueReference,
         typeNode: null, decorators
       };
     });
--- a/packages/compiler-cli/ngcc/test/host/umd_host_spec.ts
+++ b/packages/compiler-cli/ngcc/test/host/umd_host_spec.ts
@@ -1560,9 +1560,6 @@ runInEachFileSystem(() => {
 
           expect(decorators.length).toEqual(1);
           expect(decorators[0].import).toBe(mockImportInfo);
-
-          const typeIdentifier = spy.calls.mostRecent().args[0] as ts.Identifier;
-          expect(typeIdentifier.text).toBe('Inject');
         });
       });
 
--- a/packages/compiler-cli/ngcc/test/host/esm2015_host_spec.ts
+++ b/packages/compiler-cli/ngcc/test/host/esm2015_host_spec.ts
@@ -54,11 +54,10 @@ runInEachFileSystem(() => {
       SOME_DIRECTIVE_FILE = {
         name: _('/some_directive.js'),
         contents: `
-    import { Directive, Inject, InjectionToken, Input, HostListener, HostBinding } from '@angular/core';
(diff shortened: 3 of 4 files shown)


packages/compiler-cli/ngcc/src/host/esm2015_host.ts
```diff
@@ -1357,12 +1357,34 @@ export class Esm2015ReflectionHost extends TypeScriptReflectionHost implements N
           constructorParamInfo[index] :
           {decorators: null, typeExpression: null};
       const nameNode = node.name;
+
+      let typeValueReference: TypeValueReference|null = null;
+      if (typeExpression !== null) {
+        // `typeExpression` is an expression in a "type" context. Resolve it to a declared value.
+        // Either it's a reference to an imported type, or a type declared locally. Distinguish the
+        // to cases with `getDeclarationOfExpression`.
```
petebacondarwin: typo: to => two

packages/compiler-cli/ngcc/src/host/esm2015_host.ts
```diff
@@ -1357,12 +1357,34 @@ export class Esm2015ReflectionHost extends TypeScriptReflectionHost implements N
...
+        const decl = this.getDeclarationOfExpression(typeExpression);
+        if (decl !== null && decl.node !== null && decl.viaModule !== null &&
+            isNamedDeclaration(decl.node)) {
```
petebacondarwin: Can it be possible that `viaModule !== null && !isNamedDeclaration(decl.node)`?
If so, what does that mean? As this stands, we will return a "local" reference instead.

packages/compiler-cli/ngcc/test/host/umd_host_spec.ts
```diff
@@ -1555,14 +1555,12 @@ runInEachFileSystem(() => {
           const spy = spyOn(UmdReflectionHost.prototype, 'getImportOfIdentifier')
                           .and.returnValue(mockImportInfo);
 
+          debugger;
```
petebacondarwin: Oops

packages/compiler-cli/ngcc/test/host/umd_host_spec.ts
```diff
@@ -1555,14 +1555,12 @@ runInEachFileSystem(() => {
...
           const parameters = host.getConstructorParameters(classNode);
           const decorators = parameters ![2].decorators !;
 
           expect(decorators.length).toEqual(1);
           expect(decorators[0].import).toBe(mockImportInfo);
-
-          const typeIdentifier = spy.calls.mostRecent().args[0] as ts.Identifier;
-          expect(typeIdentifier.text).toBe('Inject');
```
petebacondarwin: Why is this removed?

packages/compiler-cli/ngcc/src/host/esm2015_host.ts
```diff
@@ -1357,12 +1357,34 @@ export class Esm2015ReflectionHost extends TypeScriptReflectionHost implements N
...
+            name: decl.node.name.text,
+          };
+        } else {
+          typeValueReference = {
+            local: true,
+            expression: typeExpression,
+            defaultImportStatement: null
+          };
+        }
+      }
```
petebacondarwin: NIT: this block of code is a candidate for a new private method, to keep the size of this method lean.

```ts
return {
  name: getNameText(nameNode),
  nameNode,
  typeValueReference: this.getTypeValueReference(typeExpression),
  typeNode: null,
  decorators,
};
```

petebacondarwin (approved):

packages/compiler-cli/ngcc/test/host/umd_host_spec.ts
alxhub: It fails now, since the work that `getConstructorParameters()` does calls `getImportOfIdentifier` more than this test was expecting.

In this case, the spy is configured to return incorrect information, so the previous assertion is enough to validate that the expected code path was used.

In general, this is why I think spies in tests are more trouble than they're worth. At best, this test is too sensitive to changes in the internals of the host. At worst it could be viewed as an assertion of the implementation only.

packages/compiler-cli/ngcc/src/host/esm2015_host.ts
alxhub: I don't think it's generally possible, since a declaration has to have a name to be exported in the first place. You could probably contrive some example where it's true (`export default 42`?) but in general I think a fallback option of "this could not be understood as a sane import, so just use the local identifier instead" is sane.

packages/compiler-cli/ngcc/src/host/esm2015_host.ts
alxhub: Nice catch!

packages/compiler-cli/ngcc/src/host/esm2015_host.ts
alxhub: Yeah, in general we should consider "constructor" methods for more complex types like this.

PR #33903, opened 2019-11-18
fix(service-worker): throw when using the unsupported `versionedFiles` option in config

gkalpak:
In 5d5c94d, the deprecated `versionedFiles` option from the SW asset-group configuration in `ngsw-config.json`. As a result, the option would be silently ignored and the runtime behavior of the SW would change (i.e. some files might not be cached and available offline any more). This change could be easily go unnoticed by the developer.

This commit ensures this does not happen by throwing a build-time error, when detecting the unsupported `versionedFiles` option with an error message prompting the user to use the `files` option instead.

Jira issue: [FW-1727](https://angular-team.atlassian.net/browse/FW-1727)

### DIFF
--- a/packages/service-worker/config/test/generator_spec.ts
+++ b/packages/service-worker/config/test/generator_spec.ts
@@ -12,7 +12,7 @@ import {MockFilesystem} from '../testing/mock';
 describe('Generator', () => {
   beforeEach(() => spyOn(Date, 'now').and.returnValue(1234567890123));
 
-  it('generates a correct config', done => {
+  it('generates a correct config', async() => {
     const fs = new MockFilesystem({
       '/index.html': 'This is a test',
       '/main.css': 'This is a CSS file',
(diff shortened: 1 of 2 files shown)


packages/service-worker/config/test/generator_spec.ts
```diff
@@ -67,92 +67,86 @@ describe('Generator', () => {
...
-             '/test/index.html': 'a54d88e06612d820bc3be72877c74f257b561b19',
-             '/test/main.js': '41347a66676cdc0516934c76d9d13010df420f2c',
-             '/test/main.ts': '7d333e31f0bfc4f8152732bb211a93629484c035',
-             '/test/test.txt': '18f6f8eb7b1c23d2bb61bff028b83d867a9e4643'
-           }
-         });
-         done();
-       })
-        .catch(err => done.fail(err));
+
```
IgorMinar: nice cleanup. thanks!

IgorMinar (approved):
thanks!

PR #33904, opened 2019-11-18
build: ensure package-builder finds bazel

petebacondarwin:
Since `scripts/package-builder.js` gets run from the
`aio` folder sometimes, it is important to set the cwd
for the command that finds the bazel bin directory.

### DIFF
--- a/scripts/package-builder.js
+++ b/scripts/package-builder.js
@@ -26,7 +26,7 @@ set('-e');
 const baseDir = resolve(`${__dirname}/..`);
 
 /** @type {string} The command to use for running bazel. */
-const bazelCmd = `yarn --cwd ${baseDir} --silent bazel`;
+const bazelCmd = `yarn --cwd "${baseDir}" --silent bazel`;
 
 /** @type {string} The absolute path to the bazel-bin directory. */
 const bazelBin = exec(`${bazelCmd} info bazel-bin`, true);
@@ -75,8 +75,8 @@ function buildTargetPackages(destPath, compileMode, description) {
     const pkg = target.replace(/\/\/packages\/(.*):npm_package/, '$1');
 
     // Skip any that don't have an "npm_package" target.
-    const srcDir = resolve(bazelBin, 'packages', pkg, 'npm_package');
-    const destDir = resolve(absDestPath, pkg);
+    const srcDir = `${bazelBin}/packages/${pkg}/npm_package`;
+    const destDir = `${absDestPath}/${pkg}`;
 
     if (test('-d', srcDir)) {
       console.log(`# Copy artifacts to ${destDir}`);
--- a/scripts/package-builder.js
+++ b/scripts/package-builder.js
@@ -22,7 +22,7 @@ set('-e');
 const baseDir = resolve(`${__dirname}/..`);
 
 /** @type {string} The command to use for running bazel. */
-const bazelCmd = 'yarn --silent bazel';
+const bazelCmd = `yarn --cwd ${baseDir} --silent bazel`;
 
 /** @type {string} The absolute path to the bazel-bin directory. */
 const bazelBin = exec(`${bazelCmd} info bazel-bin`, true);
@@ -35,12 +35,7 @@ const bazelBin = exec(`${bazelCmd} info bazel-bin`, true);
 const scriptPath = relative(baseDir, require.main.filename);
 
 module.exports = {
-  baseDir,
-  bazelBin,
-  bazelCmd,
-  buildTargetPackages,
-  exec,
-  scriptPath,
+    baseDir, bazelBin, bazelCmd, buildTargetPackages, exec, scriptPath,
 };
 
 /**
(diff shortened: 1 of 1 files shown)


josephperrott (approved):
LGTM

petebacondarwin:
Sorry my IDE did some auto-formatting.

scripts/package-builder.js
```diff
@@ -23,7 +23,7 @@ set('-e');
 const baseDir = resolve(`${__dirname}/..`);
 
 /** @type {string} The command to use for running bazel. */
-const bazelCmd = 'yarn --silent bazel';
+const bazelCmd = `yarn --cwd ${baseDir} --silent bazel`;
```
gkalpak: Theoretically, this is a broader issue (not strictly related to `yarn bazel`). We just happen to not be affected, because we use absolute paths in other commands (such as `cp`, `mkdir`, `rm`, etc.)

I think it would be better (and also closer to the original bash script that this JS script replaces) to use ShellJS' `cd()` somewhere near the top of the file to change the directory for all subsequent commands (and for the importing scripts, such as `build-packages-dist.js`).

petebacondarwin: But then you also need to `cd` back out again at the end, no?

gkalpak: You could do that (but then all importing scripts would also have to `cd(baseDir)` for their own commands. This would be "more correct", but maybe also easier for people to forget.

The way `package-builder.js` is used at the moment, it is fine to rely on it to `cd` to the correct directory. This would be similar to what was happening with the previous bash scripts, which were sourced in `build-packages-dist.sh` and `build-ivy-npm-packages.sh` and changed the directory for those scripts as well.

(So, theoretically it is indeed a better practice practice to cd in/out (or rather
`pushd()`/`popd()`) whenever necessary, but in this case it would also be OK (imo) to just `cd` in the `package-builder.js` script and let that propagate to other scripts. Both approaches have pros and cons and are equally error-prone :grin:)

Feel free to go either way 👈 👉

petebacondarwin: I feel it is actually safer to require that all operations are agnostic to the current working directory.
That way you are not reliant on someone else setting it correctly for you.

Each action should be using absolute file-paths. (IMO `yarn --cwd` is like using absolute paths, since it does not change the current working directory in the script that calls it).

BTW I don't think that trying to mirror the previous scripts is necessarily a good goal.

devversion: I agree with what @petebacondarwin said. In the components repository we usually specify the `cwd`, but not necessarily through a flag like `--cwd`, but rather when we call `execSync` or `spawnSync`. e.g.

```ts
const projectDir = join(__dirname, '../../');

// run yarn or other scripts/binaries
execSync('yarn', {cwd: projectDir});
spawnSync('yarn', ..., {cwd: projectDir});

// can be surely also moved into a helper function that always sets the cwd
```

gkalpak: > I feel it is actually safer to require that all operations are agnostic to the current working directory.

Fair enough. I believe both options are safe in some regard and error-prone in another.
Can you then add a comment somewhere mentioning that the script maybe be run from any directory and commands should be "cwd-agnostic"?
Also, either pass an absolute path to `buildTargetPackages()` or update [this comment][1] to mention the path is expected to be relative to `baseDir`.

##
> BTW I don't think that trying to mirror the previous scripts is necessarily a good goal.

For sure. I was just making the point that it won't be any worse than before :grin:

##
> In the components repository we usually specify the cwd, but not necessarily through a flag like `--cwd`, but rather when we call `execSync` or `spawnSync`

@devversion, that would work too (as an alternative to `yarn --cwd ...`).
(Other commands that don't go through `exec/spawnSycn`, such as ShellJS' `cp`/`mkdir`/`rm`/etc., still need to use absolute paths.)
[1]: https://github.com/angular/angular/blob/c68200cf4/scripts/package-builder.js#L49

petebacondarwin: Done

scripts/package-builder.js
```diff
@@ -22,7 +26,7 @@ set('-e');
...
```
gkalpak: Can you wrap `${baseDir}` in quotes (in case the path contains spaces, which is not uncommon on Windows)?

scripts/package-builder.js
```diff
@@ -60,22 +60,23 @@ function buildTargetPackages(destPath, compileMode, description) {
...
   if (!test('-d', absDestPath)) mkdir('-p', absDestPath);
 
   targets.forEach(target => {
     const pkg = target.replace(/\/\/packages\/(.*):npm_package/, '$1');
 
     // Skip any that don't have an "npm_package" target.
-    const srcDir = `${bazelBin}/packages/${pkg}/npm_package`;
-    const destDir = `${absDestPath}/${pkg}`;
+    const srcDir = resolve(bazelBin, 'packages', pkg, 'npm_package');
+    const destDir = resolve(absDestPath, pkg);
```
gkalpak: This change doesn't seem necessary 😕

gkalpak (approved):

scripts/package-builder.js
petebacondarwin: I know I know, I am just allergic to hand crafting paths... I'll revert.

gkalpak: Do you know what we need? We need a package that normalizes our paths, so we don't have to worry 😛

PR #33907, opened 2019-11-18
build/ci: remove remote-http-cache config for bazel

josephperrott:
We have determined that the upload time cost of remote caching via http is too costly
to be effective for our distributed usage scenario.  However this cost is still worth
it for us for Windows CI runs on a full cache hit move from a ~60 minute task to ~10
minutes.

It is worth noting that this remote http-caching is entirely separate from RBE caching
mechanisms

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -68,21 +68,6 @@ test --test_output=errors
 # any bazel target. This is a temporary flag until codebase is permanently switched to Ivy.
 build --define=compile=legacy
 
-#######################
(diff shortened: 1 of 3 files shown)


gkalpak (approved):
Please, add a short explanation on why we make these changes in the commit message.

devversion (approved):
LGTM. Agreed that there is missing context on why this change is made. i.e. what's the reasoning for disabling result uploading for non-CI RBE builds?

PR #33921, opened 2019-11-19
fix(ivy): add flag to skip non-exported classes

### LINKED ISSUE #33724: ngtsc tries to compile non-exported components while ngc did not
### Affected package

`@angular/compiler-cli` ngtsc.

### Problem description

Currently compiling a TS project which contains both test files and application code, NGC properly ignores non-exported components from test files.

When updating to v9, where ngtsc runs by default, the compilation will fail since `ngtsc` tries to compile the `@Component` annotations even for test components. Apparently since those test components are never part of a  `@NgModule`, the compilation will fail (e.g. component not found).

_accordion.spec.ts_
```ts
@Component({
  template: `<cdk-accordion>..</cdk-accordion>`
})
/** notice: not exported */ class TestComponent {}
```

One could argue that test files and application code shouldn't be part of the same compilation unit, but this has been working in View Engine and helped speeding up test compilation. There are more benefits like not having to use path mapping inside tests. i.e. relative imports to application code work just fine.

Another benefit is that the application/library components are compiled AOT. This is the primary reason why we follow this pattern (in `angular/components`).

**Note**: CLI projects follows the same pattern where tests and application code are part of the same compilation unit. The CLI just runs `tsc` instead of the Angular compiler (which we want for AOT)

### Possible solutions

* Saying that this is the expected behavior, though it will be a change in v9 that involves infrastructure changes. It doesn't seem easy to split into two compilation units, if tests and application code are located next to each other.
* Following NGC behavior by ignoring symbols which are not exported.
* Adding `jit: true` to every test component.

### Reproduction

https://github.com/devversion/ngtsc-test-component-analysis

crisbeto:
In ViewEngine we were only generating code for exported classes, however with Ivy we do it no matter whether the class has been exported or not. These changes add an extra flag that allows consumers to opt into the ViewEngine behavior. The flag works by treating non-exported classes as if they're set to `jit: true`.

Fixes #33724.

### DIFF
--- a/packages/compiler-cli/src/ngtsc/program.ts
+++ b/packages/compiler-cli/src/ngtsc/program.ts
@@ -642,7 +642,8 @@ export class NgtscProgram implements api.Program {
 
     return new IvyCompilation(
         handlers, this.reflector, this.importRewriter, this.incrementalDriver, this.perfRecorder,
-        this.sourceToFactorySymbols, scopeRegistry);
+        this.sourceToFactorySymbols, scopeRegistry,
+        this.options.compileNonExportedClasses !== false);
   }
 
   private getI18nLegacyMessageFormat(): string {
--- a/packages/compiler-cli/src/ngtsc/transform/src/compilation.ts
+++ b/packages/compiler-cli/src/ngtsc/transform/src/compilation.ts
@@ -17,7 +17,7 @@ import {PerfRecorder} from '../../perf';
 import {ClassDeclaration, ReflectionHost, isNamedClassDeclaration, reflectNameOfDeclaration} from '../../reflection';
 import {LocalModuleScopeRegistry} from '../../scope';
 import {TypeCheckContext} from '../../typecheck';
-import {getSourceFile} from '../../util/src/typescript';
+import {getSourceFile, isExported} from '../../util/src/typescript';
 
 import {AnalysisOutput, CompileResult, DecoratorHandler, DetectResult, HandlerPrecedence} from './api';
 import {DtsFileTransformer} from './declaration';
@@ -80,7 +80,8 @@ export class IvyCompilation {
       private handlers: DecoratorHandler<any, any>[], private reflector: ReflectionHost,
       private importRewriter: ImportRewriter, private incrementalDriver: IncrementalDriver,
       private perf: PerfRecorder, private sourceToFactorySymbols: Map<string, Set<string>>|null,
-      private scopeRegistry: LocalModuleScopeRegistry) {}
+      private scopeRegistry: LocalModuleScopeRegistry, private compileNonExportedClasses: boolean) {
+  }
 
 
   get exportStatements(): Map<string, Map<string, [string, string]>> { return this.reexportMap; }
@@ -90,6 +91,10 @@ export class IvyCompilation {
   analyzeAsync(sf: ts.SourceFile): Promise<void>|undefined { return this.analyze(sf, true); }
 
   private detectHandlersForClass(node: ClassDeclaration): IvyClass|null {
+    if (!this.compileNonExportedClasses && !isExported(node)) {
+      return null;
+    }
+
     // The first step is to reflect the decorators.
     const classDecorators = this.reflector.getDecoratorsOfDeclaration(node);
     let ivyClass: IvyClass|null = null;
--- a/packages/compiler-cli/src/transformers/api.ts
+++ b/packages/compiler-cli/src/transformers/api.ts
@@ -392,6 +392,12 @@ export interface CompilerOptions extends ts.CompilerOptions {
    * support these future imports.
    */
   generateDeepReexports?: boolean;
+
+  /**
+   * Whether the compiler should avoid generating code for classes that haven't been exported.
+   * This is only active when building with `enableIvy: true`. Defaults to `true`.
+   */
+  compileNonExportedClasses?: boolean;
 }
 
 export interface CompilerHost extends ts.CompilerHost {
--- a/packages/compiler-cli/test/ngtsc/ngtsc_spec.ts
+++ b/packages/compiler-cli/test/ngtsc/ngtsc_spec.ts
@@ -5181,6 +5181,60 @@ export const Foo = Foo__PRE_R3__;
         expect(jsContents).toContain('styles: ["h1[_ngcontent-%COMP%] {font-size: larger}"]');
       });
     });
+
+    describe('non-exported classes', () => {
+      beforeEach(() => env.tsconfig({compileNonExportedClasses: false}));
+
+      it('should not emit directive definitions for non-exported classes if configured', () => {
+        env.write('test.ts', `
+          import {Directive} from '@angular/core';
+
+          @Directive({
+            selector: '[test]'
+          })
+          class TestDirective {}
+        `);
+        env.driveMain();
+        const jsContents = env.getContents('test.js');
+
+        expect(jsContents).not.toContain('defineDirective(');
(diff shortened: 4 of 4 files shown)


packages/compiler-cli/src/transformers/api.ts
```diff
@@ -392,6 +392,12 @@ export interface CompilerOptions extends ts.CompilerOptions {
    * support these future imports.
    */
   generateDeepReexports?: boolean;
+
+  /**
+   * Whether the compiler should avoid generating code for classes that haven't been exported.
+   * This is only active when building with `enableIvy: true`.
+   */
+  skipNonExportedClasses?: boolean;
```
crisbeto: Some questions for the reviewer since the ticket wasn't very descriptive:

1. I'm open to suggestion on what we should call the flag.
2. Do we want to default it to `true` so it matches ViewEngine or leave it as an opt-in?

packages/compiler-cli/ngcc/src/analysis/decoration_analyzer.ts
```diff
@@ -88,18 +88,19 @@ export class DecorationAnalyzer {
...
         /* defaultPreserveWhitespaces */ false,
-        /* i18nUseExternalIds */ true, /* i18nLegacyMessageIdFormat */ '', this.moduleResolver,
-        this.cycleAnalyzer, this.refEmitter, NOOP_DEFAULT_IMPORT_RECORDER),
+        /* i18nUseExternalIds */ true, /* i18nLegacyMessageIdFormat */ '',
+        /* skipNonExportedClasses */ false, this.moduleResolver, this.cycleAnalyzer,
+        this.refEmitter, NOOP_DEFAULT_IMPORT_RECORDER),
     new DirectiveDecoratorHandler(
         this.reflectionHost, this.evaluator, this.fullRegistry, NOOP_DEFAULT_IMPORT_RECORDER,
-        this.isCore),
+        this.isCore, /* skipNonExportedClasses */ false),
```
alxhub: I think the default here should be `true`, as ngcc only processes code which was intended for View Engine.

packages/compiler-cli/test/ngtsc/ngtsc_spec.ts
```diff
@@ -3030,6 +3030,58 @@ runInEachFileSystem(os => {
       expect(jsContents).toMatch(/directives: \[DirA,\s+DirB\]/);
     });
 
+    it('should not emit directive definitions for non-exported classes if configured', () => {
```
alxhub: Can we put these all in a `describe('non-exported classes', ...)`?

devversion:
This PR will close #33724. This issue also provides some context and a reproduction for later reference.

packages/compiler-cli/src/ngtsc/annotations/src/ng_module.ts
```diff
@@ -44,7 +44,8 @@ export class NgModuleDecoratorHandler implements DecoratorHandler<NgModuleAnalys
       private scopeRegistry: LocalModuleScopeRegistry,
       private referencesRegistry: ReferencesRegistry, private isCore: boolean,
       private routeAnalyzer: NgModuleRouteAnalyzer|null, private refEmitter: ReferenceEmitter,
-      private defaultImportRecorder: DefaultImportRecorder, private localeId?: string) {}
+      private defaultImportRecorder: DefaultImportRecorder, private skipNonExportedClasses: boolean,
```
alxhub: Rather than carry `skipNonExportedClasses` around to each of the `DecoratorHandlers` (and in fact this PR only does 2 of them - non-exported classes will still be processed as `@Injectable`s for example) we should nip this in the bud.

In `IvyCompilation` inside `analyze`, we can skip the class for all decorator handlers without passing the flag down into each one.

crisbeto:

packages/compiler-cli/src/transformers/api.ts
devversion: That's a good question. I think there is still discussion going on. I'd vote for this being opt-in since long-term it feels like we need to come up with something that works better for unit testing w/ AOT. I've explained in the original Github issue.

AndrewKushnir:
- [VE Presubmit](http://test/OCL:281782004:BASE:281782075:1574361287241:5df846bb)
- [Ivy Presubmit](https://test.corp.google.com/OCL:281782004:BASE:281782303:1574361333701:8982e1df)

packages/compiler-cli/test/ngtsc/ngtsc_spec.ts
```diff
@@ -5021,6 +5021,61 @@ export const Foo = Foo__PRE_R3__;
         expect(jsContents).toContain('styles: ["h1[_ngcontent-%COMP%] {font-size: larger}"]');
       });
     });
+
+    describe('non-exported classes', () => {
+      it('should not emit directive definitions for non-exported classes if configured', () => {
+        env.tsconfig({skipNonExportedClasses: true});
```
alxhub: This can move into a `beforeEach` inside the `describe`. That centralizes the common thread between the tests.

alxhub (approved):

packages/compiler-cli/src/transformers/api.ts
alxhub: In general, I think when a flag is `true` it should do something, not prevent something.

So instead of `skipNonExportedClasses`, let's go with `compileNonExportedClasses`, and make it default to `true`. Then if users don't want the behavior, they explicitly disable it, rather than explicitly enable skipping it.

alxhub (changes requested):

crisbeto:
Updated based on the latest set of feedback.

packages/compiler-cli/src/ngtsc/transform/src/compilation.ts
```diff
@@ -80,7 +80,7 @@ export class IvyCompilation {
       private handlers: DecoratorHandler<any, any>[], private reflector: ReflectionHost,
       private importRewriter: ImportRewriter, private incrementalState: IncrementalState,
       private perf: PerfRecorder, private sourceToFactorySymbols: Map<string, Set<string>>|null,
-      private scopeRegistry: LocalModuleScopeRegistry) {}
+      private scopeRegistry: LocalModuleScopeRegistry, private compileNonExportedClasses = true) {}
```
alxhub: No default values in compiler code. Explicitly pass in the value at the callsite.

packages/compiler-cli/src/ngtsc/program.ts
```diff
@@ -607,7 +607,7 @@ export class NgtscProgram implements api.Program {
 
     return new IvyCompilation(
         handlers, this.reflector, this.importRewriter, this.incrementalState, this.perfRecorder,
-        this.sourceToFactorySymbols, scopeRegistry);
+        this.sourceToFactorySymbols, scopeRegistry, this.options.compileNonExportedClasses);
```
alxhub:
```suggestion
        this.sourceToFactorySymbols, scopeRegistry, this.options.compileNonExportedClasses !== false);
```

alxhub:
[Presubmit](http://test/OCL:281803190:BASE:281803243:1574366708702:92f3701e)
Ivy Presubmit: not needed, no a g3 behavior change.

alxhub (approved):

AndrewKushnir:
@crisbeto g3 presubmit looks good (I removed the "presubmit" label). Could you please rebase this PR and resolve a conflict? Thank you.

crisbeto:
Rebased.

PR #33922, opened 2019-11-19
fix(ivy): ensure url-based CSS variables are sanitized

matsko:
This patch ensures that [style] bindings properly sanitize
CSS variable values before they are written to the DOM. The
reasoning for this is because CSS variables may contain `url()`
entries which are a gateway for code injection. The additional
sanitization checks now fix this.

Jira issue: [FW-1684](https://angular-team.atlassian.net/browse/FW-1684)

### DIFF
--- a/packages/core/src/sanitization/sanitization.ts
+++ b/packages/core/src/sanitization/sanitization.ts
@@ -190,15 +190,8 @@ const HYPHEN_CHAR = 45;
 export const ɵɵdefaultStyleSanitizer =
     (function(prop: string, value: string|null, mode?: StyleSanitizeMode): string | boolean | null {
       mode = mode || StyleSanitizeMode.ValidateAndSanitize;
-      let doSanitizeValue = true;
-      if (mode & StyleSanitizeMode.ValidateProperty) {
-        doSanitizeValue =
-            (prop.charCodeAt(0) === HYPHEN_CHAR && prop.charCodeAt(1) === HYPHEN_CHAR) ||
-            prop === 'background-image' || prop === 'background' || prop === 'border-image' ||
-            prop === 'filter' || prop === 'list-style' || prop === 'list-style-image' ||
-            prop === 'clip-path';
-      }
-
+      const doSanitizeValue =
+          (mode & StyleSanitizeMode.ValidateProperty) ? isStyleSanitizable(prop) : true;
       if (mode & StyleSanitizeMode.SanitizeOnly) {
         return doSanitizeValue ? ɵɵsanitizeStyle(value) : value;
       } else {
@@ -227,4 +220,13 @@ export function validateAgainstEventAttributes(name: string) {
 function getSanitizer(): Sanitizer|null {
   const lView = getLView();
   return lView && lView[SANITIZER];
+}
+
+function isStyleSanitizable(prop: string): boolean {
+  // before this function is called, the provided `prop` value should
+  // have already been converted from camel-case to hyphen-case.
+  return (prop.charCodeAt(0) === HYPHEN_CHAR && prop.charCodeAt(1) === HYPHEN_CHAR) ||
+      prop === 'background-image' || prop === 'background' || prop === 'border-image' ||
+      prop === 'filter' || prop === 'list-style' || prop === 'list-style-image' ||
+      prop === 'clip-path';
 }
--- a/packages/core/src/sanitization/sanitization.ts
+++ b/packages/core/src/sanitization/sanitization.ts
@@ -179,6 +179,8 @@ export function ɵɵsanitizeUrlOrResourceUrl(unsafeUrl: any, tag: string, prop: 
   return getUrlSanitizer(tag, prop)(unsafeUrl);
 }
 
+const HYPHEN_CHAR = 45;
+
 /**
  * The default style sanitizer will handle sanitization for style properties by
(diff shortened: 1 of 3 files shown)


packages/core/src/sanitization/sanitization.ts
```diff
@@ -190,9 +192,11 @@ export const ɵɵdefaultStyleSanitizer =
       mode = mode || StyleSanitizeMode.ValidateAndSanitize;
       let doSanitizeValue = true;
       if (mode & StyleSanitizeMode.ValidateProperty) {
-        doSanitizeValue = prop === 'background-image' || prop === 'background' ||
-            prop === 'border-image' || prop === 'filter' || prop === 'list-style' ||
-            prop === 'list-style-image' || prop === 'clip-path';
+        doSanitizeValue =
```
mhevery: It troubles me that the same code is in two places. Would you be willing to refactor this in a follow up PR?

I think we should pull the common code to a new BUILD target `packages/core/src/sanitization` and than have the `packages/compiler` depend on it

mhevery (approved):

mhevery (commented):
Could you also update the SHA text with text:
1. which bug are you fixing
2. Jira issue number if applicable.
3. Motivation and reasoning behind this change.

packages/core/src/sanitization/sanitization.ts
```diff
@@ -190,9 +192,11 @@ export const ɵɵdefaultStyleSanitizer =
...
+            (prop.charCodeAt(0) === HYPHEN_CHAR && prop.charCodeAt(1) === HYPHEN_CHAR) ||
+            prop === 'background-image' || prop === 'background' || prop === 'border-image' ||
```
mhevery: Is this case sensitive? If not we need to do more were here so that we ignore mixed case as well.

mhevery:
See: [Bazel core sanitization by mhevery · Pull Request #33926 · angular/angular](https://github.com/angular/angular/pull/33926)

packages/core/src/sanitization/sanitization.ts
matsko: We do not support uppercase style/class bindings. We should have two checks for this:

- If there are uppercase `[style.prop]` bindings then the compiler should compare upper/lower case state and see if they change. Then throw an error.
- If there are uppercase `[style]` prop values (uppercase keys in the map), then there should be a runtime `ngDevMode` gated check that should complain if the upper/lower cases do not match.

packages/core/src/render3/styling/bindings.ts
```diff
@@ -1102,7 +1102,7 @@ function removeStylingValues(
     const value = getMapValue(arr, i);
     if (value) {
       const prop = getMapProp(arr, i);
-      applyFn(renderer, element, prop, false);
+      applyFn(renderer, element, prop, null);
```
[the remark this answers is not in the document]
matsko: This was code from another PR. Rebased.

packages/core/src/sanitization/sanitization.ts
matsko: Currently VE doesn't allow uppercase style bindings. It does, however, allow uppercase class bindings: https://stackblitz.com/edit/angular-aw3oz5 (Neither of these are useful for the framework).

We should throw errors when we can in a follow-up PR.

packages/core/src/sanitization/sanitization.ts
matsko: I refactored the code to be exactly the same in both core and compiler.

packages/core/src/sanitization/sanitization.ts
mhevery: The issue is that `background-image` needs to go through serializer, but it turns out that `background-IMAGE` also works when writing to `style.cssText`. The implication of this is that `background-IMAGE` will not be sanitized, and that is a security issue.

mhevery: After thinking about this more we will have to normalize the names before passing them to this function. The issue is that if we don't then writing `WIDTH: 100px` to `cssText` will always result in reading back `width: 100px` which will cause constant thrashing. But let's solve that in separate PR.

matsko:
presubmit = https://test.corp.google.com/ui#id=OCL:281991743:BASE:281991768:1574446304655:1d76367f

ivy-presubmit = https://tap.corp.google.com/ui#id=OCL:281991743:BASE:281991894:1574446353920:4771b5b9

packages/core/test/acceptance/styling_spec.ts
```diff
@@ -2410,6 +2410,53 @@ describe('styling', () => {
           .toEqual('url("https://i.imgur.com/4AiXzf8.jpg")');
     });
   });
+
+  onlyInIvy('only ivy supports CSS variables').describe('CSS variables', () => {
+    it('should sanitize all CSS variable values that contain url() content for [style.--prop] entries',
+       () => {
+         @Component({template: `<div [style.--prop]="propValue"></div>`})
+         class MyComp {
+           propValue = 'url(javascript:alert("hahaha"))';
```
IgorMinar: I believe the combo of `javascript:` and custom properties is not possible to have working on today's browsers, so the sanitization is not necessary. Is there anything else in this PR that we want to keep? e.g. the `isStyleSanitizable` refactoring seems valuable.

IgorMinar (changes requested):

PR #33926, opened 2019-11-20
Bazel core sanitization

mhevery:
NOTE: This PR best reviewed one SHA at a time

## refactor(core): add `packages/core/sanitization` bazel target

1. Breaking up monolithic build targets into smaller builds is desirable.
2. This is motivated by code duplication between `packages/core` and `packages/compiler`. By breaking `packages/core` into smaller sub-builds it will be possible for the `packages/compiler` to directly depend on `packages/core/src/sanitization` which will than allow removal of code duplication.

While creating the smaller bundle it was discovered that these two files needed to be moved:
- `packages/core/src/sanitization/sanitizer.ts`: This file was depending on DI and as such it would prevent `packages/core/src/sanitization` from being a leaf target. Moved to `packages/cor/src/sanitizer/sanitizer.ts`
- `packages/core/src/sanitization/sanitization.ts`: This file was depending on render3 directory and as such it would prevent `packages/core/src/sanitization` from being a leaf target. Moved to `packages/cor/src/render3/sanitization.ts`. The contents of this file contains Ivy instructions and so the new location is a better fit for it.

## refactor: Remove `compiler`/`core` code duplication

The `packages/core` depends on `packages/compiler`, therefore `packgase/compiler` can not depend on `packgase/core` or it would create circular dependency.

In previous SHA `packages/core/src/sanitization` sub-build was create to allow code reuse. This allows `packages/compiler` to depend on `packages/core/src/sanitization` and reuse code.

NOTE: This is a proof of concept PR, there are more places where code reuse could be implemented.

---

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No


packages/core/src/sanitization/BUILD.bazel
```diff
@@ -0,0 +1,19 @@
+load("//tools:defaults.bzl", "ts_library")
+
+package(default_visibility = [
+    "//packages/core:__subpackages__",
+    "//tools/public_api_guard:__pkg__",
```
IgorMinar: I don't believe that you need this. Can you remove?

packages/core/src/sanitization/url_sanitizer.ts
```diff
@@ -36,9 +36,6 @@ import {isDevMode} from '../util/is_dev_mode';
  */
 const SAFE_URL_PATTERN = /^(?:(?:https?|mailto|ftp|tel|file):|[^&:/?#]*(?:[/?#]|$))/gi;
 
-/* A pattern that matches safe srcset values */
-const SAFE_SRCSET_PATTERN = /^(?:(?:https?|file):|[^&:/?#]*(?:[/?#]|$))/gi;
-
```
IgorMinar: yay! dead code elimination!

packages/compiler/src/render3/view/styling_builder.ts
```diff
@@ -5,6 +5,8 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
+import {isStyleSanitizable} from '@angular/core/src/sanitization/url_sanitizer';
```
IgorMinar: this is a deep import into a path that doesn't exist at runtime.

keep in mind that we'd need to support at least 2 different runtimes:
- node.js
- browser

and for both runtimes (now that node.js supports or will soon support ESM) we need to support two module formats:
- UMD
- ESM

Deep imports are simply a no go with the current set of packaging constraints we have, which leaves you with the only known viable option to be private exports, but that doesn't solve the compilation issues breaking things up into smaller compilation units.

IgorMinar (changes requested):
Deep import across entry-points won't work as proposed.

It might still be worth breaking up the build targets though

packages/compiler/src/render3/view/styling_builder.ts
alxhub: We could have an entrypoint `@packages/core/api` or similar, which is published across all formats and only built with `tsc`, that the compiler _could_ depend on, I believe. Would that be a feasible alternative?

The last time I checked though this was challenging to do using our Bazel packaging system.

packages/core/src/sanitization/BUILD.bazel
mhevery: removed

packages/compiler/src/render3/view/styling_builder.ts
mhevery: What do you suggest as a solution then?

mhevery: Why does there have to be a package? Is there a way to tell the rollup to just inline this particular import?

mhevery: What about changing it to `import {isStyleSanitizable} from '../../../../core/src/sanitization/url_sanitizer';` would that tell rollup to just inline ?

PR #33928, opened 2019-11-20
feat(language-service): completions support for tuple array

ivanwonder:
implement the missing tuple functionality because the PR #29811 has been closed.
https://github.com/angular/angular/pull/29811#issuecomment-554812436

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/packages/language-service/src/symbols.ts
+++ b/packages/language-service/src/symbols.ts
@@ -116,9 +116,12 @@ export interface Symbol {
 
   /**
    * Return the type of the expression if this symbol is indexed by `argument`.
+   * Sometimes we need the key of arguments to get the type of the expression, for example
+   * in the case of tuples (`type Example = [string, number]`).
+   * [string, number]).
    * If the symbol cannot be indexed, this method should return `undefined`.
    */
-  indexed(argument: Symbol): Symbol|undefined;
+  indexed(argument: Symbol, key?: any): Symbol|undefined;
 }
 
 /**
--- a/packages/language-service/src/typescript_symbols.ts
+++ b/packages/language-service/src/typescript_symbols.ts
@@ -295,7 +295,8 @@ class TypeWrapper implements Symbol {
         if (nType) {
           // get the right tuple type by value, like 'var t: [number, string];'
           if (nType.isUnion()) {
-            return new TypeWrapper(nType.types[value], this.context);
+            // return undefined if array index out of bound.
+            return nType.types[value] && new TypeWrapper(nType.types[value], this.context);
           }
           return new TypeWrapper(nType, this.context);
         }
--- a/packages/language-service/src/typescript_symbols.ts
+++ b/packages/language-service/src/typescript_symbols.ts
@@ -284,7 +284,7 @@ class TypeWrapper implements Symbol {
     return selectSignature(this.tsType, this.context, types);
   }
 
-  indexed(argument: Symbol): Symbol|undefined {
+  indexed(argument: Symbol, value: any): Symbol|undefined {
     const type = argument instanceof TypeWrapper ? argument : argument.type;
     if (!(type instanceof TypeWrapper)) return;
 
@@ -292,7 +292,14 @@ class TypeWrapper implements Symbol {
     switch (typeKind) {
       case BuiltinType.Number:
         const nType = this.tsType.getNumberIndexType();
-        return nType && new TypeWrapper(nType, this.context);
+        if (nType) {
+          // get the right tuple type by value, like 'var t: [number, string];'
+          if (nType.isUnion()) {
+            return new TypeWrapper(nType.types[value], this.context);
+          }
+          return new TypeWrapper(nType, this.context);
+        }
+        return undefined;
       case BuiltinType.String:
         const sType = this.tsType.getStringIndexType();
         return sType && new TypeWrapper(sType, this.context);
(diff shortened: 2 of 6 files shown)


packages/language-service/src/symbols.ts
```diff
@@ -116,9 +116,11 @@ export interface Symbol {
 
   /**
    * Return the type of the expression if this symbol is indexed by `argument`.
+   * sometimes we need the key of arguments to get the type of the expression(type Example =
```
ayazhafiz:
```suggestion
   * Sometimes we need the key of arguments to get the type of the expression, for example
   * in the case of tuples (`type Example = [string, number]`).
```

packages/language-service/src/typescript_symbols.ts
```diff
@@ -280,15 +280,23 @@ class TypeWrapper implements Symbol {
...
     const type = argument instanceof TypeWrapper ? argument : argument.type;
     if (!(type instanceof TypeWrapper)) return;
 
     const typeKind = typeKindOf(type.tsType);
     switch (typeKind) {
       case BuiltinType.Number:
         const nType = this.tsType.getNumberIndexType();
-        return nType && new TypeWrapper(nType, this.context);
+        if (nType) {
+          // get the right tuple Types by value. like 'var t: [number, string];'
```
ayazhafiz:
```suggestion
          // get the right tuple type by value, like 'var t: [number, string];'
```

packages/language-service/test/completions_spec.ts
```diff
@@ -124,6 +124,13 @@ describe('completions', () => {
     expectContain(completions, CompletionKind.PROPERTY, ['id', 'name']);
   });
 
+  it('should be able to get property completions for members in an tuple array', () => {
+    mockHost.override(TEST_TEMPLATE, `{{ tupleArray[1].~{tuple-array-number-index}}}`);
+    const marker = mockHost.getLocationMarkerFor(TEST_TEMPLATE, 'tuple-array-number-index');
+    const completions = ngLS.getCompletionsAt(TEST_TEMPLATE, marker.start);
+    expectContain(completions, CompletionKind.PROPERTY, ['id', 'name']);
+  });
+
```
ayazhafiz: Could you a test in `packages/language-service/test/diagnostics_spec.ts` as well? There are already tests in there for index types, so you should be able to mostly just copy one.

ayazhafiz (changes requested):

packages/language-service/src/typescript_symbols.ts
```diff
@@ -280,15 +280,23 @@ class TypeWrapper implements Symbol {
...
 
     const typeKind = typeKindOf(type.tsType);
     switch (typeKind) {
       case BuiltinType.Number:
         const nType = this.tsType.getNumberIndexType();
-        return nType && new TypeWrapper(nType, this.context);
+        if (nType) {
+          // get the right tuple type by value, like 'var t: [number, string];'
+          if (nType.isUnion() && typeof value === 'number') {
+            return new TypeWrapper(nType.types[value], this.context);
```
kyliau: should we do a bounds check here?

packages/language-service/src/typescript_symbols.ts
```diff
@@ -280,15 +280,23 @@ class TypeWrapper implements Symbol {
...
+          }
+          return new TypeWrapper(nType, this.context);
+        } else {
+          return undefined;
```
kyliau: else statement and return could be omitted.

kyliau (changes requested):
Thank you very much for fixing this!

kyliau:
The commit message linter doesn't like the commit messages, could you please squash them all into one?

packages/language-service/src/typescript_symbols.ts
ivanwonder: because the tslint report a warning `Fallthrough case in switch.ts(7029)`.
if it doesn't matter. I will remove the else statement.

ivanwonder:
> The commit message linter doesn't like the commit messages, could you please squash them all into one?

ok, I will

packages/language-service/src/typescript_symbols.ts
ivanwonder: ok, I will remove the number type checker

ayazhafiz (approved):

packages/language-service/src/typescript_symbols.ts
kyliau: Sorry, I should have been more clear in my previous comment.
What I meant was, what if `nType.types[value]` is undefined?
For example, if the tuple is `t = [string, number, string]` and user indexes `t[3]`, would this report an error?
Perhaps we should add a test case for this.

ivanwonder: I see, here I will return `return nType.types[value] && new TypeWrapper(nType.types[value], this.context)`.
and then add a test case in `diagnostics_spec.ts` to check `expect(diags).toEqual([]);` if array index out of bound.

kyliau (approved):
Thank you so much!

ivanwonder:
love the language-service, and hope it better.🤝

PR #33935, opened 2019-11-20
build: change extraction of web_test_archives to run phase

josephperrott:
By moving extract to the run value, we only extract it once we are sure
it will be used.  With the build value, it is extracted during the build
phase, so we might upload it before the build completes.  If the build
fails along the way, we have uploaded it for "no reason."  By moving to
run we ensure that we only upload after we are sure something builds,
and increase the likelihood of action cache hits.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -794,101 +794,101 @@ workflows:
           filters:
             branches:
               ignore: g3
-      - lint:
-          requires:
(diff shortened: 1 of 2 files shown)


josephperrott:
Closing as it appears that the extraction needs to occur during `build` step for the chromedriver to be available correctly.

PR #33937, opened 2019-11-20
test(docs-infra): correctly clean up after `ScrollService` tests

gkalpak:
`ScrollService` subscribes to global `window` events and mutates global state in the listener (e.g. read/write values from/to `sessionStorage`). Therefore, we need to always call its `ngOnDestroy()` method to unsubscribe from these events after each test.

In f69c6e2, a new testcase was introduced that was not destroyed. As a result, random failures started to randomly happen in other, unrelated tests ([example CI failure][1]).

This commit fixes this by ensuring all `ScrollService` instances are destroyed after each tests (provided that they are created with the `createScrollService()` helper).

##
This commit also clears `sessionStorage` and unsubscribes from `Location` events, after each test to reduce the possibility of potential [spooky action at a distance][2]-type of failures in the future.

This does not have an impact on the actual app, since `ScrollService` is currently expected to live throughout the lifetime of the app. Still, unsubscribing from `Location` events keeps the code consistent with how other `ScrollService` listeners are handled (e.g. for `window` events).

[1]: https://circleci.com/gh/angular/angular/533298
[2]: https://en.wikipedia.org/wiki/Action_at_a_distance_(computer_programming)

### DIFF
--- a/aio/src/app/shared/scroll.service.spec.ts
+++ b/aio/src/app/shared/scroll.service.spec.ts
@@ -62,7 +62,10 @@ describe('ScrollService', () => {
     spyOn(window, 'scrollBy');
   });
 
-  afterEach(() => scrollServiceInstances.forEach(instance => instance.ngOnDestroy()));
+  afterEach(() => {
+    scrollServiceInstances.forEach(instance => instance.ngOnDestroy());
+    window.sessionStorage.clear();
+  });
 
   it('should debounce `updateScrollPositonInHistory()`', fakeAsync(() => {
     const updateScrollPositionInHistorySpy = spyOn(scrollService, 'updateScrollPositionInHistory');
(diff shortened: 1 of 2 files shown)


aio/src/app/shared/scroll.service.spec.ts
```diff
@@ -422,4 +425,58 @@ describe('ScrollService', () => {
...
+      expect(updateScrollLocationHrefSpy).toHaveBeenCalledTimes(2);
+
+      updateScrollLocationHrefSpy.calls.reset();
+      scrollService.ngOnDestroy();
+
+      window.dispatchEvent(new Event('beforeunload'));
+      expect(updateScrollLocationHrefSpy).not.toHaveBeenCalled();
+    });
+
+    it('should stop scrolling on `hashchage` events', () => {
```
petebacondarwin:
```suggestion
    it('should stop scrolling on `hashchange` events', () => {
```

petebacondarwin (approved):

PR #33948, opened 2019-11-21
fix(ivy): remove TNodeType assertion from `directiveInject` instruction

AndrewKushnir:
The assertion that we have in the `directiveInject` instruction is too restrictive and we came across some pattern where it throws unnecessarily. This commit removes that assertion for now and more detailed investigation is needed to decide is we need to restrict the set of TNodeType again.

This commit also adds a test which triggered the TNodeType.View to come up in the `directiveInject` instruction, so it might be useful to avoid regressions during further refactoring.

This PR resolves FW-1734.

## PR Type

- [x] Bugfix

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/render3/instructions/di.ts
+++ b/packages/core/src/render3/instructions/di.ts
@@ -46,9 +46,6 @@ export function ɵɵdirectiveInject<T>(
   // if inject utilities are used before bootstrapping.
   if (lView == null) return ɵɵinject(token, flags);
   const tNode = getPreviousOrParentTNode();
-  ngDevMode && assertNodeOfPossibleTypes(
-                   tNode, TNodeType.Container, TNodeType.Element, TNodeType.ElementContainer,
-                   TNodeType.IcuContainer);
   return getOrCreateInjectable<T>(
       tNode as TDirectiveHostNode, lView, resolveForwardRef(token), flags);
 }
--- a/packages/core/test/acceptance/di_spec.ts
+++ b/packages/core/test/acceptance/di_spec.ts
@@ -1859,4 +1859,46 @@ describe('di', () => {
 
     expect(fixture.nativeElement.innerHTML).toContain('Other value is: 2 (transformed)');
   });
+
+  it('should support dependencies in Pipes used inside i18n blocks', () => {
+    @Injectable()
+    class MyService {
+      transform(value: string): string { return `${value} (transformed)`; }
+    }
+
+    @Pipe({name: 'somePipe'})
+    class MyPipe {
+      constructor(private service: MyService) {}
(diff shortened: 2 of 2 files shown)


pkozlowski-opensource:
@AndrewKushnir thnx for the PR. I'm so sorry that I've caused so much trouble with the assertion I've added in the DI code. If you ask me now, I would remove it altogether...

Having said this I _think_ that this assert _is_ uncovering a bug / inconsistency that we should look into after the dust settles on V9. What I mean here is that my understanding of the DI code is that it _assumes_ that, when `ɵɵdirectiveInject` is called, the `previousOrParentTNode` points to a TNode on which the injection happens. 

My understanding is that i18n code is violating this assumptions. Let's have a look at the generated code for the tests you've added (corresponding to `<ng-template #source i18n>{{count | somePipe}}<span>items</span></ng-template>`):

```
 if (rf & 1) {
    ɵɵi18nStart_0(0,I18N_2);
      ɵɵpipe_1(1,'somePipe');
      ɵɵelement_2(2,'span');
    ɵɵi18nEnd_3();
  }
``` 

The assertion in question blows up here since, in pipe injection, we go past the i18nStart and up on the TView node. What is surprising things to me here is that a pipe doesn't "sit" on top of a text node (`previousOrParentTNode` would point to a text node in case of a regular pipe in interpolated).

The most probable issue here is that I lack understanding of how i18n is supposed to interact with DI so I would say:
* let's remove the assert altogether (at this point we assert on pretty much any node type... )
* let's look into more post v9 - if there is anything to look into...

packages/core/src/render3/instructions/di.ts
```diff
@@ -48,7 +48,7 @@ export function ɵɵdirectiveInject<T>(
   const tNode = getPreviousOrParentTNode();
   ngDevMode && assertNodeOfPossibleTypes(
```
pkozlowski-opensource: At this point I would probably remove this assert altogether as we start to list each and every node type here... 
This is unfortunate since we _still_ do a cast - see my other comment in the PR for more details.

AndrewKushnir: Thanks for additional investigation @pkozlowski-opensource. As we discussed, I've removed the assert for now. Thank you.

pkozlowski-opensource (approved):
LGTM

AndrewKushnir:
- [VE Presubmit](https://test.corp.google.com/ui#id=OCL:281805541:BASE:281805572:1574367479857:5e493517)
- [Ivy Presubmit](https://test.corp.google.com/ui#id=OCL:281805541:BASE:281805677:1574367514051:9d2572df)

PR #33949, opened 2019-11-21
fix(core): remove deprecated and defunct wtf* apis

### LINKED ISSUE #12509: Web Tracing Framework no longer Working in Angular

```
[X] bug report => search github for a similar issue or PR before submitting
```

**Current behavior**
1. Navigate to a angular 2 application such as: https://material2-app.firebaseapp.com/
2. Turn on [WTF](http://google.github.io/tracing-framework/)
3. notice that the application goes into a constant refresh loop (i.e. app renders then refreshes etc)

**Expected behavior**
Should come up with WTF UI.

1. Navigate to a angular 2 application such as: https://material2-app.firebaseapp.com/
2. Turn on [WTF](http://google.github.io/tracing-framework/)
3. notice that the application goes into a constant refresh loop (i.e. app renders then refreshes etc)

- **Angular version:** 2.0.X
  
  
- **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]
  
  
- **Language:** [all | TypeScript X.X | ES6/7 | ES5]
- **Node (for AoT issues):** `node --version` =

IgorMinar:
These apis have been deprecated in v8, but since they are defunct we are removing them early so that they don't take up payload size.

Closes #12509


kara (approved):
LGTM

JiaLiPassion:
@IgorMinar , should we also remove `wtfZoneSpec` from `ngZone` here https://github.com/angular/angular/blob/master/packages/core/src/zone/ng_zone.ts#L133?

IgorMinar:
hmm @JiaLiPassion I didn't realize that there was a wtf specific zone spec.

Is anyone using it? this zone spec is separate from exposing the wtf apis so I think it's ok to remove stuff from core, but keep wtf support in zone.js.

IgorMinar:
thanks for pointing this out btw

IgorMinar:
I'm marking this as blocked until we agree on the zone.js spec story

IgorMinar:
looking into this more, I think we should go ahead with this change as is without removing the zone.js integration in ng_zone. That can be done separately in the future.

IgorMinar:

matsko:
presubmit = http://test/OCL:281862610:BASE:281862635:1574383796718:82936f1b

JiaLiPassion:
@IgorMinar , yeah, sure, there is no harm to leave the `wtfZoneSpec` in ngZone for now, and I don't think any one are really using it because `wtf` itself is not continued. Thanks for looking into it.

matsko:
presubmit = https://tap.corp.google.com/ui#id=OCL:281862610:BASE:281985927:1574444478051:6a401beb

ivy-presubmit = https://test.corp.google.com/ui#id=OCL:281862610:BASE:281986438:1574444628677:e6a8c4d4

JiaLiPassion (approved):

PR #33955, opened 2019-11-21
build: update to Angular CLI 9.0.0-rc.3

filipesilva:
Followup to https://github.com/angular/angular/pull/33337 and https://github.com/angular/angular-cli/pull/16228

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No


package.json
```diff
@@ -35,10 +35,10 @@
...
   "// 1": "dependencies are used locally and by bazel",
   "dependencies": {
-    "@angular-devkit/architect": "^0.900.0-rc.0",
-    "@angular-devkit/build-optimizer": "^0.900.0-rc.0",
-    "@angular-devkit/core": "^9.0.0-rc.0",
-    "@angular-devkit/schematics": "^9.0.0-rc.0",
+    "@angular-devkit/architect": "^0.900.0-rc.3",
+    "@angular-devkit/build-optimizer": "^0.900.0-rc.3",
+    "@angular-devkit/core": "^9.0.0-rc.3",
+    "@angular-devkit/schematics": "^9.0.0-rc.3",
```
alan-agius4: Nit: it seems that `@schematics/angular` hasn't been updated for a long time as it's still on `8.0.0-beta.15`. Can you please update it?

aio/scripts/_payload-limits.json
```diff
@@ -12,7 +12,7 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 2987,
-        "main-es2015": 506857,
+        "main-es2015": 464734,
```
gkalpak: Context:
This was temporarily changed from `461159` to `506857` (pending a cli release) in 08a4f10ee.
So, there seems to be ~3.5KB regression (`461159` --> `464734`) from rc.1 to rc.3.

gkalpak (approved):

package.json
filipesilva: Done

aio/scripts/_payload-limits.json
filipesilva: Hard to say actually. It's true that the commit you linked changed the value. But that wasn't the absolute value, it was a value within +-1% of the real value.

Which is to say, I could have put that value and it would have still passed the CI. To know if there's a regression or not we'd need to check the sizes of FW+CLI rc.2 and FW+CLI rc.3 for AIO. 

It's unfortunate that we can't vary just one. Next time we do this sort of coordinated change we should instead leave one of the packages (either FW or CLI) in a intermediate compat mode (@clydin suggested this approach for the change in question, but we ended up not going with it).

gkalpak: You are right. Moreover, this size is for building with the local packages (so neither rc.1 nor rc.3) :grin:
So, my comment was invalid 😇

filipesilva: Did both builds:
- FW+CLI rc.2: 464103
- FW+CLI rc.3: 464814

So yes, there was a change of 701 bytes. I don't know if it was CLI or the FW though.

alxhub (approved):

PR #33958, opened 2019-11-21
refactor(core): update undecorated-classes-with-di migration rerun command

alan-agius4:
With Angular CLI version 9 RC 3 we can run a single migration for a package using the name of the migration schematic.

We need to pass the schematic name as a value to the `migrate-only` option.
Ex:
```
ng update @angular/core --migrate-only migration-v9-undecorated-classes-with-di
```

See: https://github.com/angular/angular-cli/pull/16174

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/core/schematics/migrations/undecorated-classes-with-di/index.ts
+++ b/packages/core/schematics/migrations/undecorated-classes-with-di/index.ts
@@ -24,7 +24,7 @@ import {UndecoratedClassesTransform} from './transform';
 import {UpdateRecorder} from './update_recorder';
 
 const MIGRATION_RERUN_MESSAGE = 'Migration can be rerun with: "ng update @angular/core ' +
-    '--from 8.0.0 --to 9.0.0 --migrate-only"';
+    '--migrate-only migration-v9-undecorated-classes-with-di"';
 
 const MIGRATION_AOT_FAILURE = 'This migration uses the Angular compiler internally and ' +
     'therefore projects that no longer build successfully after the update cannot run ' +


devversion (commented):
Nice. That is a very useful feature :100: For this transition phase though, can we be sure that people have the latest CLI version installed? I feel like we cover more cases/versions with the old command right now? Nothing wrong with re-running the other migrations as a noop.

devversion:
Also, is this really a fix? This looks more like it would fit into the `refactor` scope.

alan-agius4:
During `ng update` the latest CLI is always installed.

Updated the commit msg

devversion:
@alan-agius4 Is this new behavior? because when I do `ng update @angular/material`, it only updates Angular Material and the Angular CDK. It doesn't necessarily enforce me to update the CLI too?

alan-agius4:
It was introduced in 8.3.13 actually.

The CLI needs to be updated and be on the same version of the FW.

alan-agius4:
Also as part of the version 9 update we are suggesting that users update to the latest version 8 patch

> Run ng update @angular/core@8 @angular/cli@8 in your workspace directory to update to the latest 8.x version of @angular/core and @angular/cli and commit these changes.

https://update.angular.io/#8.0:9.0

devversion (approved):
Considering that this logic already existed in version 8, I think it's okay to show the new command.

<sup>Note: My point was mainly: why risk making it not work for some people if we can just keep the existing command that works _always_</sup>

mhevery (approved):

IgorMinar (approved):
thanks!

PR #33960, opened 2019-11-21
docs: update injectable migration guide to reflect latest schematic changes

devversion:
Since we created the migration guide for the `missing-injectable` schematic, the schematic
changed in various ways. e.g. the migration no longer migrates classes passed to `useExisting`

Additionally the migration has been expanded to handle another Ivy breaking change where
providers like `{provide: X}` will be intepreted as `{provide: X, useClass: X}`. This pattern should
be documented in the migration guide.

### DIFF
--- a/aio/content/guide/migration-injectable.md
+++ b/aio/content/guide/migration-injectable.md
@@ -1,64 +1,65 @@
 # Migration to Add Missing `@Injectable()` Decorators
 
-## What does this schematic do?
+### What does this schematic do?
 
-This schematic adds an `@Injectable()` decorator to a class
-if the class has been added as a provider anywhere in the application.
+  1. This schematic adds an `@Injectable()` decorator to classes which are provided in the
+     application but are not decorated.
+  2. The schematic updates providers which follow the `{provide: SomeToken}` pattern
+     to explicitly specify `useValue: undefined`.
+     
+**Example for missing `@Injectable()`**
 
-An example diff might look like the following:
+_Before migration:_
+```typescript
+export class MyService {...}
+export class MyOtherService {...}
+export class MyThirdClass {...}
+export class MyFourthClass {...}
+export class MyFifthClass {...}
 
-**Before:**
-
-```ts
-export class TypeCase {...}
+@NgModule({
+  providers: [
+    MyService,
+    {provide: SOME_TOKEN, useClass: MyOtherService},
+    // The following classes do not need to be decorated because they
+    // are never instantiated and just serve as DI tokens.
+    {provide: MyThirdClass, useValue: ...},
+    {provide: MyFourthClass, useFactory: ...},
+    {provide: MyFifthClass, useExisting: ...},
+  ]
+})
 ```
-
-**After:**
-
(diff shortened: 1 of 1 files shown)


devversion:
Initially I only wanted to add the new breaking change that will be handled, but I figured that we need to restructure things a bit since the guide is no longer only about _Missing injectable decorators_. Hope that is fine. I tried my best incorporating these two "schematic changes" into the guide.

aio/content/guide/migration-injectable.md
```diff
@@ -71,15 +72,37 @@ to strictly require the decorator because doing so enables further
 optimization of both the compiler and the runtime. This schematic
 adds any `@Injectable()` decorators that may be missing to future-proof your app.
 
+### Why is adding `useValue: undefined` necessary?
+
```
kapunahelewong: Maybe add an introductory sentence before the code snippet? Something like, "Consider the following pattern:"

You'd then probably want to change "Providers following this pattern..." to "Providers using this pattern..." in the next sentence just to avoid repetition.

aio/content/guide/migration-injectable.md
```diff
@@ -71,15 +72,37 @@ to strictly require the decorator because doing so enables further
...
+the `@Injectable()` decorator. In a future version of Angular, a missing `@Injectable()` decorator
+will always throw an error.
 
-## Should I update my library?
+Additionally, providers in your library that follow the described `{provide: X}` pattern should be
+updated to specify an explicit value. Without explicit value, these providers can behave
+differently based on the Angular version in applications consuming your library.
 
-Yes, if your library has any tokens that are meant to be injected, they should be updated with the `@Injectable()` decorator. In a future version of Angular, a missing `@Injectable()` decorator will always throw an error.
+[DI_TOKEN]: guide/glossary#di-token
```
kapunahelewong: How about having the reference in a sentence like this:

> For more information, see the [DI token](url) entry in the glossary.

I'd write DI token like this (instead of "DI_TOKEN") because that's how it is in the glossary. If it should instead be "DI_TOKEN" then we should change it in the glossary to reflect this way of writing it. I vote for consistency in either case.

kapunahelewong (approved):
Very nice. Thank you. 😊I have only a couple of suggestions.

aio/content/guide/migration-injectable.md
devversion: Great idea. Will do 👍

aio/content/guide/migration-injectable.md
devversion: This line is not display in the rendered page. It just serves a "variable" for the DI token glossary link. The "variable" will then be used in the page like followed: `The [DI token][DI_TOKEN] is used ...`

or do you mean something else?

kapunahelewong: Oh! I see, never mind.

IgorMinar (approved):
lgtm

IgorMinar:
this looks good to me, we need this info in the next RC, so I'll label it for merge.

PR #33968, opened 2019-11-21
build: update lockfiles for integration projects

gkalpak:
In the `integration_test` CircleCI job, we run `yarn install` on all projects in the `integration/` directory. If a project has no lockfile or if the lockfile is out-of-sync with the corresponding `package.json` file, then the installed dependency versions are no longer pinned, which can result in different versions being installed between different runs of the same job (if, for example, a new version is released for a package) and breaks hermeticity.

This could be prevented by using the `--frozen-lockfile` option with `yarn install`, but this is not possible with the current setup, because yarn needs to be able to install the locally built Angular packages, whose checksums will be different from the ones in the lockfile. Therefore, we have to manually ensure that the lockfiles remain in-sync with the corresponding `package.json` files for the rest of the dependencies.

For example, previously, [cli-hello-world-lazy/yarn.lock][1] had an entry for `@angular-devkit/build-angular@0.900.0-next.9` (pinned to `0.900.0-next.9`), but [cli-hello-world-lazy/package.json][2] specified the `@angular-devkit/build-angular` version as `^0.900.0-rc.0` (note the leading caret). As a result, since the version in the lock file does not much the one in `package.json`, the lockfile is ignored and the latest available version that matches `^0.900.0-rc.0` is installed.

This, for example, started causing unrelated CI failures ([example][3]), when `@angular-devkit/build-angular@9.0.0-rc.3` was released with a size improvement.

This PR does a number of things to guard against such failures (although still not full-proof):
- Ensure that all integration projects have a lockfile and that lockfiles are up-to-date (with the current `package.json` files).
- Use exact versions (instead of version ranges) for integration project dependencies, to reduce the probability of installing breaking versions of (direct or transitive) dependencies, in case of an out-of-sync lockfile).
- Add a script to ensure that the above remains true in the future.

See the individual commits for details.

[1]: https://github.com/angular/angular/blob/fc2f6b845/integration/cli-hello-world-lazy/yarn.lock#L13
[2]: https://github.com/angular/angular/blob/fc2f6b845/integration/cli-hello-world-lazy/package.json#L26
[3]: https://circleci.com/gh/angular/angular/535959#tests/containers/2


filipesilva (approved):
Since we don't have a way to enforce this, I think it's just a matter of time until they get out of sync again.

gkalpak:
True. I'm buying us some time, until we figure out a good way to enforce this 😛
(Any ideas?)

filipesilva:
I find it surprising that the changing FW versions are responsible for the classes of bugs we've seen. FW packages are mostly "leaf" dependencies. Changing them shouldn't cause dependent deps to change.

I'm not surprised that depending on `../../node_modules/something` causes dramatic changes in the lockfile though. We're essentially saying "this is a dynamic version determined by another lockfile".

Could that be the real cause of unwanted updates?

gkalpak:
Not sure what classes of bugs you/we've seen :grin:
Regarding the specific bug we saw the other day, which lead to PRs started to failing [like this][1], the issue was an incorrectly pinned `@angular-devkit/build-angular` dependency (and that is not related to FW packages).

Are you referring to some other type bugs?

[1]: https://circleci.com/gh/angular/angular/535959#tests/containers/2

filipesilva:
I'm talking about that type of problem. It's not obvious to me why yarn went and updated `@angular-devkit/build-angular` just because FW packages changed. But that ends up being an implementation detail of yarn.

We could also pin the direct deps themselves. That would help in part, but not completely.

gkalpak:
> It's not obvious to me why yarn went and updated `@angular-devkit/build-angular` just because FW packages changed.

No, that's not what happened. `@angular-devkit/build-angular` was updated because it was not pinned correctly: `cli-hello-world-lazy/yarn.lock` had an entry for `@angular-devkit/build-angular@0.900.0-next.9`, but `cli-hello-world-lazy/package.json` specified the `@angular-devkit/build-angular` version as `^0.900.0-rc.0` (note the leading caret). So, since the lockfile did not contain an entry for the version specified in `package.json`, yarn fetched the latest available version that satisfied the `package.json` version.

filipesilva:
Oh I see now. Then it makes sense to pin them in package.json, no?

gkalpak:
Exactly!

gkalpak:
Oh, wait, what you mean by pin them in `package.json`. We are pinning them in `yarn.lock`.
Do you mean using specific versions (e.g. `9.0.0`) instead of ranges (e.g. `^9.0.0`)?

filipesilva:
Yes, that is what I meant. It sounds better than pinning in the lockfile, since we can control it better. Am I missing something?

gkalpak:
Good idea! That's another line of defense. (It is not as full-proof as pinning in the lockfile (due to transitive deps being unpinned), but it reduces the possibility of breakages in case the lockfile gets out of sync.)

And we can at least enforce that. Let me update the PR.

filipesilva:
Yeah that's true. Pinning in both offers the best defense.

gkalpak:
@filipesilva: Updated the PR. PTAL

filipesilva (approved):
We'll need discipline on part of the reviewers in order to prevent PRs from unpinning these deps.

gkalpak:
46b6754e9 should catch most of the unpinnings, no?

filipesilva:
It should, and I totally knew that. I was just testing you!

integration/ivy-i18n/package.json
```diff
@@ -63,7 +61,9 @@
     "karma-coverage-istanbul-reporter": "2.1.0",
     "karma-jasmine": "2.0.1",
     "karma-jasmine-html-reporter": "1.4.2",
+    "npm-run-all": "4.1.5",
     "protractor": "file:../../node_modules/protractor",
+    "serve": "11.2.0",
```
IgorMinar: shouldn't there be a lockfile change for this? or devDep vs dep is not captured in the lockfile?

integration/check-dependencies.js
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env node
+'use strict';
```
IgorMinar: can you add a comment explaining what this script does? thanks

IgorMinar (approved):
small nitpicks, but overall this looks like a great improvement.

just a note that #34002 will collide with this one (but that's fine, I'll rebase mine), but also @gregmagolan's #33927 will collide, and we should consider rethinking this PR in the world where these tests run in Bazel. I'll let you and @gregmagolan figure out how to reconcile the differences between these two PRs.

gregmagolan (approved):
LGTM.

This will conflict with my https://github.com/angular/angular/pull/33927 so I'll let this land first and then rebase & sort out the best approach for keeping lock files up-to-date with most of the integration tests are run by Bazel.

integration/ivy-i18n/package.json
gkalpak: Yeah, `yarn.lock` does _not_ capture where the dependency came from (`dependencies` or `devDependencies`). So, it is expected that there is no lockfile change.

integration/check-dependencies.js
gkalpak: Done.

PR #33969, opened 2019-11-21
Size tracking improvements

kara:
*ci: tighten code owners for size tracking*

It will be easier to track regressions in size if fewer people
are approving size diffs in PRs. That way, we will have a few
people that have a fuller picture of where size changes are
coming from.
 
*ci: tighten size threshold to 1% or 500 bytes*

The size diff threshold of 1% has proven to be too lenient for us
to catch size regressions in AIO. Since the AIO main bundle is
between 400-500 KB, a size regression must be between 4-5 KB before
it will cause the tests to fail. As a result, we may merge many
changes with smaller regressions of a few KB before the size test
eventually lets us know that the number has increased. The hope is
that lowering the threshold will help us catch the smaller
regressions during code review and prevent the size tests failing at
a random later time when someone catches the size "hot potato".

---

Screen Shot 2019-11-21 at 5 17 57 PM

### DIFF
--- a/scripts/ci/payload-size.js
+++ b/scripts/ci/payload-size.js
@@ -28,16 +28,20 @@ for (const compressionType in limitSizes) {
         console.log(
             `Commit ${commit} ${compressionType} ${filename} meassurement is missing. ` +
             'Maybe the file was renamed or removed.');
-      } else if (Math.abs(actualSize - expectedSize) > expectedSize / 100) {
-        failed = true;
-        // We must also catch when the size is significantly lower than the payload limit, so
-        // we are forced to update the expected payload number when the payload size reduces.
-        // Otherwise, we won't be able to catch future regressions that happen to be below
-        // the artificially inflated limit.
-        const operator = actualSize > expectedSize ? 'exceeded' : 'fell below';
-        console.log(
-            `Commit ${commit} ${compressionType} ${filename} ${operator} expected size by >1% ` +
+      } else {
+        const absoluteSizeDiff = Math.abs(actualSize - expectedSize);
+        // If size diff is larger than 1% or 500 bytes...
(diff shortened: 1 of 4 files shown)


scripts/ci/payload-size.js
```diff
@@ -47,5 +51,5 @@ if (failed) {
   console.log(`If this is a desired change, please update the size limits in file '${limitFile}'.`);
   process.exit(1);
 } else {
-  console.log('Payload size <1% change check passed.');
+  console.log('Payload size check passed. The diff is less than 1% or 500 bytes.');
```
pkozlowski-opensource: Could we also print the actual diff here? I wanted to see it every time I saw this message... (should have changed it myself...)

kara: Ohh, that's a great idea. Will fix.

kara: https://github.com/angular/angular/pull/33969/commits/89e40d4e9fc33439f6a3da50b1a4d72cad3c1039

IgorMinar (approved):

IgorMinar (approved):
CI is not happy, but the changes look good to me

gkalpak (approved):

scripts/ci/payload-size.js
gkalpak: @pkozlowski-opensource, fyi we are also printing the sizes (including gzips) in [scripts/ci/payload-size.sh][1], so you can see the actual sizes for past builds.

<sub>(I know this is not the same as showing the diff, just mentioning it in case you've missed it.)</sub>

[1]: https://github.com/angular/angular/blob/828e5c1f7/scripts/ci/payload-size.sh#L34

pkozlowski-opensource (approved):
LGTM! Additional comments and separating success / failure messages are  👍 🥇

PR #33987, opened 2019-11-22
ci: fix tracking and checking of payload size data

gkalpak:
This PR mainly addresses two issues:
- Ensure that payload size data for angular.io is always uploaded/checked.
- Remove unreliable info (change type) from uploaded payload size data.

##
BTW, if we are concerned about recording lots of irrelevant data, we could still skip uploading/checking for code changes that could not have affected bundle sizes. For example, we could only run the script for commit ranges that touched `yarn.lock` or files inside `packages/`, `scripts/` or the target project directory (`aio/` for example).

### DIFF
--- a/aio/scripts/payload.sh
+++ b/aio/scripts/payload.sh
@@ -12,4 +12,4 @@ source ../scripts/ci/payload-size.sh
 # Provide node_modules from aio
 NODE_MODULES_BIN=$PROJECT_ROOT/aio/node_modules/.bin/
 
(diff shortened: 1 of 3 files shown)


gkalpak:
cc @kara

josephperrott (approved):
LGTM

petebacondarwin (approved):

devversion (approved):

IgorMinar (approved):
thank you!

PR #33992, opened 2019-11-22
build: make ng_module.bzl is_ivy_enabled function public

josephperrott:
We need to make is_ivy_enabled public to allow the internal i18n
build rule to rely on it rather than relying on compile_strategy.
After we move the internal i18n rule to rely on is_ivy_enabled,
compile_strategy can then be removed.


IgorMinar (approved):
please update the commit message and explain why we are making this change.

PR #33997, opened 2019-11-22
fix(common): reflect input type in context

### LINKED ISSUE #31556: Type check directive template context
# 🚀 feature request

### Relevant Package

@angular/compiler

### Description

Template contexts appear not to be checked.

```
    <ng-container *ngIf="data(); let d">
      {{ d.a }}
      {{ d.c }}
    </ng-container>
```

There is zero checking for the type of `d`.

Example: https://github.com/pauldraper/angular-template-context I would like `ng build` to fail.

Template context types are type checked.

Don't use template contexts.

I've also tried using Ivy but it seems broken.

```
ERROR in __ng_typecheck__.ts:5:22 - error TS2304: Cannot find name 'Partial'.

5 const _ctor1: (init: Partial<Pick<i1.NgIf, "ngIf" | "ngIfThen" | "ngIfElse">>) => i1.NgIf = (null!);
```

### LINKED ISSUE #33527: Ivy: fullTemplateTypeCheck errors with NgIterable

crisbeto:
Fixes `NgForOf` not reflecting the type of its input in the `NgForOfContext`. Also fixes `NgIf` which was typed to `any`.

Fixes #33527.
Fixes #31556.

### DIFF
--- a/packages/common/src/directives/ng_if.ts
+++ b/packages/common/src/directives/ng_if.ts
@@ -149,14 +149,14 @@ import {Directive, EmbeddedViewRef, Input, TemplateRef, ViewContainerRef, ɵstri
  * @publicApi
  */
 @Directive({selector: '[ngIf]'})
-export class NgIf {
-  private _context: NgIfContext = new NgIfContext();
-  private _thenTemplateRef: TemplateRef<NgIfContext>|null = null;
-  private _elseTemplateRef: TemplateRef<NgIfContext>|null = null;
-  private _thenViewRef: EmbeddedViewRef<NgIfContext>|null = null;
-  private _elseViewRef: EmbeddedViewRef<NgIfContext>|null = null;
+export class NgIf<T> {
+  private _context: NgIfContext<T> = new NgIfContext<T>();
+  private _thenTemplateRef: TemplateRef<NgIfContext<T>>|null = null;
+  private _elseTemplateRef: TemplateRef<NgIfContext<T>>|null = null;
+  private _thenViewRef: EmbeddedViewRef<NgIfContext<T>>|null = null;
+  private _elseViewRef: EmbeddedViewRef<NgIfContext<T>>|null = null;
 
-  constructor(private _viewContainer: ViewContainerRef, templateRef: TemplateRef<NgIfContext>) {
+  constructor(private _viewContainer: ViewContainerRef, templateRef: TemplateRef<NgIfContext<T>>) {
     this._thenTemplateRef = templateRef;
   }
 
@@ -164,7 +164,7 @@ export class NgIf {
    * The Boolean expression to evaluate as the condition for showing a template.
    */
   @Input()
-  set ngIf(condition: any) {
+  set ngIf(condition: T) {
     this._context.$implicit = this._context.ngIf = condition;
     this._updateView();
   }
@@ -173,7 +173,7 @@ export class NgIf {
    * A template to show if the condition expression evaluates to true.
    */
   @Input()
-  set ngIfThen(templateRef: TemplateRef<NgIfContext>|null) {
+  set ngIfThen(templateRef: TemplateRef<NgIfContext<T>>|null) {
     assertTemplate('ngIfThen', templateRef);
     this._thenTemplateRef = templateRef;
     this._thenViewRef = null;  // clear previous view if any.
@@ -184,7 +184,7 @@ export class NgIf {
    * A template to show if the condition expression evaluates to false.
    */
   @Input()
-  set ngIfElse(templateRef: TemplateRef<NgIfContext>|null) {
+  set ngIfElse(templateRef: TemplateRef<NgIfContext<T>>|null) {
     assertTemplate('ngIfElse', templateRef);
     this._elseTemplateRef = templateRef;
     this._elseViewRef = null;  // clear previous view if any.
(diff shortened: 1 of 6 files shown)


IgorMinar:
[presubmit](https://test.corp.google.com/ui#id=OCL:282499489:BASE:282499508:1574750327811:3bf9d550)
[ivy presubmit](https://test.corp.google.com/ui#id=OCL:282499489:BASE:282499650:1574750420705:1637d262)

IgorMinar (changes requested):
can you please split this into two commits and document each change and what is the user-facing fix/impact because when users read about this change in the changelog they will not have any idea about what this does, or mean to them.

I'm assuming that this is a fix for #33527. Can you please state that in the commit message?

Lastly, any chance we we can add some tests for this? I'm assuming that we currently don't have any good infra to test typechecking changes like this unless we write a ngtsc specific test. Is that right?

I've kicked off a presubmit for this change in the meantime to make sure that there are no surprises.

IgorMinar:
also, this should target the 9.0.x branch as well because otherwise we won't be able to roll this out until v10.

packages/common/src/directives/ng_if.ts
```diff
@@ -149,14 +149,14 @@ import {Directive, EmbeddedViewRef, Input, TemplateRef, ViewContainerRef, ɵstri
  * @publicApi
  */
 @Directive({selector: '[ngIf]'})
-export class NgIf {
-  private _context: NgIfContext = new NgIfContext();
-  private _thenTemplateRef: TemplateRef<NgIfContext>|null = null;
-  private _elseTemplateRef: TemplateRef<NgIfContext>|null = null;
-  private _thenViewRef: EmbeddedViewRef<NgIfContext>|null = null;
-  private _elseViewRef: EmbeddedViewRef<NgIfContext>|null = null;
+export class NgIf<T> {
```
JoostK: There's currently nothing that allows for inferring the type `T` from the condition variable, which would be the `ngIf` input. Then, to allow the inferred type `T` to flow into `NgIfContext<T>` we'd need a `ngTemplateContextGuard` similar to that of `NgForOf`.

packages/common/src/directives/ng_if.ts
```diff
@@ -230,9 +230,9 @@ export class NgIf {
 /**
  * @publicApi
  */
-export class NgIfContext {
-  public $implicit: any = null;
-  public ngIf: any = null;
+export class NgIfContext<T> {
+  public $implicit: T|null = null;
```
JoostK: I am a bit concerned about those `| null`s here, as they will be included in the type that the user needs to deal with. I reckon this is done to allow for the initial assignment to `null`, however I think we should "cheat" and assign `null !` here instead. The `NgIfContext<T>` should not be exposed before its conditions have been set, so we should be okay.

JoostK (changes requested):
Regarding tests, I think the most valuable way would still be ngtsc's template type checking tests. I feel this way because those tests exercise real Angular templates.

As an alternative, a test could mimic the code that the type checker generates, as to assert that type inference works as expected. Such tests however are most appropriate to be run using https://github.com/microsoft/dtslint, which we don't currently have anywhere AFAIK. Also, I would consider them brittle as there's no guarantee that they are accurate, hence my preference for actual ngtsc template type checking tests.

crisbeto:
I've addressed the feedback. Can you take another look @JoostK @IgorMinar?

packages/common/src/directives/ng_if.ts
```diff
@@ -149,14 +149,14 @@ import {Directive, EmbeddedViewRef, Input, TemplateRef, ViewContainerRef, ɵstri
...
-  private _thenTemplateRef: TemplateRef<NgIfContext>|null = null;
-  private _elseTemplateRef: TemplateRef<NgIfContext>|null = null;
-  private _thenViewRef: EmbeddedViewRef<NgIfContext>|null = null;
-  private _elseViewRef: EmbeddedViewRef<NgIfContext>|null = null;
+export class NgIf<T> {
+  private _context: NgIfContext<T> = new NgIfContext<T>();
+  private _thenTemplateRef: TemplateRef<NgIfContext<T>> = null !;
+  private _elseTemplateRef: TemplateRef<NgIfContext<T>> = null !;
+  private _thenViewRef: EmbeddedViewRef<NgIfContext<T>> = null !;
+  private _elseViewRef: EmbeddedViewRef<NgIfContext<T>> = null !;
```
JoostK: All these private fields can keep the `| null`, as it keeps the implementation safe against nulls.

JoostK (dismissed):
I would love to see template type checker tests. If you need help with that, please let me know.

IgorMinar (approved):
fyi: g3 is looking good. @crisbeto can you please address Joost's feedback?

crisbeto:
I have all the feedback addressed locally, but it uncovered some issues with the new signatures. I'll pair up with Joost tomorrow.

crisbeto:
All the issues and feedback should addressed now. Thank you for the help with the tests @JoostK!

JoostK (commented):
In the `NgIf` commit, could you include a fixes note for #31556?

crisbeto:
Done.

alxhub (approved):

Splaktar:
It seems like there should be a test for how this typing behaves with `*ngIf="something | async; let thing". In that case, the behavior here doesn't seem to align with one of the most common use cases of `*ngIf`.

PR #33999, opened 2019-11-22
build: remove compile_strategy from ng_module.bzl

josephperrott:
With the migration of internal i18n rule to use is_ivy_enabled complete, we
can remove compile_strategy as it is now unused.

### DIFF
--- a/packages/bazel/src/ng_module.bzl
+++ b/packages/bazel/src/ng_module.bzl
@@ -26,34 +26,6 @@ load(
 _FLAT_DTS_FILE_SUFFIX = ".bundle.d.ts"
 _R3_SYMBOLS_DTS_FILE = "src/r3_symbols.d.ts"
 
-def compile_strategy(ctx):
-    """Detect which strategy should be used to implement ng_module.
-
-    Depending on the value of the 'compile' define flag, ng_module
-    can be implemented in various ways. This function reads the configuration passed by the user and
-    determines which mode is active.
-
-    Args:
-      ctx: skylark rule execution context
-
-    Returns:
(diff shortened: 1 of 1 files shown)


packages/bazel/src/ng_module.bzl
```diff
@@ -26,34 +26,6 @@ load(
...
-      one of 'legacy' or 'aot' depending on the configuration in ctx
-    """
-
-    strategy = "legacy"
-    if "compile" in ctx.var:
-        strategy = ctx.var["compile"]
-
-    # Enable Angular targets extracted by Kythe Angular indexer to be compiled with the Ivy compiler architecture.
-    # TODO(ayazhafiz): remove once Ivy has landed as the default in g3.
-    if ctx.var.get("GROK_ELLIPSIS_BUILD", None) != None:
```
kyliau: this is still needed for Kythe to build its Angular index with Ivy, is there other location we can move this code to?

josephperrott: The logic for checking for `GROK_ELLIPSIS_BUILD` has been moved to the new [is_ivy_enabled](https://github.com/angular/angular/pull/33999/files/863eaa4927ff3eccb46f4d25749bd385f62ea060#diff-882e1b121b4d9ec070bf136d7853ffbcL76) function just below this one.

josephperrott:
Removing blocked label as all `compile_strategy` usages in google3 have been removed.

packages/bazel/src/ng_module.bzl
kyliau: LGTM, thank you!

kyliau (approved):

PR #34000, opened 2019-11-22
refactor(ivy): remove merge mode in favor of direct-mode styling

matsko:
This patch removes all usage of the merge styling algorithm
(the one that used `TStylingContext`) in favor of a direct
apply algorithm that writes directly to the style attribute
and the className property. By iterating over all bindings
one by one, the final concatenated string value can be applied
once once. This change improves the performance of the styling
features in Angular and also reduces the overall bundle size.


matsko:
presubmit = https://test.corp.google.com/ui#id=OCL:282908383:BASE:282908407:1574936772802:2b228108

ivy presubmit = https://test.corp.google.com/ui#id=OCL:282908383:BASE:282908461:1574936803814:7da45c6b

matsko:
presubmit = http://test/OCL:282960825:BASE:282960833:1574967603926:629be4e8

ivy presubmit = https://test.corp.google.com/ui#id=OCL:282960825:BASE:282960950:1574967705575:7708fc40

packages/core/src/render3/instructions/element.ts
```diff
@@ -87,7 +86,11 @@ export function ɵɵelementStart(
   if (attrs != null) {
     setUpAttributes(renderer, native, attrs);
   }
-  if ((tNode.flags & TNodeFlags.hasInitialStyling) === TNodeFlags.hasInitialStyling) {
+
+  // we don't need to render render the styling on the first create
+  // pass because it was already set in the tNode creation code above
+  if (!tView.firstCreatePass &&
```
pkozlowski-opensource: @matsko I did several refactorings to separate the `firstCreatePass` logic from other code and here we are breaking this code structure. The consequence of this is that you have to:
* read `tView.firstCreatePass` _again_
* read `lView[RENDERER]` _again_ in the `elementStartFirstCreatePass`
* add `Array.isArray(attrs)` check in `elementStartFirstCreatePass`

We need to split the logic differently and:
* do the registration logic _only_ in the `elementStartFirstCreatePass`
* do the actual DOM rendering logic _only_ in the `ɵɵelementStart`

packages/core/src/render3/instructions/element.ts
```diff
@@ -34,8 +32,9 @@ function elementStartFirstCreatePass(
   const attrs = getConstant<TAttributes>(tViewConsts, attrsIndex);
   const tNode = getOrCreateTNode(tView, lView[T_HOST], index, TNodeType.Element, name, attrs);
 
-  if (attrs !== null) {
-    registerInitialStylingOnTNode(tNode, attrs, 0);
+  if (Array.isArray(attrs)) {
+    const renderer = lView[RENDERER];
+    registerAndRenderInitialStyling(renderer, tNode, native, attrs, 0, false);
```
pkozlowski-opensource: See https://github.com/angular/angular/pull/34000/files#r353670644

I think that the `registerAndRenderInitialStyling` should be split into 2 separate functions:
* register
* render

Otherwise we are mixing the first / subsequent creation passes, reading more memory and hurting perf as the results. I think that such split is also good for code read-ability - better to nave 2 focused functions than one bigger that "does it all"

packages/core/src/render3/instructions/element_container.ts
```diff
@@ -33,7 +33,8 @@ function elementContainerStartFirstCreatePass(
   // While ng-container doesn't necessarily support styling, we use the style context to identify
   // and execute directives on the ng-container.
   if (attrs !== null) {
-    registerInitialStylingOnTNode(tNode, attrs, 0);
+    const renderer = lView[RENDERER];
+    registerAndRenderInitialStyling(renderer, tNode, null, attrs, 0, false);
```
pkozlowski-opensource: See https://github.com/angular/angular/pull/34000/files#r353671697

Here it is even more evident that we need to split `registerAndRenderInitialStyling` into 2 functions as for the `elementContainerStartFirstCreatePass` we want to do registration _only_ (there is no rendering happening)

packages/core/test/acceptance/integration_spec.ts
```diff
@@ -1072,7 +1072,7 @@ describe('acceptance integration tests', () => {
 
       it('should delegate initial styles to a [style] input binding if present on a directive on the same element',
          () => {
-           @Component({template: '<div style="width:100px;height:200px" DirWithStyle></div>'})
+           @Component({template: '<div style="width:100px; height:200px" DirWithStyle></div>'})
```
pkozlowski-opensource: This change looks suspicious - why do we need to do it?

packages/core/src/render3/styling/state.ts
```diff
@@ -47,56 +41,49 @@ export interface StylingState {
...
+    sourceStyleTail: 0,
+    sourceClassHead: 0,
+    sourceStyleHead: 0,
+  };
+}
 
-// the `0` start value is reserved for [map]-based entries
-const INDEX_START_VALUE = 1;
+/* tslint:disable */
+const _state = createStylingState();
```
mhevery: This will defeat the tree shaker. (hence `tslint:disable` is needed) Can we just inline the object. (There is no need for `createStylingState`)

packages/core/src/render3/styling/state.ts
```diff
@@ -24,18 +23,13 @@ import {TEMPLATE_DIRECTIVE_INDEX} from '../util/styling_utils';
...
  * that element).
  *
- * To learn more about the algorithm see `TStylingContext`.
+ * To learn more about the algorithm see `instructions/styling.ts`.
  *
  * --------
  */
 
 /**
  * Used as a state reference for update values between style/class binding instructions.
```
mhevery: Is this needed only during first template pass? If so can you add comment to make that clear? (Maybe some `ngDevMod` assertions stating that this is the case?)

packages/core/test/acceptance/styling_spec.ts
```diff
@@ -2437,12 +2354,397 @@ describe('styling', () => {
...
+
+        raiseWidthError = true;
+        expect(() => fixture.detectChanges()).toThrow();
+
+        raiseWidthError = false;
+        expect(() => fixture.detectChanges()).not.toThrow();
+      });
+
+  /*
+onlyInIvy('only ivy treats [class] in concert with other class bindings')
```
mhevery: Why is this commented out?

packages/core/src/render3/instructions/element.ts
```diff
@@ -34,8 +32,9 @@ function elementStartFirstCreatePass(
   const attrs = getConstant<TAttributes>(tViewConsts, attrsIndex);
   const tNode = getOrCreateTNode(tView, lView[T_HOST], index, TNodeType.Element, name, attrs);
 
-  if (attrs !== null) {
-    registerInitialStylingOnTNode(tNode, attrs, 0);
+  if (Array.isArray(attrs)) {
```
mhevery: Would `attrs` be something other than an `Array`?

mhevery (changes requested):

packages/core/src/render3/util/styling_utils.ts
```diff
@@ -302,156 +134,573 @@ export function isHostStylingActive(directiveOrSourceId: number): boolean {
...
-      addItemToStylingMap(stylingMapArr, newProp, value, true);
+  /* tslint:disable */
+  console.log(str);
+}
+
+function getPrintedBindingName(name: string, isHostBinding: boolean) {
+  return isHostBinding ? `@HostBinding("${name}")` : `[${name}]`;
+}
+
+export function printStylingTable(
```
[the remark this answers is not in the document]
matsko: We can't move these into styling_debug.ts because then I can't use them while I'm debugging locally because styling_debug.ts imports from the algorithm files (it would be a cyclic dependency).

Ideally it would go into `window.ng.debugStyling()`. But we don't have this system fully in place yet.

packages/core/src/render3/util/styling_utils.ts
```diff
@@ -302,156 +134,573 @@ export function isHostStylingActive(directiveOrSourceId: number): boolean {
...
-  stylingMapArr.push(prop, value);
-  return true;
+  if (stylesStr.length !== 0) {
+    // we add the styles in reverse so the template-level styles override
+    // the styles any directive styling (the template-level styles are
+    // always added first)
+    tNode.styles = concatString(stylesStr, tNode.styles, STYLE_ENTRIES_SEPARATOR);
+    tNode.initialStyleNames = initialStyleNames;
+    if (native && !appendStyles) {
+      setStyleAttr(renderer, native, tNode.styles);
```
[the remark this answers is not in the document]
matsko: No. Only during the `firstCreatePass` does it do it once per element and each time a directive fires that has styling. After the firstCreatePass (for a tNode) it just does it once (because it knows all the styles/classes on the element).

packages/core/src/render3/instructions/element.ts
matsko: Yeah this should be fine to move into the `elementStartFirstCreatePass` function.

packages/core/src/render3/instructions/element.ts
matsko: We can't do this because we can only write to the `className` or `attribute.style` value once (which happens just before `registerAndRenderInitialStyling` is called) and then append to it (using `classList` and/or `style`). The key thing is to remember what classes/styles are extra. Otherwise you would have to store all the additional classes/styles in an array and pass that back so it can be rendered directly after. It wouldn't be much of an additional change.

packages/core/src/render3/instructions/element_container.ts
matsko: See my previous comment.

packages/core/test/acceptance/integration_spec.ts
matsko: This shouldn't matter anymore. Reverting.

aio/scripts/_payload-limits.json
```diff
@@ -12,7 +12,7 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 2987,
-        "main-es2015": 464734,
+        "main-es2015": 509333,
```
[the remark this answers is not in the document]
matsko: Yes it looks like ES5 is bigger but other sizes are smaller.

packages/core/src/render3/styling/state.ts
matsko: No this is used all over the algorithm during each update pass.

packages/core/test/acceptance/styling_spec.ts
matsko: This was the test that me and Andrew came across. Still deciding if it's the right way to go.

packages/core/src/render3/instructions/element.ts
matsko: The benchmark tests render it as `undefined` sometimes (and it is `null` by default). Instead of having to have two checks (`undefined` and `null`), better to just have one to see if it's an array?

packages/core/src/render3/styling/state.ts
matsko: Removing.

packages/core/src/render3/util/styling_utils.ts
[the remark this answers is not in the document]
matsko: This isn't an issue for me in non-Google code, but I do use this function while testing G3 to see what's going on.

packages/core/src/render3/util/styling_utils.ts
```diff
@@ -302,156 +134,573 @@ export function isHostStylingActive(directiveOrSourceId: number): boolean {
...
-      if (propAtIndex === prop) {
-        const valueAtIndex = stylingMapArr[j];
-        if (allowOverwrite || !isStylingValueDefined(valueAtIndex)) {
-          applied = true;
-          setMapValue(stylingMapArr, j, value);
+export function setConcatenatedValue(lView: LStylingData, bindingIndex: number, value: string) {
+  lView[bindingIndex + 1] = value;
+}
+
+export function registerAndRenderInitialStyling(
```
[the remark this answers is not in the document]
matsko: As I mentioned to you in Slack. `styling_utils.ts` is still undocumented.

packages/core/src/render3/util/styling_utils.ts
```diff
@@ -161,32 +44,18 @@ export function getValue<T = any>(data: LStylingData, bindingIndex: number): T|n
...
-    a: NO_CHANGE | StylingMapArray | number | string | null | boolean | undefined | {},
-    b: NO_CHANGE | StylingMapArray | number | string | null | boolean | undefined | {}): boolean {
+    a: NO_CHANGE | number | string | null | boolean | undefined | {},
+    b: NO_CHANGE | number | string | null | boolean | undefined | {}): boolean {
   if (b === NO_CHANGE) return false;
-
-  const compareValueA = Array.isArray(a) ? a[StylingMapArrayIndex.RawValuePosition] : a;
-  const compareValueB = Array.isArray(b) ? b[StylingMapArrayIndex.RawValuePosition] : b;
-  return !Object.is(compareValueA, compareValueB);
+  return !Object.is(a, b);
```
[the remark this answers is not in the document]
matsko: Removing this causes various tests to fail. Looking into it.

packages/core/src/render3/instructions/shared.ts
```diff
@@ -32,11 +32,10 @@ import {CHILD_HEAD, CHILD_TAIL, CLEANUP, CONTEXT, DECLARATION_COMPONENT_VIEW, DE
 import {assertNodeOfPossibleTypes} from '../node_assert';
 import {isNodeMatchingSelectorList} from '../node_selector_matcher';
 import {ActiveElementFlags, enterView, executeElementExitFn, getBindingsEnabled, getCheckNoChangesMode, getIsParent, getPreviousOrParentTNode, getSelectedIndex, hasActiveElementFlag, incrementActiveDirectiveId, leaveView, leaveViewProcessExit, setActiveHostElement, setBindingIndex, setBindingRoot, setCheckNoChangesMode, setCurrentDirectiveDef, setCurrentQueryIndex, setPreviousOrParentTNode, setSelectedIndex} from '../state';
-import {renderStylingMap, writeStylingValueDirectly} from '../styling/bindings';
 import {NO_CHANGE} from '../tokens';
 import {isAnimationProp} from '../util/attrs_utils';
 import {INTERPOLATION_DELIMITER, renderStringify, stringifyForError} from '../util/misc_utils';
-import {getInitialStylingValue} from '../util/styling_utils';
+import {setClassName, setStyleAttr} from '../util/styling_utils';
```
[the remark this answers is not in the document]
matsko: Fixed.

packages/core/src/render3/instructions/element.ts
pkozlowski-opensource: In this case we should rather change the benchmark to avoid Array.isArray(attrs) check here (strong preference) or just do `attrs != null`

packages/core/src/render3/instructions/element.ts
matsko: Actually no. It won't work because the `firstCreatePass` code only runs on the first time the `tNode` is processed. We still need to render styling for all the other times the element is created.

matsko: See my other comment about the registration.

packages/core/src/render3/instructions/element.ts
matsko: So the reason why this is a problem at all is because various Material components/directives add/remove classes/styles in their constructor code (using their own classList/style[prop] code). Components/directives also expect that various classes to be defined before they are run (which is why the rendering happens first via `renderInitialStyling`).

Then each time a directive is added it will add new classes/styles in addition to what was there earlier. Those are added via classList/style[prop]. Because Material components may add their own classes we can't rewrite the attribute again (which is what happens when `renderInitialStyling` is called) ... This is why it's skipped when on the first create pass.

This explains why:
- On firstCreatePass we need to loop over all the attributes and render multiple times
- Otherwise we can write the to the attribute directly because we know the combined className/style string that will be written to the element (combined === the template + directive initial classes/styles combined into two strings).

I understand the logic is coupled, but if you want to change how `registerAndRenderInitialStyling` work then you will need to collect the new classes/styles in an array. Then render that on the element immediately after? Doesn't that mean that more memory is used?

packages/core/src/render3/styling/direct_write_algorithm.ts
```diff
@@ -0,0 +1,383 @@
...
+ *          exits for the element currently being processed.
+ */
+export function updateBindingValue(
+    lView: LView, tData: TData, state: StylingState, value: any, bindingIndex: number,
+    isClassBased: boolean): boolean {
+  const previousValue = getValue(lView, bindingIndex);
+  const updated = hasValueChanged(previousValue, value);
+  if (updated) {
+    setValue(lView, bindingIndex, value);
+    if (isBindingRegistered(tData, bindingIndex) &&
```
mhevery: could we change this `if` statement to something like this:
```
markLocationFromWhichStylesNeedToBeRecomputed(...)

/**
 * This function marks the place in the styling linked list from which the flushing needs to start.
 *
 * There are two things to keep in mind. 
 * 1. Bindings for template are before hostBindings for component which are before hostBindings for directives.
 *    We refer to this as source. `0` for Template, `1` for component, `2 for directive and so on.
 *    The implication of this is that a higher source ID is linked earlier in the styling linked list. 
 * 2. Within the source id the instructions are generated in the correct order. So an instruction which has should 
 *    overwrite a previous instruction will be after such instruction. Implication of this is that within the same
 *    source ID the lowest binding index wins.
 *
 * Putting the above two together yields these resolution rules.
 * 1. Always take the index of higher sourceID.
 * 2. If SourceID same the first binding index was the correct one.
 */
function markLocationFromWhichStylesNeedToBeRecomputed(...) {
  if (lastSourceId !== currentSourceID) {
    ngDevMode && assertLessThen(state.flusStartSourceId, currentSourceId, "expecting source ids to be increasing only");
    state.flushStartIndex = currentBindingIndex;
    state.flushStartSourceId = currentSourceId;
  }
}
```

I think this is a lot simpler to follow as to what is going on.

packages/core/src/render3/styling/state.ts
```diff
@@ -114,19 +101,19 @@ const INDEX_START_VALUE = 1;
...
-    _state.classesBitMask = BIT_MASK_START_VALUE;
-    _state.classesIndex = INDEX_START_VALUE;
-    _state.stylesBitMask = BIT_MASK_START_VALUE;
-    _state.stylesIndex = INDEX_START_VALUE;
-    _state.lastDirectClassMap = null;
-    _state.lastDirectStyleMap = null;
-  } else if (_state.directiveIndex !== directiveIndex) {
-    _state.directiveIndex = directiveIndex;
-    _state.sourceIndex++;
+  if (_state.element !== element || _state.directiveIndex !== directiveIndex) {
```
mhevery: Why do we need both directiveIndex as well as element? It seems redundant. 
- My understanding is that `element` should not be needed
- `directiveIndex` and `sourceIndex` track the same information, and should  be unified.

mhevery (changes requested):

packages/core/src/render3/styling/state.ts
matsko: we can keep only `directiveIndex` because `sourceIndex` was only used with `TStylingContext`.

matsko: `element` is used to track when the template bindings exit and/or when the host bindings are done.

matsko:
http://test/OCL:284046515:BASE:284046617:1575583106000:7e519709

packages/core/src/render3/styling/direct_write_algorithm.ts
```diff
@@ -0,0 +1,383 @@
...
+/**
+ * Whether or not any style/class bindings are queued for flushing
+ */
+export function hasBindingsToFlush(state: StylingState) {
+  return state.lastStyleBindingIndex !== 0 || state.lastClassBindingIndex !== 0;
+}
+
+const LAST_BINDING_INDEX_BITS = 16;
+const LAST_BINDING_INDEX_MASK = (1 << LAST_BINDING_INDEX_BITS) - 1;
+function getMinSourceAndBindingIndex(
```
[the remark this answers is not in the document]
matsko: I changed them all to be:
- `setDirectiveAndBindingStartIndices`
- `getBindingStartIndex`
- `getDirectiveStartIndex`

packages/core/src/render3/instructions/styling.ts
```diff
@@ -304,173 +264,57 @@ export function ɵɵstyleMap(styles: {[styleName: string]: any} | NO_CHANGE | nu
...
-    if (isStylingValueDefined(newValue) || !firstUpdatePass) {
-      const inputName: string = isClassBased ? selectClassBasedInputName(tNode.inputs !) : 'style';
-      const inputs = tNode.inputs ![inputName] !;
-      const initialValue = getInitialStylingValue(context);
-      const value = normalizeStylingDirectiveInputValue(initialValue, newValue, isClassBased);
-      setInputsForProperty(lView, inputs, inputName, value);
-      setElementExitFn(stylingApply);
-    }
-    setValue(lView, bindingIndex, newValue);
+    updated = updateBindingValue(lView, tData, state, value, bindingIndex, isClassBased);
```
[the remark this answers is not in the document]
matsko: the `updated` flag is used below to update the `ngDevMode` counters.

packages/core/src/render3/instructions/element.ts
matsko: Do you think `attrs != null` is more clear than `Array.isArray`?

packages/core/src/render3/styling/binding_concatenation.ts
```diff
@@ -0,0 +1,398 @@
...
+ *
+ * All style/class values are stored in the provided `LView` instance and each
+ * of these entries are continuously updated whenever `updateBindingValue` is called.
+ * Due to the nature of the concatenation algorithm, all of the binding values are
+ * concatenated together into a single style/className string value which is then
+ * applied directly to the element in this function.
+ */
+export function flushBindings(
+    renderer: any, native: RElement, lView: LView, tNode: TNode, tData: TData, state: StylingState,
+    sanitizer: StyleSanitizeFn | null, firstUpdatePass: boolean, hostBindingsMode: boolean): void {
```
mhevery: `firstUpdatePass` is a suspect here. The logic here should be independent of wether or not we are in `firstUpdatePass`.

Same thing for `firstUpdatePass` 

This algorithm should be agnostic to `firstUpdatePass` and `hostBindingsMode`

packages/core/src/render3/styling/binding_concatenation.ts
```diff
@@ -0,0 +1,398 @@
...
+  const previousBindingIndex = getPreviousBindingIndex(tData, bindingIndex);
+  return getConcatenatedStr(
+      lView, previousBindingIndex, isClassBased ? tNode.classes : tNode.styles);
+}
+
+/**
+ * Applies (or reapplies) each style/class binding entry into the provided `LView` (all the way
+ * until the tail).
+ */
+export function processStylingBindingsUpToPoint(
```
mhevery: maybe rename to `processStylingBindingsUpToPoint` => `processStylingBindingsUntilEnd`? 

It says upto point, but there is no point, it goes to end.

packages/core/src/render3/styling/binding_concatenation.ts
```diff
@@ -0,0 +1,398 @@
...
+    lView: LStylingData, tData: TData, tNode: TStylingNode, bindingIndex: number,
+    sanitizer: StyleSanitizeFn | null, isClassBased: boolean): void {
```
mhevery: We talked about having the sanitizer be stored in the `TData` Would that be too much work to refactor? It would really clean up this code.

mhevery (changes requested):

packages/core/src/render3/styling/binding_concatenation.ts
matsko: If you're ok to do it with this PR then I'm happy to do it.

packages/core/src/render3/styling/binding_concatenation.ts
matsko: Fixed.

packages/core/src/render3/styling/binding_concatenation.ts
matsko: It's used to determine whether or not the initial concatenated value is the initial value (used during element creation) or the most recent concatenated value.

matsko: hostBindingsMode is used to resolve whether or not the template bindings were applied in the last CD. If yes then it reads that, otherwise it gets the initial value.

packages/core/src/render3/styling/binding_concatenation.ts
```diff
@@ -0,0 +1,398 @@
...
+ * Due to the nature of the concatenation algorithm, all of the binding values are
+ * concatenated together into a single style/className string value which is then
+ * applied directly to the element in this function.
+ */
+export function flushBindings(
+    renderer: any, native: RElement, lView: LView, tNode: TNode, tData: TData, state: StylingState,
+    sanitizer: StyleSanitizeFn | null, firstUpdatePass: boolean, hostBindingsMode: boolean): void {
+  const lastStyleBindingIndex = getBindingStartIndex(state, false);
+  if (lastStyleBindingIndex !== 0) {
+    const stylesTail = getStylingTail(tNode, false);
```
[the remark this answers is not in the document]
matsko: The tail is used to determine where to read the last applied value (the concatenated value).

packages/core/src/render3/styling/binding_concatenation.ts
```diff
@@ -0,0 +1,398 @@
...
+      // the write operation may apply the value directly to the element's style
+      // attribute. Some browsers may reorder the style values on said property, therefore,
+      // the previous value will need to be updated with this entry.
+      if (result !== null) {
+        setConcatenatedValue(lView, stylesTail, result);
+      }
+    }
+  }
+
+  const lastClassBindingIndex = getBindingStartIndex(state, true);
```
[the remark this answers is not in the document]
matsko: De-duped.

packages/core/src/render3/styling/binding_concatenation.ts
```diff
@@ -0,0 +1,398 @@
...
+ * `updateBindingValue` function.
+ *
+ * All style/class values are stored in the provided `LView` instance and each
+ * of these entries are continuously updated whenever `updateBindingValue` is called.
+ * Due to the nature of the concatenation algorithm, all of the binding values are
+ * concatenated together into a single style/className string value which is then
+ * applied directly to the element in this function.
+ */
+export function flushBindings(
+    renderer: any, native: RElement, lView: LView, tNode: TNode, tData: TData, state: StylingState,
```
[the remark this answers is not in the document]
matsko: We would need to call the function twice then.

packages/core/src/render3/interfaces/view.ts
```diff
@@ -752,7 +752,7 @@ export type HookData = (number | (() => void))[];
  */
 export type TData =
     (TNode | PipeDef<any>| DirectiveDef<any>| ComponentDef<any>| number | Type<any>|
-     InjectionToken<any>| TI18n | I18nUpdateOpCodes | null | string)[];
+     InjectionToken<any>| TI18n | I18nUpdateOpCodes | null | string | {})[];
```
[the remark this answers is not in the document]
matsko: Added.

packages/core/src/render3/instructions/styling.ts
```diff
@@ -440,181 +416,10 @@ function updateDirectiveInputValue(
     if (isStylingValueDefined(newValue) || !firstUpdatePass) {
```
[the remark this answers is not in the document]
matsko: `elementStart` will write the falsy value upon creation. I will add a comment that explains.

integration/_payload-limits.json
```diff
@@ -39,7 +39,7 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 2289,
-        "main-es2015": 266648,
+        "main-es2015": 267395,
```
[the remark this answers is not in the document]
matsko: We should discuss part 1) more because I have other ideas. I don't know why the styling code is being pulled in when only static styles are used in an app. We should investigate.

mhevery (changes requested):

packages/core/src/render3/instructions/element.ts
pkozlowski-opensource: Yup! (and probably cheeper). But regardless, we should have it as `attrs !== null` (that is, fix the benchmark - we don't want to pay readabiluty / perf price of benchmarks in the prod code).

kara: Yeah, I'd agree. The `isArray` check implies that there is a value other than an array. `attrs !== null` is clearer.

packages/core/src/render3/instructions/element.ts
```diff
@@ -142,12 +145,12 @@ export function ɵɵelementEnd(): void {
     }
   }
 
-  if (hasClassInput(tNode)) {
+  if (hasClassInput(tNode) && typeof tNode.classes === 'string') {
```
pkozlowski-opensource: Why do we need to check `typeof tNode.classes === 'string'`? Shouldn't it be `tNode.classes !== null` (similar question for styles)?

packages/core/src/render3/instructions/styling.ts
```diff
@@ -304,173 +259,59 @@ export function ɵɵstyleMap(styles: {[styleName: string]: any} | NO_CHANGE | nu
  * @codeGenApi
  */
 export function ɵɵclassMap(classes: {[className: string]: any} | NO_CHANGE | string | null): void {
-  classMapInternal(getSelectedIndex(), classes);
+  stylingBindingInternal(getSelectedIndex(), null, classes, null, null, true);
 }
 
 /**
- * Internal function for applying a class string or key/value map of classes to an element.
+ * Applies the provided `[style]`, `[style.prop]`, `[class]` or `[class.name]` binding.
```
kara: This function doesn't actually apply the styles, does it? My understanding is that it just registers the bindings and we flush the styles later, but this comment makes it sound like it does both

packages/core/src/render3/styling/binding_concatenation.ts
```diff
@@ -0,0 +1,432 @@
...
+    }
+  }
+
+  setConcatenatedValue(lView, bindingIndex, concatenatedStr);
+}
+
+/**
+ * Sanitizes and normalizes the value that will be placed in the concatenation string.
+ *
+ * @returns a string value (when the value is allowed to be concatenated), otherwise a falsy
```
IgorMinar:
```suggestion
 * @returns a string value (when the value is allowed to be concatenated), otherwise a false
```

IgorMinar: unless you really mean `falsy` in which case you should be explicit and say `false|null`

packages/core/src/render3/util/styling_utils.ts
```diff
@@ -6,191 +6,94 @@
...
+ */
+export function setValue(data: LStylingData, bindingIndex: number, value: any): void {
   data[bindingIndex] = value;
 }
 
+/**
+ * Returns the value set at the provided `bindingIndex` in the `LStylingData` array
+ */
 export function getValue<T = any>(data: LStylingData, bindingIndex: number): T|null {
   return bindingIndex !== 0 ? data[bindingIndex] as T : null;
```
kara: When is binding index 0?

packages/core/src/render3/styling/binding_concatenation.ts
```diff
@@ -0,0 +1,432 @@
...
+ *          value (which signals that the value should not be concatenated).
+ */
```
IgorMinar: I have a few questions related to this fn:

1. can you please document all parameters?
2. under what circumstances can the `sanitizer` be `null`? is that a valid scenario? I see us check if sanitizer is not `null` in many places, so maybe we should ensure that it is null early on and then expect that it will not be null.
3. related to that, is `null` really a valid return value? what does it mean to return `null`? Can you please document the meaning, or even better if possible always ensure that we return a string by returning `''` instead?
4. can you consider renaming `validateAndSanitize` to `isMapBased` as it reflects the type of binding rather than what we should do with it (similar to `isClassBased`)
5. can you please document the meaning of returned `false` value?

packages/core/src/render3/instructions/styling.ts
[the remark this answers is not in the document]
kara: `updateBindingValue` looks pretty similar to `bindingUpdated` (which also returns a boolean). Seems like we might be able to re-use it instead of rewriting it from scratch.

packages/core/src/render3/styling/binding_registration.ts
```diff
@@ -0,0 +1,344 @@
...
+  if (directiveIndex === 0) return BindingSourceMode.Template;
+  const hasComponentOnHost = tNode.flags & TNodeFlags.isComponentHost;
+  return directiveIndex === 1 && hasComponentOnHost ? BindingSourceMode.Component :
+                                                      BindingSourceMode.Directive;
+}
+
+/**
+ * Simple enum used to express what mode the bindings are being applied in
+ */
+const enum BindingSourceMode {
```
kara: Can we move this to the styling interface file?

packages/core/src/render3/styling/binding_concatenation.ts
```diff
@@ -0,0 +1,432 @@
...
+function normalizeValueForConcatenation(
+    prop: string, value: string | number | boolean | null, suffix: string,
+    sanitizer: StyleSanitizeFn | null, validateAndSanitize: boolean, isClassBased: boolean): string|
+    false|null {
+  if (isStylingValueDefined(value) && value !== false) {
+    if (isClassBased) {
```
IgorMinar: Is it justified to complicate this fn using this flag? why not create two simpler functions: `normalizeStyleValueForConcatenation` and `normalizeClassValueForConcatenation`, with the second being just:

```
function normalizeClassValueForConcatenation(prop, value, sanitizer): string|false {
  return (isStylingValueDefined(value) && value !== false && value !== 0) ? false : prop;
}
```
The code assumes that I understand the meaning of `false` and `0` values, which I don't so I might have gotten that part wrong, but the general simplification suggestion should still apply. In short, I don't believe that the isClassBased flag is necessary, and it appears to make the code more convoluted.

kara (changes requested):
First pass - since this refactor is so large, it's mostly questions. Will do another pass once @matsko and I can sync up

packages/core/src/render3/styling/binding_concatenation.ts
```diff
@@ -0,0 +1,432 @@
...
+ * Sanitizes and normalizes the value that will be placed in the concatenation string.
+ *
+ * @returns a string value (when the value is allowed to be concatenated), otherwise a falsy
+ *          value (which signals that the value should not be concatenated).
+ */
+function normalizeValueForConcatenation(
+    prop: string, value: string | number | boolean | null, suffix: string,
+    sanitizer: StyleSanitizeFn | null, validateAndSanitize: boolean, isClassBased: boolean): string|
+    false|null {
+  if (isStylingValueDefined(value) && value !== false) {
```
IgorMinar: based on this logic, and because of the cast on the last line of the fn, if the value is `true` then this fn returns `true` in spite of the fn signature claiming that that can't be. that seems like a bug?

IgorMinar: I also find it confusing that `isStylingValueDefined` explicitly considers the value defined if it's set to `false` but not defined if it's set to `''` - because of this it's not clear what `isStylingValueDefined` actually means and why the comment in `isStylingValueDefined` state that the value is defined if it's set to `false`.

I'm sorry that I can't suggest what this should look like instead because I don't fully understand the significance of `''` and `false` values here, I'm just pointing out that the point of this condition is not clear and considering the `true` bug described above, we already seem to have an issue here, so it would be great to clarify this.

packages/core/src/render3/styling/binding_concatenation.ts
```diff
@@ -0,0 +1,432 @@
...
+    } else if (sanitizer !== null) {
+      const sanitizationFlag = validateAndSanitize ?
+          StyleSanitizeMode
+              .ValidateAndSanitize  // map-based values can't be validated ahead of time
+          :
+          StyleSanitizeMode.SanitizeOnly;  // prop-based values are checked during firstUpdatePass
+                                           // if they need sanitization
+      value = sanitizer(prop, value, sanitizationFlag);
+    } else {
+      value = renderStringify(value);
```
IgorMinar: shouldn't we `unwrapSafeValue` before stringifying?

IgorMinar: see related PR with a pending fix: #34286

IgorMinar (changes requested):
I've reviewed `normalizeValueForConcatenation` while reviewing #34286 and I have some suggestions/questions that I'm posting as individual review comments.

packages/core/src/render3/instructions/shared.ts
```diff
@@ -801,8 +799,11 @@ export function createTNode(
                          null,       // child: ITNode|null
                          tParent,    // parent: TElementNode|TContainerNode|null
                          null,       // projection: number|(ITNode|RNode[])[]|null
-                         null,       // styles: TStylingContext|null
-                         null,       // classes: TStylingContext|null
+                         null,       // initialStyleNames: string[]|null
+                         null,       // styles: string
```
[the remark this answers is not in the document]
matsko: @mhevery if you're going to push directly to the PR with these changes then please update the docs as well.

packages/core/src/render3/instructions/class_map_interpolation.ts
```diff
@@ -334,5 +334,9 @@ export function ɵɵclassMapInterpolate8(
 export function ɵɵclassMapInterpolateV(values: any[]): void {
   const lView = getLView();
   const interpolatedValue = interpolationV(lView, values);
-  classMapInternal(getSelectedIndex(), interpolatedValue);
+  stylingMapLocal(interpolatedValue);
+}
+
+function stylingMapLocal(value: {}) {
+  stylingBindingInternal(getSelectedIndex(), null, value, null, null, true);
```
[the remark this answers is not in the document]
matsko: We could, but there are a lot more params to pass in though. Bundle size-wise using a wrapper function would be smaller.

packages/core/src/render3/instructions/styling.ts
matsko:
```
 * Queues the provided `[style]`, `[style.prop]`, `[class]` or `[class.name]` binding to be applied during a styling flush.
 *
 * Each time any styling instruction runs the style/class values are passed
 * in through this function, but they are not yet applied to the element.
 * They are instead "queued" until the "styling flush" occurs (which happens
 * when the element is fully processed).
 *
 * When this function is called during the first update pass (when the bindings are
 * run for the very first time on an element), the binding property name and metadata
 * are registered on the associated `TData` data-structure (which is obtained from
 * the current global `LView` instance).
```

packages/core/src/render3/util/styling_utils.ts
matsko: This is used in the current implementation (in `TStylingContext`) to specify that a binding doesn't have a value in a specific directive.

matsko: fixed

packages/core/src/render3/styling/binding_registration.ts
matsko: moved

matsko:
https://test.corp.google.com/ui#id=OCL:284465544:BASE:285283537:1576192983681:e3657fdb

matsko:
https://tap.corp.google.com/ui#id=OCL:285468034:BASE:285468117:1576274879535:4f7ae053

PR #34004, opened 2019-11-23
refactor(ivy): Add style reconciliation algorithm

mhevery:
# refactor(ivy): Add style reconciliation algorithm

This change introduces class/style reconciliation algorithm for DOM elements.
NOTE: The code is not yet hooked up, it will be used by future style algorithm.

Background: 
Styling algorithm currently has [two paths](https://hackmd.io/@5zDGNGArSxiHhgvxRGrg-g/rycZk3N5S) when computing how the style should be rendered. 
1. A direct path which concatenates styling and uses `elemnent.className`/`element.style.cssText` and
2. A merge path which uses internal data structures and uses `element.classList.add/remove`/`element.style[property]`.

The situation is confusing and hard to follow/maintain. So a future PR will remove the merge-path and do everything with direct-path. This however breaks when some other code adds class or style to the element without Angular's knowledge. If this happens instead of switching from direct-path to merge-path algorithm, this change provides a different mental model whereby we always do `direct-path` but the code which writes to the DOM detects the situation and reconciles the out of bound write.

The reconciliation process is as follows:
1. Detect that no one has modified `className`/`cssText` and if so just write directly (fast path).
2. If out of bounds write did occur, switch from writing using `className`/`cssText` to `element.classList.add/remove`/`element.style[property]`. This does require that the write function computes the difference between the previous Angular expected state and current Angular state. (This requires a parser. The advantage of having a parser is that we can support `style="width: {{exp}}px"` kind of bindings.) Compute the diff and apply it in non destructive way using `element.classList.add/remove`/`element.style[property]`

Properties of approach:
- If no out of bounds style modification:
  - Very fast code path: Just concatenate string in right order and write them to DOM.
  - Class list order is preserved
- If out of bounds style modification detected:
  - Penalty for parsing 
  - Switch to non destructive modification: `element.classList.add/remove`/`element.style[property]`
  - Switch to alphabetical way of setting classes.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/render3/styling/class_differ.ts
+++ b/packages/core/src/render3/styling/class_differ.ts
@@ -1,0 +1,131 @@
+/**
+* @license
+* Copyright Google Inc. All Rights Reserved.
+*
+* Use of this source code is governed by an MIT-style license that can be
+* found in the LICENSE file at https://angular.io/license
+*/
+
+import {CharCode} from '../../util/char_code';
+import {consumeClassToken, consumeWhitespace} from './styling_parser';
+
+/**
+ * Computes the diff between two class-list strings.
+ *
+ * Example:
+ *  `oldValue` => `"A B C"`
+ *  `newValue` => `"A C D"`
+ * will result in:
+ * ```
+ * new Map([
+ *   ['A', null],
+ *   ['B', false],
+ *   ['C', null],
+ *   ['D', true]
+ * ])
+ * ```
+ *
+ * @param oldValue Previous class-list string.
+ * @param newValue New class-list string.
+ * @returns A `Map` which will be filled with changes.
+ *        - `true`: Class needs to be added to the element.
+ *        - `false: Class needs to be removed from the element.
+ *        - `null`: No change (leave class as is.)
+ */
+export function computeClassChanges(oldValue: string, newValue: string): Map<string, boolean|null> {
+  const changes = new Map<string, boolean|null>();
+  splitClassList(oldValue, changes, false);
+  splitClassList(newValue, changes, true);
+  return changes;
+}
+
+/**
+ * Splits the class list into array, ignoring whitespace and add it to corresponding categories
+ * `changes`.
+ *
+ * @param text Class list to split
+ * @param changes Map which will be filled with changes. (`false` - remove; `null` - noop;
+ *        `true` - add.)
+ * @param isNewValue `true` if we are processing new list.
+ */
+export function splitClassList(
+    text: string, changes: Map<string, boolean|null>, isNewValue: boolean): void {
+  const end = text.length;
+  let index = 0;
+  while (index < end) {
+    index = consumeWhitespace(text, index, end);
+    const tokenEnd = consumeClassToken(text, index, end);
+    if (tokenEnd !== index) {
+      processClassToken(changes, text.substring(index, tokenEnd), isNewValue);
+    }
+    index = tokenEnd;
+  }
+}
+
+/**
+ * Processes the token by adding it to the `changes` Map.
+ *
+ * @param changes Map which keeps track of what should be done with each value.
+ *        - `false` The token should be deleted. (It was in old list, but not in new list.)
+ *        - `null` The token should be ignored. (It was present in old list as well as new list.)
+ *        - `true` the token should be added. (It was only present in the new value)
+ * @param token Token to add to set.
+ * @param isNewValue True if invocation represents an addition (removal otherwise.)
+ *        - `false` means that we are processing the old value, which may need to be deleted.
+ *          Initially all tokens are labeled `false` (remove it.)
+ *        - `true` means that we are processing new value which may need to be added. If a token
+ *          with same key already exists with `false` then the resulting token is `null` (no
+ *          change.) If no token exists then the new token value is `true` (add it.)
+ */
+export function processClassToken(
+    changes: Map<string, boolean|null>, token: string, isNewValue: boolean) {
+  if (isNewValue) {
+    // This code path is executed when we are iterating over new values.
+    const existingTokenValue = changes.get(token);
+    if (existingTokenValue === undefined) {
+      // the new list has a token which is not present in the old list. Mark it for addition.
+      changes.set(token, true);
+    } else if (existingTokenValue === false) {
+      // If the existing value is `false` this means it was in the old list. Because it is in the
+      // new list as well we marked it as `null` (noop.)
+      changes.set(token, null);
+    }
+  } else {
+    // This code path is executed when we are iterating over previous values.
+    // This means that we store the tokens in `changes` with `false` (removals).
+    changes.set(token, false);
+  }
+}
+
+/**
+ * Removes a class from a `className` string.
+ *
+ * @param className A string containing classes (whitespace separated)
+ * @param classToRemove A class name to remove from the `className`
+ * @returns a new class-list which does not have `classToRemove`
+ */
+export function removeClass(className: string, classToRemove: string): string {
+  let start = 0;
+  let end = className.length;
+  while (start < end) {
+    start = className.indexOf(classToRemove, start);
+    if (start === -1) {
+      // we did not find anything, so just bail.
+      break;
+    }
+    const removeLength = classToRemove.length;
+    const hasLeadingWhiteSpace = start === 0 || className.charCodeAt(start - 1) <= CharCode.SPACE;
+    const hasTrailingWhiteSpace = start + removeLength === end ||
+        className.charCodeAt(start + removeLength) <= CharCode.SPACE;
+    if (hasLeadingWhiteSpace && hasTrailingWhiteSpace) {
+      // Cut out the class which should be removed.
+      const endWhitespace = consumeWhitespace(className, start + removeLength, end);
+      className = className.substring(0, start) + className.substring(endWhitespace, end);
+      end = className.length;
+    } else {
+      // in this case we are only a substring of the actual class, move on.
+      start = start + removeLength;
+    }
+  }
+  return className;
+}
--- a/packages/core/src/render3/styling/reconcile.ts
+++ b/packages/core/src/render3/styling/reconcile.ts
@@ -1,0 +1,175 @@
+/**
+* @license
+* Copyright Google Inc. All Rights Reserved.
+*
+* Use of this source code is governed by an MIT-style license that can be
+* found in the LICENSE file at https://angular.io/license
+*/
+
+import {ProceduralRenderer3, RElement, Renderer3, isProceduralRenderer} from '../interfaces/renderer';
+import {computeClassChanges} from './class_differ';
+import {computeStyleChanges} from './style_differ';
+
+/**
+ * Writes new `className` value in the DOM node.
+ *
+ * In its simplest form this function just writes the `newValue` into the `element.className`
+ * property.
+ *
+ * However, under some circumstances this is more complex because there could be other code which
+ * has added `class` information to the DOM element. In such a case writing our new value would
+ * clobber what is already on the element and would result in incorrect behavior.
+ *
+ * To solve the above the function first reads the `element.className` to see if it matches the
+ * `expectedValue`. (In our case `expectedValue` is just last value written into the DOM.) In this
+ * way we can detect to see if anyone has modified the DOM since our last write.
+ * - If we detect no change we simply write: `element.className = newValue`.
+ * - If we do detect change then we compute the difference between the `expectedValue` and
+ * `newValue` and then use `element.classList.add` and `element.classList.remove` to modify the
+ * DOM.
+ *
+ * NOTE: Some platforms (such as NativeScript and WebWorkers) will not have `element.className`
+ * available and reading the value will result in `undefined`. This means that for those platforms
+ * we will always fail the check and will always use  `element.classList.add` and
+ * `element.classList.remove` to modify the `element`. (A good mental model is that we can do
+ * `element.className === expectedValue` but we may never know the actual value of
+ * `element.className`)
+ *
+ * @param renderer Renderer to use
+ * @param element The element which needs to be updated.
+ * @param expectedValue The expected (previous/old) value of the class list which we will use to
+ *        check if out of bounds modification has happened to the `element`.
+ * @param newValue The new class list to write.
+ */
+export function writeAndReconcileClass(
+    renderer: Renderer3, element: RElement, expectedValue: string, newValue: string): void {
+  if (element.className === expectedValue) {
+    // This is the simple/fast case where no one has written into element without our knowledge.
+    if (isProceduralRenderer(renderer)) {
+      renderer.setAttribute(element, 'class', newValue);
+    } else {
+      element.className = newValue;
+    }
+    ngDevMode && ngDevMode.rendererSetClassName++;
+  } else {
+    // The expected value is not the same as last value. Something changed the DOM element without
+    // our knowledge so we need to do reconciliation instead.
+    reconcileClassNames(renderer, element, expectedValue, newValue);
+  }
+}
+
+/**
+* Writes new `cssText` value in the DOM node.
+*
+* In its simplest form this function just writes the `newValue` into the `element.style.cssText`
+* property.
+*
+* However, under some circumstances this is more complex because there could be other code which
+* has added `style` information to the DOM element. In such a case writing our new value would
+* clobber what is already on the element and would result in incorrect behavior.
+*
+* To solve the above the function first reads the `element.style.cssText` to see if it matches the
+* `expectedValue`. (In our case `expectedValue` is just last value written into the DOM.) In this
+* way we can detect to see if anyone has modified the DOM since our last write.
+* - If we detect no change we simply write: `element.style.cssText = newValue`
+* - If we do detect change then we compute the difference between the `expectedValue` and
+* `newValue` and then use `element.style[property]` to modify the DOM.
+*
+* NOTE: Some platforms (such as NativeScript and WebWorkers) will not have `element.style`
+* available and reading the value will result in `undefined` This means that for those platforms we
+* will always fail the check and will always use  `element.style[property]` to
+* modify the `element`. (A good mental model is that we can do `element.style.cssText ===
+* expectedValue` but we may never know the actual value of `element.style.cssText`)
+*
+* @param renderer Renderer to use
+* @param element The element which needs to be updated.
+* @param expectedValue The expected (previous/old) value of the class list to write.
+* @param newValue The new class list to write
+*/
+export function writeAndReconcileStyle(
+    renderer: Renderer3, element: RElement, expectedValue: string, newValue: string): void {
+  const style = (element as HTMLElement).style;
+  if (style != null && style.cssText === expectedValue) {
+    // This is the simple/fast case where no one has written into element without our knowledge.
+    if (isProceduralRenderer(renderer)) {
+      renderer.setAttribute(element, 'style', newValue);
+    } else {
+      style.cssText = newValue;
+    }
+    ngDevMode && ngDevMode.rendererCssText++;
+  } else {
+    // The expected value is not the same as last value. Something changed the DOM element without
+    // our knowledge so we need to do reconciliation instead.
+    reconcileStyleNames(renderer, element, expectedValue, newValue);
+  }
+}
+
+/**
+ * Writes to `classNames` by computing the difference between `oldValue` and `newValue` and using
+ * `classList.add` and `classList.remove`.
+ *
+ * NOTE: Keep this a separate function so that `writeAndReconcileClass` is small and subject to
+ * inlining. (We expect that this function will be called rarely.)
+ *
+ * @param renderer Renderer to use when updating DOM.
+ * @param element The native element to update.
+ * @param oldValue Old value of `classNames`.
+ * @param newValue New value of `classNames`.
+ */
+function reconcileClassNames(
+    renderer: Renderer3, element: RElement, oldValue: string, newValue: string) {
+  const isProcedural = isProceduralRenderer(renderer);
+  computeClassChanges(oldValue, newValue).forEach((classValue, className) => {
+    if (classValue === true) {
+      if (isProcedural) {
+        (renderer as ProceduralRenderer3).addClass(element, className);
+      } else {
+        (element as HTMLElement).classList.add(className);
+      }
+      ngDevMode && ngDevMode.rendererAddClass++;
+    } else if (classValue === false) {
+      if (isProcedural) {
+        (renderer as ProceduralRenderer3).removeClass(element, className);
+      } else {
+        (element as HTMLElement).classList.remove(className);
+      }
+      ngDevMode && ngDevMode.rendererRemoveClass++;
+    }
+  });
+}
+
+/**
+ * Writes to `styles` by computing the difference between `oldValue` and `newValue` and using
+ * `styles.setProperty` and `styles.removeProperty`.
+ *
+ * NOTE: Keep this a separate function so that `writeAndReconcileStyle` is small and subject to
+ * inlining. (We expect that this function will be called rarely.)
+ *
+ * @param renderer Renderer to use when updating DOM.
+ * @param element The DOM element to update.
+ * @param oldValue Old value of `classNames`.
+ * @param newValue New value of `classNames`.
+ */
+function reconcileStyleNames(
+    renderer: Renderer3, element: RElement, oldValue: string, newValue: string) {
+  const isProcedural = isProceduralRenderer(renderer);
+  const changes = computeStyleChanges(oldValue, newValue);
+  changes.forEach((styleValue, styleName) => {
+    const newValue = styleValue.new;
+    if (newValue === null) {
+      if (isProcedural) {
+        (renderer as ProceduralRenderer3).removeStyle(element, styleName);
+      } else {
+        (element as HTMLElement).style.removeProperty(styleName);
+      }
+      ngDevMode && ngDevMode.rendererRemoveStyle++;
+    } else if (styleValue.old !== newValue) {
+      if (isProcedural) {
+        (renderer as ProceduralRenderer3).setStyle(element, styleName, newValue);
+      } else {
+        (element as HTMLElement).style.setProperty(styleName, newValue);
+      }
+      ngDevMode && ngDevMode.rendererSetStyle++;
+    }
+  });
+}
--- a/packages/core/src/render3/styling/style_differ.ts
+++ b/packages/core/src/render3/styling/style_differ.ts
@@ -1,0 +1,172 @@
+/**
+* @license
+* Copyright Google Inc. All Rights Reserved.
+*
+* Use of this source code is governed by an MIT-style license that can be
+* found in the LICENSE file at https://angular.io/license
+*/
+
+import {CharCode} from '../../util/char_code';
+import {consumeSeparator, consumeStyleKey, consumeStyleValue, consumeWhitespace} from './styling_parser';
+
+
+
+/**
+ * Stores changes to Style values.
+ * - `key`: style name.
+ * - `value`:
+ *   - `old`: previous value (or `null`)
+ *   - `new`: new value (or `null`).
+ *
+ * If `old === new` do nothing.
+ * If `old === null` then add `new`.
+ * If `new === null` then remove `old`.
+ */
+export type StyleChangesMap = Map<string, {old: string | null, new: string | null}>;
+
+/**
+ * Computes the diff between two style strings.
+ *
+ * Example:
+ *  `oldValue` => `"a: 1; b: 2, c: 3"`
+ *  `newValue` => `"b: 2; c: 4; d: 5;"`
+ * will result in:
+ * ```
+ * changes = Map(
+ *   'a', { old:  '1', new: null },
+ *   'b', { old:  '2', new:  '2' },
+ *   'c', { old:  '3', new:  '4' },
+ *   'd', { old: null, new:  '5' },
+ * )
+ * ``
+ *
+ * @param oldValue Previous style string.
+ * @param newValue New style string.
+ * @returns `StyleChangesArrayMap`.
+ */
+export function computeStyleChanges(oldValue: string, newValue: string): StyleChangesMap {
+  const changes: StyleChangesMap = new Map<string, {old: string | null, new: string | null}>();
+  parseKeyValue(oldValue, changes, false);
+  parseKeyValue(newValue, changes, true);
+  return changes;
+}
+
+/**
+ * Splits the style list into array, ignoring whitespace and add it to corresponding categories
+ * changes.
+ *
+ * @param text Style list to split
+ * @param changes Where changes will be stored.
+ * @param isNewValue `true` if parsing new value (effects how values get added to `changes`)
+ */
+export function parseKeyValue(text: string, changes: StyleChangesMap, isNewValue: boolean): void {
+  const end = text.length;
+  let start = 0;
+  while (start < end) {
+    const keyStart = consumeWhitespace(text, start, end);
+    const keyEnd = consumeStyleKey(text, keyStart, end);
+    if (keyEnd === keyStart) {
+      // we reached an end so just quit
+      break;
+    }
+    const valueStart = consumeSeparator(text, keyEnd, end, CharCode.COLON);
+    const valueEnd = consumeStyleValue(text, valueStart, end);
+    if (ngDevMode && valueStart === valueEnd) {
+      throw malformedStyleError(text, valueStart);
+    }
+    start = consumeSeparator(text, valueEnd, end, CharCode.SEMI_COLON);
+    const key = text.substring(keyStart, keyEnd);
+    const value = text.substring(valueStart, valueEnd);
+    processStyleKeyValue(changes, key, value, isNewValue);
+  }
+}
+
+/**
+ * Appends style `key`/`value` information into the list of `changes`.
+ *
+ * Once all of the parsing is complete, the `changes` will contain a
+ * set of operations which need to be performed on the DOM to reconcile it.
+ *
+ * @param changes An `StyleChangesMap which tracks changes.
+ * @param key Style key to be added to the `changes`.
+ * @param value Style value to be added to the `changes`.
+ * @param isNewValue true if `key`/`value` should be processed as new value.
+ */
+function processStyleKeyValue(
+    changes: StyleChangesMap, key: string, value: string, isNewValue: boolean): void {
+  if (isNewValue) {
+    // This code path is executed when we are iterating over new values.
+    const existing = changes.get(key);
+    if (existing === undefined) {
+      // Key we have not seen before
+      changes.set(key, styleKeyValue(null, value));
+    } else {
+      // Already seen, update value.
+      existing.new = value;
+    }
+  } else {
+    // This code path is executed when we are iteration over previous values.
+    changes.set(key, styleKeyValue(value, null));
+  }
+}
+
+function styleKeyValue(oldValue: string | null, newValue: string | null) {
+  return {old: oldValue, new: newValue};
+}
+
+/**
+ * Removes a style from a `cssText` string.
+ *
+ * @param cssText A string which contains styling.
+ * @param styleToRemove A style (and its value) to remove from `cssText`.
+ * @returns a new style text which does not have `styleToRemove` (and its value)
+ */
+export function removeStyle(cssText: string, styleToRemove: string): string {
+  let start = 0;
+  let end = cssText.length;
+  let lastValueEnd = 0;
+  while (start < end) {
+    const possibleKeyIndex = cssText.indexOf(styleToRemove, start);
+    if (possibleKeyIndex === -1) {
+      // we did not find anything, so just bail.
+      break;
+    }
+    while (start < possibleKeyIndex + 1) {
+      const keyStart = consumeWhitespace(cssText, start, end);
+      const keyEnd = consumeStyleKey(cssText, keyStart, end);
+      if (keyEnd === keyStart) {
+        // we reached the end
+        return cssText;
+      }
+      const valueStart = consumeSeparator(cssText, keyEnd, end, CharCode.COLON);
+      const valueEnd = consumeStyleValue(cssText, valueStart, end);
+      if (ngDevMode && valueStart === valueEnd) {
+        throw malformedStyleError(cssText, valueStart);
+      }
+      const valueEndSep = consumeSeparator(cssText, valueEnd, end, CharCode.SEMI_COLON);
+      if (keyStart == possibleKeyIndex && keyEnd === possibleKeyIndex + styleToRemove.length) {
+        if (valueEndSep == end) {
+          // This is a special case when we are the last key in a list, we then chop off the
+          // trailing separator as well.
+          cssText = cssText.substring(0, lastValueEnd);
+        } else {
+          cssText = cssText.substring(0, keyStart) + cssText.substring(valueEndSep, end);
+        }
+        end = cssText.length;
+        start = keyStart;
+        break;  // rescan.
+      } else {
+        // This was not the item we are looking for, keep going.
+        start = valueEndSep;
+      }
+      lastValueEnd = valueEnd;
+    }
+  }
+  return cssText;
+}
+
+function malformedStyleError(text: string, index: number) {
+  return new Error(
+      `Malformed style at location ${index} in string '` + text.substring(0, index) + '[>>' +
+      text.substring(index, index + 1) + '<<]' + text.substr(index + 1) + '\'.');
+}
--- a/packages/core/src/render3/styling/styling_parser.ts
+++ b/packages/core/src/render3/styling/styling_parser.ts
@@ -1,0 +1,153 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {CharCode} from '../../util/char_code';
+
+/**
+ * Returns index of next non-whitespace character.
+ *
+ * @param text Text to scan
+ * @param startIndex Starting index of character where the scan should start.
+ * @param endIndex Ending index of character where the scan should end.
+ * @returns Index of next non-whitespace character (May be the same as `start` if no whitespace at
+ *          that location.)
+ */
+export function consumeWhitespace(text: string, startIndex: number, endIndex: number): number {
+  while (startIndex < endIndex && text.charCodeAt(startIndex) <= CharCode.SPACE) {
+    startIndex++;
+  }
+  return startIndex;
+}
+
+/**
+ * Returns index of last char in class token.
+ *
+ * @param text Text to scan
+ * @param startIndex Starting index of character where the scan should start.
+ * @param endIndex Ending index of character where the scan should end.
+ * @returns Index after last char in class token.
+ */
+export function consumeClassToken(text: string, startIndex: number, endIndex: number): number {
+  while (startIndex < endIndex && text.charCodeAt(startIndex) > CharCode.SPACE) {
+    startIndex++;
+  }
+  return startIndex;
+}
+
+/**
+ * Consumes all of the characters belonging to style key and token.
+ *
+ * @param text Text to scan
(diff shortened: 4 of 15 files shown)


packages/core/src/util/array_utils.ts
```diff
@@ -70,4 +72,268 @@ export function newArray<T>(size: number, value?: T): T[] {
...
+    array[index] = array[index + count];
+    index++;
+  }
+  while (count--) {
+    array.pop();  // shrink the array
+  }
+}
+
+/**
+ * Same as `Array.splice(index, 0, value)` but faster.
```
pkozlowski-opensource: This part of the comment is incorrect, right?

packages/core/src/util/array_utils.ts
```diff
@@ -70,4 +72,268 @@ export function newArray<T>(size: number, value?: T): T[] {
...
+  while (end > index) {
+    const previousEnd = end - 1;
+    array[end] = array[previousEnd];
+    end = previousEnd;
+  }
+  array[index] = value;
+}
+
+/**
+ * Same as `Array.splice2(index, 0, value1, value2)` but faster.
```
pkozlowski-opensource: it gets any index, not only `0`, right?

packages/core/test/util/array_utils_spec.ts
```diff
@@ -6,20 +6,170 @@
...
+
+      expect(arrayMapGet(map, 'A')).toEqual(0);
+      expect(arrayMapGet(map, 'B')).toEqual(1);
+      expect(arrayMapGet(map, 'C')).toEqual(2);
+      expect(arrayMapGet(map, 'AA')).toEqual(undefined);
+
+      expect(arrayMapSet(map, 'B', -1)).toEqual(2);
+      expect(map).toEqual(['A', 0, 'B', -1, 'C', 2]);
+
+      expect(arrayMapDelete(map, 'AA')).toEqual(~2);
```
pkozlowski-opensource: why using bits inverting operator `~` here?

packages/core/src/render3/styling/reconcile.ts
```diff
@@ -0,0 +1,196 @@
...
+ * @param element The DOM element to update.
+ * @param oldValue Old value of `classNames`.
+ * @param newValue New value of `classNames`.
+ */
+function reconcileClassNames(
+    renderer: Renderer3, element: RElement, oldValue: string, newValue: string) {
+  const isProcedural = isProceduralRenderer(renderer);
+  // Reset all of the state arrays.
+  removals.length = additions.length = 0;
+  computeClassChanges(oldValue, newValue, removals, additions);
```
pkozlowski-opensource: When we look at the implementation of the VE's `NgClass` one can notice that we use the existing `IterableDiffer` to figure out a set of operations to get from one state to another. What is the advantage of introducing another, specialised diffing algorithm here?

packages/core/src/util/array_utils.ts
```diff
@@ -70,4 +72,268 @@ export function newArray<T>(size: number, value?: T): T[] {
...
+ * removed. This causes memory pressure and slows down code when most of the time we don't
+ * care about the deleted items array.
+ *
+ * https://jsperf.com/fast-array-splice (About 20x faster)
+ *
+ * @param array Array to splice
+ * @param index Index of element in array to remove.
+ * @param count Number of items to remove.
+ */
+export function arraySplice(array: any[], index: number, count: number): void {
```
petebacondarwin: So this should be named `arrayRemove` or something similar. It is no longer "splicing" - neither adding nor replacing items of the array.

packages/core/src/util/array_utils.ts
```diff
@@ -70,4 +72,268 @@ export function newArray<T>(size: number, value?: T): T[] {
...
+  const length = array.length - count;
+  while (index < length) {
+    array[index] = array[index + count];
+    index++;
+  }
+  while (count--) {
+    array.pop();  // shrink the array
+  }
```
petebacondarwin: Is this faster than setting the `length` of the array?

packages/core/src/render3/styling/class_differ.ts
```diff
@@ -0,0 +1,129 @@
...
+const duplicates: string[] = [];
+
+
+
+/**
+ * Splits the class list into array, ignoring whitespace and add it to corresponding categories
+ * (addition/removals).
+ *
+ * @param text Class list to split
+ * @param parts Where the parts will be stored
```
pkozlowski-opensource: there is no `parts` argument, as far as I can tell

packages/core/src/render3/styling/class_differ.ts
```diff
@@ -0,0 +1,129 @@
...
+  splitClassList(newValue, removals, additions, duplicates);
+}
+
+/**
+ * Reusable array for storing duplicate keys when computing reconciliations.
+ *
+ * This is an internal state of the reconciliation algorithm. It is needed because if
+ * old value is `A` and new value is `A A` (which can happen when concatenating strings)
+ * the insert would try to remove `A` twice from old value incorrectly believing that there was no
+ * `A` second time. This prevents that from happening.
```
pkozlowski-opensource: Huh, this is interesting! Could you please provide an example of the real-life template where such situation would happen? Naively, I would say that a binding says `remove A` we should remove it, no matter how many times it is present... I'm clearly missing context / understanding here...

packages/core/src/render3/styling/class_differ.ts
```diff
@@ -0,0 +1,129 @@
...
+ *
+ * @param oldValue Previous class-list string.
+ * @param newValue New class-list string.
+ * @param removals Array which will be populated with tokens which require removal after
+ *        reconciliation.
+ * @param additions Array which will be populated with tokens which require addition after
+ *        reconciliation.
+ */
+export function computeClassChanges(
+    oldValue: string, newValue: string, removals: string[], additions: string[]): void {
```
pkozlowski-opensource: I'm surprised that the `oldValue` is a string - I was assuming that we are storing some sort of a list / collection of old values and would have to compare this data structure to changes computed by CD. Or is it supposed to be used for cases where the `classList` is "externally" mutated? Clearly missing context / big-picture design here...

packages/core/src/render3/styling/reconcile.ts
```diff
@@ -0,0 +1,196 @@
...
+*/
+
+import {ArrayMap} from '../../util/array_utils';
+import {ProceduralRenderer3, RElement, Renderer3, isProceduralRenderer} from '../interfaces/renderer';
+import {computeClassChanges} from './class_differ';
+import {computeStyleChanges} from './style_differ';
+
+
+/**
+ * Writes new `className` (style) value in the DOM node.
```
pkozlowski-opensource: Not sure about the `(style)` part of the comment - we are dealing with the CSS class names here only, no?

packages/core/src/render3/styling/reconcile.ts
```diff
@@ -0,0 +1,196 @@
...
+
+/**
+ * Writes to `styles` by computing the difference between `oldValue` and `newValue` and using
+ * `styles.setProperty` and `styles.removeProperty`.
+ *
+ * NOTE: Keep this a separate function so that `writeAndReconcileStyle` is small and subject to
+ * inlining. (We expect that this function will be called rarely.)
+ *
+ * @param renderer Renderer to use when updating DOM.
+ * @param element The DOM element to update.
```
pkozlowski-opensource: DOM -> native?

packages/core/src/render3/styling/reconcile.ts
```diff
@@ -0,0 +1,196 @@
...
+ * @param oldValue Old value of `classNames`.
+ * @param newValue New value of `classNames`.
+ */
+function reconcileClassNames(
+    renderer: Renderer3, element: RElement, oldValue: string, newValue: string) {
+  const isProcedural = isProceduralRenderer(renderer);
+  // Reset all of the state arrays.
+  removals.length = additions.length = 0;
```
pkozlowski-opensource: Is it "worth the trouble" to share all those arrays? How could we measure this?

packages/core/src/render3/styling/reconcile.ts
mhevery: 1. `IterableDiffer` is super expensive both in size, memory and runtime.
2. `IterableDiffer` is for arrays here we are working strings not arrays. Not sure how this would even fit her.

packages/core/src/util/array_utils.ts
mhevery: renamed.

packages/core/src/util/array_utils.ts
mhevery: good idea, changed.

packages/core/src/util/array_utils.ts
mhevery: I think it is correct. What specifically do you think is wrong?

packages/core/src/util/array_utils.ts
mhevery: Opps, fixed.

packages/core/test/render3/styling_next/reconcile_spec.ts
```diff
@@ -0,0 +1,72 @@
...
+
+        writeAndReconcileStyle(
+            renderer, element, 'width: 100px;', 'color: red; height: 100px; width: 100px;');
+        expect(element.style.cssText).toEqual('color: red; height: 100px; width: 100px;');
+
+        writeAndReconcileStyle(renderer, element, 'color: red; height: 100px; width: 100px;', '');
+        expect(element.style.cssText).toEqual('');
+      });
+
+      it('should not clobber out of bound styles', () => {
```
kara: It would be good to have an acceptance test for this as well

packages/core/test/render3/styling_next/reconcile_spec.ts
```diff
@@ -0,0 +1,72 @@
...
+        element.style.cssText = 'color: red;';
+        writeAndReconcileStyle(renderer, element, '', 'width: 100px;');
+        expect(element.style.cssText).toEqual('color: red; width: 100px;');
+
+        writeAndReconcileStyle(renderer, element, 'width: 100px;', 'width: 100px; height: 100px;');
+        expect(element.style.cssText).toEqual('color: red; width: 100px; height: 100px;');
+
+        writeAndReconcileStyle(renderer, element, 'width: 100px; height: 100px;', '');
+        expect(element.style.cssText).toEqual('color: red;');
```
kara: Would be good to have a test for the case where there is an existing style for some property (triggering the reconciliation logic) and we have multiple style values (e.g. dir writes "color", comp writes "color", etc). We would want to make sure the last value for the conflicting style is the one written.

kara (changes requested):

packages/core/src/util/array_utils.ts
```diff
@@ -70,4 +72,265 @@ export function newArray<T>(size: number, value?: T): T[] {
...
+ *   - negative index if value not found. (`~index` to get the value where it should have been
+ *     located)
+ */
+export function arrayIndexOfSorted(array: string[], value: string): number {
+  return _arrayIndexOfSorted(array, value, 0);
+}
+
+
+/**
+ * `ArrayMap` is an array where even positions contain keys and odd positions contain values.
```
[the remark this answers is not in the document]
kara: Yeah `MapAsArray` isn't great either, but for me the "as" makes it clear that it's actually an array, whereas "ArrayMap" sounds like it's ultimately a map ("array-like map"). Maybe re-ordering it to `MapArray` would be slightly better?

packages/core/src/util/array_utils.ts
```diff
@@ -70,4 +74,329 @@ export function newArray<T>(size: number, value?: T): T[] {
...
+ *
+ * @param array A sorted array to binary search.
+ * @param value The value to look for.
+ * @returns index of the value.
+ *   - positive index if value found.
+ *   - negative index if value not found. (`~index` to get the value where it should have been
+ *     located)
+ */
+export function arrayIndexOfSorted(array: string[], value: string): number {
+  return _arrayIndexOfSorted(array, value, 0);
```
kara: Is it necessary to create the `arrayIndexOfSorted` function wrapper for this? Seems like we can simplify by calling `_arrayIndexOfSorted` directly and passing through 0. Other functions are delegating directly to `_arrayIndexOfSorted` already.

packages/core/src/util/array_utils.ts
```diff
@@ -70,4 +74,329 @@ export function newArray<T>(size: number, value?: T): T[] {
...
+}
+
+/**
+ * Remove item from array (Same as `Array.splice()` but faster.)
+ *
+ * `Array.splice()` is not as fast because it has to allocate an array for the elements which were
+ * removed. This causes memory pressure and slows down code when most of the time we don't
+ * care about the deleted items array.
+ *
+ * https://jsperf.com/fast-array-splice (About 20x faster)
```
kara: This benchmark seems slightly different than the removal code you have below. Can we benchmark the code below?

packages/core/src/util/array_utils.ts
```diff
@@ -70,4 +74,329 @@ export function newArray<T>(size: number, value?: T): T[] {
...
+}
+
+/**
+ * Same as `Array.splice(index, 0, value)` but faster.
+ *
+ * `Array.splice()` is not fast because it has to allocate an array for the elements which were
+ * removed. This causes memory pressure and slows down code when most of the time we don't
+ * care about the deleted items array.
+ *
+ * https://jsperf.com/fast-array-splice (About 20x faster)
```
kara: This benchmark doesn't appear to match the code below. Can you link one that adds values?

packages/core/src/render3/styling/class_differ.ts
```diff
@@ -0,0 +1,129 @@
...
+    }
+  } else {
+    // This code path is executed when we are iterating over previous values.
+    // This means that we store the tokens in `changes` with `false` (removals).
+    arrayMapSet(changes, token, false);
+  }
+}
+
+/**
+ * Removes a class from a `className` string.
```
kara: Where is this being used or where is it intended to be used? It's not clear to me where it will fit into the algorithm

packages/core/src/render3/styling/style_differ.ts
```diff
@@ -0,0 +1,185 @@
...
+ * Example:
+ *  `oldValue` => `"a: 1; b: 2, c: 3"`
+ *  `newValue` => `"b: 2; c: 4; d: 5;"`
+ * will result in:
+ * ```
+ * changes = [
+ *   'a', false,  '1', null,
+ *   'b',  null,  '2',  '2',
+ *   'c',  true,  '3',  '4',
+ *   'd',  true, null,  '5',
```
kara: Thinking out loud - if the new value is null, wouldn't you know it's a removal? And if the old value is null, wouldn't you know it's an addition?  Is recording the second value (`true` / `false` / `null`) really necessary?

packages/core/src/render3/styling/style_differ.ts
```diff
@@ -0,0 +1,185 @@
...
+export function computeStyleChanges(oldValue: string, newValue: string): StyleChangesArrayMap {
+  const changes: StyleChangesArrayMap = [] as any;
+  parseKeyValue(oldValue, changes, false);
+  parseKeyValue(newValue, changes, true);
+  return changes;
+}
+
+/**
+ * Splits the style list into array, ignoring whitespace and add it to corresponding categories
+ * (addition/removals).
```
kara:
```suggestion
 * in `changes`
```

packages/core/src/render3/styling/style_differ.ts
```diff
@@ -0,0 +1,185 @@
...
+ *
+ * @param text Style list to split
+ * @param parts Where the parts will be stored
```
kara: Remove outdated param

```suggestion
```

packages/core/src/render3/styling/styling_parser.ts
```diff
@@ -0,0 +1,168 @@
...
+ */
+export function consumeClassToken(text: string, startIndex: number, endIndex: number): number {
+  while (startIndex < endIndex && text.charCodeAt(startIndex) > CharCode.SPACE) {
+    startIndex++;
+  }
+  return startIndex;
+}
+
+/**
+ * Consumes all of the characters belonging to style token.
```
kara: I'm not sure what this means - how do you consume characters? I'm guessing this function returns the next index after the key starts?

In general, for all these "consume" methods: "consume"  is too generic to mean something to me. It seems like all of them return an index, so that should be reflected in the name to make it clear what it's for.

packages/core/src/render3/styling/styling_parser.ts
```diff
@@ -0,0 +1,168 @@
...
+ * @param startIndex Starting index of character where the scan should start.
+ * @param endIndex Ending index of character where the scan should end.
+ * @returns Index of next non-whitespace character (May be the same as `start` if no non-whitespace
+ *          at that location, or if `startIndex === endIndex`.)
+ */
+export function consumeStyleKeySeparator(
+    text: string, startIndex: number, endIndex: number): number {
+  startIndex = consumeWhitespace(text, startIndex, endIndex);
+  if (text.charCodeAt(startIndex++) !== CharCode.COLON) {
+    throw new Error(`Expected ':' at index ${startIndex} of: ${text}`);
```
kara: Can we make this a dev mode only error so it tree shakes? Same for the others in this file.

packages/core/src/render3/styling/style_differ.ts
```diff
@@ -0,0 +1,185 @@
...
+    const valueStart = consumeStyleKeySeparator(text, keyEnd, end);
+    const valueEnd = consumeStyleValue(text, valueStart, end);
+    if (valueStart !== valueEnd) {
+      start = consumeStyleValueSeparator(text, valueEnd, end);
+      const key = text.substring(keyStart, keyEnd);
+      const value = text.substring(valueStart, valueEnd);
+      processStyleKeyValue(changes, key, value, isNewValue);
+    } else {
+      // We don't have a value, this style is malformed, error.
+      throw new Error('Malformed style: ' + text);
```
kara: Make dev mode only?

kara (commented):
Some more comments and questions here (to prime the meeting for tomorrow).

Since we're adding more code here but not using it yet, maybe we can sanity check its impact by temporarily adding `writeAndReconcileClass` and `writeAndReconcileStyle` to one of the style instructions to see how much it *would* add to something like the CLI-hello-world-ivy payload (then removing it). I think it would help to see the size impact of this pr when looking at the total size with the follow up pr

packages/core/test/render3/styling_next/reconcile_spec.ts
mhevery: Per our discussion, we can only add acceptance tests after this is all integrated with #34000

packages/core/src/render3/styling/class_differ.ts
```diff
@@ -0,0 +1,124 @@
...
+import {consumeClassToken, consumeWhitespace} from './styling_parser';
+
+/**
+ * Computes the diff between two class-list strings.
+ *
+ * Example:
+ *  `oldValue` => `"A B C"`
+ *  `newValue` => `"A C D"`
+ * will result in:
+ * `['A', null, 'B', false, 'C', null, 'D', true],
```
pkozlowski-opensource: nit: this example suggest / looks like we are returning an array while a `Map` is returned. Maybe it would be less confusing when written as `{'A': null, 'B': false, 'C': null, 'D': true}`?

packages/core/src/render3/styling/reconcile.ts
```diff
@@ -0,0 +1,170 @@
...
+ * @param expectedValue The expected (previous/old) value of the class list which we will use to
+ *        check if out of bounds modification has happened to the `element`.
+ * @param newValue The new class list to write.
+ */
+export function writeAndReconcileClass(
+    renderer: Renderer3, element: RElement, expectedValue: string, newValue: string): void {
+  if (element.className === expectedValue) {
+    // This is the simple/fast case where no one has written into element without our knowledge.
+    if (isProceduralRenderer(renderer)) {
+      (renderer as ProceduralRenderer3).setAttribute(element, 'class', newValue);
```
pkozlowski-opensource: nit: a cast to the `ProceduralRenderer3` (`(renderer as ProceduralRenderer3)`) is not needed here since the `isProceduralRenderer` has the signature of `function isProceduralRenderer(renderer: ProceduralRenderer3 | ObjectOrientedRenderer3):renderer is ProceduralRenderer3` so TS should "know" that this is a `ProceduralRenderer3`

packages/core/src/render3/styling/reconcile.ts
```diff
@@ -0,0 +1,170 @@
...
+* @param expectedValue The expected (previous/old) value of the class list to write.
+* @param newValue The new class list to write
+*/
+export function writeAndReconcileStyle(
+    renderer: Renderer3, element: RElement, expectedValue: string, newValue: string): void {
+  const style = (element as HTMLElement).style;
+  if (style != null && style.cssText === expectedValue) {
+    // This is the simple/fast case where no one has written into element without our knowledge.
+    if (isProceduralRenderer(renderer)) {
+      (renderer as ProceduralRenderer3).setAttribute(element, 'style', newValue);
```
pkozlowski-opensource: nit: `(renderer as ProceduralRenderer3)` is not needed, see https://github.com/angular/angular/pull/34004/files#r355965814

pkozlowski-opensource: nit2: we start to have this pattern repeated in some places now through the code-base:

```javascript
if (isProceduralRenderer(renderer)) {
      (renderer as ProceduralRenderer3).setAttribute(element, 'class', newValue);
    } else {
      element.className = newValue;
    }
```

what do you think of creating a util (sth like `setNativeAttribute(renderer, attrName, attrValue)` to encapsulate this logic?

packages/core/src/render3/styling/reconcile.ts
```diff
@@ -0,0 +1,170 @@
...
+function reconcileClassNames(
+    renderer: Renderer3, element: RElement, oldValue: string, newValue: string) {
+  const isProcedural = isProceduralRenderer(renderer);
+  computeClassChanges(oldValue, newValue).forEach((classValue, className) => {
+    if (classValue === true) {
+      if (isProcedural) {
+        (renderer as ProceduralRenderer3).addClass(element, className);
+      } else {
+        (element as HTMLElement).classList.add(className);
+      }
```
pkozlowski-opensource: nit: what do you think about extracting this into a `nativeAddClass` / `nativeRemoveClass` to encapsulate this logic? I think I saw those lines repeated in other places, if I'm not mistaken? Also, we used to have those `ngDevMode` DOM manipulation counters - is this something we want to keep using / adding? 

Generally speaking it shows the "cost" of the `ObjectOrientedRenderer3` and I believe that we want to remove this overhead in the future.

packages/core/src/render3/styling/style_differ.ts
```diff
@@ -0,0 +1,175 @@
...
+ * If `old === new` do nothing.
+ * If `old === null` than add `new`.
+ * If `new === null` than remove `old`.
+ */
+export type StyleChangesArrayMap = Map<string, {old: string | null, new: string | null}>;
+
+/**
+ * See: `StyleChangesArrayMap`
+ */
+export const enum StyleChangesArrayMapEnum {
```
pkozlowski-opensource: I don't see this `StyleChangesArrayMapEnum` being used anywhere, is it here for the future work?

packages/core/src/render3/styling/style_differ.ts
```diff
@@ -0,0 +1,175 @@
...
+    while (start < possibleKeyIndex + 1) {
+      const keyStart = consumeWhitespace(cssText, start, end);
+      const keyEnd = consumeStyleKey(cssText, keyStart, end);
+      if (keyEnd === keyStart) {
+        // we reached the end
+        return cssText;
+      }
+      const valueStart = consumeSeparator(cssText, keyEnd, end, CharCode.COLON);
+      const valueEnd = consumeStyleValue(cssText, valueStart, end);
+      ngDevMode && assertNotEqual(valueStart, valueEnd, 'malformed style');
```
pkozlowski-opensource: nit: could we add the string of the malformed style here for easier debugging when this assert happens to fail? `assertNotEqual(valueStart, valueEnd, `malformed style: ${cssText}`);`

packages/core/test/render3/styling_next/class_differ_spec.ts
```diff
@@ -0,0 +1,108 @@
...
+
+    it('should not remove a sub-string', () => {
+      expect(removeClass('ABC', 'A')).toEqual('ABC');
+      expect(removeClass('ABC', 'B')).toEqual('ABC');
+      expect(removeClass('ABC', 'C')).toEqual('ABC');
+      expect(removeClass('ABC', 'AB')).toEqual('ABC');
+      expect(removeClass('ABC', 'BC')).toEqual('ABC');
+    });
+  });
+});
```
pkozlowski-opensource: Great tests in this file, thank you @mhevery !

packages/core/test/render3/styling_next/reconcile_spec.ts
```diff
@@ -0,0 +1,118 @@
...
+
+        writeAndReconcileStyle(
+            renderer, element, 'width: 100px;', 'color: red; height: 100px; width: 100px;');
+        expect(getSortedStyle(element)).toEqual('color: red; height: 100px; width: 100px;');
+
+        writeAndReconcileStyle(renderer, element, 'color: red; height: 100px; width: 100px;', '');
+        expect(getSortedStyle(element)).toEqual('');
+      });
+
+      fit('should not clobber out of bound styles', () => {
```
pkozlowski-opensource: `fit` left-over, fortunately it was reported by CI: https://circleci.com/gh/angular/angular/559718?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

packages/core/test/render3/styling_next/reconcile_spec.ts
```diff
@@ -0,0 +1,118 @@
...
+    if (names.indexOf(name) === -1) {
+      names.push(name);
+    }
+  }
+  names.sort();
+  return names.join(' ');
+}
+
+function getSortedStyle(element: HTMLElement): string {
+  debugger;
```
pkozlowski-opensource: `debugger;`

packages/core/src/render3/styling/reconcile.ts
mhevery: removed in favor of new allocation each time.

packages/core/src/render3/styling/class_differ.ts
mhevery: fixed

packages/core/src/render3/styling/style_differ.ts
mhevery: removed

packages/core/test/render3/styling_next/class_differ_spec.ts
mhevery: 👍

packages/core/src/render3/styling/reconcile.ts
mhevery: 1. added the counters.
2. The other locations should be deleted. This should be the only place where this happens.

packages/core/test/render3/styling_next/reconcile_spec.ts
mhevery: removed

packages/core/test/render3/styling_next/reconcile_spec.ts
mhevery: fixed

packages/core/src/render3/styling/style_differ.ts
mhevery: done

mhevery:
[presubmit](http://test/OCL:284838716:BASE:284838770:1576013933303:2a115280)

packages/core/src/render3/styling/styling_parser.ts
```diff
@@ -0,0 +1,158 @@
...
+  return startIndex;
+}
+
+/**
+ * Consumes all of the characters belonging to style key and token.
+ *
+ * @param text Text to scan
+ * @param startIndex Starting index of character where the scan should start.
+ * @param endIndex Ending index of character where the scan should end.
+ * @returns Index of next non-whitespace character (May be the same as `start` if no non-whitespace
```
kara: Is this right? From the name, it sounds like this should return the last index of the style key
```suggestion
 * @returns Index after last style key character 
```

packages/core/src/render3/styling/styling_parser.ts
```diff
@@ -0,0 +1,158 @@
...
+  return startIndex;
+}
+
+/**
+ * Consumes all whitespace and the separator `:` after the style key.
+ *
+ * @param text Text to scan
+ * @param startIndex Starting index of character where the scan should start.
+ * @param endIndex Ending index of character where the scan should end.
+ * @returns Index of next non-whitespace character (May be the same as `start` if no non-whitespace
```
kara:
```suggestion
 * @returns Index after separator and surrounding whitespace
```

packages/core/src/render3/styling/styling_parser.ts
```diff
@@ -0,0 +1,158 @@
...
+}
+
+
+/**
+ * Consumes style value honoring `url()` and `""` text.
+ *
+ * @param text Text to scan
+ * @param startIndex Starting index of character where the scan should start.
+ * @param endIndex Ending index of character where the scan should end.
+ * @returns Index of next non-whitespace character (May be the same as `start` if no non-whitespace
```
kara:
```suggestion
 * @returns Index after last style value character
```

packages/core/src/render3/styling/styling_parser.ts
```diff
@@ -0,0 +1,158 @@
...
+  return startIndex;
+}
+
+/**
+ * Returns index of last char in class token.
+ *
+ * @param text Text to scan
+ * @param startIndex Starting index of character where the scan should start.
+ * @param endIndex Ending index of character where the scan should end.
+ * @returns last char in class token (May be the same as `start` if no non-whitespace
```
kara:
```suggestion
 * @returns Index after last char in class token
```

packages/core/src/render3/styling/styling_parser.ts
```diff
@@ -0,0 +1,158 @@
...
+}
+
+/**
+ * Consumes all of the quoted characters.
+ *
+ * @param text Text to scan
+ * @param quoteCharCode CharCode of either `"` or `'` quote or `)` for `url(...)`.
+ * @param startIndex Starting index of character where the scan should start.
+ * @param endIndex Ending index of character where the scan should end.
+ * @returns Index of next non-whitespace character (May be the same as `start` if no non-whitespace
```
kara:
```suggestion
 * @returns Index after quoted characters
```

packages/core/test/render3/styling_next/reconcile_spec.ts
```diff
@@ -0,0 +1,121 @@
...
+        element.style.cssText = 'color: red;';
+        writeAndReconcileStyle(renderer, element, '', 'width: 100px;');
+        expect(getSortedStyle(element)).toEqual('color: red; width: 100px;');
+
+        writeAndReconcileStyle(renderer, element, 'width: 100px;', 'width: 100px; height: 100px;');
+        expect(getSortedStyle(element)).toEqual('color: red; height: 100px; width: 100px;');
+
+        writeAndReconcileStyle(renderer, element, 'width: 100px; height: 100px;', '');
+        expect(getSortedStyle(element)).toEqual('color: red;');
+      });
```
kara: Can we add a test where we change the value of an existing style (through reconciliation algorithm)? I believe that would fail right now

kara (commented):
Like how this has been simplified! Comments below are mostly to update outdated comments/typos, but I believe I may have found a bug here:

https://github.com/angular/angular/pull/34004#discussion_r356325558

packages/core/src/render3/styling/class_differ.ts
```diff
@@ -0,0 +1,131 @@
...
+ *   ['A', null],
+ *   ['B', false],
+ *   ['C', null],
+ *   ['D', true]
+ * ])
+ * ```
+ *
+ * @param oldValue Previous class-list string.
+ * @param newValue New class-list string.
+ * @returns An `Map` which will be filled with changes.
```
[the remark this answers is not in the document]
mhevery: fixed

packages/core/src/render3/styling/class_differ.ts
```diff
@@ -0,0 +1,131 @@
...
+  splitClassList(newValue, changes, true);
+  return changes;
+}
+
+/**
+ * Splits the class list into array, ignoring whitespace and add it to corresponding categories
+ * `changes`.
+ *
+ * @param text Class list to split
+ * @param changes ArrayMap which will be filled with changes. (`false` - remove; `null` - noop;
```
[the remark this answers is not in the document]
mhevery: fixed

packages/core/src/render3/styling/class_differ.ts
```diff
@@ -0,0 +1,131 @@
...
+    const tokenEnd = consumeClassToken(text, index, end);
+    if (tokenEnd !== index) {
+      processClassToken(changes, text.substring(index, tokenEnd), isNewValue);
+    }
+    index = tokenEnd;
+  }
+}
+
+/**
+ * Precesses the token by adding it to the `changes` array.
```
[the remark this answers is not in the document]
mhevery: fixed

packages/core/src/render3/styling/class_differ.ts
```diff
@@ -0,0 +1,131 @@
...
+ *        - `false` The token should be deleted. (It was in old list, but not in new list.)
+ *        - `null` The token should be ignored. (It was present in old list as well as new list.)
+ *        - `true` the token should be added. (It was only present in the new value)
+ * @param token Token to add to set.
+ * @param isNewValue True if invocation represents an addition (removal otherwise.)
+ *        - `false` means that we are processing the old value, which may need to be deleted.
+ *          Initially all tokens are labeled `false` (remove it.)
+ *        - `true` means that we are processing new value which may need to be added. If a token
+ *          with same key already exists with `false` then the resulting token is `null` (no
+ *          change.) If no token exist than the new token value is `true` (add it.)
```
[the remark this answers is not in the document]
mhevery: fixed

packages/core/src/render3/styling/reconcile.ts
```diff
@@ -0,0 +1,175 @@
...
+ *
+ * However, under some circumstances this is more complex because there could be other code which
+ * has added `class` information to the DOM element. In such a case writing our new value would
+ * clobber what is already on the element and would result in incorrect behavior.
+ *
+ * To solve the above the function first reads the `element.className` to see if it matches the
+ * `expectedValue`. (In our case `expectedValue` is just last value written into the DOM.) In this
+ * way we can detect to see if anyone has modified the DOM since our last write.
+ * - If we detect no change we simply write: `element.className = newValue`.
+ * - If we do detect change than we compute the difference between the `expectedValue` and
```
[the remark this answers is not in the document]
mhevery: fixed

packages/core/src/render3/styling/reconcile.ts
```diff
@@ -0,0 +1,175 @@
...
+*
+* However, under some circumstances this is more complex because there could be other code which
+* has added `style` information to the DOM element. In such a case writing our new value would
+* clobber what is already on the element and would result in incorrect behavior.
+*
+* To solve the above the function first reads the `element.style.cssText` to see if it matches the
+* `expectedValue`. (In our case `expectedValue` is just last value written into the DOM.) In this
+* way we can detect to see if anyone has modified the DOM since our last write.
+* - If we detect no change we simply write: `element.style.cssText = newValue`
+* - If we do detect change than we compute the difference between the `expectedValue` and
```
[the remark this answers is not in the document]
mhevery: fixed

packages/core/src/render3/styling/reconcile.ts
```diff
@@ -0,0 +1,175 @@
...
+  changes.forEach((styleValue, styleName) => {
+    const newValue = styleValue.new;
+    if (newValue === null) {
+      if (isProcedural) {
+        (renderer as ProceduralRenderer3).removeStyle(element, styleName);
+      } else {
+        (element as HTMLElement).style.removeProperty(styleName);
+      }
+      ngDevMode && ngDevMode.rendererRemoveStyle++;
+    } else if (styleValue.old === null) {
```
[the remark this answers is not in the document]
mhevery: Good catch! I added a test case for this use case and a fix.

packages/core/src/render3/styling/style_differ.ts
```diff
@@ -0,0 +1,173 @@
...
+
+/**
+ * Stores changes to Style values.
+ * - `key`: style name.
+ * - `value`:
+ *   - `old`: previous value (or `null`)
+ *   - `new`: new value (or `null`).
+ *
+ * If `old === new` do nothing.
+ * If `old === null` than add `new`.
```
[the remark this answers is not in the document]
mhevery: fixed

packages/core/src/render3/styling/style_differ.ts
```diff
@@ -0,0 +1,173 @@
...
+ * If `new === null` than remove `old`.
```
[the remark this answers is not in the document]
mhevery: fixed

packages/core/src/render3/styling/style_differ.ts
```diff
@@ -0,0 +1,173 @@
...
+ */
+export type StyleChangesArrayMap = Map<string, {old: string | null, new: string | null}>;
```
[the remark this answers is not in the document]
mhevery: fixed

packages/core/src/render3/styling/style_differ.ts
```diff
@@ -0,0 +1,173 @@
...
+    }
+    start = consumeSeparator(text, valueEnd, end, CharCode.SEMI_COLON);
+    const key = text.substring(keyStart, keyEnd);
+    const value = text.substring(valueStart, valueEnd);
+    processStyleKeyValue(changes, key, value, isNewValue);
+  }
+}
+
+/**
+ * Appends style `key`/`value` information into the list of `removal`/`additions`.
```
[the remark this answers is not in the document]
mhevery: fixed

packages/core/src/render3/styling/style_differ.ts
```diff
@@ -0,0 +1,173 @@
...
+ *
+ * Once all of the parsing is complete, the `removal`/`additions` will contain a
```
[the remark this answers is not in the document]
mhevery: fixed

packages/core/src/render3/styling/style_differ.ts
```diff
@@ -0,0 +1,173 @@
...
+ * set of operations which need to be performed on the DOM to reconcile it.
+ *
+ * @param changes An `ArrayMap3` which tracks changes.
```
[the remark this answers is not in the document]
mhevery: fixed

packages/core/src/render3/styling/style_differ.ts
```diff
@@ -0,0 +1,173 @@
...
+ * @param key Style key to be added to the `removal`/`additions`.
+ * @param value Style value to be added to the `removal`/`additions`.
```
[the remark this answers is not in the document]
mhevery: fixed

packages/core/src/render3/styling/style_differ.ts
```diff
@@ -0,0 +1,173 @@
...
+}
+
+/**
+ * Appends style `key`/`value` information into the list of `removal`/`additions`.
+ *
+ * Once all of the parsing is complete, the `removal`/`additions` will contain a
+ * set of operations which need to be performed on the DOM to reconcile it.
+ *
+ * @param changes An `ArrayMap3` which tracks changes.
+ * @param key Style key to be added to the `removal`/`additions`.
```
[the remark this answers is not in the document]
mhevery: fixed

packages/core/src/render3/styling/style_differ.ts
```diff
@@ -0,0 +1,173 @@
...
+ * @param value Style value to be added to the `removal`/`additions`.
+ * @param isNewValue true if `key`/`value` should be processed as new value.
+ */
+function processStyleKeyValue(
+    changes: StyleChangesArrayMap, key: string, value: string, isNewValue: boolean): void {
+  if (isNewValue) {
+    // This code path is executed when we are iterating over previous values.
```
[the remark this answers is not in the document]
mhevery: fixed

packages/core/src/render3/styling/style_differ.ts
```diff
@@ -0,0 +1,173 @@
...
+    const existing = changes.get(key);
+    if (existing === undefined) {
+      // Key we have not seen before
+      changes.set(key, styleKeyValue(null, value));
+    } else {
+      // Already seen, update value.
+      existing.new = value;
+    }
+  } else {
+    // This code path is executed when we are iteration over new values.
```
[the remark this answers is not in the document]
mhevery: fixed

packages/core/src/render3/styling/style_differ.ts
```diff
@@ -0,0 +1,173 @@
...
+      }
+      const valueStart = consumeSeparator(cssText, keyEnd, end, CharCode.COLON);
+      const valueEnd = consumeStyleValue(cssText, valueStart, end);
+      if (ngDevMode && valueStart === valueEnd) {
+        throw malformedStyleError(cssText, valueStart);
+      }
+      const valueEndSep = consumeSeparator(cssText, valueEnd, end, CharCode.SEMI_COLON);
+      if (keyStart == possibleKeyIndex && keyEnd === possibleKeyIndex + styleToRemove.length) {
+        if (valueEndSep == end) {
+          // This is a special case when we are the last key in a list, we than chop of the
```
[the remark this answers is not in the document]
mhevery: fix

packages/core/src/render3/styling/styling_parser.ts
mhevery: fixed

packages/core/src/render3/styling/styling_parser.ts
mhevery: fixed

packages/core/src/render3/styling/styling_parser.ts
mhevery: fixed

packages/core/src/render3/styling/styling_parser.ts
mhevery: fixed

packages/core/src/render3/styling/styling_parser.ts
```diff
@@ -0,0 +1,158 @@
...
+    } else if (ch === CharCode.DOUBLE_QUOTE || ch === CharCode.SINGLE_QUOTE) {
+      lastChIndex = i = consumeQuotedText(text, ch, i, endIndex);
+    } else if (
+        startIndex ===
+            i - 4 &&  // We have seen only 4 characters so far "URL(" (Ignore "foo_URL()")
+        ch3 === CharCode.U &&
+        ch2 === CharCode.R && ch1 === CharCode.L && ch === CharCode.OPEN_PAREN) {
+      lastChIndex = i = consumeQuotedText(text, CharCode.CLOSE_PAREN, i, endIndex);
+    } else if (ch > CharCode.SPACE) {
+      // if we have a non-whitespace character than capture its location
```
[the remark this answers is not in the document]
mhevery: fixed

packages/core/src/render3/styling/styling_parser.ts
mhevery: fixed

packages/core/test/render3/styling_next/reconcile_spec.ts
mhevery: yes, done.

mhevery:
[presubmit](http://test/OCL:284838716:BASE:284996704:1576083865887:ae55f614)

kara (approved):
LGTM

mhevery:
merged as 7cd465aa9b

packages/core/src/util/array_utils.ts
mhevery: removed

PR #34014, opened 2019-11-23
Several ngcc fixes

JoostK:
See individiual commits for descriptions.

### DIFF
--- a/packages/compiler-cli/ngcc/src/main.ts
+++ b/packages/compiler-cli/ngcc/src/main.ts
@@ -12,6 +12,7 @@ import {DepGraph} from 'dependency-graph';
 import * as os from 'os';
 import * as ts from 'typescript';
 
+import {replaceTsWithNgInErrors} from '../../src/ngtsc/diagnostics';
 import {AbsoluteFsPath, FileSystem, absoluteFrom, dirname, getFileSystem, resolve} from '../../src/ngtsc/file_system';
 
 import {CommonJsDependencyHost} from './dependencies/commonjs_dependency_host';
@@ -262,11 +263,13 @@ export function mainNgcc(
       const result = transformer.transform(bundle);
       if (result.success) {
         if (result.diagnostics.length > 0) {
-          logger.warn(ts.formatDiagnostics(result.diagnostics, bundle.src.host));
+          logger.warn(replaceTsWithNgInErrors(
+              ts.formatDiagnosticsWithColorAndContext(result.diagnostics, bundle.src.host)));
         }
         fileWriter.writeBundle(bundle, result.transformedFiles, formatPropertiesToMarkAsProcessed);
       } else {
-        const errors = ts.formatDiagnostics(result.diagnostics, bundle.src.host);
+        const errors = replaceTsWithNgInErrors(
+            ts.formatDiagnosticsWithColorAndContext(result.diagnostics, bundle.src.host));
         throw new Error(
             `Failed to compile entry-point ${entryPoint.name} due to compilation errors:\n${errors}`);
       }
--- a/packages/compiler-cli/ngcc/test/analysis/migration_host_spec.ts
+++ b/packages/compiler-cli/ngcc/test/analysis/migration_host_spec.ts
@@ -5,7 +5,9 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
-import {ErrorCode} from '../../../src/ngtsc/diagnostics';
+import * as ts from 'typescript';
+
+import {ErrorCode, makeDiagnostic} from '../../../src/ngtsc/diagnostics';
 import {AbsoluteFsPath, absoluteFrom} from '../../../src/ngtsc/file_system';
 import {runInEachFileSystem} from '../../../src/ngtsc/file_system/testing';
 import {ClassDeclaration, Decorator} from '../../../src/ngtsc/reflection';
@@ -13,6 +15,7 @@ import {AnalysisOutput, CompileResult, DecoratorHandler, DetectResult, HandlerPr
 import {DefaultMigrationHost} from '../../src/analysis/migration_host';
 import {AnalyzedClass, AnalyzedFile} from '../../src/analysis/types';
 import {NgccClassSymbol} from '../../src/host/ngcc_host';
+import {createComponentDecorator} from '../../src/migrations/utils';
 
 runInEachFileSystem(() => {
   describe('DefaultMigrationHost', () => {
@@ -23,6 +26,7 @@ runInEachFileSystem(() => {
     let mockEvaluator: any = {};
     let mockClazz: any;
     let mockDecorator: any = {name: 'MockDecorator'};
+    let diagnosticHandler = () => {};
     beforeEach(() => {
       _ = absoluteFrom;
       entryPointPath = _('/node_modules/some-package/entry-point');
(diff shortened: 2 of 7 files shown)


packages/compiler-cli/ngcc/test/analysis/migration_host_spec.ts
```diff
@@ -167,11 +176,36 @@ runInEachFileSystem(() => {
...
+        const host = new DefaultMigrationHost(
+            mockHost, mockMetadata, mockEvaluator, [handler], entryPointPath, analyzedFiles,
+            diagnostic => diagnostics.push(diagnostic));
+        mockClazz.getStart = () => 0;
+        mockClazz.getWidth = () => 0;
+
+        const decorator = createComponentDecorator(mockClazz, {selector: 'comp', exportAs: null});
+        host.injectSyntheticDecorator(mockClazz, decorator);
+
+        expect(diagnostics.length).toBe(1);
```
petebacondarwin: You could hide the internals of diagnostic chains here by using a ts helper:

```ts
        expect(ts.flattenDiagnosticMessageText(diagnostics[0].messageText, '\n'))
            .toEqual(
                `test diagnostic\n` +
                `  Occurs for @Component decorator inserted by an automatic migration\n` +
                `  @Component({ template: "", selector: "comp" })`);
```

packages/compiler-cli/src/ngtsc/diagnostics/src/util.ts
```diff
@@ -8,8 +8,13 @@
 
 import * as ts from 'typescript';
```
petebacondarwin: This import is not used.

packages/compiler-cli/src/ngtsc/diagnostics/src/util.ts
```diff
@@ -8,8 +8,13 @@
...
 
-const ERROR_CODE_MATCHER = /(\u001b\[\d+m ?)TS-99(\d+: ?\u001b\[\d+m)/g;
+const ERROR_CODE_COLORIZED_MATCHER = /(\u001b\[\d+m ?)TS-99(\d+: ?\u001b\[\d+m)/g;
+const ERROR_CODE_MATCHER = /( )TS-99(\d+)/g;
 
-export function replaceTsWithNgInErrors(errors: string): string {
-  return errors.replace(ERROR_CODE_MATCHER, '$1NG$2');
+export function replaceTsWithNgInErrors(errors: string, colorized: boolean): string {
```
petebacondarwin: Could you add a jsdoc here to explain what is going on (i.e. that TypeScript doesn't allow custom prefixes so we have to use -99 and then convert TS-99 to NG).

petebacondarwin (approved):
Great stuff @JoostK

packages/compiler-cli/ngcc/test/analysis/migration_host_spec.ts
JoostK: Excellent, that is a nice one!

packages/compiler-cli/ngcc/src/main.ts
```diff
@@ -248,11 +249,13 @@ export function mainNgcc(
       const result = transformer.transform(bundle);
       if (result.success) {
         if (result.diagnostics.length > 0) {
-          logger.warn(ts.formatDiagnostics(result.diagnostics, bundle.src.host));
+          logger.warn(replaceTsWithNgInErrors(
+              ts.formatDiagnostics(result.diagnostics, bundle.src.host), /* colorized */ false));
```
gkalpak: OOC, why using `colorized: false` here?

packages/compiler-cli/src/ngtsc/diagnostics/src/util.ts
```diff
@@ -6,10 +6,24 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import * as ts from 'typescript';
+const ERROR_CODE_COLORIZED_MATCHER = /(\u001b\[\d+m ?)TS-99(\d+: ?\u001b\[\d+m)/g;
+const ERROR_CODE_MATCHER = /( )TS-99(\d+: )/g;
```
gkalpak: Is it guaranteed that there will always be a leading space?

packages/compiler-cli/src/ngtsc/diagnostics/src/util.ts
```diff
@@ -6,10 +6,24 @@
...
+ *
+ * @param errors The formatted diagnostics
+ * @param colorized should be true when the diagnostics were formatted using
+ * `ts.formatDiagnosticsWithColorAndContext`, false otherwise.
+ */
+export function replaceTsWithNgInErrors(errors: string, colorized: boolean): string {
+  if (colorized) {
+    return errors.replace(ERROR_CODE_COLORIZED_MATCHER, '$1NG$2');
+  } else {
+    return errors.replace(ERROR_CODE_MATCHER, '$1NG$2');
```
gkalpak: Why not omit the `colorized` param and use a RegExp that can match both versions?
(Having to pass in `colorized` is error prone and can easily get out-of-sync with the actual implementation.)

gkalpak (changes requested):
Some minor comments - otherwise lgtm 👌

packages/compiler-cli/ngcc/src/main.ts
JoostK: Because diagnostics are formatted using `ts.formatDiagnostics`, not `ts.formatDiagnosticsWithColorsAndContext`. This difference has been mentioned in the params documentation for `colorized`.

packages/compiler-cli/src/ngtsc/diagnostics/src/util.ts
JoostK: Yes! The `ts.formatDiagnostic` is implemented to prefix an error code with `" TS"`.

packages/compiler-cli/src/ngtsc/diagnostics/src/util.ts
JoostK: If all is well, the non-colorized regex would work just fine for colorized diagnostics, it's just that the colorized regex is less likely to incorrectly replace some random string that happens to contain " TS-99##: ". So we _could_ drop the colorized version. I kept it the way it was as I didn't want to change the logic of how it already worked in ngtsc, as I presume it was intentionally written the way it was.

For non-colorized messages that are used in ngcc though, the colorized regex wouldn't work. Hence I added an additional one. I choose to do this in the same function, as that achieves a centralized place where this replacement occurs. One could argue that it should be two separate functions, the non-colorized one potentially only present in ngcc, however I prefered the centralized logic.

I am not convinced we should change any of this, if you feel differently please let me know :-)

gkalpak: Is there any particular reason we use non-colorized formatting in ngcc?

JoostK: @gkalpak As mentioned on Slack, there really isn't. I have changed ngcc to report colorized diagnostics with colors, [you can see them in action here](https://app.circleci.com/jobs/github/angular/ngcc-validation/2292/parallel-runs/3?filterBy=ALL).

gkalpak (approved):

JoostK:
merge-assistance: Current reviews should be sufficient, as I'm code owner of _fw-compiler_ (and it's only a comment change).

AndrewKushnir:
[Presubmit](http://test/OCL:284630278:BASE:284630318:1575929095987:5ecd385e)

PR #34015, opened 2019-11-23
fix(ivy): recompile on template change in ngc watch mode on Windows

### LINKED ISSUE #32869: ngc --watch ignores changes in templates (ivy, 9.0.0-next.8)
This is a repost of #31982, since:
- the bug is still there in 9.0.0-next.8 and 8.2.8 , 
- it is reproducible (as before) with the given example and 
- it has been closed unjustified as duplicate by @alxhub.

🐞 **bug report**
**Affected Package**

**Is this a regression?**
Yes, the previous version in which this bug was not present was: **8.1.2**
(But, due to #31322, it only worked on every second change of a template.)

**Description**
I run ngc in watch mode. Output directory is out-tsc.
Ivy is enabled.
I change a component's template.
ngc says "File change detected. Starting incremental compilation."
However, the template change does not result in an according change in the *.js file in out-tsc directory.
For changes in *.ts files, it works.

🔬 **Minimal Reproduction**
Example project is at: https://github.com/stefanocke/angular-cli-app-with-rollup
Please run:

npm i
npm run ngc-app-watch

(This is nothing more than running ngc in watch mode, as the title of this bug says.)

Then, change **app.component.html** and look for **app.component.js** in tsc-out.

You will see that, although the console says correctly "File change detected. Starting incremental compilation.", nothing has changed in app.component.js.

🌍 **Your Environment**
Angular Version:

Angular CLI: 8.3.6
Node: 10.16.0
OS: win32 x64
Angular: 9.0.0-next.8
... language-service, platform-browser, router

@angular-devkit/architect         0.900.0-next.6
@angular-devkit/build-angular     0.900.0-next.6
@angular-devkit/build-optimizer   0.900.0-next.6
@angular-devkit/build-webpack     0.900.0-next.6
@angular-devkit/core              9.0.0-next.6
@angular-devkit/schematics        8.3.6
@angular/cli                      8.3.6
@ngtools/webpack                  9.0.0-next.6
@schematics/angular               8.3.6
@schematics/update                0.803.6
rxjs                              6.5.2

**Anything else relevant?**
Ivy is enabled

JoostK:
In #33551, a bug in `ngc --watch` mode was fixed so that a component is
recompiled when its template file is changed. Due to insufficient
normalization of files paths, this fix did not have the desired effect
on Windows.

Fixes #32869

### DIFF
--- a/packages/compiler-cli/src/perform_watch.ts
+++ b/packages/compiler-cli/src/perform_watch.ts
@@ -246,11 +246,13 @@ export function performWatchCompilation(host: PerformWatchHost):
   }
 
   function watchedFileChanged(event: FileChangeEvent, fileName: string) {
+    const normalizedPath = path.normalize(fileName);
+
     if (cachedOptions && event === FileChangeEvent.Change &&
         // TODO(chuckj): validate that this is sufficient to skip files that were written.
         // This assumes that the file path we write is the same file path we will receive in the
         // change notification.
-        path.normalize(fileName) === path.normalize(cachedOptions.project)) {
+        normalizedPath === path.normalize(cachedOptions.project)) {
       // If the configuration file changes, forget everything and start the recompilation timer
       resetOptions();
     } else if (
@@ -263,12 +265,12 @@ export function performWatchCompilation(host: PerformWatchHost):
     if (event === FileChangeEvent.CreateDeleteDir) {
       fileCache.clear();
     } else {
-      fileCache.delete(path.normalize(fileName));
+      fileCache.delete(normalizedPath);
     }
 
-    if (!ignoreFilesForWatch.has(path.normalize(fileName))) {
+    if (!ignoreFilesForWatch.has(normalizedPath)) {
       // Ignore the file if the file is one that was written by the compiler.
-      startTimerForRecompilation(fileName);
+      startTimerForRecompilation(normalizedPath);
     }
   }
 
--- a/packages/compiler-cli/src/ngtsc/annotations/src/component.ts
+++ b/packages/compiler-cli/src/ngtsc/annotations/src/component.ts
@@ -273,7 +273,7 @@ export class ComponentDecoratorHandler implements
         const resourceStr = this.resourceLoader.load(resourceUrl);
         styles.push(resourceStr);
         if (this.depTracker !== null) {
-          this.depTracker.addResourceDependency(node.getSourceFile(), resourceUrl);
+          this.depTracker.addResourceDependency(node.getSourceFile(), absoluteFrom(resourceUrl));
         }
       }
     }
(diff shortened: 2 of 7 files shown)


JoostK:
@stefanocke Hi Stefan, could you please test the artifacts from this PR on Windows?

Artifacts: https://app.circleci.com/jobs/github/angular/angular/540987/artifacts
Instructions: https://github.com/angular/angular/blob/master/docs/DEVELOPER.md#archives-for-each-package

I am expecting this PR to fix your issue, hoping you can confirm. Thanks!

stefanocke:
@JoostK , sorry, but I get a 404 on https://app.circleci.com/jobs/github/angular/angular/540987/artifacts

stefanocke:
@JoostK , it works for me. :-)

stefanocke:
Hello @JoostK , any chance to get it merged in the next RC? Thank you.

JoostK:
@stefanocke yeah, I can look into reviving this PR. There has been quite some work around incremental compilations and dependency tracking that conflicted heavily with this PR so it has gotten outdated.

packages/compiler-cli/src/perform_watch.ts
```diff
@@ -246,11 +246,13 @@ export function performWatchCompilation(host: PerformWatchHost):
   }
 
   function watchedFileChanged(event: FileChangeEvent, fileName: string) {
+    const normalizedPath = path.normalize(fileName);
```
petebacondarwin: Why the changes in this file? It looks like a separate bit of tidying up that are not relevant to this commit, no?
If it is necessary for this fix, then should we not be using a platform independent normalization of the path?

petebacondarwin (approved):
One minor concern.
Thanks  @JoostK !

packages/compiler-cli/src/perform_watch.ts
JoostK: The primary change here is passing the normalized path to `startTimerForRecompilation`. While I believe (I don't recall exactly) this is indeed not strictly necessary for this to work, but it makes more sense to me to be passing the normalized path.

In any case, this code resides in "ngc" where we don't have the abstracted filesystem, or did you mean something else?

petebacondarwin: Oh I didn't notice it was outside of ngtsc.
But you could try `path.posix.normalize`?

petebacondarwin (approved):

packages/compiler-cli/src/perform_watch.ts
JoostK: I think that only switching over to `path.posix.normalize` would actually also resolve this issue, even without the change to ngtsc. I made the ngtsc changes so that we can no longer go wrong in this regard, as the branded paths ensure the paths are all standardized.

Actually, come to think of it, the `CompilerHost.getModifiedResourceFiles` can be provided by third parties which may or may not use posix paths, so the ngtsc change is still valuable.

In the end I don't _think_ that changing this to `path.posix.normalize` would behave differently (apart from the paths containing forward slashes, instead of backwards slashes).

JoostK:
merge-assistance: Pete's review should be sufficient, I am code-owner of fw-compiler.

devversion (approved):
LGTM for the BUILD file update (for which pullapprove requested review from dev-infra)

mhevery:
[presubmit](http://test/OCL:293157101:BASE:293157144:1580835701660:3a873610)

PR #34017, opened 2019-11-23
fix(ivy): take styles extracted from template into account in JIT mode

AndrewKushnir:
Prior to this commit, all styles extracted from Component's template (defined using <style> tags) were ignored by JIT compiler, so only `styles` array values defined in @Component decorator were used. This change updates JIT compiler to take styles extracted from the template into account. It also ensures correct order where `styles` array values are applied first and template styles are applied second.

This PR resolves FW-1744.

## PR Type

- [x] Bugfix

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/test/acceptance/styling_spec.ts
+++ b/packages/core/test/acceptance/styling_spec.ts
@@ -2410,6 +2410,33 @@ describe('styling', () => {
           .toEqual('url("https://i.imgur.com/4AiXzf8.jpg")');
     });
   });
+
+  isBrowser && it('should process <style> tag contents extracted from template', () => {
+    @Component({
+      template: `
+        <style>
+          div { width: 10px; }
+        </style>
+        <div></div>
+      `,
+      styles: [
+        'div { width: 100px; }',
+      ]
+    })
+    class MyComp {
+    }
+
+    TestBed.configureTestingModule({
+      declarations: [MyComp],
+    });
+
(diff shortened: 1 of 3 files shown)


packages/core/test/acceptance/styling_spec.ts
```diff
@@ -2410,6 +2410,33 @@ describe('styling', () => {
           .toEqual('url("https://i.imgur.com/4AiXzf8.jpg")');
     });
   });
+
+  isBrowser && it('should process <style> tag contents extracted from template', () => {
```
mhevery: why `isBrowser`?

mhevery (approved):

packages/core/test/acceptance/styling_spec.ts
AndrewKushnir: I use `getComputedStyle` in that test to verify that the styles were actually processed and applied correctly (in the right order).

AndrewKushnir:
- [VE Presubmit](https://test.corp.google.com/ui#id=OCL:282430978:BASE:282431032:1574720023046:963f6d9b)
- [Ivy Presubmit](https://test.corp.google.com/OCL:282430978:BASE:282431116:1574720054592:156c64d0)

AndrewKushnir:
Caretaker: using Misko's approval as global (Misko confirmed) to cover for `fw-compiler` group. Thank you.

PR #34021, opened 2019-11-24
perf(ivy): support simple generic type constraints in local type ctors

JoostK:
In Ivy's template type checker, type constructors are created for all
directive types to allow for accurate type inference to work. The type
checker has two strategies for dealing with such type constructors:

1. They can be emitted local to the type check block/type check file.
2. They can be emitted as static `ngTypeCtor` field into the directive
itself.

The first strategy is preferred, as it avoids having to update the
directive type which would cause a more expensive rebuild. However, this
strategy is not suitable for directives that have constrained generic
types, as those constraints would need to be present on the local type
constructor declaration. This is not trivial, as it requires that any
type references within a type parameter's constraint are imported into
the local context of the type check block.

For example, lets consider the `NgForOf` directive from '@angular/core'
looks as follows:

```typescript
import {NgIterable} from '@angular/core';

export class NgForOf<T, U extends NgIterable<T>> {}
```

The type constructor will then have the signature:
`(o: Pick<i1.NgForOf<T, U>, 'ngForOf'>) => i1.NgForOf<T, U>`

Notice how this refers to the type parameters `T` and `U`, so the type
constructor needs to be emitted into a scope where those types are
available, _and_ have the correct constraints.

Previously, the template type checker would detect the situation where a
type parameter is constrained, and would emit the type constructor
using strategy 2; within the directive type itself. This approach makes
any type references with the generic type constraints lexically
available:

```typescript
export class NgForOf<T, U extends NgIterable<T>> {
  static ngTypeCtor<T = any, U extends NgIterable<T> = any>
    (o: Pick<NgForOf<T, U>, 'ngForOf'>): NgForOf<T, U> { return null!; }
}
```

This commit introduces the ability to reify a type parameter with
constraints into a different context, under the condition that it can
be imported from an absolute module. This allows a generic type
constructor to be emitted into a type check block or type check file
according to strategy 1, as its generic types can be reconstructed using
the reification mechanism:

```typescript
import * as i0 from '@angular/core';
import * as i1 from '@angular/common';

const _ctor1: <T = any, U extends i0.NgIterable<T> = any>
  (o: Pick<i1.NgForOf<T, U>, 'ngForOf'>) => i1.NgForOf<T, U> = null!;
```

Notice how the generic type constraint of `U` has resulted in an import
of `@angular/core`, and the `NgIterable` is transformed into a qualified
name during the reification process.

Resolves FW-1739

### DIFF
--- a/packages/compiler-cli/src/ngtsc/transform/src/declaration.ts
+++ b/packages/compiler-cli/src/ngtsc/transform/src/declaration.ts
@@ -205,6 +205,7 @@ export class IvyDeclarationDtsTransform implements DtsTransform {
     const newMembers = fields.map(decl => {
       const modifiers = [ts.createModifier(ts.SyntaxKind.StaticKeyword)];
       const typeRef = translateType(decl.type, imports);
+      emitAsSingleLine(typeRef);
       return ts.createProperty(
           /* decorators */ undefined,
           /* modifiers */ modifiers,
@@ -223,6 +224,11 @@ export class IvyDeclarationDtsTransform implements DtsTransform {
         /* heritageClauses */ clazz.heritageClauses,
         /* members */[...members, ...newMembers]);
   }
+}
+
+function emitAsSingleLine(node: ts.Node) {
+  ts.setEmitFlags(node, ts.EmitFlags.SingleLine);
+  ts.forEachChild(node, emitAsSingleLine);
 }
 
 export class ReturnTypeTransform implements DtsTransform {
--- a/packages/compiler-cli/src/ngtsc/translator/src/translator.ts
+++ b/packages/compiler-cli/src/ngtsc/translator/src/translator.ts
@@ -6,7 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {ArrayType, AssertNotNull, BinaryOperator, BinaryOperatorExpr, BuiltinType, BuiltinTypeName, CastExpr, ClassStmt, CommaExpr, CommentStmt, ConditionalExpr, DeclareFunctionStmt, DeclareVarStmt, Expression, ExpressionStatement, ExpressionType, ExpressionVisitor, ExternalExpr, ExternalReference, FunctionExpr, IfStmt, InstantiateExpr, InvokeFunctionExpr, InvokeMethodExpr, JSDocCommentStmt, LiteralArrayExpr, LiteralExpr, LiteralMapExpr, MapType, NotExpr, ReadKeyExpr, ReadPropExpr, ReadVarExpr, ReturnStatement, Statement, StatementVisitor, StmtModifier, ThrowStmt, TryCatchStmt, Type, TypeVisitor, TypeofExpr, WrappedNodeExpr, WriteKeyExpr, WritePropExpr, WriteVarExpr} from '@angular/compiler';
+import {ArrayType, AssertNotNull, BinaryOperator, BinaryOperatorExpr, BuiltinType, BuiltinTypeName, CastExpr, ClassStmt, CommaExpr, CommentStmt, ConditionalExpr, DeclareFunctionStmt, DeclareVarStmt, Expression, ExpressionStatement, ExpressionType, ExpressionVisitor, ExternalExpr, FunctionExpr, IfStmt, InstantiateExpr, InvokeFunctionExpr, InvokeMethodExpr, JSDocCommentStmt, LiteralArrayExpr, LiteralExpr, LiteralMapExpr, MapType, NotExpr, ReadKeyExpr, ReadPropExpr, ReadVarExpr, ReturnStatement, Statement, StatementVisitor, StmtModifier, ThrowStmt, TryCatchStmt, Type, TypeVisitor, TypeofExpr, WrappedNodeExpr, WriteKeyExpr, WritePropExpr, WriteVarExpr} from '@angular/compiler';
 import {LocalizedString} from '@angular/compiler/src/output/output_ast';
 import * as ts from 'typescript';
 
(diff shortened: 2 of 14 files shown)


packages/compiler-cli/src/ngtsc/typecheck/src/type_parameter_reifier.ts
```diff
@@ -0,0 +1,102 @@
...
+
+    // If the declaration corresponds with a local type parameter, the type reference can be used
+    // as is.
+    if (this.isLocalTypeParameter(declaration.node)) {
+      return type;
+    }
+
+    // If the declaration has not been imported from an external module, it cannot be reified. The
+    // reason being that it's unlikely that an import can be written to the declaration from
+    // anywhere.
```
petebacondarwin: When you say anywhere, isn't it true that the type check file will be local to the application being built, so if the original application source file can import the local declaration then the type check file ought to be able to as well?

petebacondarwin: Can you give an example of when it is not possible to import the dependency?

petebacondarwin (dismissed):
This is awesome work @JoostK . I couldn't find anything wrong with the implementation. Only nice tidy code, full tests and an excellent commit message!

---

I am sorry to bike-shed but "reify" is not the correct term to use here - and it actually made it much harder for me to work out what this PR was about.

Reifying is taking something abstract and making it concrete. This is not what this code is doing.

What this is really about is moving a type to a new file, when it is possible. I.E. the type that is being used in the file where a directive is being defined needs to be moved to the file containing the type checking code.

I believe that if you change all instances of `reify` to `relocate` then this makes a lot more sense. E.g.

```ts
  private relocateTypeParameters(declaration: ClassDeclaration<ts.ClassDeclaration>):
      ts.TypeParameterDeclaration[]|undefined {
    const relocator = new TypeParameterRelocator(declaration.typeParameters, this.reflector);
    return relocator.relocate(ref => this.referenceType(ref));
  }
```

packages/compiler-cli/src/ngtsc/typecheck/src/type_parameter_reifier.ts
```diff
@@ -0,0 +1,102 @@
...
+    //
+    // Note: strictly speaking, even declarations imported from external modules might not be
+    // referred to from anywhere due to transitive dependency constraints.
+    if (declaration.viaModule === null) {
```
JoostK: Note to self: consider always returning a `Reference` here, then decide in `TypeReifier` whether it's eligible based on `Reference.hasOwningModuleGuess`. It already throws an error in that case, so `isTypeReifiable` should be updated to return false if  `Reference.hasOwningModuleGuess` is false. This moves all eligibility checks into `TypeReifier` which is a more appropriate place.

packages/compiler-cli/src/ngtsc/typecheck/test/type_parameter_reifier_spec.ts
```diff
@@ -0,0 +1,169 @@
...
+
+      expect(reifier.isReifiable()).toBe(false);
+      expect(() => reify(reifier))
+          .toThrowError('Unable to reify a reference to a non reifiable declaration.');
+    });
+
+    it('cannot reify references to local declarations as nested type arguments', () => {
+      const reifier = createReifier(`
+          import {NgIterable} from '@angular/core';
+          import {Internal} from './internal';
```
JoostK: Note to self: remove import.

packages/compiler-cli/src/ngtsc/typecheck/test/type_parameter_reifier_spec.ts
```diff
@@ -0,0 +1,169 @@
...
+
+          export class TestClass<T, U extends NgIterable<T>> {}`);
+
+      expect(reifier.isReifiable()).toBe(true);
+      expect(reify(reifier)).toEqual('<T, U extends test.NgIterable<T>>');
+    });
+
+    it('cannot reify references to local declarations', () => {
+      const reifier = createReifier(`
+          import {Internal} from './internal';
```
JoostK: Note to self: remove import.

packages/compiler-cli/src/ngtsc/typecheck/src/type_parameter_reifier.ts
alxhub: @petebacondarwin in general, yes, we _should_ be able to import such identifiers. However, I'd like to limit the scope of this change until we've thought through all the ramifications regarding rootDirs and the other quirks of TS module resolution. I'm mostly concerned that the template type-checking file is somewhat randomly placed right now.

packages/compiler-cli/src/ngtsc/translator/src/translator.ts
```diff
@@ -475,13 +475,15 @@ export class TypeTranslatorVisitor implements ExpressionVisitor, TypeVisitor {
...
 
     const typeName = moduleImport ?
-        ts.createPropertyAccess(ts.createIdentifier(moduleImport), symbolIdentifier) :
+        ts.createQualifiedName(ts.createIdentifier(moduleImport), symbolIdentifier) :
         symbolIdentifier;
 
-    const typeArguments =
-        ast.typeParams ? ast.typeParams.map(type => type.visitType(this, context)) : undefined;
+    if (ast.typeParams === null) {
+      return typeName;
```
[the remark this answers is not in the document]
alxhub: (per discussion, there's good reason for this, so resolving).

packages/compiler-cli/src/ngtsc/transform/src/declaration.ts
```diff
@@ -225,6 +226,11 @@ export class IvyDeclarationDtsTransform implements DtsTransform {
   }
 }
 
+function emitAsSingleLine(node: ts.Node) {
+  ts.setEmitFlags(node, ts.EmitFlags.SingleLine);
```
alxhub: Whoa, cool!

packages/compiler-cli/src/ngtsc/typecheck/src/type_constructor.ts
```diff
@@ -188,9 +188,16 @@ function generateGenericArgs(params: ReadonlyArray<ts.TypeParameterDeclaration>)
...
 }
 
-export function requiresInlineTypeCtor(node: ClassDeclaration<ts.ClassDeclaration>): boolean {
-  // The class requires an inline type constructor if it has constrained (bound) generics.
-  return !checkIfGenericTypesAreUnbound(node);
+export function requiresInlineTypeCtor(
+    node: ClassDeclaration<ts.ClassDeclaration>, host: ReflectionHost): boolean {
+  // The class requires an inline type constructor if it has generic type bounds that can not be
+  // reified into a different context.
+  return !checkIfGenericTypeBoundsAreReifiable(node, host);
```
alxhub: Change the "reified" terminology here :)

alxhub (approved):

packages/compiler-cli/src/ngtsc/typecheck/src/type_constructor.ts
JoostK: Hahaha, the very first iteration of the code started with a function named `isTypeContextFree`, which became `isTypeReifiable` as it became more complete. I'll have a look at changing the `canEmit` terminology into `isContextFree` to see how clear that would be.

alxhub:
Updated:

[Presubmit](https://test.corp.google.com/ui#id=OCL:286084476:BASE:286651330:1576885012249:fb0e1c70)
[Ivy Presubmit](https://test.corp.google.com/ui#id=OCL:286084476:BASE:286651330:1576885012266:1be7585d)

PR #34024, opened 2019-11-24
fix(ivy): prevent unknown element check for AOT-compiled components

### LINKED ISSUE #33976: bug(ivy): runtime error for unknown element even with NO_ERRORS_SCHEMA
The ivy runtime throws an error if there is an unknown element, even if the NgModule contains the `NO_ERRORS_SCHEMA`.

See the reproduction here: 

```
git clone https://github.com/petebacondarwin/ivy-schema-test && cd ivy-schema-test && npm i && npm start -o
```

Open console in the browser to see the error:

```
core.js:6068 ERROR Error: 'some-element' is not a known element:
1. If 'some-element' is an Angular component, then verify that it is part of this module.
2. If 'some-element' is a Web Component then add 'CUSTOM_ELEMENTS_SCHEMA' to the '@NgModule.schemas' of this component to suppress this message.
    at validateElement (core.js:23517)
    at ɵɵelementStart (core.js:23323)
    at Module.ɵɵelement (core.js:23389)
    at AppComponent_Template (app.component.html:2)
    at executeTemplate (core.js:13880)
    at renderView (core.js:13682)
    at renderComponent (core.js:15111)
    at renderChildComponents (core.js:13484)
    at renderView (core.js:13708)
    at ComponentFactory$1.create (core.js:34077)
```

The `AppComponent` template looks like:

```html
<div>
  <some-element></some-element>
</div>
```

but the `AppModule` has the `NO_ERRORS_SCHEMA`:

```ts
@NgModule({
  declarations: [
    AppComponent
  ],
  imports: [
    BrowserModule
  ],
  providers: [],
  schemas: [NO_ERRORS_SCHEMA],
  bootstrap: [AppComponent]
})
export class AppModule { }
```

Note that if you remove the `NO_ERRORS_SCHEMA` then the build fails, as expected, during AOT compile time.

This looks like a regression that was introduced by https://github.com/angular/angular/commit/c83f5013bfb0182dc653d7e5d8a1750e09ea50e9

AndrewKushnir:
Prior to this commit, the unknown element can happen twice for AOT-compiled components: once during compilation and once again at runtime. Due to the fact that `schemas` information is not present on Component and NgModule defs after AOT compilation, the second check (at runtime) may fail, even though the same check was successful at compile time. This commit updates the code to avoid the second check for AOT-compiled components by checking whether `schemas` information is present in a logic that executes the unknown element check.

This PR resolves FW-1736.
Resolves #33976.

## PR Type

- [x] Bugfix

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/render3/instructions/element.ts
+++ b/packages/core/src/render3/instructions/element.ts
@@ -255,6 +255,14 @@ function setDirectiveStylingInput(
 
 function validateElement(
     hostView: LView, element: RElement, tNode: TNode, hasDirectives: boolean): void {
+  const schemas = hostView[TVIEW].schemas;
+
+  // If `schemas` is set to `null`, that's an indication that this Component was compiled in AOT
+  // mode where this check happens at compile time. In JIT mode, `schemas` is always present and
+  // defined as an array (as an empty array in case `schemas` field is not defined) and we should
+  // execute the check below.
+  if (schemas === null) return;
+
   const tagName = tNode.tagName;
 
   // If the element matches any directive, it's considered as valid.
--- a/packages/core/src/render3/jit/module.ts
+++ b/packages/core/src/render3/jit/module.ts
@@ -128,6 +128,13 @@ export function compileNgModuleDefs(
               schemas: ngModule.schemas ? flatten(ngModule.schemas) : null,
               id: ngModule.id || null,
             });
+        // Set `schemas` on ngModuleDef to an empty array in JIT mode to indicate that runtime
+        // should verify that there are no unknown elements in a template. In AOT mode, that check
+        // happens at compile time and `schemas` information is not present on Component and Module
+        // defs after compilation (so the check doesn't happen the second time at runtime).
+        if (!ngModuleDef.schemas) {
+          ngModuleDef.schemas = [];
+        }
       }
       return ngModuleDef;
     }
--- a/packages/compiler-cli/test/ngtsc/ngtsc_spec.ts
+++ b/packages/compiler-cli/test/ngtsc/ngtsc_spec.ts
@@ -3009,6 +3009,44 @@ runInEachFileSystem(os => {
               `/*@__PURE__*/ (function () { i0.ɵsetClassMetadata(Service, [{ type: Injectable, args: [{ providedIn: 'root' }] }], null, null); })();`);
     });
 
+    it('should not include `schemas` in component and module defs', () => {
+      env.write('test.ts', `
+        import {Component, NgModule, NO_ERRORS_SCHEMA} from '@angular/core';
+
+        @Component({
+          selector: 'comp',
+          template: '<custom-el></custom-el>',
+          schemas: [NO_ERRORS_SCHEMA],
+        })
+        class MyComp {}
+
+        @NgModule({
+          declarations: [MyComp],
+          schemas: [NO_ERRORS_SCHEMA],
+        })
+        class MyModule {}
+      `);
+
+      env.driveMain();
+      const jsContents = trim(env.getContents('test.js'));
+      expect(jsContents).toContain(trim(`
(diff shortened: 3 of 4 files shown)


AndrewKushnir:
@petebacondarwin @crisbeto could you please have a look at this draft PR (we discussed these changes offline) when you get a chance? Thank you.

packages/core/src/render3/jit/module.ts
```diff
@@ -128,6 +128,13 @@ export function compileNgModuleDefs(
               schemas: ngModule.schemas ? flatten(ngModule.schemas) : null,
               id: ngModule.id || null,
             });
+        // Set `schemas` on ngModuleDef to an empty array in JIT mode to indicate that runtime
+        // should verify that there are no unknown elements in a template. In AOT mode, that check
+        // happens at compile time and `schemas` information is not present on Component and Module
+        // defs after compilation (so the check doesn't happen the second time at runtime).
+        if (!ngModuleDef.schemas) {
+          ngModuleDef.schemas = [];
```
crisbeto: I think that this might be hard to follow if somebody were to try and work their way backwards to figure out why schemas are being skipped during runtime for AoT components. Maybe we can make it a bit more obvious by adding another parameter to `compileNgModule` like `checkSchemasAtRuntime` and then have the compiler generate the empty array based on it?

petebacondarwin: If the project is compiled via AOT then it would never succeed if the schema was not valid, right?
So there shouldn't be a situation where it could fail at runtime...

packages/core/src/render3/instructions/element.ts
```diff
@@ -245,6 +245,14 @@ function setDirectiveStylingInput(
 
 function validateElement(
     hostView: LView, element: RElement, tNode: TNode, hasDirectives: boolean): void {
+  const schemas = hostView[TVIEW].schemas;
+
+  // If `schemas` is set to `null`, that's an indication that this Component was compiled in AOT
+  // mode where this check happens at compile time. In JIT mode, `schemas` is always present and
+  // defined as an array (as an empty array in case `schemas` field is not defined) and we should
+  // execute the check below.
+  if (schemas === null) return;
```
petebacondarwin: `null` and not `undefined`?

petebacondarwin (approved):

packages/core/src/render3/instructions/element.ts
AndrewKushnir: If it's not present, it's set to `null` in [defineComponent](https://github.com/angular/angular/blob/master/packages/core/src/render3/definition.ts#L293).

packages/core/src/render3/jit/module.ts
AndrewKushnir: @crisbeto thanks for review! I agree that this is not ideal, but having extra flag that we need to carry over to compiler to satisfy this JIT-only use-case might not worth it. We also perform some "patching" of Component/Directive defs in JIT mode in other places, so that seemed acceptable to me.

My proposal is to see whether that creates issues and/or we have similar use-cases where we need to do something similar, and if that's the case, perform additional refactoring. What do you think?

crisbeto: It's mostly a nitpick, I don't feel strongly about it.

crisbeto (approved):
LGTM

packages/core/src/render3/jit/module.ts
AndrewKushnir: Thanks for the reply @crisbeto!

AndrewKushnir:
- [VE Presubmit](http://test/OCL:282442899:BASE:282442941:1574723581519:590f773a)
- [Ivy Presubmit](https://test.corp.google.com/OCL:282442899:BASE:282443148:1574723640468:a9497c56)

IgorMinar (approved):

PR #34026, opened 2019-11-25
fix(ivy): run pre-order hooks in injection order

kara:
This commit fixes a compatibility bug where pre-order lifecycle
hooks (onInit, doCheck, OnChanges) for directives on the same
host node were executed based on the order the directives were
matched, rather than the order the directives were instantiated
(i.e. injection order).

This discrepancy can cause issues with forms, where it is common
to inject NgControl and try to extract its control property in
ngOnInit. As the NgControl directive is injected, it should be
instantiated before the control value accessor directive (and
thus its hooks should run first). This ensures that the NgControl
ngOnInit can set up the form control before the ngOnInit
for the control value accessor tries to access it.

### DIFF
--- a/integration/_payload-limits.json
+++ b/integration/_payload-limits.json
@@ -12,7 +12,7 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 1485,
-        "main-es2015": 15267,
+        "main-es2015": 15783,
         "polyfills-es2015": 36808
       }
     }
--- a/packages/core/src/render3/di.ts
+++ b/packages/core/src/render3/di.ts
@@ -17,6 +17,7 @@ import {assertDefined, assertEqual} from '../util/assert';
 
 import {getFactoryDef} from './definition';
 import {NG_ELEMENT_ID, NG_FACTORY_DEF} from './fields';
+import {registerPreOrderHooks} from './hooks';
 import {DirectiveDef, FactoryFn} from './interfaces/definition';
 import {NO_PARENT_INJECTOR, NodeInjectorFactory, PARENT_INJECTOR, RelativeInjectorLocation, RelativeInjectorLocationFlags, TNODE, isFactory} from './interfaces/injector';
 import {AttributeMarker, TContainerNode, TDirectiveHostNode, TElementContainerNode, TElementNode, TNode, TNodeProviderIndexes, TNodeType} from './interfaces/node';
@@ -519,7 +520,7 @@ export function locateDirectiveOrProvider<T>(
 }
 
 /**
-* Retrieve or instantiate the injectable from the `lData` at particular `index`.
+* Retrieve or instantiate the injectable from the `LView` at particular `index`.
 *
 * This function checks to see if the value has already been instantiated and if so returns the
 * cached `injectable`. Otherwise if it detects that the value is still a factory it
@@ -542,6 +543,16 @@ export function getNodeInjectable(
     enterDI(lView, tNode);
     try {
       value = lView[index] = factory.factory(undefined, tData, lView, tNode);
+      const tView = lView[TVIEW];
+      // This code path is hit for both directives and providers.
+      // For perf reasons, we want to avoid searching for hooks on providers.
+      // It does no harm to try (the hooks just won't exist), but the extra
+      // checks are unnecessary and this is a hot path. So we check to see
+      // if the index of the dependency is in the directive range for this
+      // tNode. If it's not, we know it's a provider and skip hook registration.
+      if (tView.firstCreatePass && index >= tNode.directiveStart) {
+        registerPreOrderHooks(index, tData[index] as DirectiveDef<any>, tView);
+      }
     } finally {
       if (factory.injectImpl) setInjectImplementation(previousInjectImplementation);
       setIncludeViewProviders(previousIncludeViewProviders);
--- a/packages/core/src/render3/di.ts
+++ b/packages/core/src/render3/di.ts
@@ -15,6 +15,7 @@ import {InjectFlags} from '../di/interface/injector';
 import {Type} from '../interface/type';
 import {assertDefined, assertEqual} from '../util/assert';
 
+import {assertDirectiveDef} from './assert';
 import {getFactoryDef} from './definition';
 import {NG_ELEMENT_ID, NG_FACTORY_DEF} from './fields';
 import {registerPreOrderHooks} from './hooks';
@@ -472,7 +473,7 @@ function searchTokensOnInjector<T>(
   const injectableIdx = locateDirectiveOrProvider(
       tNode, currentTView, token, canAccessViewProviders, isHostSpecialCase);
   if (injectableIdx !== null) {
-    return getNodeInjectable(currentTView.data, lView, injectableIdx, tNode as TElementNode);
+    return getNodeInjectable(lView, currentTView, injectableIdx, tNode as TElementNode);
   } else {
     return NOT_FOUND;
   }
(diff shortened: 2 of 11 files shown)


integration/_payload-limits.json
```diff
@@ -12,7 +12,7 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 1485,
-        "main-es2015": 15267,
+        "main-es2015": 15783,
```
kara: This is likely due to the fact that pre-order hook registration is no longer tree-shakable using `LifecycleHookFeature` (which is what the ivy-minimal example uses). This will have no impact on apps using normal bootstrap.

packages/forms/test/value_accessor_integration_spec.ts
```diff
@@ -1359,7 +1369,11 @@ export class MyInput implements ControlValueAccessor {
   // TODO(issue/24571): remove '!'.
   value !: string;
 
-  constructor(cd: NgControl) { cd.valueAccessor = this; }
+  control: AbstractControl|null = null;
+
+  constructor(public cd: NgControl) { cd.valueAccessor = this; }
```
pkozlowski-opensource: nit: I know that this is the existing code but the `cd` as a name is "cryptic" to me. Can you think of a different name (sorry for not having any better proposal but I'm not sure what `cd` supposed to signify, initially).

packages/core/test/acceptance/lifecycle_spec.ts
```diff
@@ -1704,6 +1825,50 @@ describe('doCheck', () => {
     expect(doChecks).toEqual(['app', 'comp 1', 'dir 1', 'comp 2', 'dir 2']);
   });
 
+  it('should be called on directives before component if component injects directives', () => {
```
pkozlowski-opensource: thinking out loud: do you think it would make sense to another test(s) for 2 directives, ex.: `<div dirNeedsDir1 dir1>` just to assert that orders depends on injection order between directives (without component being involved)? Just to make sure that we don't have any "special" processing for components (unless we've got such tests already).

packages/core/src/render3/di.ts
```diff
@@ -542,6 +543,16 @@ export function getNodeInjectable(
     enterDI(lView, tNode);
     try {
       value = lView[index] = factory.factory(undefined, tData, lView, tNode);
+      const tView = lView[TVIEW];
```
pkozlowski-opensource: nit: what do you think of passing `TView` now as an argument and getting `TData` from there? Generally speaking I wonder if / how we could avoid memory reads here and pass `TView` / `TData` / `firstCreatePass` as an argument. Again, this is very nit-y and I don't want it to result in more complex code, but yeh, if we could limit reads for each injectable, this would be great

packages/core/src/render3/di.ts
```diff
@@ -542,6 +543,16 @@ export function getNodeInjectable(
...
+      // This code path is hit for both directives and providers.
+      // For perf reasons, we want to avoid searching for hooks on providers.
+      // It does no harm to try (the hooks just won't exist), but the extra
+      // checks are unnecessary and this is a hot path. So we check to see
+      // if the index of the dependency is in the directive range for this
+      // tNode. If it's not, we know it's a provider and skip hook registration.
+      if (tView.firstCreatePass && index >= tNode.directiveStart) {
+        registerPreOrderHooks(index, tData[index] as DirectiveDef<any>, tView);
```
pkozlowski-opensource: Thnx for leaving such a detailed comment! 
nit: what do you think about adding a `ngDevMode` check to make sure that what we get from `tData[index]` is actually a `DirectiveDef` - just to make sure that the cast we are doing here is "safe"?

packages/core/src/render3/instructions/shared.ts
```diff
@@ -1097,26 +1097,33 @@ export function resolveDirectives(
...
-      registerPreOrderHooks(
-          directiveDefIdx, def, tView, nodeIndex, initialPreOrderHooksLength,
-          initialPreOrderCheckHooksLength);
+      // Only push a node index into the preOrderHooks array if this is the first
+      // pre-order hook found on this node.
+      if (!preOrderHooksFound && (def.onChanges || def.onInit || def.doCheck)) {
+        // We will push the actual hook function into this array later during dir instantiation.
+        // We cannot do it now because we must ensure hooks are registered in the same
+        // order that directives are created (i.e. injection order).
+        (tView.preOrderHooks || (tView.preOrderHooks = [])).push(tNode.index - HEADER_OFFSET);
```
pkozlowski-opensource: test for `preOrderHooks !== null`?

packages/core/src/render3/instructions/shared.ts
```diff
@@ -1097,26 +1097,33 @@ export function resolveDirectives(
...
+        preOrderHooksFound = true;
+      }
+
+      if (!preOrderCheckHooksFound && (def.onChanges || def.doCheck)) {
+        (tView.preOrderCheckHooks || (tView.preOrderCheckHooks = [
```
pkozlowski-opensource: test for `preOrderCheckHooksFound !== null`?

pkozlowski-opensource (approved):
LGTM. I've left some super-nit comments that you can safely ignore.
What would be great is to have some idea of the perf impact (if any).
How about running benchmarks and getting numbers before / after?
The ones that could be impacted are:
* micro: `yarn bazel run --define=compile=aot //packages/core/test/render3/perf:directive_instantiate` (for this we will need https://github.com/angular/angular/pull/34031 as this benchmark is broken atm...)
* benchpress: `yarn bazel run modules/benchmarks/src/tree/ng2:perf --define=compile=aot`

atscott:
[presubmit](https://test.corp.google.com/ui#id=OCL:282344716:BASE:282344731:1574691980222:2fda84eb)
[ivy presubmit](https://test.corp.google.com/OCL:282344716:BASE:282344794:1574692023431:3cec68a0)
[ivy presubmit with large project](https://test.corp.google.com/OCL:282344716:BASE:282345051:1574692156243:f543ce6f)

packages/core/test/acceptance/lifecycle_spec.ts
kara: That's a good call, I'll add one 👍

packages/core/src/render3/di.ts
kara: We can do a sanity check by looking at the shape of the object, sure (can't do an instanceof check because `DirectiveDef` is an interface)

packages/core/src/render3/instructions/shared.ts
kara: This is a reference to `tView.preOrderHooks` (not a truthiness check), so it can't be replaced by `preOrderHooks !== null`. If we did, we would be trying to push the hook into a boolean

kara:
@pkozlowski-opensource Thanks for the detailed review! Addressed your comments

packages/core/src/render3/instructions/shared.ts
pkozlowski-opensource: Right, right, what I _wanted_ to say is re-writing it with an `if`, but yeh, probably doesn't matter, let's lave it as-is unless something comes up in the benchmarks.

pkozlowski-opensource:
Thnx @kara LGTM (already approved)

packages/core/src/render3/instructions/shared.ts
kara: Ohhh sorry, totally misread your comment. I can change it now.

kara:
@pkozlowski-opensource As discussed, here are the "best times" for before and after:

*Microbenchmark*
Before: 1.806 (worst: 1.871)
After: 1.774  (worst: 1.914)

There was some variance, but I think we've determined we haven't regressed. It's about the same.

kara:
merge-assistance: i'm a code owner for size changes

PR #34043, opened 2019-11-26
build: update payload limits for `cli-hello-world-ivy-i18n` es2015 bundle

AndrewKushnir:
Commit that removes wtf* apis (https://github.com/angular/angular/commit/ed55355363dad26778dfe61fe6fc0ffc521f0db6) decreased es2015 bundle for `cli-hello-world-ivy-i18n` app (was: 138032, actual: 137209). This commit updates the `_payload-limits.json` file to reflect that.

## PR Type
What kind of change does this PR introduce?

- [x] Build related changes

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

PR #34044, opened 2019-11-26
ci: fix code ownership for size tracking

kara:
It should be set up so either Igor OR Kara can approve
size tracking rather than both.

See https://github.com/angular/angular/pull/34043 for example


IgorMinar (approved):

IgorMinar:
the ci is failing due to payload size issues?

IgorMinar:
the PR maybe needs a rebase?

.github/CODEOWNERS
```diff
@@ -347,6 +347,12 @@
 #   - IgorMinar
 #   - josephperrott
 
+# ===========================================================
+#  @angular/fw-size-tracking
+# ===========================================================
+#
+#   - kara
+#   - IgorMinar
```
gkalpak: We use alphabetic order in all other teams 😇

gkalpak:
I pushed a fixup commit (to restore alphabetic order and consistent linebreaks 😇) and I also rebased on master, while I was at it.

kara:
@gkalpak Thanks for fixing it up!!

PR #34052, opened 2019-11-26
fix(ivy): support inserting a `viewRef` that is already present

### LINKED ISSUE #33924: ngx-formly fails at runtime with: ASSERTION ERROR: index
# 🐞 bug report

### Affected Package

### Is this a regression?

Yes this worked seemlessly with ViewEngine under Angular 8.

### Description

When using [ngx-formly](https://github.com/ngx-formly/ngx-formly) with Ivy the app breaks. The **compilation succeeds without issues** but at **runtime** immediately when opening the app, I get the exception

```
ERROR Error: ASSERTION ERROR: index
    at throwError (core.js:1222)
    at assertLessThan (core.js:1196)
    at ViewContainerRef_._adjustIndex (core.js:17181)
    at ViewContainerRef_.insert (core.js:17104)
    at ViewContainerRef_.move (core.js:17131)
    at ViewContainerRef_.insert (core.js:17108)
    at FormlyField.renderField (ngx-formly-core.js:1165)
    at ngx-formly-core.js:1155
    at FormlyWrapperFormField.set [as fieldComponent] (ngx-formly-core.js:261)
    at FieldWrapper_Query (ngx-formly-core.js:1794)
```
which starts [here](https://github.com/ngx-formly/ngx-formly/blob/v5/src/core/src/lib/components/formly.field.ts#L111) in the Formly library.

## 🔬 Minimal Reproduction

Pointing to a GitHub repo as it strangely "works" (?) on Stackblitz.

1. `git clone git@github.com:juristr/superforms-talk.git`
1. `git checkout upgrade-ng9`
1. `yarn install` (or `npm install`)
1. Go to http://localhost:4200/formly-basics/person-simple

The example is broken as the Material components don't animate properly and the devtools console gives the `ASSERTION ERROR: index` as mentioned above

## 🔥 Exception or Error
<pre><code>

ERROR Error: ASSERTION ERROR: index
    at throwError (core.js:1222)
    at assertLessThan (core.js:1196)
    at ViewContainerRef_._adjustIndex (core.js:17181)
    at ViewContainerRef_.insert (core.js:17104)
    at ViewContainerRef_.move (core.js:17131)
    at ViewContainerRef_.insert (core.js:17108)
    at FormlyField.renderField (ngx-formly-core.js:1165)
    at ngx-formly-core.js:1155
    at FormlyWrapperFormField.set [as fieldComponent] (ngx-formly-core.js:261)
    at FieldWrapper_Query (ngx-formly-core.js:1794)
</code></pre>

## 🌍  Your Environment

**Angular Version:**
<pre><code>

Angular CLI: 9.0.0-rc.2
Node: 11.6.0
OS: darwin x64
Angular: 9.0.0-rc.2
... forms, language-service, material, platform-browser

@angular-devkit/architect         0.900.0-rc.2
@angular-devkit/build-angular     0.900.0-rc.2
@angular-devkit/build-optimizer   0.900.0-rc.2
@angular-devkit/build-webpack     0.900.0-rc.2
@angular-devkit/core              0.6.8
@angular-devkit/schematics        0.6.8
@ngtools/webpack                  9.0.0-rc.2
@schematics/angular               0.6.8
@schematics/update                0.900.0-rc.2

</code></pre>

**Anything else relevant?**

Reporting it here, but could also be related to some ngx-formly implementation issue

petebacondarwin:
When inserting a `viewRef` it is possible to not provide
an `index`, which is regarded as appending to the end of
the container.

If the `viewRef` already exists in the container, then
this results in a move. But there was a fault in the logic
that computed where to insert the `viewRef` that did not
account for the fact that the `viewRef` was already in
the container, so the insertion `index` was outside the
bounds of the array.

Fixes #33924

### DIFF
--- a/packages/core/src/render3/view_engine_compatibility.ts
+++ b/packages/core/src/render3/view_engine_compatibility.ts
@@ -245,14 +245,16 @@ export function createContainerRef(
         }
         this.allocateContainerIfNeeded();
         const lView = (viewRef as ViewRef<any>)._lView !;
-        const adjustedIdx = this._adjustIndex(index);
 
         if (viewAttachedToContainer(lView)) {
           // If view is already attached, fall back to move() so we clean up
           // references appropriately.
-          return this.move(viewRef, adjustedIdx);
+          // Note that we "shift" -1 because the move will involve inserting
+          // one view but also removing one view.
+          return this.move(viewRef, this._adjustIndex(index, -1));
         }
 
+        const adjustedIdx = this._adjustIndex(index);
         insertView(lView, this._lContainer, adjustedIdx);
 
         const beforeNode = getBeforeNodeForView(adjustedIdx, this._lContainer);
--- a/packages/core/src/render3/view_engine_compatibility.ts
+++ b/packages/core/src/render3/view_engine_compatibility.ts
@@ -271,8 +271,12 @@ export function createContainerRef(
           throw new Error('Cannot move a destroyed View in a ViewContainer!');
         }
         const index = this.indexOf(viewRef);
-        if (index !== -1) this.detach(index);
-        this.insert(viewRef, newIndex);
+        if (index === -1) {
+          this.insert(viewRef, newIndex);
+        } else if (index !== newIndex) {
+          this.detach(index);
+          this.insert(viewRef, newIndex);
+        }
         return viewRef;
       }
 
--- a/packages/core/test/acceptance/view_container_ref_spec.ts
+++ b/packages/core/test/acceptance/view_container_ref_spec.ts
@@ -204,6 +204,22 @@ describe('ViewContainerRef', () => {
         expect(fixture.nativeElement.textContent).toEqual('102');
       }
     });
+
+    it('should do nothing when a view is re-inserted / moved at the same index', () => {
+      const fixture = TestBed.createComponent(ViewContainerRefApp);
+      fixture.detectChanges();
+
+      const templates = fixture.componentInstance.vcrComp.templates.toArray();
+      const viewContainerRef = fixture.componentInstance.vcrComp.vcr;
+
+      const ref0 = viewContainerRef.createEmbeddedView(templates[0]);
+
+      expect(fixture.nativeElement.textContent).toEqual('0');
+
+      // insert again at the same place but without specifying any index
+      viewContainerRef.insert(ref0);
+      expect(fixture.nativeElement.textContent).toEqual('0');
+    });
   });
 
   describe('move', () => {


packages/core/test/render3/view_container_ref_spec.ts
```diff
@@ -470,6 +472,50 @@ describe('ViewContainerRef', () => {
       expect(dynamicComp.foo.first.nativeElement).toEqual(fooEl as any);
     });
 
+    it('should support re-inserting components into view queries for dynamically created components',
```
pkozlowski-opensource: @petebacondarwin at this point we are trying to avoid hand-written "generated" code so it should be a regular acceptance test. Plus, I think that we can have it simplified when putting alongside existing acceptance tests (there are supporting components / directives in place).

I've created a draft of such test here: https://github.com/angular/angular/pull/34054/files#diff-ac402b07939713de4f9b2a6ff5615059R208

packages/core/src/render3/view_engine_compatibility.ts
```diff
@@ -245,14 +245,16 @@ export function createContainerRef(
...
         this.allocateContainerIfNeeded();
         const lView = (viewRef as ViewRef<any>)._lView !;
-        const adjustedIdx = this._adjustIndex(index);
 
         if (viewAttachedToContainer(lView)) {
           // If view is already attached, fall back to move() so we clean up
           // references appropriately.
-          return this.move(viewRef, adjustedIdx);
+          // Note that we "shift" -1 because the move will involve inserting
+          // one view but also removing one view.
```
pkozlowski-opensource: Hmmm, this is interesting. Do you think that we could avoid doing any processing in the case where we "move" a view to the _same_ index? How about https://github.com/angular/angular/pull/34054/files#diff-f0a30efe7a958f4c73731e33aa32ebfaR272-R283 instead?

pkozlowski-opensource (changes requested):
@petebacondarwin ha! Thnx for looking into it, I was wondering what was going on when I saw the error / issue on GH! 

I left 2 comments in the PR, but when I was writing those comments I wanted to verify things and ended up doing another test + fix, see https://github.com/angular/angular/pull/34054...

Totally up to you to pick it up, change this PR or do whatever is right :-) In any case we want to avoid hand-written "generated" code.

packages/core/src/render3/view_engine_compatibility.ts
petebacondarwin: I considered something similar to your suggestion but that would add yet more code to the bundle size...

I wasn't sure what the best trade off is. I guess the real question is how often would this happen. If we have change detection setup that constantly tries to insert a view into a container where it already exists (such as in ngx-formly), then this would indeed happen a lot.

BTW `index === undefined ? this.length - 1 : index)` is identical to `this._adjustIndex(index, -1)`

pkozlowski-opensource: > BTW index === undefined ? this.length - 1 : index) is identical to this._adjustIndex(index, -1)

Right! 

> I considered something similar to your suggestion but that would add yet more code to the bundle size...

Hmm, as far as I can tell the total amount of added code would amount to `if (a !== b)` (after minification) so I would consider it as "worth it". Not married to the idea, though! 

Pushed an update to https://github.com/angular/angular/pull/34054

petebacondarwin:
I took your perf and test suggestions. Thanks @pkozlowski-opensource.

pkozlowski-opensource (approved):
LGTM

AndrewKushnir:
- [VE Presubmit](https://test.corp.google.com/ui#id=OCL:282580612:BASE:282580632:1574789095121:b3f17fa8)
- [Ivy Presubmit](https://test.corp.google.com/OCL:282580612:BASE:282580757:1574789142654:831d9a00)

PR #34054, opened 2019-11-26
fix(ivy): do nothing when re-inserting a view at the same index

pkozlowski-opensource:
Fixes #34052


pkozlowski-opensource:
Closing as #34052 is incorporating those changes.

PR #34057, opened 2019-11-26
ci: do not run ng_benchmarks on CI

IgorMinar:
We don't collect/review the results from CI, so it's just a waste of CPU time.

It's sufficient if we are able to run these locally and manually.


mhevery (approved):

josephperrott (approved):
LGTM

PR #34058, opened 2019-11-26
build: remove obsolete build related tools and helper scripts

IgorMinar:
none of these files are needed any more as they were replaced by Bazel.


josephperrott (approved):
LGTM

tools/gulp-tasks/serve.js
```diff
@@ -1,26 +0,0 @@
...
- * Use of this source code is governed by an MIT-style license that can be
- * found in the LICENSE file at https://angular.io/license
- */
-
-module.exports = {
-
-  // Serve the built files
-  default: (gulp) => () => {
-    const connect = require('gulp-connect');
-    const cors = require('cors');
```
devversion: I think we can remove both `gulp-connect` and `cors` from the `package.json`.

devversion (approved):

IgorMinar:
> I think we can remove both `gulp-connect` and `cors` from the `package.json`. 

good catch! thanks

PR #34060, opened 2019-11-26
fix(common): HttpHeaders lazyInit type check

### LINKED ISSUE #34059: HttpHeaders lazyInit throws error with non-string value
# 🐞 bug report

### Affected Package

 The issue is caused by package @angular/common/http

### Is this a regression?

No

### Description

Initializing HttpHeaders with an object of key/value pairs where a value is set to null causes an error.

## 🔬 Minimal Reproduction

https://stackblitz.com/edit/angular-issue-repro2-vmkwd7

This code in particular

<pre>
<code>
return this.httpClient.get('URL GOES HERE', {
  headers: new HttpHeaders({
    'X-Key-1': '1',
    'X-Key-2': null,
  }),
});
</code>
</pre>

## 🔥 Exception or Error
<pre>
<code>
ERROR TypeError: "values is null"
    common common-http.umd.js:175
    common common-http.umd.js:169
    common common-http.umd.js:233
    common common-http.umd.js:302
    common common-http.umd.js:1496
    rxjs Observable.js:44
    rxjs Observable.js:30
    rxjs subscribeTo.js:22
    rxjs subscribeToResult.js:10
    rxjs mergeMap.js:82
    rxjs mergeMap.js:76
    rxjs mergeMap.js:59
    rxjs Subscriber.js:67
    rxjs scalar.js:6
    rxjs Observable.js:44
    rxjs Observable.js:30
    rxjs mergeMap.js:39
    rxjs Observable.js:25
    rxjs filter.js:29
    rxjs Observable.js:25
    rxjs map.js:32
    rxjs Observable.js:25
    ngOnInit app.component.ts:20
    core core.umd.js:9274
    core core.umd.js:10538
    core core.umd.js:10500
    core core.umd.js:11133
    core core.umd.js:11093
    View_AppComponent_Host_0 (index):1
    core core.umd.js:11085
    core core.umd.js:10482
    core core.umd.js:11375
    core core.umd.js:11053
    core core.umd.js:8869
    core core.umd.js:4641
    core core.umd.js:4641
    core core.umd.js:4675
    core core.umd.js:4617
    core core.umd.js:4420
    core core.umd.js:4420
    core core.umd.js:4386
    js zone.js:388
    core core.umd.js:3890
    js zone.js:387
    js zone.js:138
    js zone.js:872
    js zone.js:421
    core core.umd.js:3881
    js zone.js:420
    js zone.js:188
    js zone.js:595
preview-0578bbed66dd68addeaab.js:1:90093
</code>
</pre>

## 🌍  Your Environment

**Angular Version:**
<pre><code>
Your global Angular CLI version (8.3.14) is greater than your local
version (7.3.8). The local Angular CLI version is used.

To disable this warning use "ng config -g cli.warnings.versionMismatch false".

                |___/

Angular CLI: 7.3.8
Node: 10.16.3
OS: win32 x64
Angular: 7.2.6

@angular-devkit/architect         0.13.8
@angular-devkit/build-angular     0.13.8
@angular-devkit/build-optimizer   0.13.8
@angular-devkit/build-webpack     0.13.8
@angular-devkit/core              7.3.8
@angular-devkit/schematics        7.3.8
@angular/cdk                      7.3.3
@angular/cli                      7.3.8
@angular/flex-layout              7.0.0-beta.23
@angular/material                 7.3.3
@ngtools/webpack                  7.3.8
@schematics/angular               7.3.8
@schematics/update                0.13.8
typescript                        3.2.2
webpack                           4.29.0
</code></pre>

**Anything else relevant?**

Although I am using an older version of Angular (current work version), this problem is still relevent in the current version of Angular.

cooneyasendia:
Add an undefined and null check to the possible value(s) when running the lazyInit of HttpHeaders with a key/value pair object.

fixes #34059

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #34059

## What is the new behavior?

No error thrown when attempting to initialize HttpHeaders with a null value. Instead, the key/value is simply not added.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/common/http/src/headers.ts
+++ b/packages/common/http/src/headers.ts
@@ -75,7 +75,7 @@ export class HttpHeaders {
           if (typeof values === 'string') {
             values = [values];
           }
-          if (values !== undefined && values !== null && values.length > 0) {
+          if (values && values.length > 0) {
             this.headers.set(key, values);
             this.maybeSetNormalizedName(name, key);
           }
--- a/packages/common/http/src/headers.ts
+++ b/packages/common/http/src/headers.ts
@@ -75,7 +75,7 @@ export class HttpHeaders {
           if (typeof values === 'string') {
             values = [values];
           }
-          if (values.length > 0) {
+          if (values !== undefined && values !== null && values.length > 0) {
             this.headers.set(key, values);
             this.maybeSetNormalizedName(name, key);
           }
--- a/packages/common/http/test/headers_spec.ts
+++ b/packages/common/http/test/headers_spec.ts
@@ -47,6 +47,16 @@ import {HttpHeaders} from '@angular/common/http/src/headers';
 
         expect(headers.getAll('foo')).toEqual(['second']);
       });
+      
+      it('should initialize with null values', () => {
+        const headers = new HttpHeaders({
+          'X-Key-1': '1',
+          'X-Key-2': (null as any),
+        });
+
+        expect(headers.getAll('X-Key-1')).toEqual(['1']);
+        expect(headers.getAll('X-Key-2')).toEqual(null);
+      });
     });
 
     describe('.set()', () => {
--- a/packages/common/http/test/headers_spec.ts
+++ b/packages/common/http/test/headers_spec.ts
@@ -47,7 +47,7 @@ import {HttpHeaders} from '@angular/common/http/src/headers';
 
         expect(headers.getAll('foo')).toEqual(['second']);
       });
-      
+
       it('should initialize with null values', () => {
         const headers = new HttpHeaders({
           'X-Key-1': '1',


ajitsinghkaler:
@cooneyasendia what do you think it better if get we get an error rather than just skipping it

cooneyasendia:
> 
> 
> @cooneyasendia what do you think it better if get we get an error rather than just skipping it

@ajitsinghkaler  [Based on the code directly above this code](https://github.com/cooneyasendia/angular/blob/http-headers-lazy-init-fix/packages/common/http/src/headers.ts#L55), there are two options. If you create HttpHeaders with a string instead of a key/value pair object, it depends on if you include a colon or not. With a colon included, we would get a value of empty string. Without a colon included, the key/value pair is skipped. I chose the skipped route because I thought it was cleaner then sending an empty HTTP header value. Is there ever value in sending an HTTP header with an empty string value? Not sure what the right answer is but having the code hit a null reference exception seems wrong.

packages/common/http/src/headers.ts
```diff
@@ -75,7 +75,7 @@ export class HttpHeaders {
           if (typeof values === 'string') {
             values = [values];
           }
-          if (values.length > 0) {
+          if (values !== undefined && values !== null && values.length > 0) {
```
IgorMinar: how about:

```
if (values && values.length > 0) {
```

IgorMinar (changes requested):
can you please add a test as well to packages/common/http/test/headers_spec.ts ? thanks

packages/common/http/src/headers.ts
pkozlowski-opensource: Or better yet: `if (values != null && values.length > 0) {`?

packages/common/http/test/headers_spec.ts
```diff
@@ -47,6 +47,16 @@ import {HttpHeaders} from '@angular/common/http/src/headers';
 
         expect(headers.getAll('foo')).toEqual(['second']);
       });
+
+      it('should initialize with null values', () => {
+        const headers = new HttpHeaders({
+          'X-Key-1': '1',
+          'X-Key-2': (null as any),
```
IgorMinar: wait a minute, looking into this change more, it's surprising that you have to defeat the type-checker to get the runtime error.

the API signature of `HttpHeaders` doesn't allow for `null` or `undefined` to be passed into it. So we should either change the API signature (which seems wrong), or not do this runtime check. Having to cast the test value to `any` in a test is definitely a code smell that we need to address.

how do you get into situation where nulls or undefined are being passed into http headers?

IgorMinar (changes requested):

packages/common/http/test/headers_spec.ts
cooneyasendia: The issue is that the Angular codebase has strict null checking but when I create a project with the Angular CLI it does not have strict null checking by default. The codebase I was using to find this issue did not have strict null checking enabled (angular cli generated) and it allows the `HttpHeaders` values to be freely set to null.

I know the other solution is for me to enable strict null checking in my Angular codebase but if the Angular CLI is not going to enable this by default it seems troublesome still.

PR #31271, opened 2019-06-25
feat(ivy): error in ivy when inheriting a ctor from an undecorated base

alxhub:
Angular View Engine uses global knowledge to compile the following code:

```typescript
export class Base {
  constructor(private vcr: ViewContainerRef) {}
}

@Directive({...})
export class Dir extends Base {
  // constructor inherited from base
}
```

Here, `Dir` extends `Base` and inherits its constructor. To create a `Dir`
the arguments to this inherited constructor must be obtained via dependency
injection. View Engine is able to generate a correct factory for `Dir` to do
this because via metadata it knows the arguments of `Base`'s constructor,
even if `Base` is declared in a different library.

In Ivy, DI is entirely a runtime concept. Currently `Dir` is compiled with
an ngDirectiveDef field that delegates its factory to `getInheritedFactory`.
This looks for some kind of factory function on `Base`, which comes up
empty. This case looks identical to an inheritance chain with no
constructors, which works today in Ivy.

Both of these cases will now become an error in this commit. If a decorated
class inherits from an undecorated base class, a diagnostic is produced
informing the user of the need to either explicitly declare a constructor or
to decorate the base class.

PR #34061, opened 2019-11-26
feat(ivy): error in ivy when inheriting a ctor from an undecorated base

crisbeto:
**Note:** this is a resubmit of https://github.com/angular/angular/pull/31271. I took the code from the original PR, rebased it against master, fixed the unit test failures and addressed the PR feedback. Sorry @alxhub, but for some reason the ownership of the original commit was switched when I rebased it. My changes are isolated to the second `fixup` commit.

--------------------------

Angular View Engine uses global knowledge to compile the following code:

```typescript
export class Base {
  constructor(private vcr: ViewContainerRef) {}
}

@Directive({...})
export class Dir extends Base {
  // constructor inherited from base
}
```

Here, `Dir` extends `Base` and inherits its constructor. To create a `Dir`
the arguments to this inherited constructor must be obtained via dependency
injection. View Engine is able to generate a correct factory for `Dir` to do
this because via metadata it knows the arguments of `Base`'s constructor,
even if `Base` is declared in a different library.

In Ivy, DI is entirely a runtime concept. Currently `Dir` is compiled with
an ngDirectiveDef field that delegates its factory to `getInheritedFactory`.
This looks for some kind of factory function on `Base`, which comes up
empty. This case looks identical to an inheritance chain with no
constructors, which works today in Ivy.

Both of these cases will now become an error in this commit. If a decorated
class inherits from an undecorated base class, a diagnostic is produced
informing the user of the need to either explicitly declare a constructor or
to decorate the base class.

### DIFF
--- a/packages/compiler-cli/ngcc/test/migrations/undecorated_parent_migration_spec.ts
+++ b/packages/compiler-cli/ngcc/test/migrations/undecorated_parent_migration_spec.ts
@@ -204,38 +204,6 @@ runInEachFileSystem(() => {
          expect(decorator.import).toEqual({from: '@angular/core', name: 'Directive'});
          expect(decorator.args !.length).toEqual(0);
        });
-
-    it('should skip the base class if it is in a different package from the derived class', () => {
-      const BASE_FILENAME = _('/node_modules/other-package/index.js');
-      const {program, analysis, errors} = setUpAndAnalyzeProgram([
-        {
-          name: INDEX_FILENAME,
-          contents: `
-       import {Directive, ViewContainerRef} from '@angular/core';
-       import {BaseClass} from 'other-package';
-       export class DerivedClass extends BaseClass {
-       }
-       DerivedClass.decorators = [
-         { type: Directive, args: [{ selector: '[dir]' }] }
-       ];
-     `
-        },
-        {
-          name: BASE_FILENAME,
-          contents: `
-          export class BaseClass {
-            constructor(private vcr: ViewContainerRef) {}
-          }
-          BaseClass.ctorParameters = () => [
-            { type: ViewContainerRef, }
-          ];
-      `
-        }
-      ]);
-      expect(errors).toEqual([]);
-      const file = analysis.get(program.getSourceFile(BASE_FILENAME) !);
-      expect(file).toBeUndefined();
-    });
   });
 
   function setUpAndAnalyzeProgram(testFiles: TestFile[]) {
--- a/packages/compiler-cli/ngcc/test/migrations/undecorated_parent_migration_spec.ts
+++ b/packages/compiler-cli/ngcc/test/migrations/undecorated_parent_migration_spec.ts
@@ -204,6 +204,45 @@ runInEachFileSystem(() => {
          expect(decorator.import).toEqual({from: '@angular/core', name: 'Directive'});
          expect(decorator.args !.length).toEqual(0);
        });
+
+    it('should skip the base class if it is in a different package from the derived class', () => {
+      const BASE_FILENAME = _('/node_modules/other-package/index.js');
+      const {errors} = setUpAndAnalyzeProgram([
+        {
+          name: INDEX_FILENAME,
+          contents: `
+         import {Directive, ViewContainerRef} from '@angular/core';
+         import {BaseClass} from 'other-package';
+         export class DerivedClass extends BaseClass {
+         }
+         DerivedClass.decorators = [
+           { type: Directive, args: [{ selector: '[dir]' }] }
+         ];
+       `
+        },
+        {
+          name: BASE_FILENAME,
+          contents: `
+            export class BaseClass {
+              constructor(private vcr: ViewContainerRef) {}
+            }
+            BaseClass.ctorParameters = () => [
+              { type: ViewContainerRef, }
+            ];
+        `
+        }
+      ]);
+
+      expect(errors.length).toBe(1);
+      expect(errors[0].messageText)
+          .toBe(
+              `The directive DerivedClass inherits its constructor ` +
+              `from BaseClass, but the latter does not have an Angular ` +
+              `decorator of its own. Dependency injection will not be ` +
+              `able to resolve the parameters of BaseClass's ` +
+              `constructor. Either add a @Directive decorator to ` +
+              `BaseClass, or add an explicit constructor to DerivedClass.`);
+    });
   });
 
   function setUpAndAnalyzeProgram(testFiles: TestFile[]) {
--- a/packages/compiler-cli/src/ngtsc/annotations/src/diagnostics.ts
+++ b/packages/compiler-cli/src/ngtsc/annotations/src/diagnostics.ts
@@ -63,12 +63,7 @@ export function checkInheritanceOfDirective(
   let baseClass = readBaseClass(node, reflector, evaluator);
 
   while (baseClass) {
-    if (baseClass === 'dynamic') {
-      // return getInheritedUndecoratedCtorDiagnostic(node, baseClass, reader);
-      return null;
-    }
-
-    if (!isNamedClassDeclaration(baseClass.node)) {
+    if (baseClass === 'dynamic' || !isNamedClassDeclaration(baseClass.node)) {
       return null;
     }
 
@@ -112,32 +107,13 @@ function getConstructor(node: ts.ClassDeclaration): ts.ConstructorDeclaration|un
 }
 
 function getInheritedUndecoratedCtorDiagnostic(
-    node: ClassDeclaration & ts.ClassDeclaration, baseClass: Reference | 'dynamic',
-    reader: MetadataReader) {
+    node: ClassDeclaration & ts.ClassDeclaration, baseClass: Reference, reader: MetadataReader) {
   const subclassMeta = reader.getDirectiveMetadata(new Reference(node)) !;
   const dirOrComp = subclassMeta.isComponent ? 'Component' : 'Directive';
-  let baseClassName: string;
-
-  if (baseClass instanceof Reference) {
-    baseClassName = baseClass.debugName !;
-  } else {
-    // If we weren't able to resolve the base class, fall back to picking out the name
-    // for the diagnostic from the type that is part of the `extends` clause.
-    const extendsClause = node.heritageClauses &&
-        node.heritageClauses.find(clause => clause.token === ts.SyntaxKind.ExtendsKeyword);
-    const extendedType = extendsClause && extendsClause.types[0];
-
-    if (extendedType && ts.isIdentifier(extendedType.expression)) {
-      baseClassName = extendedType.expression.text;
-    } else {
-      // We should be guaranteed to have a name from the logic above, but just in case,
-      // use `<dynamic>` if something went wrong. This looks better than `undefined`.
-      baseClassName = '<dynamic>';
-    }
-  }
+  const baseClassName = baseClass.debugName;
 
   return makeDiagnostic(
-      ErrorCode.DIRECTIVE_INHERITS_UNDECORATED_CTOR, node,
+      ErrorCode.DIRECTIVE_INHERITS_UNDECORATED_CTOR, node.name,
       `The ${dirOrComp.toLowerCase()} ${node.name.text} inherits its constructor from ${baseClassName}, ` +
           `but the latter does not have an Angular decorator of its own. Dependency injection will not be able to ` +
           `resolve the parameters of ${baseClassName}'s constructor. Either add a @Directive decorator ` +
--- a/packages/compiler-cli/src/ngtsc/annotations/src/diagnostics.ts
+++ b/packages/compiler-cli/src/ngtsc/annotations/src/diagnostics.ts
@@ -84,7 +84,7 @@ export function checkInheritanceOfDirective(
     }
 
     const newParentClass = readBaseClass(baseClass.node, reflector, evaluator);
-    if (newParentClass) {
+    if (newParentClass !== null) {
       // If we found another parent class, keep going.
       baseClass = newParentClass;
     } else if (!baseClassConstructor) {
--- a/packages/compiler-cli/src/ngtsc/annotations/src/diagnostics.ts
+++ b/packages/compiler-cli/src/ngtsc/annotations/src/diagnostics.ts
@@ -1,0 +1,92 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import * as ts from 'typescript';
+
+import {ErrorCode, makeDiagnostic} from '../../diagnostics';
+import {Reference} from '../../imports';
+import {MetadataReader} from '../../metadata';
+import {PartialEvaluator} from '../../partial_evaluator';
+import {ClassDeclaration, isNamedClassDeclaration} from '../../reflection';
+import {LocalModuleScopeRegistry} from '../../scope';
+import {makeDuplicateDeclarationError} from './util';
+
+export function getDirectiveDiagnostics(
+    node: ClassDeclaration, reader: MetadataReader, evaluator: PartialEvaluator,
+    scopeRegistry: LocalModuleScopeRegistry): ts.Diagnostic[]|null {
+  let diagnostics: ts.Diagnostic[]|null = [];
+
+  const addDiagnostics = (more: ts.Diagnostic | ts.Diagnostic[] | null) => {
+    if (more === null) {
+      return;
+    } else if (diagnostics === null) {
+      diagnostics = Array.isArray(more) ? more : [more];
+    } else if (Array.isArray(more)) {
+      diagnostics.push(...more);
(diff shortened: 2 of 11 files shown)


packages/compiler-cli/ngcc/test/migrations/undecorated_parent_migration_spec.ts
```diff
@@ -204,38 +204,6 @@ runInEachFileSystem(() => {
          expect(decorator.import).toEqual({from: '@angular/core', name: 'Directive'});
          expect(decorator.args !.length).toEqual(0);
        });
-
-    it('should skip the base class if it is in a different package from the derived class', () => {
```
crisbeto: I deleted this test because it's being caught as a compile-time error now.

packages/compiler-cli/src/ngtsc/annotations/src/diagnostics.ts
```diff
@@ -48,36 +48,48 @@ export function checkInheritanceOfDirective(
...
   const subclassMeta = reader.getDirectiveMetadata(new Reference(node)) !;
-
   const dirOrComp = subclassMeta.isComponent ? 'Component' : 'Directive';
 
   return makeDiagnostic(
       ErrorCode.DIRECTIVE_INHERITS_UNDECORATED_CTOR, type,
       `The ${dirOrComp.toLowerCase()} ${node.name.text} inherits its constructor from ${baseClass.debugName}, ` +
           `but the latter does not have an Angular decorator of its own. Dependency injection will not be able to ` +
-          `resolve the parameters of ${baseClass.debugName}'s constructor. Either add an @Directive annotation ` +
+          `resolve the parameters of ${baseClass.debugName}'s constructor. Either add a @Directive decorator ` +
```
petebacondarwin: I would still write `an @Directive` - although I note mixed use throughout the project.

packages/compiler-cli/src/ngtsc/annotations/src/diagnostics.ts
```diff
@@ -48,36 +48,48 @@ export function checkInheritanceOfDirective(
...
   if (!(baseClass instanceof Reference) || !isNamedClassDeclaration(baseClass.node)) {
     return null;
   }
 
+  // If the base class doesn't have a constructor or the constructor
+  // doesn't have parameters, it won't be using DI so we can skip it.
+  const baseClassConstructor = getConstructor(baseClass.node);
+  if (baseClassConstructor === undefined || baseClassConstructor.parameters.length === 0) {
+    return null;
+  }
```
petebacondarwin: Should we not be checking further ancestors? The direct base class may have no constructor but its base class might...

```ts
@DIrective(...)
class Dir extends Parent {}

class Parent extends GrandParent {}

class GrandParent {
  constructor(someParam: any) {}
}
```

packages/compiler-cli/ngcc/test/migrations/undecorated_parent_migration_spec.ts
petebacondarwin: I can't remember exactly but I thought we were trying to avoid errors such as this in ngcc - i.e. it should be more tolerant than ngtsc? But perhaps this is not one of those cases?

packages/compiler-cli/src/ngtsc/annotations/src/diagnostics.ts
crisbeto: Good point. This is the logic that was in the original PR so I kept it more or less verbatim, assuming that there was a reason it was written that way.

packages/compiler-cli/ngcc/test/migrations/undecorated_parent_migration_spec.ts
crisbeto: I haven't been part of those discussions, but this is a bit of a weird case because the error contradicts the migration completely. Shouldn't the migrations be run before the code even gets to the compiler?

petebacondarwin: Migrations are run after the "analyze" phase but before "resolve" and "compile" phases.

petebacondarwin: I can see that this error is being identified in the "resolve" phase, so I would expect this migration to avoid the error from appearing??

packages/compiler-cli/src/ngtsc/annotations/src/diagnostics.ts
alxhub: @petebacondarwin is right, we should walk as far up the inheritance hierarchy as we can get.

If we find a decorated base class, we're good. If we find an undecorated class with a non-trivial constructor, that should fail. And we should add test cases for the above.

packages/compiler-cli/src/ngtsc/annotations/src/diagnostics.ts
alxhub: We didn't cover this case in my English classes...

alxhub (changes requested):

crisbeto:
Updated based on the feedback from @alxhub.

packages/compiler-cli/src/ngtsc/annotations/src/diagnostics.ts
```diff
@@ -0,0 +1,122 @@
...
+
+  return makeDiagnostic(
+      ErrorCode.DIRECTIVE_INHERITS_UNDECORATED_CTOR, node,
+      `The ${dirOrComp.toLowerCase()} ${node.name.text} inherits its constructor from ${baseClassName}, ` +
+          `but the latter does not have an Angular decorator of its own. Dependency injection will not be able to ` +
+          `resolve the parameters of ${baseClassName}'s constructor. Either add a @Directive decorator ` +
+          `to ${baseClassName}, or add an explicit constructor to ${node.name.text}.`);
+}
+
+function getParentClass(node: ts.ClassDeclaration, evaluator: PartialEvaluator): Reference|null {
```
alxhub: See `readBaseClass` in `ngtsc/annotations/src/util.ts`.

packages/compiler-cli/src/ngtsc/annotations/src/diagnostics.ts
```diff
@@ -0,0 +1,122 @@
...
+  // The extends clause is an expression which can be as dynamic as the user wants. Try to
+  // evaluate it, but fall back on ignoring the clause if it can't be understood. This is a View
+  // Engine compatibility hack: View Engine ignores 'extends' expressions that it cannot understand.
+  let parentClass = getParentClass(node, evaluator);
+
+  if (!parentClass) {
+    return null;
+  }
+
+  while (parentClass && isNamedClassDeclaration(parentClass.node)) {
```
alxhub:
```suggestion
  while (parentClass instanceof Reference && isNamedClassDeclaration(parentClass.node)) {
```

packages/compiler-cli/src/ngtsc/annotations/src/diagnostics.ts
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    const newParentClass = getParentClass(parentClass.node, evaluator);
+    if (newParentClass) {
+      // If we found another parent class, keep going.
+      parentClass = newParentClass;
+    } else if (!baseClassConstructor) {
+      // If this is the last class in the chain and it doesn't have
+      // a constructor, it can't be using DI so we shouldn't throw.
+      return null;
```
alxhub: We should distinguish between the cases where:

1) we reach the top class in the inheritance chain, and it has no constructor
2) we reach a dynamic point in the inheritance chain, and still have not found a constructor.

For case 2, I think we should error. `readBaseClass` will distinguish between no base class (returns `null`) or an unknown base class (returns `'dynamic'`).

packages/compiler-cli/src/ngtsc/diagnostics/src/code.ts
```diff
@@ -98,7 +98,13 @@ export enum ErrorCode {
...
    * An injectable already has a `ɵprov` property.
    */
-  INJECTABLE_DUPLICATE_PROV = 9001
+  INJECTABLE_DUPLICATE_PROV = 9001,
+
+  /**
+   * Raised when a Directive inherits its constructor from a base class without an Angular
+   * decorator.
+   */
+  DIRECTIVE_INHERITS_UNDECORATED_CTOR = 10001,
```
alxhub: Use the 2xxx error space above.

alxhub (changes requested):

packages/compiler-cli/src/ngtsc/annotations/src/diagnostics.ts
```diff
@@ -0,0 +1,114 @@
...
+    if (baseClassConstructor) {
+      if (baseClassConstructor.parameters.length === 0) {
+        return null;
+      } else {
+        return getInheritedUndecoratedCtorDiagnostic(node, baseClass, reader);
+      }
+    }
+
+    const newParentClass = readBaseClass(baseClass.node, reflector, evaluator);
+    if (newParentClass) {
```
alxhub: Explicit checks: `newParentClass !== null`.

packages/compiler-cli/ngcc/test/migrations/undecorated_parent_migration_spec.ts
alxhub: I think the scenario is still valid to test. Here's what I think should happen in the context of ngcc:

When compiling `BaseClass` in its separate package, there is zero indication that it's an Angular class, so no action is taken against it and no error is produced.

When compiling `DerivedClass`, then, ngcc will attempt to apply the migration for undecorated parents. It will follow `DerivedClass`' inheritance hierarchy, recognize that `BaseClass` would need to be migrated but cannot be (because it's in a separate package).

At this point, ngcc has a choice:
1) raise an error, because there's no hope this would work in production and no possibility of continuing.
2) silently ignore the problem and allow the crash to happen at runtime.

Currently we use door #2, but this PR I believe will convert us to door #1. That's fine (I think the error message will be better, for sure), but we should definitely have a test for this scenario. I would also really like to see us have an error message that makes sense in the context of ngcc ("Cannot generate a factory for 'DerivedClass' from 'test-package' as it inherits from the undecorated class 'BaseClass' from 'other-package'.").

alxhub (changes requested):

packages/compiler-cli/ngcc/test/migrations/undecorated_parent_migration_spec.ts
```diff
@@ -207,34 +207,41 @@ runInEachFileSystem(() => {
...
         }
       ]);
-      expect(errors).toEqual([]);
-      const file = analysis.get(program.getSourceFile(BASE_FILENAME) !);
-      expect(file).toBeUndefined();
+
+      expect(errors.length).toBe(1);
+      expect(errors[0].messageText)
+          .toBe(
+              `The directive DerivedClass inherits its constructor ` +
```
alxhub: I think this is okay for now, but we should revisit this later and figure out how to get ngcc to specialize the error in this case, to talk explicitly about the two packages involved and the exact situation in the context of ngcc's operations, rather than as if the user wrote both classes.

packages/compiler-cli/src/ngtsc/annotations/src/diagnostics.ts
```diff
@@ -0,0 +1,114 @@
...
+    } else {
+      // Otherwise log the diagnostic.
+      return getInheritedUndecoratedCtorDiagnostic(node, baseClass, reader);
+    }
+  }
+
+  return null;
+}
+
+/** Gets the constructor of a class declaration. */
```
alxhub: I think this should go through `ReflectionHost.getMembersOfClass` (or possibly `ReflectionHost.getConstructorParameters()` instead of the TS API, as there's no guarantee the class declaration is a `NamedClassDeclaration` as checked above.

packages/compiler-cli/src/ngtsc/annotations/src/diagnostics.ts
```diff
@@ -0,0 +1,114 @@
...
+      } else {
+        return getInheritedUndecoratedCtorDiagnostic(node, baseClass, reader);
+      }
+    }
+
+    const newParentClass = readBaseClass(baseClass.node, reflector, evaluator);
+    if (newParentClass !== null) {
+      // If we found another parent class, keep going.
+      baseClass = newParentClass;
+    } else if (!baseClassConstructor) {
```
alxhub: In the compiler, prefer explicit comparisons and don't rely on type coercion:

```suggestion
    } else if (baseClassConstructor === null) {
```

packages/compiler-cli/src/ngtsc/annotations/src/diagnostics.ts
```diff
@@ -0,0 +1,114 @@
...
+    // it's legal to inherit from any class.
+    return null;
+  }
+
+  // The extends clause is an expression which can be as dynamic as the user wants. Try to
+  // evaluate it, but fall back on ignoring the clause if it can't be understood. This is a View
+  // Engine compatibility hack: View Engine ignores 'extends' expressions that it cannot understand.
+  let baseClass = readBaseClass(node, reflector, evaluator);
+
+  while (baseClass) {
```
alxhub: Explicit checks.
```suggestion
  while (baseClass !== null) {
```

packages/compiler-cli/src/ngtsc/annotations/src/diagnostics.ts
```diff
@@ -0,0 +1,114 @@
...
+
+    // We can skip the base class if it has metadata.
+    const baseClassMeta = reader.getDirectiveMetadata(baseClass);
+    if (baseClassMeta !== null) {
+      return null;
+    }
+
+    // If the base class has a blank constructor we can skip it since it can't be using DI.
+    const baseClassConstructor = getConstructor(baseClass.node);
+    if (baseClassConstructor) {
```
alxhub: I think we could simplify this logic by reading both `baseClassConstructor` and `newParentClass` up front, and then:

```typescript
if (baseClassConstructor !== null && baseClassConstructor.parameters.length > 0) {
  // this class has a non-trivial constructor, that's an error!
  return error
} else if (baseClassConstructor !== null || newParentClass === null) {
  // this class has a trivial constructor, or no constructor + is the top of the inheritance chain, so it's okay
  return null;
}

// go up the chain and continue
baseClass = newParentClass;
```

packages/compiler-cli/src/ngtsc/metadata/src/registry.ts
```diff
@@ -30,6 +30,10 @@ export class LocalMetadataRegistry implements MetadataRegistry, MetadataReader {
     return this.pipes.has(ref.node) ? this.pipes.get(ref.node) ! : null;
   }
 
+  forEachDirective(fn: (dir: ClassDeclaration, meta: DirectiveMeta) => void): void {
```
alxhub: Seems unused?

packages/compiler-cli/test/ngtsc/ngtsc_spec.ts
```diff
@@ -5131,6 +5134,245 @@ export const Foo = Foo__PRE_R3__;
       });
     });
 
+    describe('inherited directives', () => {
```
alxhub: +1 nice test suite!

alxhub (changes requested):

packages/compiler-cli/src/ngtsc/annotations/src/diagnostics.ts
```diff
@@ -0,0 +1,98 @@
...
+    return null;
+  }
+
+  // The extends clause is an expression which can be as dynamic as the user wants. Try to
+  // evaluate it, but fall back on ignoring the clause if it can't be understood. This is a View
+  // Engine compatibility hack: View Engine ignores 'extends' expressions that it cannot understand.
+  let baseClass = readBaseClass(node, reflector, evaluator);
+
+  while (baseClass !== null) {
+    if (baseClass === 'dynamic' || !isNamedClassDeclaration(baseClass.node)) {
```
alxhub: I don't think you need `isNamedClassDeclaration` here. The rest of this code doesn't actually rely on having a `NamedClassDeclaration` (specifically, it doesn't need a `ts.ClassDeclaration`, and it won't be one in ngcc). Using `ReflectionHost` means that you can operate on a non-`class` declaration and still follow the inheritance chain.

alxhub (changes requested):
One more comment, but otherwise this is in great shape.

crisbeto:
Addressed the final comment.

alxhub:
[Presubmit](OCL:285808490:BASE:285808531:1576523276972:3e7065df)
[Ivy Presubmit](https://test.corp.google.com/ui#id=OCL:285808490:BASE:285808536:1576523276946:c7d4c229)

alxhub:
Updated:

[Presubmit](https://test.corp.google.com/ui#id=OCL:285808490:BASE:286081460:1576627730248:c9e42e0d)
[Ivy Presubmit](https://test.corp.google.com/ui#id=OCL:285808490:BASE:286081460:1576627730238:faff8fae)

alxhub:
Closing as this has been merged into #34460

PR #34065, opened 2019-11-26
fix(ivy): i18n - ensure that escaped chars are handled in localized s…

petebacondarwin:
…trings

When creating synthesized tagged template literals, one must provide both
the "cooked" text and the "raw" (unparsed) text. Previously there were no
good APIs for creating the AST nodes with raw text for such literals.
Recently the APIs were improved to support this, and they do an extra
check to ensure that the raw text parses to be equal to the cooked text.

It turns out there is a bug in this check -
see https://github.com/microsoft/TypeScript/issues/35374.

This commit works around the bug by synthesizing a "head" node and morphing
it by changing its `kind` into the required node type.

// FW-1747


AndrewKushnir (approved):
LGTM 👍 Thanks for the fix Pete!

AndrewKushnir:
- [VE Presubmit](http://test/OCL:282656451:BASE:282656483:1574812668571:6d588626)
- [Ivy Presubmit](https://test.corp.google.com/OCL:282656451:BASE:282656567:1574812709781:79e4688a)

PR #34074, opened 2019-11-27
ci: disable renovate for integration tests

IgorMinar:
the package.json in these tests points to file:dist/packages-dist which
is not materialized on renovate-bot, so the bot freaks out and fails to
update the lock files.

This means that the bot currently just opens up PRs that can't be merged.

Example: https://github.com/angular/angular/pull/34071

(In the past this worked nicely because we didn't keep the yarn.lock files
in the git repo for these projects, but that had a problem with unpined
depenedencies and undeterministic build/test runs).

For now it's better to disable the bot by removing all the package.jsons
from the config. We should revisit this in the future have the bot update
the root package.json instead.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -15,11 +15,7 @@
 		"source-map",
 		"@types/node"
 	],
-	"packageFiles": [
-		"integration/cli-hello-world-ivy-compat/package.json",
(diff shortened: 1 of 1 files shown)


josephperrott (approved):
LGTM

PR #34076, opened 2019-11-27
refactor(ivy): remove unnecessary fac wrapper

kara:
For injectables, we currently generate a factory function in the
injectable def (prov) that delegates to the factory function in
the factory def (fac). It looks something like this:

```js
factory: function(t) { return Svc.fac(t); }
```

The extra wrapper function is unnecessary since the args for
the factory functions are the same. This commit changes the
compiler to generate this instead:

```js
factory: Svc.fac
```

Because we are generating less code for each injectable, we
should see some modest code size savings. AIO's main bundle
is >1 KB smaller.

### DIFF
--- a/packages/compiler-cli/ngcc/src/analysis/decoration_analyzer.ts
+++ b/packages/compiler-cli/ngcc/src/analysis/decoration_analyzer.ts
@@ -94,6 +94,11 @@ export class DecorationAnalyzer {
     new DirectiveDecoratorHandler(
         this.reflectionHost, this.evaluator, this.fullRegistry, NOOP_DEFAULT_IMPORT_RECORDER,
         this.isCore, /* annotateForClosureCompiler */ false),
+    // Pipe handler must be before injectable handler in list so pipe factories are printed
+    // before injectable factories (so injectable factories can delegate to them)
+    new PipeDecoratorHandler(
+        this.reflectionHost, this.evaluator, this.metaRegistry, NOOP_DEFAULT_IMPORT_RECORDER,
+        this.isCore),
     new InjectableDecoratorHandler(
         this.reflectionHost, NOOP_DEFAULT_IMPORT_RECORDER, this.isCore,
         /* strictCtorDeps */ false, /* errorOnDuplicateProv */ false),
@@ -101,9 +106,6 @@ export class DecorationAnalyzer {
         this.reflectionHost, this.evaluator, this.fullMetaReader, this.fullRegistry,
         this.scopeRegistry, this.referencesRegistry, this.isCore, /* routeAnalyzer */ null,
         this.refEmitter, NOOP_DEFAULT_IMPORT_RECORDER, /* annotateForClosureCompiler */ false),
-    new PipeDecoratorHandler(
-        this.reflectionHost, this.evaluator, this.metaRegistry, NOOP_DEFAULT_IMPORT_RECORDER,
-        this.isCore),
   ];
   migrations: Migration[] = [
     new UndecoratedParentMigration(),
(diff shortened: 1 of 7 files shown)


IgorMinar (approved):
that turned out to be quite a simple change! thanks

packages/compiler/src/injectable_compiler_2.ts
```diff
@@ -120,8 +120,7 @@ export function compileInjectable(meta: R3InjectableMetadata): InjectableDef {
 function delegateToFactory(type: o.Expression) {
   return {
     statements: [],
-    // () => type.ɵfac(t)
-    factory: o.fn([new o.FnParam('t', o.DYNAMIC_TYPE)], [new o.ReturnStatement(type.callMethod(
-                                                            'ɵfac', [o.variable('t')]))])
+    // factory: type.ɵfac
+    factory: type.prop('ɵfac')
```
JoostK: Can this be a forward reference, in which case it would not yet be available if evaluated eagerly like this?

IgorMinar:
this will also need a rebase

packages/compiler/src/injectable_compiler_2.ts
kara: I don't think that's necessary. This factory is always generated next to the type referenced here, so it has to be defined.

JoostK: Actually I think this test is showing it is possible: https://github.com/angular/angular/pull/34076/files#diff-c2c3919962b7deb686e11f9115af9de0R291

kara: Ahh, I see what you mean. If the classes are different, then we will do what we did before with the wrapper. Adding a forward ref in all cases would defeat the purpose of the PR.

kara:
[VE presubmit](https://tap.corp.google.com/ui#id=OCL:282848321:BASE:282848339:1574899175458:8ce9448b)

[Ivy presubmit](https://test.corp.google.com/OCL:282848321:BASE:282848577:1574899274013:411416bc)

IgorMinar (approved):
looks good to me. I thought we were hoping for a slightly bigger impact, no? I'll take 1.2k for aio :)

IgorMinar:

crisbeto (approved):
For the record, the reason it was set up this way to begin with was that if the order of `Comp.ɵfac` and `Comp.ɵcmp` is reversed, the factory in `Comp.ɵcmp` will be `undefined` because the factory hasn't been defined yet.

packages/compiler-cli/ngcc/src/analysis/decoration_analyzer.ts
```diff
@@ -94,16 +94,18 @@ export class DecorationAnalyzer {
     new DirectiveDecoratorHandler(
         this.reflectionHost, this.evaluator, this.fullRegistry, NOOP_DEFAULT_IMPORT_RECORDER,
         this.isCore, /* annotateForClosureCompiler */ false),
+    // Pipe handler must be before injectable handler in list so pipe factories are printed
+    // before injectable factories (so injectable factories can delegate to them)
+    new PipeDecoratorHandler(
+        this.reflectionHost, this.evaluator, this.metaRegistry, NOOP_DEFAULT_IMPORT_RECORDER,
+        this.isCore),
```
petebacondarwin: We should have a test for this. Apart from anything to prevent an accidental regression (despite the comment).
I would suggest adding one here https://github.com/angular/angular/blob/2fb9b7ff1b2da43410a624e8090062a2d3bbe02d/packages/compiler-cli/ngcc/test/integration/ngcc_spec.ts

petebacondarwin (changes requested):

petebacondarwin:
> that turned out to be quite a simple change! thanks

@IgorMinar - I would say that is because @alxhub did a good job of architecting ngtsc :-)

packages/compiler-cli/ngcc/src/analysis/decoration_analyzer.ts
kara: I added a test for ngtsc, but not for ngcc. Good catch - will add.

petebacondarwin (approved):

kara:
merge-assistance: issue with `integration_test` failing on all PRs

PR #34085, opened 2019-11-27
fix(ngcc): do not output duplicate ɵprov properties

petebacondarwin:
Previously, the Angular AOT compiler would always add a
`ɵprov` to injectables. But in ngcc this resulted in duplicate `ɵprov`
properties since published libraries already have this property.

Now in ngtsc, trying to add a duplicate `ɵprov` property is an error,
while in ngcc the additional property is silently not added.

// FW-1750

### DIFF
--- a/packages/compiler-cli/ngcc/test/integration/ngcc_spec.ts
+++ b/packages/compiler-cli/ngcc/test/integration/ngcc_spec.ts
@@ -175,6 +175,36 @@ runInEachFileSystem(() => {
       expect(jsContents).toMatch(/\bvar _c0 =/);
     });
 
+    it('should add ɵfac but not duplicate ɵprov properties on injectables', () => {
+      compileIntoFlatEs5Package('test-package', {
+        '/index.ts': `
+        import {Injectable, ɵɵdefineInjectable} from '@angular/core';
+        export const TestClassToken = 'TestClassToken';
+        @Injectable({providedIn: 'module'})
+        export class TestClass {
+          static ɵprov = ɵɵdefineInjectable({ factory: () => {}, token: TestClassToken, providedIn: "module" });
+        }
+        `,
+      });
+
+      const before = fs.readFile(_(`/node_modules/test-package/index.js`));
+      const originalProp = /ɵprov[^;]+/.exec(before) ![0];
+      mainNgcc({
+        basePath: '/node_modules',
+        targetEntryPointPath: 'test-package',
+        propertiesToConsider: ['main'],
+      });
+      const after = fs.readFile(_(`/node_modules/test-package/index.js`));
+
+      expect(before).toContain(originalProp);
+      expect(countOccurrences(before, 'ɵprov')).toEqual(1);
+      expect(countOccurrences(before, 'ɵfac')).toEqual(0);
+
+      expect(after).toContain(originalProp);
+      expect(countOccurrences(after, 'ɵprov')).toEqual(1);
+      expect(countOccurrences(after, 'ɵfac')).toEqual(1);
+    });
+
     it('should add generic type for ModuleWithProviders and generate exports for private modules',
        () => {
          compileIntoApf('test-package', {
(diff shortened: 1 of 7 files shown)


packages/compiler-cli/src/ngtsc/diagnostics/src/code.ts
```diff
@@ -94,6 +94,11 @@ export enum ErrorCode {
    * No matching pipe was found for a
    */
   MISSING_PIPE = 8004,
+
+  /**
+   * An injectable already has a `ɵprov` property.
+   */
+  INJECTABLE_DUPLICAGE_PROV = 9001
```
IgorMinar: DUPLICATE_INJECTABLE_PROV?

packages/compiler-cli/ngcc/test/integration/ngcc_spec.ts
```diff
@@ -1231,3 +1261,12 @@ runInEachFileSystem(() => {
     }
   });
 });
+
+function countOccurrences(haystack: string, needle: string): number {
+  const regex = new RegExp(needle, 'g');
+  let count = 0;
+  while (regex.exec(haystack)) {
```
IgorMinar: How does this work? You are not mutating the haystack so how do you exit the loop?

IgorMinar (approved):
A few nits but overall lgtm. Thank you

packages/compiler-cli/src/ngtsc/diagnostics/src/code.ts
petebacondarwin: Apart from the typo... :-)

I was following the "sort of" convention in this file of putting the "scope" e.g. `INJECTABLE` at the start of the error code. For example:

```ts
  DECORATOR_ARG_NOT_LITERAL = 1001,
  DECORATOR_ARITY_WRONG = 1002,
  DECORATOR_NOT_CALLED = 1003,
  DECORATOR_ON_ANONYMOUS_CLASS = 1004,
  DECORATOR_UNEXPECTED = 1005,
  ...
  NGMODULE_INVALID_DECLARATION = 6001,
  NGMODULE_INVALID_IMPORT = 6002,
  NGMODULE_INVALID_EXPORT = 6003,
  ...
```

packages/compiler-cli/ngcc/test/integration/ngcc_spec.ts
petebacondarwin: Each time you execute the regex it starts from the previous position.

IgorMinar: it does?!?!? so it stores the state? how did I not know this! thanks

packages/compiler-cli/src/ngtsc/diagnostics/src/code.ts
IgorMinar: ok. sgtm

IgorMinar (approved):

IgorMinar:
[presubmit](https://test.corp.google.com/ui#id=OCL:282799081:BASE:282799123:1574880914278:85453b6b)
[ivy](https://test.corp.google.com/ui#id=OCL:282799081:BASE:282799197:1574880948184:687934)

packages/compiler-cli/ngcc/test/integration/ngcc_spec.ts
gkalpak: I am late to the party, but here are some fun facts:
- There is a (kind of) bug in `RegExp#exec()`, so that you can get stuck in an infinite loop when the RegExp accepts zero-length matches. E.g.
    ```js
    const re = /\d*/g;
    while (re.exec('1a2b')) console.log('Infinite loop \\o/');
    ```
    To avoid it, you need to manually do `re.lastIndex += 1` on zero-length matches.
- Other methods are not susceptible to this issue. So, using `haystack.match(needle)` would avoid this pitfall (and be sorter :wink:).

petebacondarwin: `String.match()` is what I was looking for!

PR #34089, opened 2019-11-27
fix(language-service): determine correct type for ngFor exported values

### LINKED ISSUE #460: index value is not number error in template
**Describe the bug**

Showing error for index value from ngFor is not numeric value.
```
(variable) i
Expected a numeric typeng(0)
```
**To Reproduce**

**Expected behavior**

**Logs**

Please attach two logs:

1. Console output (Go to toolbar --> View --> Output, and copy the contents)
2. Log file (Path is printed to the console output at startup)

**Screenshots**

image

**Additional context**

ayazhafiz:
Currently, variables of an unknown type in an `*ngFor` expression are
refined to have the type of the iterable binding of the `*ngFor`
expression. Unfortunately, this is a bug for variables aliasing
[values exported by `*ngFor`](https://angular.io/api/common/NgForOf#local-variables),
including `index` and `first`, because they are also given the type of
the binding expression, but they are not of the binding type. For
example, in

```typescript
@Component({
  selector: 'test',
  template: `
    <div *ngFor="let hero of heroes; let i = index; let isFirst = first">
      {{ hero }}
    </div>
  `
})
export class TestComponent {
  heroes: Hero[];
}
```

The local variables `i` and `isFirst` are determined to have a type of
`Hero`, when actually their types are `number` and `boolean`,
respectively.

This commit fixes this bug by checking if the value of a variable in an
`*ngFor` expression is known to be an export and assigning the variable
the type of that export value. Only if the variable does not alias an
export is it typed with the binding value of the `*ngFor` expression.

Closes https://github.com/angular/vscode-ng-language-service/issues/460

## PR Checklist

## PR Type

- [x] Bugfix

- [ ] Yes
- [x] No

### DIFF
--- a/packages/language-service/src/expression_diagnostics.ts
+++ b/packages/language-service/src/expression_diagnostics.ts
@@ -115,7 +115,7 @@ function getVarDeclarations(
           // that have been declared so far are also in scope.
           info.query.createSymbolTable(results),
         ]);
-        symbol = refinedVariableType(symbolsInScope, info.query, current);
+        symbol = refinedVariableType(variable.value, symbolsInScope, info.query, current);
       }
       results.push({
         name: variable.name,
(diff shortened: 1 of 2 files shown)


packages/language-service/src/expression_diagnostics.ts
```diff
@@ -127,16 +127,36 @@ function getVarDeclarations(
...
 }
 
+/**
+ * Gets the type of an ngFor exported value, as enumerated in
+ * https://angular.io/api/common/NgForOf#local-variables.
+ * @param value exported value name
+ * @param query type symbol query
+ */
+function getNgForExportedValueType(value: string, query: SymbolQuery): Symbol|undefined {
+  switch (value) {
```
kyliau: There're also $implicit and count.
Please see https://angular.io/api/common/NgForOfContext for the full list.
Playground example: https://stackblitz.com/edit/angular-rqjbmk

packages/language-service/src/expression_diagnostics.ts
```diff
@@ -127,16 +127,36 @@ function getVarDeclarations(
   return results;
 }
 
+/**
+ * Gets the type of an ngFor exported value, as enumerated in
+ * https://angular.io/api/common/NgForOf#local-variables.
```
kyliau: The list on that link is not complete, please use https://angular.io/api/common/NgForOfContext instead

kyliau (changes requested):

Airblader:
Why does the language service have special knowledge of ngFor at all? Shouldn't it infer proper types for all structural directives?

packages/language-service/src/expression_diagnostics.ts
ayazhafiz: Added `count`. `$implicit` has the type of the binding we are trying to deduce, so it's functionality is correct as-is.

packages/language-service/src/expression_diagnostics.ts
ayazhafiz: Added.

kyliau:
> Why does the language service have special knowledge of ngFor at all? Shouldn't it infer proper types for all structural directives?

For historical reasons, the language service does not share the same code path as the compiler for some of the checks. This is a problem that we are aware of, and we plan to fully integrate the Ivy compiler into the language service in version 10.

We are tracking this work in this issue https://github.com/angular/vscode-ng-language-service/issues/380

ayazhafiz:
> Why does the language service have special knowledge of ngFor at all?

To build on Keen's comment above, for this particular case, see https://github.com/angular/vscode-ng-language-service/issues/144. Some syntaxes of `ngFor` do not allow for immediately-evident variable types with the parsing pipeline the language service currently runs.

Airblader:
Thanks, I appreciate both of you taking the time to explain!

kyliau (approved):

Issue #15730, opened 2017-04-03
Platform Server - Attach cookies to HTTP requests


Reuvenc:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => search github for a similar issue or PR before submitting
[ ] feature request
[ ] support request => Please do not submit support request here, instead see https://github.com/angular/angular/blob/master/CONTRIBUTING.md#question
```

**Current behavior**

Currently I see no option to attach cookies to a http request. 
Using Universal + angular v2.x.x allowed setting 'Cookie' header property manually, in the current version I'm getting the following warning: 
`Refused to set unsafe header "Cookie"` ([node-xhr2](https://github.com/pwnall/node-xhr2) library logs this warning, as it disallows setting cookie header).

**Expected behavior**
It should be possible to pass cookies to a http request. 
[It was already asked in the roadmap](https://github.com/angular/angular/issues/13822#issuecomment-289734558) but no answer or suggestion was provided. 

**Minimal reproduction of the problem with instructions**

**What is the motivation / use case for changing the behavior?**

Without attaching cookies to a http request, the response on the server is different than the response on the client. It also makes the usage of transfer state dangerous, since the browser will not fire the http request and will use the response received from the server.

**Please tell us about your environment:**

* **Angular version:** 4.0.0

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

 
* **Language:**: TypeScript 2.2.2

kukjevov:
Hi this is a problem of `xhr2` module.

I dont have solution, but i have workaround that works nice :).

At the beginning of your main-server.ts file add these lines

```typescript
import * as xhr2 from 'xhr2';

//HACK - enables setting cookie header
xhr2.prototype._restrictedHeaders.cookie = false;
```

This allows you to send cookie header. Not clean solution, but it looks like only one so far.

Hope this helps :)

sangeet003:
Even retrieving cookie does not work on Server Side.
I was able to set cookies, but couldn't retrieve the same.

`Zone.current.get('cookie')`

The above used to work in angular2 universal starters but not in angular4

kukjevov:
Here you can see how to use cookies.

https://github.com/kukjevov/ng-universal-demo/tree/mine

[...]

PR #34111, opened 2019-11-27
feat(platform-server): Expose XHR client used in platform-server/http

let-aurn:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

I wish to attach cookies to HTTP requests during SSR.
But `cookie` is a [restricted header in node-xhr2](https://github.com/pwnall/node-xhr2/blob/master/src/001-xml_http_request.coffee#L369)

So I'm using the following https://github.com/angular/angular/issues/15730#issuecomment-291578814
```typescript
import * as xhr2 from 'xhr2';
xhr2.prototype._restrictedHeaders.cookie = false;
```

Everything works fine.

But then I noticed that this works only if the `xhr2` resolved in `import * as xhr2 from 'xhr2'` is the actual one used by [platform-server/src/http.ts](https://github.com/angular/angular/blob/master/packages/platform-server/src/http.ts#L10).  
Meaning that, if I add the dependency `xhr2` in the `package.json` of my project, then the imported xhr2 will be considered as a new instance. Because of that, platform-server/http's xhr2 cookies setting remains unchanged (because I'm actually changing the settings of that new instance, not the one of platform-server/http).

## What is the new behavior?

This Pull Request is to expose the XHR client used in platform-server/http, so that anyone can tweak it the way they want, especially for SSR-related issues.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/platform-server/src/http.ts
+++ b/packages/platform-server/src/http.ts
@@ -121,3 +121,10 @@ export const SERVER_HTTP_PROVIDERS: Provider[] = [
     deps: [HttpBackend, Injector]
   }
 ];
+
+/**
+ * The instance of the XHR Client used in platform-server.
+ *
+ * @publicApi
+ */
+export const PLATFORM_SERVER_XHR = xhr2;
--- a/packages/platform-server/src/platform-server.ts
+++ b/packages/platform-server/src/platform-server.ts
@@ -6,6 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
+export {PLATFORM_SERVER_XHR} from './http';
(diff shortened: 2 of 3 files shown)


IgorMinar:
hi there, thank you for the PR but this is not the right way to go about this - we can't expose the implementation of the xhr because then we would not ever be able to change it.

can you please open a new issue and describe your usecase? we can then consider either adding an api to the platform server that allows you to set cookies or internally set unrestrict the cookie header within platform-server so that you can set it via the http client apis.

in any case, exposing the xhr2 instance is not the right way to go about this.

PR #34116, opened 2019-11-28
refactor(ivy): do not generate providedIn: null

kara:
We should only generate the `providedIn` property in injectable
defs if it has a non-null value. `null` does not communicate
any information to the runtime that isn't communicated already
by the absence of the property.

This should give us some modest code size savings (~1 KB in AIO main bundle)

### DIFF
--- a/packages/compiler/src/injectable_compiler_2.ts
+++ b/packages/compiler/src/injectable_compiler_2.ts
@@ -106,11 +106,12 @@ export function compileInjectable(meta: R3InjectableMetadata): InjectableDef {
 
   const token = meta.internalType;
 
+  const injectableProps: {[key: string]: o.Expression} = {token, factory: result.factory};
+
   // Only generate providedIn property if it has a non-null value
-  const injectableProps: {[key: string]: o.Expression} =
-      (meta.providedIn as o.LiteralExpr).value === null ?
-      {token, factory: result.factory} :
-      {token, factory: result.factory, providedIn: meta.providedIn};
+  if ((meta.providedIn as o.LiteralExpr).value !== null) {
+    injectableProps.providedIn = meta.providedIn;
+  }
 
   const expression =
       o.importExpr(Identifiers.ɵɵdefineInjectable).callFn([mapToMapExpression(injectableProps)]);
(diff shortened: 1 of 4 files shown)


kara:
[VE presubmit](https://test.corp.google.com/ui#id=OCL:282854147:BASE:282854154:1574901795850:d569da83)

[Ivy presubmit](https://test.corp.google.com/ui#id=OCL:282854147:BASE:282854763:1574902160875:eefc1e92)

petebacondarwin (approved):

packages/compiler/src/injectable_compiler_2.ts
```diff
@@ -99,14 +100,20 @@ export function compileInjectable(meta: R3InjectableMetadata): InjectableDef {
...
   const token = meta.internalType;
-  const providedIn = meta.providedIn;
 
-  const expression = o.importExpr(Identifiers.ɵɵdefineInjectable).callFn([mapToMapExpression(
-      {token, factory: result.factory, providedIn})]);
+  // Only generate providedIn property if it has a non-null value
+  const injectableProps: {[key: string]: o.Expression} =
+      (meta.providedIn as o.LiteralExpr).value === null ?
+      {token, factory: result.factory} :
+      {token, factory: result.factory, providedIn: meta.providedIn};
```
IgorMinar: the ternary operator and object literal duplicate makes this code harder to read than necessary. how about:

```ts
  const injectableProps: {[key: string]: o.Expression} = {token, factory: result.factory};

  // Only generate providedIn property if it has a non-null value
  if ((meta.providedIn as o.LiteralExpr).value !== null) {
    injectableProps.providedIn = meta.providedIn;
  };
```

IgorMinar (approved):
the rest looks great.

btw I suggest that you close #34076 in favor of this PR. Since github doesn't support basing one PR on top of this other, stacking and blocking PRs like this just creates more headaches for the PR author and the caretaker.

IgorMinar:

kara:
@IgorMinar I find stacking commits to be easier for me :) I don't put the "merge" label on, so it shouldn't affect the caretaker.

packages/compiler/src/injectable_compiler_2.ts
kara: Ah, yep. I originally avoided that because I didn't want to update the type lol, but you're right. I'll update it.

kara:
merge-assistance: I am a code owner for compiler

PR #34125, opened 2019-11-28
docs: update injectable migration guide title and schematic description

devversion:
* See individual commits

### DIFF
--- a/aio/content/guide/migration-injectable.md
+++ b/aio/content/guide/migration-injectable.md
@@ -1,4 +1,4 @@
-# Migration to Add Missing `@Injectable()` Decorators
+# Migration for missing `@Injectable()` decorators and incomplete provider definitions
 
 ### What does this schematic do?
 
--- a/aio/content/guide/updating-to-version-9.md
+++ b/aio/content/guide/updating-to-version-9.md
@@ -102,7 +102,7 @@ Read about the migrations the CLI handles for you automatically:
 
 - [Migrating from `Renderer` to `Renderer2`](guide/migration-renderer)
 - [Migrating undecorated classes](guide/migration-undecorated-classes)
-- [Migrating missing `@Injectable()` decorators](guide/migration-injectable)
+- [Migrating missing `@Injectable()` decorators and incomplete provider definitions](guide/migration-injectable)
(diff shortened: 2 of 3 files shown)


kapunahelewong (approved):
LGTM. Thank you, @devversion!

kapunahelewong:
merge-assistance: Global approval for this docs-only change.

aio/content/guide/migration-injectable.md
```diff
@@ -1,4 +1,4 @@
-# Migration to Add Missing `@Injectable()` Decorators
+# Migration for missing `@Injectable()` decorators and incomplete provider definitions
```
IgorMinar: this doesn't match the capitalization used in the other guides.

@kapunahelewong is this right? should we change the other guides or fix this one?

IgorMinar (changes requested):
the title looks good, but the capitalization doesn't. Let's see what @kapunahele suggests we do.

aio/content/guide/migration-injectable.md
devversion: Good point. Happy to make it consistent. For what it's worth: looks like some migration guides are not consistent either (e.g. `migration-dynamic-flag`)

kapunahelewong: Ah, this change is correct (sentence case, that is, only the first word is capitalized and others are lower case unless they are customarily capitalized like `@Injectable()`). I believe when we changed docs teams we went from title case to sentence case. @jbogarthyde, would you like to look into our standard and correct me if I'm missing something here?

jbogarthyde: That is correct. We have been applying it consistently to internal headings, but not as much to page headings (h1).  
Note that even when the page heading uses sentence case, the entry in the left-nav TOC should still be title case.

jbogarthyde (approved):

aio/content/guide/migration-injectable.md
kapunahelewong: For reference, @jbogarthyde's PR doing a consistency sweep: https://github.com/angular/angular/pull/34196

Thanks, Judy!

Issue #34056, opened 2019-11-26
[Ivy] [i18n] Translated messages are not loaded for library components


tsvetomir:
# 🐞 bug report

### Is this a regression?
Yes, the translated messages are picked up by Angular 6-8 and in Angular 9 with Ivy disabled.

### Description
Messages for third-party components are extracted successfully when running `ng xi18n`, but the compiler fails to locate the translated strings during compilation.

Possibly related: https://github.com/angular/angular/issues/32311, https://github.com/angular/angular/pull/32881

## 🔬 Minimal Reproduction
1. Clone https://github.com/tsvetomir/ivy-i18n-messages
1. Run `npm ci`
1. Run `ng serve --configuration=es` or `ng serve --configuration=es,production`

Result:
* None of the messages declared in `CalendarModule` are translated.
* The message declared in the application ("Hello, World!") is translated.
* If `enableIvy` is set to false in `tsconfig.json` messages are translated normally.

## 🔥 Exception or Error
No run-time errors.

## 🌍  Your Environment

**Angular Version:**
<pre><code>
Angular CLI: 9.0.0-rc.3
Node: 12.13.1
OS: darwin x64
Angular: 9.0.0-rc.3
... animations, cli, common, compiler, compiler-cli, core, forms
... language-service, platform-browser, platform-browser-dynamic
... router

Package                           Version
-----------------------------------------------------------
@angular-devkit/architect         0.900.0-rc.3
@angular-devkit/build-angular     0.900.0-rc.3
@angular-devkit/build-optimizer   0.900.0-rc.3
@angular-devkit/build-webpack     0.900.0-rc.3
@angular-devkit/core              9.0.0-rc.3
@angular-devkit/schematics        9.0.0-rc.3
@angular/localize                 9.0.0-rc.0
@ngtools/webpack                  9.0.0-rc.3
@schematics/angular               9.0.0-rc.3
@schematics/update                0.900.0-rc.3
rxjs                              6.5.3
typescript                        3.6.4
webpack                           4.41.2
</code></pre>

petebacondarwin:
I think I know what is going on here...

This is what the warnings look like:

```
No translation found for "826701167553276328" ("TODAY" - "kendo.calendar.today").
No translation found for "589112377451589685" ("Increase value" - "kendo.dateinput.increment").
No translation found for "6980524091606682769" ("Decrease value" - "kendo.dateinput.decrement").
```

[...]

PR #34135, opened 2019-11-28
ngcc/i18n - legacy message id format

petebacondarwin:
This PR is triggered by the use case given in https://github.com/angular/angular/issues/34056

The idea is that ngcc will now always (unless explicitly turned off) add all the legacy message ids to templates that it compiles. Then we can try to use these for matching translations when we get to actually merging the translations at compile-time.

(The first few commits are just general tidy up and not necessary for this PR)


gkalpak (approved):
LGTM for `fw-ngcc`.
Nice, managable commits :100: :heart:

packages/compiler-cli/test/compliance/r3_view_compiler_i18n_spec.ts
```diff
@@ -3457,6 +3457,38 @@ describe('i18n support in the template compiler', () => {
...
 
+  describe('$localize legacy message ids', () => {
+    it('should add legacy message ids if `enableI18nLegacyMessageIdFormat` is true', () => {
+      const input = `<div i18n>Some Message</div>`;
+
+      const output = String.raw `
+        var $I18N_0$;
+        if (ngI18nClosureMode) { … }
+        else {
+            $I18N_0$ = $localize \`:␟ec93160d6d6a8822214060dd7938bf821c22b226␟6795333002533525253:Some Message\`;
```
josephperrott: Just to sanity check since we have had weirdness with encodings after clang format ran, `␟` is the expected character in this string?

josephperrott (approved):
LGTM, on little thing I wanted to check to make sure about encoding.

packages/compiler-cli/test/compliance/r3_view_compiler_i18n_spec.ts
petebacondarwin: :-) Yes this is the `LEGACY_ID_INDICATOR` - see https://github.com/angular/angular/pull/34135/files#diff-57fe4c8c3c329d46f463b846ab9475faR47

AndrewKushnir (approved):
LGTM 👍

AndrewKushnir:
- [VE Presubmit](http://test/OCL:283368908:BASE:283368954:1575311745111:5779a7a5)
- [Ivy Presubmit](https://test.corp.google.com/OCL:283368908:BASE:283369130:1575311794658:5b8738c1)

petebacondarwin:
Merge-assistance: the integration test failure is not related (a problem with terser dependencies).

PR #34154, opened 2019-11-29
fix(compiler): i18n - trim whitespace from i18n metadata

### LINKED ISSUE #34147: whitespace is included in trans-unit id
# 🐞 Bug report

- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [x] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?

This has never been a problem before; I think it has surfaced because Prettier has started intruducing whitespace into i18n attribute values

see
https://github.com/prettier/prettier/commit/4e46f92b8648c7f25de62bfaf4368aa3f2e588d9

### Description

In a template, the 

    i18n="description  | name@@id" 

syntax can be used to fix the id of the translation unit. If for some reason (e.g. using Prettier) the template code is changed to

    i18n="
        description  | name@@id
    " 

the id in the resulting `messages.xlf` file now _includes the trailing whitespace_. This
messes up the subsequent xliffmerge run. Also sometimes we get duplicate trans-units
(whose id's differ by whitespace only)

## 🔬 Minimal Reproduction

- generate a project with

        ng new i18n-app

- add a whitespace-id'ed tag to the app template, and look at the resulting messages file (two instances of `sillyId`!)

        cd i18n-app
        echo '<p i18n="name| @@sillyId ">silly test</p>' >> src/app/app.component.html
        echo '<p i18n="name| @@sillyId">silly test</p>' >> src/app/app.component.html
        ng xi18n
        grep sillyId messages.xlf

## 🔥 Exception or Error
<pre><code>

</code></pre>

## 🌍 Your Environment
<pre><code>

Angular CLI: 8.3.20
Node: 10.13.0
OS: linux x64
Angular: 8.2.14
... router

@angular-devkit/architect         0.803.20
@angular-devkit/build-angular     0.803.20
@angular-devkit/build-optimizer   0.803.20
@angular-devkit/build-webpack     0.803.20
@angular-devkit/core              8.3.20
@angular-devkit/schematics        8.3.20
@angular/cli                      8.3.20
@ngtools/webpack                  8.3.20
@schematics/angular               8.3.20
@schematics/update                0.803.20

</code></pre>

**Anything else relevant?**

petebacondarwin:
Fixes #34147

### DIFF
--- a/packages/compiler/src/i18n/extractor_merger.ts
+++ b/packages/compiler/src/i18n/extractor_merger.ts
@@ -516,5 +516,5 @@ function _parseMessageMeta(i18n?: string): {meaning: string, description: string
       [meaningAndDesc.slice(0, descIndex), meaningAndDesc.slice(descIndex + 1)] :
       ['', meaningAndDesc];
 
-  return {meaning, description, id};
+  return {meaning, description, id: id.trim()};
 }
--- a/packages/compiler/src/render3/view/i18n/get_msg_utils.ts
+++ b/packages/compiler/src/render3/view/i18n/get_msg_utils.ts
@@ -10,7 +10,7 @@ import {mapLiteral} from '../../../output/map_util';
 import * as o from '../../../output/output_ast';
 
 import {serializeIcuNode} from './icu_serializer';
-import {i18nMetaToDocStmt, metaFromI18nMessage} from './meta';
+import {i18nMetaToDocStmt} from './meta';
 import {formatI18nPlaceholderName} from './util';
 
 /** Closure uses `goog.getMsg(message)` to lookup translations */
@@ -32,7 +32,7 @@ export function createGoogleGetMsgStatements(
   // const MSG_... = goog.getMsg(..);
   // I18N_X = MSG_...;
   const statements = [];
-  const jsdocComment = i18nMetaToDocStmt(metaFromI18nMessage(message));
+  const jsdocComment = i18nMetaToDocStmt(message);
   if (jsdocComment !== null) {
     statements.push(jsdocComment);
   }
--- a/packages/compiler/src/render3/view/i18n/localize_utils.ts
+++ b/packages/compiler/src/render3/view/i18n/localize_utils.ts
@@ -9,7 +9,6 @@ import * as i18n from '../../../i18n/i18n_ast';
 import * as o from '../../../output/output_ast';
 
 import {serializeIcuNode} from './icu_serializer';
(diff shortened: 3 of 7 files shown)


packages/compiler/src/i18n/extractor_merger.ts
```diff
@@ -508,6 +508,8 @@ function _getI18nAttr(p: html.Element): html.Attribute|null {
 function _parseMessageMeta(i18n?: string): {meaning: string, description: string, id: string} {
   if (!i18n) return {meaning: '', description: '', id: ''};
 
+  i18n = i18n.trim();
```
AndrewKushnir: I'm wondering if we should trim whitespace for `id` part only to avoid potentially breaking change in case there is a whitespace in front of the meaning part, i.e. for these messages below:
```
<div i18n="meaning|descripion">TEST</div>

<div i18n="
    meaning|descripion">TEST</div>
```
we'll generate different message ids (due to different `meaning` values). With the changes in this PR, we will generate the same message id for both, which is better, but may result in "new" messages to be translated in translation consoles.

AndrewKushnir (changes requested):
Looks great, but this change as is might lead to some minor breaking changes. So the proposal is to trim `id` part only.

packages/compiler/src/i18n/extractor_merger.ts
petebacondarwin: I wondered about this... I really don't like the idea that `meaning` is whitespace dependent in the templates.  It makes it very brittle (as indicated by your comment).
I appreciate that there is a danger of messing up people's translation files, if they have whitespace in their templates...

How about we make the new `$localize` message digest whitespace independent? This is not a breaking change since no one is using this yet. Then when people migrate from the legacy message ids, we transform the id to be whitespace agnostic?

AndrewKushnir: @petebacondarwin making new `$localize` message digest whitespace independent sounds great 👍

AndrewKushnir (approved):
LGTM 👍

AndrewKushnir:
- [VE Presubmit](http://test/OCL:283407514:BASE:283407594:1575322785611:a02bfc3a)
- [Ivy Presubmit](https://test.corp.google.com/OCL:283407514:BASE:283408211:1575322952397:1856fc00)

PR #34169, opened 2019-12-01
fix(ngcc): several fixes

gkalpak:
This PR includes several ngcc fixes. With these fixes, ngcc should be able to successfully process NativeScript. See individual commits for details.

Jira issue: [FW-1689][1]

##
TODO:

- [ ] Determine whether it is problematic that the synthetically created [function declaration][2] for TS helpers will be recognized as [non-synthetic][3] in `StaticInterpreter` (and how to fix it).
- [x] Implement the CommonJS fixes for UMD as well.
- [ ] Add integration tests.

[1]: https://angular-team.atlassian.net/browse/FW-1689
[2]: https://github.com/angular/angular/commit/e28ac708e#diff-3814db5d5b51f278653eb80c9ac981c1R235
[3]: https://github.com/angular/angular/blob/023c9bebe/packages/compiler-cli/src/ngtsc/partial_evaluator/src/interpreter.ts#L242-L244

[FW-1689]: https://angular-team.atlassian.net/browse/FW-1689

### DIFF
--- a/packages/compiler-cli/ngcc/src/dependencies/commonjs_dependency_host.ts
+++ b/packages/compiler-cli/ngcc/src/dependencies/commonjs_dependency_host.ts
@@ -71,14 +71,11 @@ export class CommonJsDependencyHost extends DependencyHostBase {
           } else if (ts.isObjectLiteralExpression(stmt.expression.right)) {
             // Import in object literal. E.g.:
             // `module.exports = {foo: require('...')}`
-            const requireCallsFromProperties: RequireCall[] =
-                stmt.expression.right.properties
-                    .filter(
-                        (prop): prop is ts.PropertyAssignment & {initializer: RequireCall} =>
-                            ts.isPropertyAssignment(prop) && isRequireCall(prop.initializer))
-                    .map(prop => prop.initializer);
-
-            requireCalls.push(...requireCallsFromProperties);
+            stmt.expression.right.properties.forEach(prop => {
+              if (ts.isPropertyAssignment(prop) && isRequireCall(prop.initializer)) {
+                requireCalls.push(prop.initializer);
+              }
+            });
           }
         }
       }
--- a/packages/compiler-cli/ngcc/src/dependencies/commonjs_dependency_host.ts
+++ b/packages/compiler-cli/ngcc/src/dependencies/commonjs_dependency_host.ts
@@ -7,7 +7,7 @@
  */
 import * as ts from 'typescript';
 import {AbsoluteFsPath} from '../../../src/ngtsc/file_system';
-import {isRequireCall} from '../host/commonjs_host';
+import {RequireCall, isReexportStatement, isRequireCall} from '../host/commonjs_host';
 import {DependencyHostBase} from './dependency_host';
 import {ResolvedDeepImport, ResolvedRelativeModule} from './module_resolver';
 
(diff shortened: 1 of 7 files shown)


petebacondarwin:
typo in the 2nd to last commit: `findDepeendencies`

petebacondarwin:
typos in the last commit:

> In the case of emitted helpers, the `tslib.__someHelper` expression was
resolved to a function declaration of the form
`export declare function __someHelper(...args: any[][]): any[];`, which
allows `getDefinitionOfFunction()` to correctly map it to a TS helper.

I think this should be `In the case of imported helpers, ...`.

And also `functinos`

petebacondarwin:
typo in 5th commit: By failing to recongize

packages/compiler-cli/ngcc/test/dependencies/commonjs_dependency_host_spec.ts
```diff
@@ -12,6 +12,29 @@ import {loadTestFiles} from '../../../test/helpers';
...
+  // require('...');
+  forSideEffects?: string[];
+
+  // __export(require('...'));
+  reExportsWithEmittedHelper?: string[];
+
+  // tslib_1.__exportStar(require('...'), exports);
+  reExportsWithImportedHelper?: string[];
+}
+
```
petebacondarwin: NIT: I prefer not to clutter the start of a test file with supporting stuff (like helper functions and interfaces) since it makes it more difficult to quickly get to the tests when browsing later. Since interfaces are always hoisted, can we move this down below the `describe` block?

packages/compiler-cli/ngcc/src/dependencies/commonjs_dependency_host.ts
```diff
@@ -40,33 +40,66 @@ export class CommonJsDependencyHost extends DependencyHostBase {
...
+            // Import in object literal. E.g.:
+            // `module.exports = {foo: require('...')}`
+            const requireCallsFromProperties: RequireCall[] =
+                stmt.expression.right.properties
+                    .filter(
+                        (prop): prop is ts.PropertyAssignment & {initializer: RequireCall} =>
+                            ts.isPropertyAssignment(prop) && isRequireCall(prop.initializer))
+                    .map(prop => prop.initializer);
+
+            requireCalls.push(...requireCallsFromProperties);
```
petebacondarwin: NIT: this could be more simple and not require intermediate arrays to be created:

```ts
        stmt.expression.right.properties.forEach(prop => {
          if(ts.isPropertyAssignment(prop) && isRequireCall(prop.initializer)) {
            requireCalls.push(prop.initializer);
          }
        });
```

packages/compiler-cli/ngcc/src/dependencies/commonjs_dependency_host.ts
```diff
@@ -40,33 +40,66 @@ export class CommonJsDependencyHost extends DependencyHostBase {
...
+          }
+        }
+      }
+    }
+
+    const importPaths = new Set(requireCalls.map(call => call.arguments[0].text));
```
petebacondarwin: Nice!

petebacondarwin (commented):
Reviewed all but the last comment.

packages/compiler-cli/ngcc/src/host/esm5_host.ts
```diff
@@ -209,6 +218,40 @@ export class Esm5ReflectionHost extends Esm2015ReflectionHost {
...
+              ts.createParameter(
+                  undefined, undefined, ts.createToken(ts.SyntaxKind.DotDotDotToken),
+                  ts.createIdentifier('args'), undefined,
+                  ts.createArrayTypeNode(
+                      ts.createArrayTypeNode(ts.createKeywordTypeNode(ts.SyntaxKind.AnyKeyword))),
+                  undefined),
+            ],
+            ts.createArrayTypeNode(ts.createKeywordTypeNode(ts.SyntaxKind.AnyKeyword)), undefined),
+        viaModule: null,
+      };
```
petebacondarwin: NIT: I would put this whole block into its own function (e..g `createSynthesizedHelperDeclaration(id)`) with the  big comment converted to JSDOC comment. This will keep the calling function much smaller and more readable.

petebacondarwin (approved):
The PR looks great! Regarding the last commit, I can accept this approach. I don't know what other ideas you had. One thought is if we could create a new thing called a `LazyFunction`, which we would match code of the  form `(this && this.<fn>) || function () { ... }`, which could then be visited more intelligently in the StaticInterpreter?

petebacondarwin:
@gkalpak is this PR nearly ready? Is it worth splitting out the non-contentious bits so that we can land them this week?

gkalpak:
I'll split the non-contentious bits into a separate PR this week 👍

gkalpak:
@petebacondarwin: I have addressed your comments (in `fixup` commits - except for the commit message typos) and moved some of the commits into separate PRs: #34437, #34441

I need to further split this into more PRs (for easier reviewing) and add similar fixes for UMD (where appropriate).

gkalpak:
Further split this up into separate PRs per fix: #34512, #34527, #34528
<sub>(Each builds on top of the previous one, so you can either only review all commits on the last one or only review the new commits on each one.)</sub>

gkalpak:
With all the "spin-off" PRs merged, the only remaining fix for this PR is to **correctly detect emitted TS helpers in ES5** (i.e. 0b89f17c3). (I need to clean it up or possibly rework the approach.)

This will also fix angular/ngcc-validation#762.

gkalpak:
The final fix is in #35139. This PR is now obsolete. Closing...

PR #34175, opened 2019-12-02
feat(common): add camelcase and kebabcase pipes

MrJithil:
## PR Checklist

## PR Type
This PR will add 2 new case conversion pipes.
1. `CamelCasePipe`
2. `KebabCasePipe`

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/packages/common/src/common.ts
+++ b/packages/common/src/common.ts
@@ -22,7 +22,7 @@ export {parseCookieValue as ɵparseCookieValue} from './cookie';
 export {CommonModule} from './common_module';
 export {NgClass, NgClassBase, NgForOf, NgForOfContext, NgIf, NgIfContext, NgPlural, NgPluralCase, NgStyle, NgStyleBase, NgSwitch, NgSwitchCase, NgSwitchDefault, NgTemplateOutlet, NgComponentOutlet} from './directives/index';
 export {DOCUMENT} from './dom_tokens';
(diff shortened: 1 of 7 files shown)


mhevery (changes requested):
Thanks for the PR. 

We discussed this and we don't feel that this belongs into angular. We would like to keep the Angular API surface focused on the most common use cases. This one seems developer specific.

Adding this would enlarge Angular surface area which will be maintenance cost. Given the utility we don't think we want to pay that tax going forward.

We are very sorry, I am sure you put a lot of effort into this.

MrJithil:
> Thanks for the PR.
> 
> We discussed this and we don't feel that this belongs into angular. We would like to keep the Angular API surface focused on the most common use cases. This one seems developer specific.
> 
> Adding this would enlarge Angular surface area which will be maintenance cost. Given the utility we don't think we want to pay that tax going forward.
> 
> We are very sorry, I am sure you put a lot of effort into this.

Thanks 👍

petebacondarwin:
@MrJithil - this is a perfect case for a 3rd party library. Angular was always designed from the beginning to allow developers to create libraries that provide additional functionality, whether it appears in the form of services, components or (in this case) pipes.

PR #34177, opened 2019-12-02
fix(language-service): bug of accessing a string index signature using dot notation

ivanwonder:
fix bug of the pr #33884 @ayazhafiz 
```
primitiveType: {[name: string]: string} = {};
```
`{{primitiveType.test}}` this will produce errors. because the `string` type doesn't have Symbol.
https://github.com/angular/angular/blob/168abc6d6f52713383411b14980e104c99bfeef5/packages/language-service/src/typescript_symbols.ts#L509

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/packages/language-service/src/typescript_symbols.ts
+++ b/packages/language-service/src/typescript_symbols.ts
@@ -246,7 +246,13 @@ class TypeWrapper implements Symbol {
 
   get name(): string {
     const symbol = this.tsType.symbol;
-    return (symbol && symbol.name) || '<anonymous>';
+    if (symbol) {
+      return symbol.name;
+    } else {
+      // the js primitive type(e.g. 'string') doesn't have Symbol.
+      // use the ts.TypeChecker to get the type name.
+      return this.context.checker.typeToString(this.tsType);
+    }
   }
 
   public readonly kind: DeclarationKind = 'type';
--- a/packages/language-service/src/typescript_symbols.ts
+++ b/packages/language-service/src/typescript_symbols.ts
@@ -244,16 +244,7 @@ class TypeWrapper implements Symbol {
     }
   }
 
-  get name(): string {
-    const symbol = this.tsType.symbol;
-    if (symbol) {
-      return symbol.name;
-    } else {
-      // A primitive type (e.g. 'string') doesn't have Symbol,
-      // so use the ts.TypeChecker to get the type name.
-      return this.context.checker.typeToString(this.tsType);
-    }
-  }
+  get name(): string { return this.context.checker.typeToString(this.tsType); }
 
   public readonly kind: DeclarationKind = 'type';
 
@@ -617,15 +608,10 @@ class PipeSymbol implements Symbol {
         let resultType: ts.Type|undefined = undefined;
         switch (this.name) {
           case 'async':
-            switch (parameterType.name) {
-              case 'Observable':
-              case 'Promise':
-              case 'EventEmitter':
-                resultType = getTypeParameterOf(parameterType.tsType, parameterType.name);
-                break;
-              default:
-                resultType = getTsTypeFromBuiltinType(BuiltinType.Any, this.context);
-                break;
+            // Get symbol of 'Observable', 'Promise', or 'EventEmitter' type.
+            const symbol = parameterType.tsType.symbol;
+            if (symbol) {
+              resultType = getTypeParameterOf(parameterType.tsType, symbol.name);
             }
             break;
           case 'slice':
--- a/packages/language-service/src/typescript_symbols.ts
+++ b/packages/language-service/src/typescript_symbols.ts
@@ -308,6 +308,10 @@ class TypeWrapper implements Symbol {
   }
 }
 
+class StringIndexTypeWrappr extends TypeWrapper {
+  public readonly type = new TypeWrapper(this.tsType, this.context);
+}
+
 class SymbolWrapper implements Symbol {
   private symbol: ts.Symbol;
   private _members?: SymbolTable;
@@ -506,10 +510,11 @@ class SymbolTableWrapper implements SymbolTable {
       //   obj.stringIndex // equivalent to obj['stringIndex'];
       //
       // In this case, return the type indexed by an arbitrary string key.
-      const symbol = this.stringIndexType.getSymbol();
-      if (symbol) {
-        return new SymbolWrapper(symbol, this.context, this.stringIndexType);
-      }
+
+      // if stringIndexType is js primitive type(e.g. 'string'), the Symbol is undefined;
+      // and In AstType.resolvePropertyRead method, the Symbol.type should get the right type.
+      // so I add a new Symbol type, 'StringIndexTypeWrappr'
+      return new StringIndexTypeWrappr(this.stringIndexType, this.context);
     }
 
     return undefined;
--- a/packages/language-service/src/typescript_symbols.ts
+++ b/packages/language-service/src/typescript_symbols.ts
@@ -249,8 +249,8 @@ class TypeWrapper implements Symbol {
     if (symbol) {
       return symbol.name;
     } else {
-      // the js primitive type(e.g. 'string') doesn't have Symbol.
-      // use the ts.TypeChecker to get the type name.
+      // A primitive type (e.g. 'string') doesn't have Symbol,
+      // so use the ts.TypeChecker to get the type name.
       return this.context.checker.typeToString(this.tsType);
     }
   }
@@ -314,7 +314,9 @@ class TypeWrapper implements Symbol {
   }
 }
 
-class StringIndexTypeWrappr extends TypeWrapper {
+// If stringIndexType a primitive type(e.g. 'string'), the Symbol is undefined;
+// and in AstType.resolvePropertyRead method, the Symbol.type should get the right type.
+class StringIndexTypeWrapper extends TypeWrapper {
   public readonly type = new TypeWrapper(this.tsType, this.context);
 }
 
@@ -516,11 +518,7 @@ class SymbolTableWrapper implements SymbolTable {
       //   obj.stringIndex // equivalent to obj['stringIndex'];
       //
       // In this case, return the type indexed by an arbitrary string key.
-
-      // if stringIndexType is js primitive type(e.g. 'string'), the Symbol is undefined;
-      // and In AstType.resolvePropertyRead method, the Symbol.type should get the right type.
-      // so I add a new Symbol type, 'StringIndexTypeWrappr'
-      return new StringIndexTypeWrappr(this.stringIndexType, this.context);
+      return new StringIndexTypeWrapper(this.stringIndexType, this.context);
     }
 
     return undefined;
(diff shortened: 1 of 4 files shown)


packages/language-service/src/typescript_symbols.ts
```diff
@@ -308,6 +314,10 @@ class TypeWrapper implements Symbol {
   }
 }
 
+class StringIndexTypeWrappr extends TypeWrapper {
```
ayazhafiz:
```suggestion
class StringIndexTypeWrapper extends TypeWrapper {
```

Can you also add documentation for this class like you did on line 522?

packages/language-service/src/typescript_symbols.ts
```diff
@@ -506,10 +516,11 @@ class SymbolTableWrapper implements SymbolTable {
...
       //
       // In this case, return the type indexed by an arbitrary string key.
-      const symbol = this.stringIndexType.getSymbol();
-      if (symbol) {
-        return new SymbolWrapper(symbol, this.context, this.stringIndexType);
-      }
+
+      // if stringIndexType is js primitive type(e.g. 'string'), the Symbol is undefined;
+      // and In AstType.resolvePropertyRead method, the Symbol.type should get the right type.
+      // so I add a new Symbol type, 'StringIndexTypeWrappr'
```
ayazhafiz:
```suggestion
```

packages/language-service/src/typescript_symbols.ts
```diff
@@ -506,10 +516,11 @@ class SymbolTableWrapper implements SymbolTable {
       //   obj.stringIndex // equivalent to obj['stringIndex'];
       //
       // In this case, return the type indexed by an arbitrary string key.
-      const symbol = this.stringIndexType.getSymbol();
-      if (symbol) {
-        return new SymbolWrapper(symbol, this.context, this.stringIndexType);
-      }
+
+      // if stringIndexType is js primitive type(e.g. 'string'), the Symbol is undefined;
```
ayazhafiz:
```suggestion
      // If stringIndexType a primitive type(e.g. 'string'), the Symbol is undefined;
```

packages/language-service/src/typescript_symbols.ts
```diff
@@ -506,10 +516,11 @@ class SymbolTableWrapper implements SymbolTable {
...
+      // and In AstType.resolvePropertyRead method, the Symbol.type should get the right type.
```
ayazhafiz:
```suggestion
      // and in AstType.resolvePropertyRead method, the Symbol.type should get the right type.
```

packages/language-service/test/diagnostics_spec.ts
```diff
@@ -147,6 +147,14 @@ describe('diagnostics', () => {
         expect(diags[0].messageText)
             .toBe(`Identifier 'badProperty' is not defined. 'Hero' does not contain such a member`);
       });
+
+      it('should not produce errors with dot notation if stringIndexType is js primitive type',
+         () => {
+           mockHost.override(TEST_TEMPLATE, `
+        {{primitiveType.test}}`);
```
ayazhafiz: Can you place this template on one line? Also, can you add a test in `completions_spec.ts`? There are similar tests in there already.

packages/language-service/test/project/app/parsing-cases.ts
```diff
@@ -192,6 +192,7 @@ export class TemplateReference {
   tupleArray: [string, Hero] = ['test', this.hero];
   league: Hero[][] = [this.heroes];
   heroesByName: {[name: string]: Hero} = {};
+  primitiveType: {[name: string]: string} = {};
```
ayazhafiz: Can you rename to `primitiveIndexType` or similar? `primitiveType` sounds like the type (of the wrapper) itself is primitive.

ayazhafiz (changes requested):

packages/language-service/src/typescript_symbols.ts
```diff
@@ -246,7 +246,13 @@ class TypeWrapper implements Symbol {
 
   get name(): string {
     const symbol = this.tsType.symbol;
-    return (symbol && symbol.name) || '<anonymous>';
+    if (symbol) {
+      return symbol.name;
+    } else {
+      // the js primitive type(e.g. 'string') doesn't have Symbol.
```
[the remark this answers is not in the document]
ivanwonder: yes, you are right.
for example
```
tupleArray: [string, Hero];
```
this will return `[string, Hero]`

packages/language-service/src/typescript_symbols.ts
ayazhafiz: Can you fix the typo here?

packages/language-service/src/typescript_symbols.ts
```diff
@@ -249,7 +249,7 @@ class TypeWrapper implements Symbol {
     if (symbol) {
       return symbol.name;
     } else {
-      // the js primitive type(e.g. 'string') doesn't have Symbol.
+      // a primitive type (e.g. 'string') doesn't have Symbol.
       // use the ts.TypeChecker to get the type name.
```
ayazhafiz: Could you capitalize the sentences here?

packages/language-service/src/typescript_symbols.ts
```diff
@@ -517,9 +519,8 @@ class SymbolTableWrapper implements SymbolTable {
       //
       // In this case, return the type indexed by an arbitrary string key.
 
-      // if stringIndexType is js primitive type(e.g. 'string'), the Symbol is undefined;
-      // and In AstType.resolvePropertyRead method, the Symbol.type should get the right type.
-      // so I add a new Symbol type, 'StringIndexTypeWrappr'
+      // If stringIndexType a primitive type(e.g. 'string'), the Symbol is undefined;
+      // and in AstType.resolvePropertyRead method, the Symbol.type should get the right type.
```
ayazhafiz: I think you can get rid of the comment here, since you added them on the new class :)

ayazhafiz (changes requested):
LGTM, just a couple of typos.

ayazhafiz (approved):

ayazhafiz:
Thank you!

packages/language-service/src/typescript_symbols.ts
```diff
@@ -246,7 +246,13 @@ class TypeWrapper implements Symbol {
 
   get name(): string {
     const symbol = this.tsType.symbol;
-    return (symbol && symbol.name) || '<anonymous>';
+    if (symbol) {
+      return symbol.name;
```
kyliau: Do we still need to check the symbol here?
I think the else case here (getting name from the typechecker) is more generic and will always return the right name.

ivanwonder: I think it should be removed, but this code uses it as symbol name.
https://github.com/angular/angular/blob/37fa0837954865d3653bcdc4564e732fbf8b8691/packages/language-service/src/typescript_symbols.ts#L620
for example, the typechecker will return `Promise<string>`, the symbol name is `Promise`. It confuses me. or I change it to `parameterType.tsType.symbol.name`.

ayazhafiz: Checking for symbol can be removed by changing the code in `PipeSymbol#selectSignature` from the current evaluation of the piped object's type parameter:

https://github.com/angular/angular/blob/37fa0837954865d3653bcdc4564e732fbf8b8691/packages/language-service/src/typescript_symbols.ts#L619-L630

to an evaluation of the type parameter using the typechecker:

```typescript
case 'async':
  const typeNode = this.context.checker.typeToTypeNode(parameterType.tsType);
  if (typeNode && ts.isTypeReferenceNode(typeNode) && 
      ts.isIdentifier(typeNode.typeName)) {
    const name = typeNode.typeName.text;
    const result = getTypeParameterOf(parameterType.tsType, name);
    if (result) resultType = result;
  }
  break;
```

(Maybe it's worth also adding a default of `any` to stay consistent with the current implementation, but its exclusion doesn't break any tests.)

ivanwonder: I test it before, maybe the `any` has something to do with this test:
https://github.com/angular/angular/blob/37fa0837954865d3653bcdc4564e732fbf8b8691/packages/language-service/test/diagnostics_spec.ts#L568

ayazhafiz: Can you elaborate? I didn’t have a problem with `any` with the implementation I commented.

ivanwonder: https://github.com/angular/angular/blob/37fa0837954865d3653bcdc4564e732fbf8b8691/packages/language-service/src/typescript_symbols.ts#L620
In the test of 'parse a template reference', the `parameterType` type is `any`, then the code will go to line 627.
Your code works fine. I mean the code before applying your modifications.

ayazhafiz: Ah, my bad; thank you for explaining. Either way, exclusion of the branch on line 627 should be okay since an `any` type will match any signature.

ivanwonder:
```
switch (this.name) {
          case 'async':
            const symbol = parameterType.tsType.symbol;
            if (symbol) {
              resultType = getTypeParameterOf(parameterType.tsType, symbol.name);
            }
            break;
          case 'slice':
            resultType = parameterType.tsType;
            break;
        }
```
why not use `parameterType.tsType.symbol` just I have said, it's simpler and easier to read

ayazhafiz: Yes, it is much simpler, and I think it would be great to use it in the way you suggest; the only reason I suggested the other one is because it doesn't rely on the bad-typed `symbol` property and is symmetric to your new implementation of `get name()`.

packages/language-service/src/typescript_symbols.ts
```diff
@@ -608,15 +608,10 @@ class PipeSymbol implements Symbol {
...
-            switch (parameterType.name) {
-              case 'Observable':
-              case 'Promise':
-              case 'EventEmitter':
-                resultType = getTypeParameterOf(parameterType.tsType, parameterType.name);
-                break;
-              default:
-                resultType = getTsTypeFromBuiltinType(BuiltinType.Any, this.context);
-                break;
+            // It's safe to use Symbol here for the type 'Observable, Promise, EventEmitter'.
```
ayazhafiz: Could you rephrase or elaborate this comment? What do you mean by safe?
```suggestion
            // It's safe to use Symbol here for the type 'Observable', 'Promise', or 'EventEmitter'.
```

ayazhafiz (approved):

packages/language-service/src/typescript_symbols.ts
ivanwonder: this place only needs to handle these three types. It's enough to use Symbol. Maybe replace it with 'enough'.

ayazhafiz: Consider rephrasing to `Get symbol of 'Observable', 'Promise', or 'EventEmitter' type` or similar.

ivanwonder: How about `Get TypeParameter of 'Observable', 'Promise', or 'EventEmitter' type`

kyliau (approved):
Thank you very much @ivanwonder
This solution is clean and elegant, much appreciated!

PR #34179, opened 2019-12-02
fix(bazel): pin terser to 4.4.0

alan-agius4:
4.4.1 caused a regression https://github.com/terser/terser/issues/527 which is breaking our CI with such errors
```
'http://localhost:4200/main.57ce335e666ac98d65ad.js 0:370275 TypeError: e is not a function
```

Example of failing build https://circleci.com/gh/angular/angular/549169?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

### DIFF
--- a/packages/bazel/src/schematics/ng-add/index.ts
+++ b/packages/bazel/src/schematics/ng-add/index.ts
@@ -59,7 +59,7 @@ function addDevDependenciesToPackageJson(options: Schema) {
       'rollup': '^1.25.2',
       'rollup-plugin-commonjs': '^10.1.0',
       'rollup-plugin-node-resolve': '^5.2.0',
-      'terser': '^4.3.9',
+      'terser': '4.4.0',
     };
 
     const recorder = host.beginUpdate(packageJson);
--- a/integration/bazel-schematics/test.sh
+++ b/integration/bazel-schematics/test.sh
@@ -51,8 +51,13 @@ function testNonBazel() {
   rm -rf dist src/main.dev.ts src/main.prod.ts
   # disable CLI's version check (if version is 0.0.0, then no version check happens)
   yarn --cwd node_modules/@angular/cli version --new-version 0.0.0 --no-git-tag-version
-  # re-add build-angular
+
+  # Work-around for https://github.com/terser/terser/issues/527 which should be fixed in terser 4.4.2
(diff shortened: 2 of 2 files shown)


petebacondarwin (approved):

packages/bazel/src/schematics/ng-add/index.ts
```diff
@@ -59,7 +59,7 @@ function addDevDependenciesToPackageJson(options: Schema) {
       'rollup': '^1.25.2',
       'rollup-plugin-commonjs': '^10.1.0',
       'rollup-plugin-node-resolve': '^5.2.0',
-      'terser': '^4.3.9',
+      'terser': '4.4.0',
```
gkalpak: Just to make sure I understand correctly, is this only needed in case we publish a `@angular/bazel` version before terser 4.4.2 is released, right?

Once 4.4.2 is released, should this be changed to `^4.4.2` again?

integration/bazel-schematics/package.json
```diff
@@ -8,6 +8,9 @@
     "@angular/cli": "file:../../node_modules/@angular/cli",
     "@bazel/bazel": "file:../../node_modules/@bazel/bazel"
   },
+  "resolutions": {
+    "terser": "4.4.0"
```
gkalpak: Could you add a comment here (à la https://github.com/angular/angular/commit/ad753d3fa7f5d45cc72c9d90b2aeff7eb437f779#diff-0f512cda5dff463a54c9e4db5cc1ad3cR178-R180), pointing to the terser issue and mentioning when this can be removed?

alan-agius4: Yeah, I am still looking at it. The problem is that in this test we are hitting hoisting issues.

packages/bazel/src/schematics/ng-add/index.ts
alan-agius4: Yes

integration/bazel-schematics/package.json
alan-agius4: Should be fixed now

gkalpak (approved):

josephperrott (approved):
LGTM

alan-agius4:
It seems that terser fixed the bug https://github.com/terser/terser/issues/527

Thus, I am going to close this as it's no longer needed.

PR #34181, opened 2019-12-02
fix(bazel): don't rely on @angular/core being as a depedency to insta…

### LINKED ISSUE #34164: @angular/core dependency not found in package.json when ng add @angular/bazel
# 🐞 bug report

### Affected Package

 The issue is caused by package @angular/bazel

### Description

Optional `@angular/core` package may be in dependency section in `package.json` but installed in project

## 🔬 Minimal Reproduction

```
$ git clone https://github.com/ngxs/store.git
$ yarn install
$ ng add @angular/bazel
```

## 🔥 Exception or Error

```
@angular/core dependency not found in package.json
```

## 🌍  Your Environment

**Angular Version:** 
latest

alan-agius4:
fix(bazel): don't rely on @angular/core being as a depedency to install @angular/bazel

With this change we fix the logic to detect if a package is installed, removing a package and add a package by using the CLI schematic helpers.

Also we save `@angular/bazel` package directly as a `devDependency` when doing `ng-add`.

Closes #34164

### DIFF
--- a/packages/bazel/src/schematics/ng-add/index.ts
+++ b/packages/bazel/src/schematics/ng-add/index.ts
@@ -12,13 +12,15 @@ import {JsonAstObject, parseJsonAst} from '@angular-devkit/core';
 import {Rule, SchematicContext, SchematicsException, Tree, apply, applyTemplates, chain, mergeWith, url} from '@angular-devkit/schematics';
 import {NodePackageInstallTask} from '@angular-devkit/schematics/tasks';
 import {getWorkspace, getWorkspacePath} from '@schematics/angular/utility/config';
+import {NodeDependencyType, addPackageJsonDependency, getPackageJsonDependency, removePackageJsonDependency} from '@schematics/angular/utility/dependencies';
 import {findPropertyInAstObject, insertPropertyInAstObjectInOrder} from '@schematics/angular/utility/json-utils';
 import {validateProjectName} from '@schematics/angular/utility/validation';
 
-import {isJsonAstObject, removeKeyValueInAstObject, replacePropertyInAstObject} from '../utility/json-utils';
+import {isJsonAstObject, replacePropertyInAstObject} from '../utility/json-utils';
 import {findE2eArchitect} from '../utility/workspace-utils';
 
 import {Schema} from './schema';
+
 
 
 /**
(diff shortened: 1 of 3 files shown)


packages/bazel/src/schematics/ng-add/index.ts
```diff
@@ -277,39 +238,25 @@ function backupAngularJson(): Rule {
...
     if (match) {
       const [_, major, minor] = match;
-      if (major < '6' || (major === '6' && minor < '4')) {
-        const recorder = host.beginUpdate(packageJson);
-        replacePropertyInAstObject(recorder, deps, 'rxjs', '~6.4.0');
-        host.commitUpdate(recorder);
+      if (major < '6' || (major === '6' && minor < '5')) {
+        addPackageJsonDependency(host, {
+          ...rxjsNode,
+          version: '~6.5.3',
```
alan-agius4: In Angular 9 we don't support RXJS 6.4

packages/bazel/src/schematics/ng-add/index.ts
```diff
@@ -12,73 +12,59 @@ import {JsonAstObject, parseJsonAst} from '@angular-devkit/core';
...
-    if (!packageJsonContent) {
-      throw new Error('Failed to read package.json content');
-    }
-    const jsonAst = parseJsonAst(packageJsonContent.toString()) as JsonAstObject;
-    const deps = findPropertyInAstObject(jsonAst, 'dependencies') as JsonAstObject;
-    const devDeps = findPropertyInAstObject(jsonAst, 'devDependencies') as JsonAstObject;
-
-    const angularCoreNode = findPropertyInAstObject(deps, '@angular/core');
-    if (!angularCoreNode) {
+    const angularCore = getPackageJsonDependency(host, '@angular/core');
```
kyliau: Could you please add a comment here how this solves https://github.com/angular/angular/issues/34164 ?

kyliau (approved):

packages/bazel/src/schematics/ng-add/index.ts
alan-agius4: Because this methods checks all type of dependencies.

PR #34183, opened 2019-12-02
perf(ivy): avoid duplicate state lookup and default function parameters

crisbeto:
Includes a few minor performance improvements:
* In the `nextContext` instruction we assign a new LView to the `LFrame.contextLView` and then we immediately look it up to get its context. We don't need to do that since we know the view that was assigned already.
* Removes the default value for the `level` parameter of `nextContextImpl` because it generates more code in es5 and is internal-only.
* Removes the default parameter from `setActiveHostElement` since it generates extra code and it's an internal function.
* Makes a check in `setElementExitFn` more strict since we're guaranteed for the value to match the type.

### DIFF
--- a/packages/core/src/render3/state.ts
+++ b/packages/core/src/render3/state.ts
@@ -302,7 +302,7 @@ function setActiveElementFlag(flag: ActiveElementFlags) {
  * @param elementIndex the element index value for the host element where
  *                     the directive/component instance lives
  */
-export function setActiveHostElement(elementIndex: number | null = null) {
+export function setActiveHostElement(elementIndex: number | null) {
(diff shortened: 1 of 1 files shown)


crisbeto:
The `integration_test` failure will be fixed by https://github.com/angular/angular/pull/34179.

mhevery (approved):

mhevery:
[presubmit](http://test/OCL:283380436:BASE:283380463:1575314890655:c8bb0635)
[ivy presubmit](http://test/OCL:283380436:BASE:283381271:1575315119647:ecc60467)

mhevery:
[presubmit](http://test/OCL:283380436:BASE:283574374:1575398168583:fa5c3780)

PR #34190, opened 2019-12-02
test(docs-infra): add unit tests for rxjs examples

sonukapoor:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [X] No

### DIFF
--- a/aio/content/examples/observables/src/creating.spec.ts
+++ b/aio/content/examples/observables/src/creating.spec.ts
@@ -1,0 +1,55 @@
+import { docRegionFromEvent, docRegionSubscriber } from './creating';
+
+describe('observables', () => {
+  it('should create an observable using the constructor', () => {
+    const console = {log: jasmine.createSpy('log')};
+    docRegionSubscriber(console);
+    expect(console.log).toHaveBeenCalledTimes(4);
+    expect(console.log.calls.allArgs()).toEqual([
+      [1],
+      [2],
+      [3],
+      ['Finished sequence'],
+    ]);
+  });
+
+  it('should listen to input changes', () => {
+    let triggerInputChange;
+    const input = {
+      value: 'Test',
+      addEventListener: jasmine
+        .createSpy('addEvent')
+        .and.callFake((eventName: string, cb: (e) => void) => {
+          if (eventName === 'keydown') {
+            triggerInputChange = cb;
+          }
+        }),
+      removeEventListener: jasmine.createSpy('removeEventListener'),
+    };
+
+    const document = { getElementById: () => input };
+    docRegionFromEvent(document);
+    triggerInputChange({keyCode: 65});
+    expect(input.value).toBe('Test');
+
+    triggerInputChange({keyCode: 27});
+    expect(input.value).toBe('');
+  });
+
+  it('should call removeEventListener when unsubscribing', (doneFn: DoneFn) => {
+    const input = {
+      addEventListener: jasmine.createSpy('addEvent'),
+      removeEventListener: jasmine
+        .createSpy('removeEvent')
+        .and.callFake((eventName: string, cb: (e) => void) => {
+          if (eventName === 'keydown') {
+            doneFn();
+          }
+        })
+    };
+
+    const document = { getElementById: () => input };
+    const subscription = docRegionFromEvent(document);
+    subscription.unsubscribe();
+  });
+});
--- a/aio/content/examples/observables/src/creating.ts
+++ b/aio/content/examples/observables/src/creating.ts
@@ -1,37 +1,38 @@
+// #docplaster
 
 import { Observable } from 'rxjs';
 
-// #docregion subscriber
+export function docRegionSubscriber(console) {
+  // #docregion subscriber
+  // This function runs when subscribe() is called
+  function sequenceSubscriber(observer) {
+    // synchronously deliver 1, 2, and 3, then complete
+    observer.next(1);
+    observer.next(2);
+    observer.next(3);
+    observer.complete();
 
-// This function runs when subscribe() is called
-function sequenceSubscriber(observer) {
-  // synchronously deliver 1, 2, and 3, then complete
-  observer.next(1);
-  observer.next(2);
-  observer.next(3);
-  observer.complete();
+    // unsubscribe function doesn't need to do anything in this
+    // because values are delivered synchronously
+    return {unsubscribe() {}};
+  }
 
-  // unsubscribe function doesn't need to do anything in this
-  // because values are delivered synchronously
-  return {unsubscribe() {}};
+  // Create a new Observable that will deliver the above sequence
+  const sequence = new Observable(sequenceSubscriber);
+
+  // execute the Observable and print the result of each notification
+  sequence.subscribe({
+    next(num) { console.log(num); },
+    complete() { console.log('Finished sequence'); }
+  });
+
+  // Logs:
+  // 1
+  // 2
+  // 3
+  // Finished sequence
+  // #enddocregion subscriber
 }
-
-// Create a new Observable that will deliver the above sequence
-const sequence = new Observable(sequenceSubscriber);
-
-// execute the Observable and print the result of each notification
-sequence.subscribe({
-  next(num) { console.log(num); },
-  complete() { console.log('Finished sequence'); }
-});
-
-// Logs:
-// 1
-// 2
-// 3
-// Finished sequence
-
-// #enddocregion subscriber
 
 // #docregion fromevent
 
@@ -51,16 +52,18 @@ function fromEvent(target, eventName) {
 
 // #enddocregion fromevent
 
-// #docregion fromevent_use
+export function docRegionFromEvent(document) {
+  // #docregion fromevent_use
 
-const ESC_KEY = 27;
-const nameInput = document.getElementById('name') as HTMLInputElement;
+  const ESC_KEY = 27;
+  const nameInput = document.getElementById('name') as HTMLInputElement;
 
-const subscription = fromEvent(nameInput, 'keydown')
-  .subscribe((e: KeyboardEvent) => {
+  const subscription = fromEvent(nameInput, 'keydown').subscribe((e: KeyboardEvent) => {
     if (e.keyCode === ESC_KEY) {
       nameInput.value = '';
     }
   });
+  // #enddocregion fromevent_use
+  return subscription;
+}
 
-// #enddocregion fromevent_use
--- a/aio/content/examples/observables/src/multicasting.spec.ts
+++ b/aio/content/examples/observables/src/multicasting.spec.ts
@@ -1,0 +1,48 @@
+import { docRegionDelaySequence, docRegionMulticastSequence } from './multicasting';
+
+describe('multicasting', () => {
+  let console;
+  beforeEach(() => {
+    jasmine.clock().install();
+    console = {log: jasmine.createSpy('log')};
+  });
+
+  afterEach(() => {
+    jasmine.clock().uninstall();
+  });
+
+  it('should create an observable and emit in sequence', () => {
+    docRegionDelaySequence(console);
+    jasmine.clock().tick(10000);
+    expect(console.log).toHaveBeenCalledTimes(12);
+    expect(console.log.calls.allArgs()).toEqual([
+      [1],
+      ['1st subscribe: 1'],
+      ['2nd subscribe: 1'],
+      [2],
+      ['1st subscribe: 2'],
+      ['2nd subscribe: 2'],
+      [3],
+      ['Finished sequence'],
+      ['1st subscribe: 3'],
+      ['1st sequence finished.'],
+      ['2nd subscribe: 3'],
+      ['2nd sequence finished.']
+    ]);
+  });
+
+  it('should create an observable and multicast the emissions', () => {
+    docRegionMulticastSequence(console);
+    jasmine.clock().tick(10000);
+    expect(console.log).toHaveBeenCalledTimes(7);
+    expect(console.log.calls.allArgs()).toEqual([
+      ['1st subscribe: 1'],
+      ['1st subscribe: 2'],
+      ['2nd subscribe: 2'],
+      ['1st subscribe: 3'],
+      ['2nd subscribe: 3'],
+      ['1st sequence finished.'],
+      ['2nd sequence finished.']
+    ]);
+  });
+});
--- a/aio/content/examples/observables/src/multicasting.ts
+++ b/aio/content/examples/observables/src/multicasting.ts
@@ -1,155 +1,160 @@
+// #docplaster
 
 import { Observable } from 'rxjs';
 
-// #docregion delay_sequence
+export function docRegionDelaySequence(console) {
+  // #docregion delay_sequence
+  function sequenceSubscriber(observer) {
+    const seq = [1, 2, 3];
+    let timeoutId;
 
-function sequenceSubscriber(observer) {
-  const seq = [1, 2, 3];
-  let timeoutId;
+    // Will run through an array of numbers, emitting one value
+    // per second until it gets to the end of the array.
+    function doInSequence(arr, idx) {
+      timeoutId = setTimeout(() => {
+        observer.next(arr[idx]);
+        if (idx === arr.length - 1) {
+          observer.complete();
+        } else {
+          doInSequence(arr, ++idx);
+        }
+      }, 1000);
+    }
 
-  // Will run through an array of numbers, emitting one value
+    doInSequence(seq, 0);
+
+    // Unsubscribe should clear the timeout to stop execution
+    return {
+      unsubscribe() {
+        clearTimeout(timeoutId);
+      }
+    };
+  }
+
+  // Create a new Observable that will deliver the above sequence
+  const sequence = new Observable(sequenceSubscriber);
+
+  sequence.subscribe({
+    next(num) { console.log(num); },
+    complete() { console.log('Finished sequence'); }
+  });
+
+  // Logs:
+  // (at 1 second): 1
+  // (at 2 seconds): 2
+  // (at 3 seconds): 3
+  // (at 3 seconds): Finished sequence
+
+  // #enddocregion delay_sequence
+
+  // #docregion subscribe_twice
+
+  // Subscribe starts the clock, and will emit after 1 second
+  sequence.subscribe({
+    next(num) { console.log('1st subscribe: ' + num); },
+    complete() { console.log('1st sequence finished.'); }
+  });
+
+  // After 1/2 second, subscribe again.
+  setTimeout(() => {
+    sequence.subscribe({
+      next(num) { console.log('2nd subscribe: ' + num); },
+      complete() { console.log('2nd sequence finished.'); }
+    });
+  }, 500);
+
+  // Logs:
+  // (at 1 second): 1st subscribe: 1
+  // (at 1.5 seconds): 2nd subscribe: 1
+  // (at 2 seconds): 1st subscribe: 2
+  // (at 2.5 seconds): 2nd subscribe: 2
+  // (at 3 seconds): 1st subscribe: 3
+  // (at 3 seconds): 1st sequence finished
+  // (at 3.5 seconds): 2nd subscribe: 3
+  // (at 3.5 seconds): 2nd sequence finished
+
+  // #enddocregion subscribe_twice
+}
+
+export function docRegionMulticastSequence(console) {
+  // #docregion multicast_sequence
+  function multicastSequenceSubscriber() {
+    const seq = [1, 2, 3];
+    // Keep track of each observer (one for every active subscription)
+    const observers = [];
+    // Still a single timeoutId because there will only ever be one
+    // set of values being generated, multicasted to each subscriber
+    let timeoutId;
+
+    // Return the subscriber function (runs when subscribe()
+    // function is invoked)
+    return observer => {
+      observers.push(observer);
+      // When this is the first subscription, start the sequence
+      if (observers.length === 1) {
+        timeoutId = doSequence({
+          next(val) {
+            // Iterate through observers and notify all subscriptions
+            observers.forEach(obs => obs.next(val));
+          },
+          complete() {
+            // Notify all complete callbacks
+            observers.slice(0).forEach(obs => obs.complete());
+          }
+        }, seq, 0);
+      }
+
+      return {
+        unsubscribe() {
+          // Remove from the observers array so it's no longer notified
+          observers.splice(observers.indexOf(observer), 1);
+          // If there's no more listeners, do cleanup
+          if (observers.length === 0) {
+            clearTimeout(timeoutId);
+          }
+        }
+      };
+    };
+  }
+
+  // Run through an array of numbers, emitting one value
   // per second until it gets to the end of the array.
-  function doInSequence(arr, idx) {
-    timeoutId = setTimeout(() => {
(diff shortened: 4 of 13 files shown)


sonukapoor:
@gkalpak Please review draft.

gkalpak:
This lgtm for now. We will re-work it the infrastructure, before the merging the PR, but this is good enough for now so that you can keep working on the actual tests. Thx for sorting it out 👍

gkalpak:
As discussed "offline", here are my thoughts (and suggestions for next steps):

- A proper solution might involve setting up a different project type for Observable-related projects (e.g. not involving Angular dependencies and providing dependencies needed for observables only). This would be a little expensive in terms of both initial implementation and maintenance, though.

- I realized that the observables examples are not offered for download/StackBlitz, so the users only really see the code snippets (not the whole files). This gives us some flexibility on putting stuff around the docregions that will help us with testing, but without confusing the users (which will never see that extra stuff).

- I thought about using Karma for the tests. This would give us access to a real browser. I then realized that that would only solve part of the problem. We would need to add appropriate HTML files and that would take more time/maintenance/complexity and would most likely still require infrastructure changes. Not ideal.

- Finally, I noticed another issue: The code in the files executes eagerly (i.e. as soon as the file is loaded), which means that we would have a hard time testing stuff in the test files, because it would have happened before even loading the test files.

---
So, given the above, I think a good solution ("good" in the sense that it is fairly simple, will not take a lot of time to implement infrastructure for and will allow us to focus on tests (which is the primary focus for now)) would be the following:

- We could wrap each docregion (i.e. code snippet) in a function and export that function. This way, the code snippets that the user sees wouldn't change (because they would only see the function body), but the code would not be executed as soon as the file is loaded during tests.

- In tests, we could call the exported functions to run the code. This gives us the freedom to:
  a. Set up spies and mocks before executing the function.
  b. Pass mocks for globals as arguments to the function. For example, we could pass a mock `document` to the function etc.

As an example, here is what a code snippet could look like before my suggested change:

```js
// #docregion foo
const elem = document.querySelector('foo');
const subscription = fromEvent(elem, 'click').subscribe(() => console.log('*click*'));

setTimeout(() => subscription.unsubscribe(), 10000);
// #enddocregion foo
```

And here is how it would be transformed with my suggestion:

```js
// By wrapping a function around the code, we make it "lazy"
// (i.e. it won't be executed immediately).
//
// We can pass in "globals" (in TS we can give them proper types).
// Here for example, we can pass in a mock `console` to capture the logs.
// (We could even pass a mock `setTimeout()`, but that is unnecessary,
// since we can mock that with `jasmine.clock().install()` etc.)
export function docregionFoo(console, document) {

  // The code-snippet that the user will see remains unchanged.
  // (The indentation might change, but that is fine.)
  // #docregion foo
  const elem = document.querySelector('foo');
  const subscription = fromEvent(elem, 'click').subscribe(() => console.log('*click*'));

  setTimeout(() => subscription.unsubscribe(), 10000);
  // #enddocregion foo

  // Here, return the `subscription`, if we need to run assertions against it.
  return subscription;
}
```

petebacondarwin:
@sonukapoor can we close this PR?

sonukapoor:
> @sonukapoor can we close this PR?

This PR is awaiting @gkalpak's review :)

sonukapoor:
Blocked by https://github.com/angular/angular/pull/38143

sonukapoor:
@gkalpak This is ready for you to review. Thanks

aio/content/examples/observables/src/multicasting.ts
```diff
@@ -23,17 +24,31 @@ function sequenceSubscriber(observer) {
...
+    }
+  };
 }
+// #enddocregion delay_sequence
+/*
+  Because of how the code is merged together using the doc regions,
+  we need to start the indentation of the below body function at the most left
+  position.
+*/
+export function docDelaySequence(console) {
```
gkalpak, line 39:
```suggestion
export function docRegionDelaySequence(console) {
```

aio/content/examples/observables/src/multicasting.ts
```diff
@@ -23,17 +24,31 @@ function sequenceSubscriber(observer) {
...
 
 sequence.subscribe({
-  next(num) { console.log(num); },
-  complete() { console.log('Finished sequence'); }
+  next(num) {
+    console.log(num);
+  },
+  complete() {
+    console.log('Finished sequence');
+  }
```
gkalpak, line 51: These changes look unrelated. Please revert.

aio/content/examples/observables/src/multicasting.ts
```diff
@@ -48,15 +63,23 @@ sequence.subscribe({
...
 // Subscribe starts the clock, and will emit after 1 second
 sequence.subscribe({
-  next(num) { console.log('1st subscribe: ' + num); },
-  complete() { console.log('1st sequence finished.'); }
+  next(num) {
+    console.log('1st subscribe: ' + num);
+  },
+  complete() {
+    console.log('1st sequence finished.');
+  }
```
gkalpak, line 71: These changes look unrelated. Please revert.

aio/content/examples/observables/src/multicasting.ts
```diff
@@ -48,15 +63,23 @@ sequence.subscribe({
...
 setTimeout(() => {
   sequence.subscribe({
-    next(num) { console.log('2nd subscribe: ' + num); },
-    complete() { console.log('2nd sequence finished.'); }
+    next(num) {
+      console.log('2nd subscribe: ' + num);
+    },
+    complete() {
+      console.log('2nd sequence finished.');
+    }
```
gkalpak, line 82: These changes look unrelated. Please revert.

aio/content/examples/observables/src/multicasting.ts
```diff
@@ -84,20 +109,24 @@ function multicastSequenceSubscriber() {
...
+          }
         },
-        complete() {
-          // Notify all complete callbacks
-          observers.slice(0).forEach(obs => obs.complete());
-        }
-      }, seq, 0);
+        seq,
+        0
+      );
```
gkalpak, line 129: Same as above. Please revert the unrelated, formatting changes.

aio/content/examples/observables/src/multicasting.ts
```diff
@@ -126,6 +155,12 @@ function doSequence(observer, arr, idx) {
   }, 1000);
 }
 
+// #enddocregion multicast_sequence
+// Because of how the code is merged together using the doc regions,
+// we need to start the indention of the below body function at the most left
+// position.
+export function docMulticastSequence(console) {
```
gkalpak, line 162:
```suggestion
export function docRegionMulticastSequence(console) {
```

aio/content/examples/observables/src/multicasting.spec.ts
```diff
@@ -0,0 +1,27 @@
+import { docDelaySequence, docMulticastSequence } from './multicasting';
+import { toArray } from 'rxjs/operators';
+
+describe('multicasting', () => {
+  let console;
+  beforeEach(() => {
+    console = {
+      log: jasmine.createSpy
```
gkalpak, line 8: Was this supposed to be `jasmine.createSpy('log')`?

aio/content/examples/observables/src/multicasting.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+    };
+  });
+
+  it('should create an observable and emit in sequence', (doneFn: DoneFn) => {
+    const sequence = docDelaySequence(console);
+    sequence.pipe(toArray()).subscribe((output: Array<number>) => {
+      expect(output).toEqual([1, 2, 3]);
+      doneFn();
+    });
+  });
```
gkalpak, line 18: Same as above, this test does not verify what happens with the subscriptions inside `docDelaySequence()`. This is the main thing we need to assert. (Same for the test below.)

gkalpak (changes requested):
Here is my partial review. Many of the comments apply to files in all observables examples, so please make the changes to all of them.

aio/content/examples/observables/src/creating.ts
```diff
@@ -16,6 +17,14 @@ function sequenceSubscriber(observer) {
   return {unsubscribe() {}};
 }
 
+// #enddocregion subscriber
+/*
+  Because of how the code is merged together using the doc regions,
+  we need to start the indentation of the below body function at the most left
+  position.
+*/
```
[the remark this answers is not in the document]
sonukapoor: Hm... I am pretty sure, I had tried this and it was not indenting properly, which is why I had the comment in there in the first place. I have tried this again now and it seems to work fine 👍

aio/content/examples/observables/src/creating.spec.ts
```diff
@@ -0,0 +1,49 @@
+import { docRegionFromEvent, docRegionSubscriber } from './creating';
+import { cold } from 'jasmine-marbles';
+
+describe('observables', () => {
+  it('should create an observable using the constructor', () => {
+    const console = { log: jasmine.createSpy };
+    const source = docRegionSubscriber(console);
+    const expected = cold('(abc|)', { a: 1, b: 2, c: 3 });
```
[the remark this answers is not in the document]
sonukapoor: There is only one observable. Yes, It will subscribe to the emitted values and verify them. 
Ok. I will remove the marbles and verify that `console.log` was called instead.

aio/content/examples/observables/src/creating.ts
```diff
@@ -51,16 +65,19 @@ function fromEvent(target, eventName) {
...
-const ESC_KEY = 27;
-const nameInput = document.getElementById('name') as HTMLInputElement;
+  const ESC_KEY = 'Escape';
+  const nameInput = document.getElementById('name') as HTMLInputElement;
 
-const subscription = fromEvent(nameInput, 'keydown')
-  .subscribe((e: KeyboardEvent) => {
-    if (e.keyCode === ESC_KEY) {
+  const subscription = fromEvent(nameInput, 'keydown');
+  subscription.subscribe((e: KeyboardEvent) => {
```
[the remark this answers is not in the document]
sonukapoor: I am using the `subscription` to make sure that the `removeEventListener` is called when unsubscribing. See the test [here](https://github.com/sonukapoor/angular/blob/unit-tests-for-stand-alone-code/aio/content/examples/observables/src/creating.spec.ts#L32).

sonukapoor:
@gkalpak, I have addressed your review comments. However, I have some difficulties to get the doc-plasters working without the indentation when there are imports used. Please see the individual sections where I have added a screenshot with the issue.

aio/content/examples/observables/src/creating.ts
```diff
@@ -2,51 +2,38 @@
...
+    observer.next(1);
+    observer.next(2);
+    observer.next(3);
+    observer.complete();
+
+    // unsubscribe function doesn't need to do anything in this
+    // because values are delivered synchronously
+    return {unsubscribe() {}};
+  }
+  // Create a new Observable that will deliver the above sequence
```
gkalpak, line 19:
```suggestion

  // Create a new Observable that will deliver the above sequence
```

aio/content/examples/observables/src/creating.spec.ts
```diff
@@ -1,12 +1,16 @@
 import { docRegionFromEvent, docRegionSubscriber } from './creating';
-import { cold } from 'jasmine-marbles';
 
 describe('observables', () => {
   it('should create an observable using the constructor', () => {
-    const console = { log: jasmine.createSpy };
+    const consoleLogSpy  = spyOn(console , 'log');
```
gkalpak, line 5: Generally, it is preferable to use a spy object (instead of spying on the real `console.log()` (in case some other pieve of code wants to `console.log()` something):

```suggestion
    const console = {log: jasmine.createSpy('log')};
```

aio/content/examples/observables/src/creating.spec.ts
```diff
@@ -1,12 +1,16 @@
...
     const source = docRegionSubscriber(console);
-    const expected = cold('(abc|)', { a: 1, b: 2, c: 3 });
-    expect(source).toBeObservable(expected);
+    source.subscribe(() => {
```
gkalpak, line 7: Is subscribing here necessary?
Couldn't we run the assertions immediately?

aio/content/examples/observables/src/creating.ts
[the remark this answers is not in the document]
gkalpak: I don't think you need this change to test the unsubscription. You can just change [this line][1] of the test:

```diff
-    subscription.subscribe().unsubscribe();
+    subscription.unsubscribe();

[1]: https://github.com/sonukapoor/angular/blob/230b42f1bfeee39d40538c443533d6e876d2faad/aio/content/examples/observables/src/creating.spec.ts#L53

File: aio/content/examples/practical-observable-usage/src/backoff.spec.ts [hunk 15]
```diff
@@ -0,0 +1,31 @@
+import { backoff } from './backoff';
+import { map } from 'rxjs/operators';
+import { Subject } from 'rxjs';
```
gkalpak, line 3:
```suggestion
import { Subject } from 'rxjs';
import { map } from 'rxjs/operators';
import { backoff } from './backoff';
```

aio/content/examples/practical-observable-usage/src/backoff.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+          }
+          return val;
+        }),
+        backoff(3, 250)
+      )
+      .subscribe(a => {
+        // should not emit 4
+        expect(a).not.toEqual(4);
+        if (a === 3) {
+          doneFn();
```
gkalpak, line 22: This test looks broken (for a couple fo reasons).

1. How are we sure 4 is not emitted if we exit on 3? :grin:
2. Since 4 is emitted once (i.e. the observable only errors once), how are we verifying the backoff?

aio/content/examples/practical-observable-usage/src/backoff.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+        }
+      });
+
+    obs.next(1);
+    obs.next(2);
+    obs.next(3);
+    obs.next(4);
+  });
```
gkalpak, line 30: Ideally, we should also test the backing off (i.e. the increased delay between retries).
Since this is a little complicated, feel free to add a TODO comment for us to do later (at some point...maybe 😛).

aio/content/examples/practical-observable-usage/src/typeahead.ts
```diff
@@ -1,18 +1,32 @@
-import { fromEvent } from 'rxjs';
-import { ajax } from 'rxjs/ajax';
-import { debounceTime, distinctUntilChanged, filter, map, switchMap } from 'rxjs/operators';
+/*
+  Because of how the code is merged together using the doc regions,
+  we need to indent the imports with the function below.
+*/
+// #docplaster
+// #docregion
```
gkalpak, line 7:
```suggestion
```

aio/content/examples/practical-observable-usage/src/typeahead.spec.ts
```diff
@@ -0,0 +1,29 @@
+import { docRegionTypeahead } from './typeahead';
+import { of } from 'rxjs';
```
gkalpak, line 2:
```suggestion
import { of } from 'rxjs';
import { docRegionTypeahead } from './typeahead';
```

aio/content/examples/observables/src/multicasting.spec.ts
```diff
@@ -0,0 +1,45 @@
+import { docRegionDelaySequence, docRegionMulticastSequence } from './multicasting';
+import { toArray } from 'rxjs/operators';
```
gkalpak, line 2:
```suggestion
import { toArray } from 'rxjs/operators';
import { docRegionDelaySequence, docRegionMulticastSequence } from './multicasting';
```

aio/content/examples/observables/src/subscribing.spec.ts
```diff
@@ -0,0 +1,22 @@
+import { docRegionObserver } from './subscribing';
+import { toArray } from 'rxjs/operators';
```
gkalpak, line 2:
```suggestion
import { toArray } from 'rxjs/operators';
import { docRegionObserver } from './subscribing';
```

aio/content/examples/practical-observable-usage/src/typeahead.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+    const ajax = (url: string) => of('Hello World');
+
+    const typeahead = docRegionTypeahead(document, ajax);
+    typeahead.subscribe(data => {
+      expect(data).toBe('Hello World');
+      doneFn();
+    });
+
+    triggertInputChange({ target: { value: 'Hello' } });
+  });
```
gkalpak, line 28: This test changes too little of the implementation. Even though testing everything is quite complicated (and thus outside the scope of this PR), there is some low-hanging fruit:

- Make the mock `ajax` a spy so we can track calls.
- Verify that there is no AJAX call when the value is less than 3 characters.
- Verify that there is not AJAX call when the value of the input does not change (even if the `input` event is emitted on the element).
- Potentially, use jasmine's [mock clock][1] (or RxJS' [TestScheduler][2]?) to verify that events are debounced.

##
**Bonus points:**

The current example that is calling the same API endpoint without taking the emitted value into account is rather unrealistic.

- Improve the example, by using the emitted input value, for example:
  ```diff
       distinctUntilChanged(),
  -    switchMap(() => ajax('/api/endpoint'))
  +    switchMap(searchTerm => ajax(`/api/endpoint?search=${searchTerm}`))
     );
  ```
2. Update the test (or add a new test) to verify that AJAX calls are made to the correct endpoint.
3. Update the test (or add a new test) to verify that no AJAX calls are made for intermediate values when debouncing.

[1]: https://jasmine.github.io/api/3.6/Clock
[2]: https://rxjs.dev/api/testing/TestScheduler

gkalpak (changes requested):
Thx, @sonukapoor! Here's another round of review comments :smiley:
(I have reviewed all examples except for the `rx-library` one.)

Pending comments from previous review (for my reference):
- https://github.com/angular/angular/pull/34190#discussion_r482845609

aio/content/examples/observables/src/creating.ts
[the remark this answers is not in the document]
sonukapoor: Ok. Would you like me to remove the test?

aio/content/examples/observables/src/multicasting.spec.ts
```diff
@@ -1,26 +1,44 @@
...
-    console = {
-      log: jasmine.createSpy
-    };
+    consoleLogSpy = spyOn(console , 'log');
   });
 
   it('should create an observable and emit in sequence', (doneFn: DoneFn) => {
-    const sequence = docDelaySequence(console);
+    const sequence = docRegionDelaySequence(console);
     sequence.pipe(toArray()).subscribe((output: Array<number>) => {
```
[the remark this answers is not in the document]
sonukapoor: I have changed all tests to not subscribe anymore.

aio/content/examples/practical-observable-usage/src/typeahead.spec.ts
sonukapoor: I have added the scenarios into the same commit. But let me know if you want me to split this up into a new commit (which might make sense).

gkalpak (changes requested):
We are getting there :smiley:

In addition to my other comments, can you also:
- Move the change of using `searchTerm` in the typeahead request to a separate commit (as discussed in https://github.com/angular/angular/pull/34190#discussion_r484560207) and adjust the tests accordingly (i.e. each commit should include the tests that are relevant to the source code changes it introduces).
- Move the changes to the `rx-library` example to a new PR (in the interest of getting this one in faster and avoiding its getting stale).

aio/content/examples/practical-observable-usage/src/backoff.spec.ts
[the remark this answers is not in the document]
sonukapoor: I have added a `TODO` to fix the test, once this PR is merged.

aio/content/examples/practical-observable-usage/src/typeahead.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+    triggertInputChange({ target: { value: 'foo' } });
+    jasmine.clock().tick(11);
+    expect(ajax).toHaveBeenCalled();
+    ajax.calls.reset();
+    triggertInputChange({ target: { value: 'foo' } });
+    jasmine.clock().tick(11);
+    expect(ajax).not.toHaveBeenCalled();
+  });
+
+  it('should make an ajax call to the corrent endpoint', () => {
```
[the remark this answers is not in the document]
sonukapoor: Done.

aio/content/examples/observables/src/multicasting.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+    expect(console.log).toHaveBeenCalledWith('2nd subscribe: 1');
+    expect(console.log).toHaveBeenCalledWith(2);
+    expect(console.log).toHaveBeenCalledWith('1st subscribe: 2');
+    expect(console.log).toHaveBeenCalledWith('2nd subscribe: 2');
+    expect(console.log).toHaveBeenCalledWith(3);
+    expect(console.log).toHaveBeenCalledWith('Finished sequence');
+    expect(console.log).toHaveBeenCalledWith('1st subscribe: 3');
+    expect(console.log).toHaveBeenCalledWith('1st sequence finished.');
+    expect(console.log).toHaveBeenCalledWith('2nd subscribe: 3');
+    expect(console.log).toHaveBeenCalledWith('2nd sequence finished.');
```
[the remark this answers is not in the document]
sonukapoor: Yes, that makes it a lot easier 👍

aio/content/examples/practical-observable-usage/src/typeahead.ts
```diff
@@ -1,18 +1,32 @@
...
+  import { fromEvent } from 'rxjs';
+  import { ajax } from 'rxjs/ajax';
+  import { debounceTime, distinctUntilChanged, filter, map, switchMap } from 'rxjs/operators';
 
+// #enddocregion
+/* tslint:disable:no-shadowed-variable */
+/* tslint:disable:align */
```
gkalpak, line 13: Is this needed? Why?

gkalpak (approved):
Almost ready :tada:

BTW, please update the second commit message to use imperative, present tesne (per our [guidelines](https://github.com/angular/angular/blob/1d9e56eba06958e37ef8062ae7b65afa85f90632/CONTRIBUTING.md#commit)):

`improves typeahead example and adds unit test` --> `improve typeahead example and add unit test …`

aio/content/examples/practical-observable-usage/src/typeahead.ts
sonukapoor: > Is this needed? Why?

Because of the alignment of the imports. Because we are hiding the function declaration using the doc regions the import statements are misaligned with the body of the function. In the below example, the `import` aligns perfectly fine with the body `const bar...`.

```typescript
// #docplaster
// #docregion
  import {} ...
// #enddocregion
export function foo() {
  // #docRegion
  const bar = ...
  // #enddocregion
}
```

gkalpak (approved):
Nice :tada:

PR #34194, opened 2019-12-02
docs: add docs about inputs and strictNullChecks to template typecheck

kapunahelewong:
## PR Checklist

## PR Type

Adds documentation about  inputs and strictNullChecks to template typecheck doc

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local varia
- [ ] Build related changes
- [ ] CI related changes
-[x] Documentation content changes

Currently we don't have documentation on this feature.

Issue Number: N/A

## What is the new behavior?

New Docs.

- [ ] Yes
- [x] No


aio/content/guide/template-typecheck.md
```diff
@@ -129,3 +129,106 @@ If that doesn't work, an option of last resort is to turn off full mode entirely
...
+  const cmp: UserDetailComponent = null!;
+  cmp.user = ctx.user;
+}
+```
+
+Since this code is generated as part of the program in a hidden manner, any TypeScript compilation flags such as `strictNullChecks` will be taken into account during template type checking. In the above example, the assignment `cmp.user = ctx.user;` is not null safe, as `ctx.user` can be `null` whereas `UserDetailComponent`'s `user` input cannot. Hence, if `strictNullChecks` were enabled, a type check error would be produced.
+
+## Libraries and input type-checking
+
+When library directives or components do not allow `null` for inputs, there are a couple of considerations:
```
JoostK: I think we should split this chapter into one dedicated for library authors, and one for library consumers that have to deal with libraries that are compiled without `strictNullChecks`. This should help with providing a clear image for both perspectives, as they are really quite different.

aio/content/guide/template-typecheck.md
```diff
@@ -129,3 +129,106 @@ If that doesn't work, an option of last resort is to turn off full mode entirely
...
+
+## `strictNullInputTypes`
+
+The `async` pipe introduces `null`, so the application itself must insert the `!`, or the library author needs to include `null` in all inputs to allow for the usage of things like the `async` pipe.
+To make this easier, there’s a type check option called `strictNullInputTypes` that will automatically insert the `!`.
+
+<!-- Where does one use `strictNullInputTypes`? Should we tell the reader how? -->
+
+
+## `ngInputAcceptType`
```
JoostK: I noticed recently that [this is already documented pretty well here](https://github.com/angular/angular/blob/fba72f0d4df8c76e029015c742b144f323ee1d56/aio/content/guide/aot-compiler.md#input-setter-coercion). I think it would make most sense to have this information only in a single place, in which case this doc page might be more appropriate.

In any case, these should all be `ngAcceptInputType` (Accept and Input are swapped).

aio/content/guide/template-typecheck.md
```diff
@@ -129,3 +129,106 @@ If that doesn't work, an option of last resort is to turn off full mode entirely
...
+For a consumer application that has `strictNullChecks` enabled and uses a directive from a library, the assignment in the type check code would be invalid if the value could be `null`.
+
+The `async` pipe currently assumes that the Observable it subscribes to can be asynchronous, which means that it's possible that there is no value available yet. In that case, it still has to return something&mdash;which is `null`.
+In other words, the return type of the `async` pipe includes `null`, which with `strictNullChecks` will cause an error.
+
+Consider the case where the template is `<div my-dir [input]="data$ | async">`.
+Now, even when `data$`  is typed as `Observable<string>`, the return type of `asyncPipe.transform(ctx.data$)` would be `string | null`, as the `async` pipe always includes `null` in its return type.
+
+Thus, the `myDir.input`  must include `null`  to avoid a type check error from being detected, or the template would have to be adjusted to be `<div my-dir [input]="(data$ | async)!">`.
+The `!` operator at the end reduces the return type of the `async` pipe from `string | null`  to just `string`, so that it correctly type checks against `myDir.input`.
```
JoostK: I don't think using the `!` operator should be considered a _recommendation_, so we should be careful about how this is presented.

aio/content/guide/template-typecheck.md
kapunahelewong: Agreed. I'll delete this section from this page. Thank you for the correction on the name as well.

aio/content/guide/template-typecheck.md
```diff
@@ -129,3 +129,166 @@ If that doesn't work, an option of last resort is to turn off full mode entirely
...
+
+## `strictNullInputTypes`
+
+The `async` pipe introduces `null`, so the application itself must insert the `!`, or the library author needs to include `null` in all inputs to allow for the usage of things like the `async` pipe.
+To make this easier, there’s a type check option called `strictNullInputTypes` that will automatically insert the `!`.
+
+<!-- Where does one use `strictNullInputTypes`? Should we tell the reader how? -->
+
+
+### Input setter coercion
```
JoostK: Can you make this a `h2`?

aio/content/guide/template-typecheck.md
```diff
@@ -129,3 +129,214 @@ If that doesn't work, an option of last resort is to turn off full mode entirely
...
+Now, even when `data$`  is typed as `Observable<string>`, the return type of `asyncPipe.transform(ctx.data$)` would be `string | null`, as the `async` pipe always includes `null` in its return type.
+
+Thus, the `myDir.input`  must include `null`  to avoid a type check error from being detected, or the template would have to be adjusted to be `<div my-dir [input]="(data$ | async)!">`.
+The `!` operator at the end reduces the return type of the `async` pipe from `string | null`  to just `string`, so that it correctly type checks against `myDir.input`.
+ -->
+
+
+### Strict null checks
+
+When you enable `strictTemplates` and the TypeScript flag `strictNullChecks`, typecheck errors may occur for certain situations. For example:
```
JoostK: I think we lost some information here. The intent of this sentence is to introduce a potential user experience issue related to strict null checks, therefore I originally mentioned "may occur for situations that may not easily be avoided". I think this is important, as currently the sentence makes me believe that both `strictTemplates` and `strictNullChecks` are the only combination of flags that may cause typecheck errors to be produced. This is not the case, e.g. having `strictNullChecks` disabled will also produce errors "for certain situations".

I'm probably being pedantic here, but I hope we can tweak this a bit so that the intent is better communicated.

aio/content/guide/template-typecheck.md
```diff
@@ -129,3 +129,214 @@ If that doesn't work, an option of last resort is to turn off full mode entirely
...
+
+  1. A nullable value that is bound to a directive from a library which did not have `strictNullChecks` enabled.
+
+  When you compile a library without `strictNullChecks`, its declaration files will not indicate whether a field can be `null` or not.
```
JoostK: I tried to avoid referring to "you" here, as this section is not specific to library authors. Hence, this should be applicable for application developers as well, in which case he/she will not be the one compiling a library.

kapunahelewong: Ah, I see. Is this ok?

> When compiling a library without `strictNullChecks`, its declaration files will not indicate whether a field can be `null` or not.

Just trying to minimize the passive voice.

aio/content/guide/template-typecheck.md
```diff
@@ -129,3 +129,214 @@ If that doesn't work, an option of last resort is to turn off full mode entirely
...
+@Component({
+  selector: 'my-app',
+  template: '<user-detail [user]="user" />',
+})
+export class AppComponent {
+  user: User | null = null;
+}
+```
+
+Now, the template type checker generates some code to check the `[user]="user"` binding, this looks roughly as follows:
```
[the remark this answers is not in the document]
kapunahelewong: Ah, thank you. Does this edit still maintain your intent? I have to meet certain grammatical style guidelines:

```
Here, during type checking of the template for `AppComponent`, the `[user]="selectedUser"` binding corresponds with the `UserDetailComponent.user` input.
Therefore, TypeScript assigns the `selectedUser` property to `UserDetailComponent.user`, which would result in an error if their types were incompatible.
TypeScript checks the assignment according to its type system, obeying flags such as `strictNullChecks` as they are configured in the application.
```

aio/content/guide/template-typecheck.md
kapunahelewong: Certainly! Thank you for keeping me on message. I definitely want to keep your meaning intact. I may have been overzealous on the grammar. 🙂

I've put it back.

JoostK (approved):

PR #34195, opened 2019-12-02
`strictTemplates` implies `fullTemplateTypeCheck`

JoostK:
See individual commits.

### DIFF
--- a/aio/content/guide/template-typecheck.md
+++ b/aio/content/guide/template-typecheck.md
@@ -46,7 +46,7 @@ The following still have type `any`.
 ### Strict mode
 
 Angular version 9 maintains the behavior of the `fullTemplateTypeCheck` flag, and introduces a third "strict mode".
-Strict mode is accessed by setting both `fullTemplateTypeCheck` and the `strictTemplates` flag to `true`.
+Strict mode is a superset of full mode, and is accessed by setting the `strictTemplates` flag to true. This flag supersedes the `fullTemplateTypeCheck` flag.
 In strict mode, Angular version 9 adds checks that go beyond the version 8 type-checker.
 Note that strict mode is only available if using Ivy.
 
--- a/packages/compiler-cli/src/ngtsc/program.ts
+++ b/packages/compiler-cli/src/ngtsc/program.ts
@@ -462,9 +462,15 @@ export class NgtscProgram implements api.Program {
   }
 
   private getTemplateDiagnostics(): ReadonlyArray<ts.Diagnostic> {
+    // Determine the strictness level of type checking based on compiler options. As
+    // `strictTemplates` is a superset of `fullTemplateTypeCheck`, the former implies the latter.
+    // Also see `verifyCompatibleTypeCheckOptions` where it is verified that `fullTemplateTypeCheck`
+    // is not disabled when `strictTemplates` is enabled.
+    const strictTemplates = !!this.options.strictTemplates;
+    const fullTemplateTypeCheck = strictTemplates || !!this.options.fullTemplateTypeCheck;
+
     // Skip template type-checking if it's disabled.
-    if (this.options.ivyTemplateTypeCheck === false &&
-        this.options.fullTemplateTypeCheck !== true) {
+    if (this.options.ivyTemplateTypeCheck === false && !fullTemplateTypeCheck) {
       return [];
     }
 
@@ -475,8 +481,7 @@ export class NgtscProgram implements api.Program {
     // First select a type-checking configuration, based on whether full template type-checking is
     // requested.
     let typeCheckingConfig: TypeCheckingConfig;
-    if (this.options.fullTemplateTypeCheck) {
-      const strictTemplates = !!this.options.strictTemplates;
+    if (fullTemplateTypeCheck) {
       typeCheckingConfig = {
         applyTemplateContextGuards: strictTemplates,
         checkQueries: false,
--- a/packages/compiler-cli/src/ngtsc/program.ts
+++ b/packages/compiler-cli/src/ngtsc/program.ts
@@ -82,6 +82,11 @@ export class NgtscProgram implements api.Program {
       verifySupportedTypeScriptVersion();
     }
 
+    const incompatibleTypeCheckOptionsDiagnostic = verifyCompatibleTypeCheckOptions(options);
+    if (incompatibleTypeCheckOptionsDiagnostic !== null) {
+      this.constructionDiagnostics.push(incompatibleTypeCheckOptionsDiagnostic);
+    }
+
     if (shouldEnablePerfTracing(options)) {
       this.perfTracker = PerfTracker.zeroedToNow();
       this.perfRecorder = this.perfTracker;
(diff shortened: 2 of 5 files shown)


packages/compiler-cli/src/ngtsc/program.ts
```diff
@@ -746,6 +752,28 @@ function isAngularCorePackage(program: ts.Program): boolean {
...
 }
 
+/**
+ * Since "strictTemplates" is a true superset of type checking capabilities compared to
+ * "strictTemplateTypeCheck", it is required that the latter is not explicitly disabled if the
+ * former is enabled.
+ */
+function verifyCompatibleTypeCheckOptions(options: api.CompilerOptions): void {
+  if (options.fullTemplateTypeCheck === false && options.strictTemplates === true) {
+    throw new Error(
```
alxhub: We could emit this as a diagnostic in `getNgOptionDiagnostics`. I've never done an option diagnostic though - might be worth a shot.

alxhub (approved):

AndrewKushnir:
[Presubmit](http://test/OCL:284054900:BASE:284054948:1575585060622:e6005c19)

aio/content/guide/template-typecheck.md
```diff
@@ -46,7 +46,8 @@ The following still have type `any`.
 ### Strict mode
 
 Angular version 9 maintains the behavior of the `fullTemplateTypeCheck` flag, and introduces a third "strict mode".
-Strict mode is accessed by setting both `fullTemplateTypeCheck` and the `strictTemplates` flag to `true`.
+Strict mode is accessed by setting the `strictTemplates` flag to `true`, which implies that `fullTemplateTypeCheck` is also enabled.
```
IgorMinar: I'd simplify this to just say:

```suggestion
Strict mode is the a superset of full mode, and is accessed by setting the `strictTemplates` flag to true. This flag supersedes the `fullTemplateTypeCheck` flag.
```

aio/content/guide/template-typecheck.md
```diff
@@ -46,7 +46,8 @@ The following still have type `any`.
...
+It is therefore OK to not specify `fullTemplateTypeCheck` or to set it to `true`. Setting `fullTemplateTypeCheck` to `false` in this mode is an error.
```
IgorMinar: There is no need to talk about the error case. The error will be thrown when users use the settings incorrectly.
```suggestion
```

packages/compiler-cli/src/ngtsc/program.ts
```diff
@@ -427,9 +432,15 @@ export class NgtscProgram implements api.Program {
...
+    // `strictTemplates` is a superset of `fullTemplateTypeCheck`, the former implies the latter.
+    // Also see `verifyCompatibleTypeCheckOptions` where it is verified that `fullTemplateTypeCheck`
+    // is not disabled when `strictTemplates` is enabled.
+    const strictTemplates = !!this.options.strictTemplates;
+    const fullTemplateTypeCheck = strictTemplates || !!this.options.fullTemplateTypeCheck;
+
     // Skip template type-checking if it's disabled.
-    if (this.options.ivyTemplateTypeCheck === false &&
-        this.options.fullTemplateTypeCheck !== true) {
+    if (this.options.ivyTemplateTypeCheck === false && !fullTemplateTypeCheck) {
```
IgorMinar: I find this condition very misleading and confusing. I always have even before this change.

How about:
```
const templateTypeCheck = (strictTemplates || !!this.options.fullTemplateTypeCheck) && this.options.ivyTemplateTypeCheck !== false;

if (!templateTypeCheck) {
  return [];
}
```

packages/compiler-cli/src/ngtsc/program.ts
```diff
@@ -743,6 +753,37 @@ function isAngularCorePackage(program: ts.Program): boolean {
...
+      length: undefined,
+      messageText:
+          `Angular compiler option "strictTemplates" is enabled, however "fullTemplateTypeCheck" is disabled.
+
+Having the "strictTemplates" flag enabled implies that "fullTemplateTypeCheck" is also enabled, so
+the latter can not be explicitly disabled.
+
+One of the following actions is required:
+1. Enable "fullTemplateTypeCheck" by setting it to 'true', or by removing the option entirely.
+2. Disable "strictTemplates" by setting it to 'false', or by removing the option entirely.
```
IgorMinar: I don't think that it makes sense to encourage people to set both strictTemplates and fullTemplateTypeCheck in the config. Fewer flags is better, especially if we in the future want to remove the "fullTemplateTypeCheck" flag.

Can you please change this, so that the two options are:

1. remove the "fullTemplateTypeCheck" option.
2. remove "strictTemplates" or set it to "false".

packages/compiler-cli/test/ngtsc/template_typecheck_spec.ts
```diff
@@ -245,7 +245,7 @@ export declare class AnimationEvent {
       });
 
       it('should check expressions and their type when overall strictness is enabled', () => {
-        env.tsconfig({fullTemplateTypeCheck: true, strictTemplates: true});
+        env.tsconfig({strictTemplates: true});
```
IgorMinar: much better!!!

IgorMinar (dismissed):
I suggest we clean up the docs a bit more. See individual commits.

thanks for making this change.

packages/compiler-cli/src/ngtsc/program.ts
JoostK: You are not alone, it confuses me just as much as it confuses you :-)

Your suggestion, however, is not equivalent to the current logic. Currently, even if "ivyTemplateTypeCheck" is false will typechecking still be enabled if "fullTemplateTypeCheck"/"strictTemplates" is set.  Your suggestion as I interpret it would not behave the same, as having "ivyTemplateTypeCheck" set to false would completely disable type-checking under all circumstances.

JoostK:
@IgorMinar thanks for the review, your suggestions have been applied.

packages/compiler-cli/src/ngtsc/program.ts
alxhub: I'm okay with the behavior change. If you're curious, the logic worked this way because `fullTemplateTypeCheck` is easy to control as an option of the `ng_module` build rule. So I could set `ivyTemplateTypeCheck` to `false` in bazel/g3, and then set `fullTemplateTypeCheck` via the `ng_module` option when I wanted to test type-checking in g3 projects.

This logic predates the `strictTemplates` though - now setting `fullTemplateTypeCheck` isn't enough. So I'm happy to change it.

JoostK: @alxhub Can we get rid of `ivyTemplateTypeCheck` entirely? non-fTTC should be at the same level asa it was in VE, so there's shouldn't be a special flag for disabling TTC in g3?

alxhub: I've been looking into what it would take to do this in g3, and there's a number of issues preventing it at the moment. So I think we should wait for now.

alxhub:
[Presubmit](https://test.corp.google.com/ui#id=OCL:286650954:BASE:286650985:1576884867769:17a4afe5)
[Ivy Presubmit](https://test.corp.google.com/ui#id=OCL:286650954:BASE:286650985:1576884867769:8deaed0d)

PR #34200, opened 2019-12-02
fix(ivy): correctly support `ngProjectAs` on templates

### LINKED ISSUE #34120: ngProjectAs not working in Ivy v9 rc4
# 🐞 bug report

### Affected Package

### Is this a regression?
I do not know.

Did this behavior use to work in the previous version?
Yes, v8.2.14 ok

### Description
version 8 all slot is correct
v8-good

version 9 all slot is not correct 
v9-bad

above are using same code. 

## 🔬 Minimal Reproduction
v8: https://github.com/keatkeat87/ng-project-as-issue-v8
v9: https://github.com/keatkeat87/ng-project-as-issue-v9
git clone 
ng serve --open
click button

## 🌍  Your Environment
**Angular Version:**
Angular CLI: 9.0.0-rc.4
Node: 12.3.1
OS: win32 x64

Angular:
...
Ivy Workspace:

Package                      Version
------------------------------------------------------
@angular-devkit/architect    0.900.0-rc.4
@angular-devkit/core         9.0.0-rc.4
@angular-devkit/schematics   9.0.0-rc.4
@schematics/angular          9.0.0-rc.4
@schematics/update           0.900.0-rc.4
rxjs                         6.5.3

AndrewKushnir:
Prior to this commit, if a template (for example, generated using structural directive such as *ngIf) contains `ngProjectAs` attribute, it was not included into attributes array in generated code and as a result, these templates were not matched at runtime during content projection. This commit adds the logic to append `ngProjectAs` values into corresponding element's attribute arrays, so content projection works as expected.

This PR resolves #34120.

## PR Type

- [x] Bugfix

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/test/acceptance/content_spec.ts
+++ b/packages/core/test/acceptance/content_spec.ts
@@ -1165,6 +1165,77 @@ describe('projection', () => {
       expect(fixture.nativeElement).toHaveText('inline()ng-template(onetwothree)');
     });
 
+    it('should project template content with `ngProjectAs` defined', () => {
+      @Component({
+        selector: 'projector-app',
+        template: `
+            App selector: <ng-content select="app-selector"></ng-content>
+            Attribute-based: <ng-content select="[foo]"></ng-content>
+            Class-based: <ng-content select=".bar"></ng-content>
+            Other: <ng-content></ng-content>
+          `,
+      })
+      class ProjectorApp {
+      }
+
+      @Component({
+        selector: 'app-selector',
+        template: ' App {{ marker }} ',
+      })
+      class ChildApp {
+        @Input() marker: string = '';
+      }
+
+      @Component({
+        selector: 'root-comp',
+        template: `
+          <projector-app>
+            <div>Text content with no selectors</div>
+
+            <div *ngIf="show" foo> Attr A </div>
+            <ng-container *ngIf="show" ngProjectAs="[foo]">
+              <div foo> Attr B </div>
+            </ng-container>
+
+            <div *ngIf="show" class="bar"> Class A </div>
+            <ng-container *ngIf="show" ngProjectAs=".bar">
+              <div class="bar"> Class B </div>
+            </ng-container>
+
+            <app-selector *ngIf="show" marker="A"></app-selector>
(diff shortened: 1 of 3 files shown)


packages/core/test/acceptance/content_spec.ts
```diff
@@ -1165,6 +1165,77 @@ describe('projection', () => {
...
+            <ng-container *ngIf="show" ngProjectAs="[foo]">
+              <div foo> Attr B </div>
+            </ng-container>
+
+            <div *ngIf="show" class="bar"> Class A </div>
+            <ng-container *ngIf="show" ngProjectAs=".bar">
+              <div class="bar"> Class B </div>
+            </ng-container>
+
+            <app-selector *ngIf="show" marker="A"></app-selector>
```
pkozlowski-opensource: Do we need ` marker="A"` and ` marker="B"` attributes for this test?

packages/core/test/acceptance/content_spec.ts
```diff
@@ -1165,6 +1165,77 @@ describe('projection', () => {
...
+      }
+
+      @Component({
+        selector: 'root-comp',
+        template: `
+          <projector-app>
+            <div>Text content with no selectors</div>
+
+            <div *ngIf="show" foo> Attr A </div>
+            <ng-container *ngIf="show" ngProjectAs="[foo]">
```
pkozlowski-opensource: Do we need `<ng-container>` here? (that is - is it essential for the test to have it? If not I would drop it since my understanding is that the case tests here is to make sure that `<div *ngIf="show" foo>` behaves the same as `<div *ngIf="show"ngProjectAs="[foo]">`

pkozlowski-opensource (approved):
LGTM overall from the runtime point of view, but left some nit comments that I would like to see resolved before this PR gets merged.

AndrewKushnir:
Thanks for review @pkozlowski-opensource 👍 I've updated (simplified) tests based on your feedback.

AndrewKushnir:
[Presubmit](http://test/OCL:283588285:BASE:283588349:1575401892963:82056531)

alxhub (approved):

PR #34206, opened 2019-12-03
refactor(common): add defaults to new generic parameters

crisbeto:
This is a follow-up to #33997 where some new generic parameters were added without defaults which is technically a breaking change. These changes add the defaults.

### DIFF
--- a/packages/common/src/directives/ng_for_of.ts
+++ b/packages/common/src/directives/ng_for_of.ts
@@ -11,7 +11,7 @@ import {Directive, DoCheck, EmbeddedViewRef, Input, IterableChangeRecord, Iterab
 /**
  * @publicApi
  */
-export class NgForOfContext<T, U extends NgIterable<T>> {
+export class NgForOfContext<T, U extends NgIterable<T> = NgIterable<T>> {
   constructor(public $implicit: T, public ngForOf: U, public index: number, public count: number) {}
 
   get first(): boolean { return this.index === 0; }
@@ -121,7 +121,7 @@ export class NgForOfContext<T, U extends NgIterable<T>> {
  * @publicApi
  */
 @Directive({selector: '[ngFor][ngForOf]'})
-export class NgForOf<T, U extends NgIterable<T>> implements DoCheck {
+export class NgForOf<T, U extends NgIterable<T> = NgIterable<T>> implements DoCheck {
   /**
    * The value of the iterable expression, which can be used as a
    * [template input variable](guide/structural-directives#template-input-variable).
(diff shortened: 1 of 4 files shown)


packages/common/src/directives/ng_for_of.ts
```diff
@@ -11,7 +11,7 @@ import {Directive, DoCheck, EmbeddedViewRef, Input, IterableChangeRecord, Iterab
 /**
  * @publicApi
  */
-export class NgForOfContext<T, U extends NgIterable<T>> {
+export class NgForOfContext<T, U extends NgIterable<T> = any> {
```
JoostK: We should be able to use `NgIterable<T>` as default type, here and in all other places:

```suggestion
export class NgForOfContext<T, U extends NgIterable<T> = NgIterable<T>> {
```

JoostK (commented):
Could you also update the template type check type definitions?

packages/common/src/directives/ng_for_of.ts
crisbeto: Done.

JoostK (approved):

IgorMinar (commented):
Could we set the default to be `unknown` rather than `any`?

crisbeto:
Done.

packages/common/src/directives/ng_if.ts
```diff
@@ -149,7 +149,7 @@ import {Directive, EmbeddedViewRef, Input, TemplateRef, ViewContainerRef, ɵstri
  * @publicApi
  */
 @Directive({selector: '[ngIf]'})
-export class NgIf<T> {
+export class NgIf<T = unknown> {
```
gkalpak: Have you verified that this avoids the breaking change?

crisbeto: I don't have a way of verifying it, but it should, given that the error is about a required parameter not being defined.

gkalpak: I used the [build artifacts from this PR][1] to test on `ngcc-validation` (where we got a [failure due to #33997][2]) and it passed: angular/ngcc-validation#656. But then I realized that it only exercises the `NgForOfContext` change (whose type does not default to `unknown`) 🤷‍♂ 

Since I don't know how to verify this either, let's assume it's fine until proven otherwise :grin:

[1]: https://circleci.com/gh/angular/angular/551837#artifacts/containers/0
[2]: https://circleci.com/gh/angular/ngcc-validation/2188#tests/containers/1

mhevery (approved):

mhevery:
[presubmit](http://test/OCL:283605236:BASE:283605318:1575407045550:81a4316)

PR #34207, opened 2019-12-03
refactor(platform-browser): avoid mutable exports.

mprobst:
Previously, browser_util would export a mutable `let` binding that was
initialized as a side-effect of `BrowserDetection.setup()`. This change
refactors the mutable binding into a `const` binding that is immediately
initialized in its initialized.

This is functionally equivalent, but makes it easier for module
optimizers such as Closure Compiler to track down side effects and prune
modules. It is also arguably cleaner to read (no worries about later
changes to the apparently mutable but effectively const binding).


mhevery (approved):

mhevery:
[presubmit](http://test/OCL:283635944:BASE:283636025:1575415966446:79028709)

PR #34211, opened 2019-12-03
i18n: guard `ngI18nClosureMode`

petebacondarwin:
cc @filipesilva and @mhevery

### DIFF
--- a/packages/core/src/application_module.ts
+++ b/packages/core/src/application_module.ts
@@ -49,7 +49,8 @@ export function _localeFactory(locale?: string): string {
  * * Ivy enabled: use `$localize.locale`
  */
 export function getGlobalLocale(): string {
-  if (ngI18nClosureMode && typeof goog !== 'undefined' && goog.LOCALE !== 'en') {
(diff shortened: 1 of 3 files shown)


packages/core/src/application_module.ts
```diff
@@ -49,7 +49,8 @@ export function _localeFactory(locale?: string): string {
  * * Ivy enabled: use `$localize.locale`
  */
 export function getGlobalLocale(): string {
-  if (ngI18nClosureMode && typeof goog !== 'undefined' && goog.LOCALE !== 'en') {
+  if (typeof ngI18nClosureMode === 'undefined' && ngI18nClosureMode &&
```
filipesilva: Shouldn't this be `typeof ngI18nClosureMode !== 'undefined'` (`===` -> `!==`)?

filipesilva (approved):
LGTM aside from the one comment.

packages/core/src/application_module.ts
petebacondarwin: !! Yes !! Doh.

mhevery:
[presubmit](http://test/OCL:283572159:BASE:283572210:1575397546661:3a955d2f)

PR #34213, opened 2019-12-03
fix(docs-infra): avoid version mismatch error when using local Angular packages

### LINKED ISSUE #34208: AIO local build script requires tags to be fetched
To build the AIO content locally you have to run `yarn setup-local` and then `yarn start` to start the app, however `yarn start` will fail with an error like the following if the repo's tags are out of date:

```
This version of CLI is only compatible with Angular versions ^9.0.0-beta || >=9.0.0 <10.0.0,
but Angular version 2.3.0 was found instead.
Please visit the link below to find instructions on how to update Angular.
https://update.angular.io/
```

We should have something in the documentation about the need for tags to be up-to-date, or have a workaround that doesn't require the tags at all.

gkalpak:
The local Angular packages used to build `aio/` when running `yarn setup-local`/`yarn build-local` (and related commands), are built by bazel. Bazel, determines the version to use for these locally built packages based on the latest tag for a commit on the current branch. (This works as expected during the release, because the packages are built on the correct branch with up-to-date tags.)

During local development, however, this often leads to generating older versions than what the current `@angular/cli` version is compatible with, if the user has not fetched the latest tags from `angular/angular` or the branch has not been rebased recently.

Previously, the above (common) situation would result in a build error by the CLI. (Note that this would be a false error, because in this case the version set by bazel would not reflect the actual version of the local packages.) The solution would be for the user to fetch the latest tags from `angular/angular`, rebase their branch and run a bazel build again (ensuring that it would actually build anew and not emit cached artifacts). This was cumbersome and most people didn't even know about it.

This commit avoids this error and the associated pain-points by overwriting the versions of the installed local packages with fake versions based on the ones in the lockfile, which are guaranteed to be compatible with the currently used CLI version.

Fixes #34208.

### DIFF
--- a/aio/tools/ng-packages-installer/index.js
+++ b/aio/tools/ng-packages-installer/index.js
@@ -70,8 +70,12 @@ class NgPackagesInstaller {
   installLocalDependencies() {
     if (this.force || !this._checkLocalMarker()) {
       const pathToPackageConfig = path.resolve(this.projectDir, PACKAGE_JSON);
+      const packageConfigFile = fs.readFileSync(pathToPackageConfig, 'utf8');
+      const packageConfig = JSON.parse(packageConfigFile);
+
       const pathToLockfile = path.resolve(this.projectDir, YARN_LOCK);
       const parsedLockfile = this._parseLockfile(pathToLockfile);
+
       const packages = this._getDistPackages();
 
       try {
@@ -96,11 +100,11 @@ class NgPackagesInstaller {
             });
           });
 
+          // Overwrite the package's version to avoid version mismatch errors with the CLI.
+          this._overwritePackageVersion(key, tmpConfig, packageConfig, parsedLockfile);
+
           fs.writeFileSync(pkg.packageJsonPath, JSON.stringify(tmpConfig, null, 2));
         });
-
-        const packageConfigFile = fs.readFileSync(pathToPackageConfig, 'utf8');
-        const packageConfig = JSON.parse(packageConfigFile);
 
         const [dependencies, peers] = this._collectDependencies(packageConfig.dependencies || {}, packages);
         const [devDependencies, devPeers] = this._collectDependencies(packageConfig.devDependencies || {}, packages);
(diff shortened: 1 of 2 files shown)


aio/tools/ng-packages-installer/index.js
```diff
@@ -265,6 +269,35 @@ class NgPackagesInstaller {
...
+   * @param {import('@yarnpkg/lockfile').LockFileObject} projectLockfile - The projects's parsed `yarn.lock`.
+   */
+  _overwritePackageVersion(packageName, packageConfig, projectConfig, projectLockfile) {
+    const originalVersionRange = (projectConfig.dependencies || {})[packageName] ||
+                                 (projectConfig.devDependencies || {})[packageName];
+    const originalVersion =
+      (projectLockfile[`${packageName}@${originalVersionRange}`] || {}).version;
+
+    if (originalVersion !== undefined) {
+      const newVersion = `${originalVersion}+locally-overwritten-by-ngPackagesInstaller`;
```
petebacondarwin: Nice!

aio/tools/ng-packages-installer/index.js
```diff
@@ -265,6 +269,35 @@ class NgPackagesInstaller {
...
+   * @param {string} packageName - The name of the package we are updating (e.g. `'@angular/core'`).
+   * @param {{[key: string]: any}} packageConfig - The package's parsed `package.json`.
+   * @param {{[key: string]: any}} projectConfig - The project's parsed `package.json`.
+   * @param {import('@yarnpkg/lockfile').LockFileObject} projectLockfile - The projects's parsed `yarn.lock`.
+   */
+  _overwritePackageVersion(packageName, packageConfig, projectConfig, projectLockfile) {
+    const originalVersionRange = (projectConfig.dependencies || {})[packageName] ||
+                                 (projectConfig.devDependencies || {})[packageName];
+    const originalVersion =
+      (projectLockfile[`${packageName}@${originalVersionRange}`] || {}).version;
```
petebacondarwin: `originalversion` is a bit confusing, because I thought "original" would be what is in `packageConfig.version` at the start of this function...

petebacondarwin (approved):

aio/tools/ng-packages-installer/index.js
gkalpak: How about `projectVersion` or `originalProjectVersion`?

petebacondarwin: Yeah `projectVersion` sounds good.

gkalpak: Ooops, too late :smiley:
I'll make a follow-up PR.

gkalpak: For reference: #34284

Issue #33141, opened 2019-10-14
docs: Creating Libraries Guide should include guidance on CSS/assets


Splaktar:
# 📚 Docs or angular.io bug report

### Description
I spent some time in January working on a new library. One thing that is missing from the new Library guide, and not supported out of the box (but generally required by most library authors) in the Angular CLI or ng-packagr, is shipping assets or theme SCSS files with your library. This issue (https://github.com/angular/angular-cli/issues/11071) and many other linked/closed issues with hundreds of 👍  indicate that this is a major need and at minimum that the docs to work around this are not available.

It appears that the general guidance for SCSS is to use the `scss-bundle` package similar to [what Angular Material does](https://github.com/angular/components/blob/aaa656dbab41fee298244cc25351da8f44e9f614/tools/BUILD.bazel#L16-L24). For assets, it appears that manually copying them via npm scripts is the most common approach.

I can understand some hesitation to include details about library-specific implementation that could cause some kind of support burden.

However, the other perspective is that library authors see the docs and get started and then get completely stuck on assets and scss files (like many hundreds have already run into), etc with no references for how to proceed. They will then continue to open duplicate issues in the Angular CLI GitHub.

Perhaps there is some kind of happy medium between providing them with some starting points on how to handle this without providing them a fully supported solution that doesn’t exist yet. Even telling/reminding people that you can just use regular command line copy commands to move assets via NPM scripts would help people rather than them having to dig through a lot of closed GitHub issues to find examples that work.

### What's the affected URL?**
- https://next.angular.io/guide/creating-libraries
- https://angular.io/guide/creating-libraries

@jbogarthyde I haven't had time to write up some proposed copy for this, but I wanted to at least create the issue.

santoshyadav198613:
This is much needed, I also struggled with the same. I will love to help in documentation, so can we add how to use scss-bunlde @Splaktar .

santoshyadav198613:
Hi @kapunahelewong,
I will start working on this.

santoshyadav198613:
Just raised a WIP PR to move further discussion.

Splaktar:
Closed in https://github.com/angular/angular/commit/26b17aee1441534b554862129df308c780169c5f via PR https://github.com/angular/angular/pull/34217.

PR #34217, opened 2019-12-03
Add asset info to library guide

jbogarthyde:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

No mention in docs of possible ways to add assets to a library

Issue Number: https://github.com/angular/angular/issues/33141

## What is the new behavior?
Adds a section on assets to https://angular.io/guide/creating-libraries

- [ ] Yes
- [x] No

### DIFF
--- a/aio/content/guide/creating-libraries.md
+++ b/aio/content/guide/creating-libraries.md
@@ -119,6 +119,7 @@ npm publish
 
 If you've never published a package in npm before, you must create a user account. Read more in [Publishing npm Packages](https://docs.npmjs.com/getting-started/publishing-npm-packages).
 
+
 ## Linked libraries
 
 While working on a published library, you can use [npm link](https://docs.npmjs.com/cli/link) to avoid reinstalling the library on every build.
@@ -155,6 +156,7 @@ List all the peer dependencies that your library uses in the workspace TypeScrip
 ```
 
 This mapping ensures that your library always loads the local copies of the modules it needs.
+
 
 ## Using your own library in apps
 
@@ -213,3 +215,14 @@ For this reason, an app that depends on a library should only use TypeScript pat
 TypeScript path mappings should *not* point to the library source `.ts` files.
 
 </div>
+
+{@a lib-assets}
+
+### Managing library assets with ng-packagr
+
+Starting with version 9.x of the [ng-packagr](https://github.com/ng-packagr/ng-packagr/blob/master/README.md) tool, you can configure the tool to automatically copy assets into your library package as part of the build process.
+You can use this feature when your library needs to publish optional theming files, Sass mixins, or documentation (like a change log).
+
+* Learn how to [copy assets into your library as part of the build](https://github.com/ng-packagr/ng-packagr/blob/master/docs/copy-assets.md).
+
+* Learn more about how to use the tool to [embed assets in CSS](https://github.com/ng-packagr/ng-packagr/blob/master/docs/embed-assets-css.md).
--- a/aio/content/guide/creating-libraries.md
+++ b/aio/content/guide/creating-libraries.md
@@ -221,7 +221,7 @@ TypeScript path mappings should *not* point to the library source `.ts` files.
 ### Managing library assets with ng-packagr
 
 Starting with version 9.x of the [ng-packagr](https://github.com/ng-packagr/ng-packagr/blob/master/README.md) tool, you can configure the tool to automatically copy assets into your library package as part of the build process.
-You can use this feature when your library needs to publish optional theming files, Sass mixins, or documentation (like a change log).
+You can use this feature when your library needs to publish optional theming files, Sass mixins, or documentation (like a changelog).
 
 * Learn how to [copy assets into your library as part of the build](https://github.com/ng-packagr/ng-packagr/blob/master/docs/copy-assets.md).
 


jbogarthyde:
@Splaktar Please take a look and help me fill this in properly.

aio/content/guide/creating-libraries.md
```diff
@@ -156,6 +156,27 @@ List all the peer dependencies that your library uses in the workspace TypeScrip
 
 This mapping ensures that your library always loads the local copies of the modules it needs.
 
+{@a lib-assets}
+
+## Including assets with libraries
```
alan-agius4: From devrel perspective how do we usually communicate something like this?

That being, while this is documented it’s not something that we support.

//cc @mgechev

aio/content/guide/creating-libraries.md
```diff
@@ -156,6 +156,27 @@ List all the peer dependencies that your library uses in the workspace TypeScrip
...
+
+As a library author, you can process style files in any of the supported formats (CSS, SCSS, or SASS) and generate CSS files to be included with your library.
+There are various ways to do this.
+Angular Material, for example, uses a utility called [scss-bundle](https://www.npmjs.com/package/scss-bundle) to generate multiple CSS files.
+
+To use this tool, load the npm package and create a configuration file named `scss-bundle.config.json`.
```
alan-agius4: I don’t think this description is correct. Scss-bundle doesn’t generate an CSS. But rather bundles SCSS and Sass into a single file.

aio/content/guide/creating-libraries.md
```diff
@@ -156,6 +156,27 @@ List all the peer dependencies that your library uses in the workspace TypeScrip
...
+The configuration tells the tool where to get the bundle and where to put the generated CSS, and lets you provide options for how to generate the styles.
+The options in this file can be overridden by similar options in the CLI configuration file, `angular.json`. For details, see the [package documentation](https://www.npmjs.com/package/scss-bundle#configuration).
```
alan-agius4: I am not following why this line is needed and comparing it with angular.json

They are 2 unrelated tools.

aio/content/guide/creating-libraries.md
Splaktar: In what sense is this "not supported"? You just want to call out that `scss-bundle` is not owned and maintained by the Angular team?

aio/content/guide/creating-libraries.md
Splaktar: Yeah, I think that this line (169) should be removed.

aio/content/guide/creating-libraries.md
```diff
@@ -156,6 +156,27 @@ List all the peer dependencies that your library uses in the workspace TypeScrip
 
 This mapping ensures that your library always loads the local copies of the modules it needs.
 
+{@a lib-assets}
+
+## Including assets with libraries
+
+As a library author, you can process style files in any of the supported formats (CSS, SCSS, or SASS) and generate CSS files to be included with your library.
```
Splaktar: We may not want to list the formats here since I believe there are other supported formats than what we list here.

Splaktar: This isn't about generating CSS, it's about bundling SCSS files with your library. These can then be consumed by applications which will handle the generation of CSS from the combination of their SCSS and the library's bundled SCSS.

aio/content/guide/creating-libraries.md
```diff
@@ -156,6 +156,27 @@ List all the peer dependencies that your library uses in the workspace TypeScrip
...
+There are various ways to do this.
+Angular Material, for example, uses a utility called [scss-bundle](https://www.npmjs.com/package/scss-bundle) to generate multiple CSS files.
```
Splaktar:
```suggestion
Angular Material, for example, uses a utility called [scss-bundle](https://www.npmjs.com/package/scss-bundle) to include multiple SCSS theme bundles with their published library.
```

aio/content/guide/creating-libraries.md
```diff
@@ -156,6 +156,27 @@ List all the peer dependencies that your library uses in the workspace TypeScrip
...
+
+To use this tool, load the npm package and create a configuration file named `scss-bundle.config.json`.
+The configuration tells the tool where to get the bundle and where to put the generated CSS, and lets you provide options for how to generate the styles.
```
Splaktar:
```suggestion
Learn more about [`scss-bundle` configuration](https://github.com/reactway/scss-bundle#configuration).
```

aio/content/guide/creating-libraries.md
Splaktar:
```suggestion
```

aio/content/guide/creating-libraries.md
```diff
@@ -156,6 +156,27 @@ List all the peer dependencies that your library uses in the workspace TypeScrip
...
+The options in this file can be overridden by similar options in the CLI configuration file, `angular.json`. For details, see the [package documentation](https://www.npmjs.com/package/scss-bundle#configuration).
+
+When you have created the configuration file, add a script to invoke the tool.
+
+<code-example language="html" header="??name of file??">
+"build-css" : "scss-bundle -p $lt;project-path-to-scss-bundle.config.json>"
```
Splaktar:
```suggestion
"build:scss-bundle" : "scss-bundle -p $lt;project-path-to-scss-bundle.config.json>"
```

aio/content/guide/creating-libraries.md
```diff
@@ -156,6 +156,27 @@ List all the peer dependencies that your library uses in the workspace TypeScrip
...
+</code-example>
+
+A [custom CLI builder](https://github.com/linnenschmidt/build-ng-packagr) is also available that will automatically copy the assets folder created by the bundler into your library package, as part of the build process.
```
Splaktar: I'm not sure that we should include this here as it isn't directly related to `scss-bundle` and isn't as well supported/established.

Splaktar (changes requested):

aio/content/guide/creating-libraries.md
mgechev: We can add a warning box stating that this is not supported by the team, but an approach provided by a third-party library.

alan-agius4: Something like that, my point is that using a package might cause undefined behaviour, example unless changed using this package will opt you out of using sass and will force you to use the legacy node-sass since it has direct dependency on node-sass. This will have a effect on all workspace projects which might break currently working project. 

Ps: will need to confirm if they still use node sass.

Also, afaik previously we reframing from adding certain documents in aio for “unsupported” reasons, such as HMR, bootstrap etc...

alan-agius4: Maybe we should revisit and add documents such as how to configure HMR etc...

aio/content/guide/creating-libraries.md
jelbourn: I agree that it's better to not enumerate the formats here and instead link to the canonical docs for support style formats.

aio/content/guide/creating-libraries.md
alan-agius4: I agree, and using a custom builder has its cavets.

The library author will need to create migrations script for each majoy angular version.

This is because certain cli migrations will only be executed if the builders are from @angular-devkit/build-angular. The main reason for this is because if we don’t do this we might break users since custom builders might have all sort of options which we don’t know of.

aio/content/guide/creating-libraries.md
jelbourn: I'm iffy about whether we want to call attention to this. We only use `scss-bundle` at this point because it improves the performance of downstream Sass compilations (since the compiler doesn't have to process multiple files). However, the Sass team is working on fixing performance issues around file processing, so this may become unnecessary within the next six months. 

For most libraries, they'd probably be fine just including their Sass sources in the package output directly.

jelbourn: This also isn't exactly right; we use `scss-bundle` to combine _multiple_ Sass partial files into _one_ Sass partial file. Our actual theme css files are just compiled directly with the Sass compiler (via Bazel).

alan-agius4: I think i agree with this. I think what should be done is that we should add something similar to `srcs` attribute of the `npm_package` Bazel rule to `ng-packagr` the package to allow basic copying of files. 

This should be trivial to implement.

//cc @Splaktar what do you think?

Splaktar: Oh I agree that it would be much preferred to add support for this to ng-packagr rather than documenting a workaround! However, the issue has been open for a couple years now with ng-packagr and doesn't appear to be getting any attention (other than people asking for it).

alan-agius4: Now it has 😜 can you send me the issue number over slack? 

If it’s kept simple I’ll should be able to do something in the next couple of days.

alan-agius4: https://github.com/ng-packagr/ng-packagr/pull/1465

jbogarthyde:
@Splaktar @jelbourn  Now that Alan has fixed `ng-packagr` , what should the "Creating Libs" page say about assets?
Since so many people have been having trouble with it for so long, is seems like we should say something, if only to reassure them that all they need to do is update to the new release.

jelbourn:
Seems like documenting the ng-packagr API to do this would be the right thing to do, maybe adding a note like "Available as of version xxx"

jbogarthyde:
@jelbourn Does the existing `ng-packagr` doc provide enough information? 
The doc I've pointed to is old, does it need to be updated for Alan's latest change? 
Is it even the right doc?

Which functionality, exactly, is available as of v9?

aio/content/guide/creating-libraries.md
```diff
@@ -119,6 +119,14 @@ npm publish
 
 If you've never published a package in npm before, you must create a user account. Read more in [Publishing npm Packages](https://docs.npmjs.com/getting-started/publishing-npm-packages).
 
+{@a lib-assets}
+
+### Including assets with libraries
+
+The [ng-packagr](https://github.com/ng-packagr/ng-packagr/blob/master/README.md) builder can automatically copy assets into your library package as part of the build process.
+Learn more about how to [embed assets in CSS](https://github.com/ng-packagr/ng-packagr/blob/master/docs/embed-assets-css.md).
```
alan-agius4: This doc is related to a bit of a different topic. It’s how to embed an asset in a style sheet.

For the “feature” that is mentioned here, there are no docs yet.

aio/content/guide/creating-libraries.md
```diff
@@ -119,6 +119,14 @@ npm publish
 
 If you've never published a package in npm before, you must create a user account. Read more in [Publishing npm Packages](https://docs.npmjs.com/getting-started/publishing-npm-packages).
 
+{@a lib-assets}
+
+### Including assets with libraries
+
+The [ng-packagr](https://github.com/ng-packagr/ng-packagr/blob/master/README.md) builder can automatically copy assets into your library package as part of the build process.
```
alan-agius4: - ng-packagr is not a builder, it’s a tool
- @angular-devkit/ng-packagr is a builder

aio/content/guide/creating-libraries.md
jbogarthyde: @alan-agius4 Have you had a chance to look at this yet?

Splaktar:
> Does the existing ng-packagr doc provide enough information?

No, I mentioned this in @alan-agius4's [PR](https://github.com/ng-packagr/ng-packagr/pull/1465). It added the functionality, but didn't include docs. 

> The doc I've pointed to is old, does it need to be updated for Alan's latest change?

Yes.

> Which functionality, exactly, is available as of v9?

I'm hoping that @alan-agius4 can provide some of these details for documentation either here or in the ng-packagr `.md` file docs (it probably needs to be in both).

aio/content/guide/creating-libraries.md
alan-agius4: No sorry didn’t have time yet. Will try to allocate some time tomorrow.

alan-agius4: FYI: Assets and embedded assets within css are 2 different features, which serve different purposes.

The former is used for assets which are NOT referenced in Angular components templates or stylesheets. Typically such assets are `.scss` files that contain mixins which can be used for theming. While the latter is used to embed assets which are referenced in Angular Component templates and styles.

jbogarthyde: I'm not familiar with this area - can you give me more direction about what specifically needs to  change, or needs to be said?

alan-agius4: I guess, it's better to amend this once we have the "assets" documentation in ng-packagr, so that we can link it. I'll try to write up something in the coming days.

aio/content/guide/creating-libraries.md
```diff
@@ -119,6 +119,17 @@ npm publish
 
 If you've never published a package in npm before, you must create a user account. Read more in [Publishing npm Packages](https://docs.npmjs.com/getting-started/publishing-npm-packages).
 
+{@a lib-assets}
+
+### Managing library assets with ng-packagr
+
+Starting with version 9.x of the [ng-packagr](https://github.com/ng-packagr/ng-packagr/blob/master/README.md) tool, you can use the tool to automatically copy assets into your library package as part of the build process.
+
+* Learn how to [copy assets into your library as part of the build](https://github.com/ng-packagr/ng-packagr/blob/master/docs/copy-assets-css.md). *TBD*
```
IgorMinar: this link doesn't exit, and "TBD" should not be part of our content. Should we remove this or update the link and remove TBD?

aio/content/guide/creating-libraries.md
```diff
@@ -119,6 +119,17 @@ npm publish
 
 If you've never published a package in npm before, you must create a user account. Read more in [Publishing npm Packages](https://docs.npmjs.com/getting-started/publishing-npm-packages).
 
+{@a lib-assets}
+
+### Managing library assets with ng-packagr
```
IgorMinar: this section has nothing to do with "publishing your library" so it shouldn't be the subsection of this section. I suggest we move it to the very bottom of this md file as a subsection of "Building and rebuilding your library"

IgorMinar (approved):
this needs a bit more work, but once the issues are resolved this change looks good to me.

aio/content/guide/creating-libraries.md
jbogarthyde: That's why this is "blocked" -- waiting for Alan to create this page.

aio/content/guide/creating-libraries.md
jbogarthyde: Thanks @IgorMinar 
I will remove the "TBD" when the new page is added to `ng-packagr`

alan-agius4:
@jbogarthyde here's the doc link https://github.com/ng-packagr/ng-packagr/blob/master/docs/copy-assets.md sorry for the delay.

nelson6e65:
In what versions will be available this feature?

jbogarthyde:
@nelson6e65 The ngpackagr feature is available now.

Splaktar:
> @nelson6e65 The ngpackagr feature is available now.

In the latest version 9.0.0 RC release. It hasn't been released in a stable/final release yet.

aio/content/guide/creating-libraries.md
Splaktar: I think that this section on publishing apps with additional assets is related to "publishing your library". This is the step of library development where library authors run into the issue of needing to copy assets during build time into their production bundle for publishing on NPM. This often happens when an already published library wants to add theming mixins to their published library.

I think adding this to a separate "building your library" section could work, but currently the "getting started" section covers `ng build my-lib` and the "publishing your library" section covers `ng build my-lib --prod`.

The current hierarchy doesn't seem to make sense
Screen Shot 2020-02-03 at 14 02 13

It doesn't seem like "Managing library assets with ng-packagr" should be nested under the "Using your own library in apps" section.

aio/content/guide/creating-libraries.md
```diff
@@ -213,3 +215,13 @@ For this reason, an app that depends on a library should only use TypeScript pat
 TypeScript path mappings should *not* point to the library source `.ts` files.
 
 </div>
+
+{@a lib-assets}
+
+### Managing library assets with ng-packagr
+
+Starting with version 9.x of the [ng-packagr](https://github.com/ng-packagr/ng-packagr/blob/master/README.md) tool, you can use the tool to automatically copy assets into your library package as part of the build process.
```
Splaktar:
```suggestion
Starting with version 9.x of the [ng-packagr](https://github.com/ng-packagr/ng-packagr/blob/master/README.md) tool, you can configure the tool to automatically copy assets into your library package as part of the build process. You can use this feature when your library needs to publish optional theming files, Sass Mixins, or documentation (like a change log).
```

Splaktar (changes requested):
This section seems to be missing the "Why". I've suggested changes to try to provide that.

aio/content/guide/creating-libraries.md
kapunahelewong: Hey @Splaktar, out of curiosity, why just sass mixins? Does it not include other sass features?

jbogarthyde:
Igor has approved, and Kapunahele has approved for docs.

aio/content/guide/creating-libraries.md
Splaktar: Sass mixins are most likely to be provided, but not used by the library. They may be provided to allow the consumer to optionally override a theme or other behavior of the app where just overriding a single variable is not enough. If the user doesn't need to customize this behavior, they don't have to import these mixins and can save on bundle size.

Other Sass features are most likely to be used by the library directly and thus bundled into the build by the CLI.

Splaktar: Note that a library likely has lots of other Sass mixins that it uses. So there may be a split of required and optional mixin bundles.

Splaktar:
The docs hierarchy for this still seems a bit disjoint: https://github.com/angular/angular/pull/34217#discussion_r374285659

aio/content/guide/creating-libraries.md
kapunahelewong: Ah, thank you for the clarification, @Splaktar.

kapunahelewong (approved):

aio/content/guide/creating-libraries.md
```diff
@@ -213,3 +215,14 @@ For this reason, an app that depends on a library should only use TypeScript pat
 TypeScript path mappings should *not* point to the library source `.ts` files.
 
 </div>
+
+{@a lib-assets}
+
+### Managing library assets with ng-packagr
+
+Starting with version 9.x of the [ng-packagr](https://github.com/ng-packagr/ng-packagr/blob/master/README.md) tool, you can configure the tool to automatically copy assets into your library package as part of the build process.
+You can use this feature when your library needs to publish optional theming files, Sass mixins, or documentation (like a change log).
```
alan-agius4: Not: change log -> changelog

IgorMinar (approved):
lgtm. (I fixed the nit and pushed to the branch)

Splaktar:
It looks like no one was able to respond to my concerns in https://github.com/angular/angular/pull/34217#issuecomment-582713599. I don't know if they just got lost in the shuffle or if they were deemed invalid.

jbogarthyde:
@Splaktar Can you be more specific? What would you like to change? We did move the section on asset management to a different part of the document, I thought that was in response to your comment.

Splaktar:
@jbogarthyde I'll post [my comment](https://github.com/angular/angular/pull/34217#discussion_r374285659) here again since it seems to have gotten lost (no action taken and no response):

-----------------------------
I think that this section on publishing apps with additional assets is related to "publishing your library". This is the step of library development where library authors run into the issue of needing to copy assets during build time into their production bundle for publishing on NPM. This often happens when an already published library wants to add theming mixins to their published library.

I think adding this to a separate "building your library" section could work, but currently the "getting started" section covers `ng build my-lib` and the "publishing your library" section covers `ng build my-lib --prod`.

The current hierarchy doesn't seem to make sense
Screen Shot 2020-02-03 at 14 02 13

It doesn't seem like "Managing library assets with ng-packagr" should be nested under the "Using your own library in apps" section.

-----------------------------

I'm still seeing this same structure on https://next.angular.io/guide/creating-libraries.

If we want to keep this section, "Managing library assets with ng-packagr", at the end of the page, then I think that it needs to not be indented/nested under "Using your own library in apps".

PR #34220, opened 2019-12-03
build: add karma-sauce-launcher dependency for karma_web_test macro

josephperrott:
Currently our bazel saucelabs tests silently fail as it does not have
karma-sauce-launcher available from npm.  By providing it as expected
we will properly run the bazel saucelabs tests once more


gregmagolan (approved):
LGTM. Thanks for catching that. I can look into why it failed silently.

PR #34221, opened 2019-12-04
fix(ivy): properly parenthesize ternary expressions when emitted

alxhub:
Previously, ternary expressions were emitted as:

condExpr ? trueCase : falseCase

However, this causes problems when ternary operations are nested. In
particular, a template expression of the form:

a?.b ? c : d

would have compiled to:

a == null ? null : a.b ? c : d

The precedence here is incorrect:

a == null ? null : (a.b ? c : d)

when in reality the desired association is:

(a == null ? null : a.b) ? c : d

This commit adds parentheses when emitting ternary operations to preserve
the desired precedence.

### DIFF
--- a/packages/compiler-cli/src/ngtsc/translator/src/translator.ts
+++ b/packages/compiler-cli/src/ngtsc/translator/src/translator.ts
@@ -292,8 +292,36 @@ class ExpressionTranslatorVisitor implements ExpressionVisitor, StatementVisitor
   }
 
   visitConditionalExpr(ast: ConditionalExpr, context: Context): ts.ConditionalExpression {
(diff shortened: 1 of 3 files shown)


AndrewKushnir (approved):
LGTM 👍 Thanks for the fix Alex!

kara (approved):
LGTM too, though looks like one more compliance test needs to be updated (CI is unhappy)

AndrewKushnir (approved):
👍(after the latest set of changes)

AndrewKushnir:
[Presubmit](http://test/OCL:284228209:BASE:284228253:1575661085709:6916b13a)

PR #34232, opened 2019-12-04
refactor(upgrade): avoid mutable exports.

mprobst:
Previously, create_angular_testing_module would export a mutable `let`
binding. The binding is already exporting using an accessor function
though, so the export on the let variable seems like an accidental
oversight.

This is functionally equivalent, but makes it easier for module
optimizers such as Closure Compiler to track down side effects and prune
modules.

## PR Checklist

## PR Type


mprobst:
[presubmit](https://test.corp.google.com/ui#id=OCL:283757964:BASE:283758029:1575475495451:2479cd0e)

petebacondarwin (approved):

PR #34233, opened 2019-12-04
ci: update saucelabs to use angular-framework account

josephperrott:
Currently all saucelabs usage in our repos is done using the same
account, `angular-ci`.  By migrating to use individual accounts
for each repo, we can better track the usage for each repo as well
as providing concurrency limiting on a per repo basis.

Additionally, we no longer use two separate accounts based on being
on master or a PR branch, so this logic can be removed.


devversion (approved):
LGTM. Why don't we want separate accounts for PRs / master builds? Is it because we don't show the Saucelabs badge anywhere?

josephperrott:
Correct, we don't show the badge anywhere.

Additionally, the logic we had in place wasn't actually ever triggering the separate account usage, so it wasn't even being used at this point.

PR #34234, opened 2019-12-04
fix(compiler): switch to modern diagnostic formatting

alxhub:
The compiler exports a `formatDiagnostics` function which consumers can use
to print both ts and ng diagnostics. However, this function was previously
using the "old" style TypeScript diagnostics, as opposed to the modern
diagnostic printer which uses terminal colors and prints additional context
information.

This commit updates `formatDiagnostics` to use the modern formatter, plus to
update Ivy's negative error codes to Angular 'NG' errors.

The Angular CLI needs a little more work to use this function for printing
TS diagnostics, but this commit alone should fix Bazel builds as ngc-wrapped
goes through `formatDiagnostics`.

### DIFF
--- a/packages/compiler-cli/test/ngc_spec.ts
+++ b/packages/compiler-cli/test/ngc_spec.ts
@@ -11,7 +11,7 @@ import * as path from 'path';
 import * as ts from 'typescript';
 
 import {main, readCommandLineAndConfiguration, watchMode} from '../src/main';
-import {setup} from './test_support';
+import {setup, stripAnsi} from './test_support';
 
 describe('ngc transformer command-line', () => {
   let basePath: string;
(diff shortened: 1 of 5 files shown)


josephperrott (approved):
LGTM

packages/compiler-cli/test/transformers/program_spec.ts
```diff
@@ -963,7 +963,7 @@ describe('ng program', () => {
     });
   });
 
-  it('should report errors for ts and ng errors on emit with noEmitOnError=true', () => {
+  fit('should report errors for ts and ng errors on emit with noEmitOnError=true', () => {
```
alfaproject: Oops

packages/compiler-cli/test/ngc_spec.ts
```diff
@@ -105,7 +106,9 @@ describe('ngc transformer command-line', () => {
       }`);
 
       const exitCode = main(['-p', basePath], errorSpy);
-      expect(errorSpy).toHaveBeenCalledWith(
+      debugger;
```
gkalpak: 😱

clydin:
This section can probably be simplified with this change as well:
https://github.com/angular/angular/blob/f9e959e098f1c872070edb352241653d445ba8f0/packages/compiler-cli/src/main.ts#L245

alxhub:
[Presubmit](http://test/OCL:284285323:BASE:284285351:1575678381939:e67b112)
[Ivy Presubmit](http://test/OCL:284285323:BASE:284285349:1575678381883:fd46a6fd)

AndrewKushnir (approved):

PR #34235, opened 2019-12-04
include generic type for `ModuleWithProviders` in .d.ts files

JoostK:
The `ModuleWithProviders` type has an optional type parameter that
should be specified to indicate what NgModule class will be provided.
This enables the Ivy compiler to statically determine the NgModule type
from the declaration files. This type parameter will become required in
the future, however to aid in the migration the compiler will detect
code patterns where using `ModuleWithProviders` as return type is
appropriate, in which case it transforms the emitted .d.ts files to
include the generic type argument.

This should reduce the number of occurrences where `ModuleWithProviders`
is referenced without its generic type argument.

Resolves FW-389


packages/compiler-cli/src/ngtsc/transform/src/compilation.ts
```diff
@@ -246,6 +237,14 @@ export class IvyCompilation {
 
     visit(sf);
 
+    if (!preanalyze) {
```
alxhub: Is there an issue with running the scanner during a `preanalyze: true` analysis? The CLI always has `preanalyze: true` - I'm pretty sure we'll need to do it regardless.

packages/compiler-cli/src/ngtsc/transform/src/compilation.ts
```diff
@@ -246,6 +237,14 @@ export class IvyCompilation {
...
+      this.mwpScanner.scan(sf, {
```
alxhub: There's no need to pass the object with `addTypeReplacement` - just pass `this.dtsTransforms.getReturnTypeTransform(sf)` which satisfies the `DtsHandler` interface expected by the `ModuleWithProvidersScanner`.

packages/compiler-cli/src/ngtsc/transform/src/declaration.ts
```diff
@@ -6,74 +6,278 @@
...
-import {IvyCompilation} from './compilation';
+import {DtsTransform} from './api';
 import {addImports} from './utils';
 
+/**
+ * Keeps track of `DtsTransform`s per source file, so that it is known which source files need to
+ * have their declaration file transformed.
+ */
+export class DtsTransformRegistry {
+  private ivyDeclarationTransforms = new Map<string, IvyDeclarationDtsTransform>();
```
alxhub: I would rather have a `Map<ts.SourceFile, ...>` instead of using the file name as a key. Use `ts.getOriginalNode(sf)` to go from the declaration source file to the original .ts file when doing the map lookups in `getAllTransforms`.

alxhub (changes requested):
Looking good - a few comments :)

packages/compiler-cli/src/ngtsc/transform/src/compilation.ts
JoostK: Hey, this was your logic 😄

But it seems like we should just always run this, indeed. I removed the guard.

packages/compiler-cli/src/ngtsc/transform/src/compilation.ts
JoostK: Actually, this is deliberate. This way, allocating the `ReturnTypeTransform` for the source file is only done when there's actually anything to replace.
Alternatively, I could change `DtsTransformRegistry.getAllTransforms` to check whether a pending transform actually has any work to do, but I prefer not creating the transform for the source file in the first place.

packages/compiler-cli/src/ngtsc/transform/src/declaration.ts
JoostK: Oh, I am in favor of the current situation. Happy to change though 👍

packages/compiler-cli/src/ngtsc/transform/src/compilation.ts
alxhub: > Hey, this was your logic 😄

All the more reason to be skeptical of it!

> But it seems like we should just always run this, indeed. I removed the guard.

packages/compiler-cli/src/ngtsc/transform/src/declaration.ts
alxhub: Yeah. I prefer to keep `Map`s using `ts.SourceFile` objects instead of `string` paths wherever possible, as I feel like the cost is less (it's not strictly guaranteed that `string` paths will be interned in the JS engine, whereas object identity is always cheap).

packages/compiler-cli/src/ngtsc/transform/src/compilation.ts
alxhub: Makes sense to me.

alxhub (approved):

alxhub:
LGTM otherwise. Ping me on Slack when this is updated and I'll kick off the presubmits.

alxhub:
[Presubmit](http://test/OCL:284802862:BASE:284802919:1576003816015:2d2cf39f)
[Ivy Presubmit](https://test.corp.google.com/ui#id=OCL:284802862:BASE:284802919:1576003815939:59b0d080)

AndrewKushnir:
@JoostK @alxhub there was a conflict while merging this PR into patch branch, so I merged it to master only for now. Could you please create a patch-only version of this PR with the necessary changes? Thank you.

AndrewKushnir:
@JoostK quick update: Alex R. and myself identified the set of commits that were merged to master only and as a result, were causing merge conflicts in this PR. I've created a separate PR with commits from this PR as well, so we bring patch branch to a consistent state: https://github.com/angular/angular/pull/34340. I.e. no need for a separate patch-only PR for these changes. Thank you.

PR #34236, opened 2019-12-04
build: set upper version limit for yarn to <2

josephperrott:
Rather than bumping up the allowed version of yarn on each release
we should instead just allow for anything within the major version
1 range.

### DIFF
--- a/aio/package.json
+++ b/aio/package.json
@@ -80,10 +80,10 @@
     "post~~build": "yarn build-404-page",
     "~~light-server": "light-server --bind=localhost --historyindex=/index.html --no-reload"
   },
-  "//engines-comment": "Keep this in sync with ../package.json",
+  "//engines-comment": "Keep this in sync with /package.json and /aio/tools/examples/shared/package.json",
   "engines": {
     "node": ">=10.9.0 <13.0.0",
-    "yarn": ">=1.17.3 <1.20.0"
+    "yarn": ">=1.17.3 <2"
   },
   "private": true,
   "dependencies": {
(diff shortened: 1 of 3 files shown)


devversion (approved):
LGTM.

petebacondarwin (approved):

gkalpak (approved):
Can you, please, also update [aio/tools/examples/shared/package.json][1]?
(We should add a comment in `package.json` to [the one about `aio/package.json`][2].)

[1]: https://github.com/angular/angular/blob/701bf53eb/aio/tools/examples/shared/package.json#L16
[2]: https://github.com/angular/angular/blob/701bf53eb/package.json#L9

aio/package.json
```diff
@@ -80,10 +80,10 @@
     "post~~build": "yarn build-404-page",
     "~~light-server": "light-server --bind=localhost --historyindex=/index.html --no-reload"
   },
-  "//engines-comment": "Keep this in sync with ../package.json",
+  "//engines-comment": "Keep this in sync with ../package.json and aio/tools/examples/shared/package.json",
```
gkalpak: The paths in these comments are a mess (some are absolute, some are relative to current dir, some are relative to root), but I'll pretend I don't care 😛

gkalpak (approved):

aio/package.json
```diff
@@ -80,10 +80,10 @@
     "post~~build": "yarn build-404-page",
     "~~light-server": "light-server --bind=localhost --historyindex=/index.html --no-reload"
   },
-  "//engines-comment": "Keep this in sync with ../package.json",
+  "//engines-comment": "Keep this in sync with /package.json and /aio/tools/examples/shared/package.json",
```
gkalpak: ❤️ 🤗 💖

PR #34237, opened 2019-12-04
feat(docs-infra): add the ability to expose globals

crisbeto:
Adds the ability to expose global symbols in the API docs via the `@globalApi` tag. Also supports optionally setting a namespace which will be added to the name automatically (e.g. `foo` will be renamed to `ng.foo`). Relevant APIs should also be exported through the `global.ts` file which will show up under `core/global`.

### DIFF
--- a/aio/tools/transforms/angular-api-package/processors/addGlobalApiData.js
+++ b/aio/tools/transforms/angular-api-package/processors/addGlobalApiData.js
@@ -1,0 +1,28 @@
+/**
+ * @dgProcessor addGlobalApiData
+ *
+ * Marks APIs tagged with `@globalApi` as globals and
+ * prefixes them with the namespace, if there is one.
+ */
+module.exports = function addGlobalApiDataProcessor() {
+  return {
+    $runBefore: ['computing-ids'],
+    $process: function(docs) {
+      docs.forEach(doc => {
+        const globalApiTag = doc.globalApi && doc.globalApi.trim();
+
+        if (globalApiTag != null) {
+          doc.global = true;
+
+          if (globalApiTag.length > 0) {
+            // Prefix the symbol name with the global namespace,
+            // if we have one (e.g. `foo` becomes `ng.foo`).
+            doc.unprefixedName = doc.name;
+            doc.name = `${globalApiTag}.${doc.name}`;
+            doc.globalNamespace = globalApiTag;
+          }
+        }
+      });
+    }
+  };
+};
--- a/packages/core/global.ts
+++ b/packages/core/global.ts
@@ -1,0 +1,7 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
--- a/aio/tools/transforms/angular-api-package/index.js
+++ b/aio/tools/transforms/angular-api-package/index.js
@@ -24,6 +24,8 @@ module.exports =
         .processor(require('./processors/extractPipeParams'))
         .processor(require('./processors/matchUpDirectiveDecorators'))
         .processor(require('./processors/addMetadataAliases'))
+        .processor(require('./processors/addGlobalApiData'))
+        .processor(require('./processors/updateGlobalApiPath'))
         .processor(require('./processors/computeApiBreadCrumbs'))
         .processor(require('./processors/filterContainedDocs'))
         .processor(require('./processors/processClassLikeMembers'))
@@ -95,6 +97,7 @@ module.exports =
             'common/testing/index.ts',
             'common/upgrade/index.ts',
             'core/index.ts',
+            'core/global.ts',
             'core/testing/index.ts',
             'elements/index.ts',
             'forms/index.ts',
--- a/aio/tools/transforms/angular-api-package/processors/addGlobalApiData.spec.js
+++ b/aio/tools/transforms/angular-api-package/processors/addGlobalApiData.spec.js
@@ -1,0 +1,51 @@
+const testPackage = require('../../helpers/test-package');
+const processorFactory = require('./addGlobalApiData');
+const Dgeni = require('dgeni');
+
+describe('addGlobalApiData processor', () => {
+  it('should be available on the injector', () => {
+    const dgeni = new Dgeni([testPackage('angular-api-package')]);
+    const injector = dgeni.configureInjector();
+    const processor = injector.get('addGlobalApiDataProcessor');
+    expect(processor.$process).toBeDefined();
+  });
+
+  it('should run before the correct processor', () => {
+    const processor = processorFactory();
+    expect(processor.$runBefore).toEqual(['computing-ids']);
+  });
+
+  it('should mark global APIs correctly', () => {
+    const processor = processorFactory();
+    const docs = [
+      { docType: 'function', name: 'noNamespace', globalApi: '' },
+      { docType: 'function', name: 'withNamespace', globalApi: 'ng' },
+      { docType: 'function', name: 'notGlobal' },
+    ];
+    processor.$process(docs);
+    expect(docs[0].global).toBe(true);
+    expect(docs[1].global).toBe(true);
+    expect(docs[2].global).toBeUndefined();
+  });
+
+  it('should prefix global APIs with the namespace, if one is defined', () => {
+    const processor = processorFactory();
+    const docs = [
+      { docType: 'function', name: 'noNamespace', globalApi: '' },
+      { docType: 'function', name: 'withNamespace', globalApi: 'ng' },
+      { docType: 'function', name: 'notGlobal' },
+    ];
+    processor.$process(docs);
+    expect(docs[0].name).toBe('noNamespace');
+    expect(docs[1].name).toBe('ng.withNamespace');
+    expect(docs[2].name).toBe('notGlobal');
+
+    expect(docs[0].unprefixedName).toBeUndefined();
+    expect(docs[1].unprefixedName).toBe('withNamespace');
+    expect(docs[2].unprefixedName).toBeUndefined();
+
+    expect(docs[0].globalNamespace).toBeUndefined();
+    expect(docs[1].globalNamespace).toBe('ng');
+    expect(docs[2].globalNamespace).toBeUndefined();
+  });
+});
--- a/aio/tools/transforms/angular-api-package/processors/updateGlobalApiPath.js
+++ b/aio/tools/transforms/angular-api-package/processors/updateGlobalApiPath.js
@@ -1,0 +1,26 @@
+/**
+ * @dgProcessor updateGlobalApiPath
+ *
+ * If a global API has a namespace, its name will contain a dot which will cause its
+ * URL to look like a file path. This processor updates it so it's less ambiguous.
+ */
+module.exports = function updateGlobalApiPathProcessor() {
+  return {
+    $runAfter: ['computePathsProcessor'],
+    $runBefore: ['processNgModuleDocs'],
+    $process: function(docs) {
+      docs.forEach(doc => {
+        if (doc.global && doc.globalNamespace) {
+          // We need to change the path to camel case, because having a dot
+          // in the URL will make it look like a file path.
+          const name = doc.unprefixedName;
+          const fileName = doc.globalNamespace + name[0].toUpperCase() + name.slice(1);
+
+          doc.path = `${doc.moduleDoc.moduleFolder}/${fileName}`;
+          doc.outputPath =
+              `${doc.moduleDoc.moduleFolder}/${fileName}.json`;
+        }
+      });
+    }
+  };
+};
--- a/aio/tools/transforms/angular-api-package/processors/updateGlobalApiPath.spec.js
+++ b/aio/tools/transforms/angular-api-package/processors/updateGlobalApiPath.spec.js
@@ -1,0 +1,37 @@
+const testPackage = require('../../helpers/test-package');
+const processorFactory = require('./updateGlobalApiPath');
+const Dgeni = require('dgeni');
+
+describe('updateGlobalApiPath processor', () => {
+  it('should be available on the injector', () => {
+    const dgeni = new Dgeni([testPackage('angular-api-package')]);
+    const injector = dgeni.configureInjector();
+    const processor = injector.get('updateGlobalApiPathProcessor');
+    expect(processor.$process).toBeDefined();
+  });
+
+  it('should run before the correct processor', () => {
+    const processor = processorFactory();
+    expect(processor.$runBefore).toEqual(['processNgModuleDocs']);
+  });
+
+  it('should run after the correct processor', () => {
+    const processor = processorFactory();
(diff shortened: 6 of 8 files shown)


aio/src/styles/2-modules/_api-list.scss
```diff
@@ -146,7 +146,7 @@ aio-api-list {
 
 .api-filter {
   aio-select {
-    width: 200px;
+    width: 210px;
```
crisbeto: I had to bump this because "Global Function" was a little too long and was breaking into multiple lines.

packages/core/global.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
```
crisbeto: This file is a placeholder for now. I'll add the relevant exports here once the docs are ready.

aio/tools/transforms/templates/api/global-function.template.html
```diff
@@ -0,0 +1,33 @@
+{% import "lib/memberHelpers.html" as memberHelpers -%}
+{% import "lib/paramList.html" as params -%}
+{% extends 'export-base.template.html' -%}
+
```
crisbeto: This is copied verbatim from the `function.template.html`. I'm open to suggestions on how we can have it inherit the `function` template instead.

aio/tools/transforms/angular-api-package/processors/addGlobalApiData.js
```diff
@@ -0,0 +1,36 @@
...
+
+        // Prefix the symbol name with the global namespace,
+        // if we have one (e.g. `foo` becomes `ng.foo`).
+        if (globalApiTag.length > 0) {
+          doc.name = `${globalApiTag}.${doc.name}`;
+        }
+
+        // Change the doc item's type so that we can distinguish it from the non-global ones.
+        switch (doc.docType) {
+        case 'function':
```
crisbeto: The indentation here is a little weird, but it's what was enforced through the linter.

aio/tools/transforms/angular-api-package/processors/addGlobalApiData.js
```diff
@@ -0,0 +1,36 @@
...
+          doc.docType = 'global-function';
+          break;
+        }
+      });
```
petebacondarwin: Is it not simpler to just write

```ts
doc.docType = `global-${doc.docType}`;
```

and avoid the switch altogether?

aio/tools/transforms/templates/api/global-function.template.html
petebacondarwin: If this is identical then you can probably just write something like:

```
{% extends `function.template.html` -%}
```

And leave the rest empty - similar to how `const.template.html` extends `var.temlate.html`.
```

**Review by petebacondarwin — APPROVED** [review]
I would be interested to see what the entry-point pages look like (i.e. `angular.io/api/core/global`).
Perhaps you could add the imports into `global.ts` push a temporary commit to cause the preview to be generated, then remove the commit so it doesn't land when merged?

But generally this LGTM.

Nice work @crisbeto - I hope you enjoyed your foray into dgeni!

**Review by gkalpak — COMMENTED** [review]
What about globals that are not functions? E.g. if we want to document an object (such as the `ng` that can be used in DevTools console) or something like `ngDevMode`?

My point is: Should we use a more generic term than "Global Functions"?

**Comment by petebacondarwin** [comment]
@gkalpak - I think the plan was to have `Global Functions`, `Global Consts`, etc... But maybe it is simpler to just have `Globals` and be done with it.

My suggestion here would also help with support for non-functions: https://github.com/angular/angular/pull/34237#discussion_r354026426

Actually @crisbeto - now that we have these globals in their own "entry-point" I don't know if there is as much benefit in changing their `docType`. You could leave them as they are (e.g. `doc.docType === 'function'`) but then use the fact that they have `doc.global === true` to modify the way they are rendered in the templates.

**Comment by crisbeto** [comment]
I've updated it to remove the custom `docType` for globals. Now the things that set them apart are the `/global` package and a `global` badge in the description page. This means that we can't filter by whether they're global anymore, but maybe that wasn't useful anyway?

I've also exposed some global functions so they show up in the preview. I'll revert them before merging.

Can you take another look @petebacondarwin @gkalpak?

**Comment by gkalpak** [comment]
The CI `test_aio` failure is due to the SW recognizing files with a `.` in the last path segment requests for files, while this PR introduces URLs such as `/.../ng.someFunction`.

We could probably fix the SW config, but I do feel that avoiding dots in path segments that don't point to files is preferrable.

**Comment by crisbeto** [comment]
@gkalpak is there a way to change the path through a processor? I'm thinking that we still want the name to be `ng.something`, but we could change the URL to be `ngSomething`.

**Comment by petebacondarwin** [comment]
@crisbeto - yes. Paths, Ids and Names are independent of each other...

File: aio/tools/transforms/angular-api-package/processors/addGlobalApiData.js [hunk 6]
```diff
@@ -0,0 +1,27 @@
...
+      docs.forEach(doc => {
+        const globalApiTag = doc.globalApi && doc.globalApi.trim();
+
+        if (globalApiTag != null) {
+          doc.global = true;
+
+          // Prefix the symbol name with the global namespace,
+          // if we have one (e.g. `foo` becomes `ng.foo`).
+          if (globalApiTag.length > 0) {
+            doc.name = `${globalApiTag}.${doc.name}`;
```
petebacondarwin: Here you can put something like:

```ts
doc.path = `${globalApiTag}-${doc.name}`;
doc.outputPath = `${globalApiTag}-${doc.name}`;
doc.name = `${globalApiTag}.${doc.name}`;
```

petebacondarwin (changes requested):

aio/tools/transforms/angular-api-package/processors/addGlobalApiData.js
petebacondarwin: If these are not set explicitly then they are computed based on the config here:

https://github.com/angular/angular/blob/86104b82b8af87efab46db50962e6b16f3724672/aio/tools/transforms/angular-api-package/index.js#L201-L204

petebacondarwin: and here

https://github.com/angular/dgeni-packages/blob/master/typescript/src/index.ts#L77-L81

crisbeto:
Updated @petebacondarwin. Also the preview works properly now.

aio/tools/transforms/angular-api-package/processors/addGlobalApiData.js
```diff
@@ -0,0 +1,34 @@
...
+            const unprefixedName = doc.name;
+
+            // Prefix the symbol name with the global namespace,
+            // if we have one (e.g. `foo` becomes `ng.foo`).
+            doc.name = `${globalApiTag}.${unprefixedName}`;
+
+            // We need to change the path, because having a dot
+            // in the URL will make it look like a file path.
+            doc.path = `${globalApiTag}-${unprefixedName}`;
+            doc.outputPath = `${globalApiTag}-${unprefixedName}.json`;
```
gkalpak: This causes useful info to be lost (i.e. the moduleDoc path). E.g. we get `ng-getComponent` instead of `api/core/ng-getComponent`, which in turn results in the docs being rendered incorrectly (e.g. no sidenav, no table-of-contents, etc).

We should somehow retain the module path. (Not sure what the best way is. @petebacondarwin?)

aio/tools/transforms/angular-api-package/processors/addGlobalApiData.js
```diff
@@ -0,0 +1,34 @@
...
+
+        if (globalApiTag != null) {
+          doc.global = true;
+
+          if (globalApiTag.length > 0) {
+            const unprefixedName = doc.name;
+
+            // Prefix the symbol name with the global namespace,
+            // if we have one (e.g. `foo` becomes `ng.foo`).
+            doc.name = `${globalApiTag}.${unprefixedName}`;
```
gkalpak: TBH, I find it a little confusing that we prefix the name with `globalApiTag`:

core-global

How am I supposed to interpret the above info for example? Should I import this stuff from `@angular/core/global` (assuming for example I want to write a debug utility library)? If I do, do I import `ng` from `@angular/core/global`?

Is `@angular/core/global` just a convenient wya to expose the globals so that dgeni picks them up?
Would other packages have similar entry-points (e.g. `@angular/common/global`) to export their globals?

If these packages are never meant to be used directly (which I suspect is the case), we should make sure to clearly note it on the page (and explain how users are supposed to access the helpers).

gkalpak (commented):
TBH, I found the dedicated `docType` useful. For example, I can imagine a situation where I want to discover all global utils exposed by Angular. I don't think I can do this now, without the custom `docType` 🤔

aio/tools/transforms/angular-api-package/processors/addGlobalApiData.js
gkalpak: Thinking about it again, I guess the prefix as part of the name is useful for when there are several different prefixes for an entry-point's exports. So, we should just make sure it is clear to the users how these things are supposed to be accessed.

crisbeto: The `/global` entry point is there so we can separate the imported APIs from the ones that are global. We still need it to be part of a package (e.g. `core` or `common`), because you still need to import something for the utilities to be declared.

As for the prefixing, I'm not sure how else we would expose that the user has to do `ng.something` without having to rename the functions. The screenshot above looks a little confusing, because it seems to be inheriting its content from `core`, but once the functions have proper docs we should have a dedicated description for `core/global`.

crisbeto:
I don't mind either approach (with or without changing the `docType`. The one drawback I see against having a dedicated `docType` is that it's harder to maintain, because we need one for each symbol type (e.g. `global-functions`, `global-var`) which in turn means that we need dedicated templates and having to account for it in the various checks.

petebacondarwin:
One quick way to see all the globals is to write "global" into the filter of the API page:  https://pr34237-82169df.ngbuilds.io/api?query=global

Also you can use the `doc.moduleDoc.moduleFolder` to retrieve the original proper path:

```
doc.path = `${doc.moduleDoc.moduleFolder}/${globalApiTag}-${unprefixedName}`;
doc.outputPath = `${doc.moduleDoc.moduleFolder}/${globalApiTag}-${unprefixedName}.json`;
```

crisbeto:
Updated to better preserve the path.

crisbeto:
Hmm weird, the `moduleFolder` seems to be `undefined`. Maybe the processor is running too early?

petebacondarwin:
I'm afraid that it is computed in the `computePathsProcessor`: https://github.com/angular/angular/blob/86104b82b8af87efab46db50962e6b16f3724672/aio/tools/transforms/angular-api-package/index.js#L193-L199

crisbeto:
Updated again @petebacondarwin. I had to split the processor in two, because of a circular dependency. We need to run the part that prefixes the name before the IDs are computed, and the part that changes the paths after they're available.

petebacondarwin:
I thought that might be the case

aio/src/styles/_constants.scss
```diff
@@ -116,6 +116,10 @@ $api-symbols: (
     content: 'K',
     background: $mediumgray
   ),
+  global: (
+    content: 'G',
+    background: $green-800
+  ),
```
petebacondarwin: I don't think this is used any more...

petebacondarwin (approved):
This is all looking good now.  I think we can improve a few things going forward but I don't think that should stop this PR from landing...

By the way, I have always tried to move towards very small simple processors, so splitting these two is good for me :-)

Things we can improve:

* Enhance the API filtering to be able to filter based on the API item being a global. We can use the combination of the `docType` and the `global` flag to potentially create new filter types (e.g. "global function" etc) on the fly rather than hard coding them.
* Add a `packages/core/global/PACKAGE.md` to appear on https://pr34237-f875d7d.ngbuilds.io/api/core/global. And maybe even update the template for those kind of pages so that it doesn't have the "Entry point exports" heading if all the exports are "global".

petebacondarwin:
One more thing... why is `https://pr34237-f875d7d.ngbuilds.io/api/core/global/markDirty` appearing?  Perhaps it is exported from `packages/core/src/render3/global_utils_api` but not marked as `globalApi`?

crisbeto:
Correct, it's appearing because it's exported through there. `markDirty` is supposed to be exposed as a global utility eventually according to the design doc. I'll do a follow-up commit to clean up the remaining comments and to remove these exports until we've documented them properly.

crisbeto:
Setting as ready to merge. One final tweak I had to make was change the URLs to be camel cased (`ngMarkDirty` vs `ng-markDirty`), because the service worker tests didn't like URLs with dashes in them for some reason.

crisbeto:
@gkalpak FYI I've reverted the changes that export the new utilities since they don't have proper docs yet. The last preview that has them is https://pr34237-7a90b9a.ngbuilds.io/api.

AndrewKushnir:
@gkalpak @crisbeto please let me know when this PR is ready to be merged. I see "merge" label, but looks like there are some changes and "aio: preview" was added to generate a preview. I'll remove the "merge" label for now. Feel free to add it back once it's ready to go. Thank you.

gkalpak:
> One quick way to see all the globals is to write "global" into the filter of the API page: https://pr34237-82169df.ngbuilds.io/api?query=global

@petebacondarwin: That's a work-around at best 😄 (For example, we also get results from `upgrade/static`.) And also people won't necessarily know they can do that (or how likely it is that they will not miss some globa this way), unless they know how our filtering works internally.

But I am fine addressing this on another PR.

##
> One final tweak I had to make was change the URLs to be camel cased (`ngMarkDirty` vs `ng-markDirty`)

@crisbeto: That looks confusing to me tbh; I would prefer a different name (doesn't have to block this PR though). And it can also be problematic in the unlikely case that something with the same name (e.g. both `ng.markDirty` and `ngMarkDirty`) is exported from the same `*/global` entry-point (which is admittedly unlikely).

FWIW, the reason the SW tests were failing is that we [explicitly ignore URLs with a `-`][1]. This is to ignore URLs from the old website structure, which had things like `/api/common/DatePipe-class` or `/api/common/NgFor-directive.html`. (We need to ignore those from the SW, so that they are forwarded to the server, who knows how to redirect them to the new API URLs.)

##
@AndrewKushnir: I didn't intend to block merging. I added the `aio: preview` label because I wanted to check something (for future PRs), but this should not affect merging.

Adding the `merge` label back.

[1]: https://github.com/angular/angular/blob/613dc1122/aio/ngsw-config.json#L78

gkalpak:
@crisbeto, @petebacondarwin: I created #34283 to track potential future improvements to the globals API docs. Feel free to add your throughts or more items.

AndrewKushnir:
Thanks for confirmation @gkalpak 👍

AndrewKushnir (approved):

PR #34238, opened 2019-12-04
build: Enable ivy in google3 if ng_module attr is True

kyliau:
This commit creates a way for `ng_module` rule to compile with Ivy based
on the `ivy` attribute. This enables google3 targets to pick the
compiler without using `define` flags.

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/bazel/src/ng_module.bzl
+++ b/packages/bazel/src/ng_module.bzl
@@ -42,6 +42,11 @@ def is_ivy_enabled(ctx):
             attr = "ng_module",
         )
 
+    # This attribute is only defined in google's private ng_module rule and not
+    # available externally. For external users, this is effectively a no-op.
+    if hasattr(ctx.attr, "ivy") and ctx.attr.ivy == True:
+        return True
+
     # TODO(josephperrott): Remove configuration via compile=aot define flag.
     if ctx.var.get("compile", None) == "aot":
         # buildifier: disable=print


josephperrott (approved):
LGTM

devversion (commented):
@kyliau Just double-checking. Is the attribute actually declared in the rule somewhere? (I might have missed it). If not, then I don't think that will work. Also declaring it would make it show up in the docs.

josephperrott:
@devversion
This is intentionally not declared in the rule itself.  We are taking advantage of the fact that our google3 `ng_module` rule directly relies on the `ng_module_impl` function defined here.

So essentially this is a check that only is in place for usage in our google3 `ng_module` rule to use, @kyliau and I talked through it a bit yesterday and this is our best option for build target level configuration.  We would never recommend target level configuration for end users since it easily results in a mixed state of some `ng_module`s using Ivy and some using ViewEngine, which won't work.  Our google3 setup has additional logic in place to prevent this mixed state, but it isn't something that is really doable in bazel.

devversion:
👍 I saw in the commit message that `google3` has been mentioned, but I was unsure since the commit title has `feat(bazel)` and looked like an `@angular/bazel` feature. Considering this is google3 specific logic I think we should (1) Not make it show up as a feature in the changelog? and (2) add a code comment? Right now it's unclear where the attribute is declared. Thoughts?

josephperrott:
Yeah, I didn't think about how it would show up in the changelog for it. Agreed.  I'll review requesting the changes.

packages/bazel/src/ng_module.bzl
```diff
@@ -42,6 +42,9 @@ def is_ivy_enabled(ctx):
             attr = "ng_module",
         )
 
+    if hasattr(ctx.attr, "ivy") and ctx.attr.ivy == True:
```
josephperrott: Can you add a comment explaining why this is here if its not actually usable by the provided `ng_module` rule from `@angular/bazel`

josephperrott (changes requested):
As @devversion noted, we should change the commit message for this to not show up in the changelog as a feature since its not something that is actually available in `@angular/bazel`

packages/bazel/src/ng_module.bzl
kyliau: done. Also changed commit message from feat to build.

josephperrott (approved):
LGTM, thanks!

dgp1130 (approved):

devversion (approved):
LGTM.

PR #34239, opened 2019-12-04
fix(ivy): destroy componentRef unconditionally

atscott:
[Ivy presubmit](https://test.corp.google.com/ui#id=OCL:283831496:BASE:283831536:1575497192186:a1419376)
[global presubmit](https://test.corp.google.com/ui#id=OCL:283831496:BASE:284223613:1575659842200:f354a12e) (no regressions related to this change)

The added test would fail without the change made to `component_ref`.

### DIFF
--- a/packages/core/src/render3/component_ref.ts
+++ b/packages/core/src/render3/component_ref.ts
@@ -253,7 +253,7 @@ export class ComponentRef<T> extends viewEngine_ComponentRef<T> {
     if (this.destroyCbs) {
       this.destroyCbs.forEach(fn => fn());
       this.destroyCbs = null;
-      !this.hostView.destroyed && this.hostView.destroy();
+      this.hostView.destroy();
     }
   }
 
--- a/packages/core/test/acceptance/lifecycle_spec.ts
+++ b/packages/core/test/acceptance/lifecycle_spec.ts
@@ -7,7 +7,7 @@
  */
 
 import {CommonModule} from '@angular/common';
-import {ChangeDetectorRef, Component, ComponentFactoryResolver, ContentChildren, Directive, Input, NgModule, OnChanges, QueryList, SimpleChanges, TemplateRef, ViewChild, ViewContainerRef} from '@angular/core';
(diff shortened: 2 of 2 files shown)


packages/core/src/render3/component_ref.ts
```diff
@@ -256,7 +256,7 @@ export class ComponentRef<T> extends viewEngine_ComponentRef<T> {
     if (this.destroyCbs) {
       this.destroyCbs.forEach(fn => fn());
       this.destroyCbs = null;
-      !this.hostView.destroyed && this.hostView.destroy();
```
crisbeto: We should we able to remove the `destroyed` property from the `ViewRef` as well.

packages/core/test/acceptance/lifecycle_spec.ts
```diff
@@ -3684,6 +3684,45 @@ describe('onDestroy', () => {
...
+    class App {
+      componentRef !: ComponentRef<MyComponent>;
+      constructor(
+          private readonly viewContainerRef: ViewContainerRef,
+          readonly resolver: ComponentFactoryResolver, private readonly injector: Injector) {}
+
+      ngOnInit() {
+        const factory = this.resolver.resolveComponentFactory(MyComponent);
+        const div = document.createElement('div');
+        document.body.appendChild(div);
```
crisbeto: We should remove the div from the DOM at the end of the test so it doesn't leak out into other tests.

packages/core/src/render3/component_ref.ts
atscott: There are still some uses in `ViewContainerRef.insert` and `ViewContainerRef.move` right?

packages/core/test/acceptance/lifecycle_spec.ts
atscott: Done

packages/core/src/render3/component_ref.ts
crisbeto: Ah, when I added it initially this was the only place where it was being used. It would've been nice if we could drop the getter since it generates a lot of code in es5.

crisbeto (approved):
LGTM

atscott:
Old and forgotten PR. Closing.

PR #34240, opened 2019-12-04
fix(common): update closure locales to include directionality data

josephperrott:
When directionality data was added to @angular/common locales, only
the externally used locales were updated.  We need to additionally
update the closure locales which are synced into google3.


petebacondarwin (approved):
There are a lot of formatting changes. I hope that there is nothing bad hidden amongst them.

PR #34241, opened 2019-12-05
fix(ivy): do not invoke change detection for destroyed views

AndrewKushnir:
Prior to this commit, calling change detection for destroyed views resulted in errors being thrown in some cases. This commit adds a check to make sure change detection is skipped in case a given view is marked as destroyed.

Note: I also ran micro benchmarks and the main observation is that `noop_change_detection` one is 2-3% slower, which is a bit surprising since the operation that this PR adds should be cheap. Other benchmarks have no perf changes.

This PR resolves FW-1636.

## PR Type

- [x] Bugfix

- [ ] Yes
- [x] No


AndrewKushnir:
[Presubmit](http://test/OCL:284235072:BASE:284235101:1575662879495:914a58d2)

pkozlowski-opensource (approved):
LGTM. I'm a bit sad that we are adding more checks to `refreshView` but we shouldn't block a bug fix on this. I've created a follow up issue for the future refactoring work: https://github.com/angular/angular/issues/34285

PR #34242, opened 2019-12-05
test(ivy): `class_binding` benchmark fixes

AndrewKushnir:
This commit fixes a couple issues that prevent `class_binding` benchmark from running: moving constants requires by the `benchmark` function before function declaration and referencing correct consts in template instructions.

## PR Type

- [x] Bugfix

- [ ] Yes
- [x] No


pkozlowski-opensource (approved):
LGTM, thnx @AndrewKushnir for fixing this one!

PR #30169, opened 2019-04-27
fix(ivy): inherited listeners from grand super classes invoked multiple times

crisbeto:
Fixes event listeners that come from more than one level of inheritance being invoked multiple times.

This PR resolves FW-1294.

PR #34244, opened 2019-12-05
fix(ivy): skip field inheritance if InheritDefinitionFeature is present on parent def

### LINKED ISSUE #33300: [Ivy] `HostListener` decorated methods are handled multiple times on inheriting (child) components
# 🐞 bug report

### Affected Package
Seems it's caused by `@angular/core`.

### Is this a regression?
Version 8.x **without** Ivy works fine.
With Ivy enabled, fails on both 8.x and 9.x.

### Description
In a project using Angular 9, `HostListener` and `HostBinding` decorators are handled multiple times on child and grandchild components.

## 🔬 Minimal Reproduction
Here's a link to a repo with the reproduction:
[https://github.com/ViktorSlavov/angular-inheritance-repo](https://github.com/ViktorSlavov/angular-inheritance-repo)

Steps to reproduce:
 1. Run the app and open the console.
 2. Click on one of the components labeled `Child` - one entry is logged
 7. Click on one of the components labeled `Grand Child` - two entries are logged

## 🔥 Exception or Error
None
The `grand-child` component's handler is executed twice.

## 🌍  Your Environment

**Angular Version:**
<pre><code>
Angular CLI: 8.3.12
Node: 10.15.2
OS: win32 x64
Angular: 9.0.0-next.12
... router

@angular-devkit/architect         0.803.12
@angular-devkit/build-angular     0.803.12
@angular-devkit/build-optimizer   0.803.12
@angular-devkit/build-webpack     0.803.12
@angular-devkit/core              8.3.12
@angular-devkit/schematics        8.3.12
@angular/cli                      8.3.12
@ngtools/webpack                  8.3.12
@schematics/angular               8.3.12
@schematics/update                0.803.12

</code></pre>

**Additional Info**
In the repro repo, if you remove the `HostBinding`s in `(Grand)ChildComponent`, the handler is executed only once.

AndrewKushnir:
The main logic of the `InheritDefinitionFeature` is to go through the prototype chain of a given Component and merge all Angular-specific information onto that Component def. The problem happens in case there is a Component in a hierarchy that also contains the `InheritDefinitionFeature` (i.e. it extends some other Component), so it inherits all Angular-specific information from its super class. As a result, the root Component may end up having duplicate information inherited from different Components in hierarchy.

Let's consider the following structure: `GrandChild` extends `Child` that extends `Base` and the `Base` class has a `HostListener`. In this scenario `GrandChild` and `Child` will have `InheritDefinitionFeature` included into the `features` list. The processing will happend in the following order:

- `Child` inherits `HostListener` from the `Base` class
- `GrandChild` inherits `HostListener` from the `Child` class
- since `Child` has a parent, `GrandChild` also inherits from the `Base` class

The result is that the `GrandChild` def has duplicated host listener, which is not correct.

This commit introduces additional logic that checks whether we came across a def that has `InheritDefinitionFeature` feature (which means that this def already inherited information from its super classes). If that's the case, we skip further fields-related inheritance logic, but keep going though the prototype chain to look for super classes that contain other features (like NgOnChanges), that we need to invoke for a given Component def.

This PR resolves #33300.

## PR Type

- [x] Bugfix

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/render3/features/inherit_definition_feature.ts
+++ b/packages/core/src/render3/features/inherit_definition_feature.ts
@@ -25,7 +25,7 @@ export function getSuperType(type: Type<any>): Type<any>&
  */
 export function ɵɵInheritDefinitionFeature(definition: DirectiveDef<any>| ComponentDef<any>): void {
   let superType = getSuperType(definition.type);
-  let applyInheritedFeaturesOnly = false;
+  let shouldInheritFields = true;
 
   while (superType) {
     let superDef: DirectiveDef<any>|ComponentDef<any>|undefined = undefined;
@@ -41,7 +41,7 @@ export function ɵɵInheritDefinitionFeature(definition: DirectiveDef<any>| Comp
     }
 
     if (superDef) {
-      if (!applyInheritedFeaturesOnly) {
+      if (shouldInheritFields) {
         // Some fields in the definition may be empty, if there were no values to put in them that
         // would've justified object creation. Unwrap them if necessary.
         const writeableDef = definition as any;
@@ -89,10 +89,10 @@ export function ɵɵInheritDefinitionFeature(definition: DirectiveDef<any>| Comp
           // can stop merging fields from super classes. However we need to iterate through the
           // prototype chain to look for classes that might contain other "features" (like
           // NgOnChanges), which we should invoke for the original `definition`. We set the
-          // `applyInheritedFeaturesOnly` flag to indicate that, essentially skipping fields
-          // inheritance logic and only invoking functions from the "features" list.
+          // `shouldInheritFields` flag to indicate that, essentially skipping fields inheritance
+          // logic and only invoking functions from the "features" list.
           if (feature === ɵɵInheritDefinitionFeature) {
-            applyInheritedFeaturesOnly = true;
+            shouldInheritFields = false;
           }
         }
       }
@@ -117,7 +117,6 @@ function maybeUnwrapEmpty(value: any): any {
 function inheritViewQuery(
     definition: DirectiveDef<any>| ComponentDef<any>, superViewQuery: ViewQueriesFunction<any>) {
   const prevViewQuery = definition.viewQuery;
-
   if (prevViewQuery) {
     definition.viewQuery = (rf, ctx) => {
       superViewQuery(rf, ctx);
@@ -132,7 +131,6 @@ function inheritContentQueries(
     definition: DirectiveDef<any>| ComponentDef<any>,
     superContentQueries: ContentQueriesFunction<any>) {
   const prevContentQueries = definition.contentQueries;
-
   if (prevContentQueries) {
     definition.contentQueries = (rf, ctx, directiveIndex) => {
       superContentQueries(rf, ctx, directiveIndex);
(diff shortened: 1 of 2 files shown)


AndrewKushnir:
[Presubmit](https://test.corp.google.com/ui#id=OCL:284037384:BASE:284037439:1575580501913:33a293c9)

JoostK:
Related fix done earlier is #30169, it looks like the `if (superHostBindings !== prevHostBindings)` guard in `inheritHostBindings` becomes obsolete with these changes.

packages/core/src/render3/features/inherit_definition_feature.ts
```diff
@@ -25,6 +25,7 @@ export function getSuperType(type: Type<any>): Type<any>&
  */
 export function ɵɵInheritDefinitionFeature(definition: DirectiveDef<any>| ComponentDef<any>): void {
   let superType = getSuperType(definition.type);
+  let applyInheritedFeaturesOnly = false;
```
crisbeto: Maybe we can call this something similar to `hasInheritedFields`? That way it's clearer that the flag concerns the field inheritance, not the other features. Also we won't have to rename it if ever add a third kind of inheritance in here.

packages/core/test/acceptance/inherit_definition_feature_spec.ts
```diff
@@ -4225,6 +4225,77 @@ describe('inheritance', () => {
...
+      const events: string[] = [];
+
+      @Component({
+        selector: 'app-base',
+        template: 'base',
+      })
+      class BaseComponent {
+        @HostListener('click')
+        clicked() {
+          events.push('BaseComponent.clicked');  //
```
crisbeto: Leftover comment at the end of the line.

packages/core/test/acceptance/inherit_definition_feature_spec.ts
```diff
@@ -4225,6 +4225,77 @@ describe('inheritance', () => {
...
+      })
+      class ChildComponent extends BaseComponent {
+        // additional host listeners are defined here to have `hostBindings` function generated on
+        // component def, which would trigger `hostBindings` functions merge operation in
+        // InheritDefinitionFeature logic (merging Child and Base host binding functions)
+        @HostListener('focus')
+        focused() {}
+
+        clicked() {
+          events.push('ChildComponent.clicked');  //
```
crisbeto: Leftover comment at the end of the line.

packages/core/test/acceptance/inherit_definition_feature_spec.ts
```diff
@@ -4225,6 +4225,77 @@ describe('inheritance', () => {
...
+      })
+      class GrandChildComponent extends ChildComponent {
+        // additional host listeners are defined here to have `hostBindings` function generated on
+        // component def, which would trigger `hostBindings` functions merge operation in
+        // InheritDefinitionFeature logic (merging GrandChild and Child host binding functions)
+        @HostListener('blur')
+        focused() {}
+
+        clicked() {
+          events.push('GrandChildComponent.clicked');  //
```
crisbeto: Leftover comment at the end of the line.

crisbeto (approved):
LGTM, aside from a few minor nits.

AndrewKushnir:
Thanks for review @crisbeto and @JoostK. I've applied the necessary changes based on your comments.

AndrewKushnir:
[Presubmit](http://test/OCL:288067592:BASE:288067606:1578097562278:cbe8272b)

kara (approved):
LGTM

AndrewKushnir:
[Presubmit](http://test/OCL:288411016:BASE:288411042:1578361636066:339418d3)

PR #34254, opened 2019-12-05
fix(ngcc): dts declaration mapping

### LINKED ISSUE #32442: Ivy error TS-996002, when using compiled esm5 files
# 🐞 bug report

### Affected Package

The issue is caused by package @angular/cli

### Is this a regression?

It only happens when Ivy is enabled - ngcc. With stable ngc AOT build it works fine.

### Description

The error occurs when running the `ng build` command, if the source code imports a module and component from a compiled esm5 file, which has the associated `d.ts` and `metadata.json` files.

## 🔬 Minimal Reproduction

Can be found here: https://github.com/acidb/mobiscroll-ivy-test
Running `ng build` throws the mentioned error.

## 🔥 Exception or Error
<pre><code>
ERROR in src/mobiscroll/src/js/input.angular.d.ts:3:22 - error TS-996002: Appears in the NgModule.imports of AppModule, but could not be resolved to an NgModule class

3 export declare class MbscInputModule {
</code></pre>

## 🌍  Your Environment

**Angular Version:**
<pre><code>
Angular CLI: 9.0.0-next.0
Node: 12.4.0
OS: win32 x64
Angular: 9.0.0-next.3
... router

@angular-devkit/architect         0.900.0-next.0
@angular-devkit/build-angular     0.900.0-next.0
@angular-devkit/build-optimizer   0.900.0-next.0
@angular-devkit/build-webpack     0.900.0-next.0
@angular-devkit/core              9.0.0-next.0
@angular-devkit/schematics        9.0.0-next.0
@angular/cli                      9.0.0-next.0
@ngtools/webpack                  9.0.0-next.0
@schematics/angular               9.0.0-next.0
@schematics/update                0.900.0-next.0
rxjs                              6.5.2

</code></pre>

**Anything else relevant?**

Found [this issue](https://github.com/angular/angular/issues/30972) with the same error message, which is closed and fixed, but I'm not sure if it's the same cause.

### LINKED ISSUE #33593: Compatibility Compiler Fails When Dependency has Multiple Modules with Same Named Pipe
# 🐞 bug report

### Affected Package

### Is this a regression?

 Unfortunately, I am not sure. My first attempt at using Ivy has been with `next.15` and `rc.0`. I have experienced the issue in both.

### Description

 When an app pulls in a node module that has multiple modules in it, and those modules define their own pipes that have the same name, `ngcc` fails. Ex. `CoolPackage` has `ModuleA` with pipe called `CoolPipe` and `ModuleB` with pipe also called `CoolPipe`. 

## 🔬 Minimal Reproduction

Please pull down the repo [here](https://github.com/sourdoughdetzel/angular-compat-compiler-bug) and follow the `README` instructions. 

## 🔥 Exception or Error
**From our live app:**
<pre><code>

ERROR in ../node_modules/@local/angular-pattern-library/multi-select-inline-search/multi-select-inline-search.module.d.ts:3:25 - error TS7016: Could not find a declaration file for module '../fesm2015/angular-pattern-library'. 'C:/app-path/node_modules/@local/angular-pattern-library/fesm2015/angular-pattern-library.js' implicitly has an 'any' type.

3 import * as ɵngcc2 from '../fesm2015/angular-pattern-library';
</code></pre>

**From the demo app above:**
<pre><code>
ERROR in Cannot resolve type entity ɵngcc2.ɵb to symbol
</code></pre>

## 🌍  Your Environment

**Angular Version:**
<pre><code>

Angular CLI: 9.0.0-rc.0
Node: 10.16.0
OS: win32 x64
Angular: 9.0.0-rc.0

@angular/http                     7.2.15
</code></pre>

**Anything else relevant?**

Although it may be the case that the pipes probably shouldn't have the same exact name, the fact that they are contained to the modules would make me think it shouldn't matter.

### LINKED ISSUE #34191: Ivy: NGCC incorrectly compiles the "@ngxs/router-plugin" as UMD
# 🐞 bug report

### Affected Package
@angular/compiler-cli

### Is this a regression?

Yes

### Description

The NGCC incorrectly compiles the `@ngxs/router-plugin` as UMD.

## 🔬 Minimal Reproduction
https://github.com/arturovt/ivy-ngxs-router-plugin-umd-repro

* clone and install dependencies
* check the `postinstall` script

Or open the `node_modules/@ngxs/router-plugin/bundles/ngxs-router-plugin.umd` file in your IDE and see the generated code at the end of the `factory(....)` function. This has to be something like `global...modules.ngxsFeature.module`.

Providing screenshot below (this is how it looks like on my PC):

image

## 🌍  Your Environment

**Angular Version:**
<pre><code>
Angular CLI: 9.0.0-rc.4
Node: 12.11.1
OS: linux x64

Angular: 9.0.0-rc.4
... platform-browser, platform-browser-dynamic, router
Ivy Workspace: Yes

@angular-devkit/architect         0.900.0-rc.4
@angular-devkit/build-angular     0.900.0-rc.4
@angular-devkit/build-optimizer   0.900.0-rc.4
@angular-devkit/build-webpack     0.900.0-rc.4
@angular-devkit/core              9.0.0-rc.4
@angular-devkit/schematics        9.0.0-rc.4
@ngtools/webpack                  9.0.0-rc.4
@schematics/angular               9.0.0-rc.4
@schematics/update                0.900.0-rc.4
</code></pre>

---

**UPD**

I could guess that it emits such code due to the inability to compile this package at all. The compiler behaves differently, because:

* in 80% of cases it compiles without errors and generates such code
* in 20% of cases it throws during compilation with:
```
Error: Error on worker #1: Error: Failed to compile entry-point @ngxs/router-plugin due to compilation errors:
node_modules/@ngxs/router-plugin/bundles/ngxs-router-plugin.umd.js(1034,34): error TS-991010: Value at position 0 in the NgModule.imports of NgxsRouterPluginModule is not a reference: [object Object]
```

petebacondarwin:
The naïve matching algorithm we previously used to match declarations in
source files to declarations in typings files was based only on the name
of the thing being declared.  This did not handle cases where the declared
item has been exported via an alias. This is a common scenariow when one
of the two file sets (source or typings) has been flattened, while the other
has not.

The new algorithm tries to overcome this by creating two maps of export
name to declaration (i.e. `Map<string, ts.Declaration>`).
One for the source files and one for the typings files.
It then joins these two together by matching export names, resulting in a
new map that maps source declarations to typings declarations directly
(i.e. `Map<ts.Declaration, ts.Declaration>`).

This new map can handle the declaration names being different between the
source and typings as long as they are ultimately both exported with the
same alias name.

Fixes #33593
Fixes #34191
Fixes #32442

### DIFF
--- a/packages/compiler-cli/ngcc/src/analysis/private_declarations_analyzer.ts
+++ b/packages/compiler-cli/ngcc/src/analysis/private_declarations_analyzer.ts
@@ -17,7 +17,6 @@ export interface ExportInfo {
   identifier: string;
   from: AbsoluteFsPath;
   dtsFrom?: AbsoluteFsPath|null;
-  alias?: string|null;
 }
 export type PrivateDeclarationsAnalyses = ExportInfo[];
 
@@ -42,7 +41,6 @@ export class PrivateDeclarationsAnalyzer {
       rootFiles: ts.SourceFile[],
       declarations: Map<ts.Identifier, ConcreteDeclaration>): PrivateDeclarationsAnalyses {
     const privateDeclarations: Map<ts.Identifier, ConcreteDeclaration> = new Map(declarations);
-    const exportAliasDeclarations: Map<ts.Identifier, string> = new Map();
 
     rootFiles.forEach(f => {
       const exports = this.host.getExportsOfModule(f);
@@ -54,17 +52,8 @@ export class PrivateDeclarationsAnalyzer {
               if (privateDeclaration.node !== declaration.node) {
                 throw new Error(`${declaration.node.name.text} is declared multiple times.`);
               }
-
-              if (declaration.node.name.text === exportedName) {
-                // This declaration is public so we can remove it from the list
-                privateDeclarations.delete(declaration.node.name);
-              } else {
-                // The referenced declaration is exported publicly but via an alias.
-                // In some cases the original declaration is missing from the dts program, such as
-                // when rolling up (flattening) the dts files.
-                // This is because the original declaration gets renamed to the exported alias.
-                exportAliasDeclarations.set(declaration.node.name, exportedName);
-              }
+              // This declaration is public so we can remove it from the list
+              privateDeclarations.delete(declaration.node.name);
             }
           }
         });
@@ -74,11 +63,10 @@ export class PrivateDeclarationsAnalyzer {
     return Array.from(privateDeclarations.keys()).map(id => {
       const from = absoluteFromSourceFile(id.getSourceFile());
       const declaration = privateDeclarations.get(id) !;
-      const alias = exportAliasDeclarations.has(id) ? exportAliasDeclarations.get(id) ! : null;
       const dtsDeclaration = this.host.getDtsDeclaration(declaration.node);
       const dtsFrom = dtsDeclaration && absoluteFromSourceFile(dtsDeclaration.getSourceFile());
 
-      return {identifier: id.text, from, dtsFrom, alias};
+      return {identifier: id.text, from, dtsFrom};
     });
   }
 }
--- a/packages/compiler-cli/ngcc/src/analysis/private_declarations_analyzer.ts
+++ b/packages/compiler-cli/ngcc/src/analysis/private_declarations_analyzer.ts
@@ -58,33 +58,11 @@ export class PrivateDeclarationsAnalyzer {
               if (declaration.node.name.text === exportedName) {
                 // This declaration is public so we can remove it from the list
                 privateDeclarations.delete(declaration.node.name);
-              } else if (!this.host.getDtsDeclaration(declaration.node)) {
+              } else {
                 // The referenced declaration is exported publicly but via an alias.
                 // In some cases the original declaration is missing from the dts program, such as
                 // when rolling up (flattening) the dts files.
                 // This is because the original declaration gets renamed to the exported alias.
-
-                // There is a constraint on this which we cannot handle. Consider the following
-                // code:
-                //
-                // /src/entry_point.js:
-                //     export {MyComponent as aliasedMyComponent} from './a';
-                //     export {MyComponent} from './b';`
-                //
-                // /src/a.js:
-                //     export class MyComponent {}
-                //
-                // /src/b.js:
-                //     export class MyComponent {}
-                //
-                // //typings/entry_point.d.ts:
-                //     export declare class aliasedMyComponent {}
-                //     export declare class MyComponent {}
-                //
-                // In this case we would end up matching the `MyComponent` from `/src/a.js` to the
-                // `MyComponent` declared in `/typings/entry_point.d.ts` even though that
-                // declaration is actually for the `MyComponent` in `/src/b.js`.
-
                 exportAliasDeclarations.set(declaration.node.name, exportedName);
               }
             }
--- a/packages/compiler-cli/ngcc/src/host/esm2015_host.ts
+++ b/packages/compiler-cli/ngcc/src/host/esm2015_host.ts
@@ -83,9 +83,9 @@ export class Esm2015ReflectionHost extends TypeScriptReflectionHost implements N
   protected decoratorCache = new Map<ClassDeclaration, DecoratorInfo>();
 
   constructor(
-      protected logger: Logger, protected isCore: boolean, checker: ts.TypeChecker,
+      protected logger: Logger, protected isCore: boolean, src: BundleProgram,
       dts?: BundleProgram|null) {
-    super(checker);
+    super(src.program.getTypeChecker());
     this.dtsDeclarationMap =
         dts && this.computeDtsDeclarationMap(dts.path, dts.program, dts.package) || null;
   }
--- a/packages/compiler-cli/ngcc/src/host/esm2015_host.ts
+++ b/packages/compiler-cli/ngcc/src/host/esm2015_host.ts
@@ -8,7 +8,6 @@
 
 import * as ts from 'typescript';
 
-import {AbsoluteFsPath} from '../../../src/ngtsc/file_system';
 import {ClassDeclaration, ClassMember, ClassMemberKind, ConcreteDeclaration, CtorParameter, Declaration, Decorator, TypeScriptReflectionHost, TypeValueReference, isDecoratorIdentifier, reflectObjectLiteral} from '../../../src/ngtsc/reflection';
 import {isWithinPackage} from '../analysis/util';
 import {Logger} from '../logging/logger';
@@ -50,7 +49,22 @@ export const CONSTRUCTOR_PARAMS = 'ctorParameters' as ts.__String;
  *   a static method called `ctorParameters`.
  */
 export class Esm2015ReflectionHost extends TypeScriptReflectionHost implements NgccReflectionHost {
-  protected dtsDeclarationMap: Map<string, ts.Declaration>|null;
+  /**
+   * A mapping from source declarations typings declarations, which are both publicly exported.
+   *
+   * There should be one entry for every public export visible from the root file of the source
+   * tree. Note that by definition the key and value declarations will not be in the same TS
+   * program.
+   */
+  protected publicDtsDeclarationMap: Map<ts.Declaration, ts.Declaration>|null = null;
+  /**
+   * A mapping from source declarations to typings declarations, which are not publicly exported.
+   *
+   * This mapping is a best guess between declarations that happen to be exported from their file by
+   * the same name in both the source and the dts file. Note that by definition the key and value
+   * declarations will not be in the same TS program.
+   */
+  protected privateDtsDeclarationMap: Map<ts.Declaration, ts.Declaration>|null = null;
 
   /**
    * The set of source files that have already been preprocessed.
@@ -83,11 +97,9 @@ export class Esm2015ReflectionHost extends TypeScriptReflectionHost implements N
   protected decoratorCache = new Map<ClassDeclaration, DecoratorInfo>();
 
   constructor(
-      protected logger: Logger, protected isCore: boolean, src: BundleProgram,
-      dts?: BundleProgram|null) {
+      protected logger: Logger, protected isCore: boolean, protected src: BundleProgram,
+      protected dts: BundleProgram|null = null) {
     super(src.program.getTypeChecker());
-    this.dtsDeclarationMap =
-        dts && this.computeDtsDeclarationMap(dts.path, dts.program, dts.package) || null;
   }
 
   /**
@@ -484,16 +496,32 @@ export class Esm2015ReflectionHost extends TypeScriptReflectionHost implements N
    * `ts.Program` as the input declaration.
    */
   getDtsDeclaration(declaration: ts.Declaration): ts.Declaration|null {
-    if (!this.dtsDeclarationMap) {
+    if (this.dts === null) {
       return null;
     }
     if (!isNamedDeclaration(declaration)) {
       throw new Error(
           `Cannot get the dts file for a declaration that has no name: ${declaration.getText()} in ${declaration.getSourceFile().fileName}`);
     }
-    return this.dtsDeclarationMap.has(declaration.name.text) ?
-        this.dtsDeclarationMap.get(declaration.name.text) ! :
-        null;
+
+    // Try to retrieve the dts declaration from the public map
+    if (this.publicDtsDeclarationMap === null) {
+      this.publicDtsDeclarationMap = this.computePublicDtsDeclarationMap(this.src, this.dts);
+    }
+    if (this.publicDtsDeclarationMap.has(declaration)) {
+      return this.publicDtsDeclarationMap.get(declaration) !;
+    }
+
+    // No public export, try the private map
+    if (this.privateDtsDeclarationMap === null) {
+      this.privateDtsDeclarationMap = this.computePrivateDtsDeclarationMap(this.src, this.dts);
+    }
+    if (this.privateDtsDeclarationMap.has(declaration)) {
+      return this.privateDtsDeclarationMap.get(declaration) !;
+    }
+
+    // No declaration found at all
+    return null;
   }
 
   /**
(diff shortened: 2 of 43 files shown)


packages/compiler/src/render3/r3_pipe_compiler.ts
```diff
@@ -64,14 +64,14 @@ export function compilePipeFromMetadata(metadata: R3PipeMetadata) {
   definitionMapValues.push({key: 'name', value: o.literal(metadata.pipeName), quoted: false});
 
   // e.g. `type: MyPipe`
-  definitionMapValues.push({key: 'type', value: metadata.type, quoted: false});
+  definitionMapValues.push({key: 'type', value: metadata.type.value, quoted: false});
```
petebacondarwin: I think this should be `type.value` but I am not 100% sure. What do you think @alxhub ?

packages/compiler-cli/ngcc/test/rendering/dts_renderer_spec.ts
```diff
@@ -114,7 +114,7 @@ runInEachFileSystem(() => {
             `import { Directive } from '@angular/core';\nexport class A {\n    foo(x) {\n        return x;\n    }\n}\nA.decorators = [\n    { type: Directive, args: [{ selector: '[a]' }] }\n];\n`
       };
       INPUT_DTS_PROGRAM = {
-        name: _('/typings/file.d.ts'),
+        name: _('/node_modules/test-package/typings/file.d.ts'),
```
gkalpak: OOC, why is this change needed?

gkalpak (approved):
LGTM for `fw-ngcc` 👍

packages/compiler-cli/ngcc/src/host/esm2015_host.ts
```diff
@@ -83,9 +83,9 @@ export class Esm2015ReflectionHost extends TypeScriptReflectionHost implements N
   protected decoratorCache = new Map<ClassDeclaration, DecoratorInfo>();
 
   constructor(
-      protected logger: Logger, protected isCore: boolean, checker: ts.TypeChecker,
-      dts?: BundleProgram|null) {
-    super(checker);
+      protected logger: Logger, protected isCore: boolean, src: BundleProgram,
+      dts: BundleProgram|null = null) {
```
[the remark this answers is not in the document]
petebacondarwin: Before this the type signature was `BundleProgram|null|undefined` because of the `dts?:`.
There was no tangible difference between `null` and `undefined` here so I dropped the `undefined` by this change.

packages/compiler-cli/ngcc/src/packages/transformer.ts
```diff
@@ -103,15 +103,13 @@ export class Transformer {
     const typeChecker = bundle.src.program.getTypeChecker();
```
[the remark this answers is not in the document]
petebacondarwin: Thanks!
<sub>I assume you meant "not used"</sub>

packages/compiler-cli/ngcc/test/rendering/dts_renderer_spec.ts
petebacondarwin: Because of this line: https://github.com/angular/angular/pull/34254/commits/cf36bfc0d982dbfc06f4dab782e6f3ccda59342f#diff-56411fd1f865093a818fbf15a5ea15f6R1998

```


/**
 * Execute the `callBack` for each file in the bundle program, starting with the root file.
 *
 * @param bundle the bundle program whose files should be iterated over
 * @param callBack the function to all for each of the files in the program
 */
function forEachFileInProgram(bundle: BundleProgram, callBack: (file: ts.SourceFile) => void) {
  const rootFile = bundle.program.getSourceFile(bundle.path);
  if (!rootFile) {
    throw new Error(`The given rootPath ${rootFile} is not a file of the given program.`);
  }
  // First run the call back on the root file.
  callBack(rootFile);
  bundle.program.getSourceFiles().forEach(file => {
    // Ignore if the file is the root (already processed) or if the file is not part of the package.
    if (file !== rootFile && isWithinPackage(bundle.package, file)) {
      callBack(file);
    }
  });
}
```

When computing the dts mappings we now only consider files that are "within" the package.

packages/compiler-cli/ngcc/src/host/esm2015_host.ts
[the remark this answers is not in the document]
gkalpak: I see. I don't think it makes any different in the commit it was introduced in, but it does make sense in tandem with changes in a subsequent commit where `dts` is made a protected property of the class and the [this.dtsDeclarationMap = ...][1] line below is removed.

<sub>(I am fine keeping it here for the sake of reducing commit noise.)</sub>

[1]: https://github.com/angular/angular/pull/34254/commits/6c44955b4#diff-56411fd1f865093a818fbf15a5ea15f6R89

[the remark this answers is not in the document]
petebacondarwin: Oh sorry - a rebase mixup. Thanks for being generous :-)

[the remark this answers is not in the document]
petebacondarwin: I moved it to the other commit (and updated the other hosts too) since I needed to do another rebase anyway. (No fixup commits, sorry)

packages/compiler-cli/ngcc/src/analysis/private_declarations_analyzer.ts
```diff
@@ -58,33 +58,11 @@ export class PrivateDeclarationsAnalyzer {
               if (declaration.node.name.text === exportedName) {
                 // This declaration is public so we can remove it from the list
                 privateDeclarations.delete(declaration.node.name);
-              } else if (!this.host.getDtsDeclaration(declaration.node)) {
+              } else {
```
petebacondarwin: Actually I think this entire `else` block (and associated `alias` property on `ExportInfo` can now be dropped since the dts mapping is able to handle aliasing.

petebacondarwin:
Rebased on top of #34356 as that is also needed to fix #33593

packages/compiler-cli/ngcc/src/host/esm2015_host.ts
```diff
@@ -1972,3 +1979,24 @@ function getContainingStatement(node: ts.Node): ts.ExpressionStatement|null {
...
   return node || null;
 }
+
+/**
+ * Execute the `callBack` for each file in the bundle program, starting with the root file.
+ *
+ * @param bundle the bundle program whose files should be iterated over
+ * @param callBack the function to all for each of the files in the program
+ */
+function forEachFileInProgram(bundle: BundleProgram, callBack: (file: ts.SourceFile) => void) {
```
alxhub:
```suggestion
function forEachFileInProgram(bundle: BundleProgram, callback: (file: ts.SourceFile) => void) {
```

packages/compiler-cli/ngcc/src/host/esm2015_host.ts
```diff
@@ -1972,3 +1979,24 @@ function getContainingStatement(node: ts.Node): ts.ExpressionStatement|null {
...
+  const rootFile = bundle.program.getSourceFile(bundle.path);
+  if (!rootFile) {
+    throw new Error(`The given rootPath ${rootFile} is not a file of the given program.`);
+  }
+  // First run the call back on the root file.
+  callBack(rootFile);
+  bundle.program.getSourceFiles().forEach(file => {
```
alxhub: Should we check `isDeclarationFile` too?

alxhub (changes requested):

packages/compiler-cli/ngcc/src/host/esm2015_host.ts
petebacondarwin: No because this function is called twice - once with dts files and once with source files.

packages/compiler-cli/ngcc/src/host/esm2015_host.ts
```diff
@@ -1499,42 +1507,25 @@ export class Esm2015ReflectionHost extends TypeScriptReflectionHost implements N
...
-    // Now add any additional classes that are exported from individual  dts files,
-    // but are not publicly exported from the entry-point.
-    dtsProgram.getSourceFiles().forEach(sourceFile => {
-      if (!isWithinPackage(dtsPackage, sourceFile)) {
-        return;
+  protected computeDtsDeclarationMap(src: BundleProgram, dts: BundleProgram):
+      Map<ts.Declaration, ts.Declaration> {
+    const dtsMap = computeDtsExportDeclarationMap(dts);
+    const declarationMap = new Map<ts.Declaration, ts.Declaration>();
+    forEachFileInProgram(src, file => {
```
alxhub: Is this logic correct? Only the entrypoint index's exports are expected to match up to the .d.ts program, I would think.

So if the source program has a file `lib/internal/foo.js` that has `export class Foo {...}`, there is zero guarantee that this is the same `Foo` exported from the root of the `.d.ts` program.

I see that `forEachFileInProgram` will process the root file first. For that file, the logic will be correct.

For other files, though, we're making an assumption that export names will line up. That's okay, but we should be clear about such assumptions.

I think we should:

1) keep the `Map`s of certified matches vs guesses separate. So we should have a primary `Map` that matches up the root file, and a "guess" `Map` that matches up deeply exported things by name only.

2) currently, the logic actually overwrites `declarationMap` unilaterally on a match, even if previously the root file gave us a definitive (and different) answer. The split approach above will help avoid this problem.

3) we should have a comment here explaining the matching algorithm and the two different approaches.

packages/compiler-cli/ngcc/src/host/esm2015_host.ts
```diff
@@ -1863,21 +1854,45 @@ function isClassMemberType(node: ts.Declaration): node is ts.ClassElement|
...
+ * are ignored.
+ *
+ * We are most interested in classes that are publicly exported from the root file, so these
+ * are added to the map first, to ensure that they are not ignored.
+ *
+ * @param dtsRoot The path to the root file for the dts program.
+ * @param dtsProgram The program containing all the dts declarations.
+ * @returns a map of exported names to declarations.
+ */
+function computeDtsExportDeclarationMap(dts: BundleProgram): Map<string, ts.Declaration> {
```
alxhub: I think the same goes for .d.ts exports. We should keep entrypoint exports and deep exports in separate `Map`s.

packages/compiler/src/render3/r3_pipe_compiler.ts
alxhub: You're correct. :)

alxhub (approved):

alxhub (changes requested):
Sorry, wrong button.

packages/compiler-cli/ngcc/src/host/esm2015_host.ts
petebacondarwin: I have refactored this as suggested. There are now two maps: one for exports that are exposed publicly via the entry-point (i.e. the root file) and one for exports that are exposed from individual files in the source tree. 

As a result of doing this I found that there were further problems with the CommonJS and UMD re-export handling, which I have fixed in two extra commits.

Therefore the whole of this PR could probably do with another review: @alxhub , @gkalpak , @JoostK .

alxhub (approved):
👍 I really like the logic here - it's clean, easy to follow, and I think will address a lot of subtle .d.ts matching issues we have today.

alxhub:
[Presubmit](https://test.corp.google.com/ui#id=OCL:286221348:BASE:286221394:1576694400309:aa213af6)
[Ivy Presubmit](https://test.corp.google.com/ui#id=OCL:286221348:BASE:286221394:1576694400309:c0c11529)

packages/compiler-cli/ngcc/src/host/umd_host.ts
```diff
@@ -237,6 +273,15 @@ interface UmdExportDeclaration {
...
 }
 
+type ReexportStatement = ts.ExpressionStatement & {expression: {arguments: [ts.Identifier]}};
+function isReexportStatement(statement: ts.Statement): statement is ReexportStatement {
+  return ts.isExpressionStatement(statement) && ts.isCallExpression(statement.expression) &&
+      ts.isIdentifier(statement.expression.expression) &&
+      statement.expression.expression.text === '__export' &&
+      statement.expression.arguments.length === 1 &&
+      ts.isIdentifier(statement.expression.arguments[0]);
+}
```
gkalpak: Can't UMD have `__export(require('...'))` type re-exports? I thought it could.

gkalpak (approved):

packages/compiler-cli/ngcc/src/host/umd_host.ts
petebacondarwin: You can have commonjs `require` imports but they must be passed to the factory function. I dont see how you could have `require` reexports without breaking non-commonjs formats

gkalpak: My understanding is that you could also have `require(...)` calls inside the factory (what RequireJS calls [simplified CommonJS wrapping][1]). Also, the TypeScript playground seems to generate that format for UMD (but that might be because the playground doesn't do a proper build): [TS playground example][2]

[1]: https://requirejs.org/docs/whyamd.html#sugar 
[2]: https://www.typescriptlang.org/play/?target=1&module=3&ssl=1&ssc=1&pln=28&pc=1#code/JYWwDg9gTgLgBAIhgOwM4FoDG0Cm6QQAmArgDY6oD0AhmGKcJtTMBMggNwBQXOAHpFhwAVHABmUCCEQA6SvWZjoILFILtu-QfFESps+aUXLOvAdB3jJ0hHIUwlUFahjNGprRZFX9tw8ad0MEkAN2BCHChUD3MhXWsDKAoIYihMPFIIagioUzNtbz0bOWQ3EIpMKGAwGBkCEnIYgvjfOWwQdTzPOJ9iykliGEimrxa+sWBydFQATxccEBGeooMACxgYMCwGHGQYJcsVv0cQaO58rwBvLjhbuAA1YBwAdwBhI1RUABobu8eX97UT4AWRwrh+dweT2eACVkqRylAIXcAOZg-5vD7fX63NEwDGg8E4uDAVAAaWQEGeyAxyNuSRRpKGUAAouQQLsYD8AL69Aw4dmc9AMplQGZ5bBoeBKCBwAC8cAA2jIVQByACMACYAMyqgC63CAA

petebacondarwin: OK, so AMD supports a `require` argument passed through to the factory function, so 
I guess it is possible and we should perhaps support it... But in that case you cannot use the UMD templates that support globals. See https://github.com/umdjs/umd

gkalpak: I am not arguing whether it is a good idea or not to use `require()` calls in UMD, but if it is possible that the UMD files we need to process use that, then it would be better if we detected it.

I'd be happy to add support for this in a PR I am working on to fix star re-exports with imported helpers.

gkalpak: For reference: #34512

PR #34255, opened 2019-12-05
refactor(ivy): minor cleanup in the listener instruction

pkozlowski-opensource:
Super-tiny cleanup (couldn't resist...).

### DIFF
--- a/packages/core/src/render3/instructions/listener.ts
+++ b/packages/core/src/render3/instructions/listener.ts
@@ -119,6 +119,11 @@ function listenerInternal(
   const firstCreatePass = tView.firstCreatePass;
   const tCleanup: false|any[] = firstCreatePass && (tView.cleanup || (tView.cleanup = []));
 
+  // When the ɵɵlistener instruction was generated and is executed we know that there is either a
+  // native listener or a directive output on this element. As such we we know that we will have to
+  // register a listener and store its cleanup function on LView.
+  const lCleanup = getCleanup(lView);
+
   ngDevMode && assertNodeOfPossibleTypes(
                    tNode, TNodeType.Element, TNodeType.Container, TNodeType.ElementContainer);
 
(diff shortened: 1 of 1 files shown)


petebacondarwin (commented):
How expensive is `getCleanup()`?
If the `tNode` is not an element and there are no output properties then the `lCleanup` is not needed, right?

So the trade-off is between calling it (possibly) twice, when it is needed; and calling it once when it is not. How common is each case?

One way around this would be to lazy initialize `lCleanup`...

kara (approved):
LGTM

kara:
@petebacondarwin This change is in the `listenerInternal` instruction, so there is at least one listener defined on the node. My thinking is this: If it's not an element, then the listener defined has to be an output. You can't put a DOM event listener on something like a container or an ng-container. As far as I can think of, there's no case where we would be calling it unnecessarily because it has to be either an element or an output for there to be a listener instruction generated.

Unless I'm missing a case? What were you thinking would be the scenario?

pkozlowski-opensource:
@kara yep, this was exactly my assumption - as soon as we generate a `listener` instruction we will have _either_ a native listener _or_ a directive's output and in any case we need to store a cleanup function on `LView`. The only other scenarios I could think of were error ones, but then the Whole World is broken...

kara:
[presubmit](http://test/OCL:284454261:BASE:284454268:1575843011273:29be0284)

kara:
@pkozlowski-opensource Perhaps we should put this reasoning in the commit message description? It's non-obvious so would be good to have it documented in addition to noting it's a cleanup

pkozlowski-opensource:
> Perhaps we should put this reasoning in the commit message description?

@kara sure, will do. Still would like to hear from @petebacondarwin as he often sees cases I don't see, so maybe he had a particular scenario on his mind?

petebacondarwin:
I was just following the code paths:

```
const thing = ...
if (x) {
  useThing(thing);
}
if (y) {
  useThing(thing);
}
```

If it is always true that `x || y === true` then that is fine. But it is not 100% obvious from the code, IMO.
Rather than adding it to the commit message, I would suggest a small comment before the `const lCleanup` explaining why it is OK to do the work here.

Commit messages are great for retrospectively working out why something was done, but not good for preventing people for "fixing" stuff that is not broken in the future.

pkozlowski-opensource:
@petebacondarwin @kara added a comment in the code

kara (approved):
LGTM

PR #34262, opened 2019-12-06
fix(language-service): Simplify resolution logic in banner

kyliau:
Due to a bug in the existing banner, `typescript` module was require-d
instead of reusing the module passed in from tsserver.
This bug is caused by some source files in language-service that imports
`typescript` instead of `typescript/lib/tsserverlibrary`.
This is not an unsupported use case, it's just that when typescript is
resolved in the banner we have to be very careful about which modules to
"require".
The convoluted logic in the banner makes it very hard to detect
anomalies. This commit cleans it up and removes a lot of unneeded code.

This commit also removes `ts` import in typescript_host.ts and use `tss`
instead to make it less confusing.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No


ayazhafiz (approved):
Are we able rename `banner.js.txt` to `banner.js`?

PR #34270, opened 2019-12-06
ci: run saucelabs legacy tests with ivy

devversion:
WIP / DRAFT

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -667,6 +667,7 @@ jobs:
     steps:
       - custom_attach_workspace
       - init_environment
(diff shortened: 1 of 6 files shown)


devversion:
This technically works (tested locally if I build zone.js properly), but it turns out that:

1. We needed a lot of workarounds to build packages with ngtsc --> so that the `PRE_R3` and `POST_R3` switches are updated
2. We need to audit all tests which are View Engine only and should not run
3. It will still be the legacy setup. I think we should just start a new non-legacy job with fine-grained Ivy tests that should run w/ *bazel*.

PR #34277, opened 2019-12-06
ci: run acceptance tests on saucelabs with ivy

devversion:
Currently we only run Saucelabs on PRs using the legacy View Engine build. Switching that build to Ivy is not trivial and there are various options:

1. Updating the R3 switches to use `POST_R3` by default. At first glance, this doesn't look easy because the current ngtsc switch logic seems to be unidirectional (only PRE_R3 to POST_R3).
2. Updating the legacy setup to run with Ivy. This sounds like the easiest solution at first.. but it turns out to be way more complicated.
   * Packages would need to be built with ngtsc using legacy tools (i.e. first building the compiler-cli; and then building packages)
   * View Engine only tests would need to be determined and filtered out. Basically it will result in re-auditing all test targets... Contradictory to the fact that we have this information in Bazel already.
3. Creating a new job that runs tests on Saucelabs with Bazel. We specify fine-grained test targets that should run. This would be a good start (e.g. acceptance tests) and also would mean that we do not continue maintaining the legacy setup..

### DIFF
--- a/.circleci/bazel.common.rc
+++ b/.circleci/bazel.common.rc
@@ -13,3 +13,8 @@ test --flaky_test_attempts=2
 
 # More details on failures
 build --verbose_failures=true
+
+# For saucelabs tests we don't want to enable flaky test attempts. Karma has its own integrated
+# retry mechanism and we do not want to retry unnecessarily if Karma already tried multiple times.
+# Setting the option to "default" is necessary because the option is set above.
+test:saucelabs --flaky_test_attempts=default
--- a/.circleci/bazel.common.rc
+++ b/.circleci/bazel.common.rc
@@ -16,5 +16,4 @@ build --verbose_failures=true
 
 # For saucelabs tests we don't want to enable flaky test attempts. Karma has its own integrated
 # retry mechanism and we do not want to retry unnecessarily if Karma already tried multiple times.
-# Setting the option to "default" is necessary because the option is set above.
-test:saucelabs --flaky_test_attempts=default
+test:saucelabs --flaky_test_attempts=1
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -326,7 +326,7 @@ jobs:
   # [BUILD.bazel](https://github.com/angular/angular/blob/ef44f51d5/BUILD.bazel#L66-L92)).
   #
   # NOTE: This is currently limited to master builds only. See the `default_workflow` configuration.
-  test_saucelabs_bazel:
+  saucelabs_view_engine:
     executor:
       name: default-executor
       # In order to avoid the bottleneck of having a slow host machine, we acquire a better
@@ -340,9 +340,9 @@ jobs:
       - setup_bazel_rbe
       - run:
           name: Run Bazel tests in saucelabs
-          # All web tests are contained within a single //:test_web_all target for Saucelabs
-          # as running each set of tests as a separate target will attempt to acquire too
-          # many browsers on Saucelabs (7 per target currently) and some tests will always
+          # All web tests are contained within a single //:saucelabs_unit_tests_poc target
+          # for Saucelabs as running each set of tests as a separate target will attempt to acquire
+          # too many browsers on Saucelabs (7 per target currently) and some tests will always
           # fail to acquire browsers. For example:
           # 14 02 2019 19:52:33.170:WARN [launcher]: chrome beta on SauceLabs have not captured in 180000 ms, killing.
           # //packages/forms/test:web_test_sauce TIMEOUT in 315.0s
@@ -351,10 +351,33 @@ jobs:
               --tunnel-id angular-${CIRCLE_BUILD_NUM}-${CIRCLE_NODE_INDEX} \
               --username $SAUCE_USERNAME \
               --key $(echo $SAUCE_ACCESS_KEY | rev) \
-              yarn bazel test //:test_web_all
+              yarn bazel test //:saucelabs_unit_tests_poc --config=saucelabs
           no_output_timeout: 20m
       - notify_webhook_on_fail:
           webhook_url_env_var: SLACK_DEV_INFRA_CI_FAILURES_WEBHOOK_URL
+
+  saucelabs_ivy:
+    executor:
+      name: default-executor
+      # In order to avoid the bottleneck of having a slow host machine, we acquire a better
+      # container for this job. This is necessary because we launch a lot of browsers concurrently
+      # and therefore the tunnel and Karma need to process a lot of file requests and tests.
+      resource_class: xlarge
+    steps:
+      - custom_attach_workspace
+      - init_environment
+      - setup_circleci_bazel_config
+      - setup_bazel_rbe
+      - run:
+          name: Run Bazel tests on Saucelabs
+          # Runs the //:saucelabs_tests target with Saucelabs and Ivy.
+          command:  |
+            ./scripts/saucelabs/run-bazel-via-tunnel.sh \
+              --tunnel-id angular-${CIRCLE_BUILD_NUM}-${CIRCLE_NODE_INDEX} \
+              --username $SAUCE_USERNAME \
+              --key $(echo $SAUCE_ACCESS_KEY | rev) \
+              -- yarn bazel test //:saucelabs_unit_tests --config=ivy --config=saucelabs
+          no_output_timeout: 20m
 
   test_aio:
     # Needed because the AIO tests and the PWA score test depend on Chrome being available.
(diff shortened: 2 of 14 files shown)


.circleci/config.yml
```diff
@@ -815,6 +840,9 @@ workflows:
       - legacy-unit-tests-saucelabs:
           requires:
             - setup
+      - saucelabs_ivy:
```
gkalpak: Shouldn't we make this required by `publish_snapshot`?

karma-js.conf.js
```diff
@@ -145,7 +143,15 @@ module.exports = function(config) {
...
     browserNoActivityTimeout: 300000,
-  }
+  };
+
+  // Workaround for: https://github.com/bazelbuild/rules_nodejs/issues/1431. The idea is
+  // that we do no not allow `@bazel/karma` to add the `progress` reporter.
+  Object.defineProperty(conf, 'reporters', {
+    enumerable: true,
+    get: () => ['dots'],
+    set: () => {},
```
gkalpak: Snazy 👌

gkalpak (approved):
:tada:

BUILD.bazel
```diff
@@ -47,6 +47,21 @@ filegroup(
     ],
 )
 
+# We combine all tests into a single "karma_web_test" target because running them as separate
+# targets in parallel can result in to too many browsers being acquired at the same time. This
+# will then result in tests being flaky.
+karma_web_test(
+    name = "saucelabs_tests",
```
josephperrott: Should we rename both this `karma_web_test` target as well as the `test_web_all` target to express they are to run in saucelabs, and note which one is for ivy targets and which is for viewengine targets?

.circleci/config.yml
devversion: yeah. will add it as required

BUILD.bazel
devversion: @josephperrott Updated. Technically these targets are independent of whether they run with Ivy or VE because that's ultimately what the command decides, but considering that we use these as a  "filegroup" for CI-only, I added comments for clarity.

IgorMinar: as suggested elsewhere, we should rename this to saucelabs_acceptance_tests, assuming that this target will contain only "bigger" acceptance tests.

alternatively have you considered adding //packages/core/test/acceptance:acceptance_lib to saucelabs_cronjob_tests and running that target in ivy and ve modes on CI?

IgorMinar: actually, why not do that instead of having different subset of tests for VE and different for Ivy?

.circleci/config.yml
```diff
@@ -815,6 +840,9 @@ workflows:
...
+            requires:
+              - setup
```
IgorMinar: can you also require test_ivy_aot so that we don't run stuff on sauce (and drain capacity) unless we know that locally the tests pass?

IgorMinar (changes requested):
I assume that it it expected that the CI is failing because we do have legit regressions on IE10/11 with Ivy that need to be fixed independently.

Do we have a rollout plan for this change? Is there a way to mark this job as optional / OK to fail, while we fix the tests and then make it non-optional? (I'm not aware of circleci having such job option, but we could force the bazel step to always exit with 0 to workaround a missing circleci feature if it's really missing).

.circleci/config.yml
```diff
@@ -351,11 +351,36 @@ jobs:
...
+          name: Run Bazel tests on Saucelabs
+          # Runs the //:saucelabs_tests target with Saucelabs and Ivy. Note that we also
+          # pass "--flaky_test_attempts=default" because we don't want Bazel to rerun
+          # the test in case of an error. The retry mechanism is already part of Karma.
+          command:  |
+            ./scripts/saucelabs/run-bazel-via-tunnel.sh \
+              --tunnel-id angular-${CIRCLE_BUILD_NUM}-${CIRCLE_NODE_INDEX} \
+              --username $SAUCE_USERNAME \
+              --key $(echo $SAUCE_ACCESS_KEY | rev) \
+              -- yarn bazel test //:saucelabs_tests --config=ivy --flaky_test_attempts=default
```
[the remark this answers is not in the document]
devversion: haha, it's the default but we always set it to `2` in our CircleCI `bazelrc`, so this overwrites it back to their `default` mechanism. The comment above is maybe not clear enough that the test attempts are currently overwritten in Circle.

Using it for the other job above sounds good 👍

packages/common/http/test/BUILD.bazel
```diff
@@ -6,7 +6,7 @@ ts_library(
     srcs = glob(
         ["**/*.ts"],
     ),
-    # Visible to //:test_web_all target
+    # Visible to //:saucelabs_cronjob_tests target
```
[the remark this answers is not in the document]
devversion: I tried that initially but this doesn't seem to work in Bazel. I just kept the visibility like before.

This is the error:

```
 cycle in dependency graph:
    //packages/core/test/acceptance:acceptance_lib
.-> //:saucelabs_tests (host)
|   //packages/core/test/acceptance:acceptance_lib (host)
`-- //:saucelabs_tests (host)

```

BUILD.bazel
devversion: The plan was not to only have acceptance tests in there. My thinking was that we want to add more targets over time, but start with the acceptance tests since these were critical for validating #34306. 

I didn't consider combining both targets because the other one is primarily for a PoC and the target seems to be flaky quite often (see #dev-infra-ci-failures channel). This PR was just about getting validation for the Ivy IE bug fixes as fast as possible.

I think we need to spend more time deciding on the general saucelabs plan? with some auditing of what actual targets we want to run (_for every PR_)

.circleci/config.yml
devversion: Sounds like a good idea. It shouldn't be an issue for now since the sum of both jobs `test_ivy_aot` and `saucelabs_ivy` does not exceed the overall time the CI currently takes.

.circleci/config.yml
```diff
@@ -351,11 +351,36 @@ jobs:
               --tunnel-id angular-${CIRCLE_BUILD_NUM}-${CIRCLE_NODE_INDEX} \
               --username $SAUCE_USERNAME \
               --key $(echo $SAUCE_ACCESS_KEY | rev) \
-              yarn bazel test //:test_web_all
+              yarn bazel test //:saucelabs_cronjob_tests
```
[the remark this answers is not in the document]
devversion: oops. yeah good catch.

.circleci/config.yml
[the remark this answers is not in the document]
devversion: Moved the flag into a bazelrc config and tried to make clear that we really want to set `default` as value. Let me know if you think it can be made more clear. added it for the other job too 😁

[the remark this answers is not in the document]
devversion: let's discuss this in the comment by Joey above. I left a comment there on why I don't think we want this to name `acceptance_tests`.

devversion:
@IgorMinar Thanks for reviewing! Yes. it's expected that the job fails in this PR. 

We've been chatting about the best way on landing this in the [PR that fixes the actual IE issues](https://github.com/angular/angular/issues/34305). Based on Kristiyan's comments, the plan is that his IE fixes should land before, and then this PR can be merged (should be green at that point).

I replied on your proposals for naming and combining both saucelabs test targets into a single one. I totally agree that we can avoid duplication.. but it feels out of scope for this PR since we don't even know what to do with the PoC Saucelabs Bazel job.

devversion:
I rebased this PR since the IE10/IE11 fixes landed and the job is now green.

.circleci/bazel.common.rc
```diff
@@ -13,3 +13,8 @@ test --flaky_test_attempts=2
 
 # More details on failures
 build --verbose_failures=true
+
+# For saucelabs tests we don't want to enable flaky test attempts. Karma has its own integrated
+# retry mechanism and we do not want to retry unnecessarily if Karma already tried multiple times.
+# Setting the option to "default" is necessary because the option is set above.
+test:saucelabs --flaky_test_attempts=default
```
IgorMinar: if our goal is to set this to 1, then why not use `1` rather than `default` (which is ambiguous)?

IgorMinar (approved):
lgtm. consider replacing the `default` value.

IgorMinar:
@devversion I marked this for merge, but if you agree that `1` is better than `default` then please update this PR or send a follow up PR.

.circleci/bazel.common.rc
devversion: `default` has a different meaning than `1`. If the value is set to `default`, then Bazel will still re-run tests explicitly marked with `flaky = True` three times. If we'd set it to `1` explicitly, then the `flaky` attribute on targets would be a noop.

https://docs.bazel.build/versions/master/command-line-reference.html#flag--flaky_test_attempts

devversion: I think we can set it to `1` though since we can assume that every target that runs with the `saucelabs` config, actually runs with Saucelabs/Karma and doesn't need the retry mechanism.

josephperrott: I think setting it to `1` is a better option for us because failing is preferable to actually run things 3 times. Since its so costly in time/resource constraints to run things on saucelabs.

devversion: @josephperrott `1` would behave the same as `default` for the Saucelabs targets since they are not marked with `flaky = True`. The exception is that for targets which do have the attribute explicitly set, the mechanism would still work.

Anyway though I changed it to `1` because it's more obvious what it does, and we can assume that there are no targets running with `--config=saucelabs` that have `flaky = True` set.

kara (approved):

kara:
caretaker who is me: global approval

PR #34265, opened 2019-12-06
fix(compiler): ensure localized strings are ES5 compatible for JIT mode

petebacondarwin:
Previously the JIT evaluated code for ivy localized strings included
backtick tagged template strings, which are not compatible with ES5
in legacy browsers such as IE 11.

Now the generated code is ES5 compatible.

Fixes #34246

PR #34279, opened 2019-12-06
Test ivy JIT mode on IE

gkalpak:
Builds on top of #34265 and #34277.


gkalpak:
Superceded by #34305.

PR #34286, opened 2019-12-06
fix(ivy): handle SafeStyles in [style.prop] correctly

### LINKED ISSUE #34023: Ivy preview strange behavior when binding SafeStyles
If I disable Ivy I have no errors in thumbnails, but if i back enable Ivy I see only one thumbnails element.
No styles for other elements...
image

image

https://github.com/kolkov/ngx-gallery/blob/53785a725e709b10c0f716050e41d32cc3e73d68/projects/gallery/src/lib/ngx-gallery-thumbnails/ngx-gallery-thumbnails.component.html#L7

https://github.com/kolkov/ngx-gallery/blob/53785a725e709b10c0f716050e41d32cc3e73d68/projects/gallery/src/lib/ngx-gallery-thumbnails/ngx-gallery-thumbnails.component.ts#L159

image

AndrewKushnir:
Prior to this commit, values wrapped into SafeStyle were not handled correctly in [style.prop] bindings in case style sanitizer is present (when template contains some style props that require sanitization). Style sanitizer was not unwrapping values in case a given prop doesn't require sanitization.As a result, wrapped values were used as final styling values (see [styling/bindings.ts](https://github.com/angular/angular/blob/master/packages/core/src/render3/styling/bindings.ts#L620)). This commit updates the logic to unwrap safe values in sanitizer in case no sanitization is required.

This PR closes #34023.

## PR Type

- [x] Bugfix

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/sanitization/sanitization.ts
+++ b/packages/core/src/sanitization/sanitization.ts
@@ -196,7 +196,7 @@ export const ɵɵdefaultStyleSanitizer =
       }
 
       if (mode & StyleSanitizeMode.SanitizeOnly) {
-        return doSanitizeValue ? ɵɵsanitizeStyle(value) : value;
+        return doSanitizeValue ? ɵɵsanitizeStyle(value) : unwrapSafeValue(value);
       } else {
         return doSanitizeValue;
       }
--- a/packages/core/test/acceptance/styling_spec.ts
+++ b/packages/core/test/acceptance/styling_spec.ts
@@ -1598,7 +1598,7 @@ describe('styling', () => {
     const comp = fixture.componentInstance;
     const style = fixture.nativeElement.firstChild.style;
 
-    expect(style.background).toBe('url("/1.png")');
+    expect(style.background).toContain('url("/1.png")');
     expect(style.width).toBe('calc(20%)');
     expect(style.height).toBe('10px');
     expect(style.color).toBe('red');
@@ -1610,7 +1610,7 @@ describe('styling', () => {
 
     fixture.detectChanges();
 
-    expect(style.background).toBe('url("/2.png")');
+    expect(style.background).toContain('url("/2.png")');
     expect(style.width).toBe('5px');
     expect(style.height).toBe('100%');
     expect(style.color).toBe('green');
--- a/packages/core/test/acceptance/styling_spec.ts
+++ b/packages/core/test/acceptance/styling_spec.ts
@@ -5,6 +5,7 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
+import {CommonModule} from '@angular/common';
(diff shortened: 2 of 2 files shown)


packages/core/test/acceptance/styling_spec.ts
```diff
@@ -1563,6 +1564,58 @@ describe('styling', () => {
     expect(html).toMatch(/style=["|']clip-path:\s*url\(.*#test.*\)/);
   });
 
+  it('should not throw when bound to SafeValue', () => {
```
IgorMinar: can you please split this into two specs - one for sanitized properties (e.g. `background`, `background-image`, or `border-image`) and one for regular properties (e.g. `height`).

These two types of properties are handled differently, and we should have explicit tests for both.

alternatively, since it looks like the setup is quite expensive/verbose, you can also separate the `[style.xxx]` bindings in the template into two distinct groups (two divs or groups of lines separated with a blank line perhaps?) with an inline comment describing both.

packages/core/src/sanitization/sanitization.ts
```diff
@@ -196,7 +196,7 @@ export const ɵɵdefaultStyleSanitizer =
       }
 
       if (mode & StyleSanitizeMode.SanitizeOnly) {
-        return doSanitizeValue ? ɵɵsanitizeStyle(value) : value;
+        return doSanitizeValue ? ɵɵsanitizeStyle(value) : unwrapSafeValue(value);
```
IgorMinar: this change is right, but the surrounding code is super complex without any obvious reason for this complexity. :-(

I'll ask @matsko and @mhevery if they could simplify it in the future.

IgorMinar (approved):
lgtm, but please improve the tests as suggested.

AndrewKushnir:
[Presubmit](http://test/OCL:284660417:BASE:284660464:1575938048355:890aa3f6)

packages/core/test/acceptance/styling_spec.ts
AndrewKushnir: Thanks for the review @IgorMinar. I've updated the test to separate props by groups and added comments in a template to make it more explicit. Thank you.

packages/core/src/sanitization/sanitization.ts
matsko: The reasoning for this is because `styleMap()` cannot figure out ahead of time whether or not to sanitize a given property--it can only be figured out during runtime (which is why we need the big if statement above).

If this were a class then we could have two methods that do validation + sanitization (which would make it easier to follow), but we need to stick to functions because of runtime bundle size concerns.

matsko:
LGTM

matsko (approved):

packages/core/src/sanitization/sanitization.ts
IgorMinar: @matsko couldn't we have a private utility function (e.g. `isSanitizationRequired(propName: string): boolean`) that is used by both the ɵɵdefaultStyleSanitizer and `styleMap` directly?

PR #34290, opened 2019-12-07
Cleanup ivy

mhevery:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/render3/instructions/shared.ts
+++ b/packages/core/src/render3/instructions/shared.ts
@@ -55,13 +55,14 @@ const _CLEAN_PROMISE = (() => Promise.resolve(null))();
 export function setHostBindings(tView: TView, lView: LView): void {
   const selectedIndex = getSelectedIndex();
   try {
-    if (tView.expandoInstructions !== null) {
+    const expandoInstructions = tView.expandoInstructions;
+    if (expandoInstructions !== null) {
       let bindingRootIndex = setBindingIndex(tView.expandoStartIndex);
       setBindingRoot(bindingRootIndex);
       let currentDirectiveIndex = -1;
       let currentElementIndex = -1;
-      for (let i = 0; i < tView.expandoInstructions.length; i++) {
-        const instruction = tView.expandoInstructions[i];
+      for (let i = 0; i < expandoInstructions.length; i++) {
+        const instruction = expandoInstructions[i];
         if (typeof instruction === 'number') {
           if (instruction <= 0) {
             // Negative numbers mean that we are starting new EXPANDO block and need to update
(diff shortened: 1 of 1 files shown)


packages/core/src/render3/instructions/shared.ts
```diff
@@ -1216,7 +1217,10 @@ export function generateExpandoInstructionBlock(
                    tView.firstCreatePass, true,
                    'Expando block should only be generated on first create pass.');
 
-  const elementIndex = -(tNode.index - HEADER_OFFSET);
+  // Important: In JS `-x` and `0-x` is not the same! If `x===0` then `-x` will produce `-0` which
+  // requires non standard math arithmetic and it can prevent VM optimizations.
+  // `0-0` will always produce `0` and will not cause a potential deoptimization in VM.
+  const elementIndex = 0 - tNode.index - HEADER_OFFSET;
```
petebacondarwin:
```suggestion
  const elementIndex = 0 - tNode.index + HEADER_OFFSET;
```

ocombe: why not just use `HEADER_OFFSET - tNode.index`?

petebacondarwin: Even better!

packages/core/src/render3/instructions/shared.ts
```diff
@@ -1216,7 +1217,10 @@ export function generateExpandoInstructionBlock(
                    tView.firstCreatePass, true,
                    'Expando block should only be generated on first create pass.');
 
-  const elementIndex = -(tNode.index - HEADER_OFFSET);
+  // Important: In JS `-x` and `0-x` is not the same! If `x===0` then `-x` will produce `-0` which
+  // requires non standard math arithmetic and it can prevent VM optimizations.
```
gkalpak: Very curious to know how you found that out.

pkozlowski-opensource: @gkalpak I'm not sure if this is the method that @mhevery used but if you run V8 with the `--trace-deopt` option (you can pass it to the node directly or via ` --js-flags="--trace-deopt"` to Chrome) you will get logs when v8 (TurboFan, actually) decides to de-optimise (drop) the generated code when one of the assumptions / checks in the generated code fails.

I _presume_ that in this particular case v8 assumed that all the numbers coming down a particular point are smi (small integers) but `-0` is not a smi so v8 had to bail out. At least this was the only "hard" (eager) de-optimisations I've seen while running benchmarks, but I couldn't figure out what was going on!

I'm also curious about the exact path that lead to the @mhevery discovery and hist analysis, but I'm presume it will be the above. In any case huge props for nailing this one @mhevery !

mhevery: You are giving me more credit than I deserve. :-) I was debugging something and I got `-0` and was very confused as to how in the world is `-0` encoded into an integer (SMI). So did some research and realized that `-0` is encoded as floating point number. At this point I just assume that is a bad idea to mix integers and floats.

packages/core/src/render3/instructions/shared.ts
mhevery: Good idea, will change it.

pkozlowski-opensource (approved):
LGTM

PR #34296, opened 2019-12-07
fix(ivy): inherit static coercion members from base classes

### LINKED ISSUE #33830: ngtsc: acceptance coercion members should be inherited
For ngtsc, the concept of acceptance coercion members has been introduced:

```ts
export class MatSelect {
  @Input()
  get disabled(): boolean {..}
  set disabled(val: boolean) { this._disabled = coerceToBoolean(val); }

  // this allows consumers to use <select disabled="">
  static ngAcceptInputType_disabled = boolean | string;
}
```

Unfortunately it's currently not intuitive that those acceptance members are _not_ inherited. This means that consumers of the component, or library authors who share logic in base classes need to copy the acceptance members manually to derived classes. e.g.

```ts
export class CustomSelect extends MatSelect {
  // PROBLEM: this needs to be re-created.. even though we extend from MatSelect.
  static ngAcceptInputType_disabled = typeof MatSelect.ngAcceptInputType_disabled;
}
```

The acceptance members are technically inherited (at runtime) in the `CustomSelect` but just need to be properly picked up by ngtsc. This would match the mental model where the actual inputs are inherited too.

Just an idea of how it could work:

```ts
export declare class Test {
  test: boolean;
  static ngTypeCtor(init: {test: boolean|string}): Test;
}

export declare class Test2 extends Test {
  static ngTypeCtor(init: Parameters<typeof Test.ngTypeCtor>[0]): Test2;
}

Test2.ngTypeCtor({test: 'a'})
```

JoostK:
See individual commits for descriptions

Fixes #33830

### DIFF
--- a/packages/compiler-cli/src/metadata/collector.ts
+++ b/packages/compiler-cli/src/metadata/collector.ts
@@ -230,7 +230,7 @@ export class MetadataCollector {
             const property = <ts.PropertyDeclaration>member;
             if (isStatic(property)) {
               const name = evaluator.nameOf(property.name);
-              if (!isMetadataError(name)) {
+              if (!isMetadataError(name) && !shouldIgnoreStaticMember(name)) {
                 if (property.initializer) {
                   const value = evaluator.evaluateNode(property.initializer);
                   recordStaticMember(name, value);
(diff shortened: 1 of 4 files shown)


alxhub (approved):
Nice work, this will be super helpful.

alxhub:
[Presubmit](https://test.corp.google.com/ui#id=OCL:284797539:BASE:284797581:1576002335148:2143f495)
[Ivy Presubmit](https://test.corp.google.com/ui#id=OCL:284797539:BASE:284797581:1576002335078:8bb17b91)

PR #34304, opened 2019-12-08
fix(ivy): various IE issues

crisbeto:
Since we haven't been running the Ivy tests against IE for a while, we've accumulated some errors and test failures. These changes fix the issues which come from a handful of root causes:

**i18n instructions thrown off by sanitizer:**

While sanitizing on browsers that don't support the `template` element (pretty much only IE), we create an inert document and we insert content into it via `document.body.innerHTML = unsafeHTML`. The problem is that IE appears to parse the HTML passed to `innerHTML` differently, depending on whether the element has been inserted into a document or not. In particular, it seems to split some strings into multiple text nodes, which would've otherwise been a single node. This ended up throwing off some of the i18n code down the line and causing a handful of failures. I've worked around it by creating a new inert `body` element into which the HTML would be inserted.

**Proxies being used in DebugElement.classes:**

We're currently using proxies to pick up changes to element classes coming outside of Angular. Proxies aren't supported in IE and error is always thrown. I've reworked the tests where possible and skipped the rest with a TODO to come back and re-enable them once we have som kind of fallback.

**Inconsistent `typeof Node` value:**

We have a couple of cases where we use something like `typeof Node === 'function'` to figure out whether we're in a worker context. This works in most browsers, but IE returns `object` instead of `function`. I've updated all the usages to account for it.

**IE saving the attribute case:**

In `DebugElement.attributes` we return all of the attributes from the underlying DOM node. Most browsers change the attribute names to lower case, but IE preserves the case and since we use camel-cased attributes, the return value was inconsitent. I've changed it to always lower case the attribute names.

**Unable to get own function name:**

When we log DI errors we get the name of the provider via `SomeClass.name`. In IE functions that inherit from other functions don't have their own `name`, but they take the `name` from the lowest parent in the chain, before `Function`. I've added some changes to fall back to parsing out the function name from the function's string form.

**Different attribute order in innerHTML:**

We've got some tests that assert that the generate DOM looks correct. The problem is that IE changes the attribute order in `innerHTML` which caused the tests to fail. I've reworked the relevant tests not to assert directly against `innerHTML`.

**__proto__ not supported:**

IE doesn't support `__proto__`. I've added a fallback based on #34279.

### DIFF
--- a/packages/core/test/test_bed_spec.ts
+++ b/packages/core/test/test_bed_spec.ts
@@ -989,4 +989,17 @@ describe('TestBed', () => {
                  .toEqual(originalResolver);
            });
       });
+
+  onlyInIvy('VE injects undefined when provider does not have useValue or useFactory')
+      .describe('overrides provider', () => {
+        it('with empty provider object', () => {
+          @Injectable()
+          class Service {
+          }
+          TestBed.overrideProvider(Service, {});
+          // Should be able to get a Service instance because it has no dependencies that can't be
+          // resolved
+          expect(TestBed.inject(Service)).toBeDefined();
+        });
+      });
 });
--- a/packages/core/testing/src/r3_test_bed_compiler.ts
+++ b/packages/core/testing/src/r3_test_bed_compiler.ts
@@ -158,14 +158,19 @@ export class R3TestBedCompiler {
   overrideProvider(
       token: any,
       provider: {useFactory?: Function, useValue?: any, deps?: any[], multi?: boolean}): void {
-    const providerDef = provider.useFactory ?
-        {
-          provide: token,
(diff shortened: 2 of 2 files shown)


.circleci/config.yml
```diff
@@ -356,6 +356,31 @@ jobs:
       - notify_webhook_on_fail:
           webhook_url_env_var: SLACK_DEV_INFRA_CI_FAILURES_WEBHOOK_URL
 
+  saucelabs_ivy:
```
crisbeto: Note that this is copied over from https://github.com/angular/angular/pull/34277 so that I can check for more failures. I'll revert it before merging.

PR #34305, opened 2019-12-08
fix(ivy): various IE issues

crisbeto:
Since we haven't been running the Ivy tests against IE for a while, we've accumulated some errors and test failures. These changes fix the issues which come from a handful of root causes:

**i18n instructions thrown off by sanitizer:**

While sanitizing on browsers that don't support the `template` element (pretty much only IE), we create an inert document and we insert content into it via `document.body.innerHTML = unsafeHTML`. The problem is that IE appears to parse the HTML passed to `innerHTML` differently, depending on whether the element has been inserted into a document or not. In particular, it seems to split some strings into multiple text nodes, which would've otherwise been a single node. This ended up throwing off some of the i18n code down the line and causing a handful of failures. I've worked around it by creating a new inert `body` element into which the HTML would be inserted.

**Inheriting injectable definition from parent class not working in IE10 JIT mode:**

The way definitions are added in JIT mode is through `Object.defineProperty`, but the problem is that in IE10 properties defined through `defineProperty` won't be inherited which means that inheriting injectable definitions no longer works. These changes add a workaround only for JIT mode where we define a fallback method for retrieving the definition. This isn't ideal, but it should only be required until v10 where we'll no longer support inheriting injectable definitions from undecorated classes.

**Inheritance issues in IE10:**

It looks like most of the inheritance functionality in Ivy wasn't working in JIT mode because we weren't accessing things correctly. These changes fix all of the cases.

**Proxies being used in DebugElement.classes:**

We're currently using proxies to pick up changes to element classes coming outside of Angular. Proxies aren't supported in IE and an error is always thrown. I've reworked the tests where possible and skipped the rest with a TODO to come back and re-enable them once we have some kind of fallback.

**Inconsistent `typeof Node` value:**

We have a couple of cases where we use something like `typeof Node === 'function'` to figure out whether we're in a worker context. This works in most browsers, but IE returns `object` instead of `function`. I've updated all the usages to account for it.

**IE saving the attribute case:**

In `DebugElement.attributes` we return all of the attributes from the underlying DOM node. Most browsers change the attribute names to lower case, but IE preserves the case and since we use camel-cased attributes, the return value was inconsistent. I've changed it to always lower case the attribute names.

**Unable to get own function name:**

When we log DI errors we get the name of the provider via `SomeClass.name`. In IE functions that inherit from other functions don't have their own `name`, but they take the `name` from the lowest parent in the chain, before `Function`. I've added some changes to fall back to parsing out the function name from the function's string form.

**Different attribute order in innerHTML:**

We've got some tests that assert that the generated DOM looks correct. The problem is that IE changes the attribute order in `innerHTML` which caused the tests to fail. I've reworked the relevant tests not to assert directly against `innerHTML`.

**__proto__ not supported:**

IE doesn't support `__proto__`. I've added a fallback based on #34279.

### DIFF
--- a/packages/core/src/di/jit/util.ts
+++ b/packages/core/src/di/jit/util.ts
@@ -48,13 +48,17 @@ function reflectDependency(compiler: CompilerFacade, dep: any | any[]): R3Depend
       if (param === undefined) {
         // param may be undefined if type of dep is not set by ngtsc
         continue;
-      } else if (param instanceof Optional || param.__proto__.ngMetadataName === 'Optional') {
+      }
+
+      const proto = Object.getPrototypeOf(param);
+
+      if (param instanceof Optional || proto.ngMetadataName === 'Optional') {
         meta.optional = true;
-      } else if (param instanceof SkipSelf || param.__proto__.ngMetadataName === 'SkipSelf') {
+      } else if (param instanceof SkipSelf || proto.ngMetadataName === 'SkipSelf') {
         meta.skipSelf = true;
-      } else if (param instanceof Self || param.__proto__.ngMetadataName === 'Self') {
+      } else if (param instanceof Self || proto.ngMetadataName === 'Self') {
         meta.self = true;
-      } else if (param instanceof Host || param.__proto__.ngMetadataName === 'Host') {
+      } else if (param instanceof Host || proto.ngMetadataName === 'Host') {
         meta.host = true;
       } else if (param instanceof Inject) {
         meta.token = param.token;
--- a/packages/core/src/render3/instructions/element.ts
+++ b/packages/core/src/render3/instructions/element.ts
@@ -271,7 +271,10 @@ function validateElement(
     // as a custom element. Note that unknown elements with a dash in their name won't be instances
     // of HTMLUnknownElement in browsers that support web components.
     const isUnknown =
-        (typeof HTMLUnknownElement === 'function' && element instanceof HTMLUnknownElement) ||
+        // Note that we can't check for `typeof HTMLUnknownElement === 'function'`,
+        // because while most browsers return 'function', IE returns 'object'.
+        (typeof HTMLUnknownElement !== 'undefined' && HTMLUnknownElement &&
+         element instanceof HTMLUnknownElement) ||
         (typeof customElements !== 'undefined' && tagName.indexOf('-') > -1 &&
          !customElements.get(tagName));
 
--- a/packages/core/src/render3/jit/directive.ts
+++ b/packages/core/src/render3/jit/directive.ts
@@ -189,7 +189,8 @@ export function extendsDirectlyFromObject(type: Type<any>): boolean {
  */
 export function directiveMetadata(type: Type<any>, metadata: Directive): R3DirectiveMetadataFacade {
   // Reflect inputs and outputs.
-  const propMetadata = getReflect().ownPropMetadata(type);
+  const reflect = getReflect();
+  const propMetadata = reflect.ownPropMetadata(type);
 
   return {
     name: type.name,
@@ -202,7 +203,7 @@ export function directiveMetadata(type: Type<any>, metadata: Directive): R3Direc
     inputs: metadata.inputs || EMPTY_ARRAY,
     outputs: metadata.outputs || EMPTY_ARRAY,
     queries: extractQueriesMetadata(type, propMetadata, isContentQuery),
-    lifecycle: {usesOnChanges: usesLifecycleHook(type, 'ngOnChanges')},
+    lifecycle: {usesOnChanges: reflect.hasLifecycleHook(type, 'ngOnChanges')},
     typeSourceSpan: null !,
     usesInheritance: !extendsDirectlyFromObject(type),
     exportAs: extractExportAs(metadata.exportAs),
@@ -216,7 +217,7 @@ export function directiveMetadata(type: Type<any>, metadata: Directive): R3Direc
  */
 function addDirectiveDefToUndecoratedParents(type: Type<any>) {
   const objPrototype = Object.prototype;
-  let parent = Object.getPrototypeOf(type);
+  let parent = Object.getPrototypeOf(type.prototype).constructor;
 
   // Go up the prototype until we hit `Object`.
   while (parent && parent !== objPrototype) {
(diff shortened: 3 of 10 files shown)


.circleci/config.yml
```diff
@@ -356,6 +356,31 @@ jobs:
       - notify_webhook_on_fail:
           webhook_url_env_var: SLACK_DEV_INFRA_CI_FAILURES_WEBHOOK_URL
 
+  saucelabs_ivy:
```
crisbeto: Note that this is copied over from #34277 so that I can check for more failures. I'll revert it before merging.

packages/core/src/render3/instructions/element.ts
```diff
@@ -271,7 +271,10 @@ function validateElement(
     // as a custom element. Note that unknown elements with a dash in their name won't be instances
     // of HTMLUnknownElement in browsers that support web components.
     const isUnknown =
-        (typeof HTMLUnknownElement === 'function' && element instanceof HTMLUnknownElement) ||
+        // Note that we can't check for `typeof HTMLUnknownElement === 'function'`,
+        // because while most browsers return 'function', IE returns 'object'.
+        (typeof HTMLUnknownElement !== 'undefined' && HTMLUnknownElement &&
```
gkalpak: Why the `HTMLUnknownElement` check? (Wouldn't `!== 'undefined'` and `instanceof` be enough?)

packages/core/src/render3/jit/directive.ts
```diff
@@ -325,3 +323,19 @@ function shouldAddAbstractDirective(type: Type<any>): boolean {
 
   return false;
 }
+
+
+function getTypeName(type: any): string {
+  // `Function.prototype.name` behaves differently between IE and other browsers. In most browsers
+  // it'll always return the name of the function itself, no matter ho many other functions it
```
gkalpak: Typo: `ho` --> `how`

packages/core/src/render3/jit/directive.ts
```diff
@@ -325,3 +323,19 @@ function shouldAddAbstractDirective(type: Type<any>): boolean {
...
+  // inherits from. On IE the function doesn't have its own `name` property, but it takes it from
+  // the lowest level in the prototype chain. E.g. if we have `class Foo extends Parent` most
+  // browsers will evaluate `Foo.name` to `Foo` while IE will return `Parent`. We work around
+  // the issue by converting the function to a string and parsing its name out that way via a regex.
+  if (type.hasOwnProperty('name')) {
+    return type.name;
+  }
+
+  const match = (type.toString() as string).match(/^function\s*([^\s(]+)/);
```
gkalpak: Is `type` guaranteed to have a `toString()` method? (If not, something like `'' + type` would be safer.)

packages/core/src/di/jit/util.ts
```diff
@@ -51,13 +51,17 @@ function reflectDependency(compiler: CompilerFacade, dep: any | any[]): R3Depend
       if (param === undefined) {
         // param may be undefined if type of dep is not set by ngtsc
         continue;
-      } else if (param instanceof Optional || param.__proto__.ngMetadataName === 'Optional') {
+      }
+
+      const proto = param.__proto__ || Object.getPrototypeOf(param);
```
gkalpak: Since we are using `Object.getPrototypeOf()` elsewhere (and `__proto__` is considered a bad practice), why not use `Object.getPrototypeOf()` in all cases?

(Also, why care about the prototype in the first place? Why not just `param.ngMetadataName`?)

packages/core/test/acceptance/styling_spec.ts
```diff
@@ -2172,6 +2147,13 @@ describe('styling', () => {
           });
 
   it('should retrieve styles set via Renderer2', () => {
+    // TODO(FW-1767): retrieving classes set through Renderer2 currently depends on Proxy which
```
gkalpak: 😱

gkalpak (approved):

packages/core/src/render3/instructions/element.ts
crisbeto: It's unlikely, but if it's null, `typeof HTMLUnknownElement` will be `object`.

packages/core/src/di/jit/util.ts
crisbeto: I'm not sure about the reason tbh. I left it in assuming there was a reason for it. I can try dropping it to see if anything breaks.

packages/core/src/render3/instructions/element.ts
gkalpak: Makes sense :+1: (Although I have no idea why it might be `null`, it doesn't hurt to play it safe.) 
Consider making the intention more explicit by chaning it to `HTMLUnknownElement !== null` then.

.circleci/config.yml
pkozlowski-opensource: I think that we should do the following:
* squash all the commits from #34277
* pick-up (cherry-pick) one commit here.

It would be much easier to see what is part of this PR and what is not if the CI work would be a separate commit.

packages/core/src/di/jit/util.ts
pkozlowski-opensource: >  I can try dropping it to see if anything breaks.
yeh, using `getPrototypeOf` consistently would be the best.

.circleci/config.yml
crisbeto: I can do it when I get home, but it's basically just the `.yml` file changes that have been taken from #34277.

devversion: Either that, or the #34277 can be just merged afterwards (though I could see the point of having an indicator that this PR actually fixes IE issues)

pkozlowski-opensource: > though I could see the point of having an indicator that this PR actually fixes IE issues

yeh, we should be able to see that the issues are actually fixed...

devversion: Okay, my other PR is ready for merge, so Kristiyan can just pull in the `.patch` for that PR.

crisbeto:
Addressed the feedback from @pkozlowski-opensource and @gkalpak.

packages/core/src/render3/jit/directive.ts
```diff
@@ -325,3 +323,19 @@ function shouldAddAbstractDirective(type: Type<any>): boolean {
 
   return false;
 }
+
+
+function getTypeName(type: any): string {
```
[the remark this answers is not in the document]
crisbeto: This one isn't necessary and was removed yesterday. I had copied it here for easier debugging since IE10's console logs objects as strings.

pkozlowski-opensource (changes requested):
@crisbeto could you please pick up Paul's commits from https://github.com/angular/angular/pull/34277 as discussed?

Also, ideally I would like to see work-around related to `Proxy` usage (FW-1767) in a separate commit. The reason for this is that if we decide to merge #34328 (and it looks like we will) we might not want to merge your PR with work-arounds.

Generally speaking, I'm expecting that we are only going to merge parts of the work here so having small, focused commits would make the whole operation easier (I know it is more work for you, but it will help the overall effort)

crisbeto:
@pkozlowski-opensource the problem with cherry picking the commits is that the CLA bot will start complaining if there are mixed authors. I took Paul's changes just to illustrate that everything is passing, but I'll remove them before setting it as ready to merge.

Regarding the `Proxy` error: this is the reason why I left the TODO in there and mentioned it in the issue. If we get it fixed before this PR goes in, I'll rebase and remove my workarounds.

crisbeto:
The build changes have been reverted.

kara (approved):
Thanks for all these fixes!  Changes LGTM, but we should split the fixes into separate commits as @pkozlowski-opensource suggests. If there are unintended consequences (e.g. in G3) and we have to revert something, would rather not revert all changes. Maybe a natural way to split would be by section in the PR description?

crisbeto:
I've split it up into commits @kara.

kara (approved):

pkozlowski-opensource (commented):
Thnx so much for splitting things up into individual commits - this is _so_ much cleaner and easier to review (plus easier to revert individual changes if needed)! Thnx again for doing this! 

I will keep this PR as  "Request changes" so it doesn't get merged accidentally. We will land it as soon as the proxy issues is resolved.

crisbeto:
I've dropped the commit that works around the `Proxy` usages since we agreed to remove them.

petebacondarwin (approved):
LGTM for i18n

pkozlowski-opensource (approved):

mhevery (approved):

kara:
[presubmit](http://test/OCL:285457504:BASE:285457538:1576271669561:283acb72)

PR #34328, opened 2019-12-10
refactor(ivy): remove usage of Proxy for IE10/11 compatibility

pkozlowski-opensource:
Opening a PR to gauge impact of a breaking change in `DebugElement.classes`.

### Context

The `DebugElement` exposes the `classes` getter with the following structure signature: `readonly classes: {[key: string]: boolean};`. The idea behind this data structure is to expose information about CSS classes present on a given element.

To illustrate it let's consider the following example:
`<div class="foo {{exp}}"">` where `exp` is initially an empty string (`''`).

Given the above we would expect the following:
* `expect(debugElement.classes['foo']).toBe(true)`
* `expect(debugElement.classes['bar']).toBe(undefined)`
and this is the case. 

Now let's assume that `exp` flips to `'bar'` so we can expect the following:
* `expect(debugElement.classes['foo']).toBe(true)`
* `expect(debugElement.classes['bar']).toBe(true)`
and this holds true.

Now, let's flip `exp` back to `'bar'` to an empty string (`''`):
* `expect(debugElement.classes['foo']).toBe(true)`
* `expect(debugElement.classes['bar']).toBe(???)`

The question here is this: what the value of `debugElement.classes['bar']` should be? `undefined` or `false`?

In VE the answer was `false` - in other words, VE was tracking not only _presence_ of a given class but also its _state_ in the past. Tracking such state would require a testing-only, mutable data structure.

### Implementation choices

The VE implementation based on debug services was a source of memory leaks and increased complexity so we've decided to change this implementation in ivy. Unfortunately, the side effect of this implementation change is that to support VE semantics we had to resort to usage of `Proxy` [which is not supported in IE10/11](https://caniuse.com/#search=proxy).

Seems like we are facing a choice here between  IE10/11 support and a breaking change to the data structure returned by `DebugElement.classes`

### Breaking change

Removal of `Proxy` usage would result in a breaking change for the following situation:
* a CSS class is added to an element and then _subsequently removed_;
* a test asserts on the _lack_ of a CSS class in the following way:  `expect(debugElement.classes['bar']).toBe(false)`
 
A change in a test is simple (`expect(debugElement.classes['bar']).toBe(false)` => `expect(debugElement.classes['bar']).toBeFalsy()`) but it is a breaking change that might have negligible impact on the existing code.

As the additional data point - so far we've decided to go with such breaking change for the similar  `debugElement.attributes` collection (the reasoning being that the "never present" vs. "was there at one point but got removed" situations should be indistinguishable from a test point of view). 

### Notes on the API design

We are having those discussions partly to the API design choices. The `debugElement.classes`, as it stands today, exposes too much implementation details. A better API might look like follows:
`debugElement.hasClass('foo')`

### DIFF
--- a/aio/content/guide/ivy-compatibility.md
+++ b/aio/content/guide/ivy-compatibility.md
@@ -57,3 +57,7 @@ If the errors are gone, switch back to Ivy by removing the changes to the `tscon
 * ICU parsing happens at runtime, so only text, HTML tags and text bindings are allowed inside ICU cases (previously, directives were also permitted inside ICUs).
 
 * Providers formatted as `{provide: X}` without a `useValue`, `useFactory`, `useExisting`, or `useClass` property are treated like `{provide: X, useClass: X}` (previously, it defaulted to `{provide: X, useValue: undefined}`).
+
+* `DebugElement.attributes` returns `undefined` for attributes that were added and then subsequently removed (previously, attributes added and later removed would have a value of `null`).
+
+* `DebugElement.classes` returns `undefined` for classes that were added and then subsequently removed (previously, classes added and later removed would have a value of `false`).
--- a/packages/core/test/acceptance/styling_spec.ts
+++ b/packages/core/test/acceptance/styling_spec.ts
@@ -2238,9 +2238,9 @@ describe('styling', () => {
       constructor(public elementRef: ElementRef, public renderer: Renderer2) { dirInstance = this; }
 
       setStyles() {
-        this.renderer.setStyle(
-            this.elementRef.nativeElement, 'transform', 'translate3d(0px, 0px, 0px)');
-        this.renderer.addClass(this.elementRef.nativeElement, 'my-class');
+        const nativeEl = this.elementRef.nativeElement;
+        this.renderer.setStyle(nativeEl, 'transform', 'translate3d(0px, 0px, 0px)');
+        this.renderer.addClass(nativeEl, 'my-class');
       }
     }
 
@@ -2258,16 +2258,6 @@ describe('styling', () => {
     const div = fixture.debugElement.children[0];
     expect(div.styles.transform).toMatch(/translate3d\(0px\s*,\s*0px\s*,\s*0px\)/);
     expect(div.classes['my-class']).toBe(true);
-
-    div.classes['other-class'] = true;
-    div.styles['width'] = '200px';
-    expect(div.styles.width).toEqual('200px');
-    expect(div.classes['other-class']).toBe(true);
-
-    if (ivyEnabled) {
-      expect(div.nativeElement.classList.contains('other-class')).toBeTruthy();
-      expect(div.nativeElement.style['width']).toEqual('200px');
-    }
   });
 
   it('should not set classes when falsy value is passed while a sanitizer is present', () => {
--- a/packages/core/test/debug/debug_node_spec.ts
+++ b/packages/core/test/debug/debug_node_spec.ts
@@ -354,7 +354,7 @@ class TestCmptWithPropInterpolation {
       const bankElem = fixture.debugElement.children[0];
 
       expect(bankElem.classes['closed']).toBe(true);
-      expect(bankElem.classes['open']).toBe(false);
+      expect(bankElem.classes['open']).toBeFalsy();
     });
 
     it('should get element classes from host bindings', () => {
@@ -365,7 +365,7 @@ class TestCmptWithPropInterpolation {
       expect(debugElement.classes['present-class'])
           .toBe(true, 'Expected bound host CSS class "present-class" to be present');
       expect(debugElement.classes['absent-class'])
-          .toBe(false, 'Expected bound host CSS class "absent-class" to be absent');
+          .toBeFalsy('Expected bound host CSS class "absent-class" to be absent');
     });
 
     it('should list element styles', () => {
--- a/aio/src/app/app.component.spec.ts
+++ b/aio/src/app/app.component.spec.ts
@@ -957,7 +957,7 @@ describe('AppComponent', () => {
         triggerDocViewerEvent('docRendered');
         fixture.detectChanges();
         expect(component.isTransitioning).toBe(false);
-        expect(toolbar.classes['transitioning']).toBe(false);
+        expect(toolbar.classes['transitioning']).toBeFalsy();
 
         // While a document is being rendered, `isTransitoning` is set to true.
         triggerDocViewerEvent('docReady');
@@ -968,7 +968,7 @@ describe('AppComponent', () => {
         triggerDocViewerEvent('docRendered');
         fixture.detectChanges();
         expect(component.isTransitioning).toBe(false);
-        expect(toolbar.classes['transitioning']).toBe(false);
+        expect(toolbar.classes['transitioning']).toBeFalsy();
       });
 
       it('should update the sidenav state as soon as a new document is inserted (but not before)', () => {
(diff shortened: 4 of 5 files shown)


packages/core/test/acceptance/styling_spec.ts
```diff
@@ -2200,16 +2200,6 @@ describe('styling', () => {
     const div = fixture.debugElement.children[0];
     expect(div.styles.transform).toMatch(/translate3d\(0px\s*,\s*0px\s*,\s*0px\)/);
     expect(div.classes['my-class']).toBe(true);
-
-    div.classes['other-class'] = true;
-    div.styles['width'] = '200px';
-    expect(div.styles.width).toEqual('200px');
-    expect(div.classes['other-class']).toBe(true);
```
pkozlowski-opensource: Note to a reviewer: I wasn't able to resolve what we are trying to assert here :-/ The way I was reading it is that we are setting a key on an object and assert that a value was set on the said object.

packages/core/test/acceptance/styling_spec.ts
```diff
@@ -2200,16 +2200,6 @@ describe('styling', () => {
...
-
-    if (ivyEnabled) {
-      expect(div.nativeElement.classList.contains('other-class')).toBeTruthy();
-      expect(div.nativeElement.style['width']).toEqual('200px');
-    }
```
pkozlowski-opensource: So this assert is suspicious as it seems to imply that setting a value of a debug node should propagate a change to a backing native element. I'm not sure why we should have such logic, especially given that VE wasn't doing it... 
I'm certainly missing context here...

packages/core/test/debug/debug_node_spec.ts
```diff
@@ -354,7 +354,7 @@ class TestCmptWithPropInterpolation {
       const bankElem = fixture.debugElement.children[0];
 
       expect(bankElem.classes['closed']).toBe(true);
-      expect(bankElem.classes['open']).toBe(false);
+      expect(bankElem.classes['open']).toBeFalsy();
```
pkozlowski-opensource: Note to a reviewer: this would be a breaking change - we need to weight the impact on users vs. support for IE10/11. Additional data point: we've decided to do such breaking change for (non-class) `attributes` where we don't put attributes in a retuned object (regardless if the attribute was never present vs. added and removed).

packages/core/test/debug/debug_node_spec.ts
```diff
@@ -365,7 +365,7 @@ class TestCmptWithPropInterpolation {
       expect(debugElement.classes['present-class'])
           .toBe(true, 'Expected bound host CSS class "present-class" to be present');
       expect(debugElement.classes['absent-class'])
-          .toBe(false, 'Expected bound host CSS class "absent-class" to be absent');
+          .toBeFalsy('Expected bound host CSS class "absent-class" to be absent');
```
pkozlowski-opensource: Note to a reviewer: this would be a breaking change - we need to weight the impact on users vs. support for IE10/11. Additional data point: we've decided to do such breaking change for (non-class) `attributes` where we don't put attributes in a retuned object (regardless if the attribute was never present vs. added and removed).

alfaproject:
Not sure if you considered or if it's an option at all but just wondering if a polyfill would help here? Something like https://github.com/GoogleChrome/proxy-polyfill

pkozlowski-opensource:
@alfaproject yes, this would be another option (using an existing polyfill or just grabbing an essential part of it for our use-case.

kara:
[Ivy presubmit](http://test/OCL:284822930:BASE:284822972:1576009150247:55762a52)

packages/core/test/debug/debug_node_spec.ts
kara: Can we add a line item to the compatibility guide in this PR too?

https://next.angular.io/guide/ivy-compatibility

kara (approved):
LGTM pending G3 fixes and comment below

AndrewKushnir:
[Ivy Presubmit #2](http://test/OCL:285088642:BASE:285088678:1576111143991:9cbc2c29)

packages/core/test/debug/debug_node_spec.ts
pkozlowski-opensource: Done in a [separate commit](91dc85485a2b168c73726fb8fc7a17c7dcbc0510)

pkozlowski-opensource:
Assigning to @AndrewKushnir as my understanding is that we need to migrate G3 before we can land this one.

AndrewKushnir:
Results from g3 presubmits:
- Blueprint targets were updated to be compatible with this change
- Global TAP indicated no additional regressions

So the g3 status is now "green" for this PR.

aio/content/guide/ivy-compatibility.md
```diff
@@ -56,4 +56,8 @@ If the errors are gone, switch back to Ivy by removing the changes to the `tscon
 
 * ICU parsing happens at runtime, so only text, HTML tags and text bindings are allowed inside ICU cases (previously, directives were also permitted inside ICUs).
 
-* Providers formatted as `{provide: X}` without a `useValue`, `useFactory`, `useExisting`, or `useClass` property are treated like `{provide: X, useClass: X}` (previously, it defaulted to `{provide: X, useValue: undefined}`).
\ No newline at end of file
+* Providers formatted as `{provide: X}` without a `useValue`, `useFactory`, `useExisting`, or `useClass` property are treated like `{provide: X, useClass: X}` (previously, it defaulted to `{provide: X, useValue: undefined}`).
+
+* The `TestBed`'s `DebugElement.attributes` object returns different values for attributes that were added and then subsequently removed (ex. with a binding like `[attributes.aria-haspopup]="showExp"`). Previously, for such attributes, the following would be true: `expect(element.attributes['aria-haspopup']).toBe(null)`. In ivy a returned value is `undefined` so the following is true: `expect(element.attributes['aria-haspopup']).toBe(undefined)` (or `expect(element.attributes['aria-haspopup']).toBeFalsy()`).
```
kara: Can you use the same format as the other breaking changes?  The breaking change should be one sentence with the previous behavior in parentheses. e.g.


>`DebugElement.attributes` returns `undefined` for attributes that were added and then subsequently removed (previously, attributes added and later removed would have a value of `null`).

aio/content/guide/ivy-compatibility.md
```diff
@@ -56,4 +56,8 @@ If the errors are gone, switch back to Ivy by removing the changes to the `tscon
...
+
+* The `TestBed`'s `DebugElement.classes` returns different values for classes that were added and then subsequently removed (ex. with a binding like `[class.bright]="showExp"`). Previously, for such classes, the following would be true: `expect(element.classes.bright).toBe(false)`. In ivy a returned value is `undefined` so the following is true: `expect(element.classes.bright).toBe(undefined)` (or `expect(element.classes.bright).toBeFalsy()`).
```
kara: >`DebugElement.classes` returns `undefined` for attributes that were added and then subsequently removed (previously, attributes added and later removed would have a value of `false`).

kara (changes requested):
A few comments on the docs

aio/content/guide/ivy-compatibility.md
pkozlowski-opensource: Edited with your suggestions in both places. Could you please re-review @kara ?

kara (approved):
LGTM, thanks for the updates

PR #34331, opened 2019-12-10
fix(bazel): improve performance of tsHost.writeFile()

kyliau:
Removing from an array incurs O(n^2) cost, and could be mitigated with the use of a Set instead.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/bazel/src/ngc-wrapped/index.ts
+++ b/packages/bazel/src/ngc-wrapped/index.ts
@@ -199,16 +199,15 @@ export function compile({allDepsCompiledWithBazel = true, compilerOpts, tsHost, 
     throw new Error(`Couldn't find bazel bin in the rootDirs: ${compilerOpts.rootDirs}`);
   }
 
-  const writtenExpectedOuts = expectedOuts.map(p => p.replace(/\\/g, '/'));
+  const expectedOutsSet = new Set(expectedOuts.map(p => p.replace(/\\/g, '/')));
 
   const originalWriteFile = tsHost.writeFile.bind(tsHost);
   tsHost.writeFile =
       (fileName: string, content: string, writeByteOrderMark: boolean,
        onError?: (message: string) => void, sourceFiles?: ts.SourceFile[]) => {
         const relative = relativeToRootDirs(fileName.replace(/\\/g, '/'), [compilerOpts.rootDir]);
-        const expectedIdx = writtenExpectedOuts.findIndex(o => o === relative);
-        if (expectedIdx >= 0) {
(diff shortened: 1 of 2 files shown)


packages/bazel/src/ngc-wrapped/index.ts
```diff
@@ -199,16 +199,15 @@ export function compile({allDepsCompiledWithBazel = true, compilerOpts, tsHost,
...
   const originalWriteFile = tsHost.writeFile.bind(tsHost);
   tsHost.writeFile =
       (fileName: string, content: string, writeByteOrderMark: boolean,
        onError?: (message: string) => void, sourceFiles?: ts.SourceFile[]) => {
         const relative = relativeToRootDirs(fileName.replace(/\\/g, '/'), [compilerOpts.rootDir]);
-        const expectedIdx = writtenExpectedOuts.findIndex(o => o === relative);
-        if (expectedIdx >= 0) {
-          writtenExpectedOuts.splice(expectedIdx, 1);
+        if (writtenExpectedOuts.has(relative)) {
+          writtenExpectedOuts.delete(relative);
```
dgp1130: If a file is written multiple times, wouldn't this be a problem? Considering this logic was already the case it might not be a problem, but we probably shouldn't assume that the compiler will only write out to a file once.

Also the name `writtenExpectedOuts` is a little off, since they haven't been written yet on line 202. I'd recommend having a `const expectedOuts` set on line 202 with a separate `writtenExpectedOuts` set which gets added on each `tsHost.writeFile`, then take the set difference between the two at the end. I think that's a little clearer in intent though it might be a little slower, I don't think it would have significant performance implications (`tsHost.writeFile` would be effectively the same and that's the bottleneck).

dgp1130 (changes requested):

packages/bazel/src/ngc-wrapped/index.ts
kyliau: > If a file is written multiple times
I don't think we'll ever hit this case. If we do, it might be a compiler bug / setup error.

I agree having two sets would make the intent clearer, but it also requires iterating through one of the sets at the end to figure out the set difference. 
I've instead renamed `writtenExpectedOuts` to `expectedOutsSet` to make it clear that possibly not all files in the set are written out and I also added a comment.

dgp1130 (approved):

PR #34333, opened 2019-12-10
feat(ivy): throw compilation error when providing undecorated classes

crisbeto:
Adds a compilation error if the consumer tries to pass in an undecorated class into the `providers` of an `NgModule`, or the `providers`/`viewProviders` arrays of a `Directive`/`Component`.

### DIFF
--- a/packages/compiler-cli/ngcc/src/analysis/decoration_analyzer.ts
+++ b/packages/compiler-cli/ngcc/src/analysis/decoration_analyzer.ts
@@ -6,7 +6,6 @@
  * found in the LICENSE file at https://angular.io/license
  */
 import {ConstantPool} from '@angular/compiler';
-import {InjectableClassRegistry} from '@angular/compiler-cli/src/ngtsc/metadata/src/registry';
 import * as ts from 'typescript';
 import {ComponentDecoratorHandler, DirectiveDecoratorHandler, InjectableDecoratorHandler, NgModuleDecoratorHandler, PipeDecoratorHandler, ReferencesRegistry, ResourceLoader} from '../../../src/ngtsc/annotations';
 import {CycleAnalyzer, ImportGraph} from '../../../src/ngtsc/cycles';
@@ -14,6 +13,7 @@ import {isFatalDiagnosticError} from '../../../src/ngtsc/diagnostics';
 import {FileSystem, LogicalFileSystem, absoluteFrom, dirname, resolve} from '../../../src/ngtsc/file_system';
 import {AbsoluteModuleStrategy, LocalIdentifierStrategy, LogicalProjectStrategy, ModuleResolver, NOOP_DEFAULT_IMPORT_RECORDER, PrivateExportAliasingHost, Reexport, ReferenceEmitter} from '../../../src/ngtsc/imports';
 import {CompoundMetadataReader, CompoundMetadataRegistry, DtsMetadataReader, LocalMetadataRegistry} from '../../../src/ngtsc/metadata';
+import {InjectableClassRegistry} from '../../../src/ngtsc/metadata/src/registry';
 import {PartialEvaluator} from '../../../src/ngtsc/partial_evaluator';
 import {ClassDeclaration} from '../../../src/ngtsc/reflection';
 import {LocalModuleScopeRegistry, MetadataDtsModuleScopeResolver} from '../../../src/ngtsc/scope';
--- a/packages/compiler-cli/ngcc/src/analysis/decoration_analyzer.ts
+++ b/packages/compiler-cli/ngcc/src/analysis/decoration_analyzer.ts
@@ -6,6 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 import {ConstantPool} from '@angular/compiler';
+import {InjectableClassRegistry} from '@angular/compiler-cli/src/ngtsc/metadata/src/registry';
 import * as ts from 'typescript';
 import {ComponentDecoratorHandler, DirectiveDecoratorHandler, InjectableDecoratorHandler, NgModuleDecoratorHandler, PipeDecoratorHandler, ReferencesRegistry, ResourceLoader} from '../../../src/ngtsc/annotations';
 import {CycleAnalyzer, ImportGraph} from '../../../src/ngtsc/cycles';
@@ -85,6 +86,7 @@ export class DecorationAnalyzer {
       new PartialEvaluator(this.reflectionHost, this.typeChecker, /* dependencyTracker */ null);
   importGraph = new ImportGraph(this.moduleResolver);
   cycleAnalyzer = new CycleAnalyzer(this.importGraph);
+  injectableRegistry = new InjectableClassRegistry();
   handlers: DecoratorHandler<unknown, unknown, unknown>[] = [
     new ComponentDecoratorHandler(
         this.reflectionHost, this.evaluator, this.fullRegistry, this.fullMetaReader,
@@ -92,28 +94,28 @@ export class DecorationAnalyzer {
         /* defaultPreserveWhitespaces */ false,
         /* i18nUseExternalIds */ true, this.bundle.enableI18nLegacyMessageIdFormat,
         this.moduleResolver, this.cycleAnalyzer, this.refEmitter, NOOP_DEFAULT_IMPORT_RECORDER,
-        NOOP_DEPENDENCY_TRACKER, /* annotateForClosureCompiler */ false),
+        NOOP_DEPENDENCY_TRACKER, this.injectableRegistry, /* annotateForClosureCompiler */ false),
     // See the note in ngtsc about why this cast is needed.
     // clang-format off
     new DirectiveDecoratorHandler(
         this.reflectionHost, this.evaluator, this.fullRegistry, this.scopeRegistry,
-        NOOP_DEFAULT_IMPORT_RECORDER, this.isCore,
+        NOOP_DEFAULT_IMPORT_RECORDER, this.injectableRegistry, this.isCore,
         /* annotateForClosureCompiler */ false) as DecoratorHandler<unknown, unknown, unknown>,
     // clang-format on
     // Pipe handler must be before injectable handler in list so pipe factories are printed
     // before injectable factories (so injectable factories can delegate to them)
     new PipeDecoratorHandler(
         this.reflectionHost, this.evaluator, this.metaRegistry, this.scopeRegistry,
-        NOOP_DEFAULT_IMPORT_RECORDER, this.isCore),
+        NOOP_DEFAULT_IMPORT_RECORDER, this.injectableRegistry, this.isCore),
     new InjectableDecoratorHandler(
         this.reflectionHost, NOOP_DEFAULT_IMPORT_RECORDER, this.isCore,
-        /* strictCtorDeps */ false, /* errorOnDuplicateProv */ false),
+        /* strictCtorDeps */ false, this.injectableRegistry, /* errorOnDuplicateProv */ false),
     new NgModuleDecoratorHandler(
         this.reflectionHost, this.evaluator, this.fullMetaReader, this.fullRegistry,
         this.scopeRegistry, this.referencesRegistry, this.isCore, /* routeAnalyzer */ null,
         this.refEmitter,
         /* factoryTracker */ null, NOOP_DEFAULT_IMPORT_RECORDER,
-        /* annotateForClosureCompiler */ false),
+        /* annotateForClosureCompiler */ false, this.injectableRegistry),
   ];
   migrations: Migration[] = [
     new UndecoratedParentMigration(),
--- a/packages/compiler-cli/ngcc/src/analysis/decoration_analyzer.ts
+++ b/packages/compiler-cli/ngcc/src/analysis/decoration_analyzer.ts
@@ -86,7 +86,7 @@ export class DecorationAnalyzer {
       new PartialEvaluator(this.reflectionHost, this.typeChecker, /* dependencyTracker */ null);
   importGraph = new ImportGraph(this.moduleResolver);
   cycleAnalyzer = new CycleAnalyzer(this.importGraph);
-  injectableRegistry = new InjectableClassRegistry();
+  injectableRegistry = new InjectableClassRegistry(this.reflectionHost);
   handlers: DecoratorHandler<unknown, unknown, unknown>[] = [
     new ComponentDecoratorHandler(
         this.reflectionHost, this.evaluator, this.fullRegistry, this.fullMetaReader,
--- a/packages/compiler-cli/src/ngtsc/annotations/src/component.ts
+++ b/packages/compiler-cli/src/ngtsc/annotations/src/component.ts
@@ -192,18 +192,27 @@ export class ComponentDecoratorHandler implements
       }
     }, undefined) !;
 
-    const rawViewProviders = component.get('viewProviders');
-    const wrappedViewProviders: Expression|null = rawViewProviders ?
-        new WrappedNodeExpr(
-            this.annotateForClosureCompiler ? wrapFunctionExpressionsInParens(rawViewProviders) :
-                                              rawViewProviders) :
-        null;
-    const viewProvidersRequiringFactory = rawViewProviders !== undefined ?
-        resolveProvidersRequiringFactory(rawViewProviders, this.evaluator) :
-        null;
-    const providersRequiringFactory = component.has('providers') ?
-        resolveProvidersRequiringFactory(component.get('providers') !, this.evaluator) :
-        null;
+
+    // Note that we could technically combine the `viewProvidersRequiringFactory` and
+    // `providersRequiringFactory` into a single set, but we keep the separate so that
+    // we can distinguish where an error is coming from when logging the diagnostics in `resolve`.
+    let viewProvidersRequiringFactory: Set<ClassDeclaration>|null = null;
+    let providersRequiringFactory: Set<ClassDeclaration>|null = null;
+    let wrappedViewProviders: Expression|null = null;
+
+    if (component.has('viewProviders')) {
+      const viewProviders = component.get('viewProviders') !;
+      viewProvidersRequiringFactory =
+          resolveProvidersRequiringFactory(viewProviders, this.evaluator);
+      wrappedViewProviders = new WrappedNodeExpr(
+          this.annotateForClosureCompiler ? wrapFunctionExpressionsInParens(viewProviders) :
+                                            viewProviders);
+    }
+
+    if (component.has('providers')) {
+      providersRequiringFactory =
+          resolveProvidersRequiringFactory(component.get('providers') !, this.evaluator);
+    }
 
     // Parse the template.
     // If a preanalyze phase was executed, the template may already exist in parsed form, so check
--- a/packages/compiler-cli/src/ngtsc/annotations/src/component.ts
+++ b/packages/compiler-cli/src/ngtsc/annotations/src/component.ts
@@ -17,6 +17,7 @@ import {DependencyTracker} from '../../incremental/api';
 import {IndexingContext} from '../../indexer';
 import {DirectiveMeta, MetadataReader, MetadataRegistry, extractDirectiveGuards} from '../../metadata';
 import {flattenInheritedDirectiveMetadata} from '../../metadata/src/inheritance';
+import {InjectableClassRegistry} from '../../metadata/src/registry';
 import {EnumValue, PartialEvaluator} from '../../partial_evaluator';
 import {ClassDeclaration, Decorator, ReflectionHost, reflectObjectLiteral} from '../../reflection';
 import {ComponentScopeReader, LocalModuleScopeRegistry} from '../../scope';
@@ -26,10 +27,11 @@ import {tsSourceMapBug29300Fixed} from '../../util/src/ts_source_map_bug_29300';
 import {SubsetOfKeys} from '../../util/src/typescript';
 
 import {ResourceLoader} from './api';
+import {getProviderDiagnostics} from './diagnostics';
 import {extractDirectiveMetadata, parseFieldArrayValue} from './directive';
 import {compileNgFactoryDefField} from './factory';
 import {generateSetClassMetadataCall} from './metadata';
-import {findAngularDecorator, isAngularCoreReference, isExpressionForwardReference, makeDuplicateDeclarationError, readBaseClass, unwrapExpression, wrapFunctionExpressionsInParens} from './util';
+import {findAngularDecorator, isAngularCoreReference, isExpressionForwardReference, makeDuplicateDeclarationError, readBaseClass, resolveProvidersRequiringFactory, unwrapExpression, wrapFunctionExpressionsInParens} from './util';
 
 const EMPTY_MAP = new Map<string, Expression>();
 const EMPTY_ARRAY: any[] = [];
@@ -53,6 +55,8 @@ export interface ComponentAnalysisData {
   guards: ReturnType<typeof extractDirectiveGuards>;
   template: ParsedTemplateWithSource;
   metadataStmt: Statement|null;
+  providersRequiringFactory: Set<ClassDeclaration>|null;
+  viewProvidersRequiringFactory: Set<ClassDeclaration>|null;
 }
 
 export type ComponentResolutionData = Pick<R3ComponentMetadata, ComponentMetadataResolvedFields>;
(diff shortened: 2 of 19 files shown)


packages/compiler-cli/src/ngtsc/annotations/src/component.ts
```diff
@@ -170,12 +174,18 @@ export class ComponentDecoratorHandler implements
       }
     }, undefined) !;
 
-    const viewProviders: Expression|null = component.has('viewProviders') ?
+    const rawViewProviders = component.get('viewProviders');
+    const wrappedViewProviders: Expression|null = rawViewProviders ?
```
petebacondarwin: In general @alxhub does not like us to rely upon the result of a `get()` call to determine whether an item is in a map, preferring to explicitly use `has()`.

packages/compiler-cli/src/ngtsc/annotations/src/component.ts
```diff
@@ -489,7 +504,28 @@ export class ComponentDecoratorHandler implements
         this.scopeRegistry.setComponentAsRequiringRemoteScoping(node);
       }
     }
-    return {};
+    if (analysis.providersRequiringFactory !== null &&
+        analysis.meta.providers instanceof WrappedNodeExpr) {
+      result.diagnostics = result.diagnostics || [];
+      const providerDiagnostics = getProviderDiagnostics(
+          analysis.providersRequiringFactory, analysis.meta.providers !.node, this.factoryRegistry);
+      if (providerDiagnostics) {
```
petebacondarwin: The result of `getProviderDiagnostics()` is always an array, so I expect this to always be true.
Similarly below.

packages/compiler-cli/src/ngtsc/annotations/src/directive.ts
```diff
@@ -98,16 +105,35 @@ export class DirectiveDecoratorHandler implements
...
+    let diagnostics: ts.Diagnostic[]|undefined = undefined;
+
+    if (analysis.providersRequiringFactory !== null &&
+        analysis.meta.providers instanceof WrappedNodeExpr) {
+      const providerDiagnostics = getProviderDiagnostics(
+          analysis.providersRequiringFactory, analysis.meta.providers !.node, this.factoryRegistry);
+      if (providerDiagnostics) {
+        diagnostics = providerDiagnostics;
+      }
+    }
```
petebacondarwin: This block is pretty much duplicated three times: once here and twice in the component `resolve()` method.
Could it be moved into a helper, which would also help to decrease the LOC in the already very long component `resolve()` method.

packages/compiler-cli/src/ngtsc/annotations/src/util.ts
```diff
@@ -375,4 +375,44 @@ const parensWrapperTransformerFactory: ts.TransformerFactory<ts.Expression> =
  */
 export function wrapFunctionExpressionsInParens(expression: ts.Expression): ts.Expression {
   return ts.transform(expression, [parensWrapperTransformerFactory]).transformed[0];
-}
\ No newline at end of file
+}
+
+/**
+ * Resolves the classes, passed into a Provider[], that are supposed to have a factory definition.
```
petebacondarwin:
```suggestion
 * Resolves the given `rawProviders` into `ClassDeclarations` and returns a set containing those that are known to require a factory definition.
```

packages/compiler-cli/src/ngtsc/annotations/src/util.ts
```diff
@@ -375,4 +375,44 @@ const parensWrapperTransformerFactory: ts.TransformerFactory<ts.Expression> =
...
+  const resolvedProviders = evaluator.evaluate(rawProviders);
+
+  if (Array.isArray(resolvedProviders)) {
+    const handleReference = (reference: Reference) => {
+      const node = reference.node;
+
+      // Skip declarations from external libraries since the consumer can't do much about them.
+      if (!node.getSourceFile().isDeclarationFile && isNamedClassDeclaration(node)) {
+        const constructor = node.members.find(ts.isConstructorDeclaration);
+        if (constructor !== undefined && constructor.parameters.length > 0) {
```
petebacondarwin: This is the key line in this function IMO.
A comment here pointing out that we are capturing this `node` because it has a contructor and that constructor has arguments, which will need injecting, would be helpful for future readers.

packages/compiler-cli/src/ngtsc/metadata/src/registry.ts
```diff
@@ -59,3 +59,14 @@ export class CompoundMetadataRegistry implements MetadataRegistry {
     }
   }
 }
+
+/**
+ * Keeps track of class declarations that have factories
```
petebacondarwin: This description (and the name of the registry) threw me for a while (and maybe I am still confused).
As I understand it, this is a registry of "injectable things" (e.g. services, directives, components, etc) that we know about, right? The fact that these injectables may require a factory to be able to instantiate them in DI is not actually relevant here.

packages/compiler-cli/test/ngtsc/ngtsc_spec.ts
```diff
@@ -5222,6 +5222,244 @@ export const Foo = Foo__PRE_R3__;
       });
     });
 
+    describe('undecorated providers', () => {
+      it('should error when an undecorated class is provided directly in a module', () => {
```
petebacondarwin: I feel that we need to be more clear here, since a class that has no constructor or a constructor with no parameters IS allowed.

```suggestion
      it('should error when an undecorated class, with a non-trivial constructor, is provided directly in a module', () => {
```

petebacondarwin (changes requested):
In the tests there is no mention of undecorated classes that have no constructor but inherit from a base class that does have a non-trivial constructor. Is this case covered by this PR?

There are a few minor changes suggested. Also the PR is conflicted and needs a rebase.

crisbeto:
@petebacondarwin the case when a class inherits a non-trivial constructor from an undecorated base class is handled by https://github.com/angular/angular/pull/34061.

packages/compiler-cli/src/ngtsc/metadata/src/registry.ts
crisbeto: Good point. I had a hard time figuring out how to name the class. I'll think about it a bit more.

packages/compiler-cli/src/ngtsc/annotations/src/directive.ts
crisbeto: I noticed it as I was writing it as well, but it was also fairly trivial code so pulling it out into a function probably won't help much. I wrote it this way, because I noticed other places where we return `undefined` instead of an empty array of diagnostics. If it's fine to return an empty array, we can simplify it a lot.

petebacondarwin: I think it would help in having less brain-load when reading the `resolve()` methods :-)

I think it would be great if we could avoid the `undefined` when dealing with diagnostics. There doesn't really appear to be any real difference between `undefined` and `[]` in the code. It is probably just that the `diagnostics` property on `ResolveResult` is optional.

crisbeto:
Rebased and addressed the feedback from @petebacondarwin.

packages/compiler-cli/src/ngtsc/annotations/src/component.ts
petebacondarwin: Did you mean to 👍 this comment and then not change the code ? 😺

packages/compiler-cli/src/ngtsc/annotations/src/directive.ts
petebacondarwin: Similarly did you intend to make changes to this?

packages/compiler-cli/src/ngtsc/annotations/src/util.ts
```diff
@@ -375,4 +375,48 @@ const parensWrapperTransformerFactory: ts.TransformerFactory<ts.Expression> =
...
+  if (Array.isArray(resolvedProviders)) {
+    const handleReference = (reference: Reference) => {
+      const node = reference.node;
+
+      // Skip declarations from external libraries since the consumer can't do much about them.
+      if (!node.getSourceFile().isDeclarationFile && isNamedClassDeclaration(node)) {
+        const constructor = node.members.find(ts.isConstructorDeclaration);
+
+        // Note that we only want to capture providers with a non-trivial constructor,
+        // because they're the ones that might be using DI and need to be decorated.
```
petebacondarwin: 👍

petebacondarwin (approved):

packages/compiler-cli/src/ngtsc/annotations/src/component.ts
crisbeto: I kept it here, because `rawViewProviders` was used in 4 other places within the same block. I've changed it everywhere else though.

packages/compiler-cli/ngcc/src/analysis/decoration_analyzer.ts
```diff
@@ -6,6 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 import {ConstantPool} from '@angular/compiler';
+import {InjectableClassRegistry} from '@angular/compiler-cli/src/ngtsc/metadata/src/registry';
```
alxhub: Relative import path.

packages/compiler-cli/src/ngtsc/annotations/src/component.ts
```diff
@@ -170,11 +175,17 @@ export class ComponentDecoratorHandler implements
...
+    const rawViewProviders = component.get('viewProviders');
+    const wrappedViewProviders: Expression|null = rawViewProviders ?
         new WrappedNodeExpr(
-            this.annotateForClosureCompiler ?
-                wrapFunctionExpressionsInParens(component.get('viewProviders') !) :
-                component.get('viewProviders') !) :
+            this.annotateForClosureCompiler ? wrapFunctionExpressionsInParens(rawViewProviders) :
+                                              rawViewProviders) :
+        null;
+    const viewProvidersRequiringFactory = rawViewProviders ?
```
alxhub: Use explicit comparisons: `rawViewProviders !== null`.

packages/compiler-cli/src/ngtsc/metadata/src/registry.ts
```diff
@@ -59,3 +59,13 @@ export class CompoundMetadataRegistry implements MetadataRegistry {
...
   }
 }
+
+/**
+ * Registry that keeps track of classes that can be injected (e.g. injectables, components, pipes).
+ */
+export class InjectableClassRegistry {
+  private classes = new Set<ClassDeclaration>();
+
+  registerClass(declaration: ClassDeclaration): void { this.classes.add(declaration); }
```
alxhub: Let's go with `registerInjectable` and `isInjectable` to make the code which consumes `InjectableClassRegistry` even more obvious about what it's doing.

packages/compiler-cli/src/ngtsc/annotations/src/util.ts
```diff
@@ -375,4 +375,48 @@ const parensWrapperTransformerFactory: ts.TransformerFactory<ts.Expression> =
...
+        if (constructor !== undefined && constructor.parameters.length > 0) {
+          providers.add(node);
+        }
+      }
+    };
+
+    resolvedProviders.forEach(function processProviders(provider) {
```
alxhub: Use a `for...of` loop instead.

I'm trying to switch the compiler over to these as they make debugging a lot simpler.

packages/compiler-cli/src/ngtsc/annotations/src/util.ts
```diff
@@ -375,4 +375,48 @@ const parensWrapperTransformerFactory: ts.TransformerFactory<ts.Expression> =
...
+      if (provider instanceof Reference) {
+        handleReference(provider);
+      } else if (provider instanceof Map) {
+        const useExisting = provider.get('useClass');
+        if (useExisting instanceof Reference) {
+          handleReference(useExisting);
+        }
+      } else if (Array.isArray(provider)) {
+        // If we ran into an array, recurse into it until we've resolve all the classes.
+        provider.forEach(processProviders);
```
alxhub: Oooh. This is clever.

packages/compiler-cli/src/ngtsc/annotations/src/util.ts
```diff
@@ -375,4 +375,48 @@ const parensWrapperTransformerFactory: ts.TransformerFactory<ts.Expression> =
...
+          providers.add(node);
+        }
+      }
+    };
+
+    resolvedProviders.forEach(function processProviders(provider) {
+      if (provider instanceof Reference) {
+        handleReference(provider);
+      } else if (provider instanceof Map) {
+        const useExisting = provider.get('useClass');
```
alxhub: Check `.has` first.

packages/compiler-cli/src/ngtsc/annotations/src/util.ts
```diff
@@ -375,4 +375,48 @@ const parensWrapperTransformerFactory: ts.TransformerFactory<ts.Expression> =
...
+    rawProviders: ts.Expression, evaluator: PartialEvaluator): Set<ClassDeclaration> {
+  const providers = new Set<ClassDeclaration>();
+  const resolvedProviders = evaluator.evaluate(rawProviders);
+
+  if (Array.isArray(resolvedProviders)) {
+    const handleReference = (reference: Reference) => {
+      const node = reference.node;
+
+      // Skip declarations from external libraries since the consumer can't do much about them.
+      if (!node.getSourceFile().isDeclarationFile && isNamedClassDeclaration(node)) {
```
alxhub: I think we should definitely handle the case someone does `providers: [Service]` and `Service` comes from an external library.

The `InjectableClassRegistry` can look if it has a `ɵfac` field present, and consider it an injectable if so.

alxhub (changes requested):

packages/compiler-cli/src/ngtsc/annotations/src/util.ts
crisbeto: Using a `for...of` here would require more code to do the recursion below.

packages/compiler-cli/src/ngtsc/annotations/src/util.ts
crisbeto: I think that removing this check will end up introducing breaking changes somewhere. I added it in the first place, because it was erroring for [this example](https://github.com/angular/components/blob/master/src/components-examples/cdk/scrolling/cdk-virtual-scroll-custom-strategy/cdk-virtual-scroll-custom-strategy-example.ts#L16) in the CDK which in turn inherits [from here](https://github.com/angular/components/blob/master/src/cdk/scrolling/fixed-size-virtual-scroll.ts#L41) that has a non-trivial constructor with `number` parameters. We can fix the CDK quickly since it's something we control, but I can imagine other libraries out there that have the same problem.

crisbeto:
Addressed/replied to all of the feedback.

packages/compiler-cli/src/ngtsc/annotations/src/component.ts
alxhub: Sorry, but @petebacondarwin is right. We should explicitly use `.has` even for `rawViewProviders`:

```typescript
const rawViewProviders = component.has('viewProviders') ? component.get('viewProviders') ! : null;
```

In general I think this code could benefit from fewer ternaries. Maybe structured like:

```typescript
let providersRequiringFactory: Set<ClassDeclaration>|null = null;

if (component.has('providers')) {
  // populate providersRequiringFactory if needed
}

let viewProviders: Expression[]|null = null;
if (component.has('viewProviders')) {
  // append view providers that require factories to providersRequiringFactory
}
```

alxhub (changes requested):

packages/compiler-cli/src/ngtsc/annotations/src/component.ts
crisbeto: Done. Also added a comment about why `providersRequiringFactory` and `viewProvidersRequiringFactory` are kept separately.

packages/compiler-cli/src/ngtsc/annotations/src/diagnostics.ts
```diff
@@ -0,0 +1,36 @@
...
+ * @param highlightNode Node that should be highlighted in the diagnostic.
+ *    Usually the place where the providers are being used.
+ * @param registry Registry that keeps track of the registered injectable classes.
+ */
+export function getProviderDiagnostics(
+    providerClasses: Set<ClassDeclaration>, highlightNode: ts.Node,
+    registry: InjectableClassRegistry): ts.Diagnostic[] {
+  const diagnostics: ts.Diagnostic[] = [];
+
+  providerClasses.forEach(provider => {
```
alxhub: Prefer a `for...of` loop (it's easier to debug through).

packages/compiler-cli/src/ngtsc/annotations/src/diagnostics.ts
```diff
@@ -0,0 +1,36 @@
...
+    if (!registry.isInjectable(provider)) {
+      diagnostics.push(makeDiagnostic(
```
alxhub: Let's do a trick here :)

1) Capture the `Reference` instead of the `ClassDeclaration` when tracking these providers.
2) Pass through the `ts.SourceFile` for the component/directive (let's call it `hostFile`).
3) Call `Reference.getIdentityIn(hostFile)`. If it returns a `ts.Identifier`, check whether the `ts.Identifier` is within the `ts.Expression` for the providers (which you call `highlightNode`). You can do this check positionally, by checking the `pos`/`end` of each.
4) If the result of `getIdentityIn` is contained in the `highlightNode`, then it's the identifier by which the class was referenced in the providers array, and we can put the diagnostic there. Otherwise we can fall back on the `providers` array itself.

packages/compiler-cli/src/ngtsc/annotations/src/util.ts
```diff
@@ -375,4 +375,46 @@ const parensWrapperTransformerFactory: ts.TransformerFactory<ts.Expression> =
...
+          providers.add(node);
+        }
+      }
+    };
+
+    resolvedProviders.forEach(function processProviders(provider) {
+      if (provider instanceof Reference) {
+        handleReference(provider);
+      } else if (provider instanceof Map && provider.has('useClass')) {
+        const useExisting = provider.get('useClass');
```
alxhub:
```suggestion
        const useExisting = provider.get('useClass') !;
```

packages/compiler-cli/src/ngtsc/annotations/src/util.ts
```diff
@@ -375,4 +375,46 @@ const parensWrapperTransformerFactory: ts.TransformerFactory<ts.Expression> =
...
+ */
+export function resolveProvidersRequiringFactory(
+    rawProviders: ts.Expression, evaluator: PartialEvaluator): Set<ClassDeclaration> {
+  const providers = new Set<ClassDeclaration>();
+  const resolvedProviders = evaluator.evaluate(rawProviders);
+
+  if (Array.isArray(resolvedProviders)) {
+    const handleReference = (reference: Reference) => {
+      const node = reference.node;
+      if (isNamedClassDeclaration(node)) {
```
alxhub: Maybe use `ReflectionHost.getConstructorParameters` instead of the TS APIs directly.

packages/compiler-cli/src/ngtsc/annotations/src/util.ts
```diff
@@ -375,4 +375,46 @@ const parensWrapperTransformerFactory: ts.TransformerFactory<ts.Expression> =
...
+        // because they're the ones that might be using DI and need to be decorated.
+        if (constructor !== undefined && constructor.parameters.length > 0) {
+          providers.add(node);
+        }
+      }
+    };
+
+    resolvedProviders.forEach(function processProviders(provider) {
+      if (provider instanceof Reference) {
+        handleReference(provider);
```
alxhub: I think it would be cleaner if instead of a callback, we first pick out the `reference` depending on the shape of the provider, and then process it inline.

```typescript
let tokenClass: Reference<ClassDeclaration>|null = null;

if (provider instanceof Reference) {
  tokenClass = provider;
} else if (provider instanceof Map && provider.has('useClass')) {
  const useClass = provider.get('useClass') !;
  if (useClass instanceof Reference) {
    tokenClass = useClass;
  }
}

if (tokenClass === null) {
  return;
}

// process the tokenClass here.
```

alxhub (approved):

packages/compiler-cli/src/ngtsc/annotations/src/diagnostics.ts
crisbeto: Cool! I don't think we need to pass in the `SourceFile`, because we can take it off of the `highlightNode` (I'll rename it to something more appropriate).

crisbeto:
Addressed the last set of feedback.

crisbeto:
Addressed the latest set of feedback.

packages/compiler-cli/src/ngtsc/annotations/src/diagnostics.ts
```diff
@@ -0,0 +1,50 @@
...
+ * @param providersDeclaration Node that declares the providers array.
+ * @param registry Registry that keeps track of the registered injectable classes.
+ */
+export function getProviderDiagnostics(
+    providerClasses: Set<Reference<ClassDeclaration>>, providersDeclaration: ts.Node,
+    registry: InjectableClassRegistry): ts.Diagnostic[] {
+  const diagnostics: ts.Diagnostic[] = [];
+
+  for (const provider of providerClasses) {
+    if (!registry.isInjectable(provider.node)) {
```
alxhub: Nit: rather than have a deeper `if`, we can invert the condition:

```typescript
for (const provider of providerClasses) {
  if (registry.isInjectable(provider.node)) {
    continue;
  }

  // Now logic here proceeds as normal, without the need for nesting.
}

File: packages/compiler-cli/src/ngtsc/annotations/src/util.ts [hunk 22]
```diff
@@ -375,4 +375,46 @@ const parensWrapperTransformerFactory: ts.TransformerFactory<ts.Expression> =
...
+ * a set containing those that are known to require a factory definition.
+ * @param rawProviders Expression that declared the providers array in the source.
+ */
+export function resolveProvidersRequiringFactory(
+    rawProviders: ts.Expression, reflector: ReflectionHost,
+    evaluator: PartialEvaluator): Set<Reference<ClassDeclaration>> {
+  const providers = new Set<Reference<ClassDeclaration>>();
+  const resolvedProviders = evaluator.evaluate(rawProviders);
+
+  if (Array.isArray(resolvedProviders)) {
```
alxhub: Same here - we could avoid nesting with an early return here, rather than nesting the entire function body in an `if` block.

alxhub (approved):

alxhub:
[Presubmit](https://test.corp.google.com/ui#id=OCL:285804135:BASE:285804235:1576522177734:ca39b808)
[Ivy Presubmit](https://test.corp.google.com/ui#id=OCL:285804135:BASE:285804235:1576522177703:6356269f)

packages/compiler-cli/src/ngtsc/annotations/src/util.ts
```diff
@@ -375,4 +375,46 @@ const parensWrapperTransformerFactory: ts.TransformerFactory<ts.Expression> =
...
+    resolvedProviders.forEach(function processProviders(provider) {
+      let tokenClass: Reference|null = null;
+
+      if (Array.isArray(provider)) {
+        // If we ran into an array, recurse into it until we've resolve all the classes.
+        provider.forEach(processProviders);
+      } else if (provider instanceof Reference) {
+        tokenClass = provider;
+      } else if (provider instanceof Map && provider.has('useClass')) {
```
alxhub: The g3 presubmit brought up an interesting case:

```typescript
{provide: Foo, useClass: Bar, deps: [DepA, DepB]}
```

is perfectly fine even if `Bar` is not injectable, as the user is explicitly telling the injector how to call the constructor. So `useClass` with `deps` can be ignored here:

```suggestion
      } else if (provider instanceof Map && provider.has('useClass') && !provider.has('deps')) {
```

Please add a test to that effect :)

alxhub (changes requested):

crisbeto:
Fixed the presubmit issue, addressed the feedback and rebased.

alxhub:
Updated:

[Presubmit](http://test/OCL:285804135:BASE:286080470:1576627432631:4ebb8553)
[Ivy Presubmit](http://test/OCL:285804135:BASE:286080470:1576627432631:3cbb15ef)

alxhub:
Closing in favor of #34460.

PR #34334, opened 2019-12-10
perf(ivy): eagerly parse the template twice during analysis

alxhub:
A quirk of the Angular template parser is that when parsing templates in the
"default" mode, with options specified by the user, the source mapping
information in the template AST may be inaccurate. As a result, the compiler
parses the template twice: once for "emit" and once to produce an AST with
accurate sourcemaps for diagnostic production.
    
Previously, only the first parse was performed during analysis. The second
parse occurred during the template type-checking phase, just in time to
produce the template type-checking file.
    
However, with the reuse of analysis results during incremental builds, it
makes more sense to do the diagnostic parse eagerly during analysis so that
the work isn't unnecessarily repeated in subsequent builds. This commit
refactors the `ComponentDecoratorHandler` to do both parses eagerly, which
actually cleans up some complexity around template parsing as well.

### DIFF
--- a/packages/compiler-cli/ngcc/src/analysis/decoration_analyzer.ts
+++ b/packages/compiler-cli/ngcc/src/analysis/decoration_analyzer.ts
@@ -92,8 +92,8 @@ export class DecorationAnalyzer {
         /* i18nUseExternalIds */ true, this.bundle.enableI18nLegacyMessageIdFormat,
         this.moduleResolver, this.cycleAnalyzer, this.refEmitter, NOOP_DEFAULT_IMPORT_RECORDER,
         NOOP_DEPENDENCY_TRACKER, /* annotateForClosureCompiler */ false),
+    // See the note in ngtsc about why this cast is needed.
     // clang-format off
-        // See the note in ngtsc about why this cast is needed.
     new DirectiveDecoratorHandler(
         this.reflectionHost, this.evaluator, this.fullRegistry, NOOP_DEFAULT_IMPORT_RECORDER,
         this.isCore, /* annotateForClosureCompiler */ false) as DecoratorHandler<unknown, unknown, unknown>,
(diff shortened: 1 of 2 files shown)


packages/compiler-cli/src/ngtsc/annotations/src/component.ts
```diff
@@ -737,14 +696,41 @@ export class ComponentDecoratorHandler implements
       interpolation = InterpolationConfig.fromArray(value as[string, string]);
     }
 
+    const {errors, nodes: emitNodes, styleUrls, styles} = parseTemplate(templateStr, templateUrl, {
```
JoostK: It would make more sense to extract `errors` from the secondary parse.

packages/compiler-cli/src/ngtsc/annotations/src/component.ts
```diff
@@ -737,14 +696,41 @@ export class ComponentDecoratorHandler implements
...
+    //    newlines). This also destroys source mapping information.
+    //
+    // In order to guarantee the correctness of diagnostics, templates are parsed a second time with
+    // the above options set to preserve source mappings.
+
+    const {nodes: diagNodes} = parseTemplate(templateStr, templateUrl, {
+      preserveWhitespaces: true,
+      interpolationConfig: interpolation,
+      range: templateRange, escapedString,
+      enableI18nLegacyMessageIdFormat: this.enableI18nLegacyMessageIdFormat,
```
JoostK: I think we should be able to set this to `false`, always. It avoids computing message digests that we're not going to use, anyway.

JoostK (approved):
Much better!

packages/compiler-cli/ngcc/src/analysis/decoration_analyzer.ts
```diff
@@ -90,9 +91,9 @@ export class DecorationAnalyzer {
         /* defaultPreserveWhitespaces */ false,
         /* i18nUseExternalIds */ true, this.bundle.enableI18nLegacyMessageIdFormat,
         this.moduleResolver, this.cycleAnalyzer, this.refEmitter, NOOP_DEFAULT_IMPORT_RECORDER,
-        /* annotateForClosureCompiler */ false),
+        NOOP_DEPENDENCY_TRACKER, /* annotateForClosureCompiler */ false),
     // clang-format off
-    // See the note in ngtsc about why this cast is needed.
+        // See the note in ngtsc about why this cast is needed.
```
petebacondarwin: incorrect indentation?

alxhub:
[Presubmit](https://test.corp.google.com/ui#id=OCL:285252307:BASE:285252356:1576184498255:7d384d45)
[Ivy Presubmit](https://test.corp.google.com/ui#id=OCL:285252307:BASE:285252356:1576184498201:56fb41af)

PR #34335, opened 2019-12-10
refactor(ivy): avoid type coercion in `saveNameToExportMap` function

AndrewKushnir:
TypeScript 3.7 flags `if` conditions that implicitly coerce a function/method definition. While checking for the `template` presence on a def (actually checking whether we work with Component) in `saveNameToExportMap`, the `if` condition had implicit type coercion. This commit updates the condition to use the `isComponentDef` function (that checks `def.template` against `null` internally) to avoid compilation errors with TypeScript 3.7.

Note: similar change was originally proposed by @mprobst in g3. This PR applies the changes to main Angular repo, so we can sync them into g3.

## PR Type

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/render3/instructions/shared.ts
+++ b/packages/core/src/render3/instructions/shared.ts
@@ -1300,7 +1300,7 @@ function saveNameToExportMap(
         exportsMap[def.exportAs[i]] = index;
       }
     }
-    if ((def as ComponentDef<any>).template) exportsMap[''] = index;
+    if (isComponentDef(def)) exportsMap[''] = index;
   }
 }
 


packages/core/src/render3/instructions/shared.ts
```diff
@@ -1300,7 +1300,7 @@ function saveNameToExportMap(
         exportsMap[def.exportAs[i]] = index;
       }
     }
-    if ((def as ComponentDef<any>).template) exportsMap[''] = index;
+    if ((def as ComponentDef<any>).template !== null) exportsMap[''] = index;
```
pkozlowski-opensource: we've got the `isComponentDef` utility that does this check, maybe it would be good to re-use it?

AndrewKushnir:
[Presubmit](http://test/OCL:284821574:BASE:284821640:1576008746669:141970f8)

packages/core/src/render3/instructions/shared.ts
AndrewKushnir: Thanks @pkozlowski-opensource, that makes sense. I've updated the code to use `isComponentDef` instead.

AndrewKushnir:
[Presubmit](http://test/OCL:284823930:BASE:284823960:1576009469983:136de606)

pkozlowski-opensource (approved):
LGTM

PR #34337, opened 2019-12-10
build: prevent creating runfiles tree during remote builds

josephperrott:
Runfiles trees are not needed when building remotely as the trees
are actually created from in-memory manifests


gkalpak (approved):

gkalpak:
Just to be clear: LGTM as soon as CI is happy :grin:

Issue #33674, opened 2019-11-07
Add diagnostics error for template var assignment issue


kara:
If you try to assign a value to a template var, we should give you an error with a squiggly that points you to where the invalid assignment is.

kara:
https://github.com/angular/angular/pull/34339

PR #34339, opened 2019-12-11
fix(ivy): generate a better error for template var writes

alxhub:
In Ivy it's illegal for a template to write to a template variable. So the
template:

```html
<ng-template let-somevar>
  <button (click)="somevar = 3">Set var to 3</button>
</ng-template>
```

is erroneous and previously would fail to compile with an assertion error
from the `TemplateDefinitionBuilder`. This error wasn't particularly user-
friendly, though, as it lacked the context of which template or where the
error occurred.

In this commit, a new check in template type-checking is added which detects
such erroneous writes and produces a true diagnostic with the appropriate
context information.

### DIFF
--- a/packages/compiler-cli/src/ngtsc/typecheck/src/type_check_block.ts
+++ b/packages/compiler-cli/src/ngtsc/typecheck/src/type_check_block.ts
@@ -6,7 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {AST, BindingPipe, BindingType, BoundTarget, DYNAMIC_TYPE, ImplicitReceiver, MethodCall, ParseSourceSpan, ParseSpan, ParsedEventType, PropertyRead, SchemaMetadata, TmplAstBoundAttribute, TmplAstBoundEvent, TmplAstBoundText, TmplAstElement, TmplAstNode, TmplAstReference, TmplAstTemplate, TmplAstTextAttribute, TmplAstVariable} from '@angular/compiler';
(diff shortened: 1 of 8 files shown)


packages/compiler-cli/src/ngtsc/typecheck/src/type_check_block.ts
```diff
@@ -972,6 +981,16 @@ class TcbExpressionTranslator {
       // returned here to let it fall through resolution so it will be caught when the
       // `ImplicitReceiver` is resolved in the branch below.
       return this.resolveTarget(ast);
+    } else if (ast instanceof PropertyWrite && ast.receiver instanceof ImplicitReceiver) {
```
JoostK: Consider adding a test in `type_check_block_spec.ts` for this case.

JoostK (approved):
Looks good, with one comment

kara (approved):
LGTM, but can you add `Closes #33674` to the bottom of the commit?

packages/compiler-cli/src/ngtsc/typecheck/src/type_check_block.ts
alxhub: Done.

alxhub:
[Presubmit](https://test.corp.google.com/ui#id=OCL:285221233:BASE:285221288:1576175884792:b7ce2ead)
[Ivy Presubmit](https://test.corp.google.com/ui#id=OCL:285221233:BASE:285221288:1576175884739:ffaa1999)

PR #34348, opened 2019-12-11
docs: more precise docs for template statement syntax

thorn0:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/aio/content/guide/template-syntax.md
+++ b/aio/content/guide/template-syntax.md
@@ -237,16 +237,15 @@ You're free to change anything, anywhere, during this turn of the event loop.
 
 Like template expressions, template *statements* use a language that looks like JavaScript.
 The template statement parser differs from the template expression parser and
-specifically supports both basic assignment (`=`) and chaining expressions
-(with <code>;</code> or <code>,</code>).
+specifically supports both basic assignment (`=`) and chaining expressions with <code>;</code>.
 
-However, certain JavaScript syntax is not allowed:
+However, certain JavaScript and template expression syntax is not allowed:
 
 * <code>new</code>
 * increment and decrement operators, `++` and `--`
 * operator assignment, such as `+=` and `-=`
-* the bitwise operators `|` and `&`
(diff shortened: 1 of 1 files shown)


JoostK:
I think there's a bit more to it than you currently suggest, although the proposed change is already getting us in the right direction.

Pipes are simple and are indeed not allowed.

The non-null assertion operator, _can_ be used in template statements (both in a _left-hand side_ (LHS) expression and in a _right-hand side_ (RHS) expression) with the exception that LHS expressions cannot end with it, whereas RHS expressions can. As an example, `selectedName = person.name!` is fine, but `selectedName! = person.name;` is not. The latter however would also not be valid TypeScript, so that is mainly a matter of grammar, not so much restrictions.

For the safe-navigation operator, it depends on where it is used. It is perfectly fine to use it in a RHS expression, which is basically something that evaluates to a value. It cannot be used, however, in LHS expressions, which are the ones that you can assign something into. For example, the statement `toggle.enabled = !toggle.enabled` has a LHS = `toggle.enabled` and a RHS of `!toggle.enabled`. The `?.` operator can be used in the RHS as that would just result in an `undefined` value, whereas it's illegal on the left as you can't assign into something that is not available.

So if anything, we should be mentioning the safe-navigation operator here, not the non-null assertion operation. @thorn0 If you think I'm misinterpreting something here please let me know.

thorn0:
@JoostK Yeah, I included the non-null assertion operator there by mistake. Fixed.

As for `?.` in assignments, it's the same rules as in TS, so I don't think this needs additional explaining.

JoostK:
@thorn0 Agreed, looks good now! Could you squash the two commits into one?

thorn0:
@JoostK done

mhevery (approved):

PR #34112, opened 2019-11-27
build: update integration/bazel & @angular/bazel schematics to rules nodejs 0.42.2

gregmagolan:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [x] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] angular.io application / infrastructure changes
- [ ] Other... Please describe:

## What is the current behavior?

Issue Number: N/A

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

PR #34354, opened 2019-12-11
build: update to rules_nodejs@0.42.3

josephperrott:
Updating to latest rules_nodejs which properly handles stamping for
releases.  Previously rollup_bundle and npm_package were unable to
be cached remotely.
See https://github.com/bazelbuild/rules_nodejs/pull/1441 for changes.

With new requirements from changes in rules_nodejs, we now must add
--stamp to bazel builds we wish to include stamping.


kyliau (approved):

devversion (approved):
LGTM. Nit: It would be nice if the changes for `@angular/bazel` are in a separate commit with the `bazel` scope (so that they show up in the changelog).

devversion:
Looks like the PR is blocked on https://github.com/angular/angular/pull/34112 getting merged.

integration/bazel/src/BUILD.bazel
```diff
@@ -25,6 +52,62 @@ ng_module(
...
+terser_minified(
+    name = "bundle.min",
+    src = ":bundle",
+)
+
+html_insert_assets(
+    name = "inject_scripts",
+    outs = ["_/index.html"],
+    args = [
+        "--html",
```
jbedard: FYI if you use the latest `html-insert-assets` you can save some lines doing `--html=$(location :index.html)` similar to the [rules_nodejs examples](https://github.com/bazelbuild/rules_nodejs/blob/0.42.3/examples/angular/src/BUILD.bazel#L175-L189).

PR #34356, opened 2019-12-11
fix(ngcc): handle imports in dts files when processing UMD

petebacondarwin:
When statically evalulating UMD code it is possible to find
that we are looking for the declaration of an identifier that
actually came from a typings file (rather than a UMD file).

Previously, the UMD reflection host would always try to use
a UMD specific algorithm for finding identifier declarations,
but when the id is actually in a typings file this resulted in the
returned declaration being the containing file of the declaration
rather than the declaration itself.

Now the UMD reflection host will check to see if the file containing
the identifier is a typings file and use the appropriate stategy.

### DIFF
--- a/packages/compiler-cli/ngcc/test/host/umd_host_spec.ts
+++ b/packages/compiler-cli/ngcc/test/host/umd_host_spec.ts
@@ -1777,6 +1777,43 @@ runInEachFileSystem(() => {
           expect(actualDeclaration !.node).toBe(expectedDeclarationNode);
           expect(actualDeclaration !.viaModule).toBe('@angular/core');
         });
+
+        it('should return the correct declaration of an identifier imported in a typings file',
+           () => {
+
+             const FILES = [
+               {
+                 name: _('/node_modules/test-package/index.d.ts'),
+                 contents: `
+              import {SubModule} from 'sub_module';
+              export const x = SubModule;
+              `,
+               },
+               {
+                 name: _('/node_modules/packages.json'),
+                 contents: '{ "typings: "index.d.ts" }',
+               },
+               {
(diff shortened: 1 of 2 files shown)


packages/compiler-cli/ngcc/test/host/umd_host_spec.ts
```diff
@@ -1777,6 +1777,43 @@ runInEachFileSystem(() => {
...
+                 contents: '{ "typings: "index.d.ts" }',
+               },
+               {
+                 name: _('/node_modules/sub_module/index.d.ts'),
+                 contents: `export class SubModule {}`,
+               }
+             ];
+             loadTestFiles(FILES);
+             const {program, host: compilerHost} = makeTestBundleProgram(FILES[0].name);
+             const host = new UmdReflectionHost(new MockLogger(), false, program, compilerHost);
```
gkalpak: BTW, this will likely not generate a conflict with #34254, but if both are merged as is will start failing tests.

gkalpak (approved):
👌

packages/compiler-cli/ngcc/test/host/umd_host_spec.ts
petebacondarwin: Actually it will create a conflict.
So we are all good there.
:-)

petebacondarwin: No you are correct - it is not conflicted and it will break.

petebacondarwin: I am going to rebase that one on top of this, since this is ready to be merged.

PR #34363, opened 2019-12-12
feat(language-service): completions support for template reference variables

ivanwonder:
https://angular.io/guide/template-syntax#how-a-reference-variable-gets-its-value
get completions from option `exportAs` of `Directive`.

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/packages/language-service/src/completions.ts
+++ b/packages/language-service/src/completions.ts
@@ -6,7 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {AST, AstPath, AttrAst, Attribute, BoundDirectivePropertyAst, BoundElementPropertyAst, BoundEventAst, BoundTextAst, Element, ElementAst, ImplicitReceiver, NAMED_ENTITIES, Node as HtmlAst, NullTemplateVisitor, ParseSpan, PropertyRead, TagContentType, TemplateBinding, Text, getHtmlTagDefinition} from '@angular/compiler';
+import {AST, AstPath, AttrAst, Attribute, BoundDirectivePropertyAst, BoundElementPropertyAst, BoundEventAst, BoundTextAst, Element, ElementAst, ImplicitReceiver, NAMED_ENTITIES, Node as HtmlAst, NullTemplateVisitor, ParseSpan, PropertyRead, ReferenceAst, TagContentType, TemplateBinding, Text, getHtmlTagDefinition} from '@angular/compiler';
 import {$$, $_, isAsciiLetter, isDigit} from '@angular/compiler/src/chars';
 
 import {AstResult} from './common';
@@ -170,11 +170,18 @@ export function getTemplateCompletions(
             }
           },
           visitAttribute(ast) {
-            if (!ast.valueSpan || !inSpan(templatePosition, spanOf(ast.valueSpan))) {
+            const bindParts = ast.name.match(BIND_NAME_REGEXP);
+            const isReference = bindParts && bindParts[ATTR.KW_REF_IDX] !== undefined;
+            if (!isReference &&
+                (!ast.valueSpan || !inSpan(templatePosition, spanOf(ast.valueSpan)))) {
               // We are in the name of an attribute. Show attribute completions.
               result = attributeCompletions(templateInfo, path);
             } else if (ast.valueSpan && inSpan(templatePosition, spanOf(ast.valueSpan))) {
-              result = attributeValueCompletions(templateInfo, templatePosition, ast);
+              if (isReference) {
+                result = referenceAttributeValueCompletions(templateInfo, templatePosition, ast);
+              } else {
+                result = attributeValueCompletions(templateInfo, templatePosition, ast);
+              }
             }
           },
           visitText(ast) {
@@ -313,6 +320,26 @@ function attributeValueCompletions(
   return visitor.results;
 }
 
+function referenceAttributeValueCompletions(
+    info: AstResult, position: number, attr: Attribute): ng.CompletionEntry[] {
+  const path = findTemplateAstAt(info.templateAst, position);
+  if (!path.tail) {
+    return [];
+  }
+
+  // When the template parser does not find a directive with matching "exportAs",
+  // the ReferenceAst will be ignored.
+  if (!(path.tail instanceof ReferenceAst)) {
+    // The sourceSpan of an ReferenceAst is the valueSpan of the HTML Attribute.
+    path.push(new ReferenceAst(attr.name, null !, attr.value, attr.valueSpan !));
+  }
+  const dinfo = diagnosticInfoFromTemplateInfo(info);
+  const visitor =
+      new ExpressionVisitor(info, position, () => getExpressionScope(dinfo, path, false));
+  path.tail.visit(visitor, path.parentOf(path.tail));
+  return visitor.results;
+}
+
 function elementCompletions(info: AstResult): ng.CompletionEntry[] {
   const results: ng.CompletionEntry[] = [...ANGULAR_ELEMENTS];
 
@@ -444,6 +471,16 @@ class ExpressionVisitor extends NullTemplateVisitor {
       const expressionAst = new PropertyRead(span, span.toAbsolute(offset), receiver, '');
       this.addAttributeValuesToCompletions(expressionAst);
     }
+  }
+
+  visitReference(ast: ReferenceAst, context: ElementAst) {
+    context.directives.forEach(dir => {
+      const {exportAs} = dir.directive;
+      if (exportAs) {
+        this.completions.set(
+            exportAs, {name: exportAs, kind: ng.CompletionKind.REFERENCE, sortText: exportAs});
+      }
+    });
   }
 
   visitBoundText(ast: BoundTextAst) {
--- a/packages/language-service/test/completions_spec.ts
+++ b/packages/language-service/test/completions_spec.ts
@@ -681,6 +681,22 @@ describe('completions', () => {
         expectContain(completions, CompletionKind.METHOD, ['substring']);
       });
     });
+
+    describe('with template reference variables', () => {
+      it('should be able to get the completions (ref- prefix)', () => {
+        mockHost.override(TEST_TEMPLATE, `<form ref-itemForm="ngF~{reference}"></form>`);
+        const marker = mockHost.getLocationMarkerFor(TEST_TEMPLATE, 'reference');
+        const completions = ngLS.getCompletionsAt(TEST_TEMPLATE, marker.start) !;
+        expectContain(completions, CompletionKind.REFERENCE, ['ngForm']);
+      });
+
+      it('should be able to get the completions (# prefix)', () => {
+        mockHost.override(TEST_TEMPLATE, `<form #itemForm="ngF~{reference}"></form>`);
+        const marker = mockHost.getLocationMarkerFor(TEST_TEMPLATE, 'reference');
+        const completions = ngLS.getCompletionsAt(TEST_TEMPLATE, marker.start) !;
+        expectContain(completions, CompletionKind.REFERENCE, ['ngForm']);
+      });
+    });
   });
 });
 


packages/language-service/src/completions.ts
```diff
@@ -242,6 +242,25 @@ function attributeValueCompletions(
   if (!path.tail) {
     return [];
   }
+
+  const isTemplateRef = attr.name.startsWith('#') || attr.name.startsWith('ref-');
```
ayazhafiz: Please add a comment linking to reference variable documentation (https://angular.io/guide/template-syntax#template-reference-variables-var)

packages/language-service/test/completions_spec.ts
```diff
@@ -117,6 +117,26 @@ describe('completions', () => {
     expectContain(completions, CompletionKind.PROPERTY, ['id', 'name']);
   });
 
+  it('should be able to get the completions for template reference variables', () => {
+    const fileName = mockHost.addCode(`
```
ayazhafiz: please use `mockHost.override`. It generally speeds up the tests.

packages/language-service/test/completions_spec.ts
```diff
@@ -117,6 +117,26 @@ describe('completions', () => {
...
+      @Component({
+        selector: 'foo-component',
+        template: \`
+        <form #itemForm="ngF~{reference}" (ngSubmit)="onSubmit(itemForm)">
+          <label for="name"
+            >Name <input class="form-control" name="name" ngModel required />
```
ayazhafiz: is this newline intentional?

packages/language-service/src/completions.ts
```diff
@@ -242,6 +242,25 @@ function attributeValueCompletions(
...
+    let parent: TemplateAst|undefined = path.tail;
+    const result: ng.CompletionEntry[] = [];
+    if (path.tail instanceof ReferenceAst) {
+      parent = path.parentOf(path.tail);
+    }
+    if (parent && parent instanceof ElementAst) {
+      parent.directives.forEach(dir => {
+        const exportAs = dir.directive.exportAs;
+        if (exportAs) {
+          result.push({name: exportAs, kind: ng.CompletionKind.REFERENCE, sortText: exportAs});
```
ayazhafiz: nit: consider renaming `exportAs` to a more semantic name in this context, maybe `directiveName`

ayazhafiz (dismissed):

packages/language-service/src/completions.ts
kyliau: Could you please add a test for the `ref-` case?

kyliau (commented):
Thank you for contributing! Your help is much appreciated :)

packages/language-service/src/completions.ts
ivanwonder: I think it's better to use the name defined by angular https://angular.io/api/core/Directive#exportas

packages/language-service/src/completions.ts
```diff
@@ -242,6 +242,26 @@ function attributeValueCompletions(
   if (!path.tail) {
     return [];
   }
+
+  // https://angular.io/guide/template-syntax#template-reference-variables-var
+  const isTemplateRef = attr.name.startsWith('#') || attr.name.startsWith('ref-');
+  if (isTemplateRef) {
+    let parent: TemplateAst|undefined = path.tail;
+    const result: ng.CompletionEntry[] = [];
+    if (path.tail instanceof ReferenceAst) {
```
kyliau: Sorry for the delay, it took me a while to get back to this PR.
This is not the right place to do this check, because the `ExpressionVisitor` is designed specifically to traverse the TemplateAst. In this case, this visitor already knows how to go to a reference via `visitReference()`.
`ExpressionVisitor` extends `NullTemplateVisitor` and the reference node is currently not visited. This method needs to be added.
Please let me know if you need more info / clarification. I'm happy to guide you through.
Thank you again!

kyliau (changes requested):

packages/language-service/src/completions.ts
ivanwonder: I try the method with visitReference(). There is a problem.
`<form #itemForm="ngF~{reference}"></form>`
When the angular parses the HTML, it will report an error `Template parse errors:
There is no directive with "exportAs" set to "ngF"`. So the `visitReference()` method will not invoked. This is why I add the code on line 251

ivanwonder: So the `path.tail` is assigned to element `form`, and will not be regarded as a template reference.

kyliau:
@ivanwonder, you are correct. The TemplateAst parser does not parse attributes correctly. This is bad, because it means there's no 1-1 mapping from HtmlAst to TemplateAst.
More specifically, the HtmlAst would contain `Element --> Attribute`, but when it gets parsed into TemplateAst, there's only `ElementAst`. This is why the visit has no effect.
I'll fix this in a separate PR, and you should be able to implement this feature the correct way after that.
Thank you for your patience!

ivanwonder:
Thank you for your guidance. I will pick it up after you fix the problem.

ivanwonder:
@kyliau I think I can help to resolve the problem of `TemplateAst` if you have not begun to do it or no time to look into it.
I recently read the code, and my solution is to add the reference to the array `targetReferences` before reporting the error on line 596.
https://github.com/angular/angular/blob/2dffe65cfdc1329819a5ef0d0f2069efb0b07ff8/packages/compiler/src/template_parser/template_parser.ts#L593-L608

ayazhafiz:
> @kyliau I think I can help to resolve the problem of `TemplateAst` if you have not begun to do it or no time to look into it.

I think this was partially fixed in #34459, @kyliau can you confirm? I think you would have to do a little extra work to get `visitReference` to be called, since that change parses an attribute and not a reference, but you could add a call to `visitReference` in `ExpressionVisitor#visitAttr` or parse the reference where the attribute is.

> I recently read the code, and my solution is to add the reference to the array `targetReferences` before reporting the error on line 596.

I think this would work, but we will have to get a review from the compiler team to make sure it is okay. My only worry is that this would break the typings of the ReferenceAst because the value of the created reference AST would be null (since there is no matching directive to reference), so this might cause some trouble if the reference is used later in the template (but if an error is reported, I don't know if this matters).

kyliau:
yes, @ayazhafiz is right. I did the work in #34459 to partially address the problem of incomplete template AST. The ideal solution here is to visit AttrAst, create ReferenceAst, then visit it.
However, in order to get hold of the directives that apply to an Element, you have to pass the ElementAst to visitReference() as well. This can be done through the context (second parameter).
Bear in mind that it is insufficient to provide "exportAs" as the sole result set, as was done here. Other symbols that are in scope are valid candidates too.

ivanwonder:
@kyliau What are the other symbols,  can you give me an example? When I input others, it will report the error no directive with "exportAs"

ivanwonder:
I think it's better to add a new `referenceAttributeValueCompletions` here.
https://github.com/angular/angular/blob/9b9116c79d626b6356f3dfe0d48c9d990ac412a2/packages/language-service/src/completions.ts#L177
Just push the missing `ReferenceAst` to `path` if need, begin to visit it with `ElementAst` parameter in the new method. Like 
https://github.com/angular/angular/blob/9b9116c79d626b6356f3dfe0d48c9d990ac412a2/packages/language-service/src/completions.ts#L362
It's a little kind of weird to invoke the method `visitReference` in `visitAttr`, travel the `path` to get the right `ElementAst` and pass it to `visitReference`.

ivanwonder:
> yes, @ayazhafiz is right. I did the work in #34459 to partially address the problem of incomplete template AST. The ideal solution here is to visit AttrAst, create ReferenceAst, then visit it.

This is the new implementation, if need, I can change to the way as @kyliau  said above.

packages/language-service/src/completions.ts
```diff
@@ -174,7 +174,12 @@ export function getTemplateCompletions(
               // We are in the name of an attribute. Show attribute completions.
               result = attributeCompletions(templateInfo, path);
             } else if (ast.valueSpan && inSpan(templatePosition, spanOf(ast.valueSpan))) {
-              result = attributeValueCompletions(templateInfo, templatePosition, ast);
+              const bindParts = ast.name.match(BIND_NAME_REGEXP);
+              if (bindParts && bindParts[ATTR.KW_REF_IDX] !== undefined) {
+                result = referenceAttributeValueCompletions(templateInfo, templatePosition, ast);
+              } else {
+                result = attributeValueCompletions(templateInfo, templatePosition, ast);
+              }
```
ayazhafiz: I would prefer doing this one nesting up, because if we're in a reference name, we probably don't want to show completions since a new reference is being defined.

```typescript
const bindParts = ast.name.match(BIND_NAME_REGEXP);
const isReference = bindParts && bindParts[ATTR.KW_REF_IDX] !== undefined;
if (!isReference && (!ast.valueSpan || !inSpan(templatePosition, spanOf(ast.valueSpan)))) {
  result = attributeCompletions(templateInfo, path);
} else if (ast.valueSpan && inSpan(templatePosition, spanOf(ast.valueSpan))) {
  if (isReference) {
    result = referenceAttributeValueCompletions(templateInfo, templatePosition, ast);
  } else {
    result = attributeValueCompletions(templateInfo, templatePosition, ast);
  }
}
```

packages/language-service/src/completions.ts
```diff
@@ -313,6 +318,26 @@ function attributeValueCompletions(
...
 }
 
+function referenceAttributeValueCompletions(
+    info: AstResult, position: number, attr: Attribute): ng.CompletionEntry[] {
+  const path = findTemplateAstAt(info.templateAst, position);
+  if (!path.tail) {
+    return [];
+  }
+
+  // When the template parser donot find the directive with "exportAs",
```
```suggestion
  // When the template parser does not find a directive with matching "exportAs",
```

packages/language-service/src/completions.ts
```diff
@@ -313,6 +318,26 @@ function attributeValueCompletions(
...
+  // the ReferenceAst will be ignore.
+  if (!(path.tail instanceof ReferenceAst)) {
+    // The sourceSpan of an AttrAst is the valueSpan of the HTML Attribute.
```
```suggestion
    // The sourceSpan of a ReferenceAst is the valueSpan of the HTML Attribute.
```

packages/language-service/src/completions.ts
```diff
@@ -313,6 +318,26 @@ function attributeValueCompletions(
...
+    path.push(new ReferenceAst(attr.name, null !, attr.value, attr.valueSpan !));
```
ayazhafiz: Maybe we should create a dummy value for the ReferenceAst's value, but we can do that in another PR.

packages/language-service/src/completions.ts
```diff
@@ -446,6 +471,16 @@ class ExpressionVisitor extends NullTemplateVisitor {
     }
   }
 
+  visitReference(ast: ReferenceAst, context: ElementAst) {
+    context.directives.forEach(dir => {
+      const exportAs = dir.directive.exportAs;
```
ayazhafiz: nit
```suggestion
      const {exportAs} = dir.directive;
```

ayazhafiz (changes requested):

packages/language-service/src/completions.ts
ivanwonder: Agree, the `ReferenceAst ` with second parameter `null` will refer to the DOM element on which it is declared.
https://github.com/angular/angular/blob/2dffe65cfdc1329819a5ef0d0f2069efb0b07ff8/packages/compiler/test/template_parser/template_parser_spec.ts#L1318-L1321

packages/language-service/src/completions.ts
```diff
@@ -313,6 +320,26 @@ function attributeValueCompletions(
...
 
+function referenceAttributeValueCompletions(
+    info: AstResult, position: number, attr: Attribute): ng.CompletionEntry[] {
+  const path = findTemplateAstAt(info.templateAst, position);
+  if (!path.tail) {
+    return [];
+  }
+
+  // When the template parser does not find a directive with matching "exportAs",
+  // the ReferenceAst will be ignore.
```
```suggestion
  // the ReferenceAst will be ignored.
```

ayazhafiz (approved):
Thanks for the feature 🙂

kyliau (approved):

PR #34381, opened 2019-12-12
fix(ivy): improve ExpressionChangedAfterChecked error

AndrewKushnir:
Prior to this change, the ExpressionChangedAfterChecked error thrown in Ivy was missing useful information that was available in View Engine, specifically: missing property name for property bindings and also the content of the entire property interpolation (only a changed value was displayed) if one of expressions was changed unexpectedly. This commit improves the error message by including the mentioned information into the error text.

## PR Type

- [x] Bugfix

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/render3/errors.ts
+++ b/packages/core/src/render3/errors.ts
@@ -51,9 +51,10 @@ export function throwErrorIfNoChangesMode(
   if (creationMode) {
     msg +=
         ` It seems like the view has been created after its parent and its children have been dirty checked.` +
-        ` Has it been created in a change detection hook ?`;
+        ` Has it been created in a change detection hook?`;
   }
-  // TODO: include debug context
+  // TODO: include debug context, see `viewDebugError` function in
+  // `packages/core/src/view/errors.ts` for reference.
   throw new Error(msg);
 }
 
--- a/packages/core/src/render3/errors.ts
+++ b/packages/core/src/render3/errors.ts
@@ -11,7 +11,6 @@ import {stringify} from '../util/stringify';
 
 import {TNode} from './interfaces/node';
 import {LView, TVIEW} from './interfaces/view';
-import {getBindingRoot} from './state';
 import {INTERPOLATION_DELIMITER} from './util/misc_utils';
 
 
@@ -46,9 +45,9 @@ export function throwInvalidProviderError(
 /** Throws an ExpressionChangedAfterChecked error if checkNoChanges mode is on. */
 export function throwErrorIfNoChangesMode(
     creationMode: boolean, oldValue: any, currValue: any, propName?: string): never|void {
-  const field = propName ? `${propName}: ` : '';
+  const field = propName ? ` for '${propName}'` : '';
   let msg =
-      `ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked. Previous value: '${field}${oldValue}'. Current value: '${field}${currValue}'.`;
+      `ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked. Previous value${field}: '${oldValue}'. Current value: '${currValue}'.`;
   if (creationMode) {
     msg +=
         ` It seems like the view has been created after its parent and its children have been dirty checked.` +
@@ -79,7 +78,8 @@ function constructDetailsForInterpolation(
  * function description.
  */
 export function getExpressionChangedErrorDetails(
-    lView: LView, bindingIndex: number, oldValue: any, newValue: any): any {
+    lView: LView, bindingIndex: number, oldValue: any,
+    newValue: any): {propName?: string, oldValue: any, newValue: any} {
   const tData = lView[TVIEW].data;
   const metadata = tData[bindingIndex];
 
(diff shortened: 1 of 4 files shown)


kara (changes requested):
Thanks for picking this up!

packages/core/test/acceptance/change_detection_spec.ts
```diff
@@ -1248,6 +1248,103 @@ describe('change detection', () => {
...
+          () => initWithTemplate(
+              '<div id="Expressions: {{ a }} and {{ unstableStringExpression }}!"></div>'))
+          .toThrowError(
+              /Previous value: 'id: Expressions: a and initial!'. Current value: 'id: Expressions: a and changed!'/);
+    });
+
+    it('should only display a value of an expression that was changed in text interpolation',
+       () => {
+         expect(() => initWithTemplate('Expressions: {{ a }} and {{ unstableStringExpression }}!'))
+             .toThrowError(/Previous value: '.*?initial'. Current value: '.*?changed'/);
```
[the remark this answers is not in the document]
AndrewKushnir: Thanks Kara, I've created a ticket with additional improvements that we can make once this PR lands: #34454.

packages/core/src/render3/errors.ts
```diff
@@ -52,3 +42,75 @@ export function throwInvalidProviderError(
...
+    return {propName: metadata, oldValue, newValue};
+  }
+
+  // metadata is not available for this expression, check if this expressions is a part of the
+  // property interpolation by going from the current binding index left and look for a string that
+  // contains INTERPOLATION_DELIMITER, the layout in tView.data for this case will look like this:
+  // [..., 'id�Prefix � and � suffix', null, null, null, ...]
+  if (metadata === null) {
+    let idx = bindingIndex - 1;
+    while (typeof tData[idx] !== 'string' && tData[idx + 1] === null && idx >= getBindingRoot()) {
```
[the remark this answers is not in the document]
AndrewKushnir: My goal was to avoid going too far "left", i.e. beyond the index where bindings start. However that should not be critical for the error message details assembling logic, so I removed the check.

AndrewKushnir:
@kara thanks for the review! I've addressed your comments, could you please have another look?

packages/core/src/render3/errors.ts
```diff
@@ -52,3 +41,76 @@ export function throwInvalidProviderError(
...
+/** Throws an ExpressionChangedAfterChecked error if checkNoChanges mode is on. */
+export function throwErrorIfNoChangesMode(
+    creationMode: boolean, oldValue: any, currValue: any, propName?: string): never|void {
+  const field = propName ? ` for '${propName}'` : '';
+  let msg =
+      `ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked. Previous value${field}: '${oldValue}'. Current value: '${currValue}'.`;
+  if (creationMode) {
+    msg +=
+        ` It seems like the view has been created after its parent and its children have been dirty checked.` +
+        ` Has it been created in a change detection hook ?`;
```
Airblader: Not your change, but there should be no space before the question mark.

packages/core/src/render3/errors.ts
```diff
@@ -52,3 +41,76 @@ export function throwInvalidProviderError(
...
+  }
+  // TODO: include debug context
```
Airblader: Isn't this what you're doing now? Or does this mean something else?

packages/core/src/render3/errors.ts
AndrewKushnir: Thanks for noticing that, I've updated the message to remove that whitespace.

packages/core/src/render3/errors.ts
AndrewKushnir: That's a different debug context, see https://github.com/angular/angular/blob/04d4fea3e8784ea81916b24f3b153331967bad71/packages/core/src/view/errors.ts#L36. I've updated the comment, so it's more clear.

packages/core/test/acceptance/change_detection_spec.ts
```diff
@@ -1248,6 +1248,114 @@ describe('change detection', () => {
...
+
+    // Note: the test below currently fails in Ivy, but not in VE. VE behavior is correct and Ivy's
+    // logic should be fixed by the upcoming styling refactor, we keep this test to verify that.
+    //
+    // it('should NOT throw for style maps', () => {
+    //  expect(() => initWithTemplate('<div [style]="unstableStyleMapExpression"></div>'))
+    //      .not.toThrowError();
+    // });
+
+    onlyInIvy('VE doesn\'t throw in case of [class] binding change')
```
kara: Can we do the same with this test as the one above? It's also something we should fix.

packages/core/test/acceptance/change_detection_spec.ts
```diff
@@ -1248,6 +1248,114 @@ describe('change detection', () => {
...
+       });
+
+    it('should only display a value of an expression for interpolation inside an i18n property',
+       () => {
+         expect(
+             () => initWithTemplate(
+                 '<div i18n-title title="Expression: {{ unstableStringExpression }}"></div>'))
+             .toThrowError(/Previous value: '.*?initial'. Current value: '.*?changed'/);
+       });
+  });
```
kara: Can we add tests for: 

- `attr.X` cas
- Host binding case

For the "attr" case, it would be nice if we could print the interpolation string and format as "attr.X", but that should probably go in a follow-up as well.

kara (approved):
LGTM, with a few more small things

packages/core/test/acceptance/change_detection_spec.ts
AndrewKushnir: Added extra tests for `attr.x` cases and host bindings. Will improve error message for `attr.x` once this PR lands 👍

AndrewKushnir:
[Presubmit](http://test/OCL:286084253:BASE:286084291:1576628633370:2f6af284)

kara (approved):
LGTM

PR #34404, opened 2019-12-13
fix(ivy): validate the NgModule declarations field

alxhub:
This commit adds three previously missing validations to
NgModule.declarations:

1. It checks that declared classes are actually within the current
   compilation.

2. It checks that declared classes are directives, components, or pipes.

3. It checks that classes are declared in at most one NgModule.

### DIFF
--- a/packages/compiler-cli/src/ngtsc/annotations/src/ng_module.ts
+++ b/packages/compiler-cli/src/ngtsc/annotations/src/ng_module.ts
@@ -9,7 +9,7 @@
 import {CUSTOM_ELEMENTS_SCHEMA, Expression, ExternalExpr, InvokeFunctionExpr, LiteralArrayExpr, LiteralExpr, NO_ERRORS_SCHEMA, R3Identifiers, R3InjectorMetadata, R3NgModuleMetadata, R3Reference, STRING_TYPE, SchemaMetadata, Statement, WrappedNodeExpr, compileInjector, compileNgModule} from '@angular/compiler';
 import * as ts from 'typescript';
 
-import {ErrorCode, FatalDiagnosticError} from '../../diagnostics';
+import {ErrorCode, FatalDiagnosticError, makeDiagnostic} from '../../diagnostics';
 import {DefaultImportRecorder, Reference, ReferenceEmitter} from '../../imports';
 import {MetadataReader, MetadataRegistry} from '../../metadata';
 import {PartialEvaluator, ResolvedValue} from '../../partial_evaluator';
@@ -22,13 +22,14 @@ import {getSourceFile} from '../../util/src/typescript';
 
 import {generateSetClassMetadataCall} from './metadata';
 import {ReferencesRegistry} from './references_registry';
-import {combineResolvers, findAngularDecorator, forwardRefResolver, getValidConstructorDependencies, isExpressionForwardReference, toR3Reference, unwrapExpression, wrapFunctionExpressionsInParens} from './util';
+import {combineResolvers, findAngularDecorator, forwardRefResolver, getReferenceOriginForDiagnostics, getValidConstructorDependencies, isExpressionForwardReference, toR3Reference, unwrapExpression, wrapFunctionExpressionsInParens} from './util';
 
 export interface NgModuleAnalysis {
   mod: R3NgModuleMetadata;
   inj: R3InjectorMetadata;
   metadataStmt: Statement|null;
   declarations: Reference<ClassDeclaration>[];
+  rawDeclarations: ts.Expression|null;
   schemas: SchemaMetadata[];
   imports: Reference<ClassDeclaration>[];
   exports: Reference<ClassDeclaration>[];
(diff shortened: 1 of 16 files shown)


packages/compiler-cli/src/ngtsc/scope/src/local.ts
```diff
@@ -111,9 +118,42 @@ export class LocalModuleScopeRegistry implements MetadataRegistry, ComponentScop
...
+        rawDeclarations: data.rawDeclarations,
+      };
+
+      if (this.duplicateDeclarations.has(decl.node)) {
+        // This directive/pipe has already been identified as being duplicated (possibly just
+        // above). Add this module to the map of modules for which a duplicate declaration exists.
+        this.duplicateDeclarations.get(decl.node) !.set(ngModule, declData);
+
+        // Next declaration.
+        continue;
```
JoostK: Consider refactoring the loop body by extracting out the duplicate handling into a separate method:

```ts
for (const decl of data.declarations) {
  const declData: DeclarationData = {
    ngModule,
    ref: decl,
    rawDeclarations: data.rawDeclarations,
  };
  const isDuplicate = this.processPotentialDuplicate(decl, declData);

  if (!isDuplicate) {
    this.declarationToModule.set(decl.node, declData, ngModule);
  }
}

// ...

private processPotentialDuplicate(decl: ts.Declaration, declData: DeclarationData, ngModule: ts.Declaration): boolean {
  if (this.declarationToModule.has(decl)) {
    const firstDeclData = this.declarationToModule.get(decl) !;
    if (firstDeclData.ngModule !== ngModule) {
      // If the declaration occurs in the same NgModule, it is not considered duplicate.
      return false;
    }

    // This directive/pipe is already declared in another module. Initialize its
    // duplicateDeclarations entry.
    const duplicateDeclMap = new Map<ClassDeclaration, DeclarationData>();
    duplicateDeclMap.set(firstDeclData.ngModule, firstDeclData);
    this.duplicateDeclarations.set(decl, duplicateDeclMap);

    // Remove the directive/pipe from `declarationToModule` as it is considered invalid.
    this.declarationToModule.delete(decl);

    return true;
  }

  if (this.duplicateDeclarations.has(decl)) {
    // This directive/pipe has already been identified as being duplicated. Add this module
    // to the map of modules for which a duplicate declaration exists.
    this.duplicateDeclarations.get(decl) !.set(ngModule, declData);

    return true;
  }

  return false;
}
```

I find this easier to reason about.

packages/compiler-cli/src/ngtsc/annotations/src/util.ts
```diff
@@ -375,4 +376,42 @@ const parensWrapperTransformerFactory: ts.TransformerFactory<ts.Expression> =
...
+    if (decl.rawDeclarations !== null) {
+      // Attempt to find the `ts.Identifier` within the `declarations` of the NgModule for the class
+      // in question. If it can be found, it's the best choice for the error context.
+      const id = decl.ref.getIdentityInExpression(decl.rawDeclarations);
+      if (id !== null) {
+        // The compiler was able to find a `ts.Identifer` within the `declarations` field of the
+        // NgModule. This is the most logical node to highlight for context.
+        contextNode = id;
+      }
+      context.push({
```
JoostK: This must be one block higher.

packages/compiler-cli/test/ngtsc/scope_spec.ts
```diff
@@ -28,6 +28,74 @@ runInEachFileSystem(() => {
...
+
+            export class RandomClass {}
+
+            @NgModule({
+              declarations: [RandomClass],
+            })
+            export class Module {}
+          `);
+
+          env.driveMain();
```
JoostK: I don't think this test is currently passing? This should produce a diagnostic, so `driveMain` throws causing the test to fail.

packages/compiler-cli/test/ngtsc/scope_spec.ts
```diff
@@ -192,3 +261,14 @@ runInEachFileSystem(() => {
     return node as T;
   }
 });
+
+function nodeToClass(node: ts.Node): ts.ClassDeclaration {
```
JoostK: nit:

```suggestion
function findContainingClass(node: ts.Node): ts.ClassDeclaration {
```

packages/compiler-cli/test/ngtsc/scope_spec.ts
```diff
@@ -28,6 +28,74 @@ runInEachFileSystem(() => {
...
+            export class Module {}
+          `);
+          const diags = env.driveDiagnostics();
+          expect(diags.length).toBe(1);
+          const node = diagnosticToNode(diags[0], ts.isIdentifier);
+          expect(node.text).toEqual('ExternalDir');
+          expect(diags[0].messageText).toContain(`not a part of the current compilation`);
+        });
+
+        it('should detect when a declaration is shared between two modules', () => {
```
JoostK: Can you add a test for these two cases:

1. a declaration occurs twice in a single NgModule.
1. a declaration occurs twice in two NgModules, which should then have only two related infos for the two NgModules (not for each declaration, which would have resulted in either 4—or possibly 3—related info messages)

The code is specifically set up to deal with those cases and a test would help to show that.

JoostK (changes requested):

packages/compiler-cli/src/ngtsc/annotations/src/util.ts
alxhub: Actually no, our intention is to skip declarations which don't have `rawDeclarations` entirely. This is mainly .d.ts files, for which showing an error would be pointless.

JoostK: In that case, the code is misleading :-) Moving `contextNode` within the `if (decl.rawDeclarations)` block should make this far more clear.

JoostK (approved):
Thanks for the updates. CI is unhappy but otherwise LGTM!

packages/compiler-cli/src/ngtsc/annotations/src/ng_module.ts
```diff
@@ -104,13 +105,47 @@ export class NgModuleDecoratorHandler implements
...
+      declarationRefs =
+          this.resolveTypeList(rawDeclarations, declarationMeta, name, 'declarations');
+
+      // Look through the declarations to make sure they're all a part of the current compilation.
+      for (const ref of declarationRefs) {
+        if (ref.node.getSourceFile().isDeclarationFile) {
+          // This is an error: declaring something outside of the current compilation.
+          // If no better position for the error can be found, the `declarations` array from the
+          // NgModule is used.
+          let errorNode: ts.Expression = rawDeclarations;
```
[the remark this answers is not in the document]
alxhub: Good idea, done :)

packages/compiler-cli/src/ngtsc/scope/src/local.ts
JoostK: Agreed, I prefer your refactor as well :+1:

PR #34409, opened 2019-12-14
fix(animations): leaking detached nodes when parent has a leave transition

### LINKED ISSUE #25744: Animations: elements with leave transitions never cleanup causing memory leaks
## I'm submitting a...

<pre><code>
[x] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Other... Please describe:
</code></pre>
 
## Current behavior
When a component has a :leave transition the DOM Element reference is never removed from TransitionAnimationEngine.statesByElement map causing detached DOM nodes and out-of-memory crashes.
 
Debugging seems to show that the "hasAnimation" flag is preventing cleanup functions being run (see [_flushAnimations](https://github.com/angular/angular/blob/dd2a650c3455f3bc0a88f8181758a84aacb25fea/packages/animations/browser/src/render/transition_animation_engine.ts#L1204)). Elements never re-enter this function and therefore `removeNodesAfterAnimationDone` or `processLeaveNode` are not triggered.
 
 
## Expected behavior
TransitionAnimationEngine should successfully clean up elements that have left the page.
 
 
1. Create ListItemComponent which has a simple :leave transition in the @Component decorator
2. Include this component with an *ngIf toggle attached to a button in the main AppComponent to trigger the :leave animation
3. Watch detached DOM nodes grow and TransitionAnimationEngine.statesByElement grow
 
Currently this memory leak is causing performance issues and crashes on dashboard monitor type applications which run for a long time with no user input. Over the cause of a day `statesByElement` has grown from 0 elements to 10,000.
 
 
## Environment
 
<pre><code>
Angular version: 6.1.4

 
Browser:
- [x] Chrome (desktop) version 68
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX
</code></pre>

crisbeto:
In the `TransitionAnimationEngine` we keep track of the existing elements with animations and we clear the cached data when they're removed. We also have some logic where we transition away the child elements when a parent is removed, however in that case we never cleared the cached element data which resulted in a memory leak. The leak is particularly visible in Material where whenever there's an animated overlay with a component inside of it that has an animation, the child component would always be retained in memory.

Fixes #25744.

### DIFF
--- a/packages/animations/browser/src/render/transition_animation_engine.ts
+++ b/packages/animations/browser/src/render/transition_animation_engine.ts
@@ -308,11 +308,13 @@ export class AnimationTransitionNamespace {
     }
   }
 
-  private _signalRemovalForInnerTriggers(rootElement: any, context: any, animate: boolean = false) {
+  private _signalRemovalForInnerTriggers(rootElement: any, context: any) {
+    const elements = this._engine.driver.query(rootElement, NG_TRIGGER_SELECTOR, true);
+
     // emulate a leave animation for all inner nodes within this node.
     // If there are no animations found for any of the nodes then clear the cache
     // for the element.
-    this._engine.driver.query(rootElement, NG_TRIGGER_SELECTOR, true).forEach(elm => {
+    elements.forEach(elm => {
       // this means that an inner remove() operation has already kicked off
       // the animation on this element...
       if (elm[REMOVAL_FLAG]) return;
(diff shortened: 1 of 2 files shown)


packages/animations/browser/src/render/transition_animation_engine.ts
```diff
@@ -324,6 +326,11 @@ export class AnimationTransitionNamespace {
         this.clearElementCache(elm);
       }
     });
+
+    // If the child elements were removed along with the parent, their animations might not
+    // have completed. Clear all the elements from the cache so we don't end up with a memory leak.
+    this._engine.afterFlushAnimationsDone(
```
crisbeto: Note that I only have a high level understanding of how animations work so it's possible that this introduces some unintended consequences. I assumed that it's ok, because it passes all of the existing tests and I couldn't see any issues while testing it against the Material dev app.

matsko (approved):
LGTM. Thank you Kristiyan.

kara:
[presubmit](http://test/OCL:285802212:BASE:285802276:1576521608846:e5f3c94)

PR #34429, opened 2019-12-16
ci: fix saucelabs_view_engine master-only job failing

devversion:
Currently the `saucelabs_view_engine` job fails because
the Saucelabs Bazel run script thinks that `--config=saucelabs`
is a flag targeting the actual script. This is not the case and
the flag should be actually part of the bazel command.


devversion:
Example build that shows that these changes fix the job: https://circleci.com/gh/angular/angular/565795

gkalpak (approved):
LGTM

Issue #34431, opened 2019-12-16
Move Template Type Checking info from AOT compiler doc to TTC Doc


kapunahelewong:
# 📚 Docs or angular.io bug report

### Description

 
Currently, there's detailed info in the AOT Compiler doc about Template Type Checking (TTC)  and it should be moved into TTC Doc. That section should be replaced with a section about TTC in the AOT doc with a link to TTC.

### What's the affected URL?**

https://angular.io/guide/aot-compiler#phase-3-template-type-checking

### Reproduction Steps**

N/A

### Expected vs Actual Behavior**

Info should be in one place. TTC doc is new so previously there was no other place for the TTC info. Now it can go into its own doc.

## 🔥 Exception or Error
<pre><code>

</code></pre>

### Anything else relevant?
This is a follow on to https://github.com/angular/angular/pull/34194

Issue #33943, opened 2019-11-20
ngIfUseIfTypeGuard: confusing documentation and deprecation?


Airblader:
### Description

The referenced page describes the following:

> Directives that behave like *ngIf can declare that they want the same treatment by including a static member marker that is a signal to the template compiler to treat them like *ngIf. This static member for *ngIf is:
> ```
> public static ngIfUseIfTypeGuard: void;
> ```

It's not very clear that the pattern has to be `xyzUseIfTypeGuard` for input `xyz` (that's how it is – right?). At the very least this deserves to be polished, however I also wonder whether this feature deserves to be deprecated given that we now have `ngTemplateGuard_xyz` which can do the same and is much more generic.

In fact I was unable to finde a code path that actually implements this (old) logic. Does it even still exist?

### What's the affected URL?**
https://angular.io/guide/aot-compiler#custom-ngif-like-directives

kapunahelewong:
Removing this reference per https://github.com/angular/angular/pull/34435#discussion_r358441328.

JoostK:
The `ngIfUseIfTypeGuard` field is indeed no longer used in Ivy and the documentation has since been removed.

PR #34435, opened 2019-12-16
docs: move template type checking info into template type checking doc

kapunahelewong:
## PR Checklist

## PR Type

Moves Template Type Checking(TTC) info into the new TTC guide. 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #34431

## What is the new behavior?

Now the AOT compiler guide only introduces Phase 3: Template Type Checking and links to TTC guide for details.

- [ ] Yes
- [x] No

## Other information

Also fixes some broken formatting in the current AOT Compiler doc.


aio/content/guide/template-typecheck.md
```diff
@@ -90,6 +90,114 @@ In basic mode, Angular doesn't check either of them.
...
+    template: '<span *ngIf="person"> {{person.addresss.street}} </span>'
+  })
+  class MyComponent {
+    person?: Person;
+  }
+```
+
+Using `*ngIf` allows the TypeScript compiler to infer that the `person` used in the binding expression will never be `undefined`.
+
+### Custom `ngIf` like directives
```
JoostK: My vote would be to remove this as it stands now, also see #33943.

The reason being that this particular syntax is no longer considered in Ivy, as it has been replaced with a more general concept referred to as _template guards_. Alternatively, we can rewrite it using the new syntax.

aio/content/guide/template-typecheck.md
```diff
@@ -90,6 +90,114 @@ In basic mode, Angular doesn't check either of them.
...
+  })
+  class MyComponent {
+    person?: Person;
+  }
+```
+
+This produces the following error:
+
+```bash
+  my.component.ts.MyComponent.html(1,1): : Property 'addresss' does not exist on type 'Person'. Did you mean 'address'?
```
JoostK: This is the "old" formatting of an error, even including the cute double colon 😄

The new style is similar to the following:

```
my.component.html:1:11 - error TS2339: Property 'addresss' does not exist on type 'Person'. Did you mean 'address'?

1         {{person.addresss.street}}
                   ~~~~~~~~
```

aio/content/guide/template-typecheck.md
```diff
@@ -90,6 +90,114 @@ In basic mode, Angular doesn't check either of them.
 However, in full mode, Angular checks that `config` and `user` exist and assumes a type of `any`.
 In strict mode, Angular knows that the `user` in the `<span>` has a type of `User`, and that `address` is an object with a `city` property of type `string`.
 
+Template validation produces error messages when a type error is detected in a template binding  expression, similar to how type errors are reported by the TypeScript compiler against code in a `.ts` file.
```
JoostK: I think we have to play around a bit with the structure of this doc. This is now below `ngFor` checking, which I don't think is desirable.

aio/content/guide/template-typecheck.md
```diff
@@ -90,6 +90,114 @@ In basic mode, Angular doesn't check either of them.
...
+  }
+```
+
+This produces the following error:
+
+```bash
+  my.component.ts.MyComponent.html(1,1): : Property 'addresss' does not exist on type 'Person'. Did you mean 'address'?
+ ```
+
+The file name reported in the error message, `my.component.ts.MyComponent.html`, is a synthetic file
```
JoostK: For Ivy, this is no longer accurate. Either the path of the .ts file is shown for inline templates (`template`), or the path of the .html file is used for external templates (`templateUrl`).

aio/content/guide/template-typecheck.md
```diff
@@ -90,6 +90,114 @@ In basic mode, Angular doesn't check either of them.
...
+## Type narrowing
+
+The expression used in an `ngIf` directive is used to narrow type unions in the Angular
+template compiler, the same way the `if` expression does in TypeScript.
+For example, to avoid `Object is possibly 'undefined'` error in the template above, modify it to only emit the interpolation if the value of `person` is initialized as shown below:
+
+```typescript
+  @Component({
+    selector: 'my-component',
+    template: '<span *ngIf="person"> {{person.addresss.street}} </span>'
```
JoostK:
```suggestion
    template: '<span *ngIf="person"> {{person.address.street}} </span>'
```

aio/content/guide/template-typecheck.md
kapunahelewong: Removed.

aio/content/guide/template-typecheck.md
kapunahelewong: Is this better?

> The file name reported in the error message, `my.component.ts.MyComponent.html`, is either the path of the .ts file is shown for inline templates (template), or the path of the .html file is used for external templates (templateUrl).

> The line and column numbers are relative to the template string in the `@Component()` annotation of the class, `MyComponent` in this case.

aio/content/guide/template-typecheck.md
```diff
@@ -90,6 +90,114 @@ In basic mode, Angular doesn't check either of them.
...
+```typescript
+    public static ngIfUseIfTypeGuard: void;
+```
+
+This declares that the input property `ngIf` of the `NgIf` directive should be treated as a guard to the use of its template, implying that the template will only be instantiated if the `ngIf` input property is true.
+
+
+## Non-null type assertion operator
+
+Use the [non-null type assertion operator](guide/template-syntax#non-null-assertion-operator) to suppress the `Object is possibly 'undefined'` error when it is inconvenient to use `*ngIf` or when some constraint in the component ensures that the expression is always non-null when the binding expression is interpolated.
```
aikidave, line 159:
```suggestion
Use the [non-null type assertion operator](guide/template-syntax#non-null-assertion-operator) to suppress the `Object is possibly 'undefined'` error when it is inconvenient to use `*ngIf`. You can also use this operator when a constraint in the component ensures that the expression is always non-null when the binding expression is interpolated.
```

Note: Be wary of long sentences. Users tend to skip reading them.

aikithoughts (commented):
Should still review for structure, per Joost's comment. Let's do that after the other feedback is incorporated.

aio/content/guide/template-typecheck.md
```diff
@@ -90,6 +90,114 @@ In basic mode, Angular doesn't check either of them.
...
+```
+
+This produces the following error:
+
+```bash
+  my.component.ts.MyComponent.html(1,1): : Property 'addresss' does not exist on type 'Person'. Did you mean 'address'?
+ ```
+
+The file name reported in the error message, `my.component.ts.MyComponent.html`, is a synthetic file
+generated by the template compiler that holds contents of the `MyComponent` class template.
```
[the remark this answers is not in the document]
kapunahelewong: A rebase has changed this sentence significantly since review so that it now reads:

>The file name reported in the error message, `my.component.ts.MyComponent.html`, is either the path of the .ts file is shown for inline templates (template), or the path of the .html file is used for external templates (templateUrl).

I've edited it to:
>The file name reported in the error message, `my.component.ts.MyComponent.html`, is either the path of the `.ts` file for inline templates (template), or the path of the `.html` file for external templates (templateUrl).

aio/content/guide/aot-compiler.md
```diff
@@ -547,127 +547,15 @@ It does not, however, rewrite the `.d.ts` file, so TypeScript doesn't recognize
 One of the Angular compiler's most helpful features is the ability to type-check expressions within templates, and catch any errors before they cause crashes at runtime.
 In the template type-checking phase, the Angular template compiler uses the TypeScript compiler to validate the binding expressions in templates.
 
-Enable this phase explicitly by adding the compiler option `"fullTemplateTypeCheck"` in the `"angularCompilerOptions"` of the project's `tsconfig.json`
-(see [Angular Compiler Options](guide/angular-compiler-options)).
-
 <div class="alert is-helpful">
 
 In [Angular Ivy](guide/ivy), the template type checker has been completely rewritten to be more capable as well as stricter, meaning it can catch a variety of new errors that the previous type checker would not detect.
```
alxhub, line 552: The few paragraphs here seem a little confusing still to me, and I think we need to emphasize that things should be backwards compatible _unless_ you enable `strictTemplates`, so I've attempted a short refactor of the wording here:

> In [Angular Ivy](guide/ivy), the template type checker has been completely rewritten to be more capable as well as stricter, meaning it can catch a variety of new errors that the previous type checker would not detect. This stricter type-checking is not enabled by default in version 9, but can be enabled by setting the `strictTemplates` configuration option.
>
> Templates that previously compiled under View Engine should continue to work with any previous strictness settings, but may fail when `strictTemplates` is enabled.
>
> We do expect to make strict type checking the default in a future version of Angular.

aio/content/guide/template-typecheck.md
```diff
@@ -90,6 +91,104 @@ In basic mode, Angular doesn't check either of them.
...
+    person?: Person;
+  }
+```
+
+This produces the following error:
+
+```bash
+  my.component.html:1:11 - error TS2339: Property 'addresss' does not exist on type 'Person'. Did you mean 'address'?
+
+1         {{person.addresss.street}}
```
alxhub, line 113: It might be helpful to show the underline of the error span on the line here too, if we have a nice way of rendering it.
```suggestion
1         {{person.addresss.street}}
            ~~~~~~~~~~~~~~~
```

aio/content/guide/template-typecheck.md
```diff
@@ -90,6 +91,104 @@ In basic mode, Angular doesn't check either of them.
...
+ ```
+
+The file name reported in the error message, `MyComponent.html`, is either the path of the `.ts` file for inline templates (template), or the path of the `.html` file for external templates (templateUrl).
+
+If a component uses `templateUrl` instead of `template`, the errors are reported in the HTML file referenced by the `templateUrl` instead of a synthetic file.
+
+The error location is the beginning of the text node that contains the interpolation expression with the error.
```
alxhub, line 120: This section is still a little confused.

In general, errors will be correctly highlighted within the file that contains the template.

For `templateUrl`/`.html` templates, this means the error will be shown in the context of the HTML template text.

For `template`/inline templates, the error will be shown just like any other TypeScript error in the file, except it'll point inside the template string at the actual location of the binding/expression/whatever that produced the error. Basically, the _exact_ same error location is shown, but it's within the TypeScript file.

This is all minutiae, though. The real takeaway for users is that template errors are now reported like any other TypeScript error, regardless of what kind of template is in use: you get line/col number references that correctly point to the location of the error, and a nice colored highlight of the erroneous expression within the line.

aio/content/guide/template-typecheck.md
```diff
@@ -90,6 +91,104 @@ In basic mode, Angular doesn't check either of them.
...
+If the error is in an attribute binding such as `[value]="person.address.street"`, the error
+location is the location of the attribute that contains the error.
+
+The validation uses the TypeScript type checker and the options supplied to the TypeScript compiler to control how detailed the type validation is.
+For example, if the `strictTypeChecks` is specified, the error
```
alxhub, line 125:
```suggestion
For example, if the `strictNullChecks` is specified, the error
```

aio/content/guide/template-typecheck.md
```diff
@@ -90,6 +91,104 @@ In basic mode, Angular doesn't check either of them.
...
+
+The file name reported in the error message, `MyComponent.html`, is either the path of the `.ts` file for inline templates (template), or the path of the `.html` file for external templates (templateUrl).
+
+If a component uses `templateUrl` instead of `template`, the errors are reported in the HTML file referenced by the `templateUrl` instead of a synthetic file.
+
+The error location is the beginning of the text node that contains the interpolation expression with the error.
+If the error is in an attribute binding such as `[value]="person.address.street"`, the error
+location is the location of the attribute that contains the error.
+
+The validation uses the TypeScript type checker and the options supplied to the TypeScript compiler to control how detailed the type validation is.
```
alxhub, line 124:
```suggestion
The validation uses the TypeScript type checker and so the options supplied to the TypeScript compiler control how detailed the type validation is.
```
A slight change, but more accurate: because we use the TS type checker, template type-checking is governed by the same settings that apply to other TS code in the project, which are determined by the compiler options.

aio/content/guide/template-typecheck.md
```diff
@@ -90,6 +91,104 @@ In basic mode, Angular doesn't check either of them.
...
+For example, if the `strictTypeChecks` is specified, the error
+```my.component.ts.MyComponent.html(1,1): : Object is possibly 'undefined'```
```
alxhub, line 126: I don't think this is how errors in a project actually appear.

The filename is confused between a TS and HTML file, but also, normal TS errors have an error code (TS1034 or something for example), and the offset won't be at 1,1 (the very first character in the template) but will be at the location of the actual error.

This kind of resembles a View Engine template error which usually did show 1,1 and didn't have TS codes.

aio/content/guide/template-typecheck.md
```diff
@@ -90,6 +91,104 @@ In basic mode, Angular doesn't check either of them.
...
+  @Component({
+    selector: 'my-component',
+    template: '<span *ngIf="person"> {{person.address.street}} </span>'
+  })
+  class MyComponent {
+    person?: Person;
+  }
+```
+
+Using `*ngIf` allows the TypeScript compiler to infer that the `person` used in the binding expression will never be `undefined`.
```
alxhub, line 145: Beautiful example, I love this not only demonstrates how the checker indicates problems, but shows you that the errors reported actually make sense and are actionable.

aio/content/guide/template-typecheck.md
```diff
@@ -90,6 +91,104 @@ In basic mode, Angular doesn't check either of them.
...
+
+
+## Non-null type assertion operator
+
+Use the [non-null type assertion operator](guide/template-syntax#non-null-assertion-operator) to suppress the `Object is possibly 'undefined'` error when it is inconvenient to use `*ngIf`. You can also use this operator when a constraint in the component ensures that the expression is always non-null when the binding expression is interpolated.
```
alxhub, line 150: I legitimately did not know this was possible in template syntax. TIL!

aio/content/guide/template-typecheck.md
```diff
@@ -90,6 +91,104 @@ In basic mode, Angular doesn't check either of them.
...
+
+1         {{person.addresss.street}}
+ ```
+
+The file name reported in the error message, `MyComponent.html`, is either the path of the `.ts` file for inline templates (template), or the path of the `.html` file for external templates (templateUrl).
+
+If a component uses `templateUrl` instead of `template`, the errors are reported in the HTML file referenced by the `templateUrl` instead of a synthetic file.
+
+The error location is the beginning of the text node that contains the interpolation expression with the error.
+If the error is in an attribute binding such as `[value]="person.address.street"`, the error
```
alxhub, line 121: In general, the error should point at the actual part of the expression which failed type-checking.

Certain errors relate to the presence of the binding itself though, and should point at the attribute instead.

aio/content/guide/template-typecheck.md
kapunahelewong: We don't have any special styles for rendering it, but I've added `~` to see how we feel about it in the preview.

aio/content/guide/template-typecheck.md
kapunahelewong: I removed this sentence for the time being and will replace it With the correct error message if we decide to include it. At the moment, I will push up the other updates.

aio/content/guide/template-typecheck.md
kapunahelewong: Holding on this one per https://github.com/angular/angular/pull/34435#discussion_r425318202

Issue #33452, opened 2019-10-28
bazel: unable to enable ngtsc strict template type checking


devversion:
# 🐞 bug report

### Affected Package

@angular/bazel

### Is this a regression?

No

### Description

Type checking performed by ngtsc is currently matching as much as possible with what NGC did. Since ngtsc type checking can be more advanced, there are flags to enable more advanced checking of templates. 

These flags cannot be set from within the `ng_module` Bazel rule unfortunately. The issue is that `ngc-wrapped` does not redirect the new strictness flags. It will be cumbersome to redirect all individual strictness flags, so we need to think about the ideal API for these type checking flags

PR #34438, opened 2019-12-16
feat(forms): expand NgModel disabled type to work with strict template type checking

devversion:
NgModel internally coerces any arbitrary value that will assigned
to the `disabled` `@Input` to a boolean. This has been done to
support the common case where developers set the disabled attribute
without a value. For example:

```html
<input type="checkbox" [(ngModel)]="value" disabled>
```

This worked in View Engine without any errors because inputs were
not strictly checked. In Ivy though, developers can opt-in into
strict template type checking where the attribute would be flagged.

This is because the `NgModel#isDisabled` property type-wise only
accepts a `boolean`. To ensure that the common pattern described
above can still be used, and to reflect the actual runtime behavior,
we should add an acceptance member that makes it work without type
checking errors.

Using a coercion member means that this is not a breaking change.

**Note**: I tried creating a test for this.. but it looks like there is no way to AOT compile a component (replicating the scenario above) with strict template checking enabled. Also see #33452

### DIFF
--- a/packages/forms/src/directives/ng_model.ts
+++ b/packages/forms/src/directives/ng_model.ts
@@ -136,6 +136,16 @@ const resolvedPromise = (() => Promise.resolve(null))();
 export class NgModel extends NgControl implements OnChanges,
     OnDestroy {
   public readonly control: FormControl = new FormControl();
+
+  // At runtime we coerce arbitrary values assigned to the "disabled" input to a "boolean".
+  // This is not reflected in the type of the property because outside of templates, consumers
+  // should only deal with booleans. In templates, a string is allowed for convenience and to
+  // match the native "disabled attribute" semantics which can be observed on input elements.
+  // This static member tells the compiler that values of type "string" can also be assigned
+  // to the input in a template.
+  /** @nodoc */
+  static ngAcceptInputType_disabled: boolean|string;
+
   /** @internal */
   _registered = false;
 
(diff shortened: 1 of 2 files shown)


packages/forms/src/directives/ng_model.ts
```diff
@@ -136,6 +136,16 @@ const resolvedPromise = (() => Promise.resolve(null))();
 export class NgModel extends NgControl implements OnChanges,
     OnDestroy {
   public readonly control: FormControl = new FormControl();
+
+  // At runtime we coerce arbitrary values assigned to the "disabled" input to a "boolean".
+  // This is not reflected in the type of the property because outside of templates, consumers
+  // should only deal with booleans. In templates, a string is allowed for convenience and to
+  // match the native "disabled attribute" semantics which can be observed on input elements.
+  // This static members tells the compiler that values of type "string" can also be assigned
```
kara:
```suggestion
  // This static member tells the compiler that values of type "string" can also be assigned
```

kara (approved):
LGTM once typo is fixed and CI is green

alxhub (approved):

devversion:
@kara can you have another look? I need a global review or from fw-forms and fw-public-api.

kara (approved):
LGTM (global approval)

kara:
[presubmit](http://test/OCL:285845686:BASE:285845740:1576533908666:590e9b47)

PR #34440, opened 2019-12-16
fix(language-service): HTML path should include last node before cursor

kyliau:
Given the following HTML and cursor position:
```
<div c|></div>
      ^ cursor is here
```

Note that the cursor is **after** the attribute `c`.

Under the current implementation, only `Element` is included in the
path. Instead, it should be `Element -> Attribute`.

This bug occurs only for cases where the cursor is right after the Node,
and it is because the `end` position of the span is excluded from the search.
Instead, the `end` position should be included.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/language-service/test/utils_spec.ts
+++ b/packages/language-service/test/utils_spec.ts
@@ -6,8 +6,10 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
+import * as ng from '@angular/compiler';
 import * as ts from 'typescript';
-import {getDirectiveClassLike} from '../src/utils';
+
+import {getDirectiveClassLike, getPathToNodeAtPosition} from '../src/utils';
 
 describe('getDirectiveClassLike()', () => {
   it('should return a directive class', () => {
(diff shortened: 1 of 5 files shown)


packages/language-service/test/utils_spec.ts
```diff
@@ -6,8 +6,10 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
+import * as ng from '@angular/compiler';
 import * as ts from 'typescript';
-import {getDirectiveClassLike} from '../src/utils';
+
```
ayazhafiz: extra newline here?

ayazhafiz (approved):

packages/language-service/test/utils_spec.ts
kyliau: it's automatically inserted by the formatter

PR #34443, opened 2019-12-16
fix(core): Ensure checkNoChanges runs for OnPush components

atscott:
Add additional CheckNoChanges flag to indicate the view needs to be checked during CheckNoChanges pass. This fixes the problem of the dirty flag being cleared during regular change detection and the view subsequently being skipped in CheckNoChanges mode because it's not dirty.

### DIFF
--- a/packages/core/src/render3/instructions/shared.ts
+++ b/packages/core/src/render3/instructions/shared.ts
@@ -1679,7 +1679,7 @@ function refreshComponent(hostLView: LView, componentHostIdx: number): void {
     refreshView(componentTView, componentView, componentTView.template, componentView[CONTEXT]);
 =======
       (componentView[FLAGS] & (LViewFlags.CheckAlways | LViewFlags.Dirty) ||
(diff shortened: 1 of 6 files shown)


atscott:
[VE presubmit](https://test.corp.google.com/ui#id=OCL:285868848:BASE:285880393:1576544097577:e63db1bb)
[Ivy presubmit](https://test.corp.google.com/ui#id=OCL:285868848:BASE:285880397:1576544097621:c3e71a6a)

atscott:
Because [detectChanges can specifically be called without checkNoChanges](https://github.com/angular/angular/blob/5cecd97493025cd940c9ade4ea9f1836d5b05cc8/packages/core/testing/src/component_fixture.ts#L116) -- even in dev mode -- this solution isn't sufficient but helps gives an idea of how breaky the change would be with the presubmits.

atscott:
Closing since this can't be feasibly merged as an error. We could consider making it a warning in the future and should be aware of this issue when investigating zoneless

PR #34445, opened 2019-12-17
fix(language-service): Proper completions for properties and events

### LINKED ISSUE #398: [feature] Support for "bind-", "on-" and "bind-on-" syntax in template
Hi,

angular supports an alternative way of binding properties and events with the prefixes "bind-", "on-" and "bind-on-" on attributes. A support by this plugin would be nice.

example:
```ts
class AnyComponent {
  foo: String = "bar";
}
```

template with angular syntax:
```html
<other-component [foo]="foo"></other-component>
```
alternative html conform way:
```html
<other-component bind-foo="foo"></other-component>
```

### LINKED ISSUE #474: Event binding behavior very weird in VSCode
**Describe the bug**

Please watch this video I recorded.

https://youtu.be/QxY4O76Zl4A

**To Reproduce**

1. find a tag
2. add a `(click)=""` event binding
3. Unable to find right event when typing

**Expected behavior**

I expected to enter right event binding format.

**Logs**

N/A

**Screenshots**

https://youtu.be/QxY4O76Zl4A

**Additional context**

I'm using version `0.900.0`. When I downgrade to version `0.802.3`, the problem is gone.

kyliau:
This commit fixes autocompletions for properties and events bindings.

The language service will no longer provide bindings like (click) or [id].
Instead, it'll infer the context based on the brackets and provide suggestions
without any brackets.

This fix also adds support for alternative binding syntax such as
`bind-`, `on-`, and `bindon`.

PR closes https://github.com/angular/vscode-ng-language-service/issues/398
PR closes https://github.com/angular/vscode-ng-language-service/issues/474

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/language-service/src/completions.ts
+++ b/packages/language-service/src/completions.ts
@@ -45,6 +45,33 @@ const ANGULAR_ELEMENTS: ReadonlyArray<ng.CompletionEntry> = [
   },
 ];
 
+// This is adapted from packages/compiler/src/render3/r3_template_transform.ts
+// to allow empty binding names.
+const BIND_NAME_REGEXP =
+    /^(?:(?:(?:(bind-)|(let-)|(ref-|#)|(on-)|(bindon-)|(@))(.*))|\[\(([^\)]*)\)\]|\[([^\]]*)\]|\(([^\)]*)\))$/;
+enum ATTR {
+  // Group 1 = "bind-"
+  KW_BIND_IDX = 1,
+  // Group 2 = "let-"
+  KW_LET_IDX = 2,
+  // Group 3 = "ref-/#"
+  KW_REF_IDX = 3,
+  // Group 4 = "on-"
+  KW_ON_IDX = 4,
+  // Group 5 = "bindon-"
+  KW_BINDON_IDX = 5,
+  // Group 6 = "@"
+  KW_AT_IDX = 6,
+  // Group 7 = the identifier after "bind-", "let-", "ref-/#", "on-", "bindon-" or "@"
+  IDENT_KW_IDX = 7,
+  // Group 8 = identifier inside [()]
+  IDENT_BANANA_BOX_IDX = 8,
+  // Group 9 = identifier inside []
+  IDENT_PROPERTY_IDX = 9,
+  // Group 10 = identifier inside ()
+  IDENT_EVENT_IDX = 10,
+}
+
 function isIdentifierPart(code: number) {
   // Identifiers consist of alphanumeric characters, '_', or '$'.
   return isAsciiLetter(code) || isDigit(code) || code == $$ || code == $_;
@@ -139,7 +166,7 @@ export function getTemplateCompletions(
             } else if (templatePosition < startTagSpan.end) {
               // We are in the attribute section of the element (but not in an attribute).
               // Return the attribute completions.
-              result = attributeCompletions(templateInfo, path);
+              result = attributeCompletionsForElement(templateInfo, ast.name);
             }
           },
           visitAttribute(ast) {
@@ -190,11 +217,52 @@ export function getTemplateCompletions(
 }
 
 function attributeCompletions(info: AstResult, path: AstPath<HtmlAst>): ng.CompletionEntry[] {
-  const item = path.tail instanceof Element ? path.tail : path.parentOf(path.tail);
-  if (item instanceof Element) {
-    return attributeCompletionsForElement(info, item.name);
+  const attr = path.tail;
+  const elem = path.parentOf(attr);
+  if (!(attr instanceof Attribute) || !(elem instanceof Element)) {
+    return [];
   }
-  return [];
+
+  // TODO: Consider parsing the attrinute name to a proper AST instead of
+  // matching using regex. This is because the regexp would incorrectly identify
+  // bind parts for cases like [()|]
+  //                              ^ cursor is here
+  const bindParts = attr.name.match(BIND_NAME_REGEXP);
+  // TemplateRef starts with '*'. See https://angular.io/api/core/TemplateRef
+  const isTemplateRef = attr.name.startsWith('*');
+  const isBinding = bindParts !== null || isTemplateRef;
+
+  if (!isBinding) {
+    return attributeCompletionsForElement(info, elem.name);
+  }
+
+  const results: string[] = [];
+  const ngAttrs = angularAttributes(info, elem.name);
+  if (!bindParts) {
+    // If bindParts is null then this must be a TemplateRef.
+    results.push(...ngAttrs.templateRefs);
+  } else if (
+      bindParts[ATTR.KW_BIND_IDX] !== undefined ||
+      bindParts[ATTR.IDENT_PROPERTY_IDX] !== undefined) {
+    // property binding via bind- or []
+    results.push(...propertyNames(elem.name), ...ngAttrs.inputs);
+  } else if (
+      bindParts[ATTR.KW_ON_IDX] !== undefined || bindParts[ATTR.IDENT_EVENT_IDX] !== undefined) {
+    // event binding via on- or ()
+    results.push(...eventNames(elem.name), ...ngAttrs.outputs);
+  } else if (
+      bindParts[ATTR.KW_BINDON_IDX] !== undefined ||
+      bindParts[ATTR.IDENT_BANANA_BOX_IDX] !== undefined) {
+    // banana-in-a-box binding via bindon- or [()]
+    results.push(...ngAttrs.bananas);
+  }
+  return results.map(name => {
+    return {
+      name,
+      kind: ng.CompletionKind.ATTRIBUTE,
+      sortText: name,
+    };
+  });
 }
 
 function attributeCompletionsForElement(
@@ -212,26 +280,15 @@ function attributeCompletionsForElement(
     }
   }
 
-  // Add html properties
-  for (const name of propertyNames(elementName)) {
-    results.push({
-      name: `[${name}]`,
-      kind: ng.CompletionKind.ATTRIBUTE,
-      sortText: name,
-    });
-  }
-
-  // Add html events
-  for (const name of eventNames(elementName)) {
-    results.push({
-      name: `(${name})`,
-      kind: ng.CompletionKind.ATTRIBUTE,
-      sortText: name,
-    });
-  }
-
   // Add Angular attributes
-  results.push(...angularAttributes(info, elementName));
+  const ngAttrs = angularAttributes(info, elementName);
+  for (const name of ngAttrs.others) {
+    results.push({
+      name,
+      kind: ng.CompletionKind.ATTRIBUTE,
+      sortText: name,
+    });
+  }
 
   return results;
 }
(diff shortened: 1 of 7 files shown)


packages/language-service/test/completions_spec.ts
```diff
@@ -117,46 +118,21 @@ describe('completions', () => {
     expectContain(completions, CompletionKind.PROPERTY, ['id', 'name']);
   });
 
-  describe('property completions for members of an indexed type', () => {
```
kyliau: This whole section is moved to the bottom of the test file.

packages/language-service/src/completions.ts
```diff
@@ -190,11 +217,71 @@ export function getTemplateCompletions(
...
-    return attributeCompletionsForElement(info, item.name);
+  const attr = path.tail;
+  const elem = path.parentOf(attr);
+  if (!(attr instanceof Attribute) || !(elem instanceof Element)) {
+    return [];
   }
-  return [];
+
+  const isTemplateRef = attr.name.startsWith('*');
+  const bindParts = attr.name.match(BIND_NAME_REGEXP);
```
kyliau: One downside of using this regular expression (instead of a proper AST) is that 
```
[()|]
   ^ cursor is here
```
would still provide results for as if the cursor is inside the banana.
I think it's ok for now, because there's actually no such parser in the compiler.
See code in `packages/compiler/src/render3/r3_template_transform.ts`

packages/language-service/src/completions.ts
```diff
@@ -190,11 +217,71 @@ export function getTemplateCompletions(
...
+  if (!isTemplateRef && !bindParts) {
+    return attributeCompletionsForElement(info, elem.name);
+  }
+
+  const results: ng.CompletionEntry[] = [];
+  const ngAttrs = angularAttributes(info, elem.name);
+  if (!bindParts) {
+    // If bindParts is null then this must be a template reference.
```
kyliau: Somehow TypeScript is not able to infer that if `isTemplateRef` is true here then `bindParts` in the following else statements must be truthy based on the logic on line 228.

ayazhafiz: I agree; this doesn't seem too much of a worry, as users likely will not be typing in that region.

We do not need to do this now, but wouldn't it be possible to check if the cursor is between a `BoundAttribute` and a `BoundEvent` template node to deduce the location in this case?

ayazhafiz (changes requested):

packages/language-service/src/completions.ts
kyliau: > `bindParts` could still be falsy, because if `templateRef` is truthy and `bindParts` is falsy the branch on 228 will still evaluate to true.

I think this statement is incorrect. Both `templateRef` and `bindParts` must be falsy in order for line 228 to evaluate to true.

Let's make this a little more abstract.
Say I have the following code:
```typescript
const isTemplateRef = attr.name.startsWith('*');
const hasBindParts = bindParts !== null;
const isBinding = isTemplateRef || hasBindParts;
if (!isBinding) {
  return;  // isTemplateRef and hasBindParts are both false
}
// At this point, it is one of three possibilities:
// 1. isTemplateRef = true,  hasBindParts = false
// 2. isTemplateRef = false, hasBindParts = true
// 3. isTemplateRef = true,  hasBindParts = true

if (isTemplateRef) {
  // isTemplateRef must be true
} else {
  // isTemplateRef must be false, and therefore, hasBindParts must be true
}
```

However, if the code is written this way, TypeScript could not deduce that `bindParts` is truthy in the `else` statement. So we have to write 
```typescript
if (!bindParts) {
  // ...
} else {
  // can access bindParts[...] here without the non-null assertion operator
}
```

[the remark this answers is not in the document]
kyliau: Yes, I meant the former. Added a link to the code. Thanks for doign the work to disambiguate the two, they are indeed confusing.

packages/language-service/src/completions.ts
```diff
@@ -190,11 +217,71 @@ export function getTemplateCompletions(
...
+      bindParts[ATTR.IDENT_BANANA_BOX_IDX] !== undefined) {
+    // banana-in-a-box binding via bindon- or [()]
+    for (const name of ngAttrs.bananas) {
+      results.push({
+        name,
+        kind: ng.CompletionKind.ATTRIBUTE,
+        sortText: name,
+      });
+    }
+  }
```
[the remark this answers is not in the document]
kyliau: Will make the code more compact for now. I was thinking of making `CompletionKind` more fine-grained, so I left the `kind` there.

packages/language-service/src/completions.ts
```diff
@@ -484,24 +560,54 @@ function getSourceText(template: ng.TemplateSource, span: ng.Span): string {
   return template.source.substring(span.start, span.end);
 }
 
-function angularAttributes(info: AstResult, elementName: string): ng.CompletionEntry[] {
+interface AngularAttributes {
+  /**
+   * Attributes that support the * syntax
+   */
+  templateRefs: Set<string>;
```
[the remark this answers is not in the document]
kyliau: Done

packages/language-service/src/utils.ts
```diff
@@ -57,12 +57,10 @@ export function isNarrower(spanA: Span, spanB: Span): boolean {
...
 export function hasTemplateReference(type: CompileTypeMetadata): boolean {
-  if (type.diDeps) {
-    for (let diDep of type.diDeps) {
-      if (diDep.token && diDep.token.identifier &&
-          identifierName(diDep.token !.identifier !) === 'TemplateRef')
-        return true;
-    }
+  for (const diDep of type.diDeps) {
+    if (diDep.token && diDep.token.identifier &&
+        identifierName(diDep.token.identifier) === 'TemplateRef')
```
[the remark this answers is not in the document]
kyliau: Done

packages/language-service/src/completions.ts
```diff
@@ -45,6 +45,33 @@ const ANGULAR_ELEMENTS: ReadonlyArray<ng.CompletionEntry> = [
...
+  KW_AT_IDX = 6,
+  // Group 7 = the identifier after "bind-", "let-", "ref-/#", "on-", "bindon-" or "@"
+  IDENT_KW_IDX = 7,
+  // Group 8 = identifier inside [()]
+  IDENT_BANANA_BOX_IDX = 8,
+  // Group 9 = identifier inside []
+  IDENT_PROPERTY_IDX = 9,
+  // Group 10 = identifier inside ()
+  IDENT_EVENT_IDX = 10,
+}
```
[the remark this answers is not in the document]
ayazhafiz: Ah, makes sense. Thanks!

When you say this will be going away during integration with Ivy, do you mean that the whole idea of these matching groups will be going away, or just the duplication in the Language Service?

packages/language-service/src/completions.ts
ayazhafiz: Sorry; what I meant to say is that 228 will evaluate to false. But yes, my calculation was still incorrect - bindParts has to be true.

packages/language-service/src/completions.ts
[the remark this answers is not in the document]
ayazhafiz: Ah, makes sense. Thanks!

When you say this will be going away with the Ivy integration, do you mean the whole idea of matching groups or their duplication in the language service?

packages/language-service/src/utils.ts
```diff
@@ -57,12 +57,10 @@ export function isNarrower(spanA: Span, spanB: Span): boolean {
...
 export function hasTemplateReference(type: CompileTypeMetadata): boolean {
-  if (type.diDeps) {
-    for (let diDep of type.diDeps) {
-      if (diDep.token && diDep.token.identifier &&
-          identifierName(diDep.token !.identifier !) === 'TemplateRef')
-        return true;
-    }
+  for (const diDep of type.diDeps) {
+    if (diDep.token && diDep.token.identifier &&
+        identifierName(diDep.token.identifier) === Identifiers.TemplateRef.name)
```
ayazhafiz: `identifierName` can take a null parameter:

https://github.com/angular/angular/blob/6bfe214346b5b7c0e7ede3848c2975a707fbf9ab/packages/compiler/src/compile_metadata.ts#L27

So you can remove the `diDep.token.identifier` check 🙂

packages/language-service/test/completions_spec.ts
```diff
@@ -117,46 +118,21 @@ describe('completions', () => {
...
+    const marker = mockHost.getLocationMarkerFor(TEST_TEMPLATE, 'cursor');
+    const completions = ngLS.getCompletionsAt(TEST_TEMPLATE, marker.start);
+    expectContain(completions, CompletionKind.ATTRIBUTE, [
+      'ngFor',
+      'ngForOf',
+      'ngIf',
+      'ngSwitchCase',
+      'ngSwitchDefault',
+      'ngPluralCase',
+    ]);
```
ayazhafiz: Can we add tests for custom TemplateRefs? This can be done in another PR if needed.

packages/language-service/test/completions_spec.ts
```diff
@@ -383,13 +357,41 @@ describe('completions', () => {
...
+      mockHost.override(TEST_TEMPLATE, `<div number-model (~{cursor})></div>`);
+      const m5 = mockHost.getLocationMarkerFor(TEST_TEMPLATE, 'cursor');
+      const c5 = ngLS.getCompletionsAt(TEST_TEMPLATE, m5.start);
+      expectContain(c5, CompletionKind.ATTRIBUTE, ['outputAlias']);
+
+      // Event binding via on-
+      mockHost.override(TEST_TEMPLATE, `<div number-mode on-~{cursor}></div>`);
+      const m6 = mockHost.getLocationMarkerFor(TEST_TEMPLATE, 'cursor');
+      const c6 = ngLS.getCompletionsAt(TEST_TEMPLATE, m6.start);
+      expectContain(c6, CompletionKind.ATTRIBUTE, ['outputAlias']);
```
ayazhafiz: Considering splitting these into three sets of test cases: input and output bindings, input bindings, output bindings.

packages/language-service/src/completions.ts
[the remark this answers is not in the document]
kyliau: We'll fix the compiler so that it could provide features needed by the language service.
In this case I'd imagine either changing the compiler to parse AST, or expose the regexes and constants required by the language service.

packages/language-service/src/utils.ts
kyliau: done.

packages/language-service/test/completions_spec.ts
kyliau: Will do this in another PR. There isn't any existing directives that provide custom `TemplateRef`.

packages/language-service/test/completions_spec.ts
kyliau: done.

ayazhafiz (approved):

kyliau:
caretaker: ci/angular:size got stuck

PR #31812, opened 2019-07-23
fix(ivy): support attribute selectors in dynamic component creation

kara:
This commit fixes a bug where we assumed all dynamically created
components would have tag-name selectors, so we passed through the
"tag name" to the renderer as the first index of the selector. For
components with attribute selectors, the tag name would be "", so
the renderer would try to create an element with tag name "" and
throw. Now we default to a "div" element the same way that View
Engine did.

Closes #31785

PR #34450, opened 2019-12-17
fix(ivy): properly bootstrap components with attribute selectors

### LINKED ISSUE #34349: Attribute selector when running with Ivy
# 🐞 bug report

### Affected Package
ivy renderer

### Is this a regression?

Yes, the behavior changed between 8.0 (no-ivy) and 9.0 (ivy)

### Description

When the bootstrap component uses an attribute selector  (e.g. '[app-component]') instead of an element selector, the first `<div>` on the page is used to render the component.

## 🔬 Minimal Reproduction

https://github.com/dartzki/angular9-ivy-atttribute-selector

clone, npm install, ng serve
=> the content of AppComponent is rendered inside the first `<div>` instead of the second `<div>` where the attribute is set.
If Ivy is disabled in tsconfig.json, AppComponent is rendered in the second `<div>` as expected.
It seems as if attribute selectors are completely ignored by Ivy.

## 🔥 Exception or Error
the content of AppComponent is rendered inside the first `<div>` instead of the second `<div>` where the attribute is set.

## 🌍  Your Environment

**Angular Version:**
<pre><code>
Angular CLI: 9.0.0-rc.5
Node: 12.12.0
OS: linux x64

Angular: 9.0.0-rc.5
... router
Ivy Workspace: Yes

@angular-devkit/architect         0.900.0-rc.5
@angular-devkit/build-angular     0.900.0-rc.5
@angular-devkit/build-optimizer   0.900.0-rc.5
@angular-devkit/build-webpack     0.900.0-rc.5
@angular-devkit/core              9.0.0-rc.5
@angular-devkit/schematics        9.0.0-rc.5
@ngtools/webpack                  9.0.0-rc.5
@schematics/angular               9.0.0-rc.5
@schematics/update                0.900.0-rc.5

</code></pre>

**Anything else relevant?**

pkozlowski-opensource:
Before this change ivy didn't properly bootstrap components that used selectors more complex than a tag name. The crux of the issue is that we were trying to use the `selectors` property of a component definition in 2 different code paths:
* bootstrap;
* dynamic instantiation.

It tuns out that selector should be interpreted differently for those 2 scenarios:
* bootstrap: we need a selector as string (as written by a directive author);
* dynamic instantiation: we need a tag name only (defaulting to `div` if a given selector doesn't contain tag name).

This PR tries to fix the identified issue by clearly separating 2 usage scenario (and hence having 2 separate fields on a `ComponentFactory`: `selector` and `tagName`). Sadly this PR has one obvious issue - it doesn't _fully_ recreate a selector's string (as written by a user) - to do it properly we would have to add more code (non tree-shakable) to stringify a parsed selector in all cases.  

So it seems like we need to decide on the following trade-offs:
* correctness (as in this PR);
* fixed code size in the framework (fully re-create a selector string from its parsed version);
* generated code size (we could add oryginal selector string to a directive definition). 

I'm leaning on the "be correct and add more code to the framework) side, but opening a draft PR so we can discuss. 

Fixes #34349

### DIFF
--- a/packages/core/src/render3/component_ref.ts
+++ b/packages/core/src/render3/component_ref.ts
@@ -29,7 +29,7 @@ import {ComponentDef} from './interfaces/definition';
 import {TContainerNode, TElementContainerNode, TElementNode} from './interfaces/node';
 import {RNode, RendererFactory3, domRendererFactory3, isProceduralRenderer} from './interfaces/renderer';
 import {LView, LViewFlags, TVIEW, TViewType} from './interfaces/view';
-import {stringifyCSSSelector} from './node_selector_matcher';
+import {stringifyCSSSelectorList} from './node_selector_matcher';
 import {enterView, leaveView} from './state';
 import {defaultScheduler} from './util/misc_utils';
 import {getTNode} from './util/view_utils';
@@ -114,7 +114,7 @@ export class ComponentFactory<T> extends viewEngine_ComponentFactory<T> {
       private componentDef: ComponentDef<any>, private ngModule?: viewEngine_NgModuleRef<any>) {
     super();
     this.componentType = componentDef.type;
-    this.selector = stringifyCSSSelector(componentDef.selectors[0]);
+    this.selector = stringifyCSSSelectorList(componentDef.selectors);
     this.ngContentSelectors =
         componentDef.ngContentSelectors ? componentDef.ngContentSelectors : [];
     this.isBoundToModule = !!ngModule;
@@ -134,7 +134,7 @@ export class ComponentFactory<T> extends viewEngine_ComponentFactory<T> {
 
     const hostRNode = rootSelectorOrNode ?
         locateHostElement(rendererFactory, rootSelectorOrNode, this.componentDef.encapsulation) :
-        // Determine a tag name used to creating host elements when this component is created
+        // Determine a tag name used for creating host elements when this component is created
         // dynamically. Default to 'div' if this component did not specify any tag name in its
         // selector.
         elementCreate(
--- a/packages/core/src/render3/component_ref.ts
+++ b/packages/core/src/render3/component_ref.ts
@@ -36,6 +36,10 @@ import {getTNode} from './util/view_utils';
 import {createElementRef} from './view_engine_compatibility';
 import {RootViewRef, ViewRef} from './view_ref';
 
+// Default tag name to be used while creating a component dynamically in case tag name is not
+// specified in selector.
+const DEFAULT_TAG_NAME = 'div';
+
 export class ComponentFactoryResolver extends viewEngine_ComponentFactoryResolver {
   /**
    * @param ngModule The NgModuleRef to which all resolved factories are bound.
@@ -114,7 +118,7 @@ export class ComponentFactory<T> extends viewEngine_ComponentFactory<T> {
       private componentDef: ComponentDef<any>, private ngModule?: viewEngine_NgModuleRef<any>) {
     super();
     this.componentType = componentDef.type;
-    this.selector = stringifyCSSSelectorForBootstrap(componentDef.selectors[0]);
+    this.selector = stringifyCSSSelectorForBootstrap(componentDef.selectors[0]) || DEFAULT_TAG_NAME;
     this.ngContentSelectors =
         componentDef.ngContentSelectors ? componentDef.ngContentSelectors : [];
     this.isBoundToModule = !!ngModule;
@@ -138,7 +142,7 @@ export class ComponentFactory<T> extends viewEngine_ComponentFactory<T> {
         // dynamically. Default to 'div' if this component did not specify any tag name in its
         // selector.
         elementCreate(
-            this.componentDef.selectors[0][0] as string || 'div',
+            this.componentDef.selectors[0][0] as string || DEFAULT_TAG_NAME,
             rendererFactory.createRenderer(null, this.componentDef), null);
 
     const rootFlags = this.componentDef.onPush ? LViewFlags.Dirty | LViewFlags.IsRoot :
--- a/packages/core/src/render3/component_ref.ts
+++ b/packages/core/src/render3/component_ref.ts
@@ -36,7 +36,6 @@ import {getTNode} from './util/view_utils';
 import {createElementRef} from './view_engine_compatibility';
 import {RootViewRef, ViewRef} from './view_ref';
 
-
 export class ComponentFactoryResolver extends viewEngine_ComponentFactoryResolver {
   /**
    * @param ngModule The NgModuleRef to which all resolved factories are bound.
--- a/packages/core/src/render3/component_ref.ts
+++ b/packages/core/src/render3/component_ref.ts
@@ -29,16 +29,13 @@ import {ComponentDef} from './interfaces/definition';
 import {TContainerNode, TElementContainerNode, TElementNode} from './interfaces/node';
 import {RNode, RendererFactory3, domRendererFactory3, isProceduralRenderer} from './interfaces/renderer';
 import {LView, LViewFlags, TVIEW, TViewType} from './interfaces/view';
-import {stringifyCSSSelectorForBootstrap} from './node_selector_matcher';
+import {stringifyCSSSelector} from './node_selector_matcher';
 import {enterView, leaveView} from './state';
 import {defaultScheduler} from './util/misc_utils';
 import {getTNode} from './util/view_utils';
 import {createElementRef} from './view_engine_compatibility';
 import {RootViewRef, ViewRef} from './view_ref';
 
-// Default tag name to be used while creating a component dynamically in case tag name is not
-// specified in selector.
-const DEFAULT_TAG_NAME = 'div';
 
 export class ComponentFactoryResolver extends viewEngine_ComponentFactoryResolver {
   /**
@@ -118,7 +115,7 @@ export class ComponentFactory<T> extends viewEngine_ComponentFactory<T> {
       private componentDef: ComponentDef<any>, private ngModule?: viewEngine_NgModuleRef<any>) {
     super();
     this.componentType = componentDef.type;
-    this.selector = stringifyCSSSelectorForBootstrap(componentDef.selectors[0]) || DEFAULT_TAG_NAME;
+    this.selector = stringifyCSSSelector(componentDef.selectors[0]);
     this.ngContentSelectors =
         componentDef.ngContentSelectors ? componentDef.ngContentSelectors : [];
     this.isBoundToModule = !!ngModule;
@@ -142,7 +139,7 @@ export class ComponentFactory<T> extends viewEngine_ComponentFactory<T> {
         // dynamically. Default to 'div' if this component did not specify any tag name in its
         // selector.
         elementCreate(
-            this.componentDef.selectors[0][0] as string || DEFAULT_TAG_NAME,
+            this.componentDef.selectors[0][0] as string || 'div',
             rendererFactory.createRenderer(null, this.componentDef), null);
 
     const rootFlags = this.componentDef.onPush ? LViewFlags.Dirty | LViewFlags.IsRoot :
--- a/packages/core/src/render3/component_ref.ts
+++ b/packages/core/src/render3/component_ref.ts
@@ -29,6 +29,7 @@ import {ComponentDef} from './interfaces/definition';
 import {TContainerNode, TElementContainerNode, TElementNode} from './interfaces/node';
 import {RNode, RendererFactory3, domRendererFactory3, isProceduralRenderer} from './interfaces/renderer';
 import {LView, LViewFlags, TVIEW, TViewType} from './interfaces/view';
+import {stringifyCSSSelectorForBootstrap} from './node_selector_matcher';
 import {enterView, leaveView} from './state';
 import {defaultScheduler} from './util/misc_utils';
 import {getTNode} from './util/view_utils';
@@ -113,9 +114,7 @@ export class ComponentFactory<T> extends viewEngine_ComponentFactory<T> {
       private componentDef: ComponentDef<any>, private ngModule?: viewEngine_NgModuleRef<any>) {
     super();
     this.componentType = componentDef.type;
-
-    // default to 'div' in case this component has an attribute selector
-    this.selector = componentDef.selectors[0][0] as string || 'div';
+    this.selector = stringifyCSSSelectorForBootstrap(componentDef.selectors[0]);
     this.ngContentSelectors =
         componentDef.ngContentSelectors ? componentDef.ngContentSelectors : [];
     this.isBoundToModule = !!ngModule;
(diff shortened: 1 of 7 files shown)


pkozlowski-opensource:
Related to #31812

packages/core/src/render3/component_ref.ts
```diff
@@ -114,8 +122,27 @@ export class ComponentFactory<T> extends viewEngine_ComponentFactory<T> {
...
+    const selectorFirstPart = componentDef.selectors[0];
+    let selectorString = selectorFirstPart[0] as string;
+
+    // Determine a tag name used to creating host elements when this component is created
+    // dynamically. Default to 'div' in case this component did not specify any tag name in
+    // its selector.
+    this.tagName = selectorString || 'div';
+
+    // Stringify a selector used to bootstrap a component.
+    for (let i = 1; i < selectorFirstPart.length; i += 2) {
```
AndrewKushnir: Thanks for the PR @pkozlowski-opensource!

I understand that it's a WIP/Draft PR, but just want to share some minor feedback:

- we iterate through the `selectorFirstPart` with +2 increment. May be add a test in case there are more items in `selectorFirstPart` array?
- may be move this logic to a separate function as a private on a class or outside of a class (as helper method)?

Overall LGTM, thanks again 👍

pkozlowski-opensource: Thnx for going over this PR @AndrewKushnir - definitively agree with both comments - initially I've sent just a "quick and dirty" version to kick off discussion. Having agreed on a direction I've cleaned up this PR a bit. Could you please have another look?

packages/core/src/render3/node_selector_matcher.ts
```diff
@@ -297,3 +297,42 @@ export function isSelectorInSelectorList(selector: CssSelector, list: CssSelecto
...
+      if (mode === SelectorFlags.ATTRIBUTE) {
+        const attrValue = selector[i + 1] as string;
+        result += '[' + valueOrMarker + (attrValue.length > 0 ? '="' + attrValue + '"' : '') + ']';
+        i += 2;
+      } else if (mode === SelectorFlags.CLASS) {
+        result += '.' + valueOrMarker;
+        i++;
+      }
+    } else {
+      ngDevMode &&
```
AndrewKushnir: I'm wondering how break-y this assert might be, since the function will be invoked for dynamically created components as well. Will run g3 presubmit to see if there are some targets that might be affected.

packages/core/src/render3/node_selector_matcher.ts
```diff
@@ -297,3 +297,42 @@ export function isSelectorInSelectorList(selector: CssSelector, list: CssSelecto
...
+          assertEqual(
+              valueOrMarker === SelectorFlags.ATTRIBUTE || valueOrMarker === SelectorFlags.CLASS,
+              true,
+              'Only element, attribute and class selectors are supported for components to be bootstrapped');
+      mode = valueOrMarker;
+      i++;
+    }
+  }
+
```
AndrewKushnir: It looks like there is a couple extra whitespaces.

packages/core/src/render3/node_selector_matcher.ts
```diff
@@ -297,3 +297,42 @@ export function isSelectorInSelectorList(selector: CssSelector, list: CssSelecto
...
+                   ':not() selectors are not supported for components to be bootstrapped.');
+  let result = selector[0] as string;
+  let i = 1;
+  let mode = SelectorFlags.ATTRIBUTE;
+  while (i < selector.length) {
+    let valueOrMarker = selector[i];
+    if (typeof valueOrMarker === 'string') {
+      if (mode === SelectorFlags.ATTRIBUTE) {
+        const attrValue = selector[i + 1] as string;
+        result += '[' + valueOrMarker + (attrValue.length > 0 ? '="' + attrValue + '"' : '') + ']';
```
AndrewKushnir: Should we escape double quotes in `attrValue`, may be add a test to check that?

packages/core/src/render3/node_selector_matcher.ts
```diff
@@ -297,3 +297,42 @@ export function isSelectorInSelectorList(selector: CssSelector, list: CssSelecto
...
+        i += 2;
+      } else if (mode === SelectorFlags.CLASS) {
```
AndrewKushnir: We may also need to take `SelectorFlags.ELEMENT` into account here as it might be present in a selector like `[attr] .class span`, so `i` would not progress if we come across `SelectorFlags.ELEMENT`. May be move `i++` outside of inner `if` block (to make sure we always increment it), like:
```
  if (typeof valueOrMarker === 'string') {
    if (mode === SelectorFlags.ATTRIBUTE) {
      // ...
    } else if (mode === SelectorFlags.CLASS) {
      // ...
    }
    i++;  // <----
  } else {
    // ...
  }
```

AndrewKushnir (changes requested):
Thanks for the PR @pkozlowski-opensource.

Looks great, just left a couple comments that might need attention/tests.

packages/core/src/render3/component_ref.ts
AndrewKushnir: Thanks Pawel.

packages/core/src/render3/node_selector_matcher.ts
pkozlowski-opensource: Oh, good point, definitively needs a test! Thnx!

packages/core/src/render3/node_selector_matcher.ts
pkozlowski-opensource: hmm, my assumption (probably wrong!?) was that this will parse to 3 different `CssSelector` and this function deals with only the first one...

packages/core/src/render3/node_selector_matcher.ts
pkozlowski-opensource: > I'm wondering how break-y this assert might be, since the function will be invoked for dynamically created components as well

Hmm, I'm not sure I got you... What would be the exact selector that you've got on your mind?

AndrewKushnir: Selector like `div:not(.classA)`, which would be represented as `[['div', 9, '.classA']]` will fail this assertion. But that component would currently work for dynamic components (created via `vcr.createComponent`). The goal is to support most frequently used selectors, but for dynamically created components bootstrap selector is not important/used, so we may throw unnecessarily for these components. I was thinking maybe convert selector to a getter function, so the `stringifyCSSSelectorForBootstrap` is called (with assertions) only when `selector` is accessed?

pkozlowski-opensource: > `div:not(.classA)`

oh, right, this is a good one, thnx!

AndrewKushnir: I've started [g3 presubmit](http://test/OCL:286223797:BASE:286223842:1576695122096:c710a34d) to see if there are any affected targets, so may be it's a "false alarm" from my side :)

AndrewKushnir: FYI, just confirmed that there are some failing targets in g3 caused by this assert, an example selector is using `:not`: `[attrA]:not([attrB])`, which is represented as `[['','attrA','',3,'attrB','']]` in parsed form.

packages/core/src/render3/node_selector_matcher.ts
AndrewKushnir: I tested a use-case where attribute value contains double quotes and it turned out that CSS parser only extracts a part of the value after the quote (which looks like a possible bug), so no additional changes are required here at this moment.

packages/core/src/render3/node_selector_matcher.ts
AndrewKushnir: I moved `i++` at the very end of the `while` loop, so we always progress.

packages/core/src/render3/node_selector_matcher.ts
AndrewKushnir: I removed asserts and return `null` in case `:not(...)` is present in selector, that triggers a logic to fall back to default `div` selector in `component_ref.ts`. That should avoid breaking changes for selectors with `:not(...)` inside.

AndrewKushnir:
@pkozlowski-opensource @kara I've applied a few changes (see https://github.com/angular/angular/pull/34450/commits/8cebb0744f9e3bc5b763f9126a500e92353237a8 commit) to remove asserts and take care of a case with `:not(...)` rules inside a selector (fall back to `div` for such selectors). Could you please have a look at the code again and let me know if you have any feedback? Thank you.

packages/core/src/render3/component_ref.ts
```diff
@@ -113,9 +118,7 @@ export class ComponentFactory<T> extends viewEngine_ComponentFactory<T> {
       private componentDef: ComponentDef<any>, private ngModule?: viewEngine_NgModuleRef<any>) {
     super();
     this.componentType = componentDef.type;
-
-    // default to 'div' in case this component has an attribute selector
-    this.selector = componentDef.selectors[0][0] as string || 'div';
+    this.selector = stringifyCSSSelectorForBootstrap(componentDef.selectors[0]) || DEFAULT_TAG_NAME;
```
pkozlowski-opensource: @AndrewKushnir clarifying question: the way I'm reading / understanding it is as follows: if we bump into a CSS selector form that we can't stringify (ex. has `:not(...)` part) than don't try to stringily and default to `div` (element) element selector. Is this the intention? If so, is this behaviour consistent with VE? Do you think we could / should have integration test for this particular case?

AndrewKushnir: Thanks for the comment @pkozlowski-opensource 👍 As we discussed, I will add extra logic to support selectors that contain `:not` - in this case the function would be able to stringify all selectors produced by CSS Parser.

AndrewKushnir:
@pkozlowski-opensource as we discussed, I've updated the logic to handle `:not()` selectors as well. Could you please have a look at the changes? Thank you.

AndrewKushnir:
[Presubmit](http://test/OCL:288804718:BASE:288804753:1578531852265:a8e916b5)

pkozlowski-opensource:
LGTM, thnx @AndrewKushnir  for all the updates!

packages/core/src/render3/component_ref.ts
```diff
@@ -135,7 +134,12 @@ export class ComponentFactory<T> extends viewEngine_ComponentFactory<T> {
 
     const hostRNode = rootSelectorOrNode ?
         locateHostElement(rendererFactory, rootSelectorOrNode, this.componentDef.encapsulation) :
-        elementCreate(this.selector, rendererFactory.createRenderer(null, this.componentDef), null);
+        // Determine a tag name used to creating host elements when this component is created
```
kara:
```suggestion
        // Determine a tag name used for creating host elements when this component is created
```

packages/core/test/render3/node_selector_matcher_spec.ts
```diff
@@ -508,3 +508,70 @@ describe('css selector matching', () => {
...
+
+  it('should stringify selector with class names', () => {
+    expect(stringifyCSSSelector(['', SelectorFlags.CLASS, 'foo'])).toBe('.foo');
+    expect(stringifyCSSSelector(['button', SelectorFlags.CLASS, 'foo'])).toBe('button.foo');
+
+    expect(stringifyCSSSelector([
+      'button', SelectorFlags.CLASS, 'foo', 'bar'
+    ])).toBe('button.foo.bar');
+
+    expect(stringifyCSSSelector(['button', SelectorFlags.CLASS, 'foo'])).toBe('button.foo');
```
kara: Dupe of line 528?

packages/core/src/render3/component_ref.ts
```diff
@@ -113,9 +114,7 @@ export class ComponentFactory<T> extends viewEngine_ComponentFactory<T> {
       private componentDef: ComponentDef<any>, private ngModule?: viewEngine_NgModuleRef<any>) {
     super();
     this.componentType = componentDef.type;
-
-    // default to 'div' in case this component has an attribute selector
-    this.selector = componentDef.selectors[0][0] as string || 'div';
+    this.selector = stringifyCSSSelector(componentDef.selectors[0]);
```
kara: I think we still have a problem here because we're only supporting the first selector. VE supports bootstrapping with any selector in the list for that component. @pkozlowski-opensource @AndrewKushnir Is that what is meant in the PR description about trading correctness for size? Either way, should be a follow-up PR to fix/add tests or document a breaking change.

packages/core/src/render3/node_selector_matcher.ts
```diff
@@ -297,3 +297,60 @@ export function isSelectorInSelectorList(selector: CssSelector, list: CssSelecto
   }
   return false;
 }
+
+function maybeWrapInNotSelector(isNegativeMode: boolean, chunk: string): string {
+  return isNegativeMode ? ':not(' + chunk.trim() + ')' : chunk;
+}
+
+export function stringifyCSSSelector(selector: CssSelector): string {
```
kara: Nit: Add JSdoc?

kara (approved):
LGTM, some minor things below

AndrewKushnir:
Hi @kara, thanks for the review!

I've addressed your comments:

- implemented support for selectors that contain multiple chunks (e.g. `.foo, .bar, [baz]`)
- added extra tests to verify that it works as expected
- added JSDoc for the helper function

Could you please have a look once again and let me know if there is any additional feedback.

Thank you.

AndrewKushnir:
[Presubmit](http://test/OCL:289191495:BASE:289191512:1578704263960:20ec2070)

packages/core/src/render3/node_selector_matcher.ts
```diff
@@ -297,3 +297,75 @@ export function isSelectorInSelectorList(selector: CssSelector, list: CssSelecto
...
+  }
+  return result;
+}
+
+/**
+ * Generates string representation of CSS selector in parsed form.
+ *
+ * ComponentDef and DirectiveDef are generated with the selector in parsed form to avoid doing
+ * additional parsing at runtime. However in some cases (for example while bootstrapping a component
+ * on a page), a string version of the selector is required. This function takes parsed form of a
```
kara:
```suggestion
 * ), a string version of the selector is required to query for the host element on the page. This function takes the parsed form of a
```

packages/core/src/render3/node_selector_matcher.ts
```diff
@@ -297,3 +297,75 @@ export function isSelectorInSelectorList(selector: CssSelector, list: CssSelecto
...
+    result += maybeWrapInNotSelector(isNegativeMode, currentChunk);
+  }
+  return result;
+}
+
+/**
+ * Generates string representation of CSS selector in parsed form.
+ *
+ * ComponentDef and DirectiveDef are generated with the selector in parsed form to avoid doing
+ * additional parsing at runtime. However in some cases (for example while bootstrapping a component
```
kara:
```suggestion
 * additional parsing at runtime (for example, for directive matching). However in some cases (for example while bootstrapping a component
```

packages/core/test/acceptance/bootstrap_spec.ts
```diff
@@ -7,16 +7,28 @@
...
-         expect(document.body.textContent).toEqual('works!');
+         const ngModuleRef = await platformBrowserDynamic().bootstrapModule(IdSelectorAppModule);
+         expect(document.body.textContent).toEqual('before|works!');
+         ngModuleRef.destroy();
+       } catch (err) {
+         console.error(err);
+       }
+     }));
+
+  it('should bootstrap using one of selectors from the list',
```
kara: Nice, thanks for adding this test 👍

kara (approved):
LGTM, some nits on comments

AndrewKushnir (approved):

PR #34453, opened 2019-12-17
docs(ivy): document global debugging utilities and clean up API

crisbeto:
Cleans up the public API of the global debugging utilities, documents them and exposes them in the API docs.

### DIFF
--- a/packages/core/src/render3/instructions/change_detection.ts
+++ b/packages/core/src/render3/instructions/change_detection.ts
@@ -32,9 +32,6 @@ export function detectChanges(component: {}): void {
  * detection can be scheduled per component tree.
  *
  * @param component Component to mark as dirty.
- *
- * @publicApi
- * @globalApi ng
  */
 export function markDirty(component: {}): void {
   ngDevMode && assertDefined(component, 'component');
--- a/packages/core/src/render3/instructions/change_detection.ts
+++ b/packages/core/src/render3/instructions/change_detection.ts
@@ -15,44 +15,35 @@ import {detectChangesInternal, markViewDirty, scheduleTick, tickRootContext} fro
 /**
  * Synchronously perform change detection on a component (and possibly its sub-components).
  *
- * This function triggers change detection in a synchronous way on a component. There should
- * be very little reason to call this function directly since a preferred way to do change
- * detection is to {@link markDirty} the component and wait for the scheduler to call this method
- * at some future point in time. This is because a single user action often results in many
- * components being invalidated and calling change detection on each component synchronously
- * would be inefficient. It is better to wait until all components are marked as dirty and
- * then perform single change detection across all of the components
+ * This function triggers change detection in a synchronous way on a component.
  *
  * @param component The component which the change detection should be performed on.
  */
-export function detectChanges<T>(component: T): void {
+export function detectChanges(component: {}): void {
   const view = getComponentViewByInstance(component);
-  detectChangesInternal<T>(view, component);
+  detectChangesInternal(view, component);
 }
 
 /**
- * Mark the component as dirty (needing change detection).
+ * Marks the component as dirty (needing change detection). Marking a component dirty will
+ * schedule a change detection on it at some point in the future.
  *
- * Marking a component dirty will schedule a change detection on this
- * component at some point in the future. Marking an already dirty
- * component as dirty is a noop. Only one outstanding change detection
- * can be scheduled per component tree. (Two components bootstrapped with
- * separate `renderComponent` will have separate schedulers)
- *
- * When the root component is bootstrapped with `renderComponent`, a scheduler
- * can be provided.
+ * Marking an already dirty component as dirty won't do anything. Only one outstanding change
+ * detection can be scheduled per component tree.
  *
  * @param component Component to mark as dirty.
  *
  * @publicApi
+ * @globalApi ng
  */
-export function markDirty<T>(component: T) {
+export function markDirty(component: {}): void {
   ngDevMode && assertDefined(component, 'component');
   const rootView = markViewDirty(getComponentViewByInstance(component)) !;
 
   ngDevMode && assertDefined(rootView[CONTEXT], 'rootContext should be defined');
   scheduleTick(rootView[CONTEXT] as RootContext, RootContextFlags.DetectChanges);
 }
+
 /**
  * Used to perform change detection on the whole application.
  *
--- a/packages/core/src/render3/util/change_detection_utils.ts
+++ b/packages/core/src/render3/util/change_detection_utils.ts
@@ -1,0 +1,24 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {detectChanges, markDirty} from '../instructions/all';
+import {getRootComponents} from './discovery_utils';
+
+/**
+ * Marks a component for check (in case of OnPush components) and synchronously
+ * performs change detection on the application this component belongs to.
+ *
+ * @param component Component to {@link ChangeDetectorRef#markForCheck mark for check}.
+ *
+ * @publicApi
+ * @globalApi ng
+ */
+export function applyChanges(component: {}): void {
+  markDirty(component);
+  getRootComponents(component).forEach(rootComponent => detectChanges(rootComponent));
+}
--- a/packages/core/src/render3/util/discovery_utils.ts
+++ b/packages/core/src/render3/util/discovery_utils.ts
@@ -24,30 +24,34 @@ import {getTNode, unwrapRNode} from './view_utils';
 
 
 /**
- * Returns the component instance associated with a given DOM host element.
- * Elements which don't represent components return `null`.
+ * Retrieves the component instance associated with a given DOM element.
  *
- * @param element Host DOM element from which the component should be retrieved.
- *
- * ```
+ * @usageNotes
+ * Given the following DOM structure:
+ * ```html
  * <my-app>
- *   #VIEW
- *     <div>
- *       <child-comp></child-comp>
- *     </div>
- * </mp-app>
- *
- * expect(getComponent(<child-comp>) instanceof ChildComponent).toBeTruthy();
- * expect(getComponent(<my-app>) instanceof MyApp).toBeTruthy();
+ *   <div>
+ *     <child-comp></child-comp>
+ *   </div>
+ * </my-app>
  * ```
+ * Calling `getComponent` on `<child-comp>` will return the instance of `ChildComponent`
+ * associated with this DOM element.
+ *
+ * Calling the function on `<my-app>` will return the `MyApp` instance.
+ *
+ *
+ * @param element DOM element from which the component should be retrieved.
+ * @returns Component instance associated with the element or `null` if there
+ *    is no component associated with it.
  *
  * @publicApi
+ * @globalApi ng
  */
-export function getComponent<T = {}>(element: Element): T|null {
-  if (!(element instanceof Node)) throw new Error('Expecting instance of DOM Node');
+export function getComponent<T>(element: Element): T|null {
+  assertDomElement(element);
   const context = loadLContext(element, false);
   if (context === null) return null;
-
 
   if (context.component === undefined) {
     context.component = getComponentAtNodeIndex(context.nodeIndex, context.lView);
@@ -56,56 +60,42 @@ export function getComponent<T = {}>(element: Element): T|null {
   return context.component as T;
 }
 
+
 /**
- * Returns the component instance associated with a given DOM host element.
- * Elements which don't represent components return `null`.
+ * If inside an embedded view (e.g. `*ngIf` or `*ngFor`), retrieves the context of the embedded
+ * view that the element is part of. Otherwise retrieves the instance of the component whose view
+ * owns the element (in this case, the result is the same as calling `getOwningComponent`).
  *
- * @param element Host DOM element from which the component should be retrieved.
- *
- * ```
- * <my-app>
- *   #VIEW
- *     <div>
- *       <child-comp></child-comp>
- *     </div>
- * </mp-app>
- *
- * expect(getComponent(<child-comp>) instanceof ChildComponent).toBeTruthy();
- * expect(getComponent(<my-app>) instanceof MyApp).toBeTruthy();
- * ```
+ * @param element Element for which to get the surrounding component instance.
+ * @returns Instance of the component that is around the element or null if the element isn't
+ *    inside any component.
  *
  * @publicApi
+ * @globalApi ng
  */
-export function getContext<T = {}>(element: Element): T|null {
-  if (!(element instanceof Node)) throw new Error('Expecting instance of DOM Node');
+export function getContext<T>(element: Element): T|null {
+  assertDomElement(element);
   const context = loadLContext(element, false);
-  if (context === null) return null;
-
-  return context.lView[CONTEXT] as T;
+  return context === null ? null : context.lView[CONTEXT] as T;
 }
 
 /**
- * Returns the component instance associated with view which owns the DOM element (`null`
- * otherwise).
+ * Retrieves the component instance whose view contains the DOM element.
  *
- * @param element DOM element which is owned by an existing component's view.
+ * For example, if `<child-comp>` is used in the template of `<app-comp>`
+ * (i.e. a `ViewChild` of `<app-comp>`), calling `getOwningComponent` on `<child-comp>`
+ * would return `<app-comp>`.
  *
- * ```
- * <my-app>
- *   #VIEW
- *     <div>
- *       <child-comp></child-comp>
- *     </div>
- * </mp-app>
- *
- * expect(getViewComponent(<child-comp>) instanceof MyApp).toBeTruthy();
- * expect(getViewComponent(<my-app>)).toEqual(null);
- * ```
+ * @param elementOrDir DOM element, component or directive instance
+ *    for which to retrieve the root components.
+ * @returns Component instance whose view owns the DOM element or null if the element is not
+ *    part of a component view.
  *
  * @publicApi
+ * @globalApi ng
  */
-export function getViewComponent<T = {}>(element: Element | {}): T|null {
-  const context = loadLContext(element, false);
+export function getOwningComponent<T>(elementOrDir: Element | {}): T|null {
+  const context = loadLContext(elementOrDir, false);
   if (context === null) return null;
 
   let lView = context.lView;
@@ -119,27 +109,32 @@ export function getViewComponent<T = {}>(element: Element | {}): T|null {
 }
 
 /**
- * Retrieve all root components.
- *
+ * Retrieves all root components associated with a DOM element, directive or component instance.
  * Root components are those which have been bootstrapped by Angular.
  *
- * @param target A DOM element, component or directive instance.
+ * @param elementOrDir DOM element, component or directive instance
+ *    for which to retrieve the root components.
+ * @returns Root components associated with the target object.
  *
  * @publicApi
+ * @globalApi ng
  */
-export function getRootComponents(target: {}): any[] {
-  return [...getRootContext(target).components];
+export function getRootComponents(elementOrDir: Element | {}): {}[] {
+  return [...getRootContext(elementOrDir).components];
 }
 
 /**
- * Retrieves an `Injector` associated with the element, component or directive.
+ * Retrieves an `Injector` associated with an element, component or directive instance.
  *
- * @param target A DOM element, component or directive instance.
+ * @param elementOrDir DOM element, component or directive instance for which to
+ *    retrieve the injector.
+ * @returns Injector associated with the element, component or directive instance.
  *
  * @publicApi
+ * @globalApi ng
  */
-export function getInjector(target: {}): Injector {
-  const context = loadLContext(target, false);
+export function getInjector(elementOrDir: Element | {}): Injector {
+  const context = loadLContext(elementOrDir, false);
   if (context === null) return Injector.NULL;
 
   const tNode = context.lView[TVIEW].data[context.nodeIndex] as TElementNode;
@@ -150,7 +145,6 @@ export function getInjector(target: {}): Injector {
  * Retrieve a set of injection tokens at a given DOM node.
  *
  * @param element Element for which the injection tokens should be retrieved.
- * @publicApi
  */
 export function getInjectionTokens(element: Element): any[] {
   const context = loadLContext(element, false);
@@ -176,26 +170,43 @@ export function getInjectionTokens(element: Element): any[] {
 }
 
 /**
- * Retrieves directives associated with a given DOM host element.
+ * Retrieves directive instances associated with a given DOM element. Does not include
+ * component instances.
  *
- * @param target A DOM element, component or directive instance.
+ * @usageNotes
+ * Given the following DOM structure:
+ * ```
+ * <my-app>
+ *   <button my-button></button>
+ *   <my-comp></my-comp>
+ * </my-app>
+ * ```
+ * Calling `getDirectives` on `<button>` will return an array with an instance of the `MyButton`
+ * directive that is associated with the DOM element.
+ *
+ * Calling `getDirectives` on `<my-comp>` will return an empty array.
+ *
+ * @param element DOM element for which to get the directives.
+ * @returns Array of directives associated with the element.
  *
  * @publicApi
+ * @globalApi ng
  */
-export function getDirectives(target: {}): Array<{}> {
-  const context = loadLContext(target) !;
+export function getDirectives(element: Element): {}[] {
+  const context = loadLContext(element) !;
 
   if (context.directives === undefined) {
     context.directives = getDirectivesAtNodeIndex(context.nodeIndex, context.lView, false);
   }
 
-  return context.directives || [];
+  // The `directives` in this case are a named array called `LComponentView`. Clone the
+  // result so we don't expose an internal data structure in the user's console.
+  return context.directives === null ? [] : [...context.directives];
 }
 
 /**
  * Returns LContext associated with a target passed as an argument.
  * Throws if a given target doesn't have associated LContext.
- *
  */
 export function loadLContext(target: {}): LContext;
 export function loadLContext(target: {}, throwOnNotFound: false): LContext|null;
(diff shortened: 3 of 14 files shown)


packages/core/src/render3/util/discovery_utils.ts
```diff
@@ -57,52 +46,29 @@ export function getComponent<T = {}>(element: Element): T|null {
...
- * @param element Host DOM element from which the component should be retrieved.
- *
- * ```
- * <my-app>
- *   #VIEW
- *     <div>
- *       <child-comp></child-comp>
- *     </div>
- * </mp-app>
+ * Retrieves the instance of the component that an element is inside of.
```
kara: What's the difference between this function and `getViewComponent`?  I think this needs an example that shows it.

packages/core/src/render3/util/discovery_utils.ts
```diff
@@ -277,25 +251,13 @@ export function isBrowserEvents(listener: Listener): boolean {
...
- * ```
+ * Retrieves a list of event listeners associated with a DOM element.
  *
  * @param element Element for which the DOM listeners should be retrieved.
+ * @returns Array of event listeners on the DOM element.
+ *
  * @publicApi
+ * @globalApi ng
  */
 export function getListeners(element: Element): Listener[] {
```
kara: It feels like we should also have docs for `Listener` since we are listing it as a return value. There doesn't seem to be a way to click through right now.

kara (changes requested):
Thanks for picking this up! Some thoughts below

packages/core/src/render3/util/discovery_utils.ts
crisbeto: I wanted to raise this in the next meeting, because I'm not exactly sure what the difference is either. From my testing the return values were identical.

packages/core/src/render3/util/discovery_utils.ts
```diff
@@ -24,31 +24,20 @@ import {getTNode, unwrapRNode} from './view_utils';
...
- * <my-app>
- *   #VIEW
- *     <div>
- *       <child-comp></child-comp>
- *     </div>
- * </mp-app>
- *
- * expect(getComponent(<child-comp>) instanceof ChildComponent).toBeTruthy();
- * expect(getComponent(<my-app>) instanceof MyApp).toBeTruthy();
- * ```
```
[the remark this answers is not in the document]
crisbeto: In general I dropped the examples, because we don't seem to have any other inline examples on AIO. Maybe @kapunahelewong, @gkalpak or @petebacondarwin can say what the convention is?

packages/core/src/render3/instructions/change_detection.ts
```diff
@@ -31,28 +31,25 @@ export function detectChanges<T>(component: T): void {
...
+ * Marking an already dirty component as dirty won't do anything. Only one outstanding change
+ * detection can be scheduled per component tree.
  *
  * @param component Component to mark as dirty.
  *
  * @publicApi
+ * @globalApi ng
  */
-export function markDirty<T>(component: T) {
+export function markDirty<T>(component: T): void {
```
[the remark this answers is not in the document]
crisbeto: I think I missed renaming it to `applyChangeDetection`. That being said, isn't applying change detection different from marking it as dirty? When I hear "apply change detection" I assume that change detection would be run immediately, but that's not the case here.

packages/core/src/render3/util/discovery_utils.ts
crisbeto: My thinking when I decided not to document it was that the way people interact with these utilities is usually in the browser console where they won't have access to interfaces anyway.

packages/core/src/render3/instructions/change_detection.ts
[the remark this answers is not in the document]
kara: Yeah, agree that the name makes it sound like it should be synchronous and local (rather than scheduling an async application tick). It feels like if we want to expose anything, it really should be `detectChanges`.

packages/core/src/render3/util/discovery_utils.ts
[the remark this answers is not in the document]
kara: I think the problem is that the one line explanations are not sufficient without examples. Can we see what it would look like to re-add them with backticks?

packages/core/src/render3/util/discovery_utils.ts
kara: It also means that the return value of this function is completely opaque. You can't click through, so people won't know what this function returns, except that it's an array of an object called `Listener[]`. I think we need to document it.

packages/core/src/render3/util/discovery_utils.ts
kara: @mhevery ?

packages/core/src/render3/util/discovery_utils.ts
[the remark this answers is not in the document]
kapunahelewong: @crisbeto, you're right, we do usually try to have functioning examples, especially in the guides, but we do have some inline examples in cases where it's hypothetical or doesn't need the context of an app. I think the inline ones are more common in the api docs, but maybe @gkalpak or @petebacondarwin would know for sure. (I've worked more in the guides and less in the api docs).

packages/core/src/render3/util/discovery_utils.ts
```diff
@@ -57,52 +46,29 @@ export function getComponent<T = {}>(element: Element): T|null {
...
- *
- * expect(getViewComponent(<child-comp>) instanceof MyApp).toBeTruthy();
- * expect(getViewComponent(<my-app>)).toEqual(null);
- * ```
+ * @returns Component view that owns the DOM element or null if it's not inside a component.
  *
  * @publicApi
+ * @globalApi ng
  */
 export function getViewComponent<T = {}>(element: Element | {}): T|null {
```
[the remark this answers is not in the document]
kara: Re - Misko/I talked offline and we agreed on a compromise - `getOwnerComponent`

packages/core/src/render3/util/discovery_utils.ts
kara: Talked to Misko offline about this and he reminded me of embedded views. This will return the context of the closest embedded view if the element is inside an embedded view. So maybe:

```suggestion
 * If inside an embedded view (e.g. *ngIf or *ngFor), retrieves the context of the embedded view that owns the element.
 * If not, retrieves the instance of the component that owns the element (in this case, the result should be the same as `getOwnerComponent`

```

packages/core/src/render3/instructions/change_detection.ts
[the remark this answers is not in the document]
kara: Chatted a bit with @mhevery about this and we decided:
 - We shouldn't expose `markDirty` (remove it from packages/core/src/render3/global_utils_api.ts in a separate commit)
- We should expose `detectChanges` instead and document it here

crisbeto:
I've reworked it based on the feedback @kara.

packages/core/src/render3/util/discovery_utils.ts
```diff
@@ -119,26 +114,30 @@ export function getViewComponent<T = {}>(element: Element | {}): T|null {
...
  *
- * @param target A DOM element, component or directive instance.
+ * @param target DOM element, component or directive instance for which to retrieve the injector.
+ * @returns Injector associated with the element, component or directive instance.
  *
  * @publicApi
+ * @globalApi ng
  */
-export function getInjector(target: {}): Injector {
+export function getInjector(target: Element | {}): Injector {
```
IgorMinar: plz rename `target` to `element` so that we use consistent naming across all functions

IgorMinar: do we need `| {}` here? why? maybe some of our internal tests need it today because of mocks, but that should not affect the public api which should be just `Element`. Please fix across all functions.

IgorMinar (changes requested):

packages/core/src/render3/instructions/change_detection.ts
```diff
@@ -15,44 +15,34 @@ import {detectChangesInternal, markViewDirty, scheduleTick, tickRootContext} fro
...
- * components being invalidated and calling change detection on each component synchronously
- * would be inefficient. It is better to wait until all components are marked as dirty and
- * then perform single change detection across all of the components
+ * This function triggers change detection in a synchronous way on a component.
  *
  * @param component The component which the change detection should be performed on.
+ * @publicApi
+ * @globalApi ng
  */
 export function detectChanges<T>(component: T): void {
```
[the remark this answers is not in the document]
IgorMinar: but more importantly is this the right function to expose? I discussed this briefly with @kara yesterday and we have two different types of users that need to trigger change detection:
- application developers - those almost always want to trigger global change detection including flushing microtask queue that might have been filled during change detection
- component developers - those might want to trigger local change detection only for a subset of a component tree (similar to [ComponentFixture#detectchanges](https://pr34453-889cf5b.ngbuilds.io/api/core/testing/ComponentFixture#detectchanges))

I argue that unlike in unit tests where ComponentFixture#detectchanges is handy, during debugging triggering local change detection is less important and in fact sometimes undesirable because it could put the app into an inconsistent state.

For these reasons I propose that we initially expose a function that addresses the application developer needs (triggers global change detection) and only if we get requests from component developers expose another one that is triggers local change detection. 

This application developer specific function, should be implemented similar to how `markDirty` works today by triggering a global change detection.

In fact we could go as far as implementing it as:
```
function applyChanges(component?: {}): void
```
where if `component` is provided, we mark it as dirty and trigger synchronous change detection on the root component the provided component belongs to.

if the `component` is not provided, we trigger synchronous change detection on all root components (retrieved via `getRootComponents`).

IgorMinar (changes requested):
added a few more comments.. sorry for all the feedback, but I think the API and the docs really could use a significant cleanup compared to the current state.

packages/core/src/render3/util/discovery_utils.ts
```diff
@@ -24,87 +24,82 @@ import {getTNode, unwrapRNode} from './view_utils';
...
  * ```
  *
+ * @param element DOM element from which the component should be retrieved.
+ * @returns Component instance associated with the element or `null` if there
+ *    is no component associated with it.
+ *
  * @publicApi
+ * @globalApi ng
  */
 export function getComponent<T = {}>(element: Element): T|null {
```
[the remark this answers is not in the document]
crisbeto: I'll just change it now. There are 24 usages of it and none of them are breaking for now.

packages/core/src/render3/util/discovery_utils.ts
crisbeto: It's not just an element though, it can also be a directive/component instance.

crisbeto: The `| {}` here is to signal that component/directive instances are accepted, in addition to a DOM element.

packages/core/src/render3/util/discovery_utils.ts
```diff
@@ -263,10 +277,18 @@ export function loadLContextFromNode(node: Node): LContext {
   return loadLContext(node) !;
 }
 
+/**
+ * Represents an event listener configuration returned from `getListeners`.
+ * @publicApi
```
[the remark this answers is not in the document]
crisbeto: I was thinking of exporting it through there, but it seemed wrong to export an interface through a package that is supposed to be only global functions available at runtime. I don't feel strongly against doing it, but it seems like it would stand out next to the rest of the functions.

packages/core/src/render3/util/discovery_utils.ts
```diff
@@ -263,10 +277,18 @@ export function loadLContextFromNode(node: Node): LContext {
...
+ */
 export interface Listener {
+  /** Name of the event listener. */
   name: string;
+  /** Element that the listener is bound to. */
   element: Element;
+  /** Callback that is invoked when the event is triggered. */
   callback: (value: any) => any;
+  /** Whether the listener is using event capturing. */
   useCapture: boolean|null;
```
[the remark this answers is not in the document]
crisbeto: I agree that it looks weird, but from what I can tell the `null` is there so we can distinguish between browser events and outputs which is used later on inside `DebugNode.listeners`. I see a couple of alternatives:

1. Add a separate flag on `Listener` that indicates whether it's an output.
2. Exclude outputs from the `getListeners` result.

packages/core/src/render3/instructions/change_detection.ts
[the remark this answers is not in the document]
crisbeto: I'm not sure what the action item is for this PR. Should I implement the new utility or leave it as is until we've had a chance to discuss it?

crisbeto (commented):
I've implemented or responded to all of the feedback @IgorMinar.

tools/public_api_guard/global_utils.d.ts
```diff
@@ -1,19 +1,17 @@
...
 
-export declare function getInjector(target: {}): Injector;
+export declare function getInjector(target: Element | {}): Injector;
 
 export declare function getListeners(element: Element): Listener[];
 
-export declare function getRootComponents(target: {}): any[];
-
-export declare function getViewComponent<T = {}>(element: Element | {}): T | null;
+export declare function getOwnerComponent<T = {}>(element: Element | {}): T | null;
```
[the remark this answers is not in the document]
kara: Ha, Misko and I couldn't decide which one of those two sounded better, so we ended up with `getOwnerComponent`. `getOwningComponent` is fine with me too.

packages/core/src/render3/instructions/change_detection.ts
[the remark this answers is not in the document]
kara: Let's implement `applyChanges`. I'm good with what Igor proposes above

packages/core/src/render3/util/discovery_utils.ts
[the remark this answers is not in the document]
kara: Agree that we should probably just export it

packages/core/src/render3/instructions/change_detection.ts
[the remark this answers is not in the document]
kara: Let's implement `applyChanges`. I'm good with what Igor proposes above

packages/core/src/render3/util/discovery_utils.ts
kara: Maybe `elementOrDir`?

packages/core/src/render3/instructions/change_detection.ts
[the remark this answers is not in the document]
crisbeto: I've added `applyChanges` based on what Igor mentioned above, but I'm not sure how the case where a `component` isn't provided will work since, from what I can tell, we don't have a utility that allows us to get all of the root components without a reference DOM node.

crisbeto:
@kara @IgorMinar, I've addressed the latest set of feedback, can you take another look?

packages/core/src/render3/instructions/change_detection.ts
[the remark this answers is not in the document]
IgorMinar: aha! that is interesting. let's move the discussion to `applyChanges` - I'll start a new thread there.

packages/core/src/render3/util/change_detection_utils.ts
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Marks a component as dirty and synchronously performs change detection on it.
+ *
+ * @param component Component for which to detect changes.
+ *
+ * @publicApi
+ * @globalApi ng
+ */
+export function applyChanges(component: {}): void {
```
IgorMinar: continuing @crisbeto's discussion from markDirty:
>I'm not sure how the case where a component isn't provided will work since, from what I can tell, we don't have a utility that allows us to get all of the root components without a reference DOM node.

aha! that is interesting. Doesn't protractor have a way to get hold of all roots via https://next.angular.io/api/core/TestabilityRegistry ? could we use that and change detect all the root components?

IgorMinar (changes requested):
this is much better @crisbeto a few more changes requested but we are close. thanks!

packages/core/src/render3/util/discovery_utils.ts
```diff
@@ -230,17 +253,17 @@ export function getLocalRefs(target: {}): {[key: string]: any} {
...
 /**
- * Retrieve the host element of the component.
- *
- * Use this function to retrieve the host element of the component. The host
- * element is the element which the component is associated with.
+ * Retrieves the host element of a component or directive instance.
+ * The host element is the DOM element that matched the selector of the directive.
  *
- * @param directive Component or Directive for which the host element should be retrieved.
+ * @param target Component or directive instance for which the host element should be retrieved.
```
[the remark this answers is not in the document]
crisbeto: It's not just a component though, it can also be a directive. I named it `target`, because it's consistent with all of the other functions.

tools/public_api_guard/global_utils.d.ts
```diff
@@ -1,19 +1,24 @@
...
-export declare function getHostElement<T>(directive: T): Element;
+export declare function getHostElement<T>(target: T): Element;
 
-export declare function getInjector(target: {}): Injector;
+export declare function getInjector(elementOrDir: Element | {}): Injector;
 
 export declare function getListeners(element: Element): Listener[];
 
-export declare function getRootComponents(target: {}): any[];
+export declare function getOwningComponent<T>(elementOrDir: Element | {}): T | null;
```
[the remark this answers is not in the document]
crisbeto: I saw your comment after the meeting so it's too late to bring it up. I have some suggestions, but they aren't much better. We can call it something like `getContainingComponent` or `getClosestComponent`.

packages/core/src/render3/util/change_detection_utils.ts
crisbeto: I looked into it and we should be able to use `TestabilityRegistry` to get all the root elements, but the problem is that we need an injector so that we can get a hold of a `TestabilityRegistry` in the first place, and we can't do that without a DOM node or a component instance.

crisbeto:
@IgorMinar I've addressed/replied to all the recent feedback.

packages/core/src/render3/util/change_detection_utils.ts
IgorMinar: I see. let's skip this for now. We might come back to this later, but it can be done after 9.0

packages/core/src/render3/util/discovery_utils.ts
```diff
@@ -259,46 +279,59 @@ export function getRenderedText(component: any): string {
...
   callback: (value: any) => any;
-  useCapture: boolean|null;
-}
-
-export function isBrowserEvents(listener: Listener): boolean {
-  // Browser events are those which don't have `useCapture` as boolean.
-  return typeof listener.useCapture === 'boolean';
+  /** Whether the listener is using event capturing. */
+  useCapture: boolean;
+  /** Whether this is a listener for a DOM event. */
```
IgorMinar:
```suggestion
  /** `true` if this is a listener for a DOM event, `false` if this is an `@Output` listener registered via event binding. */
```

tools/public_api_guard/global_utils.d.ts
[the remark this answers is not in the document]
IgorMinar: `getContainingComponent` is not too bad. In fact "containing" it's used more frequently in https://github.com/microsoft/TypeScript/blob/master/lib/lib.dom.d.ts than `owning`. @kara what do you think?

packages/core/src/render3/util/discovery_utils.ts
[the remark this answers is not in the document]
IgorMinar: actually, while iterating on this PR, we renamed all "target" arguments to something else and this is the only one remaining how about `componentOrDirective` if you think that `component` is too constraining?

tools/public_api_guard/global_utils.d.ts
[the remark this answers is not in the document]
crisbeto: I ended up renaming it to `getContainingComponent`.

crisbeto:
The feedback from last night has been addressed.

packages/core/src/render3/util/discovery_utils.ts
```diff
@@ -24,88 +24,88 @@ import {getTNode, unwrapRNode} from './view_utils';
...
  * <my-app>
- *   #VIEW
- *     <div>
- *       <child-comp></child-comp>
- *     </div>
- * </mp-app>
- *
- * expect(getViewComponent(<child-comp>) instanceof MyApp).toBeTruthy();
- * expect(getViewComponent(<my-app>)).toEqual(null);
+ *   <div>
```
kara: Can we add back the "VIEW" marker here? It's a vital distinction because content children do not work this way. As written, `<child-comp>` would get whatever component `my-app` is inside of.

```suggestion
 * #VIEW
 *   <div>
```

tools/public_api_guard/global_utils.d.ts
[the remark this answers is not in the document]
kara: To be honest, I don't love `getContainingComponent` because it's confusing in the case of content children. For ex, below it looks like the containing component is `<comp>`, but it's actually the component that "owns" `<child-comp>` (whatever template `<comp>` and `<child-comp>` are inside of.

```
<comp>
   <child-comp></child-comp>
</comp>
```

I'm good with `getOwnerComponent` though

packages/core/src/render3/util/change_detection_utils.ts
```diff
@@ -0,0 +1,23 @@
...
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {detectChanges, markDirty} from '../instructions/all';
+import {getRootComponents} from './discovery_utils';
+
+/**
+ * Marks a component for check and synchronously performs change detection on it.
+ *
+ * @param component Component for which to detect changes.
```
kara: This doc makes it sound like we're only performing change detection on the component that's passed in, when in fact we're triggering change detection on the whole application. Can we tweak to reflect that? Maybe:

```
Marks a component for check and synchronously performs change detection on the application.

@param component Component to mark dirty 
```

IgorMinar: as previously discussed, let's not use "dirty" terminology, let's stick to "mark for check" which is already part of our api. But otherwise I like the improvement:

```
Marks a component for check (in case of OnPush components) and synchronously performs change detection on the application this component belongs to.

@param component Component to [mark for check](ChangeDetectorRef#markForCheck).
```

packages/core/src/render3/util/discovery_utils.ts
IgorMinar: @kara the dom snippet is supposed to represent what people see and debug in the browser's elements view. There will be no marker there. How can we make this example more clear while writing it from the perspective of a developer debugging the app?

IgorMinar (changes requested):
@crisbeto Kara has a point about the "containing" terminology. But now I'm confused if she's suggesting to go back to "owner" or "owning"? Let's discuss in the fw sync today and get this done.

packages/core/src/render3/util/discovery_utils.ts
```diff
@@ -24,88 +24,88 @@ import {getTNode, unwrapRNode} from './view_utils';
...
-  return context.lView[CONTEXT] as T;
+  return context === null ? null : context.lView[CONTEXT] as T;
 }
 
 /**
- * Returns the component instance associated with view which owns the DOM element (`null`
- * otherwise).
- *
- * @param element DOM element which is owned by an existing component's view.
+ * Retrieves the component instance whose view contains the DOM element.
```
kara:
```suggestion
 * Retrieves the component instance whose template contains the DOM element.
```

packages/core/src/render3/util/change_detection_utils.ts
kara: Yeah, that's better

packages/core/src/render3/util/discovery_utils.ts
kara: I think in this case having the example from a DOM perspective might be more confusing. Instead, maybe we could clarify with more text?  Something like:

```
For example, if `<child-comp>` is used in the template for `<app-comp>` (i.e. a `ViewChild` of `<app-comp>`), calling `getOwningComponent` on `<child-comp>` would return `<app-comp>`.
```

packages/core/src/render3/util/discovery_utils.ts
IgorMinar: This seems a bit confusing. The component instance doesn't directly have a template, it has a view and is an instance of a component class that has a template.

packages/core/src/render3/util/discovery_utils.ts
IgorMinar: sgtm

packages/core/src/render3/util/discovery_utils.ts
kara: Discussed offline, we'll leave it as is

IgorMinar:
in an epic bikeshedding session we settled on getOwningComponent because getOwnerComponent has two consecutive nouns.

IgorMinar (approved):
lgtm after `getContainingComponent` is renamed to `getOwningComponent`

AndrewKushnir:
[Presubmit](http://test/OCL:292372356:BASE:292372393:1580407967978:caa364ca)

packages/core/src/render3/instructions/change_detection.ts
```diff
@@ -15,44 +15,34 @@ import {detectChangesInternal, markViewDirty, scheduleTick, tickRootContext} fro
...
- * detection is to {@link markDirty} the component and wait for the scheduler to call this method
- * at some future point in time. This is because a single user action often results in many
- * components being invalidated and calling change detection on each component synchronously
- * would be inefficient. It is better to wait until all components are marked as dirty and
- * then perform single change detection across all of the components
+ * This function triggers change detection in a synchronous way on a component.
  *
  * @param component The component which the change detection should be performed on.
+ * @publicApi
+ * @globalApi ng
```
kara: `detectChanges` shouldn't be public API. Remove?

IgorMinar:
```suggestion
```

packages/core/src/render3/instructions/change_detection.ts
```diff
@@ -15,44 +15,34 @@ import {detectChangesInternal, markViewDirty, scheduleTick, tickRootContext} fro
...
- * be very little reason to call this function directly since a preferred way to do change
- * detection is to {@link markDirty} the component and wait for the scheduler to call this method
- * at some future point in time. This is because a single user action often results in many
- * components being invalidated and calling change detection on each component synchronously
- * would be inefficient. It is better to wait until all components are marked as dirty and
- * then perform single change detection across all of the components
+ * This function triggers change detection in a synchronous way on a component.
  *
  * @param component The component which the change detection should be performed on.
+ * @publicApi
```
IgorMinar:
```suggestion
```

kara (approved):
LGTM (global)

crisbeto:
Good to go from my end.

PR #34472, opened 2019-12-18
build: use vendored yarn version for bazel node toolchain

devversion:
Instead of downloading Yarn separately for Bazel, we could use
our existing copy of Yarn that we maintain for CI jobs that do
not use Bazel. This ensures we use consistent versions of Yarn
across our CI jobs, and also avoids unnecessary downloads to the
Bazel mirror server, or Yarn download servers.


gkalpak (approved):
Yay :tada:

PR #34473, opened 2019-12-18
fix(language-service): completions after "let x of |" in ngFor

kyliau:
This commit fixes a bug in which we do testing for completions.
Subsequently, this exposes another bug in our implementation whereby
suggestions are not provided in "ngFor" where there should have been.

Currently, multiple test cases are grouped together in a single
template. This requires the template to be somewhat complete so that
test cases that depend on variables declared earlier would pass.

Consider the following case:

```html
  template: `
    <div *ngFor="let ~{for-person}person of ~{for-people}people">
      <span>Name: {{~{for-interp-person}person.~{for-interp-name}name}}</span>
      <span>Age: {{person.~{for-interp-age}age}}</span>
    </div>`,
```

In order to test `~{for-interp-person}`, `people` has to be included after
`~{for-people}`. This means the test case for `~{for-people}` is not
reflective of the actual use case because the variable is already there!
In real case, the expression would be incomplete, and our implementation
failed to take that into account.

This commit breaks such test into individual tests, and fix the bugs in
the underlying implementation.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No


packages/language-service/src/completions.ts
```diff
@@ -531,6 +531,18 @@ class ExpressionVisitor extends NullTemplateVisitor {
       this.addAttributeValuesToCompletions(binding.expression.ast);
       return;
     }
+
+    // If the expression is incomplete, for example *ngFor="let x of |"
+    // binding.expression is null. We could still try to provide suggestions
+    // by looking for symbols that are in scope.
+    const ofLocation = attr.value.indexOf(' of ');
+    if (ofLocation > 0 && valueRelativePosition >= ofLocation + ' of '.length) {
```
ayazhafiz: nit: can you pull `' of '` out as a constant?

ayazhafiz (changes requested):

packages/language-service/src/completions.ts
kyliau: done.

ayazhafiz (approved):

PR #34475, opened 2019-12-18
refactor(ivy): move all element attribute creation code into shared class

matsko:
This patch moves all element attribute creation code (e.g. style, class,
binding names, template refs, etc...) into a shared class which is used
by compiler code that creates element and host-binding element
instructions.

### DIFF
--- a/packages/compiler/src/render3/view/static_attributes_builder.ts
+++ b/packages/compiler/src/render3/view/static_attributes_builder.ts
@@ -41,9 +41,15 @@ export class StaticAttributesBuilder {
     }
   }
 
-  setClassAttribute(value: string) { this._classAttr = value; }
+  setClassAttribute(value: string) {
+    value = value.trim();
+    this._classAttr = value.length === 0 ? null : value;
+  }
 
-  setStyleAttribute(value: string) { this._stylesAttr = value; }
+  setStyleAttribute(value: string) {
+    value = value.trim();
+    this._stylesAttr = value.length === 0 ? null : value;
+  }
 
   private _registerAttribute(attrName: string, value: any) {
     if (!this._keyValueAttrs) {
--- a/packages/compiler/src/render3/view/static_attributes_builder.ts
+++ b/packages/compiler/src/render3/view/static_attributes_builder.ts
@@ -26,7 +26,7 @@ export class StaticAttributesBuilder {
   private _i18nNameAttrs: string[]|null = null;
 
   setAttribute(attrName: string, value: string|o.Expression) {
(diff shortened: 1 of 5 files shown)


packages/compiler/src/render3/view/static_attributes_builder.ts
```diff
@@ -0,0 +1,243 @@
...
+  private _keyValueAttrs: KeyValueEntry[]|null = null;
+  private _classAttr: string|null = null;
+  private _stylesAttr: string|null = null;
+  private _templateNameAttrs: string[]|null = null;
+  private _bindingNameAttrs: string[]|null = null;
+  private _projectAsSelectors: (string|number)[]|null = null;
+  private _i18nNameAttrs: string[]|null = null;
+
+  setAttribute(attrName: string, value: string|o.Expression) {
+    if (!this._keyValueAttrs) {
```
alxhub: I generally prefer explicit checks for `null`ness instead of relying on coercion to a boolean value.

alxhub (approved):

packages/compiler/src/render3/view/static_attributes_builder.ts
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    if (!(value instanceof o.Expression)) {
+      value = Array.isArray(value) ? toLiteralArr(value) : o.literal(value);
+    }
+    attrs.push(value);
+  }
+}
+
+function populateNameAttributeEntries(attrs: o.Expression[], names: any[]): void {
```
alxhub: Can we remove the `any`?

packages/compiler/src/render3/view/static_attributes_builder.ts
```diff
@@ -0,0 +1,243 @@
...
+  for (let i = 0; i < names.length; i++) {
+    attrs.push(o.literal(names[i]));
+  }
+}
+
+function toLiteralArr(values: any[]) {
```
alxhub: Same here.

matsko:
presubmit = https://test.corp.google.com/ui#id=OCL:286311063:BASE:286311085:1576725390254:4c91173d

mhevery (commented):
Minor nit: 1bd70f0812

I defer to @alxhub for final approval as this is a compiler change.

packages/compiler/src/render3/view/static_attributes_builder.ts
alxhub: I still see `names: any[]` here.

alxhub (changes requested):

PR #34477, opened 2019-12-18
refactor: remove extraneous then call

mikebolt:
The purpose of this "then" call is to ensure that the return type is Promise< void >, however this may be done by adding a type assertion instead of adding a call to "then". Using a type assertion leads to more efficient code.

## PR Checklist

- [no] Tests for the changes have been added (for bug fixes / features)
- [no] Docs have been added / updated (for bug fixes / features)

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

unnecessary "then" call is included in transpiled JS output

## What is the new behavior?

it isn't

- [ ] Yes
- [x] No

### DIFF
--- a/packages/compiler-cli/src/ngtsc/transform/src/compilation.ts
+++ b/packages/compiler-cli/src/ngtsc/transform/src/compilation.ts
@@ -127,7 +127,7 @@ export class TraitCompiler {
     visit(sf);
 
     if (preanalyze && promises.length > 0) {
-      return Promise.all(promises).then(() => undefined as void);
+      return Promise.all(promises) as unknown as Promise<void>;
     } else {
       return undefined;
     }


packages/compiler-cli/src/ngtsc/transform/src/compilation.ts
```diff
@@ -127,7 +127,7 @@ export class TraitCompiler {
     visit(sf);
 
     if (preanalyze && promises.length > 0) {
-      return Promise.all(promises).then(() => undefined as void);
+      return Promise.all(promises) as unknown as Promise<void>;
     } else {
```
alfaproject: Since you are here, you can also remove this `else` block entirely, I believe

mikebolt: I don't know if it can be removed without looking at all usages of this function.

petebacondarwin (approved):
LGTM

JoostK:
Heya, to me the current implementation is properly implementing the specified return type and casting it like this would be returning the internal Promise values to the caller at runtime. Although it does indeed introduce an extra microtask, I don't think that warrants the unsound cast here. I'll go ahead and close this, but feel free to convince me below (I don't feel too strongly about it).

PR #34480, opened 2019-12-18
refactor(ivy): remove styleSanitizer instruction in favor of an inline param

matsko:
This patch removes the need for the `styleSanitizer()` instruction in
favor of passing the sanitizer into directly into the styleProp
instruction.

This patch also increases the binding index size for all style/class bindings in preparation for #34418

### DIFF
--- a/packages/compiler-cli/test/compliance/r3_view_compiler_styling_spec.ts
+++ b/packages/compiler-cli/test/compliance/r3_view_compiler_styling_spec.ts
@@ -377,7 +377,6 @@ describe('compiler compliance: styling', () => {
               $r3$.ɵɵelement(0, "div");
             }
             if (rf & 2) {
-              $r3$.ɵɵstyleSanitizer($r3$.ɵɵdefaultStyleSanitizer);
               $r3$.ɵɵstyleMap($ctx$.myStyleExp);
             }
           }
@@ -511,7 +510,6 @@ describe('compiler compliance: styling', () => {
                   $r3$.ɵɵelement(0, "div", 0);
                 }
                 if (rf & 2) {
-                  $r3$.ɵɵstyleSanitizer($r3$.ɵɵdefaultStyleSanitizer);
                   $r3$.ɵɵstyleMap($ctx$.myStyleExp);
                   $r3$.ɵɵstyleProp("width", $ctx$.myWidth)("height", $ctx$.myHeight);
                   $r3$.ɵɵattribute("style", "border-width: 10px", $r3$.ɵɵsanitizeStyle);
@@ -557,8 +555,7 @@ describe('compiler compliance: styling', () => {
                 $r3$.ɵɵelement(0, "div");
               }
               if (rf & 2) {
-                $r3$.ɵɵstyleSanitizer($r3$.ɵɵdefaultStyleSanitizer);
-                $r3$.ɵɵstyleProp("background-image", ctx.myImage);
+                $r3$.ɵɵstyleProp("background-image", ctx.myImage, $r3$.ɵɵdefaultStyleSanitizer);
               }
             },
             encapsulation: 2
@@ -816,7 +813,6 @@ describe('compiler compliance: styling', () => {
               $r3$.ɵɵelement(0, "div");
             }
             if (rf & 2) {
-              $r3$.ɵɵstyleSanitizer($r3$.ɵɵdefaultStyleSanitizer);
               $r3$.ɵɵstyleMap($ctx$.myStyleExp);
               $r3$.ɵɵclassMap($ctx$.myClassExp);
             }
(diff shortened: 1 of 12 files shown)


matsko:
presubmit = https://test.corp.google.com/ui#id=OCL:286310939:BASE:286310947:1576725301307:f88467c3

packages/core/src/render3/instructions/styling.ts
```diff
@@ -177,12 +162,15 @@ export function ɵɵclassProp(className: string, value: boolean | null): typeof
 function stylingProp(
     tNode: TNode, firstUpdatePass: boolean, lView: LView, bindingIndex: number, prop: string,
     value: boolean | number | SafeValue | string | null | undefined | NO_CHANGE,
-    isClassBased: boolean): boolean {
+    isClassBased: boolean, sanitizer: StyleSanitizeFn | null): boolean {
   let updated = false;
 
+  if (sanitizer) {
+    setCurrentStyleSanitizer(ɵɵdefaultStyleSanitizer);
```
AndrewKushnir: Should `sanitizer` be used here instead?
```
  setCurrentStyleSanitizer(sanitizer);
```

matsko: Fixed. Thank you. It's the same sanitizer so it's difficult to detect that in tests.

AndrewKushnir: Thanks for the fix @matsko 👍 

I was also curious if we need to keep track of the current sanitizer? If sanitizer is now a part of the necessary instruction, this information should be available when it's needed, similar to `attribute` and `property` instructions?

matsko: We do need to keep track of it because the existing algorithm sanitizes values during the flush operation (which doesn't have its own instruction). Right now we can rely on storing and reusing the same sanitizer because there is no way to have multiple sanitizers in Angular.

Misko's follow-up patch removes the need to keep track of the sanitizer, but this PR enables the instructions to be ready so he doesn't have to change them in his patch.

AndrewKushnir: Thanks for additional information Matias!

mhevery (changes requested):
See my comments: https://github.com/angular/angular/commit/2d3b6c659fdc23d136b7439fc7195fdbc570df1e

matsko:
presubmit = https://tap.corp.google.com/ui#id=OCL:290793940:BASE:290794041:1579637043668:84c7c1a0

packages/compiler-cli/test/compliance/r3_view_compiler_styling_spec.ts
```diff
@@ -1424,6 +1414,46 @@ describe('compiler compliance: styling', () => {
...
+              myRepeat = 'no-repeat';
+            }
+          `
+           }
+         };
+
+         const template = `
+            …
+            if (rf & 2) {
+              $r3$.ɵɵstylePropInterpolate1("background", "url(", ctx.myUrl1, ")", $r3$.ɵɵdefaultStyleSanitizer);
```
mhevery: There is no need to pass in sanitizer here. All interpolation will always have the sanitization, so it can just internally include it. This way we can save on the size of the `template` function.

packages/compiler-cli/test/compliance/r3_view_compiler_styling_spec.ts
```diff
@@ -1937,11 +1965,10 @@ describe('compiler compliance: styling', () => {
          expectEmit(result.source, template, 'Incorrect template');
        });
 
-    it('should generate a `styleSanitizer` instruction when a `styleMap` instruction is used',
-       () => {
-         const files = {
-           app: {
-             'spec.ts': `
+    it('should not add a sanitizer param  when a `styleMap` instruction is used', () => {
```
mhevery: just like interpolation, `styleMap` does not need sanitizer to be passed in. This is because `styleMap` always need sanitizer and so we can just reference it from the `styleMap` directly without the need for passing it in as instruction. (We don't gain anything by passing it in, we only lose on size)

packages/compiler-cli/test/compliance/r3_view_compiler_styling_spec.ts
```diff
@@ -1954,25 +1981,24 @@ describe('compiler compliance: styling', () => {
...
 
-         const result = compile(files, angularFiles);
-         expectEmit(result.source, template, 'Incorrect template');
-       });
+      const result = compile(files, angularFiles);
+      expectEmit(result.source, template, 'Incorrect template');
+    });
 
-    it('shouldn\'t generate a `styleSanitizer` instruction when class-based instructions are used',
+    it('shouldn\'t generate a sanitizer param into the styling instruction when class-based instructions are used',
```
mhevery: same here

packages/core/src/render3/interfaces/styling.ts
```diff
@@ -295,33 +295,23 @@ import {LView} from './view';
  * Sanitization is used to prevent invalid style values from being applied to
  * the element.
  *
- * It is enabled in two cases:
+ * It is enabled in three cases:
```
mhevery: why `three`? We only list two.

mhevery (changes requested):

packages/compiler-cli/test/compliance/r3_view_compiler_styling_spec.ts
mhevery: My fault, I missread, this is correct.

packages/compiler-cli/test/compliance/r3_view_compiler_styling_spec.ts
mhevery: sorry I got confused, this is OK.

mhevery (approved):

PR #34481, opened 2019-12-19
fix(ivy): add attributes and classes to host elements based on selector

### LINKED ISSUE #34256: [RC5 IVY] Nested SVG doesn't work anymore
# 🐞 bug report

### Affected Package
I think it's in IVY .. 

### Is this a regression?

YES.
It is working when you disable IVY with Angular 9, and it's working with Angular 8 (without Ivy)

### Description

I have a project that generate a SVG to reflect the state of an hydraulic powerplant. There is a root <SVG> tag then Angular dynamically adds component into it, that are SVG as well.

The structure of the HTML is something like 
```
<body>
<my-app>
  <svg>
    <g>
      <svg>
        <svg>
          <rect></rect>
        </svg>
      </svg>
    </g>
  </svg>
</my-app>
</body>
```

With IVY enabled, you can see the svg in the inspector but it is not rendered by the browser. If you disable IVY it works.

## 🔬 Minimal Reproduction

Since IVY is not supported by Stackblitz, here's a repo : https://github.com/Yohandah/Nested-SVG-Angular-9-IVY 

Steps :

- Clone the repo
- Install dependencies (npm install)
- run `ng serve`
- You can see a blank page, but if you inspect, you will see the SVG
- Cancel `ng serve`
- add the following to `tsconfig.json`

```
"compilerOptions": {.....},
"angularCompilerOptions": {
  "enableIvy": false
}
```
- run `ng serve` 
- You can see the SVG produced (with your eyes, and in the inspector)

## 🔥 Exception or Error
there is no error

## 🌍  Your Environment

**Angular Version:** 
<pre><code>
Angular CLI: 9.0.0-rc.5
Node: 12.13.1
OS: win32 x64

Angular: 9.0.0-rc.5
... cli, common, compiler, compiler-cli, core, forms
... router
Ivy Workspace: Yes

@angular-devkit/architect         0.900.0-rc.5
@angular-devkit/build-angular     0.900.0-rc.5
@angular-devkit/build-optimizer   0.900.0-rc.5
@angular-devkit/build-webpack     0.900.0-rc.5
@angular-devkit/core              9.0.0-rc.5
@angular-devkit/schematics        9.0.0-rc.5
@ngtools/webpack                  9.0.0-rc.5
@schematics/angular               9.0.0-rc.5
@schematics/update                0.900.0-rc.5

</code></pre>

### LINKED ISSUE #35445: Ivy: Component's class and attribute selectors are missing in created html element when router outlet is used
# 🐞 bug report

### Affected Package
Ivy

### Description
If component has class or attribute selectors they are removed when component is created via routing.

Component:
```typescript
import { Component } from "@angular/core";

@Component({
    selector: "my-component.main[test]",
    template: "My component"
})
export class MyComponent {
}
```

When it's used in another template like this:
```html
<my-component class="main" [test]></my-component>
```
then it's created as expected.
```html
<my-component class="main" test>My component</my-component>
```

When it's created via routing in router-outlet, then class and attribute are missing in resulted html:
```html
<my-component>My component</my-component>
```

There is no such problem when Ivy is not enabled.

## 🔬 Minimal Reproduction

https://stackblitz.com/edit/angular-v2cpwe

## 🌍  Your Environment

**Angular Version:**
<pre><code>
Angular CLI: 9.0.2
Node: 12.14.1
OS: win32 x64

Angular: 9.0.1
... router
Ivy Workspace: Yes

@angular-devkit/architect          0.900.2
@angular-devkit/build-angular      0.900.2
@angular-devkit/build-ng-packagr   0.900.2
@angular-devkit/build-optimizer    0.900.2
@angular-devkit/build-webpack      0.900.2
@angular-devkit/core               9.0.2
@angular-devkit/schematics         9.0.2
@angular/cli                       9.0.2
@ngtools/webpack                   9.0.2
@schematics/angular                9.0.2
@schematics/update                 0.900.2
</code></pre>

AndrewKushnir:
In View Engine, host element of dynamically created component received attributes and classes extracted from component's selector. For example, if component selector is `[attr] .class`, the `attr` attribute and `.class` class will be add to host element. This commit adds similar logic to Ivy, to make sure this behavior is aligned with View Engine.

This PR resolves #35445.

## PR Type

- [x] Bugfix

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/render3/component.ts
+++ b/packages/core/src/render3/component.ts
@@ -20,7 +20,7 @@ import {CLEAN_PROMISE, addHostBindingsToExpandoInstructions, addToViewTree, crea
 import {ComponentDef, ComponentType, RenderFlags} from './interfaces/definition';
 import {TElementNode, TNode, TNodeType} from './interfaces/node';
 import {PlayerHandler} from './interfaces/player';
-import {RElement, Renderer3, RendererFactory3, domRendererFactory3, isProceduralRenderer} from './interfaces/renderer';
+import {RElement, Renderer3, RendererFactory3, domRendererFactory3} from './interfaces/renderer';
 import {CONTEXT, HEADER_OFFSET, LView, LViewFlags, RootContext, RootContextFlags, TVIEW, TViewType} from './interfaces/view';
 import {writeDirectClass, writeDirectStyle} from './node_manipulation';
 import {enterView, getPreviousOrParentTNode, leaveView, setSelectedIndex} from './state';
@@ -139,7 +139,7 @@ export function renderComponent<T>(
   try {
     if (rendererFactory.begin) rendererFactory.begin();
     const componentView = createRootComponentView(
-        hostRNode, componentDef, rootView, rendererFactory, renderer, null, sanitizer);
+        hostRNode, componentDef, rootView, rendererFactory, renderer, sanitizer);
     component = createRootComponent(
         componentView, componentDef, rootView, rootContext, opts.hostFeatures || null);
 
@@ -169,8 +169,8 @@ export function renderComponent<T>(
  */
 export function createRootComponentView(
     rNode: RElement | null, def: ComponentDef<any>, rootView: LView,
-    rendererFactory: RendererFactory3, hostRenderer: Renderer3, addVersion: string | null,
-    sanitizer: Sanitizer | null): LView {
+    rendererFactory: RendererFactory3, hostRenderer: Renderer3,
+    sanitizer?: Sanitizer | null): LView {
   const tView = rootView[TVIEW];
   ngDevMode && assertDataInRange(rootView, 0 + HEADER_OFFSET);
   rootView[0 + HEADER_OFFSET] = rNode;
@@ -188,14 +188,8 @@ export function createRootComponentView(
       }
     }
   }
-  const viewRenderer = rendererFactory.createRenderer(rNode, def);
-  if (rNode !== null && addVersion) {
-    ngDevMode && ngDevMode.rendererSetAttribute++;
-    isProceduralRenderer(hostRenderer) ?
-        hostRenderer.setAttribute(rNode, 'ng-version', addVersion) :
-        rNode.setAttribute('ng-version', addVersion);
-  }
 
+  const viewRenderer = rendererFactory.createRenderer(rNode, def);
   const componentView = createLView(
       rootView, getOrCreateTComponentView(def), null,
       def.onPush ? LViewFlags.Dirty : LViewFlags.CheckAlways, rootView[HEADER_OFFSET], tNode,
--- a/packages/core/src/render3/component_ref.ts
+++ b/packages/core/src/render3/component_ref.ts
@@ -30,8 +30,10 @@ import {TContainerNode, TElementContainerNode, TElementNode} from './interfaces/
 import {RNode, RendererFactory3, domRendererFactory3} from './interfaces/renderer';
 import {LView, LViewFlags, TVIEW, TViewType} from './interfaces/view';
 import {MATH_ML_NAMESPACE, SVG_NAMESPACE} from './namespaces';
-import {stringifyCSSSelectorList} from './node_selector_matcher';
+import {writeDirectClass} from './node_manipulation';
+import {extractAttrsAndClassesFromSelector, stringifyCSSSelectorList} from './node_selector_matcher';
 import {enterView, leaveView} from './state';
+import {setUpAttributes} from './util/attrs_utils';
 import {defaultScheduler} from './util/misc_utils';
 import {getTNode} from './util/view_utils';
 import {createElementRef} from './view_engine_compatibility';
@@ -165,7 +167,6 @@ export class ComponentFactory<T> extends viewEngine_ComponentFactory<T> {
     const rootLView = createLView(
         null, rootTView, rootContext, rootFlags, null, null, rendererFactory, hostRenderer,
         sanitizer, rootViewInjector);
-    const addVersion = rootSelectorOrNode && hostRNode ? VERSION.full : null;
 
     // rootView is the parent when bootstrapping
     // TODO(misko): it looks like we are entering view here but we don't really need to as
@@ -179,7 +180,21 @@ export class ComponentFactory<T> extends viewEngine_ComponentFactory<T> {
 
     try {
       const componentView = createRootComponentView(
-          hostRNode, this.componentDef, rootLView, rendererFactory, hostRenderer, addVersion, null);
+          hostRNode, this.componentDef, rootLView, rendererFactory, hostRenderer);
+
+      if (hostRNode) {
+        // extract attributes and classes from the first selector only to match VE behavior
+        const {attrs, classes} = extractAttrsAndClassesFromSelector(this.componentDef.selectors[0]);
+        if (rootSelectorOrNode) {
+          attrs.push('ng-version', VERSION.full);
+        }
+        if (attrs) {
+          setUpAttributes(hostRenderer, hostRNode, attrs);
+        }
+        if (classes && classes.length > 0) {
+          writeDirectClass(hostRenderer, hostRNode, classes.join(' '));
+        }
+      }
 
       tElementNode = getTNode(rootLView[TVIEW], 0) as TElementNode;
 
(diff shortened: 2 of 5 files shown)


AndrewKushnir:
Note: this PR is blocked on PR #34450, since functions that work with selectors (in this PR and in PR #34450) should be combined/refactored.

packages/core/test/acceptance/view_container_ref_spec.ts
```diff
@@ -160,6 +160,80 @@ describe('ViewContainerRef', () => {
       fixture.detectChanges();
       expect(fixture.debugElement.nativeElement.innerHTML).toContain('Hello');
     });
+
+    it('should apply attributes and classes to host element based on selector', () => {
```
kara: Add a test with an SVG namespace? :)

kara (changes requested):
Thanks for picking this up! Some thoughts below

packages/core/test/acceptance/view_container_ref_spec.ts
```diff
@@ -160,6 +160,80 @@ describe('ViewContainerRef', () => {
...
+      @Component({
+        selector: '[attr-a=a] .class-a :not(.class-b) :not([attr-b=b]) .class-c [attr-c]',
```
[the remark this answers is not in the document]
AndrewKushnir: Thanks for the note, updated component selector.

packages/core/src/render3/component.ts
```diff
@@ -188,12 +190,17 @@ export function createRootComponentView(
...
-        rNode.setAttribute('ng-version', addVersion);
+  if (rNode !== null) {
+    if (hostAttrs) {
+      for (let i = 0; i < hostAttrs.length; i += 2) {
+        nativeSetAttribute(hostRenderer, rNode, hostAttrs[i], hostAttrs[i + 1]);
+      }
+    }
+    if (hostClasses && hostClasses.length > 0) {
+      writeDirectClass(hostRenderer, rNode, hostClasses.join(' '));
+    }
```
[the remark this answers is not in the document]
AndrewKushnir: Good point, moved this logic outside this function.

packages/core/src/render3/node_manipulation.ts
```diff
@@ -596,6 +596,13 @@ function nativeAppendOrInsertBefore(
   }
 }
 
+export function nativeSetAttribute(
```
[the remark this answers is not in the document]
AndrewKushnir: It turned out this logic is not needed, I reused existing `setUpAttributes` function.

packages/core/src/render3/component.ts
```diff
@@ -188,12 +190,17 @@ export function createRootComponentView(
...
     }
   }
+
   const viewRenderer = rendererFactory.createRenderer(rNode, def);
-  if (rNode !== null && addVersion) {
-    ngDevMode && ngDevMode.rendererSetAttribute++;
-    isProceduralRenderer(hostRenderer) ?
-        hostRenderer.setAttribute(rNode, 'ng-version', addVersion) :
-        rNode.setAttribute('ng-version', addVersion);
+  if (rNode !== null) {
```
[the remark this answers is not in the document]
AndrewKushnir: This logic was moved to the function that is used to create components dynamically.

packages/core/test/acceptance/view_container_ref_spec.ts
AndrewKushnir: This is done as a part of a followup PR (#35136) that handles namespaces for dynamically created components.

AndrewKushnir:
Thanks for the review @kara. I've updated the code according to the comments/suggestions. Could you please have another look when you get a chance? Thank you.

pkozlowski-opensource:
@AndrewKushnir in the description of this PR you've referenced issue #34256 but #34256 is closed now (and looks like a bit different issue from what this PR is solving?).

On the other hand it seems like this PR should fix #35445

AndrewKushnir:
@pkozlowski-opensource issue #34256 was resolved by a different PR (#35136, to fix namespaces issue with dynamically created components). You are right, the problem described in the #35445 should be fixed by the changes in this PR. I've rebased it, waiting for final review from Kara. Thank you.

packages/core/src/render3/component_ref.ts
```diff
@@ -179,7 +180,21 @@ export class ComponentFactory<T> extends viewEngine_ComponentFactory<T> {
 
     try {
       const componentView = createRootComponentView(
-          hostRNode, this.componentDef, rootLView, rendererFactory, hostRenderer, addVersion, null);
+          hostRNode, this.componentDef, rootLView, rendererFactory, hostRenderer);
+
+      if (hostRNode) {
```
kara: There are three cases that I know of: 

1) User creates component dynamically with `ViewContainerRef.createComponent` or `ComponentFactory.create` (with no selector).  In this case, we will create the component, so we should be adding appropriate attrs / classes to the new element.

2) App is bootstrapping. `ApplicationRef` will call `ComponentFactory.create` with the selector found in the component factory. Because there's a selector, we will locate the correct host element using the selector. In this case, we should not be adding attrs / classes (if we've located it based on the selector, it already has those attrs/classes, so it's wasted work).

3) User creates a component dynamically using `ComponentFactory.create` and passes in a selector. This selector will be used to locate the host element in which to bootstrap the component. In this case, I think we should also avoid adding attrs / classes. We should treat the passed selector as a replacement for the component's selector and ignore what is in the component selector. Otherwise, we could run into cases where the passed selector and the component def's selector don't match and there's no way to reconcile the two (e.g. if they both have tag selectors like 'my-comp').

In this PR, it looks like we would add the attrs / classes in all 3 cases. At the least, I think we should avoid adding attrs / classes in case 2 (bootstrap). I think it's most correct to also avoid adding attrs in case 3 as well, but that could be considered breaking to VE. What do you think?

AndrewKushnir: Thanks for review @kara!

Your proposal makes a lot of sense.

From `ComponentFactory.create` standpoint we should consider 2 cases:

- when **selector or host node is passed** into the function, which means that host node already exists on a page and we just "insert" our component into it. Since that node was already created (inside or outside of the framework), I think we should not "touch" it and just perform component bootstrap.

- when **selector or host node is not passed** into the function - we create a host node ourselves and in this case it makes sense to add attributes and classes based on selector.

The main motivation for the current implementation was to match VE behavior as closely as possible, but I agree that having different logic based on selector presence makes more sense.

pkozlowski-opensource: I would also argue for only adding attrs for the case (2) [as outlined by Kara](https://github.com/angular/angular/pull/34481#discussion_r379682699).

I think that there reasoning of "we don't touch a host node if we don't create it" is very reasonable.

kara: @pkozlowski-opensource You mean only add attrs in case 1? Case 2 is the bootstrap case.

pkozlowski-opensource: @kara sorry, this is what I've meant... My brain is half half-asleep today. Edited my comment...

AndrewKushnir: @kara and @pkozlowski-opensource thanks for reviewing this PR. I've applied the necessary changes to avoid adding attrs/classes in case selector is provided. Could you please take another look when you get a chance? Thank you.

AndrewKushnir:
[Presubmit](http://test/OCL:295775064:BASE:295775132:1582052711278:7f719618)

packages/core/test/acceptance/view_container_ref_spec.ts
```diff
@@ -251,6 +251,98 @@ describe('ViewContainerRef', () => {
...
+      expect(vcrHostElement.getAttribute('attr-b')).toBe(null);
+      expect(vcrHostElement.getAttribute('attr-c')).toBe('');
+
+      // Verify host element for a component created using `factory.createComponent` method when
+      // also passing element selector as an argument.
+      const factoryHostElement = fixture.nativeElement.querySelector('#factory');
+
+      if (ivyEnabled) {
+        // In Ivy, if selector is passed when component is created, matched host node (found using
+        // this selector) retains all atts/classes and selector-based attrs/classes should *not* be
```
kara:
```suggestion
        // this selector) retains all attrs/classes and selector-based attrs/classes should *not* be
```

kara (approved):
LGTM

PR #34482, opened 2019-12-19
fix(ivy): type-checking of properties which map to multiple fields

alxhub:
It's possible to declare multiple inputs for a directive/component which all
map to the same property name. This is usually done in error, as only one of
any bindings to the property will "win".

In the template type-checker, an error was previously being raised as a
result of this ambiguity. Specifically, a type constructor was produced
which required a binding for each field, but only one of the fields had
a value via the binding. TypeScript would (rightfully) error on missing
values for the remaining fields. This ultimately was happening when the
code which generated the default values for "unset" inputs belonging to
directives or pipes used the final mapping from properties to fields as
a source for field names.

Instead, this commit uses the original list of fields to generate unset
input values, which correctly provides values for fields which shared a
property name but didn't receive the final binding.

### DIFF
--- a/packages/compiler-cli/src/ngtsc/typecheck/src/type_check_block.ts
+++ b/packages/compiler-cli/src/ngtsc/typecheck/src/type_check_block.ts
@@ -1187,7 +1187,10 @@ function tcbGetDirectiveInputs(
 
   // To determine which of directive's inputs are unset, we keep track of the set of field names
   // that have not been seen yet. A field is removed from this set once a binding to it is found.
-  const unsetFields = new Set(propMatch.values());
+  // Note: it's actually important here that `propMatch.values()` isn't used, as there can be
+  // multiple fields which share the same property name and only one of them will be listed as a
+  // value in `propMatch`.
+  const unsetFields = new Set(Object.keys(dir.inputs));
 
   el.inputs.forEach(processAttribute);
   el.attributes.forEach(processAttribute);
--- a/packages/compiler-cli/test/ngtsc/template_typecheck_spec.ts
+++ b/packages/compiler-cli/test/ngtsc/template_typecheck_spec.ts
@@ -174,6 +174,32 @@ export declare class AnimationEvent {
           .toEqual(`Argument of type 'string' is not assignable to parameter of type 'number'.`);
     });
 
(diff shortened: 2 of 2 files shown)


packages/compiler-cli/test/ngtsc/template_typecheck_spec.ts
```diff
@@ -174,6 +174,32 @@ export declare class AnimationEvent {
           .toEqual(`Argument of type 'string' is not assignable to parameter of type 'number'.`);
     });
 
+    it('should support one input property mapping to multiple fields', () => {
+      env.write('test.ts', `
+        import {Component, Directive, Input, NgModule} from '@angular/core';
+
+        @Directive({
+          selector: '[dir]',
+          inputs: ['propertyName'],
```
JoostK: I don't think it's relevant that this input is declared here, instead of using `@Input('propertyName')` on an additional field? Having an additional field using the same property name would make it more clear that there's a single input that maps to multiple fields. The test currently depends on the semantics of how `Directive.inputs` and `@Input()` decorators interact with each other: the reader (read: me) has to infer that `inputs: ['propertyName']` is mapped into a field `propertyName` that is not declared, instead of it coalescing with the `@Input('propertyName')` below (which would have meant that it's mapped into `fieldName`, which wouldn't expose the bug).

packages/compiler-cli/src/ngtsc/typecheck/src/type_check_block.ts
```diff
@@ -1187,7 +1187,10 @@ function tcbGetDirectiveInputs(
 
   // To determine which of directive's inputs are unset, we keep track of the set of field names
   // that have not been seen yet. A field is removed from this set once a binding to it is found.
-  const unsetFields = new Set(propMatch.values());
+  // Note: it's actually important here that `propMatch.values()` isn't used, as there can be
+  // multiple fields which share the same property name and only one of them will be listed as a
+  // value in `propMatch`.
+  const unsetFields = new Set(Object.keys(dir.inputs));
```
JoostK: nit:

```suggestion
  const unsetFields = new Set(Object.keys(inputs));
```

alxhub:
Superseded by #34649

PR #34484, opened 2019-12-19
refactor(ivy): move all projection attribute creation code into shared class

matsko:
This patch moves all projection attribute creation code (e.g. attributes,
and ngProjectAs entries) into a shared class which is used by the compiler
code that creates projection instruction code.


mhevery (commented):
Minor fixups: 6b552822f2

LGTM, Since this is compiler I defer final approval to @alxhub

PR #34495, opened 2019-12-19
fix(ivy): do not reset view dirty state in check no changes mode

### LINKED ISSUE #34471: Ivy: components with OnPush cannot be marked as dirty in view lifecycle hooks
With View Engine, it was possible for developers to mark a component as dirty within the view init/checked lifecycle hooks using "markForCheck". This no longer seems to work in Ivy because the dirty-state of the component is _reset_ after the view lifecycle hooks have been executed. 

In View Engine, the order seems slightly different and a component dirty state is reset when it's children have been checked. The lifecycle hooks for the component execute when the parent view is checked. This means that a component can be marked from within the view lifecycle hooks since the dirty state is _not_ reset afterwards (unlike in Ivy)

https://github.com/angular/angular/blob/8eb05964639ea66e602f3c145728cb63ced706cd/packages/core/src/render3/instructions/shared.ts#L478 

**Reproduction**:

[View Engine Demo](https://ng-run.com/edit/Hq5Gbi5wOIsr6Bdymk1q?aot=false&open=app%2Fhello%2Fhello.component.ts)
[Ivy Demo](https://ng-run.com/edit/Hq5Gbi5wOIsr6Bdymk1q?aot=true&open=app%2Fhello%2Fhello.component.ts)

Also here is an acceptance test that shows the issue: https://github.com/angular/angular/pull/30634

The test replicates a scenario where `MatCheckbox` calls `markForCheck` in the ngAfterViewInit lifecycle hook. Below is the sequence of actions View Engine performs in that test cae:

1.  Runs AfterViewInit hooks for children of: MatCheckbox (no children in this test scenario)
2.  Resets "ChecksEnabled"/Dirty-Flag for MatCheckbox instance
3.  Runs AfterViewInit hook for children of: MyTestComp (only children is the checkbox; so it runs the checkbox hook)
4.  AfterViewInit Hook of MatCheckbox marked the checkbox as dirty
5.  Runs AfterViewInit hooks for children of [ROOT]

-> After the change detection cycle, `MatCheckbox` is left as dirty. This is not the case in Ivy.

Reference ticket: FW-1354

devversion:
In Ivy we currently also reset the dirty state of components in the check
no changes change detection cycle. This is not the case in View Engine.

This behavior in Ivy results in inconsistencies and unpredictable behavior of
components using the `OnPush` change detection strategy. One example
is that components can no longer be marked as dirty from view lifecycle hooks
because the dirty flag is reset afterwards in the check-no-changes CD cycle.

Resolves FW-1354. Fixes #34471

### DIFF
--- a/packages/core/src/render3/instructions/shared.ts
+++ b/packages/core/src/render3/instructions/shared.ts
@@ -384,6 +384,7 @@ export function refreshView<T>(
   const flags = lView[FLAGS];
   if ((flags & LViewFlags.Destroyed) === LViewFlags.Destroyed) return;
   enterView(lView, lView[T_HOST]);
+  const checkNoChangesMode = getCheckNoChangesMode();
   try {
     resetPreOrderHookFlags(lView);
 
@@ -392,11 +393,10 @@ export function refreshView<T>(
       executeTemplate(lView, templateFn, RenderFlags.Update, context);
     }
 
-    const checkNoChangesMode = getCheckNoChangesMode();
     const hooksInitPhaseCompleted =
         (flags & LViewFlags.InitPhaseStateMask) === InitPhaseState.InitPhaseCompleted;
 
-    // execute pre-order hooks (OnInit, OnChanges, DoChanges)
+    // execute pre-order hooks (OnInit, OnChanges, DoCheck)
     // PERF WARNING: do NOT extract this to a separate function without running benchmarks
     if (!checkNoChangesMode) {
       if (hooksInitPhaseCompleted) {
@@ -475,7 +475,17 @@ export function refreshView<T>(
     if (tView.firstUpdatePass === true) {
       tView.firstUpdatePass = false;
     }
-    lView[FLAGS] &= ~(LViewFlags.Dirty | LViewFlags.FirstLViewPass);
+
+    // Do not reset the dirty state when running in check no changes mode. We don't want components
+    // to behave differently depending on whether check no changes is enabled or not. For example:
+    // Marking an OnPush component as dirty from within the `ngAfterViewInit` hook in order to
+    // refresh a `NgClass` binding should work. If we would reset the dirty state in the check
+    // no changes cycle, the component would be not be dirty for the next update pass. This would
+    // be different in production mode where the component dirty state is not reset.
+    if (!checkNoChangesMode) {
+      lView[FLAGS] &= ~(LViewFlags.Dirty | LViewFlags.FirstLViewPass);
+    }
+
     leaveViewProcessExit();
   }
 }
--- a/packages/core/test/acceptance/integration_spec.ts
+++ b/packages/core/test/acceptance/integration_spec.ts
@@ -9,7 +9,6 @@ import {CommonModule} from '@angular/common';
 import {Component, ContentChild, Directive, ElementRef, EventEmitter, HostBinding, HostListener, Input, NgModule, OnInit, Output, Pipe, QueryList, TemplateRef, ViewChild, ViewChildren, ViewContainerRef} from '@angular/core';
 import {TVIEW} from '@angular/core/src/render3/interfaces/view';
 import {getLView} from '@angular/core/src/render3/state';
-import {loadLContext} from '@angular/core/src/render3/util/discovery_utils';
 import {ngDevModeResetPerfCounters} from '@angular/core/src/util/ng_dev_mode';
 import {TestBed} from '@angular/core/testing';
 import {By} from '@angular/platform-browser';
--- a/packages/core/test/acceptance/change_detection_spec.ts
+++ b/packages/core/test/acceptance/change_detection_spec.ts
@@ -9,10 +9,10 @@
 
 import {CommonModule} from '@angular/common';
 import {ApplicationRef, ChangeDetectionStrategy, ChangeDetectorRef, Component, ComponentFactoryResolver, ComponentRef, Directive, DoCheck, EmbeddedViewRef, ErrorHandler, Input, NgModule, OnInit, QueryList, TemplateRef, Type, ViewChild, ViewChildren, ViewContainerRef} from '@angular/core';
-import {AfterContentChecked, AfterViewChecked} from '@angular/core/src/core';
+import {AfterViewChecked} from '@angular/core/src/core';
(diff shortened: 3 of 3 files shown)


packages/core/test/acceptance/integration_spec.ts
```diff
@@ -1973,4 +1972,103 @@ describe('acceptance integration tests', () => {
...
+            selector: 'on-push-comp',
+            changeDetection: ChangeDetectionStrategy.OnPush,
+            template: `<p>{{text}}</p>`,
+          })
+          class OnPushComp {
+            text = 'initial';
+
+            constructor(private _cdRef: ChangeDetectorRef){}
+
+                [hookName]() {
```
devversion: FWIW: The weird formatting comes from clang-format.

packages/core/src/render3/instructions/shared.ts
```diff
@@ -475,7 +475,15 @@ export function refreshView<T>(
...
       tView.firstUpdatePass = false;
     }
-    lView[FLAGS] &= ~(LViewFlags.Dirty | LViewFlags.FirstLViewPass);
+    lView[FLAGS] &= ~LViewFlags.FirstLViewPass;
+
+    // Do not reset the dirty state when running in check no changes mode. Resetting would
+    // result in inconsistent and unpredictable behavior when check no changes is disabled.
+    // View Engine did not reset the dirty state in check no changes mode either.
+    if (!checkNoChangesMode) {
+      lView[FLAGS] &= ~LViewFlags.Dirty;
```
devversion: I was unsure if we would want to do the same for `FirstLViewPass` flag unset, but I decided to keep that as is.

packages/core/src/render3/instructions/shared.ts
```diff
@@ -475,7 +475,15 @@ export function refreshView<T>(
     if (tView.firstUpdatePass === true) {
       tView.firstUpdatePass = false;
     }
-    lView[FLAGS] &= ~(LViewFlags.Dirty | LViewFlags.FirstLViewPass);
+    lView[FLAGS] &= ~LViewFlags.FirstLViewPass;
+
+    // Do not reset the dirty state when running in check no changes mode. Resetting would
+    // result in inconsistent and unpredictable behavior when check no changes is disabled.
+    // View Engine did not reset the dirty state in check no changes mode either.
```
pkozlowski-opensource: I would drop the "View Engine did not reset the dirty state in check no changes mode either." from the documentation comment (and maybe move it to the commit message) - as the time goes by we will delete VE and those comments won't be relevant any more.

packages/core/src/render3/instructions/shared.ts
```diff
@@ -475,7 +475,15 @@ export function refreshView<T>(
     if (tView.firstUpdatePass === true) {
       tView.firstUpdatePass = false;
     }
-    lView[FLAGS] &= ~(LViewFlags.Dirty | LViewFlags.FirstLViewPass);
+    lView[FLAGS] &= ~LViewFlags.FirstLViewPass;
+
+    // Do not reset the dirty state when running in check no changes mode. Resetting would
+    // result in inconsistent and unpredictable behavior when check no changes is disabled.
```
pkozlowski-opensource: nit: when I read this comment it sounds like "don't touch or bad things will happen!". While it serves as a warning I think it would be better to express it as follows: we are doing things this way since we want the following scenario to work correctly (describe scenario here).

Can you think of re-phrasing this comment so we've got the exact scenario here?

packages/core/src/render3/instructions/shared.ts
pkozlowski-opensource: I would rather move both flags setting into this if - otherwise we have to read / write to the  `lView[FLAGS]` multiple times (and this is expensive).

pkozlowski-opensource (changes requested):
@devversion the change LGTM overall but I would like to move flags resetting to one place (perf).

packages/core/test/acceptance/integration_spec.ts
```diff
@@ -1973,4 +1972,103 @@ describe('acceptance integration tests', () => {
...
+
+  describe('OnPush markForCheck in lifecycle hooks', () => {
+    describe('with check no changes enabled', () => createOnPushMarkForCheckTests(true));
+    describe('with check no changes disabled', () => createOnPushMarkForCheckTests(false));
+
+    function createOnPushMarkForCheckTests(checkNoChanges: boolean) {
+      const detectChanges = (f: ComponentFixture<any>) => f.detectChanges(checkNoChanges);
+
+      // 1. ngAfterViewInit and ngAfterViewChecked lifecycle hooks run after "OnPushComp" has
+      //    been refreshed. They can leave the component as dirty. Meaning that the "OnPushComp"
```
kara:
```suggestion
      //    been refreshed. They can mark the component as dirty. Meaning that the "OnPushComp"
```

packages/core/test/acceptance/integration_spec.ts
```diff
@@ -1973,4 +1972,103 @@ describe('acceptance integration tests', () => {
...
+      //    can be checked/refreshed in a subsequent change detection cycle.
+      // 2. ngDoCheck and ngAfterContentChecked lifecycle hooks run before "OnPushComp" is
+      //    refreshed. This means that those hooks cannot leave the component as dirty because
```
kara:
```suggestion
      //    refreshed. This means that those hooks cannot mark the component as dirty because
```

kara (approved):
LGTM once my comments and Pawel's are addressed

packages/core/src/render3/instructions/shared.ts
devversion: Done. good idea with providing a concrete scenario.

devversion:
@kara @pkozlowski-opensource Done. Thanks for reviewing! Yeah, I was thinking about combining both flag resets for perf, but wasn't sure if it semantically has any impact. I combined them now.

AndrewKushnir:
[Presubmit](http://test/OCL:289271760:BASE:289271763:1578782848973:ff60298d)

pkozlowski-opensource (approved):
LGTM, thnx for addressing all the comments @devversion !

PR #34498, opened 2019-12-19
docs(ivy): minor changes to version 9 guide

kara:
The main change here was to remove the updating instructions for
version 9 and instead point to update.angular.io. This ensures
we only have one "source of truth" for update instructions.

This commit also includes updates to error message text to
keep them up-to-date with live error messages (and thus keep
them searchable).


IgorMinar (approved):
thanks for the cleanup!

PR #34502, opened 2019-12-19
fix(forms): disabled input acceptance member not applied properly

devversion:
With 5cecd97493025cd940c9ade4ea9f1836d5b05cc8 we intended to expand the input type of the `disabled` input
of the `NgModel` directive. Read more about the reason for this in the actual commit
message.

Currently though, the acceptance coercion member does not have any effect. This
is because the acceptance member needs to refer to the actual input property name,
and not to the public input name. `disabled` corresponds to the `isDisabled` property


alxhub (approved):

IgorMinar (approved):
lgtm!

IgorMinar:

I kicked off a presubmit at http://test/OCL:288730916:BASE:288730959:1578509637885:78b7faa4

PR #34505, opened 2019-12-20
feat(ivy): improve `ExpressionChangedAfterChecked` error message for attributes

AndrewKushnir:
This commit improves `ExpressionChangedAfterChecked` error message for attributes by including attribute name and the content of the entire expression that contains interpolation(s). In order to achieve that, metadata is now stored in `TData` array when `attribute` and `attributeInterpolate` instructions are being called (similar to `property` and `propertyInterpolate` instructions).

## PR Type

- [x] Bugfix

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/render3/instructions/attribute.ts
+++ b/packages/core/src/render3/instructions/attribute.ts
@@ -7,9 +7,10 @@
  */
 import {bindingUpdated} from '../bindings';
 import {SanitizerFn} from '../interfaces/sanitization';
+import {TVIEW} from '../interfaces/view';
 import {getLView, getSelectedIndex, nextBindingIndex} from '../state';
 
-import {elementAttributeInternal} from './shared';
+import {elementAttributeInternal, storePropertyBindingMetadata} from './shared';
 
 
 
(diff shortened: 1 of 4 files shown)


packages/core/src/render3/instructions/attribute.ts
```diff
@@ -7,9 +7,10 @@
  */
 import {bindingUpdated} from '../bindings';
 import {SanitizerFn} from '../interfaces/sanitization';
+import {TVIEW} from '../interfaces/view';
 import {getLView, getSelectedIndex, nextBindingIndex} from '../state';
 
-import {elementAttributeInternal} from './shared';
+import {elementAttributeInternal, storePropertyBindingMetadata} from './shared';
```
pkozlowski-opensource: nit: what do you think of renaming `storePropertyBindingMetadata` to `storeBindingMetadata` (since now it is applied to both property bindings, attributes, interpolations etc.)?

pkozlowski-opensource (approved):
LGTM, left one nit comment to be resolved at your discretion

pkozlowski-opensource:
Oh, I can see that you've changed the `storePropertyBindingMetadata` name in https://github.com/angular/angular/pull/34520/ so we can land this one. Thnx for doing the rename!

AndrewKushnir:
[Presubmit](http://test/OCL:288929683:BASE:288929723:1578596225918:ec33b24c)

AndrewKushnir:
Thanks for the review @pkozlowski-opensource! 👍

PR #34512, opened 2019-12-20
fix(ngcc): fix UMD inline exports and re-exports

gkalpak:
Several fixes regarding exports in UMD.
See individual commits for more info.

### DIFF
--- a/packages/compiler-cli/ngcc/src/host/umd_host.ts
+++ b/packages/compiler-cli/ngcc/src/host/umd_host.ts
@@ -12,16 +12,20 @@ import {absoluteFrom} from '../../../src/ngtsc/file_system';
 import {Declaration, Import} from '../../../src/ngtsc/reflection';
 import {Logger} from '../logging/logger';
 import {BundleProgram} from '../packages/bundle_program';
-import {getOrDefault, stripExtension} from '../utils';
+import {FactoryMap, stripExtension} from '../utils';
 import {ExportDeclaration, ExportStatement, ReexportStatement, findNamespaceOfIdentifier, findRequireCallReference, isExportStatement, isReexportStatement, isRequireCall} from './commonjs_umd_utils';
 import {Esm5ReflectionHost, stripParentheses} from './esm5_host';
 
 export class UmdReflectionHost extends Esm5ReflectionHost {
-  protected umdModules = new Map<ts.SourceFile, UmdModule|null>();
-  protected umdExports = new Map<ts.SourceFile, Map<string, Declaration>|null>();
-  protected umdImportPaths = new Map<ts.ParameterDeclaration, string|null>();
+  protected umdModules =
+      new FactoryMap<ts.SourceFile, UmdModule|null>(sf => this.computeUmdModule(sf));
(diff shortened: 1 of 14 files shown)


packages/compiler-cli/ngcc/src/host/umd_host.ts
```diff
@@ -54,47 +55,42 @@ export class UmdReflectionHost extends Esm5ReflectionHost {
...
+      if (imports === null) {
+        return null;
+      }
 
-    for (const i of imports) {
-      this.umdImportPaths.set(i.parameter, i.path);
-      if (i.parameter === importParameter) {
-        return i.path;
+      for (const i of imports) {
+        this.umdImportPaths.set(i.parameter, i.path);
```
petebacondarwin: Arguably this line is no longer needed since the `getOrDefault()` function will write the return value of this factory to the map.
That being said, this is a minor optimisation, since it adds additional imports, that precede the desired one, to the map, speeding up future lookups on those imports. But in that case perhaps we should complete the iteration over all the imports and add them all the first time we get here?

packages/compiler-cli/ngcc/src/utils.ts
```diff
@@ -64,6 +64,13 @@ export function findAll<T>(node: ts.Node, test: (node: ts.Node) => node is ts.No
   }
 }
 
+export function getOrDefault<K, V>(map: Map<K, V>, key: K, factory: (key: K) => V): V {
```
petebacondarwin: I know that I originally named this function but I wonder if "default" is the correct term here?
What this function is really doing is memoizing a complex factory result...

Perhaps we could create a MemoizedMap class that looks like:

```ts
class MemoizedMap<K,V> extends Map {
  constructor(private factory: (key: T) => V) {}
  get() {
    if (!this.has(key)) {
      this.set(key, factory(key));
    }
    return this.get(key) !;
  }
}
```

petebacondarwin (approved):
Couple of nits - nice tidying up.

packages/compiler-cli/ngcc/src/host/umd_host.ts
gkalpak: This looked intentional, so I left it as is.
Adding all imports to the map sounds good to me (we will almost certainly use them anyway).

gkalpak: Done.

packages/compiler-cli/ngcc/src/utils.ts
gkalpak: Done.

(Extending `Map` causes errors in transpiled ES5 code, so I used a `Map` internally instead. I also went with `FactoryMap`, because that felt more appropriate.)

gkalpak:
@petebacondarwin, thx for the review. PTAL at the last 3 commits, which are new.

gkalpak:
Rebased on master.

petebacondarwin:
CI does not seem happy...

gkalpak:
Added another commit to fix the CI failure (caused by a cyrcular dependency resolution issue). PTAL

petebacondarwin (approved):

petebacondarwin:
LGTM please add "merge" label when you are ready.

PR #34514, opened 2019-12-20
fix(ivy): ensure eventListeners added outside angular context are not called

atscott:
View Engine debug node does not call event listeners added outside the Angular context. This PR updates Ivy to mimic that behavior (not calling those listeners either when using `debugElement.triggerEventHandler`)

### DIFF
--- a/packages/core/src/debug/debug_node.ts
+++ b/packages/core/src/debug/debug_node.ts
@@ -410,7 +410,7 @@ class DebugElement__POST_R3__ extends DebugNode__POST_R3__ implements DebugEleme
     this.listeners.forEach(listener => {
       if (listener.name === eventName) {
         const callback = listener.callback;
-        callback(eventObj);
+        callback.call(node, eventObj);
         invokedListeners.push(callback);
       }
     });
@@ -419,11 +419,20 @@ class DebugElement__POST_R3__ extends DebugNode__POST_R3__ implements DebugEleme
     // that Zone.js only adds to `EventTarget` in browser environments.
     if (typeof node.eventListeners === 'function') {
       // Note that in Ivy we wrap event listeners with a call to `event.preventDefault` in some
-      // cases. We use `Function` as a special token that gives us access to the actual event
+      // cases. We use '__ngUnwrap__' as a special token that gives us access to the actual event
       // listener.
       node.eventListeners(eventName).forEach((listener: Function) => {
-        const unwrappedListener = listener(Function);
-        return invokedListeners.indexOf(unwrappedListener) === -1 && unwrappedListener(eventObj);
+        // In order to ensure that we can detect the special __ngUnwrap__ token described above, we
+        // use `toString` on the listener and see if it contains the token. We use this approach to
+        // ensure that it still worked with compiled code since it cannot remove or rename string
+        // literals. We also considered using a special function name (i.e. if(listener.name ===
+        // special)) but that was more cumbersome and we were also concerned the compiled code could
+        // strip the name, turning the condition in to ("" === "") and always returning true.
+        if (listener.toString().indexOf('__ngUnwrap__') !== -1) {
+          const unwrappedListener = listener('__ngUnwrap__');
+          return invokedListeners.indexOf(unwrappedListener) === -1 &&
+              unwrappedListener.call(node, eventObj);
+        }
       });
     }
   }
--- a/packages/core/test/debug/debug_node_spec.ts
+++ b/packages/core/test/debug/debug_node_spec.ts
@@ -12,7 +12,7 @@ import {Component, DebugElement, DebugNode, Directive, ElementRef, EmbeddedViewR
 import {NgZone} from '@angular/core/src/zone';
 import {ComponentFixture, TestBed, async} from '@angular/core/testing';
 import {By} from '@angular/platform-browser/src/dom/debug/by';
-import {hasClass} from '@angular/platform-browser/testing/src/browser_util';
+import {createMouseEvent, hasClass} from '@angular/platform-browser/testing/src/browser_util';
 import {expect} from '@angular/platform-browser/testing/src/matchers';
 import {ivyEnabled, onlyInIvy} from '@angular/private/testing';
 
@@ -1252,5 +1252,24 @@ class TestCmptWithPropInterpolation {
                         .createComponent(Wrapper);
     expect(fixture.debugElement.query(e => e.name === 'myComponent')).toBeTruthy();
     expect(fixture.debugElement.query(e => e.name === 'div')).toBeTruthy();
+  });
+
+  it('does not call event listeners added outside angular context', () => {
+    let listenerCalled = false;
+    const eventToTrigger = createMouseEvent('mouseenter');
+    function listener() { listenerCalled = true; }
+    @Component({template: ''})
+    class MyComp {
+      constructor(private readonly zone: NgZone, private readonly element: ElementRef) {}
+      ngOnInit() {
+        this.zone.runOutsideAngular(
+            () => { this.element.nativeElement.addEventListener('mouseenter', listener); });
+      }
+    }
+    const fixture =
+        TestBed.configureTestingModule({declarations: [MyComp]}).createComponent(MyComp);
+    fixture.detectChanges();
+    fixture.debugElement.triggerEventHandler('mouseenter', eventToTrigger);
+    expect(listenerCalled).toBe(false);
   });
 }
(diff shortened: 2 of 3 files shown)


packages/core/src/debug/debug_node.ts
```diff
@@ -418,8 +439,16 @@ class DebugElement__POST_R3__ extends DebugNode__POST_R3__ implements DebugEleme
       // cases. We use `Function` as a special token that gives us access to the actual event
       // listener.
       node.eventListeners(eventName).forEach((listener: Function) => {
-        const unwrappedListener = listener(Function);
-        return invokedListeners.indexOf(unwrappedListener) === -1 && unwrappedListener(eventObj);
+        if (listener.name === getIdentifiableWrappedEventHandlerName()) {
+          const unwrappedListener = listener(Function);
+          return invokedListeners.indexOf(unwrappedListener) === -1 &&
+              unwrappedListener.call(node, eventObj);
```
JiaLiPassion: we use `call(node,...)` here, should we also use `callback.call(node)` in https://github.com/angular/angular/blob/cb94a07c6775d05da86a83b75bc507a4447071a2/packages/core/src/debug/debug_node.ts#L430 ?

packages/core/test/debug/debug_node_spec.ts
```diff
@@ -1215,4 +1215,28 @@ class TestCmptWithPropInterpolation {
...
   });
+
+  it('works with event listeners added outside angular context', () => {
+    const eventToTrigger = new MouseEvent('mouseenter');
+    // Note: This listener is only called if this test is run with karma (and Ivy).
+    // Karma locally: `yarn bazel test //packages/core/test:test_web --config=ivy`
+    // Karma in CI: saucelabs_unit_tests_poc
+    function listener(event: any) {
+      // @ts-ignore: The `this` context should be the node the event was triggered on.
+      expect(this).toBeDefined();
```
JiaLiPassion: should we check `expect(this).toBe(fixture.debugElement)`?

atscott: Yea, that's a better assertion. Thanks for the suggestion.

packages/core/src/debug/debug_node.ts
atscott: From what I can see, the listeners here are retrieved from the `LView` and those would have been registered in the components/component templates. I think the `this` context in those functions would be the component class. I made the change because I think it's more correct but I don't think it ends up having downstream effects.

JiaLiPassion: Thanks, I will try to verify it.

packages/core/src/core_private_export.ts
```diff
@@ -27,6 +28,7 @@ export {_sanitizeHtml as ɵ_sanitizeHtml} from './sanitization/html_sanitizer';
 export {_sanitizeStyle as ɵ_sanitizeStyle} from './sanitization/style_sanitizer';
 export {_sanitizeUrl as ɵ_sanitizeUrl} from './sanitization/url_sanitizer';
 export {global as ɵglobal} from './util/global';
+
```
mhevery:
```suggestion
```

packages/core/test/debug/debug_node_spec.ts
```diff
@@ -1215,4 +1215,28 @@ class TestCmptWithPropInterpolation {
...
   });
+
+  it('works with event listeners added outside angular context', () => {
+    const eventToTrigger = createMouseEvent('mouseenter');
+    // Note: This listener is only called if this test is run with karma (and Ivy).
+    // Karma locally: `yarn bazel test //packages/core/test:test_web --config=ivy`
+    // Karma in CI: saucelabs_unit_tests_poc
+    function listener(event: any) {
+      // @ts-ignore: The `this` context should be the node the event was triggered on.
+      expect(this).toBe(fixture.nativeElement);
```
mhevery: If for some reason the `listener` does not get executed, than the assertions don't get executed either. Could you update the test so that lack of `listener` execution fails the test? One way to do that would be to move the `expect` to the `it` closure and than just leave code here which copies tho values from `listener` to the outside closure (so that it can be asserted)

packages/core/src/debug/debug_node.ts
```diff
@@ -418,8 +439,16 @@ class DebugElement__POST_R3__ extends DebugNode__POST_R3__ implements DebugEleme
       // cases. We use `Function` as a special token that gives us access to the actual event
       // listener.
       node.eventListeners(eventName).forEach((listener: Function) => {
-        const unwrappedListener = listener(Function);
-        return invokedListeners.indexOf(unwrappedListener) === -1 && unwrappedListener(eventObj);
+        if (listener.name === getIdentifiableWrappedEventHandlerName()) {
```
mhevery: I don't think this is safe. When minifier runs it strips the function names. So you will just be checking that `"" === ""` which is not a good check.

packages/platform-browser/src/dom/dom_renderer.ts
```diff
@@ -50,7 +50,11 @@ export function flattenStyles(
 }
 
 function decoratePreventDefault(eventHandler: Function): Function {
-  return (event: any) => {
+  // `DebugNode.triggerEventHandler` needs to know if the listener was created with
+  // decoratePreventDefault or is a listener added outside the Angular context so it can handle the
+  // two differently. In the first case, the special `Function` token is passed to the unwrap the
+  // listener (see below).
+  return ɵcreateIdentifiableWrappedEventHandler((event: any) => {
```
mhevery: does this creates additional wrapper? If so what is the perf impact of that?

atscott: Yes, this is an additional wrapper. Do we have performance tests that would cover this and how do I run them? I could also conditionally wrap this based on NG_DEV_MODE since we only need to identify these functions for debug_node.

alxhub: Doing it conditionally seems reasonable.

alxhub (commented):
This looks good to me as a solution to the problem we worked on.

I'll defer to Misko, et al. for runtime approval though.

alxhub (approved):

atscott:
[presubmit](https://test.corp.google.com/ui#id=OCL:290307856:BASE:290307883:1579289109700:cfeeb5eb)

packages/core/src/debug/debug_node.ts
```diff
@@ -415,11 +415,14 @@ class DebugElement__POST_R3__ extends DebugNode__POST_R3__ implements DebugEleme
     // that Zone.js only adds to `EventTarget` in browser environments.
     if (typeof node.eventListeners === 'function') {
       // Note that in Ivy we wrap event listeners with a call to `event.preventDefault` in some
-      // cases. We use `Function` as a special token that gives us access to the actual event
+      // cases. We use '__ngUnwrap__' as a special token that gives us access to the actual event
       // listener.
       node.eventListeners(eventName).forEach((listener: Function) => {
-        const unwrappedListener = listener(Function);
-        return invokedListeners.indexOf(unwrappedListener) === -1 && unwrappedListener(eventObj);
+        if (listener.toString().indexOf('__ngUnwrap__') !== -1) {
```
mhevery: Could you add a text above explaining how this logic wars. Why `listener.toSttring()` should cantain `__ngUnwrap__` etc..

packages/platform-browser/src/dom/dom_renderer.ts
```diff
@@ -50,10 +50,15 @@ export function flattenStyles(
 }
 
 function decoratePreventDefault(eventHandler: Function): Function {
-  return (event: any) => {
-    // Ivy uses `Function` as a special token that allows us to unwrap the function
+  // `DebugNode.triggerEventHandler` needs to know if the listener was created with
+  // decoratePreventDefault or is a listener added outside the Angular context so it can handle the
+  // two differently. In the first case, the special '__ngUnwrap__' token is passed to the unwrap
+  // the
```
mhevery: fix line break

packages/platform-browser/src/dom/dom_renderer.ts
```diff
@@ -50,10 +50,15 @@ export function flattenStyles(
...
+  // listener (see below).
+  const wrappedEventHandler = (event: any) => {
+    // Ivy uses '__ngUnwrap__' as a special token that allows us to unwrap the function
```
mhevery: add text here how this is meant to be `toString()` and why it will survive minification.

mhevery (changes requested):

AndrewKushnir:
@atscott FYI, it looks like size checks are failing in `test_aio_local` CircleCI job, so I'm adding "cleanup" label for now.

PR #34515, opened 2019-12-20
feat(language-service): Append symbol type to hover tooltip

kyliau:
Now that https://github.com/angular/angular/pull/34177 fixed the `TypeWrapper`
to have a proper name, we have the information needed to show the type
name in a hover tooltip.

Screen Shot 2019-12-20 at 10 32 22 AM

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/language-service/test/hover_spec.ts
+++ b/packages/language-service/test/hover_spec.ts
@@ -36,7 +36,7 @@ describe('hover', () => {
     expect(quickInfo).toBeTruthy();
     const {textSpan, displayParts} = quickInfo !;
     expect(textSpan).toEqual(marker);
-    expect(toText(displayParts)).toBe('(property) MyComponent.name');
+    expect(toText(displayParts)).toBe('(property) MyComponent.name: string');
   });
 
   it('should be able to find a field in a attribute reference', () => {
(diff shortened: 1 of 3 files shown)


packages/language-service/test/hover_spec.ts
```diff
@@ -113,7 +113,7 @@ describe('hover', () => {
     expect(quickInfo).toBeTruthy();
     const {textSpan, displayParts} = quickInfo !;
     expect(textSpan).toEqual(marker);
-    expect(toText(displayParts)).toBe('(directive) StringModel');
+    expect(toText(displayParts)).toBe('(directive) StringModel: typeof StringModel');
```
ayazhafiz: hm.. this seems a bit unexpected, no? This suggests `StringModel`'s type is registered as the typeof the value and not the `StringModel` class. Do you know why they would be different?

kyliau: This surprised me too when I first saw it. 
It is actually correct, because here we are pointing to the class itself (specifically, the _constructor function_), and not to an **instance** of the class.
This [section](https://www.typescriptlang.org/docs/handbook/classes.html#constructor-functions) in the TypeScript handbook does a pretty good job on explaining the difference between the two.

ayazhafiz: I see. I guess my confusion was because classes create both a type and a value, so yes here we do expect to get the type of the class value rather than the type of the class instance, as you mention.

ayazhafiz (approved):

PR #34521, opened 2019-12-20
misc fixes that improve the build system and the code base

IgorMinar:
plz review individual commits

### DIFF
--- a/packages/core/src/render3/instructions/shared.ts
+++ b/packages/core/src/render3/instructions/shared.ts
@@ -49,7 +49,7 @@ import {LCleanup, LViewBlueprint, MatchesArray, TCleanup, TNodeConstructor, TNod
  * A permanent marker promise which signifies that the current CD tree is
  * clean.
  */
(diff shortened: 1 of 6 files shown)


packages/core/src/render3/instructions/shared.ts
```diff
@@ -1702,8 +1702,8 @@ export function scheduleTick(rootContext: RootContext, flags: RootContextFlags)
   rootContext.flags |= flags;
 
   if (nothingScheduled && rootContext.clean == _CLEAN_PROMISE) {
-    let res: null|((val: null) => void);
-    rootContext.clean = new Promise<null>((r) => res = r);
+    let res: (() => void);
```
gkalpak:
```suggestion
    let res: () => void;
```

jessicajaniuk:
Closing this as it's out of date and hasn't seen updates since February 2020, but we can re-open if necessary to land any changes in here after updates.

PR #34524, opened 2019-12-21
fix(ivy): warn instead of throwing for unknown elements

crisbeto:
Follow-up from [this discussion](https://github.com/angular/angular/pull/33419#discussion_r339296216). In Ivy we don't use the schema to validate tag names, but instead we use feature detection to figure out whether an element is supported. While this should generally be more accurate, it'll also end up throwing for some more innocent cases. E.g. now Ivy throws an error for `main` elements in IE which is accurate since IE doesn't support the element, but is annoying since there is no functionality attached.

These changes switch to logging a warning instead, similarly to what we're doing for unknown properties.

### DIFF
--- a/packages/core/test/acceptance/ng_module_spec.ts
+++ b/packages/core/test/acceptance/ng_module_spec.ts
@@ -166,36 +166,69 @@ describe('NgModule', () => {
       }).not.toThrow();
     });
 
-    it('should throw unknown element error without CUSTOM_ELEMENTS_SCHEMA for element with dash in tag name',
-       () => {
-         @Component({template: `<custom-el></custom-el>`})
-         class MyComp {
-         }
+    onlyInIvy('unknown element check logs a warning rather than throwing')
+        .it('should warn about unknown element without CUSTOM_ELEMENTS_SCHEMA for element with dash in tag name',
+            () => {
 
-         TestBed.configureTestingModule({declarations: [MyComp]});
+              @Component({template: `<custom-el></custom-el>`})
+              class MyComp {
+              }
 
-         expect(() => {
-           const fixture = TestBed.createComponent(MyComp);
-           fixture.detectChanges();
-         }).toThrowError(/'custom-el' is not a known element/);
-       });
+              const spy = spyOn(console, 'warn');
+              TestBed.configureTestingModule({declarations: [MyComp]});
+              const fixture = TestBed.createComponent(MyComp);
+              fixture.detectChanges();
+              expect(spy.calls.mostRecent().args[0]).toMatch(/'custom-el' is not a known element/);
+            });
 
-    it('should throw unknown element error without CUSTOM_ELEMENTS_SCHEMA for element without dash in tag name',
-       () => {
-         @Component({template: `<custom></custom>`})
-         class MyComp {
-         }
+    modifiedInIvy('unknown element error thrown instead of warning')
+        .it('should throw unknown element error without CUSTOM_ELEMENTS_SCHEMA for element with dash in tag name',
+            () => {
+              @Component({template: `<custom-el></custom-el>`})
+              class MyComp {
+              }
 
-         TestBed.configureTestingModule({declarations: [MyComp]});
+              TestBed.configureTestingModule({declarations: [MyComp]});
 
-         expect(() => {
-           const fixture = TestBed.createComponent(MyComp);
-           fixture.detectChanges();
-         }).toThrowError(/'custom' is not a known element/);
-       });
+              expect(() => {
+                const fixture = TestBed.createComponent(MyComp);
+                fixture.detectChanges();
+              }).toThrowError(/'custom-el' is not a known element/);
+            });
+
+    onlyInIvy('unknown element check logs a warning rather than throwing')
+        .it('should warn about unknown element without CUSTOM_ELEMENTS_SCHEMA for element without dash in tag name',
+            () => {
+              @Component({template: `<custom></custom>`})
+              class MyComp {
+              }
+
+              const spy = spyOn(console, 'warn');
+              TestBed.configureTestingModule({declarations: [MyComp]});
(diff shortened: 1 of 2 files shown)


packages/core/test/acceptance/ng_module_spec.ts
```diff
@@ -166,36 +166,69 @@ describe('NgModule', () => {
...
-         }
-
-         TestBed.configureTestingModule({declarations: [MyComp]});
-
-         expect(() => {
-           const fixture = TestBed.createComponent(MyComp);
-           fixture.detectChanges();
-         }).toThrowError(/'custom' is not a known element/);
-       });
+    onlyInIvy('unknown element check logs a warning rather an throwing')
```
kara:
```suggestion
    onlyInIvy('unknown element check logs a warning rather than throwing')
```

packages/core/test/acceptance/ng_module_spec.ts
```diff
@@ -166,36 +166,69 @@ describe('NgModule', () => {
...
+              }
+
+              const spy = spyOn(console, 'warn');
+              TestBed.configureTestingModule({declarations: [MyComp]});
+              const fixture = TestBed.createComponent(MyComp);
+              fixture.detectChanges();
+              expect(spy.calls.mostRecent().args[0]).toMatch(/'custom-el' is not a known element/);
+            });
+
+    modifiedInIvy('unknown element error thrown instead warning')
```
kara: Same below

```suggestion
    modifiedInIvy('unknown element error thrown instead of warning')
```

packages/core/test/acceptance/ng_module_spec.ts
```diff
@@ -166,36 +166,69 @@ describe('NgModule', () => {
...
+              expect(() => {
+                const fixture = TestBed.createComponent(MyComp);
+                fixture.detectChanges();
+              }).toThrowError(/'custom' is not a known element/);
+            });
 
     onlyInIvy('test relies on Ivy-specific AOT format')
-        .it('should not throw unknown element error for AOT-compiled components', () => {
+        .it('should not log unknown element warning for AOT-compiled components', () => {
+          const spy = spyOn(console, 'warn');
```
kara: I think we lost the `modifiedInIvy` version of this test (unless I'm missing it)? Let's add one

packages/core/test/acceptance/ng_module_spec.ts
```diff
@@ -253,88 +286,173 @@ describe('NgModule', () => {
             imports: [MyModule],
           });
 
+          const fixture = TestBed.createComponent(MyComp);
+          fixture.detectChanges();
+          expect(spy).not.toHaveBeenCalled();
+        });
+
+    onlyInIvy('unknown element check logs a warning rather than throwing')
+        .it('should not throw unknown element error with CUSTOM_ELEMENTS_SCHEMA', () => {
```
kara:
```suggestion
        .it('should not warn about unknown elements with CUSTOM_ELEMENTS_SCHEMA', () => {
```

packages/core/test/acceptance/ng_module_spec.ts
```diff
@@ -253,88 +286,173 @@ describe('NgModule', () => {
...
             fixture.detectChanges();
           }).not.toThrow();
         });
 
-    it('should not throw unknown element error with CUSTOM_ELEMENTS_SCHEMA', () => {
-      @Component({template: `<custom-el></custom-el>`})
-      class MyComp {
-      }
+    onlyInIvy('unknown element check logs a warning rather than throwing')
+        .it('should not throw unknown element error with NO_ERRORS_SCHEMA', () => {
```
kara:
```suggestion
        .it('should not warn about unknown elements with NO_ERRORS_SCHEMA', () => {
```

packages/core/test/acceptance/ng_module_spec.ts
```diff
@@ -253,88 +286,173 @@ describe('NgModule', () => {
...
 
-    it('should not throw unknown element error if element matches a directive', () => {
-      @Component({
-        selector: 'custom-el',
-        template: '',
-      })
-      class CustomEl {
-      }
+    onlyInIvy('unknown element check logs a warning rather than throwing')
+        .it('should not throw unknown element error if element matches a directive', () => {
```
kara:
```suggestion
        .it('should not warn about unknown elements if element matches a directive', () => {
```

packages/core/test/acceptance/ng_module_spec.ts
```diff
@@ -253,88 +286,173 @@ describe('NgModule', () => {
...
+          TestBed.configureTestingModule({declarations: [MyComp, CustomEl]});
+
+          expect(() => {
+            const fixture = TestBed.createComponent(MyComp);
+            fixture.detectChanges();
+          }).not.toThrow();
+        });
+
+    onlyInIvy('unknown element check logs a warning rather than throwing')
+        .it('should not throw for HTML elements inside an SVG foreignObject', () => {
```
kara:
```suggestion
        .it('should not warn for HTML elements inside an SVG foreignObject', () => {
```

kara (changes requested):
Mostly looks good, just need to update the Ivy-only test descriptions

crisbeto:

AndrewKushnir:
[Presubmit](http://test/OCL:288523430:BASE:288523461:1578421553445:c090ba54)

kara (approved):
LGTM

kara:
Caretaker note: This PR requires a patch [here](https://cs.corp.google.com/piper///depot/google3/javascript/angular2/testing/catalyst/integration_test.ts?l=144&ws=tap-prod-presubmit/58690893&snapshot=2) because the project is explicitly testing that the framework throws an error.

AndrewKushnir:
Thanks @kara. I've created a [patch](http://cl/288543848) that disables that test temporary. The patch can be submitted prior to landing this PR.

PR #34527, opened 2019-12-21
fix(ngcc): recognize re-exports with imported TS helpers in CommonJS and UMD

gkalpak:
~~_Builds on top of #34512 (so the first 7 commits are from there)._~~
_Rebased on master now that #34512 has been merged._

##
Previously, the `CommonJsReflectionHost` and `UmdReflectionHost` would only recognize re-exports of the form `__export(...)`. This is what re-exports look like, when the TypeScript helpers are emitted inline (i.e. when compiling with the default [TypeScript compiler options][1] that include `noEmitHelpers: false` and `importHelpers: false`).

However, when compiling with `importHelpers: true` and [tslib][2] (which is the recommended way for optimized bundles), the re-exports will look like: `tslib_1.__exportStar(..., exports)`
These types of re-exports were previously not recognized by the CommonJS/UMD `ReflectionHost`s and thus ignored.

This commit fixes this by ensuring both re-export formats are recognized.

[1]: https://www.typescriptlang.org/docs/handbook/compiler-options.html
[2]: https://www.npmjs.com/package/tslib

### DIFF
--- a/packages/compiler-cli/ngcc/test/host/commonjs_host_spec.ts
+++ b/packages/compiler-cli/ngcc/test/host/commonjs_host_spec.ts
@@ -118,7 +118,7 @@ exports.SomeDirective = SomeDirective;
         contents: `
         var core = require('@angular/core');
         var CtorDecoratedAsArray = (function() {
-        function CtorDecoratedAsArray(arg1) {
+          function CtorDecoratedAsArray(arg1) {
           }
           CtorDecoratedAsArray.ctorParameters = [{ type: ParamType, decorators: [{ type: Inject },] }];
           return CtorDecoratedAsArray;
@@ -511,6 +511,7 @@ var c = file_a.a;
           var b_module = require('./b_module');
           var xtra_module = require('./xtra_module');
           var wildcard_reexports = require('./wildcard_reexports');
+          var wildcard_reexports_imported_helpers = require('./wildcard_reexports_imported_helpers');
           `
         },
         {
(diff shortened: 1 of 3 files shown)


packages/compiler-cli/ngcc/test/host/commonjs_host_spec.ts
```diff
@@ -554,13 +555,22 @@ exports.xtra2 = xtra2;
         {
           name: _('/wildcard_reexports.js'),
```
petebacondarwin: Perhaps rename this file to be `wildcard_reexports_inline_helpers.js`?

packages/compiler-cli/ngcc/test/host/commonjs_host_spec.ts
```diff
@@ -1763,7 +1773,7 @@ exports.ExternalModule = ExternalModule;
               ]);
         });
 
-        it('should handle wildcard re-exports of other modules', () => {
+        it('should handle wildcard re-exports of other modules (with emitted helpers)', () => {
```
petebacondarwin: NIT: emitted -> inline

(Unless I am wrong about the convention for naming these.)

petebacondarwin (approved):
LGTM (for last commit) - couple of minor nits.

packages/compiler-cli/ngcc/test/host/commonjs_host_spec.ts
gkalpak: Both `emitted` and `inline` coule be used (the helpers are emitted inline by TS). I'll keep `emitted` to avoid confusion with `inline exports` (which is something different).

packages/compiler-cli/ngcc/test/host/commonjs_host_spec.ts
gkalpak: :+1: I'll go with `*_emitted_helpers` (see comment below).

gkalpak:
Thx for the review, @petebacondarwin! Rebased on master and addressed feedback.
Marking this as blocked on #34512 (to ensure they are merged in order).

gkalpak:
Rebased on master.

petebacondarwin:
CI?

gkalpak:
#34512 has been merged. Rebased on master and marked for merging.

PR #34528, opened 2019-12-21
fix(ngcc): correctly detect dependencies in CommonJS

gkalpak:
~~_Builds on top of #34527 (so the first 9 commits are from there)._~~
Rebased on master now that #34527 has been merged.

##
Previously, `CommonJsDependencyHost.findDependencies()` would only find dependencies via imports of the form `var foo = require('...');` or `var foo = require('...'), bar = require('...');` However, CommonJS files can have imports in many different forms. By failing to recognize other forms of imports, the associated dependencies were missed, which in turn resulted in entry-points being compiled out-of order and failing due to that.

While we cannot easily capture all different types of imports, this commit enhances `CommonJsDependencyHost` to recognize the following common forms of imports:

- Imports in property assignments. E.g.:
  `exports.foo = require('...');` or
  `module.exports = {foo: require('...')};`

- Imports for side-effects only. E.g.:
  `require('...');`

- Star re-exports (with both emitted and imported heleprs). E.g.:
  `__export(require('...'));` or
  `tslib_1.__exportStar(require('...'), exports);`

### DIFF
--- a/packages/compiler-cli/ngcc/src/dependencies/commonjs_dependency_host.ts
+++ b/packages/compiler-cli/ngcc/src/dependencies/commonjs_dependency_host.ts
@@ -7,7 +7,7 @@
  */
 import * as ts from 'typescript';
 import {AbsoluteFsPath} from '../../../src/ngtsc/file_system';
-import {isRequireCall} from '../host/commonjs_umd_utils';
+import {RequireCall, isReexportStatement, isRequireCall} from '../host/commonjs_umd_utils';
 import {DependencyHostBase} from './dependency_host';
 import {ResolvedDeepImport, ResolvedRelativeModule} from './module_resolver';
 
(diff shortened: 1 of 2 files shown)


petebacondarwin (approved):
LTGM (for the last two commits)

Do we need to support side-effect only imports in UMD too?

gkalpak:
(Rebased on master.)

> Do we need to support side-effect only imports in UMD too?

Theoretically, I believe we should. I originally set out to fix that in UMD too, but then I realized we currently [only detect dependencies][1] if they are passed as arguments to the factory function.

So, we would miss dependencies altogether in some UMD variations (e.g. only passing `require` to factory and importing dependencies via `require()` calls). Since it is not a trivial change to cover all possible UMD variations, I think it might make sense to wait until we have an actual failure.

[1]: https://github.com/angular/angular/blob/9b9116c79/packages/compiler-cli/ngcc/src/host/umd_host.ts#L242-L252

gkalpak:
Rebased on master.

petebacondarwin:
Similar CI issues?

PR #34353, opened 2019-12-11
fix(ngcc): render UMD imports even if no prior imports

petebacondarwin:
Previously the UMD rendering formatter assumed that
there would already be import (and an export) arguments
to the UMD factory function.

This commit adds support for this corner case.

Fixes #34138

PR #34545, opened 2019-12-23
fix(ngcc): do not add trailing commas in UMD imports

### LINKED ISSUE #34525: ngcc: processed umd bundles fail with syntax errors in IE11
# 🐞 bug report

### Affected Package

@angular/compiler-cli - ngcc

### Is this a regression?

Yes, this doesn't seem to happen in Angular v9.0.0-rc.6

### Description

Processed UMD bundles sometimes seem to have superfluous comma tokens that break evaluation of the file in IE11.

```js
// First broken code                                           ↓  
... factory(exports, require('rxjs'), require('rxjs/operators'),) :

// Second superfluous comma                                     ↓  
... define('@angular/core', ['exports', 'rxjs', 'rxjs/operators',]

// Third broken code                       ↓  
}(this, (function (exports, rxjs, operators,) { 'use strict';
```

Note: This all seems to happen in the canonical AMD module header. It seems to happen three times per bundle. Basically every time the dependencies of the AMD module are specified.

## 🔬 Minimal Reproduction

https://github.com/devversion/ie11-ngcc-compat

## 🔥 Exception or Error
```
SCRIPT1002: Syntax Error
core.umd.js (8,128)
```

## 🌍  Your Environment

**Angular Version:**

Angular v9.0.0-rc.7

gkalpak:
Previously, if `UmdRenderingFormatter#addImports()` was called with an empty list of imports to add (i.e. no new imports were needed), it would add trailing commas in several locations (arrays, function arguments, function parameters), thus making the code imcompatible with legacy browsers such as IE11.

This commit fixes it by ensuring that no trailing commas are added if `addImports()` is called with an empty list of imports.
This is a follow-up to #34353.

Fixes #34525.

### DIFF
--- a/packages/compiler-cli/ngcc/src/rendering/commonjs_rendering_formatter.ts
+++ b/packages/compiler-cli/ngcc/src/rendering/commonjs_rendering_formatter.ts
@@ -30,6 +30,11 @@ export class CommonJsRenderingFormatter extends Esm5RenderingFormatter {
    *  Add the imports below any in situ imports as `require` calls.
    */
   addImports(output: MagicString, imports: Import[], file: ts.SourceFile): void {
+    // Avoid unnecessary work if there are no imports to add.
+    if (imports.length === 0) {
+      return;
+    }
+
     const insertionPoint = this.findEndOfImports(file);
     const renderedImports =
         imports.map(i => `var ${i.qualifier} = require('${i.specifier}');\n`).join('');
--- a/integration/ngcc/test.sh
+++ b/integration/ngcc/test.sh
@@ -125,6 +125,21 @@ assertSucceeded "Expected 'ngcc' to log 'Compiling'."
   assertSucceeded "Expected '@angular/cdk/a11y' (umd) to actually have decorators via static properties."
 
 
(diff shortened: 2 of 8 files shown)


packages/compiler-cli/ngcc/src/rendering/commonjs_rendering_formatter.ts
```diff
@@ -30,6 +30,11 @@ export class CommonJsRenderingFormatter extends Esm5RenderingFormatter {
    *  Add the imports below any in situ imports as `require` calls.
    */
   addImports(output: MagicString, imports: Import[], file: ts.SourceFile): void {
+    // Avoid unnecessary work if there are no imports to add.
```
petebacondarwin: NIT: this comment is not really necessary IMO.

petebacondarwin (approved):

packages/compiler-cli/ngcc/src/rendering/commonjs_rendering_formatter.ts
gkalpak: Removed.

PR #34550, opened 2019-12-23
build: cleanup and unify tsconfigs used across the repo

IgorMinar:
The tsconfig-defaults.json is the main tsconfig from which all the other tsconfig will extend.

This allows us to have just a single place where we set common flags, like "strict", etc.


jessicajaniuk:
Closing this since it's seen no updates for 2 years and is out of date. We can re-open and land if the change is still valid.

PR #34553, opened 2019-12-24
docs: set default file property in stackblitz.json

### LINKED ISSUE #22357: Docs: Link to stackblitz examples with filename on aio
## I'm submitting a...

<pre><code>
[X] Documentation feature request
</code></pre>

## Current behavior
When visiting any page with a stackblitz link such as https://angular.io/guide/router there is a link at the top to open a live example like this one: https://angular.io/generated/live-examples/router/stackblitz.html

This also opens main.ts in the editor.

## Expected behavior
It would be nice if the examples linked directly to a more relevant file from the example, such as app-routing.module.ts (https://stackblitz.com/angular/ovkvayplpbm?file=src%2Fapp%2Fapp-routing.module.ts).

This would make it easier to engage with the examples more quickly, and to see what the example is attempting to demonstrate.

sonukapoor:
Previously, when no default file is defined in the `stackblitz.json`, stackblitz
will set it automatically. It used to be `main.ts` and nowadays it seems
to be `app.component.ts`. However, this behaviour can change in the future. To overcome
this, the stackblitz API allows specifying the file to open in the editor by passing
a query param. This PR sets the default file in the `stackblitz.json` through the
`file` property. If still no file is specified, the `builder.js` will try to default to
`app.component.html`, `app.component.ts` or `main.ts`.

Closes #22357

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #22357

- [ ] Yes
- [x] No

### DIFF
--- a/aio/tools/stackblitz-builder/builder.js
+++ b/aio/tools/stackblitz-builder/builder.js
@@ -116,13 +116,26 @@ class StackblitzBuilder {
     }
   }
 
-  _createBaseStackblitzHtml(config) {
-    var file = '';
-
-    // TODO: Doesn't work properly yet
+  _getPrimaryFile(config) {
     if (config.file) {
-      file = `?file=${config.file}`;
+      if (!this._existsSync(path.join(config.basePath, config.file))) {
+        throw new Error(`The specified primary file (${config.file}) does not exist in '${config.basePath}'.`);
+      }
+      return config.file;
+    } else {
+      const defaultPrimaryFiles = ['src/app/app.component.html', 'src/app/app.component.ts', 'src/app/main.ts'];
+      const primaryFile = defaultPrimaryFiles.find(fileName =>  this._existsSync(path.join(config.basePath, fileName)));
+  
+      if (!primaryFile) {
+        throw new Error(`None of the default primary files (${defaultPrimaryFiles.join(', ')}) exists in '${config.basePath}'.`);
+      }
+ 
+      return primaryFile;
     }
+  }
+
+  _createBaseStackblitzHtml(config) {
+    var file = `?file=${this._getPrimaryFile(config)}`;
     var action = `https://run.stackblitz.com/api/angular/v1${file}`;
     var html = `<!DOCTYPE html><html lang="en"><body>
     <form id="mainForm" method="post" action="${action}" target="_self"></form>
--- a/aio/content/examples/accessibility/stackblitz.json
+++ b/aio/content/examples/accessibility/stackblitz.json
@@ -5,6 +5,5 @@
     "!**/*.js",
     "!**/*.[1,2].*"
   ],
-  "file": "src/app/app.component.ts",
   "tags": ["Accessibility"]
 }
--- a/aio/content/examples/animations/stackblitz.json
+++ b/aio/content/examples/animations/stackblitz.json
@@ -5,5 +5,6 @@
     "!**/*.js",
     "!**/*.[1,2,3].*"
   ],
+  "file": "src/app/app.component.ts",
   "tags": ["animations"]
 }
--- a/aio/content/examples/architecture/stackblitz.json
+++ b/aio/content/examples/architecture/stackblitz.json
@@ -5,5 +5,5 @@
     "!**/*.js",
     "!**/*.[1].*"
   ],
-  "file": "src/app/app.module.ts"
+  "file": "src/app/hero-list.component.html"
 }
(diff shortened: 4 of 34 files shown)


StephenFluin (approved):

aio/tools/stackblitz-builder/builder.js
```diff
@@ -119,10 +119,16 @@ class StackblitzBuilder {
...
-    // TODO: Doesn't work properly yet
     if (config.file) {
       file = `?file=${config.file}`;
+    } else {
+      var defaultSelectableFiles = ['src/app/app.component.html', 'src/app/app.component.ts', 'src/app/main.ts'];
+      var selectedFile = defaultSelectableFiles.find(fileName =>  this._existsSync(path.join(config.basePath,fileName)));
+      if (!selectedFile) {
+        throw new Error(`No default file found in ${config.basePath}`);
+      }
     }
```
gkalpak: To make this easier to follow, could you please move the logic to determine the `file` to a separate function?
Also, check the file's existence for `config.file` as well. Something like:

```js
function getPrimaryFile(config) {
  if (config.file) {
    if (!this._existsSync(path.join(config.basePath, config.file)) {
      throw new Error(`The specified primary file (${config.file}) does not exist in '${config.basePath}'.);
    }
    return config.file;
  } else {
    const defaultPrimaryFiles = ['src/app/app.component.html', 'src/app/app.component.ts', 'src/app/main.ts'];
    const primaryFile = defaultPrimaryFiles.find(fileName =>  this._existsSync(path.join(config.basePath, fileName)));

    if (!primaryFile) {
      throw new Error(`None of the default primary files (${defaultPrimaryFiles.join(', ')}) exists in '${config.basePath}'.);
    }

    return primaryFile;
  }
}
```

gkalpak (changes requested):
Thx for working on this ❤️
I did an initial pass:

- Please move the changes to the builder to a commit of their own.
- Was adding an empty `aio/content/examples/docs-style-guide/src/main.2.ts` intentional? If not , please remove.

aio/tools/stackblitz-builder/builder.js
sonukapoor: @gkalpak: Yes, the empty `main.2.ts` is intentional. The reason behind the file is that we are creating angular Stackblitz projects. If we include the default `main.ts`, then we will also need to include every file that is imported in the main.ts - e.g: `environment.ts`, `app.module`. It will also try to bootstrap, the application, which we do not want for the `aio\content\examples\docs-style-guide\second.stackblitz.json`. That file is used to showcase how we generate Stackblitz examples. Hope that makes sense. I have created a secondary commit for the builder as well.

gkalpak (changes requested):
Nice work, @sonukapoor! (Sorry it took a while to review.)
I left a few minor comments. Also, please change the commit message of the first commit to something like:

> build(docs-infra): always specify a landing file for StackBlitz projects
> 
> The StackBlitz API (which we use to generate StackBlitz projects on the
> fly, when a user clicks on a live example link in the docs) allows
> specifying the file to open in the editor by passing a query param. If
> no file is specified, StackBlitz opens a default one. In the past, it
> used to be \`main.ts\` and nowadays it seems to be \`app.component.ts\`.
> 
> [StackBlitz builder][1], the tool that we use to generate the StackBlitz
> projects, allows specifying the primary file by setting the \`file\`
> property in the corresponding \`stackblitz.json\`.
> 
> Previously, if the \`file\` property was not set, StackBlitz builder would
> not specify a file, thus falling back on StackBlitz's default behavior.
> This was not great, because the default behavior may change unexpectedly
> is the future to something that is less useful for our users.
> 
> This commit change StackBlitz builder to always specify a primary file.
> If no file is specified in \`stackblitz.json\`, the builder will look
> for the first file that exists in the example from a list of predefined
> files.
> 
> Partially addresses #22357.
> 
> [1]: https://github.com/angular/angular/blob/9b9116c79/aio/tools/stackblitz-builder/builder.js

...and that of the second commit to something like:

> build(docs-infra): specify more relevant landing files for StackBlitz projects
> 
> The StackBlitz API (which we use to generate StackBlitz projects on the
> fly, when a user clicks on a live example link in the docs) allows
> specifying the file to open in the editor by passing a query param. If
> no file is specified, StackBlitz opens a default one.
> 
> This commit specifies relevant landing files for some of the examples to
> make it easier to engage with the examples more quickly, and to see what
> the example is attempting to demonstrate.
> 
> Fixes #22357.

sonukapoor:
@gkalpak I have made all the requested changes. Can you please re-review?

gkalpak (approved):
Awesome! Thx, @sonukapoor ❤️

gkalpak:
@caretaker: Global approval for the docs-only changes.

PR #34554, opened 2019-12-24
fix(ivy): support SkipSelf when injecting ViewContainerRef, ElementRef and ChangeDetectorRef

### LINKED ISSUE #34066: In Ivy, SkipSelf() is not accessing the parent's ViewContainerRef, ElementRef, etc
# 🐞 bug report

### Affected Package

 The issue is caused by Angular Ivy

### Description

Before Ivy, I was able to access the `ViewContainerRef`, `ElementRef`, etc of the parent directive using the `@SkipSelf` decorator in the child directive. With Ivy integrated in my project, it's no longer skipping its own injections and accessing the ones in its parent. 

For example, I have the following code: 

In app.component.ts
```ts
@Directive({ selector: '[parentDirective]' })
export class ParentDirective {
  constructor(private vc: ViewContainerRef){
    console.log(vc.element);
  }
}

@Directive({ selector: '[childDirective]' })
export class ChildDirective {
  constructor(@SkipSelf() private vc: ViewContainerRef){
    console.log(vc.element);
  }
}
```
in app.component.html

```html
<div parentDirective>parent <span childDirective>child</span></div>
```

These two directive classes should log the same ViewContainerRef element. However, with Ivy, you can see that's no longer the case. Here is the [reproduction](https://ivy.ng-run.com/edit/xMfu7xoF1t4b8hFvoU9N?open=app%2Fapp.component.html). This reproduction is made in Angular Ivy 8.1.0-next.3. But I also see this behavior in my project that's on "@angular/core": "9.0.0-rc.3",

## 🔬 Minimal Reproduction

[Minimal Reproduction](https://ivy.ng-run.com/edit/xMfu7xoF1t4b8hFvoU9N?open=app%2Fapp.component.html)

## 🌍  Your Environment

**Angular Version:**
<pre><code>

"@angular/core": "9.0.0-rc.3",

</code></pre>

crisbeto:
Currently Ivy doesn't support adding `@SkipSelf` for some of the special DI tokens (e.g. `ViewContainerRef`, `ElementRef`, `ChangeDetectorRef`). These changes add support for it since it's something we support in ViewEngine.

Fixes #34066.

### DIFF
--- a/packages/core/src/render3/di.ts
+++ b/packages/core/src/render3/di.ts
@@ -333,7 +333,15 @@ export function getOrCreateInjectable<T>(
     // If the ID stored here is a function, this is a special object like ElementRef or TemplateRef
     // so just call the factory function to create it.
     if (typeof bloomHash === 'function') {
-      enterDI(lView, tNode);
+      if (flags & InjectFlags.SkipSelf) {
+        return getOrCreateInjectable(
+            tNode.parent, getParentInjectorView(getParentInjectorLocation(tNode, lView), lView),
+            token, flags & InjectFlags.Optional ? InjectFlags.Optional : InjectFlags.Default,
+            notFoundValue);
+      } else {
+        enterDI(lView, tNode);
+      }
+
       try {
         const value = bloomHash();
         if (value == null && !(flags & InjectFlags.Optional)) {
--- a/packages/core/test/acceptance/di_spec.ts
+++ b/packages/core/test/acceptance/di_spec.ts
@@ -698,6 +698,85 @@ describe('di', () => {
         expect(dirA.dirB.value).toEqual('parent');
       });
 
+      it('should support @SkipSelf when injecting ViewContainerRef', () => {
+        let parentViewContainer: ViewContainerRef|undefined;
+        let childViewContainer: ViewContainerRef|undefined;
+
+        @Directive({selector: '[parent]'})
+        class ParentDirective {
+          constructor(vc: ViewContainerRef) { parentViewContainer = vc; }
+        }
+
+        @Directive({selector: '[child]'})
+        class ChildDirective {
+          constructor(@SkipSelf() vc: ViewContainerRef) { childViewContainer = vc; }
+        }
+
+        @Component({template: '<div parent>parent <span child>child</span></div>'})
+        class MyComp {
+        }
+
+        TestBed.configureTestingModule({declarations: [ParentDirective, ChildDirective, MyComp]});
+        const fixture = TestBed.createComponent(MyComp);
+        fixture.detectChanges();
+
+        // Assert against the `element` since Ivy always returns a new ViewContainerRef.
+        expect(parentViewContainer !.element.nativeElement)
+            .toBe(childViewContainer !.element.nativeElement);
+      });
+
+      it('should support @SkipSelf when injecting ElementRef', () => {
+        let parentRef: ElementRef|undefined;
+        let childRef: ElementRef|undefined;
+
+        @Directive({selector: '[parent]'})
+        class ParentDirective {
+          constructor(elementRef: ElementRef) { parentRef = elementRef; }
+        }
+
+        @Directive({selector: '[child]'})
+        class ChildDirective {
+          constructor(@SkipSelf() elementRef: ElementRef) { childRef = elementRef; }
+        }
+
+        @Component({template: '<div parent>parent <span child>child</span></div>'})
+        class MyComp {
+        }
+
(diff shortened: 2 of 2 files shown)


packages/core/src/render3/di.ts
```diff
@@ -334,7 +334,15 @@ export function getOrCreateInjectable<T>(
     // If the ID stored here is a function, this is a special object like ElementRef or TemplateRef
     // so just call the factory function to create it.
     if (typeof bloomHash === 'function') {
-      enterDI(lView, tNode);
+      if (flags & InjectFlags.SkipSelf) {
+        return getOrCreateInjectable(
+            tNode.parent, getParentInjectorView(getParentInjectorLocation(tNode, lView), lView),
```
AndrewKushnir: @crisbeto I worked on a fix to support @SkipSelf for Injectors and noticed that `tNode.parent` may not be available when we cross component boundary, [see a test here](https://github.com/angular/angular/pull/34839/files#diff-881d964a81b3db8c8cb5311b8e3546ecR644). It looks like this code should be updated to handle this case as well ([similar to this line](https://github.com/angular/angular/pull/34839/files#diff-a29d4642602b40bdabe81990eed99e50R353)).)

crisbeto: Hmm, would that be a problem for `ViewContainerRef`, `ChangeDetectorRef` and `ElementRef` though (crossing component boundaries)? A node should always be inside another element or a view container so we'd always resolve the value before we have to cross a component boundary. That being said, I might be missing some context here.

IgorMinar:
is this PR still relevant now that Andrew has #34839 in the merge queue?

crisbeto:
I think so. #34839 handles `SkipSelf` only for the `Injector` token, whereas these changes do it for `ElementRef`, `ViewContainerRef` etc.

crisbeto:
Rebased which removed the need to bump the size limit.

packages/core/src/render3/di.ts
kara: I believe @AndrewKushnir is talking about the case where the `child` element requesting `SkipSelf` is a `ViewChild` of `parent` instead of a `ContentChild`. 

```ts
@Component({selector: 'parent', template: '<child></child>' }) 
export class Parent {...}
```

If `child` is a root node of the `parent` template, `tNode.parent` would be null because the `parent` element is in a different view and we can't cross component boundaries with the `tNode` tree (because `tNode.parent` would change with each usage of the component).  So `ViewContainerRef` would not be found.

In this case, we should be falling back to the `T_HOST` saved in the `LView`:

```suggestion
            tNode.parent || lView[T_HOST}, getParentInjectorView(getParentInjectorLocation(tNode, lView), lView),
```

packages/core/test/acceptance/di_spec.ts
```diff
@@ -698,6 +698,85 @@ describe('di', () => {
         expect(dirA.dirB.value).toEqual('parent');
       });
 
+      it('should support @SkipSelf when injecting ViewContainerRef', () => {
```
kara: Can we add a test where the child element requesting `ViewContainerRef` is a ViewChild of the parent element?

kara (changes requested):
Sorry for the delayed review! Comments below

packages/core/src/render3/di.ts
kara: One more thing to consider: when crossing component boundaries, we have to respect `@Host` config flags as well. In other words, if there is `@SkipSelf` and `Host` together, the `ViewContainerRef` should not be found if it would cross a component boundary.

kara: One more thing to consider: when crossing component boundaries, we have to respect `@Host` config flags as well. In other words, if there are `@SkipSelf` and `@Host` flags used together, the `ViewContainerRef` should not be found if it would cross a component boundary.

packages/core/src/render3/di.ts
```diff
@@ -333,7 +333,15 @@ export function getOrCreateInjectable<T>(
...
```
kara: One more thing to consider: when crossing component boundaries, we have to respect `@Host` config flags as well. In other words, if there are `@SkipSelf` and `@Host` flags used together, the `ViewContainerRef` should not be found if it would cross a component boundary.

AndrewKushnir:
FYI, it turned out that adding support or @Host decorator required extracting common logic into separate functions, so the commit from this PR is now cherry-pick'ed into #34839 and further review (and fixes if needed) will happen in #34839. Thank you.

PR #33264, opened 2019-10-19
refactor(ivy): remove ngBaseDef

crisbeto:
Removes `ngBaseDef` from the compiler and any runtime code that was still referring to it. In the cases where we'd previously generate a base def we now generate a definition for an abstract directive.

PR #34557, opened 2019-12-25
fix(ivy): don't run decorator handlers against declaration files

crisbeto:
Currently the decorator handlers are run against all `SourceFile`s in the compilation, but we shouldn't be doing it against declaration files. This initially came up as a CI issue in #33264 where it was worked around only for the `DirectiveDecoratorHandler`. These changes move the logic into the `TraitCompiler` and `DecorationAnalyzer` so that it applies to all of the handlers.

### DIFF
--- a/packages/compiler-cli/src/ngtsc/transform/test/compilation_spec.ts
+++ b/packages/compiler-cli/src/ngtsc/transform/test/compilation_spec.ts
@@ -1,0 +1,51 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import {absoluteFrom} from '../../file_system';
+import {runInEachFileSystem} from '../../file_system/testing';
+import {NOOP_INCREMENTAL_BUILD} from '../../incremental';
+import {NOOP_PERF_RECORDER} from '../../perf';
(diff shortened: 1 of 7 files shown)


petebacondarwin (approved):
One NIT.

crisbeto:

packages/compiler-cli/src/ngtsc/transform/test/compilation_spec.ts
```diff
@@ -0,0 +1,51 @@
...
+      const {program} = makeProgram([{
+        name: _('/lib.d.ts'),
+        contents: `export declare class SomeDirective {}`,
+      }]);
+
+      const checker = program.getTypeChecker();
+      const reflectionHost = new TypeScriptReflectionHost(checker);
+      const fakeDecoratorHandler: DecoratorHandler<{}|null, unknown, unknown> = {
+        name: 'FakeDecoratorHandler',
+        precedence: HandlerPrecedence.PRIMARY,
```
alxhub: I would prefer to avoid using jasmine spies in ngtsc. They're not type-safe and make refactoring and debugging tests unnecessarily difficult.

Instead, let's declare/implement a `FakeDecoratorHandler` for this test directly. It can throw if any of its methods are called.

crisbeto:

alxhub (approved):

atscott:
[Presubmit](http://test/OCL:289171339:BASE:289171369:1578697030353:8c05d388)

PR #34564, opened 2019-12-26
feat(language-service): support hover/definitions for structural directive

ivanwonder:
support hover/definitions for structural directive(e.g. `ngIf`, `ngFor`).

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/packages/language-service/src/locate_symbol.ts
+++ b/packages/language-service/src/locate_symbol.ts
@@ -150,16 +150,14 @@ function findAttribute(info: AstResult, position: number): Attribute|undefined {
 
 function findInputBinding(
     info: AstResult, path: TemplateAstPath, binding: BoundDirectivePropertyAst): Symbol|undefined {
-  const element = path.first(ElementAst);
-  if (element) {
-    for (const directive of element.directives) {
-      const invertedInput = invertMap(directive.directive.inputs);
-      const fieldName = invertedInput[binding.templateName];
-      if (fieldName) {
-        const classSymbol = info.template.query.getTypeSymbol(directive.directive.type.reference);
-        if (classSymbol) {
-          return classSymbol.members().get(fieldName);
-        }
+  const directive = path.parentOf(path.tail);
+  if (directive instanceof DirectiveAst) {
+    const invertedInput = invertMap(directive.directive.inputs);
+    const fieldName = invertedInput[binding.templateName];
+    if (fieldName) {
+      const classSymbol = info.template.query.getTypeSymbol(directive.directive.type.reference);
+      if (classSymbol) {
+        return classSymbol.members().get(fieldName);
       }
     }
   }
--- a/packages/language-service/src/locate_symbol.ts
+++ b/packages/language-service/src/locate_symbol.ts
@@ -148,14 +148,14 @@ function findAttribute(info: AstResult, position: number): Attribute|undefined {
   return path.first(Attribute);
 }
 
-function findParentOfDirectivePropertyAst(ast: TemplateAst[], binding: BoundDirectivePropertyAst) {
+function findParentOfBinding(ast: TemplateAst[], binding: BoundDirectivePropertyAst) {
   let res: DirectiveAst|undefined;
   const visitor = new class extends RecursiveTemplateAstVisitor {
     visitDirective(ast: DirectiveAst) {
       const result = this.visitChildren(ast, visit => { visit(ast.inputs); });
       return result;
     }
-    visitDirectiveProperty(ast: BoundDirectivePropertyAst, context: any) {
+    visitDirectiveProperty(ast: BoundDirectivePropertyAst, context: DirectiveAst) {
       if (ast === binding) {
         res = context;
       }
@@ -167,7 +167,7 @@ function findParentOfDirectivePropertyAst(ast: TemplateAst[], binding: BoundDire
 
 function findInputBinding(
     info: AstResult, path: TemplateAstPath, binding: BoundDirectivePropertyAst): Symbol|undefined {
-  const directiveAst = findParentOfDirectivePropertyAst(info.templateAst, binding);
+  const directiveAst = findParentOfBinding(info.templateAst, binding);
   if (directiveAst) {
     const invertedInput = invertMap(directiveAst.directive.inputs);
     const fieldName = invertedInput[binding.templateName];
--- a/packages/language-service/src/locate_symbol.ts
+++ b/packages/language-service/src/locate_symbol.ts
@@ -6,7 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {AST, Attribute, BoundDirectivePropertyAst, BoundEventAst, CompileTypeSummary, CssSelector, DirectiveAst, ElementAst, SelectorMatcher, TemplateAstPath, tokenReference} from '@angular/compiler';
+import {AST, Attribute, BoundDirectivePropertyAst, BoundEventAst, CompileTypeSummary, CssSelector, DirectiveAst, ElementAst, RecursiveTemplateAstVisitor, SelectorMatcher, TemplateAst, TemplateAstPath, templateVisitAll, tokenReference} from '@angular/compiler';
 
 import {AstResult} from './common';
 import {getExpressionScope} from './expression_diagnostics';
@@ -148,14 +148,31 @@ function findAttribute(info: AstResult, position: number): Attribute|undefined {
   return path.first(Attribute);
 }
 
+function findParentOfDirectivePropertyAst(ast: TemplateAst[], binding: BoundDirectivePropertyAst) {
+  let res: DirectiveAst|undefined;
+  const visitor = new class extends RecursiveTemplateAstVisitor {
+    visitDirective(ast: DirectiveAst) {
+      const result = this.visitChildren(ast, visit => { visit(ast.inputs); });
+      return result;
+    }
+    visitDirectiveProperty(ast: BoundDirectivePropertyAst, context: any) {
+      if (ast === binding) {
+        res = context;
+      }
+    }
+  };
+  templateVisitAll(visitor, ast);
+  return res;
+}
+
 function findInputBinding(
     info: AstResult, path: TemplateAstPath, binding: BoundDirectivePropertyAst): Symbol|undefined {
-  const directive = path.parentOf(path.tail);
-  if (directive instanceof DirectiveAst) {
-    const invertedInput = invertMap(directive.directive.inputs);
+  const directiveAst = findParentOfDirectivePropertyAst(info.templateAst, binding);
+  if (directiveAst) {
+    const invertedInput = invertMap(directiveAst.directive.inputs);
     const fieldName = invertedInput[binding.templateName];
     if (fieldName) {
-      const classSymbol = info.template.query.getTypeSymbol(directive.directive.type.reference);
+      const classSymbol = info.template.query.getTypeSymbol(directiveAst.directive.type.reference);
       if (classSymbol) {
         return classSymbol.members().get(fieldName);
       }
--- a/packages/language-service/src/locate_symbol.ts
+++ b/packages/language-service/src/locate_symbol.ts
@@ -148,12 +148,16 @@ function findAttribute(info: AstResult, position: number): Attribute|undefined {
   return path.first(Attribute);
 }
 
+// TODO: remove this function after the path includes 'DirectiveAst'.
+// Find the directive that corresponds to the specified 'binding'
+// at the specified 'position' in the 'ast'.
 function findParentOfBinding(
-    ast: TemplateAst[], binding: BoundDirectivePropertyAst, position: number) {
+    ast: TemplateAst[], binding: BoundDirectivePropertyAst, position: number): DirectiveAst|
+    undefined {
   let res: DirectiveAst|undefined;
   const visitor = new class extends RecursiveTemplateAstVisitor {
     visit(ast: TemplateAst): any {
-      let span = spanOf(ast);
+      const span = spanOf(ast);
       if (!inSpan(position, span)) {
         // Returning a value here will result in the children being skipped.
         return true;
--- a/packages/language-service/src/locate_symbol.ts
+++ b/packages/language-service/src/locate_symbol.ts
@@ -6,7 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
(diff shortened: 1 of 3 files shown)


packages/language-service/src/locate_symbol.ts
```diff
@@ -150,16 +150,14 @@ function findAttribute(info: AstResult, position: number): Attribute|undefined {
...
+  const directive = path.parentOf(path.tail);
+  if (directive instanceof DirectiveAst) {
+    const invertedInput = invertMap(directive.directive.inputs);
+    const fieldName = invertedInput[binding.templateName];
+    if (fieldName) {
+      const classSymbol = info.template.query.getTypeSymbol(directive.directive.type.reference);
+      if (classSymbol) {
+        return classSymbol.members().get(fieldName);
       }
     }
```
ivanwonder: Because the `binding` type is `BoundDirectivePropertyAst`, so the parent should be `DirectiveAst`. No need to creates a loop iterating over the head of `path`. And for the structural directive, the head of `path` is `EmbeddedTemplateAst`.

ayazhafiz (dismissed):

packages/language-service/src/locate_symbol.ts
ivanwonder: @ayazhafiz finds a bug for my code. I will fix it.

ivanwonder: when hovering the two-way binding, the parent of the tail is not a type of `DirectiveAst`.

packages/language-service/src/locate_symbol.ts
```diff
@@ -148,18 +148,33 @@ function findAttribute(info: AstResult, position: number): Attribute|undefined {
   return path.first(Attribute);
 }
 
+function findParentOfDirectivePropertyAst(ast: TemplateAst[], binding: BoundDirectivePropertyAst) {
+  let res: DirectiveAst|undefined;
```
ayazhafiz:
```suggestion
  let res: DirectiveAst|undefined = undefined;
```

packages/language-service/src/locate_symbol.ts
```diff
@@ -148,18 +148,33 @@ function findAttribute(info: AstResult, position: number): Attribute|undefined {
...
+  const visitor = new class extends RecursiveTemplateAstVisitor {
+    visitDirective(ast: DirectiveAst) {
+      const result = this.visitChildren(ast, visit => { visit(ast.inputs); });
+      return result;
+    }
+    visitDirectiveProperty(ast: BoundDirectivePropertyAst, context: any) {
```
ayazhafiz:
```suggestion
    visitDirectiveProperty(ast: BoundDirectivePropertyAst, context: DirectiveAst) {
```

packages/language-service/src/locate_symbol.ts
```diff
@@ -148,18 +148,33 @@ function findAttribute(info: AstResult, position: number): Attribute|undefined {
   return path.first(Attribute);
 }
 
+function findParentOfDirectivePropertyAst(ast: TemplateAst[], binding: BoundDirectivePropertyAst) {
```
ayazhafiz:
```suggestion
function findParentOfBinding(ast: TemplateAst[], binding: BoundDirectivePropertyAst) {
```

packages/language-service/src/locate_symbol.ts
```diff
@@ -148,18 +148,33 @@ function findAttribute(info: AstResult, position: number): Attribute|undefined {
...
+    }
+    visitDirectiveProperty(ast: BoundDirectivePropertyAst, context: any) {
+      if (ast === binding) {
+        res = context;
+      }
+    }
+  };
+  templateVisitAll(visitor, ast);
+  return res;
+}
```
ayazhafiz: > When hovering the two-way binding, the parent of the tail is not a type of DirectiveAst.

What is the parent in that case? I wonder if there is a way to find the parent from just traversing up the path rather than going through the entire template, which is expensive.

packages/language-service/test/definitions_spec.ts
```diff
@@ -262,6 +262,71 @@ describe('definitions', () => {
     }
   });
 
+  it('should be able to find a structural directive', () => {
+    const fileName = mockHost.addCode(`
```
ayazhafiz: Can you use `mockHost.override` instead? `addCode` is slower for test running.

packages/language-service/test/definitions_spec.ts
```diff
@@ -262,6 +262,71 @@ describe('definitions', () => {
...
+    const refFileName = '/node_modules/@angular/common/common.d.ts';
+    const def = definitions ![0];
+    expect(def.fileName).toBe(refFileName);
+    expect(def.name).toBe('ngIf');
+    expect(def.kind).toBe('property');
+    // Not asserting the textSpan of definition because it's external file
+  });
+
+  it('should be able to find a two-way binding', () => {
+    const fileName = mockHost.addCode(`
```
ayazhafiz: prefer `mockHost.override`

packages/language-service/test/hover_spec.ts
```diff
@@ -148,6 +148,36 @@ describe('hover', () => {
     expect(toText(displayParts)).toBe('(property) TestComponent.name: string');
   });
 
+  it('should be able to find a structural directive', () => {
+    const fileName = mockHost.addCode(`
```
ayazhafiz: prefer `mockHost.override`

packages/language-service/test/hover_spec.ts
```diff
@@ -148,6 +148,36 @@ describe('hover', () => {
...
+    const marker = mockHost.getDefinitionMarkerFor(fileName, 'ngIf');
+    const quickInfo = ngLS.getHoverAt(fileName, marker.start);
+    expect(quickInfo).toBeTruthy();
+    const {textSpan, displayParts} = quickInfo !;
+    expect(textSpan).toEqual(marker);
+    expect(toText(displayParts)).toBe('(property) NgIf<T>.ngIf: T');
+  });
+
+  it('should be able to find a reference to a two-way binding', () => {
+    const fileName = mockHost.addCode(`
```
ayazhafiz: prefer `mockHost.override`

ayazhafiz (changes requested):

packages/language-service/test/definitions_spec.ts
ivanwonder: Oh, sorry, I just copy the test case from here. It confuses me, there is `mockHost.addCode` and `mockHost.override` both here. Now I know and will replace my new test case.

packages/language-service/src/locate_symbol.ts
ivanwonder:
```
// 0: ElementAst
name: "test-comp"
attrs: [AttrAst]
inputs: []
outputs: [
  // 0: BoundEventAst
  name: "modelChange"
  start: ParseLocation {file: ParseSourceFile, offset: 24, line: 0, col: 24}
  end: ParseLocation {file: ParseSourceFile, offset: 40, line: 0, col: 40}
  fullName: "modelChange"
]
references: []
directives: (2) [
  // 0: DirectiveAst
  directive: {summaryKind: 1, type: {…}, isComponent: false, selector: "[string-model]", exportAs: null, …}
  inputs: Array(1) [
    // 0: BoundDirectivePropertyAst
    directiveName: "model"
    templateName: "model"
    start: ParseLocation {file: ParseSourceFile, offset: 24, line: 0, col: 24}
    end: ParseLocation {file: ParseSourceFile, offset: 40, line: 0, col: 40}
  ]
  hostProperties: []
  sourceSpan: ParseSourceSpan
    start: ParseLocation {file: ParseSourceFile, offset: 0, line: 0, col: 0}
    end: ParseLocation {file: ParseSourceFile, offset: 41, line: 0, col: 41}
]
sourceSpan: ParseSourceSpan {start: ParseLocation, end: ParseLocation, details: null}
endSourceSpan: ParseSourceSpan {start: ParseLocation, end: ParseLocation, details: null}
```
This is the templateAst for `<test-comp string-model ~{start-my}[(«model»)]="test"~{end-my}></test-comp>`.
Firstly, the `visitor` will the `ElementAst` and put it into `path`, then the `outputs` and record the `BoundEventAst`, when the code run here for the `DirectiveAst`
https://github.com/angular/angular/blob/9b9116c79d626b6356f3dfe0d48c9d990ac412a2/packages/language-service/src/utils.ts#L109
the `DirectiveAst` will be ignored because `DirectiveAst` contains `BoundEventAst`, At last, the  `BoundDirectivePropertyAst` will be put into `path`.
`path: (3) [ElementAst, BoundEventAst, BoundDirectivePropertyAst]`
You can debug the code here
https://github.com/angular/angular/blob/74bfe03d3a82433dcf9f2ff75a06beac2820a561/packages/language-service/test/definitions_spec.ts#L297

packages/language-service/src/locate_symbol.ts
ivanwonder: The default value is already undefined. No need to reassign.

packages/language-service/src/locate_symbol.ts
ayazhafiz: That makes sense, thanks. I think it would be better if we add the `DirectiveAst` to the path when it's encountered so the entire template doesn't have to be traversed, but this breaks some tests so we can do it in another commit.

packages/language-service/src/locate_symbol.ts
```diff
@@ -148,18 +148,57 @@ function findAttribute(info: AstResult, position: number): Attribute|undefined {
   return path.first(Attribute);
 }
 
+function findParentOfBinding(
+    ast: TemplateAst[], binding: BoundDirectivePropertyAst, position: number) {
```
ayazhafiz:
```suggestion
    ast: TemplateAst[], binding: BoundDirectivePropertyAst, position: number): DirectiveAst|undefined {
```

Just to make it easier when the functions. I think this will allow you to get rid of the typing on 153 as well.

packages/language-service/src/locate_symbol.ts
ivanwonder: Maybe I should add a comment here, `TODO: remove the function after the path includes 'DirectiveAst '`.

packages/language-service/src/locate_symbol.ts
ivanwonder: You can't get rid of it. If I remove the type of `res`, the typescript will regard it as `any` which is compatible with ` DirectiveAst|undefined`. No need to change here.

ayazhafiz: If you specify the return type, TypeScript will automatically infer the type of `res`. No need to change the type of `res`, but can you add the return type on the function definition itself, since this is a large function?

packages/language-service/src/locate_symbol.ts
```diff
@@ -148,18 +148,58 @@ function findAttribute(info: AstResult, position: number): Attribute|undefined {
...
+        visit(ast.children);
+      });
+    }
+
+    visitElement(ast: ElementAst, context: any): any {
+      return this.visitChildren(context, visit => {
+        visit(ast.directives);
+        visit(ast.children);
+      });
+    }
```
ayazhafiz: I don't think this is necessary; for succinctness it think it would be okay to inherit these methods from `RecursiveTemplateAstVisitor`.

ayazhafiz (commented):
Just a couple more comments. Thank you for this PR!

packages/language-service/src/locate_symbol.ts
ivanwonder: But it's useful for efficiency.

packages/language-service/src/locate_symbol.ts
```diff
@@ -148,18 +148,59 @@ function findAttribute(info: AstResult, position: number): Attribute|undefined {
   return path.first(Attribute);
 }
 
+// TODO: remove this function after the path includes 'DirectiveAst'.
+function findParentOfBinding(
+    ast: TemplateAst[], binding: BoundDirectivePropertyAst, position: number): DirectiveAst|
+    undefined {
```
ivanwonder: This is formatted automatically.

packages/language-service/src/locate_symbol.ts
ayazhafiz: I don't think it's too much of a big deal if we're going through the entire template anyway and the todo comment you've added to remove this after `DirectiveAst` is included in the path. I'm just worried about bloat in the codebase.

ayazhafiz (approved):
LGTM but let's wait for a review from @kyliau as well.

Thank you for the contributions!

packages/language-service/test/definitions_spec.ts
```diff
@@ -262,6 +264,63 @@ describe('definitions', () => {
...
+    expect(definitions !.length).toBe(1);
+    const def = definitions ![0];
+
+    const refFileName = '/app/parsing-cases.ts';
+    expect(def.fileName).toBe(refFileName);
+    expect(def.name).toBe('model');
+    expect(def.kind).toBe('property');
+    const content = mockHost.readFile(refFileName) !;
+    const ref = `@Input() model: string = 'model';`;
+    expect(def.textSpan).toEqual({
```
kyliau: It'd be clearer if this is written as
```ts
expect(content.substring(def.textSpan.start, def.textSpan.start + def.textSpan.length)).toEqual(`@Input() model: string = 'model';`);
```

packages/language-service/src/locate_symbol.ts
```diff
@@ -148,18 +148,59 @@ function findAttribute(info: AstResult, position: number): Attribute|undefined {
   return path.first(Attribute);
 }
 
+// TODO: remove this function after the path includes 'DirectiveAst'.
+function findParentOfBinding(
```
kyliau: Please provide brief documentation for the function.
e.g. "Find the directive that corresponds to the specified `binding` at the specified `position` in the `ast`."

packages/language-service/src/locate_symbol.ts
kyliau: I think it's ok to leave this as is, since we will likely come back to this very soon with the work to
(1) include `DirectiveAst` in the path
(2) change the impl to use Ivy

kyliau (changes requested):

kyliau (approved):

PR #34570, opened 2019-12-26
feat(language-service): completions for output $event properties

ayazhafiz:
This commit adds support for completions of properties on `$event`
variables in bound outputs.

This is the second major PR to support completions for `$event`
variables (angular/vscode-ng-language-service#531).
The final completion support that must be provided is for `$event`
variables in bindings targeting DOM events, like `(click)`.

## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [x] Feature

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

PR #34571, opened 2019-12-26
refactor(language-service): adds Symbol#typeArguments and does cleanup

ayazhafiz:
Adds a `typeArguments` method to the `Symbol` interface, cleaning up how
type parameters of a TypeScript type are currently found. This will be
necessary for providing completions for `$event` variables' properties
(#34570).

This commit also performs some fly-by cleanups seen while implementing
the `typeArguments` methods. There is more clean up to do in the
`typescript_symbols` file, but the scope of this commit didn't need to
get larger.

## PR Checklist

## PR Type

- [x] Feature

- [ ] Yes
- [x] No

### DIFF
--- a/packages/language-service/src/symbols.ts
+++ b/packages/language-service/src/symbols.ts
@@ -129,6 +129,11 @@ export interface Symbol {
    * If the symbol cannot be indexed, this method should return `undefined`.
    */
   indexed(argument: Symbol, key?: any): Symbol|undefined;
+
+  /**
+   * Returns the type arguments of a Symbol, if any.
+   */
+  typeArguments(): Symbol[]|undefined;
 }
 
 /**
--- a/packages/language-service/src/typescript_symbols.ts
+++ b/packages/language-service/src/typescript_symbols.ts
@@ -313,6 +313,7 @@ class TypeWrapper implements Symbol {
   }
 
   typeArguments(): Symbol[]|undefined {
+    // TODO: use checker.getTypeArguments when TS 3.7 lands in the monorepo.
     const typeArguments: ReadonlyArray<ts.Type> = (this.tsType as any).typeArguments;
     if (!typeArguments) return undefined;
     return typeArguments.map(ta => new TypeWrapper(ta, this.context));
--- a/packages/language-service/src/typescript_symbols.ts
+++ b/packages/language-service/src/typescript_symbols.ts
@@ -125,10 +125,10 @@ class TypeScriptSymbolQuery implements SymbolQuery {
 
   getElementType(type: Symbol): Symbol|undefined {
     if (type instanceof TypeWrapper) {
-      const elementType = getTypeParameterOf(type.tsType, 'Array');
-      if (elementType) {
-        return new TypeWrapper(elementType, type.context);
-      }
+      const tSymbol = type.tsType.symbol;
+      const tArgs = type.typeArguments();
+      if (!tSymbol || tSymbol.name !== 'Array' || !tArgs || tArgs.length != 1) return;
+      return tArgs[0];
     }
   }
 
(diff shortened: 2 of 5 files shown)


packages/language-service/src/symbols.ts
```diff
@@ -122,6 +122,11 @@ export interface Symbol {
    * If the symbol cannot be indexed, this method should return `undefined`.
    */
   indexed(argument: Symbol, key?: any): Symbol|undefined;
+
+  /**
+   * Returns the type arguments of a Symbol, if any.
+   */
+  typeArguments(): Symbol[]|undefined;
```
andrius-pra: The `typeArguments` property has been removed from the TypeReference interface in Typescipt 3.7. See #32962 (comment). I think it's better to move this method to `TypeScriptSymbolQuery`, otherwise you will need to add `TypeChecker` parameter to this method.

ayazhafiz: The class has access to the `TypeChecker` via the `context` property, so a new parameter would not be necessary.

andrius-pra: you are right

packages/language-service/src/typescript_symbols.ts
```diff
@@ -206,19 +206,19 @@ class TypeScriptSymbolQuery implements SymbolQuery {
...
 
-  private getTypeWrapper(symbol: Symbol): TypeWrapper|undefined {
-    let type: TypeWrapper|undefined = undefined;
-    if (symbol instanceof TypeWrapper) {
-      type = symbol;
-    } else if (symbol.type instanceof TypeWrapper) {
-      type = symbol.type;
-    }
-    return type;
+function getTypeWrapper(symbol: Symbol): TypeWrapper|undefined {
```
kyliau: I think we could remove this helper function entirely, or change it to
```ts
return symbol instanceof TypeWrapper ? symbol : symbol.type instanceof TypeWrapper ? symbol.type : undefined;
```

kyliau (changes requested):

packages/language-service/src/symbols.ts
ayazhafiz: Fixed.

packages/language-service/src/typescript_symbols.ts
```diff
@@ -303,6 +303,12 @@ class TypeWrapper implements Symbol {
         return sType && new TypeWrapper(sType, this.context);
     }
   }
+
+  typeArguments(): Symbol[]|undefined {
+    const typeArguments: ReadonlyArray<ts.Type> = (this.tsType as any).typeArguments;
```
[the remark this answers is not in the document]
ayazhafiz: Do you mean get `checker.getTypeArguments`? If so, looks like it's only available on TS 3.7 ([3.6.4 TypeChecker](https://github.com/microsoft/TypeScript/blob/d0340313982fe15561f7ca7ca16c7141e658bc6a/lib/typescript.d.ts#L1936) vs [3.7 TypeChecker](https://github.com/microsoft/TypeScript/blob/17f445035a100d846df313c92537050a74913d4b/lib/typescript.d.ts#L1977)).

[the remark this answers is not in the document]
kyliau: I was thinking of the proper way to get the type arguments from a `Type` without casting it to `any`.
In TS 3.6, it involves getting the `typeNode` from `type` (by querying the type checker), then checking if the `typeNode` `isExpressionWithTypeArguments`. We can then safely access `typeArguments` property.

Considering we are upgrading to TS 3.7 very soon, it's not strictly necessary to do this. If you decide to leave the code as is, could you please add a TODO and clean up the code once TS 3.7 lands?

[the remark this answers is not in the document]
ayazhafiz: I see. I tried the `isExpressionWithTypeArguments` but it began breaking tests, so I think I'll just leave as is for now. Comment added.

kyliau (approved):

PR #34574, opened 2019-12-27
fix(ivy): injecting incorrect provider when re-providing injectable with useClass

### LINKED ISSUE #34110: Undocumented Ivy difference for tree-shakeable providers
I've noticed an undocumented run time difference between Ivy and ViewEngine, (as of 9.0.0-rc3) which might be unintended, but which could make sense as a deliberate breaking change. I'm reporting this so that it can be evaluated, and likely either documented or changed. 

Please note, that I am **not** reporting that I have an application that broke because of this. I actually only discovered this because it interacts with a pull request I am working on.

### Test Case

Here is a minimal test case (run with the angular cli default main.ts):
```typescript
import { BrowserModule } from '@angular/platform-browser';
import { NgModule, DoBootstrap, ApplicationRef, Injectable, Inject} from '@angular/core';

@Injectable({providedIn: 'root'})
export class BarService{
  public getMessage(): string{
    return "Bar!";
  }
}

@Injectable({
  providedIn: 'root',
  useClass: BarService
})
export class FooService {
  public getMessage(): string{
    return "Foo?";
  }
}

@NgModule({
  imports: [BrowserModule],
  providers: [{provide: 'string_token', useClass: FooService}
]
})
export class AppModule implements DoBootstrap {
  //This is just to avoid needing a component in this minimal test case.
  ngDoBootstrap(appRef: ApplicationRef): void {} 

  constructor(@Inject('string_token') serv: FooService) {
    console.log(serv.getMessage());
  }
}
```

When run in IVY mode (either AOT or JIT), the message printed is "Bar!". When run in ViewEngine mode (again either AOT or JIT) the output is "Foo?".

The same difference also occurs if you drop the `@Inject('string_token')` decorator, and change the provider array to `[{provider: FooService, useClass: FooService}]` Or even to just `[FooService]`. If the provider array is removed entirely (while removing the `@Inject` decorator) then both Ivy and View engine will return "Bar!".

### Analysis

I understand code-wise why this difference occurs. Ivy's injector is coded code so that a type provider that injects a type that is a tree-shakeable provider will use the tree-shakeable provider's `ɵprov.factory` function. Ivy's injector also does the same for the case where `useClass` is present if no `dep` array is provided.

Looking into the commit history, it does not look like this change was necessarily deliberate. It not well covered by unit tests [Footnote 1]. It is not mentioned in the Ivy compatibility guide. 

The effective result of this is is that under Ivy, the presence or absence of any `useClass`, `deps`, `useValue`, or `useExisting` properties in a tree-shakeable provider is an implementation detail that is not visible to users of the injectable and users can treat the class as though they are not present. This is unlike with ViewEngine, where the difference can become visible if the class explicitly gets added to some other provider list, perhaps with the intention of getting a separate instance. 

So the Ivy behavior is reasonable. I believe the odds of this behavior breaking anybody are minuscule, for the following reasons: 

1. The `@Injectable` options other than `providedIn` are not very widely known, since they [don't show up in the documentation](https://github.com/angular/angular/issues/34107). 
2. It requires that somebody "redundantly" specify a tree-shakeable provider in a provider list, while it has one or more of these other properties, and have the resulting code work correctly if the type is instantiated directly instead of delegated (VE behavior) and not work if the delegation occurs (Ivy behavior). This seems rather unlikely to happen. Honestly, I suspect most people would be expecting the delegation if they tried this, and when it did not work, would have either removed the redundant provider, or made it also specify the delegated type.
3. As far as I can tell nobody else has reported this difference in behavior, and it presumably must not have affected google3. 

----------

### Footnotes
[1]: It looks like only [one test](https://github.com/angular/angular/blob/40a0666651331d427069cc770a15a875822e6db4/packages/core/test/acceptance/providers_spec.ts#L364) covers this, and that seems kind-of accidental, as what it was trying to test was a bug in code run by the `ɵprov.factory` for a tree-shakeable provider. While this case would have indeed have tested that codepath, so would the simple case of not specifying any provider, since as a tree-shakeable provider, the token itself is the provider.

crisbeto:
If an injectable has a `useClass`, Ivy injects the token in `useClass`, rather than the original injectable, if the injectable is re-provided under a different token. The correct behavior is that it should inject the re-provided token, no matter whether it has `useClass`.

Fixes #34110.

### DIFF
--- a/packages/core/src/di/r3_injector.ts
+++ b/packages/core/src/di/r3_injector.ts
@@ -12,6 +12,7 @@ import {OnDestroy} from '../interface/lifecycle_hooks';
 import {Type} from '../interface/type';
 import {getFactoryDef} from '../render3/definition';
 import {throwCyclicDependencyError, throwInvalidProviderError, throwMixedMultiProviderError} from '../render3/errors';
+import {FactoryFn} from '../render3/interfaces/definition';
 import {deepForEach, newArray} from '../util/array_utils';
 import {stringify} from '../util/stringify';
 import {resolveForwardRef} from './forward_ref';
@@ -407,7 +408,7 @@ export class R3Injector {
   }
 }
 
-function injectableDefOrInjectorDefFactory(token: Type<any>| InjectionToken<any>): () => any {
+function injectableDefOrInjectorDefFactory(token: Type<any>| InjectionToken<any>): FactoryFn<any> {
   // Most tokens will have an injectable def directly on them, which specifies a factory directly.
   const injectableDef = getInjectableDef(token);
   const factory = injectableDef !== null ? injectableDef.factory : getFactoryDef(token);
@@ -496,7 +497,8 @@ export function providerToFactory(
       if (hasDeps(provider)) {
         factory = () => new (classRef)(...injectArgs(provider.deps));
       } else {
-        return injectableDefOrInjectorDefFactory(classRef);
+        const classFactoryDef = getFactoryDef(classRef);
+        return classFactoryDef || injectableDefOrInjectorDefFactory(classRef);
       }
     }
   }
--- a/packages/core/src/di/r3_injector.ts
+++ b/packages/core/src/di/r3_injector.ts
@@ -479,7 +479,8 @@ export function providerToFactory(
     provider: SingleProvider, ngModuleType?: InjectorType<any>, providers?: any[]): () => any {
   let factory: (() => any)|undefined = undefined;
   if (isTypeProvider(provider)) {
-    return injectableDefOrInjectorDefFactory(resolveForwardRef(provider));
+    const unwrappedProvider = resolveForwardRef(provider);
+    return getFactoryDef(unwrappedProvider) || injectableDefOrInjectorDefFactory(unwrappedProvider);
   } else {
     if (isValueProvider(provider)) {
       factory = () => resolveForwardRef(provider.useValue);
@@ -497,8 +498,7 @@ export function providerToFactory(
       if (hasDeps(provider)) {
         factory = () => new (classRef)(...injectArgs(provider.deps));
       } else {
-        const classFactoryDef = getFactoryDef(classRef);
-        return classFactoryDef || injectableDefOrInjectorDefFactory(classRef);
+        return getFactoryDef(classRef) || injectableDefOrInjectorDefFactory(classRef);
       }
     }
   }
--- a/packages/core/test/acceptance/di_spec.ts
+++ b/packages/core/test/acceptance/di_spec.ts
@@ -1153,6 +1153,23 @@ describe('di', () => {
         expect(overriddenProvider !.dep.name).toBe('FooServiceDep');
       }
     });
+
+    it('should use constructor config directly when token is explicitly provided as a type provider',
+       () => {
+         let provider: FooService|BarService;
+
+         @Component({template: ''})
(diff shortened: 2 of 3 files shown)


packages/core/src/di/r3_injector.ts
```diff
@@ -496,7 +497,8 @@ export function providerToFactory(
       if (hasDeps(provider)) {
         factory = () => new (classRef)(...injectArgs(provider.deps));
       } else {
-        return injectableDefOrInjectorDefFactory(classRef);
+        const classFactory = injectableDefOrInjectorDefFactory(classRef);
+        factory = () => classFactory(classRef);
```
kara: This solution works well if the `classRef` doesn't use DI, but if `FooService` (token) and `BarService` (useClass in injectable def) have different DI signatures, we'll end up using `BarService`'s DI setup when instantiating `FooService`. So that case would be broken. Can we add a test for it?

In terms of a fix, it seems like the root problem is that we're going through the injectable def by default for anything that's explicitly provided. It seems like provider factories should not be doing this. Is there a reason we can't remove the `injectableDefOrInjectorDefFactory` call and replace it with `getFactoryDef` + the `getUndecoratedInjectableFactory` logic?

kara (changes requested):
Thanks for picking this up! Some Qs below

crisbeto:
Sorry for the delay @kara, the feedback has been addressed. Also you were right about the injected provider using the DI setup from the original class. I've implemented your comment on using the `getFactoryDef`, but I had to wrap the assertions in the tests with `if (ivyEnabled)`, because ViewEngine still has the behavior where it uses the wrong DI setup.

packages/core/test/acceptance/di_spec.ts
```diff
@@ -1066,6 +1066,95 @@ describe('di', () => {
...
+      expect(overriddenProvider !.getMessage()).toBe('foo');
+
+      // ViewEngine incorrectly uses the original class DI config, instead of the one from useClass.
+      if (ivyEnabled) {
+        expect(directProvider !.dep.name).toBe('BarServiceDep');
+        expect(overriddenProvider !.dep.name).toBe('FooServiceDep');
+      }
+    });
+  });
+
```
kara: I think we're still missing a test for the explicit type provider case, e.g.

```
providers: [FooService]
```

packages/core/src/di/r3_injector.ts
```diff
@@ -496,7 +497,8 @@ export function providerToFactory(
       if (hasDeps(provider)) {
         factory = () => new (classRef)(...injectArgs(provider.deps));
       } else {
-        return injectableDefOrInjectorDefFactory(classRef);
+        const classFactoryDef = getFactoryDef(classRef);
+        return classFactoryDef || injectableDefOrInjectorDefFactory(classRef);
```
kara: I can't comment on the line, but for type providers, it seems like we would also want to update line 482. Otherwise, if you provide using `providers: [FooService]` pattern, the injectable def would still be used by default.

crisbeto:
The latest feedback has been addressed @kara.

kara (approved):
LGTM

kara:
[presubmit](http://test/OCL:297229463:BASE:297229506:1582676298616:fa96c17c)

kara:
merge-assistance: my approval should be sufficient since there are only changes in core files

Issue #16223, opened 2017-04-21
feat(aio): show a "back to top" link in narrow view


wardbell:
**Feature suggestion - not a release blocker**

While dgeni is adding anchor/links to headings, it could also add this nifty "Back to top" widget just like they have on google web docs:

![image](https://cloud.githubusercontent.com/assets/129061/25259056/7941ece2-25f8-11e7-9402-ae031e0bc7a8.png)

See [this page for example](https://developers.google.com/fonts/docs/technical_considerations).

In google web docs, this widget is only available on H2s and is only visible in medium screen widths.  Seems reasonable.

gkalpak:
👍 For having a "Back to top" button".

Things that I don't get/would do differently:

- Why only show it on medium screens. What's wrong big wide screens?
- Why not have it always visible as a floating/fixed button?

petebacondarwin:
> Why not have it always visible as a floating/fixed button?
+1 for this, as long as it doesn't block contents

That being said, once we have the floating table of contents, there is less need for this feature at all IMO, since one can always click the first item of the ToC to get back to the top.

gkalpak:
Except for small screens where the TOC won't be floating.

petebacondarwin:
> Except for small screens where the TOC won't be floating.

OK, in that case (where the screen is narrow) we should have a floating "back to top" anchor that is stuck to the bottom of the screen. There is nothing special about headings as take off points for getting back to the top.

petebacondarwin:
Anyone want to take this on? It should be fairly straightforward to do...

sonukapoor:
@petebacondarwin I am taking on this one.

sonukapoor:
@petebacondarwin / @gkalpak What size is considered small screen?

gkalpak:
We decided not to implement this feature for now. See https://github.com/angular/angular/pull/34580#issuecomment-921032398.

PR #34580, opened 2019-12-27
feat(docs-infra): show "back to top" link on narrow screens

sonukapoor:
## PR Checklist

## PR Type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #16223

- [ ] Yes
- [X] No

### DIFF
--- a/aio/src/app/app.component.html
+++ b/aio/src/app/app.component.html
@@ -61,6 +61,10 @@
   <aio-lazy-ce selector="aio-toc"></aio-lazy-ce>
 </div>
 
+<div class="back-to-top-link" *ngIf="!hasFloatingToc">
+    <a (click)="scrollToTop()" class="material-icons" data-title="Back to top"></a>
+</div>
+
 <footer class="no-print">
   <aio-footer [nodes]="footerNodes" [versionInfo]="versionInfo"></aio-footer>
 </footer>
--- a/aio/src/app/app.component.html
+++ b/aio/src/app/app.component.html
@@ -61,9 +61,8 @@
   <aio-lazy-ce selector="aio-toc"></aio-lazy-ce>
 </div>
 
-<div class="back-to-top-link" *ngIf="!hasFloatingToc">
-    <a (click)="scrollToTop()" (keyup)="scrollToTop()" class="material-icons" data-title="Back to top" aria-label="Back to top">
-    </a>
+<div class="back-to-top-link no-print" *ngIf="!hasFloatingToc">
+    <a class="material-icons" title="Back to top" aria-label="Back to top" (click)="scrollToTop()" (keyup)="scrollToTop()">arrow_upward</a>
 </div>
 
 <footer class="no-print">
--- a/aio/src/app/app.component.html
+++ b/aio/src/app/app.component.html
@@ -62,7 +62,8 @@
 </div>
 
 <div class="back-to-top-link" *ngIf="!hasFloatingToc">
-    <a (click)="scrollToTop()" class="material-icons" data-title="Back to top"></a>
+    <a (click)="scrollToTop()" (keyup)="scrollToTop()" class="material-icons" data-title="Back to top" aria-label="Back to top">
+    </a>
 </div>
 
 <footer class="no-print">
--- a/aio/src/styles/2-modules/_api-pages.scss
+++ b/aio/src/styles/2-modules/_api-pages.scss
@@ -305,3 +305,27 @@
 .deprecated-api-item {
   text-decoration: line-through;
 }
+
+.back-to-top-link {
+  display: block;
+  margin-left: auto;
+  width: 25px;
+  z-index: 9999;
+  position: -webkit-sticky;
+  position: sticky;
+  bottom: 10px;
+
+  a {
+    color: $blue;
+    cursor: pointer;
+
+    &::before {
+      content: "arrow_upward";
+    }
+  
+    &:hover {
+      color: $mediumgray;
+      text-decoration: none;
+    }
+  }
+}
--- a/aio/src/styles/2-modules/_api-pages.scss
+++ b/aio/src/styles/2-modules/_api-pages.scss
@@ -314,18 +314,10 @@
   position: -webkit-sticky;
   position: sticky;
   bottom: 10px;
+  right: 10px;
 
   a {
     color: $blue;
     cursor: pointer;
-
-    &::before {
-      content: "arrow_upward";
-    }
-  
-    &:hover {
-      color: $mediumgray;
-      text-decoration: none;
-    }
   }
 }
--- a/aio/src/app/app.component.spec.ts
+++ b/aio/src/app/app.component.spec.ts
@@ -152,6 +152,40 @@ describe('AppComponent', () => {
         component.onResize(hideToCBreakPoint + 1);
         expect(component.hasFloatingToc).toBe(true);
       });
+
+      it('should contain a back to top link on narrow screens', () => {
+        component.onResize(hideToCBreakPoint - 1);
+        tocService.tocList.next([{}, {}, {}] as TocItem[]);
+        fixture.detectChanges();
+
+        const backToTopDiv = fixture.debugElement.query(By.css('.back-to-top-link'));
+        expect(backToTopDiv).not.toBeNull();
+        expect(backToTopDiv.nativeElement).toBeDefined();
+      });
+
+      it('should call scrollToTop when clicking the back to top link on narrow screens', () => {
+        const scrollService: ScrollService = fixture.debugElement.injector.get<ScrollService>(ScrollService);
+        const scrollToTopSpy: jasmine.Spy = spyOn(scrollService, 'scrollToTop');
+
+        component.onResize(hideToCBreakPoint - 1);
+        tocService.tocList.next([{}, {}, {}] as TocItem[]);
+        fixture.detectChanges();
+
+        const backToTopDiv  = fixture.debugElement.query(By.css('.back-to-top-link')).nativeElement as HTMLDivElement;
+        const anchorLink = backToTopDiv.firstChild as HTMLAnchorElement;
+        anchorLink.click();
+
+        expect(scrollToTopSpy).toHaveBeenCalled();
+      });
+
+      it('should not contain a back to top link on wider screens', () => {
+        component.onResize(hideToCBreakPoint + 1);
+        tocService.tocList.next([{}, {}, {}] as TocItem[]);
+        fixture.detectChanges();
+
+        const backToTopDiv = fixture.debugElement.query(By.css('.back-to-top-link'));
+        expect(backToTopDiv).toBeNull();
+      });
     });
 
     describe('isSideBySide', () => {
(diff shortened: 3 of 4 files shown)


aio/src/styles/2-modules/_api-pages.scss
```diff
@@ -305,3 +305,27 @@
...
 }
+
+.back-to-top-link {
+  display: block;
+  margin-left: auto;
+  width: 25px;
+  z-index: 9999;
+  position: -webkit-sticky;
+  position: sticky;
+  bottom: 10px;
```
gkalpak: I think it would better with an equivalent right offset: `right: 10px;`

aio/src/styles/2-modules/_api-pages.scss
```diff
@@ -305,3 +305,27 @@
...
+    color: $blue;
+    cursor: pointer;
+
+    &::before {
+      content: "arrow_upward";
+    }
+  
+    &:hover {
+      color: $mediumgray;
+      text-decoration: none;
```
gkalpak: This is covered by [this style][1]. (If anything, we should not be setting it for hover only anyway.)

[1]: https://github.com/angular/angular/blob/65974154e/aio/src/styles/0-base/_typography.scss#L131-L133

aio/src/styles/2-modules/_api-pages.scss
```diff
@@ -305,3 +305,27 @@
...
+  display: block;
+  margin-left: auto;
+  width: 25px;
+  z-index: 9999;
+  position: -webkit-sticky;
+  position: sticky;
+  bottom: 10px;
+
+  a {
+    color: $blue;
```
gkalpak: I wonder if it would make more sense to make it less distracting (e.g. partially faded out or a less bright color).

aio/src/styles/2-modules/_api-pages.scss
```diff
@@ -305,3 +305,27 @@
...
+    cursor: pointer;
+
+    &::before {
+      content: "arrow_upward";
```
gkalpak: Here too, I am on the fence on whether it would be better to have the more explicit `Back to top` text (instead of an icon), especially given that on the narrow screens that this is going to be displayed, there is no hover.

gkalpak (changes requested):
This is a great start, thx, @sonukapoor :tada:
I've left some comments, but it would be good to get @sjtrimble's opinion on the visual aspect.

BTW, alternatively, we could show a "back to top" button next to each section header, similar to what Google docs do (on narrow screens - [example][1]):

google-docs-back-to-top

[1]: https://developers.google.com/docs/api/quickstart/nodejs

aio/src/app/app.component.html
```diff
@@ -61,6 +61,11 @@
   <aio-lazy-ce selector="aio-toc"></aio-lazy-ce>
 </div>
 
+<div class="back-to-top-link" *ngIf="!hasFloatingToc">
```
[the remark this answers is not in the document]
gkalpak: Also, please add the [no-print][1] class.

[1]: https://github.com/angular/angular/blob/65974154e/aio/src/styles/_print.scss#L42

aio/src/styles/2-modules/_api-pages.scss
sonukapoor: I personally don't think that the colour is distracting. It matches with the rest of the icons (like github, edit). We can change it though if you like, below is a colour palette. The current colour is at 46%.

image

aio/src/app/app.component.html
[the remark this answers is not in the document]
sonukapoor: I agree with the logic to show/hide the ToC is very confusing and needs to be cleaned up. I believe the reason for the arrow link to appear on narrow screens, was so that users have a way to scroll to top. On normal screens, the arrow becomes redundant, because the user can use the ToC to scroll to top. The `no-print` class was added.

aio/src/app/app.component.html
```diff
@@ -61,6 +61,11 @@
...
+    <a (click)="scrollToTop()" (keyup)="scrollToTop()" class="material-icons" data-title="Back to top" aria-label="Back to top">
```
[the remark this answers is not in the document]
sonukapoor: Below are the answers in same order:

- Done
- `keyup` is required for accessibility. Linting fails with the error `click must be accompanied by either keyup, keydown or keypress event for accessibility` if it's not provided. 
- Done
- You are right. I simplified this to use `<a ...><i class="material-icons">arrow_upward</i></a>`
- Currently, the main page is `app.component.html`. Here we render the document with all sections. I do not see a use-case where a full-fledged component could be re-used. I am happy to change it if you see it being useful.

aio/src/styles/2-modules/_api-pages.scss
sonukapoor: Done.

aio/src/styles/2-modules/_api-pages.scss
sonukapoor: Good point. I have removed it.

aio/src/styles/2-modules/_api-pages.scss
sonukapoor: I have removed this entirely in favour of `<a ...><i class="material-icons">arrow_upward</i></a>`

aio/src/app/app.component.html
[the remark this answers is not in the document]
gkalpak: > You are right. I simplified this to use `<a ...><i class="material-icons">arrow_upward</i></a>`

That's not what I see in the code 😕 

> Currently, the main page is `app.component.html`. Here we render the document with all sections. I do not see a use-case where a full-fledged component could be re-used. I am happy to change it if you see it being useful.

If we decided to use the alternative mentioned in https://github.com/angular/angular/pull/34580#pullrequestreview-340525269, for example, we would have multiple instances of it. Other than that, I can't think of other re-use opportunities indeed.

The more I think about it, the more I lean toward `href="#top-of-page"`, which too makes it self-contained (i.e. you don't need the component to expose a `scrollToTop()` method), gets rid of the event handlers (and the associated overhead) and relies more on native browser capabilities (which is better for a11y too). WDYT?

aio/src/app/app.component.html
```diff
@@ -61,6 +61,10 @@
   <aio-lazy-ce selector="aio-toc"></aio-lazy-ce>
 </div>
 
+<div class="back-to-top-link no-print" *ngIf="!hasFloatingToc">
+    <a class="material-icons" title="Back to top" aria-label="Back to top" (click)="scrollToTop()" (keyup)="scrollToTop()">arrow_upward</a>
```
gkalpak: - Inconsistent indentation.
- I think for a11y it would be better to hide this altogether.

aio/src/app/app.component.html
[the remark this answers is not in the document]
gkalpak: Actually, depending on what behavior we want to implement, another benefit of a component is that it is easier to encapsulate more logic. E.g. we could have more rules on what pages it should and ahould not appear or have logic to hide the icon when the page is already scrolled to top.

aio/src/styles/2-modules/_api-pages.scss
gkalpak: I was referring to the `text-decoration` style only, but I see you removed `color` as well.
I would normally want to have the same hover behavior as other similar icon-links (such as the GitHub links), but given that this will only be shown on screens where there is no real hover, I think the current styles are fine.

aio/src/app/app.component.html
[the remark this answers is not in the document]
gkalpak: > I believe the reason for the arrow link to appear on narrow screens, was so that users have a way to scroll to top.

Indeed. But I am not sure about how I feel about the icon appearing on some "marketing" pages, such as the home page 🤔 

> On normal screens, the arrow becomes redundant, because the user can use the ToC to scroll to top.

Yes, but currently the icon appears even on pages that do not have a ToC anyway (so there was not way to scroll to top on any screen - why would we want to fix it on small screens but not larger ones). Most of these screens are relatively short, so scrolling manually is not a big deal.

Also, one of the main reasons to scroll to top is to get to the ToC (when it's not floating on the right), so that you can easily jump to a different section of the doc. On pages that have no ToC, scrolling to the top is less useful. So, I think I lean more towards only showing this on non-marketing docs.

aio/src/styles/2-modules/_api-pages.scss
gkalpak: My comment was not about where the `arrow_upward` will go, but more about whether we should have an arrow icon or a "Back to top" text.

aio/src/styles/2-modules/_api-pages.scss
gkalpak: I meant more like a gray-ish shade (to not draw that much attention. But let's keep the current color for now (we can always change it later).

aio/src/styles/2-modules/_api-pages.scss
```diff
@@ -305,3 +305,19 @@
...
+
+.back-to-top-link {
+  display: block;
+  margin-left: auto;
+  width: 25px;
+  z-index: 9999;
+  position: -webkit-sticky;
+  position: sticky;
+  bottom: 10px;
+  right: 10px;
```
gkalpak: I now realize that `right: 10px` will make the arrow slightly overlap with the content (which is not ideal).
Given that the main content has a `3rem` padding on the right, it might make sense to try an keep the arrow on that padding area (with something like `width: 2rem; right: 0.5rem; bottom: 0.5rem;`).

gkalpak (changes requested):

sonukapoor:
@gkalpak Do we still want to proceed with this? I sent you a google doc some time ago with some details on this.

gkalpak:
I still think it is a useful feature, but haven't had the time to look at the doc 😞
Let's keep it open for now and I'll try to get to it at some point (in the not very distant future :grin:).

gkalpak:
Sorry it took me so long to get to this, @sonukapoor 😓

I went through the doc and the comments in this PR and here are my thoughts/summary:
- The main usefulness of the link would lie in getting to an embedded ToC at the top of the page in order to navigate to a different part of the page.
- The link would not be particularly useful on marketing pages, because they don't have a ToC and are usually quite short (so manually scrolling is not an issue).
- On pages that have a floating ToC on the right (i.e. docs pages on larger screens), the user can easily navigate to different parts of the page (including to the top) using the ToC. So, again a "back to top" link would not add value there.
- This leaves us with docs pages on small screens (such as mobile devices).

(On top of that, we would have to figure out where to place the link, how to make it accessible (if at all), how to hide when already at the top, etc.)

##
Based on the above and taking into account that a very small portion of our visits come from mobile devices (according to Google analytics), I am thinking that the addition of a "back to top" link would benefit very few users and is not worth the effort. 

Also note that the "back to top" links seem to have been removed from google web docs ([example][1]), which served as an inspiration for the original feature request (see #16223) 😛

Therefore, I think should not move forward with this.

##
Thank you for the work you've put into it, @sonukapoor ❤️
It is greatly appreciated 🏅

<sub>(Even if it didn't end up getting merged, the work that leads to finding out you don't want to move forward with a feature is equally important with the actual implementation of a feature :wink:)</sub>

[1]: https://developers.google.com/fonts/docs/technical_considerations

PR #34589, opened 2019-12-29
Rules nodejs 1.0.1

gregmagolan:
Update to first stable release of rules_nodejs.

* Folds in rules_nodejs breaking changes that were made for 1.0 after BazelCoreWeb API review. The major one that affects the angular repo is the removal of the `bootstrap` attribute in nodejs_binary, nodejs_test and jasmine_node_test in favor of using templated_args `--node_options=--require=/path/to/script`. The side-effect of this is that the bootstrap script does not get the require.resolve patches with explicitly loading the targets `_loader.js` file.

* Brings in the fix for stamping which was preventing many targets from getting cached.

Breaking changes in rules_nodejs will be more controlled in the future.

This PR also updates integration/bazel and @angular/bazel schematics to rules_nodejs 1.0.0

For the purposes of the integration test the zone.js script & bundle script tags can just go into the source index.html itself. The purpose of the integration test is is to test @angular/bazel & ng_module & ng_package so there is no need to exercise html_insert_assets in integration/bazel.

--------------------------------------------------------

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No


gregmagolan:
@josephperrott `bazel test ...` is now fully cached with this update.

```
$ yarn test-non-ivy ...

INFO: Build completed successfully, 1 total action
//modules/benchmarks/src/class_bindings:perf_chromium           (cached) PASSED in 3.0s
//modules/benchmarks/src/expanding_rows:perf_chromium           (cached) PASSED in 54.0s
//modules/benchmarks/src/js-web-frameworks/ng2:perf_chromium    (cached) PASSED in 3.9s
//modules/benchmarks/src/largeform/ng2:perf_chromium            (cached) PASSED in 2.7s
//modules/benchmarks/src/largetable/baseline:perf_chromium      (cached) PASSED in 3.0s
//modules/benchmarks/src/largetable/incremental_dom:perf_chromium (cached) PASSED in 2.9s
//modules/benchmarks/src/largetable/iv:perf_chromium            (cached) PASSED in 3.4s
//modules/benchmarks/src/largetable/ng2:perf_chromium           (cached) PASSED in 2.6s
//modules/benchmarks/src/largetable/ng2_switch:perf_chromium    (cached) PASSED in 2.8s
//modules/benchmarks/src/tree/baseline:perf_chromium            (cached) PASSED in 4.5s
//modules/benchmarks/src/tree/incremental_dom:perf_chromium     (cached) PASSED in 3.8s
//modules/benchmarks/src/tree/iv:perf_chromium                  (cached) PASSED in 3.8s
//modules/benchmarks/src/tree/ng1:perf_chromium                 (cached) PASSED in 3.8s
//modules/benchmarks/src/tree/ng2:perf_chromium                 (cached) PASSED in 2.5s
//modules/benchmarks/src/tree/ng2_next:perf_chromium            (cached) PASSED in 3.7s
//modules/benchmarks/src/tree/ng2_static:perf_chromium          (cached) PASSED in 3.8s
//modules/benchmarks/src/tree/ng2_switch:perf_chromium          (cached) PASSED in 2.6s
//modules/benchmarks/src/tree/render3_function:perf_chromium    (cached) PASSED in 5.4s
//modules/playground/e2e_test/async:protractor_tests_chromium   (cached) PASSED in 3.6s
//modules/playground/e2e_test/hello_world:protractor_tests_chromium (cached) PASSED in 2.9s
//modules/playground/e2e_test/http:protractor_tests_chromium    (cached) PASSED in 3.9s
//modules/playground/e2e_test/jsonp:protractor_tests_chromium   (cached) PASSED in 3.3s
//modules/playground/e2e_test/key_events:protractor_tests_chromium (cached) PASSED in 2.5s
//modules/playground/e2e_test/model_driven_forms:protractor_tests_chromium (cached) PASSED in 3.5s
//modules/playground/e2e_test/order_management:protractor_tests_chromium (cached) PASSED in 3.5s
//modules/playground/e2e_test/person_management:protractor_tests_chromium (cached) PASSED in 2.3s
//modules/playground/e2e_test/relative_assets:protractor_tests_chromium (cached) PASSED in 2.3s
//modules/playground/e2e_test/routing:protractor_tests_chromium (cached) PASSED in 2.8s
//modules/playground/e2e_test/sourcemap:protractor_tests_chromium (cached) PASSED in 3.0s
//modules/playground/e2e_test/svg:protractor_tests_chromium     (cached) PASSED in 2.2s
//modules/playground/e2e_test/template_driven_forms:protractor_tests_chromium (cached) PASSED in 2.7s
//modules/playground/e2e_test/upgrade:protractor_tests_chromium (cached) PASSED in 3.2s
//modules/playground/e2e_test/web_workers/animations:protractor_tests_chromium (cached) PASSED in 3.2s
//modules/playground/e2e_test/web_workers/input:protractor_tests_chromium (cached) PASSED in 3.1s
//modules/playground/e2e_test/web_workers/kitchen_sink:protractor_tests_chromium (cached) PASSED in 3.4s
//modules/playground/e2e_test/web_workers/message_broker:protractor_tests_chromium (cached) PASSED in 2.6s
//modules/playground/e2e_test/web_workers/router:protractor_tests_chromium (cached) PASSED in 3.2s
//modules/playground/e2e_test/web_workers/todo:protractor_tests_chromium (cached) PASSED in 2.8s
//modules/playground/e2e_test/zippy_component:protractor_tests_chromium (cached) PASSED in 2.4s
//packages/animations/browser/test:test                         (cached) PASSED in 0.1s
//packages/animations/browser/test:test_web_chromium            (cached) PASSED in 2.0s
//packages/animations/test:test                                 (cached) PASSED in 0.1s
//packages/animations/test:test_web_chromium                    (cached) PASSED in 3.0s
//packages/bazel/src/schematics:test                            (cached) PASSED in 0.1s
//packages/bazel/test/ng_package:common_package                 (cached) PASSED in 0.1s
//packages/bazel/test/ng_package:core_package                   (cached) PASSED in 0.1s
//packages/bazel/test/ng_package:example_package                (cached) PASSED in 0.9s
//packages/bazel/test/ngc-wrapped:flat_module_test              (cached) PASSED in 0.1s
//packages/bazel/test/ngc-wrapped:ngc_test                      (cached) PASSED in 0.1s
//packages/benchpress/test:test                                 (cached) PASSED in 0.1s
//packages/common/http/test:test                                (cached) PASSED in 0.1s
//packages/common/http/test:test_web_chromium                   (cached) PASSED in 4.4s
//packages/common/http/testing/test:test                        (cached) PASSED in 0.1s
//packages/common/http/testing/test:test_web_chromium           (cached) PASSED in 4.8s
//packages/common/test:test                                     (cached) PASSED in 0.1s
//packages/common/test:test_web_chromium                        (cached) PASSED in 4.0s
//packages/common/upgrade/test:test                             (cached) PASSED in 0.1s
//packages/compiler-cli/integrationtest:integrationtest         (cached) PASSED in 115.6s
//packages/compiler-cli/integrationtest/bazel/injectable_def/app/test:test (cached) PASSED in 0.1s
//packages/compiler-cli/integrationtest/bazel/ng_module:test    (cached) PASSED in 0.6s
//packages/compiler-cli/ngcc/test:integration                   (cached) PASSED in 280.8s
  Stats over 4 runs: max = 280.8s, min = 275.1s, avg = 277.9s, dev = 2.7s
//packages/compiler-cli/ngcc/test:test                          (cached) PASSED in 0.0s
//packages/compiler-cli/src/ngtsc/annotations/test:test         (cached) PASSED in 0.0s
//packages/compiler-cli/src/ngtsc/cycles/test:test              (cached) PASSED in 0.0s
//packages/compiler-cli/src/ngtsc/entry_point/test:test         (cached) PASSED in 0.0s
//packages/compiler-cli/src/ngtsc/file_system/test:test         (cached) PASSED in 0.0s
//packages/compiler-cli/src/ngtsc/imports/test:test             (cached) PASSED in 0.0s
//packages/compiler-cli/src/ngtsc/indexer/test:test             (cached) PASSED in 0.0s
//packages/compiler-cli/src/ngtsc/partial_evaluator/test:test   (cached) PASSED in 0.0s
//packages/compiler-cli/src/ngtsc/reflection/test:test          (cached) PASSED in 0.0s
//packages/compiler-cli/src/ngtsc/scope/test:test               (cached) PASSED in 0.0s
//packages/compiler-cli/src/ngtsc/shims/test:test               (cached) PASSED in 0.0s
//packages/compiler-cli/src/ngtsc/typecheck/test:test           (cached) PASSED in 0.0s
//packages/compiler-cli/src/ngtsc/util/test:test                (cached) PASSED in 0.0s
//packages/compiler-cli/test:extract_i18n                       (cached) PASSED in 3.6s
//packages/compiler-cli/test:ngc                                (cached) PASSED in 2.0s
//packages/compiler-cli/test:perform_compile                    (cached) PASSED in 0.1s
//packages/compiler-cli/test:perform_watch                      (cached) PASSED in 0.1s
//packages/compiler-cli/test/diagnostics:check_types            (cached) PASSED in 3.5s
//packages/compiler-cli/test/diagnostics:typescript_version     (cached) PASSED in 0.3s
//packages/compiler-cli/test/metadata:test                      (cached) PASSED in 0.1s
//packages/compiler-cli/test/ngtsc:ngtsc                        (cached) PASSED in 0.0s
  Stats over 4 runs: max = 0.0s, min = 0.0s, avg = 0.0s, dev = 0.0s
//packages/compiler-cli/test/transformers:test                  (cached) PASSED in 3.9s
//packages/compiler/test:test                                   (cached) PASSED in 3.6s
//packages/compiler/test:test_web_chromium                      (cached) PASSED in 29.2s
//packages/compiler/test/css_parser:css_parser                  (cached) PASSED in 0.1s
//packages/compiler/test/css_parser:css_parser_web_chromium     (cached) PASSED in 3.6s
//packages/compiler/test/expression_parser:expression_parser    (cached) PASSED in 0.1s
//packages/compiler/test/expression_parser:expression_parser_web_chromium (cached) PASSED in 2.1s
//packages/compiler/test/ml_parser:ml_parser                    (cached) PASSED in 0.1s
//packages/compiler/test/ml_parser:ml_parser_web_chromium       (cached) PASSED in 4.1s
//packages/compiler/test/render3:test                           (cached) PASSED in 0.1s
//packages/compiler/test/selector:selector                      (cached) PASSED in 0.1s
//packages/compiler/test/selector:selector_web_chromium         (cached) PASSED in 3.0s
//packages/core/schematics/test:test                            (cached) PASSED in 0.1s
//packages/core/schematics/test/google3:google3                 (cached) PASSED in 0.1s
//packages/core/test:test                                       (cached) PASSED in 0.1s
  Stats over 4 runs: max = 0.1s, min = 0.1s, avg = 0.1s, dev = 0.0s
//packages/core/test:test_web_chromium                          (cached) PASSED in 3.4s
//packages/core/test/acceptance:acceptance                      (cached) PASSED in 0.1s
//packages/core/test/acceptance:acceptance_web_chromium         (cached) PASSED in 4.3s
//packages/core/test/render3:render3                            (cached) PASSED in 4.8s
//packages/core/test/render3:render3_web_chromium               (cached) PASSED in 4.0s
//packages/core/test/render3/perf:perf                          (cached) PASSED in 0.1s
//packages/core/test/strict_types:strict_types                  (cached) PASSED in 0.1s
//packages/core/test/view:view                                  (cached) PASSED in 3.3s
//packages/core/test/view:view_web_chromium                     (cached) PASSED in 14.4s
//packages/elements/schematics/ng-add:test                      (cached) PASSED in 0.0s
//packages/elements/test:test_chromium                          (cached) PASSED in 4.2s
//packages/examples/common:protractor_tests_chromium            (cached) PASSED in 3.2s
//packages/examples/core:protractor_tests_chromium              (cached) PASSED in 2.2s
//packages/examples/core:test                                   (cached) PASSED in 0.1s
//packages/examples/core/testing/ts:test                        (cached) PASSED in 0.1s
//packages/examples/forms:protractor_tests_chromium             (cached) PASSED in 2.9s
//packages/examples/service-worker/push:protractor_tests_chromium (cached) PASSED in 5.4s
//packages/examples/service-worker/registration-options:protractor_tests_chromium (cached) PASSED in 4.0s
//packages/examples/upgrade/static/ts/full:full_protractor_chromium (cached) PASSED in 3.1s
//packages/examples/upgrade/static/ts/lite:lite_protractor_chromium (cached) PASSED in 4.0s
//packages/examples/upgrade/static/ts/lite-multi:lite-multi_protractor_chromium (cached) PASSED in 2.3s
//packages/examples/upgrade/static/ts/lite-multi-shared:lite-multi-shared_protractor_chromium (cached) PASSED in 2.5s
//packages/forms/test:test                                      (cached) PASSED in 0.1s
//packages/forms/test:test_web_chromium                         (cached) PASSED in 3.0s
//packages/http/test:test                                       (cached) PASSED in 0.1s
//packages/http/test:test_web_chromium                          (cached) PASSED in 2.9s
//packages/language-service/test:diagnostics                    (cached) PASSED in 15.0s
//packages/language-service/test:test                           (cached) PASSED in 68.8s
//packages/localize/schematics/ng-add:test                      (cached) PASSED in 0.1s
//packages/localize/src/localize/test:test                      (cached) PASSED in 0.0s
//packages/localize/src/tools/test:test                         (cached) PASSED in 0.0s
//packages/localize/src/tools/test/translate/integration:integration (cached) PASSED in 0.1s
//packages/localize/src/utils/test:test                         (cached) PASSED in 0.0s
//packages/localize/test:test                                   (cached) PASSED in 0.0s
//packages/platform-browser-dynamic/test:test                   (cached) PASSED in 0.1s
//packages/platform-browser-dynamic/test:test_web_chromium      (cached) PASSED in 4.9s
//packages/platform-browser/animations/test:test                (cached) PASSED in 0.1s
//packages/platform-browser/animations/test:test_web_chromium   (cached) PASSED in 3.3s
//packages/platform-browser/test:test                           (cached) PASSED in 0.1s
//packages/platform-browser/test:test_web_chromium              (cached) PASSED in 3.1s
//packages/platform-server/test:test                            (cached) PASSED in 0.1s
//packages/platform-webworker/test:test                         (cached) PASSED in 0.1s
//packages/platform-webworker/test:test_web_chromium            (cached) PASSED in 3.2s
//packages/router/test:test                                     (cached) PASSED in 0.1s
//packages/router/test:test_web_chromium                        (cached) PASSED in 3.2s
//packages/router/test/aot_ngsummary_test:test                  (cached) PASSED in 0.1s
//packages/router/upgrade/test:test_web_chromium                (cached) PASSED in 2.7s
//packages/service-worker/config/test:test                      (cached) PASSED in 0.1s
//packages/service-worker/test:test                             (cached) PASSED in 0.1s
//packages/service-worker/test:test_web_chromium                (cached) PASSED in 2.4s
//packages/service-worker/worker/test:test                      (cached) PASSED in 0.1s
//packages/upgrade/src/common/test:test_chromium                (cached) PASSED in 4.1s
//packages/upgrade/src/dynamic/test:test_chromium               (cached) PASSED in 4.4s
//packages/upgrade/static/test:test_chromium                    (cached) PASSED in 2.5s
//packages/upgrade/static/testing/test:test_chromium            (cached) PASSED in 5.3s
//packages/zone.js/test:browser_green_test_karma_jasmine_test_chromium (cached) PASSED in 13.9s
//packages/zone.js/test:browser_legacy_test_karma_jasmine_test_chromium (cached) PASSED in 14.3s
//packages/zone.js/test:browser_shadydom_karma_jasmine_test_chromium (cached) PASSED in 8.2s
//packages/zone.js/test:browser_test_karma_jasmine_test_chromium (cached) PASSED in 12.5s
//packages/zone.js/test:karma_jasmine_test_ci_chromium          (cached) PASSED in 7.3s
//packages/zone.js/test:test_node                               (cached) PASSED in 0.1s
//packages/zone.js/test:test_node_bluebird                      (cached) PASSED in 0.0s
//packages/zone.js/test:test_node_error_disable_policy          (cached) PASSED in 0.0s
//packages/zone.js/test:test_node_error_lazy_policy             (cached) PASSED in 0.1s
//packages/zone.js/test:test_node_no_jasmine_clock              (cached) PASSED in 0.1s
//packages/zone.js/test:test_npm_package                        (cached) PASSED in 0.1s
//tools/public_api_guard:animations_api                         (cached) PASSED in 0.0s
//tools/public_api_guard:animations_browser_api                 (cached) PASSED in 2.9s
//tools/public_api_guard:animations_browser_testing_api         (cached) PASSED in 3.6s
//tools/public_api_guard:common_api                             (cached) PASSED in 4.2s
//tools/public_api_guard:common_http_api                        (cached) PASSED in 5.0s
//tools/public_api_guard:common_http_testing_api                (cached) PASSED in 3.2s
//tools/public_api_guard:common_testing_api                     (cached) PASSED in 2.3s
//tools/public_api_guard:common_upgrade_api                     (cached) PASSED in 4.5s
//tools/public_api_guard:core_api                               (cached) PASSED in 4.1s
//tools/public_api_guard:core_testing_api                       (cached) PASSED in 3.0s
//tools/public_api_guard:elements_api                           (cached) PASSED in 3.7s
//tools/public_api_guard:forms_api                              (cached) PASSED in 5.3s
//tools/public_api_guard:http_api                               (cached) PASSED in 4.2s
//tools/public_api_guard:http_testing_api                       (cached) PASSED in 2.8s
//tools/public_api_guard:localize_api                           (cached) PASSED in 3.0s
//tools/public_api_guard:ng_global_utils_api                    (cached) PASSED in 0.0s
//tools/public_api_guard:platform-browser-dynamic_api           (cached) PASSED in 2.6s
//tools/public_api_guard:platform-browser-dynamic_testing_api   (cached) PASSED in 4.4s
//tools/public_api_guard:platform-browser_animations_api        (cached) PASSED in 3.2s
//tools/public_api_guard:platform-browser_api                   (cached) PASSED in 3.4s
//tools/public_api_guard:platform-browser_testing_api           (cached) PASSED in 3.1s
//tools/public_api_guard:platform-server_api                    (cached) PASSED in 2.8s
//tools/public_api_guard:platform-server_testing_api            (cached) PASSED in 3.2s
//tools/public_api_guard:platform-webworker-dynamic_api         (cached) PASSED in 2.5s
//tools/public_api_guard:platform-webworker_api                 (cached) PASSED in 3.6s
//tools/public_api_guard:router_api                             (cached) PASSED in 4.1s
//tools/public_api_guard:router_testing_api                     (cached) PASSED in 4.2s
//tools/public_api_guard:router_upgrade_api                     (cached) PASSED in 3.6s
//tools/public_api_guard:service-worker_api                     (cached) PASSED in 3.0s
//tools/public_api_guard:service-worker_config_api              (cached) PASSED in 3.8s
//tools/public_api_guard:upgrade_api                            (cached) PASSED in 3.0s
//tools/public_api_guard:upgrade_static_api                     (cached) PASSED in 4.6s
//tools/public_api_guard:upgrade_static_testing_api             (cached) PASSED in 2.5s
//tools/size-tracking:test                                      (cached) PASSED in 0.1s
//tools/symbol-extractor:test                                   (cached) PASSED in 0.0s
//tools/ts-api-guardian:tests                                   (cached) PASSED in 47.3s
//tools/validate-commit-message:validate-commit-message         (cached) PASSED in 0.0s
```
I'd recommend all angular developers turn on disk_cache in their `~/.bazelrc` file:
```
build --disk_cache=~/.cache/bazel-disk-cache
```

gregmagolan:
material-unit-tests CI job waiting on https://github.com/angular/components/pull/18064 to land

josephperrott:
@gregmagolan Just to confirm, what you show here confirms that the results of the test targets being build is stable and leads to us having full cache hits for `test` action executions. 

Did you confirm that we also are seeing (as fully as possible), remote cache hits for the `build` actions themselves?  I believe we were already seeing everything return as cache hits for our test executions, but we determined that we were having to redo alot of work in the build process to get there still.

This is what I want to start looking into showing in our logs with something like https://github.com/angular/angular/pull/34439, so we can see how many of the individual spawned actions can be cache hits.

gregmagolan:
On a clean build with --disk_cash enabled and pre-populated I got,

```
$ yarn test-non-ivy ... --disk_cache=~/.cache/bazel-disk-cache

INFO: Elapsed time: 320.595s, Critical Path: 101.72s
INFO: 2620 processes: 2425 remote cache hit, 195 darwin-sandbox.
INFO: Build completed successfully, 44363 total actions

All tests cached & PASSED

Executed 0 out of 200 tests: 200 tests pass.
INFO: Build completed successfully, 44363 total actions
✨  Done in 321.95s.
```

On initial clean build with --config=remote to populate cache (OSX cross-platform) and with no --disk_cache,

```
$ yarn test-non-ivy ... --config=remote --google_credentials=/Users/greg/gcp_rbe.json --cpu=k8 --host_cpu=k8

INFO: Elapsed time: 1147.257s, Critical Path: 1078.01s
INFO: 2617 processes: 465 remote cache hit, 2152 remote.
INFO: Build completed successfully, 44361 total actions

All tests executed and PASSED

Executed 200 out of 200 tests: 200 tests pass.
INFO: Build completed successfully, 44361 total actions
✨  Done in 1147.83s.
```

On subsequent clean build with --config=remote to use the now populated cache (OSX cross-platform) and with no --disk_cache,

```
$ yarn test-non-ivy ... --config=remote --google_credentials=/Users/greg/gcp_rbe.json --cpu=k8 --host_cpu=k8

INFO: Elapsed time: 781.324s, Critical Path: 757.82s
INFO: 2617 processes: 2422 remote cache hit, 195 remote.
INFO: Build completed successfully, 44361 total actions

All tests cached & PASSED

Executed 0 out of 200 tests: 200 tests pass.
INFO: Build completed successfully, 44361 total actions
✨  Done in 783.08s.
```
This takes longer than using --disk_cache due to the extra network transmissino time

gregmagolan:
@josephperrott @IgorMinar Just a heads up that I initially updated to @angular/cli rc.7 but that breaks the cli-hello-world-ivy-i18n integration test. Other integration tests still pass so something particular about that one.

This PR now just updated to @angular/cli rc.4 to pick up the "0.0.0" @angular/core version fix otherwise the following is observed.
```
This version of CLI is only compatible with Angular versions ^9.0.0-beta || >=9.0.0 <10.0.0,
but Angular version 0.0.0 was found instead.
```

Somewhere between @angular/cli rc.5 & rc.7 the integration test breaks.

gregmagolan:
Same benchmark on current `master` has fewer remote cache hits and a number of actions that are forced local.

Clean build with --disk_cash enabled and pre-populated:

```
INFO: Elapsed time: 296.378s, Critical Path: 85.78s
INFO: 2407 processes: 2102 remote cache hit, 274 darwin-sandbox, 31 local.
INFO: Build completed successfully, 43512 total actions

All tests cached & PASSED

Executed 0 out of 200 tests: 200 tests pass.
There were tests whose specified size is too big. Use the --test_verbose_timeout_warnings command line option to see which ones these are.
INFO: Build completed successfully, 43512 total actions
```

On clean build with --config=remote (OSX cross-platform) enabled and pre-populated and with no --disk_cache:


```
INFO: Elapsed time: 733.490s, Critical Path: 698.26s
INFO: 2404 processes: 2100 remote cache hit, 31 local, 273 remote.
INFO: Build completed successfully, 43510 total actions

All tests cached & PASSED

Executed 0 out of 200 tests: 200 tests pass.
There were tests whose specified size is too big. Use the --test_verbose_timeout_warnings command line option to see which ones these are.
INFO: Build completed successfully, 43510 total actions
```

gregmagolan:
Updated to latest material commit which includes https://github.com/angular/components/pull/1806. This PR is ready to land.

gregmagolan:
Ugh. The dependency sandwich with material is tricky. The patches to @angular/bazel fail to apply since the changes are already made in this context. I'll have to make another PR to components to try/catch that patch.

gregmagolan:
Next pre-req PR made to components: https://github.com/angular/components/pull/18118

gregmagolan:
Update to material master & ready to land now for real.

gregmagolan:
Turned on Windows CI for this PR here: https://github.com/angular/angular/pull/34657

packages/animations/browser/test/BUILD.bazel
```diff
@@ -16,10 +16,11 @@ ts_library(
 
 jasmine_node_test(
     name = "test",
-    bootstrap = ["angular/tools/testing/init_node_spec.js"],
+    templated_args = ["--node_options=--require=$(rlocation $(location //tools/testing:node_es5))"],
```
devversion: I'm wondering if we could avoid repeating this for many `jasmine_node_test` targets by adding to `defaults.bzl`. `jasmine_node_test` seems to be already passed-through there. 

It looks like the whole `--node_options` argument is somewhat cumbersome and we repeat it quite often in codebase now. Going through the macro will avoid a lot of changes in the future too (i.e. avoiding multiple changes like in this PR which are spread across multiple files). Also it seems to bloat the actual readable build/test configuration files IMO.

<sup>We could differentiate based on a custom parameter whether node_no_angular, node etc.)

packages/core/test/render3/load_domino.ts
```diff
@@ -6,6 +6,12 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
+// bootstrap the bazel require patch since this bootstrap script is loaded with
+// `--node_options=--require=$(rlocation $(location script.js))`
+if (process.env['BAZEL_NODE_RUNFILES_HELPER']) {
+  require(process.env['BAZEL_NODE_RUNFILES_HELPER'] as string).patchRequire();
+}
```
devversion: Should we expand the comment to provide slightly more context? e.g.

```ts
// Load the Bazel Node runfiles helper script and patch the NodeJS module resolution. This is necessary
// to ensure that imports are properly resolved in Bazel. By default, the module resolution is patched in
// `nodejs_binary` targets, but since this file is loaded as bootstrap script through `node --require`,
// we need to manually setup the patched module resolution.

```

Also I asked myself if "not having to do this" would have been a strong enough point to keep the `bootstrap` attribute. There might be other stronger reasons on why the attribute has been removed though.

gregmagolan: Sure. I can add that. Thanks!

We're in a valley with the runfiles support right now having remove `bootstrap` but not applied the linker to run targets yet in rules_nodejs. In a near future rules_nodejs release, the runfiles helper won't be required as standard node module resolution works with the linker.

I suppose we could have left the `bootstrap` in for now and removed it for 2.0 but since we had an alternate approach that wasn't going to be broken in the future by the linker being added we decided to pull it for 1.0.

.circleci/config.yml
```diff
@@ -752,7 +751,6 @@ jobs:
       - restore_cache:
           keys:
             - *material_unit_tests_cache_key
-            - *material_unit_tests_cache_key_fallback
```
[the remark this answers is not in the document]
gregmagolan: Sounds good. I'll put in back in and add a comment to env.sh that it should be bumped if switching between forks.

integration/bazel/src/BUILD.bazel
```diff
@@ -64,14 +67,14 @@ terser_minified(
...
-web_package(
+pkg_web(
     name = "prodapp",
-    assets = [
-        # do not sort
-        "@npm//:node_modules/zone.js/dist/zone.min.js",
+    srcs = [
+        "index.html",
         ":bundle.min",
+        "@npm//:node_modules/zone.js/dist/zone.min.js",
```
IgorMinar: does the order no longer matter?

integration/bazel/src/BUILD.bazel
```diff
@@ -64,14 +67,14 @@ terser_minified(
...
     ],
-    index_html = "index.html",
+    additional_root_paths = ["npm/node_modules/zone.js/dist"],
```
IgorMinar: oh no.. do we really have to use fs paths rather than labels here? what is the relationship between `srcs` and `additional_root_paths`? Why do we need both?

integration/bazel/src/BUILD.bazel
```diff
@@ -64,14 +67,14 @@ terser_minified(
     src = ":bundle",
 )
 
-web_package(
+pkg_web(
```
IgorMinar: the changes to pkg_web and ts_devserver rules made the use of these rules much more cryptic :-( is there a good justification for the changes?

packages/bazel/src/builders/files/src/BUILD.bazel.template
```diff
@@ -64,18 +65,38 @@ terser_minified(
     src = ":bundle",
 )
 
-web_package(
+html_insert_assets(
+    name = "inject_scripts",
```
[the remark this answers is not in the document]
IgorMinar: +1 or even better something like: `asset_injected_index_html` - the label should preferably describe what the label represents.

packages/animations/browser/test/BUILD.bazel
IgorMinar: yup. please move this to the defaults.bzl.

IgorMinar (changes requested):
posting the comments I have so far... several more commits still to be reviewed.

packages/core/test/render3/load_domino.ts
IgorMinar: this additional information should also be included in the expanded comment so that we understand how we got to doing path-remapping magic in a helper script, even if it's just temporary.

packages/zone.js/test/node_bluebird_entry_point.ts
```diff
@@ -6,6 +6,15 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
+if (process.env['TEST_SRCDIR']) {
+  // bootstrap the bazel require resolve patch since this
+  // script is a bootstrap script loaded with --node_options=--require=...
```
IgorMinar: this (and the changes below) is the same code as the one above where we said that the comment should be expanded. isn't there  a better way to reuse all of this and ideally hide it via `defaults.bzl`, rather than sprinkle it all throughout the code base?

tools/defaults.bzl
```diff
@@ -108,13 +108,28 @@ def ts_library(tsconfig = None, testonly = False, deps = [], module_name = None,
...
 
+    native.filegroup(
+        name = "%s_es5" % name,
+        srcs = [":%s" % name],
+        testonly = testonly,
+        output_group = "es5_sources",
+    )
+
+    native.filegroup(
+        name = "%s_es6" % name,
```
IgorMinar: we standardized on "es2015" across Angular in favor of the unofficial "es6". What's the significance of this name? Can it be renamed to `_es2015`?

tools/ts-api-guardian/BUILD.bazel
```diff
@@ -68,18 +71,36 @@ ts_library(
...
+    name = "bootstrap",
+    testonly = True,
+    srcs = ["test/bootstrap.ts"],
+    tsconfig = "//tools:tsconfig-test",
+    deps = ["@npm//@types/node"],
+)
+
+filegroup(
+    name = "bootstrap_es5",
+    testonly = True,
```
IgorMinar: this could use a documentation - we are relying on the `testonly` flag to select only es5 sources (if I understand this correctly). For someone how doesn't know how the rules are implemented this is really hard to understand without any kind of documentation about what this target is about and how it works.

tools/ts-api-guardian/BUILD.bazel
```diff
@@ -68,18 +71,36 @@ ts_library(
...
+    srcs = [
+        ":test_lib",
+    ],
     data = glob([
         "test/fixtures/*.ts",
         "test/fixtures/*.patch",
     ]) + [
+        ":bootstrap_es5",
         ":ts-api-guardian",
         # TODO: remove this once the boostrap.js workaround has been removed.
```
IgorMinar: this comment no longer makes sense since the boostrap.js workaround no longer exists in the original form. Is this comment or the dependency it documents still relevant?

.circleci/config.yml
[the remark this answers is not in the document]
IgorMinar: @devversion what does this fallback cache do and how does it work? do you know? could you propose a doc comment explaining that Greg could incorporate in this PR (so that we don't create a conflicting PR with just the docs comment)

IgorMinar (changes requested):
submitting the rest of the comments. overall lgtm, I just hope we can reduce the duplication of the bootstrap code and document it better. thanks!

packages/animations/browser/test/BUILD.bazel
gregmagolan: Done. Good call. Since different tests require different bootstrap scripts I added `bootstrap` to the `jasmine_node_test` macro in defaults.bzl which takes the label and expands it to `templated_args --node_options=--require` and adds the `dep`. API is similar except bootstrap now takes a label instead of an fs path which is a plus.

packages/core/test/render3/load_domino.ts
devversion: I just thought about this more: could we have a separate script in `tools/` that does this, but we just _always_ add it to the `bootstrap` attribute (as first entry) in the `defaults.bzl`? That way, we can have a single place where this "code" needs to live?

(same

integration/bazel/src/BUILD.bazel
gregmagolan: pkg_web no longer does index.html injection (it just copies files as a pkg_* should do) so the order of `srcs` does not matter as these are now just a list of files to add to the package. Previously the order of `assets` mattered as these were scripts injected into index.html.

integration/bazel/src/BUILD.bazel
gregmagolan: This is so that `npm/node_modules/zone.js/dist/zone.min.js` is copied to `zone.min.js` in the package as the index.html script tag loads it from that path `<script src="/zone.min.js"></script>`. This is actually not a regression here as `additional_root_paths` was already in web_package & ts_devserver. Looking at the code before this PR, it looks like the injected script path was `<script src="/npm/node_modules/zone.js/dist/zone.min.js"></script>` since `additional_root_paths` was not being used. I'll change it to that to keep the diff at a minimum.

.circleci/config.yml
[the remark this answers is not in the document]
devversion: The fallback cache becomes useful if we update the `angular/components` SHA we test against. It basically ensures that we do not need to clone the _whole_ repository again. Instead, CircleCI provides us the latest outdated repository folder, we can update to the desired SHA.

Cloning takes longer than usually in this CI setup, because we don't clone with `--depth=1`, but rather clone the whole repository. This is because we want to use a specific SHA (and not a branch).

Here is my proposed comment. Also I'd still add the comment about bumping the cache key when switching the remote in `env.sh`.

```yml
# Whenever the `angular/components` SHA is updated, the cache key will no longer match. The
# fallback cache will still match, and CircleCI will restore the most recently cached
# repository folder. Without the fallback cache, we'd need to download the repository
# from scratch and it would slow down the job. This is because we can't clone the
# repository with reduced "--depth", but rather need to clone the whole repository to be
# able to support arbitrary SHA's.
- *material_unit_tests_cache_key_fallback
```

Let me know if something is unclear.

integration/bazel/src/BUILD.bazel
gregmagolan: The main justification was separation of concerns so that ts_devserver & pkg_web no longer did index.html injection. Its now done by another rule the output index.html can be piped into ts_devserver & pkg_web instead. Less magical this way and the API is cleaner.

Some more context on this:
* previously ts_devserver & pkg_web used a common helper function which has now been factored out into a pure js re-usable `html-insert-assets` npm package (https://www.npmjs.com/package/html-insert-assets).
* instead of maintaining this feature as part of ts_devserver & pkg_web in rules_nodejs, you can now use any npm tool (via the auto-generated npm_package_bin target) to perform html script injection and feed the output of that into ts_devserver or pkg_web.
* ts_devserver in google3 (which the oss ts_devserver shares code with) does not have the injection feature; it was added in rules_nodejs but we decided to remove it for 1.0 so that the rules did not diverge on that

tools/defaults.bzl
gregmagolan: Good call. It can be renamed to the standard 👍

packages/core/test/render3/load_domino.ts
gregmagolan: Done and done. Thanks for the suggestion @devversion. Worked like a charm and removed the churn in the bootstrap scripts.

.circleci/config.yml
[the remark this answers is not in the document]
gregmagolan: Looks good to me. Also added the comment about bumping the version on fork or branch change.

packages/zone.js/test/node_bluebird_entry_point.ts
gregmagolan: managed to hide this complication in defaults.bzl with help from @devversion

tools/ts-api-guardian/BUILD.bazel
gregmagolan: The testonly flag is required as the srcs target `srcs = [":bootstrap"],` is marked testonly and all downstream targets of a testonly target must also be marked testonly.

The output selection happens with `output_group = "es5_sources"` in that filegroup and this selects the es5 js output of ts_library to use in the `bootstrap` of the jasmine_node_test below: `bootstrap = [":bootstrap_es5"],`

The complication is unfortunate with ts_library that has 2 output flavors. I'm working on a simplified replacement tsc rule that only has a single output type and the `.js` files are in the DefaultInfo instead of the `.d.ts`. It would simplify much of this repo and get rid of the ts_library complication. Will have more info on that soon.

tools/ts-api-guardian/BUILD.bazel
gregmagolan: cleaned this target up so that it no longer depends on `../package.json` to determine the runfiles package path.

packages/bazel/src/builders/files/src/BUILD.bazel.template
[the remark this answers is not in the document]
gregmagolan: 👍

gregmagolan:
> submitting the rest of the comments. overall lgtm, I just hope we can reduce the duplication of the bootstrap code and document it better. thanks!

Thanks for the reviews @IgorMinar & @devversion. Updated based on your comments and got rid of all of the bootstrap code duplication --- its nicely abstracted by defaults.bzl now.

tools/ts-api-guardian/test/bootstrap.ts
```diff
@@ -8,11 +8,10 @@
...
 
-// Resolve the path to the package.json of the ts-api-guardian. We need to resolve an actual
+// Change directories to the path of the ts-api-guardian runfiles. We need to resolve an actual
 // path of a runfile because we want to determine the path to the directory that includes all
 // test fixture runfiles. On Windows this is usually the original non-sandboxed disk location,
 // otherwise this just refers to the runfile directory with all the proper symlinked files.
 // TODO: remove the whole bootstrap file once the tests are Bazel and Windows compatible.
-const runfilesDirectory = path.dirname(require.resolve('../package.json'));
-
+const runfilesDirectory = path.posix.join(process.env['RUNFILES'], 'angular/tools/ts-api-guardian');
```
devversion: IIRC we needed to use the runfile resolution (w/ manifest if needed) so that the tests work on Windows. Is there a specific reason we need to use the runfile tree?

packages/zone.js/test/BUILD.bazel
```diff
@@ -100,68 +131,91 @@ ts_library(
...
 
 jasmine_node_test(
     name = "test_node_error_lazy_policy",
-    bootstrap = [
-        "angular/packages/zone.js/test/node_error_lazy_policy_entry_point.js",
-    ],
+    bootstrap = [":node_error_lazy_policy_entry_point_es5"],
     deps = [
-        ":error_spec",
+        ":node_error_lazy_policy_entry_point",
```
devversion: I think the deps are unnecessary now, right?. We could do the same for all other targets that use `bootstrap`. I saw a couple of references to `//tools/testing:node` in other targets that use `bootstrap`. what do you think? 

<sup>It looks like you partially already did this; so there are just a few leftovers I guess</sup>

devversion (approved):
LGTM. nice work!

packages/zone.js/test/BUILD.bazel
gregmagolan: This particular one is necessary because the ":node_error_lazy_policy_entry_point_es5" target is a filegroup that isolates the bootstrap file and doesn't bring in the transitives. This is needed with ts_library (unfortunately) because it doesn't provide a predeclared output label for its .js outputs.

It's something that can be cleaned up in the future when a label to `:node_error_lazy_policy_entry_point.js` is available.

But adding `//tools/testing:node` to defaults.bzl is a good idea.

tools/ts-api-guardian/test/bootstrap.ts
gregmagolan: I'm not familiar with this test but I would guess that the test is assuming it is running the the output directory so it doesn't currently work on Windows. This change just cleans it up so its less hacky by not using the location of package.json to figure out where runfiles are. Much simpler to just using the RUNFILES env.

integration/bazel/src/BUILD.bazel
IgorMinar: I just find it really non-intuitive to see us using raw fs paths anywhere especially when it's not clear what the base dir of a relative url like `npm/...` in this case. I can keep the address space of targets and labels in my head, and I understand what `@npm//...` address resolves to but when I see "npm/..." I can read the path, but I'd struggle to write this without a lot of fs probing and testing. 

Maybe I'm just thinking about all of this incorrectly and don't have the right mental model, in which case I'd love to get some insights from you.

integration/bazel/src/BUILD.bazel
IgorMinar: ok. that makes sense. I'm just sad that a very common scenario got more complicated.

tools/ts-api-guardian/BUILD.bazel
IgorMinar: sgtm

.circleci/config.yml
```diff
@@ -752,6 +752,12 @@ jobs:
       - restore_cache:
           keys:
             - *material_unit_tests_cache_key
+            # Whenever the `angular/components` SHA is updated, the cache key will no longer match. The
+            # fallback cache will still match, and CircleCI will restore the most recently cached
+            # repository folder. Without the fallback cache, we'd need to download the repository
+            # from scratch and it would slow down the job. This is because we can't clone the
+            # repository with reduced "--depth", but rather need to clone the whole repository to be
+            # able to support arbitrary SHA's.
```
IgorMinar: awesome comment! thanks

.circleci/env.sh
```diff
@@ -72,6 +72,8 @@ setPublicVar SAUCE_READY_FILE_TIMEOUT 120
 # completely isolated from angular/angular in order to avoid any bad interactions between
 # their separate build setups.
 setPublicVar MATERIAL_REPO_TMP_DIR "/tmp/material2"
+# **NOTE**: When switching forks or branches, also bump the cache key & fallback cache key versions
+# in the CircleCI "config.yml".
 setPublicVar MATERIAL_REPO_URL "https://github.com/angular/material2.git"
```
IgorMinar: @devversion the repo was moved to angular/components a while back, can you please update this in a separate PR to the new URL and also update the MATERIAL_REPO_TMP_DIR path to `/tmp/angular-components`? thanks!

tools/defaults.bzl
```diff
@@ -256,10 +256,11 @@ def nodejs_binary(data = [], **kwargs):
         **kwargs
     )
 
-def jasmine_node_test(deps = [], **kwargs):
+def jasmine_node_test(bootstrap = [], **kwargs):
```
IgorMinar: can you add docs for the bootstrap attribute since it is custom and no longer documented upstream? thanks

tools/defaults.bzl
```diff
@@ -346,7 +346,7 @@ def rollup_bundle(name, testonly = False, **kwargs):
             name + ".js",
         ],
         args = [
-            "$(location :%s.es2015.js)" % name,
+            "$(execpath :%s.es2015.js)" % name,
```
IgorMinar: should we use `execpath` in jasmine_node_test as well? you just added `rlocation` and `location` there.

gregmagolan:
@IgorMinar I agree that the fs paths in `additional_root_paths` in ts_devserver & pkg_web are not ideal and I'd rather use labels everywhere. It may be possible to change these to labels same as we did for g3. I'll add this to my post-1.0 list. 

The ts_devserver rule has the same attribute I believe (@alexeagle would know.) as it originally came from there and the values are passed to the devserver go code through the `-packages` argument. `pkgs            = flag.String("packages", "", "root package(s) to serve, comma-separated")`.

tools/defaults.bzl
```diff
@@ -281,6 +281,11 @@ def jasmine_node_test(bootstrap = [], **kwargs):
     for label in bootstrap:
         deps += [label]
         templated_args += ["--node_options=--require=$(rlocation $(location %s))" % label]
+        if label.endswith("_es5"):
+            # If this label is a filegroup derived from a ts_library then automtically
```
IgorMinar: automtically -> automatically

tools/ts-api-guardian/test/bootstrap.ts
devversion: I added the `require.resolve` in the past for windows compatiblity. That's why I'm asking.

The Windows CI jobs don't seem to run the `ts-api-guardian` tests, so I don't think that they continue working on Windows. Usually on Windows, experimental runfile support is enabled, so I'm quite sure this won't work, unless I miss something?

.circleci/env.sh
devversion: will do.

tools/defaults.bzl
```diff
@@ -281,6 +281,11 @@ def jasmine_node_test(bootstrap = [], **kwargs):
...
+            # add the ts_library target (which is the label sans `_es5`) to deps so we pull
+            # in all of its transitive deps
```
IgorMinar: good comments don't just describe the code, but also explain why the code is there. Consider adding something like: "This removes the need for duplicate `deps` on the target and makes the usage of this rule less verbose."

IgorMinar (approved):
great work! just minor tweaks remain. I'm approving this for now, feel free to remove the cleanup label once the last set of comments are addressed.

IgorMinar:

IgorMinar:
@gregmagolan you'll also need to add the merge-assistance label once ready (in addition to removing the cleanup label). thanks!

tools/ts-api-guardian/test/bootstrap.ts
gregmagolan: Yup. This test is broken on Windows with the current `require.resolve` code as well as on Windows CI the runfiles tree is not laid out. I think the require patches will always be needed for Windows for users that don't turn on Windows symlinks but I'd like to reduce the scope of the require patch to the output tree and not have it deal with node_modules and modules maps.

For this particular test, it is currently depending on running in the input tree looks like which I think we can make work on Windows with a chdir to `./node_modules/angular/path/in/input/tree` once we enable the linker. I'll play with it once linker is enabled in rules_nodejs.

tools/defaults.bzl
gregmagolan: 👍

tools/defaults.bzl
gregmagolan: For now `$(rlocation)` can only handle `$(location)` but I've got a PR out to rules_nodejs to have `$(rlocation)` handle `$(rootpath)` which is more standard. Bazel recommends not using location:

https://docs.bazel.build/versions/master/be/make-variables.html#predefined_label_variables
```
location: A synonym for either execpath or rootpath, depending on the attribute being expanded. This is legacy pre-Starlark behavior and not recommended unless you really know what it does for a particular rule. See #2475 for details.
```

And currently we have a discrepancy in rules_nodejs with what `location` returns in `templated_args` vs `args` so I'd like to deprecate its usage in favor of $(rootpath) for runfiles (bazel run & bazel test) and $(execpath) for builds tools (bazel build).

That just a tweak in this repo as there are not many $(location) uses left.

tools/defaults.bzl
gregmagolan: Thanks. I will keep that in mind in the future.

tools/ts-api-guardian/test/bootstrap.ts
devversion: I ran it locally, and it currently passes on my end. I think what you are saying is that the tests just won't work in the new `rules_nodejs` version anymore, because the bootstrap script runs without the `requirePatches`, right?

 If that is the case, then we could consider using `jasmine_node_test` from `defaults.bzl` (currently not the case; probably why it won't work). Either way, I don't feel like it is too important.. I agree that the current `require.resolve` is not clean, but back at that time, it was the only way to get it work 😄

tools/defaults.bzl
gregmagolan: Roger.
```
    Args:
      bootstrap: A list of labels of scripts to run before the entry_point.

                 The labels can either be individual files or a filegroup that contain a single
                 file.
                 
                 The label is automatically added to the deps of jasmine_node_test.
                 If the label ends in `_es5` which by convention selects the es5 outputs
                 of a ts_library rule, then corresponding ts_library target sans `_es5`
                 is also added to the deps of jasmine_node_test.

                 For example with,

                 jasmine_node_test(
                     name = "test",
                     bootstrap = ["//tools/testing:node_es5"],
                     deps = [":test_lib"],
                 )

                 the `//tools/testing:node` target will automatically get added to deps
                 by this macro. This removes the need for duplicate deps on the
                 target and makes the usage of this rule less verbose.
```

gregmagolan:
Windows CI tested explicitly here: https://github.com/angular/angular/pull/34657 -- all GREEN

PR #34590, opened 2019-12-29
fix(ivy): handle overloaded constructors in ngtsc

crisbeto:
Currently ngtsc looks for the first `ConstructorDeclaration` when figuring out what the parameters are so that it can generate the DI instructions. The problem is that if a constructor has overloads, it'll have several `ConstructorDeclaration` members with a different number of parameters. These changes tweak the logic so it looks for the constructor implementation.

### DIFF
--- a/packages/compiler-cli/src/ngtsc/reflection/src/typescript.ts
+++ b/packages/compiler-cli/src/ngtsc/reflection/src/typescript.ts
@@ -35,8 +35,12 @@ export class TypeScriptReflectionHost implements ReflectionHost {
   getConstructorParameters(clazz: ClassDeclaration): CtorParameter[]|null {
     const tsClazz = castDeclarationToClassOrDie(clazz);
 
-    // First, find the constructor.
-    const ctor = tsClazz.members.find(ts.isConstructorDeclaration);
+    // First, find the constructor with a `body`. The constructors without a `body` are overloads
+    // whereas we want the implementation since it's the one that'll be executed and which can
+    // have decorators.
+    const ctor = tsClazz.members.find(
+        (member): member is ts.ConstructorDeclaration =>
+            ts.isConstructorDeclaration(member) && member.body !== undefined);
     if (ctor === undefined) {
       return null;
     }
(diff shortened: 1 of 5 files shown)


packages/compiler-cli/src/ngtsc/reflection/src/typescript.ts
```diff
@@ -35,8 +35,12 @@ export class TypeScriptReflectionHost implements ReflectionHost {
   getConstructorParameters(clazz: ClassDeclaration): CtorParameter[]|null {
     const tsClazz = castDeclarationToClassOrDie(clazz);
 
-    // First, find the constructor.
-    const ctor = tsClazz.members.find(ts.isConstructorDeclaration);
+    // First, find the constructor with a `body`. The constructors without a `body` are overloads
+    // whereas we want the implementation since it's the one that'll be executed and which can
+    // have decorators.
+    const ctor = tsClazz.members.find(
+        (member): member is ts.ConstructorDeclaration =>
```
crisbeto: I declared the type here, instead of having `as ts.ConstructorDeclaration | undefined` at the end, because clang was formatting it in a weird way that resulted in a syntax error.

alxhub: I think this is a perfect approach.

alxhub (approved):

alxhub:
[Presubmit](http://test/OCL:288946916:BASE:288946953:1578601206222:83bc1558)

atscott:
@crisbeto, There seems to be a failing test in google3 as a result of this PR. @alxhub, can you investigate?

AndrewKushnir:
@atscott FYI I re-ran [presubmit](https://test.corp.google.com/ui#id=OCL:289696445:BASE:289696578:1579030845473:160fa2cd) for this PR and the failures seem to be pre-existing or caused by flaky tests, so I'm marking g3 status as "green". Thank you.

atscott:
Thanks @AndrewKushnir

PR #34597, opened 2019-12-30
feat(language-service): support completions for animation

ivanwonder:
Support completions for animation.
https://angular.io/guide/animations
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No


ivanwonder:
I don't know if it's already prepared for review. Maybe I should wait until the completion support the full binding syntax (e.g. `<button [style.font-size.em]="isSpecial ? 3 : 1" >Big</button>`).
https://angular.io/guide/template-syntax#binding-types-and-targets

packages/language-service/src/completions.ts
```diff
@@ -265,6 +283,84 @@ function attributeCompletions(info: AstResult, path: AstPath<HtmlAst>): ng.Compl
...
+  }
+  for (const part of parts) {
+    if (start <= position && position <= start + part.length) {
+      return index;
+    }
+    index++;
+    start += part.length + 1;
+  }
+  return -1;
+}
```
ivanwonder: I don't find the right parser for it.

ayazhafiz:
Happy New Year, @ivanwonder. I will do my best to submit a review in the next day.

ivanwonder:
> Happy New Year, @ivanwonder. I will do my best to submit a review in the next day.

Happy New Year. 🎉

packages/language-service/src/completions.ts
```diff
@@ -265,6 +283,84 @@ function attributeCompletions(info: AstResult, path: AstPath<HtmlAst>): ng.Compl
...
 
+function isAnimationAttribute(name: string) {
+  return name.startsWith(ANIMATE_PROP_PREFIX) || name.startsWith(SECOND_ANIMATE_PROP_PREFIX);
+}
+
+// `cursorPosition` represents location of a cursor in the template(start with 0).
+//    k|ey.entry
+//     ^---------- cursor, return 0.
+//    key.ent|ry
+//           ^---- cursor, return 1.
```
ayazhafiz: Why does this return 1? Is it returning the cursor position in the nearest AST node? In either case, why not just take the difference between the template position and the start of the AST instead of this function?

packages/language-service/src/completions.ts
```diff
@@ -265,6 +283,84 @@ function attributeCompletions(info: AstResult, path: AstPath<HtmlAst>): ng.Compl
...
+    if (start <= position && position <= start + part.length) {
+      return index;
+    }
+    index++;
+    start += part.length + 1;
+  }
+  return -1;
+}
+
+function completionForAnimation(
```
```suggestion
function animationCompletions(
```

packages/language-service/src/completions.ts
```diff
@@ -265,6 +283,84 @@ function attributeCompletions(info: AstResult, path: AstPath<HtmlAst>): ng.Compl
...
+  let isSpecialAnimationControl = false;
+
+  if (name.startsWith(ANIMATE_PROP_PREFIX)) {
+    animationPrefixLength = ANIMATE_PROP_PREFIX.length;
+  }
+  if (name.startsWith(SECOND_ANIMATE_PROP_PREFIX)) {
+    animationPrefixLength = SECOND_ANIMATE_PROP_PREFIX.length;
+  }
+  name = name.substring(animationPrefixLength);
+  namePosition += animationPrefixLength;
```
ayazhafiz: I think it would be better if the name was just chopped off properly so the beginning of the name was at index 0 without also keeping track of a `namePosition`.

packages/language-service/src/completions.ts
```diff
@@ -265,6 +283,84 @@ function attributeCompletions(info: AstResult, path: AstPath<HtmlAst>): ng.Compl
...
+        result.push(...ANIMATIONEVENT);
+      }
+      break;
+    default:
+      break;
+  }
+  return result;
+}
+
+function completionForAnimationTrigger(info: AstResult): string[] {
```
```suggestion
function animationTriggerCompletions(info: AstResult): string[] {
```

packages/language-service/test/project/app/animations-case.ts
```diff
@@ -0,0 +1,28 @@
...
+      [@.disabled~{disabled}]
+      [animate-~{animate-prefix}]
+      (@openClose~{trigger}.done)="onAnimationEvent($event)"
+      (@openClose.~{event})="onAnimationEvent($event)"
+    ></div>
+  `,
+  animations: [trigger('openClose')]
+})
+export class AnimationsCaseComponent {
+}
```
ayazhafiz: Please place this inline in the test via `mockHost.override`. We will eventually move away from tests in `project/app`.

ayazhafiz (changes requested):

packages/language-service/src/completions.ts
```diff
@@ -72,6 +72,11 @@ enum ATTR {
   IDENT_EVENT_IDX = 10,
 }
 
+const ANIMATE_PROP_PREFIX = 'animate-';
```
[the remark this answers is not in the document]
ivanwonder: https://github.com/angular/angular/blob/9b9116c79d626b6356f3dfe0d48c9d990ac412a2/packages/compiler/src/template_parser/binding_parser.ts#L202

packages/language-service/src/completions.ts
ivanwonder:
```
<div (@openClose.done)="onAnimationEvent($event)"></div>
```
this is an example of `ast` of  ElementAst's `outputs` for the template above.
```
0: BoundEventAst
name: "openClose"
target: null
phase: "done"
handler: ASTWithSource {span: ParseSpan, sourceSpan: AbsoluteSourceSpan, ast: MethodCall, source: "onAnimationEvent($event)", location: "/app/animations-case.ts@4:25", …}
sourceSpan: ParseSourceSpan {start: ParseLocation, end: ParseLocation, details: null}
handlerSpan: ParseSourceSpan {start: ParseLocation, end: ParseLocation, details: null}
fullName: "@openClose.done"
isAnimation: true
```
I need to know where the user is, in the name `openClose` or phase `done`. So I return 1 which represent the phasse, 0 represent name. 
BTW, there is no `ast` output for template `<div (@openClose.done)></div>`

packages/language-service/src/completions.ts
ivanwonder: But the better place to achieve this is here
https://github.com/angular/angular/blob/0f7c08ae08f6478e591aaff06ba3c52f5f9e76c1/packages/language-service/src/completions.ts#L236
the different condition has different `namePosition`.
https://github.com/angular/angular/blob/0f7c08ae08f6478e591aaff06ba3c52f5f9e76c1/packages/language-service/src/completions.ts#L266
For `on-` it should be `offset + 3`, for `()`, it should be `offset + 1` (there is a bug for my code here, I will fix it ).

ivanwonder: I think for the binding type like  `<button [style.font-size.em]="isSpecial ? 3 : 1" >Big</button>` the name is chopped off properly

packages/language-service/test/project/app/parsing-cases.ts
```diff
@@ -141,9 +145,7 @@ export class TestComponent {
   @Output('test') testEvent = new EventEmitter();
 } /*EndTestComponent*/
 
-@Component({
-  templateUrl: 'test.ng',
-})
+@Component({templateUrl: 'test.ng', animations: [trigger('openClose')]})
```
ayazhafiz:
```suggestion
@Component({templateUrl: 'test.ng', animations: [trigger('openClose')],})
```

nit: just to make easier to read ))

ayazhafiz (changes requested):

packages/language-service/src/completions.ts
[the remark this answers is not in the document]
ayazhafiz: Wow, we should add some public documentation for this. It is only used in one test in the whole monorepo, and only with `bind-`.

packages/language-service/src/completions.ts
```diff
@@ -72,6 +72,10 @@ enum ATTR {
   IDENT_EVENT_IDX = 10,
 }
 
+// packages/compiler/src/template_parser/binding_parser.ts#L25
+const ANIMATE_PROP_PREFIX = 'animate-';
+const SECOND_ANIMATE_PROP_PREFIX = '@';
```
ayazhafiz: Consider semantic renaming, maybe something like `AT_ANIMATE_PROP_PREFIX`, or putting both prefixes in an array and then when you have to use them in something like

```typescript
// https://github.com/angular/angular/pull/34597/files#diff-124f548ae63f6d523acad1f921efecbeR336-R341
if (name.startsWith(ANIMATE_PROP_PREFIX)) {
  animationPrefixLength = ANIMATE_PROP_PREFIX.length;
}
if (name.startsWith(SECOND_ANIMATE_PROP_PREFIX)) {
  animationPrefixLength = SECOND_ANIMATE_PROP_PREFIX.length;
}
```

refactoring as something like

```typescript
const ANIMATE_PREFIXES = ['animate-', '@'];
/// ...
const prefix = ANIMATE_PREFIXES.find(pre => name.startsWith(pre));
const animationPrefixLength = prefix ? prefix.length : 0;
```

packages/language-service/src/completions.ts
```diff
@@ -241,15 +246,22 @@ function attributeCompletions(info: AstResult, path: AstPath<HtmlAst>): ng.Compl
...
+        bindParts[ATTR.IDENT_PROPERTY_IDX], info, position, attr.sourceSpan.start.offset + 1, elem,
+        ngAttrs));
+  } else if (bindParts[ATTR.KW_ON_IDX] !== undefined) {
+    results.push(...eventBindingCompletions(
+        bindParts[ATTR.KW_ON_IDX], info, position, attr.sourceSpan.start.offset + 3, elem,
+        ngAttrs));
+  } else if (bindParts[ATTR.IDENT_EVENT_IDX] !== undefined) {
+    results.push(...eventBindingCompletions(
+        bindParts[ATTR.IDENT_EVENT_IDX], info, position, attr.sourceSpan.start.offset + 1, elem,
+        ngAttrs));
```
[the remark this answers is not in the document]
ivanwonder: Before you can use `bindParts[ATTR.KW_IDENT_IDX]`, for example, firstly, get the value from `bindParts[ATTR.IDENT_EVENT_IDX]`. then `bindParts[ATTR.IDENT_EVENT_IDX].math(BIND_NAME_REGEXP)`.
The value of `bindParts[ATTR.KW_IDENT_IDX]` is undefined at the first name match.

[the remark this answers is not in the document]
ayazhafiz: Yes, sorry. For matches on groups 1-6, you can use `ATTR.KW_IDENT_IDX`. Otherwise, you must fallback on groups 8/9/10.

ivanwonder:
https://github.com/angular/angular/blob/9b9116c79d626b6356f3dfe0d48c9d990ac412a2/packages/compiler/src/template_parser/binding_parser.ts#L202
https://github.com/angular/angular/blob/9b9116c79d626b6356f3dfe0d48c9d990ac412a2/packages/compiler/src/template_parser/binding_parser.ts#L343
I don't know if it's a mistake or a feat. When parse the event binding, it only supports `@` prefix.

packages/language-service/src/completions.ts
ivanwonder: > Why not substring the name at the template position and check that if contains a . to see if we are in the name or the phase? Bounds checking are not necessary because they have already been handled when we determined that we were in an event binding AST. This would also reduce the function to a one-line flag.

What I think here is when the cursor is in `bind-`, no need to provide completions.
https://github.com/angular/angular/blob/9b9116c79d626b6356f3dfe0d48c9d990ac412a2/packages/language-service/src/completions.ts#L228
Maybe you are right, it should not be checked here.

ayazhafiz:
> don't know if it's a mistake or a feat. When parse the event binding, it only supports `@` prefix.

Yeah, I noticed the same thing. I'm not sure if this intentional or not. @alxhub @JoostK can you provide some insight? Seems that animation bindings can be specified with the `@` or `animate-` prefix, but `animate-` only works in property bindings (and I can't seem to find any public documentation on it).

packages/language-service/src/completions.ts
```diff
@@ -265,6 +271,115 @@ function attributeCompletions(info: AstResult, path: AstPath<HtmlAst>): ng.Compl
...
+
+class BoundElementPropertyInfo {
+  readonly isAnimation: boolean;
+  constructor(
+      public name: string, public type: PropertyBindingType, public unit: string|null,
+      public attribute: Attribute) {
+    this.isAnimation = this.type === PropertyBindingType.Animation;
+  }
+}
+
```
ivanwonder: This code copy from here and remove some useless property
https://github.com/angular/angular/blob/9b9116c79d626b6356f3dfe0d48c9d990ac412a2/packages/compiler/src/template_parser/template_ast.ts#L111
https://github.com/angular/angular/blob/9b9116c79d626b6356f3dfe0d48c9d990ac412a2/packages/compiler/src/template_parser/template_ast.ts#L86

ivanwonder:
> > don't know if it's a mistake or a feat. When parse the event binding, it only supports `@` prefix.
> 
> Yeah, I noticed the same thing. I'm not sure if this intentional or not. @alxhub @JoostK can you provide some insight? Seems that animation bindings can be specified with the `@` or `animate-` prefix, but `animate-` only works in property bindings (and I can't seem to find any public documentation on it).

Maybe we should open an issue for it.

JoostK:
> can you provide some insight? Seems that animation bindings can be specified with the `@` or `animate-` prefix, but `animate-` only works in property bindings (and I can't seem to find any public documentation on it).

@ayazhafiz I have literally no clue about how this is supposed to be.

packages/language-service/src/completions.ts
```diff
@@ -265,6 +271,115 @@ function attributeCompletions(info: AstResult, path: AstPath<HtmlAst>): ng.Compl
...
+function animationEventBindingCompletions(
+    bound: BoundEventInfo, info: AstResult, templatePosition: number): string[] {
+  if (isInAnimationName(bound, templatePosition)) {
+    return animationTriggerCompletions(info);
+  }
+  if (isInAnimationPhase(bound, templatePosition)) {
+    const ANIMATION_PHASES = ['start', 'done'];
+    return ANIMATION_PHASES;
```
ayazhafiz: nit: consider removing the constant entirely in this case

packages/language-service/src/completions.ts
```diff
@@ -265,6 +271,115 @@ function attributeCompletions(info: AstResult, path: AstPath<HtmlAst>): ng.Compl
...
+  }
+}
+
+function animationPropertyBindingCompletions(
+    bound: BoundElementPropertyInfo, info: AstResult): string[] {
+  // A special animation control binding called @.disabled.
+  // https://angular.io/guide/transition-and-triggers#disabling-an-animation-on-an-html-element
+  if (bound.name.startsWith('.')) {
+    const SPECIAL_ANIMATION_CONTROL = ['disabled'];
+    return SPECIAL_ANIMATION_CONTROL;
```
ayazhafiz: nit: consider removing the constant entirely in this case, and adding the comment on 331-332 inside the if branch.

packages/language-service/src/completions.ts
```diff
@@ -265,6 +271,115 @@ function attributeCompletions(info: AstResult, path: AstPath<HtmlAst>): ng.Compl
...
+    return animationTriggerCompletions(info);
+  }
+}
+
+function animationEventBindingCompletions(
+    bound: BoundEventInfo, info: AstResult, templatePosition: number): string[] {
+  if (isInAnimationName(bound, templatePosition)) {
+    return animationTriggerCompletions(info);
+  }
+  if (isInAnimationPhase(bound, templatePosition)) {
```
ayazhafiz: if we're not in the animation name, we have to be in the animation phase right? If so, I think we should remove the functions and `isIn...` function and perform the check in this function.

packages/language-service/src/completions.ts
```diff
@@ -265,6 +271,115 @@ function attributeCompletions(info: AstResult, path: AstPath<HtmlAst>): ng.Compl
...
+      name = name.substring(ANIMATE_PROP_PREFIX.length);
+    }
+    if (name.startsWith(AT_ANIMATE_PROP_PREFIX)) {
+      name = name.substring(AT_ANIMATE_PROP_PREFIX.length);
+    }
+    return new BoundElementPropertyInfo(name, PropertyBindingType.Animation, null, attr);
+  } else {
+    return new BoundElementPropertyInfo(name, PropertyBindingType.Property, null, attr);
+  }
+}
```
ayazhafiz: I don't think we should duplicate the job of the parser in the language service, especially since this could be done by creating and using a binding parser. Were you unable to extract the event/binding name using `KW_IDENT_IDX` and friends?

ayazhafiz (changes requested):
Hi @ivanwonder, sorry for the delay. I have recently started a new job and been busy with the move and such. I will make an effort to be more active. I hope we can get this PR in soon. Keen should be back soon for a second pair of eyes as well.

ayazhafiz:
> Maybe we should open an issue for it.

Agreed!

packages/language-service/src/completions.ts
ivanwonder: For me, create a new one is simple and readable.
One way to get it is to visit the `TemplateAst`(now the code use HtmlAst), but it has a problem as the reference ASTs in my earlier PR.(this way is better, but still need to create the `BoundEventAst ` when encountering an error).
If I just use the exported classes, I need to create `public handler: AST` for it, and `this.isAnimation = !!this.phase;` the `phase` maybe null or an empty string.
https://github.com/angular/angular/blob/9b9116c79d626b6356f3dfe0d48c9d990ac412a2/packages/compiler/src/template_parser/template_ast.ts#L115-L212

ivanwonder:
> Hi @ivanwonder, sorry for the delay. I have recently started a new job and been busy with the move and such. I will make an effort to be more active. I hope we can get this PR in soon. Keen should be back soon for a second pair of eyes as well.

Thanks for your reviews so much. I will do my best to finish this pr.

packages/language-service/src/completions.ts
ivanwonder: https://github.com/angular/angular/blob/9b9116c79d626b6356f3dfe0d48c9d990ac412a2/packages/language-service/src/completions.ts#L50-L51
Only contain `@` here, or I add `animate-` to it.

ivanwonder: https://github.com/angular/angular/pull/34597#discussion_r364052767
The `TemplateAst` uses the binding parser, but just like I said, `this.isAnimation = !!this.phase;`, this is a problem. This is why the new `BoundEventInfo` I create has the fifth parameter.

ivanwonder:
> > Maybe we should open an issue for it.
> 
> Agreed!

@ayazhafiz https://github.com/angular/angular/issues/34673 I open a feature request issue

ayazhafiz:
I think we should wait on figuring out the disparity between the `@` and `animate-` prefixes before moving forward with this PR, because the diff could be large depending on whether the framework wants to support both or one prefix for events or bindings.

kyliau:
> > don't know if it's a mistake or a feat. When parse the event binding, it only supports `@` prefix.
> 
> Yeah, I noticed the same thing. I'm not sure if this intentional or not. @alxhub @JoostK can you provide some insight? Seems that animation bindings can be specified with the `@` or `animate-` prefix, but `animate-` only works in property bindings (and I can't seem to find any public documentation on it).

@matsko Do you know the specifics about the `@` and `animate-` prefix?

AndrewKushnir:
@kyliau @ayazhafiz, just want to check if the changes in this PR are still relevant in a context of upcoming Ivy-based Language Service? Thank you.

ivanwonder:
I think this should be resolved in the template parser. Now the parse will break when encounter the error, for example, `<div (@¦)="onAnimationEvent($event)"></div>`, the `outputs` of the `TemplateAst` is empty. @ayazhafiz

ayazhafiz:
> I think this should be resolved in the template parser. Now the parse will break when encounter the error, for example, `<div (@¦)="onAnimationEvent($event)"></div>`, the `outputs` of the `TemplateAst` is empty. @ayazhafiz

I agree this should be a parser-side fix, and then re 

> @kyliau @ayazhafiz, just want to check if the changes in this PR are still relevant in a context of upcoming Ivy-based Language Service? Thank you.

We should probably just use this work for the ivy-native language service.

@ivanwonder do you want to create a patch for the binding parser to handle this case? I think the r3_template_transform code is the only place we need to worry about this, if we are not interested in the fix for ViewEngine.

ivanwonder:
> @ivanwonder do you want to create a patch for the binding parser to handle this case?

yes, I will try

ivanwonder:
I find the test case support the `bind-animate-eventName`, but `on-animate-name.phase` doesn't.
https://github.com/angular/angular/blob/eac85b4e235fc99306945012a62ba4333e524718/packages/compiler/test/template_parser/template_parser_spec.ts#L726-L736

ivanwonder:
> > I think this should be resolved in the template parser. Now the parse will break when encounter the error, for example, `<div (@¦)="onAnimationEvent($event)"></div>`, the `outputs` of the `TemplateAst` is empty. @ayazhafiz
> 
> I agree this should be a parser-side fix, and then re
> 
> > @kyliau @ayazhafiz, just want to check if the changes in this PR are still relevant in a context of upcoming Ivy-based Language Service? Thank you.
> 
> We should probably just use this work for the ivy-native language service.
> 
> @ivanwonder do you want to create a patch for the binding parser to handle this case? I think the r3_template_transform code is the only place we need to worry about this, if we are not interested in the fix for ViewEngine.

@ayazhafiz  submit a PR #39925

AndrewKushnir:
@ivanwonder should we close this PR since #39925 has landed? Thank you.

ivanwonder:
> @ivanwonder should we close this PR since #39925 has landed? Thank you.

I will try to close this by this [PR](https://github.com/angular/angular/pull/40300#issuecomment-761553514)

jessicajaniuk:
Closing this since View Engine is no more and the [Ivy PR](https://github.com/angular/angular/pull/40300#issuecomment-761553514) is still around.

PR #34599, opened 2019-12-30
docs: fix lazy routes code snippet format

alan-agius4:


gkalpak (approved):
This is better, but ideally these examples should be [extracted from docregions][1] in [aio/content/examples/lazy-loading-ngmodules/src/app/app-routing.module.ts][2] :wink:

[1]: https://angular.io/guide/docs-style-guide#code-snippet-from-a-file
[2]: https://github.com/angular/angular/blob/9b9116c79/aio/content/examples/lazy-loading-ngmodules/src/app/app-routing.module.ts#L8-L22

alan-agius4:
How is that possible? Ie: having a file with a before and after a schematic has been executed?

gkalpak:
@googlebot I consent.

gkalpak:
**merge-assistance:** Global approval for this docs-only change.

aio/content/examples/lazy-loading-ngmodules/src/app/app-routing.module.ts
```diff
@@ -4,23 +4,26 @@ import { NgModule } from '@angular/core';
 import { Routes, RouterModule } from '@angular/router';
 
 
-// #docregion const-routes
+// #docregion const-routes, routes-customers, routes-customers-orders
 const routes: Routes = [
   {
     path: 'customers',
     loadChildren: () => import('./customers/customers.module').then(mod => mod.CustomersModule)
```
alan-agius4: Nit: probably we should change `mod` to `m` just to be inline with the actual code generated by the CLI.

alan-agius4:
@gkalpak thanks for this!

aio/content/examples/lazy-loading-ngmodules/src/app/app-routing.module.ts
gkalpak: Good catch! Feel free to make the change (it can also be on a separate PR - there are probably several examples that need the same change).

gkalpak (approved):

gkalpak:
**merge-assistance:** Global approval for this docs-only change.

PR #34613, opened 2020-01-02
feat: performance improvement for eventListeners

JiaLiPassion:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Since now `DebugNode__PRE_R3__.triggerEventListener` will use `eventListeners` method here https://github.com/angular/angular/blob/f79110c6375d39b148339b7b4af38d99a9c039d9/packages/core/src/debug/debug_node.ts#L414 and it will use the method provided by `zone.js`, the current implementation is not performance efficient because it will loop all properties of the target and use regular expression to check the property name.

This PR will do better.
1. if the `eventName` parameter is not null (which is the most cases), we can directly get the event tasks from target.
2. if the `eventName` parameter is null, we will use the original logic.


mhevery (approved):

mhevery:
[presubmit](http://test/OCL:292584220:BASE:292584247:1580499222718:337f249b)

PR #34202, opened 2019-12-03
fix(ivy): Add `style="{{exp}}"` based interpolation

mhevery:
Fixes #33575

Add support for interpolation in styles as shown:
```
<div style="color: {{exp1}}; width: {{exp2}};">
```

## PR Checklist
Please check if your PR fulfills the following requirements:

- [ ] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] angular.io application / infrastructure changes
- [ ] Other... Please describe:

## What is the current behavior?

Issue Number: N/A

## Does this PR introduce a breaking change?

- [ ] Yes
- [ ] No

PR #34616, opened 2020-01-02
refactor(ivy): Switch styling to reconcile algorithm

mhevery:
`refactor(ivy): Switch styling to reconcile algorithm`

This change deletes old styling code and replaces it with a simplified styling algorithm.

The mental model for the new algorithm is:
- Create a linked list of styling bindings in the order of priority. All styling bindings ere executed in compiled order and than a linked list of bindings is created in priority order.
- Flush the style bindings at the end of `advance()` instruction. This implies that there are two flush events. One at the end of template `advance` instruction in the template. Second one at the end of `hostBindings` `advance` instruction when processing host bindings (if any).
- Each binding instructions effectively updates the string to represent the string at that location. Because most of the bindings are additive, this is a cheap strategy in most cases. In rare cases the strategy requires removing tokens from the styling up to this point. (We expect that to be rare case) Because, the bindings are presorted in the order of priority, it is safe to resume the processing of the concatenated string from the last change binding.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No


packages/common/test/directives/ng_class_spec.ts
```diff
@@ -346,7 +347,12 @@ import {ComponentFixture, TestBed, async} from '@angular/core/testing';
            detectChangesAndExpectClassName('init foo bar');
 
            cmp.strExpr = 'baz';
-           detectChangesAndExpectClassName('init bar baz foo');
+           detectChangesAndExpectClassName(
+               'init bar baz' + (ivyEnabled ?
+                                     '' :    // This is correct behavior because there is no `foo`.
+                                     ' foo'  // This seems to be a bug in VE where `foo` gets
+                                             // persisted even though it was replaced by `baz`.
```
pkozlowski-opensource: @mhevery one could argue that the behaviour is correct in VE since `foo` is part of the `objExpr` bound to `ngClass`.  A user states in `objExpr` that she _wants_ "foo" to be part of the existing classes.... 

It sounds to me like we've got a breaking change here where manipulating `class` bindings influences results of `ngClass`...

I also wonder what would happen if we swap order of bindings, that is  `<div class="init" [ngClass]="objExpr" [class]="strExpr"></div>'` vs. `<div class="init" [class]="strExpr" [ngClass]="objExpr" ></div>'` - would we get the same results? Could we add a test for this case?

packages/core/src/render3/instructions/element.ts
```diff
@@ -87,8 +87,11 @@ export function ɵɵelementStart(
   if (attrs != null) {
     setUpAttributes(renderer, native, attrs);
   }
-  if ((tNode.flags & TNodeFlags.hasInitialStyling) === TNodeFlags.hasInitialStyling) {
-    renderInitialStyling(renderer, native, tNode, false);
+  if (tNode.classes) {
```
pkozlowski-opensource: Could we have a more precise check here and below (`!== null` or whatever the condition is)?

packages/core/src/render3/instructions/element.ts
```diff
@@ -87,8 +87,11 @@ export function ɵɵelementStart(
   if (attrs != null) {
     setUpAttributes(renderer, native, attrs);
   }
-  if ((tNode.flags & TNodeFlags.hasInitialStyling) === TNodeFlags.hasInitialStyling) {
```
pkozlowski-opensource: Since we are not using the `hasInitialStyling` flag here, I wonder if it is still needed?

packages/core/test/util/string_spec.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import {concatStringsWithSpace} from '@angular/core/src/util/stringify';
+
+describe('string', () => {
```
pkozlowski-opensource: Could we name this file `stringify_spec.ts` (to match the source file name) and have `describe` with "concatStringsWithSpace"?

mhevery:
[presubmit](http://test/OCL:289180023:BASE:289180111:1578699968195:ec80fac2)

mhevery:
[presubmit](http://test/OCL:289180023:BASE:289540489:1578960250844:d1923382)

kara (commented):
A few initial comments. Will do another pass tomorrow!

mhevery:
[presubmit](http://test/OCL:289180023:BASE:289666010:1579022371070:2076a017)

mhevery:
[presubmit](http://test/OCL:289180023:BASE:289693126:1579029948541:3c8916fd)

kara (commented):
Another round of comments (trying to release in chunks so you dont have to wait until I'm done to see feedback)

mhevery:
[presubmit](http://test/OCL:289180023:BASE:289797394:1579070758522:85cb44f2)

mhevery:
[presubmit](http://test/OCL:289180023:BASE:289883711:1579112033638:6bbfd922)

mhevery:
[presubmit](http://test/OCL:289180023:BASE:289889048:1579113544749:a8051fc1)

packages/core/src/render3/instructions/styling.ts
```diff
@@ -138,39 +111,9 @@ export function ɵɵclassProp(className: string, value: boolean | null): typeof
...
-  if (firstUpdatePass) {
-    patchConfig(tNode, TNodeFlags.hasStyleMapBindings);
-    patchHostStylingFlag(tNode, isHostStyling(), false);
-  }
-
-  stylingMap(
-      context, tNode, firstUpdatePass, lView, bindingIndex, styles, false, hasDirectiveInput);
+export function ɵɵstyleMap(styles: {[styleName: string]: any} | string | null): typeof ɵɵstyleMap {
+  checkStylingMap(STYLE_MAP_STYLING_KEY, styles, false);
+  return ɵɵstyleMap;
```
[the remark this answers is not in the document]
mhevery: there is no reason why it would chain, I just made it consistent with others. Will revert.

packages/core/src/render3/instructions/styling.ts
```diff
@@ -304,323 +135,178 @@ export function ɵɵstyleMap(styles: {[styleName: string]: any} | NO_CHANGE | nu
...
-    if (isStylingValueDefined(newValue) || !firstUpdatePass) {
-      const inputName: string = isClassBased ? selectClassBasedInputName(tNode.inputs !) : 'style';
-      const inputs = tNode.inputs ![inputName] !;
-      const initialValue = getInitialStylingValue(context);
-      const value = normalizeStylingDirectiveInputValue(initialValue, newValue, isClassBased);
-      setInputsForProperty(lView, inputs, inputName, value);
-      setElementExitFn(stylingApply);
-    }
-    setValue(lView, bindingIndex, newValue);
+* Common code between `ɵɵclassMap` and `ɵɵstyleStyle`.
```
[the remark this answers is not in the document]
mhevery: fixed

packages/core/src/render3/instructions/styling.ts
```diff
@@ -304,323 +135,178 @@ export function ɵɵstyleMap(styles: {[styleName: string]: any} | NO_CHANGE | nu
...
+* @param tStylingMapKey See `STYLE_MAP_STYLING_KEY` and `CLASS_MAP_STYLING_KEY`.
+* @param value binding value.
+* @param isClassBased `true` if `class` change (`false` if `style`)
+*/
+export function checkStylingMap(
+    tStylingMapKey: TStylingMapKey, value: {[className: string]: any} | string | null,
+    isClassBased: boolean): void {
+  const lView = getLView();
+  const tView = lView[TVIEW];
+  const bindingIndex = incrementBindingIndex(2);
```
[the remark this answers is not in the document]
mhevery: comment added.

packages/core/src/render3/interfaces/view.ts
```diff
@@ -471,7 +471,17 @@ export interface TView {
   /** Whether or not this template has been processed in creation mode. */
   firstCreatePass: boolean;
 
-  /** Whether or not the first update for this template has been processed. */
+  /**
+   *  Flag indicating if there is more work to be done on `firstUpdatePass`.
```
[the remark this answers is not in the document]
mhevery: fixed

packages/core/src/render3/interfaces/view.ts
```diff
@@ -471,7 +471,17 @@ export interface TView {
...
+   *
+   * `firstUpdatePass` is used by styling to set up `TData` to contain metadata about the styling
+   * instructions. (Mainly to build up a linked list of styling priority order.)
+   *
+   * Typically this function gets cleared after first execution. If exception is thrown than this
```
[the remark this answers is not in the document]
mhevery: fixed

packages/core/src/render3/state.ts
```diff
@@ -97,17 +96,9 @@ interface LFrame {
   /**
    * Used when processing host bindings.
    */
+  // TODO(misko): this is needed for first template pass only, we should move it out of `LFrame`
```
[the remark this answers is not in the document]
mhevery: Yes! ❤️ deleting code.

packages/core/test/util/string_spec.ts
mhevery: done.

packages/common/test/directives/ng_class_spec.ts
mhevery: Yes VE and Ivy are now aligned.

packages/core/src/render3/instructions/element.ts
mhevery: fixed

packages/core/src/render3/component.ts
```diff
@@ -220,24 +226,24 @@ export function createRootComponent<T>(
...
-
-    setActiveHostElement(null);
+  }
+  const renderer = rootLView[RENDERER];
+  const native = getNativeByTNode(rootTNode, rootLView) as RElement;
+  if (rootTNode.classes) {
+    writeDirectClass(renderer, native, rootTNode.classes);
+  }
+  if (rootTNode.styles) {
+    writeDirectStyle(renderer, native, rootTNode.styles);
```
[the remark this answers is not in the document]
mhevery: reverted

packages/core/src/util/assert.ts
```diff
@@ -14,68 +14,78 @@ import {stringify} from './stringify';
 
 export function assertNumber(actual: any, msg: string) {
   if (typeof actual != 'number') {
-    throwError(msg);
+    throwError(msg, typeof actual, 'number', '===');
```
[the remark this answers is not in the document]
mhevery: https://github.com/angular/angular/pull/34799

AndrewKushnir:
- [VE Presubmit](https://test.corp.google.com/ui#id=OCL:289972406:BASE:289972519:1579138540956:6f524746)
- [Ivy Presubmit](https://test.corp.google.com/ui#id=OCL:289972406:BASE:289972521:1579138540978:3ecf53c6)
- [Ivy Global Presubmit](https://test.corp.google.com/ui#id=OCL:289972406:BASE:289983559:1579144275012:9f3eddf4)

packages/core/test/acceptance/host_binding_spec.ts
```diff
@@ -254,18 +254,17 @@ describe('host bindings', () => {
               }
 
               TestBed.configureTestingModule(
-                  {declarations: [MyApp, ParentDir, ChildDir, SiblingDir]});
+                  {declarations: [MyApp, ParentDir, SiblingDir, ChildDir]});
```
pkozlowski-opensource: Why did the order of declarations change?

packages/core/test/acceptance/host_binding_spec.ts
```diff
@@ -254,18 +254,17 @@ describe('host bindings', () => {
...
 
               const childElement = element.querySelector('div');
 
               // width/height values were set in all directives, but the sub-class directive
-              // (ChildDir)
-              // had priority over the parent directive (ParentDir) which is why its value won. It
-              // also
-              // won over Dir because the SiblingDir directive was evaluated later on.
+              // (ChildDir) had priority over the parent directive (ParentDir) which is why its
+              // value won. It also won over Dir because the SiblingDir directive was evaluated
```
pkozlowski-opensource: question: when you say " SiblingDir directive was evaluated later", what do you mean, exactly? Does it have anything to do with directives declaration order in a module?

packages/core/test/acceptance/integration_spec.ts
```diff
@@ -1033,7 +1033,7 @@ describe('acceptance integration tests', () => {
 
         fixture.componentInstance.value = false;
         fixture.detectChanges();
-        expect(structuralCompEl.getAttribute('class')).toEqual('');
+        expect(structuralCompEl.getAttribute('class')).toBeFalsy();
```
pkozlowski-opensource: question: why do we need to change this test? Is this because of different DOM manipulation?

packages/core/test/acceptance/styling_spec.ts
```diff
@@ -7,16 +7,338 @@
...
+      fixture.detectChanges();
+
+      const [div] = fixture.nativeElement.querySelectorAll('div');
+      expect(getSortedClassName(div)).toEqual('dynamic');
+      expect(getSortedStyle(div)).toEqual('color: blue; width: 100px;');
+    });
+
+    it('should perform interpolation bindings', () => {
+      @Component({
+        // TODO(misko): change `style-x` to `style` once #34202 lands
```
pkozlowski-opensource: question: do we still plan to have #34202 in v9 / later?

packages/core/test/acceptance/styling_spec.ts
```diff
@@ -7,16 +7,338 @@
...
+      }
+
+      @Directive({
+        selector: '[my-host-bindings-2]',
+        host: {'class': 'HOST_STATIC_2', 'style': 'font-family: "c2"'}
+      })
+      class Dir2 {
+      }
+
+      TestBed.configureTestingModule({declarations: [Cmp, Dir1, Dir2]});
```
pkozlowski-opensource: Could we change this test so that directives declaration order in a module and their usage in a template is different? Currently it is not clear what matters as the order of override.

packages/core/test/acceptance/styling_spec.ts
```diff
@@ -7,16 +7,338 @@
...
+      const fixture = TestBed.createComponent(Cmp);
+      fixture.detectChanges();
+
+      const [div] = fixture.nativeElement.querySelectorAll('div');
```
pkozlowski-opensource: nit: could we do `querySelector` here?

packages/core/test/acceptance/styling_spec.ts
```diff
@@ -7,16 +7,338 @@
...
 import {TestBed} from '@angular/core/testing';
+import {getElementClasses, getElementStyles, getSortedClassName, getSortedStyle} from '@angular/core/testing/src/styling';
 import {By, DomSanitizer, SafeStyle} from '@angular/platform-browser';
 import {expect} from '@angular/platform-browser/testing/src/matchers';
 import {ivyEnabled, onlyInIvy} from '@angular/private/testing';
 
 describe('styling', () => {
   beforeEach(ngDevModeResetPerfCounters);
 
+  onlyInIvy('style merging is ivy only feature').describe('apply in prioritization order', () => {
```
pkozlowski-opensource: Could we have `onlyInIvy` on each and every individual test instead of the whole `describe`? Otherwise it is hard to see where we do introduce breaking changes (WRT to VE).

packages/core/test/acceptance/styling_spec.ts
```diff
@@ -7,16 +7,338 @@
...
+      })
+      class Dir extends SuperDir {
+      }
+
+      TestBed.configureTestingModule({declarations: [Cmp, Dir]});
+      const fixture = TestBed.createComponent(Cmp);
+      fixture.detectChanges();
+
+      const [div] = fixture.nativeElement.querySelectorAll('div');
+      expect(getSortedClassName(div)).toEqual('HOST_STATIC STATIC SUPER_STATIC');
```
pkozlowski-opensource: question: sounds like it would be a breaking change as compared to VE behaviour (it wouldn't add `SUPER_STATIC`)

packages/core/test/acceptance/styling_spec.ts
```diff
@@ -7,16 +7,338 @@
...
+      expect(getSortedStyle(div)).toEqual('color: blue; font-family: host; width: 1px;');
+    });
+
+    it('should apply template values in correct order', () => {
+      @Component({
+        template: `
+        <div class="STATIC" style="color: blue"></div>
```
pkozlowski-opensource: nit: look like the case of static class / style combination was exercised in the previous tests so we don't need to repeat it here. Delete?

packages/core/test/acceptance/styling_spec.ts
```diff
@@ -7,16 +7,338 @@
...
+        <div class="STATIC DELETE_MAP_A DELETE_PROP_B"
+             [class]="{foo: true, DELETE_MAP_A: false}"
+             [class.bar]="true"
+             [class.DELETE_PROP_B]="false"></div>
+        <div style="width: 100px; height: 200px: color: red; background-color: yellow"
```
pkozlowski-opensource: nit: could we split this test into 2 separate ones (one for class and another one for style)?

packages/core/test/acceptance/styling_spec.ts
```diff
@@ -7,16 +7,338 @@
...
+
+    it('should work with ngClass/ngStyle', () => {
+      @Component(
+          {template: `<div [ngClass]="['dynamic']" [ngStyle]="{'font-family': 'dynamic'}"></div>`})
+      class Cmp {
+      }
+      TestBed.configureTestingModule({declarations: [Cmp]});
+      const fixture = TestBed.createComponent(Cmp);
+      fixture.detectChanges();
+      fixture.detectChanges();
```
pkozlowski-opensource: question: do we need to detect changes 2x?

packages/core/test/acceptance/styling_spec.ts
```diff
@@ -7,16 +7,338 @@
...
+
+      const [div] = fixture.nativeElement.querySelectorAll('div');
```
pkozlowski-opensource: nit: `querySelectorAll` -> `querySelector`

packages/core/test/acceptance/styling_spec.ts
```diff
@@ -7,16 +7,338 @@
...
+    // `[class]="dynamic"` but never both. This is determined at compile time. Due to locality we
+    // don't know if `[class]` is coming if we see `class` only. So we need to combine the two
+    // This results in slightly different calling sequence, but should result in the same final DOM.
+    expect(div1.getAttribute('shadow-class')).toEqual(ivyEnabled ? 's1 d1' : 'd1');
+
+    expect(div2.className).toEqual('');
+    expect(div2.getAttribute('shadow-class')).toEqual('s2 d2');
+  });
+
+  it('should mimic VE [style] bindings behavior if shadowed', () => {
```
kara: Same here - it doesn't mimic VE behavior here, so we should split it into two tests.

kara (commented):
Another set of comments (still reviewing PR)

pkozlowski-opensource (changes requested):
Could we have `onlyInIvy` on each and every individual test instead of the whole describe? Otherwise it is hard to see where we do introduce breaking changes (WRT to VE).

packages/core/test/acceptance/styling_spec.ts
mhevery: Correct. Enabled for VE and put the assertions in the `ivyEnabled` condition to better demonstrate the difference.

packages/core/test/acceptance/styling_spec.ts
mhevery: fixed

packages/core/test/acceptance/styling_spec.ts
```diff
@@ -7,16 +7,338 @@
...
+      }
+    }
+
+    TestBed.configureTestingModule({declarations: [Cmp, DirectiveShadowsClassInput]});
+    const fixture = TestBed.createComponent(Cmp);
+    fixture.detectChanges();
+
+    const [div1, div2] = fixture.nativeElement.querySelectorAll('div');
+    expect(div1.className).toEqual('s1');
+    // VE has weird behavior where it calls the @Input('class') with either `class="static` ol
```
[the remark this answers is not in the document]
mhevery: fixed

packages/core/test/acceptance/styling_spec.ts
mhevery: nope, deleted!

packages/core/test/acceptance/styling_spec.ts
mhevery: not sure why you think it is duplicate.

packages/core/test/acceptance/styling_spec.ts
mhevery: Split into two tests.

packages/core/test/acceptance/styling_spec.ts
```diff
@@ -7,16 +7,338 @@
...
+      fixture.detectChanges();
+      fixture.detectChanges();
+
+      const [div] = fixture.nativeElement.querySelectorAll('div');
+      expect(getSortedClassName(div)).toEqual('dynamic');
+      expect(getSortedStyle(div)).toEqual('font-family: dynamic;');
+    });
+  });
+
+  it('should mimic VE [class] bindings behavior if shadowed', () => {
```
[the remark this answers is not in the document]
mhevery: I am happy to do the work, but I would like to push back on splitting it like that. 

If you split the tests into two it is hard to know what is the difference in behavior between the two systems. As implemented the difference is very clear with an assertion which is different with explanation. If we split it into two separate tests, it is easy to miss that the two tests are related or to figure out at a glance what the difference in behavior is. 

Could we keep it as is?

packages/core/test/acceptance/styling_spec.ts
```diff
@@ -7,16 +7,338 @@
...
+        this.elementRef.nativeElement.setAttribute('shadow-class', value);
+      }
+    }
+
+    TestBed.configureTestingModule({declarations: [Cmp, DirectiveShadowsClassInput]});
+    const fixture = TestBed.createComponent(Cmp);
+    fixture.detectChanges();
+
+    const [div1, div2] = fixture.nativeElement.querySelectorAll('div');
+    expect(div1.className).toEqual('s1');
```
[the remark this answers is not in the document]
mhevery: I was trying to explain the reason in the comment, but I don't think I succeeded. Could you help me reword it.

The difference in behavior is not by choice as it is side-effect of locality. The issue is when we see static `class` we don't know if `[class]` is going to come later. But we need to make a decision if we should bind or not. Since we don't have the information we are forced to always bind on static `class`. Once we get the the dynamic version `[style]` we could either ignore static and overwrite or combine. If we ignore and overwrite then `ngClass` thinks that static portion is being removed and it deletes it. Which is the worst possible outcome. So instead we merge the two, which allows the `ngClass` to behave without changing.

packages/core/test/acceptance/styling_spec.ts
mhevery: As described above, I would like to keep it, or at least discuss it.

packages/core/test/acceptance/host_binding_spec.ts
pkozlowski-opensource: I see. Longer it might make sense to have it in-the-template order, but this is a separate discussion.

packages/core/test/acceptance/integration_spec.ts
pkozlowski-opensource: Yeh, this is what I was suspecting, thnx!

packages/core/test/acceptance/styling_spec.ts
```diff
@@ -7,16 +7,373 @@
...
+import {getElementClasses, getElementStyles, getSortedClassName, getSortedStyle} from '@angular/core/testing/src/styling';
 import {By, DomSanitizer, SafeStyle} from '@angular/platform-browser';
 import {expect} from '@angular/platform-browser/testing/src/matchers';
 import {ivyEnabled, onlyInIvy} from '@angular/private/testing';
 
 describe('styling', () => {
   beforeEach(ngDevModeResetPerfCounters);
 
+  describe('apply in prioritization order', () => {
+    onlyInIvy('style merging is ivy only feature').it('should perform static bindings', () => {
```
pkozlowski-opensource: I don't think that this test should be `onlyInIvy`, we are not exercising any prioritisation here.

packages/core/test/acceptance/styling_spec.ts
```diff
@@ -7,16 +7,373 @@
...
+
+      TestBed.configureTestingModule({declarations: [Cmp]});
+      const fixture = TestBed.createComponent(Cmp);
+
+      const [staticDiv] = fixture.nativeElement.querySelectorAll('div');
+      expect(getSortedClassName(staticDiv)).toEqual('STATIC');
+      expect(getSortedStyle(staticDiv)).toEqual('color: blue;');
+    });
+
+    onlyInIvy('style merging is ivy only feature').it('should perform prop bindings', () => {
```
pkozlowski-opensource: I don't think that this test should be ivy only, things should work the same in VE if I'm not mistaken.

packages/core/test/acceptance/styling_spec.ts
```diff
@@ -7,16 +7,373 @@
...
+      fixture.detectChanges();
+
+      const div = fixture.nativeElement.querySelector('div');
+      expect(getSortedClassName(div)).toEqual('dynamic');
+      expect(getSortedStyle(div)).toEqual('color: blue; width: 100px;');
+    });
+
+    onlyInIvy('style merging is ivy only feature').it('should perform map bindings', () => {
+      @Component({
+        template: `<div [class]="{dynamic: true}"
```
pkozlowski-opensource: question: we couldn't bind like this in VE to `class`, right? My understanding is that in VE binding to `[class]` was an equivalent of binding to `className` (so the object literal would be simply stringified). I'm assuming that this is a deliberate change in ivy?

pkozlowski-opensource (commented):
Another batch of reviews - this time mostly for the existing acceptance tests (packages/core/test/acceptance/styling_spec.ts)

packages/core/test/acceptance/styling_spec.ts
```diff
@@ -7,16 +7,373 @@
...
+
+    TestBed.configureTestingModule({declarations: [Cmp, DirectiveShadowsClassInput]});
+    const fixture = TestBed.createComponent(Cmp);
+    fixture.detectChanges();
+
+    const [div1, div2] = fixture.nativeElement.querySelectorAll('div');
+    expect(div1.className).toEqual('s1');
+    // VE has weird behavior where it calls the @Input('class') with either `class="static` or
+    // `[class]="dynamic"` but never both. This is determined at compile time. Due to locality we
+    // don't know if `[class]` is coming if we see `class` only. So we need to combine the two
```
pkozlowski-opensource: For the future consideration / helping to re-word the comment / stating the obvious: this behaviour is _not_ specific to `class` / `[class]` but applies to _any_ property binding that has a corresponding attribute... This is another manifestation of the syntax / "shadowing rules" where the left-hand side doesn't identify binding target _precisely enough_. In this sense VE is not particularly "weird" - it is just how the syntax works / can be abused atm....

packages/core/test/acceptance/styling_spec.ts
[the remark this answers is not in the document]
pkozlowski-opensource: tl;dr;
*  I would tend to side with Kara in saying that the test name doesn't match asserts so we should at least rename the test; 
* TBH I would also preferred to have it split in 2 tests;

Longer explanation: I can see and don't see those 2 cases as related (!). I'm saying this because when one de-sugars interpolation, we end up with:
* `<div class="s1" [class]=" 'd1' " dir-shadows-class-input>` (a static attribute _and_ a property binding)
* `<div [class]="'s2 ' + 'd2'" dir-shadows-class-input>`(just property binding).

The way I'm looking at it there are 2 use-cases:
*  a static attribute _and_ a property binding
* a property binding (that happens to be in a form of interpolation)

I _think_ that we are having this debate because one can see those cases as related (or not!) depending on how we _think_ about them - it is true that those 2 cases are similar syntax-wise (_visually_ look the same - in both cases `class="...."` shows up in a template but _semantically_ very different! 

At the risk of sounding as a breaking record I would say that this is another manifestation of syntax problems where we write visually-similar things to mean different concepts.

packages/core/test/acceptance/styling_spec.ts
pkozlowski-opensource: This use-case is exercised in several other places (ex.: https://github.com/angular/angular/pull/34616/files#diff-0483f62e84f992bd169dee41626776e1R23)

packages/core/test/acceptance/styling_spec.ts
```diff
@@ -7,16 +7,373 @@
...
+          const fixture = TestBed.createComponent(Cmp);
+          fixture.detectChanges();
+
+          const [staticDiv, styleDiv] = fixture.nativeElement.querySelectorAll('div');
+          expect(getSortedStyle(staticDiv)).toEqual('color: blue;');
+          expect(getSortedStyle(styleDiv))
+              .toEqual('background-color: yellow; color: blue; width: 110px;');
+        });
+
+    onlyInIvy('style merging is ivy only feature').it('should work with ngClass/ngStyle', () => {
```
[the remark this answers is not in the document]
mhevery: yes works, removed `onlyInIvy`

packages/core/test/acceptance/styling_spec.ts
```diff
@@ -7,16 +7,373 @@
...
+      @Component(
+          {template: `<div [ngClass]="['dynamic']" [ngStyle]="{'font-family': 'dynamic'}"></div>`})
+      class Cmp {
+      }
+      TestBed.configureTestingModule({declarations: [Cmp]});
+      const fixture = TestBed.createComponent(Cmp);
+      fixture.detectChanges();
+      // fixture.detectChanges();
```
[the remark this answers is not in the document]
mhevery: removed

packages/core/test/acceptance/styling_spec.ts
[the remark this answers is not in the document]
mhevery: > * I would tend to side with Kara in saying that the test name doesn't match asserts so we should at least rename the test;
Test renamed
> * TBH I would also preferred to have it split in 2 tests;
Test split

packages/core/test/acceptance/styling_spec.ts
mhevery: It is specific to `class` in the sense that we do the concatenation only in `class`.

It is unclear what is my action item here.

packages/core/test/acceptance/styling_spec.ts
mhevery: Split into two, changed name per discussion.

packages/core/test/acceptance/styling_spec.ts
```diff
@@ -496,23 +849,18 @@ describe('styling', () => {
          }
        }
 
-       // Ivy does an extra `[class]` write with a falsy value since the value
-       // is applied during creation mode. This is a deviation from VE and should
-       // be (Jira Issue = FW-1467).
-       let totalWrites = ivyEnabled ? 1 : 0;
```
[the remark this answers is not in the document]
mhevery: no problem, renamed.

packages/core/test/acceptance/styling_spec.ts
```diff
@@ -829,6 +1190,14 @@ describe('styling', () => {
             expect(element.style.opacity).toBeFalsy();
```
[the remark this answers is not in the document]
mhevery: - yes this is technically a breaking change.
- you can think of `null` vs `undefined` as a new feature. It is being introduced now, because doing it later will be harder, because as people will start using priorities they will use `null` or `undefined` interchangeably which will make introducing this feature harder.  
- Let's discuss on VC with @kara

packages/core/test/acceptance/styling_spec.ts
```diff
@@ -867,17 +1236,17 @@ describe('styling', () => {
         const element = fixture.nativeElement.querySelector('div');
         expect(element.style.width).toEqual('100px');
 
-        fixture.componentInstance.w0 = null;
+        fixture.componentInstance.w0 = undefined;
         fixture.detectChanges();
 
-        expect(element.style.width).toEqual('200px');
+        expect(element.style.width).toEqual('300px');
```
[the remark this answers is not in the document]
mhevery: - Let's discuss
- What is wrong with the assert? (Test pass, looks right to me)

packages/core/test/acceptance/styling_spec.ts
```diff
@@ -958,8 +1320,8 @@ describe('styling', () => {
             expect(element.style.height).toEqual('100px');
             expect(element.style.fontSize).toEqual('50px');
 
-            // there is no need to flush styling since the styles are applied directly
-            expect(ngDevMode !.flushStyling).toEqual(0);
+            // once for the template flush and again for the host bindings
+            expect(ngDevMode !.flushStyling).toEqual(2);
```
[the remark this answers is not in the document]
mhevery: - It is not clear to me which `onlyInIvy` do you want me to remove (Should I remove all and see which tests pass?)
- I thought per earlier comment you did not like when I put conditional asserts between VE/Ivy. 

Sorry I am just confused as to what I should do here.

packages/core/test/acceptance/styling_spec.ts
```diff
@@ -2486,25 +2714,389 @@ describe('styling', () => {
...
+        expect(root.className).toEqual('');
+
+        fixture.componentInstance.classes = '1 2 3';
+        fixture.detectChanges();
+        expect(root.className.split(/\s+/).sort().join(' ')).toEqual('1 2 3');
+
+        root.classList.add('0');
+        expect(root.className.split(/\s+/).sort().join(' ')).toEqual('0 1 2 3');
+
+        // TODO(pk): why ???
```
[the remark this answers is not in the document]
mhevery: removed

packages/core/test/acceptance/styling_spec.ts
```diff
@@ -2486,25 +2714,389 @@ describe('styling', () => {
...
+    const fixture = TestBed.createComponent(MyComp);
+    fixture.detectChanges();
+
+    const classList = (fixture.nativeElement.querySelector('div') as HTMLDivElement).classList;
+    expect(classList.contains('fooBar')).toBeTruthy();
+    expect(classList.contains('barFoo')).toBeTruthy();
+  });
+
+  onlyInIvy('[style] bindings are ivy only')
+      .it('should convert camelCased style property names to snake-case', () => {
```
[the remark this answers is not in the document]
mhevery: As discussed this came from previews PR from @matsko. I think this is a good feature and we should keep it, but there is no particular reason why this feature has to be included.

packages/core/src/render3/instructions/styling.ts
```diff
@@ -304,323 +136,221 @@ export function ɵɵstyleMap(styles: {[styleName: string]: any} | NO_CHANGE | nu
...
   }
-
-  // we check for this in the instruction code so that the context can be notified
-  // about prop or map bindings so that the direct apply check can decide earlier
-  // if it allows for context resolution to be bypassed.
-  if (firstUpdatePass) {
-    patchConfig(tNode, TNodeFlags.hasClassMapBindings);
-    patchHostStylingFlag(tNode, isHostStyling(), true);
+  if (value !== NO_CHANGE && bindingUpdated(lView, bindingIndex, value)) {
+    lView[bindingIndex] = value;
```
kara: Setting the value here shouldn't be necessary, given that `bindingUpdated` will already set `lView[bindingIndex]` to the `value`

```suggestion
```

packages/core/src/render3/instructions/styling.ts
```diff
@@ -304,323 +136,221 @@ export function ɵɵstyleMap(styles: {[styleName: string]: any} | NO_CHANGE | nu
...
-    }
-
-    setElementExitFn(stylingApply);
-  }
-
-  if (ngDevMode) {
-    isClassBased ? ngDevMode.classMap : ngDevMode.styleMap++;
-    if (valueHasChanged) {
-      isClassBased ? ngDevMode.classMapCacheMiss : ngDevMode.styleMapCacheMiss++;
+      lView[bindingIndex] = value;
```
kara: This set should be unnecessary because `bindingUpdated` already does this

```suggestion
```

packages/core/src/render3/instructions/styling.ts
```diff
@@ -304,323 +136,221 @@ export function ɵɵstyleMap(styles: {[styleName: string]: any} | NO_CHANGE | nu
...
- * it is also called from `ɵɵclassMapInterpolate`.
+ * @param prop property name.
+ * @param value binding value.
+ * @param suffixOrSanitizer suffix or sanitization function
+ * @param isClassBased `true` if `class` change (`false` if `style`)
  */
-export function classMapInternal(
-    elementIndex: number, classes: {[className: string]: any} | NO_CHANGE | string | null): void {
+export function checkStylingProperty(
+    prop: string, value: string | number | SafeValue | null,
```
kara: It seems like `value` should have a type that includes `NO_CHANGE`, given line 172. Seems like style prop interpolations would sometimes pass that through.

packages/core/src/render3/instructions/styling.ts
```diff
@@ -304,323 +136,221 @@ export function ɵɵstyleMap(styles: {[styleName: string]: any} | NO_CHANGE | nu
...
+    tView: TView, prop: string | TStylingMapKey,
+    suffixOrSanitization: SanitizerFn | string | null | undefined, bindingIndex: number,
+    isClassBased: boolean): void {
+  ngDevMode && assertFirstUpdatePass(tView);
+  const tData = tView.data;
+  if (tData[bindingIndex + 1] === null) {
+    // The above check is necessary because we don't clear first update pass until first successful
+    // (ne exception) template execution. This prevents the styling instruction from double adding
+    // itself to the list.
+    const tNode = tData[getSelectedIndex() + HEADER_OFFSET] as TNode;
```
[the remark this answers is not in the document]
kara: I'm OK with keeping it (esp since now that I look at it again, it means we can avoid the function call on subsequent template passes). But if we're going to break the convention, can we add a brief comment here so others don't refactor this code later by mistake?

packages/core/src/render3/instructions/styling.ts
```diff
@@ -304,323 +136,221 @@ export function ɵɵstyleMap(styles: {[styleName: string]: any} | NO_CHANGE | nu
...
+    if (hasStylingInputShadow(tNode, isClassBased) && typeof prop === 'object' &&
+        !isInHostBindings(tView, bindingIndex)) {
```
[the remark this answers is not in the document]
kara: Sure, same deal - let's add a brief comment that this is intentional to prevent later refactoring

packages/core/src/render3/instructions/styling.ts
```diff
@@ -304,323 +136,221 @@ export function ɵɵstyleMap(styles: {[styleName: string]: any} | NO_CHANGE | nu
...
+        assertLessThan(styleLastWrittenValueIndex, lView.length, 'Reading past end of LView');
+    const lastValue: string|NO_CHANGE = lView[styleLastWrittenValueIndex];
+    const newValue = flushStyleBinding(tData, tNode, lView, styleBindingIndex, false);
+    if (lastValue !== newValue) {
+      if (tNode.type === TNodeType.Element) {
+        writeAndReconcileStyle(
+            renderer, element, lastValue === NO_CHANGE ? tNode.styles || '' : lastValue as string,
+            newValue);
+      }
+      lView[styleLastWrittenValueIndex] = newValue;
```
kara: Lines 334 -352 bear a lot of resemblance to lines 312-330. I suspect that we could share more code between styles & classes by refactoring into a shared function. WDYT?

packages/core/test/acceptance/change_detection_spec.ts
```diff
@@ -1435,16 +1435,15 @@ describe('change detection', () => {
        });
 
     it('should include style prop name in case of style binding', () => {
-      const message = ivyEnabled ?
-          `Previous value for 'style.color': 'red'. Current value: 'green'` :
-          `Previous value: 'color: red'. Current value: 'color: green'`;
+      const message = ivyEnabled ? `Previous value for 'color': 'red'. Current value: 'green'` :
```
kara: Why revert the `style.color` change? I think it makes easier to map what you wrote in the template.

packages/core/test/acceptance/styling_spec.ts
```diff
@@ -7,16 +7,471 @@
...
+            this.elementRef.nativeElement.setAttribute('shadow-class', value);
+          }
+        }
+
+        TestBed.configureTestingModule({declarations: [Cmp, DirectiveShadowsClassInput]});
+        const fixture = TestBed.createComponent(Cmp);
+        fixture.detectChanges();
+
+        const [div1, div2] = fixture.nativeElement.querySelectorAll('div');
+        expect(div1.className).toEqual('s1');
```
kara: It doesn't let me respond to the thread anymore, so to respond to your Q here: https://github.com/angular/angular/pull/34616#discussion_r368177121

Do you mean the comment below this assertion (line 248)? I think I missed it because it was below - can we move it up a line for visibility?

In terms of the actual behavior, I think we are still crossing wires somewhere. In your comment in this thread and the comment on line 248, you mention that the static class is being "combined" with the dynamic class, but the assertion says that only "s1" ends up on the DOM node. If it were combined, I would expect to see "s1 d1" (with the dynamic class as well). Can you clarify a bit? Maybe this is a good one for VC?

packages/core/src/render3/styling/style_binding_list.ts
```diff
@@ -197,9 +202,7 @@ import {getLastParsedKey, parseClassName, parseClassNameNext, parseStyle, parseS
 export function insertTStylingBinding(
```
kara: It looks like we are keeping the "head" marker pointing to the "template head" and inserting new host bindings into the middle (in front of the template head but behind previous host bindings). I take it we are not prepending host bindings to maintain their priority order (last directive has highest priority)?

packages/core/test/render3/styling_next/style_binding_list_spec.ts
```diff
@@ -438,34 +437,47 @@ describe('TNode styling linked list', () => {
...
+      fixture.setBinding(0, 'red');
+      fixture.setBinding(1, undefined);
+      expect(fixture.flush(0)).toEqual('color: red;');
+    });
+
+    it('should ignore falsy values', () => {
+      const fixture = new StylingFixture([['color']], false);
+      fixture.setBinding(0, null);
+      expect(fixture.flush(0)).toEqual('');
     });
```
kara: Nit: it "should remove falsy values" and it should "ignore falsy values". the test description names seem to conflict. Can we make the descriptions a bit more specific?

packages/core/src/render3/instructions/styling.ts
```diff
@@ -304,323 +136,221 @@ export function ɵɵstyleMap(styles: {[styleName: string]: any} | NO_CHANGE | nu
...
+  const classBindingIndex = getClassBindingChanged();
+  if (classBindingIndex > 0) {
+    const classLastWrittenValueIndex = getTStylingRangeTail(tNode.classBindings) + 1;
+    ngDevMode &&
+        assertGreaterThan(
+            classLastWrittenValueIndex, 1,
+            'Ignoring `class` binding because there is no `class` metadata associated with the element. ' +
+                '(Was exception thrown during `firstUpdatePass` which prevented the metadata creation?)');
+    ngDevMode &&
+        assertLessThan(classLastWrittenValueIndex, lView.length, 'Reading past end of LView');
```
kara: Nit: seems like we could use the "data in range" function we already have?

```suggestion
        assertDataInRange(lView, classLastWrittenValueIndex);
```

packages/core/src/render3/state.ts
```diff
@@ -120,6 +105,18 @@ interface LFrame {
...
   currentQueryIndex: number;
+
+  /**
+   * Stores the index of the style binding which changed first.
+   *
+   * A change in styling binding implies that all bindings starting with this index need to be
+   * recomputed. See: `flushStylingOnElementExit` and `markStylingBindingDirty` functions for
+   * details.
+   *
+   * If this value is set than `flushStylingOnElementExit` needs to execute during the `advance`
```
kara:
```suggestion
   * If this value is set then `flushStylingOnElementExit` needs to execute during the `advance`
```

packages/core/src/render3/styling/style_binding_list.ts
```diff
@@ -355,6 +359,8 @@ function markDuplicates(
       // if we are a Map (and we have statics) we must assume duplicate
       foundDuplicate = true;
     } else if (staticValues != null) {
+      // If we found non-map than we iterate over its keys to determine if any of them match ours
```
kara:
```suggestion
      // If we found non-map then we iterate over its keys to determine if any of them match ours
```

packages/core/src/render3/styling/style_binding_list.ts
```diff
@@ -392,18 +399,21 @@ export function flushStyleBinding(
   // value and look up the previous concatenation as a starting point going forward.
   const lastUnchangedValueIndex = getTStylingRangePrev(tStylingRangeAtIndex);
   let text = lastUnchangedValueIndex === 0 ?
-      ((isClassBinding ? tNode.classes : tNode.styles) as string) :
-      lView[lastUnchangedValueIndex + 1] as string;
+      (isClassBinding ? tNode.classes : tNode.styles) :
+      lView[lastUnchangedValueIndex + 1] as string | NO_CHANGE;
+  if (text === null || text === NO_CHANGE) text = '';
+  ngDevMode && assertString(text, 'Last unchanged value value should be a string');
```
kara:
```suggestion
  ngDevMode && assertString(text, 'Last unchanged value should be a string');
```

kara (commented):
Two more typos

packages/core/test/acceptance/styling_spec.ts
mhevery: updated text to make ti more clear. Please have a look. (happy to VC as well)

packages/core/test/render3/styling_next/style_differ_spec.ts
```diff
@@ -31,6 +30,13 @@ describe('style differ', () => {
       expectParseValue(':  text1 text2  ;🛑').toBe('text1 text2');
     });
 
+    it('should parse empty vale', () => {
```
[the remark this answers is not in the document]
mhevery: removed in next PR, ignoring.

packages/core/src/render3/styling/style_binding_list.ts
mhevery: fixed

packages/core/src/render3/styling/style_binding_list.ts
mhevery: I don't fully follow your question.

The items get inserted into the list in a way which results bindings to be sorted in priority order. 

This means that component/directive get inserted in front of e "head" so that the final list has the desired order.

Should we VC?

packages/core/src/render3/styling/style_binding_list.ts
mhevery: fixed

packages/core/test/render3/styling_next/style_binding_list_spec.ts
mhevery: updated text.

packages/core/src/render3/state.ts
mhevery: fixed

packages/core/test/acceptance/change_detection_spec.ts
kara: Yeah, I'd agree. Thanks for explaining.

packages/core/src/render3/styling/style_binding_list.ts
kara: Probably easier to discuss over VC, yeah. I was asking why we insert new host bindings behind previous host bindings but before template bindings (instead of just prepending to the very beginning)

mhevery:
Merged

PR #34617, opened 2020-01-02
fix(ivy): Ensure ngProjectAs marker name appears at even attribute index

atscott:
The `getProjectAsAttrValue` in `node_selector_matcher` finds the
`ProjectAs` marker and then additionally checks that the marker appears in
an even index of the node attributes because "attribute names are stored
at even indexes". This is true for "regular" attribute bindings but
classes, styles, bindings, templates, and i18n do not necessarily follow
this rule because there can be an uneven number of them, causing the
next "special" attribute "name" to appear at an odd index. To address
this issue, ensure `ngProjectAs` is placed right after "regular"
attributes.

### DIFF
--- a/packages/compiler/src/render3/view/template.ts
+++ b/packages/compiler/src/render3/view/template.ts
@@ -599,14 +599,9 @@ export class TemplateDefinitionBuilder implements t.Visitor<void>, LocalResolver
       attributes.push(...getAttributeNameLiterals(attr.name), o.literal(attr.value));
     });
 
-    // Keep ngProjectAs next to the other name, value pairs so we can verify that we match
-    // ngProjectAs marker in the attribute name slot.
-    if (ngProjectAsAttr) {
-      attributes.push(...getNgProjectAsLiteral(ngProjectAsAttr));
-    }
     // add attributes for directive and projection matching purposes
     attributes.push(...this.prepareNonRenderAttrs(
-        allOtherInputs, element.outputs, stylingBuilder, [], i18nAttrs));
+        allOtherInputs, element.outputs, stylingBuilder, [], i18nAttrs, ngProjectAsAttr));
     parameters.push(this.addAttrsToConsts(attributes));
 
     // local refs (ex.: <div #foo #bar="baz">)
@@ -876,13 +871,9 @@ export class TemplateDefinitionBuilder implements t.Visitor<void>, LocalResolver
       }
       attrsExprs.push(asLiteral(attr.name), asLiteral(attr.value));
     });
-    // Keep ngProjectAs next to the other name, value pairs so we can verify that we match
-    // ngProjectAs marker in the attribute name slot.
-    if (ngProjectAsAttr) {
-      attrsExprs.push(...getNgProjectAsLiteral(ngProjectAsAttr));
-    }
     attrsExprs.push(...this.prepareNonRenderAttrs(
-        template.inputs, template.outputs, undefined, template.templateAttrs, undefined));
+        template.inputs, template.outputs, undefined, template.templateAttrs, undefined,
+        ngProjectAsAttr));
     parameters.push(this.addAttrsToConsts(attrsExprs));
 
     // local refs (ex.: <ng-template #foo>)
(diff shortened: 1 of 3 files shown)


packages/compiler/src/render3/view/template.ts
```diff
@@ -1311,10 +1319,6 @@ export class TemplateDefinitionBuilder implements t.Visitor<void>, LocalResolver
       templateAttrs.forEach(attr => addAttrExpr(attr.name));
     }
 
-    if (ngProjectAsAttr) {
```
crisbeto: Maybe we can keep the this logic inside `prepareNonRenderAttrs`, by moving the `push` further up? That way we can keep all the logic for the non-render attributes together.

atscott: I was waffling back and forth on this one. I moved it out because having it inside sort of makes some assumptions about the context in which the function is called that the function doesn't necessarily know about. It felt like the only real way to guarantee placement right after the other k/v attribute pairs was to have the code next to each other.

I did move it back though to keep it with the other non-render attributes.

crisbeto (approved):
LGTM

packages/compiler/src/render3/view/template.ts
AndrewKushnir: As an option, we can refactor the code a bit and extend the function in `template.ts` that assembles attribute array (which is currently called `prepareNonRenderAttrs`, but we may need to find a better name during refactoring) to also accepts set of attributes (that should be at the very beginning of an array) as an argument. In this case, the right order/shape of the attributes array would be encapsulated within the `prepareNonRenderAttrs` function. We can reuse this function in the `visitContent` function as well for consistency.

atscott: Refactored to single function based on suggestion from @AndrewKushnir

packages/compiler/src/render3/view/template.ts
```diff
@@ -1240,24 +1223,37 @@ export class TemplateDefinitionBuilder implements t.Visitor<void>, LocalResolver
...
    *
    * Note that this function will fully ignore all synthetic (@foo) attribute values
    * because those values are intended to always be generated as property instructions.
    */
-  private prepareNonRenderAttrs(
-      inputs: t.BoundAttribute[], outputs: t.BoundEvent[], styles?: StylingBuilder,
-      templateAttrs: (t.BoundAttribute|t.TextAttribute)[] = [],
-      i18nAttrs: (t.BoundAttribute|t.TextAttribute)[] = [],
-      ngProjectAsAttr?: t.TextAttribute): o.Expression[] {
+  private getAttributeExpressions(
```
AndrewKushnir: Thanks for this refactoring @atscott, it looks great!

I think we should be able to use this function in the `visitContent` function as well, WDYT?

packages/compiler/src/render3/view/template.ts
```diff
@@ -1240,24 +1223,37 @@ export class TemplateDefinitionBuilder implements t.Visitor<void>, LocalResolver
...
+      renderAttributes: t.TextAttribute[], inputs: t.BoundAttribute[], outputs: t.BoundEvent[],
+      styles?: StylingBuilder, templateAttrs: (t.BoundAttribute|t.TextAttribute)[] = [],
+      i18nAttrs: (t.BoundAttribute|t.TextAttribute)[] = [], ): o.Expression[] {
```
AndrewKushnir: Extra comma at the end of the argument list.

packages/core/test/acceptance/content_spec.ts
```diff
@@ -955,6 +955,46 @@ describe('projection', () => {
...
+    class Card {
+    }
+
+    @Component({
+      selector: 'card-with-title',
+      template: `
+        <card>
+         <h1 [color]="'red'" [margin]="10" ngProjectAs="[card-title]">Title</h1>
+         <div style="font-color: blue;" ngProjectAs="[card-content]">content</div>
+         <div [color]="'blue'" ngProjectAs="[card-footer]">footer</div>
```
AndrewKushnir: Could you plz add an attribute with a namespace, so we make sure projectAs attribute is always at the right location?

AndrewKushnir (changes requested):

packages/core/test/acceptance/content_spec.ts
atscott: Done.

packages/compiler/src/render3/view/template.ts
atscott: Done.

packages/compiler/src/render3/view/template.ts
atscott: Done.

AndrewKushnir (approved):
LGTM 👍

alxhub (approved):

atscott:
[Ivy Presubmit](https://test.corp.google.com/ui#id=OCL:287998244:BASE:288341489:1578339247698:a76daa54)

PR #34649, opened 2020-01-06
Batch of Template Type-checking fixes

alxhub:
This is a batch PR fixing several template type-checking issues that were discovered in the g3 codebase:

* Fixes an issue with bindings which map to more than one field. `@Input('x') x; @Input('x') y;` for example (amazingly, this is not an error).
* Fixes an issue with repeated bindings to fields (`[x]="..." [x]="..."`).
* Corrects the precedence of `as any` casts generated by the template type-checker in certain cases (`a ? b : c as any` was previously generated where the intent was `(a ? b : c) as any`).
* Disables the aliasing system when generating template type-checking code, as it won't work.
* Adds a new strictness flag for context generics, so `Component<T>` has `T` inferred as `any` unless `strictTemplates` is enabled.


packages/compiler-cli/test/ngtsc/template_typecheck_spec.ts
```diff
@@ -174,6 +174,32 @@ export declare class AnimationEvent {
           .toEqual(`Argument of type 'string' is not assignable to parameter of type 'number'.`);
     });
 
+    it('should support one input property mapping to multiple fields', () => {
+      env.write('test.ts', `
+        import {Component, Directive, Input, NgModule} from '@angular/core';
+
+        @Directive({
+          selector: '[dir]',
+          inputs: ['propertyName'],
```
JoostK: Please see https://github.com/angular/angular/pull/34482 for a suggestion on how to make this setup more clear.

packages/compiler-cli/src/ngtsc/typecheck/src/ts_util.ts
```diff
@@ -9,7 +9,46 @@
...
+  ts.SyntaxKind.NumericLiteral,
+  ts.SyntaxKind.TrueKeyword,
+  ts.SyntaxKind.FalseKeyword,
+  ts.SyntaxKind.NullKeyword,
+  ts.SyntaxKind.UndefinedKeyword,
+]);
+
 export function tsCastToAny(expr: ts.Expression): ts.Expression {
+  // Wrap `expr` in parentheses if needed (see `SAFE_TO_CAST_WITHOUT_PARENS` above).
+  if (!SAFE_TO_CAST_WITHOUT_PARENS.has(expr.kind)) {
```
JoostK: Why go through the trouble of avoiding the parens if not necessary? There's no code side concern in the TCB generated code and they already littered with parens because of the positional markers.

Also, I'm surprised TypeScript doesn't handle this properly in the first place, as it can do for other `ts.Node` factory functions. Perhaps they have a bug in their precedence logic.

packages/compiler-cli/src/ngtsc/typecheck/src/host.ts
```diff
@@ -19,12 +19,18 @@ export class TypeCheckProgramHost implements ts.CompilerHost {
...
+  readonly resolveModuleNames?: ts.CompilerHost['resolveModuleNames'];
+
   constructor(sfMap: Map<string, ts.SourceFile>, private delegate: ts.CompilerHost) {
     this.sfMap = sfMap;
 
     if (delegate.getDirectories !== undefined) {
       this.getDirectories = (path: string) => delegate.getDirectories !(path);
     }
+
+    if (delegate.resolveModuleNames !== undefined) {
```
JoostK: Oh no, not this again 😱 At some point we should definitely clean up all these hosts so they act as a proper proxy, as this is super error prone :(

packages/compiler-cli/test/ngtsc/template_typecheck_spec.ts
```diff
@@ -1175,6 +1175,72 @@ export declare class AnimationEvent {
...
+          }
+         `);
+
+         // 'beta' re-exports AlphaModule from alpha.
+         env.write('beta.d.ts', `
+          import {ɵɵNgModuleDefWithMeta} from '@angular/core';
+          import {AlphaModule} from './alpha';
+
+          export declare class BetaModule {
+            static ɵmod: ɵɵNgModuleDefWithMeta<AlphaModule, never, never, [typeof AlphaModule]>;
```
JoostK:
```suggestion
            static ɵmod: ɵɵNgModuleDefWithMeta<BetaModule, never, never, [typeof AlphaModule]>;
```

aio/content/guide/template-typecheck.md
```diff
@@ -120,6 +120,7 @@ In case of a false positive like these, there are a few options:
 |`strictDomLocalRefTypes`|Whether local references to DOM elements will have the correct type. If disabled `ref` will be of type `any` for `<input #ref>`.|
 |`strictOutputEventTypes`|Whether `$event` will have the correct type for event bindings to component/directive an `@Output()`, or to animation events. If disabled, it will be `any`.|
 |`strictDomEventTypes`|Whether `$event` will have the correct type for event bindings to DOM events. If disabled, it will be `any`.|
+|`strictContextGenerics`|Whether the type parameters of generic components will be inferred correctly (including any generic bounds). If disabled, any type parameters will be any will be `any`.|
```
JoostK:
```suggestion
|`strictContextGenerics`|Whether the type parameters of generic components will be inferred correctly (including any generic bounds). If disabled, any type parameters will be `any`.|
```

packages/compiler-cli/src/ngtsc/typecheck/src/type_check_block.ts
```diff
@@ -925,12 +925,20 @@ class Scope {
...
   // Check if the component is generic, and pass generic type parameters if so.
   if (node.typeParameters !== undefined) {
-    typeArguments =
-        node.typeParameters.map(param => ts.createTypeReferenceNode(param.name, undefined));
+    if (useGenericType) {
+      typeArguments =
+          node.typeParameters.map(param => ts.createTypeReferenceNode(param.name, undefined));
+    } else {
+      typeArguments = [];
+      for (let i = 0; i < node.typeParameters.length; i++) {
```
JoostK: No `.map` here? :-)

packages/compiler-cli/src/transformers/api.ts
```diff
@@ -213,6 +213,19 @@ export interface CompilerOptions extends ts.CompilerOptions {
    */
   strictDomEventTypes?: boolean;
 
+  /**
+   * Whether to include the generic type of components when type-checking the template.
+   *
+   * If a component has no generic type parameters, this setting has no effect.
```
JoostK:
```suggestion
   * If no components have generic type parameters, this setting has no effect.
```

JoostK (changes requested):
Some minor things that need some cleaning up.

packages/compiler-cli/src/ngtsc/typecheck/src/ts_util.ts
alxhub: Readability. Template type-checking code is already kind of tricky to follow. I want to try to keep it as readable as possible.

Ideally, I'd like to even eliminate some of the unnecessary parens for the positional markers, as well as add more descriptive variable names and even comments. We should have a compiler feature to dump this code for debugging purposes.

packages/compiler-cli/src/ngtsc/typecheck/src/host.ts
alxhub: 👍

JoostK (approved):
LGTM, but I'd like to see https://github.com/angular/angular/pull/34649/files#r363443363 addressed.

alxhub:
Done :)

alxhub:
[Presubmit](http://test/OCL:290821930:BASE:290844186:1579651524364:2047ee29)

PR #34288, opened 2019-12-06
Incrementality for the win!

alxhub:

PR #34652, opened 2020-01-06
fix(ngcc): do not attempt compilation when analysis fails

### LINKED ISSUE #34500: ngcc: packages that fail with errors causes a crash
# 🐞 bug report

### Affected Package

### Is this a regression?

Yes, the previous version in which this bug was not present was: 9.0.0-rc.6

### Description

When ngcc processes a package that fails to compile, a fatal error that crashes the compiler occurs.

## 🔬 Minimal Reproduction

For example `angular-tree-component` fails to compile:
https://app.circleci.com/jobs/github/angular/ngcc-validation/2459/parallel-runs/1?filterBy=ALL

## 🔥 Exception or Error
<pre><code>
ERROR in Cannot read property 'injectorImports' of undefined
</code></pre>

## 🌍  Your Environment

**Angular Version:**
<pre><code>
Angular: 9.0.0-rc.7
</code></pre>

**Anything else relevant?**
n/a

JoostK:
In #34288, ngtsc was refactored to separate the result of the analysis
and resolve phase for more granular incremental rebuilds. In this model,
any errors in one phase transition the trait into an error state, which
prevents it from being ran through subsequent phases. The ngcc compiler
on the other hand did not adopt this strict error model, which would
cause incomplete metadata—due to errors in earlier phases—to be offered
for compilation that could result in a hard crash.

This commit changes ngcc so that it starts using ngtsc's `Trait` concept
to track transitions between the analysis, resolve and compile phase
much more explicitly. This approach makes it illegal to attempt the
compile phase when either analysis or resolve has produced errors,
avoiding the crash.

Fixes #34500
Resolves FW-1788

### DIFF
--- a/packages/compiler-cli/ngcc/src/analysis/types.ts
+++ b/packages/compiler-cli/ngcc/src/analysis/types.ts
@@ -9,7 +9,7 @@ import {ConstantPool} from '@angular/compiler';
 import * as ts from 'typescript';
 import {Reexport} from '../../../src/ngtsc/imports';
 import {ClassDeclaration, Decorator} from '../../../src/ngtsc/reflection';
-import {CompileResult, DecoratorHandler, DetectResult} from '../../../src/ngtsc/transform';
+import {CompileResult, Trait} from '../../../src/ngtsc/transform';
 
 export interface AnalyzedFile {
   sourceFile: ts.SourceFile;
@@ -20,8 +20,8 @@ export interface AnalyzedClass {
   name: string;
   decorators: Decorator[]|null;
   declaration: ClassDeclaration;
-  diagnostics?: ts.Diagnostic[];
-  matches: MatchingHandler<unknown, unknown, unknown>[];
+  traits: Trait<unknown, unknown, unknown>[];
+  metaDiagnostics?: ts.Diagnostic[];
 }
 
 export interface CompiledClass extends AnalyzedClass {
@@ -41,10 +41,3 @@ export interface CompiledFile {
 
 export type DecorationAnalyses = Map<ts.SourceFile, CompiledFile>;
 export const DecorationAnalyses = Map;
-
-export interface MatchingHandler<D, A, R> {
-  handler: DecoratorHandler<D, A, R>;
-  detected: DetectResult<D>;
-  analysis: Readonly<A>;
-  resolution: Readonly<R>;
-}
--- a/packages/compiler-cli/ngcc/test/analysis/decoration_analyzer_spec.ts
+++ b/packages/compiler-cli/ngcc/test/analysis/decoration_analyzer_spec.ts
@@ -44,6 +44,7 @@ runInEachFileSystem(() => {
         const handler = jasmine.createSpyObj<DecoratorHandlerWithResolve>('TestDecoratorHandler', [
           'detect',
           'analyze',
+          'register',
           'resolve',
           'compile',
         ]);
@@ -388,6 +389,10 @@ runInEachFileSystem(() => {
           analyzer.analyzeProgram();
           expect(diagnosticLogs.length).toEqual(1);
           expect(diagnosticLogs[0]).toEqual(jasmine.objectContaining({code: -999999}));
+          expect(testHandler.analyze).toHaveBeenCalled();
+          expect(testHandler.register).not.toHaveBeenCalled();
+          expect(testHandler.resolve).not.toHaveBeenCalled();
+          expect(testHandler.compile).not.toHaveBeenCalled();
         });
 
         it('should report resolve diagnostics to the `diagnosticHandler` callback', () => {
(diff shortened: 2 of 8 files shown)


packages/compiler-cli/ngcc/src/analysis/types.ts
```diff
@@ -20,8 +20,8 @@ export interface AnalyzedClass {
   name: string;
   decorators: Decorator[]|null;
   declaration: ClassDeclaration;
-  diagnostics?: ts.Diagnostic[];
-  matches: MatchingHandler<unknown, unknown, unknown>[];
+  traits: Trait<unknown, unknown, unknown>[];
+  metaDiagnostics?: ts.Diagnostic[];
```
petebacondarwin: Why this name, `metaDiagnostics`? It seems like this is only diagnostics from migrations, in which case it would make more sense to call it `migrationDiagnostics`? In any case it would be very helpful to have a jsdoc description for this property.

packages/compiler-cli/ngcc/test/analysis/decoration_analyzer_spec.ts
```diff
@@ -372,25 +372,41 @@ runInEachFileSystem(() => {
...
-             expect(diagnosticLogs.length).toEqual(2);
-             expect(diagnosticLogs[0]).toEqual(jasmine.objectContaining({code: -999999}));
-             expect(diagnosticLogs[1]).toEqual(jasmine.objectContaining({code: -999998}));
-           });
+                },
+              ],
+              {analyzeError: true, resolveError: true});
+          analyzer.analyzeProgram();
+          expect(diagnosticLogs.length).toEqual(1);
+          expect(diagnosticLogs[0]).toEqual(jasmine.objectContaining({code: -999999}));
```
petebacondarwin: Should we be testing whether `resolve()` is even called if `analyze()` has errors?

packages/compiler-cli/ngcc/src/analysis/types.ts
JoostK: I took the name from `ClassRecord` as it exists in ngtsc, which has more or less the same purpose as ngcc's `AnalyzedClass`.

One other `metaDiagnostic` would be when multiple decorators are used together in an incompatible fashion, e.g. both `@Component` and `@Directive`. There's actually still a TODO for this to be an actual diagnostic, so there's indeed only migration diagnostics for now.

packages/compiler-cli/ngcc/test/analysis/decoration_analyzer_spec.ts
JoostK: Yeah, this is where I may want to improve the tests a bit. I do think however that it would be a separate test from the diagnostics tests here.

packages/compiler-cli/ngcc/test/analysis/migration_host_spec.ts
```diff
@@ -187,18 +181,23 @@ runInEachFileSystem(() => {
...
         const decorator = createComponentDecorator(mockClazz, {selector: 'comp', exportAs: null});
         host.injectSyntheticDecorator(mockClazz, decorator);
 
-        expect(diagnostics.length).toBe(1);
-        expect(ts.flattenDiagnosticMessageText(diagnostics[0].messageText, '\n'))
+        expect(analyzedFiles.length).toBe(1);
+        expect(analyzedFiles[0].analyzedClasses.length).toBe(1);
+        const migratedClass = analyzedFiles[0].analyzedClasses[0];
+        expect(migratedClass.traits.length).toBe(1);
+        const migratedTrait = migratedClass.traits[0] as ErroredTrait<unknown, unknown, unknown>;
```
petebacondarwin: Rather than this explicit cast we can rely upon the discriminated type system by doing:

```
        const migratedTrait = migratedClass.traits[0];
        if (migratedTrait.state !== TraitState.ERRORED) {
          return fail('Expected migrated class trait to be in an error state');
        }
```

It automatically knows that the trait must be an `ErroredTrait` if we get past this point.

petebacondarwin:
Looking good so far. I left a few comments.
What is the result of this? Is it that packages that have errors do not complete their processing by ngcc but that other packages can continue doing so? What happens about the errored package? Is it marked as processed? What about packages that rely upon the errored package? Will they complete as successful? Or will they error out too?

JoostK:
@petebacondarwin 

> What is the result of this? Is it that packages that have errors do not complete their processing by ngcc but that other packages can continue doing so?

In ngcc's main, we were already throwing an error to fail the entry-point's task when an error diagnostic is reported:

https://github.com/angular/angular/blob/918d8c990904b35ba67ea60c4b985f22fab1e853/packages/compiler-cli/ngcc/src/main.ts#L263-L275

This causes the execution mechanism to also error out immediately, preventing additional tasks from completing (even if some task may not be dependent on the failed entry-point).

From that perspective, nothing really changes. The primary change is that the internal compilation pipeline that is run in `DecoratorAnalyzer` no longer initiates a compilation phase if a prior phase already errored out, following the same strategy as ngtsc.

The crash would occur in the `compile` phase of `NgModuleHandler`, as its `resolve` phase had produced errors such that no resolution data is available. Since `compile` expects to be called when `analysis` and `resolve` have succeeded, it reads from the resolution data but crashes as it's not available. By not invoking the compilation the `compile` phase at all, this is avoided.

> What happens about the errored package? Is it marked as processed?

This aspect has not changed, but the failing entry-point will not be marked as processed.

> What about packages that rely upon the errored package? Will they complete as successful? Or will they error out too?

None of the remaining entry-points will be processed after an entry-point has failed processing.

petebacondarwin (approved):
LGTM from the ngcc side.

JoostK:
Superseded by #34889

Issue #34378, opened 2019-12-12
Internal compiler error on host binding with async pipe with Ivy


jonrimmer:
# 🐞 bug report

### Is this a regression?

Yes, it works in Angular v8.2 or in v9 with Ivy disabled.

### Description

In v8.2 it is possible to use the async pipe in a component's host binding to bind to an observable:

```ts
@Component({
  host: {
    '[class.active]': '(active$ | async) == true'
  }
})
export class MyComponent {
   active$: Observable<boolean>;
}
```
But this is broken in v9.

## 🔬 Minimal Reproduction

https://github.com/jonrimmer/ivy-binding-error

See the instructions in the README, and the source of https://github.com/jonrimmer/ivy-binding-error/blob/master/src/app/hello.component.ts

## 🔥 Exception or Error
<pre><code>
ERROR in Error: Internal Error: Unexpected node
    at error (ivy-binding-error/node_modules/@angular/compiler/bundles/compiler.umd.js:2721:15)
    at ValueConverter.allocateSlot (ivy-binding-error/node_modules/@angular/compiler/bundles/compiler.umd.js:18820:88)
    at ValueConverter.visitPipe (ivy-binding-error/node_modules/@angular/compiler/bundles/compiler.umd.js:17851:29)
    at BindingPipe.visit (ivy-binding-error/node_modules/@angular/compiler/bundles/compiler.umd.js:7493:28)
    at ValueConverter.AstMemoryEfficientTransformer.visitBinary (ivy-binding-error/node_modules/@angular/compiler/bundles/compiler.umd.js:7928:33)
    at Binary.visit (ivy-binding-error/node_modules/@angular/compiler/bundles/compiler.umd.js:7562:28)
    at ASTWithSource.visit (ivy-binding-error/node_modules/@angular/compiler/bundles/compiler.umd.js:7662:29)
    at ivy-binding-error/node_modules/@angular/compiler/bundles/compiler.umd.js:13611:41
    at Array.forEach (<anonymous>)
    at StylingBuilder._buildSingleInputs (ivy-binding-error/node_modules/@angular/compiler/bundles/compiler.umd.js:13609:20)
    at StylingBuilder._buildClassInputs (ivy-binding-error/node_modules/@angular/compiler/bundles/compiler.umd.js:13657:29)
    at StylingBuilder.buildUpdateLevelInstructions (ivy-binding-error/node_modules/@angular/compiler/bundles/compiler.umd.js:13696:69)
[...]

PR #34655, opened 2020-01-07
fix(ivy): more accurate detection of pipes in host bindings

AndrewKushnir:
Pipes in host binding expressions are not supported in View Engine and Ivy, but in some more complex cases (like `(value | pipe) === true`) compiler was not reporting errors. This commit extends Ivy logic to detect pipes in host binding expressions and throw in cases bindings are present. View Engine behavior remains the same.

This PR resolves FW-1787.
PR is related to #34378.

## PR Type

- [x] Bugfix

- [ ] Yes
- [x] No

### DIFF
--- a/packages/compiler-cli/test/ngtsc/ngtsc_spec.ts
+++ b/packages/compiler-cli/test/ngtsc/ngtsc_spec.ts
@@ -2250,29 +2250,67 @@ runInEachFileSystem(os => {
           }
         })
         class FooCmp {}
-    `);
+      `);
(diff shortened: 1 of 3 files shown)


alxhub (approved):
This is okay for now.

In the future we should really migrate host bindings onto the existing template type-checking machinery, and enable the production of rich diagnostic errors for them.

AndrewKushnir:
[Presubmit](http://test/OCL:290350340:BASE:290350368:1579302607544:d98aff09)

PR #34660, opened 2020-01-07
fix(ngcc): handle UMD factories that do not use all params

### LINKED ISSUE #34653: NGCC: generates broken UMD for @angular/fire/forestore
# 🐞 bug report

### Affected Package
@angular/compiler-cli

### Is this a regression?
Not sure

### Description

## 🔬 Minimal Reproduction
If you run ngcc, though, then you will get an error when you run your server like this:
```js
/Users/jeffwhelpley/repos/criticly/frontend/dist/apps/criticly/server/main.js:92536
AngularFirestore.ɵprov = ɵngcc0.ɵɵdefineInjectable({ token: AngularFirestore, factory: function (t) { return AngularFirestore.ɵfac(t); } });
                                ^

TypeError: ɵngcc0.ɵɵdefineInjectable is not a function
    at /Users/jeffwhelpley/repos/criticly/frontend/dist/apps/criticly/server/main.js:92536:33
    at /Users/jeffwhelpley/repos/criticly/frontend/dist/apps/criticly/server/main.js:92567:6
    at /Users/jeffwhelpley/repos/criticly/frontend/dist/apps/criticly/server/main.js:92235:13
    at Object.../../node_modules/@angular/fire/bundles/firestore.umd.js (/Users/jeffwhelpley/repos/criticly/frontend/dist/apps/criticly/server/main.js:92237:2)
    at __webpack_require__ (/Users/jeffwhelpley/repos/criticly/frontend/dist/apps/criticly/server/main.js:20:30)
```
When I look at the UMD that is generated for Firestore by ngcc, I see this at the beginning:
```js
(function (global, factory) {
     true ? factory(exports, __webpack_require__(/*! rxjs */ "../../node_modules/rxjs/index.js"), __webpack_require__(/*! rxjs/operators */ "../../node_modules/rxjs/operators/index.js"), __webpack_require__(/*! @angular/fire */ "../../node_modules/@angular/fire/bundles/core.umd.js"), __webpack_require__(/*! @angular/core */ "../../node_modules/@angular/core/bundles/core.umd.js"), __webpack_require__(/*! @angular/common */ "../../node_modules/@angular/common/bundles/common.umd.js"), __webpack_require__(/*! firebase/app */ "../../node_modules/firebase/app/dist/index.cjs.js"), __webpack_require__(/*! firebase/firestore */ "../../node_modules/firebase/firestore/dist/index.cjs.js"),__webpack_require__(/*! @angular/core */ "../../node_modules/@angular/core/bundles/core.umd.js")) :
    undefined;
}(this, (function (exports,rxjs,operators,fire,core,common,firebase,ɵngcc0) { 'use strict';
```
I was debugging this code for awhile to try and figure out what is going on until I realized that the call to the factory() function is sending in more input params (9) than there is actually in the function definition (8). he function has exports,rxjs,operators,fire,core,common,firebase,ɵngcc0 which is 8 params, but here is what the factory is sending in:
```js
1. exports
2. __webpack_require__(/*! rxjs */ "../../node_modules/rxjs/index.js")
3. __webpack_require__( "../../node_modules/rxjs/operators/index.js")
4. __webpack_require__( "../../node_modules/@angular/fire/bundles/core.umd.js")
5. __webpack_require__( "../../node_modules/@angular/core/bundles/core.umd.js")
6. __webpack_require__("../../node_modules/@angular/common/bundles/common.umd.js")
7. __webpack_require__("../../node_modules/firebase/app/dist/index.cjs.js")
8. __webpack_require__("../../node_modules/firebase/firestore/dist/index.cjs.js")
9. __webpack_require__(/*! @angular/core */ "../../node_modules/@angular/core/bundles/core.umd.js")
```

When you look at the original UMD for Firestore, this starts to make sense because you can see there is an extra param there as well:
```js
(function (global, factory) {
    typeof exports === 'object' && typeof module !== 'undefined' ? factory(exports, require('rxjs'), require('rxjs/operators'), require('@angular/fire'), require('@angular/core'), require('@angular/common'), require('firebase/app'), require('firebase/firestore')) :
    typeof define === 'function' && define.amd ? define(['exports', 'rxjs', 'rxjs/operators', '@angular/fire', '@angular/core', '@angular/common', 'firebase/app', 'firebase/firestore'], factory) :
    (factory((global.angularfire2 = global.angularfire2 || {}, global.angularfire2.firestore = {}),global.rxjs,global.rxjs.operators,global.angularfire2,global.ng.core,global.ng.common,global.firebase));
}(this, (function (exports,rxjs,operators,fire,core,common,firebase) { 'use strict';
```

You can see here that `require('firebase/firestore')` is at the end of the `factory()` call, but there is no param for that. So, in the original UMD there is no harm because that extra param isn't used. But, ngcc must be adding an extra param without checking for this.

So, I believe at least part of the fix is figuring out why the normal Firestore UMD has that references to /firebase/firestore even though it is not being used.

## 🌍  Your Environment

**Angular Version:** 9.0.0-rc.7

From https://github.com/angular/angularfire/issues/2280

//cc @jeffwhelpley

petebacondarwin:
In some cases, where a module imports a dependency
but does not actually use it, UMD bundlers may remove
the dependency parameter from the UMD factory function
definition.

For example:

```
import * as x from 'x';
import * as z from 'z';
export const y = x;
```

may result in a UMD bundle including:

```
(function (global, factory) {
	typeof exports === 'object' && typeof module !== 'undefined' ? factory(exports, require('x'), require('z')) :
	typeof define === 'function' && define.amd ? define(['exports', 'x', 'z'], factory) :
	(global = global || self, factory(global.myBundle = {}, global.x));
}(this, (function (exports, x) { 'use strict';
...
})));
```

Note that while the `z` dependency is provide in the call,
the factory itself only accepts `exports` and `x` as parameters.

Previously ngcc appended new dependencies to the end of the
factory function, but this breaks in the above scenario. Now
the new dependencies are prefixed in front of any dependencies
already in place.

Fixes #34653

### DIFF
--- a/packages/compiler-cli/ngcc/src/rendering/umd_rendering_formatter.ts
+++ b/packages/compiler-cli/ngcc/src/rendering/umd_rendering_formatter.ts
@@ -193,7 +193,7 @@ function renderGlobalDependencies(
     return;
   }
   const injectionPoint = globalFactoryCall.arguments.length > 0 ?
-      // Add extra dependencies after the first argument
+      // Add extra dependencies before the first argument
       globalFactoryCall.arguments[0].getFullStart() :
       // Backup one char to account for the closing parenthesis on the call
       globalFactoryCall.getEnd() - 1;
--- a/packages/compiler-cli/ngcc/src/rendering/umd_rendering_formatter.ts
+++ b/packages/compiler-cli/ngcc/src/rendering/umd_rendering_formatter.ts
@@ -27,7 +27,24 @@ export class UmdRenderingFormatter extends Esm5RenderingFormatter {
   constructor(protected umdHost: UmdReflectionHost, isCore: boolean) { super(umdHost, isCore); }
 
   /**
-   *  Add the imports to the UMD module IIFE.
+   * Add the imports to the UMD module IIFE.
+   *
+   * Note that imports at "prepended" to the start of the parameter list of the factory function,
+   * and so also to the arguments passed to it when it is called.
+   * This is because there are scenarios where the factory function does not accept as many
+   * parameters as are passed as argument in the call. For example:
+   *
+   * ```
+   * (function (global, factory) {
+   *     typeof exports === 'object' && typeof module !== 'undefined' ?
+   *         factory(exports,require('x'),require('z')) :
+   *     typeof define === 'function' && define.amd ?
+   *         define(['exports', 'x', 'z'], factory) :
+   *     (global = global || self, factory(global.myBundle = {}, global.x));
+   * }(this, (function (exports, x) { ... }
+   * ```
+   *
+   * (See that the `z` import is not being used by the factory function.)
    */
   addImports(output: MagicString, imports: Import[], file: ts.SourceFile): void {
     if (imports.length === 0) {
@@ -126,12 +143,12 @@ function renderCommonJsDependencies(
   }
   const factoryCall = conditional.whenTrue;
   const injectionPoint = factoryCall.arguments.length > 0 ?
-      // Add extra dependencies after the first argument
-      factoryCall.arguments[0].getEnd() :
+      // Add extra dependencies before the first argument
+      factoryCall.arguments[0].getFullStart() :
       // Backup one char to account for the closing parenthesis on the call
       factoryCall.getEnd() - 1;
   const importString = imports.map(i => `require('${i.specifier}')`).join(',');
-  output.appendLeft(injectionPoint, (factoryCall.arguments.length > 0 ? ',' : '') + importString);
+  output.appendLeft(injectionPoint, importString + (factoryCall.arguments.length > 0 ? ',' : ''));
 }
 
 /**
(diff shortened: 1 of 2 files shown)


gkalpak (changes requested):
This code seems to assume that the first argument/array item (if any) will be `exports`. In cases where there is no `exports` parameter, we will still run into the same issue. E.g.:

```js
(function (global, factory) {
  typeof exports === 'object' && typeof module !== 'undefined' ? factory(require('some/dep/for/side-effects')) :
  typeof define === 'function' && define.amd ? define('file', ['some/dep/for/side-effects'], factory) :
  (factory(global.someDepForSideEffects));
}(this, (function () { ... })));
```

<sub>(There is also the case where the first two parameters are `require` and `exports`, but we do not fully support it anyway and the current implementation doesn't break it (it just puts the extra deps in-between `require` and `exports`).)</sub>

How about putting the extra deps at the very beginning instead?

petebacondarwin:
I felt uncomfortable about putting the parameters at the start, since it is so common to have an `exports` parameter; and I was worried that some other tooling might expect the first parameter not to be exports.
But I guess this is not a real concern, since UMDs can have so many formats with various bits included or excluded.

So OK, I will change it to put the new deps at the start of the parameter list.

gkalpak (approved):
👌

PR #34661, opened 2020-01-07
fix(ngcc): libraries using spread in object literals cannot be processed

### LINKED ISSUE #34659: ngcc: libraries using spread operator in object literals cannot be processed
Consider a library that uses a shared constant for host bindings. e.g.

```ts
export const BASE_BINDINGS= {
  '[class.mat-themed]': '_isThemed',
}

----

@Directive({
  host: {...BASE_BINDINGS, '(click)': '...'}
})
export class Dir1 {}

@Directive({
  host: {...BASE_BINDINGS, '(click)': '...'}
})
export class Dir2 {}
```

Previously when these components were shipped as part of the library to NPM, consumers were able to consume `Dir1` and `Dir2`. No errors showed up.

Now with Ivy, when ngcc tries to process the library, an error will be thrown. The error is stating that the host bindings should be an object (which they obviously are). This happens because TypeScript transforms the object spread to individual `Object.assign` calls (for compatibility).

The partial evaluator used by the `@Directive` annotation handler is unable to process this expression because there is no integrated support for `Object.assign`. In View Engine, this was not a problem because the `metadata.json` files from the library were used to compute the host bindings.

### Why should Object.assign work?

Currently ngtsc works with object spread operators. Just ngcc which runs into the transpiled `Object.assign` calls will not support object spreads. This results in inconsistent behavior.

Also, this previously worked in View Engine, so it would help avoiding a breaking change.

### How could it be fixed?

We could add another `BuiltinFn` into the partial evaluator. It will be somewhat difficult to detect whether a expression with `Object.assign` is actually resolving to the default ES6 `Object.assign`.

Since the default libs are disabled in the ngcc program, we cannot simply use the type checker. Joost had the idea of providing these types on our own in the program. That would make the detection easy and also work for ES5 files that rely on the `Object.assign` polyfill.

devversion:
Consider a library that uses a shared constant for host bindings. e.g.

```ts
export const BASE_BINDINGS= {
  '[class.mat-themed]': '_isThemed',
}

----

@Directive({
  host: {...BASE_BINDINGS, '(click)': '...'}
})
export class Dir1 {}

@Directive({
  host: {...BASE_BINDINGS, '(click)': '...'}
})
export class Dir2 {}
```

Previously when these components were shipped as part of the
library to NPM, consumers were able to consume `Dir1` and `Dir2`.
No errors showed up.

Now with Ivy, when ngcc tries to process the library, an error
will be thrown. The error is stating that the host bindings should
be an object (which they obviously are). This happens because
TypeScript transforms the object spread to individual
`Object.assign` calls (for compatibility).

The partial evaluator used by the `@Directive` annotation handler
is unable to process this expression because there is no
integrated support for `Object.assign`. In View Engine, this was
not a problem because the `metadata.json` files from the library
were used to compute the host bindings.

To fix this, we add support for `Object.assign`. Unfortunately there
is no safe and easy way to detect if a property access resolves to
the `assign` method of the global `Object` variable. To ensure that
the partial evaluator can resolve both the "lhs" and "rhs" of the property
access, we add the needed `Object.assign` types to the ngcc program.

Fixes #34659

### DIFF
--- a/packages/compiler-cli/ngcc/src/host/esm2015_host.ts
+++ b/packages/compiler-cli/ngcc/src/host/esm2015_host.ts
@@ -8,7 +8,7 @@
 
 import * as ts from 'typescript';
 
-import {ClassDeclaration, ClassMember, ClassMemberKind, ConcreteDeclaration, CtorParameter, Declaration, Decorator, TypeScriptReflectionHost, TypeValueReference, isDecoratorIdentifier, reflectObjectLiteral} from '../../../src/ngtsc/reflection';
+import {BuiltinDeclaration, ClassDeclaration, ClassMember, ClassMemberKind, ConcreteDeclaration, CtorParameter, Declaration, Decorator, TypeScriptReflectionHost, TypeValueReference, isDecoratorIdentifier, reflectObjectLiteral} from '../../../src/ngtsc/reflection';
 import {isWithinPackage} from '../analysis/util';
 import {Logger} from '../logging/logger';
 import {BundleProgram} from '../packages/bundle_program';
@@ -351,6 +351,25 @@ export class Esm2015ReflectionHost extends TypeScriptReflectionHost implements N
       if (aliasedIdentifier !== null) {
         return this.getDeclarationOfIdentifier(aliasedIdentifier);
       }
+    }
+
+    // We detect if the resolved declaration refers to the JavaScript global `Object` by checking
+    // if the declaration node is named `Object`, resolves to a type `ObjectConstructor`, and has
+    // been defined in a default library source file.
+    if (superDeclaration && superDeclaration.node !== null &&
+        ts.isVariableDeclaration(superDeclaration.node) &&
+        ts.isIdentifier(superDeclaration.node.name) &&
+        superDeclaration.node.name.text === 'Object' && superDeclaration.node.type &&
+        ts.isTypeReferenceNode(superDeclaration.node.type) &&
+        ts.isIdentifier(superDeclaration.node.type.typeName) &&
+        superDeclaration.node.type.typeName.text === 'ObjectConstructor' &&
+        this.src.program.isSourceFileDefaultLibrary(superDeclaration.node.getSourceFile())) {
+      return {
+        builtin: BuiltinDeclaration.JsGlobalObject,
+        expression: id,
+        viaModule: null,
+        node: null,
+      };
     }
 
     return superDeclaration;
--- a/packages/compiler-cli/ngcc/src/host/esm2015_host.ts
+++ b/packages/compiler-cli/ngcc/src/host/esm2015_host.ts
@@ -8,7 +8,7 @@
 
 import * as ts from 'typescript';
 
-import {BuiltinDeclaration, ClassDeclaration, ClassMember, ClassMemberKind, ConcreteDeclaration, CtorParameter, Declaration, Decorator, TypeScriptReflectionHost, TypeValueReference, isDecoratorIdentifier, reflectObjectLiteral} from '../../../src/ngtsc/reflection';
+import {ClassDeclaration, ClassMember, ClassMemberKind, ConcreteDeclaration, CtorParameter, Declaration, Decorator, KnownDeclaration, TypeScriptReflectionHost, TypeValueReference, isDecoratorIdentifier, reflectObjectLiteral,} from '../../../src/ngtsc/reflection';
 import {isWithinPackage} from '../analysis/util';
 import {Logger} from '../logging/logger';
 import {BundleProgram} from '../packages/bundle_program';
@@ -353,19 +353,11 @@ export class Esm2015ReflectionHost extends TypeScriptReflectionHost implements N
       }
     }
 
-    // We detect if the resolved declaration refers to the JavaScript global `Object` by checking
-    // if the declaration node is named `Object`, resolves to a type `ObjectConstructor`, and has
-    // been defined in a default library source file.
-    if (superDeclaration && superDeclaration.node !== null &&
-        ts.isVariableDeclaration(superDeclaration.node) &&
-        ts.isIdentifier(superDeclaration.node.name) &&
-        superDeclaration.node.name.text === 'Object' && superDeclaration.node.type &&
-        ts.isTypeReferenceNode(superDeclaration.node.type) &&
-        ts.isIdentifier(superDeclaration.node.type.typeName) &&
-        superDeclaration.node.type.typeName.text === 'ObjectConstructor' &&
-        this.src.program.isSourceFileDefaultLibrary(superDeclaration.node.getSourceFile())) {
+    // If the identifier resolves to the global JavaScript `Object`, return a
+    // declaration that denotes it as the known `JsGlobalObject` declaration.
+    if (superDeclaration !== null && this.isJavaScriptObjectDeclaration(superDeclaration)) {
       return {
-        builtin: BuiltinDeclaration.JsGlobalObject,
+        known: KnownDeclaration.JsGlobalObject,
         expression: id,
         viaModule: null,
         node: null,
@@ -1715,6 +1707,30 @@ export class Esm2015ReflectionHost extends TypeScriptReflectionHost implements N
 
     const exportDecl = namespaceExports.get(expression.name.text) !;
     return {...exportDecl, viaModule: namespaceDecl.viaModule};
+  }
+
+  /** Checks if the specified declaration resolves to the known JavaScript global `Object`. */
+  protected isJavaScriptObjectDeclaration(decl: Declaration): boolean {
+    if (decl.node === null) {
+      return false;
+    }
+    const node = decl.node;
+    // The default TypeScript library types the global `Object` variable through
+    // a variable declaration with a type reference resolving to `ObjectConstructor`.
+    if (!ts.isVariableDeclaration(node) || !ts.isIdentifier(node.name) ||
+        node.name.text !== 'Object' || node.type === undefined) {
+      return false;
+    }
+    const typeNode = node.type;
+    // If the variable declaration does not have a type resolving to `ObjectConstructor`,
+    // we cannot guarantee that the declaration resolves to the global `Object` variable.
+    if (!ts.isTypeReferenceNode(typeNode) || !ts.isIdentifier(typeNode.typeName) ||
+        typeNode.typeName.text !== 'ObjectConstructor') {
+      return false;
+    }
+    // Finally, check if the type definition for `Object` originates from a default library
+    // definition file. This requires default types to be enabled for the host program.
+    return this.src.program.isSourceFileDefaultLibrary(node.getSourceFile());
   }
 }
 
--- a/packages/compiler-cli/ngcc/src/host/esm5_host.ts
+++ b/packages/compiler-cli/ngcc/src/host/esm5_host.ts
@@ -8,11 +8,12 @@
 
 import * as ts from 'typescript';
 
-import {ClassDeclaration, ClassMember, ClassMemberKind, Declaration, Decorator, FunctionDefinition, isNamedVariableDeclaration, Parameter, reflectObjectLiteral, TsHelperFn,} from '../../../src/ngtsc/reflection';
+import {ClassDeclaration, ClassMember, ClassMemberKind, Declaration, Decorator, FunctionDefinition, Parameter, TsHelperFn, isNamedVariableDeclaration, reflectObjectLiteral} from '../../../src/ngtsc/reflection';
 import {getNameText, hasNameIdentifier, stripDollarSuffix} from '../utils';
 
-import {Esm2015ReflectionHost, getPropertyValueFromSymbol, isAssignment, isAssignmentStatement, ParamInfo} from './esm2015_host';
+import {Esm2015ReflectionHost, ParamInfo, getPropertyValueFromSymbol, isAssignment, isAssignmentStatement} from './esm2015_host';
 import {NgccClassSymbol} from './ngcc_host';
+
 
 /**
  * ESM5 packages contain ECMAScript IIFE functions that act like classes. For example:
@@ -655,12 +656,12 @@ function getTsHelperFn(node: ts.NamedDeclaration): TsHelperFn|null {
       null;
 
   switch (name) {
+    case '__assign':
+      return TsHelperFn.Assign;
     case '__spread':
       return TsHelperFn.Spread;
     case '__spreadArrays':
       return TsHelperFn.SpreadArrays;
-    case '__assign':
-      return TsHelperFn.Assign;
     default:
       return null;
   }
--- a/packages/compiler-cli/ngcc/src/host/esm5_host.ts
+++ b/packages/compiler-cli/ngcc/src/host/esm5_host.ts
@@ -8,10 +8,10 @@
 
 import * as ts from 'typescript';
 
-import {ClassDeclaration, ClassMember, ClassMemberKind, Declaration, Decorator, FunctionDefinition, Parameter, TsHelperFn, isNamedVariableDeclaration, reflectObjectLiteral} from '../../../src/ngtsc/reflection';
+import {ClassDeclaration, ClassMember, ClassMemberKind, Declaration, Decorator, FunctionDefinition, isNamedVariableDeclaration, Parameter, reflectObjectLiteral, TsHelperFn,} from '../../../src/ngtsc/reflection';
 import {getNameText, hasNameIdentifier, stripDollarSuffix} from '../utils';
 
-import {Esm2015ReflectionHost, ParamInfo, getPropertyValueFromSymbol, isAssignment, isAssignmentStatement} from './esm2015_host';
+import {Esm2015ReflectionHost, getPropertyValueFromSymbol, isAssignment, isAssignmentStatement, ParamInfo} from './esm2015_host';
 import {NgccClassSymbol} from './ngcc_host';
 
 /**
@@ -659,6 +659,8 @@ function getTsHelperFn(node: ts.NamedDeclaration): TsHelperFn|null {
       return TsHelperFn.Spread;
     case '__spreadArrays':
       return TsHelperFn.SpreadArrays;
+    case '__assign':
+      return TsHelperFn.Assign;
     default:
       return null;
   }
--- a/packages/compiler-cli/ngcc/src/packages/entry_point_bundle.ts
+++ b/packages/compiler-cli/ngcc/src/packages/entry_point_bundle.ts
@@ -6,7 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 import * as ts from 'typescript';
-import {AbsoluteFsPath, FileSystem, NgtscCompilerHost, absoluteFrom} from '../../../src/ngtsc/file_system';
+import {AbsoluteFsPath, FileSystem, NgtscCompilerHost} from '../../../src/ngtsc/file_system';
 import {PathMappings} from '../utils';
 import {BundleProgram, makeBundleProgram} from './bundle_program';
 import {EntryPoint, EntryPointFormat} from './entry_point';
@@ -50,8 +50,7 @@ export function makeEntryPointBundle(
   const rootDir = entryPoint.package;
   const options: ts.CompilerOptions = {
     allowJs: true,
-    maxNodeModuleJsDepth: Infinity,
-    noLib: true, rootDir, ...pathMappings
+    maxNodeModuleJsDepth: Infinity, rootDir, ...pathMappings
   };
   const srcHost = new NgccSourcesCompilerHost(fs, options, entryPoint.path);
   const dtsHost = new NgtscCompilerHost(fs, options);
--- a/packages/compiler-cli/src/ngtsc/partial_evaluator/src/builtin.ts
+++ b/packages/compiler-cli/src/ngtsc/partial_evaluator/src/builtin.ts
@@ -46,19 +46,15 @@ export class ObjectAssignBuiltinFn extends BuiltinFn {
     if (args.length === 0) {
       return DynamicValue.fromUnsupportedSyntax(node);
     }
-    const target = args[0];
-    if (target instanceof DynamicValue) {
-      return DynamicValue.fromDynamicInput(node, target);
-    } else if (!(target instanceof Map)) {
-      return DynamicValue.fromUnsupportedSyntax(node);
-    }
-    for (let i = 1; i < args.length; i++) {
-      const source = args[i];
-      if (source instanceof DynamicValue) {
-        return DynamicValue.fromDynamicInput(node, source);
-      } else if (!(source instanceof Map)) {
+    for (const arg of args) {
+      if (arg instanceof DynamicValue) {
+        return DynamicValue.fromDynamicInput(node, arg);
+      } else if (!(arg instanceof Map)) {
         return DynamicValue.fromUnsupportedSyntax(node);
       }
+    }
+    const [target, ...sources] = args as Map<string, ResolvedValue>[];
+    for (const source of sources) {
       source.forEach((value, key) => target.set(key, value));
     }
     return target;
--- a/packages/compiler-cli/src/ngtsc/partial_evaluator/src/builtin.ts
+++ b/packages/compiler-cli/src/ngtsc/partial_evaluator/src/builtin.ts
@@ -12,25 +12,25 @@ import {DynamicValue} from './dynamic';
 import {BuiltinFn, ResolvedValue, ResolvedValueArray} from './result';
 
 export class ArraySliceBuiltinFn extends BuiltinFn {
-  constructor(private node: ts.Node, private lhs: ResolvedValueArray) { super(); }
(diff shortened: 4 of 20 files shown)


packages/compiler-cli/src/ngtsc/partial_evaluator/src/interpreter.ts
```diff
@@ -367,7 +367,13 @@ export class StaticInterpreter {
       return lhs[rhs];
     } else if (lhs instanceof Reference) {
       const ref = lhs.node;
-      if (this.host.isClass(ref)) {
+      const identifier = identifierOfNode(ref);
+      // Handles `Object.assign`. The interpreter intends to support the spread operator, but
+      // should also handle a TypeScript transpiled spread object literal. TypeScript transforms
+      // the spread operator into `Object.assign` calls when targeting anything older than ES2018.
+      if (identifier && identifier.text === 'Object' && rhs === 'assign') {
```
devversion: This is still unsafe. I'm not sure how we can reliably ensure that this `Reference` is the global `Object`. I don't want to check if the source-file that declares `Reference#node` matches the name of the fake source file we added to the program. This is specific to ngcc, and the isolated interpreter should not know about this implementation detail.

Any ideas?

devversion: Improved by using a global definition that can be identified.

gkalpak (commented):
TBH, I would expect (and prefer) `Object.assign()` to be treated as a [BuiltinFn][1] and `__assign()` to be treated as a [TsHelperFn][2]. That would also ensure that fixing/improving the respective features (`BuiltinFn`/`TsHelperFn`) will apply to `assign` too.

(For example, there is work in progress to support imported helpers (e.g. `tslib.__<helper>()`). With the current implementation, this fix should be applied separately to `assign`.

[1]: https://github.com/angular/angular/blob/725148a44/packages/compiler-cli/src/ngtsc/partial_evaluator/src/result.ts#L56
[2]: https://github.com/angular/angular/blob/725148a44/packages/compiler-cli/src/ngtsc/reflection/src/host.ts#L334

devversion:
@gkalpak Thanks for your initial feedback. I initially was trying to use a `BuiltinFn`, but it turned out that it doesn't even get to [`accessHelper`](https://github.com/angular/angular/blob/725148a44db896f919fc0b6d7e256dbc35af3ec6/packages/compiler-cli/src/ngtsc/partial_evaluator/src/interpreter.ts#L328) which runs the built-in functions. This is because `Object` will resolve to a `DynamicValue`. This might be avoidable with the global type definition, or by introducing a new way that allows mapping identifiers to `ts.Node`'s. This might be useful for your PR that fixes inlined ts helper functions too. I chatted with Alex about this, and I'll be experiment with this.

In general, I agree that it would be good to handle `__assign` through a `TsHelperFn`. Though, for `Object.assign` I feel slightly different. I think we don't want to add support for it in the `Interpreter`, but rather only handle it in ngcc (with the reason of supporting transpiled libraries). If we'd add it to the interpreter, then I could imagine us arguing adding more functions like `Object.keys` etc. 

Also I wanted to mention that I think that most of the evaluation code for built-in functions or helper functions could be simplified, by providing a way to transform the call expression into a corresponding TS expression. e.g. `__spreadArray(a, b)` becomes `[...a, ...b]`. The evaluator will then handle it as expected without us having to worry about the implementation.

gkalpak:
> I initially was trying to use a `BuiltinFn`, but it turned out that it doesn't even get to `accessHelper` [...]
> This might be avoidable with the global type definition [...]

Yes, I believe so.

##
> or by introducing a new way that allows mapping identifiers to `ts.Node`'s. This might be useful for your PR that fixes inlined ts helper functions too.

Indeed. Looking forward to see what you'll come up with.

##
> I think we don't want to add support for `Object.assign` in the `Interpreter`, but rather only handle it in ngcc (with the reason of supporting transpiled libraries).

TBH, I don't see why :grin:
`Object.assign()` is not much different than other `BuiltinFn`s (such as `Array.prototype.slice()`) and if there are `Object.assign()` instances that are _not_ a result of transpilation, it would be weird to support them in ngcc, but not ngtsc.

##
> If we'd add it to the interpreter, then I could imagine us arguing adding more functions like `Object.keys` etc. 

Which is bad, because... 😕 :grin:

packages/compiler-cli/src/ngtsc/partial_evaluator/src/builtin.ts
```diff
@@ -8,29 +8,45 @@
...
+ * Resolves the specified builtin declaration to a resolved value. For example,
+ * the known JavaScript global `Object` will resolve to a `Map` that provides the
+ * `assign` method with a builtin function. This enables evaluation of `Object.assign`.
+ */
+export function resolveBuiltinDeclaration(builtinDecl: BuiltinDeclaration): ResolvedValue {
+  switch (builtinDecl) {
+    case BuiltinDeclaration.JsGlobalObject:
+      return new Map([['assign', new ObjectAssignBuiltinFn()]]);
+    default:
+      throw new Error(`Cannot resolve unknown builtin declaration ${builtinDecl}`);
```
alxhub:
```suggestion
      throw new Error(`Cannot resolve unknown builtin declaration ${BuiltinDeclaration[builtinDecl]}`);
```

(gives the name, rather than the enum index)

packages/compiler-cli/src/ngtsc/partial_evaluator/src/builtin.ts
```diff
@@ -8,29 +8,45 @@
...
+  }
+}
+
 export class ArraySliceBuiltinFn extends BuiltinFn {
-  constructor(private node: ts.Node, private lhs: ResolvedValueArray) { super(); }
+  constructor(private lhs: ResolvedValueArray) { super(); }
 
-  evaluate(args: ResolvedValueArray): ResolvedValue {
+  evaluate(node: ts.Node, args: ResolvedValueArray): ResolvedValue {
```
alxhub: This is a `ts.CallExpression`, right? We can type it more narrowly.

packages/compiler-cli/src/ngtsc/reflection/src/host.ts
```diff
@@ -395,6 +409,12 @@ export interface BaseDeclaration<T extends ts.Declaration = ts.Declaration> {
    * TypeScript reference to the declaration itself, if one exists.
    */
   node: T|null;
+
+  /**
+   * If set, describes the type of builtin JavaScript declaration this
+   * declaration resolves to.
+   */
+  builtinDecl?: BuiltinDeclaration
```
alxhub: In the ngtsc codebase, we don't use optional structure members (at least not often). Additionally, I think `Decl` as a suffix is a bit redundant given this is in a "Declaration" interface.
```suggestion
  builtin: BuiltinDeclaration|null;
```

packages/compiler-cli/src/ngtsc/partial_evaluator/test/utils.ts
```diff
@@ -16,17 +16,20 @@ import {getDeclaration, makeProgram} from '../../testing';
...
-  const {program, options, host} = makeProgram([
-    {name: absoluteFrom('/entry.ts'), contents: `${code}; const target$ = ${expr};`},
-    ...supportingFiles
-  ]);
+  const {program, options, host} = makeProgram(
+      [
+        {name: absoluteFrom('/entry.ts'), contents: `${code}; const target$ = ${expr};`},
+        ...supportingFiles
+      ],
+      undefined, undefined, checkForErrors);
```
alxhub: For readability, I tend to comment each primitive/unspecified value with its argument name:

```suggestion
      /* options */ undefined, /* host */ undefined, checkForErrors);
```

packages/compiler-cli/src/ngtsc/partial_evaluator/test/evaluator_spec.ts
```diff
@@ -13,9 +13,8 @@ import {Reference} from '../../imports';
 import {DependencyTracker} from '../../incremental/api';
 import {FunctionDefinition, TsHelperFn, TypeScriptReflectionHost} from '../../reflection';
 import {getDeclaration, makeProgram} from '../../testing';
-import {DynamicValue} from '../src/dynamic';
+import {DynamicValue, EnumValue} from '../index';
```
alxhub: Don't import from `index`, use a direct import to the file.

packages/compiler-cli/src/ngtsc/partial_evaluator/src/builtin.ts
```diff
@@ -8,29 +8,45 @@
...
 
+/**
+ * Resolves the specified builtin declaration to a resolved value. For example,
+ * the known JavaScript global `Object` will resolve to a `Map` that provides the
+ * `assign` method with a builtin function. This enables evaluation of `Object.assign`.
+ */
+export function resolveBuiltinDeclaration(builtinDecl: BuiltinDeclaration): ResolvedValue {
+  switch (builtinDecl) {
+    case BuiltinDeclaration.JsGlobalObject:
+      return new Map([['assign', new ObjectAssignBuiltinFn()]]);
```
alxhub: Rather than construct a new `Map` each time, let's have a shared global one.

alxhub (changes requested):

packages/compiler-cli/src/ngtsc/partial_evaluator/src/builtin.ts
devversion: Good idea. Worth mentioning: I took this from the `TsHelperFn` code 😄 I'll do the same there too.

packages/compiler-cli/src/ngtsc/partial_evaluator/test/utils.ts
devversion: Sure, sounds good to me 👍

packages/compiler-cli/src/ngtsc/partial_evaluator/src/builtin.ts
devversion: I had that initially, but figured that one could also argue that it unnecessarily allocates memory since it runs as a top-level module side-effect. Should be minimal though, so let's do it.

gkalpak (changes requested):
Nice! This approach turned out quite clean :tada:

packages/compiler-cli/src/ngtsc/partial_evaluator/src/ts_helpers.ts
```diff
@@ -10,17 +10,29 @@ import * as ts from 'typescript';
...
     case TsHelperFn.Spread:
     case TsHelperFn.SpreadArrays:
       return evaluateTsSpreadHelper(node, args);
+    case TsHelperFn.Assign:
+      // Use the same implementation we use for `Object.assign`. Semantically these
+      // functions are the same, so they can also share the same evaluation code.
+      return objectAssignBuiltinFn.evaluate(node, args);
     default:
-      throw new Error(`Cannot evaluate unknown helper ${helper} inline`);
+      throw new Error(`Cannot evaluate unknown helper ${TsHelperFn[helper]} inline`);
```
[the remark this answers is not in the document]
devversion: I did this upon a review comment. I don't think it's necessarily to print out the name, so let's just keep it like before, and do the same for `BuiltinDeclaration`.

packages/compiler-cli/src/ngtsc/reflection/src/host.ts
```diff
@@ -340,6 +340,20 @@ export enum TsHelperFn {
    * Indicates the `__spreadArrays` function.
    */
   SpreadArrays,
+  /**
+   * Indicates the `__assign` function.
+   */
+  Assign,
```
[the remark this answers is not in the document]
devversion: I'm using the chronological order 😄 Will change!

packages/compiler-cli/ngcc/src/host/esm2015_host.ts
```diff
@@ -353,6 +353,25 @@ export class Esm2015ReflectionHost extends TypeScriptReflectionHost implements N
       }
     }
 
+    // We detect if the resolved declaration refers to the JavaScript global `Object` by checking
+    // if the declaration node is named `Object`, resolves to a type `ObjectConstructor`, and has
+    // been defined in a default library source file.
+    if (superDeclaration && superDeclaration.node !== null &&
```
JoostK: In the compiler we tend to use explicit `null`/`undefined` checks, even if's more verbose.

```suggestion
    if (superDeclaration !== null && superDeclaration.node !== null &&
```

And yes, ngcc doesn't follow this convention very closely 😅

packages/compiler-cli/ngcc/src/host/esm2015_host.ts
```diff
@@ -353,6 +353,25 @@ export class Esm2015ReflectionHost extends TypeScriptReflectionHost implements N
...
+        ts.isVariableDeclaration(superDeclaration.node) &&
+        ts.isIdentifier(superDeclaration.node.name) &&
+        superDeclaration.node.name.text === 'Object' && superDeclaration.node.type &&
+        ts.isTypeReferenceNode(superDeclaration.node.type) &&
+        ts.isIdentifier(superDeclaration.node.type.typeName) &&
+        superDeclaration.node.type.typeName.text === 'ObjectConstructor' &&
+        this.src.program.isSourceFileDefaultLibrary(superDeclaration.node.getSourceFile())) {
```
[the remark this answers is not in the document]
JoostK: I agree with George here, let's put this whole thing in a function so we can have some early returns and assign intermediate variables where it makes sense.

packages/compiler-cli/ngcc/src/host/esm5_host.ts
```diff
@@ -8,10 +8,10 @@
 
 import * as ts from 'typescript';
 
-import {ClassDeclaration, ClassMember, ClassMemberKind, Declaration, Decorator, FunctionDefinition, Parameter, TsHelperFn, isNamedVariableDeclaration, reflectObjectLiteral} from '../../../src/ngtsc/reflection';
+import {ClassDeclaration, ClassMember, ClassMemberKind, Declaration, Decorator, FunctionDefinition, isNamedVariableDeclaration, Parameter, reflectObjectLiteral, TsHelperFn,} from '../../../src/ngtsc/reflection';
```
JoostK:
```suggestion
import {ClassDeclaration, ClassMember, ClassMemberKind, Declaration, Decorator, FunctionDefinition, isNamedVariableDeclaration, Parameter, reflectObjectLiteral, TsHelperFn} from '../../../src/ngtsc/reflection';
```

packages/compiler-cli/src/ngtsc/partial_evaluator/test/utils.ts
```diff
@@ -16,17 +16,20 @@ import {getDeclaration, makeProgram} from '../../testing';
 import {ForeignFunctionResolver, PartialEvaluator} from '../src/interface';
 import {ResolvedValue} from '../src/result';
 
-export function makeExpression(code: string, expr: string, supportingFiles: TestFile[] = []): {
+export function makeExpression(
+    code: string, expr: string, supportingFiles: TestFile[] = [], checkForErrors?: boolean): {
```
JoostK: Is this change still needed?

packages/compiler-cli/ngcc/src/host/esm2015_host.ts
```diff
@@ -353,6 +353,25 @@ export class Esm2015ReflectionHost extends TypeScriptReflectionHost implements N
...
+      return {
+        builtin: BuiltinDeclaration.JsGlobalObject,
+        expression: id,
+        viaModule: null,
+        node: null,
+      };
+    }
```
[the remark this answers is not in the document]
petebacondarwin: I think the idea is to keep ngtsc as lightweight as possible?

packages/compiler-cli/src/ngtsc/partial_evaluator/src/builtin.ts
```diff
@@ -40,3 +40,27 @@ export class ArrayConcatBuiltinFn extends BuiltinFn {
...
+    } else if (!(target instanceof Map)) {
+      return DynamicValue.fromUnsupportedSyntax(node);
+    }
+    for (let i = 1; i < args.length; i++) {
+      const source = args[i];
+      if (source instanceof DynamicValue) {
+        return DynamicValue.fromDynamicInput(node, source);
+      } else if (!(source instanceof Map)) {
+        return DynamicValue.fromUnsupportedSyntax(node);
+      }
```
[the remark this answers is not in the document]
devversion: great idea. thx!

packages/compiler-cli/ngcc/src/host/esm2015_host.ts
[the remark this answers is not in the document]
devversion: The idea was that we just want to support `__assign` and `Object.assign` in ngcc for backwards compatibility. Previously the spread operator worked fine because the actual spread AST got put into the `metadata.json`. So, explicitly using `Object.assign` didn't work in ngc either.

I don't feel strongly about it. I just don't feel like there is a compelling reason to put it into ngtsc. Happy to change it if you all feel strongly about it.

packages/compiler-cli/ngcc/src/host/esm5_host.ts
```diff
@@ -659,6 +659,8 @@ function getTsHelperFn(node: ts.NamedDeclaration): TsHelperFn|null {
       return TsHelperFn.Spread;
     case '__spreadArrays':
       return TsHelperFn.SpreadArrays;
+    case '__assign':
```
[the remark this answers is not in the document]
devversion: 😄

packages/compiler-cli/ngcc/src/host/esm5_host.ts
```diff
@@ -8,10 +8,10 @@
 
 import * as ts from 'typescript';
 
-import {ClassDeclaration, ClassMember, ClassMemberKind, Declaration, Decorator, FunctionDefinition, Parameter, TsHelperFn, isNamedVariableDeclaration, reflectObjectLiteral} from '../../../src/ngtsc/reflection';
+import {ClassDeclaration, ClassMember, ClassMemberKind, Declaration, Decorator, FunctionDefinition, isNamedVariableDeclaration, Parameter, reflectObjectLiteral, TsHelperFn,} from '../../../src/ngtsc/reflection';
 import {getNameText, hasNameIdentifier, stripDollarSuffix} from '../utils';
 
-import {Esm2015ReflectionHost, ParamInfo, getPropertyValueFromSymbol, isAssignment, isAssignmentStatement} from './esm2015_host';
+import {Esm2015ReflectionHost, getPropertyValueFromSymbol, isAssignment, isAssignmentStatement, ParamInfo} from './esm2015_host';
```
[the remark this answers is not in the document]
devversion: wouldn't clang-format complain if it would be case-sensitive? I'm changing it, but I don't think it's a good rule if it's not enforced somehow. FWIW: Webstorm ordered it that way for me.

packages/compiler-cli/src/ngtsc/partial_evaluator/src/builtin_declaration.ts
```diff
@@ -0,0 +1,30 @@
...
+ * the known JavaScript global `Object` will resolve to a `Map` that provides the
+ * `assign` method with a builtin function. This enables evaluation of `Object.assign`.
+ */
+export function resolveBuiltinDeclaration(builtinDecl: BuiltinDeclaration): ResolvedValue {
+  switch (builtinDecl) {
+    case BuiltinDeclaration.JsGlobalObject:
+      return jsGlobalObjectValue;
+    default:
+      throw new Error(
+          `Cannot resolve unknown builtin declaration ${BuiltinDeclaration[builtinDecl]}`);
```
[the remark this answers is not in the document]
devversion: See comment above.

packages/compiler-cli/src/ngtsc/partial_evaluator/test/utils.ts
devversion: good catch. removed

packages/compiler-cli/src/ngtsc/reflection/src/host.ts
```diff
@@ -340,6 +340,20 @@ export enum TsHelperFn {
...
+}
+
+/**
+ * Possible declarations which are known.
+ */
+export enum BuiltinDeclaration {
```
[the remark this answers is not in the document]
devversion: done. agreed

packages/compiler-cli/ngcc/src/packages/entry_point_bundle.ts
```diff
@@ -50,8 +50,7 @@ export function makeEntryPointBundle(
   const rootDir = entryPoint.package;
   const options: ts.CompilerOptions = {
     allowJs: true,
-    maxNodeModuleJsDepth: Infinity,
-    noLib: true, rootDir, ...pathMappings
+    maxNodeModuleJsDepth: Infinity, rootDir, ...pathMappings
```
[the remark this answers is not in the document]
devversion: I didn't. It has been enabled in the past too, but just seems to have been disabled when the mocked file system has been introduced. Can surely do some measurements if needed.

devversion:
@gkalpak @JoostK Thanks for reviewing. Addressed the feedback.

packages/compiler-cli/ngcc/src/host/esm2015_host.ts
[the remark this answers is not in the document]
gkalpak: Makes sense. Let's keep it in ngcc only for now.

packages/compiler-cli/ngcc/src/host/esm5_host.ts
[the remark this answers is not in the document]
gkalpak: I am pretty sure clang-format used to put uppercase imports before lowercase ones.
Maybe it doesn't care any more 🤷‍♂

packages/compiler-cli/ngcc/src/packages/entry_point_bundle.ts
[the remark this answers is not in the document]
gkalpak: Fair enough. If it has been there before, it can't be too bad :grin:

@petebacondarwin, any idea why was this removed in https://github.com/angular/angular/commit/16d7dde2ad89eb4610c74da63e0e4b00afb8b577#diff-ac19368dcc489893e10fcaa6ff9436aeR48 (just to make sure there is nothing we should take into account)?

gkalpak (approved):
LGTM! Great work, @devversion :100:

(I still wonder about `noLib` (https://github.com/angular/angular/pull/34661#discussion_r367637113), but that shouldn't block the PR from landing.)

JoostK (approved):
Excellent work!

devversion:
@gkalpak I can do some measurements, but I doubt that `noLib` affects performance significantly.

gkalpak:
I am not concerned about performance. My question (https://github.com/angular/angular/pull/34661#discussion_r367637113) was more about whether thare might any other reason to add/remove it.

devversion:
Fair enough. Maybe @petebacondarwin remembers why the change has been made. IIRC we already chatted about this on Slack, and it seemed like it was because of the mocked file system which did not contain the TypeScript default library files.

petebacondarwin:
I believe I did add it for one of two reasons:
a) it would be faster to process if we loaded in as few files into the program as possible
b) the mock file systems did not include these files

If everything works as expected with this removed, and there is no significant performance penalty, then I am happy.

packages/compiler-cli/ngcc/src/packages/entry_point_bundle.ts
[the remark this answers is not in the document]
alxhub: Probably because it was a fair bit of work to introduce the default library typings into the abstract filesystem.

I doubt including an extra .d.ts file or two will cause any performance issues.

packages/compiler-cli/src/ngtsc/partial_evaluator/src/known_declaration.ts
```diff
@@ -0,0 +1,29 @@
...
+ * Resolves the specified known declaration to a resolved value. For example,
+ * the known JavaScript global `Object` will resolve to a `Map` that provides the
+ * `assign` method with a builtin function. This enables evaluation of `Object.assign`.
+ */
+export function resolveKnownDeclaration(decl: KnownDeclaration): ResolvedValue {
+  switch (decl) {
+    case KnownDeclaration.JsGlobalObject:
+      return jsGlobalObjectValue;
+    default:
+      throw new Error(`Cannot resolve known declaration. Received: ${decl}.`);
```
alxhub:
```suggestion
      throw new Error(`Cannot resolve known declaration. Received: ${KnownDeclaration[decl]}.`);
```
(prints the name, not the numeric value)

alxhub (approved):

AndrewKushnir:
[Presubmit](http://test/OCL:291025335:BASE:291025371:1579730515280:5bc45ba7)

AndrewKushnir:
@devversion, could you please rebase this PR? Thank you.

devversion:
@AndrewKushnir Rebased!

gkalpak (approved):

Issue #28443, opened 2019-01-30
Service worker cache.match options


vitaliy-patsay:
# 🚀 feature request

### Relevant Package
@angular/service-worker
@angular/pwa

### Description
At the moment there is no support for [cache.match options](https://developer.mozilla.org/en-US/docs/Web/API/Cache/match)  in listed above package. I have a scenario with images which have some dynamic hashes as url parameters. 

### Desired behaviour 
The same image 

- https://cdn.example.com/image1.png?foo=bar1
- https://cdn.example.com/image1.png?foo=bar2
- https://cdn.example.com/image1.png?foo=bar3

to be cached as 

- https://cdn.example.com/image1.png

### Describe the solution you'd like
It's good to have possibility to specify some properties in ngsw-config.json assetGroup like
```
interface AssetGroup {
  name: string;
  installMode?: 'prefetch' | 'lazy';
  ...  
  cacheOptions?: {
    ignoreSearch?: boolean;
    ignoreMethod?: boolean;
    ignoreVary?: boolean;
  };
  ...
}

```

tpiros:
I have a similar scenario which warrants for adding especially `ignoreVary` to the `AssetGroup` interface. I am using images coming from Cloudinary who have a special flag called `f_auto` to dynamagically return the right image format to the browser - i.e. WebP for Chrome, JPG or JPEG 2000 to IE so on and so forth. When requesting an image asset with the `f_auto` flag it is adding the `Vary: User-Agent` header to the response which means that even though I have an item cached it will not return anything and instead - in offline mode - will try to access the image, resulting in a 504 Gateway error.

spike-rabbit:
As were are facing a similiar issue in my project I would like to provide a pull request.
Suggested API:
```ts
interface AssetGroup {
  name: string;
  installMode?: 'prefetch' | 'lazy';
  ...  
  cacheMatchOptions?: ClientMatchOptions;
  ...
}

export interface DataGroup {
  name: string;
  ...
  cacheMatchOptions?: ClientMatchOptions,
  cacheConfig: {
    ...
  };
}

// from DOM lib...
export interface ClientMatchOptions {
  ignoreMethod?: boolean;
  ignoreSearch?: boolean;
  ignoreVary?: boolean;
}
```

Since the ngsw always creates new requests for fetching resources without copying the original headers I would also suggest to provide default ClientMatchOptions = `{ignoreVary: true}`. Otherwise requests will never match any cached resources if a header which is listed in Vary is present.

[...]

PR #34663, opened 2020-01-07
feat(service-worker): support cache match options in ngsw config

spike-rabbit:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #28443

## What is the new behavior?
Cache#match options can be configured for assets and data groups.
Default options: `{ignoreVary: true}` 

- [ ] Yes
- [x] No


aio/content/guide/service-worker-config.md
```diff
@@ -123,7 +123,7 @@ These options are used to modify that matching behavior of requests.
 
 * `ignoreMethod` prevents validating of http method. Default false
 
-* `ignoreVary` do not perform VARY header matching. Default false
+* `ignoreVary` do not perform VARY header matching. Default true
```
gkalpak: Besides what's mentioned in the commit, by reading [this article][1] (and the linked resources, including [this Chrome bug][2]), I believe that the SW would not really work correctly with `Vary` headers anyway. Given that and the fact that not all browsers implement `Vary` headers correctly wrt ServiceWorkers and Caches (see aforementioned Chrome bug), I wonder if it might be better to always use `ignoreVary : true` and not allow the user to change it.

Or at least document that some SW functionalities will not work with `ignoreVary: false` if the server does indeed send `Vary` headers in responses.

[1]: https://www.smashingmagazine.com/2017/11/understanding-vary-header
[2]: https://bugs.chromium.org/p/chromium/issues/detail?id%3D756796

packages/service-worker/config/src/generator.ts
```diff
@@ -150,3 +150,14 @@ function withOrderedKeys<T extends{[key: string]: any}>(unorderedObj: T): T {
   Object.keys(unorderedObj).sort().forEach(key => orderedObj[key] = unorderedObj[key]);
   return orderedObj as T;
 }
+
+function buildCacheMatchOptions(inOptions?: CacheQueryOptions): CacheQueryOptions {
+  if (!inOptions) {
+    return {ignoreVary: true};
```
gkalpak: Actually, I believe this is a breaking change. Even if the SW would not work correctly with `Vary` headers before, it would at least not serve a response it was not meant to. With this change, the SW may start serving cached responses for requests with different headers when it is not supposed to (according to the responses `Vary` headers).

So, we need a [breaking change notice][1] in the commit message and need to land this in a new major version (and I'm afraid we may have missed the 9.0 train 😞).

[1]: https://github.com/angular/angular/blob/24b2f1da2/CONTRIBUTING.md#footer

gkalpak (changes requested):
Thx for putting this together, @spike-rabbit :heart:

I have left a few comments/suggestions, but this is off to a great start. Ultimately, we need to have more thorough testing, but let's agree on the implementation details first :smiley:

gkalpak:
Also, I think we need to pass `ignoreVary` to `delete()` calls too; e.g. here:

https://github.com/angular/angular/blob/24b2f1da2bfaa8afa6e01920d94e00574fd4d5a3/packages/service-worker/worker/src/data.ts#L564-L565

aio/content/guide/service-worker-config.md
spike-rabbit: With the current SW I don't see any way how to use `{ignoreVary: true}` beneficial. I just exposed the whole CacheQueryOptions without actually questioning if it makes sense.

For AssetsGroups it (currently) makes no sense due to the previous described reasons.

The issue described in the commit does not apply to DataGroups. But based on your linked resources and the fact that DataGroups are always using the URL as key internally I expect that `{ignoreVary: true}` would propably cause some unwanted behaviors if you're application actually relies on `Vary`.

So, I think you're right. It's better to always ignore Vary and not allow the user to change it

packages/service-worker/worker/testing/cache.ts
```diff
@@ -117,6 +117,19 @@ export class MockCache {
...
     if (res !== undefined) {
+      let varyHeaders = res.headers.get('Vary');
+      if (varyHeaders && (!options || !options.ignoreVary)) {
+        if (typeof request === 'string') {
+          return null !;
+        }
+        for (let varyHeader of varyHeaders.split(/, ?/)) {
+          let requestVary = request.headers.get(varyHeader);
+          let resVary = res.headers.get(varyHeader);
+          if (requestVary !== resVary) {
```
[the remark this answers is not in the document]
spike-rabbit: ok, I understand what is wrong here.
But If we decide to always set `ignoreVary: true` wouldn't it be sufficient to assert `ignoreVary === true` and drop my changes performing the vary check`?

Otherwise the MockCache needs to be modified heavily as it's currently not storing the cached Request. I don't think the effort is worth it since a Cache#match will never take Vary headers into account.

packages/service-worker/config/src/generator.ts
spike-rabbit: I don't like this (especially that we missed 9). But I understand your point. Once all other tasks regarding the implementation details are closed I will rebase -i my commits and add a breaking change notice

aio/content/guide/service-worker-config.md
```diff
@@ -110,6 +115,16 @@ This section describes the resources to cache, broken up into the following grou
 * `urls` includes both URLs and URL patterns that will be matched at runtime. These resources are not fetched directly and do not have content hashes, but they will be cached according to their HTTP headers. This is most useful for CDNs such as the Google Fonts service.<br>
   _(Negative glob patterns are not supported and `?` will be matched literally; i.e. it will not match any character other than `?`.)_
 
+### `cacheMatchOptions`
+
+These options are used to modify that matching behavior of requests.
+
+* `ignoreSearch` ignore query parameters: Default: false
+
+* `ignoreMethod` prevents validating of http method. Default false
```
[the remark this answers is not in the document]
spike-rabbit: You're right. To get other methods working more changes would be necessary, so I will remove ignoreMethod for both as this is out of the PRs scope.

But anyway I guess this is a valid use case:
In our company we have some applications that are fetching cachable resources via POST. Those resources are generated based on some input parameters. In order to prevent the creation and sharing of links with malicious parameters POST is used as you cannot create a link that performs a post.

spike-rabbit:
thanks for taking care of this PR @gkalpak

 I did most of the changes you requested. Can you please decide on how to proceed with ignoreVary so that I can update the missing ones as well

aio/content/guide/service-worker-config.md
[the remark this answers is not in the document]
gkalpak: Sounds good. Happy to explore supporting `ignoreMethod` in a future PR.

aio/content/guide/service-worker-config.md
```diff
@@ -74,6 +74,11 @@ interface AssetGroup {
     files?: string[];
     urls?: string[];
   };
+  cacheMatchOptions?: {
```
gkalpak, line 77: You missed one :smiley:

```suggestion
  cacheQueryOptions?: {
```

packages/service-worker/worker/testing/cache.ts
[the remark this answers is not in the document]
gkalpak: Dropping these changes (and just asserting `ignoreVary: true` sgtm based on [this discussion][1].

[1]: https://github.com/angular/angular/pull/34663#discussion_r370741772

gkalpak (changes requested):
Sorry for not realizing sooner that this was waiting for my input, @spike-rabbit 😅
On the bright side, master only opened up for v10 changes a few days ago (so this couldn't have gone in sooner anyway :grin:).

Are you still interested/available to work on this?

If so, I think we can move forward with only making `ignoreSearch` configurable and enforcing `ignoreVary: true` (because it is a better default and `ignoreVary: false` won't work correctly for the SW anyway) and document it as a (minor) breaking change (even though it will not affect most people (if any)).

spike-rabbit:
yes, I am still interested in working on this PR.

I will do changes you requested within the next days

gkalpak:
@spike-rabbit, is this ready for another review?

spike-rabbit:
@gkalpak I am not sure. The changes you request are all done except for the breaking change notice.

But the pipeline failed due to formatting errors, so I reformatted the code with the command provided in pipeline output. Unfortunately, this modified dozens lines of code.

packages/service-worker/config/test/generator_spec.ts
```diff
@@ -92,6 +92,7 @@ describe('Generator', () => {
           '\\/some\\/url\\?with\\+escaped\\+chars',
           '\\/test\\/relative\\/[^/]*\\.txt',
         ],
+        cacheQueryOptions: {ignoreVary: true}
```
gkalpak, line 95: Can you also add some tests with a config that contains `cacheQueryOptions`?

packages/service-worker/worker/test/data_spec.ts
```diff
@@ -6,6 +6,8 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
+import {MockCache} from '@angular/service-worker/worker/testing/cache';
+
 import {CacheDatabase} from '../src/db-cache';
 import {Driver} from '../src/driver';
 import {Manifest} from '../src/manifest';
```
gkalpak, line 13:
```suggestion
import {CacheDatabase} from '../src/db-cache';
import {Driver} from '../src/driver';
import {Manifest} from '../src/manifest';
import {MockCache} from '../testing/cache';
```

packages/service-worker/worker/test/data_spec.ts
```diff
@@ -14,283 +16,315 @@ import {MockFileSystemBuilder, MockServerStateBuilder, tmpHashTableForFs} from '
...
+  dataGroups: [
+    {
+      name: 'testPerf',
+      maxSize: 3,
+      strategy: 'performance',
+      patterns: ['^/api/.*$'],
+      timeoutMs: 1000,
+      maxAge: 5000,
+      version: 1,
+      cacheQueryOptions: {ignoreVary: true},
```
gkalpak, line 72: Can you add `ignoreSearch` to one of the `cacheQueryOptions` and verify that the appropriate `cacheQueryOptions` are used for different requests.

packages/service-worker/worker/test/data_spec.ts
```diff
@@ -14,283 +16,315 @@ import {MockFileSystemBuilder, MockServerStateBuilder, tmpHashTableForFs} from '
...
+      expect(await driver.checkForUpdate()).toEqual(true);
+      await driver.updateClient(await scope.clients.get('default'));
+      expect(await makeRequest(scope, '/api/test')).toEqual('version 2');
+    });
+
+    it('CacheQueryOptions are passed through', async () => {
       await driver.initialized;
+      const matchSpy = spyOn(MockCache.prototype, 'match').and.callThrough();
+      await makeRequest(scope, '/api/a');
+      await makeRequest(scope, '/api/a');
```
gkalpak, line 205: Can you add a comment on why we need two requests?

packages/service-worker/worker/test/data_spec.ts
```diff
@@ -14,283 +16,315 @@ import {MockFileSystemBuilder, MockServerStateBuilder, tmpHashTableForFs} from '
...
+
+      // Pause the server to force the cached (opaque) response to be returned.
+      serverUpdate.pause();
+      const [res2] = makePendingRequest(scope, '/fresh/data');
+      await serverUpdate.nextRequest;
+      scope.advance(2000);
+
+      expect(await res2).toBe('');
+    });
+    it('CacheQueryOptions are passed through when falling back to cache', async () => {
```
gkalpak, line 306:
```suggestion

    it('CacheQueryOptions are passed through when falling back to cache', async () => {
```

packages/service-worker/worker/test/prefetch_spec.ts
```diff
@@ -6,84 +6,103 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
+import {MockCache} from '@angular/service-worker/worker/testing/cache';
+import {MockRequest} from '@angular/service-worker/worker/testing/fetch';
+
 import {PrefetchAssetGroup} from '../src/assets';
 import {CacheDatabase} from '../src/db-cache';
 import {IdleScheduler} from '../src/idle';
```
gkalpak, line 14:
```suggestion
import {PrefetchAssetGroup} from '../src/assets';
import {CacheDatabase} from '../src/db-cache';
import {IdleScheduler} from '../src/idle';
import {MockCache} from '../testing/cache';
import {MockRequest} from '../testing/fetch';
```

packages/service-worker/worker/test/prefetch_spec.ts
```diff
@@ -6,84 +6,103 @@
...
+    const badServer =
+        new MockServerStateBuilder().withManifest(manifest).withStaticFiles(badHashFs).build();
+    const badScope = new SwTestHarnessBuilder().withServerState(badServer).build();
+    group = new PrefetchAssetGroup(
+        badScope, badScope, idle, manifest.assetGroups![0], tmpHashTable(manifest),
+        new CacheDatabase(badScope, badScope), 'test');
+    const err = await errorFrom(group.initializeFully());
+    expect(err.message).toContain('Hash mismatch');
+  });
+  it('respond with a resource even if a header listed in vary is present', async () => {
```
gkalpak, line 90: Does this test add any value, given that it is handled by a mock version of the cache that ignores options anyway?

packages/service-worker/worker/src/data.ts
```diff
@@ -561,8 +564,8 @@ export class DataGroup {
   private async clearCacheForUrl(url: string): Promise<void> {
     const [cache, ageTable] = await Promise.all([this.cache, this.ageTable]);
     await Promise.all([
-      cache.delete(this.adapter.newRequest(url, {method: 'GET'})),
-      cache.delete(this.adapter.newRequest(url, {method: 'HEAD'})),
+      cache.delete(this.adapter.newRequest(url, {method: 'GET'}), this.config.cacheQueryOptions),
+      cache.delete(this.adapter.newRequest(url, {method: 'HEAD'}), this.config.cacheQueryOptions),
       ageTable.delete(url),
```
gkalpak, line 569: Actually, I think we also need to take `ignoreSearch` into account in the `ageTable` (and potentially other `CacheTable`s/methods). Because, if `/api/foo` and `/api/foo?bar` are treated as different entries in the cache, then we need to also track their metadata (such as age) separately.

gkalpak (changes requested):
Thx, @spike-rabbit :+1: I've left some more comments, but it's looking really good overall.
Note to self: We still need the breaking change notice.

packages/service-worker/worker/src/data.ts
spike-rabbit: The cacheQueryOptions are now also passed to the LruTable and AgeTable. I guess this is sufficient as the cache within the tables will handle the matching.

packages/service-worker/worker/test/prefetch_spec.ts
spike-rabbit: no. This was just a leftover from my previous attempt to test vary header. I removed it.

spike-rabbit:
@gkalpak this is ready for next review. Every requested change should be done (except for the breaking change notice).

kapunahelewong (approved):
Thanks for updating the doc, too, @spike-rabbit! 😊

packages/service-worker/config/src/generator.ts
```diff
@@ -143,8 +146,16 @@ function joinUrls(a: string, b: string): string {
...
+  const orderedObj = {} as {[key: string]: any};
   Object.keys(unorderedObj).sort().forEach(key => orderedObj[key] = unorderedObj[key]);
   return orderedObj as T;
 }
+
+function buildCacheQueryOptions(inOptions?: Pick<CacheQueryOptions, 'ignoreSearch'>):
+    CacheQueryOptions {
+  return {
+    ...inOptions,
+    ignoreVary: true,
```
gkalpak, line 159: I was thinking about this some more. I think it might be a good idea to actually put `ignoreVary: true` first. This way, we offer an escape hatch for the unlikely case that some people were relying in th previous behavior of `ignoreVary: false`.

We should keep everything else the same: Document that `only `ignoreSearch` is supported, `ignoreVary` is set to `true`, etc.
But if someone really needs it, there will be a way to have `ignoreVary: false` for their groups.

WDYT?

gkalpak (changes requested):
Thx again, @spike-rabbit!
I've left a few more comments. We are really close :wink:

Can you also rebase on master and add `Fixes #36638` and a breaking change notice at the bottom of the commit message of the commit that turns on `ignoreVary`. Example commit message:

> feat(service-worker): use \`ignoreVary: true\` when retrieving responses from cache
>
> The Angular ServiceWorker always uses a copy of the request without
> headers for caching assets (in order to avoid issues with opaque
> responses). Therefore, it was previously not possible to retreive
> resources from the cache if the response contained [Vary][1] headers.
>
> In addition to that, \`Vary\` headers do not work in all browsers (or work
> differently) and may not work as intended with ServiceWorker caches. See
> [this article][2] and the linked resources for more info.
>
> This commit avoids the aforementioned issues by making sure the Angular
> ServiceWorker always sets the \`ignoreVary\` option passed to
> [Cache#match()][3] to \`true\`. This allows the ServiceWorker to correctly
> retrieve cached responses with \`Vary\` headers, which was previosuly not
> possible.
>
> [1]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Vary
> [2]: https://www.smashingmagazine.com/2017/11/understanding-vary-header
> [3]: https://developer.mozilla.org/en-US/docs/Web/API/Cache/match
>
> Fixes #36638
>
> BREAKING CHANGE:
>
> Previously, [Vary](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Vary)
> headers would be taken into account when retrieving resources from the
> cache, completely preventing the retrieval of cached assets (due to
> ServiceWorker implementation details) and leading to unpredictable
> behavior due to inconsistent/buggy implementations in different
> browsers.
>
> Now, \`Vary\` headers are ignored when retrieving resources from the
> ServiceWorker caches, which can result in resources being retrieved even
> when their headers are different. If your application needs to
> differentiate its responses based on request headers, please make sure
> the Angular ServiceWorker is [configured](https://angular.io/guide/service-worker-config)
> to avoid caching the affected resources.

packages/service-worker/config/src/generator.ts
spike-rabbit: This sound reasonable to me. 
Wouldn't it make sense to print a warning to the console if someone is actually using `ignoreVary : true` or `ignoreMethod: true`?

packages/service-worker/worker/testing/cache.ts
```diff
@@ -116,19 +135,28 @@ export class MockCache {
     }
     // TODO: cleanup typings. Typescript doesn't know this can resolve to undefined.
     let res = this.cache.get(url);
+    if (!res && options?.ignoreSearch) {
+      // check if cache has url by ignoring search
+      url = new URL(`http://dummyhost${url}`).pathname;
+      res = Array.from(this.cache.entries())
+                .find(([key]) => url === new URL(`http://dummyhost${key}`).pathname)
+                ?.[1];
```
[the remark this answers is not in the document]
spike-rabbit: Typescript does not like your solution. So I modified it a little. Hope it's ok as well

spike-rabbit:
@gkalpak The changes you request are done. I also squashed my commits and used your commit message. 
But I have some doubts about the commit message: It's currently not mentioned that `ignoreSearch` can be configured and technically also the other two options.

packages/service-worker/config/src/in.ts
```diff
@@ -51,6 +52,9 @@ export interface DataGroup {
   urls: Glob[];
   version?: number;
   cacheConfig: {
-    maxSize: number; maxAge: Duration; timeout?: Duration; strategy?: 'freshness' | 'performance';
+    maxSize: number; maxAge: Duration;
```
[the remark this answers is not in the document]
gkalpak: I guess not 🤷

packages/service-worker/worker/testing/cache.ts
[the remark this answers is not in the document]
gkalpak: Oh, I see. TypeScript got smarter. Nice!
I still don't think `this.cache.get(matchingReq!)` is correct. Using a non-null assertion here is a hack (because `matchingReq` _can_ actually be `undefined`).

How about:

```diff
-res = this.cache.get(matchingReq!);
+if (matchingReq !== undefined) res = this.cache.get(matchingReq);
```

packages/service-worker/worker/testing/cache.ts
```diff
@@ -174,6 +189,11 @@ export class MockCache {
     });
     return dehydrated;
   }
+
+  /** remove the query/hash part from a url*/
+  private sanitizeUrl(url: string): string {
```
gkalpak, line 194: Nit: `sanitizeUrl` sounds a bit strong. I'd prefer the more explicit `stripQueryAndHash`.

gkalpak (changes requested):
Oh, no, I didn't mean you should squash all commits into one :grin:
My commit message (which I have updated to include `Fixes #36638` btw 😇) was explicitly intended for the commit that turn `ignoreVary` to true:

> Can you [...] add [...] a breaking change notice at the bottom of the commit message of the commit that turns on `ignoreVary`.

##
Basically, we need two commits:
- One that adds support for `ignoreSearch` (and the associated tests and docs).
- One that switches `ignoreVary` to true by default (and the associated tests and docs). This second commit should include the breaking change notice (as seen in my proposed commit message).

spike-rabbit:
@gkalpak the changes you requested are done and there are two commits now with the related changes.
How can I resolve the `ci/angular: size` pipeline failure? For me it makes sense that the bundle size increased as the ngsw-config includes more content by default and the ngsw got a couple of more lines as well

packages/service-worker/worker/src/data.ts
```diff
@@ -566,8 +566,8 @@ export class DataGroup {
   private async clearCacheForUrl(url: string): Promise<void> {
     const [cache, ageTable] = await Promise.all([this.cache, this.ageTable]);
     await Promise.all([
-      cache.delete(this.adapter.newRequest(url, {method: 'GET'})),
-      cache.delete(this.adapter.newRequest(url, {method: 'HEAD'})),
+      cache.delete(this.adapter.newRequest(url, {method: 'GET'}), this.config.cacheQueryOptions),
+      cache.delete(this.adapter.newRequest(url, {method: 'HEAD'}), this.config.cacheQueryOptions),
```
gkalpak, line 570: Shouldn't these be part of the first commit?

gkalpak (approved):
Thx, @spike-rabbit! This is almost ready :tada:

Regarding the commit messages:
- In the first commit, I would remove the following:
  > Although it is not documented and will not work it is technical possible provide values for `ignoreVary` and `ignoreMethod` as well.

  (Even if it is true, I don't think it is a good idea to mention it in the commit message, as it might give the wrong message. Both `ignoreVary` and `ignoreMethod` will not work 100% correctly.)

- In the second commit message, the `Fixes #36638` part needs to be above the `BREAKING CHANGE` (see https://github.com/angular/angular/pull/34663#pullrequestreview-398422351). Otherwise, the `Fixes ...` will be considered as part of the breaking change notice.

Regarding the `ci/angular: size`: This is unrelated to your PR. Can you try rebasing on master once more (this might help fix it).

spike-rabbit:
@gkalpak 
- both commit messages are updated 
- I moved the mentioned changes from commit 2 to 1
- rebased on master again

gkalpak:
Awesome work! Thank you, @spike-rabbit :heart:
This just needs a `public API` approval from @IgorMinar and is good to go :shipit:

IgorMinar (approved):
Lgtm! Thank you

Reviewed-for: public-api

PR #34664, opened 2020-01-07
test: run styling benchmarks in benchpress 

devversion:
Runs the styling benchmarks that have been added with 2e0b237
in benchpress. The goal is that these benchmarks can be wired up in
Latency Lab.

### DIFF
--- a/modules/benchmarks/src/styling/BUILD.bazel
+++ b/modules/benchmarks/src/styling/BUILD.bazel
@@ -1,0 +1,1 @@
+package(default_visibility = ["//visibility:public"])
--- a/modules/benchmarks/src/styling/BUILD.bazel
+++ b/modules/benchmarks/src/styling/BUILD.bazel
@@ -1,1 +1,15 @@
-package(default_visibility = ["//visibility:public"])
+package(default_visibility = ["//modules/benchmarks:__subpackages__"])
+
+load("//tools:defaults.bzl", "ts_library")
+
+ts_library(
+    name = "tests_lib",
+    testonly = True,
+    srcs = ["styling_perf.spec.ts"],
+    tsconfig = "//modules/benchmarks:tsconfig-e2e.json",
+    deps = [
+        "//modules/e2e_util",
+        "@npm//@types/jasminewd2",
+        "@npm//protractor",
+    ],
+)
--- a/modules/benchmarks/src/styling/ng2/index.html
+++ b/modules/benchmarks/src/styling/ng2/index.html
@@ -1,0 +1,39 @@
+<!doctype html>
+<html>
+
+<head>
+  <!-- Prevent the browser from requesting any favicon. -->
+  <link rel="icon" href="data:,">
+</head>
+
+<body>
+
+  <h2>Styling bindings benchmark</h2>
+  <p>
+    <select id="scenario-select">
+      <option value="0">(0) [title]="exp" (no styling involved)</option>
+      <option value="1">(1) class="foo" (static class)</option>
+      <option value="2">(2) class="foo {{exp}}" (className binding)</option>
+      <option value="3">(3) [class.foo]="boolExp" binding</option>
+      <option value="4">(4) class="foo" [class.bar]="boolExp" (mix of static and class. bindings</option>
+      <option value="5">(5) class="foo" [ngClass]="{bar: boolExp}" (mix of static class and ngClass binding)</option>
+      <option value="6">(6) class="foo" [ngStyle]="{width: 10px}" [style.background-color]="exp" (Sierpinski's triangle)
+      </option>
+    </select>
+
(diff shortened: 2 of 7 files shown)


modules/benchmarks/src/styling/ng2/index.html
```diff
@@ -11,9 +11,10 @@
   <h2>Styling bindings benchmark</h2>
   <p>
     <select id="scenario-select">
+      <!-- *Note*: When adding/removing scenarios, ensure the e2e test is also updated. -->
       <option value="0">(0) [title]="exp" (no styling involved)</option>
       <option value="1">(1) class="foo" (static class)</option>
-      <option value="2">(2) class="foo {{exp}}" (className binding)</option>
+      <option value="2">(2) class="foo {{exp}}" (static class with interpolation)</option>
```
pkozlowski-opensource: nit: could we drop the "static" from the name here? It is a binding, or, to be more precise, an interpolation. How about "class interpolation"?

modules/benchmarks/src/styling/styling_perf.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+const SCENARIOS = [
+  {optionIndex: 0, id: 'no_styling_involved'}, {optionIndex: 1, id: 'static_class'},
+  {optionIndex: 2, id: 'static_class_with_interpolation'}, {optionIndex: 3, id: 'class_binding'},
+  {optionIndex: 4, id: 'static_class_and_class_binding'},
+  {optionIndex: 5, id: 'static_class_and_ngclass_binding'},
+  {optionIndex: 6, id: 'static_class_and_ngstyle_binding_and_style_binding'}
+];
+
+/** Benchmarks that should run against all configured scenarios. */
+const BENCHMARKS = {
```
pkozlowski-opensource: nit: do we need this const now that we always run the same work for each scenario?

modules/benchmarks/src/styling/styling_perf.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+        // Scenario always contain work functions and an optional initialization function
+        // that can be used to perform actions without affecting the measurements.
+        const {work, init} = BENCHMARKS[benchmarkName];
+        it(`should run benchmark: ${benchmarkName}`, (done) => {
+          runStylingBenchmark(`styling.${id}.${benchmarkName}`, {
+            work,
+            prepare: () => {
+              // Switch to the current test mode by clicking the corresponding option.
+              element.all(by.tagName('option')).get(optionIndex).click();
+              if (init !== undefined) {
```
pkozlowski-opensource: Are there cases where `init` is undefined?

pkozlowski-opensource (approved):
LGTM overall. Left a couple of nit comments as I think that we can simplify the test a bit since we are doing the same work for all scenarios. 

Thnx @devversion !

modules/benchmarks/src/styling/ng2/index.html
devversion: good point. fixed.

modules/benchmarks/src/styling/styling_perf.spec.ts
devversion: Not really. I kept it since I initially had more cases like `create & detect changes` and it could've been useful in the future. I removed it for now though, since I doubt there will be more soon.

modules/benchmarks/src/styling/styling_perf.spec.ts
devversion: Related to my answer in the other comment. It was intended to have optional inits. e.g. when testing the `create` button, there are no preparation/init actions needed

devversion: Removed for simplicity 😄

modules/benchmarks/src/styling/BUILD.bazel
```diff
@@ -1 +1,15 @@
-package(default_visibility = ["//visibility:public"])
+package(default_visibility = ["//modules/benchmarks:__subpackages__"])
+
+load("//tools:defaults.bzl", "ts_library")
+
+ts_library(
+    name = "tests_lib",
+    testonly = 1,
```
IgorMinar: please use `True` instead - we prefer it because of better readability.

IgorMinar (approved):
the rest looks good to me. thanks!

modules/benchmarks/src/styling/BUILD.bazel
IgorMinar: I fixed this and forcepushed this this branch

devversion: @IgorMinar Thanks for fixing this. I was unexpectedly out yesterday, so couldn't address it myself.

PR #34668, opened 2020-01-08
docs: deprecate undecorated base classes that use Angular features

kara:
This PR adds a section to the deprecation guide about undecorated base classes that links to the existing guide.


IgorMinar (approved):
Lgtm, thanks!

PR #34683, opened 2020-01-08
refactor(ivy): move hostVars/hostAttrs from instruction to DirectiveDef

mhevery:
This PR is part of sequence of work which is meant to be reviewed together:
- YOU ARE HERE => [#34683 - refactor(ivy): move hostVars/hostAttrs from instruction to DirectiveDef](https://github.com/angular/angular/pull/34683)
- [#34708 - refactor(ivy): delete `ɵɵallocHostVars` instruction](https://github.com/angular/angular/pull/34708)
- [#34717 - refactor(ivy): delete `ɵɵelementHostAttrs` instruction](https://github.com/angular/angular/pull/34708)

# refactor(ivy): move `hostVars`/`hostAttrs` from instruction to `DirectiveDef`
    
This change moves information from instructions to declarative position:
- `ɵɵallocHostVars(vars)` => `DirectiveDef.hostVars`
- `ɵɵelementHostAttrs(attrs)` => `DirectiveDef.hostAttrs`
    
When merging directives it is necessary to know about `hostVars` and `hostAttrs`. Before this change that information was stored in the `hostBindings` function. This was problematic, because in order to get to that information the `hostBindings` would have to be executed. In order for `hostBindings` to be executed the directives would have to be instantiated. This means that the directive instantiation would happen before we had knowledge about the `hostAttrs` and as a result the directive could observe in the constructor that not all of the `hostAttrs` have been applied. This further complicates the runtime as we have to apply `hostAttrs` in parts over many invocations.
    
`ɵɵallocHostVars` was unnecessarily complicated because it would have to update the `LView` (and Blueprint) while existing directives are already executing. By moving it out of `hostBindings` function we can access it statically and we can create correct `LView` (and Blueprint) in a single pass.
    
This change only changes how the instructions are generated, but does not change the runtime much. (We cheat by emulating the old behavior by calling `ɵɵallocHostVars` and `ɵɵelementHostAttrs`) Subsequent change will refactor the runtime to take advantage of the static information.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No


packages/core/test/acceptance/inherit_definition_feature_spec.ts
```diff
@@ -7,11 +7,107 @@
...
+import {ivyEnabled} from '@angular/core/src/ivy_switch';
+import {getDirectiveDef} from '@angular/core/src/render3/definition';
+import {appendHostAttrs, mergeHostAttrs} from '@angular/core/src/render3/features/inherit_definition_feature';
+import {AttributeMarker, TAttributes} from '@angular/core/src/render3/interfaces/node';
 import {TestBed} from '@angular/core/testing';
 import {By} from '@angular/platform-browser';
 import {onlyInIvy} from '@angular/private/testing';
 
 describe('inheritance', () => {
+  describe('appendHostAttrs', () => {
```
pkozlowski-opensource: Those look like low-level tests digging into ivy internals so I would probably move those out of `acceptance` into `src/render3/features/inherit_definition_feature_spec.ts`

packages/core/src/render3/instructions/shared.ts
```diff
@@ -1204,7 +1207,18 @@ export function invokeHostBindingsInCreationMode(
   const previousExpandoLength = expando.length;
   setCurrentDirectiveDef(def);
   const elementIndex = tNode.index - HEADER_OFFSET;
-  def.hostBindings !(RenderFlags.Create, directive, elementIndex);
+  // TODO(misko-next): This is a temporary work around for the fact that we moved the information
+  // from instruction to declaration. The workaround is to just call the instruction as if it was
+  // part of the `hostAttrs`.
+  if (def.hostVars !== 0) {
+    ɵɵallocHostVars(def.hostVars);
```
pkozlowski-opensource: I guess we should rename those functions and drop the `ɵɵ` prefix as those are not the instructions any more. Same for `ɵɵelementHostAttrs`

packages/core/src/render3/instructions/shared.ts
```diff
@@ -1184,7 +1187,7 @@ function invokeDirectivesHostBindings(tView: TView, viewData: LView, tNode: TNod
     for (let i = start; i < end; i++) {
       const def = tView.data[i] as DirectiveDef<any>;
       const directive = viewData[i];
-      if (def.hostBindings) {
+      if (def.hostBindings || def.hostVars !== 0 || def.hostAttrs !== null) {
```
pkozlowski-opensource: Use `TNodeFlags.hasHostBindings` since it was already set in resolve directives?

kara (changes requested):

packages/core/src/render3/instructions/shared.ts
mhevery: These will be deleted in follow up PRs, so I don't think it is worth creating the churn. (See: https://github.com/angular/angular/pull/34708)

packages/core/test/bundling/hello_world/bundle.golden_symbols.json
```diff
@@ -137,12 +140,18 @@
...
+    "name": "addItemToStylingMap"
+  },
   {
     "name": "addToViewTree"
   },
   {
     "name": "allocLFrame"
   },
+  {
+    "name": "allocStylingMapArray"
```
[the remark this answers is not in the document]
mhevery: These will be deleted in follow up PR.

packages/core/src/render3/instructions/shared.ts
pkozlowski-opensource: oh, sure, no need to rename if those go away!

packages/core/test/acceptance/inherit_definition_feature_spec.ts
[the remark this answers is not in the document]
mhevery: moved

packages/core/src/render3/features/inherit_definition_feature.ts
```diff
@@ -100,6 +117,104 @@ export function ɵɵInheritDefinitionFeature(definition: DirectiveDef<any>| Comp
...
+    for (let i = 0; i < src.length; i++) {
+      const item = src[i];
+      if (typeof item === 'number') {
+        srcMarker = item;
+      } else {
+        if (srcMarker === AttributeMarker.Classes) {
+          appendHostAttrs(dst, srcMarker, item as string, '');
+        } else {
+          ngDevMode && assertEqual(
+                           srcMarker === -1 || srcMarker === AttributeMarker.Styles, true,
```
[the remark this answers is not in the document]
mhevery: fixed

packages/core/src/render3/instructions/shared.ts
mhevery: That is not possible. The if statement is on per directive basis, where as `TNodeFlags.hasHostBindings` is on `TNode` level. So it will either flag all or no directives. In subsequent PRs some of the `||` portions will be removed so it will become `if (def.hostBindings !== null)`

kara (approved):
LGTM

packages/core/src/render3/features/inherit_definition_feature.ts
```diff
@@ -6,17 +6,26 @@
...
 
 export function getSuperType(type: Type<any>): Type<any>&
     {ɵcmp?: ComponentDef<any>, ɵdir?: DirectiveDef<any>} {
   return Object.getPrototypeOf(type.prototype).constructor;
 }
 
+/// Infer writable `DirectiveDef` type.
+/// This will be tree shaken because it is reference as type only.
+const writableDefValue = false as true &&
+    (null as any as returnWritableType)(null as any as DirectiveDef<any>| ComponentDef<any>);
```
JoostK: Instead of the weird trick with the `returnWritableType` function type, you could use a mapped type instead that removes the readonly marker:

```ts
type Writable<T> = { -readonly [K in keyof T]: T[K] };
```

mhevery: WHAT! That is possible! I love it. Will change

mhevery: Thanks @JoostK I learned something new!

packages/core/src/render3/features/inherit_definition_feature.ts
```diff
@@ -6,17 +6,26 @@
...
 import {isComponentDef} from '../interfaces/type_checks';
+import {mergeHostAttrs} from '../util/attrs_utils';
 
 export function getSuperType(type: Type<any>): Type<any>&
     {ɵcmp?: ComponentDef<any>, ɵdir?: DirectiveDef<any>} {
   return Object.getPrototypeOf(type.prototype).constructor;
 }
 
+/// Infer writable `DirectiveDef` type.
+/// This will be tree shaken because it is reference as type only.
```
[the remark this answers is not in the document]
mhevery: removed per @JoostK  suggestion

mhevery:
merged in 82a41af062

PR #34686, opened 2020-01-09
refactor(compiler): refactor and export Bazel host creation.

dgp1130:
## PR Type

Issue Number: N/A

## What is the new behavior?

Refactored Bazel host creation to export some of the relevant utility functions. google3 needs to make some modifications to the Bazel host to enable the Ivy migration. The functionality exported here is still needed in g3, but it allows g3 code to create the Bazel host, make necessary modifications, and then provide it to the external codebase here while everything is still compatible. For external users, this is a no-op.

- [X] No

### DIFF
--- a/packages/bazel/src/ngc-wrapped/index.ts
+++ b/packages/bazel/src/ngc-wrapped/index.ts
@@ -7,7 +7,7 @@
  */
 
 import * as ng from '@angular/compiler-cli';
-import {BazelOptions, CachedFileLoader, CompilerHost, FileCache, FileLoader, UncachedFileLoader, constructManifest, debug, parseTsconfig, resolveNormalizedPath, runAsWorker, runWorkerLoop} from '@bazel/typescript';
+import {BazelOptions, CachedFileLoader, CompilerHost as BazelHost, FileCache, FileLoader, UncachedFileLoader, constructManifest, debug, parseTsconfig, resolveNormalizedPath, runAsWorker, runWorkerLoop} from '@bazel/typescript';
 import * as fs from 'fs';
 import * as path from 'path';
 import * as tsickle from 'tsickle';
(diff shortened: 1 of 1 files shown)


packages/bazel/src/ngc-wrapped/index.ts
```diff
@@ -433,6 +379,68 @@ export function compile({allDepsCompiledWithBazel = true, compilerOpts, tsHost,
   return {program, diagnostics};
 }
 
+/** A module resolver for handling generated files in Bazel. */
+export function generatedFileModuleResolver(
```
kyliau: Is this purely moving code around, or are there changes to the logic as well?

dgp1130: This *should* be a pure refactor. I did change the syntax slightly to be consistent will how I'll be doing most of the host wrapping, but I believe it should not affect functionality.

dgp1130:
I also ran a global presubmit of this commit within google3 and only encountered 4 failures, which appear to be flakes.

kyliau (approved):

kyliau:
Changing the target from patch-only to master & patch. It'll make it easier to cherry pick changes to patch branch in case subsequent PRs need to target patch.

dgp1130:
Sounds good @kyliau. I also forced the google3 tests to pass as the only failures were a couple flakes. Should be ready to merge.

IgorMinar (approved):
the commit message explaining the reasoning for this change is quite sparse, but the PR has a great description. @dgp1130 can you please update the commit message to contain the explanation from the PR description?

I'm fine with this going to patch branch since it affects only bazel integrations for which we try to keep the diff between master and patch minimal

thanks!

dgp1130:
Updated commit message to include the description from this PR.

atscott:
[presubmit](https://test.corp.google.com/ui#id=OCL:289127981:BASE:289699733:1579031725440:fefc88a9)

PR #34687, opened 2020-01-09
fix(ivy): Prevent errors when querying DebugElement roots that were o…

atscott:
…utside angular context (or with Renderer2)

DebugElement.query also matches elements that may have been created
outside of Angular (ex: with `document.appendChild`). If those matched
DebugElements are in turn used to query for more elements, an error
occurs because the first step in queryAll is to load the LContext.

### DIFF
--- a/packages/core/src/debug/debug_node.ts
+++ b/packages/core/src/debug/debug_node.ts
@@ -470,10 +470,14 @@ function _queryAllR3(
 function _queryAllR3(
     parentElement: DebugElement, predicate: Predicate<DebugElement>| Predicate<DebugNode>,
     matches: DebugElement[] | DebugNode[], elementsOnly: boolean) {
-  const context = loadLContext(parentElement.nativeNode) !;
-  const parentTNode = context.lView[TVIEW].data[context.nodeIndex] as TNode;
-  _queryNodeChildrenR3(
-      parentTNode, context.lView, predicate, matches, elementsOnly, parentElement.nativeNode);
+  const context = loadLContext(parentElement.nativeNode, false);
+  if (!context) {
+    _queryNativeNodeDescendants(parentElement.nativeNode, predicate, matches, elementsOnly);
+  } else {
+    const parentTNode = context.lView[TVIEW].data[context.nodeIndex] as TNode;
+    _queryNodeChildrenR3(
+        parentTNode, context.lView, predicate, matches, elementsOnly, parentElement.nativeNode);
+  }
 }
 
 /**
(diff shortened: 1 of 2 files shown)


packages/core/test/debug/debug_node_spec.ts
```diff
@@ -664,6 +666,13 @@ class TestCmptWithPropInterpolation {
 
       it('when searching by injector',
          () => { expect(() => el.query(e => e.injector === null)).not.toThrow(); });
+
+      onlyInIvy('VE does not match elements created outside Angular context')
```
pkozlowski-opensource: I'm kind of torn here: maybe we should match VE behaviour for now and skip those elements as well? WDYT?

atscott: I was a little torn as well, but I also just realized that in Ivy, we don't (as far as I know, so correct me if I'm wrong) have a way of differentiating between nodes that were created with `Renderer2` and nodes created with `document.createElement`. This poses a problem because VE would be able to find `Renderer2` created elements but not `document.create/appendChild`. So Ivy either has to be able to find both or neither since it doesn't know the difference. I added another test case to the spec to document this behavior.

AndrewKushnir (approved):

packages/core/src/debug/debug_node.ts
```diff
@@ -470,10 +470,14 @@ function _queryAllR3(
 function _queryAllR3(
     parentElement: DebugElement, predicate: Predicate<DebugElement>| Predicate<DebugNode>,
     matches: DebugElement[] | DebugNode[], elementsOnly: boolean) {
-  const context = loadLContext(parentElement.nativeNode) !;
-  const parentTNode = context.lView[TVIEW].data[context.nodeIndex] as TNode;
-  _queryNodeChildrenR3(
-      parentTNode, context.lView, predicate, matches, elementsOnly, parentElement.nativeNode);
+  const context = loadLContext(parentElement.nativeNode, false);
+  if (!context) {
```
pkozlowski-opensource: nit: `if (context !== null)` ?
* more explicit check
* "positive" case first?

packages/core/src/debug/debug_node.ts
```diff
@@ -470,10 +470,14 @@ function _queryAllR3(
...
+    _queryNativeNodeDescendants(parentElement.nativeNode, predicate, matches, elementsOnly);
```
pkozlowski-opensource: leave a comment briefly describing when `LContext` is not available?

pkozlowski-opensource (approved):
LGTM, left a couple of nit comments that you might want to have a look at.

packages/core/src/debug/debug_node.ts
atscott: Done.

packages/core/src/debug/debug_node.ts
atscott: Done.

atscott:
[presubmit](http://test/OCL:288803454:BASE:289443388:1578931633014:6b7c1933)

PR #34690, opened 2020-01-09
fix(compiler): record correct end of expression

### LINKED ISSUE #433: Diagnostics span should not include whitespace
Test case:

```typescript
  it('should not include whitespace in the interpolation span', () => {
    const content = mockHost.override(TEST_TEMPLATE, `
      <div>{{ titlex }}</div>
    `);
    const diags = ngLS.getDiagnostics(TEST_TEMPLATE);
    expect(diags.length).toBe(1);
    const {messageText, start, length} = diags[0];
    expect(messageText)
        .toBe(
            `Identifier 'titlex' is not defined. ` +
            `The component declaration, template variable declarations, ` +
            `and element references do not contain such a member`);
    expect(start).toBe(content.indexOf('titlex'));
    // Currently the length is 'titlex ', with the trailing whitespace included.
    expect(length).toBe('titlex'.length);
  });
```

### LINKED ISSUE #33477: Extraneous whitespace in expression ASTs
# 🐞 bug report

### Affected Package

@angular/compiler

### Is this a regression?

No

### Description

The expression parser leaves trailing whitespace when generating some expressions ASTs. This is already a known issue:

https://github.com/angular/angular/blob/8b5ca670ad74ccc6ca1edd0eea06b9520143d557/packages/compiler/test/render3/r3_ast_absolute_span_spec.ts#L60-L68

This extra whitespace should probably be stripped so that e.g. diagnostic messages do not have trailing whitespace. See https://github.com/angular/vscode-ng-language-service/issues/433 for an example of this affecting the language service.

This is unrelated to #31898.

## 🔬 Minimal Reproduction

https://github.com/angular/angular/blob/8b5ca670ad74ccc6ca1edd0eea06b9520143d557/packages/compiler/test/render3/r3_ast_absolute_span_spec.ts#L60-L68

## 🔥 Exception or Error

N/A

## 🌍  Your Environment

**Angular Version:**
```
9.0.0-next.*
```

**Anything else relevant?**

N/A

ayazhafiz:
This commit fixes a bug with the expression parser wherein the end index
of an expression node was recorded as the start index of the next token,
not the end index of the current token.

Closes #33477
Closes angular/vscode-ng-language-service#433

## PR Checklist

## PR Type

- [x] Bugfix

- [x] No

### DIFF
--- a/packages/compiler/src/expression_parser/parser.ts
+++ b/packages/compiler/src/expression_parser/parser.ts
@@ -312,7 +312,14 @@ export class _ParseAST {
    */
   get currentAbsoluteOffset(): number { return this.absoluteOffset + this.inputIndex; }
 
-  span(start: number) { return new ParseSpan(start, this.inputIndex); }
+  span(start: number) {
+    // `end` is either the
+    //   - end index of the current token
+    //   - start of the first token (this can happen e.g. when creating an implicit receiver)
+    const curToken = this.peek(-1);
+    const end = curToken ? curToken.end + this.offset : this.inputIndex;
+    return new ParseSpan(start, end);
+  }
 
   sourceSpan(start: number): AbsoluteSourceSpan {
     const serial = `${start}@${this.inputIndex}`;
@@ -740,7 +747,6 @@ export class _ParseAST {
           const value = this.parseConditional();
           return new PropertyWrite(this.span(start), this.sourceSpan(start), receiver, id, value);
         } else {
-          const span = this.span(start);
           return new PropertyRead(this.span(start), this.sourceSpan(start), receiver, id);
         }
       }
@@ -887,11 +893,7 @@ export class _ParseAST {
       return null;
     }
     const ast = this.parsePipe();  // example: "condition | async"
-    const {start} = ast.span;
-    // Getting the end of the last token removes trailing whitespace.
-    // If ast has the correct end span then no need to peek at last token.
-    // TODO(ayazhafiz): Remove this in https://github.com/angular/angular/pull/34690
-    const {end} = this.peek(-1);
+    const {start, end} = ast.span;
     const value = this.input.substring(start, end);
     return new ASTWithSource(ast, value, this.location, this.absoluteOffset + start, this.errors);
   }
--- a/packages/compiler/src/expression_parser/parser.ts
+++ b/packages/compiler/src/expression_parser/parser.ts
@@ -317,7 +317,7 @@ export class _ParseAST {
     //   - end index of the current token
     //   - start of the first token (this can happen e.g. when creating an implicit receiver)
     const curToken = this.peek(-1);
-    const end = curToken ? curToken.end + this.offset : this.inputIndex;
+    const end = this.index > 0 ? curToken.end + this.offset : this.inputIndex;
     return new ParseSpan(start, end);
   }
 
--- a/packages/language-service/src/completions.ts
+++ b/packages/language-service/src/completions.ts
@@ -6,7 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {AST, ASTWithSource, AstPath, AttrAst, Attribute, BoundDirectivePropertyAst, BoundElementPropertyAst, BoundEventAst, BoundTextAst, Element, ElementAst, HtmlAstPath, NAMED_ENTITIES, Node as HtmlAst, NullTemplateVisitor, ReferenceAst, TagContentType, TemplateBinding, Text, VariableBinding, getHtmlTagDefinition} from '@angular/compiler';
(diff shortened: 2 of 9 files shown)


JoostK (approved):

packages/compiler/src/expression_parser/parser.ts
```diff
@@ -312,7 +312,14 @@ export class _ParseAST {
    */
   get currentAbsoluteOffset(): number { return this.absoluteOffset + this.inputIndex; }
 
-  span(start: number) { return new ParseSpan(start, this.inputIndex); }
+  span(start: number) {
+    // `end` is either the
+    //   - end index of the current token
+    //   - start of the first token (this can happen e.g. when creating an implicit receiver)
+    const curToken = this.peek(-1);
+    const end = curToken ? curToken.end + this.offset : this.inputIndex;
```
kyliau: `curToken` will always be truthy, so instead of doing this check, I think we should handle the case where `this.index === 0`

packages/language-service/src/completions.ts
```diff
@@ -575,21 +575,21 @@ class ExpressionVisitor extends NullTemplateVisitor {
...
-    // If the expression is incomplete, for example *ngFor="let x of |"
-    // binding.expression is null. We could still try to provide suggestions
-    // by looking for symbols that are in scope.
-    const KW_OF = ' of ';
-    const ofLocation = attr.value.indexOf(KW_OF);
-    if (ofLocation > 0 && valueRelativePosition >= ofLocation + KW_OF.length) {
-      const expressionAst = this.info.expressionParser.parseBinding(
-          attr.value, attr.sourceSpan.toString(), attr.sourceSpan.start.offset);
-      this.processExpressionCompletions(expressionAst);
+    } else if (binding instanceof ExpressionBinding) {
```
kyliau: this check shouldn't be necessary since `TemplateBinding = VariableBinding | ExpressionBinding`
but please add a comment that clarifies that this is an ExpressionBinding

packages/language-service/src/completions.ts
```diff
@@ -575,21 +575,21 @@ class ExpressionVisitor extends NullTemplateVisitor {
...
+      if (inSpan(this.position, binding.value?.ast.sourceSpan)) {
+        this.processExpressionCompletions(binding.value !.ast);
+        return;
+      } else if (!binding.value && this.position > binding.key.span.end) {
+        // No expression is defined for the value of the key expression binding, but the cursor is
+        // in a location where the expression would be defined. This can happen in a case like
+        //   let i of |
+        //            ^-- cursor
+        // In this case, backfill the value to be an empty expression and retrieve completions.
```
kyliau: nice work!

kyliau (approved):
Two minor comments, thank you for fixing this! It's awesome =D

packages/compiler/src/expression_parser/parser.ts
ayazhafiz: This is equivalent to that check, since `let arr = []; arr[-1] === undefined`.

packages/language-service/src/completions.ts
ayazhafiz: I meant for this to be a comment in code, so that it is easier to see what the definition of `ExpressionBinding` is just by going to the definition if reading in an editor with tsserver. Let me know if you would prefer a simple `else` with a comment

packages/compiler/src/expression_parser/parser.ts
kyliau: ah, I was mistaken by the return type of peek(). It should really be `Token|undefined`.

packages/language-service/src/completions.ts
kyliau: I think the editor is able to infer that `binding` is this `else` scope is `ExpressionBinding`, so if you right click and select "Go to type definition" it'd bring you to `ExpressionScope`.
It's ok to leave it as is. Thanks!

packages/compiler/src/expression_parser/parser.ts
ayazhafiz: I decided to go with your first suggestion, since updating the type might be confusing since it generally should apply to this specific use case (in all other cases, we will never look beyond the array).

ayazhafiz:
Caretaker: could you run g3sync?

AndrewKushnir:
@ayazhafiz, sure, I will run a presubmit, but it looks like rebase needs to happen first (the `setup` CI job did not start). Could you please rebase and ping this thread again once CI is "green", so that we can run presubmit? Thank you.

ayazhafiz:
@AndrewKushnir sure, done :)

AndrewKushnir:
[Presubmit](http://test/OCL:301503070:BASE:301503076:1584498107838:70e6e403)

kyliau:
This is breaking the golden test for codelyzer in g3, I'll fix it.

AndrewKushnir:
Thanks @kyliau! I've added the `blocked` label for now, please let us know once the problem in g3 is fixed. Thank you.

kyliau:
presubmit: http://test/OCL:301692741:BASE:301692766:1584573968008:b85feb03

PR #34695, opened 2020-01-09
ngcc: handle deep-imports in `DtsDependencyHost`

petebacondarwin:
When searching the typings program for a package for imports a
distinction is drawn between missing entry-points and deep imports.

Previously in the `DtsDependencyHost` these deep imports may be
marked as missing if there was no typings file at the deep import path.
Instead there may be a javascript file instead. In practice this means
the import is "deep" and not "missing".

Now the `DtsDependencyHost` will also consider `.js` files when checking
for deep-imports.

This was causing the `igniteui-angular` package to fail to be processed by
ngcc.

### DIFF
--- a/packages/compiler-cli/ngcc/src/dependencies/dts_dependency_host.ts
+++ b/packages/compiler-cli/ngcc/src/dependencies/dts_dependency_host.ts
@@ -5,7 +5,7 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
-import {FileSystem} from '../../../src/ngtsc/file_system';
+import {AbsoluteFsPath, FileSystem} from '../../../src/ngtsc/file_system';
 import {PathMappings} from '../utils';
 import {EsmDependencyHost} from './esm_dependency_host';
 import {ModuleResolver} from './module_resolver';
@@ -15,6 +15,18 @@ import {ModuleResolver} from './module_resolver';
  */
 export class DtsDependencyHost extends EsmDependencyHost {
   constructor(fs: FileSystem, pathMappings?: PathMappings) {
-    super(fs, new ModuleResolver(fs, pathMappings, ['', '.d.ts', '/index.d.ts']));
+    super(
+        fs, new ModuleResolver(fs, pathMappings, ['', '.d.ts', '/index.d.ts', '.js', '/index.js']));
+  }
+
+  /**
+   * Attempts to process the `importPath` directly and also inside `@types/...`.
+   */
+  protected processImport(
+      importPath: string, file: AbsoluteFsPath, dependencies: Set<AbsoluteFsPath>,
+      missing: Set<string>, deepImports: Set<string>, alreadySeen: Set<AbsoluteFsPath>): boolean {
+    return super.processImport(importPath, file, dependencies, missing, deepImports, alreadySeen) ||
+        super.processImport(
+            `@types/${importPath}`, file, dependencies, missing, deepImports, alreadySeen);
   }
 }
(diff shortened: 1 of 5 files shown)


packages/compiler-cli/ngcc/test/dependencies/dts_dependency_host_spec.ts
```diff
@@ -0,0 +1,231 @@
...
+        // a .js file. This test exercises this particular scenario.
+        const {dependencies, missing, deepImports} = createDependencyInfo();
+        host.collectDependencies(
+            _('/external/deep-import/index.d.ts'), {dependencies, missing, deepImports});
+
+        expect(dependencies.size).toBe(0);
+        expect(missing.size).toBe(0);
+        expect(deepImports.size).toBe(1);
+        expect(deepImports.has(_('/node_modules/lib-1/deep/import'))).toBe(true);
+      });
```
petebacondarwin: This is the key test of the fix here.

packages/compiler-cli/ngcc/src/dependencies/dts_dependency_host.ts
```diff
@@ -15,6 +15,7 @@ import {ModuleResolver} from './module_resolver';
  */
 export class DtsDependencyHost extends EsmDependencyHost {
   constructor(fs: FileSystem, pathMappings?: PathMappings) {
-    super(fs, new ModuleResolver(fs, pathMappings, ['', '.d.ts', '/index.d.ts']));
+    super(
+        fs, new ModuleResolver(fs, pathMappings, ['', '.d.ts', '/index.d.ts', '.js', '/index.js']));
```
petebacondarwin: I played with other approaches such as ignoring all missing packages in dts mode - but I think that this is the most appropriate fix. It is lenient for deep imports but will still pick up "type-only" dependencies that are completely missing, which I think is an important diagnostic for users of ngcc.

gkalpak (approved):

JoostK (approved):

petebacondarwin:
I need to add a bit more behaviour to deal with deep type-only imports that only hit `@types/...` paths. See  https://github.com/angular/angular/issues/34720#issuecomment-573913883

petebacondarwin:
I have updated the PR with the extra `@types/...` behaviour. Please take another look @gkalpak and @JoostK

packages/compiler-cli/ngcc/test/dependencies/dts_dependency_host_spec.ts
```diff
@@ -0,0 +1,253 @@
...
+    let _: typeof absoluteFrom;
+    let host: DtsDependencyHost;
+    beforeEach(() => {
+      _ = absoluteFrom;
+      setupMockFileSystem();
+      const fs = getFileSystem();
+      host = new DtsDependencyHost(fs);
+    });
+
+    describe('getDependencies()', () => {
```
gkalpak:
```suggestion
    describe('collectDependencies()', () => {
```

gkalpak (approved):
LGTM 👍
~~Is the issue with the docs examples tests resolved or is it just flakey (so the tests happened to pass this time, but they might fail on the next run)?~~ _Wrong PR; see below_ 😅

packages/compiler-cli/ngcc/src/dependencies/esm_dependency_host.ts
```diff
@@ -44,29 +44,44 @@ export class EsmDependencyHost extends DependencyHostBase {
...
+   * Resolve the given `importPath` from `file` and add it to the appropriate set.
+   *
+   * @returns `true` if the import was resolved (to an entry-point, a local import, or a
+   * deep-import).
+   */
+  protected processImport(
+      importPath: string, file: AbsoluteFsPath, dependencies: Set<AbsoluteFsPath>,
+      missing: Set<string>, deepImports: Set<string>, alreadySeen: Set<AbsoluteFsPath>): boolean {
+    const resolvedModule = this.moduleResolver.resolveModuleImport(importPath, file);
+    if (resolvedModule) {
```
JoostK: nit: consider inverting this condition to have an early return

JoostK (approved):

petebacondarwin:
> Is the issue with the docs examples tests resolved or is it just flakey (so the tests happened to pass this time, but they might fail on the next run)?

This is a problem with a different PR.

packages/compiler-cli/ngcc/test/dependencies/dts_dependency_host_spec.ts
petebacondarwin: Good spot - and there are others!

PR #34706, opened 2020-01-09
test(ivy): add test for class setters being invoked when not used

crisbeto:
Adds a test that we should make pass once the latest styling refactor has landed.


kara (approved):
LGTM

PR #34707, opened 2020-01-09
ci: fix saucelabs bazel tests terminating after 5min

devversion:
Currently Bazel always terminates the Saucelabs Bazel tests
after 5mins. This is causing failing tests as Saucelabs tests
usually run longer than 5mins. Our current Saucelabs idle timeout
is 10min, so we should ensure that we don't exit early without
properly shutting down Karma / disconnecting the browsers.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -55,6 +55,11 @@ filegroup(
 # flaky. This target runs in CI with Saucelabs and Ivy.
 karma_web_test(
     name = "saucelabs_unit_tests",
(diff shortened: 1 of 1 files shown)


gkalpak (approved):
Is there a way to ensure that Saucelab resources are properly released when/if Bazel decides to terminate the operation (e.g. due to timeout)?

gregmagolan (approved):
LGTM

devversion:
@gkalpak I don't think so. I checked in their code and in the documentation, but didn't find anything.

PR #34709, opened 2020-01-09
docs(ivy): document breaking change with <select> and binding order

### LINKED ISSUE #34470: Ivy: Can't set default value for 'select' when options are generated dynamically through ngFor
### Is this a regression?

Yes, it works in ViewEngine

### Description
While working with Ivy I noticed one incompatibility with ViewEngine.
Let's say we want to dynamically generate options for select and also assign default value

```
<select [value]="3">
  <option [value]="item" *ngFor="let item of [1,2,3]">{{ item }}</option>
</select>
```

ViewEngine will render 3 as selected value(https://ng-run.com/edit/rn9WBjLKXzXUEI8CSE3d) but in Ivy the first item will be selected(https://ng-run.com/edit/TXJC0ZhJTzrN8aYgUmtH?aot=true).

It's happening because the order of checking is different.In ViewEngine updateDirectives is called first(which triggers ngFor directive to render all options) and then updateRenderer is called which sets value to the select.
Ivy checks template from top to bottom so that it first sets default value and since there is no any options rendered it doesn't actually set it. And after that options are rendered.

ViewEngine behavior looks rational for me so I wonder if it's need to be fixed for Ivy. The workaround is using `selected` binding on `option` tag.

Please do note that similar React code(https://codesandbox.io/s/priceless-pare-lynml)

```
<select value={3}>
 {[1, 2, 3].map(item => <option value={item}>{item}</option>)}
</select>
```

works the same way as in ViewEngine

## 🔬 Minimal Reproduction

ViewEngine https://ng-run.com/edit/rn9WBjLKXzXUEI8CSE3d
Ivy 9.0.0-rc.6 https://ng-run.com/edit/TXJC0ZhJTzrN8aYgUmtH?aot=true

kara:
Closes #34470

### DIFF
--- a/aio/content/guide/ivy-compatibility-examples.md
+++ b/aio/content/guide/ivy-compatibility-examples.md
@@ -229,3 +229,58 @@ export class BaseMenu {
 @Directive({selector: '[settingsMenu]'})
 export class SettingsMenu extends BaseMenu {}
 ```
+
+{@a select-value-binding}
+## Cannot Bind to `value` property of `<select>` with `*ngFor`
+
+
+### Basic example of change 
+
+
+```html
+<select [value]="someValue">
+  <option *ngFor="let option of options" [value]="option"> {{ option }} <option>
+</select>
+```
+
+In the previous runtime, the above code would set the initial value of the `<select>` as expected. 
+In Ivy, the initial value would not be set at all in this case.
+
+
+### Background
+
+In the previous runtime, directive input bindings were always executed in their own change detection pass before any DOM bindings were processed. 
+This was an implementation detail that supported the use case in question:
+    
+```html
+<select [value]="someValue">
+  <option *ngFor="let option of options" [value]="option"> {{ option }} <option>
+</select>
+```
+
+It happened to work because the `*ngFor` would be checked first, during the directive input binding pass, and thus create the options first.
+Then the DOM binding pass would run, which would check the `value` binding. 
+At this time, it would be able to match the value against one of the existing options, and set the value of the `<select>` element in the DOM to display that option. 
+
+In Ivy, bindings are checked in the order they are defined in the template, regardless of whether they are directive input bindings or DOM bindings.
+This change makes change detection easier to reason about for debugging purposes, since bindings will run from top to bottom.
+
+In this case, it means that the `value` binding will be checked before the `*ngFor` is checked, as it is above the `*ngFor` in the template. 
+Consequently, the value of the `<select>` element will be set before any options are created, and it won't be able to match and display the correct option in the DOM.
+
+### Example of error
+
+There is no error thrown, but the `<select>` in question will not have the correct initial value displayed in the DOM.
+
+
+### Recommended fix
+
+To fix this problem, we recommend binding to the `selected` property on the `<option>` instead of the `value` on the `<select>`.
+
+```html
+<select>
+  <option *ngFor="let option of options" [value]="option" [selected]="someValue == option">
+    {{ option }}
+  <option>
+</select>
+```
--- a/aio/content/guide/ivy-compatibility-examples.md
+++ b/aio/content/guide/ivy-compatibility-examples.md
@@ -243,13 +243,13 @@ export class SettingsMenu extends BaseMenu {}
 </select>
 ```
 
-In the previous runtime, the above code would set the initial value of the `<select>` as expected. 
+In the View Engine runtime, the above code would set the initial value of the `<select>` as expected. 
 In Ivy, the initial value would not be set at all in this case.
 
 
 ### Background
 
-In the previous runtime, directive input bindings were always executed in their own change detection pass before any DOM bindings were processed. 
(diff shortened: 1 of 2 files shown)


aio/content/guide/ivy-compatibility-examples.md
```diff
@@ -228,4 +228,59 @@ export class BaseMenu {
...
+### Basic example of change 
+
+
+```html
+<select [value]="someValue">
+  <option *ngFor="let option of options" [value]="option"> {{ option }} <option>
+</select>
+```
+
+In the previous runtime, the above code would set the initial value of the `<select>` as expected.
```
IgorMinar: "previous runtime" seem ambiguous for anyone who deep-links into this section. Maybe say "in View Engine runtime" ?

aio/content/guide/ivy-compatibility-examples.md
```diff
@@ -228,4 +228,59 @@ export class BaseMenu {
...
+</select>
+```
+
+In the previous runtime, the above code would set the initial value of the `<select>` as expected. 
+In Ivy, the initial value would not be set at all in this case.
+
+
+### Background
+
+In the previous runtime, directive input bindings were always executed in their own change detection pass before any DOM bindings were processed.
```
IgorMinar:
```suggestion
Prior to Ivy, directive input bindings were always executed in their own change detection pass before any DOM bindings were processed. 
```

aio/content/guide/ivy-compatibility-examples.md
```diff
@@ -228,4 +228,59 @@ export class BaseMenu {
...
+  <option *ngFor="let option of options" [value]="option"> {{ option }} <option>
+</select>
+```
+
+It happened to work because the `*ngFor` would be checked first, during the directive input binding pass, and thus create the options first.
+Then the DOM binding pass would run, which would check the `value` binding. 
+At this time, it would be able to match the value against one of the existing options, and set the value of the `<select>` element in the DOM to display that option. 
+
+In Ivy, bindings are checked in the order they are defined in the template, regardless of whether they are directive input bindings or DOM bindings.
+This change makes change detection easier to reason about for debugging purposes, since bindings will run from top to bottom.
```
IgorMinar:
```suggestion
This change makes change detection easier to reason about for debugging purposes, since bindings will be checked in the depth-first order as declared in the template.
```

aio/content/guide/ivy-compatibility-examples.md
```diff
@@ -228,4 +228,59 @@ export class BaseMenu {
...
+
+In this case, it means that the `value` binding will be checked before the `*ngFor` is checked, as it is above the `*ngFor` in the template.
```
IgorMinar:
```suggestion
For our example `select` snippet above, this change means that the `value` binding will be checked before the `*ngFor` is checked, as it is declared above the `*ngFor` in the template. 
```

aio/content/guide/ivy-compatibility-examples.md
```diff
@@ -228,4 +228,59 @@ export class BaseMenu {
...
+
+### Example of error
+
+There is no error thrown, but the `<select>` in question will not have the correct initial value displayed in the DOM.
+
+
+### Recommended fix
+
+To fix this problem, we recommend binding to the `selected` property on the `<option>` instead of the `value` on the `<select>`.
+
```
IgorMinar: consider repeating the original snippet here so that it's easy to compare before and after:

Original code:

... snippet....

Recommended fix:

.... snippet....

aio/content/guide/ivy-compatibility.md
```diff
@@ -60,4 +60,6 @@ If the errors are gone, switch back to Ivy by removing the changes to the `tscon
 
 * `DebugElement.attributes` returns `undefined` for attributes that were added and then subsequently removed (previously, attributes added and later removed would have a value of `null`).
 
-* `DebugElement.classes` returns `undefined` for classes that were added and then subsequently removed (previously, classes added and later removed would have a value of `false`).
\ No newline at end of file
+* `DebugElement.classes` returns `undefined` for classes that were added and then subsequently removed (previously, classes added and later removed would have a value of `false`).
+
+* If using a native `<select>` with options created through `*ngFor`, the `[selected]` property of an `<option>` must be bound rather than the `[value]` property of the parent `<select>` (previously, you could bind to either.) [details](guide/ivy-compatibility-examples#select-value-binding)
```
IgorMinar:
```suggestion
* If selecting the native `<option>` element in a `<select>` where the `<option>`s are created via `*ngFor`, use the `[selected]` property of an `<option>` instead of binding to the `[value]` property of the `<select>` element (previously, you could bind to either.) [details](guide/ivy-compatibility-examples#select-value-binding)
```

IgorMinar (approved):
a few suggestions, otherwise lgtm

kara:
@IgorMinar Comments addressed in 20e92dec711f8cf23bc348704b7eaed9dd4509b9

kara:
merge-assistance: having trouble getting SL to pass due to flakes but this has no SL implications since it's a docs-only change

PR #34711, opened 2020-01-09
Windows ci tests

gregmagolan:


gregmagolan:
Verified that `//tools/ts-api-guardian/...` tests pass on Windows CI on master.

PR #34715, opened 2020-01-10
refactor(ivy): remove unused embedded view instructions

AndrewKushnir:
This commit performs a cleanup and removes unused embedded view instructions and corresponding tests.

## PR Type

- [ ] Yes
- [x] No


packages/core/src/render3/interfaces/container.ts
```diff
@@ -48,9 +48,7 @@ export const CONTAINER_HEADER_OFFSET = 9;
 
 
 /**
- * Used to track:
- *  - Inline embedded views (see: `ɵɵembeddedViewStart`)
- *  - Transplanted `LView`s (see: `LView[DECLARATION_COMPONENT_VIEW])`
+ * Used to track Transplanted `LView`s (see: `LView[DECLARATION_COMPONENT_VIEW])`
  */
 export const enum ActiveIndexFlag {
```
pkozlowski-opensource: This flag / field should be probably renamed / cleaned up (for example, do we still need `INCREMENT`? Put simply, this field was originally having JS block embedded view index and was later re-purposed to track info about transplanted views. But with JS blocks gone we should make the intention behind the `ACTIVE_INDEX` more clear?

packages/core/test/render3/query_spec.ts
```diff
@@ -1367,66 +1366,6 @@ describe('query', () => {
     });
   });
 
-  describe('queryList', () => {
```
pkozlowski-opensource: It would be good to re-write those as integration / acceptance tests using ngIf / ngFor

packages/core/test/render3/providers_spec.ts
```diff
@@ -1316,99 +1146,6 @@ describe('providers', () => {
       expect(injector.get(Some).location).toEqual('From app component');
     });
   });
-
-  describe('lifecycles', () => {
```
pkozlowski-opensource: probably would make sense to move it to a `TestBed` acceptance test (if we don't have any similar one).

packages/core/test/render3/providers_spec.ts
```diff
@@ -859,176 +859,6 @@ describe('providers', () => {
...
-        });
-      }
-
-      const fixture = new ComponentFixture(ComponentWithProviders);
-      expect(fixture.html)
-          .toEqual(
-              '<div><repeated>foo1,2</repeated><repeated>foo1,2</repeated><repeated>foo1,2</repeated></div>');
-    });
-
-    it('should have access to viewProviders of the repeated component', () => {
```
pkozlowski-opensource: probably would make sense to move it to a `TestBed` acceptance test (if we don't have any similar one).

packages/core/test/render3/providers_spec.ts
```diff
@@ -859,176 +859,6 @@ describe('providers', () => {
     });
   });
 
-  describe('- embedded views', () => {
-    it('should have access to viewProviders of the host component', () => {
```
pkozlowski-opensource: probably would make sense to move it to a `TestBed` acceptance test (if we don't have any similar one).

pkozlowski-opensource:
@AndrewKushnir thnx for starting this cleanup - soooooo many things will be easier when this cleanup lands! 

I did the first pass on this PR and left some small comment. More importantly, though, I think that we should migrate more of the tests from the `test/render3` folder to `TestBed`. I've left some comments / suggestions but IMO we should migrate all tests that:
* are not specifically testing for DOM insertion with JS blocks;
* don't exist as an acceptance test already. 

I know that it will be more effort but would be great not to loose test coverage....

packages/core/src/render3/interfaces/container.ts
AndrewKushnir: I did a quick, search and it looks like `INCREMENT` is the only unused field at this moment, so I removed it. I believe the use of other fields might also be refactored and they can be removed, but I think that might be the next step in refactoring - to revisit existing logic and data structures to clean them up once this PR lands.

AndrewKushnir:
Hi @pkozlowski-opensource, thanks for the review. I've refactored and moved the tests that you mentioned to the `acceptance` folder. Could you please let me know if there are some other tests that would benefit from moving to acceptance folder? Thank you.

packages/core/test/render3/renderer_factory_spec.ts
```diff
@@ -104,102 +104,4 @@ describe('renderer factory lifecycle', () => {
...
-
-    expect(t.html).toEqual('<div><span></span><span></span><span></span></div>');
-
-    condition = false;
-    t.update();
-    expect(t.html).toEqual('<div></div>');
-    expect(ngDevMode).toHaveProperties({rendererDestroy: 0, rendererDestroyNode: 3});
-  });
-
-  it('should call renderer.destroy for each component destroyed', () => {
```
pkozlowski-opensource: @AndrewKushnir I would expect that we've got already an acceptance test like this but if not I would see it as valuable to migrate.

pkozlowski-opensource (approved):
This LGTM overall, just left one comment regarding a test that we might want to migrate. Other tests seem to be either JS-blocks specific or exercising very basic scenarios (which, I would assume, are already covered in other acceptance tests).

packages/core/test/render3/renderer_factory_spec.ts
AndrewKushnir: Thanks @pkozlowski-opensource. I moved these tests to `acceptance/renderer_factory_spec.ts`, so we have the necessary coverage.

AndrewKushnir:
[Presubmit](http://test/OCL:291081011:BASE:291081036:1579749317262:c12a275d)

AndrewKushnir:
[Presubmit](http://test/OCL:291485328:BASE:291485339:1579923712184:67d42e1b)

mhevery (approved):
I saw a lot of Green, but I am assuming that most of it is you just moving things around so I did not look very deeply into it.

packages/core/test/render3/listeners_spec.ts
```diff
@@ -287,277 +287,6 @@ describe('event listeners', () => {
     expect(ctx.showing).toBe(false);
   });
 
-  it('should support listeners in views', () => {
```
kara, line 290: Seems like this test could be moved to acceptance/listener_spec.ts

packages/core/test/render3/listeners_spec.ts
```diff
@@ -287,277 +287,6 @@ describe('event listeners', () => {
...
-    expect(comp.counter).toEqual(2);
-
-    // the listener should be removed when the view is removed
-    comp.showing = false;
-    fixture.update();
-    button.click();
-    expect(comp.counter).toEqual(2);
-  });
-
-  it('should destroy listeners in for loops', () => {
```
kara, line 403: Seems like this test could be moved to acceptance/listener_spec.ts

packages/core/test/render3/listeners_spec.ts
```diff
@@ -718,254 +447,6 @@ describe('event listeners', () => {
     expect(comp.counter).toEqual(6);
   });
 
-  it('should destroy listeners in nested views', () => {
```
kara, line 721: Seems like this test could be moved to acceptance/listener_spec.ts

kara (approved):
LGTM, just need to rebase and add the requested tests

AndrewKushnir:
Thanks for review @kara 👍 I've added requested tests in a fixup commit https://github.com/angular/angular/pull/34715/commits/9d09e376c897abf6f4b6a9f3712b4c65c26460da.

AndrewKushnir:
[Presubmit](https://test.corp.google.com/ui#id=OCL:310475215:BASE:310475287:1588900898928:bd222a5b)

Issue #32431, opened 2019-08-31
Issues with parallel build using ivy


cromefire:
# 🐞 Bug report

### Command (mark with an `x`)

```
- [ ] new
- [x] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Is this a regression?

Kind of, this worked without ivy

### Description

It seems like the cli argues with itself, because both instances want to create the same directory and don't check if it already exists:
```
ERROR in EEXIST: file already exists, mkdir '/path/to/project/node_modules/@angular/core/__ivy_ngcc__'
```
or
```
ERROR in Tried to overwrite /path/to/project/node_modules/@angular/core/core.d.ts.__ivy_ngcc_bak with an ngcc back up file, which is disallowed.
```

## 🔬 Minimal Reproduction

Build app and universal server:
```
ng build --prod & ng run <project>:server:production
```

## 🔥 Exception or Error

```
ERROR in EEXIST: file already exists, mkdir '/path/to/project/node_modules/@angular/core/__ivy_ngcc__'

ERROR in Tried to overwrite /path/to/project/node_modules/@angular/core/core.d.ts.__ivy_ngcc_bak with an ngcc back up file, which is disallowed.
```

**The app/repo must be clean (not already built; to reproduce it again remove `node_modules` and `yarn install` again)**

## 🌍 Your Environment

```
     _                      _                 ____ _     ___
    / \   _ __   __ _ _   _| | __ _ _ __     / ___| |   |_ _|
   / △ \ | '_ \ / _` | | | | |/ _` | '__|   | |   | |    | |
  / ___ \| | | | (_| | |_| | | (_| | |      | |___| |___ | |
 /_/   \_\_| |_|\__, |\__,_|_|\__,_|_|       \____|_____|___|
                |___/
    

Angular CLI: 8.3.2
Node: 12.9.1
OS: linux x64
Angular: 8.2.4
... animations, common, compiler, compiler-cli, core, forms
... platform-browser, platform-browser-dynamic, platform-server
... router, service-worker

Package                                    Version
--------------------------------------------------------------------
@angular-devkit/architect                  0.803.2
@angular-devkit/build-angular              0.803.2
@angular-devkit/build-optimizer            0.803.2
@angular-devkit/core                       8.3.2
[...]

PR #34722, opened 2020-01-10
fix(ngcc): implement lockfile

petebacondarwin:
Previously, it was possible for multiple instance of ngcc to be running
at the same time, but this is not supported and can cause confusing and
flakey errors at build time.

Now, only one instance of ngcc can run at a time. If a second instance
tries to execute it fails with an appropriate error message.

See https://github.com/angular/angular/issues/32431#issuecomment-571825781

### DIFF
--- a/packages/compiler-cli/ngcc/src/execution/lock_file.ts
+++ b/packages/compiler-cli/ngcc/src/execution/lock_file.ts
@@ -1,0 +1,119 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import * as process from 'process';
+import {FileSystem} from '../../../src/ngtsc/file_system';
+
+/**
+ * The LockFile is used to prevent more than one instance of ngcc executing at the same time.
+ *
+ * When ngcc starts executing, it creates a file in the `compiler-cli/ngcc` folder. If it finds one
+ * is already there then it fails with a suitable error message.
+ * When ngcc completes executing, it removes the file so that future ngcc executions can start.
+ */
+export class LockFile {
+  lockFilePath =
+      this.fs.resolve(require.resolve('@angular/compiler-cli/ngcc'), '../__ngcc_lock_file__');
+
+  constructor(private fs: FileSystem) {}
+
+  /**
+   * Run a function guarded by the lock file.
+   *
+   * Note that T can be a Promise. If so, we run the `remove()` call in the promise's `finally`
+   * handler. Otherwise we run the `remove()` call in the `try...finally` block.
+   *
+   * @param fn The function to run.
+   */
+  lock<T>(fn: () => T): T {
+    let isAsync = false;
+    this.create();
+    try {
+      const result = fn();
+      if (result instanceof Promise) {
+        isAsync = true;
+        // The cast is necessary because TS cannot deduce that T is now a promise here.
+        return result.finally(() => this.remove()) as unknown as T;
+      } else {
+        return result;
+      }
+    } finally {
+      if (!isAsync) {
+        this.remove();
+      }
+    }
+  }
+
+  /**
+   * Write a lock file to disk, or error if there is already one there.
+   */
+  protected create() {
+    try {
+      this.addSignalHandlers();
+      // To avoid race conditions, we check for existence of the lockfile
+      // by actually trying to create it exclusively
+      this.fs.writeFile(this.lockFilePath, process.pid.toString(), /* exclusive */ true);
+    } catch (e) {
+      this.removeSignalHandlers();
+      if (e.code !== 'EEXIST') {
+        throw e;
+      }
+
+      // The lockfile already exists so raise a helpful error.
+      // It is feasible that the lockfile was removed between the previous check for existence
+      // and this file-read. If so then we still error but as gracefully as possible.
+      let pid: string;
+      try {
+        pid = this.fs.readFile(this.lockFilePath);
+      } catch {
+        pid = '{unknown}';
+      }
+
+      throw new Error(
+          `ngcc is already running at process with id ${pid}.\n` +
+          `(If you are sure no ngcc process is running then you should delete the lockfile at ${this.lockFilePath}.)`);
+    }
+  }
+
+  /**
+   * Remove the lock file from disk.
+   */
+  protected remove() {
+    this.removeSignalHandlers();
+    if (this.fs.exists(this.lockFilePath)) {
+      this.fs.removeFile(this.lockFilePath);
+    }
+  }
+
+  protected addSignalHandlers() {
+    process.once('SIGINT', this.signalHandler);
+    process.once('SIGHUP', this.signalHandler);
+  }
+
+  protected removeSignalHandlers() {
+    process.removeListener('SIGINT', this.signalHandler);
+    process.removeListener('SIGHUP', this.signalHandler);
+  }
+
+  /**
+   * This handle needs to be defined as a property rather than a method
+   * so that it can be passed around as a bound function.
+   */
+  protected signalHandler =
+      () => {
+        this.remove();
+        this.exit(1);
+      }
+
+  /**
+   * This function wraps `process.exit()` which makes it easier to manage in unit tests,
+   * since it is not possible to mock out `process.exit()` when it is called from signal handlers.
+   */
+  protected exit(code: number): void {
+    process.exit(code);
+  }
+}
--- a/packages/compiler-cli/ngcc/src/main.ts
+++ b/packages/compiler-cli/ngcc/src/main.ts
@@ -149,6 +149,20 @@ export function mainNgcc(
   // NOTE: Avoid eagerly instantiating anything that might not be used when running sync/async or in
   //       master/worker process.
   const fileSystem = getFileSystem();
+
+
+  // Bail out early if the work is already done.
+  const supportedPropertiesToConsider = ensureSupportedProperties(propertiesToConsider);
+  const absoluteTargetEntryPointPath =
+      targetEntryPointPath !== undefined ? resolve(basePath, targetEntryPointPath) : undefined;
+  if (absoluteTargetEntryPointPath !== undefined &&
+      hasProcessedTargetEntryPoint(
+          fileSystem, absoluteTargetEntryPointPath, supportedPropertiesToConsider,
+          compileAllFormats)) {
+    logger.debug('The target entry-point has already been processed');
+    return;
+  }
+
   // NOTE: To avoid file corruption, ensure that each `ngcc` invocation only creates _one_ instance
   //       of `PackageJsonUpdater` that actually writes to disk (across all processes).
   //       This is hard to enforce automatically, when running on multiple processes, so needs to be
@@ -159,8 +173,6 @@ export function mainNgcc(
   const analyzeEntryPoints: AnalyzeEntryPointsFn = () => {
     logger.debug('Analyzing entry-points...');
     const startTime = Date.now();
-
-    const supportedPropertiesToConsider = ensureSupportedProperties(propertiesToConsider);
 
     const moduleResolver = new ModuleResolver(fileSystem, pathMappings);
     const esmDependencyHost = new EsmDependencyHost(fileSystem, moduleResolver);
@@ -180,7 +192,7 @@ export function mainNgcc(
     const config = new NgccConfiguration(fileSystem, dirname(absBasePath));
     const {entryPoints, graph} = getEntryPoints(
         fileSystem, pkgJsonUpdater, logger, dependencyResolver, config, absBasePath,
-        targetEntryPointPath, pathMappings, supportedPropertiesToConsider, compileAllFormats);
+        absoluteTargetEntryPointPath, pathMappings);
 
     const unprocessableEntryPointPaths: string[] = [];
     // The tasks are partially ordered by virtue of the entry-points being partially ordered too.
@@ -348,13 +360,12 @@ function getExecutor(
 function getEntryPoints(
     fs: FileSystem, pkgJsonUpdater: PackageJsonUpdater, logger: Logger,
     resolver: DependencyResolver, config: NgccConfiguration, basePath: AbsoluteFsPath,
-    targetEntryPointPath: string | undefined, pathMappings: PathMappings | undefined,
-    propertiesToConsider: string[], compileAllFormats: boolean):
-    {entryPoints: PartiallyOrderedEntryPoints, graph: DepGraph<EntryPoint>} {
+    targetEntryPointPath: AbsoluteFsPath | undefined, pathMappings: PathMappings |
+        undefined): {entryPoints: PartiallyOrderedEntryPoints, graph: DepGraph<EntryPoint>} {
   const {entryPoints, invalidEntryPoints, graph} = (targetEntryPointPath !== undefined) ?
       getTargetedEntryPoints(
           fs, pkgJsonUpdater, logger, resolver, config, basePath, targetEntryPointPath,
-          propertiesToConsider, compileAllFormats, pathMappings) :
+          pathMappings) :
       getAllEntryPoints(fs, config, logger, resolver, basePath, pathMappings);
   logInvalidEntryPoints(logger, invalidEntryPoints);
   return {entryPoints, graph};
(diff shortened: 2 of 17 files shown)


packages/compiler-cli/ngcc/test/helpers/mock_lock_file.ts
```diff
@@ -0,0 +1,27 @@
...
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import {MockFileSystemNative} from '../../../src/ngtsc/file_system/testing';
+import {LockFile} from '../../src/execution/lock_file';
+
+export class MockLockFile extends LockFile {
+  log: string[] = [];
+  constructor(private options: {throwOnCreate?: boolean, throwOnRemove?: boolean} = {}) {
+    super(new MockFileSystemNative());
```
JoostK: Could this be `getFileSystem` instead, as we run inside `runInEachFileSystem`?

packages/compiler-cli/ngcc/src/execution/lock_file.ts
```diff
@@ -0,0 +1,99 @@
...
+   */
+  protected create() {
+    if (this.fs.exists(this.lockFilePath)) {
+      const pid = this.fs.readFile(this.lockFilePath);
+      throw new Error(
+          `ngcc is already running at process with id ${pid}.\n` +
+          `(If you are sure no ngcc process is running then you should delete the lockfile at ${this.lockFilePath}.)`);
+    }
+    this.fs.ensureDir(this.fs.dirname(this.lockFilePath));
+    this.fs.writeFile(this.lockFilePath, process.pid.toString());
```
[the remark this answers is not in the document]
petebacondarwin: I have changed the algorithm to use an exclusive write (and no exists check) which removes the change of a race condition, so I think we can get away with not adding the PID check here.

packages/compiler-cli/ngcc/test/helpers/mock_lock_file.ts
petebacondarwin: I have added a comment to the code. This `MockLockFile` is not used in tests that run `runInEachFileSystem`.

packages/compiler-cli/ngcc/src/execution/lock_file.ts
```diff
@@ -0,0 +1,99 @@
...
+   */
+  protected remove() {
+    if (this.fs.exists(this.lockFilePath)) {
+      this.fs.removeFile(this.lockFilePath);
+    }
+  }
+
+  protected addSignalHandlers() {
+    process.once('SIGINT', this.signalHandler);
+    process.once('SIGHUP', this.signalHandler);
```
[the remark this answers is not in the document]
petebacondarwin: No idea! I am hoping that our CI (or one of our Windows gurus) will test that for me :-)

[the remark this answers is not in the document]
petebacondarwin: But to be serious, I looked into it and in theory this should be enough...

packages/compiler-cli/ngcc/src/execution/lock_file.ts
```diff
@@ -0,0 +1,99 @@
...
+    this.fs.writeFile(this.lockFilePath, process.pid.toString());
+  }
+
+  /**
+   * Remove the lock file from disk.
+   */
+  protected remove() {
+    if (this.fs.exists(this.lockFilePath)) {
+      this.fs.removeFile(this.lockFilePath);
+    }
```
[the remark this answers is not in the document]
petebacondarwin: Sadly we can't remove the handlers here because it breaks the unit tests. This is because in the unit tests we don't actually kill the process so `remove()` still gets called :-/

packages/compiler-cli/ngcc/src/execution/lock_file.ts
```diff
@@ -0,0 +1,116 @@
...
+      }
+
+      // The lockfile already exists so raise a helpful error.
+      // It is feasible that the lockfile was removed between the previous check for existence
+      // and this file-read. If so then we still error but as gracefully as possible.
+      let pid: string;
+      try {
+        pid = this.fs.readFile(this.lockFilePath);
+      } catch (e) {
+        pid = '{unkown}';
```
JoostK:
```suggestion
        pid = '{unknown}';
```

packages/compiler-cli/ngcc/src/execution/lock_file.ts
```diff
@@ -0,0 +1,116 @@
...
+        throw e;
+      }
+
+      // The lockfile already exists so raise a helpful error.
+      // It is feasible that the lockfile was removed between the previous check for existence
+      // and this file-read. If so then we still error but as gracefully as possible.
+      let pid: string;
+      try {
+        pid = this.fs.readFile(this.lockFilePath);
+      } catch (e) {
```
JoostK:
```suggestion
      } catch {
```

petebacondarwin: Woh! I was sure that one had to provide the `(e)` even if you didn't use it. But I see that this is not the case. YLSED!

JoostK: @petebacondarwin it's been supported since TS2.5! https://mariusschulz.com/blog/optional-catch-binding-in-typescript

packages/compiler-cli/ngcc/src/execution/lock_file.ts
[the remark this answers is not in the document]
petebacondarwin: Hmm, but I just realised that the CLI will not exit the process after each ngcc run (i.e. they use `main.ts` directly. Which means that each time the CLI runs ngcc we will add more handlers...
So perhaps I do need to find a way to remove the handlers here - and still get it to work in the unit tests...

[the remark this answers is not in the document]
petebacondarwin: And I can indeed run the `removeSignalHandlers()` here!

packages/compiler-cli/ngcc/src/main.ts
```diff
@@ -148,6 +149,20 @@ export function mainNgcc(
...
+  // Bail out early if the work is already done.
+  const supportedPropertiesToConsider = ensureSupportedProperties(propertiesToConsider);
+  const absoluteTargetEntryPointPath =
+      targetEntryPointPath !== undefined ? resolve(basePath, targetEntryPointPath) : undefined;
+  if (absoluteTargetEntryPointPath !== undefined &&
+      hasProcessedTargetEntryPoint(
+          fileSystem, absoluteTargetEntryPointPath, supportedPropertiesToConsider,
+          compileAllFormats)) {
+    logger.debug('The target entry-point has already been processed');
+    return;
```
gkalpak: `mainNgcc` is expected to return a promise if `async` is true (returning `undefined` could break someone using it programatically and expecting a promise - not that there's anyone doing this, but better to be safe :grin:):

```suggestion
    return async ? Promise.resolve() : undefined;
```

packages/compiler-cli/ngcc/src/execution/lock_file.ts
```diff
@@ -0,0 +1,117 @@
...
+
+/**
+ * The LockFile is used to prevent more than one instance of ngcc executing at the same time.
+ *
+ * When ngcc starts executing, it creates a file next to this source file. If it finds one is
+ * already there then it fails with a suitable error message.
+ * When ngcc completes executing, it removes the file so that future ngcc executions can start.
+ */
+export class LockFile {
+  lockFilePath = this.fs.resolve(__dirname, '__ngcc_lock_file__');
```
gkalpak: Why not create it in a more "central" place, such as `compiler-cli/` or `compiler-cli/ngcc/`?
That would also be better DX imo (in case one needs to remove it manually after a crash or something).

packages/compiler-cli/ngcc/src/execution/lock_file.ts
```diff
@@ -0,0 +1,117 @@
...
+  lock<T>(fn: () => T): T {
+    let isAsync = false;
+    this.addSignalHandlers();
+    this.create();
+    try {
+      const result = fn();
+      if (result instanceof Promise) {
+        isAsync = true;
+        // The cast is necessary because TS cannot deduce that T is now a promise here.
+        return result.finally(() => this.remove()) as unknown as T;
```
gkalpak: [According to MDN][1], `Promise.finally` is available in Node.js since v10. Should we play it safe and refrain from using it?

[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally#Browser_compatibility

packages/compiler-cli/ngcc/src/execution/lock_file.ts
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  /**
+   * This handle needs to be defined as a property rather than a method
+   * so that it can be passed around as a bound function.
+   */
+  protected signalHandler =
+      () => {
+        this.remove();
+        this.exit(1);
```
gkalpak: Given that `removeSignalHandlers` is now called in `this.remove()`, this could be just `process.exit(1)` (and the `exit()` method could be dropped).

packages/compiler-cli/ngcc/src/execution/lock_file.ts
```diff
@@ -0,0 +1,117 @@
...
+   *
+   * Note that T can be a Promise. If so, we run the `remove()` call in the promise's `finally`
+   * handler. Otherwise we run the `remove()` call in the `try...finally` block.
+   *
+   * @param fn The function to run.
+   */
+  lock<T>(fn: () => T): T {
+    let isAsync = false;
+    this.addSignalHandlers();
+    this.create();
```
gkalpak: So, if `this.create()` errors (and the process does not exit, because the caller handles the error), shouldn't signal handlers still be removed?

packages/compiler-cli/ngcc/src/execution/lock_file.ts
```diff
@@ -0,0 +1,117 @@
...
+        this.remove();
+      }
+    }
+  }
+
+  /**
+   * Write a lock file to disk, or error if there is already one there.
+   */
+  protected create() {
+    this.fs.ensureDir(this.fs.dirname(this.lockFilePath));
```
gkalpak: Is this necessary, given that it's `__dirname`?

gkalpak (changes requested):
Looking good! Some minor comments/questions/suggestions.

packages/compiler-cli/ngcc/src/execution/lock_file.ts
petebacondarwin: I did consider that but I was concerned about hard-coding the path to a more central place into this file. If folders/files get moved around then we would need to update the logic here, which IMO is brittle.

The path to the file is printed in the error message, so it is a simple cut-and-paste to delete it. And IMO there is not a huge difference in complexity for the developer between:

```
/Users/pete/dev/angular/tmp/ngcc-lockfile/integration/ngcc/node_modules/@angular/compiler-cli/ngcc/src/execution/__ngcc_lock_file__
```

and

```
/Users/pete/dev/angular/tmp/ngcc-lockfile/integration/ngcc/node_modules/@angular/compiler-cli/ngcc/__ngcc_lock_file__
```

packages/compiler-cli/ngcc/src/execution/lock_file.ts
petebacondarwin: The `package.json` for Angular indicates node.js >= 10.9; in CLI it is >=10.13. Moreover the CLI will actively error if the node.js version is below 10.13. See https://github.com/angular/angular-cli/blob/43e43bfacb7b59624b726ae1976e4d63d89bebf9/packages/angular/cli/lib/cli/index.ts#L26

So I think we are safe using this.

packages/compiler-cli/ngcc/src/execution/lock_file.ts
petebacondarwin: You are right, we can remove the call to `removeSignalHandlers()` from `exit()` but I want to keep that method because it is not possible to mock out `process.exit()` directly in unit tests: I found that the `process` object/namespace is not the same inside signal handlers - so if you replace `process.exit` in unit tests, the original one is still called inside the handler.

packages/compiler-cli/ngcc/src/execution/lock_file.ts
petebacondarwin: You are right that during normal operation this is not required. Good spot.

I threw this in because in many unit tests we are using a mock file-system that does not include the source files (and therefore doesn't contain this directory). If we don't do this here we have to change "a lot" of unit tests.

But I have just looked again, and I realise that I can add this to the `LockFileUnderTest` class rather than it appearing in the "real" `LockFile` class.

packages/compiler-cli/ngcc/src/execution/lock_file.ts
gkalpak: The `package.json` at the repository root (which I assume is what you mean by "package.json for Angular") only applies during development (i.e. when one works in the repository). It does not affect the published packages (and their consumers). The [package.json][1] of the `compiler-cli` package currently requires Node.js >=8.

Also, while the latest CLI does require >=10.13, isn't it possible (although probably not recommended) that users could use the latest Angular with an older CLI version (or without the CLI at all)?

##
Maybe we just need to update `compiler-cli/package.json` to require Node.js v10 or newer. I think that would be best for v9.

[1]: https://github.com/angular/angular/blob/f2df1c79d/packages/compiler-cli/package.json#L31

petebacondarwin: I agree that we should update the compiler-cli package.json. I'll add a commit to this PR.

petebacondarwin: The last time it was changed was to align it with the CLI: https://github.com/angular/angular/commit/c602563589a7e14f0c57ee5b191979039aa370db

packages/compiler-cli/ngcc/src/execution/lock_file.ts
petebacondarwin: Actually a whole load of the integration tests also need this folder to exist :-/

petebacondarwin: I tweaked the integration tests to also ensure this folder exists...

packages/compiler-cli/ngcc/src/execution/lock_file.ts
JoostK: @petebacondarwin Establishing the path of the ngcc root would be quite stable with `require.resolve('@angular/compiler-cli/ngcc')`?

petebacondarwin: Done

packages/compiler-cli/ngcc/test/execution/lock_file_spec.ts
```diff
@@ -0,0 +1,207 @@
...
+           expect(lockFile.log).toEqual(['create()', 'fn()', 'remove()']);
+         });
+
+      it('should remove the lockfile if CTRL-C is triggered', () => {
+        const fs = getFileSystem();
+        const lockFile = new LockFileUnderTest(fs, /* handleSignals */ true);
+
+        lockFile.lock(() => {
+          lockFile.log.push('SIGINT');
+          process.emit('SIGINT', 'SIGINT');
```
gkalpak: OOC, why the second arg?

gkalpak (approved):
👌

packages/compiler-cli/ngcc/test/execution/lock_file_spec.ts
petebacondarwin: I don't know if this is needed for the runtime but the typings for this overload (i.e. on that uses a signal string in the first parameter) looks like:

```ts
        emit(event: Signals, signal: Signals): boolean;
```

So at the very least, the second parameter is needed for TS compilation.

JoostK (approved):

matsko:
presubmit = https://test.corp.google.com/ui#id=OCL:290769792:BASE:290769840:1579630888550:8a26eed7

petebacondarwin:
@matsko - what is the reason for the presubmit failure?

AndrewKushnir:
@petebacondarwin it looks like there were some pre-existing failing tests. I marked "google3" check as passing for this PR. Thank you.

matsko:
@petebacondarwin please rebase.

petebacondarwin:
@matsko - done.

PR #34724, opened 2020-01-10
I18n default currency from locale

petebacondarwin:
This PR is for v10 only. It adds a breaking change that is not wanted in v9.

### DIFF
--- a/goldens/public-api/core/core.d.ts
+++ b/goldens/public-api/core/core.d.ts
@@ -1064,7 +1064,7 @@ export declare function ɵɵstyleSanitizer(sanitizer: StyleSanitizeFn | null): v
 
 export declare function ɵɵtemplate(index: number, templateFn: ComponentTemplate<any> | null, decls: number, vars: number, tagName?: string | null, attrsIndex?: number | null, localRefsIndex?: number | null, localRefExtractor?: LocalRefExtractor): void;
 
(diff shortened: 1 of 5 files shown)


AndrewKushnir:
@petebacondarwin quick comment: it looks like unrelated fixup commit was added to this PR.

I've also started a [Blueprint Presubmit](http://test/OCL:306335418:BASE:306335455:1586821289177:d3c38b51) for this PR as you requested (by adding the "presubmit" label).

Thank you.

petebacondarwin:
Thanks @AndrewKushnir - The fixup commit is for this PR but for some reason it picked up the wrong commit message. I'll change that.

petebacondarwin:
I need to write a migration to go with this change that will add in a provider for the default to ensure that current projects do not change unexpectedly.

petebacondarwin:
After further discussion, it is decided that this is not actually a desirable change. 
The currency being displayed should not depend upon the current locale - at least not by default.

If the application wishes to display prices in locale specific currencies then it would not only need to change the currency symbol being displayed but also the value of the currency based on the current exchange rate.

So automatically setting the default currency to the current locale is not enough on its own, without further logic to update the value.

It is already possible to set the default currency code by providing the `DEFAULT_CURRENCY_CODE`, and if wanted it can even be set as the current locale's currency via a factory provider:

```
{provide: DEFAULT_CURRENCY_CODE, useFactory: getDefaultCurrencyCode, deps: [LOCALE_ID]},
```

Therefore we do not intend to land this PR. Closing.

PR #34726, opened 2020-01-10
refactor(docs-infra): simplify Firebase redirect config

gkalpak:
This simplifies the Firebase redirect config, removing a redundant pattern (`/api/http`) which is covered by the previous pattern and dropping an unused named captured segment (`:rest*`).

This change does not affect the redirection behavior for the deployed apps, but allows the Firebase hosting emulator to work correctly.
(See [here][1] for more info.)

NOTE:
Although we are not currently using the Firebase emulator for development/testing, we might want to use it in the future.

[1]: https://github.com/firebase/superstatic/issues/231#issuecomment-573034515


IgorMinar (approved):
lgtm, thanks!

PR #34743, opened 2020-01-11
refactor(language-service): create attr for missing attr, bound tmpl

ayazhafiz:
Currently the language service constructs an `AttrAst` anytime it is
missing from a `TemplateAst` path. However, this should only be done
when the path does not contain an "attribute-like" AST, which can
includes bound properties or bound events.

This commit also refactors `visitAttr` to parse bindings only for
microsyntax expressions and does some other minor cleanup to make
linters happy.

This is some cleanup to help the language service eventually use
`BoundDirectivePropertyAst`s for providing completions for template
bindings rather than performing the manual parsing currently done.

## PR Checklist

## PR Type

- [ ] Yes
- [x] No

### DIFF
--- a/packages/language-service/src/completions.ts
+++ b/packages/language-service/src/completions.ts
@@ -6,7 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {AST, AstPath, AttrAst, Attribute, BoundDirectivePropertyAst, BoundElementPropertyAst, BoundEventAst, BoundTextAst, Element, ElementAst, ImplicitReceiver, NAMED_ENTITIES, Node as HtmlAst, NullTemplateVisitor, ParseSpan, PropertyRead, ReferenceAst, TagContentType, TemplateBinding, Text, getHtmlTagDefinition} from '@angular/compiler';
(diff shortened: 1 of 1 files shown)


packages/language-service/src/completions.ts
```diff
@@ -441,36 +442,34 @@ class ExpressionVisitor extends NullTemplateVisitor {
...
-    const valueRelativePosition = this.position - ast.sourceSpan.start.offset;
-    // Find the template binding that contains the position
-    const binding = templateBindings.find(b => inSpan(valueRelativePosition, b.span));
-
-    if (!binding) {
-      return;
-    }
-
     if (ast.name.startsWith('*')) {
+      // This a template binding given by micro syntax expression.
```
kyliau: Eventually this should be migrated to the proper visitor. I'm actually working on this, will submit a follow up PR soon.

kyliau (approved):

packages/language-service/src/completions.ts
ayazhafiz: Awesome! I was planning to do that after this PR, glad to hear you are on it :)

PR #34745, opened 2020-01-12
refactor(common): delete obsolete NgClassBase and NgStyleBase

IgorMinar:
These classes are no longer needed.

These classes were accidently exported as public api in the past, but fortunately since they are
useless, we don't expect anyone to depend on them, so I'm simply deleting them.


IgorMinar:
this PR is obsolete and will be resolved by @mhevery's PRs on the `styling` branch.

PR #34747, opened 2020-01-12
fix(language-service): include directives in template AST paths

ayazhafiz:
When constructing a path to a template node, the language service
currently expunges all directives ASTs. This commit includes them again,
which simplifies location of symbols where directives are, in fact, the
most specific symbol (like element selectors). This commit will also
help in the implementation of #34564, which relies on determination of
the nearest AST in a path.

One inconsistency this introduces is the location of symbols referencing
components between opening and closing tags in an element. In the past,
querying the definition for both opening and closing tags would result
in a query for the symbol encompassing the entire element.

```text
      v                 cursor
<app-c|omp></app-comp>
~~~~~~~~~~~~~~~~~~~~~~  symbol location
                 v      cursor
<app-comp></app-c|omp>
~~~~~~~~~~~~~~~~~~~~~~  symbol location
```

Now, this behavior only occurs for closing tags; for opening tags, the
queried symbol is the opening tag precisely because that's where the
component selector applies.

```text
      v                 cursor
<app-c|omp></app-comp>
~~~~~~~~~~~             symbol location
                 v      cursor
<app-comp></app-c|omp>
~~~~~~~~~~~~~~~~~~~~~~  symbol location
```

## PR Checklist

## PR Type

- [x] Bugfix

- [ ] Yes
- [x] No

### DIFF
--- a/packages/language-service/src/utils.ts
+++ b/packages/language-service/src/utils.ts
@@ -139,10 +139,6 @@ export function findTemplateAstAt(ast: TemplateAst[], position: number): Templat
     visitDirective(ast: DirectiveAst, context: any): any {
       // Ignore the host properties of a directive
       const result = this.visitChildren(context, visit => { visit(ast.inputs); });
-      // We never care about the diretive itself, just its inputs.
-      if (path[path.length - 1] === ast) {
-        path.pop();
-      }
       return result;
     }
   };
--- a/packages/language-service/test/definitions_spec.ts
+++ b/packages/language-service/test/definitions_spec.ts
@@ -136,7 +136,7 @@ describe('definitions', () => {
       // Get the marker for «test-comp» in the code added above.
       const marker = mockHost.getReferenceMarkerFor(fileName, 'test-comp');
 
-      const result = ngService.getDefinitionAt(fileName, marker.start);
+      const result = ngService.getDefinitionAndBoundSpan(fileName, marker.start);
       expect(result).toBeDefined();
       const {textSpan, definitions} = result !;
 
@@ -173,7 +173,7 @@ describe('definitions', () => {
       // Get the marker for «test-comp» in the code added above.
       const marker = mockHost.getReferenceMarkerFor(fileName, 'test-comp');
 
-      const result = ngService.getDefinitionAt(fileName, marker.start);
+      const result = ngService.getDefinitionAndBoundSpan(fileName, marker.start);
       expect(result).toBeDefined();
       const {textSpan, definitions} = result !;
 
--- a/packages/language-service/test/definitions_spec.ts
+++ b/packages/language-service/test/definitions_spec.ts
@@ -125,40 +125,79 @@ describe('definitions', () => {
     expect(def.textSpan).toEqual(mockHost.getDefinitionMarkerFor(fileName, 'include'));
   });
 
-  it('should be able to find a reference to a component', () => {
-    const fileName = mockHost.addCode(`
+  describe('find a reference to a component', () => {
+    it('should be able to find the reference from an opening tag', () => {
+      const fileName = mockHost.addCode(`
       @Component({
-        template: '~{start-my}<«test-comp»></test-comp>~{end-my}'
+        template: '~{start-my}<«test-comp»>~{end-my}</test-comp>'
       })
       export class MyComponent { }`);
 
-    // Get the marker for «test-comp» in the code added above.
-    const marker = mockHost.getReferenceMarkerFor(fileName, 'test-comp');
+      // Get the marker for «test-comp» in the code added above.
+      const marker = mockHost.getReferenceMarkerFor(fileName, 'test-comp');
 
-    const result = ngService.getDefinitionAndBoundSpan(fileName, marker.start);
-    expect(result).toBeDefined();
-    const {textSpan, definitions} = result !;
+      const result = ngService.getDefinitionAt(fileName, marker.start);
+      expect(result).toBeDefined();
+      const {textSpan, definitions} = result !;
 
-    // Get the marker for bounded text in the code added above.
-    const boundedText = mockHost.getLocationMarkerFor(fileName, 'my');
-    expect(textSpan).toEqual(boundedText);
+      // Get the marker for bounded text in the code added above.
+      const boundedText = mockHost.getLocationMarkerFor(fileName, 'my');
+      expect(textSpan).toEqual(boundedText);
 
-    // There should be exactly 1 definition
-    expect(definitions).toBeDefined();
-    expect(definitions !.length).toBe(1);
-    const def = definitions ![0];
+      // There should be exactly 1 definition
+      expect(definitions).toBeDefined();
+      expect(definitions !.length).toBe(1);
+      const def = definitions ![0];
 
-    const refFileName = '/app/parsing-cases.ts';
-    expect(def.fileName).toBe(refFileName);
-    expect(def.name).toBe('TestComponent');
-    expect(def.kind).toBe('component');
-    const content = mockHost.readFile(refFileName) !;
-    const begin = '/*BeginTestComponent*/ ';
-    const start = content.indexOf(begin) + begin.length;
-    const end = content.indexOf(' /*EndTestComponent*/');
-    expect(def.textSpan).toEqual({
-      start,
-      length: end - start,
+      const refFileName = '/app/parsing-cases.ts';
+      expect(def.fileName).toBe(refFileName);
+      expect(def.name).toBe('TestComponent');
+      expect(def.kind).toBe('component');
+      const content = mockHost.readFile(refFileName) !;
+      const begin = '/*BeginTestComponent*/ ';
+      const start = content.indexOf(begin) + begin.length;
+      const end = content.indexOf(' /*EndTestComponent*/');
+      expect(def.textSpan).toEqual({
+        start,
+        length: end - start,
+      });
+    });
+
+    it('should be able to find the reference from a closing tag', () => {
+      const fileName = mockHost.addCode(`
+      @Component({
+        template: '~{start-my}<test-comp></«test-comp»>~{end-my}'
+      })
+      export class MyComponent { }`);
+
+      // Get the marker for «test-comp» in the code added above.
+      const marker = mockHost.getReferenceMarkerFor(fileName, 'test-comp');
+
+      const result = ngService.getDefinitionAt(fileName, marker.start);
+      expect(result).toBeDefined();
+      const {textSpan, definitions} = result !;
+
+      // Get the marker for bounded text in the code added above.
+      const boundedText = mockHost.getLocationMarkerFor(fileName, 'my');
+      expect(textSpan).toEqual(boundedText);
+
+      // There should be exactly 1 definition
+      expect(definitions).toBeDefined();
+      expect(definitions !.length).toBe(1);
+      const def = definitions ![0];
+
+      const refFileName = '/app/parsing-cases.ts';
+      expect(def.fileName).toBe(refFileName);
+      expect(def.name).toBe('TestComponent');
+      expect(def.kind).toBe('component');
+      const content = mockHost.readFile(refFileName) !;
+      const begin = '/*BeginTestComponent*/ ';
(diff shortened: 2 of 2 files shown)


ayazhafiz:
cc @ivanwonder. This PR touches your impl of reference completions. As there is enough information in `referenceAttributeValueCompletions` to construct the completions, I think we can get rid of the visitor. What do you think?

packages/language-service/src/completions.ts
```diff
@@ -336,7 +336,14 @@ function referenceAttributeValueCompletions(
   const dinfo = diagnosticInfoFromTemplateInfo(info);
   const visitor =
       new ExpressionVisitor(info, position, () => getExpressionScope(dinfo, path, false));
-  path.tail.visit(visitor, path.parentOf(path.tail));
+
+  // Find the reference's element, which contains all directives assignable to the reference.
+  let element: TemplateAst|undefined = path.tail;
+  while (!(element instanceof ElementAst) && path.parentOf(element)) {
+    element = path.parentOf(element);
```
ivanwonder: https://github.com/angular/angular/pull/34564#discussion_r363012255
the `path` is `path: (3) [ElementAst, BoundEventAst, BoundDirectivePropertyAst]`. You can use `path.first(ElementAst)`.
But for the template `<div *ngIf="true"></div>`, the `path` is `path: (3) [EmbeddedTemplateAst, DirectiveAst, BoundDirectivePropertyAst]`.

ayazhafiz: True, I didn’t think about that. I’ll check it out soon, on mobile now.

ivanwonder:
I think the root cause of this problem is structural directives(e.g. `ngIf`, `ngFor`). The better way to handle this as described below.
https://github.com/angular/angular/blob/6d534f10e69e0c7763bc44c108017eba1bd230b5/packages/language-service/src/utils.ts#L102
`function findTemplateAstAt(ast: TemplateAst[], position: number, isStructuralDirectives?: boolean)`
`function isStructuralDirectives(templateString:string, position: number):boolean`
check if the `templateString[position - n]` is 'ref-' or '*'.
https://github.com/angular/angular/blob/6d534f10e69e0c7763bc44c108017eba1bd230b5/packages/language-service/src/utils.ts#L118-L125
If the position is on structural directives, no need to visit the directives again.
@ayazhafiz What do you think of this solution.

ayazhafiz:
For clarity, what is the problem you mention here @ivanwonder? I think we may be thinking of different problems.

ivanwonder:
Maybe it's different problems. I talk about my pr #34564 . The structural directives and two-way binding are shorthand form. 
For example `<test-comp string-model «[(ᐱmodelᐱ)]="title"»></test-comp>`, the `model` and `modelChange` share the same span. So 
https://github.com/angular/angular/blob/6d534f10e69e0c7763bc44c108017eba1bd230b5/packages/language-service/src/utils.ts#L127-L137
If ` visit(ast.outputs); ` find a path(`modelChange`), I think it's safe to return here.
For structural directives `<div «*ᐱngIfᐱ="true"»></div>`, the directive `ngIf` and element `div` share the same span. If  the `visit(ast.directives);` can location the `ngIf` directive, no need to visit children.
https://github.com/angular/angular/blob/6d534f10e69e0c7763bc44c108017eba1bd230b5/packages/language-service/src/utils.ts#L118-L125

ivanwonder:
a
 /   \\
b    c
If a to b can be confirmed that is the result. we don't need to visit a to c.

patilharshal16 (approved):

ayazhafiz:
@ivanwonder I see what you are saying. I don't think it will be necessary to limit what we visit (and I'm worried doing so might run into issues down the road). I will try to take a look today or tomorrow; need to catch up on some stuff in my personal life first.

packages/language-service/src/utils.ts
```diff
@@ -139,10 +139,6 @@ export function findTemplateAstAt(ast: TemplateAst[], position: number): Templat
     visitDirective(ast: DirectiveAst, context: any): any {
       // Ignore the host properties of a directive
       const result = this.visitChildren(context, visit => { visit(ast.inputs); });
-      // We never care about the diretive itself, just its inputs.
-      if (path[path.length - 1] === ast) {
-        path.pop();
-      }
       return result;
     }
   };
```
ivanwonder: I don't think the code here need to be removed. An `ElementAst` may contain lots of directives. this will cause lots of meaningless directives in `path`.
I think we should care about the directive itself only when its input is what we need.
```
visitDirective(ast: DirectiveAst, context: any): any {
      const length = path.length;
      // Ignore the host properties of a directive
      const result = this.visitChildren(context, visit => { visit(ast.inputs); });
      // We never care about the diretive itself, just its inputs.
      if (path[path.length - 1] === ast) {
        path.pop();
      }
      if (path.length > length && path[path.length - 2] !== ast) {
        path.splice(path.length - 1, 0, ast);
      }
      return result;
    }
```

ayazhafiz:
I disagree. Knowing the directives on an element is useful because it provides a path that includes all the applied directives. This makes application of the most “specific” directive easier. For example, see the commit message - we can now show that a component is applied because of the opening tag of an element.

The other reason is directive outputs matter too; for example, in the case of a definition for a two-way binding that your PR added support for. Requesting a definition for “[(prop)]” should give a definition for the prop member, but it should also give a definition for the “propChange” event.


________________________________
От: ivan <notifications@github.com>
Отправлено: вторник, января 14, 2020 6:41 AM
Кому: angular/angular
Копия: hafiz; Mention
Тема: Re: [angular/angular] fix(language-service): include directives in template AST paths (#34747)


@ivanwonder commented on this pull request.

________________________________

In packages/language-service/src/utils.ts<https://github.com/angular/angular/pull/34747#discussion_r366376014>:

>      visitDirective(ast: DirectiveAst, context: any): any {
       // Ignore the host properties of a directive
       const result = this.visitChildren(context, visit => { visit(ast.inputs); });
-      // We never care about the diretive itself, just its inputs.
-      if (path[path.length - 1] === ast) {
-        path.pop();
-      }
       return result;
     }
   };


I don't think the code here need to be removed. An ElementAst may contain lots of directives. this will cause lots of meaningless directives in path.
I think we should care about the directive itself only when its input is what we need.

visitDirective(ast: DirectiveAst, context: any): any {
      const length = path.length;
      // Ignore the host properties of a directive
      const result = this.visitChildren(context, visit => { visit(ast.inputs); });
      // We never care about the diretive itself, just its inputs.
      if (path[path.length - 1] === ast) {
        path.pop();
      }
      if (path.length > length && path[path.length - 2] !== ast) {
        path.splice(path.length - 1, 0, ast);
      }
      return result;
    }

packages/language-service/src/completions.ts
ayazhafiz: Resolving this for now as you posted a longer comment of discussion.

ivanwonder:
Ok, I will wait for your future PR.
Just give some info I consider, I hope it's useful for you.
`<test-comp string-model></test-comp>`, the `path` will output `path: (3) [ElementAst, DirectiveAst(string-model), DirectiveAst(test-comp)]`, the last one is what we need, but it depends on 
the parser' order(Maybe this is not a problem).
`<test-comp *string-model="let element;"></test-comp>`(Oh, find a new bug... check the hover for `string-model`), I think it's the same with the pre one.

ayazhafiz:
You are right. The order is non-deterministic. Maybe we should compare the element name to find the matching component (this is what we used to do and I think what we do to find matching directives on an attribute).

Is the bug you mention that string-model hover does not reference the directive? If so, is string-model a structural directive? If it’s not, I wouldn’t expect any hover definition for it to show up. I will check when I am on desktop.

________________________________
От: ivan <notifications@github.com>
Отправлено: вторник, января 14, 2020 6:17 PM
Кому: angular/angular
Копия: hafiz; Mention
Тема: Re: [angular/angular] fix(language-service): include directives in template AST paths (#34747)


Ok, I will wait for your future PR.
Just give some info I consider, I hope it's useful for you.
<test-comp string-model></test-comp>, the path will outputpath: (3) [ElementAst, DirectiveAst(string-model), DirectiveAst(test-comp)], the last one is what we need, but it depends on
the parser' order(Maybe this is not a problem).
<test-comp *string-model="let element;"></test-comp>(Oh, find a new bug... check the hover forstring-model), I think it's the same with the pre one.

ivanwonder:
@ayazhafiz sorry for the delay, I test on the master branch and only remove the code
```
 if (path[path.length - 1] === ast) {
        path.pop();
      }
``` 
the test case is `<test-comp *«string-model»="let ele;"></test-comp>`.
the `path` is `path: (3) [EmbeddedTemplateAst, DirectiveAst, DirectiveAst]`.
the output is
`Expected '(component) AppModule.TestComponent: class' to be '(directive) StringModel: typeof StringModel'.`
I think the `*string-model` is a structural directive. Just like the `ngFor, ngIf`

ayazhafiz:
Structural directives require a TemplateRef and view container (https://angular.io/guide/structural-directives#write-a-structural-directive), which StringModel doesn’t have. I don’t think this is a bug; the diagnostics in this case say that *stringModel is unknown on the element.


________________________________
От: ivan <notifications@github.com>
Отправлено: среда, января 15, 2020 5:24 PM
Кому: angular/angular
Копия: hafiz; Mention
Тема: Re: [angular/angular] fix(language-service): include directives in template AST paths (#34747)


@ayazhafiz<https://github.com/ayazhafiz> sorry for the delay, I test on the master branch and only remove the code

 if (path[path.length - 1] === ast) {

        path.pop();

      }



the test case is <test-comp *«string-model»="let ele;"></test-comp>.
the path is path: (3) [EmbeddedTemplateAst, DirectiveAst, DirectiveAst].
the output is
Expected '(component) AppModule.TestComponent: class' to be '(directive) StringModel: typeof StringModel'.
I think the *string-model is a structural directive. Just like the ngFor, ngIf

ivanwonder:
I think the parser doesn't care about the templateRef and view container which only is used in js runtime or used to generate runtime code(I think). it just parses the asterisk (*) syntax into a `<ng-template>`.
Or you can create a new structural directive for the test.
The directive can be attribute directives or structural directive, depend on how you use it.
See the code here.
https://stackblitz.com/edit/angular-kge9x7?file=src%2Fapp%2Funless.directive.ts

ayazhafiz:
The parser does care about view containers (see https://github.com/angular/angular/blob/436e7388c3487ec98f6eaa6d6a87ced6e0bb2af7/packages/compiler/src/template_parser/template_parser.ts#L385-L411). But you're right, this is not the root cause of the discrepancy you are seeing. The real cause is that the application of a structural directive is an input. For example, the `NgIf` directive has the `[ngIf]` selector and an input of name `ngIf`. Application of `*ngIf="some-expr"` on an element wraps the element in a template and passes `some-expr` to the directive input.

The reason this doesn't work on `StringModel` is because `StringModel`'s input name is `model`. Change the selector and input name to be the same, like `stringModel`, and the definitions show up fine.

Note  that the issue you were noticing does not exist for "real" structural directives. Querying the definition on `*ngIf` or `*appUnless` (per your stackblitz link) gives the correct definition.

ivanwonder:
> The reason this doesn't work on `StringModel` is because `StringModel`'s input name is `model`. Change the selector and input name to be the same, like `stringModel`, and the definitions show up fine.

`<test-comp *string-model="let element;"></test-comp>` If the `some-expr` is `let element;`, you donn't need rename `model`.

packages/language-service/test/definitions_spec.ts
```diff
@@ -125,40 +125,79 @@ describe('definitions', () => {
     expect(def.textSpan).toEqual(mockHost.getDefinitionMarkerFor(fileName, 'include'));
   });
 
-  it('should be able to find a reference to a component', () => {
-    const fileName = mockHost.addCode(`
+  describe('find a reference to a component', () => {
+    it('should be able to find the reference from an opening tag', () => {
+      const fileName = mockHost.addCode(`
```
kyliau: While you're here, could you please change moved the html to TEST_TEMPLATE instead? Thank you!

PR #34753, opened 2020-01-13
ci: do not run benchmark measurements in circleci

devversion:
Currently we run all benchmark perf tests in CircleCI. Since we do not
collect any results, we unnecessarily waste CI/RBE resources. Instead,
we should just not run benchmark perf tests in CI, but still run the
functionality e2e tests which ensure that benchmarks are not broken.

We can do this by splitting the perf and e2e tests into separate
files/targets.


modules/e2e_util/e2e_util.ts
```diff
@@ -12,7 +12,7 @@ import {browser} from 'protractor';
 const yargs = require('yargs');
 import * as webdriver from 'selenium-webdriver';
 
-let cmdArgs: {'bundles': boolean};
+let cmdArgs: {'bundles': boolean}|undefined;
```
IgorMinar: I believe that his is obsolete. In the past you used to be able to run the benchmarks with js code bundled or not. with bazel you can only run them as bundled (because the build is fast enough compared to the pre-bazel build).

I suggest, you remove this option.

modules/benchmarks/src/largeform/BUILD.bazel
```diff
@@ -3,10 +3,21 @@ package(default_visibility = ["//modules/benchmarks:__subpackages__"])
...
+    deps = [
+        "//modules/e2e_util",
+        "@npm//protractor",
+    ],
+)
+
+ts_library(
+    name = "e2e_tests_lib",
+    testonly = 1,
+    srcs = ["largeform.spec.ts"],
```
IgorMinar: can we disambiguate the ".spec.ts" files, by calling them ".e2e-spec.ts" and ".perf-spec.ts"?

modules/benchmarks/e2e_test.bzl
```diff
@@ -0,0 +1,17 @@
...
+  with `@angular/benchpress`.
+"""
+
+def e2e_test(name, server, deps, **kwargs):
+    protractor_web_test_suite(
+        name = name,
+        on_prepare = "//modules/benchmarks:start-server.js",
+        server = server,
+        # `yargs` is needed as runtime dependency for the e2e utils.
+        deps = ["@npm//yargs"] + deps,
```
IgorMinar: do we still need it after you remove the command line option (see comment below)?

IgorMinar (changes requested):
thanks @devversion!  I found a few things that I think we could improve, please take a look at the individual comments.

modules/e2e_util/e2e_util.ts
devversion: Good idea. Will do! I was unsure if this is still needed somewhere.

modules/benchmarks/src/largeform/BUILD.bazel
devversion: Sounds good to me.

modules/benchmarks/e2e_test.bzl
devversion: Yeah, this can be removed then. Nice idea 👍

devversion:
@IgorMinar Thanks for reviewing! I addressed your feedback.

IgorMinar (approved):

IgorMinar:
thank you!

IgorMinar:
Thanks Paul!

IgorMinar:
this PR fails in g3 in a catastrophic way with build errors due to internal latency lab tests. we'll need to investigate this and fix it next week.

devversion:
I don't have enough insight, but I'd guess it's because we either changed the file names of the spec files, or since we updated the test library target names (depending on how it is set up in g3)

josephperrott:
@devversion The change in file names caused the files to be included in a `glob` that they shouldn't have been in.  I created the internal change to prevent them from being included and then we should be able to re-presubmit this and get it merged.

AndrewKushnir:
I started a [new g3 presubmit](http://test/OCL:292038789:BASE:292038823:1580257804362:f6241609).

@devversion could you plz rebase your PR on top of the latest master, so we rerun CI with the most up-to-date version?

Thank you.

devversion:
@AndrewKushnir Rebased.

PR #34759, opened 2020-01-13
perf: optimize findMatchnigIndex

Connormiha:
Optimize finding matched index for selector in extractProjectableNodes

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

I just optimized function for finding matched selectors.

Issue Number: N/A

## What is the new behavior?
N/A

- [ ] Yes
- [x] No

### DIFF
--- a/packages/elements/src/extract-projectable-nodes.ts
+++ b/packages/elements/src/extract-projectable-nodes.ts
@@ -38,17 +38,13 @@ export function extractProjectableNodes(host: HTMLElement, ngContentSelectors: s
 }
 
 function findMatchingIndex(node: Node, selectors: string[], defaultIndex: number): number {
-  let matchingIndex = defaultIndex;
-
   if (isElement(node)) {
-    selectors.some((selector, i) => {
-      if ((selector !== '*') && matchesSelector(node, selector)) {
-        matchingIndex = i;
-        return true;
+    for (let i = 0; i < selectors.length; i++) {
+      if (selectors[i] !== '*' && matchesSelector(node, selectors[i])) {
+        return i;
       }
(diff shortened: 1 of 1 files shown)


gkalpak:
Thx for submitting a PR, @Connormiha, trying to improve the performance of `@angular/elements`. Can you share some more info on why this improves performance and the motivation behind this change (for example, do you have found this prticular piece of code to be causing a slow-down).

TBH, I find it hard to believe that this change can in any way noticeably improve the performance of an app, because:
- The main difference is using a `for`-loop vs using `Array.prototype.some()` (and the associated function calls), which VMs are quite good at handling these days.
- The `selectors` array will be empty in most cases, leading to `selectors.some()` being a no-op.
- In the few cases where `selectors` is not empty, it will usually contain just a handful of items.
- In the very rare cases where `selectors` contains many items, there will be other operations that will be equally or more expensive than `findMatchingIndex()`, including iterating `selectors` with `Array.prototype.some()` [a few lines above][1].

##
If anything, I feel we should take a more holistic approach (for example, replacing `Array` methods with `for`-loops in the whole file (or package or across packages) instead of just one occurrence).

Based on the above, I will close this PR, since I believe the downsides of merging this isolated change (such as the resulting code inconsistency) outweight the benefits. However, I would like to better understand your motivation/reasoning, so feel free to continue the discussion below (or open a new issue or PR to discuss).

[1]: https://github.com/angular/angular/blob/748129eb1a2450a2304b62300f3af73f23977735/packages/elements/src/extract-projectable-nodes.ts#L20-L26

Issue #34703, opened 2020-01-09
APP_INITIALIZER


mswilson4040:
# 📚 Docs or angular.io bug report

### Description

 A clear and concise description of the problem...
In the [docs](https://angular.io/api/core/APP_INITIALIZER), the APP_INITIALIZER is mentioned. However, there is only a single sentence and that's it. The sentence says:

> A function that will be executed when an application is initialized.

Based on many other 3rd party articles (such as Medium, SOF, etc..), everyone talks about the APP_INITIALIZER as the place to put code that you need to be loaded before your application is initialized. However, the sentence in the documentation makes it seem like it's a function that runs when the app is initialized.

There are other important parts of information such as what you can put inside the function. Does it have to return anything specific? A Promise? An Observable? Anything?

I've been working with angular for 6 years now and just ran into this. It's going to solve a lot of my problems if it does what the 3rd party articles say it does. It would be very helpful if more documentation was provided by the Angular team on how to actually use it the right way. This feature is very hidden. It would be nice if this was included in the Tour of Heroes as there is almost always something that needs to be pre-fetched to make the app work. In my case, it's the router configuration.

### What's the affected URL?**
 https://angular.io/...
https://angular.io/api/core/APP_INITIALIZER

## 🔥 Exception or Error
<pre><code>

</code></pre>

thefliik:
I also didn't know about this. Just looking at the [source code](https://github.com/angular/angular/blob/8.2.14/packages/core/src/application_init.ts#L19-L72) (linked to from the API docs), it looks like you can provide multiple `APP_INITIALIZER`s (`multi: true`) and, if any app initializers return a promise, the application will wait for all promises to resolve.

The `APP_INITIALIZER` injection token's type signature is actually somewhat confusing, as it is typed as `Array<() => void>`:

```ts
export const APP_INITIALIZER = new InjectionToken<Array<() => void>>('Application Initializer');
```

`() => void` makes it seem like Angular will ignore the `APP_INITIALIZER`'s return when, in fact, Angular checks to see if a promise is returned and, if so, waits for it to resolve. If the return type was `Array<() => void | Promise<void>>` it would be more clear.

smnbbrv:
> Based on many other 3rd part articles (such as Medium, SOF, etc..), everyone talks about the APP_INITIALIZER as the place to put code that you need to be loaded before your application is initialized. However, the sentence in the documentation makes it seem like it's a function that runs when the app is initialized.

[...]

PR #34760, opened 2020-01-13
Expand doc for APP_INITIALIZER token

jbogarthyde:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

The description of https://angular.io/api/core/APP_INITIALIZER is brief and uninformative.

Issue Number:  https://github.com/angular/angular/issues/34703

## What is the new behavior?
Expands the API doc description. 
**Needs an example or pointer to additional doc to be really useful.**
The subject does not appear to be covered anywhere in the current doc, but lots of external discussion.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/application_init.ts
+++ b/packages/core/src/application_init.ts
@@ -12,13 +12,15 @@ import {Inject, Injectable, InjectionToken, Optional} from './di';
 
 
 /**
- * An injection token that allows you to inject one or more functions to be executed during
+ * An injection token that allows you to provide one or more initialization functions.
+ * These function are injected at application startup and executed during
  * app initialization. If any of these functions returns a Promise, initialization
  * does not complete until the Promise is resolved.
  *
- * You can, for example, create a factory function that loads an external configuration
- * or language data, and provide that function to a service using this token, so that it is
- * executed during the application bootstrap process.
+ * You can, for example, create a factory function that loads language data
+ * or an external configuration, and provide that function to the `APP_INITIALIZER` token.
+ * That way, the function is executed during the application bootstrap process,
+ * and the needed data is available on startup.
  *
  * @publicApi
  */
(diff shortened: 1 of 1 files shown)


AndrewKushnir (approved):
👍 LGTM, thanks for the update.

I think it'd be great to add an example in the followup PR on how this token can be used.

packages/core/src/application_init.ts
```diff
@@ -12,7 +12,13 @@ import {Inject, Injectable, InjectionToken, Optional} from './di';
 
 
 /**
- * A function that will be executed when an application is initialized.
+ * An injection token that allows you to inject one or more functions to be executed during
```
CaerusKaru: I think we should say `provide one or more functions to be executed`. Those functions then get injected at application startup and run. This verbiage is similar to how the value is actually used in code:
```ts
{
   provide: APP_INITIALIZER,
   useValue: () => {}
}
```

packages/core/src/application_init.ts
```diff
@@ -12,7 +12,13 @@ import {Inject, Injectable, InjectionToken, Optional} from './di';
...
+ * app initialization. If any of these functions returns a Promise, initialization
+ * does not complete until the Promise is resolved.
+ *
+ * You can, for example, create a factory function that loads an external configuration
```
CaerusKaru: This whole example has me a little confused, mostly I think just by the wording. If I get your intention correctly, maybe this would be more direct?

```
You can, for example, create a factory function that loads language data or an external configuration, and provide that function to the APP_INITIALIZER token. 
That way, the function is executed during the application bootstrap process, and the needed data is available on startup.
```

jbogarthyde: Thank you!

PR #34769, opened 2020-01-14
Simplify bazel saucelabs targets

gregmagolan:
refactor: simplify bazel saucelabs targets using karma pre-test wrapper and shared saucelabs connection between tests

* Added a /tools/saucelabs/sauce-service.sh script that manages the sauce-connect as a service which is used by the karma-saucelabs.js wrapper to start the service.
* Added /tools/saucelabs/README.md that covers the details of SauceLabs karma testing with Bazel.


gregmagolan:
Added cleanup tag to remind me to remove the temporary CircleCI config changes before merging

gregmagolan:
Since `saucelabs_view_engine` CI job is master-only, a separate PR turns that job on https://github.com/angular/angular/pull/34937 for CI testing.

.circleci/config.yml
```diff
@@ -315,19 +315,22 @@ jobs:
...
-          # fail to acquire browsers. For example:
-          # 14 02 2019 19:52:33.170:WARN [launcher]: chrome beta on SauceLabs have not captured in 180000 ms, killing.
-          # //packages/forms/test:web_test_sauce TIMEOUT in 315.0s
-          command:  |
-            ./scripts/saucelabs/run-bazel-via-tunnel.sh \
-              --tunnel-id angular-${CIRCLE_BUILD_NUM}-${CIRCLE_NODE_INDEX} \
-              --username $SAUCE_USERNAME \
-              --key $(echo $SAUCE_ACCESS_KEY | rev) \
-              -- yarn bazel test //:saucelabs_unit_tests_poc --config=saucelabs
+          name: Preparing environment for running tests on Saucelabs.
```
josephperrott: We should just move setting this environment variable into the `env.sh` script

.circleci/config.yml
```diff
@@ -670,17 +681,20 @@ jobs:
             setPublicVar KARMA_JS_BROWSERS $(node -e 'console.log(require("./browser-providers.conf").sauceAliases.CI_REQUIRED.join(","))')
             setSecretVar SAUCE_ACCESS_KEY $(echo $SAUCE_ACCESS_KEY | rev)
       - run:
-          name: Starting Saucelabs tunnel
-          command: ./scripts/saucelabs/start-tunnel.sh
+          name: Starting Saucelabs tunnel service
```
josephperrott: I think this doesn't need to be `background: true` anymore since the `sauce-service.sh` exits after start now

Additionally, I think we can move this to be after the building of the packages, modules and zone.js

BUILD.bazel
```diff
@@ -47,68 +48,149 @@ filegroup(
...
+    karma_web_test(
+        name = "saucelabs_unit_tests_poc_%s" % test.replace("/", "_").replace(":", "_").replace(".", "_"),
+        # Default timeout is moderate (5min). This causes the test to be terminated while
+        # Saucelabs browsers keep running. Ultimately resulting in failing tests and browsers
+        # unnecessarily being acquired. Our specified Saucelabs idle timeout is 10min, so we use
+        # Bazel's long timeout (15min). This ensures that Karma can shut down properly.
+        timeout = "long",
+        karma = ":karma-sauce-connect",
+        tags = [
+            "exclusive",
```
josephperrott: Technically, I think the `no-remote-exec` isn't necessary since `exclusive` will already make it so everything is local and in a serial order.

BUILD.bazel
```diff
@@ -47,68 +48,149 @@ filegroup(
     ],
 )
 
-# To run a karma_web_test target manually, run the "./scripts/saucelabs/run-bazel-via-tunnel.sh"
-# script. Note: If you are on MacOS or Windows, you need to manually start the Saucelabs tunnel
-# because the script only supports the linux Saucelabs tunnel binary. We combine all tests into
-# a single "karma_web_test" target because running them as separate targets in parallel can result
-# in to too many browsers being acquired at the same time. This will then result in tests being
-# flaky. This target runs in CI with Saucelabs and Ivy.
+sh_library(
```
josephperrott: Should we move all of these `sh_library`/`sh_binary` targets into something line `/tools/saucelabs`

We can also include a README file there that explains how to use the saucelabs setup?

josephperrott (changes requested):

BUILD.bazel
gregmagolan: Oh. Seems you are correct, although it doesn't hurt to have the redundant tag since you'd have to go to the docs on `exclusive` to determine that.

.circleci/config.yml
gregmagolan: `./scripts/saucelabs/sauce-service.sh run` is the synchronous command for that script. Just runs the sc and doesn't exit. I had problems with CircleCI closing the process before the next `run` if I used `./scripts/saucelabs/sauce-service.sh start`

BUILD.bazel
gregmagolan: Done & added the README as well.

.circleci/config.yml
gregmagolan: SAUCE_ACCESS_KEY is there but in reverse so it is obfuscated hence the `$(echo $SAUCE_ACCESS_KEY | rev)`. Really it should be set in the CircleCI secret environment variables so that it is not in the code base at all. Who has access to the CircleCI account?

.circleci/config.yml
gregmagolan: With regards to the order, its done before the build intentionally AFAIK so that the the tunnel is connecting (which can take a bit of time) during build so by the the time `./scripts/saucelabs/sauce-service.sh ready-wait` is called is should be instantaneous. This legacy saucelabs job will go away in a follow up PR as well.

tools/saucelabs/README.md
```diff
@@ -0,0 +1,51 @@
+# Saucelabs testing with Bazel
+
+## Local testing
+
+Setup your SAUCE_USERNAME, SAUCE_ACCESS_KEY & SAUCE_TUNNEL_IDENTIFIER environment variables. These are required. On OSX, also set SAUCE_CONNECT to the path of your `sc` binary.
```
josephperrott: can we style the env variables?

Setup your `SAUCE_USERNAME`, `SAUCE_ACCESS_KEY`...

tools/saucelabs/README.md
```diff
@@ -0,0 +1,51 @@
...
+```
+
+or if the tests are combined into a test suite:
+
+```
+yarn bazel run //tools/saucelabs:sauce_service_setup
+yarn bazel test //path/to:saucelabs_test_suite --config=saucelabs [--config=ivy]
+```
+
+To see the test output while the tests are running as these are long tests, add the `--test_output=streamed` option. Note, this option will also prevent bazel from using he test cache and will force the test to run.
```
josephperrott: /s/using he test cache/using the test cache

tools/saucelabs/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+The `//tools/saucelabs:sauce_service_setup` target does not start the Sauce Connect proxy but it does start process which will that then listens for the start signal from the service manager script. This signal is sent by the karma wrapper script ``//tools/saucelabs:karma-saucelabs` which calls `./tools/saucelabs/sauce-service.sh start`. This is necessary as the Sauce Connect Proxy process must be started outside of `bazel test` as Bazel will automatically kill any processes spwaned during a test when that tests completes.
```
josephperrott: There is a double \`\` before `//tools/saucelabs:karama-saucelabs`

tools/saucelabs/sauce-service.sh
```diff
@@ -0,0 +1,408 @@
...
+}
+
+# Called by ready-wait action
+service-ready-wait() {
+  if [[ ! -f "${SAUCE_PID_FILE}" ]]; then
+    @fail "Sauce Connect not running"
+  fi
+  if [[ ! -f "${SAUCE_READY_FILE}" ]]; then
+    # Wait for saucelabs tunnel to connect
+    printf "# Waiting for saucelabs tunnel to connect (${SAUCE_READY_FILE})"
```
josephperrott: Should this be `@echo "Waiting for saucelabs tunnel to connect (${SAUCE_READY_FILE})"`

tools/saucelabs/README.md
gregmagolan: 👍

tools/saucelabs/README.md
gregmagolan: 👍

tools/saucelabs/README.md
gregmagolan: 👍

tools/saucelabs/sauce-service.sh
gregmagolan: Intentional printf so that its followed by ....... on the same line below with `printf "."`

tools/saucelabs/sauce-service.sh
```diff
@@ -0,0 +1,414 @@
...
+  local sauce_args=(
+    "--no-ssl-bump-domains all"
+    "--logfile ${SAUCE_LOG_FILE}"
+    "--pidfile ${SAUCE_PID_FILE}"
+    "--readyfile ${SAUCE_READY_FILE}"
+    "--tunnel-identifier ${SAUCE_TUNNEL_IDENTIFIER}"
+    "--user ${SAUCE_USERNAME}"
+    # Don't add the --api-key here so we don't echo it out in service-pre-start
+  )
+  echo "${SAUCE_CONNECT} ${sauce_args[@]} --api-key ${SAUCE_ACCESS_KEY}"
```
josephperrott: Is this going to echo out the key?

josephperrott (approved):
LGTM

tools/saucelabs/sauce-service.sh
gregmagolan: No. `service-command` is echo'ing out into a local variable in the service code
```
local command=$(service-command)
```
This is so that only the thread that actually runs SauceConnect needs the SAUCE_* environment variables set. They are not needed by the shell script in the general case.

Removed that comment there as it was outdated and confusing.

gregmagolan: On second thought I'll avoid that pattern and set a SERVICE_COMMAND var in `setup-service-command` instead and I can still echo out the command without the api-key at `setup`.
Cleanup up.

PR #34772, opened 2020-01-14
docs: replace non-breaking with regular spaces in `resources.json`

gkalpak:
By using non-breaking spaces, the description of Amexio UI components was overflowing its container in the "Resources" page.

**Before:**
amexio before

**After:**
amexio after


IgorMinar (approved):
lgtm. thanks!

PR #34782, opened 2020-01-15
feat(language-service): support multiple symbol definitions

ayazhafiz:
In Angular, symbol can have multiple definitions (e.g. a two-way
binding). This commit adds support for for multiple definitions for a
queried location in a template.

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/packages/language-service/src/expressions.ts
+++ b/packages/language-service/src/expressions.ts
@@ -148,8 +148,9 @@ export function getExpressionSymbol(
     },
     visitPropertyWrite(ast) {
       const receiverType = getType(ast.receiver);
+      const {start} = ast.span;
       symbol = receiverType && receiverType.members().get(ast.name);
-      span = ast.span;
+      span = {start, end: start + ast.name.length};
     },
     visitQuote(ast) {},
     visitSafeMethodCall(ast) {
--- a/packages/language-service/src/expressions.ts
+++ b/packages/language-service/src/expressions.ts
@@ -150,6 +150,11 @@ export function getExpressionSymbol(
       const receiverType = getType(ast.receiver);
       const {start} = ast.span;
       symbol = receiverType && receiverType.members().get(ast.name);
+      // A PropertyWrite span includes both the LHS (name) and the RHS (value) of the write. In this
+      // visit, only the name is relevant.
+      //   prop=$event
+      //   ^^^^        name
+      //        ^^^^^^ value; visited separately as a nested AST
       span = {start, end: start + ast.name.length};
     },
     visitQuote(ast) {},
--- a/packages/language-service/src/locate_symbol.ts
+++ b/packages/language-service/src/locate_symbol.ts
@@ -12,7 +12,7 @@ import {AstResult} from './common';
 import {getExpressionScope} from './expression_diagnostics';
 import {getExpressionSymbol} from './expressions';
 import {Definition, DirectiveKind, Span, Symbol} from './types';
-import {diagnosticInfoFromTemplateInfo, findTemplateAstAt, getPathToNodeAtPosition, inSpan, offsetSpan, spanOf} from './utils';
+import {diagnosticInfoFromTemplateInfo, findTemplateAstAt, getPathToNodeAtPosition, inSpan, isNarrower, offsetSpan, spanOf} from './utils';
 
 export interface SymbolInfo {
   symbol: Symbol;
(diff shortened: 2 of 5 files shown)


ayazhafiz:
Note that in some cases, there a symbol can have multiple of the same definition -- for example, in the attribute value of a two way binding. `[(ngModel)]="prop"` really expands to `[ngModel]="prop" (ngModel)="prop=$event"`, so when the definition for `prop` is requested, it is twice-referencing the `prop` member of the template's component. I did not include pruning duplicate entries in this commit because the LSP engines I tested this on (Vim via CoC and VSCode) handled these cases gracefully, combining the two definitions, but we can add code to prune them if desired..

packages/language-service/src/expressions.ts
```diff
@@ -148,8 +148,9 @@ export function getExpressionSymbol(
     },
     visitPropertyWrite(ast) {
       const receiverType = getType(ast.receiver);
+      const {start} = ast.span;
       symbol = receiverType && receiverType.members().get(ast.name);
-      span = ast.span;
+      span = {start, end: start + ast.name.length};
```
kyliau: How is this different from the original `ast.span`? Could you please add a comment here?

packages/language-service/src/locate_symbol.ts
```diff
@@ -21,124 +21,138 @@ export interface SymbolInfo {
...
+  while (!path.empty && isNarrower(spanOf(path.tail.sourceSpan), narrowest)) {
+    toVisit.push(path.tail);
+    path.pop();
+  }
+
+  return toVisit.map(ast => locateSymbol(ast, path, info, position))
+      .filter((sym): sym is SymbolInfo => sym !== undefined);
+}
+
+function locateSymbol(ast: TemplateAst, path: TemplateAstPath, info: AstResult, position: number):
```
kyliau: the `position` parameter is not needed, because it's available in the `path`.

kyliau: Could you please add a brief function doc and list the assumptions here?
e.g. `path` must not be empty.

packages/language-service/src/locate_symbol.ts
```diff
@@ -21,124 +21,138 @@ export interface SymbolInfo {
...
-          const dinfo = diagnosticInfoFromTemplateInfo(info);
-          const scope = getExpressionScope(dinfo, path);
-          if (attribute.valueSpan) {
-            const result = getExpressionSymbol(scope, ast, templatePosition, info.template.query);
-            if (result) {
-              symbol = result.symbol;
-              const expressionOffset = attribute.valueSpan.start.offset;
-              span = offsetSpan(result.span, expressionOffset);
-            }
+  let symbol: Symbol|undefined = undefined;
```
kyliau: GitHub does a bad job at showing the diff here. Is there any change in code besides the indentation?

packages/language-service/test/definitions_spec.ts
```diff
@@ -64,13 +64,14 @@ describe('definitions', () => {
 
     expect(textSpan).toEqual(marker);
     expect(definitions).toBeDefined();
-    expect(definitions !.length).toBe(1);
-    const def = definitions ![0];
+    expect(definitions !.length).toBe(2);
```
kyliau: Please add a comment on why there are two definitions here.

kyliau: Since you're here, do you mind refactoring the spec to use the test template?
```ts
mockHost.override(TEST_TEMPLATE, `<input [(ngModel)]="title">`)
```
Thank you :)

packages/language-service/test/definitions_spec.ts
```diff
@@ -64,13 +64,14 @@ describe('definitions', () => {
...
 
-    expect(def.fileName).toBe(fileName);
-    expect(def.name).toBe('name');
-    expect(def.kind).toBe('property');
-    expect(def.textSpan).toEqual(mockHost.getDefinitionMarkerFor(fileName, 'name'));
+    for (const def of definitions !) {
+      expect(def.fileName).toBe(fileName);
+      expect(def.name).toBe('name');
+      expect(def.kind).toBe('property');
+      expect(def.textSpan).toEqual(mockHost.getDefinitionMarkerFor(fileName, 'name'));
```
kyliau: Suggestion:
read the content of the file,
then assert content.substring(span.start, span.start + span.length) to be "name: string;"
That way we don't need to add markers to the TS file.

packages/language-service/test/definitions_spec.ts
```diff
@@ -304,18 +305,24 @@ describe('definitions', () => {
...
+    expect(def1.kind).toBe('property');
+    let content = mockHost.readFile(refFileName) !;
+    expect(content.substring(def1.textSpan.start, def1.textSpan.start + def1.textSpan.length))
         .toEqual(`@Input() model: string = 'model';`);
+
+    expect(def2.fileName).toBe(refFileName);
+    expect(def2.name).toBe('modelChange');
+    expect(def2.kind).toBe('event');
+    content = mockHost.readFile(refFileName) !;
+    expect(content.substring(def2.textSpan.start, def2.textSpan.start + def2.textSpan.length))
```
kyliau: +1

kyliau (changes requested):

packages/language-service/src/expressions.ts
ayazhafiz: Sure! Thanks for reminding me.

packages/language-service/src/locate_symbol.ts
ayazhafiz: Nope.

packages/language-service/test/definitions_spec.ts
ayazhafiz: It's not possible to solely use the test template for this test because we need to grab a member of the component class. If you'd like, we can for instance override 'app.component.ts'.

ayazhafiz: Ah, I spoke too soon. Read your comment below. Nice, thanks!

packages/language-service/test/definitions_spec.ts
```diff
@@ -304,18 +306,24 @@ describe('definitions', () => {
...
-    expect(def.fileName).toBe(refFileName);
-    expect(def.name).toBe('model');
-    expect(def.kind).toBe('property');
-    const content = mockHost.readFile(refFileName) !;
-    expect(content.substring(def.textSpan.start, def.textSpan.start + def.textSpan.length))
+    expect(def1.fileName).toBe(refFileName);
+    expect(def1.name).toBe('model');
+    expect(def1.kind).toBe('property');
+    let content = mockHost.readFile(refFileName) !;
+    expect(content.substr(def1.textSpan.start, def1.textSpan.length))
```
kyliau: `substr` is considered a legacy function and should be avoided. Please use `substring` instead.
See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr

packages/language-service/test/definitions_spec.ts
```diff
@@ -49,28 +50,29 @@ describe('definitions', () => {
...
+    // property and event bindings of the two-way binding. The two-way binding is effectively
+    // syntactic sugar for `[ngModel]="name" (ngModel)="name=$event"`.
+    expect(definitions !.length).toBe(2);
+    for (const def of definitions !) {
+      expect(def.fileName).toBe(PARSING_CASES);
+      expect(def.name).toBe('title');
+      expect(def.kind).toBe('property');
+
+      const fileContent = mockHost.readFile(def.fileName);
+      expect(fileContent !.substr(def.textSpan.start, def.textSpan.length))
```
kyliau: `substr` is considered legacy function and should be avoided. Please use `substring` instead.
See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr

packages/language-service/test/definitions_spec.ts
```diff
@@ -304,18 +306,24 @@ describe('definitions', () => {
...
+    expect(def1.kind).toBe('property');
+    let content = mockHost.readFile(refFileName) !;
+    expect(content.substr(def1.textSpan.start, def1.textSpan.length))
         .toEqual(`@Input() model: string = 'model';`);
+
+    expect(def2.fileName).toBe(refFileName);
+    expect(def2.name).toBe('modelChange');
+    expect(def2.kind).toBe('event');
+    content = mockHost.readFile(refFileName) !;
+    expect(content.substr(def2.textSpan.start, def2.textSpan.length))
```
kyliau: ditto

packages/language-service/test/definitions_spec.ts
```diff
@@ -304,18 +306,24 @@ describe('definitions', () => {
...
     expect(textSpan).toEqual(boundedText);
 
-    // There should be exactly 1 definition
     expect(definitions).toBeDefined();
-    expect(definitions !.length).toBe(1);
-    const def = definitions ![0];
+    expect(definitions !.length).toBe(2);
+    const [def1, def2] = definitions !;
 
     const refFileName = '/app/parsing-cases.ts';
```
kyliau: replace with PARSING_CASES

packages/language-service/test/definitions_spec.ts
```diff
@@ -49,28 +50,29 @@ describe('definitions', () => {
...
-    expect(definitions !.length).toBe(1);
-    const def = definitions ![0];
 
-    expect(def.fileName).toBe(fileName);
-    expect(def.name).toBe('name');
-    expect(def.kind).toBe('property');
-    expect(def.textSpan).toEqual(mockHost.getDefinitionMarkerFor(fileName, 'name'));
+    // There are exactly two, indentical definitions here, corresponding to the "name" on the
+    // property and event bindings of the two-way binding. The two-way binding is effectively
+    // syntactic sugar for `[ngModel]="name" (ngModel)="name=$event"`.
```
kyliau: use `title` instead of `name` to be consistent with the test case?
`(ngModel)="name=$event"` should be `(ngModelChange)="$event=title"`

kyliau (changes requested):

packages/language-service/test/definitions_spec.ts
ayazhafiz: Went ahead and replaced all string literal instances with PARSING_CASES.

packages/language-service/test/definitions_spec.ts
ayazhafiz: I need to pay more attention. Thank you.

packages/language-service/src/locate_symbol.ts
```diff
@@ -21,124 +21,146 @@ export interface SymbolInfo {
...
```
kyliau: the RHS assignment is unnecessary.

packages/language-service/src/locate_symbol.ts
```diff
@@ -21,124 +21,146 @@ export interface SymbolInfo {
...
+  let span: Span|undefined = undefined;
```
kyliau: ditto

packages/language-service/test/definitions_spec.ts
kyliau: This is still `substr`, did you forget to push your local changes?

kyliau (changes requested):

kyliau (approved):

PR #34786, opened 2020-01-15
docs: fix public api tool name

alan-agius4:
The tool in question is named `TS API Guardian` and not `public API guard`.

### DIFF
--- a/docs/PUBLIC_API.md
+++ b/docs/PUBLIC_API.md
@@ -51,7 +51,6 @@ Angular tracks the status of the public API in *golden files*. These files are s
 If you modify any part of a public API in one of the supported public packages, you must update the golden files that track the public API surface.
 If the golden files are not updated correctly, our test suite will detect the problem and the PR will fail to pass the tests.
 In such case a test error message will provide instructions for a command to run to update the golden files.
-The TS API Guardian provides a Bazel target that updates the current status of a given package. If you add to or modify the public API in any way, you must use [yarn](https://yarnpkg.com/) to execute the Bazel target in your terminal shell of choice (a recent version of `bash` is recommended).
 
 ```shell
 yarn bazel run //tools/public_api_guard:<modified_package>_api.accept
--- a/docs/PUBLIC_API.md
+++ b/docs/PUBLIC_API.md
@@ -48,9 +48,9 @@ Our peer dependencies (such as TypeScript, Zone.js, or RxJS) are not considered 
 
 ## Golden files
 
-Angular tracks the status of the public API in *golden files*. These files are stored in [tools/public_api_guard/](../tools/public_api_guard)` and are maintained with a tool called [TS API Guardian](https://www.npmjs.com/package/ts-api-guardian). If you modify any part of a public API in one of the supported public packages, the PR can fail a Circle CI test with failures in `testlogs/tools/public_api_guard`, and an error message that instructs you to accept the golden file.
+Angular tracks the status of the public API in *golden files*. These files are stored in [tools/public_api_guard/](../tools/public_api_guard) and are maintained with a tool called [TS API Guardian](https://www.npmjs.com/package/ts-api-guardian). If you modify any part of a public API in one of the supported public packages, the PR can fail a Circle CI test with failures in `testlogs/tools/public_api_guard`, and an error message that instructs you to accept the golden file.
 
-The TS API Guardian provides a Bazel target that updates the current status of a given package. If you add to or modify the public API in any way, you must use [yarn](https://yarnpkg.com/) to execute the Bazel target in your terminal shell of choice (a recent version of `bash` is recommended). 
+The TS API Guardian provides a Bazel target that updates the current status of a given package. If you add to or modify the public API in any way, you must use [yarn](https://yarnpkg.com/) to execute the Bazel target in your terminal shell of choice (a recent version of `bash` is recommended).
 
 ```shell
 yarn bazel run //tools/public_api_guard:<modified_package>_api.accept
--- a/docs/PUBLIC_API.md
+++ b/docs/PUBLIC_API.md
@@ -47,9 +47,10 @@ Our peer dependencies (such as TypeScript, Zone.js, or RxJS) are not considered 
 <a name="golden-files"></a>
 
 ## Golden files
-
-Angular tracks the status of the public API in *golden files*. These files are stored in [tools/public_api_guard/](../tools/public_api_guard) and are maintained with a tool called [TS API Guardian](https://www.npmjs.com/package/ts-api-guardian). If you modify any part of a public API in one of the supported public packages, the PR can fail a Circle CI test with failures in `testlogs/tools/public_api_guard`, and an error message that instructs you to accept the golden file.
-
+Angular tracks the status of the public API in *golden files*. These files are stored in [tools/public_api_guard/](../tools/public_api_guard) and are maintained with a tool called [TS API Guardian](https://www.npmjs.com/package/ts-api-guardian).
+If you modify any part of a public API in one of the supported public packages, you must update the golden files that track the public API surface.
+If the golden files are not updated correctly, our test suite will detect the problem and the PR will fail to pass the tests.
+In such case a test error message will provide instructions for a command to run to update the golden files.
 The TS API Guardian provides a Bazel target that updates the current status of a given package. If you add to or modify the public API in any way, you must use [yarn](https://yarnpkg.com/) to execute the Bazel target in your terminal shell of choice (a recent version of `bash` is recommended).
 
 ```shell
--- a/docs/PUBLIC_API.md
+++ b/docs/PUBLIC_API.md
@@ -48,9 +48,9 @@ Our peer dependencies (such as TypeScript, Zone.js, or RxJS) are not considered 
 
 ## Golden files
 
-Angular tracks the status of the public API in a *golden file*, maintained with a tool called the *TS API Guardian*.  If you modify any part of a public API in one of the supported public packages, the PR can fail a Circle CI test with failures in `testlogs/tools/public_api_guard`, and an error message that instructs you to accept the golden file.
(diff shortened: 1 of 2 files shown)


gkalpak (approved):

gkalpak (approved):
While you're at it, it might be helpful to also mention that the "golden files" are kept in the `tools/public_api_guard/` directory.

alan-agius4:
@gkalpak PTAL.

docs/PUBLIC_API.md
```diff
@@ -48,9 +48,9 @@ Our peer dependencies (such as TypeScript, Zone.js, or RxJS) are not considered
 
 ## Golden files
 
-Angular tracks the status of the public API in a *golden file*, maintained with a tool called the *public API guard*.  If you modify any part of a public API in one of the supported public packages, the PR can fail a Circle CI test with failures in `testlogs/tools/public_api_guard`, and an error message that instructs you to accept the golden file.
+Angular tracks the status of the public API in *golden files*. These files are stored in `tools/public_api_guard` and are maintained with a tool called [TS API Guardian](https://www.npmjs.com/package/ts-api-guardian). If you modify any part of a public API in one of the supported public packages, the PR can fail a Circle CI test with failures in `testlogs/tools/public_api_guard`, and an error message that instructs you to accept the golden file.
```
gkalpak:
```suggestion
Angular tracks the status of the public API in *golden files*. These files are stored in [tools/public_api_guard/](../tools/public_api_guard)` and are maintained with a tool called [TS API Guardian](https://www.npmjs.com/package/ts-api-guardian). If you modify any part of a public API in one of the supported public packages, the PR can fail a Circle CI test with failures in `testlogs/tools/public_api_guard`, and an error message that instructs you to accept the golden file.
```

gkalpak: Wow, I didn't know we publish TS API Guardian 👍

gkalpak (approved):

docs/PUBLIC_API.md
alan-agius4: Yeah we do :)

alan-agius4: Change done!

docs/PUBLIC_API.md
```diff
@@ -48,9 +48,9 @@ Our peer dependencies (such as TypeScript, Zone.js, or RxJS) are not considered
...
+Angular tracks the status of the public API in *golden files*. These files are stored in [tools/public_api_guard/](../tools/public_api_guard)` and are maintained with a tool called [TS API Guardian](https://www.npmjs.com/package/ts-api-guardian). If you modify any part of a public API in one of the supported public packages, the PR can fail a Circle CI test with failures in `testlogs/tools/public_api_guard`, and an error message that instructs you to accept the golden file.
 
-The public API guard provides a Bazel target that updates the current status of a given package. If you add to or modify the public API in any way, you must use [yarn](https://yarnpkg.com/) to execute the Bazel target in your terminal shell of choice (a recent version of `bash` is recommended).
+The TS API Guardian provides a Bazel target that updates the current status of a given package. If you add to or modify the public API in any way, you must use [yarn](https://yarnpkg.com/) to execute the Bazel target in your terminal shell of choice (a recent version of `bash` is recommended).
```
gkalpak: 😱 😱 😱 

```suggestion
The TS API Guardian provides a Bazel target that updates the current status of a given package. If you add to or modify the public API in any way, you must use [yarn](https://yarnpkg.com/) to execute the Bazel target in your terminal shell of choice (a recent version of `bash` is recommended).
```

docs/PUBLIC_API.md
```diff
@@ -48,9 +48,9 @@ Our peer dependencies (such as TypeScript, Zone.js, or RxJS) are not considered
 
 ## Golden files
 
-Angular tracks the status of the public API in a *golden file*, maintained with a tool called the *public API guard*.  If you modify any part of a public API in one of the supported public packages, the PR can fail a Circle CI test with failures in `testlogs/tools/public_api_guard`, and an error message that instructs you to accept the golden file.
+Angular tracks the status of the public API in *golden files*. These files are stored in [tools/public_api_guard/](../tools/public_api_guard)` and are maintained with a tool called [TS API Guardian](https://www.npmjs.com/package/ts-api-guardian). If you modify any part of a public API in one of the supported public packages, the PR can fail a Circle CI test with failures in `testlogs/tools/public_api_guard`, and an error message that instructs you to accept the golden file.
```
gkalpak: My bad. Forgot to remove the closing `` ` `` in my previous suggestion.

```suggestion
Angular tracks the status of the public API in *golden files*. These files are stored in [tools/public_api_guard/](../tools/public_api_guard) and are maintained with a tool called [TS API Guardian](https://www.npmjs.com/package/ts-api-guardian). If you modify any part of a public API in one of the supported public packages, the PR can fail a Circle CI test with failures in `testlogs/tools/public_api_guard`, and an error message that instructs you to accept the golden file.
```

gkalpak (approved):

alan-agius4:
@gkalpak PTAL.

docs/PUBLIC_API.md
```diff
@@ -48,9 +48,9 @@ Our peer dependencies (such as TypeScript, Zone.js, or RxJS) are not considered
 
 ## Golden files
 
-Angular tracks the status of the public API in a *golden file*, maintained with a tool called the *public API guard*.  If you modify any part of a public API in one of the supported public packages, the PR can fail a Circle CI test with failures in `testlogs/tools/public_api_guard`, and an error message that instructs you to accept the golden file.
+Angular tracks the status of the public API in *golden files*. These files are stored in [tools/public_api_guard/](../tools/public_api_guard) and are maintained with a tool called [TS API Guardian](https://www.npmjs.com/package/ts-api-guardian). If you modify any part of a public API in one of the supported public packages, the PR can fail a Circle CI test with failures in `testlogs/tools/public_api_guard`, and an error message that instructs you to accept the golden file.
```
IgorMinar:
```suggestion
Angular tracks the status of the public API in *golden files*. These files are stored in [tools/public_api_guard/](../tools/public_api_guard) and are maintained with a tool called [TS API Guardian](https://www.npmjs.com/package/ts-api-guardian).
If you modify any part of a public API in one of the supported public packages, you must update the golden files that track the public API surface.
If the golden files are not updated correctly, our test suite will detect the problem and the PR will fail to pass the tests.
In such case a test error message will provide instructions for a command to run to update the golden files.
```

docs/PUBLIC_API.md
```diff
@@ -48,9 +48,9 @@ Our peer dependencies (such as TypeScript, Zone.js, or RxJS) are not considered
...
 
-The public API guard provides a Bazel target that updates the current status of a given package. If you add to or modify the public API in any way, you must use [yarn](https://yarnpkg.com/) to execute the Bazel target in your terminal shell of choice (a recent version of `bash` is recommended).
+The TS API Guardian provides a Bazel target that updates the current status of a given package. If you add to or modify the public API in any way, you must use [yarn](https://yarnpkg.com/) to execute the Bazel target in your terminal shell of choice (a recent version of `bash` is recommended).
```
IgorMinar: I suggest that you drop this as the error message should give you the right instructions. Or do you think something is missing?

IgorMinar (changes requested):

docs/PUBLIC_API.md
alan-agius4: The command is already printed in the error message 
https://github.com/angular/angular/blob/1fd4d4f0d4e9ef6d05815e999533e65cff26c51c/tools/ts-api-guardian/lib/cli.ts#L82-L93

docs/PUBLIC_API.md
alan-agius4: Updated

docs/PUBLIC_API.md
alan-agius4: Though In all fairness that document link shouldn't be there, because not all API guardian users use the same commands. So that document is only valid for Angular/Angular

alan-agius4: Should we remove it?

IgorMinar: so in that case, we don't need to to add this paragraph to doc, and we should likely drop that link or even better print the link only if the guardian is running in the angular repo (which can be inferred from ENV or even the BAZEL_TARGET itself.

docs/PUBLIC_API.md
```diff
@@ -47,10 +47,10 @@ Our peer dependencies (such as TypeScript, Zone.js, or RxJS) are not considered
...
-
-The public API guard provides a Bazel target that updates the current status of a given package. If you add to or modify the public API in any way, you must use [yarn](https://yarnpkg.com/) to execute the Bazel target in your terminal shell of choice (a recent version of `bash` is recommended).
+Angular tracks the status of the public API in *golden files*. These files are stored in [tools/public_api_guard/](../tools/public_api_guard) and are maintained with a tool called [TS API Guardian](https://www.npmjs.com/package/ts-api-guardian).
+If you modify any part of a public API in one of the supported public packages, you must update the golden files that track the public API surface.
+If the golden files are not updated correctly, our test suite will detect the problem and the PR will fail to pass the tests.
+In such case a test error message will provide instructions for a command to run to update the golden files.
```
Splaktar: I think that we should leave the note about "a recent version of `bash` is recommended" as I personally ran into that issue and it was a pain to track down without a hint like this. It affects all macOS users as the Apple Bash isn't 'recent' enough.

Splaktar: I'm fine with removing Yarn since it's been added to the error message now. That part was added before the error message was updated.

alan-agius4: I personally never encountered the issue were I had to update bash. But I am fine with re-adding that part if @IgorMinar & @gkalpak agree.

kapunahelewong: A typo fix and a wording suggestion:
```suggestion
In such cases, a test error message will provide instructions for a command that updates the golden files.
```

alan-agius4: Done

IgorMinar (approved):
lgtm! thank you!

PR #34792, opened 2020-01-15
ngtsc as a plugin!

alxhub:
It's happening!!

This PR implements `NgTscPlugin` on top of some refactored compiler APIs. Hopefully this will allow integration with `ts_library` in the near future.

There's lots of rough edges left here, but it looks promising already!


packages/compiler-cli/src/ngtsc/tsc_plugin.ts
```diff
@@ -6,77 +6,88 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {PluginCompilerHost, TscPlugin} from '@bazel/typescript/internal/tsc_wrapped/plugin_api';
+import {PluginCompilerHost} from '@bazel/typescript/internal/tsc_wrapped/plugin_api';
 import * as ts from 'typescript';
 
-import {SyntheticFilesCompilerHost} from './synthetic_files_compiler_host';
```
alexeagle: I think you can delete this file now

packages/compiler-cli/src/ngtsc/tsc_plugin.ts
```diff
@@ -6,77 +6,88 @@
...
+  setupCompilation(program: ts.Program, oldProgram?: ts.Program): void;
+
+  getDiagnostics(file?: ts.SourceFile): ts.Diagnostic[];
+
+  getOptionDiagnostics(): ts.Diagnostic[];
+
+  getNextProgram(): ts.Program;
+
+  prepareEmit(): {
+    transformers: {
```
alexeagle: could you use the type `ts.CustomTransformers` here?

packages/compiler-cli/src/ngtsc/tsc_plugin.ts
```diff
@@ -6,77 +6,88 @@
...
-function createProxy<T>(delegate: T): T {
-  const proxy = Object.create(null);
-  for (const k of Object.keys(delegate)) {
-    proxy[k] = function() { return (delegate as any)[k].apply(delegate, arguments); };
+/**
+ * Mirrors the plugin interface from tsc_wrapped which is currently under active development. To
+ * enable progress to be made in parallel, the upstream interface isn't implemented directly.
+ * Instead, `TscPlugin` here is structurally assignable to what tsc_wrapped expects.
+ */
+interface TscPlugin {
```
alexeagle: could you add a member `name = 'ngtsc'` so that it matches the interface LitPlugin introduced for gathering diagnostics?

packages/compiler-cli/src/ngtsc/tsc_plugin.ts
```diff
@@ -6,77 +6,88 @@
...
+  wrapHost(
+      host: PluginCompilerHost, inputFiles: ReadonlyArray<string>,
```
alexeagle: why does this need to be a `PluginCompilerHost`? in our work last week we wrapped a `ts.CompilerHost` which was the only thing constructed at this point

packages/compiler-cli/src/ngtsc/tsc_plugin.ts
```diff
@@ -6,77 +6,88 @@
...
+      options: ts.CompilerOptions): ts.CompilerHost;
```
alexeagle: and I think it needs to return a PluginCompilerHost so we can locate the inputFiles when we construct a ts.Program

packages/compiler-cli/src/ngtsc/core/src/compiler.ts
```diff
@@ -0,0 +1,809 @@
...
+      if (this.entryPoint !== null && compilation.exportReferenceGraph !== null) {
+        this.diagnostics.push(...checkForPrivateExports(
+            this.entryPoint, this.tsProgram.getTypeChecker(), compilation.exportReferenceGraph));
+      }
+    }
+
+    if (file === undefined) {
+      return this.diagnostics;
+    } else {
+      return this.diagnostics.filter(diag => diag.file === file);
```
alexeagle: this is broken - diag.file is the .ng.html file but file is the .ts file

packages/compiler-cli/src/ngtsc/core/src/host.ts
```diff
@@ -0,0 +1,248 @@
...
+    Omit<RequiredCompilerHostDelegations, 'getSourceFile'|'fileExists'> {
+  constructor(protected delegate: ExtendedTsCompilerHost) {}
+
+  private delegateMethod<M extends keyof ExtendedTsCompilerHost>(name: M):
+      ExtendedTsCompilerHost[M] {
+    return this.delegate[name] !== undefined ? (this.delegate[name] as any).bind(this.delegate) :
+                                               undefined;
+  }
+
+  // Excluded are 'getSourceFile' and 'fileExists', which are actually implemented by NgCompilerHost
```
JoostK: Did you consider collecting all delegate methods into an object literal, and copying over the ones that are present in the delegate host but not on the prototype as custom methods? This way, we can avoid these exclusions _and_ fully omit setting properties that the delegate doesn't have anyways.

Something like this:

```ts
type RequiredCompilerHostDelegations = {
   [M in keyof Required<ExtendedTsCompilerHost>]: null;
}

const delegateMethods: RequiredCompilerHostDelegations  = {
  createHash: null,
  directoryExists: null,
  // ... etc.
}

export class DelegatingCompilerHost implements ExtendedTsCompilerHost {
  constructor(protected delegate: ExtendedTsCompilerHost) {
    for (const delegateMethod of Object.keys(delegateMethods)) {
      if (!(delegateMethod in this) && (delegate as any)[delegateMethod] !== undefined) {
       (this as any)[delegateMethod] = (delegate as any)[delegateMethod].bind(delegate);
      }
    }
  }

}
```

JoostK (commented):
Stellar work! I like how the compiler objects are nicely separated and how the shim generator host has been centralized into the `NgCompilerHost`. Some minor comments for your consideration.

packages/compiler-cli/src/ngtsc/tsc_plugin.ts
alxhub: Done.

packages/compiler-cli/src/ngtsc/core/src/host.ts
alxhub: Yes, but I prefer the explicitness of adding each method directly. The above won't actually type-check because dynamically adding the methods won't satisfy the interface implementation, so you'll need to add a lie to the type system to address that.

packages/compiler-cli/src/ngtsc/tsc_plugin.ts
```diff
@@ -6,77 +6,88 @@
...
-                         ts.createSourceFile(factoryPath, 'contents', ts.ScriptTarget.ES5),
-      };
-    });
+  private options: NgCompilerOptions|null = null;
+  private host: NgCompilerHost|null = null;
+  private _compiler: NgCompiler|null = null;
+
+  get compiler(): NgCompiler {
+    if (this._compiler === null) {
+      throw new Error('Lifecycle error: getDiagnostics() called before setupCompilation()');
```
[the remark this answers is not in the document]
alxhub: Fixed :)

packages/compiler-cli/src/ngtsc/core/src/host.ts
```diff
@@ -0,0 +1,248 @@
...
+// the delegating host doesn't implement or delegate them. This causes subtle runtime failures. No
+// more. This infrastructure ensures that failing to delegate a method is a compile-time error.
+
+/**
+ * Represents the `ExtendedTsCompilerHost` interface, with a transformation applied that turns all
+ * methods (even optional ones) into required fields (which may be `undefined`, if the method was
+ * optional).
+ */
+export type RequiredCompilerHostDelegations = {
+  [M in keyof Required<ExtendedTsCompilerHost>]: ExtendedTsCompilerHost[M]
```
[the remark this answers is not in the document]
alxhub: 👍

packages/compiler-cli/src/ngtsc/core/src/compiler.ts
```diff
@@ -0,0 +1,809 @@
...
+      new ComponentDecoratorHandler(
+          reflector, evaluator, metaRegistry, metaReader, scopeReader, scopeRegistry, isCore,
+          this.resourceManager, this.host.rootDirs, this.options.preserveWhitespaces || false,
+          this.options.i18nUseExternalIds !== false,
+          this.options.enableI18nLegacyMessageIdFormat !== false, this.moduleResolver,
+          this.cycleAnalyzer, refEmitter, defaultImportTracker, this.incrementalDriver.depGraph,
+          injectableRegistry, this.closureCompilerEnabled),
+      // TODO(alxhub): understand why the cast here is necessary (something to do with `null`
+      // not
+      // being assignable to `unknown` when wrapped in `Readonly`).
```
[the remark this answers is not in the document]
alxhub: 👍

packages/compiler-cli/src/ngtsc/core/src/compiler.ts
```diff
@@ -0,0 +1,809 @@
...
+    // The formatter really doesn't appreciate these casts, and no amount of begging or pleading
+    // with parentheses seems to convince it to leave these lines alone. Thus, it must be disabled
+    // for each one.
+
+    const before = [
+      ivyTransformFactory(
+          compilation.traitCompiler, compilation.reflector, importRewriter,
+          compilation.defaultImportTracker, compilation.isCore, this.closureCompilerEnabled),
+      // clang-format off
+      aliasTransformFactory(compilation.traitCompiler.exportStatements) as ts.TransformerFactory<ts.SourceFile>,
```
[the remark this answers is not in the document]
alxhub: Done. Much cleaner!

packages/compiler-cli/src/ngtsc/tsc_plugin.ts
alxhub: Done.

packages/compiler-cli/src/ngtsc/tsc_plugin.ts
alxhub: I've refactored it to take a `ts.CompilerHost` plus the optional `fileNameToModuleName`.

packages/compiler-cli/src/ngtsc/tsc_plugin.ts
alxhub: Done.

packages/compiler-cli/src/ngtsc/core/src/compiler.ts
alxhub: Fixed.

packages/compiler-cli/src/ngtsc/core/api.ts
```diff
@@ -0,0 +1,360 @@
...
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import * as ts from 'typescript';
+
+/**
+ * A host backed by a build system which has a unified view of the module namespace.
+ *
+ * Such a build system is able to map be supports the `moduleNameToFileName` method provided by
```
JoostK: I don't understand this comment, `able to map be supports the` seems incorrect.

packages/compiler-cli/src/ngtsc/core/src/compiler.ts
```diff
@@ -0,0 +1,810 @@
...
+      }
+    }
+
+    if (file === undefined) {
+      return this.diagnostics;
+    } else {
+      return this.diagnostics.filter(diag => {
+        if (diag.file === file) {
+          return true;
+        } else if (isTemplateDiagnostic(diag) && diag.componentFile === file) {
```
JoostK: This would be worth having in a unit test, I don't think we have such a test at the moment.

JoostK (approved):
LGTM, with some more minor comments

alxhub:
This is blocked on the upstream g3 work to standardize the plugin API, as well as #34887 landing first.

alexeagle:
Should be unblocked for g3sync now - I commented out the relevant bit in tsc_wrapped

packages/compiler-cli/src/ngtsc/core/src/compiler.ts
alxhub: I agree, but I'm gonna defer it because we don't have any testing infrastructure that invokes the compiler at this level yet.

alxhub:
[Presubmit](http://test/OCL:290825946:BASE:293477683:1580948480305:c5a0aede)

alxhub:
Caretaker: I cannot force g3 status ("This SHA and context has reached the maximum number of statuses" error). Presubmits are good (only Ivy failures are screenshot diffs & timeouts).

LayZeeDK:
So this is something Bazel-related? Would anyone care to explain what the big deal is? 🤔

PR #34797, opened 2020-01-15
docs: do not add auto-code-links to observables example

### LINKED ISSUE #34793: More hyperlinks in code comments

### Description

On this page
https://angular.io/guide/observables
the following code comment contains the whole bunch of hyperrefs (create, a, sequence)
// Create a new Observable that will deliver the above sequence

https://angular.io/guide/observables

## 🔥 Exception or Error
<pre><code>

</code></pre>

petebacondarwin:
Fixes #34793

See https://pr34797-e671b3b.ngbuilds.io/guide/observables

### DIFF
--- a/aio/content/guide/observables.md
+++ b/aio/content/guide/observables.md
@@ -17,7 +17,7 @@ To execute the observable you have created and begin receiving notifications, yo
 
 Here's an example that demonstrates the basic usage model by showing how an observable could be used to provide geolocation updates.
 
-<code-example path="observables/src/geolocation.ts" header="Observe geolocation updates"></code-example>
+<code-example class="no-auto-link" path="observables/src/geolocation.ts" header="Observe geolocation updates"></code-example>
 
 ## Defining observers
 


IgorMinar (approved):
lgtm, but shouldn't this be the default for all examples?

my guess is that most examples will contain bogus links because the auto-linking is primarily meant to work for API docs. no?

IgorMinar:

petebacondarwin:
@IgorMinar - there may well be a number of false-positives, but they are not that noticeable (and so do not negatively impact on the docs) since you only realise the link is there if you hover over it. This is evidenced by how few issues we receive like this one.

I think that there is enough value in generally having code linked to API docs, that it is worth having a few false positives.

One mitigation could be to create a stop-list of terms that should not be linked.

gkalpak:
The stop list might be worth for things that are likely to generate false positives. E.g. things such as `a`, `group`, `sequence`.

PR #34798, opened 2020-01-15
fix: work around 'noImplicityAny' incompatibility due to ts3.7 update

IgorMinar:
Typescript 3.7 now emits d.ts files for getters differently than prior versions,
and there seems to be a bug in how it strips private types without replacing them
with explicit 'any' type. This then leads to compilation failures in projects compiled
against our packages that don't have skipLibCheck turned on but do have strict or
noImplicitAny check on.

I'm working around this by marking the affected getters as @Internal and
adding a test to prevent future regressions.

Typescript bug report: https://github.com/microsoft/TypeScript/issues/36216


josephperrott (approved):
LGTM

IgorMinar:

JoostK:
Upstream fix in TypeScript has been backported to 3.6.5: https://github.com/microsoft/TypeScript/releases/tag/v3.6.5

PR #34814, opened 2020-01-16
ci: Migrate from CODEOWNERS to PullApprove for PR review management

josephperrott:
We are migrating to PullApprove for our PR review management in an attempt
to allow for more granular and equitable code review assignments across the
team.  Currently this migration is equivalent in the review assignments
it will create. Once stable, our expectation is that we will be able to
take advantage of PullApprove's additional features for things like staged
reviews.


.pullapprove.yml
```diff
@@ -0,0 +1,961 @@
...
+          'packages/compiler-cli/src/ngtools/**',
+          'aio/content/guide/cli-builder.md',
+          'aio/content/guide/ivy.md',
+          'aio/content/guide/web-worker.md'
+          ])
+    reviewers:
+      users:
+        - filipesilva
+        - mgechev
+        - vikerman
```
CaerusKaru: Are we keeping users who are no longer on the team? (Also noticed AlexE above)

.circleci/config.yml
```diff
@@ -278,7 +278,6 @@ jobs:
               (echo -e "\n.bzl files have lint errors. Please run ''yarn bazel:lint-fix''"; exit 1)'
 
       - run: yarn gulp lint
-      - run: node tools/verify-codeownership
```
gkalpak: 😞 Shouldn't we have a similar check for pullapprove?

josephperrott: Yes, The plan is to do this logic/check in a follow up PR.

.pullapprove.yml
```diff
@@ -0,0 +1,955 @@
...
+#  dgp1130 - Doug Parker
+#  filipesilva - Filipe Silva
+#  gkalpak - Georgios Kalpakas
+#  gregmagolan - Greg Magolan
+#  igorminar - Igor Minar
+#  jbogarthyde - Judy Bogart
+#  jelbourn - Jeremy Elbourn
+#  jialipassion - Jia Li
+#  josephperrott - Joey Perrott
+#  juleskremer - Jules Kremer
```
JoostK: I have gone missing from this list for some reason (I'm still in the compiler group below)

josephperrott: Sorry! I used the listing at `aio/content/marketing/contributors.json` to create the list and must have accidentally dropped you out of the list.

CaerusKaru: Does this include collaborators (noticed JiaLi in the list)? If so, I'm missing too!

IgorMinar (changes requested):
Posting comments from my first pass... I also pushed a commit with some fixes I made locally.

Things we should not forget to update:

- [ ] ng-bot config to require pullapprove status (`.github/angular-robot.yml`)
- [ ] `docs/COMMITTER.md` which makes references to codeowners
- [ ] `docs/TRIAGE_AND_LABELS.md` which also makes references to codeowners
- [ ] what to do with existing gh groups? delete them?

IgorMinar:
@josephperrott can you please review my commit + rebase this? thanks

.pullapprove.yml
josephperrott: The intention is for it to be a list that contains all of the usernames that are present as an owner for any of the groups defined in the file.

.pullapprove.yml
```diff
@@ -0,0 +1,955 @@
...
+    explanation: "Waiting to send reviews as PR is WIP"
+
+groups:
+  # =========================================================
+  #  Animations
+  # =========================================================
+  animations:
+    conditions:
+      - >
+        contains_any_globs(files, [
```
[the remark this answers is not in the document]
josephperrott: `matches_any_globs` probably makes more sense, but is not the name of the method that they went with.  I think it comes from a consistency in naming conventions with there other matcher methods.

Yes, all of the files modified in the PR.

josephperrott (commented):
Created https://github.com/angular/angular/issues/34956 for the verify-codeownership followup.

.pullapprove.yml
CaerusKaru: @IgorMinar @josephperrott Something I’ve been wondering for a while then, can we add the Googlers and Collaborators for Universal as members on platform-server? (That would be me, @alan-agius4, @kyliau, and @manfredsteyer)

IgorMinar (approved):
I've rebased this and pushed a last set of fixup commits. I think this is good to go.

I'm marking this as "blocked" so that it is merged only once @josephperrott is around to babysit this PR and ensure that it doesn't break the world. 😄

IgorMinar:
live long and prosper pullaprove 🖖

PR #27038, opened 2018-11-10
feat: remove @angular/http

CaerusKaru:
BREAKING CHANGE:

* This removes `@angular/http` from the list of supported Angular core APIs. The new official API for HTTP in Angular is `@angular/common/http`

PR #34817, opened 2020-01-16
refactor(http): fix a strictBindCallApply issue.

mprobst:
`String.fromCharCode`'s type signature requires a regular `number[]`.

## PR Type

### DIFF
--- a/packages/http/src/body.ts
+++ b/packages/http/src/body.ts
@@ -57,9 +57,13 @@ export abstract class Body {
     if (this._body instanceof ArrayBuffer) {
       switch (encodingHint) {
         case 'legacy':
-          return String.fromCharCode.apply(null, new Uint16Array(this._body));
+          // TODO: Argument of type 'Uint16Array' is not assignable to parameter of type
+          // 'number[]'.
+          return String.fromCharCode.apply(null, new Uint16Array(this._body) as any);
         case 'iso-8859':
-          return String.fromCharCode.apply(null, new Uint8Array(this._body));
+          // TODO: Argument of type 'Uint8Array' is not assignable to parameter of type
+          // 'number[]'.
+          return String.fromCharCode.apply(null, new Uint8Array(this._body) as any);
         default:
           throw new Error(`Invalid value for encodingHint: ${encodingHint}`);
       }


CaerusKaru:
The HTTP package is no longer published or supported, and I'd hate to set a precedent that we're still updating it. If this is a blocker for `strict` support elsewhere, my vote is for finally deleting this package via #27038.

cc @IgorMinar

kapunahelewong:
Thank you for requesting my review, @mprobst. I'd like to wait until engineering has had a chance to review and then come to check it from a docs perspective. Will you tag me once the engineering review process is done (in case I miss it)?

mprobst:
@CaerusKaru this PR fixes a trivial compilation issue with a (not that) new TS strictness flag. I understand that a more general cleanup is planned, but given this is a one line essentially, I don't think we should block on the more general cleanup (which I suspect is hard - we'd need to remove all users of the code, not just the code itself!).

@kapunahelewong ack, but I think there's no docs impact in this PR.

CaerusKaru:
@mprobst HTTP hasn’t been updated since v7, and that won’t change. The users of HTTP (granted, only those outside g3) have already been dealt with. I very strongly believe we should just continue with the removal, especially since most of the work has been done to do so (in the PR linked above)

mprobst:
@CaerusKaru sure. What I'm fixing is that currently Angular does not compile with `strictBindCallApply`. I think that's orthogonal to whether or not to remove this code. Given that the timeline to remove the code as I understand it is not on the order of this week, I think it's worth landing this and then deleting the code later.

CaerusKaru:
@mprobst I totally understand your point of view, which is why I’d like @IgorMinar to weigh in before we merge this. I don’t think the work is on the order of a week (unless you mean google3 work), but that’s just my opinion.

kapunahelewong:
@mprobst ah yes, thank you! This was the first one that auto requested me. I'm catching on to the new notifications. Never mind! 🙃

IgorMinar:
@CaerusKaru this change is needed in g3 to compile the monorepo with this flag. External users/programs are not affected because they don't compile this package from sources.

In short, we should merge this and sync it to g3 but not release it on npm.

IgorMinar (approved):
Lgtm.

I'll create a kanban task for us to delete this code and stop syncing it. For g3, we should make g3 the source of truth of this code.

Issue #34794, opened 2020-01-15
Concluding note on hyperlinks in comments and not in code


pasha-bolokhov:
# 📚 Docs or angular.io bug report

### Description

While I have noticed that multiple angular doc webpages contain sudden hyperlinks, at the same time, the places where I would *really* love to see hyperrefs do not have them: such as `pipe()` or, `filter()`, or the library `rxjs/operators` itself

Here is the next one I'm reporting — the following page
https://angular.io/guide/rx-library
in the following comment also contains unnecessary hyperlinks
// Create a function that accepts an Observable.

I suspect many more pages have the same issue

### What's the affected URL?**
https://angular.io/guide/rx-library

## 🔥 Exception or Error
<pre><code>

</code></pre>

PR #34835, opened 2020-01-17
build(docs-infra): add auto-link-filter for generic words

kapunahelewong:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #34794 

## What is the new behavior?

Adds filter to prevent autolinking words like "create" 

- [ ] Yes
- [x] No

### DIFF
--- a/aio/tools/transforms/angular-base-package/services/auto-link-filters/ignoreGenericWords.js
+++ b/aio/tools/transforms/angular-base-package/services/auto-link-filters/ignoreGenericWords.js
@@ -1,0 +1,10 @@
+
+/**
+ * This service is used by the autoLinkCode post-processor to ignore generic words
+ * that could be mistaken as classes, functions, etc.
+ */
+
+module.exports = function ignoreGenericWords() {
+  const ignoredWords = new Set(['a', 'create', 'error', 'group', 'request', 'value']);
+  return (docs, words, index) => ignoredWords.has(words[index].toLowerCase()) ? [] : docs;
+};
--- a/aio/tools/transforms/angular-base-package/index.js
+++ b/aio/tools/transforms/angular-base-package/index.js
@@ -40,6 +40,7 @@ module.exports = new Package('angular-base', [
   .factory(require('./services/auto-link-filters/filterPipes'))
   .factory(require('./services/auto-link-filters/filterAmbiguousDirectiveAliases'))
   .factory(require('./services/auto-link-filters/ignoreHttpInUrls'))
+  .factory(require('./services/auto-link-filters/ignoreGenericWords'))
 
   .factory(require('./post-processors/add-image-dimensions'))
   .factory(require('./post-processors/auto-link-code'))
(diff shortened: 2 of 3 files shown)


aio/tools/transforms/angular-base-package/services/auto-link-filters/ignoreGenericWords.js
```diff
@@ -0,0 +1,10 @@
+
+/**
+ * This service is used by the autoLinkCode post-processor to ignore generic words
+ * that could be mistaken as classes, functions, etc.
+ */
+
+module.exports = function ignoreGenericTerms() {
```
gkalpak: Please name this the same as the file (either rename the function or the file :smiley:)

kapunahelewong: oh oops!

petebacondarwin (approved):
This is awesome! I wonder if it might be worth making use of the `ignore.words` file? Or perhaps it is a different enough scenario that it warrants a separate words list?

gkalpak (approved):
LGTM :tada:
I believe the commit subject should be `build(docs-infra): ...` btw.

aio/tools/transforms/angular-base-package/services/auto-link-filters/ignoreGenericWords.js
```diff
@@ -0,0 +1,10 @@
+
+/**
+ * This service is used by the autoLinkCode post-processor to ignore generic words
+ * that could be mistaken as classes, functions, etc.
+ */
+
+module.exports = function ignoreGenericWords() {
+  const ignoredWords = ['a', 'create', 'error', 'request', 'value', 'group'];
```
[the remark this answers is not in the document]
kapunahelewong: Oh I see, alphabetical. Got it. Thanks!

aio/tools/transforms/angular-base-package/services/auto-link-filters/ignoreGenericWords.js
```diff
@@ -0,0 +1,10 @@
+
+/**
+ * This service is used by the autoLinkCode post-processor to ignore generic words
+ * that could be mistaken as classes, functions, etc.
+ */
+
+module.exports = function ignoreGenericWords() {
+  const ignoredWords = ['a', 'create', 'error', 'group', 'request', 'value'];
+  return (docs, words, index) => ignoredWords.includes(words[index].toLowerCase()) ? [] : docs;
```
petebacondarwin: I wonder if we should be considering performance here?

This function will get called for every single word in every example. Do we have any idea of the performance impact of doing this? If it is negligible (since it is a small array) then no problem. But if not, perhaps we should consider a `Set` instead?

gkalpak: Good spot! +1 for using a `Set` (it can't hurt even if not necessary :grin:)

petebacondarwin: Just don't create a new `Set` on each function call :-)

gkalpak: Even that can't be much worse 😛

kapunahelewong: I've changed it to a `Set`

gkalpak: You also need to change `ignoredWords.includes` (which is an `Array` method) with [ignoredWords.has][1] (which is a `Set` method).

[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/has

gkalpak (approved):
👌

aio/tools/transforms/angular-base-package/services/auto-link-filters/ignoreGenericWords.js
kapunahelewong: Changed the method to `.has()` and ran `yarn docs-only` in `angular/aio` to check performance (thank you both!!). These were my results:

On `master`:

- Done in 26.63s.
- Done in 26.34s.
- Done in 28.17s.

On my branch:

- Done in 26.61s.
- Done in 25.18s.
- Done in 25.04s.

PR #34839, opened 2020-01-17
fix(ivy): support @SkipSelf while injecting special DI tokens

### LINKED ISSUE #34066: In Ivy, SkipSelf() is not accessing the parent's ViewContainerRef, ElementRef, etc

### LINKED ISSUE #34819: SkipSelf annotation to get parent component Injector no longer working in Ivy (Angular 9 rc.9)
# 🐞 bug report

### Affected Package

### Is this a regression?
Yes, with Angular 8 this was still possible.

### Description
I have a component B that is a child of component A. With previous Angular versions I was able to inject the `Injector` of component A into component B by simply using the `@SkipSelf` annotation on the injected service. With that, I could then resolve services in the context of the parent component.

In my use case, both A and B provide the same injection token with different fixed values and component B needs to know what this value was for its parent component.

## 🔬 Minimal Reproduction
https://stackblitz.com/edit/angular-parent-injector-gx5pha

The example on StackBlitz is of course in Angular 8 (since I can't seem to do Angular 9 rc projects) and works as expected. The output on the console is:

<pre><code>
this is "HELLO COMPONENT" and parent has name "THIS IS ROOT"
this is "HELLO COMPONENT" and parent has name "THIS IS ROOT"
</code></pre>

When I download the example and upgrade it to Angular 9 rc.9 I get different output:

<pre><code>
this is "HELLO COMPONENT" and parent has name "THIS IS ROOT"
this is "HELLO COMPONENT" and parent has name "HELLO COMPONENT"
</code></pre>

The second line is produced with the parent `Injector` injection approach.

## 🌍  Your Environment

**Angular Version:**
I tested this with Angular 9 rc.7, rc.8 and rc.9 and the example fails in all of them.

**Anything else relevant?**
In the past I have tried other aproaches to solve this issue, one of them was to use the `Injector.get` method and specifying the `InjectFlags` argument as documented in the [Injector](https://angular.io/api/core/Injector). It seems that in all versions of Angular (that I tested), the injector instance that I get when requesting an `Injector` on a component simply _ignores_ the inject flags - in the source code I can see that the method simply does not use this argument. Somehow I find this very weird since the documentation does not seem to mention that.

AndrewKushnir:
Currently Ivy doesn't support adding `@SkipSelf` for some of the special DI tokens (e.g. `Injector`, `ViewContainerRef`, `ElementRef`, `ChangeDetectorRef`). These changes add support for `@SkipSelf` flag and align the behavior with View Engine.

Closes #34066.
Closes #34819.

## PR Type

- [x] Bugfix

- [ ] Yes
- [x] No

### DIFF
--- a/aio/scripts/_payload-limits.json
+++ b/aio/scripts/_payload-limits.json
@@ -12,7 +12,7 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 2987,
-        "main-es2015": 450017,
+        "main-es2015": 450763,
         "polyfills-es2015": 52195
       }
     }
--- a/integration/_payload-limits.json
+++ b/integration/_payload-limits.json
@@ -3,7 +3,7 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 1485,
-        "main-es2015": 141569,
+        "main-es2015": 142106,
         "polyfills-es2015": 36657
       }
     }
@@ -21,7 +21,7 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 1485,
-        "main-es2015": 147647,
+        "main-es2015": 148184,
         "polyfills-es2015": 36657
       }
     }
@@ -30,7 +30,7 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 1485,
-        "main-es2015": 136777,
+        "main-es2015": 137380,
         "polyfills-es2015": 37334
       }
     }
@@ -39,7 +39,7 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 2289,
-        "main-es2015": 247198,
+        "main-es2015": 247815,
         "polyfills-es2015": 36657,
         "5-es2015": 751
       }
@@ -49,7 +49,7 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 2289,
-        "main-es2015": 226144,
+        "main-es2015": 226689,
         "polyfills-es2015": 36657,
         "5-es2015": 779
       }
--- a/packages/core/src/render3/di.ts
+++ b/packages/core/src/render3/di.ts
@@ -23,7 +23,7 @@ import {DirectiveDef, FactoryFn} from './interfaces/definition';
 import {NO_PARENT_INJECTOR, NodeInjectorFactory, PARENT_INJECTOR, RelativeInjectorLocation, RelativeInjectorLocationFlags, TNODE, isFactory} from './interfaces/injector';
 import {AttributeMarker, TContainerNode, TDirectiveHostNode, TElementContainerNode, TElementNode, TNode, TNodeProviderIndexes, TNodeType} from './interfaces/node';
 import {isComponentDef, isComponentHost} from './interfaces/type_checks';
-import {DECLARATION_COMPONENT_VIEW, DECLARATION_VIEW, INJECTOR, LView, TData, TVIEW, TView, T_HOST} from './interfaces/view';
+import {DECLARATION_COMPONENT_VIEW, DECLARATION_VIEW, INJECTOR, LView, TData, TVIEW, TView, TViewType, T_HOST} from './interfaces/view';
 import {assertNodeOfPossibleTypes} from './node_assert';
 import {enterDI, leaveDI} from './state';
 import {isNameOnlyAttributeMarker} from './util/attrs_utils';
@@ -313,6 +313,12 @@ function getParentTNode(
   // Falling back to `T_HOST` in case we cross component boundary (thus `tNode.parent` is not
   // defined in this case), but do that only if `@Host()` is not present.
   if (parent === null && !(flags & InjectFlags.Host)) {
+    const tView = lView[TVIEW];
+    if (tView.type === TViewType.Embedded) {
+      // TODO: we should go up the chain until we find Element or ElementContainer node
+      return (tView as any).declarationTNode;
+    }
+    // TODO: check other TViewType types
     return lView[T_HOST] as TDirectiveHostNode;
   }
   return parent;
--- a/packages/core/src/render3/di.ts
+++ b/packages/core/src/render3/di.ts
@@ -15,7 +15,7 @@ import {InjectFlags} from '../di/interface/injector';
 import {Type} from '../interface/type';
 import {assertDefined, assertEqual} from '../util/assert';
 
-import {assertDirectiveDef} from './assert';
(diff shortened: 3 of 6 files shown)


AndrewKushnir:
Related to https://github.com/angular/angular/pull/34554.

integration/_payload-limits.json
```diff
@@ -3,7 +3,7 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 1485,
-        "main-es2015": 141476,
+        "main-es2015": 141985,
```
AndrewKushnir: I believe this is an accumulated size regression (i.e. not caused specifically by the changes in this PR).

IgorMinar: hmm.. ok

IgorMinar (approved):
lgtm

IgorMinar:

AndrewKushnir:
[Presubmit](http://test/OCL:291656781:BASE:291656800:1580104645768:c20ee630)

packages/core/src/render3/di.ts
```diff
@@ -345,10 +345,16 @@ export function getOrCreateInjectable<T>(
         leaveDI();
       }
     } else if (typeof bloomHash == 'number') {
+      // `-1` is a special value used to identify `Injector` types.
```
kara: Thank you for moving the comment up. More visible this way! :)

packages/core/src/render3/di.ts
```diff
@@ -345,10 +345,16 @@ export function getOrCreateInjectable<T>(
...
       if (bloomHash === -1) {
-        // `-1` is a special value used to identify `Injector` types.
+        if (flags & InjectFlags.SkipSelf) {
+          const _tNode = tNode;
+          tNode = tNode.parent || lView[T_HOST] as TDirectiveHostNode;
```
kara: Good that we are supporting crossing component boundaries, but if we're going to do this, we also need to respect `@Host` config flags. In other words, if you have:

```ts
@SkipSelf() @Host() public vcr: ViewContainerRef 
```

The `ViewContainerRef` should only be found if the parent element is in the same view.

Can we add a test for this as well?

kara (changes requested):

AndrewKushnir:
@kara thanks for the review! I've added extra logic to take `@Host()` into account. Could you please take another look? Thank you.

aio/scripts/_payload-limits.json
```diff
@@ -12,7 +12,7 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 2987,
-        "main-es2015": 450017,
+        "main-es2015": 450763,
```
kara: How much of this is accumulation and how much from this PR?

packages/core/src/render3/di.ts
```diff
@@ -345,10 +376,19 @@ export function getOrCreateInjectable<T>(
         leaveDI();
       }
     } else if (typeof bloomHash == 'number') {
+      // `-1` is a special value used to identify `Injector` types.
       if (bloomHash === -1) {
-        // `-1` is a special value used to identify `Injector` types.
+        if (flags & InjectFlags.SkipSelf) {
+          const parentInjectorLocation = getParentInjectorLocation(tNode, lView);
```
kara: Nit: Can you move this line down below the `if (tNode === null)` block? No need to look up the parent injector if we are going to return the not found value.

kara (approved):
LGTM

packages/core/src/render3/di.ts
AndrewKushnir: Thanks, updated!

aio/scripts/_payload-limits.json
AndrewKushnir: @kara here are the numbers for "main-es2015":
- baseline: 450017
- master: 450469 (**452b** over baseline)
- this branch: 450763 (**294b** over master)

So master on its own is very close to reaching **500b** threshold.

AndrewKushnir:
[Presubmit](http://test/OCL:291656781:BASE:296535856:1582331634314:457bd993)

AndrewKushnir:
@kara thanks for the review!

I'd like to ask you to have a look at this PR again, since I had to update a bunch of payload limits, which I believe for the most part is an accumulation of payload size increase introduced in other commits merged to master prior to this change, see https://github.com/angular/angular/pull/34839#discussion_r382851840.

Thank you.

kara (approved):
LGTM

AndrewKushnir:
Quick update: investigation of g3 failures revealed a couple use-cases that work incorrectly with the changes in this PR due to the `tView` structure implementation that doesn't allow to easily go *up* the chain (in case it's a cross-boundary read). In order to make further progress, some updates to `tView` structure is required. As a pre-requisite to that work, it'd be helpful to land PR #34715 to minimize the number of affected lines of code and avoid dealing with unnecessary tests. I'll clean up #34715 and request a review shortly. Thank you.

AndrewKushnir:
Closing this PR in favor of #39464.

PR #34846, opened 2020-01-18
fix(ivy): don't detect changes on detached child embedded views

### LINKED ISSUE #34816: [Ivy] detach view doesn't work
# 🐞 bug report

### Is this a regression?

Yes, worked perfectly fine in latest v8 without Ivy. Works also without Ivy in v9.

### Description

I have a directive that renders a template and detaches it from the CD until some criteria is true. Once true I reattach it again. However in Ivy the whole detach doesn't work. After calling `view.detach()` the view is still checked in change detector runs and thus uses resources. In non-Ivy that works correctly: The view is never used in change detector runs when detach was called.

## 🔬 Minimal Reproduction

https://stackblitz.com/edit/angular-issue-repro2-xd8kbu

This Stackblitz uses Angular 9.0.0rc9, **however** the issue is not reproducible directly in Stackblitz (only god knows why). In Stackblitz you see the correct behavior:
Screenshot 2020-01-16 at 19 00 05

But if you download that Stackblitz and run `npm install && ./node_modules/.bin/ng serve` then you see the actual issue:

Screenshot 2020-01-16 at 19 01 29

The second line behind the `:` is detached and should never display anything.

Code of that wrapper:

```
    <div>This should increase w/ each click: <cd-counter></cd-counter></div>
    <div *wrapper>This should stay empty or 0: <cd-counter></cd-counter></div>
```

```
@Directive({
  selector: "[wrapper]"
})
export class WrapperComponent implements OnInit {
  constructor(
    protected template: TemplateRef<any>,
    protected viewContainer: ViewContainerRef
  ) {}

  ngOnInit() {
    const view = this.viewContainer.createEmbeddedView(this.template);
    view.detach();
  }
}
```

## Versions

9.0.0: I tested rc7, rc8, rc9. None worked correctly. When I opt out of Ivy using `"enableIvy": false` it works correctly (only tested with with rc9).

Screenshot 2020-01-16 at 19 08 30

## 🌍  Your Environment

**Angular Version:**
<pre><code>Angular CLI: 9.0.0-rc.9
Node: 12.14.1
OS: darwin x64

Angular: 9.0.0-rc.9
... router
Ivy Workspace: Yes

@angular-devkit/architect         0.900.0-rc.9
@angular-devkit/build-angular     0.900.0-rc.9
@angular-devkit/build-optimizer   0.900.0-rc.9
@angular-devkit/build-webpack     0.900.0-rc.9
@angular-devkit/core              9.0.0-rc.9
@angular-devkit/schematics        9.0.0-rc.9
@angular/cdk                      9.0.0-rc.7
@ngtools/webpack                  9.0.0-rc.9
@schematics/angular               9.0.0-rc.9
@schematics/update                0.900.0-rc.9
</code></pre>

**Anything else relevant?**

It's interesting that in Stackblitz it works as intended, but when using Chrome and `ng serve` locally it doesn't work. I assume that Stackblitz uses some v8 components or not the Ivy renderer, but that's just a guess as I have no knowledge about how Stackblitz actually starts the angular dev server and whether they overwrite some settings.

crisbeto:
Fixes Ivy detecting changes inside child embedded views, even though they're detached.

Note that there's one subtlety here: I made the changes inside `refreshDynamicEmbeddedViews` rather than `refreshView`, because we support detecting changes on a detached view (evidenced by a couple of unit tests), but only if it's triggered directly from the view's `ChangeDetectorRef`, however we shouldn't be detecting changes in the detached child view when something happens in the parent.

**Note:** I had to bump the size limit, because it looks like master was right on the edge before these changes. I haven't introduced any new imports and the new logic is only a couple of lines so I doubt that these changes would've added too much to the size.

Fixes #34816.

### DIFF
--- a/packages/core/src/render3/instructions/shared.ts
+++ b/packages/core/src/render3/instructions/shared.ts
@@ -1525,7 +1525,10 @@ function refreshDynamicEmbeddedViews(lView: LView) {
         const embeddedLView = viewOrContainer[i];
         const embeddedTView = embeddedLView[TVIEW];
         ngDevMode && assertDefined(embeddedTView, 'TView must be allocated');
-        refreshView(embeddedLView, embeddedTView, embeddedTView.template, embeddedLView[CONTEXT] !);
+        if (viewAttachedToChangeDetector(embeddedLView)) {
+          refreshView(
+              embeddedLView, embeddedTView, embeddedTView.template, embeddedLView[CONTEXT] !);
+        }
       }
       if ((activeIndexFlag & ActiveIndexFlag.HAS_TRANSPLANTED_VIEWS) !== 0) {
         // We should only CD moved views if the component where they were inserted does not match
--- a/packages/core/test/acceptance/change_detection_spec.ts
+++ b/packages/core/test/acceptance/change_detection_spec.ts
@@ -22,10 +22,10 @@ describe('change detection', () => {
     @Directive({selector: '[viewManipulation]', exportAs: 'vm'})
     class ViewManipulation {
       constructor(
-          private _tplRef: TemplateRef<{}>, private _vcRef: ViewContainerRef,
+          private _tplRef: TemplateRef<{}>, public vcRef: ViewContainerRef,
           private _appRef: ApplicationRef) {}
 
-      insertIntoVcRef() { this._vcRef.createEmbeddedView(this._tplRef); }
+      insertIntoVcRef() { return this.vcRef.createEmbeddedView(this._tplRef); }
 
       insertIntoAppRef(): EmbeddedViewRef<{}> {
         const viewRef = this._tplRef.createEmbeddedView({});
@@ -43,11 +43,8 @@ describe('change detection', () => {
     class TestCmpt {
     }
 
-    beforeEach(() => {
-      TestBed.configureTestingModule({declarations: [TestCmpt, ViewManipulation]});
-    });
-
     it('should detect changes for embedded views inserted through ViewContainerRef', () => {
+      TestBed.configureTestingModule({declarations: [TestCmpt, ViewManipulation]});
       const fixture = TestBed.createComponent(TestCmpt);
       const vm = fixture.debugElement.childNodes[0].references['vm'] as ViewManipulation;
 
@@ -58,6 +55,7 @@ describe('change detection', () => {
     });
 
     it('should detect changes for embedded views attached to ApplicationRef', () => {
+      TestBed.configureTestingModule({declarations: [TestCmpt, ViewManipulation]});
       const fixture = TestBed.createComponent(TestCmpt);
       const vm = fixture.debugElement.childNodes[0].references['vm'] as ViewManipulation;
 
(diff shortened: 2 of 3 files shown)


packages/core/test/acceptance/change_detection_spec.ts
```diff
@@ -70,6 +68,49 @@ describe('change detection', () => {
       expect(viewRef.rootNodes[0]).toHaveText('change-detected');
     });
 
+    it('should not detect changes in child embedded views while they are detached', () => {
+      const counters = {outsideTemplate: 0, insideTemplate: 0};
```
pkozlowski-opensource: nit: personally I would name those to be named `componentView` and `embeddedView` as usage of the "template" in those names made me think that there are 2 `<ng-template>` elements here. It might be just me, though...

packages/core/test/acceptance/change_detection_spec.ts
```diff
@@ -70,6 +68,49 @@ describe('change detection', () => {
...
+        changeDetection: ChangeDetectionStrategy.OnPush
+      })
+      class App {
+        increment(counter: 'outsideTemplate'|'insideTemplate') { counters[counter]++; }
+        noop() {}
+      }
+
+      TestBed.configureTestingModule({declarations: [App, ViewManipulation]});
+      const fixture = TestBed.createComponent(App);
+      const vm: ViewManipulation = fixture.debugElement.childNodes[2].references['vm'];
```
pkozlowski-opensource: super-nit / personal preference: in cases like this I usually do `debugElement.query(By.directive(ViewManipulation)).injector.get(ViewManipulation)` as I find this approach less "brittle". (BTW, for quite a bit of time I wanted to have a better API to do just this in `TestBed` but this is a separate story....).

packages/core/test/acceptance/change_detection_spec.ts
```diff
@@ -70,6 +68,49 @@ describe('change detection', () => {
...
+      const viewRef = vm.insertIntoVcRef();
+      fixture.detectChanges();
+
+      expect(counters).toEqual({outsideTemplate: 1, insideTemplate: 1});
+
+      button.click();
+      fixture.detectChanges();
+      expect(counters).toEqual({outsideTemplate: 2, insideTemplate: 2});
+
+      viewRef.detach();
```
pkozlowski-opensource: yep, yep, yep.... corner case but definitively possible given our APIs....

packages/core/test/acceptance/change_detection_spec.ts
```diff
@@ -70,6 +68,49 @@ describe('change detection', () => {
...
+      expect(counters).toEqual({outsideTemplate: 3, insideTemplate: 2});
+
+      // Re-attach the view to ensure that the process can be reversed.
+      viewRef.reattach();
+      button.click();
+      fixture.detectChanges();
+
+      expect(counters).toEqual({outsideTemplate: 4, insideTemplate: 3});
+    });
+
```
pkozlowski-opensource: question / idea: do you think it would make sense to add another test where we insert a component view (and not only an embedded view) to make sure that we cover all the cases?

packages/core/src/render3/instructions/shared.ts
```diff
@@ -1525,7 +1525,10 @@ function refreshDynamicEmbeddedViews(lView: LView) {
         const embeddedLView = viewOrContainer[i];
         const embeddedTView = embeddedLView[TVIEW];
         ngDevMode && assertDefined(embeddedTView, 'TView must be allocated');
-        refreshView(embeddedLView, embeddedTView, embeddedTView.template, embeddedLView[CONTEXT] !);
+        if (viewAttachedToChangeDetector(embeddedLView)) {
```
pkozlowski-opensource: thinking out loud: I wonder if it wouldn't be better to move this check to the `refreshView` so we don't need to read `lView[FLAGS]` here _and_ inside `refreshView` for the most common case. WDYT?

pkozlowski-opensource (approved):
LGTM overall, just left few comments / questions that I would like have your thoughts on.

packages/core/src/render3/instructions/shared.ts
crisbeto: I commented on this in the PR description: I initially had it inside of `refreshView`, but that means that we can't detect changes on detached views which we seems to support (there are two unit tests for it).

packages/core/test/acceptance/change_detection_spec.ts
crisbeto: EDIT: I tried it out and `query` didn't manage to find the directive for some reason. Maybe because I'm trying to access it before the first change detection run?

Sure, I don't feel strongly about it. I did it this way, because the other unit tests were doing the same.

packages/core/test/acceptance/change_detection_spec.ts
crisbeto: Not sure what you mean by a component view. Do you mean through `ViewContainerRef.createComponent`?

crisbeto (commented):
I've addressed/replied to all of the feedback @pkozlowski-opensource.

packages/core/test/acceptance/change_detection_spec.ts
pkozlowski-opensource: yep, `ViewContainerRef.createComponent`

packages/core/src/render3/instructions/shared.ts
pkozlowski-opensource: Ohhhh, I see, this makes sense, you are right! Thnx for clarifying!

crisbeto:
I've added the extra unit test.

packages/core/test/acceptance/change_detection_spec.ts
crisbeto: Done.

pkozlowski-opensource (approved):
LGTM. The size increase is unfortunate but I believe that it is mostly because of the creeping size increase from the previous PRs.

AndrewKushnir:
[Presubmit](http://test/OCL:291382586:BASE:291382654:1579885557647:7fc18c5f)

IgorMinar (approved):
lgtm  - global approval (the size diff is not caused by this change)

PR #34847, opened 2020-01-18
feat(language-service): provide hover for microsyntax in structural directive

ivanwonder:
In #34564, I only support hover over the directive name. And have a problem, when hovering over `*ngFor="exp"`, it will show `ngForOf` property of `NgForOf` class, this pr will show `(directive) CommonModule.NgForOf: class`. This PR also supports the microsyntax(e.g. `trackBy `, `of`).

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/packages/language-service/src/locate_symbol.ts
+++ b/packages/language-service/src/locate_symbol.ts
@@ -55,19 +55,14 @@ function locateSymbol(ast: TemplateAst, path: TemplateAstPath, info: AstResult):
   let symbol: Symbol|undefined;
   let span: Span|undefined;
   let staticSymbol: StaticSymbol|undefined;
-  const attributeValueSymbol = (ast: AST): boolean => {
+  const attributeValueSymbol = (): boolean => {
     const attribute = findAttribute(info, position);
     if (attribute) {
       if (inSpan(templatePosition, spanOf(attribute.valueSpan))) {
-        const dinfo = diagnosticInfoFromTemplateInfo(info);
-        const scope = getExpressionScope(dinfo, path);
-        if (attribute.valueSpan) {
-          const result = getExpressionSymbol(scope, ast, templatePosition, info.template.query);
-          if (result) {
-            symbol = result.symbol;
-            const expressionOffset = attribute.valueSpan.start.offset;
-            span = offsetSpan(result.span, expressionOffset);
-          }
+        const result = getSymbolInAttributeValue(info, path, attribute);
+        if (result) {
+          symbol = result.symbol;
+          span = offsetSpan(result.span, attribute.valueSpan !.start.offset);
         }
         return true;
       }
@@ -105,13 +100,13 @@ function locateSymbol(ast: TemplateAst, path: TemplateAstPath, info: AstResult):
         },
         visitVariable(ast) {},
         visitEvent(ast) {
-          if (!attributeValueSymbol(ast.handler)) {
+          if (!attributeValueSymbol()) {
             symbol = findOutputBinding(info, path, ast);
             symbol = symbol && new OverrideKindSymbol(symbol, DirectiveKind.EVENT);
             span = spanOf(ast);
           }
         },
-        visitElementProperty(ast) { attributeValueSymbol(ast.value); },
+        visitElementProperty(ast) { attributeValueSymbol(); },
         visitAttr(ast) {
           const element = path.head;
           if (!element || !(element instanceof ElementAst)) return;
@@ -155,9 +150,22 @@ function locateSymbol(ast: TemplateAst, path: TemplateAstPath, info: AstResult):
           span = spanOf(ast);
         },
         visitDirectiveProperty(ast) {
-          if (!attributeValueSymbol(ast.value)) {
-            symbol = findInputBinding(info, templatePosition, ast);
-            span = spanOf(ast);
+          if (!attributeValueSymbol()) {
+            const directive = findParentOfBinding(info.templateAst, ast, templatePosition);
+            const attribute = findAttribute(info, position);
+            if (directive && attribute) {
+              if (attribute.name.startsWith('*')) {
+                const compileTypeSummary = directive.directive;
+                symbol = info.template.query.getTypeSymbol(compileTypeSummary.type.reference);
+                symbol = symbol && new OverrideKindSymbol(symbol, DirectiveKind.DIRECTIVE);
+                // Use 'attribute.sourceSpan' instead of the directive's,
+                // because the span of the directive is the whole opening tag of an element.
+                span = spanOf(attribute.sourceSpan);
+              } else {
+                symbol = findInputBinding(info, ast.templateName, directive);
+                span = spanOf(ast);
+              }
+            }
           }
         }
       },
@@ -169,6 +177,42 @@ function locateSymbol(ast: TemplateAst, path: TemplateAstPath, info: AstResult):
       span: tss.createTextSpanFromBounds(start, end), staticSymbol,
     };
   }
+}
+
+// Get the symbol in attribute value at template position.
+function getSymbolInAttributeValue(info: AstResult, path: TemplateAstPath, attribute: Attribute):
+    {symbol: Symbol, span: Span}|undefined {
+  if (!attribute.valueSpan) {
+    return;
+  }
+  let result: {symbol: Symbol, span: Span}|undefined;
+  const {templateBindings} = info.expressionParser.parseTemplateBindings(
+      attribute.name, attribute.value, attribute.sourceSpan.toString(),
+      attribute.valueSpan.start.offset);
+  // Find where the cursor is relative to the start of the attribute value.
+  const valueRelativePosition = path.position - attribute.valueSpan.start.offset;
+
+  // Find the symbol that contains the position.
+  templateBindings.filter(tb => !tb.keyIsVar).forEach(tb => {
+    if (inSpan(valueRelativePosition, tb.expression?.ast.span)) {
+      const dinfo = diagnosticInfoFromTemplateInfo(info);
+      const scope = getExpressionScope(dinfo, path);
+      result = getExpressionSymbol(scope, tb.expression !, path.position, info.template.query);
+    } else if (inSpan(valueRelativePosition, tb.span)) {
+      const template = path.first(EmbeddedTemplateAst);
+      if (template) {
+        // One element can only have one template binding.
+        const directiveAst = template.directives[0];
+        if (directiveAst) {
+          const symbol = findInputBinding(info, tb.key.substring(1), directiveAst);
+          if (symbol) {
+            result = {symbol, span: tb.span};
+          }
+        }
+      }
+    }
+  });
+  return result;
 }
 
 function findAttribute(info: AstResult, position: number): Attribute|undefined {
@@ -222,17 +266,15 @@ function findParentOfBinding(
   return res;
 }
 
-function findInputBinding(
-    info: AstResult, position: number, binding: BoundDirectivePropertyAst): Symbol|undefined {
-  const directiveAst = findParentOfBinding(info.templateAst, binding, position);
-  if (directiveAst) {
-    const invertedInput = invertMap(directiveAst.directive.inputs);
-    const fieldName = invertedInput[binding.templateName];
-    if (fieldName) {
-      const classSymbol = info.template.query.getTypeSymbol(directiveAst.directive.type.reference);
-      if (classSymbol) {
-        return classSymbol.members().get(fieldName);
-      }
+// Find the symbol of input binding in 'directiveAst' by 'name'.
+function findInputBinding(info: AstResult, name: string, directiveAst: DirectiveAst): Symbol|
+    undefined {
+  const invertedInput = invertMap(directiveAst.directive.inputs);
+  const fieldName = invertedInput[name];
+  if (fieldName) {
+    const classSymbol = info.template.query.getTypeSymbol(directiveAst.directive.type.reference);
+    if (classSymbol) {
+      return classSymbol.members().get(fieldName);
     }
   }
 }
--- a/packages/language-service/src/locate_symbol.ts
+++ b/packages/language-service/src/locate_symbol.ts
@@ -29,6 +29,8 @@ export function locateSymbols(info: AstResult, position: number): SymbolInfo[] {
   const templatePosition = position - info.template.span.start;
   // TODO: update `findTemplateAstAt` to use absolute positions.
   const path = findTemplateAstAt(info.templateAst, templatePosition);
+  const attribute = findAttribute(info, position);
+
   if (!path.tail) return [];
 
   const narrowest = spanOf(path.tail);
@@ -36,6 +38,11 @@ export function locateSymbols(info: AstResult, position: number): SymbolInfo[] {
   for (let node: TemplateAst|undefined = path.tail;
        node && isNarrower(spanOf(node.sourceSpan), narrowest); node = path.parentOf(node)) {
     toVisit.push(node);
+  }
+
+  // For the structural directive, only care about the last template AST.
+  if (attribute?.name.startsWith('*')) {
+    toVisit.splice(0, toVisit.length - 1);
   }
 
   return toVisit.map(ast => locateSymbol(ast, path, info))
--- a/packages/language-service/test/definitions_spec.ts
+++ b/packages/language-service/test/definitions_spec.ts
@@ -266,29 +266,86 @@ describe('definitions', () => {
     }
   });
 
-  it('should be able to find a structural directive', () => {
-    mockHost.override(TEST_TEMPLATE, `<div ~{start-my}*«ngIf»="true"~{end-my}></div>`);
+  describe('in structural directive', () => {
+    it('should be able to find the directive', () => {
+      mockHost.override(
+          TEST_TEMPLATE, `<div ~{start-my}*«ngFor»="let item of heroes;"~{end-my}></div>`);
 
-    // Get the marker for ngIf in the code added above.
-    const marker = mockHost.getReferenceMarkerFor(TEST_TEMPLATE, 'ngIf');
+      // Get the marker for ngFor in the code added above.
+      const marker = mockHost.getReferenceMarkerFor(TEST_TEMPLATE, 'ngFor');
 
-    const result = ngService.getDefinitionAndBoundSpan(TEST_TEMPLATE, marker.start);
-    expect(result).toBeDefined();
-    const {textSpan, definitions} = result !;
+      const result = ngService.getDefinitionAndBoundSpan(TEST_TEMPLATE, marker.start);
+      expect(result).toBeDefined();
+      const {textSpan, definitions} = result !;
 
-    // Get the marker for bounded text in the code added above
-    const boundedText = mockHost.getLocationMarkerFor(TEST_TEMPLATE, 'my');
-    expect(textSpan).toEqual(boundedText);
+      // Get the marker for bounded text in the code added above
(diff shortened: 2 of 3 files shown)


packages/language-service/src/locate_symbol.ts
```diff
@@ -20,6 +20,43 @@ export interface SymbolInfo {
...
+  let result: {symbol: Symbol, span: Span}|undefined;
+  const {templateBindings} = info.expressionParser.parseTemplateBindings(
+      attribute.name, attribute.value, attribute.sourceSpan.toString(),
+      attribute.valueSpan.start.offset);
+  // Find where the cursor is relative to the start of the attribute value.
+  const valueRelativePosition: number = templatePosition - attribute.valueSpan.start.offset;
+
+  // Find the symbol that contains the position.
+  templateBindings.forEach(tb => {
+    if (!tb.keyIsVar) {
```
ayazhafiz: Prefer filtering by `keyIsVar` first or bailing out if `keyIsVar` is true: `if (tb.keyIsVar) return;`, because there are a lot of nested ifs.

packages/language-service/src/locate_symbol.ts
```diff
@@ -20,6 +20,43 @@ export interface SymbolInfo {
...
+      if (tb.expression && inSpan(valueRelativePosition, tb.expression.ast.span)) {
```
ayazhafiz: Now that we are on TS 3.7, you can do

```suggestion
      if (inSpan(valueRelativePosition, tb.expression?.ast?.span)) {
```

packages/language-service/src/locate_symbol.ts
```diff
@@ -36,14 +73,24 @@ export function locateSymbol(info: AstResult, position: number): SymbolInfo|unde
...
-          const dinfo = diagnosticInfoFromTemplateInfo(info);
-          const scope = getExpressionScope(dinfo, path);
-          if (attribute.valueSpan) {
-            const result = getExpressionSymbol(scope, ast, templatePosition, info.template.query);
-            if (result) {
-              symbol = result.symbol;
-              const expressionOffset = attribute.valueSpan.start.offset;
-              span = offsetSpan(result.span, expressionOffset);
+          if (attribute.name.startsWith('*')) {
+            if (attribute.valueSpan) {
```
ayazhafiz: This check is done in `getSymbolInMicrosyntax`; please remove it in one place.

packages/language-service/src/locate_symbol.ts
```diff
@@ -36,14 +73,24 @@ export function locateSymbol(info: AstResult, position: number): SymbolInfo|unde
...
+          } else {
+            const dinfo = diagnosticInfoFromTemplateInfo(info);
+            const scope = getExpressionScope(dinfo, path);
+            if (attribute.valueSpan) {
+              const result = getExpressionSymbol(scope, ast, templatePosition, info.template.query);
+              if (result) {
+                symbol = result.symbol;
+                const expressionOffset = attribute.valueSpan.start.offset;
+                span = offsetSpan(result.span, expressionOffset);
+              }
```
ayazhafiz: Please assign to one `result` variable and perform this procedure simultaneously for both branches rather than duplicating.

packages/language-service/src/locate_symbol.ts
```diff
@@ -130,12 +177,25 @@ export function locateSymbol(info: AstResult, position: number): SymbolInfo|unde
...
             if (!attributeValueSymbol(ast.value)) {
-              symbol = findInputBinding(info, templatePosition, ast);
-              span = spanOf(ast);
+              const directive = findParentOfBinding(info.templateAst, ast, templatePosition);
+              const attribute = findAttribute(info, position);
+              if (directive && attribute) {
+                if (attribute.name.startsWith('*')) {
+                  compileTypeSummary = directive.directive;
+                  symbol = info.template.query.getTypeSymbol(compileTypeSummary.type.reference);
+                  // Use 'attribute.sourceSpan' instead of directive,
```
ayazhafiz:
```suggestion
                  // Use 'attribute.sourceSpan' instead of the directive's,
```

packages/language-service/src/locate_symbol.ts
```diff
@@ -130,12 +177,25 @@ export function locateSymbol(info: AstResult, position: number): SymbolInfo|unde
...
+                  // because the span of the directive is the whole opening tag of an element.
+                  span = spanOf(attribute.sourceSpan);
+                } else {
+                  symbol = findInputBinding(info, ast.templateName, directive);
+                  span = spanOf(ast);
+                }
+              }
```
ayazhafiz: What happens if you get rid of this and just do

```typescript
if (!directive) return;

symbol = findInputBinding(info.templateAst, ast, templatePosition);
span = spanOf(ast);
```

The logic here seems very similar to what you have written and changed for `attributeValueSymbol`. Is it necessary if we are already on a directive property?

ayazhafiz (changes requested):

packages/language-service/src/locate_symbol.ts
ivanwonder: For example `*ngFor="exp"`, the `attributeValueSymbol` only checks the `exp`, if it returns false, for the structural directive, we should get the `ngFor` directive, for attribute directives `[model]="value"`, we need the directive property `model` from `string-model`.

packages/language-service/src/locate_symbol.ts
```diff
@@ -36,15 +36,10 @@ export function locateSymbol(info: AstResult, position: number): SymbolInfo|unde
...
-              symbol = result.symbol;
-              const expressionOffset = attribute.valueSpan.start.offset;
-              span = offsetSpan(result.span, expressionOffset);
-            }
+          const result = getSymbolInAttributeValue(info, templatePosition, path, attribute);
+          if (result) {
+            symbol = result.symbol;
+            span = offsetSpan(result.span, attribute.valueSpan !.start.offset);
           }
           return true;
```
ivanwonder: rename the function `getSymbolInMicrosyntax` to `getSymbolInAttributeValue`.  the `getExpressionSymbol` have been done in `getSymbolInAttributeValue`, so remove it.

ivanwonder: If it's ok, then I can remove the arguments `ast`, `inEvent` of function `attributeValueSymbol`, they are unused now.

ayazhafiz: Yes, feel free to remove them!

packages/language-service/test/definitions_spec.ts
```diff
@@ -264,29 +264,84 @@ describe('definitions', () => {
...
-    const def = definitions ![0];
-    expect(def.fileName).toBe(refFileName);
-    expect(def.name).toBe('ngIf');
-    expect(def.kind).toBe('property');
-    // Not asserting the textSpan of definition because it's external file
+      const refFileName = '/node_modules/@angular/common/common.d.ts';
+      const def = definitions ![0];
+      expect(def.fileName).toBe(refFileName);
+      expect(def.name).toBe('NgForOf');
+      expect(def.kind).toBe('property');
```
ayazhafiz: Hm, this doesn't look right.. `NgForOf` is the directive, `ngForOf` is the property on that directive:

https://github.com/angular/angular/blob/e0ad9ecda0b8a541b405d2ab35335b90ceb21fd1/packages/common/src/directives/ng_for_of.ts#L124-L133

This is why I asked about the code change in `visitDirectiveProperty`; it looks like it is grabbing the directive name rather than the property.

packages/language-service/src/locate_symbol.ts
```diff
@@ -130,18 +125,67 @@ export function locateSymbol(info: AstResult, position: number): SymbolInfo|unde
...
         },
         null);
+
     if (symbol && span) {
       return {symbol, span: offsetSpan(span, info.template.span.start), compileTypeSummary};
     }
   }
 }
 
+function getSymbolInAttributeValue(
```
ayazhafiz: Nice! I like this function a lot.

ayazhafiz (commented):
My apologies for the delayed review. I will try to respond as soon as possible next time..

packages/language-service/test/definitions_spec.ts
ivanwonder: I think it's the bug here
https://github.com/angular/angular/blob/e0ad9ecda0b8a541b405d2ab35335b90ceb21fd1/packages/language-service/src/typescript_symbols.ts#L358
The hover info is correct because of the code here.
https://github.com/angular/angular/blob/e0ad9ecda0b8a541b405d2ab35335b90ceb21fd1/packages/language-service/src/hover.ts#L139

packages/language-service/src/locate_symbol.ts
ivanwonder: The arguments `ast`, `inEvent` have been removed.

packages/language-service/test/definitions_spec.ts
ivanwonder: > This is why I asked about the code change in `visitDirectiveProperty`; it looks like it is grabbing the directive name rather than the property.

Yes, I use the directive name here, see the hover info `(directive) CommonModule.NgForOf: class`.

ivanwonder: If you want to see the `ngForOf` property, hover over the attribute value
`<div *ngFor="let item «ᐱofᐱ heroes»; trackBy: test;"></div>`.
I have added the test for `trackBy`.

packages/language-service/src/locate_symbol.ts
```diff
@@ -129,19 +124,68 @@ export function locateSymbol(info: AstResult, position: number): SymbolInfo|unde
...
+  // Find the symbol that contains the position.
+  templateBindings.filter(tb => !tb.keyIsVar).forEach(tb => {
+    if (inSpan(valueRelativePosition, tb.expression?.ast.span)) {
+      const dinfo = diagnosticInfoFromTemplateInfo(info);
+      const scope = getExpressionScope(dinfo, path);
+      result = getExpressionSymbol(scope, tb.expression !, templatePosition, info.template.query);
+    } else if (inSpan(valueRelativePosition, tb.span)) {
+      const template = path.first(EmbeddedTemplateAst);
+      if (template) {
+        // One element can only have one template bindings.
```
ayazhafiz:
```suggestion
        // One element can only have one template binding.
```

packages/language-service/src/locate_symbol.ts
```diff
@@ -129,19 +124,68 @@ export function locateSymbol(info: AstResult, position: number): SymbolInfo|unde
...
+        const directiveAst = template.directives[0];
```
ayazhafiz: What happens if this directive does not have the matching input binding?

ayazhafiz (changes requested):
lgtm just a couple notes

packages/language-service/src/locate_symbol.ts
ivanwonder: It will be `undefined` as before. Here, firstly I will check the input value, then check the property.
for example `<div *ngFor="let item of heroes; «ᐱtrackByᐱ: test»;">`, check `test`, and then `trackBy`.
For `<div *ngIf="«include»"></div>`, it only check the input value, the `tb.span` wil be `{start:0, end:0}`. So here I only try to get the `EmbeddedTemplateAst`.

ayazhafiz (approved):
thanks!

packages/language-service/src/locate_symbol.ts
```diff
@@ -129,19 +124,68 @@ export function locateSymbol(info: AstResult, position: number): SymbolInfo|unde
...
         null);
+
     if (symbol && span) {
       return {symbol, span: offsetSpan(span, info.template.span.start), compileTypeSummary};
     }
   }
 }
 
+function getSymbolInAttributeValue(
+    info: AstResult, templatePosition: number, path: TemplateAstPath,
```
kyliau: `templatePosition = path.position`, so it is not needed as a function parameter

packages/language-service/src/locate_symbol.ts
```diff
@@ -129,19 +124,68 @@ export function locateSymbol(info: AstResult, position: number): SymbolInfo|unde
...
         },
         null);
+
     if (symbol && span) {
       return {symbol, span: offsetSpan(span, info.template.span.start), compileTypeSummary};
     }
   }
 }
 
+function getSymbolInAttributeValue(
```
kyliau: Could you please add function doc?

packages/language-service/test/definitions_spec.ts
```diff
@@ -264,29 +264,84 @@ describe('definitions', () => {
...
+      const result = ngService.getDefinitionAt(TEST_TEMPLATE, marker.start);
+      expect(result).toBeDefined();
+      const {textSpan, definitions} = result !;
+
+      // Get the marker for bounded text in the code added above
+      const boundedText = mockHost.getLocationMarkerFor(TEST_TEMPLATE, 'my');
+      expect(textSpan).toEqual(boundedText);
+
+      expect(definitions).toBeDefined();
+      expect(definitions !.length).toBe(2);
```
kyliau: Could you please add an assertion for `definitions[1]`? 
Is it the same as `definitions[0]`?
If so, could you please add a comment to explain why this is the case?

packages/language-service/src/locate_symbol.ts
```diff
@@ -193,17 +237,14 @@ function findParentOfBinding(
...
-  const directiveAst = findParentOfBinding(info.templateAst, binding, position);
-  if (directiveAst) {
-    const invertedInput = invertMap(directiveAst.directive.inputs);
-    const fieldName = invertedInput[binding.templateName];
-    if (fieldName) {
-      const classSymbol = info.template.query.getTypeSymbol(directiveAst.directive.type.reference);
-      if (classSymbol) {
-        return classSymbol.members().get(fieldName);
-      }
+function findInputBinding(info: AstResult, name: string, directiveAst: DirectiveAst): Symbol|
```
kyliau: Please add function doc.

kyliau (changes requested):
Minor changes. Thank you for adding this feature!

packages/language-service/src/locate_symbol.ts
ivanwonder: I think need the `templatePosition` to locate the symbol.

packages/language-service/test/definitions_spec.ts
ivanwonder: yes, it's the same as the `definitions[0]`. comment has added.

ayazhafiz: Can you add tests for the second definition?

ayazhafiz (changes requested):

packages/language-service/test/definitions_spec.ts
ivanwonder: The two definitions are setter and getter of 'ngForTrackBy'. @ayazhafiz

ivanwonder: I just add a comment, do we need to test the getter and setter?

ayazhafiz: Yes; could you add expectations in this test method for `definitions[1]`? You can loop over all definitions and make sure that they have the same fileName, name , and kind.

ivanwonder: OK, How about removing the `expect(definitions !.length).toBe(2);`, just make sure they have the same fileName, name, and kind.

ayazhafiz: I think the expectation that there is 2 is good, because of the comment that you have added. And we also want to verify that both of the definitions are correct (that their file and name and kind is what you expect)

ayazhafiz (approved):

packages/language-service/src/locate_symbol.ts
kyliau: Sorry, I meant to say that `templatePosition` is the same as `path.position`, so anywhere `templatePosition` is used in this function, they can be replaced with `path.position`.

ivanwonder: I don’t think the path has property position. Do you mean the path.tail?

ayazhafiz: it does:

https://github.com/angular/angular/blob/32b72f39f061773dfc455446106ad0bbcd5be88c/packages/compiler/src/ast_path.ts#L27

ivanwonder: my mistake.

ivanwonder:
@ayazhafiz I just rebase to the master. can you look at the test case error? are they the expected way? I think it's a bug for `definitions`.

ayazhafiz:
The `property` for NgForOf I think is. I have noticed the same thing with a commit I was working on today. I will take a look and fix in a few hours if that is okay. My apologies for the trouble.

ayazhafiz:
@ivanwonder I think for now, the easiest and most correct thing to do for this PR will be to overwrite the symbol found for the structural directive in `visitDirectiveProperty` to be a directive symbol, via e.g. `symbol = symbol && new OverrideKindSymbol(symbol, DirectiveKind.DIRECTIVE)`. This doesn't fix the `CommonModule.` prefix missing, but that requires a PR that I don't want to block this feature on. Let me know if that sounds good..

ivanwonder:
@ayazhafiz So, I also need to make the assertion work for the length of `definitions`?

ayazhafiz:
Yes. I do not think that is a bug.

packages/language-service/src/locate_symbol.ts
```diff
@@ -38,6 +40,11 @@ export function locateSymbols(info: AstResult, position: number): SymbolInfo[] {
     toVisit.push(node);
   }
 
+  // For the structural directive, only care about the last template AST.
+  if (attribute?.name.startsWith('*')) {
+    toVisit.splice(0, toVisit.length - 1);
+  }
```
ivanwonder: For the structural directive, the 'path' will contain multiple `BoundDirectivePropertyAst` which depends on the number of directive property in the attribute value(e.g. '*ngFor="let item of []; trackBy: test;"', it has 2 `BoundDirectivePropertyAst`, 'ngForOf' and 'ngForTrackBy').

ivanwonder: @ayazhafiz I think this is why the length of `definitions` has doubled.

ivanwonder: You can review my PR again. If it’s ok, can you approve it?
I test it in vscode, the number of definitions show in the tooltip doesn’t ignore the duplication.

ivanwonder: image
https://github.com/angular/angular/blob/3d4b1431ecbf6e6d6605fea08016baf905fc3299/packages/language-service/test/definitions_spec.ts#L66
When you click it, it will show only one definition.

ivanwonder: Maybe we should filter the `definitions` here.
https://github.com/angular/angular/blob/3d4b1431ecbf6e6d6605fea08016baf905fc3299/packages/language-service/src/definitions.ts#L68-L71

ayazhafiz: Yes, we will filter the definitions. The PR looks good to me. I think for now this is fine, will fixup after the definition-pruning PR.

ayazhafiz (approved):

packages/language-service/src/locate_symbol.ts
kyliau: There is a `uniqueBySpan` function in `diagnostics.ts` that could be useful here.

kyliau (approved):

PR #34855, opened 2020-01-19
Update hero.service.ts

LuisEscamillaMartin:
Forgotten semicolon.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/aio/content/examples/toh-pt6/src/app/hero.service.ts
+++ b/aio/content/examples/toh-pt6/src/app/hero.service.ts
@@ -37,7 +37,7 @@ export class HeroService {
   /** GET heroes from the server */
   // #docregion getHeroes-2
   getHeroes (): Observable<Hero[]> {
-    return this.http.get<Hero[]>(this.heroesUrl)
+    return this.http.get<Hero[]>(this.heroesUrl);
   // #enddocregion getHeroes-1
       .pipe(
         // #enddocregion getHeroes-2


ericmartinezr:
I wouldn't say this is correct even though it is. In this [snippet](https://angular.io/tutorial/toh-pt6#get-heroes-with-httpclient) (note that is not necessary, see this [answer](https://stackoverflow.com/a/4002248/4933038)). 

Also that snippet is extended later on [here](https://angular.io/tutorial/toh-pt6#error-handling) and [here](https://angular.io/tutorial/toh-pt6#tap-into-the-observable) where those sentences are terminated with a semicolon.

Airblader:
This change is incorrect and will cause a syntax error. There's a comment in between, but look at the second line after the return line.

petebacondarwin:
I agree. We cannot break the actual code in the examples in order to make the code snippet look right in the docs. We must find another way...

gkalpak:
IMO, we should leave it as is. Missing an (optional) semicolon in an intermediate code snippet is better than maintaining two copies of the same file (having to keep them in-sync, not being able to test both, etc.).

petebacondarwin:
How about a compromise using a commented out block right next to the working code?

```ts
  // #docregion getHeroes
  /** GET heroes from the server */
  // #docregion getHeroes-2
  getHeroes(): Observable<Hero[]> {
    return this.http.get<Hero[]>(this.heroesUrl)
      .pipe(
        // #enddocregion getHeroes-2
        tap(_ => this.log('fetched heroes')),
        // #docregion getHeroes-2
        catchError(this.handleError<Hero[]>('getHeroes', []))
      );
  }
  // #enddocregion getHeroes, getHeroes-2
  // #enddocregion
  /*
    THIS COMMENT BLOCK IS TO HOLD THE getHeroes-1 docregion
  // #docregion getHeroes-1
    getHeroes(): Observable<Hero[]> {
      return this.http.get<Hero[]>(this.heroesUrl);
    }
  // #enddocregion getHeroes-1
    END OF COMMENT BLOCK
  */
  // #docregion
```

gkalpak:
Not worth it imo, but not a hill I'll die on :wink:

petebacondarwin:
Nor me 😁 
Perhaps we can let the docs-team (e.g. @kapunahelewong) take view.

kapunahelewong:
Thank you for this, @LuisEscamillaMartin. Though it did appear to be a simple fix, as others point out, that snippet is pulled from the larger file and has to compile, so we can't put the semicolon there. While you could create a separate file or add a comment block as @petebacondarwin describes, yet another option is to edit the header of that snippet to say:

```
src/app/hero.service.ts (excerpt)
```

That would be in [the md file](https://github.com/angular/angular/blob/master/aio/content/tutorial/toh-pt6.md) and you'd change the `header` like this:

```
<code-example path="toh-pt6/src/app/hero.service.ts" header="src/app/hero.service.ts (excerpt)" region="getHeroes-1">
</code-example>
```

So, you could leave it as is, or try one of the options we've shared here. What are your thoughts?

petebacondarwin (approved):

petebacondarwin:
Closing as the PR is stale, no consensus was found on how to resolve the problem, and the OP did not sign the CLA.

PR #34870, opened 2020-01-20
fix(ngcc): update `package.json` deterministically

### LINKED ISSUE #34635: inconsistent __processed_by_ivy_ngcc__ order breaks caching
# 🐞 bug report

### Affected Package
`@angular/*`, `@angular/bazel` users especially

### Is this a regression?

Yes, in ivy at least.

### Description

#32427 made the `ngcc` tasks run in parallel (FYI @gkalpak 👋). I assume this is what causes the `__processed_by_ivy_ngcc__` entries in `package.json` to be defined in a random order, causing inconsistent `package.json` content. Tools such as bazel which use file hashes for caching purposes will then get cache misses (almost) every time `ngcc` runs on a package.

## 🔬 Minimal Reproduction

Run `yarn install` a couple times and observe the `package.json` files being inconsistent.

If using bazel you can use `--execution_log_json_file=executions.log` on a clean build a few times and see the difference, the hashes of many `@angular/*` `package.json` files changes.

**Angular Version:**

9.0.0-rc.7, I think it also existed back to rc.0 and earlier.

gkalpak:
Ngcc adds properties to the `package.json` files of the entry-points it processes to mark them as processed for a format and point to the created Ivy entry-points (in case of `--create-ivy-entry-points`). When running ngcc in parallel mode (which is the default for the standalone ngcc command), multiple formats can be processed simultaneously for the same entry-point and the order of completion is not deterministic.

Previously, ngcc would append new properties at the end of the target object in `package.json` as soon as the format processing was completed. As a result, the order of properties in the resulting `package.json` (when processing multiple formats for an entry-point in parallel) was not deterministic. For tools that use file hashes for caching purposes (such as Bazel), this lead to a high probability of cache misses.

This commit fixes the problem by ensuring that the position of properties added to `package.json` files is deterministic and independent of the order in which each format is processed.

Jira issue: [FW-1801](https://angular-team.atlassian.net/browse/FW-1801)

Fixes #34635.

### DIFF
--- a/packages/compiler-cli/ngcc/src/writing/package_json_updater.ts
+++ b/packages/compiler-cli/ngcc/src/writing/package_json_updater.ts
@@ -10,7 +10,8 @@ import {AbsoluteFsPath, FileSystem, dirname} from '../../../src/ngtsc/file_syste
 import {JsonObject, JsonValue} from '../packages/entry_point';
 
 
-export type PackageJsonChange = [string[], JsonValue];
+export type PackageJsonChange = [string[], JsonValue, PackageJsonPropertyPositioning];
+export type PackageJsonPropertyPositioning = 'unimportant' | 'alphabetic' | {before: string};
 export type WritePackageJsonChangesFn =
     (changes: PackageJsonChange[], packageJsonPath: AbsoluteFsPath, parsedJson?: JsonObject) =>
         void;
@@ -23,8 +24,9 @@ export type WritePackageJsonChangesFn =
  * const updatePackageJson = packageJsonUpdater
  *     .createUpdate()
  *     .addChange(['name'], 'package-foo')
- *     .addChange(['scripts', 'foo'], 'echo FOOOO...')
- *     .addChange(['dependencies', 'bar'], '1.0.0')
+ *     .addChange(['scripts', 'foo'], 'echo FOOOO...', 'unimportant')
+ *     .addChange(['dependencies', 'baz'], '1.0.0', 'alphabetic')
+ *     .addChange(['dependencies', 'bar'], '2.0.0', {before: 'baz'})
  *     .writeChanges('/foo/package.json');
  *     // or
  *     // .writeChanges('/foo/package.json', inMemoryParsedJson);
@@ -33,13 +35,13 @@ export type WritePackageJsonChangesFn =
 export interface PackageJsonUpdater {
   /**
    * Create a `PackageJsonUpdate` object, which provides a fluent API for batching updates to a
-   * `package.json` file. (Batching the updates is useful, because it avoid unnecessary I/O
+   * `package.json` file. (Batching the updates is useful, because it avoids unnecessary I/O
    * operations.)
    */
   createUpdate(): PackageJsonUpdate;
 
   /**
-   * Write a set of changes to the specified `package.json` file and (and optionally a pre-existing,
+   * Write a set of changes to the specified `package.json` file (and optionally a pre-existing,
    * in-memory representation of it).
    *
    * @param changes The set of changes to apply.
@@ -65,15 +67,27 @@ export class PackageJsonUpdate {
   constructor(private writeChangesImpl: WritePackageJsonChangesFn) {}
 
   /**
-   * Record a change to a `package.json` property. If the ancestor objects do not yet exist in the
-   * `package.json` file, they will be created.
+   * Record a change to a `package.json` property.
    *
-   * @param propertyPath The path of a (possibly nested) property to update.
+   * If the ancestor objects do not yet exist in the `package.json` file, they will be created. The
+   * positioning of the property can also be specified. (If the property already exists, it will be
+   * moved accordingly.)
+   *
+   * NOTE: Property positioning is only guaranteed to be respected in the serialized `package.json`
+   *       file. Positioning will not be taken into account when updating in-memory representations.
+   *
+   * NOTE 2: Property positioning only affects the last property in `propertyPath`. Ancestor
+   *         objects' positioning will not be affected.
+   *
+   * @param propertyPath The path of a (possibly nested) property to add/update.
    * @param value The new value to set the property to.
+   * @param position The desired position for the added/updated property.
    */
-  addChange(propertyPath: string[], value: JsonValue): this {
+  addChange(
+      propertyPath: string[], value: JsonValue,
+      positioning: PackageJsonPropertyPositioning = 'unimportant'): this {
     this.ensureNotApplied();
-    this.changes.push([propertyPath, value]);
+    this.changes.push([propertyPath, value, positioning]);
     return this;
   }
 
@@ -121,15 +135,16 @@ export class DirectPackageJsonUpdater implements PackageJsonUpdater {
 
     // Apply all changes to both the canonical representation (read from disk) and any pre-existing,
     // in-memory representation.
-    for (const [propPath, value] of changes) {
+    for (const [propPath, value, positioning] of changes) {
       if (propPath.length === 0) {
         throw new Error(`Missing property path for writing value to '${packageJsonPath}'.`);
       }
 
-      applyChange(parsedJson, propPath, value);
+      applyChange(parsedJson, propPath, value, positioning);
 
       if (preExistingParsedJson) {
-        applyChange(preExistingParsedJson, propPath, value);
+        // No need to take property positioning into account for in-memory representations.
+        applyChange(preExistingParsedJson, propPath, value, 'unimportant');
       }
     }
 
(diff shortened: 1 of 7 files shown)


packages/compiler-cli/ngcc/src/writing/package_json_updater.ts
```diff
@@ -157,4 +174,42 @@ export function applyChange(ctx: JsonObject, propPath: string[], value: JsonValu
   }
 
   ctx[lastProp] = value;
+  positionProperty(ctx, lastProp, positioning);
```
petebacondarwin: _Functional programming: for the win!_

Instead of using "enumeration" and a switch statement you could do this by creating instances of `isNextProp` function instead of `positioning`. I would envisage something like:

```
const unimportant = (propToInsert: string, currentProp: string) => boolean = false;
const alphabetic = (propToInsert: string, currentProp: string) => boolean = currentProp > propToInsert;
const createBefore = (beforeProp: string) => (_propToInsert: string, currentProp: string) => currentProp === beforeProp;
```

This would also avoid the awkward union of string and object.

packages/compiler-cli/ngcc/test/execution/cluster/package_json_updater_spec.ts
```diff
@@ -88,21 +104,31 @@ runInEachFileSystem(() => {
         it('should send an `update-package-json` message to the master process', () => {
```
petebacondarwin: Should there be an equivalent check to 

```ts
afterEach(() => expect(processSendSpy).not.toHaveBeenCalled());
```

which is in the master version of this code. But in this case we should check that the `writeChangesSpy` was not called, right?

packages/compiler-cli/ngcc/test/integration/ngcc_spec.ts
```diff
@@ -759,6 +759,67 @@ runInEachFileSystem(() => {
...
+           expect(pkgKeys).toContain('|esm5_ivy_ngcc|esm5|');
+           expect(pkgKeys).toContain('|fesm2015_ivy_ngcc|fesm2015|');
+           expect(pkgKeys).toContain('|fesm5_ivy_ngcc|fesm5|');
+           expect(stringifyKeys(pkg.__processed_by_ivy_ngcc__ !))
+               .toMatch(/\|esm5\|(?:.+\|)?fesm2015\|(?:.+\|)?fesm5\|/);
+
+           // Helpers
+           function expectNotToHaveProp(obj: object, prop: string) {
+             expect(obj.hasOwnProperty(prop))
+                 .toBe(false, `Expected object not to have property '${prop}'.`);
```
petebacondarwin: Is it worth stringifying the object in the error message too?

packages/compiler-cli/ngcc/test/integration/ngcc_spec.ts
```diff
@@ -759,6 +759,67 @@ runInEachFileSystem(() => {
...
+           expectToHaveProp(pkg, 'fesm5_ivy_ngcc');
+           expectToHaveProp(pkg.__processed_by_ivy_ngcc__ !, 'esm5');
+
+           // Ensure the properties are in deterministic order (regardless of processing order).
+           const pkgKeys = stringifyKeys(pkg);
+           expect(pkgKeys).toContain('|esm5_ivy_ngcc|esm5|');
+           expect(pkgKeys).toContain('|fesm2015_ivy_ngcc|fesm2015|');
+           expect(pkgKeys).toContain('|fesm5_ivy_ngcc|fesm5|');
+           expect(stringifyKeys(pkg.__processed_by_ivy_ngcc__ !))
+               .toMatch(/\|esm5\|(?:.+\|)?fesm2015\|(?:.+\|)?fesm5\|/);
```
petebacondarwin: Why is it not possible to simply check the array of keys directly here? What pre-existing properties might exist on this?
This regex matching seems overkill, no?

packages/compiler-cli/ngcc/test/writing/package_json_updater_spec.ts
```diff
@@ -151,5 +152,137 @@ runInEachFileSystem(() => {
...
+               .writeChanges(jsonPath);
+
+           expectJsonEquals(jsonPath, {
+             p2: '2',
+             p0: '0-added',
+             p1: {p10: '1.0-added', p12: '1.2-updated', p11: '1.1'},
+           });
+
+           // Verify that trying to add before non-existent property, puts updated property at the
+           // end.
```
petebacondarwin: 💯

petebacondarwin (changes requested):
Nice work @gkalpak - it was not a trivial fix after all.

I would be much happier (unless you can convince me otherwise) if there was no switch statement.

packages/compiler-cli/ngcc/test/integration/ngcc_spec.ts
gkalpak: Because processing one format will also marked all other formats pointing to the same entry-point as processed. For example, the `fesm5` property may point to the same file as the `module` property and thus both properties will be marked as processed once either of them gets processed. Same for `fesm2015` and `es2015`. Also, `typings` gets processed the first time we process a property.

So, after having explicitly processed `esm5`, `fesm2015` and `fesm5`, the `__processed_by_ivy_ngcc__` object will also contain `es2015`, `module` and `typings`.

packages/compiler-cli/ngcc/test/execution/cluster/package_json_updater_spec.ts
gkalpak: Good idea :+1:

packages/compiler-cli/ngcc/src/writing/package_json_updater.ts
gkalpak: I am not sure I 100% understand what you are suggesting :grin:

The problem (if I understand correctly) is that for `unimportant` I want to leave existing properties at their old positions (to avoid unnecessary changes to files). This can't be coded as an `isNextProp()` function. Also, in the future there might be different positionings, such as `{after: <prop>}`, which again cannot be coded as an `isNextProp()` function.

We could come up with ways to work around these issues, but I don't think there will be much benefit in doing it (as it will rather complicate the code than simplify it).

In general, I think it is just a coincidence that two (out of three) positioning options (`alphabetic` and `{before: ...}`) can be implemented with similar logic (`movePropBefore()`).
<sub>
_**Fun fact:**_
_Original, I specified positioning values as `unimportant`, `alphabetic` and `{after: <prop>}`, but switched to `{before: <prop>}` (instead of `after`), because it allowed the implementations to share more code and I didn't really care as long as the new prop was next to the old prop_ :grin:
</sub>

##
There is always the case that I might be missing something, so feel free to suggest an implementation.

packages/compiler-cli/ngcc/test/integration/ngcc_spec.ts
gkalpak: Done.

packages/compiler-cli/ngcc/test/integration/ngcc_spec.ts
petebacondarwin: We do know exactly what will be in the array for this test right?
I guess it is a trade off between complexity of the test vs brittleness of the test.

* In this case the test is harder to read and reason about.
* In the other case the test is more prone to failing if something about the other properties that are added changes.

But I would argue that we need to ensure that "all" properties added to the `__processed_by_ivy_ngcc__` are deterministic and so I don't think it would be brittle to check all fo them.

Perhaps I am missing something...

gkalpak: Yes, we know which properties will be there and we can check with a string match.

> In this case the test is harder to read and reason about.

Well...that is subjective 😛
It is a little unexpected why there are other props there, so we would have to have a comment explaining what is going on and some people claim that reading a simple RegExp would be more straight forward 😛

I don't think it makes a big difference anyway, so happy to switch to exact string match 👍

packages/compiler-cli/ngcc/src/writing/package_json_updater.ts
petebacondarwin: This is the basic idea: https://github.com/angular/angular/commit/6891ddeb48605965a4f7e4576c58dff1c6d7441e

petebacondarwin: I missed a few places that would also need updating but you get the general gist :-)

gkalpak: Oh, I see what you mean now. Well, the thing is that in parallel mode the recorded changes are collected on the worker side, serialized to JSON and send over to the cluster master which applies them to the actual file. Therefore, they need to be JSON-serializable (and functions are not (in a reasonable way :wink:)).

petebacondarwin: I see. That is a shame. Leave it as it is then.

Although, I could envisage changing the messaging between worker and master so that the message is abstracted more to the update task rather than needing to know about the JSON structure. I.E. where the property is placed could be owned by the cluster master and the messages could be more domain oriented.

gkalpak: I am not sure I agree, but that's a discussion for another day :smiley:

packages/compiler-cli/ngcc/test/integration/ngcc_spec.ts
gkalpak: Done.

petebacondarwin (approved):

PR #34889, opened 2020-01-21
fix(ngcc): switch ngcc over to use the `TraitCompiler`

### LINKED ISSUE #34500: ngcc: packages that fail with errors causes a crash

JoostK:
fix(ngcc): do not attempt compilation when analysis fails

In #34288, ngtsc was refactored to separate the result of the analysis
and resolve phase for more granular incremental rebuilds. In this model,
any errors in one phase transition the trait into an error state, which
prevents it from being ran through subsequent phases. The ngcc compiler
on the other hand did not adopt this strict error model, which would
cause incomplete metadata—due to errors in earlier phases—to be offered
for compilation that could result in a hard crash.

This commit updates ngcc to take advantage of ngtsc's `TraitCompiler`,
that internally manages all Ivy classes that are part of the
compilation. This effectively replaces ngcc's own `AnalyzedFile` and
`AnalyzedClass` types, together with all of the logic to drive the
`DecoratorHandler`s. All of this is now handled in the `TraitCompiler`,
benefiting from its explicit state transitions of `Trait`s so that the
ngcc crash is a thing of the past.

Fixes #34500
Resolves FW-1788

### DIFF
--- a/packages/compiler-cli/ngcc/src/analysis/decoration_analyzer.ts
+++ b/packages/compiler-cli/ngcc/src/analysis/decoration_analyzer.ts
@@ -132,11 +132,9 @@ export class DecorationAnalyzer {
    */
   analyzeProgram(): DecorationAnalyses {
     for (const sourceFile of this.program.getSourceFiles()) {
-      if (sourceFile.isDeclarationFile || !isWithinPackage(this.packagePath, sourceFile)) {
-        continue;
+      if (!sourceFile.isDeclarationFile && isWithinPackage(this.packagePath, sourceFile)) {
+        this.compiler.analyzeFile(sourceFile);
       }
-
-      this.compiler.analyzeFile(sourceFile);
     }
 
     this.applyMigrations();
--- a/packages/compiler-cli/ngcc/src/analysis/decoration_analyzer.ts
+++ b/packages/compiler-cli/ngcc/src/analysis/decoration_analyzer.ts
@@ -15,20 +15,19 @@ import {FileSystem, LogicalFileSystem, absoluteFrom, dirname, resolve} from '../
 import {AbsoluteModuleStrategy, LocalIdentifierStrategy, LogicalProjectStrategy, ModuleResolver, NOOP_DEFAULT_IMPORT_RECORDER, PrivateExportAliasingHost, Reexport, ReferenceEmitter} from '../../../src/ngtsc/imports';
 import {CompoundMetadataReader, CompoundMetadataRegistry, DtsMetadataReader, InjectableClassRegistry, LocalMetadataRegistry} from '../../../src/ngtsc/metadata';
 import {PartialEvaluator} from '../../../src/ngtsc/partial_evaluator';
-import {ClassDeclaration} from '../../../src/ngtsc/reflection';
 import {LocalModuleScopeRegistry, MetadataDtsModuleScopeResolver} from '../../../src/ngtsc/scope';
-import {CompileResult, DecoratorHandler} from '../../../src/ngtsc/transform';
-import {NgccClassSymbol, NgccReflectionHost} from '../host/ngcc_host';
+import {DecoratorHandler} from '../../../src/ngtsc/transform';
+import {NgccReflectionHost} from '../host/ngcc_host';
 import {Migration} from '../migrations/migration';
 import {MissingInjectableMigration} from '../migrations/missing_injectable_migration';
 import {UndecoratedChildMigration} from '../migrations/undecorated_child_migration';
 import {UndecoratedParentMigration} from '../migrations/undecorated_parent_migration';
 import {EntryPointBundle} from '../packages/entry_point_bundle';
-import {isDefined} from '../utils';
 
 import {DefaultMigrationHost} from './migration_host';
-import {AnalyzedClass, AnalyzedFile, CompiledClass, CompiledFile, DecorationAnalyses} from './types';
-import {NOOP_DEPENDENCY_TRACKER, analyzeDecorators, isWithinPackage} from './util';
+import {NgccTraitCompiler} from './ngcc_trait_compiler';
+import {CompiledClass, CompiledFile, DecorationAnalyses} from './types';
+import {NOOP_DEPENDENCY_TRACKER, isWithinPackage} from './util';
 
 
 
@@ -57,10 +56,6 @@ export class DecorationAnalyzer {
   private packagePath = this.bundle.entryPoint.package;
   private isCore = this.bundle.isCore;
 
-  /**
-   * Map of NgModule declarations to the re-exports for that NgModule.
-   */
-  private reexportMap = new Map<ts.Declaration, Map<string, [string, string]>>();
   moduleResolver =
       new ModuleResolver(this.program, this.options, this.host, /* moduleResolutionCache */ null);
   resourceManager = new NgccResourceLoader(this.fs);
@@ -118,6 +113,7 @@ export class DecorationAnalyzer {
         /* factoryTracker */ null, NOOP_DEFAULT_IMPORT_RECORDER,
         /* annotateForClosureCompiler */ false, this.injectableRegistry),
   ];
+  compiler = new NgccTraitCompiler(this.handlers, this.reflectionHost);
   migrations: Migration[] = [
     new UndecoratedParentMigration(),
     new UndecoratedChildMigration(),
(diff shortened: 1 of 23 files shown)


JoostK:
This PR is a replacement for #34652, with a different approach. Instead of making use of the `Trait`  types from within ngcc, this change goes one step further to drive the compilation using the `TraitCompiler`. The benefit of this is that the protocol around the calling of `DecoratorHandlers` isn't duplicated between ngtsc and ngcc, with the drawback of introducing a larger diff with more risk involved.

alxhub (approved):

packages/compiler-cli/ngcc/src/analysis/ngcc_trait_compiler.ts
```diff
@@ -0,0 +1,63 @@
...
+  analyzeFile(sf: ts.SourceFile): void {
+    const ngccClassSymbols = this.ngccReflector.findClassSymbols(sf);
+    for (const classSymbol of ngccClassSymbols) {
+      this.analyzeClass(classSymbol.declaration.valueDeclaration, null);
+    }
+
+    return undefined;
+  }
+
+  injectSyntheticDecorator(clazz: ClassDeclaration, decorator: Decorator, flags?: HandlerFlags):
```
petebacondarwin: Can you add a jsdoc description of this method since, unlike the `DefaultMigrationHost` it cannot fallback on the decription in the base `MigrationHost` interface. Copying the one from that interface would be fine by me (although that is also missing explanation of the `flags` param!):

```
  /**
   * Associate a new synthesized decorator, which did not appear in the original source, with a
   * given class.
   * @param clazz the class to receive the new decorator.
   * @param decorator the decorator to inject.
   */
   ```

File: packages/compiler-cli/ngcc/src/rendering/dts_renderer.ts [hunk 2]
```diff
@@ -135,9 +136,11 @@ export class DtsRenderer {
...
-              dtsFile.fileName.replace(/\.d\.ts$/, '.js')) {
-            renderInfo.reexports.push(...compiledClass.reexports);
+          if (!appliedReexports &&
+              compiledClass.declaration.getSourceFile().fileName ===
+                  dtsFile.fileName.replace(/\.d\.ts$/, '.js')) {
+            renderInfo.reexports.push(...compiledFile.reexports);
+            appliedReexports = true;
           }
           dtsMap.set(dtsFile, renderInfo);
         }
```
petebacondarwin: Consider moving the adding of re-exports outside of the loop:

```ts
decorationAnalyses.forEach(compiledFile => {
  let addReexports = false;
  compiledFile.compiledClasses.forEach(compiledClass => {
    const dtsDeclaration = this.host.getDtsDeclaration(compiledClass.declaration);
    if (dtsDeclaration) {
      const dtsFile = dtsDeclaration.getSourceFile();
      const renderInfo = dtsMap.has(dtsFile) ? dtsMap.get(dtsFile) ! : new DtsRenderInfo();
      renderInfo.classInfo.push({dtsDeclaration, compilation: compiledClass.compilation});
      dtsMap.set(dtsFile, renderInfo);
      // Only add re-exports if the .d.ts tree is overlayed with the .js tree, as re-exports in
      // ngcc are only used to support deep imports into e.g. commonjs code. For a deep import
      // to work, the typing file and JS file must be in parallel trees. This logic will detect
      // the simplest version of this case, which is sufficient to handle most commonjs
      // libraries.
      if (compiledClass.declaration.getSourceFile().fileName === dtsFile.fileName.replace(/\.d\.ts$/, '.js')) {
        addReexports = true;
      }
    }
  });
  if (addReexports) {
    renderInfo.reexports.push(...compiledFile.reexports);
  }
});
```

packages/compiler-cli/ngcc/test/analysis/decoration_analyzer_spec.ts
```diff
@@ -117,7 +119,7 @@ runInEachFileSystem(() => {
             getFileSystem(), bundle, reflectionHost, referencesRegistry,
             (error) => diagnosticLogs.push(error));
         testHandler = createTestHandler(options);
-        analyzer.handlers = [testHandler];
+        analyzer.handlers.splice(0, analyzer.handlers.length, testHandler);
```
petebacondarwin: I know that splice is probably more efficient but since this is just a test, perhaps 

```ts
analyzer.handlers = [...analyzer.handlers, testHandler];
```

is a bit more reasonable?

OOC Can you explain why we cannot just wipe out the standard handlers for this test?

packages/compiler-cli/ngcc/test/analysis/decoration_analyzer_spec.ts
```diff
@@ -410,7 +436,7 @@ runInEachFileSystem(() => {
             name: _('/node_modules/test-package/index.d.ts'),
             contents: 'export declare class SomeDirective {}',
           }]);
-          analyzer.handlers = [new FakeDecoratorHandler()];
+          analyzer.handlers.splice(0, analyzer.handlers.length, new FakeDecoratorHandler());
```
petebacondarwin:
```ts
analyzer.handlers = [...analyzer.handlers.length, new FakeDecoratorHandler()];
```

packages/compiler-cli/ngcc/test/analysis/migration_host_spec.ts
```diff
@@ -5,62 +5,67 @@
...
-import {NgccClassSymbol} from '../../src/host/ngcc_host';
+import {NgccTraitCompiler} from '../../src/analysis/ngcc_trait_compiler';
+import {Esm2015ReflectionHost} from '../../src/host/esm2015_host';
 import {createComponentDecorator} from '../../src/migrations/utils';
+import {EntryPointBundle} from '../../src/packages/entry_point_bundle';
+import {MockLogger} from '../helpers/mock_logger';
+import {makeTestEntryPointBundle} from '../helpers/utils';
 
 runInEachFileSystem(() => {
   describe('DefaultMigrationHost', () => {
```
petebacondarwin: It seems to me that most of this spec file is actually testing the `NgccTraitCompiler`. The `DefaultMigrationHost` mostly just delegates to that.

How about simplifying this spec just to check that the delegation occurs as expected, and then create a spec for `NgccTraitCompiler` that exercises the tests here?

packages/compiler-cli/src/ngtsc/transform/src/api.ts
```diff
@@ -155,6 +155,7 @@ export interface DecoratorHandler<D, A, R> {
 
 export interface DetectResult<M> {
   trigger: ts.Node|null;
+  decorator: Decorator|null;
   metadata: Readonly<M>;
 }
```
petebacondarwin: I know this had none before, but some jsdoc descriptions of this interface and its fields would be great!

packages/compiler-cli/src/ngtsc/transform/src/compilation.ts
```diff
@@ -6,24 +6,23 @@
...
 import {TypeCheckContext} from '../../typecheck';
 import {getSourceFile, isExported} from '../../util/src/typescript';
 
-import {AnalysisOutput, CompileResult, DecoratorHandler, DetectResult, HandlerPrecedence, ResolveResult} from './api';
+import {AnalysisOutput, CompileResult, DecoratorHandler, HandlerFlags, HandlerPrecedence, ResolveResult} from './api';
 import {DtsTransformRegistry} from './declaration';
 import {Trait, TraitState} from './trait';
 
 
+
```
petebacondarwin: Unecessary extra blank line

packages/compiler-cli/src/ngtsc/transform/src/compilation.ts
```diff
@@ -123,7 +122,7 @@ export class TraitCompiler {
     }
 
     const visit = (node: ts.Node): void => {
-      if (isNamedClassDeclaration(node)) {
+      if (this.reflector.isClass(node)) {
```
petebacondarwin: :100:

packages/compiler-cli/src/ngtsc/transform/src/compilation.ts
```diff
@@ -425,7 +441,7 @@ export class TraitCompiler {
 
   compile(clazz: ts.Declaration, constantPool: ConstantPool): CompileResult[]|null {
     const original = ts.getOriginalNode(clazz) as typeof clazz;
-    if (!isNamedClassDeclaration(clazz) || !isNamedClassDeclaration(original) ||
+    if (!this.reflector.isClass(clazz) || !this.reflector.isClass(original) ||
```
petebacondarwin: :+1:

petebacondarwin (changes requested):
Nice refactoring! It is great to see so much code removed from ngcc.
Can you add the `NgccTraitCompiler` spec and simplify the `DefaultMigrationHost` spec as mentioned below.
Otherwise LGTM!

packages/compiler-cli/ngcc/src/analysis/decoration_analyzer.ts
```diff
@@ -135,56 +131,56 @@ export class DecorationAnalyzer {
...
-                              .filter(sourceFile => isWithinPackage(this.packagePath, sourceFile))
-                              .map(sourceFile => this.analyzeFile(sourceFile))
-                              .filter(isDefined);
+    for (const sourceFile of this.program.getSourceFiles()) {
+      if (sourceFile.isDeclarationFile || !isWithinPackage(this.packagePath, sourceFile)) {
+        continue;
+      }
+
+      this.compiler.analyzeFile(sourceFile);
+    }
```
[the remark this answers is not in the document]
JoostK: In this case I don't really mind either way. I updated as suggested.

packages/compiler-cli/ngcc/src/analysis/decoration_analyzer.ts
```diff
@@ -194,67 +190,45 @@ export class DecorationAnalyzer {
...
-    const reexports: Reexport[] = [];
-    if (this.reexportMap.has(declaration)) {
-      this.reexportMap.get(declaration) !.forEach(([fromModule, symbolName], asAlias) => {
-        reexports.push({asAlias, fromModule, symbolName});
+      compiledClasses.push({
+        name: record.node.name.text,
+        decorators: this.compiler.getAllDecorators(record.node),
+        declaration: record.node, compilation
       });
     }
```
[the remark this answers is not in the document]
JoostK: Here I do actually prefer the early `continue`, as I usually deal with "edge-cases" with an early control flow statement to keep the main body of work at the top.

packages/compiler-cli/ngcc/src/analysis/ngcc_trait_compiler.ts
JoostK: Added, including `flags` in both places.

packages/compiler-cli/ngcc/src/rendering/dts_renderer.ts
JoostK: That is indeed much nicer but when I applied this change I noticed that `renderInfo` is only available in the loop. I tried to introduce

```ts
      let dtsRenderInfo: DtsRenderInfo|null = null;
```

instead of `addReexports` to capture the render info, but then you'd need to workaround a TS limitation where it doesn't recognize that `dtsRenderInfo` could have been reassigned within the closure, so it would narrow `dtsRenderInfo` to `never` after a `dtsRenderInfo !== null`. At that point I don't think the readability is better than what it currently is, so I kept it unchanged.

packages/compiler-cli/ngcc/test/analysis/decoration_analyzer_spec.ts
JoostK: Oooh, so this is actually trickier than it looks (and perhaps even trickier than it should be, honestly). The `analyze.handlers` is passed into the `NgccTraitCompiler` and is shared through the array reference. Replacing the array would therefore not make the new handlers available to the trait compiler.

If I just append/prepend the handler some tests start to fail, hence the `slice`.

petebacondarwin: Oh! I forgot that the second parameter is the `delete`, which means that my suggestion is incorrect.
Perhaps a comment in the code to stop people like me from worrying about it?

packages/compiler-cli/src/ngtsc/transform/src/api.ts
JoostK: Added some

packages/compiler-cli/ngcc/test/analysis/migration_host_spec.ts
JoostK: I'm really not fond of testing just delegation behaviour, as doing so starts to be testing an implementation rather than some desired behaviour. I do see however where you're coming from here, and have moved most of the `injectSyntheticDecorator` and `getAllDecorators` tests to an `NgccTraitCompiler` spec file, while retaining some in the `DefaultMigrationHost` specs to test the logic in the context of the migration host.

JoostK:
This is green on CI: angular/ngcc-validation#792

@petebacondarwin PTAL

packages/compiler-cli/ngcc/src/analysis/decoration_analyzer.ts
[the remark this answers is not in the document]
petebacondarwin: As you like 😄

petebacondarwin (approved):
Grand!

JoostK:
merge-assistence: Alex has approved on behalf of fw-compiler when it was still draft.

AndrewKushnir:
[Presubmit](http://test/OCL:291236444:BASE:291236504:1579816192037:1b275da2)

PR #34890, opened 2020-01-22
build: add --stamp flag to the release config

josephperrott:
In a recent upgrade of rules_nodejs an update was made the
stamping logic in npm_package.  Triggering stamping for npm_package
now require the --stamp flag.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -43,6 +43,7 @@ test --incompatible_strict_action_env
 # This command assumes node on the path and is a workaround for
 # https://github.com/bazelbuild/bazel/issues/4802
 build:release --workspace_status_command="node ./tools/bazel_stamp_vars.js"
+build:release --stamp
 
 ###############################
 # Output                      #


gregmagolan (approved):
Ahh. Right. That was it. I thought we had landed that change already in this repo as we had released it before 1.0 but I believe but this repo was now a few versions behind.
Thanks for tracking it down!

gregmagolan:
Guess we should add a test for this :)

kyliau (approved):

devversion (approved):
LGTM. I agree that we should check for this. Worth mentioning: In the past, I added a version validation to avoid `a.b.c-local-changes` versions. We can expand this to error if `0.0.0(-PLACEHOLDER)?` matches?

https://github.com/angular/angular/blob/master/scripts/release/pre-check

IgorMinar (approved):
thanks!

PR #34896, opened 2020-01-22
fix(common): remove extra & in http params

cexbrayat:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

HTTP params like
```typescript
{a: '', b: [], c: '3'}
```
result in a request with `a=&&c=3` (note the extra `&` introduced by `b` and its empty array)

Issue Number: couldn't find one

## What is the new behavior?

There are two ways to fix this:
- be coherent and generate `a=&b=&c=3`. This would be the best but it is a breaking change as ome people may rely on the current behavior of not having the array parameter sent if it is empty.
- be compatible and just remove the extra & in the URL `a=&c=3`

After discussing with @petebacondarwin we went for the simple and compatible fix.

- [ ] Yes
- [x] No, although some developers may have some unit tests that need to be fixed if they are checking the generated requests of their services.

### DIFF
--- a/packages/common/http/src/params.ts
+++ b/packages/common/http/src/params.ts
@@ -224,9 +224,15 @@ export class HttpParams {
     return this.keys()
         .map(key => {
           const eKey = this.encoder.encodeKey(key);
+          // `a: ['1']` produces `'a=1'`
+          // `b: []` produces `''`
+          // `c: ['1', '2']` produces `'c=1&c=2'`
           return this.map !.get(key) !.map(value => eKey + '=' + this.encoder.encodeValue(value))
               .join('&');
         })
+        // filter out empty values because `b: []` produces `''`
+        // which results in `a=1&&c=1&c=2` instead of `a=1&c=1&c=2` if we don't
+        .filter(param => param !== '')
         .join('&');
   }
 
--- a/packages/common/http/test/params_spec.ts
+++ b/packages/common/http/test/params_spec.ts
@@ -76,5 +76,20 @@ import {HttpParams} from '@angular/common/http/src/params';
         expect(body.keys()).toEqual(['a', 'b', 'c', 'd']);
       });
     });
+
+    describe('toString', () => {
+      it('should stringify string params', () => {
+        const body = new HttpParams({fromObject: {a: '', b: '2', c: '3'}});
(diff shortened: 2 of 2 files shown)


petebacondarwin (approved):
Please squash into one commit.

AndrewKushnir:
[Presubmit](http://test/OCL:291383109:BASE:291383143:1579885737482:21c869a5)

AndrewKushnir:
FYI, g3 presubmit is successful.

IgorMinar (approved):
lgtm, thanks @cexbrayat for describing the two solutions you considered. this non-breaking one makes sense.

naveedahmed1:
Probably because of this update I am seeing some strange behavior in my app:

For urls like this `/profile/6s4cr9d648f53278035412d4c/?applicant=true` in my app, the http interceptor is not applying the Authorization headers.

```
let newHeaders = req.headers.set('X-Requested-With', 'XMLHttpRequest').set('Content-Type', 'application/json');

        if (token) {
            newHeaders = newHeaders.set('Authorization', 'Bearer ' + token);
        }

        // Clone the request to add the new header.
        if (req.body) {
            modifiedRequest = req.clone({ headers: newHeaders, url: url, body: this.parseModel(req.body) });

        } else {
            modifiedRequest = req.clone({ headers: newHeaders, url: url });
        }
```

`parseModel` function removes empty properties

```
parseModel(model: any) {
        let parsedModel = {};
        for (var key in model) {
            if (key.indexOf('Data') === -1) {
                if (Object.prototype.toString.call(model[key]) == '[object Object]') {
                    parsedModel[key] = this.parseModel(model[key]);
                } else
                    if (!(this.isNullOrEmpty(model[key]))) {
                        parsedModel[key] = model[key];
                    }
            }
        }
        return parsedModel;
    }
```

Another section of my app in which I am seeing this issue is the one that uses below code, when `this.modelId` is null:

```
let model = {
                    id: photoId,
                    modelId: this.modelId,
                    typeId: this.typeId
                };
                this.http.post('/photo/remove/', model)
               .......
               .......
```

cexbrayat:
@naveedahmed1 I don't think this is related... Can you please open an issue and provide a minimal repro so we can investigate? Thanks

PR #34900, opened 2020-01-22
fix(docs-infra): switch over to img[loading=lazy] for front page images

IgorMinar:
This should speed up the perceived loading latency as the image loading won't block rendering.

https://addyosmani.com/blog/lazy-loading/

Screen Shot 2020-01-22 at 7 01 30 AM

### DIFF
--- a/aio/content/marketing/index.html
+++ b/aio/content/marketing/index.html
@@ -9,7 +9,7 @@
 
     <!-- LOGO -->
     <div class="hero-logo">
-      <img src="assets/images/logos/angular/angular.svg" alt="Angular" loading="lazy">
+      <img src="assets/images/logos/angular/angular.svg" alt="Angular">
     </div>
 
     <!-- CONTAINER -->
--- a/aio/content/marketing/index.html
+++ b/aio/content/marketing/index.html
@@ -9,7 +9,7 @@
 
     <!-- LOGO -->
     <div class="hero-logo">
-      <img src="assets/images/logos/angular/angular.svg" alt="Angular"/>
+      <img src="assets/images/logos/angular/angular.svg" alt="Angular" loading="lazy">
     </div>
 
     <!-- CONTAINER -->
@@ -35,7 +35,7 @@
     <div layout="row" layout-xs="column" class="home-row homepage-container">
       <div class="promo-img-container promo-1">
         <div>
-          <img height="222" width="340" src="generated/images/marketing/home/responsive-framework.svg" alt="responsive framework">
+          <img height="222" width="340" src="generated/images/marketing/home/responsive-framework.svg" alt="responsive framework" loading="lazy">
         </div>
       </div>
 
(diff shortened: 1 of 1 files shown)


IgorMinar:
hmm.. this doesn't seem to work in https://pr34900-985c0b4.ngbuilds.io/ in a recent browser that does support loading=lazy. no idea why.

IgorMinar:
it seems to work correctly here: https://mathiasbynens.be/demo/img-loading-lazy I'm not sure what's the difference.

petebacondarwin:
What do you expect to see? If the images are on the screen at start up they are not lazy loaded.
I also note that a certain number of the cat images get loaded any way.

Screenshot 2020-01-22 at 16 51 24

So perhaps we have so few images on the home page it does not bother waiting?

IgorMinar:
maybe, but then why does lighthouse complain. Seems like there is a disconnect somewhere...

IgorMinar:
to be clear, lighthouse complains about the images even with loading=lazy - can be seen when auditing https://pr34900-985c0b4.ngbuilds.io/

gkalpak:
Based on [this][1] (and the corresponding [Chromium source code][2]), my understanding is that images "below the fold" will be loaded as soon as they reach 3000 to 8000 pixels from the viewport (depending on network type). Given that our home page is less than 2300px long, it is no surprise that all images are loaded asap.

I agree that Lighthouse should not be complaining though :grin:

[1]: https://web.dev/native-lazy-loading/#load-in-distance-threshold
[2]: https://cs.chromium.org/chromium/src/third_party/blink/renderer/core/frame/settings.json5?l=971-1003&rcl=e8f3cf0bbe085fee0d1b468e84395aad3ebb2cad

IgorMinar:
@gkalpak thank you! I ❤️ that you checked the chromium source code.

@addyosmani @housseindjirdeh is this something that should be fixed in Lighthouse audits? The issue is that lighthouse complains about us not deferring loading of offscreen images (see screenshot above), and the audit warning doesn't go away when we add `loading=lazy` to the problematic images.

housseindjirdeh:
@IgorMinar I can see how that audit suggestion can be a little confusing if the loading attribute is being used correctly (`loading="lazy"`).

Opened an issue to discuss on the Lighthouse repo to discuss some possible ways to approach this: https://github.com/GoogleChrome/lighthouse/issues/10259

housseindjirdeh:
[Fix](https://github.com/GoogleChrome/lighthouse/pull/10117) merged in Lighthouse repo. I tested locally on latest master and all the front page images are not flagged as expected 👍🏾.

Changes should be available in the next LH release (v5.7.0)!

IgorMinar:
thank you @housseindjirdeh !

IgorMinar:
@gkalpak I think this PR is still useful because it is explicit about our intent regardless of browser heuristics. Can you please review?

gkalpak (approved):
👌

jessicajaniuk:
@IgorMinar What's the status of this PR?

gkalpak:
This is still relevant and should be good to go. I've rebased it on latest master and resolved conflicts.

PR #34902, opened 2020-01-22
build: vendor yarn using the .yarnrc yarn-path value

josephperrott:
Rather than enforcing yarn versioning using `package.json`'s
engines value.  We can utilize yarn's `yarn-path` value to
ensure that the version of yarn used at execution time is
consistent for everyone who uses our repo.  This is the first
step in this wider vendoring process.  We will use this same
vendoring mechanism for CI after https://github.com/bazelbuild/rules_nodejs/pull/1569 lands


gregmagolan:
Cool. There may be a bit of work to do to also use this within the bazel integration tests after https://github.com/angular/angular/pull/33927 lands but can deal with that then.

gregmagolan (approved):

.yarnrc
```diff
@@ -0,0 +1,5 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
+
+
+yarn-path ".yarn/releases/yarn-1.21.1.js"
```
devversion: So, now we will have two copies of Yarn right? We should be able to either point this one to the one in `third_party` (the `cli.js` file IIRC), or have the Bazel one refer to the `.yarn/releases` folder?

gregmagolan: Bazel won't see `yarn-path` until https://github.com/bazelbuild/rules_nodejs/pull/1569 lands but come to think of it, the vendored_yarn would probably not work with this path structure as it assumes the distribution format. But once https://github.com/bazelbuild/rules_nodejs/pull/1569 we can remove vendored_yarn entirely and just have bazel run the default rules_nodejs yarn version which will automatically launch `.yarn/releases/yarn-1.21.1.js` when it looks at the `.yarnrc`

devversion: My thinking was that we still need the Yarn vendored version for the CircleCI jobs (e.g. to download bazel) Though we might be able to just use the default Yarn version from the CircleCI images then? (which will then resolve to the vendored due to the `rc` file too)

devversion (approved):

.yarnrc
josephperrott: Correct, as long as the version of bazel is `>1.0.0` it will work.  And from my understanding Circle's yarn install is far past `1.0.0`.

Splaktar: > as long as the version of bazel is >1.0.0 it will work. And from my understanding Circle's yarn install is far past 1.0.0.

Did you mean yarn instead of bazel in the first sentence?

josephperrott: Correct, needs to be be `yarn > 1.0`

PR #33613, opened 2019-11-06
docs: app-name not coming on browser

santoshyadavdev:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [X] Documentation content changes
- [ ] angular.io application / infrastructure changes
- [ ] Other... Please describe:

## What is the current behavior?

Issue Number: #33612 

## Does this PR introduce a breaking change?

- [ ] Yes
- [X] No

PR #34908, opened 2020-01-22
docs: fix syntax of code example for ES5 configuration

cbejensen:
## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

At the end of the second code example under "**Configuring serve for ES5**", Markdown correctly formats `&lt;app-name&gt;:build:es5` as `<app-name>:build:es5`, but the rendered code sample online at angular.io doesn't show `<app-name>` at all, so it ends up being `"browserTarget": ":build:es5"`, which makes the note below the code sample confusing as it says to replace `<app-name>`, which doesn't exist.

## What is the new behavior?
Code example is rendered correctly to match the hint underneath. I don't know if there's a way to render "less than" and "greater than" symbols correctly online, so I used all caps with an underscore delimiter.

- [ ] Yes
- [x] No

### DIFF
--- a/aio/content/guide/deployment.md
+++ b/aio/content/guide/deployment.md
@@ -663,14 +663,14 @@ In `angular.json` add two new configuration sections under the `build` and `serv
      ...
     },
     "es5": {
-      "browserTarget": "&lt;app-name&gt;:build:es5"
+      "browserTarget": "YOUR_APP_NAME:build:es5"
     }
   }
 },
 
 </code-example>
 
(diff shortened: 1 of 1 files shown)


gkalpak:
Thx for submitting a pull request, @cbejensen! 
The problem you are trying to address has been fixed in #33613. The reason it doesn't show up on https://angular.io/ yet is that the fix was not backported to the 8.x branch, which is the latest stable version (while 9.x is in RC) and gets deployed to angular.io.

You can see the fix live on our staging environment at https://next.angular.io/. The fix will land to angular.io once 9.0.0 is released (and 9.x becomes the new stable version).

##
Given that the issue is already fixed on master, I am going to close this PR.
(If you want to see it land to angular.io sooner, you can try to backport #33613 to the 8.2.x branch.)

PR #34910, opened 2020-01-22
revert: refactor(ivy): remove styleSanitizer instruction in favor of an inline param (#34480)

matsko:
This reverts commit 84d24c08e1cd5ff3cb28e079ef03a347abf6d98b.


matsko:
No need for a presubmit because G3 master is exactly where it needs to be without this patch.

AndrewKushnir (approved):
LGTM

PR #34912, opened 2020-01-22
Fix bug removing a component from an NgModule during an incremental build

alxhub:


packages/compiler-cli/test/ngtsc/incremental_spec.ts
```diff
@@ -251,6 +251,73 @@ runInEachFileSystem(() => {
...
+        export class Module {}
+      `);
+
+      env.driveMain();
+      env.flushWrittenFileTracking();
+
+      // Remove the component from the module and recompile.
+      env.write('module.ts', `
+        import {NgModule} from '@angular/core';
+        import {Cmp} from './cmp';
```
JoostK: This import can and should be removed

JoostK (approved):
LGTM, but ci is unhappy (leftover `debugger`)

PR #34923, opened 2020-01-23
perf(ivy): add more styling use-cases to benchmarks

pkozlowski-opensource:
New benchmarks focus on style bindings (style and ngStyle).

### DIFF
--- a/modules/benchmarks/src/styling/ng2/styling.ts
+++ b/modules/benchmarks/src/styling/ng2/styling.ts
@@ -20,9 +20,12 @@ import {BrowserModule} from '@angular/platform-browser';
     <ng-template #t5><button class="foo" [ngClass]="{bar: exp ==='bar'}"></button></ng-template>
     <ng-template #t6><button class="foo" [ngStyle]="staticStyle" [style.background-color]="exp"></button></ng-template>
     <ng-template #t7><button style="color: red"></button></ng-template>
+    <ng-template #t8><button [style.width.px]="exp ==='bar' ? 10 : 20" [style.color]="exp"></button></ng-template>
+    <ng-template #t9><button style="width: 10px" [style.color]="exp"></button></ng-template>
+    <ng-template #t10><button [ngStyle]="{'width.px': exp ==='bar' ? 10 : 20, color: exp}"></button></ng-template>
     
     <div>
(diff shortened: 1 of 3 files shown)


modules/benchmarks/src/styling/ng2/styling.ts
```diff
@@ -20,9 +20,12 @@ import {BrowserModule} from '@angular/platform-browser';
     <ng-template #t5><button class="foo" [ngClass]="{bar: exp ==='bar'}"></button></ng-template>
     <ng-template #t6><button class="foo" [ngStyle]="staticStyle" [style.background-color]="exp"></button></ng-template>
     <ng-template #t7><button style="color: red"></button></ng-template>
+    <ng-template #t8><button [style.width.px]="exp ==='bar' ? 10 : 20" [style.color]="exp"></button></ng-template>
+    <ng-template #t9><button style="width: 10px" [style.color]="exp"></button></ng-template>
+    <ng-template #t10><button [ngStyle]="{'width.px': exp ==='bar' ? 10 : 20, color: exp}"></button></ng-template>
```
IgorMinar: the unfortunate side effect of keeping the code simple and piggy-backing on `exp` is that we now have to use ternary operator which adds extra overhead. I don't know how significant it is but it rubs me the wrong way just a bit because the benchmark is not focused just on ngStyle (or `[style]` above). Are you not concerned that this is going to blur the results a bit? Without knowing how much overhead the ternary operator adds, I can't tell how significant this is.

IgorMinar (approved):
Can you please update the e2e tests? otherwise lgtm

modules/benchmarks/src/styling/ng2/styling.ts
IgorMinar: I'll leave it up to you to use your judgement on this one.

pkozlowski-opensource: Here is my thinking: the way I see the purpose of this test is to get 2 data-points:
* gap to VE;
* evolution over time.

In the 2 above cases:
* `exp ==='bar' ? 10 : 20` is exactly the same (in other words, it should contribute to the overall time in the same way, so should not make any difference when measuring _deltas_;
* based on my understanding of things evaluating `exp ==='bar' ? 10 : 20` is _really_ fast (we are not accessing any memory here etc.) and should be negligible given all the other work we are doing.

So, I hear what you are saying and can change things but I also tend to think that "it doesn't matter".

IgorMinar:
sgtm

note: e2e tests will be added in #34753

PR #34929, opened 2020-01-23
refactor(ivy): remove styleSanitizer instruction in favor of an inline param

matsko:
Resurrection of https://github.com/angular/angular/pull/34480


gkalpak:
@googlebot I consent.

PR #34934, opened 2020-01-23
docs: use an explicit product id in the getting started example 

### LINKED ISSUE #34738: Routing Tutorial
Seems to be an issue with this tutorial. Maybe I am doing something wrong but the products.ts does not have a product id variable in its array and basically the end result is this tutorial does not completely work.

troelslenda:
Simplyfing the example by prodiving an id for each
product instead of relying on their index in the array.

Closes #34738

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #34738

## What is the new behavior?

Instead of using the index to extract the product from the array, each product now has an id.

- [ ] Yes
- [ ] No

### DIFF
--- a/aio/content/examples/getting-started/src/app/product-details/product-details.component.1.ts
+++ b/aio/content/examples/getting-started/src/app/product-details/product-details.component.1.ts
@@ -23,8 +23,11 @@ export class ProductDetailsComponent implements OnInit {
   // #enddocregion props-methods
   // #docregion get-product
   ngOnInit() {
-    this.route.paramMap.subscribe(params => {
-      this.product = products[+params.get('productId')];
+    // First get the product id from the current route.
+    const productIdFromRoute = this.route.snapshot.paramMap.get('productId');
+    // Find the product that correspond with the id provided in route.
+    this.product = products.find(product => {
+      return product.id === Number(productIdFromRoute);
     });
   // #docregion product-prop
   }
--- a/aio/content/examples/getting-started/src/app/product-details/product-details.component.ts
+++ b/aio/content/examples/getting-started/src/app/product-details/product-details.component.ts
@@ -30,8 +30,11 @@ export class ProductDetailsComponent implements OnInit {
 // #docregion get-product
   ngOnInit() {
 // #enddocregion props-methods
-    this.route.paramMap.subscribe(params => {
-      this.product = products[+params.get('productId')];
+    // First get the product id from the current route.
+    const productIdFromRoute = this.route.snapshot.paramMap.get('productId');
+    // Find the product that correspond with the id provided in route.
+    this.product = products.find(product => {
+      return product.id === Number(productIdFromRoute);
     });
 // #docregion props-methods
   }
--- a/aio/content/examples/getting-started/src/app/product-list/product-list.component.html
+++ b/aio/content/examples/getting-started/src/app/product-list/product-list.component.html
@@ -1,10 +1,10 @@
 <h2>Products</h2>
 
 <!-- #docregion router-link -->
-<div *ngFor="let product of products; index as productId">
+<div *ngFor="let product of products">
 
   <h3>
-    <a [title]="product.name + ' details'" [routerLink]="['/products', productId]">
+    <a [title]="product.name + ' details'" [routerLink]="['/products', product.id]">
       {{ product.name }}
     </a>
   </h3>
--- a/aio/content/examples/getting-started/src/app/products.ts
+++ b/aio/content/examples/getting-started/src/app/products.ts
@@ -1,15 +1,18 @@
 export const products = [
   {
+    id: 1,
     name: 'Phone XL',
     price: 799,
     description: 'A large phone with one of the best screens'
   },
   {
+    id: 2,
     name: 'Phone Mini',
     price: 699,
     description: 'A great phone with one of the best cameras'
   },
   {
+    id: 3,
     name: 'Phone Standard',
     price: 299,
     description: ''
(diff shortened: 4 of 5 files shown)


aio/content/examples/getting-started/src/app/product-details/product-details.component.1.ts
```diff
@@ -23,7 +23,7 @@ export class ProductDetailsComponent implements OnInit {
   // #docregion get-product
   ngOnInit() {
     this.route.paramMap.subscribe(params => {
-      this.product = products[+params.get('productId')];
+      this.product = products.find(product => product.id === +params.get('productId'));
```
IgorMinar: should it be better to switch the data-structure to map instead of doing this inefficient (and for newcomers not easy to comprehend) lookup?

IgorMinar (commented):
@troelslenda  thank you for the PR, would you mind changing the commit description to "fix(docs): use an explicit product id in the getting started example"?

"docs-infra" is used only for the infrastructure we use for building the docs site, but not the content (which include example code).

I do like this change, but I don't know how it affects the entire getting started example/guide, so I'll leave it up to @aikidave and @StephenFluin to make the final decision.

aio/content/examples/getting-started/src/app/product-details/product-details.component.1.ts
troelslenda: Yes. @IgorMinar It would make this easier to understand. But how about the iteration in the template? Would you suggest `keyvalue` pipe? Would that be to magical for newcomers?

troelslenda:
> @troelslenda thank you for the PR, would you mind changing the commit description to "fix(docs): use an explicit product id in the getting started example"?
> 
> "docs-infra" is used only for the infrastructure we use for building the docs site, but not the content (which include example code).
> 
> I do like this change, but I don't know how it affects the entire getting started example/guide, so I'll leave it up to @aikidave and @StephenFluin to make the final decision.

Thank you for your reply and review @IgorMinar. the scope __docs__ seems to be invalid according to CONTRIBUTING.md. Should I change the commit anyway?

aio/content/examples/getting-started/src/app/product-details/product-details.component.1.ts
gkalpak: FWIW, I think a map would come with its own challenges (such as iterating on it in the template).
Also, having products in an array is more similar to what happens most often in real apps.

gkalpak:
I believe the appropriate commit message type is `docs` (with no scope): `docs: ...`

troelslenda:
Thank you @gkalpak. I changed the commit message accordingly.

aio/content/examples/getting-started/src/app/product-details/product-details.component.1.ts
troelslenda: Maybe Observables and subscription to them is out of context in this tutorial? We could skip the subscription to the paramMap Observable and use snapshot instead. Also use `forEach` instead of `find` to make it even more expressive.
```typescript
const productIdFromRoute = this.route.snapshot.paramMap.get('productId');
products.forEach(product => {
  if (product.id === productIdFromRoute) {
    this.product = product;
  }
})
```
However, I think that knowledge regarding the array methods is required in order to understand this tutorial.
So maybe just
```typescript
// First get the product id from the current route.
const productIdFromRoute = this.route.snapshot.paramMap.get('productId');
// Find the product that correspond with the id provided in route.
this.product = products.find(product => {
  return product.id === productIdFromRoute;
})

**Comment by troelslenda** [comment]
How should we proceed with this?

**Comment by kapunahelewong** [comment]
@troelslenda thank you for the bump. I'm bringing your PR up to the docs team to request another review. 🙂

File: aio/content/examples/getting-started/src/app/product-details/product-details.component.1.ts (hunk 1 above)
Comment (aikidave) [reply aio/content/examples/getting-started/src/app/product-details/product-details.component.1.ts]: I would prioritize using code that represents good programming habits with Angular. :) 

I think the code comment you've provided should be sufficient enough, as you state clearly what the code is doing.

**Review by aikithoughts — APPROVED** [review]
I'm fine with these changes! If we want to have a deeper discussion of how to improve the code overall, we can do that separately.

**Comment by troelslenda** [comment]
@IgorMinar can you clarify what cleanup is needed?
@aikidave and @kapunahelewong what's the next step?

**Comment by kapunahelewong** [comment]
@troelslenda, at the moment, we just need an engineering approval. We also needed a rebase, to resolve the conflicts, which I just took care of for you and pushed up. Thank you again for this work!

File: aio/content/examples/getting-started/src/app/product-details/product-details.component.1.ts [hunk 2]
```diff
@@ -22,8 +22,11 @@ export class ProductDetailsComponent implements OnInit {
   // #enddocregion props-methods
   // #docregion get-product
   ngOnInit() {
-    this.route.paramMap.subscribe(params => {
-      this.product = products[+params.get('productId')];
+    // First get the product id from the current route.
+    const productIdFromRoute = this.route.snapshot.paramMap.get('productId');
+    // Find the product that correspond with the id provided in route.
+    this.product = products.find(product => {
+      return product.id === +productIdFromRoute;
```
atscott, line 29: I know this already existed in some form, but could you please use `Number(productIdFromRoute)` here instead of the unary `+`? This is part of the Google style guide:

> Code must not use unary plus (+) to coerce strings to numbers. Parsing numbers can fail, has surprising corner cases, and can be a code smell (parsing at the wrong layer). A unary plus is too easy to miss in code reviews given this.

atscott (approved):
LGTM - I agree that this approach is much better. Relying on the array order is not very production-like and our examples should to try be as real-world as possible.

aio/content/examples/getting-started/src/app/product-details/product-details.component.1.ts
troelslenda: done

PR #34952, opened 2020-01-24
build: vendor yarn using the .yarnrc yarn-path value

gkalpak:
This is a backport of #34902 to the 8.2.x branch (except that yarn v1.17.3 is used to satisfy the yarn version requirements of the 8.2.x branch).
This will ensure that an appropriate yarn version will be used in the `aio_monitoring_stable` CI job, which checks out files from the git branch corresponding to the latest stable Angular version (currently the 8.2.x branch).

(Example of the current failure: https://circleci.com/gh/angular/angular/598455)

PR #34954, opened 2020-01-24
ci(docs-infra): use the correct version of yarn in `aio_monitoring_stable`

gkalpak:
The `aio_monitoring_stable` CI job needs to check out (and execute commands on) some files from the stable branch. Therefore, it needs to use a version of `yarn` that is compatible with the `engines` version specified in `aio/package.json`.

In #34902 (and subsequently #34952 for the 8.2.x branch), the way to access our vendored version of yarn changed. This commit ensures that we checkout the necessary files from the stable branch to use an appropriate yarn version.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -621,14 +621,10 @@ jobs:
       - init_environment
       - run: setPublicVar_CI_STABLE_BRANCH
       - run:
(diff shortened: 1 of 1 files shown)


gkalpak:
This is half of the fix. The other half is #34952.

josephperrott (approved):
LGTM

PR #34969, opened 2020-01-25
perf(ivy): remove unused argument in hostBindings function

crisbeto:
We had some logic for generating and passing in the `elIndex` parameter into the `hostBindings` function, but it wasn't actually being used for anything. The only place left that had a reference to it was the `StylingBuilder` and it only stored it without referencing it again.


pkozlowski-opensource (approved):
LGTM from the runtime point of view. 

Note: I would love to in-line `invokeHostBindingsInCreationMode` (and limit checks on `def.hostBindings !== null)` but as @mhevery pointed out in multiple comments we should rather refactor how we keep track of host bindings instructions to execute (store directive indexes and don't store `null`)

AndrewKushnir:
[Presubmit](http://test/OCL:291755545:BASE:291755609:1580151187743:e98436bf)

AndrewKushnir (approved):
LGTM 👍

PR #34974, opened 2020-01-26
build: add missing @babel/generator bazel dep to fix  //packages/localize/src/tools/test:test

IgorMinar:
It's not clear how this ever worked (npm hoisting is a suspect though), but this dep is required because
one of the tests imports @babel/generator in the ts sources.

The lack of this dep is breaking builds on the master branch.

More discussion about this issue on Slack: https://angular-team.slack.com/archives/C07DT5M6V/p1579934766007500


gkalpak (approved):

packages/localize/src/tools/test/BUILD.bazel
```diff
@@ -11,6 +11,7 @@ ts_library(
         "//packages/compiler",
         "//packages/localize",
         "//packages/localize/src/tools",
+        "@npm//@babel/generator",
```
gkalpak: Do we need to have these packages in `package.json` or is `BUILD.bazel` enough?

gkalpak (approved):

gregmagolan:
There are quite a few @babel/generators in node_modules on master.
```
./@angular-devkit/build-angular/node_modules/@babel/core/node_modules/@babel/generator/package.json
./@angular-devkit/build-angular/node_modules/@babel/generator/package.json
./@angular-devkit/build-angular/node_modules/@babel/traverse/node_modules/@babel/generator/package.json
./@angular-devkit/build-angular/node_modules/@babel/helpers/node_modules/@babel/generator/package.json
./@babel/generator/package.json
./@babel/traverse/node_modules/@babel/generator/package.json
```
The test that is causing trouble transitively depends on @babel/core which transitively depends on @babel/generator so its import from @babel/generator would only work if the @babel/core dep was being hoisted to the root.... which on master that looks to be the case:
```
cat node_modules/@babel/generator/package.json
{
  "name": "@babel/generator",
  "version": "7.6.4",
```
But if it wasn't hoisted due to package.json changes that could break the test
I don't think this is sufficient to fix the problem as it is possible that there will be no `@npm//@babel/generator` layed out.

The file that uses `@babel/generator` is `packages/localize/src/tools/test/translate/source_files/source_file_utils_spec.ts`. So I think the correct shape for the fix is

1) to add `@babel/generator` and `@npm//@babel/core` to the `test_lib` target `packages/localize/src/tools/test/BUILD.bazel`. Its currently getting both transitively from `//packages/localize/src/tools` but only working as `@babel/generator` from `@bable/core` is being hoisted to the top.

2) explicitely add `@babel/generator` to the root package.json so we're not depending on the package manager hoisting it to the top

gregmagolan:
@IgorMinar Threw up a quick https://github.com/angular/angular/pull/34980 with my suggestion

packages/localize/src/tools/test/BUILD.bazel
IgorMinar: I think you are right. I added it as well as the @types/babel__generator package in a separate commit.

IgorMinar:

IgorMinar:
@gregmagolan George noticed the same thing, and I addressed it in follow up commits.

packages/localize/src/tools/test/translate/source_files/es5_translate_plugin_spec.ts
```diff
@@ -140,9 +140,9 @@ describe('makeEs5Plugin', () => {
          expect(diagnostics.hasErrors).toBe(true);
          expect(diagnostics.messages[0]).toEqual({
            type: 'error',
-           message: '/app/dist/test.js: `$localize` called without any arguments.\n' +
```
IgorMinar: @petebacondarwin @gkalpak this is unexpected and likely due to the @babel version update in https://github.com/angular/angular/pull/34974/commits/0a5a944824353ea985b91fc28c2835a0ab26a5a4 can you please look into why the path disappeared and fix it in a follow up PR if possible? I'd like to get this PR landed so that it doesn't block other PRs.

petebacondarwin: Not sure, yet, why the file path has disappeared. But the other changes to the error messages (i.e. the extra `^^^^` etc) are due to this commit in Babel: https://github.com/babel/babel/commit/f1bc6c4e18dbff8aad8e0b53d80c787c9dc9870c

petebacondarwin: Ahah! The filename  was removed in this commit: https://github.com/babel/babel/commit/dcf7d89b8e9b6b364aae17652f94a4f0d3fe0950#diff-7a78d3752af461cdbf66d58d59a2026d

PR #34978, opened 2020-01-26
feat(docs-infra): add useful links if landed on 404 page

### LINKED ISSUE #31532: Improve bad URL experience

### Description

Per discussion in https://github.com/angular/angular/pull/31390:

Improve bad URL experience for users by giving them relevant data instead of 'no results found' for something they didn't _actually_ search for.

Example: https://angular.io/lemonade

### Reproduction Steps**
Go to https://angular.io/lemonade to see page and message

Don't actually show the "I have nothing to show you" message if we don't have any suggestions OR even better, show them a few items where they might want to spend some time on like the blog, the guide, etc.

## 📷Screenshot
Screen Shot 2019-07-04 at 12 32 16 PM

ajitsinghkaler:
Added additional links useful links which can help user find the things he is looking for on the 404 page
Fixes #31532

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

On no link found no additional shown which can be helpful

Issue Number: 31532

## What is the new behavior?
Added additional links which can be helpful to users to find what they are looking for

- [ ] Yes
- [x] No

### DIFF
--- a/aio/src/app/shared/search-results/search-results.component.html
+++ b/aio/src/app/shared/search-results/search-results.component.html
@@ -1,30 +1,57 @@
-<div class="search-results">
-  <div *ngIf="searchAreas.length; then searchResults; else notFound"></div>
+<div class="search-results" [ngSwitch]="searchState">
+
+  <ng-container *ngSwitchCase="'in-progress'">
+    <p class="no-results">Searching ...</p>
+  </ng-container>
+
+  <ng-container *ngSwitchCase="'results-found'">
+    <h2 class="visually-hidden">Search Results</h2>
+    <div class="search-area" *ngFor="let area of searchAreas">
+      <h3 class="search-section-header">{{area.name}} ({{area.pages.length + area.priorityPages.length}})</h3>
+      <ul class="priority-pages" >
+        <li class="search-page" *ngFor="let page of area.priorityPages">
+          <a class="search-result-item" href="{{ page.path }}" (click)="onResultSelected(page, $event)">
+            <span class="symbol {{page.type}}" *ngIf="area.name === 'api'"></span>
+            <span [class.deprecated-api-item]="page.deprecated">{{ page.title }}</span>
+          </a>
+        </li>
+      </ul>
+      <ul>
+        <li class="search-page" *ngFor="let page of area.pages">
+          <a class="search-result-item" href="{{ page.path }}" (click)="onResultSelected(page, $event)">
+            <span class="symbol {{page.type}}" *ngIf="area.name === 'api'"></span>
+            <span [class.deprecated-api-item]="page.deprecated">{{ page.title }}</span>
+          </a>
+        </li>
+      </ul>
+    </div>
+  </ng-container>
+  
+  <ng-container *ngSwitchCase="'no-results-found'">
+    <div class="search-area">
+      <p class="no-results">
+        No results found.<br>
+        Here are a few links that might be helpful in finding what you are looking for:
+      </p>
+      <ul class="priority-pages">
+        <li class="search-page">
+          <a class="search-result-item" href="api">API reference</a>
+        </li>
+        <li class="search-page">
+          <a class="search-result-item" href="resources">Resources</a>
+        </li>
+        <li class="search-page">
+          <a class="search-result-item" href="guide/glossary">Glossary</a>
+        </li>
+        <li class="search-page">
+          <a class="search-result-item" href="guide/cheatsheet">Cheat-sheet</a>
+        </li>
+        <li class="search-page">
+          <a class="search-result-item" href="https://blog.angular.io/">Angular blog</a>
+        </li>
+      </ul>
+    </div>
+  </ng-container>
 </div>
 
-<ng-template #searchResults>
-  <h2 class="visually-hidden">Search Results</h2>
-  <div class="search-area" *ngFor="let area of searchAreas">
-    <h3 class="search-section-header">{{area.name}} ({{area.pages.length + area.priorityPages.length}})</h3>
-    <ul class="priority-pages" >
-      <li class="search-page" *ngFor="let page of area.priorityPages">
-        <a class="search-result-item" href="{{ page.path }}" (click)="onResultSelected(page, $event)">
-          <span class="symbol {{page.type}}" *ngIf="area.name === 'api'"></span>
-          <span [class.deprecated-api-item]="page.deprecated">{{ page.title }}</span>
-        </a>
-      </li>
-    </ul>
-    <ul>
-      <li class="search-page" *ngFor="let page of area.pages">
-        <a class="search-result-item" href="{{ page.path }}" (click)="onResultSelected(page, $event)">
-          <span class="symbol {{page.type}}" *ngIf="area.name === 'api'"></span>
-          <span [class.deprecated-api-item]="page.deprecated">{{ page.title }}</span>
-        </a>
-      </li>
-    </ul>
-  </div>
-</ng-template>
 
-<ng-template #notFound>
-  <p class="not-found">{{notFoundMessage}}</p>
-</ng-template>
--- a/aio/src/app/shared/search-results/search-results.component.spec.ts
+++ b/aio/src/app/shared/search-results/search-results.component.spec.ts
@@ -170,6 +170,12 @@ describe('SearchResultsComponent', () => {
     expect(getText()).toContain('Searching ...');
   });
 
+  it('should not display default links while searching', () => {
+    fixture.detectChanges();
+    const resultLinks = fixture.debugElement.queryAll(By.css('.search-page a'));
+    expect(resultLinks.length).toEqual(0);
+  });
+
   describe('when a search result anchor is clicked', () => {
     let searchResult: SearchResult;
     let selected: SearchResult|null;
@@ -214,9 +220,25 @@ describe('SearchResultsComponent', () => {
   });
 
   describe('when no query results', () => {
-    it('should display "not found" message', () => {
+    beforeEach(() => {
       setSearchResults('something', []);
+    });
+
+    it('should display "not found" message', () => {
       expect(getText()).toContain('No results');
+    });
+
+    it('should contain reference links', () => {
+      const resultLinks = fixture.debugElement.queryAll(By.css('.search-page a'));
+      const resultHrefs = resultLinks.map(a => a.nativeNode.getAttribute('href'));
+      expect(resultHrefs.length).toEqual(5);
+      expect(resultHrefs).toEqual([
+        'api',
+        'resources',
+        'guide/glossary',
+        'guide/cheatsheet',
+        'https://blog.angular.io/',
+      ]);
     });
   });
 });
--- a/aio/src/app/shared/search-results/search-results.component.ts
+++ b/aio/src/app/shared/search-results/search-results.component.ts
@@ -1,6 +1,12 @@
 import { Component, EventEmitter, Input, OnChanges, Output } from '@angular/core';
 import { SearchResult, SearchResults, SearchArea } from 'app/search/interfaces';
 
+enum SearchState {
+  InProgress = 'in-progress',
+  ResultsFound = 'results-found',
+  NoResultsFound = 'no-results-found'
+}
+
 /**
  * A component to display search results in groups
  */
@@ -23,11 +29,18 @@ export class SearchResultsComponent implements OnChanges {
   resultSelected = new EventEmitter<SearchResult>();
 
   readonly defaultArea = 'other';
-  notFoundMessage = 'Searching ...';
+  searchState: SearchState = SearchState.InProgress;
   readonly topLevelFolders = ['guide', 'tutorial'];
   searchAreas: SearchArea[] = [];
 
   ngOnChanges() {
+    if (this.searchResults === null) {
+      this.searchState = SearchState.InProgress;
+    } else if (this.searchResults.results.length) {
+      this.searchState = SearchState.ResultsFound;
+    } else {
+      this.searchState = SearchState.NoResultsFound;
+    }
     this.searchAreas = this.processSearchResults(this.searchResults);
   }
 
@@ -43,7 +56,6 @@ export class SearchResultsComponent implements OnChanges {
     if (!search) {
       return [];
     }
-    this.notFoundMessage = 'No results found.';
     const searchAreaMap: { [key: string]: SearchResult[] } = {};
     search.results.forEach(result => {
       if (!result.title) { return; } // bad data; should fix
--- a/aio/src/styles/2-modules/_search-results.scss
+++ b/aio/src/styles/2-modules/_search-results.scss
@@ -67,10 +67,15 @@ aio-search-results {
       }
     }
 
-    .not-found {
+    .no-results {
       color: $white;
       text-align: center;
       margin: 16px;
+    }
+
+    a {
+      color: $white;
+      font-weight: 500;
     }
 
     @media (max-width: 600px) {
@@ -103,6 +108,10 @@ aio-search-results {
 
       .not-found {
         color: $darkgray;
+      }
+
+      a {
+        color: $blue;
       }
     }
   }
--- a/aio/scripts/_payload-limits.json
+++ b/aio/scripts/_payload-limits.json
@@ -3,7 +3,7 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 2987,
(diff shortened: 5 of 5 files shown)


gkalpak (changes requested):
Thx for taking this on, @ajitsinghkaler! Per the discussion in #31532, the generic links would appear if there can be no relevant search results. (I would be fine putting them below the search result too.)

So, they need to be somehow incorparated into the `aio-file-not-found-search` component.

ajitsinghkaler:
@gkalpak I've made the changes please have a look

ajitsinghkaler:
@gkalpak I've made the changes please have a look I've incorporated it in the aio-file-not-found-search component.

ajitsinghkaler:
@gkalpak can you please explain why e2e tests are failing

ajitsinghkaler:
@gkalpak can you please look into this too

gkalpak (changes requested):
Regarding why the e2e tests are failing:

The `SearchResultsComponent` will initially have an empty result set (since it hasn't received the search results yet) and will therefore show the [#notFound template][1]. At that point, however, the `notFoundMessage` property is [set to "Searching..."][2].

Then, once the results are received, the `notFoundMessage` is [changed to "No results found"][3] and either the [#searchResults][4] or the [#notFound][1] templates are shown (the latter now showing "No results found" instead of "Searching...").

So, while we pretend there are two states (no results or results), there are actually three: searching, no results, results
But we are piggybacking on the "no results" state to handle the "searching" case too.

##
Now, why this causes the e2e to fail?

In the e2e tests, we [send a search query][5] and run assertions against the [search results][6]. The way we retrieve the search results is by [waiting][7] for [elements that match `.search-results li`][8].

Previously, such elements would only be present when there were some actual search results. With your changes, these elements are also present whenever the [#notFound][1] template is shown, which (as explained above) is also showing while the search is in progress.

As a result, the e2e tests will look at the generic links and think they are the search results returned by the search query and thus try to run assertions on them. These assertions will fail, of course, because these "results" do not contain the expected terms.

---

**How to fix this?**

I think the proper fix is to have three templates:
- One for when there are results (i.e. the `#searchResults` template we already have).
- One for when there are no results showing the generic links (i.e. the `#notFound` template, except with the `notFoundMessage` hard-coded to "No results found.").
- One for when the search is in progress without the generic links (and with a hard-coded "Searching..." message).

Then we would also need a property on the component to track whether a search is in progress or not.

LMK if that makes sense and whether you need any help with the implementation.

[1]: https://github.com/angular/angular/blob/c35671c0a/aio/src/app/shared/search-results/search-results.component.html#L28-L30
[2]: https://github.com/angular/angular/blob/c35671c0a/aio/src/app/shared/search-results/search-results.component.ts#L26
[3]: https://github.com/angular/angular/blob/c35671c0a/aio/src/app/shared/search-results/search-results.component.ts#L46
[4]: https://github.com/angular/angular/blob/c35671c0a/aio/src/app/shared/search-results/search-results.component.html#L5-L26

[5]: https://github.com/angular/angular/blob/c35671c0a/aio/tests/e2e/src/search.e2e-spec.ts#L12
[6]: https://github.com/angular/angular/blob/c35671c0a/aio/tests/e2e/src/search.e2e-spec.ts#L13
[7]: https://github.com/angular/angular/blob/c35671c0a/aio/tests/e2e/src/app.po.ts#L83
[8]: https://github.com/angular/angular/blob/c35671c0a/aio/tests/e2e/src/app.po.ts#L82

aio/src/app/shared/search-results/search-results.component.html
```diff
@@ -26,5 +26,25 @@ <h3 class="search-section-header">{{area.name}} ({{area.pages.length + area.prio
...
+        </li>
+        <li class="search-page">
+            <a class="search-result-item" href="guide/glossary#glossary">Glossary</a>
+        </li>
+        <li  class="search-page">
+            <a class="search-result-item" href="guide/cheatsheet#cheat-sheet">Cheat-sheet</a>
+        </li>
+        <li  class="search-page">
+            <a  class="search-result-item" href="https://blog.angular.io/">Angular blog</a>
+        </li>
```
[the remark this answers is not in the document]
ajitsinghkaler: YS WILL DO

aio/src/styles/2-modules/_search-results.scss
```diff
@@ -103,6 +111,14 @@ aio-search-results {
       .not-found {
         color: $darkgray;
       }
+
+      .not-found-block {
+        flex-direction: column;
+        a {
+          color: $blue;
+          font-weight: 500;
```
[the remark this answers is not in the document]
ajitsinghkaler: there we applied color white which is not visible when you enter a wrong URL on the aio site so that is why I applied color blue as for font-weight I'll remove it.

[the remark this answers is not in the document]
gkalpak: Yes, I was referring to the `font-weigth` line specifically. Thx!

ajitsinghkaler:
@gkalpak I've made the changes pleas have a look

ajitsinghkaler:
@gkalpak  why these hooks are failing can you please explain this.

aio/src/app/shared/search-results/search-results.component.html
```diff
@@ -1,30 +1,59 @@
 <div class="search-results">
-  <div *ngIf="searchAreas.length; then searchResults; else notFound"></div>
-</div>
+  <div>
```
gkalpak: This extra `div` seems unnecessary. Can we get rid of it (or does it serve a purpose I missed)?

aio/src/app/shared/search-results/search-results.component.html
```diff
@@ -1,30 +1,59 @@
...
+    <div [ngSwitch]="searchState">
```
gkalpak: Actually, we shouldn't need this `div` either. Can you please move the `ngSwitch` to the top-level `div` instead?

aio/src/app/shared/search-results/search-results.component.html
```diff
@@ -1,30 +1,59 @@
...
+
+      <div *ngSwitchCase="1">
```
gkalpak: Let's avoid unnecessary DOM elements (here and below):

```suggestion
      <ng-container *ngSwitchCase="1">
```

aio/src/app/shared/search-results/search-results.component.html
```diff
@@ -1,30 +1,59 @@
...
+        <p class="not-found">Searching ...</p>
```
gkalpak: Let's change the class name to something more generic (e.g. `no-results`). `not-found` implies that we searched for andf couldn't find them, which is not the case here.

aio/src/app/shared/search-results/search-results.component.html
```diff
@@ -1,30 +1,59 @@
...
+                <span class="symbol {{page.type}}" *ngIf="area.name === 'api'"></span>
+                <span [class.deprecated-api-item]="page.deprecated">{{ page.title }}</span>
+              </a>
+            </li>
+          </ul>
+        </div>
+      </div>
+      
+      <div *ngSwitchCase="3">
+        <div>
```
gkalpak: Again, let's avoid this unnecessary `div` and convert the parent `div` to `ng-container`.

aio/src/app/shared/search-results/search-results.component.html
```diff
@@ -1,30 +1,59 @@
...
+          <p class="not-found">No results found.</p>
+          <p class="not-found">Here are a few links that might be helpful in finding what you are looking for:</p>
```
gkalpak: Wrap these two elements in a parent element with `class="..."` (instead of having the class on each element (just to set their color)).

aio/src/app/shared/search-results/search-results.component.ts
```diff
@@ -23,11 +23,24 @@ export class SearchResultsComponent implements OnChanges {
...
+   * 3 or greater is not found state
+   */
+  searchState = '1';
   readonly topLevelFolders = ['guide', 'tutorial'];
   searchAreas: SearchArea[] = [];
 
   ngOnChanges() {
+    if (this.searchResults == null) {
+      this.searchState = '1';
+    } else if (this.searchResults['results'].length) {
```
gkalpak:
```suggestion
    } else if (this.searchResults.results.length) {
```

aio/src/app/shared/search-results/search-results.component.ts
```diff
@@ -23,11 +23,24 @@ export class SearchResultsComponent implements OnChanges {
...
 
   readonly defaultArea = 'other';
-  notFoundMessage = 'Searching ...';
+  /**
+   * searchState
+   * 1 is searching state
+   * 2 is results found state
+   * 3 or greater is not found state
+   */
+  searchState = '1';
```
gkalpak: Let's use a string enum instead, so that the values are more descriptive (and we get better type-checking):

```ts
enum SearchState {
  InProgress = 'in-progress',
  ResultsFound = 'results-found',
  NoResultsFound = 'no-results-found'
}
```

gkalpak (changes requested):
Thx for implementing this, @ajitsinghkaler! It is much cleaner now.
Can you also update the unit tests to at least verify when the extra links are shown?

ajitsinghkaler:
@gkalpak can you please review this I've made the changes

aio/src/app/shared/search-results/search-results.component.spec.ts
```diff
@@ -214,9 +216,39 @@ describe('SearchResultsComponent', () => {
...
+      expect(getHrefs(anchors)[2]).toContain('guide/glossary');
+    })
+
+    it('should contain cheatsheet reference link', () => {
+      expect(getHrefs(anchors)[3]).toContain('guide/cheatsheet');
+    })
+
+    it('should contain blog reference link', () => {
+      expect(getHrefs(anchors)[4]).toContain('https://blog.angular.io/');
+    })
```
gkalpak: Let's combine all these tests into one:

```ts
const resultLinks = fixture.debugElement.queryAll(By.css('.search-page a'));
const resultHrefs = resultLinks.map(a => a.nativeNode.href);
expect(resultHrefs).toEqual([...]);
```

aio/src/app/shared/search-results/search-results.component.spec.ts
```diff
@@ -214,9 +216,39 @@ describe('SearchResultsComponent', () => {
   });
 
   describe('when no query results', () => {
-    it('should display "not found" message', () => {
+    let anchors: DebugElement[];
```
gkalpak: Since this will only be used on one test (see my other comment), we don't need to define it here.
Same for the `getHrefs()` helper.

gkalpak (changes requested):
Thx for the tests. Can you also add some tests to verify that the generic links are not shown in the other states (i.e. while searching is in progress and when there are results found)?

Also, please, make sure each commit message follows the [commit message guidelines][1] and represents a self-contained unit of work (e.g. tests should be on the same commit as the code changes that are tested).

[1]: https://github.com/angular/angular/blob/939e2f76f/CONTRIBUTING.md#commit

ajitsinghkaler:
@gkalpak I 

> Thx for the tests. Can you also add some tests to verify that the generic links are not shown in the other states (i.e. while searching is in progress and when there are results found)?
> 
> Also, please, make sure each commit message follows the [commit message guidelines](https://github.com/angular/angular/blob/939e2f76f/CONTRIBUTING.md#commit) and represents a self-contained unit of work (e.g. tests should be on the same commit as the code changes that are tested).

I'm just pushing changes right now when we get this finalized I'll squash all the commits into one until them I'm pushing new changes on this branch if that is not much of a problem. Otherwise, I can squash them right now.

gkalpak:
That's OK :+1:

aio/src/app/shared/search-results/search-results.component.html
```diff
@@ -1,30 +1,57 @@
-<div class="search-results">
-  <div *ngIf="searchAreas.length; then searchResults; else notFound"></div>
+<div class="search-results" [ngSwitch]="searchState">
+
+  <ng-container *ngSwitchCase="'in-progress'">
+    <p class="not-found">Searching ...</p>
```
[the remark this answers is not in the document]
ajitsinghkaler: Done

aio/src/app/shared/search-results/search-results.component.html
```diff
@@ -1,30 +1,57 @@
...
+            <span class="symbol {{page.type}}" *ngIf="area.name === 'api'"></span>
+            <span [class.deprecated-api-item]="page.deprecated">{{ page.title }}</span>
+          </a>
+        </li>
+      </ul>
+    </div>
+  </ng-container>
+  
+  <ng-container *ngSwitchCase="'no-results-found'">
+    <div>
```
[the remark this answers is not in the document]
ajitsinghkaler: Yes this is needed if we remove this div the p tag and the ul tag comes side by rather than one below anther because of search results flex

ajitsinghkaler:
@gkalpak made the changes please have a look

aio/src/app/shared/search-results/search-results.component.spec.ts
```diff
@@ -216,39 +220,19 @@ describe('SearchResultsComponent', () => {
...
-      expect(getHrefs(anchors)[3]).toContain('guide/cheatsheet');
-    })
-
-    it('should contain blog reference link', () => {
-      expect(getHrefs(anchors)[4]).toContain('https://blog.angular.io/');
+    it('should contain reference links', () => {
+      const resultLinks = fixture.debugElement.queryAll(By.css('.search-page a'));
+      const resultHrefs = resultLinks.map(a => a.nativeNode.href);
+      expect(resultHrefs.length).toEqual(5);
+      expect(resultHrefs[4]).toEqual("https://blog.angular.io/");
```
gkalpak: Why only test one entry? Let's test all:

```ts
expect(resultHrefs).toEqual([
  ...entry 1...
  ...entry 1...
  ...
]);
```

aio/src/app/shared/search-results/search-results.component.html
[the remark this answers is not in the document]
gkalpak: I see. Then, how about adding the `search-area` class to it?

gkalpak (changes requested):

aio/src/app/shared/search-results/search-results.component.spec.ts
ajitsinghkaler: How do I get base href of the app

gkalpak: Hm...you can just use `a.nativeNode.getAttribute('href')` and check the `href` attributes (which is what we set in the HTML, not the full resolved URL.

ajitsinghkaler:
@gkalpak can you please look at it I made the changes

ajitsinghkaler:
@gkalpak can you please look at it I made the changes

aio/src/app/shared/search-results/search-results.component.spec.ts
```diff
@@ -230,9 +230,15 @@ describe('SearchResultsComponent', () => {
...
-      const resultHrefs = resultLinks.map(a => a.nativeNode.href);
+      const resultHrefs = resultLinks.map(a => a.nativeNode.getAttribute('href'));
       expect(resultHrefs.length).toEqual(5);
-      expect(resultHrefs[4]).toEqual("https://blog.angular.io/");
+      expect(resultHrefs).toEqual([
+        'api', 
+        'resources', 
+        'guide/glossary', 
+        'guide/cheatsheet', 
+        'https://blog.angular.io/'
```
gkalpak: Nit:
```suggestion
        'https://blog.angular.io/',
```

gkalpak (approved):
Thx, @ajitsinghkaler! LGTM - don't forget to squash commits.

ajitsinghkaler:
@gkalpak made all the changes and did the cleanup please have a look

gkalpak:
Hm...this PR seems to increase the main bundle by 1KB. Investigating whether that's expected...

gkalpak:
@googlebot I consent.

gkalpak:
OK, it seems that the increase is expected (and is caused by the changes and extra content in the `aio-search-results` component). For Ivy, it is ~1KB (uncompressed).

I rebased on master, updated payload size limits for current master (to make it easy to see the increase by this PR) and bumped the payload size limits due to this PR's changes.

PR #34980, opened 2020-01-26
build: add missing @babel/generator bazel dep to fix //packages/localize/src/tools/test:test

gregmagolan:
Based on https://github.com/angular/angular/pull/34974 with the addition of the @babel/core dep & the package.json change.

* Also add @babel/generator to root package.json. This doesn't change the yarn.lock since yarn is already hoisting that version the top anyway but it will prevent yarn from not-hoisting in the future.


IgorMinar:
already fixed by #34974

PR #34981, opened 2020-01-26
feat(docs-infra): add stable to the list of api statuses

### LINKED ISSUE #30396: API reference: filter to include the APIs that I should use

### Description

Currently, the API reference can be filtered by:
* All 
* Deprecated
* Security Risk

What I'm most likely to want is this: All the APIs that I **should** use.

It would be nice to have a filter that excludes deprecated and security risks. It might be nice to have that be the default, to help encourage use of non-deprecated and secure APIs. Name ideas: "Current" or "Recommended" or "Stable"?

* Stable
* Deprecated
* Security Risk
* All

If "Experimental" or "Labs" is ever added, this new option would also exclude those, to focus readers on the most stable APIs. 

## 🔬 Minimal Reproduction
See above

`https://angular.io/api`

### Reproduction Steps**

See above

See above

## 📷Screenshot

n/a

## 🔥 Exception or Error
<pre><code>

</code></pre>

ajitsinghkaler:
Earlier api can be filtered based on security risk and deprecated now added stable as a status for better user experience

Fixes #30396

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Earlier only api can be filtered based on deprecated and security
Issue Number: #30396
## What is the new behavior?
Api can be filtered using stable also

- [ ] Yes
- [x] No

### DIFF
--- a/aio/src/app/custom-elements/api/api-list.component.ts
+++ b/aio/src/app/custom-elements/api/api-list.component.ts
@@ -57,6 +57,7 @@ export class ApiListComponent implements OnInit {
 
   statuses: Option[] = [
     { value: 'all', title: 'All' },
+    { value: 'stable', title: 'Stable'},
     { value: 'deprecated', title: 'Deprecated' },
     { value: 'security-risk', title: 'Security Risk' }
   ];


aio/src/app/custom-elements/api/api-list.component.ts
```diff
@@ -117,7 +118,7 @@ export class ApiListComponent implements OnInit {
     const matchesQuery = (item: ApiItem) =>
       sectionNameMatches || item.name.indexOf(query!) !== -1;
     const matchesStatus = (item: ApiItem) =>
-      status === 'all' || status === item.stability || (status === 'security-risk' && item.securityRisk);
+      status === 'all' || status === item.stability || (status === 'security-risk' && item.securityRisk) ||  (status === item.stability && !item.securityRisk);
```
gkalpak: The last condition seems superfluous, since the earlier `status === item.stability` would have matched already.

gkalpak (changes requested):
Thx, @ajitsinghkaler! Can you also add some tests for the new feature?

aio/src/app/custom-elements/api/api-list.component.ts
ajitsinghkaler: I added the last one so that the security risk ones does not mix with the stable ones but you are right this condition is not right.

@gkalpak need your views on this should we add security risk items to the stable filter

ajitsinghkaler: @gkalpak and tests have already been added on line 82-85 it was just that the menu item needed to be added.

If we remove security risk from the stable API then the tests need to be updated otherwise these seem enough to me please have a look at the tests and tell me if any other tests are needed

gkalpak: Yes, security risk is orthogonal to stability, so `stable` should include `securityRisk` items as well. I.e. the condition should remain: `status === 'all' || status === item.stability || (status === 'security-risk' && item.securityRisk)`

Fair enough about the tests. Ideally we should have e2e tests covering this feature, but we don't 😅.
(If you want to add some e2e tests, I won't stop you, but it isn't required for this PR :grin:)

ajitsinghkaler:
@gkalpak made the changes I'll add e2e tests later will make a new issue for that

gkalpak (approved):

PR #34982, opened 2020-01-26
Build/bazel flags cleanup

IgorMinar:
wip

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -65,12 +65,23 @@ test --test_output=errors
 ################################
 # Temporary Settings for Ivy   #
 ################################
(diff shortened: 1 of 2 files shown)


IgorMinar:
hey @josephperrott what do you think about cleaning up the bazel flags and consolidating them in the bazelrc? I came across some issues as I was trying this and I left comments in the diff explaining what they were... the biggest issue seems to be that the build/test_tag_filter flags are not additive so if we specify them in bazelrc and also on the command line or package.json then the bazelrc flags will be overwritten.

jessicajaniuk:
Closing this as it's out of date and no longer necessary.

PR #34995, opened 2020-01-27
fix(language-service): prune duplicate returned definitions

ayazhafiz:
Sometimes, a request for definitions will return multiple of the same
definition. This can happen in at least the cases of

- two-way bindings (one of the same definition for the property and
  event binding)
- multiple template binding expressions in the same attribute
  - something like "*ngFor="let i of items; trackBy: test" has two
    template bindings, resulting in two template binding ASTs at the
    same location (the attribute span). The language service then parses
    both of these bindings individually, resulting in two independent
    but identical definitions. For more context, see https://github.com/angular/angular/pull/34847#discussion_r371006680.

This commit prunes duplicate definitions by signing definitions with
their location, and checking if that location signature has been seen in
a previous definition returned to the client.

## PR Checklist

## PR Type

- [x] Bugfix

- [ ] Yes
- [x] No

### DIFF
--- a/packages/language-service/src/definitions.ts
+++ b/packages/language-service/src/definitions.ts
@@ -39,6 +39,7 @@ export function getDefinitionAndBoundSpan(
     return;
   }
 
+  const seen = new Set<string>();
   const definitions: ts.DefinitionInfo[] = [];
   for (const symbolInfo of symbols) {
     const {symbol} = symbolInfo;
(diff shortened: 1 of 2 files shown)


packages/language-service/src/definitions.ts
```diff
@@ -53,16 +54,28 @@ export function getDefinitionAndBoundSpan(
...
+    for (const {fileName, span} of locations) {
+      const textSpan = ngSpanToTsTextSpan(span);
+      // In cases like two-way bindings, a request for the definitions of an expression may return
+      // two of the same definition:
+      //    [(ngModel)]="prop"
+      //                 ^^^^  -- one definition for the property binding, one for the event binding
+      // To prune duplicate definitions, tag definitions with unique location signatures and ignore
+      // definitions whose locations have already been seen.
+      const signature = `${textSpan.start}:${textSpan.length}@${fileName}`;
+      if (seen.has(signature)) continue;
```
ayazhafiz: @kyliau I don't think we'll be able to use `uniqueBySpan` because we need to sign the filename as well. I can make a util function extending `uniqueBySpan` in either this or a future PR.

kyliau (approved):

PR #34996, opened 2020-01-27
ci: prevent benchmark_tests from executing on RBE

josephperrott:
Since we don't do anything with the results of the benchmark_test
executions on CI, we should not run them.  Additionally, since
benchmarks are meant to test in a consistent environment, we should
execute the benchmark tests locally rather than on RBE executors.

### DIFF
--- a/modules/benchmarks/benchmark_test.bzl
+++ b/modules/benchmarks/benchmark_test.bzl
@@ -17,7 +17,13 @@ def benchmark_test(name, server, tags = [], **kwargs):
         on_prepare = "//modules/benchmarks:start-server.js",
         server = server,
         # Benchmark targets should not run on CI by default.
-        tags = tags + ["manual"],
-        test_suite_tags = ["manual"],
+        tags = tags + [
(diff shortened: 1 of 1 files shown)


devversion:
We've talked about this in the framework meeting in the past, and decided to actually not disable benchmarks with their functionality tests. There is a PR for that: #34753

pkozlowski-opensource:
Yeh, I would like to side with @devversion here - the issue is perf benchmarks (as written today) also act as e2e test (making sure that the scenario is actually valid). #34753 contains the work to split e2e (correctness) tests from perf tests (both are written using the same tech but have different purpose.

After the split proposed in #34753 we could disable the perf side of things but keep the verification (e2e) part running on CI.

josephperrott:
@devversion @pkozlowski-opensource I updated to my PR to include the change that Paul's PR didn't cover.  PTAL

devversion (approved):
LGTM. thanks for helping with the other PR!

PR #35001, opened 2020-01-28
fix(ngcc): improve lockfile error message

### LINKED ISSUE #35000: Allow specifying custom NGCC lockfile location (cannot build in parallel with Ivy)
# 🚀 feature request

### Relevant Package

 This feature request is for @angular/compiler-cli

### Description
I am running multiple webpack build pipelines simultaneously  in CI. It does work without Ivy, but with Ivy (Angular 9 rc 11) it fails with:
```
(If you are sure no ngcc process is running then you should delete the lockfile at /drone/src/github.com/Rush/angular-app/node_modules/@angular/compiler-cli/ngcc/__ngcc_lock_file__.)
```

This line should allow overwriting the lock file location:
https://github.com/angular/angular/blob/d15cf60c4969c4da081d0efdbc818abfd3fe43eb/packages/compiler-cli/ngcc/src/execution/lock_file.ts#L20
Perhaps like so:
```
 lockFilePath = process.env.ANGULAR_NGCC_LOCK_FILE_PATH || 
      this.fs.resolve(require.resolve('@angular/compiler-cli/ngcc'), '../__ngcc_lock_file__');
```

I think this is safe cause all simultaneous compiles are happening in Webpack so they're not conflicting with each other. @petebacondarwin @AndrewKushnir 

Not having this would cause my builds running 10x more slowly due to a lot of parallelism.

Disabling the lock file with a config option.

petebacondarwin:
The message now gives concrete advice to developers who
experience the error due to running multiple simultaneous builds
via webpack.

Fixes #35000

### DIFF
--- a/packages/compiler-cli/ngcc/src/execution/lock_file.ts
+++ b/packages/compiler-cli/ngcc/src/execution/lock_file.ts
@@ -75,6 +75,8 @@ export class LockFile {
 
       throw new Error(
           `ngcc is already running at process with id ${pid}.\n` +
+          `If you are running multiple builds (e.g. webpack) in parallel then you should pre-process your node_modules via the command line ngcc tool before starting the builds;\n` +
+          `See https://v9.angular.io/guide/ivy#speeding-up-ngcc-compilation.\n` +
           `(If you are sure no ngcc process is running then you should delete the lockfile at ${this.lockFilePath}.)`);
     }
   }
--- a/packages/compiler-cli/ngcc/src/execution/lock_file.ts
+++ b/packages/compiler-cli/ngcc/src/execution/lock_file.ts
@@ -75,7 +75,7 @@ export class LockFile {
 
       throw new Error(
           `ngcc is already running at process with id ${pid}.\n` +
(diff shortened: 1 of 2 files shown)


packages/compiler-cli/ngcc/src/execution/lock_file.ts
```diff
@@ -75,6 +75,8 @@ export class LockFile {
 
       throw new Error(
           `ngcc is already running at process with id ${pid}.\n` +
+          `If you are running multiple builds (e.g. webpack) in parallel then you should pre-process your node_modules via the command line ngcc tool before starting the builds;\n` +
```
gkalpak: I think `webpack` might be confusing here. As far as I understand, the parallel builds are most probably run by the cli (and the fact that it runs webpack under the hood is an implementation detail).

gkalpak (approved):

packages/compiler-cli/ngcc/src/execution/lock_file.ts
petebacondarwin: Shall we just remove the webpack mention then?

PR #30130, opened 2019-04-25
fix(ivy): remove debug utilities from ivy production builds

kara:
Prior to this commit, we were pulling DebugNode and DebugElement
into production builds because BrowserModule automatically pulled
in NgProbe and thus getDebugNode. In Ivy, this is not necessary
because Ivy has its own set of debug utilities. We should use these
existing tools instead of NgProbe.

This commit adds an Ivy switch so we do not pull in NgProbe utilities
when running with Ivy. This saves us ~8KB in prod builds.

Note: Master is red for unrelated size regressions :-(

PR #35003, opened 2020-01-28
fix(ivy): ensure `DebugNode`/`DebugElement` are tree-shakeable in Ivy

gkalpak:
There are different `DebugNode`/`DebugElement` implementations (and associated helper functions) for ViewEngine and Ivy. Additionally, these classes/functions, which are defined inside the `core` package, are imported by the `platform-browser` package.

Previously, this code was not tree-shaken as expected in Ivy. #30130 partially addressed the issue, but only for the case where `core` and `platform-browser` end up in the same closure after webpack's scope hoisting. In cases where this is not the case, our webpack/terser based tooling is not capable of tree-shaking it.

This commit fixes the problem, by ensuring that the code retained in Ivy mode (due to the cross-package import) does not unnecessarily reference `DebugNode`/`DebugElement`, allowing the code to be tree-shaken away. This results in a 7.6KB reduction in the size of the main angular.io
bundle.

Jira issue: [FW-1802](https://angular-team.atlassian.net/browse/FW-1802)


packages/core/src/debug/debug_node.ts
```diff
@@ -727,6 +727,18 @@ export function getDebugNode__POST_R3__(nativeNode: any): DebugNode|null {
...
+
+export function getDebugNodeR2__PRE_R3__(nativeNode: any): DebugNode|null {
+  return getDebugNode__PRE_R3__(nativeNode);
+}
+
+export function getDebugNodeR2__POST_R3__(_nativeNode: any): DebugNode|null {
+  return null;
+}
+
+export const getDebugNodeR2: (nativeNode: any) => DebugNode | null = getDebugNodeR2__PRE_R3__;
```
gkalpak: Not sure what would be an appropriate name :innocent:

aio/scripts/_payload-limits.json
```diff
@@ -12,7 +12,7 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 2987,
-        "main-es2015": 456604,
+        "main-es2015": 448956,
```
IgorMinar: woot!!!

integration/_payload-limits.json
```diff
@@ -39,7 +39,7 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 2289,
-        "main-es2015": 254657,
+        "main-es2015": 247225,
```
IgorMinar: sweet!

packages/core/src/debug/debug_node.ts
IgorMinar: lgtm

IgorMinar (approved):
George, you rock!

thanks for figuring this out.

❤️

PR #35008, opened 2020-01-28
build(localize): sync up babel dependencies

petebacondarwin:
In #34974 the top level dependency on `@babel/core` was bumped to
7.8.3. This commit ensures that the package.json that gets included
in the `@angular/localize` distributable is at the same version.

### DIFF
--- a/packages/localize/package.json
+++ b/packages/localize/package.json
@@ -32,7 +32,7 @@
     "node": ">=8.0"
   },
   "dependencies": {
-    "@babel/core": "^7.5.5",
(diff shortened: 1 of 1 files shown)


AndrewKushnir (approved):
LGTM 👍

petebacondarwin:
Merge-assistance: I am a member of the `angular/framework-global-approvers-for-docs-only-changes` code-owners group. (And I think that code-owners are no longer relevant given that we have moved over to pullapprove?)

PR #23015, opened 2018-03-27
docs(common): add `HttpParamsOptions` to the public API (#20332)

gkalpak:
Resubmitting #20332, since it was reverted (due to an [error in g3][1]).

## PR Checklist
- [x] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [ ] ~~Tests for the changes have been added (for bug fixes / features)~~
- [x] Docs have been added / updated (for bug fixes / features)

## PR Type
```
[x] Documentation content changes
```

## What is the current behavior?
`HttpParamsOptions` is not documented as part of the public API.
Issue Number: #20276

## What is the new behavior?
`HttpParamsOptions` is documented as part of the public API.

## Does this PR introduce a breaking change?
```
[ ] Yes
[x] No
```

## Other information
Fixes #20276.

[1]: https://github.com/angular/angular/pull/20332#issuecomment-364574160

PR #35015, opened 2020-01-28
ci: minor PullApprove fixes/improvements

gkalpak:
This is a follow-up to #34814 to fix some typos in patterns and make them more similar to the old patterns from `.github/CODEOWNERS`.

### DIFF
--- a/.pullapprove.yml
+++ b/.pullapprove.yml
@@ -183,23 +183,6 @@ groups:
 
 
   # =========================================================
-  #  Framework: Compiler + CLI integration
-  # =========================================================
-  fw-cli-integration:
-    conditions:
-      - >
-        contains_any_globs(files, [
-          'packages/compiler-cli/src/ngtools/**',
-          ])
-    reviewers:
-      users:
-        - filipesilva
-      teams:
-        - ~framework-global-approvers
-        - ~framework-global-approvers-for-docs-only-changes
-
-
-  # =========================================================
   #  Framework: Core
   # =========================================================
   fw-core:
@@ -893,24 +876,29 @@ groups:
       - >
         contains_any_globs(files, [
           '*',
-          '.buildkite/**',
           '.circleci/**',
           '.devcontainer/**',
           '.github/**',
           '.vscode/**',
+          '.yarn/**',
           'docs/BAZEL.md',
           'packages/*',
           'packages/examples/test-utils/**',
           'packages/private/**',
           'scripts/**',
           'third_party/**',
+          'tools/brotli-cli/**',
+          'tools/browsers/**',
           'tools/build/**',
+          'tools/circular_dependency_test/**',
           'tools/gulp-tasks/**',
+          'tools/ng_rollup_bundle/**',
           'tools/ngcontainer/**',
           'tools/npm/**',
-          'tools/public_api_guard/BUILD',
+          'tools/public_api_guard/BUILD.bazel',
           'tools/public_api_guard/public_api_guard.bzl',
           'tools/rxjs/**',
+          'tools/saucelabs/**',
           'tools/size-tracking/**',
           'tools/source-map-test/**',
           'tools/symbol-extractor/**',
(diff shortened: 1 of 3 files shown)


scripts/github/merge-pr
```diff
@@ -131,8 +131,8 @@ CHERRY_PICK_PR="git cherry-pick merge_pr_base..merge_pr"
 #
 # This check is used to enforce that we don't merge PRs that have not been rebased recently and could result in merging
 # of non-approved or otherwise bad changes.
-REQUIRED_BASE_SHA_MASTER="a03a9236f2aed5d00012d25f032aa43a046d91da" # pullapprove => CODEOWNERS migration
-REQUIRED_BASE_SHA_PATCH="a03a9236f2aed5d00012d25f032aa43a046d91da"  # pullapprove => CODEOWNERS migration
+REQUIRED_BASE_SHA_MASTER="296dc0622f0e8c4e803ff4f19a5c6fe02a2ae66e" # CODEOWNERS => PullApprove migration
+REQUIRED_BASE_SHA_PATCH="110f6c91b904819cab639861b54b6a989e176942"  # CODEOWNERS => PullApprove migration
```
devversion: Out of curiosity: @josephperrott mentioned in #fw-caretaker that pullapprove walks through PRs in the backlog to assign revieweers. Does that mean that we don't need to bump the SHA here? I could see the script incorrectly blocking old PRs that have been updated by pullapprove?

.pullapprove.yml
```diff
@@ -73,6 +76,8 @@
 # =========================================================
 # @angular/framework-global-approvers
 # =========================================================
+#  Used for approving minor changes, large-scale refactorings, and emergency situations.
```
devversion:
```suggestion
#  Used for approving minor changes, large-scale refactorings, and in emergency situations.
```

devversion (approved):
LGTM. One comment proposal, and a question.

scripts/github/merge-pr
gkalpak: Not sure either, so I opted to err on the side of caution :grin:

For example, this super old PR, #23015, still has no PullApprove status, so not sure what PullApprove is doing. Or maybe in these cases the PR will be on CODEOWNERS, so they will have the necessary reviews requested.

@josephperrott, wdyt?

josephperrott: After talking with PullApprove a bit, it looks like only the most recent 200 are immediately processed when the config changes.  We should include this required base update.

.pullapprove.yml
```diff
@@ -913,8 +920,8 @@ groups:
           'tools/validate-commit-message/**',
           'tools/yarn/**',
           'tools/*',
-          '*.bzl',
-          '*.BAZEL'
+          '**/*.bzl',
```
josephperrott: Can you also:
**remove:**
  - `.buildkite/**`

**change:**
  - `tools/public_api_guard/BUILD` -> `tools/public_api_guard/BUILD.bazel`
  - `**/*.BAZEL` -> `**/*.bazel`

**add:**
  - `.yarn/**`
  - `tools/brotli-cli/**`
  - `tools/browsers/**`
  - `tools/circular_dependency_test/**`
  - `tools/ng_rollup_bundle/**`
  - `tools/saucelabs/**`

josephperrott (changes requested):
Can you also remove the `fw-cli-integration` group as its glob does not match any files in the repo.

.pullapprove.yml
IgorMinar: all of this SGTM

.pullapprove.yml
```diff
@@ -934,7 +941,7 @@ groups:
     conditions:
       - >
         contains_any_globs(files, [
-          'tools/material-ci/*'
+          'tools/components-repo-ci/**'
```
IgorMinar: ah! good catch.

scripts/github/merge-pr
IgorMinar: this change is sooooo ironic :-D

IgorMinar (approved):
lgtm after cleanup. thanks for all these fixes George!

.pullapprove.yml
gkalpak: Done.

josephperrott (approved):
LGTM

PR #35033, opened 2020-01-29
ci(docs-infra): ignore node version in `aio_monitoring_stable` CI job (for real)

gkalpak:
In #35004, we started ignoring yarn's engines check for `yarn install` in AIO's `test-production.sh` script to fix a failure in the `aio_monitoring_stable` CI job. (See #35004 for details.)

It turns out that the version of yarn used on the stable branch (1.17.3) `--ignore-engines` is needed on all yarn commands (including `yarn run`). Thus, #35004 is not enough to fix the failures.
New example failure: https://circleci.com/gh/angular/angular/604341

This commit turns of the engines check for the whole `aio_monitoring_stable` CI job to fix the failure and make the job more robust.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -623,6 +623,8 @@ jobs:
             git checkout --force origin/$CI_STABLE_BRANCH -- aio/ .yarn/ .yarnrc
         # Overwrite yarn again to use the version from the checked out branch.
       - overwrite_yarn
+        # Since we check out `aio/package.json` from the stable branch, ignore yarn's engines check.
+      - run: yarn config set ignore-engines true
       - run:
           name: Run tests against https://angular.io/
           command: ./aio/scripts/test-production.sh https://angular.io/ $CI_AIO_MIN_PWA_SCORE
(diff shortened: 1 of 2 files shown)


gkalpak:
You can see a successful run of the `aio_monitoring_stable` job (with the changes from this PR) [here](https://circleci.com/gh/angular/angular/604376).

<sub>NOTE: This change is not required on the 8.2.x branch, since the job runs with the CircleCI config from master.</sub>

devversion (approved):
LGTM

.circleci/config.yml
```diff
@@ -623,6 +623,8 @@ jobs:
             git checkout --force origin/$CI_STABLE_BRANCH -- aio/ .yarn/ .yarnrc
         # Overwrite yarn again to use the version from the checked out branch.
       - overwrite_yarn
+        # Since we check out `aio/package.json` from the stable branch, ignore yarn's engines check.
```
IgorMinar:
```suggestion
        # Ignore yarn's engines check, because we checked out `aio/package.json` from the stable branch and there could be a node version skew, which is acceptable in this monitoring job.
```

IgorMinar (approved):
please take a look at the suggestion, otherwise lgtm

.circleci/config.yml
gkalpak: Done.

PR #35035, opened 2020-01-29
ci: update commit message linting

### LINKED ISSUE #35020: Do not run "validate-commit-message" lint check for upstream branches
Currently the `validate-commit-message` lint check is executed for PRs as well as upstream branches (master, 9.0.x, 8.2.x). There is no need to run this check for upstream branches, since all the necessary validation happens at PR stage. Note: all other lint checks (like formatting) should still be executed in upstream branches.

### LINKED ISSUE #35021: "validate-commit-message" lint check should only verify commits that belong to a given PR
Currently the "validate-commit-message" lint check for PRs verify all commits from v9.0.x to HEAD (which is ~850 commits atm). Instead, the "validate-commit-message" lint check should only verify messages for commits that belong to a given PR.

josephperrott:
See commits for specific changes, sets linting to only check commit messages on PRs and only checks commits from the PR.

Fixes #35020 and Fixes #35021

### DIFF
--- a/tools/gulp-tasks/validate-commit-message.js
+++ b/tools/gulp-tasks/validate-commit-message.js
@@ -37,10 +37,10 @@ module.exports = (gulp) => async () => {
 
     const commitsByLine = result.trim().split(/\n/).filter(line => line != '');
 
-    console.log(`Examining ${commitsByLine.length} commit(s) between ${baseBranch} and HEAD`);
+    console.log(`Examining ${commitsByLine.length} commit(s) between ${target.base.ref} and HEAD`);
 
     if (commitsByLine.length == 0) {
-      console.log(`There are zero new commits between ${baseBranch} and HEAD`);
+      console.log(`There are zero new commits between ${target.base.ref} and HEAD`);
     }
 
     const disallowSquashCommits = true;
--- a/tools/gulp-tasks/validate-commit-message.js
+++ b/tools/gulp-tasks/validate-commit-message.js
@@ -8,19 +8,18 @@
 
 
 // tslint:disable:no-console
-module.exports = (gulp) => async () => {
+module.exports = (gulp) => async() => {
   try {
     if (process.env['CIRCLECI'] === 'true' && !process.env['CIRCLE_PR_NUMBER']) {
       console.info(
-        `Since commit messages are validated as part of the PR review process,\n` +
-        `we do not need to commit messages on CI runs on upstream branches.\n\n` +
-        `Skipping validate-commit-message check`
-        )
+          `Since commit messages are validated as part of the PR review process,\n` +
+          `we do not need to commit messages on CI runs on upstream branches.\n\n` +
+          `Skipping validate-commit-message check`);
       process.exit();
     }
     const validateCommitMessage = require('../validate-commit-message');
     const shelljs = require('shelljs');
-    const getRefsAndShasForTarget = require('../utils/get-refs-and-shas-for-target')
+    const getRefsAndShasForTarget = require('../utils/get-refs-and-shas-for-target');
 
     shelljs.set('-e');  // Break on error.
 
--- a/tools/gulp-tasks/validate-commit-message.js
+++ b/tools/gulp-tasks/validate-commit-message.js
@@ -22,6 +22,9 @@ module.exports = (gulp) => async() => {
     if (process.env['CI'] === 'true') {
       // Validation of commit messages on CI
       if (process.env['CI_PULL_REQUEST'] === 'false') {
+        // Skip commit message validation on CI for non-PR branches as we are not testing new
+        // unreviewed commits.  By enforcing correctness on the incoming changes in the PR
+        // branches, we are able to render this check unnecessary on non-PR branches.
         console.info(
             `Since valid commit messages are enforced by PR linting on CI,\n` +
             `we do not need to validate commit messages on CI runs on upstream branches.\n\n` +
--- a/tools/gulp-tasks/validate-commit-message.js
+++ b/tools/gulp-tasks/validate-commit-message.js
@@ -10,10 +10,10 @@
 // tslint:disable:no-console
 module.exports = (gulp) => async() => {
   try {
-    if (process.env['CIRCLECI'] === 'true' && !process.env['CIRCLE_PR_NUMBER']) {
+    if (process.env['CI'] === 'true' && process.env['CI_PULL_REQUEST'] === 'false') {
       console.info(
-          `Since commit messages are validated as part of the PR review process,\n` +
-          `we do not need to commit messages on CI runs on upstream branches.\n\n` +
+          `Since valid commit messages are enforced by PR linting on CI,\n` +
+          `we do not need to validate commit messages on CI runs on upstream branches.\n\n` +
           `Skipping validate-commit-message check`);
       process.exit();
     }
--- a/tools/gulp-tasks/validate-commit-message.js
+++ b/tools/gulp-tasks/validate-commit-message.js
@@ -10,25 +10,54 @@
 // tslint:disable:no-console
 module.exports = (gulp) => async() => {
   try {
-    if (process.env['CI'] === 'true' && process.env['CI_PULL_REQUEST'] === 'false') {
-      console.info(
-          `Since valid commit messages are enforced by PR linting on CI,\n` +
-          `we do not need to validate commit messages on CI runs on upstream branches.\n\n` +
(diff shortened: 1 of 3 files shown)


tools/gulp-tasks/validate-commit-message.js
```diff
@@ -8,41 +8,38 @@
...
-        shelljs
-            .exec(`git ls-remote --heads origin ${currentVersion.major}.${currentVersion.minor}.*`)
-            .trim()
-            .split('\n')
-            .pop();
-    if (baseHead) {
-      const match = /refs\/heads\/(.+)/.exec(baseHead);
-      baseBranch = match && match[1] || baseBranch;
-    }
+    const target = await getRefsAndShasForTarget(process.env['CIRCLE_PR_NUMBER']);
```
gkalpak: What about running the script locally (where `CIRCLE_PR_NUMBER` is not defined)?
IMO, we should still allow this (or at least make sure that `gulp lint` does not fail locally).

tools/gulp-tasks/validate-commit-message.js
```diff
@@ -8,41 +8,38 @@
 
 
 // tslint:disable:no-console
-module.exports = (gulp) => () => {
+module.exports = (gulp) => async() => {
   try {
+    if (process.env['CIRCLECI'] === 'true' && !process.env['CIRCLE_PR_NUMBER']) {
```
gkalpak: 1. _Regarding `CIRCLECI`:_
  Consider using `CI` instead, because it is available in all popular CI environments (so more portable than `CIRCLECI`) and we use it in other places (such as [here][1] and [there][2]).

2. _Regarding `CIRCLE_PR_NUMBER`:_
  We should avoid using CircleCI-specific environment variables outside the `.circleci/` directory s it makes our scripts less portable. (This was a significant pain point when migrating from TravisCI to CircleCI.) Instead, we should use variables exported by the `.circleci/env.sh` script; in this case [CI_PULL_REQUEST][3]. 

[1]: https://github.com/angular/angular/blob/a3de110b5/tools/check-environment.js#L106
[2]: https://github.com/angular/angular/blob/a3de110b5/integration/run_tests.sh#L6
[3]: https://github.com/angular/angular/blob/a3de110b5/.circleci/env.sh#L35

tools/gulp-tasks/validate-commit-message.js
```diff
@@ -8,41 +8,38 @@
...
+      console.info(
+          `Since commit messages are validated as part of the PR review process,\n` +
```
gkalpak: > as part of the PR review process

This sounds like the reviewer is responsible for validating the commit messages :grin:

tools/gulp-tasks/validate-commit-message.js
```diff
@@ -8,41 +8,38 @@
...
+          `we do not need to commit messages on CI runs on upstream branches.\n\n` +
```
gkalpak:
```suggestion
          `we do not need to validate commit messages on CI runs on upstream branches.\n\n` +
```

gkalpak (changes requested):
Yay for re-usability 👌

tools/gulp-tasks/validate-commit-message.js
josephperrott: Migrated to using `CI` AND `CI_PULL_REQUEST`.

tools/gulp-tasks/validate-commit-message.js
```diff
@@ -8,41 +8,67 @@
...
+      const {stdout: headRef} = shelljs.exec(
+          'git branch 2> /dev/null | sed -e \'/^[^*]/d\' -e \'s/* \(.*\)/\1/\'', {silent: true});
+      const {stdout: baseSha} = shelljs.exec(`git rev-parse origin/master`, {silent: true});
+      const {stdout: headSha} = shelljs.exec(`git rev-parse HEAD`, {silent: true});
+      const {stdout: commonAncestorSha} =
+          shelljs.exec(`git merge-base origin/master ${headSha}`, {silent: true});
+      target = {
+        base: {
+          ref: baseRef,
+          sha: baseSha,
```
gkalpak:
```suggestion
          sha: baseSha.trim(),
```

tools/gulp-tasks/validate-commit-message.js
```diff
@@ -8,41 +8,67 @@
...
+          sha: headSha.trim(),
+        },
+        commonAncestorSha: commonAncestorSha.trim(),
+        latestShaOfTargetBranch: baseSha.trim(),
+        latestShaOfPrBranch: headSha.trim(),
+      };
     }
 
+
+    shelljs.set('-e');  // Break on error.
```
gkalpak: Shouldn't this be above the if block above (since it now runs `shelljs.exec()` too)?

gkalpak (approved):
A couple of minor comments; otherwise LGTM :+1:
This should be much more reliable now :tada:

One thing to keep in mind:
Making GitHub API requests without a GitHub token has a relatively low rate limit. If we start seeing rate-limit errors, we could consider making one call per PR (e.g. in the `setup` job) and share the results with all subsequent jobs.

tools/gulp-tasks/validate-commit-message.js
```diff
@@ -8,41 +8,67 @@
...
+
     // We need to fetch origin explicitly because it might be stale.
     // I couldn't find a reliable way to do this without fetch.
```
devversion: This comment seems to be out of date now. Should we update/remove it?

tools/gulp-tasks/validate-commit-message.js
```diff
@@ -8,41 +8,67 @@
...
+            `we do not need to validate commit messages on CI runs on upstream branches.\n\n` +
+            `Skipping validate-commit-message check`);
+        process.exit();
+      }
+      target = await getRefsAndShasForTarget(process.env['CIRCLE_PR_NUMBER']);
+    } else {
+      // Validation of commit messages locally
+      const baseRef = 'master';
+      const {stdout: headRef} = shelljs.exec(
+          'git branch 2> /dev/null | sed -e \'/^[^*]/d\' -e \'s/* \(.*\)/\1/\'', {silent: true});
```
devversion: This command seems to fail for me on Windows. Mostly because of the stdout/stderr redirect, and because of the pipe. IMO `shelljs.exec` should just run a single command (to ensure it works on windows too), and any output transformation should be made in JS. 

In general, the goal is just to get the current branch name, right? In `angular/components` we use [the following command](https://github.com/angular/components/blob/master/tools/release/git/git-client.ts#L25). Could that be used here too?

josephperrott: Yes, your command is a much better option.  I honestly just took this from my .bashrc where I put the git branch in my terminal prompt.

tools/gulp-tasks/validate-commit-message.js
josephperrott: I found that I actually can do this on the `shelljs.exec` call because the `ShellString` that if returns is just a `string` that has properties added on it.

tools/gulp-tasks/validate-commit-message.js
```diff
@@ -8,41 +8,64 @@
...
-            .split('\n')
-            .pop();
-    if (baseHead) {
-      const match = /refs\/heads\/(.+)/.exec(baseHead);
-      baseBranch = match && match[1] || baseBranch;
+
+    let target = {};
+    if (process.env['CI'] === 'true') {
+      // Validation of commit messages on CI
+      if (process.env['CI_PULL_REQUEST'] === 'false') {
```
devversion: I think we should add a comment on why we don't want to run for master. what do you think?

devversion (approved):
LGTM.

tools/gulp-tasks/validate-commit-message.js
IgorMinar: done in the console message bellow

IgorMinar (approved):
this is excellent! one thing we could consider in the future is populating the CI_COMPARE_URL using the newly extracted script so that we can use it elsewhere easily without having to make all the http calls.

IgorMinar:
I rebased this PR

tools/gulp-tasks/validate-commit-message.js
devversion: Right. My thinking was that it can be just part of the comment above that previously didn't provide any meaningful context. I agree though that it's not necessary. Joey added a comment anyway.. and it shouldn't hurt 😄

PR #35036, opened 2020-01-29
fix(language-service): suggest, not error, on missing context members

### LINKED ISSUE #572: The template context does not define a member called 'item'
**Describe the bug**

Not sure if this issue is related to this extension, but the code does compile and work even if the language service gives me an error in the console.

In a component.html file I have this:
```
<ng-template ng-label-tmp ng-option-tmp let-item="item">
    {{ item.name }} -- {{ item.description }} -- {{ item.price | currency }}
</ng-template>
```
the `let-item="item"` piece is red highlighted and this is the description:

> The template context does not define a member called 'item'

**To Reproduce**
Use any `*-tmp` directives along with `let-*` ones.

**Expected behavior**
No Errors should be highlighted for the `let-item="item"` directive.

**Logs**
1. Console output
```
[Info  - 17:40:22] Angular language server process ID: 21305
[Info  - 17:40:22] Using typescript v3.7.4 from /home/user/.vscode/extensions/angular.ng-template-0.900.5/node_modules/typescript/lib/tsserverlibrary.js
[Info  - 17:40:24] Using @angular/language-service v9.0.0-rc.9 from /home/user/.vscode/extensions/angular.ng-template-0.900.5/server/node_modules/@angular/language-service/bundles/language-service.umd.js
[Info  - 17:40:25] Log file: /home/user/.config/Code/logs/20200123T153034/exthost9/Angular.ng-template/nglangsvc.log
[Error - 17:40:39] No config file for /home/user/git/projects/my-app/src/app/ui-widgets/app-card/add-item-card/add-item-card.component.html
```

2. Log file 
[nglangsvc.log](https://github.com/angular/vscode-ng-language-service/files/4104555/nglangsvc.log)

**Additional context**
If I remove  both `ng-label-tmp` and `ng-option-tmp` directives the error disappears.

**UPDATE**
Downgrading the extension to `v0.900.4` the message error disappears.
The `v0.900.5` starts to show this error.

ayazhafiz:
The language service reports an error when a directive's template
context is missing a member that is being used in a template (e.g. if
`$implicit` is being used with a template context typed as `any`).
While this diagnostic message is valuable, typing template contexts
loosely as `any` or `object` is very widespread in community packages,
and often still compiles correctly, so reporting the diagnostic as an
error may be misleading to users.

This commit changes the diagnostic to be a warning, and adds additional
information about how the user can eliminate the warning entirely -- by
refining the template context type.

A screenshot demonstrating the change with VSCode's LSP is attached to
the PR for this commit.

Screen Shot 2020-01-29 at 10 17 15 AM

Closes https://github.com/angular/vscode-ng-language-service/issues/572

## PR Checklist

## PR Type

- [x] Bugfix
- [x] Feature

- [ ] Yes
- [x] No

### DIFF
--- a/packages/language-service/src/expression_diagnostics.ts
+++ b/packages/language-service/src/expression_diagnostics.ts
@@ -292,16 +292,12 @@ class ExpressionDiagnosticsVisitor extends RecursiveTemplateAstVisitor {
     if (directive && ast.value) {
       const context = this.info.query.getTemplateContext(directive.type.reference) !;
       if (context && !context.has(ast.value)) {
-        if (ast.value === '$implicit') {
-          this.reportError(
-              `The template context of '${directive.type.reference.name}' does not define an implicit value.\n` +
-                  `If the context type is a base type, consider refining it to a more specific type.`,
-              spanOf(ast.sourceSpan));
-        } else {
-          this.reportError(
-              `The template context of '${directive.type.reference.name}' does not define a member called '${ast.value}'`,
-              spanOf(ast.sourceSpan));
-        }
+        const missingMember =
+            ast.value === '$implicit' ? 'an implicit value' : `a member called '${ast.value}'`;
+        this.reportError(
+            `The template context of '${directive.type.reference.name}' does not define ${missingMember}.\n` +
+                `If the context type is a base type, consider refining it to a more specific type.`,
+            spanOf(ast.sourceSpan), ts.DiagnosticCategory.Warning);
       }
     }
   }
@@ -353,10 +349,11 @@ class ExpressionDiagnosticsVisitor extends RecursiveTemplateAstVisitor {
 
   private pop() { this.path.pop(); }
 
-  private reportError(message: string, span: Span|undefined) {
+  private reportError(
+      message: string, span: Span|undefined,
+      kind: ts.DiagnosticCategory = ts.DiagnosticCategory.Error) {
     if (span) {
-      this.diagnostics.push(
-          {span: offsetSpan(span, this.info.offset), kind: ts.DiagnosticCategory.Error, message});
+      this.diagnostics.push({span: offsetSpan(span, this.info.offset), kind, message});
     }
   }
 }
--- a/packages/language-service/src/expression_diagnostics.ts
+++ b/packages/language-service/src/expression_diagnostics.ts
@@ -294,10 +294,10 @@ class ExpressionDiagnosticsVisitor extends RecursiveTemplateAstVisitor {
       if (context && !context.has(ast.value)) {
         const missingMember =
             ast.value === '$implicit' ? 'an implicit value' : `a member called '${ast.value}'`;
-        this.reportError(
+        this.reportDiagnostic(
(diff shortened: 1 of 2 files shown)


packages/language-service/src/expression_diagnostics.ts
```diff
@@ -353,10 +349,11 @@ class ExpressionDiagnosticsVisitor extends RecursiveTemplateAstVisitor {
 
   private pop() { this.path.pop(); }
 
-  private reportError(message: string, span: Span|undefined) {
+  private reportError(
+      message: string, span: Span|undefined,
```
kyliau: Unrelated to the issue addressed by this PR, but good to clean up while you're here:
This method is only used in two places, and both call sites provide a valid span, so `undefined` is not necessary.
The if check below can also be removed.

packages/language-service/src/expression_diagnostics.ts
```diff
@@ -353,10 +349,11 @@ class ExpressionDiagnosticsVisitor extends RecursiveTemplateAstVisitor {
 
   private pop() { this.path.pop(); }
 
-  private reportError(message: string, span: Span|undefined) {
+  private reportError(
```
kyliau: Since this method now accepts `kind`, perhaps it's more meaningful to name it `reportDiagnostic`?

packages/language-service/src/expression_diagnostics.ts
```diff
@@ -292,16 +292,12 @@ class ExpressionDiagnosticsVisitor extends RecursiveTemplateAstVisitor {
...
-          this.reportError(
-              `The template context of '${directive.type.reference.name}' does not define a member called '${ast.value}'`,
-              spanOf(ast.sourceSpan));
-        }
+        const missingMember =
+            ast.value === '$implicit' ? 'an implicit value' : `a member called '${ast.value}'`;
+        this.reportError(
+            `The template context of '${directive.type.reference.name}' does not define ${missingMember}.\n` +
+                `If the context type is a base type, consider refining it to a more specific type.`,
+            spanOf(ast.sourceSpan), ts.DiagnosticCategory.Warning);
```
kyliau: Thank you for attaching the screenshot, it is exactly what I wanted to verify.
Given that the compiler does not complain about the error whatsoever (it can be `blah` and the compilation will still pass), I'm afraid making this a warning is not enough. Users will still find it confusing due to the way a warning is presented in the IDE.
Perhaps it is better to reword this message and make it a Suggestion instead.
```
It appears the template context of ${directive...} does not define ${missingMember}.
If the context type is a base type or any, consider refining it to a more specific type.
```
What do you think?

kyliau (changes requested):

packages/language-service/src/expression_diagnostics.ts
ayazhafiz: I tried this out; at least in VSCode, the suggestion is not very obvious. I agree though, it is more of a suggestion that anything else. I will fix this.

ayazhafiz:
@kyliau updated screen:

Screen Shot 2020-01-29 at 11 23 21 AM

kyliau (approved):

PR #35037, opened 2020-01-29
fix(ivy): PoC draft - containsdirty

atscott:
Only check transplanted views at insertion point

### DIFF
--- a/packages/core/src/render3/instructions/shared.ts
+++ b/packages/core/src/render3/instructions/shared.ts
@@ -28,7 +28,7 @@ import {AttributeMarker, InitialInputData, InitialInputs, LocalRefExtractor, Pro
 import {RComment, RElement, RNode, RText, Renderer3, RendererFactory3, isProceduralRenderer} from '../interfaces/renderer';
 import {SanitizerFn} from '../interfaces/sanitization';
 import {isComponentDef, isComponentHost, isContentQueryHost, isLContainer, isRootView} from '../interfaces/type_checks';
-import {CHILD_HEAD, CHILD_TAIL, CLEANUP, CONTEXT, DECLARATION_COMPONENT_VIEW, DECLARATION_VIEW, FLAGS, HEADER_OFFSET, HOST, INJECTOR, InitPhaseState, LView, LViewFlags, NEXT, PARENT, RENDERER, RENDERER_FACTORY, RootContext, RootContextFlags, SANITIZER, TData, TVIEW, TView, TViewType, T_HOST} from '../interfaces/view';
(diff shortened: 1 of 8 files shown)


packages/core/src/render3/instructions/shared.ts
```diff
@@ -1663,10 +1691,70 @@ function refreshComponent(hostLView: LView, componentHostIdx: number): void {
...
+      refreshView(componentView, tView, tView.template, componentView[CONTEXT]);
+    } else if (
+        componentView[FLAGS] & LViewFlags.ContainsDirty ||
+        componentView[CONTAINS_ALWAYS_COUNT] > 0) {
+      refreshContainsDirtyComponent(hostLView, componentHostIdx);
+    }
+  }
+}
+
+function refreshContainsDirtyComponent(hostLView: LView, componentHostIdx: number): void {
```
mhevery:
```suggestion
function descendComponent(hostLView: LView, componentHostIdx: number): void {
```

1. I would like to reserve `refresh` for actually running CD, and descend to just go into, but not refresh.
2. also it is not `containsDirty` as it could also have `alwaysCount`.

packages/core/src/render3/instructions/shared.ts
```diff
@@ -1663,10 +1691,70 @@ function refreshComponent(hostLView: LView, componentHostIdx: number): void {
...
+  ngDevMode && assertEqual(isCreationMode(hostLView), false, 'Should be run in update mode');
+  const componentView = getComponentLViewByIndex(componentHostIdx, hostLView);
+  // Only attached components that are CheckAlways or OnPush and dirty should be refreshed
+  if (viewAttachedToChangeDetector(componentView) &&
```
mhevery:
```suggestion
  if (isViewAttached(componentView) &&
```
I know not part of your PR, but it would make the code so much easier to read.

packages/core/src/render3/instructions/shared.ts
```diff
@@ -1663,10 +1691,70 @@ function refreshComponent(hostLView: LView, componentHostIdx: number): void {
...
+      embeddedLView[DECLARATION_COMPONENT_VIEW]
+}
+
+function isCheckAlwaysTransplantedView(lContainer: LContainer, embeddedLView: LView) {
+  return isTransplantedView(lContainer, embeddedLView) &&
+      (embeddedLView[DECLARATION_COMPONENT_VIEW][FLAGS] & LViewFlags.CheckAlways);
+}
+
+function refreshContainsDirtyView(lView: LView) {
+  forEachDynamicEmbeddedView(lView, (lContainer: LContainer) => {
```
mhevery: creating closure to pass into `forEach` allocates memory and may create megamorphic call sites. Simple for/while loop is strongly preferred for perf reasons.

PR #34455, opened 2019-12-17
Update CircleCI config to use new determinisitc image types

eddiewebb:
New `cimg` collection of images from CircleCI have the benefits of:
* More deterministic versioning
* Smaller images sizes
* More shared layers to other languages (more cache hits!)

More info on the new images can be found on https://github.com/CircleCI-Public/cimg-node

All this means more confidence and faster builds on CircleCI.

## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [x] CI related changes
- [ ] Documentation content changes
- [ ] angular.io application / infrastructure changes
- [ ] Other... Please describe:

## What is the current behavior?
Uses `circleci` image types that have known concerns with determinism and stability.

Issue Number: N/A

## What is the new behavior?
 Use new more determinant image type

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

## Other information

The image is currently beta, but base image is being used ~20k builds each week, and gorwing every day.

Use and feedback from a community like Angular would be an important milestone for the new types.

PR #35049, opened 2020-01-30
test: use puppeteer in integration tests and to download correct chromedriver

gregmagolan:
This means integration tests no longer need to depend on a $CI_CHROMEDRIVER_VERSION_ARG environment variable to specify which chromedriver version to download to match the locally installed chrome. This was bad DX and not having it specified was not reliable as webdriver-manager would not always download the chromedriver version to work with the locally installed chrome.

`webdriver-manager update --gecko=false --standalone=false $CI_CHROMEDRIVER_VERSION_ARG` is now replaced with `node webdriver-manager-update.js` in the root package.json, which checks which version of chrome puppeteer has come bundled with & downloads informs webdriver-manager to download the corresponding chrome driver version.

Integration tests now use `"webdriver-manager": "file:../../node_modules/webdriver-manager"` so they don't have to waste time calling webdriver-manager update in postinstall
```
"// resolutions": "Ensure a single version of webdriver-manager which comes from root node_modules that has already run webdriver-manager update",
"resolutions": {
"**/webdriver-manager": "file:../../node_modules/webdriver-manager"
}
```
This should speed up each integration postinstall by a few seconds.

Further, integration test package.json files link puppeteer via `file:../../node_modules/puppeteer` which is the ideal situation as the puppeteer post-install won't download chrome if it is already downloaded. In CI, since node_modules is cached it should not need to download Chrome either unless the node_modules cache is busted.

NB: each version of puppeteer comes bundles with a specific version of chrome. Root package.json & yarn.lock currently pull down puppeteer 2.1.0 which comes with chrome 80. See https://github.com/puppeteer/puppeteer#q-which-chromium-version-does-puppeteer-use for more info.

Only two references to CI_CHROMEDRIVER_VERSION_ARG left in integration tests at `integration/bazel-schematics/test.sh` which I'm not entirely sure how to get rid of it <== @kyliau

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [x] Other... Please describe: integration test changes 

Issue Number: N/A

- [ ] Yes
- [x] No


gregmagolan:
BONUS: updated the aio build to use puppeteer as well so it no longer depends on `CI_CHROMEDRIVER_VERSION_ARG`.

Only two references to `CI_CHROMEDRIVER_VERSION_ARG` left:
1) integration/bazel-schematics/test.sh which I'm not entirely sure how to get rid of it <== @kyliau 
2) aio/tools/examples/shared/package.json which I'm not hesitant to touch as its used for examples

aio/scripts/webdriver-manager-update.js
```diff
@@ -0,0 +1,23 @@
...
+  const version = (await browser.version()).split('/')[1];
+  browser.close();
+  process.argv.push(...[
+    'update',
+    '--gecko=false',
+    '--standalone=false',
+    '--versions.chrome',
+    version,
+  ]);
+  require('webdriver-manager/built/lib/webdriver');
```
filipesilva: Clever way of triggering the update!

integration/ng_update_migrations/src/app/migration-tests/providers-test-module.ts
```diff
@@ -1,8 +1,11 @@
-import {Component, Directive, NgModule, NgZone} from '@angular/core';
+import { Component, Directive, NgModule, NgZone, Injectable } from '@angular/core';
```
filipesilva: Were the changes in `ng_update_migrations` changes intentional? It looks like the result of an update.

aio/scripts/webdriver-manager-update.js
filipesilva: Clever way of triggering the update!

filipesilva (changes requested):

aio/scripts/webdriver-manager-update.js
gkalpak: Actually, the more I think about it, the more I believe this is not worth it. The overhead of running the proper command (`yarn webdriver-manager update ...`) should be negligible.

integration/cli-hello-world-ivy-compat/package.json
```diff
@@ -46,6 +45,11 @@
     "protractor": "file:../../node_modules/protractor",
     "ts-node": "8.3.0",
     "tslint": "5.18.0",
-    "typescript": "file:../../node_modules/typescript"
+    "typescript": "file:../../node_modules/typescript",
+    "webdriver-manager": "file:../../node_modules/webdriver-manager"
+  },
+  "// resolutions": "Ensure a single version of webdriver-manager which comes from root node_modules that has already run webdriver-manager update",
+  "resolutions": {
+    "webdriver-manager": "file:../../node_modules/webdriver-manager"
```
gkalpak: Normally, `webdriver-manager` comes in as a dependency of protractor. This means that it can be in `node_modules/protractor/node_modules/`, unless the package manager decides to hoist it to `node_modules/`. Maybe it would be safer to use `"**/webdriver-manager": ...`.

integration/cli-hello-world-ivy-i18n/e2e/protractor.conf.js
```diff
@@ -16,7 +17,7 @@ exports.config = {
     browserName: 'chrome',
     chromeOptions: {
       binary: process.env.CHROME_BIN,
```
gkalpak: Oh, I see what happens. This seems to be an artifact of back when we used to [install Chrome(-ium) ourselves][1] and [expose it to scripts][2] via the `CHROME_BIN` env var. So, the first integration project (added in e9a89c069) used `CHROME_BIN` in the [protractor config][3] and then subsequent integration projects followed suit.

This isn't needed for a long time and now with puppeteer we are overwriting `process.env.CHROME_BIN` anyway, so there is no need to use an anv var. I would drop the env var and use `binary: require('puppeteer').executablePath()` directly.

[1]: https://github.com/angular/angular/blob/e9a89c069/scripts/ci/install_chromium.sh
[2]: https://github.com/angular/angular/blob/e9a89c069/scripts/ci-lite/env.sh#L79
[3]: https://github.com/angular/angular/blob/e9a89c069/integration/hello_world__closure/e2e/protractor.config.js#L9

gkalpak (changes requested):
Yay! One step closer to deterministic builds :tada: 💯

Follow-up tasks (mostly a note to self):
- [x] Swtch all AIO scripts to the puppeteer setup (e.g. `test-a11y-score`, `test-pwa-score`).
- [x] Switch all docs examples scripts to the puppeteer setup.
- [x] Get rid of the `CI_CHROMEDRIVER_VERSION_ARG` environment variable.

integration/cli-hello-world-ivy-i18n/e2e/protractor.conf.js
josephperrott: +1, Agreed

aio/scripts/webdriver-manager-update.js
```diff
@@ -0,0 +1,23 @@
+/**
```
josephperrott: Rather than maintaining this file in two places, can we just have aio use the main `/scripts/webdriver-manager-update.js` script?

aio/scripts/webdriver-manager-update.js
```diff
@@ -0,0 +1,23 @@
...
+  // browser.version() will return a string such as HeadlessChrome/80.0.3987.0
+  const version = (await browser.version()).split('/')[1];
+  browser.close();
+  process.argv.push(...[
+    'update',
+    '--gecko=false',
+    '--standalone=false',
+    '--versions.chrome',
+    version,
+  ]);
```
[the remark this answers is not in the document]
gregmagolan: Hah. Seems I like my `...` operator a little too much.

integration/cli-hello-world-ivy-compat/e2e/protractor.conf.js
```diff
@@ -13,7 +14,11 @@ exports.config = {
     './src/**/*.e2e-spec.ts'
   ],
   capabilities: {
-    'browserName': 'chrome'
+    browserName: 'chrome',
+    chromeOptions: {
+      binary: process.env.CHROME_BIN,
+      args: ['--no-sandbox', '--headless', '--disable-gpu', '--disable-dev-shm-usage']
```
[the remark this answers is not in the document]
gregmagolan: * ( In CI we get `Root cause: org.openqa.selenium.WebDriverException: unknown error: DevToolsActivePort file doesn't exist` which `--disable-dev-shm-usage` resolves.
* `--no-sandbox` is needed in some cases in CI as well and possibly locally as well; it was already set in a few of the integration tests so I opted to make them consistent
* `--disable-gpu` is probably unnecessary; I'll try removing it
* `--headless` is just nice to have IMO so when these tests are running locally you don't have chrome browsers popping up and tearing down on your desktop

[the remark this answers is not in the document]
gregmagolan: https://stackoverflow.com/questions/50642308/webdriverexception-unknown-error-devtoolsactiveport-file-doesnt-exist-while-t has more info on the `DevToolsActivePort` failure

integration/ng_update_migrations/src/app/migration-tests/providers-test-module.ts
gregmagolan: Those snuck in. Thanks.

integration/cli-hello-world-ivy-i18n/e2e/protractor.conf.js
gregmagolan: Sounds good. I'll remove the old `binary: process.env.CHROME_BIN` artifacts

aio/scripts/webdriver-manager-update.js
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+const puppeteer = require('puppeteer');
+
+(async () => {
+  const browser = await puppeteer.launch({headless: true});
+  // browser.version() will return a string such as HeadlessChrome/80.0.3987.0
+  const version = (await browser.version()).split('/')[1];
```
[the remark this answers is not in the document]
gregmagolan: Good call. I added a regex check to make sure the version was sane & linked to the docs as well in a comment.

aio/scripts/webdriver-manager-update.js
```diff
@@ -0,0 +1,23 @@
...
+  browser.close();
+  process.argv.push(...[
```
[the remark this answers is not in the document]
gregmagolan: 👍 I'll just error out if someone passes an argument

aio/scripts/webdriver-manager-update.js
gregmagolan: I was thinking aio needed to be completely self contained ... also if it executes the postinstall at `../scripts/webdriver-manager-update.js` then that script will end up calling into the root `/node_modules/webdriver-manager` and running the update there instead of the `aio/node_modules/webdriver-manager` as intended 🤔

integration/cli-hello-world-ivy-compat/package.json
gregmagolan: I thought of doing
```
"postinstall": "yarn webdriver-manager update ... --versions.chrome $(node get-chrome-version-from-puppeteer.js)"
```
but that would require the postinstall running in bash which would fail on Windows so I opted to just run webdriver-manager via a `require`.
I was avoiding a spawning a new process and calling `yarn webdriver-manager update ...` from within node.
Either way you'd needed a hoisted `node_modules/webdriver-manager` so it seems like we just need to depend on it directly to make sure it is there?

integration/cli-hello-world-ivy-compat/package.json
```diff
@@ -46,6 +45,11 @@
     "protractor": "file:../../node_modules/protractor",
     "ts-node": "8.3.0",
     "tslint": "5.18.0",
-    "typescript": "file:../../node_modules/typescript"
+    "typescript": "file:../../node_modules/typescript",
+    "webdriver-manager": "file:../../node_modules/webdriver-manager"
```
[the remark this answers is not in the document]
gregmagolan: Optimization so that we get the webdriver manager from the root /node_modules which has already run update and we don't have to run the `webdriver-manager update` at all for the integration tests.

aio/tests/deployment/e2e/protractor.conf.js
```diff
@@ -1,6 +1,7 @@
 // @ts-check
 // Protractor configuration file, see link for more information
 // https://github.com/angular/protractor/blob/master/lib/config.ts
+process.env.CHROME_BIN = require('puppeteer').executablePath();
```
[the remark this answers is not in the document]
gregmagolan: Cleaned up. This is only strictly needed for karma config as the karma-chrome-launcher looks for the env var.

aio/scripts/webdriver-manager-update.js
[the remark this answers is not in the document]
gregmagolan: The contents of the `webdriver-manager` bin looked like something to avoid requiring as it did has some logic based on the cwd() to determine where to require `built/lib/webdriver` from but looks like it will work. 👍

aio/scripts/webdriver-manager-update.js
gregmagolan: I'll move the aio changes to a separate PR as I think there is enough to review on just the integration changes

integration/cli-hello-world-ivy-compat/package.json
gregmagolan: 👍

aio/scripts/webdriver-manager-update.js
gregmagolan: I thought of doing
```
"postinstall": "yarn webdriver-manager update ... --versions.chrome $(node get-chrome-version-from-puppeteer.js)"
```
but that would require the postinstall running in bash which would fail on Windows so I opted to just run webdriver-manager via a require.

I was avoiding a spawning a new process and calling yarn webdriver-manager update ... from within node.

Either way you'd needed a hoisted node_modules/webdriver-manager so it seems like we just need to depend on it directly to make sure it is there?

gkalpak: Protractor (which is a top-level dependency) provides a `webdriver-manager` script, so you don't have to rely on `webdriver-manager` being a top-level dependency (or hoisted) for there to exist `node_modules/.bin/webdriver-manager`. It is safe to use the script (i.e. it will work with or without `webdriver-manager` being hoisted and the overhead of spawning an extra process once every `postinstall` script is negligible.

integration/cli-hello-world/package.json
```diff
@@ -42,8 +42,13 @@
     "karma-jasmine": "2.0.1",
     "karma-jasmine-html-reporter": "1.4.2",
     "protractor": "file:../../node_modules/protractor",
+    "puppeteer": "file:../../node_modules/puppeteer",
```
gkalpak: Isn't this change needed in all integration projects?

.circleci/config.yml
```diff
@@ -133,12 +133,29 @@ commands:
...
+      - run:
+          name: Install shared Chrome libs
+          command: |
+            sudo apt-get update
+            # Install GTK+ graphical user interface (libgtk-3-0), advanced linux sound architecture (libasound2)
+            # and network security service libraries (libnss3) & X11 Screen Saver extension library (libssx1)
+            # which are dependendies of chrome & needed for karma & protractor headless chrome tests.
+            # This is a very small install which takes around 7s in comparing to using the full
+            # circleci/node:x.x.x-browsers image.
+            sudo apt-get -y install libgtk-3-0 libasound2 libnss3 libxss1
```
gkalpak: Why is this change needed? Aren't the CircleCI images we use to run those tests providing the necessary deps? How did they work before?

gkalpak (dismissed):
Getting there :smiley:
Good call moving the AIO stuff to a different PR :+1:

.circleci/config.yml
gregmagolan: In short, there were not needed before as Chrome was not being run in the job but now that the postinstall webdriver-manager update script runs puppeteers Chrome in headless to find out it's version the shared libs are needed in all containers.

Added a comment above as to why as realized that was missing:

These are now needed in all containers as the webdriver-manager update script which is run in the root postinstall
now starts puppeteer's Chrome in headless to check its version so the correct chromedriver is downloaded. See `/scripts/webdriver-manager-update.js`.

gregmagolan: And no, the base circleci/node image does not have the required shared libs for Chrome to run. The circleci/node-browsers image does but it also includes the browsers which would add more time to each job to setup the environment than doing this apt-get here.

integration/cli-hello-world/package.json
gregmagolan: Yes. All the ones that do browser testing.

gregmagolan: Oh. Good point. I must have removed them from the other on cleanup. It does work without it there as require('puppeteer') would fall back to the root node_modules .... but once the next step of npm_integration_test is added then these tests are sandboxed and it would fail at that point.

aio/scripts/webdriver-manager-update.js
gregmagolan: Ohhhh.  I didn't realize that. That is where that bin entry comes from. That clears things up thanks!
In that case that is the bin to use here.

package.json
```diff
@@ -173,8 +173,10 @@
     "vrsource-tslint-rules": "5.1.1"
   },
   "// 4": "Overwrite graceful-fs to a version that does not rely on the 'natives' package. This fixes gulp for >= 10.13, more information: #28213",
+  "// 5": "Resolve to a single version of webdriver-manager that yarn should hoist from protractor#webdriver-manager",
```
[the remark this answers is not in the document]
gregmagolan: I was under the impression that webdriver-manager needed to be hoisted to work but since there is a protractor webdriver-manager bin entry then its not. Thanks George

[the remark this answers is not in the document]
gregmagolan: Actually. Circling back to this. The integration tests are relying on webdriver-manager to be found at `../../node_modules/webdriver-manager` so I think we need to have "**/webdriver-manager" in resolutions to ensure that there is only one version that will get hoisted.

Is there another way to ensure this?

scripts/webdriver-manager-update.js
```diff
@@ -0,0 +1,41 @@
...
+  }
+  browser.close();
+  process.argv.push(
+    'update',
+    '--gecko=false',
+    '--standalone=false',
+    '--versions.chrome',
+    version,
+  );
+  require('webdriver-manager');
```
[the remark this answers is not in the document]
gregmagolan: Done.

scripts/webdriver-manager-update.js
```diff
@@ -0,0 +1,41 @@
...
+  // browser.version() will return a string such as HeadlessChrome/80.0.3987.0
+  // See https://github.com/puppeteer/puppeteer/blob/v2.1.0/docs/api.md#browserversion.
+  // NB: the format may change in a future version of Chromium so we test for
+  // the version number below.
+  const version = (await browser.version()).split('/')[1];
+  if (!/^[0-9]+\.[0-9]+\.[0-9]+\.[0-9]$/.test(version)) {
+    console.error(`[webdriver-manager-update.js] Error: could not determine Chrome version from puppeteer. Expected 'x.x.x.x' but got '${version}'.`);
+    process.exit(1);
+  }
+  browser.close();
```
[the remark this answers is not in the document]
gregmagolan: Good catch. Thanks.

aio/scripts/webdriver-manager-update.js
gregmagolan: I made the script a little more portable so that it can be used by /aio which is now in a separate PR

package.json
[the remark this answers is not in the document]
gregmagolan: Further to that, when npm_integration_test is brought in, it will depend on the `@npm//webdriver-manager` label so we will depend on the hoisting.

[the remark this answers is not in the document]
gkalpak: I don't know of another way to ensure that. (I am not even sure this way ensures it, but it is probably our best bet :grin:)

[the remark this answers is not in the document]
gregmagolan: Cool. yarn could break it in the future but it will be obvious as the `@npm//webdriver-manager` label will no longer exist.

.circleci/config.yml
gkalpak: Could we switch to the non-browser CircleCI docker images in some jobs (e.g. the integration one) now that we use headless Chrome?

gregmagolan: Almost. the integration/bazel-schematics test is the last that still uses local chrome but I'll need some help from @kyliau to remove that dependency so likely to be a follow up PR

filipesilva: I don't quite follow here. I understand that when running the webdriver-update script we need to run chrome, so we need the shared libs. But I don't understand why it's necessary to run the script for all jobs. Jobs that use chrome need it, but jobs that don't need chrome don't need it.

It looks like we decided to use a postinstall at the root to update webdriver. I think that makes sense and simplifies the setup. But doing so is also a tradeoff: jobs that are never going to use chrome are still running this postinstall.

So the real reason for installing these libraries is not that the new webdriver install script requires running chrome. Instead, the reason is that we opted to install webdriver in all jobs, even those that don't need it, in order to simplify our setup. Changing the script just exposed this.

In the past having steps that require `sudo` hasn't been ideal. It usually only comes up when we need to change some parts of our CI setup and hit lower permissions, like in https://github.com/angular/angular/pull/34455.

filipesilva: We can partially work around this by having an alternate method of looking up the chrome version. Puppeteer includes in its `package.json` a field that indicates the chrome revision, and from that we can get to the version number somehow. I think there's a online API for that? I remember seeing it once, but can't find it now.

This still wouldn't address the fact we're installing puppeteer/webdriver/chrome deps on jobs that never use it. It would just help to only install chrome deps on jobs that do use it. It would also make the setup a bit more complicated because now those jobs would need to explicitly install these deps.

But if we'd get to the point where we're saying "this job needs chrome deps for chromedriver", why not go all the way and say "this job needs chromedriver" instead, and only then update webdriver?

I suppose the downside is that local installs are not privy to this logic. But we already have an answer to that too:  `./.circleci/env.sh` contains variables only present in circleci. Maybe that can be leveraged to only run the webdriver update on jobs that need it, and to always run it outside circleci.

filipesilva: Reviewing this approach makes me think it's somewhat worse than just using the circleci images with chrome when needed. Instead of installing chrome deps we could say "if this container has chrome, install webdriver".

gregmagolan: With puppeteer in the root package.json, its postinstall step to download chrome will happen in the setup job as well and it is actually quite slow. The webdriver update is fast in comparison but doing it in the postinstall of the root package.json simplifies the setup for all integration tests so they can just depend on `../../node_modules/puppeteer` && `../../node_modules/webdriver-manager` and puppeteer won't download Chrome since it is already downloaded and webdriver-manager update won't need to be called. IMO that's enough reason to do this in the root package.json in order to simplify the integration tests and 10s or so for each one... although the webdriver-manager update is quite fast in comparison to the puppeteer post-install so it could still be done in each integration test, however, these will soon run in the the same CircleCI `test` job as everything else.

With https://github.com/angular/angular/pull/33927, all integration tests will be run in the `test` && `test_ivy_aot` jobs by Bazel in CircleCI and the legacy integration tests job & shell script will be removed in a follow up PR. Eventually some or all of the aio jobs could get moved to run under bazel as well using npm_integration_test so with moving to a setup where `bazel test ...` can test everything it makes more sense for the root package.json to run the webdriver-update.

NB: for cross-platform RBE, the integration tests puppeteer postinstall will still download chrome since they will need the linux Chrome while the repository_rule (which always runs on the local host machine) would have downloaded the OSX binaries.

But.. if there is an easy way to get the Chrome version from the puppeteer package.json with an API to map the version number or to vendor in a mapping file from somewhere so it is available locally then that would mean we wouldn't need to have these shared libs installed in the `setup` job but could move it to only the jobs that actually need to run the puppeteer (or rules_webtesting) provisioned Chrome for testing.

gregmagolan: Ok. After some thought I opted for a very simple mapping up puppeteer semver version to Chrome version:
```
// Mapping of puppeteer releases to their default Chrome version
// derived from https://github.com/puppeteer/puppeteer/blob/master/docs/api.md.
// The puppeteer package.json file the Chrome revision such as
// "chromium_revision": "722234" but this does not map easily to the Chrome
// so we use this mapping here instead.
module.exports = {
  "2.1.0": "80.0.3987.0",
  "2.0.0": "79.0.3942.0",
  "1.20.0": "78.0.3882.0",
  "1.19.0": "77.0.3803.0",
  "1.17.0": "76.0.3803.0",
  "1.15.0": "75.0.3765.0",
  "1.13.0": "74.0.3723.0",
  "1.12.2": "73.0.3679.0",
};
```
This will be easy to maintain (maybe it should be upstreamed to puppeteer) and when puppeteer is updated to a new version the webdriver-manager-update script will error out and prompt you to update the mappings file.

gregmagolan: This means that the root yarn install postinstall setup no longer needs to have chrome shared libs available because it needs to launch headless chrome.

gregmagolan:
aio changes included back into this PR at Igor's request after all the review comments & Igor reviewed both.

https://github.com/angular/angular/pull/35102 now just testing CI job for this PR that only run on master

gregmagolan:
@gkalpak I switch the PWA score test to use protractor. Couldn't figure out the doc examples so will leave that to you :)

IgorMinar (approved):
this looks good to me now. Thanks @gregmagolan for working through all the review feedback.

IgorMinar:
let's see if @filipesilva or @gkalpak have anything else to add. I believe that all their feedback has been addressed.

integration/cli-hello-world-ivy-compat/karma.conf.js
```diff
@@ -1,6 +1,11 @@
 // Karma configuration file, see link for more information
 // https://karma-runner.github.io/1.0/config/configuration-file.html
 
+// Env var CHROME_BIN is later picked up by karma-chrome-launcher that is triggered by
```
filipesilva: Maybe this comment could also be in `integration/README.md` instead of repeated on karma configs.

integration/dynamic-compiler/package.json
```diff
@@ -42,5 +42,9 @@
     "rxjs": "file:../../node_modules/rxjs",
     "systemjs": "file:../../node_modules/systemjs",
     "zone.js": "file:../../dist/zone.js-dist/zone.js"
+  },
+  "//resolutions-comment": "Ensure a single version of webdriver-manager which comes from root node_modules that has already run webdriver-manager update",
```
filipesilva: Maybe this comment could also be in integration/README.md instead of repeated on package.json.

filipesilva (approved):
LGTM, thanks! I left a few non blocking comments, but it's up to you if they are a good suggestion or not.

integration/cli-hello-world-ivy-compat/karma.conf.js
gregmagolan: Seems to me like its better in this file even if there is some repetition as in the README.md you'd loose the context

aio/tests/deployment/e2e/protractor.conf.js
```diff
@@ -14,6 +14,11 @@ exports.config = {
   suite: 'full',
   capabilities: {
     browserName: 'chrome',
+    chromeOptions: process.env['CI'] ? {
+      binary: require('puppeteer').executablePath(),
+      // See /integration/README.md#Browser tests for more info on these args
```
josephperrott: nit: The link should actually be `README.md#browser-tests`

aio/package.json
```diff
@@ -155,6 +154,7 @@
     "lunr": "^2.1.0",
     "npm-run-all": "^4.1.5",
     "protractor": "^5.4.2",
+    "puppeteer": "^2.1.0",
```
josephperrott: We should have this to be pinned to `2.1.0` so we don't accidentally update puppeteer when someone updates something else in `package.json`

josephperrott (approved):
LGTM, just a couple nits

aio/tests/deployment/e2e/protractor.conf.js
gregmagolan: 👍

aio/package.json
gregmagolan: 👍 Pinning to 2.1.1

PR #35057, opened 2020-01-30
fix(ngcc): do not lock if the target is not compiled by Angular

### LINKED ISSUE #35000: Allow specifying custom NGCC lockfile location (cannot build in parallel with Ivy)

petebacondarwin:
To support parallel CLI builds we instruct developers to pre-process
their node_modules via ngcc at the command line.

Despite doing this ngcc was still trying to set a lock when it was being
triggered by the CLI for packages that are not going to be processed,
since they are not compiled by Angular for instance.

This commit checks whether a target package needs to be compiled
at all before attempting to set the lock.

Fixes #35000

### DIFF
--- a/packages/compiler-cli/ngcc/src/entry_point_finder/targeted_entry_point_finder.ts
+++ b/packages/compiler-cli/ngcc/src/entry_point_finder/targeted_entry_point_finder.ts
@@ -71,7 +71,7 @@ export class TargetedEntryPointFinder implements EntryPointFinder {
       }
     }
     // All `propertiesToConsider` that appear in this entry-point have been processed.
-    // In other word, there were no properties that need processing.
+    // In other words, there were no properties that need processing.
     return false;
   }
 
--- a/packages/compiler-cli/ngcc/src/entry_point_finder/targeted_entry_point_finder.ts
+++ b/packages/compiler-cli/ngcc/src/entry_point_finder/targeted_entry_point_finder.ts
@@ -51,29 +51,28 @@ export class TargetedEntryPointFinder implements EntryPointFinder {
     return entryPoints;
   }
 
-  hasProcessedTargetEntryPoint(
+  targetNeedsProcessingOrCleaning(
       propertiesToConsider: EntryPointJsonProperty[], compileAllFormats: boolean): boolean {
     const entryPoint = this.getEntryPoint(this.targetPath);
     if (entryPoint === null || !entryPoint.compiledByAngular) {
-      return true;
+      return false;
     }
 
     for (const property of propertiesToConsider) {
       if (entryPoint.packageJson[property]) {
-        // Here is a property that should be processed
+        // Here is a property that should be processed.
         if (!hasBeenProcessed(entryPoint.packageJson, property)) {
-          // `hasBeenProcessed()` may return `null`, which means that this entry-point package needs
-          // cleaning and therefore needs to be processed.
-          return false;
+          return true;
         }
         if (!compileAllFormats) {
-          // It has been processed and we only need one, so we are done.
-          return true;
+          // This property has been processed, and we only need one.
+          return false;
         }
       }
     }
-    // All formats need to be compiled and there were none that were not,
-    return true;
+    // All `propertiesToConsider` that appear in this entry-point have been processed.
+    // In other word, there were no properties that need processing.
+    return false;
   }
 
   private processNextPath(): void {
--- a/packages/compiler-cli/ngcc/src/entry_point_finder/targeted_entry_point_finder.ts
+++ b/packages/compiler-cli/ngcc/src/entry_point_finder/targeted_entry_point_finder.ts
@@ -8,8 +8,9 @@
 import {AbsoluteFsPath, FileSystem, PathSegment, join, relative, relativeFrom} from '../../../src/ngtsc/file_system';
 import {DependencyResolver, SortedEntryPointsInfo} from '../dependencies/dependency_resolver';
 import {Logger} from '../logging/logger';
+import {hasBeenProcessed} from '../packages/build_marker';
 import {NgccConfiguration} from '../packages/configuration';
-import {EntryPoint, getEntryPointInfo} from '../packages/entry_point';
+import {EntryPoint, EntryPointJsonProperty, getEntryPointInfo} from '../packages/entry_point';
 import {PathMappings} from '../utils';
 import {EntryPointFinder} from './interface';
 import {getBasePaths} from './utils';
@@ -37,8 +38,42 @@ export class TargetedEntryPointFinder implements EntryPointFinder {
       this.processNextPath();
     }
     const targetEntryPoint = this.unsortedEntryPoints.get(this.targetPath);
-    return this.resolver.sortEntryPointsByDependency(
+    const entryPoints = this.resolver.sortEntryPointsByDependency(
         Array.from(this.unsortedEntryPoints.values()), targetEntryPoint);
+
+    const invalidTarget =
+        entryPoints.invalidEntryPoints.find(i => i.entryPoint.path === this.targetPath);
+    if (invalidTarget !== undefined) {
+      throw new Error(
+          `The target entry-point "${invalidTarget.entryPoint.name}" has missing dependencies:\n` +
+          invalidTarget.missingDependencies.map(dep => ` - ${dep}\n`).join(''));
+    }
+    return entryPoints;
+  }
+
+  hasProcessedTargetEntryPoint(
+      propertiesToConsider: EntryPointJsonProperty[], compileAllFormats: boolean): boolean {
+    const entryPoint = this.getEntryPoint(this.targetPath);
+    if (entryPoint === null || !entryPoint.compiledByAngular) {
+      return true;
+    }
+
+    for (const property of propertiesToConsider) {
+      if (entryPoint.packageJson[property]) {
+        // Here is a property that should be processed
+        if (!hasBeenProcessed(entryPoint.packageJson, property)) {
+          // `hasBeenProcessed()` may return `null`, which means that this entry-point package needs
+          // cleaning and therefore needs to be processed.
+          return false;
+        }
+        if (!compileAllFormats) {
+          // It has been processed and we only need one, so we are done.
+          return true;
+        }
+      }
+    }
+    // All formats need to be compiled and there were none that were not,
+    return true;
   }
 
   private processNextPath(): void {
--- a/packages/compiler-cli/ngcc/src/main.ts
+++ b/packages/compiler-cli/ngcc/src/main.ts
@@ -16,12 +16,13 @@ import {replaceTsWithNgInErrors} from '../../src/ngtsc/diagnostics';
 import {AbsoluteFsPath, FileSystem, absoluteFrom, dirname, getFileSystem, resolve} from '../../src/ngtsc/file_system';
 
 import {CommonJsDependencyHost} from './dependencies/commonjs_dependency_host';
-import {DependencyResolver, InvalidEntryPoint, PartiallyOrderedEntryPoints, SortedEntryPointsInfo} from './dependencies/dependency_resolver';
+import {DependencyResolver, InvalidEntryPoint} from './dependencies/dependency_resolver';
 import {DtsDependencyHost} from './dependencies/dts_dependency_host';
 import {EsmDependencyHost} from './dependencies/esm_dependency_host';
 import {ModuleResolver} from './dependencies/module_resolver';
 import {UmdDependencyHost} from './dependencies/umd_dependency_host';
 import {DirectoryWalkerEntryPointFinder} from './entry_point_finder/directory_walker_entry_point_finder';
+import {EntryPointFinder} from './entry_point_finder/interface';
 import {TargetedEntryPointFinder} from './entry_point_finder/targeted_entry_point_finder';
 import {AnalyzeEntryPointsFn, CreateCompileFn, Executor, PartiallyOrderedTasks, Task, TaskProcessingOutcome, TaskQueue} from './execution/api';
 import {ClusterExecutor} from './execution/cluster/executor';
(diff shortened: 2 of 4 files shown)


petebacondarwin:
This PR looks a lot more complicated than it is. Mostly it is just a refactoring - moving the `hasProcessedTargetEntryPoint()` check into the `EntryPointFinder`.

packages/compiler-cli/ngcc/src/entry_point_finder/targeted_entry_point_finder.ts
```diff
@@ -37,8 +38,42 @@ export class TargetedEntryPointFinder implements EntryPointFinder {
...
+          // cleaning and therefore needs to be processed.
+          return false;
+        }
+        if (!compileAllFormats) {
+          // It has been processed and we only need one, so we are done.
+          return true;
+        }
+      }
+    }
+    // All formats need to be compiled and there were none that were not,
```
gkalpak: `,` --> `.` (or was there more to say?)
Actually, I think we should also mention that this method will also return `true` if the package has none of the requested properties.

packages/compiler-cli/ngcc/test/integration/ngcc_spec.ts
```diff
@@ -480,20 +480,47 @@ runInEachFileSystem(() => {
         expect(loadPackage('@angular/common/testing').__processed_by_ivy_ngcc__).toBeUndefined();
       });
 
-      it('should mark a non-Angular package target as processed', () => {
+      it('should not mark a non-Angular package as processed if it is the target', () => {
```
gkalpak: Why not? 😕

packages/compiler-cli/ngcc/test/integration/ngcc_spec.ts
```diff
@@ -480,20 +480,47 @@ runInEachFileSystem(() => {
...
+        // * `test-package` has no Angular and is not marked as processed.
+        expect(loadPackage('test-package').__processed_by_ivy_ngcc__).toBeUndefined();
+
+        // * `core` is a dependency of `test-package`, but it is also not processed, since
+        // test-package was not processed.
+        expect(loadPackage('@angular/core').__processed_by_ivy_ngcc__).toBeUndefined();
+      });
+
+      it('should not mark a non-Angular package as processed if it is a dependency', () => {
+        // a package that depends upon test-package, which no metadata.json file so not compiled by
```
gkalpak:
```suggestion
        // a package that depends upon test-package, which has no metadata.json file so not compiled by
```

packages/compiler-cli/ngcc/test/integration/ngcc_spec.ts
```diff
@@ -480,20 +480,47 @@ runInEachFileSystem(() => {
...
-          esm2015: '0.0.0-PLACEHOLDER',
-          esm5: '0.0.0-PLACEHOLDER',
-          fesm2015: '0.0.0-PLACEHOLDER',
-          fesm5: '0.0.0-PLACEHOLDER',
-          main: '0.0.0-PLACEHOLDER',
-          module: '0.0.0-PLACEHOLDER',
+          typings: '0.0.0-PLACEHOLDER',
         });
 
+        // * `test-package` has no Angular but is marked as processed.
```
gkalpak: Is it? :smiley:

gkalpak (changes requested):

packages/compiler-cli/ngcc/src/main.ts
```diff
@@ -188,18 +192,18 @@ export function mainNgcc(
 
     const absBasePath = absoluteFrom(basePath);
```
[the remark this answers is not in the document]
petebacondarwin: Yes! The result of a poor rebase conflict resolution on my part. That'll teach for me for trying to do a quick rebase late at night. I'll clean all this up now.

packages/compiler-cli/ngcc/src/main.ts
```diff
@@ -147,16 +148,19 @@ export function mainNgcc(
...
   const absoluteTargetEntryPointPath =
-      targetEntryPointPath !== undefined ? resolve(basePath, targetEntryPointPath) : undefined;
-  if (absoluteTargetEntryPointPath !== undefined &&
-      hasProcessedTargetEntryPoint(
-          fileSystem, absoluteTargetEntryPointPath, supportedPropertiesToConsider,
-          compileAllFormats)) {
+      targetEntryPointPath !== undefined ? resolve(basePath, targetEntryPointPath) : null;
+  const finder = getEntryPointFinder(
+      fileSystem, logger, dependencyResolver, config, absBasePath, absoluteTargetEntryPointPath,
+      pathMappings);
```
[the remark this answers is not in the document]
petebacondarwin: I agree this is not ideal for workers but I think we need to a more significant refactoring of the code base to make workers more lightweight in general. Watch this space for a refactoring PR once we are through 9.0.0...

packages/compiler-cli/ngcc/src/entry_point_finder/targeted_entry_point_finder.ts
```diff
@@ -37,8 +38,42 @@ export class TargetedEntryPointFinder implements EntryPointFinder {
...
+    const entryPoint = this.getEntryPoint(this.targetPath);
+    if (entryPoint === null || !entryPoint.compiledByAngular) {
+      return true;
+    }
+
+    for (const property of propertiesToConsider) {
+      if (entryPoint.packageJson[property]) {
+        // Here is a property that should be processed
+        if (!hasBeenProcessed(entryPoint.packageJson, property)) {
+          // `hasBeenProcessed()` may return `null`, which means that this entry-point package needs
```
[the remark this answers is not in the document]
petebacondarwin: Yeah - sorry. This comment was left over from an alternative approach that I was playing with.

packages/compiler-cli/ngcc/src/main.ts
```diff
@@ -364,47 +368,22 @@ function getExecutor(
...
+      fileSystem, logger, {
+        esm5: esmDependencyHost,
+        esm2015: esmDependencyHost,
+        umd: umdDependencyHost,
+        commonjs: commonJsDependencyHost
+      },
+      dtsDependencyHost);
 }
 
 function hasProcessedTargetEntryPoint(
```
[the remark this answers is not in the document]
petebacondarwin: Yep - more rebase disaster :-)

packages/compiler-cli/ngcc/test/integration/ngcc_spec.ts
petebacondarwin: So now we bail early if the target is not an Angular package before we get into the lockfile control.
We do not want to do any writing to the file-system (e.g. updating the package.json as being processed) outside of lockfile protection since this could cause corruption.
This check is about as fast as checking for the build marker, so there is not much loss in performance. Plus in CLI integration where this performance is more important, the CLI will track the file as processed itself so won't trigger ngcc again for it.

packages/compiler-cli/ngcc/test/integration/ngcc_spec.ts
petebacondarwin: 😱

packages/compiler-cli/ngcc/test/integration/ngcc_spec.ts
petebacondarwin: Also in general I think it is best to avoid touching a package's files if it is not one that we are actively processing.

packages/compiler-cli/ngcc/test/entry_point_finder/targeted_entry_point_finder_spec.ts
```diff
@@ -227,32 +228,257 @@ runInEachFileSystem(() => {
...
+             };
+             fs.writeFile(packageJsonPath, JSON.stringify(packageJson));
+
+             const finder = new TargetedEntryPointFinder(
+                 fs, config, logger, resolver, _Abs('/sub_entry_points/node_modules'), targetPath,
+                 undefined);
+             expect(finder.targetNeedsProcessingOrCleaning(['fesm2015', 'esm5'], false)).toBe(true);
+           });
+
+        it('should return true if the first of the properties to consider (that actually appear in the package.json) has been processed',
```
gkalpak:
```suggestion
        it('should return false if the first of the properties to consider (that actually appear in the package.json) has been processed',
```

packages/compiler-cli/ngcc/test/entry_point_finder/targeted_entry_point_finder_spec.ts
```diff
@@ -227,32 +228,257 @@ runInEachFileSystem(() => {
...
+             const packageJson = JSON.parse(fs.readFile(packageJsonPath));
+             packageJson.__processed_by_ivy_ngcc__ = {
+               fesm2015: NGCC_VERSION,
+             };
+             fs.writeFile(packageJsonPath, JSON.stringify(packageJson));
+
+             const finder = new TargetedEntryPointFinder(
+                 fs, config, logger, resolver, _Abs('/sub_entry_points/node_modules'), targetPath,
+                 undefined);
+             expect(finder.targetNeedsProcessingOrCleaning(['fesm2015', 'esm5'], true)).toBe(true);
```
gkalpak:
```suggestion
             expect(finder.targetNeedsProcessingOrCleaning(['fesm2015', 'esm5'], false)).toBe(false);
```

packages/compiler-cli/ngcc/test/entry_point_finder/targeted_entry_point_finder_spec.ts
```diff
@@ -227,32 +228,257 @@ runInEachFileSystem(() => {
...
+             // Add build markers to the package.json
+             const packageJsonPath = _Abs(`${targetPath}/package.json`);
+             const packageJson = JSON.parse(fs.readFile(packageJsonPath));
+             packageJson.__processed_by_ivy_ngcc__ = {
+               fesm2015: NGCC_VERSION,
+             };
+             fs.writeFile(packageJsonPath, JSON.stringify(packageJson));
+
+             const finder = new TargetedEntryPointFinder(
+                 fs, config, logger, resolver, _Abs('/sub_entry_points/node_modules'), targetPath,
```
gkalpak: ?
```suggestion
                 fs, config, logger, resolver, basePath, targetPath,
```

gkalpak (approved):
A couple of minor (and few super-minor) comments. Otherwise lgtm 👌

packages/compiler-cli/ngcc/src/main.ts
```diff
@@ -172,34 +176,18 @@ export function mainNgcc(
...
-      entryPointInfo = getEntryPoints(
-          fileSystem, pkgJsonUpdater, logger, dependencyResolver, config, absBasePath,
-          absoluteTargetEntryPointPath, pathMappings);
+      entryPointInfo = finder.findEntryPoints();
+    }
+
+    const {entryPoints, invalidEntryPoints, graph} = entryPointInfo;
+    logInvalidEntryPoints(logger, invalidEntryPoints);
+    if (entryPoints.length === 0 && absoluteTargetEntryPointPath !== null) {
+      markNonAngularPackageAsProcessed(fileSystem, pkgJsonUpdater, absoluteTargetEntryPointPath);
```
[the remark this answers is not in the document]
petebacondarwin: You're right! Another rebase mistake.

packages/compiler-cli/ngcc/test/entry_point_finder/targeted_entry_point_finder_spec.ts
```diff
@@ -227,32 +228,257 @@ runInEachFileSystem(() => {
...
+            ...createPackage(fs.resolve(basePath, 'common'), 'testing', ['common']),
+          ]);
+
+          // Add build markers to the package.json
+          const packageJsonPath = _Abs(`${targetPath}/package.json`);
+          const packageJson = JSON.parse(fs.readFile(packageJsonPath));
+          packageJson.__processed_by_ivy_ngcc__ = {
+            fesm2015: NGCC_VERSION,
+            esm5: NGCC_VERSION,
+            main: NGCC_VERSION,
```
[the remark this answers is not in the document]
petebacondarwin: What do you mean? It is true that the test will not fail without it. But equally, I wanted to throw in a property that is not one of the `propertiesToConsider` just in case that messes something up.

packages/compiler-cli/ngcc/test/entry_point_finder/targeted_entry_point_finder_spec.ts
petebacondarwin: Nice spot!

packages/compiler-cli/ngcc/test/entry_point_finder/targeted_entry_point_finder_spec.ts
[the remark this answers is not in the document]
gkalpak: :100:

That's fine. I was just wondering whether there was some hidden connection I was missing.

PR #35060, opened 2020-01-30
ci: add verification of the pullapprove config

### LINKED ISSUE #34956: Replace the functionality found in the verify-codeownership script we had for CODEOWNERS
As above, Replace the functionality found in the verify-codeownership script we had for CODEOWNERS.

We need to make sure the same goals are met for PullApprove that we had for CODEOWNERS.

josephperrott:
After approval of the verification script itself, another commit will be added to resolve outstanding missing pullapprove configuration.

Passing verification example output:
```
┌──────────────────────────────────────────────────────────────────────────────┐
│                                    Result                                    │
└──────────────────────────────────────────────────────────────────────────────┘
PullApprove verification succeeded!
┌──────────────────────────────────────────────────────────────────────────────┐
│                        PullApprove file match results                        │
└──────────────────────────────────────────────────────────────────────────────┘
Matched Files (8677 files)
Unmatched Files (0 files)
┌──────────────────────────────────────────────────────────────────────────────┐
│                           PullApprove group matches                          │
└──────────────────────────────────────────────────────────────────────────────┘
Matched Groups (32 groups)
Unmatched Groups (0 groups)

```

Closes #34956

### DIFF
--- a/tools/pullapprove/verify.js
+++ b/tools/pullapprove/verify.js
@@ -1,0 +1,153 @@
+#!/usr/bin/env node
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+const parse = require('yaml').parse;
+const readFileSync = require('fs').readFileSync;
+const Minimatch = require('minimatch').Minimatch;
+const exec = require('shelljs').exec;
+const path = require('path');
+
+
+// Whether to log verbosely
+const VERBOSE_MODE = !!process.argv.find(arg =>  arg === `-v`);
+// Full path to PullApprove config file
+const PULL_APPROVE_YAML_PATH = path.resolve(__dirname, '../../.pullapprove.yml')
+// All relative path file names in the git repo
+const ALL_FILES = exec('git ls-tree --full-tree -r --name-only HEAD', {silent: true}).trim().split('\n').filter(_ => _);
+if (!ALL_FILES.length) {
+  console.error(
+    `No files were found to be in the git tree, did you run this command from \n` +
+    `inside the angular repository?`);
+  process.exit(1);
+}
+
+/** Gets the glob matching information from each group's condition. */
+function getGlobMatcherFromConditions(groupName, condition) {
+  return condition
+    .trim().split('\n').slice(1,-1)
+    .map((glob) => {
+      const match = glob.match(/'(.*)'/);
+      return match && match[1];
+    })
+    .filter(globString => !!globString)
+    .map(globString => ({
+      group: groupName,
+      glob: globString,
+      matcher: new Minimatch(globString, {dot: true}),
+      matchCount: 0,
+    }));
+}
+
+/** Create logs for each review group. */
+function logGroups(groups) {
+  const globGroups = Array.from(groups.entries()).sort((a, b) => a[0] > b[0] ? 1 : -1);
+  for (let globGroup of globGroups) {
+    console.groupCollapsed(globGroup[0]);
+    const globs = Array.from(globGroup[1].values()).sort((a,b) => b.matchCount - a.matchCount);
+    for (let glob of globs) {
+      console.log(`${glob.glob} - ${glob.matchCount}`);
+    }
+    console.groupEnd();
+  }
+}
+
+/** Logs a header within a text drawn box. */
+function logHeader(...params) {
+  const totalWidth = 80;
+  const fillWidth = totalWidth - 2;
+  const headerText = [...params].join(' ').substr(0, fillWidth);
+  const leftSpace = Math.ceil((fillWidth - headerText.length) / 2);
+  const rightSpace = fillWidth - leftSpace - headerText.length;
+  const fill = (count, content) => Array(++count).join(content);
+
+  console.log(`┌${fill(fillWidth, '─')}┐`);
+  console.log(`│${fill(leftSpace, ' ')}${headerText}${fill(rightSpace, ' ')}│`);
+  console.log(`└${fill(fillWidth, '─')}┘`);
+}
+
+/** Runs the pull approve verification check on provided files. */
+function runVerification(files) {
+  // The pull approve config file.
+  const pullApprove = readFileSync(PULL_APPROVE_YAML_PATH, {encoding: 'utf8'});
+  // All of the PullApprove groups, parsed from the PullApprove yaml file.
+  const parsedPullApproveGroups = parse(pullApprove).groups;
+  // All of the globs created for each group's conditions.
+  const allGlobs = [];
+  // All files which were found to match a condition in PullApprove.
+  const matchedFiles = new Set();
+  // All files which were not found to match a condition in PullApprove.
+  const unmatchedFiles = new Set();
+  // All PullApprove groups which matched at least one file.
+  const matchedGroups = new Map();
+  // All PullApprove groups which did not match at least one file.
+  const unmatchedGroups = new Map();
+
+  // Get all of the globs from the PullApprove group conditions.
+  Object.entries(parsedPullApproveGroups).map(([groupName, group]) => {
+    for (let condition of group.conditions) {
+      allGlobs.push(...getGlobMatcherFromConditions(groupName, condition))
+    }
+  })
+
+  // Check each file for if it is matched by a PullApprove condition.
+  for (let file of files) {
+    const matched = allGlobs.filter(glob => glob.matcher.match(file));
+    matched.length ? matchedFiles.add(file) : unmatchedFiles.add(file);
+    matched.forEach(glob => glob.matchCount++);
+  }
+
+  // Add each glob for each group to a map either matched or unmatched.
+  allGlobs.forEach(glob => {
+    const groupsMap = glob.matchCount ? matchedGroups : unmatchedGroups;
+    let groupMap = groupsMap.get(glob.group) || new Map()
+    groupsMap.set(glob.group, groupMap);
+    groupMap.set(glob.glob, glob);
+  })
+
+  // PullApprove is considered verified if no files or groups are found to be unsed.
+  const verificationSucceeded = !(unmatchedFiles.size || unmatchedGroups.size);
+
+  /**
+   * Overall result
+   */
+  logHeader('Result')
+  if (verificationSucceeded) {
+    console.log(`PullApprove verification failed.\n`);
+    console.log(`Please update '.pullapprove.yml' to ensure that all necessary`);
+    console.log(`files/directories have owners and all patterns that appear in`);
+    console.log(`the file correspond to actual files/directories in the repo.`);
+  } else {
+    console.log('PullApprove verification succeeded!');
+  }
+  /**
+   * File by file Summary
+   */
+  logHeader('PullApprove file match results');
+  console.groupCollapsed(`Matched Files (${matchedFiles.size} files)`);
+  VERBOSE_MODE && Array.from(matchedFiles).forEach(file => console.log(file));
+  console.groupEnd();
+  console.groupCollapsed(`Unmatched Files (${unmatchedFiles.size} files)`);
+  for (let file of Array.from(unmatchedFiles)) { console.log(file); }
+  console.groupEnd();
+
+  /**
+   * Group by group Summary
+   */
+  logHeader('PullApprove group matches');
+  console.groupCollapsed(`Matched Groups (${matchedGroups.size} groups)`);
+  VERBOSE_MODE && logGroups(matchedGroups);
+  console.groupEnd();
+  console.groupCollapsed(`Unmatched Groups (${unmatchedGroups.size} groups)`);
+  logGroups(unmatchedGroups);
+  console.groupEnd();
+
+  // Provide correct exit code based on verification success.
+  process.exit(verificationSucceeded ? 0 : 1);
+}
+
+runVerification(ALL_FILES);
--- a/tools/pullapprove/verify.js
+++ b/tools/pullapprove/verify.js
@@ -16,6 +16,9 @@ const path = require('path');
 // Exit early on shelljs errors
 set('-e');
 
+// Regex Matcher for contains_any_globs conditions
+const CONTAINS_ANY_GLOBS_REGEX = /^'([^']+)',?$/;
+
 // Full path of the angular project directory
 const ANGULAR_PROJECT_DIR = path.resolve(__dirname, '../..');
 // Change to the Angular project directory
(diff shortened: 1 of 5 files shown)


tools/pullapprove/verify.js
```diff
@@ -0,0 +1,153 @@
...
+const exec = require('shelljs').exec;
+const path = require('path');
+
+
+// Whether to log verbosely
+const VERBOSE_MODE = !!process.argv.find(arg =>  arg === `-v`);
+// Full path to PullApprove config file
+const PULL_APPROVE_YAML_PATH = path.resolve(__dirname, '../../.pullapprove.yml')
+// All relative path file names in the git repo
+const ALL_FILES = exec('git ls-tree --full-tree -r --name-only HEAD', {silent: true}).trim().split('\n').filter(_ => _);
```
devversion: Optional: Add a comment on why we just don't use a glob? (i.e. to respect the gitignore file)

tools/pullapprove/verify.js
```diff
@@ -0,0 +1,153 @@
+#!/usr/bin/env node
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+const parse = require('yaml').parse;
```
devversion:
```suggestion
const parseYaml = require('yaml').parse;
```

tools/pullapprove/verify.js
```diff
@@ -0,0 +1,153 @@
...
+// Whether to log verbosely
+const VERBOSE_MODE = !!process.argv.find(arg =>  arg === `-v`);
+// Full path to PullApprove config file
+const PULL_APPROVE_YAML_PATH = path.resolve(__dirname, '../../.pullapprove.yml')
+// All relative path file names in the git repo
+const ALL_FILES = exec('git ls-tree --full-tree -r --name-only HEAD', {silent: true}).trim().split('\n').filter(_ => _);
+if (!ALL_FILES.length) {
+  console.error(
+    `No files were found to be in the git tree, did you run this command from \n` +
+    `inside the angular repository?`);
```
devversion: I'd propose we always `chdir` into the project directory? (like we do it commonly in scripts in `angular/components`)

tools/pullapprove/verify.js
```diff
@@ -0,0 +1,153 @@
...
+  process.exit(1);
+}
+
+/** Gets the glob matching information from each group's condition. */
+function getGlobMatcherFromConditions(groupName, condition) {
+  return condition
+    .trim().split('\n').slice(1,-1)
+    .map((glob) => {
+      const match = glob.match(/'(.*)'/);
```
devversion: I'd propose that we also match for `contains_any_globs`. The way it is right now is not very future-proof IMO. Expanding the RegEx should make things more safe. Maybe also a code snippet example for this function would be nice.

tools/pullapprove/verify.js
```diff
@@ -0,0 +1,153 @@
...
+  }
+  /**
+   * File by file Summary
+   */
+  logHeader('PullApprove file match results');
+  console.groupCollapsed(`Matched Files (${matchedFiles.size} files)`);
+  VERBOSE_MODE && Array.from(matchedFiles).forEach(file => console.log(file));
+  console.groupEnd();
+  console.groupCollapsed(`Unmatched Files (${unmatchedFiles.size} files)`);
+  for (let file of Array.from(unmatchedFiles)) { console.log(file); }
```
devversion: Optional nit: The sets already have a `forEach` method. This one could be used instead of the for loop with braces.

tools/pullapprove/verify.js
```diff
@@ -0,0 +1,153 @@
...
+}
+
+/** Logs a header within a text drawn box. */
+function logHeader(...params) {
+  const totalWidth = 80;
+  const fillWidth = totalWidth - 2;
+  const headerText = [...params].join(' ').substr(0, fillWidth);
+  const leftSpace = Math.ceil((fillWidth - headerText.length) / 2);
+  const rightSpace = fillWidth - leftSpace - headerText.length;
+  const fill = (count, content) => Array(++count).join(content);
```
devversion: I think the `fill` logic is currently flawed (hence you need the `++count`). That is because `join(content)` doesn't necessarily result in the needed amount of content being returned. How about just using `content.repeat(count)`?

tools/pullapprove/verify.js
```diff
@@ -0,0 +1,153 @@
...
+      glob: globString,
+      matcher: new Minimatch(globString, {dot: true}),
+      matchCount: 0,
+    }));
+}
+
+/** Create logs for each review group. */
+function logGroups(groups) {
+  const globGroups = Array.from(groups.entries()).sort((a, b) => a[0] > b[0] ? 1 : -1);
+  for (let globGroup of globGroups) {
```
devversion: you could destructure this into `[groupName, unsortedGlobs]`.

tools/pullapprove/verify.js
```diff
@@ -0,0 +1,153 @@
...
+  // Check each file for if it is matched by a PullApprove condition.
+  for (let file of files) {
+    const matched = allGlobs.filter(glob => glob.matcher.match(file));
+    matched.length ? matchedFiles.add(file) : unmatchedFiles.add(file);
+    matched.forEach(glob => glob.matchCount++);
+  }
+
+  // Add each glob for each group to a map either matched or unmatched.
+  allGlobs.forEach(glob => {
+    const groupsMap = glob.matchCount ? matchedGroups : unmatchedGroups;
```
devversion: This reads a little confusingly to me (personally) because `groupsMap` and `groupMap` are nearly the same. How about changing the names to `targetMap` and just `groupGlobs`?

devversion (commented):
LGTM in general. Just a few optional/nit comments.

tools/pullapprove/verify.js
josephperrott: I dont know how I didn't know a `repeat` method existed...

tools/pullapprove/verify.js
```diff
@@ -0,0 +1,176 @@
...
+  });
+
+  // PullApprove is considered verified if no files or groups are found to be unsed.
+  const verificationSucceeded = !(unmatchedFiles.size || unmatchedGroups.size);
+
+  /**
+   * Overall result
+   */
+  logHeader('Result');
+  if (verificationSucceeded) {
```
gkalpak:
```suggestion
  if (!verificationSucceeded) {
```

tools/pullapprove/verify.js
```diff
@@ -0,0 +1,176 @@
...
+            .forEach(glob => console.log(`${glob.glob} - ${glob.matchCount}`));
+        console.groupEnd();
+      });
+}
+
+/** Logs a header within a text drawn box. */
+function logHeader(...params) {
+  const totalWidth = 80;
+  const fillWidth = totalWidth - 2;
+  const headerText = [...params].join(' ').substr(0, fillWidth);
```
gkalpak:
```suggestion
  const headerText = params.join(' ').substr(0, fillWidth);
```

tools/pullapprove/verify.js
```diff
@@ -0,0 +1,176 @@
...
+             glob: globString,
+             matcher: new Minimatch(globString, {dot: true}),
+             matchCount: 0,
+           }));
+}
+
+/** Create logs for each review group. */
+function logGroups(groups) {
+  Array.from(groups.entries())
+      .sort((a, b) => { return a[0] > b[0] ? 1 : -1; })
```
gkalpak:
```suggestion
      .sort((a, b) => a[0] > b[0] ? 1 : -1)
```

gkalpak: Or even:
```suggestion
      .sort()
```

tools/pullapprove/verify.js
```diff
@@ -0,0 +1,176 @@
...
+/** Gets the glob matching information from each group's condition. */
+function getGlobMatcherFromConditions(groupName, condition) {
+  const trimmedCondition = condition.trim();
+  // If the condition doesn't start with contains_any_globs, no
+  // globs should be evaluated.
+  if (!trimmedCondition.startsWith('contains_any_globs')) {
+    return [];
+  }
+  return trimmedCondition.split('\n')
+      .slice(1, -1)
```
gkalpak: Why the `(1, -1)`?

tools/pullapprove/verify.js
```diff
@@ -0,0 +1,176 @@
...
+      .map((glob) => {
+        const match = glob.match(/'(.*)'/);
+        return match && match[1];
```
gkalpak: Should we rather fail if the glob pattern doesn't match the expected format? Are there other formats that we expect to use?

gkalpak (changes requested):
Nice! Gotta love the boxed text :smiley:

tools/pullapprove/verify.js
devversion: My thinking is that this has been done to omit the leading `contains_any_globs(files, [` line, and the trailing closing `])`. Maybe all of this can be more obvious if we just have a regex, and `eval` the actual array?

tools/pullapprove/verify.js
```diff
@@ -0,0 +1,176 @@
...
+    for (let condition of group.conditions) {
+      ALL_GLOBS.push(...getGlobMatcherFromConditions(groupName, condition));
+    }
+  });
+
+  // Check each file for if it is matched by a PullApprove condition.
+  for (let file of files) {
+    const matched = ALL_GLOBS.filter(glob => glob.matcher.match(file));
+    matched.length ? matchedFiles.add(file) : unmatchedFiles.add(file);
+    matched.forEach(glob => glob.matchCount++);
```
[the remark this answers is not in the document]
josephperrott: PullApprove will require reviews from both groups

tools/pullapprove/verify.js
josephperrott: I was guarding against how ugly clang format wanted it to be and didn't realize that I wouldn't need a custom sort:

```js
function logGroups(groups) {
  Array.from(groups.entries()).sort((a, b) => a[0] > b[0] ? 1 : -1).forEach(([
                                                                              groupName, globs
                                                                            ]) => {
    console.groupCollapsed(groupName);
...
```

tools/pullapprove/verify.js
josephperrott: The first and last line of the conditions rules for `contains_any_globs` are `contains_any_globs([` and `])` so we drop them from the check.  I removed this and instead allowed for the regex to remove these from considered glob lines.

tools/pullapprove/verify.js
josephperrott: The intention here is to just grab the glob itself

We go from:

```yaml
          'tools/pullapprove/**',
```

to 
```yaml
tools/pullapprove/**
```

tools/pullapprove/verify.js
gkalpak: `clang-format` never ceases to amaze me with its bad taste 😄

tools/pullapprove/verify.js
```diff
@@ -44,44 +42,40 @@ if (!ALL_FILES.length) {
...
 function getGlobMatcherFromConditions(groupName, condition) {
   const trimmedCondition = condition.trim();
-  // If the condition doesn't start with contains_any_globs, no
-  // globs should be evaluated.
-  if (!trimmedCondition.startsWith('contains_any_globs')) {
-    return [];
+  // If the condition should be starts with contains_any_globs, evaluate all of the globs
+  if (trimmedCondition.startsWith('contains_any_globs')) {
+    return trimmedCondition.split('\n')
+        .map((glob) => {
```
gkalpak: Super-nit (😇):
```suggestion
        .map(glob => {
```

tools/pullapprove/verify.js
gkalpak: Not sure how I missed the `(1, -1)` purpose :grin:
Somehow I thought it was operating on each pattern, not the whole condition string 🤷‍♂

FWIW, I prefer the old approach (with `.slice(1, -1)`), because it allows us to be more strict about the patterns.

tools/pullapprove/verify.js
gkalpak: Yes, I understand that. But what about patterns that (intentionally or not) do not follow that pattern? For example `"tools/pullapprove/**"`. If it is valid, we should match that too (currently we don't). If it is invalid, we should throw (currently we just ignore it).

I suggest we make the RegExp stricter and expect it to match each pattern; i.e. throw if a like does not match the RegExp. (We can expand it if we need to allow more glob patterns in the future.)

```js
      .map(glob => {
        const match = glob.trim().match(/^'([^']+)',?$/);
        if (!match) throw new Error(...);
        return match[1];
      })
```

devversion (approved):
LGTM. Last minor comments from @gkalpak seem good. I agree that we should throw/error if a condition doesn't strictly match what we expect.

tools/pullapprove/verify.js
```diff
@@ -0,0 +1,203 @@
...
+  process.exit(1);
+}
+
+/** Gets the glob matching information from each group's condition. */
+function getGlobMatchersFromConditions(groupName, condition) {
+  const trimmedCondition = condition.trim();
+  const globMatchers = [];
+  const badConditionLines = [];
+
+  // If the condition should be starts with contains_any_globs, evaluate all of the globs
```
gkalpak:
```suggestion
  // If the condition starts with contains_any_globs, evaluate all of the globs
```

tools/pullapprove/verify.js
```diff
@@ -0,0 +1,203 @@
...
+                      .filter(_ => _);
+if (!ALL_FILES.length) {
+  console.error(
+      `No files were found to be in the git tree, did you run this command from \n` +
+      `inside the angular repository?`);
+  process.exit(1);
+}
+
+/** Gets the glob matching information from each group's condition. */
+function getGlobMatchersFromConditions(groupName, condition) {
```
gkalpak:
```suggestion
function getGlobMatchersFromCondition(groupName, condition) {
```

tools/pullapprove/verify.js
```diff
@@ -0,0 +1,203 @@
...
+  const unmatchedGroups = new Map();
+  // All condition lines which were not able to be correctly parsed, by group.
+  const badConditionLinesByGroup = new Map();
+  // Total number of condition lines which were not able to be correctly parsed.
+  let badConditionLineCount = 0;
+
+  // Get all of the globs from the PullApprove group conditions.
+  Object.entries(parsedPullApproveGroups).map(([groupName, group]) => {
+    for (const condition of group.conditions) {
+      const [matchers, badConditions] = getGlobMatchersFromConditions(groupName, condition);
```
gkalpak:
```suggestion
      const [matchers, badConditions] = getGlobMatchersFromCondition(groupName, condition);
```

tools/pullapprove/verify.js
```diff
@@ -0,0 +1,203 @@
...
+        badConditionLinesByGroup.set(groupName, badConditions);
+        badConditionLineCount += badConditions.length;
+      }
+      allGlobs.push(...matchers);
+    }
+  });
+
+  if (badConditionLineCount) {
+    console.log(`Discovered ${badConditionLineCount} parsing errors in PullApprove conditions`);
+    console.log(`Attempted parsing using: /^'([^']+)',?$/`);
```
gkalpak: To avoid showing a confusing error in case we modify the RegExp in the future, you could store it in a variable and use it in the error message too.

```js
const re = /^'([^']+)',?$/;
...
console.log(`Attempted parsing using: ${re}`);
```

gkalpak (approved):
Some minor comments (feel free to ignore). Otherwise lgtm ✨
Nice work. I ❤️ all the comments 👌

IgorMinar (approved):
Lgtm but no tests? Hmmm 🤷🏼

Didn't the script you delete have tests?

PR #35070, opened 2020-01-30
fix(language-service): always use non-strict mode

### LINKED ISSUE #589: Truthy value check, but red highlighting for undefined
**Describe the bug**

If you check a variable in HTML has a truthy value (not null, not undefined, ...) and call then a method on it, I get an error with unknown method because VSCode believes object can be undefined.

Example:

`{{ foo && foo.bar() }}`

In my example, I consider `foo : Type | undefined;`

**To Reproduce**

Just implement a similar simple `foo` attribute in an Angular component and try to call a method in the template checking truthy first.

**Expected behavior**

No red highlighting as value is checked using `foo &&`

**Screenshots**

image

In this case, `startModelInstance` is defined as `public startModelInstance?: ModelInstance;`

kyliau:
Under strict mode, the language service fails to typecheck nullable
Symbols that have already verified to be non-null.

This generates incorrect and confusing diagnostics for users.

To work around this issue in the short term, this commit makes a few changes:

1. Always use strict=false internally in the language service
2. Rename MockHost.overrideOptions() to setCompilerOptions() to conform
to the interface of ts.server.Project

PR closes https://github.com/angular/vscode-ng-language-service/issues/589

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/language-service/src/ts_plugin.ts
+++ b/packages/language-service/src/ts_plugin.ts
@@ -12,7 +12,7 @@ import {createLanguageService} from './language_service';
 import {TypeScriptServiceHost} from './typescript_host';
 
 export function create(info: tss.server.PluginCreateInfo): tss.LanguageService {
-  const {languageService: tsLS, languageServiceHost: tsLSHost, config} = info;
+  const {languageService: tsLS, languageServiceHost: tsLSHost, config, project} = info;
   // This plugin could operate under two different modes:
   // 1. TS + Angular
   //    Plugin augments TS language service to provide additional Angular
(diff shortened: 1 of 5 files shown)


kyliau:
@ayazhafiz @andrius-pra I'm not sure if this approach is too "heavy-handed".
The other solution is to just get the non-nullable type of the receiver, but that feels more hacky than just turning off strict mode. Any thoughts?

rbuzzibarrios (approved):

ayazhafiz:
I think this is better than assuming the type is not null.

ayazhafiz (approved):

andrius-pra:
I prefer to keep strict mode based on settings in tsconfig.json. All reported errors are potential errors in run-time. I prefer to solve strict mode the issues with `!` and `?` operators. So i'm sure that i will not get `Cannot read property 'a' of undefined` and similar errors in run-time.

kyliau:
Good call actually. Maybe we could turn the Error into a Suggestion that prompts user to change their code to use the non-null assertion operator (!) and optional chaining (?).
It is a better code pattern anyway (I think).

ayazhafiz:
I feel that providing such a suggestion may have a very high false positive rate. This suggestion would appear every time a method/property is not found, even if a safe method call/property access is not appropriate (e.g. in the case of a misspelling). The other workaround is to look exactly for this kind of pattern, but this seems hacky.

PR #35079, opened 2020-01-31
ngcc: cleanup outdated artifacts

### LINKED ISSUE #35082: clean previous ngcc ivy entrpoints works when running ng update
# 🚀 Feature request

### Command

- [x] serve
- [x] update

### Description
Actually updating a project with `ng update` is not straightforward, if an update exist it update `package.json` versions and run `npm i` correctly. But now running `ng serve` always throw this known error:

> ERROR in The ngcc compiler has changed since the last ngcc build.
> Please remove "C:/Users/e.../ngcc-has-changed/node_modules" and try again.

Here we realize we need again remove `node_modules` and rerun `npm i`
This task  is long and not automatic, or something escapes me?

### Repro
Following my repro project which has fixed ng version to `9.0.0-rc.9`
1. run `git clone https://github.com/elvisbegovic/ngcc-has-changed.git && cd ngcc-has-changed && npm i && ng s`
2. kill `ng s` with Ctrl+C
3. run `ng update --next  @angular/cli @angular/core --allow-dirty && ng s`
4. error thrown

Clean previous ngcc ivy entrypoints works in `node_modules` during `ng update` command, so they could be again compiled when rerun `ng s` after update finish.

petebacondarwin:
Fixes #35082

I looked at various approaches of opting it or doing a special one-off cleaning pass, but I found that it  we can completely automate the cleaning so that the developer doesn't even need to be aware of it - and by always doing it in this way we are avoiding a whole range of awkward corner cases where we need to decide whether (and when) to run the clean up.

Note that this approach of using the backup files as the markers for cleanup means that we must clean whole packages. We cannot simply clean a single entry-point to a package. I think this is actually reasonable. Entry-points that have not been processed with be left untouched anyway by the cleaning. And if you have one entry-point that is outdated then you would want to fix all other outdated ones in the package anyway.

One thing this doesn't do is guarantee that, after you have run ngcc, that all packages in the `node_modules` folder are clean. Only those that would have been processed by ngcc are cleaned. But since we will now always clean on every run of ngcc, if there are any packages that were not cleaned on the one run of ngcc (because they were not needed) they will just get cleaned up they are needed. (The use case here is CLI).

One thing that we might need to check with the CLI team is whether they need to track changes to files due to the clean up. I think they track files that are changed by ngcc in general already so perhaps this will just work out of the box.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -727,7 +727,7 @@ jobs:
           command: node scripts/ci/update-deps-to-dist-packages.js ${COMPONENTS_REPO_TMP_DIR}/package.json dist/packages-dist/
       - run:
           name: "Running `angular/components` unit tests"
-          command: ./scripts/ci/run_angular_components_unit_tests.sh
+          command: ./scripts/ci/run_angular_components_unit_tests.sh | exit 0
 
   test_zonejs:
     executor:
@@ -871,9 +871,10 @@ workflows:
             - build-npm-packages
             - build-ivy-npm-packages
             - legacy-unit-tests-saucelabs
-      - components-repo-unit-tests:
-          requires:
-            - build-npm-packages
+      # FIXME - uncomment this job once https://github.com/angular/components/pull/18355 lands
+      # - components-repo-unit-tests:
+      #     requires:
+      #       - build-npm-packages
       - test_zonejs:
           requires:
             - setup
--- a/packages/compiler-cli/ngcc/src/packages/build_marker.ts
+++ b/packages/compiler-cli/ngcc/src/packages/build_marker.ts
@@ -6,10 +6,49 @@
  * found in the LICENSE file at https://angular.io/license
  */
 import {AbsoluteFsPath} from '../../../src/ngtsc/file_system';
+import {NGCC_PROPERTY_EXTENSION} from '../writing/new_entry_point_file_writer';
 import {PackageJsonUpdater} from '../writing/package_json_updater';
 import {EntryPointPackageJson, PackageJsonFormatProperties} from './entry_point';
 
 export const NGCC_VERSION = '0.0.0-PLACEHOLDER';
+
+/**
+ * Returns true if there is a format in this entry-point that was compiled with an outdated version
+ * of ngcc.
+ *
+ * @param packageJson The parsed contents of the package.json for the entry-point
+ */
+export function needsCleaning(packageJson: EntryPointPackageJson): boolean {
+  return Object.keys(packageJson.__processed_by_ivy_ngcc__ || {})
+      .some(property => packageJson.__processed_by_ivy_ngcc__ ![property] !== NGCC_VERSION);
+}
+
+/**
+ * Clean an build marker artifacts from the given `packageJson` object.
+ * @param packageJson The parsed contents of the package.json to modify
+ */
+export function cleanPackageJson(packageJson: EntryPointPackageJson): void {
+  if (packageJson.__processed_by_ivy_ngcc__ !== undefined) {
+    // Remove the actual marker
+    delete packageJson.__processed_by_ivy_ngcc__;
+    // Remove new format properties that have been added by ngcc
+    for (const prop of Object.keys(packageJson)) {
+      if (prop.endsWith(NGCC_PROPERTY_EXTENSION)) {
+        delete packageJson[prop];
+      }
+    }
+
+    // Also remove the prebulish script if we modified it
+    const scripts = packageJson.scripts;
+    if (scripts !== undefined && scripts.prepublishOnly) {
+      delete scripts.prepublishOnly;
+      if (scripts.prepublishOnly__ivy_ngcc_bak !== undefined) {
+        scripts.prepublishOnly = scripts.prepublishOnly__ivy_ngcc_bak;
+        delete scripts.prepublishOnly__ivy_ngcc_bak;
+      }
+    }
+  }
+}
 
 /**
  * Check whether ngcc has already processed a given entry-point format.
--- a/packages/compiler-cli/ngcc/src/packages/build_marker.ts
+++ b/packages/compiler-cli/ngcc/src/packages/build_marker.ts
@@ -5,7 +5,7 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
-import {AbsoluteFsPath, basename, dirname, isRoot} from '../../../src/ngtsc/file_system';
+import {AbsoluteFsPath} from '../../../src/ngtsc/file_system';
 import {PackageJsonUpdater} from '../writing/package_json_updater';
 import {EntryPointPackageJson, PackageJsonFormatProperties} from './entry_point';
 
@@ -14,33 +14,15 @@ export const NGCC_VERSION = '0.0.0-PLACEHOLDER';
 /**
  * Check whether ngcc has already processed a given entry-point format.
  *
- * The entry-point is defined by the package.json contents provided.
- * The format is defined by the provided property name of the path to the bundle in the package.json
- *
  * @param packageJson The parsed contents of the package.json file for the entry-point.
  * @param format The entry-point format property in the package.json to check.
- * @returns true if the entry-point and format have already been processed with this ngcc version.
- * @throws Error if the `packageJson` property is not an object.
- * @throws Error if the entry-point has already been processed with a different ngcc version.
+ * @returns true if the `format` in the entry-point has already been processed by this ngcc version,
+ * false otherwise.
  */
 export function hasBeenProcessed(
-    packageJson: EntryPointPackageJson, format: PackageJsonFormatProperties,
-    entryPointPath: AbsoluteFsPath): boolean {
-  if (!packageJson.__processed_by_ivy_ngcc__) {
-    return false;
-  }
-  if (Object.keys(packageJson.__processed_by_ivy_ngcc__)
-          .some(property => packageJson.__processed_by_ivy_ngcc__ ![property] !== NGCC_VERSION)) {
-    let nodeModulesFolderPath = entryPointPath;
-    while (!isRoot(nodeModulesFolderPath) && basename(nodeModulesFolderPath) !== 'node_modules') {
-      nodeModulesFolderPath = dirname(nodeModulesFolderPath);
-    }
-    throw new Error(
-        `The ngcc compiler has changed since the last ngcc build.\n` +
-        `Please remove "${isRoot(nodeModulesFolderPath) ? entryPointPath : nodeModulesFolderPath}" and try again.`);
-  }
-
-  return packageJson.__processed_by_ivy_ngcc__[format] === NGCC_VERSION;
+    packageJson: EntryPointPackageJson, format: PackageJsonFormatProperties): boolean {
+  return packageJson.__processed_by_ivy_ngcc__ !== undefined &&
+      packageJson.__processed_by_ivy_ngcc__[format] === NGCC_VERSION;
 }
 
 /**
--- a/packages/compiler-cli/ngcc/src/packages/build_marker.ts
+++ b/packages/compiler-cli/ngcc/src/packages/build_marker.ts
@@ -19,12 +19,12 @@ export const NGCC_VERSION = '0.0.0-PLACEHOLDER';
  * @param packageJson The parsed contents of the package.json for the entry-point
  */
 export function needsCleaning(packageJson: EntryPointPackageJson): boolean {
-  return Object.keys(packageJson.__processed_by_ivy_ngcc__ || {})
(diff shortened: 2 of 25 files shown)


packages/compiler-cli/ngcc/src/writing/cleaning/cleaning_strategies.ts
```diff
@@ -0,0 +1,62 @@
...
+ */
+export class PackageJsonCleaner implements CleaningStrategy {
+  constructor(private fs: FileSystem) {}
+  canClean(_path: AbsoluteFsPath, basename: PathSegment): boolean {
+    return basename === 'package.json';
+  }
+  clean(path: AbsoluteFsPath, _basename: PathSegment): void {
+    const packageJson = JSON.parse(this.fs.readFile(path));
+    cleanPackageJson(packageJson);
+    this.fs.writeFile(path, JSON.stringify(packageJson));
```
alxhub: Do we do this already? `JSON.stringify` will return a one-line blob with no whitespace. It's not a good experience for users if ngcc compresses their package.json like that.

I think we should at _least_ do 
```suggestion
    this.fs.writeFile(path, JSON.stringify(packageJson, null, 2));
```
which will apply some prettier formatting. But I'd prefer if we could avoid reformatting any parts of the file besides the keys we touch.

alxhub (approved):

packages/compiler-cli/ngcc/src/writing/cleaning/utils.ts
```diff
@@ -0,0 +1,23 @@
...
+ * found in the LICENSE file at https://angular.io/license
+ */
+import {AbsoluteFsPath, FileSystem} from '../../../../src/ngtsc/file_system';
+
+/**
+* Returns true if the given `path` is a directory (not a symlink) and actually exists.
+*
+* @param fs the current filesystem
+* @param path the path to check
+*/
```
gkalpak: Incorrect indentation.

packages/compiler-cli/ngcc/src/writing/cleaning/cleaning_strategies.ts
petebacondarwin: Good spot. Currently we do this:

```
      this.fs.writeFile(path, `${JSON.stringify(packageJson, null, 2)}\n`);
```

I will update this bit of code to follow this pattern - but also not to write the file at all if we didn't change it.

packages/compiler-cli/ngcc/src/writing/cleaning/utils.ts
```diff
@@ -0,0 +1,23 @@
...
+export function isLocalDirectory(fs: FileSystem, path: AbsoluteFsPath): boolean {
+  if (fs.exists(path)) {
+    const stat = fs.lstat(path);
+    return (stat.isDirectory() && !stat.isSymbolicLink());
```
gkalpak: The `isSymbolicLink()` check shouldn't be needed, since we are using `lstat`.

packages/compiler-cli/ngcc/src/writing/cleaning/cleaning_strategies.ts
gkalpak: That's what the `PackageJsonUpdater` does (`JSON.stringify(..., null, 2)` + `'\n'`). I think npm/yarn may do the same, so I believe it is fine in this case.

gkalpak: FWIW, I believe it would be preferable to do the `package,json` modifications through `PackageJsonUpdater`, but this should not block this PR anyway.

packages/compiler-cli/ngcc/src/writing/cleaning/package_cleaner.ts
```diff
@@ -0,0 +1,76 @@
...
+    for (const basename of basenames) {
+      const path = this.fs.resolve(directory, basename);
+      for (const cleaner of this.cleaners) {
+        if (cleaner.canClean(path, basename)) {
+          cleaner.clean(path, basename);
+          break;
+        }
+      }
+      // Recurse into subdirectories (note that a cleaner may have removed this path)
+      if (basename !== 'node_modules' && isLocalDirectory(this.fs, path)) {
```
gkalpak: Nit: We could probably move the `basename !== 'node_modules'` check at the top of the `for` loop and save the work associated with building the full path and iterating over all cleaners for this path.

gkalpak (approved):
Nice clean-up :tada:
I left several comments, but almost all of them are non-blocking (except for  the ones about adding `fs-extra` as a dependency). 

<sub>I noticed that you pushed some new commits as I was about to finish my review, so some of my comments might have already been addressed.</sub>

packages/compiler-cli/ngcc/test/packages/build_marker_spec.ts
```diff
@@ -194,5 +195,103 @@ runInEachFileSystem(() => {
       it('should return false if no markers exist',
          () => { expect(hasBeenProcessed({name: 'test'}, 'module')).toBe(false); });
     });
+
+    describe('needsCleaning()', () => {
+      it('should return true if any format has been compiled with a different version.', () => {
```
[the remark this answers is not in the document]
petebacondarwin: No idea. One probably sneaked in there and then I probably cut and pasted it everywhere.

IgorMinar (commented):
> Note that this approach of using the backup files as the markers for cleanup means that we must clean whole packages. We cannot simply clean a single entry-point to a package. I think this is actually reasonable. Entry-points that have not been processed with be left untouched anyway by the cleaning. And if you have one entry-point that is outdated then you would want to fix all other outdated ones in the package anyway.

sgtm, package-level consistency is desirable

> One thing this doesn't do is guarantee that, after you have run ngcc, that all packages in the node_modules folder are clean. Only those that would have been processed by ngcc are cleaned. But since we will now always clean on every run of ngcc, if there are any packages that were not cleaned on the one run of ngcc (because they were not needed) they will just get cleaned up they are needed. (The use case here is CLI).

that is ok, we don't need node_module level consistency. if someone cares about that, they can use yarn instead of npm.

packages/compiler-cli/ngcc/src/packages/build_marker.ts
```diff
@@ -6,11 +6,50 @@
...
 
+/**
+ * Returns true if there is a format in this entry-point that was compiled with an outdated version
+ * of ngcc.
+ *
+ * @param packageJson The parsed contents of the package.json for the entry-point
+ */
+export function needsCleaning(packageJson: EntryPointPackageJson): boolean {
+  return Object.keys(packageJson.__processed_by_ivy_ngcc__ || {})
+      .some(property => packageJson.__processed_by_ivy_ngcc__ ![property] !== NGCC_VERSION);
```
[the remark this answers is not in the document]
petebacondarwin: Nice!

packages/compiler-cli/ngcc/src/packages/build_marker.ts
```diff
@@ -6,11 +6,50 @@
...
+}
+
+/**
+ * Clean an build marker artifacts from the given `packageJson` object.
```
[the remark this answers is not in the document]
petebacondarwin: Actually I was going for:

```suggestion
 * Clean any build marker artifacts from the given `packageJson` object.
```

packages/compiler-cli/src/ngtsc/file_system/BUILD.bazel
```diff
@@ -9,7 +9,9 @@ ts_library(
     ]),
     deps = [
         "//packages:types",
+        "@npm//@types/fs-extra",
         "@npm//@types/node",
+        "@npm//fs-extra",
```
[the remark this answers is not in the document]
petebacondarwin: It is - I added in one of the fixup commits.

packages/compiler-cli/src/ngtsc/file_system/src/cached_file_system.ts
```diff
@@ -88,6 +88,12 @@ export class CachedFileSystem implements FileSystem {
     }
   }
 
+  removeDir(path: AbsoluteFsPath): void {
+    this.delegate.removeDir(path);
+    this.existsCache.set(path, false);
```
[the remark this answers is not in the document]
petebacondarwin: Hmm good point...

packages/compiler-cli/src/ngtsc/file_system/testing/src/mock_file_system.ts
```diff
@@ -131,6 +131,21 @@ export abstract class MockFileSystem implements FileSystem {
...
+    const [folderPath, basename] = this.splitIntoFolderAndFile(path);
+    const {entity} = this.findFromPath(folderPath);
+    if (entity === null || !isFolder(entity)) {
+      throw new MockFileSystemError(
+          'ENOENT', path,
+          `Unable to remove folder "${path}". The containing folder does not exist.`);
+    }
+    if (!isFolder(entity[basename])) {
+      throw new MockFileSystemError(
+          'ENOTDIR', path, `Unable to remove "${path}" - it is not a directory.`);
```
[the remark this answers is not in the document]
petebacondarwin: Hmm... I was copying the standard NodeJS `rmDirSync()` behaviour. But since the recursive form of that is only available in Node 12+ then we couldn't use it here. The fs-extra version can indeed remove files too.

How about we change the function name to `removeDeep()` and allow it to remove files or directories?

packages/compiler-cli/src/ngtsc/file_system/test/BUILD.bazel
```diff
@@ -12,6 +12,7 @@ ts_library(
         "//packages:types",
         "//packages/compiler-cli/src/ngtsc/file_system",
         "//packages/compiler-cli/src/ngtsc/file_system/testing",
+        "@npm//@types/fs-extra",
```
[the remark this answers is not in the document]
petebacondarwin: The vagaries of Bazel deps always confuse me.

So to "build" this TS library we only need the typings.
But to "run" code that uses the built JS of this library you need the actual package.
So I think that for this rule, it is correct. But then for things like the unit tests we need to add in the actual package.

packages/compiler-cli/ngcc/src/writing/cleaning/cleaning_strategies.ts
petebacondarwin: I guess the benefit is that it will be more consistent in writing the files out...

packages/compiler-cli/ngcc/src/writing/cleaning/package_cleaner.ts
petebacondarwin: That makes the assumption that no cleaning strategies will want to do something with `node_modules` :-P

(But since that is not the case, I can move it).

filipesilva (approved):
I don't think the CLI needs followup changes from this PR. From the CLI point of view, if ngcc runs we just purge the package.json from our file systems and reload it.

packages/compiler-cli/ngcc/src/writing/cleaning/cleaning_strategies.ts
petebacondarwin: But unfortunately these modification require removal of properties from package.json, which the current updater doesn't handle.

So I think this is too big an extra for this PR.

packages/compiler-cli/src/ngtsc/file_system/test/BUILD.bazel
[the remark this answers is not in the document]
petebacondarwin: Actually, what I mean is that the runtime package gets used because the src BUILD.bazel mentions it.
We just need the typings here so that Bazel doesn't complain that we have a non-direct dependency.

.circleci/config.yml
```diff
@@ -727,7 +727,7 @@ jobs:
           command: node scripts/ci/update-deps-to-dist-packages.js ${COMPONENTS_REPO_TMP_DIR}/package.json dist/packages-dist/
       - run:
           name: "Running `angular/components` unit tests"
-          command: ./scripts/ci/run_angular_components_unit_tests.sh
+          command: ./scripts/ci/run_angular_components_unit_tests.sh | exit 0
```
IgorMinar: in a follow up PR: can you add a FIXME comment here as well so that we don't forget to remove `| exit 0` plz

IgorMinar (approved):
at high level this LGTM, I requested one more change but let's do that in a separate PR so that we don't block this any more

.circleci/config.yml
petebacondarwin: Oops, I meant to remove that, as I commented out the job...

petebacondarwin: https://github.com/angular/angular/pull/35090/commits/285e841736451bf1bb297fa95c775b04ef69d80c

PR #35083, opened 2020-01-31
ci: remove old vendoring solution in favor of relying on yarn-path

josephperrott:
Now that bazel respects the yarn-path value found in .yarnrc, we can
remove the last remaining reliances on our vendoring in
//third_party/github.com/yarnpkg/yarn/


gregmagolan (approved):
So nice when a PR just removes unnecessary complication 👍 👍

IgorMinar (approved):

.circleci/config.yml
```diff
@@ -621,11 +603,6 @@ jobs:
           command: |
             git fetch origin $CI_STABLE_BRANCH
             git checkout --force origin/$CI_STABLE_BRANCH -- aio/ .yarn/ .yarnrc
-        # Overwrite yarn again to use the version from the checked out branch.
-      - overwrite_yarn
-        # Ignore yarn's engines check, because we checked out `aio/package.json` from the stable
-        # branch and there could be a node version skew, which is acceptable in this monitoring job.
-      - run: yarn config set ignore-engines true
```
gkalpak: This line (and the associated comments above) is still required.
(NOTE: This is about ignore the check for the Node.js version.)

gkalpak (approved):
One comment that needs to be addressed.
Otherwise ❤️ 💯 😍 🍾 🎊

.circleci/config.yml
IgorMinar: I pushed a fixup commit for this and rebased the PR

PR #34374, opened 2019-12-12
WIP - build(docs-infra): update docs examples to latest v9 and Ivy

gkalpak:
Jira issue: [FW-1609](https://angular-team.atlassian.net/browse/FW-1609)

##
Blocked on:
- [x] ~~#33484 (+updating `aio/tools/examples/shared/boilerplate/elements/package.json` according to ` aio/tools/examples/shared/boilerplate/cli/package.json`)~~ _Landed in #36015._
- [x] ~~#34508~~ (+updating `zone.js` dependency version) _Landed in #36015._
- [x] ~~#35778~~
- [x] ~~#35780~~
- [x] ~~#35858~~
- [x] ~~#36015~~ (orignally part of this PR - split off for easier review/merging)
- [x] ~~#36143~~ (orignally part of this PR - split off for easier review/merging)
- [x] ~~#36483 (orignally part of this PR - split off for easier review/merging)~~

##
Fixes #34382, #34383 and #35289.

This PR may also partially address: #35776

PR #35091, opened 2020-01-31
ci: remove test_docs_examples_ivy job

josephperrott:
Removing test_docs_examples_ivy job from CI as it is testing a scenario that
does not currently exist and likely will not exist in the future, of
applications building based on libraries using ivy instructions directly. We
expect to continue to instread rely on ngcc converted code for usage of
libraries.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -420,10 +420,6 @@ jobs:
       - run: ./aio/aio-builds-setup/scripts/test.sh
 
   test_docs_examples:
-    parameters:
-      ivy:
-        type: boolean
-        default: false
     executor:
       # Needed because the example e2e tests depend on Chrome.
       name: browsers-executor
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -434,19 +430,10 @@ jobs:
...
       - run: yarn --cwd aio install --frozen-lockfile --non-interactive
-      - when:
-          condition: << parameters.ivy >>
-          steps:
-              # Rename the Ivy packages dist folder to "dist/packages-dist" as the AIO
-              # package installer picks up the locally built packages from that location.
-              # *Note*: We could also adjust the packages installer, but given we won't have
-              # two different folders of Angular distributions in the future, we should keep
-              # the packages installer unchanged.
-            - run: mv dist/packages-dist-ivy-aot dist/packages-dist
```
gkalpak: This is not about testing the scenario of using the ivy-built libraries. What this job does is check the docs examples mini-apps in Ivy mode. Using the ivy-build Angular packages is just a way of speeding up the build (i.e. not requiring ngcc to compile the regular Angular libraries).

If you want to get rid of the `packages-dist-ivy-aot/` dir, you can just remove this line (in which case the examples will use the regular Angular packages from `packages-dist/` and cmpile them to Ivy via ngcc). But this job should not be removed imo.

gkalpak (changes requested):

.circleci/config.yml
IgorMinar: @gkalpak I see. thanks for the clarification. The issue I have with the current test setup is that it doesn't test what users use in CLI and thus it doesn't represent a real use-case. Can you please send a PR to have the Ivy examples tested against the dist/packages-dist output instead as part of #34374 thanks!

IgorMinar:
closing per discussion above - let's resolve the pending work in #34374

PR #35096, opened 2020-02-01
feat(language-service): provide completions for microsyntax

ivanwonder:
Provide completions for microsyntax, for example, `ngFor` completions will include `of`, `trackBy`, etc. `ngIf` completions will include `then`, `else`.
```
*ngFor="let item of items; | ;"
```
Here provide template context(`index`, `odd`, etc.) and directive input property(`of`, `trackBy`, etc.).
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/packages/language-service/src/completions.ts
+++ b/packages/language-service/src/completions.ts
@@ -6,7 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {AST, AstPath, AttrAst, Attribute, BoundDirectivePropertyAst, BoundElementPropertyAst, BoundEventAst, BoundTextAst, Element, ElementAst, HtmlAstPath, NAMED_ENTITIES, Node as HtmlAst, NullTemplateVisitor, ReferenceAst, TagContentType, TemplateBinding, Text, getHtmlTagDefinition} from '@angular/compiler';
+import {AST, AstPath, AttrAst, Attribute, BoundDirectivePropertyAst, BoundElementPropertyAst, BoundEventAst, BoundTextAst, CompileDirectiveSummary, Element, ElementAst, HtmlAstPath, NAMED_ENTITIES, Node as HtmlAst, NullTemplateVisitor, ReferenceAst, TagContentType, TemplateBinding, Text, getHtmlTagDefinition} from '@angular/compiler';
 import {$$, $_, isAsciiLetter, isDigit} from '@angular/compiler/src/chars';
 
 import {AstResult} from './common';
@@ -15,7 +15,7 @@ import {getExpressionCompletions} from './expressions';
 import {attributeNames, elementNames, eventNames, propertyNames} from './html_info';
 import {InlineTemplate} from './template';
 import * as ng from './types';
-import {diagnosticInfoFromTemplateInfo, findTemplateAstAt, getPathToNodeAtPosition, getSelectors, hasTemplateReference, inSpan, spanOf} from './utils';
+import {diagnosticInfoFromTemplateInfo, findTemplateAstAt, getPathToNodeAtPosition, getSelectorByName, getSelectors, hasTemplateReference, inSpan, spanOf} from './utils';
 
 const HIDDEN_HTML_ELEMENTS: ReadonlySet<string> =
     new Set(['html', 'script', 'noscript', 'base', 'body', 'title', 'head', 'link']);
@@ -457,14 +457,30 @@ class ExpressionVisitor extends NullTemplateVisitor {
           ast.name, ast.value, ast.sourceSpan.toString(), ast.sourceSpan.start.offset);
       // Find where the cursor is relative to the start of the attribute value.
       const valueRelativePosition = this.position - ast.sourceSpan.start.offset;
-      // Find the template binding that contains the position.
-      const binding = templateBindings.find(b => inSpan(valueRelativePosition, b.span));
 
-      if (!binding) {
+      const key = ast.name.substring(1);  // remove leading asterisk
+
+      // Find the selector - eg ngFor, ngIf, etc
+      const selectorInfo = getSelectors(this.info);
+      const selector = getSelectorByName(key, selectorInfo.selectors);
+
+      if (!selector) {
         return;
       }
 
-      this.microSyntaxInAttributeValue(ast, binding);
+      const directiveMetadata = selectorInfo.map.get(selector) !;
+
+      // Find the template binding that contains the position.
+      const binding = templateBindings.find(b => inSpan(valueRelativePosition, b.span));
+      if (!binding) {
+        // '*ngFor="let item of items; | ;"'
+        //                             ^------ cursor
+        this.addTemplateContextCompletions(directiveMetadata);
+        this.addDirectiveInputCompletions(ast, directiveMetadata);
+        return;
+      }
+
+      this.microSyntaxInAttributeValue(ast, binding, directiveMetadata);
     } else {
       const expressionAst = this.info.expressionParser.parseBinding(
           ast.value, ast.sourceSpan.toString(), ast.sourceSpan.start.offset);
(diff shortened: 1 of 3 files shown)


packages/language-service/src/completions.ts
```diff
@@ -460,11 +460,34 @@ class ExpressionVisitor extends NullTemplateVisitor {
...
+
+      // Find the selector - eg ngFor, ngIf, etc
+      const selectorInfo = getSelectors(this.info);
+      const selector = selectorInfo.selectors.find(s => {
+        // attributes are listed in (attribute, value) pairs
+        for (let i = 0; i < s.attrs.length; i += 2) {
+          if (s.attrs[i] === key) {
+            return true;
+          }
+        }
```
ayazhafiz: I think we use this pattern in several other places in the LS. we don't have to do it in this PR, but should consider a util function

packages/language-service/src/completions.ts
```diff
@@ -460,11 +460,34 @@ class ExpressionVisitor extends NullTemplateVisitor {
...
+        return;
+      }
+
+      const directiveMetadata = selectorInfo.map.get(selector) !;
+
       if (!binding) {
+        // '*ngFor="let item of items; | ;"'
+        //                             ^------ cursor
+        this.addTemplateContextToCompletions(directiveMetadata);
+        this.addStructuralDirectiveInputToCompletions(ast, directiveMetadata);
```
ayazhafiz: nit: change to `addTemplateContextCompletions` and `addDirectiveInputCompletions` for brevity

packages/language-service/src/completions.ts
```diff
@@ -460,11 +460,34 @@ class ExpressionVisitor extends NullTemplateVisitor {
       // Find the template binding that contains the position.
       const binding = templateBindings.find(b => inSpan(valueRelativePosition, b.span));
```
ayazhafiz: Can you move this down to line 482 since there is now the large block to find the directive?

packages/language-service/src/completions.ts
```diff
@@ -528,59 +551,106 @@ class ExpressionVisitor extends NullTemplateVisitor {
...
-
-    if (!selector) {
-      return;
-    }
-
+  private microSyntaxInAttributeValue(
+      attr: AttrAst, binding: TemplateBinding, directiveMetadata: CompileDirectiveSummary) {
     const valueRelativePosition = this.position - attr.sourceSpan.start.offset;
 
+    const bindingSource = attr.value.substring(binding.span.start, binding.span.end);
```
ayazhafiz: can we do `toString` on the binding? I wouldn't be surprised if we can't, just wondering..

ayazhafiz: consider calling `trim` on the construction of the source here instead of on the next line

packages/language-service/src/completions.ts
```diff
@@ -528,59 +551,106 @@ class ExpressionVisitor extends NullTemplateVisitor {
...
-            // This adds symbols like $implicit, index, count, etc.
-            this.addSymbolsToCompletions(contextTable.values());
-            return;
-          }
+      const equalLocation = bindingSource.indexOf('=');
+      const asLocation = bindingSource.indexOf(' as ');
+      if (equalLocation > 0 && valueRelativePosition > equalLocation + binding.span.start ||
+          (asLocation > 0 && asLocation + binding.span.start >= valueRelativePosition)) {
+        // We are after the '=' or before the 'as' in a let clause. The valid values here are the
+        // members of the template reference's type parameter.
```
```suggestion
        // members of the directive's template context.
```

would you mind changing this while you're here? i think the original comment obscures what we're trying to do.

packages/language-service/src/completions.ts
```diff
@@ -528,59 +551,106 @@ class ExpressionVisitor extends NullTemplateVisitor {
...
+        this.addTemplateContextToCompletions(directiveMetadata);
+      }
+      if (binding.key && binding.name === '$implicit' && endOfSyntax) {
+        // '*ngFor="let x |;'
+        //                ^------ cursor
+        // The binding span contain the white space before ';',
+        // so provide template context and directive input property in here.
```
ayazhafiz: This doesn't happen only if we are before a semicolon, just if we're at the end of the binding right? Consider rewording the comment to something like `We are at the end of the binding span, so...`. Please correct me if I'm wrong.

packages/language-service/src/completions.ts
```diff
@@ -528,59 +551,106 @@ class ExpressionVisitor extends NullTemplateVisitor {
...
+        this.addTemplateContextToCompletions(directiveMetadata);
+        this.addStructuralDirectiveInputToCompletions(attr, directiveMetadata);
+        return;
+      }
+    } else {
+      // '*ngFor="let x of item; trackBy| ;"'
+      //                                ^------ cursor
+      // The binding expression should be null,
```
ayazhafiz: What do you mean by should be null? Wouldn't this branch be hit if the binding name wasn't implicit or we were before the end of the binding as well?

packages/language-service/src/completions.ts
```diff
@@ -528,59 +551,106 @@ class ExpressionVisitor extends NullTemplateVisitor {
...
-          }
+      const equalLocation = bindingSource.indexOf('=');
+      const asLocation = bindingSource.indexOf(' as ');
+      if (equalLocation > 0 && valueRelativePosition > equalLocation + binding.span.start ||
+          (asLocation > 0 && asLocation + binding.span.start >= valueRelativePosition)) {
+        // We are after the '=' or before the 'as' in a let clause. The valid values here are the
+        // members of the template reference's type parameter.
+        this.addTemplateContextToCompletions(directiveMetadata);
+      }
+      if (binding.key && binding.name === '$implicit' && endOfSyntax) {
```
```suggestion
      else if (binding.key && binding.name === '$implicit' && endOfSyntax) {
```

packages/language-service/src/completions.ts
```diff
@@ -528,59 +551,106 @@ class ExpressionVisitor extends NullTemplateVisitor {
...
+        this.addTemplateContextToCompletions(directiveMetadata);
+        this.addStructuralDirectiveInputToCompletions(attr, directiveMetadata);
+        return;
+      }
+    } else {
+      // '*ngFor="let x of item; trackBy| ;"'
+      //                                ^------ cursor
+      // The binding expression should be null,
+      // but the expression here exist and 'binding.expression.span' is {start:0, end: 0}.
+      const expressionExist =
```
```suggestion
      const expressionExists =
```

packages/language-service/src/completions.ts
```diff
@@ -528,59 +551,106 @@ class ExpressionVisitor extends NullTemplateVisitor {
...
+          this.addTemplateContextToCompletions(directiveMetadata);
+          this.addStructuralDirectiveInputToCompletions(attr, directiveMetadata);
+          return;
         }
+        this.processExpressionCompletions(binding.expression !.ast);
+        return;
+      }
+
+      // '*ngFor="let item of items;"'
+      // The 'of' have no span, so get it here.
```
```suggestion
      // The 'of' has no span, so get it here.
```

packages/language-service/src/completions.ts
```diff
@@ -528,59 +551,106 @@ class ExpressionVisitor extends NullTemplateVisitor {
...
+      const bindingKeySpan: ng.Span = {
+        start: binding.span.start,
+        end: binding.span.start + binding.key.length - attr.name.length
+      };
+      if (inSpan(valueRelativePosition, bindingKeySpan)) {
+        if (!expressionExist) {
+          // '*ngFor="let item of items; trackBy| ;"'
+          //                                    ^------ cursor
+          // The template context should be provided here because of 'index as i'.
```
ayazhafiz: I am confused by this comment. Can you elaborate? How is this different from the branch on line 581?

packages/language-service/src/completions.ts
```diff
@@ -528,59 +551,106 @@ class ExpressionVisitor extends NullTemplateVisitor {
...
+          this.addTemplateContextToCompletions(directiveMetadata);
+        }
+        // '*ngFor="let item of items; trackBy|: myTrack;"'
+        //                                    ^------ cursor
+        this.addStructuralDirectiveInputToCompletions(attr, directiveMetadata);
+        return;
+      }
+
+      if (attr.value.substring(valueRelativePosition, binding.span.end).indexOf(':') === -1) {
```
ayazhafiz: I wonder if there is a way to rearrange these branches so that it is more clear that this is the only option if `:` is not found in the binding. I will think about it a little bit more; it is not a big deal.

packages/language-service/src/completions.ts
```diff
@@ -528,59 +551,106 @@ class ExpressionVisitor extends NullTemplateVisitor {
...
+        //                                      ^------ cursor
+        this.addSymbolsToCompletions(this.getExpressionScope().values());
       }
     }
+  }
 
-    if (binding.expression && inSpan(valueRelativePosition, binding.expression.ast.span)) {
-      this.processExpressionCompletions(binding.expression.ast);
-      return;
+  // Add the structural directive's template context(e.g. 'ngFor' has 'index', 'odd', etc.) to
```
```suggestion
  // Add the structural directive's template context (e.g. 'ngFor' has 'index', 'odd', etc.) to
```

packages/language-service/src/completions.ts
```diff
@@ -528,59 +551,106 @@ class ExpressionVisitor extends NullTemplateVisitor {
...
-      const expressionAst = this.info.expressionParser.parseBinding(
-          attr.value, attr.sourceSpan.toString(), attr.sourceSpan.start.offset);
-      this.processExpressionCompletions(expressionAst);
+  // Add the structural directive's input property(e.g. 'ngFor' has 'of', 'trackBy', etc.) to
+  // completions.
+  private addStructuralDirectiveInputToCompletions(
+      attr: AttrAst, directiveMetadata: CompileDirectiveSummary) {
+    const key = attr.name.substring(1);
+    for (const input of Object.values(directiveMetadata.inputs)) {
+      const _input = input.substring(key.length);
```
```suggestion
      const unprefixedInput = input.substring(key.length);
```

packages/language-service/src/completions.ts
```diff
@@ -528,59 +551,106 @@ class ExpressionVisitor extends NullTemplateVisitor {
...
+      if (_input) {
+        const _inputInsert = _input[0].toLowerCase() + _input.substring(1);
```
```suggestion
        const inputName = _input[0].toLowerCase() + _input.substring(1);
```

packages/language-service/src/completions.ts
ayazhafiz: Consider also adding an example (`ngForTrackBy` -> `trackBy`). This is one of the most magical things in the compiler and can be kind of confusing.

packages/language-service/test/completions_spec.ts
```diff
@@ -320,6 +304,53 @@ describe('completions', () => {
       expectContain(completions, CompletionKind.VARIABLE, ['x']);
     });
 
+    it('should provide suggestion for template context and input name', () => {
+      mockHost.override(
+          TEST_TEMPLATE,
+          `<div *ngFor="let x	~{cursor-one}; of heroes ~{cursor-two}; tr~{cursor-three};"></div>`);
```
ayazhafiz: For cursor-three, it looks like we also want to see `trackBy`. Is that correct?

packages/language-service/test/completions_spec.ts
```diff
@@ -320,6 +304,53 @@ describe('completions', () => {
...
+      const marker = mockHost.getLocationMarkerFor(TEST_TEMPLATE, 'cursor');
+      const completions = ngLS.getCompletionsAtPosition(TEST_TEMPLATE, marker.start);
+      expectContain(completions, CompletionKind.PROPERTY, ['of', 'trackBy']);
+    });
+
+    it('should not provide suggestion before the : sign', () => {
+      mockHost.override(TEST_TEMPLATE, `<div *ngFor="let x; trackBy ~{cursor}: myTrack;"></div>`);
+      const marker = mockHost.getLocationMarkerFor(TEST_TEMPLATE, 'cursor');
+      const completions = ngLS.getCompletionsAtPosition(TEST_TEMPLATE, marker.start);
+      expect(completions).toBeUndefined();
```
ayazhafiz: :+1:

ayazhafiz (changes requested):
Great feature! I'm excited for this one )

packages/language-service/src/completions.ts
ivanwonder: `*ngFor="let x of item; tr| ;"`
the templateBindings is
```
span: ParseSpan {start: 20, end: 22}
key: "*ngForTr"
keyIsVar: false
name: null
expression: ASTWithSource
  span: ParseSpan {start: 0, end: 0}
  sourceSpan: AbsoluteSourceSpan {start: 35, end: 35}
  ast: EmptyExpr {span: ParseSpan, sourceSpan: AbsoluteSourceSpan}
  source: ""
  location: "let x	; of heroes ; tr;"
  errors: (2) [ParserError, ParserError]
```
`*ngFor="let x of item; tr| "`
If remove the `;`, the templateBindings is
```
span: ParseSpan {start: 20, end: 22}
key: "*ngForTr"
keyIsVar: false
name: null
expression: null
```
Before this PR,  it only check `binding.expression && inSpan()`.

packages/language-service/src/completions.ts
ivanwonder: `*ngFor="let item of items; trackBy| ;"`
If the cursor is here, there are two conditions, one is `index as i`, another is `trackBy: myTrack`.
And the expression doesn't exist here or is null, so I will provide `index` or `trackBy`. 
`*ngFor="let item of items; trackBy| : myTrack ;"`
Now the expression exists, it's `myTrack`, I only provide `trackBy`, `of`, etc.

packages/language-service/test/completions_spec.ts
ivanwonder: Yes, here I only list `of` to represent the directive input and  `index` to represent template context.
Need to list all of them?

packages/language-service/src/completions.ts
ivanwonder: return [object object]....

packages/language-service/src/completions.ts
ivanwonder: In this branch, there are only three conditions, in directive input name, in expression, and after the `:`.

packages/language-service/test/completions_spec.ts
ayazhafiz: I would recommend changing the test case or adding `trackBy`. It could be confusing to someone reading the code

packages/language-service/src/completions.ts
ayazhafiz: So in this situation we would only want to provide `... as ...` completions, is that correct? If so, could you reword the comment?

ivanwonder: No, here we need to provide the template context(e.g. `index`, `odd`, etc.) and directive input(e.g. `of`, `trackBy`).

packages/language-service/test/completions_spec.ts
ivanwonder: Just add comment there.

packages/language-service/src/completions.ts
ivanwonder: `*ngFor="let item of items; trackBy| ;"`
the completions include template context and directive input.
`*ngFor="let item of items; trackBy| : myTrack ;"`
the completions only include directive input.

ayazhafiz: Okay, I understand. Sorry, the comment was confusing me -- it sounds like only the template context should be provided, because of the example `index as i`, and it is not clear where `index as i` comes from. Could you reword the comment to explain that in this case, we should provide both template context member and input completions?

packages/language-service/src/completions.ts
ivanwonder: Does it better now?

jessicajaniuk:
Closing this since View Engine is no more making this no longer necessary.

PR #35097, opened 2020-02-01
perf(ivy): remove unused event argument in listener instructions

crisbeto:
Currently Ivy always generates the `$event` function argument, even if it isn't being used by the listener expressions. This can lead to unnecessary bytes being generated, because optimizers won't remove unused arguments by default. These changes add some logic to avoid adding the argument when it isn't required.

### DIFF
--- a/packages/compiler/src/render3/view/template.ts
+++ b/packages/compiler/src/render3/view/template.ts
@@ -69,13 +69,14 @@ export function prepareEventListenerParameters(
         Supported list of global targets: ${Array.from(GLOBAL_TARGET_RESOLVERS.keys())}.`);
   }
 
+  const eventArgumentName = '$event';
+  const implicitReceiverAccesses = new Set<string>();
   const implicitReceiverExpr = (scope === null || scope.bindingLevel === 0) ?
       o.variable(CONTEXT_NAME) :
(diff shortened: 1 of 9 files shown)


packages/compiler/src/render3/view/template.ts
```diff
@@ -86,9 +87,13 @@ export function prepareEventListenerParameters(
   const eventName: string =
       type === ParsedEventType.Animation ? prepareSyntheticListenerName(name, phase !) : name;
   const fnName = handlerName && sanitizeIdentifier(handlerName);
-  const fnArgs = [new o.FnParam('$event', o.DYNAMIC_TYPE)];
-  const handlerFn = o.fn(fnArgs, statements, o.INFERRED_TYPE, null, fnName);
+  const fnArgs: o.FnParam[] = [];
 
+  if (implicitReceiverAccesses.has(eventArgumentName)) {
```
crisbeto: I had a nicer alternate solution which detected usages of `$event` through the `BindingScope`, but the problem was that we don't have a binding scope for host bindings hence why I ended up doing it through the visitor.

alxhub (approved):

mhevery:
[presubmit](http://test/OCL:296246108:BASE:296246152:1582224229916:91123448)

Issue #35095, opened 2020-02-01
Compilation breaks after update to 9.0.0-rc.13: `Duplicate identifier 'ɵngcc0'`


cyrilletuzi:
# 🐞 bug report

### Affected Package
I think the issue is in `@angular/compiler-cli` as it's about `ngcc`, but not sure.

### Is this a regression?

Yes, it was OK in previous RC (rc.12 and below).

### Reproduction

- `npm install @angular/cli@next -g`
- `ng new helloworld --skipGit` (git is disabled to avoid ng commands blocking because repo is not clean, otherwise you have to commit every change)
- `cd helloworld`
- `npm install @angular/{animations,common,compiler,core,forms,platform-browser,platform-browser-dynamic,router,compiler-cli,language-service}@9.0.0-rc.12` (rollback to rc.12, as CLI installed rc.13)
- add any library, for example: `ng add @angular/material@next` and add `MatCardModule` in `AppModule` imports
- `npm start` (to run the compilation with rc.12)
- `ng update @angular/core --next` (will update to rc.13)
- `npm start` 

## 🔥 Exception or Error

Compilation breaks with a lot of errors like this one (it's just an example):

<pre><code>
import * as ɵngcc0 from '@angular/core';
                   ~~~~~~
    node_modules/@angular/material/core/selection/pseudo-checkbox/pseudo-checkbox.d.ts:13:13 -
error TS2300: Duplicate identifier 'ɵngcc0'.
</code></pre>


## 🌍  Your Environment

**Angular Version:**
<pre><code>
Angular CLI: 9.0.0-rc.12
Node: 12.14.1
OS: darwin x64

Angular: 9.0.0-rc.13
... animations, common, compiler, compiler-cli, core, forms
... language-service, platform-browser, platform-browser-dynamic
... router
Ivy Workspace: Yes

Package                           Version
-----------------------------------------------------------
@angular-devkit/architect         0.900.0-rc.12
@angular-devkit/build-angular     0.900.0-rc.12
@angular-devkit/build-optimizer   0.900.0-rc.12
@angular-devkit/build-webpack     0.900.0-rc.12
@angular-devkit/core              9.0.0-rc.12
@angular-devkit/schematics        9.0.0-rc.12
@angular/cdk                      9.0.0-rc.9
@angular/cli                      9.0.0-rc.12
@angular/material                 9.0.0-rc.9
@ngtools/webpack                  9.0.0-rc.12
@schematics/angular               9.0.0-rc.12
[...]

PR #35101, opened 2020-02-01
integration test for #35095

IgorMinar:
mimics:

```
ng new ngcc-cleanup
yarn add @angular/{animations,common,compiler,core,forms,platform-browser,platform-browser-dynamic,router,compiler-cli,language-service}@9.0.0-rc.12
ng add @angular/material@next
ng generate @angular/material:nav my-navigation
ng generate @angular/material:dashboard my-dashboard
ng build
ng update @angular/core@next
ng build
```


IgorMinar:
I can't reproduce the issue for some reason. @cyrilletuzi can you please take a look at https://github.com/angular/angular/pull/35101/files#diff-52db63aae44ee042dbf0c1ff4536cd0eR8-R9 and check if I missed something.

gkalpak:
In #35095, @cyrilletuzi mentioned this happens when updating to rc.13 while `ng serve` is running (so it might be the case that `ng build` is not affected). Could it be that the CLI (or ngtsc) is caching some files during `ng serve`, so they do not see the updated content after ngcc cleaning up?

cyrilletuzi:
@IgorMinar I can definitely reproduce the issue:
- with your steps in PR message
- with your PR git strategy (rollback `package.json` and install again).

The only difference is that I use npm (I don't have yarn). So I would suspect that `yarn install` does more cleaning that `npm install`? 

@gkalpak Nope, I can reproduce with `ng build` too.

Doing a reproduction repo right now, I'll post it in a few seconds.

cyrilletuzi:
Repro: [https://github.com/cyrilletuzi/ngcc-cleanup](https://github.com/cyrilletuzi/ngcc-cleanup)

To reproduce:
```bash
git clone https://github.com/cyrilletuzi/ngcc-cleanup
cd ngcc-cleanup
npm run issue
```

Steps done to generate project:
```bash
ng new ngcc-cleanup
cd ngcc-cleanup
ng add @angular/material@next
ng generate @angular/material:nav my-navigation
git add -A
git commit -m "test"
```

elvisbegovic:
Hm If we consider this https://github.com/angular/angular/issues/35082#issuecomment-579823144 Pete’s comment Yarn doesn’t have the problem but npm yes.

cyrilletuzi:
@elvisbegovic Yeah I think it's the issue.

cyrilletuzi:
Note that it only affects projects which are upgrading from v8 with Ivy or v9-beta/rc, ie. experimental scenarios. So it's a blocker for a 9.0.1 release but for the 9.0.0 release it's not that critical.

About the issue:  how does #35079 detect which packages must be cleaned up? Is that possible that librairies need to be rebuilt with the very last rc to be detected and cleaned up?

IgorMinar:
As was already mentioned, it just hit me this morning that I have to use npm and not yarn to repro this because yarn discards and resets node_modules during each `install` or `add/remove`.

gkalpak:
@IgorMinar, I added a quick (and hacky) test in the `ngcc` integration project in #35106.
We can add a proper integration test (featuring `ng build`) in a follow-up PR.

IgorMinar:
@gkalpak added a sufficient integration test in #35106 so I'm going to close this PR.

PR #35105, opened 2020-02-02
fix(core): make subclass inherit developer-defined data

### LINKED ISSUE #34999: Animations metadata is not inherited to sub-classed components
In Ivy, the `@Component({...})` metadata is inherited from a parent class. However, it appears that `animations` is not included.

This causes an issue when a parent class component defines animations host bindings and an animations definition, but the subclassed component only receives the host bindings.

The renderer throws an error but it is not helpful:
`Found the synthetic property @state. Please include either "BrowserAnimationsModule" or "NoopAnimationsModule" in your application.`

Reproduction: https://stackblitz.com/edit/ivy-ejv3ha?file=src/app/hello.component.ts

Zuckjet:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

## Other information
Fixes  [#34999](https://github.com/angular/angular/issues/34999)

### DIFF
--- a/packages/core/src/render3/features/inherit_definition_feature.ts
+++ b/packages/core/src/render3/features/inherit_definition_feature.ts
@@ -7,11 +7,10 @@
  */
 
 import {Type, Writable} from '../../interface/type';
-import {assertEqual} from '../../util/assert';
 import {fillProperties} from '../../util/property';
 import {EMPTY_ARRAY, EMPTY_OBJ} from '../empty';
 import {ComponentDef, ContentQueriesFunction, DirectiveDef, DirectiveDefFeature, HostBindingsFunction, RenderFlags, ViewQueriesFunction} from '../interfaces/definition';
-import {AttributeMarker, TAttributes} from '../interfaces/node';
+import {TAttributes} from '../interfaces/node';
 import {isComponentDef} from '../interfaces/type_checks';
 import {mergeHostAttrs} from '../util/attrs_utils';
 
@@ -70,6 +69,11 @@ export function ɵɵInheritDefinitionFeature(definition: DirectiveDef<any>| Comp
         fillProperties(definition.inputs, superDef.inputs);
         fillProperties(definition.declaredInputs, superDef.declaredInputs);
         fillProperties(definition.outputs, superDef.outputs);
+
+        // Merge developer-defined data
+        (superDef as ComponentDef<any>).data &&
+            inheritData(
+                (definition as ComponentDef<any>).data, (superDef as ComponentDef<any>).data);
 
         // Inherit hooks
         // Assume super class inheritance feature has already run.
@@ -179,4 +183,9 @@ function inheritHostBindings(
   } else {
     definition.hostBindings = superHostBindings;
   }
+}
+
+function inheritData(defData: {[kind: string]: any}, superData: {[kind: string]: any}) {
+  superData.animation &&
+      (defData.animation = (defData.animation || []).concat(superData.animation));
 }
--- a/packages/core/src/render3/features/inherit_definition_feature.ts
+++ b/packages/core/src/render3/features/inherit_definition_feature.ts
@@ -70,10 +70,14 @@ export function ɵɵInheritDefinitionFeature(definition: DirectiveDef<any>| Comp
         fillProperties(definition.declaredInputs, superDef.declaredInputs);
         fillProperties(definition.outputs, superDef.outputs);
 
-        // Merge developer-defined data
-        (superDef as ComponentDef<any>).data &&
-            inheritData(
-                (definition as ComponentDef<any>).data, (superDef as ComponentDef<any>).data);
+        // Merge animations metadata.
+        // If `superDef` is a Component, the `data` field is present (defaults to an empty object).
+        if (isComponentDef(superDef) && superDef.data.animation) {
+          // If super def is a Component, the `definition` is also a Component, since Directives can
+          // not inherit Components (we throw an error above and cannot reach this code).
+          const defData = (definition as ComponentDef<any>).data;
+          defData.animation = (defData.animation || []).concat(superDef.data.animation);
+        }
 
         // Inherit hooks
         // Assume super class inheritance feature has already run.
@@ -183,9 +187,4 @@ function inheritHostBindings(
   } else {
     definition.hostBindings = superHostBindings;
   }
-}
-
-function inheritData(defData: {[kind: string]: any}, superData: {[kind: string]: any}) {
-  superData.animation &&
-      (defData.animation = (defData.animation || []).concat(superData.animation));
 }
--- a/packages/core/test/acceptance/inherit_definition_feature_spec.ts
+++ b/packages/core/test/acceptance/inherit_definition_feature_spec.ts
@@ -3518,110 +3518,6 @@ describe('inheritance', () => {
       });
     });
 
-    describe('animations', () => {
-      onlyInIvy('View Engine does not provide this check')
-          .it('should work with inherited host bindings', () => {
-            @Directive({
-              selector: '[super-dir]',
-              host: {
-                '[@animation1]': 'colorExp',
-              },
-            })
-            class SuperDirective {
-              colorExp = 'color';
-            }
-
-            class BareClass extends SuperDirective {}
-
-            @Component({
-              selector: 'my-comp',
-              template: `<h3>my-comp</h3>`,
-              animations: [trigger(
-                  'animation1',
-                  [
-                    state('color', style({color: 'red'})),
-                  ])],
-            })
-            class MyComponent extends BareClass {
-            }
-
-            @Component({template: `<my-comp>app</my-comp>`})
-            class App {
-            }
-
-            TestBed.configureTestingModule({
-              declarations: [App, MyComponent, SuperDirective],
-              imports: [NoopAnimationsModule],
-            });
-            const fixture = TestBed.createComponent(App);
-            fixture.detectChanges();
-            const queryResult = fixture.debugElement.query(By.css('my-comp'));
-
-            expect(queryResult.nativeElement.style.color).toBe('red');
-          });
-      onlyInIvy('View Engine does not provide this check')
-          .it('should work with composed host bindings', () => {
-            @Directive({
-              selector: '[super-dir]',
-              host: {
-                '[@animation1]': 'colorExp',
-                '[@animation2]': 'opacityExp',
-              },
-            })
-            class SuperDirective {
-              colorExp = 'color';
-              opacityExp = 'opacity';
-            }
-
-            class BareClass extends SuperDirective {}
-
-            @Component({
-              selector: 'my-comp',
-              template: `<h3>my-comp</h3>`,
-              host: {
-                '[@animation1]': 'colorExp',
-                '[@animation3]': 'bgExp',
-              },
-              animations: [
-                trigger(
-                    'animation1',
-                    [
-                      state('color', style({color: 'blue'})),
-                    ]),
-                trigger(
-                    'animation2',
-                    [
-                      state('opacity', style({opacity: '0.5'})),
-                    ]),
-                trigger(
-                    'animation3',
-                    [
-                      state('bg', style({backgroundColor: 'green'})),
-                    ]),
-              ],
-            })
-            class MyComponent extends BareClass {
-              bgExp = 'bg';
-            }
-
-            @Component({template: `<my-comp>app</my-comp>`})
-            class App {
-            }
-
-            TestBed.configureTestingModule({
-              declarations: [App, MyComponent, SuperDirective],
-              imports: [NoopAnimationsModule],
-            });
-            const fixture = TestBed.createComponent(App);
-            fixture.detectChanges();
-            const queryResult = fixture.debugElement.query(By.css('my-comp'));
-
-            expect(queryResult.nativeElement.style.color).toBe('blue');
-            expect(queryResult.nativeElement.style.opacity).toBe('0.5');
-            expect(queryResult.nativeElement.style.backgroundColor).toBe('green');
-          });
-    });
-
     describe('host bindings (style related)', () => {
       // TODO: sub and super HostBinding same property but different bindings
       // TODO: sub and super HostBinding same binding on two different properties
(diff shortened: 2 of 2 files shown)


Zuckjet:
this PR failed on CircleCI with message:

>   ✗ should be active with a blue background
        - Failed: Wait timed out after 2011ms

It seems I just need rerun the "test_docs_examples_ivy" workflow, but I don't  have the permissions.

packages/core/src/render3/features/inherit_definition_feature.ts
```diff
@@ -70,7 +70,10 @@ export function ɵɵInheritDefinitionFeature(definition: DirectiveDef<any>| Comp
         fillProperties(definition.inputs, superDef.inputs);
         fillProperties(definition.declaredInputs, superDef.declaredInputs);
         fillProperties(definition.outputs, superDef.outputs);
-
+        if ('data' in superDef) {
```
mhevery: 1. This change needs tests.
2. I don't understand why we need to check that `data` is present in `superDef`. Can this just be done always?

mhevery (changes requested):

packages/core/src/render3/features/inherit_definition_feature.ts
Zuckjet: Thanks for review. I have update my code. But when I add test file, I cannot pass `ci/circleci` workflow. it show error:

> The selector "root-ng-internal-isolated-122" didn't match any elements

I still cannot figure out where the problem is, would you mind giving me some advice to fix it ?

crisbeto (commented):
We also need some tests covering multiple levels of inheritance.

Zuckjet:
> We also need some tests covering multiple levels of inheritance.

you mean that I nedd add more test, such as "child comp extends super super component"?

crisbeto:
Yes. We've had issues in the past where one level of inheritance works fine, but things break down with multiple.

Zuckjet:
@crisbeto  OK, I get your point. If you don't mind, I will add  more test later. But now I need to fix the  
ci/circleci error, I still have no idea about it.

crisbeto:
This is the error in your test, not the one you posted above:

```
Found the synthetic property @testAnimation. Please include either "BrowserAnimationsModule" or "NoopAnimationsModule" in your application.
```

AndrewKushnir:
@Zuckjet, I looked at the CI errors and it looks like there is a couple things you can try:

- switch from `BrowserAnimationsModule` to `NoopAnimationsModule` - that should help avoid  the error that you mentioned (`The selector "root-ng-internal-isolated-122" didn't match any elements`)
- the issue that @crisbeto mentioned (`Found the synthetic property @testAnimation. Please include either "BrowserAnimationsModule" or "NoopAnimationsModule" in your application.
`) is likely to be a View Engine behavior (or a bug) that doesn't handle animation metadata inheritance correctly. I've created a [Stackblitz demo](https://stackblitz.com/edit/angular-osbmex?file=src%2Fapp%2Fhello.component.ts) that reproduces the same error. I believe this issue would be fixed by the changes in this PR for Ivy, so that particular test case can be wrapped into `onlyInIvy(...)` block, [for example as shown here](https://github.com/angular/angular/blob/master/packages/core/test/acceptance/ng_module_spec.ts#L81).

Also, I think it'd be great to add more tests as @crisbeto proposed that would include multiple levels of inheritance and components with host bindings (which was mentioned in [Stackblitz demo](https://stackblitz.com/edit/ivy-ejv3ha?file=src%2Fapp%2Fhello.component.ts) attached to the original issue).

Please let us know if any additional information is required.

Thank you.

Zuckjet:
@AndrewKushnir  Thanks for offering so many detail info, it helped me a lot.
I have added more tests. If there is something I should change, let me know.

packages/core/src/render3/features/inherit_definition_feature.ts
Zuckjet: Fixed

AndrewKushnir:
Hi @Zuckjet, thanks for adding extra tests.

I looked at the changes and I have a few comments:

- we should add a test with **host bindings** as it was described in original issue and demonstrated in [Stackblitz demo](https://stackblitz.com/edit/ivy-ejv3ha?file=src%2Fapp%2Fhello.component.ts). This test should work correctly with View Engine and Ivy. It'd be also great to have 3 separate tests for this: 1) with simple 1 level inheritance (Parent-Child); 2) similar test, but with multiple levels of inheritance and 3) test where parent and child components have animations with the same name

- it looks like your current fix just overrides `data` field value altogether. In case Parent and Child components have different set of animations, I'm wondering if we should merge them rather than overriding. For example, if `Parent` has animations `A` and `B` and `Child` has `B` and `C`, should the final set be `B, C` or `A, B, C` (with just one `B` present). @crisbeto could you please let us know what's the expected behavior?

Thank you.

packages/core/test/acceptance/inherit_definition_feature_spec.ts
```diff
@@ -4819,6 +4864,52 @@ describe('inheritance', () => {
...
+                    'animation1',
+                    [
+                      state('useAnimation1', style({color: 'red', opacity: 0.5})),
+                    ]),
+              ],
+            })
+            class SuperSuperComponent {
+            }
+
+            class SuperComponent extends SuperSuperComponent {}
```
AndrewKushnir: Could you please add `@Directive()` as a decorator to the `SuperComponent` class?

packages/core/test/acceptance/inherit_definition_feature_spec.ts
```diff
@@ -4819,6 +4864,52 @@ describe('inheritance', () => {
       });
     });
 
+    describe('developer-defined data', () => {
+      onlyInIvy('View Engine does not provide this check')
+          .it('should inherit developer-defined data', () => {
```
AndrewKushnir: Test description is the same as the test added above. Could you please update test descriptions to have more information about the test itself, for example:

```
.it('should inherit developer-defined data in case of immediate parent-child case', () => {
```

```
.it('should inherit developer-defined data across multiple inheritance levels', () => {
```

packages/core/test/acceptance/inherit_definition_feature_spec.ts
```diff
@@ -4819,6 +4864,52 @@ describe('inheritance', () => {
...
+
+            @Component({
+              selector: 'my-comp',
+              template: `<h3 [@animation1]="'useAnimation1'">my-comp</h3>`,
+            })
+            class MyComponent extends SuperComponent {
+            }
+
+            @Component({
+              template: `
```
AndrewKushnir: Please update `template` formatting and may be make it consistent with other components in this test (template in one line).

AndrewKushnir (changes requested):
Thanks for adding extra tests, please see my comments below and also additional feedback in this comment: https://github.com/angular/angular/pull/35105#issuecomment-583509068. Thank you.

crisbeto:
@AndrewKushnir my understanding was that we could merge the animations so the end array looks like `[B, C, A, B]` and we could avoid de-duplcating the triggers.

Zuckjet:
@AndrewKushnir Sorry for the late reply, since we have a different timezone, I often can't reply you in time.

I have read your review advice, I will take your advice and do some changes as soon as possible.
But before that I want to have a discussion with you about this PR.

this PR is aim to fix the bug that the original issue points out. I dive into angular souce code, and find out the reason. So I just add one line code:
https://github.com/angular/angular/blob/61035c9a84950812c0dc0739400d88f8f8fe7464/packages/core/src/render3/features/inherit_definition_feature.ts#L73-L74
We can see the real reason is that the compDef doesn't inherit superCompDef's data property.

> we should add a test with host bindings as it was described in original issue

I don't understand why we should add host bindings in the test. the error original issue points out still happen even if not use host bindings:
```
template: `<div>component</div>`,
  host: {
    '[@state]': 'expression',
  },
```
will equal to:
```
<div [@state]='expression'>component</div>
```
in my opinion, host binding relative test should not be include this test.

packages/core/test/acceptance/inherit_definition_feature_spec.ts
Zuckjet: you mean: 
```
@Directive('xx')
class SuperComponent {
}
```
? but `Directives `cannot inherit `Components`, SuperComponent extends SuperSuperComponnet

Zuckjet: https://github.com/angular/angular/blob/ef905f728f2200b7926dfdb617b8de4cd53c637e/packages/core/test/acceptance/inherit_definition_feature_spec.ts#L3523-L3533

May be fixed ?

packages/core/test/acceptance/inherit_definition_feature_spec.ts
Zuckjet: there already have some top category, such as 
https://github.com/angular/angular/blob/ef905f728f2200b7926dfdb617b8de4cd53c637e/packages/core/test/acceptance/inherit_definition_feature_spec.ts#L3847
https://github.com/angular/angular/blob/ef905f728f2200b7926dfdb617b8de4cd53c637e/packages/core/test/acceptance/inherit_definition_feature_spec.ts#L4636

so can I just describe it such as "animations should inherit animations" ? because they all under sub-category, and this desc can keep consistent with other similar test case such as inputs outputs. of course, If you still think we should change the description as you said, I will change it as soon as possible.

packages/core/test/acceptance/inherit_definition_feature_spec.ts
Zuckjet: Fixed

Zuckjet:
> @AndrewKushnir my understanding was that we could merge the animations so the end array looks like `[B, C, A, B]` and we could avoid de-duplcating the triggers.

thanks for advice. Since angular compilers don't de-duplcate the triggers, and two same name triggers the first will work.  I merge the animations through below functions:
https://github.com/angular/angular/blob/ef905f728f2200b7926dfdb617b8de4cd53c637e/packages/core/src/render3/features/inherit_definition_feature.ts#L188-L191

If this implement isn't what you want, please let me know.

Zuckjet:
> Hi @Zuckjet, thanks for adding extra tests.
> 
> I looked at the changes and I have a few comments:
> 
> * we should add a test with **host bindings** as it was described in original issue and demonstrated in [Stackblitz demo](https://stackblitz.com/edit/ivy-ejv3ha?file=src%2Fapp%2Fhello.component.ts). This test should work correctly with View Engine and Ivy. It'd be also great to have 3 separate tests for this: 1) with simple 1 level inheritance (Parent-Child); 2) similar test, but with multiple levels of inheritance and 3) test where parent and child components have animations with the same name
> * it looks like your current fix just overrides `data` field value altogether. In case Parent and Child components have different set of animations, I'm wondering if we should merge them rather than overriding. For example, if `Parent` has animations `A` and `B` and `Child` has `B` and `C`, should the final set be `B, C` or `A, B, C` (with just one `B` present). @crisbeto could you please let us know what's the expected behavior?
> 
> Thank you.

I have read your review suggestion and think it helpful. I also commited some changes and replyed you in corresponding comment. 

what's more, I have a `ci/circleci: saucelabs_ivy` CI error:
```
A platform with a different configuration has been created. Please destroy it first.
```
From the error info, I have tried to use `TestBed.resetTestingModule();` or `destroyPlatform()`, but still cannot work. Could you please help me fix this ? Many thanks!

AndrewKushnir (changes requested):
Thanks for additional changes @Zuckjet, I think we are very close! 👍 

I left a few additional comments, please see below.

Regarding the CI failure: I think you should remove the `TestBed.resetTestingModule();` call - it's unnecessary, since it's called by TestBed n between tests runs. Could you please try to remove it and push the changes, so that the tests are executed again on CI?

packages/core/test/acceptance/inherit_definition_feature_spec.ts
```diff
@@ -4132,6 +4247,125 @@ describe('inheritance', () => {
       });
     });
 
+    describe('animations', () => {
```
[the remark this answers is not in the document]
Zuckjet: fix

packages/core/test/acceptance/inherit_definition_feature_spec.ts
```diff
@@ -3516,6 +3518,119 @@ describe('inheritance', () => {
...
+
+        @Component({
+          template: `
+              <my-comp>app</my-comp>
+            `
+        })
+        class App {
+        }
+
+        TestBed.resetTestingModule();
```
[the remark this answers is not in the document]
Zuckjet: fix

packages/core/test/acceptance/inherit_definition_feature_spec.ts
```diff
@@ -3516,6 +3518,119 @@ describe('inheritance', () => {
...
+              [
+                state('color', style({color: 'red'})),
+              ])],
+        })
+        class MyComponent extends BareClass {
+        }
+
+        @Component({
+          template: `
+              <my-comp>app</my-comp>
```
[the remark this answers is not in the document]
Zuckjet: fix

packages/core/test/acceptance/inherit_definition_feature_spec.ts
```diff
@@ -3516,6 +3518,119 @@ describe('inheritance', () => {
       });
     });
 
+    describe('animations', () => {
+      onlyInIvy('View Engine does not provide this check').it('should inherit animations', () => {
```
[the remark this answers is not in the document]
Zuckjet: I have tried remove  `onlyInIvy`, after that it cannot pass test.

packages/core/test/acceptance/inherit_definition_feature_spec.ts
```diff
@@ -3516,6 +3518,119 @@ describe('inheritance', () => {
...
+          imports: [NoopAnimationsModule],
+        });
+        const fixture = TestBed.createComponent(App);
+        fixture.detectChanges();
+        const queryResult = fixture.debugElement.query(By.css('my-comp'));
+
+        expect(queryResult.nativeElement.style.color).toBe('red');
+      });
+
+      onlyInIvy('View Engine does not provide this check').it('should compose animations', () => {
```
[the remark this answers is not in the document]
Zuckjet: https://github.com/angular/angular/pull/35105#discussion_r377814880

packages/core/src/render3/features/inherit_definition_feature.ts
```diff
@@ -180,3 +184,8 @@ function inheritHostBindings(
     definition.hostBindings = superHostBindings;
   }
 }
+
+function inheritData(defData: {[kind: string]: any}, superData: {[kind: string]: any}) {
+  (superData.animation) &&
```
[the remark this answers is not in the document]
Zuckjet: fix

AndrewKushnir:
Hi @Zuckjet, thanks for additional changes. I will test View Engine behavior and get back to you with additional information (to see which tests we can run with both Ivy and View Engine). Thank you.

packages/core/test/acceptance/inherit_definition_feature_spec.ts
AndrewKushnir: Thanks for the comment, it makes sense to keep them as is then (given that we have section description above in `describe` call).

AndrewKushnir:
@Zuckjet, quick update:

- I verified that `host` metadata was not inherited in View Engine, that's why the tests failed with View Engine as you mentioned. I tried updating it to `@HostBinding` decorator, but I got a different error message (related to unexpected synthetic property)

- also I looked at the `saucelabs_ivy` CircleCI job that is failing and the problem only happens in IE10. Example failure:

```
IE 10.0.0 (Windows 8.0.0) inheritance of a directive inherited by a bare class and then by a component animations should work with composed host bindings FAILED
	Error: Expected '' to be '0.5'.
	    at <Jasmine>
```

It looks like something is not fully correct while running newly added tests in IE10. I checked basic animations case by [creating a test PR](https://github.com/angular/angular/pull/35356) and the same `saucelabs_ivy` CircleCI job [was successful](https://circleci.com/gh/angular/angular/621920).

There are also errors like this one in the `saucelabs_ivy` CircleCI job in this PR:

```
A platform with a different configuration has been created. Please destroy it first.
```

But they seem unrelated to your changes and most likely pre-existing.

@Zuckjet, we may need to refactor newly added tests in this PR a bit. Would it be ok if I push a couple fixup commits to this PR? Please let me know.

Thank you.

Zuckjet:
> @Zuckjet, quick update:
> 
> * I verified that `host` metadata was not inherited in View Engine, that's why the tests failed with View Engine as you mentioned. I tried updating it to `@HostBinding` decorator, but I got a different error message (related to unexpected synthetic property)
> * also I looked at the `saucelabs_ivy` CircleCI job that is failing and the problem only happens in IE10. Example failure:
> 
> ```
> IE 10.0.0 (Windows 8.0.0) inheritance of a directive inherited by a bare class and then by a component animations should work with composed host bindings FAILED
> 	Error: Expected '' to be '0.5'.
> 	    at <Jasmine>
> ```
> 
> It looks like something is not fully correct while running newly added tests in IE10. I checked basic animations case by [creating a test PR](https://github.com/angular/angular/pull/35356) and the same `saucelabs_ivy` CircleCI job [was successful](https://circleci.com/gh/angular/angular/621920).
> 
> There are also errors like this one in the `saucelabs_ivy` CircleCI job in this PR:
> 
> ```
> A platform with a different configuration has been created. Please destroy it first.
> ```
> 
> But they seem unrelated to your changes and most likely pre-existing.
> 
> @Zuckjet, we may need to refactor newly added tests in this PR a bit. Would it be ok if I push a couple fixup commits to this PR? Please let me know.
> 
> Thank you.

Of course you can , thanks for your helping.

AndrewKushnir:
Hi @Zuckjet, quick update: I've refactored tests to make descriptions more explicit, updated scenarios in a couple cases and also removed tests that were testing same (or very similar) use-cases. The tests should work correctly on CI now. Next step is to have reviewers to look at this PR once again (I will ping them and ask for review). Will keep you updated. Thank you.

Zuckjet:
@AndrewKushnir Thanks. finally, this PR seems will be close. I should say you have done most of work for this PR, I am sorry for bothering you so many times.

packages/core/src/render3/features/inherit_definition_feature.ts
```diff
@@ -180,3 +184,8 @@ function inheritHostBindings(
     definition.hostBindings = superHostBindings;
   }
 }
+
+function inheritData(defData: {[kind: string]: any}, superData: {[kind: string]: any}) {
+  superData.animation &&
```
crisbeto: Since this is a one-liner and it's only used in one place, couldn't we just do it inline? I feel like that would make it more readable as well as generate less code. E.g.

```ts
const superData = (superDef as ComponentDef<any>).data;

if (superData) {
  (defData.animation = (defData.animation || []).concat(superData.animation));
}
```

crisbeto (approved):
LGTM, aside from the last comment.

AndrewKushnir:
Hey @crisbeto, thanks for review! I've inlined the code (with a slight change to check if superDef is a Component) as you proposed and added extra comments. Could you plz take a final look? Thank you.

AndrewKushnir:
Hi @Zuckjet, no problem, thanks for your help! 👍 I've applied the changes as Kristiyan proposed and [started a presubmit](http://test/OCL:295210706:BASE:295210745:1581714883367:4473fcdd) to verify the changes in google3. If everything goes well with testing, this PR should be ready for merge. Thank you.

crisbeto (approved):
LGTM

mhevery (approved):

AndrewKushnir (approved):

PR #35106, opened 2020-02-02
fix(ngcc): correctly invalidate cache when moving/removing files/directories

### LINKED ISSUE #35095: Compilation breaks after update to 9.0.0-rc.13: `Duplicate identifier 'ɵngcc0'`

gkalpak:
One particular scenario where this was causing problems was when the [BackupFileCleaner][1] restored a file (such as a `.d.ts` file) by [moving the backup file][2] to its original location, but the modified content was kept in the cache.

Fixes #35095.

[1]: https://github.com/angular/angular/blob/4d36b2f6e/packages/compiler-cli/ngcc/src/writing/cleaning/cleaning_strategies.ts#L54
[2]: https://github.com/angular/angular/blob/4d36b2f6e/packages/compiler-cli/ngcc/src/writing/cleaning/cleaning_strategies.ts#L61


integration/ngcc/test.sh
```diff
@@ -140,6 +160,18 @@ assertSucceeded "Expected 'ngcc' to log 'Compiling'."
   assertSucceeded "Expected 'ngcc' to not add trailing commas to factory function parameters in UMD."
 
 
+# Can it correctly clean up and re-compile when dependencies are already compiled by a different version?
+# (Use sync mode to ensure all tasks share the same `CachedFileSystem` instance.)
+  assertNotEquals "1.3.3.7" `node --print "require('@angular/material/button/package.json').__processed_by_ivy_ngcc__.main"`
```
IgorMinar: Add a note about why you picked 1.3.3.7

integration/ngcc/test.sh
```diff
@@ -3,7 +3,7 @@
 # Do not immediately exit on error to allow the `assertSucceeded` function to handle the error.
 #
 # NOTE:
-# Each statement should be followed by an `assertSucceeded`/`assertFailed` or `exit 1` statement.
+# Each statement should be followed by an `assert*` or `exit 1` statement.
```
IgorMinar: Not for this PR but a general note: Can we please stop writing these scripts in bash?

Shelljs is the way to go. Please don't let more bash scripts into the repo during core reviews.

integration/ngcc/mock-ngcc-version.js
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * This script replaces the ngcc version that is used as marker to make already processed packages
+ * appear as if they were processed by a different version. This allows testing the clean-up logic
+ * without the overhead of actually installing an older version, compiling the dependencies, then
+ * installing the current version and compiling again.
+ */
+
+const {readFileSync, writeFileSync} = require('fs');
+
+const mockVersion = process.argv[2] || 'MOCK.VERSION';
```
IgorMinar: The default value seems dubious. Maybe just require the an argument is provided?

integration/ngcc/mock-ngcc-version.js
```diff
@@ -0,0 +1,47 @@
...
+const buildMarkerPath = require.resolve('@angular/compiler-cli/ngcc/src/packages/build_marker');
+const buildMarkerVersionRe = /\bNGCC_VERSION = '([^']+)';/;
+
+const oldContent = readFileSync(buildMarkerPath, 'utf8');
```
IgorMinar: older -> original? Please rename throughout the file.

Older could also refer to semver lower version which is not necessarily the case.

IgorMinar (changes requested):

integration/ngcc/test.sh
gkalpak: Amen to that :smiley:

packages/compiler-cli/src/ngtsc/file_system/test/cached_file_system_spec.ts
```diff
@@ -252,6 +252,24 @@ describe('CachedFileSystem', () => {
...
     });
+
+    it('should delete the `to` "readFile" cache (if `from` was not cached)', () => {
+      spyOn(delegate, 'moveFile');
+      const readFileSpy = spyOn(delegate, 'readFile');
+
+      // Fill the xyz "readFile" cache
+      readFileSpy.and.returnValue('xyz content');
+      fs.readFile(xyzPath);
+      readFileSpy.calls.reset();
```
petebacondarwin: I don't think this is necessary, since you are rewriting the spy below.

petebacondarwin (changes requested):
The source code changes look perfect! Thanks for tidying this up @gkalpak.

What is our policy for updating `integration/ngcc/yarn.lock`?

Could we change the integration test to modify the build markers in `@angular/material/button/package.json` and then run ngcc again, rather than hacking ngcc itself to have a different version?
Modifying ngcc is much more brittle to internal code changes in the ngcc source code - as we found wth the components repo unit tests.

integration/ngcc/mock-ngcc-version.js
gkalpak: I had it like this originally <sub>(pun intented)</sub>, but liked the `old*`/`new*` symmetry.
Changed it to `original*`/`updated*`.

IgorMinar: Sgtm

integration/ngcc/mock-ngcc-version-marker.js
```diff
@@ -0,0 +1,50 @@
...
+ * **Usage:**
+ * ```sh
+ * node mock-ngcc-version-marker <entry-point> <mock-ngcc-version>
+ *
+ * # Example:
+ * # node mock-ngcc-version-marker @angular/material/button 3.0.0
+ * ```
+ *
+ * Ngcc marks entry-points as processed by adding marker properties in their `package.json`. It uses
+ * a version as marker in order to be able to distinguish entry-points processed by a different ngcc
```
petebacondarwin:
```suggestion
 * a version as a marker in order to be able to distinguish entry-points processed by a different ngcc
```

integration/ngcc/test.sh
```diff
@@ -140,6 +160,27 @@ assertSucceeded "Expected 'ngcc' to log 'Compiling'."
   assertSucceeded "Expected 'ngcc' to not add trailing commas to factory function parameters in UMD."
 
 
+# Can it correctly clean up and re-compile when dependencies are already compiled by a different version?
+  readonly actualNgccVersion=`node --print "require('@angular/compiler-cli/package.json').version"`
+  readonly mockNgccVersion="3.0.0"
+
+  # Mock the ngcc version marker on a package make it appear as if compiled by a different ngcc version.
```
petebacondarwin:
```suggestion
  # Mock the ngcc version marker on a package to make it appear as if it is compiled by a different ngcc version.
```

petebacondarwin (approved):

IgorMinar (approved):

mhevery:
[presubmit](http://test/OCL:292996966:BASE:292997002:1580766832586:91b3c27a)

gkalpak:
BTW, regarding the `yarn.lock` file update:

These lockfiles depend on the locally built Angular packages (and their dependencies), because the integration projects have dependencies such as `"@angular/core": "file:../../dist/packages-dist/core"`.

However, we do not go and update these lockfiles every time the dependencies of the Angular packages change (such as after 37f39d6db). So, they get updated whenever someone happens to run the integration project locally, but that should be fine. (The same updates happen every time they run on CI.)

PR #35115, opened 2020-02-03
ci: remove components-repo-ci blocklist

devversion:
Previously we needed the `components-repo-ci` blocklist to disable
tests that were failing during the development of Ivy. Since we fixed
all those failing tests, and we don't want to regress, we can remove the
blocklist logic.

Resolves FW-1807


IgorMinar (approved):

IgorMinar:
Thanks Paul!

PR #35131, opened 2020-02-03
feat(ngcc): pause async processing if another process has the lockfile

petebacondarwin:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/compiler-cli/ngcc/src/execution/lock_file.ts
+++ b/packages/compiler-cli/ngcc/src/execution/lock_file.ts
@@ -7,77 +7,37 @@
  */
 import * as process from 'process';
 import {FileSystem} from '../../../src/ngtsc/file_system';
+import {Logger} from '../logging/logger';
 
-/**
- * The LockFile is used to prevent more than one instance of ngcc executing at the same time.
- *
- * When ngcc starts executing, it creates a file in the `compiler-cli/ngcc` folder. If it finds one
- * is already there then it fails with a suitable error message.
- * When ngcc completes executing, it removes the file so that future ngcc executions can start.
- */
-export class LockFile {
+export abstract class LockFileBase {
   lockFilePath =
       this.fs.resolve(require.resolve('@angular/compiler-cli/ngcc'), '../__ngcc_lock_file__');
 
-  constructor(private fs: FileSystem) {}
+  constructor(protected fs: FileSystem) {}
 
-  /**
-   * Run a function guarded by the lock file.
-   *
-   * Note that T can be a Promise. If so, we run the `remove()` call in the promise's `finally`
-   * handler. Otherwise we run the `remove()` call in the `try...finally` block.
-   *
-   * @param fn The function to run.
-   */
-  lock<T>(fn: () => T): T {
-    let isAsync = false;
-    this.create();
+  protected writeLockFile(): void {
     try {
-      const result = fn();
-      if (result instanceof Promise) {
-        isAsync = true;
-        // The cast is necessary because TS cannot deduce that T is now a promise here.
-        return result.finally(() => this.remove()) as unknown as T;
-      } else {
-        return result;
-      }
-    } finally {
-      if (!isAsync) {
-        this.remove();
-      }
+      this.addSignalHandlers();
+      // To avoid race conditions, we check for existence of the lockfile
+      // by actually trying to create it exclusively.
+      return this.fs.writeFile(this.lockFilePath, process.pid.toString(), /* exclusive */ true);
+    } catch (e) {
+      this.removeSignalHandlers();
+      throw e;
     }
   }
 
   /**
-   * Write a lock file to disk, or error if there is already one there.
+   * Read the pid from the lockfile.
+   *
+   * It is feasible that the lockfile was removed between the previous check for existence
+   * and this file-read. If so then we still error but as gracefully as possible.
    */
-  protected create() {
+  protected readLockFile(): string {
     try {
-      this.addSignalHandlers();
-      // To avoid race conditions, we check for existence of the lockfile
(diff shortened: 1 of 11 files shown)


packages/compiler-cli/ngcc/src/execution/lock_file.ts
```diff
@@ -119,3 +86,105 @@ export class LockFile {
...
+        `(If you are sure no ngcc process is running then you should delete the lockfile at ${this.lockFilePath}.)`);
+  }
+}
+
+/**
+ * LockFileAsync is used to prevent more than one instance of ngcc executing at the same time,
+ * when being called in an asynchronous context.
+ *
+ * * When ngcc starts executing, it creates a file in the `compiler-cli/ngcc` folder.
+ * * If it finds one is already there then it fails with a suitable error message.
```
gkalpak: No, it doesn't :wink:

packages/compiler-cli/ngcc/src/execution/lock_file.ts
```diff
@@ -119,3 +86,105 @@ export class LockFile {
...
+   *
+   * @param fn The function to run.
+   */
+  async lock<T>(fn: () => Promise<T>): Promise<T> {
+    await this.create();
+    return await fn().finally(() => this.remove());
+  }
+
+  protected async create() {
+    for (let attempts = 0; attempts < this.retryAttempts; attempts++) {
```
gkalpak: I think we should track `pid` and reset `attempts` if the `pid` changes. If there more than 2 ngcc processes waiting for each other, one of them will have to wait for all others to complete.

gkalpak (changes requested):
LGTM, except I think we should reset the waiting period when the `pid` in the lockfile changes.
Really excited to incorporate this into the CLI workflow \o/

packages/compiler-cli/ngcc/test/execution/lock_file_spec.ts
```diff
@@ -335,10 +335,40 @@ runInEachFileSystem(() => {
...
+        fs.writeFile(lockFile.lockFilePath, '188');
+        const promise = lockFile.lock(async() => lockFile.log.push('fn()'));
+        // The lock is now waiting on the lockfile becoming free, so no `fn()` in the log.
+        expect(lockFile.log).toEqual(['create()']);
+        expect(lockFile.getLogger().logs.info).toEqual([[
+          'Another process, with id 188, is currently running ngcc.\nWaiting up to 1s for it to finish.'
+        ]]);
+        // We need to write to the rawFs to ensure that we don't update the cache at this point
+        rawFs.writeFile(lockFile.lockFilePath, '444');
+        await new Promise(resolve => setTimeout(resolve, 250));
```
gkalpak: This looks like a potential source of flakes :grin:
Could we use [fakeAsync()][1] and [tick()][2] to make this test deterministic?

[1]: https://angular.io/api/core/testing/fakeAsync
[2]: https://angular.io/api/core/testing/tick

gkalpak (approved):

packages/compiler-cli/ngcc/test/execution/lock_file_spec.ts
petebacondarwin: Yeah, I was a little bit worried about this. We cannot use the Angular `fakeAsync` stuff, since that is  driven via Zone.js and we definitely not going to bring that in here just for testing...

I tried to make the timeouts as reliable as possible so that it would be extremely unlikely for it to flake:

* having proven that the class has already checked the lockfile (and read its value) we are immediately writing the new one synchronously so there is no chance for the `setTimeout` inside `LockFile` to be triggered.
* we then set a timeout for 250ms before the next check but the class has been configured to check the contents of the lockfile every 100ms. So there should be at least 2 opportunities for a `setTimeout` to be triggered before this testing one is triggered.
* The class is setup not to fail until 10 attempts have occurred, which is a full second, so there is very little chance that the 250ms timeout is not triggered before we run out of attempts.

Do you have any better ideas?

kara:
@petebacondarwin Are you still waiting on a review from @alxhub or is this ready to presubmit/merge?

petebacondarwin:
Just a presubmit please @kara

packages/compiler-cli/ngcc/test/execution/lock_file_spec.ts
gkalpak: Fair enough. I am fine leaving this as is and address it in the future if we start seeing flakes.

For future reference, one way to make this deterministic would be using Jasmine's [Clock][1]. There is one twist, because we need to also flush microtasks after each `retryDelay` ms, so it would be something like:

```ts
const clock = jasmine.clock();
const moveToNextIteration = () => {
  clock().tick(100);     // Or whatever `retryDelay` we have specified.
  return Promise.resolve();  // To wait for microtask queue to be flushed.
};

clock.withMock(async () => {
  const rawFs = getFileSystem();
  const fs = new CachedFileSystem(rawFs);
  const lockFile = new LockFileUnderTest(fs);

  fs.writeFile(lockFile.lockFilePath, '188');
  const promise = lockFile.lock(async () => lockFile.log.push('fn()'));
  expect(lockFile.log).toEqual(['create()']);
  expect(lockFile.getLogger().logs.info).toEqual([...]);

  rawFs.writeFile(lockFile.lockFilePath, '444');
  await moveToNextIteration();
  await moveToNextIteration();
  expect(lockFile.getLogger().logs.info).toEqual([...]]);

  ...
});
```

[1]: https://jasmine.github.io/api/3.5/Clock

PR #35132, opened 2020-02-03
ngcc: source map flattening

petebacondarwin:
The library used by ngcc to update the source files (MagicString) is able
to generate a source-map but it is not able to account for any previous
source-map that the input text is already associated with.

There have been various attempts to fix this but none have been very
successful, since it is not a trivial problem to solve.

This PR contains a novel approach that is able to load up a tree of
source-files connected by source-maps and flatten them down into a single
source-map that maps directly from the final generated file to the original
sources referenced by the intermediate source-maps.

(Note that there are probably (definitely) some terrible performance characteristics in this PR. But I wanted to go for correctness before stressing about speed.)

FW-1558

### DIFF
--- a/packages/compiler-cli/ngcc/BUILD.bazel
+++ b/packages/compiler-cli/ngcc/BUILD.bazel
@@ -35,6 +35,7 @@ ts_library(
         "@npm//magic-string",
         "@npm//semver",
         "@npm//source-map",
+        "@npm//sourcemap-codec",
         "@npm//typescript",
     ],
 )
--- a/packages/compiler-cli/ngcc/src/sourcemaps/segment_marker.ts
+++ b/packages/compiler-cli/ngcc/src/sourcemaps/segment_marker.ts
@@ -28,9 +28,10 @@ export function compareSegments(a: SegmentMarker, b: SegmentMarker): number {
   return a.line === b.line ? a.column - b.column : a.line - b.line;
 }
 
-// The `1` is to indicate a newline character between the lines
-// Note that in the actual contents there could be more than one character -
-// e.g. \r\n - that indicate a newline but that is not important here.
+// The `1` is to indicate a newline character between the lines.
+// Note that in the actual contents there could be more than one character that indicates a newline
+// - e.g. \r\n - but that is not important here since segment-markers are in line/column pairs and
+// so differences in length due to extra `\r` characters do not affect the algorithms.
 const NEWLINE_MARKER_OFFSET = 1;
 
 /**
@@ -59,11 +60,12 @@ export function segmentDiff(lineLengths: number[], a: SegmentMarker, b: SegmentM
 }
 
 /**
- * Return a new SegmentMarker that is offset by the given number of characters.
+ * Return a new segment-marker that is offset by the given number of characters.
  *
- * @param lineLengths the lines of the source file whose segment marker we are offsetting
- * @param marker The segment to offset
- * @param offset The number of character to offset by
+ * @param lineLengths The length of each line in the source file whose segment-marker we are
+ * offsetting.
+ * @param marker The segment to offset.
+ * @param offset The number of character to offset by.
  */
 export function offsetSegment(lineLengths: number[], marker: SegmentMarker, offset: number) {
   if (offset === 0) {
--- a/packages/compiler-cli/ngcc/src/sourcemaps/segment_marker.ts
+++ b/packages/compiler-cli/ngcc/src/sourcemaps/segment_marker.ts
@@ -1,0 +1,86 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+
+/**
+* A marker that indicates the start of a segment in a mapping.
+*
+* The end of a segment is indicated by the the first segment-marker of another mapping whose start
+* is greater or equal to this one.
+*/
+export interface SegmentMarker {
+  readonly line: number;
+  readonly column: number;
+}
+
+/**
+ * Compare two segment-markers, for use in a search or sorting algorithm.
+ *
+ * @returns a positive number if `a` is after `b`, a negative number if `b` is after `a`
+ * and zero if they are at the same position.
+ */
+export function compareSegments(a: SegmentMarker, b: SegmentMarker): number {
+  return a.line === b.line ? a.column - b.column : a.line - b.line;
+}
+
+// The `1` is to indicate a newline character between the lines
+// Note that in the actual contents there could be more than one character -
+// e.g. \r\n - that indicate a newline but that is not important here.
+const NEWLINE_MARKER_OFFSET = 1;
+
+/**
+ * Compute the difference between two segment markers in a source file.
+ *
+ * @param lineLengths the lengths of each line of content of the source file where we are computing
+ * the difference
+ * @param a the start marker
+ * @param b the end marker
+ * @returns the number of characters between the two segments `a` and `b`
+ */
+export function segmentDiff(lineLengths: number[], a: SegmentMarker, b: SegmentMarker) {
+  let diff = b.column - a.column;
+
+  // Deal with `a` being before `b`
+  for (let lineIndex = a.line; lineIndex < b.line; lineIndex++) {
+    diff += lineLengths[lineIndex] + NEWLINE_MARKER_OFFSET;
+  }
+
+  // Deal with `a` being after `b`
+  for (let lineIndex = a.line - 1; lineIndex >= b.line; lineIndex--) {
+    // The `+ 1` is the newline character between the lines
+    diff -= lineLengths[lineIndex] + NEWLINE_MARKER_OFFSET;
+  }
+  return diff;
+}
+
+/**
+ * Return a new SegmentMarker that is offset by the given number of characters.
+ *
+ * @param lineLengths the lines of the source file whose segment marker we are offsetting
+ * @param marker The segment to offset
+ * @param offset The number of character to offset by
+ */
+export function offsetSegment(lineLengths: number[], marker: SegmentMarker, offset: number) {
+  if (offset === 0) {
+    return marker;
+  }
+
+  let line = marker.line;
+  let column = marker.column + offset;
+
+  while (line < lineLengths.length - 1 && column > lineLengths[line]) {
+    column -= lineLengths[line] + NEWLINE_MARKER_OFFSET;
+    line++;
+  }
+  while (line > 0 && column < 0) {
+    line--;
+    column += lineLengths[line] + NEWLINE_MARKER_OFFSET;
+  }
+
+  return {line, column};
+}
--- a/packages/compiler-cli/ngcc/src/sourcemaps/source_file.ts
+++ b/packages/compiler-cli/ngcc/src/sourcemaps/source_file.ts
@@ -1,0 +1,313 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import {removeComments, removeMapFileComments} from 'convert-source-map';
+import {SourceMapMappings, SourceMapSegment, decode, encode} from 'sourcemap-codec';
+import {AbsoluteFsPath, dirname, relative} from '../../../src/ngtsc/file_system';
+import {RawSourceMap} from './raw_source_map';
+import {SegmentMarker, compareSegments, offsetSegment, segmentDiff} from './segment_marker';
+
+export function removeSourceMapComments(contents: string): string {
+  return removeMapFileComments(removeComments(contents)).replace(/\n\n$/, '\n');
+}
+
+export class SourceFile {
+  /**
+   * The parsed mappings that have been flattened so that any intermediate source mappings have been
+   * flattened.
+   *
+   * The result is that any source file mentioned in the flattened mappings have no source map (are
+   * pure original source files).
+   */
+  readonly flattenedMappings: Mapping[];
+  readonly lineLengths: number[];
+
+  constructor(
+      /** The path to this source file. */
+      readonly sourcePath: AbsoluteFsPath,
+      /** The contents of this source file. */
+      readonly contents: string,
+      /** The raw source map (if any) associated with this source file. */
+      readonly rawMap: RawSourceMap|null,
+      /** Whether this source files' source map was inline or external. */
+      readonly inline: boolean,
+      /** Any source files referenced by the raw source map associated with this source file. */
+      readonly sources: (SourceFile|null)[]) {
+    this.contents = removeSourceMapComments(contents);
+    this.lineLengths = computeLineLengths(this.contents);
+    this.flattenedMappings = this.flattenMappings();
+  }
+
+  /**
+   * Render the raw source map generated from the flattened mappings.
+   */
+  renderFlattenedSourceMap(): RawSourceMap {
+    const sources: SourceFile[] = [];
+    const names: string[] = [];
+
+    // Ensure a mapping line array for each line in the generated source.
+    const mappings: SourceMapMappings = this.lineLengths.map(() => []);
+
+    for (const mapping of this.flattenedMappings) {
+      const mappingLine = mappings[mapping.generatedSegment.line];
+      const sourceIndex = findIndexOrAdd(sources, mapping.originalSource);
+      const mappingArray: SourceMapSegment = [
+        mapping.generatedSegment.column,
+        sourceIndex,
+        mapping.originalSegment.line,
+        mapping.originalSegment.column,
+      ];
+      if (mapping.name !== undefined) {
+        const nameIndex = findIndexOrAdd(names, mapping.name);
+        mappingArray.push(nameIndex);
+      }
+      mappingLine.push(mappingArray);
+    }
+
+    const sourcePathDir = dirname(this.sourcePath);
+    const sourceMap: RawSourceMap = {
+      version: 3,
+      file: relative(sourcePathDir, this.sourcePath),
+      sources: sources.map(sf => relative(sourcePathDir, sf.sourcePath)), names,
+      mappings: encode(mappings),
+      sourcesContent: sources.map(sf => sf.contents),
+    };
+    return sourceMap;
+  }
+
+  /**
+   * Flatten the parsed mappings for this source file, so that all the mappings are to pure original
+   * source files with no transitive source maps.
+   */
+  private flattenMappings(): Mapping[] {
+    const mappings = parseMappings(this.rawMap, this.sources);
+    const originalSegments = extractOriginalSegments(mappings);
+    const flattenedMappings: Mapping[] = [];
+    for (let mappingIndex = 0; mappingIndex < mappings.length; mappingIndex++) {
+      const aToBmapping = mappings[mappingIndex];
+      const bSource = aToBmapping.originalSource;
+      if (bSource.flattenedMappings.length === 0) {
+        // The b source file has no mappings of its own (i.e. it a pure original file)
+        // so just use the mapping as-is.
+        flattenedMappings.push(aToBmapping);
+        continue;
+      }
+
+      // The `incomingStart` and `incomingEnd` are the `SegmentMarker`s in `B` that represent the
+      // section of `B` source file that is being mapped to by the current `aToBmapping`.
+      //
+      // For example, consider the mappings from A to B:
+      //
+      // src A   src B     mapping
+      //
+      //   a ----- a       [0, 0]
+      //   b       b
+      //   f -  /- c       [4, 2]
+      //   g  \ /  d
+      //   c -/\   e
+      //   d    \- f       [2, 5]
+      //   e
+      //
+      // For mapping [0,0] the incoming start and end are 0 and 2 (i.e. the range a, b, c)
+      // For mapping [4,2] the incoming start and end are 2 and 5 (i.e. the range c, d, e, f)
+      //
+      const incomingStart = aToBmapping.originalSegment;
+      const incomingEndIndex = originalSegments.indexOf(incomingStart) + 1;
+      const incomingEnd = incomingEndIndex < originalSegments.length ?
+          originalSegments[incomingEndIndex] :
+          undefined;
+
+      // The `outgoingStartIndex` and `outgoingEndIndex` are the indices of the range of mappings
+      // that leave `b` that we are interested in merging with the aToBmapping.
+      // We actually care about all the markers from the last bToCmapping directly before the
+      // `incomingStart` to the last bToCmaping directly before the `incomingEnd`, inclusive.
+      //
+      // For example, if we consider the range 2 to 5 from above (i.e. c, d, e, f) with the
+      // following mappings from B to C:
+      //
+      //   src B   src C     mapping
+      //     a
+      //     b ----- b       [1, 0]
+      //   - c       c
+      //  |  d       d
+      //  |  e ----- 1       [4, 3]
+      //   - f  \    2
+      //         \   3
+      //          \- e       [4, 6]
+      //
+      // The range with `incomingStart` at 2 and `incomingEnd` at 5 has outgoing start mapping of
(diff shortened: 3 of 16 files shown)


filipesilva (commented):
I know you mentioned you wanted to go for correctness before speed, but can you check what the perf difference is? We shouldn't merge this yet if it makes everyone's first build 3x as long.

petebacondarwin:
@filipesilva absolutely!

JoostK:
Additionally, could this work be avoided if source map generation is disabled?

JoostK:
@petebacondarwin Sorry, I was a little vague. What I meant to say is that when an app is built without source maps enabled (the TS `sourcemap` option is `false` or `sourceMap` is disabled in the CLI configuration), the user is likely not interested in source maps coming from ngcc as well.  Therefore, the additional work of merging source maps in ngcc may be for nothing. I see your point when there's already an existing sourcemap, in which case not merging source maps would result in an incorrect result (probably just as broken as what is generated by ngcc as of today)

petebacondarwin:
Also we have to be careful in monorepo scenarios where the ngcc generated files might be used in different compilations that are configured differently. So I think assuming that there is no major performance issues it is safer to always process the source file...

petebacondarwin:
I have now tweaked this algorithm a bit more and I believe it is working as desired.

Regarding performance, I ran ngcc against the entire AIO `node_modules` folder (all package and all formats) and compared the time taken on master and this PR. The average time on master was 140 seconds. The average time for this PR was 149 seconds. So while there is an increase in processing time it is around only 6%, which I argue is acceptable. It is possible that we can do some further performance tuning to improve things further later on.

filipesilva:
Yeah 6% is pretty much within just normal variance even.

filipesilva (approved):
LGTM as far as the CLI and perf are concerned, also LGTM from a dev-infra review perspective.

packages/compiler-cli/ngcc/src/sourcemaps/source_file.ts
```diff
@@ -0,0 +1,302 @@
...
+      //   g  \ /  d
+      //   c -/\   e
+      //   d    \- f       [2, 5]
+      //   e
+      //
+      // For mapping [0,0] the incoming start and end are 0 and 2 (i.e. the range a, b, c)
+      // For mapping [4,2] the incoming start and end are 2 and 5 (i.e. the range c, d, e, f)
+      //
+      const incomingStart = aToBmapping.originalSegment;
+      const incomingEnd = originalSegments[originalSegments.indexOf(incomingStart) + 1];
```
JoostK: nit: since this can be an out of bound read, `incomingEnd` can be `undefined` here (which you have dealt with in all places, so that's all good). AFAIK TypeScript will infer the type of `incomingEnd` to be `SegmentMarker`, so including `undefined` explicitly in the type makes this a bit more safe:

```suggestion
      const incomingEnd: SegmentMarker|undefined = originalSegments[originalSegments.indexOf(incomingStart) + 1];
```

packages/compiler-cli/ngcc/src/sourcemaps/source_file.ts
```diff
@@ -0,0 +1,302 @@
...
+      const aToBmapping = mappings[mappingIndex];
+      const bSource = aToBmapping.originalSource;
+      if (bSource.flattenedMappings.length === 0) {
+        // The b source file has no mappings of its own (i.e. it a pure original file)
+        // so just use the mapping as-is.
+        flattenedMappings.push(aToBmapping);
+        continue;
+      }
+
+      // The `incomingStart` and `incomingEnd` are the `SegmentMarker`s in `B` that represent the
```
JoostK: Thanks for these comments, this is super helpful!

packages/compiler-cli/ngcc/src/sourcemaps/source_file.ts
```diff
@@ -0,0 +1,302 @@
...
+      const outgoingEndIndex = incomingEnd !== undefined ?
+          findLastIndex(
+              bSource.flattenedMappings,
+              mapping => compareSegments(mapping.generatedSegment, incomingEnd) < 0) :
+          bSource.flattenedMappings.length - 1;
+
+      for (let bToCmappingIndex = outgoingStartIndex; bToCmappingIndex <= outgoingEndIndex;
+           bToCmappingIndex++) {
+        const bToCmapping = bSource.flattenedMappings[bToCmappingIndex];
+        if (bToCmapping) {
```
JoostK: This would only be `undefined` if `bToCmappingIndex` is out of bounds wrt `bSource.flattenedMappings`, but it looks like the loop iterates only known indices of `bSource.flattenedMappings` (as `outgoingStartIndex` and `outgoingEndIndex` are directly derived from that array).

packages/compiler-cli/ngcc/src/sourcemaps/source_file.ts
```diff
@@ -0,0 +1,302 @@
...
+      return index;
+    }
+  }
+  return -1;
+}
+
+/**
+ * A Mapping consists of two segment markers one in the generated source and one in the original
+ * source, which indicate the start of each segment. The end of a segment is indicated by the
+ * the first segment marker of another mapping whose start is greater or equal to this one.
```
JoostK: double "the":

```suggestion
 * first segment marker of another mapping whose start is greater or equal to this one.
```

packages/compiler-cli/ngcc/src/sourcemaps/source_file.ts
```diff
@@ -0,0 +1,302 @@
...
+  for (let index = items.length - 1; index >= 0; index--) {
+    if (predicate(items[index])) {
+      return index;
+    }
+  }
+  return -1;
+}
+
+/**
+ * A Mapping consists of two segment markers one in the generated source and one in the original
```
JoostK:
```suggestion
 * A Mapping consists of two segment markers: one in the generated source and one in the original
```

packages/compiler-cli/ngcc/src/sourcemaps/source_file.ts
```diff
@@ -0,0 +1,302 @@
...
+  readonly originalSource: SourceFile;
+  readonly originalSegment: SegmentMarker;
+  readonly name?: string;
+}
+
+/**
+ * Find `item` in `items` or, if it is not found, push it to the end of the array.
+ * @param item the item to look for.
+ * @param items the collection in which to look for `item`.
+ * @returns the index of the `item` in `items`.
```
JoostK: Many _items_ in this docblock 😆

packages/compiler-cli/ngcc/src/sourcemaps/source_file.ts
```diff
@@ -0,0 +1,302 @@
...
+}
+
+/**
+ * Parse the raw mappings into a collection of parsed mappings (stored in the `mappings`) and a
+ * tree of `SourceFile`s (stored in the `sources` property).
+ */
+export function parseMappings(
+    rawMap: RawSourceMap | null, sources: (SourceFile | null)[]): Mapping[] {
+  const mappings: Mapping[] = [];
+  if (rawMap !== null) {
```
JoostK: nit: considering an early return here

packages/compiler-cli/ngcc/src/sourcemaps/source_file.ts
```diff
@@ -0,0 +1,302 @@
...
+    const rawMappings = decode(rawMap.mappings);
+    if (rawMappings !== null) {
```
JoostK: Together with an early return here would save two levels of indentation

packages/compiler-cli/ngcc/src/sourcemaps/source_file_loader.ts
```diff
@@ -0,0 +1,132 @@
...
+ * Since a source map might reference other source files, these are also loaded as needed.
+ *
+ * This is done recursively. The result is a "tree" of `SourceFile` objects, each containing
+ * mappings to other `SourceFile` objects as necessary.
+ */
+export class SourceFileLoader {
+  constructor(private fs: FileSystem) {}
+
+  /**
+   * Load a source file, compute its source map, and recursively load and referenced source files.
```
JoostK:
```suggestion
   * Load a source file, compute its source map, and recursively load any referenced source files.
```

JoostK (commented):
I haven't been through the tests yet, but this is awesome work. Hats off to you! 👏

packages/compiler-cli/ngcc/src/sourcemaps/segment_marker.ts
```diff
@@ -0,0 +1,86 @@
...
+  while (line > 0 && column < 0) {
+    column += lines[line - 1].length + 1;
+    line--;
+  }
+
+  return {line, column};
+}
+
+
+type Writable<T> = {
```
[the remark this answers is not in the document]
petebacondarwin: 🤷‍♂ - it was great while it lasted.

packages/compiler-cli/ngcc/src/sourcemaps/source_file.ts
```diff
@@ -0,0 +1,302 @@
...
+      /** The contents of this source file. */
+      readonly contents: string,
+      /** The raw source map (if any) associated with this source file. */
+      readonly rawMap: RawSourceMap|null,
+      /** Whether this source files' source map was inline or external. */
+      readonly inline: boolean,
+      /** Any source files referenced by the raw source map associated with this source file. */
+      readonly sources: (SourceFile|null)[]) {
+    this.contents = removeSourceMapComments(contents).replace(/\n\n$/, '\n');
+    this.lines = this.contents.split('\n');
```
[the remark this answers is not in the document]
petebacondarwin: You're right. I might do that anyway.

packages/compiler-cli/ngcc/src/sourcemaps/source_file.ts
```diff
@@ -0,0 +1,302 @@
...
+import {RawSourceMap} from './raw_source_map';
+import {SegmentMarker, compareSegments, offsetSegment, segmentDiff} from './segment_marker';
+
+export function removeSourceMapComments(contents: string): string {
+  return removeMapFileComments(removeComments(contents)).replace(/\n\n$/, '\n');
+}
+
+export class SourceFile {
+  /**
+   * The parsed mappings that have been flattened so that and intermediate source mappings have been
```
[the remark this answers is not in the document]
petebacondarwin: Me neither :-)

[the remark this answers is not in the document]
petebacondarwin:
```suggestion
   * The parsed mappings that have been flattened so that any intermediate source mappings have been
```

packages/compiler-cli/ngcc/src/sourcemaps/source_file.ts
```diff
@@ -0,0 +1,302 @@
...
+
+  /**
+   * Render the raw source map generated from the flattened mappings.
+   */
+  renderFlattenedSourceMap(): RawSourceMap {
+    const sources: SourceFile[] = [];
+    const names: string[] = [];
+
+    // Ensure a mapping line array for each line in the generated source.
+    const mappings: SourceMapMappings = new Array(this.lines.length);
```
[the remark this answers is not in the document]
petebacondarwin: The next line fills every value. Does that not make the array non-holey?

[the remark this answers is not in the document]
JoostK: In V8, it is not possible for an array to transition from holey to packed, even if all slots are filled. There's a lot of details to learn [in this post on the V8 blog](https://v8.dev/blog/elements-kinds), where this exact scenario is mentioned.

packages/compiler-cli/ngcc/src/sourcemaps/source_file.ts
petebacondarwin: The result of `findLastIndex()` can be -1, so `outgoingStartIndex` can also be -1, which would mean that the first `bToCmapping` could be `undefined`.

petebacondarwin: I'll add a comment

packages/compiler-cli/ngcc/src/sourcemaps/source_file.ts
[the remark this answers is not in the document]
petebacondarwin: I wonder if I can get a `PACKED_ELEMENTS` array and quickly populate it by using...

```
    const mappings: SourceMapMappings = this.lines.map(() => []);
```

packages/compiler-cli/ngcc/src/sourcemaps/source_file.ts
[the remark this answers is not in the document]
petebacondarwin: I did do it and I like it.

packages/compiler-cli/ngcc/src/sourcemaps/source_file.ts
petebacondarwin: I changed the code to

```
      const incomingEnd = incomingEndIndex < originalSegments.length ?
          originalSegments[incomingEndIndex] :
          undefined;
```

to avoid the out-of-bounds lookup - and this also had the benefit of adding the `| undefined` to the type automatically.

packages/compiler-cli/ngcc/src/sourcemaps/source_file.ts
```diff
@@ -0,0 +1,312 @@
...
+      // The range with `incomingStart` at 2 and `incomingEnd` at 5 has outgoing start mapping of
+      // [1,0] and outgoing end mapping of [4, 6], which also includes [4, 3].
+      //
+      const outgoingStartIndex = findLastIndex(
+          bSource.flattenedMappings,
+          mapping => compareSegments(mapping.generatedSegment, incomingStart) <= 0);
+      const outgoingEndIndex = incomingEnd !== undefined ?
+          findLastIndex(
+              bSource.flattenedMappings,
+              mapping => compareSegments(mapping.generatedSegment, incomingEnd) < 0) :
```
JoostK: This looks like the only call site of `compareSegments` where the second argument is of type `SegmentMarker|undefined`, but there's a guard that `incomingEnd !== undefined` right before so we can be safe to pass `incomingEnd!` here and remove the `undefined` check from `compareSegments`.

packages/compiler-cli/ngcc/src/sourcemaps/source_file.ts
```diff
@@ -0,0 +1,312 @@
...
+          bSource.flattenedMappings.length - 1;
+
+      for (let bToCmappingIndex = outgoingStartIndex; bToCmappingIndex <= outgoingEndIndex;
+           bToCmappingIndex++) {
+        if (bToCmappingIndex !== -1) {
```
JoostK: Could we set `outgoingStartIndex` to 0 if it starts out as -1? Then we can remove this check inside the loop.

packages/compiler-cli/ngcc/src/sourcemaps/source_file.ts
```diff
@@ -0,0 +1,312 @@
...
+  }
+  return mappings;
+}
+
+export function extractOriginalSegments(mappings: Mapping[]): SegmentMarker[] {
+  return mappings.map(mapping => mapping.originalSegment).sort(compareSegments);
+}
+
+export function computeLineLengths(str: string): number[] {
+  return (str.split(/\r?\n/)).map(s => s.length);
```
JoostK: Wouldn't the optional `\r` cause all subsequent computations to be off? There's typically a `+1` to account for the `\n`, but we can't know afterwards whether there was a `\r`, right?

JoostK: Also, using an `indexOf` check in a loop would avoid the allocations for the intermediate strings:

```ts
const lineLengths = [];
let offset = 0;
let newlineIndex = str.indexOf('\n');
while (newlineIndex !== -1) {
  lineLengths.push(newlineIndex - offset);
  offset = newlineIndex + 1;
  newlineIndex = str.indexOf('\n', offset);
}
return lineLengths;
```

(from the top of my head, no idea if it works ;))

packages/compiler-cli/ngcc/test/sourcemaps/source_file_loader_spec.ts
```diff
@@ -0,0 +1,207 @@
...
+        fs.ensureDir(_('/foo/src'));
+        fs.writeFile(_('/foo/src/index.js.map'), JSON.stringify(sourceMap));
+        const sourceFile = registry.loadSourceFile(_('/foo/src/index.js'), 'some inline content');
+        if (sourceFile === null) {
+          return fail('Expected source file to be defined');
+        }
+        expect(sourceFile.rawMap).toEqual(sourceMap);
+      });
+
+      // it('should handle missing files, etc');
```
JoostK: should it? :-)

packages/compiler-cli/ngcc/test/sourcemaps/source_file_spec.ts
```diff
@@ -0,0 +1,212 @@
...
+        });
+
+        it('should be empty array for source files with no source map mappings', () => {
+          const rawSourceMap: RawSourceMap = {mappings: '', names: [], sources: [], version: 3};
+          const sourceFile =
+              new SourceFile(_('/foo/src/index.js'), 'index contents', rawSourceMap, false, []);
+          expect(sourceFile.flattenedMappings).toEqual([]);
+        });
+
+        // it('should be the same as non-flat mappings if there is only one level of source map', ()
```
JoostK: Can this be enabled/removed?

packages/compiler-cli/ngcc/src/sourcemaps/source_file.ts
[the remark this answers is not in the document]
JoostK: I have no idea what the characteristics of that would be. At least I would expect it to result in a single allocation _and_ a packed array, so it sounds like a good approach to me.

packages/compiler-cli/ngcc/src/sourcemaps/source_file.ts
petebacondarwin: And there we not even any tests for `b` being `undefined` 🙀

packages/compiler-cli/ngcc/src/sourcemaps/source_file.ts
petebacondarwin: Great idea!

packages/compiler-cli/ngcc/src/sourcemaps/source_file.ts
petebacondarwin: It turns out that the +1 is only needed as a marker that there is an end of line. It doesn't matter how many actual characters there are at the end of the line. I have updated the `Renderer` tests to prove it.

As far as the `computeLineLengths()` implementation goes... I spent ages trying out lots of different approaches, including using `indexOf`. They all seems to have some corner cases the just didn't work properly. I can't remember them all but consider, `''`, `'abc'`, `'\n'`, `'\nabc'`, `'abc\n'`. In the case of the implementation you proposed it doesn't capture the last line if there is no final `\n`. I'll add some tests to show it.

(You can probably improve it but then you are competing with the speed of the built-in `spit()` and `map()` functions, which I expect would be pretty fast - and the intermediate array can be thrown away immediately. Without performance testing I am not sure what is better - but I like the simplicity of `split` + `map`.)

packages/compiler-cli/ngcc/test/sourcemaps/source_file_loader_spec.ts
petebacondarwin: yes I guess it should... (it might) ...but I should create tests for that

packages/compiler-cli/ngcc/test/sourcemaps/source_file_spec.ts
petebacondarwin: I was trying to decide how I could keep this test, now that I am not exposing the `mappings` in the `SourceFile` object. But I see that I can just run `parseMappings()` and use its output.

👍

packages/compiler-cli/ngcc/src/sourcemaps/source_file.ts
JoostK: Ah, I see. I realise now that mappings never refer to absolute positions from the start, only line/column numbers so that they are invariant to `\r\n` vs `\n`.

Regarding the implementation, your edge cases are indeed very valid and the extra tests are really useful. Let's keep it simple with `split` + `map` as long as there's no evidence of major perf downsides 👍

JoostK (approved):
The changes in this PR LGTM, but I just realized something:

When rendering the generated statements to JavaScript code to insert into the JS bundle, any source mapping information present in the statements is _not_ retained as all statement outputs are concatenated into a single string:

https://github.com/angular/angular/blob/7659f2e24b5caf6f4c28256b12d9a2187515ff8f/packages/compiler-cli/ngcc/src/rendering/renderer.ts#L186-L191

Although this PR should address the unmapped bytes problem as reported in FW-1558 (did we actually verify that?), it does not improve the granularity of the source maps of the generated code.

petebacondarwin:
There are a few issues.. 

1. the previous source-map basically dropped all the references to the previous original source files, which is what the JIRA issue is primarily complaining about.
2. the mappings in that source-map were only showing the lowest resolution between the merged source maps, which results in them being of minimal use.
3. the output from ngcc compilation is indeed just a single string with no additional source-map information, which means that we do not benefit from things like template source map information.

This PR fixes both 1 and 2. I have checked both.

Number 3 is a more complex issue, and I feel that given that ngcc is only providing backward compatibility to libraries, and hopefully should only be for a limited amount of time, it may not be worth investing time in.

JoostK:
> This PR fixes both 1 and 2. I have checked both.

Excellent! 💯

> Number 3 is a more complex issue, and I feel that given that ngcc is only providing backward compatibility to libraries, and hopefully should only be for a limited amount of time, it may not be worth investing time in.

Agreed, just wanted to make sure this does indeed solve the unmapped bytes problem.

petebacondarwin:
Here is a link to a document that gives a high level explanation of what is going on here: https://docs.google.com/document/d/1x1laiDVFgJtpiw49ozwr91LG7hqohytWDiptJ93nJcE

gkalpak (approved):
Nice! +1 for the ASCII art 😛

packages/compiler-cli/ngcc/BUILD.bazel
```diff
@@ -35,6 +35,7 @@ ts_library(
         "@npm//magic-string",
         "@npm//semver",
         "@npm//source-map",
+        "@npm//sourcemap-codec",
```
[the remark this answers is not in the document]
petebacondarwin: Yes! Good spot. We really should have a way of automating this... like generating the dependencies automatically via Bazel...

packages/compiler-cli/ngcc/src/sourcemaps/source_file.ts
```diff
@@ -0,0 +1,313 @@
...
+  readonly originalSegment: SegmentMarker;
+  readonly name?: string;
+}
+
+/**
+ * Find the index of `item` in the `items` array.
+ * If it is not found, then push `item` to the end of the array and returns its new index.
+ *
+ * @param item the item to look for.
+ * @param items the collection in which to look for `item`.
```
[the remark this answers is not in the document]
petebacondarwin: Nope :-)

packages/compiler-cli/ngcc/src/sourcemaps/segment_marker.ts
```diff
@@ -0,0 +1,86 @@
...
+ * @returns a positive number if `a` is after `b`, a negative number if `b` is after `a`
+ * and zero if they are at the same position.
+ */
+export function compareSegments(a: SegmentMarker, b: SegmentMarker): number {
+  return a.line === b.line ? a.column - b.column : a.line - b.line;
+}
+
+// The `1` is to indicate a newline character between the lines
+// Note that in the actual contents there could be more than one character -
+// e.g. \r\n - that indicate a newline but that is not important here.
```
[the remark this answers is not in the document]
petebacondarwin: improved the comment

packages/compiler-cli/ngcc/BUILD.bazel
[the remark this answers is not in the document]
petebacondarwin: added (and to the top level `package.json` since it was only available in the node_modules as a transitive dependency).

PR #35146, opened 2020-02-04
fix(ivy): template type-check errors from TS should not use NG error codes

alxhub:
A bug previously caused the template type-checking diagnostics produced by
TypeScript for template expressions to use -99-prefixed error codes. These
codes are converted to "NG" errors instead of "TS" errors during diagnostic
printing. This commit fixes the issue.

### DIFF
--- a/packages/compiler-cli/src/ngtsc/typecheck/src/diagnostics.ts
+++ b/packages/compiler-cli/src/ngtsc/typecheck/src/diagnostics.ts
@@ -8,7 +8,6 @@
 import {AbsoluteSourceSpan, ParseSourceSpan} from '@angular/compiler';
 import * as ts from 'typescript';
 
-import {ErrorCode, ngErrorCode} from '../../diagnostics';
 import {getTokenAtPosition} from '../../util/src/typescript';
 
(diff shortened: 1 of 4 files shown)


packages/compiler-cli/src/ngtsc/typecheck/src/diagnostics.ts
```diff
@@ -216,7 +215,7 @@ export function makeTemplateDiagnostic(
     return {
       source: 'ngtsc',
       category,
-      code: ngErrorCode(code), messageText,
+      code: code, messageText,
```
AndrewKushnir: nit: we can just say `code,` here instead of `code: code,`.

AndrewKushnir (approved):
LGTM (just one nit) 👍

packages/compiler-cli/src/ngtsc/typecheck/src/diagnostics.ts
alxhub: Done!

AndrewKushnir: 👍

PR #35147, opened 2020-02-04
refactor(benchpress): made all it callbacks async

wagnermaciel:
## PR Checklist

## PR Type
- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

`it` callbacks in `//modules/benchmarks` are not async.

## What is the new behavior?
`it` callbacks in `//modules/benchmarks` are async.

- [ ] Yes
- [x] No

## Other information
This is part of a stream of PRs that are hoping to hoping to unify internal and external performance benchmark drivers.


modules/benchmarks/e2e_test/old/compiler_perf.ts
```diff
@@ -15,7 +15,7 @@ describe('ng2 compiler benchmark', function() {
 
   afterEach(verifyNoBrowserErrors);
 
-  it('should log withBindings stats', function(done) {
+  it('should log withBindings stats', async() => {
```
jelbourn: Looks like this still needs to `await` the `runBenchmark` call, otherwise the test will complete before `runBenchmark` finished. 

(before it directly called `done` with the `then`)

Here and elsewhere

wagnermaciel: Yep, you're right. I didn't want to have to implement that in this PR to keep it small but ¯\_(ツ)_/¯

I should note there's a way to have bazel catch this for us by just specifying the `tsconfig` arg of the `ts_library` rule. I'll make those changes in this PR as well.

wagnermaciel: I don't think we need to be updating `//modules/benchmarks/e2e_test`. It is already outdated, using functions that no longer exist. Any idea who I can ask about the purpose of this folder?

wagnermaciel: For the sake of documenting our conversation - This folder doesn't seem to be used anymore. We are comfortable deleting it.

modules/benchmarks/src/tree/BUILD.bazel
```diff
@@ -5,13 +5,15 @@ package(default_visibility = ["//visibility:public"])
 ts_library(
     name = "util_lib",
     srcs = ["util.ts"],
+    tsconfig = "//modules/benchmarks:tsconfig-e2e.json",
```
devversion: Is there any reason why the e2e tsconfig need to be used? Isn't the default tsconfig we provide for `ts_library` sufficient?

devversion (commented):
Nice cleanup! Should we consider disabling the selenium promise manager to enforce that async/await is consistently used? (and to avoid that promises are magically awaited)

```js
  // https://www.protractortest.org/#/async-await
  SELENIUM_PROMISE_MANAGER: false,
```

jelbourn (approved):
LGTM aside from Paul's comments

modules/benchmarks/src/tree/BUILD.bazel
wagnermaciel: No, the default `tsconfig` we provide for `ts_library` disables the `must-use-promises` rule. See [//packages/tsconfig-test.json](https://github.com/angular/angular/blob/master/packages/tsconfig-test.json).

devversion: I see. It would be nice if we have a comment for this somewhere, but we don't want to add a comment for every `ts_library`. Maybe it would be worth mentioning in the commit at least?

wagnermaciel: Yeah, I agree. I changed the commit message for [commit 69ce145](https://github.com/angular/angular/pull/35147/commits/69ce1458f0d8fb230307c39bd49b81da237c8d2d) to document this behavior.

wagnermaciel: Hmm, seems like I'm getting a [lint error from CircleCi](https://circleci.com/gh/angular/angular/614658?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) now...
"ERROR: The commit message header is longer than 120 characters"

devversion: There are multiple options:

1. Splitting the changes into two individual commits. First using `async/await`, and second specifying the e2e tsconfig.
2. Just reducing the commit message title to what it was before, and mentioning the tsconfig change only in the commit message body.

Going with the first option would definitely be ideal, but I don't think it matters too much 😄

wagnermaciel: I fixed it - the commit message just needed to have two newlines between what's considered the message header and the message body (I only had one).

Lint is passing now 👍

wagnermaciel:
> Nice cleanup! Should we consider disabling the selenium promise manager to enforce that async/await is consistently used? (and to avoid that promises are magically awaited)
> 
> ```js
>   // https://www.protractortest.org/#/async-await
>   SELENIUM_PROMISE_MANAGER: false,
> ```

@jelbourn What are your thoughts? I know very little about what this does but it sounds like it's something we could use

jelbourn:
I don't know anything about it, but in general I would avoid any "it just magically works" features. Can try turning that off in a follow-up change.

devversion (approved):

pkozlowski-opensource (approved):

jelbourn:
Presubmit looks good, one failing target looks like a flake

kara:
@wagnermaciel What branches should this PR target?

jelbourn:
I added the target label

PR #35153, opened 2020-02-05
refactor: depend on bazelisk rather than directly on Bazel

alexeagle:
This has a couple benefits:
- we now use a .bazelversion file rather than package.json to pin the version of bazel we want. This means even if you install bazel on your computer rather than via yarn, you'll still get a warning if your bazel version is wrong.
- you no longer end up downloading three copies of bazel due to bugs in both npm and yarn where they download all tarballs before checking the metadata to see which are usable on the local platform.
- bazelisk correctly handles the tools/bazel trick for wrapping functionality, which we want to use to instrument developer build latencies
- it should fix the strange issue with Windows CI seeing bazel being unlinked out from under the binary; I removed the workaround

Note that @angular/bazel package isn't updated in this PR; I'm working on a plan to remove the opt-in preview bits with schematics/builder for running bazel this way.


filipesilva:
CI is failing because `integration/bazel` is still trying to run bazel:
```
$ bazel build ... --noshow_progress && bazel test ...
/bin/sh: 1: bazel: not found
```

philwo:
@alexeagle What's necessary to get this in and can I help in some way?

alexeagle:
thanks @philwo I think I just need to work through misconfigurations of the integration tests wrt. how they find bazel binary

alexeagle:
@googlebot I signed it!

gregmagolan:
Replaced by https://github.com/angular/angular/pull/36078

PR #35156, opened 2020-02-05
fix(ivy): `LFrame` needs to release memory on `leaveView()`

### LINKED ISSUE #35148: Ivy doesn’t deallocate memory of destroyed components correctly
# 🐞 bug report

### Affected Package

The issue is caused by package @angular/core@9.0.0-rc.14, earlier versions probably as well.

### Is this a regression?

This issue does not occur in Angular 8 + View Engine.

### Description

Ivy does not seem to correctly deallocate memory of destroyed components in all cases. We’ve noticed an issue that when navigating away from a route, at least some of the loaded components are getting destroyed **but not** deallocated and garbage collected. (The issue might not be restricted to routing.)

The problem is that unnecessary memory is not deallocated, which might lead to problems on mobile and other low-memory devices and might also have security implications (e.g. when sensitive state remains in memory, even after a user has logged out).

## 🔬 Minimal Reproduction

Here is a minimal repo of the problem. It’s a plain Angular CLI application with three routes and four components:

* `/#/route1` (`Route1Component`) includes the `ContentComponent` as a child that loads the contents of a large file retrieved via the network into a field of its class during `ngOnInit`, consuming some memory. This component logs to the console when it’s being initialized or destroyed.
* `/#/route2` (`Route2Component`) and `route3` (`Route3Component`) don’t include this component.

**Repro steps:**

```
git clone https://github.com/thinktecture-labs/ivy-mem-leak.git
cd ivy-mem-leak
npm i
npm start
```

* Open http://localhost:4200/#/route3
* Open Chrome dev tools > Memory
* Click 🔴 _Take Heap Snapshot_
* Open http://localhost:4200/#/route2
* Take a snapshot (size should stay roughly the same)
* Open http://localhost:4200/#/route1, wait until the large file has been downloaded (check the _Network_ tab)
* Take a snapshot (size should be larger than before)
* Open http://localhost:4200/#/route2
* Take a snapshot (size should not significantly change, even though the `ContentComponent` was destroyed, check the _Console_ tab)

Stop `ng serve` and compare this with the behavior of Angular 8:

```
git checkout ng8
npm i
npm start
```

* Click 🚫 _Clear All Profiles_
* Open http://localhost:4200/#/route3
* Take a snapshot
* Open http://localhost:4200/#/route2
* Take a snapshot (size should stay roughly the same)
* Open http://localhost:4200/#/route1
* Take a snapshot (should be larger than before)
* Open http://localhost:4200/#/route2
* Take a snapshot (should be similar to the first two snapshots again)

## 🔥 Exception or Error

The memory of `route1`’s components doesn’t seem to be correctly deallocated when leaving `route1` for `route2` on Angular 9/Ivy:

| Angular 9/Ivy  | Angular 8/View Engine |
| ------------- | ------------- |
| image  | image |

As per DevTools, (at least) the component instance of the `ContentComponent` seems to be retained by a `contextLView` array even after it has been destroyed:

image

Is this behavior intended or is this a bug?

Researched by @ManuelRauber, @thomashilzendegen and me.

## 🌍  Your Environment

**Angular Version:**
<pre><code>
                |___/

Angular CLI: 9.0.0-rc.12
Node: 12.14.1
OS: darwin x64

Angular: 9.0.0-rc.14
... router
Ivy Workspace: Yes

@angular-devkit/architect         0.900.0-rc.12
@angular-devkit/build-angular     0.900.0-rc.12
@angular-devkit/build-optimizer   0.900.0-rc.12
@angular-devkit/build-webpack     0.900.0-rc.12
@angular-devkit/core              9.0.0-rc.12
@angular-devkit/schematics        9.0.0-rc.12
@angular/cli                      9.0.0-rc.12
@ngtools/webpack                  9.0.0-rc.12
@schematics/angular               9.0.0-rc.12
@schematics/update                0.900.0-rc.12
</code></pre>

**Anything else relevant?**

—

mhevery:
Root cause is that for perf reasons we cache `LFrame` so that we don't have to allocate it all the time. To be extra fast we clear the `LFrame` on `enterView()` rather that on `leaveView()`. The implication of this strategy is that the deepest `LFrame` will retain objects until the `LFrame` allocation depth matches the deepest object.

The fix is to simply clear the `LFrame` on `leaveView()` rather then on `enterView()`

Fix #35148

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/render3/state.ts
+++ b/packages/core/src/render3/state.ts
@@ -7,7 +7,7 @@
  */
 
 import {StyleSanitizeFn} from '../sanitization/style_sanitizer';
-import {assertDefined} from '../util/assert';
+import {assertDefined, assertEqual} from '../util/assert';
 import {assertLViewOrUndefined} from './assert';
 import {TNode} from './interfaces/node';
 import {CONTEXT, DECLARATION_VIEW, LView, OpaqueViewState, TVIEW, TView} from './interfaces/view';
@@ -374,28 +374,7 @@ export function enterDI(newView: LView, tNode: TNode) {
   instructionState.lFrame = newLFrame;
   newLFrame.previousOrParentTNode = tNode !;
   newLFrame.lView = newView;
-  if (ngDevMode) {
-    // resetting for safety in dev mode only.
-    newLFrame.isParent = DEV_MODE_VALUE;
-    newLFrame.selectedIndex = DEV_MODE_VALUE;
-    newLFrame.contextLView = DEV_MODE_VALUE;
-    newLFrame.elementDepthCount = DEV_MODE_VALUE;
-    newLFrame.currentNamespace = DEV_MODE_VALUE;
-    newLFrame.currentSanitizer = DEV_MODE_VALUE;
-    newLFrame.bindingRootIndex = DEV_MODE_VALUE;
-    newLFrame.currentQueryIndex = DEV_MODE_VALUE;
-  }
 }
-
-const DEV_MODE_VALUE: any =
-    'Value indicating that DI is trying to read value which it should not need to know about.';
-
-/**
- * This is a light weight version of the `leaveView` which is needed by the DI system.
- *
- * Because the implementation is same it is only an alias
- */
-export const leaveDI = leaveView;
 
 /**
  * Swap the current lView with a new lView.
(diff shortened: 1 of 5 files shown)


mhevery:
[presubmit](http://test/OCL:293473862:BASE:293473903:1580947413999:38408d37)

mhevery:
[presubmit](http://test/OCL:293473862:BASE:294505492:1581456088184:75b192af)

pkozlowski-opensource (changes requested):
The code LGTM but I'm surprised that there is code size increase (and it is even bigger than what is written in the code as the CI check fails: https://circleci.com/gh/angular/angular/621857#tests/containers/3)

mhevery:
> The code LGTM but I'm surprised that there is code size increase (and it is even bigger than what is written in the code as the CI check fails: https://circleci.com/gh/angular/angular/621857#tests/containers/3)

I think this is the case where the limit was close to 500 bytes and this CL took os over, so now you have the 500 bytes plus this CL. `226144 - 225584 = 560`

pkozlowski-opensource (approved):

packages/core/src/render3/state.ts
```diff
@@ -461,8 +444,52 @@ function createLFrame(parent: LFrame | null): LFrame {
...
+function leaveViewLight(): LFrame {
+  const oldLFrame = instructionState.lFrame;
+  instructionState.lFrame = oldLFrame.parent;
+  oldLFrame.previousOrParentTNode = null !;
+  oldLFrame.lView = null !;
+  return oldLFrame;
+}
+
+/**
+ * This is a light weight version of the `leaveView` which is needed by the DI system.
```
kara:
```suggestion
 * This is a lightweight version of the `leaveView` which is needed by the DI system.
```

packages/core/src/render3/state.ts
```diff
@@ -461,8 +444,52 @@ function createLFrame(parent: LFrame | null): LFrame {
...
+  oldLFrame.tView = null !;
+  oldLFrame.selectedIndex = 0;
+  oldLFrame.contextLView = null !;
+  oldLFrame.elementDepthCount = 0;
+  oldLFrame.currentDirectiveIndex = -1;
+  oldLFrame.currentNamespace = null;
+  oldLFrame.currentSanitizer = null;
+  oldLFrame.bindingRootIndex = -1;
+  oldLFrame.bindingIndex = -1;
+  oldLFrame.currentQueryIndex = 0;
```
kara: I wonder if we could share more code with `createLFrame`? It sets properties in similar ways.

kara (approved):
LGTM

packages/core/src/render3/state.ts
mhevery: fixed

packages/core/src/render3/state.ts
mhevery: Yes, that would be nice. The issue is that in `createLFrame` we are creating an object where as here we are modifying it. We could create an empty object and then set properties on it, but I am concern that it may have negative perf impact as we are adding properties one by one rather than in bulk.

kara: Yeah, that was what I was thinking, but thought maybe there could be a perf reason. Might be good to have that in a comment.

PR #35171, opened 2020-02-05
ci: ensure saucelabs browsers can load karma test page

devversion:
[Based on the Saucelabs docs](https://support.saucelabs.com/hc/en-us/articles/225106887-Safari-and-Internet-Explorer-Won-t-Load-Website-When-Using-Sauce-Connect-on-Localhost), Saucelabs runs a process that usually takes up all requests to `localhost` / `127.0.0.1` and forwards them through the tunnel. It looks like this process somehow stopped working for Saucelabs Android devices. I tried updating to Android 9 and 10, but the same issue surfaced.

In the past, we had similar issues and the issue magically got resolved after we opened up a support ticket. They mentioned that they had issues (no real details), and that everything should work again.

In general though, I think we could improve the stability of our tests by not relying on this Saucelabs process. Even though this issue doesn't surface too often, we had it a couple of times now, and identifying the issue / temporarily disabling tests slows down the overall productivity. Not mentioning the fact that all PRs are red..

I personally think it would be worth doing such a change, so that we don't need to worry about it in the future, don't need to spend resources investigating, and that we aren't dependent on the Saucelabs support to fix/investigate it. The impact of such outage seems to cause unnecessary slow-down for the caretaker, dev-infra (in terms of investigating) and the PR submitters (who need to rebase etc.). We can make this more stable long-term by adding an alias for `localhost` to the CircleCI hosts, and using that from within the Saucelabs browsers. These requests will always go through the tunnel (might be even faster that way). Similar solutions seem to be recommended by SauceLabs if specifically needed ports aren't forwarded by the Saucelabs VM process, so the solution seems somewhat proven.

### DIFF
--- a/tools/saucelabs/karma-saucelabs.js
+++ b/tools/saucelabs/karma-saucelabs.js
@@ -21,22 +21,18 @@ try {
   // KARMA_WEB_TEST_MODE is set which informs /karma-js.conf.js that it should
   // run the test with the karma saucelabs launcher
   process.env['KARMA_WEB_TEST_MODE'] = 'SL_REQUIRED';
+  // Saucelabs parameters read from a temporary file that is created by the `sauce-service`. This
+  // will be `null` if the test runs locally without the `sauce-service` being started.
+  const saucelabsParams = readLocalSauceConnectParams();
   // Setup required SAUCE_* env if they are not already set
   if (!process.env['SAUCE_USERNAME'] || !process.env['SAUCE_ACCESS_KEY'] ||
       !process.env['SAUCE_TUNNEL_IDENTIFIER']) {
-    try {
-      // The following path comes from /tools/saucelabs/sauce-service.sh.
-      // We setup the required saucelabs environment variables here for the karma test
-      // from a json file under /tmp/angular/sauce-service  so that we don't break the
-      // test cache with a changing SAUCE_TUNNEL_IDENTIFIER provided through --test_env
-      const scParams = require('/tmp/angular/sauce-service/sauce-connect-params.json');
-      process.env['SAUCE_USERNAME'] = scParams.SAUCE_USERNAME;
-      process.env['SAUCE_ACCESS_KEY'] = scParams.SAUCE_ACCESS_KEY;
-      process.env['SAUCE_TUNNEL_IDENTIFIER'] = scParams.SAUCE_TUNNEL_IDENTIFIER;
-    } catch (e) {
-      console.error(e.stack || e);
-      console.error(
-          `!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+    // We print a helpful error message below if the required Saucelabs parameters have not
+    // been specified in test environment, and the `sauce-service` params file has not been
+    // created either.
+    if (saucelabsParams === null) {
+      console.error(`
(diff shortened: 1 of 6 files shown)


IgorMinar (approved):
Lgtm. Thanks Paul!

IgorMinar:
Fyi @gregmagolan @josephperrott  - this might affect some of your prs.

tools/saucelabs/sauce-service.sh
```diff
@@ -147,6 +147,12 @@ service-setup-command() {
     "--user ${SAUCE_USERNAME}"
     # Don't add the --api-key here so we don't echo it out in service-pre-start
   )
+
+  if [[ -n "${SAUCE_LOCALHOST_ALIAS_DOMAIN}" ]]; then
+    # Ensures that requests to the localhost alias domain are always resolved through the tunnel.
+    sauce_args+=("--tunnel-domains ${SAUCE_LOCALHOST_ALIAS_DOMAIN}")
```
gregmagolan: Add comment that this depends on a line in `/etc/hosts`?

tools/saucelabs/karma-saucelabs.js
```diff
@@ -23,7 +23,7 @@ try {
   process.env['KARMA_WEB_TEST_MODE'] = 'SL_REQUIRED';
   // Setup required SAUCE_* env if they are not already set
   if (!process.env['SAUCE_USERNAME'] || !process.env['SAUCE_ACCESS_KEY'] ||
-      !process.env['SAUCE_TUNNEL_IDENTIFIER']) {
+      !process.env['SAUCE_TUNNEL_IDENTIFIER'] || !process.env['SAUCE_LOCALHOST_ALIAS_DOMAIN']) {
```
gregmagolan: Hmmmm. The logic here seems wrong now since SAUCE_LOCALHOST_ALIAS_DOMAIN is optional and I'm guessing it won't be set when running this locally since it would require stomping over 127.0.0.1 in `/etc/hosts` on your local machine.

This made sense with SAUCE_USERNAME, SAUCE_ACCESS_KEY & SAUCE_TUNNEL_IDENTIFIER being required.

On this note, are we toast now for running SauceLabs tests locally without the localhost alias?

gregmagolan: Also, whichever way we go with that new env the following comment may need updating in this file
```
          `!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
!!! Make sure that you have run "yarn bazel run //tools/saucelabs:sauce_service_setup"
!!! (or "./tools/saucelabs/sauce-service.sh setup") before the test target. Alternately
!!! you can provide the required SAUCE_* environment variables (SAUCE_USERNAME, SAUCE_ACCESS_KEY &
!!! SAUCE_TUNNEL_IDENTIFIER) to the test with --test_env or --define but this may prevent bazel from
!!! using cached test results.
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!`);
```

devversion: Updated based on what we discussed on Slack. In summary (for later reference): Running the tests locally still works. Using the localhost alias domain is optional, but commonly done on CI to improve overall stability. We don't want to run into issues with the localhost resolution again on CI.

PR #35172, opened 2020-02-05
fix(ivy): ensure module imports are instantiated before the module being declared

atscott:


packages/core/src/di/r3_injector.ts
```diff
@@ -290,7 +290,6 @@ export class R3Injector {
     }
 
     // Track the InjectorType and add a provider for it.
```
AndrewKushnir: nit: it looks like comment should also be updated/rephrased to reflect code change.

atscott: Removed.

packages/core/src/di/r3_injector.ts
```diff
@@ -330,6 +326,9 @@ export class R3Injector {
         }
       }
     }
+    // Track the InjectorType and add a provider for it.
```
alxhub: Add a comment here about the ordering being important - recursing into children and _then_ adding the injectortype to the set.

packages/core/test/acceptance/ng_module_spec.ts
```diff
@@ -77,6 +77,25 @@ describe('NgModule', () => {
     });
   });
 
+  it('initializes the module imports before the module itself', () => {
```
alxhub: This test is great, but it would be nice to have a simple r3_injector_spec test as well.

alxhub (approved):

atscott:
[presubmit](https://test.corp.google.com/ui#id=OCL:293702733:BASE:293703076:1581034970421:73b064f1) (failing tests passed in a rerun)

PR #35176, opened 2020-02-06
docs: fix routing code to use `navigate`

### LINKED ISSUE #34917: Potential bug in the routing example stackblitz program

### Description

The example presented in https://angular.io/generated/live-examples/router/stackblitz.html (live example) does not exhibit the behaviour claimed in the documentation at https://angular.io/guide/router#query-parameters-and-fragments. Namely, this:

> Now, you can click on the Admin button, which takes you to the Login page with the provided queryParamMap and fragment. After you click the login button, notice that you have been redirected to the Admin Dashboard page with the query parameters and fragment still intact in the address bar.

The query parameter and the fragment both get lost when you are forwarded to the "admin" page by the login button. You can see that the "Admin" page displays `Null` both for `Session ID` and for `Token`, as they are also absent in the address bar. 

The cause seems to be the `navigateByUrl()` call in `auth/login/login.component.ts` at line 40. This is what the official doc says for this function referring to its `extras` parameter:

>  NavigationExtras | An object containing properties that modify the navigation strategy. The function ignores any properties in the NavigationExtras that would change the provided URL.

So if you replace line 40 
```
this.router.navigateByUrl(redirect, navigationExtras);
```
with
```
this.router.navigate([this.authService.redirectUrl], navigationExtras);
```
the app behaviour is then consistent with the expected description in the angular write-up. While `navigate()` is normally doing a relative navigation, in the absence of a "start" (`relativeTo`) position it will do the absolute navigation, but without the adverse effect of `navigateByUrl()`. Maybe not the most optimal solution though

But notice that if this solution is adopted, the `UrlTree` `redirect` variable is then not necessary

Also the Angular Doc write–up has to be updated as well to whatever solution is adopted

https://angular.io/generated/live-examples/router/stackblitz.html

and the documentation at

https://angular.io/guide/router#query-parameters-and-fragments

### Reproduction Steps**
Navigate to https://angular.io/generated/live-examples/router/stackblitz.html, then to `Admin` tab and click the `Login` button. You will see both `Session ID` and `Token` being `None`

`Session ID` should be `123456789`, and `Token` should be `anchor`, and not `None`

Otherwise how are they persistent?

## 🔥 Exception or Error
<pre><code>

</code></pre>

sonukapoor:
Previously, the example in the `router` guide was not
preserving the query params after logging in.

This commit fixes this by using `navigate` instead of
using `navigateByUrl`, which ignores any properties in
the `NavigationExtras` that would change the provided URL.

Fixes #34917

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #34917

- [ ] Yes
- [X] No

### DIFF
--- a/aio/content/examples/router/src/app/auth/login/login.component.ts
+++ b/aio/content/examples/router/src/app/auth/login/login.component.ts
@@ -25,15 +25,9 @@ export class LoginComponent {
     this.authService.login().subscribe(() => {
       this.setMessage();
       if (this.authService.isLoggedIn) {
-        // Get the redirect URL from our auth service
-        // If no redirect has been set, use the default
-        let redirectUrl = this.authService.redirectUrl || '/admin';
-
-        // If the redirectUrl contains an aux outlet name /admin(popup:compose)
-        // Then get rid of that and just take the first part of the Url.
-        if (redirectUrl.indexOf('(') !== -1) {
-          redirectUrl = redirectUrl.split('(')[0];
-        }
+        // Usually you would use the redirect URL from the auth service.
+        // However to keep the example simple, we will always redirect to `/admin`.
+        const redirectUrl = '/admin';
 
         // #docregion preserve
         // Set our navigation extras object
--- a/aio/content/examples/router/src/app/auth/login/login.component.ts
+++ b/aio/content/examples/router/src/app/auth/login/login.component.ts
@@ -1,8 +1,7 @@
 // #docregion
-import { Component }        from '@angular/core';
-import { Router,
-         NavigationExtras } from '@angular/router';
-import { AuthService }      from '../auth.service';
+import { Component } from '@angular/core';
+import { NavigationExtras, Router } from '@angular/router';
+import { AuthService } from '../auth.service';
 
 @Component({
   selector: 'app-login',
@@ -28,7 +27,13 @@ export class LoginComponent {
       if (this.authService.isLoggedIn) {
         // Get the redirect URL from our auth service
         // If no redirect has been set, use the default
-        let redirect = this.authService.redirectUrl ? this.router.parseUrl(this.authService.redirectUrl) : '/admin';
+        let redirectUrl = this.authService.redirectUrl || '/admin';
+
+        // If the redirectUrl contains an aux outlet name /admin(popup:compose)
+        // Then get rid of that and just take the first part of the Url.
+        if (redirectUrl.indexOf('(') !== -1) {
+          redirectUrl = redirectUrl.split('(')[0];
+        }
 
         // #docregion preserve
         // Set our navigation extras object
@@ -39,7 +44,7 @@ export class LoginComponent {
         };
 
         // Redirect the user
-        this.router.navigateByUrl(redirect, navigationExtras);
+        this.router.navigate([redirectUrl], navigationExtras);
         // #enddocregion preserve
       }
     });
--- a/aio/content/guide/router.md
+++ b/aio/content/guide/router.md
@@ -3302,11 +3302,11 @@ Although it doesn't actually log in, it has what you need for this discussion.
 It has an `isLoggedIn` flag to tell you whether the user is authenticated.
 Its `login` method simulates an API call to an external service by returning an
 observable that resolves successfully after a short pause.
-The `redirectUrl` property will store the attempted URL so you could navigate to it after authenticating.
+The `redirectUrl` property stores the URL that the user wanted to access so you can navigate to it after authentication.
 
 <div class="alert is-helpful">
 
-To keep the stackblitz example simple, we will always redirect an un-authenticated user to `/admin`. 
+To keep things simple, this example redirects unauthenticated users to `/admin`. 
 
 </div>
 
--- a/aio/content/guide/router.md
+++ b/aio/content/guide/router.md
@@ -3302,7 +3302,13 @@ Although it doesn't actually log in, it has what you need for this discussion.
 It has an `isLoggedIn` flag to tell you whether the user is authenticated.
 Its `login` method simulates an API call to an external service by returning an
 observable that resolves successfully after a short pause.
-The `redirectUrl` property will store the attempted URL so you can navigate to it after authenticating.
+The `redirectUrl` property will store the attempted URL so you could navigate to it after authenticating.
+
+<div class="alert is-helpful">
+
+To keep the stackblitz example simple, we will always redirect an un-authenticated user to `/admin`. 
+
+</div>
 
 Revise the `AuthGuard` to call it.
 
--- a/aio/content/examples/router/src/app/auth/login/login.component.1.ts
+++ b/aio/content/examples/router/src/app/auth/login/login.component.1.ts
@@ -25,9 +25,9 @@ export class LoginComponent {
     this.authService.login().subscribe(() => {
       this.setMessage();
       if (this.authService.isLoggedIn) {
-        // Get the redirect URL from our auth service
-        // If no redirect has been set, use the default
-        const redirectUrl = this.authService.redirectUrl || '/admin';
+        // Usually you would use the redirect URL from the auth service.
+        // However to keep the example simple, we will always redirect to `/admin`.
+        const redirectUrl = '/admin';
 
         // Redirect the user
         this.router.navigate([redirectUrl]);
--- a/aio/content/examples/router/src/app/auth/login/login.component.1.ts
+++ b/aio/content/examples/router/src/app/auth/login/login.component.1.ts
@@ -1,6 +1,6 @@
 // #docregion
-import { Component }   from '@angular/core';
-import { Router }      from '@angular/router';
+import { Component } from '@angular/core';
+import { Router } from '@angular/router';
 import { AuthService } from '../auth.service';
 
 @Component({
@@ -27,10 +27,10 @@ export class LoginComponent {
       if (this.authService.isLoggedIn) {
         // Get the redirect URL from our auth service
         // If no redirect has been set, use the default
-        let redirect = this.authService.redirectUrl ? this.router.parseUrl(this.authService.redirectUrl) : '/admin';
+        const redirectUrl = this.authService.redirectUrl || '/admin';
 
         // Redirect the user
-        this.router.navigateByUrl(redirect);
+        this.router.navigate([redirectUrl]);
       }
     });
   }


sonukapoor:
@gkalpak Not sure why the `test_aio` step is failing. It's reporting a test failure, that I did not touch. Can you take a look?

aio/content/examples/router/src/app/auth/login/login.component.ts
```diff
@@ -28,7 +27,13 @@ export class LoginComponent {
       if (this.authService.isLoggedIn) {
         // Get the redirect URL from our auth service
         // If no redirect has been set, use the default
-        let redirect = this.authService.redirectUrl ? this.router.parseUrl(this.authService.redirectUrl) : '/admin';
+        let redirectUrl = this.authService.redirectUrl || '/admin';
+
+        // If the redirectUrl contains an aux outlet name /admin(popup:compose)
```
kapunahelewong: Does aux mean auxiliary? If so, can we spell it out?

aio/content/examples/router/src/app/auth/login/login.component.ts
```diff
@@ -28,7 +27,13 @@ export class LoginComponent {
...
+        // Then get rid of that and just take the first part of the Url.
```
kapunahelewong:
```suggestion
        // Remove auxiliary outlet name and take only the first part of the URL.
```

kapunahelewong (changes requested):
Just couple of small copy suggestions.

aio/content/examples/router/src/app/auth/login/login.component.ts
```diff
@@ -28,7 +27,13 @@ export class LoginComponent {
       if (this.authService.isLoggedIn) {
         // Get the redirect URL from our auth service
         // If no redirect has been set, use the default
-        let redirect = this.authService.redirectUrl ? this.router.parseUrl(this.authService.redirectUrl) : '/admin';
+        let redirectUrl = this.authService.redirectUrl || '/admin';
```
gkalpak:
```suggestion
        const redirectUrl = this.authService.redirectUrl || '/admin';
```

aio/content/examples/router/src/app/auth/login/login.component.ts
```diff
@@ -28,7 +27,13 @@ export class LoginComponent {
...
+
+        // If the redirectUrl contains an aux outlet name /admin(popup:compose)
+        // Then get rid of that and just take the first part of the Url.
+        if (redirectUrl.indexOf('(') !== -1) {
+          redirectUrl = redirectUrl.split('(')[0];
```
gkalpak: This is a hack and we should not show this to users :grin:
We need to figure out the correct way to handle this. Can you give the steps to reproduce the issue with aux outlets?

gkalpak (changes requested):

aio/content/examples/router/src/app/auth/login/login.component.ts
gkalpak: By taking a brief look at the docs, I couldn't figure out the correct way to do this 😕
If there is indeed no way to do this, we should change the guide (or at least the example temporarily to not show something that will not work 😱

gkalpak: cc @atscott (as the router owner) - any ideas?

atscott: If the auxiliary route is active, is there any reason we *have* to clear it here? How about just keeping the popup? It's sticking around for the other navigations.

Otherwise, you'd have to clear the auxiliary routes with something like: `this.router.navigate([{outlets: {popup: null, primary: redirectUrl}}], navigationExtras);`

sonukapoor: @atscott Keeping the auxiliary route creates problems with the redirect URL. The URL already contains the aux path and calling `navigate` with that URL, adds another aux path on top of that. I have tried your solution, but that does not work either. Here is a reproducible stackblitz:

https://stackblitz.com/edit/angular-mx8rau?file=src/app/auth/login/login.component.ts

Steps to follow:

1. Click on the Login tab
2. Click on the Login button
3. See an error in the console.

atscott: You would need to use `admin` instead of `/admin` for the redirectUrl. Can you give me the steps to reproduce the case where you're getting two auxiliary routes? I'm only ever getting one after the redirect.

sonukapoor: @atscott. I replaced the code in the stackblitz with the original we had in place: `this.router.navigate([redirectUrl], navigationExtras);`

Please see below the steps to see two auxiliary paths in the URL. 
Make sure that you are logged out. 

1. Click on Contacts
2. Click on Admin
3. Click on Login 
4. Click on the Login button
5. You will end up with an URL like this:

https://angular-mx8rau.stackblitz.io/admin%28popup:compose%29(popup:compose)?session_id=123456789#anchor

sonukapoor: > You would need to use 'admin' instead of '/admin' for the redirectUrl.

This seems to work. I will update the PR.

atscott: We could either modify the `auth.guard` to only store the url segment for the primary outlet:
`this.authService.redirectUrl = this.router.parseUrl(url).root.children['primary'].toString();`
or we can modify the login component to only retrieve the url segment for the primary outlet stored in the redirectUrl:
`        const redirect = this.authService.redirectUrl ? this.router.parseUrl(this.authService.redirectUrl).root.children['primary'].toString() : '/admin';`

@gkalpak - what do you think here? It's getting a little hairy for a simple example. Maybe we should remove the part where authService stores a redirectUrl and just land them on '/admin' when they log in no matter where they were trying to go.

gkalpak: Thx for the info, @atscott. It is indeed getting complicated and (while not ideal) I agree it would be better to just redirect them to `/admin`.

sonukapoor: Thanks @atscott. @gkalpak, I have made the changes to the code only and added a comment. I did not change the documentation for routing now. I think that that would be a somewhat bigger change.

aio/content/examples/router/src/app/auth/login/login.component.ts
```diff
@@ -26,9 +25,9 @@ export class LoginComponent {
     this.authService.login().subscribe(() => {
       this.setMessage();
       if (this.authService.isLoggedIn) {
-        // Get the redirect URL from our auth service
-        // If no redirect has been set, use the default
-        let redirect = this.authService.redirectUrl ? this.router.parseUrl(this.authService.redirectUrl) : '/admin';
+        // Usually you would use the redirect URL from the auth service.
```
aikidave: Is this information reflected in the user documentation that accompanies the example? I think that's important.

Having code comments is great, but it's better when it supports/compliments the content in the user doc. There are a number of reasons for this, but probably the most important one is the simplest: We never know what users are going to read, or when, or even where! (Makes writing documentation interesting...)

gkalpak (changes requested):
The code changes lgtm, but as @aikidave said we should update the docs accordingly.

aio/content/examples/router/src/app/auth/login/login.component.ts
sonukapoor: @aikidave Makes sense to me as well.  I have added something very simple. Take a look.

aio/content/guide/router.md
```diff
@@ -3302,7 +3302,13 @@ Although it doesn't actually log in, it has what you need for this discussion.
 It has an `isLoggedIn` flag to tell you whether the user is authenticated.
 Its `login` method simulates an API call to an external service by returning an
 observable that resolves successfully after a short pause.
-The `redirectUrl` property will store the attempted URL so you can navigate to it after authenticating.
+The `redirectUrl` property will store the attempted URL so you could navigate to it after authenticating.
+
+<div class="alert is-helpful">
+
+To keep the stackblitz example simple, we will always redirect an un-authenticated user to `/admin`.
```
aikidave: perhaps:

"To keep things simple, this example redirects unauthenticated users to `/admin`."

aio/content/guide/router.md
```diff
@@ -3302,7 +3302,13 @@ Although it doesn't actually log in, it has what you need for this discussion.
 It has an `isLoggedIn` flag to tell you whether the user is authenticated.
 Its `login` method simulates an API call to an external service by returning an
 observable that resolves successfully after a short pause.
-The `redirectUrl` property will store the attempted URL so you can navigate to it after authenticating.
+The `redirectUrl` property will store the attempted URL so you could navigate to it after authenticating.
```
aikidave: Perhaps:

The `redirectUrl` property stores the URL that the user wanted to access so you can navigate to it after authentication.

Side note: In general, content is clearer if you use present tense ("stores" in this case) instead of future tense ("will store").

aikithoughts (commented):
Thanks for adding some content! I've made a few suggestions.

aio/content/guide/router.md
sonukapoor: That sounds better. I will make the change tonight.

sonukapoor: That sounds better. I will make the change tonight.

aio/content/guide/router.md
sonukapoor: But the code in the doc does not redirect the user to /admin. It’s only the stack blitz. I think it’s beneficial to keep that code in the doc.

gkalpak: The code in the doc and the StackBlitz example should both be generated based on the same example? Are you saying that is not the case? 😱 
We should definitely not have different code in StackBlitz vs the guide.

sonukapoor: @gkalpak Actually, the code in the docs is the same as in the example - my mistake. I have made the changes.

gkalpak (approved):
The code changes lgtm. (@aikidave and @kapunahelewong are more qualified to speak about the changes in the guide.)
Thx, @sonukapoor :+1:

aikithoughts (approved):
Changes look good. Thank  you, @sonukapoor!

kapunahelewong (approved):

aio/content/examples/router/src/app/auth/login/login.component.ts
atscott: FYI - this issue is now tracked by https://github.com/angular/angular/issues/35414.

sonukapoor:
Could you please add the `merge` labels?

kapunahelewong:
Per offline conversation with @sonukapoor, we're going to hold until @atscott gets a chance to look it over since he'd been involved in the development of this PR.

atscott (approved):
LGTM. Thanks for updating this.

PR #35177, opened 2020-02-06
refactor(router): Allow routerLinkWithHref directive to be extended

jmandreslopez:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

This will allow anybody that wants to extend the routerLink directive to override or modify any of the properties. 

Issue Number: N/A

## What is the new behavior?
Right now I'm trying to create a new feature that detects if the routerLink contains an external URL and router takes another path because of that. I can't continue because all the important methods are private to prevent the extended class to override them.

- [ ] Yes
- [X] No

### DIFF
--- a/packages/router/src/directives/router_link.ts
+++ b/packages/router/src/directives/router_link.ts
@@ -126,12 +126,12 @@ export class RouterLink {
   // TODO(issue/24571): remove '!'.
   @Input() replaceUrl !: boolean;
   @Input() state?: {[k: string]: any};
-  private commands: any[] = [];
+  protected commands: any[] = [];
   // TODO(issue/24571): remove '!'.
-  private preserve !: boolean;
+  protected preserve !: boolean;
 
   constructor(
-      private router: Router, private route: ActivatedRoute,
+      protected router: Router, protected route: ActivatedRoute,
       @Attribute('tabindex') tabIndex: string, renderer: Renderer2, el: ElementRef) {
     if (tabIndex == null) {
       renderer.setAttribute(el.nativeElement, 'tabindex', '0');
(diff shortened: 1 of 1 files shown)


jmandreslopez:
@googlebot I signed it!

atscott:
Hi @jmandreslopez, we discourage people from extend our classes unless we documented that the class is meant to be subclassed:

extending any of our classes 

> unless the support for this is specifically documented in the API docs 
> from: https://github.com/angular/angular/blob/master/docs/PUBLIC_API.md

Have you considered using router guards instead?

jmandreslopez:
Hi @atscott, yes I've already tried Router Guards and that won't work on my project. Basically what I'm trying to do is to make the router smart enough to be able to detect if the href/routerLink belongs to an internal route or it is an external link, and handle them differently based on that. I managed to extend the RouterLinkWithHref to accomplish that but because that class listens to navigation events the href gets modified every time I go to another route. What I want with this PR is to be able to override the private methods in order to stop the update href functionality if it detect that it is an external link. If you know a better implementation I'm open to it. Thanks for checking this matter btw 😀

atscott:
Hi @josenicomaia, can you help me understand a bit more about why some of these solutions don't work for your project? 

There are two options described in the answers here (guards and resolvers):
https://stackoverflow.com/questions/40150393/how-to-redirect-to-an-external-url-from-angular2-route-without-using-component

This post describes in a bit more detail how to use the guard with a directive to help make things a bit easier: https://medium.com/@adrianfaciu/using-the-angular-router-to-navigate-to-external-links-15cc585b7b88

If neither of these work because you want the link to act outside the router completely, why not use `href` directly instead of `routerLink`?

jmandreslopez:
Hi @atscott, I can't use `href` directly because I'm handling dynamic links and some of them are internal routes coming directly from the API and others are external links. I wanted to have a flexible directive that could identify those links and either trigger the router to navigate to that route internally or, in case of an external route, to just override the window url and force a redirect. 

The link you recommended made me think that I was overthinking my code and I manage to accomplish what I mentioned using another directly (not extending `routerLink`) and whenever an internal route gets detected I'm just triggering the router using the `navigate` method, otherwise a forced redirect makes the trick. 

Thanks for taking the time and helping me find a better alternative. I will close this PR as it is not valid anymore.

PR #35185, opened 2020-02-06
build(docs-infra): upgrade cli command docs sources to d452f0873

gkalpak:
Updating [angular#master](https://github.com/angular/angular/tree/master) from [cli-builds#master](https://github.com/angular/cli-builds/tree/master).

##
Relevant changes in [commit range](https://github.com/angular/cli-builds/compare/94d07909c...d452f0873):

**Modified**
- help/generate.json

##


IgorMinar (approved):
lgtm but shouldn't this be master&patch?

gkalpak:
We use different SHAs on master and patch. On master, we use the latest SHA from `angular/cli-builds#master` and on patch we use the latest SHA from the highest `angular/cli-builds` branch that has the same major as `angular/angular` patch. So, for our 9.x.x patch branch, we use a SHA from the highest 9.y.y `cli-builds` branch.

PR #35191, opened 2020-02-06
fix(ngcc): correctly detect emitted TS helpers in ES5

gkalpak:
In ES5 code, TypeScript requires certain helpers (such as `__spreadArrays()`) to be able to support ES2015+ features. These helpers can be either imported from `tslib` (by setting the `importHelpers` TS compiler option to `true`) or emitted inline (by setting the `importHelpers` and `noEmitHelpers` TS compiler options to `false`, which is the default value for both).

Ngtsc's `StaticInterpreter` (which is also used during ngcc processing) is able to statically evaluate some of these helpers (currently `__assing()`, `__spread()` and `__spreadArrays()`), as long as `ReflectionHost#getDefinitionOfFunction()` correctly detects the declaration of the helper. For this to happen, the left-hand size of the corresponding call expression (i.e. `__spread(...)` or `tslib.__spread(...)`) must be evaluated as a function declaration for `getDefinitionOfFunction()` to be called with.

In the case of imported helpers, the `tslib.__someHelper` expression was resolved to a function declaration of the form `export declare function __someHelper(...args: any[][]): any[];`, which allows `getDefinitionOfFunction()` to correctly map it to a TS helper.

In contrast, in the case of emitted helpers (and regardless of the module format: `CommonJS`, `ESNext`, `UMD`, etc.)), the `__someHelper` identifier was resolved to a variable declaration of the form `var __someHelper = (this && this.__someHelper) || function () { ... }`, which upon further evaluation was categorized as a `DynamicValue` (prohibiting further evaluation by the `getDefinitionOfFunction()`).

As a result of the above, emitted TypeScript helpers were not evaluated in ES5 code.

---
This commit changes the detection of TS helpers to leverage the existing `KnownFn` feature (previously only used for built-in functions). `Esm5ReflectionHost`'s is changed to always return `KnownDeclaration`s for TS helpers, both imported (`getExportsOfModule()`) as well as emitted (`getDeclarationOfIdentifier()`).

The `KnownDeclaration`s are then mapped to `KnownFn`s in `StaticInterpreter`, allowing it to statically evaluate call expressions involving any kind of TS helpers.

Jira issue: [FW-1689][1]

---
_I tested it on `ngcc-validation` and it does fix the `mobx-angular@3` failure (without breaking anything else 😛):_
- _angular/ngcc-validation#838_
- _[CI run][2]_

[1]: https://angular-team.atlassian.net/browse/FW-1689
[2]: https://circleci.com/workflow-run/0cf476fd-319e-4821-93f9-2f3716761c22

[FW-1689]: https://angular-team.atlassian.net/browse/FW-1689

### DIFF
--- a/packages/compiler-cli/ngcc/src/host/esm5_host.ts
+++ b/packages/compiler-cli/ngcc/src/host/esm5_host.ts
@@ -274,8 +274,8 @@ export class Esm5ReflectionHost extends Esm2015ReflectionHost {
       |null {
     const superDeclaration = super.getDeclarationOfSymbol(symbol, originalId);
 
-    if ((superDeclaration !== null) && (superDeclaration.node !== null) &&
-        (superDeclaration.known === null)) {
+    if (superDeclaration !== null && superDeclaration.node !== null &&
+        superDeclaration.known === null) {
       superDeclaration.known = getTsHelperFnFromDeclaration(superDeclaration.node);
     }
 
--- a/packages/compiler-cli/ngcc/src/host/esm5_host.ts
+++ b/packages/compiler-cli/ngcc/src/host/esm5_host.ts
@@ -8,8 +8,8 @@
 
 import * as ts from 'typescript';
 
-import {ClassDeclaration, ClassMember, ClassMemberKind, Declaration, Decorator, FunctionDefinition, Parameter, TsHelperFn, isNamedVariableDeclaration, reflectObjectLiteral} from '../../../src/ngtsc/reflection';
-import {getNameText, hasNameIdentifier, stripDollarSuffix} from '../utils';
+import {ClassDeclaration, ClassMember, ClassMemberKind, Declaration, Decorator, FunctionDefinition, Parameter, isNamedVariableDeclaration, reflectObjectLiteral} from '../../../src/ngtsc/reflection';
+import {getNameText, getTsHelperFnFromDeclaration, hasNameIdentifier} from '../utils';
 
 import {Esm2015ReflectionHost, ParamInfo, getPropertyValueFromSymbol, isAssignment, isAssignmentStatement} from './esm2015_host';
 import {NgccClassSymbol} from './ngcc_host';
@@ -118,6 +118,7 @@ export class Esm5ReflectionHost extends Esm2015ReflectionHost {
     // Return the statement before the IIFE return statement
     return iifeBody.statements[returnStatementIndex - 1];
   }
+
   /**
    * In ES5, the implementation of a class is a function expression that is hidden inside an IIFE,
    * whose value is assigned to a variable (which represents the class to the rest of the program).
@@ -245,23 +246,7 @@ export class Esm5ReflectionHost extends Esm2015ReflectionHost {
    */
   getDefinitionOfFunction(node: ts.Node): FunctionDefinition|null {
     if (!ts.isFunctionDeclaration(node) && !ts.isMethodDeclaration(node) &&
-        !ts.isFunctionExpression(node) && !ts.isVariableDeclaration(node)) {
-      return null;
-    }
-
-    const tsHelperFn = getTsHelperFn(node);
-    if (tsHelperFn !== null) {
-      return {
-        node,
-        body: null,
-        helper: tsHelperFn,
-        parameters: [],
-      };
-    }
-
-    // If the node was not identified to be a TypeScript helper, a variable declaration at this
-    // point cannot be resolved as a function.
-    if (ts.isVariableDeclaration(node)) {
+        !ts.isFunctionExpression(node)) {
       return null;
     }
 
@@ -276,11 +261,26 @@ export class Esm5ReflectionHost extends Esm2015ReflectionHost {
       return !lookingForParamInitializers;
     });
 
-    return {node, body: statements || null, helper: null, parameters};
+    return {node, body: statements || null, parameters};
   }
 
 
   ///////////// Protected Helpers /////////////
+  /**
+   * Resolve a `ts.Symbol` to its declaration and detect whether it corresponds with a known
+   * TypeScript helper function.
+   */
+  protected getDeclarationOfSymbol(symbol: ts.Symbol, originalId: ts.Identifier|null): Declaration
+      |null {
+    const superDeclaration = super.getDeclarationOfSymbol(symbol, originalId);
+
+    if ((superDeclaration !== null) && (superDeclaration.node !== null) &&
+        (superDeclaration.known === null)) {
+      superDeclaration.known = getTsHelperFnFromDeclaration(superDeclaration.node);
+    }
+
+    return superDeclaration;
+  }
 
   /**
    * Get the inner function declaration of an ES5-style class.
@@ -643,28 +643,6 @@ function getReturnStatement(declaration: ts.Expression | undefined): ts.ReturnSt
 
 function reflectArrayElement(element: ts.Expression) {
   return ts.isObjectLiteralExpression(element) ? reflectObjectLiteral(element) : null;
-}
-
-/**
- * Inspects a function declaration to determine if it corresponds with a TypeScript helper function,
- * returning its kind if so or null if the declaration does not seem to correspond with such a
- * helper.
- */
-function getTsHelperFn(node: ts.NamedDeclaration): TsHelperFn|null {
-  const name = node.name !== undefined && ts.isIdentifier(node.name) ?
-      stripDollarSuffix(node.name.text) :
-      null;
-
-  switch (name) {
-    case '__assign':
-      return TsHelperFn.Assign;
-    case '__spread':
-      return TsHelperFn.Spread;
-    case '__spreadArrays':
-      return TsHelperFn.SpreadArrays;
-    default:
-      return null;
-  }
 }
 
 /**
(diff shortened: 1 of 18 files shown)


petebacondarwin (commented):
typo in `fix(ngcc): correctly detect emitted TS helpers in ES5` commit:

`__assing` -> `__assign`

packages/compiler-cli/src/ngtsc/partial_evaluator/test/evaluator_spec.ts
```diff
@@ -665,36 +805,52 @@ runInEachFileSystem(() => {
...
+        };
+      }
+
+      return superDeclaration;
     }
   }
 
-  function getTsHelperFn(node: ts.FunctionDeclaration): TsHelperFn|null {
-    const name = node.name !== undefined && ts.isIdentifier(node.name) && node.name.text;
+  function getTsHelperFn(node: ts.Declaration): KnownDeclaration|null {
```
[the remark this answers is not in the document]
petebacondarwin: Perhaps move both helpers into `packages/compiler-cli/src/ngtsc/reflection/src/host.ts`?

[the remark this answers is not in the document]
gkalpak: Actually, I'm thinking about keeping this one (and possibly simplifying it), because we don't need the actual implementation here. I would rather keep the "declaration node"-->"TS helper" mapping inside `ngcc/`.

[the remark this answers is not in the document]
petebacondarwin: OK

packages/compiler-cli/ngcc/src/host/esm5_host.ts
```diff
@@ -650,18 +659,24 @@ function reflectArrayElement(element: ts.Expression) {
  * returning its kind if so or null if the declaration does not seem to correspond with such a
  * helper.
  */
-function getTsHelperFn(node: ts.NamedDeclaration): TsHelperFn|null {
-  const name = node.name !== undefined && ts.isIdentifier(node.name) ?
-      stripDollarSuffix(node.name.text) :
-      null;
+export function getTsHelperFn(node: ts.Declaration): KnownDeclaration|null {
```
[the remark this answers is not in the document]
gkalpak: Done.

packages/compiler-cli/ngcc/src/host/esm5_host.ts
```diff
@@ -200,6 +216,15 @@ export class Esm5ReflectionHost extends Esm2015ReflectionHost {
...
     }
 
+    const tsHelperFn = getTsHelperFn(superDeclaration.node);
+    if (tsHelperFn !== null) {
+      return {
+        known: tsHelperFn,
+        node: id,
+        viaModule: null,
+      };
+    }
```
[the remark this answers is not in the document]
gkalpak: Done.

gkalpak:
Rebased on master and addressed initial feedback. PTAL.

packages/compiler-cli/ngcc/test/host/commonjs_host_spec.ts
```diff
@@ -1744,6 +1744,239 @@ exports.ExternalModule = ExternalModule;
...
+                    (node.initializer as ts.CallExpression).expression as ts.Identifier;
+                const helperDeclaration = host.getDeclarationOfIdentifier(helperIdentifier);
+
+                expect(helperDeclaration).toEqual({
+                  known: knownAs,
+                  node: getDeclaration(
+                      bundle.program, file.name, helperName, ts.isFunctionDeclaration),
+                  viaModule: null,
+                });
+              };
```
petebacondarwin: This code pattern is used over and over. I wonder if we could DRY it up with something like `const testForHelper = createTestForHelper(host, bundle, file, ts.isFunctionDeclaration);`?

packages/compiler-cli/ngcc/test/utils_spec.ts
```diff
@@ -50,6 +52,121 @@ describe('FactoryMap', () => {
   });
 });
 
+describe('getTsHelperFnFromDeclaration()', () => {
+  const createFunctionDeclaration = (fnName?: string) => ts.createFunctionDeclaration(
+      undefined, undefined, undefined, fnName, undefined, [], undefined, undefined);
+  const createVariableDeclaration = (varName: string) =>
+      ts.createVariableDeclaration(varName, undefined, undefined);
+
+  it('should recongize the `__assign` helper as function declaration', () => {
```
petebacondarwin:
```suggestion
  it('should recognize the `__assign` helper as function declaration', () => {
```

And throughout the file :-)

packages/compiler-cli/src/ngtsc/partial_evaluator/test/evaluator_spec.ts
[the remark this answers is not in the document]
petebacondarwin: Actually if ngtsc's code never uses this stuff then why are we even testing it here?
Or if there is some code path in ngtsc that needs to be exercised in these tests, why both with a complex function that returns different TS helpers? Why not just support one of them in these tests?

petebacondarwin (approved):
Loving this refactoring @gkalpak - nice work.

packages/compiler-cli/ngcc/test/host/commonjs_host_spec.ts
gkalpak: The problem is that there are many variations and the helpers would be a little complicated.
I pushed a fixup commit with one helper per spec file (having helpers that work across format would be too complicated and hurt readability imo).

I am fine with or without that last commit. LMK what you think.

petebacondarwin (approved):
I think this last commit is just right. The tests are much easier to read now and the amount of duplication is significantly reduced.

packages/compiler-cli/src/ngtsc/partial_evaluator/test/evaluator_spec.ts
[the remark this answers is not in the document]
gkalpak: As discussed "offline" these tests verify that if a `ReflectionHost` implements the TS helper detection in the appropriate methods, then `StaticInterpreter` will be able to statically evaluate them. 

This does not guarantee that the `NgccReflectionHost`s will implement TS helper detection in the correct way (since we are not using the actual `NgccReflectionHost`s in the tests). So, the tests do not add much value, but they don't hurt either.

(Note, we integration tests that verify this works end-to-end (using the actual `NgccReflectionHost`s) in [integration/ngcc_spec.ts][1].)

[1]: https://github.com/angular/angular/blob/d6bc63fa3/packages/compiler-cli/ngcc/test/integration/ngcc_spec.ts

packages/compiler-cli/ngcc/src/host/esm5_host.ts
```diff
@@ -276,11 +261,26 @@ export class Esm5ReflectionHost extends Esm2015ReflectionHost {
...
+  /**
+   * Resolve a `ts.Symbol` to its declaration and detect whether it corresponds with a known
+   * TypeScript helper function.
+   */
+  protected getDeclarationOfSymbol(symbol: ts.Symbol, originalId: ts.Identifier|null): Declaration
+      |null {
+    const superDeclaration = super.getDeclarationOfSymbol(symbol, originalId);
+
+    if ((superDeclaration !== null) && (superDeclaration.node !== null) &&
+        (superDeclaration.known === null)) {
```
JoostK: nit: why the braces around the comparisons? 🙊

packages/compiler-cli/ngcc/src/utils.ts
```diff
@@ -132,6 +133,40 @@ export function resolveFileWithPostfixes(
...
+/**
+ * Determine whether a function declaration corresponds with a TypeScript helper function, returning
+ * its kind if so or null if the declaration does not seem to correspond with such a helper.
+ */
+export function getTsHelperFnFromDeclaration(decl: ts.Declaration): KnownDeclaration|null {
+  if (!ts.isFunctionDeclaration(decl) && !ts.isVariableDeclaration(decl)) {
+    return null;
+  }
+
+  if ((decl.name === undefined) || !ts.isIdentifier(decl.name)) {
```
JoostK: Here too, I don't recall seeing such braces in other areas of the code.

packages/compiler-cli/src/ngtsc/partial_evaluator/test/evaluator_spec.ts
```diff
@@ -538,69 +538,209 @@ runInEachFileSystem(() => {
...
+      it('should evaluate `__assign()` (named import)', () => {
+        const map: Map<string, boolean> = evaluateExpression(
+            `
+              import {__assign} from 'tslib';
+              const a = {a: true};
+              const b = {b: true};
+            `,
+            '__assign(a, b)');
+
+        expect([...map]).toEqual([
```
JoostK: Clever!

JoostK (approved):
Neat! I like how the `BuiltinFn` vs `TsHelperFn` concepts have been unified into `KnownFn`.

packages/compiler-cli/ngcc/src/host/esm5_host.ts
gkalpak: It's a personal convention - I always add them automatically 😇
Let me remove them.

packages/compiler-cli/ngcc/src/utils.ts
gkalpak: Removed.

alxhub:
[Presubmit](http://test/OCL:295802858:BASE:295802915:1582059817262:ce8b54a)

PR #35193, opened 2020-02-06
fix(ivy): repeat template guards to narrow types in event handlers

### LINKED ISSUE #35073: Ivy template type checking doesn't narrow types in output positions
# 🐞 bug report

### Scope
@angular/compiler-cli
Specifically: Ivy template type checking

### Is this a regression?
No.

### Description / Reproduction
Given a template such as
```
<some-component
    *ngIf="myVar"
    [anInput]="myVar.otherVal"
    (anOutput)="myFunc(myVar)"
></some-component>
```
And the owning component class contains
<pre><code>    public myVar?: {otherVal: number};
    public myFunc(aDefinedVal: {otherVal: number}) {}
</code></pre>

The type checker will correctly know that `myVar.otherVal` is a safe operation under `strictNullChecks` because of the `*ngIf`, but the checker will incorrectly throw an error on the `myFunc(myVar)`, claiming that `myVar` is of type `{otherVal: number} | undefined` which is not assignable to type `{otherVal: number}` (diagnostic code `NG2345`).

## 🌍  Environment
"@angular/common": "9.0.0-rc.11",
"@angular/compiler": "9.0.0-rc.11",
"@angular/compiler-cli": "9.0.0-rc.11",
"@angular/core": "9.0.0-rc.11",

Compiled using Bazel and `--define=angular_ivy_enabled=True`

JoostK:
In Ivy's template type checker, event bindings are checked in a closure
to allow for accurate type inference of the `$event` parameter. Because
of the closure, any narrowing effects of template guards will no longer
be in effect when checking the event binding, as TypeScript assumes that
the guard outside of the closure may no longer be true once the closure
is invoked. For more information on TypeScript's Control Flow Analysis,
please refer to https://github.com/microsoft/TypeScript/issues/9998.

In Angular templates, it is known that an event binding can only be
executed when the view it occurs in is currently rendered, hence the
corresponding template guard is known to hold during the invocation of
an event handler closure. As such, it is desirable that any narrowing
effects from template guards are still in effect within the event
handler closure.

This commit tweaks the generated Type-Check Block (TCB) to repeat all
template guards within an event handler closure. This achieves the
narrowing effect of the guards even within the closure.

Fixes #35073

### DIFF
--- a/packages/compiler-cli/src/ngtsc/typecheck/src/type_check_block.ts
+++ b/packages/compiler-cli/src/ngtsc/typecheck/src/type_check_block.ts
@@ -51,7 +51,7 @@ export function generateTypeCheckBlock(
     oobRecorder: OutOfBandDiagnosticRecorder): ts.FunctionDeclaration {
   const tcb = new Context(
       env, domSchemaChecker, oobRecorder, meta.id, meta.boundTarget, meta.pipes, meta.schemas);
-  const scope = Scope.forNodes(tcb, null, tcb.boundTarget.target.template !);
+  const scope = Scope.forNodes(tcb, null, tcb.boundTarget.target.template !, /* guard */ null);
   const ctxRawType = env.referenceType(ref);
   if (!ts.isTypeReferenceNode(ctxRawType)) {
     throw new Error(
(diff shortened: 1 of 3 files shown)


alxhub (approved):

kara:
[presubmit](http://test/OCL:293858751:BASE:293858790:1581104143917:8607ebcc)

PR #35200, opened 2020-02-06
fix(language-service): Suggest ? and ! operator on nullable receiver

### LINKED ISSUE #301: Non-null check in ngIf not respected for variables
I'm using VSCode version 1.29.1 and Angular CLI version 7.0.4
I have the following issue

In my typescript component file I have declared an object as the following
```
object: Type | undefined
```

In my HTML I wanted to use that object as the following
```
<div *ngIf="object">
	{{ object.someField }}
</div>
```

but I'm having error highlights with the following message
```
[Angular] The expression might be null
```

### LINKED ISSUE #589: Truthy value check, but red highlighting for undefined

kyliau:
Under strict mode, the language service fails to typecheck nullable
symbols that have already been verified to be non-null.

This generates incorrect (false positive) and confusing diagnostics
for users.

To work around this issue in the short term (before properly fixing it in Ivy), 
this commit changes the diagnostic message from an error to a suggestion, 
and prompts users to use the safe navigation operator (?) or 
non-null assertion operator (!).

For example, instead of

```typescript
{{ optional && optional.toString() }}
```

the following is cleaner:

```typescript
{{ optional?.toString() }}
{{ optional!.toString() }}
```

Note that with this change, users who legitimately make a typo in their
code will no longer see an error. I think this is a reasonable trade-off, since
false positive is worse than false negative. However, if users follow
the suggestion, add ? or ! to their code, then the error will be surfaced.

References:

1. Safe navigation operator (?)
   https://angular.io/guide/template-syntax#the-safe-navigation-operator----and-null-property-paths
2. Non-null assertion operator (!)
   https://angular.io/guide/template-syntax#the-non-null-assertion-operator---

Thank you @andrius-pra for the suggestion =)

PR closes https://github.com/angular/angular/pull/35070
PR closes https://github.com/angular/vscode-ng-language-service/issues/589
PR closes https://github.com/angular/vscode-ng-language-service/issues/301

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/language-service/src/expression_type.ts
+++ b/packages/language-service/src/expression_type.ts
@@ -366,22 +366,19 @@ export class AstType implements AstVisitor {
     if (this.isAny(receiverType)) {
       return this.anyType;
     }
-
-    // The type of a method is the selected methods result type.
-    const method = receiverType.members().get(ast.name);
-    if (!method) {
-      this.reportDiagnostic(`Unknown method '${ast.name}'`, ast);
-      return this.anyType;
-    }
-    if (!method.type) {
+    const methodType = this.resolvePropertyRead(receiverType, ast);
+    if (!methodType) {
       this.reportDiagnostic(`Could not find a type for '${ast.name}'`, ast);
       return this.anyType;
     }
-    if (!method.type.callable) {
+    if (this.isAny(methodType)) {
+      return this.anyType;
+    }
+    if (!methodType.callable) {
       this.reportDiagnostic(`Member '${ast.name}' is not callable`, ast);
       return this.anyType;
     }
-    const signature = method.type.selectSignature(ast.args.map(arg => this.getType(arg)));
+    const signature = methodType.selectSignature(ast.args.map(arg => this.getType(arg)));
     if (!signature) {
       this.reportDiagnostic(`Unable to resolve signature for call of method ${ast.name}`, ast);
       return this.anyType;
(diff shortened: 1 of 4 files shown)


packages/language-service/src/expression_type.ts
```diff
@@ -370,7 +370,13 @@ export class AstType implements AstVisitor {
     // The type of a method is the selected methods result type.
     const method = receiverType.members().get(ast.name);
     if (!method) {
-      this.reportDiagnostic(`Unknown method '${ast.name}'`, ast);
+      if (receiverType.nullable) {
+        this.reportDiagnostic(
+            `Consider using the safe navigation operator (?) or non-null assertion operator (!) to access '${ast.name}'.`,
```
ayazhafiz: Could we provide an example based off the receiver and ast name? It would be very nice if we could suggest `(myobj?.value)` rather than just `(?)`

packages/language-service/src/expression_type.ts
```diff
@@ -370,7 +370,13 @@ export class AstType implements AstVisitor {
     // The type of a method is the selected methods result type.
     const method = receiverType.members().get(ast.name);
     if (!method) {
-      this.reportDiagnostic(`Unknown method '${ast.name}'`, ast);
+      if (receiverType.nullable) {
```
ayazhafiz: ohhhh nice. this fixes the actual typo case. sweet!

ayazhafiz (approved):
just one comment

packages/language-service/src/expression_type.ts
kyliau: Good idea! I had to narrow down the type of `ast.receiver` to access the receiver name.
In the future we could implement [quick fixes](https://code.visualstudio.com/docs/languages/typescript#_quick-fixes) for suggestions like this.

packages/language-service/src/expression_type.ts
```diff
@@ -366,11 +366,18 @@ export class AstType implements AstVisitor {
     if (this.isAny(receiverType)) {
       return this.anyType;
     }
-
+    const receiverName = (ast.receiver as PropertyRead).name;
```
ayazhafiz: Can we do a sanity check and bail out in case this isn’t a property read? same on line 403.

ayazhafiz (changes requested):

packages/language-service/src/expression_type.ts
kyliau: I went one step further, there's really no need to do the same check in multiple places. resolveMethodCall should call resolvePropertyRead, then verify that the type is callable. The code is a lot simpler now ;)

kyliau:
@ayazhafiz PTAL, thanks :)

packages/language-service/src/expression_type.ts
```diff
@@ -393,34 +390,32 @@ export class AstType implements AstVisitor {
...
-        return this.reportDiagnostic(`The expression might be null`, ast.receiver);
+      if (receiverType.name === '$implicit') {
+        this.reportDiagnostic(
+            `Identifier '${ast.name}' is not defined. ` +
+                `The component declaration, template variable declarations, and element references do not contain such a member`,
+            ast);
+      } else if (receiverType.nullable && ast.receiver instanceof PropertyRead) {
+        this.reportDiagnostic(
+            `Consider using the safe navigation operator (${ast.receiver.name}?.${ast.name}) ` +
+                `or non-null assertion operator (${ast.receiver.name}!.${ast.name}).`,
```
ayazhafiz: Sorry, just one more nit -- when we report this diagnostic, it may not be immediately clear to the user why they should use the safe navigation/assertion operator. Can we add a sentence at the beginning saying that the receiver might be null? e.g.

```typescript
`${ast.receiver.name} may not be defined. Consider...`
```

ayazhafiz (changes requested):

packages/language-service/src/expression_type.ts
kyliau: Good idea, done!

ayazhafiz (approved):

PR #35211, opened 2020-02-07
fix(elements): schematics fails with schema.json not found error

### LINKED ISSUE #35154: `ng add @angular/elements@next` fails with schema.json not found error
# 🐞 bug report

### Affected Package
The issue is caused by package @angular/elements

### Is this a regression?

Yes. Doing `ng add @angular/elements` works fine.

### Description

Creating a new workspace project using angular 9 RC 12. Everything installing fine, but when doing `ng add @angular/elements@next` I get: 

```
`An unhandled exception occurred: Path "<my_dir>/node_modules/@angular/elements/schematics/ng-add/schema.json" does not exist.
See "/private/var/folders/9c/scyzyh01349gnx6qgkp_twzh0000gn/T/ng-GPnlhg/angular-errors.log" for further details.
```

Log shows: 

```
[error] Error: Path "/Users/hisham/src/ess-app/frontend/node_modules/@angular/elements/schematics/ng-add/schema.json" does not exist.
    at readFile (/Users/hisham/src/ess-app/frontend/node_modules/@angular-devkit/schematics/tools/file-system-utility.js:17:15)
    at Object.readJsonFile (/Users/hisham/src/ess-app/frontend/node_modules/@angular-devkit/schematics/tools/file-system-utility.js:45:29)
    at NodeModulesEngineHost.createSchematicDescription (/Users/hisham/src/ess-app/frontend/node_modules/@angular-devkit/schematics/tools/file-system-engine-host-base.js:183:48)
    at SchematicEngine.createSchematic (/Users/hisham/src/ess-app/frontend/node_modules/@angular-devkit/schematics/src/engine/engine.js:219:38)
    at CollectionImpl.createSchematic (/Users/hisham/src/ess-app/frontend/node_modules/@angular-devkit/schematics/src/engine/engine.js:69:29)
    at AddCommand.getSchematic (/Users/hisham/src/ess-app/frontend/node_modules/@angular/cli/models/schematic-command.js:134:27)
    at AddCommand.runSchematic (/Users/hisham/src/ess-app/frontend/node_modules/@angular/cli/models/schematic-command.js:296:32)
    at AddCommand.executeSchematic (/Users/hisham/src/ess-app/frontend/node_modules/@angular/cli/commands/add-impl.js:176:31)
    at AddCommand.run (/Users/hisham/src/ess-app/frontend/node_modules/@angular/cli/commands/add-impl.js:60:29)
    at async AddCommand.validateAndRun (/Users/hisham/src/ess-app/frontend/node_modules/@angular/cli/models/command.js:134:28)
    at async Object.runCommand (/Users/hisham/src/ess-app/frontend/node_modules/@angular/cli/models/command-runner.js:201:24)
    at async default_1 (/Users/hisham/src/ess-app/frontend/node_modules/@angular/cli/lib/cli/index.js:62:31)
```

**Anything else relevant?**

OSX 10.5.12, npm 6.13.2, node 13.2.0

alan-agius4:
Fixes #35154


integration/ng_elements-schematics/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "elements-schematics",
+  "version": "1.0.0",
+  "main": "index.js",
```
gkalpak:
```suggestion
```
(to avoid confusion)

integration/ng_elements-schematics/test.sh
```diff
@@ -0,0 +1,40 @@
...
+
+
+function test() {
+  # Set up
+  ng version
+  rm -rf demo
+  # Create project
+  ng new demo --skip-git --skip-install --style=css
+  cd demo
+  ng add @angular/elements
```
gkalpak: Will this not install the dependencies?

integration/ng_elements-schematics/package.json
```diff
@@ -0,0 +1,13 @@
...
+  "license": "MIT",
+  "devDependencies": {
+    "@angular/elements": "file:../../dist/packages-dist/elements",
+    "@angular/cli": "file:../../node_modules/@angular/cli"
```
gkalpak: Are these dependencies used at all?
(If not, do we even need a `package.json`/`yarn.lock`?)

integration/ng_elements-schematics/test.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
```
gkalpak: We try to avoid adding more bash scripts into the repo, because they are harder to understand and work with across OSes. We favor JS scripts (with `ShellJS` if needed) instead.

gkalpak (changes requested):
Thx for adding the integration test :100:

Can you explain somewhere in the integration test (e.g. in a `README.md` or in a comment in the test script) what this test is testing. (I understand from the context that it verifies the `ng-add` schematic works correctly, but it would be hard to infer by only looking at the integration test.

integration/ng_elements-schematics/test.sh
alan-agius4: No, because it's already installed. `ng add` will use the local version

integration/ng_elements-schematics/package.json
alan-agius4: Yes they are used.

1) CLI: because we need the local `ng`
2) Elements to ensure that we use the package from `packages-dist/elements` instead of `npm`

integration/ng_elements-schematics/test.sh
alan-agius4: I wasn't aware of that on the move to node script. But I totally dig that!

integration/ng_elements_schematics/test.ts
```diff
@@ -0,0 +1,49 @@
...
+  'animations',
+  'common',
+  'compiler',
+  'forms',
+  'platform-browser',
+  'platform-browser-dynamic',
+  'router',
+  'elements',
+  'compiler-cli',
+  'language-service',
```
gkalpak: Do we need all these packages (forms, router, etc.)? Also, where is `core`? 🤔
How about having just `common`, `core`, `elements`, `platform-browser`, `platform-browser-dynamic`?

integration/ng_elements_schematics/test.ts
```diff
@@ -0,0 +1,49 @@
...
+].map(p => `"@angular/${p}@file:${pwd}/../../dist/packages-dist/${p}"`);
```
gkalpak: Wouldn't it be safer to use `__driname`? `pwd` depends on where the script is run from, so these paths could be wrong.

integration/ng_elements_schematics/test.ts
```diff
@@ -0,0 +1,49 @@
...
+  ...ngPackages,
+  ...nodePackages,
+].join(' ');
+
+// Clean up previously run test
+shx.rm('-rf', 'demo');
+
+// Set up demo project.
+shx.exec('ng version');
+shx.exec('ng new demo --skip-git --skip-install --style=css');
```
gkalpak: How come this does not ask about things like routing pn CI? (It does for me when run locally.)

integration/ng_elements_schematics/test.ts
```diff
@@ -0,0 +1,49 @@
...
+shx.cd('demo');
```
gkalpak: Here too, consider using `__dirname`, so that this script does not need to be run from a specific directory.

integration/ng_elements_schematics/test.ts
```diff
@@ -0,0 +1,49 @@
...
+
+// Add @angular/elements
+shx.exec('ng add "../../../dist/packages-dist/elements"');
+
+// Install Angular packages that are built locally from HEAD.
+shx.exec(`yarn add --ignore-scripts --silent ${packages}`);
```
gkalpak: Why are we installing the packages after running `ng add`? Couldn't we do it the other way around? (It would feel more close to what happens in reality.)
If there is a reason for the current order of commands, can you please add comments explaining that?

integration/ng_elements_schematics/test.ts
```diff
@@ -0,0 +1,49 @@
...
+
+// Test that build is successful after adding elements.
+shx.exec('ng build --no-source-map');
```
gkalpak: By default, ShellJS `exec()` (and other methods) will not cause the script to exit with an error (which is what we want here). We should use [set('-e')][1] at the top of the file.

[1]: https://documentup.com/shelljs/shelljs#setoptions

gkalpak (changes requested):

integration/ng_elements_schematics/test.ts
alan-agius4: For CI environments it non interactive, I’ll add the non interactive option

integration/ng_elements_schematics/test.ts
alan-agius4: Yeah agree.

integration/ng_elements_schematics/test.ts
alan-agius4: The reason being as the the CLI will install all these packages, and we don't want to have tests failure because of incompatibilities.
https://github.com/angular/angular-cli/blob/54b79ade65bfe3749e9ec2cd4fc52f6a74c26e94/packages/schematics/angular/workspace/files/package.json.template#L13-L44

alan-agius4:
@gkalpak PTAL

integration/ng_elements_schematics/test.ts
alan-agius4: No particular reason

integration/ng_elements_schematics/test.ts
gkalpak: I see, thx for clarifying. Could you mention that in a comment (so that people know why these packages are in the list and how to update the list in the future if needed)?

integration/ng_elements_schematics/test.ts
```diff
@@ -1,49 +1,51 @@
 #!/usr/bin/env node
 
-import * as shx from 'shelljs';
+import {cd, exec, set, rm} from 'shelljs';
```
gkalpak:
```suggestion
import {cd, exec, rm, set} from 'shelljs';
```

😇

integration/ng_elements_schematics/test.ts
```diff
@@ -1,49 +1,51 @@
...
+].map(p => `"${p}@file:${__dirname}/../../node_modules/${p}"`);
 
 const packages = [
   ...ngPackages,
   ...nodePackages,
 ].join(' ');
 
 // Clean up previously run test
-shx.rm('-rf', 'demo');
+rm('-rf', 'demo');
```
gkalpak:
```suggestion
rm('-rf', `${__dirname}/demo`);
```

or 

```suggestion
cd(__dirname);
rm('-rf', 'demo');
```

integration/ng_elements_schematics/test.ts
alan-agius4: done.

integration/ng_elements_schematics/test.ts
alan-agius4: Done

integration/ng_elements_schematics/test.ts
```diff
@@ -0,0 +1,52 @@
...
+  ...nodePackages,
+].join(' ');
+
+// Clean up previously run test
+rm('-rf', `${__dirname}/demo`);
+
+// Set up demo project
+exec('ng version');
+exec('ng new demo --skip-git --skip-install --style=css --no-interactive');
+cd('demo');
```
gkalpak: If you don't `cd(__dirname)` above, then you need to update this as well.
```suggestion
cd(`${__dirname}/demo`);
```

gkalpak (approved):

mhevery (approved):

PR #35251, opened 2020-02-08
feat(docs-infra): feature to download docs for offline viewing

### LINKED ISSUE #27492: feat(docs-infra): add option to cache all docs (e.g. for offline access)
We are currently able to persistently [cache via the SW][1] any doc (guide, tutorial, API page, etc.) that a user visits.
We could utilize that to provide an option to cache all docs eagerly (so that they are available offline, load faster on first visit, etc.).

Currently it is possible to hack your way there by manually running [this code][2] (e.g. via DevTools or as a bookmarklet).

##
**Possible options:**
- A button "hidden" somewhere (e.g. in the footer).
- A separate page with some (minimal or not so minimal) UI and explanation of the implications.

##
**Considerations:**
- How to handle errors. E.g.:
  * Sometimes some file requests randomly fail. (Re-download? Just log?)
- There is currently no reliable way to undo this. Once all files are cached, they will continue to be re-fetched and cached every time there is a change.
   (Fortunately, only the files that _did_ change are fetched.)

[1]: https://github.com/angular/angular/blob/8973dd5a7e6ede6d29a98684255088afac294a81/aio/ngsw-config.json#L60-L70
[2]: https://twitter.com/gkalpakas/status/1065929759593562113

ajitsinghkaler:
added feature to download docs for offline viewing featching all files tat need to be downloaded from ngsw.json files and downloading them all using fetch which leads to cacheing via service worker

Fixes #27492

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Right now cannot download docs for offline download.

Issue Number: #27492

## What is the new behavior?
Added a new para at the bottom for offline download of the docs

- [ ] Yes
- [x] No

### DIFF
--- a/aio/src/app/layout/footer/footer.component.html
+++ b/aio/src/app/layout/footer/footer.component.html
@@ -9,6 +9,8 @@
   </div>
 </div>
 
+<p (click)="downloadDocs()">Download docs for offline usage</p>
+
 <p>
   Super-powered by Google ©2010-2020.
   Code licensed under an <a href="license" title="License text">MIT-style License</a>.
--- a/aio/src/app/layout/footer/footer.component.html
+++ b/aio/src/app/layout/footer/footer.component.html
@@ -9,8 +9,6 @@
   </div>
 </div>
 
-<p (click)="downloadDocs()">Download docs for offline usage</p>
-
 <p>
   Super-powered by Google ©2010-2020.
   Code licensed under an <a href="license" title="License text">MIT-style License</a>.
--- a/aio/src/app/layout/footer/footer.component.ts
+++ b/aio/src/app/layout/footer/footer.component.ts
@@ -9,4 +9,25 @@ import { NavigationNode, VersionInfo } from 'app/navigation/navigation.service';
 export class FooterComponent {
   @Input() nodes: NavigationNode[];
   @Input() versionInfo: VersionInfo;
+
+  async downloadDocs(){
+    try {
+      const manifest = await fetch('ngsw.json').then(res => res.json());
+      const lazyUrls = manifest.assetGroups.find((g:any) => g.name === 'docs-lazy').urls;
+      const confirmed = confirm(`Download all ${lazyUrls.length} URLs?`);
+
+      if (confirmed) {
+        const start = performance.now();
+        const responses = await Promise.all(lazyUrls.map((u:any) => fetch(u)));
+        const failed = responses.filter((r:any) => !r.ok);
+        const failedCount = failed.length;
+        const duration = Math.round((performance.now() - start) / 1000);
+        failed.forEach((failure:any)=> console.log(`Failed to fetch ${failure.url}.`))
+        alert(`Downloaded ${lazyUrls.length} URLs in ${duration}s. (${failedCount} failed.)`);
+      }
+    } catch (err) {
+      console.error(err);
+      alert(`An error occurred:\n\n${err.stack}`);
+    }
+  }
 }
--- a/aio/src/app/layout/footer/footer.component.ts
+++ b/aio/src/app/layout/footer/footer.component.ts
@@ -9,25 +9,4 @@ import { NavigationNode, VersionInfo } from 'app/navigation/navigation.service';
 export class FooterComponent {
   @Input() nodes: NavigationNode[];
   @Input() versionInfo: VersionInfo;
-
-  async downloadDocs(){
-    try {
-      const manifest = await fetch('ngsw.json').then(res => res.json());
-      const lazyUrls = manifest.assetGroups.find((g:any) => g.name === 'docs-lazy').urls;
-      const confirmed = confirm(`Download all ${lazyUrls.length} URLs?`);
-
-      if (confirmed) {
-        const start = performance.now();
-        const responses = await Promise.all(lazyUrls.map((u:any) => fetch(u)));
-        const failed = responses.filter((r:any) => !r.ok);
-        const failedCount = failed.length;
-        const duration = Math.round((performance.now() - start) / 1000);
-        failed.forEach((failure:any)=> console.log(`Failed to fetch ${failure.url}.`))
-        alert(`Downloaded ${lazyUrls.length} URLs in ${duration}s. (${failedCount} failed.)`);
-      }
-    } catch (err) {
-      console.error(err);
-      alert(`An error occurred:\n\n${err.stack}`);
-    }
-  }
 }
--- a/aio/src/app/app.component.html
+++ b/aio/src/app/app.component.html
@@ -32,6 +32,7 @@
         <mat-icon svgIcon="logos:github"></mat-icon>
       </a>
     </div>
+    <aio-lazy-ce selector="aio-offline"></aio-lazy-ce>
   </mat-toolbar-row>
 </mat-toolbar>
 
--- a/aio/src/app/custom-elements/element-registry.ts
+++ b/aio/src/app/custom-elements/element-registry.ts
@@ -40,6 +40,10 @@ export const ELEMENT_MODULE_LOAD_CALLBACKS_AS_ROUTES = [
   {
     selector: 'live-example',
     loadChildren: () => import('./live-example/live-example.module').then(m => m.LiveExampleModule)
+  },
+  {
+    selector: 'aio-offline',
+    loadChildren: () => import('./offline/offline.module').then(m => m.OfflineModule)
   }
 ];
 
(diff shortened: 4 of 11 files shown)


ajitsinghkaler:
I will add tests once we are sure how this functionality has to be implemented. Right now using only @gkalpak gist to download all the files. I need additional input on how can we implement it better.

aio/src/app/layout/footer/footer.component.html
```diff
@@ -9,6 +9,8 @@ <h3>{{node.title}}</h3>
   </div>
 </div>
 
+<p (click)="downloadDocs()">Download docs for offline usage</p>
```
IgorMinar: We might want tweak this text to:
- [ ] make it shorter
- [ ] clarify that this is full offline usage (we already have partial offline usage which is on by default)
- [ ] add a download icon of some sort (@sjtrimble can help with that after we come to a conclusion that this PR is viable)
- [ ] decide on the position of the icon+text (again @sjtrimble can help with that after other technical issues are resolved)
- [ ] we should make this accessible by not using a click handler on `p` but instead use a `button` or an `a` or use the proper `aria` roles to communicate properly with screen readers.

aio/src/app/layout/footer/footer.component.ts
```diff
@@ -9,4 +9,25 @@ import { NavigationNode, VersionInfo } from 'app/navigation/navigation.service';
 export class FooterComponent {
   @Input() nodes: NavigationNode[];
   @Input() versionInfo: VersionInfo;
+
+  async downloadDocs(){
+    try {
+      const manifest = await fetch('ngsw.json').then(res => res.json());
+      const lazyUrls = manifest.assetGroups.find((g:any) => g.name === 'docs-lazy').urls;
+      const confirmed = confirm(`Download all ${lazyUrls.length} URLs?`);
```
IgorMinar: we should provide some kind of estimate about the download size to set the expectations. I ran a test and currently we download about 17MB of json files, without any images.

we should also modify the text to say that once downloaded, future updates will be incremental and automatic - this is actually huge win for the users because once they have the docs downloaded, they will always have a recent version of docs content available offline without any more manual downloads, as long as they occasionally connect to the internet and navigate to angular.io (which kicks in the automated incremental SW update).

aio/src/app/layout/footer/footer.component.ts
```diff
@@ -9,4 +9,25 @@ import { NavigationNode, VersionInfo } from 'app/navigation/navigation.service';
 export class FooterComponent {
   @Input() nodes: NavigationNode[];
   @Input() versionInfo: VersionInfo;
+
+  async downloadDocs(){
+    try {
+      const manifest = await fetch('ngsw.json').then(res => res.json());
+      const lazyUrls = manifest.assetGroups.find((g:any) => g.name === 'docs-lazy').urls;
```
IgorMinar: bah! it makes me a bit sad that we have to handle it like this because the whole reason for us designing assetGroups was so that we can eventually provide a new https://angular.io/api/service-worker/SwUpdate API that would trigger download of `lazy` assetGroups. So instead of all of this code, you would be able to just to `swUpdate.fetch('docs-lazy')` and be done.

We've however haven't gotten to implementing this (trivial) functionality - something we might consider as a follow up on this PR?

aio/src/app/layout/footer/footer.component.ts
```diff
@@ -9,4 +9,25 @@ import { NavigationNode, VersionInfo } from 'app/navigation/navigation.service';
...
+      const confirmed = confirm(`Download all ${lazyUrls.length} URLs?`);
+
+      if (confirmed) {
+        const start = performance.now();
+        const responses = await Promise.all(lazyUrls.map((u:any) => fetch(u)));
+        const failed = responses.filter((r:any) => !r.ok);
+        const failedCount = failed.length;
+        const duration = Math.round((performance.now() - start) / 1000);
+        failed.forEach((failure:any)=> console.log(`Failed to fetch ${failure.url}.`))
+        alert(`Downloaded ${lazyUrls.length} URLs in ${duration}s. (${failedCount} failed.)`);
```
IgorMinar: it took about 20sec for me to download everything on a relatively fast connection. Users with slower connections might need significantly more time, so we should consider providing some kind of progress indicator. We might need to replace the confirm/alert with `MatDialog` but we should be careful about not increasing the critical path bundle size in the process of doing that. Maybe the code (and the MatDialog) for full offline download could be moved to a lazy-loaded chunk?

I don't want to overcomplicate this PR, so let's discuss this one and whether my concern about lack of download progress indicator is a blocker or could be a follow up PR.

IgorMinar (changes requested):
Awesome work @ajitsinghkaler! I tested this PR and confirmed that it works reasonably well:
 
Screen Shot 2020-02-08 at 5 59 28 AM

I noted the "transfer" (compressed) payload size which is something we should make people aware of before the confirm that they really want to do this.

A few additional things I noticed that should be investigated/resolved:
- [ ] images are not being fetched, which results in poor UX because the app is not fully offline usable without them. We should include the images in the manifest file (currently they are not listed there)
- [ ] we need to confirm that as we modify the content and images in the future, the incremental update works automatically and is really incremental (only modified content and images should be automatically refetched) - this should work out of the box, but we should verify.
- [ ] we need some kind of test for this, a proper e2e test will be hard to build, but we should at minimum verify that the app is not broken after the full offline mode is activated.

Again awesome work @ajitsinghkaler! 🚀
I ❤️seeing all your improvements coming in.

aio/src/app/layout/footer/footer.component.ts
gkalpak: > they will always have a recent version of docs content available offline without any more manual downloads

Note that this is only true for existing guides/API docs. New guides/API docs needs to be explicitly downloaded.

aio/src/app/layout/footer/footer.component.ts
gkalpak: With the current approach of kicking off all downloads at once, I don't think it is possible to report progress, because the requests are downloaded from the SW (which first fully downloads the resource, then caches it and finally returns it to the page - so I don't think it shares progress with the page).

We could show some rough progress if we manually batched the requests (e.g. download 10 resources at a time), but that would not allow the browser to optimize the request batching.

gkalpak (commented):
Nice! 🚀 

I generally agree with Igor's comments (I left some additional comments inline and below).
A couple more issues I see:

- Once cached, there is no way to remove the cached resources (and stop them from being re-fetched on every update). The only option is to clear the caches (e.g. via DevTools), but that is not easy on mobile devices.

- The error messages (when a URL fails to be downloaded) are not very helpful.

##
> images are not being fetched, which results in poor UX because the app is not fully offline usable without them. We should include the images in the manifest file (currently they are not listed there)

AFAICT, the images are in the manifest and caches ad expected. Which images did you see missing?

sjtrimble:
Couple of options:

1) Add the "download offline" icon button to the toolbar next to the social icons. Only show this when the user is on a 'true' docs site such as `/docs`, `/guide`, and '/api`

2) Add a "Download Docs Offline" menu button to the menu (menu only appears when on 'true' docs sites so no further logic needed there.

docs

3) Add a link to the footer.
footer

I think both work - the second and third options are more subtle and less discoverable which might be good as I don't think this is a feature EVERYONE will need.

ajitsinghkaler:
@IgorMinar @gkalpak Made the changes put the offline content into a lazy-loaded chunk put it at the  top position. Added styles and mat dialog.

Now need some guidance or help on the following things

Icon from @sjtrimble can you please send me the SVG or should I download it my self from some free icon site

Tests: What kind of test do you think we should add here. Should we check that all requests are sent from the service worker because the test that these are saved correctly should already be present

Adding size to app data: should I change the build command to add size on each build

Progress bar: Should we show a progress bar based on the completion of each request or think about this on the follow-up pr

Errors: What to do with files that fail to download on these requests. Download again by saving failure cases or give an option to the user to try again once if that also fails, he can start the whole process over again

ajitsinghkaler:
@gkalpak @IgorMinar can you please review this.

ajitsinghkaler:
@gkalpak @IgorMinar can you please review this

IgorMinar:
Hi @ajitsinghkaler, I tested it on my phone and while I like the new dialog the download didn't seem to work. The UI doesn't look good on phone screen and when I clicked download nothing happened and the dialog remained open for several minutes until I closed it.
Screenshot_20200311-094613
Screenshot_20200311-094619
Screenshot_20200311-094657

Can you please take a look at these issues?

PR #35263, opened 2020-02-09
docs(forms): update the ngForm deprecation notice

IgorMinar:
we should be documenting when an API is eligible for removal and not when it will be removed.

The actual removal depends on many factors, e.g. if we were able to automate the refactoring to
the recommended API in time or not.


kara (approved):
LGTM, thanks for cleaning it up!

PR #35271, opened 2020-02-09
feat(language-service): improve non-callable error message

ayazhafiz:
This commit improves the context of a non-callable function error
message by providing the affected call target and its non-callable type.

## PR Checklist

## PR Type

- [x] Feature

- [ ] Yes
- [x] No

### DIFF
--- a/packages/language-service/src/expression_diagnostics.ts
+++ b/packages/language-service/src/expression_diagnostics.ts
@@ -21,6 +21,7 @@ export interface DiagnosticTemplateInfo {
   members: SymbolTable;
   htmlAst: Node[];
   templateAst: TemplateAst[];
+  source: string;
 }
 
 export function getTemplateExpressionDiagnostics(info: DiagnosticTemplateInfo): ng.Diagnostic[] {
@@ -103,7 +104,7 @@ function getVarDeclarations(
           // that have been declared so far are also in scope.
           info.query.createSymbolTable(results),
         ]);
-        symbol = refinedVariableType(variable.value, symbolsInScope, info.query, current);
+        symbol = refinedVariableType(variable.value, symbolsInScope, info, current);
       }
       results.push({
         name: variable.name,
@@ -143,11 +144,11 @@ function getVariableTypeFromDirectiveContext(
  * case for `ngFor` and `ngIf`. If resolution fails, return the `any` type.
  * @param value variable value name
  * @param mergedTable symbol table for all variables in scope
- * @param query
+ * @param info available template information
  * @param templateElement
  */
 function refinedVariableType(
-    value: string, mergedTable: SymbolTable, query: SymbolQuery,
+    value: string, mergedTable: SymbolTable, info: DiagnosticTemplateInfo,
     templateElement: EmbeddedTemplateAst): Symbol {
   if (value === '$implicit') {
     // Special case: ngFor directive
@@ -159,9 +160,10 @@ function refinedVariableType(
       const ngForOfBinding = ngForDirective.inputs.find(i => i.directiveName == 'ngForOf');
       if (ngForOfBinding) {
         // Check if there is a known type for the ngFor binding.
-        const bindingType = new AstType(mergedTable, query, {}).getType(ngForOfBinding.value);
+        const bindingType =
+            new AstType(mergedTable, info.query, {}, info.source).getType(ngForOfBinding.value);
         if (bindingType) {
-          const result = query.getElementType(bindingType);
+          const result = info.query.getElementType(bindingType);
           if (result) {
             return result;
           }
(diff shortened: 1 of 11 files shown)


ayazhafiz:
One thing we should start evaluating a design for is isolating error messages in a separate (json) file with unique IDs, descriptions, and even examples. Many modern language tools like tsserver, Rust's compiler, and Shellcheck follow this model. It will also provide a natural way for us to support enabling/disabling particular error messages as users have requested.

kyliau:
> One thing we should start evaluating a design for is isolating error messages in a separate (json) file with unique IDs, descriptions, and even examples. Many modern language tools like tsserver, Rust's compiler, and Shellcheck follow this model. It will also provide a natural way for us to support enabling/disabling particular error messages as users have requested.

In Ivy, each error already has a unique ID, but description is still generated on the fly.
It's good idea to pull them out into JSON.

https://github.com/angular/angular/blob/master/packages/compiler-cli/src/ngtsc/diagnostics/src/error_code.ts

packages/language-service/test/diagnostics_spec.ts
```diff
@@ -885,4 +885,17 @@ describe('diagnostics', () => {
     const ngDiags = ngLS.getSemanticDiagnostics(fileName);
     expect(ngDiags).toEqual([]);
   });
+
+  describe('function calls', () => {
+    it('should report error for non-callable function call', () => {
+      mockHost.override(TEST_TEMPLATE, `{{myClick()()}}`);
```
kyliau: This case checks for a function call of which the receiver is an ImplicitReceiver.
We should add another check for which the receiver is a property read.
like `title.toString()()`

packages/language-service/src/expression_type.ts
```diff
@@ -197,15 +197,18 @@ export class AstType implements AstVisitor {
     const args = ast.args.map(arg => this.getType(arg));
     const target = this.getType(ast.target !);
```
kyliau: `ast.target` is of type `AST` here, but really it's a `MethodCall` AST, of which the receiver is an `ImplicitReceiver`.
The `MethodCall` AST contains the name of the function call, so it looks like a better fix here is to make a safe cast to retrieve the name.

Alternatively, we could fix the typings upstream. Right now, many parameters are just typed `AST`. They should have been more specific.

If we go with the cast, we'll have to account for the other case which is a PropertyRead receiver. Please see next comment =)

kyliau (changes requested):

packages/language-service/src/expression_type.ts
ayazhafiz: Yeah, you're right. I'll see if we can type the ast specially.

ayazhafiz: The problem with pulling out the name is it only contains the name for the most local AST. So for something like `title.toUpperCase()()`, the property read AST name would only give us `toUpperCase`, and we would have to continue backing up and reconstruct what we think the source code looks like to get `title.toUpperCase()`. For this reason, I think it's better to parse out the source code from the AST span.

ayazhafiz:
ptal @kyliau

packages/language-service/src/expression_diagnostics.ts
```diff
@@ -332,9 +335,9 @@ class ExpressionDiagnosticsVisitor extends RecursiveTemplateAstVisitor {
     return ast.sourceSpan.start.offset;
   }
 
-  private diagnoseExpression(ast: AST, offset: number, event: boolean) {
-    const scope = this.getExpressionScope(this.path, event);
-    const analyzer = new AstType(scope, this.info.query, {event});
+  private diagnoseExpression(ast: AST, offset: number, inEvent: boolean) {
+    const scope = this.getExpressionScope(this.path, inEvent);
+    const analyzer = new AstType(scope, this.info.query, {inEvent}, this.info.source);
```
kyliau: We don't have to pass in the entire template, because the AST we got from the TemplateAst is actually ASTWithSource, so we already have the expression string here. This info combined with the span of the expression node should be sufficient for us to retrieve the string of interest.
Before we do that though, it might be better to update the type for all ASTs in `template_ast.ts` from AST to ASTWithSource. Please let me know if you need more clarification, thank you!

ayazhafiz: I'm not sure this will work, because AstType may visit child nodes of the AST, which may not be `ASTWithSource`s. We could pass the source of just the expression, but I am weary of this it would require using more relative spans. What do you think?

packages/language-service/src/expression_type.ts
```diff
@@ -360,6 +362,8 @@ export class AstType implements AstVisitor {
     return this.resolvePropertyRead(this.query.getNonNullableType(this.getType(ast.receiver)), ast);
   }
 
+  private sourceOf(ast: AST): string { return this.source.substring(ast.span.start, ast.span.end); }
```
kyliau: You brought up a very good point in https://github.com/angular/angular/pull/35271/files#r390700195
Could you please double check that `ast.span` here is relative to the whole template?
I think it might be relative to the expression in the template.
In the test case that you wrote in packages/language-service/test/diagnostics_spec.ts the expression is the template itself so it works. Could you please test this with a template that has multiple expressions?

ayazhafiz: Good call. Fixed to use absolute spans.

kyliau (approved):

Issue #35119, opened 2020-02-03
EventEmitter.emit() should only accept T, not T | undefined.


hmil:
# 🐞 bug report

### Affected Package

@angular/core

### Is this a regression?

No

### Description

The signature of `EventEmiter.emit` was modelled according to the signature of rxjs' `Subject.next` (see #10058 ). But the signature of `Subject.next` has a [design flaw](https://github.com/ReactiveX/rxjs/issues/2852) which has been copied to `EventEmitter.emit`.

This issue is essentially the same as https://github.com/ReactiveX/rxjs/issues/2852 , but applied to angular's EventEmitter. See repro below for details.

## 🔬 Minimal Reproduction

```ts
const em = new EventEmitter<string>();
em.asObservable().subscribe(s => {
    s.substring(0, 1); // boom
});
em.emit();
```

### Expected behavior

A compile error at line 5

### Actual behavior

A runtime error at line 3

### Details

The signature of `emit` allows emitting `undefined` on an event emitter which declares a non-nullable type.
This defeats the whole purpose of typescript's --strictNullCheck flag.

## 🔥 Exception or Error

No error, that's the problem.

## 🌍  Your Environment

All versions since #10058

## 🔧 Suggested fix

The suggested fix is simple enough, but it is a breaking change:
- Make the parameter of `emit` non-optional. One can still write `em.emit()` provided the EventEmitter is of type `EventEmitter<void>`.
- Make the default type of T `void` instead of `any` (see [this](https://github.com/ReactiveX/rxjs/issues/5066#issuecomment-581462164)) for details.

```diff
diff --git a/packages/core/src/event_emitter.ts b/packages/core/src/event_emitter.ts
index 07143d9774..eaf84b7ad0 100644
--- a/packages/core/src/event_emitter.ts
+++ b/packages/core/src/event_emitter.ts
@@ -61,7 +61,7 @@ import {Subject, Subscription} from 'rxjs';
  * @see [Observables in Angular](guide/observables-in-angular)
  * @publicApi
  */
-export class EventEmitter<T extends any> extends Subject<T> {
+export class EventEmitter<T extends any = void> extends Subject<T> {
   /**
    * @internal
    */
@@ -83,7 +83,7 @@ export class EventEmitter<T extends any> extends Subject<T> {
    * Emits an event containing a given value.
    * @param value The value to emit.
    */
-  emit(value?: T) { super.next(value); }
+  emit(value: T) { super.next(value); }
 
   /**
[...]

PR #35300, opened 2020-02-10
Fix/35119 strict emit type

hmil:
## PR Checklist

## PR Type

- [x] Bugfix

Change the type of EventEmitter to make it "more" type safe.

Issue Number: #35119

This error is not caught and will cause bugs at runtime:
```ts
const em = new EventEmitter<number>;
em.next(); // <- No value provided will feed `undefined` where `number` is expected
```

## What is the new behavior?

This error is caught by TypeScript
```ts
const em = new EventEmitter<number>;
em.next(); // <- Expected 1 argument to `next()` but 0 arguments provided
```

However, this valid behavior is still permitted:
```ts
const em = new EventEmitter();
em.next(); // <- This is OK because we did not specify what em is supposed to emit
```

In this PR, the default template type of `EventEmitter` is set to `EventEmitter<void>`. This is a strongly opinionated choice which favors safety. It might break code which expected the previous typing.

It's also perfectly valid to keep the previous default of `EventEmitter<any>`.

I find that using `<void>` is the safest alternative. It's more straightforward: `emit()` takes as argument exactly the type you specified in the template brackets, including no argument if you specified nothing. `<any>` is dangerous because the template argument may have been omitted by mistake and the user is unaware that the EventEmitter's data is not type checked.

The previous design was loose by default and let you opt-in strictness: defaults to `any`, allows more specific type to be provided. (noob friendly, but gets in the way of advanced users)
The new design is strict by default and lets you opt-out of strictness: defaults to `void`, but allows `any` to be specified. (ninja-friendly, but gets in the way of noobs).

What to chose depends on who is the primary audience of angular. Either way it's a breaking change.

- [x] Yes
- [ ] No

It's no longer possible to have an event emitter type which sometimes accepts arguments and sometimes not. It's either one but not both:

```ts
const em = new EventEmitter<void>();
em.next(); // OK
em.next(something); // Not OK

const em2 = new EventEmitter<any>();
em2.next(); // Not OK
em2.next(something); // OK
```

In the second example above, `em2.next();` should be replaced with `em2.next(undefined);`

Note that if you want to stick to the same API as RxJS, then wait on this PR to know what they chose to do: https://github.com/ReactiveX/rxjs/pull/5307

### DIFF
--- a/modules/playground/src/order_management/index.ts
+++ b/modules/playground/src/order_management/index.ts
@@ -139,7 +139,7 @@ export class OrderListComponent {
 })
 export class OrderItemComponent {
   @Input() item: OrderItem;
-  @Output() delete = new EventEmitter();
+  @Output() delete = new EventEmitter<OrderItem>();
 
   onDelete(): void { this.delete.emit(this.item); }
 }
--- a/packages/core/src/event_emitter.ts
+++ b/packages/core/src/event_emitter.ts
@@ -83,7 +83,7 @@ export class EventEmitter<T extends any> extends Subject<T> {
    * Emits an event containing a given value.
    * @param value The value to emit.
    */
-  emit(value?: T) { super.next(value); }
+  emit(value: T) { super.next(value); }
 
   /**
    * Registers handlers for events emitted by this instance.
--- a/packages/core/src/event_emitter.ts
+++ b/packages/core/src/event_emitter.ts
@@ -61,7 +61,7 @@ import {Subject, Subscription} from 'rxjs';
  * @see [Observables in Angular](guide/observables-in-angular)
  * @publicApi
  */
-export class EventEmitter<T extends any> extends Subject<T> {
+export class EventEmitter<T extends any = void> extends Subject<T> {
   /**
    * @internal
    */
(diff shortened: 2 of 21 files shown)


mhevery (approved):

mhevery:
Can you please look into the CI failures, so that this can be merged.

mhevery:
[presubmit](http://test/OCL:297278087:BASE:297278105:1582695589626:521c07c9)

atscott:
Hi @hmil, can you take a look at the CI failures and address them? Looks like you need to regenerate the public api golden

> yarn bazel run //tools/public_api_guard:core_api.accept

Thanks!

hmil:
Hey, as I am working through the changes to the test files, I'm thinking this will most likely break the vast majority of existing angular apps.

What is the plan to provide a smooth upgrade path? Is there a way to automate away the necessary refactoring? Or at least there should be an entry on https://update.angular.io/ explaining what to do.

amakhrov:
Hi there! Is there any plan to move forward with this improvement? Anything that would require some extra help?

hmil:
Hi @amakhrov , this branch needs to be rebased and failing tests need to be fixed. I don't have lots of time to take care of this these days. Feel free to pick it up from here.

amakhrov:
@hmil thanks. I rebased your branch and created a new PR.
Since the commits are co-authored by you, your consent on that PR is required. If you're ok with that, could you please leave a comment saying `@googlebot I consent.` in that PR: https://github.com/angular/angular/pull/38916 :)

AndrewKushnir:
Closing this PR in favor of #38916.

PR #20310, opened 2017-11-09
fix(forms): disable radio inputs in template driven forms

mohamed-gara:
In template driven forms, disable radio controls when they are initially disabled using data binding.

closes #18206

## PR Checklist
Please check if your PR fulfills the following requirements:

- [X] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [X] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

```
[X] Bugfix
[ ] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Documentation content changes
[ ] angular.io application / infrastructure changes
[ ] Other... Please describe:
```

## What is the current behavior?
https://github.com/angular/angular/issues/18206

Issue Number: 18206

## What is the new behavior?
In template driven forms, the radio inputs are correctly disabled when they are initially disabled using data binding.

## Does this PR introduce a breaking change?
```
[ ] Yes
[X] No
```

PR #18575, opened 2017-08-07
fix(forms): fix disabled attribute handling

dzonatan:
Update the disabled attribute of DOM element when setting a new FormControl instance either directly or via setControl().

Fixes #17806

## PR Checklist
Please check if your PR fulfills the following requirements:

- [X] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [X] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

```
[X] Bugfix
[ ] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Documentation content changes
[ ] angular.io application / infrastructure changes
[ ] Other... Please describe:
```

## What is the current behavior?

```
myGroup.setControl('ctrl', new FormControl({ value: 'Control is disabled', disabled: true }));
// Input dom element has disabled attribute. Correct.

myGroup.setControl('ctrl', new FormControl({ value: 'Control should be enabled', disabled: false }));
// Input dom element still has disabled attribute. WRONG! Should not have.
```

Full issue with plunker demo: #17806

## What is the new behavior?

```
myGroup.setControl('ctrl', new FormControl({ value: 'Control is disabled', disabled: true }));
// Input dom element has disabled attribute. Correct.

myGroup.setControl('ctrl', new FormControl({ value: 'Control should be enabled', disabled: false }));
// Input dom element does not have disabled attribute anymore. Correct.
```

## Does this PR introduce a breaking change?
```
[ ] Yes
[X] No
```

Issue #17806, opened 2017-06-28
FormControl does not reflect the new disabled value


dzonatan:
## I'm submitting a ...
<pre><code>[X] Bug report</code></pre>

## Current behavior
When setting new `FormControl` instance to `[formControl]`/`formControlName` directive the **disabled** property/attribute of control DOM element **does not update** to the newest value.

## Expected behavior
When setting new `FormControl` instance to `[formControl]`/`formControlName` directive the **disabled** property/attribute of control DOM element **should reflect** the newest value.

## Minimal reproduction of the problem with instructions
http://plnkr.co/edit/g5yxZRs8PvJWPAIag7wj?p=preview

## Please tell us about your environment
<pre><code>Angular version: 4.2.4

Browser:
- [X] Chrome (desktop) version 58.0
</code></pre>

alantreadway:
Since this issue and PR don't seem to be going anywhere, here's a workaround that seems to be working for me which can be easily removed if/when an actual fix is released:
https://gist.github.com/alantreadway/773c4880585bc9eb27703b3073c7fcd2

Caveats (YMMV):
- Not sure if there is a race/conflict between the update of the `disabled` state on the host element, although all update sources should be setting identical values if they are based on `FormControl.disabled` anyway, even after the proper fix.
- Guessing this is sub-optimal in terms of change-detection due to the use of a method call on every change-detector pass.

rzvjerkovic:
Any updates? I just wasted a day trying to debug this.
I replaced `formControlName` with `[formControl]`, but I don't think that's the solution.

chandrasagarjain:
Is there any update on this bug?

pavankjadda:
I just wasted couple of hours on this, Any update?

etay2000:
Figured I'd add a comment here just to prevent the conversation from being ignored and then automatically locked as per usual. How is it possible to have 587 pending pull requests? Is Google unwilling to pay more employees to work on Angular? I realize it is open source, but seriously wtf?

AndrewKushnir:
Hi, it looks like this issue has the same root cause as #35309, so I'm closing this PR in favor of #35309. Thank you.

PR #35311, opened 2020-02-10
fix(forms): Call setDisabledState in in reactive directives when appropriate

### LINKED ISSUE #35309: FormControlDirective and FormControlName do not call setDisabledState on ControlValueAcessor instances when the CVA is added and is enabled
# 🐞 bug report

### Affected Package

@angular/forms

### Is this a regression?

Not that I know of.

### Description

When you add a CVA to a form via the FormControlDirective directive or FormControlName directive with reactive forms and the CVA has implemented setDisabledState the the setDisabledState and the FormControl instance is enabled, setDisabledState is not called on the CVA instance, it is however called with the FormControl is disabled.

## 🔬 Minimal Reproduction

https://stackblitz.com/edit/angular-issue-repro2-3fagzc

## 🌍  Your Environment

**Angular Version:**
<pre><code>

google3 head

</code></pre>

**Anything else relevant?**

Nope

jermowery:
When a CVA is added a form and that CVA implements setDisabledState the
reactive form directives should call setDisabledState with true when the
FormControl is disabled and false when it is enabled. Currently,
setDisabledState is only called when the FormControl is disabled.

Fixes #35309

## PR Checklist

- [ x] Tests for the changes have been added (for bug fixes / features)
- [ x] Docs have been added / updated (for bug fixes / features)

## PR Type

- [ x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

setDisabledState when adding a control to a form is called with true with the FormControl is disabled and not called when the FormControl is enabled.

Issue Number: #35309 

## What is the new behavior?

setDisabledState is called when adding a control to a form with false when the FormControl is enabled and true with the FormControl is disabled.

- [ ] Yes
- [x ] No

### DIFF
--- a/packages/forms/test/value_accessor_integration_spec.ts
+++ b/packages/forms/test/value_accessor_integration_spec.ts
@@ -64,7 +64,7 @@ import {dispatchEvent} from '@angular/platform-browser/testing/src/browser_util'
       fixture.detectChanges();
 
       const input = fixture.debugElement.query(By.css('input'));
-      form.valueChanges.subscribe({next: (value) => { throw 'Should not happen'; }});
+      form.valueChanges.subscribe({next: (value) => { throw new Error('Should not happen'); }});
       input.nativeElement.value = 'updatedValue';
 
       dispatchEvent(input.nativeElement, 'change');
@@ -481,7 +481,8 @@ import {dispatchEvent} from '@angular/platform-browser/testing/src/browser_util'
         const assertOptionElementSelectedState = (selectedStates: boolean[]): void => {
           const options = fixture.debugElement.queryAll(By.css('option'));
           if (options.length !== selectedStates.length) {
-            throw 'the selected state values to assert does not match the number of options';
+            throw new Error(
+                'the selected state values to assert does not match the number of options');
           }
           for (let i = 0; i < selectedStates.length; i++) {
             expect(options[i].nativeElement.selected).toBe(selectedStates[i]);
@@ -1056,6 +1057,38 @@ import {dispatchEvent} from '@angular/platform-browser/testing/src/browser_util'
              expect(fixture.componentInstance.control.status).toEqual('DISABLED');
            });
 
+        describe('should support custom accessors with setDisabledState - formControlName', () => {
+          let fixture: ComponentFixture<CvaWithDisabledStateForm>;
+
+          beforeEach(() => { fixture = initTest(CvaWithDisabledStateForm, CvaWithDisabledState); });
+
+          it('sets the disabled state when the control is initally disabled', () => {
+            fixture.componentInstance.form = new FormGroup({
+              'login': new FormControl({value: 'aa', disabled: true}),
+            });
+            fixture.detectChanges();
+
+            expect(fixture.componentInstance.form.status).toEqual('DISABLED');
+            expect(fixture.componentInstance.form.get('login') !.status).toEqual('DISABLED');
+            expect(fixture.debugElement.query(By.directive(CvaWithDisabledState))
+                       .nativeElement.textContent)
+                .toEqual('DISABLED');
+          });
+
+          it('sets the enabled state when the control is initally enabled', () => {
+            fixture.componentInstance.form = new FormGroup({
+              'login': new FormControl({value: 'aa', disabled: false}),
+            });
+            fixture.detectChanges();
+
+            expect(fixture.componentInstance.form.status).toEqual('VALID');
+            expect(fixture.componentInstance.form.get('login') !.status).toEqual('VALID');
+            expect(fixture.debugElement.query(By.directive(CvaWithDisabledState))
+                       .nativeElement.textContent)
+                .toEqual('ENABLED');
+          });
+        });
+
         it('should populate control in ngOnInit when injecting NgControl', () => {
           const fixture = initTest(MyInputForm, MyInput);
           fixture.componentInstance.form = new FormGroup({'login': new FormControl('aa')});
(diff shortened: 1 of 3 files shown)


packages/forms/test/value_accessor_integration_spec.ts
```diff
@@ -15,7 +15,7 @@ import {dispatchEvent} from '@angular/platform-browser/testing/src/browser_util'
 {
   describe('value accessors', () => {
 
-    function initTest<T>(component: Type<T>, ...directives: Type<any>[]): ComponentFixture<T> {
+    function initTest<T>(component: Type<T>, ...directives: Array<Type<any>>): ComponentFixture<T> {
```
AndrewKushnir: We tend to use a short version of type notation wrt arrays, e.g. `string[]` vs `Array<string>`, so we should probably revert this change.

packages/forms/test/value_accessor_integration_spec.ts
```diff
@@ -1056,6 +1057,38 @@ import {dispatchEvent} from '@angular/platform-browser/testing/src/browser_util'
...
+            fixture.detectChanges();
+
+            expect(fixture.componentInstance.form.status).toEqual('DISABLED');
+            expect(fixture.componentInstance.form.get('login') !.status).toEqual('DISABLED');
+            expect(fixture.debugElement.query(By.directive(CvaWithDisabledState))
+                       .nativeElement.textContent)
+                .toEqual('DISABLED');
+          });
+
+          it('sets the disabled state when the control is initally enabled', () => {
```
AndrewKushnir: I'm a bit confused by the test description, should it say `sets the enabled state...` (seems like this is what you are also checking in the test)?

packages/forms/test/value_accessor_integration_spec.ts
```diff
@@ -481,7 +481,8 @@ import {dispatchEvent} from '@angular/platform-browser/testing/src/browser_util'
         const assertOptionElementSelectedState = (selectedStates: boolean[]): void => {
           const options = fixture.debugElement.queryAll(By.css('option'));
           if (options.length !== selectedStates.length) {
-            throw 'the selected state values to assert does not match the number of options';
+            throw new Error(
```
AndrewKushnir: 👍

AndrewKushnir (changes requested):
@jermowery thanks for creating a PR! 👍 I left a couple nits, please have a look.

It looks like this PR also introduces behavior change for Forms, where `setDisabledState` is called in situations where it was not called before. Could you please try to run Global TAP with this change to see if there are targets that might be potentially affected?

Thank you.

jermowery:
> It looks like this PR also introduces behavior change for Forms, where `setDisabledState` is called in situations where it was not called before. Could you please try to run Global TAP with this change to see if there are targets that might be potentially affected?

There are 3 specs failing as a result of this change. 2 of which are in my own codebase. I will put together a CL that can be patched with the fixes for the Angular caretaker doing the sync

AndrewKushnir (commented):
Thanks for additional changes @jermowery!

> There are 3 specs failing as a result of this change. 2 of which are in my own codebase. I will put
> together a CL that can be patched with the fixes for the Angular caretaker doing the sync

Thanks for verifying this in g3 👍Given that there are some (minor) breakages, I'd like to ask @kara to have a look at this PR too, to make sure we assess the risks accordingly.

Thank you.

jermowery:
I just took another look at my Global TAP and it looks like there are more changes than anticipated. I will discuss with you all tomorrow

sonukapoor:
Hi @jermowery, I was working on a similar PR, but noticed that your PR already covers what I was going to change as well. Is there anything I can help with for this PR?

jermowery:
Unfortunately, this breaks a significant number of tests inside Google, and I don't have to bandwidth to fix them, so this PR is pretty much abandoned

AndrewKushnir:
FYI, potentially related PR #18575 and related (duplicate?) issue #17806.

AndrewKushnir:
Hi @jermowery, since this is a breaking change and it affects a lot of targets internally in google3, I was thinking if we should consider adding a config option similar to the `warnOnNgModelWithFormControl` one [defined here](https://github.com/angular/angular/blob/d1ea1f4c7f3358b730b0d94e65b00bc28cae279c/packages/forms/src/form_providers.ts#L55-L64). That should keep existing behavior, allow enabling new/corrected one and simplify the rollout of this change (make it incremental). WDYT?

jermowery:
That seems reasonable. Sadly I don't have the bandwidth right now to incorporate that change but feel free to adopt this PR if it is a priority

grosch:
What's the status of this being implemented?

mohamed-gara:
There is already this PR #20310. Maybe it's the same fix.

Jakartoine:
@AndrewKushnir,

Sorry for disturbing you but do you have any update about this point?

To workaround this, I had to recreate a custom radio button component with an override on NG_VALUE_ACCESSOR.

Already 3 years since this bug has been discovered without any fix...

Thank you in advance

AndrewKushnir:
Hi @Jakartoine,

Unfortunately we don't have any updates on this issue at this moment.

The problem with the fix is that it requires a breaking change: it changes the behavior (to be more correct), but this affects existing apps since some of them rely on the current sequence of calls to the `setDisabledState` methods defined on CVAs. Running tests in Google's codebase indicated that the breaking change is quite severe and there is no easy way to auto-correct the code with schematic/migration. We have to come up with a migration strategy, such as introducing a flag (something like `ReactiveFormsModule.withConfig({setDisabledStateBehavior: 'legacy' /* or 'new' */})`) for apps to opt into using the "correct" behavior, wait for 1-2 major versions and only after that flip the flag to make "correct" behavior default. This work requires some planning and additional investigations and we didn't have a chance to prioritize this yet. We'll keep this PR updated once we have more info available.

Thank you.

Jakartoine:
@AndrewKushnir

Alright, it sounds clearer now. Thank you for your response mate.

dylhunn:
I'm going to run a TGP to find out how severe the breakage is. I suspect it's quite bad, but the only way to know is to try!

PR #35312, opened 2020-02-10
test(ngcc): add missing `UmdReflectionHost#getExportsOfModule()` tests

gkalpak:
Support for re-exports in UMD were added in e9fb5fdb8. This commit adds some tests for re-exports (similar to the ones used for `CommonJsReflectionHost`).

<sub>(Just something I came across while working on other stuff.)</sub>


petebacondarwin (approved):

petebacondarwin (approved):
Actually I think the commit message title should be `getExportsOfModule` not `getExportsOrModule`?

PR #35317, opened 2020-02-11
fix(language-service): get the right 'ElementAst' in the nested HTML tag

ivanwonder:
For example, `<div><p string-model~{cursor}></p></div>`, when provide the hover info for `string-model`, the `path.head` is root tag `div`. Use the parent of `path.tail` instead.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/packages/language-service/test/hover_spec.ts
+++ b/packages/language-service/test/hover_spec.ts
@@ -99,7 +99,8 @@ describe('hover', () => {
   });
 
   it('should be able to find a reference to a directive', () => {
-    const content = mockHost.override(TEST_TEMPLATE, `<div string-model~{cursor}></div>`);
+    const content =
+        mockHost.override(TEST_TEMPLATE, `<div><div string-model~{cursor}></div></div>`);
     const marker = mockHost.getLocationMarkerFor(TEST_TEMPLATE, 'cursor');
     const quickInfo = ngLS.getQuickInfoAtPosition(TEST_TEMPLATE, marker.start);
     expect(quickInfo).toBeDefined();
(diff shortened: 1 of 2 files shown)


ayazhafiz (approved):
Can we do `path.first` instead?

packages/language-service/test/hover_spec.ts
```diff
@@ -99,7 +99,8 @@ describe('hover', () => {
   });
 
   it('should be able to find a reference to a directive', () => {
-    const content = mockHost.override(TEST_TEMPLATE, `<div string-model~{cursor}></div>`);
+    const content =
```
kyliau: It is not immediately apparent what's the buggy behavior that this fix addresses.
Consider creating another component with a selector like `input[type=text]` (or anything that targets a particular element) then try to use it in the test template?
If the element is matched incorrectly the language service will not be able to provide a quick info.

ivanwonder: Do you mean `<div number-model><div string-model~{cursor}></div></div>`, but If the element is matched incorrectly the language service will provide nothing. Can it be tested?

ivanwonder:
> Can we do `path.first` instead?

Yes, I will use it instead

packages/language-service/test/hover_spec.ts
kyliau: ah, I get what you mean now. If the Element is not matched correctly then we would not be able to get the directives on the element.

kyliau (approved):

PR #35339, opened 2020-02-11
fix(core): undecorated-classes-with-decorated-fields migration does not decorate derived classes

### LINKED ISSUE #34376: Migration autofix not adding @Directive to intermediary component abstract classes
Testing 9.0.0-rc.6 upgrade from 8@latest

The migration logic to add @Directive to abstract classes that contain @Input is not working in all cases.

Where we have a component that has 2 layers of abstract class below it. 

The @Input is on the lowest level.

The migration adds the @Directive to the abstract class at the lowest level but it is also needed on the middle abstract class (which does not have any @Input in it).

Manually, adding the @Directive to the middle abstract class resolves the prod build error.

devversion:
The `undecorated-classes-with-decorated-fields` migration has been
introduced with https://github.com/angular/angular/commit/904a2018e0d3394ad91ffb6472f8271af7845295, but misses logic for decorating derived classes
of undecorated classes which use Angular features. Example scenario:

```ts
export abstract class MyBaseClass {
  @input() someInput = true;
}

export abstract class BaseClassTwo extends MyBaseClass {}

@component(...)
export class MyButton extends BaseClassTwo {}
```

Both abstract classes would need to be migrated. Previously, the migration
only added `@Directive()` to `MyBaseClass`, but with this change, it also
decorates `BaseClassTwo`.

This is necessary because the Angular Compiler requires `BaseClassTwo` to
have a directive definition when it flattens the directive metadata for
`MyButton` in order to perform type checking. Technically, not decorating
`BaseClassTwo` does not break at runtime because the inherit definition feature
walks up the prototype chain. Though we want to enforce consistent use of
`@Directive` to simplify the mental model. [See the migration guide for motivation](https://angular.io/guide/migration-undecorated-classes#migrating-classes-that-use-field-decorators).

Fixes #34376.

**Note**: It should be easier and less verbose to to review the commits individually.


packages/core/schematics/migrations/undecorated-classes-with-decorated-fields/google3/tslint_update_recorder.ts
```diff
@@ -0,0 +1,46 @@
...
+    // Adding a decorator should be the last replacement. Replacements/rule failures
+    // are handled in reverse and in case a decorator and import are inserted at
+    // the start of the file, the class decorator should come after the import.
+    this.failures.unshift(new RuleFailure(
+        this.sourceFile, node.getStart(), 0, `Class needs to be decorated with ` +
+            `"${decoratorText}" because it uses Angular features.`,
+        this.ruleName, Replacement.appendText(node.getStart(), `${decoratorText}\n`)));
+  }
+
+  addNewImport(start: number, importText: string) {
```
crisbeto: Rather than passing in the start, maybe we should pass in the AST node here instead? I'm not sure that it's obvious what `start` means in this context.

packages/core/schematics/migrations/undecorated-classes-with-decorated-fields/transform.ts
```diff
@@ -0,0 +1,157 @@
...
+    };
+  }
+
+  /**
+  * Checks whether the given decorator resolves to an abstract directive. An directive is
+  * considered "abstract" if there is no selector specified.
+  */
+  private _isAbstractDirective({node}: NgDecorator): boolean {
+    const metadataArgs = node.expression.arguments;
+    if (metadataArgs.length === 0) {
```
crisbeto: A bit of an edge case, but the `node.expression` isn't guaranteed to be a call expression, it could also be an identifier if the user did something like this:

```
@Directive
class Foo {}
```

Maybe we should null check the arguments?

crisbeto (approved):
A couple of nits, but otherwise LGTM.

packages/core/schematics/migrations/undecorated-classes-with-decorated-fields/transform.ts
JoostK: This is taken care of when creating the `NgDecorator` type: https://github.com/angular/angular/blob/ea27239b4d4bccc54c9910c2baaaac423a91a9d5/packages/core/schematics/utils/typescript/decorators.ts#L12-L23

devversion: Yep. We also have it strongly typed that `NgDecorator#node` is a `CallExpression`. I was also trying out if the above snippet works, but it seems like it won't, so I think there is no action on this.

packages/core/schematics/migrations/undecorated-classes-with-decorated-fields/google3/tslint_update_recorder.ts
devversion: What AST node should we pass? I don't think passing the node would allow us to drop `start`. The node is synthetically created and doesn't have any start/end.

I'll try to make it more clear by adding method comments.

devversion:
@alxhub A similar discussion came up for this in the `ngcc` channel. It looks like it's unclear whether intermediary classes (as shown in the PR description) need to be decorated or not. It might be good to have a look at that discussion, so that we can have a consistent solution for both ngcc and `ng update`.

alxhub (approved):

alxhub:
Nice work!

josephperrott (approved):
LGTM

josephperrott (approved):
LGTM

Reviewed-for: dev-infra

devversion:
Caretaker: Can you please run presubmit for this? I'm unsure whether schematic code is synced into g3. If it is not, then we should update the bot to exclude the folder.

kara:
[presubmit](http://test/OCL:304274599:BASE:304274640:1585780064499:92ed9455)

PR #35343, opened 2020-02-11
fix(ivy): error if directive with synthetic property binding is on same node as directive that injects ViewContainerRef

### LINKED ISSUE #35342: Can't attach a matTooltip to mat-expansion-panel-header since 9.0.0
#### Reproduction

Steps to reproduce:
1. Clone and run https://github.com/tadzik/angular-material-bug
2. Observe the browser console

Alternatively: try to attach a `matTooltip` to a `mat-expansion-panel-header`, as you could in angular/material 8.
 
#### Expected Behavior

Everything works as expected without exceptions thrown.

#### Actual Behavior

Component (?) crashes:

```
ERROR TypeError: "renderer is undefined"
    Angular 8
        isProceduralRenderer
        listenerInternal
        componentHostSyntheticListener
        MatExpansionPanelHeader_HostBindings
        invokeHostBindingsInCreationMode
        invokeDirectivesHostBindings
        createDirectivesInstances
        elementStart
    AppComponent_Template app.component.html:2
    Angular 20
core.js:5828:19
```

Removing the `matTooltip` on `mat-expansion-panel-header` fixes the issue – but removes the tooltip too :)

#### Environment

- Angular: 9
- CDK/Material: 9
- Browser(s): Firefox
- Operating System (e.g. Windows, macOS, Ubuntu): Ubuntu

crisbeto:
In the `loadRenderer` we make an assumption that the value will always be an `LView`, but if there's a directive on the same node which injects `ViewContainerRef` the `LView` will be wrapped in an `LContainer`. These changes add a call to unwrap the value before we try to read the value off of it.

Fixes #35342.

### DIFF
--- a/packages/core/test/acceptance/property_binding_spec.ts
+++ b/packages/core/test/acceptance/property_binding_spec.ts
@@ -5,8 +5,9 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
+import {state, style, trigger} from '@angular/animations';
 import {CommonModule} from '@angular/common';
-import {Component, Directive, EventEmitter, Input, Output} from '@angular/core';
+import {Component, Directive, EventEmitter, Input, Output, ViewContainerRef} from '@angular/core';
 import {TestBed} from '@angular/core/testing';
 import {By, DomSanitizer, SafeUrl} from '@angular/platform-browser';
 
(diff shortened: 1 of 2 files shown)


packages/core/test/acceptance/property_binding_spec.ts
```diff
@@ -598,4 +599,32 @@ describe('property bindings', () => {
...
+       @Component({template: '<my-comp my-dir></my-comp>'})
+       class App {
+       }
+
+       TestBed.configureTestingModule({declarations: [App, MyDir, MyComp]});
+
+       expect(() => {
+         const fixture = TestBed.createComponent(App);
+         fixture.detectChanges();
+       }).not.toThrow();
```
pkozlowski-opensource: I don't think that you need explicit `not.toThrow();` a test would fail in case of an exception thrown.

mhevery: I actually prefer this as it makes it very clear what you are testing. Without this it is harder to understand what exactly you are testing. I think of it as a marker.

mhevery (approved):

mhevery:
[presubmit](http://test/OCL:294484873:BASE:294484926:1581449995342:cc20e5ed)

packages/core/test/acceptance/property_binding_spec.ts
pkozlowski-opensource: Hmmm, I hear what you say about marker and this is a good point. The problem I've got with such assertion, though, is that this test is going to pass even if nothing happens /  gets rendered. I could pretty much delete the entire template and the test would still pass.

This is why, personally, I prefer to have a more concrete assert... Definitively don't want to split a hair over this one, just sharing personal preference / how I think about it...

PR #35348, opened 2020-02-11
ci: use pipeline values to define the CI_COMMIT_RANGE

josephperrott:
Taken from CircleCI's [Pipeline Values](https://circleci.com/docs/2.0/pipeline-variables/#pipeline-values) example, this is the expected method for getting the values formerly found in the `CIRCLE_COMPARE_URL`


.circleci/env.sh
```diff
@@ -27,11 +27,7 @@ setPublicVar CI_CHROMEDRIVER_VERSION_ARG "--versions.chrome 79.0.3945.130";
 setPublicVar CI_COMMIT "$CIRCLE_SHA1";
 # `CI_COMMIT_RANGE` is only used on push builds (a.k.a. non-PR, non-scheduled builds and rerun
 # workflows of such builds).
-# NOTE: With [CircleCI Pipelines](https://circleci.com/docs/2.0/build-processing) enabled,
-#       `CIRCLE_COMPARE_URL` is no longer available and the commit range cannot be reliably
-#       detected. Fall back to only considering the last commit (which is accurate in the majority
-#       of cases for push builds).
-setPublicVar CI_COMMIT_RANGE "`[[ ${CIRCLE_PR_NUMBER:-false} != false ]] && echo "" || echo "$CIRCLE_SHA1~1...$CIRCLE_SHA1"`";
+setPublicVar CI_COMMIT_RANGE "`[[ ${CIRCLE_PR_NUMBER:-false} != false ]] && echo "" || echo "$CIRCLE_GIT_BASE_REVISION..$CIRCLE_GIT_REVISION"`";
```
IgorMinar: Is it `..` or `...`? I keep on getting the two mixed up. #git

IgorMinar (approved):
Otherwise lgtm. @gkalpak should take a look too as he's been dealing with this for aio previews and payload size monitoring scripts which we could update as a result of this.

.circleci/env.sh
josephperrott: We will have to confirm with @gkalpak, but two dots is what Circle used.



Three dots `A...B` will show all of the commits after the common ancestor in A and B minus its intersection (`(A + B) - A∩B`)

Two dots `A..B` will show all of the commits in B after its common ancestor with A

I am not 100% sure of what we are looking for here.

gkalpak: Confusingly, the dots can mean different things in [git diff][1] than [git log][2]. [We are using it][3] with `git log`, so what Joey said applies.

We sued to extract the range from CircleCI's compare URL (which is supposed to work sort of like `git diff`), so it contained three dots. In this case it shouldn't matter, because the first commit should represent the merge-base between the two, so both notations should be equivalent for our purpose.

[1]: https://git-scm.com/docs/git-diff#Documentation/git-diff.txt-emgitdiffemltoptionsgtltcommitgtltcommitgt--ltpathgt82308203
[2]: https://git-scm.com/docs/gitrevisions#Documentation/gitrevisions.txt-emltrev1gtltrev2gtem
[3]: https://github.com/angular/angular/blob/eef047bc5/scripts/ci/payload-size.sh#L78

gkalpak: If this works for PR and non-PR builds, there is no reason to differentiate based on `$CIRCLE_PR_NUMBER`.
(We needed to do this before, because `CIRCLE_COMPARE_URL` was not always available.)

gkalpak (approved):

.circleci/env.sh
gkalpak: BTW, we used to use the range for `git diff`, but this was problematic (for several reasons) and was removed in 9b5299131.

gkalpak:
BTW, we can also remove [this script][1] (as I don't forsee it being useful in the future).

[1]: https://github.com/angular/angular/blob/4aeb6c4e7/.circleci/get-commit-range.js

Issue #35219, opened 2020-02-07
sometime form .get return null


mix5003:
# 🐞 bug report
sometime form .get return null. i have no idea why? 
for example in my code i just add `withholdingTax.items` in to form and i try to subscribe valueChanges but it said i try to access valueChanges of null value 
![valueChangesOfNull1](https://user-images.githubusercontent.com/1129524/74016932-d76f4500-49c5-11ea-993d-9d40e933d9c7.png)
![valueChangesOfNull2](https://user-images.githubusercontent.com/1129524/74016939-d9390880-49c5-11ea-9a8d-625d0e15ebba.png)

After i try to find a re-produce step to fix it i found that sometime `form.get` will return null. as you can see in next screenshot line 195 and 199 is same code (`console.log(this.form.get('withholdingTax.items'));` and not touch any controls in the form but result is different 
![valueChangesOfNull3](https://user-images.githubusercontent.com/1129524/74017193-5f554f00-49c6-11ea-9a9c-9d69be955d14.png)
 
list of error that i found
- `Cannot read property 'controls' of null`
- `Cannot read property 'valueChanges' of null`
- `Cannot read property 'setValue' of null`
- `Cannot read property 'value' of null`
- `Cannot find control with name`
- `Cannot read property 'length' of null` (FormArray)

### Affected Package
i think it should be `@angular/forms`

### Description

 A clear and concise description of the problem...
```
TypeError: Cannot read property 'valueChanges' of null
    at FormPaymentVoucherComponent.push../src/app/purchase/payment-voucher/form-payment-voucher/form-payment-voucher.component.ts.FormPaymentVoucherComponent.initForm (form-payment-voucher.component.ts:193)
    at FormPaymentVoucherComponent.push../src/app/component/base-form/base-form.component.ts.BaseFormComponent.ngOnInit (base-form.component.ts:36)
    at FormPaymentVoucherComponent.push../src/app/purchase/shared/base-purchase-form/base-purchase-form.component.ts.BasePurchaseFormComponent.ngOnInit (base-purchase-form.component.ts:49)
    at FormPaymentVoucherComponent.push../src/app/purchase/payment-voucher/form-payment-voucher/form-payment-voucher.component.ts.FormPaymentVoucherComponent.ngOnInit (form-payment-voucher.component.ts:99)
    at checkAndUpdateDirectiveInline (core.js:21096)
    at checkAndUpdateNodeInline (core.js:29494)
    at checkAndUpdateNode (core.js:29456)
[...]

PR #35349, opened 2020-02-11
fix(forms): change Array.reduce usage to Array.forEach

kara:
There is currently a bug in Chrome 80 that makes Array.reduce
not work according to spec. The functionality in forms that
retrieves controls from FormGroups and FormArrays (`form.get`)
relied on Array.reduce, so the Chrome bug broke forms for
many users.

This commit refactors our forms code to rely on Array.forEach
instead of Array.reduce to fix forms while we are waiting
for the Chrome fix to go live.

See https://bugs.chromium.org/p/chromium/issues/detail?id=1049982.

### DIFF
--- a/packages/forms/src/model.ts
+++ b/packages/forms/src/model.ts
@@ -52,17 +52,21 @@ function _find(control: AbstractControl, path: Array<string|number>| string, del
   }
   if (Array.isArray(path) && path.length === 0) return null;
 
-  return path.reduce((v: AbstractControl | null, name) => {
-    if (v instanceof FormGroup) {
-      return v.controls.hasOwnProperty(name as string) ? v.controls[name] : null;
+  // Not using Array.reduce here due to a Chrome 80 bug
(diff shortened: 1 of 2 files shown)


AndrewKushnir (approved):
LGTM 👍 Thanks for the fix!

AndrewKushnir (approved):

kara:
[presubmit](http://test/OCL:294544945:BASE:294544986:1581466466080:2e35e334)

samvloeberghs:
is this retrofitted somehow in for example ng v6/7-v8?

maartentibau:
> is this retrofitted somehow in for example ng v6/7-v8 (if necessary?)? @kara

Same here

mstislavm:
will there be fix for Angular 7?

kara:
This PR fixes the problem for version 9. For earlier versions, it's possible to fix the problem by patching in a polyfill - see https://github.com/angular/angular/issues/35219#issuecomment-583465676.

PR #35356, opened 2020-02-12
[DO NOT MERGE] test: check animations behavior in IE10

AndrewKushnir:
This is a test-only PR, DO NOT MERGE.

The goal of this PR is to check whether a simple animations case passes IE10 tests on CI.


AndrewKushnir:
Result: basic animation test is [passing in all browsers](https://circleci.com/workflow-run/6e686357-51d7-41bf-805f-5357e5313cc3).

PR #35379, opened 2020-02-12
docs-infra: update AIO dependencies

petebacondarwin:


petebacondarwin:
Interesting that the size checks failed:

```
Verifying sizes against '/home/circleci/ng/aio/scripts/_payload-limits.json'...
SUCCESS: Commit 0682487602fcf0ef82259df8d7abfd632be07273 uncompressed runtime-es2015 did NOT cross size threshold of 500 bytes or >1% (expected: 2987, actual: 2987).
FAIL: Commit 0682487602fcf0ef82259df8d7abfd632be07273 uncompressed main-es2015 fell below expected size by 500 bytes or >1% (expected: 455897, actual: 449223).
FAIL: Commit 0682487602fcf0ef82259df8d7abfd632be07273 uncompressed polyfills-es2015 exceeded expected size by 500 bytes or >1% (expected: 52195, actual: 55321).
```

`main-es2015` went down by too much; while `polyfills-es2015` went up!

I think this is reasonable. Updating the PR.

aio/package.json
```diff
@@ -87,28 +87,28 @@
...
-    "@angular/compiler": "9.0.0-rc.11",
-    "@angular/core": "9.0.0-rc.11",
-    "@angular/elements": "9.0.0-rc.11",
-    "@angular/forms": "9.0.0-rc.11",
+    "@angular/common": "9.0.0",
+    "@angular/compiler": "9.0.0",
+    "@angular/core": "9.0.0",
+    "@angular/elements": "9.0.0",
+    "@angular/forms": "9.0.0",
     "@angular/material": "9.0.0-rc.8",
```
gkalpak: Maybe update CDK/Material too?

gkalpak (approved):

aio/scripts/_payload-limits.json
```diff
@@ -3,26 +3,26 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 2987,
-        "main-es2015": 455897,
-        "polyfills-es2015": 52195
+        "main-es2015": 449223,
+        "polyfills-es2015": 55321
```
gkalpak: I just noticed this: Do we know what is causing it?

aio/package.json
petebacondarwin: Let me try...

aio/scripts/_payload-limits.json
petebacondarwin: The old and the new files seem really quite different. The `git diff` does not help much.

aio/package.json
petebacondarwin: Added

aio/scripts/_payload-limits.json
petebacondarwin: The reason was the update to `@webcomponents/custom-element` which I have reverted.

mhevery (approved):

aio/scripts/_payload-limits.json
```diff
@@ -12,7 +12,7 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 2987,
-        "main-es2015": 448928,
+        "main-es2015": 449438,
```
kara: Do we know where the 500 bytes are coming from? Did we add more

petebacondarwin: So even on master right now it is 449289, so this PR's change is actually really only 149 bytes. I am not sure what the difference is though... I'll look into it a bit.

petebacondarwin: I looked through the diff and could not see any obvious changes that I could identify as being the culprit. I suspect it is a combination of the styling refactor and changes to Material/CDK.

kara (approved):

PR #35381, opened 2020-02-12
build: several minor fixes related to using `puppeteer`

gkalpak:
This PR is a follow-up to #35049 and contains several fixes related to using the browser provided by `puppeteer` to run tests. It contains the following commits:

---
**build: several minor fixes related to using `puppeteer`**

This is a follow-up to #35049 with a few minor fixes related to using the browser provided by `puppeteer` to run tests. Included fixes:

- Make the `webdriver-manager-update.js` really portable. (Previously,
  it needed to be run from the directory that contained the
  `node_modules/` directory. Now, it can be executed from a subdirectory
  and will correctly resolve dependencies.)

- Use the `puppeteer`-based setup in AIO unit and e2e tests to ensure
  that the downloaded ChromeDriver version matches the browser version
  used in tests.

- Use the `puppeteer`-based setup in the `aio_monitoring_stable` CI job
  (as happens with `aio_monitoring_next`).

- Use the [recommended way][1] of getting the browser port when using
  `puppeteer` with `lighthouse` and avoid hard-coding the remote
  debugging port (to be able to handle multiple instances running
  concurrently).

---
**build(docs-infra): use `puppeteer` to get a browser for docs examples tests**

In #35049, integration and AIO tests were changed to use the browser provided by `puppeteer` in tests. This commit switches the docs examples tests to use the same setup.

IMPLEMENTATION NOTE:
The examples are used to create ZIP archives that docs users can download to experiment with. Since we want the downloaded projects to resemble an `@angular/cli` generated project, we do not want to affect the project's Protractor configuration in order to use `puppeteer`.

To achieve this, a second Protractor configuration is created (which is ignored when creating the ZIP archives) that extends the original one and passes the approperiate arguments to use the browser provided by `puppeteer`. This new configuration (`protractor-puppeteer.conf.js`) is used when running the docs examples tests (on CI or locally during development).

---
**ci: get rid of the `CI_CHROMEDRIVER_VERSION_ARG` env var**

Previously, we needed to manually specify a ChromeDriver version to download on CI that would be compatible with the browser version provided by the docker image used to run the tests. This was kept in the `CI_CHROMEDRIVER_VERSION_ARG` environment variable.

With recent commits, we use the browser provided by `puppeteer` and can determine the correct ChromeDriver version programmatically. Therefore, we no longer need the `CI_CHROMEDRIVER_VERSION_ARG` environment variable.

NOTE:
There is still one place (the `bazel-schematics` integration project) where a hard-coded ChromeDriver version is necessary. Since I am not sure what is the best way to refactor the tests to not rely on a hard-coded version, I left it as a TODO for a follow-up PR.

[1]: https://github.com/GoogleChrome/lighthouse/blame/51df179a0/docs/puppeteer.md#L49

### DIFF
--- a/.circleci/env.sh
+++ b/.circleci/env.sh
@@ -19,11 +19,6 @@ setPublicVar CI_AIO_MIN_PWA_SCORE "95";
 # This is the branch being built; e.g. `pull/12345` for PR builds.
 setPublicVar CI_BRANCH "$CIRCLE_BRANCH";
 setPublicVar CI_BUILD_URL "$CIRCLE_BUILD_URL";
-# ChromeDriver version compatible with the Chrome version included in the docker image used in
-# `.circleci/config.yml`. See http://chromedriver.chromium.org/downloads for a list of versions.
-# This variable is intended to be passed as an arg to the `webdriver-manager update` command (e.g.
-# `"postinstall": "webdriver-manager update $CI_CHROMEDRIVER_VERSION_ARG"`).
-setPublicVar CI_CHROMEDRIVER_VERSION_ARG "--versions.chrome 79.0.3945.130";
 setPublicVar CI_COMMIT "$CIRCLE_SHA1";
 # `CI_COMMIT_RANGE` is only used on push builds (a.k.a. non-PR, non-scheduled builds and rerun
 # workflows of such builds).
--- a/aio/karma.conf.js
+++ b/aio/karma.conf.js
@@ -35,10 +35,10 @@ module.exports = function (config) {
       ChromeHeadlessNoSandbox: {
         base: 'ChromeHeadless',
         // See /integration/README.md#browser-tests for more info on these args
-        flags: ['--no-sandbox', '--headless', '--disable-gpu', '--disable-dev-shm-usage', '--hide-scrollbars', '--mute-audio']
-      }
+        flags: ['--no-sandbox', '--headless', '--disable-gpu', '--disable-dev-shm-usage', '--hide-scrollbars', '--mute-audio'],
+      },
     },
-    browsers: [process.env['CI'] ? 'ChromeHeadlessNoSandbox' : 'Chrome'],
+    browsers: ['ChromeHeadlessNoSandbox'],
     browserNoActivityTimeout: 60000,
     singleRun: false,
     restartOnFileChange: true,
--- a/aio/scripts/audit-web-app.js
+++ b/aio/scripts/audit-web-app.js
@@ -27,10 +27,10 @@
  */
 
 // Imports
-const puppeteer = require('puppeteer');
 const lighthouse = require('lighthouse');
 const printer = require('lighthouse/lighthouse-cli/printer');
 const logger = require('lighthouse-logger');
+const puppeteer = require('puppeteer');
 
 // Constants
 const AUDIT_CATEGORIES = ['accessibility', 'best-practices', 'performance', 'pwa', 'seo'];
@@ -83,10 +83,8 @@ function formatScore(score) {
 }
 
 async function launchChromeAndRunLighthouse(url, flags, config) {
-  const browser = await puppeteer.launch({
-    headless: true,
-    args: ['--remote-debugging-port=9222']});
-  flags.port = browser.port;
+  const browser = await puppeteer.launch();
+  flags.port = (new URL(browser.wsEndpoint())).port;
 
   try {
     return await lighthouse(url, flags, config);
--- a/scripts/webdriver-manager-update.js
+++ b/scripts/webdriver-manager-update.js
@@ -10,17 +10,19 @@
  */
 // Use process.cwd() so that this script is portable and can be used in /aio
 // where this will require /aio/node_modules/puppeteer
-const puppeteerVersion = require(`${process.cwd()}/node_modules/puppeteer/package.json`).version;
+const puppeteerPkgPath = require.resolve('puppeteer/package.json', {paths: [process.cwd()]});
+const puppeteerVersion = require(puppeteerPkgPath).version;
 const chromeVersionMap = require('./puppeteer-chrome-versions');
 const spawnSync = require('child_process').spawnSync;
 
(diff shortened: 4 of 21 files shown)


.circleci/env.sh
```diff
@@ -19,11 +19,6 @@ setPublicVar CI_AIO_MIN_PWA_SCORE "95";
 # This is the branch being built; e.g. `pull/12345` for PR builds.
 setPublicVar CI_BRANCH "$CIRCLE_BRANCH";
 setPublicVar CI_BUILD_URL "$CIRCLE_BUILD_URL";
-# ChromeDriver version compatible with the Chrome version included in the docker image used in
-# `.circleci/config.yml`. See http://chromedriver.chromium.org/downloads for a list of versions.
-# This variable is intended to be passed as an arg to the `webdriver-manager update` command (e.g.
-# `"postinstall": "webdriver-manager update $CI_CHROMEDRIVER_VERSION_ARG"`).
-setPublicVar CI_CHROMEDRIVER_VERSION_ARG "--versions.chrome 79.0.3945.130";
```
gregmagolan: I think we still need this for one last test which is the integration/bazel-schematics. Once that is updated to puppeteer we can finally kill it.

aio/karma.conf.js
```diff
@@ -35,10 +35,10 @@ module.exports = function (config) {
       ChromeHeadlessNoSandbox: {
         base: 'ChromeHeadless',
         // See /integration/README.md#browser-tests for more info on these args
-        flags: ['--no-sandbox', '--headless', '--disable-gpu', '--disable-dev-shm-usage', '--hide-scrollbars', '--mute-audio']
-      }
+        flags: ['--no-sandbox', '--headless', '--disable-gpu', '--disable-dev-shm-usage', '--hide-scrollbars', '--mute-audio'],
+      },
     },
-    browsers: [process.env['CI'] ? 'ChromeHeadlessNoSandbox' : 'Chrome'],
+    browsers: ['ChromeHeadlessNoSandbox'],
```
gregmagolan: For local testing I suppose you're no relying on the browser popping up and if you need it you can always change this temporarily?

aio/scripts/audit-web-app.js
```diff
@@ -83,10 +83,8 @@ function formatScore(score) {
 }
 
 async function launchChromeAndRunLighthouse(url, flags, config) {
-  const browser = await puppeteer.launch({
-    headless: true,
-    args: ['--remote-debugging-port=9222']});
-  flags.port = browser.port;
+  const browser = await puppeteer.launch();
+  flags.port = (new URL(browser.wsEndpoint())).port;
```
gregmagolan: 😎

scripts/webdriver-manager-update.js
```diff
@@ -10,17 +10,19 @@
  */
 // Use process.cwd() so that this script is portable and can be used in /aio
```
gregmagolan: nit: comment now out of date

scripts/webdriver-manager-update.js
```diff
@@ -30,13 +32,12 @@ const args = [
   ...process.argv.slice(2),
 ];
 
-const isWindows = process.platform === 'win32';
-const result = spawnSync(`${process.cwd()}/node_modules/.bin/webdriver-manager${isWindows ? '.cmd' : ''}`, args, {stdio: 'inherit'});
+const result = spawnSync('yarn', args, {shell: true, stdio: 'inherit'});
```
gregmagolan: I was avoiding yarn as in a follow up PR i'll run the aio build in npm_integration_test so under Bazel I want to ensure that the correct version of yarn is used. However, under Bazel I think `yarn` will be on the path and it will redirect to the Bazel provisioned yarn to be hermetic.

However, this may now be more an issue outside of Bazel now since we're relying on the root `.yarnrc` to use yarn 1.21.1 and in this context there will be no `.yarnrc` in the cwd. + @josephperrott 👀

gregmagolan: I think this will end up using the locally installed yarn now which may be out of date

.circleci/env.sh
gregmagolan: Ahh. I see you hard-coded it there instead 👍 👍

aio/karma.conf.js
gkalpak: Yeah, exactly. I still needed to pass some flags for the `puppeteer`-provided Chrome to work locally (at least on Windows) and I realized we hardly ever need the browser. For debugging you can always open a browser and point it to the Karma URL (or change this config temporarily as you mentioned).

scripts/webdriver-manager-update.js
gkalpak: Yarn is clever enough to look at parent directories and find the closest `.yarnrc` (if any). So, it will use our vendored version regardless of the directory.

If this turns out to be a problem for bazel, we can use `require.resolve()` to find the relative path to the closest `node_modules/` directory (but `yarn` was easier for now and from what I understand it might work in bazel too).

gregmagolan: That's true for .yarnrc as I had to sift through that code to get yarn_install to honor .yarnrc and the `yarn-path`. However, there is special treatment for the `yarn-path` setting as it is a start-up setting; it may only if it is in a `.yarnrc` in the cwd as otherwise you can get into infinite recursion situations. You may be right however but should verify that is the case.

josephperrott: Confirmed that it picks up the version from `.yarnrc` as desired.

gregmagolan: 👍 Thanks @josephperrott

gregmagolan (approved):
LGTM. Will be some conflict with https://github.com/angular/angular/pull/33927 but this will probably land first tho so I'll handle the conflicts on rebase.

gregmagolan:
After https://github.com/angular/angular/pull/33927 and this land I should be able to run aio in an npm_integration_test :)

josephperrott (approved):
LGTM

Issue #24571, opened 2018-06-18
Post strictPropertyInitialization flag flip cleanup


rkirov:
## I'm submitting a...

[x] Other... Please describe: Tracking issue for internal cleanup.
</code></pre>

## What needs to be done

In order to quickly turn on strictPropertyInitialization flag in the whole code base, we introduced `!` on every non-initialized class field. Each one of those fields needs to be individually checked and decided whether:

- the field should be initialized. Thus `foo!: string`, should become `foo = 'default'`.
- the field should be marked as optional. Thus `foo!: string`, should become `foo?: string`. 
- the field is reliably initialized, but TS cannot infer that (initialization can occur outside the ctor, for example), so `!` should be kept.

alan-agius4:
@rkirov when an input is required, would you mark it as optional `?`? As in my option in you don't know that the consumer will provide that value, that thus, a component should be doing check if that value has been actually provided.

And seeing the comment from @mhevery it seems that we are on the same page;

> Nothing forces @Input() to be set. A user of the your component may simply chose not to bind to that input in which case your @Input() is not initialize and you should have checks for non-initialized case. I know we disagree, but I think this is the correct behavior.

https://github.com/angular/angular/issues/21571

alan-agius4:
@rkirov when an input is required, would you mark with a band `!`? As in my option in you don't know that the consumer will provide that value, thus, a component should be doing check if that value has been actually provided. `!` should only be used if the value is initialised such as on `ngOnInit`. that said though, if one used that component using the `viewChild` etc... the property will be marked that it will never be `undefined`. Which in some cases it might be wrong, based on which lifecycle hook the value is initialized.

And seeing the comment from @mhevery in https://github.com/angular/angular/issues/21571 it seems that we are on the same page;

> Nothing forces @Input() to be set. A user of the your component may simply chose not to bind to that input in which case your @Input() is not initialize and you should have checks for non-initialized case. I know we disagree, but I think this is the correct behavior.

Truth to the told I am not a particular fan of using the `!`

alan-agius4:
Thanks for the detailed explaination.

a-deeb:
[...]

PR #35393, opened 2020-02-12
feat(router): add skipNavigation option to routerLink

### LINKED ISSUE #13980: routerLink and `null` value should remove `href` attribute from link
```
[ ] bug report
[ x ] feature request
[ ] support request
```

Currently null or empty value passed to `routerLink` adds current page `href` value to link that is not useful. There is a lot of usecases for 'disabled' links that should not navigate anywhere and should not has `href` attribute at all.
It would be useful to have null/empty value support by `routerLink` that will results with no `href` attribute on link at all to prevent any navigation/link copying.

### LINKED ISSUE #31154: Add 'skipNavigation' boolean to routerLink to disable navigating
# 🚀 feature request

### Relevant Package
This feature request is for `@angular/router` (specifically the `routerLink` directive)

### Description
At the moment it doesn't seem possible to combine the power of `routerLinkActive` (for class changes) with a `routerLink` that doesn't navigate.  
This can be useful for instance when you have a navigation of multiple levels, and only the child levels are allowed to actually navigate (but you still want to leverage the `routerLinkActive` directive to add/remove classes on the links that shouldn't navigate.

* Add `skipNavigation` to the `routerLink` directive to block the actual navigation of the `onClick` handler.

``` typescript
  @HostListener('click', ['$event.button', '$event.ctrlKey', '$event.metaKey', '$event.shiftKey'])
  onClick(button: number, ctrlKey: boolean, metaKey: boolean, shiftKey: boolean): boolean {
    if (button !== 0 || ctrlKey || metaKey || shiftKey) {
      return true;
    }

    if (typeof this.target === 'string' && this.target !== '_self') {
      return true;
    }

    const extras = {
      skipLocationChange: attrBoolValue(this.skipLocationChange),
      replaceUrl: attrBoolValue(this.replaceUrl),
      state: this.state
    };
+    if(!attrBoolValue(this.skipNavigation)) {
+      this.router.navigateByUrl(this.urlTree, extras);
+    }
    return false;
  }
```

* making the `routerLink` of the non-navigating links empty (or /) doesn't help, since `routerLinkActive` will match all of them when using `{ exact: false }`.
* using `{exact: true}` works for not matching all the root levels, but this will in turn never actually show a clear view of what links are currently active based on the URL (current link + all it's parents should be set to active, not only the clicked link).
* good practice would be using a guard, but for this specific situation it seems to be quite the overkill when it can just be fixed with a boolean that is false by default (so doesn't impact anything but the actual case).

### Related Issues
#10431
#22588
#13980

alexduer:
The RouterLink directive now has a 'skipNavigation' boolean which disables navigating.

Closes #31154
Closes #13980

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: 31154

- [ ] Yes
- [x] No

## Other information
I could not find documentation for the 'skipLocationChange' boolean, so I did not add any documentation for the 'skipNavigation' boolean either.

### DIFF
--- a/packages/router/src/directives/router_link.ts
+++ b/packages/router/src/directives/router_link.ts
@@ -124,6 +124,8 @@ export class RouterLink {
   // TODO(issue/24571): remove '!'.
   @Input() skipLocationChange !: boolean;
   // TODO(issue/24571): remove '!'.
+  @Input() disableNavigation !: boolean;
+  // TODO(issue/24571): remove '!'.
   @Input() replaceUrl !: boolean;
   @Input() state?: {[k: string]: any};
   private commands: any[] = [];
@@ -160,6 +162,10 @@ export class RouterLink {
 
   @HostListener('click')
   onClick(): boolean {
+    if (attrBoolValue(this.disableNavigation)) {
+      return true;
+    }
+
     const extras = {
       skipLocationChange: attrBoolValue(this.skipLocationChange),
       replaceUrl: attrBoolValue(this.replaceUrl),
@@ -206,6 +212,8 @@ export class RouterLinkWithHref implements OnChanges, OnDestroy {
   // TODO(issue/24571): remove '!'.
   @Input() skipLocationChange !: boolean;
   // TODO(issue/24571): remove '!'.
+  @Input() disableNavigation !: boolean;
+  // TODO(issue/24571): remove '!'.
   @Input() replaceUrl !: boolean;
   @Input() state?: {[k: string]: any};
   private commands: any[] = [];
@@ -249,6 +257,10 @@ export class RouterLinkWithHref implements OnChanges, OnDestroy {
 
   @HostListener('click', ['$event.button', '$event.ctrlKey', '$event.metaKey', '$event.shiftKey'])
   onClick(button: number, ctrlKey: boolean, metaKey: boolean, shiftKey: boolean): boolean {
+    if (attrBoolValue(this.disableNavigation)) {
+      return false;
+    }
+
     if (button !== 0 || ctrlKey || metaKey || shiftKey) {
       return true;
     }
--- a/packages/router/test/integration.spec.ts
+++ b/packages/router/test/integration.spec.ts
@@ -1758,6 +1758,37 @@ describe('Integration', () => {
          expect(location.path()).toEqual('/team/22');
        })));
 
+    it('should support skipping navigation for anchor router links',
+       fakeAsync(inject([Router, Location], (router: Router, location: Location) => {
+         const fixture = TestBed.createComponent(RootCmp);
+         advance(fixture);
+
+         router.resetConfig([{path: 'team/:id', component: NoNavCmp}]);
+
+         router.navigateByUrl('/team/22');
+         advance(fixture);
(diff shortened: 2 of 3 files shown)


packages/router/src/directives/router_link.ts
```diff
@@ -164,7 +166,9 @@ export class RouterLink {
       skipLocationChange: attrBoolValue(this.skipLocationChange),
       replaceUrl: attrBoolValue(this.replaceUrl),
     };
-    this.router.navigateByUrl(this.urlTree, extras);
+    if (!attrBoolValue(this.skipNavigation)) {
```
atscott: I think this would be better as an early return at the top like what's done with the `onClick` of `RouterLinkWithHref`

packages/router/src/directives/router_link.ts
```diff
@@ -262,7 +268,9 @@ export class RouterLinkWithHref implements OnChanges, OnDestroy {
       replaceUrl: attrBoolValue(this.replaceUrl),
       state: this.state
     };
-    this.router.navigateByUrl(this.urlTree, extras);
+    if (!attrBoolValue(this.skipNavigation)) {
```
atscott: Move this to be an early return at the top with the others. Additionally, I think you should be returning `true` instead of `false` if you're not navigating.

packages/router/src/directives/router_link.ts
```diff
@@ -124,6 +124,8 @@ export class RouterLink {
   // TODO(issue/24571): remove '!'.
   @Input() skipLocationChange !: boolean;
   // TODO(issue/24571): remove '!'.
+  @Input() skipNavigation !: boolean;
```
atscott: I'm not sure how I feel about `skipNavigation` since navigating is kind of the only thing a `routerLink` *does* in the first place. Maybe `disabled` or `disableNavigation` would be a more appropriate name. What do you think?

atscott (changes requested):

packages/router/src/directives/router_link.ts
alexduer: Good point! 'disableNavigation' is definitely clearer.

atscott:
Can you take a look at the CI failures and try to fix them? Also adding @gkalpak for some input on how this should be documented, if merged.

packages/router/src/directives/router_link.ts
```diff
@@ -124,6 +124,8 @@ export class RouterLink {
   // TODO(issue/24571): remove '!'.
   @Input() skipLocationChange !: boolean;
   // TODO(issue/24571): remove '!'.
+  @Input() disableNavigation !: boolean;
```
atscott: Rather than continuing to use `!` and adding to the technical debt for #24571, how about using `?` since Inputs are optional?

jdgarvey:
@alexduer are you still working on this? If you are that's cool, but if you aren't I can pick this up and get it through the rest of the process.

alexduer:
@jdgarvey you are welcome to pick up the issue

atscott:
Closing, as it looks like the work was handed off and is now in #39916

PR #35396, opened 2020-02-12
ci: reenable draft mode conditioning for pullapprove

josephperrott:
Previously there was a regression in PullApprove which preventing draft mode
from being respected correctly for PullApprove processing.  This regression
has been remedied and we should be able to once again respect draft mode for
PRs.


IgorMinar (approved):
Lgtm

PR #35428, opened 2020-02-13
(Draft): Proof of concept - transplanted view CD, unified counter

atscott:


packages/core/src/render3/node_manipulation.ts
```diff
@@ -6,10 +6,13 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
+import {getLView} from '@angular/core/src/render3/state';
```
pkozlowski-opensource: I guess it should be a relative import?

atscott:
[Presubmit](https://test.corp.google.com/ui#id=OCL:297858477:BASE:298967882:1583367699268:134f34a8)

PR #35436, opened 2020-02-14
Incremental, multi-file template type-checking!

alxhub:


packages/compiler-cli/src/ngtsc/typecheck/src/context.ts
```diff
@@ -75,7 +84,34 @@ export class TypeCheckContext {
...
+      this.files.set(sf, record);
+    } else {
+      const rec = this.files.get(sf) !;
+      if (rec.type === 'final') {
+        throw new Error(
+            `AssertionError: cannot add type-checking code for template of ${ref.debugName} (${sf.fileName}) to reused type-checking file ${rec.file.fileName}.`);
+      }
+
+      record = rec;
+    }
```
JoostK: nit: consider extracting loading the record into a private method, that would make it easier to read.

packages/compiler-cli/src/ngtsc/typecheck/src/shim.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  static forRootFiles(files: ReadonlyArray<AbsoluteFsPath>, oldProgram: ts.Program|null):
+      TypeCheckShimGenerator {
+    const typeCheckPaths = new Map<AbsoluteFsPath, AbsoluteFsPath>();
+    for (const file of files) {
+      if (!isNonDeclarationTsPath(file)) {
+        continue;
+      }
+
+      typeCheckPaths.set(file, file.replace(/\.ts$/, '.__ng_typecheck__.ts') as AbsoluteFsPath);
```
JoostK: `isNonDeclarationTsPath` supports `.tsx` extensions, which would not be properly replaced here.

packages/compiler-cli/test/ngtsc/incremental_spec.ts
```diff
@@ -393,6 +448,263 @@ runInEachFileSystem(() => {
...
+      });
+
+      it('should recompile when a remote change happens to a scope', () => {
+        // The premise of this test is that the component Cmp has a template error (a binding to an
+        // unknown property). Cmp is in ModuleA, which imports ModuleB, which declares Dir that has
+        // the property. Because ModuleB doesn't export Dir, it's not visible to Cmp - hence the
+        // error.
+        // In the test, during the incremental rebuild Dir is exported from ModuleB. This is a
+        // change to the scope of ModuleA made by changing ModuleB (hence, a "remote change"). The
+        // test verifies that incremental template type-checking.
```
JoostK: Incomplete comment

packages/compiler-cli/test/ngtsc/incremental_spec.ts
```diff
@@ -393,6 +448,263 @@ runInEachFileSystem(() => {
...
+
+            export interface Keys {
+              alpha: string;
+              beta: string;
+            }
+
+            @Directive({
+              selector: '[dir]'
+            })
+            export class Dir<T extends keyof Keys> {
```
JoostK: Ha, the `keyof` constraint would actually be fairly simple to start supporting in the `TypeEmitter` so that an inline type ctor wouldn't be necessary anymore. A safer alternative to force an inline ctor would be not to export `Keys`.

JoostK (approved):
LGTM, just some some minor comments. And CI appears to fail for legitimate reasons.

It appears that the work was inadvertently squashed into a single commit, hopefully this can be restored.

alxhub:
Closing this as it turns out the solution doesn't work in practice. Hidden in the implementation is an assumption that the `rootFiles` passed into the `CompilerHost` contains all the TS files in the program. This is true in our nicely defined unit testing environments, but not in real projects where TS files are discovered via imports, etc.

This will warrant some additional thinking.

PR #33787, opened 2019-11-13
fix(ivy): avoid using stale cache in TestBed if module overrides are defined

AndrewKushnir:
NgModule compilation in JIT mode (that is also used in TestBed) caches module scopes on NgModule defs (using `transitiveCompileScopes` field). Module overrides (defined via TestBed.overrideModule) may invalidate this data by adding/removing items in `declarations` list. This commit forces TestBed to recalculate transitive scopes in case module overrides are present, so TestBed always gets the most up-to-date information.

This PR resolves #33735.

## PR Type
What kind of change does this PR introduce?

- [x] Bugfix

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

PR #35454, opened 2020-02-14
perf(core): avoid recursive scope recalculation when TestBed.overrideModule is used

### LINKED ISSUE #35395: Unit Tests - Overriding large modules slows down createComponent
# 🐞 bug report

### Affected Package

### Is this a regression?

This is a regression from:

https://github.com/angular/angular/pull/33787
https://github.com/angular/angular/issues/33735

### Description

The angular style guide recommends using shared modules for re-usable features.
https://angular.io/guide/styleguide#shared-feature-module

As a result, in our production app, we have very large shared modules containing many components, pipes, services, directives, common imports for angular material, etc.

When testing one of our feature component outside the shared module, it takes 8-9 seconds to create the component when we use overrideModule.

Our use case is as follows.

- Lazy Loaded Module
-- Module 1
-- Module 2 - imports module 1, shared module, declares feature component, but doesn't export it
--- Feature Component Being Tested - spec file imports module 2, overrides module 2 to export feature component
- Shared Modules
-- contains lots of components, service, pipes, etc...

Our spec file contains 21 unit tests and these are the times for each test being run.

This is logging just the createComponent function.

image

This is performance tab for just 1 unit test.
image

## 🔬 Minimal Reproduction
We are unable to share our production application, however I have a minimal reproduction, the scope is smaller, but still represents the slow down factor even for a small shared module.

The order of the tests being run and the component caches affects the times, but this shows with both the fast test running first and the slow test running first.

image

image

https://github.com/ChristopherKiss/angular-unit-test-repo

* run `npm i`
* run `ng test`
* open console and note the timings
* refresh to run tests a few times

For reference the tests are being run from `src/app/test/test/test.component.spec.ts`

## 🔥 Exception or Error
<pre><code>

</code></pre>

## 🌍  Your Environment

**Angular Version:**
<pre><code>

                |___/
    

Angular CLI: 9.0.1
Node: 12.13.1
OS: win32 x64

Angular: 9.0.0
... animations, cdk, common, compiler, compiler-cli, core, forms
... language-service, material, platform-browser
Ivy Workspace: Yes

@angular-devkit/architect         0.900.1
@angular-devkit/build-angular     0.900.1
@angular-devkit/build-optimizer   0.900.1
@angular-devkit/build-webpack     0.900.1
@angular-devkit/core              9.0.1
@angular-devkit/schematics        9.0.1
@angular/cli                      9.0.1
@angular/flex-layout              9.0.0-beta.29
@ngtools/webpack                  9.0.1
@schematics/angular               9.0.1
@schematics/update                0.900.1
</code></pre>

**Anything else relevant?**

Our shared module components make heavy use of angular material components in them.

AndrewKushnir:
Currently if TestBed detects that TestBed.overrideModule was used, transitive scopes are recalculated recursively for all modules and previously calculated data (stored in cache) is ignored. This behavior was introduced in https://github.com/angular/angular/pull/33787 and causes performance issues in case modules dependency tree is large. This commit updates the logic to avoid unnecessary recursive recalculation in nested modules.

Fixes #35395.

## PR Type

- [x] Other... Please describe: perf improvement for Ivy TestBed.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/testing/src/r3_test_bed_compiler.ts
+++ b/packages/core/testing/src/r3_test_bed_compiler.ts
@@ -355,8 +355,11 @@ export class R3TestBedCompiler {
             // are present, always re-calculate transitive scopes to have the most up-to-date
             // information available. The `moduleToScope` map avoids repeated re-calculation of
             // scopes for the same module.
-            const forceRecalc = !isTestingModule && this.hasModuleOverrides;
-            moduleToScope.set(moduleType, transitiveScopesFor(realType, forceRecalc));
+            if (!isTestingModule && this.hasModuleOverrides) {
+              this.storeFieldOfDefOnType(moduleType as any, NG_MOD_DEF, 'transitiveCompileScopes');
+              (moduleType as any)[NG_MOD_DEF].transitiveCompileScopes = null;
+            }
+            moduleToScope.set(moduleType, transitiveScopesFor(realType));
           }
           return moduleToScope.get(moduleType) !;
         };
--- a/packages/core/src/render3/jit/module.ts
+++ b/packages/core/src/render3/jit/module.ts
@@ -427,25 +427,19 @@ export function patchComponentDefWithScope<C>(
 /**
  * Compute the pair of transitive scopes (compilation scope and exported scope) for a given module.
  *
- * By default this operation is memoized and the result is cached on the module's definition. You
- * can avoid memoization and previously stored results (if available) by providing the second
- * argument with the `true` value (forcing transitive scopes recalculation).
- *
- * This function can be called on modules with components that have not fully compiled yet, but the
- * result should not be used until they have.
+ * This operation is memoized and the result is cached on the module's definition. This function can
+ * be called on modules with components that have not fully compiled yet, but the result should not
+ * be used until they have.
  *
  * @param moduleType module that transitive scope should be calculated for.
- * @param forceRecalc flag that indicates whether previously calculated and memoized values should
- * be ignored and transitive scope to be fully recalculated.
  */
-export function transitiveScopesFor<T>(
-    moduleType: Type<T>, forceRecalc: boolean = false): NgModuleTransitiveScopes {
+export function transitiveScopesFor<T>(moduleType: Type<T>): NgModuleTransitiveScopes {
   if (!isNgModule(moduleType)) {
     throw new Error(`${moduleType.name} does not have a module def (ɵmod property)`);
   }
   const def = getNgModuleDef(moduleType) !;
 
-  if (!forceRecalc && def.transitiveCompileScopes !== null) {
+  if (def.transitiveCompileScopes !== null) {
     return def.transitiveCompileScopes;
   }
 
@@ -486,7 +480,7 @@ export function transitiveScopesFor<T>(
 
     // When this module imports another, the imported module's exported directives and pipes are
     // added to the compilation scope of this module.
-    const importedScope = transitiveScopesFor(importedType, forceRecalc);
+    const importedScope = transitiveScopesFor(importedType);
     importedScope.exported.directives.forEach(entry => scopes.compilation.directives.add(entry));
     importedScope.exported.pipes.forEach(entry => scopes.compilation.pipes.add(entry));
   });
@@ -505,7 +499,7 @@ export function transitiveScopesFor<T>(
     if (isNgModule(exportedType)) {
       // When this module exports another, the exported module's exported directives and pipes are
       // added to both the compilation and exported scopes of this module.
-      const exportedScope = transitiveScopesFor(exportedType, forceRecalc);
+      const exportedScope = transitiveScopesFor(exportedType);
       exportedScope.exported.directives.forEach(entry => {
         scopes.compilation.directives.add(entry);
         scopes.exported.directives.add(entry);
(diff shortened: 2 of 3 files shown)


packages/core/testing/src/r3_test_bed_compiler.ts
```diff
@@ -355,8 +355,11 @@ export class R3TestBedCompiler {
             // are present, always re-calculate transitive scopes to have the most up-to-date
             // information available. The `moduleToScope` map avoids repeated re-calculation of
             // scopes for the same module.
-            const forceRecalc = !isTestingModule && this.hasModuleOverrides;
-            moduleToScope.set(moduleType, transitiveScopesFor(realType, forceRecalc));
+            if (!isTestingModule && this.hasModuleOverrides) {
+              this.storeFieldOfDefOnType(moduleType as any, NG_MOD_DEF, 'transitiveCompileScopes');
+              (moduleType as any)[NG_MOD_DEF].transitiveCompileScopes = null;
```
IgorMinar: this cast to `any` in order to reset the memized values is a code smell. Why do you think that this is the proper way to deal with this scenario? The explicit `forceRecalc` flag api was self-documenting and intuitive, why not keep it and ensure that we pass it in only when needed.

I don't quite follow how this change decreases the number of forced recals. Can you please explain? thanks!

alxhub: I'm having a hard time understanding it too.

I _think_ the substantial change is that by calling `transitiveScopesFor()` without setting `forceRecalc = true`, the resulting scopes are memoized on the module, and later calls to `transitiveScopesFor()` by e.g. `compileComponent` don't cause a recalculation as a result.

To me, this indicates that the `if` guard around https://github.com/angular/angular/pull/35454/files#diff-fd48f1abc4e7fe99ac0dbb7df845ff45L524-L526 may be undesirable - do we _always_ want to memoize even forced recalculations? That to me has the same effect as clearing the memoization before calling `transitiveScopesFor`.

alxhub:
Additionally, I think the commit message is a bit sparse. It should explain in detail under what circumstances unnecessary recalculation happens, and what's being changed to eliminate it.

alxhub (changes requested):

alxhub:
Oh, weird, Github posted my comment twice.

packages/core/testing/src/r3_test_bed_compiler.ts
AndrewKushnir: Thanks for review @IgorMinar and @alxhub!

The problem was caused by the fact that the `forceRecalc` flag was passed to recursive `transitiveScopesFor` calls, forcing each call to ignore memoized values (so that `transitiveScopesFor` calls become "pure" and avoid the need to store/restore original `transitiveCompileScopes` value). As a result, for complex/deep module hierarchies that results in perf regression and in some cases excessive recalculation of scopes for the same modules (for example if a few modules in the hierarchy depend on Angular Material).

However it looks like we just need to recalc transitive scopes only for a module that has overrides (for ex. in case `declarations` field is updated) and keep the rest of the modules (and memoized scopes) untouched. So this PR does the following:

- stores current value of the `transitiveScopesFor` field (so that we can restore it back to original value after the test)
- clears `transitiveScopesFor` field value to trigger scopes recalculation

Since `transitiveScopesFor` function is used in JIT, I think having TestBed-specific flags there might not be ideal and we can just operate with ModuleDef fields instead (similar to how we patch Component and Directive defs in case of provider overrides).

alxhub (approved):

alxhub:
Per offline discussion:

1) the logic in this PR is correct: there is no need to clear scopes of any children modules of an overridden module, as they cannot be affected by the override. This is what the change accomplishes.
2) the perf issue comes from a "diamond" problem, where module X is overridden which imports modules A and B, which both import module C. Under previous logic, module C gets its transitive deps recomputed multiple times, during the recompute for both A and B. For deep graphs this can be super costly.
3) @AndrewKushnir will update the commit message to more thoroughly describe the perf issue, how it occurs under the current logic, and how the new logic fixes it.
4) we might have _another_ bug, where if module A imports B and B has overrides, module A's transitive scopes need to be recomputed as well. It's unclear if `R3TestBed` currently does this - @AndrewKushnir will write a test and see how we do on this case, but we should land this PR regardless.

Approved!

AndrewKushnir:
[Presubmit](http://test/OCL:296530260:BASE:296530349:1582329799537:4015ce2d)

AndrewKushnir:
FYI, Blueprint and Global TAP runs are successful, so marking this PR with "merge" label.

AndrewKushnir:
MERGE ASSISTANCE: this PR is good to go in it's current state (see https://github.com/angular/angular/pull/35454#issuecomment-589427985), followup actions will be performed outside of the scope of this PR. Thank you.

AndrewKushnir:
> 4. we might have another bug, where if module A imports B and B has overrides, module A's transitive scopes need to be recomputed as well. It's unclear if R3TestBed currently does this - @AndrewKushnir will write a test and see how we do on this case, but we should land this PR regardless.

@alxhub I investigated this scenario and it looks like it works as expected and we already [have a test that proves it](https://github.com/angular/angular/blob/master/packages/core/test/test_bed_spec.ts#L292-L343). Thank you.

PR #35462, opened 2020-02-14
Two template type-checking fixes

alxhub:


packages/compiler-cli/src/ngtsc/typecheck/src/expression.ts
```diff
@@ -219,39 +224,103 @@ class AstTranslator implements AstVisitor {
...
+ * In View Engine's template type-checker, certain receivers of safe navigation operations will
+ * cause a temporary variable to be allocated as part of the checking expression, to save the value
+ * of the receiver and use it more than once in the expression. This temporary variable has type
+ * 'any'. In practice, this means certain receivers cause View Engine to not check the full
+ * expression, and other receivers will receive more complete checking.
+ *
+ * For compatibility, this logic is adapted from View Engine's expression_converter.ts so that the
+ * Ivy checker can emulate this bug when needed.
+ */
+class VeSafeLhsInferenceBugDetector implements AstVisitor {
```
JoostK: 😰

packages/compiler-cli/src/ngtsc/typecheck/src/expression.ts
```diff
@@ -219,39 +224,103 @@ class AstTranslator implements AstVisitor {
...
+  private static SINGLETON = new VeSafeLhsInferenceBugDetector();
+
+  static veWillInferAnyFor(ast: SafeMethodCall|SafePropertyRead) {
```
JoostK: If this only accepts `SafeMethodCall|SafePropertyRead`, wouldn't the result of visiting the node just call into `visitSafeMethodCall`/`visitSafePropertyRead` which return `true` and `false` respectively? I.e. the visitor doesn't do a lot here.

packages/compiler-cli/src/ngtsc/typecheck/test/type_check_block_spec.ts
```diff
@@ -557,19 +558,21 @@ describe('type check blocks', () => {
     });
 
     describe('config.strictSafeNavigationTypes', () => {
-      const TEMPLATE = `{{a?.b}} {{a?.method()}}`;
+      const TEMPLATE = `{{a?.b}} {{a?.method()}} {{a.method()?.b}}`;
```
JoostK: nit: there's no tests for the VE bug output here

packages/compiler-cli/src/ngtsc/core/api.ts
```diff
@@ -278,6 +278,14 @@ export interface StrictTemplateOptions {
...
    */
   strictContextGenerics?: boolean;
+
+  /**
+   * Whether object or array literals defined in templates use their inferred type, or are
+   * interpreted as `any`.
+   *
+   * Defaults to `false` unless `fullTemplateTypeCheck` or `strictTemplates` are set.
+   */
+  strictLiteralTypes?: boolean;
```
JoostK: I'm not sure how useful this particular option would be for users. For g3 compat I understand that setting this to `false` is desirable, but I don't expect g3 to set these individual properties at all?

JoostK (changes requested):
I am a bit confused by the VE bug detector logic, there appears to be a bug there 😛

packages/compiler-cli/src/ngtsc/typecheck/src/expression.ts
alxhub: Ah, you're right. I just refactored this into a class, and it should be `ast.receiver.visit(...)`.

packages/compiler-cli/src/ngtsc/typecheck/test/type_check_block_spec.ts
alxhub: There should be. The `a.method()?.b` expression with `strictSafeNavigationTypes: false` gets translated to:

```typescript
((((ctx).a).method()), (null as any)).b
```

Note that here, the `.b` is applied to the result of `(...receiver expr, (null as any))` which will be of type `any`. This is the VE bug - because the receiver will be assigned to a temp variable, it will be treated as type `any` in the resulting expression. This contrasts with the input `a?.b`, which is translated to:

```typescript
(((ctx).a)!.b, (null as any))
```

Here the `.b` happens on the non-null asserted value `ctx.a!`, which will have its true type, and so it will be checked that `ctx.a!.b` is a legal property access given the type of `ctx.a!`, even though the result of the _overall_ access will be `any` because of the way VE writes its type-check blocks.

packages/compiler-cli/src/ngtsc/core/api.ts
alxhub: It's like all of the other options we have - it's useful as you're migrating your application onto stricter type-checking. In g3 there are _lots_ of cases of templates which do things like:

```html
{{ {'invalid': 'Invalid Thing', 'valid': 'Valid Thing'}[validity] }}
```

We'll be using the individual options to migrate users progressively onto stricter checking in g3, just as application developers will do externally.

packages/compiler-cli/src/ngtsc/typecheck/src/expression.ts
alxhub: Fixed.

packages/compiler-cli/src/ngtsc/typecheck/src/expression.ts
alxhub: That was my reaction too 😂

packages/compiler-cli/src/ngtsc/typecheck/test/type_check_block_spec.ts
JoostK: Ooh, that is correct. But `a.method()?.b` is only the `SafePropertyRead`. For `SafeMethodCall` there's currently the `a?.method()` test, but its receiver does not require a temporary. So what I think is missing is something like `a()?.method()` to exercise VE bug behaviour for `SafeMethodCall`

packages/compiler-cli/src/ngtsc/core/api.ts
JoostK: That example is a good reason to have this option, indeed!

JoostK (approved):

Issue #31495, opened 2019-07-10
Custom decorators not working using AoT and Ivy


robisim74:
# 🐞 bug report

### Affected Package

 The issue is caused by package @angular/compiler-cli (I guess)

### Is this a regression?

 I had never tried AoT with Ivy

### Description

 I have a custom property decorator like this:
```
export function Language(): PropertyDecorator {
    function DecoratorFactory(target: any): void {
        const targetNgOnInit: () => void = target.ngOnInit;
        function ngOnInit(this: any): void {
            
            // Just for example
            console.log('onInit');

            if (targetNgOnInit) {
                targetNgOnInit.apply(this);
            }
        }
        target.ngOnInit = ngOnInit;
    }
    return DecoratorFactory;
}
```
used as:

```
@Language() lang: string;

ngOnInit() {
  // For aot
}
```

- With AoT compiler, _it works_ (it writes _onInit_ in console)
- If I enable Ivy and disable AoT, _it works_
- _If I enable Ivy and AoT, it doesn't work_

## 🔬 Minimal Reproduction

[angular-ivy-decorators](https://github.com/robisim74/angular-ivy-decorators)

## 🔥 Exception or Error
No errors during compilation or in console. The `ngOnInit `method in decorator function is ignored.

## 🌍  Your Environment

**Angular Version:**
<pre><code>
Angular CLI: 8.1.1
Node: 10.14.2
OS: win32 x64
Angular: 8.1.1
... animations, cli, common, compiler, compiler-cli, core, forms
... language-service, platform-browser, platform-browser-dynamic
... router

Package                           Version
-----------------------------------------------------------
@angular-devkit/architect         0.801.1
@angular-devkit/build-angular     0.801.1
@angular-devkit/build-optimizer   0.801.1
@angular-devkit/build-webpack     0.801.1
@angular-devkit/core              8.1.1
@angular-devkit/schematics        8.1.1
@ngtools/webpack                  8.1.1
@schematics/angular               8.1.1
@schematics/update                0.801.1
rxjs                              6.4.0
typescript                        3.4.5
webpack                           4.35.2
</code></pre>

**Anything else relevant?**
Tried without success also the _next_ version of Angular.

Thanks

JoostK:
[Version 8.0.4 of the CLI](https://github.com/angular/angular-cli/releases/tag/v8.0.4) removed `emitDecoratorMetadata` from `tsconfig.json` by default, your project also doesn't have it enabled. Could you try again with that option enabled?

[...]

PR #35464, opened 2020-02-15
fix(core): Allow modification of lifecycle hooks any time before bootstrap

### LINKED ISSUE #30497: Allow custom decorators that use lifecycle hooks in AoT - these are our version of hooks, but we can't use them
When writing a custom decorator, lifecycle methods you patch are removed in AoT because the compiler doesn't check that custom decorators are using them.

This results in an inability to create custom decorators with any meaningful component specific behaviour.

React uses hooks to help patch into setup/teardown of your functional components, decorators are Angulars form of hooks. It would be super useful to have this functionality in Angular.

mhevery:
## Overview

Currently we read lifecycle hooks eagerly during `ɵɵdefineComponent`. The result is that it is not possible to do any sort of meta-programing such as mixins or adding lifecycle hooks using custom decorators since any such code executes after `ɵɵdefineComponent` has extracted the lifecycle hooks from the prototype. Additionally the behavior is inconsistent between AOT and JIT mode. In JIT mode overriding lifecycle hooks is possible because the whole `ɵɵdefineComponent` is placed in getter which is executed lazily. This is because JIT mode must compile a template which can be specified as `templateURL` and those we are waiting for its resolution.

## Size tradeoffs
- `+` `ɵɵdefineComponent` becomes smaller as it no longer needs to copy lifecycle hooks from prototype to `ComponentDef`
- `-` `ɵɵNgOnChangesFeature` feature is now always included with the codebase as it is no longer tree shakable.

## Perf tradeoffs
Previously we have read lifecycle hooks from prototype in the `ɵɵdefineComponent` so that lifecycle hook access would be monomorphic. This decision was made before we had `T*` data structures. By not reading the lifecycle hooks we are moving the megamorhic read form `ɵɵdefineComponent` to instructions. However, the reads happen on `firstTemplatePass` only and are subsequently cached in the `T*` data structures. The result is that the overall performance should be same (or slightly better as the intermediate `ComponentDef` has been removed)

## Next steps
- [ ] Remove `ɵɵNgOnChangesFeature` from compiler. (It will no longer be a feature.)
- [ ] Discuss the future of `Features` as they hinder meta-programing.

Fix #30497

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/goldens/size-tracking/aio-payloads.json
+++ b/goldens/size-tracking/aio-payloads.json
@@ -12,7 +12,7 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 2987,
-        "main-es2015": 450883,
+        "main-es2015": 450301,
         "polyfills-es2015": 52630
       }
     }
--- a/goldens/size-tracking/integration-payloads.json
+++ b/goldens/size-tracking/integration-payloads.json
@@ -12,8 +12,8 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 1485,
-        "main-es2015": 16959,
-        "polyfills-es2015": 36938
+        "main-es2015": 17362,
+        "polyfills-es2015": 36657
       }
     }
   },
@@ -21,7 +21,7 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 1485,
-        "main-es2015": 147314,
+        "main-es2015": 147573,
         "polyfills-es2015": 36571
       }
     }
@@ -30,7 +30,7 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 1485,
-        "main-es2015": 135533,
+        "main-es2015": 136168,
         "polyfills-es2015": 37248
       }
     }
@@ -39,7 +39,7 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 2289,
-        "main-es2015": 246044,
+        "main-es2015": 245351,
         "polyfills-es2015": 36938,
         "5-es2015": 751
       }
--- a/packages/core/src/render3/features/ng_onchanges_feature.ts
+++ b/packages/core/src/render3/features/ng_onchanges_feature.ts
@@ -11,14 +11,6 @@ import {SimpleChange, SimpleChanges} from '../../interface/simple_change';
 import {EMPTY_OBJ} from '../empty';
 import {DirectiveDef, DirectiveDefFeature} from '../interfaces/definition';
 
-const PRIVATE_PREFIX = '__ngOnChanges_';
-
-type OnChangesExpando = OnChanges&{
-  __ngOnChanges_: SimpleChanges|null|undefined;
-  // tslint:disable-next-line:no-any Can hold any value
-  [key: string]: any;
-};
-
 /**
  * The NgOnChangesFeature decorates a component with support for the ngOnChanges
  * lifecycle hook, so it should be included in any component that implements
@@ -41,12 +33,15 @@ type OnChangesExpando = OnChanges&{
  *
  * @codeGenApi
  */
+export function ɵɵNgOnChangesFeature<T>(): DirectiveDefFeature {
+  return NgOnChangesFeatureImpl;
+}
 
-export function ɵɵNgOnChangesFeature<T>(definition: DirectiveDef<T>): void {
+export function NgOnChangesFeatureImpl<T>(definition: DirectiveDef<T>) {
   if (definition.type.prototype.ngOnChanges) {
     definition.setInput = ngOnChangesSetInput;
-    (definition as {onChanges: Function}).onChanges = wrapOnChanges();
   }
+  return rememberChangeHistoryAndInvokeOnChangesHook;
 }
 
 // This option ensures that the ngOnChanges lifecycle hook will be inherited
@@ -55,27 +50,36 @@ export function ɵɵNgOnChangesFeature<T>(definition: DirectiveDef<T>): void {
 // tslint:disable-next-line:no-toplevel-property-access
 (ɵɵNgOnChangesFeature as DirectiveDefFeature).ngInherit = true;
 
-function wrapOnChanges() {
-  return function wrapOnChangesHook_inPreviousChangesStorage(this: OnChanges) {
-    const simpleChangesStore = getSimpleChangesStore(this);
-    const current = simpleChangesStore && simpleChangesStore.current;
+/**
+ * This is a synthetic lifecycle hook which gets inserted into `TView.preOrderHooks` to simulate
+ * `ngOnChanges`.
+ *
+ * The hook reads the `NgSimpleChangesStore` data from the component instance and if changes are
+ * found it invokes `ngOnChanges` on the component instance.
+ *
+ * @param this Component instance. Because this function gets inserted into `TView.preOrderHooks`,
+ *     it is guaranteed to be called with component instance.
+ */
+function rememberChangeHistoryAndInvokeOnChangesHook(this: OnChanges) {
+  const simpleChangesStore = getSimpleChangesStore(this);
+  const current = simpleChangesStore?.current;
 
-    if (current) {
-      const previous = simpleChangesStore!.previous;
-      if (previous === EMPTY_OBJ) {
-        simpleChangesStore!.previous = current;
-      } else {
-        // New changes are copied to the previous store, so that we don't lose history for inputs
-        // which were not changed this time
-        for (let key in current) {
-          previous[key] = current[key];
-        }
+  if (current) {
+    const previous = simpleChangesStore!.previous;
+    if (previous === EMPTY_OBJ) {
+      simpleChangesStore!.previous = current;
+    } else {
+      // New changes are copied to the previous store, so that we don't lose history for inputs
+      // which were not changed this time
+      for (let key in current) {
+        previous[key] = current[key];
       }
-      simpleChangesStore!.current = null;
-      this.ngOnChanges(current);
     }
-  };
+    simpleChangesStore!.current = null;
+    this.ngOnChanges(current);
+  }
 }
+
 
 function ngOnChangesSetInput<T>(
     this: DirectiveDef<T>, instance: T, value: any, publicName: string, privateName: string): void {
@@ -102,6 +106,10 @@ function setSimpleChangesStore(instance: any, store: NgSimpleChangesStore): NgSi
   return instance[SIMPLE_CHANGES_STORE] = store;
 }
 
+/**
+ * Data structure which is monkey-patched on the component instance and used by `ngOnChanges`
+ * life-cycle hook to track previous input values.
+ */
 interface NgSimpleChangesStore {
   previous: SimpleChanges;
   current: SimpleChanges|null;
--- a/packages/core/src/render3/hooks.ts
+++ b/packages/core/src/render3/hooks.ts
@@ -6,9 +6,10 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
+import {AfterContentChecked, AfterContentInit, AfterViewChecked, AfterViewInit, DoCheck, OnChanges, OnDestroy, OnInit} from '../interface/lifecycle_hooks';
 import {assertEqual, assertNotEqual} from '../util/assert';
-
 import {assertFirstCreatePass} from './assert';
+import {NgOnChangesFeatureImpl} from './features/ng_onchanges_feature';
 import {DirectiveDef} from './interfaces/definition';
 import {TNode} from './interfaces/node';
 import {FLAGS, HookData, InitPhaseState, LView, LViewFlags, PREORDER_HOOK_FLAGS, PreOrderHookFlags, TView} from './interfaces/view';
@@ -31,20 +32,23 @@ import {getCheckNoChangesMode} from './state';
 export function registerPreOrderHooks(
     directiveIndex: number, directiveDef: DirectiveDef<any>, tView: TView): void {
   ngDevMode && assertFirstCreatePass(tView);
-  const {onChanges, onInit, doCheck} = directiveDef;
+  const {ngOnChanges, ngOnInit, ngDoCheck} =
+      directiveDef.type.prototype as OnChanges & OnInit & DoCheck;
 
-  if (onChanges) {
-    (tView.preOrderHooks || (tView.preOrderHooks = [])).push(directiveIndex, onChanges);
-    (tView.preOrderCheckHooks || (tView.preOrderCheckHooks = [])).push(directiveIndex, onChanges);
+  if (ngOnChanges as Function | undefined) {
+    const wrappedOnChanges = NgOnChangesFeatureImpl(directiveDef);
+    (tView.preOrderHooks || (tView.preOrderHooks = [])).push(directiveIndex, wrappedOnChanges);
+    (tView.preOrderCheckHooks || (tView.preOrderCheckHooks = []))
+        .push(directiveIndex, wrappedOnChanges);
   }
 
-  if (onInit) {
-    (tView.preOrderHooks || (tView.preOrderHooks = [])).push(-directiveIndex, onInit);
+  if (ngOnInit) {
+    (tView.preOrderHooks || (tView.preOrderHooks = [])).push(0 - directiveIndex, ngOnInit);
   }
 
-  if (doCheck) {
-    (tView.preOrderHooks || (tView.preOrderHooks = [])).push(directiveIndex, doCheck);
-    (tView.preOrderCheckHooks || (tView.preOrderCheckHooks = [])).push(directiveIndex, doCheck);
+  if (ngDoCheck) {
+    (tView.preOrderHooks || (tView.preOrderHooks = [])).push(directiveIndex, ngDoCheck);
+    (tView.preOrderCheckHooks || (tView.preOrderCheckHooks = [])).push(directiveIndex, ngDoCheck);
   }
 }
 
@@ -73,27 +77,36 @@ export function registerPostOrderHooks(tView: TView, tNode: TNode): void {
   // hooks for projected components and directives must be called *before* their hosts.
   for (let i = tNode.directiveStart, end = tNode.directiveEnd; i < end; i++) {
     const directiveDef = tView.data[i] as DirectiveDef<any>;
-    if (directiveDef.afterContentInit) {
-      (tView.contentHooks || (tView.contentHooks = [])).push(-i, directiveDef.afterContentInit);
+    const lifecycleHooks: AfterContentInit&AfterContentChecked&AfterViewInit&AfterViewChecked&
+        OnDestroy = directiveDef.type.prototype;
+    const {
+      ngAfterContentInit,
+      ngAfterContentChecked,
+      ngAfterViewInit,
+      ngAfterViewChecked,
+      ngOnDestroy
+    } = lifecycleHooks;
+
+    if (ngAfterContentInit) {
+      (tView.contentHooks || (tView.contentHooks = [])).push(-i, ngAfterContentInit);
     }
 
-    if (directiveDef.afterContentChecked) {
-      (tView.contentHooks || (tView.contentHooks = [])).push(i, directiveDef.afterContentChecked);
-      (tView.contentCheckHooks || (tView.contentCheckHooks = []))
-          .push(i, directiveDef.afterContentChecked);
+    if (ngAfterContentChecked) {
+      (tView.contentHooks || (tView.contentHooks = [])).push(i, ngAfterContentChecked);
+      (tView.contentCheckHooks || (tView.contentCheckHooks = [])).push(i, ngAfterContentChecked);
     }
 
-    if (directiveDef.afterViewInit) {
-      (tView.viewHooks || (tView.viewHooks = [])).push(-i, directiveDef.afterViewInit);
+    if (ngAfterViewInit) {
+      (tView.viewHooks || (tView.viewHooks = [])).push(-i, ngAfterViewInit);
     }
 
-    if (directiveDef.afterViewChecked) {
-      (tView.viewHooks || (tView.viewHooks = [])).push(i, directiveDef.afterViewChecked);
-      (tView.viewCheckHooks || (tView.viewCheckHooks = [])).push(i, directiveDef.afterViewChecked);
+    if (ngAfterViewChecked) {
+      (tView.viewHooks || (tView.viewHooks = [])).push(i, ngAfterViewChecked);
+      (tView.viewCheckHooks || (tView.viewCheckHooks = [])).push(i, ngAfterViewChecked);
     }
 
-    if (directiveDef.onDestroy != null) {
-      (tView.destroyHooks || (tView.destroyHooks = [])).push(i, directiveDef.onDestroy);
+    if (ngOnDestroy != null) {
+      (tView.destroyHooks || (tView.destroyHooks = [])).push(i, ngOnDestroy);
     }
   }
 }
--- a/packages/core/src/render3/instructions/shared.ts
+++ b/packages/core/src/render3/instructions/shared.ts
@@ -7,6 +7,7 @@
  */
 import {Injector} from '../../di';
 import {ErrorHandler} from '../../error_handler';
+import {DoCheck, OnChanges, OnInit} from '../../interface/lifecycle_hooks';
 import {CUSTOM_ELEMENTS_SCHEMA, NO_ERRORS_SCHEMA, SchemaMetadata} from '../../metadata/schema';
 import {ViewEncapsulation} from '../../metadata/view';
 import {validateAgainstEventAttributes, validateAgainstEventProperties} from '../../sanitization/sanitization';
@@ -1168,9 +1169,11 @@ export function resolveDirectives(
         if (def.hostBindings !== null || def.hostAttrs !== null || def.hostVars !== 0)
           tNode.flags |= TNodeFlags.hasHostBindings;
 
+        const lifeCycleHooks: OnChanges&OnInit&DoCheck = def.type.prototype;
         // Only push a node index into the preOrderHooks array if this is the first
         // pre-order hook found on this node.
-        if (!preOrderHooksFound && (def.onChanges || def.onInit || def.doCheck)) {
+        if (!preOrderHooksFound &&
+            (lifeCycleHooks.ngOnChanges || lifeCycleHooks.ngOnInit || lifeCycleHooks.ngDoCheck)) {
           // We will push the actual hook function into this array later during dir instantiation.
           // We cannot do it now because we must ensure hooks are registered in the same
           // order that directives are created (i.e. injection order).
@@ -1178,7 +1181,7 @@ export function resolveDirectives(
           preOrderHooksFound = true;
         }
 
-        if (!preOrderCheckHooksFound && (def.onChanges || def.doCheck)) {
+        if (!preOrderCheckHooksFound && (lifeCycleHooks.ngOnChanges || lifeCycleHooks.ngDoCheck)) {
           (tView.preOrderCheckHooks || (tView.preOrderCheckHooks = []))
               .push(tNode.index - HEADER_OFFSET);
           preOrderCheckHooksFound = true;
--- a/packages/core/test/acceptance/lifecycle_spec.ts
+++ b/packages/core/test/acceptance/lifecycle_spec.ts
@@ -1124,6 +1124,46 @@ describe('onChanges', () => {
   });
 });
 
+describe('meta-programing', () => {
+  it('should allow adding lifecycle hook methods any time before first instance creation', () => {
+    const events: any[] = [];
+
+    @Component({template: `<child name="value"></child>`})
+    class App {
+    }
+
+    @Component({selector: 'child', template: `empty`})
+    class Child {
+      @Input() name: string = '';
+    }
+
+    (Child.prototype as any).ngOnInit = () => events.push('onInit');
+    (Child.prototype as any).ngOnChanges = (e: SimpleChanges) => {
+      const name = e['name'];
+      expect(name.previousValue).toEqual(undefined);
+      expect(name.currentValue).toEqual('value');
+      expect(name.firstChange).toEqual(true);
+      events.push('ngOnChanges');
+    }, (Child.prototype as any).ngDoCheck = () => events.push('ngDoCheck');
+    (Child.prototype as any).ngAfterContentInit = () => events.push('ngAfterContentInit');
+    (Child.prototype as any).ngAfterContentChecked = () => events.push('ngAfterContentChecked');
+    (Child.prototype as any).ngAfterViewInit = () => events.push('ngAfterViewInit');
+    (Child.prototype as any).ngAfterViewChecked = () => events.push('ngAfterViewChecked');
+    (Child.prototype as any).ngOnDestroy = () => events.push('ngOnDestroy');
+
+    TestBed.configureTestingModule({
+      declarations: [App, Child],
+    });
+    const fixture = TestBed.createComponent(App);
+    fixture.detectChanges();
+    fixture.destroy();
+    expect(events).toEqual([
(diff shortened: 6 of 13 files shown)


bryanrideshark:
Thank you so much for fixing this. Angular Ivy is awesome - Being able to upgrade our app, which heavily uses custom decorators, will be possible once this PR is merged.

RolginRoman:
Really looking forward to this PR 👍

ef32:
Really looking forward to this PR being merged

qtis:
🙄

kumaresan-subramani:
Eagerly waiting for this MR to get merged. Once this get merged so many issue must get resolved i think.

lurock:
When using Angular 9.1.1 with Ivy now when I "ng serve" all of my custom class decorators are called and working the way they did with view engine but when I run "ng build --prod" and run the application only a few of my custom decorators are called.  @mhevery will this pull request fix that?  Meaning will all my decorators get called when the application is first loaded.

hakimio:
@mhevery any idea when this might get merged and what's stopping it from getting reviewed?
It's quite important issue for us too.

bryanrideshark:
@hakimio it seems like different members of the team have different opinions about how important Metaprogramming is for users of Angular.

Thankfully @mhevery  seems to realize just how important it is to us; but he might have a challenge convincing other members. Hopefully everyone on the Angular team will realize just how prevalent metaprogramming is among Angular users.

Maximaximum:
The merge delay is quite confusing. I really hope that this PR will get into master.

maxime1992:
Without this MR, Ivy misses a little bit to be backward compatible :) 

That'd really help us too at CloudNC to work on a new version of [ngx-sub-form](https://www.npmjs.com/package/ngx-sub-form) where we're trying to get rid of inheritance.

:crossed_fingers: for this to get merged soon

LayZeeDK:
Please don't kill [component features](https://indepth.dev/component-features-with-angular-ivy/).

Please stop worrying about custom decorators. Custom decorators are experimental and should be treated as such.

The original ECMAScript decorators proposal has been abandoned. There's a new proposal in early stages, but it has different syntax and semantics.

hakimio:
> Please don't kill [component features](https://indepth.dev/component-features-with-angular-ivy/).

AFAIK it's Angular private API and they can do whatever they want with private APIs.

> Custom decorators are experimental.

Yet they are one of the most widely used TypeScript features and many TypeScript libraries depend on them.

mlc-mlapis:
> AFAIK it's Angular private API and they can do whatever they want with private APIs.

It's private now, but because of the phase for stabilization inner structures and finalize the public API. It was always mentioned as the public API at some point.

> Yet they are one of the most widely used TypeScript features and many TypeScript libraries depend on them.

Yep, but it was meant from official ECMAScript spec.

klemenoslaj:
> Please stop worrying about custom decorators. Custom decorators are experimental and should be treated as such.

Why is everyone fixating on decorators only? The point is, that components are JS classes. Methods on classes are members of prototype. Overwriting a prototype method is an absolutely valid thing to do. Angular currently makes this impossible.

And yes, sure, decorators are one use case of it and I wouldn't make custom decorators for reasons you mentions, but there are many more use cases.
One use case is a `takeUntilDestroy` which is basically a `takeUntil` that patches `ngOnDestroy` hook in Angular. Impossible right now.

bryanrideshark:
If we can't use custom decorators, you might as well get rid of them across the board in Angular. No `@Input`, `@Output`, `@HostBinding`. I'm not going to keep using a framework if the authors of the framework take a "good for me, but not for thee" approach to something that's a key language feature. Angular originally was going to write its own language just to make decorators possible (Remember AtScript?) and now, because we've got an optimized compiler and some pet private APIs, we're going to say decorators aren't impossible or good for users to use?

It's not a good look when libraries are built around the idea that developers have to use a subset of a language.

pkozlowski-opensource:
Just FYI, the intention is to review and merge this PR. It is in progress.

bryanrideshark:
@klemenoslaj , you're absolutely right. Currently, we use a custom `takeUntil` implementation which (until Angular 9) added the ngOnDestroy method if it hadn't already been defined in the prototype. Now, I'm having to add a blank `ngOnDestroy()` method to all my components, just for the runtime to recognize that we're using the `ngOnDestroy()` hook. My compiled code has a bunch of useless `KlassComponent.prototype.ngOnDestroy = function() {}` code in it, just to convince the Angular runtime that it's actually using ngOnDestroy. Sure, maybe there's a minute performance boost in the startup time for the application, but my entire bundle has redundant code in it now, so am I really getting a win here?

bryanrideshark:
Thank you @pkozlowski-opensource  for the clarification. It's been a bit unclear as to whether or not this was going to land when you see people marked as "Contributor" sending mixed messages. I suppose "Contributor" just means you've committed code to the project, whereas "Members" have a decision making role inside Angular?

Regardless, I'm glad this is being looked at and addressed by the Angular team.

LayZeeDK:
>If we can't use custom decorators, you might as well get rid of them across the board in Angular. No @Input, @Output, @HostBinding. I'm not going to keep using a framework if the authors of the framework take a "good for me, but not for thee" approach to something that's a key language feature. Angular originally was going to write its own language just to make decorators possible (Remember AtScript?) and now, because we've got an optimized compiler and some pet private APIs, we're going to say decorators are off limits for end users?

Angular's TypeScript decorators are special in that they are compiler instructions, meaning they are annotations rather than decorators. They are converted to static properties at compile time, so they aren't used at runtime like most other custom decorators.

LayZeeDK:
>Thank you @pkozlowski-opensource for the clarification. It's been a bit unclear as to whether or not this was going to land when you see people marked as "Contributor" sending mixed messages. I suppose "Contributor" just means you've committed code to the project, whereas "Members" have a decision making role inside Angular?

I'm just voicing my opinion. Very few people are decision makers when it comes to Angular.

I think anyone's opinion should be valued, no matter whether they are an Angular team member, an Angular Collaborator, an Angular GDE, an Angular contributor, or any other form of Angular community member.

klemenoslaj:
@bryanrideshark since it was just confirmed that this MR is going to get done/merged, you can change your decorator to account for `ɵcmp`. Then you do not need to create an empty `ngOnDestroy` function.

I already use it and it works great, however, as stated above - by using private API. When this MR is merged, my private API usage can be removed too.

You can do something along these lines:

```
if (instanceConstructor.ɵcmp) {
  // Using private API until the corresponding Angular issues are resolved.
  // Issues:  https://github.com/angular/angular/issues/31495, https://github.com/angular/angular/issues/30497
  // PR:      https://github.com/angular/angular/pull/35464
  destroy = <Function>instanceConstructor.ɵcmp.onDestroy ?? noop;
  instanceConstructor.ɵcmp.onDestroy = __takeUntilDestroy__destroy__;
} else {
  destroy = <Function>instanceConstructor.prototype.ngOnDestroy ?? noop;
  instanceConstructor.prototype.ngOnDestroy = __takeUntilDestroy__destroy__;
}

function __takeUntilDestroy__destroy__() {
  destroy();
  // do something fancy
}
```

zakhenry:
For anyone that can't wait for this PR to close and be released, I've released a library that makes hooking into the angular lifecycle hooks a breeze. It does use internal APIs to get this working, and does require a class level decorator, but it works cleanly with ivy.

See https://github.com/cloudnc/ngx-observable-lifecycle

Even once this PR lands, I will continue to support this library; it will just get a bit simpler eliminating the need for implementers to use the class decorator

integration/_payload-limits.json
```diff
@@ -21,7 +21,7 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 1485,
-        "main-es2015": 147647,
+        "main-es2015": 154647,
```
kara, line 24: This is quite a big size jump.

kara (dismissed):
Can we add some tests to this PR illustrating what we're trying to fix?

bryanrideshark:
@kara the following snippet is the decorator code which no longer works for our team, along with a usage example.

### Decorator:
````ts
// NgUnsubscribe (Yes, we're terrible for using the Ng prefix)

import { Subject, Observable } from 'rxjs';

export interface NgUnsubscribe {
  readonly ngUnsubscribe: Observable<void>;
}

export function NgUnsubscribe() {
  return function(constructor) {
    const originalNgOnDestroy = constructor.prototype.ngOnDestroy;

    if (constructor.prototype.__hasSetupNgUnsubscribe) {
      // Handles errors that arise from using the decorator on inherited component classes.
      throw new Error(
        `${constructor.name} used @NgUnsubscribe already. Remove the decorator from ${constructor.name}`
      );
    }
    constructor.prototype.__hasSetupNgUnsubscribe = true;

    // Actually sets up the observable.
    Object.defineProperty(constructor.prototype, 'ngUnsubscribe', {
      get() {
        if (!this.__ngUnsubscribe) {
          this.__ngUnsubscribe = new Subject<void>();
        }
        return this.__ngUnsubscribe;
      }
    });

    // Do the unsubscription
    constructor.prototype.ngOnDestroy = function() {
      try {
        if (originalNgOnDestroy && typeof originalNgOnDestroy === 'function') {
          originalNgOnDestroy.apply(this, arguments);
        }
        if (this.__ngUnsubscribe) {
          const ngUnsubscribe = this.__ngUnsubscribe;
          ngUnsubscribe.next();
          ngUnsubscribe.complete();
          ngUnsubscribe.unsubscribe();
        }
      } catch {}
    };
  };
}


`````

### In Use:
````ts

@Injectable()
export class LogService {

   logNgUnsubscribeEmitted() {
   }

   logNgOnDestroyCalled() {
   }

}


@Component({
selector: 'foo-bar',
template: '',
styles: []
})
@NgUnsubscribe()
export class FooBarComponent implements OnInit, NgUnsubscribe {
   readonly ngUnsubscribe: Observable<void>;

   constructor(private _log: LogService) {}

   ngOnInit() {
       this.ngUnsubscribe.subscribe(() => this._log.logNgUnsubscribeEmitted());
   }

   ngOnDestroy() {
      this._log.logNgOnDestroyCalled();
   }
}



````

I'm no pro at writing tests, but the test would assert
 - That you can create and initialize FooBarComponent without errors
 - That when FooBarComponent is destroyed, both `logNgUnsubscribeEmitted` and `logNgOnDestroyCalled` have been called once.


That only uses `ngOnInit` and `ngOnDestroy`. To get a full sense of the usage, you could use the following decorator in a test environment.

````ts
export interface Mocks {
  onNgOnChanges: () => void;
  onNgOnInit: () => void;
  onNgOnDoCheck: () => void;
  onNgAfterContentInit: () => void;
  onNgAfterContentChecked: () => void;
  onNgAfterViewInit: () => void;
  onNgAfterViewChecked: () => void;
  onNgOnDestroy: () => void; 
}

const MOCKS: Mocks = null;

export function setTestMocks(mocks: Mocks) {
    MOCKS = mocks;
}


export function AugmentComponentHooks() {
   return function augmentComponentHooksDecorator(Original) {
    const originalNgOnChanges = Original.prototype.ngOnChanges;
    const originalNgOnInit = Original.prototype.ngOnInit;
    const originalNgOnDoCheck = Original.prototype.ngOnDoCheck;
    const originalNgAfterContentInit = Original.prototype.ngAfterContentInit;
    const originalNgAfterContentChecked = Original.prototype.ngAfterContentChecked;
    const originalNgAfterViewInit = Original.prototype.ngAfterViewInit;
    const originalNgAfterViewChecked = Original.prototype.ngAfterViewChecked;
    const originalNgOnDestroy = Original.prototype.ngOnDestroy;

    if (originalNgOnChanges) {
        Original.prototype.ngOnChanges = function(...args: Array<any>) {
            originalNgOnChanges.apply(this, arguments);
            MOCKS.onNgOnChanges();
        };
    }
    if (originalNgOnInit) {
        Original.prototype.ngOnInit = function(...args: Array<any>) {
            originalNgOnInit.apply(this, arguments);
            MOCKS.onNgOnInit();
        };
    }
    if (originalNgOnDoCheck) {
        Original.prototype.ngOnDoCheck = function(...args: Array<any>) {
            originalNgOnDoCheck.apply(this, arguments);
            MOCKS.onNgOnDoCheck();
        };
    }
    if (originalNgAfterContentInit) {
        Original.prototype.ngAfterContentInit = function(...args: Array<any>) {
            originalNgAfterContentInit.apply(this, arguments);
            MOCKS.onNgAfterContentInit();
        };
    }
    if (originalNgAfterContentChecked) {
        Original.prototype.ngAfterContentChecked = function(...args: Array<any>) {
            originalNgAfterContentChecked.apply(this, arguments);
            MOCKS.onNgAfterContentChecked();
        };
    }
    if (originalNgAfterViewInit) {
        Original.prototype.ngAfterViewInit = function(...args: Array<any>) {
            originalNgAfterViewInit.apply(this, arguments);
            MOCKS.onNgAfterViewInit();
        };
    }
    if (originalNgAfterViewChecked) {
        Original.prototype.ngAfterViewChecked = function(...args: Array<any>) {
            originalNgAfterViewChecked.apply(this, arguments);
            MOCKS.onNgAfterViewChecked();
        };
    }
    if (originalNgOnDestroy) {
        Original.prototype.ngOnDestroy = function(...args: Array<any>) {
            originalNgOnDestroy.apply(this, arguments);
            MOCKS.onNgOnDestroy();
        };
    }




   }
}


@Component({
    selector: 'uses-augmented-component',
    template: ``,
    styles: []
})
@AugmentComponentHooks()
export class UsesAugmentedComponent {

}

````

I realize that the team might want to implement another way of augmenting the lifecycle hooks - If there was another method, perhaps another compiler API we called in order to agument the hooks at runtime or compile time, that would be sufficient for me.

I don't think we really care about patching the prototype specifically - I think we care more about being able to augment the behavior using the convenient syntax Decorators provide.

File: integration/_payload-limits.json (hunk 1 above)
Comment (mhevery) [reply integration/_payload-limits.json]: Yes, that is because we always have to include `NgOnChangesFeature` as we don't know it will be added. Not sure we can do much about that.

**Comment by mhevery** [comment]
I have added tests, ready for another round of reviews.

**Comment by mhevery** [comment]
[presubmit](http://test/OCL:314996754:BASE:314996828:1591392505499:1e74d877)

File: packages/core/src/render3/features/ng_onchanges_feature.ts [hunk 2]
```diff
@@ -55,28 +50,27 @@ export function ɵɵNgOnChangesFeature<T>(definition: DirectiveDef<T>): void {
 // tslint:disable-next-line:no-toplevel-property-access
 (ɵɵNgOnChangesFeature as DirectiveDefFeature).ngInherit = true;
 
-function wrapOnChanges() {
-  return function wrapOnChangesHook_inPreviousChangesStorage(this: OnChanges) {
-    const simpleChangesStore = getSimpleChangesStore(this);
-    const current = simpleChangesStore && simpleChangesStore.current;
+function wrapOnChangesHook_inPreviousChangesStorage(this: OnChanges) {
```
IgorMinar, line 53: what is this function name trying to convey? could you please add a comment with docs and consider simplifying the function name? now that `wrapOnChanges` function is gone?

goldens/size-tracking/aio-payloads.json
```diff
@@ -12,7 +12,7 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 2987,
-        "main-es2015": 453386,
+        "main-es2015": 452712,
```
IgorMinar, line 15: overall this change is a size reduction for bigger apps even if for minimalistic apps this change causes a minor size regression.

that's a good news!

IgorMinar (changes requested):
This seems to be heading down the right path. Can you please resolve the comments and ensure that @pkozlowski-opensource also reviews the runtime part of this change? thanks!

IgorMinar:
the g3 failures seem to be unrelated to this change

IgorMinar:
@LayZeeDK I've read https://indepth.dev/component-features-with-angular-ivy/ I appreciate that you are exploring Angular truly in depth. It's cool what you can do with the pattern you documented, but I can't say strongly enough that all these features are private and highly unsable / experimental. These Ivy apis can change from one patch version to the next, so we don't advise anyone to rely on them. 

That being said, yet.. the flexibility is quite awesome and we do want to keep on making Angular better a more flexible to support these use-cases, it just takes a lot of time to do this right.

LayZeeDK:
@IgorMinar I think there's a sweet spot where Angular can allow consumers to hook into the component creation process or override the prototype while having access to the node injector without exposing the whole component definition or other internals/low level APIs.

I wanted to chime in because of this mention in the description:
>Discuss the future of Features as they hinder meta-programing.

I think it's quite the opposite. Features can enable very dynamic and reusable mixins, if we hit that sweet spot.

mlc-mlapis:
[**@IgorMinar**](https://github.com/IgorMinar) **@LayZeeDK** I always supposed that's `Component Features` is on the road to be public API when it'll be stabilized enough to avoid surprising breaking changes. Isn't the plan already?

LayZeeDK:
@mlc-mlapis they have been around for at least 2.5 years, but the primært focus for finishing the first stable release of Ivy was to maintain a proper amount of backwards compatibility. What's on the roadmap for Angular v11+, I think is still being discussed.

packages/core/src/render3/features/ng_onchanges_feature.ts
mhevery: I have added a comment with explanation. 

NOTE: This was existing code I only removed the un-needed `wrapOnChanges`

packages/core/src/render3/hooks.ts
```diff
@@ -73,27 +77,36 @@ export function registerPostOrderHooks(tView: TView, tNode: TNode): void {
   // hooks for projected components and directives must be called *before* their hosts.
   for (let i = tNode.directiveStart, end = tNode.directiveEnd; i < end; i++) {
     const directiveDef = tView.data[i] as DirectiveDef<any>;
-    if (directiveDef.afterContentInit) {
-      (tView.contentHooks || (tView.contentHooks = [])).push(-i, directiveDef.afterContentInit);
+    const lifecycleHooks: AfterContentInit&AfterContentChecked&AfterViewInit&AfterViewChecked&
```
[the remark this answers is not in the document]
mhevery: Yes, I agree with you that it is more readable, but `clang-format` removes them.

packages/core/src/render3/instructions/shared.ts
```diff
@@ -1183,17 +1184,19 @@ export function resolveDirectives(
         if (def.hostBindings !== null || def.hostAttrs !== null || def.hostVars !== 0)
           tNode.flags |= TNodeFlags.hasHostBindings;
 
+        const lifeCycleHooks: OnChanges&OnInit&DoCheck = def.type.prototype;
```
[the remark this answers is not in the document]
mhevery: That is how `clang-format` formats the code.

packages/core/test/acceptance/lifecycle_spec.ts
```diff
@@ -1124,6 +1125,33 @@ describe('onChanges', () => {
   });
 });
 
+describe('meta-programing', () => {
+  it('should allow adding lifecycle hook methods any time before bootstrap', () => {
```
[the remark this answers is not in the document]
mhevery: Good point, changed the text.

IgorMinar (changes requested):
@mhevery can you please make the CI pass. thanks

IgorMinar:
@LayZeeDK I hear you and agree about the potential and the sweet spot. But until View Engine is not fully deleted from Angular, I don't think that we have brain cycles to explore these ideas. 😞 🤯

pkozlowski-opensource (changes requested):
@mhevery I did the first pass on this PR and I agree with the general direction.

Having said this I'm marking this PR as "Request changes", as I believe that we need more tests (see in-line comments for specifics).

Also, the commit message only references #30497 but my understanding was that the main driver here was to fix #31495 (ivy + AoT). Could you please confirm that the both issues are being solved and amend the commit message accordingly?

pkozlowski-opensource (changes requested):
Reviewed-For: fw-core

packages/core/src/render3/features/ng_onchanges_feature.ts
```diff
@@ -55,28 +50,36 @@ export function ɵɵNgOnChangesFeature<T>(definition: DirectiveDef<T>): void {
...
+ * `ngOnChanges`.
+ *
+ * The hooke reads the `NgSimpleChangesStore` data from the component instance and if changes are
+ * found it invokes `ngOnChanges` on the component instance.
+ *
+ * @param this Component instance.
+ */
+function wrapOnChangesHook(this: OnChanges) {
+  const simpleChangesStore = getSimpleChangesStore(this);
+  const current = simpleChangesStore && simpleChangesStore.current;
```
jelbourn, line 64:
```suggestion
  const current = simpleChangesStore?.current;
```

packages/core/src/render3/features/ng_onchanges_feature.ts
```diff
@@ -55,28 +50,36 @@ export function ɵɵNgOnChangesFeature<T>(definition: DirectiveDef<T>): void {
...
+/**
+ * This is a synthetic lifecycle hook which gets inserted into `TView.preOrderHooks` to simulate
+ * `ngOnChanges`.
+ *
+ * The hooke reads the `NgSimpleChangesStore` data from the component instance and if changes are
+ * found it invokes `ngOnChanges` on the component instance.
+ *
+ * @param this Component instance.
+ */
+function wrapOnChangesHook(this: OnChanges) {
```
jelbourn, line 62: Does the name of this function still apply? The previous version did "wrap" the changes hook in another function, but now this appears to just be called directly. Would it be more descriptive to name it `rememberChangeHistoryAndInvokeOnChangesHook`?

packages/core/src/render3/features/ng_onchanges_feature.ts
```diff
@@ -41,12 +33,15 @@ type OnChangesExpando = OnChanges&{
...
+  return NgOnChangesFeatureImpl;
+}
 
-export function ɵɵNgOnChangesFeature<T>(definition: DirectiveDef<T>): void {
+export function NgOnChangesFeatureImpl<T>(definition: DirectiveDef<T>) {
   if (definition.type.prototype.ngOnChanges) {
     definition.setInput = ngOnChangesSetInput;
-    (definition as {onChanges: Function}).onChanges = wrapOnChanges();
   }
+  return wrapOnChangesHook;
```
jelbourn, line 44: Maybe add a comment here explaining that the `this` context for this function will be bound elsewhere (mentioning where that is).

packages/core/test/acceptance/lifecycle_spec.ts
```diff
@@ -1124,6 +1125,33 @@ describe('onChanges', () => {
...
+    (Child.prototype as any).ngOnInit = () => events.push('onInit');
+    (Child.prototype as any).ngOnChanges = (e: SimpleChanges) => events.push(e);
+
+    TestBed.configureTestingModule({
+      declarations: [App, Child],
+    });
+    const fixture = TestBed.createComponent(App);
+    fixture.detectChanges();
+    expect(events.length).toEqual(2);
+    expect(events[0].name.currentValue).toEqual('value');
```
jelbourn, line 1150: Would it be valuable to add a test that looks at the previous value as well?

packages/core/test/acceptance/lifecycle_spec.ts
```diff
@@ -1124,6 +1125,33 @@ describe('onChanges', () => {
...
+    class App {
+    }
+
+    @Component({selector: 'child', template: `empty`})
+    class Child {
+      @Input() name: string = '';
+    }
+
+    (Child.prototype as any).ngOnInit = () => events.push('onInit');
+    (Child.prototype as any).ngOnChanges = (e: SimpleChanges) => events.push(e);
```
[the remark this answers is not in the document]
mhevery: test extended

packages/core/test/acceptance/lifecycle_spec.ts
mhevery: added

packages/core/test/acceptance/lifecycle_spec.ts
```diff
@@ -1124,6 +1125,33 @@ describe('onChanges', () => {
...
+    TestBed.configureTestingModule({
+      declarations: [App, Child],
+    });
+    const fixture = TestBed.createComponent(App);
+    fixture.detectChanges();
+    expect(events.length).toEqual(2);
+    expect(events[0].name.currentValue).toEqual('value');
+    expect(events[1]).toEqual('onInit');
+  });
+});
```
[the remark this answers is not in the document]
mhevery: Happy to add a test, but it is not clear to me what I would be testing. The framework logic simply reads the prototype chain. All this refactoring is doing is delaying the reading from eager to lazy. So I don't know what such a test would assert other than that it works the same as normally declared code. Please advise.

goldens/size-tracking/aio-payloads.json
```diff
@@ -12,8 +12,13 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 2987,
+<<<<<<< HEAD
```
tonivj5, line 15: Merge markers here

goldens/size-tracking/integration-payloads.json
```diff
@@ -3,26 +3,41 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 1485,
+<<<<<<< HEAD
```
tonivj5, line 6: Here

goldens/size-tracking/integration-payloads.json
```diff
@@ -3,26 +3,41 @@
...
+        "polyfills-es2015": 36963
+>>>>>>> d863db05cc... fix(core): Allow modification of lifecycle hooks any time before bootstrap
       }
     }
   },
   "cli-hello-world-ivy-minimal": {
     "master": {
       "uncompressed": {
         "runtime-es2015": 1485,
+<<<<<<< HEAD
```
tonivj5, line 20: Here

goldens/size-tracking/integration-payloads.json
```diff
@@ -3,26 +3,41 @@
...
+        "polyfills-es2015": 36657
+>>>>>>> d863db05cc... fix(core): Allow modification of lifecycle hooks any time before bootstrap
       }
     }
   },
   "cli-hello-world-ivy-compat": {
     "master": {
       "uncompressed": {
         "runtime-es2015": 1485,
+<<<<<<< HEAD
```
tonivj5, line 34: And here :+1:

jelbourn (approved):
LGTM

Reviewed-for: public-api

Splaktar:
Re-ran CI and AIO local tests are now passing.

packages/core/test/acceptance/lifecycle_spec.ts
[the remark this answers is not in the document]
jelbourn: I think the rationale in adding tests for this would be that the tests don't know/care that the framework is just reading the prototype; that's just an implementation detail. The tests should verify that the expected behaviors work, even if that would seem "obvious" from the code.

packages/core/test/acceptance/lifecycle_spec.ts
```diff
@@ -1124,6 +1124,46 @@ describe('onChanges', () => {
...
+    @Component({template: `<child name="value"></child>`})
+    class App {
+    }
+
+    @Component({selector: 'child', template: `empty`})
+    class Child {
+      @Input() name: string = '';
+    }
+
+    (Child.prototype as any).ngOnInit = () => events.push('onInit');
```
jelbourn, line 1140: You could avoid some repetition here by doing something like
```ts
const componentPrototype = Child.prototype as any;
```

jelbourn (approved):
Still LGTM for public-api

Reviewed-for: public-api

mhevery:
[presubmit](http://test/OCL:318334378:BASE:318334429:1593116185622:cb382320)

packages/core/test/acceptance/lifecycle_spec.ts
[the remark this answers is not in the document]
pkozlowski-opensource: Yeh, my thinking is in-line with @jelbourn - things might be obvious today with the current implementation but not so obvious n months down the line with a modified implementation. Personally I always start to review PRs by reading tests to understand what is the use-case being solved and at this point I have no context / details on the underlying implementation details (nor should have I!).

So yeh, even if it sounds like a super-obvious from the code author point of view, I still think that it would be beneficial to have a test - even if it acts as documentation / non regression. I would even say that we should have test(s) that use a custom decorator and class inheritance to make sure that we really cover the use-case that was reported.

pkozlowski-opensource (approved):
The code change looks good to me. Still I would love to see few cleanups before this PR goes in.

* [ ] This bug fix is all about the AoT mode but we currently don't run tests in the AoT mode. The net effect here is that I can run the code with the new test and _no_ code change from this PR.... and the test will still pass... Can you think of having a test that fails without the code change?
* [ ] I would love to see more test(s) as [discussed](https://github.com/angular/angular/pull/35464#discussion_r441567036) 
* [ ] add reference to the issue #31495 (fixes #31495 or similar) since this is the actual issue being fixed here, no? (also mentioned this in [my previous review](https://github.com/angular/angular/pull/35464#pullrequestreview-432427887))

Reviewed-for: public-api
Reviewed-for: fw-core

pkozlowski-opensource (approved):
Reviewed-for: size-tracking

packages/core/src/render3/hooks.ts
```diff
@@ -31,20 +32,23 @@ import {getCheckNoChangesMode} from './state';
...
+    const wrappedOnChanges = NgOnChangesFeatureImpl(directiveDef);
+    (tView.preOrderHooks || (tView.preOrderHooks = [])).push(directiveIndex, wrappedOnChanges);
+    (tView.preOrderCheckHooks || (tView.preOrderCheckHooks = []))
+        .push(directiveIndex, wrappedOnChanges);
   }
 
-  if (onInit) {
-    (tView.preOrderHooks || (tView.preOrderHooks = [])).push(-directiveIndex, onInit);
+  if (ngOnInit) {
+    (tView.preOrderHooks || (tView.preOrderHooks = [])).push(0 - directiveIndex, ngOnInit);
```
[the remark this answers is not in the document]
mhevery: if `directiveIndex` is `0` than `-0` is a special kind of number which de-optimizes the VM. Where ase `0-0` is always `0`.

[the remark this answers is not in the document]
alfaproject: Out of curiosity, will Terser try to be smart here and change to `-0` anyway?

ramtob:
Hi guys, thanks for the good work.

I suppose, when this PR is finalized, it will find its way to the new version 10 of Angular as well?

IgorMinar (approved):
LGTM - the increase is minimal

Reviewed-for: size-tracking

jelbourn (approved):
LGTM

Reviewed-for: size-tracking

jelbourn (approved):
LGTM, using global approvers because the public-api change here is a false positive

mhevery:
[presubmit](http://test/OCL:318334378:BASE:321406361:1594838868604:64806800)

alxhub (approved):

packages/core/test/acceptance/lifecycle_spec.ts
mhevery: fixed

packages/core/test/acceptance/lifecycle_spec.ts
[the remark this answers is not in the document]
mhevery: tests added.

AndrewKushnir:
Hi @mhevery, this PR was merged to master only, since there was a conflict while merging into a patch branch. Could you please create a new PR that targets patch branch? Thank you.

PR #35466, opened 2020-02-15
fix(language-service): provide completions for the structural directive that only injects the 'ViewContainerRef'

ivanwonder:
The completions list don't include the 'ngTemplateOutlet'. The directive is a structural directive when it has injected the 'TemplateRef' or 'ViewContainerRef'.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/language-service/src/completions.ts
+++ b/packages/language-service/src/completions.ts
@@ -15,7 +15,7 @@ import {getExpressionCompletions} from './expressions';
 import {attributeNames, elementNames, eventNames, propertyNames} from './html_info';
 import {InlineTemplate} from './template';
 import * as ng from './types';
-import {diagnosticInfoFromTemplateInfo, findTemplateAstAt, getPathToNodeAtPosition, getSelectors, hasTemplateReference, inSpan, spanOf} from './utils';
+import {diagnosticInfoFromTemplateInfo, findTemplateAstAt, getPathToNodeAtPosition, getSelectors, inSpan, isStructuralDirective, spanOf} from './utils';
 
 const HIDDEN_HTML_ELEMENTS: ReadonlySet<string> =
     new Set(['html', 'script', 'noscript', 'base', 'body', 'title', 'head', 'link']);
@@ -629,11 +629,11 @@ function angularAttributes(info: AstResult, elementName: string): AngularAttribu
       continue;
     }
     const summary = selectorMap.get(selector) !;
-    const isTemplateRef = hasTemplateReference(summary.type);
+    const hasTemplateRef = isStructuralDirective(summary.type);
     // attributes are listed in (attribute, value) pairs
     for (let i = 0; i < selector.attrs.length; i += 2) {
       const attr = selector.attrs[i];
-      if (isTemplateRef) {
+      if (hasTemplateRef) {
         templateRefs.add(attr);
       } else {
         others.add(attr);
--- a/packages/language-service/src/utils.ts
+++ b/packages/language-service/src/utils.ts
@@ -56,9 +56,11 @@ export function isNarrower(spanA: Span, spanB: Span): boolean {
   return spanA.start >= spanB.start && spanA.end <= spanB.end;
 }
 
-export function hasTemplateReference(type: CompileTypeMetadata): boolean {
+export function isStructuralDirective(type: CompileTypeMetadata): boolean {
   for (const diDep of type.diDeps) {
-    if (diDep.token && identifierName(diDep.token.identifier) === Identifiers.TemplateRef.name) {
+    const diDepName = identifierName(diDep.token?.identifier);
+    if (diDepName === Identifiers.TemplateRef.name ||
+        diDepName === Identifiers.ViewContainerRef.name) {
       return true;
     }
   }
--- a/packages/language-service/test/completions_spec.ts
+++ b/packages/language-service/test/completions_spec.ts
@@ -129,6 +129,7 @@ describe('completions', () => {
       'ngSwitchCase',
       'ngSwitchDefault',
       'ngPluralCase',
+      'ngTemplateOutlet',
     ]);
   });
 


packages/language-service/test/completions_spec.ts
```diff
@@ -123,12 +123,8 @@ describe('completions', () => {
...
     const completions = ngLS.getCompletionsAtPosition(TEST_TEMPLATE, marker.start);
     expectContain(completions, CompletionKind.ATTRIBUTE, [
-      'ngFor',
-      'ngForOf',
-      'ngIf',
-      'ngSwitchCase',
-      'ngSwitchDefault',
-      'ngPluralCase',
+      'ngFor', 'ngForOf', 'ngIf', 'ngSwitchCase', 'ngSwitchDefault', 'ngPluralCase',
+      'ngTemplateOutlet'
```
ayazhafiz: Can you add a comma at the end here to keep it in a vertical list

packages/language-service/src/utils.ts
```diff
@@ -58,7 +58,7 @@ export function isNarrower(spanA: Span, spanB: Span): boolean {
 
 export function hasTemplateReference(type: CompileTypeMetadata): boolean {
   for (const diDep of type.diDeps) {
-    if (diDep.token && identifierName(diDep.token.identifier) === Identifiers.TemplateRef.name) {
+    if (identifierName(diDep.token?.identifier) === Identifiers.TemplateRef.name || identifierName(diDep.token?.identifier) === Identifiers.ViewContainerRef.name) {
```
ayazhafiz: Please make (identifierName(diDep.token?.identifier) a local variable

ayazhafiz (commented):
Well structural directives still always have view containers and template refs... `NgTemplateOutput` just uses an input:

https://github.com/angular/angular/blob/755d2d572f00efa4b7362c71d0299bef1655788f/packages/common/src/directives/ng_template_outlet.ts#L50

This LGTM, but let's rename the function to something like `isStructuralDirective` since that's what we're trying to deduce now.

ivanwonder:
There is another problem about the structural directive. We should only provide the structural directive for the `ng-template` element. For example the `ngFor` use case below.
```
<ng-template ngFor let-item [ngForOf]="items" let-i="index" [ngForTrackBy]="trackByFn">
  <li>...</li>
</ng-template>
```
Now it only provide the attribute directives, I think it's meaningless for the `ng-template`.

packages/language-service/src/completions.ts
```diff
@@ -629,7 +629,7 @@ function angularAttributes(info: AstResult, elementName: string): AngularAttribu
       continue;
     }
     const summary = selectorMap.get(selector) !;
-    const isTemplateRef = hasTemplateReference(summary.type);
+    const isTemplateRef = isStructuralDirective(summary.type);
```
ayazhafiz: nit, for semantics (it [probably] has a template ref, but the directive itself is not one) 

```suggestion
    const hasTemplateRef = isStructuralDirective(summary.type);
```

ayazhafiz (approved):

ayazhafiz:
> There is another problem about the structural directive. We should only provide the structural directive for the `ng-template` element.

Well, structural directives are basically a syntactic wrapper for `ng-template`. So the microsyntax attributes like `*ngFor` should be provided for elements, as long as `NgForOf` is in the application scope.

> ```html
> <ng-template ngFor let-item [ngForOf]="items" let-i="index" [ngForTrackBy]="trackByFn">
>   <li>...</li>
> </ng-template>
> ```

This is legal. But maybe it doesn't make sense to provide microsyntax completions on template elements themselves. For example I'm not sure this works:

```html
<ng-template *ngIf="true">hi</ng-template>
```

It would get lowered to

```html
<ng-template [ngIf]="true">
  <ng-template>hi</ng-template>
</ng-template>
```

I don't know if the compiler accepts this.

packages/language-service/src/completions.ts
ivanwonder: Maybe I can remove the local variable because it's only used in the 'if' condition

ayazhafiz: It runs through a loop though, right?

ivanwonder: Yes, I will apply your suggestion.

ivanwonder:
> ~~Now it only provide the attribute directives,~~ I think it's meaningless for the `ng-template`.

```
<ng-template ngFor~{cursor}></ng-template>
```
My mistake, there is just no `ngFor` provided here.

I think the `attribute directives` is meaningless for `ng-template`. Because we only provide `structural directives` after `*`.
```
<div *ngClass="'className'"></div>
```
It will compile to code below.
```
<ng-template [ngClass]="'className'">
  <div></div>
</ng-template>
```
So do we need provide `ngClass` for `ng-template`? and  the attribute directive changes the appearance or behavior of a DOM element.

ivanwonder:
Test on stackblit, angular accept all the code.😂

ayazhafiz:
Well, `ngClass` is an input on `ng-template`, and in fact it is a directive that can be matched on an element or template.

I think what is more related to your first comment is the scope provided by a directive applied on elements and templates. In the case of an element, a structural directive context is applied via a microsyntax attribute. In the case of a template, it's applied in the same way a normal attribute is declared. In both cases, it's the same directive being applied, and the same inputs being brought into scope -- just different syntaxes for doing so. And in the case of the template, bringing a directive into scope also scopes all of its inputs, like `ngTrackBy` in `<div ngFor [ngTrackBy]="...">`. Note that `ngTrackBy` is an input to `NgFor`, not an attribute on the element.

ivanwonder:
This discussion is not related to the PR, I will think about it again.

ayazhafiz:
Sounds good. Feel free to open an issue/PR if needed.

________________________________
От: ivan <notifications@github.com>
Отправлено: Sunday, February 16, 2020 2:07:27 AM
Кому: angular/angular <angular@noreply.github.com>
Копия: hafiz <ayaz.hafiz.1@gmail.com>; Review requested <review_requested@noreply.github.com>
Тема: Re: [angular/angular] fix(language-service): provide completions for the structural directive that only injects the 'ViewContainerRef' (#35466)


This discussion is not related to the PR, I will think about it again.

kyliau (approved):

PR #35474, opened 2020-02-15
build(core/labs):  Set up `@angular/core/labs` package for experimentation

mhevery:
This change creates a new sub-package `@angular/core/labs` the purpose of which is to allow us to have experimental API in a way where the community can try them out without us being committed to supporting them long term.

Before the developer can use any of the APIs in the `@angular/core/labs` package they have to agree to the conditions which are clearly spelled out in both the way of agreeing as well as a warning in the console.

```
import {iWantToUseExperimentalAPIs} from `@angular/core/labs`

iWantToUseExperimentalAPIs({
  iUnderstand: ['not ready for production', 'unstable API']
})
```

The purpose of the above code is to make it very clear that the developer is about to use experimental APIs which should not be used in production.

Additionally this change sets up testing harness for `@angular/core/labs` where the specs are next to the actual code to simplify the directory structure.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/compiler/test/aot/test_util.ts
+++ b/packages/compiler/test/aot/test_util.ts
@@ -622,8 +622,17 @@ export function setup(options: {
       // If running under bazel then we get the compiled version of the files from the bazel package
       // output.
       const bundles = new Set([
-        'bundles', 'esm2015', 'esm5', 'testing', 'testing.d.ts', 'testing.metadata.json', 'browser',
-        'browser.d.ts'
+        'bundles',
+        'esm2015',
+        'esm5',
+        'testing',
+        'testing.d.ts',
+        'testing.metadata.json',
+        'browser',
+        'browser.d.ts',
+        'labs',
+        'labs.d.ts',
+        'labs.metadata.json',
       ]);
       const skipDirs = (name: string) => bundles.has(name);
       if (options.compileAngular) {
--- a/packages/core/BUILD.bazel
+++ b/packages/core/BUILD.bazel
@@ -27,6 +27,8 @@ ng_package(
     name = "npm_package",
     srcs = glob(["**/*.externs.js"]) + [
         "package.json",
+        "//packages/core/labs:package.json",
+        "//packages/core/labs:README.md",
         "//packages/core/testing:package.json",
     ],
     entry_point = ":index.ts",
@@ -51,6 +53,7 @@ ng_package(
     ],
     deps = [
         ":core",
+        "//packages/core/labs",
         "//packages/core/testing",
     ],
 )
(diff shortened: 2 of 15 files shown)


IgorMinar (changes requested):
We should merge this with https://github.com/angular/angular/pull/26003

The two PRs use different designs to create the labs. The original PR follows a design that ahs gone through several iterations so it would be great if you could capture why that design is inferior to the one used in this PR.

IgorMinar:
Also, is it intentional that this PR is in a draft state but has reviewers assigned? What kind of feedback are you looking for from the reviewers on a PR that is not ready yet? Reviewing PRs take a lot of time so it would be great if you set the expectations upfront so that the reviewers know how deep they should go.

Thanks for the PR description that one is very helpful, it just doesn't answer what state this PR is in.

mhevery:
> Also, is it intentional that this PR is in a draft state but has reviewers assigned? What kind of feedback are you looking for from the reviewers on a PR that is not ready yet? Reviewing PRs take a lot of time so it would be great if you set the expectations upfront so that the reviewers know how deep they should go.

I think the short answer is that I don't know how to use Draft button. My thinking was that this is more of a discussion rather than ready to be merged hence I wanted to  signal with the Draft.

mhevery:
> The two PRs use different designs to create the labs. The original PR follows a design that ahs gone through several iterations so it would be great if you could capture why that design is inferior to the one used in this PR.

The biggest difference I see is:
1. is it `@angular/core/labs` or `@angular/labs/core`. I know the design doc says `@angular/labs/core`, but I think `@angular/core/core` makes more sense from the build file organization.
2. I also suggest rules for test files unification in same directory
3. propose a way to agree to a disclaimer.

robwormald:
my PR used `@angular/labs/package-name` for a few reasons:
- it required an explicit opt-in package install of `@angular/labs` vs experimental APIs being added to existing packages.
- `@angular/labs` package easier to exclude from builds / CI
- easy promotion / merging to top-level packages, eg `@angular/labs/http-client` => `@angular/http-client`
- potential for automated releasing outside of normal release cycle

This was mostly set up for feature package development, like a unified `@angular/forms` rewrite or a standalone animations engine. For experimental core / Ivy APIs, IMHO we should decide where we want to be in a couple of years in terms of package structure, especially the dependencies between the `@angular/core`, `@angular/platform-*`, and `@angular/common` packages and stuff like `BrowserModule`

PR #35481, opened 2020-02-16
fix(ivy): incorrectly generating shared pure function between null and object literal

### LINKED ISSUE #35298: Angular 9 confuses literal null with undefined when wrapped in object in input
# 🐞 bug report

### Affected Package

The issue is caused by package @angular/.... I guess compiler

### Is this a regression?

### Description

Angular 9 confuses literal `null` with `undefined` in inputs when the literal is wrapped in object and there are at least 2 instances of given component. In my project this breaks routing because there is a difference between

```html
<a [routerLink]="['', { outlets: { aux: null } }]"></a>
<a [routerLink]="['', { outlets: { aux: undefined } }]"></a>
```

## 🔬 Minimal Reproduction

1. execute following:

   ```sh
   git clone https://github.com/bgoscinski/angular-bug-repro1.git
   cd angular-bug-repro1
   npm install
   ng serve
   ```

1. observe the difference: 
image

1. see https://github.com/bgoscinski/angular-bug-repro1/blob/master/src/app/app.component.html#L17-L22

## 🌍 Your Environment

**Angular Version:**

<pre><code>
                |___/
    

Angular CLI: 9.0.1
Node: 12.7.0
OS: win32 x64

Angular: 9.0.0
... router
Ivy Workspace: Yes

@angular-devkit/architect         0.900.1
@angular-devkit/build-angular     0.900.1
@angular-devkit/build-optimizer   0.900.1
@angular-devkit/build-webpack     0.900.1
@angular-devkit/core              9.0.1
@angular-devkit/schematics        9.0.1
@angular/cli                      9.0.1
@ngtools/webpack                  9.0.1
@schematics/angular               9.0.1
@schematics/update                0.900.1
</code></pre>

**Anything else relevant?**

I guess the easiest workaround is to declare property on component instance and set its value to `null` and use it instead of literal `null` in templates

crisbeto:
In #33705 we made it so that we generate pure functions for object/array literals in order to avoid having them be shared across elements/views. The problem this introduced is that further down the line the `ContantPool` uses the generated literal in order to figure out whether to share an existing factory or to create a new one. `ConstantPool` determines whether to share a factory by creating a key from the AST node and using it to look it up in the factory cache, however the key generation function didn't handle function invocations and replaced them with `null`. This means that the key for `{foo: pureFunction0(...)}` and `{foo: null}` are the same.

These changes rework the logic so that instead of generating a `null` key for function invocations, we generate a variable called `<unknown>` which shouldn't be able to collide with anything.

Fixes #35298.

### DIFF
--- a/packages/compiler/src/constant_pool.ts
+++ b/packages/compiler/src/constant_pool.ts
@@ -127,16 +127,17 @@ export class ConstantPool {
 
   getLiteralFactory(literal: o.LiteralArrayExpr|o.LiteralMapExpr):
       {literalFactory: o.Expression, literalFactoryArguments: o.Expression[]} {
-    // Create a pure function that builds an array of a mix of constant  and variable expressions
+    // Create a pure function that builds an array of a mix of constant and variable expressions
     if (literal instanceof o.LiteralArrayExpr) {
-      const argumentsForKey = literal.entries.map(e => e.isConstant() ? e : o.literal(null));
+      const argumentsForKey =
+          literal.entries.map(e => e.isConstant() ? e : o.variable('<unknown>'));
       const key = this.keyOf(o.literalArr(argumentsForKey));
       return this._getLiteralFactory(key, literal.entries, entries => o.literalArr(entries));
     } else {
       const expressionForKey = o.literalMap(
           literal.entries.map(e => ({
                                 key: e.key,
-                                value: e.value.isConstant() ? e.value : o.literal(null),
+                                value: e.value.isConstant() ? e.value : o.variable('<unknown>'),
                                 quoted: e.quoted
                               })));
       const key = this.keyOf(expressionForKey);
--- a/packages/compiler/src/constant_pool.ts
+++ b/packages/compiler/src/constant_pool.ts
@@ -11,6 +11,16 @@ import {OutputContext, error} from './util';
 
 const CONSTANT_PREFIX = '_c';
 
+/**
+ * `ConstantPool` tries to reuse literal factories when two or more literals are identical.
+ * We determine whether literals are identical by creating a key out of their AST using the
+ * `KeyVisitor`. This constant is used to replace dynamic expressions which can't be safely
+ * converted into a key. E.g. given an expression `{foo: bar()}`, since we don't know what
+ * the result of `bar` will be, we create a key that looks like `{foo: <unknown>}`. Note
+ * that we use a variable, rather than something like `null` in order to avoid collisions.
+ */
+const UNKNOWN_VALUE_KEY = o.variable('<unknown>');
+
 export const enum DefinitionKind {Injector, Directive, Component, Pipe}
 
 /**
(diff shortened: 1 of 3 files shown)


packages/core/test/acceptance/pure_function_spec.ts
```diff
@@ -559,5 +559,65 @@ describe('components using pure function instructions internally', () => {
...
+      const values = fixture.componentInstance.directives.map(directive => directive.value);
+
+      expect(values).toEqual([{foo: null}, {foo: []}]);
+    });
+
+    it('should not confuse function calls and null inside of a literal', () => {
+      @Component({
+        template: `
+          <div [dir]="{foo: null}"></div>
+          <div [dir]="{foo: getFoo()}"></div>
```
crisbeto: I think this one has been an issue even before #33705, but we just never hit it.

alxhub (approved):

packages/compiler/src/constant_pool.ts
```diff
@@ -255,12 +256,25 @@ class KeyVisitor implements o.ExpressionVisitor {
     return `TYPEOF:${node.expr.visitExpression(this, context)}`;
   }
 
+  visitInvokeFunctionExpr(node: o.InvokeFunctionExpr, context: any): string {
+    return `FN_CALL:${node.fn.visitExpression(this, context)}:ARGS:${node.args.length}`;
+  }
+
+  visitInvokeMethodExpr(node: o.InvokeMethodExpr, context: any): string {
+    const receiver = node.receiver.visitExpression(this, context);
+    return `METHOD_CALL:${receiver}${node.name}:ARGS:${node.args.length}`;
```
alxhub: Should this not be
```suggestion
    return `METHOD_CALL:${receiver}:${node.name}:ARGS:${node.args.length}`;
```

alxhub (changes requested):

packages/compiler/src/constant_pool.ts
crisbeto: I think I meant that one to be `${receiver}.${node.name}`.

packages/compiler/src/constant_pool.ts
```diff
@@ -127,16 +127,17 @@ export class ConstantPool {
...
+      const argumentsForKey =
+          literal.entries.map(e => e.isConstant() ? e : o.variable('<unknown>'));
       const key = this.keyOf(o.literalArr(argumentsForKey));
       return this._getLiteralFactory(key, literal.entries, entries => o.literalArr(entries));
     } else {
       const expressionForKey = o.literalMap(
           literal.entries.map(e => ({
                                 key: e.key,
-                                value: e.value.isConstant() ? e.value : o.literal(null),
+                                value: e.value.isConstant() ? e.value : o.variable('<unknown>'),
```
alxhub: Maybe extract the `o.variable(...)` to a constant, and add a comment about what's going on there.

alxhub (approved):
Please update the commit message as well - it's now out of date since the changes we made yesterday :-)

mhevery:
[presubmit](http://test/OCL:296245846:BASE:296245880:1582224164789:469b435e)

kara (approved):
LGTM. Thanks for the detailed commit message! :)

PR #35307, opened 2020-02-10
fix(ivy): support dynamic query tokens in AOT mode

JoostK:
For view and content queries, the Ivy compiler attempts to statically
evaluate the predicate token so that string predicates containing
comma-separated reference names can be split into an array of strings
during compilation. When the predicate is a dynamic value that cannot be
statically interpreted at compile time, the compiler would previously
produce an error. This behavior breaks a use-case where an `InjectionToken`
is being used as query predicate, as the usage of the `new` keyword
prevents such predicates from being statically evaluated.

This commit changes the behavior to no longer produce an error for
dynamic values. Instead, the expression is emitted as is into the
generated code, postponing the evaluation to happen at runtime.

Fixes #34267
Resolves FW-1828

PR #35529, opened 2020-02-18
fix(ngcc): Corrected error message with correct variable.

### LINKED ISSUE #35504: ERROR in The given rootPath undefined is not a file of the program.
When upgrading to Angular 9:

Screenshot 2020-02-17 at 16 52 39

sacgrover:
Fixes #35504

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #35504

## What is the new behavior?
Error message updated with correct variable.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/compiler-cli/ngcc/src/packages/entry_point_bundle.ts
+++ b/packages/compiler-cli/ngcc/src/packages/entry_point_bundle.ts
@@ -7,7 +7,7 @@
  */
 import * as ts from 'typescript';
 import {AbsoluteFsPath, FileSystem, NgtscCompilerHost} from '../../../src/ngtsc/file_system';
-import {PathMappings} from '../utils';
+import {PathMappings, resolveFileWithPostfixes} from '../utils';
 import {BundleProgram, makeBundleProgram} from './bundle_program';
 import {EntryPoint, EntryPointFormat} from './entry_point';
 import {NgccSourcesCompilerHost} from './ngcc_compiler_host';
@@ -56,16 +56,20 @@ export function makeEntryPointBundle(
   const dtsHost = new NgtscCompilerHost(fs, options);
 
   // Create the bundle programs, as necessary.
-  const absFormatPath = fs.resolve(entryPoint.path, formatPath);
-  const typingsPath = fs.resolve(entryPoint.path, entryPoint.typings);
+  const absFormatPath = resolveFileWithPostfixes(
+      fs, fs.resolve(entryPoint.path, formatPath), ['', '.js', '/index.js']);
+  if (absFormatPath === null) {
+    throw new Error(
+        `The source file ${formatPath} could not be resolved within entry-point ${entryPoint.path}`);
+  }
   const src = makeBundleProgram(
       fs, isCore, entryPoint.package, absFormatPath, 'r3_symbols.js', options, srcHost);
   const additionalDtsFiles = transformDts && mirrorDtsFromSrc ?
-      computePotentialDtsFilesFromJsFiles(fs, src.program, absFormatPath, typingsPath) :
+      computePotentialDtsFilesFromJsFiles(fs, src.program, absFormatPath, entryPoint.typings) :
       [];
   const dts = transformDts ? makeBundleProgram(
-                                 fs, isCore, entryPoint.package, typingsPath, 'r3_symbols.d.ts',
-                                 options, dtsHost, additionalDtsFiles) :
+                                 fs, isCore, entryPoint.package, entryPoint.typings,
+                                 'r3_symbols.d.ts', options, dtsHost, additionalDtsFiles) :
                              null;
   const isFlatCore = isCore && src.r3SymbolsFile === null;
 
--- a/packages/compiler-cli/ngcc/test/packages/entry_point_bundle_spec.ts
+++ b/packages/compiler-cli/ngcc/test/packages/entry_point_bundle_spec.ts
@@ -287,7 +287,7 @@ runInEachFileSystem(() => {
         name: 'secondary',
         packageJson: {name: 'secondary'},
         package: absoluteFrom('/node_modules/primary'),
-        path: absoluteFrom('/node_modules/primary/secondary'),
+        path: absoluteFrom('/node_modules/primary/esm2015/secondary'),
         typings: absoluteFrom('/node_modules/primary/secondary/index.d.ts'),
         compiledByAngular: true,
         ignoreMissingDependencies: false,
@@ -297,6 +297,30 @@ runInEachFileSystem(() => {
           fs, entryPoint, './index.js', false, 'esm2015', /* transformDts */ true,
           /* pathMappings */ undefined, /* mirrorDtsFromSrc */ true);
       expect(bundle.rootDirs).toEqual([absoluteFrom('/node_modules/primary')]);
+    });
+
+    it('resolve source file path without .js extension in formatPath', () => {
+      setupMockFileSystem();
+      const fs = getFileSystem();
+      const entryPoint: EntryPoint = {
+        name: 'test',
+        packageJson: {name: 'test'},
+        package: absoluteFrom('/node_modules/test'),
+        path: absoluteFrom('/node_modules/test'),
+        typings: absoluteFrom('/node_modules/test/index.d.ts'),
+        compiledByAngular: true,
+        ignoreMissingDependencies: false,
+        generateDeepReexports: false,
+      };
+      // source file path without .js extension
+      const formatPath = './index';
+      const esm5bundle = makeEntryPointBundle(
+          fs, entryPoint, formatPath, false, 'esm2015', /* transformDts */ true,
+          /* pathMappings */ undefined, /* mirrorDtsFromSrc */ false);
+      expect(esm5bundle.src.program.getSourceFiles().map(sf => sf.fileName))
+          .toContain(absoluteFrom('/node_modules/test/index.js'));
+      expect(esm5bundle.dts !.program.getSourceFiles().map(sf => sf.fileName))
+          .toContain(absoluteFrom('/node_modules/test/index.d.ts'));
     });
   });
 });
--- a/packages/compiler-cli/ngcc/src/host/esm2015_host.ts
+++ b/packages/compiler-cli/ngcc/src/host/esm2015_host.ts
@@ -1549,10 +1549,8 @@ export class Esm2015ReflectionHost extends TypeScriptReflectionHost implements N
       Map<ts.Declaration, ts.Declaration> {
     const declarationMap = new Map<ts.Declaration, ts.Declaration>();
     const dtsDeclarationMap = new Map<string, ts.Declaration>();
-    const rootDts = getRootFileOrFail(dts);
-    this.collectDtsExportedDeclarations(dtsDeclarationMap, rootDts, dts.program.getTypeChecker());
-    const rootSrc = getRootFileOrFail(src);
-    this.collectSrcExportedDeclarations(declarationMap, dtsDeclarationMap, rootSrc);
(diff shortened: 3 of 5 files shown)


JoostK (changes requested):
Although this does address the incorrect error message (thanks for opening a PR!), the underlying issue would still occur. I had a brief look at the cause of the bug and we can actually fix ngcc so that it will no longer error out.

Are you interested in contributing more to this PR? I can definitely help to provide pointers. Alternatively, I could push some commits to this PR to address the problem.

sacgrover:
@JoostK Yes, I am interested in it. Please provide me the pointers so that I can dig more into it.

JoostK:
Cool, thanks!

So the issue here is that the `typings` field in the package.json of `angular-sortablejs` points to `./dist/index`, which is resolved by the TypeScript compiler to the `./dist/index.d.ts` file. Unfortunately, ngcc fails to perform this resolution and then searches for `./dist/index` which doesn't exist, so that it errors out.

The absolute paths are constructed here:

https://github.com/angular/angular/blob/6b468f9b2e4b99cd9e784c2515484bce47147476/packages/compiler-cli/ngcc/src/packages/entry_point_bundle.ts#L59-L60

We can adjust this code to perform the resolution:

```ts
  const absFormatPath = resolveFileWithPostfixes(fs, fs.resolve(entryPoint.path, formatPath), ['', '.js', '/index.js']);
  if (absFormatPath === null) {
    throw new Error(`The source file ${formatPath} could not be resolved within entry-point ${entryPoint.path}`);
  }

  const typingsPath = resolveFileWithPostfixes(fs, fs.resolve(entryPoint.path, entryPoint.typings), ['', '.d.ts', '/index.d.ts']);
  if (typingsPath === null) {
    throw new Error(`The typings file ${entryPoint.typings} could not be resolved within entry-point ${entryPoint.path}`);
  }
```

Then, we should be able to update this:

https://github.com/angular/angular/blob/5b42084912f6bc8bdc419b0893fe3e25351719dc/packages/compiler-cli/ngcc/src/host/esm2015_host.ts#L1552-L1555

to the following:

```ts
    this.collectDtsExportedDeclarations(dtsDeclarationMap, dts.file, dts.program.getTypeChecker());
    this.collectSrcExportedDeclarations(declarationMap, dtsDeclarationMap, src.file);
```

and voila, we can actually remove `getRootFileOrFail` altogether :)

For tests, we can probably add a test to `packages/compiler-cli/ngcc/test/packages/entry_point_bundle_spec.ts` to verify that the paths get resolved as expected. If you need hints there, feel free to ping me here.

sacgrover:
Thanks, I will make that changes with test.

sacgrover:
@JoostK I made the changes. CI/CD wasn't triggered after new ommit.

packages/compiler-cli/ngcc/test/packages/entry_point_bundle_spec.ts
```diff
@@ -298,5 +302,28 @@ runInEachFileSystem(() => {
           /* pathMappings */ undefined, /* mirrorDtsFromSrc */ true);
       expect(bundle.rootDirs).toEqual([absoluteFrom('/node_modules/primary')]);
     });
+
+    // https://github.com/angular/angular/issues/35504
+    it('should work when typings has no mentioned .d.ts in the path ', () => {
```
JoostK:
```suggestion
    it('resolve a typings path without .d.ts extension', () => {
```

packages/compiler-cli/ngcc/test/packages/entry_point_bundle_spec.ts
```diff
@@ -298,5 +302,28 @@ runInEachFileSystem(() => {
...
+        packageJson: {name: 'test'},
+        package: absoluteFrom('/node_modules/test'),
+        path: absoluteFrom('/node_modules/test'),
+        typings: absoluteFrom('/node_modules/test/index'),
+        compiledByAngular: true,
+        ignoreMissingDependencies: false,
+        generateDeepReexports: false,
+      };
+      const esm5bundle = makeEntryPointBundle(
+          fs, entryPoint, './index.js', false, 'esm2015', /* transformDts */ true,
```
JoostK: Could you add another test where the `./index.js` that is passed here doesn't have an extension?

JoostK (commented):
This is looking really good, thank you so much for working on this PR! 💯

I have two additional requests:
- Could you add an additional test as mentioned in a review comment?
- A good practice for any commit in the Angular repo is that the commit message provides some context into what the problem is, and what change is made to fix the bug. An example can be found here: #35307. Again, if you need help just let me know.

Other than that, it looks like you may need to fetch the latest `master` commit and rebase this PR on top of it, as the size failure on CI is definitely not caused by this PR.

Again, thank you so much for taking the time to contribute, it is really appreciated! 🎉

packages/compiler-cli/ngcc/test/packages/entry_point_bundle_spec.ts
sacgrover: Thanks, changing as per suggestion.

packages/compiler-cli/ngcc/test/packages/entry_point_bundle_spec.ts
sacgrover: Adding this test case.

sacgrover:
@JoostK Thanks for review and guidance.

packages/compiler-cli/ngcc/test/packages/entry_point_bundle_spec.ts
```diff
@@ -154,6 +154,10 @@ runInEachFileSystem(() => {
           name: _('/node_modules/primary/secondary/index.d.ts'),
           contents: 'export declare function secondary();'
         },
+        {
```
JoostK: Is this change still needed? I don't see why it has been added.

packages/compiler-cli/ngcc/test/packages/entry_point_bundle_spec.ts
```diff
@@ -298,5 +302,50 @@ runInEachFileSystem(() => {
...
+        packageJson: {name: 'test'},
+        package: absoluteFrom('/node_modules/test'),
+        path: absoluteFrom('/node_modules/test'),
+        typings: absoluteFrom('/node_modules/test/index.d.ts'),
+        compiledByAngular: true,
+        ignoreMissingDependencies: false,
+        generateDeepReexports: false,
+      };
+      const esm5bundle = makeEntryPointBundle(
+          fs, entryPoint, './index', false, 'esm2015', /* transformDts */ true,
```
JoostK: For clarity, consider introducing a variable for the `'./index'` string with a comment, so that the scenario that is tested here becomes a lot easier to follow.

packages/compiler-cli/ngcc/test/packages/entry_point_bundle_spec.ts
```diff
@@ -298,5 +302,50 @@ runInEachFileSystem(() => {
...
+          /* pathMappings */ undefined, /* mirrorDtsFromSrc */ false);
+      expect(esm5bundle.src.program.getSourceFiles().map(sf => sf.fileName))
```
JoostK: Could you add an expect for the `esm5bundle.file` property? It should be the entry-point file:

```ts
expect(esm5bundle.file.fileName).toEqual(absoluteFrom('/node_modules/test/index.js'));
```

JoostK (commented):
Thanks @sacgrover. I realized there's something incorrect going on with the typings path, please see the individual review comments.

Here's a suggestion to extend the commit message with a bit more context, and some typo's fixed:

> A package.json file of a library may refer to source and typings files without extension, in which case the TypeScript compiler is able to resolve the files according the a set of common file extensions. ngcc however would fail to perform this resolution, which could cause an error because the files could not be found.
> 
> This commit changes the behavior to allow file paths without an extension to be resolved in a similar way, preventing the crash.

packages/compiler-cli/ngcc/test/packages/entry_point_bundle_spec.ts
sacgrover: ok

packages/compiler-cli/ngcc/test/packages/entry_point_bundle_spec.ts
sacgrover: ok

packages/compiler-cli/ngcc/test/packages/entry_point_bundle_spec.ts
sacgrover: @JoostK  Yes, this is needed because after changes in getting absolute path in following line 
https://github.com/sacgrover/angular/blob/rootpath/packages/compiler-cli/ngcc/src/packages/entry_point_bundle.ts#L59
It will throw an error because absolute path will be null as there is no file. so already written test case will show that error. however we are not expecting it in this test.
https://github.com/sacgrover/angular/blob/rootpath/packages/compiler-cli/ngcc/test/packages/entry_point_bundle_spec.ts#L294

JoostK: Ah, I see. In that case, I think the testcase should be updated instead so that `'./index.js'` is changed into `'../esm2015/secondary/index.js'`, so that it actually points to an existing file?

JoostK: But it reminds me, we should have a test for the scenario where the file does indeed not exist, in which case we expect the exception to be thrown.

sacgrover: yes.. we can do that..

packages/compiler-cli/ngcc/src/packages/entry_point_bundle.ts
```diff
@@ -56,8 +56,18 @@ export function makeEntryPointBundle(
...
-  const absFormatPath = fs.resolve(entryPoint.path, formatPath);
-  const typingsPath = fs.resolve(entryPoint.path, entryPoint.typings);
+  const absFormatPath = resolveFileWithPostfixes(
+      fs, fs.resolve(entryPoint.path, formatPath), ['', '.js', '/index.js']);
+  if (absFormatPath === null) {
+    throw new Error(
+        `The source file ${formatPath} could not be resolved within entry-point ${entryPoint.path}`);
+  }
+  const typingsPath = resolveFileWithPostfixes(
+      fs, fs.resolve(entryPoint.path, entryPoint.typings), ['', '.d.ts', '/index.d.ts']);
```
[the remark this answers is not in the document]
sacgrover: ok

packages/compiler-cli/ngcc/test/packages/entry_point_bundle_spec.ts
sacgrover: esm5bundle.file this doesn't exist.

sacgrover:
@JoostK  It seems, we can't resolve the typing path in entry_point.ts. A lot  of test cases are failing.

jjamaral:
Hi @sacgrover!

I'm using latest version of Angular (9.1.1).

I'm getting this error running local server when importing for example _**MatTableModule**_:
`ERROR in The given rootPath undefined is not a file of the program.`

If I don't import this module it works perfectly. I also test with another module from @angular/components (MatInputModule) and it's working. This error happens when I import MatTableModule (and paginator, sort, etc).

Do you think this PR will fix this also?

Thank you in advance!

sacgrover:
Hi @jjamaral,

This PR doesn't have the right code yet. Maybe someone has already worked on it in another PR.

JoostK:
Hi @sacgrover, sorry for never getting back to this PR. Since it is now quite outdated and the issue doesn't really surface anymore, I'll close this PR. Thank you for your contributions!

PR #35568, opened 2020-02-20
fix(core): handle synthetic props in Directive host bindings correctly

### LINKED ISSUE #35501: Cannot attach animation trigger using @HostBinding - renderer is undefined
# 🐞 bug report

### Affected Package

angular/core ?
Exactly similar to this issue here - https://github.com/angular/angular/issues/35342, however i am able to reproduce it with Angular - 9.0.1

### Is this a regression?

Yes, works as expected in Angular 8.x

### Description

Using the `@HostBinding` decorator to attach animation triggers to the host element does not work.

## 🔬 Minimal Reproduction

1) Clone this repo https://github.com/greyskulll/angular-animation-directive-hostbinding-issue
2) `ng serve` and check the browser console

In short, using `@HostBinding('@<animation trigger name>')` fails

## 🔥 Exception or Error
In Chrome
<pre><code>
core.js:5845 ERROR TypeError: Cannot read property 'listen' of undefined
    at isProceduralRenderer (core.js:4338)
    at elementPropertyInternal (core.js:12342)
    at Module.ɵɵupdateSyntheticHostBinding (core.js:25196)
    at SmoothHeightChangeDirective_HostBindings (uik-smooth-height-change.directive.ts:6)
    at setHostBindingsByExecutingExpandoInstructions (core.js:11385)
    at refreshView (core.js:11752)
    at refreshComponent (core.js:13145)
    at refreshChildComponents (core.js:11436)
    at refreshView (core.js:11757)
    at renderComponentOrTemplate (core.js:11831)
</code></pre>

In Firefox
<pre><code>
ERROR TypeError: "renderer is undefined"
    Angular 3
        isProceduralRenderer
        elementPropertyInternal
        updateSyntheticHostBinding
</code></pre>

## 🌍  Your Environment

**Angular Version:**
<pre><code>

                |___/

Angular CLI: 9.0.2
Node: 12.14.1
OS: darwin x64

Angular: 9.0.1
... router
Ivy Workspace: Yes

@angular-devkit/architect         0.900.2
@angular-devkit/build-angular     0.900.2
@angular-devkit/build-optimizer   0.900.2
@angular-devkit/build-webpack     0.900.2
@angular-devkit/core              9.0.2
@angular-devkit/schematics        9.0.2
@angular/cli                      9.0.2
@ngtools/webpack                  9.0.2
@schematics/angular               9.0.2
@schematics/update                0.900.2

</code></pre>

**Anything else relevant?**
Was working as expected prior to IVY.

AndrewKushnir:
Prior to this change, animations-related runtime logic assumed that the `@HostBinding` and `@HostListener` with synthetic (animations) props are used for Components only. However having `@HostBinding` and `@HostListener` with synthetic props on Directives is supported by View Engine. This commit updates the logic to select correct renderer to execute instructions (current renderer for Directives and sub-component renderer for Components).

This PR resolves #35501.

## PR Type

- [x] Bugfix

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/render3/instructions/host_property.ts
+++ b/packages/core/src/render3/instructions/host_property.ts
@@ -8,7 +8,7 @@
 import {bindingUpdated} from '../bindings';
 import {SanitizerFn} from '../interfaces/sanitization';
 import {RENDERER} from '../interfaces/view';
-import {getLView, getSelectedTNode, getTView, nextBindingIndex} from '../state';
+import {getCurrentDirectiveDef, getLView, getSelectedTNode, getTView, nextBindingIndex} from '../state';
 import {NO_CHANGE} from '../tokens';
 
 import {elementPropertyInternal, loadComponentRenderer, storePropertyBindingMetadata} from './shared';
@@ -42,7 +42,7 @@ export function ɵɵhostProperty<T>(
 
 
 /**
- * Updates a synthetic host binding (e.g. `[@foo]`) on a component.
+ * Updates a synthetic host binding (e.g. `[@foo]`) on a component or directive.
  *
  * This instruction is for compatibility purposes and is designed to ensure that a
  * synthetic host binding (e.g. `@HostBinding('@foo')`) properly gets rendered in
@@ -70,7 +70,8 @@ export function ɵɵupdateSyntheticHostBinding<T>(
   if (bindingUpdated(lView, bindingIndex, value)) {
     const tView = getTView();
     const tNode = getSelectedTNode();
-    const renderer = loadComponentRenderer(tNode, lView);
+    const currentDef = getCurrentDirectiveDef(tView.data);
+    const renderer = loadComponentRenderer(currentDef, tNode, lView);
     elementPropertyInternal(tView, tNode, lView, propName, value, renderer, sanitizer, true);
     ngDevMode && storePropertyBindingMetadata(tView.data, tNode, propName, bindingIndex);
   }
--- a/packages/core/src/render3/instructions/listener.ts
+++ b/packages/core/src/render3/instructions/listener.ts
@@ -15,7 +15,7 @@ import {GlobalTargetResolver, isProceduralRenderer, RElement, Renderer3} from '.
 import {isDirectiveHost} from '../interfaces/type_checks';
 import {CLEANUP, FLAGS, LView, LViewFlags, RENDERER, TView} from '../interfaces/view';
 import {assertNodeOfPossibleTypes} from '../node_assert';
-import {getLView, getPreviousOrParentTNode, getTView} from '../state';
+import {getCurrentDirectiveDef, getLView, getPreviousOrParentTNode, getTView} from '../state';
 import {getComponentLViewByIndex, getNativeByTNode, unwrapRNode} from '../util/view_utils';
 
 import {getLCleanup, handleError, loadComponentRenderer, markViewDirty} from './shared';
@@ -48,7 +48,7 @@ export function ɵɵlistener(
 }
 
 /**
- * Registers a synthetic host listener (e.g. `(@foo.start)`) on a component.
+ * Registers a synthetic host listener (e.g. `(@foo.start)`) on a component or directive.
  *
  * This instruction is for compatibility purposes and is designed to ensure that a
  * synthetic host listener (e.g. `@HostListener('@foo.start')`) properly gets rendered
@@ -73,8 +73,9 @@ export function ɵɵcomponentHostSyntheticListener(
     eventTargetResolver?: GlobalTargetResolver): typeof ɵɵcomponentHostSyntheticListener {
   const tNode = getPreviousOrParentTNode();
   const lView = getLView();
-  const renderer = loadComponentRenderer(tNode, lView);
   const tView = getTView();
+  const currentDef = getCurrentDirectiveDef(tView.data);
+  const renderer = loadComponentRenderer(currentDef, tNode, lView);
   listenerInternal(
       tView, lView, renderer, tNode, eventName, listenerFn, useCapture, eventTargetResolver);
   return ɵɵcomponentHostSyntheticListener;
--- a/packages/core/src/render3/instructions/shared.ts
+++ b/packages/core/src/render3/instructions/shared.ts
@@ -31,7 +31,7 @@ import {isComponentDef, isComponentHost, isContentQueryHost, isLContainer, isRoo
 import {CHILD_HEAD, CHILD_TAIL, CLEANUP, CONTEXT, DECLARATION_COMPONENT_VIEW, DECLARATION_VIEW, FLAGS, HEADER_OFFSET, HOST, InitPhaseState, INJECTOR, LView, LViewFlags, NEXT, PARENT, RENDERER, RENDERER_FACTORY, RootContext, RootContextFlags, SANITIZER, T_HOST, TData, TVIEW, TView, TViewType} from '../interfaces/view';
 import {assertNodeOfPossibleTypes} from '../node_assert';
 import {isInlineTemplate, isNodeMatchingSelectorList} from '../node_selector_matcher';
-import {enterView, getBindingsEnabled, getCheckNoChangesMode, getIsParent, getPreviousOrParentTNode, getSelectedIndex, getTView, leaveView, setBindingIndex, setBindingRootForHostBindings, setCheckNoChangesMode, setCurrentQueryIndex, setPreviousOrParentTNode, setSelectedIndex} from '../state';
+import {enterView, getBindingsEnabled, getCheckNoChangesMode, getCurrentDirectiveIndex, getIsParent, getPreviousOrParentTNode, getSelectedIndex, getTView, leaveView, setBindingIndex, setBindingRootForHostBindings, setCheckNoChangesMode, setCurrentDirectiveIndex, setCurrentQueryIndex, setPreviousOrParentTNode, setSelectedIndex} from '../state';
 import {NO_CHANGE} from '../tokens';
 import {isAnimationProp, mergeHostAttrs} from '../util/attrs_utils';
 import {INTERPOLATION_DELIMITER, renderStringify, stringifyForError} from '../util/misc_utils';
@@ -1272,11 +1272,13 @@ function invokeDirectivesHostBindings(tView: TView, lView: LView, tNode: TNode) 
   const expando = tView.expandoInstructions!;
   const firstCreatePass = tView.firstCreatePass;
   const elementIndex = tNode.index - HEADER_OFFSET;
+  const currentDirectiveIndex = getCurrentDirectiveIndex();
   try {
     setSelectedIndex(elementIndex);
-    for (let i = start; i < end; i++) {
-      const def = tView.data[i] as DirectiveDef<any>;
-      const directive = lView[i];
+    for (let dirIndex = start; dirIndex < end; dirIndex++) {
+      const def = tView.data[dirIndex] as DirectiveDef<any>;
+      const directive = lView[dirIndex];
+      setCurrentDirectiveIndex(dirIndex);
       if (def.hostBindings !== null || def.hostVars !== 0 || def.hostAttrs !== null) {
         invokeHostBindingsInCreationMode(def, directive);
       } else if (firstCreatePass) {
@@ -1285,6 +1287,7 @@ function invokeDirectivesHostBindings(tView: TView, lView: LView, tNode: TNode) 
     }
   } finally {
     setSelectedIndex(-1);
+    setCurrentDirectiveIndex(currentDirectiveIndex);
   }
 }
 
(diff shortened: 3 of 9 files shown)


packages/core/src/render3/instructions/shared.ts
```diff
@@ -1263,11 +1263,13 @@ function invokeDirectivesHostBindings(tView: TView, lView: LView, tNode: TNode)
...
+  const currentDirectiveIndex = getCurrentDirectiveIndex();
   try {
     setSelectedIndex(elementIndex);
-    for (let i = start; i < end; i++) {
-      const def = tView.data[i] as DirectiveDef<any>;
-      const directive = lView[i];
+    for (let dirIndex = start; dirIndex < end; dirIndex++) {
+      const def = tView.data[dirIndex] as DirectiveDef<any>;
+      const directive = lView[dirIndex];
+      setCurrentDirectiveIndex(dirIndex);
```
kara: Hmm this means that every host binding, regardless of whether there are any synthetic host bindings, will pay the cost of keeping track of the directive index global, right? If I read the comment right, we can't generate something different at compile-time because we don't know if a directive might be extended by a component....

Can we run benchmarks on this change to see impact on apps with lots of host bindings? If we don't have a micro-benchmark for host bindings, we should create one.

packages/core/src/render3/instructions/host_property.ts
```diff
@@ -68,8 +70,17 @@ export function ɵɵupdateSyntheticHostBinding<T>(
...
+    // avoid passing `loadComponentRenderer` function to `elementPropertyInternal` call), since
+    // this is where animations are defined.
+    // Note: this logic exists at runtime (not at compile time), since a Component can inherit a
+    // Directive (hence inherit its host bindings) and it would require loading component's
+    // renderer.
+    const loadComponentRendererFn =
+        currentDef === null || isComponentDef(currentDef) ? loadComponentRenderer : null;
     elementPropertyInternal(
-        tView, lView, nodeIndex, propName, value, sanitizer, true, loadComponentRenderer);
+        tView, lView, nodeIndex, propName, value, sanitizer, true, loadComponentRendererFn);
```
kara: If there is a component with a directive on the same node, and the directive has a synthetic host binding, I take it that the correct renderer is *not* the component renderer? Is this why we can't just fall back to `lView[RENDERER]` if the component view isn't found?  Can we add a test that showcases this?

kara (commented):
Mostly questions

packages/core/test/acceptance/host_binding_spec.ts
```diff
@@ -175,6 +178,266 @@ describe('host bindings', () => {
     });
   });
 
+  describe('with synthetic (animations) props', () => {
```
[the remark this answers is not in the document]
AndrewKushnir: 👍

packages/core/src/render3/instructions/shared.ts
AndrewKushnir: Thanks for the review @kara. We don't seem to have micro benchmarks for host bindings, so I created one (in a separate PR): https://github.com/angular/angular/pull/35705. Once that PR lands, I will measure perf impact of these changes. Thank you.

AndrewKushnir:
Hi @kara, I performed performance measurements using newly added benchmark for host bindings. I tested 2 scenarios:

- 1 directive with 1 host binding + 1 host listeners attached to an element. The results (master vs this branch) are almost identical.
- 100 directives with 1 host binding + 1 host listeners attached to an element (to emulate a case that you mentioned in the comment). The result of this change is a bit more noticeable, but still low - around **0.5%** (12.33us vs 12.26us in best cases).

Thank you.

pkozlowski-opensource:
I'm putting this PR in the "cleanup" state for a tiny bit since with @AndrewKushnir we've observed some funky numbers when running perf tests. We talked and for 99,99% this is just a case of the rebase that needs to be done, but I will confirm this tomorrow when the rebase is done.

AndrewKushnir:
Thanks for looking at the changes in this PR @pkozlowski-opensource! I've rebased this PR, so it should be up-to-date with the current master (the `host_binding` benchmark should be the same now). Thank you.

packages/core/test/acceptance/host_binding_spec.ts
```diff
@@ -6,10 +6,13 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
+import {animate, state, style, transition, trigger} from '@angular/animations';
```
mhevery: nit: `animate` is un-needed import

packages/core/src/render3/instructions/host_property.ts
mhevery: Discussed with @AndrewKushnir offline
- [ ] We think that `elementPropertyInternal` should not take `loadComponentRendererFn` but instead take the `Renderer` directly. The implication of this is that the renderer lookup logic should move from `elementPropertyInternal` to `ɵɵupdateSyntheticHostBinding`
- [ ] The above will require that instead of passing `nodeIndex` we pass in `TNode`. Implication is that `const nodeIndex = getSelectedIndex();` should change to `const tNode = getSelectedTNode();`

packages/core/src/render3/instructions/listener.ts
```diff
@@ -72,8 +74,15 @@ export function ɵɵcomponentHostSyntheticListener(
...
   const tNode = getPreviousOrParentTNode();
   const lView = getLView();
-  const renderer = loadComponentRenderer(tNode, lView);
   const tView = getTView();
+  const currentDef = getCurrentHostDirectiveDef(tView.data);
+  // For synthetic host listeners defined on Directives, use current component's renderer, since
+  // this is where animations are defined.
+  // Note: this logic exists at runtime (not at compile time), since a Component can inherit a
+  // Directive (hence inherit its host listeners) and it would require loading component's renderer.
+  const renderer = currentDef === null || isComponentDef(currentDef) ?
```
mhevery: this logic is same as in `ɵɵupdateSyntheticHostBinding`. Perhaps we could just move that into `loadComponentRenderer` directly?

pkozlowski-opensource:
@AndrewKushnir I can confirm that after the rebase the perf results from master and your PR are about the same so it is all good 👍 

I will do another review pass after you address Misko's review.

packages/core/src/render3/instructions/host_property.ts
```diff
@@ -70,7 +70,8 @@ export function ɵɵupdateSyntheticHostBinding<T>(
   if (bindingUpdated(lView, bindingIndex, value)) {
     const tView = getTView();
     const tNode = getSelectedTNode();
-    const renderer = loadComponentRenderer(tNode, lView);
+    const currentDef = getCurrentDirectiveDef(tView.data);
+    const renderer = loadComponentRenderer(currentDef, tNode, lView);
```
mhevery: I wonder if we could do this a different way, can we discuss this over VC.

AndrewKushnir:
@mhevery @matsko thanks for the help and additional information. I've updated this PR to include extra comments to describe the `currentDef === null` check (due to root component creation process). Could you please have a look at this PR when you get a chance? Thank you.

AndrewKushnir:
[Presubmit](http://test/OCL:304101930:BASE:304101938:1585711363554:fdc07289)

mhevery (approved):

packages/core/src/render3/instructions/shared.ts
```diff
@@ -1928,9 +1931,12 @@ function getTViewCleanup(tView: TView): any[] {
  * There are cases where the sub component's renderer needs to be included
  * instead of the current renderer (see the componentSyntheticHost* instructions).
  */
-export function loadComponentRenderer(tNode: TNode, lView: LView): Renderer3 {
-  const componentLView = unwrapLView(lView[tNode.index]) !;
-  return componentLView[RENDERER];
+export function loadComponentRenderer(
+    currentDef: DirectiveDef<any>| null, tNode: TNode, lView: LView): Renderer3 {
+  if (currentDef === null || isComponentDef(currentDef)) {
```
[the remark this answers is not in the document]
AndrewKushnir: @mhevery there is a comment above this check, please let me know if you want me to include additional info or rephrase it. Thank you.

integration/_payload-limits.json
```diff
@@ -49,7 +49,7 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 2289,
-        "main-es2015": 226144,
+        "main-es2015": 226686,
```
AndrewKushnir, line 52: This looks like an accumulated size increase, since the same happened in a [different PR (#36381)](https://github.com/angular/angular/pull/36381/files#diff-55315ab5c9e23bee7dc42a45a3517105R52).

kara:
@AndrewKushnir Can you update payload limits?

AndrewKushnir:
@kara thanks for the note. I've just updated payload limits: the payload size limit in integration tests in the `master` branch was higher that the one that was set in this PR (so I reverted my change) :) However there is still a payload limit increase for AIO, which is an accumulated increase from previous changes. I've also added a second commit with clang-related formatting changes. Thank you.

packages/core/src/render3/instructions/shared.ts
```diff
@@ -1260,11 +1260,13 @@ function invokeDirectivesHostBindings(tView: TView, lView: LView, tNode: TNode)
...
   const firstCreatePass = tView.firstCreatePass;
   const elementIndex = tNode.index - HEADER_OFFSET;
+  const currentDirectiveIndex = getCurrentDirectiveIndex();
   try {
     setSelectedIndex(elementIndex);
-    for (let i = start; i < end; i++) {
-      const def = tView.data[i] as DirectiveDef<any>;
-      const directive = lView[i];
+    for (let dirIndex = start; dirIndex < end; dirIndex++) {
+      const def = tView.data[dirIndex] as DirectiveDef<any>;
```
matsko, line 1267: could we get away with using `DirectiveRef<unkown>` here? That way the behavior of `any` won't seep into the framework.

packages/core/src/render3/instructions/shared.ts
```diff
@@ -1273,6 +1275,7 @@ function invokeDirectivesHostBindings(tView: TView, lView: LView, tNode: TNode)
     }
   } finally {
     setSelectedIndex(-1);
+    setCurrentDirectiveIndex(currentDirectiveIndex);
```
matsko, line 1278: Is there a reason why need this?

matsko: Is it because we no longer rely on the state being cleaned up once the `LFrame` is out of use?

packages/core/src/render3/state.ts
```diff
@@ -356,6 +357,21 @@ export function getCurrentDirectiveIndex(): number {
   return instructionState.lFrame.currentDirectiveIndex;
 }
 
+export function setCurrentDirectiveIndex(currentDirectiveIndex: number): void {
```
matsko, line 360: Can you add some documentation for this?

matsko (changes requested):

packages/core/src/render3/instructions/shared.ts
AndrewKushnir: I've used `DirectiveRef<unknown>` here as you proposed.
The `DirectiveRef<any>` is used more widely in codebase, but I guess that's mainly for historical reasons.

packages/core/src/render3/instructions/shared.ts
AndrewKushnir: The goal is to keep the global `state` in a clean state once we are done with host bindings processing (since there are other operations that happen afterwards while we process the view).

matsko (approved):

AndrewKushnir:
[Presubmit](http://test/OCL:306702047:BASE:306702106:1586980511193:fc2bc634).

packages/core/test/acceptance/host_binding_spec.ts
```diff
@@ -175,6 +178,274 @@ describe('host bindings', () => {
...
+           imports: [NoopAnimationsModule],
+         });
+         const fixture = TestBed.createComponent(Comp);
+         fixture.detectChanges();
+         await fixture.whenStable();  // wait for animations to complete
+         expect(fixture.nativeElement.style.color).toBe('yellow');
+         expect(events).toEqual(['@myAnimation.start', '@myAnimation.done']);
+       });
+  });
+
```
kara, line 448: I don't see a test for a component with a directive on the same node - looks like https://github.com/angular/angular/pull/35568#discussion_r382349635 still needs to be addressed?

kara (approved):
LGTM once test is added

AndrewKushnir:
[Global Presubmit](https://test.corp.google.com/ui#id=OCL:306702047:BASE:307525392:1587436783212:38d20375)

AndrewKushnir:
[Presubmit](http://test/OCL:308670981:BASE:308671028:1588014546093:bf8fee70)

PR #35592, opened 2020-02-20
fix(ngcc): capture path-mapped entry-points that start with same string

### LINKED ISSUE #35536: Error NG6002, caused by enableIvy with libraries of a specific naming scheme
# 🐞 bug report

### Affected Package

 

The issue is caused by package `@angular/compiler` maybe? ngcc

### Is this a regression?

Yes, the previous version in which this bug was not present was: 8.x.x

Introduced by adding `"enableIvy": true` to an application in a monorepo.

### Description

The following error occurs under a very specific setup in an Angular CLI monorepo:

```
error NG6002: Appears in the NgModule.imports of AppModule, but could not be resolved to an NgModule class
```

This is related to several other reported issues, but never under this scenario AFAIK:

* Create a monorepo, where a library name matches a substring of another repo (ie: "**my-lib**" and "**my-lib**-second").
* Build the libraries with `"enableIvy": false` .
* Build the app, referencing both libraries in their **dist** folder.

## 🔬 Minimal Reproduction

https://github.com/zbream/angular-ivy-paths-issue

Pull, follow the README.

## 🔥 Exception or Error
<pre><code>
ERROR in dist/my-lib-second/lib/my-lib-second.module.d.ts:1:22 - error NG6002: Appears in the NgModule.imports of AppModule, but could not be resolved to an NgModule class

1 export declare class MyLibSecondModule {
                       ~~~~~~~~~~~~~~~~~
</code></pre>

## 🌍  Your Environment

**Angular Version:**
<pre><code>
<!-- run `ng version` and paste output below -->

                |___/


Angular CLI: 9.0.2
Node: 12.14.0
OS: win32 x64

Angular: 9.0.1
... router
Ivy Workspace: Yes

@angular-devkit/architect          0.900.2
@angular-devkit/build-angular      0.900.2
@angular-devkit/build-ng-packagr   0.900.2
@angular-devkit/build-optimizer    0.900.2
@angular-devkit/build-webpack      0.900.2
@angular-devkit/core               9.0.2
@angular-devkit/schematics         9.0.2
@angular/cli                       9.0.2
@ngtools/webpack                   9.0.2
@schematics/angular                9.0.2
@schematics/update                 0.900.2

</code></pre>

**Anything else relevant?**
<!-- ✍️Is this a browser specific issue? If so, please specify the browser and version. -->

<!-- ✍️Do any of these matter: operating system, IDE, package manager, HTTP server, ...? If so, please mention it below. -->

I took note of some issues I noticed in the result of `ngcc` [in the repo](https://github.com/zbream/angular-ivy-paths-issue#observations).

petebacondarwin:
Previously if there were two path-mapped libraries that are in
different directories but the path of one started with same string
as the path of the other, we would incorrectly return the shorter
path - e.g. `dist/my-lib` and `dist/my-lib-second`. This was because
the list of `basePaths` was searched in ascending alphabetic order and
we were using `startsWith()` to match the path.

Now the `basePaths` are searched in reverse alphabetic order so the
longer path will be matched correctly.

// FW-1873

Fixes #35536

### DIFF
--- a/packages/compiler-cli/ngcc/src/entry_point_finder/utils.ts
+++ b/packages/compiler-cli/ngcc/src/entry_point_finder/utils.ts
@@ -5,7 +5,7 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
-import {AbsoluteFsPath, getFileSystem, join, resolve} from '../../../src/ngtsc/file_system';
+import {AbsoluteFsPath, getFileSystem, join, relative, resolve} from '../../../src/ngtsc/file_system';
 import {PathMappings} from '../utils';
 
 /**
@@ -42,7 +42,7 @@ export function getBasePaths(
       basePaths.push(basePath);
     }));
   }
-  basePaths.sort();  // Get the paths in order with the shorter ones first.
+  basePaths.sort().reverse();  // Get the paths in order with the longer ones first.
   return basePaths.filter(removeDeeperPaths);
 }
 
@@ -56,16 +56,19 @@ function extractPathPrefix(path: string) {
 }
 
 /**
- * A filter function that removes paths that are already covered by higher paths.
+ * A filter function that removes paths that are contained by other paths.
+ *
+ * For example given `['a/b/c', 'a/b', 'd/e', 'd']` we will end up with `['a/b', 'd']`.
+ *
+ * We only need to check the following path since the `array` is sorted in reverse alphabetic order.
  *
  * @param value The current path.
  * @param index The index of the current path.
- * @param array The array of paths (sorted alphabetically).
- * @returns true if this path is not already covered by a previous path.
+ * @param array The array of paths (sorted in reverse alphabetical order).
+ * @returns true if this path is not contained by another path.
  */
 function removeDeeperPaths(value: AbsoluteFsPath, index: number, array: AbsoluteFsPath[]) {
-  for (let i = 0; i < index; i++) {
-    if (value.startsWith(array[i])) return false;
-  }
-  return true;
+  // Use `relative()` rather than `startsWith()` to avoid false positives for paths like
+  // `abc/defg` and `abc/def` since the former is not contained by the latter.
+  return index === array.length - 1 || relative(value, array[index + 1]).startsWith('../');
 }
--- a/packages/compiler-cli/ngcc/src/entry_point_finder/utils.ts
+++ b/packages/compiler-cli/ngcc/src/entry_point_finder/utils.ts
@@ -43,7 +43,7 @@ export function getBasePaths(
     }));
   }
   basePaths.sort().reverse();  // Get the paths in order with the longer ones first.
-  return basePaths.filter(removeDeeperPaths);
+  return basePaths.filter(removeContainedPaths);
 }
 
 /**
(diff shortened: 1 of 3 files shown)


packages/compiler-cli/ngcc/src/entry_point_finder/utils.ts
```diff
@@ -56,16 +56,19 @@ function extractPathPrefix(path: string) {
...
+ * @returns true if this path is not contained by another path.
  */
 function removeDeeperPaths(value: AbsoluteFsPath, index: number, array: AbsoluteFsPath[]) {
-  for (let i = 0; i < index; i++) {
-    if (value.startsWith(array[i])) return false;
-  }
-  return true;
+  // Use `relative()` rather than `startsWith()` to avoid false positives for paths like
+  // `abc/defg` and `abc/def` since the former is not contained by the latter.
+  return index === array.length - 1 || relative(value, array[index + 1]).startsWith('../');
```
gkalpak: 1. AFAICT, `relative('/foo/bar', '/foo')` will return `..` (without a trailing `/`) and thus the `.startsWith('../')` check will fail.
  (Although, I amy have missed something, because I would expect a test to be failing due to this 😕)

2. By not checking against all following paths, we will incorrectly retain `a/b/d` in `['a/b/d', 'a/b/c', 'a/b']`.

gkalpak (changes requested):

packages/compiler-cli/ngcc/src/entry_point_finder/utils.ts
petebacondarwin: 😱

That'll teach me for holiday coding

😇

I'll fix that  ASAP

petebacondarwin:
@gkalpak - fixed the code and tightened up the test. PTAL

packages/compiler-cli/ngcc/src/entry_point_finder/utils.ts
```diff
@@ -58,17 +58,23 @@ function extractPathPrefix(path: string) {
 /**
  * A filter function that removes paths that are contained by other paths.
  *
- * For example given `['a/b/c', 'a/b', 'd/e', 'd']` we will end up with `['a/b', 'd']`.
- *
- * We only need to check the following path since the `array` is sorted in reverse alphabetic order.
+ * For example:
+ * Given `['a/b/c', 'a/b/x', 'a/b', 'd/e', 'd/f']` we will end up with `['a/b', 'd/e', 'd/f]`.
+ * (Note that we do not get `d` even though `d/e` and `d/f` share a base directory, since `d` is not
+ * one of the base paths.)
```
gkalpak: 💯

gkalpak (approved):
👌

JoostK (approved):

fkolar:
@petebacondarwin Is this already in 9.0.6. ?

fkolar:
I can confirm this does not fix ##35536, as I need to run `ngcc && ngcc -s dist` on my local libs so I can use them in apps.

petebacondarwin:
It was included in 9.0.3 - https://github.com/angular/angular/blob/master/CHANGELOG.md#903-2020-02-27

@fkolar Please provide a reproduction of your problem in a new issue for us to investigate.

fkolar:
Let me comit my changes.. I will create new issues. thanks allot.

PR #35593, opened 2020-02-20
fix(localize): improve placeholder mismatch error message

petebacondarwin:
The original error message was confusing since often it is the
translation that is at fault not the message.

cc @ocombe

### DIFF
--- a/packages/localize/src/utils/src/translations.ts
+++ b/packages/localize/src/utils/src/translations.ts
@@ -68,7 +68,8 @@ export function translate(
         return message.substitutions[placeholder];
       } else {
         throw new Error(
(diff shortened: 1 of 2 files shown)


AndrewKushnir (approved):
LGTM, thanks for improving error messages Pete! 👍

ocombe (approved):
Much more clear, that being said it might be a bit too verbose? Text from error messages isn't minifiable, this means that you increase the size of the framework if you use error messages that are too long

petebacondarwin:
Of course if you use compile time translation, this will never get to the browser.

Feel free to offer an improvement :smiley: if you are concerned about the runtime scenario.

It just needs to be neutral as regards which side is wrong.

PR #35597, opened 2020-02-21
TEST ONLY: revert recent ngcc changes

AndrewKushnir:
DO NOT MEGRE.

This PR reverts 3 commits that triggered `integration_test` CircleCI job to start failing. The goal of this PR is to see whether these commits actually affect `integration_test` CircleCI job or there are other issue related to CI and ngcc changes are ok.


AndrewKushnir:
Result: CircleCI fails with the same error message even with these commits reverted.

PR #35619, opened 2020-02-21
refactor(compiler): Remove NullAstVisitor and visitAstChildren

kyliau:
This commit removes the `NullAstVisitor` and `visitAstChildren` exported
from `packages/compiler/src/expression_parser/ast.ts` because they
contain duplicate and buggy implementation, and their use cases could be
sufficiently covered by `RecursiveAstVisitor` if the latter implements the
`visit` method. This use case is only needed in the language service.

With this change, any visitor that extends `RecursiveAstVisitor` could
just define their own `visit` function and the parent class will behave
correctly.

### A bit of historical context:
In language service, we need a way to tranverse the expression AST in a
selective manner based on where the user's cursor is. This means we need a
"filtering" function to decide which node to visit and which node to not
visit. Instead of refactoring `RecursiveAstVisitor` to support this,
`visitAstChildren` was created. `visitAstChildren` duplicates the
implementation of `RecursiveAstVisitor`, but introduced some bugs along
the way. For example, in `visitKeyedWrite`, it visits
```
obj -> key -> obj
```
instead of
```
obj -> key -> value
```

Moreover, because of the following line
```
visitor.visit && visitor.visit(ast, context) || ast.visit(visitor, context);
```
`visitAstChildren` visits every node **twice**.

### DIFF
--- a/packages/compiler/src/expression_parser/ast.ts
+++ b/packages/compiler/src/expression_parser/ast.ts
@@ -311,109 +311,85 @@ export interface AstVisitor {
   visitSafeMethodCall(ast: SafeMethodCall, context: any): any;
   visitSafePropertyRead(ast: SafePropertyRead, context: any): any;
   visitASTWithSource?(ast: ASTWithSource, context: any): any;
+  /**
+   * This function is optionally defined to allow classes that implement this
+   * interface to selectively decide if the specified `ast` should be visited.
+   * @param ast node to visit
+   * @param context context that gets passed to the node and all its children
+   */
   visit?(ast: AST, context?: any): any;
 }
 
-export class NullAstVisitor implements AstVisitor {
-  visitBinary(ast: Binary, context: any): any {}
-  visitChain(ast: Chain, context: any): any {}
-  visitConditional(ast: Conditional, context: any): any {}
-  visitFunctionCall(ast: FunctionCall, context: any): any {}
-  visitImplicitReceiver(ast: ImplicitReceiver, context: any): any {}
-  visitInterpolation(ast: Interpolation, context: any): any {}
-  visitKeyedRead(ast: KeyedRead, context: any): any {}
-  visitKeyedWrite(ast: KeyedWrite, context: any): any {}
-  visitLiteralArray(ast: LiteralArray, context: any): any {}
-  visitLiteralMap(ast: LiteralMap, context: any): any {}
-  visitLiteralPrimitive(ast: LiteralPrimitive, context: any): any {}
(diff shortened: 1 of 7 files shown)


packages/compiler/test/expression_parser/ast_spec.ts
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {AST, Lexer, Parser, RecursiveAstVisitor} from '@angular/compiler';
+import {ImplicitReceiver, MethodCall, PropertyRead} from '@angular/compiler/src/compiler';
+
+describe('RecursiveAstVisitor', () => {
+  it('should call visit() on every node', () => {
```
ayazhafiz:
```suggestion
  it('should visit every node', () => {
```

nit: this tests the expected result, not if visit() is actually called

ayazhafiz (approved):

packages/compiler/src/expression_parser/ast.ts
```diff
@@ -311,109 +311,85 @@ export interface AstVisitor {
   visitSafeMethodCall(ast: SafeMethodCall, context: any): any;
   visitSafePropertyRead(ast: SafePropertyRead, context: any): any;
   visitASTWithSource?(ast: ASTWithSource, context: any): any;
+  /**
+   * This function is optionally defined to allow classes that implement this
+   * interface to selectively decide if the specified `ast` should be visited.
+   * @param ast
+   * @param context
```
ayazhafiz: prefer elaborating or removing these comments, I feel like they aren't very useful on their own

packages/compiler/test/expression_parser/ast_spec.ts
```diff
@@ -0,0 +1,42 @@
...
+
+class Visitor extends RecursiveAstVisitor {
+  visit(node: AST, path: AST[]) {
+    path.push(node);
+    node.visit(this, path);
+  }
+}
+
+type Newable = new (...args: any) => any;
+function assertType<T extends Newable>(val: any, t: T): asserts val is InstanceType<T> {
```
ayazhafiz:
```suggestion
function expectType<T extends Newable>(val: any, t: T): asserts val is InstanceType<T> {
```

packages/compiler/src/expression_parser/ast.ts
kyliau: done

ayazhafiz (approved):

alxhub (approved):

Issue #35630, opened 2020-02-22
We cannot connect to angular.iodocs server


Elkaboverde:
# 📚 Docs or angular.io bug report

### Description

When change version from "stable (v9.0.2)" to "stable (v9.0.2)' URL changet to "https://angular.iodocs/" and
When change version from "stable (v9.0.2)" to "next" URL changet to "https://next.angular.iodocs/";

### What's the affected URL?**
 https://angular.io/docs

### Reproduction Steps**
change version from "stable (v9.0.2)" to "stable (v9.0.2)' ||
change version from "stable (v9.0.2)" to "next"

## 📷Screenshot

![angular io](https://user-images.githubusercontent.com/26579451/75099502-70f84280-55d3-11ea-95c2-230046e72c1a.png)

## 🔥 Exception or Error
<pre><code>

</code></pre>

### Browser info
Not browser specific. But I tested in Google Chrome v80.0.3987.116  and Firefox: Developer Edition v74.0b6.
https://angular.io/ : Version 9.0.3-local+sha.bb09cd0e41.

sonukapoor:
I will take a look.

PR #35318, opened 2020-02-11
fix(docs-infra): preserves query and hash when switching angular versions

sonukapoor:
Previously, when switching angular versions through the
version selector in the sidenav, the path is lost, and the
user has to manually navigate to the original location.

This commit fixes this issue and maintains the `pathname`
when navigating between different versions.

Closes #24495

## PR Checklist
Please check if your PR fulfills the following requirements:

- [X] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [X] Tests for the changes have been added (for bug fixes / features)
- [X] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [X] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] angular.io application / infrastructure changes
- [ ] Other... Please describe:

## What is the current behavior?

Issue Number: #24495

## Does this PR introduce a breaking change?

- [ ] Yes
- [X] No

PR #35632, opened 2020-02-22
fix(docs-infra): fix redirect in angular version selector

sonukapoor:
## PR Checklist

## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #35630

- [ ] Yes
- [X] No

### DIFF
--- a/aio/src/app/app.component.spec.ts
+++ b/aio/src/app/app.component.spec.ts
@@ -417,6 +417,15 @@ describe('AppComponent', () => {
         selectElement.triggerEventHandler('change', { option: versionWithoutUrl, index: versionWithoutUrlIndex });
         expect(locationService.go).not.toHaveBeenCalled();
       });
+
+      it('should navigate when change to a version with a url that does not end with `/`', async () => {
+        await setupSelectorForTesting();
+        locationService.urlSubject.next('docs#section-1');
+        const versionWithoutSlashIndex = component.docVersions.length;
+        const versionWithoutSlashUrl = component.docVersions[versionWithoutSlashIndex] = { url: 'https://next.angular.io', title: 'foo' };
+        selectElement.triggerEventHandler('change', { option: versionWithoutSlashUrl, index: versionWithoutSlashIndex });
+        expect(locationService.go).toHaveBeenCalledWith('https://next.angular.io/docs#section-1');
+      });
     });
 
     describe('currentDocument', () => {
--- a/aio/src/app/app.component.ts
+++ b/aio/src/app/app.component.ts
@@ -149,8 +149,8 @@ export class AppComponent implements OnInit {
     ]).subscribe(([versionInfo, versions]) => {
       // TODO(pbd): consider whether we can lookup the stable and next versions from the internet
       const computedVersions: NavigationNode[] = [
-        { title: 'next', url: 'https://next.angular.io' },
-        { title: 'stable', url: 'https://angular.io' },
+        { title: 'next', url: 'https://next.angular.io/' },
+        { title: 'stable', url: 'https://angular.io/' },
       ];
       if (this.deployment.mode === 'archive') {
         computedVersions.push({ title: `v${versionInfo.major}` });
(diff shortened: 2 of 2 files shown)


aio/src/app/app.component.ts
```diff
@@ -149,7 +149,7 @@ export class AppComponent implements OnInit {
     ]).subscribe(([versionInfo, versions]) => {
       // TODO(pbd): consider whether we can lookup the stable and next versions from the internet
       const computedVersions: NavigationNode[] = [
-        { title: 'next', url: 'https://next.angular.io' },
+        { title: 'next', url: 'https://next.angular.io/' },
         { title: 'stable', url: 'https://angular.io' },
```
ericmartinezr: You should change both, as both have the same problem.

gkalpak (changes requested):
From https://github.com/angular/angular/pull/35318#discussion_r379393795:

> Currently, all `docVersions` URLs contains a trailing `/`, but if we wanted to be cautious, we should check (and if necessary add a trailing `/`).

And that's why one should always err on the side of caution :grin:

##
So, in addition to the current changes, I still believe it is better to ensure there is always exactly one `/` between `version.url` and `currentUrl`:

https://github.com/angular/angular/blob/e084835fb111147ac7497036d38a0b5c012d5cd6/aio/src/app/app.component.ts#L235

gkalpak:
Also, please make sure the commit message follows our [guidelines][1] (e.g. in this case include a `Fixes #35630` footer).

[1]: https://github.com/angular/angular/blob/e084835fb/CONTRIBUTING.md#commit

aio/src/app/app.component.ts
```diff
@@ -232,7 +232,8 @@ export class AppComponent implements OnInit {
   onDocVersionChange(versionIndex: number) {
     const version = this.docVersions[versionIndex];
     if (version.url) {
-      this.locationService.go(`${version.url}${this.currentUrl}`);
+      const versionUrl = version.url  + (!version.url.endsWith('/') ? '/' : '');
```
gkalpak: :+1:
Can you also add a test for this?

aio/src/app/app.component.spec.ts
```diff
@@ -417,6 +417,13 @@ describe('AppComponent', () => {
         selectElement.triggerEventHandler('change', { option: versionWithoutUrl, index: versionWithoutUrlIndex });
         expect(locationService.go).not.toHaveBeenCalled();
       });
+
+      it('should navigate when change to a version with a url that does not end with `/`', async () => {
+        const versionWithoutSlashIndex = component.docVersions.length;
+        const versionWithoutSlashUrl = component.docVersions[versionWithoutSlashIndex] = { url: 'https://next.angular.io', title: 'foo' };
+        selectElement.triggerEventHandler('change', { option: versionWithoutSlashUrl, index: versionWithoutSlashIndex });
+        expect(locationService.go).not.toHaveBeenCalledWith('https://next.angular.io/');
```
gkalpak: Can you check with a path (to better highlight what the problem would be if no `/` were added)?

sonukapoor: Sure. Will take a look.

sonukapoor: @gkalpak All done. There are some strange CI errors, that I am not too sure about.

aio/src/app/app.component.spec.ts
```diff
@@ -417,6 +417,14 @@ describe('AppComponent', () => {
         selectElement.triggerEventHandler('change', { option: versionWithoutUrl, index: versionWithoutUrlIndex });
         expect(locationService.go).not.toHaveBeenCalled();
       });
+
+      it('should navigate when change to a version with a url that does not end with `/`', async () => {
+        locationService.urlSubject.next('https://angular.io/docs#section-1');
```
gkalpak: `locationService.urlSubject` is supposed to include the path part only (see other occurrences of `urlSubject.next()` in this file).

gkalpak (changes requested):

aio/src/app/app.component.spec.ts
sonukapoor: My bad. I fixed it. The test was actually not accurate.

gkalpak (approved):
Thank you, sonukapoor ✨

PR #35637, opened 2020-02-23
fix(ivy): error when accessing NgModuleRef.componentFactoryResolver in constructor

### LINKED ISSUE #35580: [IVY]: Fail to get `componentFactoryResolver` from a child NgModuleRef
# 🐞 bug report

### Affected Package

### Is this a regression?

Worked in 8 and below

### Description

When trying to access `componentFactoryResolver` on an instance of `NgModuleRef` that
is not the root module there is an exception thrown:

<details>
<summary>TypeError: Cannot read property 'get' of undefined</summary>
<p>
   at NgModuleRef$1.get (core.umd.js:24253)
    at NgModuleRef$1.get [as componentFactoryResolver] (core.umd.js:24257)
    at new FeatModule (feat-module.ts:8)
    at Object.FeatModule_Factory [as factory] (feat-module.ts:12)
    at R3Injector.hydrate (core.umd.js:11505)
    at R3Injector.get (core.umd.js:11328)
    at eval (core.umd.js:11268)
    at Set.forEach (<anonymous>)
    at new R3Injector (core.umd.js:11268)
    at createInjector (core.umd.js:11233)
</p>
</details>

The issue is that there is no value in the `_r3Injector` property... it exists only in the root module.

## 🔬 Minimal Reproduction
Go to https://stackblitz.com/edit/angular-9-abstract-g6cugw?file=src%2Fapp%2Ffeat-module%2Ffeat-module.ts

There is a debugger statement in line 22, the issue will throw in line 24.

## 🌍  Your Environment

**Angular Version:**
<pre><code>
Angular CLI: 9.0.2
Node: 12.9.1
OS: darwin x64

Angular: 9.0.1
... platform-browser-dynamic, platform-server, router
Ivy Workspace: Yes

Package                                    Version
--------------------------------------------------------------------
@angular-devkit/architect                  0.900.2
@angular-devkit/build-angular              0.900.2
@angular-devkit/build-ng-packagr           0.900.2
@angular-devkit/build-optimizer            0.900.2
@angular-devkit/build-webpack              0.900.2
@angular-devkit/core                       9.0.2
@angular-devkit/schematics                 9.0.2
@angular/cdk                               9.0.0
@angular/cdk-experimental                  9.0.0
@angular/cli                               9.0.2
@angular/flex-layout                       9.0.0-beta.29
@angular/material                          9.0.0
@ngtools/webpack                           9.0.2
@nguniversal/common                        9.0.0
@nguniversal/express-engine                9.0.0
@nguniversal/module-map-ngfactory-loader   8.1.1
@schematics/angular                        9.0.2
@schematics/update                         0.900.2
ng-packagr                                 9.0.1
rxjs                                       6.5.4
typescript                                 3.7.5
webpack                                    4.41.2
</code></pre>

Also worth noting, that if requesting an injection of `ComponentFactoryResolver` it is totally fine.
You get an instance of `ComponentFactoryResolver` that points to it's `NgModuleRef` (`ngModule` prop) and that `NgModuleRef` will **NOT** point to that `ComponentFactoryResolver` we just got... it will throw when trying to access it's `componentFactoryResolver ` property.

crisbeto:
Currently we resolve the `NgModuleRef.componentFactoryResolver` by going through the injector, but the problem is that `ComponentFactoryResolver` has a dependency on `NgModuleRef`, which means that if the module that's attached to the ref tries to inject  `ComponentFactoryResolver` in its constructor, we'll create a circular dependency which throws at runtime.

These changes resolve the issue by creating the `ComponentFactoryResolver` manually ahead of time without going through the injector. We can do this safely, because the only dependency for the resolver is the current module ref which is providing it.

Aside from fixing the issue, another advantage to this approach is that it should reduce the amount of generated JS, because it removes a getter and a provider definition.

Fixes #35580.

### DIFF
--- a/packages/core/src/render3/ng_module_ref.ts
+++ b/packages/core/src/render3/ng_module_ref.ts
@@ -9,7 +9,6 @@
 import {Injector} from '../di/injector';
 import {INJECTOR} from '../di/injector_compatibility';
 import {InjectFlags} from '../di/interface/injector';
-import {StaticProvider} from '../di/interface/provider';
 import {R3Injector, createInjector} from '../di/r3_injector';
 import {Type} from '../interface/type';
 import {ComponentFactoryResolver as viewEngine_ComponentFactoryResolver} from '../linker/component_factory_resolver';
@@ -26,12 +25,6 @@ import {maybeUnwrapFn} from './util/misc_utils';
 
 export interface NgModuleType<T = any> extends Type<T> { ɵmod: NgModuleDef<T>; }
 
-const COMPONENT_FACTORY_RESOLVER: StaticProvider = {
-  provide: viewEngine_ComponentFactoryResolver,
-  useClass: ComponentFactoryResolver,
-  deps: [viewEngine_NgModuleRef],
-};
-
 export class NgModuleRef<T> extends viewEngine_NgModuleRef<T> implements InternalNgModuleRef<T> {
   // tslint:disable-next-line:require-internal-with-underscore
   _bootstrapComponents: Type<any>[] = [];
@@ -40,6 +33,7 @@ export class NgModuleRef<T> extends viewEngine_NgModuleRef<T> implements Interna
   injector: Injector = this;
   instance: T;
   destroyCbs: (() => void)[]|null = [];
+  readonly componentFactoryResolver: ComponentFactoryResolver;
 
   constructor(ngModuleType: Type<T>, public _parent: Injector|null) {
     super();
@@ -49,20 +43,23 @@ export class NgModuleRef<T> extends viewEngine_NgModuleRef<T> implements Interna
                      `NgModule '${stringify(ngModuleType)}' is not a subtype of 'NgModuleType'.`);
 
     const ngLocaleIdDef = getNgLocaleIdDef(ngModuleType);
-    if (ngLocaleIdDef) {
-      setLocaleId(ngLocaleIdDef);
-    }
-
+    ngLocaleIdDef && setLocaleId(ngLocaleIdDef);
     this._bootstrapComponents = maybeUnwrapFn(ngModuleDef !.bootstrap);
-    const additionalProviders: StaticProvider[] = [
-      {
-        provide: viewEngine_NgModuleRef,
-        useValue: this,
-      },
-      COMPONENT_FACTORY_RESOLVER
-    ];
+
+    // When bootstrapping a module we have a dependency graph that looks like this:
+    // ApplicationRef -> ComponentFactoryResolver -> NgModuleRef. The problem is that if the
+    // module being resolved tries to inject the ComponentFactoryResolver, it'll create a
+    // circular dependency which will result in a runtime error, because the injector doesn't
+    // exist yet. We work around the issue by creating the ComponentFactoryResolver ourselves
+    // and providing it, rather than letting the injector resolve it.
+    this.componentFactoryResolver = new ComponentFactoryResolver(this);
     this._r3Injector = createInjector(
-        ngModuleType, _parent, additionalProviders, stringify(ngModuleType)) as R3Injector;
+        ngModuleType, _parent,
+        [
+          {provide: viewEngine_NgModuleRef, useValue: this},
+          {provide: viewEngine_ComponentFactoryResolver, useValue: this.componentFactoryResolver}
+        ],
+        stringify(ngModuleType)) as R3Injector;
     this.instance = this.get(ngModuleType);
   }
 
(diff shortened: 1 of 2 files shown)


shlomiassaf:
@crisbeto Did you try to create the component using the factory itself? (not the testbed)

The problem that I see is in [line 56](https://github.com/angular/angular/pull/35637/files#diff-44fd0871b4117393a79ffaa1bb998e2aR56):

```
 this._r3Injector = createInjector(`  ...
```

Assume you are creating the module-ref for `AppModule` and an import module in `AppModule` (let call it M) requests an injection of `NgModuleRef` which internally will try to create a component using the component factory resolver.

The module-ref for `AppModule` will do

```
 this._r3Injector = createInjector(`  ...
```

Which will instantiate the other modules, including module M... at that point the constructor for module M will run and try to call `create()` on the component factory created by the resolver.

This will throw an error because the first call we are still in (`createInjector`) still did not return, having null in `_r3Injector` thus throwing [HERE](https://github.com/angular/angular/blob/f95b8ce07e71cf0b736cec1f617349a44e41f939/packages/core/src/render3/component_ref.ts#L139-L141)

Since the `ComponentFactory` of `Render3` requires injections from the root view injector (`_r3Injector`) which is still not in place...

I hope i'm clear, not sure at all I am... but just gave it a shot.

crisbeto:
I'm not sure I follow, but changing the test to this seems to work just fine. Maybe the case oversimplified though:

```
it('', () => {
  let factory: ComponentFactory<TestCmp>;

  @NgModule({
    declarations: [TestCmp],
    exports: [TestCmp],
    entryComponents: [TestCmp]  // Only necessary for ViewEngine
  })
  class MyModule {
    constructor(ngModuleRef: NgModuleRef<any>, injector: Injector) {
      factory = ngModuleRef.componentFactoryResolver.resolveComponentFactory(TestCmp);
      const instance = factory.create(injector);

      // Logs "hello"
      console.log(instance.location.nativeElement.textContent);
    }
  }

  TestBed.configureTestingModule({imports: [MyModule]});
  TestBed.createComponent(TestCmp);
  expect(factory !.componentType).toBe(TestCmp);
});
```

shlomiassaf:
@crisbeto Since this is not part of the release yet I can't really post an example or test it properly.

However, try the following:

Create a feature module the is in the app root module.  
In the Ctor of the feature module, inject the component factory resolver and the `Injector` and use them to resolve a component from a 3rd parry lib, let's say material.

```
import { NgModule, ComponentFactoryResolver, Injector } from '@angular/core';
import { CommonModule } from '@angular/common';
import { MatPaginator, MatPaginatorModule } from '@angular/material/paginator';
import { MatSelectModule } from '@angular/material/select';
import { MatTooltipModule } from '@angular/material/tooltip';
import { MatButtonModule } from '@angular/material/button';

@NgModule({
  imports: [ CommonModule, MatPaginatorModule, MatSelectModule, MatTooltipModule, MatButtonModule ],
})
export class MyPaginatorModule {
  constructor(cf: ComponentFactoryResolver, injector: Injector) {
    cf.resolveComponentFactory(MatPaginator).create(injector);
  }
}
``` 

`MyPaginatorModule` should be a sub module of the app root module.

If this works, I guess it's fine... I've tried this locally with your fix and it didn't but that wasn't a pure release of the fix so i'm not 100% sure here.

crisbeto:
It seems to work when injecting the `ComponentFactoryResolver` as well. If you have a repo somewhere I can try out the solution against it.

shlomiassaf:
@crisbeto I guess I'll wait for the release to check it with a complex scenario like I have in my library.

I'll update when and if I find something. Thank you!

packages/core/src/render3/ng_module_ref.ts
```diff
@@ -40,6 +33,7 @@ export class NgModuleRef<T> extends viewEngine_NgModuleRef<T> implements Interna
   injector: Injector = this;
   instance: T;
   destroyCbs: (() => void)[]|null = [];
+  readonly componentFactoryResolver: ComponentFactoryResolver;
```
mhevery: Why not just do this?

```suggestion
  readonly componentFactoryResolver: ComponentFactoryResolver = new ComponentFactoryResolver(this);
```

mhevery (approved):
with cleanup

crisbeto:
I've addressed the feedback.

mhevery:
[presubmit](http://test/OCL:296978067:BASE:296978094:1582584378959:a56cebcd)

kara (approved):
LGTM. Thanks for the descriptive commit message!

shlomiassaf:
@crisbeto My previous concern seems to be true.

While i'm not sure what you've fixed is 100% related, the issue remains, can't create a component instance inside a feature module constructor.

https://github.com/angular/angular/issues/35677

Issue #35041, opened 2020-01-29
Create CI check to prevent additional typescript circular dependencies from being created


josephperrott:
Create a CI check that uses madge to check circular dependencies between Typescript source files. 

We should have:
- [ ] A known list of circular deps to burn down
- [ ] Enforce ensure that we don't add new circular deps
- [ ] Provide a script for users to run, automatically updating the known circular deps list
- [ ] A code ownership group to specifically look at changes to the circular deps list

PR #35647, opened 2020-02-24
Create tool to check for circular dependencies in source files

devversion:
Creates a tool for validating TypeScript circular dependencies. The tool
has been designed in a way that allows us to slowly burn down the amount
of circular dependencies while ensuring that we don't regress.

For more context. See: #35041 and [this comment](https://github.com/angular/angular/pull/34774#issuecomment-577237684).


tools/ts-circular-dependencies/index.ts
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  const expected: Golden = JSON.parse(readFileSync(goldenFile, 'utf8'));
+  const {fixedCircularDeps, newCircularDeps} = compareGoldens(actual, expected);
+  const isMatching = fixedCircularDeps.length === 0 && newCircularDeps.length === 0;
+
+  if (isMatching) {
+    console.info(chalk.green('✅  Golden matches current circular dependencies.'));
+    console.info(
+        chalk.green(`   Current amount of cycles: ${chalk.yellow(cycles.length.toString())}`));
```
josephperrott: Can we change this to be "Current number of cycles".

Additionally, can we always print out the number of cycles in the repo whenever the tool is run.

josephperrott (commented):
Before we approve and commit this, we are also talking about if this should actually live in dev-infra repo and then be depended on via npm.  So we can do the same process for components and tooling.

devversion:
@josephperrott I heard about these plans in the components sync. Concerns about not having another repository for this stuff had been raised. Also personally I'm slightly concerned about going through NPM since it could end up as with `ts-api-guardian` where it's difficult/non-standard process to get a release. It might be good to just use these things through a Bazel `http_repository`.

There are surely benefits/downsides but the primary benefit would be that we don't need any `builds` repositories for `dev-infra` and that we could easily use shared Bazel rules (like `npm_integration`).

Anyway though; TL;DR: Is there anything I need to do? Should I move this already over, or should we just get this started now and move it over once we decided on the process?

packages/circular-dependencies-golden.json
```diff
@@ -0,0 +1,5285 @@
+[
```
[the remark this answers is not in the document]
devversion: Done. Added a small readme. I think the convention should be to have a description for each golden we add there. WDYT?

package.json
```diff
@@ -34,7 +34,10 @@
     "lint": "yarn -s tslint && yarn gulp lint",
     "tslint": "tsc -p tools/tsconfig.json && tslint -c tslint.json \"+(packages|modules|scripts|tools)/**/*.+(js|ts)\"",
     "public-api:check": "node goldens/public-api/manage.js test",
-    "public-api:update": "node goldens/public-api/manage.js accept"
+    "public-api:update": "node goldens/public-api/manage.js accept",
+    "ts-circular-deps": "ts-node dev-infra/ts-circular-dependencies/index.ts",
```
devversion: @josephperrott For now, until your packaging PR lands, I think we should be also fine just using `ts-node` here. It's minimal and allows us to test the changes immediately. We can always change that single line when we have the long-term plan.

dev-infra/BUILD.bazel
```diff
@@ -1,38 +1,38 @@
...
 
 pkg_npm(
     name = "npm_package",
-    srcs = [
-        "package.json",
-    ],
     visibility = ["//visibility:public"],
     deps = [
-        ":bundle",
+        ":cli",
```
josephperrott: Moving to a dependency on the ts_library outputs results in requiring the full umd name which doesn't work as needed.

For instance if you build the `npm_package` current and add it to try to run:
```sh
$ yarn -s ng-dev
/Users/jperrott/js/angular/node_modules/.bin/ng-dev: line 1: syntax error near unexpected token `('
/Users/jperrott/js/angular/node_modules/.bin/ng-dev: line 1: `(function (factory) {'
```

We might have to wait on https://github.com/bazelbuild/rules_typescript/pull/492 for us to be able to do something like this.

josephperrott (changes requested):

package.json
devversion: The generic one is actually used so that the `ts-node <..>` command doesn't need to be repeated all the time.

dev-infra/BUILD.bazel
devversion: I've explained in the commit message what the original problem was. There is nothing wrong with having the right AMD module paths. That's actually how the Angular NPM package output works too. The AMD module names just need to match the module name of the actual NPM package.

Also I added the shebang, so that it can be launched without having to use `node <script>`.

josephperrott (approved):
LGTM

PR #35668, opened 2020-02-25
fix(core): treat `[class]` and `[className]` as unrelated bindings

### LINKED ISSUE #35577: Adding `class` attribute to the component's host element also sets the same value to it's `className` input for some reason
# 🐞 bug report

### Affected Package

Not sure

### Is this a regression?

### Description
For example we have this component:
```ts
import {
  Component,
  Input
} from '@angular/core';

@Component({
  selector: 'app-test',
  template: 'className is "{{ className }}" but should be undefined'
})
export class TestComponent {
  @Input() className: string;
}
```
If we use it like this:
```html
<app-test class="foo"></app-test>
```
`TestComponent` will render `className is "foo" but should be undefined` which is not correct as `className` input was not provided.

## 🔬 Minimal Reproduction

Couldn't reproduce it on Stackblitz (it uses JIT compiler instead of AoT?) so created a separate repo with a minimal reproduction: https://github.com/th0r/ng9-classname-bug

Steps:
1. Checkout repo and run `npm ci && npm start`
2. Open `http://localhost:4200/` in the browser
3. Page renders `className is "foo" but should be undefined`

## 🌍  Your Environment

**Angular Version:**
<pre><code>
Angular CLI: 9.0.3
Node: 12.16.0
OS: darwin x64

Angular: 9.0.2
... router
Ivy Workspace: Yes

@angular-devkit/core              9.0.3
@angular-devkit/schematics        9.0.3
@angular/cli                      9.0.3
@schematics/angular               9.0.3
@schematics/update                0.900.3
</code></pre>

**Anything else relevant?**
Bug doesn't depend on the browser used.

mhevery:
fix(core): treat `[class]` and `[className]` as unrelated bindings

Before this change `[class]` and `[className]` were both converted into `ɵɵclassMap`. The implication of this is that at runtime we could not differentiate between the two and as a result we treated `@Input('class')` and `@Input('className)` as equivalent.

This change makes `[class]` and `[className]` distinct. The implication of this is that `[class]` becomes `ɵɵclassMap` instruction but  `[className]` becomes `ɵɵproperty' instruction. This means that `[className]` will no longer participate in styling and will overwrite the DOM `class` value.

Fix #35577

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [x] Code style update (formatting, local variables)
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/packages/compiler/src/render3/view/styling_builder.ts
+++ b/packages/compiler/src/render3/view/styling_builder.ts
@@ -202,8 +202,7 @@ export class StylingBuilder {
     let binding: BoundStylingEntry|null = null;
     const prefix = name.substring(0, 6);
     const isStyle = name === 'style' || prefix === 'style.' || prefix === 'style!';
-    const isClass = !isStyle &&
-        (name === 'class' || name === 'className' || prefix === 'class.' || prefix === 'class!');
+    const isClass = !isStyle && (name === 'class' || prefix === 'class.' || prefix === 'class!');
     if (isStyle || isClass) {
       const isMapBased = name.charAt(5) !== '.';         // style.prop or class.prop makes this a no
       const property = name.substr(isMapBased ? 5 : 6);  // the dot explains why there's a +1
--- a/packages/core/test/acceptance/styling_spec.ts
+++ b/packages/core/test/acceptance/styling_spec.ts
@@ -3460,13 +3460,12 @@ describe('styling', () => {
           class MyApp {
             // When the first view in the list gets CD-ed, everything works.
             // When the second view gets CD-ed, the styling has already created the data structures
-            // in the `TView`. As a result when
-            // `[class.foo]` runs it already knows that `[class]` is a duplicate and hence it
-            // should check with it. While the resolution is happening it reads the value of the
-            // `[class]`, however `[class]` has not yet executed and therefore it does not have
-            // normalized value in its `LView`. The result is that the assertions fails as it
-            // expects an
-            // `KeyValueArray`.
+            // in the `TView`. As a result when `[class.foo]` runs it already knows that `[class]`
+            // is a duplicate and hence it can overwrite the `[class.foo]` binding. While the
+            // styling resolution is happening the algorithm  reads the value of the `[class]`
+            // (because it overwrites `[class.foo]`), however  `[class]` has not yet executed and
+            // therefore it does not have normalized value in its `LView`. The result is that the
+            // assertions fails as it expects an `KeyValueArray`.
           }
 
           TestBed.configureTestingModule({declarations: [MyApp, MyCmp, HostStylingsDir]});
--- a/packages/core/test/acceptance/styling_spec.ts
+++ b/packages/core/test/acceptance/styling_spec.ts
@@ -1129,7 +1129,7 @@ describe('styling', () => {
   });
 
   onlyInIvy('only ivy combines static and dynamic class-related attr values')
-      .it('should write to a `class` input binding, when static `class` is present', () => {
+      .it('should write combined class attribute and class binding to the class input', () => {
         @Component({
           selector: 'comp',
           template: `{{className}}`,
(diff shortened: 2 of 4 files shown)


packages/core/test/acceptance/styling_spec.ts
```diff
@@ -3411,13 +3385,11 @@ describe('styling', () => {
...
             // When the second view gets CD-ed, the styling has already created the data structures
-            // in the `TView`. As a result when
-            // `[class.foo]` runs it already knows that `[class]` is a duplicate and hence it
-            // should check with it. While the resolution is happening it reads the value of the
-            // `[class]`, however `[class]` has not yet executed and therefore it does not have
-            // normalized value in its `LView`. The result is that the assertions fails as it
-            // expects an
-            // `KeyValueArray`.
+            // in the `TView`. As a result when `[class.foo]` runs it already knows that `[class]`
+            // is a duplicate and hence it should check with it. While the resolution is happening
```
matsko: Please clarify this: "and hence it should check with it"

matsko (approved):

mhevery:
[presubmit](http://test/OCL:297150685:BASE:297150720:1582656063028:afc12938)

packages/core/test/acceptance/styling_spec.ts
mhevery: Good catch, updated for clarity.

mhevery:
Playground showing old behavior: https://stackblitz.com/edit/angular-yvr67j?file=src/app/app.component.ts

packages/compiler/src/render3/view/styling_builder.ts
```diff
@@ -202,8 +202,7 @@ export class StylingBuilder {
     let binding: BoundStylingEntry|null = null;
     const prefix = name.substring(0, 6);
     const isStyle = name === 'style' || prefix === 'style.' || prefix === 'style!';
-    const isClass = !isStyle &&
-        (name === 'class' || name === 'className' || prefix === 'class.' || prefix === 'class!');
+    const isClass = !isStyle && (name === 'class' || prefix === 'class.' || prefix === 'class!');
```
pkozlowski-opensource: question out of curiosity / building up my understanding: - how do we get the `class!` / `style!` prefix?

packages/core/test/acceptance/styling_spec.ts
```diff
@@ -3427,6 +3400,25 @@ describe('styling', () => {
...
+      }
+      @Component({template: `<my-cmp [class]="'bound'"></my-cmp>`})
+      class MyApp {
+      }
+
+      TestBed.configureTestingModule({declarations: [MyApp, MyCmp]});
+      const fixture = TestBed.createComponent(MyApp);
+      fixture.detectChanges();
+      expect(fixture.nativeElement.textContent).toEqual('className = unbound');
+    });
```
pkozlowski-opensource: nit question: would it makes sense, as a sanity check, to add a test that a static `class` is not bound to the `@Input("className")`?

packages/core/test/acceptance/styling_spec.ts
```diff
@@ -1129,17 +1129,17 @@ describe('styling', () => {
   });
 
   onlyInIvy('only ivy combines static and dynamic class-related attr values')
-      .it('should write to a `className` input binding, when static `class` is present', () => {
+      .it('should write to a `class` input binding, when static `class` is present', () => {
```
pkozlowski-opensource: nit: I guess that at this point this test demonstrates that we are writing merged `class="static" [class]="exp"` so maybe the name of this test should reflect this? Sth like "should write combined class attribute and class binding to the class input"?

pkozlowski-opensource (approved):
LGTM (just left a couple of nit comments).

Side note: this is not in the scope of this PR, but more long term, I guess we would still want to merge / throw when both `[class]` and `[className]` are present **_without directives with class / className inputs_**. Obviously it would require that the compiler knows about all directives and their inputs on a given element and this goes back to the locality discussion.

packages/compiler/src/render3/view/styling_builder.ts
mhevery: No idea.

packages/core/test/acceptance/styling_spec.ts
mhevery: fixed

packages/core/test/acceptance/styling_spec.ts
mhevery: done

kara (approved):
LGTM

mhevery:
[presubmit](http://test/OCL:297150685:BASE:297957153:1582934831908:6eadc521)

PR #35669, opened 2020-02-25
Npm integration test manifest

gregmagolan:
Run /integration/ng_elements_schematics test with bazel using the MANIFEST file.

@filipesilva This manifest file should be useful in the CLI repo for your integration tests

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes
- [x] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -78,20 +78,6 @@ executors:
     resource_class: << parameters.resource_class >>
     working_directory: ~/ng
 
-  browsers-executor:
-    parameters:
-      resource_class:
-        type: string
-        default: medium
-    docker:
-      # The browser docker image comes with Chrome and Firefox preinstalled. This is just
-      # needed for jobs that run tests without Bazel. Bazel runs tests with browsers that will be
-      # fetched by the Webtesting rules. Therefore for jobs that run tests with Bazel, we don't need a
-      # docker image with browsers pre-installed.
-      - image: circleci/node:12.14.1-browsers@sha256:792797ab9be3179be7c9fc38a0931a3349288e699467c8d646d7c54e148ae46c
-    resource_class: << parameters.resource_class >>
-    working_directory: ~/ng
-
   windows-executor:
     working_directory: ~/ng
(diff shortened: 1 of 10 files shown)


josephperrott (approved):
LGTM

.circleci/config.yml
```diff
@@ -78,20 +78,6 @@ executors:
     resource_class: << parameters.resource_class >>
     working_directory: ~/ng
 
-  browsers-executor:
```
gregmagolan: Goodbye browsers container

filipesilva (approved):

filipesilva:
Nice! Will have to take a look at the produced format later, but I have a pretty good idea from seeing how its consumed.

gregmagolan:
> Nice! Will have to take a look at the produced format later, but I have a pretty good idea from seeing how its consumed.

Just a JSON file with a map of npm package name to absolute path on disk to the .tar.gz for that package. Archive either comes from a generated npm_package or ng_package rule or a pkg_tar of a root `node_modules/foo` folder so you can share dependencies from your root node_modules with your integration tests automatically so they are all on the same version and so that any post-install steps that can be cached will be.

Puppeteer, for example, doesn't need to download Chrome again for the integrations tests since the root `/node_modules/puppeteer` is passed to the integration tests via .tar.gz. Integration tests also do not need to run `webdriver-manager update` as `**/webdriver-manager` is resolved to the archive of `/node_modules/webdriver-manager`.

gregmagolan:
Hmmm. Thinking out loud. If the filename of the .tar.gz that is used could include a hash of the contents then would we still need the `.yarn-local-cache` work-around? If the contents of the package changed then the `file:/path/to/package-<hash>.tar.gz` reference would as well.

filipesilva:
@gregmagolan nifty stuff, thanks for doing this!

I *think* yarn only checks package and version name when determining if a cached version should be used. I think @gkalpak might have debugged this in depth in the past.

PR #35671, opened 2020-02-25
test(compiler): add correct use case of ngFor in r3 ast

kyliau:
The only test case for `ngFor` exercises an incorrect usage which causes
two bound attributes to be generated . This commit adds a canonical and
correct usage to show the difference between the two.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/compiler/test/render3/r3_ast_spans_spec.ts
+++ b/packages/compiler/test/render3/r3_ast_spans_spec.ts
@@ -225,6 +225,24 @@ describe('R3 AST source spans', () => {
   // TODO(joost): improve spans of nodes extracted from macrosyntax
   describe('inline templates', () => {
     it('is correct for attribute and bound attributes', () => {
+      // Desugared form is
+      // <ng-template ngFor [ngForOf]="items" let-item>
+      //   <div></div>
+      // </ng-template>
+      expectFromHtml('<div *ngFor="let item of items"></div>').toEqual([
+        ['Template', '0:32', '0:32', '32:38'],
+        ['TextAttribute', '5:31', '<empty>'],
+        ['BoundAttribute', '5:31', '<empty>'],
+        ['Variable', '5:31', '<empty>'],
+        ['Element', '0:38', '0:32', '32:38'],
+      ]);
+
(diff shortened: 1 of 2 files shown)


JoostK (approved):

packages/compiler/test/render3/r3_ast_spans_spec.ts
```diff
@@ -225,6 +225,24 @@ describe('R3 AST source spans', () => {
...
+      // </ng-template>
+      expectFromHtml('<div *ngFor="let item of items"></div>').toEqual([
+        ['Template', '0:32', '0:32', '32:38'],
+        ['TextAttribute', '5:31', '<empty>'],
+        ['BoundAttribute', '5:31', '<empty>'],
+        ['Variable', '5:31', '<empty>'],
+        ['Element', '0:38', '0:32', '32:38'],
+      ]);
+
+      // Note that this test exercises an *incorrect* usage of the ngFor
```
ayazhafiz: Can we specify that this is invalid (e.g. does not compile)? We have seen before that incorrect !== incompatible

ayazhafiz (approved):

packages/compiler/test/render3/r3_ast_spans_spec.ts
JoostK: It does parse though. Whether it compiles or not depends on context, which is not provided in this test.

kyliau: reword incorrect -> invalid, with emphasis that this code does not compile

kyliau: hmm.. that's true actually. This test only deals with parsing of the Angular syntax and subsequent compilation depends on context, like the actual implementation of `NgFor`. I removed the comment that it does not compile, and I think _incorrect usage_ is the right term here.

ayazhafiz (approved):

PR #35674, opened 2020-02-25
fix(ivy): type-checking error for duplicate variables in templates

### LINKED ISSUE #35186: Better error handling for template variable redefinitions
# 🚀 Feature request: Better error handling for template variable redefinitions

### Command

- [ ] new
- [x] build
- [x] serve
- [x] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Description
The new Ivy pipeline handles template variable redefinitions differently than in previous Angular versions. It would be great if issues stemming from this user error could be better highlighted at development stages with the Angular Language Service in editors or at compile time with a better error description from the compiler.

#### Incorrect code sample
This issue occurs if you redefine a template variable, like so:
image
Component:
image
Notice that both loop variables are called _i_. Now obviously this is a user error and it is not an intended use of the loop variables.
However neither the Angular Language Service in VS Code or the compiler handle it well.
- The language service:
  - Doesn't highlight the variable redefinition, even as a warning.
  - Does highlight the property accessor being incorrect
  - Thinks that the latter defnition takes precedence over the former, since it thinks _i_ is going to be a number in this scope:
image
    When in fact before Ivy this code did not produce any errors whatsoever, instead at runtime the loop variable held the former value and correctly iterated over the collection in the component and accessed the individual properties. Here is the rendered page for the above template:
image
- The compiler:
  - After Ivy, the above template produces a compiler error. It produces:
    - no error when calling `ng build` in any version or configuration
    - a runtime error when calling `ng serve` with AOT turned off and Ivy turned on in Angular 8
    - a compile time error when calling `ng serve` with  both AOT and Ivy turned on in Angular 8
    - a compile time error when calling `ng serve` in Angular 9
    - a compile time error when calling `ng test` in Angular 9
image
image
This error message does not allow for effective debugging, since it does not mention in which template the problem occured. In fact it does not even state explicitly that it is a template error. The only way I managed to find the component was by modifying the compiler code and writing extra information to the console.

#### Updating to Ivy
Since previous Angular versions did not break on this error we only found it when upgrading a medium/large sized app to Angular 9 and suddenly a completely cryptic compiler error message prevented both testing and compiling. This feature might come in handy for others planning to update to Angular 9.

- The compiler should create a more verbose template error (template parse error?)
- The Angular Language Service should warn about variable redefinitions

No real alternatives to consider.

### Sample
I have attached a sample component and template for testing in a completely new project:
[app.component.ts.txt](https://github.com/angular/angular-cli/files/4164894/app.component.ts.txt)
[app.component.html.txt](https://github.com/angular/angular-cli/files/4164893/app.component.html.txt)

alxhub:
It's an error to declare a variable twice on a specific template:

```html
<div *ngFor="let i of items; let i = index">
</div>
```

This commit introduces a template type-checking error which helps to detect
and diagnose this problem.

Fixes #35186


alxhub:
[Presubmit](http://test/OCL:297209996:BASE:297210100:1582671192754:ce6fe764)

JoostK (approved):

packages/compiler-cli/test/ngtsc/template_typecheck_spec.ts
```diff
@@ -1229,6 +1229,35 @@ export declare class AnimationEvent {
...
+        @NgModule({
+          declarations: [TestCmp],
+          imports: [CommonModule],
+        })
+        export class Module {}
+      `);
+
+      const diags = env.driveDiagnostics();
+      expect(diags.length).toEqual(1);
+      expect(getSourceCodeForDiagnostic(diags[0])).toContain('let i of items;');
```
petebacondarwin: Worth checking the error code here too?

petebacondarwin (approved):
Nice work @alxhub - it is very satisfying when the architecture of the code allows a change to be effected with so little churn!
A couple of suggestions.

IgorMinar (commented):
lgtm - post submit. let's please not submit unreviewed PRs in the future.

PR #35675, opened 2020-02-25
fix(compiler): fine-grained source span for microsyntax template

kyliau:
Currently, when an inline template is parsed, every symbol in the
    template assumes the source span of the entire attribute:
    
    ```
      <div *ngFor="let item of items">
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
           entire source span of the attribute
    ```
    
    This is problematic for many reasons:
    
    1. Unable to pinpoint a specific location if an error occurs
    2. Unable to traverse the complete AST for a particular position
    3. Microsyntax keywords (let & as) intermingled with directive selector
       like "of", which is not a keyword
    
    This commit attempts to rectify this issue by making the source span for
    each symbol in an inline template as fine grained as possible.
    
    As a result of this change, the new source spans are as follows:
    
    ```
      <div *ngFor="let item of items">
            ^^^^^      ^^^^ ^^ ^^^^^
    ```
    
    1. `ngFor` refers to the attribute
    2. `item` refers to template variable
    3. `of` refers to `[ngForOf]`
    4. `items` refers to the bound target of `ngForOf`
    
    This structure allows the source spans to better represent the desugared
    form.
    
    Other minor changes include:
    
    1. Trailing whitespaces no longer part of source span
    2. `let` and `as` no longer part of source span
    3. VariableAst now contains the source span of the value it refers to

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/packages/compiler/src/expression_parser/ast.ts
+++ b/packages/compiler/src/expression_parser/ast.ts
@@ -285,7 +285,7 @@ export class ASTWithSource extends AST {
 
 export class TemplateBinding {
   constructor(
-      public span: ParseSpan, sourceSpan: AbsoluteSourceSpan, public key: string,
+      public span: ParseSpan, public sourceSpan: AbsoluteSourceSpan, public key: string,
       public keyIsVar: boolean, public name: string, public expression: ASTWithSource|null) {}
 }
 
@@ -783,7 +783,18 @@ export class ParsedEvent {
 }
 
 export class ParsedVariable {
-  constructor(public name: string, public value: string, public sourceSpan: ParseSourceSpan) {}
+  /**
+   * Create a new instance of a parsed variable declaration. A variable could be
+   * with or without a value, i.e. "let x" or "let x = y"
+   * @param name the variable name, “x“ in the example above
+   * @param value the variable value, “y” in the example above. If there is no
+   * RHS in the variable declaration, value must be "$implicit".
+   * @param sourceSpan source span of "x"
+   * @param valueSpan source span of "y", optional
+   */
+  constructor(
+      public name: string, public value: string, public sourceSpan: ParseSourceSpan,
+      public valueSpan?: ParseSourceSpan) {}
 }
 
 export const enum BindingType {
--- a/packages/compiler/src/expression_parser/parser.ts
+++ b/packages/compiler/src/expression_parser/parser.ts
@@ -118,12 +118,31 @@ export class Parser {
         span, span.toAbsolute(absoluteOffset), prefix, uninterpretedExpression, location);
   }
 
-  parseTemplateBindings(tplKey: string, tplValue: string, location: any, absoluteOffset: number):
-      TemplateBindingParseResult {
+  /**
+   * Parse microsyntax template bindings and return errors, if any.
+   *
+   * Given the following example:
+   *   <div *ngFor="let item of items">
+   *         ^      ^ absoluteValueOffset for `tplValue`
+   *         absoluteKeyOffset for `tplKey`
+   *
+   * The `absoluteKeyOffset` is needed so that the correct source span for the
+   * attribute itself could be generated.
+   *
+   * @param tplKey name of directive, without the * prefix. For example: ngIf, ngFor
+   * @param tplValue string value for the inline template
+   * @param location
+   * @param absoluteKeyOffset absolute offset of the `tplKey`.
+   * @param absoluteValueOffset absolute offset of the `tplValue`
+   */
+  parseTemplateBindings(
+      tplKey: string, tplValue: string, location: any, absoluteKeyOffset: number,
+      absoluteValueOffset: number): TemplateBindingParseResult {
     const tokens = this._lexer.tokenize(tplValue);
     return new _ParseAST(
-               tplValue, location, absoluteOffset, tokens, tplValue.length, false, this.errors, 0)
-        .parseTemplateBindings(tplKey);
+               tplValue, location, absoluteValueOffset, tokens, tplValue.length, false, this.errors,
+               0)
+        .parseTemplateBindings(tplKey, absoluteKeyOffset);
   }
 
   parseInterpolation(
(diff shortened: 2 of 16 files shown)


kyliau:
reviewers: I had to disable one test and modify two others in language service, but fixing them in this PR would make it too big. Will submit subsequent PRs once this is merged.

packages/compiler/src/expression_parser/parser.ts
```diff
@@ -737,56 +747,93 @@ export class _ParseAST {
...
+        start = this.inputIndex;
         name = rawKey;
         key = this.expectTemplateBindingKey();  // read local var name
         isVar = true;
       } else if (this.next !== EOF && !this.peekKeywordLet()) {
         const start = this.inputIndex;
         const ast = this.parsePipe();
-        const source = this.input.substring(start - this.offset, this.inputIndex - this.offset);
+        // Getting the end of the last token removes trailing whitespace
+        const source = this.input.substring(start - this.offset, this.peek(-1).end - this.offset);
```
ayazhafiz: I have longer term solution that will do this for all expressions in draft PR... interestingly its only blockers are template bindings!

ayazhafiz: Can you add a TODO here to remove with #34690?

packages/compiler/src/template_parser/binding_parser.ts
```diff
@@ -479,6 +488,35 @@ export class BindingParser {
   }
 }
 
+/**
+ * Compute a new ParseSourceSpan for an inline template binding based off an
+ * original `sourceSpan` by computing a relative offset using the specified
+ * `absoluteSpan`. This should be used only for inline templates because the
+ * implementation assumes that the offset occurs on the same line.
+ * Inline templates are, by definition, on the same line.
```
ayazhafiz: We also use inline templates to describe HTML templates declared on a component decorator. So maybe we can provide a less ambiguous name for this? Like inline expression or similar. Also, are we sure that template bindings can’t be stretched across multiple lines?

packages/language-service/src/utils.ts
```diff
@@ -111,6 +111,11 @@ export function findTemplateAstAt(ast: TemplateAst[], position: number): Templat
         if (!len || isNarrower(span, spanOf(path[len - 1]))) {
           path.push(ast);
         }
+      } else if (
+          ast instanceof BoundDirectivePropertyAst && inSpan(position, ast.value.sourceSpan)) {
+        // The BoundDirectivePropertyAst itself refers to "of" (ngForOf) in
+        // the source text but its value refers to the symbol after "of".
+        path.push(ast);
```
ayazhafiz: TODO: test this util function

ayazhafiz (changes requested):

kyliau (commented):
I cleaned up some changes in the "meat" of the PR in `parseTemplateBindings()` in `parser.ts` and added more comments. 
All Ayaz's comments have been addressed.

packages/compiler/src/expression_parser/ast.ts
```diff
@@ -783,7 +783,18 @@ export class ParsedEvent {
...
+   * with or without a value, i.e. "let x" or "let x = y"
+   * @param name the variable name, “x“ in the example above
+   * @param value the variable value, “y” in the example above. If there is no
+   * RHS in the variable declaration, value must be "$implicit".
+   * @param sourceSpan source span of "x"
+   * @param valueSpan source span of "y", optional
+   */
+  constructor(
+      public name: string, public value: string, public sourceSpan: ParseSourceSpan,
+      public valueSpan?: ParseSourceSpan) {}
```
AndrewKushnir: Quick question: the `sourceSpan` now only contains the var itself and `valueSpan` - the value (if present). If I understand correctly, the `sourceSpan` used to have a meaning of the "span for entire expression". For example, it looks like the case for `TemplateBinding` class above. If my understanding is correct, I'd propose to keep things aligned, where `sourceSpan` is the entire span and introduce extra spans (as needed) for pieces of the expression (like `keySpan`, `valueSpan`). WDYT?

packages/compiler/src/expression_parser/parser.ts
```diff
@@ -118,12 +118,31 @@ export class Parser {
...
+   *
+   * The `absoluteKeyOffset` is needed so that the correct source span for the
+   * attribute itself could be generated.
+   *
+   * @param tplKey name of directive, without the * prefix. For example: ngIf, ngFor
+   * @param tplValue string value for the inline template
+   * @param location
+   * @param absoluteValueOffset absolute offset of the `tplValue`
+   * @param absoluteKeyOffset absolute offset of the `tplKey`.
+   */
```
AndrewKushnir: Thanks for all the comments Keen, great job 👍

packages/compiler/src/expression_parser/parser.ts
```diff
@@ -118,12 +118,31 @@ export class Parser {
...
+  parseTemplateBindings(
+      tplKey: string, tplValue: string, location: any, absoluteValueOffset: number,
+      absoluteKeyOffset: number): TemplateBindingParseResult {
```
AndrewKushnir: really minor nit: I'd propose to put `absoluteKeyOffset` first and `absoluteValueOffset` second (to respect the order in which they appear in expression).

packages/compiler/src/expression_parser/parser.ts
```diff
@@ -737,56 +756,86 @@ export class _ParseAST {
...
+          const token = this.peek(-1);             // token for 'b'
+          const parseSpan = new ParseSpan(token.index, token.end);
+          // The actual AST here doesn't matter much because the RHS of an "="
+          // expression can only be an identifier or string. No expression allowed.
+          // This AST is generated solely to capture the source span of the RHS.
+          const ast = new AST(parseSpan, parseSpan.toAbsolute(this.absoluteOffset));
+          expression =
+              new ASTWithSource(ast, name, this.location, ast.sourceSpan.start, this.errors);
         } else {
-          name = '\$implicit';
```
AndrewKushnir: Just curios: why backslash is not needed? :)

packages/compiler/src/template_parser/binding_parser.ts
```diff
@@ -129,20 +129,27 @@ export class BindingParser {
       targetMatchableAttrs: string[][], targetProps: ParsedProperty[],
       targetVars: ParsedVariable[]) {
     const bindings = this._parseTemplateBindings(tplKey, tplValue, sourceSpan, absoluteValueOffset);
-
     for (let i = 0; i < bindings.length; i++) {
       const binding = bindings[i];
+      // Generate a new ParseSourceSpan for the binding based off the original sourceSpan
+      const bindingSourceSpan = offsetParseSourceSpanForMicrosyntax(sourceSpan, binding.sourceSpan);
+      // If the binding has a target (value), generate a ParseSourceSpan for the target as well
+      const valueSpan = binding.expression ?
```
AndrewKushnir: It looks like `valueSpan` is used in 2 out of 3 cases below. Would it make sense to move calculation to respective branches (1 and 2)? In the second branch the `binding.expression` check would not be required (as it's a condition to enter the branch).

packages/compiler/src/template_parser/binding_parser.ts
```diff
@@ -479,6 +487,35 @@ export class BindingParser {
   }
 }
 
+/**
+ * Compute a new ParseSourceSpan for a microsyntax inline template binding
+ * based off an original `sourceSpan` by computing a relative offset using the
+ * specified `absoluteSpan`. This should be used only for inline templates
+ * because the implementation assumes that the offset occurs on the same line.
+ * Microsyntax inline templates are, by definition, on the same line.
+ * @param sourceSpan original source span of the attribute
```
AndrewKushnir: May be add a new line between function description and first `@param`?

AndrewKushnir (changes requested):
Great job @kyliau 👍 Also thanks for additional comments/documentation, it's very helpful.

I left a few comments, plz let me know if you have questions. Thank you.

packages/compiler/src/expression_parser/parser.ts
kyliau: done.

packages/compiler/src/expression_parser/parser.ts
kyliau: backslash is used to escape the character, but in javascript string '$' has no special meaning so whether it is escaped or not does not make any difference.

packages/compiler/src/template_parser/binding_parser.ts
kyliau: done

packages/compiler/test/render3/r3_ast_spans_spec.ts
```diff
@@ -245,26 +245,26 @@ describe('R3 AST source spans', () => {
...
       ]);
     });
 
     it('is correct for variables via as ...', () => {
       expectFromHtml('<div *ngIf="expr as local"></div>').toEqual([
         ['Template', '0:27', '0:27', '27:33'],
-        ['BoundAttribute', '5:26', '<empty>'],
-        ['Variable', '5:26', '<empty>'],
+        ['BoundAttribute', '6:10', '12:16'],  // ngIf -> expr
+        ['Variable', '20:25', '<empty>'],     // local
```
ivanwonder:
```
['Variable', '20:25', '6:10'],     // local -> ngIf
```
https://angular.io/guide/structural-directives#microsyntax-examples
I dont know if the position `6:10` is suitable here. But I think the `local` should not be empty. It points to `ngIf`.🤔

kyliau: yes this is a bug :) I'll fix this in the following fixup!

PR #35683, opened 2020-02-26
fix(ngcc): allow deep-import warnings to be ignored

### LINKED ISSUE #35615: Entry point 'xxx' contains deep imports into 'yyy'
# 🐞 bug report

### Is this a regression?

### Description
Multiple warnings like "Entry point 'devextreme-angular/...' contains deep imports into '/node_modules/devextreme/...'. This is probably not a problem, but may cause the compilation of entry points to be out of order." occur when building an Angular v9 project with Ivy enabled.

## 🔬 Minimal Reproduction
[https://github.com/onebalaban/dxangular9](https://github.com/onebalaban/dxangular9)
npm install
ng build

## 🌍  Your Environment
<pre><code>
Angular CLI: 9.0.3
Node: 13.9.0
OS: win32 x64

Angular: 9.0.2
... router
Ivy Workspace: Yes

@angular-devkit/core              7.3.9
@angular-devkit/schematics        7.3.9
@angular/cli                      9.0.3
@schematics/angular               8.3.20
@schematics/update                0.900.3
</code></pre>

**Anything else relevant?**
There are two packages in the project. The first one is [devextreme-angular](https://github.com/DevExpress/devextreme-angular). It is created according to Angular library requirements. The second package is **devextreme**. It is not an Angular library. The thing is that warnings are related to long paths of this second package. We think these warnings are false-positive.

petebacondarwin:
This commit adds a new ngcc configuration, `ignorableDeepImportMatchers`
for packages. This is a list of regular expressions matching deep imports
that can be safely ignored from that package. Deep imports that are not
ignored cause a warning to be logged.

// FW-1892

Fixes #35615

### DIFF
--- a/packages/compiler-cli/ngcc/src/packages/configuration.ts
+++ b/packages/compiler-cli/ngcc/src/packages/configuration.ts
@@ -13,7 +13,12 @@ import {PackageJsonFormatPropertiesMap} from './entry_point';
 /**
  * The format of a project level configuration file.
  */
-export interface NgccProjectConfig<T = NgccPackageConfig> { packages: {[packagePath: string]: T}; }
+export interface NgccProjectConfig<T = NgccPackageConfig> {
+  /**
+   * The packages that are configured by this project config.
+   */
+  packages: {[packagePath: string]: T};
+}
 
 /**
  * The format of a package level configuration file.
@@ -27,6 +32,11 @@ export interface NgccPackageConfig {
    * will be absolute.
    */
   entryPoints: {[entryPointPath: string]: NgccEntryPointConfig;};
+  /**
+   * A collection of regexes that match deep imports to ignore, for this package, rather than
+   * displaying a warning.
+   */
+  ignorableDeepImportMatchers?: RegExp[];
 }
 
 /**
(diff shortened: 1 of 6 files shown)


petebacondarwin:
Here is a demo of the kind of ngcc.config.js that can now be used to fix these unwanted warnings: https://github.com/petebacondarwin/angular-issue-35615/commit/acdc35733362a2c5e01477b8874753b245c0c4ac

packages/compiler-cli/ngcc/src/packages/configuration.ts
```diff
@@ -27,6 +32,11 @@ export interface NgccPackageConfig {
    * will be absolute.
    */
   entryPoints: {[entryPointPath: string]: NgccEntryPointConfig;};
+  /**
+   * A collection of regexes that match deep imports to ignore, for this package, rather than
```
alxhub:
```suggestion
   * A collection of regexes that match deep imports to ignore, for this project, rather than
```
This seems to be a project level configuration, not a package one?

alxhub (approved):

packages/compiler-cli/ngcc/src/packages/configuration.ts
petebacondarwin: Actually it is at the package level. You define the regexes for each package that will do the importing. This way we can isolate turning off warnings only to specific packages.

JoostK (approved):

gkalpak (approved):

stevethemacguy:
@petebacondarwin How do we know whether it's safe to ignore the warning or if we should actually try to fix it? Thank you!

petebacondarwin:
This warning can happily be ignored if the package is not compiled by Angular (i.e. has not `...metadata.json` files) and does not depend upon packages that are compiled by Angular.
Furthermore, even if the package is compiled by Angular, ngcc will still process it but there is a chance that the order in which packages are processed might be incorrect. In this case it would most likely lead to a processing error in ngcc, a compile-time error in `ng build` (and similar) or a runtime error. If you are getting known of these, and you can see that the order in which packages are being processed is correct then it is also safe to ignore these warnings.

PR #35688, opened 2020-02-26
test(language-service): differentiate feature and internal infra tests

ayazhafiz:
This commit differentiates language service feature and language service
infrastructure tests. This is primarily to make testing of different
components at the development level easier. This commit continues a
small effort to expand our test coverage and normalize testing
structure.

Also adds test coverage to language service tests. We have quite a bit
to go on that front :slightly_smiling_face:.

## PR Checklist

## PR Type

- [x] Build related changes

- [ ] Yes
- [x] No

### DIFF
--- a/packages/language-service/test/BUILD.bazel
+++ b/packages/language-service/test/BUILD.bazel
@@ -8,24 +8,11 @@ circular_dependency_test(
 )
 
 ts_library(
-    name = "test_lib",
+    name = "test_utils_lib",
     testonly = True,
     srcs = [
-        "completions_spec.ts",
-        "definitions_spec.ts",
-        "diagnostics_spec.ts",
-        "global_symbols_spec.ts",
-        "hover_spec.ts",
-        "html_info_spec.ts",
-        "language_service_spec.ts",
-        "reflector_host_spec.ts",
-        "template_spec.ts",
         "test_utils.ts",
-        "ts_plugin_spec.ts",
-        "typescript_host_spec.ts",
-        "utils_spec.ts",
     ],
-    data = glob(["project/**/*"]),
     deps = [
         "//packages/compiler",
         "//packages/compiler-cli/test:test_utils",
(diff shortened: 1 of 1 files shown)


packages/language-service/test/BUILD.bazel
```diff
@@ -8,34 +8,95 @@ circular_dependency_test(
...
+        ":test_utils_lib",
         "//packages/compiler",
-        "//packages/compiler-cli/test:test_utils",
         "//packages/language-service",
         "@npm//typescript",
     ],
 )
 
+jasmine_node_test(
+    name = "features_test",
```
kyliau: could you please rename this so that it becomes the default test target of the package?
i.e. one could just run `bazel test //packages/language-service/test`

packages/language-service/test/BUILD.bazel
```diff
@@ -8,34 +8,95 @@ circular_dependency_test(
...
+        "no-ivy-aot",
+    ],
+    deps = [
+        ":infra_test_lib",
+    ],
+)
+
 jasmine_node_test(
     name = "test",
+    coverage = True,
```
kyliau: do we have coverage infrastructure in the open source setup?

packages/language-service/test/BUILD.bazel
```diff
@@ -8,34 +8,95 @@ circular_dependency_test(
...
         "//packages/language-service",
         "@npm//typescript",
     ],
 )
 
+jasmine_node_test(
+    name = "features_test",
+    coverage = True,
+    data = [
+        "//packages/common:npm_package",
```
kyliau: I don't think we have to depend on the `npm package`, they requires minification etc and take a much longer time to build. It's a good chance now to clean up the build dependencies.

kyliau (changes requested):

packages/language-service/test/BUILD.bazel
devversion: I was going to comment on that too. I don't think we decided on how we want to use the coverage on framework. It comes with a performance overhead of 20% ([according to the commit that introduced this feature](https://github.com/bazelbuild/rules_nodejs/commit/9abc78ffb596f1a78d0671cd51ec9b398dbc3f67)). Maybe it should be manually enabled if coverage is desired (for now)?

packages/language-service/test/BUILD.bazel
ayazhafiz: Do we want this to be the default target? There is a test that executes all specs which I feel should be the default.

packages/language-service/test/BUILD.bazel
ayazhafiz: Fair. Will elide for now.

packages/language-service/test/BUILD.bazel
ayazhafiz: Will fix.

ayazhafiz: Actually, we do need the npm package since that is what we use to load dependencies for the test project...

kyliau (approved):

ayazhafiz:
@devversion do you mind taking a look again?

devversion (approved):
LGTM

PR #35691, opened 2020-02-26
build: allow auto-discover all typings files in npm package by ts-api-guardian

josephperrott:
Adds a new feature to ts-api-guardian allowing for automatically discovering all
entry point d.ts files from all package.json files in a provided directory.

### DIFF
--- a/tools/ts-api-guardian/lib/cli.ts
+++ b/tools/ts-api-guardian/lib/cli.ts
@@ -14,7 +14,7 @@ const chalk = require('chalk');
 import * as minimist from 'minimist';
 import * as path from 'path';
 
-import {SerializationOptions, generateGoldenFile, verifyAgainstGoldenFile} from './main';
+import {SerializationOptions, generateGoldenFile, verifyAgainstGoldenFile, discoverAllEntrypoints} from './main';
 
 const CMD = 'ts-api-guardian';
 
@@ -44,6 +44,12 @@ export function startCli() {
       banned: ['experimental', 'publicApi', 'codeGenApi'],
       toCopy: ['deprecated']
     };
+  }
+
+  // In autoDiscoverEntrypoints mode we set the inputed files as the discovered entrypoints
+  // for the rootDir
+  if (argv['autoDiscoverEntrypoints']) {
+    argv._ = Array.from(discoverAllEntrypoints(argv['rootDir']));
   }
 
   for (const error of errors) {
@@ -110,7 +116,7 @@ export function parseArguments(input: string[]):
       'allowModuleIdentifiers'
     ],
     boolean: [
-      'help', 'useAngularTagRules',
+      'help', 'useAngularTagRules', 'autoDiscoverEntrypoints',
       // Options used by chalk automagically
       'color', 'no-color'
     ],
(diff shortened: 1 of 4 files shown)


tools/ts-api-guardian/lib/main.ts
```diff
@@ -16,14 +16,18 @@ export {SerializationOptions, publicApi} from './serializer';
...
+  }
   ensureDirectory(path.dirname(outFile));
   fs.writeFileSync(outFile, output);
 }
 
 export function verifyAgainstGoldenFile(
     entrypoint: string, goldenFile: string, options: SerializationOptions = {}): string {
   const actual = publicApi(entrypoint, options);
-  const expected = fs.readFileSync(goldenFile).toString();
+  const expected = fs.existsSync(goldenFile) ? fs.readFileSync(goldenFile).toString() : '';
```
devversion: This was an issue quite some time. Nice improvement 👍

tools/ts-api-guardian/lib/cli.ts
```diff
@@ -147,15 +153,26 @@ export function parseArguments(input: string[]):
...
-  } else if (argv._.length > 1 && !argv['outDir'] && !argv['verifyDir']) {
-    errors.push(`More than one input specified. Use --${modes[0]}Dir instead.`);
-    modes = ['help'];
+  if (argv['autoDiscoverEntrypoints']) {
+    if (!argv['rootDir']) {
+      errors.push(`rootDir must be provided with autoDiscoverEntrypoints enabled.`);
+      modes = ['help'];
+    }
+    if (!argv['outDir'] && !argv['verifyDir']) {
+      errors.push(`outDir or verifyDir must be used with autoDiscoverEntrypoints.`);
```
devversion: Prefix the options with `--` so that it's clear that the options were meant?

tools/ts-api-guardian/lib/cli.ts
```diff
@@ -46,6 +46,12 @@ export function startCli() {
     };
   }
 
+  // In autoDiscoverEntrypoints mode we set the inputed files as the discovered entrypoints
+  // for the rootDir
+  if (argv['autoDiscoverEntrypoints']) {
+    argv._ = Array.from(discoverAllEntrypoints(argv['rootDir']));
+  }
+
```
devversion: Could we make this more robust by just having a new variable here (e.g. `targetFiles`). This one is then set appropriately and passed to `resolveFileNamePairs` below (instead of passing the argv object around)

tools/ts-api-guardian/lib/main.ts
```diff
@@ -16,14 +16,18 @@ export {SerializationOptions, publicApi} from './serializer';
 export function generateGoldenFile(
     entrypoint: string, outFile: string, options: SerializationOptions = {}): void {
   const output = publicApi(entrypoint, options);
+
+  if (process.env['BUILD_WORKSPACE_DIRECTORY']) {
+    outFile = path.join(process.env['BUILD_WORKSPACE_DIRECTORY'], outFile);
+  }
```
devversion: Is this is a symptom from an issue where the golden file does not yet exist, so `require.resolve` (runfile resolution) won't work? I think we should respect this environment variable in the `resolveBazelFilePath` variable if the runfile resolution fails. What do you think?

devversion:
So, that means that the API golden tests always need to happen for `npm_package` targets now?  This could slow down our API golden tests. Alternatively we could detect entry-points in the non-`npm_package` targets by looking for `public-api.ts` files. I'm fine either way though.

tools/ts-api-guardian/lib/main.ts
```diff
@@ -43,3 +47,46 @@ function ensureDirectory(dir: string) {
...
+  } catch {
+    return false;
+  }
+}
+
+/**
+ * Gets an interable of paths to the typings files for each of the recursively discovered
+ * package.json
+ * files from the directory provided.
+ */
```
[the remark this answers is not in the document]
josephperrott: Now that I look at it, its an artifact from trying to provide the entrypoints directly to `resolveFileNamePairs`

tools/ts-api-guardian/lib/main.ts
josephperrott: `BUILD_WORKSPACE_DIRECTORY` will only be available during `bazel run`, so it only really comes into play for us when we are attempting to generate and write the file.

During a `bazel test` this environment variable is not available, so we would have to fallback anyway in this case inside of the `resolveBazelFilePath` function.

josephperrott:
@devversion Yes, our API goldens would depend on the `npm_package` targets.

Relying on the `npm_package` target allows us to actually test the public API we are shipping to our users, rather than a facsimile of what we expect it will be. We actually have already discovered issues we were catching with our current set up (see #35690)

For the case of CI, we already have to build these targets for our tests, so I don't feel that it will make a difference in cost, because it will just reorder and run the api tests later.

devversion:
@josephperrott Sounds good. If we make sure that we don't build the packages multiple times unnecessarily, then this is definitely better. Agreed.

tools/ts-api-guardian/lib/main.ts
devversion: That makes sense. Optional: Add a comment for this here, and maybe also add a comment in `resolveBazelFilePath` that the try/catch is needed for resolution of files within tree artifacts?

You might able to just make this a noop-catch and `return fileName;` should do the same. I think we needed `path.relative` only for the absolute resolved path from the manifest.

josephperrott: Added in the comments.

devversion: Thanks for adding these! Could we make those more concrete/helpful by saying that:

In `resolveBazelFilePath`: We need the `try/catch` to handle file resolution for tree artifacts where individual files inside the tree artifact (e.g. `npm_package`) are not declared outputs and cannot be resolved through the manifest?

The BUILD_WORKSPACE_DIRECTORY variable is needed only for approving the golden in case the golden is not existing and the golden file is not symlinked/part of the manifest. Right?

devversion (approved):
LGTM. Replied on the comments for file path resolution, but that shouldn't be blocking.

PR #35695, opened 2020-02-26
feat(compiler): add dependency info and ng-content selectors to metadata

JoostK:
This commit augments the `FactoryDef` declaration of Angular decorated 
classes to contain information about the parameter decorators used in 
the constructor. If no constructor is present, or none of the parameters
have any Angular decorators, then this will be represented using the 
`null` type. Otherwise, a tuple type is used where the entry at index `i`
corresponds with parameter `i`. Each tuple entry can be one of two types:

1. If the associated parameter does not have any Angular decorators,
   the tuple entry will be the `null` type.
2. Otherwise, a type literal is used that may declare at least one of 
   the following properties:
   - "attribute": if `@Attribute` is present. The injected attribute's 
   name is used as string literal type, or the `unknown` type if the
   attribute name is not a string literal. 
   - "self": if `@Self` is present, always of type `true`.
   - "skipSelf": if `@SkipSelf` is present, always of type `true`.
   - "host": if `@Host` is present, always of type `true`.
   - "optional": if `@Optional` is present, always of type `true`.
   
   A property is only present if the corresponding decorator is used.
   
   Note that the `@Inject` decorator is currently not included, as it's
   non-trivial to properly convert the token's value expression to a 
   type that is valid in a declaration file. 

Additionally, the `ComponentDefWithMeta` declaration that is created for
Angular components has been extended to include all selectors on 
`ng-content` elements within the component's template.

This additional metadata is useful for tooling such as the Angular
Language Service, as it provides the ability to offer suggestions for
directives/components defined in libraries. At the moment, such
tooling extracts the necessary information from the _metadata.json_
manifest file as generated by ngc, however this metadata representation
is being replaced by the information emitted into the declaration files.

Resolves FW-1870

### DIFF
--- a/packages/compiler/src/render3/r3_factory.ts
+++ b/packages/compiler/src/render3/r3_factory.ts
@@ -142,6 +142,13 @@ export interface R3DependencyMetadata {
   token: o.Expression;
 
   /**
+   * If an @Attribute decorator is present, this is the literal type of the attribute name, or
+   * the unknown type if no literal type is available (e.g. the attribute name is an expression).
+   * Will be null otherwise.
+   */
+  attribute: o.Expression|null;
+
+  /**
    * An enum indicating whether this dependency has special meaning to Angular and needs to be
    * injected specially.
    */
@@ -180,6 +187,7 @@ export interface R3FactoryFn {
 export function compileFactoryFunction(meta: R3FactoryMetadata): R3FactoryFn {
   const t = o.variable('t');
   const statements: o.Statement[] = [];
+  let ctorDepsType: o.Type = o.NONE_TYPE;
 
   // The type to instantiate via constructor invocation. If there is no delegated factory, meaning
   // this type is always created by constructor invocation, then this is the type-to-create
@@ -197,6 +205,8 @@ export function compileFactoryFunction(meta: R3FactoryMetadata): R3FactoryFn {
       ctorExpr = new o.InstantiateExpr(
           typeForCtor,
           injectDependencies(meta.deps, meta.injectFn, meta.target === R3FactoryTarget.Pipe));
+
+      ctorDepsType = createCtorDepsType(meta.deps);
     }
   } else {
     const baseFactory = o.variable(`ɵ${meta.name}_BaseFactory`);
@@ -269,8 +279,9 @@ export function compileFactoryFunction(meta: R3FactoryMetadata): R3FactoryFn {
         [new o.FnParam('t', o.DYNAMIC_TYPE)], body, o.INFERRED_TYPE, undefined,
         `${meta.name}_Factory`),
     statements,
-    type: o.expressionType(
-        o.importExpr(R3.FactoryDef, [typeWithParameters(meta.type.type, meta.typeArgumentCount)]))
+    type: o.expressionType(o.importExpr(
+        R3.FactoryDef,
+        [typeWithParameters(meta.type.type, meta.typeArgumentCount), ctorDepsType]))
   };
 }
 
@@ -319,6 +330,49 @@ function compileInjectDependency(
   }
 }
 
+function createCtorDepsType(deps: R3DependencyMetadata[]): o.Type {
+  let hasTypes = false;
+  const attributeTypes = deps.map(dep => {
+    const type = createCtorDepType(dep);
+    if (type !== null) {
+      hasTypes = true;
+      return type;
+    } else {
+      return o.literal(null);
+    }
+  });
+
+  if (hasTypes) {
+    return o.expressionType(o.literalArr(attributeTypes));
+  } else {
+    return o.NONE_TYPE;
+  }
+}
+
+function createCtorDepType(dep: R3DependencyMetadata): o.LiteralMapExpr|null {
+  const entries: {key: string, quoted: boolean, value: o.Expression}[] = [];
+
+  if (dep.resolved === R3ResolvedDependencyType.Attribute) {
+    if (dep.attribute !== null) {
+      entries.push({key: 'attribute', value: dep.attribute, quoted: false});
+    }
+  }
+  if (dep.optional) {
+    entries.push({key: 'optional', value: o.literal(true), quoted: false});
+  }
+  if (dep.host) {
+    entries.push({key: 'host', value: o.literal(true), quoted: false});
+  }
+  if (dep.self) {
+    entries.push({key: 'self', value: o.literal(true), quoted: false});
+  }
+  if (dep.skipSelf) {
+    entries.push({key: 'skipSelf', value: o.literal(true), quoted: false});
+  }
+
+  return entries.length > 0 ? o.literalMap(entries) : null;
+}
+
 /**
  * A helper function useful for extracting `R3DependencyMetadata` from a Render2
  * `CompileTypeMetadata` instance.
@@ -348,7 +402,7 @@ export function dependenciesFromGlobalMetadata(
       // Construct the dependency.
       deps.push({
         token,
-        resolved,
+        attribute: null, resolved,
         host: !!dependency.isHost,
         optional: !!dependency.isOptional,
         self: !!dependency.isSelf,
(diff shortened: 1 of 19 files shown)


piotrtomiak:
@JoostK would it be possible to include types for `@Attribute` arguments? Or is it too complicated?

JoostK:
@piotrtomiak AFAIK, `@Attribute` injections can only be of type string (or `null`) as it only works for static attributes, not bindings.

piotrtomiak:
@JoostK - ultimately the underlying object type would be `string` indeed, but you can have a more specific `string` type, like:
```
import { Component, Attribute } from '@angular/core';

export type ButtonType = 'primary' | 'secondary';

@Component({
  selector: 'app-button',
  template: `<button [ngClass]="type"><ng-content></ng-content></button>`
})
export class ButtonComponent {
  constructor(@Attribute('type') public type: ButtonType = 'primary') { }
}
```
We will not be able to perform additional validation of the input, or more precise code completion, if we don't know the exact type of attribute. It would be great, if for such a component definition we could get:
```
import * as i0 from "@angular/core";
export declare type ButtonType = 'primary' | 'secondary';
export declare class ButtonComponent {
    type: ButtonType;
    constructor(type?: ButtonType);
    static ɵfac: i0.ɵɵFactoryDef<ButtonComponent>;
    static ɵcmp: i0.ɵɵComponentDefWithMeta<ButtonComponent, "app-button", never, {}, {}, never, {'type': ButtonType}, never>;
}
```

JoostK:
@piotrtomiak Ah, that is an excellent observation! I hadn't thought about union types. Also, you made me realize that this is currently a blind spot in the template type checker as well.

Anyway, I think it is indeed possible to implement your proposed syntax of using an object literal type, as referred types should indeed be simply copyable.

piotrtomiak:
@JoostK Awesome, thanks!

JoostK:
@piotrtomiak I've discussed the format with @alxhub and we decided to make some changes. Initially there was a concern that user types may need to be imported for inherited constructors, but this turned out not to be the case as any `@Attribute` in a baseclass' constructor would not be part of the subclass' definition. Anyhow, this lead to an alternative encoding which would be a tuple type of strings and nulls corresponding with the indices of the constructor parameter, e.g.:

```ts
constructor(a: Foo, @Attribute("type") type: string, b: Bar) {}
```

would be encoding using `[null, "type", null]` so that the type information can be extracted from the constructor declaration. This avoids us having to include user types that are available elsewhere in the class declaration.

Note, also, the interesting scenario of having a baseclass with an `@Attribute` decorator. If the subclass has a constructor, it will _not_ inherit the `@Attribute` but it either has to have an `@Attribute` decorator of its own, or pass in some string value to the superclass' constructor. If there's no constructor, any injected attributes of the superclass _are_ inherited. Both cases are encoded as `unknown` right now, so this is a case that you'd need to take into account :-)

piotrtomiak:
@JoostK The change looks good. It's indeed more straightforward and avoids any troubles with types. Do you still plan to backport the change to Angular 9?

JoostK:
@piotrtomiak My current understanding is that this can land in master, so that would correspond with the 9.1 release.

piotrtomiak:
@JoostK Great! 

I've just noticed one more problem related to injections - probably worth a separate issue. I am now looking into several Angular related issues in WebStorm and there is an interesting one with Transloco directive (https://github.com/ngneat/transloco/blob/master/projects/ngneat/transloco/src/lib/transloco.directive.ts), which can be both regular and structural directive, since `TemplateRef` parameter is optional. I need to determine, whether a directive should be provided for code completion as regular or as structural or as both, otherwise `ngIf` and `ngClass` would be listed in code completion and validated in both contexts. Ivy metadata doesn't hold any information about optional parameters and neither it holds info about custom injection tokens. Do you think that `@Optional` parameters should be forced to include `| null` in their types, so that one can determine which parameters are optional simply from their type, or should you include information about optional parameters in `ɵɵFactoryDef`, e.g.:
```
static ɵfac: i0.ɵɵFactoryDef<TranslocoDirective, [false, true, true, true, true, false, false, false]>;
```
What about injection tokens? At some point we might want to add support for validating modules to check whether all required injection tokens have been provided, or to provide smart code completion with available injection tokens.

piotrtomiak:
@JoostK It seems though that heuristics based on presence of `ElementRef`, `TemplateRef` and `ViewContainerRef` types in the constructor are enough to determine kind of directive, so it doesn't seem that I would need anything more now in Ivy metadata. However, I think that it might be worth considering adding info for `@Optional` and `@Inject` for a future use.

JoostK:
@piotrtomiak I had an idea to encode constructor decorators compactly and consistently:

```ts
constructor(@Attribute('type') type: string, @Optional() @Inject(TOKEN) foo: unknown, bar: Bar) {}
```

```ts
[ɵɵAttribute<'type'>, ɵɵInject<typeof TOKEN> | ɵɵOptional, null]
```

piotrtomiak:
@JoostK This looks pretty nice and consistent. So far the best idea, I would say. I guess it would need to be placed into `ɵɵFactoryDef`, which again could make all more consistent, as `@Attribute` is also an injector related information.

IgorMinar (approved):
lgtm for instruction api change - I'm deferring to other reviewers for the rest of the PR

piotrtomiak:
@JoostK Recent changes looks great. When do you plan to merge to into master?

JoostK:
@piotrtomiak Well, it's quite broken currently and we decided to cut down on the functionality for now. The reason is that emitting `@Inject` has some difficulties:

1. the `typeof` operator needs to be followed with a qualified name, so an arbitrary expression cannot be inserted verbatim (this is also true for `@Attribute`s but probably less prevalent). 
1. A token value may be internal to the file, in which case it won't be available in the .d.ts (so it's impossible to include in the metadata).
1. Imports may need to be emitted into the .d.ts file in some situations.

So for now, we decided not to include `@Inject` tokens in the metadata.


Compared to my previous suggestion, the format has been tweaked to become an object literal type instead:

```ts
{ attribute: 'attr', optional: true, self: true, skipSelf: true, host: true }
```

What's left to do is to move this data into the `FactoryDef` as you suggested, as that makes the most sense going forward. We may later revisit including an `@Inject` token where possible.

piotrtomiak:
@JoostK Thanks! I think that with flexibility of proposed API, addition of another property should not be a problem at a later stage.

piotrtomiak:
@JoostK We are closing to our release date of 2020.1, with RC coming on 31.03. I would love to merge my part into WebStorm codebase and have full Ivy support in 2020.1. When will you be able to merge the change to master?

packages/compiler/src/render3/r3_factory.ts
```diff
@@ -141,6 +141,13 @@ export interface R3DependencyMetadata {
    */
   token: o.Expression;
 
+  /**
+   * If an @Attribute decorator is present, this is the literal type of the attribute name, or
+   * the unknown type if no literal type is available (e.g. the attribute name is an expression).
+   * Will be null otherwise.
+   */
+  type: o.Expression|null;
```
alxhub: Shouldn't this be:
```suggestion
  attribute: o.Expression|null;
```
or
```suggestion
  attribute: string|null;
```

I think we should just pass the attribute name as a `string` here.

alxhub (approved):

IgorMinar (approved):
lgtm for codegen api changes.

JoostK:
merge-assistance: please run presubmit and merge if green 🍏

mhevery:
[presubmit](http://test/OCL:302694105:BASE:302694170:1585070666202:a6044229)

JoostK:
@piotrtomiak Sorry it took me so long to confirm, I didn't have an estimate. The good news is that this has landed for the 9.1 release!

piotrtomiak:
@JoostK It's great to see this PR merged for 9.1! Thank you for implementing this :) I have managed to include my part in WS 2020.1 :)

JoostK:
@piotrtomiak Excellent! As a JetBrains fan and customer I am delighted to hear that :-) Thanks for your input on the design here.

piotrtomiak:
@JoostK You're welcome, I am happy to have my small contribution to the Angular project :)

PR #35705, opened 2020-02-27
perf(core): adding micro benchmark for host bindings

AndrewKushnir:
This commit adds micro benchmark for host bindings, so that we can assess the impact of changes related to host bindings (for example PR #35568).

## PR Type

- [x] Other: adding new perf benchmark

- [ ] Yes
- [x] No


AndrewKushnir:
FYI this micro benchmark is needed to measure perf impact of #35568 (and any other future changes to host bindings).

mhevery (approved):

PR #35711, opened 2020-02-27
build: enable IDE type checking of $localize code

petebacondarwin:
The `packages/localize/src/tools` folder was excluded
from the top level `tsconfig.json` which meant that in IDEs
these source files were not being given the correct configuration.
It was originally excluded because it required the native `node` typings
but this is no longer a requirement.

Removing this folder from the exclusion list exposed a new issue
where there was a typings mismatch between `@babel/...` sources
and the associated `@types/babel__...` typings packages.

A clean up of the package.json and yarn.lock appears to fix this.


devversion (approved):
LGTM. For later reference: The cache had to be invalidated because the fallback cache brings in nested versions of `@babel/types` (e.g. in `node_modules/@types/babel__core/node_modules/@babel/types`).

This caused a version mismatch in `d.ts` module resolution, and caused TypeScript errors. To fix this caching issue, the cache needs to be invalidated. It might be a good idea to explore/investigate this more in the future. Created DEV-51 for tracking.

devversion:
cc. @josephperrott. It might be good if you have another look too.

Issue #35717, opened 2020-02-27
NGCC async mode master issue


alan-agius4:
This is a master issue of issues/improvements needed for NGCC async mode to be more successful with the CLI integration.

1) Allow termination (CTRL+C) - Currently it is not possible to terminate NGCC async process. 
2) Clustering issue - Fork clustering without properly setting up the master is not ideal when used via the API. More context in: TOOL-1339 (Can be solved by using `cluster.setupMaster`)
3) Unable to provide a custom logger. As a workaround and interim for the issue 2. In the CLI we will most likely setup the cluster master ourselves in the webpack plugin. However with this approach we cannot use a custom logger

```ts
cluster.setupMaster({
  exec: require.resolve('@angular/compiler-cli/ngcc/main-ngcc.js'),
  args: [
    '--basePath',
    this._nodeModulesDirectory,
    '--propertiesToConsider',
    ...this.propertiesToConsider,
    '--compileAllFormats',
    'false',
    '--async',
  ],
});
```
3) Reduce worker code path - Reduce workers by pointing the cluster workers to start from a different entrypoint (Can use `cluster.setupMaster()`), this should speed up the worker boot time.
4) ~~Don't spawn unneeded workers - Spawning workers is expensive and currently we are spawning unwanted workers, when 1, there are no packages to process, 2, when we need to process a number of packages which is less than the available CPU count - 1~~. Addressed in #35719
5) Running the new integration NGCC async as opposed to NGCC sync during module resolution causes a build time regression of 8/10 seconds on an `ng new` application with Angular Material when all modules have been processed previously. (Seems to be related to `analyzeEntryPoints`) Note: in my app I am importing all material entrypoints. 
```ts
import { BrowserModule } from '@angular/platform-browser';
import { NgModule } from '@angular/core';

import { AppRoutingModule } from './app-routing.module';
import { AppComponent } from './app.component';
import { BrowserAnimationsModule } from '@angular/platform-browser/animations';


import { MatExpansionModule } from '@angular/material/expansion';
import { MatGridListModule } from '@angular/material/grid-list';
import { MatTreeModule } from '@angular/material/tree';
import { MatBottomSheetModule } from '@angular/material/bottom-sheet';
import { MatButtonModule } from '@angular/material/button';
import { MatButtonToggleModule } from '@angular/material/button-toggle';
import { MatCardModule } from '@angular/material/card';
import { MatCheckboxModule } from '@angular/material/checkbox';
[...]

PR #35719, opened 2020-02-27
perf(ngcc): spawn workers eagerly

alan-agius4:
With this change we spawn workers eagerly based on the amount of work that needs to be done and changed the behaviour in `analyzeEntryPoints`, to only create tasks for non-processed formats.

Before this change we spawned the maximum of workers possible. However, in some cases there are less tasks than the max number of workers which resulted in created unnecessary workers

Reference: #35717

### DIFF
--- a/packages/compiler-cli/ngcc/src/execution/cluster/master.ts
+++ b/packages/compiler-cli/ngcc/src/execution/cluster/master.ts
@@ -41,6 +41,10 @@ export class ClusterMaster {
   }
 
   run(): Promise<void> {
+    if (this.taskQueue.allTasksCompleted) {
+      return Promise.resolve();
+    }
+
     // Set up listeners for worker events (emitted on `cluster`).
     cluster.on('online', this.wrapEventHandler(worker => this.onWorkerOnline(worker.id)));
 
@@ -51,10 +55,8 @@ export class ClusterMaster {
         'exit',
         this.wrapEventHandler((worker, code, signal) => this.onWorkerExit(worker, code, signal)));
 
-    // Start the workers.
-    for (let i = 0; i < this.workerCount; i++) {
-      cluster.fork();
-    }
+    // Since we have pending tasks at the very minimum we need a single worker.
+    cluster.fork();
 
     return this.finishedDeferred.promise.then(() => this.stopWorkers(), err => {
       this.stopWorkers();
@@ -98,11 +100,16 @@ export class ClusterMaster {
       isWorkerAvailable = false;
     }
 
-    // If there are no available workers or no available tasks, log (for debugging purposes).
     if (!isWorkerAvailable) {
-      this.logger.debug(
-          `All ${this.taskAssignments.size} workers are currently busy and cannot take on more ` +
-          'work.');
+      if (this.taskAssignments.size < this.workerCount) {
+        this.logger.debug('Spawning another worker process as there is more work to be done.');
+        cluster.fork();
+      } else {
+        // If there are no available workers or no available tasks, log (for debugging purposes).
+        this.logger.debug(
+            `All ${this.taskAssignments.size} workers are currently busy and cannot take on more ` +
+            'work.');
+      }
     } else {
       const busyWorkers = Array.from(this.taskAssignments)
                               .filter(([_workerId, task]) => task !== null)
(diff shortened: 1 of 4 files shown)


packages/compiler-cli/ngcc/src/main.ts
```diff
@@ -207,6 +207,10 @@ export function mainNgcc(
       }
 
       for (const formatProperty of propertiesToProcess) {
+        if (hasBeenProcessed(entryPoint.packageJson, formatProperty)) {
+          continue;
+        }
```
gkalpak: Suggested changes:
- Remove the `hasBeenProcessed` check from the `CompileFn` below.
- Retain the current logging behavior: ``logger.debug(`Skipping ${entryPoint.name} : ${formatProperty} (already compiled).`);``
- Remove [TaskProcessingOutcome.AlreadyProcessed][1], as it is not used any more. (Theoretically, we could remove `TaskProcessingOutcome` altogether, but I would keep it in case we need to add other outcomes in the future (e.g. handle errors differently).)

[1]: https://github.com/angular/angular/blob/d3c0d92d2/packages/compiler-cli/ngcc/src/execution/api.ts#L78-L79

packages/compiler-cli/ngcc/src/main.ts
```diff
@@ -33,7 +33,7 @@ import {ParallelTaskQueue} from './execution/task_selection/parallel_task_queue'
 import {SerialTaskQueue} from './execution/task_selection/serial_task_queue';
 import {ConsoleLogger, LogLevel} from './logging/console_logger';
 import {Logger} from './logging/logger';
-import {hasBeenProcessed, markAsProcessed} from './packages/build_marker';
+import {hasBeenProcessed} from './packages/build_marker';
```
gkalpak: 💯

gkalpak (changes requested):
Thx, @alan-agius4 :tada:

Can you please split this up into two commits (one for lazily workers and one for checking `hasBeenProcessed` upfront)?
Also, in the commit message: `eagerly` --> `lazily` :grin:

packages/compiler-cli/ngcc/src/execution/cluster/master.ts
```diff
@@ -256,11 +256,11 @@ export class ClusterMaster {
    * Wrap an event handler to ensure that `finishedDeferred` will be rejected on error (regardless
    * if the handler completes synchronously or asynchronously).
    */
-  private wrapEventHandler<Args extends unknown[]>(fn: (...args: Args) => void|Promise<void>):
```
gkalpak: Why this change? Are there performance (or other) implications?
If not, I would prefer to keep it, because otherwise it is easy to introduce an error by making a handler async.

gkalpak (changes requested):

packages/compiler-cli/ngcc/src/execution/cluster/master.ts
alan-agius4: I don't think that is correct, `cluster.on` is not a callback based methods using `async/await` in it might result in unexpected behaviour. Though probably, in this case might not, since you are not interested in the resolution of the `wrapEventHandler` itself.

alan-agius4: Or am I missing something?

alan-agius4: Actually I take that back!

gkalpak: So, what is your current opinion on it, I'm confused 😕 :grin:

alan-agius4: I reverted the commit

gkalpak (approved):
Nice! Thx, @alan-agius4 :+1:

AndrewKushnir:
@gkalpak @alan-agius4 quick question: should this change be master-only (i.e. go to `9.1.0`) or it should also be included in `9.0.x`? Currently it's merged to master only (according to the `PR target` label).

alan-agius4:
@AndrewKushnir, I checked and it should apply cleanly. So let's merge it to the branch path as well.

atscott:
Hi @alan-agius4, since this was already merged and closed, can you open a new PR that targets patch only and copy these changes in?

alan-agius4:
@atscott here you go https://github.com/angular/angular/pull/35832

PR #35736, opened 2020-02-28
perf(core): use multiple directives in host bindings micro benchmark

AndrewKushnir:
This commit updates the host bindings micro benchmark to run tests with multiple directives (where each directive contains host bindings). The number of directives is configurable as a constant in the micro benchmark file. This change is needed to have an ability to measure/compare perf in different scenarios.

## PR Type

- [x] Other: micro benchmark update.

- [ ] Yes
- [x] No


pkozlowski-opensource (approved):
LGTM

PR #35756, opened 2020-02-28
perf(ngcc): reduce directory traversing

alan-agius4:
This reduces the time that `findEntryPoints` takes from `9701.143ms` to `4177.278ms`, by reducing the file operations done.

Reference: #35717

### DIFF
--- a/packages/compiler-cli/ngcc/src/entry_point_finder/directory_walker_entry_point_finder.ts
+++ b/packages/compiler-cli/ngcc/src/entry_point_finder/directory_walker_entry_point_finder.ts
@@ -50,16 +50,19 @@ export class DirectoryWalkerEntryPointFinder implements EntryPointFinder {
     }
 
     if (entryPoints.length > 0) {
-      // The `sourceDirectory` is an entry-point itself so no need to search its sub-directories.
-      // Also check for any nested node_modules in this package.
-      const nestedNodeModulesPath = this.fs.join(sourceDirectory, 'node_modules');
-      if (this.fs.exists(nestedNodeModulesPath)) {
-        entryPoints.push(...this.walkDirectoryForEntryPoints(nestedNodeModulesPath));
+      // The `sourceDirectory` is an entry point itself so no need to search its sub-directories.
+      // Also check for any nested node_modules in this package but only if it was compiled by
+      // Angular.
+      // It is unlikely that a non Angular entry point has a dependency on an Angular library.
+      if (entryPoints.some(e => e.compiledByAngular)) {
+        const nestedNodeModulesPath = this.fs.join(sourceDirectory, 'node_modules');
+        if (this.fs.exists(nestedNodeModulesPath)) {
+          entryPoints.push(...this.walkDirectoryForEntryPoints(nestedNodeModulesPath));
+        }
       }
 
       return entryPoints;
     }
-
 
     this.fs
         .readdir(sourceDirectory)
--- a/packages/compiler-cli/ngcc/src/entry_point_finder/directory_walker_entry_point_finder.ts
+++ b/packages/compiler-cli/ngcc/src/entry_point_finder/directory_walker_entry_point_finder.ts
@@ -9,13 +9,11 @@ import {AbsoluteFsPath, FileSystem, join, resolve} from '../../../src/ngtsc/file
 import {DependencyResolver, SortedEntryPointsInfo} from '../dependencies/dependency_resolver';
 import {Logger} from '../logging/logger';
 import {NgccConfiguration} from '../packages/configuration';
-import {EntryPoint, getEntryPointInfo} from '../packages/entry_point';
+import {EntryPoint, INVALID_ENTRY_POINT, NO_ENTRY_POINT, getEntryPointInfo} from '../packages/entry_point';
 import {PathMappings} from '../utils';
 import {NGCC_DIRECTORY} from '../writing/new_entry_point_file_writer';
-
 import {EntryPointFinder} from './interface';
 import {getBasePaths} from './utils';
-
 
 /**
  * An EntryPointFinder that searches for all entry-points that can be found given a `basePath` and
@@ -96,11 +94,11 @@ export class DirectoryWalkerEntryPointFinder implements EntryPointFinder {
         getEntryPointInfo(this.fs, this.config, this.logger, packagePath, packagePath);
 
     // If there is no primary entry-point then exit
-    if (topLevelEntryPoint === undefined) {
+    if (topLevelEntryPoint === NO_ENTRY_POINT) {
       return [];
     }
 
-    if (topLevelEntryPoint === null) {
+    if (topLevelEntryPoint === INVALID_ENTRY_POINT) {
       return null;
     }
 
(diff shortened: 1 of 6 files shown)


packages/compiler-cli/ngcc/src/packages/entry_point.ts
```diff
@@ -80,17 +80,18 @@ export const SUPPORTED_FORMAT_PROPERTIES: EntryPointJsonProperty[] =
  *
  * @param packagePath the absolute path to the containing npm package
  * @param entryPointPath the absolute path to the potential entry-point.
- * @returns An entry-point if it is valid, `null` otherwise.
+ * @returns An entry-point if it is valid, `undefined` when it's not an entrypoint, `null`
+ * otherwise.
```
petebacondarwin: Could you be a bit more explicit by what "not an entry-point" and "otherwise" mean?
* `undefined`: there is no package.json at the path and there is no config to force an entry-point
* `null`: there is a package.json but it is not a valid Angular compiled entry-point

packages/compiler-cli/ngcc/test/packages/entry_point_spec.ts
```diff
@@ -128,7 +128,7 @@ runInEachFileSystem(() => {
       const entryPoint = getEntryPointInfo(
           fs, config, new MockLogger(), SOME_PACKAGE,
           _('/project/node_modules/some_package/missing_package_json'));
-      expect(entryPoint).toBe(null);
+      expect(entryPoint).toBe(undefined);
     });
```
petebacondarwin: Can we have a test for the case when `null` is returned?

alan-agius4: There is https://github.com/angular/angular/blob/5bc39f8c8d5238a9be9bd968cf18ea4b738bd6be/packages/compiler-cli/ngcc/test/packages/entry_point_spec.ts#L58

packages/compiler-cli/ngcc/src/packages/entry_point.ts
alan-agius4: Updated!

packages/compiler-cli/ngcc/src/entry_point_finder/directory_walker_entry_point_finder.ts
```diff
@@ -113,26 +131,25 @@ export class DirectoryWalkerEntryPointFinder implements EntryPointFinder {
...
             // We are not interested in symbolic links
             return;
           }
 
-          fn(path, stat.isDirectory());
-
-          if (stat.isDirectory()) {
-            this.walkDirectory(packagePath, path, fn);
+          const containsEntryPoint = fn(absolutePath, stat.isDirectory());
+          if (containsEntryPoint) {
```
petebacondarwin: This check seems suspicious... isn't it possible to have an entry-point more than one directory below the package with no entry-point in between? E.g. `package/directory/entry-point`?

packages/compiler-cli/ngcc/test/packages/entry_point_spec.ts
petebacondarwin: ...except that you have changed that test to check for `undefined`!

petebacondarwin (changes requested):
Thanks for this work @alan-agius4 - I can already see we will get benefits from this.

I get very nervous when we have code that starts using `null` and `undefined` to mean significant things. They are too magical and it is not easy to see immediately what they mean.

How about we modify the return value of these functions to be an object that has various flags to indicate the state of the result?

packages/compiler-cli/ngcc/src/entry_point_finder/directory_walker_entry_point_finder.ts
alan-agius4: I guess there is nothing to stop a user from creating such format, though I don't think I ever saw this.

alan-agius4:
Test for `null` -> https://github.com/angular/angular/blob/41b5b2714eb19a4dea334a92bccf1cb4ac66f214/packages/compiler-cli/ngcc/test/packages/entry_point_spec.ts#L168

> I get very nervous when we have code that starts using null and undefined to mean significant things. They are too magical and it is not easy to see immediately what they mean.

Totally agree, though my worry would be that this will require some refactoring in the tests mainly. I'll look at this tomorrow.

packages/compiler-cli/ngcc/src/entry_point_finder/directory_walker_entry_point_finder.ts
alan-agius4: And in such case most likely it will get transformed by the sync version of the NGCC, when integrated with the CLI.

packages/compiler-cli/ngcc/src/entry_point_finder/directory_walker_entry_point_finder.ts
```diff
@@ -40,66 +43,81 @@ export class DirectoryWalkerEntryPointFinder implements EntryPointFinder {
...
+  walkDirectoryForEntryPoints(sourceDirectory: AbsoluteFsPath): EntryPoint[] {
     const entryPoints = this.getEntryPointsForPackage(sourceDirectory);
+    if (entryPoints === null) {
+      return [];
+    }
+
     if (entryPoints.length > 0) {
       // The `sourceDirectory` is an entry-point itself so no need to search its sub-directories.
+      // Also check for any nested node_modules in this package.
+      const nestedNodeModulesPath = this.fs.join(sourceDirectory, 'node_modules');
```
alan-agius4: I was thinking of adding a check here, that we only try to traverse nested  `node_modules`, if one of the entrypoints is compiled by Angular.

It is very unlikely that a non Angular entry point has a dependency on an Angular library. 

Thoughts @petebacondarwin?

petebacondarwin: I was thinking about this too. There is a very small chance, maybe, that someone creates a holding library that aggregates together a bunch of Angular libraries and re-exports them. But I think that this might well break ngcc anyway, since we would need to process the re-exported typings for it to work with ngtsc anyway.

So perhaps we can go with this idea.

packages/compiler-cli/ngcc/src/entry_point_finder/directory_walker_entry_point_finder.ts
```diff
@@ -11,9 +11,12 @@ import {Logger} from '../logging/logger';
 import {NgccConfiguration} from '../packages/configuration';
 import {EntryPoint, getEntryPointInfo} from '../packages/entry_point';
 import {PathMappings} from '../utils';
+import {NGCC_DIRECTORY} from '../writing/new_entry_point_file_writer';
+
```
petebacondarwin: I know these extra blank lines are added by clang but you can remove them as they are uneeded, especially the one below.

packages/compiler-cli/ngcc/src/entry_point_finder/directory_walker_entry_point_finder.ts
petebacondarwin: OK let's go with this for now and revert if we get a complaint.

petebacondarwin (approved):

PR #35780, opened 2020-02-29
build(docs-infra): use local version of Zone.js when testing against local packages

gkalpak:
In some cases, we want to test the AIO app or docs examples against the locally built Angular packages (for example to ensure that the changes in a commit do not introduce a breaking change). In order to achieve this, we have the `ng-packages-installer` script that handles updating a project's `package.json` file to use the locally built Angular packages (and appropriate versions for their (dev-/peer-)dependencies).

Previously, `ng-packages-installer` would only consider the locally built Angular packages (from `dist/packages-dist/`). However, given that Zone.js is now part of the `angular/angular` repo, it makes sense to also use the locally built Zone.js package (from `dist/zone.js-dist/`). Otherwise, the tests might fail for commits that update both the Angular packages (and related docs examples) and the Zone.js package. An example of such a simultaneous change (that would have broken tests) is #33838.

This commit updates the script to install the locally built Zone.js package (in addition to the Angular ones). The commit ensures that the Zone.js package will always be available alongside the Angular packages (i.e. that the Zone.js package will be built by the same script that builds the Angular packages and that the `dist/zone.js-dist/` directory will be cached on CI).

Note: This problem was discovered while enabling docs examples unit tests in #34374.


scripts/build/build-packages-dist.js
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env node
```
IgorMinar: Most of us have the original path to this file in our muscle memory. Can you please leave a placeholder script in the old location that will error instructing the developers to use the new path?

Also please leave a comment there that the placeholder can be removed two months from now (early May).

IgorMinar (approved):
Lgtm with a nit.

petebacondarwin (approved):

PR #35784, opened 2020-03-01
perf(core): add micro benchmark for destroy hook invocation

crisbeto:
Adds a micro benchmark which excercises the logic that invokes `ngOnDestroy` hooks.

### DIFF
--- a/packages/core/test/render3/perf/setup.ts
+++ b/packages/core/test/render3/perf/setup.ts
@@ -19,10 +19,12 @@ const rendererFactory: RendererFactory3 =
     isBrowser ? domRendererFactory3 : new MicroBenchmarkRendererFactory;
 const renderer = rendererFactory.createRenderer(null, null);
 
-export function createAndRenderLView(parentLView: LView, tView: TView, hostTNode: TViewNode) {
+export function createAndRenderLView(
+    parentLView: LView, tView: TView, hostTNode: TViewNode): LView {
   const embeddedLView = createLView(
       parentLView, tView, {}, LViewFlags.CheckAlways, null, hostTNode, rendererFactory, renderer);
   renderView(tView, embeddedLView, null);
+  return embeddedLView;
 }
 
 export function setupRootViewWithEmbeddedViews(
--- a/packages/core/test/render3/perf/view_destroy_hook/index.ts
+++ b/packages/core/test/render3/perf/view_destroy_hook/index.ts
@@ -1,0 +1,78 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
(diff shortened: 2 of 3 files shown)


packages/core/test/render3/perf/view_destroy_hook/index.ts
```diff
@@ -0,0 +1,27 @@
...
+
+// scenario to benchmark
+const viewDestroy = createBenchmark('view destroy hook');
+const refreshTime = viewDestroy('destroy');
+const destroyHook = () => {};
+
+// create a view with 1000 sub-views that have onDestroy hooks and destroy them
+console.profile('view_destroy_hook');
+while (refreshTime()) {
+  const rootLView = setupRootViewWithEmbeddedViews(
```
crisbeto: Note: I'm not totally sure whether that this is the correct approach, because it's timing both creation and destruction. Ideally we'd do `while (refreshTime()) destroyViewTree()`, but from what I can tell, that'll destroy the views once and then keep trying to destroy the already-destroyed views. I'm open to suggestions on how to do it better.

devversion (approved):
LGTM for dev-infra.

packages/core/test/render3/perf/setup.ts
```diff
@@ -79,7 +82,11 @@ export function setupTestHarness(
   // create embedded views and add them to the container
   for (let i = 0; i < noOfViews; i++) {
     const lView = createEmbeddedLView();
-    insertView(lView[TVIEW], lView, lContainer, i);
+    const tView = lView[TVIEW];
+    insertView(tView, lView, lContainer, i);
+    if (destroyHook !== null) {
+      (tView.destroyHooks || (tView.destroyHooks = [])).push(i, destroyHook);
+    }
```
pkozlowski-opensource: Hmmmmm, I think that I would rather prefer that you create a real component / directive with a destroy hook and use it in an embedded view. This way:
* we can test different scenarios easily (directives, directives with providers that have onDestroy hook etc.);
* we don't have to modify `setupRootViewWithEmbeddedViews`;
* we've got less dependencies on the internal impl details (`destroyViewTree`)

WDYT?

packages/core/test/render3/perf/view_destroy_hook/index.ts
pkozlowski-opensource: Yeh, the way I understand the current perf harness is that we don't have a place for prepare / tear-down logic that wouldn't be measured. But I think that it is fine in this particular case as registering destroy hooks takes time as well.

pkozlowski-opensource (commented):
I think that I would rather prefer that you create a real component / directive with a destroy hook and use it in an embedded view, see: https://github.com/angular/angular/pull/35784/files#r386395833

WDYT?

crisbeto:
I've reworked it based on the discussions @pkozlowski-opensource. Can you take another look?

pkozlowski-opensource (approved):
LGTM, thnx!

PR #35792, opened 2020-03-02
fix(bazel): ng_package rule creates incorrect UMD module exports 

### LINKED ISSUE #18652: [a11y] Typo in @angular/cdk/a11y package breaks build
#### Reproduction

Steps to reproduce:
1. `npm i @angular/cdk`
2. Open node_modules/@angular/cdk/bundles/cdk-a11y.umd.js
3. There you will find the following piece of code:

`...((global.ng = global.ng || {}, global.ng.cdk = global.ng.cdk || {}, global.ng.cdk.a11Y = {}), global.ng.common, ....`
 
Especially the UMD ID `global.ng.cdk.a11Y` is wrong, since every other module of the @angular/cdk or @angular/material framework looks for an UMD ID of `global.ng.cdk.a11y` (small writte 'y')
 
#### Expected Behavior

What behavior were you expecting to see?

UMD ID of `global.ng.cdk.a11y`

#### Actual Behavior

What behavior did you actually see?

UMD ID of `global.ng.cdk.a11Y`

#### Environment

- CDK/Material: 9.1.0

devversion:
The `ng_package` rule currently creates incorrect UMD module exports
if an entry-point has a module name with numbers included.

For example, consider an entry-point called `@angular/cdk/a11y`. The UMD
module name should be `ng.cdk.a11y`. Instead, `ng_package` currently generates
an UMD module export called `ng.cdk.a11Y`.

This is because the logic for converting dash-case to camel case is
invalid as it uses Starlark's `title()` method. The title method
converts text to title case while we actually just want to capitalize
the first letter of a dash-case segment.

Fixes angular/components#18652.

### DIFF
--- a/packages/bazel/src/ng_package/ng_package.bzl
+++ b/packages/bazel/src/ng_package/ng_package.bzl
@@ -77,7 +77,7 @@ def _convert_dash_case_to_camel_case(s):
     parts = s.split("-")
 
     # First letter in the result is always unchanged
-    return s[0] + "".join([p.title() for p in parts])[1:]
+    return s[0] + "".join([p.capitalize() for p in parts])[1:]
 
 # Convert from a package name on npm to an identifier that's a legal global symbol
 #  @angular/core -> ng.core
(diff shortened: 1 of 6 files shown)


josephperrott (approved):
LGTM

PR #35793, opened 2020-03-02
fix(localize): translation file parser selection

### LINKED ISSUE #35525: Update Angular 8 to 9 problem with angular.json translation file format
# 🐞 bug report

### Affected Package

### Is this a regression?

### Description
After update to Angular 9 appears when I try to build with a configuration appears this error:
`ng serve client --configuration=default` or `ng build client --configuration=default`
**An unhandled exception occurred: Unsupported translation file format.**
image
image
image

Before update to Angular 9 (Work fine!)
image
After update to Angular 9
image

`package.json`
```
      "root": "apps/client",
      "sourceRoot": "apps/client/src",
      "prefix": "app",
      "architect": {
        "build": {
          "builder": "@angular-builders/custom-webpack:browser",
          "options": {
            "indexTransform": "apps/client/index-html-transform.js",
            "outputPath": "dist/apps/client",
            "index": "apps/client/src/index.html",
            "main": "apps/client/src/main.ts",
            "polyfills": "apps/client/src/polyfills.ts",
            "tsConfig": "apps/client/tsconfig.app.json",
...
            "default": {
              "fileReplacements": [
                {
                  "replace": "apps/client/src/environments/environment.ts",
                  "with": "apps/client/src/environments/environment.prod.ts"
                }
              ],
              "optimization": true,
              "outputHashing": "none",
              "sourceMap": false,
              "extractCss": true,
              "namedChunks": false,
              "aot": true,
              "extractLicenses": true,
              "vendorChunk": false,
              "buildOptimizer": true,
              "budgets": [
                {
                  "type": "initial",
                  "maximumWarning": "2mb",
                  "maximumError": "5mb"
                }
              ],
              "i18nFile": "i18n/default.xliff",
              "i18nFormat": "xlf",
              "i18nLocale": "en",
              "i18nMissingTranslation": "warning"
            },
...
        "extract-i18n": {
          "builder": "@angular-devkit/build-angular:extract-i18n",
          "options": {
            "browserTarget": "client:build",
            "i18nFormat": "xlf",
            "outputPath": "../../i18n",
            "outFile": "messages.xliff"
          }
        },
...
```

Update:
Working without IVY, but **I want IVY** for `ng build --configuration=default` but `ng serve` entryComponents error with;
```
  "angularCompilerOptions": {
    "enableIvy": false
  }
```

petebacondarwin:
These changes should fix #35525, and should be compatible with current CLI integration code (@clydin  to confirm), although the `hint` based approach with accompanied change to remove the need to pass `diagnostics` to constructors should also improve the performance and maintainability.

This should be completely backward compatible with previous versions.

FW-1888

### DIFF
--- a/packages/localize/src/tools/src/translate/translation_files/translation_parsers/translation_parser.ts
+++ b/packages/localize/src/tools/src/translate/translation_files/translation_parsers/translation_parser.ts
@@ -6,6 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 import {ɵMessageId, ɵParsedTranslation} from '@angular/localize/private';
+import {Diagnostics} from '../../../diagnostics';
 
 /**
 * An object that holds translations that have been parsed from a translation file.
@@ -13,6 +14,7 @@ import {ɵMessageId, ɵParsedTranslation} from '@angular/localize/private';
 export interface ParsedTranslationBundle {
   locale: string|undefined;
   translations: Record<ɵMessageId, ɵParsedTranslation>;
+  diagnostics: Diagnostics;
 }
 
 /**
--- a/packages/localize/src/tools/src/translate/translation_files/translation_parsers/translation_parser.ts
+++ b/packages/localize/src/tools/src/translate/translation_files/translation_parsers/translation_parser.ts
@@ -19,21 +19,57 @@ export interface ParsedTranslationBundle {
 
 /**
  * Implement this interface to provide a class that can parse the contents of a translation file.
+ *
+ * The `canParse()` method can return a hint that can be used by the `parse()` method to speed up
+ * parsing. This allows the parser to do significant work to determine if the file can be parsed
+ * without duplicating the work when it comes to actually parsing the file.
+ *
+ * Example usage:
+ *
+ * ```
+ * const parser: TranslationParser = getParser();
+ * const result = parser.canParse(filePath, content);
+ * if (result) {
+ *   return parser.parse(filePath, content, result);
+ * }
+ * ```
  */
-export interface TranslationParser {
+export interface TranslationParser<Hint = true> {
   /**
-   * Returns true if this parser can parse the given file.
+   * Can this parser parse the given file?
    *
    * @param filePath The absolute path to the translation file.
    * @param contents The contents of the translation file.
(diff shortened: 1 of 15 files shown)


mhevery (approved):
LGTM, But I am not sure if I am the best person to review, as I am not familiar with the codebase. I did not see anything which would jump out at me.

packages/localize/src/tools/src/translate/translation_files/translation_parsers/translation_parser.ts
```diff
@@ -6,32 +6,70 @@
...
+   * have been provided as the return result from calling `canParse()`.
+   * @returns The translation bundle parsed from the file.
+   * @throws No errors. If there was a problem with parsing the bundle will contain errors
+   * in the `diagnostics` property.
+   */
+  parse(filePath: string, contents: string, hint: Hint): ParsedTranslationBundle;
+  /**
+   * Parses the given file, extracting the target locale and translations.
+   *
+   * @deprecated This overload is kept for backward compatibility. Going forward use the Hint
```
clydin: Nit: I could see this overload still being useful if the format is already known.

clydin (approved):

packages/localize/src/tools/src/translate/translation_files/translation_parsers/translation_parser.ts
petebacondarwin: I see what you are saying, but I think it is easier, in the future, to un-deprecate this than to deprecate it, right?

clydin: Very true.

AndrewKushnir (approved):
LGTM, thanks @petebacondarwin! 👍

AndrewKushnir:
@petebacondarwin quick question: when can we remove deprecated APIs (after the corresponding release on CLI side)? Could you please create a ticket on that, so that we clean them up when possible? Thank you.

petebacondarwin:
@AndrewKushnir - regarding the deprecated overload. Since this is not actually a public API it does not have to follow our official deprecation rules. I just added the deprecation tag to remind the CLI to change their implementation. Once they have done so, we should be able to safely remove this overload. I will create a JIRA.

PR #35796, opened 2020-03-02
fix(core): allow null / undefined values in query results

### LINKED ISSUE #35673: Querying for child that provides itself as undefined throws under Ivy
#### Bug, feature request, or proposal:
Bug

#### What is the expected behavior?
Beeing able to use ViewChild to view children of cdkDropListGroup

#### What is the current behavior?
After updating to angular/material 9.0.1 and angular/cdk 9.0.1, I can no longer access cdkDropListGroup via ViewChild.

#### What are the steps to reproduce?

Use ViewChild on simple cdkDropList within a cdkDropListGroup:
`@ViewChild(CdkDropListGroup) listGroup: CdkDropListGroup<CdkDropList>;`

Console shows:
`ASSERTION ERROR: materialized query result should be defined [Expected=> null != undefined <=Actual]`

#### Which versions of Angular, Material, OS, TypeScript, browsers are affected?
"@angular/core": "9.0.2"
"@angular/material": "^9.0.1"
"@angular/cdk": "^9.0.1"
All browsers

pkozlowski-opensource:
Before this change ngIvy implementation of queries would throw upon
encountering null / undefined query result collected from an embedded
view. It turns out that we might have a provider that explicitly provides
a null / undefined value in a place of a token queried for.

This commit removes a check from the ngIvy query implementation that was
asserting on a query result to be defined.

Fixes #35673

### DIFF
--- a/packages/core/src/render3/query.ts
+++ b/packages/core/src/render3/query.ts
@@ -363,9 +363,7 @@ function collectQueryResults<T>(tView: TView, lView: LView, queryIndex: number, 
     for (let i = 0; i < tQueryMatches.length; i += 2) {
       const tNodeIdx = tQueryMatches[i];
       if (tNodeIdx > 0) {
-        const viewResult = lViewResults[i / 2];
-        ngDevMode && assertDefined(viewResult, 'materialized query result should be defined');
-        result.push(viewResult as T);
+        result.push(lViewResults[i / 2] as T);
       } else {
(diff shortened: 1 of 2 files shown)


mhevery (approved):

mhevery:
[presubmit](http://test/OCL:298375148:BASE:298375200:1583171165411:89a04a08)

PR #35801, opened 2020-03-02
docs-infra: include CLI options in search

petebacondarwin:
The first commit is just a reformatting - the real changes are in the second commit.

You can test it in the preview at https://pr35801-a121ba5.ngbuilds.io/. Try searching for `deleteOutputPath`. Looks like:

Screenshot 2020-03-02 at 18 07 44

### DIFF
--- a/aio/tools/transforms/cli-docs-package/processors/processCliCommands.spec.js
+++ b/aio/tools/transforms/cli-docs-package/processors/processCliCommands.spec.js
@@ -7,11 +7,7 @@ describe('processCliCommands processor', () => {
 
   const navigationStub = {
     docType: 'navigation-json',
-    data: {
-      SideNav: [{
-        children: [{'url': 'cli'}]
-      }]
-    }
+    data: {SideNav: [{children: [{'url': 'cli'}]}]}
   };
 
   beforeEach(() => {
@@ -21,17 +17,13 @@ describe('processCliCommands processor', () => {
     createDocMessage = injector.get('createDocMessage');
   });
 
-  it('should be available on the injector', () => {
-    expect(processor.$process).toBeDefined();
-  });
+  it('should be available on the injector', () => { expect(processor.$process).toBeDefined(); });
 
-  it('should run after the correct processor', () => {
-    expect(processor.$runAfter).toEqual(['extra-docs-added']);
-  });
+  it('should run after the correct processor',
+     () => { expect(processor.$runAfter).toEqual(['extra-docs-added']); });
 
-  it('should run before the correct processor', () => {
-    expect(processor.$runBefore).toEqual(['rendering-docs']);
-  });
+  it('should run before the correct processor',
+     () => { expect(processor.$runBefore).toEqual(['rendering-docs']); });
 
   it('should collect the names (name + aliases)', () => {
     const doc = {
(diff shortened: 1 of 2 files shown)


aio/tools/transforms/cli-docs-package/processors/processCliCommands.spec.js
```diff
@@ -95,6 +95,22 @@ describe('processCliCommands processor', () => {
...
+        commandAliases: [],
+        options: [
+          {name: 'named1'},
+          {name: 'positional1', positional: 0},
+          {name: 'named2', hidden: true},
+          {name: 'positional2', hidden: true, positional: 1},
+        ],
+      };
+      processor.$process([doc, navigationStub]);
+      expect(doc.optionKeywords).toEqual('named1  positional1');
```
gkalpak: Hm...why do we end up with two consecutive spaces?

gkalpak (approved):
I think it is worth having a test with subcommands too. Otherwise lgtm :+1:

aio/tools/transforms/cli-docs-package/processors/processCliCommands.spec.js
petebacondarwin: I wasn't quite sure - I think it might be the way that `Array.from(someSet)` works; but it doesn't impact the result and I couldn't be bothered to spend time working out why. 🙀

gkalpak: I doubt it is `Array.from()` alone. The only potential causes I can think of are that either one of the entries already contains a space or there is an empty option. I believe both sound problematic, so I would rather we figure it out 😇

petebacondarwin: 😆 agreed
It turns out it was the second reason.

petebacondarwin: This is because `aliases` are often `undefined`...

petebacondarwin: Done

petebacondarwin:
Added the extra test

aio/tools/transforms/cli-docs-package/processors/processCliCommands.spec.js
```diff
@@ -301,4 +301,36 @@ describe('processCliCommands processor', () => {
...
+            ],
+          }
+        },
+      }],
+    };
+    processor.$process([doc, navigationStub]);
+    expect(doc.optionKeywords)
+        .toEqual(
+            'supercommand subcommand1 subcommand1-option1 subcommand1-option2 subcommand2 subcommand2-option1 subcommand2-option2');
+  });
```
gkalpak: :100:

gkalpak (approved):

aio/tools/transforms/cli-docs-package/processors/processCliCommands.spec.js
gkalpak: Thx for looking into it ❤️

atscott:
Hi @petebacondarwin, I see that this is pending review from @IgorMinar. Could you remove him from the reviewers list if his review is not needed, get an approval from him, or add the merge assistance label and explain why his review is not needed? Thank you!

petebacondarwin:
I really did try to remove him but the GH UI was not letting me. I'll try again now.

petebacondarwin:
Nope. Still no good.

Merge-assistance: I cannot remove @IgorMinar from the list of reviewers. It appears to be a GH glitch. Notice that he appears twice in the list. The review from @gkalpak is enough to approve this pull request.

(Note to pull approve configurers - it is frustrating that PullApprove randomly chooses a single user from the list of candidates, since it invariably chooses someone inappropriate like @IgorMinar who is not really the front line reviewer for this sort of PR. It is then incumbent upon the PR author to search through to find a more appropriate reviewer. With CodeOwners this was less of a problem because all code-owners were assigned and then unassigned when their bit of code had been approved by someone else in their team.)

PR #35802, opened 2020-03-02
build: use nonstrict python toolchain to support PY2/PY3

josephperrott:


IgorMinar (changes requested):
CI failed with legit error

PR #35809, opened 2020-03-02
docs: remove `ivy` and mention `ve` label in docs

gkalpak:
This is a follow-up to #35799.


gkalpak:
The second commit assumes that there will be a `comp: ve` label (which currently doesn't exist). Not sure if we want that or not.

josephperrott (approved):
LGTM

Additionally, I will create the `comp: ve` label

PR #35810, opened 2020-03-02
fix(animations): allow computeStyle to work on elements created in Node

matsko:
This patch is a follow-up patch to 35c9f0dc2f3665d4f9d9ece328cee4559bbec9c6.
It changes the `computeStyle` function to handle situations where
non string based values are returned from `window.getComputedStyle`.
This situation usually ocurrs in Node-based test environments where
the element or `window.getComputedStyle` is mocked out.


alxhub (approved):

alxhub:
[Presubmit (normal)](http://test/OCL:298735704:BASE:298735777:1583286453516:ca996af4)

PR #35812, opened 2020-03-03
refactor(compiler): simplify parseTemplateBindings() for microsyntax

kyliau:
This commit is purely a refactoring of the logic in
`parseTemplateBindings` method for parsing the microsyntax expression.
This is done to enable the introduction of `keySpan` and `valueSpan` in
subsequent PR.

For a detailed explanation of this work and the subsequent work items,
please see https://docs.google.com/document/d/1mEVF2pSSMSnOloqOPQTYNiAJO0XQxA1H0BZyESASOrE/edit?usp=sharing

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/compiler/src/expression_parser/parser.ts
+++ b/packages/compiler/src/expression_parser/parser.ts
@@ -118,12 +118,38 @@ export class Parser {
         span, span.toAbsolute(absoluteOffset), prefix, uninterpretedExpression, location);
   }
 
-  parseTemplateBindings(tplKey: string, tplValue: string, location: any, absoluteOffset: number):
-      TemplateBindingParseResult {
-    const tokens = this._lexer.tokenize(tplValue);
+  /**
+   * Parse microsyntax template expression and return a list of bindings or
+   * parsing errors in case the given expression is invalid.
+   *
+   * For example,
+   * ```
+   *   <div *ngFor="let item of items">
+   *                ^ `absoluteOffset` for `tplValue`
+   * ```
+   * contains three bindings:
+   * 1. ngFor -> null
+   * 2. item -> NgForOfContext.$implicit
+   * 3. ngForOf -> items
+   *
+   * This is apparent from the de-sugared template:
+   * ```
+   *   <ng-template ngFor let-item [ngForOf]="items">
+   * ```
+   *
+   * @param templateKey name of directive, without the * prefix. For example: ngIf, ngFor
+   * @param templateValue RHS of the microsyntax attribute
+   * @param templateUrl template filename if it's external, component filename if it's inline
+   * @param absoluteOffset absolute offset of the `tplValue`
+   */
+  parseTemplateBindings(
+      templateKey: string, templateValue: string, templateUrl: string,
+      absoluteOffset: number): TemplateBindingParseResult {
+    const tokens = this._lexer.tokenize(templateValue);
     return new _ParseAST(
-               tplValue, location, absoluteOffset, tokens, tplValue.length, false, this.errors, 0)
-        .parseTemplateBindings(tplKey);
+               templateValue, templateUrl, absoluteOffset, tokens, templateValue.length,
+               false /* parseAction */, this.errors, 0 /* relative offset */)
+        .parseTemplateBindings(templateKey);
   }
 
   parseInterpolation(
@@ -721,11 +747,12 @@ export class _ParseAST {
   }
 
   /**
-   * An identifier, a keyword, a string with an optional `-` in between.
+   * Parses an identifier, a keyword, a string with an optional `-` in between.
    */
-  expectTemplateBindingKey(): string {
+  expectTemplateBindingKey(): {key: string, keySpan: ParseSpan} {
     let result = '';
     let operatorFound = false;
+    const start = this.inputIndex;
     do {
       result += this.expectIdentifierOrKeywordOrString();
       operatorFound = this.optionalOperator('-');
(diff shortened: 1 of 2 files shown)


packages/compiler/src/expression_parser/parser.ts
```diff
@@ -118,7 +118,20 @@ export class Parser {
         span, span.toAbsolute(absoluteOffset), prefix, uninterpretedExpression, location);
   }
 
-  parseTemplateBindings(tplKey: string, tplValue: string, location: any, absoluteOffset: number):
+  /**
+   * Parse microsyntax template expression and return errors, if any.
```
AndrewKushnir: May be rephrase this a bit to say something like:
```
Parse microsyntax template expression and return either an AST that represents this expression or parsing errors in case given expression in invalid.
```

packages/compiler/src/expression_parser/parser.ts
```diff
@@ -118,7 +118,20 @@ export class Parser {
...
+   * For example,
+   * ```
+   *   <div *ngFor="let item of items">
+   *                ^ absoluteOffset for `tplValue`
+   * ```
+   *
+   * @param tplKey name of directive, without the * prefix. For example: ngIf, ngFor
+   * @param tplValue string value for the microsyntax expression
+   * @param location
```
AndrewKushnir: Can we add a quick description of what this parameter is (to be consistent with other params)?

packages/compiler/src/expression_parser/parser.ts
```diff
@@ -723,77 +736,168 @@ export class _ParseAST {
...
+   * @param rawKey keyword in the expression, for example "of", "trackBy"
+   * @param rawKeySpan relative span of the `rawKey` in the attribute value
+   * @param absoluteOffset absolute offset of the attribute value
+   */
+  private parseDirectiveKeyword(
+      templateKey: string, rawKey: string, rawKeySpan: ParseSpan,
+      absoluteOffset: number): TemplateBinding[] {
+    const bindings: TemplateBinding[] = [];
+    let key = templateKey;
+    if (rawKey !== templateKey) {
```
AndrewKushnir: Just wondering in which case it would not be the case (may be add a comment on this)?

packages/compiler/src/expression_parser/parser.ts
```diff
@@ -723,77 +736,168 @@ export class _ParseAST {
...
+      this.optionalCharacter(chars.$COLON);  // trackBy: trackByFunction
+    }
+    const valueExpr = this.getDirectiveBoundTarget();
+    const span = new ParseSpan(rawKeySpan.start, this.inputIndex);
+    bindings.push(new TemplateBinding(
+        span, span.toAbsolute(absoluteOffset), key, false /* keyIsVar */, valueExpr?.source || '', valueExpr));
+    // The binding could optionally be followed by "as". For example,
+    // *ngIf="cond | pipe as x". In this case, the key in the "as" binding
+    // is "x" and the value is the template key itself ("ngIf").
+    const asBinding = this.parseValueAsKey(key, rawKeySpan, absoluteOffset);
```
AndrewKushnir: I was a bit confused by the function name (it sounds like we parse a value as a key) :) May be we can rename it to be more clear on what "as" stands for in function name, like `parseValueOfAsExpression`?

packages/compiler/src/expression_parser/parser.ts
```diff
@@ -723,77 +736,168 @@ export class _ParseAST {
...
+    if (asBinding) {
+      bindings.push(asBinding);
+    }
+    if (!this.optionalCharacter(chars.$SEMICOLON)) {
```
AndrewKushnir: I think this if statement might benefit from additional comments.

packages/compiler/src/expression_parser/parser.ts
```diff
@@ -723,77 +736,168 @@ export class _ParseAST {
...
+   * Return the binding for a variable declaration using `let`.
+   * For example,
+   * ```
+   * *ngFor="let item of items; let i=index;"
+   *         ^^^^^^^^           ^^^^^^^^^^^
+   * ```
+   * In the first binding, `item` is bound to `NgForOfContext.$implicit`.
+   * In the second binding, `i` is bound to `NgForOfContext.index`.
+   */
+  private parseLetDeclaration(): TemplateBinding|undefined {
```
AndrewKushnir: I glanced over other functions in the `parser.ts` script and it looks like we typically return `null` when we hit exit condition without any work performed. I think we should follow this in the newly added functions as well.

AndrewKushnir (changes requested):
Thanks a lot for separating this change out in a standalone PR @kyliau!

FYI I've added a few comments, nothing major, but may worth looking into.

Thank you.

packages/compiler/src/expression_parser/parser.ts
```diff
@@ -118,7 +118,20 @@ export class Parser {
         span, span.toAbsolute(absoluteOffset), prefix, uninterpretedExpression, location);
   }
 
-  parseTemplateBindings(tplKey: string, tplValue: string, location: any, absoluteOffset: number):
+  /**
+   * Parse microsyntax template expression and return errors, if any.
+   * For example,
+   * ```
+   *   <div *ngFor="let item of items">
+   *                ^ absoluteOffset for `tplValue`
```
ayazhafiz: Consider mentioning that absoluteOffset is the start of the template binding relative to the start of the source file it's in. This example doesn't make it clear what absoluteOffset actually is.

packages/compiler/src/expression_parser/parser.ts
```diff
@@ -723,77 +736,168 @@ export class _ParseAST {
   /**
    * An identifier, a keyword, a string with an optional `-` in between.
```
ayazhafiz:
```suggestion
   * Parses an identifier, a keyword, a string with an optional `-` in between.
```

packages/compiler/src/expression_parser/parser.ts
```diff
@@ -723,77 +736,168 @@ export class _ParseAST {
...
-    return result.toString();
+    return {
+      key: result,
+      keySpan: new ParseSpan(start, start + result.length),
+    };
   }
 
-  // Parses the AST for `<some-tag *tplKey=AST>`
+  /**
+   * Parses the bindings for microsyntax `<some-tag *tplKey=AST>`
```
ayazhafiz:
```suggestion
   * Parses the bindings for a microsyntax expression; e.g. the AST in <some-tag *tplKey=AST>
```

packages/compiler/src/expression_parser/parser.ts
```diff
@@ -723,77 +736,168 @@ export class _ParseAST {
...
+   * For example,
+   * ```
+   *   <div *ngFor="let item of items; index as i; trackBy: func">
+   *                ^ absoluteOffset for `this.input`
+   * ```
```
ayazhafiz: Maybe add an example of what we expect to be parsed out of here? (let items, of expression, "as i" expression, trackBy key expression)

packages/compiler/src/expression_parser/parser.ts
```diff
@@ -723,77 +736,168 @@ export class _ParseAST {
...
+    // In *ngIf="cond | pipe", key = "ngIf", value = "cond | pipe"
+    bindings.push(...this.parseDirectiveKeyword(
+        tplKey, tplKey, new ParseSpan(0, tplKey.length), this.absoluteOffset));
+
+    while (this.index < this.tokens.length) {
+      const letBinding = this.parseLetDeclaration();
+      if (letBinding) {
+        bindings.push(letBinding);
+      } else {
+        const {key, keySpan} = this.expectTemplateBindingKey();
```
ayazhafiz: Why do we expect a template binding key if this is not a let expression?

packages/compiler/src/expression_parser/parser.ts
```diff
@@ -723,77 +736,168 @@ export class _ParseAST {
...
+   */
+  private parseDirectiveKeyword(
+      templateKey: string, rawKey: string, rawKeySpan: ParseSpan,
+      absoluteOffset: number): TemplateBinding[] {
+    const bindings: TemplateBinding[] = [];
+    let key = templateKey;
+    if (rawKey !== templateKey) {
+      // Transform rawKey to actual key. Eg. of -> ngForOf, trackBy -> ngForTrackBy
+      key = templateKey + rawKey[0].toUpperCase() + rawKey.substring(1);
+      this.optionalCharacter(chars.$COLON);  // trackBy: trackByFunction
```
ayazhafiz: We should probably rename this to consumeOptionalCharacter or similar

packages/compiler/src/expression_parser/parser.ts
ayazhafiz: +1 and consider renaming to consumeOptionalCharacter

packages/compiler/src/expression_parser/parser.ts
```diff
@@ -723,77 +736,168 @@ export class _ParseAST {
...
+      return null;
+    }
+    const ast = this.parsePipe();  // example: "condition | async"
+    const {start, end} = ast.span;
+    const value = this.input.substring(start, end);
+    return new ASTWithSource(ast, value, this.location, this.absoluteOffset + start, this.errors);
+  }
+
+  private parseValueAsKey(value: string, valueSpan: ParseSpan, absoluteOffset: number):
+      TemplateBinding|undefined {
```
ayazhafiz: nit for consistency

```suggestion
      TemplateBinding|null {
```

packages/compiler/src/expression_parser/parser.ts
ayazhafiz: +1

packages/compiler/src/expression_parser/parser.ts
kyliau: Will rename this in standalone PR

packages/compiler/src/expression_parser/parser.ts
kyliau: done

kyliau:
I've addressed all comments, PTAL.
After an offline discussion with @AndrewKushnir, I went back to look at the existing test suites for the expression parser, and found that while there are many test cases, all of them are "fictional", so I added some test cases that better represent real world usage of the two most common microsyntax directives - `ngIf` and `ngFor`.
Please let me know what you think, thank you!

packages/compiler/src/expression_parser/parser.ts
```diff
@@ -118,11 +118,32 @@ export class Parser {
...
+   *                ^ absoluteOffset for `tplValue`
+   * ```
+   * contains three bindings:
+   * 1. ngFor -> null
+   * 2. item -> NgForOfContext.$implicit
+   * 3. ngForOf -> items
+   *
+   * @param tplKey name of directive, without the * prefix. For example: ngIf, ngFor
+   * @param tplValue RHS of the microsyntax attribute
+   * @param templateUrl template filename if it's external, component filename if it's inline
```
AndrewKushnir: nit: I assume `tpl` in `tplKey` and `tplValue` is a short version of `template`. If that's the case, the `templateUrl` (with full `template` word) seems a bit inconsistent. I think `location` (that you had before) was fine, it just needed a `@param` description.

packages/compiler/src/expression_parser/parser.ts
```diff
@@ -118,11 +118,32 @@ export class Parser {
...
+  /**
+   * Parse microsyntax template expression and return a list of bindings or
+   * parsing errors in case the given expression is invalid.
+
+   * For example,
+   * ```
+   *   <div *ngFor="let item of items">
+   *                ^ absoluteOffset for `tplValue`
+   * ```
+   * contains three bindings:
```
AndrewKushnir: May be we can also include the de-sugared version of the same example (de-sugar the `*ngFor`), so it'd be clear why we have 3 bindings?

packages/compiler/test/expression_parser/parser_spec.ts
```diff
@@ -317,6 +325,44 @@ describe('parser', () => {
       expect(bindings[0].expression !.location).toEqual('location');
     });
 
+    it('should support common usage of ngIf', () => {
```
AndrewKushnir: Thanks for adding extra tests! 👍

packages/compiler/test/expression_parser/parser_spec.ts
```diff
@@ -270,6 +270,14 @@ describe('parser', () => {
           binding => binding.expression != null ? binding.expression.source : null);
     }
 
+    function humanize(bindings: TemplateBinding[]): Array<[string, string | null, boolean]> {
+      return bindings.map(binding => {
+        const {key, expression, name, keyIsVar} = binding;
+        const value = keyIsVar ? name : expression ? expression.source : expression;
```
AndrewKushnir: May be wrap inner ternary expression into `()`, so that it's easier to read?

packages/compiler/src/expression_parser/parser.ts
AndrewKushnir: @kyliau could you plz add a quick comment on why this `if` statement is here and what's the goal? (I agree that renaming `optionalCharacter` can be a part of a separate PR).

AndrewKushnir (changes requested):

packages/compiler/src/expression_parser/parser.ts
kyliau: done

packages/compiler/src/expression_parser/parser.ts
kyliau: I prefer `templateUrl` over `location`, as the former is less ambiguous. To keep everything consistent I renamed `tplKey` and `tplValue` to `templateKey` and `templateValue`, respectively.

packages/compiler/src/expression_parser/parser.ts
kyliau: Instead of a comment, I refactored this into a standalone method called `consumeStatementTerminator()` and added documentation there.

packages/compiler/test/expression_parser/parser_spec.ts
kyliau: done. Actually, I'd like to use optional chaining and nullish coalescing here, but our version of clang-format doesn't support this yet. I filed a bug with dev infra.

kyliau (commented):
Addressed Andrew's comments

packages/compiler/src/expression_parser/parser.ts
```diff
@@ -721,79 +747,199 @@ export class _ParseAST {
...
+  /**
+   * Parse microsyntax template expression and return a list of bindings or
+   * parsing errors in case the given expression is invalid.
+
+   * For example,
+   * ```
+   *   <div *ngFor="let item of items; index as i; trackBy: func">
+   * ```
+   * contains five bindings:
+   * 1. ngFor -> null
```
AndrewKushnir: May be add de-sugared version here as well?

AndrewKushnir (approved):
Thanks for addressing all the comments @kyliau and once again thanks for additional documentation and tests 👍 The changes LGTM, but as we discussed offline, it'd be great if @alxhub would have a look at this PR too, since this PR touches pretty fundamental expression parsing logic (adding Alex as a reviewer to this PR). Thank you.

packages/compiler/src/expression_parser/parser.ts
ayazhafiz: No, there is no other meaning. I just mean that the comment is non-obvious in that it doesn't tell us much about why the absoluteOffset is noted in the example and why it would be useful. I think it would be helpful to add this information to the comment. But this is a nit.

packages/compiler/src/expression_parser/parser.ts
ayazhafiz: Also, consider adding a link to the template syntax grammar (https://angular.io/guide/structural-directives#grammar or https://gist.github.com/mhevery/d3530294cff2e4a1b3fe15ff75d08855). It may not be immediately clear why the binding for ngFor is null.

packages/compiler/src/expression_parser/parser.ts
```diff
@@ -721,79 +747,199 @@ export class _ParseAST {
...
+    bindings.push(...this.parseDirectiveKeyword(
+        templateKey, new ParseSpan(0, templateKey.length), this.absoluteOffset));
+
+    while (this.index < this.tokens.length) {
+      // If it starts with 'let', then this must be variable declaration
+      const letBinding = this.parseLetDeclaration();
+      if (letBinding) {
+        bindings.push(letBinding);
+      } else {
+        // Two possible cases here, either "value as key" or
```
ayazhafiz: nit: 
```suggestion
        // Two possible cases here, either `value "as" key` or
```

to make more clear that "as" is a literal rather than a variable. If you choose to apply this, the next two lines should be updated too.

packages/compiler/src/expression_parser/parser.ts
```diff
@@ -721,79 +747,199 @@ export class _ParseAST {
...
+        const {key, keySpan} = this.expectTemplateBindingKey();
+        // Peek at the next token, if it is "as" then this must be variable
+        // declaration.
+        const binding = this.parseValueAsKey(key, keySpan, this.absoluteOffset);
+        if (binding) {
+          bindings.push(binding);
         } else {
-          name = '\$implicit';
+          // Otherwise the key must be a directive keyword, like "of". Transform
+          // the key to actual key. Eg. of -> ngForOf, trackBy -> ngForTrackBy
```
ayazhafiz: nice comment!

packages/compiler/src/expression_parser/parser.ts
```diff
@@ -721,79 +747,199 @@ export class _ParseAST {
...
 
-    return new TemplateBindingParseResult(bindings, warnings, this.errors);
+  /**
+   * Return the binding for a variable declared using `as`. Note that the order
+   * of the key-value pair in this declaration is reversed. For example,
+   * ```
+   * *ngIf="cond | pipe as foo"
+   *  ^^^^                 ^^^
+   *  value                key
+   * ```
```
ayazhafiz: Is this always the case? For example, in the branch on line 801 - 810, it looks like the value could be "cond | pipe". Maybe there is a better way to name this function, like `parseAsBinding`?

packages/compiler/src/expression_parser/parser.ts
kyliau: ah I see what you mean. `absoluteOffset` is actually an input parameter to this function, which is why I called it out in the example. I've quoted it in backticks to make this more obvious. Let me know if this is better.
(in the next PR I'm going to differentiate between absoluteKeyOffset and absoluteValueOffset)

packages/compiler/src/expression_parser/parser.ts
kyliau: added the second link

packages/compiler/src/expression_parser/parser.ts
kyliau: No, the value cannot be "cond | pipe". In an "as" binding **that follows an expression**, the value is always the template key (ngFor, ngForOf, etc). This is documented in `parseDirectiveKeyword`. I've updated the example here to reflect the case in line 801-810.

kyliau: I've renamed
1. `parseDirectiveKeyword` -> `parseDirectiveKeywordBindings`
2. `parseLetDeclaration` -> `parseLetBinding`
3. `parseValueAsKey` -> `parseAsBinding`

Note that (1) is in plural form to reflect the fact that there could be two bindings.

packages/compiler/src/expression_parser/parser.ts
```diff
@@ -118,12 +118,38 @@ export class Parser {
         span, span.toAbsolute(absoluteOffset), prefix, uninterpretedExpression, location);
   }
 
-  parseTemplateBindings(tplKey: string, tplValue: string, location: any, absoluteOffset: number):
-      TemplateBindingParseResult {
-    const tokens = this._lexer.tokenize(tplValue);
+  /**
+   * Parse microsyntax template expression and return a list of bindings or
+   * parsing errors in case the given expression is invalid.
+
```
AndrewKushnir: Missing `*` char.

packages/compiler/src/expression_parser/parser.ts
```diff
@@ -721,79 +747,203 @@ export class _ParseAST {
...
+    };
   }
 
-  // Parses the AST for `<some-tag *tplKey=AST>`
-  parseTemplateBindings(tplKey: string): TemplateBindingParseResult {
-    let firstBinding = true;
+  /**
+   * Parse microsyntax template expression and return a list of bindings or
+   * parsing errors in case the given expression is invalid.
+
```
AndrewKushnir: Missing `*` char.

packages/compiler/src/expression_parser/parser.ts
```diff
@@ -721,79 +747,203 @@ export class _ParseAST {
...
+   * For example, "of items", "trackBy: func".
+   * The bindings are: ngForOf -> items, ngForTrackBy -> func
+   * There could be an optional "as" binding that follows the expression.
+   * For example,
+   * ```
+   * *ngFor="let item of items | slice:0:1 as collection".`
+   *                  ^^ ^^^^^^^^^^^^^^^^^ ^^^^^^^^^^^^^
+   *             keyword    bound target   optional 'as' binding
+   * ```
+
```
AndrewKushnir: Missing `*` char.

alxhub (approved):
Fantastic work!

kyliau:
g3 presubmit: http://test/OCL:298480280:BASE:298966468:1583367302889:60602a0a

ayazhafiz (approved):

PR #35832, opened 2020-03-03
perf(ngcc): only create tasks for non-processed formats (#35719)

alan-agius4:
Patch version of https://github.com/angular/angular/pull/35719


alan-agius4:
@caretaker, these changes have already been approved and landed on master (#35719).

atscott (approved):

PR #35834, opened 2020-03-03
build: creates a script to get contributor stats from across the angular org

josephperrott:
This script gets all of the current users for the organization and retrieves
information about PR/Issue contributions/authorship since a provided date.
Returning this information as a CSV.


tools/contributing-stats/get-data.ts
```diff
@@ -0,0 +1,254 @@
...
+ */
+
+import {graphql as unauthenticatedGraphql} from '@octokit/graphql';
+import * as minimist from 'minimist';
+import {alias, params, query as graphqlQuery, types} from 'typed-graphqlify';
+
+// The organization to be consider for the queries.
+const ORG = 'angular';
+// The repositories to be consider for the queries.
+const REPOS = ['angular', 'components', 'angular-cli'];
```
devversion: The `:repo` filter requires the user to be included to. e.g. `repo:angular/components`. Currently repo data does not work.

tools/contributing-stats/get-data.ts
```diff
@@ -0,0 +1,254 @@
...
+  // Whether the updated or created timestamp should be used.
+  const updatedOrCreated = args['use-created'] ? 'created' : 'updated';
+  let key_val_pairs: {[key: string]: {query: string, label: string}} = {};
+  // Add queries and params for all values queried for each repo.
+  for (let repo of REPOS) {
+    key_val_pairs = {
+      ...key_val_pairs,
+      [`${repo.replace(/[\/\-]/g, '_')}_issue_author`]: {
+        query: `repo:${repo} is:issue author:${username} ${updatedOrCreated}:>${date}`,
+        label: 'Issue Authored'
```
devversion: Should these column labels explicitly mention the repo they are representing? Right now it's hard to differentiate.

tools/contributing-stats/get-data.ts
```diff
@@ -0,0 +1,254 @@
...
+
+  /**
+   * Gets the graphql query object for the GraphQL query.
+   */
+  function getQuery(pairs: typeof key_val_pairs) {
+    const output: {[key: string]: {}} = {};
+    Object.keys(pairs).forEach((key: string) => {
+      output[alias(key, 'search')] = params(
+          {
+            query: `$${key}`,
```
devversion: Can't the queries be inlined here?

tools/contributing-stats/get-data.ts
```diff
@@ -0,0 +1,254 @@
...
+
+/**
+ * Build metadata for making requests for a specific user and date.
+ *
+ * Builds GraphQL query string, Query Params and Labels for making queries to GraphQl.
+ */
+function buildQueryAndParams(username: string, date: string) {
+  // Whether the updated or created timestamp should be used.
+  const updatedOrCreated = args['use-created'] ? 'created' : 'updated';
+  let key_val_pairs: {[key: string]: {query: string, label: string}} = {};
```
devversion: Nit: How about changing this variable to something more meaningful? e.g. `dataQueries`.

tools/contributing-stats/get-data.ts
devversion: I agree that what you proposed would be ideal. I agree that it will be more verbose in the CSV file, but once integrated somewhere, it's actually better to reason about it. The current way results in the CSV not being self-contained as one needs to understand how it has been generated.

tools/contributing-stats/get-data.ts
```diff
@@ -0,0 +1,254 @@
...
+  process.exit(1);
+}
+
+/**
+ * Authenticated instance of Github GraphQl API service, relies on a
+ * personal access token being available in the TOKEN environment variable.
+ */
+const graphql = unauthenticatedGraphql.defaults({
+  headers: {
+    authorization: `token ${process.env.TOKEN}`,
```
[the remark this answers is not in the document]
devversion: That's fair. Optional: Add a short comment about this. We could also use `TOKEN` as a fallback and have something more specific primarily.

tools/contributing-stats/get-data.ts
josephperrott: Done

tools/contributing-stats/get-data.ts
```diff
@@ -0,0 +1,241 @@
...
+ * The script expects the following flag(s):
+ *
+ * required:
+ *   --since [date] The data after which contributions are queried for.
+ *
+ * optional:
+ *  --use-created [boolean] If the created timestamp should be used for
+ *     time comparisons, defaults otherwise to the updated timestamp.
+ *     Uses githubs search format for dates, e.g. "2020-01-21".
+ *     See
```
devversion: Looks like this comment has accidentally moved to the wrong option.

devversion (approved):
LGTM. One minor comment (and needs lint fix)

IgorMinar (approved):
lgtm - global approval

PR #35840, opened 2020-03-03
fix(core): ngOnDestroy on multi providers called with incorrect context

### LINKED ISSUE #35231: Invalid ngOnDestroy invocation on multiple provision
# 🐞 bug report
### Affected Package
core? 9.0.0. Would like to know
### Is this a regression?
It occured after 

### Description
Not sure whether it is connected with https://github.com/angular/angular/issues/35167 or not, but following screen shows an error:
image
in summary, those services in array - all of them implements some interface and are provided with the same token, under provision with `multiple: true`. As you see, `ngOnDestroy` is invoked on that array, but this `this`... eh, I don't think that this array extends `Subscriber`. I have also noticed that this is not always the case.

## 🔥 Exception or Error
<pre><code>
ERROR Error: Uncaught (in promise): TypeError: this.clearSubscriptions is not a function
TypeError: this.clearSubscriptions is not a function
    at Array.ngOnDestroy (subscriber.ts:28)
    at executeOnDestroys (core.js:14024)
    at cleanUpView (core.js:13931)
    at destroyViewTree (core.js:13705)
    at destroyLView (core.js:13882)
    at RootViewRef.destroy (core.js:14737)
    at ComponentRef$1.destroy (core.js:33528)
    at RouterOutlet.deactivate (router.js:8912)
    at ActivateRoutes.deactivateRouteAndOutlet (router.js:3963)
    at ActivateRoutes.deactivateRouteAndItsChildren (router.js:3921)
    at resolvePromise (zone-evergreen.js:793)
    at resolvePromise (zone-evergreen.js:752)
    at zone-evergreen.js:854
    at ZoneDelegate.invokeTask (zone-evergreen.js:400)
    at Object.onInvokeTask (core.js:40744)
    at ZoneDelegate.invokeTask (zone-evergreen.js:399)
    at Zone.runTask (zone-evergreen.js:168)
    at drainMicroTaskQueue (zone-evergreen.js:570)
    at invokeTask (zone-evergreen.js:485)
    at ZoneTask.invoke (zone-evergreen.js:470)
</code></pre>

## 🌍  Your Environment

**Angular Version:**
<pre><code>
Angular CLI: 9.0.1
Node: 12.13.1
OS: win32 x64

Angular: 9.0.0
... animations, cdk, common, compiler, compiler-cli, core
... elements, forms, language-service, material
... platform-browser, platform-browser-dynamic, router
Ivy Workspace: Yes

@angular-devkit/architect          0.900.1
@angular-devkit/build-angular      0.900.1
@angular-devkit/build-ng-packagr   0.900.1
@angular-devkit/build-optimizer    0.900.1
@angular-devkit/build-webpack      0.900.1
@angular-devkit/core               8.3.23
@angular-devkit/schematics         8.3.23
@angular/cli                       9.0.1
@ngtools/webpack                   9.0.1
@schematics/angular                8.3.23
@schematics/update                 0.900.1
</code></pre>

crisbeto:
Currently destroy hooks are stored in memory as `[1, hook, 5, hook]` where the numbers represent the index at which to find the context and `hook` is the function to be invoked. This breaks down for `multi` providers, because the value at the index will be an array of providers, resulting in the hook being invoked with an array of all the multi provider values, rather than the provider that was destroyed. In ViewEngine `ngOnDestroy` wasn't being called for `multi` providers at all.

These changes fix the issue by changing the structure of the destroy hooks to `[1, hook, 5, [0, hook, 3, hook]]` where the indexes inside the inner array point to the provider inside of the multi provider array. Note that this is slightly different from the original design which called for the structure to be `[1, hook, 5, [hook, hook]`, because in the process of implementing it, I realized that we wouldn't get passing the correct context if only some of the `multi` providers have `ngOnDestroy` and others don't.

I've run the newly-added `view_destroy_hooks` benchmark against these changes and compared it to master. The difference seems to be insignificant (between 1% and 2% slower).

Fixes #35231.

### DIFF
--- a/packages/core/src/render3/di_setup.ts
+++ b/packages/core/src/render3/di_setup.ts
@@ -10,6 +10,7 @@
 import {resolveForwardRef} from '../di/forward_ref';
 import {ClassProvider, Provider} from '../di/interface/provider';
 import {isClassProvider, isTypeProvider, providerToFactory} from '../di/r3_injector';
+import {assertDefined} from '../util/assert';
 
 import {diPublicInInjector, getNodeInjectable, getOrCreateNodeInjectorForNode} from './di';
 import {ɵɵdirectiveInject} from './instructions/all';
@@ -17,7 +18,7 @@ import {DirectiveDef} from './interfaces/definition';
 import {NodeInjectorFactory} from './interfaces/injector';
 import {TContainerNode, TDirectiveHostNode, TElementContainerNode, TElementNode, TNodeProviderIndexes} from './interfaces/node';
 import {isComponentDef} from './interfaces/type_checks';
-import {LView, TData, TVIEW, TView} from './interfaces/view';
+import {DestroyHookData, LView, TData, TVIEW, TView} from './interfaces/view';
 import {getLView, getPreviousOrParentTNode, getTView} from './state';
 
 
@@ -149,7 +150,7 @@ function resolveProvider(
         if (!isViewProvider && doesViewProvidersFactoryExist) {
           lInjectablesBlueprint[existingViewProvidersFactoryIndex].providerFactory = factory;
         }
-        registerDestroyHooksIfSupported(tView, provider, tInjectables.length);
+        registerDestroyHooksIfSupported(tView, provider, tInjectables.length, 0);
         tInjectables.push(token);
         tNode.directiveStart++;
         tNode.directiveEnd++;
@@ -160,16 +161,19 @@ function resolveProvider(
         lView.push(factory);
       } else {
         // Cases 1.b and 2.b
-        registerDestroyHooksIfSupported(
-            tView, provider, existingProvidersFactoryIndex > -1 ?
-                existingProvidersFactoryIndex :
-                existingViewProvidersFactoryIndex);
-        multiFactoryAdd(
-            lInjectablesBlueprint ![isViewProvider ? existingViewProvidersFactoryIndex : existingProvidersFactoryIndex],
+        const indexInFactory = multiFactoryAdd(
+            lInjectablesBlueprint!
+                [isViewProvider ? existingViewProvidersFactoryIndex :
+                                  existingProvidersFactoryIndex],
             providerFactory, !isViewProvider && isComponent);
+        registerDestroyHooksIfSupported(
+            tView, provider,
+            existingProvidersFactoryIndex > -1 ? existingProvidersFactoryIndex :
+                                                 existingViewProvidersFactoryIndex,
+            indexInFactory);
       }
       if (!isViewProvider && isComponent && doesViewProvidersFactoryExist) {
-        lInjectablesBlueprint[existingViewProvidersFactoryIndex].componentProviders !++;
+        lInjectablesBlueprint[existingViewProvidersFactoryIndex].componentProviders!++;
       }
     }
   }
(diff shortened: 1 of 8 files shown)


packages/core/src/render3/di_setup.ts
```diff
@@ -180,15 +181,34 @@ function resolveProvider(
...
+      if (!providerIsTypeProvider && ((provider as ClassProvider)).multi) {
+        ngDevMode &&
+            assertDefined(
+                indexInFactory, 'indexInFactory when registering multi factory destroy hook');
+        const existingCallbacksIndex = hooks.indexOf(contextIndex);
+
+        if (existingCallbacksIndex === -1) {
+          hooks.push(contextIndex, [indexInFactory, ngOnDestroy]);
+        } else {
+          (hooks[existingCallbacksIndex + 1] as(number | (() => void))[])
```
pkozlowski-opensource: nit: do you think it would make sense to create a type / types for  `(number | (() => void))[])`? We seem to be using similar casts in other places and having a type with a name / description would aid understanding, maybe?

crisbeto: It would definitely make things easier to read, but I'm not sure what we'd call it.

pkozlowski-opensource: Right. So I think that we've got multiple types here:

* `HookCallback` / `HookFn`: `() => void`
* `HookEntry` / `HookData`: `number | HookFn`
* `HooksData` / `Hooks`: `HookData[]`
* `DestroyHookData`: (HookData | HookData[])[]

This is just hand-wavy proposal but yeh, I've found the current types and type casts hard to read.

pkozlowski-opensource (changes requested):
The change LGTM overall but I _really_ think that we should adjust types as the code gets hard to follow. I've suggested some adjustments but I think that you've got more context than me at this point.

Happy to chat more about it if needed.

crisbeto:
Thank you for the feedback @pkozlowski-opensource, it definitely made it easier to follow 👍 Can you take another look?

Also regarding the payload increase, I think it's due to master being right on the edge, rather than the changes in this PR which are fairly minimal and don't introduce new dependencies.

packages/core/src/render3/interfaces/view.ts
```diff
@@ -734,7 +743,20 @@ export interface RootContext {
...
+ * Array of destroy hooks that should be executed for a view and their directive indices.
+ *
+ * The array is set up as a series of number/function or number/(number|function)[]:
+ * - Even indices represent the context with which hooks should be called.
+ * - Odd indices are the hook functions themselves. If a value at an odd index is an array,
+ *   it represents the destroy hooks of a `multi` provider where:
+ *     - Even indices represent the index of the provider for which we've registered a destroy hook,
+ *       inside of the `multi` provider array.
+ *     - Odd indices are the destroy hook functions.
+ */
```
pkozlowski-opensource: nit: I think that it would be good to add an example, showing an example service(s) and the LView structure.

pkozlowski-opensource (approved):
LGTM, you might consider adding an example (see the comment in-line)

packages/core/src/render3/interfaces/view.ts
crisbeto: Added.

kara (changes requested):
A few minor things

Can we also add a note to the commit that notes the differences from VE and how we expect this fix to make it more compatible?

crisbeto:
Sorry for the increased diff, `clang-format` reformatted all the files I touched after I rebased. The formatting changes are in the last `fixup` commit and my latest changes are in the second-to-last commit.

kara (approved):
LGTM

(George shouldn't need to approve this one since AIO is only size changes)

kara:
[presubmit](http://test/OCL:305044634:BASE:305044655:1586189869333:3d7d903e)

kara:
@crisbeto Just realized the commit message hasn't been updated yet. Can you update it re https://github.com/angular/angular/pull/35840#pullrequestreview-374620905

crisbeto:
Fixed @kara. I had to squash the commits myself, because I messed up the initial rebase.

PR #35841, opened 2020-03-03
fix(bazel): do not use manifest paths for generated imports within compilation unit

devversion:
Currently, the `ng_module` rule incorrectly uses manifest paths for
generated imports from the Angular compiler.

This breaks packaging as prodmode output (i.e. `esnext`) is copied in
various targets (`es5` and `es2015`) to the npm package output. Example:

_node_modules/my-pkg/es2015/imports/public-api.js_
```ts
import * as i1 from "angular/packages/bazel/test/ng_package/example/imports/second";
```

The above import is invalid in NPM and leaks information about the source project
structure. The import should be relative:

```ts
import * as i1 from "./second";
```

The imports can, and **should** be relative so that the files of the same compilation unit are self-contained and do not rely on custom module resolution.

### DIFF
--- a/packages/bazel/src/ngc-wrapped/index.ts
+++ b/packages/bazel/src/ngc-wrapped/index.ts
@@ -113,19 +113,17 @@ export function runOneBuild(args: string[], inputs?: {[path: string]: string}): 
     }
   }
 
-  const expectedOuts = config['angularCompilerOptions']['expectedOut'];
+  // These are options passed through from the `ng_module` rule which aren't supported
+  // by the `@angular/compiler-cli` and are only intended for `ngc-wrapped`.
+  const {expectedOut, _useManifestPathsAsModuleName} = config['angularCompilerOptions'];
 
   const {basePath} = ng.calcProjectFileAndBasePath(project);
   const compilerOpts = ng.createNgCompilerOptions(basePath, config, tsOptions);
   const tsHost = ts.createCompilerHost(compilerOpts, true);
   const {diagnostics} = compile({
     allDepsCompiledWithBazel: ALL_DEPS_COMPILED_WITH_BAZEL,
-    compilerOpts,
-    tsHost,
-    bazelOpts,
-    files,
-    inputs,
-    expectedOuts
+    useManifestPathsAsModuleName: _useManifestPathsAsModuleName,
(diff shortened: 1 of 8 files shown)


gregmagolan (approved):
LGTM. This code is much too complicated. We will be able to replace this with ts_library with angular as a plugin in the future once we can drop
ViewEngine support.

devversion:
Yep. Totally agreed. And yes! I'm waiting for that to happen. The current system is too complicated. I tried to improve this by adding comments. Also the fact that we rely on summaries, but not on metadata files, makes ngc-wrapped even more complicated.

kyliau (approved):

kyliau:
g3 presubmit: http://test/OCL:298679132:BASE:298679262:1583270817811:1d706ca6

kyliau:
g3 presubmit: http://test/OCL:298679132:BASE:298853465:1583338254767:e12dceb4

kyliau:
g3 presubmit: http://test/OCL:298679132:BASE:298918004:1583355304817:c6251c4b

kyliau:
g3 presubmit: http://test/OCL:298679132:BASE:299157401:1583438236266:2d80d68f

packages/bazel/src/ngc-wrapped/index.ts
```diff
@@ -290,20 +291,32 @@ export function compile({allDepsCompiledWithBazel = true, compilerOpts, tsHost,
...
-    const result = doFileNameToModuleName(importedFilePath);
-    fileNameToModuleNameCache.set(importedFilePath, result);
+    const result = doFileNameToModuleName(importedFilePath, containingFilePath);
+    fileNameToModuleNameCache.set(cacheKey, result);
     return result;
   };
 
-  function doFileNameToModuleName(importedFilePath: string): string {
+  function doFileNameToModuleName(importedFilePath: string, containingFilePath?: string): string {
+    const relativeTargetPath =
```
kyliau: this needs to be moved to line 358 to preserve the original behavior.

kyliau:
g3 presubmit: http://test/OCL:298679132:BASE:299217178:1583452908596:d0fe404c

gkalpak (approved):

PR #35843, opened 2020-03-03
fix(core): use console to show error if ErrorHandler not present

mhevery:
The errors used to be swallowed if no `Injector` or `ErrorHandler` were present. The new implementation falls back to `console` or re-throws if no `console` is present.

Using classical bootstrap we are guaranteed to have `Injector`. This can only happen if we bootstrap using `renderComponent` which is not public yet and so it it is used in tests only.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/core/src/render3/instructions/shared.ts
+++ b/packages/core/src/render3/instructions/shared.ts
@@ -2021,7 +2021,15 @@ export function loadComponentRenderer(
 export function handleError(lView: LView, error: any): void {
   const injector = lView[INJECTOR];
   const errorHandler = injector ? injector.get(ErrorHandler, null) : null;
-  errorHandler && errorHandler.handleError(error);
+  if (errorHandler !== null) {
+    errorHandler.handleError(error);
+    // tslint:disable-next-line:no-console
+  } else if (typeof console === 'object' && console !== null) {
+    // tslint:disable-next-line:no-console
+    (console.error || console.log)(error);
+  } else {
+    throw error;
+  }
 }
 
 /**
--- a/packages/core/test/render3/instructions/shared_spec.ts
+++ b/packages/core/test/render3/instructions/shared_spec.ts
@@ -6,13 +6,54 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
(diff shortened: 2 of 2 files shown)


packages/core/src/render3/instructions/shared.ts
```diff
@@ -2021,7 +2021,15 @@ export function loadComponentRenderer(
...
-  errorHandler && errorHandler.handleError(error);
+  if (errorHandler !== null) {
+    errorHandler.handleError(error);
+    // tslint:disable-next-line:no-console
+  } else if (typeof console === 'object' && console !== null) {
+    // tslint:disable-next-line:no-console
+    (console.error || console.log)(error);
+  } else {
+    throw error;
+  }
```
AndrewKushnir, line 2032: It looks like this function is called in one place and we already throw there:

https://github.com/angular/angular/blob/00291ae0543c2cc142479ea2580a90c74e5a63f5/packages/core/src/render3/instructions/shared.ts#L1889-L1900

So the main goal of this function is to communicate the error to the `ErrorHandler` if it's present (which it does). I'm wondering if this change is needed?

Quick comment: we may also consider wrapping the `handleError` into `ngDevMode` to tree-shake away this function in prod builds (that should also avoid keeping a reference to the `ErrorHandler` token).

jessicajaniuk: I think @AndrewKushnir has a handler on this one.

mhevery: What happens if `errorHandler` is not present? As of right now the exception is eaten, an this is what this is trying to solve.

AndrewKushnir: If the `errorHandler` is not present, it looks like the original error would still be re-thrown in the `detectChangesInternal` function.

mhevery:
Closing as not on issue any more

PR #35844, opened 2020-03-03
Revert "fix(compiler): type-checking error for duplicate variables in…

atscott:
… templates (#35674)"

This reverts commit 12072954dea2dd4fab05c0391182d503ae132d2c.

Because this is a new error, it shouldn't go in the patch branch.


IgorMinar (approved):
lgtm - global approval

IgorMinar:
please document the revert reason in the commit message (not just the PR description)

atscott:
Just spoke with @alxhub and he said that this was [already an error](https://github.com/angular/angular/blob/00f3c58bb929d9afa10ab3c4f1b9ee7a0077ba57/packages/compiler/src/render3/view/template.ts#L1640) that wasn't very useful and was made into a proper diagnostic error so we should not revert the commit.

PR #35798, opened 2020-03-02
fix(core): log error instead of warning for unknown properties and elements

crisbeto:
Changes the Ivy unknown element/property messages from being logged with `console.warn` to `console.error`. This should make them a bit more visible without breaking existing apps. Furthermore, a lot of folks filter out warning messages in the dev tools' console, whereas errors are usually still shown.

Fixes #35699.

PR #35845, opened 2020-03-03
Revert "fix(core): log error instead of warning for unknown propertie…

atscott:
…s and elements (#35798)"

This reverts commit 00f3c58bb929d9afa10ab3c4f1b9ee7a0077ba57.

rolling back because it could be breaking e2e tests that assert that there are no errors in the console after the assertions have run. we can readd this in v10.


IgorMinar (approved):
lgtm - global approval,

but please add the revert reason into the commit message.

PR #35848, opened 2020-03-04
feat(bazel): transform generated shims (in Ivy) with tsickle

alxhub:
Currently, when Angular code is built with Bazel and with Ivy, generated
factory shims (.ngfactory files) are not processed via the majority of
tsickle's transforms. This is a subtle effect of the build infrastructure,
but it boils down to a TsickleHost method `shouldSkipTsickleProcessing`.

For ngc_wrapped builds (Bazel + Angular), this method is defined in the
`@bazel/typescript` (aka bazel rules_typescript) implementation of
`CompilerHost`. The default behavior is to skip tsickle processing for files
which are not present in the original `srcs[]` of the build rule. In
Angular's case, this includes all generated shim files.

For View Engine factories this is probably desirable as they're quite
complex and they've never been tested with tsickle. Ivy factories however
are smaller and very straightforward, and it makes sense to treat them like
any other output.

This commit adjusts two independent implementations of
`shouldSkipTsickleProcessing` to enable transformation of Ivy shims:

* in `@angular/bazel` aka ngc_wrapped, the upstream `@bazel/typescript`
  `CompilerHost` is patched to treat .ngfactory files the same as their
  original source file, with respect to tsickle processing.

  It is currently not possible to test this change as we don't have any test
  that inspects tsickle output with bazel. It will be extensively tested in
  g3.

* in `ngc`, Angular's own implementation is adjusted to allow for the
  processing of shims when compiling with Ivy. This enables a unit test to
  be written to validate the correct behavior of tsickle when given a host
  that's appropriately configured to process factory shims.

For ngtsc-as-a-plugin, a similar fix will need to be submitted upstream in
tsc_wrapped.

### DIFF
--- a/packages/bazel/src/ngc-wrapped/index.ts
+++ b/packages/bazel/src/ngc-wrapped/index.ts
@@ -184,7 +184,7 @@ export function compile({allDepsCompiledWithBazel = true, compilerOpts, tsHost, 
   }
 
   // Detect from compilerOpts whether the entrypoint is being invoked in Ivy mode.
-  const isInIvyMode = compilerOpts.enableIvy === 'ngtsc';
+  const isInIvyMode = !!compilerOpts.enableIvy;
 
   // Disable downleveling and Closure annotation if in Ivy mode.
   if (isInIvyMode) {
@@ -245,9 +245,18 @@ export function compile({allDepsCompiledWithBazel = true, compilerOpts, tsHost, 
         files, compilerOpts, bazelOpts, tsHost, fileLoader, generatedFileModuleResolver);
   }
 
-  // Also need to disable decorator downleveling in the BazelHost in Ivy mode.
   if (isInIvyMode) {
+    // Also need to disable decorator downleveling in the BazelHost in Ivy mode.
     bazelHost.transformDecorators = false;
+
+    const delegate = bazelHost.shouldSkipTsickleProcessing.bind(bazelHost);
+    bazelHost.shouldSkipTsickleProcessing = (fileName: string) => {
+      // The base implementation of shouldSkipTsickleProcessing checks whether `fileName` is part of
+      // the original `srcs[]`. For Angular (Ivy) compilations, ngfactory/ngsummary files that are
+      // shims for original .ts files in the program should be treated identically. Thus, strip the
+      // '.ngfactory' or '.ngsummary' part of the filename away before calling the delegate.
+      return delegate(fileName.replace(/\.(ngfactory|ngsummary)\.ts$/, '.ts'));
+    };
   }
 
   // Prevent tsickle adding any types at all if we don't want closure compiler annotations.
--- a/packages/compiler-cli/src/main.ts
+++ b/packages/compiler-cli/src/main.ts
@@ -103,8 +103,9 @@ function createEmitCallback(options: api.CompilerOptions): api.TsEmitCallback|un
       tsickle.TsickleHost, 'shouldSkipTsickleProcessing'|'pathToModuleName'|
       'shouldIgnoreWarningsForPath'|'fileNameToModuleId'|'googmodule'|'untyped'|
       'convertIndexImportShorthand'|'transformDecorators'|'transformTypesToClosure'> = {
-    shouldSkipTsickleProcessing: (fileName) =>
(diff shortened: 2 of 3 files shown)


alxhub:
I intend to run a global presubmit on this PR.

alxhub:
[Presubmit(normal)](http://test/OCL:298735704:BASE:298736582:1583286737814:cb5618b3)

alxhub:
[Presubmit(global)](https://test.corp.google.com/ui#id=OCL:299029045:BASE:299029499:1583393104423:5d45a80c)

packages/bazel/src/ngc-wrapped/index.ts
```diff
@@ -245,9 +245,18 @@ export function compile({allDepsCompiledWithBazel = true, compilerOpts, tsHost,
...
+    // Also need to disable decorator downleveling in the BazelHost in Ivy mode.
     bazelHost.transformDecorators = false;
+
+    const delegate = bazelHost.shouldSkipTsickleProcessing.bind(bazelHost);
+    bazelHost.shouldSkipTsickleProcessing = (fileName: string) => {
+      // The base implementation of shouldSkipTsickleProcessing checks whether `fileName` is part of
+      // the original `srcs[]`. For Angular (Ivy) compilations, ngfactory/ngsummary files that are
+      // shims for original .ts files in the program should be treated identically. Thus, strip the
+      // '.ngfactory' or '.ngsummary' part of the filename away before calling the delegate.
+      return delegate(fileName.replace(/\.(ngfactory|ngsummary)\.ts$/, ''));
```
AndrewKushnir: The string replace operation will also strip `.ts` extension, i.e. `test.ngfactory.ts` -> `test`. If this is desirable, let's may be  update a comment above (last sentence) to say that we also strip the `.ts` part of the filename?

packages/compiler-cli/src/main.ts
```diff
@@ -103,8 +103,8 @@ function createEmitCallback(options: api.CompilerOptions): api.TsEmitCallback|un
       tsickle.TsickleHost, 'shouldSkipTsickleProcessing'|'pathToModuleName'|
       'shouldIgnoreWarningsForPath'|'fileNameToModuleId'|'googmodule'|'untyped'|
       'convertIndexImportShorthand'|'transformDecorators'|'transformTypesToClosure'> = {
-    shouldSkipTsickleProcessing: (fileName) =>
-                                     /\.d\.ts$/.test(fileName) || GENERATED_FILES.test(fileName),
+    shouldSkipTsickleProcessing: (fileName) => /\.d\.ts$/.test(fileName) ||
```
AndrewKushnir: May be add a quick comment here too on why we exclude generated files from Tsickle processing in VE? It's explained in the PR description and I think it'd be great to have a quick note here as well.

packages/compiler-cli/test/ngtsc/ngtsc_spec.ts
```diff
@@ -4594,6 +4594,32 @@ runInEachFileSystem(os => {
           expect(trim(jsContents).startsWith(trim(fileoverview))).toBeTruthy();
         });
 
+        it('should be produced for generated factory files', () => {
+          env.tsconfig({
+            'annotateForClosureCompiler': true,
+            'generateNgFactoryShims': true,
+
```
AndrewKushnir: nit: extra empty line

AndrewKushnir (changes requested):
Thanks for a very detailed description of the problem and the fix @alxhub 👍 

I've left a couple comments, could you plz have a look? Thank you.

packages/bazel/src/ngc-wrapped/index.ts
alxhub: Fixed!

alxhub:
[Presubmit](http://test/OCL:299029045:BASE:299419231:1583528823927:d03654a8) with the minor changes.

AndrewKushnir (approved):
LGTM, thanks for updates @alxhub! 👍

IgorMinar (approved):
lgtm - thanks!!

IgorMinar:
@alxhub consider making this a master&patch PR

EatingW:
Thanks, Alex!

PR #35850, opened 2020-03-04
fix(compiler): process `imports` first and `declarations` second while calculating scopes

### LINKED ISSUE #35502: User defined pipes vs builtin pipes
# 🐞 bug report

### Affected Package

 The issue is caused by package 
@angular/common

### Is this a regression?

I think it is.

### Description

User can define his own pipe with name `currency` for example. Previously angular used *user* pipe instead of builtin. Ivy does opposite silently.

## 🔬 Minimal Reproduction

You can copy & paste any pipe from `number_pipe.ts` modify its transform method and see that pipe will be ignored.

## 🔥 Exception or Error

No error, but it should be.

## 🌍  Your Environment

**Angular Version:**
<pre><code>

Angular: 9.0.1
... common, compiler, compiler-cli, core, forms
... router

Package                      Version
------------------------------------------------------
@angular-devkit/architect    0.803.9 (cli-only)
@angular-devkit/core         8.3.9 (cli-only)
@angular-devkit/schematics   8.3.9 (cli-only)
@angular/cdk                 9.0.0
@schematics/angular          8.3.9 (cli-only)
@schematics/update           0.803.9 (cli-only)
rxjs                         6.5.4
typescript                   3.7.5
webpack                      4.41.6
</code></pre>

AndrewKushnir:
Prior to this commit, while calculating the scope for a module, Ivy compiler processed `declarations` field first and `imports` after that. That results in a couple issues:

* for Pipes with the same `name` and present in `declarations` and in an imported module, Pipe from imported module was selected. In View Engine the logic is opposite: Pipes from `declarations` field receive higher priority.
* for Directives with the same selector and present in `declarations` and in an imported module, we first invoked the logic of a Directive from `declarations` field and after that - imported Directive logic. In View Engine, it was the opposite and the logic of a Directive from the `declarations` field was invoked last.

In order to align Ivy and View Engine behavior, this commit updates the logic in which we populate module scope: we first process all imports and after that handle `declarations` field. As a result, in Ivy both use-cases listed above work similar to View Engine.

Resolves #35502.

## PR Type

- [x] Bugfix

- [ ] Yes
- [x] No

### DIFF
--- a/packages/compiler-cli/src/ngtsc/scope/src/local.ts
+++ b/packages/compiler-cli/src/ngtsc/scope/src/local.ts
@@ -296,9 +296,10 @@ export class LocalModuleScopeRegistry implements MetadataRegistry, ComponentScop
     const exportPipes = new Map<ts.Declaration, PipeMeta>();
 
     // The algorithm is as follows:
-    // 1) Add directives/pipes declared in the NgModule to the compilation scope.
-    // 2) Add all of the directives/pipes from each NgModule imported into the current one to the
-    //    compilation scope. At this point, the compilation scope is complete.
+    // 1) Add all of the directives/pipes from each NgModule imported into the current one to the
+    //    compilation scope.
+    // 2) Add directives/pipes declared in the NgModule to the compilation scope. At this point, the
+    //    compilation scope is complete.
     // 3) For each entry in the NgModule's exports:
     //    a) Attempt to resolve it as an NgModule with its own exported directives/pipes. If it is
     //       one, add them to the export scope of this NgModule.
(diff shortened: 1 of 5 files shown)


AndrewKushnir:
[Presubmit](http://test/OCL:298744012:BASE:298744044:1583289517197:cc94e12b)

alxhub (approved):
Hmmm. I really don't like relying on the orderedness of `Set`, but this will work for now.

AndrewKushnir:
- [Presubmit](http://test/OCL:299720523:BASE:299720539:1583704686840:efd0c3a8)
- [Global Presubmit (Ivy only)](https://test.corp.google.com/ui#id=OCL:299720523:BASE:299720726:1583704838123:3fbbfbea)

AndrewKushnir:
FYI, both Blueprint and Global TAP runs (with Ivy) with the changes from this PR are successful.

pkozlowski-opensource (approved):
LGTM

PR #35856, opened 2020-03-04
build: create angular-internal-dev-infra npm package

josephperrott:


dev-infra/pullapprove/verify.js
```diff
@@ -18,9 +18,8 @@ set('-e');
 
 // Regex Matcher for contains_any_globs conditions
 const CONTAINS_ANY_GLOBS_REGEX = /^'([^']+)',?$/;
-
 // Full path of the angular project directory
-const ANGULAR_PROJECT_DIR = path.resolve(__dirname, '../..');
+const ANGULAR_PROJECT_DIR = process.cwd();
```
gregmagolan: ANGULAR_PROJECT_DIR could just be removed if it depends on `process.cwd()`. It could then also assume there is a `.pullapropve` file in the cwd below.

dev-infra/pullapprove/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "angular-internal-dev-infra-pullapprove",
+  "version": "0.0.0",
+  "bin": {
+    "verify": "./cli.js"
```
gregmagolan: the `.bin` name will end up in `node_modules/.bin/verify` in this case so I think it should be more description such as `verify-pullapprove`

gregmagolan: or would it as this would be a nested package.json 🤔

dev-infra/BUILD.bazel
```diff
@@ -0,0 +1,16 @@
+load("@build_bazel_rules_nodejs//:index.bzl", "pkg_npm")
+
+pkg_npm(
+    name = "npm_package",
+    srcs = [
```
gregmagolan: may be easier to maintain without the nested package and just use
```
srcs = glob(["**")
```

dev-infra/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "angular-internal-dev-infra",
+  "version": "0.0.0",
+  "description": "INTERNAL USE ONLY - Angular internal DevInfra tooling/scripts - INTERNAL USE ONLY",
+  "license": "MIT",
+  "bin": {
+    "ng-dev": "./cli.js"
+  },
+  "devDependencies": {
+    "pullapprove": "file:./pullapprove"
```
gregmagolan: I'd avoid this pattern. Seems like it could lead to yarn cache issues.

Maybe there should only be a root package.json and its "bin" would contain the root `ng-dev` bin as well as any nested folder cli.js bin's that you want to expose such as `"ng-dev-verify-pullapprove" : "./pullapprove/cli.js"`?

dev-infra/pullapprove/verify.js
josephperrott: Agreed, I think it makes sense to do this kind of thing in a follow up as  I was just trying to make the minimal change to this file in this PR, since it is mostly about moving to this new package.

PR #35862, opened 2020-03-04
build: create dev-infra-private npm package

josephperrott:
Creates the scaffolding for an @angular/dev-infra-private package
which will not be published to npm but will be pushed to
https://github.com/angular/dev-infra-private-builds repo for each
commit to master.

The contents of this npm package will then be depended on via
package.json dependency for angular/angular angular/angular-cli and
angular/components.

### DIFF
--- a/.pullapprove.yml
+++ b/.pullapprove.yml
@@ -928,6 +928,7 @@ groups:
           '.github/**',
           '.vscode/**',
           '.yarn/**',
+          'dev-infra/**',
           'docs/BAZEL.md',
           'docs/CARETAKER.md',
           'docs/COMMITTER.md',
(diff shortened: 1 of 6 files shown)


gregmagolan:
From slack:

Integration tests still flaking out.

https://app.circleci.com/jobs/github/angular/angular/643436

Ohhhh. I think I know what it is. Parallel tests are using a shared global yarn cache. (edited) 

That only started to happen when the integration tests were added

This is the giveaway:
```
error An unexpected error occurred: "ENOTEMPTY: directory not empty, rmdir '/home/circleci/.cache/yarn/v6/npm-rxjs-6.5.4-e0777fe0d184cec7872df147f303572d414e211c-integrity/node_modules/rxjs/add'".
```

yarn cache is not mutex protected unless you ask it to be with the --mutex network flag which we add in yarn_install

So options are:

1. set --mutex network for yarn install in all integration tests
2. use local yarn cache for all integration tests

#1 is probably more efficient

devversion (approved):
LGTM

IgorMinar (approved):
lgtm - global approval

PR #35864, opened 2020-03-05
build: typescript 3.8 support

alxhub:
This commit adds support in the Angular monorepo and in the Angular
compiler(s) for TypeScript 3.8. All packages can now compile with
TS 3.8.

For most of the repo, only a handful few typings adjustments were needed.

For compiler-cli and schematics (which use the TypeScript AST) a major
breaking change was the introduction of the export form:

```typescript
export * as foo from 'bar';
```

This is a `ts.NamespaceExport`, and the `exportClause` of a
`ts.ExportDeclaration` can now take this type as well as `ts.NamedExports`.
This broke a lot of places where `exportClause` was assumed to be
`ts.NamedExports`.

For the most part these breakages were in cases where it is not necessary
to handle the new `ts.NamedExports` anyway. ngtsc's design uses the
`ts.TypeChecker` APIs to understand syntax and so automatically supports the
new form of exports.

The View Engine compiler on the other hand extracts TS structures into
metadata.json files, and that format was not designed for namespaced
exports. As a result it will take a nontrivial amount of work if we want to
support such exports in View Engine. For now, these new exports are not
accounted for in metadata.json, and so using them in "folded" Angular
expressions will result in errors (probably claiming that the referenced
exported namespace doesn't exist).

Care was taken to only use TS APIs which are present in 3.7/3.6, as Angular
needs to remain compatible with these for the time being.

This commit does not update angular.io.


ayazhafiz (approved):
lgtm for language service

packages/elements/src/create-custom-element.ts
```diff
@@ -31,7 +31,7 @@ export interface NgElementConstructor<P> {
    * Initializes a constructor instance.
    * @param injector The source component's injector.
    */
-  new (injector: Injector): NgElement&WithProperties<P>;
+  new (injector?: Injector): NgElement&WithProperties<P>;
```
AndrewKushnir: Quick question: why do we need to mark `injector` argument as optional?

integration/typings_test_ts38/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "angular-integration",
+  "description": "Assert that users with TypeScript 3.7 can type-check an Angular application",
```
IgorMinar:
```suggestion
  "description": "Assert that users with TypeScript 3.8 can type-check an Angular application",
```

package.json
```diff
@@ -132,10 +132,10 @@
     "sourcemap-codec": "^1.4.8",
     "systemjs": "0.18.10",
     "terser": "^4.4.0",
-    "tsickle": "0.38.0",
+    "tsickle": "^0.38.1",
```
IgorMinar:
```suggestion
    "tsickle": "0.38.1",
```
we don't want to update tsickle by accident/unintentionally.

integration/typings_test_ts38/package.json
JoostK:
```suggestion
  "description": "Assert that users with TypeScript 3.8 can type-check an Angular application",
```

packages/elements/src/create-custom-element.ts
IgorMinar: why this change now?

if this is the right change, then we should also document the behavior when injector is not provided

package.json
```diff
@@ -132,10 +132,10 @@
...
     "tslib": "^1.10.0",
     "tslint": "6.0.0",
-    "typescript": "~3.7.4",
+    "typescript": "3.8",
```
IgorMinar:
```suggestion
    "typescript": "~3.8.3",
```

packages/language-service/test/diagnostics_spec.ts
```diff
@@ -636,8 +636,7 @@ describe('diagnostics', () => {
     const tsDiags = tsLS.getSemanticDiagnostics(APP_COMPONENT);
     expect(tsDiags.length).toBe(1);
     const msgText = ts.flattenDiagnosticMessageText(tsDiags[0].messageText, '\n');
-    expect(msgText).toBe(
-        `Type 'null[]' is not assignable to type 'Provider[]'.\n  Type 'null' is not assignable to type 'Provider'.`);
+    expect(msgText).toBe(`Type 'null' is not assignable to type 'Provider'.`);
```
IgorMinar: the commit message doesn't have any description of this change - why did we change the test in this way?

IgorMinar: also this is not `fix` but `test` commit type

integration/language_service_plugin/matcher.ts
```diff
@@ -5,15 +5,16 @@ const goldens: string[] = process.argv.slice(2);
 export const goldenMatcher: jasmine.CustomMatcherFactories = {
   toMatchGolden(util: jasmine.MatchersUtil): jasmine.CustomMatcher {
     return {
-      compare(actual: {command: string}, golden: string): jasmine.CustomMatcherResult {
+      compare(actual: {body?: {}}, golden: string): jasmine.CustomMatcherResult {
```
IgorMinar: this seems to be an actual code change and not a test change + we are missing a description of the change - please always try to answer the question "why this change?" in the commit message body.

IgorMinar (approved):
lgtm with nits - global approval once comments are resolved.

JoostK (approved):
LGTM, minor typo

integration/language_service_plugin/matcher.ts
alxhub: This is in an integration test, so it's a test change. My guess is that the object/value returned by the TS Language Service has changed in 3.8, so the Jasmine matcher here is being updated to reflect that change.

packages/language-service/test/diagnostics_spec.ts
alxhub: The commit messages themselves are unimportant as they're getting squashed into my `build` commit.

packages/elements/src/create-custom-element.ts
alxhub: In TS 3.7 `CustomElements.define()` accepts a `Function` for the element constructor, but 3.8 declares a new type:

```typescript
interface CustomElementConstructor {
    new (): HTMLElement;
}
```

This enforces that the custom element constructor has zero arguments. `NgElementConstructor`'s implementation has always supported being called with zero arguments, it was just never expressed in the type signature.

integration/typings_test_ts38/package.json
alxhub: Nice catch!

packages/elements/src/create-custom-element.ts
alxhub: See my comment to @AndrewKushnir above.

AndrewKushnir: Thanks for explaining this Alex!

alxhub:
Caretaker, per @IgorMinar's comment above, this has global approval.

I've squashed @kyliau's commits into mine, and added commit message notes about the language service changes.

alxhub:
[Presubmit](http://test/OCL:299899605:BASE:300155212:1583869207227:a819d0de)

PR #35870, opened 2020-03-05
feat(service-worker): change default SW strategy behavior

### LINKED ISSUE #34464: feat(service-worker): improve default `SwRegistrationStrategy`
**The problem**

In order to avoid negatively affecting the initial page-load time, `@angular/service-worker` will (by default) wait for the [app to stabilize][1], before registering the SW  script. This can lead to a situation where the SW will never be registered when there is a long-running task (such as an interval or recurring timeout), which can often be started by a 3rd-party dependency (without the developer realizing). Related issues: #20970, #34461

This can easily be missed by the developer and there is no other indication other than the SW is not in effect. This is particularly hard to detect, because the SW is typically not used during development and on production builds a previous SW instance might be in effect.

##
**Possible solutions**

Since 4cfba5807, it is possible to overwrite this default behavior by [providing a different `SwRegistrationStrategy`][2]. Things we could do to remedy the situation (non-exhaustive list; up for discussion):

- Change the default strategy to a custom one that wait for the app to stabilize, but registers anyway after X seconds (no matter whether stabilized or not).

- Keep the current behavior, but print a warning after X seconds that the app has not stabilized yet and the SW has not been registered. (This won't change the current behavior, but will make it easier for developers to identify the problem.)

[1]: https://angular.io/api/core/ApplicationRef#is-stable-examples
[2]: https://angular.io/api/service-worker/SwRegistrationOptions#registrationStrategy

sonukapoor:
Previously, there was a chance that the SW will never register
when there is a long-running task or recurring timeout. This commit
fixes this, by first trying to wait for the app to stabilize, however,
if that does not happen, then the SW will register after 30 seconds.

Closes #34464

## PR Checklist

## PR Type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #34464

- [ ] Yes
- [X] No

### DIFF
--- a/packages/service-worker/src/module.ts
+++ b/packages/service-worker/src/module.ts
@@ -7,7 +7,7 @@
  */
 
 import {isPlatformBrowser} from '@angular/common';
-import {APP_INITIALIZER, ApplicationRef, InjectionToken, Injector, ModuleWithProviders, NgModule, PLATFORM_ID} from '@angular/core';
+import {APP_INITIALIZER, ApplicationRef, InjectionToken, Injector, ModuleWithProviders, NgModule, NgZone, PLATFORM_ID} from '@angular/core';
 import {Observable, merge, of } from 'rxjs';
 import {delay, filter, take} from 'rxjs/operators';
 
@@ -108,11 +108,14 @@ export function ngswAppInitializer(
           readyToRegister$ = of (null);
           break;
         case 'registerWithDelay':
-          readyToRegister$ = delayWithTimeout(+args[0] || 0);
+          readyToRegister$ =
+              injector.get<NgZone>(NgZone).runOutsideAngular(() => delayWithTimeout(+args[0] || 0));
           break;
         case 'registerWhenStable':
-          readyToRegister$ = !args[0] ? whenStable(injector) :
-                                        merge(whenStable(injector), delayWithTimeout(+args[0]));
+          readyToRegister$ = !args[0] ?
+              whenStable(injector) :
+              injector.get<NgZone>(NgZone).runOutsideAngular(
+                  () => merge(whenStable(injector), delayWithTimeout(+args[0])));
           break;
         default:
           // Unknown strategy.
--- a/packages/service-worker/src/module.ts
+++ b/packages/service-worker/src/module.ts
@@ -108,14 +108,11 @@ export function ngswAppInitializer(
           readyToRegister$ = of (null);
           break;
         case 'registerWithDelay':
-          readyToRegister$ =
-              injector.get<NgZone>(NgZone).runOutsideAngular(() => delayWithTimeout(+args[0] || 0));
+          readyToRegister$ = delayWithTimeout(+args[0] || 0);
           break;
         case 'registerWhenStable':
-          readyToRegister$ = !args[0] ?
-              whenStable(injector) :
-              injector.get<NgZone>(NgZone).runOutsideAngular(
-                  () => merge(whenStable(injector), delayWithTimeout(+args[0])));
+          readyToRegister$ = !args[0] ? whenStable(injector) :
+                                        merge(whenStable(injector), delayWithTimeout(+args[0]));
           break;
         default:
           // Unknown strategy.
@@ -125,10 +122,15 @@ export function ngswAppInitializer(
     }
 
     // Don't return anything to avoid blocking the application until the SW is registered.
+    // Also, run outside the Angular zone to avoid preventing the app from stabilizing (especially
+    // given that some registration strategies wait for the app to stabilize).
     // Catch and log the error if SW registration fails to avoid uncaught rejection warning.
-    readyToRegister$.pipe(take(1)).subscribe(
-        () => navigator.serviceWorker.register(script, {scope: options.scope})
-                  .catch(err => console.error('Service worker registration failed with:', err)));
+    const ngZone = injector.get(NgZone);
+    ngZone.runOutsideAngular(
+        () => readyToRegister$.pipe(take(1)).subscribe(
+            () =>
+                navigator.serviceWorker.register(script, {scope: options.scope})
+                    .catch(err => console.error('Service worker registration failed with:', err))));
   };
   return initializer;
 }
--- a/packages/service-worker/src/module.ts
+++ b/packages/service-worker/src/module.ts
@@ -100,7 +100,9 @@ export function ngswAppInitializer(
     if (typeof options.registrationStrategy === 'function') {
       readyToRegister$ = options.registrationStrategy();
     } else {
-      const [strategy, ...args] = (options.registrationStrategy || 'registerWhenStable:30000').split(':');
+      const [strategy, ...args] =
+          (options.registrationStrategy || 'registerWhenStable:30000').split(':');
+
       switch (strategy) {
         case 'registerImmediately':
           readyToRegister$ = of (null);
--- a/packages/service-worker/src/module.ts
+++ b/packages/service-worker/src/module.ts
@@ -8,7 +8,7 @@
 
 import {isPlatformBrowser} from '@angular/common';
 import {APP_INITIALIZER, ApplicationRef, InjectionToken, Injector, ModuleWithProviders, NgModule, PLATFORM_ID} from '@angular/core';
-import {Observable, of } from 'rxjs';
+import {Observable, merge, of } from 'rxjs';
 import {delay, filter, take} from 'rxjs/operators';
 
 import {NgswCommChannel} from './low_level';
@@ -57,6 +57,9 @@ export abstract class SwRegistrationOptions {
    *
    * - `registerWhenStable`: Register as soon as the application stabilizes (no pending
    *      micro-/macro-tasks).
+   * - `registerWhenStable:<timeout>`: Register when the app becomes stable or with a delay
+   *     of `<timeout>` milliseconds. This will ensure that the ServiceWorker will register
+   *     after the specified timeout, regardless of if the app is stable or not.
    * - `registerImmediately`: Register immediately.
    * - `registerWithDelay:<timeout>`: Register with a delay of `<timeout>` milliseconds. For
    *     example, use `registerWithDelay:5000` to register the ServiceWorker after 5 seconds. If
@@ -102,11 +105,10 @@ export function ngswAppInitializer(
           readyToRegister$ = of (null);
           break;
         case 'registerWithDelay':
-          readyToRegister$ = of (null).pipe(delay(+args[0] || 0));
+          readyToRegister$ = delayWithTimeout(+args[0]);
           break;
         case 'registerWhenStable':
-          const appRef = injector.get<ApplicationRef>(ApplicationRef);
-          readyToRegister$ = appRef.isStable.pipe(filter(stable => stable));
+          readyToRegister$ = merge(whenStable(injector), delayWithTimeout(+args[0]));
           break;
         default:
           // Unknown strategy.
@@ -122,6 +124,15 @@ export function ngswAppInitializer(
                   .catch(err => console.error('Service worker registration failed with:', err)));
   };
   return initializer;
+}
+
+function delayWithTimeout(timeout: number) {
+  return of (null).pipe(delay(timeout || 0));
+}
+
+function whenStable(injector: Injector) {
+  const appRef = injector.get<ApplicationRef>(ApplicationRef);
+  return appRef.isStable.pipe(filter(stable => stable));
 }
 
 export function ngswCommChannelFactory(
--- a/packages/service-worker/src/module.ts
+++ b/packages/service-worker/src/module.ts
@@ -55,11 +55,12 @@ export abstract class SwRegistrationOptions {
    * registered (e.g. there might be a long-running timeout or polling interval, preventing the app
    * to stabilize). The available option are:
    *
-   * - `registerWhenStable`: Register as soon as the application stabilizes (no pending
-   *      micro-/macro-tasks).
-   * - `registerWhenStable:<timeout>`: Register when the app becomes stable or with a delay
-   *     of `<timeout>` milliseconds. This will ensure that the ServiceWorker will register
-   *     after the specified timeout, regardless of if the app is stable or not.
+   * - `registerWhenStable:<timeout>`: Register as soon as the application stabilizes (no pending
+   *     micro-/macro-tasks) but no later than `<timeout>` milliseconds. If the app hasn't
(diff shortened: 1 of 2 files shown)


sonukapoor:
@alxhub / @gkalpak My local lint and tests are all passing - not sure why CI is complaining. Can you help with this?

gkalpak:
Thx, @sonukapoor, for taking this on. Can you please remove the formatting changes? They make the PR hard to review and the formatter is unhappy (you are probably running the wrong commands locally if `lint` passes).

Also, how are you running the tests locally? The failures on CI seem expected.

sonukapoor:
@gkalpak. I am running yarn lint and yarn test in the aio folder.

packages/service-worker/src/module.ts
```diff
@@ -105,9 +105,11 @@ export function ngswAppInitializer(
           readyToRegister$ = of (null).pipe(delay(+args[0] || 0));
           break;
         case 'registerWhenStable':
-          const appRef = injector.get<ApplicationRef>(ApplicationRef);
-          readyToRegister$ = appRef.isStable.pipe(filter(stable => stable));
-          break;
+          // Our default is when the app is stable, or after 30 seconds
```
gkalpak: This is not the place to document the default behavior. This branch is about `registerWhenStable`. (The fact that we use `registerWhenStable` as the default value is done in a [different part of the code][1].)

[1]: https://github.com/angular/angular/blob/f714a54f018277c1113cbac11f5c4735e9e08da6/packages/service-worker/src/module.ts#L99

packages/service-worker/src/module.ts
```diff
@@ -105,9 +105,11 @@ export function ngswAppInitializer(
...
+          // This ensures that the SW is still registered, in the case of a long running 
+          // or recurring timeout. 
+          readyToRegister$ = merge(whenStable(injector), of(null).pipe(delay(+args[0] || 30000)));
```
gkalpak: Thinking about it again, I don't think we should change the behavior of `registerWhenStable` (in case it is used explicitly by someone), only the default value. It should also be possible for people to wait for stability indefinitely (if they so wish).

Therefore, I suggest the following changes:
- When not passing an argument (e.g. `'registerWhenStable'`), keep the current behavior (i.e. wait for stability indefinitely).
- When passing a timeout argument (e.g. `'registerWhenStable:10000'`), wait for stability or for the timeout to expire.
- Change the [default strategy][1] from `'registerWhenStable'` to `'registerWhenStable:30000'` (timeout up for debate).

[1]: https://github.com/angular/angular/blob/f714a54f018277c1113cbac11f5c4735e9e08da6/packages/service-worker/src/module.ts#L99

gkalpak: Hm...I just realized: Does this `delay` prevent the app from stabilizing? If that's the case (which it probably is), we should run it outside the `NgZone`. Same for other `registerWithDelay` strategy.

gkalpak (commented):
Here are some more comments from a preliminary high-level review.

gkalpak:
> @gkalpak. I am running yarn lint and yarn test in the aio folder. When I try to run similar commands within root then I get all kind of strange errors from bazel

These changes are outside the `aio/` directory, so you need to run commands at the root of the repo :grin:
Can you open a separate issue with the errors you are getting from bazel?

packages/service-worker/src/module.ts
sonukapoor: @gkalpak Good catch 👍

sonukapoor:
> Can you open a separate issue with the errors you are getting from bazel?

@gkalpak, Please see here: #35891

packages/service-worker/src/module.ts
```diff
@@ -7,8 +7,8 @@
  */
 
 import {isPlatformBrowser} from '@angular/common';
-import {APP_INITIALIZER, ApplicationRef, InjectionToken, Injector, ModuleWithProviders, NgModule, PLATFORM_ID} from '@angular/core';
-import {Observable, of } from 'rxjs';
+import {APP_INITIALIZER, ApplicationRef, InjectionToken, Injector, ModuleWithProviders, NgModule, PLATFORM_ID, NgZone} from '@angular/core';
```
gkalpak: Please order imported symbols.

packages/service-worker/src/module.ts
```diff
@@ -96,17 +97,27 @@ export function ngswAppInitializer(
     if (typeof options.registrationStrategy === 'function') {
       readyToRegister$ = options.registrationStrategy();
     } else {
-      const [strategy, ...args] = (options.registrationStrategy || 'registerWhenStable').split(':');
+      // Our default is when the app is stable, or after 30 seconds
+      // This ensures that the SW is still registered, in the case of a long running 
+      // or recurring timeout. 
+      const [strategy, ...args] = (options.registrationStrategy || 'registerWhenStable:30000').split(':');
```
gkalpak: Move this change to a separate commit.

packages/service-worker/src/module.ts
```diff
@@ -96,17 +97,27 @@ export function ngswAppInitializer(
...
       switch (strategy) {
         case 'registerImmediately':
           readyToRegister$ = of (null);
           break;
         case 'registerWithDelay':
-          readyToRegister$ = of (null).pipe(delay(+args[0] || 0));
+          ngZone.runOutsideAngular(() => {
+            readyToRegister$ = of (null).pipe(delay(+args[0] || 0));
+          });
```
gkalpak: Have you verified that:
a. This prevents the app from stabilizng without `runOutsideangular`?
b. The app can stabilize with `runOutsideAngular` here (vs where the observable is subscribed to)?

We also need some test(s) to verify the above.

packages/service-worker/src/module.ts
```diff
@@ -96,17 +97,27 @@ export function ngswAppInitializer(
...
         case 'registerWhenStable':
-          const appRef = injector.get<ApplicationRef>(ApplicationRef);
-          readyToRegister$ = appRef.isStable.pipe(filter(stable => stable));
+            ngZone.runOutsideAngular(() => {
+              if (+args[0]) {
+                readyToRegister$ = merge(whenStable(injector), of(null).pipe(delay(+args[0])));
+              } else {
+                readyToRegister$ = whenStable(injector);
+              }
+            });
```
gkalpak:
```suggestion
          ngZone.runOutsideAngular(() => {
            if (+args[0]) {
              readyToRegister$ = merge(whenStable(injector), of(null).pipe(delay(+args[0])));
            } else {
              readyToRegister$ = whenStable(injector);
            }
          });
```

packages/service-worker/src/module.ts
```diff
@@ -75,7 +75,8 @@ export const SCRIPT = new InjectionToken<string>('NGSW_REGISTER_SCRIPT');
 
 export function ngswAppInitializer(
     injector: Injector, script: string, options: SwRegistrationOptions,
-    platformId: string): Function {
+    platformId: string, 
+    ngZone: NgZone): Function {
```
gkalpak: In order to avoid unnecessary work, I suggest we use the `injector` to get `NgZone` only when/if needed.

packages/service-worker/src/module.ts
```diff
@@ -96,17 +97,27 @@ export function ngswAppInitializer(
...
-          readyToRegister$ = of (null).pipe(delay(+args[0] || 0));
+          ngZone.runOutsideAngular(() => {
+            readyToRegister$ = of (null).pipe(delay(+args[0] || 0));
+          });
           break;
         case 'registerWhenStable':
-          const appRef = injector.get<ApplicationRef>(ApplicationRef);
-          readyToRegister$ = appRef.isStable.pipe(filter(stable => stable));
+            ngZone.runOutsideAngular(() => {
+              if (+args[0]) {
```
gkalpak: This will treat 0 the same as not providing an argument. I think this is not correct. Explicitly specifying `0` as an argument should be respected.

packages/service-worker/src/module.ts
gkalpak: Or rather:

```suggestion
          readyToRegister$ = (args[0] === undefined) ?
              whenStable(injector) :
              ngZone.runOutsideAngular(() => merge(whenStable(injector), of(null).pipe(delay(+args[0]))));
```

But, again, we should verify whether `runOutsideAngular()` is needed here or when subscribing to the observable.

packages/service-worker/src/module.ts
```diff
@@ -117,13 +128,18 @@ export function ngswAppInitializer(
 
     // Don't return anything to avoid blocking the application until the SW is registered.
     // Catch and log the error if SW registration fails to avoid uncaught rejection warning.
-    readyToRegister$.pipe(take(1)).subscribe(
+    readyToRegister$!.pipe(take(1)).subscribe(
```
gkalpak: This should not be needed with my suggested refactoring.

packages/service-worker/src/module.ts
```diff
@@ -96,17 +97,27 @@ export function ngswAppInitializer(
...
+      // This ensures that the SW is still registered, in the case of a long running 
+      // or recurring timeout. 
+      const [strategy, ...args] = (options.registrationStrategy || 'registerWhenStable:30000').split(':');
       switch (strategy) {
         case 'registerImmediately':
           readyToRegister$ = of (null);
           break;
         case 'registerWithDelay':
-          readyToRegister$ = of (null).pipe(delay(+args[0] || 0));
+          ngZone.runOutsideAngular(() => {
```
gkalpak: Please move this change to a separate commit.

packages/service-worker/src/module.ts
```diff
@@ -96,17 +97,27 @@ export function ngswAppInitializer(
...
+            readyToRegister$ = of (null).pipe(delay(+args[0] || 0));
+          });
           break;
         case 'registerWhenStable':
-          const appRef = injector.get<ApplicationRef>(ApplicationRef);
-          readyToRegister$ = appRef.isStable.pipe(filter(stable => stable));
+            ngZone.runOutsideAngular(() => {
+              if (+args[0]) {
+                readyToRegister$ = merge(whenStable(injector), of(null).pipe(delay(+args[0])));
```
gkalpak: Might be worth creating a helper for these `of(null).pipe(delay(+args[0]))`.

packages/service-worker/test/module_spec.ts
```diff
@@ -177,6 +177,21 @@ describe('ServiceWorkerModule', () => {
            expect(swRegisterSpy).toHaveBeenCalledWith('sw.js', {scope: undefined});
          }));
 
+        it('registers the SW when the app doesnt stabilizes with `registerWhenStable`, but uses the timeout', fakeAsync(() => {
```
gkalpak:
```suggestion
        it('registers the SW when the app does not stabilize after `<timeout>` with `registerWhenStable:<timeout>`', fakeAsync(() => {
```

gkalpak (changes requested):

gkalpak:
BTW, @sonukapoor, you don't need to run all the tests when you only update a specific package. For this PR for example, you could run `yarn bazel test //packages/service-worker/...` or `yarn bazel test //packages/service-worker/test`.
(Replace `bazel` with `ibazel` in the above commands to run in "watch  mode".)

packages/service-worker/src/module.ts
sonukapoor: I was going to test all of this using unit tests but since I can’t run bazel I am stuck on it.

sonukapoor:
Thanks. I will do that once I get bazel working. Until then it’s very difficult to run any tests. I will make the code changes for now and test everything once the bazel issue is resolved.

packages/service-worker/src/module.ts
sonukapoor: Good point

packages/service-worker/src/module.ts
sonukapoor: I tried that, but typescript is still complaining about the error `Variable 'readyToRegister$' is used before being assigned.ts(2454)`

packages/service-worker/src/module.ts
sonukapoor: @gkalpak, you are right. The app will never stabilize. Here is the documentation:

https://angular.io/api/core/ApplicationRef#isstable-examples-and-caveats

> the application will never be stable if you start any kind of recurrent asynchronous task when the application starts (for example for a polling process, started with a setInterval, a setTimeout or using RxJS operators like interval);

By wrapping the code block in the `runOutsideAngular` function, we should be fine.

sonukapoor:
@gkalpak, I think I have addressed all of your comments. However, I am not able to run any tests, because of the previously mentioned errors with bazel. I tried the single tests but get the same errors. Let me know if the code/commits are otherwise fine.

sonukapoor:
@gkalpak, something is wrong with CI, my PWA score locally with the same arguments is at passing:

image

sonukapoor:
@gkalpak I borrowed a mac for this PR. It seems that the tests are failing because of the timeout that we have set on the service worker. It does not complete in time with the timeout set in protractor/jasmine. It seems to be set to `10secs`. I have tried to set that but couldn't figure it out. I tried setting it directly in the `it`s third function argument or using `jasmine.DEFAULT_TIMEOUT_INTERVAL = 31000`, but nothing worked. I then simply reduced the time in the service worker to 8 secs instead of 30. Now the tests seem to pass fine - see [here](https://app.circleci.com/pipelines/github/angular/angular/11854/workflows/dd226ce1-5290-41d1-b628-d15e6084491a/jobs/658504). However, the lighthouse audits are failing now. They are passing for me on windows and mac. I am not sure why `CI` is complaining about this.

packages/service-worker/src/module.ts
gkalpak: It turns out that we need to `runOutsideAngular()` when subscribing to the observable (because that is where the timeouts are created). Creating the observable (but not subscribing to it) outside the zone does nothing to allow the app to stabilize :grin:

I will move the `runOutsideAngular()` call to where we subscribe.

gkalpak:
@googlebot I consent.

gkalpak:
Rebased on master, updated commit messages and fixed up commits. Let's see what CI thinks :smiley:
@sonukapoor, please take a look and make sure I didn't miss anything 🙏

sonukapoor:
> Rebased on master, updated commit messages and fixed up commits. Let's see what CI thinks 😃
> @sonukapoor, please take a look and make sure I didn't miss anything 🙏

Looks good to me. Thanks for the updates - very much appreciated.

gkalpak (approved):
Thx, @sonukapoor ✨

PR #31443, opened 2019-07-07
fix(zone.js): fix #27840, don't wrap uncaught promise error.

JiaLiPassion:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] angular.io application / infrastructure changes
- [ ] Other... Please describe:

## What is the current behavior?

Issue Number: #27840 

`zone.js` will catch `uncaught promise rejection`, such as this case.

```
Promise.reject(new Error('test'));
```

And when `zone onHandleError` catch the error, the error will look like

```
{
  message: 'Uncaught (in promise): test',
  reason: Error('test')
}
```

because `zone.js` wrap the original error into a new Error with the reason property equals to the original error. It will make developer hard to use, because developer expect the original error will be caught in error handler.

## What is the new behavior?
If the `uncaught Promise rejection` is `instanceof Error or has a message property`. 
To keep the backward compatibility, I also set the `reason` property to the `error`.

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

PR #35873, opened 2020-03-05
feat(zone.js): add a zone config to allow user disable wrapping uncaught promise rejection

### LINKED ISSUE #27840: new behavior for global ErrorHandler and uncaught errors from promises
# 🐞 bug report

### Affected Package

 Well I thought this was zone.js wrapping the errors, but I rolled zone back only zone and it didn't change.

### Is this a regression?

 Yes, the previous version in which this bug was not present was 5.x

### Description

Previously, unhandled errors from promises were handed to a custom error handler unchanged.  But now, it seems the error is wrapped and you need to unwrap it to get the real error that's sitting on the "rejection" property of the error.

The docs make no mention of this here https://angular.io/api/core/ErrorHandler and neither does any googling surprisingly.

I noticed this only after upgrading from angular 5 to angular 7 and one of my e2e tests that relied on my general error handler started failing.

The only workaround I found was to have something like this at the top of the error handler's handleError() method like this:

```typescript
    while(error.rejection) {
      error = error.rejection;
    }

```
but...

1. I'm quite sure there's a better way - now I have to type error as any
2. I'm not sure if there are any other situations that will require unwrapping (e.g. from observables?).

Either way, if this was some kind of intentional feature, it's definitely a gotcha that I think should be mentioned (with the RIGHT way to handle it) in the doc link I pasted.

## 🔬 Minimal Reproduction

 

1. follow the docs and implement a global error handler and try to handle unhandled requests uniformly - something like this:

```typescript
@Injectable()
export class SpiffyErrorHandlerService implements ErrorHandler {
  constructor(private injector: Injector, private snackBar: MatSnackBar) {}
  public handleError(error: Error) {
    if (error instanceof HttpErrorResponse) {
      if (error.status === 401) {
        this.snackBar.open(
          'Oops.  It looks like you tried to access a logged-in resource while logged out.  Please log in again.',
          '',
          { duration: 3000 }
        );
    }
  }
}
```

2. inject HttpClient

3. await this.http.get('/some-404-returning-url').toPromise();

## 🔥 Exception or Error

EXPECTED: error handler works, gets the HttpErrorResponse, and shows the snackbar
ACTUAL: the HttpErrorResponse is on error.rejection instead of right on error.

## 🌍  Your Environment

**Angular Version:**
<pre><code>

Angular CLI: 7.1.2
Node: 10.14.2
OS: linux x64
Angular: 7.1.3
... platform-browser-dynamic, router, service-worker

@angular-devkit/architect         0.10.7
@angular-devkit/build-angular     0.10.7
@angular-devkit/build-optimizer   0.10.7
@angular-devkit/build-webpack     0.10.7
@angular-devkit/core              7.0.7
@angular-devkit/schematics        7.1.2
@angular/cdk                      7.1.1
@angular/cli                      7.1.2
@angular/flex-layout              7.0.0-beta.19
@angular/material                 7.1.1
@ngtools/webpack                  7.0.7
@schematics/angular               7.1.2
@schematics/update                0.11.2
rxjs                              6.3.3
typescript                        3.1.6
webpack                           4.19.1
    
</code></pre>

**Anything else relevant?**

JiaLiPassion:
Close #27840.

By default, `zone.js` wrap uncaught promise error and wrap it to a new Error object with some
additional information includes the value of the error and the stack trace.

Consider the following example:

```
Zone.current
  .fork({
    name: 'promise-error',
    onHandleError: (delegate: ZoneDelegate, current: Zone, target: Zone, error: any): boolean => {
      console.log('caught an error', error);
      delegate.handleError(target, error);
      return false;
    }
}).run(() => {
  const originalError = new Error('testError');
  Promise.reject(originalError);
});
```

The `promise-error` zone catches a wrapped `Error` object whose `reject` property equals
to the `originalError`, and the message will be `Uncaught (in promise): testError...`.
You can disable this wrapping behavior by defining a global configuraiton
` __zone_symbol__DISABLE_WRAPPING_UNCAUGHT_PROMISE_REJECTION = true;` before importing `zone.js`.

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #28740

This PR was original submitted here https://github.com/angular/angular/pull/31443, and it was reverted as it is a breaking change, and in this new PR, I changed the logic and keep the original behavior and use a configuration to let user change the behavior.

### DIFF
--- a/packages/zone.js/lib/common/promise.ts
+++ b/packages/zone.js/lib/common/promise.ts
@@ -20,6 +20,8 @@ Zone.__load_patch('ZoneAwarePromise', (global: any, Zone: ZoneType, api: _ZonePr
 
   const __symbol__ = api.symbol;
   const _uncaughtPromiseErrors: UncaughtPromiseError[] = [];
+  const isDisableWrappingUncaughtPromiseRejection =
+      global[__symbol__('DISABLE_WRAPPING_UNCAUGHT_PROMISE_REJECTION')] === true;
   const symbolPromise = __symbol__('Promise');
   const symbolThen = __symbol__('then');
   const creationTrace = '__creationTrace__';
@@ -41,13 +43,11 @@ Zone.__load_patch('ZoneAwarePromise', (global: any, Zone: ZoneType, api: _ZonePr
 
   api.microtaskDrainDone = () => {
     while (_uncaughtPromiseErrors.length) {
-      while (_uncaughtPromiseErrors.length) {
-        const uncaughtPromiseError: UncaughtPromiseError = _uncaughtPromiseErrors.shift() !;
-        try {
-          uncaughtPromiseError.zone.runGuarded(() => { throw uncaughtPromiseError; });
-        } catch (error) {
-          handleUnhandledRejection(error);
-        }
+      const uncaughtPromiseError: UncaughtPromiseError = _uncaughtPromiseErrors.shift() !;
+      try {
+        uncaughtPromiseError.zone.runGuarded(() => { throw uncaughtPromiseError; });
+      } catch (error) {
+        handleUnhandledRejection(error);
       }
     }
   };
@@ -58,7 +58,7 @@ Zone.__load_patch('ZoneAwarePromise', (global: any, Zone: ZoneType, api: _ZonePr
     api.onUnhandledError(e);
     try {
       const handler = (Zone as any)[UNHANDLED_PROMISE_REJECTION_HANDLER_SYMBOL];
-      if (handler && typeof handler === 'function') {
+      if (typeof handler === 'function') {
         handler.call(this, e);
       }
     } catch (err) {
@@ -176,20 +176,28 @@ Zone.__load_patch('ZoneAwarePromise', (global: any, Zone: ZoneType, api: _ZonePr
         }
         if (queue.length == 0 && state == REJECTED) {
           (promise as any)[symbolState] = REJECTED_NO_CATCH;
-          try {
-            // try to print more readable error log
-            throw new Error(
-                'Uncaught (in promise): ' + readableObjectToString(value) +
-                (value && value.stack ? '\n' + value.stack : ''));
-          } catch (err) {
-            const error: UncaughtPromiseError = err;
-            error.rejection = value;
-            error.promise = promise;
-            error.zone = Zone.current;
-            error.task = Zone.currentTask !;
-            _uncaughtPromiseErrors.push(error);
-            api.scheduleMicroTask();  // to make sure that it is running
+          let uncaughtPromiseError = value;
+          if (!isDisableWrappingUncaughtPromiseRejection) {
+            // If disable wrapping uncaught promise reject
+            // and the rejected value is an Error object,
+            // use the value instead of wrapping it.
+            try {
+              // Here we throws a new Error to print more readable error log
+              // and if the value is not an error, zone.js builds an `Error`
+              // Object here to attach the stack information.
+              throw new Error(
+                  'Uncaught (in promise): ' + readableObjectToString(value) +
+                  (value && value.stack ? '\n' + value.stack : ''));
+            } catch (err) {
+              uncaughtPromiseError = err;
+            }
           }
+          uncaughtPromiseError.rejection = value;
+          uncaughtPromiseError.promise = promise;
+          uncaughtPromiseError.zone = Zone.current;
+          uncaughtPromiseError.task = Zone.currentTask !;
+          _uncaughtPromiseErrors.push(uncaughtPromiseError);
+          api.scheduleMicroTask();  // to make sure that it is running
         }
       }
     }
--- a/packages/zone.js/lib/zone.configurations.api.ts
+++ b/packages/zone.js/lib/zone.configurations.api.ts
@@ -529,6 +529,17 @@ interface ZoneGlobalConfigurations {
    * The preceding code makes all scroll event listeners passive.
    */
   __zone_symbol__PASSIVE_EVENTS?: boolean;
+
+  /**
+   * Disable wrapping uncaught promise rejection.
+   *
+   * By default, `zone.js` wraps the uncaught promise rejection in a new `Error` object
+   * which contains additional information such as a value of the rejection and a stack trace.
+   *
+   * If you set `__zone_symbol__DISABLE_WRAPPING_UNCAUGHT_PROMISE_REJECTION = true;` before
+   * importing `zone.js`, `zone.js` will not wrap the uncaught promise rejection.
+   */
+  __zone_symbol__DISABLE_WRAPPING_UNCAUGHT_PROMISE_REJECTION?: boolean;
 }
 
 /**
--- a/packages/zone.js/test/common/promise-disable-wrap-uncaught-promise-rejection.spec.ts
+++ b/packages/zone.js/test/common/promise-disable-wrap-uncaught-promise-rejection.spec.ts
@@ -1,0 +1,79 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+class TestRejection {
+  prop1?: string;
+  prop2?: string;
+}
+
+describe('disable wrap uncaught promise rejection', () => {
+  it('should notify Zone.onHandleError if promise is uncaught', (done) => {
+    let promiseError: Error|null = null;
+    let zone: Zone|null = null;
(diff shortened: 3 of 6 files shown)


integration/_payload-limits.json
```diff
@@ -3,8 +3,8 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 1485,
-        "main-es2015": 141569,
-        "polyfills-es2015": 36657
+        "main-es2015": 141943,
+        "polyfills-es2015": 37081
```
mhevery: Why such large changes in size, given that code changes are relatively minor.

packages/zone.js/lib/common/promise.ts
```diff
@@ -176,18 +178,30 @@ Zone.__load_patch('ZoneAwarePromise', (global: any, Zone: ZoneType, api: _ZonePr
...
-                (value && value.stack ? '\n' + value.stack : ''));
-          } catch (err) {
-            const error: UncaughtPromiseError = err;
-            error.rejection = value;
-            error.promise = promise;
-            error.zone = Zone.current;
-            error.task = Zone.currentTask !;
-            _uncaughtPromiseErrors.push(error);
+          if (queue.length == 0 && state == REJECTED) {
+            (promise as any)[symbolState] = REJECTED_NO_CATCH;
```
mhevery: The above two lines are same as the two lines above that. Is that intentional?

packages/zone.js/lib/common/promise.ts
```diff
@@ -176,18 +178,30 @@ Zone.__load_patch('ZoneAwarePromise', (global: any, Zone: ZoneType, api: _ZonePr
...
+              // use the value instead of wrapping it.
+              uncaughtPromiseError = value;
+            } else {
+              try {
+                // try to print more readable error log
+                throw new Error(
+                    'Uncaught (in promise): ' + readableObjectToString(value) +
+                    (value && value.stack ? '\n' + value.stack : ''));
+              } catch (err) {
+                uncaughtPromiseError = err;
```
mhevery: What is the benefit of throwing and than catching the error? If there is a reason, can you document it please so others are not confused like me.

packages/zone.js/lib/zone.configurations.api.ts
```diff
@@ -530,6 +530,17 @@ interface ZoneGlobalConfigurations {
    * to let all `scroll` event listeners passive.
    */
   __zone_symbol__PASSIVE_EVENTS?: boolean;
+
+  /**
+   * Disable wrapping uncaught promise rejection.
+   *
+   * By default, `zone.js` wrap the uncaught promise rejection to a new `Error` object
```
mhevery:
```suggestion
   * By default, `zone.js` wraps the uncaught promise rejection in a new `Error` object
```

packages/zone.js/lib/zone.configurations.api.ts
```diff
@@ -530,6 +530,17 @@ interface ZoneGlobalConfigurations {
...
+   * with additional information, the value of the rejection and the stack trace.
```
mhevery:
```suggestion
   * which contains additional information such as a value of the rejection and a stack trace.
```

packages/zone.js/lib/zone.configurations.api.ts
```diff
@@ -530,6 +530,17 @@ interface ZoneGlobalConfigurations {
...
+   *
+   * If you set `__zone_symbol__DISABLE_WRAPPING_UNCAUGHT_PROMISE_REJECTION = true;` before
+   * importing `zone.js`, `zone.js` does not wrap the uncaught promise rejection.
```
mhevery:
```suggestion
   * importing `zone.js`, `zone.js` will not wrap the uncaught promise rejection.
```

packages/zone.js/test/common/promise-disable-wrap-uncaught-promise-rejection.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+                               delegate.handleError(target, error);
+                               return false;
+                             }
+        })
+        .run(() => {
+          zone = Zone.current;
+          task = Zone.currentTask;
+          error = new Error('rejectedErrorShouldBeHandled');
+          try {
+            // throw so that the stack trace is captured
```
mhevery: we don't seem to assert anything about a stack. So this `try-catch` seems unneeded?

packages/zone.js/test/common/promise-disable-wrap-uncaught-promise-rejection.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+class TestRejection {
+  prop1?: string;
+  prop2?: string;
+}
+
+describe('disable wrap uncaught promise rejection', () => {
+  it('should notify Zone.onHandleError if no one catches promise', (done) => {
```
mhevery:
```suggestion
  it('should notify Zone.onHandleError if promise is uncaught', (done) => {
```

Is `uncaught` the right word? you mean no one processes rejection??? Or is that the same thing?

packages/zone.js/test/common/promise-disable-wrap-uncaught-promise-rejection.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+    setTimeout(() => {
+      expect(promiseError).toBe(error);
+      expect((promiseError as any)['rejection']).toBe(error);
+      expect((promiseError as any)['zone']).toBe(zone);
+      expect((promiseError as any)['task']).toBe(task);
+      done();
+    });
+  });
+
+  it('should print original information when throw a not error object with a message property',
```
mhevery:
```suggestion
  it('should print original information when a non-Error object is used for rejection',
```

mhevery (changes requested):

packages/zone.js/lib/common/promise.ts
JiaLiPassion: Sorry, those lines are not needed, deleted.

packages/zone.js/lib/common/promise.ts
JiaLiPassion: I believe the reason of the logic here is if the `rejected value` is not an `Error` object, we will build an `Error` object to add more readable message and `stack trace`.

packages/zone.js/lib/zone.configurations.api.ts
JiaLiPassion: Got it, updated. Thanks.

packages/zone.js/lib/zone.configurations.api.ts
JiaLiPassion: Got it, updated, thanks.

packages/zone.js/lib/zone.configurations.api.ts
JiaLiPassion: Got it, updated, thanks.

packages/zone.js/test/common/promise-disable-wrap-uncaught-promise-rejection.spec.ts
JiaLiPassion: The `try/catch` is for `IE` browser, because the `stack` property will only be there when `throw` in IE.

packages/zone.js/test/common/promise-disable-wrap-uncaught-promise-rejection.spec.ts
JiaLiPassion: I believe `uncaught` and `no one processes rejection` is the same thing here.

packages/zone.js/test/common/promise-disable-wrap-uncaught-promise-rejection.spec.ts
JiaLiPassion: Got it, updated, thanks.

integration/_payload-limits.json
JiaLiPassion: I changed the logic a bit, now the size change will under the limit, so we don't need to update this file for now. And the reason the size is bigger because last time the `Passive event` PR's changes also included.

mhevery (approved):

mhevery:
[presubmit](http://test/OCL:299849439:BASE:299849484:1583768489601:792a0f63)

AndrewKushnir:
FYI, changes in this PR were tested against g3 as a part of https://github.com/angular/angular/pull/36043, using Global TAP run, [which was successful](https://test.corp.google.com/ui#id=OCL:300603599:BASE:301077666:1584325797358:d8800cfc&include_skipped=true). Thank you.

PR #35876, opened 2020-03-05
feat(ngcc): support JSON formatted console logging

petebacondarwin:
This commit adds a `Logger` that will format messages as JSON
strings before writing them to the console. This should allow tools
to parse the messages if running ngcc in a separate process.

FW-1960

(I have set the target as "master-only" since this is arguably a new public API for ngcc, so I expect it to go into 9.1.0)

### DIFF
--- a/packages/compiler-cli/ngcc/main-ngcc.ts
+++ b/packages/compiler-cli/ngcc/main-ngcc.ts
@@ -11,7 +11,8 @@ import * as yargs from 'yargs';
 import {resolve, setFileSystem, CachedFileSystem, NodeJSFileSystem} from '../src/ngtsc/file_system';
 import {mainNgcc} from './src/main';
 import {ConsoleLogger} from './src/logging/console_logger';
-import {LogLevel} from './src/logging/logger';
+import {LogLevel, Logger} from './src/logging/logger';
+import {JsonConsoleLogger} from './src/logging/json_console_logger';
 
 // CLI entry point
 if (require.main === module) {
@@ -76,6 +77,11 @@ if (require.main === module) {
             describe: 'The lowest severity logging message that should be output.',
             choices: ['debug', 'info', 'warn', 'error'],
           })
+          .option('log-format', {
+            describe: 'Choose the format of the logging messages.',
+            choices: ['default', 'json'],
+            default: 'default',
+          })
           .help()
           .parse(args);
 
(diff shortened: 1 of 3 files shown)


packages/compiler-cli/ngcc/main-ngcc.ts
```diff
@@ -92,11 +97,13 @@ if (require.main === module) {
   const compileAllFormats = !options['first-only'];
   const createNewEntryPointFormats = options['create-ivy-entry-points'];
   const logLevel = options['l'] as keyof typeof LogLevel | undefined;
+  const jsonLogging = options['json'];
   const enableI18nLegacyMessageIdFormat = options['legacy-message-ids'];
 
   (async() => {
     try {
-      const logger = logLevel && new ConsoleLogger(LogLevel[logLevel]);
+      const logger = logLevel && (jsonLogging ? new JsonConsoleLogger(LogLevel[logLevel]) :
```
stcom77: may be to use something like fabric of loggers? Now is 2 types of loggers. But if will needs additional logger types?

alan-agius4 (commented):
I think this is good. Though can we wait until we figure out how we are going to integrate this in the CLI. Since if it will not be used by the CLI, there would be no reason why these should be merged.

Thanks @petebacondarwin :)

packages/compiler-cli/ngcc/main-ngcc.ts
petebacondarwin: I think you are suggesting that we should. make this more general to handle other logger kinds?
Maybe in the future we would refactor this if we had more logging options at the command line interface, but until then I would rather not add more complexity here.

(The programmatic API already allows custom loggers to be provided)

packages/compiler-cli/ngcc/main-ngcc.ts
```diff
@@ -75,6 +76,10 @@ if (require.main === module) {
             describe: 'The lowest severity logging message that should be output.',
             choices: ['debug', 'info', 'warn', 'error'],
           })
+          .option('json', {
```
gkalpak: It might be a good idea imo to use a more specific name (something that indicates it is about logging), in case we want to add some other JSON-related option in the future.

gkalpak (approved):
One minor (optional) suggestion. Otherwise lgtm :+1:

packages/compiler-cli/ngcc/main-ngcc.ts
petebacondarwin: changed to `log-format`

packages/compiler-cli/ngcc/src/logging/json_console_logger.ts
```diff
@@ -0,0 +1,38 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import {LogLevel} from './console_logger';
```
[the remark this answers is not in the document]
petebacondarwin: rebased and picked that up

packages/compiler-cli/ngcc/main-ngcc.ts
petebacondarwin: In the end the comment by @gkalpak to make the option more "log" oriented encouraged me to go for a `log-format` option, which can only be `default` or `json`. This would give scope for other loggers in the future.

petebacondarwin:
Blocked on whether CLI actually needs it or not...

petebacondarwin:
Closing as this is no longer needed.

PR #35881, opened 2020-03-05
test(zone.js): sets rollup_bundle to use the --silent flag in karma tests

josephperrott:


IgorMinar (approved):

matsko:
@josephperrott this fails on `patch`. Please open a new PR that targets that branch specifically.

josephperrott:
It is intended to be `master-only`.

PR #35884, opened 2020-03-05
refactor(core): misc refactoring to support upcoming changes to improve synthetic host bindings

AndrewKushnir:
This commit performs a few updates to internal functions that would be required in upcoming changes to support synthetic host bindings in Directives.

* the `elementPropertyInternal` function was refactored to accept renderer as an argument (prior to that, there was a function that loads the renderer in some specific way for animation bindings)
* `elementPropertyInternal`, `elementAttributeInternal` and `listenerInternal` functions were updated to have a fixed set of arguments (for better performance)
* `elementPropertyInternal` and `elementAttributeInternal` functions were updated to take `tNode` as an argument instead of passing node index (that was used to retrieve `tNode` internally), in some cases we already have `tNode` available or we can retrieve it from the state

The refactoring was triggered by the need to pass different renderers to the `elementPropertyInternal` to support synthetic host bindings in Directives (see this comment for additional context: https://github.com/angular/angular/pull/35568/files#r388034584).

## PR Type

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/render3/instructions/attribute_interpolation.ts
+++ b/packages/core/src/render3/instructions/attribute_interpolation.ts
@@ -6,7 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 import {SanitizerFn} from '../interfaces/sanitization';
-import {getBindingIndex, getLView, getSelectedIndex, getTView} from '../state';
+import {getBindingIndex, getLView, getSelectedTNode, getTView} from '../state';
 import {NO_CHANGE} from '../tokens';
 import {interpolation1, interpolation2, interpolation3, interpolation4, interpolation5, interpolation6, interpolation7, interpolation8, interpolationV} from './interpolation';
 import {elementAttributeInternal, storePropertyBindingMetadata} from './shared';
@@ -43,13 +43,12 @@ export function ɵɵattributeInterpolate1(
   const lView = getLView();
   const interpolatedValue = interpolation1(lView, prefix, v0, suffix);
   if (interpolatedValue !== NO_CHANGE) {
-    const nodeIndex = getSelectedIndex();
-    const tView = getTView();
+    const tNode = getSelectedTNode();
     elementAttributeInternal(
-        nodeIndex, attrName, interpolatedValue, tView, lView, sanitizer, namespace);
+        tNode, lView, attrName, interpolatedValue, sanitizer || null, namespace || null);
     ngDevMode &&
         storePropertyBindingMetadata(
-            tView.data, nodeIndex, 'attr.' + attrName, getBindingIndex() - 1, prefix, suffix);
+            getTView().data, tNode, 'attr.' + attrName, getBindingIndex() - 1, prefix, suffix);
   }
   return ɵɵattributeInterpolate1;
 }
@@ -86,13 +85,12 @@ export function ɵɵattributeInterpolate2(
   const lView = getLView();
   const interpolatedValue = interpolation2(lView, prefix, v0, i0, v1, suffix);
   if (interpolatedValue !== NO_CHANGE) {
-    const nodeIndex = getSelectedIndex();
-    const tView = getTView();
+    const tNode = getSelectedTNode();
     elementAttributeInternal(
-        nodeIndex, attrName, interpolatedValue, tView, lView, sanitizer, namespace);
+        tNode, lView, attrName, interpolatedValue, sanitizer || null, namespace || null);
     ngDevMode &&
         storePropertyBindingMetadata(
-            tView.data, nodeIndex, 'attr.' + attrName, getBindingIndex() - 2, prefix, i0, suffix);
+            getTView().data, tNode, 'attr.' + attrName, getBindingIndex() - 2, prefix, i0, suffix);
   }
   return ɵɵattributeInterpolate2;
 }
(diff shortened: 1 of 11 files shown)


packages/core/src/render3/instructions/attribute_interpolation.ts
```diff
@@ -132,12 +130,11 @@ export function ɵɵattributeInterpolate3(
   const lView = getLView();
   const interpolatedValue = interpolation3(lView, prefix, v0, i0, v1, i1, v2, suffix);
   if (interpolatedValue !== NO_CHANGE) {
-    const nodeIndex = getSelectedIndex();
-    const tView = getTView();
+    const tNode = getSelectedTNode();
     elementAttributeInternal(
-        nodeIndex, attrName, interpolatedValue, tView, lView, sanitizer, namespace);
+        tNode, lView, attrName, interpolatedValue, sanitizer || null, namespace || null);
```
mhevery: Not sure I am fan of the `|| null` pattern. Could we just make `elementAttributeInternal` take `undefined` as well? Not sure what the right answer is, just asking.

packages/core/src/render3/state.ts
```diff
@@ -532,6 +534,12 @@ export function setSelectedIndex(index: number) {
   instructionState.lFrame.selectedIndex = index;
 }
 
+/**
+ * Gets the `tNode` that represents currently selected element.
+ */
+export function getSelectedTNode() {
+  return getTNode(instructionState.lFrame.tView, instructionState.lFrame.selectedIndex);
```
mhevery:
```suggestion
  const lFrame = instructionState.lFrame;
  return getTNode(lFrame.tView, lFrame.selectedIndex);
```

mhevery (changes requested):

AndrewKushnir:
Thanks for the review @mhevery. I've applied the changes based on your feedback. Could you please take another look when you get a chance? Thank you.

packages/core/src/render3/instructions/shared.ts
```diff
@@ -941,17 +941,15 @@ function mapPropName(name: string): string {
 }
 
 export function elementPropertyInternal<T>(
-    tView: TView, lView: LView, index: number, propName: string, value: T,
-    sanitizer?: SanitizerFn | null, nativeOnly?: boolean,
-    loadRendererFn?: ((tNode: TNode, lView: LView) => Renderer3) | null): void {
+    tView: TView, tNode: TNode, lView: LView, propName: string, value: T, renderer: Renderer3,
+    sanitizer?: SanitizerFn | null, nativeOnly?: boolean): void {
```
mhevery: I see these are optional as well. Not sure if you want to take on them or not.

packages/core/src/render3/instructions/shared.ts
```diff
@@ -1438,19 +1435,18 @@ function addComponentLogic<T>(lView: LView, hostTNode: TElementNode, def: Compon
 }
 
 export function elementAttributeInternal(
-    index: number, name: string, value: any, tView: TView, lView: LView,
-    sanitizer?: SanitizerFn | null, namespace?: string) {
+    tNode: TNode, lView: LView, name: string, value: any, sanitizer?: SanitizerFn | null,
+    namespace?: string | null) {
```
mhevery: same here, optional args.

packages/core/src/render3/state.ts
```diff
@@ -8,10 +8,12 @@
 
 import {StyleSanitizeFn} from '../sanitization/style_sanitizer';
 import {assertDefined, assertEqual} from '../util/assert';
+
```
mhevery: nit: revert extra blank line

mhevery (approved):

AndrewKushnir:
[Presubmit](http://test/OCL:300433799:BASE:300433828:1583969868877:9a619582)

pkozlowski-opensource (changes requested):
I'm marking this PR as "Request changes" since I would like to discuss some of the refactorings with @AndrewKushnir .

More specifically, I see how the "the elementPropertyInternal function was refactored to accept renderer as an argument (prior to that, there was a function that loads the renderer in some specific way for animation bindings)" refactoring is useful / required but I'm less convinced about other ones.

Also, would love to learn more about "elementPropertyInternal, elementAttributeInternal and listenerInternal functions were updated to have a fixed set of arguments (for better performance)" - I'm curious about the benchmark that showed the speed-up and the exact numbers.

Finally I'm not sure how I feel abut "elementPropertyInternal and elementAttributeInternal functions were updated to take tNode as an argument instead of passing node index (that was used to retrieve tNode internally), in some cases we already have tNode available or we can retrieve it from the state" - it seems to me like previously we were getting `TNode` instance only in dev mode (this is the only place where we need it, right?) but now we do so systematically. Would love to chat about it more!

I think that it would be _much_ better to split this PR into multiple commits (one per refactoring) if possible to aid reviewing.

AndrewKushnir:
Hi @pkozlowski-opensource, thanks for the review and comments. Please see replies below.

> Also, would love to learn more about "elementPropertyInternal, elementAttributeInternal and listenerInternal functions were updated to have a fixed set of arguments (for better performance)" - I'm curious about the benchmark that showed the speed-up and the exact numbers.

Prior to this change, the `elementPropertyInternal` and `elementAttributeInternal` functions were sometimes called with variable number of arguments, since function signature allowed it (there was a couple optional arguments at the end of these functions). This refactoring enforces the presence of all arguments. I don't have exact numbers (perf was not the main goal of this refactoring), but I can run a suite of micro benchmarks to see if there is any noticeable impact.

> Finally I'm not sure how I feel abut "elementPropertyInternal and elementAttributeInternal functions were updated to take tNode as an argument instead of passing node index (that was used to retrieve tNode internally), in some cases we already have tNode available or we can retrieve it from the state" - it seems to me like previously we were getting TNode instance only in dev mode (this is the only place where we need it, right?) but now we do so systematically.

In `elementPropertyInternal`, the `tNode` is [always retrieved](https://github.com/angular/angular/blob/master/packages/core/src/render3/instructions/shared.ts#L949), so there should be no perf difference. In the `elementAttributeInternal` function, the `tNode` is retrieved when [we update the value](https://github.com/angular/angular/blob/master/packages/core/src/render3/instructions/shared.ts#L1453), but for that we always access `tView` (as it's passed as one of the arguments). In this updated code, instead of accessing and passing `tView`, we do access and pass `tNode`, which should presumably be at the same "cost".

pkozlowski-opensource (approved):
LGTM after @AndrewKushnir additional info - thnx!

AndrewKushnir:
[Presubmit #2](http://test/OCL:300433799:BASE:301214448:1584386859015:6f79aa4d)

PR #35768, opened 2020-02-28
ci: move public-api goldens to goldens directory

josephperrott:
Addresses DEV-16

PR #35885, opened 2020-03-05
Public API guard for angularCompilerOptions

alxhub:
Refactor and introduction of a public API guard for `angularCompilerOptions` in tsconfig.


packages/compiler-cli/src/ngtsc/core/api/src/options.ts
```diff
@@ -0,0 +1,72 @@
...
+   * Whether the compiler should avoid generating code for classes that haven't been exported.
+   * This is only active when building with `enableIvy: true`. Defaults to `true`.
+   */
+  compileNonExportedClasses?: boolean;
+
+  /**
+   * Whether to remove blank text nodes from compiled templates. It is `false` by default starting
+   * from Angular 6.
+   */
+  preserveWhitespaces?: boolean;
```
JoostK: This was here before this change, but this option is already declared in `LegacyNgOptions`.

packages/compiler-cli/src/ngtsc/core/api/src/options.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  /**
+   * Disable TypeScript Version Check.
+   */
+  disableTypeScriptVersionCheck?: boolean;
```
JoostK: This one should probably be under the public API guard.

packages/compiler-cli/src/ngtsc/core/api/src/public_options.ts
```diff
@@ -6,63 +6,13 @@
...
-export interface ExtendedTsCompilerHost extends ts.CompilerHost, Partial<ResourceHost>,
-    Partial<UnifiedModulesHost> {}
-
 /**
  * Options supported by the legacy View Engine compiler, which are still consumed by the Angular Ivy
  * compiler for backwards compatibility.
  *
  * These are expected to be removed at some point in the future.
+ *
+ * @publicApi
```
JoostK: This is here for the golden generation to include this type, however the type itself is not really public API from a programmatic usage perspective. I'm not sure what can be done about it, but maybe include a disclaimer in the comment.

JoostK (changes requested):
Some minor comments. Also, merging this needs to be coordinated with #35768.

packages/compiler-cli/src/ngtsc/core/api/src/public_options.ts
alxhub: There is no exported API/documentation generated for these types anyway, so the marker is more "in spirit".

packages/compiler-cli/src/ngtsc/core/api/src/options.ts
alxhub: Good catch - deleted.

packages/compiler-cli/src/ngtsc/core/api/src/options.ts
alxhub: Done.

packages/compiler-cli/src/ngtsc/core/api/src/public_options.ts
JoostK: True, true

JoostK (approved):
LGTM

alxhub:
[Presubmit](http://test/OCL:299424667:BASE:299424701:1583530293538:c4eb3453)

IgorMinar (approved):
lgtm for public api changes/tracking. thanks!

gkalpak (approved):

PR #35889, opened 2020-03-06
fix(core): Host classes should not be fed back into `@Input`

### LINKED ISSUE #35383: Input with name "class" and host binding for class have a different behaviour in ng8 and ng9
# 🐞 bug report

### Affected Package

### Is this a regression?
Yes, in angular before version 9 this bug was not presented.
Before ng9 @Input properties don't be affected with host.class.

### Description

```
@Component({
  ...
  host: { class: 'attributes' },
})
export class SomeComponent {
  @Input('class')
  public klass: string;
}
```

In angular before ng9 for pice of code above we had a bit different behaviour.

In case, when we use our component like `<some-component class="test"></some-component>` for angular 8, after ngOnInit, we had only "test" in "klass" variable. But for angular 9 we have concatenated input property with host.class and now it equals klass === "attributes test" for example above.

It looks strange and logically at the same time. Possibly, there is feature of Ivy, but I want to clarify and be sure that its not a bug of Ivy and ng9.

My examples for reproduction is totally the same with only one thing different - angular version.

## 🔬 Minimal Reproduction
For angular version 8 and early: https://ng-run.com/edit/urRUMm5njGb9EAOsgtuh

For angular version 9: https://ng-run.com/edit/y5YZ2sHv9uC0N900mHAz

## 🔥 Exception or Error
I expect that it works like before. So, @Input('class') doesn't affected with host.class

## 🌍  Your Environment

**Angular Version:**

angular 8.2.14 and angular 9.0

mhevery:
Previously all static styling information (including the ones from component/directive host bindings) would get merged into a single value before it would be written into the `@Input('class'/'style')`. The new behavior specifically excludes host values from the `@Input` bindings.

Fix #35383

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No


packages/core/src/render3/instructions/styling.ts
```diff
@@ -257,11 +261,15 @@ export function checkStylingMap(
       // the binding has removed it. This would confuse `[ngStyle]`/`[ngClass]` to do the wrong
       // thing as it would think that the static portion was removed. For this reason we
       // concatenate it so that `[ngStyle]`/`[ngClass]`  can continue to work on changed.
-      let staticPrefix = isClassBased ? tNode.classes : tNode.styles;
-      ngDevMode && isClassBased === false && staticPrefix !== null &&
+      if (tNode.classesNoHost == null) {
+        computeStaticStyling(tNode, tNode.attrs, true);
+      }
+      let staticPrefix = isClassBased ? tNode.classesNoHost : tNode.stylesNoHost;
+      debugger;
```
krassx: `debugger;` statement here looks like leftover

packages/core/src/render3/styling/static_styling.ts
```diff
@@ -19,13 +17,12 @@ import {getTView} from '../state';
...
-      assertFirstCreatePass(getTView(), 'Expecting to be called in first template pass only');
-  let styles: string|null = tNode.styles;
-  let classes: string|null = tNode.classes;
+export function computeStaticStyling(
+    tNode: TNode, attrs: TAttributes | null, noHost: boolean): void {
+  let styles: string|null = noHost ? null : tNode.styles;
+  let classes: string|null = noHost ? null : tNode.classes;
   let mode: AttributeMarker|0 = 0;
-  for (let i = 0; i < attrs.length; i++) {
+  for (let i = 0; attrs !== null && i < attrs.length; i++) {
```
krassx: We can just do `i < attrs?.length` here

packages/core/src/render3/instructions/styling.ts
mhevery: This is not ready for review yet.

packages/core/src/render3/styling/static_styling.ts
mhevery: If we did that than it would be `i < undefined` which means that the VM would not always see a number at that location and it would cause deoptimization.

pkozlowski-opensource:
@mhevery it looks like this one needs rebase (or at least this is what CI is trying to say)

mhevery:
[presubmit](http://test/OCL:300141938:BASE:300141992:1583865996036:cb9cf5c3)

alxhub (approved):
This LGTM from a behavior standpoint.

@mhevery I'd like to suggest expanding the commit message a bit. I remember the context from the FW sync, but I think it'd be useful if the commit message:

* described the situation (a directive which both declares an `@Input('class')` or `@Input('style')` and a host binding to style or class properties
* described the previous behavior's effect (the message right now talks about merging of styling information, but isn't really clear that a directive's host style/class bindings could be visible in its own inputs).

mhevery:
[presubmit](http://test/OCL:310641674:BASE:310641712:1588977351386:6f644731)

packages/core/src/render3/interfaces/node.ts
```diff
@@ -502,14 +502,25 @@ export interface TNode {
...
   styles: string|null;
 
+
+  /**
+   * A collection of all `style` static values for an element excluding host sources.
+   *
+   * Populated when there are one or more initial `style`s on an element
+   * (e.g. `<div style="width:200px;">`)
+   */
+  stylesNoHost: string|null;
```
kara, line 522: Nit: might read easier this way?

```suggestion
  stylesWithoutHost: string|null;
```

packages/core/src/render3/interfaces/node.ts
```diff
@@ -540,14 +551,24 @@ export interface TNode {
...
    */
   classes: string|null;
 
+  /**
+   * A collection of all class static values for an element excluding host sources.
+   *
+   * Populated when there are one or more initial classes on an element
+   * (e.g. `<div class="SOME_CLASS">`)
+   */
+  classesNoHost: string|null;
```
kara, line 570:
```suggestion
  classesWithoutHost: string|null;
```

packages/core/src/render3/interfaces/node.ts
```diff
@@ -502,14 +502,25 @@ export interface TNode {
...
+   *   (e.g. `@Directive({host: {style: "width:200px;" } }`)
    */
   styles: string|null;
 
+
+  /**
+   * A collection of all `style` static values for an element excluding host sources.
+   *
+   * Populated when there are one or more initial `style`s on an element
+   * (e.g. `<div style="width:200px;">`)
```
kara, line 520: Might want to add the reason why this needs to be stored separately from `tNode.styles`, e.g.

```suggestion
   * (e.g. `<div style="width:200px;">`)
   * Must be stored separately from `tNode.styles` to facilitate setting directive 
   * inputs that shadow the `style` property. 
   * If we used `tNode.styles` as is for shadowed inputs, we would feed host styles 
   * back into directives as "inputs". 
   * If we used `tNode.attrs`, we would have to concatenate the attributes on 
   * every template pass. Instead, we process once on first create pass and store here.
   
```

kara (changes requested):
General approach LGTM, just some smaller comments below

packages/core/test/acceptance/styling_spec.ts
```diff
@@ -308,14 +308,87 @@ describe('styling', () => {
...
+      template: `
+          <div style="width: 1px;" dir-shadows-class-input></div>
+          <div style="width: 1px;" [style]=" 'height:1px;' " dir-shadows-class-input></div>
+          `
+    })
+    class Cmp {
+    }
+
+    @Directive({selector: '[dir-shadows-class-input]', host: {'style': 'color: red;'}})
+    class DirectiveShadowsClassInput {
```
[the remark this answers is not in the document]
mhevery: fixed

packages/core/src/render3/interfaces/node.ts
```diff
@@ -502,14 +502,25 @@ export interface TNode {
...
 
   /**
-   * A collection of all style static values for an element.
+   * A collection of all `style` static values for an element (including from host).
    *
    * This field will be populated if and when:
    *
-   * - There are one or more initial styles on an element (e.g. `<div style="width:200px">`)
+   * - There are one or more initial `style`s on an element (e.g. `<div style="width:200px;">`)
+   * - There are one or more initial `style`s on an directive/component host
```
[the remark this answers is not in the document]
mhevery: fixed

packages/core/src/render3/instructions/shared.ts
```diff
@@ -851,8 +851,10 @@ export function createTNode(
                          tParent,    // parent: TElementNode|TContainerNode|null
                          null,       // projection: number|(ITNode|RNode[])[]|null
                          null,       // styles: string|null
+                         null,       // stylesNoHost: string|null
```
kara, line 854:
```suggestion
                         null,       // stylesWithoutHost: string|null
```

packages/core/src/render3/instructions/shared.ts
```diff
@@ -851,8 +851,10 @@ export function createTNode(
...
                          undefined,  // residualStyles: string|null
                          null,       // classes: string|null
+                         null,       // classesNoHost: string|null
```
kara, line 857:
```suggestion
                         null,       // classesWithoutHost: string|null
```

packages/core/src/render3/interfaces/node.ts
```diff
@@ -540,14 +556,29 @@ export interface TNode {
...
   classes: string|null;
 
+  /**
+   * A collection of all class static values for an element excluding host sources.
+   *
+   * Populated when there are one or more initial classes on an element
+   * (e.g. `<div class="SOME_CLASS">`)
+   * Must be stored separately from `tNode.classes` to facilitate setting directive
+   * inputs that shadow the `class` property. If we used `tNode.classes` as is for shadowed inputs,
+   * we would feed host class back into directives as "inputs". If we used `tNode.attrs`, we would
```
kara, line 576:
```suggestion
   * we would feed host classes back into directives as "inputs". If we used `tNode.attrs`, we would
```

kara (approved):
LGTM

packages/core/src/render3/instructions/shared.ts
mhevery: fixed

packages/core/src/render3/instructions/shared.ts
mhevery: fixed

packages/core/src/render3/interfaces/node.ts
mhevery: fixed

mhevery:
[presubmit](http://test/OCL:310641674:BASE:311581077:1589483655306:c70d2795)

PR #35895, opened 2020-03-06
fix(language-service): resolve the real path for symlink

### LINKED ISSUE #657: Cannot resolve definitions from pnpm modules
**Describe the bug**

We are getting an error when using some pipes since we updated Angular Language Service to 0.900.14. 

For example, in one of our components we have the following variable:

```
export class LuBadgeComponent implements OnInit {
    @Input() label: string;
    ...
}
```

Then, in the template we use the uppercase pipe on it:

```
<p>
    {{ label | uppercase }}
</p>
```

And we're getting the following error in VS Code:

image

**To Reproduce**

1. Create a component with a string variable
2. Use the uppercase pipe on the variable inside the template

(see the example above)

**Expected behavior**

We expect no error from this piece of code since the code compiles correctly:

image

**Logs**

Please attach two logs:

1. Console output 

```
[Info  - 18:18:10] Angular language server process ID: 8428
[Info  - 18:18:10] Using typescript v3.7.4 from c:\Users\lucas\.vscode\extensions\angular.ng-template-0.900.14\node_modules\typescript\lib\tsserverlibrary.js
[Info  - 18:18:10] Using @angular/language-service v9.0.2 from c:\Users\lucas\.vscode\extensions\angular.ng-template-0.900.14\server\node_modules\@angular\language-service\bundles\language-service.umd.js
[Info  - 18:18:10] Log file: c:\Users\lucas\AppData\Roaming\Code\logs\20200221T181704\exthost1\Angular.ng-template\nglangsvc.log
```

2. Log file
[nglangsvc.log](https://github.com/angular/vscode-ng-language-service/files/4239649/nglangsvc.log)

**Screenshots**
image

**Additional context**

ivanwonder:
when AOT resolves the module name, it will preserve the path of the symlink (because of no `host.realpath`), but the ts-server will return the real path for symlink(https://github.com/angular/vscode-ng-language-service/blob/9120631a10fc9d49b4dd16c9976cc4222b98f3ff/server/src/server_host.ts#L132).

I don't know if it needs a test case.

fixed: https://github.com/angular/vscode-ng-language-service/issues/657

https://github.com/microsoft/TypeScript/blob/6769313eee2f2403387ea935a4fef4dfddae7671/src/compiler/moduleNameResolver.ts#L951

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/packages/language-service/src/reflector_host.ts
+++ b/packages/language-service/src/reflector_host.ts
@@ -19,12 +19,17 @@ class ReflectorModuleModuleResolutionHost implements ts.ModuleResolutionHost, Me
   });
 
   readonly directoryExists?: (directoryName: string) => boolean;
+  // Resolve a symbolic link.
+  realpath?: (path: string) => string;
 
   constructor(
       private readonly tsLSHost: ts.LanguageServiceHost,
       private readonly getProgram: () => ts.Program) {
     if (tsLSHost.directoryExists) {
       this.directoryExists = directoryName => tsLSHost.directoryExists !(directoryName);
+    }
+    if (tsLSHost.realpath) {
+      this.realpath = path => tsLSHost.realpath !(path);
     }
   }
 


packages/language-service/src/reflector_host.ts
```diff
@@ -28,6 +28,9 @@ class ReflectorModuleModuleResolutionHost implements ts.ModuleResolutionHost, Me
     }
   }
 
+  // Resolve a symbolic link.
+  realpath(path: string) { return this.tsLSHost.realpath !(path); }
```
ayazhafiz: Is this interface return value optional? If so, let’s use ? instead to avoid failures down the line

ayazhafiz (approved):
LGTM, great job figuring this out!

packages/language-service/src/reflector_host.ts
ivanwonder: https://github.com/angular/vscode-ng-language-service/blob/9120631a10fc9d49b4dd16c9976cc4222b98f3ff/server/src/server_host.ts#L132
I think the `tsLSHost.realpath` come from here.

ayazhafiz: Yes, but we shouldn’t expect that the language service will only be used with that extension. Looks like the interface is defined as: https://sourcegraph.com/github.com/microsoft/TypeScript@931e33f194049f7f56eab90fe89b5c9acd100dd1/-/blob/lib/tsserverlibrary.d.ts#L2841:22. Can we define this method optionally depending on if realPath is defined in the tsLSHost?

packages/language-service/src/reflector_host.ts
```diff
@@ -28,6 +28,9 @@ class ReflectorModuleModuleResolutionHost implements ts.ModuleResolutionHost, Me
     }
```
ivanwonder: @ayazhafiz I see, I will rewrite it like here later.

packages/language-service/src/reflector_host.ts
ivanwonder: Yes, the interface is there.

ayazhafiz (approved):

PR #35897, opened 2020-03-06
feat(compiler): Add sourceSpan and keySpan to TemplateBinding

kyliau:
This commit adds fine-grained text spans to TemplateBinding for microsyntax expressions.

1. Source span
   By convention, source span refers to the entire span of the binding,
   including its key and value.
2. Key span
   Span of the binding key, without any whitespace or keywords like `let`
3. Value span
   The value span is captured by the value expression AST.

These spans are primarily needed by the language service to provide precise autocomplete
suggestions and quick hover info at particular position.

The tests for microsyntax expressions are totally overhauled for better coverage, conciseness, and organization.

This is part of a series of PRs to fix source span mapping in microsyntax expression.
For more info, please see the design doc https://docs.google.com/document/d/1mEVF2pSSMSnOloqOPQTYNiAJO0XQxA1H0BZyESASOrE/edit?usp=sharing

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/compiler/src/expression_parser/ast.ts
+++ b/packages/compiler/src/expression_parser/ast.ts
@@ -283,10 +283,58 @@ export class ASTWithSource extends AST {
   toString(): string { return `${this.source} in ${this.location}`; }
 }
 
-export class TemplateBinding {
+/**
+ * TemplateBinding refers to a particular key-value pair in a microsyntax
+ * expression. A few examples are:
+ *
+ *   |---------------------|--------------|---------|--------------|
+ *   |     expression      |     key      |  value  | binding type |
+ *   |---------------------|--------------|---------|--------------|
+ *   | 1. let item         |    item      |  null   |   variable   |
+ *   | 2. of items         |   ngForOf    |  items  |  expression  |
+ *   | 3. let x = y        |      x       |    y    |   variable   |
+ *   | 4. index as i       |      i       |  index  |   variable   |
+ *   | 5. trackBy: func    | ngForTrackBy |   func  |  expression  |
+ *   | 6. *ngIf="cond"     |     ngIf     |   cond  |  expression  |
+ *   |---------------------|--------------|---------|--------------|
+ *
+ * (6) is a notable exception because it is a binding from the template key in
+ * the LHS of a HTML attribute to the expression in the RHS. All other bindings
+ * in the example above are derived solely from the RHS.
+ */
+export type TemplateBinding = VariableBinding | ExpressionBinding;
+
+export class VariableBinding {
+  /**
+   * @param sourceSpan entire span of the binding.
+   * @param key name of the LHS along with its span.
+   * @param value optional value for the RHS along with its span.
+   */
   constructor(
-      public span: ParseSpan, sourceSpan: AbsoluteSourceSpan, public key: string,
-      public keyIsVar: boolean, public name: string, public value: ASTWithSource|null) {}
+      public readonly sourceSpan: AbsoluteSourceSpan,
+      public readonly key: TemplateBindingIdentifier,
+      public readonly value: TemplateBindingIdentifier|null) {}
+}
+
+export class ExpressionBinding {
+  /**
+   * @param sourceSpan entire span of the binding.
+   * @param key binding name, like ngForOf, ngForTrackBy, ngIf, along with its
+   * span. Note that the length of the span may not be the same as
+   * `key.source.length`. For example,
+   * 1. key.source = ngFor, key.span is for "ngFor"
+   * 2. key.source = ngForOf, key.span is for "of"
+   * 3. key.source = ngForTrackBy, key.span is for "trackBy"
+   * @param value optional expression for the RHS.
+   */
+  constructor(
+      public readonly sourceSpan: AbsoluteSourceSpan,
+      public readonly key: TemplateBindingIdentifier, public readonly value: ASTWithSource|null) {}
+}
+
+export interface TemplateBindingIdentifier {
+  source: string;
+  span: AbsoluteSourceSpan;
 }
 
 export interface AstVisitor {
--- a/packages/compiler/src/expression_parser/parser.ts
+++ b/packages/compiler/src/expression_parser/parser.ts
@@ -10,7 +10,7 @@ import * as chars from '../chars';
 import {DEFAULT_INTERPOLATION_CONFIG, InterpolationConfig} from '../ml_parser/interpolation_config';
 import {escapeRegExp} from '../util';
 
-import {AST, ASTWithSource, AbsoluteSourceSpan, AstVisitor, Binary, BindingPipe, Chain, Conditional, EmptyExpr, FunctionCall, ImplicitReceiver, Interpolation, KeyedRead, KeyedWrite, LiteralArray, LiteralMap, LiteralMapKey, LiteralPrimitive, MethodCall, NonNullAssert, ParseSpan, ParserError, PrefixNot, PropertyRead, PropertyWrite, Quote, SafeMethodCall, SafePropertyRead, TemplateBinding} from './ast';
+import {AST, ASTWithSource, AbsoluteSourceSpan, AstVisitor, Binary, BindingPipe, Chain, Conditional, EmptyExpr, ExpressionBinding, FunctionCall, ImplicitReceiver, Interpolation, KeyedRead, KeyedWrite, LiteralArray, LiteralMap, LiteralMapKey, LiteralPrimitive, MethodCall, NonNullAssert, ParseSpan, ParserError, PrefixNot, PropertyRead, PropertyWrite, Quote, SafeMethodCall, SafePropertyRead, TemplateBinding, TemplateBindingIdentifier, VariableBinding} from './ast';
 import {EOF, Lexer, Token, TokenType, isIdentifier, isQuote} from './lexer';
 
 export class SplitInterpolation {
@@ -125,7 +125,8 @@ export class Parser {
    * For example,
    * ```
    *   <div *ngFor="let item of items">
-   *                ^ `absoluteOffset` for `tplValue`
+   *         ^      ^ absoluteValueOffset for `templateValue`
+   *         absoluteKeyOffset for `templateKey`
    * ```
    * contains three bindings:
    * 1. ngFor -> null
@@ -140,16 +141,20 @@ export class Parser {
    * @param templateKey name of directive, without the * prefix. For example: ngIf, ngFor
    * @param templateValue RHS of the microsyntax attribute
    * @param templateUrl template filename if it's external, component filename if it's inline
-   * @param absoluteOffset absolute offset of the `tplValue`
+   * @param absoluteKeyOffset start of the `templateKey`
+   * @param absoluteValueOffset start of the `templateValue`
    */
   parseTemplateBindings(
-      templateKey: string, templateValue: string, templateUrl: string,
-      absoluteOffset: number): TemplateBindingParseResult {
+      templateKey: string, templateValue: string, templateUrl: string, absoluteKeyOffset: number,
+      absoluteValueOffset: number): TemplateBindingParseResult {
     const tokens = this._lexer.tokenize(templateValue);
-    return new _ParseAST(
-               templateValue, templateUrl, absoluteOffset, tokens, templateValue.length,
-               false /* parseAction */, this.errors, 0 /* relative offset */)
-        .parseTemplateBindings(templateKey);
+    const parser = new _ParseAST(
+        templateValue, templateUrl, absoluteValueOffset, tokens, templateValue.length,
+        false /* parseAction */, this.errors, 0 /* relative offset */);
+    return parser.parseTemplateBindings({
+      source: templateKey,
+      span: new AbsoluteSourceSpan(absoluteKeyOffset, absoluteKeyOffset + templateKey.length),
+    });
   }
 
   parseInterpolation(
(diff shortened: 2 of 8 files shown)


packages/language-service/src/completions.ts
```diff
@@ -453,12 +453,19 @@ class ExpressionVisitor extends NullTemplateVisitor {
     if (ast.name.startsWith('*')) {
       // This a template binding given by micro syntax expression.
       // First, verify the attribute consists of some binding we can give completions for.
+      // The sourceSpan of AttrAst points to the RHS of the attribute
+      const TEMPLATE_ATTR_PREFIX = '*';
```
AndrewKushnir: @kyliau it looks like this const can be moved to the top level (outside of the class) and also used in the `if` statement above.

kyliau: good call, done!

packages/compiler/src/expression_parser/parser.ts
```diff
@@ -901,13 +914,15 @@ export class _ParseAST {
       return null;
     }
     this.advance();  // consume the 'as' keyword
-    const {key} = this.expectTemplateBindingKey();
+    const {key, keySpan} = this.expectTemplateBindingKey();
```
AndrewKushnir: It looks like in most cases when we retrieve a `keySpan`, we then convert it to an absolute one (i.e. keySpan.toAbsolute(this.absoluteOffset)). I'm wondering if we can return absolute form from `expectTemplateBindingKey` call instead and just pass it further? WDYT?

packages/compiler/src/expression_parser/parser.ts
```diff
@@ -847,18 +851,23 @@ export class _ParseAST {
     const bindings: TemplateBinding[] = [];
     this.consumeOptionalCharacter(chars.$COLON);  // trackBy: trackByFunction
     const valueExpr = this.getDirectiveBoundTarget();
-    const span = new ParseSpan(keySpan.start, this.inputIndex);
-    bindings.push(new TemplateBinding(
-        span, span.toAbsolute(absoluteOffset), key, false /* keyIsVar */, valueExpr?.source || '', valueExpr));
+    const spanStart = absoluteOffset + keySpan.start;
+    let spanEnd = this.absoluteOffset + this.inputIndex;
```
AndrewKushnir: Just an idea: may be we can have a getter function called for ex. `currentAbsoluteOffset` that we can use to get the current offset value?

AndrewKushnir (changes requested):
Thanks for the changes and additional docs @kyliau  👍 

I've added a couple comments, could you please have a look?

Thank you.

packages/compiler/src/expression_parser/parser.ts
```diff
@@ -140,16 +141,17 @@ export class Parser {
...
   parseTemplateBindings(
-      templateKey: string, templateValue: string, templateUrl: string,
-      absoluteOffset: number): TemplateBindingParseResult {
+      templateKey: string, templateValue: string, templateUrl: string, absoluteKeyOffset: number,
+      absoluteValueOffset: number): TemplateBindingParseResult {
     const tokens = this._lexer.tokenize(templateValue);
     return new _ParseAST(
-               templateValue, templateUrl, absoluteOffset, tokens, templateValue.length,
+               templateValue, templateUrl, absoluteValueOffset, tokens, templateValue.length,
                false /* parseAction */, this.errors, 0 /* relative offset */)
```
ayazhafiz: I think this should make `offset` on the _ParseAST constructor obsolete, no? Could we remove it?

packages/compiler/src/expression_parser/parser.ts
ayazhafiz: This is a bit difficult to reason about. `absoluteOffset` refers to the offset of the key, passed as a parameter. `this.absoluteOffset` is the offset of the template binding value, which is passed statefully to the parser on construction. In general, I don't think we should assume that `this.absoluteOffset` plays that role, because this parser does more than just parse template bindings. I would recommend renaming the two types of absoluteOffsets used here and passing them as parameters instead.

On the other hand, shouldn't we know the location of the value after `valueExpr` is parsed?

packages/language-service/src/completions.ts
```diff
@@ -450,15 +451,21 @@ class ExpressionVisitor extends NullTemplateVisitor {
   }
 
   visitAttr(ast: AttrAst) {
-    if (ast.name.startsWith('*')) {
+    if (ast.name.startsWith(TEMPLATE_ATTR_PREFIX)) {
```
ayazhafiz: nit: place this constant at the top of the method since this method is the only place its used

packages/language-service/src/completions.ts
```diff
@@ -450,15 +451,21 @@ class ExpressionVisitor extends NullTemplateVisitor {
...
       // This a template binding given by micro syntax expression.
       // First, verify the attribute consists of some binding we can give completions for.
+      // The sourceSpan of AttrAst points to the RHS of the attribute
+      const templateKey = ast.name.substring(TEMPLATE_ATTR_PREFIX.length);
+      const templateValue = ast.sourceSpan.toString();
+      const templateUrl = ast.sourceSpan.start.file.url;
+      // TODO(kyliau): We are unable to determine the absolute offset of the key
+      // but it is okay here, because we are only looking at the RHS of the attr
+      const absKeyOffset = 0;
```
ayazhafiz: shouldn't the key offset be the start of the attribute name?

packages/language-service/src/locate_symbol.ts
```diff
@@ -200,28 +200,35 @@ function getSymbolInMicrosyntax(info: AstResult, path: TemplateAstPath, attribut
...
         const directiveAst = template.directives[0];
         if (directiveAst) {
           const symbol = findInputBinding(info, tb.key.substring(1), directiveAst);
           if (symbol) {
-            result = {symbol, span: tb.span};
+            result = {
+              symbol,
+              // the span here has to be relative to the start of the template
+              // value so deduct the absolute offset.
+              span: offsetSpan(tb.keySpan, -absValueOffset),
```
ayazhafiz: Let's add a TODO, I think we should move to only dealing with one type of span

ayazhafiz (changes requested):

kyliau (commented):
@AndrewKushnir @ayazhafiz Thank you for the valuable feedback!
After analyzing the structure of the code, I realized much of the complexity comes from the decision to use `ASTWithSource` to represent the value of the binding for __both__ variable binding and expression binding. 
`ASTWithSource` is messy to work with because it contains both relative span and absolute span, so we had to keep track of both.
The fundamental problem stems from the usage of __field__ variable `keyIsVar` to differentiate the data types for variable binding and expression binding. The right thing to do here is to use the __type information__ to represent the data, and I've created two separate classes, `VariableBinding` and `ExpressionBinding` to reflect this.
With this change, there is no more confusing relative spans to deal with. All spans handled within the microsyntax expression parsing are absolute spans. The implementation should be much cleaner now. Please take a look and let me know what you think. Thank you!

packages/compiler/src/expression_parser/ast.ts
```diff
@@ -283,10 +283,57 @@ export class ASTWithSource extends AST {
...
+   */
+  constructor(
+      public readonly sourceSpan: AbsoluteSourceSpan, public readonly key: TemplateBindingData,
+      public readonly value: ASTWithSource|null) {}
+}
+
+export interface TemplateBindingData {
+  source: string;
+  span: AbsoluteSourceSpan;
 }
```
ayazhafiz: nit: consider rename to `TemplateBindingContext`

packages/compiler/test/expression_parser/parser_spec.ts
```diff
@@ -247,200 +247,305 @@ describe('parser', () => {
...
+        ['*a="b"', 'a', 'b', false],
+        ['*a-b="c"', 'a-b', 'c', false],
+        ['*a="1+1"', 'a', '1+1', false],
+      ];
+      for (const [attr, key, value, keyIsVar] of cases) {
+        const bindings = parseTemplateBindings(attr);
+        expect(humanize(bindings)).toEqual([
+          [key, value, keyIsVar],
+        ]);
+      }
```
ayazhafiz: Can we extend these tests to also check the spans?

packages/language-service/src/completions.ts
ayazhafiz: Ah, I was under the impression AttrAst contained the info we needed. Thank you for explaining.

ayazhafiz (changes requested):
LGTM, just a comment regarding testing spans

packages/compiler/test/expression_parser/parser_spec.ts
kyliau: done

packages/compiler/src/expression_parser/ast.ts
kyliau: renamed to `TemplateBindingIdentifier`, thanks for pointing this out!

packages/compiler/src/expression_parser/parser.ts
```diff
@@ -886,35 +900,30 @@ export class _ParseAST {
...
     }
     this.advance();  // consume the 'as' keyword
-    const {key} = this.expectTemplateBindingKey();
-    const valueAst = new AST(valueSpan, valueSpan.toAbsolute(absoluteOffset));
-    const valueExpr = new ASTWithSource(
-        valueAst, value, this.location, absoluteOffset + valueSpan.start, this.errors);
-    const span = new ParseSpan(valueSpan.start, this.inputIndex);
-    return new TemplateBinding(
-        span, span.toAbsolute(absoluteOffset), key, true /* keyIsVar */, value, valueExpr);
+    const key = this.expectTemplateBindingKey();
```
AndrewKushnir: nit: I'm wondering if it make sense to rename `expectTemplateBindingKey` -> `consumeTemplateBindingKey` in a followup PR?

kyliau:
@ayazhafiz @AndrewKushnir I've addressed all comments, could you please take another look? Thank you!

packages/language-service/src/completions.ts
AndrewKushnir: My proposal was based on the assumption that this const can be used in other places as well, which seems like the case, for ex. here: https://github.com/angular/angular/blob/ac8dd2c32b7e4990c606e6a44653cc5c8dea2888/packages/language-service/src/completions.ts#L235
I'm fine keeping it in the function scope as well and refactor later. Up to you Keen :)

packages/compiler/src/expression_parser/ast.ts
```diff
@@ -283,10 +283,58 @@ export class ASTWithSource extends AST {
...
+ *
+ * (6) is a notable exception because it is a binding from the template key in
+ * the LHS of a HTML attribute to the expression in the RHS. All other bindings
+ * in the example above are derived solely from the RHS.
+ */
+export type TemplateBinding = VariableBinding | ExpressionBinding;
+
+export class VariableBinding {
+  /**
+   * @param sourceSpan entire span of the binding.
```
AndrewKushnir: I think it'd be great to have a quick example of the binding that would be treated as `Variable` vs `Expression`. As an option, we can extend the table that you have above and annotate each line: whether a given expression is a variable or expression binding.

AndrewKushnir (approved):
LGTM, thanks @kyliau 👍 

I left a few comments (mostly docs and minor refactoring proposals), please have a look when you get a chance.

Thank you.

AndrewKushnir:
@kyliau I think it'd make sense to also run a Global TAP (in addition to Blueprint one) to make sure everything looks good. Thank you.

packages/compiler/src/expression_parser/ast.ts
kyliau: done

packages/compiler/src/expression_parser/parser.ts
kyliau: Good call, I'll do this in a follow up PR

packages/language-service/src/completions.ts
kyliau: ah I didn't see that there's another usage of `*` in this file. I moved it back to the top level.

kyliau:
Global TAP: https://test.corp.google.com/OCL:300148674:BASE:300248062:1583902188182:84a773d3

packages/compiler/test/expression_parser/parser_spec.ts
```diff
@@ -247,200 +247,309 @@ describe('parser', () => {
...
-      const bindings = parseTemplateBindings('a', 'let b');
-      expect(keyValues(bindings)).toEqual(['a', 'let b=$implicit']);
+    it('should parse key and value', () => {
+      const cases: Array<[string, string, string | null, boolean, string, string, string | null]> =
+          [
+            // expression, key, value, VariableBinding, source span, key span, value span
+            ['*a=""', 'a', null, false, 'a="', 'a', null],
+            ['*a="b"', 'a', 'b', false, 'a="b', 'a', 'b'],
+            ['*a-b="c"', 'a-b', 'c', false, 'a-b="c', 'a-b', 'c'],
+            ['*a="1+1"', 'a', '1+1', false, 'a="1+1', 'a', '1+1'],
```
ayazhafiz: Is the missing trailing quote in the source span intentional?

ayazhafiz (commented):
Could we add span tests for all test cases? Sorry..

packages/compiler/test/expression_parser/parser_spec.ts
kyliau: Yes, due to the way the string is tokenized, the quotes never appear in the tokens at all.

kyliau:
> Could we add span tests for all test cases? Sorry..

My intention was to separate the testing for key value and their spans, so that we have tests that are focused. There is a dedicated section for testing spans only from line 482-552.

packages/compiler/test/expression_parser/parser_spec.ts
ayazhafiz: Oh, interesting. Thanks!

ayazhafiz (approved):

PR #35911, opened 2020-03-06
build: Add pullapprove verification tool to dev-infra-private package

josephperrott:
Migrates pullapprove verification tool to be available in the dev-infra-private
package

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/dev-infra/pullapprove/verify.ts
+++ b/dev-infra/pullapprove/verify.ts
@@ -1,0 +1,215 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import {readFileSync} from 'fs';
+import {IMinimatch, Minimatch} from 'minimatch';
+import * as path from 'path';
+import {cd, exec, set} from 'shelljs';
+import {parse as parseYaml} from 'yaml';
+
+interface GlobMatcher {
+  group: string;
+  glob: string;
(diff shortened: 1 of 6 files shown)


dev-infra/pullapprove/verify.ts
```diff
@@ -0,0 +1,217 @@
...
+  matchCount: number;
+}
+
+// Exit early on shelljs errors
+set('-e');
+
+// Regex Matcher for contains_any_globs conditions
+const CONTAINS_ANY_GLOBS_REGEX = /^'([^']+)',?$/;
+// Full path of the angular project directory
+const ANGULAR_PROJECT_DIR = process.cwd();
```
devversion: Would it better to accept this through an argument from the CLI? It feels brittle to rely on `process.cwd()`.

devversion (approved):
I assume this is just moving the script over right now. Will there be a follow-up to remove the one under `tools/pullapprove`?

dev-infra/pullapprove/verify.ts
josephperrott: Agreed, that it would be better to do something a little more robust, or at least allow for something more.

My goal in this PR is just migrate the script over, only making the changes needed for it to continue to work, rather than fixing things like this as well.

I think it makes perfect sense in a follow up PR.

devversion: SGTM.

devversion (approved):

PR #35916, opened 2020-03-06
refactor(core): throw more descriptive error message in case of invalid host element

### LINKED ISSUE #35240: Provide more descriptive error message in case Component matches <ng-container>
This issue is a followup from https://github.com/angular/angular/issues/34548. If Component has a selector that matches `<ng-container>`, the following error message is thrown at runtime:
```
ASSERTION ERROR: Should be one of Element but got ElementContainer.
```
Framework works as expected (`<ng-container>` can not act as a host element for Components), but we should improve this error message to make it useful/friendly and provide more info about the Component that is causing the problem.

AndrewKushnir:
This commit replaces an assert with more descriptive error message that is thrown in case `<ng-template>` or `<ng-container>` is used as host element for a Component.

Resolves #35240.

## PR Type

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/render3/instructions/shared.ts
+++ b/packages/core/src/render3/instructions/shared.ts
@@ -1381,8 +1381,9 @@ function findDirectiveDefMatches(
           ngDevMode &&
               assertNodeOfPossibleTypes(
                   tNode, [TNodeType.Element],
-                  `Element with "${tNode.tagName}" tag name is not a valid host ` +
-                      `for ${stringify(def.type)} component.`);
+                  `"${tNode.tagName}" tags cannot be used as component hosts. ` +
+                      `Please use a different tag to activate the ${
+                          stringify(def.type)} component.`);
           if (tNode.flags & TNodeFlags.isComponentHost) throwMultipleComponentError(tNode);
           markAsComponentHost(tView, tNode);
           // The component is always stored first with directives after.
--- a/packages/core/src/render3/instructions/shared.ts
+++ b/packages/core/src/render3/instructions/shared.ts
@@ -15,6 +15,7 @@ import {assertDataInRange, assertDefined, assertDomNode, assertEqual, assertGrea
 import {createNamedArrayType} from '../../util/named_array_type';
 import {initNgDevMode} from '../../util/ng_dev_mode';
 import {normalizeDebugBindingName, normalizeDebugBindingValue} from '../../util/ng_reflect';
+import {stringify} from '../../util/stringify';
 import {assertFirstCreatePass, assertLContainer, assertLView} from '../assert';
 import {attachPatchData} from '../context_discovery';
 import {getFactoryDef} from '../definition';
@@ -272,7 +273,7 @@ export function assignTViewNodeToLView(
   let tNode = tView.node;
   if (tNode == null) {
     ngDevMode && tParentNode &&
-        assertNodeOfPossibleTypes(tParentNode, TNodeType.Element, TNodeType.Container);
+        assertNodeOfPossibleTypes(tParentNode, [TNodeType.Element, TNodeType.Container]);
     tView.node = tNode = createTNode(
                              tView,
                              tParentNode as TElementNode | TContainerNode | null,  //
@@ -1278,7 +1279,7 @@ function instantiateAllDirectives(
     const isComponent = isComponentDef(def);
 
     if (isComponent) {
-      ngDevMode && assertNodeOfPossibleTypes(tNode, TNodeType.Element);
+      ngDevMode && assertNodeOfPossibleTypes(tNode, [TNodeType.Element]);
       addComponentLogic(lView, tNode as TElementNode, def as ComponentDef<any>);
     }
 
@@ -1366,7 +1367,7 @@ function findDirectiveDefMatches(
   ngDevMode && assertFirstCreatePass(tView);
   ngDevMode &&
       assertNodeOfPossibleTypes(
-          tNode, TNodeType.Element, TNodeType.ElementContainer, TNodeType.Container);
+          tNode, [TNodeType.Element, TNodeType.ElementContainer, TNodeType.Container]);
   const registry = tView.directiveRegistry;
   let matches: any[]|null = null;
   if (registry) {
@@ -1377,6 +1378,11 @@ function findDirectiveDefMatches(
         diPublicInInjector(getOrCreateNodeInjectorForNode(tNode, viewData), tView, def.type);
 
         if (isComponentDef(def)) {
+          ngDevMode &&
+              assertNodeOfPossibleTypes(
+                  tNode, [TNodeType.Element],
+                  `Element with "${tNode.tagName}" tag name is not a valid host ` +
+                      `for ${stringify(def.type)} component.`);
           if (tNode.flags & TNodeFlags.isComponentHost) throwMultipleComponentError(tNode);
           markAsComponentHost(tView, tNode);
           // The component is always stored first with directives after.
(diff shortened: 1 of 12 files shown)


packages/core/src/render3/errors.ts
```diff
@@ -25,6 +26,12 @@ export function throwMultipleComponentError(tNode: TNode): never {
   throw new Error(`Multiple components match node with tagname ${tNode.tagName}`);
 }
 
+/** Called when invalid host element is used for a Component */
+export function throwInvalidHostElementError(tNode: TNode, def: ComponentDef<any>): never {
+  throw new Error(
+      `Element with "${tNode.tagName}" tagname is not a valid host for ${stringify(def.type)} component`);
```
mhevery:
```suggestion
      `Element with "${tNode.tagName}" tagname is not a valid host for ${stringify(def.type)} component.`);
```

packages/core/src/render3/instructions/shared.ts
```diff
@@ -1239,7 +1239,9 @@ function instantiateAllDirectives(
     const isComponent = isComponentDef(def);
 
     if (isComponent) {
-      ngDevMode && assertNodeOfPossibleTypes(tNode, TNodeType.Element);
+      if (ngDevMode && tNode.type !== TNodeType.Element) {
```
mhevery: Why is this not just an assert?

```
ngDevMode && assertNodeOfPossibleTypes(tNode, ...);
```

If the assert error needs to be made better we should do that.

AndrewKushnir: The goal is to make this error message more user friendly and explain what went wrong, when assert operates with internal data structure concepts (as it should).

pkozlowski-opensource (changes requested):
I agree with @mhevery - I believe that the existing assert should not be modified. I think that the proper solution is to throw this error much earlier, when the directives are being matched in `findDirectiveDefMatches` (we do some other verifications there already).

packages/core/src/render3/instructions/shared.ts
mhevery: Are you arguing that asserts should only be used for framework level assertions, and not for developer level?

In that case I would argue we should have `assertDevFoo(condition, friendlyMsg)` methods which we can use to differentiate which asserts are meant for us and which are for the developer. But at the end of the day I feel like the two are same and hence we should keep them consistent.

AndrewKushnir:
Hi @mhevery and @pkozlowski-opensource, thanks for the feedback! I addressed both comments and used an assert + moved it to the `findDirectiveDefMatches` function to detect this situation earlier. This change is also consistent with similar change in #37111. Please let me know if everything looks good or any updates are required. Thank you.

packages/core/src/render3/node_assert.ts
```diff
@@ -26,12 +26,14 @@ export function assertNodeType(tNode: TNode, type: TNodeType): asserts tNode is
...
 
-export function assertNodeOfPossibleTypes(tNode: TNode|null, ...types: TNodeType[]): void {
+export function assertNodeOfPossibleTypes(
+    tNode: TNode|null, types: TNodeType[], message?: string): void {
   assertDefined(tNode, 'should be called with a TNode');
   const found = types.some(type => tNode.type === type);
   assertEqual(
       found, true,
-      `Should be one of ${types.map(typeName).join(', ')} but got ${typeName(tNode.type)}`);
+      message ??
```
kara, line 35:
```suggestion
      message ||
```

packages/core/src/render3/node_manipulation.ts
```diff
@@ -778,10 +778,10 @@ function applyNodes(
     renderParent: RElement|null, beforeNode: RNode|null, isProjection: boolean) {
   while (tNode != null) {
     ngDevMode && assertTNodeForLView(tNode, lView);
-    ngDevMode &&
-        assertNodeOfPossibleTypes(
-            tNode, TNodeType.Container, TNodeType.Element, TNodeType.ElementContainer,
-            TNodeType.Projection, TNodeType.Projection, TNodeType.IcuContainer);
+    ngDevMode && assertNodeOfPossibleTypes(tNode, [
+      TNodeType.Container, TNodeType.Element, TNodeType.ElementContainer, TNodeType.Projection,
+      TNodeType.Projection, TNodeType.IcuContainer
```
kara, line 783: While you're here, fix duplicate of 'projection'?

```suggestion
      TNodeType.IcuContainer
```

kara (dismissed):
Thanks for fixing!

packages/core/src/render3/node_assert.ts
AndrewKushnir: I've added this for consistency with the `assertNodeNotOfTypes` function [here](https://github.com/angular/angular/pull/37111/files#diff-31d46cb6aed13f0120b5db8c6e67f5f4R42), i.e. providing an empty string should result in rendering empty message (aside from the one generated from `assertEqual`).

AndrewKushnir:
[Presubmit](http://test/OCL:315391417:BASE:315391452:1591664079393:b6e4a433)

packages/core/test/acceptance/component_spec.ts
```diff
@@ -189,6 +189,51 @@ describe('component', () => {
     });
   });
 
+  onlyInIvy('Ivy throws an exception when invalid host elements are used for Components')
```
pkozlowski-opensource, line 192: super-nit: it would be totally awesome to have a 1-sentence comment about what VE does (ignores such components silently?).

pkozlowski-opensource (approved):
LGTM with one minor super-nit (up to you if you want to clean it up or not).
Happy to see that the checks happen now only when we match components (and not when we instantiate them!)

packages/core/test/acceptance/component_spec.ts
AndrewKushnir: Thanks for the review @pkozlowski-opensource 👍 

In VE it was also an error and I've updated acceptance test to capture VE behavior as well.

AndrewKushnir:
[Presubmit #2](http://test/OCL:315391417:BASE:318089827:1593019314262:bf76f6ee)

AndrewKushnir:
Hi @IgorMinar, after rebasing this PR I had to update the payload size limit for the `hello_world__closure` test app ([see here](https://github.com/angular/angular/pull/35916/files#diff-2ace7eeef6d66fa246807c16e7585578R65)). That's a bit unexpected, since all the code touched in this PR is behind the `ngDevMode` flag. Looking at the app size (~1200kb when regular cli-hello-world one is ~140kb) I'm wondering if the `ngDevMode` doesn't work there atm at all? Please let me know if it's a known issue, given the fact that advanced optimizations are disabled (based on the [comment here](https://github.com/angular/angular/pull/35916/files#diff-2ace7eeef6d66fa246807c16e7585578R64)) and if that's the case, could you please approve this payload size change? Thank you.

IgorMinar:
Our closure setup is currently broken and doesn't dead code eliminate code behind the ngDevMode flag. Feel free to ignore this. I've been meaning to look into this anyway so I'll see if I can fix it today.

IgorMinar (approved):
Reviewed-for: code-size

IgorMinar (approved):
Reviewed-for: size-tracking

IgorMinar (approved):
LGTM - this increase is due to a bug in our closure setup that causes code behind the ngDevMode flag to retained. I'll try to fix this bug today.

Reviewed-for: size-tracking

AndrewKushnir:
Thanks @IgorMinar for the review and additional information 👍

Note to Caretaker: this PR was approved by Pawel and was ready to go. After rebasing, the payload size issue appeared, which is described by Igor above and was approved. This PR is ready for merge now.

IgorMinar (approved):
LGTM! Thank you!

PR #35918, opened 2020-03-06
build(docs-infra): update the dependency versions of the examples

petebacondarwin:
On OS/X The examples were failing to build due to a version mismatch
with fsevents that results in errors that look like:

```
$ ng serve
...
/.../angular/aio/tools/examples/shared/node_modules/chokidar/lib/fsevents-handler.js:28
  return (new fsevents(path)).on('fsevent', callback).start();
          ^
TypeError: fsevents is not a constructor
    at createFSEventsInstance (/.../angular/aio/tools/examples/shared/node_modules/chokidar/lib/fsevents-handler.js:28:11)
    at setFSEventsListener (/.../angular/aio/tools/examples/shared/node_modules/chokidar/lib/fsevents-handler.js:82:16)
    ...
error Command failed with exit code 1.
```

This commit updates the version in yarn.lock to fix this problem.

### DIFF
--- a/aio/tools/examples/shared/package.json
+++ b/aio/tools/examples/shared/package.json
@@ -45,7 +45,7 @@
     "zone.js": "~0.10.2"
   },
   "devDependencies": {
-    "@angular-devkit/build-angular": "0.800.0",
+    "@angular-devkit/build-angular": "^0.803.25",
     "@angular/cli": "^8.0.0",
     "@angular/compiler-cli": "^8.0.0",
     "@angular/language-service": "^8.0.0",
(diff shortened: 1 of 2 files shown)


gkalpak (approved):

gkalpak:
LGTM, but CI didn't look happy. Wasn't sure if the failures were legit, so I restarted it :grin:

petebacondarwin:
The CI is not a flake. I think it needs all the patch versions to align...

gkalpak:
BTW, I think this would be addressed by #36015.

petebacondarwin:
Yeah, given that #36015 is nearing completion, we might as well close this.

PR #35920, opened 2020-03-06
docs: Add useful tip in testing guide

bampakoa:
Inspired from [Protractor style guide](https://www.protractortest.org/#/style-guide), I added a note about how E2E tests perform in speed when compared with unit tests. I think that this note will make more clear to the readers about differences between two.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/aio/content/guide/testing.md
+++ b/aio/content/guide/testing.md
@@ -3637,7 +3637,7 @@ what the test is testing. If it's doing more, then it doesn't belong here.
 E2E tests are great for high-level validation of the entire system.
 But they can't give you the comprehensive test coverage that you'd expect from unit tests.
 
-E2E tests are difficult to write and perform poorly compared to unit tests.
+E2E tests are difficult to write and perform poorly compared to unit tests. Unit tests are much faster than E2E tests.
 They break easily, often due to changes or misbehavior far removed from the site of breakage.
 
 E2E tests can't easily reveal how your components behave when things go wrong,


sonukapoor (approved):

aio/content/guide/testing.md
```diff
@@ -3637,7 +3637,7 @@ what the test is testing. If it's doing more, then it doesn't belong here.
 E2E tests are great for high-level validation of the entire system.
 But they can't give you the comprehensive test coverage that you'd expect from unit tests.
 
-E2E tests are difficult to write and perform poorly compared to unit tests.
+E2E tests are difficult to write and perform poorly compared to unit tests. Unit tests are much faster than E2E tests.
```
kapunahelewong, line 3640:
```suggestion
E2E tests are difficult to write and perform poorly compared to unit tests, which are much faster than E2E tests.
```

kapunahelewong (approved):
Thank you, @bampakoa! I have just one small suggestion, but we will need engineering to approve this one before moving forward.

aio/content/guide/testing.md
bampakoa: @kapunahelewong thanks very much for your feedback 😃. I have the feeling that your proposed change does not give proper focus that unit tests are indeed faster than e2e, compared to the initial change. What do you think?

kapunahelewong: Hm, it if we leave it as 
> Unit tests are much faster than E2E tests. 

Then the very next sentence is:

> "They break easily... 

"They" here makes it sound like we're talking about unit tests.

What if in the next sentence you replace "they" with "tests", so that you have:

> E2E tests break easily, often due to changes...

Does that help? :)

bampakoa: Sure! You are right @kapunahelewong 👍 👍. Did not notice the next sentence. Sometimes you need to check the whole context of something in order to act properly.  

Do I need to make the change by myself or wait for the engineering approval?

kapunahelewong: It is an easy thing to miss! 🙂 When you make the change is totally up to you, @bampakoa. If you would prefer to wait until engineering confirms the meaning of the change before changing the wording that's fine. Then you can do all the changes at one time. 

Sometimes engineering has other thoughts which might make you change the original wording anyway. That's why I like to see what engineering says first, but this time I just happened to notice that next sentence. I can review again after they've had a chance to look, though. 🌸

pkozlowski-opensource (changes requested):
TBH I'm not sure I see the value in adding "Unit tests are much faster than E2E tests." => for me this is just a repetition of "perform poorly" in the previous sentence ("E2E tests are difficult to write and perform poorly compared to unit tests.").

I'm totally fine with re-wording things so they are easier to understand but I don't think that the proposed change is the way to go - IMO it adds noise by repeating parts of the original sentence (again totally fine with proper attempts to re-word the original sentence). 

@kapunahelewong I will let you work with @bampakoa if you guys still think that it needs improving. But yeh, I would prefer _not_ to merge this PR in the current form.

bampakoa:
@pkozlowski-opensource thanks very much for your feedback. Your comment makes sense. What forced me to open the PR in the first place was:

1. Poor performance may refer to many things such as memory consumption and speed. I am not really sure if there is a memory problem when using E2E. 
2. Most developers, not all, want their tests to be fast, this is what they care at the end of the day. So I feel that the phrase _poor performance_ does not convey the real benefit of unit tests which is speed. The wording also feels like emitting negativity to the reader.

I understand your concerns which are totally respected. I may also exaggerate and over-analyze the whole point of a single phrase. I just wanted to make my intentions clear. @kapunahelewong let me know your thoughts and I would not mind closing the PR. Thanks a lot 🙏

kapunahelewong:
Thank you for your thoughtful response, @bampakoa. Because I'm not as knowledgeable in test performance, I'll have to follow @pkozlowski-opensource here. Thank you for all your contributions. Even if this one isn't merged like your others, it can still serve as a reference for others in the future. I'll leave it up to you to close if you like.

bampakoa:
@kapunahelewong  yeah, you know, this gives me 💪 and motivation to try harder for my next PR's. We all, in Angular community, need to do a better job each single day and make contributions that maintain the framework in top quality. And I know that this PR was not 😃 👍 . Thank you all for your feedback!

sonukapoor:
> @kapunahelewong yeah, you know, this gives me 💪 and motivation to try harder for my next PR's. We all, in Angular community, need to do a better job each single day and make contributions that maintain the framework in top quality. And I know that this PR was not 😃 👍 . Thank you all for your feedback!

That's good energy 👍

PR #35922, opened 2020-03-07
refactor(benchpress): delete broken code

wagnermaciel:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

The [firefox_extension](https://github.com/angular/angular/blob/master/packages/benchpress/src/firefox_extension/lib/main.ts#L9) has become outdated. It currently uses the `chrome` api which [has not worked since November 2017](https://developer.mozilla.org/en-US/docs/Archive/Add-ons/Add-on_SDK/Tutorials/Chrome_Authority).

## What is the new behavior?
Remove it.

- [ ] Yes
- [x] No


jelbourn (approved):
LGTM

matsko:
presubmit looks to have some breakages = https://tap.corp.google.com/ui#id=OCL:300135810:BASE:300135849:1583864423273:1b843442

jelbourn:
@matsko those failures all look super unrelated to me

AndrewKushnir:
Hi, it looks like broken targets were removed from g3 (thus restarting them would not give a clear indication), so I [started a new presubmit](http://test/OCL:300368370:BASE:300368418:1583951120684:e642f575) and replaced "merge" label with "presubmit" to indicate that.

Also, according to the process we need approvals from all reviewers before we can proceed with merging. @alxhub could you please have a look at this PR when you get a chance?

Thank you.

alxhub (approved):

AndrewKushnir:
@wagnermaciel just want to check if this PR should be merged into master and patch branches or master only? It looks like it removes some code (can someone technically depend on it?), so I'd assume it might be safer to land it in master only (to include it into 9.1.0), but PR also says that it's not a breaking change. Please let me know (I will put "cleanup" label for now to avoid merging it accidentally). Thank you.

wagnermaciel:
@AndrewKushnir I think going with the safer route is good because there is still a lot that needs to be done before the full feature I’m adding gets finished anyways.

I should repeat here that I don't think this will break anything because the firefox_extension is severely outdated. It currently uses the chrome api [which has not worked since November 2017](https://developer.mozilla.org/en-US/docs/Archive/Add-ons/Add-on_SDK/Tutorials/Chrome_Authority).

AndrewKushnir:
@wagnermaciel, thanks for the reply. Given that we've just released `9.1.0-rc.0` and created a new patch branch (`9.1.x`), we can keep "master & patch" label, so that the changes are included into `9.1.0` release. Thank you.

PR #35931, opened 2020-03-07
perf(ngcc): implement entry-point manifest

petebacondarwin:
The `DirectoryWalkerEntryPointFinder` has to traverse the
entire node_modules library everytime it executes in order to
identify the entry-points that need to be processed. This is
very time consuming (several seconds for big projects on
Windows).

This commit changes the `DirectoryWalkerEntryPointFinder` to
use the `EntryPointManifest` to store the paths to entry-points
that were found when doing this initial node_modules traversal
in a file to be reused for subsequent calls.

This dramatically speeds up ngcc processing when it has been run once already.

FW-1962

### DIFF
--- a/packages/compiler-cli/ngcc/main-ngcc.ts
+++ b/packages/compiler-cli/ngcc/main-ngcc.ts
@@ -79,7 +79,7 @@ if (require.main === module) {
           .option('invalidate-entry-point-manifest', {
             describe:
                 'If this is set then ngcc will not read an entry-point manifest file from disk.\n' +
-                'Instead it will walking the directory tree as normal looking for entry-points, and then write a new manifest file.',
+                'Instead it will walk the directory tree as normal looking for entry-points, and then write a new manifest file.',
             type: 'boolean',
             default: false,
           })
--- a/packages/compiler-cli/ngcc/main-ngcc.ts
+++ b/packages/compiler-cli/ngcc/main-ngcc.ts
@@ -76,6 +76,13 @@ if (require.main === module) {
             describe: 'The lowest severity logging message that should be output.',
             choices: ['debug', 'info', 'warn', 'error'],
           })
+          .option('invalidate-entry-point-manifest', {
+            describe:
+                'If this is set then ngcc will not read an entry-point manifest file from disk.\n' +
+                'Instead it will walking the directory tree as normal looking for entry-points, and then write a new manifest file.',
+            type: 'boolean',
+            default: false,
+          })
           .strict()
           .help()
           .parse(args);
@@ -95,6 +102,7 @@ if (require.main === module) {
   const createNewEntryPointFormats = options['create-ivy-entry-points'];
   const logLevel = options['l'] as keyof typeof LogLevel | undefined;
   const enableI18nLegacyMessageIdFormat = options['legacy-message-ids'];
+  const invalidateEntryPointManifest = options['invalidate-entry-point-manifest'];
 
   (async() => {
     try {
@@ -108,7 +116,7 @@ if (require.main === module) {
         createNewEntryPointFormats,
         logger,
         enableI18nLegacyMessageIdFormat,
-        async: options['async'],
+        async: options['async'], invalidateEntryPointManifest,
       });
 
       if (logger) {
--- a/packages/compiler-cli/ngcc/src/execution/cluster/master.ts
+++ b/packages/compiler-cli/ngcc/src/execution/cluster/master.ts
@@ -70,7 +70,7 @@ export class ClusterMaster {
 
     // First, check whether all tasks have been completed.
     if (this.taskQueue.allTasksCompleted) {
-      const duration = Math.round((Date.now() - this.processingStartTime) / 1000);
+      const duration = Math.round((Date.now() - this.processingStartTime) / 100) / 10;
       this.logger.debug(`Processed tasks in ${duration}s.`);
 
       return this.finishedDeferred.resolve();
--- a/packages/compiler-cli/ngcc/src/main.ts
+++ b/packages/compiler-cli/ngcc/src/main.ts
@@ -39,6 +39,7 @@ import {hasBeenProcessed} from './packages/build_marker';
 import {NgccConfiguration} from './packages/configuration';
 import {EntryPoint, EntryPointJsonProperty, EntryPointPackageJson, SUPPORTED_FORMAT_PROPERTIES, getEntryPointFormat} from './packages/entry_point';
 import {makeEntryPointBundle} from './packages/entry_point_bundle';
+import {EntryPointManifest} from './packages/entry_point_manifest';
 import {Transformer} from './packages/transformer';
 import {PathMappings} from './utils';
 import {cleanOutdatedPackages} from './writing/cleaning/package_cleaner';
@@ -153,14 +154,15 @@ export function mainNgcc(
   const absBasePath = absoluteFrom(basePath);
   const config = new NgccConfiguration(fileSystem, dirname(absBasePath));
   const dependencyResolver = getDependencyResolver(fileSystem, logger, config, pathMappings);
+  const entryPointManifest = new EntryPointManifest(fileSystem, config, logger);
 
   // Bail out early if the work is already done.
   const supportedPropertiesToConsider = ensureSupportedProperties(propertiesToConsider);
   const absoluteTargetEntryPointPath =
       targetEntryPointPath !== undefined ? resolve(basePath, targetEntryPointPath) : null;
   const finder = getEntryPointFinder(
-      fileSystem, logger, dependencyResolver, config, absBasePath, absoluteTargetEntryPointPath,
-      pathMappings);
+      fileSystem, logger, dependencyResolver, config, entryPointManifest, absBasePath,
+      absoluteTargetEntryPointPath, pathMappings);
   if (finder instanceof TargetedEntryPointFinder &&
       !finder.targetNeedsProcessingOrCleaning(supportedPropertiesToConsider, compileAllFormats)) {
     logger.debug('The target entry-point has already been processed');
@@ -376,14 +378,15 @@ function getDependencyResolver(
 
 function getEntryPointFinder(
     fs: FileSystem, logger: Logger, resolver: DependencyResolver, config: NgccConfiguration,
-    basePath: AbsoluteFsPath, absoluteTargetEntryPointPath: AbsoluteFsPath | null,
+    entryPointManifest: EntryPointManifest, basePath: AbsoluteFsPath,
+    absoluteTargetEntryPointPath: AbsoluteFsPath | null,
     pathMappings: PathMappings | undefined): EntryPointFinder {
   if (absoluteTargetEntryPointPath !== null) {
     return new TargetedEntryPointFinder(
         fs, config, logger, resolver, basePath, absoluteTargetEntryPointPath, pathMappings);
   } else {
     return new DirectoryWalkerEntryPointFinder(
-        fs, config, logger, resolver, basePath, pathMappings);
+        fs, config, logger, resolver, entryPointManifest, basePath, pathMappings);
   }
 }
 
--- a/packages/compiler-cli/ngcc/src/main.ts
+++ b/packages/compiler-cli/ngcc/src/main.ts
@@ -231,7 +231,7 @@ export function mainNgcc(
           unprocessableEntryPointPaths.map(path => `\n  - ${path}`).join(''));
     }
 
-    const duration = Math.round((Date.now() - startTime) / 1000);
+    const duration = Math.round((Date.now() - startTime) / 100) / 10;
     logger.debug(
         `Analyzed ${entryPoints.length} entry-points in ${duration}s. ` +
         `(Total tasks: ${tasks.length})`);
--- a/packages/compiler-cli/ngcc/src/main.ts
+++ b/packages/compiler-cli/ngcc/src/main.ts
@@ -119,13 +119,11 @@ export interface SyncNgccOptions {
   enableI18nLegacyMessageIdFormat?: boolean;
 
   /**
-   * Whether to read an entry-point manifest file from disk, rather than walking the directory tree
-   * looking for entry-points.
+   * Whether to invalidate any entry-point manifest file that is on disk. Instead, walk the
+   * directory tree looking for entry-points, and then write a new entry-point manifest, if
+   * possible.
    *
-   * If `true` then any manifest that exists will be replaced with a new one created from walking
-   * the directory tree.
-   *
-   * Default: `false` (i.e. the manifest will be read)
+   * Default: `false` (i.e. the manifest will be used if available)
    */
   invalidateEntryPointManifest?: boolean;
 }
--- a/packages/compiler-cli/ngcc/src/main.ts
+++ b/packages/compiler-cli/ngcc/src/main.ts
@@ -39,7 +39,7 @@ import {hasBeenProcessed} from './packages/build_marker';
 import {NgccConfiguration} from './packages/configuration';
 import {EntryPoint, EntryPointJsonProperty, EntryPointPackageJson, SUPPORTED_FORMAT_PROPERTIES, getEntryPointFormat} from './packages/entry_point';
 import {makeEntryPointBundle} from './packages/entry_point_bundle';
-import {EntryPointManifest} from './packages/entry_point_manifest';
+import {EntryPointManifest, NullEntryPointManifest} from './packages/entry_point_manifest';
 import {Transformer} from './packages/transformer';
 import {PathMappings} from './utils';
 import {cleanOutdatedPackages} from './writing/cleaning/package_cleaner';
@@ -117,6 +117,17 @@ export interface SyncNgccOptions {
    * legacy message ids will all be stripped during translation.
    */
   enableI18nLegacyMessageIdFormat?: boolean;
+
+  /**
+   * Whether to read an entry-point manifest file from disk, rather than walking the directory tree
+   * looking for entry-points.
+   *
+   * If `true` then any manifest that exists will be replaced with a new one created from walking
+   * the directory tree.
+   *
+   * Default: `false` (i.e. the manifest will be read)
+   */
+  invalidateEntryPointManifest?: boolean;
 }
 
 /**
@@ -139,11 +150,12 @@ export type NgccOptions = AsyncNgccOptions | SyncNgccOptions;
  */
 export function mainNgcc(options: AsyncNgccOptions): Promise<void>;
 export function mainNgcc(options: SyncNgccOptions): void;
-export function mainNgcc(
-    {basePath, targetEntryPointPath, propertiesToConsider = SUPPORTED_FORMAT_PROPERTIES,
-     compileAllFormats = true, createNewEntryPointFormats = false,
-     logger = new ConsoleLogger(LogLevel.info), pathMappings, async = false,
-     enableI18nLegacyMessageIdFormat = true}: NgccOptions): void|Promise<void> {
+export function mainNgcc({basePath, targetEntryPointPath,
+                          propertiesToConsider = SUPPORTED_FORMAT_PROPERTIES,
+                          compileAllFormats = true, createNewEntryPointFormats = false,
+                          logger = new ConsoleLogger(LogLevel.info), pathMappings, async = false,
+                          enableI18nLegacyMessageIdFormat = true,
+                          invalidateEntryPointManifest = false}: NgccOptions): void|Promise<void> {
   // Execute in parallel, if async execution is acceptable and there are more than 1 CPU cores.
   const inParallel = async && (os.cpus().length > 1);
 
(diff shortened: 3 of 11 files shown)


packages/compiler-cli/ngcc/src/packages/entry_point_manifest.ts
```diff
@@ -0,0 +1,139 @@
...
+import {AbsoluteFsPath, FileSystem} from '../../../src/ngtsc/file_system';
+import {Logger} from '../logging/logger';
+
+import {NGCC_VERSION} from './build_marker';
+import {NgccConfiguration} from './configuration';
+import {EntryPoint, INVALID_ENTRY_POINT, NO_ENTRY_POINT, getEntryPointInfo} from './entry_point';
+
+/**
+ * Manages reading a write a manifest file that contains a list of all the entry-points that were
+ * found below a give basePath.
```
JoostK:
```suggestion
 * Manages reading and writing a manifest file that contains a list of all the entry-points that were
 * found below a given basePath.
```

packages/compiler-cli/ngcc/src/packages/entry_point_manifest.ts
```diff
@@ -0,0 +1,139 @@
...
+ *
+ * This is as super-set of the entry-points that are actually processed for a given run of ngcc,
```
JoostK:
```suggestion
 * This is a super-set of the entry-points that are actually processed for a given run of ngcc,
```

packages/compiler-cli/ngcc/src/packages/entry_point_manifest.ts
```diff
@@ -0,0 +1,139 @@
...
+            getEntryPointInfo(this.fs, this.config, this.logger, packagePath, entryPointPath);
+        if (result === NO_ENTRY_POINT || result === INVALID_ENTRY_POINT) {
+          throw new Error(
+              `The entry-point manifest at ${manifestPath} contained an invalid pair of package paths: [${packagePath}, ${entryPointPath}]`);
+        } else {
+          entryPoints.push(result);
+        }
+      }
+      const duration = Math.round((Date.now() - startTime) / 100) / 10;
+      this.logger.debug(`Reading entry-points directly took ${duration}s.`);
```
JoostK: nit: Mentioning _manifest_ here is more clear than _directly_

```suggestion
      this.logger.debug(`Reading entry-points from manifest took ${duration}s.`);
```

packages/compiler-cli/ngcc/src/packages/entry_point_manifest.ts
```diff
@@ -0,0 +1,139 @@
...
+      return entryPoints;
+    } catch (e) {
+      this.logger.warn(`Unable to read the entry-point manifest for ${basePath}`, e.toString());
```
JoostK: I would include a colon before the error message:

```suggestion
      this.logger.warn(`Unable to read the entry-point manifest for ${basePath}: ${e.toString()}`);
```

packages/compiler-cli/ngcc/src/packages/entry_point_manifest.ts
```diff
@@ -0,0 +1,139 @@
...
+    if (lockFileHash === null) {
+      return;
+    }
+    const manifest: EntryPointManifestFile = {
+      ngccVersion: NGCC_VERSION,
+      configFileHash: this.config.hash,
+      lockFileHash: lockFileHash,
+      entryPointPaths: entryPoints.map(entryPoint => [entryPoint.package, entryPoint.path]),
+    };
+    this.fs.writeFile(this.getEntryPointManifestPath(basePath), JSON.stringify(manifest));
```
JoostK: Consider writing this as readable JSON:
```suggestion
    this.fs.writeFile(this.getEntryPointManifestPath(basePath), JSON.stringify(manifest, null, 2));
```

packages/compiler-cli/ngcc/src/packages/entry_point_manifest.ts
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  private getEntryPointManifestPath(basePath: AbsoluteFsPath) {
+    return this.fs.resolve(basePath, '__ngcc_entry_points__.json');
+  }
+
+  private computeLockFileHash(basePath: AbsoluteFsPath): string|null {
```
JoostK: Should this be cached? Otherwise it would have to be computed twice if there's no manifest file.

packages/compiler-cli/ngcc/test/packages/entry_point_manifest_spec.ts
```diff
@@ -0,0 +1,238 @@
...
+        fs.writeFile(
+            _Abs('/project/node_modules/__ngcc_entry_points__.json'), JSON.stringify(manifestFile));
+        const entryPoints = manifest.readEntryPointsUsingManifest(_Abs('/project/node_modules'));
+        expect(entryPoints).toBe(null);
+      });
+
+      it('should return null if the lockfile hash does not match', () => {
+        fs.ensureDir(_Abs('/project/node_modules'));
+        fs.writeFile(_Abs('/project/yarn.lock'), 'LOCK FILE CONTENTS');
+        manifestFile.lockFileHash = 'bad-version';
```
JoostK:
```suggestion
        manifestFile.lockFileHash = 'bad-hash';
```

packages/compiler-cli/ngcc/test/packages/entry_point_manifest_spec.ts
```diff
@@ -0,0 +1,238 @@
...
+        fs.writeFile(
+            _Abs('/project/node_modules/__ngcc_entry_points__.json'), JSON.stringify(manifestFile));
+        const entryPoints = manifest.readEntryPointsUsingManifest(_Abs('/project/node_modules'));
+        expect(entryPoints).toBe(null);
+      });
+
+      it('should return an array of entry-points if all the checks pass', () => {
```
JoostK: Could you duplicate this test for `package-lock.json`?

packages/compiler-cli/ngcc/test/packages/entry_point_manifest_spec.ts
```diff
@@ -0,0 +1,238 @@
...
+        fs.ensureDir(_Abs('/project/node_modules'));
+        fs.writeFile(_Abs('/project/yarn.lock'), 'LOCK FILE CONTENTS');
+        manifestFile.configFileHash = 'bad-hash';
+        fs.writeFile(
+            _Abs('/project/node_modules/__ngcc_entry_points__.json'), JSON.stringify(manifestFile));
+        const entryPoints = manifest.readEntryPointsUsingManifest(_Abs('/project/node_modules'));
+        expect(entryPoints).toBe(null);
+      });
+
+      it('should return null if the lockfile hash does not match', () => {
```
JoostK: Could you duplicate this test for `package-lock.json`?

packages/compiler-cli/ngcc/src/packages/entry_point_manifest.ts
petebacondarwin: Strictly the entry-points were not read from the manifest, only their paths are read from the manifest. We still read the entry-point info from the files on disk.

packages/compiler-cli/ngcc/src/packages/entry_point_manifest.ts
petebacondarwin: I avoided this because:

a) It is not supposed to be read by humans. And if they want to read it, it is trivial to format it in an editor
b) These files can get quite large. Adding extra whitespace slows down the stringification/parsing for now programmatic benefit. Given that this change is supposed to be for performance reasons that seemed a strange thing to do.

packages/compiler-cli/ngcc/src/packages/entry_point_manifest.ts
petebacondarwin: That is an interesting idea...

* In the case we are trying to optimise (when the manifest does exist) there would be no benefit.
* In the other case it would speed things up. By how much I don't know:
  * The FileSystem will have cached the contents of the file so there would be no actual file access
  * I don't know how time consuming the hash computation is but I expect it is done in C++ code so is probably pretty fast.

I am a bit worried about memoizing this sort of thing... What happens if there happens to be two separate `basePath`s that are `node_modules` in the processing - we would need to store a lookup of lockfile hash per basePath. The complexity seems over-kill.

I will run a few performance checks to see what the cost of this is...

packages/compiler-cli/ngcc/src/packages/entry_point_manifest.ts
petebacondarwin:
```suggestion
      this.logger.debug(`Reading entry-points using the manifest entries took ${duration}s.`);
```

JoostK: Oh, yeah that is true. I was thinking that if I were to read that debug message, I wouldn't know what _directly_ would actually mean.

packages/compiler-cli/ngcc/src/packages/entry_point_manifest.ts
petebacondarwin: On my laptop, running in `ngcc -l debug` in the `aio` folder I get:

```
Debug: Analyzing entry-points...
computeLockFileHash: 2.956ms
Debug: No manifest found for /Users/pete/dev/angular/tmp/ngcc-analysis-manifest/aio/node_modules so walking the directories for entry-points.
...
Debug: Walking directories took 1.4s.
computeLockFileHash: 0.919ms
```

And then when running a second time:

```
Debug: Analyzing entry-points...
computeLockFileHash: 2.769ms
Debug: Entry-point manifest found for /Users/pete/dev/angular/tmp/ngcc-analysis-manifest/aio/node_modules so loading entry-point information directly.
Debug: Reading entry-points using the manifest entries took 0.1s.
```

So it seems that we would, at the most, save 2ms by caching the hash, which I think we can agree is not worth the additional complexity.

JoostK: That's not worth the complexity, indeed! Thanks for looking into it 👍

JoostK (approved):

petebacondarwin:
Marking for merge but waiting to see if @gkalpak or @alxhub want to comment.

petebacondarwin:
Add an `--invalidate-manifest` flag

petebacondarwin:
Added the new `--invalidate-entry-point-manifest` flag. PTAL @JoostK et al

packages/compiler-cli/ngcc/src/packages/entry_point_manifest.ts
```diff
@@ -128,6 +128,18 @@ export class EntryPointManifest {
   }
 }
 
+/**
+ * A specialized implementation of the `EntryPointManifest` that always returns `null` rather than a
+ * set of entry-points from a manifest file.
+ *
+ * This is used if you want to ignore a manifest file that is written to disc.
```
JoostK:
```suggestion
 * This is used if you want to ignore a manifest file that is written to disk.
```

Or are you holding on to portable media 💿😄

packages/compiler-cli/ngcc/src/packages/entry_point_manifest.ts
```diff
@@ -128,6 +128,18 @@ export class EntryPointManifest {
...
+ * Note that it will still write a new manifest file with the entry-points that were found by
+ * walking the directories.
+ */
+export class NullEntryPointManifest extends EntryPointManifest {
```
JoostK: May I propose `IgnoredEntryPointManifest`, as it's not entirely a null implementation.

```suggestion
export class IgnoredEntryPointManifest extends EntryPointManifest {
```

packages/compiler-cli/ngcc/main-ngcc.ts
```diff
@@ -76,6 +76,13 @@ if (require.main === module) {
             describe: 'The lowest severity logging message that should be output.',
             choices: ['debug', 'info', 'warn', 'error'],
           })
+          .option('invalidate-entry-point-manifest', {
+            describe:
+                'If this is set then ngcc will not read an entry-point manifest file from disk.\n' +
+                'Instead it will walking the directory tree as normal looking for entry-points, and then write a new manifest file.',
```
JoostK:
```suggestion
                'Instead it will walk the directory tree as normal looking for entry-points, and then write a new manifest file.',
```

packages/compiler-cli/ngcc/test/integration/ngcc_spec.ts
```diff
@@ -982,6 +982,39 @@ runInEachFileSystem(() => {
...
+          basePath: '/node_modules',
+          propertiesToConsider: ['esm2015'],
+          logger: new MockLogger(),
+          invalidateEntryPointManifest: true,
+        });
+        // Check that common/testing ES2015 is now processed, despite the manifest not listing it
+        commonTesting =
+            JSON.parse(fs.readFile(_('/node_modules/@angular/common/testing/package.json')));
+        expect(hasBeenProcessed(commonTesting, 'esm5')).toBe(true);
+        expect(hasBeenProcessed(commonTesting, 'esm2015')).toBe(true);
```
JoostK: Consider adding an assertion that the manifest file has been updated with the newly scanned entry-points.

packages/compiler-cli/ngcc/src/packages/entry_point_manifest.ts
```diff
@@ -0,0 +1,151 @@
...
+import {EntryPoint, INVALID_ENTRY_POINT, NO_ENTRY_POINT, getEntryPointInfo} from './entry_point';
+
+/**
+ * Manages reading and write a manifest file that contains a list of all the entry-points that were
+ * found below a given basePath.
+ *
+ * This is a super-set of the entry-points that are actually processed for a given run of ngcc,
+ * since some may already be processed, or excluded if they do not have the required format.
+ */
+export class EntryPointManifest {
```
alan-agius4: Nit: for consistence, should you always use `EntryPoints` or `EntryPoint`

- readEntryPointsUsingManifest
- writeEntryPointManifest
- getEntryPointManifestPath
- invalidateEntryPointManifest

I think case, probably it's probably it's prefer the plural versions, since you have a manfest which contains entry-points data.

packages/compiler-cli/ngcc/src/packages/entry_point_manifest.ts
petebacondarwin: Can I counter propose `InvalidatingEntryPointManifest`, since that is what it is doing?

packages/compiler-cli/ngcc/src/packages/entry_point_manifest.ts
petebacondarwin: Actually I believe it is already consistent. The manifest is an "entry-point manifest" (like at school you might have a "student register" not a "students register"). It is referred in this way consistently throughout, unless I have missed something. I feel the name `EntryPointsManifest` is overly verbose.

The only difference is the `readEntryPointsUsingManifest()` method but this also is consistent: it is reading "entry-points" using the "entry-point manifest" as a shortcut.

Let me know if there are other places that are inconsistent with this naming.

packages/compiler-cli/ngcc/test/integration/ngcc_spec.ts
petebacondarwin: Done

packages/compiler-cli/ngcc/src/packages/entry_point_manifest.ts
JoostK: Sure, go ahead 👍

alxhub (approved):

packages/compiler-cli/ngcc/src/packages/entry_point_manifest.ts
alan-agius4: Thanks for the detailed explanation.

packages/compiler-cli/ngcc/src/packages/entry_point_manifest.ts
```diff
@@ -0,0 +1,139 @@
...
+
+  /**
+   * Try to get the entry-point info from a manifest file for the given `basePath` if it exists and
+   * is not out of date.
+   *
+   * Reasons for the manifest to be out of date are:
+   *
+   * * the file does not exist
+   * * the ngcc version has changed
+   * * the package lock-file (i.e. yarn.lock or package-lock.json) has changed
```
gkalpak: Wouldn't `project` (instead of `package`) be more consistent with how we use it when referring to ngcc configurations?
(Here and in other method docs below.)

packages/compiler-cli/ngcc/src/packages/entry_point_manifest.ts
```diff
@@ -0,0 +1,139 @@
...
+  readEntryPointsUsingManifest(basePath: AbsoluteFsPath): EntryPoint[]|null {
+    try {
+      if (this.fs.basename(basePath) !== 'node_modules') {
+        return null;
+      }
+
+      const computedLockFileHash = this.computeLockFileHash(basePath);
+      if (computedLockFileHash === null) {
+        return null;
+      }
```
gkalpak: Nit: I would move this below the `mnifestPath` check, because this check it more expensive and less likely to fail (if the next one succeeds).

packages/compiler-cli/ngcc/src/packages/entry_point_manifest.ts
```diff
@@ -0,0 +1,139 @@
...
+   * Try to get the entry-point info from a manifest file for the given `basePath` if it exists and
+   * is not out of date.
+   *
+   * Reasons for the manifest to be out of date are:
+   *
+   * * the file does not exist
+   * * the ngcc version has changed
+   * * the package lock-file (i.e. yarn.lock or package-lock.json) has changed
+   * * the project configuration has changed
+   * * one or more entry-points in the manifest are not valid
```
gkalpak: AFAICT, this case results in an error being thrown (not the manifest being considered "out-of-date").

packages/compiler-cli/ngcc/src/packages/entry_point_manifest.ts
```diff
@@ -0,0 +1,139 @@
...
+              `The entry-point manifest at ${manifestPath} contained an invalid pair of package paths: [${packagePath}, ${entryPointPath}]`);
+        } else {
+          entryPoints.push(result);
+        }
+      }
+      const duration = Math.round((Date.now() - startTime) / 100) / 10;
+      this.logger.debug(`Reading entry-points using the manifest entries took ${duration}s.`);
+      return entryPoints;
+    } catch (e) {
+      this.logger.warn(`Unable to read the entry-point manifest for ${basePath}:`, e.toString());
```
gkalpak: Why not just pass `e`? `e.toString()` loses the stack trace info, which might be helpful in debugging.

packages/compiler-cli/ngcc/src/main.ts
```diff
@@ -117,6 +117,17 @@ export interface SyncNgccOptions {
    * legacy message ids will all be stripped during translation.
    */
   enableI18nLegacyMessageIdFormat?: boolean;
+
+  /**
+   * Whether to read an entry-point manifest file from disk, rather than walking the directory tree
+   * looking for entry-points.
```
gkalpak: This reads backwards to me 😇
How about inverting it:

```suggestion
   * Whether to walk the directory tree looking for entry-points rather than trying to read an
   * entry-point manifest file from disk.
```

gkalpak (approved):
I am a little late to the party, but since I had started the review yesterday 🤷‍♂
Nothing blocking, so feel free to just add the `merge` label back without addressing my comments 😇

packages/compiler-cli/ngcc/src/execution/cluster/master.ts
```diff
@@ -70,7 +70,7 @@ export class ClusterMaster {
 
     // First, check whether all tasks have been completed.
     if (this.taskQueue.allTasksCompleted) {
-      const duration = Math.round((Date.now() - this.processingStartTime) / 1000);
+      const duration = Math.round((Date.now() - this.processingStartTime) / 100) / 10;
```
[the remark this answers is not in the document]
petebacondarwin: Feel free to do a refactor PR after this lands to clean this up.

packages/compiler-cli/ngcc/src/packages/entry_point_manifest.ts
petebacondarwin: By "package lock-file" I wanted to indicate that we are talking about file that locks down the package versions - rather than our other lock-file which is used to lock down the current running instance of ngcc. IMO project would be even more ambiguous.

packages/compiler-cli/ngcc/src/packages/entry_point_manifest.ts
petebacondarwin: `Logger.warn()` expects strings

petebacondarwin: We could do `e.stack` possibly, falling back on `e.toString()` if there is none.

packages/compiler-cli/ngcc/src/main.ts
petebacondarwin: Yes, looking back this doesn't read well.

packages/compiler-cli/ngcc/src/packages/entry_point_manifest.ts
petebacondarwin: Note the big try-catch block around the whole function. It never throws but returns `null` if there was an error.

packages/compiler-cli/ngcc/src/main.ts
petebacondarwin: Going for:

```
   * Whether to invalidate any entry-point manifest file that is on disk. Instead, walk the
   * directory tree looking for entry-points, and then write a new entry-point manifest, if
   * possible.
```

petebacondarwin:
Thanks @gkalpak .

packages/compiler-cli/ngcc/src/packages/entry_point_manifest.ts
gkalpak: Ooops, missed that :grin:

PR #35932, opened 2020-03-07
refactor(compiler): add `@nocollapse` annotation using a synthetic comment

JoostK:
In Ivy, Angular decorators are compiled into static fields that are
inserted into a class declaration in a TypeScript transform. When
targeting Closure compiler such fields need to be annotated with
`@nocollapse` to prevent them from being lifted from a static field into
a variable, as that would prevent the Ivy runtime from being able to
find the compiled definitions.

Previously, there was a bug in TypeScript where synthetic comments added
in a transform would not be emitted at all, so as a workaround a global
regex-replace was done in the emit's `writeFile` callback that would add
the `@nocollapse` annotation to all static Ivy definition fields. This
approach is no longer possible when ngtsc is running as TypeScript
plugin, as a plugin cannot control emit behavior.

The workaround is no longer necessary, as synthetic comments are now
properly emitted, likely as of
https://github.com/microsoft/TypeScript/pull/22141 which has been
released with TypeScript 2.8.

This change is required for running ngtsc as TypeScript plugin in
Bazel's `ts_library` rule, to move away from the custom `ngc_wrapped`
approach.

Resolves FW-1952

---

Note: marking for merge to master only, as patch would likely not benefit from this change.


alxhub (approved):
You can revert the non-ngtsc part of this change - we don't maintain the old implementation of the Ivy compiler inside src/transformers/.... We honestly should just delete it.

alxhub:
Oh, I do think we should merge this to `patch` too - it'll keep the trees more in sync, and it could result in a performance improvement.

alxhub:
[Presubmit](https://test.corp.google.com/ui#id=OCL:299880004:BASE:299880080:1583776231005:db71e24f)

alxhub:
[Presubmit](https://test.corp.google.com/ui#id=OCL:299959293:BASE:299959334:1583795252778:765560f2) without an unrelated failing commit

alxhub:
[Fresh presubmit](http://test/OCL:299959293:BASE:302735399:1585081881280:f3e346ee)

(the old one seemed fine, just an abundance of precaution)

PR #35933, opened 2020-03-07
ci: exclude ngcc directory from fw-compiler pull-approve rule

petebacondarwin:
ngcc has its own pull-approve group

### DIFF
--- a/.pullapprove.yml
+++ b/.pullapprove.yml
@@ -123,10 +123,10 @@ pullapprove_conditions:
 groups:
   # =========================================================
   #  Global Approvers
-  # 
+  #
   # All reviews performed for global approvals require using
   # the `Reviewed-for:` specifier to set the approval
-  # specificity as documented at: 
+  # specificity as documented at:
   # https://docs.pullapprove.com/reviewed-for/
   # =========================================================
   global-approvers:
@@ -141,10 +141,10 @@ groups:
 
   # =========================================================
   #  Global Approvers For Docs
-  # 
+  #
   # All reviews performed for global docs approvals require
   # using the `Reviewed-for:` specifier to set the approval
-  # specificity as documented at: 
+  # specificity as documented at:
   # https://docs.pullapprove.com/reviewed-for/
   # =========================================================
   global-docs-approvers:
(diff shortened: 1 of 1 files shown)


.pullapprove.yml
```diff
@@ -155,6 +155,7 @@ groups:
           'aio/content/guide/aot-metadata-errors.md',
           'aio/content/guide/template-typecheck.md '
           ])
+      - "files.exclude('ngcc')"
```
IgorMinar: is it just `'ngcc'` or `'*ngcc*'`?

the docs are not clear: https://docs.pullapprove.com/config/conditions/ 

// @josephperrott

IgorMinar (approved):
lgtm if we can get the confirmation on how the argument to `exclude` behaves.

.pullapprove.yml
josephperrott: I can't explain why exactly, but the current change doesn't work as expected.

When I tested it, the most readable option I found is
```yaml
fw-compiler:
  conditions:
    - >
      contains_any_globs(files, [
        'packages/compiler/**',
        'packages/examples/compiler/**',
        'packages/compiler-cli/**',
        'aio/content/guide/angular-compiler-options.md',
        'aio/content/guide/aot-compiler.md',
        'aio/content/guide/aot-metadata-errors.md',
        'aio/content/guide/template-typecheck.md '
        ])
    - >
      not contains_any_globs(files, [
        'packages/compiler-cli/ngcc/**'
        ])
```

This also will be _relatively_ easy to manage in our verification script, rather than having an entirely new method of matching to mimic.

petebacondarwin:
@josephperrott - is this now unblocked?

josephperrott:
Sorry, yes this is unblocked.

.pullapprove.yml
josephperrott: Now that we are probably handling this in our pullapprove verification tooling, it will verify it is as expected.  However we should move to using the `contains_any_globs` for our exclude lists as well.

Can we change 
```
    - "files.exclude('ngcc')"
```

to 

```
    - >
      not contains_any_globs(files, [
        'packages/compiler-cli/ngcc/**'
        ])
```

josephperrott (changes requested):

josephperrott (approved):
LGTM

PR #35861, opened 2020-03-04
fix(ngcc): re-enable killing ngcc (and CLI) via Ctrl-C

petebacondarwin:
Fixes #35761

For the moment I have left the old implementation in there too. We should probably remove it either as part of this PR or soonish, unless we decide that we want to make it configurable and allow people the different options.

PR #35934, opened 2020-03-07
Ngcc better lockfile patch

petebacondarwin:
This is a patch version of https://github.com/angular/angular/pull/35861

It does not include any `AsyncLocker` functionality but otherwise is the same.

Probably still needs a review though.


packages/compiler-cli/ngcc/src/execution/single_process_executor.ts
```diff
@@ -10,47 +10,56 @@ import {Logger} from '../logging/logger';
...
-export class AsyncSingleProcessExecutor extends SingleProcessExecutor {
-  async execute(...args: Parameters<Executor['execute']>): Promise<void> {
-    return super.execute(...args);
+export class SingleProcessExecutorAsync extends SingleProcessorExecutorBase implements Executor {
+  constructor(logger: Logger, pkgJsonUpdater: PackageJsonUpdater, private lockFile: SyncLocker) {
+    super(logger, pkgJsonUpdater);
+  }
+  async execute(analyzeEntryPoints: AnalyzeEntryPointsFn, createCompileFn: CreateCompileFn):
+      Promise<void> {
+    await this.lockFile.lock(async() => this.doExecute(analyzeEntryPoints, createCompileFn));
```
JoostK: I think it's strange to pass an async fn to the synchronous locker, given that the `SyncLocker` does not await its result before unlocking the lock file. That's actually fine because all work still executes synchronously, but I think it should be enough to have just the `async` marker on the `execute` method as that ought to be sufficient to achieve asynchronous completion.

```suggestion
    this.lockFile.lock(() => this.doExecute(analyzeEntryPoints, createCompileFn));
```

JoostK (commented):
I just scanned through most of this, given that it has already landed in master, but I do have one comment though. Also, the commit message of ``separate `SyncLocker` and `LockFile` `` refers to "Apart from avoiding the awkward base" but it doesn't look like that was the case in the patch branch.

petebacondarwin:
@Joostk PTAL

packages/compiler-cli/ngcc/src/execution/lock_file.ts
```diff
@@ -119,3 +108,56 @@ export class LockFile {
     process.exit(code);
   }
 }
+
+/**
+ * SyncLocker is used to prevent more than one instance of ngcc executing at the same time,
+ * when being called in a synchronous context.
+ *
+ * * When ngcc starts executing, it creates a file in the `compiler-cli/ngcc` folder.
```
JoostK: Ha, these asterisks are duplicated but that's actually also the case in master ;)

packages/compiler-cli/ngcc/src/locking/lock_file_with_signal_handlers.ts
```diff
@@ -0,0 +1,83 @@
...
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import * as process from 'process';
+import {CachedFileSystem, FileSystem} from '../../../src/ngtsc/file_system';
+import {LockFile, getLockFilePath} from './lock_file';
+
+export class LockFileWithSignalHandlers implements LockFile {
```
JoostK: This thing isn't used anymore as the child process strategy is used unconditionally. The same is true for master, though, so I suggest not to include this in the patch branch and then delete it in master exclusively.

packages/compiler-cli/ngcc/src/execution/single_process_executor.ts
petebacondarwin: Actually it seems that these are actually needed. CI errors:

```
ERROR: /home/circleci/ng/packages/compiler-cli/ngcc/BUILD.bazel:5:1: Compiling TypeScript (devmode) //packages/compiler-cli/ngcc:ngcc failed (Exit 1)
packages/compiler-cli/ngcc/src/execution/single_process_executor.ts:63:5 - error TS21225: [tsetse] All Promises in async functions must either be awaited or used in an expression.
	See http://tsetse.info/must-use-promises

63     this.lockFile.lock(() => this.doExecute(analyzeEntryPoints, createCompileFn));
```

packages/compiler-cli/ngcc/src/execution/lock_file.ts
petebacondarwin: Actually they are bullet points!

packages/compiler-cli/ngcc/src/locking/lock_file_with_signal_handlers.ts
petebacondarwin: Yeah, I left it in both initially in case we decided that we needed to make it configurable.
But once we are happy that we won't go back to this version we can create simple PRs for master and patch to remove it.

petebacondarwin: Here is the PR for master: https://github.com/angular/angular/pull/35938
I will just cherry-pick that commit directly into this PR for patch.

packages/compiler-cli/ngcc/src/execution/single_process_executor.ts
JoostK: I guess that is because `doExecute` has a return type of `void|Promise<void>`, whereas it never returns a Promise in reality (that would be an issue for the sync variant). If that return type was changed into just `void`, the `SyncLocker.lock` call would also return `void` so tsetse would no longer flag it as needing to be awaited.

JoostK (approved):

petebacondarwin:
Merge-assistance: I am told by the `dev-infra` team that the `ci/angular: size` failure is not valid: as can be seen by the fact that the size increase is the same as the entire file, which indicates that the expected file-size is 0 bytes!

IgorMinar (approved):
lgtm - the CI failure is invalid, we should proceed with the merge and fix the infra separately.

AndrewKushnir:
This PR is now merged into `9.0.x` branch.

PR #35943, opened 2020-03-08
perf(language-service): cache variable and event symbols found in path

ayazhafiz:
When computing the diagnostics for a template, the language service
traverses through all paths in the template and computes diagnostics at
each location an Angular expression is found. The context of the
expression to be diagnosed is generated partially by examining the
variables, references, and events introduced by template nodes in a
current path. Today, this context is re-computed everytime an expression
is evaluated. This often results in a wasteful re-computation of data
for template nodes whose scope contribution is already known. For
example, in the template

```html
<main #mysection>
  {{ title }}
  {{ subtitle }}
</main>
```

The expression scope is computed twice, once for each interpolation. But
because the template is static, the contribution of the element "main"
to the expression scope should only need to be computed once.

To reduce unnecessary work in computing the expression scope, cache
references, variables, and events found in template path nodes.

Note that calculation of expression scope is likely to be one of the
hottest paths in the language service aside actual parsing of the
template. I have not benchmarked this change, but I expect at least a
minor performance improvement.

## PR Checklist

## PR Type

- [ ] Yes
- [x] No

### DIFF
--- a/packages/language-service/src/expression_diagnostics.ts
+++ b/packages/language-service/src/expression_diagnostics.ts
@@ -242,7 +242,7 @@ function getEventDeclaration(
     })();
 
     cache.set(event, eventDeclaration);
-  };
+  }
 
   return cache.get(event);
 }
--- a/packages/language-service/src/expression_diagnostics.ts
+++ b/packages/language-service/src/expression_diagnostics.ts
@@ -24,8 +24,10 @@ export interface DiagnosticTemplateInfo {
 }
 
 export function getTemplateExpressionDiagnostics(info: DiagnosticTemplateInfo): ng.Diagnostic[] {
+  const variableCache = new Map<TemplateAst, SymbolDeclaration[]>();
+  const eventCache = new Map<BoundEventAst, SymbolDeclaration>();
   const visitor = new ExpressionDiagnosticsVisitor(
-      info, (path: TemplateAstPath) => getExpressionScope(info, path));
+      info, (path: TemplateAstPath) => getExpressionScope(info, path, variableCache, eventCache));
   templateVisitAll(visitor, info.templateAst);
   return visitor.diagnostics;
 }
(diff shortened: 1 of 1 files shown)


ayazhafiz:
Edit: I was able to observe up to a ~20% speedup with a template like

```html
<form #f="ngForm" (ngSubmit)="onSubmit(f)" novalidate>
  <input name="first" ngModel required #first="ngModel">
  <input name="last" ngModel>
  <button>Submit</button>
</form>
<div #div>
  <span #span>My label</span>
  <button (ngSubmit)="onSubmit(f)">
    {{ span.innerText }}
  </button>
</div>
<p>First name value: {{ first.value }}</p>
<p>First name valid: {{ first.valid }}</p>
<p>Form value: {{ f.value | json }}</p>
<p>Form valid: {{ f.valid }}</p>

<div fxLayout="row" fxLayoutGap="20px grid" fxLayoutAlign="space-evenly center" fxFlex style="text-align: center;">
  <div *ngFor="let table of title; let i = index" style='cursor: pointer;'>
      <div *ngIf='span.innerText.length == 0' style='text-align: center;'> Aucune table n'est disponnible dans cette base de donnée </div>
      <div *ngIf='span.innerText.length != 0'> <img src='assets/images/table.png' width="50px"><br /> {{span.innerText}} </div>
      <div *ngIf="i == 7" class='break'></div>
  </div>
</div>

<ng-container *ngFor="let t of div.innerText; index as i">
    {{ i.as + 'a' }}
    {{ t.aasdasd }}
</ng-container>

<ng-container *ngIf="div.innerText; let title;">
   {{ title.foo() }}
</ng-container>
```

| | Time |
|--|--|
|Before|29.847850918769836 ms|
|After|24.013825058937073 ms|

packages/language-service/src/expression_diagnostics.ts
```diff
@@ -225,12 +255,19 @@ function getEventDeclaration(
  * @param path path of template nodes narrowing to the context the expression scope should be
  * derived for.
  */
-export function getExpressionScope(
-    info: DiagnosticTemplateInfo, path: TemplateAstPath): SymbolTable {
+export function getExpressionScope(info: DiagnosticTemplateInfo, path: TemplateAstPath, cache: {
+  references: Map<TemplateAst, SymbolDeclaration[]>,
```
kyliau: Can we consolidate all of these into a single `Map<TemplateAst, SymbolDeclaration[]>`?
If not, I think it is better if we create a separate class to represent this entire cache as a single entity.

ayazhafiz: I think there is an even better way to do this. We can keep a dirty flag in the expression diagnostics visitor to check if ASTs in a path have not been visited before. If they have not, we look up the diagnostic scope the contribute and merge it with the diagnostics of the existing nodes in the path.

I will look into this more, but before I do that I will refactor the getExpressionScope method to a new file and add tests specifically for it.

PR #35962, opened 2020-03-09
test(core): re-enable test on Sauce Labs

pkozlowski-opensource:
I was not able to reproduce IE 10/11 failure of the disabled
tests on SauceLabs any more. I did some cleanup of the test
in question but I doubt it was the root cause of the problem.


devversion (approved):
I tried locally on my IE11 version but wasn't able to reproduce the failure too. LGTM

alxhub (approved):

PR #35964, opened 2020-03-09
refactor(core): Take advantage of 'assert functions' for `ngDevMode` asserts

mhevery:
As of TypeScript 3.7, TypeScript supports [Assert Functions](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions). This change adds assert types to our `assert*` functions.

We can't fully take advantage of this due to [Assert functions do not constraint type when they are guarded by a truthy expression.](https://github.com/microsoft/TypeScript/issues/37295)

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/render3/assert.ts
+++ b/packages/core/src/render3/assert.ts
@@ -15,6 +15,10 @@ import {TNode} from './interfaces/node';
 import {isLContainer, isLView} from './interfaces/type_checks';
 import {LView, TVIEW, TView} from './interfaces/view';
 
+// [Assert functions do not constraint type when they are guarded by a truthy
+// expression.](https://github.com/microsoft/TypeScript/issues/37295)
+
+
 export function assertTNodeForLView(tNode: TNode, lView: LView) {
   tNode.hasOwnProperty('tView_') && assertEqual(
                                         (tNode as any as{tView_: TView}).tView_, lView[TVIEW],
--- a/packages/core/src/render3/assert.ts
+++ b/packages/core/src/render3/assert.ts
@@ -85,7 +85,7 @@ export function assertFirstUpdatePass(tView: TView, errMessage?: string) {
  * This is a basic sanity check that an object is probably a directive def. DirectiveDef is
  * an interface, so we can't do a direct instanceof check.
  */
-export function assertDirectiveDef(obj: any): asserts obj is DirectiveDef<any> {
+export function assertDirectiveDef<T>(obj: any): asserts obj is DirectiveDef<T> {
   if (obj.type === undefined || obj.selectors == undefined || obj.inputs === undefined) {
     throwError(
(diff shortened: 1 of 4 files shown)


packages/core/src/render3/assert.ts
```diff
@@ -82,7 +85,7 @@ export function assertFirstUpdatePass(tView: TView, errMessage?: string) {
  * This is a basic sanity check that an object is probably a directive def. DirectiveDef is
  * an interface, so we can't do a direct instanceof check.
  */
-export function assertDirectiveDef(obj: any) {
+export function assertDirectiveDef(obj: any): asserts obj is DirectiveDef<any> {
```
pkozlowski-opensource: What do you think of adding a generic type for directive type: `assertDirectiveDef<T>(obj: any): asserts obj is DirectiveDef<T>` ?

packages/core/src/util/ng_dev_mode.ts
```diff
@@ -23,6 +23,8 @@ declare global {
    * - The URL contains a `ngDevMode=false` text.
    * Finally, ngDevMode may not have been defined at all.
    */
+  // [Assert functions do not constraint type when they are guarded by a truthy
+  // expression.](https://github.com/microsoft/TypeScript/issues/37295)
```
pkozlowski-opensource: Oh, this means that we actually won't see the benefits of those new assert types, right :-/
I wonder if we should keep this comment here (there are multiple places in our code-base where asserts are guarded with a truthy check (`ngDevMode`). Move it as a comment to the assert file?

pkozlowski-opensource (approved):
LGTM with 2 cleanup comments.
The new assert types as super-cool, just sad that we can't take advantage of them...

packages/core/src/util/ng_dev_mode.ts
mhevery: I was concerned that it will just be polluting the codebase as there are many such places. If you feel strongly I can add it there as well.

pkozlowski-opensource: Oh, no, no, definitively don't add it across the code-base!  What I'm suggesting is to _remove_ this comment from here and move it to the packages/core/src/render3/assert.ts file

packages/core/src/render3/assert.ts
mhevery: Added, but I don't think it ads much value, since there is no way for TS to guess `<T>`.

mhevery:
[presubmit](http://test/OCL:300147187:BASE:300147204:1583867274836:85d06a66)

packages/core/src/util/ng_dev_mode.ts
mhevery: ok, done

PR #35971, opened 2020-03-09
build: enable forms tests on IE

crisbeto:
Enables running the `forms` unit tests against IE on Saucelabs. It seems like original failures have been resolved.

**Note:** these tests usually don't run against PRs, but I've verified that they pass by:
1. Running them against IE on my local machine.
2. Enabling them for PRs temporarily (see the first commit).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -894,11 +894,11 @@ workflows:
       # This isn't strictly necessary as there is no artifact dependency, but helps economize
       # CI resources by not attempting to build when we know should fail.
       - test_win:
-          # <<: *skip_on_pull_requests
+          <<: *skip_on_pull_requests
           requires:
             - test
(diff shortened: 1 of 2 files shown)


AndrewKushnir (approved):
LGTM, thanks @crisbeto 👍

AndrewKushnir:
FYI, it looks like @IgorMinar was added by PullApprove because there was a BUILD file modified in the original set of changes to run the tests on windows. The BUILD file was reverted, so I removed Igor from reviewers list. Thank you.

AndrewKushnir:
Sorry, my previous message is incorrect. Igor was added because the `*.bazel` file was changed, so we still need an approval from @IgorMinar (adding "review" label for now).

IgorMinar (approved):
Lgtm

PR #35975, opened 2020-03-09
feat(bazel): transform generated shims (in Ivy) with tsickle

alxhub:
# Take 2 of the PR, with a g3 fix

Currently, when Angular code is built with Bazel and with Ivy, generated
factory shims (.ngfactory files) are not processed via the majority of
tsickle's transforms. This is a subtle effect of the build infrastructure,
but it boils down to a TsickleHost method `shouldSkipTsickleProcessing`.

For ngc_wrapped builds (Bazel + Angular), this method is defined in the
`@bazel/typescript` (aka bazel rules_typescript) implementation of
`CompilerHost`. The default behavior is to skip tsickle processing for files
which are not present in the original `srcs[]` of the build rule. In
Angular's case, this includes all generated shim files.

For View Engine factories this is probably desirable as they're quite
complex and they've never been tested with tsickle. Ivy factories however
are smaller and very straightforward, and it makes sense to treat them like
any other output.

This commit adjusts two independent implementations of
`shouldSkipTsickleProcessing` to enable transformation of Ivy shims:

* in `@angular/bazel` aka ngc_wrapped, the upstream `@bazel/typescript`
  `CompilerHost` is patched to treat .ngfactory files the same as their
  original source file, with respect to tsickle processing.

  It is currently not possible to test this change as we don't have any test
  that inspects tsickle output with bazel. It will be extensively tested in
  g3.

* in `ngc`, Angular's own implementation is adjusted to allow for the
  processing of shims when compiling with Ivy. This enables a unit test to
  be written to validate the correct behavior of tsickle when given a host
  that's appropriately configured to process factory shims.

For ngtsc-as-a-plugin, a similar fix will need to be submitted upstream in
tsc_wrapped.

### DIFF
--- a/packages/compiler-cli/src/ngtsc/shims/src/factory_generator.ts
+++ b/packages/compiler-cli/src/ngtsc/shims/src/factory_generator.ts
@@ -56,21 +56,16 @@ export class FactoryGenerator implements ShimGenerator {
                             .map(decl => decl.name !.text);
 
 
+    let sourceText = '';
+
     // If there is a top-level comment in the original file, copy it over at the top of the
     // generated factory file. This is important for preserving any load-bearing jsdoc comments.
-    let comment: string = '';
-    if (original.statements.length > 0) {
-      const firstStatement = original.statements[0];
-      // Must pass SourceFile to getLeadingTriviaWidth() and getFullText(), otherwise it'll try to
-      // get SourceFile by recursively looking up the parent of the Node and fail,
-      // because parent is undefined.
-      const leadingTriviaWidth = firstStatement.getLeadingTriviaWidth(original);
-      if (leadingTriviaWidth > 0) {
-        comment = firstStatement.getFullText(original).substr(0, leadingTriviaWidth);
-      }
+    const leadingComment = getFileoverviewComment(original);
+    if (leadingComment !== null) {
+      // Leading comments must be separated from the rest of the contents by a blank line.
+      sourceText = leadingComment + '\n\n';
     }
 
-    let sourceText = comment;
     if (symbolNames.length > 0) {
       // For each symbol name, generate a constant export of the corresponding NgFactory.
       // This will encompass a lot of symbols which don't need factories, but that's okay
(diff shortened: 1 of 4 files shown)


alxhub:
[Presubmit](http://test/OCL:299958532:BASE:299958576:1583795052496:8767c4b8)

packages/compiler-cli/src/ngtsc/shims/src/factory_generator.ts
```diff
@@ -248,3 +236,23 @@ function transformFactorySourceFile(
...
   return file;
 }
+
+
+/**
+ * Parses and returns the comment text of a \@fileoverview comment in the given source file. Returns
+ * the empty string if no such comment can be found.
+ */
+function getFileoverviewComment(sourceFile: ts.SourceFile): string {
+  const trivia = sourceFile.getFullText(sourceFile).substring(0, sourceFile.getStart());
```
AndrewKushnir: Can we omit the argument in the `getFullText` call (since we call it on the same instance)? Also it looks like we can call it once and store as a const (we call it 3 times in this function below).

packages/compiler-cli/src/ngtsc/shims/src/factory_generator.ts
```diff
@@ -248,3 +236,23 @@ function transformFactorySourceFile(
...
+  const leadingComments = ts.getLeadingCommentRanges(trivia, 0);
+  if (!leadingComments || leadingComments.length === 0) return '';
+  const comment = leadingComments[0];
+  if (comment.kind !== ts.SyntaxKind.MultiLineCommentTrivia) return '';
+  // Only comments separated with a \n\n from the file contents are considered file-level comments
+  // in TypeScript.
+  if (sourceFile.getFullText().substring(comment.end, comment.end + 2) !== '\n\n') return '';
+  const commentText = sourceFile.getFullText(sourceFile).substring(comment.pos, comment.end);
+  // Closure Compiler ignores @suppress and similar if the comment contains @license.
+  if (commentText.indexOf('@license') !== -1) return '';
```
AndrewKushnir: May be add a quick test to capture this behavior?

AndrewKushnir (approved):
LGTM, thanks for the fixes @alxhub 👍

I've left a couple comments (non-blocking), please have a look when you get a chance.

packages/compiler-cli/src/ngtsc/shims/src/factory_generator.ts
alxhub: I didn't write this method actually - it's copied from VE (which doesn't use the positional argument, but there were some comments about it being needed in our impl so I added it).

packages/compiler-cli/src/ngtsc/shims/src/factory_generator.ts
alxhub: sure thing!

alxhub:
[Presubmit](http://test/OCL:299958532:BASE:300154913:1583869132983:c43c1c0c)

IgorMinar (approved):
lgtm!!

AndrewKushnir:
@alxhub thanks for the fix 👍 

Could you please rebase this PR (so that we also rerun CI) and may be also consider running another presubmit (since the last one was 7 days ago)?

Thank you.

AndrewKushnir:
Discussed with Alex offline: the presubmit should still be valid, since the code touches files that are rarely changed. Thank you.

PR #35976, opened 2020-03-09
test(router): use pageYOffset in testing when scrollY is not available

atscott:
IE 9, 10, and 11 use the non-standard name `pageYOffset` instead of `scrollY`.
https://caniuse.com/#search=scrolly


josephperrott (approved):
LGTM

Issue #31684, opened 2019-04-19
zone-patch-rxjs - Unsubscription error - Cannot read property 'apply' of null


AleG94:
I'm trying to implement a retry http interceptor for Angular 7. 

Without `import 'zone.js/dist/zone-patch-rxjs';` it works fine but when I add it, I get:

```
ERROR {"message":"1 errors occurred during unsubscription:\n1) TypeError: Cannot read property 'apply' of null","name":"UnsubscriptionError","errors":[{}]}
```

This is an example repo to reproduce the issue: 

[https://stackblitz.com/edit/angular-s3iv5u?file=polyfills.ts](https://stackblitz.com/edit/angular-s3iv5u?file=polyfills.ts)

Comment and uncomment the line `import 'zone.js/dist/zone-patch-rxjs';` in `polyfills.ts` to see the error appear and disappear.

arturovt:
@AleG94 just checked the example you've provided with Angular 8 and it works fine. Probably it was fixed but PR didn't mention this issue.

xinbaihui:
I update zone.js from v0.9.1 to v0.10.3  to avoid this error. But I encounter another problem that it won't stop when retry times up to [count] times.
```
return source.pipe(
    retryWhen(errs =>
        errs.pipe(
          take(count),
          mergeMap((err, i) => {
            if (i < count - 1) {  // i === 0 all the time
              return of(err);
            }
            return throwError(err);
          }),
          delay(delayTime)
        )
      )
    );
```

JiaLiPassion:
@xinbaihui, please provide a reproduce repo and a simple step to explain what is the expect result. thanks.

xinbaihui:
This is an example from AleG94, I update zone.js to v0.10.3
[https://stackblitz.com/edit/angular-wmyhre](url)

I hope  the request will be stopped and catch the error after retry 3 times. But actually the retry won't stop even retry time up to maxAttempts. 

Thanks very much.

JiaLiPassion:
@xinbaihui , thank you for the sample, I have created a PR https://github.com/angular/angular/pull/37091 to fix this one.

xinbaihui:
@JiaLiPassion Thank you for resolving this problem so fast. It helps a lot. Thanks again.

PR #35990, opened 2020-03-10
fix(zone.js): zone.js patches rxjs should check null for unsubscribe

### LINKED ISSUE #31687: Patch-rxjs: _unsubscribe function can be null
here https://github.com/angular/zone.js/blob/9c65d25f583548d8366d94a633a8fceaecc6d1ca/lib/rxjs/rxjs.ts#L120
_unsbscribe is applied before null check
for retryWhen Operator it is set to be null for brief time see here 

https://github.com/ReactiveX/rxjs/blob/d229235f4812dd5140209300e7cddf55553d5f53/src/internal/operators/retryWhen.ts#L108

JiaLiPassion:
Close #31687, #31684

Zone.js patches rxjs `_subscribe` and `_unsubscribe` methods, but zone.js doesn't do null check,
so in some operator such as `retryWhen`, the `_unsubscribe` will be set to null, and will cause
zone patched version throw error.

In this PR, if `_subscribe` and `_unsubscribe` is null, will not do the patch.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

### DIFF
--- a/packages/zone.js/lib/rxjs/rxjs.ts
+++ b/packages/zone.js/lib/rxjs/rxjs.ts
@@ -50,22 +50,28 @@ type ZoneSubscriberContext = {
         },
         set: function(this: Observable<any>, subscribe: any) {
           (this as any)._zone = Zone.current;
-          (this as any)._zoneSubscribe = function(this: ZoneSubscriberContext) {
-            if (this._zone && this._zone !== Zone.current) {
-              const tearDown = this._zone.run(subscribe, this, arguments as any);
-              if (tearDown && typeof tearDown === 'function') {
-                const zone = this._zone;
-                return function(this: ZoneSubscriberContext) {
-                  if (zone !== Zone.current) {
-                    return zone.run(tearDown, this, arguments as any);
-                  }
-                  return tearDown.apply(this, arguments);
-                };
+          if (!subscribe) {
+            (this as any)._zoneSubscribe = subscribe;
+          } else {
+            (this as any)._zoneSubscribe = function(this: ZoneSubscriberContext) {
+              if (this._zone && this._zone !== Zone.current) {
+                const tearDown = this._zone.run(subscribe, this, arguments as any);
+                if (typeof tearDown === 'function') {
+                  const zone = this._zone;
+                  return function(this: ZoneSubscriberContext) {
+                    if (zone !== Zone.current) {
+                      return zone.run(tearDown, this, arguments as any);
+                    }
+                    return tearDown.apply(this, arguments);
+                  };
+                } else {
+                  return tearDown;
+                }
+              } else {
+                return subscribe.apply(this, arguments);
               }
-              return tearDown;
-            }
-            return subscribe.apply(this, arguments);
-          };
+            };
+          }
         }
       },
       subjectFactory: {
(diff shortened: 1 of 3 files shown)


packages/zone.js/test/rxjs/rxjs.Observable.retry.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+      observable1.subscribe(
+          (result: any) => {
+            log.push(result);
+            expect(Zone.current.name).toEqual(subscriptionZone.name);
+          },
+          (err: any) => { fail('should not call error'); },
+          () => {
+            log.push('completed');
+            expect(Zone.current.name).toEqual(subscriptionZone.name);
+            expect(log).toEqual([1, 2, 1, 2, 3, 'completed']);
```
mhevery: Can you move the `expect(log)` to the `it` top level. That way we can also assort that the `subscribe` callbacks actual runs. As it stands right now the test would pass if the subscribe never calls the callbacks.

packages/zone.js/lib/rxjs/rxjs.ts
```diff
@@ -50,22 +50,26 @@ type ZoneSubscriberContext = {
...
+                    return tearDown.apply(this, arguments);
+                  };
+                }
+                return tearDown;
               }
-              return tearDown;
-            }
-            return subscribe.apply(this, arguments);
-          };
+              return subscribe.apply(this, arguments);
```
mhevery: can we add `else` here. The early returns make this code confusing to read. 
```suggestion
             } else { return subscribe.apply(this, arguments); }
```

mhevery (changes requested):

packages/zone.js/test/rxjs/rxjs.Observable.retry.spec.ts
JiaLiPassion: Sure, updated, thanks.

packages/zone.js/lib/rxjs/rxjs.ts
JiaLiPassion: yes, you are right, updated, thanks.

packages/zone.js/test/rxjs/rxjs.Observable.retry.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+    observable1 = constructorZone1.run(() => {
+      return of (1, 2, 3).pipe(
+          map(v => {
+            if (v > 2 && !isErrorHandled) {
+              isErrorHandled = true;
+              throw v;
+            }
+            return v;
+          }),
+          retryWhen(err => err.pipe(delayWhen(v => timer(v * 100)))));
```
mhevery:
```suggestion
          retryWhen(err => err.pipe(delayWhen(v => timer(v)))));
```
Could we change this to be faster?

packages/zone.js/test/rxjs/rxjs.Observable.retry.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+          (err: any) => { fail('should not call error'); },
+          () => {
+            log.push('completed');
+            expect(Zone.current.name).toEqual(subscriptionZone.name);
+          });
+    });
+    setTimeout(() => {
+      expect(log).toEqual([1, 2, 1, 2, 3, 'completed']);
+      done();
+    }, 1000);
```
mhevery: This test will take 1 second to run. That seems a lot.

I am sorry I did not understand that this test was async. In that case, really sorry but can I ask you to put the `expect(log)` back but also include the `done()`. This way we have both, a fast test and `done()` will ensure that it actually runs.

mhevery (approved):
I am really sorry but I am going to ask you to go back to previous code, as I miss-understood the test. Really sorry about that.

LGTM and leaving `cleanup`.
Remove `cleanup` after fixed.

packages/zone.js/test/rxjs/rxjs.Observable.retry.spec.ts
JiaLiPassion: Yeah, you are right, we don't need to `*100` here, updated.

packages/zone.js/test/rxjs/rxjs.Observable.retry.spec.ts
JiaLiPassion: Yes, sorry, it was my fault,  in the previous code, the `done()` is not there, sorry for the confusion. Updated to move `done()` and `expect()` back to the `subscribe`

AndrewKushnir:
FYI, changes in this PR were tested against g3 as a part of https://github.com/angular/angular/pull/36043, using Global TAP run, [which was successful](https://test.corp.google.com/ui#id=OCL:300603599:BASE:301077666:1584325797358:d8800cfc&include_skipped=true). Thank you.

PR #35991, opened 2020-03-10
fix(bazel): prod server doesn't serve files in windows 

alan-agius4:
In Windows runfiles are not always available thus we need to use the real location using rlocation.

Related to bazelbuild/rules_nodejs#1699


alan-agius4:
Targeting master only because the fix is also dependent on https://github.com/bazelbuild/rules_nodejs/commit/d635dca which is available only in bazel nodejs rules 1.4.0.

gregmagolan (approved):
LGTM

packages/bazel/src/schematics/ng-add/index.ts
```diff
@@ -39,14 +39,14 @@ function addDevDependenciesToPackageJson(options: Schema) {
     const devDependencies: [string, string][] = [
       ['@angular/bazel', angularCore.version],
       ['@bazel/bazel', '2.1.0'],
-      ['@bazel/ibazel', '0.11.1'],
+      ['@bazel/ibazel', '0.12.2'],
```
gregmagolan: nit: also update in root package.json?

kyliau (approved):

packages/bazel/src/schematics/ng-add/index.ts
alan-agius4: I’ll do that tmr cause I am current AFK

PR #35992, opened 2020-03-10
feat(dev-infra): add dev-infra to the commit message scopes

josephperrott:
Adds dev-infra to the commit message scopes.  Also, sets the scope to be ignored
in changelogs.

### DIFF
--- a/tools/gulp-tasks/changelog.js
+++ b/tools/gulp-tasks/changelog.js
@@ -10,6 +10,7 @@ module.exports = (gulp) => () => {
   const conventionalChangelog = require('gulp-conventional-changelog');
   const ignoredScopes = [
     'aio',
+    'dev-infra',
     'docs-infra',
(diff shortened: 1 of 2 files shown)


filipesilva (approved):

gkalpak:
When updating commit message scopes, we also need to update [CONTRIBUTING.md#scope][1].
When additionally we add a new GitHub label for the scope (`comp: <new-scope>`), we need to update [docs/TRIAGE_AND_LABELS.md#components][2].

I would updae these myself for `dev-infra`, but I am not sure what exactly it will be used for 😅 

[1]: https://github.com/angular/angular/blob/e4b1e6c62/CONTRIBUTING.md#scope
[2]: https://github.com/angular/angular/blob/e4b1e6c62/docs/TRIAGE_AND_LABELS.md#components

PR #35996, opened 2020-03-10
fix(dev-infra): use @angular/dev-infra-private package for pullapprove verification

josephperrott:
Adds devDependency on @angular/dev-infra-private and removes the verify script
from tools, relying instead on the script from ng-dev.


package.json
```diff
@@ -144,6 +144,7 @@
   "// 3": "when updating @bazel/bazel version you also need to update the RBE settings in .bazelrc (see https://github.com/angular/angular/pull/27935)",
   "devDependencies": {
     "@angular/cli": "9.0.3",
+    "@angular/dev-infra-private": "angular/dev-infra-private-builds#3485f9b",
```
devversion: Oh. That's somewhat inconvenient. Does that mean whenever we make changes to `dev-infra`, we need a first PR to make the change, and then a second (after builds have been pushed), to use the new builds output?

I think it's fine for now in the interim, but long-term it feels like we need alternatives?

IgorMinar: I had the same concern. Couldn't we consume dev-infra from dist/ instead?

IgorMinar (changes requested):

package.json
josephperrott: We can't consume it from dist/ because it will fail on first `yarn` install since `dist` doesn't exist.

devversion: Couldn't we have a `nodejs_binary` and run the CLI that way through Bazel? e.g.

```yaml
- run: yarn bazel run dev-infra:cli -- pullapprove:verify

```

josephperrott: Thats what I have been doing for local development.

This is part of the decision/consideration we have to make with the idea of having a single common resource that is colocated with the angular/angular code.

From the angular/angular perspective it seems odd that we have to make the change and then consume it in another PR, but that is the same experience that all other consumers have.

I personally have mixed feelings about how we should handle this.

IgorMinar: I think that the right trade off here is to consume all the code from the npm package built by bazel, but do not install the package code npm/yarn

devversion: For other repositories like `angular/components`, the minimal steps will always be that changes need to be made in a separate repository, and then the `builds` version needs to be updated. (unless `angular/components` becomes part of the `angular/angular` monorepo).

If we acknowledge that limitation, I still think that it would be beneficial to make sure that dev-infra changes become effective as soon as changes are merged into the `angular/angular` repo.

AndrewKushnir:
FYI, I'm removing the "merge" label for now, since there is a conflict with master branch (the `.pullapprove.yml` file) and it looks like some changes are requested from reviewers. @josephperrott feel free to add the "merge" label back once everything is resolved. Thank you.

package.json
josephperrott: Updated to run it locally from the build artifacts, will talk with bazel web team about if this is something we should work to build into the `pkg_npm` bazel rule directly.

josephperrott:
@IgorMinar ready for another look.  Per our discussion in the dev-infra meeting, I think it might make sense to use this method for now while we determine our long term plan.

devversion (approved):
I'd agree that this is a good interim solution for now and doesn't require a lot of setup. 

After playing with the TS circular deps, and moving that to `dev-infra/`, I hit the issues I suspected initially. It's difficult to test and debug my changes. I personally feel strongly that we need to go with a different approach long-term. I'm looking forward to our discussions about the long-term plan.

IgorMinar (approved):

PR #36001, opened 2020-03-10
test(core): Add benchpress test for transplanted view change detection

atscott:
Adds a performance benchmark for transplanted views views (those which are declared in one component and inserted into another).

### DIFF
--- a/modules/benchmarks/src/change_detection/transplanted_views/transplanted_views.ts
+++ b/modules/benchmarks/src/change_detection/transplanted_views/transplanted_views.ts
@@ -8,7 +8,7 @@
 
 import {CommonModule} from '@angular/common';
 import {ChangeDetectionStrategy, Component, Input, NgModule, TemplateRef, ɵdetectChanges} from '@angular/core';
-import {numViews} from '../util';
+import {newArray, numViews} from '../util';
 
 @Component({
   selector: 'declaration-component',
@@ -37,10 +37,11 @@ export class DeclarationComponent {
 })
 export class InsertionComponent {
   @Input() template !: TemplateRef<{}>;
-  views: undefined[] = [];
+  views: any[] = [];
   @Input()
-  set viewCount(n: number) { this.views = new Array(n); }
+  set viewCount(n: number) { this.views = n > 0 ? newArray<any>(n) : []; }
 
+  // use trackBy to ensure profile isn't affected by the cost to refresh ngFor.
   trackByIndex(index: number, item: any) { return index; }
 }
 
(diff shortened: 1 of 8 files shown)


gkalpak (approved):
Commit message typo: `views views` --> `views` plus a couple of minor questions/suggestions.
Otherwise lgtm ✨

modules/benchmarks/src/change_detection/transplanted_views/transplanted_views.ts
```diff
@@ -0,0 +1,70 @@
...
+import {ChangeDetectionStrategy, Component, Input, NgModule, TemplateRef, ɵdetectChanges} from '@angular/core';
+import {numViews} from '../util';
+
+@Component({
+  selector: 'declaration-component',
+  template: `
+  <ng-template #template>{{trackTemplateRefresh()}}</ng-template>
+  <ng-container *ngFor="let n of views; trackBy: trackByIndex">
+    <insertion-component [template]="template" [viewCount]="viewCount"></insertion-component>
+  </ng-container>
```
mhevery:
```suggestion
    <insertion-component *ngFor="let n of views; trackBy: trackByIndex" [template]="template" [viewCount]="viewCount"></insertion-component>
```

mhevery: perhaps remove it here and just use the `*ngFor` inside the `InsertionComponent`

modules/benchmarks/src/change_detection/transplanted_views/transplanted_views.ts
```diff
@@ -0,0 +1,70 @@
...
+  `,
+})
+export class DeclarationComponent {
+  views: undefined[] = [];
+  @Input()
+  set viewCount(n: number) { this.views = new Array(n); }
+  get viewCount() { return this.views.length; }
```
mhevery: clean up the component as there are no longer many views.

modules/benchmarks/src/change_detection/transplanted_views/transplanted_views.ts
```diff
@@ -0,0 +1,70 @@
...
+  template: `
+    <ng-container *ngFor="let n of views; template: template; trackBy: trackByIndex"></ng-container>
+  `,
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class InsertionComponent {
+  @Input() template !: TemplateRef<{}>;
+  views: undefined[] = [];
+  @Input()
+  set viewCount(n: number) { this.views = new Array(n); }
```
mhevery: `new Array(x)` creates a holey array which has slower access than proper array. In this case it probably makes no difference however, I am worried about showing the right example. Replace with `newArray` helper.

modules/benchmarks/src/change_detection/transplanted_views/transplanted_views.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  trackByIndex(index: number, item: any) { return index; }
```
mhevery: leave a comment as to why this is here.

mhevery (approved):
Clean up

PR #36002, opened 2020-03-11
fix(compiler): set structural directive input as `null` when binding is empty

### LINKED ISSUE #35726: For structural directives with empty attribute the corresponding @Input-method is not called
# 🐞 bug report

### Is this a regression?

Yes, this works at least in angular 7.1, 8.2.9

### Description

When using a structural directive, but with an empty attribute, the corresponding setter for that property is not called. This might be intentional and not a bug, however it would then be an API-change which I did not find anything in the migration guide about and it is also inconsistent with the behaviour of non-structural directives.

## 🔬 Minimal Reproduction

The following example is a custom directive which uses `*appShow` to show some content and `*appShow="false"` to hide the content. The directive itself is of course rather useless, however it is the simplest showcase I came up with. The test fails on angular 9.0.0 and 9.0.2, but passes at least on 8.2.9 and on 7.1.0:

```typescript
import { Component, DebugElement, Directive, Input, Predicate, TemplateRef, ViewContainerRef } from '@angular/core';
import { ComponentFixture, TestBed } from '@angular/core/testing';
import { By } from '@angular/platform-browser';

@Component({
    template: `
        <div *appShow>
            <p id="show">This should show</p>
        </div>
        <div *appShow="false">
            <p id="noShow">This should hide</p>
        </div>`
})
class TestShowDirectiveComponent {
}

@Directive({
    selector: '[appShow]'
})
class ShowDirective {

    constructor(private templateRef: TemplateRef<any>,
                private viewContainer: ViewContainerRef) {}

    @Input()
    set appShow(show: boolean | '') {
        if (show !== false) {
            this.viewContainer.createEmbeddedView(this.templateRef);
        } else {
            this.viewContainer.clear();
        }
    }
}


fdescribe('AppShowDirective', () => {
    let fixture: ComponentFixture<TestShowDirectiveComponent>;

    beforeEach(() => {
        TestBed.configureTestingModule({
            imports: [],
            declarations: [TestShowDirectiveComponent, ShowDirective],
        });

        fixture = TestBed.createComponent(TestShowDirectiveComponent);
        fixture.detectChanges();
    });

    function elementExists(selector: Predicate<DebugElement>): boolean {
        return fixture.debugElement.queryAll(selector).length === 1;
    }

    it('should show only content without param', () => {
        expect(elementExists(By.css('#show'))).toBe(true);
        expect(elementExists(By.css('#noShow'))).toBe(false);
    });
});
```
## 🌍  Your Environment

**Angular Version:**
```
                |___/
    

Angular CLI: 9.0.2
Node: 12.15.0
OS: linux x64

Angular: 9.0.2
Ivy Workspace: Yes

@angular-devkit/core              9.0.3
@angular-devkit/schematics        9.0.2
@schematics/angular               9.0.2
@schematics/update                0.900.2
```

AndrewKushnir:
Prior to this commit, inputs for structural directives that do not have a binding (e.g. `<input *dir>`) were not set, which is different from View Engine behavior that set such inputs (with `null`). This commit updates the logic to set structural directive inputs as `null` when binding is empty. Corresponding tests were also added to verify the behavior in the described case as well as behavior in case of other false-y values.

Fixes #35726.

## PR Type

- [x] Bugfix

- [ ] Yes
- [x] No

### DIFF
--- a/packages/compiler-cli/test/compliance/r3_compiler_compliance_spec.ts
+++ b/packages/compiler-cli/test/compliance/r3_compiler_compliance_spec.ts
@@ -841,7 +841,7 @@ describe('compiler compliance', () => {
           selectors: [["my-component"]],
           decls: 3,
           vars: 0,
-          consts: [["foo", ""], [${AttributeMarker.Template}, "if"]],
+          consts: [["foo", ""], [${AttributeMarker.TemplateUnboundAttrs}, "if"]],
           template:  function MyComponent_Template(rf, ctx) {
             if (rf & 1) {
               $r3$.ɵɵelementStart(0, "ul", null, 0);
@@ -1316,8 +1316,8 @@ describe('compiler compliance', () => {
           }
           const $_c4$ = [[["span", "title", "tofirst"]], "*"];
           …
-          consts: [["id", "second", ${AttributeMarker.Template}, "ngIf"], ["id", "third", ${
-            AttributeMarker.Template}, "ngIf"], ["id", "second"], ["id", "third"]],
+          consts: [["id", "second", ${AttributeMarker.TemplateBindings}, "ngIf"], ["id", "third", ${
+            AttributeMarker.TemplateBindings}, "ngIf"], ["id", "second"], ["id", "third"]],
           template: function Cmp_Template(rf, ctx) {
             if (rf & 1) {
               $r3$.ɵɵprojectionDef($_c4$);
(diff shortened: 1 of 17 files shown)


AndrewKushnir:
[Initial g3 presubmit](http://test/OCL:303833318:BASE:303833357:1585606529879:3d648960) to verify that solution works as expected.

AndrewKushnir:
Closing this PR for now. The fix would introduce extra size to **all** bundles, so we can not land it as is. Extra investigation is needed to figure out if there is a solution that would have no extra size downsides.

PR #36007, opened 2020-03-11
build(dev-infra): handle async failures in `get-data.ts`

alan-agius4:
Wrap the await block in a try/catch so not to have unhandled promise rejections.


josephperrott (approved):
LGTM

PR #36008, opened 2020-03-11
fix(compiler-cli): TypeScript peer dependency range

alan-agius4:
This commit https://github.com/angular/angular/commit/95c729f introduced TypeScript 3.8 support however it is not reflected in the `peerDependencies` section of the compiler-cli package.


josephperrott (approved):
LGTM

PR #36010, opened 2020-03-11
Show helpful error when providing an invalid option 

alan-agius4:
Currently, when running the ngcc binary directly and provide an invalid option ngcc will not error out and the user might have a hard time telling why ngcc is behaving not as expected.

With this change we now output an actionable error:
```
 yarn ngcc --unknown-option
Options:
  --version                          Show version number               [boolean]
  -s, --source                       A path (relative to the working directory)
                                     of the `node_modules` folder to process.
                                                     [default: "./node_modules"]
  -p, --properties                   An array of names of properties in
                                     package.json to compile (e.g. `module` or
                                     `es2015`)
                                     Each of these properties should hold the
                                     path to a bundle-format.
                                     If provided, only the specified properties
                                     are considered for processing.
                                     If not provided, all the supported format
                                     properties (e.g. fesm2015, fesm5, es2015,
                                     esm2015, esm5, main, module) in the
                                     package.json are considered.        [array]
  -t, --target                       A relative path (from the `source` path) to
                                     a single entry-point to process (plus its
                                     dependencies).
  --first-only                       If specified then only the first matching
                                     package.json property will be compiled.
                                                                       [boolean]
  --create-ivy-entry-points          If specified then new `*_ivy_ngcc`
                                     entry-points will be added to package.json
                                     rather than modifying the ones in-place.
                                     For this to work you need to have custom
                                     resolution set up (e.g. in webpack) to look
                                     for these new entry-points.
                                     The Angular CLI does this already, so it is
                                     safe to use this option if the project is
                                     being built via the CLI.          [boolean]
  --legacy-message-ids               Render `$localize` messages with legacy
                                     format ids.
                                     The default value is `true`. Only set this
                                     to `false` if you do not want legacy
                                     message ids to
                                     be rendered. For example, if you are not
                                     using legacy message ids in your
                                     translation files
                                     AND are not doing compile-time inlining of
                                     translations, in which case the extra
                                     message ids
                                     would add unwanted size to the final source
                                     bundle.
                                     It is safe to leave this set to true if you
                                     are doing compile-time inlining because the
                                     extra
                                     legacy message ids will all be stripped
                                     during translation.
                                                       [boolean] [default: true]
  --async                            Whether to compile asynchronously. This is
                                     enabled by default as it allows
                                     compilations to be parallelized.
                                     Disabling asynchronous compilation may be
                                     useful for debugging.
                                                       [boolean] [default: true]
  -l, --loglevel                     The lowest severity logging message that
                                     should be output.
                                     [choices: "debug", "info", "warn", "error"]
  --invalidate-entry-point-manifest  If this is set then ngcc will not read an
                                     entry-point manifest file from disk.
                                     Instead it will walking the directory tree
                                     as normal looking for entry-points, and
                                     then write a new manifest file.
                                                      [boolean] [default: false]
  --help                             Show help                         [boolean]
Unknown arguments: unknown-option, unknownOption
```

### DIFF
--- a/integration/ngcc/test.sh
+++ b/integration/ngcc/test.sh
@@ -42,9 +42,11 @@ function assertNotEquals {
   fi
 }
 
-
 ngcc --help
 assertSucceeded "Expected 'ngcc --help' to succeed."
+
+ngcc --unknown-option 2>&1 | grep 'Unknown arguments: unknown-option, unknownOption'
+assertSucceeded "Expected ngcc to report bad option."
 
 # node --inspect-brk $(npm bin)/ngcc -f esm2015
 # Run ngcc and check it logged compilation output as expected
--- a/package.json
+++ b/package.json
@@ -140,7 +140,7 @@
     "typescript": "~3.8.3",
     "xhr2": "0.1.4",
     "yaml": "^1.7.2",
-    "yargs": "13.1.0"
+    "yargs": "15.3.0"
   },
   "// 2": "devDependencies are not used under Bazel. Many can be removed after test.sh is deleted.",
   "// 3": "when updating @bazel/bazel version you also need to update the RBE settings in .bazelrc (see https://github.com/angular/angular/pull/27935)",
--- a/packages/compiler-cli/ngcc/main-ngcc.ts
+++ b/packages/compiler-cli/ngcc/main-ngcc.ts
@@ -76,6 +76,7 @@ if (require.main === module) {
             describe: 'The lowest severity logging message that should be output.',
             choices: ['debug', 'info', 'warn', 'error'],
           })
+          .strict()
           .help()
           .parse(args);
 
--- a/packages/compiler-cli/package.json
+++ b/packages/compiler-cli/package.json
@@ -22,7 +22,7 @@
     "semver": "^6.3.0",
     "source-map": "^0.6.1",
     "sourcemap-codec": "^1.4.8",
-    "yargs": "13.1.0"
+    "yargs": "15.3.0"
   },
   "peerDependencies": {
     "@angular/compiler": "0.0.0-PLACEHOLDER",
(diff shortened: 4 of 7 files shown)


integration/ngcc/test.sh
```diff
@@ -42,10 +42,12 @@ function assertNotEquals {
   fi
 }
 
-
 ngcc --help
 assertSucceeded "Expected 'ngcc --help' to succeed."
 
+ngcc --unknown-option | grep 'Unknown argument: unknown-option'
```
petebacondarwin: Looks like the error message is actually:

```
Unknown arguments: unknown-option, unknownOption
```

petebacondarwin (changes requested):

integration/ngcc/test.sh
petebacondarwin: See https://app.circleci.com/pipelines/github/angular/angular/11371/workflows/13b540e8-9730-4e5d-9ac8-0d2497949b50/jobs/651144

petebacondarwin (approved):
@IgorMinar - can this land in 9.1.0, since it could cause builds that are using invalid options to fail?

alan-agius4:
@IgorMinar can you please chime in on this? Should we wait until version 10 or we can get in version 9.1?

@petebacondarwin concern is that if users are currently passing invalid options to the `ngcc` binary it works, but with this change it will fail with an error and a non zero exit code. IMO, it's better to fail with a meaningful error message rather than continue running ngcc with unknown options as it might result in unexpected behaviour. This happened to me recently, when integrating with the CLI and spent a good amount of time why ngcc was outputting unexpected files, only to notice afterwards that I was providing an invalid option.

IgorMinar (approved):
I think this one is fine because the likelihood of things working correctly with invalid options is low, so erroring early only makes an existing likely fatal error easier to debug.

IgorMinar:
Plz apply the merge label of this is ready for merge

alan-agius4:
@IgorMinar done 😀

PR #36011, opened 2020-03-11
fix(core): fix memory leak caused by chrome dev tool.

### LINKED ISSUE #35575: Ivy doesn't clear the component after the component is destroyed.
# 🐞 bug report

### Affected Package

### Is this a regression?

### Description

The reproduce repo is here, https://github.com/JiaLiPassion/ivy-memory-leak

Basically we have two routes with two components. `Login` and `Hello` component.
1. The homepage will redirect to `LoginComponent`
Screenshot from 2020-02-20 20-21-58

2. Click a button will navigate to `HelloComponent`.
3. in chrome devtools memory tab, GC->take heap snapshot, but can still find `LoginComponent` in the heap.
Screenshot from 2020-02-20 20-22-44

## 🔬 Minimal Reproduction

 
https://github.com/JiaLiPassion/ivy-memory-leak

## 🔥 Exception or Error
<pre><code>

</code></pre>

## 🌍  Your Environment

**Angular Version:**
<pre><code>

I assume this issue related to this one https://github.com/angular/angular/issues/35148, but I updated to the newest version `9.1.0-next.1`, but the issue still exists.

</code></pre>

**Anything else relevant?**

JiaLiPassion:
Close #35575

In ivy, Angular attaches lView data to the DOM node, and when the DOM node is removed ,
the DOM node and the attached lView data should be garbage collected. But in Chrome,
if the chrome dev tools is opened, the last clicked dom element will be kept in the memory,
and because the DOM node holds the instance of the lView data, and the lView data holds the
component data, so a lot of data is kept in the memory which will cause memory leak and also
make memory analyze with memory snapshot difficult.

So this PR will clear the attached lView data(__ngContext__) from the DOM node when the DOM
node is detached.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

@mhevery , please review, thank you.

### DIFF
--- a/packages/core/src/render3/node_manipulation.ts
+++ b/packages/core/src/render3/node_manipulation.ts
@@ -11,7 +11,7 @@ import {ViewEncapsulation} from '../metadata/view';
 import {addToArray, removeFromArray} from '../util/array_utils';
 import {assertDefined, assertDomNode, assertEqual, assertString} from '../util/assert';
 import {assertLContainer, assertLView, assertTNodeForLView} from './assert';
-import {attachPatchData} from './context_discovery';
+import {attachPatchData, detachPatchData} from './context_discovery';
 import {ACTIVE_INDEX, ActiveIndexFlag, CONTAINER_HEADER_OFFSET, LContainer, MOVED_VIEWS, NATIVE, unusedValueExportToPlacateAjd as unused1} from './interfaces/container';
 import {ComponentDef} from './interfaces/definition';
 import {NodeInjectorFactory} from './interfaces/injector';
@@ -106,6 +106,14 @@ function applyToElementOrContainer(
       nativeInsertBefore(renderer, parent, rNode, beforeNode || null);
     } else if (action === WalkTNodeTreeAction.Detach) {
       nativeRemoveNode(renderer, rNode, isComponent);
+      // Clear __ngContext__ from node to avoid memory leak.
+      // In Chrome, if the chrome dev tool is open, the last
+      // clicked element will be kept in the memory event the element
+      // is removed from it's parentNode, it will make memory
+      // analyze very difficult, so we clear the __ngContext__
+      // here, for details, please check this issue,
+      // https://github.com/angular/angular/issues/35575
+      detachPatchData(rNode);
     } else if (action === WalkTNodeTreeAction.Destroy) {
       ngDevMode && ngDevMode.rendererDestroyNode++;
       (renderer as ProceduralRenderer3).destroyNode !(rNode);
--- a/packages/core/src/render3/context_discovery.ts
+++ b/packages/core/src/render3/context_discovery.ts
@@ -177,6 +177,13 @@ export function attachPatchData(target: any, data: LView | LContext) {
   target[MONKEY_PATCH_KEY_NAME] = data;
 }
 
+/**
+ * Clear the monkey-patched data from the given target
+ */
+export function detachPatchData(target: any) {
+  target[MONKEY_PATCH_KEY_NAME] = null;
+}
+
 export function isComponentInstance(instance: any): boolean {
   return instance && instance.constructor && instance.constructor.ɵcmp;
 }


packages/core/src/render3/node_manipulation.ts
```diff
@@ -106,6 +106,14 @@ function applyToElementOrContainer(
...
     } else if (action === WalkTNodeTreeAction.Detach) {
       nativeRemoveNode(renderer, rNode, isComponent);
+      // Clear __ngContext__ from node to avoid memory leak.
+      // In Chrome, if the chrome dev tool is open, the last
+      // clicked element will be kept in the memory event the element
+      // is removed from it's parentNode, it will make memory
+      // analyze very difficult, so we clear the __ngContext__
+      // here, for details, please check this issue,
+      // https://github.com/angular/angular/issues/35575
+      detachPatchData(rNode);
```
mhevery: This is a megamorphic write which will happen on each destruction slowing it down. Not sure how I feel about it. @kara  and @pkozlowski-opensource what do you think?

mhevery (changes requested):
I am hesitant to work around Chrome DevTools issue like this because it will have a negative perf impact. 
- [ ] At the very least we should have a perf tests which shows the impact.  
- [ ] If we chose to do this we would need tests which asserts that we are removing the path data.
- [ ] Also have concern in that just because a context is disconnected, it may be reconnected later in which case the `LView`s should be retained. So we would need a test to verify that once the removed DOM is re-inserted the `LView`s are restored.

packages/core/src/render3/node_manipulation.ts
pkozlowski-opensource: Tend to agree with @mhevery - this is likely to slow down destroy / detach scenarios. So, as the very minimum we should run benchmarks here (existing largetable / tree would be a good start. BUT, more importantly, if we are _sure_ that this is chrome dev tools issue we should rather follow up with this team and not work-around it in Angular.

JiaLiPassion: @pkozlowski-opensource , thanks and it seems Chrome will not fix this one, 
https://bugs.chromium.org/p/v8/issues/detail?id=10297#c5

mhevery: Given that this happens with Chrome DevTools open and is transient, do we even need to fix it? End users will not notice anything.

alfaproject: It sucks if you are debugging memory leaks. Maybe can work-around under `ngDevMode`?

JohnnyDevNull: This pretty sucks when debugging... it's now the fourth time i've dived into debugging memory leaks because of collages reported to me as main Angular Developere there would be Memory Leaks and i'am not able to argue to my team lead why this happens and the phrase "It's just because it's dev mode" is never a accaptable answer to Senior Lead Devs!

I would also suggest to workaround with ngDevMode and do there some additional cleanup when Chrom won't fix it. This would save a lot of debugging time over and over again running in the same result as described above.

arturovt: @mhevery @JiaLiPassion @pkozlowski-opensource hey guys. It's been 10 months since this pull request was opened. Basically, it's _mostly_ impossible to debug memory leaks in Chrome, other browsers, like Firefox or Safari, don't provide such rich functionality.

Can we give this issue a higher priority? At least as @JohnnyDevNull proposed to invoke this function only in development mode.

mhevery:
I agree that the current situation sucks. The issue is that chrome dev tools retains the last clicked item and with it the whole angular internal `LView`. The issue is that we can't really do much about it even if we would hide it behind `ngDevMode`. Imagine if a click removes a view, in such a situation we could clear the patched data, but 
1. it would be slow, 
2. we could do the slow part behind `ngDevMode` but
3. It would break the application. Imagine if a view is removed (we clear the patch data) and then the view is re-added. We would have to now somehow restore the data wu cleared, but there is no easy way to get that data back.

So this PR is really not a feasible solution. I hear the community that this is problematic, but it is feature of chrome dev tools, and I really can't think of a way to work around this feature.

PR #36013, opened 2020-03-11
fix(bazel): update typescript peer dependency range

devversion:
95c729f5d130a1b903d47304c0aa5b5d0a9b8df2 added support for TypeScript v3.8. Since
these versions are now supported, the `typescript` peer dependency
range in `@angular/bazel` needs to be updated to not report unsatisfied
peer dependencies


alan-agius4:
Also related to https://github.com/angular/angular/pull/36008

alan-agius4 (approved):

josephperrott (approved):
LGTM

Issue #34383, opened 2019-12-12
Update Universal guide code download to v9


vikerman:
Similar to #34382.

The completed code download for Universal guide points to 8.x verison of the project in next.angular.io - https://next.angular.io/generated/zips/universal/universal.zip

gkalpak:
I am working on that as part of [FW-1609][1].
It will be fixed by #34374 (still WIP).

[1]: https://angular-team.atlassian.net/browse/FW-1609

PR #33484, opened 2019-10-30
Add elements project type for boilerplate example projects

yallen011:
Added elements project type to the boilerplate sample projects to fix the issue with the angular elements example project in the Angular docs needing to be updated due to the angular elements package not being present in the `package.json` file when the zip file is downloaded.

## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [x] Documentation content changes
- [ ] angular.io application / infrastructure changes
- [ ] Other... Please describe:

## What is the current behavior?
When downloading the zip from the [https://angular.io/generated/zips/elements/elements.zip](https://angular.io/generated/zips/elements/elements.zip) the `@angular/elements` package is not included in the `package.json`

Issue Number: #31332 

## What is the new behavior? 
the `@angular/elements` package should now be included when downloading the following zip mentioned above.

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

PR #36015, opened 2020-03-11
build(docs-infra): update docs examples to latest v9

### LINKED ISSUE #34382: Update Tour of Heroes completed project ZIP file for v9
The link in completed example to download  - https://next.angular.io/generated/zips/toh-pt6/toh-pt6.zip  in next.angular.io still points to the 8.0 version.

This needs to be updated to the 9.x versions.

gkalpak:
This PR cleans-up the docs examples and updates them to v9 (and also updates the code to match what the latest CLI would generate for a new project).

This was originally part of #34374 (and split off for easier reviewing).
Partly addresses: [FW-1609](https://angular-team.atlassian.net/browse/FW-1609)

##
Fixes #34382 and #34383.


aio/content/examples/template-expression-operators/src/app/app.component.html
```diff
@@ -72,8 +72,8 @@ <h3>Using safe navigation operator (<code>?</code>)</h3>
 <h2>Non-null assertion operator (<code>!</code>)</h2>
 <div>
   <!-- #docregion non-null -->
-  <!--No color, no error -->
-  <p *ngIf="item">The item's color is: {{item!.color}}</p>
+  <!-- Assert color is defined, even if according to the `Item` type it could be undefined. -->
```
petebacondarwin: 👍

aio/tools/examples/shared/package.json
```diff
@@ -18,74 +18,74 @@
...
+    "@nguniversal/common": "~9.0.1",
+    "@nguniversal/express-engine": "~9.0.1",
+    "@nguniversal/module-map-ngfactory-loader": "~9.0.0-next.9",
+    "angular": "1.7.9",
+    "angular-in-memory-web-api": "~0.9.0",
+    "angular-route": "1.7.9",
     "core-js": "^2.5.4",
-    "express": "^4.14.1",
-    "rxjs": "^6.5.1",
+    "express": "^4.17.1",
```
petebacondarwin: Not particularly important for this PR but it is worth noting that `express` should be a `devDependency` right, along with the `@nguniversal/express-engine`...

petebacondarwin (approved):
LGTM:

I assume that the reason the commits are out of order in the GH UI is because of rebases holding on to old dates... This can be remedied with `git rebase master --ignore-date` but not important.

* commit 10/12 does a bit more than update the tsconfig.json - perhaps add some more explanation to the commit message?
* commit 12/12 !fixup seems like it should be its own commit with its own message, not a fixup...

IgorMinar (approved):
lgtm - global approval

but as discussed offline this is blocked until the next zone.js patch release is out.

aio/tools/examples/shared/package.json
gkalpak: No, these are added to `dependencies` by the CLI. I guess it makes sense, because these are prod dependencies for your server side (where they are used to prerender the page before sending it to the client).

gkalpak:
@petebacondarwin 

> This can be remedied with `git rebase master --ignore-date` but not important.

Wow, thx, I didn't know `--ignore-date` works with regular `rebase`. I thought it only works with `--interactive`, which does not work on Windows. But without `--interactive` it works fine :tada:
(Updated my rebase alias :wink:)

> commit 10/12 does a bit more than update the tsconfig.json - perhaps add some more explanation to the commit message?
> commit 12/12 !fixup seems like it should be its own commit with its own message, not a fixup...

In each commit, I make the described change (e.g. update `tsconfig.json`) and all changes required as a result of that to keep the tests passing. For example, updating `tsconfig.json` with some stricter type-checking flags made some template changes necessary. I'll update the commit messages with more info.

[1]: https://github.com/angular/angular/pull/34374/commits/89a37a0202ee3f2131477d697f35c26b0c3dc7a5?file-filters%5B%5D=.ts&hide-deleted-files=true#diff-3e334b5f20890d0e67e19178e1c1cd55R21

gkalpak:
Rebased on master, added more info in commit messages (see previous comment) and added a fixup commit (158929e313264ee131833aabfb1e8fd519763d34) to update `boilerplate/elements/package.json` that was recently merged with #33484.

gkalpak:
Rebased on master and updated `zone.js` to 0.10.3. I also took the opportunity to update `@angular/*` from 9.0.3 to the currently latest 9.0.6.

Marking for merge...

PR #36016, opened 2020-03-11
fix: update exports to result in correct public api

josephperrott:


IgorMinar (commented):
My attempt at setting the expectations: I started reviewing this but realized that now that we expose so many new symbols, we will need to rename some of them for external API naming consistency and that will take more time than just a casual review (and might require consultations with various people on the team).  I will need to review this PR again with a fresh brain, which it is not right now.

petebacondarwin:
Needs a rebase and conflict resolution.

atscott (approved):
reviewed-for: fw-router

While we could make the documentation on better `Tree` and `TreeNode`, the links to them at the moment are completely broken so this is still an improvement.

petebacondarwin:
@josephperrott is this PR still relevant?

josephperrott:
I will go ahead and close this and I can make issues for each of the symbols that are errantly in our goldens.  I think that will end up being the best way to get traction on fixing them individually

PR #36022, opened 2020-03-11
fix(zone.js): zone.js patch jest should handle done correctly

JiaLiPassion:
`zone.js` support jest `test.each()` methods, but it
introduces a bug, which is the `done()` function will not be handled correctly.

```
it('should work with done', done => {
  // done will be undefined.
});
```

The reason is the logic of monkey patching `test` method is different from `jasmine` patch

// jasmine patch
```
return testBody.length === 0
   ? () => testProxyZone.run(testBody, null)
   : done => testProxyZone.run(testBody, null, [done]);
```

// jest patch
```
 return function(...args) {
   return testProxyZone.run(testBody, null, args);
 };
```

the purpose of this change is to handle the following cases.

```
test.each([1, 2])('test.each', (arg1, arg2) => {
  expect(arg1).toBe(1);
  expect(arg2).toBe(2);
});
```

so in jest, it is a little complex, because the `testBody`'s parameter may be bigger than 1, so the
logic in `jasmine`

```
return testBody.length === 0
   ? () => testProxyZone.run(testBody, null)
   : done => testProxyZone.run(testBody, null, [done]);
```
will not work for `test.each` in jest.

So in this PR, I created a dynamic `Function` to return the correct length of paramters (which is required by jest core), to handle
1. normal `test` with or without `done`.
2. each with parameters with or without done.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: The issue is found at `jest-preset-angular` repo here, https://github.com/thymikee/jest-preset-angular/issues/356

### DIFF
--- a/packages/zone.js/lib/jest/jest.ts
+++ b/packages/zone.js/lib/jest/jest.ts
@@ -43,10 +43,9 @@ Zone.__load_patch('jest', (context: any, Zone: ZoneType) => {
 
   function wrapTestFactoryInZone(originalJestFn: Function) {
     return function(this: unknown, ...tableArgs: any[]) {
-      const testFn = originalJestFn.apply(this, tableArgs);
       return function(this: unknown, ...args: any[]) {
         args[1] = wrapTestInZone(args[1]);
-        return testFn.apply(this, args);
+        return originalJestFn.apply(this, tableArgs).apply(this, args);
       };
     };
   }
@@ -64,16 +63,21 @@ Zone.__load_patch('jest', (context: any, Zone: ZoneType) => {
   /**
    * Gets a function wrapping the body of a jest `it/beforeEach/afterEach` block to
    * execute in a ProxyZone zone.
-   * This will run in the `testProxyZone`.
+   * This will run in the `proxyZone`.
    */
   function wrapTestInZone(testBody: Function): Function {
     if (typeof testBody !== 'function') {
       return testBody;
     }
-    // The `done` callback is only passed through if the function expects at least one argument.
-    // Note we have to make a function with correct number of arguments, otherwise jest will
-    // think that all functions are sync or async.
-    return function(this: unknown, ...args: any[]) { return proxyZone.run(testBody, this, args); };
+    const wrappedFunc = function() {
+      return proxyZone.run(testBody, null, arguments as any);
+    };
+    // Update the length of wrappedFunc to be the same as the length of the testBody
+    // So jest core can handle whether the test function has `done()` or not correctly
+    Object.defineProperty(
+        wrappedFunc, 'length', {configurable: true, writable: true, enumerable: false});
+    wrappedFunc.length = testBody.length;
+    return wrappedFunc;
   }
 
   ['describe', 'xdescribe', 'fdescribe'].forEach(methodName => {
--- a/packages/zone.js/test/jest/jest.spec.js
+++ b/packages/zone.js/test/jest/jest.spec.js
@@ -6,10 +6,18 @@ function assertInsideSyncDescribeZone() {
 }
 describe('describe', () => {
   assertInsideSyncDescribeZone();
-  beforeEach(() => { assertInsideProxyZone(); });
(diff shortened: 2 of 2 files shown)


packages/zone.js/lib/jest/jest.ts
```diff
@@ -64,16 +63,24 @@ Zone.__load_patch('jest', (context: any, Zone: ZoneType) => {
...
     }
-    // The `done` callback is only passed through if the function expects at least one argument.
-    // Note we have to make a function with correct number of arguments, otherwise jest will
-    // think that all functions are sync or async.
-    return function(this: unknown, ...args: any[]) { return proxyZone.run(testBody, this, args); };
+    const wrappedFunc = function() { return proxyZone.run(testBody, null, arguments as any); };
+    try {
+      // Update the length of wrappedFunc to be the same with the length of the testBody
+      // So jest core can handle whether the test function have `done()` or not correctly
+      Object.defineProperty(
```
mhevery: That seems to me like a very complicated way to make sure that he function length property matches. 

Could we just always do the catch block. That should be much simpler.

JiaLiPassion: I should delete the `catch` block, because it is some kind of fallback, the `catch` block only handle some cases, not all cases, because `jest` have such cases, 
```
test.each([[1, 2], [2, 3]])((arg1, arg2)=> {
})
```

So the `testFunc.length` will > 1, but still not a `done` case.

mhevery (approved):

mhevery:
[presubmit](https://test.corp.google.com/ui#cl=303990265)

packages/zone.js/lib/jest/jest.ts
```diff
@@ -64,16 +63,19 @@ Zone.__load_patch('jest', (context: any, Zone: ZoneType) => {
...
   function wrapTestInZone(testBody: Function): Function {
     if (typeof testBody !== 'function') {
       return testBody;
     }
-    // The `done` callback is only passed through if the function expects at least one argument.
-    // Note we have to make a function with correct number of arguments, otherwise jest will
-    // think that all functions are sync or async.
-    return function(this: unknown, ...args: any[]) { return proxyZone.run(testBody, this, args); };
+    const wrappedFunc = function() { return proxyZone.run(testBody, null, arguments as any); };
+    // Update the length of wrappedFunc to be the same with the length of the testBody
```
kara, line 73:
```suggestion
    // Update the length of wrappedFunc to be the same as the length of the testBody
```

packages/zone.js/lib/jest/jest.ts
```diff
@@ -64,16 +63,19 @@ Zone.__load_patch('jest', (context: any, Zone: ZoneType) => {
...
+    // So jest core can handle whether the test function have `done()` or not correctly
```
kara, line 74:
```suggestion
    // So jest core can handle whether the test function has `done()` or not correctly
```

kara (changes requested):
Small grammatical changes

Can we also update the commit message to change 'support' to 'supports':

`zone.js support jest `test.each()` methods` => `zone.js supports jest `test.each()` methods`

packages/zone.js/lib/jest/jest.ts
JiaLiPassion: Got it, updated, thanks, I also update the commit message.

packages/zone.js/lib/jest/jest.ts
JiaLiPassion: Got it, updated, thanks.

JiaLiPassion:
@kara , thanks for the review, I just updated the comment and the commit message.

kara (approved):
LGTM, thanks!

kara:
[new presubmit](http://test/OCL:305350416:BASE:305350443:1586297767394:fccb8f7d)

PR #36025, opened 2020-03-11
ci: prevent api golden tests on windows CI

josephperrott:
Updates the package.json scripts to not run the golden api checks on windows CI.  We historically have not run them, and they are not expected to pass.


matsko (approved):

package.json
```diff
@@ -29,8 +29,8 @@
...
     "bazel": "bazel",
     "//circleci-win-comment": "See the test-win circleci job for why these are needed. If they are not needed anymore, remove them.",
-    "circleci-win-ve": "bazel test --build_tag_filters=-ivy-only --test_tag_filters=-ivy-only,-browser:chromium-local //packages/compiler-cli/... //tools/ts-api-guardian/...",
-    "circleci-win-ivy": "bazel test --config=ivy --build_tag_filters=-no-ivy-aot,-fixme-ivy-aot --test_tag_filters=-no-ivy-aot,-fixme-ivy-aot,-browser:chromium-local //packages/compiler-cli/... //tools/ts-api-guardian/...",
+    "circleci-win-ve": "bazel test --build_tag_filters=-ivy-only --test_tag_filters=-ivy-only,-browser:chromium-local -- //packages/compiler-cli/... //tools/ts-api-guardian/... -//packages/compiler-cli:error_code_api -//packages/compiler-cli:compiler_options_api",
+    "circleci-win-ivy": "bazel test --config=ivy --build_tag_filters=-no-ivy-aot,-fixme-ivy-aot --test_tag_filters=-no-ivy-aot,-fixme-ivy-aot,-browser:chromium-local -- //packages/compiler-cli/... //tools/ts-api-guardian/... -//packages/compiler-cli:error_code_api -//packages/compiler-cli:compiler_options_api",
```
gkalpak: Could we use a tag instead, so we don't have to list specific targets in multiple places?

gkalpak (approved):
One minor suggestion. Otherwise lgtm :+1:

package.json
josephperrott: We can do so in a followup, lets get this in so it unblocks master

PR #36027, opened 2020-03-11
perf(ngcc): speed up source map flattening

petebacondarwin:
See individual commits for more details.

### DIFF
--- a/packages/compiler-cli/ngcc/src/sourcemaps/segment_marker.ts
+++ b/packages/compiler-cli/ngcc/src/sourcemaps/segment_marker.ts
@@ -16,6 +16,7 @@
 export interface SegmentMarker {
   readonly line: number;
   readonly column: number;
+  next: SegmentMarker|undefined;
 }
 
 /**
@@ -64,5 +65,5 @@ export function offsetSegment(
     line--;
   }
   const column = newPos - startOfLinePositions[line];
-  return {line, column};
+  return {line, column, next: undefined};
 }
--- a/packages/compiler-cli/ngcc/src/sourcemaps/segment_marker.ts
+++ b/packages/compiler-cli/ngcc/src/sourcemaps/segment_marker.ts
@@ -28,61 +28,41 @@ export function compareSegments(a: SegmentMarker, b: SegmentMarker): number {
   return a.line === b.line ? a.column - b.column : a.line - b.line;
 }
 
-// The `1` is to indicate a newline character between the lines.
-// Note that in the actual contents there could be more than one character that indicates a newline
-// - e.g. \r\n - but that is not important here since segment-markers are in line/column pairs and
-// so differences in length due to extra `\r` characters do not affect the algorithms.
-const NEWLINE_MARKER_OFFSET = 1;
-
 /**
  * Compute the difference between two segment markers in a source file.
  *
- * @param lineLengths the lengths of each line of content of the source file where we are computing
- * the difference
+ * @param startOfLinePositions the position of the start of each line of content of the source file
+ * where we are computing the difference
  * @param a the start marker
  * @param b the end marker
  * @returns the number of characters between the two segments `a` and `b`
  */
-export function segmentDiff(lineLengths: number[], a: SegmentMarker, b: SegmentMarker) {
-  let diff = b.column - a.column;
-
-  // Deal with `a` being before `b`
-  for (let lineIndex = a.line; lineIndex < b.line; lineIndex++) {
-    diff += lineLengths[lineIndex] + NEWLINE_MARKER_OFFSET;
-  }
-
-  // Deal with `a` being after `b`
-  for (let lineIndex = a.line - 1; lineIndex >= b.line; lineIndex--) {
-    // The `+ 1` is the newline character between the lines
-    diff -= lineLengths[lineIndex] + NEWLINE_MARKER_OFFSET;
-  }
-  return diff;
+export function segmentDiff(startOfLinePositions: number[], a: SegmentMarker, b: SegmentMarker) {
+  return startOfLinePositions[b.line] - startOfLinePositions[a.line] + b.column - a.column;
 }
 
 /**
  * Return a new segment-marker that is offset by the given number of characters.
  *
- * @param lineLengths The length of each line in the source file whose segment-marker we are
- * offsetting.
- * @param marker The segment to offset.
- * @param offset The number of character to offset by.
+ * @param startOfLinePositions the position of the start of each line of content of the source file
+ * whose segment-marker we are offsetting.
+ * @param marker the segment to offset.
+ * @param offset the number of character to offset by.
  */
-export function offsetSegment(lineLengths: number[], marker: SegmentMarker, offset: number) {
+export function offsetSegment(
+    startOfLinePositions: number[], marker: SegmentMarker, offset: number) {
   if (offset === 0) {
     return marker;
   }
 
   let line = marker.line;
-  let column = marker.column + offset;
-
-  while (line < lineLengths.length - 1 && column > lineLengths[line]) {
-    column -= lineLengths[line] + NEWLINE_MARKER_OFFSET;
+  const newPos = startOfLinePositions[line] + marker.column + offset;
+  while (line < startOfLinePositions.length - 1 && startOfLinePositions[line + 1] <= newPos) {
     line++;
   }
-  while (line > 0 && column < 0) {
+  while (line > 0 && startOfLinePositions[line] > newPos) {
     line--;
-    column += lineLengths[line] + NEWLINE_MARKER_OFFSET;
   }
-
+  const column = newPos - startOfLinePositions[line];
   return {line, column};
 }
--- a/packages/compiler-cli/ngcc/src/sourcemaps/segment_marker.ts
+++ b/packages/compiler-cli/ngcc/src/sourcemaps/segment_marker.ts
@@ -16,6 +16,7 @@
 export interface SegmentMarker {
   readonly line: number;
   readonly column: number;
+  readonly position: number;
   next: SegmentMarker|undefined;
 }
 
@@ -26,20 +27,7 @@ export interface SegmentMarker {
  * and zero if they are at the same position.
  */
 export function compareSegments(a: SegmentMarker, b: SegmentMarker): number {
-  return a.line === b.line ? a.column - b.column : a.line - b.line;
-}
-
-/**
- * Compute the difference between two segment markers in a source file.
- *
- * @param startOfLinePositions the position of the start of each line of content of the source file
- * where we are computing the difference
- * @param a the start marker
- * @param b the end marker
- * @returns the number of characters between the two segments `a` and `b`
- */
-export function segmentDiff(startOfLinePositions: number[], a: SegmentMarker, b: SegmentMarker) {
-  return startOfLinePositions[b.line] - startOfLinePositions[a.line] + b.column - a.column;
+  return a.position - b.position;
 }
 
 /**
@@ -51,19 +39,19 @@ export function segmentDiff(startOfLinePositions: number[], a: SegmentMarker, b:
  * @param offset the number of character to offset by.
  */
 export function offsetSegment(
-    startOfLinePositions: number[], marker: SegmentMarker, offset: number) {
+    startOfLinePositions: number[], marker: SegmentMarker, offset: number): SegmentMarker {
   if (offset === 0) {
     return marker;
   }
 
   let line = marker.line;
-  const newPos = startOfLinePositions[line] + marker.column + offset;
-  while (line < startOfLinePositions.length - 1 && startOfLinePositions[line + 1] <= newPos) {
+  const position = marker.position + offset;
+  while (line < startOfLinePositions.length - 1 && startOfLinePositions[line + 1] <= position) {
     line++;
   }
-  while (line > 0 && startOfLinePositions[line] > newPos) {
+  while (line > 0 && startOfLinePositions[line] > position) {
     line--;
   }
-  const column = newPos - startOfLinePositions[line];
-  return {line, column, next: undefined};
+  const column = position - startOfLinePositions[line];
+  return {line, column, position, next: undefined};
 }
--- a/packages/compiler-cli/ngcc/src/sourcemaps/source_file.ts
+++ b/packages/compiler-cli/ngcc/src/sourcemaps/source_file.ts
@@ -9,7 +9,7 @@ import {removeComments, removeMapFileComments} from 'convert-source-map';
 import {SourceMapMappings, SourceMapSegment, decode, encode} from 'sourcemap-codec';
 import {AbsoluteFsPath, dirname, relative} from '../../../src/ngtsc/file_system';
 import {RawSourceMap} from './raw_source_map';
-import {SegmentMarker, compareSegments, offsetSegment, segmentDiff} from './segment_marker';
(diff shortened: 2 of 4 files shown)


petebacondarwin:
Screenshot 2020-03-11 at 21 21 32

This is the current state of affairs. The biggest consumer of time is the `flattenMappings()` method. It is not clear whether this is taking a long time because of the calls it makes or whether it is just a method that takes a long time because it has a lot to do.

packages/compiler-cli/ngcc/src/sourcemaps/segment_marker.ts
```diff
@@ -16,6 +16,8 @@
 export interface SegmentMarker {
   readonly line: number;
   readonly column: number;
+  readonly position: number;
+  next?: SegmentMarker;
```
JoostK: Quick observation: consider making this non-optional to achieve a monomorphic types.

packages/compiler-cli/ngcc/src/sourcemaps/source_file.ts
```diff
@@ -302,22 +322,89 @@ export function parseMappings(
...
+}
+
+/**
+ * Clean out the links between original segments of each of the given `mappings`.
+ *
+ * @param mappings the mappings whose segments are to be cleaned.
+ */
+export function removeOriginalSegmentLinks(mappings: Mapping[]) {
+  for (const mapping of mappings) {
+    delete mapping.originalSegment.next;
```
JoostK: This doesn't appear to be used anymore, but using `delete` causes all sorts of deoptimizations (I believe V8 does not track transitions to hidden classes when `delete` has been used, so you'll end up with slow reads as there's no shared hidden class)

JoostK (commented):
A quick glance yields two comments. To get more insight into the flamegraph, you can run node with the `--no-turbo-inlining` flag to disable inlining to see in which functions CPU time is really spent.

packages/compiler-cli/ngcc/src/sourcemaps/source_file.ts
petebacondarwin: This is only used in tests to clean out the mutations for checking the mappings against what was parsed from the source-map file. I put it here in case it was useful elsewhere. So it is not a performance issue.

packages/compiler-cli/ngcc/src/sourcemaps/segment_marker.ts
petebacondarwin: If I made it `SegmentMarker|undefined` would it be monomorphic? The problem we have is that this property is only known after the mappings have all been instantiated and also the last item in the list will be `undefined`.

JoostK: Yes, that would result in only a single hidden class (where all props are present) instead of two hidden classes (one with three props, the other with four)

petebacondarwin: 🎉

gkalpak: Well...how we type this is not really relevant :grin:

There _will_ be two hidden classes, because we initially have `SegmentMarker`s without the `next` prop and then we iterate over them and add the `next` prop (to some value).

The important thing is that each function is called with _one_ of these two "types" (either only called with `SegmentMarker`s without the `next` prop or only called with `SegmentMarker`s with the `next` prop). So, unless I am missing something, we need to at least ensure that, once processed, all `SegmentMarker`s will have a `next` prop (potentially set to `undefined`) - see my comment below.

packages/compiler-cli/ngcc/src/sourcemaps/source_file.ts
```diff
@@ -361,6 +356,33 @@ export function extractOriginalSegments(mappings: Mapping[]): Map<SourceFile, Se
...
+/**
+ * Update the original segments of each of the given `mappings` to include a link to the next
+ * segment in the source file.
+ *
+ * @param mappings the mappings whose segments should be updated
+ */
+export function ensureOriginalSegmentLinks(mappings: Mapping[]): void {
+  const segmentsBySource = extractOriginalSegments(mappings);
+  segmentsBySource.forEach(markers => {
+    for (let i = 0; i < markers.length - 1; i++) {
```
gkalpak: By not iterating over all markers, will end up with a collection of markers of different shapes: Most have a `next` property and the last one does not. This will result to polymorphic calls afaict. (I doubt it will noticeably affect performance, but since it was discussed above...)

So, I suggest ensuring _all_ markers will have a `next` property (even it is is set to `undefined` or `null`).

```suggestion
    for (let i = 0; i < markers.length; i++) {
```

packages/compiler-cli/ngcc/src/sourcemaps/source_file.ts
gkalpak: Except, if it turns out to be useful elsewhere in the future, it may become a performance issue :grin:
Maybe add a comment on that here (or set `next` to `undefined` if that is acceptable).

gkalpak (changes requested):
Nice perf improvements ⚡️ ❤️ 🏇

packages/compiler-cli/ngcc/src/sourcemaps/segment_marker.ts
JoostK: @gkalpak I would expect it to become a requirement to set `next` to `undefined` in an object literal typed as such, so that there's never a hidden class without the `next` prop (well, there is because of hidden class transitions when creating the object, but let's not get into that 😛)

packages/compiler-cli/ngcc/src/sourcemaps/source_file.ts
JoostK: I would prefer to set `next` to `undefined` here, as well.

If this is test only, can it be moved to a test file instead?

packages/compiler-cli/ngcc/src/sourcemaps/source_file.ts
```diff
@@ -169,13 +166,35 @@ export class SourceFile {
...
+  let lowerIndex = -1;
+  let upperIndex = mappings.length - 1;
+  let index = upperIndex;
+  let test = exclusive ? -1 : 0;
+
+  // Binary search the mappings, which is sorted by generatedSegment, for the
+  // highest generated segment that is not higher than the given marker.
+  while (lowerIndex < index) {
+    index = (upperIndex + lowerIndex) >> 1;
+    if (index === -1) {
```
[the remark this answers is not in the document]
petebacondarwin: This whole function is messed up. I am going to rework it. Thanks

packages/compiler-cli/ngcc/src/sourcemaps/source_file.ts
```diff
@@ -169,13 +166,35 @@ export class SourceFile {
...
+      return -1;
+    }
+    if (compareSegments(mappings[index].generatedSegment, marker) <= test) {
+      lowerIndex = index;
```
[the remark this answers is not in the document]
petebacondarwin: Thanks @gkalpak - I have been struggling to get the logic right in this function. I'll give your suggestion a shot. (Plus I'll actually add some unit tests too)

packages/compiler-cli/ngcc/src/sourcemaps/source_file.ts
```diff
@@ -357,6 +360,21 @@ export function extractOriginalSegments(mappings: Mapping[]): Map<SourceFile, Se
...
+  // The `1` is to indicate a newline character between the lines.
+  // Note that in the actual contents there could be more than one character that indicates a
+  // newline
+  // - e.g. \r\n - but that is not important here since segment-markers are in line/column pairs and
+  // so differences in length due to extra `\r` characters do not affect the algorithms.
+  const NEWLINE_MARKER_OFFSET = 1;
+  const lineLengths = computeLineLengths(str);
+  const startPositions = [0];  // First line starts at position 0
+  for (let i = 0; i < lineLengths.length; i++) {
+    startPositions.push(startPositions[i] + lineLengths[i] + NEWLINE_MARKER_OFFSET);
```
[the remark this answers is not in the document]
petebacondarwin: Can you explain what you mean here a bit more?

packages/compiler-cli/ngcc/src/sourcemaps/source_file.ts
petebacondarwin: The downside of going up to `markers.length` is that we end up doing a read on `markers[length]` which converts the array to unpacked.
Instead I will change the `SegmentMarker.next` to be `SegmentMarker|undefined` which means that the `next` property will have to be set (to `undefined`) at creation time so no need to update it here.

packages/compiler-cli/ngcc/src/sourcemaps/source_file.ts
petebacondarwin: I'll move it back to the test file (and change it to set the value to `undefined` rather than deleting the property.

packages/compiler-cli/ngcc/src/sourcemaps/segment_marker.ts
petebacondarwin: 👍

packages/compiler-cli/ngcc/src/sourcemaps/source_file.ts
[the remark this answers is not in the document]
petebacondarwin: Worked a treat. Thanks!

packages/compiler-cli/ngcc/src/sourcemaps/source_file.ts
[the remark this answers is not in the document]
gkalpak: Assuming with have `lineLengths: [1, 2, 3]` (i.e. 3 lines with length 1, 2 and 3 respectively), we will end up with `startPositions: [0, 1, 3, 6]`. The 4th entry (6) does not really correspond to a start position (it is basically the end of the file).

If we do not need that, we should only add to `startPositions` while `i < lineLengths.length - 1` (instead of `i < lineLenghts.length`).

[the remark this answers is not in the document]
petebacondarwin: We do need that as I remember. I am pretty sure tests fail if you remove it.  But I'm on my phone now so can't double check.

gkalpak (approved):
LGTM 👍
Awesome work!

(My only unresolved comment is https://github.com/angular/angular/pull/36027#discussion_r391789217, but that is not blocking by any means.)

packages/compiler-cli/ngcc/src/sourcemaps/source_file.ts
```diff
@@ -169,13 +164,38 @@ export class SourceFile {
...
-function findLastIndex<T>(items: T[], predicate: (item: T) => boolean): number {
-  for (let index = items.length - 1; index >= 0; index--) {
-    if (predicate(items[index])) {
-      return index;
+/**
+ *
+ * @param mappings The collection of mappings whose segment-markers we are searching.
+ * @param marker The segment-marker to match against those of the given `mappings`.
+ * @param exclusive If exclusive then we must find a mapping with a segment-marker that is
+ * exclusively earlier that the given `marker`.
```
JoostK:
```suggestion
 * exclusively earlier than the given `marker`.
```

packages/compiler-cli/ngcc/src/sourcemaps/source_file.ts
```diff
@@ -302,22 +325,79 @@ export function parseMappings(
         }
         const generatedColumn = rawMapping[0];
         const name = rawMapping.length === 5 ? rawMap.names[rawMapping[4]] : undefined;
-        const mapping: Mapping = {
-          generatedSegment: {line: generatedLine, column: generatedColumn},
-          originalSource,
-          originalSegment: {line: rawMapping[2] !, column: rawMapping[3] !}, name
+        const line = rawMapping[2] !;
+        const column = rawMapping[3] !;
+        const generatedSegment = {
```
JoostK:
```suggestion
        const generatedSegment: SegmentMarker = {
```

packages/compiler-cli/ngcc/src/sourcemaps/source_file.ts
```diff
@@ -302,22 +325,79 @@ export function parseMappings(
...
+          line: generatedLine,
+          column: generatedColumn,
+          position: generatedSourceStartOfLinePositions[generatedLine] + generatedColumn,
+          next: undefined,
         };
-        mappings.push(mapping);
+        const originalSegment = {
```
JoostK:
```suggestion
        const originalSegment: SegmentMarker = {
```

packages/compiler-cli/ngcc/src/sourcemaps/segment_marker.ts
```diff
@@ -25,64 +27,31 @@ export interface SegmentMarker {
...
     return marker;
   }
 
   let line = marker.line;
-  let column = marker.column + offset;
-
-  while (line < lineLengths.length - 1 && column > lineLengths[line]) {
-    column -= lineLengths[line] + NEWLINE_MARKER_OFFSET;
+  const position = marker.position + offset;
+  while (line < startOfLinePositions.length - 1 && startOfLinePositions[line + 1] <= position) {
```
JoostK: Related to ([this thread](https://github.com/angular/angular/pull/36027/files#r391797993)), because `startOfLinePositions` appears to contain one additional entry, this would currently allow for the line to be incremented beyond the last line in the source (only if an offset is provided that exceeds the remaining number of characters when on the last line)

JoostK (approved):
Some comments but other than that LGTM.

Would be wise to run this against ngcc-validation prior to landing.

packages/compiler-cli/ngcc/src/sourcemaps/segment_marker.ts
petebacondarwin: I think you are correct but I think this would only happen if the source-map being flattened are invalid. I have added a test for falling off the end of the file and we are without the last `startOfLinePositions` element, so we should be resilient to such invalid source-maps.

packages/compiler-cli/ngcc/src/sourcemaps/source_file.ts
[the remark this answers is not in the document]
petebacondarwin: The tests that fail were only ones checking that this last element actually exists!
So I have removed it and it appears that all is well. Thanks @gkalpak

packages/compiler-cli/ngcc/src/sourcemaps/source_file.ts
```diff
@@ -169,13 +164,38 @@ export class SourceFile {
...
+/**
+ *
+ * @param mappings The collection of mappings whose segment-markers we are searching.
+ * @param marker The segment-marker to match against those of the given `mappings`.
+ * @param exclusive If exclusive then we must find a mapping with a segment-marker that is
+ * exclusively earlier that the given `marker`.
+ * If not exclusive then we can return the highest mappings with an equivalent segment-marker to the
+ * given `marker`.
+ * @param lowerIndex If provided, this is used as a hint that the marker we are searching for has an
+ * index that is no lower than this.
```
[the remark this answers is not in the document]
petebacondarwin: Actually, I did mean "no lower than this" here :-)

petebacondarwin:
Triggered a test on ngcc-validation: https://app.circleci.com/pipelines/github/angular/ngcc-validation/2385/workflows/0cf973bb-8aa7-4909-a39d-605b2860a0f1

PR #36033, opened 2020-03-12
build: update ibazel to `0.12.2`

alan-agius4:
Follow up on https://github.com/angular/angular/pull/35991#discussion_r391023871

//cc @gregmagolan


devversion (approved):
FYI: We initially updated to `ibazel` to `0.12.2` in `angular/components`, but it broke `@bazel/bazelisk` support. It's not yet used in this repo, but I figured I'd mention it.

PR #36034, opened 2020-03-12
Windows TS API Guardian

alan-agius4:
Check individual commits


alan-agius4:
//cc @filipesilva

packages/animations/BUILD.bazel
```diff
@@ -42,10 +42,10 @@ ng_package(
...
 ts_api_guardian_test_npm_package(
     name = "animations_api",
-    actualDir = "packages/animations/npm_package",
+    actualDir = "angular/packages/animations/npm_package",
     data = [
         ":npm_package",
         "//goldens:public-api",
     ],
-    goldenDir = "goldens/public-api/animations",
+    goldenDir = "angular/goldens/public-api/animations",
```
josephperrott: By providing `angular/` at the beginning of the golden file, when bazel runs the `.accept` target, if the file does not already exist in the `goldens/` folder, it creates the `angular/goldens/` directory and places the golden there.

alan-agius4: Umm that's a though one. Because it will not exist in the manifest :(

filipesilva:
Nice work tracking down the path issue!

josephperrott (approved):
LGTM, I think we should get this in so that it works on windows as it currently blocks developers on windows from updating golden apis with the command locally.

As discussed offline, I will in another PR change ts-api-guardian to fail with better messaging if the `.accept` target is run without the golden file(s) already existing.

PR #36044, opened 2020-03-12
build: ng_rollup_bundle internal rule should support ngcc

devversion:
The `ng_rollup_bundle` rule currently is only consumed in the
Angular framework repository. This means that Angular packages
are built from source, and ngcc is never needed to build rollup
bundles using Ivy.

Though, this rule is planned to be shared with other repositories
to support common benchmark code. This means that ngcc needs to be
handled as these other repositories cannot build Angular from source,
but instead consume Angular through NPM (with ngcc enabling Ivy).

The `ng_rollup_bundle` rule needs to dynamically prioritize `ngcc`
generated main resolution fields if `--define=angular_ivy_enabled=True`
is set (or with the alias: `--config=ivy`).


wagnermaciel (approved):

jelbourn (approved):
LGTM

wagnermaciel:
LGTM - I tested these changes in the `angular/components` repo and was able to run an angular application with it.

alan-agius4 (approved):

josephperrott (approved):
LGTM

PR #36047, opened 2020-03-13
feat(compiler): Propagate source span and value span to Variable AST

kyliau:
This commit propagates the `sourceSpan` and `valueSpan` of a `VariableBinding`
in a microsyntax expression to `ParsedVariable`, and subsequently to
View Engine Variable AST and Ivy Variable AST.

Note that this commit does not propagate the `keySpan`, because it involves
significant changes to the template AST.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/compiler-cli/test/ngtsc/template_typecheck_spec.ts
+++ b/packages/compiler-cli/test/ngtsc/template_typecheck_spec.ts
@@ -1331,7 +1331,7 @@ export declare class AnimationEvent {
       const diags = env.driveDiagnostics();
       expect(diags.length).toEqual(1);
       expect(diags[0].code).toEqual(ngErrorCode(ErrorCode.DUPLICATE_VARIABLE_DECLARATION));
-      expect(getSourceCodeForDiagnostic(diags[0])).toContain('let i of items;');
+      expect(getSourceCodeForDiagnostic(diags[0])).toContain('let i = index');
     });
 
     it('should still type-check when fileToModuleName aliasing is enabled, but alias exports are not in the .d.ts file',
--- a/packages/compiler/src/template_parser/binding_parser.ts
+++ b/packages/compiler/src/template_parser/binding_parser.ts
@@ -8,11 +8,11 @@
 
 import {CompileDirectiveSummary, CompilePipeSummary} from '../compile_metadata';
 import {SecurityContext} from '../core';
-import {ASTWithSource, BindingPipe, BindingType, BoundElementProperty, EmptyExpr, ParsedEvent, ParsedEventType, ParsedProperty, ParsedPropertyType, ParsedVariable, ParserError, RecursiveAstVisitor, TemplateBinding, VariableBinding} from '../expression_parser/ast';
+import {ASTWithSource, AbsoluteSourceSpan, BindingPipe, BindingType, BoundElementProperty, EmptyExpr, ParsedEvent, ParsedEventType, ParsedProperty, ParsedPropertyType, ParsedVariable, ParserError, RecursiveAstVisitor, TemplateBinding, VariableBinding} from '../expression_parser/ast';
 import {Parser} from '../expression_parser/parser';
 import {InterpolationConfig} from '../ml_parser/interpolation_config';
 import {mergeNsAndName} from '../ml_parser/tags';
-import {ParseError, ParseErrorLevel, ParseSourceSpan} from '../parse_util';
+import {ParseError, ParseErrorLevel, ParseLocation, ParseSourceSpan} from '../parse_util';
 import {ElementSchemaRegistry} from '../schema/element_schema_registry';
 import {CssSelector} from '../selector';
 import {splitAtColon, splitAtPeriod} from '../util';
@@ -21,7 +21,7 @@ const PROPERTY_PARTS_SEPARATOR = '.';
 const ATTRIBUTE_PREFIX = 'attr';
 const CLASS_PREFIX = 'class';
 const STYLE_PREFIX = 'style';
-
+const TEMPLATE_ATTR_PREFIX = '*';
 const ANIMATE_PROP_PREFIX = 'animate-';
 
 /**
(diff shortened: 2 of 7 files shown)


ayazhafiz:
> Note that this commit does not propagate the `keySpan`, because it involves
> significant changes to the template AST.

Is there a plan to propagate the `keySpan` later on?

packages/compiler/src/template_parser/binding_parser.ts
```diff
@@ -129,16 +129,22 @@ export class BindingParser {
...
         tplKey, tplValue, sourceSpan, absoluteKeyOffset, absoluteValueOffset);
 
-    for (let i = 0; i < bindings.length; i++) {
-      const binding = bindings[i];
+    for (const binding of bindings) {
+      // sourceSpan is for the entire HTML attribute. bindingSpan is for a particular
+      // binding within the microsyntax expression so it's more narrow than sourceSpan.
+      const bindingSpan = offsetParseSourceSpanForMicrosyntax(sourceSpan, binding.sourceSpan);
       const key = binding.key.source;
+      const keySpan = offsetParseSourceSpanForMicrosyntax(sourceSpan, binding.key.span);
```
ayazhafiz: As an aside: when we first added absolute spans, we named them `sourceSpan`s to avoid interference with the existing relative spans named `span`s. Now, as we add new `span`s they are absolute spans, but the naming could be confusing to readers without looking up the type information. We should consider if it would be possible to completely eliminate the relative spans named `span`s. This could be a very long process, but maybe it's the correct one.

packages/compiler/src/template_parser/binding_parser.ts
```diff
@@ -518,3 +524,33 @@ export function calcPossibleSecurityContexts(
...
+ */
+function offsetParseSourceSpanForMicrosyntax(
+    sourceSpan: ParseSourceSpan, absoluteSpan: AbsoluteSourceSpan): ParseSourceSpan {
+  // The difference of two absolute offsets provide the relative offset
+  const startDiff = absoluteSpan.start - sourceSpan.start.offset;
+  const endDiff = absoluteSpan.end - sourceSpan.end.offset;
+  return new ParseSourceSpan(
+      new ParseLocation(
+          sourceSpan.start.file,                // file
+          sourceSpan.start.offset + startDiff,  // absolute offset
```
ayazhafiz: sourceSpan.start.offset + absoluteSpan.start - sourceSpan.start.offset = absoluteSpan.start, no? I think we can simplify all of these.

packages/compiler-cli/test/ngtsc/template_typecheck_spec.ts
```diff
@@ -1331,7 +1331,7 @@ export declare class AnimationEvent {
       const diags = env.driveDiagnostics();
       expect(diags.length).toEqual(1);
       expect(diags[0].code).toEqual(ngErrorCode(ErrorCode.DUPLICATE_VARIABLE_DECLARATION));
-      expect(getSourceCodeForDiagnostic(diags[0])).toContain('let i of items;');
+      expect(getSourceCodeForDiagnostic(diags[0])).toContain('let i = index');
```
ayazhafiz: Interesting. Do you know why it was the other way before?

ayazhafiz (commented):
Looks great! Just one comment regarding simplification.

packages/compiler/src/template_parser/binding_parser.ts
```diff
@@ -518,3 +524,33 @@ export function calcPossibleSecurityContexts(
   });
   return ctxs.length === 0 ? [SecurityContext.NONE] : Array.from(new Set(ctxs)).sort();
 }
+
+/**
+ * Compute a new ParseSourceSpan for a microsyntax inline template binding
+ * based off an original `sourceSpan` by computing a relative offset using the
+ * specified `absoluteSpan`. This should be used only for microsyntax expressions
+ * because the implementation assumes that the offset occurs on the same line.
+ * Microsyntax inline templates are, by definition, on the same line.
```
AndrewKushnir: Can we add a quick test (or find existing one) to make sure we have a good handling of this case in compiler? I.e. having a microsyntax expression that spans multiple lines causes compiler error.

AndrewKushnir (commented):
LGTM, just left one comment (to see if we can add extra test or find existing one). Thank you.

packages/compiler-cli/test/ngtsc/template_typecheck_spec.ts
kyliau: The expectation uses `toContain` to search for the string in the entire RHS of the attribute, so either `let i of items` or `let i = index` would have passed.
Now the span is narrower,  and we know that it is specifically pointing to the second binding that redeclares the variable `i`.

packages/compiler/src/template_parser/binding_parser.ts
kyliau: Agreed. Eventually, there should only be `span`s, and they are absolute. I am also thinking of eventually deprecating `ParseSourceSpan`, because 
1. It contains both line+col and absolute offset, and
2. Every `ParseSourceSpan` internally stores the entire string of the source file.

packages/compiler/src/template_parser/binding_parser.ts
kyliau: It's a sanity check to make sure that `startDiff` is indeed the right calculation for column (otherwise the absolute offset would have been wrong). I'll remove it here and on line 552

kyliau:
> > Note that this commit does not propagate the `keySpan`, because it involves
> > significant changes to the template AST.
> 
> Is there a plan to propagate the `keySpan` later on?

Yes! I'm already working on it. It's a little tricky because I want to do it right and not make `keySpan` optional, which means I need to calculate the keySpan for all variables and not just the ones in microsyntax.

packages/compiler/src/template_parser/binding_parser.ts
kyliau: I've used ParseLocation.move() to compute the ParseSourceSpan instead. This solution is much nicer. Also renamed this function to `moveParseSourceSpan`

AndrewKushnir (approved):
LGTM, thanks @kyliau 👍 

For future updates (while adding `keySpan`), I'm wondering if we can look at grouping spans into new data structure(s) (if it makes sense) to clean up the code more and move the logic of working with spans into separate functions. Something like this:

```
interface ExpressionSpans {
  full: ParseSourceSpan;
  key: ParseSourceSpan;
  value?: ParseSourceSpan;
  ...
}
```
and then in the code we can pass them around together and also perform operations (like "move") on all of them as a group:
```
  const spans = this.shiftExpressionSpans(sourceSpan, binding);
  targetVars.push(new ParsedVariable(key, value, spans));
```

ayazhafiz (approved):

kyliau:
g3 presubmit: https://test.corp.google.com/ui#id=OCL:300816613:BASE:300816643:1584131346752:a4df0fa

PR #36051, opened 2020-03-13
fix(core): properly get root nodes from embedded views with <ng-content>

### LINKED ISSUE #35967: In Ivy, accessing rootNodes results in throwing error when its embedded view has a hole
# 🐞 bug report

Angular v9.0.3

### Is this a regression?

Possibly. This used to work fine with ViewEngine.

### Description

I have this component which I dynamically create and provide its projected nodes. But these projected nodes are placed inside a `ng-template` which I would use to create an embedded view. However, after creating the embedded view, I am no longer could access its `rootNodes` as it's throwing the following error:

```
ERROR Error: ASSERTION ERROR: Should be one of Element, Container, Projection, ElementContainer, IcuContainer but got <unknown> [Expected=> true == false <=Actual]
```

This is the minimal reproduction of the dynamic component: 
```ts
@Component({
  selector: 'dynamic-portal',
  template: `<ng-template #portal><ng-content></ng-content></ng-template>`,
})
export class DynamicPortal implements AfterViewInit {
  @ViewChild('portal') templateRef: TemplateRef<void>;
  view: EmbeddedViewRef<void>;

  ngAfterViewInit() {
    this.view = this.templateRef.createEmbeddedView();
    // Simply trying to access the view's rootNodes would throw an error
    console.log(this.view.rootNodes);
  }
}
```

## 🔬 Minimal Reproduction

 https://stackblitz.com/...

## 🔥 Exception or Error
<pre><code>
ERROR Error: ASSERTION ERROR: Should be one of Element, Container, Projection, ElementContainer, IcuContainer but got <unknown> [Expected=> true == false <=Actual]

</code></pre>

## 🌍  Your Environment

**Angular Version:**
<pre><code>
Angular v9.0.3
Node v12.15.0
</code></pre>

**Anything else relevant?**

pkozlowski-opensource:
This commit fixes 2 separate issues related to root nodes retrieval from
embedded views with `<ng-content>`:

1) we did not account for the case where there were no projectable nodes
for a given `<ng-content>`;

2) we did not account for the case where projectable nodes for a given
`<ng-content>` were represented as an array of native nodes (happens in
the case of dynamically created components with projectable nodes);

Fixes #35967


Shijir (commented):
Thank you, @pkozlowski-opensource! This is awesome. Looking forward to its merge.

packages/core/src/render3/view_ref.ts
```diff
@@ -359,11 +359,19 @@ function collectNativeNodes(
...
       const componentHost = componentView[T_HOST] as TElementNode;
-      const parentView = getLViewParent(componentView);
-      let firstProjectedNode: TNode|null =
-          (componentHost.projection as (TNode | null)[])[tNode.projection as number];
-      if (firstProjectedNode !== null && parentView !== null) {
-        collectNativeNodes(parentView[TVIEW], parentView, firstProjectedNode, result, true);
+      const projectionSlots = componentHost.projection;
+      const slotIdx = tNode.projection as number;
+
+      if (projectionSlots !== null && projectionSlots.length > slotIdx) {
```
mhevery, line 365: Could you add a comment explaining this condition.

packages/core/src/render3/view_ref.ts
```diff
@@ -359,11 +359,19 @@ function collectNativeNodes(
...
+        const nodesInSlot = projectionSlots[slotIdx];
+        if (Array.isArray(nodesInSlot)) {
+          result.push(...nodesInSlot);
+        } else {
+          const parentView = getLViewParent(componentView);
```
mhevery, line 370: comments what is going on here would be helpfull.

mhevery (changes requested):
just comment nits

mhevery (approved):
with nits

packages/core/src/render3/view_ref.ts
pkozlowski-opensource: Simplified the code - turns out that most of checks were not necessary (added an assert). This code is gone so things should be easier to reason about.

packages/core/src/render3/view_ref.ts
pkozlowski-opensource: Simplified the code and added an assert that should serve as documentation.

kara (approved):
LGTM

kara:
[presubmit](http://test/OCL:309837005:BASE:309837034:1588633811262:89e7a659)

kara:
merge-assistance: can't remove the pending review

Issue #31595, opened 2019-07-16
ReferenceError: ngDevMode is not defined


mcalmus:
# 🐞 bug report

### Affected Package

The issue is caused by package @angular/core

### Is this a regression?

 Not sure

### Description

 A clear and concise description of the problem...
If there are compilation errors in a template (while using "ng serve"), I am seeing the error "ReferenceError: ngDevMode is not defined" when running in Firefox (60.6.1esr) or Edge. Chrome shows the actual compilation errors.

## 🔬 Minimal Reproduction

 https://stackblitz.com/...
Create simple app with a component. Introduce compilation errors. Try to run app in browser.

The error (with Angular 8.1.1 is in core.js:13393

## 🔥 Exception or Error
<pre><code>

ReferenceError: ngDevMode is not defined
</code></pre>

## 🌍  Your Environment

Angular Version: 8.1.1 (also 8.1.0)
<pre><code>

</code></pre>

**Anything else relevant?**

miroslavkovac:
I am seeing the same error after upgrading from Angular 6 after a long time. The error is in generated `vendors~main.a7289f63f220e84f1b85.chunk.js` and it looks like couple of `ngDevMode` variables names are wrongly put inside generated JS file:
```
...
ngDevMode, ngDevMode, ngDevMode, ngDevMode, ngDevMode, ngDevMode, ngDevMode, ngDevMode, ngDevMode, ngDevMode;
var qr = function() {
    return Promise.resolve(null)
}();
function Kr(r) {
...
```

mezhik91:
same issue. @angular/core 8.1.2
Is there any way to fix it?

VlasovSergey:
Same here,
I have this issue in only in production build.
Tried to find the reason. Looks like the issue here
https://github.com/angular/angular/blob/cb848b94102855d1e773892b25375304102cab69/packages/core/src/render3/instructions/lview_debug.ts#L60
I think it should be replaced with 
`export const LViewArray = (typeof ngDevMode !== 'undefined') && createNamedArrayType('LView');`
cause ngDevMode does not exist in prodaction build.

mcalmus:
FWIW: I was not using a production build when I encountered the issue.

jfborie:
Same error with --prod, it looks like this class is gone in 8.2.0-rc.0.

CapLonelyFlaw:
Temporary fixed this issue by declaring ngDevMode inside index.html before all scripts only for prod build
[...]

PR #36055, opened 2020-03-13
Do extra check to be sure ngDevMode is defined

mcalmus:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #31595

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/util/ng_dev_mode.ts
+++ b/packages/core/src/util/ng_dev_mode.ts
@@ -115,7 +115,7 @@ export function initNgDevMode(): boolean {
     if (typeof ngDevMode !== 'object') {
       ngDevModeResetPerfCounters();
     }
-    return !!ngDevMode;
+    return typeof ngDevMode !== 'undefined' && !!ngDevMode;
   }
   return false;
 }


packages/core/src/util/ng_dev_mode.ts
```diff
@@ -115,7 +115,7 @@ export function initNgDevMode(): boolean {
     if (typeof ngDevMode !== 'object') {
       ngDevModeResetPerfCounters();
     }
-    return !!ngDevMode;
+    return typeof ngDevMode === 'undefined' ? false : !!ngDevMode;
```
alfaproject:
```suggestion
    return typeof ngDevMode !== 'undefined';
```

mhevery: LGTM, could you add a comment as to why this is needed?

mhevery (approved):

mhevery:
Also please change the git commit message to:
```
fix(core): guard reading of global `ngDevMode` for undefined.

When reading globals such as `ngDevMode` the read should be guarded by `typeof ngDevMode` otherwise it will throw if not defined in `"use strict"` mode.

```

mcalmus:
I know resources are limited, but any chance someone could get the final review done?

AndrewKushnir:
Hi @mcalmus, thanks for creating this PR.

Could you please perform a couple updates:

- squash commits into a single commit (keeping the one that starts with `fix(core): ...`)
- add a comment in the code to describe why there is an additional `typeof ngDevMode` check even though we set it in `ngDevModeResetPerfCounters` as Misko requested in https://github.com/angular/angular/pull/36055#discussion_r392315839 (use the content from the commit body).

Thank you.

mcalmus:
Done

AndrewKushnir:
Thanks for the update @mcalmus. I looked at the comment and I think it'd be great to expand it to provide more context. It's still a bit unclear why this extra check is needed. I agree that this approach is much safer, but it's not clear what the main reason is and having a detailed comment would help a lot. Feel free to share what the problem was in your use-case and I can help put together a comment that we can add to the code. Thank you.

mcalmus:
@AndrewKushnir Here is my comment in the original issue: https://github.com/angular/angular/issues/31595#issuecomment-598698425
Essentially, I was porting legacy code that had an included .js file with a var named 'global'. I had declared this in global.d.ts. When running the app in any browser but Chrome it would blow up with an "undefined" error.

packages/core/src/util/ng_dev_mode.ts
```diff
@@ -115,7 +115,8 @@ export function initNgDevMode(): boolean {
     if (typeof ngDevMode !== 'object') {
       ngDevModeResetPerfCounters();
     }
-    return !!ngDevMode;
+    // This can still be undefined if code has defined global in glboal.d.ts
```
mhevery, line 118: The change seems reasonable, but the comment can't be correct. `.d.ts` can not have any impact on the runtime of code. (It is type checking only information) So regardless of what is declared in `.d.ts` wether or not `ngDevMode` is undefined is orthogonal. 


I would just remove the comment.

mhevery (changes requested):

mcalmus:
Clearly I rushed that comment, even with a typo. I guess I should have said "if defined globally elsewhere" or something like that. In any event it has been removed.

packages/core/src/util/ng_dev_mode.ts
```diff
@@ -115,7 +115,7 @@ export function initNgDevMode(): boolean {
     if (typeof ngDevMode !== 'object') {
       ngDevModeResetPerfCounters();
     }
-    return !!ngDevMode;
+    return typeof ngDevMode !== 'undefined';
```
mhevery, line 118:
```suggestion
    return typeof ngDevMode !== 'undefined' && !ngDevMode;
```

I think this is more correct. Consider if `ngDevMode === 'false'` In that case function should return `false` not `true`

mhevery (changes requested):

mhevery (approved):

packages/core/src/util/ng_dev_mode.ts
```diff
@@ -115,7 +115,7 @@ export function initNgDevMode(): boolean {
     if (typeof ngDevMode !== 'object') {
       ngDevModeResetPerfCounters();
     }
-    return !!ngDevMode;
+    return typeof ngDevMode !== 'undefined' && !ngDevMode;
```
alfaproject, line 118: Shouldn't it be?
```suggestion
    return typeof ngDevMode !== 'undefined' && ngDevMode;
```

(before it was `!!ngDevMode`)

packages/core/src/util/ng_dev_mode.ts
```diff
@@ -115,7 +115,7 @@ export function initNgDevMode(): boolean {
     if (typeof ngDevMode !== 'object') {
       ngDevModeResetPerfCounters();
     }
-    return !!ngDevMode;
+    return typeof ngDevMode !== 'undefined' && ngDevMode;
```
AndrewKushnir, line 118: It looks like the `!!` is needed to make sure the return value type is aligned with function type signature (there are some tests that are failing on CI because of that):
```suggestion
    return typeof ngDevMode !== 'undefined' && !!ngDevMode;
```

AndrewKushnir:
Hi @mcalmus, thanks for applying the changes, the code looks good.

Could you please squash all commits in 1 commit and rebase this PR on top of the current master?

Thank you.

AndrewKushnir:
[Presubmit](http://test/OCL:337384921:BASE:337384943:1602797826566:8fcc7ff8).

PR #36058, opened 2020-03-13
test(core): Update transplanted views benchpress test to support VE

atscott:
Remove uses of render3 internal `detectChanges` calls to allow test to
be run against VE code as well as Ivy.

### DIFF
--- a/modules/benchmarks/src/change_detection/transplanted_views/BUILD.bazel
+++ b/modules/benchmarks/src/change_detection/transplanted_views/BUILD.bazel
@@ -5,14 +5,13 @@ load("//modules/benchmarks:e2e_test.bzl", "e2e_test")
 package(default_visibility = ["//modules/benchmarks:__subpackages__"])
 
 ng_module(
-    name = "ng2",
+    name = "transplanted_views_lib",
     srcs = glob(["*.ts"]),
     generate_ve_shims = True,
     tsconfig = "//modules/benchmarks:tsconfig-build.json",
     deps = [
         "//modules/benchmarks/src:util_lib",
         "//modules/benchmarks/src/change_detection:util_lib",
-        "//packages/common",
         "//packages/core",
         "//packages/platform-browser",
         "//packages/platform-browser-dynamic",
@@ -23,7 +22,7 @@ ng_rollup_bundle(
     name = "bundle",
     entry_point = ":index_aot.ts",
     deps = [
-        ":ng2",
+        ":transplanted_views_lib",
         "@npm//rxjs",
     ],
 )
--- a/modules/benchmarks/src/change_detection/transplanted_views/BUILD.bazel
+++ b/modules/benchmarks/src/change_detection/transplanted_views/BUILD.bazel
@@ -1,55 +1,49 @@
-package(default_visibility = ["//visibility:public"])
-
 load("//tools:defaults.bzl", "ng_module", "ng_rollup_bundle", "ts_devserver")
 load("//modules/benchmarks:benchmark_test.bzl", "benchmark_test")
 load("//modules/benchmarks:e2e_test.bzl", "e2e_test")
 
+package(default_visibility = ["//modules/benchmarks:__subpackages__"])
+
 ng_module(
-    name = "transplanted_views_lib",
-    srcs = [
-        "index_aot.ts",
-        "transplanted_views.ts",
-    ],
-    tags = ["ivy-only"],
+    name = "ng2",
+    srcs = glob(["*.ts"]),
+    generate_ve_shims = True,
+    tsconfig = "//modules/benchmarks:tsconfig-build.json",
     deps = [
         "//modules/benchmarks/src:util_lib",
         "//modules/benchmarks/src/change_detection:util_lib",
-        "//packages:types",
         "//packages/common",
         "//packages/core",
+        "//packages/platform-browser",
+        "//packages/platform-browser-dynamic",
     ],
 )
 
 ng_rollup_bundle(
     name = "bundle",
     entry_point = ":index_aot.ts",
-    tags = ["ivy-only"],
     deps = [
-        ":transplanted_views_lib",
+        ":ng2",
         "@npm//rxjs",
     ],
 )
 
 ts_devserver(
-    name = "devserver",
+    name = "prodserver",
+    bootstrap = ["//packages/zone.js/dist:zone.js"],
     port = 4200,
     static_files = ["index.html"],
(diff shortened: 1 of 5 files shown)


pkozlowski-opensource (approved):
LGTM but needs cleanup (lint job is failing)

modules/benchmarks/src/change_detection/transplanted_views/transplanted_views.ts
```diff
@@ -45,21 +47,10 @@ export class InsertionComponent {
...
-  component.viewCount = numViews;
-  ɵdetectChanges(component);
-}
-
-export function detectChanges(component: DeclarationComponent) {
-  ɵdetectChanges(component);
+@NgModule({
+  declarations: [DeclarationComponent, InsertionComponent],
+  bootstrap: [DeclarationComponent],
+  imports: [CommonModule, BrowserModule]
```
gkalpak: Do we need both? My understanding is that `BrowserModule` includes `CommonModule` (among other things).

modules/benchmarks/src/change_detection/transplanted_views/init.ts
```diff
@@ -0,0 +1,44 @@
...
+  function noop() {}
+
+  const injector = moduleRef.injector;
+  appRef = injector.get(ApplicationRef);
+
+  declaration = appRef.components[0].instance;
+  bindAction('#destroyDom', destroyDom);
+  bindAction('#createDom', createDom);
+  bindAction('#detectChanges', detectChanges);
+  bindAction('#detectChangesProfile', profile(detectChanges, noop, 'detectChanges'));
```
gkalpak: Nit: I find the following more readable:

```ts
  const appRef: ApplicationRef = moduleRef.injector.get(ApplicationRef);
  const declaration: DeclarationComponent = appRef.components[0].instance;

  bindAction('#destroyDom', destroyDom);
  bindAction('#createDom', createDom);
  bindAction('#detectChanges', detectChanges);
  bindAction('#detectChangesProfile', profile(detectChanges, noop, 'detectChanges'));

  // Helpers
  function createDom() {
    declaration.viewCount = numViews;
    appRef.tick();
  }

  function destroyDom() {
    declaration.viewCount = 0;
    appRef.tick();
    declaration.templateRefreshCount = 0;
    appRef.tick();
  }

  function detectChanges() { appRef.tick(); }

  function noop() {}
```

modules/benchmarks/src/change_detection/transplanted_views/init.ts
```diff
@@ -0,0 +1,44 @@
...
+
+import {DeclarationComponent, TransplantedViewsModule} from './transplanted_views';
+
+export function init(moduleRef: NgModuleRef<TransplantedViewsModule>) {
+  let declaration: DeclarationComponent;
+  let appRef: ApplicationRef;
+
+  function destroyDom() {
+    declaration.viewCount = 0;
+    appRef.tick();
```
gkalpak: Is this needed?

modules/benchmarks/src/change_detection/transplanted_views/BUILD.bazel
```diff
@@ -1,55 +1,49 @@
...
+package(default_visibility = ["//modules/benchmarks:__subpackages__"])
+
 ng_module(
-    name = "transplanted_views_lib",
-    srcs = [
-        "index_aot.ts",
-        "transplanted_views.ts",
-    ],
-    tags = ["ivy-only"],
+    name = "ng2",
```
gkalpak: I find this name a little vague/confusing. What was wrong with `transplanted_views_lib`?

gkalpak (approved):
Some minor comments. Otherwise lgtm :+1:

modules/benchmarks/src/change_detection/transplanted_views/init.ts
atscott: Unfortunately, yes. Because Ivy currently refreshes transplanted views at the declaration _and_ at the insertion, when I set the `viewCount` to 0, it still refreshes the transplanted view because the declaration refreshes first (even though the `ngFor` would remove the view). So in order to get my `templateRefreshCount` to actually be 0, I have to first clear the views, then update the counter. This won't be necessary once https://github.com/angular/angular/pull/35968 lands.

modules/benchmarks/src/change_detection/transplanted_views/transplanted_views.ts
atscott: Indeed, you are correct. I should have removed `CommonModule` when I added `BrowserModule`.

modules/benchmarks/src/change_detection/transplanted_views/BUILD.bazel
atscott: Good catch. Copy-paste mistake from another package named `ng2`.

modules/benchmarks/src/change_detection/transplanted_views/init.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  function detectChanges() { appRef.tick(); }
+
+  function noop() {}
+
+  const injector = moduleRef.injector;
+  appRef = injector.get(ApplicationRef);
+
+  declaration = appRef.components[0].instance;
+}
```
gkalpak: Sorry, my previous multi-line comment was a little confusing (probably because I also made a comment inside the range, which threw GitHub off). What I meant is that (in addition to the changes you made already), the `init` function could be cleaner wrt the variable declarations/initialization.

```diff
 export function init(moduleRef: NgModuleRef<TransplantedViewsModule>) {
-  let declaration: DeclarationComponent;
-  let appRef: ApplicationRef;
+  const appRef: ApplicationRef = moduleRef.injector.get(ApplicationRef);
+  const declaration: DeclarationComponent = appRef.components[0].instance;

   bindAction('#destroyDom', destroyDom);
   bindAction('#createDom', createDom);
   bindAction('#detectChanges', detectChanges);
   bindAction('#detectChangesProfile', profile(detectChanges, noop, 'detectChanges'));
 
   // helpers
   function destroyDom() {
     declaration.viewCount = 0;
     appRef.tick();
     declaration.templateRefreshCount = 0;
     appRef.tick();
   }
 
   function createDom() {
     declaration.viewCount = numViews;
     appRef.tick();
   }
 
   function detectChanges() { appRef.tick(); }
 
   function noop() {}

-  const injector = moduleRef.injector;
-  appRef = injector.get(ApplicationRef);
-
-  declaration = appRef.components[0].instance;
}

PR #36071, opened 2020-03-15
fix(docs-infra): include correct dependencies in StackBlitz examples

gkalpak:
Previously, all StackBlitz examples included the default dependencies for `cli`-type projects. However, different example types may have different `package.json` files with different dependencies.
For example, the [boilerplate `package.json`][1] for `elements` examples includes an extra dependency on `@angular/elements`.

This commit changes `StackblitzBuilder` to use the dependencies that correspond to each example type.
(NOTE: Manually verified the changes.)

Jira issue: [FW-2002][2]

[1]: https://github.com/angular/angular/blob/05d058622/aio/tools/examples/shared/boilerplate/elements/package.json
[2]: https://angular-team.atlassian.net/browse/FW-2002

[FW-2002]: https://angular-team.atlassian.net/browse/FW-2002

### DIFF
--- a/aio/tools/stackblitz-builder/builder.js
+++ b/aio/tools/stackblitz-builder/builder.js
@@ -13,16 +13,8 @@ class StackblitzBuilder {
     this.basePath = basePath;
     this.destPath = destPath;
 
-    // Extract npm package dependencies
-    const packageJson = require(path.join(__dirname, '../examples/shared/boilerplate/cli/package.json'));
-    this.examplePackageDependencies = packageJson.dependencies;
-
-    // Add unit test packages from devDependency for unit test examples
-    const devDependencies = packageJson.devDependencies;
-    this.examplePackageDependencies['jasmine-core'] = devDependencies['jasmine-core'];
-    this.examplePackageDependencies['jasmine-marbles'] = devDependencies['jasmine-marbles'];
-
     this.copyrights = this._buildCopyrightStrings();
+    this._boilerplatePackageJsons = {};
   }
 
   build() {
@@ -42,8 +34,34 @@ class StackblitzBuilder {
     });
   }
 
-  _addDependencies(postData) {
-    postData['dependencies'] = JSON.stringify(this.examplePackageDependencies);
+  _addDependencies(config, postData) {
+    // Extract npm package dependencies
+    const exampleType = this._getExampleType(config.basePath);
+    const packageJson = this._getBoilerplatePackageJson(exampleType) || this._getBoilerplatePackageJson('cli');
+    const exampleDependencies = packageJson.dependencies;
+
+    // Add unit test packages from devDependencies for unit test examples
+    const devDependencies = packageJson.devDependencies;
+    ['jasmine-core', 'jasmine-marbles'].forEach(dep => exampleDependencies[dep] = devDependencies[dep]);
+
+    postData.dependencies = JSON.stringify(exampleDependencies);
+  }
+
+  _getExampleType(exampleDir) {
+    const configPath = `${exampleDir}/example-config.json`;
+    const configSrc = fs.existsSync(configPath) && fs.readFileSync(configPath, 'utf-8').trim();
+    const config = configSrc ? JSON.parse(configSrc) : {};
+
+    return config.projectType || 'cli';
+  }
+
+  _getBoilerplatePackageJson(exampleType) {
+    if (!this._boilerplatePackageJsons.hasOwnProperty(exampleType)) {
+      const pkgJsonPath = `${__dirname}/../examples/shared/boilerplate/${exampleType}/package.json`;
+      this._boilerplatePackageJsons[exampleType] = fs.existsSync(pkgJsonPath) ? require(pkgJsonPath) : null;
+    }
+
+    return this._boilerplatePackageJsons[exampleType];
   }
 
   _buildCopyrightStrings() {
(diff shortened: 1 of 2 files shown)


aio/tools/stackblitz-builder/builder.js
```diff
@@ -68,30 +67,29 @@ class StackblitzBuilder {
...
-    var outputFileName = `stackblitz.no-link.html`;
-    outputFileName = configFileName.replace(/stackblitz\.json$/, outputFileName);
-    var altFileName;
+    const outputFileName = configFileName.replace(/stackblitz\.json$/, 'stackblitz.no-link.html');
+    let altFileName;
     if (this.destPath && this.destPath.length > 0) {
-      var partPath = path.dirname(path.relative(this.basePath, outputFileName));
-      var altFileName = path.join(this.destPath, partPath, path.basename(outputFileName)).replace('.no-link.', '.');
+      const partPath = path.dirname(path.relative(this.basePath, outputFileName));
+      altFileName = path.join(this.destPath, partPath, path.basename(outputFileName)).replace('.no-link.', '.');
```
petebacondarwin: 👍

aio/tools/stackblitz-builder/builder.js
```diff
@@ -118,85 +116,87 @@ class StackblitzBuilder {
...
+            var form = document.getElementById('mainForm');
+            var action = form.action;
+            var actionHasParams = action.indexOf('?') > -1;
+            var symbol = actionHasParams ? '&' : '?'
+            form.action = form.action + symbol + embedded;
+          }
+          document.getElementById("mainForm").submit();
+        </script>
+      </body></html>
+    `.trim();
```
petebacondarwin: Is the `trim()` really needed?

aio/tools/stackblitz-builder/builder.js
```diff
@@ -118,85 +116,87 @@ class StackblitzBuilder {
...
     config.fileNames.forEach((fileName) => {
-      var content;
-      var extn = path.extname(fileName);
-      if (extn == '.png') {
+      let content;
+      const extn = path.extname(fileName);
+      if (extn === '.png') {
         content = this._encodeBase64(fileName);
-        fileName = fileName.substr(0, fileName.length - 4) + '.base64.png'
+        fileName = `${fileName.slice(0, -extn.length)}.base64${extn}`;
```
petebacondarwin: 🤓

aio/tools/stackblitz-builder/builder.js
```diff
@@ -208,65 +208,56 @@ class StackblitzBuilder {
       postData[`files[${relativeFileName}]`] = content;
     });
 
-    var tags = ['angular', 'example'].concat(config.tags || []);
-    tags.forEach(function(tag,ix) {
-      postData['tags[' + ix + ']'] = tag;
-    });
+    const tags = ['angular', 'example'].concat(config.tags || []);
```
petebacondarwin: Perhaps:
```suggestion
    const tags = ['angular', 'example', ...config.tags || []];
```

aio/tools/stackblitz-builder/builder.js
gkalpak: Probably not, but added it just in case :grin:

aio/tools/stackblitz-builder/builder.js
```diff
@@ -208,65 +208,56 @@ class StackblitzBuilder {
...
-    var baseHtml = this._createBaseStackblitzHtml(config);
-    var doc = jsdom.jsdom(baseHtml);
-    var form = doc.querySelector('form');
-    _.forEach(postData, (value, key) => {
-      var ele = this._htmlToElement(doc, '<input type="hidden" name="' + key + '">');
+    const baseHtml = this._createBaseStackblitzHtml(config);
+    const doc = jsdom.jsdom(baseHtml);
+    const form = doc.querySelector('form');
+
+    Object.entries(postData).forEach(([key, value]) => {
```
petebacondarwin: We can now also do:

```suggestion
    for(const [key, value] of postData) {
```

aio/tools/stackblitz-builder/builder.js
```diff
@@ -208,65 +208,56 @@ class StackblitzBuilder {
...
   _initConfigAndCollectFileNames(configFileName) {
-    var configDir = path.dirname(configFileName);
-    var configSrc = fs.readFileSync(configFileName, 'utf-8');
+    const configDir = path.dirname(configFileName);
+    const configSrc = fs.readFileSync(configFileName, 'utf-8');
+    let config;
+
     try {
-      var config = (configSrc && configSrc.trim().length) ? JSON.parse(configSrc) : {};
+      config = (configSrc && configSrc.trim().length) ? JSON.parse(configSrc) : {};
```
petebacondarwin: It's a shame that JS doesn't have a good way of initializing a const variable inside a try-catch block.

petebacondarwin: I guess the simplest way would be to put it in a function and call it:

```js
const config = parseConfig(configSrc);
```

petebacondarwin (approved):
LGTM - nothing blocking

aio/tools/stackblitz-builder/builder.js
gkalpak: You mean `... of Object.entries(postData)`, right?
We can indeed. The question is, should we? 😛
<sub>(I did anyway - just because I can 😛 </sub>

gkalpak:
Thx for reviewing, @petebacondarwin ✨
Rebased on master and addressed comments. Marking for merge...

aio/tools/stackblitz-builder/builder.js
petebacondarwin: No! What I wrote works in node 12.14.1 as-is.

gkalpak: On 12.15.0, I get `postData is not iterable` 😕

mhevery:
MERGE_ASSISTANCE: the PR is ready for merge but the bot seems stuck.

PR #36072, opened 2020-03-15
fix(core): prevent unknown property check for AOT-compiled components

### LINKED ISSUE #35945: Ability to bind to any DOM property for storing arbitrary data
Hi,
    I got this error after I upgraded my app to angular 9 when the AOT is enabled.  It is working in JIT.

   Error:  Can't bind to '_data' since it isn't a known property of 'span'.

Code snippet:
`   <span #tmpSpan
                *ngFor="let word of wrods; let i = index; trackBy: trackWord"
                class="__word"
                [attr.key]="i"
                [_data]="word"
                [ngStyle]="getOnWordStyle(word, tmpSpan)"></span>
`
Currently, I can only use a directive with Input to assign the data object to the native element. 
` this.el.nativeElement._data = this._data; `. But I am worry about the performance impact. Is there a way to disable this validation somewhere?

AndrewKushnir:
Prior to this commit, the unknown property check was unnecessarily invoked for AOT-compiled components (for these components, the check happens at compile time). This commit updates the code to avoid unknown property verification for AOT-compiled components by checking whether schemas information is present (as a way to detect whether this is JIT or AOT compiled component).

Note: there is a similar check performed for unknown elements check as well (added in https://github.com/angular/angular/pull/34024).

Resolves #35945.

## PR Type

- [x] Bugfix

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/render3/instructions/shared.ts
+++ b/packages/core/src/render3/instructions/shared.ts
@@ -1038,6 +1038,12 @@ export function setNgReflectProperties(
 function validateProperty(
     tView: TView, lView: LView, element: RElement|RComment, propName: string,
     tNode: TNode): boolean {
+  // If `schemas` is set to `null`, that's an indication that this Component was compiled in AOT
+  // mode where this check happens at compile time. In JIT mode, `schemas` is always present and
+  // defined as an array (as an empty array in case `schemas` field is not defined) and we should
+  // execute the check below.
+  if (tView.schemas === null) return true;
+
   // The property is considered valid if the element matches the schema, it exists on the element
   // or it is synthetic, and we are in a browser context (web worker nodes should be skipped).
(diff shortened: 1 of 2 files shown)


JoostK (commented):
I'd like to explore the idea of refactoring `matchingSchemas` to take just `SchemaMetadata[]` instead of `TView`. It does check for `tView.schemas` being null and returns `false` in that case, but that is not desirable if `null` means an AOT compiled component.

AndrewKushnir:
Hi @JoostK, thanks for the comment.

Great idea, it makes sense. There is also some code around these checks that can be refactored/unified and I can take care of that in a separate (followup) PR.

Thank you.

mhevery (approved):

AndrewKushnir:
[Presubmit](http://test/OCL:306700931:BASE:306700956:1586980156807:e4eb85ce).

Issue #34748, opened 2020-01-12
Wrong error type in httpclient doc


wildsalarian:
# 📚 Docs or angular.io bug report

### Description
A type was changed in a recent commit in angular which is not yet represented in the docs.

### What's the affected URL?**
https://angular.io/guide/http#getting-error-details

### Reproduction Steps**
In the code snippet app/config/config.service.ts there is this line:

`if (error.error instanceof ErrorEvent) `

In case of a network error (or similar like a CORS error) the error property of the HttpErrorResponse  is should be according to the docs an ErrorEvent. But in my tests always seems to be a ProgressEvent.

In this recent commit 
https://github.com/angular/angular/commit/0139b11227fdd93983a8c6bd870d579d798fa244#diff-26683cfb34fa97883213282055aef1e7
in the file
packages/common/http/src/xhr.ts
the type of the error property of the returned HttpErrorResponse was changed from ErrorEvent to ProgressEvent.

Unfortunately unlike ErrorEvent the ProgressEvent has no message property. It is only usable as a indicator for a network error. The message property of the HttpErrorResponse itself has to be queried to retrieve a human readable error message.

The behaviour can be tested in
https://stackblitz.com/edit/error-handling-httpclient?file=src%2Fapp%2Fuser.service.ts

I am using Angular 8.2.14

devversion:
@wildsalarian Thanks for this issue. I think this has always been an issue because the `HttpClient` never emitted an `ErrorEvent`, but instead always used `ProgressEvent`. It was just incorrectly assumed that `ErrorEvent` will be returned (which has been flagged as incorrect by TypeScript in 0139b11)

https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/error_event describes `error` is of type `ProgressEvent`, so I think we should change:

1. Documentation to use `ProgressEvent` in the `instanceof` check (seen in [this guide](https://angular.io/guide/http#getting-error-details))
2. Change `TestRequest` to return `ProgressEvent` instead of `ErrorEvent` (which is incorrect).

This is an interesting question: https://stackoverflow.com/questions/51451508/why-is-a-progressevent-object-passed-to-a-javascript-ajax-error-event-handler.

kapunahelewong:
Status: When https://github.com/angular/angular/pull/36082 merges, this will be resolved.

HarelM:
I just found out the the problem is not only in the docs.
[...]

PR #36082, opened 2020-03-16
fix(common): incorrect error type for XHR errors in `TestRequest`

devversion:
Currently the `HttpClient` always wraps errors from XHR requests, but
the underlying errors are always of type `ProgressEvent`, or don't have
a native error if the status code is just indicating failure (e.g. 404).

This behavior does not match in the `TestRequest` class provided by
`@angular/common/http/testing` where errors are considered being
of type `ErrorEvent`. This is incorrect because `ErrorEvent`s provide
information for errors in scripts or files which are evaluated. Since
the `HttpClient` never evaluates scripts/files, and also since XHR requests
clearly are documented to emit `ProgressEvent`'s, we should change the
`TestSupport` to retrieve such `ProgressEvent`'s instead of incompatible
objects of type `ErrorEvent`.

Resources:
  * https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/error_event
  * https://developer.mozilla.org/en-US/docs/Web/API/ErrorEvent
  * https://xhr.spec.whatwg.org/#event-xhr-errpr

Related to: https://github.com/angular/angular/issues/34748.

### DIFF
--- a/aio/content/examples/http/src/testing/http-client.spec.ts
+++ b/aio/content/examples/http/src/testing/http-client.spec.ts
@@ -137,30 +137,20 @@ describe('HttpClient testing', () => {
   // #enddocregion 404
 
   // #docregion network-error
-  it('can test for network error', () => {
-    const emsg = 'simulated network error';
+  it('can test for network error', done => {
+    // Create mock ProgressEvent with type `error`, raised when something goes wrong
+    // at network level. e.g. Connection timeout, DNS error, offline, etc.
+    const mockError = new ProgressEvent('error');
 
     httpClient.get<Data[]>(testUrl).subscribe(
       data => fail('should have failed with the network error'),
       (error: HttpErrorResponse) => {
-        expect(error.error.message).toEqual(emsg, 'message');
+        expect(error.error).toBe(mockError);
+        done();
       }
     );
 
     const req = httpTestingController.expectOne(testUrl);
-
-    // Create mock ErrorEvent, raised when something goes wrong at the network level.
-    // Connection timeout, DNS error, offline, etc
-    const mockError = new ErrorEvent('Network error', {
-      message: emsg,
-      // #enddocregion network-error
-      // The rest of this is optional and not used.
-      // Just showing that you could provide this too.
-      filename: 'HeroService.ts',
-      lineno: 42,
-      colno: 21
-    // #docregion network-error
-    });
 
     // Respond with mock error
     req.error(mockError);
--- a/aio/content/guide/deprecations.md
+++ b/aio/content/guide/deprecations.md
@@ -38,6 +38,7 @@ v13 -> v16
 | `@angular/common`                   | [`ReflectiveInjector`](#reflectiveinjector)                                                                | <!--v8--> v11         |
 | `@angular/common`                   | [`CurrencyPipe` - `DEFAULT_CURRENCY_CODE`](api/common/CurrencyPipe#currency-code-deprecation)              | <!--v9--> v11         |
 | `@angular/common/http`              | [`XhrFactory`](api/common/http/XhrFactory)                                                                 | <!--v12--> v15        |
+| `@angular/common/http/testing`      | [`TestRequest` accepting `ErrorEvent` for error simulation](#testrequest-errorevent)                       | <!--v13--> v16        |
 | `@angular/core`                     | [`DefaultIterableDiffer`](#core)                                                                           | <!--v7--> v11         |
 | `@angular/core`                     | [`ReflectiveKey`](#core)                                                                                   | <!--v8--> v11         |
 | `@angular/core`                     | [`RenderComponentType`](#core)                                                                             | <!--v7--> v11         |
@@ -469,6 +470,26 @@ In ViewEngine, [JIT compilation](https://angular.io/guide/glossary#jit) required
 
 Important note: this deprecation doesn't affect JIT mode in Ivy (JIT remains available with Ivy, however we are exploring a possibility of deprecating it in the future. See [RFC: Exploration of use-cases for Angular JIT compilation mode](https://github.com/angular/angular/issues/43133)).
 
+{@a testrequest-errorevent}
+
+### `TestRequest` accepting `ErrorEvent`
+
+Angular provides utilities for testing `HttpClient`. The `TestRequest` class from
+`@angular/common/http/testing` mocks HTTP request objects for use with `HttpTestingController`.
+
+`TestRequest` provides an API for simulating an HTTP response with an error. In earlier versions
+of Angular, this API accepted objects of type `ErrorEvent`, which does not match the type of
+error event that browsers return natively. If you use `ErrorEvent` with `TestRequest`,
+you should switch to `ProgressEvent`.
+
+Here is an example using a `ProgressEvent`:
+
+```ts
+const mockError = new ProgressEvent('error');
+const mockRequest = httpTestingController.expectOne(..);
+
+mockRequest.error(mockError);
+```
 
 {@a deprecated-cli-flags}
 
--- a/aio/content/examples/testing/src/app/model/hero.service.spec.ts
+++ b/aio/content/examples/testing/src/app/model/hero.service.spec.ts
@@ -186,27 +186,21 @@ describe('HeroesService (with mocks)', () => {
       req.flush(msg, {status: 404, statusText: 'Not Found'});
     });
 
-    it('should turn network error into user-facing error', () => {
-      const emsg = 'simulated network error';
+    it('should turn network error into user-facing error', done => {
+      // Create mock ProgressEvent with type `error`, raised when something goes wrong at
+      // the network level. Connection timeout, DNS error, offline, etc.
+      const errorEvent = new ProgressEvent('error');
 
       const updateHero: Hero = { id: 1, name: 'A' };
       heroService.updateHero(updateHero).subscribe(
         heroes => fail('expected to fail'),
-        error => expect(error.message).toContain(emsg)
+        error => {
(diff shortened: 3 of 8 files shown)


alxhub (approved):

jelbourn (approved):
LGTM - does anything need to go into the deprecation doc?

Reviewed-for: public-api

devversion:
@jelbourn Yeah, good point. Added an entry to the deprecation guide.

jelbourn:
Adding @aikidave for the deprecation guide

devversion:
FYI: CI is red due to failures in `master`: https://github.com/angular/angular/pull/37679, but PR is still ready for review. I'll rebase when there is more feedback / or it's ready for merge.

aio/content/guide/deprecations.md
```diff
@@ -489,6 +490,30 @@ If you rely on the behavior that the same object instance should cause change de
 - Clone the resulting value so that it has a new identity.
 - Explicitly call [`ChangeDetectorRef.detectChanges()`](api/core/ChangeDetectorRef#detectchanges) to force the update. 
 
+
+{@a testrequest-errorevent}
+### `TestRequest` accepting `ErrorEvent`
+
+The `@angular/common` package provides utilities for testing the `HttpClient`. One part of these
+testing utilities is the `TestRequest` class. It serves as mock for HTTP requests and is primarily
+used in the `HttpTestingController`.
```
jelbourn, line 499:
```suggestion
The `@angular/common` package provides utilities for testing `HttpClient`. The `TestRequest` class from `@angular/common/http/testing` mocks HTTP request objects for use with `HttpTestingController.
```

aio/content/guide/deprecations.md
```diff
@@ -489,6 +490,30 @@ If you rely on the behavior that the same object instance should cause change de
...
+
+`TestRequest` provides an interface for resolving to mocked content, or simulating an errored http
+request. Currently, the method for simulating an error accepts objects of type `ErrorEvent`. This
+does not match with error events that are returned by browsers natively, so we intend to no longer
+accept `ErrorEvent`.
```
jelbourn, line 504:
```suggestion
`TestRequest` provides an API for simulating an HTTP response with an error. In earlier versions of Angular, this API accepted objects of type `ErrorEvent`, which
does not match the type of error event that are returned by browsers natively. If you use `ErrorEvent` with `TestRequest`, you should switch to `ProgressEvent`. A future version of Angular will remove support for using `ErrorEvent` with `TestRequest`. 
```

aio/content/guide/deprecations.md
devversion: Yeah, that's definitely better. thx!

aio/content/guide/deprecations.md
devversion: Thanks! way better 👍

goldens/public-api/common/http/testing/testing.d.ts
```diff
@@ -25,13 +25,8 @@ export declare class TestRequest {
...
     constructor(request: HttpRequest<any>, observer: Observer<HttpEvent<any>>);
-    error(error: ErrorEvent, opts?: {
-        headers?: HttpHeaders | {
-            [name: string]: string | string[];
-        };
-        status?: number;
-        statusText?: string;
-    }): void;
+    /** @deprecated */ error(error: ErrorEvent, opts?: TestRequestErrorOptions): void;
+    error(error: ProgressEvent, opts?: TestRequestErrorOptions): void;
```
alfaproject, line 29: `ProgressEvent` sounds very specific. Is this intended?

aio/content/guide/deprecations.md
```diff
@@ -489,6 +490,29 @@ If you rely on the behavior that the same object instance should cause change de
...
+Angular provides utilities for testing `HttpClient`. The `TestRequest` class from
+`@angular/common/http/testing` mocks HTTP request objects for use with `HttpTestingController`.
+
+`TestRequest` provides an API for simulating an HTTP response with an error. In earlier versions
+of Angular, this API accepted objects of type `ErrorEvent`, which does not match the type of
+error event being returned by browsers natively. If you use `ErrorEvent` with `TestRequest`,
+you should switch to `ProgressEvent`. A future version of Angular will remove support for
+using `ErrorEvent` with `TestRequest`.
+
+Here is an example using a `ProgressEvent`:
```
jelbourn, line 506:
```suggestion
The following example demonstrates use of `ProgressEvent` with `HttpTestingController`:
```

jelbourn (approved):
LGTM

Reviewed-for: public-api

goldens/public-api/common/http/testing/testing.d.ts
devversion: @alfaproject yeah, that is true. What would you expect in those cases? The goal here was to match with native error being constructed when a request fails. https://xhr.spec.whatwg.org/#event-xhr-error

It seems ambiguous that `TestRequest.error` could theoretically also deal with non-network level errors. e.g. just a simple 404 error. In that case though it's unclear what `ErrorEvent` or `ProgressEvent` (with this change) should be though. Ideally only network errors are handled here and other server responses are simulated with `flush`. This is ambiguous though as the method logic supports status responses, status text and headers.

alfaproject: I guess I was expecting to see the `Event` base class in the public API but I suppose that's too broad as well. A `ProgressEvent` is not necessarily an error and that's why I got a bit confused.

`ErrorEvent` also extends from `Event` but as you mentioned that's not really emitted.

¯\_(ツ)_/¯

IgorMinar (approved):
the deprecation lgtm

Reviewed-for: docs-packaging-and-releasing

IgorMinar (approved):
the deprecation lgtm

Reviewed-for: docs-packaging-and-releasing, public-api

pkozlowski-opensource (commented):
LGTM overall but I think that @alxhub has more context so I'm not going to submit this as approving review (so Alex has a chance of rubber-stamping it).

pkozlowski-opensource (approved):
Oh, actually @alxhub reviewed this already. Again, this LGTM.

Reviewed-for: public-api

aio/content/guide/deprecations.md
```diff
@@ -489,6 +490,29 @@ If you rely on the behavior that the same object instance should cause change de
...
+
+{@a testrequest-errorevent}
+### `TestRequest` accepting `ErrorEvent`
+
+Angular provides utilities for testing `HttpClient`. The `TestRequest` class from
+`@angular/common/http/testing` mocks HTTP request objects for use with `HttpTestingController`.
+
+`TestRequest` provides an API for simulating an HTTP response with an error. In earlier versions
+of Angular, this API accepted objects of type `ErrorEvent`, which does not match the type of
+error event being returned by browsers natively. If you use `ErrorEvent` with `TestRequest`,
```
aikidave, line 502:
```suggestion
error event that browsers return natively. If you use `ErrorEvent` with `TestRequest`,
```
Passive voice. Hard to avoid sometimes!

aio/content/guide/deprecations.md
```diff
@@ -489,6 +490,29 @@ If you rely on the behavior that the same object instance should cause change de
...
+you should switch to `ProgressEvent`. A future version of Angular will remove support for
+using `ErrorEvent` with `TestRequest`.
```
aikidave, line 504:
```suggestion
you should switch to `ProgressEvent`.
```

In general, I advise against statements such as "a future version..." or "in the future..." They often don't add as much value as we think--often, readers care about the state of the product as it is right now, not how it will be at some point in the future.

In this case, I think there's a second reason to remove this sentence: We're already in the deprecations topic. The fact that something might be removed is probably well understood. :)

aikithoughts (approved):
Just a few suggestions. Otherwise, LGTM.

jessicajaniuk:
@devversion This PR was approved, but looks like it stalled. Can you take a look at it again?

aio/content/guide/deprecations.md
devversion: thanks. that's much better 👍

aio/content/examples/http/src/testing/http-client.spec.ts
```diff
@@ -138,30 +138,19 @@ describe('HttpClient testing', () => {
...
-    const emsg = 'simulated network error';
+    // Create mock ProgressEvent with type `error`, raised when something goes wrong
+    // at network level. e.g. Connection timeout, DNS error, offline, etc.
+    const mockError = new ProgressEvent('error');
 
     httpClient.get<Data[]>(testUrl).subscribe(
       data => fail('should have failed with the network error'),
       (error: HttpErrorResponse) => {
-        expect(error.error.message).toEqual(emsg, 'message');
+        expect(error.error).toBe(mockError);
```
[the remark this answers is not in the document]
devversion: I think this is not necessarily related to this PR, but I added the `done` completion. seems like a good improvement.

devversion:
@jessicajaniuk I've just resurrected this PR. Addressed feedback etc. I just need one more public API approval.

jessicajaniuk (approved):
LGTM :cookie:

reviewed-for: public-api

jessicajaniuk:
I'll run a TGP to make sure we're safe in google3 with this change.

alxhub (approved):

jessicajaniuk:
@devversion Can you rebase this and repush? It seems to fail to build in google3 now.

jessicajaniuk:
The TGP passed.

jessicajaniuk:
This PR was merged into the repository by commit 489cf42cd07ac2b549ea958127590b30f1b3d28f.

Issue #17606, opened 2017-06-19
MODULE_INITIALIZER like APP_INITIALIZER 


pantonis:
## I'm submitting a ...

<pre><code>
[ ] Regression (behavior that used to work and stopped working in a new release)
[ ] Bug report 
[X] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead see https://github.com/angular/angular/blob/master/CONTRIBUTING.md#question
</code></pre>

I was wondering if like APP_INITIALIZER a MODULE_INITIALIZER can be implemented. 
I have a scenario where multiple lazy load modules exist. Each module has a service that has injected in its constructor a config of type ConfigA. ConfigA is fetched from server. This service is then injected into several module components. With APP_INITIALIZER I cannot do this since the same type ConfigA is used for all modules and a singleton will be created.

vicb:
Could you please describe your use case more precisely ?

How do you lazy load your module ? 
If you use the router you might want to use a resolver to fetch data from the serve ?

pantonis:
Hi,

Please see the sample code I created above that is implemented with resolvers and is not working
````
@NgModule({
    providers: [
        StompServiceProvider,        
        ConfigResolve
    ]

})
export class FeatureModule { }


import { Http } from '@angular/http';
import { OverrideStompService } from "./services/override-stomp.service";
import { ActivatedRoute } from '@angular/router';

export let StompServiceProvider =
    {
        provide: OverrideStompService,
        useClass: OverrideStompService
    };


import { Http } from "@angular/http";
import { Injectable } from '@angular/core';
import { StompService, StompConfig } from "@stomp/ng2-stompjs";


@Injectable()
export class OverrideStompService extends StompService {
    constructor(stompConfig: StompConfig) {
        super(stompConfig);
    }   
}


import { Http, Response } from "@angular/http";
import { Injectable } from '@angular/core';
import { Resolve, ActivatedRouteSnapshot } from '@angular/router';
import { StompConfig } from '@stomp/ng2-stompjs';
import 'rxjs/add/operator/map';

@Injectable()
export class ConfigResolve implements Resolve<StompConfig>{
    constructor(private http: Http) {
    }

    resolve(route: ActivatedRouteSnapshot) {
        return this.http.get('app/config/brokerConfigFile.json')
            .map((response: Response) =>{ 
                debugger;
                return response.json()});        
    }
}
````

pantonis:
[...]

Issue #18093, opened 2017-07-13
Ergonomic component-level code splitting and lazy loading


aalbericio:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[X] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead see https://github.com/angular/angular/blob/master/CONTRIBUTING.md#question
</code></pre>

## Current behavior

Router Module is optional and, depending on the nature of the app, it does not naturally fit.

If we don't use the RouterModule, I can't see an easy way to take advantage of the "lazy loading" feature for modules.

I'd like to, somehow, identify an NgModule to be "lazy loaded" at definition time (just like you define its providers, etc).

Once the module has been marked as "lazy loaded", the build should generate an independent chunk for it and that module becomes eligible for being lazy-loaded at some execution time.

Of course, If I have the choice of marking a module "lazy", I also need a mechanism to manually load those "lazy loaded" modules and start using them.

In summary, the feature request is: make the "lazy loading" feature available to developers outside the RouterModule since this module is optional but the feature should still be available.

Of course this should output AOT compatible chunks, etc, just like the RouterModule internally does.

## Expected behavior

In summary, the feature request is: make the "lazy loading" feature available to developers outside the RouterModule since this module is optional but the feature should still be available.

Of course this should output AOT compatible chunks, etc, just like the RouterModule internally does.

## What is the motivation / use case for changing the behavior?

Our application and functionally can not be Router driven thus we need some way to make use of the "lazy loading" strategy that is internally used by the Router.

## Environment

<pre><code>
Angular version: X.Y.Z

Browser:
- [ ] Chrome (desktop) version XX
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX
 
For Tooling issues:
- Node version: XX  
- Platform:  

Others:

</code></pre>

Toxicable:
You can already do this, here is a small example of how
```
System.import('./lazy.module.ngfactory.ts').then((moduleFactories: any) => {
   const moduleFactory: NgModuleFactory<any> = moduleFactories['LazyModuleNgFactory'];
   const moduleRef = moduleFactory.create(injector);
[...]

PR #36084, opened 2020-03-16
feat(router): route initializer

kristofdegrave:
This is an initial POC to provide an initialization hook for lazy loaded Modules. Implementation is Open for Discussion!

Additional Docs will be added when a consent about the solution is made.

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

A lazy route is now just loaded without being block by initialization logic

Issue Number: #17606

## What is the new behavior?
A lazy route is now loaded with an option being block by initialization logic. If all initialization logic has run, the route gets loaded

- [ ] Yes
- [x] No

## Other information
It introduces a new injection toke ROUTE_INITIALIZER.

### DIFF
--- a/packages/router/src/router_config_loader.ts
+++ b/packages/router/src/router_config_loader.ts
@@ -10,7 +10,7 @@
 import {Compiler, InjectionToken, Injector, NgModuleFactory, NgModuleFactoryLoader} from '@angular/core';
 import {isPromise} from '@angular/core/src/util/lang';
 import {Observable, from, of } from 'rxjs';
-import {mergeMap, switchMap} from 'rxjs/operators';
+import {map, mergeMap} from 'rxjs/operators';
 
 import {LoadChildren, LoadedRouterConfig, Route, standardizeConfig} from './config';
 import {flatten, wrapIntoObservable} from './utils/collection';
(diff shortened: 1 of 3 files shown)


kristofdegrave:
@googlebot I signed it!

steve-todorov:
Hey @atscott, have you had a chance to checkout this implementation yet? :)

packages/router/src/router_config_loader.ts
```diff
@@ -6,19 +6,37 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
+// TODO(i): switch to fromPromise once it's expored in rxjs
 import {Compiler, InjectionToken, Injector, NgModuleFactory, NgModuleFactoryLoader} from '@angular/core';
+import {isPromise} from '@angular/core/src/util/lang';
```
atscott, line 11: This should be imported from `@angular/core` like what's done in `utils/collection.ts`

packages/router/src/router_config_loader.ts
```diff
@@ -6,19 +6,37 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
+// TODO(i): switch to fromPromise once it's expored in rxjs
```
atscott, line 9: let's remove this comment

atscott (changes requested):
Does this provide value beyond what `CanActivate` at the `forChild` root could do?
```
RouterModule.forChild([
  {path: '', canActivate: [MyInitializerGuard], children: [...]}
]);
```
The `canActivate` there would allow for asynchronous initialization before attempting to activate any child routes in the module. If you think of `canActivate` conceptually like "can this route activate yet?" rather than "is it allowed to activate at all?" it makes a bit more sense to do this.

kristofdegrave:
@atscott I think the canActivate hook will be to late. The loading of the module depends on the loading strategy and can be preloaded. The case I need it for is to make sure that I can fetch configuration for example before the the module and it's dependencies are loaded, because they can depend on that configuration. 
I will look into the review you gave asap.

atscott:
> @atscott I think the canActivate hook will be to late. The loading of the module depends on the loading strategy and can be preloaded. The case I need it for is to make sure that I can fetch configuration for example before the the module and it's dependencies are loaded, because they can depend on that configuration.

Outside of preloading, CanLoad could be used in this case. With preloading, it seems like you could actually build this functionality into your preloader. I think you’re going to need to provide more concrete examples before we can consider moving this PR forward. It seems like the existing router guards allow for handling the behavior desired. This also does not resolve the linked issue since modules can be loaded without the router.

kristofdegrave:
@atscott never looked this way at it, you could have a point. The desired behavior is the same as the APP_INITIALIZER I want to have, and ideal it shouldn't matter if the module is loaded immediately or trough a lazy route.
I wanted this behavior to prepare my application to move to a MicroFrontend architecture. So every (routed) module would be a micro frontend in this case, with it's own configuration file. So to get this behavior right I should write some logic in the preloader and the canLoad guard that gets the APP_INITIALIZER tokens from the module that needs to get loaded and initialize them, but is this even possible? 
The module would need to be loaded before I can get them in those cases? Also I would need something to determine if the initialization has run

atscott:
> The module would need to be loaded before I can get them in those cases? Also I would need something to determine if the initialization has run

That's right, I was really just suggesting `canLoad` because you said `canActivate` is too late. I guess what you meant was that if the module is preloaded, `canActivate` still won't run until you try navigating to a route in that module but you would want the initializers to run before that. 

Here's an [over-engineered] example of how this could all work: 
https://stackblitz.com/edit/angular-ivy-72wgbc?file=src%2Fapp%2Finitializing-router.module.ts
Unfortunately, there's still a "small" hack since `_loadedConfig` is internal and the preloader doesn't technically have any way to access the module that was loaded. Maybe there's a way to do this that I'm not thinking of... Other than that, I think the would be what you're looking for.

atscott:
Closing this PR per the comments mentioned above, namely:
* This does not resolve the linked issue since modules can be loaded without the router.
* Some of this behavior can be achieved with the router guards
* I think this should be addressed in a more complete way along with #18093 rather than extending the router API to only partially fix the issues.

Issue #7557, opened 2016-03-11
Backspace and delete keys on input text doesn't update model on IE9.


miguelangelorocha:

MarkPieszak:
If internally it's using a `.change() event then yes, it ignores those. (More of an IE9 bug / bizarre difference) but maybe it should be changed to a keyup/down for this jnstance?

miguelangelorocha:
All keys change the model except delete and backspace. When events keyup/down are debugged, after delete or backspace, the model don't change too.

renehamburger:
We've got a similar issue with input validation not being triggered on IE9 for backspace, delete or cut. The reason is that IE9 only triggers a `keyup` but no `input` event for these operations (see https://developer.mozilla.org/en-US/docs/Web/Events/input#Browser_compatibility), while Angular only binds `input` and `blur` events to the handler.

zoechi:
You should be able to workaround by providing a custom `ControlValueAccessor`

Delagen:
Please fix this. This is a breaking issue for using any form in IE 9.

kara:
Can someone post a repro case for this?  

@Delagen If this issue is important to you, feel free to submit a PR.

Delagen:
@kara Plunker: http://plnkr.co/edit/Q1MNpMgjuYwBcnYx4AA0?p=preview
Open run frame in IE 9.
Try to change input. Changing value using backspace and delete keys does not update model.

pkozlowski-opensource:
The root cause here is that IE9 doesn't fire `input` events on backspace / delete. In Angular 1.x we were working around it by listening to the change event as well.

Delagen:
@pkozlowski-opensource Yes. You are right. I found it when I learn code of NgModel directive.
http://caniuse.com/#feat=input-event

pkozlowski-opensource:
@kara I'm going to remove the "Need repro" label as the issue is known - see https://github.com/angular/angular/issues/7557#issuecomment-253723850

thomasmerieux:
Hi guys,

Do you have any update on this issue (or a workaround) ?

Thanks

ncroma:
Using **_FormControl_** and **_formControlName_** on IE9 doesn't update the model either when using backspace. Is this getting worked on soon? This is making text input for IE9 users really unreliable.

Fabyao:
[...]

PR #36087, opened 2020-03-16
fix(forms): number input fires valueChanges twice.

### LINKED ISSUE #12540: Number input fires valueChanges twice
May be bug.

ReactiveFormsModule. Input with type "number" fires valueChanges event twice when changed from browser.

I expect it to fire valueChanges once.

Here is a plunkr example: https://plnkr.co/edit/HYCtlP72W3edsaZe9HtR?p=preview

I use ngrx/store to manage state and subscribe my store to valueCahnges observable. When I change value of input, reducers fires twice. It is not optimal, I suppose.

Angular version: 2.1.0

Browser: Chrome

TidianeRolland:
Prior to this commit, input field of type number used to fire valueChanges twice. First time after typing in the input field and second time when the input field loses the focus.

PR Close #12540

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Input field of type number emits valueChanges twice. First time after typing in the input field and second time when the input field loses the focus.

Issue Number: #12540

## What is the new behavior?
Input field of type number emits valueChanges once.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/forms/src/directives/number_value_accessor.ts
+++ b/packages/forms/src/directives/number_value_accessor.ts
@@ -43,11 +43,7 @@ export const NUMBER_VALUE_ACCESSOR: any = {
 @Directive({
   selector:
       'input[type=number][formControlName],input[type=number][formControl],input[type=number][ngModel]',
-  host: {
-    '(change)': 'onChange($event.target.value)',
-    '(input)': 'onChange($event.target.value)',
-    '(blur)': 'onTouched()'
-  },
+  host: {'(input)': 'onChange($event.target.value)', '(blur)': 'onTouched()'},
   providers: [NUMBER_VALUE_ACCESSOR]
 })
 export class NumberValueAccessor implements ControlValueAccessor {
--- a/packages/forms/test/value_accessor_integration_spec.ts
+++ b/packages/forms/test/value_accessor_integration_spec.ts
@@ -149,6 +149,23 @@ import {dispatchEvent} from '@angular/platform-browser/testing/src/browser_util'
         expect(control.value).toEqual(0);
       });
 
+      it('should ignore the change event', () => {
+        const fixture = initTest(FormControlNumberInput);
+        const control = new FormControl();
+        fixture.componentInstance.control = control;
+        fixture.detectChanges();
+
+        control.valueChanges.subscribe({
+          next: (value) => {
+            throw 'Input[number] should not react to change event';
+          }
+        });
+        const input = fixture.debugElement.query(By.css('input'));
+
+        input.nativeElement.value = '5';
+        dispatchEvent(input.nativeElement, 'change');
+      });
+
       it('when value is cleared programmatically', () => {
         const fixture = initTest(FormControlNumberInput);
         const control = new FormControl(10);


petebacondarwin (approved):
LGTM
- the linting failure is because you need to format the test file correctly. I setup my VS Code to autorun the clang formatter on every save, which sorts this out
- also it would be good if we were able to explain, in the commit message, why the `(change)` event can now be removed; it was obviously included originally for a reason...

petebacondarwin:
I think this is the reason... https://github.com/angular/angular/issues/7557#issuecomment-253723850

petebacondarwin:
According to this we still support IE9 - https://angular.io/guide/browser-support
But I think that in this case there should be a polyfill to cover the problem of `input` not being triggered...

AndrewKushnir:
Hi @TidianeRolland, thanks for the PR!

Could you please provide a bit more information on this change to help me understand the problem better? Currently we fire 2 events: one after typing in the input field (as a result of the `oninput` event handler) and another one when the input field loses the focus (due to `onchange` event handler). Having both events fire at the same time is undesirable, but if I understand correctly, removing one of them might be a breaking change (losing focus will no longer trigger value change event, which might be used by developers)? I'm not proposing to keep the `change` host listener, just want to understand the impact of this change better.

I also [started a g3 presubmit](http://test/OCL:301937835:BASE:301937852:1584667844527:7412e067) (internal only link) to see if this change affects any targets in g3.

Thank you.

TidianeRolland:
Hi @AndrewKushnir thanks for the reply!

Yes, currently we fire 2 events: one after typing in the input field (as a result of the oninput event handler) and another one when the input field loses the focus (due to onchange event handler). Having both events fire at the same time is undesirable and I think having the onchange event trigger just when the input field loses the focus is also undesirable, according to this issue #12540. 
Because a lot of developers are subscribing to the valueChanges observable. valueChanges observable emits twice the same value of the input field.

I noticed that we're triggering the change event **only** for `input type number`. 
The other types of fields aren't concerned. 
I think the input event is enough for `input type number` too.
I hope I explained well.

Thanks for your help

petebacondarwin:
My understanding, although it may be lost in the annals of time - perhaps as far back as AngularJS, is that the extra handler was only added to support IE9, which appeared not to trigger the `input` event on some occasions. But @TidianeRolland was not able to recreate that issue with IE9 now, so it may have been fixed in the browser?

AndrewKushnir (changes requested):
Hi @TidianeRolland and @petebacondarwin,

Running global TAP indicated that there are few tests in g3 that rely on the `change` event to be present, here is an example code (simplified) from test harness:

```
setTextValue(element, text) {
    element.value = text;
    const changeEvent = new Event('change');
    element.dispatchEvent(changeEvent);
}
```

as a result, form control doesn't pick up an updated value. This is similar to the test that you've added as a part of this PR.

It can probably be refactored to trigger `input` event instead, but that would require additional time and investigation. I'm marking this PR with the "breaking change" flag, since there might be other cases outside g3 with the same behavior/expectations.

@TidianeRolland I also looked at the original issue report and I think it'd be great to add one more test using Forms API that would replicate original problem (and make sure that only one event is fired), so I set the status of my review to "requested changes".

Thank you.

atscott:
Thanks to @petebacondarwin for the tip on needing to test IE9 and @gkalpak for helping to test [the demo](https://stackblitz.com/edit/number-input-and-change-events?file=index.ts). 

It appears that the `input` event does not fire on IE9 when text in the input is removed via backspace or a cut action. Because IE9 is still a supported browser, I don't think we can merge this change. It would certainly be useful to add a comment about this in the code, noting the `change` event is needed for IE9 and can be removed when we drop support for it.

atscott:
What's the contract on the `valueChanges` event of the forms API? Are we able to change it to pipe through a `distinctUntilChanged`?

AndrewKushnir:
@atscott thanks for additional investigation! 👍

Just want to check if this problem only happens in IE9 or also present in IE10/11?

Thank you.

atscott:
> Just want to check if this problem only happens in IE9 or also present in IE10/11?

That's right - I was able to reproduce this missing `input` event for backspace in a protractor test. The only failure is in IE9, while it passes in chrome, IE10, and IE11.

TidianeRolland:
> Thanks to @petebacondarwin for the tip on needing to test IE9 and @gkalpak for helping to test [the demo](https://stackblitz.com/edit/number-input-and-change-events?file=index.ts).
> 
> It appears that the `input` event does not fire on IE9 when text in the input is removed via backspace or a cut action. 

@atscott it's really strange because I wasn't able to reproduce this bug using IE9 emulated in IE11 and IE9 on a Windows 7 machine. The input event was triggered

gkalpak:
@TidianeRolland, fwiw I was able to reproduce it using IE9 emulated in IE11 on Windows 10.
Note that the issue only appears when you delete characters from the value (e.g. hitting backspace or cutting). The `input` event works fine when typing more characters into the field.

atscott:
@TidianeRolland can you confirm that backspace/cut *does* fire the input event for you in IE9? I find it curious that it's working for you but not for me or George. You can use the [demo linked above](https://stackblitz.com/edit/number-input-and-change-events?file=index.ts) to test this.

TidianeRolland:
> @TidianeRolland can you confirm that backspace/cut _does_ fire the input event for you in IE9? I find it curious that it's working for you but not for me or George. You can use the [demo linked above](https://stackblitz.com/edit/number-input-and-change-events?file=index.ts) to test this.

I redid my tests. Backspace does not fire input event for emulated IE9.
If I weren't with a friend when I was doing my tests I would have thought I am crazy.
@atscott I confirm, stop listening to change event is not a good idea if we're supporting IE9

atscott:
> I redid my tests. Backspace does not fire input event for emulated IE9.

Thanks for retesting that. I'm adding the blocked label on this PR for now. @AndrewKushnir is adding an action item to discuss with the team what the plan is for IE9 support in the future. We'll get back to you with what can be done with this after the discussion.

atscott:
Hi @TidianeRolland - We discussed this issue today and decided that we can make this change and add a note in the breaking change on what devs can do if they want to maintain IE9 support.

Please update your commit message to include in your footer:

```
BREAKING CHANGE: <description> (tests that could break because they rely on 'change' event, requirement for devs to add their own `(change)` listener if they want to support IE9 in the same way it is now, etc)
```

TidianeRolland:
> Hi @TidianeRolland - We discussed this issue today and decided that we can make this change and add a note in the breaking change on what devs can do if they want to maintain IE9 support.
> 
> Please update your commit message to include in your footer:
> 
> ```
> BREAKING CHANGE: <description> (tests that could break because they rely on 'change' event, requirement for devs to add their own `(change)` listener if they want to support IE9 in the same way it is now, etc)
> ```

Hi @atscott - It's done

atscott:
One of the internal tests failures took me down a deep rabbit hole. It appears that this change may also break some selenium-webdriver tests if you have an old version. 

`WebElement.clear()` ([code link](https://github.com/pilwon/selenium-webdriver/blob/master/java/client/src/org/openqa/selenium/htmlunit/HtmlUnitWebElement.java#L265)) and `WebElement.sendKeys` ([code link](https://github.com/pilwon/selenium-webdriver/blob/master/java/client/src/org/openqa/selenium/htmlunit/HtmlUnitKeyboard.java#L77)) methods  call `htmlInput.setValueAttribute("");` which, until 2017 ([commit reference](https://github.com/HtmlUnit/htmlunit/commit/109bf5fba9988deccad6a1b6a05ed15f7d6cc550)), would trigger the `'change'` event on the input. From what I can tell right now, this means `sendKeys` and `clear` would trigger the `onChange` handler of the number input but no longer do.

Still investigating what can be done to resolve this issue.

atscott:
[global](https://test.corp.google.com/OCL:309466345:BASE:310163045:1588781226126:b824326) presubmit from a few days ago with a rerun of the failing started today. These should all be green now. New TGP scheduled for the 12:00 train.

AndrewKushnir (approved):

atscott:
[green TGP](https://test.corp.google.com/ui#id=OCL:310159075:BASE:310265362:1588813194323:a7033cca)

PR #36089, opened 2020-03-16
fix(ngcc): consistently delegate to TypeScript host for typing files

### LINKED ISSUE #35078: [Ivy] Incompatibility of Esm2015ReflectionHost with libraries compiled with NGCC
# 🐞 bug report

### Affected Package
The issue is caused by @angular/compiler-cli

### Is this a regression?
Sort of. ~~It is a regression if you happen to upgrade from a < 3.7 to a >= 3.7 version of TypeScript~~
It did work with Angular 6 and it also worked with previous Angular 9 rc's (I'm pretty sure that rc.7 still worked).

### Description
This was _very_ tricky to track down. The problem expressed in the form that the compiler complains when running ngcc on some of the project libraries that "selector/template has to be a string". This happened for cases where we use a constant from another package as the selector or template for a @Component annotation.

A lot of debugging of the build process and some guesswork resulted in the following findings:
* In essence, the Esm2015ReflectionHost is not capable of resolving expressions of the form `MyClass.MyPublicStaticStringProperty`... in _some situations_.
* In the cases where it fails, the `reflectMember` check for `isPropertyAccess` simply does not work anymore because the syntax kind of the node is something completely unexpected.
* In `getMembersOfClass` method, the nodes of `symbol.implementation.exports` for public static properties are of syntax kind `PropertyDeclaration` when "normally" (when things work), they are of kind `PropertyAccessExpression`
* In the end, it all breaks down in the `StaticInterpreter.accessHelper` method because it ends up returning a `Reference` instead of the resolved string constant. This then triggers various extremely unhelpful downstream errors.

## 🔬 Minimal Reproduction
I don't have a minimal reproduction right now since I'm not sure how I can "build" this situation from scratch, but as soon as I find one, I will add it here. See my next issue comment for reproduction.

An "abstraction" of the code I am working with contains a component in a library in the "packages" directory
```
@Component({
    selector: SharedConstants.SecondSelector,
    template: `<div>...</div>`,
})
export class SharedDerivedComponent extends BaseComponent {
}
```
This is referencing a constant from another library in the "packages" directory
```
export declare class SharedConstants {
    static readonly SecondSelector = "lib-shared-derived";
}
```

## 🔥 Exception or Error
Some of the messages that result from this are:
<pre><code>
NG1010: selector must be a string
</code></pre>

<pre><code>
NG1010: template must be a string
</code></pre>

<pre><code>
NG1006: Two incompatible decorators on class
</code></pre>

## 🌍  Your Environment

**Angular Version:**
<pre><code>
Angular CLI: 9.0.0-rc.11
Node: 10.16.0
OS: win32 x64

Angular: 9.0.0-rc.11
Ivy Workspace: Yes

@angular-devkit/architect          0.900.0-rc.11
@angular-devkit/build-angular      0.900.0-rc.11
@angular-devkit/build-ng-packagr   0.900.0-rc.11
@angular-devkit/build-optimizer    0.900.0-rc.11
@angular-devkit/build-webpack      0.900.0-rc.11
@angular-devkit/core               9.0.0-rc.11
@angular-devkit/schematics         9.0.0-rc.11
@ngtools/webpack                   9.0.0-rc.11
@schematics/angular                9.0.0-rc.11
@schematics/update                 0.900.0-rc.11
ng-packagr                         9.0.0-rc.7
webpack                            4.8.3
</code></pre>

JoostK:
When ngcc is compiling an entry-point, it uses a `ReflectionHost` that
is specific to its format, e.g. ES2015, ES5, UMD or CommonJS. During the
compilation of that entry-point however, the reflector may be used to
reflect into external libraries using their declaration files.

Up until now this was achieved by letting all `ReflectionHost` classes
consider their parent class for reflector queries, thereby ending up in
the `TypeScriptReflectorHost` that is a common base class for all
reflector hosts. This approach has proven to be prone to bugs, as
failing to call into the base class would cause incompatibilities with
reading from declaration files.

The observation can be made that there's only two distinct kinds of
reflection host queries:
1. the reflector query is about code that is part of the entry-point
   that is being compiled, or
2. the reflector query is for an external library that the entry-point
   depends on, in which case the information is reflected
   from the declaration files.

The `ReflectionHost` that was chosen for the entry-point should serve
only reflector queries for the first case, whereas a regular
`TypeScriptReflectionHost` should be used for the second case. This
avoids avoids the problem where a format-specific `ReflectionHost` fails
to handle the second case correctly, as it isn't even considered for
such reflector queries.

This commit introduces a `ReflectionHost` that delegates to the
`TypeScriptReflectionHost` for AST nodes within declaration files,
otherwise delegating to the format-specific `ReflectionHost`.

Fixes #35078
Resolves FW-1859

### DIFF
--- a/packages/compiler-cli/ngcc/src/host/commonjs_host.ts
+++ b/packages/compiler-cli/ngcc/src/host/commonjs_host.ts
@@ -34,11 +34,6 @@ export class CommonJsReflectionHost extends Esm5ReflectionHost {
   }
 
   getImportOfIdentifier(id: ts.Identifier): Import|null {
-    const superImport = super.getImportOfIdentifier(id);
-    if (superImport !== null) {
-      return superImport;
-    }
-
     const requireCall = this.findCommonJsImport(id);
     if (requireCall === null) {
       return null;
@@ -47,8 +42,7 @@ export class CommonJsReflectionHost extends Esm5ReflectionHost {
   }
 
   getDeclarationOfIdentifier(id: ts.Identifier): Declaration|null {
-    return (!id.getSourceFile().isDeclarationFile && this.getCommonJsImportedDeclaration(id)) ||
-        super.getDeclarationOfIdentifier(id);
+    return this.getCommonJsImportedDeclaration(id) || super.getDeclarationOfIdentifier(id);
   }
 
   getExportsOfModule(module: ts.Node): Map<string, Declaration>|null {
--- a/packages/compiler-cli/ngcc/src/host/delegating_host.ts
+++ b/packages/compiler-cli/ngcc/src/host/delegating_host.ts
@@ -1,0 +1,147 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import * as ts from 'typescript';
+
+import {ClassDeclaration, ClassMember, CtorParameter, Declaration, Decorator, FunctionDefinition, Import, ReflectionHost} from '../../../src/ngtsc/reflection';
+import {isFromDtsFile} from '../../../src/ngtsc/util/src/typescript';
+
+import {ModuleWithProvidersFunction, NgccClassSymbol, NgccReflectionHost, SwitchableVariableDeclaration} from './ngcc_host';
+
+export class DelegatingReflectionHost implements NgccReflectionHost {
+  constructor(private tsHost: ReflectionHost, private ngccHost: NgccReflectionHost) {}
+
+  getConstructorParameters(clazz: ClassDeclaration): CtorParameter[]|null {
+    if (isFromDtsFile(clazz)) {
+      return this.tsHost.getConstructorParameters(clazz);
+    }
+    return this.ngccHost.getConstructorParameters(clazz);
+  }
+
+  getDeclarationOfIdentifier(id: ts.Identifier): Declaration|null {
+    if (isFromDtsFile(id)) {
+      return this.tsHost.getDeclarationOfIdentifier(id);
+    }
+    return this.ngccHost.getDeclarationOfIdentifier(id);
+  }
+
+  getDecoratorsOfDeclaration(declaration: ts.Declaration): Decorator[]|null {
+    if (isFromDtsFile(declaration)) {
+      return this.tsHost.getDecoratorsOfDeclaration(declaration);
+    }
+    return this.ngccHost.getDecoratorsOfDeclaration(declaration);
+  }
+
+  getDefinitionOfFunction(fn: ts.Node): FunctionDefinition|null {
+    if (isFromDtsFile(fn)) {
+      return this.tsHost.getDefinitionOfFunction(fn);
+    }
+    return this.ngccHost.getDefinitionOfFunction(fn);
+  }
+
+  getDtsDeclaration(declaration: ts.Declaration): ts.Declaration|null {
+    if (isFromDtsFile(declaration)) {
+      return this.tsHost.getDtsDeclaration(declaration);
+    }
+    return this.ngccHost.getDtsDeclaration(declaration);
+  }
+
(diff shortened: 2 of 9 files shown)


petebacondarwin (approved):
Great stuff @JoostK ! 
Did you do any performance benchmarks? I don't expect it to be a big difference since the majority of the cost is in creating the original TS program, and you have removed some complexity from the individual ngcc reflectionhosts, but it would be interesting to see...

JoostK:
@petebacondarwin I have not, but I'm not expecting it to become slower. Reflected nodes are typically not deep in the AST so asking for the source file should be relatively few parent pointer reads.

JoostK:
It's green on ngcc-validation: https://github.com/angular/ngcc-validation/pull/1015

packages/compiler-cli/ngcc/src/host/delegating_host.ts
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  isClass(node: ts.Node): node is ClassDeclaration {
+    if (isFromDtsFile(node)) {
+      return this.tsHost.isClass(node);
+    }
+    return this.ngccHost.isClass(node);
+  }
+
+  findClassSymbols(sourceFile: ts.SourceFile): NgccClassSymbol[] {
```
gkalpak: Consider adding a comment here that these are `NgccReflectionHost`-specific methods (not present in `ReflectionHost`) for future readers. It caught me off-guard for a couple of minutes 😅

packages/compiler-cli/ngcc/test/integration/util.ts
```diff
@@ -90,7 +96,7 @@ function compileIntoFlatPackage(
   const pkgJson: unknown = {
     name: pkgName,
     version: '0.0.1',
-    main: './index.js',
+    [options.formatProperty]: './index.js',
```
gkalpak: :100:

gkalpak (approved):
Nice 👍 ✨ 👌
(Just a couple minor questions/suggestions.)

gkalpak:
Nit: There is also a typo in the first commit message: `avoids avoids` --> `avoids`

packages/compiler-cli/ngcc/test/host/umd_host_spec.ts
```diff
@@ -1942,39 +1942,6 @@ runInEachFileSystem(() => {
...
-           if (x.initializer === undefined || !ts.isIdentifier(x.initializer)) {
-             return fail('Expected constant `x` to have an identifer as an initializer.');
-           }
-           const decl = host.getDeclarationOfIdentifier(x.initializer);
-           if (decl === null) {
-             return fail('Expected to find a declaration for ' + x.initializer.getText());
-           }
-           expect(decl.viaModule).toEqual('sub_module');
-           expect(decl.node).toBe(expectedDeclaration);
-         });
```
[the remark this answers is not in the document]
JoostK: Yeah, because the functionality has been removed from the hosts' implementation it won't work for AST nodes in declaration files anymore.

packages/compiler-cli/ngcc/src/host/commonjs_host.ts
```diff
@@ -34,11 +34,6 @@ export class CommonJsReflectionHost extends Esm5ReflectionHost {
   }
 
   getImportOfIdentifier(id: ts.Identifier): Import|null {
-    const superImport = super.getImportOfIdentifier(id);
-    if (superImport !== null) {
-      return superImport;
-    }
-
```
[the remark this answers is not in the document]
JoostK: Good spot, I removed this. Git blame attributed this to a commit to fix a bug with missing support for declaration files, so it can indeed be removed.

packages/compiler-cli/ngcc/src/host/delegating_host.ts
JoostK: I added a comment in between the regular `ReflectionHost` methods and the `NgccReflectionHost` queries, as well as a brief JSDoc on the class itself.

PR #36091, opened 2020-03-16
feat(dev-infra): create common config file loading util

josephperrott:
Create a common config file loading utility function and the
necessary util directory.  This util directory can provide common
utility functions for usage inside of the dev-infra package.


dev-infra/utils/config.ts
```diff
@@ -0,0 +1,46 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import {parse} from 'comment-json';
```
devversion: Do you want to use `json5`? It supports comments as in TypeScript tsconfig files (which is pretty useful; and would be consistent): https://www.npmjs.com/package/json5

dev-infra/utils/BUILD.bazel
```diff
@@ -0,0 +1,16 @@
+load("@npm_bazel_typescript//:index.bzl", "ts_library")
```
devversion: Optional nit: Use `defaults.bzl`?

dev-infra/utils/config.ts
```diff
@@ -0,0 +1,46 @@
...
+export function getAngularDevConfig(): DevInfraConfig {
+  let configPath = '';
+  let rawConfig = '';
+  try {
+    configPath = join(getRepoBaseDir(), '.dev-infra.json');
+    rawConfig = readFileSync(configPath, 'utf8');
+  } catch {
+    throw Error(
+        `Unable to find config file at:\n` +
+        `  ${configPath}`);
```
gkalpak: This error will be confusing if `getRepoBaseDir()` throws.
I suggest:

```suggestion
  const configPath = join(getRepoBaseDir(), '.dev-infra.json');
  let rawConfig = '';
  try {
    rawConfig = readFileSync(configPath, 'utf8');
  } catch {
    throw Error(
        `Unable to find config file at:\n` +
        `  ${configPath}`);
```

dev-infra/utils/config.ts
gkalpak: `comment-json` is the only one that preserves the comments, though. That sounds quite useful (e.g. for when you want to modify a config file in place) - not used here yet, but :grin:

From their [README][1]:

> There are many other libraries that can deal with JSON with comments, such as `json5`, or `strip-json-comments`, but none of them can stringify the parsed object and return back a JSON string the same as the original content.

[1]: https://www.npmjs.com/package/comment-json#why

dev-infra/utils/BUILD.bazel
josephperrott: My reasoning for not using the `defaults.bzl` version is that it brings in unnecessary dependencies, specifically in the `testonly` attribute case that causes a dependency on the angular specific pieces of the toolchain.

This could lead to less cachability and longer compilation times specific to these rules.

dev-infra/utils/config.ts
josephperrott: Let's go ahead and use `json5` for now since we aren't relying on modifying in place at this point.  But we can keep that in mind for if we need the capability.

Does that sound good?

dev-infra/utils/BUILD.bazel
devversion: 👍

dev-infra/utils/config.ts
gkalpak: Works for me :+1:

dev-infra/utils/BUILD.bazel
```diff
@@ -0,0 +1,16 @@
...
+ts_library(
+    name = "config",
+    srcs = [
+        "config.ts",
+    ],
+    visibility = ["//dev-infra:__subpackages__"],
+    deps = [
+        "@npm//@types/node",
+        "@npm//@types/shelljs",
+        "@npm//comment-json",
```
gkalpak: This now needs to be switched to `json5` (and `@types/json5`?) too.

```suggestion
        "@npm//@types/json5",
        "@npm//@types/node",
        "@npm//@types/shelljs",
        "@npm//json5",
```

gkalpak (approved):
LGTM as soon as CI is happy 👍

devversion (approved):
LGTM (aside from @gkalpak's comment)

PR #36096, opened 2020-03-17
build: fix ts-api-guardian golden approval not working on windows

devversion:
Currently on Windows, it's not possible to approve goldens in
`ts-api-guardian`. This is because paths are resolved relatively
to the working directory. In Windows, golden files are resolved
to the actual workspace directory. The current logic tries to
compute a relative path to the runfile from the working directory.

This causes the file paths to have a lot of parent directory
path segments. Eventually, when joined with the build workspace
directory, the paths end up being incorrect. e.g.

```
fileName = ../../../../../../projects/angular/golden/<..>/common.d.ts`
outFile = BUILD_WORKSPACE_DIR + fileName;
```

To fix this, we no longer deal with confusing relative paths, but
instead always use absolute file system paths.

Additionally, this fixes that new goldens are generated at the wrong
location on all platforms.

### DIFF
--- a/tools/ts-api-guardian/lib/cli.ts
+++ b/tools/ts-api-guardian/lib/cli.ts
@@ -16,7 +16,22 @@ import * as path from 'path';
 
 import {SerializationOptions, generateGoldenFile, verifyAgainstGoldenFile, discoverAllEntrypoints} from './main';
 
+/** Name of the CLI */
 const CMD = 'ts-api-guardian';
+
+/** Name of the Bazel workspace that runs the CLI. */
+const bazelWorkspaceName = process.env.BAZEL_WORKSPACE;
+/**
+ * Path to the Bazel workspace directory. Only set if the CLI is run with `bazel run`.
+ * https://docs.bazel.build/versions/master/user-manual.html#run.
+ */
+const bazelWorkspaceDirectory = process.env.BUILD_WORKSPACE_DIRECTORY;
+/**
+ * Regular expression that matches Bazel manifest paths that start with the
+ * current Bazel workspace, followed by a path delimiter.
+ */
+const bazelWorkspaceManifestPathRegex =
+    bazelWorkspaceName ? new RegExp(`^${bazelWorkspaceName}[/\\\\]`) : null;
 
 export function startCli() {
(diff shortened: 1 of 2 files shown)


tools/ts-api-guardian/lib/cli.ts
```diff
@@ -210,44 +225,57 @@ Options:
...
+    return path.resolve(fileName);
+  }
+  // In Bazel, we first try to resolve the file through the runfiles. We do this by calling
+  // the `require.resolve` function that is patched by the Bazel NodeJS rules. Note that we
+  // need to catch errors because files inside tree artifacts cannot be resolved through
+  // runfile manifests. Hence, we need to have alternative resolution logic when resolving
+  // file paths. Additionally, it could happen that manifest paths which aren't part of the
+  // runfiles are specified (i.e. golden is approved but does not exist in the workspace yet).
+  try {
+    return require.resolve(fileName);
```
josephperrott: While I believe this will work on both unix and windows systems, I don't believe that it corrects the issue of not creating a new file when `.accept` is run.

Since the `require.resolve` correctly "finds" the location that exists in the runfiles, but that does not yet symlinked to a file that its in the repo, so it gets left in the runfiles and isn't copied back over.

This causes it to fails to create the new file silently.

josephperrott (changes requested):

tools/ts-api-guardian/lib/cli.ts
devversion: I tested manually on Windows and macOS. The `require.resolve` handles the resolution for runfiles. The code below handles the case where a new golden is generated (if it not exists yet).

Writing to the file in symlinked position seems to behave as expected, and the content of the actual target file is updated correctly.

josephperrott: I am testing again now, it failed for me on macOS, though I am wondering if I saw a failure because I was running it with RBE

josephperrott: Confirmed, I saw the failure because I was running with RBE.  I think we should be good to commit this now since that is the case.

josephperrott (approved):
LGTM

PR #36098, opened 2020-03-17
ci: cache downloaded bazel version in circleci

devversion:
Similarly to what is done in `angular/components`, we should
cache the downloaded Bazel version (from `bazelisk`).

This reduces the overhead of downloading Bazel, and also avoids
the dependency on the external download server. We should avoid
external server dependencies as much as possible (see how the yarn
registry was flaky in the past).


gregmagolan (approved):
LGTM. Nice optimization!

josephperrott (approved):
LGTM, good call.  Thanks!

PR #36106, opened 2020-03-17
fix(compiler): handle type references to namespace imports correctly

### LINKED ISSUE #36006: ng serve fails with "ERROR in Debug failure: no local name provided for NamespaceImport"
# 🐞 Bug report

- [ ] new
- [ ] build
- [x] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?

Yes, the same app was working with Angular 8 and CLI 8.

### Description

Running ng serve consistently fails with the stack trace reported below.

## 🔬 Minimal Reproduction
npm install
ng serve
It's a private repo and the exception does not refer to any specific code in the repo. It seems to be a generic error in compiler-cli itself.

## 🔥 Exception or Error
<pre><code>
ERROR in Debug failure: no local name provided for NamespaceImport
F:\webstr\node_modules\@angular\compiler-cli\src\ngtsc\core\src\compiler.js:342
                finally { if (e_2) throw e_2.error; }
                                   ^

Error: Debug failure: no local name provided for NamespaceImport
    at extractModuleAndNameFromImport (F:\webstr\node_modules\@angular\compiler-cli\src\ngtsc\reflection\src\type_to_value.js:173:27)
    at Object.typeToValue (F:\webstr\node_modules\@angular\compiler-cli\src\ngtsc\reflection\src\type_to_value.js:59:28)
    at F:\webstr\node_modules\@angular\compiler-cli\src\ngtsc\reflection\src\typescript.js:76:58
    at Array.map (<anonymous>)
    at TypeScriptReflectionHost.getConstructorParameters (F:\webstr\node_modules\@angular\compiler-cli\src\ngtsc\reflection\src\typescript.js:55:36)
    at Object.getConstructorDependencies (F:\webstr\node_modules\@angular\compiler-cli\src\ngtsc\annotations\src\util.js:32:36)
    at extractInjectableCtorDeps (F:\webstr\node_modules\@angular\compiler-cli\src\ngtsc\annotations\src\injectable.js:233:38)
    at InjectableDecoratorHandler.analyze (F:\webstr\node_modules\@angular\compiler-cli\src\ngtsc\annotations\src\injectable.js:71:31)
    at TraitCompiler.analyzeTrait (F:\webstr\node_modules\@angular\compiler-cli\src\ngtsc\transform\src\compilation.js:341:40)
    at analyze (F:\webstr\node_modules\@angular\compiler-cli\src\ngtsc\transform\src\compilation.js:293:58)

</code></pre>

## 🌍 Your Environment
<pre><code>
Angular CLI: 9.0.5
Node: 12.13.1
OS: win32 x64

Angular: 9.0.5
... router
Ivy Workspace: Yes

@angular-devkit/architect         0.900.5
@angular-devkit/build-angular     0.900.5
@angular-devkit/build-optimizer   0.900.5
@angular-devkit/build-webpack     0.900.5
@angular-devkit/core              9.0.5
@angular-devkit/schematics        9.0.5
@angular/cdk                      9.1.2
@angular/material                 9.1.2
@ngtools/webpack                  9.0.5
@schematics/angular               9.0.5
@schematics/update                0.900.5

</code></pre>

**Anything else relevant?**

JoostK:
fix(compiler): handle type references to namespaced symbols correctly

When the compiler needs to convert a type reference to a value
expression, it may encounter a type that refers to a namespaced symbol.
Such namespaces need to be handled specially as there's various forms
available. Consider a namespace named "ns":

1. One can refer to a namespace by itself: `ns`. A namespace is only
   allowed to be used in a type position if it has been merged with a
   class, but even if this is the case it may not be possible to convert
   that type into a value expression depending on the import form. More
   on this later (case a below)
2. One can refer to a type within the namespace: `ns.Foo`. An import
   needs to be generated to `ns`, from which the `Foo` property can then
   be read.
3. One can refer to a type in a nested namespace within `ns`:
   `ns.Foo.Bar` and possibly even deeper nested. The value
   representation is similar to case 2, but includes additional property
   accesses.

The exact strategy of how to deal with these cases depends on the type
of import used. There's two flavors available:

a. A namespaced import like `import * as ns from 'ns';` that creates
   a local namespace that is irrelevant to the import that needs to be
   generated (as said import would be used instead of the original
   import).

   If the local namespace "ns" itself is referred to in a type position,
   it is invalid to convert it into a value expression. Some JavaScript
   libraries publish a value as default export using `export = MyClass;`
   syntax, however it is illegal to refer to that value using "ns".
   Consequently, such usage in a type position *must* be accompanied by
   an `@Inject` decorator to provide an explicit token.

b. An explicit namespace declaration within a module, that can be
   imported using a named import like `import {ns} from 'ns';` where the
   "ns" module declares a namespace using `declare namespace ns {}`.
   In this case, it's the namespace itself that needs to be imported,
   after which any qualified references into the namespace are converted
   into property accesses.

Before this change, support for namespaces in the type-to-value
conversion was limited and only worked  correctly for a single qualified
name using a namespace import (case 2a). All other cases were either
producing incorrect code or would crash the compiler (case 1a).

Crashing the compiler is not desirable as it does not indicate where
the issue is. Moreover, the result of a type-to-value conversion is
irrelevant when an explicit injection token is provided using `@Inject`,
so referring to a namespace in a type position (case 1) could still be
valid.

This commit introduces logic to the type-to-value conversion to be able
to properly deal with all type references to namespaced symbols.

Fixes #36006
Resolves FW-1995

### DIFF
--- a/packages/compiler-cli/src/ngtsc/reflection/src/host.ts
+++ b/packages/compiler-cli/src/ngtsc/reflection/src/host.ts
@@ -243,8 +243,24 @@ export type TypeValueReference = {
   local: true; expression: ts.Expression; defaultImportStatement: ts.ImportDeclaration | null;
 }|{
   local: false;
-  name: string;
+
+  /**
+   * The module specifier from which the `importedName` symbol should be imported.
+   */
   moduleName: string;
+
+  /**
+   * The name of the top-level symbol that is imported from `moduleName`. If `nestedPath` is also
+   * present, a nested object is being referenced from the top-level symbol.
+   */
+  importedName: string;
+
+  /**
+   * If present, represents the symbol names that are referenced from the top-level import.
+   * When `null` or empty, the `importedName` itself is the symbol being referenced.
+   */
+  nestedPath: string[]|null;
+
   valueDeclaration: ts.Declaration;
 };
 
--- a/packages/compiler-cli/src/ngtsc/reflection/src/type_to_value.ts
+++ b/packages/compiler-cli/src/ngtsc/reflection/src/type_to_value.ts
@@ -42,30 +42,75 @@ export function typeToValue(
   // Look at the local `ts.Symbol`'s declarations and see if it comes from an import
   // statement. If so, extract the module specifier and the name of the imported type.
   const firstDecl = local.declarations && local.declarations[0];
+  if (firstDecl !== undefined) {
+    if (ts.isImportClause(firstDecl) && firstDecl.name !== undefined) {
+      // This is a default import.
+      //   import Foo from 'foo';
 
-  if (firstDecl && ts.isImportClause(firstDecl) && firstDecl.name !== undefined) {
-    // This is a default import.
-    return {
-      local: true,
-      // Copying the name here ensures the generated references will be correctly transformed along
-      // with the import.
-      expression: ts.updateIdentifier(firstDecl.name),
-      defaultImportStatement: firstDecl.parent,
-    };
-  } else if (firstDecl && isImportSource(firstDecl)) {
-    const origin = extractModuleAndNameFromImport(firstDecl, symbols.importName);
-    return {local: false, valueDeclaration: decl.valueDeclaration, ...origin};
-  } else {
-    const expression = typeNodeToValueExpr(typeNode);
-    if (expression !== null) {
       return {
         local: true,
-        expression,
-        defaultImportStatement: null,
+        // Copying the name here ensures the generated references will be correctly transformed
+        // along with the import.
+        expression: ts.updateIdentifier(firstDecl.name),
+        defaultImportStatement: firstDecl.parent,
       };
-    } else {
-      return null;
+    } else if (ts.isImportSpecifier(firstDecl)) {
+      // The symbol was imported by name
+      //   import {Foo} from 'foo';
+      // or
+      //   import {Foo as Bar} from 'foo';
+
+      // Determine the name to import (`Foo`) from the import specifier, as the symbol names of
+      // the imported type could refer to a local alias (like `Bar` in the example above).
+      const importedName = (firstDecl.propertyName || firstDecl.name).text;
+
+      // The first symbol name refers to the local name, which is replaced by `importedName` above.
+      // Any remaining symbol names make up the complete path to the value.
+      const [_localName, ...nestedPath] = symbols.symbolNames;
+
+      const moduleName = extractModuleName(firstDecl.parent.parent.parent);
+      return {
+        local: false,
+        valueDeclaration: decl.valueDeclaration,
+        moduleName,
+        importedName,
+        nestedPath
+      };
+    } else if (ts.isNamespaceImport(firstDecl)) {
+      // The import is a namespace import
+      //   import * as Foo from 'foo';
+
+      if (symbols.symbolNames.length === 1) {
+        // The type refers to the namespace itself, which cannot be represented as a value.
+        return null;
+      }
+
+      // The first symbol name refers to the local name of the namespace, which is is discarded
+      // as a new namespace import will be generated. This is followed by the symbol name that needs
+      // to be imported and any remaining names that constitute the complete path to the value.
+      const [_ns, importedName, ...nestedPath] = symbols.symbolNames;
+
+      const moduleName = extractModuleName(firstDecl.parent.parent);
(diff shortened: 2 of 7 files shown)


packages/compiler-cli/src/ngtsc/reflection/src/type_to_value.ts
```diff
@@ -42,30 +42,57 @@ export function typeToValue(
...
+    } else if (ts.isImportSpecifier(firstDecl)) {
+      // The symbol was imported by name
+      //   import {Foo} from 'foo';
+      // or
+      //   import {Foo as Bar} from 'foo';
+
+      // The name is the aliased name if available (`Bar`) or the declared name if omitted (`Foo`).
+      const name = (firstDecl.propertyName || firstDecl.name).text;
+      const moduleName = extractModuleName(firstDecl.parent.parent.parent);
+      return {local: false, valueDeclaration: decl.valueDeclaration, moduleName, name, path: null};
```
JoostK: I just learned that we do need to do something with `symbols.path`, as the imported symbol may be a namespace so that a qualified type reference of more than two parts could be used.

packages/compiler-cli/src/ngtsc/reflection/src/host.ts
```diff
@@ -246,6 +246,7 @@ export type TypeValueReference = {
   local: false;
   name: string;
   moduleName: string;
+  path: string[]|null;
```
alxhub: The existence of `path` here changes the meaning of `name` - sometimes `name` is the thing being referenced, and sometimes it's just the top-level import of a longer access path.

I think we should rename `name` to `importedName` as a result, and `path` to `nestedPath`, and add some docs here clarifying that `importedName` is always the name of the top-level symbol being imported, and if `nestedPath` is present then a nested object is being referenced from the top-level import.

packages/compiler-cli/src/ngtsc/reflection/src/type_to_value.ts
```diff
@@ -42,31 +42,63 @@ export function typeToValue(
...
+      //   import {Foo} from 'foo';
+      // or
+      //   import {Foo as Bar} from 'foo';
+
+      // Determine the name to import (`Foo`) from the import specifier, as the symbol names of
+      // the imported type could refer to a local alias (like `Bar` in the example above).
+      const name = (firstDecl.propertyName || firstDecl.name).text;
+
+      // The remaining symbol names constitute the complete path to the value.
+      const path = symbols.symbolNames.slice(1);
```
alxhub: For symmetry:
```suggestion
      const [_ns, ...path] = symbols.symbolNames;
```

alxhub (approved):

packages/compiler-cli/src/ngtsc/reflection/src/type_to_value.ts
JoostK: That's what I initially had, but changed into this. But I actually like that explicitness, so I'll update per your suggestion.

packages/compiler-cli/src/ngtsc/reflection/src/host.ts
JoostK: Thanks for the suggestions, those names do indeed make it more clear.

IgorMinar:
hey @JoostK can you please rebase this and @alxhub can you please presubmit? thanks

JoostK:
@IgorMinar Done.

alxhub:
[Presubmit](https://test.corp.google.com/ui#id=OCL:303404593:BASE:303404624:1585343651674:64581d28)

scott-ho:
any update?

kara:
@JoostK rebase?

kara:
G3 failed with

```
third_party/javascript/angular2/rc/packages/compiler-cli/src/ngtsc/annotations/src/util.ts:141:48 - error TS2339: Property 'moduleName' does not exist on type 'TypeValueReference'.
  Property 'moduleName' does not exist on type '{ local: true; expression: Expression; defaultImportStatement: ImportDeclaration; }'.

141         new ExternalExpr({moduleName: valueRef.moduleName, name: valueRef.importedName});
```

PR #36114, opened 2020-03-17
fix(elements): a couple of small fixes

### LINKED ISSUE #30848: Angular Elements doesn't bind existing property values to component inputs
# 🐞 bug report

### Affected Package
The issue is caused by package @angular/elements

### Is this a regression?
Unsure.

### Description
If a DOM element has existing properties prior to being loaded/initialized by Angular Elements, they are not bound to component inputs.

## 🔬 Minimal Reproduction
https://stackblitz.com/edit/angular-elements-test-existing-bindings

## 🔥 Exception or Error
No error occurs, properties are not bound.

## 🌍  Your Environment

**Angular Version:**
<pre><code>
Angular CLI: 7.3.9
Node: 10.14.2
OS: darwin x64
Angular: 7.2.15
... forms, http, language-service, platform-browser

@angular-devkit/architect         0.13.9
@angular-devkit/build-angular     0.13.9
@angular-devkit/build-optimizer   0.13.9
@angular-devkit/build-webpack     0.13.9
@angular-devkit/core              7.3.9
@angular-devkit/schematics        7.3.9
@angular/cdk                      7.2.2
@angular/cli                      7.3.9
@ngtools/webpack                  7.3.9
@schematics/angular               7.3.9
@schematics/update                0.13.9
typescript                        3.2.4
webpack                           4.29.0
</code></pre>

gkalpak:
Several small fixes and refactorings. See individual commits for details.
This is an alternative to #31416 (with the feedback there addressed).

Jira issue: [FW-2006][1]

Fixes #30848.

##
NOTE:
The last commit is a fix to the public API. It should not affect users in any way (afaict), but we might still want to only merge that last commit to master-only. (The rest is good for both master and patch.)
Happy to move the last commit to a separate PR, if we decide it shouldn't go into patch.

_UPDATE: The last commit has been removed. See https://github.com/angular/angular/pull/36114#issuecomment-622378247._

[1]: https://angular-team.atlassian.net/browse/FW-2006

[FW-2006]: https://angular-team.atlassian.net/browse/FW-2006

### DIFF
--- a/packages/elements/src/create-custom-element.ts
+++ b/packages/elements/src/create-custom-element.ts
@@ -144,8 +144,9 @@ export function createCustomElement<P>(
       //
       // TODO(andrewseguin): Add e2e tests that cover cases where the constructor isn't called. For
       // now this is tested using a Google internal test suite.
-      if (this._ngElementStrategy === null) {
-        const strategy = this._ngElementStrategy = strategyFactory.create(this.injector);
+      if (!this._ngElementStrategy) {
+        const strategy = this._ngElementStrategy =
+            strategyFactory.create(this.injector || config.injector);
 
         // Collect pre-existing values on the element to re-apply through the strategy.
         const preExistingValues =
@@ -173,12 +174,10 @@ export function createCustomElement<P>(
       return this._ngElementStrategy!;
     }
 
-    private readonly injector: Injector;
-    private _ngElementStrategy: NgElementStrategy|null = null;
+    private _ngElementStrategy?: NgElementStrategy;
 
-    constructor(injector?: Injector) {
+    constructor(private readonly injector?: Injector) {
       super();
-      this.injector = injector || config.injector;
     }
 
     attributeChangedCallback(
--- a/packages/elements/src/create-custom-element.ts
+++ b/packages/elements/src/create-custom-element.ts
@@ -145,9 +145,32 @@ export function createCustomElement<P>(
       // TODO(andrewseguin): Add e2e tests that cover cases where the constructor isn't called. For
       // now this is tested using a Google internal test suite.
       if (this._ngElementStrategy === null) {
-        this._ngElementStrategy = strategyFactory.create(this.injector);
+        const strategy = this._ngElementStrategy = strategyFactory.create(this.injector);
+
+        // Collect pre-existing values on the element to re-apply through the strategy.
+        const preExistingValues =
+            inputs.filter(({propName}) => this.hasOwnProperty(propName)).map(({propName}): [
+              string, any
+            ] => [propName, (this as any)[propName]]);
+
+        // In some browsers (e.g. IE10), `Object.setPrototypeOf()` (which is required by some Custom
+        // Elements polyfills) is not defined and is thus polyfilled in a way that does not preserve
+        // the prototype chain. In such cases, `this` will not be an instance of `NgElementImpl` and
+        // thus not have the component input getters/setters defined on `NgElementImpl.prototype`.
+        if (!(this instanceof NgElementImpl)) {
+          // Add getters and setters to the instance itself for each property input.
+          defineInputGettersSetters(inputs, this);
+        } else {
+          // Delete the property from the instance, so that it can go through the getters/setters
+          // set on `NgElementImpl.prototype`.
+          preExistingValues.forEach(([propName]) => delete (this as any)[propName]);
+        }
+
+        // Re-apply pre-existing values through the strategy.
+        preExistingValues.forEach(([propName, value]) => strategy.setInputValue(propName, value));
       }
-      return this._ngElementStrategy;
+
+      return this._ngElementStrategy!;
     }
 
     private readonly injector: Injector;
@@ -193,16 +216,26 @@ export function createCustomElement<P>(
   // Update the property descriptor of `NgElementImpl#ngElementStrategy` to make it enumerable.
   Object.defineProperty(NgElementImpl.prototype, 'ngElementStrategy', {enumerable: true});
 
-  // Add getters and setters to the prototype for each property input. If the config does not
-  // contain property inputs, use all inputs by default.
-  inputs.map(({propName}) => propName).forEach(property => {
-    Object.defineProperty(NgElementImpl.prototype, property, {
-      get: function() { return this.ngElementStrategy.getInputValue(property); },
-      set: function(newValue: any) { this.ngElementStrategy.setInputValue(property, newValue); },
+  // Add getters and setters to the prototype for each property input.
+  defineInputGettersSetters(inputs, NgElementImpl.prototype);
+
+  return (NgElementImpl as any) as NgElementConstructor<P>;
+}
+
+// Helpers
+function defineInputGettersSetters(
+    inputs: {propName: string, templateName: string}[], target: object): void {
+  // Add getters and setters for each property input.
+  inputs.forEach(({propName}) => {
+    Object.defineProperty(target, propName, {
+      get(): any {
+        return this.ngElementStrategy.getInputValue(propName);
+      },
+      set(newValue: any): void {
+        this.ngElementStrategy.setInputValue(propName, newValue);
+      },
       configurable: true,
       enumerable: true,
     });
   });
-
-  return (NgElementImpl as any) as NgElementConstructor<P>;
 }
--- a/packages/elements/src/create-custom-element.ts
+++ b/packages/elements/src/create-custom-element.ts
@@ -187,9 +187,15 @@ export function createCustomElement<P>(
   inputs.map(({propName}) => propName).forEach(property => {
     Object.defineProperty(NgElementImpl.prototype, property, {
       get: function() {
+        if (!this.ngElementStrategy) {
+          this.ngElementStrategy = strategyFactory.create(config.injector);
+        }
         return this.ngElementStrategy.getInputValue(property);
       },
       set: function(newValue: any) {
+        if (!this.ngElementStrategy) {
+          this.ngElementStrategy = strategyFactory.create(config.injector);
+        }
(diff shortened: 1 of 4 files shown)


antch:
@gkalpak As requested I created #36130 for the `firstChange` property issue. I also found another issue which can be seen in the repro -- the output events are not hooked to the DOM event listeners during the component lifecycle events `ngOnInit`, `ngOnChanges`, `ngAfterViewInit`, and `ngAfterContentInit`. Should I create a separate issue for that?

gkalpak:
Thx, @antch! Yes, please create a separate issue for output events not firing (since it is a different issue). You can use the same repro.

packages/elements/src/create-custom-element.ts
```diff
@@ -136,31 +136,58 @@ export function createCustomElement<P>(
...
+          defineInputGettersSetters(inputs, this);
+        } else {
+          // Delete the property from the instance, so that it can go through the getters/setters
+          // set on `NgElementImpl.prototype`.
+          preExistingValues.forEach(([propName]) => delete (this as any)[propName]);
+        }
+
+        // Re-apply pre-existing values through the strategy.
+        preExistingValues.forEach(([propName, value]) => strategy.setInputValue(propName, value));
+      }
```
gkalpak: This new block of logic is what causes the payload size increase (~170B).

andrewseguin (approved):
LGTM

goldens/public-api/elements/elements.d.ts
```diff
@@ -23,7 +23,7 @@ export declare interface NgElementStrategy {
     connect(element: HTMLElement): void;
     disconnect(): void;
     getInputValue(propName: string): any;
-    setInputValue(propName: string, value: string): void;
+    setInputValue(propName: string, value: any): void;
```
IgorMinar, line 26: Is `any` absolutely necessary? I don't think it is, we could use `unknown` instead. right?

example: https://stackblitz.com/edit/setter-with-unknown-param

In general, we should be using `unknown` whenever possible instead of `any` (and we have been slowly introducing it into our code base and public api surface already).

Does that work for you?

IgorMinar (changes requested):

goldens/public-api/elements/elements.d.ts
gkalpak: I changed to `any` to match the `getInputValue()` signature (which returns `any`).
Happy to go with `unknown` in both, although that has the slight potential of breaking people (in the unlikely case that they are implementing their own `NgElementStrategy`).

I added a fixup commit. PTAL

goldens/public-api/elements/elements.d.ts
```diff
@@ -22,8 +22,8 @@ export declare interface NgElementStrategy {
     events: Observable<NgElementStrategyEvent>;
     connect(element: HTMLElement): void;
     disconnect(): void;
-    getInputValue(propName: string): any;
-    setInputValue(propName: string, value: string): void;
+    getInputValue(propName: string): unknown;
```
IgorMinar, line 25: This is a breaking change with possibly a significant impact. Do we have some evidence that we could get away with this in google3 as well as externally?

We could intentionally make this change, as it's likely a small impact, and use it as a pilot for fixing other `any`s in our public API surface.

To a developer updating to v10, this will look like TypeScript compiler getting better at type checking, which might be a good justification for making this breaking change (in addition to the expected impact being minimal since this is not a frequently used API today compared to other APIs in Angular).

In any case, this should be documented as a breaking change.

IgorMinar (approved):
LGTM for API changes. I'm a bit worried about the change to the getter, but I'm ok with this being a guinea pig change which could be rolled back if the impact is too severe. Before we proceed lets verify that this doesn't break google3 and do a cursory research on github via code search. Thanks!

Reviewed-for: public-api

gkalpak:
At this point, I feel like the public API fix is not worth the time invested. Therefore, I am going to remove the public API changes and create an issue (in case we want to deal with it in the future).

---
For future reference:
I have search GitHub for custom implementation of `NgElementStrategy`. I have found at least three (that also implement `get/getInputValue()`):
- [spaceman27/angular9-d3: ng-zone-ng-element-strategy-factory.ts](https://github.com/spaceman27/angular9-d3/blob/c81b645a288f079be9326408e0e0be5ba2df3edb/libs/shared/utils/src/lib/elements-utils/ng-zone-ng-element-strategy-factory.ts)
- [tnicola/my-custom-element: elements-modified-zone-strategy.ts](https://github.com/tnicola/my-custom-element/blob/37ff1666579751a16589b9c92295401ca46ca73c/src/app/elements-modified-zone-strategy.ts)
- [jKelio/angular-element-example: custom-element-strategy.ts](https://github.com/jKelio/angular-element-example/blob/0cb8aba4ef0cea739120b09029f4224745223853/src/app/custom-strategy/custom-element-strategy.ts)

While I don't think these would break by the API changes, I can't be 100% sure. More importantly, the thing that matters more is where people might be using `get/setInputValue()`. This is quite hard to search for on GitHub (unless I miss something), since these terms are generic and yield too many results (unrelated to Angular Elements).

For reference, a earch for ["getInputValue("](https://github.com/search?q=%22getInputValue%28%22&type=Code) yields more than 300.000 results and for ["setInputValue("](https://github.com/search?q=%22setInputValue%28%22&type=Code) even more than that :grin:
Even trying to narrow them down by adding "angular" to the search, [I got 500+ results](https://github.com/search?q=%22setInputValue%28%22+angular&type=Code) and of course that does not guarantee that we are not missing valid results 🤷

packages/elements/src/create-custom-element.ts
```diff
@@ -182,20 +209,32 @@ export function createCustomElement<P>(
...
-  // contain property inputs, use all inputs by default.
-  inputs.map(({propName}) => propName).forEach(property => {
-    Object.defineProperty(NgElementImpl.prototype, property, {
-      get: function() {
-        return this.ngElementStrategy.getInputValue(property);
+  // TypeScript 3.9+ defines getters/setters as configurable but non-enumerable properties (in
+  // compliance with the spec). This breaks emulated inheritance in ES5 on environments that do not
+  // natively support `Object.setPrototypeOf()` (such as IE 9-10).
+  // Update the property descriptor of `NgElementImpl#ngElementStrategy` to make it enumerable.
+  Object.defineProperty(NgElementImpl.prototype, 'ngElementStrategy', {enumerable: true});
```
gkalpak, line 216: This is an example of the error that this addition fixes: https://circleci.com/gh/angular/angular/714093

gkalpak:
I have added two new commits at the end (currently f67b1b28126d9ae913083b4998b929114d73e6b3 and 1f32e73d0f094352c82764c9d002097dcad95153) to fix an issue with TS 3.9 and an issue with certain polyfills that do not call the constructor (and was affecting g3 - need new presubmit to confirm).

@andrewseguin, ptal.

AndrewKushnir:
[New Presubmit](http://test/OCL:312122416:BASE:312122454:1589827421784:758287a2).

packages/elements/src/create-custom-element.ts
```diff
@@ -136,31 +136,57 @@ export function createCustomElement<P>(
...
+        const strategy = this._ngElementStrategy =
+            strategyFactory.create(this.injector || config.injector);
+
+        // Collect pre-existing values on the element to re-apply through the strategy.
+        const preExistingValues =
+            inputs.filter(({propName}) => this.hasOwnProperty(propName)).map(({propName}): [
+              string, any
+            ] => [propName, (this as any)[propName]]);
+
+        // In some browsers (e.g. IE10), `Object.setPrototypeOf()` (which is required by some Custom
```
AndrewKushnir, line 157: @gkalpak just want to check if IE 9/10 are affected only or this logic is needed for some other browsers as well? It looks like `Object.setPrototypeOf` is natively supported in modern browsers and given the fact that IE 9/10 will deprecated since v10, I'm wondering if this logic is needed?

gkalpak: AFAICT, this isn;t needed for other browsers. However, this is going into v9 as well and also with IE<11 being deprecated doesn't mean we should break them :wink:
(Also, our current tests are failing without this fix; see https://github.com/angular/angular/pull/36114#discussion_r426700119.)

AndrewKushnir: Sounds good, thanks @gkalpak 👍 
May be we can add a reference to [FW-2141](https://angular-team.atlassian.net/browse/FW-2141) (internal ticket to keep track of IE-specific code that we can refactor/cleanup later) as an inline comment?

AndrewKushnir:
FYI, Blueprint presubmit is successful. Started a [Global presubmit](https://test.corp.google.com/ui#id=OCL:312122416:BASE:312198700:1589853835175:fd069e3e) as well.

packages/elements/src/create-custom-element.ts
gkalpak: :+1: Added a [comment][1] on FW-2141.

[1]: https://angular-team.atlassian.net/browse/FW-2141?focusedCommentId=14257

AndrewKushnir:
Global presubmit was successful as well 👍 (I've updated g3 status for this PR)

packages/elements/src/create-custom-element.ts
AndrewKushnir: Thanks @gkalpak!

andrewseguin (approved):
LGTM

kara:
@gkalpak Can you rebase this? It's not applying cleanly to master for some reason. Also can you clarify if it should be in patch? Not sure what the resolution was for the last commit.

gkalpak:
I have removed the public API commit, so this is good for master _and_ patch. (I have updated the original comment.) I have also rebased and resolved conflicts.

PR #36117, opened 2020-03-17
feat(dev-infra): create commit-message validation script/tooling

josephperrott:
Addresses DEV-43

### DIFF
--- a/dev-infra/BUILD.bazel
+++ b/dev-infra/BUILD.bazel
@@ -8,7 +8,9 @@ ts_library(
     ],
     module_name = "@angular/dev-infra-private",
     deps = [
+        "//dev-infra/commit-message",
         "//dev-infra/pullapprove",
+        "//dev-infra/utils:config",
         "@npm//@types/node",
     ],
 )
@@ -33,6 +35,7 @@ pkg_npm(
     deps = [
         ":cli",
         ":package-json",
+        "//dev-infra/commit-message",
         "//dev-infra/ts-circular-dependencies",
     ],
 )
--- a/dev-infra/cli.ts
+++ b/dev-infra/cli.ts
@@ -20,9 +20,7 @@ switch (args[0]) {
   case 'pullapprove:verify':
     verify();
     break;
-  case 'commit-message:validate':
-    // The expectation is this command will be used for automatic commit
-    // message validation during git commit process.
+  case 'commit-message:pre-commit-validate':
     const commitMessage = readFileSync(join(getRepoBaseDir(), '.git/COMMIT_EDITMSG'), 'utf8');
     if (validateCommitMessage(commitMessage)) {
       console.info('√  Valid commit message');
--- a/dev-infra/cli.ts
+++ b/dev-infra/cli.ts
@@ -6,7 +6,11 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
+import {readFileSync} from 'fs';
+import {join} from 'path';
 import {verify} from './pullapprove/verify';
+import {validateCommitMessage} from './commit-message/validate';
+import {getRepoBaseDir} from './utils/config';
 
 const args = process.argv.slice(2);
 
@@ -15,6 +19,14 @@ const args = process.argv.slice(2);
 switch (args[0]) {
   case 'pullapprove:verify':
     verify();
+    break;
+  case 'commit-message:validate':
+    // The expectation is this command will be used for automatic commit
+    // message validation during git commit process.
+    const commitMessage = readFileSync(join(getRepoBaseDir(), '.git/COMMIT_EDITMSG'), 'utf8');
+    if (validateCommitMessage(commitMessage)) {
+      console.info('√  Valid commit message');
+    }
     break;
   default:
     console.info('No commands were matched');
--- a/dev-infra/commit-message/validate.ts
+++ b/dev-infra/commit-message/validate.ts
@@ -1,0 +1,131 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import {getAngularDevConfig} from '../utils/config';
+import {CommitMessageConfig} from './config';
+
+const FIXUP_PREFIX_RE = /^fixup! /i;
+const SQUASH_PREFIX_RE = /^squash! /i;
+const REVERT_PREFIX_RE = /^revert:? /i;
+const TYPE_SCOPE_RE = /^(\w+)(?:\(([^)]+)\))?\:\s(.+)$/;
+const COMMIT_HEADER_RE = /^(.*)/i;
+const COMMIT_BODY_RE = /^.*\n\n(.*)/i;
+
+/** Parse a full commit message into its composite parts. */
+export function parseCommitMessage(commitMsg: string) {
+  let header = '';
+  let body = '';
+  let type = '';
+  let scope = '';
+  let title = '';
+
+  if (COMMIT_HEADER_RE.test(commitMsg)) {
+    header = COMMIT_HEADER_RE.exec(commitMsg) ![1]
+                 .replace(FIXUP_PREFIX_RE, '')
+                 .replace(SQUASH_PREFIX_RE, '');
+  }
+  if (COMMIT_BODY_RE.test(commitMsg)) {
+    body = COMMIT_BODY_RE.exec(commitMsg) ![1];
+  }
+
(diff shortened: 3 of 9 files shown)


dev-infra/utils/config.ts
```diff
@@ -42,4 +41,6 @@ export function getAngularDevConfig(): DevInfraConfig {
 }
 
 // Interface exressing the expected structure of the DevInfraConfig.
-export interface DevInfraConfig {}
\ No newline at end of file
+export interface DevInfraConfig {
+  commitMessage: {maxLength: number; minBodyLength: number; types: string[]; scopes: string[];};
```
devversion: Optional idea: Declare another interface that is scoped to `dev-infra/commit-message` and use it here. That will make the config more readable in the future, and changing the config can be done while working in `commit-message`.

dev-infra/BUILD.bazel
```diff
@@ -33,6 +35,14 @@ pkg_npm(
...
         "//dev-infra/ts-circular-dependencies",
     ],
 )
+
+ts_config(
+    name = "tsconfig",
+    src = "tsconfig.json",
+    visibility = ["//dev-infra:__subpackages__"],
+    deps = [],
+)
```
devversion: What's the reason for us needing a custom `tsconfig` here now?

dev-infra/cli.ts
```diff
@@ -16,6 +20,12 @@ switch (args[0]) {
   case 'pullapprove:verify':
     verify();
     break;
+  case 'commit-message:validate':
+    const commitMessage = readFileSync(join(getRepoBaseDir(), '.git/COMMIT_EDITMSG'), 'utf8');
+    if (validateCommitMessage(commitMessage)) {
```
devversion: How is that intended to be used? It seems to be different to how the commit message validation is used right now. i.e. now it reads from `.git/COMMIT_EDITMSG` and also doesn't provide the right `priorNonFixupCommits`.

dev-infra/commit-message/validate.ts
```diff
@@ -0,0 +1,128 @@
...
+  let header = '';
+  let body = '';
+  let type = '';
+  let scope = '';
+  let title = '';
+  let headerLine = '';
+
+  if (COMMIT_HEADER_RE.test(commitMsg)) {
+    headerLine = COMMIT_HEADER_RE.exec(commitMsg) ![1];
+    header = headerLine.replace(FIXUP_PREFIX_RE, '').replace(SQUASH_PREFIX_RE, '');
```
devversion: Would it be better to split the message into lines and use the first one as header? It seems to be unnecessary to use a Regex for this. The lines could then be used for the body too (avoiding another regex). Additionally, lines could be passed back so that the max length per body line check doesn't need to split again.

dev-infra/commit-message/validate.ts
```diff
@@ -0,0 +1,128 @@
...
+    body = COMMIT_BODY_RE.exec(commitMsg) ![1];
+  }
+
+  if (TYPE_SCOPE_RE.test(header)) {
+    const parsedCommitHeader = TYPE_SCOPE_RE.exec(header) !;
+    type = parsedCommitHeader[1];
+    scope = parsedCommitHeader[2];
+    title = parsedCommitHeader[3];
+  }
+  const output = {
```
devversion: Object can be returned directly.

dev-infra/BUILD.bazel
josephperrott: The custom tsconfig  is used by the `//dev-infra/commit-message:validate-test` target.  By default, the `tsconfig` that `ts_library` will select in our repo will bring in unwanted and unnecessary angular dependencies.

That is my understanding of it, @gregmagolan may have better insight.

dev-infra/cli.ts
josephperrott: This mimics the current action we do for our `commitmsg` npm script.

For the usages where we want to provide `priorNonFixupCommits`, we can integrate the function via importing from `@angular/dev-infra-private/commit-message` into what we already have, or create another command.

That is my idea at the moment at least, though I am not 100% convinced by my own idea

dev-infra/commit-message/validate.ts
josephperrott: Yep, this was from some testing I was doing, there used to be a `debugger` on line 47 😄

dev-infra/commit-message/validate.ts
```diff
@@ -0,0 +1,128 @@
...
+        `INVALID COMMIT MSG: \n` +
+        `${Array(40).fill('─').join('')}\n` +
+        `${commitMsg}\n` +
+        `${Array(40).fill('─').join('')}\n` +
+        `ERROR: \n` +
+        `  ${errorMessage}` +
+        `\n\n` +
+        `The expected format for a commit is: \n` +
+        `<type>(<scope>): <subject>\n\n<body>`);
+  }
```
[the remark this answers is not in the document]
josephperrott: My reasoning for having it inside the `validateCommitMessage` as a closure was to prevent needing to pass the `commitMsg` value in at each call.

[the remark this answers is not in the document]
devversion: That's fair.

dev-infra/BUILD.bazel
devversion: I don't think that the `tsconfig` brings in unnecessary dependencies. It basically helps ensuring that we have consistent settings in the repo: https://github.com/angular/angular/blob/master/packages/tsconfig-build.json.

We also rely on other stuff outside of `dev-infra/`. e.g. the `tools/testing` stuff. I don't think there is anything wrong with sharing stuff here.

dev-infra/cli.ts
devversion: Importing the function sounds good to me. I just checked, it looks like the ´commitmsg` Yarn script is picked up by Husky (though this has changed in more recent versions of `husky`).

Can we document this somewhere? e.g. in the cli switch case? e.g. that this is usually run by Husky for pre-commit validation?

josephperrott: I think the fact that `commitmsg` being deprecated for husky is orthogonal to this change.

I'm happy to document that it will be used there, but I am unsure if there is a ton of value in documenting where the commands are used only in angular/angular.

devversion: Yeah, I just mentioned that since it's relevant to why the `commitmsg` npm script exists (as you mentioned that as the use-case for this CLI command).


I don't mind. I think that it would be nice to mention such things, as otherwise it's not clear why CLI commands for given things exist / whether they are actually used somewhere. Up to you 😄

devversion (approved):

dev-infra/cli.ts
josephperrott: I added a general command about when we expect it to be used.

dev-infra/commit-message/validate.ts
josephperrott: I think it might be something worth visiting in a follow up, but I am not quite sure how to best go about it because part of making sure the body is defined is ensuring that we have two newlines between the commit header and the body.

I did move to checking the `commitMsg` string for the fixup, squash and revert regex rather than just the first line. Which allowed me to get rid of the awkward `headerLine` variable that was in place.

devversion: Okay. That works for me.

gkalpak (approved):
Yay for having the script in TS :grin:

One question:
Is the idea to eventually replace [tools/validate-commit-message/validate-commit-message.js][1]? Why not now? What is missing?

[1]: https://github.com/angular/angular/blob/c0143cb2abdd172de1b95fd1d2c4cfc738640e28/tools/validate-commit-message/validate-commit-message.js

dev-infra/commit-message/validate.ts
```diff
@@ -0,0 +1,131 @@
...
+
+/** Validate a commit message against using the local repo's config. */
+export function validateCommitMessage(
+    commitMsg: string, disallowSquash: boolean = false, nonFixupCommitHeaders?: string[]) {
+  function error(errorMessage: string) {
+    console.error(
+        `INVALID COMMIT MSG: \n` +
+        `${Array(40).fill('─').join('')}\n` +
+        `${commitMsg}\n` +
+        `${Array(40).fill('─').join('')}\n` +
```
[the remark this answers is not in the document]
josephperrott: One day I will remember that the `fill` method exists...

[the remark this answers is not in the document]
gkalpak: Except it's called `repeat` 😛

josephperrott:
Yes, the plan is to migrate to relying on this, but we have to wait until we move to depending on the version of the dev-infra-private package that has this included.

AndrewKushnir:
FYI, adding "cleanup" label, since a couple CircleCI jobs is failing.

Issue #31586, opened 2019-07-16
[i18n] Unexpected character "EOF" 


aguadotzn:
# 🐞 bug report

### Affected Package
I can not pin-point any package because I'm not sure about it.

### Is this a regression?
I think was not working in previous versions but i'm not sure 100% about it.

### Description
I'm trying to internationalise an application. I create a `<span>` and put several information there. When we have info that starts with '(' the compilation fails.

## 🔬 Minimal Reproduction
You can find small reproduction of the issue here with the same output:
https://stackblitz.com/edit/angular-issue-repro2-u3gnzz

## 🔥 Exception or Error
```
Template parse errors:
@xxx/core: Unexpected character "EOF" (Do you have an unescaped "{" in your template? Use "{{ '{' }}") to escape it.) (" transport {International transport} (Inter)national transport {(Inter)national transport} } </span> [ERROR ->]"): 67
}
</span>
```

## 🌍  Your Environment

**Angular Version:**
<pre>Angular CLI: 7.0.6
Node: 8.11.1
OS: darwin x64
Angular: 7.0.4
... animations, cdk, common, compiler, compiler-cli, core, forms
... http, platform-browser, platform-browser-dynamic, router

Package                           Version
-----------------------------------------------------------
@angular-devkit/architect         0.10.6
@angular-devkit/build-angular     0.10.6
@angular-devkit/build-optimizer   0.10.6
@angular-devkit/build-webpack     0.10.6
@angular-devkit/core              7.0.6
@angular-devkit/schematics        7.0.6
@angular/cli                      7.0.6
@ngtools/webpack                  7.0.6
@schematics/angular               7.0.6
@schematics/update                0.10.6
rxjs                              6.3.3
typescript                        3.1.6
webpack                           4.19.1
<code>

</code></pre>

**Anything else relevant?**
IDE: IntelliJ IDEA
OS: macOS

aguadotzn:
UPDATE: Same problems if the sentence stars with **%**. 

So maybe happen with more characters...

petebacondarwin:
I can see the reproduction on stackblitz but equally I can't find a version of Angular for which this ever worked.

sonukapoor:
@petebacondarwin As discussed offline, I will work on this.

PR #36123, opened 2020-03-18
fix(localize): allow ICU expansion to start with any character except…

sonukapoor:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #31586

- [ ] Yes
- [x] No

### DIFF
--- a/packages/compiler/src/ml_parser/lexer.ts
+++ b/packages/compiler/src/ml_parser/lexer.ts
@@ -747,7 +747,7 @@ function isNamedEntityEnd(code: number): boolean {
 }
 
 function isExpansionCaseStart(peek: number): boolean {
-  return peek === chars.$EQ || chars.isAsciiLetter(peek) || chars.isDigit(peek);
+  return peek === chars.$EQ || peek !== chars.$RBRACE;
 }
 
 function compareCharCodeCaseInsensitive(code1: number, code2: number): boolean {
--- a/packages/compiler/src/ml_parser/lexer.ts
+++ b/packages/compiler/src/ml_parser/lexer.ts
@@ -747,7 +747,7 @@ function isNamedEntityEnd(code: number): boolean {
 }
 
 function isExpansionCaseStart(peek: number): boolean {
-  return peek === chars.$EQ || peek !== chars.$RBRACE;
+  return peek !== chars.$RBRACE;
 }
 
 function compareCharCodeCaseInsensitive(code1: number, code2: number): boolean {
--- a/packages/compiler/test/ml_parser/html_parser_spec.ts
+++ b/packages/compiler/test/ml_parser/html_parser_spec.ts
@@ -313,6 +313,26 @@ import {humanizeDom, humanizeDomSourceSpans, humanizeLineColumn} from './ast_spe
           expect(p.errors.length).toEqual(0);
         });
 
+        it(`should support ICU expressions with cases that contain any character except '}'`,
+           () => {
+             const p = parser.parse(
+                 `{a, select, b {foo} % bar {% bar}}`, 'TestComp', {tokenizeExpansionForms: true});
+             expect(p.errors.length).toEqual(0);
+           });
+
+        it('should error when expansion case is not properly closed', () => {
+          const p = parser.parse(
+              `{a, select, b {foo} % { bar {% bar}}`, 'TestComp', {tokenizeExpansionForms: true});
+          expect(humanizeErrors(p.errors)).toEqual([
+            [
+              6,
+              'Unexpected character "EOF" (Do you have an unescaped "{" in your template? Use "{{ \'{\' }}") to escape it.)',
+              '0:36'
+            ],
+            [null, 'Invalid ICU message. Missing \'}\'.', '0:22']
+          ]);
+        });
+
         it('should error when expansion case is not closed', () => {
           const p = parser.parse(
               `{messages.length, plural, =0 {one`, 'TestComp', {tokenizeExpansionForms: true});


packages/compiler/src/ml_parser/lexer.ts
```diff
@@ -747,7 +747,7 @@ function isNamedEntityEnd(code: number): boolean {
 }
 
 function isExpansionCaseStart(peek: number): boolean {
-  return peek === chars.$EQ || chars.isAsciiLetter(peek) || chars.isDigit(peek);
+  return peek === chars.$EQ || peek !== chars.$RBRACE;
```
petebacondarwin:
```suggestion
  return peek !== chars.$RBRACE;
```

petebacondarwin (changes requested):
No need for the additional check now.

In the commit message, please change references to `expansion` to `expansion case` since this is what the commit is fixing.

packages/compiler/src/ml_parser/lexer.ts
sonukapoor: Yes, of course...

petebacondarwin (approved):
Great! Thanks

AndrewKushnir:
Hi @sonukapoor, sorry I didn't have a chance to look at the changes in this PR today. Even though the diff is relatively small, I'd like to understand the code around it and see if there are any corner cases and why the function was not implemented like this before. I've started [g3 presubmit](http://test/OCL:301691009:BASE:301691112:1584573408494:76d9f45a) (internal only link) for this change as well to see if there are any cases in g3 that might be affected. Thank you.

sonukapoor:
Hi @AndrewKushnir, no problem at all:

I have discussed this offline with @petebacondarwin to make sure that we are safe to change this.

The template in question is:

```
<span i18n>{
  gOptionLabel, select,
  National transport {National transport}
  % by which guarantee amount is to be reduced {% by which guarantee amount is to be reduced}
  }
</span>
```

Once the expansion has started, we continue until we find the closing character `}`. In between that we only allow for alphanumeric characters. However, in the above example, we can see that `%` is being used instead, other characters could be used as well.  

To allow the usage of any character, except for the closing right bracket `}`, we have changed the condition. This resolves the issue and the current existing tests are still passing as well. Once, we hit the end character `}`, we close the expansion and can move on. This logic can be seen in this function (added comment blocks below only to make it easier to follow);

```
private _tokenizeExpansionForm(): boolean {
  // Expansion starts here
  if (this.isExpansionFormStart()) {
    this._consumeExpansionFormStart();
    return true;
  }

  // find the next char, and allow anything except `}`
  if (isExpansionCaseStart(this._cursor.peek()) && this._isInExpansionForm()) {
    this._consumeExpansionCaseStart();
    return true;
  }

  // Close the expansion as the current char is `}`
  if (this._cursor.peek() === chars.$RBRACE) {
    if (this._isInExpansionCase()) {
      this._consumeExpansionCaseEnd();
      return true;
    }

    if (this._isInExpansionForm()) {
      this._consumeExpansionFormEnd();
      return true;
    }
  }
  return false;
}
```

[Here](https://github.com/angular/angular/commit/a3a54299afd41cc2049953c52ff869ef9c36fc37)is a previous commit from Olivier that added support for numerical digits in expansion cases. 

@petebacondarwin please add anything else  I may have missed.

petebacondarwin:
@AndrewKushnir - yes @sonukapoor and I spent a lot of time discussing this offline and working through the possibilities. The [ICU spec](http://userguide.icu-project.org/formatparse/messages) is pretty vague:

> Use a "select" argument to select sub-messages via a fixed set of keywords.

It does not specify what can be a "keyword" but from looking at the surrounding syntax it appears that it can indeed be any string that does not contain a `}` character.

AndrewKushnir (approved):
Thanks for additional information @sonukapoor and @petebacondarwin, the change LGTM 👍 

The g3 presubmit is successful, I will start a global run tonight as well to test all affected targets in g3 just to make sure everything is ok.

Thank you.

sonukapoor:
Thanks 🙏

AndrewKushnir:
[Link to Global TAP run](https://test.corp.google.com/ui#id=OCL:301691009:BASE:301943198:1584670359389:bfb2773f) (internal only).

AndrewKushnir:
@sonukapoor, @petebacondarwin global presubmit looks good 👍

I have a couple more questions/notes:

- can we extend commit message to include info that Pete mentioned in https://github.com/angular/angular/pull/36123#issuecomment-601074010, that'd help to provide more context for future references

- previously whitespace symbols were not treated as a possible ICU case start. With this change whitespace symbols can technically be treated as ICU case start. Can we check if tis change doesn't affect i18n extraction mechanism and that i18n message IDs remain the same for ICUs after this change (otherwise it would trigger re-translation for all messages that contain ICUs)? Can we add a test (or find existing ones) to verify that?

Sorry for additional questions, just want to make sure we are not missing anything.

Thank you.

petebacondarwin:
Hi @AndrewKushnir 
I believe that we covered with regards to whitespace by this line: https://github.com/angular/angular/blob/f78876695dd46f28b99246536f3503d5e452bdb5/packages/compiler/src/ml_parser/lexer.ts#L624

When we consume an expansion case start we also trim whitespace off the front and back.
The only scenario I can think of would be to incorrectly identify a block of "only whitespace" as an expansion case. But this is actually covered by this line: https://github.com/angular/angular/blob/f78876695dd46f28b99246536f3503d5e452bdb5/packages/compiler/src/ml_parser/lexer.ts#L619 and this line https://github.com/angular/angular/blob/f78876695dd46f28b99246536f3503d5e452bdb5/packages/compiler/src/ml_parser/lexer.ts#L640

where consuming an "expansion form start" and an "expansion case" always end by skipping over whitespace.

sonukapoor:
@AndrewKushnir I have updated the commit message.

AndrewKushnir:
Thanks for additional information and updates @sonukapoor and @petebacondarwin, I'm marking this PR as ready for merge. Thank you.

PR #36129, opened 2020-03-18
build: enable service-worker tests on saucelabs

crisbeto:
Enables the `service-worker` tests on Saucelabs and fixes some issues that were preventing them from running on IE. The issues were:
1. We were serving es2017 code during tests. I've set it to es5.
2. The check which was verifying whether the environment is supported ended up hitting a `require` call in the browser which caused it to fail on browsers that don't support the `URL` API.


packages/service-worker/worker/tsconfig.json
```diff
@@ -18,6 +18,7 @@
     "types": []
   },
   "bazelOptions": {
+    "devmodeTargetOverride": "es5",
```
gkalpak: Any idea why is this needed? We don't run the `service-worker/worker/test/` tests in the browser afaict.

crisbeto: I think some tests end up being run in the browser and in the ones where all the APIs aren't supported, we still serve the test files and skip the tests. This is necessary so those older browsers can still load the files.

gkalpak: 🤷‍♂

crisbeto: For what it's worth, we seem to be doing this kind of override in other targets as well. It's possible that this one was overlooked when it was added to the others.

devversion: For context, `ts_library` by default sets the devmode target to `es2015`. [See code snippet](https://github.com/bazelbuild/rules_nodejs/blob/master/packages/typescript/src/internal/build_defs.bzl#L222-L224).

So setting the `devmodeTargetOverride` makes sense if the `worker/` folder is intended to be used in browsers. I'm actually wondering why we need the custom tsconfig at all? @crisbeto did you try removing the tsconfig in favor of the default one that sets our `devmodeTargetOverride` (as done in all other packages). The tsconfig looks like a leftover from pre-Bazel.

crisbeto: I'm guessing this tsconfig exists, because it's targeting es2017. I'm not familiar with why it's set up this way though.

devversion: The `target` property is not respected at all in Bazel. That's why I'm wondering. It's most likely a leftover from the `build.sh` script.

crisbeto: Thanks for the help, dropping the tsconfig also works.

devversion (approved):
LGTM for dev-infra.

alxhub (approved):

AndrewKushnir:
@crisbeto, it looks like presubmit for this PR is not required (ngbot indicates that this change doesn't affect g3)?

crisbeto:
I wanted to be safe since I changed something in `service-worker/testing` and I wasn't aware whether it was being used in g3.

AndrewKushnir:
@crisbeto ok, got it, thanks. I just confirmed that the `packages/service-worker/**` pattern is excluded in g3 sync script, so the changes in files that match the pattern don't affect g3.

crisbeto:
Thank you for checking. Removing the presubmit label.

kara (approved):
LGTM (though I'm not sure why I was auto-selected as a reviewer on this one)

PR #36145, opened 2020-03-19
build(docs-infra): update @angular/{cli,core,material} to latest versions and TS to 3.8

gkalpak:
☝️ 

Jira issue: [FW-1970](https://angular-team.atlassian.net/browse/FW-1970)

##
Blocking issues:

- [ ] Confirm that the payload size increases are acceptable.
- [x] ~~This seems to be very prone to [FW-2008][1].
    We should address FW-2008 first, before merging this.~~

[1]: https://angular-team.atlassian.net/browse/FW-2008

[FW-2008]: https://angular-team.atlassian.net/browse/FW-2008


aio/scripts/_payload-limits.json
```diff
@@ -3,7 +3,7 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 2987,
-        "main-es2015": 450612,
```
filipesilva: So the size begins at 450676. Updating CLI bumps it to 450677, then updating FW bumps it to 451655, then updating material bumps it to 451935.

filipesilva (approved):

aio/scripts/_payload-limits.json
gkalpak: Exactly! Thx for summarizing the changes :+1:

goldens/size-tracking/aio-payloads.json
```diff
@@ -3,26 +3,26 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 2987,
-        "main-es2015": 450612,
-        "polyfills-es2015": 52195
+        "main-es2015": 454032,
```
kara, line 6: Oof, looks like this got worse since the last time we updated this PR...

kara (approved):
Not happy about the size regression, but given that this PR doesn't cause it and this is blocking other things, LGTM. 

I'll create a Jira to investigate the regression separately.

petebacondarwin (approved):
But I don't understand why the examples are left at 9.1.x while the rest of AIO is updated to 10.0.0-next.x

gkalpak:
> But I don't understand why the examples are left at 9.1.x while the rest of AIO is updated to 10.0.0-next.x

Examples are what people will use to play with (either on StackBlitz or locally), so we generally stick to stable versions for examples. The angular.io app itself can be on any version we are happy with (and currently we need 10-next, because it is the only one that contains #36626 until the next 9.x release).

PR #36149, opened 2020-03-19
fix(animations): allow computeStyle to work on elements created in Node

matsko:
This patch is a follow-up patch to 35c9f0dc2f3665d4f9d9ece328cee4559bbec9c6.
It changes the `computeStyle` function to handle situations where
non string based values are returned from `window.getComputedStyle`.
This situation usually ocurrs in Node-based test environments where
the element or `window.getComputedStyle` is mocked out.

### DIFF
--- a/packages/animations/browser/src/render/shared.ts
+++ b/packages/animations/browser/src/render/shared.ts
@@ -237,3 +237,53 @@ export function hypenatePropsObject(object: {[key: string]: any}): {[key: string
   });
   return newObj;
 }
+
+/**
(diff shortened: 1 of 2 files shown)


matsko:
resurrected version of https://github.com/angular/angular/commit/35c9f0dc2f3665d4f9d9ece328cee4559bbec9c6

matsko:
presubmit = https://test.corp.google.com/ui#id=OCL:301885626:BASE:301885669:1584650281567:8cf68c5a

PR #36157, opened 2020-03-20
fix(forms): handle numeric values properly in the validator 

### LINKED ISSUE #35591: minLength(1) validator returns invalid on a valid input
# 🐞 bug report

### Affected Package

 The issue is caused by package @angular/forms

### Description

 Validators.minLength(1) on a form field that has number as type returns invalid when the number entered is 0.

## 🔬 Minimal Reproduction

 https://stackblitz.com/edit/angular-gjs5kh

## 🌍  Your Environment

**Angular Version:**
<pre><code>

8.2.14

</code></pre>

sonukapoor:
Previously, the behaviour of the `minLength` and `maxLength` validator
caused confusion, and making it appear as they work with numeric values.
This commit fixes this, by always passing the validation when a numeric value
is used.

BREAKING CHANGES:

* The `minLength` and `maxLength` validators now verify that a value has
numeric `length` property and invoke validation only if that's the case.
Previously, falsey values without the length property (such as `0` or
`false` values) were triggering validation errors. If your code relies on
the old behavior, you can include other validators such as [min][1] or
[requiredTrue][2] to the list of validators for a particular field.

[1]: https://angular.io/api/forms/Validators#min
[2]: https://angular.io/api/forms/Validators#requiredTrue

Closes #35591

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #35591

- [ ] Yes
- [x] No

### DIFF
--- a/packages/forms/src/validators.ts
+++ b/packages/forms/src/validators.ts
@@ -299,7 +299,7 @@ export class Validators {
    *
    */
   static minLength(minLength: number): ValidatorFn {
-    return (control: AbstractControl): ValidationErrors | null => {
+    return (control: AbstractControl): ValidationErrors|null => {
       if (isEmptyInputValue(control.value) || !hasValidLength(control.value)) {
         // don't validate empty values to allow optional controls
         // don't validate values without `length` property
@@ -340,7 +340,7 @@ export class Validators {
    *
    */
   static maxLength(maxLength: number): ValidatorFn {
-    return (control: AbstractControl): ValidationErrors | null => {
+    return (control: AbstractControl): ValidationErrors|null => {
       return hasValidLength(control.value) && control.value.length > maxLength ?
           {'maxlength': {'requiredLength': maxLength, 'actualLength': control.value.length}} :
           null;
--- a/packages/forms/src/validators.ts
+++ b/packages/forms/src/validators.ts
@@ -11,11 +11,16 @@ import {forkJoin, from, Observable} from 'rxjs';
 import {map} from 'rxjs/operators';
 
 import {AsyncValidatorFn, ValidationErrors, Validator, ValidatorFn} from './directives/validators';
-import {AbstractControl, FormControl} from './model';
+import {AbstractControl} from './model';
 
 function isEmptyInputValue(value: any): boolean {
   // we don't check for string here so it also works with arrays
   return value == null || value.length === 0;
+}
+
+function hasValidLength(value: any): boolean {
+  // non-strict comparison is intentional, to check for both `null` and `undefined` values
+  return value != null && typeof value.length === 'number';
 }
 
 /**
@@ -294,13 +299,15 @@ export class Validators {
    *
    */
   static minLength(minLength: number): ValidatorFn {
-    return (control: AbstractControl): ValidationErrors|null => {
-      if (isEmptyInputValue(control.value)) {
-        return null;  // don't validate empty values to allow optional controls
+    return (control: AbstractControl): ValidationErrors | null => {
+      if (isEmptyInputValue(control.value) || !hasValidLength(control.value)) {
+        // don't validate empty values to allow optional controls
+        // don't validate values without `length` property
+        return null;
       }
-      const length: number = control.value ? control.value.length : 0;
-      return length < minLength ?
-          {'minlength': {'requiredLength': minLength, 'actualLength': length}} :
+
+      return control.value.length < minLength ?
+          {'minlength': {'requiredLength': minLength, 'actualLength': control.value.length}} :
           null;
     };
   }
@@ -333,10 +340,9 @@ export class Validators {
    *
    */
   static maxLength(maxLength: number): ValidatorFn {
-    return (control: AbstractControl): ValidationErrors|null => {
-      const length: number = control.value ? control.value.length : 0;
-      return length > maxLength ?
-          {'maxlength': {'requiredLength': maxLength, 'actualLength': length}} :
+    return (control: AbstractControl): ValidationErrors | null => {
+      return hasValidLength(control.value) && control.value.length > maxLength ?
+          {'maxlength': {'requiredLength': maxLength, 'actualLength': control.value.length}} :
           null;
     };
   }
--- a/packages/forms/test/validators_spec.ts
+++ b/packages/forms/test/validators_spec.ts
@@ -232,6 +232,16 @@ describe('Validators', () => {
       expect(Validators.minLength(1)(new FormControl(-1))).toBeNull();
       expect(Validators.minLength(1)(new FormControl(+1))).toBeNull();
     });
+
+    it('should return null when passing an object that contains a length property', () => {
+      const value = {length: 5, someValue: [1, 2, 3, 4, 5]};
+      expect(Validators.minLength(1)(new FormControl(value))).toBeNull();
+    });
+
+    it('should return null when passing a boolean', () => {
+      expect(Validators.minLength(1)(new FormControl(true))).toBeNull();
+      expect(Validators.minLength(1)(new FormControl(false))).toBeNull();
+    });
   });
 
   describe('maxLength', () => {
@@ -257,11 +267,6 @@ describe('Validators', () => {
       });
     });
 
-    it('should return null when passing an object that contains a length property', () => {
-      const value = {length: 5, someValue: [1, 2, 3, 4, 5]};
-      expect(Validators.minLength(1)(new FormControl(value))).toBeNull();
-    });
-
     it('should not error when FormArray has valid length', () => {
       const fa = new FormArray([new FormControl(''), new FormControl('')]);
       expect(Validators.maxLength(2)(fa)).toBeNull();
@@ -284,6 +289,11 @@ describe('Validators', () => {
     it('should return null when passing an object that contains a length property', () => {
       const value = {length: 1, someValue: [1, 2, 3, 4, 5]};
       expect(Validators.maxLength(1)(new FormControl(value))).toBeNull();
+    });
+
+    it('should return null when passing a boolean', () => {
+      expect(Validators.maxLength(1)(new FormControl(true))).toBeNull();
+      expect(Validators.maxLength(1)(new FormControl(false))).toBeNull();
     });
   });
 
--- a/packages/forms/test/validators_spec.ts
+++ b/packages/forms/test/validators_spec.ts
@@ -233,9 +233,12 @@ describe('Validators', () => {
       expect(Validators.minLength(1)(new FormControl(+1))).toBeNull();
     });
 
-    it('should return null when passing an object that contains a length property', () => {
+    it('should trigger validation for an object that contains numeric length property', () => {
       const value = {length: 5, someValue: [1, 2, 3, 4, 5]};
       expect(Validators.minLength(1)(new FormControl(value))).toBeNull();
+      expect(Validators.minLength(10)(new FormControl(value))).toEqual({
+        'minlength': {'requiredLength': 10, 'actualLength': 5}
+      });
     });
 
     it('should return null when passing a boolean', () => {
@@ -286,9 +289,12 @@ describe('Validators', () => {
       expect(Validators.maxLength(1)(new FormControl(+1))).toBeNull();
     });
 
-    it('should return null when passing an object that contains a length property', () => {
-      const value = {length: 1, someValue: [1, 2, 3, 4, 5]};
-      expect(Validators.maxLength(1)(new FormControl(value))).toBeNull();
+    it('should trigger validation for an object that contains numeric length property', () => {
+      const value = {length: 5, someValue: [1, 2, 3, 4, 5]};
+      expect(Validators.maxLength(10)(new FormControl(value))).toBeNull();
+      expect(Validators.maxLength(1)(new FormControl(value))).toEqual({
+        'maxlength': {'requiredLength': 1, 'actualLength': 5}
+      });
     });
 
     it('should return null when passing a boolean', () => {
--- a/packages/forms/test/validators_spec.ts
+++ b/packages/forms/test/validators_spec.ts
@@ -225,6 +225,13 @@ describe('Validators', () => {
         'minlength': {'requiredLength': 2, 'actualLength': 1}
       });
     });
+
+    it('should always return null with numeric values', () => {
+      expect(Validators.minLength(1)(new FormControl(0))).toBeNull();
+      expect(Validators.minLength(1)(new FormControl(1))).toBeNull();
+      expect(Validators.minLength(1)(new FormControl(-1))).toBeNull();
+      expect(Validators.minLength(1)(new FormControl(+1))).toBeNull();
+    });
   });
 
   describe('maxLength', () => {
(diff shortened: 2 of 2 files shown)


petebacondarwin (approved):
This change is an improvement, generally, since it is always safer (and more performant) to compare against `undefined` and I am happy for it to land.

But we should be aware, and encourage people (perhaps via documentation) not to use `minlength` with `number` fields, precisely for some of the reasons laid out in the comments for the original issue. Use cases where the string length of the number are important seem to me that they should really be considered strings that are restricted to certain patterns, which can be dealt with by the `pattern` validator. For example a telephone number is not really a "number" in the HTML forms sense of the word. It is really a string of digits, which could be validated with something like `Validators.pattern(/\d{9,11}/)` for example.

packages/forms/src/validators.ts
```diff
@@ -289,7 +289,7 @@ export class Validators {
       if (isEmptyInputValue(control.value)) {
         return null;  // don't validate empty values to allow optional controls
       }
-      const length: number = control.value ? control.value.length : 0;
+      const length: number = control.value !== undefined ? control.value.length : 0;
```
AndrewKushnir: @sonukapoor, there is the `isEmptyInputValue` function call above this line that should catch the case where `control.value` is `undefined` and return `null` (so we should not reach this line in case `control.value` is `undefined`), so I'm not sure if this check is correct. Could you please have a look at the actual value type that we have here in your case?

AndrewKushnir (changes requested):

packages/forms/src/validators.ts
sonukapoor: @AndrewKushnir The problem is that we need to return `0` if the value is undefined. The condition above would return `null`, making the field appear as `invalid` when `maxLength` is set to `1`.

AndrewKushnir: It also seems like we should make similar changes in the `maxLength` validator as well (and add tests to verify the change).

sonukapoor: @AndrewKushnir That's not the problem. The issue is that we need to explicitly check for undefined here,  because if `control.value` is set to `0`, it will evaluate to false.

sonukapoor: > It also seems like we should make similar changes in the `maxLength` validator as well (and add tests to verify the change).

Good catch. I will make a new commit.

AndrewKushnir: @sonukapoor, thanks for additional information. I did some debugging and here are some comments:

- the `isEmptyInputValue` specifically contains the non-strict equal (`value == null`) that captures both `null` and `undefined` values and for these values we have an early return in the function (that returns `null`). Based on that, the `control.value` can not be `undefined` in the code that you updated.

- based on the previous point, if looks like your change forces all values to go through the `control.value.length` path (at line 292), since the `undefined` values are filtered out at the beginning of the function.

- for any number other than `0`, the original code at line 292 (i.e. this one: `const length: number = control.value ? control.value.length : 0`) returns `undefined` and after that the comparison on line 292 becomes `false`, thus triggering `null` to be returned from the validator function.

- `0` should not be any different from other numbers (it just happens to be false-y in ternary, thus it triggers a different behavior), so given that this validator returns `null` for any number it should also return `null` for `0`.

- in order to support this case correctly in non-breaking way, I think we may need to update the `return` condition to check if the `length` is not `undefined` (it'd be better to check if `length` is actually a number, but that might be potentially breaking for cases where `length` is a function on a class and it returns some non-number type that triggers `length < minLength` to be true), i.e.:

```
      return length !== undefined && length < minLength ?
          {'minlength': {'requiredLength': minLength, 'actualLength': length}} :
          null;
```

- I think we should add tests for other numbers (positive and negative) other than `0` to check validator behavior and may be add tests for boolean values (the `false` value would trigger the same code path as `0`)

Please let me know what do you think.

Thank you.

AndrewKushnir (changes requested):

packages/forms/src/validators.ts
sonukapoor: @AndrewKushnir Thanks for sharing all of this. I think it makes sense. I will run some tests over the weekend and let you know.

sonukapoor: @AndrewKushnir, All of this makes sense, however the new suggested return statement will not work, because the ternary operator at line 292, will set the `length` variable to 0. Therefore the function will return the object and not null. 

Would it be better if we convert the value to a string before we start the validation and comparison? It seems to work fine.

```typescript
const controlValue = control.value.toString();
const length = controlValue ? controlValue.length : 0;

return length !== undefined && length < minLength ? 
  {'minlength': {'requiredLength': minLength, 'actualLength': length}} : 
  null;
```

Let me know what you think about this.

AndrewKushnir: @sonukapoor you are right about line 292, so the solution I proposed would not work :(

I don't think that calling `toString` would work either, since the min/max length validators also work with arrays and calling `toString` on array will produce a string that would contain a different length, for ex. `[1,2,3].length` is 3, when `[1,2,3].toString().length` is 5. That would also cause a breaking change when previously these validators returned `null` for numbers and after the `toString` change we'll start actually validating numbers and checking the length of strings that represent a number.

May be the easiest and less break-y solution for now might be checking whether `control.value` type is a number and exit early if that's the case, something like this?

```
if (isEmptyInputValue(control.value) || typeof control.value === 'number') {
  return null; 
}
```

sonukapoor: We would need to add an additional check for `0` as we do not want to return null for any number without checking its length. `0` is a special case. 

```typescript
if (isEmptyInputValue(control.value) || (typeof control.value === 'number' && control.value === 0) ) {
  return null; // don't validate empty values to allow optional controls
}
```

sonukapoor: @AndrewKushnir Few more observations below:

The reason, why the array `[1,2,3].toString().length` is returning 5 is because it's counting the commas. We would have to strip them off if we want to use that logic.

I ran some more tests and noticed that if the value is numeric, then `control.value.length` will always be `undefined`. I think we need to implement custom logic for numeric values. 

For the `0` case we should also check for the minLength and exit early:

```typescript
if (isEmptyInputValue(control.value) || (control.value === 0 && minLength === 1)) {
  return null; // don't validate empty values to allow optional controls
}
```

If the value is not `0`, then we still need to check for the length, which can only be done properly after converting the value to a `string`.

```typescript
const controlValue = control.value;
let length;
if (typeof controlValue === 'number') {
  length = controlValue.toString().length;
} else {
  length = controlValue.length;
}
```

Since `control.value` can not be `undefined`, there is no need for the `ternary` operator anymore to check if `control.value` is defined or not. 

What do you think?

AndrewKushnir: Hi @sonukapoor, thanks for additional information.

I think we need to separate:
- **fix** for the `0` value and ideally come up with a common solution for all numeric (and may be other primitive) values, without special-casing any values if possible
- **feature** to support numeric values in min/max length validators (that might require separate discussion/decision)

What is a bit confusing me is this line:
https://github.com/angular/angular/blob/9d415f9c3f12cb2c29298ba414f55ce230201700/packages/forms/src/validators.ts#L292

The thing is that the portion of false-y values (`null`, `undefined` and empty string) except `false` and `0` are filtered out above (and `null` is returned):
https://github.com/angular/angular/blob/9d415f9c3f12cb2c29298ba414f55ce230201700/packages/forms/src/validators.ts#L289

But for both `0` and `false`, we actually want to return `null` as well (i.e. do not trigger error). Looking at the history of this code, it seems like it was updated in https://github.com/angular/angular/commit/e9f307f9488e44879dc027e5f4436fb6bc046fa4#diff-c754958cc98f6c0fe7edca2617efc69bL87 to handle cases other than a string and after that, the `isEmptyInputValue` was introduced in https://github.com/angular/angular/commit/d22eeb70b8eec5ed29606154edb2f8ebe02dbab0#diff-c754958cc98f6c0fe7edca2617efc69bR73, which made the `control.value ? control.value.length : 0` check obsolete for a number of cases.

I'd like to explore whether we can simplify the mental model around the `minLength` and `maxLength` validators a bit and say that these validators are actually supported for values that have the `length` property and this property is a number. Otherwise, there are some comparison in the code where we check `undefined` against a `number` (when comparing lengths) and rely on the comparison results, which is not intuitive.

Here is a diff of a rough implementation of the idea: check whether we have a value and a value has a valid length (`typeof value.length === 'number'`) and trigger validator logic only in this case (otherwise it's a noop and we return `null`).

```
--- a/packages/forms/src/validators.ts
+++ b/packages/forms/src/validators.ts
@@ -17,6 +17,11 @@ function isEmptyInputValue(value: any): boolean {
   return value == null || value.length === 0;
 }
 
+function hasValidLength(value: any): boolean {
+  // non-strict comparison is intentional, to check for both `null` and `undefined` values
+  return value != null && typeof value.length === 'number';
+}
+
 /**
  * @description
  * An `InjectionToken` for registering additional synchronous validators used with `AbstractControl`s.
@@ -286,12 +291,12 @@ export class Validators {
    */
   static minLength(minLength: number): ValidatorFn {
     return (control: AbstractControl): ValidationErrors | null => {
-      if (isEmptyInputValue(control.value)) {
-        return null;  // don't validate empty values to allow optional controls
+      if (isEmptyInputValue(control.value) || !hasValidLength(control.value)) {
+        return null;  // don't validate empty values to allow optional controls and also
+                      // don't validate values without `length` property
       }
-      const length: number = control.value ? control.value.length : 0;
-      return length < minLength ?
-          {'minlength': {'requiredLength': minLength, 'actualLength': length}} :
+      return control.value.length < minLength ?
+          {'minlength': {'requiredLength': minLength, 'actualLength': control.value.length}} :
           null;
     };
   }
@@ -324,9 +329,8 @@ export class Validators {
    */
   static maxLength(maxLength: number): ValidatorFn {
     return (control: AbstractControl): ValidationErrors | null => {
-      const length: number = control.value ? control.value.length : 0;
-      return length > maxLength ?
-          {'maxlength': {'requiredLength': maxLength, 'actualLength': length}} :
+      return hasValidLength(control.value) && control.value.length > maxLength ?
+          {'maxlength': {'requiredLength': maxLength, 'actualLength': control.value.length}} :
           null;
     };
   }
```

@sonukapoor @petebacondarwin what do you think about this?

Thank you.

petebacondarwin: I like the idea of simplifying this but I believe that such a change would be breaking, since according to the original issue, developers are actually using `minlength` validation with numbers and expect it to be able to invalidate the control if the number has too few digits.

It can be argued that developers should not be using minlength (and maxlength) with numerical inputs, since if the number of digits is important perhaps they should be using a string input and applying a regular expression pattern?

If you are happy with such a breaking change, this would have to wait for v10.

sonukapoor: I agree with @petebacondarwin. I personally find `minLength` very useful on numbers. Considering the above would essentially allow all numeric values to pass regardless of what `minLength` was set to.

sonukapoor:
@AndrewKushnir / @petebacondarwin Something is wrong with CI. It's failing with a 403 error.

petebacondarwin:
Not your fault @sonukapoor - the setup task makes requests to the Github API in order to check some stuff related to the current master commits, etc. Since this is done via the unauthenticated API, we occasionally overrun the number of allowed requests.  I'll re-run the job and see if it gets through.

packages/forms/src/validators.ts
AndrewKushnir: As we discussed offline, the `minLength` and `maxLength` validators do not work currently with number values, so the change should not be breaking.

AndrewKushnir (changes requested):
The changes looks good, thanks @sonukapoor 👍 

I requested changes to update the "breaking changes" doc to include a recommendation on what should be the action items for someone who faced this breaking change. Here is a draft version, feel free to change it as you see fit:

* The `minLength` and `maxLength` validators now verify that a value has numeric `length` property and invoke validation only if that's the case. Previously, falsey values without the `length` property (such as `0` or `false` values) were triggering validation errors. If your code relies on the old behavior, you can include other validators such as [min](https://angular.io/api/forms/Validators#min) or [requiredTrue](https://angular.io/api/forms/Validators#requiredTrue) to the list of validators for a particular field.

Note: I've merged 2 items into one, since they might appear as 2 separate entities in the changelog (and we need to keep them together).

Thank you.

sonukapoor:
@AndrewKushnir Thanks, your version sounds good to me. I will update both, the commit and the PR description.

sonukapoor:
@AndrewKushnir Updated.

packages/forms/test/validators_spec.ts
```diff
@@ -225,6 +225,23 @@ describe('Validators', () => {
...
     });
+
+    it('should always return null with numeric values', () => {
+      expect(Validators.minLength(1)(new FormControl(0))).toBeNull();
+      expect(Validators.minLength(1)(new FormControl(1))).toBeNull();
+      expect(Validators.minLength(1)(new FormControl(-1))).toBeNull();
+      expect(Validators.minLength(1)(new FormControl(+1))).toBeNull();
+    });
+
+    it('should return null when passing an object that contains a length property', () => {
```
AndrewKushnir, line 236: Could you please rename this test to indicate that we run validation logic for such cases? Maybe also add a case where validation would fail (due to insufficient `length` value), for example:

```
    it('should trigger validation for an objects that contains numeric length property', () => {
      const value = {length: 5, someValue: [1, 2, 3, 4, 5]};
      expect(Validators.minLength(1)(new FormControl(value))).toBeNull();
      expect(Validators.minLength(10)(new FormControl(value))).toEqual({
        'minlength': {'requiredLength': 10, 'actualLength': 5}
      });
    });
```

The `maxLength` validator test would require similar change as well.

AndrewKushnir (approved):
Thanks for updating breaking changes section Sonu! 👍 

The changes LGTM, I just left a final comment on the tests, could you please have a look when you get a chance?

AndrewKushnir (approved):

petebacondarwin (approved):

kara (approved):
Change LGTM, but can we update the first paragraph of the commit message to read:

```
Previously, the behavior of the `minLength` and `maxLength` validators
caused confusion, as they appeared to work with numeric values but 
did not in fact produce consistent results. This commit fixes the issue 
by skipping validation altogether when a numeric value is used.
```

AndrewKushnir:
@sonukapoor I'll start Global Presubmit tonight and if it goes well, we'll be able to add this PR to the merge queue. I noticed that it was rebased 25 days ago, so it'd be great to do one more rebase, so this PR is in the up-to-date state and we have a fresh CI run. Thank you.

AndrewKushnir:
- [Blueprint Presubmit](http://test/OCL:309863291:BASE:309863316:1588643956562:6da34b8c)
- [Global Presubmit](https://test.corp.google.com/OCL:309863291:BASE:309863824:1588644221027:c594a2c8)

sonukapoor:
@Kara I have changed the message.

@AndrewKushnir All done and rebased.

AndrewKushnir:
Thanks for the updates @sonukapoor 👍 

FYI, presubmits went well, so I'm marking this PR as ready for merge.

PR #36162, opened 2020-03-20
feat(dev-infra): handle excluding files via globs in pullapprove

josephperrott:
Updates the pullapprove verification script to handle
cases of excluding globs from groups.

Unblocks #35933

### DIFF
--- a/dev-infra/pullapprove/group.ts
+++ b/dev-infra/pullapprove/group.ts
@@ -1,0 +1,158 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import {IMinimatch, Minimatch, match} from 'minimatch';
+
+import {PullApproveGroupConfig} from './parse-yaml';
+
+/** A condition for a group. */
+interface GroupCondition {
+  glob: string;
+  matcher: IMinimatch;
+  matchedFiles: Set<string>;
+}
+
+/** Result of testing files against the group. */
+export interface PullApproveGroupResult {
+  groupName: string;
+  matchedIncludes: GroupCondition[];
+  matchedExcludes: GroupCondition[];
+  matchedCount: number;
+  unmatchedIncludes: GroupCondition[];
+  unmatchedExcludes: GroupCondition[];
+  unmatchedCount: number;
+}
+
+// Regex Matcher for contains_any_globs conditions
+const CONTAINS_ANY_GLOBS_REGEX = /^'([^']+)',?$/;
+
+const CONDITION_TYPES = {
+  INCLUDE_GLOBS: /^contains_any_globs/,
+  EXCLUDE_GLOBS: /^not contains_any_globs/,
+  ATTR_LENGTH: /^len\(.*\)/,
+};
+
+/** A PullApprove group to be able to test files against. */
+export class PullApproveGroup {
+  // Lines which were not able to be parsed as expected.
+  private misconfiguredLines: string[] = [];
+  // Conditions for the group for including files.
+  private includeConditions: GroupCondition[] = [];
+  // Conditions for the group for excluding files.
+  private excludeConditions: GroupCondition[] = [];
+  // Whether the group has file matchers.
+  public hasMatchers = false;
+
+  constructor(public groupName: string, group: PullApproveGroupConfig) {
+    for (let condition of group.conditions) {
+      condition = condition.trim();
+
+      if (condition.match(CONDITION_TYPES.INCLUDE_GLOBS)) {
+        const [conditions, misconfiguredLines] = getLinesForContainsAnyGlobs(condition);
+        conditions.forEach(globString => this.includeConditions.push({
+          glob: globString,
+          matcher: new Minimatch(globString, {dot: true}),
+          matchedFiles: new Set<string>(),
+        }));
+        this.misconfiguredLines.push(...misconfiguredLines);
+        this.hasMatchers = true;
+      } else if (condition.match(CONDITION_TYPES.EXCLUDE_GLOBS)) {
+        const [conditions, misconfiguredLines] = getLinesForContainsAnyGlobs(condition);
+        conditions.forEach(globString => this.excludeConditions.push({
+          glob: globString,
+          matcher: new Minimatch(globString, {dot: true}),
+          matchedFiles: new Set<string>(),
+        }));
+        this.misconfiguredLines.push(...misconfiguredLines);
+        this.hasMatchers = true;
+      } else if (condition.match(CONDITION_TYPES.ATTR_LENGTH)) {
+        // Currently a noop as we do not take any action on this condition type.
+      } else {
+        const errMessage = `Unrecognized condition found, unable to parse the following condition: \n\n` +
+        `From the [${groupName}] group:\n` +
+        ` - ${condition}` +
+        `\n\n` +
+        `Known condition regexs:\n` + 
+        `${Object.entries(CONDITION_TYPES).map(([k, v]) => `  ${k}  -  ${v}`).join('\n')}` +
+        `\n\n`;
+        console.error(errMessage);
+        process.exit(1);
+      }
+    }
+  }
+
+  /** Retrieve all of the lines which were not able to be parsed. */
+  getBadLines(): string[] { return this.misconfiguredLines; }
+
+  /** Retrieve the results for the Group, all matched and unmatched conditions. */
+  getResults(): PullApproveGroupResult {
+    const matchedIncludes = this.includeConditions.filter(c => !!c.matchedFiles.size);
+    const matchedExcludes = this.excludeConditions.filter(c => !!c.matchedFiles.size);
+    const unmatchedIncludes = this.includeConditions.filter(c => !c.matchedFiles.size);
+    const unmatchedExcludes = this.excludeConditions.filter(c => !c.matchedFiles.size);
+    const unmatchedCount = unmatchedIncludes.length + unmatchedExcludes.length;
+    const matchedCount = matchedIncludes.length + matchedExcludes.length;
+    return {
+      matchedIncludes,
+      matchedExcludes,
+      matchedCount,
+      unmatchedIncludes,
+      unmatchedExcludes,
+      unmatchedCount,
+      groupName: this.groupName,
+    };
+  }
+
+  /**
+   * Tests a provided file path to determine if it would be considered matched by
+   * the pull approve group's conditions.
+   */
+  testFile(file: string) {
+    let matched = false;
+    this.includeConditions.forEach((includeCondition: GroupCondition) => {
+      if (includeCondition.matcher.match(file)) {
+        let matchedExclude = false;
+        this.excludeConditions.forEach((excludeCondition: GroupCondition) => {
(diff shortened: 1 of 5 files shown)


petebacondarwin:
Just to be clear. PullApprove supports the negation but our verification script did not?

josephperrott:
Correct, our verification script did not handle the case.  Pullapprove has always allowed negation in its conditions.

dev-infra/pullapprove/verify.ts
```diff
@@ -6,210 +6,111 @@
...
-
-    // Provide correct exit code based on verification success.
-    process.exit(verificationSucceeded ? 0 : 1);
+  });
+  // Results for each group
+  const resultsByGroup = groupsWithMatchers.map(group => group.getResults());
+  // Whether all group condition lines match at least one file and all files
+  // are matched by at least one group.
+  const verificationSucceeded =
+      resultsByGroup.every(r => !r.unmatchedCount) && !unmatchedFiles.length;
```
gkalpak: This verifies that all groups have at least one condition that matched at least one file. Previously that groups only had "include" conditions that was enough to verify that a group matched at least one file. Now, however, that groups can also have "exclude" conditions it is not (because the "include" conditions may have matched files, which the "exclude" conditions have then excluded, making the group not match anything).

So, we also need to capture the matched files for each group (i.e. the files for which `group.testFile()` return true).

gkalpak (changes requested):
Yay for splitting into smaller, focused files :tada:

dev-infra/pullapprove/verify.ts
```diff
@@ -6,210 +6,111 @@
...
-  // Exit early on shelljs errors
-  set('-e');
-
-  // Regex Matcher for contains_any_globs conditions
-  const CONTAINS_ANY_GLOBS_REGEX = /^'([^']+)',?$/;
+  // Whether to log verbosely
+  const VERBOSE_MODE = process.argv.includes('-v');
   // Full path of the angular project directory
-  const ANGULAR_PROJECT_DIR = process.cwd();
+  const ANGULAR_PROJECT_DIR = exec(`git rev-parse --show-toplevel`, {silent: true}).trim();
```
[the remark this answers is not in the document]
josephperrott: This is both to allow for usage in other repos as well as properly handling if the script is run from a different directory.

dev-infra/pullapprove/verify.ts
```diff
@@ -6,210 +6,111 @@
...
-  /** Create logs for each review group. */
-  function logGroups(groups: Map<string, Map<string, GlobMatcher>>) {
-    Array.from(groups.entries()).sort().forEach(([groupName, globs]) => {
-      console.groupCollapsed(groupName);
-      Array.from(globs.values())
-          .sort((a, b) => b.matchCount - a.matchCount)
-          .forEach(glob => console.info(`${glob.glob} - ${glob.matchCount}`));
-      console.groupEnd();
+  const REPO_FILES =
+      exec('git ls-files', {silent: true}).trim().split('\n').filter((_: string) => !!_);
```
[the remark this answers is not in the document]
josephperrott: By using `ls-files` we get the files both committed and staged, whereas with `ls-tree` we only get the files committed in the repo.

dev-infra/pullapprove/verify.ts
```diff
@@ -6,210 +6,111 @@
...
-  const ALL_FILES = exec('git ls-tree --full-tree -r --name-only HEAD', {silent: true})
-                        .trim()
-                        .split('\n')
-                        .filter((_: string) => !!_);
-  if (!ALL_FILES.length) {
-    console.error(
-        `No files were found to be in the git tree, did you run this command from \n` +
-        `inside the angular repository?`);
-    process.exit(1);
-  }
```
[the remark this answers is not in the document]
josephperrott: We not longer need to perform this check since we automatically move our working directory to the root directory of the git project

dev-infra/pullapprove/verify.ts
josephperrott: I agree, that the case of files being simultaneously matched by two conditions is something we need to consider.

I believe this case is covered by the fact that we require all files to pass the `group.testFile()` method.

For instance, if I add this condition to the dev-infra group conditions:
```yaml
- >
  not contains_any_globs(files, [
    'dev-infra/pullapprove/**'
    ])
```

The output from the script is:
```
┌──────────────────────────────────────────────────────────────────────────────┐
│                                Overall Result                                │
└──────────────────────────────────────────────────────────────────────────────┘
PullApprove verification failed.

Please update '.pullapprove.yml' to ensure that all necessary
files/directories have owners and all patterns that appear in
the file correspond to actual files/directories in the repo.
┌──────────────────────────────────────────────────────────────────────────────┐
│                          PullApprove results by file                         │
└──────────────────────────────────────────────────────────────────────────────┘
Matched Files (9070 files)
Unmatched Files (5 files)
  dev-infra/pullapprove/BUILD.bazel
  dev-infra/pullapprove/group.ts
  dev-infra/pullapprove/logging.ts
  dev-infra/pullapprove/parse-yaml.ts
  dev-infra/pullapprove/verify.ts
┌──────────────────────────────────────────────────────────────────────────────┐
│                         PullApprove results by group                         │
└──────────────────────────────────────────────────────────────────────────────┘
Groups without matchers (1 groups)
Matched conditions by Group (34 groups)
Unmatched conditions by Group (0 groups)
```

.pullapprove.yml
```diff
@@ -917,6 +917,10 @@ groups:
   # =========================================================
   dev-infra:
     conditions:
+      - >
+        not contains_any_globs(files, [
+          'dev-infra/pullapprove/**'
+          ])
```
gkalpak: Is this for testing purposes?

dev-infra/pullapprove/verify.ts
```diff
@@ -6,210 +6,112 @@
...
-  matchCount: number;
-}
+import {PullApproveGroup} from './group';
+import {logGroup, logHeader} from './logging';
+import {parsePullApproveYaml} from './parse-yaml';
+import {getRepoBaseDir} from '../utils/config';
 
 export function verify() {
-  // Exit early on shelljs errors
-  set('-e');
```
gkalpak: Why remove this?

dev-infra/pullapprove/verify.ts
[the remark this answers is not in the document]
gkalpak: Doh! I didn't notice the `ls-tree` to `ls-files` change :grin:

dev-infra/pullapprove/group.ts
```diff
@@ -0,0 +1,140 @@
...
+          glob: globString,
+          matcher: new Minimatch(globString, {dot: true}),
+          matchedFiles: new Set<string>(),
+        }));
+        this.misconfiguredLines.push(...misconfiguredLines);
+        this.hasMatchers = true;
+        continue;
+      }
+
+      if (condition.startsWith('not contains_any_globs')) {
```
gkalpak: Nit: The `continue` statements are redundant here (and below).

```suggestion
      } else if (condition.startsWith('not contains_any_globs')) {
```

.pullapprove.yml
gkalpak: Is this for testing purposes?

dev-infra/pullapprove/verify.ts
gkalpak: Why remove this?

dev-infra/pullapprove/verify.ts
```diff
@@ -6,210 +6,112 @@
...
+  // The pull approve config file.
+  const pullApproveYamlRaw = readFileSync(PULL_APPROVE_YAML_PATH, 'utf8');
+  // JSON representation of the pullapprove yaml file.
+  const pullApprove = parsePullApproveYaml(pullApproveYamlRaw);
+  // All of the groups defined in the pullapprove yaml.
+  const groups = Object.entries(pullApprove.groups).map(([groupName, group]) => {
+    return new PullApproveGroup(groupName, group);
+  });
+  // PullApprove groups without matchers.
+  const groupsWithoutMatchers = groups.filter(group => !group.hasMatchers);
```
gkalpak: Personally, I feel that groups without matchers (or actually even unknown group conditions) should cause a failure. For example, it is easy to misspell `contains_any_globs` and never match any files (which might be matched by other groups).

dev-infra/pullapprove/verify.ts
gkalpak: It is true that in most cases this misconfiguration will result in some error. But there are situation where verification will succeed with invalid config (i.e. if the files are covered by another group).

For example, in #35933 the intention is to exclude `packages/compiler-cli/ngcc/` from the `compiler-cli` group by adding the following condition to the [`fw-compiler-cli` group][1]:

```yaml
- >
  not contains_any_globs(files, [
    'packages/compiler-cli/ngcc/**'
    ])
```

If we were to accidentally add it to the [`fw-ngcc` group][2], then the verification would pass, because `packages/compiler-cli/ngcc/` files would be matched by `fw-compiler-cli`, and both conditions of `fw-ngcc` would match files. Yet, the config would not be the intended one (and we could have caught that, because the `fw-ngcc` group would not match any files - although both of its conditions would match files).

[1]: https://github.com/angular/angular/blob/9d415f9c3f12cb2c29298ba414f55ce230201700/.pullapprove.yml#L146-L166
[2]: https://github.com/angular/angular/blob/9d415f9c3f12cb2c29298ba414f55ce230201700/.pullapprove.yml#L172-L186

gkalpak (changes requested):

dev-infra/pullapprove/verify.ts
josephperrott: We actually have a case already where we have a group without matchers.

We have our fallback group, which only has a condition that no other groups are matched.

gkalpak: My main concern is that we might misspell a condition and never realize it. So, I suggest we have a way to detect groups with unknown conditions and fail if there are any.

dev-infra/pullapprove/verify.ts
josephperrott: I think I got it with my change I just made, PTAL.

dev-infra/pullapprove/verify.ts
```diff
@@ -6,210 +6,114 @@
...
-  matchCount: number;
-}
+import {PullApproveGroup} from './group';
+import {logGroup, logHeader} from './logging';
+import {parsePullApproveYaml} from './parse-yaml';
+import {getRepoBaseDir} from '../utils/config';
 
 export function verify() {
-  // Exit early on shelljs errors
+  // Exit early on shelljs errors
```
gkalpak: 🤔 
```suggestion
  // Exit early on shelljs errors
```

dev-infra/pullapprove/group.ts
```diff
@@ -0,0 +1,158 @@
...
+        let matchedExclude = false;
+        this.excludeConditions.forEach((excludeCondition: GroupCondition) => {
+          if (excludeCondition.matcher.match(file)) {
+            // Add file as a discovered exclude as it is negating a matched
+            // include condition.
+            excludeCondition.matchedFiles.add(file);
+            matchedExclude = true;
+          }
+        });
+        // An include condition is only considered matched if it no exclude
```
gkalpak:
```suggestion
        // An include condition is only considered matched if no exclude
```

gkalpak (approved):
LGTM 👌

PR #36165, opened 2020-03-20
fix(dev-infra): prep ts-circular-deps to load via node_modules

josephperrott:
To run ts-circular-deps via installed node_modules, we needed to set
the hashbang of the script to be a node environment, and discover the
project directory based on where the script is run rather than the
scripts file location.


devversion (approved):

josephperrott:
@devversion can you take another look as I had to add another commit.

devversion (approved):
👍 Looks still good to me!

PR #36172, opened 2020-03-20
fix(dev-infra): use commit message validation from @angular/dev-infra-private

josephperrott:


yarn.lock
```diff
@@ -153,6 +153,10 @@
     universal-analytics "^0.4.20"
     uuid "^3.3.2"
 
+"@angular/dev-infra-private@angular/dev-infra-private-builds#204e328":
```
filipesilva: Is this right? I don't think the yarn lock should have the package here if it's not on package.json.

gkalpak: It _is_ in [package.json#L49][1] 😕 

[1]: https://github.com/angular/angular/blob/c8c6386a45dd936189126cb82dc00229341b934d/package.json#L49

filipesilva: Oh jeez I missed it, thanks George!

filipesilva (approved):

josephperrott:
Blocked by https://github.com/angular/angular/pull/36326

josephperrott:
@IgorMinar @gkalpak @filipesilva This is ready for another look as I had to do a few updates to build it from local.

dev-infra/commit-message/cli.ts
```diff
@@ -7,13 +7,27 @@
...
+      .strict()
+      .command(
+          'pre-commit-validate', 'Validate the most recent commit message', {},
+          () => {
+            validateFile('.git/COMMIT_EDITMSG');
+          })
+      .command(
+          'validate-range', 'Validate a range of commit messages', {
+            'range': {
+              description: 'The range of commits check',
```
IgorMinar: maybe provide an example as well so that it's easier to use?

dev-infra/commit-message/cli.ts
```diff
@@ -7,13 +7,27 @@
  */
 import * as yargs from 'yargs';
 import {validateFile} from './validate-file';
-
```
gkalpak, line 10: 😞

dev-infra/commit-message/cli.ts
gkalpak, line 23:
```suggestion
              description: 'The range of commits to check',
```

dev-infra/commit-message/validate-range.ts
```diff
@@ -0,0 +1,46 @@
...
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import {exec} from 'shelljs';
+import {parseCommitMessage, validateCommitMessage, ValidateCommitMessageOptions} from './validate';
+
+// Whether the provided commit is a fixup commit.
+const isNonFixup = (m: string) => parseCommitMessage(m).isFixup;
```
gkalpak, line 12:
```suggestion
const isNonFixup = (m: string) => !parseCommitMessage(m).isFixup;
```

:grin:

dev-infra/commit-message/validate-range.ts
```diff
@@ -0,0 +1,46 @@
...
+  const gitLogFormat = `%s%n%n%b${randomValueSeparator}`;
+
+  // Retrieve the commits in the provided range.
+  const result = exec(`git log --reverse --format=${gitLogFormat} ${range}`, {silent: true});
+  if (result.code) {
+    throw new Error(`Failed to get all commits in the range: \n  ${result.stderr}`);
+  }
+
+  // Separate the commits from a single string into individual commits
+  const commits = result.split(randomValueSeparator).map(l => l.trim()).filter(line => line != '');
```
gkalpak, line 29: Nit:

```suggestion
  const commits = result.split(randomValueSeparator).map(l => l.trim()).filter(line => line !== '');
```

dev-infra/commit-message/validate-range.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  console.info(`Examining ${commits.length} commit(s) in the provided range: ${range}`);
+
+  // Check if each commit in the range is valid, providing the possible commits that could be
+  // fixed up into.
```
gkalpak, line 34: Into?

dev-infra/commit-message/validate-range.ts
```diff
@@ -0,0 +1,46 @@
...
+  const allCommitsInRangeValid = commits.every((m, i) => {
+    const options: ValidateCommitMessageOptions = {
+      disallowSquash: true,
+      nonFixupCommitHeaders: isNonFixup(m) ? undefined : commits.slice(0, i).filter(isNonFixup),
+      ignoreHeaderLength: true,
```
gkalpak, line 39: Doesn't this mean that we can end up with too long commits (if people decide to trick our validation)?
Not the end of the world, but why not keep the previous behavior: Validate on CI for PR-builds only and do not ignore header length?

gkalpak (changes requested):
Yay for deleting code 🤤 🎉

IgorMinar (approved):
ironically - can you please update the commit message for this change to answer the "why this change?" question?

thanks


lgtm otherwise.

dev-infra/commit-message/validate-range.ts
```diff
@@ -0,0 +1,45 @@
...
+    throw new Error(`Failed to get all commits in the range: \n  ${result.stderr}`);
+  }
+
+  // Separate the commits from a single string into individual commits
+  const commits = result.split(randomValueSeparator).map(l => l.trim()).filter(line => !!line);
+
+  console.info(`Examining ${commits.length} commit(s) in the provided range: ${range}`);
+
+  // Check if each commit in the range is valid, providing the possible commits that could be
+  // FIXUPs could be linked to.
```
gkalpak, line 34: Something is off with this last clause 😕

gkalpak (approved):
:+1:

PR #36180, opened 2020-03-20
fix(ngcc): use path-mappings from tsconfig in dependency resolution

### LINKED ISSUE #36119: NGCC create-ivy-entry-points for all project libraries before parallel test
We based our new build pipeline on standard `ng cli` commands. Especially for parallel builds, we create a dependency tree reading library's package.json file and run ng build command on our libraries in a parallel way.

For this particular task, the `ngcc` call before all compilation steps works like a charm (see https://angular.io/guide/ivy#speeding-up-ngcc-compilation).

The main problem now is that we would like to run tests also on parallel way after we build our libraries. Actually, `ng test libraryName` command, just tries to compile in 'ivy' way all our libraries it depends on, every time we execute tests.

Running our CI pipeline with a sequential test approach doesn't raise errors (but it is slower). Feature requested is just related to a way to compile our project libraries on dist folder in 'ivy' way before run tests to support also parallel tests.

See for example error from our teamcity agent:

```
Start to test projects parallel with 4 cpus
08:13:46
    Start test of @aulos/core with dependencies: []
08:13:46
    Start test of @aulos/common with dependencies: []
08:13:46
    Start test of @aulos/common-validators with dependencies: []
08:13:55
    JavaScript Unit Tests
08:13:55
      18 03 2020 08:13:55.510:INFO [karma-server]: Karma v4.2.0 server started at http://0.0.0.0:9877/
08:13:55
      18 03 2020 08:13:55.526:INFO [launcher]: Launching browsers ChromeHeadless with concurrency unlimited
08:13:55
      18 03 2020 08:13:55.545:INFO [launcher]: Starting browser ChromeHeadless
08:13:55
      JavaScript Unit Tests
08:13:56
      ERROR in ngcc is already running at process with id 4565.
08:13:56
      If you are running multiple builds in parallel then you should pre-process your node_modules via the command line ngcc tool before starting the builds;
08:13:56
      See https://v9.angular.io/guide/ivy#speeding-up-ngcc-compilation
08:13:56
      (If you are sure no ngcc process is running then you should delete the lockfile at /opt/teamcity-agent/work/d27e3f1950cca49c/node_modules/@angular/compiler-cli/ngcc/__ngcc_lock_file__.)
08:13:56
      End test process of @aulos/core with code 1
08:13:56
      We have to test 3 libraries with 2 free cpu.
08:13:56
      Start test of @aulos/security with dependencies: [@aulos/core]
08:13:56
      Start test of @aulos/globalization with dependencies: [@aulos/core]
08:13:56
      Start test of @aulos/binary-resources with dependencies: [@aulos/core]
08:13:59
      
08:13:59
      Compiling @aulos/common : es2015 as esm2015
08:14:01
      18 03 2020 08:14:01.074:INFO [karma-server]: Karma v4.2.0 server started at http://0.0.0.0:9878/
08:14:01
      18 03 2020 08:14:01.075:INFO [launcher]: Launching browsers ChromeHeadless with concurrency unlimited
08:14:01
      18 03 2020 08:14:01.084:INFO [launcher]: Starting browser ChromeHeadless
```

As you can see we run 3 concurrency tests but they require building our libraries (for example @aulos/common).

So request here is to enable `ngcc ..... --create-ivy-entry-points` also on ./dist/libraries folders.

petebacondarwin:
When computing the dependencies between packages which are not in
node_modules, we may need to rely upon path-mappings to find the path
to the imported entry-point.

This commit allows ngcc to use the path-mappings from a tsconfig
file to find dependencies. By default any tsconfig.json file in the directory
above the `basePath` is loaded but it is possible to use a path to a
specific file by providing the `tsConfigPath` property to mainNgcc,
or to turn off loading any tsconfig file by setting `tsConfigPath` to `null`.
At the command line this is controlled via the `--tsconfig` option.

Fixes #36119

### DIFF
--- a/packages/compiler-cli/ngcc/main-ngcc.ts
+++ b/packages/compiler-cli/ngcc/main-ngcc.ts
@@ -121,8 +121,9 @@ if (require.main === module) {
   const invalidateEntryPointManifest = options['invalidate-entry-point-manifest'];
   const errorOnFailedEntryPoint = options['error-on-failed-entry-point'];
   // yargs is not so great at mixed string+boolean types, so we have to test tsconfig against a
-  // string "false" to capture the `tsconfig=false` option
-  const tsConfigPath = (options['tsconfig'] === 'false') ? null : options['tsconfig'];
+  // string "false" to capture the `tsconfig=false` option.
+  // And we have to convert the option to a string to handle `no-tsconfig`, which will be `false`.
+  const tsConfigPath = `${options['tsconfig']}` === 'false' ? null : options['tsconfig'];
 
   (async() => {
     try {
--- a/packages/compiler-cli/ngcc/main-ngcc.ts
+++ b/packages/compiler-cli/ngcc/main-ngcc.ts
@@ -92,6 +92,19 @@ if (require.main === module) {
             type: 'boolean',
             default: false,
           })
+          .option('tsconfig', {
+            describe:
+                'A path to a tsconfig.json file that will be used to configure the Angular compiler and module resolution used by ngcc.\n' +
+                'If not provided, ngcc will attempt to read a `tsconfig.json` file from the folder above that given by the `-s` option.\n' +
+                'Set to false (via `--no-tsconfig`) if you do not want ngcc to use any `tsconfig.json` file.',
+            type: 'string',
+            default: '',
+          })
+          .option('no-tsconfig', {
+            describe: 'This option is to support forcing ngcc not to use any `tsconfig.json` file.',
+            type: 'boolean',
+            hidden: true,
+          })
           .strict()
           .help()
           .parse(args);
@@ -113,6 +126,9 @@ if (require.main === module) {
   const enableI18nLegacyMessageIdFormat = options['legacy-message-ids'];
   const invalidateEntryPointManifest = options['invalidate-entry-point-manifest'];
   const errorOnFailedEntryPoint = options['error-on-failed-entry-point'];
+  // yargs is not so great at mixed string+boolean types, so we have to test tsconfig against a
+  // string "false" to capture the `no-tsconfig` option
+  const tsConfigPath = (options['tsconfig'] === 'false') ? null : options['tsconfig'];
 
   (async() => {
     try {
@@ -126,7 +142,7 @@ if (require.main === module) {
         createNewEntryPointFormats,
         logger,
         enableI18nLegacyMessageIdFormat,
-        async: options['async'], invalidateEntryPointManifest, errorOnFailedEntryPoint,
+        async: options['async'], invalidateEntryPointManifest, errorOnFailedEntryPoint, tsConfigPath
       });
 
       if (logger) {
--- a/packages/compiler-cli/ngcc/main-ngcc.ts
+++ b/packages/compiler-cli/ngcc/main-ngcc.ts
@@ -98,12 +98,6 @@ if (require.main === module) {
                 'If not provided, ngcc will attempt to read a `tsconfig.json` file from the folder above that given by the `-s` option.\n' +
                 'Set to false (via `--no-tsconfig`) if you do not want ngcc to use any `tsconfig.json` file.',
             type: 'string',
-            default: '',
-          })
-          .option('no-tsconfig', {
-            describe: 'This option is to support forcing ngcc not to use any `tsconfig.json` file.',
-            type: 'boolean',
-            hidden: true,
           })
           .strict()
           .help()
@@ -127,7 +121,7 @@ if (require.main === module) {
   const invalidateEntryPointManifest = options['invalidate-entry-point-manifest'];
   const errorOnFailedEntryPoint = options['error-on-failed-entry-point'];
   // yargs is not so great at mixed string+boolean types, so we have to test tsconfig against a
-  // string "false" to capture the `no-tsconfig` option
+  // string "false" to capture the `tsconfig=false` option
   const tsConfigPath = (options['tsconfig'] === 'false') ? null : options['tsconfig'];
 
   (async() => {
--- a/packages/compiler-cli/ngcc/src/main.ts
+++ b/packages/compiler-cli/ngcc/src/main.ts
@@ -95,6 +95,9 @@ export interface SyncNgccOptions {
   /**
    * Paths mapping configuration (`paths` and `baseUrl`), as found in `ts.CompilerOptions`.
    * These are used to resolve paths to locally built Angular libraries.
+   *
+   * Note that `pathMappings` specified here take precedence over any `pathMappings` loaded from a
+   * TS config file.
    */
   pathMappings?: PathMappings;
 
@@ -146,15 +149,16 @@ export interface SyncNgccOptions {
   invalidateEntryPointManifest?: boolean;
 
   /**
-   * An absolte path to a TS config file (e.g. `tsconfig.json`) or a directory containing one that
-   * will be used to configure module resolution with things like path mappings.
+   * An absolute path to a TS config file (e.g. `tsconfig.json`) or a directory containing one, that
+   * will be used to configure module resolution with things like path mappings, if not specified
+   * explicitly via the `pathMappings` property to `mainNgcc`.
    *
    * If `undefined`, ngcc will attempt to load a `tsconfig.json` file from the directory above the
    * `basePath`.
    *
    * If `null`, ngcc will not attempt to load any TS config file at all.
    */
-  tsConfigPath?: AbsoluteFsPath|null;
+  tsConfigPath?: string|null;
 }
 
 /**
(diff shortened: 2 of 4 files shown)


packages/compiler-cli/ngcc/main-ngcc.ts
```diff
@@ -92,6 +92,19 @@ if (require.main === module) {
...
           })
+          .option('tsconfig', {
+            describe:
+                'A path to a tsconfig.json file that will be used to configure the Angular compiler and module resolution used by ngcc.\n' +
+                'If not provided, ngcc will attempt to read a `tsconfig.json` file from the folder above that given by the `-s` option.\n' +
+                'Set to false (via `--no-tsconfig`) if you do not want ngcc to use any `tsconfig.json` file.',
+            type: 'string',
+            default: '',
+          })
+          .option('no-tsconfig', {
```
alan-agius4: Nice workaround for negation of string arguments.

packages/compiler-cli/ngcc/src/main.ts
```diff
@@ -185,6 +197,19 @@ export function mainNgcc({basePath, targetEntryPointPath,
...
   const config = new NgccConfiguration(fileSystem, dirname(absBasePath));
+  const tsConfig =
+      tsConfigPath !== null ? readConfiguration(tsConfigPath || dirname(absBasePath)) : null;
+
+  // If pathMapping is not provided directly and there is a loaded tsConfig with path mappings then
+  // use them instead.
+  if (tsConfig !== null && pathMappings === undefined && tsConfig.options.baseUrl !== undefined &&
+      tsConfig.options.paths) {
+    pathMappings = {
+      baseUrl: resolve(absBasePath, tsConfig.options.baseUrl || './'),
```
JoostK: Is this correct?

If `tsConfig.options.baseUrl` is set, I would expect it to be relative to the path where the tsconfig file is stored. However, `absBasePath` is typically a `node_modules` directory.

JoostK: Also, the guard tests for `tsConfig.options.baseUrl !== undefined` so is the fallback of `'./'` relevant?

packages/compiler-cli/ngcc/test/integration/ngcc_spec.ts
```diff
@@ -1246,6 +1246,58 @@ runInEachFileSystem(() => {
...
+           mainNgcc({
+             basePath: '/dist',
+             propertiesToConsider: ['es2015'],
+             tsConfigPath: null,
+           });
+           expect(loadPackage('local-package', _('/dist')).__processed_by_ivy_ngcc__).toEqual({
+             es2015: '0.0.0-PLACEHOLDER',
+             typings: '0.0.0-PLACEHOLDER',
+           });
+           expect(loadPackage('local-package-2', _('/dist')).__processed_by_ivy_ngcc__)
```
JoostK: Please add a comment here that `local-package-2` should not have been processed, as without path mappings the `@app/local-package` import is assumed to be missing.

packages/compiler-cli/ngcc/src/main.ts
petebacondarwin: Good question about the `basePath`. I'll double check what is happening there.

packages/compiler-cli/ngcc/test/integration/ngcc_spec.ts
petebacondarwin: I'll also consider a check of the logger to show that it complained about the missing import.

packages/compiler-cli/ngcc/src/main.ts
petebacondarwin: You were right that using the `basePath` was incorrect but it turns out that `tsConfig.options.baseUrl` is already absolute even though it is relative in the tsconfig.json file.

packages/compiler-cli/ngcc/test/integration/ngcc_spec.ts
```diff
@@ -1246,6 +1246,58 @@ runInEachFileSystem(() => {
...
+               .toBeUndefined();
+         });
```
gkalpak: I think it would be useful to have some tests showing how `tsConfigPath` works together with `pathMappings`.

gkalpak (changes requested):
Yesss :tada:

packages/compiler-cli/ngcc/main-ngcc.ts
petebacondarwin: I created a test project for this:

```
const yargs = require('yargs');

const options = yargs
  .option('tsconfig', {describe: 'specify tsconfig file', type: 'string', default: ''})
  .option('no-tsconfig', {describe: 'do not use tsconfig', type: 'boolean', hidden: true})
  .help()
  .parse(process.argv.slice(1));

console.log(options);
console.log(options.tsconfig);
console.log('force no tsconfig', '' + options.tsconfig === 'false')
```

Here is some example output:

```
➜  yargs-tester node index.js --tsconfig
{
  _: [ '/Users/pete/dev/angular/tmp/yargs-tester/index.js' ],
  tsconfig: '',
  '$0': 'index.js'
}

force no tsconfig false
➜  yargs-tester node index.js --tsconfig abc
{
  _: [ '/Users/pete/dev/angular/tmp/yargs-tester/index.js' ],
  tsconfig: 'abc',
  '$0': 'index.js'
}
abc
force no tsconfig false
➜  yargs-tester node index.js --tsconfig=false
{
  _: [ '/Users/pete/dev/angular/tmp/yargs-tester/index.js' ],
  tsconfig: 'false',
  '$0': 'index.js'
}
false
force no tsconfig true
➜  yargs-tester node index.js --no-tsconfig      
{
  _: [ '/Users/pete/dev/angular/tmp/yargs-tester/index.js' ],
  tsconfig: false,
  '$0': 'index.js'
}
false
force no tsconfig true
```

petebacondarwin: OMG! I just removed the `no-tsconfig` line and it works the same....

petebacondarwin: I spent ages playing around with these different options and I was sure that this was the only solution, but it looks like you are correct @gkalpak... Yet more kudos to you.

packages/compiler-cli/ngcc/main-ngcc.ts
```diff
@@ -113,6 +126,9 @@ if (require.main === module) {
   const enableI18nLegacyMessageIdFormat = options['legacy-message-ids'];
   const invalidateEntryPointManifest = options['invalidate-entry-point-manifest'];
   const errorOnFailedEntryPoint = options['error-on-failed-entry-point'];
+  // yargs is not so great at mixed string+boolean types, so we have to test tsconfig against a
+  // string "false" to capture the `no-tsconfig` option
```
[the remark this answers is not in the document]
petebacondarwin: Setting `--tsconfig=false` results in `tsconfig` being the string `"false"`, so we do need to check for the string version. But it is true that we do not need the `no-tsconfig` option.

packages/compiler-cli/ngcc/src/main.ts
```diff
@@ -143,6 +144,17 @@ export interface SyncNgccOptions {
...
+  /**
+   * An absolte path to a TS config file (e.g. `tsconfig.json`) or a directory containing one that
+   * will be used to configure module resolution with things like path mappings.
+   *
+   * If `undefined`, ngcc will attempt to load a `tsconfig.json` file from the directory above the
+   * `basePath`.
+   *
+   * If `null`, ngcc will not attempt to load any TS config file at all.
+   */
+  tsConfigPath?: AbsoluteFsPath|null;
```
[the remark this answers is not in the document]
petebacondarwin: Cool. Good shout.

packages/compiler-cli/ngcc/src/main.ts
```diff
@@ -143,6 +144,17 @@ export interface SyncNgccOptions {
...
+
+  /**
+   * An absolte path to a TS config file (e.g. `tsconfig.json`) or a directory containing one that
+   * will be used to configure module resolution with things like path mappings.
+   *
+   * If `undefined`, ngcc will attempt to load a `tsconfig.json` file from the directory above the
+   * `basePath`.
+   *
+   * If `null`, ngcc will not attempt to load any TS config file at all.
+   */
```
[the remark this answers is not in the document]
petebacondarwin: Or both!

packages/compiler-cli/ngcc/test/integration/ngcc_spec.ts
petebacondarwin: OK

JoostK (approved):

packages/compiler-cli/ngcc/main-ngcc.ts
```diff
@@ -113,6 +120,9 @@ if (require.main === module) {
   const enableI18nLegacyMessageIdFormat = options['legacy-message-ids'];
   const invalidateEntryPointManifest = options['invalidate-entry-point-manifest'];
   const errorOnFailedEntryPoint = options['error-on-failed-entry-point'];
+  // yargs is not so great at mixed string+boolean types, so we have to test tsconfig against a
+  // string "false" to capture the `tsconfig=false` option
+  const tsConfigPath = (options['tsconfig'] === 'false') ? null : options['tsconfig'];
```
gkalpak: Unless I am missing something, we also need to set `tsConfigPath` to `null` when `tsconfig` is negated with `--no-tsconfig`. So, I would expect this to be:

```suggestion
  const tsConfigPath = (options['tsconfig'] === false || options['tsconfig'] === 'false') ? null : options['tsconfig'];
```

...or something like...

```suggestion
  const tsConfigPath = (`${options['tsconfig']}` === 'false') ? null : options['tsconfig'];
```

gkalpak (approved):
One last comment. Otherwise lgtm 👍 ✨ 🥇

packages/compiler-cli/ngcc/main-ngcc.ts
petebacondarwin: Darn it! You are right. Thanks.

petebacondarwin: Note that in my test project I was doing:

```
console.log('force no tsconfig', '' + options.tsconfig === 'false')
```

PR #36189, opened 2020-03-21
ngcc: preserve whitespace configuration

petebacondarwin:
Tracking at FW-2013

### DIFF
--- a/packages/compiler-cli/ngcc/src/analysis/decoration_analyzer.ts
+++ b/packages/compiler-cli/ngcc/src/analysis/decoration_analyzer.ts
@@ -8,6 +8,7 @@
 import {ConstantPool} from '@angular/compiler';
 import * as ts from 'typescript';
 
+import {ParsedConfiguration} from '../../..';
 import {ComponentDecoratorHandler, DirectiveDecoratorHandler, InjectableDecoratorHandler, NgModuleDecoratorHandler, PipeDecoratorHandler, ReferencesRegistry, ResourceLoader} from '../../../src/ngtsc/annotations';
 import {CycleAnalyzer, ImportGraph} from '../../../src/ngtsc/cycles';
 import {isFatalDiagnosticError} from '../../../src/ngtsc/diagnostics';
@@ -55,6 +56,7 @@ export class DecorationAnalyzer {
   private rootDirs = this.bundle.rootDirs;
   private packagePath = this.bundle.entryPoint.package;
   private isCore = this.bundle.isCore;
+  private compilerOptions = this.tsConfig !== null? this.tsConfig.options: {};
 
   moduleResolver =
       new ModuleResolver(this.program, this.options, this.host, /* moduleResolutionCache */ null);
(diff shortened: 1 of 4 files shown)


gkalpak (approved):
LGTM 👍

packages/compiler-cli/ngcc/src/analysis/decoration_analyzer.ts
```diff
@@ -55,6 +56,7 @@ export class DecorationAnalyzer {
   private rootDirs = this.bundle.rootDirs;
   private packagePath = this.bundle.entryPoint.package;
   private isCore = this.bundle.isCore;
+  private compilerOptions = this.tsConfig !== null? this.tsConfig.options: {};
```
JoostK: 0 points for the code formatter here!

packages/compiler-cli/ngcc/src/main.ts
```diff
@@ -143,6 +144,17 @@ export interface SyncNgccOptions {
    * Default: `false` (i.e. the manifest will be used if available)
    */
   invalidateEntryPointManifest?: boolean;
+
+  /**
+   * An absolte path to a TS config file (e.g. `tsconfig.json`) or a directory containing one that
```
devversion:
```suggestion
   * An absolute path to a TS config file (e.g. `tsconfig.json`) or a directory containing one that
```

JoostK (approved):

devversion (approved):
LGTM for dev-infra

packages/compiler-cli/ngcc/src/analysis/decoration_analyzer.ts
gkalpak: ??? 😕

JoostK: @gkalpak I've not come across a ternary that is formatted as badly as this one. The question mark as well as the colon only has whitespace to their right, it looks so weird.

I'm fine with keeping as is, for sure!

gkalpak: Wow, I didn't notice the missing whitespace and it looked fine to me 🤦‍♂

petebacondarwin: Nothing to be done in this PR about this.

alxhub (approved):

packages/compiler-cli/ngcc/src/main.ts
```diff
@@ -143,6 +147,18 @@ export interface SyncNgccOptions {
    * Default: `false` (i.e. the manifest will be used if available)
    */
   invalidateEntryPointManifest?: boolean;
+
+  /**
+   * An absolute path to a TS config file (e.g. `tsconfig.json`) or a directory containing one, that
+   * will be used to configure module resolution with things like path mappings, if not specified
+   * explicitly via the `pathMappings` property to `mainNgcc`.
+   *
+   * If `undefined`, ngcc will attempt to load a `tsconfig.json` file from the directory above the
```
alxhub: I don't think ngcc should select a tsconfig implicitly - it should either be passed explicitly, or not use one.

In Angular it's rare that the top-level tsconfig is intended for compilation.

petebacondarwin: It turns out that in CLI apps the top level tsconfig does actually commonly contain the information we require. Also I am talking to the CLI team to see if they can apply the correct tsconfig file during the async integration.

petebacondarwin: The CLI can pass the correct configuration: https://github.com/angular/angular-cli/pull/17295

Issue #20756, opened 2017-12-03
Service worker 6.x does not work correctly in Edge and Samsung Internet on Android


dmostert:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead see https://github.com/angular/angular/blob/master/CONTRIBUTING.md#question
</code></pre>

## Current behavior

App with service worker loads fine the first time in Chrome 56 (And Edge Android, Samsung Internet). If you reload the page, the app fails with:

ERR_FAILED

`Uncaught (in promise) DOMException: Failed to execute 'waitUntil' on 'ExtendableEvent': The event handler is already finished.
    at Driver.handleFetch (http://127.0.0.1:8887/ngsw-worker.js:1941:15)
handleFetch @ ngsw-worker.js:1941
`

## Expected behavior

Service worker enabled sites should work correctly on Chrome 56, Edge and Samsung Internet browsers

## Minimal reproduction of the problem with instructions

ng new repro --service-worker
ng build --prod
Use Web server for Chrome or related to browse the dist folder.

## What is the motivation / use case for changing the behavior?

Service worker should work correctly on all supported browsers.

## Environment

<pre><code>
Angular CLI: 1.6.0-rc.1
Node: 8.9.1
OS: win32 x64
Angular: 5.0.5
... animations, common, compiler, compiler-cli, core, forms
... http, language-service, platform-browser
... platform-browser-dynamic, router, service-worker

@angular/cli: 1.6.0-rc.1
@angular-devkit/build-optimizer: 0.0.34
@angular-devkit/core: 0.0.22
@angular-devkit/schematics: 0.0.38
@ngtools/json-schema: 1.1.0
@ngtools/webpack: 1.9.0-rc.1
@schematics/angular: 0.1.8
typescript: 2.4.2
webpack: 3.8.1

Browser:
- [x] Chrome (desktop) version 56
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [x] Edge version  ANDROID 
 
For Tooling issues:
- Node version: v8.9.1  
- Platform:  Windows 

Others:

`
`
</code></pre>

mlc-mlapis:
[...]

PR #36194, opened 2020-03-22
fix(service-worker): handle quota exceeded

### LINKED ISSUE #36086: Service Worker "Quota exceeded" can break all JS
# 🐞 bug report

### Affected Package

 The issue is caused by package @angular/service-worker

### Is this a regression?

 Not sure

### Description

The service worker fails to load core JS files with `DOMException: Quota exceeded.`

This results in just a white page being shown, with a broken app.

Angular 9.0.4
Chromium 80

image

image

I've migrated from a custom workbox based serviceworker to the angular native service worker and now running into this.

Force reloading (`ctrl+f5`) makes everything load, but the error returns on the next normal load.

After using DevTools to debug it, I found that the error is `DOMException: Quota exceeded.`  and occurs in this block: 

image

## 🔬 Minimal Reproduction

Hard to reproduce unfortunately, needs low disk space or a big app being cached.

## 🔥 Exception or Error
<pre><code>
The FetchEvent for ".../runtime-es2015.91c3f469886b8efa1fca.js" resulted in a network error response: the promise was rejected.
Promise.then (async)
onFetch @ ngsw-worker.js:1993
(anonymous) @ ngsw-worker.js:1928
</code></pre>

## 🌍  Your Environment

**Angular Version:**
<pre><code>

9.0.4
</code></pre>

**Anything else relevant?**

Might be related to #21412 and #33793 which both appear fixed.

I'm not sure what the right fix here would be, I think fallback to network might be appropriate rather than completely failing the request.

andreialecu:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Failing to write to the cache because of Quota Exceeded results in catastrophic failure.

Issue Number: Fixes #36086

## What is the new behavior?

Don't blow up when writing to the cache.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/service-worker/worker/src/driver.ts
+++ b/packages/service-worker/worker/src/driver.ts
@@ -198,6 +198,8 @@ export class Driver implements Debuggable, UpdateSource {
       // Even though the worker is in safe mode, idle tasks still need to happen so
       // things like update checks, etc. can take place.
       event.waitUntil(this.idle.trigger());
+      // Should still return a response even while in SAFE_MODE
+      event.respondWith(this.handleFetch(event));
       return;
     }
 
@@ -516,8 +518,10 @@ export class Driver implements Debuggable, UpdateSource {
           table.write('assignments', assignments),
           table.write('latest', latest),
         ]);
-      } catch (err) {
-        this.debugger.log(err, `initialize - save initial state`);
+      } catch (error) {
+        this.state = DriverReadyState.SAFE_MODE;
+        this.stateMessage = `Initialization failed due to error: ${errorToString(error)}`;
+        this.debugger.log(error, `initialize - save initial state`);
       }
     }
 
@@ -763,8 +767,10 @@ export class Driver implements Debuggable, UpdateSource {
       // The latest manifest is broken. This means that new clients are at the mercy of the
       // network, but caches continue to be valid for previous versions. This is
       // unfortunate but unavoidable.
-      this.state = DriverReadyState.EXISTING_CLIENTS_ONLY;
-      this.stateMessage = `Degraded due to: ${errorToString(err)}`;
+      if (this.state === DriverReadyState.NORMAL) {
+        this.state = DriverReadyState.EXISTING_CLIENTS_ONLY;
+        this.stateMessage = `Degraded due to: ${errorToString(err)}`;
+      }
 
       // Cancel the binding for the affected clients.
       affectedClients.forEach(clientId => this.clientVersionMap.delete(clientId));
@@ -841,8 +847,11 @@ export class Driver implements Debuggable, UpdateSource {
     } catch (err) {
       this.debugger.log(err, `Error occurred while updating to manifest ${hash}`);
 
-      this.state = DriverReadyState.EXISTING_CLIENTS_ONLY;
-      this.stateMessage = `Degraded due to failed initialization: ${errorToString(err)}`;
+      // ensure we only update the state if previously in the NORMAL state
+      if (this.state === DriverReadyState.NORMAL) {
+        this.state = DriverReadyState.EXISTING_CLIENTS_ONLY;
+        this.stateMessage = `Degraded due to failed initialization: ${errorToString(err)}`;
+      }
 
       return false;
     }
@@ -879,8 +888,10 @@ export class Driver implements Debuggable, UpdateSource {
 
   private async safeSync(): Promise<void> {
     try {
-      return this.sync();
+      return await this.sync();
     } catch (err) {
+      this.state = DriverReadyState.SAFE_MODE;
+      this.stateMessage = `Sync failed due to error: ${errorToString(err)}`;
       this.debugger.log(err, `Driver.sync()`);
     }
   }
--- a/packages/service-worker/worker/src/driver.ts
+++ b/packages/service-worker/worker/src/driver.ts
@@ -384,7 +384,7 @@ export class Driver implements Debuggable, UpdateSource {
 
     // Set the current version used by the client, and sync the mapping to disk.
     this.clientVersionMap.set(client.id, this.latestHash !);
-    await this.sync();
+    await this.safeSync();
 
     // Notify the client about this activation.
     const current = this.versions.get(this.latestHash !) !;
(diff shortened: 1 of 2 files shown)


gkalpak (changes requested):
Thx for taking this on, @andreialecu (and sorry for my slow turn-around 😅).
I have left some commnents, but all in all I think the current approach is a little too aggressive.

I have looked through all calls of [CacheTable#write()][1] throughout the `service-worker` code and tried to trace them up to were they are handled and/or can affect the observed behavior of the SW.

Details below (mostly for personal reference - not in a particularly useful format :grin:), but the main finding is that the only failure that needs to be handled is:

https://github.com/angular/angular/blob/c8bef1259c185fab24ef483ed626bc27ba61d775/packages/service-worker/worker/src/driver.ts#L419-L421

For example:

```diff
     // Decide which version of the app to use to serve this request. This is asynchronous as in
     // some cases, a record will need to be written to disk about the assignment that is made.
-    const appVersion = await this.assignVersion(event);
+    let appVersion = null;
+    try {
+      appVersion = await this.assignVersion(event);
+    } catch (err) {
+      this.debugger.log(err, `Failed to assign an app version: Driver.fetch(${event.request.url})`);
+    }
```

WDYT?
Also, can you try that the above change does indeed solve the problem for you?

##
<details>
<summary><code>CacheTable#write()</code> calls</summary>

- In the list below, its indentation level lists the call-sites of the method in the previous indentation level.
- An error thrown in a method is propagated to the caller/rethrown, unless otherwise mentioned after the method (separated with ` - `). E.g. `SomeClass#someMethod() - Log it` means that the error thrown inside `SomeClass`' `someMethod()` is logged (i.e. suppressed).
- A method surrounded by `[...]` is one that no further analysis is needed for. This can be either because it appears elsewhere in the list (so it is already analyzed) or because it is handled in an adequate way. 

```
CacheTable#write()
  AssetGroup#fetchAndCacheOnce() - throw SwCriticalError
    IdleScheduler#schedule() callback
      IdleScheduler#execute() - DebugHandler#log()
        IdleScheduler#trigger()
          [Driver#... - never awaited]
    AssetGroup#handleFetch()
      AppVersion#handleFetch()
        [Driver#handleFetch() (with SwCriticalError) - Driver#versionFailed() (NORMAL => EXISTING_CLIENTS_ONLY) + Driver#safeFetch()]
    [PrefetchAssetGroup#initializeFully()]
    LazyAssetGroup#initializeFully()
      [AppVersion#initializeFully() - AppVersion#_okay = false + rethrow]
  AssetGroup#maybeUpdate()
    [PrefetchAssetGroup#initializeFully()]
    [LazyAssetGroup#initializeFully()]
  PrefetchAssetGroup#initializeFully()
    AppVersion#initializeFully() - AppVersion#_okay = false + rethrow
      [IdleScheduler#schedule callback - DebugHandler#log() + Driver#versionFailed()]
      [Driver#setupUpdate()]
  [DataGroup#syncLru() - DebugHandler#log()]
  DataGroup#cacheResponse()
    [DataGroup#safeCacheResponse() - DebugHandler#log()]
  Driver#initialize()
    Driver#ensureInitialized() - SAFE_MODE + rethrow
      [Driver#handleFetch() - Driver#safeFetch()]
  Driver#sync()
    Driver#assignVersion()
      Driver#handleFetch()
        Driver#onFetch() - FetchEvent#respondWith(...)  <-- Unhandled failure
    Driver#cleanupCaches()
      Driver#initialize()
        [IdleScheduler#schedule callback - DebugHandler#log()]
    Driver#setupUpdate()
      Driver#checkForUpdate() - DebugHandler#log() + EXISTING_CLIENTS_ONLY
        Driver#handleFetch()
          [IdleScheduler#schedule() callback - never awaited]
        [Driver#handleMessage()]
        Driver#initialize()
          [IdleScheduler#schedule() callback - never awaited]
    Driver#updateClient()
      [Driver#assignVersion()]
      Driver#handleMessage()
        [Driver#repostStatus()]
    [Driver#versionFailed() - DebugHandler#log()]
```

</details>

[1]: https://github.com/angular/angular/blob/c8bef1259c185fab24ef483ed626bc27ba61d775/packages/service-worker/worker/src/db-cache.ts#L67

andreialecu:
Even with the changes in this PR I think something is still missing.

I've made a build of the app that was initially affected by #36086 using this service-worker PR. 

On a 2019 generation LG OLED TV the service worker still completely breaks the app on the second load. First load is fine, but on reload nothing loads any more. It shows a "This page is not available" generic browser error. The LG TV runs Chromium 53.

I haven't been able to debug the TV to see what the actual error is, but it seems similar to `ERR_FAILED` that I've seen before even on desktop chrome.

andreialecu:
@gkalpak since I'm already here, would it make sense to add the polyfills from #20756 as part of this PR? Reminder of your own comment here: https://github.com/angular/angular/issues/20756#issuecomment-386315924

The problem is that some devices are stuck on older Chromium versions and there's no way to update them. This is valid for Smart TVs and also other Smart TV Box type devices and kiosks.

Ref: https://github.com/jakearchibald/async-waituntil-polyfill

gkalpak:
I wish we could implement option (b) (from https://github.com/angular/angular/issues/20756#issuecomment-386315924), but it didn't get traction :grin:

I am a little concerned about monkey-patching such basic APIs (i.e. `waitUntil()` and `respondWith()`) for everyone (versus just for those needing it). IS there a way to feature-detect this?

In any case, feel free to create a separate PR for that fix and we can discuss there.
(I'd rather keep this on a separate PR, since it is an unrelated issue and potentially controversial too :grin:)

gkalpak:
Hey, @andreialecu, would you have time/interest to move this forward? (Or should someone else take over?)

andreialecu:
I could, but maybe it's easier for you to take it over yourself if you have time yourself?

Allow edits is on. The polyfill turns out is more important for our use case. As mentioned previously, all LG TV's before 2020 need it. They cannot be updated so nothing can be done about getting angular+service-worker not completely blow up on them. I'd be interested in a solution for that as well, because kiosk/informational apps are broken otherwise (other TVs/boxes have the same issue)

andreialecu:
Chromium version for webOS releases, for reference:
http://webostv.developer.lge.com/discover/specifications/web-engine/

andreialecu:
For Samsung TVs, 2018 and older won't work either: 
https://developer.samsung.com/smarttv/develop/specifications/web-engine-specifications.html

PR #36205, opened 2020-03-23
fix(benchpress): update dependencies

wagnermaciel:
* updated, added, and removed dependencies in package.json
* added dependencies to BUILD.bazel

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Why are we making these changes?
The `package.json` and `BUILD.bazel` for `benchpress` are both outdated.
* `package.json` is using a very old version of `@angular/core`, defines dependencies that are no longer used, and is lacking some dependencies.
* `BUILD.bazel` is lacking some dependencies in the `benchpress` `ts_library` which is causing bazel to throw errors.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/benchpress/package.json
+++ b/packages/benchpress/package.json
@@ -6,11 +6,12 @@
   "typings": "./index.d.ts",
   "strictNullChecks": true,
   "dependencies": {
-    "@angular/core": "^2.0.0-rc.7",
+    "@angular/core": "^9.0.0",
+    "@types/node": "^12.11.1",
+    "@types/q": "^1.5.2",
+    "protractor": "^5.4.2",
+    "q": "^1.5.1",
     "reflect-metadata": "^0.1.2",
-    "rxjs": "^6.5.3",
-    "jpm": "1.1.4",
-    "firefox-profile": "0.4.0",
     "selenium-webdriver": "^2.53.3"
   },
   "repository": {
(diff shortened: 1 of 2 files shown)


packages/benchpress/package.json
```diff
@@ -6,11 +6,12 @@
   "typings": "./index.d.ts",
   "strictNullChecks": true,
   "dependencies": {
-    "@angular/core": "^2.0.0-rc.7",
+    "@angular/core": "~9.0.6",
```
josephperrott: Is there a reason we need to `~9.0.6` and not something like `^9.0.0`?

wagnermaciel: I just copied what the rest of angular/angular seemed to be using

wagnermaciel: Is `^9.0.0` what this should be using?

josephperrott: Well for most of our packages we actually have a dependency pinned to the same version, but I don't know that it makes sense here.

My assumption/guess is that `^9.0.0` is probably our best bet since we don't have anything in benchpress that relies on anything brought in after the `9.0.0` release.

wagnermaciel: Ok, I'll fix this :+1:

josephperrott (approved):
LGTM

jelbourn (approved):
LGTM

PR #36207, opened 2020-03-23
build: ensure that refs and shas for PRs only need to be requested once

josephperrott:


.circleci/env.sh
```diff
@@ -27,6 +34,9 @@ setPublicVar CI_PULL_REQUEST "${CIRCLE_PR_NUMBER:-false}";
 setPublicVar CI_REPO_NAME "$CIRCLE_PROJECT_REPONAME";
 setPublicVar CI_REPO_OWNER "$CIRCLE_PROJECT_USERNAME";
 
+# Store a PR's refs and shas so they don't need to be requested multiple times.
+setPublicVar GITHUB_REFS_AND_SHAS $(node tools/utils/get-refs-and-shas-for-target.js 36207 | awk '{ gsub(/"/,"\\\"") } 1');
```
devversion: Should this be using the actual CI pull request number?

josephperrott: Yes, it requests the refs and SHAs for the specific PR via the Github API.  We actually already do this in a number of places in our CI run, this just will do it once, and then load it from the environment if its already defined.

Long term we need to find a way to get rid of this request entirely.

josephperrott: Now I see what you were saying, a copy paste error.  Thank you.

tools/utils/get-refs-and-shas-for-target.js
```diff
@@ -16,7 +16,7 @@ const exec = util.promisify(child_process.exec);
 
 async function requestDataFromGithub(url) {
   // GitHub requires a user agent: https://developer.github.com/v3/#user-agent-required
-  const options = {headers: {'User-Agent': 'angular'}};
+  let options = {headers: {'User-Agent': 'angular'}};
```
IgorMinar: shouldn't we also provide a read-only auth token to raise the github throttling limits? it's not secure because anyone can use that token, but if it's read-only then it doesn't really matter as the worst thing that could happen is that someone will eat into our api request quota.

IgorMinar (approved):
lgtm, with a few suggestions.

thanks Joey!

tools/utils/get-refs-and-shas-for-target.js
josephperrott: I initially was attempting to do that, thus the change to `let` here, but it turns out that github doesn't do read-only tokens.

The closest thing we could do is create a github account that we give no permissions to at all, then create an access token for the account.  It would still have write access under its own account, but would not be able to interact with the Angular org as a writer since it doesn't have the necessary permissions.

tools/utils/get-refs-and-shas-for-target.js
```diff
@@ -84,6 +94,19 @@ module.exports = async function getRefsAndShasForTarget(prNumber) {
     latestShaOfTargetBranch: latestShaOfTargetBranch.trim(),
     latestShaOfPrBranch: latestShaOfPrBranch.trim(),
   };
-
   return output;
-};
+}
+
+// If the script is called directly, log the output of the refs and sha for the
```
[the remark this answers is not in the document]
josephperrott: Done

tools/utils/get-refs-and-shas-for-target.js
IgorMinar: yup.. that's exactly what I meant. We already have one or two accounts like that. e.g. @chimney-sweeper

josephperrott: SGTM, Let's do it in a follow-up, if you want to create a jira entry in dev infra we can track it there, or I can create one tomorrow.

.circleci/env.sh
```diff
@@ -3,72 +3,86 @@
...
-setPublicVar COMPONENTS_REPO_TMP_DIR "/tmp/angular-components-repo"
-setPublicVar COMPONENTS_REPO_URL "https://github.com/angular/components.git"
-setPublicVar COMPONENTS_REPO_BRANCH "master"
-# **NOTE**: When updating the commit SHA, also update the cache key in the CircleCI `config.yml`.
-setPublicVar COMPONENTS_REPO_COMMIT "598db096e668aa7e9debd56eedfd127b7a55e371"
+readonly bashEnvCachePath="$projectDir/.circleci/bash_env_cache";
+
+if [ -f $bashEnvCachePath ]; then
+  # Since a bash env cache is present, load this into the $BASH_ENV
+  cat "$bashEnvCachePath" >> $BASH_ENV;
```
gkalpak: Nit: I think it would be helpful to log a message mentioning that `$BASH_ENV` is restored from a cached copy.

gkalpak (approved):
I took me a while to realize how this PR helps prevent computing refs/SHAs multiple times. Could you add some more info in the commit message (i.e. that the computed env vars are stored in `~/ng/.circleci` in the `setup` job and are picked up by subsequent jobs because the `~/ng` is persisted in the workspace)?

tools/utils/get-refs-and-shas-for-target.js
```diff
@@ -6,6 +6,11 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
+// NOTE: When invoked directly via node, this script will take the first positional
+// arguement as to be the PR number, and log out the ref and sha information in its
```
devversion: Nit: 

```suggestion
// argument as to be the PR number, and log out the ref and sha information in its
```

.circleci/env.sh
```diff
@@ -3,72 +3,87 @@
...
+  echo "BASH environment loaded from cached value at $bashEnvCachePath";
+else
```
devversion: Can we have a banner somewhere saying that variables are cached? I'm worried about cases where a new variable is defined that is specific to individual jobs (and not to the workflow). e.g. `CIRCLE_NODE_INDEX`. 

We always advocate to put CI variables into `env.sh` to be CI agnostic, so could be technically something that happens.

devversion (approved):
LGTM. Nice work! A few very minor comments.

PR #36212, opened 2020-03-23
fix(compiler): Microsyntax expressions should allow unbound attributes

kyliau:
According to the microsyntax grammar [1] that Misko wrote, a `keyExp`
could have an empty `expression` (the RHS). Such binding is known as a
literal attribute or unbound attribute. For example, in
```
*ngFor="let item of items; foo; bar"
```
`ngFor`, `ngForFoo`, and `ngForBar` are literal attributes because the desugared form is
```
<ng-template ngFor let-item [ngForOf]="items" ngForFoo ngForBar>
             ^^^^^                            ^^^^^^^^ ^^^^^^^^
             literal attribute / unbound attribute
```
Currently, the parser allows empty expression for `ngFor` but not `ngForFoo` and `ngForBar`.
This PR fixes the bug.

[1]: https://gist.github.com/mhevery/d3530294cff2e4a1b3fe15ff75d08855

### DIFF
--- a/packages/compiler/src/expression_parser/parser.ts
+++ b/packages/compiler/src/expression_parser/parser.ts
@@ -858,6 +858,11 @@ export class _ParseAST {
    */
   private parseDirectiveKeywordBindings(key: TemplateBindingIdentifier): TemplateBinding[] {
     const bindings: TemplateBinding[] = [];
+    if (this.consumeStatementTerminator()) {
+      const sourceSpan = new AbsoluteSourceSpan(key.span.start, this.currentAbsoluteOffset);
+      bindings.push(new ExpressionBinding(sourceSpan, key, null /* value */));
+      return bindings;
+    }
     this.consumeOptionalCharacter(chars.$COLON);  // trackBy: trackByFunction
     const value = this.getDirectiveBoundTarget();
     let spanEnd = this.currentAbsoluteOffset;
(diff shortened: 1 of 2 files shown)


packages/compiler/src/expression_parser/parser.ts
```diff
@@ -858,6 +858,11 @@ export class _ParseAST {
    */
   private parseDirectiveKeywordBindings(key: TemplateBindingIdentifier): TemplateBinding[] {
     const bindings: TemplateBinding[] = [];
+    if (this.consumeStatementTerminator()) {
+      const sourceSpan = new AbsoluteSourceSpan(key.span.start, this.currentAbsoluteOffset);
```
AndrewKushnir: nit: may be add a comment on this case with an example of an expression that would trigger this codepath?

AndrewKushnir (approved):
LGTM, thanks Keen! 👍 

I'd like to ask @alxhub to also have a quick look at these changes (added Alex to the list of reviewers) to make sure we are not missing something.

Thank you.

AndrewKushnir:
Hi @kyliau, we just discussed this PR with Kara and I've added Kara to the list of reviewers to see if we want to introduce this behavior vs having a better error message (added a "blocked" label for now). Thank you.

PR #36226, opened 2020-03-24
build: allow custom module resolution for ts-circular-deps tests

devversion:
Currently the `ts-circular-deps` tool uses a hard-coded module resolver
that only works in the `angular/angular` repository.

If the tool is consumed in other repositories through the shared
dev-infra package, the module resolution won't work, and a few
resolvable imports (usually cross-entry-points) are accidentally skipped.
For each test, the resolution might differ, so tests can now configure their
module resolution in a configuration file.

Note that we intentionally don't rely on tsconfig's for module
resolution as parsing their mappings rather complicates the
circular dependency tool. Additionally, not every test has a
corresponding tsconfig file.

Also, hard-coding mappings to `@angular/*` while accepting a path
to the packages folder would work, but it would mean that the circular
deps tool is no longer self-contained. Rather, and also for better flexibility,
a custom resolver should be specified.

### DIFF
--- a/dev-infra/ts-circular-dependencies/config.ts
+++ b/dev-infra/ts-circular-dependencies/config.ts
@@ -6,11 +6,13 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {isAbsolute, dirname, resolve} from 'path';
+import {dirname, isAbsolute, resolve} from 'path';
+
 import {ModuleResolver} from './analyzer';
 
+
 /** Configuration for a circular dependencies test. */
-export interface TestConfig {
+export interface CircularDependenciesTestConfig {
   /** Base directory used for shortening paths in the golden file. */
   baseDir: string;
   /** Path to the golden file that is used for checking and approving. */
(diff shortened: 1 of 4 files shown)


dev-infra/ts-circular-dependencies/config.ts
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {isAbsolute, dirname, resolve} from 'path';
+import {ModuleResolver} from './analyzer';
+
+/** Configuration for a circular dependencies test. */
+export interface TestConfig {
```
josephperrott: Can we rename this to something noting that its the config for TS Circular Deps

dev-infra/ts-circular-dependencies/index.ts
```diff
@@ -90,17 +71,21 @@ export function main(
     return 1;
   }
 
+  const warningsAmount = analyzer.unresolvedFiles.size + analyzer.unresolvedModules.size;
```
josephperrott: Can we rename this to `warningsCount`?

josephperrott (approved):
LGTM

Long term I would love to move this into a common config location so that we have all of our tooling configuration in one place in the repo.

dev-infra/ts-circular-dependencies/config.ts
devversion: 👍 Done.

dev-infra/ts-circular-dependencies/index.ts
devversion: Yep. Done!

devversion:
@josephperrott I don't think it would make sense to move this stuff to a common config location. Primarily because there _can_ be multiple circular dependency tests in a repository, whereas each test has a different configuration. We can chat about this in one of the meetings.

PR #36232, opened 2020-03-24
build: fix bad pullapprove rule

josephperrott:
The dev-infra pull approve group contained a condition with
the glob having a string concatenation rather than a comma
separated list. This corrects this error, and in another PR
we will correct the verification scripts failure to catch
this.


IgorMinar (approved):

devversion (approved):
👍 for updating the script to catch this. <sup>PullApprove should error too?</sup>

josephperrott:
@devversion yes, this will be done in a followup.

PR #34830, opened 2020-01-17
fix(ivy): ensure that `LOCALE_ID` is set after app initializers

petebacondarwin:
Before ivy it was possible to configure a mutable service value
in an application initializer (by providing an `APP_INITIALIZER`)
that could be read in the provider of `LOCALE_ID`. This is a common
scenario if you wanted to load the locale id asynchronously from
an HTTP request for instance.

When using the ivy, the runtime needs to be told what the current
locale is, which is done by calling the `setLocaleId()` function with
the value injected by the `LOCALE_ID` token. Previously this was
being done before the application initializers were run, which meant
that the `LOCALE_ID` provider was being executed before the
app initializers had a chance to get a new value for it.

Now this initalization of the locale for the ivy runtime is done after the
application initializers have been run.

Closes #34701

PR #36237, opened 2020-03-25
fix(core): run `APP_INITIALIZER`s before accessing `LOCALE_ID` token in Ivy TestBed

### LINKED ISSUE #36230: LOCALE_ID accessed in tests before APP_INITIALIZER runs (with Ivy)
# 🐞 bug report

### Affected Package

The issue is caused by package @angular/core.

### Is this a regression?

Yes, the previous version in which this bug was not present was: ^8.2.0 (not using Ivy)

This also works if @angular/core is not processed by `ngcc` (when the tests use ViewEngine instead of Ivy) - we use jest, which does not call `ngcc` automatically like the default Angular CLI setup with karma, so we can test both scenarios easily in our private project repository.

### Description
There was a pull request (#34830) that ensured all `APP_INITIALIZER` functions are executed before `LOCALE_ID` is first accessed. This works correctly when the application is served (whether the initializer is synchronous or asynchronous).

But when running tests, this is not happening. The `LOCALE_ID` is accessed before `APP_INITIALIZER` functions are executed, thus ending in the locale being undefined.

## 🔬 Minimal Reproduction
https://github.com/IgorAufricht/angular-locale-id-test-repro

The steps are included in the readme, but for clarity, here they are:
* `git clone https://github.com/IgorAufricht/angular-locale-id-test-repro`
* `npm install`
* `ng serve` (application loads correctly, console confirms that `LOCALE_ID = fr-FR`)
* `ng test` (fails, `Error: Expected '' to equal 'fr-FR'`)

**Angular Version:**
<pre><code>
Angular CLI: 9.0.7
Node: 12.13.1
OS: win32 x64

Angular: 9.0.7
... router
Ivy Workspace: Yes

@angular-devkit/architect         0.900.7
@angular-devkit/build-angular     0.900.7
@angular-devkit/build-optimizer   0.900.7
@angular-devkit/build-webpack     0.900.7
@angular-devkit/core              9.0.7
@angular-devkit/schematics        9.0.7
@ngtools/webpack                  9.0.7
@schematics/angular               9.0.7
@schematics/update                0.900.7
</code></pre>

**Anything else relevant?**
Related to #34830.

AndrewKushnir:
Prior to this commit, Ivy TestBed was accessing locale ID before `APP_INITIALIZER` functions were called. This execution order is not consistent with the app bootstrap logic in the `application_ref.ts` file. This commit updates Ivy TestBed execution order to call initializers first (since they might affect `LOCALE_ID` token value) and accessing and setting locale ID after that.

Related changes were performed in the `application_ref.ts` file in PR #34830.

Fixes #36230.

## PR Type

- [x] Bugfix

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/test/test_bed_spec.ts
+++ b/packages/core/test/test_bed_spec.ts
@@ -6,7 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {ChangeDetectorRef, Compiler, Component, Directive, ErrorHandler, Inject, Injectable, InjectionToken, Injector, Input, ModuleWithProviders, NgModule, Optional, Pipe, Type, ViewChild, ɵsetClassMetadata as setClassMetadata, ɵɵdefineComponent as defineComponent, ɵɵdefineInjector as defineInjector, ɵɵdefineNgModule as defineNgModule, ɵɵsetNgModuleScope as setNgModuleScope, ɵɵtext as text} from '@angular/core';
+import {APP_INITIALIZER, ChangeDetectorRef, Compiler, Component, Directive, ErrorHandler, Inject, Injectable, InjectionToken, Injector, Input, LOCALE_ID, ModuleWithProviders, NgModule, Optional, Pipe, Type, ViewChild, ɵsetClassMetadata as setClassMetadata, ɵɵdefineComponent as defineComponent, ɵɵdefineInjector as defineInjector, ɵɵdefineNgModule as defineNgModule, ɵɵsetNgModuleScope as setNgModuleScope, ɵɵtext as text} from '@angular/core';
 import {TestBed, getTestBed} from '@angular/core/testing/src/test_bed';
 import {By} from '@angular/platform-browser';
 import {expect} from '@angular/platform-browser/testing/src/matchers';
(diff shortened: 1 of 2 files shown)


AndrewKushnir:
[Presubmit](http://test/OCL:302785529:BASE:302785561:1585097583202:ffcdc0b8)

petebacondarwin (approved):
This looks good for synchronous initialisers. So this is better than the current situation. Therefore I'm happy to land this for now.

But it is possible for initialisers to be asynchronous by returning a `Promise`. It looks to me like the `TestBed` will currently only work with synchronous initialisers since there is no code to wait for the `ApplicationInitStatus` to be resolved.  The corresponding code in the `PlatformRef` class waits for the status object to resolve the `donePromise` before setting the locale id. See https://github.com/angular/angular/blob/e342ffd855ffeb8af7067b42307ffa320d82177e/packages/core/src/application_ref.ts#L327-L337

AndrewKushnir:
Thanks for review @petebacondarwin.

You are right, TestBed doesn't work with async app initializers in Ivy and View Engine (see `test_bed.ts` for View Engine version):
https://github.com/angular/angular/blob/fc3e5cb6d3ef61aaa13bd9788b9b1c7b97af94fe/packages/core/testing/src/test_bed.ts#L413

Implementing this support may require extending TestBed API and it's outside of the scope of this PR (since the current logic is consistent in Ivy and VE).

AndrewKushnir:
FYI, Global and Blueprint presubmits are successful.

PR #36242, opened 2020-03-25
fix(forms): clarify that `minLength` and `maxLength` should be used w…

sonukapoor:
…ith strings/arrays only

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/aio/content/guide/form-validation.md
+++ b/aio/content/guide/form-validation.md
@@ -100,6 +100,14 @@ Note that:
 * This example adds a few getter methods. In a reactive form, you can always access any form control through the `get` method on its parent group, but sometimes it's useful to define getters as shorthands
 for the template.
 
+<div class="alert is-helpful">
+
+  #### A quick note about `minLength` and `maxLength`
+
+  * The `minLength` and `maxLength` validators are _only_ applicable to string and array types
+  * The `minLength` validator does not check for empty values. Use the `required` validator instead
+
+</div>
 
 If you look at the template for the name input again, it is fairly similar to the template-driven example.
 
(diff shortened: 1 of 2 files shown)


packages/forms/src/validators.ts
```diff
@@ -262,7 +262,8 @@ export class Validators {
    * @description
    * Validator that requires the length of the control's value to be greater than or equal
    * to the provided minimum length. This validator is also provided by default if you use the
-   * the HTML5 `minlength` attribute.
+   * the HTML5 `minlength` attribute. Note that the `minLength` validator only works with
```
AndrewKushnir: I'd propose to rephrase it a bit to mention the `length` property, for example:

```
Note that the `minLength` validator only works for types that have numeric `length` property, such as strings or arrays.

The `minLength` validator logic is also not invoked for values when their `length` property is 0 (for example in case of an empty string or an empty array), to support optional controls. You can use the standard `required` validator if empty values should not be considered valid.
```

This is just a proposal, please feel free to update is as you see fit.

aio/content/guide/form-validation.md
```diff
@@ -100,6 +100,14 @@ Note that:
 * This example adds a few getter methods. In a reactive form, you can always access any form control through the `get` method on its parent group, but sometimes it's useful to define getters as shorthands
 for the template.
 
+<div class="alert is-helpful">
```
AndrewKushnir: I think we can keep the `Built-in validators` generic and move additional information from this section into individual validator descriptions. You already updated them, so may be we can just expand them a little bit?

AndrewKushnir (changes requested):
Thanks for the PR @sonukapoor! I left a couple comments, please have a look when you get a chance.

Also I think that both commit messages should start with `docs(forms):`, since we are updating docs in both of them.

aio/content/guide/form-validation.md
sonukapoor: Closed this PR, in favor of the previous one:

https://github.com/angular/angular/pull/36157

Issue #27369, opened 2018-11-30
AsyncPipe template compilation error if value is Promise|Observable


MikeMatusz:
# 🐞 bug report

### Affected Package
The issue is caused by package @angular/core

### Is this a regression?
No.

### Description
Similar to #22100, but if a value is typed as (Promise<any> | Observable<any>) the template compile fails.  I also like the [comment](https://github.com/angular/angular/issues/22100#issuecomment-364294572) on that issue about accepting non-async values as pass-through similar to the async keyword, though that would technically be a feature request.  That's actually the exact functionality I was trying to implement by using an intermediate pipe when I ran into this issue.

As a workaround, I'm converting Promises to Observables inside my Pipe so the type isn't ambiguous, but it seems like that shouldn't be necessary since AsyncPipe.transform has a signature that should accept either.

## 🔬 Minimal Reproduction
https://stackblitz.com/edit/angular-issue-repro2-kpsdeu

## 🔥 Exception or Error
<pre><code>
Argument of type 'Promise<any> | Observable<any>' is not assignable to parameter of type 'Promise<any>'.
Type 'Observable<any>' is not assignable to type 'Promise<any>'.
Property 'then' is missing in type 'Observable<any>'.
</code></pre>

## 🌍  Your Environment

**Angular Version:**
<pre><code>
Node: 8.11.3
OS: win32 x64
Angular: 6.1.10
... animations, common, compiler, compiler-cli, core, forms
... http, language-service, platform-browser
... platform-browser-dynamic, platform-server, router, upgrade

@angular/cdk: 6.1.0
@angular/material: 6.1.0
@angular-devkit/build-optimizer: 0.0.42
@angular-devkit/core: 0.0.29
@angular-devkit/schematics: 0.0.52
@ngtools/json-schema: 1.1.0
@ngtools/webpack: 1.9.6
@schematics/angular: 0.1.17
typescript: 2.9.2
webpack-bundle-analyzer: 2.9.0
webpack-karma-die-hard: 1.0.4
webpack-merge: 4.1.0
webpack: 3.10.0
</code></pre>

dawidgarus:
From what I see async pipe can accept either Promise or Observable, but not Promise|Observable.
I think this two lines:
https://github.com/angular/angular/blob/7.1.1/packages/common/src/pipes/async_pipe.ts#L89-L90
should be merged into:
```typescript
transform<T>(obj: Observable<T>|Promise<T>|null|undefined): T|null;
```

[...]

Issue #36259, opened 2020-03-26
Case conversion pipes do not compose under strict TS checking


ranma42:
# 🐞 bug report

### Affected Package
`@angular/common`
The issue seems to be related to an incoherence between the implementation and signature of `LowerCasePipe`, `TitleCasePipe`, and `UpperCasePipe`.
https://github.com/angular/angular/blob/f8096d499324cf0961f092944bbaedd05364eea1/packages/common/src/pipes/case_conversion_pipes.ts#L32-L38

### Is this a regression?
No, this works as intended when either
 - templates are not typechecked
 - typescript is not strict

Long ago Angular performed very limited checking of the templates, so it was much harder to get to this issue.

### Description
The `lowercase` (/`titlecase`/`uppercase`) pipe is implemented to handle `null` values (which might come from an `async` or a `date` pipe), but its signature does not reflect this.

## 🔬 Minimal Reproduction

https://github.com/ranma42/async-lowercase

## 🔥 Exception or Error
<pre><code>ERROR in src/app/app.component.html:1:4 - error TS2345: Argument of type 'string | null' is not assignable to parameter of type 'string'.
  Type 'null' is not assignable to type 'string'.

1 {{ v | date | lowercase }}
     ~~~~~~~~

  src/app/app.component.ts:7:16
    7   templateUrl: './app.component.html',
                     ~~~~~~~~~~~~~~~~~~~~~~
    Error occurs in the template of component AppComponent.</code></pre>

## 🌍  Your Environment

**Angular Version:**
<pre><code>Angular CLI: 9.1.0
Node: 12.16.0
OS: darwin x64

Angular: 9.1.0
... animations, cli, common, compiler, compiler-cli, core, forms
... language-service, platform-browser, platform-browser-dynamic
... router
Ivy Workspace: Yes

Package                           Version
-----------------------------------------------------------
@angular-devkit/architect         0.901.0
@angular-devkit/build-angular     0.901.0
@angular-devkit/build-optimizer   0.901.0
@angular-devkit/build-webpack     0.901.0
@angular-devkit/core              9.1.0
@angular-devkit/schematics        9.1.0
@ngtools/webpack                  9.1.0
@schematics/angular               9.1.0
@schematics/update                0.901.0
rxjs                              6.5.4
typescript                        3.8.3
webpack                           4.42.0</code></pre>

JoostK:
Tracking as FW-2032

ranma42:
[...]

PR #36277, opened 2020-03-27
Improve typing of common pipes

ranma42:
## PR Checklist

## PR Type

- [x] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #36259
Several pipes have incomplete or incorrect typing.

## What is the new behavior?
The common pipes should now all have a stricter typing.

- [X] Yes
- [x] No

It depends on what is considered a breaking change: some code that would compile will now trigger a compilation error. On the other hand, the code that is made invalid by this change, such as passing an object to a numeric pipe, was incorrect in the first place (it caused errors at runtime instead of at compile time).

Fixing the fallout should be trivial.

### DIFF
--- a/packages/language-service/src/typescript_host.ts
+++ b/packages/language-service/src/typescript_host.ts
@@ -175,8 +175,15 @@ export class TypeScriptServiceHost implements LanguageServiceHost {
       }
     };
     const programFiles = this.program.getSourceFiles().map(sf => sf.fileName);
-    this.analyzedModules =
-        analyzeNgModules(programFiles, analyzeHost, this.staticSymbolResolver, this.resolver);
+
+    try {
+      this.analyzedModules =
+          analyzeNgModules(programFiles, analyzeHost, this.staticSymbolResolver, this.resolver);
+    } catch (e) {
+      // Analyzing modules may throw; in that case, reuse the old modules.
+      this.error(`Analyzing NgModules failed. ${e}`);
+      return this.analyzedModules;
+    }
 
     // update template references and fileToComponent
     const urlResolver = createOfflineCompileUrlResolver();
--- a/packages/language-service/test/test_utils.ts
+++ b/packages/language-service/test/test_utils.ts
@@ -101,6 +101,7 @@ export class MockTypescriptHost implements ts.LanguageServiceHost {
   private readonly overrideDirectory = new Set<string>();
   private readonly existsCache = new Map<string, boolean>();
   private readonly fileCache = new Map<string, string|undefined>();
+  errors: string[] = [];
 
   constructor(
       private readonly scriptNames: string[],
@@ -397,6 +398,10 @@ export class MockTypescriptHost implements ts.LanguageServiceHost {
       };
     }
     throw new Error(`Failed to find marker '${selector}' in ${fileName}`);
+  }
+
+  error(msg: string) {
+    this.errors.push(msg);
   }
 }
 
--- a/packages/language-service/test/typescript_host_spec.ts
+++ b/packages/language-service/test/typescript_host_spec.ts
@@ -6,6 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
+import * as ngc from '@angular/compiler';
 import * as ts from 'typescript';
 
 import {TypeScriptServiceHost} from '../src/typescript_host';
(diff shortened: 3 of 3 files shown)


ranma42:
@googlebot I signed it!

ranma42:
I just noticed that this likely fixes https://github.com/angular/angular/issues/27369

packages/common/src/pipes/async_pipe.ts
```diff
@@ -84,10 +84,7 @@ export class AsyncPipe implements OnDestroy, PipeTransform {
     }
   }
 
-  transform<T>(obj: null): null;
-  transform<T>(obj: undefined): undefined;
-  transform<T>(obj: Observable<T>|null|undefined): T|null;
-  transform<T>(obj: Promise<T>|null|undefined): T|null;
+  transform<T>(obj: Observable<T>|Promise<T>|null|undefined): T|null;
```
ranma42: I just noticed that this likely fixes #27369

packages/language-service/test/completions_spec.ts
```diff
@@ -479,7 +479,7 @@ describe('completions', () => {
       }
     });
 
-    it('should be able to resolve lowercase', () => {
+    it('should be able to resolve json', () => {
```
ranma42: As pointed out in the commit message, this seems wrong. I need some guidance to fix this properly, as the error message looks pretty generic (and not very readable, sorry):
```
Expected [ Object({ name: 'ng-container', kind: 'angular element', sortText: 'ng-container', replacementSpan: Object({ start: 760, length: 9 }) }), Object({ name: 'ng-content', kind: 'angular element', sortText: 'ng-content', replacementSpan: Object({ start: 760, length: 9 }) }), Object({ name: 'ng-template', kind: 'angular element', sortText: 'ng-template', replacementSpan: Object({ start: 760, length: 9 }) }), Object({ name: 'a', kind: 'html element', sortText: 'a', replacementSpan: Object({ start: 760, length: 9 }) }), Object({ name: 'abbr', kind: 'html element', sortText: 'abbr', replacementSpan: Object({ start: 760, length: 9 }) }), Object({ name: 'acronym', kind: 'html element', sortText: 'acronym', replacementSpan: Object({ start: 760, length: 9 }) }), Object({ name: 'address', kind: 'html element', sortText: 'address', replacementSpan: Object({ start: 760, length: 9 }) }), Object({ name: 'area', kind: 'html element', sortText: 'area', replacementSpan: Object({ start: 760, length: 9 }) ... to contain <jasmine.objectContaining(Object({ name: 'substring', kind: 'method' }))>.
```

I believe this might be related to an issue in resolving the type of the pipe when multiple overloads are available.

ranma42: Yep, the language service does not take argument types into account when determining the type of the result. Apparently it can only:
 - resolve the type for pipes with no overload
 - resolve the type for `async` and `slice`, which are [hardcoded](https://github.com/angular/angular/blob/master/packages/language-service/src/typescript_symbols.ts#L622-L644) (possibly in an incorrect way... it looks like the `async` is hardcoded as if it the pipe never returned `null`)

petebacondarwin:
@ranma42 - please can you rebase and fix the conflicts?

ranma42:
@petebacondarwin sure, will do ASAP.

I would also like to solve the issue I pointed out in https://github.com/angular/angular/pull/36277#discussion_r399341614 (aka the one from the last commit). Should I just replace the test with the `json` one?

packages/language-service/test/completions_spec.ts
petebacondarwin: @kyliau do you have any thoughts here? Should we wait for the new integration with ngtsc?

kyliau: This has been on my radar for a while. I'd love to spend some cycles this week to see if there's anything we can do to improve the logic for picking function signatures. Integration with ngtsc looks like it's gonna take a while.
@ranma42 Could you please rebase? A lot have changed since your last commit, and this test case in particular has been refactored quite a bit.

ranma42: I rebased the branch, which is currently running on the CI
Before merging I would love to improve the last commit, but if there are no short-term plans for completing the integration with ngtsc we might prefer to merge it (and possibly add a "should-fail" test that uses lowercase and that will "unexpectedly" start passing when the integration is completed).

I do not know this test system very well, so I do not know if/how such tests are done. Should I just test the opposite and have a comment explaining that the test is bad but reflects the current state of the codebase? or is there a specific syntax for well-known failures?

ranma42:
Apparently the lint fails for a file that this PR does not touch at all.
I guess something changed in the build system; I will try to look into it in the next few days.

kyliau:
The CI failure seems unrelated to your change. If you don't mind, could you please enable the option for collaborators to push commit to your branch? I'll try to fix it up. The option to give permission is on the right of the page (you'll have to scroll to the top).

ranma42:
`Allow edits by maintainers` is already enabled. Is any other flag needed to allow you to push modifications?

ranma42:
Good news! It looks like the lint failure was fixed in master overnight. Rebasing was sufficient to get rid of it :D

kyliau:
> Good news! It looks like the lint failure was fixed in master overnight. Rebasing was sufficient to get rid of it :D

I'll push a commit to fix the test failure in language service then this PR should be good to go.
I don't think we should change the test because like you mentioned, it'd hide a bug in the language service. I'll disable the test for now and add a todo for myself to improve function signature selection.

kyliau:
UGH, I'm terribly sorry! I messed up the `git push`, and as a result your master branch was reset and the PR was closed. Could you please reopen the PR?

In case things go wrong again I pushed a backup to a branch on my repo:
https://github.com/kyliau/angular/tree/ranma42

packages/language-service/test/completions_spec.ts
kyliau: This is resolved in https://github.com/kyliau/angular/commit/3f64f0b0b3f6e03d38d10e64d175b94cbd41a48f

ranma42:
I pushed a new PR with the same commits (with "Allow edits by maintainers" enabled). If you can fix the latest commit, it's awesome :D otherwise I will try to do it tomorrow (right now it's 2AM in my time zone, so I would not trust me to change the code correctly XD )

Issue #36278, opened 2020-03-27
ngcc crashing on every CI build


paustint:
# 🐞 bug report

### Affected Package

 The issue is caused by package @angular/compiler-cli - specifically `ngcc`

### Is this a regression?

 Yes, the previous version in which this bug was not present was: 9.0 (or maybe 8.2)

### Description

 when ngcc runs on circle-ci it fails every time, but if I SSH into the container and run ngcc it completes successfully.

## 🔬 Minimal Reproduction
https://github.com/paustint/test-ivy-app

Example failing builds:
https://app.circleci.com/pipelines/github/paustint/test-ivy-app/1/workflows/e3719bbe-f02c-40c0-8f6f-40c7e61f09ef/jobs/1

https://app.circleci.com/pipelines/github/paustint/test-ivy-app/1/workflows/e3719bbe-f02c-40c0-8f6f-40c7e61f09ef/jobs/2

**Note** This fails 100% of the time, but fails while compiling seemingly different packages (e.x. it is not consistently failing while compiling any specific package.

## 🔥 Exception or Error
<pre><code>
Debug: Forking unlocker child-process
Debug: Attemping to write lock-file at /home/circleci/app/node_modules/@angular/compiler-cli/ngcc/__ngcc_lock_file__ with PID 192
Debug: Written lock-file at /home/circleci/app/node_modules/@angular/compiler-cli/ngcc/__ngcc_lock_file__ with PID 192
Debug: Running ngcc on ClusterExecutor (using 8 worker processes).
Debug: Analyzing entry-points...
Debug: No manifest found for /home/circleci/app/node_modules so walking the directories for entry-points.
Debug: Starting unlocker at process 199 on behalf of process 192
Debug: The lock-file path is /home/circleci/app/node_modules/@angular/compiler-cli/ngcc/__ngcc_lock_file__
Debug: Walking directories took 0.5s.
Debug: Invalid entry-point /home/circleci/app/node_modules/@angular/common/upgrade. It is missing required dependencies:
 - @angular/upgrade/static
Debug: Invalid entry-point /home/circleci/app/node_modules/@angular/router/upgrade. It is missing required dependencies:
 - @angular/upgrade/static
Debug: Analyzed 51 entry-points in 2.1s. (Total tasks: 255)
Debug: Processing tasks...
Debug: Spawning another worker process as there is more work to be done.
Debug: Spawning another worker process as there is more work to be done.
Debug: Spawning another worker process as there is more work to be done.
[...]

PR #36280, opened 2020-03-27
fix(ngcc): do not spawn more processes than intended in parallel mode

gkalpak:
When running in parallel mode, ngcc spawns multiple worker processed to process the various entry-points. The number of max allowed processes is determined by the number of CPU cores available to the OS. There is also currently an [upper limit of 8][1]. The number of active workers is in turn inferred by the number of [task assignments][2].

In the past, counting the entries of `ClusterMaster#taskAssignments` was enough, because worker processes were spawned eagerly at the beginning and corresponding entries were created in `taskAssignments`. Since #35719 however, worker processes are spawned lazily on an as needed basis. Because there is some delay between [spawning a process][3] and [inserting it][4] into `taskAssignments`, there is a short period of time when `taskAssignment.size` does not actually represent the number of spawned processes. This can result in spawning more than `ClusterMaster#workerCount` processes.

An example of this can be seen in #36278, where the debug logs indicate 9 worker processes had been spawned (`All 9 workers are currently busy`) despite the hard limit of 8.

This commit fixes this by using `cluster.workers` to compute the number of spawned worker processes. `cluster.workers` is updated synchronously with `cluster.fork()` and thus reflects the number of spawned workers accurately at all times.

[1]: https://github.com/angular/angular/blob/b8e9a30d3b6/packages/compiler-cli/ngcc/src/main.ts#L429
[2]: https://github.com/angular/angular/blob/b8e9a30d3b6/packages/compiler-cli/ngcc/src/execution/cluster/master.ts#L108
[3]: https://github.com/angular/angular/blob/b8e9a30d3b6/packages/compiler-cli/ngcc/src/execution/cluster/master.ts#L110
[4]: https://github.com/angular/angular/blob/b8e9a30d3b6/packages/compiler-cli/ngcc/src/execution/cluster/master.ts#L199

### DIFF
--- a/packages/compiler-cli/ngcc/src/execution/cluster/master.ts
+++ b/packages/compiler-cli/ngcc/src/execution/cluster/master.ts
@@ -105,14 +105,14 @@ export class ClusterMaster {
     }
 
     if (!isWorkerAvailable) {
-      if (this.taskAssignments.size < this.workerCount) {
+      const spawnedWorkerCount = Object.keys(cluster.workers).length;
+      if (spawnedWorkerCount < this.workerCount) {
         this.logger.debug('Spawning another worker process as there is more work to be done.');
         cluster.fork();
       } else {
         // If there are no available workers or no available tasks, log (for debugging purposes).
         this.logger.debug(
-            `All ${this.taskAssignments.size} workers are currently busy and cannot take on more ` +
-            'work.');
+            `All ${spawnedWorkerCount} workers are currently busy and cannot take on more work.`);
       }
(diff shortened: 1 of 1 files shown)


petebacondarwin (approved):
LGTM.

Perhaps we should rename `workerCount` to `maxWorkerCount` now?

alxhub (approved):

gkalpak:
> Perhaps we should rename workerCount to maxWorkerCount now?

Too late :grin:

PR #36284, opened 2020-03-27
fix(ngcc): correctly detect imported TypeScript helpers

gkalpak:
The `NgccReflectionHost`'s have logic for detecting certain known declarations (such as `Object.assign()` and TypeScript helpers), which allows the `PartialEvaluator` to evaluate expressions it would not be able to evaluate otherwise.

In #36089, `DelegatingReflectionHost` was introduced, which delegates to a TypeScript `ReflectionptHost` when reflecting on TypeScript files, which for ngcc's case means `.d.ts` files of dependencies. As a result, ngcc lost the ability to detect imported TypeScript helpers from `tslib`, because `DelegatingReflectionHost` was not able to apply the known declaration detection logic while reflecting on `tslib`'s `.d.ts` files.

This commit fixes this by ensuring `DelegatingReflectionHost` calls the `NgccReflectionHost`'s `addKnownDeclaration()` method as necessary, even when using the TypeScript `ReflectionHost`.

##

### DIFF
--- a/packages/compiler-cli/ngcc/src/host/esm2015_host.ts
+++ b/packages/compiler-cli/ngcc/src/host/esm2015_host.ts
@@ -599,10 +599,11 @@ export class Esm2015ReflectionHost extends TypeScriptReflectionHost implements N
    * @param decl The `Declaration` to check.
    * @return The passed in `Declaration` (potentially enhanced with a `KnownDeclaration`).
    */
-  addKnownDeclaration<T extends Declaration|null>(decl: T): T {
-    if (decl !== null && decl.known === null &&
-        // For some reason, TS cannot infer that `decl` is not `null`chere :/
-        this.isJavaScriptObjectDeclaration(decl!)) {
+  detectKnownDeclaration(decl: null): null;
+  detectKnownDeclaration<T extends Declaration>(decl: T): T;
+  detectKnownDeclaration<T extends Declaration>(decl: T|null): T|null;
+  detectKnownDeclaration<T extends Declaration>(decl: T|null): T|null {
+    if (decl !== null && decl.known === null && this.isJavaScriptObjectDeclaration(decl)) {
       // If the identifier resolves to the global JavaScript `Object`, update the declaration to
       // denote it as the known `JsGlobalObject` declaration.
       decl.known = KnownDeclaration.JsGlobalObject;
@@ -620,7 +621,7 @@ export class Esm2015ReflectionHost extends TypeScriptReflectionHost implements N
    */
   protected getDeclarationOfSymbol(symbol: ts.Symbol, originalId: ts.Identifier|null): Declaration
       |null {
-    return this.addKnownDeclaration(super.getDeclarationOfSymbol(symbol, originalId));
+    return this.detectKnownDeclaration(super.getDeclarationOfSymbol(symbol, originalId));
   }
 
   /**
--- a/packages/compiler-cli/ngcc/src/host/esm2015_host.ts
+++ b/packages/compiler-cli/ngcc/src/host/esm2015_host.ts
@@ -8,13 +8,13 @@
 
 import * as ts from 'typescript';
 
-import {ClassDeclaration, ClassMember, ClassMemberKind, ConcreteDeclaration, CtorParameter, Declaration, Decorator, KnownDeclaration, TypeScriptReflectionHost, TypeValueReference, isDecoratorIdentifier, reflectObjectLiteral,} from '../../../src/ngtsc/reflection';
+import {ClassDeclaration, ClassMember, ClassMemberKind, ConcreteDeclaration, CtorParameter, Declaration, Decorator, isDecoratorIdentifier, KnownDeclaration, reflectObjectLiteral, TypeScriptReflectionHost, TypeValueReference,} from '../../../src/ngtsc/reflection';
 import {isWithinPackage} from '../analysis/util';
 import {Logger} from '../logging/logger';
 import {BundleProgram} from '../packages/bundle_program';
 import {findAll, getNameText, hasNameIdentifier, isDefined, stripDollarSuffix} from '../utils';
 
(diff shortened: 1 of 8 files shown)


packages/compiler-cli/ngcc/test/host/commonjs_host_spec.ts
```diff
@@ -45,6 +47,12 @@ runInEachFileSystem(() => {
     let TYPINGS_DTS_FILES: TestFile[];
     let MODULE_WITH_PROVIDERS_PROGRAM: TestFile[];
 
+    // Helpers
+    const createHost = (bundle: BundleProgram, ngccHost: CommonJsReflectionHost) => {
```
JoostK, line 51: Interestingly, I did this exact same thing when starting work on #36089 but it didn't end up in the end result. My reasoning for not going with it originally is that these tests are specific the the bundle format, not any .d.ts interop. However, this PR shows that there _is_ some sort of an interaction between the two, in which case I agree that changing the tests like this is the best way to go.

packages/compiler-cli/ngcc/src/host/ngcc_host.ts
```diff
@@ -128,4 +128,13 @@ export interface NgccReflectionHost extends ReflectionHost {
...
   getEndOfClass(classSymbol: NgccClassSymbol): ts.Node;
+
+  /**
+   * Check whether a `Declaration` corresponds with a known declaration and set its `known` property
+   * to the appropriate `KnownDeclaration`.
+   *
+   * @param decl The `Declaration` to check or `null` if there is no declaration.
+   * @return The passed in `Declaration` (potentially enhanced with a `KnownDeclaration`).
+   */
+  addKnownDeclaration<T extends Declaration|null>(decl: T): T;
```
JoostK, line 139: I am not sold on the `add` part of the name, as my association is that this would record and track the declaration somehow, whereas it's more about the _detection_ of known declarations. So I would suggest something along the line of `detectKnownDeclaration` or `augmentKnownDeclaration`.

JoostK (commented):
Thank you for fixing what I broke! The approach overall looks good, is there a way we can test the last commit in this PR? Or would the refactored tests start to fail without it? If that were the cases, I think we should mention this in the commit message.

packages/compiler-cli/ngcc/src/host/esm2015_host.ts
```diff
@@ -602,8 +592,37 @@ export class Esm2015ReflectionHost extends TypeScriptReflectionHost implements N
...
+  /**
+   * Check whether a `Declaration` corresponds with a known declaration, such as `Object`, and set
+   * its `known` property to the appropriate `KnownDeclaration`.
+   *
+   * @param decl The `Declaration` to check.
+   * @return The passed in `Declaration` (potentially enhanced with a `KnownDeclaration`).
+   */
+  addKnownDeclaration<T extends Declaration|null>(decl: T): T {
+    if (decl !== null && decl.known === null &&
+        // For some reason, TS cannot infer that `decl` is not `null`chere :/
```
petebacondarwin, line 604: If you change the method declaration throughout to this:

``` 
addKnownDeclaration<T extends Declaration>(decl: T|null): T|null;
```

Then TS is able to infer the non-null

petebacondarwin (dismissed):
I am also not keen on the naming of `addKnownDeclaration()` but I would go further and suggest that we make this lazy...  How about we have a method on NgccReflectionHost:

```
getKnownDeclaration(decl: Declaration): ResolvedValue|null;
```

Then in the ES2015 reflectionhost it would look like:

```
getKnownDeclaration(decl: Declaration): ResolvedValue|null {
  if(decl.known !== undefined) {
    return decl.known;
  }
  if (this.isJavaScriptObjectDeclaration(decl)) {
    decl.known = resolveKnownDeclaration(KnownDeclaration.JsGlobalObject);
  } else {
    decl.known = null;
  }
  return decl;
}
```

```
getKnownDeclaration(decl: Declaration): ResolvedValue|null {
  if(decl.known !== undefined) {
    return decl.known;
  }
  const known = super.getKnownDeclaration(decl);
  if (known !== null) {
    const helperDecl = getTsHelperFnFromDeclaration(decl.node);
    decl.known = helperDecl && resolveKnownDeclaration(helperDecl);
  }
  return decl.known;
}
```

Then wherever there is currently a `decl.known` property reference in this PR, we would just call `host.getKnownDeclaration(decl)`.

packages/compiler-cli/ngcc/src/host/esm2015_host.ts
gkalpak: I went with `addKnownDeclaration<T extends Declaration|null>(decl: T): T`, because I wanted TS to be able to infer that when `null` is passed in, `null` will be returned (vs `Declaration | null`).

I now realize your suggestion combined with type overloads will work:

```ts
addKnownDeclaration(decl: null): null;
addKnownDeclaration<T extends Declaration>(decl: T): T;
addKnownDeclaration<T extends Declaration>(decl: T | null): T | null { ... }
```

Thx!

gkalpak:
@JoostK, regarding the tests, yes, the last commit fixes the failing tests that were exposed by the previous commit. I have updated the commit messages to explain that.

JoostK (approved):
LGTM. There's few typos in the commit message of cacc0f532638480ca7cf1642b4b38547d3ec9909:

- Paragraph 2 refers to `ReflectionptHost`, instead of `ReflectionHost`.
- Paragraph 3 refers to `addKnownDeclaration` which was renamed
- Paragraph 4 contains `tests'`, which _I think_ should be without the apostrophe.

<sup>I wish GH had suggestions for commit messages!</sup>

gkalpak:
Thx for the thorough review, @JoostK :100:
I've fixed the typos (and a couple more I noticed - <sub>Note to self: RTFCM before you submit</sub>) except for the `tests'`, which I am pretty sure is also correct (even if an unusual form :grin:).

petebacondarwin (approved):

PR #36286, opened 2020-03-28
refactor(core): use more narrow `QueryList` import to avoid circular deps issue

AndrewKushnir:
Prior to this commit, the `packages/core/src/render3/interfaces/query.ts` file used to import `QueryList` using `../../linker`, which contains a lot of re-exports and as a result, this one import caused a lot of circular deps cycles reported by the tool that checks such deps. In other places in the code the `QueryList` is imported using more narrow import (`linker/query_list`), so this commit uses the same pattern. This change allowed to reduce the number of known cycles from 343 to 207, the golden file was updated accordingly.

This issue was originally found [in a different PR](https://github.com/angular/angular/pull/35568/files#diff-53230a72fded5e37a99eaf2b84393b06R12) while adding an import from the `interfaces/definition.ts`.

## PR Type

- [ ] Yes
- [x] No


AndrewKushnir:
Note: the `lint` CircleCI job should be fixed in #36287. However this PR needs to be rebased once #36287 is merged, since the circlular deps check also happens as a part of the `lint` job.

devversion (approved):
Nice! LGTM

pkozlowski-opensource (approved):
LGTM, I can see that we import from `../linker/xyz` in other parts of the code.

mhevery (approved):

PR #36301, opened 2020-03-29
refactor(language-service): provide service for attribute binding type

ayazhafiz:
This commit refactors the process for determining the type of an Angular
attribute to be use a function that takes an attribute name and returns
the Angular attribute kind and name, rather than requiring the user to
query match the attribute name with the regex and query the matching
array.

This refactor prepares for a future change that will improve the
experience of completing attributes in `()`, `[]`, or `[()]` contexts.

## PR Checklist

## PR Type

- [ ] Yes
- [x] No

### DIFF
--- a/packages/language-service/src/compiler_utils.ts
+++ b/packages/language-service/src/compiler_utils.ts
@@ -1,0 +1,61 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+/**
+ * Matches an Angular attribute to a binding type. See `ATTR` for more details.
+ *
+ * This is adapted from packages/compiler/src/render3/r3_template_transform.ts
+ * to allow empty binding names and match template attributes.
+ */
+const BIND_NAME_REGEXP =
+    /^(?:(?:(?:(bind-)|(let-)|(ref-|#)|(on-)|(bindon-)|(@)|(\*))(.*))|\[\(([^\)]*)\)\]|\[([^\]]*)\]|\(([^\)]*)\))$/;
+/**
+ * Represents possible Angular attribute bindings, as indices on a match of `BIND_NAME_REGEXP`.
+ */
+export enum ATTR {
+  /** "bind-" */
+  KW_BIND = 1,
+  /** "let-" */
+  KW_LET = 2,
+  /** "ref-/#" */
+  KW_REF = 3,
+  /** "on-" */
+  KW_ON = 4,
+  /** "bindon-" */
+  KW_BINDON = 5,
+  /** "@" */
+  KW_AT = 6,
+  /**
+   * "*"
+   * Microsyntax template starts with '*'. See https://angular.io/api/core/TemplateRef
+   */
+  KW_TEMPLATE_ATTR = 7,
+  /** The identifier after "bind-", "let-", "ref-/#", "on-", "bindon-", "@", or "*" */
+  IDENT_KW = 8,
+  /** Identifier inside [()] */
+  IDENT_BANANA_BOX = 9,
+  /** Identifier inside [] */
+  IDENT_PROPERTY = 10,
+  /** Identifier inside () */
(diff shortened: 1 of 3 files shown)


packages/language-service/src/completions.ts
```diff
@@ -10,6 +10,7 @@ import {AST, AbsoluteSourceSpan, AstPath, AttrAst, Attribute, BoundDirectiveProp
 import {$$, $_, isAsciiLetter, isDigit} from '@angular/compiler/src/chars';
 
 import {AstResult} from './common';
+import {ATTR, getBindingKind} from './compiler_utils';
```
kyliau: consider using a more specific filename, like `binding_utils`.

packages/language-service/src/compiler_utils.ts
```diff
@@ -0,0 +1,61 @@
...
+}
+
+/**
+ * Returns an Angular binding kind and name of a given attribute, or undefined if the attribute is
+ * not an Angular attribute.
+ */
+export function getBindingKind(attribute: string): {bindingKind: ATTR, bindingName: string}|
+    undefined {
+  const bindParts = attribute.match(BIND_NAME_REGEXP);
+  if (!bindParts) return undefined;
```
kyliau: `return` is enough, no need to `return undefined`

packages/language-service/src/compiler_utils.ts
```diff
@@ -0,0 +1,61 @@
...
+  IDENT_PROPERTY = 10,
+  /** Identifier inside () */
+  IDENT_EVENT = 11,
+}
+
+/**
+ * Returns an Angular binding kind and name of a given attribute, or undefined if the attribute is
+ * not an Angular attribute.
+ */
+export function getBindingKind(attribute: string): {bindingKind: ATTR, bindingName: string}|
```
kyliau: 1. Consider using a more description function name, since it returns not only `bindingKind`, but also `bindingName`.
Perhaps `getBindingDescriptor`?
2. Please provide a standalone interface for the return type.

packages/language-service/src/compiler_utils.ts
```diff
@@ -0,0 +1,61 @@
...
+    undefined {
+  const bindParts = attribute.match(BIND_NAME_REGEXP);
+  if (!bindParts) return undefined;
+  const bindingKind = bindParts.findIndex((val, i) => i !== 0 && val !== undefined);
```
kyliau: 1. Please handle the case where `findIndex` returns -1
2. Please add a comment on why the first element is skipped. 
3. nit: consider `i > 0`

packages/language-service/src/completions.ts
```diff
@@ -232,31 +204,24 @@ function attributeCompletions(info: AstResult, path: AstPath<HtmlAst>): ng.Compl
...
+    // This is a normal HTML attribute, not an Angular attribute.
     return attributeCompletionsForElement(info, elem.name);
   }
 
+  const {bindingKind} = binding;
   const results: string[] = [];
   const ngAttrs = angularAttributes(info, elem.name);
-  if (!bindParts) {
-    // If bindParts is null then this must be a TemplateRef.
+  if (bindingKind === ATTR.KW_TEMPLATE_ATTR) {
```
kyliau: Do you think it's better to use a switch statement here?

packages/language-service/src/compiler_utils.ts
```diff
@@ -0,0 +1,61 @@
...
+  KW_ON = 4,
+  /** "bindon-" */
+  KW_BINDON = 5,
+  /** "@" */
+  KW_AT = 6,
+  /**
+   * "*"
+   * Microsyntax template starts with '*'. See https://angular.io/api/core/TemplateRef
+   */
+  KW_TEMPLATE_ATTR = 7,
```
kyliau: I think "template attribute" is a very overloaded term, how about we just call it `KW_MICROSYNTAX`? I think that's the least ambiguous term.

kyliau (changes requested):

ayazhafiz:
@kyliau PTAL

kyliau (approved):

Issue #34614, opened 2020-01-02
incorrect component passed to canDeactivate guard 


FDIM:
# 🐞 bug report

### Affected Package

 The issue is caused by package @angular/router

### Description

For some reason CanDeactivate guard gets called with an incorrect component instance.
In this case it should get called with a component that is mounted in a named outlet, but gets called with one from primary outlet. See example in stackblitz.

I'd expect that with configuration like below:
![image](https://user-images.githubusercontent.com/1985429/71674645-0b7a9900-2d7c-11ea-96d6-671a8b7aeb89.png)
The guard should only be called when `AdminHeaderComponent` is being deactivated and only with this component instance.

One notable thing is that the `currentRoute` param is correct!
![image](https://user-images.githubusercontent.com/1985429/71675799-e8051d80-2d7e-11ea-8259-64dbe9015af0.png)

EDIT: found a workaround:
```
canDeactivate(component: HasComponentPendingChanges, currentRoute: ActivatedRouteSnapshot, currentState: RouterStateSnapshot) {
    if (currentRoute.outlet === OVERLAY_OUTLET && this.router['rootContexts'] && this.router['rootContexts'].getContext) {
        const context: OutletContext = this.router['rootContexts'].getContext(OVERLAY_OUTLET);
        if (context && context.outlet && context.outlet.isActivated && context.outlet['activated'].instance) {
            component = context.outlet['activated'].instance;
        }
    }
```

## 🔬 Minimal Reproduction

https://stackblitz.com/edit/angular-named-outlet-yfgsip?file=app%2Fadmin%2Fadmin.module.ts

- Open the link
- Open the console
- Click on `admin` link
- Click on `create user` link
- Observe that following message is printed:
```
deactivating UsersListComponent {}
```
- Now click on `hide navbar` link
![image](https://user-images.githubusercontent.com/1985429/71674811-9065b280-2d7c-11ea-8e51-1b345075419d.png)

- Observe that following messages are printed:
```
deactivating CreateUserComponent {}
deactivating AdminHeaderComponent {}
```
^ all good
- Now the fun begins, click on `admin` and `create user` links again
- Click on `Component 2`
- Observe that following messages are printed:
```
deactivating UsersListComponent {}
deactivating CreateUserComponent {}
deactivating CreateUserComponent {}
```
I'd argue that the last message should state `AdminHeaderComponent`

## 🌍  Your Environment

**Angular Version:**
<pre><code>
8.2.14
</code></pre>

[...]

PR #36302, opened 2020-03-29
fix(router): Pass correct component to canDeactivate checks when using two sibling router-outlets

martinsik:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #34614

There's an edge case where if I use two (or more) sibling `<router-outlet>`s in two (or more) child routes where their parent route doesn't have a component then preactivation will trigger all `canDeactivate` checks with the same component because it will use wrong `OutletContext`.

```
{
 path: 'a',
 children: [
   {
     path: 'b1',
     component: BlankCmp,
     canDeactivate: ['RecordingDeactivate'],
   },
   {
     path: 'b2',
     canDeactivate: ['RecordingDeactivate'],
     component: SimpleCmp,
     outlet: 'aux',
   },
 ],
},
{
  path: 'c',
  component: BlankCmp,
},
```

Navigating from `'/a/(b1//aux:b2)'` to `'/c'` will call both `canDeactivate` guards with `BlankCmp`.

Replication steps are described in #34614.
Updated demo for Angular 9: https://stackblitz.com/edit/angular-named-outlet-enygni?file=app%2Fapp.component.html

## What is the new behavior?

`canDeactivate` check is called with correct component.

- [ ] Yes
- [x] No

## Other information

This fix will pass extra parameter to `deactivateRouteAndItsChildren` that is used in case a route is without a component because there's probably no other way to determine the correct `OutletContext` beforehand. The child routers are iterated inside `deactivateRouteAndItsChildren` where it can access only child `OutletContext`s. However, in this case the router-outlets are siblings.

### DIFF
--- a/packages/router/src/utils/preactivation.ts
+++ b/packages/router/src/utils/preactivation.ts
@@ -10,10 +10,10 @@ import {Injector} from '@angular/core';
 
 import {LoadedRouterConfig, RunGuardsAndResolvers} from '../config';
 import {ChildrenOutletContexts, OutletContext} from '../router_outlet_context';
-import {ActivatedRouteSnapshot, RouterStateSnapshot, equalParamsAndUrlSegments} from '../router_state';
+import {ActivatedRouteSnapshot, equalParamsAndUrlSegments, RouterStateSnapshot} from '../router_state';
 import {equalPath} from '../url_tree';
 import {forEach, shallowEqual} from '../utils/collection';
-import {TreeNode, nodeChildrenAsMap} from '../utils/tree';
+import {nodeChildrenAsMap, TreeNode} from '../utils/tree';
 
 export class CanActivate {
   readonly route: ActivatedRouteSnapshot;
@@ -66,9 +66,8 @@ function getClosestLoadedConfig(snapshot: ActivatedRouteSnapshot): LoadedRouterC
 }
 
 function getChildRouteGuards(
-    futureNode: TreeNode<ActivatedRouteSnapshot>, currNode: TreeNode<ActivatedRouteSnapshot>| null,
-    contexts: ChildrenOutletContexts | null, futurePath: ActivatedRouteSnapshot[],
-    checks: Checks = {
+    futureNode: TreeNode<ActivatedRouteSnapshot>, currNode: TreeNode<ActivatedRouteSnapshot>|null,
+    contexts: ChildrenOutletContexts|null, futurePath: ActivatedRouteSnapshot[], checks: Checks = {
       canDeactivateChecks: [],
       canActivateChecks: []
     }): Checks {
@@ -82,15 +81,16 @@ function getChildRouteGuards(
 
   // Process any children left from the current route (not active for the future route)
   forEach(
-      prevChildren, (v: TreeNode<ActivatedRouteSnapshot>, k: string) =>
-                        deactivateRouteAndItsChildren(v, contexts !.getContext(k), checks));
+      prevChildren,
+      (v: TreeNode<ActivatedRouteSnapshot>, k: string) =>
+          deactivateRouteAndItsChildren(v, contexts!.getContext(k), contexts, checks));
 
   return checks;
 }
 
 function getRouteGuards(
     futureNode: TreeNode<ActivatedRouteSnapshot>, currNode: TreeNode<ActivatedRouteSnapshot>,
-    parentContexts: ChildrenOutletContexts | null, futurePath: ActivatedRouteSnapshot[],
+    parentContexts: ChildrenOutletContexts|null, futurePath: ActivatedRouteSnapshot[],
     checks: Checks = {
       canDeactivateChecks: [],
       canActivateChecks: []
@@ -102,7 +102,7 @@ function getRouteGuards(
   // reusing the node
   if (curr && future.routeConfig === curr.routeConfig) {
     const shouldRun =
-        shouldRunGuardsAndResolvers(curr, future, future.routeConfig !.runGuardsAndResolvers);
+        shouldRunGuardsAndResolvers(curr, future, future.routeConfig!.runGuardsAndResolvers);
     if (shouldRun) {
       checks.canActivateChecks.push(new CanActivate(futurePath));
     } else {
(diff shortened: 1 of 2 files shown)


martinsik:
The failing test is because the bundle size increased but it looks like the expected size was already outdated. [It expects `226144` bytes](https://github.com/angular/angular/blob/master/integration/_payload-limits.json#L52) but when I run the same integration test locally the same bundle has `226631`. With this PR the same bundle has `226659`.

packages/router/src/utils/preactivation.ts
```diff
@@ -172,17 +173,21 @@ function shouldRunGuardsAndResolvers(
...
+    route: TreeNode<ActivatedRouteSnapshot>, context: OutletContext | null,
+    parentContexts: ChildrenOutletContexts, checks: Checks): void {
   const children = nodeChildrenAsMap(route);
   const r = route.value;
 
   forEach(children, (node: TreeNode<ActivatedRouteSnapshot>, childName: string) => {
     if (!r.component) {
-      deactivateRouteAndItsChildren(node, context, checks);
+      deactivateRouteAndItsChildren(
+          node, parentContexts ? parentContexts.getContext(childName) : context, parentContexts,
```
atscott: I'm curious why the type for `parentContexts` doesn't reflect that it could be falsy. It seems like maybe it *should* be `ChildrenOutletContexts | null`. If this is the case, might it make sense to revert your change to the type for `parentContexts` in `getRouteGuards` above as well?

atscott:
[presubmit](https://test.corp.google.com/ui#id=OCL:303769901:BASE:303769920:1585589261595:7ff7eed0)

atscott:
[global presubmit](https://test.corp.google.com/ui#id=OCL:303769901:BASE:303737378:1585602789022:d3bbf01d) passed

atscott:
As for the size failure, those need to be updated every once in a while. Lucky PR authors increase the size just enough to break the 500B threshold and have to update the `_payload-limits.json` :)

packages/router/src/utils/preactivation.ts
martinsik: You're right, it should be `ChildrenOutletContexts | null`.

martinsik:
I had to rebase to master and rerun `yarn gulp format` on `integration.spec.ts` to pass `ci/circleci: lint`. Although, I don't know why it made so many changes.

atscott:
@martinsik - thanks for the update. The formatting changes are expected as clang format was just updated in https://github.com/angular/angular/commit/c5c57f673745e9b66a771c49e778a7d2e8d8c56a. 
Going to run a subset of presubmits as a final sanity check. This PR should get marked for merge later today.

packages/router/src/utils/preactivation.ts
```diff
@@ -66,31 +66,32 @@ function getClosestLoadedConfig(snapshot: ActivatedRouteSnapshot): LoadedRouterC
...
       canDeactivateChecks: [],
       canActivateChecks: []
     }): Checks {
   const prevChildren = nodeChildrenAsMap(currNode);
 
   // Process the children of the future route
   futureNode.children.forEach(c => {
-    getRouteGuards(c, prevChildren[c.value.outlet], contexts, futurePath.concat([c.value]), checks);
+    getRouteGuards(
+        c, prevChildren[c.value.outlet], contexts!, futurePath.concat([c.value]), checks);
```
atscott, line 79: nit: remove `!` (don't forget to run `yarn gulp format` again)

packages/router/src/utils/preactivation.ts
```diff
@@ -66,31 +66,32 @@ function getClosestLoadedConfig(snapshot: ActivatedRouteSnapshot): LoadedRouterC
...
     delete prevChildren[c.value.outlet];
   });
 
   // Process any children left from the current route (not active for the future route)
   forEach(
-      prevChildren, (v: TreeNode<ActivatedRouteSnapshot>, k: string) =>
-                        deactivateRouteAndItsChildren(v, contexts !.getContext(k), checks));
+      prevChildren,
+      (v: TreeNode<ActivatedRouteSnapshot>, k: string) =>
+          deactivateRouteAndItsChildren(v, contexts!.getContext(k), contexts!, checks));
```
atscott, line 87: nit: remove `!`

atscott (commented):
Just two nits to remove null assertion operator now that the type includes `null`

packages/router/src/utils/preactivation.ts
martinsik: I'm not sure here. This `contexts !.getContext(k)` was there before my changes and `context` is still defined as `contexts: ChildrenOutletContexts | null`. Should I update the function declaration as well?

martinsik: done

packages/router/src/utils/preactivation.ts
martinsik: done

atscott (approved):

atscott:
FYI - I also just ran the size test locally at master and got 226611 for the size so the increase is only 48 between the PR and master.

kara (approved):
LGTM for size changes (given that it's only 48)

atscott:
@martinsik - friendly ping to rebase. This is otherwise ready for merge.

martinsik:
@atscott Rebased

atscott:
Hi @martinsik - A failure was detected internally so this PR is being rolled back in #36697. I added a test in #36699 to cover the failing use-case.

martinsik:
@atscott Should I open another PR when that test is merged and I fix this PR?

matsko:
Hey @martinsik yes please if you could rebase + update this PR with the test that Andrew provided then we can reland this fix.

martinsik:
#36699

petebacondarwin:
@martinsik - sorry that this PR got stuck. Could you rebase on master to check that it is still working after the new test was added, and to resolve the conflict? Then I think we should be able to land it.

petebacondarwin:
Actually this PR has been resolved via other commits.

PR #36312, opened 2020-03-30
fix(language-service): infer type of elements of array-like objects

### LINKED ISSUE #36191: [language service] support ReadonlyArray and tuple in ngFor
# 🐞 bug report

### Affected Package

 The issue is caused by package @angular/compiler, @angular/language-service

### Is this a regression?

 No
### Description

Only Array are supported in ngFor by current language service.

## 🔬 Minimal Reproduction

```ts
@Component({
  selector: 'my-app',
  template: ` <div *ngFor="let item of items">{{ item }}</div> `, // (variable) item: any
})
export class AppComponent {
  items = [{ name: 'foo' }, { name: 'bar' }] as const;
}
```
If we remove `as const`:

```ts
@Component({
  selector: 'my-app',
  template: ` <div *ngFor="let item of items">{{ item }}</div> `, // (variable) item: { name: string; }
})
export class AppComponent {
  items = [{ name: 'foo' }, { name: 'bar' }];
}
```

If we change it into ReadonlyArray:

```ts
@Component({
  selector: 'my-app',
  template: ` <div *ngFor="let item of items">{{ item }}</div> `, // (variable) item: any
})
export class AppComponent {
  items: ReadonlyArray<{ name: string }> = [{ name: 'foo' }, { name: 'bar' }];
}
```

ayazhafiz:
Currently the language service only provides support for determining the
type of array-like members when the array-like object is an `Array`.
However, there are other kinds of array-like objects, including
`ReadonlyArray`s and `readonly`-property arrays. This commit adds
support for retrieving the element type of arbitrary array-like objects.

Closes #36191

## PR Checklist

## PR Type

- [x] Bugfix

- [ ] Yes
- [x] No

### DIFF
--- a/packages/language-service/src/typescript_symbols.ts
+++ b/packages/language-service/src/typescript_symbols.ts
@@ -127,10 +127,13 @@ class TypeScriptSymbolQuery implements SymbolQuery {
 
   getElementType(type: Symbol): Symbol|undefined {
     if (type instanceof TypeWrapper) {
-      const tSymbol = type.tsType.symbol;
-      const tArgs = type.typeArguments();
-      if (!tSymbol || tSymbol.name !== 'Array' || !tArgs || tArgs.length != 1) return;
-      return tArgs[0];
+      const ty = type.tsType;
+      const tyArgs = type.typeArguments();
+      // The type should be Array-like, like Array<T> or ReadonlyArray<T>.
+      // Since arrays are actually objects in JavaScript (and TypeScript doesn't expose a more
+      // specific type), check if the type is object-like and has one type argument.
+      if (!(ty.flags & ts.TypeFlags.Object) || tyArgs?.length != 1) return;
+      return tyArgs[0];
     }
   }
 
--- a/packages/language-service/src/typescript_symbols.ts
+++ b/packages/language-service/src/typescript_symbols.ts
@@ -129,10 +129,9 @@ class TypeScriptSymbolQuery implements SymbolQuery {
     if (type instanceof TypeWrapper) {
       const ty = type.tsType;
       const tyArgs = type.typeArguments();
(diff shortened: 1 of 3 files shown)


packages/language-service/src/typescript_symbols.ts
```diff
@@ -127,10 +127,13 @@ class TypeScriptSymbolQuery implements SymbolQuery {
...
     if (type instanceof TypeWrapper) {
-      const tSymbol = type.tsType.symbol;
-      const tArgs = type.typeArguments();
-      if (!tSymbol || tSymbol.name !== 'Array' || !tArgs || tArgs.length != 1) return;
-      return tArgs[0];
+      const ty = type.tsType;
+      const tyArgs = type.typeArguments();
+      // The type should be Array-like, like Array<T> or ReadonlyArray<T>.
+      // Since arrays are actually objects in JavaScript (and TypeScript doesn't expose a more
+      // specific type), check if the type is object-like and has one type argument.
```
kyliau: this doesn't check if `type` is actually `Array`?
`ts.TypeChecker` actually has an internal method `isArrayLikeType`, which does the right check.
Do you think it's worthwhile to call the internal method, or maybe reimplement the logic here?

kyliau: https://raw.githubusercontent.com/microsoft/TypeScript/master/src/compiler/checker.ts

Unfortunately the file is too big, and GitHub refuses to show its content so I cannot link to the line number

ayazhafiz: Issue tracking exposing the function here: https://github.com/microsoft/TypeScript/issues/37711. For now let's read into the internal method, because reimplementing the logic is cumbersome (it relies on local variables in the state of the checker)

kyliau (approved):

PR #36324, opened 2020-03-30
feat(dev-infra): add support for new global approvers in pullapprove

josephperrott:
Pullapprove as added a few new features to allow for us to better
execute our expectation for global approvals. We need to allow for
an expectation that our global approver groups are not in the list
of approved groups. Additionally, since approval groups apply to
all files in the repo, the global approval groups also do not have
conditions defined for them, which means pullapprove verification
need to allow for no conditions need to be defined.

### DIFF
--- a/dev-infra/pullapprove/group.ts
+++ b/dev-infra/pullapprove/group.ts
@@ -74,15 +74,19 @@ export class PullApproveGroup {
         } else if (condition.match(CONDITION_TYPES.ATTR_LENGTH)) {
           // Currently a noop as we do not take any action on this condition type.
         } else if (condition.match(CONDITION_TYPES.GLOBAL_APPROVAL)) {
-          // Currently a noop as we don't take any for global approval conditions.
+          // Currently a noop as we don't take any action for global approval conditions.
         } else {
-          const errMessage = `Unrecognized condition found, unable to parse the following condition: \n\n` +
-          `From the [${groupName}] group:\n` +
-          ` - ${condition}` +
-          `\n\n` +
-          `Known condition regexs:\n` + 
-          `${Object.entries(CONDITION_TYPES).map(([k, v]) => `  ${k}  -  ${v}`).join('\n')}` +
-          `\n\n`;
+          const errMessage =
+              `Unrecognized condition found, unable to parse the following condition: \n\n` +
+              `From the [${groupName}] group:\n` +
+              ` - ${condition}` +
+              `\n\n` +
+              `Known condition regexs:\n` +
+              `${Object.entries(CONDITION_TYPES).map(([k, v]) => ` ${k} - $ {
+            v
+          }
+          `).join('\n')}` +
+              `\n\n`;
           console.error(errMessage);
           process.exit(1);
         }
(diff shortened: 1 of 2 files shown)


dev-infra/pullapprove/group.ts
```diff
@@ -48,43 +49,43 @@ export class PullApproveGroup {
...
+            glob: globString,
+            matcher: new Minimatch(globString, {dot: true}),
+            matchedFiles: new Set<string>(),
+          }));
+          this.misconfiguredLines.push(...misconfiguredLines);
+          this.hasMatchers = true;
+        } else if (condition.match(CONDITION_TYPES.ATTR_LENGTH)) {
+          // Currently a noop as we do not take any action on this condition type.
+        } else if (condition.match(CONDITION_TYPES.GLOBAL_APPROVAL)) {
+          // Currently a noop as we don't take any for global approval conditions.
```
gkalpak:
```suggestion
          // Currently a noop as we don't take any action for global approval conditions.
```

dev-infra/pullapprove/group.ts
```diff
@@ -48,43 +49,43 @@ export class PullApproveGroup {
...
+        } else {
+          const errMessage = `Unrecognized condition found, unable to parse the following condition: \n\n` +
+          `From the [${groupName}] group:\n` +
+          ` - ${condition}` +
+          `\n\n` +
+          `Known condition regexs:\n` + 
+          `${Object.entries(CONDITION_TYPES).map(([k, v]) => `  ${k}  -  ${v}`).join('\n')}` +
```
gkalpak: You wish 😛

dev-infra/pullapprove/parse-yaml.ts
```diff
@@ -8,10 +8,10 @@
 import {parse as parseYaml} from 'yaml';
 
 export interface PullApproveGroupConfig {
-  conditions: string;
-  reviewers: {
-    users: string[],
-    teams: string[],
+  conditions?: string;
+  reviewers?: {
```
gkalpak: Can a group really have no reviewers? How so?
I would expect this type to be something like:

```diff
 export interface PullApproveGroupConfig {
   conditions?: string;
-  reviewers?: {
-    users?: string[],
-    teams?: string[],
-  };
+  reviewers: {users: string[], teams?: string[]} | {teams: string[]};
 }
```

dev-infra/pullapprove/group.ts
josephperrott: I'm not quite sure what you are seeing on this one, can you elaborate a little bit?

dev-infra/pullapprove/parse-yaml.ts
josephperrott: I suppose thats true, it will always require the reviewers to be defined.

dev-infra/pullapprove/group.ts
gkalpak: It looks fine to me, but clang-format has a different opinion 🤷‍♂

dev-infra/pullapprove/parse-yaml.ts
```diff
@@ -8,9 +8,12 @@
 import {parse as parseYaml} from 'yaml';
 
 export interface PullApproveGroupConfig {
-  conditions: string;
+  conditions?: string;
   reviewers: {
     users: string[],
+    teams?: string[],
+  }|{
+    users?: string[],
```
gkalpak, line 16: You shouldn't need `users?` here, since the case of both `users` and `teams` being defined is captured in the first type:

```suggestion
```

gkalpak (approved):

PR #36326, opened 2020-03-30
feat(dev-infra): standard CLI commands using yargs

josephperrott:
Creates a standard CLI command for `ng-dev`, and switches to executing from local sources rather than @angular/dev-infra-private builds.

### DIFF
--- a/dev-infra/cli.ts
+++ b/dev-infra/cli.ts
@@ -6,26 +6,17 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
-import {readFileSync} from 'fs';
-import {join} from 'path';
-import {verify} from './pullapprove/verify';
-import {validateCommitMessage} from './commit-message/validate';
-import {getRepoBaseDir} from './utils/config';
+import * as yargs from 'yargs';
+import {tsCircularDependenciesBuilder} from './ts-circular-dependencies/index';
+import {buildPullapproveParser} from './pullapprove/cli';
+import {buildCommitMessageParser} from './commit-message/cli';
 
-const args = process.argv.slice(2);
-
-
-// TODO(josephperrott): Set up proper cli flag/command handling
-switch (args[0]) {
-  case 'pullapprove:verify':
-    verify();
-    break;
-  case 'commit-message:pre-commit-validate':
-    const commitMessage = readFileSync(join(getRepoBaseDir(), '.git/COMMIT_EDITMSG'), 'utf8');
-    if (validateCommitMessage(commitMessage)) {
-      console.info('√  Valid commit message');
-    }
-    break;
-  default:
-    console.info('No commands were matched');
-}
+yargs.scriptName('ng-dev')
+    .demandCommand()
+    .recommendCommands()
+    .command('ts-circular-deps <command>', '', tsCircularDependenciesBuilder)
+    .command('pullapprove <command>', '', buildPullapproveParser)
+    .command('commit-message <command>', '', buildCommitMessageParser)
+    .wrap(120)
+    .strict()
+    .parse();
--- a/dev-infra/commit-message/cli.ts
+++ b/dev-infra/commit-message/cli.ts
@@ -1,0 +1,20 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import * as yargs from 'yargs';
+import {validateFile} from './validate-file';
+
+/** Build the parser for the commit-message commands. */
+export function buildCommitMessageParser(localYargs: yargs.Argv) {
+  return localYargs.help().strict().command(
+      'pre-commit-validate', 'Validate the most recent commit message', {},
+      () => { validateFile('.git/COMMIT_EDITMSG'); });
+}
+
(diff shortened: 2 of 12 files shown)


dev-infra/ts-circular-dependencies/index.ts
```diff
@@ -19,22 +19,31 @@ import {compareGoldens, convertReferenceChainToGolden, Golden} from './golden';
...
-          .option('warnings', {type: 'boolean', description: 'Prints all warnings.'})
-          .argv;
-  const configPath = isAbsolute(configArg) ? configArg : resolve(configArg);
-  const config = loadTestConfig(configPath);
-  const isApprove = command.includes('approve');
-  process.exit(main(isApprove, config, warnings));
+
+export function tsCircularDependenciesBuilder(localYargs: yargs.Argv) {
+  return localYargs.help()
+      .strict()
```
alan-agius4: Nice <3

dev-infra/cli.ts
```diff
@@ -6,26 +6,15 @@
...
-    break;
-  default:
-    console.info('No commands were matched');
-}
+yargs.demandCommand()
+    .recommendCommands()
+    .command('ts-circular-deps <command>', '', tsCircularDependenciesBuilder)
+    .command('pullapprove <command>', '', buildPullapproveParser)
+    .command('commit-message <command>', '', buildCommitMessageParser)
+    .wrap(120)
```
alan-agius4: Maybe enable `strictCommands` which is similar to `strict` but for commands.
https://yargs.js.org/docs/#api-strictcommandsenabledtrue

dev-infra/pullapprove/cli.ts
```diff
@@ -0,0 +1,19 @@
...
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import * as yargs from 'yargs';
+import {verify} from './verify';
+
+/** Build the parser for the pullapprove commands. */
+export function buildPullapproveParser(localYargs: yargs.Argv) {
+  return localYargs.demandCommand().command(
+      'verify', 'Verify the pullapprove config', {}, () => verify());
```
alan-agius4: Can we add `.strict` here too?

dev-infra/commit-message/cli.ts
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import * as yargs from 'yargs';
+import {validateFile} from './validate-file';
+
+/** Build the parser for the commit-message commands. */
+export function buildCommitMessageParser(localYargs: yargs.Argv) {
+  return localYargs.command(
```
alan-agius4: Same

dev-infra/cli.ts
josephperrott: `strictCommands` appears to not be present in the typings.  I think that we are fine to just have it be `strict` and be strict to both flags and commands.

dev-infra/ng-dev
```diff
@@ -0,0 +1,11 @@
...
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+// Loads the ng-dev cli, automatically executing it.
+require('./cli.ts');
```
devversion: Add a comment to this file that explains that this is the repo-specific entry-point? Also did you check how it impacts the pre-commit hook? ts-node usually has a small overhead for transpiling?

package.json
```diff
@@ -37,7 +37,8 @@
     "public-api:update": "node goldens/public-api/manage.js accept",
     "ts-circular-deps": "ts-node dev-infra/ts-circular-dependencies/index.ts --config ./packages/circular-deps-test.conf.js",
     "ts-circular-deps:check": "yarn -s ts-circular-deps check",
-    "ts-circular-deps:approve": "yarn -s ts-circular-deps approve"
+    "ts-circular-deps:approve": "yarn -s ts-circular-deps approve",
+    "ng-dev": "dev-infra/ng-dev"
```
devversion: Can we use `ts-node` specifically here too? On Windows, the shebang cannot be resolved properly. Using `ts-node` here will make `yarn ng-dev` always work.

```
Paul angular (josephperrott/dev-infra-cli) $ yarn ng-dev
yarn run v1.21.1
$ dev-infra/ng-dev
'dev-infra\ng-dev' is not recognized as an internal or external command,
```

devversion (approved):
Nice! 💯 I like that changes are now effective immediately. 

I added a few minor comments, other than that. LGTM.

package.json
josephperrott: I removed the `ng-dev` script from dev-infra.  My testing was that it worked to execute using the shebang whenever you are running it through `yarn`.

I just removed it entirely and we will just call the `cli.ts` file from the `package.json`

dev-infra/ng-dev
josephperrott: I saw ~350ms of added time when I compared this to the current script.

devversion (approved):

alxhub:
@josephperrott this has a merge conflict in package.json - can you rebase/squash the fixup commits?

josephperrott:
@alxhub Rebased and ready to go, thanks!

PR #36327, opened 2020-03-30
[WIP] feat(bazel): add devmode_target, devmode_module, prodmode_target & prodmode_module to ng_module

gregmagolan:
These were added to ts_library in rules_nodejs 1.5.0 and this change updates ng_module to use the same pattern. Repo is updated to no longer use deprecated tsconfig bazelOpts devmodeTargetOverride value.

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [x] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/dev-infra/tsconfig.json
+++ b/dev-infra/tsconfig.json
@@ -1,5 +1,5 @@
 /**
- * Root tsconfig file for use building all Angular packages. Note there is no rootDir
+ * Root tsconfig file for use building all dev-infra-private packages. Note there is no rootDir
  * and therefore any tsconfig in the package directory will need to define its own
  * rootDir.
  */
--- a/packages/core/schematics/migrations/missing-injectable/BUILD.bazel
+++ b/packages/core/schematics/migrations/missing-injectable/BUILD.bazel
@@ -3,6 +3,7 @@ load("//tools:defaults.bzl", "ts_library")
 ts_library(
     name = "missing-injectable",
     srcs = glob(["**/*.ts"]),
+    devmode_target = "es2015",
     tsconfig = "//packages/core/schematics:tsconfig.json",
     visibility = [
         "//packages/core/schematics:__pkg__",
--- a/packages/zone.js/lib/tsconfig.json
+++ b/packages/zone.js/lib/tsconfig.json
@@ -1,5 +1,5 @@
 /**
- * Root tsconfig file for use building all Angular packages. Note there is no rootDir
+ * Root tsconfig file for use building all zone.js packages. Note there is no rootDir
  * and therefore any tsconfig in the package directory will need to define its own
  * rootDir.
  */
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -11,11 +11,6 @@ exports_files([
     "package.json",
 ])
 
-alias(
-    name = "tsconfig.json",
-    actual = "//packages:tsconfig-build.json",
-)
-
 filegroup(
     name = "web_test_bootstrap_scripts",
     # do not sort
--- a/dev-infra/BUILD.bazel
+++ b/dev-infra/BUILD.bazel
@@ -1,12 +1,16 @@
 load("@build_bazel_rules_nodejs//:index.bzl", "pkg_npm")
 load("@npm_bazel_typescript//:index.bzl", "ts_library")
 
+exports_files(["tsconfig.json"])
+
 ts_library(
     name = "cli",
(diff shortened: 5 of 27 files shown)


gregmagolan:
@filipesilva This will unblock you on angular-cli bazel upgrade

@josephperrott @JiaLiPassion I removed the usage of the deprecated tsconfig bazelOpts devmodeTargetOverride value as part of this PR. I also removed the root `//:tsconfig.json` alias which pointed to `//packages:tsconfig-build.json`. That alias is the implicit tsconfig that ts_library & ng_module use and without it we now have to be explicit. Anywhere using defaults.bzl ts_libary & ng_nodule will automatically get the same value but this change teased out the build targets that may have been using that tsconfig unnecessarily. The ones that stood out are the new dev-infra built targets & zone.js. Can you please TAL and comment on what you want separate tsconfig settings for these build targets and if devmode should be default "es2015" or using the "es5" override that targets using defaults.bzl will get?

dev-infra/tsconfig.json
```diff
@@ -0,0 +1,33 @@
+/**
+ * Root tsconfig file for use building all Angular packages. Note there is no rootDir
```
josephperrott: s/all Angular packages/the @angular/dev-infra-private

packages/zone.js/lib/tsconfig.json
```diff
@@ -0,0 +1,33 @@
+/**
+ * Root tsconfig file for use building all Angular packages. Note there is no rootDir
```
josephperrott: s/all Angular packages/zone.js package

josephperrott (commented):
LGTM

dev-infra/tsconfig.json
gregmagolan: 👍

packages/zone.js/lib/tsconfig.json
gregmagolan: 👍

JiaLiPassion:
@gregmagolan, I believe currently `zone.js` use `ts_library` with the `/packages/zone.js/tsconfig.json` , but in my understanding the `target` in the tsconfig.json is not used, the `ts_library` will generate both `es5` and `es2015` javascript files, so with the changes in this PR, `devmode_target = "es5"` will only generate the `es5` output?

dev-infra/tsconfig.json
```diff
@@ -0,0 +1,33 @@
+/**
+ * Root tsconfig file for use building all dev-infra-private packages. Note there is no rootDir
+ * and therefore any tsconfig in the package directory will need to define its own
+ * rootDir.
+ */
+{
+  "compilerOptions": {
```
devversion: I wish we could have kept a single canonical tsconfig in `//:tsconfig` (doesn't need to be an alias). It ensured that most of the build targets have consistent compiler settings. e.g. `strict: true` and we'd avoid duplication.

packages/core/schematics/migrations/missing-injectable/BUILD.bazel
```diff
@@ -3,6 +3,7 @@ load("//tools:defaults.bzl", "ts_library")
 ts_library(
     name = "missing-injectable",
     srcs = glob(["**/*.ts"]),
+    devmode_target = "es2015",
```
devversion: Nit: Did you decide to use `es2015` for devmode schematics since that's how they were built before? I think there would be nothing wrong with them using the default `es5`. Can we change that? That way we don't need to set the attribute on each `ts_library` for migrations?

gregmagolan: Initially I tried to keep the settings the same as they were which is why that ended up there. Was waiting for feedback on this as it is cross cutting PR. I think I'll leave it around until Thursday and we can discuss the best approach in the dev-infra sync.

devversion: Sounds good. I agree that it's best to keep the behavior as is for now. I just wanted to provide let you know as author of the schematics that there was no particular reason why devmode needs to be `es2015`.

petebacondarwin:
@gregmagolan - is this PR still viable? It looks like it has conflicts and hasn't been worked on for 9 months.

AndrewKushnir:
Closing this PR for now, please reopen if needed.

PR #36329, opened 2020-03-30
build: remove typescript 3.6 and 3.7 support

atscott:
BREAKING CHANGE: typescript 3.6 and 3.7 are no longer supported, please
update to typescript 3.8


filipesilva (changes requested):
Is the next master a minor or a major?

If it's a minor, this change raises discoverability problems. A user won't know to update TS when updating Angular. They can only know after they get an error for the AOT compiler. For setups that skip the TS version check, they might not even get that error.

If it's a major, then `ng update` should have a migration for TS. Users are instructed to run `ng update` between majors but not between minors.

atscott:
@IgorMinar can correct me if I'm wrong, but I believe the next release on master is 10.0.0-next.0. Either way, I'm going to split this PR up into two commits and we can drop the second one if it's not the right time for it:

1. updating the angular versions to 9.1 and the typescript versions to 3.8.3
2. dropping the 3.6 and 3.7 support

atscott:
@filipesilva - I've encountered frequent failures on the test_aio/aio_preview jobs due to OOM/unexpected crashes. Could TS 3.8 have more memory pressure that's causing our CI to be unstable?

Some example failures:
https://app.circleci.com/pipelines/github/angular/angular/12455/workflows/ec50f5b4-0651-4348-9be0-0240dfcf77dd/jobs/667485
https://app.circleci.com/pipelines/github/angular/angular/12475/workflows/3c3bcacc-8389-4810-a53b-2d5afc7fc2da/jobs/667753
https://app.circleci.com/pipelines/github/angular/angular/12510/workflows/35da99da-f4f1-48af-837f-58e9a886fcf2/jobs/668391

filipesilva:
@atscott I think @clydin also noticed something similar, with regards to Build Optimizer. Correct me if I'm wrong though.

clydin:
The issue I'm currently investigating relates to a runtime error due to incorrect import eliding.  It appears that TS 3.8 when used within build optimizer is incorrectly marking certain imports as unreferenced.  However, this only appears to occur within the very specific build optimizer usage.
In regards to memory, I haven't noticed any issues.

From the logs, the errors appear to be occurring during the ngcc parallel worker phase (one of them looks to be with a standalone ngcc execution).  Do these errors occur with 9.1 when using TS 3.7?  It looks like ngcc will limit the number of process workers to 8 but that may be too much for that CI instance (4 GB of memory).  CPU counts on CI systems tend to be inaccurate due to the number of underlying virtual CPUs being quite high (16/32/128/etc.) so even though it's a 2 CPU instance the count will most likely not be 2.

atscott:
@clydin - based on my very unscientific testing in #36380, it does seem like the issue happens with the update to 9.1, not the TS 3.8 upgrade: https://app.circleci.com/pipelines/github/angular/angular/12538/workflows/5f75318c-510a-4de9-aa86-8862a0efcc30/jobs/668882

JoostK:
FYI, @gkalpak has been investigating ngcc worker crashes, which have also been observed on ngcc-validation CI since 9.1. There's FW-2008 to track that work.

clydin:
From dealing with similar issues within the CLI's build pipeline, the minimum of 8 workers may be too large in relation to the memory usage profile of the ngcc workers.

gkalpak:
BTW, updating aio to latest Angular/cli/Material is also done in #36145 (which is currently blocked on [FW-2008](https://angular-team.atlassian.net/browse/FW-2008)).

atscott:
Thanks @gkalpak - I'll wait for #36145 to merge and rebase from there.

IgorMinar:
@filipesilva @atscott the master is open for v10 changes now, so this PR is good as long as it's master-only.

filipesilva (approved):

atscott:
Adding blocked label for #36145 (which is subsequently blocked on [FW-2008](https://angular-team.atlassian.net/browse/FW-2008))

aio/package.json
```diff
@@ -87,28 +87,28 @@
...
+    "@angular/service-worker": "9.1.0",
     "@webcomponents/custom-elements": "1.2.1",
     "rxjs": "^6.5.3",
     "tslib": "^1.10.0",
     "zone.js": "~0.10.3"
   },
   "devDependencies": {
-    "@angular-devkit/build-angular": "0.900.1",
+    "@angular-devkit/build-angular": "0.901.0",
     "@angular/cli": "9.0.1",
```
IgorMinar, line 109: @atscott whenever you update `build-angular` you should update `@angular/cli` as well because they are released together and interdependent (not a great user experience if you edit these files by hand, that's why `ng update` is recommended way to update Angular projects even for minor versions)

```suggestion
    "@angular/cli": "9.1.0",
```

aio/scripts/_payload-limits.json
```diff
@@ -12,7 +12,7 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 2987,
-        "main-es2015": 451469,
+        "main-es2015": 452273,
```
IgorMinar, line 15: we regressed by another 700b just by updating the cli and typescript? is this expected? @kara, @filipesilva / @clydin ?

integration/_payload-limits.json
```diff
@@ -49,7 +49,7 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 2289,
-        "main-es2015": 226144,
+        "main-es2015": 221697,
```
IgorMinar, line 52: interesting...

integration/bazel/yarn.lock
```diff
@@ -3,10 +3,10 @@
 
 
 "@angular/animations@file:../../dist/packages-dist/animations":
-  version "9.1.0-rc.0"
+  version "9.1.0-next.2"
```
IgorMinar, line 6: note/fyi (no action required): this is a bit misleading because it's really the local version of the package and the version just happened to be 9.1.0-next.2 when you refreshed the lock file, but once this runs on master it will be whatever the HEAD version currently is.

packages/compiler-cli/package.json
```diff
@@ -26,8 +26,8 @@
   },
   "peerDependencies": {
     "@angular/compiler": "0.0.0-PLACEHOLDER",
-    "tslib": "^1.10.0",
-    "typescript": ">=3.6 <3.9"
+    "tslib": "^1.11.1",
```
IgorMinar, line 29: if you are updating tslib here (which is good imo), then you should also update it in the main `//package.json`. Can you please do that?

IgorMinar (approved):
I have a few suggestions, but the rest looks good to me. 

btw, could you summarize all the changes you made and create a "playbook" for future tsc version removals so that it's easier for others to do them? thanks! // @kara

integration/_payload-limits.json
IgorMinar: do we know why this happened decreased? I don't see a clear candidate change that would explain this.

aio/scripts/_payload-limits.json
filipesilva: I'm looking into the value above (`aio`) instead of this value (`aio-local`) because the increase is more pronounced (around 1500b instead of 700), because it better reflects versions that users use, and because it's way easier to change the version.

It's true that `aio-local` should be reflecting only the size increase from changes between FW master and this PR, but the size comparison only reports if the difference is bigger than either 1% (~4500b here) or 500b, so there's wiggle room here that's hard to account for without doing multiple bazel builds.

As far as AIO is concerned, this PR does four things:
- updates material from `^9.0.0` to `^9.1.0`
- updates framework from `9.0.0` to `9.1.0`
- updates typescript from `~3.7.4` to `~3.8.3`
- updates the cli build system from `0.900.1` to `0.901.0`

To get to the bottom of it, I'll measure size differences by applying the changes above in order. I chose that order because some of the changes depend on each other.

- starting point: 450676
- updating material from `^9.0.0` to `^9.1.0`: 451097 (+421)
- updating framework from `9.0.0` to `9.1.0`: 452128 (+1031)
- updating typescript from `~3.7.4` to `~3.8.3`: 452128 (+0)
- updating the cli build system from `0.900.1` to `0.901.0`: 452129 (+1)

So overall it looks like framework and material got bigger.

aio/package.json
atscott: Good catch. Not sure how this one got missed.

integration/_payload-limits.json
filipesilva: Size changes in this test generally mean more code was moved from the main bundle, to lazy bundles. In this case it doesn't look like the lazy bundle changed though.

Another reason for the change is that there is less interdependent code between lazy bundle and main bundle.

That it changed in the rollup setup but not the CLI setup points towards advanced cross-chunk optimizations that only really happen in rollup.

filipesilva: Although... this comment makes me think we're not actually running `cli-hello-world-lazy` anymore:
https://github.com/angular/angular/blob/ea53b5a4869ea07a6cfef309b727ba6a3d1ff465/integration/BUILD.bazel#L56

packages/compiler-cli/package.json
atscott: Updating tslib in main `package.json` is causing the `zone.js` build to fail: https://app.circleci.com/pipelines/github/angular/angular/12627/workflows/790e9f96-7a62-4341-908f-3eda1e70128d/jobs/669919

atscott:
Hi @filipesilva and @IgorMinar  - PTAL. This PR has been rebased with master since [FW-2008](https://angular-team.atlassian.net/browse/FW-2008) was resolved. I was also having issues with the main `yarn.lock` so @josephperrott recommended updating the keys to v7 in `config.yml`.

gregmagolan:
> Hi @filipesilva and @IgorMinar - PTAL. This PR has been rebased with master since [FW-2008](https://angular-team.atlassian.net/browse/FW-2008) was resolved. I was also having issues with the main `yarn.lock` so @josephperrott recommended updating the keys to v7 in `config.yml`.

Glad that fixed it. Another +1 to remove the fallback cache. Having an incremental node_modules update on yarn.lock changes has broken things a few times now. @alan-agius4 did you have some measurements on fallback cache performance?

IgorMinar (approved):
I'm still good with this change.

LGTM! Thank you!

Reviewed-for: global-approverse

filipesilva:
@alan-agius4 looked into the Angular CLI fallback cache in https://github.com/angular/angular-cli/pull/17533 and removed it because it wasn't worth it.

filipesilva (approved):

gkalpak:
In order to avoid potential CI flakes, this should not be merged before #36145 (which in turn is blocked on getting a released version that includes the fix from #36626).

kara (approved):
LGTM

atscott:
[presubmit](https://test.corp.google.com/ui#id=OCL:304052704:BASE:310000080:1588707429006:d31d1aad)

PR #36361, opened 2020-04-01
build: sort module resolution warnings in ts-circular-deps tool

devversion:
For better overview of modules that cannot be resolved in the
`ts-circular-deps` tool, the warnings are now sorted.

Additionally, an empty line between fixed and new circular dependencies
is now printed. That should slightly help with distinguishing.


josephperrott (approved):
LGTM

PR #36362, opened 2020-04-01
refactor(compiler): create a new root `BindingScope` for each template

### LINKED ISSUE #35180: ngcc is non-deterministic
# 🐞 bug report

### Affected Package

`@angular/*`

### Is this a regression?

Not sure

### Description

I'm finding `ngcc` is outputting non-deterministic results. Internally I've seen the ivy code swap variable names back and forth between runs, I haven't been able to reproduce this in isolation  though. In other cases I'm seeing the exact same code, but with a single line of whitespace at the end which causes a different source-map hash. I've also seen the `*.__ivy_ngcc_bak` files sometimes being left behind, this normally happens alongside the single line of whitespace.

Non-deterministic results can break caching with tools like bazel. In the case of different variable names it could go a step further and produce different resulting .js files and bust browser caches etc.

Essentially has the same consequences as #34635 but seems less common.

## 🔬 Minimal Reproduction

Running the following almost always produces different results between some of the `ngcc` runs.
```
# Init
yarn add @angular/compiler@9.0.0-rc.14 @angular/compiler-cli@9.0.0-rc.14 @angular/cdk@9.0.0-rc.9 tslib typescript

# Initial ngcc
./node_modules/.bin/ngcc
cp -R node_modules/@angular @angular-0-init

# Redo with fresh node_modules but no changes
rm -Rf node_modules/
yarn
./node_modules-first/.bin/ngcc
cp -R node_modules/@angular @angular-1-rerun

# Redo with new package
yarn add @angular/core@9.0.0-rc.14
./node_modules/.bin/ngcc
cp -R node_modules/@angular @angular-2-new-package

# Redo with new package and out of date yarn.lock
yarn add --no-lockfile @angular/material@9.0.0-rc.9
./node_modules/.bin/ngcc
cp -R node_modules/@angular @angular-3-bad-lock

# Redo with updated yarn.lock
yarn
./node_modules/.bin/ngcc
cp -R node_modules/@angular @angular-4-better-lock

# Redo fresh
rm -Rf node_modules yarn.lock
yarn
./node_modules/.bin/ngcc
cp -R node_modules/@angular @angular-5-redo
```

## 🌍  Your Environment

**Angular Version:**
9.0.0-rc.14, pretty sure it was also happening with 9.0.0-rc.10

petebacondarwin:
Previously we had a singleton `ROOT_SCOPE` object, from
which all `BindingScope`s derived. But this caused ngcc to
produce non-deterministic output when running multiple workers
in parallel, since each process had its own `ROOT_SCOPE`.

In reality there is no need for `BindingScope` reference names
to be unique across an entire application (or in the case of ngcc
across all the libraries). Instead we just need uniqueness within
a template.

This commit changes the compiler to create a new root `BindingScope`
each time it compiles a component's template.

Resolves #35180

### DIFF
--- a/packages/compiler/src/render3/view/compiler.ts
+++ b/packages/compiler/src/render3/view/compiler.ts
@@ -176,7 +176,7 @@ export function compileComponentFromMetadata(
 
   const template = meta.template;
   const templateBuilder = new TemplateDefinitionBuilder(
-      constantPool, BindingScope.ROOT_SCOPE, 0, templateTypeName, null, null, templateName,
+      constantPool, BindingScope.createRootScope(), 0, templateTypeName, null, null, templateName,
       directiveMatcher, directivesUsed, meta.pipes, pipesUsed, R3.namespaceHTML,
       meta.relativeContextFilePath, meta.i18nUseExternalIds);
 
--- a/packages/compiler/src/render3/view/template.ts
+++ b/packages/compiler/src/render3/view/template.ts
@@ -1587,13 +1587,8 @@ export class BindingScope implements LocalResolver {
   private map = new Map<string, BindingData>();
   private referenceNameIndex = 0;
   private restoreViewVariable: o.ReadVarExpr|null = null;
-  private static _ROOT_SCOPE: BindingScope;
-
-  static get ROOT_SCOPE(): BindingScope {
-    if (!BindingScope._ROOT_SCOPE) {
-      BindingScope._ROOT_SCOPE = new BindingScope().set(0, '$event', o.variable('$event'));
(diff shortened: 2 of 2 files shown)


petebacondarwin:
I couldn't find any suitable place to add additional tests of this change; and since it has no real change to the behaviour of the generated code I decided it was really just a refactor (or perhaps a build related change).

alxhub (approved):

alxhub:
[Presubmit](http://test/OCL:305524617:BASE:305524667:1586372455655:3b46b35d)

PR #36366, opened 2020-04-01
fix(platform-server): update `xhr2` dependency

### LINKED ISSUE #36358: update @angular/platform-server dependency xhr2 to 0.2.0
@angular/platform-server is still depending on "xhr2": "^0.1.4". This version of xhr2 is using the depreciated "new Buffer()". Please update this dependency to "0.2.0". The only difference between these two versions is the use of "Buffer.from()" instead of "new Buffer()"

alan-agius4:
Previous versions of xhr2 used the depreciated "new Buffer()".

Closes #36358


josephperrott (approved):
LGTM, global since this is all dev-infra stuff

PR #36369, opened 2020-04-01
fix(core): missing-injectable migration should not migrate `@NgModule` classes

### LINKED ISSUE #35700: ng update adds extraneous @Injectable() ref
# 🐞 bug report

Issue #35282 which was closed as non-reproducible appears to be an issue with running ng-update when upgrading from ng8 to ng9. As shown in my comments in that closed ticket, i have a screenshot which definitely proves ng-update added the extraneous @Injectable() reference. Suggest someone review the ng-update code to see why it would do this.

### Affected Package

Issue occurred when using the highest 8.x version of ng-update that is used to upgrade from ng8 to ng9.

### Is this a regression?

n/a as it's a migration issue (i.e. specific to going from ng8 to ng9)

### Description

Please see referenced issue for complete details.

## 🔬 Minimal Reproduction
<!--
Please create and share minimal reproduction of the issue starting with this template: 

I followed the official docs on upgrading from ng8 to ng9 using ng-update script. First I did a minor ng8 upgrade to get to the highest ng8 package level, then i ran ng-update to go to ng9.

## 🔥 Exception or Error
detailed in referenced issue.

## 🌍  Your Environment

**Angular Version:**
(see above)

devversion:
Based on the migration guide, provided classes which don't have
either `@Injectable`, `@Directive`, `@Component` or `@Pipe` need
to be migrated. This does not seem to be correct as provided classes
with an `@NgModule` also have a factory function that can be read by
the R3 injector. It's unclear in which cases the `@NgModule` decorator is
used for provided classes, but this scenario has been reported.

We should either fix this in the migration, or make sure to report
this as unsupported in the Ivy compiler (when checking providers)

Fixes #35700.

### DIFF
--- a/packages/core/schematics/migrations/missing-injectable/transform.ts
+++ b/packages/core/schematics/migrations/missing-injectable/transform.ts
@@ -18,10 +18,14 @@ import {ResolvedDirective, ResolvedNgModule} from './definition_collector';
 import {ProviderLiteral, ProvidersEvaluator} from './providers_evaluator';
 import {UpdateRecorder} from './update_recorder';
 
-
-
-/** Name of decorators which imply that a given class does not need to be migrated. */
-const NO_MIGRATE_DECORATORS = ['Injectable', 'Directive', 'Component', 'Pipe'];
+/**
+ * Name of decorators which imply that a given class does not need to be migrated.
+ *    - `@Injectable`, `@Directive`, `@Component` and `@Pipe` instruct the compiler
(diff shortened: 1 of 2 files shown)


crisbeto (approved):
LGTM

alxhub (approved):

alxhub:
[Presubmit](http://test/OCL:307436024:BASE:307436096:1587406132959:176bf255)

PR #36370, opened 2020-04-01
fix(ngcc): allow ngcc configuration to match pre-release versions of packages

gkalpak:
Ngcc supports providing a project-level configuration to affect how certain dependencies are processed and also has a built-in fallback configuration for some unmaintained packages. Each entry in these configurations could be scoped to specific versions of a package by providing a version range. If no version range is provided for a package, it defaults to `*` (with the intention of matching any version).

Previously, the installed version of a package was tested against the version range using the [semver][1] package's `satisfies()` function with the default options. By default, `satisfies()` does not match pre-releases (see [here][2] for more details on reasoning). While this makes sense when determining what version of a dependency to install (trying to avoid unexpected breaking changes), it is not desired in the case of ngcc.

This commit fixes it by explicitly specifying that pre-release versions should be matched normally.

##

[1]: https://www.npmjs.com/package/semver
[2]: https://github.com/npm/node-semver#prerelease-tags

### DIFF
--- a/packages/compiler-cli/ngcc/test/packages/configuration_spec.ts
+++ b/packages/compiler-cli/ngcc/test/packages/configuration_spec.ts
@@ -214,7 +214,7 @@ runInEachFileSystem(() => {
                  .toEqual({versionRange: '*', entryPoints: {}});
            });
 
-        it('should match pre-release versions of a package', () => {
+        it('should correctly handle pre-release versions and version ranges', () => {
           loadTestFiles([
             {
               name: _Abs('/project-1/ngcc.config.js'),
@@ -226,6 +226,16 @@ runInEachFileSystem(() => {
                         './entry-point-1': {},
                       },
                     },
+                    'package-2@1.0.0-beta.2': {
+                      entryPoints: {
+                        './entry-point-2': {},
+                      },
+                    },
+                    'package-3@>=1.0.0-beta.2': {
+                      entryPoints: {
+                        './entry-point-3': {},
+                      },
+                    },
                   },
                 };
               `,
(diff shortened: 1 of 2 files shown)


petebacondarwin (approved):
Is it worth testing that we do not match a prerelease in the config if the package is not a pre-release, and vice versa?

* if in `ngcc.config.js` we have `package-1@1.0.0-rc.0` what do we get for `configuration.getConfig(_Abs('/project-1/node_modules/package-1'), '1.0.0');`
* if in `ngcc.config.js` we have `package-1@1.0.0` what do we get for `configuration.getConfig(_Abs('/project-1/node_modules/package-1'), '1.0.0-rc.0');`

gkalpak:
Good call :+1: I'll add some more tests.

gkalpak:
Added some more tests. PTAL to make sure the behavior matches your expectations :grin:

packages/compiler-cli/ngcc/test/packages/configuration_spec.ts
```diff
@@ -287,13 +395,13 @@ runInEachFileSystem(() => {
       });
 
       describe('at the default level', () => {
-        const originalDefaultConfig = DEFAULT_NGCC_CONFIG.packages['package-1'];
+        const originalDefaultConfig = JSON.stringify(DEFAULT_NGCC_CONFIG.packages);
```
gkalpak, line 398: Using `JSON.parse(JSON.stringify(DEFAULT_NGCC_CONFIG.packages))` here, somehow resulted in `originalDefaultConfig === DEFAULT_NGCC_CONFIG.packages` 😱

petebacondarwin: No??

petebacondarwin (approved):

PR #36372, opened 2020-04-01
fix(ngcc): correctly identify relative Windows-style import paths

gkalpak:
Previously, `isRelativePath()` assumed paths are *nix-style. This caused Windows-style paths (such as `C:\foo\some-package\some-file.js`) to not be recognized as "relative" imports.

This commit fixes this by using the OS-agnostic `isRooted()` helper and also accounting for both styles of path delimiters: `/` and `\`

##
This is needed for angular/ngcc-validation#1029 (in case you want to run the tests locally on Windows).

### DIFF
--- a/packages/compiler-cli/ngcc/test/utils_spec.ts
+++ b/packages/compiler-cli/ngcc/test/utils_spec.ts
@@ -7,6 +7,8 @@
  */
 
 import * as ts from 'typescript';
+import {absoluteFrom as _abs} from '../../src/ngtsc/file_system';
+import {runInEachFileSystem} from '../../src/ngtsc/file_system/testing';
 import {KnownDeclaration} from '../../src/ngtsc/reflection';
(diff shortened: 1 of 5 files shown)


packages/compiler-cli/ngcc/test/utils_spec.ts
```diff
@@ -7,6 +7,8 @@
  */
 
 import * as ts from 'typescript';
+import {absoluteFrom as _abs} from '../../src/ngtsc/file_system';
+import {runInEachFileSystem} from '../../src/ngtsc/file_system/testing';
```
petebacondarwin, line 11: Ahah!!

petebacondarwin (approved):
Nice catch @gkalpak

JoostK (approved):

kara:
[presubmit](http://test/OCL:305303025:BASE:305303073:1586284399374:707a3f6f)

PR #36382, opened 2020-04-01
docs: update ng-conf announcement to remove livestream link

kara:
Since the livestream for ng-conf is not public this year,
(and is only available to ng-conf attendees), we are
removing the link from the angular.io homepage.

Instead, we are now pointing to the ng-conf homepage for
more information.

### DIFF
--- a/aio/content/marketing/announcements.json
+++ b/aio/content/marketing/announcements.json
@@ -9,9 +9,9 @@
   {
     "startDate": "2020-04-01",
     "endDate": "2020-04-03",
-    "message": "Watch ng-conf live stream <br/>April 1st-3rd, 2020",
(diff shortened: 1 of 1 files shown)


petebacondarwin (approved):

petebacondarwin:
We could also update the logo if we got a new one from the ng-config team @kara ...

kara:
@petebacondarwin Definitely open to that too, but let's update the link while we're waiting. I think the "ng-conf 2020" one will work for the time being.

petebacondarwin:
Definitely agree! The AIO preview looks good.

IgorMinar (approved):
lgtm!

PR #36384, opened 2020-04-01
update pullapprove config to better handle global approvals

josephperrott:
Beginning this this change, global approvals will now require the approver
to include `Reviewed-for: global-approvers`, and a docs global approval
requires `Reviewed-for: global-docs-approvers`.

Historically, providing a review by any member of the global reviewer
groups would automatically be considered a global review.  This change
enforces that global approval should be an intentional, explicit action.

The global-approvers and global-docs-approvers group will not be
requested as reviews by pullapprove.

### DIFF
--- a/.pullapprove.yml
+++ b/.pullapprove.yml
@@ -97,6 +97,12 @@
 
 version: 3
 
+# Meta field that goes unused by PullApprove to allow for defining aliases to be
+# used throughout the config.
+meta:
+  1: &can-be-global-approved "\"global-approvers\" not in groups.approved"
+  2: &can-be-global-docs-approved "\"global-docs-approvers\" not in groups.approved"
+
 # turn on 'draft' support
 # https://docs.pullapprove.com/config/github-api-version/
 # https://developer.github.com/v3/previews/#draft-pull-requests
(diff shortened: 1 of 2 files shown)


.pullapprove.yml
```diff
@@ -1050,28 +1048,26 @@ groups:
...
   fallback:
     conditions:
+      - *can-be-global-approved
       # Groups which are found to have matching conditions are `active`
       # according to PullApprove. If no groups are matched and considered
       # active, we still want to have a review occur.
       - len(groups.active) == 0
     reviewers:
       users:
         - IgorMinar
```
IgorMinar, line 1073: I know that this is unrelated, but could we actually remove me from here (in a separate PR) and have the PR fail because no reviewer has been assigned to it and force us to update the rules to cover the missing paths?

currently some of these reviews come to me and I update the pullrequest, but other times I don't update it because I don't realize that I'm asked to review this because I'm the final fallback. I'd prefer to get myself out of this and let the PR crash and require pullapprove config change (with global approvers still being around for emergency situations)

.pullapprove.yml
```diff
@@ -650,33 +666,29 @@ groups:
...
-        - ~framework-global-approvers
-        - ~framework-global-approvers-for-docs-only-changes
 
 
   # =========================================================
   #  Benchpress
   # =========================================================
   benchpress:
     conditions:
+      - *can-be-global-approved
```
IgorMinar, line 676: no docs-only? is that intentional?

.pullapprove.yml
```diff
@@ -115,11 +121,39 @@ pullapprove_conditions:
 
 
 groups:
+  # =========================================================
+  #  Global Approvers
```
IgorMinar, line 125: can you add docs here on how to approve to satisfy the `reviewed_for` condition? thanks

.pullapprove.yml
```diff
@@ -115,11 +121,39 @@ pullapprove_conditions:
...
+    reviewers:
+      teams:
+        - framework-global-approvers
+    reviews:
+      request: 0
+      required: 1
+      reviewed_for: required
+
+  # =========================================================
+  #  Global Approvers For Docs
```
IgorMinar, line 138: can you add docs here on how to approve to satisfy the `reviewed_for` condition? thanks

IgorMinar (approved):
lgtm with a few suggestions.

IgorMinar:
we should also let all the global approvers know about this change because it affects their workflow when approving changes:

cc: @kara @mhevery @aikidave @kapunahelewong @petebacondarwin @gkalpak

dev-infra/pullapprove/group.ts
```diff
@@ -34,7 +34,7 @@ const CONDITION_TYPES = {
   INCLUDE_GLOBS: /^contains_any_globs/,
   EXCLUDE_GLOBS: /^not contains_any_globs/,
   ATTR_LENGTH: /^len\(.*\)/,
-  GLOBAL_APPROVAL: /^global-(docs-)?approvers not in groups.approved$/,
+  GLOBAL_APPROVAL: /^\"global-(docs-)?approvers\" not in groups.approved$/,
```
gkalpak, line 37: Nit: You don't need to escape `"`:

```suggestion
  GLOBAL_APPROVAL: /^"global-(docs-)?approvers" not in groups.approved$/,
```

gkalpak (approved):
:tada:

There's a typo in the first commit message: `this this` --> `this`

.pullapprove.yml
josephperrott: I was so close to getting them all correct.

.pullapprove.yml
josephperrott: Will do

PR #36390, opened 2020-04-02
build: add commit message scope for migration changes

devversion:
This is a proposal commit that adds a separate scope for migration changes.

The motivation is that migrations aren't necessarily always affecting `@angular/core`
(e.g. the `move-document` migration is affecting platform-browser). Migrations are just
stored in the `@angular/core` package in favor of a canonical location when someone
runs `ng update`. Additionally, it rather seems confusing that migration changes are listed
under `core` in the changelog. Long commit message titles are needed to make clear that
commits do _not_ affect `@angular/core` runtime logic, but just the ng update migrations.

### DIFF
--- a/.ng-dev/config.ts
+++ b/.ng-dev/config.ts
@@ -32,6 +32,7 @@ const commitMessage = {
     'http',
     'language-service',
     'localize',
+    'migrations',
     'ngcc',
     'packaging',
     'platform-browser',
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -236,6 +236,7 @@ There are currently a few exceptions to the "use package name" rule:
 * **docs-infra**: used for docs-app (angular.io) related changes within the /aio directory of the
   repo
 * **dev-infra**: used for dev-infra related changes within the directories /scripts, /tools and /dev-infra
+* **migrations**: used for changes to the `ng update` migrations.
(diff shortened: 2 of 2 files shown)


gkalpak:
OOC, why not use the proper scope (instead of `core`) in the commit messages (even if the changes are in `packages/core/`)?

devversion:
That's a good point. I think it would be still ambiguous if we would use e.g. `platform-browser` if the migration affects the platform-browser package as the changes are made in `packages/core/`.

Additionally it would still require longer commit messages (providing the context of migrations), and in the changelog, migration changes would not be grouped. I guess it would be good to have migrations show up separately in the changelog as they don't affect `core` runtime issues. Similarly to the `ngcc` scope that could be `compiler-cli` too.

devversion:
@IgorMinar Would you be able to provide some feedback on this? It's not urgent though 😄

IgorMinar (approved):
SGTM but we should announce this change and in order to enforce it we will need reviewers to look for the current scope to be applied. Changing the pullapprove settings to create a new group for migrations would make that easier.

devversion:
@IgorMinar Thanks for reviewing! Yeah, we added a separate code owner group for migrations already ([see here](https://github.com/angular/angular/blob/master/.pullapprove.yml#L227-L236)).

Definitely agreed that we should announce this. I'm wondering if we could do better at enforcing these commit scopes, but that might be a good thing to look into from the dev-infra side.

PR #36391, opened 2020-04-02
build: enable `strictTemplates` in AIO

alan-agius4:
Related to https://github.com/angular/angular-cli/pull/17372

This is to verify that all examples in AIO are complainant with `strictTemplates` setting.


gkalpak (approved):
LGTM :+1:

One question: Why limit it to master only, @alan-agius4?

alan-agius4:
@gkalpak, we can target patch as well 🙂

PR #36383, opened 2020-04-01
Update and edit web-worker page

jbogarthyde:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [ ] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [x] Documentation content changes
- [ ] angular.io application / infrastructure changes
- [ ] Other... Please describe:

## What is the current behavior?
Previous attempt got munged when merged (https://github.com/angular/angular/pull/36026)

Issue Number:  https://angular-team.atlassian.net/browse/DOCS-777

## What is the new behavior?
Corrects previously approved content of https://angular.io/guide/web-worker

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

PR #36397, opened 2020-04-02
docs: add missing command in `WebWorker` guide

gkalpak:
The command was accidentally removed in #36383.


jbogarthyde:
This is not needed, the page is now correct - the bad one was merged briefly, but the correction was made with a couple of hours.

jbogarthyde (approved):
Yes, this example should be included.

PR #36418, opened 2020-04-03
fix(ngcc): detect non-emitted, non-imported TypeScript helpers

gkalpak:
When TypeScript downlevels ES2015+ code to ES5, it uses some helpers functions to emulated some ES2015+ features, such as spread syntax. The TypeScript compiler can be configured to emit these helpers into the transpiled code (which is controlled by the `noEmitHelpers` option - false by default). It can also be configured to import these helpers from the `tslib` module (which is controlled by the `importHelpers` option - false by default).

While most of the time the helpers will be either emitted or imported, it is possible that one configures their app to neither emit nor import them. In that case, the helpers could, for example, be made available on the global object. This is what `@nativescript/angular` v9.0.0-next-2019-11-12-155500-01 does. See, for example, [common.js][1].

Ngcc must be able to detect and statically evaluate these helpers. Previously, it was only able to detect emitted or imported helpers.

This commit adds support for detecting these helpers if they are neither emitted nor imported . It does this by checking identifiers for which no declaration (either concrete or inline) can be found against a list of known TypeScript helper function names.

##

[1]: https://unpkg.com/browse/@nativescript/angular@9.0.0-next-2019-11-12-155500-01/common.js

### DIFF
--- a/packages/compiler-cli/ngcc/src/host/esm5_host.ts
+++ b/packages/compiler-cli/ngcc/src/host/esm5_host.ts
@@ -9,7 +9,7 @@
 import * as ts from 'typescript';
 
 import {ClassDeclaration, ClassMember, ClassMemberKind, Declaration, Decorator, FunctionDefinition, isNamedVariableDeclaration, Parameter, reflectObjectLiteral} from '../../../src/ngtsc/reflection';
-import {getNameText, getTsHelperFnFromDeclaration, hasNameIdentifier} from '../utils';
+import {getNameText, getTsHelperFnFromDeclaration, getTsHelperFnFromIdentifier, hasNameIdentifier} from '../utils';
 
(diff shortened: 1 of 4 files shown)


gkalpak:
Marking this as blocked on #36284, since the latter has to go in first.

petebacondarwin:
A couple of typos in the commit message:

`When TypeScript downlevels ES2015+ code to ES5, it uses some helper~~s~~`

`functions to emulate~~d~~ some ES2015+ features, such as spread syntax.`

`emitted nor imported~~ ~~. It does this by checking identifiers for which no`

petebacondarwin (approved):
LGTM save the typos in the commit message

JoostK (approved):
💯

gkalpak:
Thx for the reviews, @petebacondarwin and @JoostK ✨

PR #36420, opened 2020-04-03
docs: add playbook for adding and removing support for a TS version

atscott:


docs/UPDATING_TYPESCRIPT_VERSIONS.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Adding support for a TypeScript version
+
+1.  Update `MAX_TS_VERSION` in `packages/compiler-cli/src/typescript_support.ts`
+2.  Update `peerDependencies` in `packages/bazel/package.json` and
+    `packages/compiler-cli/package.json` to reflect the new TypeScript version
+    requirements.
+3.  Create tests for versions of TS being added, i.e. tests sources in
+    `integration/typings_test_ts39/*` and add a test in
+    `integration/BUILD.bazel`
```
petebacondarwin, line 11: I note in the BUILD.bazel file there are some things like:

```
    "typings_test_ts38": {
        # Special case for `typings_test_ts38` test as we want to pin
        # `typescript` at version 3.8.x for that test and not link to the
        # root @npm//typescript package.
        "pinned_npm_packages": ["typescript"],
    },
```

Should this be mentioned here?

petebacondarwin (approved):

docs/UPDATING_TYPESCRIPT_VERSIONS.md
atscott: You mean the `pinned_npm_packages`? Sure, I can add that. My assumption was that there'd always be one to copy from, but I also realize that this isn't necessarily true if this process isn't followed (like now, we forgot to add the one for TS 3.8)

PR #36422, opened 2020-04-03
fix(compiler): avoid generating i18n attributes in plain form

AndrewKushnir:
Prior to this change, there was a problem while matching template attributes, which mistakenly took i18n attributes (that might be present in attrs array after template ones) into account. This commit updates the logic to avoid template attribute matching logic from entering the i18n section and as a result this also allows generating proper i18n attributes sections instead of keeping these attribute in plain form (with their values) in attribute arrays.

Note: acceptance test that verifies this change is already present in the [i18n_spec.ts](https://github.com/angular/angular/blob/master/packages/core/test/acceptance/i18n_spec.ts#L1393-L1424) file. The test verifies that directives are matched on the actual node (on the `<div>` node in case of `<div *ngIf="exp">` template) and not on the `<ng-template>` generated in case a structural directive is used. The test passes in master currently due to the fact that i18n attributes are generated in plain form in front of the template ones (i.e. `['i18nAttr', '', 4 /* AttributeMarker.Template */, 'ngIf']`), so they are not taken into account. This PR moves them into a special i18n section (`[4 /* AttributeMarker.Template */, 'ngIf', 6 /* AttributeMarker.I18n */, 'i18nAttr']`) where they should be located and also updates the runtime code to account for it, so the same test verifies the new behavior too.

## PR Type

- [x] Bugfix

- [ ] Yes
- [x] No

### DIFF
--- a/packages/compiler/src/render3/view/template.ts
+++ b/packages/compiler/src/render3/view/template.ts
@@ -26,7 +26,7 @@ import {ParseError, ParseSourceSpan} from '../../parse_util';
 import {DomElementSchemaRegistry} from '../../schema/dom_element_schema_registry';
 import {CssSelector, SelectorMatcher} from '../../selector';
 import {BindingParser} from '../../template_parser/binding_parser';
-import {error} from '../../util';
+import {error, partitionArray} from '../../util';
 import * as t from '../r3_ast';
 import {Identifiers as R3} from '../r3_identifiers';
 import {htmlAstToRender3Ast} from '../r3_template_transform';
@@ -36,7 +36,7 @@ import {I18nContext} from './i18n/context';
 import {createGoogleGetMsgStatements} from './i18n/get_msg_utils';
 import {createLocalizeStatements} from './i18n/localize_utils';
 import {I18nMetaVisitor} from './i18n/meta';
-import {assembleBoundTextPlaceholders, assembleI18nBoundString, declareI18nVariable, getTranslationConstPrefix, I18N_ICU_MAPPING_PREFIX, i18nFormatPlaceholderNames, icuFromI18nMessage, isI18nRootNode, isSingleI18nIcu, placeholdersToParams, TRANSLATION_PREFIX, wrapI18nPlaceholder} from './i18n/util';
+import {assembleBoundTextPlaceholders, assembleI18nBoundString, declareI18nVariable, getTranslationConstPrefix, hasI18nMeta, I18N_ICU_MAPPING_PREFIX, i18nFormatPlaceholderNames, icuFromI18nMessage, isI18nRootNode, isSingleI18nIcu, placeholdersToParams, TRANSLATION_PREFIX, wrapI18nPlaceholder} from './i18n/util';
 import {StylingBuilder, StylingInstruction} from './styling_builder';
 import {asLiteral, chainedInstruction, CONTEXT_NAME, getAttrsForDirectiveMatching, getInterpolationArgsLength, IMPLICIT_REFERENCE, invalid, NON_BINDABLE_ATTR, REFERENCE_PREFIX, RENDER_FLAGS, trimTrailingNulls, unsupported} from './util';
 
@@ -848,11 +848,11 @@ export class TemplateDefinitionBuilder implements t.Visitor<void>, LocalResolver
     this.matchDirectives(NG_TEMPLATE_TAG_NAME, template);
 
     // prepare attributes parameter (including attributes used for directive matching)
-    // TODO (FW-1942): exclude i18n attributes from the main attribute list and pass them
-    // as an `i18nAttrs` argument of the `getAttributeExpressions` function below.
+    const [i18nStaticAttrs, staticAttrs] =
+        partitionArray<t.TextAttribute>(template.attributes, hasI18nMeta);
     const attrsExprs: o.Expression[] = this.getAttributeExpressions(
-        template.attributes, template.inputs, template.outputs, undefined, template.templateAttrs,
-        undefined);
+        staticAttrs, template.inputs, template.outputs, undefined /* styles */,
+        template.templateAttrs, i18nStaticAttrs);
     parameters.push(this.addAttrsToConsts(attrsExprs));
 
     // local refs (ex.: <ng-template #foo>)
@@ -896,12 +896,8 @@ export class TemplateDefinitionBuilder implements t.Visitor<void>, LocalResolver
 
     // Only add normal input/output binding instructions on explicit <ng-template> elements.
     if (template.tagName === NG_TEMPLATE_TAG_NAME) {
-      const inputs: t.BoundAttribute[] = [];
-      const i18nAttrs: (t.TextAttribute|t.BoundAttribute)[] =
-          template.attributes.filter(attr => !!attr.i18n);
-
-      template.inputs.forEach(
-          (input: t.BoundAttribute) => (input.i18n ? i18nAttrs : inputs).push(input));
+      const [i18nInputs, inputs] = partitionArray<t.BoundAttribute>(template.inputs, hasI18nMeta);
+      const i18nAttrs = [...i18nStaticAttrs, ...i18nInputs];
 
       // Add i18n attributes that may act as inputs to directives. If such attributes are present,
       // generate `i18nAttributes` instruction. Note: we generate it only for explicit <ng-template>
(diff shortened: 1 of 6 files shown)


AndrewKushnir:
[Initial g3 presubmit](http://test/OCL:304699761:BASE:304699790:1585951180359:88bf1269)

packages/compiler/src/util.ts
```diff
@@ -262,4 +262,18 @@ export function newArray<T>(size: number, value?: T): T[] {
...
+}
+
+/**
+ * Splits a given array into 2 arrays, based on a boolean value returned by the condition function.
+ *
+ * @param arr Input array that should be split
+ * @param conditionFn Condition function that is called for each item in a given array and returns a
+ * boolean value.
+ */
+export function splitArray<T = any>(arr: T[], conditionFn: (value: T) => boolean): [T[], T[]] {
```
JoostK, line 282: May I suggest naming this `partitionArray`, as _split_ suggests to me there is a single position in the array that acts as pivot.

packages/core/src/render3/node_selector_matcher.ts
```diff
@@ -289,6 +289,9 @@ function matchTemplateAttribute(attrs: TAttributes, name: string): number {
   if (i > -1) {
     i++;
     while (i < attrs.length) {
+      // Return in case we checked all template attrs and are switching to the next section in the
+      // attrs array (that starts with a number that represents an attribute marker).
+      if (typeof attrs[i] === 'number') return -1;
```
JoostK, line 295: Consider storing `attrs[i]` in a variable to read it only once

JoostK (commented):
Could this be tested in an acceptance test as well?

AndrewKushnir:
@JoostK thanks for the comments! I've applied the changes based on your suggestions. Regarding an acceptance test: there is an existing test that verifies the behavior (see [i18n_spec.ts](https://github.com/angular/angular/blob/master/packages/core/test/acceptance/i18n_spec.ts#L1393-L1424) file). I've extended PR description to mention that and described why we can rely on it. Thanks for pointing that out!

packages/compiler/src/render3/view/template.ts
```diff
@@ -848,11 +848,11 @@ export class TemplateDefinitionBuilder implements t.Visitor<void>, LocalResolver
     this.matchDirectives(NG_TEMPLATE_TAG_NAME, template);
 
     // prepare attributes parameter (including attributes used for directive matching)
-    // TODO (FW-1942): exclude i18n attributes from the main attribute list and pass them
-    // as an `i18nAttrs` argument of the `getAttributeExpressions` function below.
+    const [i18nStaticAttrs, staticAttrs] =
+        partitionArray<t.TextAttribute>(template.attributes, hasI18nMeta);
```
petebacondarwin, line 852: No need to provide a type parameter here.

packages/compiler/src/render3/view/template.ts
```diff
@@ -896,12 +896,8 @@ export class TemplateDefinitionBuilder implements t.Visitor<void>, LocalResolver
 
     // Only add normal input/output binding instructions on explicit <ng-template> elements.
     if (template.tagName === NG_TEMPLATE_TAG_NAME) {
-      const inputs: t.BoundAttribute[] = [];
-      const i18nAttrs: (t.TextAttribute|t.BoundAttribute)[] =
-          template.attributes.filter(attr => !!attr.i18n);
-
-      template.inputs.forEach(
-          (input: t.BoundAttribute) => (input.i18n ? i18nAttrs : inputs).push(input));
+      const [i18nInputs, inputs] = partitionArray<t.BoundAttribute>(template.inputs, hasI18nMeta);
```
petebacondarwin, line 899: No need to provide a type parameter here.

petebacondarwin (approved):
LGTM! Two small suggestions

JoostK (approved):

packages/compiler/src/render3/view/template.ts
AndrewKushnir: @petebacondarwin if we omit type here, it would not work correctly with `strictFunctionTypes` option, see [sandbox example here](https://www.typescriptlang.org/play/?noImplicitAny=false&strictNullChecks=false&strictPropertyInitialization=false&strictBindCallApply=false&noImplicitThis=false&noImplicitReturns=false&alwaysStrict=false&esModuleInterop=false&declaration=false#code/MYGwhgzhAEBCkFNoG9oF8BQpIwILQQA8AXBAOwBMZ4IlVMMAzAVzOGIEsB7M6AC0gBJCgAoADmABOYALYAuOImgAyFGAD8CsGQCeaAJQKARly4gE2lNAzRb0SQmLNJvYpOYIA3BgZExXSWJoFjZOHmgJQI4wslxJaR0AHgAVaABeaG0dAD4RKUkFZIBtAF0AGmhgHgpo7jIAMTIFEQA3MBAPQv107OgTMwsyQ2gi4vLoMZKUG0qeCCC3ZmI+HULS9JGS71sqsnng9ohVifWM0u3M+IA6RgCAUTBgPhFohBke6btoEV2amMaXqQZN11NBFssdNAFIxDjp9FcxMwIM9XsCLgYLg4nC4RuCVhUYSAjlsfBgsHMgggiRsimQEAB3aC4ET6Cp0xnM-QkjAAegAVHybHyRiIaEhVMgNFpdJ4DKUykLbKKlBKpZkZXKSlMhTzyXsgmANpFODE4gkRFSIBUBBBhPpvLyBYqirh5UzSiUdXr9kYjVITXUzWAkrhcpbrUIKPagA).

petebacondarwin: I thought we agreed that we could do [this](https://www.typescriptlang.org/play/?noImplicitAny=false&strictNullChecks=false&strictPropertyInitialization=false&strictBindCallApply=false&noImplicitThis=false&noImplicitReturns=false&alwaysStrict=false&esModuleInterop=false&declaration=false&ssl=1&ssc=1&pln=28&pc=38#code/MYGwhgzhAEBCkFNoG9oF8BQpIwILQQA8AXBAOwBMZ4IlVMMAzAVzOGIEsB7M6AC0gBJCgAoADmABOYALYAuOImgAyFGAD8CsGQCeaAJQKARly4gE2lNAzRb0SQmLNJvYpOYIA3BgZExXSWJoFjZOHmgJQI4wslxJaR0AHgAVAD4RKUkFZIBtAF0AGmhgHgpo7jIAMTIFRIBpAhJyKmg0kQA3MBAPBTr9aABeVOgTMwsyQ2gc3MLW-LyUG2KeCCC3ZmI+HWz8wam871sSslXgrohtuYWB-cPoTIA6RgCAUTBgPhFohBlB4eQlrYRMcyjFql9SDJ+upoOtNjpoApGOcdPoHmJmBBPt8oXcDHcHE4XFM4VsisiQBcDj4MFgVkEEJS9jkyAgAO7QXAifRFVkcrn6akYAD0ACpRTZRVMRDQkKpkBotLpPAZ8gVJUDZSo1Jp7srVXkFpLhXSTkEwHtIpwYnEEiJGRAigIIMJ9N4ReKNTlcGrOfNjabTkZLVJrRVbWAdPbKU6hBQ3UA).

AndrewKushnir: Thanks @petebacondarwin. Agreed, I will update the code as you proposed. Thank you.

AndrewKushnir:
@petebacondarwin, I've applied the change that you proposed in a fixup commit (https://github.com/angular/angular/pull/36422/commits/71b4541a18e060fcf433f09f1fa5400c82a36dcc). Could you please have a quick look?

AndrewKushnir:
[Presubmit #2](http://test/OCL:306356415:BASE:306356431:1586829494429:457a01da).

packages/compiler/src/util.ts
```diff
@@ -276,11 +276,12 @@ export function newArray<T>(size: number, value?: T): T[] {
  * Partitions a given array into 2 arrays, based on a boolean value returned by the condition
  * function.
  *
- * @param arr Input array that should be split
+ * @param arr Input array that should be partitioned
  * @param conditionFn Condition function that is called for each item in a given array and returns a
  * boolean value.
  */
-export function partitionArray<T = any>(arr: T[], conditionFn: (value: T) => boolean): [T[], T[]] {
+export function partitionArray<T = any>(
```
petebacondarwin, line 283: You can also remove the `= any` as well 😸

petebacondarwin: Also thanks to this change, you can now remove the typings from the places where this function is called.

petebacondarwin (approved):

packages/compiler/src/util.ts
AndrewKushnir: Thanks Pete! The code is now updated (sorry I missed that yesterday!).

AndrewKushnir:
[Presubmit #3](http://test/OCL:306544972:BASE:306544994:1586909446869:5813462d) (after the latest set of changes, to make sure there are no broken targets in g3).

mhevery (approved):

Issue #27966, opened 2019-01-07
[TRACKING] Re-enable source-map-test on CircleCI


devversion:
We disabled the `source-map-test` in #27937 and should re-enable at some point again.

devversion:
Looked into the source-map-test issue, and apparently `rules_typescript` always inlines source maps, so we'd need to update that test and re-enable it.

Probably should be just converted to a Bazel test which handles this as well.

devversion:
Tracking in JIRA with DEV-122

PR #36425, opened 2020-04-03
build: remove source-map-test gulp script

josephperrott:
`source-map-test` is no longer used or referenced in our repository,
additionally it fails to discover source maps as expected anymore
when executed.


IgorMinar (commented):
Hmmm... That's unfortunate. We still need this test and it's not great that we somehow stopped running them. @kara, maybe someone from the FW team could take on fixing the code and moving it to run under Bazel? Sgty?

josephperrott:
@Kara, any thoughts on how we should proceed on this one?

devversion:
A long time ago I created an issue on this and also looked into this. See: https://github.com/angular/angular/issues/27966#issuecomment-452432578

IgorMinar:
Aha!!! @devversion could you please create a JIRA out of #27966 and see if we can get it into the next sprint?

It's concerning that we don't test the source map correctness any more - this could explain some of the issues I've seen in profiles shared with us by our customers.

btw can we please convert this gulp test to a bazel target in the process of fixing it? thanks!

IgorMinar (changes requested):
let's not remove this just yet and instead convert it to a bazel test either in this PR or a separate PR as discussed with @devversion above.

IgorMinar:
hey Joey, I asked @devversion to resurrect this test in https://github.com/angular/angular/pull/37623#issuecomment-646327933. Once this is done, this PR will become obsolete.

devversion:
@IgorMinar Sounds good to me. I've also created a JIRA ticket: DEV-122 for tracking. I imagine that this test should be generic enough so that it can be used in the components repo too.

AndrewKushnir:
FYI, I'm removing the "merge" label for now, since there are conflicts and it looks like there is a feedback that needs to be addressed.

PR #36428, opened 2020-04-04
fix(core): pipe injecting provider from incorrect context when used in a component input

### LINKED ISSUE #36146: In Ivy, pipes uses providers from a host component where ViewEngine did not
# 🐞 bug report

### Affected Package
@angular/core

### Is this a regression?
Yes, works without ivy

### Description
I think it's related to this https://github.com/angular/angular/issues/34227 but i'm not really sure.
What i have is a component let's say `component1` that passes an Input to another component `component2` and `component2`shows it as-is.
The problem is that this input gets transformed by a pipe that retrieves via injector the informations to transform this Input. This informations is stored component-wise. `component2` also provides it, but for his own data.
Unfortunately, the pipe in question, used in the template of `component1` ends up using the dependency declared in `component2`, probably because the execution gets delegated after change detection has been run, therefore the context of the pipe changes

I'm not honestly sure my English is good enough to explain this properly but luckily i have a repro

## 🔬 Minimal Reproduction
repo: https://github.com/EmaGht/ivypipe
ng-run version: https://ng-run.com/edit/32k314W605UdYOdVEPCE
Steps:
git clone https://github.com/EmaGht/ivypipe.git
cd ivypipe
npm install
ng serve
check out the on-screen debug informations

For now i think i can work around it telling the pipe explicitly which context should be used but... meh

crisbeto:
Given a structure of `<parent><child [input]="value | pipe"></child></parent>` the injector that is associated with the pipe will inject as if it's placed inside of the `child` component. The issue comes from the fact that we don't have a `TNode` tied to pipe instances and the current `TNode` points to the `child`.

These changes work around the issue by having a flag that indicates whether we're currently "inside of" a pipe, and if we are, DI is essentially treated as if it has `SkipSelf`.

Fixes #36146.

### DIFF
--- a/packages/core/src/render3/state.ts
+++ b/packages/core/src/render3/state.ts
@@ -121,6 +121,9 @@ interface LFrame {
    * `LView[currentDirectiveIndex]` is directive instance.
    */
   currentDirectiveIndex: number;
+
+  /** Whether a pipe is currently being executed. */
+  isPipe: boolean;
 }
 
 /**
@@ -287,6 +290,14 @@ export function setIsParent(): void {
   instructionState.lFrame.isParent = true;
 }
 
+export function setIsPipe(isPipe: boolean) {
+  instructionState.lFrame.isPipe = isPipe;
+}
+
+export function getIsPipe() {
+  return instructionState.lFrame.isPipe;
+}
+
 export function getContextLView(): LView {
   return instructionState.lFrame.contextLView;
 }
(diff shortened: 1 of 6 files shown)


packages/core/src/render3/state.ts
```diff
@@ -121,6 +121,9 @@ interface LFrame {
    * `LView[currentDirectiveIndex]` is directive instance.
    */
   currentDirectiveIndex: number;
+
+  /** Whether a pipe is currently being executed. */
+  isPipe: boolean;
```
crisbeto, line 126: Note for reviewer: I don't feel very good about adding this to the state. I'll add a note to the agenda for the next sync meeting so we can discuss it.

crisbeto:
Setting back to "in progress" based on the discussions.

crisbeto:
Closing in favor of https://github.com/angular/angular/pull/36512.

PR #36434, opened 2020-04-04
Dev infra benchmarking

wagnermaciel:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] Yes
- [x] No

## Why are we making this change?
We would like to reuse the benchmarking code in `angular/angular` in `angular/components`.


dev-infra/tools/component_benchmark/defaults/BUILD.bazel
```diff
@@ -1,5 +1,10 @@
 package(default_visibility = ["//visibility:public"])
 
+filegroup(
```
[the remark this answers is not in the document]
wagnermaciel: Should I comment each `filegroup`? All of them are doing the same thing - grouping the files so that they are all included in the `pkg_npm`.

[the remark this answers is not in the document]
jelbourn: That would be a good enough comment for me ("Make source files available for distribution via `pkg_npm`")

ayazhafiz (approved):
👍 for language service

jelbourn (approved):
LGTM

dev-infra/BUILD.bazel
```diff
@@ -34,10 +34,24 @@ genrule(
 
 pkg_npm(
     name = "npm_package",
+    srcs = [
+        "BUILD.bazel",
+        "//dev-infra/tools:files",
+    ],
+    substitutions = {
+        "//dev-infra/": "@npm_dev_infra//",
```
devversion, line 42: Can you please add a comment on why we need this? e.g. so that these utilties can be directly accessed from other bazel targets in the `angular/angular` repo?

dev-infra/tmpl-package.json
```diff
@@ -8,14 +8,28 @@
     "ng-dev": "./cli.js",
     "ts-circular-deps": "./ts-circular-dependencies/index.js"
   },
+  "dependencies": {
+    "fs-extra": "<from-root>",
+    "node-uuid": "<from-root>",
+    "reflect-metadata": "<from-root>"
```
devversion, line 14: Do you know where this dependency is used? I couldn't find it. Seems to be already a dep of benchpress?

dev-infra/tmpl-package.json
```diff
@@ -8,14 +8,28 @@
     "ng-dev": "./cli.js",
     "ts-circular-deps": "./ts-circular-dependencies/index.js"
   },
+  "dependencies": {
+    "fs-extra": "<from-root>",
```
devversion, line 12: We could omit this dependency by just using `shelljs` for `mkdir -p`.

dev-infra/tmpl-package.json
```diff
@@ -8,14 +8,28 @@
...
+    "node-uuid": "<from-root>",
+    "reflect-metadata": "<from-root>"
+  },
   "peerDependencies": {
+    "@angular/benchpress": "^0.1.0",
```
devversion, line 17: This also has a dependency on `@bazel/protractor` now, right? I see it used in the start server script.

dev-infra/BUILD.bazel
```diff
@@ -34,10 +34,24 @@ genrule(
...
+        "//packages/benchpress": "@npm//@angular/benchpress",
+        "//packages/bazel/src:esm5.bzl": "@npm_angular_bazel//src:esm5.bzl",
+        "//packages/zone.js/dist:zone.js": "@npm//:node_modules/zone.js/dist/zone.js",
+        "//tools:defaults.bzl": "@npm_angular_bazel//:index.bzl",
```
devversion, line 46: This replacement is kind of magic. Definitely needs a comment. It makes the assumption that only TypeScript stuff is imported from `tools/defaults.bzl`. We could easily regress here and the downstream package will fail, right?

dev-infra/tmpl-package.json
devversion: There actually seem to be more: https://github.com/angular/angular/blob/a04e8a2c0f831dbc098d867b3e9307f51101d719/dev-infra/tools/ng_rollup_bundle/BUILD.bazel#L13-L18.

I'm unsure what our policy should be on managing dependencies for the dev-infra package.

dev-infra/benchmark/driver-utilities/BUILD.bazel
```diff
@@ -0,0 +1,26 @@
...
+        "@npm//@types/selenium-webdriver",
+        "@npm//fs-extra",
+        "@npm//node-uuid",
+        "@npm//protractor",
+        "@npm//selenium-webdriver",
+    ],
+)
+
+# Make source files available for distribution via pkg_npm
+filegroup(
```
devversion, line 23: Is this `filegroup` and the one in `benchmark/BUILD.bazel` needed? They aren't used anywhere from what I can tell.

dev-infra/BUILD.bazel
```diff
@@ -34,10 +34,24 @@ genrule(
 
 pkg_npm(
     name = "npm_package",
+    srcs = [
+        "BUILD.bazel",
+        "//dev-infra/tools:files",
```
devversion, line 39: Are are we all set on the location of these utilities? Everything inside `dev-infra/*` could be already considered tools. 

Out of curiosity: why is `component_benchmark` not part of `dev-infra/benchmark`?

devversion (approved):
LGTM. Nice work! 🎉 A few minor nits.

dev-infra/BUILD.bazel
wagnermaciel: We can change the location of the utilities. I just named it the same as what folder they were under in angular/angular.

I placed `component_benchmark` in `dev-infra/tools` instead of `dev-infra/benchmark` so that it would live with the other macros we're defining, but I can move it.

dev-infra/BUILD.bazel
wagnermaciel: That's a great observation! I'll make the substitutions more verbose to avoid a regression like this

dev-infra/tmpl-package.json
wagnermaciel: Yeah, this is kind of tricky to get right because when I'm testing this (by using it in the angular/components repo) some of these dependencies are already defined, which makes figuring out which ones are missing difficult.

Is there an easy way to track down which dependencies are missing?

dev-infra/BUILD.bazel
wagnermaciel: How does this sound?

"angular/angular should not consume it's own packages, so we use substitutions to replace these in the deployed version of dev-infra."

dev-infra/tmpl-package.json
wagnermaciel: I'm not sure where, but I do remember getting errors about some files missing from `reflect-metadata` and adding this to dev-infra's deps fixed it.

dev-infra/tmpl-package.json
wagnermaciel: We can, but I'd rather we do this in a follow up PR if that's ok with you

dev-infra/BUILD.bazel
devversion: 👍

dev-infra/tmpl-package.json
devversion: It might be worth double-checking. Most of the time we use `core-js` as a better replacement for `reflect-metadata` too.

dev-infra/tmpl-package.json
devversion: Sure, fine by me.

dev-infra/tmpl-package.json
devversion: I just went through the folder and looked for potential imports. I don't think there is anything easy that can help us enforcing this

dev-infra/BUILD.bazel
devversion: I don't mind too much. I'd like to have a policy on where which stuff should live. I could definitely imagine the component benchmark macro to be part of `dev-infra/benchmark`.

dev-infra/benchmark/driver-utilities/BUILD.bazel
wagnermaciel: Nope. I think I left this in on accident when trying to get `npm_package` right. I'll remove it!

dev-infra/BUILD.bazel
wagnermaciel: @jelbourn Would you like to weigh in here. I definitely agree with @devversion that these could be structured better

@josephperrott I know you probably have strong opinions about this as well

josephperrott: My initial thought/reaction is that we should have all the benchmarking specific stuff in dev-infra should be under `//dev-infra/benchmarking/`. Having things be in `//dev-infra/tools` to me implies that it is an end user (of the dev-infra package) tool or a common tool for usage within the dev-infra package.  But to my understanding, all of these things are specific to benchmarking, so they should live there (at least until they are needed by something else).

dev-infra/tmpl-package.json
wagnermaciel: Just checked and didn't get the errors anymore. Even tried running `bazel clean --expunge`. I'm gonna remove it 👍

dev-infra/BUILD.bazel
wagnermaciel: That sounds great to me. So unless anyone has anything else to add, the plan is to delete `//dev-infra/tools` and move it's contents to `//dev-infra/benchmarking`.

dev-infra/tmpl-package.json
wagnermaciel: I went through and added all of the deps that I could find to `tmpl_package.json` but this really should be double checked

dev-infra/tmpl-package.json
```diff
@@ -8,14 +8,52 @@
...
+    "@types/events": "^3.0.0",
+    "@types/fs-extra": "<from-root>",
+    "@types/glob": "^7.1.1",
+    "@types/jasmine": "<from-root>",
+    "@types/json5": "<from-root>",
+    "@types/minimatch": "^3.0.3",
+    "@types/node": "<from-root>",
+    "@types/shelljs": "<from-root>",
+    "@types/yaml": "<from-root>",
+    "@types/yargs": "<from-root>",
```
devversion, line 33: We could skip all the `@types` dependencies here. These are only needed for creation of the `pkg_npm`, but not needed by consumers.

dev-infra/tmpl-package.json
devversion: Awesome, thanks!

I'm unsure how we should manage the dependencies on `@angular/*` packages in the long-term, but for now that should be fine.

dev-infra/tools/BUILD.bazel
```diff
@@ -0,0 +1,12 @@
+package(default_visibility = ["//visibility:public"])
```
devversion, line 1: We should be able to delete that file now?

devversion (approved):
Two minor nits, then this should be ready to go!

PR also needs a rebase for CI.

BUILD.bazel
```diff
@@ -2,7 +2,6 @@ package(default_visibility = ["//visibility:public"])
 
 exports_files([
```
IgorMinar, line 3: can you please update the commit message as well similar to how I asked for before. in this case you might want to say something like this:
```
build(core): use dev-infra's component_benchmark rule throughout the repo

This change is a follow up on the earlier PoC that uses the newly created
`//dev-infra/tools/component_benchmark` rule in the remaining performance tests
in the repo.
```

dev-infra/tmpl-package.json
```diff
@@ -9,8 +9,9 @@
     "ts-circular-deps": "./ts-circular-dependencies/index.js"
   },
   "dependencies": {
+    "fs-extra": "<from-root>",
     "node-uuid": "<from-root>",
-    "fs-extra": "<from-root>"
+    "reflect-metadata": "<from-root>"
```
IgorMinar: why? none of our tests should require reflect-metadata. Can we please review this as it might be a symptom of an underlying bug that should be fixed instead of this dependency being provided.

IgorMinar: also, please update the commit message capturing the "why" :-) 

this is a great example of how documenting the "why" in commit messages can speed up code reviews - if the info is provided up front, we don't need to go back and forth as much.

dev-infra/BUILD.bazel
```diff
@@ -45,6 +45,7 @@ pkg_npm(
         "//packages/zone.js/dist:zone.js": "@npm//:node_modules/zone.js/dist/zone.js",
```
IgorMinar, line 47: `fix(dev-infra): ....`

https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit

dev-infra/benchmark/browsers/BUILD.bazel
```diff
@@ -33,7 +33,7 @@ web_test_archive(
         "@io_bazel_rules_webtesting//common/conditions:mac": {"CHROMIUM": "chrome-mac/Chromium.app/Contents/MacOS/chromium"},
```
IgorMinar, line 33: see! this is a great example of how awesome git is at tracking file moves if you use `mv` instead of `cp`. We now have the file history preserved and git even showed us what you changed in the file in addition to the move.

IgorMinar (changes requested):
Thanks @wagnermaciel, this looks like an awesome cleanup!

I left a bunch of comments - please don't get discouraged by their number they are mostly related to the commit message guidelines that you don't seem to be following in this PR and the fix should be simple.

Additionally, I suggest that you squash all the `cp` and `rm` commits, which should allow git to stitch the history together and emulate `mv` operation (more details in other commits).

Lastly, let's revisit the need to use `reflect-metadata` as our dep in these benchmarks. That doesn't look right.

Thanks for all the hard work!

dev-infra/benchmark/browsers/BUILD.bazel
wagnermaciel: I actually didn't know git was clever enough to track copying files. I just used what came to mind when trying to move the files over.

I'm going to go through and see if I can revert these commits and redo them properly

dev-infra/tmpl-package.json
wagnermaciel: @devversion I made the changes you requested. They are not in separate commits because they were squashed together to make this PR cleaner.

If you look at [`//dev-infra/tmpl-package.json`](https://github.com/angular/angular/pull/36434/files#diff-0a136973c4e9e2f95711f04ba1e719e7) you can see that I have removed all of the `@types` dependencies.

Thanks for helping me get these deps right!

JiaLiPassion (approved):
LGTM for zone.js part.

dev-infra/tmpl-package.json
devversion: Thanks! I also added an item to the dev-infra sync on how we want to manage these dependencies. In the current state, it's pretty inconvenient and difficult to enforce.

dev-infra/tmpl-package.json
```diff
@@ -8,14 +8,42 @@
     "ng-dev": "./cli.js",
     "ts-circular-deps": "./ts-circular-dependencies/index.js"
   },
+  "dependencies": {
+    "fs-extra": "<from-root>",
+    "node-uuid": "<from-root>"
+  },
   "peerDependencies": {
+    "@angular/bazel": "^9.0.0",
```
IgorMinar, line 16: just an FYI @wagnermaciel (no action needed): @devversion will update some of these `peerDependencies` to `dependencies` after your PR lands. Is is happening due to an unrelated dev-infra improvement work that @devversion is working on. We do want to land your PR before that happens so that we don't block you any longer. If you have any questions about this, please follow up with Paul. thanks!

dev-infra/tools/BUILD.bazel
```diff
@@ -1,12 +0,0 @@
-package(default_visibility = ["//visibility:public"])
```
IgorMinar: I think this diff doesn't belong to this commit, it seems to belong to the previous commit. I'm hower fine with keeping it here just so that you don't have to rebase this PR one more time.

modules/benchmarks/src/class_bindings/BUILD.bazel
```diff
@@ -1,7 +1,7 @@
 package(default_visibility = ["//modules/benchmarks:__subpackages__"])
 
 load("@io_bazel_rules_sass//sass:sass.bzl", "sass_binary")
-load("//tools/components:component_benchmark.bzl", "component_benchmark")
+load("//dev-infra/benchmark/component_benchmark:component_benchmark.bzl", "component_benchmark")
```
IgorMinar, line 4: yay! so much cleanup before we finally get to this PoC one liner. It might look like you haven't done much, but in fact this one liner is so significant because of all the underlying cleanup that happened to get here, and more importantly this oneliner makes a lot of our future work just a bunch of oneliners as well. Great work!

IgorMinar (approved):
This PR so now so much easier to review and the commit messages are quite good. The only thing I'd recommend in the future (not needed to fix in this PR) is to add one more sentence to each of the PR that answers the "why this change?" question - currently the commit messages primarily capture "what is this change?" which is good, but "why?" is more important.

In any case, this PR is great! Thanks for all the work! Let's get this in.

IgorMinar:
merge-assistance: if this PR doesn't cleanly cherry-pick on the patch branch, then please merge this into master-only and notify @wagnermaciel to create a separate PR that targets only the patch branch.

modules/benchmarks/src/class_bindings/BUILD.bazel
devversion: +100. awesome work!

AndrewKushnir:
@wagnermaciel, could you please rebase this PR on top of the current master? There are no conflicts from code perspective, but it'd be great to re-run CI again with the latest changes from master (that contain some CI/infra-related changes). Thank you.

wagnermaciel:
@AndrewKushnir Just finished rebasing on top of the current master as per your request 👍

IgorMinar:
@wagnermaciel please remove the "cleanup" label in the future after the PR is updated, that gives the signal to the caretaker that the PR is good to be merged now

AndrewKushnir:
Thanks for rebasing @wagnermaciel 👍

IgorMinar:
@wagnermaciel in general the caretaker will not pick up a PR that is not fully green which looks like this:

Screen Shot 2020-04-23 at 1 25 54 PM

It is the responsibility of a PR author (and secondarily the reviewers) to get the PR into a green state. If you can't and need help, you can ask anyone on the team or the caretaker to help get

PR #36452, opened 2020-04-06
ngcc: source map overflow

petebacondarwin:


packages/compiler-cli/ngcc/src/sourcemaps/source_file_loader.ts
```diff
@@ -43,40 +46,42 @@ export class SourceFileLoader {
       sourcePath: AbsoluteFsPath, contents: string|null, mapAndPath: null,
       previousPaths: AbsoluteFsPath[]): SourceFile|null;
```
JoostK, line 47: This overload signature + the param's JSDoc can be removed, which triggers removal of `previousPaths` from `processSources`.

packages/compiler-cli/ngcc/src/sourcemaps/source_file_loader.ts
```diff
@@ -97,20 +102,37 @@ export class SourceFileLoader {
       try {
         const fileName = external[1] || external[2];
         const externalMapPath = this.fs.resolve(this.fs.dirname(sourcePath), fileName);
-        return {map: this.loadRawSourceMap(externalMapPath), mapPath: externalMapPath};
-      } catch {
+        const map = this.loadRawSourceMap(externalMapPath);
+        this.trackPath(externalMapPath);
```
JoostK, line 106: I'm not sold on this call (and the one for `impliedMapPath` below) being done here. Instead, I'd move it up one level to keep the responsibility of tracking active resolutions in one place.

Looking into it more, a somewhat larger refactoring would make sense to me.  In these branches, the resolved map's path is tracked in the resolution stack. In `loadSourceFile` when `content` is null however, the source file itself is tracked without yet knowing whether it contains an inline source map or an external source map. So that seems like the incorrect place to be tracking paths (it would make it easier to follow if we track only map paths).

Since `processSources` is the only one who calls `loadSourceFile` where `content` is possibly null (besides a heap of tests), it would make most sense to me to move the loading of content if null to `processSources` somehow. We may then move the resolution stack tracking business into a single place in `loadSourceFile`:

```ts
      if (mapAndPath !== null) {
        const basePath = mapAndPath.mapPath || sourcePath;
        this.trackPath(basePath); // <-- we know for sure we are pushing a path of something that actually contains the source map.
        sources = this.processSources(basePath, mapAndPath.map, previousPaths);
        map = mapAndPath.map;
        inline = mapAndPath.mapPath === null;
       }
```

Note: I'm not sure whether the above pans out, I may be missing something here. I am okay with approving this without addressing this at this moment.

JoostK (changes requested):

packages/compiler-cli/ngcc/src/sourcemaps/source_file_loader.ts
```diff
@@ -97,20 +102,37 @@ export class SourceFileLoader {
...
 
     return null;
   }
 
+  /**
+   * Track source file paths if we have loaded them from disk so that we don't get into an infinite
+   * recursion.
+   */
+  private trackPath(path: AbsoluteFsPath): void {
+    if (this.currentPaths.includes(path)) {
```
gkalpak, line 129: I wonder if it will make a noticeable difference to replace the array with a `Set` (for faster lookups).

packages/compiler-cli/ngcc/src/sourcemaps/source_file_loader.ts
```diff
@@ -43,40 +46,42 @@ export class SourceFileLoader {
...
       previousPaths: AbsoluteFsPath[]): SourceFile|null;
   loadSourceFile(
-      sourcePath: AbsoluteFsPath, contents: string|null = null, mapAndPath: MapAndPath|null = null,
-      previousPaths: AbsoluteFsPath[] = []): SourceFile|null {
-    if (contents === null) {
-      if (!this.fs.exists(sourcePath)) {
-        return null;
+      sourcePath: AbsoluteFsPath, contents: string|null = null,
+      mapAndPath: MapAndPath|null = null): SourceFile|null {
+    const previousPaths = this.currentPaths.slice();
```
gkalpak, line 51: I wonder how much this slicing affects performance. Would it be worh trying to optimize it away?

gkalpak (commented):
Looks reasonable :grin:
Is there a Jira ticket for this?

packages/compiler-cli/ngcc/src/sourcemaps/source_file_loader.ts
JoostK: We _do_ rely on the insertion order info when logging the resolution path in case of an error, I don't what the `Set`s iterator guarantees are wrt insertion order?

petebacondarwin:
Related ticket FW-2041

packages/compiler-cli/ngcc/src/sourcemaps/source_file_loader.ts
petebacondarwin: The previous algorithm was doing a `.concat()` each time, which has the same cost, and I didn't see this being a significant part of the time it takes.

packages/compiler-cli/ngcc/src/sourcemaps/source_file_loader.ts
petebacondarwin: The previous algorithm was also doing this lookup and it was not a significant part of the cost of the flattening. (Also in general this array never gets larger than 2 or 3 items.

JoostK (approved):

PR #36456, opened 2020-04-06
ci: remove IgorMinar from reviewers list for pullapprove fallback group

josephperrott:
Historically we have had a pullapprove group `fallback` which acted as
a catch all for files which did not match any other groups.  This
group assigned reviews to IgorMinar, however it was not apparent that
this group was assigned.  This change removes this assignment.  This
group as active should always coincide with failures of the pullapprove
verification script. We continue to have this group as a secondary test
ensuring all files in the repo are captured by the pullapprove config.

### DIFF
--- a/.pullapprove.yml
+++ b/.pullapprove.yml
@@ -1126,30 +1126,32 @@ groups:
   #  Catch all for if no groups match the code change
   # ====================================================
   fallback:
+    # A group is considered to be `active` for a PR if at least one of group's
+    # conditions matches the PR.
+    #
+    # The PullApprove CI check should fail if a PR has no `active` groups, as
+    # this indicates the PR is modifying a file that has no owner.
+    #
+    # This is enforced through the pullapprove verification check done
+    # as part of the CircleCI lint job.  Failures in this lint job should be
+    # fixed as part of the PR.  This can be done by updating the
+    # `.pullapprove.yml` file cover the unmatched path.
+    # The pullapprove verification script is part of the ng-dev tool and can be
+    # run locally with the command: `yarn -s ng-dev pullapprove verify`
+    #
+    # For cases in which the verification check fails to ensure coverage, this
+    # group will be active.  The expectation is that this should be remedied
+    # before merging the PR as described above.  In an emergency situation
+    # `global-approvers` can still approve PRs that match this `fallback` rule,
+    # but that should be an exception and not an expectation.
     conditions:
       - *can-be-global-approved
-      # Groups which are found to have matching conditions are `active`
-      # according to PullApprove.
-      #
-      # If not groups are matched and `active`, it should be considered a CI
-      # failure.  This is done through the pullapprove verification check done
-      # as part of the CircleCI lint job.  Failures in this lint job should be
-      # fixed as part of the PR.  This can be done by updating the
-      # `.pullapprove.yml` file cover the unmatched path.
-      #
-      # For cases in which the verification check fails to ensure coverage, this
-      # group will be active.  The expectation is that this should be remedied
-      # before merging the PR as described above.  In an emergency situation
-      # `global-approvers` can still approve PRs that match this `fallback` rule,
-      # but that should be an exception and not an expectation.
-      #
-      # The following groups have no conditions:
+      # The following groups have no conditions and will be `active` on all PRs
       # - `global-approvers`
       # - `global-docs-approvers`
       #
-      # Because these 2 groups have no conditions, they will be `active` on all PRs.
-      # Because the minimum number of active groups a PR can have is 2, this `fallback`
-      # rule should be matched anytime the number of active groups is at or below this
-      # minumum.  This is indended as a protection to ensure that pullapprove does not
-      # incidently mark a PR as passing without meeting the review criteria.
+      # Since this means the minimum number of active groups a PR can have is 2, this
+      # `fallback` group should be matched anytime the number of active groups is at or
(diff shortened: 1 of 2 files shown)


.pullapprove.yml
```diff
@@ -1081,7 +1081,9 @@ groups:
       # Groups which are found to have matching conditions are `active`
       # according to PullApprove. If no groups are matched and considered
       # active, we still want to have a review occur.
```
IgorMinar, line 1083:
```suggestion
      # active, we want the pullapprove check to fail. To resolve the failure, the `.pullapprove.yml`
      # config update should be made to cover the unmatched path (the current PR will be
      # intentionally blocked until this change is processed). In an emergency situation `global-approvers`
      # can still approve PRs that match this `fallback` rule, but that should be an exception and not
      # an expectation.
```

.pullapprove.yml
```diff
@@ -1081,7 +1081,9 @@ groups:
...
-      - len(groups.active) == 0
-    reviewers:
-      users:
-        - IgorMinar
+      #
+      # Since our global and global docs approval groups have no conditions
+      # it is set to match all changes, the groups.active value always will
+      # be at least 2, we should consider this matched if 2 or less groups
+      # are active.
```
IgorMinar, line 1088: this isn't clear to me... I'll try to correct.. please review and accept/reword further..

```suggestion
      # Since `global-approvers` and `global-docs-approvers` groups have no `conditions`,
      # these groups will match changes to any files and become `active`.
      # The `groups.active` value will therefore always be at least 2 for any PR that matches
      # the "fallback" group.
      # In order for us to recognize that a PR has no requested reviewer, we should
      # check if the PR has at least 2 active reviewer groups.
```

.pullapprove.yml
```diff
@@ -1081,7 +1081,9 @@ groups:
...
+      - len(groups.active) <= 2
```
IgorMinar, line 1089: even after rewriting the docs above, I'm not sure if `<= 2` is actually correct. Why 2? and why `=` and not `<`?

This is very confusing, can you please help by explaining more what's going on here? thanks!

.pullapprove.yml
IgorMinar: I also suggest mentioning the pullapprove test script here and that this group is just a fallback that would be triggered only if that test is disabled, bypassed, or broken in some other way. thanks!

IgorMinar (changes requested):
Can you please help by clarifying this change?

I tried to update the docs, but I'm not sure I understand the possibilities and limitations of pullapprove and your thinking behind implementing this change. thanks!

josephperrott:
@IgorMinar Update the commenting in the `fallback` group to expand on the intention fo the group.  PTAL.

.pullapprove.yml
```diff
@@ -1083,9 +1083,27 @@ groups:
     conditions:
       - *can-be-global-approved
       # Groups which are found to have matching conditions are `active`
```
IgorMinar, line 1085:
```suggestion
      # A group is considered to be `active` for a PR if at least one of group's conditions matches the PR.
```

.pullapprove.yml
```diff
@@ -1083,9 +1083,27 @@ groups:
...
+      # but that should be an exception and not an expectation.
+      #
+      # The following groups have no conditions:
+      # - `global-approvers`
+      # - `global-docs-approvers`
+      #
+      # Because these 2 groups have no conditions, they will be `active` on all PRs.
+      # Because the minimum number of active groups a PR can have is 2, this `fallback`
+      # rule should be matched anytime the number of active groups is at or below this
+      # minumum.  This is indended as a protection to ensure that pullapprove does not
```
IgorMinar, line 1107:
```suggestion
      # minimum. This is intended as a protection to ensure that pullapprove does not
```

.pullapprove.yml
```diff
@@ -1083,9 +1083,27 @@ groups:
...
       # Groups which are found to have matching conditions are `active`
-      # according to PullApprove. If no groups are matched and considered
-      # active, we still want to have a review occur.
-      - len(groups.active) == 0
-    reviewers:
-      users:
-        - IgorMinar
+      # according to PullApprove.
+      #
+      # If not groups are matched and `active`, it should be considered a CI
```
IgorMinar, line 1088:
```suggestion
      # The PullApprove CI check should fail if a PR has no `active` groups, as that indicates that the PR is modifying a file that has no owner.
```

.pullapprove.yml
```diff
@@ -1083,9 +1083,27 @@ groups:
     conditions:
       - *can-be-global-approved
```
IgorMinar, line 1084: hey Joey, I started by making a few small suggestions (below), but after rereading this whole doc below even with my suggestion, it still doesn't make sense.

Can you please split the doc into two sections:

One that describes this rule, what is its purpose and what is its relationship to the pullapprove verification (could you also add a pointer where this verification script lives?).

And the second one that describes the `active` groups and how we determined that `len(groups.active) <= 2` is the right condition here.

The first doc can be posted just below the `fallback` header above, and the second can be placed just about the condition as you have it now.

IgorMinar (changes requested):

.pullapprove.yml
josephperrott: Done.

IgorMinar (approved):

IgorMinar:
lgtm now. thank you!

PR #36464, opened 2020-04-06
build: update to latest version of yarn

josephperrott:


devversion (approved):
LGTM. Should we update version ranges in the `package.json`? Adding `PR action: cleanup` so you can react to my feedback.

josephperrott:
@devversion I updated them

I am unsure of how we should be handling this since yarn should be doing the vendoring for us now, but it doesn't hurt to update it.

devversion:
@josephperrott Yeah, I'm not sure either. It sounds like we could just remove the range for `Yarn`.

atscott:
@josephperrott - Is this ready to merge? If so, can you please remove Igor from the reviewers list and/or add the merge assistance label if pullapprove is still not happy?

Issue #36414, opened 2020-04-03
Slower Build Performance after Upgrade from 9.0.7 to 9.1.0


sbradl:
Since the update from Angular 9.0.7 to 9.1.0 our build times increased a lot. Here are some times:

v9.0.7 - prod build: 8:41
v9.1.0 - prod build: 13:09

These times are on a windows machine with node v12

On linux and mac the difference is much smaller - only a few seconds

petebacondarwin:
Tracking at FW-2047 - hopefully we can get this prioritised for next week.

sbradl:
> So the actual ngcc processing should still only run once per package, so unless you modify your node_modules, the processing of these packages should not happen a second time.

This seems to work fine for me

> Regarding 1. - this might well be the problem, since on Windows it is quite slow to traverse the file-system checking for entry-points that "might" need processing. This is happening even if it turns out that they are all already processed. That being said, I recently implemented an `__ngcc_entry_point_manifest__.json` file that should be written to the top of your `node_modules` directory that should avoid having to do this traversal... Can you check to see if one of these is appearing?

Can't see such a file

sbradl:
For the parallel execution of ngcc I can see the following:
As the libs are processed I can see node-processes beeing spawned. Most of them take around 50MBs of memory but only one of them takes CPU time (around 10%).

blackholegalaxy:
> That being said, if a module was already processed by the async NGCC should bail out pretty fast, it will just check for the build marker and exit.
> 

The problem is when running on a pipeline with no node_modules cache, or on new developer computer, it takes a LOT of time. Indeed after node_modules has been checked and compiled once everything runs pretty fast. But it doesn't change the fact going from 9.0.7 to 9.1.0 time has increased even if ngcc concurrent was implemented. It should had reduced our initial build time.

> 
> > Fun fact: adding a new dependency in the project (dotenv) caused the entire ngcc compile to occur again even if it was the only changed dependency.
> 
> This is expected if you use `Yarn` as your package manager. This is because `Yarn` will remove the entire `node_modules` and re-install all dependencies when you add a new dependency. Which means that NGCC will need to re-run on all modules, which in this case will only run in sync mode.
> 

Indeed we use yarn. So :+1:

> @blackholegalaxy, can you please share the output of `ng version` also can you explain how are you checking if NGCC is being run in parallel or not?

[...]

PR #36486, opened 2020-04-07
perf(ngcc): read dependencies from entry-point manifest

petebacondarwin:
Previously, even if an entry-point did not need to be processed,
ngcc would always parse the files of the entry-point to compute
its dependencies. This can take a lot of time for large node_modules.

Now these dependencies are cached in the entry-point manifest,
and read from there rather than computing them every time.

See https://github.com/angular/angular/issues/36414#issuecomment-608401834
FW-2047

### DIFF
--- a/packages/compiler-cli/ngcc/src/dependencies/dependency_host.ts
+++ b/packages/compiler-cli/ngcc/src/dependencies/dependency_host.ts
@@ -6,6 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 import {AbsoluteFsPath, FileSystem, PathSegment} from '../../../src/ngtsc/file_system';
+import {EntryPoint} from '../packages/entry_point';
 import {resolveFileWithPostfixes} from '../utils';
 
 import {ModuleResolver} from './module_resolver';
@@ -19,6 +20,11 @@ export interface DependencyInfo {
   dependencies: Set<AbsoluteFsPath>;
   missing: Set<AbsoluteFsPath|PathSegment>;
   deepImports: Set<AbsoluteFsPath>;
+}
+
+export interface EntryPointWithDependencies {
+  entryPoint: EntryPoint;
+  dependencies: DependencyInfo;
 }
 
 export function createDependencyInfo(): DependencyInfo {
--- a/packages/compiler-cli/ngcc/src/dependencies/dependency_host.ts
+++ b/packages/compiler-cli/ngcc/src/dependencies/dependency_host.ts
@@ -24,7 +24,7 @@ export interface DependencyInfo {
 
 export interface EntryPointWithDependencies {
   entryPoint: EntryPoint;
-  dependencies: DependencyInfo;
+  depInfo: DependencyInfo;
 }
 
 export function createDependencyInfo(): DependencyInfo {
--- a/packages/compiler-cli/ngcc/src/dependencies/dependency_resolver.ts
+++ b/packages/compiler-cli/ngcc/src/dependencies/dependency_resolver.ts
@@ -14,7 +14,7 @@ import {NgccConfiguration} from '../packages/configuration';
 import {EntryPoint, EntryPointFormat, getEntryPointFormat, SUPPORTED_FORMAT_PROPERTIES} from '../packages/entry_point';
 import {PartiallyOrderedList} from '../utils';
 
-import {createDependencyInfo, DependencyHost, DependencyInfo} from './dependency_host';
+import {createDependencyInfo, DependencyHost, EntryPointWithDependencies} from './dependency_host';
 
 const builtinNodeJsModules = new Set<string>(require('module').builtinModules);
 
@@ -94,7 +94,7 @@ export class DependencyResolver {
    * @param target If provided, only return entry-points depended on by this entry-point.
    * @returns the result of sorting the entry points by dependency.
    */
-  sortEntryPointsByDependency(entryPoints: EntryPoint[], target?: EntryPoint):
+  sortEntryPointsByDependency(entryPoints: EntryPointWithDependencies[], target?: EntryPoint):
       SortedEntryPointsInfo {
     const {invalidEntryPoints, ignoredDependencies, graph} =
         this.computeDependencyGraph(entryPoints);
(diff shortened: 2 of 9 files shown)


JoostK (approved):
Nice!

packages/compiler-cli/ngcc/src/dependencies/dependency_host.ts
```diff
@@ -21,6 +22,11 @@ export interface DependencyInfo {
   deepImports: Set<AbsoluteFsPath>;
 }
 
+export interface EntryPointWithDependencies {
+  entryPoint: EntryPoint;
+  dependencies: DependencyInfo;
```
gkalpak, line 27: It is a little confuding that there is a `dependencies` property that has a `dependencies` property. How about renaming this to `dependencyInfo` (or `depInfo` 😱)?

gkalpak (approved):
It would be great to see some numbers. Other than that looks great ✨

packages/compiler-cli/ngcc/src/dependencies/dependency_resolver.ts
```diff
@@ -120,18 +120,21 @@ export class DependencyResolver {
...
-  getEntryPointDependencies(entryPoint: EntryPoint): DependencyInfo {
-    const formatInfo = this.getEntryPointFormatInfo(entryPoint);
-    const host = this.hosts[formatInfo.format];
-    if (!host) {
-      throw new Error(
-          `Could not find a suitable format for computing dependencies of entry-point: '${
-              entryPoint.path}'.`);
+  getEntryPointWithDependencies(entryPoint: EntryPoint): EntryPointWithDependencies {
+    const dependencies = createDependencyInfo();
+    if (entryPoint.compiledByAngular) {
```
gkalpak, line 125: This seems like a independent fix (which I think would also fix #36524). Shouldn't it be on a separate commit?

packages/compiler-cli/ngcc/src/dependencies/dependency_host.ts
petebacondarwin: Sure!

packages/compiler-cli/ngcc/src/dependencies/dependency_resolver.ts
petebacondarwin: This is not actually a fix. This was already being handled in `computeDependencyGraph()` where we only add nodes to the graph if they are compiled by Angular.

Previously, this method (in the form of `getEntryPointDependencies()`) was only called on nodes in the graph, which had already been filtered by being compiled by Angular. Now this method is called on every entry-point, so we need to guard against it here too.

With that in mind, I don't think this would solve #36524. See my comment there...

packages/compiler-cli/ngcc/src/entry_point_finder/directory_walker_entry_point_finder.ts
```diff
@@ -32,11 +33,11 @@ export class DirectoryWalkerEntryPointFinder implements EntryPointFinder {
    * all package entry-points.
    */
   findEntryPoints(): SortedEntryPointsInfo {
-    const unsortedEntryPoints: EntryPoint[] = [];
+    const unsortedEntryPoints: EntryPointWithDependencies[] = [];
     for (const basePath of this.basePaths) {
       const entryPoints = this.entryPointManifest.readEntryPointsUsingManifest(basePath) ||
           this.walkBasePathForPackages(basePath);
-      unsortedEntryPoints.push(...entryPoints);
+      entryPoints.forEach(e => unsortedEntryPoints.push(e));
```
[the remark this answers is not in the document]
petebacondarwin: When I was looking at the generated JS code for this library, I noticed that the current code gets downleveled to ([line 43](https://unpkg.com/browse/@angular/compiler-cli@9.1.1/ngcc/src/entry_point_finder/directory_walker_entry_point_finder.js)):

```js
unsortedEntryPoints.push.apply(unsortedEntryPoints, __spread(entryPoints));
```

See [this playground](https://www.typescriptlang.org/play/?downlevelIteration=true&target=1#code/MYewdgzgLgBApmKAnAngBRAS0RGBeGAbQF0BuAWAChRJYBXSEJKOAEwFFFUNspcCSFSgwhMWHLuiw4AdAAc6EABYAKGeoTIpvCAEpSQA)

where `__spread()` is implemented like:

```js
var __read = (this && this.__read) || function (o, n) {
    var m = typeof Symbol === "function" && o[Symbol.iterator];
    if (!m) return o;
    var i = m.call(o), r, ar = [], e;
    try {
        while ((n === void 0 || n-- > 0) && !(r = i.next()).done) ar.push(r.value);
    }
    catch (error) { e = { error: error }; }
    finally {
        try {
            if (r && !r.done && (m = i["return"])) m.call(i);
        }
        finally { if (e) throw e.error; }
    }
    return ar;
};
var __spread = (this && this.__spread) || function () {
    for (var ar = [], i = 0; i < arguments.length; i++) ar = ar.concat(__read(arguments[i]));
    return ar;
};
```

This seemed to me highly sub-optimal and the approach in this PR is much simpler - most likely faster and certainly creates less memory pressure.

---

This brings up a side-point.  If ngcc (and ngc) is always going to be run with node.js >= 10.0 then perhaps we should change our downleveling settings?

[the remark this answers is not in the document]
JoostK: Yeah, that's a fair point. `__spread` usually shows up in profiling sessions quite noticeable so I would love to get rid of it. Emitting to ES2015 seems like the best way to achieve that, indeed.

petebacondarwin:
Initial tests on AIO... I first built all the packages in the `node_modules` directory by running `ngcc -l debug`. Then ran the same command again a few times, which represents the "noop" ngcc run.

  | Master | PR
-- | -- | --
Full | 308.61 | 321.08
No manifest | 4.27 | 6.68
Noop 1 | 5.26 | 1.21
Noop 1 | 5.16 | 1.72
Noop 1 | 4.92 | 1.78
Noop 1 | 4.46 | 1.75
**Average noop** | **4.95** | **1.615**

So it can be seen that there is no real difference if the manifest has to be created.
But where there is a manifest, you can see that this PR is about 3x faster than `master`.

petebacondarwin:
Further improvements, particularly only loading the `Transformer` code if it is actually required knocks a bit more time off the noop runs:

  | Master | PR (no lazy) | PR (with lazy)
-- | -- | -- | --
noop 1 | 5.26 | 1.21 | 0.97
noop 2 | 5.16 | 1.72 | 0.96
noop 3 | 4.92 | 1.78 | 0.89
noop 4 | 4.46 | 1.75 | 0.97
**Average noop** | **4.95** | **1.615** | **0.9475**

petebacondarwin:
@JoostK and @gkalpak - I added a couple of commits. PTAL

packages/compiler-cli/ngcc/src/packages/entry_point_manifest.ts
```diff
@@ -156,24 +171,29 @@ export class EntryPointManifest {
...
     return null;
   }
 }
 
+export type EntryPointPaths = [
+  string,
+  string,
+  Array<AbsoluteFsPath>?,
+  Array<AbsoluteFsPath|PathSegment>?,
+  Array<AbsoluteFsPath>?,
```
gkalpak, line 188: Would it make sense to use relative paths for these too?

packages/compiler-cli/ngcc/src/main.ts
```diff
@@ -315,6 +314,7 @@ export function mainNgcc({
   const createCompileFn: CreateCompileFn = onTaskCompleted => {
     const fileWriter = getFileWriter(
         fileSystem, logger, pkgJsonUpdater, createNewEntryPointFormats, errorOnFailedEntryPoint);
+    const {Transformer} = require('./packages/transformer');
```
gkalpak, line 317: I assume we can do more of this. Is `Trasnformer` special in some way (e.g. particularly heavy to load)?

gkalpak (approved):

packages/compiler-cli/ngcc/src/main.ts
petebacondarwin: Yes, Transformer brings in all the trait compiler stuff, all the reflection hosts etc.

petebacondarwin: I looked at other imports but most of them are needed in the top level main function :-/

packages/compiler-cli/ngcc/src/packages/entry_point_manifest.ts
petebacondarwin: I thought about it for `dependencies` and `deep-imports` even though there is a chance that they might start with `..` (but that probably doesn't matter). But I didn't want to do it for `missing` since some of these are `PathSegment`s but relative to the import place in the code, so we couldn't be sure, when reading in the manifest, whether we are reading in an AbsolutePath that needs de-relativising or a PathSegment that needs to be left alone.

JoostK (approved):

PR #36497, opened 2020-04-07
Correct commit message linting issue for git worktree users

josephperrott:
@petebacondarwin If possible could you confirm this works as expected now for you?


petebacondarwin:
@josephperrott - I pushed a fixup commit, because the `filePath` can actually be absolute, so using `join()` resulted in a broken path. After that it works for me locally.

gkalpak:
FYI, here is a similar alternative that works better on Windows: #36507

josephperrott:
Closing in favor of #36507

PR #36498, opened 2020-04-08
fix(upgrade): update $locationShim to handle Location changes before …

### LINKED ISSUE #36492: $locationShim.onChange writes wrong url when redirects are used with deferred bootstrapping
# 🐞 bug report

### Affected Package
The issue is caused by package @angular/common/upgrade

### Is this a regression?
No, this behavior never worked

### Description
The $locationShim has an option to provide the AngularJS $injector as a promise which means that the [initialize](https://github.com/angular/angular/blob/6de4cbdd4141ac3ade2fcc3fffb3a18e75978460/packages/common/upgrade/src/location_shim.ts#L71) will be run when that promise resolves. Now that we are deferring the bootstrapping of AngularJS we run into a situation where the $locationShim is initialized after the Location.onUrlChange listeners have been notified.

If the user is navigating to a URL that redirects (for example /hello redirecting to /hello/world) then we see that $locationShim emits an onChange event for /hello when it shouldn't be. This is because a call to $location.search in the AngularJS side of our application triggers the $locationShim to call composeUrls  which thinks there has been a change from /hello/world to /hello.

## 🔬 Minimal Reproduction
This reproduces within a complex dual bootstrapped application within Google. I have a fix for the issue so I'm just posting this bug report for tracking purposes.

## 🌍  Your Environment

**Angular Version:**
Running at head

agale123:
…initialization

Updates the $locationShim to receive the most recent Location change
made, even if it happened before initialize() is called. This is
important when AngularJS bootstrapping is deferred and there is a delay
between when $locationShim is constructed and when it is initialized.
With this change, the $locationShim will correctly reflect any redirects
that occurred between construction and initialization.

Closes #36492

## PR Checklist

**There weren't any unit tests for the existing behavior and it appears that the test mocks aren't set up well for testing Location changes. I tried for a bit to make it work, but that seemed like a large separate change that needs to happen.**
**No doc changes are needed because the external API isn't changing**

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Any Location changes (for example the results of redirects) that happen between $locationShim being constructed and initialized are dropped. This can lead to the onChange function being called with incorrect values. There are also some style changes from running the code formatter.

Issue Number: #36492

## What is the new behavior?
Any Location changes (for example the results of redirects) that happen between $locationShim being constructed and initialized are taken into account. This means that incorrect onChange events won't be emitted.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/common/upgrade/src/location_shim.ts
+++ b/packages/common/upgrade/src/location_shim.ts
@@ -8,6 +8,7 @@
 
 import {Location, LocationStrategy, PlatformLocation} from '@angular/common';
 import {UpgradeModule} from '@angular/upgrade/static';
+import {ReplaySubject} from 'rxjs';
 
 import {UrlCodec} from './params';
 import {deepEqual, isAnchor, isPromise} from './utils';
@@ -50,7 +51,7 @@ export class $locationShim {
 
   private cachedState: unknown = null;
 
-
+  private urlChanges = new ReplaySubject<{newUrl: string, newState: unknown}>(1);
 
   constructor(
       $injector: any, private location: Location, private platformLocation: PlatformLocation,
@@ -71,6 +72,10 @@ export class $locationShim {
     this.cacheState();
     this.$$state = this.browserState();
 
+    this.location.onUrlChange((newUrl, newState) => {
+      this.urlChanges.next({newUrl, newState});
+    });
+
     if (isPromise($injector)) {
       $injector.then($i => this.initialize($i));
     } else {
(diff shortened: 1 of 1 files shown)


agale123:
Green TAP presubmit: https://test.corp.google.com/ui#id=OCL:305370204:BASE:305356164:1586304237385:e5065f4b

gkalpak (approved):
LGTM :+1:
It would be nice if a test was added for the change (somewhere [here][1] I assume).

[1]: https://github.com/angular/angular/blob/554420f83436826fef52a43579bdaceb449e7214/packages/common/upgrade/test/upgrade.spec.ts

agale123:
That is the file where I was originally trying to add unit tests but it appears that there are no existing tests for the logic to read state from `Location.onUrlChanges`. My guess is that it was already complex mocking out all of the AngularJS logic so adding mocks for the Angular interaction is difficult to add on top of that.

mhevery (approved):

mhevery:
[presubmit](http://test/OCL:306301006:BASE:306301023:1586810735620:e602d96d)

PR #36500, opened 2020-04-08
ci: remove reliance on Github API for CI setup

josephperrott:


josephperrott:
@gkalpak and @devversion I wanted to get your thoughts on this idea for getting the data we need on our rebases, rather than using the Github API.

.circleci/env.sh
```diff
@@ -22,6 +22,7 @@ else
   ####################################################################################################
   # See https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables for more info.
   ####################################################################################################
+  setPublicVar CI "$CI"
```
gkalpak, line 25: 🤔

tools/utils/get-refs.js
```diff
@@ -0,0 +1,128 @@
...
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+const {execSync} = require('child_process');
+
+/** A regex to select a ref that matches our semver refs. */
+const semverRegex = /^(\d+)\.(\d)+\.x$/;
+
+/**
+ * Syncronously executes the command
```
gkalpak, line 14:
```suggestion
 * Synchronously executes the command.
```

tools/utils/get-refs.js
```diff
@@ -0,0 +1,128 @@
...
+ *
+ * Return the trimmed stdout as a string, with an added attribute of the exit code.
+ */
+function exec(command) {
+  let output = new String();
+  try {
+    output += execSync(command,{stdio: ['pipe','pipe','ignore']}).toString().trim();
```
gkalpak, line 21:
```suggestion
    output += execSync(command, {stdio: ['pipe','pipe','ignore']}).toString().trim();
```
:grin:

tools/utils/get-refs.js
```diff
@@ -0,0 +1,128 @@
...
+/** Gets the refs and shas for the base and target of the current environment. */
+function getRefs() {
+  let base, target;
+  if (process.env['CI']) {
+    base = getRefAndShas(process.env['CI_GIT_BASE_REVISION'], process.env['CI_REPO_OWNER']);
+    target = getRefAndShas(process.env['CI_GIT_REVISION'], process.env['CI_PR_USERNAME']);
+  } else {
+    const originSha = exec(`git rev-parse HEAD`);
+    base = getRefAndShas(originSha, 'origin');
+    target = getParentBranchForHead('upstream');
```
gkalpak, line 118: I am a little confused. Should `base` and `target` be reversed here?

tools/utils/get-refs.js
```diff
@@ -0,0 +1,128 @@
...
+}
+
+/**
+ * Sort a list of fullpath refs into a list and then provide the first entry.
+ *
+ * The sort order will first find master ref, and then any semver ref, followed
+ * by the rest of the refs in the order provided.
+ */
+function getRefFromBranchList(gitOutput) {
+  const branches = gitOutput.trim().split('\n').map(x => x.trim().split('/')[1]);
```
gkalpak, line 36: Can't there be branch names that contain `/`?

tools/utils/get-refs.js
```diff
@@ -0,0 +1,128 @@
...
+  return branches.sort((a, b) => {
+    if (a === 'master') {
+      return -1;
+    }
+    if (b === 'master') {
+      return 1;
+    }
+    if (a.match(semverRegex) && b.match(semverRegex)) {
```
gkalpak, line 44: Since you don't care about the captured groups:

```suggestion
    if (semverRegex.test(a) && semverRegex.test(b)) {
```

tools/utils/get-refs.js
```diff
@@ -0,0 +1,128 @@
...
+      const [_, aMajor, aMinor] = a.match(semverRegex);
+      const [__, bMajor, bMinor] = b.match(semverRegex);
```
gkalpak, line 46:
```suggestion
      const [, aMajor, aMinor] = a.match(semverRegex);
      const [, bMajor, bMinor] = b.match(semverRegex);
```

tools/utils/get-refs.js
```diff
@@ -0,0 +1,128 @@
...
+      return parseInt(bMajor) - parseInt(aMajor) || parseInt(aMinor) - parseInt(bMinor) || 0;
+    }
+    if (a.match(semverRegex)) {
```
gkalpak, line 49:
```suggestion
    if (semverRegex.test(a)) {
```

tools/utils/get-refs.js
```diff
@@ -0,0 +1,128 @@
...
+      return -1;
+    }
+    if (b.match(semverRegex)) {
```
gkalpak, line 52:
```suggestion
    if (semverRegex.test(b)) {
```

tools/utils/get-refs.js
```diff
@@ -0,0 +1,128 @@
...
+    if (a === 'master') {
+      return -1;
+    }
+    if (b === 'master') {
+      return 1;
+    }
+    if (a.match(semverRegex) && b.match(semverRegex)) {
+      const [_, aMajor, aMinor] = a.match(semverRegex);
+      const [__, bMajor, bMinor] = b.match(semverRegex);
+      return parseInt(bMajor) - parseInt(aMajor) || parseInt(aMinor) - parseInt(bMinor) || 0;
```
gkalpak, line 47: It is [generally safer][1] to always specify the radix, when using `parseInt()`. Not liekly to matter here, but you never know.

[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt#Description

tools/utils/get-refs.js
```diff
@@ -0,0 +1,128 @@
...
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+const {execSync} = require('child_process');
+
+/** A regex to select a ref that matches our semver refs. */
+const semverRegex = /^(\d+)\.(\d)+\.x$/;
```
gkalpak, line 11:
```suggestion
const semverRegex = /^(\d+)\.(\d+)\.x$/;
```

gkalpak (commented):
If it works, I am fine with it. OOC, what is the benefit? Not having to use a GitHub token or suffer from API rate-limiting?

BTW, it would make the code easier to follow, if we included examples of what the output of each git command will look like in comments 🙏

.circleci/env.sh
josephperrott: Yep, its provided by Circle as an environment variable, and is set to true when we are on CI.

gkalpak: I know. All CI providers I know of set this. I just wonder what is the benefit of setting the `$CI` env var to the value of the `$CI` env var :grin:
(I guess, the benefit is being explicit, which I am definitely on-board with 👍)

tools/utils/get-refs.js
```diff
@@ -0,0 +1,185 @@
...
+      return parseInt(bMajor, 10) - parseInt(aMajor, 10) ||
+          parseInt(aMinor, 10) - parseInt(bMinor, 10) || 0;
+    }
+    if (aIsSemver) {
+      return -1;
+    }
+    if (bIsSemver) {
+      return 1;
+    }
+    return 0;
```
devversion, line 64: I wish we could use the `semver` package for this, but I think these scripts can't rely on the `node_modules` as of now.

tools/utils/get-refs.js
```diff
@@ -0,0 +1,185 @@
...
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+const {execSync} = require('child_process');
+
+/** A regex to select a ref that matches our semver refs. */
+const semverRegex = /^(\d+)\.(\d+)\.x$/;
+const branchRegex = /^\w+\/(.*)/;
```
devversion, line 12: Can you please add a comment for the `branchRegex` too.

tools/utils/get-refs.js
```diff
@@ -0,0 +1,185 @@
...
+/**
+ * Sort a list of fullpath refs into a list and then provide the first entry.
+ *
+ * The sort order will first find master ref, and then any semver ref, followed
+ * by the rest of the refs in the order provided.
+ */
+function getRefFromBranchList(gitOutput) {
+  const branches = gitOutput.split('\n')
+                       .map(b => {
+                         const branch = branchRegex.exec(b);
```
devversion, line 39: Would it be more safe to just omit the remote name by passing the remote to this function?

tools/utils/get-refs.js
```diff
@@ -0,0 +1,185 @@
...
+  return output;
+}
+
+/**
+ * Sort a list of fullpath refs into a list and then provide the first entry.
+ *
+ * The sort order will first find master ref, and then any semver ref, followed
+ * by the rest of the refs in the order provided.
+ */
+function getRefFromBranchList(gitOutput) {
```
devversion, line 36: Maybe we should explain why we actually want to sort? i.e. if `master` is found, then that will be used since it's our active development branch, otherwise if `master` did not match, we try to find the most recent publish branch in case X has been just merged into the patch branch.

devversion (approved):
The approach looks reasonable to me. I'd prefer the file name to be changed to something more specific that mentions at least `git`. `getRefs` is pretty ambiguous IMO.

.circleci/env.sh
josephperrott: Yeah, I figured it wouldn't hurt to just pass forward the environment variables since it could theoretically contain information in the future.

tools/utils/get-refs.js
josephperrott: We likely won't be able to rely on it at any point unless we were to host the package ourselves in `third_party` so that its available without a yarn install.

tools/utils/get-refs.js
```diff
@@ -0,0 +1,184 @@
...
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+const {execSync} = require('child_process');
+
+/** A regex to select a ref that matches our semver refs. */
+const semverRegex = /^(\d+)\.(\d+)\.x$/;
+
+/**
+ * Synchronously executes the command
```
gkalpak, line 14:
```suggestion
 * Synchronously executes the command.
```

tools/utils/get-refs.js
```diff
@@ -0,0 +1,184 @@
...
+ *
+ * Return the trimmed stdout as a string, with an added attribute of the exit code.
+ */
+function exec(command) {
+  let output = new String();
+  try {
+    output += execSync(command, {stdio: ['pipe', 'pipe', 'ignore']}).toString().trim();
```
gkalpak, line 21: Why ignore stderr? Would it be a problem to use `inherit` for it (in case there's useful info there)?

tools/utils/get-refs.js
```diff
@@ -0,0 +1,184 @@
...
+    output.code = err.status || 1;
+  }
+  return output;
+}
+
+/**
+ * Sort a list of fullpath refs into a list and then provide the first entry.
+ *
+ * The sort order will first find master ref, and then any semver ref, followed
+ * by the rest of the refs in the order provided.
```
gkalpak, line 33: > followed by the rest of the refs in the order provided

This is not necessarily true. According to the spec the sorting must be stable, but implementations are free to choose the sorting algorithm, which means the original relative order is not necessarily preserved. If order of the rest of the refs matter, consider sorting lexicographically.

gkalpak (approved):

devversion:
@josephperrott Just a heads up in case you didn't see my comment. I'm proposing that we rename the file to something more obvious. `get-refs` seems somewhat ambiguous to me. Not sure how others feel about this 😄

tools/utils/get-refs.js
josephperrott: It looks like as of ES2019, `Array.sort` is not a stable sort [see here](https://github.com/tc39/ecma262/pull/1340)

It looks like this was added and is in all Node versions 12+, so our usage should actually be fine to consider it as a stable sort.

I did not know this going into and only learned it as I was researching it based on your comment.

tools/utils/get-refs.js
josephperrott: There are instances where we are running a command just to find out if it passes or fails, not wanting to get the stdout or stderr.

I updated it to allow for a call to suppress the stderr, otherwise it will be logged.

tools/utils/get-refs.js
gkalpak: Oh, I just realized that stability in sorting means "equal" elements retain their original relative order. I thought "stable" meant deterministic 🤦‍♂️ 
I learned something too :tada:

PR #36502, opened 2020-04-08
build: remove `fullTemplateTypeCheck` from aio tsconfig

alan-agius4:
`fullTemplateTypeCheck` is no longer required since we now use `strictTemplates` which is a superset of the former option.

---

Follow up on https://github.com/angular/angular/commit/04f61c0c3ee9b3616b1719b4ae9c3eb097698218#r38354112

//cc @IgorMinar

### DIFF
--- a/aio/tsconfig.json
+++ b/aio/tsconfig.json
@@ -37,7 +37,6 @@
   ],
   "angularCompilerOptions": {
     "disableTypeScriptVersionCheck": true,
-    "fullTemplateTypeCheck": true,
     "strictInjectionParameters": true,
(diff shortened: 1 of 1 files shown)


gkalpak (approved):
✨

Commit message typo: `requires` --> `required`
(Also, it would be useful to mention this is a follow-up to https://github.com/angular/angular/commit/04f61c0c3ee9b3616b1719b4ae9c3eb097698218#r38354112 in the commit message.)

alan-agius4:
@gkalpak commit message updated.

PR #36505, opened 2020-04-08
build: ts-circular-dependencies tool should normalize golden

devversion:
Currently the golden output of the circular-deps tool is purely based
on the order of source files passed to the tool, and on the order and amount
of imports inside source files.

This is actually resulting in deterministic output as running the tool multiple
times without any changes to source files, results in the same output.

*Though* it seems like the tool is too strict and we can avoid a few unnecessary
golden changes if:

* A source file that is part of a cycle is imported earlier (in terms
of how the analyzer visits them). This could result in the cycle path
starting with a different source file. Resulting in unnecessary diff in the golden.
* Source files which are not part of a cycle are imported earlier
(in terms of how the analyzer visits them). This could result in moved
cycle paths in the golden if re-approved.

To fix this, we normalize the cycle path that serves as serializable data
structure for the text-based goldens. Since the paths represents a cycle,
the path can be shifted in a deterministic way so that cycles don't change
unnecessarily in the golden, and to simplify comparison of cycles.

Additionally, we sort the cycles in a deterministic way so that the golden
doesn't change unnecessarily as described in the second bullet point.


dev-infra/ts-circular-dependencies/golden.ts
```diff
@@ -44,6 +53,49 @@ export function compareGoldens(actual: Golden, expected: Golden) {
...
+  if (indexFirstNode === 0) {
+    return path;
+  }
+
+  // Move the determined first node (as of alphabetical order) to the start of a new
+  // path array. The nodes before the first node in the old path are then concatenated
+  // to the end of the new path. This is possible because the path represents a cycle.
+  return [...path.slice(indexFirstNode), ...path.slice(0, indexFirstNode)];
+}
+
```
petebacondarwin, line 98: This could be marginally simplified to avoid the `null` checks:

```ts
function normalizeCircularDependency(path: CircularDependency): CircularDependency {
  if (path.length <= 1) {
    return path;
  }

  let indexFirstNode = 0;
  let valueFirstNode = path[0];

  // Find a node in the cycle path that precedes all other elements
  // in terms of alphabetical order.
  for (let i = 1; i < path.length; i++) {
    const value = path[i];
    if (value.localeCompare(valueFirstNode, 'en') < 0) {
      indexFirstNode = i;
      valueFirstNode = value;
    }
  }

  // If the alphabetically first node is already at start of the path, just
  // return the actual path as no changes need to be made.
  if (indexFirstNode === 0) {
    return path;
  }

  // Move the determined first node (as of alphabetical order) to the start of a new
  // path array. The nodes before the first node in the old path are then concatenated
  // to the end of the new path. This is possible because the path represents a cycle.
  return [...path.slice(indexFirstNode), ...path.slice(0, indexFirstNode)];
}
```

petebacondarwin (approved):
LGTM - thanks for the quick turnaround @devversion !
One minor nit.

dev-infra/ts-circular-dependencies/golden.ts
devversion: Thx Pete! That looks perfect :+1:

<sup>I didn't realize the strict null check until CI failed</sup>

josephperrott (approved):
LGTM

PR #36507, opened 2020-04-08
fix(dev-infra): fix commit message validation in git worktrees

gkalpak:
Previously, the `pre-commit-validate` command (used in the `commit-msg` git hook) assumed that the commit message was stored in `.git/COMMIT_EDITMSG` file. This is usually true, but not when using [git worktrees](https://git-scm.com/docs/git-worktree), where `.git` is a file containing the path to the actual git directory.

This commit fixes it by taking advantage of the fact that git passes the actual path of the file holding the commit message to the `commit-msg` hook and husky exposes the arguments passed by git as `$HUSKY_GIT_PARAMS`.

NOTE:
We cannot use the environment variable directly in the `commit-msg` hook command, because environment variables need to be referenced differently on Windows (`%VAR_NAME%`) vs macOS/Linux (`$VAR_NAME`). Instead, we pass the name of the environment variable and the validation script reads the variable's value off of `process.env`.

##
This is an alternative to #36497.

### DIFF
--- a/dev-infra/commit-message/cli.ts
+++ b/dev-infra/commit-message/cli.ts
@@ -14,9 +14,18 @@ export function buildCommitMessageParser(localYargs: yargs.Argv) {
   return localYargs.help()
       .strict()
       .command(
-          'pre-commit-validate', 'Validate the most recent commit message', {},
-          () => {
-            validateFile('.git/COMMIT_EDITMSG');
+          'pre-commit-validate <commit-message-file|commit-message-file-env>',
+          'Validate the most recent commit message',
+          args => {
+            args.positional('commit-message-file|commit-message-file-env', {
+              type: 'string',
+              describe:
+                  'The path to a file containing the commit message to validate or the name of ' +
+                  'an environment variable that holds the path to such a file.',
+            });
+          },
+          args => {
+            validateFile(process.env[args.commitMessageFileEnv] || args.commitMessageFile);
           })
       .command(
           'validate-range', 'Validate a range of commit messages', {
--- a/dev-infra/commit-message/cli.ts
+++ b/dev-infra/commit-message/cli.ts
@@ -14,18 +14,29 @@ export function buildCommitMessageParser(localYargs: yargs.Argv) {
   return localYargs.help()
       .strict()
       .command(
-          'pre-commit-validate <commit-message-file|commit-message-file-env>',
-          'Validate the most recent commit message',
-          args => {
-            args.positional('commit-message-file|commit-message-file-env', {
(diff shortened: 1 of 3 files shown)


dev-infra/commit-message/cli.ts
```diff
@@ -14,9 +14,18 @@ export function buildCommitMessageParser(localYargs: yargs.Argv) {
...
+          args => {
+            args.positional('commit-message-file|commit-message-file-env', {
+              type: 'string',
+              describe:
+                  'The path to a file containing the commit message to validate or the name of ' +
+                  'an environment variable that holds the path to such a file.',
+            });
+          },
+          args => {
+            validateFile(process.env[args.commitMessageFileEnv] || args.commitMessageFile);
```
petebacondarwin, line 28: For the reference of anyone reviewing this. @gkalpak has use positional arg aliases. See https://github.com/yargs/yargs/blob/master/docs/advanced.md#positional-argument-aliases

petebacondarwin (approved):
Personally, if it is not being used then I would just get rid of the `commitMessageFile` argument altogether. Or alternatively use named options to the arg rather than positional so that we don't have to worry about a filename happening to match an environment variable name.

Otherwise LGTM.

dev-infra/commit-message/cli.ts
```diff
@@ -14,9 +14,18 @@ export function buildCommitMessageParser(localYargs: yargs.Argv) {
...
           })
```
josephperrott, line 29: Can we use a flag instead of a positional here, and then set a default otherwise.

So it would default to `.git/COMMIT_EDITMSG`, unless a flag is provided.

And we should split the valid flags into `file` and `file-path-env-variable` (or something similar).  We should be able to have `yargs` set them as conflicting so that both can't be used at the same time.

```suggestion
      .command(
          'pre-commit-validate',
          'Validate the most recent commit message',
          {
            'file': {
              type: 'string',
              conflicts: ['file-env-variable'],
              description: 'The path of the commit message file.'
            },
            'file-env-variable': {
              type: 'string',
              conflicts: ['file'],
              description: 'The key of the environment variable for the path commit message file.',
              coerce: arg => {
                const file = process.env[arg];
                if (!file) {
                  throw Error(`Provided environment variable "${arg}" was not found.`);
                }
                return file;
              }
            }
          },
          args => {
            const file = args.file || process.env[args['file-env-variable']] || `.git/COMMIT_EDITMSG`;
            validateFile(file);
          })
```

gkalpak: Changed. PTAL

josephperrott (approved):
LGTM

PR #36512, opened 2020-04-08
fix(core): pipes injecting viewProviders when used on a component host node

### LINKED ISSUE #36146: In Ivy, pipes uses providers from a host component where ViewEngine did not

crisbeto:
The flag that determines whether something should be able to inject from `viewProviders` is opt-out and the pipes weren't opted out, resulting in them being able to see the `viewProviders` if they're placed on a component host node.

Fixes #36146.

### DIFF
--- a/packages/core/src/render3/pipe.ts
+++ b/packages/core/src/render3/pipe.ts
@@ -11,6 +11,7 @@ import {PipeTransform} from '../change_detection/pipe_transform';
 import {setInjectImplementation} from '../di/injector_compatibility';
 
 import {getFactoryDef} from './definition';
+import {setIncludeViewProviders} from './di';
 import {store, ɵɵdirectiveInject} from './instructions/all';
 import {PipeDef, PipeDefList} from './interfaces/definition';
 import {HEADER_OFFSET, LView, TVIEW} from './interfaces/view';
@@ -47,7 +48,12 @@ export function ɵɵpipe(index: number, pipeName: string): any {
 
   const pipeFactory = pipeDef.factory || (pipeDef.factory = getFactoryDef(pipeDef.type, true));
   const previousInjectImplementation = setInjectImplementation(ɵɵdirectiveInject);
+
+  // DI for pipes is supposed to behave like directives when placed on a component
+  // host node, which means that we have to disable access to `viewProviders`.
+  const previousIncludeViewProviders = setIncludeViewProviders(false);
   const pipeInstance = pipeFactory();
+  setIncludeViewProviders(previousIncludeViewProviders);
   setInjectImplementation(previousInjectImplementation);
   store(tView, getLView(), index, pipeInstance);
   return pipeInstance;
--- a/packages/core/src/render3/di.ts
+++ b/packages/core/src/render3/di.ts
@@ -71,7 +71,7 @@ import {stringifyForError} from './util/misc_utils';
  */
 let includeViewProviders = true;
 
-function setIncludeViewProviders(v: boolean): boolean {
+export function setIncludeViewProviders(v: boolean): boolean {
   const oldValue = includeViewProviders;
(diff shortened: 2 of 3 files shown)


AndrewKushnir (commented):
Hi @crisbeto, thanks for the fix 👍 

If it's not too much work, could you please move all clang changes to a separate commit? That'd help with review and also with keeping git history more verbose (for ex. if you `git blame` later, you'll see that a certain line was changed due to code style update).

I was also wondering if the code in the `ɵɵpipe` instruction can be unified (i.e. common logic be extracted into a function and used in 2 places) with the `getNodeInjectable` function logic in the [di.ts](https://github.com/angular/angular/blob/master/packages/core/src/render3/di.ts#L535-L563) file (to also enter/leave DI, check for circular deps and have a try-catch block around the factory invocation)? I might be missing something (and it might not be a good idea overall), so I'm curious what's your opinion on that.

Thank you.

crisbeto:
Regarding the formatting changes: I'm not sure how to do that easily without reverting the clang version, making my changes and then redoing them.

As for extracting it into a separate function, from what I can tell we can only share the `setIncludeViewProviders` and `setInjectImplementation` calls which are fairly basic.

crisbeto:
I've split up the formatting changes. The actual code change is in the fixup commit.

AndrewKushnir (approved):
Thanks for splitting the format change and the actual change 👍 

The changes LGTM, just left 1 nit (to add a comment).

kara (approved):
LGTM, but FYI looks like both commits are fixup commits, so you'll want to squash them

packages/core/src/render3/pipe.ts
```diff
@@ -47,7 +48,12 @@ export function ɵɵpipe(index: number, pipeName: string): any {
 
   const pipeFactory = pipeDef.factory || (pipeDef.factory = getFactoryDef(pipeDef.type, true));
   const previousInjectImplementation = setInjectImplementation(ɵɵdirectiveInject);
+
+  // DI for pipes is supposed to behave like directives, when place on a component
```
kara, line 52:
```suggestion
  // DI for pipes is supposed to behave like directives when placed on a component
```

AndrewKushnir:
[Presubmit](http://test/OCL:306982250:BASE:306982268:1587099223696:7e00f8e7)

JoostK:
Note that this doesn't technically fix #36146, as the reported issue expected that providers from a host component are not visible to pipes. The change that is being made here allows the issue to be solved by using `viewProviders` instead of `providers` in the host component, but it's still a change in behaviour compared to VE. This should be documented in the Ivy breaking changes doc.

crisbeto:
I added an entry to the breaking changes doc a few days ago.

PR #36523, opened 2020-04-08
build: update matching regex for bazel stamping

josephperrott:
Previously, the bazel stamping regex only matched on versions
0-9 for major and minor numbers, this update allows for matching
on any number for major, minor or patch.

### DIFF
--- a/tools/bazel_stamp_vars.js
+++ b/tools/bazel_stamp_vars.js
@@ -43,7 +43,7 @@ console.log(`BUILD_SCM_LOCAL_CHANGES ${LOCAL_CHANGES}`);
 // This will ignore non-version tags which would break unit tests expecting a valid version
 // number in the package headers
 const BUILD_SCM_VERSION_RAW =
-    _exec(`git describe --match [0-9].[0-9].[0-9]* --abbrev=7 --tags HEAD`);
+    _exec(`git describe --match [0-9]*.[0-9]*.[0-9]* --abbrev=7 --tags HEAD`);
 
 // Reformat `git describe` version string into a more semver-ish string
 //   From:   5.2.0-rc.0-57-g757f886


kara (approved):
LGTM

tools/bazel_stamp_vars.js
```diff
@@ -43,7 +43,7 @@ console.log(`BUILD_SCM_LOCAL_CHANGES ${LOCAL_CHANGES}`);
 // This will ignore non-version tags which would break unit tests expecting a valid version
 // number in the package headers
 const BUILD_SCM_VERSION_RAW =
-    _exec(`git describe --match [0-9].[0-9].[0-9]* --abbrev=7 --tags HEAD`);
+    _exec(`git describe --match [0-9]*.[0-9]*.[0-9]* --abbrev=7 --tags HEAD`);
```
devversion, line 46: Nit so that it matches with [pre-check](https://github.com/angular/angular/blob/cab0c9babcc2732fd79e08a4e1cf7a12a44cdd74/scripts/release/pre-check#L12)

```suggestion
    _exec(`git describe --match [0-9]+.[0-9]+.[0-9]+ --abbrev=7 --tags HEAD`);
```

josephperrott:
@kara I believe the group name is `global-approvers` rather than `approval`

Might be something we should change to make it read better.

devversion (approved):
LGTM.
<sup>lol, had to laugh because it still re-sent the request</sup>

kara:
@josephperrott Gah, thanks for letting me know. Clearly haven't had my coffee yet. Fixed it, though I see @devversion has reviewed it anyway, so we're good to go.

tools/bazel_stamp_vars.js
josephperrott: I could not tell you why, but it seems that the `--match` regex matching doesn't support `+`?

devversion: Yeah, looks like it. http://man7.org/linux/man-pages/man7/glob.7.html. It's fine with `*` too.

PR #36527, opened 2020-04-08
build: run pre-check before publishing

josephperrott:
Previously, our process included running the pre-check script before
releasing.  With our new publishing process this was dropped.  This
change adds in automatically executing this check before publish for
both next and latest

### DIFF
--- a/scripts/release/pre-check
+++ b/scripts/release/pre-check
@@ -3,13 +3,9 @@
 # Runs the Bazel workspace status command in order to ensure that the version placeholder
 # will be replaced with the proper version name when building the release output.
 
-# Ensure that the stamping script is executed within the project directory.
-cd $(dirname ${0})/../..
-
-# The "0.0.0-PLACEHOLDER" is based on the value of the "BUILD_SCM_VERSION" status variable.
 versionName=$(node tools/bazel_stamp_vars.js | grep -e "BUILD_SCM_VERSION" | cut -d " " -f2)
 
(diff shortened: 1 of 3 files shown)


scripts/release/pre-check
```diff
@@ -3,27 +3,13 @@
...
 # will be replaced with the proper version name when building the release output.
 
-# Ensure that the stamping script is executed within the project directory.
-cd $(dirname ${0})/../..
-
-# The "0.0.0-PLACEHOLDER" is based on the value of the "BUILD_SCM_VERSION" status variable.
 versionName=$(node tools/bazel_stamp_vars.js | grep -e "BUILD_SCM_VERSION" | cut -d " " -f2)
 
-if [[ ! ${versionName} =~ ^[0-9]+\.[0-9]+\.[0-9]+(-(alpha|beta|rc).[0-9]+)?$ ]]; then
+if [[ ! ${versionName} =~ ^[0-9]+\.[0-9]+\.[0-9]+(-(alpha|beta|rc|next).[0-9]+)?$ ]]; then
```
IgorMinar, line 8:
```suggestion
if [[ ! ${versionName} =~ ^[0-9]+\.[0-9]+\.[0-9]+(-(rc|next).[0-9]+)?$ ]]; then
```

IgorMinar: we currently don't have `alpha` or `beta` releases. if we need them in the future, we can add them.

IgorMinar (approved):
lgtm otherwise! thanks

PR #36520, opened 2020-04-08
Reformat of the compiler/compiler-cli codebase

alxhub:
This PR bundles two commits, constituting a reformat of packages/compiler and packages/compiler-cli with the new clang-format.

PR #36549, opened 2020-04-09
style(core): reformat all of packages/core with new clang format version

kara:
This commit updates the formatting of all packages/core/*
files at once to comply with the new version of clang.
Doing it all at once this way will cause some short-term
rebasing of existing core PRs, but will prevent head-
aches when reviewing PRs down the line (because we
will not have to filter out the noise of unrelated
formatting changes as these files are touched in the
future).


IgorMinar (commented):
Can we please do this ones across the whole repo and not just within packages/core as the entire repo is affected in the same way?

Additionally we should consider writing a script that would go over all the existing PRs that currently don't have a merge conflict and:
-  check them out
- rebase them on the latest master
- reformat them
- add a "style:" commit with the reformatting diff
-and force push to the PR branch

That way we know that all the existing PRs that are currently in a good enough shape, will remain in a good enough shape.

This will still leave us with a lot of PRs that currently have a merge conflict and can't be cleanly rebased, in which case we need to come up with a different reformatting strategy for these....

kara:
@IgorMinar In the Monday sync, we agreed to do the repo in smaller pieces (see https://github.com/angular/angular/pull/36520 for another example). Let's chat offline.

IgorMinar (approved):
We discussed this with @kara and @josephperrott  in person and agreed that the best course of action is to merge this PR and schedule follow up work to reformat the rest of the monorepo.

@josephperrott is looking into possibility to update the PRs automatically in the future, but we will not block this PR on that.

@josephperrott is also going to look into using prettier instead.

IgorMinar (approved):

IgorMinar:
one more update: we decided to hold off on this PR until tomorrow, while @josephperrott and I are investigating feasibility of switching over to prettier and reformatting the whole repo with it.

kara:
Update: we're going to switch to Prettier, so this PR is out of date

PR #36550, opened 2020-04-09
fix(ngcc): recognize enum declarations emitted in JavaScript

### LINKED ISSUE #35584: ES6 template literal expressions not working past v9
# 🐞 bug report

### Is this a regression?

 Yes, the previous version in which this bug was not present was: ^8

### Description

Using a template literal expression used to work until Angular 9, for as long as the template lived in the TS source not in HTML (_obviously_).

Since Angular 9 that does not work any longer in one scenario. I am mostly focusing on library builds here:

**Working**
- build library with Angular 9 Ivy & serve the Angular 9 application
- build library with Angular 8 View Engine & serve the Angular 8 application
- build library with Angular 8 View Engine & serve the Angular 9 application (`ngcc`)

**Not working**
- build library with Angular 9 View Engine (`ng build lib --prod`) & serve the Angular 9 application (`ngcc`)

## 🔬 Minimal Reproduction

In any Angular workspace create a library and do the following:

```ts
const enum Demo {
  First
}

@Component({
  template: `
    Static variable: ${ Demo.First  }
  `
})
export class DemoComponent {}
```

## 🔥 Exception or Error
```
error NG1010: template must be a string
```

## 🌍  Your Environment

**Angular Version:**
```
Angular CLI: 9.0.3
Node: 12.15.0
OS: darwin x64

Angular: 9.0.2
... router
Ivy Workspace: Yes

@angular-devkit/architect          0.900.3
@angular-devkit/build-angular      0.900.3
@angular-devkit/build-ng-packagr   0.900.3
@angular-devkit/build-optimizer    0.900.3
@angular-devkit/build-webpack      0.900.3
@angular-devkit/core               9.0.3
@angular-devkit/schematics         9.0.3
@angular/cdk                       9.0.1
@angular/cli                       9.0.3
@angular/material                  9.0.1
@ngtools/webpack                   9.0.3
@schematics/angular                9.0.3
@schematics/update                 0.900.3
ng-packagr                         9.0.1
```

**Anything else relevant?**

I am not 100% sure this was ever officially supported, even though I would have hoped so, since it is native ES6 feature. If it wasn't supported, please close the issue.

Interesting observation is, that View Engine compiled library from Angular 8 is correctly processed by `ngcc`, while View Engine from Angular 9 surprisingly fails.

---

**EDIT**: Seems like an easy reproduction mentioned above is not reproducible any more. I can still reproduce it, if I use literal expressions while binding to input, e.g.: `[title]="${variable}"`.

Reproduction:
- Clone: https://github.com/klemenoslaj/template-literal-expressions-issue
- Execute: `yarn && yarn reproduce`

Resulting in error:

```
dist/example/fesm2015/example.js:29:19 - error NG1010: template must be a string

29         template: `
                     ~
30     <button [title]="${ButtonType.Button}">Button</button>
   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
31   `
   ~~~
dist/example/fesm2015/example.js:39:24 - error NG6001: The class 'ExampleComponent' is listed in the declarations of the NgModule 'ExampleModule', but is not a directive, a component, or a pipe.
```

JoostK:
An enum declaration in TypeScript code will be emitted into JavaScript
as a regular variable declaration, with the enum members being declared
inside an IIFE. For ngcc to support interpreting such variable
declarations as enum declarations with its members, ngcc needs to
recognize the enum declaration emit structure and extract all member
from the statements in the IIFE.

This commit extends the `ConcreteDeclaration` structure in the
`ReflectionHost` abstraction to be able to capture the enum members
on a variable declaration, as a substitute for the original
`ts.EnumDeclaration` as it existed in TypeScript code. The static
interpreter has been extended to handle the extracted enum members
as it would have done for `ts.EnumDeclaration`.

Fixes #35584
Resolves FW-2069

### DIFF
--- a/packages/compiler-cli/ngcc/src/host/esm2015_host.ts
+++ b/packages/compiler-cli/ngcc/src/host/esm2015_host.ts
@@ -8,13 +8,14 @@
 
 import * as ts from 'typescript';
 
-import {ClassDeclaration, ClassMember, ClassMemberKind, ConcreteDeclaration, CtorParameter, Declaration, Decorator, isDecoratorIdentifier, KnownDeclaration, reflectObjectLiteral, TypeScriptReflectionHost, TypeValueReference,} from '../../../src/ngtsc/reflection';
+import {ClassDeclaration, ClassMember, ClassMemberKind, ConcreteDeclaration, CtorParameter, Declaration, Decorator, EnumMember, isDecoratorIdentifier, KnownDeclaration, reflectObjectLiteral, SpecialDeclarationKind, TypeScriptReflectionHost, TypeValueReference} from '../../../src/ngtsc/reflection';
 import {isWithinPackage} from '../analysis/util';
 import {Logger} from '../logging/logger';
 import {BundleProgram} from '../packages/bundle_program';
 import {findAll, getNameText, hasNameIdentifier, isDefined, stripDollarSuffix} from '../utils';
 
 import {ClassSymbol, isSwitchableVariableDeclaration, ModuleWithProvidersFunction, NgccClassSymbol, NgccReflectionHost, PRE_R3_MARKER, SwitchableVariableDeclaration} from './ngcc_host';
+import {stripParentheses} from './utils';
 
 export const DECORATORS = 'decorators' as ts.__String;
 export const PROP_DECORATORS = 'propDecorators' as ts.__String;
@@ -344,13 +345,28 @@ export class Esm2015ReflectionHost extends TypeScriptReflectionHost implements N
   getDeclarationOfIdentifier(id: ts.Identifier): Declaration|null {
     const superDeclaration = super.getDeclarationOfIdentifier(id);
 
+    // If no declaration was found or it's an inline declaration, return as is.
+    if (superDeclaration === null || superDeclaration.node === null) {
+      return superDeclaration;
+    }
+
+    // If the declaration already has traits assigned to it, return as is.
+    if (superDeclaration.known !== null || superDeclaration.identity !== null) {
+      return superDeclaration;
+    }
+
     // The identifier may have been of an additional class assignment such as `MyClass_1` that was
     // present as alias for `MyClass`. If so, resolve such aliases to their original declaration.
-    if (superDeclaration !== null && superDeclaration.node !== null &&
-        superDeclaration.known === null) {
-      const aliasedIdentifier = this.resolveAliasedClassIdentifier(superDeclaration.node);
-      if (aliasedIdentifier !== null) {
-        return this.getDeclarationOfIdentifier(aliasedIdentifier);
+    const aliasedIdentifier = this.resolveAliasedClassIdentifier(superDeclaration.node);
+    if (aliasedIdentifier !== null) {
+      return this.getDeclarationOfIdentifier(aliasedIdentifier);
+    }
+
+    // Variable declarations may represent an enum declaration, so attempt to resolve its members.
+    if (ts.isVariableDeclaration(superDeclaration.node)) {
+      const enumMembers = this.resolveEnumMembers(superDeclaration.node);
+      if (enumMembers !== null) {
+        superDeclaration.identity = {kind: SpecialDeclarationKind.DownleveledEnum, enumMembers};
       }
     }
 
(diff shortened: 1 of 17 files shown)


packages/compiler-cli/ngcc/src/host/esm2015_host.ts
```diff
@@ -1763,6 +1775,140 @@ export class Esm2015ReflectionHost extends TypeScriptReflectionHost implements N
...
+    if (!ts.isBlock(block) && !ts.isSourceFile(block)) return null;
+
+    const declarationIndex = block.statements.findIndex(statement => statement === variableStmt);
+    if (declarationIndex === -1 || declarationIndex === block.statements.length - 1) return null;
+
+    const subsequentStmt = block.statements[declarationIndex + 1];
+    if (!ts.isExpressionStatement(subsequentStmt)) return null;
+
+    const iife = stripParentheses(subsequentStmt.expression);
+    if (!ts.isCallExpression(iife)) return null;
```
petebacondarwin, line 1810: I think we should check that the first argument to the call expression is the variable being declared, i.e. of the form `Enum || (Enum = {})`, where `Enum` is the name of the declared variable.

Otherwise there is a small chance that we get a false positive.

packages/compiler-cli/ngcc/src/host/esm2015_host.ts
```diff
@@ -1763,6 +1775,140 @@ export class Esm2015ReflectionHost extends TypeScriptReflectionHost implements N
...
+    if (fn.parameters.length !== 1) return null;
+
+    const enumName = fn.parameters[0].name;
+    if (!ts.isIdentifier(enumName)) return null;
+
+    const enumMembers: EnumMember[] = [];
+    for (const statement of fn.body.statements) {
+      const enumMember = this.reflectEnumMember(enumName, statement);
+      if (enumMember === null) {
+        return null;
```
petebacondarwin, line 1840: So if any of the statements in this function do not have the correct form, we bail out completely?
Is there a chance that there might be some innocuous additional statement in there? I guess it is unlikely.

packages/compiler-cli/ngcc/src/host/esm2015_host.ts
```diff
@@ -1763,6 +1775,140 @@ export class Esm2015ReflectionHost extends TypeScriptReflectionHost implements N
...
+    // Check for the `Enum[Enum["ValueA"] = 0] = "ValueA"` case.
+    if (!ts.isBinaryExpression(expression) ||
+        expression.operatorToken.kind !== ts.SyntaxKind.EqualsToken ||
+        !ts.isElementAccessExpression(expression.left)) {
+      return null;
+    }
+
+    // Verify that the outer assignment corresponds with the enum declaration.
+    const enumIdentifier = expression.left.expression;
+    if (!ts.isIdentifier(enumIdentifier) || enumIdentifier.text !== enumName.text) return null;
```
petebacondarwin, line 1880: This block is basically repeated in `reflectEnumMemberAssignment()`.
If you pull this out into its own method, then it could be checked up front at the top of this method and that would avoid a) duplicating the code and b) running the code twice for numeric valued enums.

packages/compiler-cli/ngcc/src/host/esm2015_host.ts
```diff
@@ -2106,3 +2252,7 @@ function getNonRootPackageFiles(bundle: BundleProgram): ts.SourceFile[] {
   return bundle.program.getSourceFiles().filter(
       f => (f !== rootFile) && isWithinPackage(bundle.package, f));
 }
+
+export function stripParentheses(node: ts.Node): ts.Node {
+  return ts.isParenthesizedExpression(node) ? node.expression : node;
+}
```
petebacondarwin, line 2258: This function already exists in `esm5_host.ts`, no?

petebacondarwin (changes requested):
Only had time to check the ngcc source (and not the tests).

packages/compiler-cli/ngcc/src/host/esm2015_host.ts
JoostK: I did initially check for it, but decided to remove it. The UMD/CJS formats use `exports.Enum || (exports.Enum = {})` so there's slight variations that we'd need to account for.

packages/compiler-cli/ngcc/src/host/esm2015_host.ts
JoostK: I honestly don't know whether there's a possibility of innocuous statements. In the spirit of preventing false positives I decided to bail when there's an unrecognized statement.

packages/compiler-cli/ngcc/src/host/esm2015_host.ts
JoostK: Yes, this check is the same as in `reflectEnumMemberAssignment` but it checks a different identifier.

Given `Enum[Enum["ValueA"] = 0] = "ValueA"`, this check verifies the outer `Enum` whereas `reflectEnumMemberAssignment` check verifies the inner `Enum` identifier. Or are you suggesting that the outer identifier check already occurs within the first call to `reflectEnumMemberAssignment`, such that we don't have to repeat it? If we were to do that, I think we'd need to know why the first invocation of `reflectEnumMemberAssignment` failed to match.

packages/compiler-cli/ngcc/src/host/esm2015_host.ts
JoostK: It does! Perhaps now is the time to pull it out into a `ast_utils.ts` file.

packages/compiler-cli/ngcc/src/host/esm2015_host.ts
petebacondarwin: Looks like it is checking the exact same `expression` to me...

petebacondarwin: I would imagine we would do something like:

```
if not `expression` matches enum name
  return null
if `x` of `Enum[x]` is a string
  return processed RHS
if not `x` of `Enum[x]` is a binary expression
  return null
if not `x` matches enum name
  return null
return processed x RHS
```

packages/compiler-cli/ngcc/src/host/esm2015_host.ts
```diff
@@ -1862,56 +1863,101 @@ export class Esm2015ReflectionHost extends TypeScriptReflectionHost implements N
 
     const expression = statement.expression;
 
-    // Check for the `Enum["ValueA"] = "a"` case.
-    const assignment = this.reflectEnumMemberAssignment(enumName, expression);
-    if (assignment !== null) {
+    // Check for the `Enum["ValueA"] = "a";` case.
```
petebacondarwin, line 1866:
```suggestion
    // Check for `Enum[...] = "...";`
```

petebacondarwin (approved):
Marvellous - much easier to understand. Even at the end of the day!

packages/compiler-cli/src/ngtsc/reflection/src/host.ts
```diff
@@ -434,6 +450,12 @@ export interface BaseDeclaration<T extends ts.Declaration = ts.Declaration> {
 export interface ConcreteDeclaration<T extends ts.Declaration = ts.Declaration> extends
     BaseDeclaration<T> {
   node: T;
+
+  /**
+   * If the declaration is reflected to represent an enum declaration, all its members will be
+   * stored here. This will be null for declarations of type `ts.EnumDeclaration`, where
```
alxhub, line 456: This comment is incomplete!

The weirdness of this situation (that a real `ts.EnumDeclaration` doesn't need to be represented in this way) is a bit funny.

I think we should refine the way we represent this a little bit to make this a little less weird. Instead of `enumMembers` directly on `ConcreteDeclaration`, we can introduce a new `SpecialDeclarationIdentity` type with a `kind` enum. There will be one possible value (`SpecialEnum` maybe?) with `enumMembers: EnumMember[]`.

I think this approach will make it a lot more clear in the model that this is _one_ way of representing enums, but not the only way.

Alternatively, we could introduce a true `EnumDeclaration` subtype of `ConcreteDeclaration`, with `members` that may or may not have initializers, and reflect _all_ enums (whether from JS or `ts.EnumDeclaration`) that way. This is most in line with the spirit of the `ReflectionHost`, but might be too heavy-handed for this particular challenge.

alxhub (approved):
👍 this looks awesome

alxhub:
[Presubmit](http://test/OCL:308854806:BASE:308854843:1588096471736:89d90dec)

packages/compiler-cli/src/ngtsc/partial_evaluator/src/interpreter.ts
```diff
@@ -225,10 +225,12 @@ export class StaticInterpreter {
     }
     if (decl.known !== null) {
       return resolveKnownDeclaration(decl.known);
-    } else if (
-        decl.node !== null && decl.identity !== null &&
-        decl.identity.kind === SpecialDeclarationKind.DownleveledEnum) {
-      return this.getResolvedEnum(decl.node, decl.identity.enumMembers, context);
+    } else if (decl.node !== null) {
```
alxhub, line 228: Ahh, I think we should refactor this a bit. In general these "trap" `if` statements scare me.

What happens here is that this branch becomes true if `decl.node !== null`, even though the full case we intended to check here is really that and `decl.identity !== null`. This means that no other branches of the outer `if` will be checked even if `identity` IS null.

Instead, let's create a type guard function `isConcreteDeclaration` and use that to write the full condition in the outer `if` directly.

alxhub (approved):

AndrewKushnir:
@JoostK @alxhub it looks like a lot of targets in g3 presubmit failed with the following error:
```
packages/compiler-cli/src/ngtsc/partial_evaluator/src/interpreter.ts:229:36 - error TS2339: Property 'identity' does not exist on type 'Declaration<Declaration>'.
  Property 'identity' does not exist on type 'InlineDeclaration'.

229         decl.node !== null && decl.identity !== null &&
                                       ~~~~~~~~
packages/compiler-cli/src/ngtsc/partial_evaluator/src/interpreter.ts:230:14 - error TS2339: Property 'identity' does not exist on type 'Declaration<Declaration>'.
  Property 'identity' does not exist on type 'InlineDeclaration'.

230         decl.identity.kind === SpecialDeclarationKind.DownleveledEnum) {
                 ~~~~~~~~
packages/compiler-cli/src/ngtsc/partial_evaluator/src/interpreter.ts:231:51 - error TS2339: Property 'identity' does not exist on type 'Declaration<Declaration>'.
  Property 'identity' does not exist on type 'InlineDeclaration'.

231       return this.getResolvedEnum(decl.node, decl.identity.enumMembers, context);
                                                      ~~~~~~~~             
```
Could you please have a look at this issue when you get a chance?

(I will remove the "merge" label for now)

JoostK:
@AndrewKushnir Thanks Andrew, we spotted the same. The problem should have been fixed already, but it does need a new presubmit.

AndrewKushnir:
@JoostK got it, thanks! I've started a new [g3 presubmit](http://test/OCL:308903090:BASE:308903122:1588110947139:aa840389) and will let you know.

PR #36552, opened 2020-04-10
Use prettier for formatting

josephperrott:


docs/DEVELOPER.md
```diff
@@ -90,9 +90,10 @@ All the tests are executed on our Continuous Integration infrastructure. PRs can
 merged if the code is formatted properly and all tests are passing.
 
 <a name="clang-format"></a>
+<a name="code-formatting"></a>
```
IgorMinar, line 93: thanks for adding this!

modules/benchmarks/src/expanding_rows/expanding_row.ts
```diff
@@ -95,14 +109,15 @@ export interface ExpandingRowHostBase {
...
   styles: [expanding_row_css],
-  template: `
-  <div #expandingRowMainElement
-       class="cfc-expanding-row"
-       cdkMonitorSubtreeFocus
-       [attr.tabindex]="isExpanded ? '0' : '-1'"
-       [class.cfc-expanding-row-has-focus]="isFocused"
-       [class.cfc-expanding-row-is-expanded]="isExpanded"
-       ve="CfcExpandingRow">
+  template: ` <div
```
IgorMinar, line 112: hmm. this is an example of string template (backtick string) reformatting that is not great for two reasons:
- it changes semantic meaning of the string because of whitespace sensitivity of html and Angular templates
- the current formatting rule makes the code less readable (potentially subjective)

it would be better not to reformat template strings at all and deal with this after the fact.

modules/benchmarks/src/largetable/ng2/table.ts
```diff
@@ -16,21 +16,30 @@ let trustedGreyColor: SafeStyle;
...
   selector: 'largetable',
-  template: `<table><tbody>
-    <tr *ngFor="let row of data; trackBy: trackByIndex">
-      <td *ngFor="let cell of row; trackBy: trackByIndex" [style.backgroundColor]="getColor(cell.row)">
-      {{cell.value}}
-      </td>
-    </tr>
-  </tbody></table>`,
+  template: `<table
+    ><tbody>
```
IgorMinar, line 20: oufff

modules/benchmarks/src/expanding_rows/expanding_row.ts
josephperrott: Looks like this is a feature that will be coming very soon, but is not possible at the moment.

We need for https://github.com/prettier/prettier/pull/7875 to land.

.clang-format
```diff
@@ -1,3 +0,0 @@
-Language:        JavaScript
```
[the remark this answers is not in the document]
josephperrott: I think it makes sense in the first commit as that is where all of the clang references are being removed from config/docs.

modules/benchmarks/src/expanding_rows/expanding_row.ts
bakkot: > it changes semantic meaning of the string because of whitespace sensitivity of html and Angular templates

Drive-by comment: Prettier does change the whitespace here, but it shouldn't affect the semantic interpretation for HTML or Angular templates in particular. (The relevance is usually just whether or not there is whitespace between tags, not what that whitespace is, and Prettier should always ensure that presence or absence of such whitespace is preserved.) If there's a place where it's actually changing the interpretation as an HTML or Angular template, Prettier would consider that a bug.

gkalpak: Another drive-by-comment: If nothing else, whether HTML whitespace is important depends on the value (default or manually set) of the `white-space` CSS property and this is not something that can be statically evaluated.

petebacondarwin: Actually it is also reliant on the [`preserveWhitespaces` property of the `@Component()` decorator](https://angular.io/api/core/Component#preserveWhitespaces) (and to a certain extent the same in the `tsconfig.json` for the project). It is definitely not acceptable for Prettier to be touching whitespace within backticked strings. In fact if it is doing that then I would argue that is a significant bug within Prettier.

bakkot: > It is definitely not acceptable for Prettier to be touching whitespace within backticked strings.

It doesn't do so for arbitrary template literals, only for very specific ones, like the `template` property of `@Component`s (apparently implemented [by request](https://github.com/prettier/prettier/pull/4361) of multiple people at ng-conf). I agree this feature is dangerous, which is why I just added an option to turn it off (should be in 2.1, whenever that's released), but it is quite popular. /shrug

IgorMinar: Clarification:

- html is whitespace sensitive, but it's nuanced and as was mentioned here and in most cases it's still ok to format html as long as some rules are followed.
- Angular template compiler by default doesn't preserve all the whitespace and triee to strike the right balance between removing unnecessary whitespace and keep that which would otherwise break Dom rendering
- Angular i18n, which extracts and merges translation strings from/to Angular's templates is fully whitespace sensitive, so reformatting templates fully invalidates any translation strings extracted from that template (which typically means that the translator needs to retranslate the string). This is a bug and we have a fix in works that will be rolled out in one of the upcoming major versions (we are sorting out the migration strategy got this change and are also busy with other things, but this is definitely coming as it's a highly requested feature at Google)

As I said, whitespace sensitivity in Angular is nuanced. 🙃

What does this mean?

For now we should opt-out all localized templates from any formatting. We simply can't format them because it would be a breaking change requiring tests to be updated and that should not happen in a major repo-wide refactoring like this one. We could consider it as a follow-up task if we are confident that that would not change the original meaning of tests - in some cases it would, so we must be careful.

We can format all the other non-i18n-ed templates as long as we follow the nuanced html whitespace sensitivity rules. The main concern here is that many of our tests use templates with whitespace that results in HTML nodes to be created once the template is parsed into AST. These nodes then use up slots in indices and offset other nodes. And our tests use the indices to verify that the right thing appeared in the right place in DOM or the template AST. If we were to reformat the templates without this consideration we could break tests, or worse  change the meaning of a test and still have it accidentally pass. And only later when a human reads the test we'll wonder how it worked at all or what was it trying to test in the first place.

This second problem is even more dangerous then the first but can be avoided if we ensure that no matter how we reformat the templates in the end we will always preserve node indices in the template AST/DOM. I don't believe that this is the case with prettier today but it could be implemented before we actually reformat the repo.

I'm sure there is a bit more to all of this than what I captured but the remaining issues are likely insignificant and should not result in large-scale issues.

josephperrott:
Closing as we are continuing to use `clang-format` at this time.

PR #36562, opened 2020-04-10
feat(router): allow using single route resolver instead of map of resolvers

### LINKED ISSUE #25430: Allow Route['resolve'] to be a single value/observable/promise
## I'm submitting a...

<pre><code>
[ ] Bug report  
[ ] Performance issue
[x] Feature request
[ ] Other... Please describe:
</code></pre>

## Current behavior
`Route['resolve']` property now must be an object containing resolvers.

## Expected behavior
Allow `Route['resolve']` to be a single resolver and store resolved value right into `route.snapshot.data`.

I think that adding a separate resolver class stored in a separate file for every single route just for fetching a data (and in most cases it's just a single call to a service method) is a perfect example of overengineering (sorry). Recently I came up with an interesting idea - use page component classes as route resolvers. Here is an example:
```ts
interface Post {
  id: number;
  title: string;
}

const routes: Route[] = [
  {
    path: 'post/:id',
    component: PostPageComponent,
    resolve: {
      post: PostPageComponent
    }
  }
];

@Component({
  selector: 'app-post-page',
  template: 'Title: {{ post.title }}'
})
class PostPageComponent implements OnInit, Resolve<Post> {
  post: Post;

  constructor(
    private posts: PostsService,
    private route: ActivatedRoute
  ) {}

  async resolve(route: ActivatedRouteSnapshot): Promise<Post> {
    return this.posts.get(route.params.id);
  }

  ngOnInit() {
    this.post = this.route.snapshot.data.post;
  }
}

@Injectable()
class PostsService {
  async get(id: Post['id']): Promise<Post> {
    // ...fetch post...
    return {id: 1, title: 'Foo'};
  }
}
```
So I just use `PostPageComponent` as resolver for post page data and it works great except one little nuance: I have to provide a name for resolved resource in route configuration (`post` key in `resolve` object) but I don't need it because I want `PostComponent` to fetch all the necessary data and put it straight into the `route.snapshot.data` object. Here is the example:
```ts
class PostPageComponent implements OnInit, Resolve<PostPageData> {
 // ...
  async resolve(route: ActivatedRouteSnapshot): Promise<PostPageData> {
    const author = this.authors.get(route.params.authorId);
    const post = this.posts.get(route.params.postId);
    return {
      author: await author,
      post: await post
    };
  }
  
  ngOnInit() {
    const {author, post} = this.route.snapshot.data as PostPageData;
    this.author = author;
    this.post = post;
  }
  // ...
}
```

Of course I can rename `post` key into `data` but in this case accessing it will look like `route.snapshot.data.data` and it bothers me a little.

So what I propose is to allow `Route['resolve']` property to be a single value / observable / promise. In this case it's resolved value will be stored right into `route.snapshot.data`. If route contains a static `data` object it can be merged into the data from the resolver.

## Environment

<pre><code>
Angular: 6.1.1
... platform-browser, platform-browser-dynamic, router, upgrade
</code></pre>

martinsik:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #25430

The only way to use route resolvers is by passing an object that maps resolvers to `data` properties:

```
{
  path: 'simple',
  resolve: { two: 'resolveTwo' },
}
```

Resolved data is then available in `ActivatedRoute.data.two`.

## What is the new behavior?

This PR allows using route resolves without defining a map of resolvers. The sorter notation is like this:

```
{
  path: 'simple',
  resolve: 'resolveTwo',
}
```

Resolved values are then available straight in `ActivatedRoute.data`.

There's one special case when defining static route `data` together with this new simplified resolver notation. If the resolver returns an object it'll be merged with `data` using route's `paramsInheritanceStrategy` (it'll work exactly the same way as using an object with resolvers like before this PR). If the resolver returns anything other than an object then `data` is replaced with the resolver value.  
I'm not sure if I'm not making it too complicated.

So there are two use-cases went using with static router `data`:

1. Resolver returns an object:

   ```
   {
     path: 'simple',
     data: { foo: 1, bar: 2 },
     resolve: 'resolveObject', // resolves with `{ foo: 4 }`
   }
   ```

   `ActivatedRoute.data` will contain the following object `{ foo: 4, bar: 2 }`.

2. Resolver returns non object:

   ```
   {
     path: 'simple',
     data: { foo: 1, bar: 2 },
     resolve: 'resolveTwo', // resolves with `2`
   } 
   ```

   `ActivatedRoute.data` will contain the following object `2`.

- [ ] Yes
- [x] No

I had to change public API for `ResolveData` so it's now:

```
export declare type ResolveData = {
   [name: string]: any;
} | any;
```

instead of:

```
export declare type ResolveData = {
   [name: string]: any;
};
```

I think resolvers can be only service names right now so shouldn't there be `string`s instead of `any`s.

## Other information

I didn't want to update documentation until I have approval from somebody that these changes make sense :). 

Closes #25430

### DIFF
--- a/goldens/public-api/router/router.d.ts
+++ b/goldens/public-api/router/router.d.ts
@@ -266,7 +266,7 @@ export declare interface Resolve<T> {
 
 export declare type ResolveData = {
     [name: string]: any;
-};
+} | any;
 
 export declare class ResolveEnd extends RouterEvent {
     state: RouterStateSnapshot;
--- a/packages/router/src/config.ts
+++ b/packages/router/src/config.ts
@@ -89,7 +89,7 @@ export type Data = {
  */
 export type ResolveData = {
   [name: string]: any
-};
+}|any;
 
 /**
  *
--- a/packages/router/src/operators/resolve_data.ts
+++ b/packages/router/src/operators/resolve_data.ts
@@ -44,10 +44,14 @@ function runResolve(
   return resolveNode(resolve, futureARS, futureRSS, moduleInjector)
       .pipe(map((resolvedData: any) => {
         futureARS._resolvedData = resolvedData;
-        futureARS.data = {
-          ...futureARS.data,
-          ...inheritedParamsDataResolve(futureARS, paramsInheritanceStrategy).resolve
-        };
+        if (typeof resolvedData === 'object') {
+          futureARS.data = {
+            ...futureARS.data,
+            ...inheritedParamsDataResolve(futureARS, paramsInheritanceStrategy).resolve
+          };
+        } else {
+          futureARS.data = resolvedData;
+        }
         return null;
       }));
 }
@@ -55,6 +59,10 @@ function runResolve(
 function resolveNode(
     resolve: ResolveData, futureARS: ActivatedRouteSnapshot, futureRSS: RouterStateSnapshot,
     moduleInjector: Injector): Observable<any> {
+  if (isSingleValueResolver(resolve)) {
+    return getResolver(resolve, futureARS, futureRSS, moduleInjector);
+  }
+
   const keys = Object.keys(resolve);
   if (keys.length === 0) {
     return of({});
@@ -83,4 +91,8 @@ function getResolver(
   const resolver = getToken(injectionToken, futureARS, moduleInjector);
   return resolver.resolve ? wrapIntoObservable(resolver.resolve(futureARS, futureRSS)) :
                             wrapIntoObservable(resolver(futureARS, futureRSS));
+}
+
+function isSingleValueResolver(resolve: any) {
+  return typeof resolve === 'string' || typeof resolve === 'function';
 }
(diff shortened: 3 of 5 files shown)


martinsik:
There're unfortunately many more changes that I did myself because `yarn gulp format` did some reformatting.

atscott:
Hi @martinsik - Thanks for taking a stab at this one. I don't think there's a good enough signal that this is something we should implement at this time (I've gone ahead and removed the `community-help` label from the issue as well). I think we should only extend the router API when it's clear that the missing feature is a real painpoint for developers and there's no easy workaround. 

I don't think requiring that the resolvers be an object is terribly inconvenient and given the questions around what this means for merging with the `data`, I don't think we should move forward with it at this time. 

We may re-open this in the future if it becomes clear that the feature provides justifiable value.

PR #36569, opened 2020-04-10
fix(ngcc): do not spawn unlocker processes on cluster workers

gkalpak:
The current ngcc lock-file strategy spawns a new process in order to capture a potential `SIGINT` and remove the lock-file. For more information see #35861.

Previously, this unlocker process was spawned as soon as the `LockFile` was instantiated in order to have it available as soon as possible (given that spawning a process is an asynchronous operation). Since the `LockFile` was instantiated and passed to the `Executor`, this meant that an unlocker process was spawned for each cluster worker, when running ngcc in parallel mode. These processes were not needed, since the `LockFile` was not used in cluster workers, but we still had to pay the overhead of each process' own memory and V8 instance.
(NOTE: This overhead was small compared to the memory consumed by ngcc's normal operations, but still unnecessary.)

This commit avoids the extra processes by only spawning an unlocker process when first creating a lock-file (which only happens on the cluster master process). Since instantiating the `LockFile` and attempting to lock happen on the same tick, this change is unlikely to cause any difference in actual behavior.

### DIFF
--- a/packages/compiler-cli/ngcc/src/execution/cluster/lock_file_with_child_process.ts
+++ b/packages/compiler-cli/ngcc/src/execution/cluster/lock_file_with_child_process.ts
@@ -1,0 +1,42 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+/// <reference types="node" />
+
+import {ChildProcess} from 'child_process';
+import * as cluster from 'cluster';
+
+import {AbsoluteFsPath} from '../../../../src/ngtsc/file_system';
+import {LockFileWithChildProcess} from '../../locking/lock_file_with_child_process';
+
+
+/**
+ * A `LockFileWithChildProcess` that is `cluster`-aware and does not spawn unlocker processes from
+ * worker processes (only from the master process, which does the locking).
+ */
+export class ClusterLockFileWithChildProcess extends LockFileWithChildProcess {
+  write(): void {
+    if (!cluster.isMaster) {
+      // This is a worker process:
+      // This method should only be on the master process.
+      throw new Error('Tried to create a lock-file from a worker process.');
+    }
+
+    return super.write();
+  }
+
+  protected createUnlocker(path: AbsoluteFsPath): ChildProcess|null {
+    if (cluster.isMaster) {
+      // This is the master process:
+      // Create the unlocker.
+      return super.createUnlocker(path);
+    }
+
+    return null;
+  }
+}
--- a/packages/compiler-cli/ngcc/src/locking/lock_file_with_child_process/index.ts
+++ b/packages/compiler-cli/ngcc/src/locking/lock_file_with_child_process/index.ts
@@ -5,7 +5,8 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
-import {ChildProcess, fork} from 'child_process';
+import {ChildProcess, ChildProcessByStdio, fork} from 'child_process';
+import {Readable, Writable} from 'stream';
 
 import {AbsoluteFsPath, CachedFileSystem, FileSystem} from '../../../../src/ngtsc/file_system';
 import {Logger, LogLevel} from '../../logging/logger';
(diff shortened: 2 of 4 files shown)


petebacondarwin (commented):
The later the unlocker child process is created the more chance there is of it not being in-place in case the main process dies (or is killed). When I was testing different positions for the creation of the child process, it was quite possible to Ctrl-C before the unlocker had been created. Can you test this? Especially for the sync version of ngcc.

petebacondarwin (changes requested):
Instead of changing when the unlocker process is created, could we not just guard it with a check for `cluster.isMaster`?

gkalpak:
The code between instantiating the `LockFile` (which is where the unlocker process was spawn before this PR) and calling `LockFile#write()` (which where the unlocker process will be spawned after this PR) is all synchronous. So, I assumed there would be a tiny delay between the two (~10ms).

I tested it (which always beats "assuming" 😛) and I am seeing a delay between 40ms and 90ms (which is significantly higher than I thought. I'll reword the PR.

<sub>Thx for the feedback ❤️</sub>

##
> Instead of changing when the unlocker process is created, could we not just guard it with a check for `cluster.isMaster`?

We could, but I don't want to sprinkle the code with `cluster`-specific APIs as it will make it more difficult to migrate to a different parallelization mechanism. So, I try to limit the use of `cluster` APIs inside `ngcc/src/execution/cluster/`.

<sub>Sounds like we might need a `cluster`-specific `LockFile` :grin:</sub>

petebacondarwin:
> We could, but I don't want to sprinkle the code with cluster-specific APIs as it will make it more difficult to migrate to a different parallelization mechanism. So, I try to limit the use of cluster APIs inside ngcc/src/execution/cluster/.

This is specifically why I did not implement that previously. (I did know about the extra unlockers but I kept quiet about it 🙀). Are there any cases where `cluster.isMaster` would be false when we are not using clustering?

gkalpak:
I think `cluster.isMaster` will always be true when not running on a cluster worker afaict, but I would still prefer to refrain from using `cluster.XYZ` outside the `ngcc/src/execution/cluster/` directory. If we move to a different parallelization mechanism in the future, it will be easy to miss a `cluster`-specific API and continue to operate under the wrong assumption.

I changed the first commit to use a `ClusterLockFileWithChildProcess` \o/
PTAL

packages/compiler-cli/ngcc/src/execution/cluster/lock_file_with_child_process.ts
```diff
@@ -0,0 +1,42 @@
...
+
+import {AbsoluteFsPath} from '../../../../src/ngtsc/file_system';
+import {LockFileWithChildProcess} from '../../locking/lock_file_with_child_process';
+
+
+/**
+ * A `LockFileWithChildProcess` that is `cluster`-aware and does not spawn unlocker processes from
+ * worker processes (only from the master process, which does the locking).
+ */
+export class ClusterLockFileWithChildProcess extends LockFileWithChildProcess {
```
petebacondarwin, line 22: Rather than extending `LockFileWithChildProcess`, how about using composition instead...

```ts
export class ClusterLockFile implements LockFile {
  constructor(private delegate: LockFileWithChildProcess) {
    if (cluster.isMaster) {
      delegate.createUnlocker();
    }
  }
  ... etc
}
```

Obviously we would need to make `createUnlocker()` public, but this would make testing a lot cleaner, avoiding all those spies, since you could just pass a mock `delegate` in for testing.

petebacondarwin: It might need a bit of tweaking.... to work properly but you get the idea...

petebacondarwin (approved):
On further thought, over the weekend, I am happy to approve this.
My big gripe, outside of this PR, is that we are kicking off the worker processes with the same entry-point as the master process, which requires us to jump through these hoops.
I would like to propose that we land this but then refactor the whole clustering approach to avoid having to do this.

packages/compiler-cli/ngcc/src/execution/cluster/lock_file_with_child_process.ts
gkalpak: > Rather than extending `LockFileWithChildProcess`, how about using composition instead...

That would be my preference too, but the `delegate` would have spawned the process in its constructor. There are ways around that but they either make things more complicated than they already are or end up delaying the creation of the unlocker (which we are trying to avoid here).

> Or perhaps we could move this logic into the Locker classes themselves? Saving the middleman?

Not sure what you mean by that 😅

Issue #35488, opened 2020-02-17
Allow to enable native form validation globally


reda-alaoui:
# 🚀 feature request

### Relevant Package

@angular/forms

### Description

When using FormsModule, in order to enable native form validation,we are forced to add `ngNativeValidate` to each form.

### Describe the solution you'd like

Please provide a way to enable native form validation globally.

dylhunn:
While we appreciate your contribution, there doesn't seem to be much community demand for this feature. We're hesitant to expand this API like this without a stronger signal, given there is already a way for apps to do this.

PR #36591, opened 2020-04-12
feat(forms): allow to enable native form validation globally

reda-alaoui:
By default, native form validation is disabled on all forms by NgNoValidate
directive. Re-enabling native form validation requires to pass
'ngNativeValidate' to the form. This is not practical on projects having many
forms.

This change alleviates this difficulty by adding a FormsModule configuration
 option. When enabled, the native form validation will be the default
 validation of all forms.

## PR Checklist

## PR Type

- [X] Feature

Issue Number: #35488

## What is the new behavior?

This change adds a FormsModule configuration option. When enabled, the native form validation will be the default validation of all forms.

- [X] No

### DIFF
--- a/goldens/public-api/forms/forms.d.ts
+++ b/goldens/public-api/forms/forms.d.ts
@@ -326,7 +326,11 @@ export declare class FormGroupName extends AbstractFormGroupDirective implements
     constructor(parent: ControlContainer, validators: any[], asyncValidators: any[]);
 }
 
+export declare interface FormsConfigurationOptions extends SharedFormsConfigurationOptions {
+}
+
 export declare class FormsModule {
+    static withConfig(opts: FormsConfigurationOptions): ModuleWithProviders<FormsModule>;
 }
 
 export declare class MaxLengthValidator implements Validator, OnChanges {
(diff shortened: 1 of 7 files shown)


reda-alaoui:
Hello? :)

reda-alaoui:
Hello? :)

reda-alaoui:
v

reda-alaoui:
Hello?

jelbourn:
Hey @reda-alaoui, thanks for your patience on this PR. I think the delay comes from uncertainty about whether we'd want to move forward with an API like this. We try to be mindful that any new options like this increases the complexity of the system, so we try to focus on adding features based on feedback from a wide swath of community members. I think at this point we don't necessarily have a strong enough signal for this API.

reda-alaoui:
Hey @jelbourn ,

I am not asking for a sophisticated feature here. I just want to be able to use native browser form validation.
I want Angular to let me use the web standard and not stand in the way.

IMO, this is the kind of "feature" (add a thousand quotes if needed) that should no be implemented because of its popularity but because *we believe in web standards*.

dylhunn:
As per the linked issue:

> While we appreciate your contribution, there doesn't seem to be much community demand for this feature. We're hesitant to expand this API like this without a stronger signal, given there is already a way for apps to do this.

PR #36596, opened 2020-04-12
rollback clang update changes

josephperrott:
see individual commits


IgorMinar (approved):
LGTM

Context: https://docs.google.com/document/d/1W4FzJ2-OannSE78IrvefUiscQNfoQTevNKEnjiatK_w/edit#

IgorMinar:
hmm.. the github UI doesn't allow me to remove other reviewers maybe because the PR is too big?

This is a  global-approvers PR so the only other approver should be @kara. We might need to use the merge-assistance label unless we do manage to get all the other people with pending reviews off of this PR.

IgorMinar:
merge-assistance: due to what seems to be a github bug I'm not able to remove the existing reviewers now that this PR was approved with global approvers rights.

I also wanted @kara to review this, but she's OOO and can't review this. In the meantime the current reformatting setup is affecting our operations so the rollback is time sensitive. Let's proceed.

atscott:
Adding blocked label. If this PR is not merged along with a similar one that targets path-only, all PRs targetting both master and patch would have issues.

PR #36601, opened 2020-04-13
fix(docs-infra): add visual signal for external links

### LINKED ISSUE #17620: Provide a visual signal for external links
## Current behavior
Currently inspecting the target url is the only way for a reader of our docs to know if clicking on a link will take them to an external resource. This is time consuming and often leads to readers leaving angular.io unintentionally.

## Expected behavior
I'd like to see a visual cue near each external link that would tell me if clicking on the link will take me to an external resource or not.

Several pages have solved this via arrows pointing up and to the right. The following is from developers.google.com:

---

screen shot 2017-06-19 at 3 24 10 pm

---

In my opinion this particular design is too heavy weight and too distracting, something more lightweight, but still able to communicate the same information would be much more preferable.

ajitsinghkaler:
Clicking on a link may take us to an external source, which may lead to the user leaving angular.io unintentionally.

Added visual cues on external links so that the user knows which links are external and which are internal to angular.io.

Fixes #17620

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

No visual cue for external links

Issue Number: #17620

## What is the new behavior?
Visual cue added for external links

- [ ] Yes
- [x] No


ajitsinghkaler:
@IgorMinar said not to use this link but this is the link that users are most familiar with for external links

To automate this we can also use css selectors like a[href^="http://"],a[href^="https://"]

aio/src/app/layout/footer/footer.component.html
```diff
@@ -11,9 +11,9 @@ <h3>{{node.title}}</h3>
 
 <p>
   Super-powered by Google ©2010-2020.
-  Code licensed under an <a href="license" title="License text">MIT-style License</a>.
+  Code licensed under an <a title="License text">MIT-style License</a>.
```
gkalpak, line 14: This change looks invalid. Please revert.

aio/src/app/layout/footer/footer.component.html
```diff
@@ -11,9 +11,9 @@ <h3>{{node.title}}</h3>
...
   Documentation licensed under
-  <a href="http://creativecommons.org/licenses/by/4.0/">CC BY 4.0</a>.
+  <a class="external-link" href="license" href="http://creativecommons.org/licenses/by/4.0/">CC BY 4.0</a>.
```
gkalpak, line 16: We can actually match all external links with the following selector (which I would prefer, since we don't have to add a class to every current and future external link):

```css
a[href^="http:"]::after, a[href^="https:"]::after {
  ...
}
```

BTW, the `[<attr>^="<value>"]` CSS selector matches elements whose `<attr>` attribute value begins exactly with the string `<value>`.

aio/src/styles/0-base/_typography.scss
```diff
@@ -192,3 +192,13 @@ code {
...
 }
+
+.external-link::after {
+  font-family: "Material Icons";
+  content: "open_in_new";
+  display: inline-block;
+  height: 13px;
+  margin-left: 4px;
+  width: 15px;
+  position: relative;
```
gkalpak, line 203: 1. Does `display`, `position`, `height`, `width` make any difference? I couldn't notice any.
2. Please also add `vertical-align: bottom;`.

gkalpak (changes requested):
Nice! Thx, @ajitsinghkaler :+1:

BTW, I believe this is more of a `fix` that a `feat`.

aio/src/app/layout/footer/footer.component.html
IgorMinar: Yup. It would be better for us not to explicitly add all the links because we'll miss some or people will accidentally use the class even on some internal links.

aio/src/styles/0-base/_typography.scss
IgorMinar: Yeah. After seeing this on the preview site, I was going to say that it needs to be bottom aligned. 🙃

IgorMinar (commented):
Thanks for the change! This will be great once all of the minor issues George brought up are fixed.

// Fyi: @aikidave

aio/src/styles/0-base/_typography.scss
sjtrimble: Also probably add `@include line-height(24)` here .... not sure if height is also needed.
I think this makes a difference with the focus outline and how that is placed.
Screen Shot 2020-04-13 at 4 16 40 PM

IgorMinar: that looks great! thanks @sjtrimble!

ajitsinghkaler:
I tried the selectors but a few problems that I faced 
Icons on toolbar too icon on GitHub edit link in docs and contribute also failing

I think we need to go page by page what are your thoughts on this @IgorMinar @gkalpak

gkalpak:
It will be very few external links that we don't want to have the icon. We should either add a class to them to turn off the icon or add more styles to exclude those. (My preference is for the former.)

E.g.:

```css
// 1 - Add a `no-external-link-icon` class to each external link we don't want to have the icon.
//     Then, in Scss:
a:not(.no-external-link-icon) {
  &[href^="http:"],
  &[href^="https:"]: {
    ...
  }
}

// 2 - Add rules to exclude certain elements:
a[href^="http:"]::after,
a[href^="https:"]::after: {
  ...
}

.github-links, 
.toolbar-external-icons-container {
  & a[href^="http:"]::after,
  & a[href^="https:"]::after: {
    display: none;  /* ...or some other way of deisabling the icon. */
  }
}
```

IgorMinar:
@ajitsinghkaler I'm sorry, but I don't understand your comment. Can you please explain? Maybe a screenshot would help? Also I don't see the changes we suggested implemented in this PR so I'm not sure if you tried it locally and it didn't work or we didn't explain correctly what we asked you to change. Can you please clarify?

ajitsinghkaler:
> @ajitsinghkaler I'm sorry, but I don't understand your comment. Can you please explain? Maybe a screenshot would help? Also I don't see the changes we suggested implemented in this PR so I'm not sure if you tried it locally and it didn't work or we didn't explain correctly what we asked you to change. Can you please clarify?

I've pushed the changes I did on local you can have a look

ajitsinghkaler:
> It will be very few external links that we don't want to have the icon. We should either add a class to them to turn off the icon or add more styles to exclude those. (My preference is for the former.)
> 
> E.g.:
> 
> ```css
> // 1 - Add a `no-external-link-icon` class to each external link we don't want to have the icon.
> //     Then, in Scss:
> a:not(.no-external-link-icon) {
>   &[href^="http:"],
>   &[href^="https:"]: {
>     ...
>   }
> }
> 
> // 2 - Add rules to exclude certain elements:
> a[href^="http:"]::after,
> a[href^="https:"]::after: {
>   ...
> }
> 
> .github-links, 
> .toolbar-external-icons-container {
>   & a[href^="http:"]::after,
>   & a[href^="https:"]::after: {
>     display: none;  /* ...or some other way of deisabling the icon. */
>   }
> }
> ```

Yes I also will prefer class methods I tried the second way today always leads to unintended consequences in the code

@gkalpak will ping you once done

IgorMinar:
Aha! I think I see the problem you mentioned. And @gkalpak already posted a suggestion on how to resolve it via an explicit opt-out using classes for links that are external but we don't want them to be marked as such via the icon because it is distracting/noise - e.g. in the app footer.

@gkalpak by the other method using selectors - did you mean scoping the external link icon only to external links found in the doc-viewer content? If so, why not use that approach instead of the explicit opt-out you are recommending? Unless I'm missing something using scoped selector would resolve the issue with less css class noise in the app shell part of the app. Can you please clarify?

gkalpak:
> did you mean scoping the external link icon only to external links found in the doc-viewer content?

No, we have to exclude some links that are _inside_ the doc-viewer, such as the "view/edit source" icon buttons shown for each guide/API doc/method/etc:

icon-buttons

So, we can't just scope the icons to `doc-viewer`.

##
> we don't want them to be marked as such via the icon because it is distracting/noise - e.g. in the app footer.

Just a note: The footer contains both internal and external links.

IgorMinar:
@gkalpak am I missing something when I look at the dom structure and think to see an opportunity for us to use `aio-doc-viewer > .content` to scope the external link icons?

ajitsinghkaler:
> @gkalpak am I missing something when I look at the dom structure and think to see an opportunity for us to use `aio-doc-viewer > .content` to scope the external link icons?

@IgorMinar it will still apply in about page, contributors page, contribute page etc

So this will also not suffice

IgorMinar:
@ajitsinghkaler aha! Thanks for pointing that out. Hmm... Tricky situation...

ajitsinghkaler:
@IgorMinar @gkalpak  tried a hack, do you think this will suffice

aio/src/styles/0-base/_typography.scss
```diff
@@ -192,3 +192,21 @@ code {
...
+  a[href^="https:"]::after {
+    font-family: "Material Icons";
+    content: "open_in_new";
+    position: relative;
+    @include line-height(24);
+    vertical-align: bottom; 
+  }
+}
+
+aio-doc-viewer .content *[ng-version] {
```
IgorMinar, line 207: Is this because the special pages are Angular Elements which one bootstrapped leave ng-version tag on the host element of the Angular Element?

This seems pretty hacky but maybe we could tighten it a bit more by using `>` to make the selector more specific + document this hack and leave it at that.

@gkalpak thoughts?

gkalpak: Sorry, I don't think this is a good hack :grin:

Actually, I think we need to take a step back and determine where we want to show the icon and where not.

For example, the current setup does not show links in API docs and CLI docs (because their containers don't have the `.content` class for some reason). This doesn't look right to me. Furthermore, this could easily change in the future (i.e. someone decides to add the `.content` class for consistency), without us realizing that that would start showing icons for github links in API docs for example:
 
<p align="center">


---
I think we can consider showing icons as the default behavior (applied to all links) and start listing exceptions (i.e. cases where we don't want an icon).

Exceptions I can think of:

- [ ] `.github-links *` (on docs pages (API docs, guides, tutorials, etc.).
  <p align="center">
  

##
- [ ] `.toolbar-external-icons-container *` (on all pages).
  <p align="center">
  <br />
  <sub>(See Twitter and GitHub icon-buttons.)</sub>

##
- [ ] `.contributor-info *` (on [/about](https://pr36601-f573f6b.ngbuilds.io/about)).
  <p align="center">
  

##
- [ ] Maybe `aio-top-menu *` (on all pages)?
  <p align="center">
  <br />
  <sub>(See "Blog" menu item.)</sub>

##
- [ ] Maybe `.contribute-container .button` (on [/contribute](https://pr36601-f573f6b.ngbuilds.io/contribute))?
  <p align="center">
  

##
- [ ] Maybe `.resources-container .resource-row-link` (on [/resources](https://pr36601-f573f6b.ngbuilds.io/resources))?
  <p align="center">
  

---
Or maybe we can decide that we only want to show icons for links on docs (non-marketing) pages (with exceptions), in which case we can use scope the styles using the `.folder-{api,cli,guide,start,tutorial}` class added to the `<aio-shell>`.

gkalpak (commented):
BTW, it seems that is some places the `Material Icons` font is not applied correctly:

- contribute
-  topbar

ajitsinghkaler:
@gkalpak did you see the this link https://pr36601-f573f6b.ngbuilds.io/. it only show links on contribute page

ajitsinghkaler:
I think you saw the wrong preview

gkalpak:
> @gkalpak did you see the this link https://pr36601-f573f6b.ngbuilds.io/. it only show links on contribute page

I didn't say there are links in the places I showed in the images. I am just starting a discussion on where we do and don't want links, so that we can then decide what is the best (and more safe/future-proof) way to show them :smiley:

ajitsinghkaler:
Ohhh! okay my mistake

ajitsinghkaler:
> Or maybe we can decide that we only want to show icons for links on docs (non-marketing) pages (with exceptions), in which case we can use scope the styles using the .folder-{api,cli,guide,start,tutorial} class added to the <aio-shell>.

I like this option better because Someone can include new links with other classes later on so including the current classes in a :not is not right I think we go page by page and add icon to the footer also.

Because the main reason we are adding this is users can see he difference between external and internal links footer also has a mix of external and internal lin ks so I think we should put it in footer too. Rest of the marketing  pages all have external or internal links not both. Only docs pages have mix of these links. So the two places that I would like to add these is in docs and footer.

@gkalpak @IgorMinar thoughts?

IgorMinar:
Oh my! I totally missed that "folder" comment from @gkalpak.

I was actually trying to suggest that it is totally sufficient for us to distinguish only the external links in API, guide, and tutorial content within the doc viewer.

The thinking here is that those are the only links that a reader would click on while "in the flow" which would take them out of the flow.

It's unlikely that a reader would click on the footer or header link while "in the flow" [1]

So I'm totally fine with using the approach @gkalpak suggested and relying on the folder css classes on the doc-viewer which I didn't even know that existed.

[1] https://en.m.wikipedia.org/wiki/Flow_(psychology)

ajitsinghkaler:
@gkalpak @IgorMinar made the changes please have a look

ajitsinghkaler:
@gkalpak @IgorMinar made the changes and everything looks in good shape now. You can have a look.

gkalpak (approved):
One minor comment. Otherwise lgtm :tada:
Nice work, @ajitsinghkaler :+1:

aio/src/styles/0-base/_typography.scss
```diff
@@ -192,3 +192,32 @@ code {
...
+  aio-doc-viewer{
+
+    a {
+      &[href^="http:"]::after,
+      &[href^="https:"]::after {
+        font-family: "Material Icons";
+        content: "open_in_new";
+        position: relative;
+        @include line-height(24);
+        vertical-align: bottom;
```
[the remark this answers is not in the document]
ajitsinghkaler: And I don't know why these tests are failing

ajitsinghkaler:
@IgorMinar can you please have a look

ajitsinghkaler:
@IgorMinar please have a look

ajitsinghkaler:
@gkalpak can you please check why tests are failing I checked but did not find any answer

gkalpak:
@ajitsinghkaler, rebasing on latest master should fix the issue.

gkalpak:
It might be due to an issue that was recently fixed in #36825.
Feel free to ping me again if it fails after the rebase.

ajitsinghkaler:
@gkalpak can you please remove the review label. I think this pr is ready

ajitsinghkaler:
@IgorMinar can you please have a look

ajitsinghkaler:
@IgorMinar can you please have a look at this too it is also ready.

aio/src/styles/0-base/_typography.scss
```diff
@@ -192,3 +192,32 @@ code {
     color: $mediumgray;
   }
 }
+
```
IgorMinar, line 195: This looks great! Do you think that you could add a comment explaining how this works as a follow up PR? That way when we change stuff in the future we don't need to reverse engineer the stylesheet.

IgorMinar (approved):
This is awesome. I'm so excited about this change finally landing!

Great job @ajitsinghkaler thanks for patiently iterating on this change and getting us to this simple but effective solution. 🙇🏼‍♂️

aio/src/styles/0-base/_typography.scss
ajitsinghkaler: Yes, I'll add the comments in a follow-up pr

IgorMinar:
Hmm. Actually the CLA not did not trigger on your last push. I think that this is because of an outage the CLA not had a few days ago. Can you please add the docs to this PR and rebase the PR to master. That should retrigger the bit and the PR should go green.

I'm sorry about this hassle..

IgorMinar:
Please comment on this pr when you push the update and I'll monitor the CLA bot.

IgorMinar:
CLA bot looks happy now! Let's hope the tests will not flake.

ajitsinghkaler:
Yeah, I just was going to ping you. Just wanted to see if any other errors may have occur because of rebasing

Thank you for all your help @IgorMinar

PR #36608, opened 2020-04-13
docs(docs-infra): fix handling of client-side errors in networking

wojteko22:
Previously example for handling client-side errors in networking didn't work when there was no Internet connection.
Caught error is ProgressEvent in such case, not ErrorEvent and error.error.message is undefined.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/aio/content/guide/http.md
+++ b/aio/content/guide/http.md
@@ -327,7 +327,7 @@ Two types of errors can occur.
 
 * The server backend might reject the request, returning an HTTP response with a status code such as 404 or 500. These are error _responses_.
 
-* Something could go wrong on the client-side such as a network error that prevents the request from completing successfully or an exception thrown in an RxJS operator. These errors produce JavaScript `ErrorEvent` objects.
+* Something could go wrong on the client-side such as a network error that prevents the request from completing successfully or an exception thrown in an RxJS operator. These errors have `status` set to `0` and the `error` property contains a `ProgressEvent` object, whose `type` might provide further information.
 
 `HttpClient` captures both kinds of errors in its `HttpErrorResponse`. You can inspect that response to identify the error's cause.
 
--- a/aio/content/examples/http/src/app/config/config.service.ts
+++ b/aio/content/examples/http/src/app/config/config.service.ts
@@ -72,9 +72,9 @@ export class ConfigService {
 
   // #docregion handleError
   private handleError(error: HttpErrorResponse) {
-    if (error.error instanceof ErrorEvent) {
+    if (error.status === 0) {
       // A client-side or network error occurred. Handle it accordingly.
-      console.error('An error occurred:', error.error.message);
+      console.error('An error occurred:', error.error);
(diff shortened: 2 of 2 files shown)


petebacondarwin:
So I assume that this `ProgressError` is the one passed by this bit of code:

https://github.com/angular/angular/blob/de4fafb818b1648813aef227df883c7502b825fb/packages/common/http/src/xhr.ts#L209-L221

But I note in that code, that the status is grabbed from the `xhr` object. So it is possible that the status is not 0, right?

@wojteko22 - are you able to demonstrate that this would always be 0 in this scenario?

wojteko22:
@petebacondarwin I don't know which statuses are possible in case of no Internet connection but it would be strange to have sth different than zero. But the thing here is how to distinguish these situations:
1. "client-side or network error" 
2. "The backend returned an unsuccessful response code"

Or, in other words, how to write condition to ensure that it is "client-side or network error". For sure `error.error instanceof ErrorEvent` isn't good fit, because ErrorEvent isn't thrown when there is no Internet connection. And the previous version of example logs "The backend returned an unsuccessful response code" in such case,  which is not true. 

As I already said, I don't know if `error.status === 0` is OK.

So maybe we can use `error.error instanceof ErrorEvent || error.status === 0` or `error.error instanceof ErrorEvent || error.error instanceof ProgressEvent`? Or maybe it is currently not possible at all to ensure that it is a client-side or network error and the example should be simplified to handle all exceptions in the similar way?

aio/content/guide/http.md
```diff
@@ -327,7 +327,7 @@ Two types of errors can occur.
 
 * The server backend might reject the request, returning an HTTP response with a status code such as 404 or 500. These are error _responses_.
 
-* Something could go wrong on the client-side such as a network error that prevents the request from completing successfully or an exception thrown in an RxJS operator. These errors produce JavaScript `ErrorEvent` objects.
+* Something could go wrong on the client-side such as a network error that prevents the request from completing successfully or an exception thrown in an RxJS operator.
```
petebacondarwin, line 330:
```suggestion
* Something could go wrong on the client-side such as a network error that prevents the request from completing successfully or an exception thrown in an RxJS operator.  These errors have `status` set to `0` and the `error` property contains a `ProgressEvent` object, whose `type` might provide further information.
```

petebacondarwin (approved):
OK, so I played around with this a bit and I think you are right.
The `error.error` is definitely a `ProgressEvent` and `status` always appear to be `0`.

So let's go with you changes.
I have suggested that you just update the text in the guide rather than removing it.

petebacondarwin (approved):

AndrewKushnir:
FYI, I've rebased this PR to restart the CI (the `ci/circleci: setup` was failing).

PR #26521, opened 2018-10-17
Redirect if a guard returns a UrlTree

jasonaden:
Fixes #24618

PR #36610, opened 2020-04-13
feat(router): allow CanLoad guard to return UrlTree

### LINKED ISSUE #28306: Router: CanLoad Guard can't return UrlTree
### Affected Package
angular/router

### Description

The `CanActivate `interface allows returning a `UrlTree`, but the `CanLoad `interface only allows `boolean`, `Observable<boolean>` or `Promise<boolean>`

## 🌍  Your Environment

**Angular Version:**
<pre><code>
Angular CLI: 7.2.1
Node: 8.10.0
OS: win32 x64
Angular: 7.2.0

@angular-devkit/architect         0.12.1
@angular-devkit/build-angular     0.12.1
@angular-devkit/build-optimizer   0.12.1
@angular-devkit/build-webpack     0.12.1
@angular-devkit/core              7.2.1
@angular-devkit/schematics        7.2.1
@angular/cdk                      7.2.1
@angular/cli                      7.2.1
@angular/flex-layout              7.0.0-beta.23
@angular/material                 7.2.1
@ngtools/webpack                  7.2.1
@schematics/angular               7.2.1
@schematics/update                0.12.1
rxjs                              6.3.3
typescript                        3.2.2
webpack                           4.23.1

</code></pre>

atscott:
A `CanLoad` guard returning `UrlTree` cancels current navigation and redirects.
This matchers the behavior available to `CanActivate` guards added in https://github.com/angular/angular/pull/26521

fixes #28306

### DIFF
--- a/packages/router/src/apply_redirects.ts
+++ b/packages/router/src/apply_redirects.ts
@@ -8,7 +8,7 @@
 
 import {Injector, NgModuleRef} from '@angular/core';
 import {EmptyError, from, Observable, Observer, of} from 'rxjs';
-import {catchError, concatAll, every, first, map, mergeMap} from 'rxjs/operators';
+import {catchError, concatAll, every, first, map, mergeMap, tap} from 'rxjs/operators';
 
 import {LoadedRouterConfig, Route, Routes} from './config';
 import {CanLoadFn} from './interfaces';
@@ -16,7 +16,7 @@ import {RouterConfigLoader} from './router_config_loader';
 import {defaultUrlMatcher, navigationCancelingError, Params, PRIMARY_OUTLET} from './shared';
 import {UrlSegment, UrlSegmentGroup, UrlSerializer, UrlTree} from './url_tree';
 import {forEach, waitForMap, wrapIntoObservable} from './utils/collection';
-import {isCanLoad, isFunction} from './utils/type_guards';
+import {isCanLoad, isFunction, isUrlTree} from './utils/type_guards';
 
 class NoMatch {
   public segmentGroup: UrlSegmentGroup|null;
@@ -300,9 +300,9 @@ class ApplyRedirects {
         return of(route._loadedConfig);
       }
 
-      return runCanLoadGuard(ngModule.injector, route, segments)
-          .pipe(mergeMap((shouldLoad: boolean) => {
-            if (shouldLoad) {
+      return this.runCanLoadGuards(ngModule.injector, route, segments)
+          .pipe(mergeMap((shouldLoadResult: boolean) => {
+            if (shouldLoadResult) {
               return this.configLoader.load(ngModule.injector, route)
                   .pipe(map((cfg: LoadedRouterConfig) => {
                     route._loadedConfig = cfg;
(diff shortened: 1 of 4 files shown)


atscott:
[presubmit](https://test.corp.google.com/ui#id=OCL:306282210:BASE:306282241:1586805068712:10e055f4)

punpat2000 (approved):

petebacondarwin (approved):
This LGTM. I have two questions about the approach in general:

1) how does this impact preloading strategies? At the moment an eager preload will just run `canLoad` guards in the background (if I understand correctly) before an attempt to actually navigate to the lazy routes and these routes will not be loaded if not yet allowed. But with this new feature, if any of these guards return a `UrlTree` will it not cause a random navigation to the new `UrlTree` without the user actually requesting to navigate to the guarded URL?
2. If multiple `canLoad` guards are async observables that return `UrlTree` results It seems that which ever observable emits first will win. This does not appear to be deterministic as to which `UrlTree` will be navigated to. Is this a concern?

packages/router/src/apply_redirects.ts
```diff
@@ -316,6 +316,38 @@ class ApplyRedirects {
...
+        guardVal = guard(route, segments);
+      } else {
+        throw new Error('Invalid CanLoad guard');
+      }
+      return wrapIntoObservable(guardVal);
+    }));
+
+    return obs.pipe(
+        concatAll(),
+        tap((result: UrlTree|Boolean) => {
```
JoostK, line 339:
```suggestion
        tap((result: UrlTree|boolean) => {
```

JoostK (approved):
LGTM

atscott:
@petebacondarwin 
> 1. how does this impact preloading strategies? At the moment an eager preload will just run canLoad guards in the background (if I understand correctly) before an attempt to actually navigate to the lazy routes and these routes will not be loaded if not yet allowed.

Thanks for bringing this up! I knew I was missing something. I'll do some research and get back to you. My gut reaction is that it would be treated the same as a `false` but I actually might be missing a `catchError` somewhere for this case.

> 2. If multiple canLoad guards are async observables that return UrlTree results It seems that which ever observable emits first will win. This does not appear to be deterministic as to which UrlTree will be navigated to. Is this a concern?

This would be handled by the use of `concatAll` so guards will be processed in the order they appear in the config.

petebacondarwin:
> This would be handled by the use of concatAll so guards will be processed in the order they appear in the config.

Ahah! Thanks. I forgot how `concatAll` works. This is perfect.

atscott:
@petebacondarwin - I double checked the behavior with preloading and it turns out `canLoad` is documented as incompatible with preloading regardless of what it returns: https://angular.io/guide/router#canload-blocks-preload. I added a test to verify that the `canLoad` guard is not executed as part of preloading.

petebacondarwin:
👍 I did read that doc but I misunderstood what it was saying. I thought it meant that `canLoad` returning false blocked the preload. But I see now that it actually means that even having a `canLoad` guard prevents it from being preloaded.

atscott:
[presubmit](https://test.corp.google.com/ui#id=OCL:306282210:BASE:307680406:1587504354817:7877c677)

IgorMinar (approved):
LGTM. Thanks!

Reviewed-for: code-ownership
LGTM! Thanks!

Reviewed-for: public-api

IgorMinar (approved):
+global approval, since we don't have a fw review nut @atscott is the code owner

rakishii13:
Will this be released in a 9.x version or in V10?

atscott:
Hi @rakishii13, this will be available in the v10 release.

PR #36611, opened 2020-04-13
fix(common): format day-periods that cross midnight

### LINKED ISSUE #36566: formatDate returns AM/PM for "B" pattern in locale "en"
# 🐞 bug report

### Affected Package
The issue is caused by package **@angular/common** (locales/extra).

### Is this a regression?

This does not appear to be a regression.

### Description

If you try to use `formatDate` with the "B" placeholder ("at night", "in the morning", etc.), this does not work for locale "en" (and sublocales) between 21:00 and 06:00. Instead of getting "at night", you'll get "AM/PM".

This is caused by a misinterpretation of the time rule `['23:00', '06:00']` in `@angular/common/locales/extra/en*.ts`, as it will never match: https://github.com/angular/angular/blob/fee316161d5de6d5b8f77dc13278db779ea5b4c5/packages/common/src/i18n/format_date.ts#L286-L290

This will happen for any rule that straddles midnight (although a quick glance only shows English to be affected).

## 🔬 Minimal Reproduction
```
formatDate('2018-06-27T04:23:00', 'h B', 'en') + ', ' + formatDate('2018-06-27T06:23:00', 'h B', 'en')
```
Expected: `4 at night, 6 in the morning`, actual: `4 AM, 6 in the morning`.

Compared to Dutch: `4 's nachts, 6 's ochtends` (not `4 a.m., 6 's ochtends`).

## 🌍  Your Environment

**Angular Version:**
Angular 6–9

petebacondarwin:
When formatting a time with the `b` or `B` format codes,
the resulting string should represent a period of the day.

The logic for computing this period is based on start and end
times, but the code was not able to cope with a period that
crossed midnight.

Fixes #36566

### DIFF
--- a/packages/common/src/i18n/format_date.ts
+++ b/packages/common/src/i18n/format_date.ts
@@ -277,26 +277,40 @@ function getDateTranslation(
       if (extended) {
         const rules = getLocaleExtraDayPeriodRules(locale);
         const dayPeriods = getLocaleExtraDayPeriods(locale, form, width);
-        let result;
-        rules.forEach((rule: Time|[Time, Time], index: number) => {
+        const index = rules.findIndex(rule => {
           if (Array.isArray(rule)) {
             // morning, afternoon, evening, night
-            const {hours: hoursFrom, minutes: minutesFrom} = rule[0];
-            const {hours: hoursTo, minutes: minutesTo} = rule[1];
-            if (currentHours >= hoursFrom && currentMinutes >= minutesFrom &&
-                (currentHours < hoursTo ||
-                 (currentHours === hoursTo && currentMinutes < minutesTo))) {
-              result = dayPeriods[index];
+            const [from, to] = rule;
+            const afterFrom = currentHours >= from.hours && currentMinutes >= from.minutes;
+            const beforeTo =
+                (currentHours < to.hours ||
+                 (currentHours === to.hours && currentMinutes < to.minutes));
+            // We must account for normal rules that span a period during the day (e.g. 6am-9am)
+            // where `from` is less (earlier) than `to`. But also rules that span midnight (e.g.
+            // 10pm - 5am) where `from` is greater (later!) than `to`.
+            //
+            // In the first case the current time must be BOTH after `from` AND before `to`
+            // (e.g. 8am is after 6am AND before 10am).
+            //
+            // In the second case the current time must be EITHER after `from` OR before `to`
+            // (e.g. 4am is before 5am but not after 10pm; and 11pm is not before 5am but it is
+            // after 10pm).
+            if (from.hours < to.hours) {
+              if (afterFrom && beforeTo) {
+                return true;
+              }
+            } else if (afterFrom || beforeTo) {
+              return true;
             }
           } else {  // noon or midnight
-            const {hours, minutes} = rule;
-            if (hours === currentHours && minutes === currentMinutes) {
-              result = dayPeriods[index];
+            if (rule.hours === currentHours && rule.minutes === currentMinutes) {
+              return true;
             }
           }
+          return false;
         });
-        if (result) {
-          return result;
+        if (index !== -1) {
+          return dayPeriods[index];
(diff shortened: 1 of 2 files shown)


packages/common/src/i18n/format_date.ts
```diff
@@ -277,26 +277,30 @@ function getDateTranslation(
...
-            if (currentHours >= hoursFrom && currentMinutes >= minutesFrom &&
-                (currentHours < hoursTo ||
-                 (currentHours === hoursTo && currentMinutes < minutesTo))) {
-              result = dayPeriods[index];
+            const [from, to] = rule;
+            const afterFrom = currentHours >= from.hours && currentMinutes >= from.minutes;
+            const beforeTo =
+                (currentHours < to.hours ||
+                 (currentHours === to.hours && currentMinutes < to.minutes));
+            if (from.hours < to.hours) {
```
AndrewKushnir, line 288: I'd propose to add some comments to this code block to describe the `if` and `else if` logic.

packages/common/test/i18n/format_date_spec.ts
```diff
@@ -185,13 +202,31 @@ describe('Format date', () => {
...
       Object.keys(dateFixtures).forEach((pattern: string) => {
         expectDateFormatAs(date, pattern, dateFixtures[pattern]);
       });
 
       Object.keys(isoStringWithoutTimeFixtures).forEach((pattern: string) => {
         expectDateFormatAs(isoStringWithoutTime, pattern, isoStringWithoutTimeFixtures[pattern]);
       });
+
+      const nightTime = new Date(2015, 5, 15, 2, 3, 1, 550);
+      Object.keys(midnightCrossingPeriods).forEach(pattern => {
```
AndrewKushnir, line 227: nit: maybe use `Object.entries` here?

AndrewKushnir (approved):
Thanks for the fix @petebacondarwin! 👍 

I just left a couple minor comments, please have a look when you get a chance.

Quick question: should we consider this fix as a (minor) breaking change? Even though it's a bug fix, the behavior is slightly changed, so code that relied on the old behavior might need to be updated. What do you think @petebacondarwin?

AndrewKushnir:
I've started [Blueprint Presubmit](http://test/OCL:306333739:BASE:306333761:1586820870548:cfca9edc) and will also start a Global one tonight (and will update this comment with the link). Thank you.

packages/common/src/i18n/format_date.ts
petebacondarwin: Yes, I'll add a comment.

If you have any suggestions on how to make this logic simpler/smaller/faster I would be interested to hear. I spent some time trying to find an optimal approach and this was the best I came up with.

packages/common/test/i18n/format_date_spec.ts
petebacondarwin: I also considered this but I decided to keep the approach aligned to the rest of the code in this test, for consistency.

petebacondarwin:
Personally, I don't see this as a breaking change. Sure the behaviour changed - but that is because it was broken before, no? I will change the commit message to highlight this change to what is output.

petebacondarwin:
I added a comment and updated the commit message - no code changes, so the presubmits should stand.

packages/common/src/i18n/format_date.ts
AndrewKushnir: Thanks for adding a comment @petebacondarwin 👍 It looks great!

AndrewKushnir:
@petebacondarwin FYI, Blueprint and Global presubmits are successful 👍

petebacondarwin:
@AndrewKushnir do we need to run presubmit again, or can we add merge-assistance to push this through?

AndrewKushnir:
@petebacondarwin, I think it's safe to reuse existing g3 global presubmit, so I forced "green" g3 status for this PR. Thank you.

PR #35125, opened 2020-02-03
fix(ivy): narrow `NgIf` context variables in template type checker

JoostK:
fix(ivy): narrow `NgIf` context variables in template type checker

When the `NgIf` directive is used in a template, its context variables
can be used to capture the bound value. This is typically used together
with a pipe or function call, where the resulting value is captured in a
context variable. There's two syntax forms available:

1. Binding to `NgIfContext.ngIf` using the `as` syntax:
```html
<span *ngIf="(user$ | async) as user">{{user.name}}</span>
```

2. Binding to `NgIfContext.$implicit` using the `let` syntax:
```html
<span *ngIf="user$ | async; let user">{{user.name}}</span>
```

Because of the semantics of `ngIf`, it is known that the captured
context variable is non-nullable, however the template type checker
would not consider them as such and still report errors when
`strictNullTypes` is enabled.

This commit updates `NgIf`'s context guard to make the types of the
context variables non-nullable, avoiding the issue.

Fixes #34572

PR #36627, opened 2020-04-14
fix(common): narrow `NgIf` context variables in template type checker

ranma42:
When the `NgIf` directive is used in a template, its context variables
can be used to capture the bound value. This is sometimes used in
complex expressions, where the resulting value is captured in a
context variable. There's two syntax forms available:

1. Binding to `NgIfContext.ngIf` using the `as` syntax:
```html
<span *ngIf="enabled && user as u">{{u.name}}</span>
```

2. Binding to `NgIfContext.$implicit` using the `let` syntax:
```html
<span *ngIf="enabled && user; let u">{{u.name}}</span>
```

Because of the semantics of `ngIf`, it is known that the captured
context variable is truthy, however the template type checker
would not consider them as such and still report errors when
`strict` is enabled.

This commit updates `NgIf`'s context guard to make the types of the
context variables truthy, avoiding the issue.

Based on https://github.com/angular/angular/pull/35125

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

When compiling in `strict` mode, the compiler rejects

```html
<span *ngIf="enabled && user as u">{{u.name}}</span>
```

## What is the new behavior?

The typing of `ngIf` is more accurate, excluding most falsy values (the main exception should be NaNs) from the type passed to the template body.

- [ ] Yes
- [x] No

### DIFF
--- a/goldens/public-api/common/common.d.ts
+++ b/goldens/public-api/common/common.d.ts
@@ -243,7 +243,7 @@ export declare class NgIf<T = unknown> {
     set ngIfThen(templateRef: TemplateRef<NgIfContext<T>> | null);
     constructor(_viewContainer: ViewContainerRef, templateRef: TemplateRef<NgIfContext<T>>);
     static ngTemplateGuard_ngIf: 'binding';
-    static ngTemplateContextGuard<T>(dir: NgIf<T>, ctx: any): ctx is NgIfContext<NonNullable<T>>;
+    static ngTemplateContextGuard<T>(dir: NgIf<T>, ctx: any): ctx is NgIfContext<Exclude<T, false | 0 | '' | null | undefined>>;
 }
 
 export declare class NgIfContext<T = unknown> {
--- a/packages/common/src/directives/ng_if.ts
+++ b/packages/common/src/directives/ng_if.ts
@@ -232,7 +232,8 @@ export class NgIf<T = unknown> {
    * The presence of this method is a signal to the Ivy template type-check compiler that the
    * `NgIf` structural directive renders its template with a specific context type.
    */
-  static ngTemplateContextGuard<T>(dir: NgIf<T>, ctx: any): ctx is NgIfContext<NonNullable<T>> {
+  static ngTemplateContextGuard<T>(dir: NgIf<T>, ctx: any):
+      ctx is NgIfContext<Exclude<T, false|0|''|null|undefined>> {
     return true;
   }
 }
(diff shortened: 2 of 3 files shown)


JoostK:
Hey @ranma42,

I didn't consider this case when opening #35125 as my understanding was that `NonNullable` was sufficient, but your example shows otherwise.

Could you please rebase so that CI can run?

JoostK:
Additionally, please change the commit scope from "ivy" to "common", as the "ivy" scope is no longer used (and will fail CI checks)

ranma42:
@JoostK thank you for your guidance :)
After following your suggestions I found out that some tests were failing because of the indentation in the goldens, but AFAICT they now look good.

ranma42:
I believe I have fixed the commit message and rebased as requested.
If this is this still blocked on my side, please ping me ;)

alxhub (approved):

ranma42:
Are the "failing" checks an issue in the changes?
Or are they just blocked on the review?

petebacondarwin:
The latter.

ranma42:
Any news? Do I need to rebase this on a more recent master?

petebacondarwin:
@ranma42 - please do rebase on master - I will ping @IgorMinar to do the public API review.

alxhub:
[Presubmit](http://test/OCL:308874029:BASE:314745928:1591288923246:9f8748fb)

jelbourn (approved):
LGTM

Reviewed-for: public-api

ranma42:
Is there anything I should do to make the PR process proceed?

jelbourn:
This is just awaiting other reviewers

petebacondarwin (commented):
Should we include `NaN` in the excluded list too?

JoostK:
> Should we include `NaN` in the excluded list too?

That's not a type AFAIK?

JoostK:
Indeed, NaN is only a value, and using `typeof NaN` would be the `number` type.

petebacondarwin:
Gosh!

ranma42:
Literal `NaN` is not supported (yet?) in typescript; it is being discussed/tracked in https://github.com/microsoft/TypeScript/issues/28682 but if possible I would not block this PR on that.

While the new typing is not yet exact, it is as narrow as possible with the current types supported by TS.

petebacondarwin (approved):
Reviewed-for: public-api

atscott (approved):
Reviewed-for: public-api

atscott:
[presubmit](http://test/OCL:323624358:BASE:323624400:1595963305431:1deca4a9)

PR #36628, opened 2020-04-14
build: reformat repo to new clang@1.4.0

josephperrott:
Upgrading to clang-format@1.4.0 would bring angular/angular to the latest available version of clang-format and include support for newer ts/js features like the ones noted in the background section.  However because the previous version of the clang-format the angular/angular repository was pinned to, 1.0.41, was nearly 4 years old the number of styling changes made by the formatter was greater than desired.  After running the new version of clang-format across the formatted code in angular/angular it was found that ~2500 files would be affected by the new formatting.  At the time, the presumption was this was too large of a single change to land and instead the change would need to be able to be rolled out incrementally.

The gulp tasks which run the formatter were updated to only format files which were found to be changed between the current HEAD of a ref and its previous HEAD.  This works out to be either all of the changes in a single PR being looked at, or all of the changes in a commit as it lands in a branch.  This allowed for the updated formatting to only happen in files already changed by a PR, rather than changing all ~2500 files in one change and invalidating nearly every outstanding PR.

However it was found that this created more noise than expected in each PR and instead our best action was to reformat the remaining formatted files in the repo in a single commit.

Patch version of #36613


josephperrott:
Caretaker: The lint failure is caused because the change is too large for gulpLint to process.

IgorMinar (approved):
on behalf of Angular leads

PR #36629, opened 2020-04-14
build: fix circular deps failure

josephperrott:
With the large scale refactoring of the repo to the new version of `clang-format`, some import orders were changed.

Specifically the imports found in [this range](https://github.com/angular/angular/blob/b01084910266deaa7d8e4424398062e6905199e8/packages/core/testing/src/r3_test_bed.ts#L37-L38).

The file previously read: 
```ts
import {TestBed} from './test_bed';
import {ComponentFixtureAutoDetect, ComponentFixtureNoNgZone, TestBedStatic, TestComponentRenderer, TestModuleMetadata} from './test_bed_common';
import {R3TestBedCompiler} from './r3_test_bed_compiler';
```

and now reads:
```ts
import {R3TestBedCompiler} from './r3_test_bed_compiler';
import {TestBed} from './test_bed';
import {ComponentFixtureAutoDetect, ComponentFixtureNoNgZone, TestBedStatic, TestComponentRenderer, TestModuleMetadata} from './test_bed_common';
```

This change in order cause the circular dependency to be entered earlier and changed the golden file for our circular deps discovery.

### DIFF
--- a/goldens/packages-circular-deps.json
+++ b/goldens/packages-circular-deps.json
@@ -1807,9 +1807,10 @@
     "packages/core/src/render3/interfaces/view.ts"
   ],
   [
-    "packages/core/testing/src/r3_test_bed.ts",
+    "packages/core/testing/src/r3_test_bed_compiler.ts",
     "packages/core/testing/src/test_bed_common.ts",
-    "packages/core/testing/src/test_bed.ts"
+    "packages/core/testing/src/test_bed.ts",
+    "packages/core/testing/src/r3_test_bed.ts"
   ],
   [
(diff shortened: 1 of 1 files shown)


AndrewKushnir:
@josephperrott should this PR target both master and patch (patch branch seems to have the same problem)?

josephperrott:
@AndrewKushnir Correct, clicked the wrong one.

devversion (approved):
LGTM. I will have a look tomorrow on how we can improve this. It's okay that cycles shift if a file is visited earlier by the analyzer, but it's surprising that the length of the cycle changed.

kara (approved):
LGTM

PR #36630, opened 2020-04-14
build: move circular deps golden to a subfolder

josephperrott:
Moves the circular deps golden for packages into a subfolder of goldens,
`/goldens/circular-deps/` to more easily target the files for
ownership.


IgorMinar (approved):
lgtm. thanks!

PR #36632, opened 2020-04-14
Require commit message body to be at least 100 characters

josephperrott:

### DIFF
--- a/dev-infra/commit-message/validate.ts
+++ b/dev-infra/commit-message/validate.ts
@@ -82,20 +82,29 @@ export function validateCommitMessage(
   ////////////////////////////////////
   // Checking revert, squash, fixup //
   ////////////////////////////////////
+
+  // All revert commits are considered valid.
   if (commit.isRevert) {
     return true;
   }
 
-  if (commit.isSquash && options.disallowSquash) {
-    error('The commit must be manually squashed into the target commit');
-    return false;
+  // All squashes are considered valid, as the commit will be squashed into another in
+  // the git history anyway, unless the options provided to not allow squash commits.
+  if (commit.isSquash) {
+    if (options.disallowSquash) {
+      error('The commit must be manually squashed into the target commit');
+      return false;
+    }
+    return true;
   }
 
-  // If it is a fixup commit and `nonFixupCommitHeaders` is not empty, we only care to check whether
-  // there is a corresponding non-fixup commit (i.e. a commit whose header is identical to this
-  // commit's header after stripping the `fixup! ` prefix).
-  if (commit.isFixup && options.nonFixupCommitHeaders) {
-    if (!options.nonFixupCommitHeaders.includes(commit.header)) {
+  // Fixups commits are considered valid, unless nonFixupCommitHeaders are provided to check
+  // against. If `nonFixupCommitHeaders` is not empty, we check whether there is a corresponding
+  // non-fixup commit (i.e. a commit whose header is identical to this commit's header after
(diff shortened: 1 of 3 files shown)


kara (approved):
LGTM

IgorMinar (approved):

petebacondarwin:
Does the min length rule include WIP commits?

dev-infra/commit-message/validate.ts
```diff
@@ -20,7 +20,7 @@ const SQUASH_PREFIX_RE = /^squash! /i;
 const REVERT_PREFIX_RE = /^revert:? /i;
 const TYPE_SCOPE_RE = /^(\w+)(?:\(([^)]+)\))?\:\s(.+)$/;
 const COMMIT_HEADER_RE = /^(.*)/i;
-const COMMIT_BODY_RE = /^.*\n\n(.*)/i;
+const COMMIT_BODY_RE = /^.*\n\n((.*|\n)*)/i;
```
gkalpak, line 23: Won't matter much here, but this can lead to [catastrophic backtracking][1] in non-matching commit messages.
A better RegExp would be:

```suggestion
const COMMIT_BODY_RE = /^.*\n\n([\s\S]*)$/;
```

...or:

```suggestion
const COMMIT_BODY_RE = /^.*\n\n([^.]*)$/i;
```

##
In modern JS, one can also use the [dotAll (s) flag][2] to make `.` match newlines as well, but here it would make the first `.*` also match newlines, so you would need something like: `/^[\n]*\n\n(.*)$/s`

[1]: https://www.rexegg.com/regex-explosive-quantifiers.html
[2]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/dotAll

gkalpak (approved):

josephperrott:
@petebacondarwin Yes, this would require that WIP commits also contain 100 characters in the body.  We actually do not have any specific provisions for WIP commits.

josephperrott:
@gkalpak I had to rearrange handling of  `fixup`, `revert` and `squash` commits a little bit, and update the tests in response.  Can you take another look?

dev-infra/commit-message/validate.ts
```diff
@@ -79,20 +79,32 @@ export function validateCommitMessage(
...
-    error('The commit must be manually squashed into the target commit');
-    return false;
+  // All squashes are considered valid, as the commit will be squashed into another in
+  // the git history anyway, unless the options provided to not allow squash commits.
+  if (commit.isSquash) {
+    if (options.disallowSquash) {
+      error('The commit must be manually squashed into the target commit');
+      return false;
+    }
+    return true;
```
gkalpak, line 98: Not a big fan of these early returns (here and for fixup too).
Why not validate the commit message header as before?

gkalpak (changes requested):

dev-infra/commit-message/validate.ts
josephperrott: Personally, I think its much easier to reason about rather than having an escape based on something we already checked about earlier in the process.  But no harm in checking the message headers, I reverted the fixup I had that added the early returns.

gkalpak: You have a point about being easier to reason about. I am fine either way :grin:
So, feel free to leave as is or put 96f36c74bb28a4056c17920ff46692f8ac821a45 back in.

gkalpak (approved):
LGTM :+1:

Adding the `cleanup` label so you get a chance to act on https://github.com/angular/angular/pull/36632#discussion_r413332577 if you want to.

PR #36633, opened 2020-04-15
refactor(common): remove WrappedValue from AsyncPipe

### LINKED ISSUE #29927: RFC: AsyncPipe doesn't necessarily need WrappedValue
# 🐞 bug report

### Affected Package

 The issue is caused by package @angular/common

### Is this a regression?

No, it's likely a design failure from the very beginning.

### Description

In [`WrappedValue`](https://angular.io/api/core/WrappedValue) documentation, it's saying:

> Indicates that the result of a [`Pipe`](https://angular.io/api/core/Pipe) transformation has changed even though the reference has not changed.

But inside [`AsyncPipe`](https://angular.io/api/common/AsyncPipe) whether or not returning an `WrappedValue` is fully based on reference changed or not:

https://github.com/angular/angular/blob/9147092a156203d46cd1c865e06fa72a1e2d4137/packages/common/src/pipes/async_pipe.ts#L106-L111

It will only wrap the first changed value into `WrappedValue`, all the subsequent unchanged values will retain unwrapped, even changing the observable reference won't make any effect:

https://stackblitz.com/edit/typescript-bzkarv

Otherwise, even the same value being re-emitted many times during change detection, it won't be wrapped and no error is reported:

https://stackblitz.com/edit/angular-kqkpym?file=src%2Fapp%2Fapp.component.ts

I don't know what's the difference if `AsyncPipe` just return the unwrapped value and let the change detector to check it. Please correct me if I am wrong.

And then, what's the meaning of `WrappedValue` as not need by Angular itself?

## 🔬 Minimal Reproduction

Linked above.

## 🔥 Exception or Error
<pre><code>N/A
</code></pre>

## 🌍  Your Environment

**Angular Version:**
<pre><code>N/A
</code></pre>

**Anything else relevant?**

The question was found when digging into https://github.com/angular/angular/pull/22305.

atscott:
[global presubmit](https://test.corp.google.com/ui#id=OCL:306275934:BASE:306434549:1586898352449:9c35923d)

AsyncPipe only uses `WrappedValue` when the latest value from the `Promise` or `Observable` is different from the previous one. This is already enough to trigger change detection so the `WrappedValue` is not necessary.

Fixes #29927

BREAKING CHANGE: This change could result in
ExpressionChangedAfterItHasBeenChecked errors that were not detected
before. The error could previously have gone undetected because two
`WrappedValues` are considered "equal" in all cases for the purposes of
the check, even if their respective unwrapped values are not.

### DIFF
--- a/packages/common/src/pipes/async_pipe.ts
+++ b/packages/common/src/pipes/async_pipe.ts
@@ -6,7 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {ChangeDetectorRef, EventEmitter, OnDestroy, Pipe, PipeTransform, WrappedValue, ɵisObservable, ɵisPromise, ɵlooseIdentical} from '@angular/core';
+import {ChangeDetectorRef, EventEmitter, OnDestroy, Pipe, PipeTransform, ɵisObservable, ɵisPromise} from '@angular/core';
 import {Observable, SubscriptionLike} from 'rxjs';
 import {invalidPipeArgumentError} from './invalid_pipe_argument_error';
 
@@ -81,7 +81,6 @@ const _observableStrategy = new ObservableStrategy();
 @Pipe({name: 'async', pure: false})
 export class AsyncPipe implements OnDestroy, PipeTransform {
   private _latestValue: any = null;
-  private _latestReturnedValue: any = null;
 
   private _subscription: SubscriptionLike|Promise<any>|null = null;
   private _obj: Observable<any>|Promise<any>|EventEmitter<any>|null = null;
@@ -104,7 +103,6 @@ export class AsyncPipe implements OnDestroy, PipeTransform {
       if (obj) {
         this._subscribe(obj);
       }
-      this._latestReturnedValue = this._latestValue;
       return this._latestValue;
     }
 
@@ -113,12 +111,7 @@ export class AsyncPipe implements OnDestroy, PipeTransform {
       return this.transform(obj as any);
     }
 
-    if (ɵlooseIdentical(this._latestValue, this._latestReturnedValue)) {
-      return this._latestReturnedValue;
-    }
-
-    this._latestReturnedValue = this._latestValue;
-    return WrappedValue.wrap(this._latestValue);
+    return this._latestValue;
   }
 
   private _subscribe(obj: Observable<any>|Promise<any>|EventEmitter<any>): void {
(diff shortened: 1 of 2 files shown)


mhevery (changes requested):
This is an excellent start, but I think we can do more cleanup:
- [ ] `packages/core/src/util/WrappedValue.ts` file can be totally deleted (as it is a duplicate of `packages/core/src/change_detection/change_detection_util.ts`
- [ ] `packages/core/src/change_detection/change_detection_util.ts` should have `WrappedValue` deleted (that is the version which is actually used.)
- [ ] The only thing we care about is `WrappedValue.wrap` as that is the only place where wrapping can happen. `async` was the only place where the wrapping was happening. This means that there is no more wrapping in production, which means that it is safe to remove all calls to `WrappedValue.unwrap` and `WrapedValue.isWrapped`.
  -  `packages/common/src/pipes/async_pipe.ts` which you have already removed
  - `packages/core/test/linker/change_detection_integration_spec.ts` spec file which needs to be cleaned up. 
- [ ] Remove all calls to `WrappedValue.unwrap` and `WrapedValue.isWrapped` per comment above.

atscott:
@mhevery, I think we should discuss the removal of `WrappedValue` from other places as part of a different PR. Because `WrappedValue` is public API, there dependencies on it in the wild that we would need to have a plan for. I think removing `WrappedValue` from the async pipe is a separate issue, as it's not being used in a way that `WrappedValue` was designed for.

mhevery (approved):

AndrewKushnir:
Discussed with @atscott, the plan is:
- start another Global TAP on Sunday
- merge this PR and sync to g3 separately (only this change) on Monday (4/27)

I will put a "blocked" label for now to indicate that it should not be merged until we re-run Global TAP.

AndrewKushnir:
- [Blueprint presubmit](http://test/OCL:308541527:BASE:308541533:1587952070490:788e5336)
- [Global presubmit](https://test.corp.google.com/ui#id=OCL:308541527:BASE:308541614:1587952149686:d5d6fa23)

AndrewKushnir:
FYI, Global Presubmit indicated that there is one failing target that needs further investigation, so keeping the "blocked" label for now.

AndrewKushnir:
FYI, removing the "merge" label for now, since further investigation revealed a use-case that would require additional discussion/attention.

AndrewKushnir:
[Global presubmit #2](https://test.corp.google.com/OCL:308935766:BASE:308942854:1588125939472:920f12a0)

Issue #35826, opened 2020-03-03
Directive ngOnInit is called after parent component template is checked for changes in Ivy


ashley-hunter:
# 🐞 bug report

### Affected Package

I assume `@angular/compiler`

### Is this a regression?

Yes, this works when using View Engine.

### Description

I have put together a very simple reproduction, for example we have a directive that is exported:

```typescript
@Directive({
  selector: '[appItemSelected]',
  exportAs: 'item-selected'
})
export class ItemSelectedDirective {

  isSelected: boolean;

  ngOnInit(): void {
    this.isSelected = true;
  }

}
```

And we use the exported reference in a template, to toggle a class (which turns the text blue) based on the value of isSelected:
```html
<button [class.selected]="dir.isSelected">
  <div appItemSelected #dir="item-selected">Content</div>
</button>
```

In View Engine this works fine, the class is applied and the button appears with blue text:
https://stackblitz.com/edit/angular-cysqo9?file=src%2Fapp%2Fapp.component.html

In Ivy however this is not applied as the change detection pass is run before `ngOnInit` in the directive is run.

https://stackblitz.com/edit/angular-v2lxz8?file=src%2Fapp%2Fapp.component.html

This may be an expected change in Ivy, I didn't see it mentioned in the breaking changes. Its not a blocker for us as we can workaround it but it is a regression.

Thanks in advance!

## 🔬 Minimal Reproduction

VE (working): https://stackblitz.com/edit/angular-cysqo9?file=src%2Fapp%2Fapp.component.html
Ivy (not working): https://stackblitz.com/edit/angular-v2lxz8?file=src%2Fapp%2Fapp.component.html

## 🌍  Your Environment

**Angular Version:**
<pre><code>
@angular-devkit/architect         0.900.3
@angular-devkit/build-angular     0.900.3
@angular-devkit/build-optimizer   0.900.3
@angular-devkit/build-webpack     0.900.3
@angular-devkit/core              9.0.3
@angular-devkit/schematics        9.0.3 (cli-only)
@angular/cdk                      9.1.0
@ngtools/webpack                  9.0.3
@schematics/angular               9.0.3 (cli-only)
@schematics/update                0.900.3 (cli-only)
ng-packagr                        9.0.1
rxjs                              6.5.4
typescript                        3.7.5
webpack                           4.41.6

</code></pre>

**Anything else relevant?**

pkozlowski-opensource:
[...]

PR #36635, opened 2020-04-15
docs: template ref variables new behaviour documented

### LINKED ISSUE #34821: [ivy] using a template reference before its declaration causes ExpressionChangedAfterItHasBeenCheckedError
# 🐞 bug report

### Affected Package
Angular 9-rc.9 with Ivy

### Is this a regression?

This issue is present only when Ivy is enabled. I have to say that I am not 100% sure it's a bug. It could be the pattern we have been using might not have been supported by Angular in the first place. But I have my own doubts as I explained below.

### Description

Below is a simplified reproduction of the pattern that's causing the `ExpressionChangedAfterItHasBeenCheckedError` error. The following snippet would work just in any Angular version as long as Ivy is not enabled:
```html
<h1>Greetings, {{hello.name}}!</h1>
<hello-comp #hello [name]="'John'"></hello-comp>
```
As you can see, I'm simply template-referencing a component and interpolating its `name` property right above it. 

One thing to note here is that if you move the interpolation down below the `hello-comp`, I see there is no error:
```html
<hello-comp #hello [name]="'John'"></hello-comp>
<h1>Greetings, {{hello.name}}!</h1>
```

So basically, Ivy is not happy when the template reference is used before its declaration. However, we had been using this pattern without any issues before Ivy. Also, I suspected that this pattern (using a template reference before its declaration) could have been something Angular has never supported. But in the documentation, I read these:

- _The scope of a reference variable is the entire template._
- _You can refer to a template reference variable anywhere in the component's template_

So it seems to me at least using a template reference before its declaration has always been supported/allowed. But when Ivy is enabled, it starts throwing the error. As such, I am suspecting this could be another ivy related bug.

## 🔬 Minimal Reproduction

https://stackblitz.com/edit/angular-8-template-reference-interpolation

ajitsinghkaler:
Template ref variable's scope used to be the whole template but in Ivy it changed now template reference variables cannot be used before its declaration in template . Documented this behaviour  in template-sytax.md file.

Fixes #34821, #35826

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

No documentation of new behaviour for template ref variables.

Issue Number: #34821, #35826

## What is the new behavior?
Documented new behavior for template ref variables.

- [ ] Yes
- [x] No

### DIFF
--- a/aio/content/guide/template-syntax.md
+++ b/aio/content/guide/template-syntax.md
@@ -1815,9 +1815,27 @@ The scope of a reference variable is the entire template. So, don't define the s
 You can use the `ref-` prefix alternative to `#`.
 This example declares the `fax` variable as `ref-fax` instead of `#fax`.
 
-
 <code-example path="template-reference-variables/src/app/app.component.html" region="ref-fax" header="src/app/app.component.html"></code-example>
 
+#### Breaking Change (Ivy)
+
+In the View Engine compiler when a template reference variable is declared it can be used any where in the template ie scope of these variables was the entire template.
+
+In Ivy the binding evaluation order is changed to have a simpler / more predictable evaluation order. In this new implementation template reference variables cannot be used before its declaration. For example
+
+```html
+  <h1>Greetings, {{hello.name}}</h1>
+  <hello-comp #hello [name]="'John'"></hello-comp>
+```
+
+In Ivy this does not work.
+
+The correct sequence for using template reference variable in Ivy is
+
+```html
+  <hello-comp #hello [name]="'John'"></hello-comp>
+  <h1>Greetings, {{hello.name}}</h1>
+```
 
 <hr/>
 


aio/content/guide/template-syntax.md
```diff
@@ -1815,9 +1815,27 @@ The scope of a reference variable is the entire template. So, don't define the s
...
+  <hello-comp #hello [name]="'John'"></hello-comp>
+```
+
+In Ivy this does not work.
+
+The correct sequence for using template reference variable in Ivy is
+
+```html
+  <h1>Greetings, {{hello.name}}</h1>
+  <hello-comp #hello [name]="'John'"></hello-comp>
```
mlc-mlapis, line 1837: Probably a mistake? This exactly the same as the previous incorrect example.

ajitsinghkaler: Thnks @mlc-mlapis

ajitsinghkaler:
closing this in favour of #36643

PR #36637, opened 2020-04-15
ngcc: new cluster worker entry-point

petebacondarwin:
This refactoring simplifies the clustering by providing a new entry-point file for cluster workers.

This reduces the complexity of the code (not having to check for `cluster.isMaster` so much) and also removes the need to do parse and execute as much code in the worker processes.


packages/compiler-cli/ngcc/src/execution/cluster/master.ts
```diff
@@ -10,7 +10,7 @@
 
 import * as cluster from 'cluster';
 
-import {resolve} from '../../../../src/ngtsc/file_system';
+import {FileSystem} from '../../../../src/ngtsc/file_system';
```
gkalpak, line 13: OOC, why this change?

packages/compiler-cli/ngcc/src/locking/lock_file_with_child_process/index.ts
```diff
@@ -82,14 +82,15 @@ export class LockFileWithChildProcess implements LockFile {
...
-
+    const isWindows = process.platform === 'win32';
+    const unlocker = fork(
+                         this.fs.resolve(__dirname, './unlocker.js'), [path, logLevel],
+                         {detached: true, stdio: isWindows ? 'pipe' : 'inherit'}) as
+        ChildProcessByStdio<Writable, Readable, Readable>;
+    if (isWindows) {
+      unlocker.stdout.on('data', process.stdout.write.bind(process.stdout));
+      unlocker.stderr.on('data', process.stderr.write.bind(process.stderr));
+    }
```
gkalpak, line 93: The `ChildProcessByStdio<Writable, Readable, Readable>` type is not correct any more. (It only has stderr/stdout on Windows.)

You could check for `stderr`/`stdout` explicitly:

```suggestion
    const unlocker = fork(
                         this.fs.resolve(__dirname, './unlocker.js'), [path, logLevel],
                         {detached: true, stdio: isWindows ? 'pipe' : 'inherit'});
    if (unlocker.stdout !== null && unlocker.stderr !== null) {
      unlocker.stdout.on('data', process.stdout.write.bind(process.stdout));
      unlocker.stderr.on('data', process.stderr.write.bind(process.stderr));
    }
```

Or even:

```ts
unlocker.stdout?.on('data', process.stdout.write.bind(process.stdout));
unlocker.stderr?.on('data', process.stderr.write.bind(process.stderr));
```

packages/compiler-cli/ngcc/src/main.ts
```diff
@@ -105,11 +104,7 @@ export function mainNgcc({
     return;
   }
 
-  // NOTE: To avoid file corruption, ensure that each `ngcc` invocation only creates _one_ instance
-  //       of `PackageJsonUpdater` that actually writes to disk (across all processes).
-  //       This is hard to enforce automatically, when running on multiple processes, so needs to be
-  //       enforced manually.
```
gkalpak, line 245: I still think this comment is valuable.

packages/compiler-cli/ngcc/test/execution/cluster/package_json_updater_spec.ts
```diff
@@ -23,204 +23,175 @@ runInEachFileSystem(() => {
...
       processSendSpy = jasmine.createSpy('process.send');
       mockProcessSend(processSendSpy);
+    });
 
-      delegate = new MockPackageJsonUpdater();
-      updater = new ClusterPackageJsonUpdater(delegate);
+    describe('constructor()', () => {
+      it('should throw an error if used on a cluster master', () => {
+        runAsClusterMaster(true);
+        expect(() => new ClusterWorkerPackageJsonUpdater()).toThrowError('');
```
gkalpak, line 35: Why use an empty string? We should either use the actual error or not pass an argument at all.

packages/compiler-cli/ngcc/src/execution/cluster/worker.ts
```diff
@@ -5,58 +5,102 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
-
 /// <reference types="node" />
 
+import {readConfiguration} from '@angular/compiler-cli/src/perform_compile';
```
gkalpak, line 10: Shouldn't this be a relative import?

packages/compiler-cli/ngcc/test/execution/cluster/worker_spec.ts
```diff
@@ -35,131 +35,116 @@ describe('ClusterWorker', () => {
...
+    beforeEach(() => runAsClusterMaster(true));
 
-      it('should throw an error', () => {
-        expect(() => new ClusterWorker(mockLogger, createCompileFnSpy))
-            .toThrowError('Tried to instantiate `ClusterWorker` on the master process.');
-        expect(createCompileFnSpy).not.toHaveBeenCalled();
-      });
+    it('should throw an error', async () => {
+      await expectAsync(runWorker(mockLogger, createCompileFnSpy))
+          .toBeRejectedWithError('Tried to run cluster worker on the master process.');
```
gkalpak, line 43: Jasmine 3.5 #ftw :tada:

packages/compiler-cli/ngcc/src/execution/cluster/worker.ts
```diff
@@ -5,58 +5,102 @@
...
+  process.title = 'ngcc (worker)';
+  setFileSystem(new CachedFileSystem(new NodeJSFileSystem()));
+  let {
+    basePath,
+    targetEntryPointPath,
+    createNewEntryPointFormats = false,
+    logger = new ConsoleLogger(LogLevel.info),
+    pathMappings,
+    errorOnFailedEntryPoint = false,
+    enableI18nLegacyMessageIdFormat = true,
```
gkalpak, line 38: I would rather bake the defaults into `parseCommandLineOptions()`. Otherwise, we may end up with different values on the master vs workers with unexpected behavior (in the future).

WDYT?

packages/compiler-cli/ngcc/src/execution/cluster/worker.ts
```diff
@@ -5,58 +5,102 @@
...
-export class ClusterWorker {
-  private compile: CompileFn;
+      const fileSystem = getFileSystem();
+      const absBasePath = absoluteFrom(basePath);
+      const projectPath = fileSystem.dirname(absBasePath);
+      const tsConfig =
+          tsConfigPath !== null ? readConfiguration(tsConfigPath || projectPath) : null;
+
+      if (pathMappings === undefined) {
+        pathMappings = getPathMappingsFromTsConfig(tsConfig, projectPath);
```
gkalpak, line 55: I am little worried about having identical logic in two places. Everything in this `try` block up to this point is also done in `main.ts`. Instead of having to keep the logic consistent between two places, I would rather have it in one place and share it between master/workers.

gkalpak (approved):
Yay! 🎊 👌 💯

packages/compiler-cli/ngcc/src/execution/cluster/master.ts
petebacondarwin: The primary reason was that I started using `resolve()` in the constructor of the `ClusterMaster` and tests started failing because the tests had not called `setFileSystem()`. While I could fix the tests, it made me realise that if this class is relying upon `FileSystem` being setup we ought to make that dependency explicit.

packages/compiler-cli/ngcc/src/locking/lock_file_with_child_process/index.ts
petebacondarwin: Really like the second option! Thanks.

packages/compiler-cli/ngcc/src/main.ts
petebacondarwin: Yes but it doesn't feel right to have it there... since the PackageJsonUpdater` that is used in `main` will always write to disk... 

I will put a version of this comment in `worker.ts` instead. OK?

packages/compiler-cli/ngcc/test/execution/cluster/package_json_updater_spec.ts
petebacondarwin: Oops!

packages/compiler-cli/ngcc/src/execution/cluster/worker.ts
petebacondarwin: Yep!

packages/compiler-cli/ngcc/test/execution/cluster/worker_spec.ts
petebacondarwin: 💪

packages/compiler-cli/ngcc/src/execution/cluster/worker.ts
petebacondarwin: The trouble with that is that the sync form of the CLI calls `mainNgcc()` directly and so we need to keep that API stable.

packages/compiler-cli/ngcc/src/execution/cluster/worker.ts
petebacondarwin: Yeah, this bothered me too. I will see if I can make it more DRY.

petebacondarwin: Done.

packages/compiler-cli/ngcc/src/execution/cluster/worker.ts
petebacondarwin: I moved all the default calculations into the `setupShared()` function so that should keep main and workers aligned.

PR #36556, opened 2020-04-10
docs: add note on publishing libraries in ivy 

ajitsinghkaler:
Libraries are still built with view engine even after Ivy being the default engine for angular apps stating that explicitly and the reason for it.

Fixes #35625

## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [ ] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [x] Documentation content changes
- [ ] angular.io application / infrastructure changes
- [ ] Other... Please describe:

## What is the current behavior?

Libries built using VE not stated explicitly

Issue Number: N/A

## What is the new behavior?
Libraries built using VE stated explicitly

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

PR #36643, opened 2020-04-15
docs: add example to illustrate binding order differences in ivy

crisbeto:
Adds some examples and an explanation about the differences in binding order between Ivy and ViewEngine.

### DIFF
--- a/aio/content/guide/ivy-compatibility-examples.md
+++ b/aio/content/guide/ivy-compatibility-examples.md
@@ -321,18 +321,21 @@ In Ivy, the `myDir.name` binding will throw an `ExpressionChangedAfterItHasBeenC
 
 ### Background
 
-In the ViewEngine runtime, input property bindings and regular data bindings were executed in different stages. This means that Angular would go through the template once and assign the `name` binding, as well as any other directive inputs. Afterwards it would go through the template again and execute the `myDir.name` binding and any other data bindings.
+In the ViewEngine runtime, directive input bindings and element bindings were executed in different stages. Angular would process the template one full time to check directive inputs only (e.g. `[name]`), then process the whole template again to check element and text bindings only (e.g.`{{ myDir.name }}`). This meant that the `name` directive input would be checked before the `myDir.name` text binding despite their relative order in the template, which some users felt to be counterintuitive.
 
-On the other hand, Ivy will go through the template once and process both the `myDir.name` and `name` bindings during the same pass. In practice, this means that the `myDir.name` binding will have `undefined` as its initial value and then get the value from `myDir` once change detection runs again.
+In contrast, Ivy processes the template in just one pass, so that bindings are checked in the same order that they are written in the template.   In this case, it means that the `myDir.name` binding will have be checked before the `name` input sets the property on the directive (and thus it will be `undefined`). Since the `myDir.name` property will be set by the time the next change detection pass runs, a change detection error is thrown.
 
 ### Example of error
 
-Assuming that the value for `myName` is `Angular`, you should see an error that looks like `Error: ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked. Previous value: 'undefined'. Current value: 'Angular'`.
+Assuming that the value for `myName` is `Angular`, you should see an error that looks like
 
+```
+Error: ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked. Previous value: 'undefined'. Current value: 'Angular'.
+```
 
 ### Recommended fix
 
-To fix this problem, we recommend either getting the information for the binding from a different place (e.g. the `myName` property from our example) or to put the data binding after the directive has been declared so that the initial value is available on the first pass.
+To fix this problem, we recommend either getting the information for the binding directly from the host component (e.g. the `myName` property from our example) or to move the data binding after the directive has been declared so that the initial value is available on the first pass.
 
 *Before*
 ```html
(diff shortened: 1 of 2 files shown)


ajitsinghkaler:
We are working on the same thing  #36556 I'm closing my pull request in favour of this because it's more detailed than my documentation tell me if you found anything else

aio/content/guide/ivy-compatibility-examples.md
```diff
@@ -291,4 +291,57 @@ To fix this problem, we recommend binding to the `selected` property on the `<op
...
+<div myDir #myDir="myDir" [name]="myName"></div>
+```
+
+In the View Engine runtime, the above code would print out the name without any errors.
+In Ivy, the `myDir.name` binding will throw an `ExpressionChangedAfterItHasBeenCheckedError`.
+
+
+### Background
+
+In the ViewEngine runtime, input property bindings and regular data bindings were executed in different stages. This means that Angular would go through the template once and assign the `name` binding, as well as any other directive inputs. Afterwards it would go through the template again and execute the `myDir.name` binding and any other data bindings.
```
kara, line 324:
```suggestion
In the ViewEngine runtime, directive input bindings and element bindings were executed in different stages. Angular would process the template one full time to check directive inputs only (e.g. `[name]`), then process the whole template again to check element and text bindings only (e.g.`{{ myDir.name }}`). This meant that the `name` directive input would be checked before the `myDir.name` text binding despite their relative order in the template, which some users felt to be counterintuitive.
```

aio/content/guide/ivy-compatibility-examples.md
```diff
@@ -291,4 +291,57 @@ To fix this problem, we recommend binding to the `selected` property on the `<op
...
+
+On the other hand, Ivy will go through the template once and process both the `myDir.name` and `name` bindings during the same pass. In practice, this means that the `myDir.name` binding will have `undefined` as its initial value and then get the value from `myDir` once change detection runs again.
```
kara, line 326:
```suggestion
In contrast, Ivy processes the template in just one pass, so that bindings are checked in the same order that they are written in the template.   In this case, it means that the `myDir.name` binding will have be checked before the `name` input sets the property on the directive (and thus it will be `undefined`). Since the `myDir.name` property will be set by the time the next change detection pass runs, a change detection error is thrown.
```

aio/content/guide/ivy-compatibility-examples.md
```diff
@@ -291,4 +291,57 @@ To fix this problem, we recommend binding to the `selected` property on the `<op
...
+
+### Example of error
+
+Assuming that the value for `myName` is `Angular`, you should see an error that looks like `Error: ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked. Previous value: 'undefined'. Current value: 'Angular'`.
+
+
+### Recommended fix
+
+To fix this problem, we recommend either getting the information for the binding from a different place (e.g. the `myName` property from our example) or to put the data binding after the directive has been declared so that the initial value is available on the first pass.
```
kara, line 335:
```suggestion
To fix this problem, we recommend either getting the information for the binding directly from the host component (e.g. the `myName` property from our example) or to move the data binding after the directive has been declared so that the initial value is available on the first pass.
```

kara (changes requested):
Thanks for picking this up! Some suggestions below.

aio/content/guide/ivy-compatibility-examples.md
```diff
@@ -291,4 +291,57 @@ To fix this problem, we recommend binding to the `selected` property on the `<op
...
+
+### Background
+
+In the ViewEngine runtime, input property bindings and regular data bindings were executed in different stages. This means that Angular would go through the template once and assign the `name` binding, as well as any other directive inputs. Afterwards it would go through the template again and execute the `myDir.name` binding and any other data bindings.
+
+On the other hand, Ivy will go through the template once and process both the `myDir.name` and `name` bindings during the same pass. In practice, this means that the `myDir.name` binding will have `undefined` as its initial value and then get the value from `myDir` once change detection runs again.
+
+### Example of error
+
+Assuming that the value for `myName` is `Angular`, you should see an error that looks like `Error: ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked. Previous value: 'undefined'. Current value: 'Angular'`.
```
kara, line 330: Can we move the error down into its own code block? Hard to scan if it's inline:

```suggestion
Assuming that the value for `myName` is `Angular`, you should see an error that looks like 

```
Error: ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked. Previous value: 'undefined'. Current value: 'Angular'`.
```

**Comment by crisbeto** [comment]

File: aio/content/guide/ivy-compatibility-examples.md [hunk 5]
```diff
@@ -291,4 +291,60 @@ To fix this problem, we recommend binding to the `selected` property on the `<op
...
+
+In the View Engine runtime, the above code would print out the name without any errors.
+In Ivy, the `myDir.name` binding will throw an `ExpressionChangedAfterItHasBeenCheckedError`.
+
+
+### Background
+
+In the ViewEngine runtime, directive input bindings and element bindings were executed in different stages. Angular would process the template one full time to check directive inputs only (e.g. `[name]`), then process the whole template again to check element and text bindings only (e.g.`{{ myDir.name }}`). This meant that the `name` directive input would be checked before the `myDir.name` text binding despite their relative order in the template, which some users felt to be counterintuitive.
+
+In contrast, Ivy processes the template in just one pass, so that bindings are checked in the same order that they are written in the template.   In this case, it means that the `myDir.name` binding will have be checked before the `name` input sets the property on the directive (and thus it will be `undefined`). Since the `myDir.name` property will be set by the time the next change detection pass runs, a change detection error is thrown.
```
kara, line 326: ### 
```suggestion
In contrast, Ivy processes the template in just one pass, so that bindings are checked in the same order that they are written in the template.  In this case, it means that the `myDir.name` binding will be checked before the `name` input sets the property on the directive (and thus it will be `undefined`). Since the `myDir.name` property will be set by the time the next change detection pass runs, a change detection error is thrown.
```

kara (approved):
Thanks for the updates. Just one more typo fix, and LGTM

matsko:
@crisbeto this fails on patch. Please open up a new PR and target patch directly there.

PR #36646, opened 2020-04-15
Update lexer.ts

vmagalhaes:
## PR Type

Helpful

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes
- [ ] CI related changes

https://github.com/ng-packagr/ng-packagr/issues/976

- [ ] Yes
- [x] No

## Other information
Problem solved by @johnnyprogger

### DIFF
--- a/packages/compiler/src/ml_parser/lexer.ts
+++ b/packages/compiler/src/ml_parser/lexer.ts
@@ -110,7 +110,11 @@ export interface TokenizeOptions {
 export function tokenize(
     source: string, url: string, getTagDefinition: (tagName: string) => TagDefinition,
     options: TokenizeOptions = {}): TokenizeResult {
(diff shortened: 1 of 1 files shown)


vmagalhaes:
@googlebot I signed it!

JoostK:
I am sorry to tell you that this is not a proper fix. The typings of Angular require a non-null `string` to be returned from `CompilerHost.readResource`:

https://github.com/angular/angular/blob/0a69a2832b0d6d08ee4b301517e447dd3232339b/packages/compiler-cli/src/ngtsc/core/api/src/interfaces.ts#L45

It looks like [ng-packagr returns `undefined`](https://github.com/ng-packagr/ng-packagr/blob/7a8c8d5d3e2599cb940777d683974d36c9cc300c/src/lib/ts/cache-compiler-host.ts#L125-L138) when a file is not available, which is not allowed according to the expected return type. Solving that in ng-packagr would be the proper solution.

PR #36648, opened 2020-04-15
Revert "fix(common): `locales/global/*.js` are not ES5 compliant (#36…

atscott:
…342)"

This reverts commit c8f3fa9f3e2dd41b4e15b38786268093ce556557.


IgorMinar (approved):
Lgtm. As discussed, we want to roll this back in the patch branch only and use the next.2 release to validate that this change is desirable. Then we can consider cherry picking it back to 9.1.x.

IgorMinar (commented):

IgorMinar (approved):

PR #36649, opened 2020-04-15
fix(core): properly identify modules affected by overrides in TestBed

### LINKED ISSUE #36619: TestBed.overrideModule affects other tests
# 🐞 bug report

### Affected Package

### Is this a regression?

 Yes, the previous version in which this bug was not present was: 8.2.13

### Description

All test cases after a call to `TestBed.overrideModule` seem to be affected by its modifications.

In the provided Example project I have two tests: `one.spec.ts` and `two.spec.ts`. Both create an instance of their own component which both contain a `TestComponent`. In `one.spec.ts` I replace `TestComponent` with a Mock using `overrideModule`. If `one.spec.ts` is executed first by the test runner, `two.spec.ts` also uses the Mock component.

Additionally this example also does not work if `two.spec.ts` is executed first. In that case the `overrideModule`-call seems to be ignored.

I cannot say for sure wether this is a problem with Angular or Karma. If you think I should open an issue on Karma, please let me know.

## 🔬 Minimal Reproduction

https://github.com/kalsowerus/repro-app

The `TestComponent` I mentioned earlier prints `'test'` in its `ngOnInit()`, while the mock component inside `one.spec.ts` prints `'mock'`. To see the problem just `npm install` and then `ng test`. The exact command I used to execute the tests is `ng test --browsers ChromeForWSL --watch=false`. `ChromeForWSL` being a custom Launcher based on Chrome which disables the "VizDisplayCompositor"-feature since that does not seem to work in a WSL (Windows subsystem for Linux) environment.

The expected result would be seeing both `'test'` and `'mock'` in the console once in any order. But depending on the execution order either of them are printed twice while the other one is not printed at all.

## 🔥 Exception or Error
No exception or error

## 🌍  Your Environment

**Angular Version:**
<pre><code>
Angular CLI: 9.1.1
Node: 12.14.0
OS: linux x64

Angular: 9.1.1
... router
Ivy Workspace: Yes

@angular-devkit/architect         0.901.1
@angular-devkit/build-angular     0.901.1
@angular-devkit/build-optimizer   0.901.1
@angular-devkit/build-webpack     0.901.1
@angular-devkit/core              9.1.1
@angular-devkit/schematics        9.1.1
@ngtools/webpack                  9.1.1
@schematics/angular               9.1.1
@schematics/update                0.901.1
</code></pre>

**Anything else relevant?**

AndrewKushnir:
When module overrides (via `TestBed.overrideModule`) are present, it might affect all modules that import (even transitively) an overridden one. For all affected modules we need to recalculate their scopes for a given test run and restore original scopes at the end. Prior to this change, we were recalculating module scopes only for components that are used in a test, without taking into account module hierarchy. This commit updates Ivy TestBed logic to calculate all potentially affected modules are reset cached scopes information for them (so that scopes are recalculated as needed).

Resolves #36619.

## PR Type

- [x] Bugfix

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/test/test_bed_spec.ts
+++ b/packages/core/test/test_bed_spec.ts
@@ -360,6 +360,10 @@ describe('TestBed', () => {
   });
 
   describe('nested module overrides using TestBed.overrideModule', () => {
+    // Set up an NgModule hierarchy with two modules, A and B, each with their own component.
+    // Module B additionally re-exports module A. Also declare two mock components which can be
+    // used in tests to verify that overrides within this hierarchy are working correctly.
+
     // ModuleA content:
 
     @Component({
--- a/packages/core/test/test_bed_spec.ts
+++ b/packages/core/test/test_bed_spec.ts
@@ -359,6 +359,122 @@ describe('TestBed', () => {
     });
   });
 
+  describe('nested module overrides using TestBed.overrideModule', () => {
+    // ModuleA content:
+
+    @Component({
+      selector: 'comp-a',
+      template: 'comp-a content',
+    })
+    class CompA {
+    }
+
+    @Component({
+      selector: 'comp-a',
+      template: 'comp-a mock content',
+    })
+    class MockCompA {
+    }
+
+    @NgModule({
+      declarations: [CompA],
+      exports: [CompA],
+    })
+    class ModuleA {
+    }
+
+    // ModuleB content:
+
+    @Component({
+      selector: 'comp-b',
+      template: 'comp-b content',
+    })
+    class CompB {
+    }
+
+    @Component({
+      selector: 'comp-b',
+      template: 'comp-b mock content',
+    })
+    class MockCompB {
+    }
+
+    @NgModule({
+      imports: [ModuleA],
+      declarations: [CompB],
+      exports: [CompB, ModuleA],
+    })
+    class ModuleB {
+    }
+
+    // AppModule content:
+
+    @Component({
+      selector: 'app',
+      template: `
+        <comp-a></comp-a>
+        <comp-b></comp-b>
+      `,
+    })
+    class App {
+    }
+
+    @NgModule({
+      imports: [ModuleB],
+      exports: [ModuleB],
+    })
+    class AppModule {
+    }
+
+    it('should detect nested module override', () => {
+      TestBed
+          .configureTestingModule({
+            declarations: [App],
+            // AppModule -> ModuleB -> ModuleA (to be overridden)
+            imports: [AppModule],
+          })
+          .overrideModule(ModuleA, {
+            remove: {declarations: [CompA], exports: [CompA]},
+            add: {declarations: [MockCompA], exports: [MockCompA]}
+          })
+          .compileComponents();
+
+      const fixture = TestBed.createComponent(App);
+      fixture.detectChanges();
+
+      // CompA is overridden, expect mock content.
+      expect(fixture.nativeElement.textContent).toContain('comp-a mock content');
+
+      // CompB is not overridden, expect original content.
+      expect(fixture.nativeElement.textContent).toContain('comp-b content');
+    });
+
+    it('should detect chained modules override', () => {
+      TestBed
+          .configureTestingModule({
+            declarations: [App],
+            // AppModule -> ModuleB (to be overridden) -> ModuleA (to be overridden)
+            imports: [AppModule],
+          })
+          .overrideModule(ModuleA, {
+            remove: {declarations: [CompA], exports: [CompA]},
+            add: {declarations: [MockCompA], exports: [MockCompA]}
+          })
+          .overrideModule(ModuleB, {
+            remove: {declarations: [CompB], exports: [CompB]},
+            add: {declarations: [MockCompB], exports: [MockCompB]}
+          })
+          .compileComponents();
+
+      const fixture = TestBed.createComponent(App);
+      fixture.detectChanges();
+
+      // Both CompA and CompB are overridden, expect mock content for both.
+      expect(fixture.nativeElement.textContent).toContain('comp-a mock content');
+      expect(fixture.nativeElement.textContent).toContain('comp-b mock content');
+    });
+  });
+
   describe('multi providers', () => {
     const multiToken = new InjectionToken<string[]>('multiToken');
     const singleToken = new InjectionToken<string>('singleToken');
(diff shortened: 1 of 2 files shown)


AndrewKushnir:
Also checked this fix with repros from #36619 and #35395, everything looks good.

AndrewKushnir:
[Initial g3 presubmit](http://test/OCL:306741523:BASE:306741539:1586992838696:cf39081a).

packages/core/test/test_bed_spec.ts
```diff
@@ -359,6 +359,122 @@ describe('TestBed', () => {
     });
   });
 
+  describe('nested module overrides using TestBed.overrideModule', () => {
```
alxhub, line 362: One thing I've started doing recently which might be useful here is a human-language description of testing scenarios such as this. Essentially, starting the test or a long setup section with a comment describing what's going on at a very high level.

Example:
```typescript
// Set up an NgModule hierarchy with two modules, A and B, each with their own component.
// Module B additionally re-exports module A. Also declare two mock components which can be
// used in tests to verify that overrides within this hierarchy are working correctly.
```

packages/core/testing/src/r3_test_bed_compiler.ts
```diff
@@ -532,6 +539,46 @@ export class R3TestBedCompiler {
...
   }
 
+  // When module overrides (via `TestBed.overrideModule`) are present, it might affect all modules
+  // that import (even transitively) an overridden one. For all affected modules we need to
+  // recalculate their scopes for a given test run and restore original scopes at the end. The goal
+  // of this function is to collect all affected modules in a set for further processing. Example:
+  // if we have the following module hierarchy: A -> B -> C (where `->` means `imports`) and module
+  // `C` is overridden, we consider `A` and `B` as affected, since their scopes might become
+  // invalidated with the override.
+  private collectModulesAffectedByOverride(arr: any[]): Set<NgModuleType<any>> {
```
alxhub, line 549:
```suggestion
  private collectModulesAffectedByOverrides(arr: any[]): Set<NgModuleType<any>> {
```

alxhub (approved):

packages/core/test/test_bed_spec.ts
AndrewKushnir: Thanks @alxhub, that's a great suggestion! I've added comments as you proposed.

AndrewKushnir:
Hi @ChristopherKiss, I've created this PR as a followup fix after #35395. It tuned out that the optimization was not taking into account full module hierarchy and as a result module scopes were not re-calculated in some cases (that was reported in #36619). This PR resolves the issue by going through the module tree and calculates all affected modules that require their scopes to be updated. I've checked this fix against the repro that you provided in the original ticket and performance looks good (the performance is on par with master branch).

Since you have a setup with a large number of components and modules, I was wondering if you have a chance to verify the performance of your project with the change in this PR? That'd be very helpful.

To test your app with the changes from this PR, you can follow an instruction in [this document on how to use PR artifacts](https://github.com/angular/angular/blob/master/docs/DEVELOPER.md#archives-for-each-package). An archive with updated `@angular/core` package that contains changes from the PR is available [here](https://686837-24195339-gh.circle-artifacts.com/0/angular/core-pr36649-a2577b6a2d.tgz) (please make sure you update other `@angular` packages to `9.1.2` to avoid version conflicts). **Note**: the mentioned archive with updated `@angular/core` package is for **testing purposes only** and not suitable for production usage (that archive will be removed once this PR is closed).

Thank you.

AndrewKushnir:
[Global Presubmit](https://test.corp.google.com/ui#id=OCL:306741523:BASE:307524971:1587436482723:7b8ea947)

ChristopherKiss:
@AndrewKushnir,

Thank you for reaching out on this PR.

image

Performance remains acceptable with these changes.

Noting `ng version` for reference.

```
                |___/
    

Angular CLI: 9.1.1
Node: 12.13.1
OS: win32 x64

Angular: 10.0.0-next.2
... core
Ivy Workspace: Yes

Package                             Version
-------------------------------------------------------------
@angular-devkit/architect           0.901.1
@angular-devkit/build-angular       0.901.1
@angular-devkit/build-optimizer     0.901.1
@angular-devkit/build-webpack       0.901.1
@angular-devkit/core                9.1.1
@angular-devkit/schematics          9.1.1
@angular/animations                 9.1.2
@angular/cdk                        9.2.1
@angular/cli                        9.1.1
@angular/common                     9.1.2
@angular/compiler                   9.1.2
@angular/compiler-cli               9.1.2
@angular/flex-layout                9.0.0-beta.29
@angular/forms                      9.1.2
@angular/language-service           9.1.2
@angular/material                   9.2.1
@angular/platform-browser           9.1.2
@angular/platform-browser-dynamic   9.1.2
@angular/platform-server            9.1.2
@angular/router                     9.1.2
@ngtools/webpack                    9.1.1
@schematics/angular                 9.1.1
@schematics/update                  0.901.1
rxjs                                6.5.4
typescript                          3.7.5
webpack                             4.39.2
```

AndrewKushnir:
Hi @ChristopherKiss, thanks for taking the time to help test these changes and sharing the result! 👍

packages/core/testing/src/r3_test_bed_compiler.ts
```diff
@@ -348,21 +350,26 @@ export class R3TestBedCompiler {
   }
 
   private applyTransitiveScopes(): void {
+    if (this.overriddenModules.size > 0) {
+      // Module overrides (via `TestBed.overrideModule`) might affect scopes that were previously
+      // calculated and stored in `transitiveCompileScopes`. If module overrides are present,
+      // collect all affected modules and reset scopes to force their re-calculatation.
+      const testingModuleDef = (this.testModuleType as any)[NG_MOD_DEF];
```
IgorMinar, line 357: can we slowly try to avoid using `as any` all over the place? it's a major code smell that just tells us that we are using the typesystem incorrectly. If `NG_MOD_DEF` is "private" api that we have to use the we should have an interface that is a superset of the `NgModuleType` called `NgModulePrivateType` that exposes the extra API surface to our internal code.

I'm not going to block this PR on this because it's a major refactoring, but I'd love for you to start thinking about writing the code in this way in the future and slowly refactor the codebase where you see `as any`.

Also please help to educate others on this. Thank you!

IgorMinar (approved):
LGTM with a suggestion on how to refactor this in the future.

packages/core/testing/src/r3_test_bed_compiler.ts
AndrewKushnir: That makes sense, thanks Igor!

PR #36651, opened 2020-04-16
build: list feat commits in patch branch in relase review script

atscott:
It would be useful for release review to have the `compare-master-to-patch.js` script also list any commits to the patch branch that were features so we can quickly check that we don't release features to stable.


IgorMinar (approved):

IgorMinar:
thanks @atscott!

AndrewKushnir (approved):
LGTM, thanks Andrew 👍

PR #36652, opened 2020-04-16
fix(ngcc): use annotateForClosureCompiler option

### LINKED ISSUE #36618: platformBrowser().bootstrapModule does not work under Closure Compiler ADVANCED_OPTIMIZATIONS
# 🐞 bug report

### Affected Package
@angular/platform-browser, @angular/compiler-cli, @angular/core

### Is this a regression?
Lack of parity between Ivy and ViewEngine.

### Description
* When compiling an application with Ivy (aot: true) with an entry point that contains `platformBrowser().bootstrapModule(AppModule)` and then running it through the Closure Compiler's advanced optimizations:
  * `BrowserModule`'s providers are removed as dead code which causes the application to fail to bootstrap with the error `"No ErrorHandler. Is platform module (BrowserModule) included?"`

* Compiling the same application with Closure's simple optimizations:
  * Application renders as expected.

* Changing the entry point to use `ɵrenderComponent(AppComponent)` and using advanced optimizations:
  * Simple application renders as expected.
  * Complex app fails with `'${type.name}' is neither 'ComponentType' or 'DirectiveType'.` for just about every Angular-owned dependency (e.g. RouterModule) 

## 🌍  Environment

**Angular Version:**
<pre><code>"@angular/animations": "9.1.1",
"@angular/bazel": "9.1.1",
"@angular/common": "9.1.1",
"@angular/compiler": "9.1.1",
"@angular/compiler-cli": "9.1.1",
"@angular/core": "9.1.1",
"@angular/forms": "9.1.1",
"@angular/platform-browser": "9.1.1",
"@angular/platform-browser-dynamic": "9.1.1",
"@angular/router": "9.1.1",</code></pre>

DavidANeil:
Adds @nocollapse to static properties added by ngcc iff annotateForClosureCompiler is true.

Resolves #36618.

## PR Checklist

## PR Type

- [X] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #36618 

## What is the new behavior?
When `ngcc` is invoked with a tsconfig that contains the option `"annotateForClosureCompiler": true` then `ngcc` will output closure compatible javascript.

- [X] No

### DIFF
--- a/packages/compiler-cli/ngcc/src/rendering/renderer.ts
+++ b/packages/compiler-cli/ngcc/src/rendering/renderer.ts
@@ -5,12 +5,13 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
-import {ConstantPool, Expression, Statement, WrappedNodeExpr, WritePropExpr} from '@angular/compiler';
+import {CommentStmt, ConstantPool, Expression, Statement, WrappedNodeExpr, WritePropExpr} from '@angular/compiler';
 import MagicString from 'magic-string';
 import * as ts from 'typescript';
 
 import {FileSystem} from '../../../src/ngtsc/file_system';
 import {ImportManager} from '../../../src/ngtsc/translator';
+import {ParsedConfiguration} from '../../../src/perform_compile';
 import {PrivateDeclarationsAnalyses} from '../analysis/private_declarations_analyzer';
 import {SwitchMarkerAnalyses, SwitchMarkerAnalysis} from '../analysis/switch_marker_analyzer';
 import {CompiledClass, CompiledFile, DecorationAnalyses} from '../analysis/types';
@@ -32,7 +33,8 @@ import {FileToWrite, getImportRewriter, stripExtension} from './utils';
 export class Renderer {
   constructor(
       private host: NgccReflectionHost, private srcFormatter: RenderingFormatter,
-      private fs: FileSystem, private logger: Logger, private bundle: EntryPointBundle) {}
+      private fs: FileSystem, private logger: Logger, private bundle: EntryPointBundle,
+      private tsConfig: ParsedConfiguration|null = null) {}
 
   renderProgram(
       decorationAnalyses: DecorationAnalyses, switchMarkerAnalyses: SwitchMarkerAnalyses,
@@ -82,8 +84,9 @@ export class Renderer {
       this.srcFormatter.removeDecorators(outputText, decoratorsToRemove);
 
       compiledFile.compiledClasses.forEach(clazz => {
-        const renderedDefinition =
-            this.renderDefinitions(compiledFile.sourceFile, clazz, importManager);
+        const renderedDefinition = this.renderDefinitions(
+            compiledFile.sourceFile, clazz, importManager,
+            !!this.tsConfig?.options.annotateForClosureCompiler);
         this.srcFormatter.addDefinitions(outputText, clazz, renderedDefinition);
 
         const renderedStatements =
@@ -160,12 +163,14 @@ export class Renderer {
    * @param imports An object that tracks the imports that are needed by the rendered definitions.
    */
   private renderDefinitions(
-      sourceFile: ts.SourceFile, compiledClass: CompiledClass, imports: ImportManager): string {
+      sourceFile: ts.SourceFile, compiledClass: CompiledClass, imports: ImportManager,
+      annotateForClosureCompiler: boolean): string {
     const name = this.host.getInternalNameOfClass(compiledClass.declaration);
-    const statements: Statement[] = compiledClass.compilation.map(c => {
-      return createAssignmentStatement(name, c.name, c.initializer);
+    const statements: Statement[][] = compiledClass.compilation.map(c => {
+      return createAssignmentStatements(
+          name, c.name, c.initializer, annotateForClosureCompiler ? '* @nocollapse ' : undefined);
     });
-    return this.renderStatements(sourceFile, statements, imports);
+    return this.renderStatements(sourceFile, Array.prototype.concat.apply([], statements), imports);
   }
 
   /**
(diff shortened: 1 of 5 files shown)


JoostK (changes requested):
Thanks for looking into this and opening a PR to resolve the issue! 🎉 

It looks to me like we don't have to make the changes to `TraitCompiler`, `CompileResult`, `WritePropExpr` and `ExpressionTranslatorVisitor`, instead keeping the necessary changes local to ngcc's renderer.

My idea is to only change `Renderer.renderDefinitions`/`Renderer.renderStatements` such that the `@nocollapse` comment is added there, if `annotateForClosureCompiler` is enabled. This is similar to ngtsc's approach of adding the comment in a TypeScript transformer, which would be the renderer in ngcc's case.

DavidANeil:
> My idea is to only change `Renderer.renderDefinitions`/`Renderer.renderStatements` such that the `@nocollapse` comment is added there, if `annotateForClosureCompiler` is enabled. This is similar to ngtsc's approach of adding the comment in a TypeScript transformer, which would be the renderer in ngcc's case.

Thanks for the quick feedback @JoostK!
How are you imagining that transform to happen?  The renderer just calls
```typescript
this.srcFormatter.printStatement(stmt, sourceFile, imports);
```
Which goes all the way from `Statement` (which you are wanting to not change) to `string`.  It could be possible to just prepend the `@nocollapse` to the string, but that seems hacky at best.
So should we just figure out which type of `stmt` need these added in the renderer?
To make it that `Statement` supports leading comments would still require these changes in `ExpressionTranslatorVisitor`, and I'm afraid that by the time we get to printing it is much more ambiguous which statements even need the `@nocollapse` annotation.  At the very best we would have to do something like
```typescript
if (this.annotateForClosureCompiler && stmt.expr instanceof WritePropExpr && someEsotericCheckThatThisIsAStaticPropertyAndNotAnyOtherKind(stmt)) {
    stmt.expr.leadingComment = '* @nocollapse ';
}
this.srcFormatter.printStatement(stmt, sourceFile, imports);
```
Which still requires the changes in `WritePropExpr` and `visitWritePropExpr`.

----

While writing the above I tried implementing it and was unsuccessful.  I'm happy to be convinced that there is a better way, I'm just not sure I'm knowledgeable enough already to be able to implement that way.
I'm happy to collaborate on this PR with commits authored by others.

JoostK:
Thanks for the reply. I agree that tacking on the comment to a string would be a hack, but I see some other ways:

1. Emit a `o.CommentStmt` in front of a definition statement in `createAssignmentStatement` (would need to start returning an array). This would require updating `ExpressionTranslatorVisitor` as support for that statement kind is not currently implemented.

2. Introduce a new type in ngcc to represent a statement, which can have comments associated with it. `RendererFormatter.printStatement` can then take this new type and emit synthetic comments as desired. Something like this:
```ts
export type EmitStatement {
  stmt: o.Statement;
  comments: string[];
}
```
<sub>I'm not sold on the name but I can't think of something better</sub>

3. Change `RendererFormatter.printStatement` such that it accepts the `leadingComment` you added to `WritePropExpr`. The implementations can then add the comment synthetically to the `ts.Node` before printing it as a string.
<sub>This is still a bit of a hack.</sub>

My personal preference would be in the order as presented above. If you run into problems feel free to follow up here.

DavidANeil:
I should be able to get to this toward the end of next week.

petebacondarwin:
@DavidANeil - do you still have time to work on this?

DavidANeil:
Yup sorry for the radio silence, my plan was to come back to it as soon as I enabled Ivy for our code base, and I thought that "the end of next week" would be a good estimate for when that would be done, but it turns out "the end of next month" was probably a more accurate estimate.  I think I am about 3/4 done with enabling Ivy, so I should be circling back to this in a week or so?

Sorry I can't get to it sooner.

petebacondarwin:
No problem. Good luck with your ivy-ifying of your code.

IgorMinar:
Let's discuss the overall "Angular + Closure Compiler outside of google3" story before we start adding more flags and closure support into ngc and especially ngcc.

I have a meeting with @DavidANeil scheduled for next week, so I'll bring up this PR there and will post an update on this PR afterwards.

DavidANeil:
Sounds good, that's what I was thinking too.

I just wanted to get this PR into a better state before I entirely forgot my mental model of the ngcc renderer.

IgorMinar (approved):
After discussing the build setup with @DavidANeil I think that this is a reasonable change to make. 

@DavidANeil can you please update this PR with:
- updat the commit message to explain "The Why" - the motivation behind this change
- rebase the PR (it should make the saucelabs flake go away)

LGTM otherwise. Thanks!

The rationale behind this decision is concistency. `ngc` already supports the flag, and `ngcc` exists only to convert code from VE to Ivy, so there is no good reason for `ngcc` not to support this flag - it doesn't increase our support burden or feature set. In fact one could argue that this change fixes a regression caused by the Ivy and ngcc rollout in v9.

packages/compiler-cli/ngcc/src/rendering/renderer.ts
```diff
@@ -9,6 +9,8 @@ import {ConstantPool, Expression, Statement, WrappedNodeExpr, WritePropExpr} fro
 import MagicString from 'magic-string';
 import * as ts from 'typescript';
 
+import {ParsedConfiguration} from '../../../../compiler-cli';
+import {CommentStmt} from '../../../../compiler/src/output/output_ast';
```
JoostK, line 13: Can you move this import to `@angular/compiler` above?

JoostK (approved):
I had completely missed the refactored commit, thanks for the cleanup!

Also, the linter is complaining about the commit message being too short (needs to be 100 chars at least). Could you also change `Resolves #36618` into `Closes #36618`, that's the term we typically use.

One small nit about an import, otherwise LGTM!

PR #36661, opened 2020-04-16
ci: add codeowner group for migrations

devversion:
**Please see individual commits**. 

The PR quickly evolved into something larger. When I tested the new codeowner group for migrations (which we decided to create based on a recent FW meeting), I noticed that:

1. Reviews for `fw-compiler` are sometimes not sent if ngcc files have been touched.
2. We need to refactor our pullapprove verification script so that we can fix (1) and that we can use simpler conditions.


.pullapprove.yml
```diff
@@ -229,6 +222,22 @@ groups:
...
+    conditions:
+      - *can-be-global-approved
+      - *can-be-global-docs-approved
+      - files.include("packages/core/schematics/**")
+    reviewers:
+      users:
+        - alxhub
+        - crisbeto
+        - devversion
+        - kara
```
devversion, line 238: @alxhub @crisbeto @kara Can you please let me know if you are fine with being in this group?

alxhub (approved):

crisbeto (approved):

filipesilva (approved):

dev-infra/pullapprove/condition_evaluator.ts
```diff
@@ -0,0 +1,93 @@
...
+ * Converts a given condition to a function that accepts a set of files. The returned
+ * function can be called to check if the set of files matches the condition.
+ */
+export function convertConditionToFunction(expr: string): (files: string[]) => boolean {
+  // Creates a dynamic function with the specified expression. The first parameter will
+  // be `files` as that corresponds to the supported `files` variable that can be accessed
+  // in PullApprove condition expressions. The followed parameters correspond to other
+  // context variables provided by PullApprove for conditions.
+  const evaluateFn = new Function('files', ...Object.keys(conditionContext), `
+    return !!(${transformExpressionToJs(expr)});
```
josephperrott, line 38: It looks like we run into an issue here with the difference between type coercion in Python vs Javascript.

A `List` in python with zero elements is falsy, but in javascript an `Array` with zero elements is truthy.

```py
bool([])
# evalates true
```

```js
!![];
// evaluates false
```


So in cases where the condition is just `files.include` or `files.exclude` it returns the resulting list/array to be coerced to a boolean.  PullApprove will treat empty lists as `false`, but our test treats them as `true`.

This means that our test treats *all* lists as `true`.

We might need to do something like this in our dynamic function body

```suggestion
    return !!(${transformExpressionToJs(expr)});
    const conditionEvalResult = (${transformExpressionToJs(expr)});
    if (Array.isArray(conditionEvalResult)) {
      return !!conditionEvalResult.length;
    }
    return conditionEvalResult;
```

josephperrott (changes requested):

dev-infra/pullapprove/condition_evaluator.ts
devversion: Good catch @josephperrott. Fixed!

dev-infra/pullapprove/group.ts
```diff
@@ -5,162 +5,100 @@
...
+import {convertConditionToFunction} from './condition_evaluator';
 import {PullApproveGroupConfig} from './parse-yaml';
 
 /** A condition for a group. */
 interface GroupCondition {
-  glob: string;
-  matcher: IMinimatch;
-  matchedFiles: Set<string>;
+  expression: string;
+  checkFn: (files: string[]) => boolean, matchedFiles: Set<string>;
```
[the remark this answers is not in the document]
devversion: Fixed. Clang-format did this due to the comma.

josephperrott (approved):
LGTM, thanks!

I think this method of evaluating the config will be a better option long term.

IgorMinar (approved):
LGTM

Reviewed-for: `.pullapprove.yml`

devversion:
Caretaker: This PR has been approved by Igor, but it looks like pull approve did not capture the review due to the actual group being called `code-ownership`. The PR is ready for merge.

IgorMinar (approved):
ooops... reapproving: LGTM

Reviewed-for: code-ownership

matsko:
@devversion please rebase

devversion:
@matsko Done. Thx for the heads up!

PR #36316, opened 2020-03-30
build(docs-infra): fix `autoLinkCode` to ignore docs without a path

sonukapoor:
Closes #36260

## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [X] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [X] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] angular.io application / infrastructure changes
- [ ] Other... Please describe:

## What is the current behavior?

Issue Number: #36260

## Does this PR introduce a breaking change?

- [ ] Yes
- [X] No

PR #36686, opened 2020-04-17
refactor(docs-infra): refactor `autoLinkCode`

sonukapoor:
This PR is based on https://github.com/angular/angular/pull/36316. It should be merged after that one is merged. 

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [X] Code style update (formatting, local variables)
- [ ] Build related changes
- [ ] CI related changes

- [ ] Yes
- [X] No

### DIFF
--- a/aio/tools/transforms/angular-base-package/post-processors/auto-link-code.js
+++ b/aio/tools/transforms/angular-base-package/post-processors/auto-link-code.js
@@ -35,8 +35,8 @@ module.exports = function autoLinkCode(getDocFromAlias) {
         }
 
         visit(node, 'text', (node, ancestors) => {
-          const isInsideOtherLinks = isInsideLink(ancestors);
-          if (isInsideOtherLinks) {
+          const isInLink = isInsideLink(ancestors);
+          if (isInLink) {
             return;
           }
 
--- a/aio/tools/transforms/angular-base-package/post-processors/auto-link-code.js
+++ b/aio/tools/transforms/angular-base-package/post-processors/auto-link-code.js
@@ -30,49 +30,62 @@ module.exports = function autoLinkCode(getDocFromAlias) {
   function autoLinkCodeImpl() {
     return (ast, file) => {
       visit(ast, 'element', (node, ancestors) => {
-        // Only interested in code elements that:
-        // * do not have `no-auto-link` class
-        // * do not have an ignored language
-        // * are not inside links
-        if (autoLinkCodeImpl.codeElements.some(elementType => is(node, elementType)) &&
-            (!node.properties.className || !node.properties.className.includes('no-auto-link')) &&
-            !autoLinkCodeImpl.ignoredLanguages.includes(node.properties.language) &&
-            ancestors.every(ancestor => !is(ancestor, 'a'))) {
-          visit(node, 'text', (node, ancestors) => {
-            // Only interested in text nodes that are not inside links
-            if (ancestors.every(ancestor => !is(ancestor, 'a'))) {
-              const parent = ancestors[ancestors.length - 1];
-              const index = parent.children.indexOf(node);
-
-              // Can we convert the whole text node into a doc link?
-              const docs = getDocFromAlias(node.value);
-              if (foundValidDoc(docs, node.value, file)) {
-                parent.children.splice(index, 1, createLinkNode(docs[0], node.value));
-              } else {
-                // Parse the text for words that we can convert to links
-                const nodes =
-                    textContent(node)
-                        .split(/([A-Za-z0-9_.-]+)/)
-                        .filter(word => word.length)
-                        .map((word, index, words) => {
-                          // remove docs that fail the custom filter tests
-                          const filteredDocs = autoLinkCodeImpl.customFilters.reduce(
-                              (docs, filter) => filter(docs, words, index), getDocFromAlias(word));
-                          return foundValidDoc(filteredDocs, word, file) ?
-                              // Create a link wrapping the text node.
-                              createLinkNode(filteredDocs[0], word) :
-                              // this is just text so push a new text node
-                              {type: 'text', value: word};
-                        });
-
-                // Replace the text node with the links and leftover text nodes
-                Array.prototype.splice.apply(parent.children, [index, 1].concat(nodes));
-              }
-            }
-          });
+        if (!isValidCodeElement(node, ancestors)) {
+          return;
         }
+
+        visit(node, 'text', (node, ancestors) => {
+          // Only interested in text nodes that are not inside links
+          const isInsideLink = ancestors.every(ancestor => !is(ancestor, 'a'));
+          if (!isInsideLink) {
+            return;
+          }
+
+          const parent = ancestors[ancestors.length - 1];
+          const index = parent.children.indexOf(node);
+
+          // Can we convert the whole text node into a doc link?
+          const docs = getDocFromAlias(node.value);
+          if (foundValidDoc(docs, node.value, file)) {
+            parent.children.splice(index, 1, createLinkNode(docs[0], node.value));
+          } else {
+            // Parse the text for words that we can convert to links
+            const nodes = getNodes(node, file);
+            // Replace the text node with the links and leftover text nodes
+            Array.prototype.splice.apply(parent.children, [index, 1].concat(nodes));
+          }
+        });
       });
     };
+  }
+
+  function isValidCodeElement(node, ancestors) {
+    // Only interested in code elements that:
+    // * do not have `no-auto-link` class
+    // * do not have an ignored language
+    // * are not inside links
+    const containsElement = autoLinkCodeImpl.codeElements.some(elementType => is(node, elementType));
+    const hasNoAutoLink = !node.properties.className || !node.properties.className.includes('no-auto-link');
+    const isLanguageSupported = !autoLinkCodeImpl.ignoredLanguages.includes(node.properties.language);
+    const isInsideLink = ancestors.every(ancestor => !is(ancestor, 'a'));
+    return containsElement && hasNoAutoLink && isLanguageSupported && isInsideLink;
+  }
+
+  function getNodes(node, file) {
+    return textContent(node)
+      .split(/([A-Za-z0-9_.-]+)/)
+      .filter(word => word.length)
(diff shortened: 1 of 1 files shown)


sonukapoor:
@gkalpak / @petebacondarwin Please take a look at this PR. I have refactored `autoLinkCode`. 🙏

aio/tools/transforms/angular-base-package/post-processors/auto-link-code.js
```diff
@@ -30,51 +30,64 @@ module.exports = function autoLinkCode(getDocFromAlias) {
...
+  function isValidCodeElement(node, ancestors) {
+    // Only interested in code elements that:
+    // * do not have `no-auto-link` class
+    // * do not have an ignored language
+    // * are not inside links
+    const containsElement = autoLinkCodeImpl.codeElements.some(elementType => is(node, elementType));
+    const hasNoAutoLink = !node.properties.className || !node.properties.className.includes('no-auto-link');
+    const isLanguageSupported = !autoLinkCodeImpl.ignoredLanguages.includes(node.properties.language);
+    const isInsideLink = ancestors.every(ancestor => !is(ancestor, 'a'));
+    return containsElement && hasNoAutoLink && isLanguageSupported && isInsideLink;
```
gkalpak, line 71:
```suggestion
    const isCodeElement = autoLinkCodeImpl.codeElements.some(elementType => is(node, elementType));
    const hasNoAutoLink = node.properties.className && node.properties.className.includes('no-auto-link');
    const isLanguageSupported = !autoLinkCodeImpl.ignoredLanguages.includes(node.properties.language);
    const isInsideLink = ancestors.some(ancestor => is(ancestor, 'a'));
    return isCodeElement && !hasNoAutoLink && isLanguageSupported && !isInsideLink;
```

BTW, one problem (which is not terribly important here, I guess) with splitting this up into intermediate variables is that we lose the shortcircuiting capabilities of `&&`/`||`. I.e. now we will always traverse all ancestors even if this is not a code element.

Can you check whether this makes any difference in the time it takes to generate the docs?

aio/tools/transforms/angular-base-package/post-processors/auto-link-code.js
```diff
@@ -30,51 +30,64 @@ module.exports = function autoLinkCode(getDocFromAlias) {
...
-                          return foundValidDoc(filteredDocs, word, file) ?
-                              // Create a link wrapping the text node.
-                              createLinkNode(filteredDocs[0], word) :
-                              // this is just text so push a new text node
-                              {type: 'text', value: word};
-                        });
+        visit(node, 'text', (node, ancestors) => {
+          // Only interested in text nodes that are not inside links
+          const isInsideLink = ancestors.every(ancestor => !is(ancestor, 'a'));
+          if (!isInsideLink) {
```
gkalpak, line 40:
```suggestion
          const isInsideLink = ancestors.some(ancestor => is(ancestor, 'a'));
          if (isInsideLink) {
```

BTW, since this same logic is used in `isValidCodeElement()`, you could extract it to a standalone function.

aio/tools/transforms/angular-base-package/post-processors/auto-link-code.js
```diff
@@ -30,51 +30,64 @@ module.exports = function autoLinkCode(getDocFromAlias) {
...
+    const hasNoAutoLink = !node.properties.className || !node.properties.className.includes('no-auto-link');
+    const isLanguageSupported = !autoLinkCodeImpl.ignoredLanguages.includes(node.properties.language);
+    const isInsideLink = ancestors.every(ancestor => !is(ancestor, 'a'));
+    return containsElement && hasNoAutoLink && isLanguageSupported && isInsideLink;
+  }
+
+  function getNodes(node, file) {
+    return textContent(node)
+      .split(/([A-Za-z0-9_.-]+)/)
+      .filter(word => word.length)
```
gkalpak, line 77: While you are at it, let's do one more improvement:
The current implementation will also consider whitespace-only tokens. For example, `"foo bar"` will result into 3 words - "foo", " " and "bar". Looking for docs to match whitespace-only nodes is unnecessary. Let's fix it with:

```suggestion
      .filter(word => word.trim().length > 0)
```

gkalpak (changes requested):
Nice 😍 Thx, @sonukapoor!

BTW, the commit message type should be `refactor` (not `style`).

sonukapoor:
Wow...Love your suggestions. Will make the changes 👍

aio/tools/transforms/angular-base-package/post-processors/auto-link-code.js
sonukapoor: When I ran the script I did not notice any significant delays. I could try to revert the code and use `console.time` to see how long it takes to generate the docs. Would that be sufficient?

sonukapoor: We could always exit early? 

```typescript
const isCodeElement = ...
if (!isCodeElement) {
  return false;
}
const hasNoAutoLink = ...
```

gkalpak: Let's see if this introduces any noticeable delay first.
Any way of timing it witl and without the change is fine.

aio/tools/transforms/angular-base-package/post-processors/auto-link-code.js
sonukapoor: Is the proposed reverted logic line 40 correct? Exit out if we found a link? I thought the logic should run only for links. 😕

aio/tools/transforms/angular-base-package/post-processors/auto-link-code.js
sonukapoor: There is no significant delay. Before the change, it takes `63.63s` to run the whole process. After the change, it takes `51.22s.` I have run it a few times and don't see much difference. 

I am printing the times that the entire process takes. It does that automatically.

aio/tools/transforms/angular-base-package/post-processors/auto-link-code.js
sonukapoor: I tried that, but it seems we need the spaces. Without that, the code is converted to:
```typescript
abstractclassChangeDetectorRef {
  abstract markForCheck(): voidabstract detach(): voidabstract detectChanges(): voidabstract checkNoChanges(): voidabstract reattach(): void
}
```
vs

```typescript
abstract class ChangeDetectorRef {
  abstract markForCheck(): void
  abstract detach(): void
  abstract detectChanges(): void
  abstract checkNoChanges(): void
  abstract reattach(): void
}
```

aio/tools/transforms/angular-base-package/post-processors/auto-link-code.js
gkalpak: OK, if it doesn't make any difference, let's keep it as is.
It is strange though that it takes less time with the change. Where does the speed-up come from?

aio/tools/transforms/angular-base-package/post-processors/auto-link-code.js
gkalpak: This post-processor generates links from code inside code-blocks. We do not want to generate any links inside other links.

So, yes, the proposed change in my previous comment is correct.

aio/tools/transforms/angular-base-package/post-processors/auto-link-code.js
gkalpak: Did you try my proposed change exaclty? It is not possible for my changes to have changed the resulting HTML (since strings are immutable and we only use `.trim()` in `filter` (which is not mutating the array)).

sonukapoor: I did. It's not working for me as expected.

aio/tools/transforms/angular-base-package/post-processors/auto-link-code.js
sonukapoor: Sounds good

aio/tools/transforms/angular-base-package/post-processors/auto-link-code.js
sonukapoor: Not entirely sure TBH. I see that the times are very random. Sometimes it takes `40`s, and sometimes `60`s. Either way, I do not see much difference, after running it a few times.

gkalpak: :+1:

aio/tools/transforms/angular-base-package/post-processors/auto-link-code.js
```diff
@@ -35,9 +35,8 @@ module.exports = function autoLinkCode(getDocFromAlias) {
         }
 
         visit(node, 'text', (node, ancestors) => {
-          // Only interested in text nodes that are not inside links
-          const isInsideLink = ancestors.every(ancestor => !is(ancestor, 'a'));
-          if (!isInsideLink) {
+          const isInsideOtherLinks = isInsideLink(ancestors);
```
gkalpak, line 38:
```suggestion
          const isInsideOtherLink = isInsideLink(ancestors);
```

...or (to be consistent with the other call below):

```suggestion
          const isInLink = isInsideLink(ancestors);
```

aio/tools/transforms/angular-base-package/post-processors/auto-link-code.js
gkalpak: Oh, I see - I didn't realize we are using these nodes to replace the original code element. We need the whitespace then, yes.

Let's leave it as is then. Thx for trying it out and sorry for the noise 😅

gkalpak (approved):
LGTM ✨
Thx for following through, @sonukapoor!

aio/tools/transforms/angular-base-package/post-processors/auto-link-code.js
sonukapoor: Done: https://github.com/angular/angular/pull/36686/commits/47df12e967998944ee4fb8f2515dd3d3c2c5caa3

PR #36687, opened 2020-04-17
fix(ngcc): do not use cached file-system

petebacondarwin:
The cached file-system was implemented to speed up ngcc
processing, but in reality most files are not accessed many times
and there is no noticeable degradation in speed by removing it.

Benchmarking `ngcc -l debug` for AIO on a local machine
gave a range of 196-236 seconds with the cache and 197-224
seconds without the cache.

Moreover, when running in parallel mode, ngcc has a separate
file cache for each process. This results in excess memory usage.
Notably the master process, which only does analysis of entry-points
holds on to up to 500Mb for AIO when using the cache compared to
only around 30Mb when not using the cache.

Finally, the file-system cache being incorrectly primed with file
contents before being processed has been the cause of a number
of bugs. For example https://github.com/angular/angular-cli/issues/16860#issuecomment-614694269.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No


packages/compiler-cli/ngcc/src/locking/lock_file_with_child_process/index.ts
```diff
@@ -58,11 +57,6 @@ export class LockFileWithChildProcess implements LockFile {
 
   read(): string {
     try {
-      if (this.fs instanceof CachedFileSystem) {
```
JoostK, line 61: I would be in favour of keeping this check as long as `CachedFileSystem` is still around. Note that `CachedFileSystem` is no longer used with this change, though, so we may choose to remove it entirely.

JoostK (commented):
There's a remaining comment in `integration/ngcc/test.sh` that refers to `CachedFileSystem`

packages/compiler-cli/ngcc/src/locking/lock_file_with_child_process/index.ts
petebacondarwin: I don't like to leave code around that is not needed, which is why I removed this. I would be willing to remove the `CachedFileSystem` altogether but balked because it is outside ngcc...

JoostK: I see. I realise now that this has always been kind of a smell, as all implementors of `FileSystem` were free to cache as they wanted such that this call would always have been relevant (i.e. be part of the `FileSytem` interface itself)

JoostK (approved):

integration/ngcc/test.sh
```diff
@@ -173,7 +173,6 @@ assertSucceeded "Expected 'ngcc' to log 'Compiling'."
   assertEquals 1 `cat node_modules/@angular/material/button/button.d.ts | grep 'import \* as ɵngcc0' | wc -l`
 
   # Re-compile packages (which requires cleaning up those compiled by a different ngcc version).
-  # (Use sync mode to ensure all tasks share the same `CachedFileSystem` instance.)
```
gkalpak, line 176: Why not also remove `--no-async` then? :grin:

gkalpak (approved):
Awesome! Very excited about the memory savings 🎉

integration/ngcc/test.sh
petebacondarwin: Fair enough

kara (approved):
LGTM for integration/

matsko:
presubmit = http://test/OCL:307078160:BASE:307078193:1587147233826:f275ee04

Just issuing this to see if the g3 scripts pick anything up.

PR #36722, opened 2020-04-20
test(core): Add benchmark for transplanted views when insertion is dirty

atscott:
The current benchmark for transplanted views only exercises the path
when the declaration location is dirty and the insertion is not. This
test adds a benchmark for when both insertion and declaration are dirty.


mhevery (approved):

pkozlowski-opensource (approved):
The benchmark changes LGTM! I'm still looking at the actual results and their analysis but it shouldn't block merging of this new perf scenario.

atscott:
merge assistance: not sure what the pending code review is. Existing approvals seem sufficient.

PR #36723, opened 2020-04-20
fix(dev-infra): exit non-zero if commit message validation failed

devversion:
Currently the `commit-message` validation script does not exit
with a non-zero exit code if the commit message validation failed.

This means that invalid commit messages are currently not
causing CI to be red. See: https://circleci.com/gh/angular/angular/686008

@josephperrott I think we _could_ update the required base sha in the merge script one this PR lands. Not sure if we want to do that though (as it introduces slow-down due to rebasing I guess)


matsko (approved):

josephperrott (approved):
LGTM

PR #36726, opened 2020-04-20
Migrate from gulp format to ng-dev format

josephperrott:

___
Beginning with #36726 `gulp` is no longer used for running `clang-format` across the repo.  Instead this task is now performed using the `ng-dev` command.  The formatting enforced and performed is identical and does not result in any change to the source code of the repository.

This is being done to prevent OOM issues when using gulp to perform formatting of large changes.  Additionally, this introduces some measure of parallelization to the task speeding up the performance of the task.

As we no longer rely on `gulp` the `gulp` commands have been removed, if they are called a deprecation message will be returned.

Example usage changes:
`yarn gulp format:all:enforce` -> `yarn ng-dev format all --check`
`yarn gulp format` -> `yarn ng-dev format changed`


dev-infra/format/run-commands-parallel.ts
```diff
@@ -0,0 +1,75 @@
...
+            // If more files exist in the list, run again to work on the next file,
+            // using the same slot.
+            if (files.length) {
+              return runCommandInSlot(slot);
+            }
+            // If not more files are available, mark the slot as unused.
+            slots[slot] = false;
+            // If all of the slots are false, as they are unused, mark the progress bar
+            // completed and resolve the promise.
+            if (!slots.reduce((a, b) => a || b)) {
```
devversion, line 60:
```suggestion
            if (slots.every(active => !active)) {
```

dev-infra/format/format.ts
```diff
@@ -0,0 +1,123 @@
...
+import {getAngularDevConfig, getRepoBaseDir} from '../utils/config';
+
+import {FormatConfig} from './config';
+import {runInParallel} from './run-commands-parallel';
+
+/** By default, run the formatter on all javascript and typescript files. */
+const DEFAULT_MATCHERS = ['**/*.{t,j}s', '*.{t,j}s'];
+
+/** The number of threads to use for parallelization of formatting. */
+const THREAD_LIMIT = 8;
```
devversion, line 22: Should we limit this on the number of available core's? Otherwise the tool might have bad performance

dev-infra/format/format.ts
```diff
@@ -0,0 +1,123 @@
...
+  // All files which formatting should be applied to.
+  const files = filterFilesByMatchers(unfilteredFiles);
+
+  console.info(`Formatting ${files.length} file(s)`);
+
+
+  // Run the formatter to format the files in place, split across 8 processes.
+  await runInParallel(
+      files, THREAD_LIMIT, `${getFormatterBinary()} -i -style=file`, (file, code, _, stderr) => {
+        if (code) {
```
devversion, line 39: Nit: Can we be more explicit what exit code we expect here? Using `code !== 0` would already help IMO (same for `checkFiles`)

devversion (approved):
Nice work! Hopefully we can remove gulp soon. I added a few minor comments below.

josephperrott:
Caretaker Note:  This PR is ready for merge, but needs to be merged alongside an announcement to about the change.

___
Beginning with #36726 `gulp` is no longer used for running `clang-format` across the repo.  Instead this task is now performed using the `ng-dev` command.  The formatting enforced and performed is identical and does not result in any change to the source code of the repository.

This is being done to prevent OOM issues when using gulp to perform formatting of large changes.  Additionally, this introduces some measure of parallelization to the task speeding up the performance of the task.

As we no longer rely on `gulp` the `gulp` commands have been removed, if they are called a deprecation message will be returned.

Example usage changes:
`yarn gulp format:all:enforce` -> `yarn ng-dev format all --check`
`yarn gulp format` -> `yarn ng-dev format changed`

dev-infra/format/format.ts
```diff
@@ -0,0 +1,134 @@
...
+  console.info(`Formatting ${files.length} file(s)`);
+
+
+  // Run the formatter to format the files in place, split across up to 8 processes.
+  // The task is done in multiple processess to speed up the overall time of the task,
+  // as running accross entire repositories takes a large amount of time.
+  // As a data point for illustration, using 8 process rather than 1 cut the execution
+  // time from 276 seconds to 39 seconds for the same 2700 files
+  await runInParallel(
+      files, THREAD_LIMIT, `${getFormatterBinary()} -i -style=file`, (file, code, _, stderr) => {
```
devversion, line 42: Isn't this running sequentially now? The `THREAD_LIMIT` has been updated to `1`. Also the comment above seems to be out-of-date now.

dev-infra/tmpl-package.json
```diff
@@ -12,6 +12,9 @@
     "chalk": "<from-root>",
     "glob": "<from-root>",
     "minimatch": "<from-root>",
+    "cli-progress": "<from-root>",
+    "multimatch": "<from-root>",
+    "prompts": "<from-root>",
```
devversion, line 17: I think replacing `prompts` with `inquirer` here fixes CI.

devversion (approved):
💯 Two final comments. Other than that, LGTM.

dev-infra/format/format.ts
josephperrott: I had put it to one to retest and get the numbers for my comment, forgot to put it bad.

devversion: Do you think we should always just run at maximum available cores? Even if it's more than 8?

josephperrott: I suppose we could, for some reason I felt like we shouldn't.  But I guess there isn't really a reason not to do that.

.dev-infra.json
```diff
@@ -43,5 +43,28 @@
...
+      "!packages/common/locales/**/*.{js,ts}",
+      "!packages/common/src/i18n/available_locales.ts",
+      "!packages/common/src/i18n/currencies.ts",
+      "!packages/common/src/i18n/locale_en.ts",
+      "modules/benchmarks/**/*.{js,ts}",
+      "modules/e2e_util/**/*.{js,ts}",
+      "modules/playground/**/*.{js,ts}",
+      "tools/**/*.{js,ts}",
+      "tools/gulp-tasks/cldr/extract.js",
+      "tools/public_api_guard/**/*.d.ts",
```
gkalpak, line 61: Did you mean:

```suggestion
      "!tools/gulp-tasks/cldr/extract.js",
      "!tools/public_api_guard/**/*.d.ts",
```

Otherwise, these two shouldn't be needed (they are covered by `tools/**/*.{js,ts}`).

dev-infra/format/run-commands-parallel.ts
```diff
@@ -0,0 +1,77 @@
...
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {Bar} from 'cli-progress';
+import {cpus} from 'os';
+import {exec} from 'shelljs';
+
+const AVAILABLE_THREADS = Math.min(cpus().length - 1, 1);
```
gkalpak, line 13: Did you mean:

```suggestion
const AVAILABLE_THREADS = Math.max(cpus().length - 1, 1);
```

:grin:

dev-infra/format/run-commands-parallel.ts
```diff
@@ -0,0 +1,77 @@
...
+
+type CallbackFunction = (file: string, code?: number, stdout?: string, stderr?: string) => void;
+
+/**
+ * Run the provided commands in parallel for each provided file.
+ *
+ * A promise is returned, completed when the command has completed runing for each file.
```
gkalpak, line 20:
```suggestion
 * A promise is returned, completed when the command has completed running for each file.
```

dev-infra/format/format.ts
```diff
@@ -0,0 +1,130 @@
...
+import * as multimatch from 'multimatch';
+import {join} from 'path';
+
+import {getAngularDevConfig, getRepoBaseDir} from '../utils/config';
+
+import {FormatConfig} from './config';
+import {runInParallel} from './run-commands-parallel';
+
+/** By default, run the formatter on all javascript and typescript files. */
+const DEFAULT_MATCHERS = ['**/*.{t,j}s', '*.{t,j}s'];
```
gkalpak, line 19: Doesn't `**/*{t,j}s` include `*.{t,j}s`?

dev-infra/format/format.ts
```diff
@@ -0,0 +1,130 @@
...
+
+  console.info(`Formatting ${files.length} file(s)`);
+
+
+  // Run the formatter to format the files in place, split across (number of available
+  // cpu threads - 1) processess. The task is done in multiple processess to speed up
+  // the overall time of the task, as running accross entire repositories takes a large
+  // amount of time.
+  // As a data point for illustration, using 8 process rather than 1 cut the execution
+  // time from 276 seconds to 39 seconds for the same 2700 files
```
gkalpak, line 38: Have you checked whether this has a negative impact for less than 20 files (which is what we usually have on a PR)?

dev-infra/format/format.ts
```diff
@@ -0,0 +1,130 @@
...
+  await runInParallel(files, `${getFormatterBinary()} -i -style=file`, (file, code, _, stderr) => {
+    if (code !== 0) {
+      formatFailed = true;
+      console.error(`Error running clang-format on: ${file}`);
+      console.error(stderr);
+      console.error();
+      process.exit(1);
```
gkalpak, line 45: Is this intentional? Shouldn't we only exit if `formatFailed` is false at the end?

dev-infra/format/format.ts
```diff
@@ -0,0 +1,130 @@
...
+ */
+export async function checkFiles(unfilteredFiles: string[]) {
+  // All files which formatting should be applied to.
+  const files = filterFilesByMatchers(unfilteredFiles);
+  // Files which are currently not formatted correctly.
+  const failures: string[] = [];
+
+  console.info(`Checking format of ${files.length} file(s)`);
+
+  // Run the formatter to format the files in place, split across up to 8 processes.
```
gkalpak, line 69: This doesn't seem to be the case any more.

dev-infra/format/format.ts
```diff
@@ -0,0 +1,130 @@
...
+  // The task is done in multiple processess to speed up the overall time of the task,
+  // as running accross entire repositories takes a large amount of time.
```
gkalpak, line 71:
```suggestion
  // as running across entire repositories takes a large amount of time.
```

package.json
```diff
@@ -30,7 +30,7 @@
     "//circleci-win-comment": "See the test-win circleci job for why these are needed. If they are not needed anymore, remove them.",
     "circleci-win-ve": "bazelisk test --build_tag_filters=-ivy-only --test_tag_filters=-ivy-only,-browser:chromium-local //packages/compiler-cli/... //tools/ts-api-guardian/...",
     "circleci-win-ivy": "bazelisk test --config=ivy --build_tag_filters=-no-ivy-aot,-fixme-ivy-aot --test_tag_filters=-no-ivy-aot,-fixme-ivy-aot,-browser:chromium-local //packages/compiler-cli/... //tools/ts-api-guardian/...",
-    "lint": "yarn -s tslint && yarn gulp lint",
+    "lint": "yarn -s tslint && yarn -s ng-dev format change --check",
```
gkalpak, line 33:
```suggestion
    "lint": "yarn -s tslint && yarn -s ng-dev format changed --check",
```

???

dev-infra/format/cli.ts
```diff
@@ -0,0 +1,45 @@
...
+      .command(
+          'all', 'Run the formatter on all files in the repository', {},
+          ({check}) => {
+            const executionCmd = check ? checkFiles : formatFiles;
+            executionCmd(allFiles());
+          })
+      .command(
+          'changed [shaOrRef]', 'Run the formatter on files changed since the provide sha/ref', {},
+          ({shaOrRef, check}) => {
+            let sha = shaOrRef || `master`;
```
gkalpak, line 33: Nit:

```suggestion
            const sha = shaOrRef || 'master';
```

dev-infra/format/cli.ts
```diff
@@ -0,0 +1,45 @@
...
+        description: 'Run the formatter to check formatting rather than updating code format'
+      })
+      .command(
+          'all', 'Run the formatter on all files in the repository', {},
+          ({check}) => {
+            const executionCmd = check ? checkFiles : formatFiles;
+            executionCmd(allFiles());
+          })
+      .command(
+          'changed [shaOrRef]', 'Run the formatter on files changed since the provide sha/ref', {},
```
gkalpak, line 31:
```suggestion
          'changed [shaOrRef]', 'Run the formatter on files changed since the provided sha/ref', {},
```

gkalpak (changes requested):
Drive-by review :grin:

dev-infra/format/format.ts
josephperrott: Evidently thats true, I have run into a situation in the past where `**` was 1 or more, rather than 0 or more

dev-infra/format/format.ts
josephperrott: It appears to still be of value.

Running format for this change on itself (enforced on 8 of 24 files):
1 process: 2.04s
8 processes: 1.49s

Running format for this change 19 random files in the repo:
1 process: 2.81s
8 processes: 1.29s

josephperrott:
@gkalpak Comments addressed, PTAL.

gkalpak (approved):

AndrewKushnir:
Hi @josephperrott, could you please rebase this PR and resolve conflicts? Thank you.

josephperrott:
@googlebot Look again please!

PR #36729, opened 2020-04-20
build: update to @bazel/bazelisk@^1.4.0

josephperrott:
Upgrading `@bazel/bazelisk` to version `1.4.0` as this introduces the
`bazel` binary.  This prevents the need to have a `bazel` script defined
in package.json to point to `bazelisk`, instead it is just available
on install.


IgorMinar (approved):
lgtm. thanks!

PR #36731, opened 2020-04-20
feat(common): introduce the `watchCollection` pipe

matsko:
The `watchCollection` pipe is used to track the contents of an array or key/value map and then create a new copy once a change is detected.

This pipe can be used to signal input bindings whenever the contents of a collection change.

```html
<!-- [values] only gets fired when the `myValues` input
     value changes in identity (when the value is rewritten) -->
<my-comp [values]="myValues"></my-comp>

<!-- [values] will now get updated whenever the `myValues`
     collection changes in identity or content -->
<my-comp [values]="myValues | watchCollection"></my-comp>
```

### DIFF
--- a/goldens/public-api/common/common.d.ts
+++ b/goldens/public-api/common/common.d.ts
@@ -416,6 +416,12 @@ export declare abstract class ViewportScroller {
     static ɵprov: never;
 }
 
+export declare class WatchCollectionPipe implements PipeTransform {
+    transform(value: {
+        [key: string]: any;
+    } | any[] | null | undefined): unknown;
+}
+
 export declare enum WeekDay {
     Sunday = 0,
     Monday = 1,
--- a/packages/common/test/pipes/value_harness_spec.ts
+++ b/packages/common/test/pipes/value_harness_spec.ts
@@ -1,0 +1,144 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import {ArrayHarness, DefaultValueHarness, MapHarness} from '@angular/common/src/pipes/value_harness';
+
+fdescribe('ValueHarness', () => {
+  describe('ArrayHarness', () => {
+    it('should produce a new array whenever the contents differ', () => {
+      const a = [1, 2, 3];
+      const h = new ArrayHarness(a);
+
+      const v1 = h.value;
+      expect(v1).toEqual([1, 2, 3]);
+
+      h.update([1, 2, 3]);
+
+      const v2 = h.value;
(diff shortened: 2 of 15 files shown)


mlc-mlapis:
@matsko I understand the basic logic of it, but still, I don't understand why it should be introduced right now. It's not a pure pipe, so there is a performance hit and so. Can you share a bit more about the background of the origin?

matsko:
@mlc-mlapis we're trying to change things around with `ngStyle` / `ngClass`.

packages/common/test/pipes/value_harness_spec.ts
```diff
@@ -0,0 +1,144 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import {ArrayHarness, DefaultValueHarness, MapHarness} from '@angular/common/src/pipes/value_harness';
+
+fdescribe('ValueHarness', () => {
```
IgorMinar, line 10:
```suggestion
describe('ValueHarness', () => {
```

goldens/public-api/common/common.d.ts
```diff
@@ -416,6 +416,12 @@ export declare abstract class ViewportScroller {
     static ɵprov: never;
 }
 
+export declare class WatchCollectionPipe implements PipeTransform {
+    transform(value: {
+        [key: string]: any;
+    } | any[] | null | undefined): unknown;
+}
```
IgorMinar, line 423: how about this more type-safe version:

```
export class WatchCollectionPipe implements PipeTransform {
    transform<T extends { [key: string]: unknown } | unknown[] | null | undefined>(value: T): T {
        return value;
    }
}
```
[Playground Link](https://www.typescriptlang.org/play/?ssl=26&ssc=2&pln=22&pc=1#)

IgorMinar (changes requested):
This PR still seems as WIP. Could you please label is as such? and consider turning it into a draft PR so that it's not occupying our review queues? thanks

Reviewed-for: public-api

packages/core/test/render3/perf/ng_for/util.ts
```diff
@@ -0,0 +1,148 @@
...
+      '-----------------------------------------------\n' + name + '\n' +
+      '-----------------------------------------------');
+}
+
+/**
+ * Returns and resolves a promise once the program is less busy.
+ *
+ * See `setImmediate`
+ */
+export function untilLessBusy() {
```
atscott, line 117: Can we remove this now? I don't know that it ended up making things more stable and it also doesn't work when running in the browser (`setImmediate` is not defined)

jelbourn:
I'm unclear on the status of this PR. Is it ready for review? Could it be put back into draft until the CI is green?

PR #36733, opened 2020-04-20
fix(dev-infra): extract commit headers before checking commit message validity

AndrewKushnir:
This commit fixes an issue where adding `fixup` commits was triggering a lint error. The problem was caused by the fact that we used the entire message body while checking whether `fixup` commit has a corresponding "parent" commit in a range. This issue was found after enforcing a check that exits the process if there is an invalid commit message found (https://github.com/angular/angular/commit/4341743b4a6d7e23c6f944aa9e34166b701369a1).

## PR Type

- [x] Bugfix

- [ ] Yes
- [x] No


AndrewKushnir:
cc @josephperrott

josephperrott (approved):
LGTM

devversion (approved):

PR #36739, opened 2020-04-21
fix(core): support `Attribute` DI decorator in `deps` section of a token

### LINKED ISSUE #36479: Attribute in deps stopped working with Ivy
# 🐞 bug report

### Affected Package

### Is this a regression?

 Yes, the previous version in which this bug was not present was: 8

### Description

You can use `Attribute` decorator in `deps` for providers like this:

```
{
    provide: TOKEN,
    deps: [[new Attribute('token')]],
    useFactory: factory,
}
```
It stopped working with Ivy and Angular 9

## 🔬 Minimal Reproduction

 
Working Angular 8:
https://stackblitz.com/edit/angular-attribute-deps

Broken Angular 9:
https://ng-run.com/edit/hyppU3cXJCaRYLYVDAQp

## 🔥 Exception or Error
<pre><code>
R3InjectorError(AppModule)[[object Object] -> [object Object] -> [object Object]]: 
  NullInjectorError: No provider for [object Object]!
</code></pre>

## 🌍  Your Environment

**Angular Version:**
<pre><code>9.1.0</code></pre>

sonukapoor:
Closes #36479

## PR Checklist

## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #36479

- [ ] Yes
- [X] No


packages/core/src/di/util.ts
```diff
@@ -55,3 +59,47 @@ export function convertInjectableProviderToFactory(
...
+
+      args.push(injectFn(type!, flags));
+    } else {
+      args.push(ɵɵinject(arg));
+    }
+  }
+  return args;
+}
+
+function isValidInstance(
```
AndrewKushnir, line 101: A few notes on this function:
- may be call it `isValidDecorator`, since the objects that are used represent DI-related decorators (and also rename the second argument)
- for the type of the second argument, let's remove `any` and add `Inject` - that should cover all types used in the `injectArgs` function
- nit: rename `metaDataName` to `metadataName` (lowercase "D") - it's be more inline with the `ngMetadataName` filed name
- there is an additional check `meta !== ''` added to the function (it was nor present in original function) - do we need it?

sonukapoor: I tried `Inject` as the second parameter but get the below typescript error for this condition `meta instanceof instance`:

> The right-hand side of an 'instanceof' expression must be of type 'any' or of a type assignable to the 'Function' interface type.

I have added the `meta !== ''` for the third `else if` statement. We do not want to check for the `ngMetadataName` in that case.

I will make the other changes.

AndrewKushnir: > I tried Inject as the second parameter but get the below typescript error for this condition meta instanceof instance

Ok, thanks for additional info. I think we still need to have stricter types for the `isValidDecorator` function arguments and we can relax this inside the function if needed, like `meta instanceof (decorator as any)`, so the function may look like:

```typescript
function isValidDecorator(
    meta: any, decorator: Optional|SkipSelf|Self|Attribute|Inject, metadataName: string = '') {
  return meta instanceof (decorator as any) || meta.ngMetadataName === metadataName ||
      (meta !== '' && meta === decorator);
} 
```

However I think we can improve it further if we use corresponding types, i.e. instead of `Self` use `SelfDecorator`, `SkipSelf` -> `SkipSelfDecorator`, etc. These types are available in the same `metadata.ts` file, so the end result will look like this:

```typescript
function isValidDecorator(
    meta: any, decorator: OptionalDecorator|SkipSelfDecorator|SelfDecorator|AttributeDecorator|InjectDecorator, metadataName: string = '') {
  return meta instanceof decorator || meta.ngMetadataName === metadataName ||
      (meta !== '' && meta === decorator);
} 
```

Also, it looks like the `meta !== ''` condition was added to the wrong place, shouldn't it guard the `meta.ngMetadataName === metadataName` check?

sonukapoor: Thanks for the suggestions @AndrewKushnir - I will make the changes to the input types. 

WRT the last condition: The purpose of adding the check was to allow the `Inject` condition to pass without the `metadataName` check. 

Therefore, the correct condition I believe should be: `(metadataName !== '' && meta instanceof decorator)`. WDYT?

packages/core/src/di/util.ts
```diff
@@ -8,11 +8,15 @@
...
 import {Type} from '../interface/type';
 import {ReflectionCapabilities} from '../reflection/reflection_capabilities';
+import {ɵɵinjectAttribute} from '../render3/instructions/di';
 import {getClosureSafeProperty} from '../util/property';
 
 import {resolveForwardRef} from './forward_ref';
-import {injectArgs, ɵɵinject} from './injector_compatibility';
+import {InjectionToken} from './injection_token';
+import {ɵɵinject} from './injector_compatibility';
+import {InjectFlags} from './interface/injector';
```
sonukapoor, line 17: These three tokens are imported separately to avoid circular dependencies and to keep the payload size low.

goldens/size-tracking/integration-payloads.json
```diff
@@ -30,7 +30,7 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 1485,
-        "main-es2015": 137320,
+        "main-es2015": 138095,
```
sonukapoor, line 33: The payload increase is caused by the `injectAttribute` function.

packages/core/src/render3/instructions/di.ts
```diff
@@ -5,12 +5,13 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
-import {InjectFlags, InjectionToken, resolveForwardRef} from '../../di';
```
sonukapoor, line 8: We are avoiding to import from `di` to reduce the circular dependencies.

JoostK:
Drive by comment:

`injectArgs` is shared among VE and Ivy, so pulling in `injectAttribute` would probably be a major size regression in VE as it would be referencing Ivy instructions.

Additionally, given that this usage pattern should be really rare, I would like to explore a design where the injectAttribute is patched onto the `Attribute` decorator somehow, so that it's only referenced when the decorator survives. This should allow tree-shaking of `injectAttribute` for the common case. Notice that there's also FW-490 which may interact with this PR, although I'm not sure what the plan is on that Jira.

sonukapoor:
> Additionally, given that this usage pattern should be really rare, I would like to explore a design where the injectAttribute is patched onto the `Attribute` decorator somehow, so that it's only referenced when the decorator survives. This should allow tree-shaking of `injectAttribute` for the common case. Notice that there's also [FW-490](https://angular-team.atlassian.net/browse/FW-490) which may interact with this PR, although I'm not sure what the plan is on that Jira.

@JoostK Thanks for the feedback.

@AndrewKushnir  / @kara Is this the approach we want to take? I have no visibility into the Jira yet.

waterplea:
Any news on when this might make it into release? Lack of Attribute support blocks my migration into Ivy on several projects :( Not that they are big, but still a bummer :) Thanks for a quick pickup, by the way!

AndrewKushnir:
Hi @waterplea, thanks for reporting the issue.

We are looking for a more optimized way of handing this (to allow for better tree-shaking), so it may take a bit more time. If that's the only blocker for you to transition to Ivy, may be check if you could avoid using `Attribute` in `deps` array for now and revert it back to the `Attribute` once solution is developed and the fix is released.

Thank you.

waterplea:
I use it in directives declaring some DI items for the lower views and it's the only way to customize it. I used this technique in quite a lot of my open source libraries for native APIs: ng-web-apis.github.io

I can query ElementRef.nativeElement, of course, but that's such a hack :-(

AndrewKushnir:
Closing in favor of #37085.

PR #36741, opened 2020-04-21
fix(compiler): normalize line endings in ICU expansions

### LINKED ISSUE #36725: Angular 9 during language build causes "No translation found for "<ID>" (" {$ICU} ").
# 🐞 bug report

### Affected Package
With Ivy disabled the language build succeeds

### Is this a regression?
Yes, this error started appearing after our upgrade from 8.2 to 9.1.1.  Disabling Ivy is our current workaround.

### Description
The issue only appears to affect translations that make use of the ICU select format. After further testing it seems that ivy doesn't like it when a value is passed to a component via Input().

```
<!-- app.component.html -->
<app-boolean-translation [value]="value"></app-boolean-translation>
```

```
// app.component.ts

import { Component } from '@angular/core';

@Component({
  selector: 'app-root',
  templateUrl: './app.component.html',
})
export class AppComponent {
  value = 'TRUE';
}

```

```
// boolean-translation.component.ts

import { Component, Input, ChangeDetectionStrategy } from '@angular/core';

@Component({
  selector: 'app-boolean-translation',
  template: `
    <div i18n>
      {
        value, select,
        TRUE {Yes}
        FALSE {No}
      }
    </div>
  `,
  changeDetection: ChangeDetectionStrategy.OnPush,
})
export class BooleanTranslationComponent {
  @Input() value: string;
}
```

## 🔬 Minimal Reproduction
https://github.com/FlashBanistan/translation-error-repro

1. Clone repo.
2. run 'npm install'
3. run 'ng build --configuration=es'

## 🔥 Exception or Error
<pre><code>
No translation found for "8856905278208146821" (" {$ICU} ").
No translation found for "4422277569170046485" ("{VAR_SELECT, select, TRUE {Yes} FALSE {No}}").
</code></pre>

## 🌍  Your Environment

**Angular Version:**
<pre><code>
Angular CLI: 9.1.1
Node: 12.16.2
OS: win32 x64

Angular: 9.1.1
Ivy Workspace: Yes

@angular-devkit/architect          0.901.1
@angular-devkit/build-angular      0.901.1
@angular-devkit/build-ng-packagr   0.901.1
@angular-devkit/build-optimizer    0.901.1
@angular-devkit/build-webpack      0.901.1
@angular-devkit/core               9.1.1
@angular-devkit/schematics         9.1.1
@angular/flex-layout               9.0.0-beta.29
@angular/material                  9.2.0
@angular/tsc-wrapped               4.4.7
@ngtools/webpack                   9.1.1
@schematics/angular                9.1.1
@schematics/update                 0.901.1
ng-packagr                         9.1.1
rxjs                               6.5.3
</code></pre>

petebacondarwin:
The html parser already normalizes line endings (converting
`\r\n` to `\n`) for most text in templates but it was missing
the expressions of ICU expansions.

In ViewEngine backticked literal strings used to define inline
templates were already normalized by the TypeScript parser.
In Ivy we are parsing the raw text of the source file directly so
the line ending need to be manually normalized.

This change in behaviour was causing the computed i18n message
id for ICUs to change depending upon the current line ending settings
of the environment extracting or merging translations.

Closes #36725

----

Capturing the warning message content here:

When parsing an external template file "TestComp" one or more ICU expressions were found that contain `\\r\\n` line-endings.
These would have been normalized to `\\n` but this would change the computed id for their i18n messages.
The line endings were left unnormalized for backward compatibility on this occasion.
In a future release, these line endings will be normalized, which will be a breaking change.
It is suggested that you try opting into this behavior now to see what effect it will have on your message ids by setting `"i18nNormalizeExternalTemplateLineEndingsInICUs": true` in the `angularCompilerOptions` of `tsconfig.json`
The ICU expressions that need normalizing are:


packages/compiler/test/ml_parser/html_parser_spec.ts
```diff
@@ -278,6 +312,36 @@ import {humanizeDom, humanizeDomSourceSpans, humanizeLineColumn} from './ast_spe
...
+                  `    plural,\r\n` +
+                  `    =0 {You have \r\nno\r\n messages}\r\n` +
+                  `    =1 {One {{message}}}}\r\n` +
+                  `</div>`,
+              'TestComp', {tokenizeExpansionForms: true});
+
+          expect(humanizeDom(parsed)).toEqual([
+            [html.Element, 'div', 0],
+            [html.Text, '\n  ', 1],
+            [html.Expansion, '\n    messages.length', 'plural', 1],
```
petebacondarwin, line 329: @AndrewKushnir - note that while the expansion cases have their whitespace trimmed, the expression does not.

I think this is actually a bug, but we cannot change it without breaking all the message ids. So in order to do so we would need to come up with a migration plan for all the translation files that are already relying upon the "wrong" message id.

petebacondarwin:
It turns out that things are really broken as they stand, but that also means that this would be a breaking change for a specific scenario...

On a machine where line endings are set to be `\r\n`, there is a difference between inline and external templates. TypeScript will normalise line endings in backtick string literals to be `\n`. So in ViewEngine the line ending doesn't matter. But in external templates (i.e. `templateUrl`) TypeScript does not get a look in, and we process the contents of the file directly, which means that the line endings are taken into account.

As shown in this PR, the only thing that is affected is the ICU `expression` string - the rest of ICUs and in fact the rest of templates already have the line endings normalised when we parse templates.

This is quite broken, right now, since two people working on different machines with different line ending settings would generate different message ids for the same external templates.

The right thing to do is to fix this but then Windows users with `\r\n` line endings in their external templates will find that their ICU message IDs have changed.

petebacondarwin:
I added a fixup commit that should make this change non-breaking - and provides a opt-in switch that we can flip in the future.

packages/compiler-cli/ngcc/src/analysis/decoration_analyzer.ts
```diff
@@ -96,8 +96,9 @@ export class DecorationAnalyzer {
         this.scopeRegistry, this.scopeRegistry, this.isCore, this.resourceManager, this.rootDirs,
         !!this.compilerOptions.preserveWhitespaces,
         /* i18nUseExternalIds */ true, this.bundle.enableI18nLegacyMessageIdFormat,
-        this.moduleResolver, this.cycleAnalyzer, this.refEmitter, NOOP_DEFAULT_IMPORT_RECORDER,
-        NOOP_DEPENDENCY_TRACKER, this.injectableRegistry, /* annotateForClosureCompiler */ false),
+        /* normalizeExternalTemplateLineEndingsInICUs */ false, this.moduleResolver,
```
petebacondarwin, line 99: In ngcc we are always working with templates that are inline, since they have been previously built by ViewEngine, so really this could be `true|false|undefined` here and we would get the same result.

packages/compiler/src/ml_parser/lexer.ts
```diff
@@ -608,7 +622,13 @@ class _Tokenizer {
 
     this._beginToken(TokenType.RAW_TEXT);
     const condition = this._readUntil(chars.$COMMA);
-    this._endToken([condition]);
+    if (this._escapedString || this._normalizeExternalTemplateLineEndingsInICUs) {
+      this._endToken([this._processCarriageReturns(condition)]);
+    } else {
+      // This expression is in an external template and for backward compatibility we are not
```
AndrewKushnir, line 628: Hi @petebacondarwin, I did a first review pass (will complete it tomorrow) and I have a quick question on this section: may be we can extend this a bit and log a message that would explain that we come across a case that would be affected by the breaking change (when we decide to flip the default) and encourage opting-in? Something like:
```
  const processedCondition = this._processCarriageReturns(condition);
  if (processedCondition !== condition) {
    // ... log a message here ...
  }
```

petebacondarwin: Great idea!

packages/compiler-cli/src/ngtsc/annotations/src/component.ts
```diff
@@ -258,6 +258,11 @@ export class ComponentDecoratorHandler implements
           `Errors parsing template: ${template.errors.map(e => e.toString()).join(', ')}`);
     }
 
+    // Report any informational warnings to the developer
+    for (const warning of template.warnings) {
+      console.warn(warning);
+    }
```
petebacondarwin, line 265: @alxhub - is this the best approach to surface these warnings? I couldn't see any other non-fatal mechanism for reporting non-errors.

petebacondarwin (commented):
@AndrewKushnir I worked the PR over quite substantially, so I didn't bother to retain fixup commits. (Sorry). PTAL.

packages/compiler/src/render3/view/template.ts
```diff
@@ -1986,6 +1986,15 @@ export interface ParseTemplateOptions {
    * `$localize` message id format and you are not using compile time translation merging.
    */
   enableI18nLegacyMessageIdFormat?: boolean;
+  /**
+   * If this text is stored in an external template (e.g. via `templateUrl`) then we need to decide
+   * whether or not to normalize the line-endings (from `\r\n` to `\n`) when processing ICU
+   * expressions.
+   *
+   * If `true` then we will normalize ICU expression line endings.
+   * The default is `false`, but this will be switched in a future major release.
```
JoostK, line 1995: Is there a Jira to track this?

packages/compiler/src/ml_parser/lexer.ts
```diff
@@ -204,7 +219,22 @@ class _Tokenizer {
...
+          `When parsing an external template file "${this._file.url}" ` +
+          'one or more ICU expressions were found that contain `\\r\\n` line-endings.\n' +
+          'These should have been normalized to `\\n` but this would change the computed id for their i18n messages.\n' +
+          'The line endings were left un-normalized for backward compatibility on this occasion.\n' +
+          'In a future release, these line-endings will be normalized, which will be a breaking change.\n' +
+          'It is suggested that you try opting into this behavior now to see what effect it will have on your message ids\n' +
+          'by setting `"normalizeExternalTemplateLineEndingsInICUs": true` in the `angularCompilerOptions` of `tsconfig.json`.\n' +
+          'The ICU expressions that need normalizing are:\n' +
+          this.nonNormalizedIcuExpressions.map(
+              t => ` - "${t.sourceSpan.toString().replace('\n', '\\n').replace('\r', '\\r')}"\n`));
```
ayazhafiz, line 235: bunch of nits
```suggestion
          `When parsing an external template file "${this._file.url}" ` +
          'one or more ICU expressions were found that contain `\\r\\n` line-endings.\n' +
          'These would have been normalized to `\\n` but this would change the computed id for their i18n messages.\n' +
          'The line endings were left unnormalized for backward compatibility on this occasion.\n' +
          'In a future release, these line endings will be normalized, which will be a breaking change.\n' +
          'It is suggested that you try opting into this behavior now to see what effect it will have on your message ids\n' +
          'by setting `"normalizeExternalTemplateLineEndingsInICUs": true` in the `angularCompilerOptions` of `tsconfig.json`.\n' +
          'The ICU expressions that need normalizing are:\n' +
          this.nonNormalizedIcuExpressions.map(
              t => ` - "${t.sourceSpan.toString().replace('\n', '\\n').replace('\r', '\\r')}"\n`));
```

packages/compiler/test/ml_parser/html_parser_spec.ts
ayazhafiz: ExpansionCases should not have their whitespace trimmed in this context, correct?

ayazhafiz (approved):

ayazhafiz:
LGTM after @JoostK's comments are addressed

goldens/public-api/compiler-cli/compiler_options.d.ts
```diff
@@ -7,6 +7,7 @@ export interface I18nOptions {
     enableI18nLegacyMessageIdFormat?: boolean;
     i18nInLocale?: string;
     i18nUseExternalIds?: boolean;
+    normalizeExternalTemplateLineEndingsInICUs?: boolean;
```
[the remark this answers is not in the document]
petebacondarwin: Good idea.

packages/compiler/src/render3/view/template.ts
petebacondarwin: There is now :-) FW-2106

packages/compiler/test/ml_parser/html_parser_spec.ts
petebacondarwin: ExpansionCases have always had their whitespace trimmed because of these lines:

https://github.com/angular/angular/blob/master/packages/compiler/src/ml_parser/lexer.ts#L626-L631

petebacondarwin:
Marked as `master-only` because it touches the public API (of the compiler options).

goldens/public-api/compiler-cli/compiler_options.d.ts
```diff
@@ -7,6 +7,7 @@ export interface I18nOptions {
     enableI18nLegacyMessageIdFormat?: boolean;
     i18nInLocale?: string;
     i18nUseExternalIds?: boolean;
+    i18nNormalizeExternalTemplateLineEndingsInICUs?: boolean;
```
AndrewKushnir, line 10: I'd propose adding a few tests to verify/illustrate that there are no other cases where line endings might become a problem for message extraction and id generation. If we identify other cases, we can consider doing similar changes and reusing this flag (rename it slightly) to provide backwards compatibility.

Some basic cases that I can think of:
- i18n blocks:
```
<div i18n> \r\n
 Some content \r\n
 across multiple lines \r\n
</div>
```
- i18n attributes:
```
<div i18n-title title=" \r\n
Some title \r\n
across multiple lines \r\n
">
```
- i18n ICUs (you have this case in other tests in this PR)
- nested i18n ICUs:
```
{
  messages.length, plural,
  =0 {zero \r\n {type, select, A {type \r\n A} other {OTHER}}}
  =1 {one \r\n {type, select, B {type \r\n B} other {OTHER}}}
}
```
What do you think Pete?

packages/compiler-cli/src/ngtsc/annotations/src/component.ts
AndrewKushnir: @petebacondarwin may be add a `TODO(FW-963): ...` to take care of it in a separate PR?

packages/compiler/src/ml_parser/lexer.ts
```diff
@@ -204,7 +219,22 @@ class _Tokenizer {
...
+      warnings.push(
+          `When parsing an external template file "${this._file.url}" ` +
+          'one or more ICU expressions were found that contain `\\r\\n` line-endings.\n' +
+          'These would have been normalized to `\\n` but this would change the computed id for their i18n messages.\n' +
+          'The line endings were left unnormalized for backward compatibility on this occasion.\n' +
+          'In a future release, these line endings will be normalized, which will be a breaking change.\n' +
+          'It is suggested that you try opting into this behavior now to see what effect it will have on your message ids\n' +
+          'by setting `"i18nNormalizeExternalTemplateLineEndingsInICUs": true` in the `angularCompilerOptions` of `tsconfig.json`.\n' +
+          'The ICU expressions that need normalizing are:\n' +
+          this.nonNormalizedIcuExpressions.map(
```
AndrewKushnir, line 234: It looks like for nested ICUs we might output the same ICU multiple times, for example:

```
{
  messages.length, plural,
  =0 {zero \r\n {type, select, A {type \r\n A} other {OTHER}}}
  =1 {one \r\n {type, select, B {type \r\n B} other {OTHER}}}
}
```
May be use `Set` instead of array to store problematic ICUs to get rid of duplicates?

AndrewKushnir (changes requested):
Thanks for the fix and expanding test coverage @petebacondarwin! 👍

I've added a couple comments to see if we can add extra tests to verify that this problem doesn't exist in other cases that might affect message id generation. Please let me know what you think about it Pete.

Thank you.

goldens/public-api/compiler-cli/compiler_options.d.ts
IgorMinar, line 10: my attempt at making this shorter and based on existing naming precedences... what do you think?
```suggestion
    i18nNormalizeExternalTemplateEOLsInICUs?: boolean;
```

I don't love this, because it's still too cryptic and long, also consider `i18nNormalizeEOLsInICUs` - yes, it's more ambiguous but we can clarify in the docs that this applies only to external templates (the definition of "external" will need to be part of the docs anyway, as most people don't understand what "external template" means).

JoostK:
Something that came to mind: if there's a lot of components for which a warning is generated, the compilation will generate a lot of long warning messages. Maybe we can find a way to accumulate all occurrences and log them once.

packages/compiler/src/ml_parser/lexer.ts
petebacondarwin: I'll add a test and try to avoid that.

goldens/public-api/compiler-cli/compiler_options.d.ts
petebacondarwin: Text within blocks have tests already.
I will add for attributes, and for nested ICUs.

petebacondarwin: Or are you suggesting that html_parser_spec tests are not enough and you want to test that message ids are computed correctly still?

petebacondarwin: My suggestion is `i18nNormalizeLineEndingsInICUs`.
It is not 100% accurate because line endings in ICUs are always normalized for inline templates - and this flag only affects the normalization of ICU expressions in external templates. But I think that it is close enough and short enough.


Igor replied:

> I like your suggestion, but can we also take into consideration that tsconfig.json has a somewhat related precedence for this: https://www.typescriptlang.org/v2/en/tsconfig#newLine we might want to take that into account when deciding the name and intentionally decide to either follow it if the concepts are the same, or intentionally decide to diverge if we are trying to disambiguate between two similar concepts.

petebacondarwin: This is actually not really related that that standard tsconfig setting. That is only about outputting line endings. Whatever the value of that setting, TS always normalizes line endings in backticked strings.

packages/compiler/src/ml_parser/lexer.ts
petebacondarwin: Actually it is preferable to list both instances annotated with their location in the file, since people would need to know where the expressions are that need fixing.

goldens/public-api/compiler-cli/compiler_options.d.ts
IgorMinar: great thanks for the context in that case `i18nNormalizeLineEndingsInICUs` sgtm

IgorMinar: please re-request the review when the changes are made.

petebacondarwin:
> Something that came to mind: if there's a lot of components for which a warning is generated, the compilation will generate a lot of long warning messages. Maybe we can find a way to accumulate all occurrences and log them once.

Actually thinking it, this is a significant issue. There is no straightforward way to aggregate these warnings without making the compiler know more than it ought to about how the warning is structured...

I wonder if we could hide this warning behind a verbose option?
Or perhaps post a simple one line warning and write some documentation for it? Perhaps a link to this PR?
Or perhaps for now we should just not display the warning?

petebacondarwin:
I decided to go with not writing any warning to the console at this stage. Apart from anything, this does not really play nice with CLI integration.

packages/compiler/test/ml_parser/html_parser_spec.ts
```diff
@@ -309,6 +479,71 @@ import {humanizeDom, humanizeDomSourceSpans, humanizeLineColumn} from './ast_spe
...
+            [html.Expansion, '\n         p.gender', 'select', 0],
+            [html.ExpansionCase, 'male', 1],
+            [html.Text, '\n     ', 0],
+          ]);
+
+          expect(parsed.errors).toEqual([]);
+          expect(parsed.nonNormalizedIcuExpressions).toEqual([]);
+        });
+
+        it('should not normalize line endings in nested expansion forms for external templates',
```
JoostK, line 511: This behaviour is also dependent on the `normalizeExternalTemplateLineEndingsInICUs` flag, no? If so, I would include it in this spec to make it clear that it's still backwards compatible behaviour and not necessarily the desired behaviour.

packages/compiler-cli/src/ngtsc/core/api/src/public_options.ts
```diff
@@ -313,6 +313,19 @@ export interface I18nOptions {
...
+   * whether or not to normalize the line-endings (from `\r\n` to `\n`) when processing ICU
+   * expressions.
+   *
+   * Ideally we would always normalize, but for backward compatibility this flag allows the template
+   * parser to avoid normalizing line endings in ICU expressions.
+   *
+   * If `true` then we will normalize ICU expression line endings.
+   * The default is `false`, but this will be switched in a future major release.
+   */
+  i18nNormalizeLineEndingsInICUs?: boolean;
```
JoostK, line 328: What do you think about adding an ngtsc spec that exercises this flag?

JoostK (commented):
Now that no warning is being generated, this should be fine to go into patch once again.

packages/compiler/src/ml_parser/parser.ts
```diff
@@ -23,39 +23,41 @@ export class TreeError extends ParseError {
 }
 
 export class ParseTreeResult {
-  constructor(public rootNodes: html.Node[], public errors: ParseError[]) {}
+  constructor(
+      public rootNodes: html.Node[], public errors: ParseError[],
+      public nonNormalizedIcuExpressions?: lex.Token[]) {}
```
AndrewKushnir, line 28: I'm curious if we can leave different forms of parsed ICUs as an internal implementation of Lexer and do not propagate it to Parser (it looks like Parser just proxy them into the final output)?

petebacondarwin: Yes these are just proxying them out. The reason I did this is that I thought that if we were to want to pick up these non-normalized expressions it would be done at the template parser level...
But given that we are not doing anything with them right now, I will remove them from the parser and move the tests for them into the lexer_spec

packages/compiler-cli/src/ngtsc/core/api/src/public_options.ts
petebacondarwin: Good idea

packages/compiler/test/ml_parser/html_parser_spec.ts
petebacondarwin: do you mean add `'when `normalizeExternalTemplateLineEndingsInICUs` is not set' to the test description or something else?

JoostK: Yes, that would make it more clear :+1:

packages/compiler/src/ml_parser/parser.ts
AndrewKushnir: Thanks Pete, sounds good!

goldens/public-api/compiler-cli/compiler_options.d.ts
AndrewKushnir: Thanks for the reply Pete. My main idea is to check whether we have additional cases like this. I think html_parser_spec that you mentioned gives sufficient level of verification and we can do more checks in a separate task/PR. Thank you.

AndrewKushnir (approved):

packages/compiler-cli/ngcc/src/analysis/decoration_analyzer.ts
IgorMinar: Instead of undefined could we specify what the default value is if one is not explicitly provided? Often this is an option that makes the api simpler and less ambiguous but we don't realize that we have that option available. Could this be one of those cases?

goldens/public-api/compiler-cli/compiler_options.d.ts
```diff
@@ -6,6 +6,7 @@ export interface BazelAndG3Options {
 export interface I18nOptions {
     enableI18nLegacyMessageIdFormat?: boolean;
     i18nInLocale?: string;
+    i18nNormalizeLineEndingsInICUs?: boolean;
```
IgorMinar, line 9: I like the name we ended with! Thanks for iterating on it.

IgorMinar (commented):
Let's see if we can consider using a default value in the public API instead of undefined.

petebacondarwin:
@IgorMinar can you clarify what you mean by:

> Let's see if we can consider using a default value in the public API instead of undefined.

petebacondarwin:
> Now that no warning is being generated, this should be fine to go into patch once again.

@JoostK - the `master-only` label is due to the public API change, not because of the warning.

IgorMinar:
> @IgorMinar can you clarify what you mean by:
> 
> > Let's see if we can consider using a default value in the public API instead of undefined.

Please take a look at: https://github.com/angular/angular/pull/36741#discussion_r415136980

packages/compiler-cli/ngcc/src/analysis/decoration_analyzer.ts
petebacondarwin: I am not sure what is being asked for here...

Are you saying that the decision about the default (i.e. whether the switch defaults to `true` or `false` when it is not provided) should be done at the highest level possible?

petebacondarwin: All the `TokenizeOptions` non-required, and this gets passed through to there. I felt that it was better to follow this pattern; and in that case it is better for there to be only one place where the `undefined` gets converted to `true` or `false`. See https://github.com/angular/angular/blob/master/packages/compiler/src/ml_parser/lexer.ts#L64-L108

If we make this option required at the top level then you would have a single required option in `TokenizeOptions`, which doesn't feel right.

petebacondarwin:
I added integration tests for ViewEngine and Ivy that check the generated message ids.

goldens/public-api/compiler-cli/compiler_options.d.ts
IgorMinar, line 9: What I was trying to say in https://github.com/angular/angular/pull/36741/files#r415136980 was: could this public API, that is proposed to be an optional prop in this PR, be instead changed to be a prop with a default value? That way we don't need to support the `undefined` value type which is a big win as it makes the API more explicit and self-documenting.

packages/compiler-cli/ngcc/src/analysis/decoration_analyzer.ts
IgorMinar, line 99: plz update the comment with the new prop name

IgorMinar: ah.. I missed that the `I18nOptions` is an interface.

IgorMinar (approved):
LGTM! I have no more concerns except for that one inline comment update. Thanks!

Reviewed-for: public-api

packages/compiler-cli/ngcc/src/analysis/decoration_analyzer.ts
petebacondarwin: This is actually an internal argument name for the `ComponentHandler`. And I thought we might leave it as the more explicit name, since it is not public. But given that you mentioned it, I am happy to change it to be in sync with the public property.

packages/compiler-cli/src/ngtsc/core/api/src/public_options.ts
petebacondarwin: I added integration tests for Ivy i18n that check this.

gkalpak (approved):
Reviewed-for: fw-ngcc

AndrewKushnir:
[Presubmit](http://test/OCL:308852114:BASE:308852155:1588095793261:2278bccc)

PR #36742, opened 2020-04-21
test(router): fix router test failing on IE

atscott:
This was originally fixed in #35976, but one of the `window.scrollY`
assertions was missed. Also updated tests to use toBeGreater/LessThan
to improve failure messages.


gregmagolan (approved):
LGTM

gkalpak (approved):

atscott:
merge assistance: I'm the only owner of router so pullapprove can't be green without global approver.

PR #36745, opened 2020-04-21
refactor(localize): miscellaneous refactorings

petebacondarwin:


packages/localize/src/tools/src/source_file_utils.ts
```diff
@@ -355,3 +354,12 @@ export function buildCodeFrameError(path: NodePath, e: BabelParseError): string
...
   return `${filename}: ${message}`;
 }
+
+export function getLocation(path: NodePath): ɵSourceLocation|undefined {
+  return path.node.loc && path.hub.file.opts.filename ? {
+    start: {...path.node.loc.start},
+    end: {...path.node.loc.end},
+    file: path.hub.file.opts.filename
+  } :
+                                                        undefined;
```
AndrewKushnir, line 364: Formatting is a bit unusual :) Maybe do an `if` instead?

packages/localize/src/utils/src/messages.ts
```diff
@@ -38,6 +38,52 @@ export type TargetMessage = string;
...
+   */
+  legacyIds?: string[];
+  /**
+   * The meaning of the `message`, used to distinguish identical `messageString`s.
+   */
+  meaning?: string;
+  /**
+   * The description of the `message`, used to aid translation.
+   */
+  description?: string|undefined;
```
AndrewKushnir, line 80: I guess we don't need `undefined` here?

AndrewKushnir (approved):
LGTM, thanks @petebacondarwin 👍  Just left a couple minor comments.

matsko:
@petebacondarwin this fails on patch. Please open a new PR to address the `patch` branch directly.

matsko:
@petebacondarwin please rebase.

petebacondarwin:
@matsko  - done.

petebacondarwin:
@matsko - I assume that you removed the `patch` target label because this will not land cleanly on patch rather than because it is not valid for patch?

PR #36749, opened 2020-04-21
build: fix the compare master and patch script output

AndrewKushnir:
Currently the commit message and corresponding version are flipped, which makes it hard to review the changes. This commit updates the script to properly recognize the order of arguments.

## PR Type

- [x] Bugfix

- [ ] Yes
- [x] No


IgorMinar (approved):
LGTM - but we should start thinking about creating a test suite for this script to prevent future regressions.

PR #36751, opened 2020-04-22
revert: fix(common): format day-periods that cross midnight (#36611)

matsko:
This reverts commit 1756cced4a59bc1185653c6c9b8f661636fd6116.

The reason for this revert is because of the change being too risky for
a patch release of Angular. Changing date formatting proves to be a
breaking change and can only be apart of the next release.


kara (changes requested):
Can you add the reason why this commit is being reverted to the commit message?

kara (approved):
LGTM, thanks!

PR #36654, opened 2020-04-16
feat(router): Add RouteConfigReady

pgammans:
* Add new event to provide notification of a route being added and  ready for use.
* Improve DRY-ness of setting of _loadedConfig

## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] angular.io application / infrastructure changes
- [ ] Other... Please describe:

## What is the current behavior?
When routes are loaded a RouteConfigLoadStart/End get fired. The RouteConfigLoadEnd event occur at the end of loading not when the route is fully added to the router and 

This has been discussed in #16282,  #14036

Issue Number:  <N/A>

## What is the new behavior?

This now add a new Event as per @jasonaden  request on #16308 but also moves the setting of the internal _loadedConfig to inside the loader so the event can be sent from one place.

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

PR #26557, opened 2018-10-18
fix(router): unsubscribe preload if the route has been preloaded

Wykks:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [X] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [X] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] angular.io application / infrastructure changes
- [ ] Other... Please describe:

## What is the current behavior?
When a lazyloaded route is loaded with a user interaction for example, the observable returned by `preload` from a `PreloadingStrategy` is not unsubscribed. Which do not stop the preload strategy from loading the module again.

This basic preload strategy load the module twice if the user navigate to the route before 5 seconds:
```
@Injectable({
  providedIn: 'root'
})
export class PreloadAfterAWhile implements PreloadingStrategy {
  preload(route: Route, load: () => Observable<any>) {
    return of(true).pipe(delay(5000), switchMap(load));
  }
}

```
Reproduced here: https://stackblitz.com/edit/angular-preload-issue
Step to reproduce:

1. Click on customers
2. Look at the console "CustomersModule Loaded!"
3. Wait at least 5 seconds
4. The CustomersModule is loaded again, because the delay observable from the preload strategy has not been canceled.

## What is the new behavior?
Cancel the preload observable if the route is loaded.

## Does this PR introduce a breaking change?

- [ ] Yes
- [X] No

## Other information

Also, the observable returned by `load()` is not cold, the loading start when the function is called, even if the observable is not subscribed. Is this intended ?
https://github.com/angular/angular/blob/c9488b543251980c96c7384f69042d958bde1bae/packages/router/src/router_config_loader.ts#L47-L60
`this.loader.load(loadChildren)` should be wrapped inside an observable.

Issue #22842, opened 2018-03-17
`RouteConfigLoadEnd` event emitted multiple times


peterbakonyi05:
## I'm submitting a...
<pre><code>
[X] Bug report
</code></pre>

## Current behavior
When lazy-loaded modules are being preloaded (using `PreloadAllModules` strategy or a custom strategy) and a navigation starts to a module that is currently being preloaded, `RouteConfigLoadEnd` event is emitted twice because `RouterConfigLoader.load` is called twice and it calls `onLoadEndListener` twice ([code here](https://github.com/angular/angular/blob/5.2.8/packages/router/src/router_config_loader.ts#L38)).

## Expected behavior
`Router` emits `RouteConfigLoadEnd` event only once for a lazy-loaded route.

## Minimal reproduction of the problem with instructions
Use `PreloadAllModules` strategy and after `RouteConfigLoadStart` even is emitted, but before `RotueConfigLoadEnd` happens, call `router.navigate` or `router.navigateByUrl` to the currently pre-loaded module. If you enable tracing, you will see two `RouteConfigLoadEnd` event is being logged in the console.

## What is the motivation / use case for changing the behavior?
When listening to `Router` events, it is not expected that this event is emitted multiple times for the same route. In our case, we need to create a lookup for route config and without having an extra safe-check in our code-base, items will be added multiple times.

## Environment

<pre><code>
Angular version: 5.2.8
Chrome (desktop) version 65
</code></pre>

danielcrisp:
I'm experiencing this too.

I'm finding that child modules of lazy-loaded modules are instantiated multiple times (once for each `RouteConfigLoadEnd` event) as a result, and this is leading to odd behaviour.

atscott:
Confirmed with repro here: https://angular-preload-all-qmhasc.stackblitz.io/A
https://stackblitz.com/edit/angular-preload-all-qmhasc?file=app%2Fapp.component.ts

1. Navigate to /A route (preloading for lazy loaded module starts and is returned in a 5 second timeout)
2. Navigate to /B before the 5 second timeout
-=> see that `RouteConfigLoadEnd` is fired twice.

atscott:
Fixed by https://github.com/angular/angular/commit/4906bf6a50f307fb69080d276c3b91a60ba7fbae

PR #36760, opened 2020-04-22
Fix module loading to handle interaction between route preload strategy  and route navigation  

pgammans:
## PR Checklist

## PR Type

- [x] Bugfix
- [x] Feature

## What is the current behaviour?

This PR addresses the issues found in the issues  #22842, and pr #26557

It also replaces the PR for #26557 which has a number of issues
 * By forcing the subscription to close if a module has a lazy sub-module then the these are passed to the PreloadingStrategy class as the observable has already terminated.
 * It doesn't actually prevent the load as the fetch may overlap with the preload fetch. Ie both the pre-loader and navigation can start a load both will finish .

Issue Number:  #26557, #22842

## What is the new behaviour?

By Prevent loading a route module twice, we prevent the sending of multiple  RouteConfigLoadEnd events, and the duplicate console log as show in the stackbiz example on #26557 

We also add a new feature to SpyNgModuleFactoryLoader to enable us to emulate slow loading modules, this is required for testing.

- [ ] Yes
- [x] No

## Other information / Questions

* I have a version of this based on #36654
* I need to add test to enable the and support error in loading the module. Ie if the initial load request fails, the stored load-observable need to be removed so a new one is created.
* I need suggestions for the name of the private var attached to the route to store the loading observable currently named '_loader'. It attached to the route as both the preloader and router create the own instance of the RouterConfigLoader

### DIFF
--- a/packages/router/src/config.ts
+++ b/packages/router/src/config.ts
@@ -486,6 +486,11 @@ export interface Route {
    * @internal
    */
   _loadedConfig?: LoadedRouterConfig;
+  /**
+   * Filled for routes with `loadChildren` during load
+   * @internal
+   */
+  _loader$?: Observable<LoadedRouterConfig>;
 }
 
 export class LoadedRouterConfig {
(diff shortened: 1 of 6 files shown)


atscott (changes requested):
The implementation is simple enough and looks relatively good to me. The tests need some work though so I'll re-review the implementation once those are redone. Please see the comments.

atscott:
Closing due to lack of activity. If you still want to contribute a change, please open a new PR and I'd be happy to review it. Thanks!

PR #36792, opened 2020-04-24
feat(localize): support merging multiple translation files

petebacondarwin:
Previously only one translation file per locale could be loaded.
Now the user can specify multiple files per locale, and the translations
from each of these files will be merged together by message id.

---

Note that the CLI uses the lower level `TranslationParser` classes directly, so will not inherit this new behaviour automatically - cc @clydin.

### DIFF
--- a/packages/localize/src/tools/src/translate/main.ts
+++ b/packages/localize/src/tools/src/translate/main.ts
@@ -84,7 +84,7 @@ if (require.main === module) {
   const translationFilePaths: string[] = options['t'];
   const outputPathFn = getOutputPathFn(options['o']);
   const diagnostics = new Diagnostics();
-  const missingTranslation: MissingTranslationStrategy = options['m'];
+  const missingTranslation: DiagnosticHandlingStrategy = options['m'];
   const sourceLocale: string|undefined = options['l'];
   const translationFileLocales: string[] = options['target-locales'] || [];
 
@@ -134,7 +134,7 @@ export interface TranslateFilesOptions {
   /**
    * How to handle missing translations.
    */
-  missingTranslation: MissingTranslationStrategy;
+  missingTranslation: DiagnosticHandlingStrategy;
   /**
    * The locale of the source files.
    * If this is provided then a copy of the application will be created with no translation but just
--- a/packages/localize/src/tools/src/translate/main.ts
+++ b/packages/localize/src/tools/src/translate/main.ts
@@ -10,8 +10,7 @@ import * as glob from 'glob';
 import {resolve} from 'path';
 import * as yargs from 'yargs';
 
-import {Diagnostics} from '../diagnostics';
-import {MissingTranslationStrategy} from '../source_file_utils';
+import {DiagnosticHandlingStrategy, Diagnostics} from '../diagnostics';
 import {AssetTranslationHandler} from './asset_files/asset_translation_handler';
 import {getOutputPathFn, OutputPathFn} from './output_path';
 import {SourceFileTranslationHandler} from './source_files/source_file_translation_handler';
@@ -51,7 +50,10 @@ if (require.main === module) {
             array: true,
             describe:
                 'A list of paths to the translation files to load, either absolute or relative to the current working directory.\n' +
-                'E.g. "-t src/locale/messages.en.xlf src/locale/messages.fr.xlf src/locale/messages.de.xlf".',
+                'E.g. `-t src/locale/messages.en.xlf src/locale/messages.fr.xlf src/locale/messages.de.xlf`.\n' +
+                'If you want to merge multiple translation files for each locale, then provide the list of files in an array.\n' +
+                'Note that the arrays must be in double quotes if you include any whitespace within the array.\n' +
+                'E.g. `-t "[src/locale/messages.en.xlf, src/locale/messages-2.en.xlf]" [src/locale/messages.fr.xlf,src/locale/messages-2.fr.xlf]`',
           })
 
           .option('target-locales', {
@@ -74,6 +76,14 @@ if (require.main === module) {
             choices: ['error', 'warning', 'ignore'],
             default: 'warning',
           })
+
+          .option('d', {
+            alias: 'duplicateTranslation',
+            describe: 'How to handle duplicate translations.',
+            choices: ['error', 'warning', 'ignore'],
+            default: 'warning',
+          })
+
           .strict()
           .help()
           .parse(args);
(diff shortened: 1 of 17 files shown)


packages/localize/src/tools/src/translate/translation_files/translation_loader.ts
```diff
@@ -31,46 +31,87 @@ export class TranslationLoader {
...
         }
-
-        return {locale, translations, diagnostics};
       }
-      throw new Error(
-          `There is no "TranslationParser" that can parse this translation file: ${filePath}.`);
-    });
+      Object.keys(nextBundle.translations).forEach(messageId => {
+        if (bundle.translations[messageId] !== undefined) {
+          this.diagnostics?.error(
```
AndrewKushnir, line 108: @petebacondarwin I'm wondering if this behavior should be configurable (a-la https://angular.io/guide/i18n#report-missing-translations), but default to "error". An example I can think of is if you want to use a 3rd party component in your app and this component also includes translations, but you want to change some of the messages. In this case it might be convenient to provide component's translation file and your own translation file that overrides some of the messages.

What do you think about it?

AndrewKushnir (approved):
LGTM, thanks @petebacondarwin 👍 Just left one question/proposal to consider.

packages/localize/src/tools/src/translate/translation_files/translation_loader.ts
petebacondarwin: Actually that sounds like a very common scenario... Perhaps we should default to "warn"?

petebacondarwin: I added options to change whether this errors or not and defaulted it to warning.
(also took the opportunity to add a refactor commit to the start of the PR)

packages/localize/src/tools/src/translate/translation_files/translation_loader.ts
```diff
@@ -16,6 +17,7 @@ import {TranslationParser} from './translation_parsers/translation_parser';
 export class TranslationLoader {
   constructor(
       private translationParsers: TranslationParser<any>[],
+      private duplicateTranslation: DiagnosticHandlingStrategy = 'warning',
```
AndrewKushnir, line 20: nit: I think `TranslationLoader` should not make a decision on how to handle duplicate translations if nothing is provided. I'd propose to set default value in a different place, where we call `new TranslationLoader`.

packages/localize/src/tools/src/translate/translation_files/translation_loader.ts
```diff
@@ -31,46 +33,88 @@ export class TranslationLoader {
...
+      if (parsedLocale !== undefined && providedLocale !== undefined &&
+          parsedLocale !== providedLocale) {
+        diagnostics.warn(
+            `The provided locale "${providedLocale}" does not match the target locale "${
+                parsedLocale}" found in the translation file "${filePath}".`);
+      }
+
+      // If we were passed a diagnostics object then copy the messages over to it.
+      if (this.diagnostics) {
+        this.diagnostics.messages.push(...diagnostics.messages);
```
AndrewKushnir, line 82: The `messages` is marked as `readonly` on the `Diagnostics` class (which doesn't prevent new items from being added) so I'm wondering if adding a method to `Diagnostics` class would be better (and also making `messages` a private field)?

packages/localize/src/tools/src/translate/main.ts
```diff
@@ -117,7 +129,7 @@ export interface TranslateFilesOptions {
    * An array of paths to the translation files to load, either absolute or relative to the current
    * working directory.
    */
-  translationFilePaths: string[];
+  translationFilePaths: (string|string[])[];
```
AndrewKushnir, line 132: I think it'd be great to add a note regarding duplicate translations rule, i.e. if we have 2 bundles that contain translations with the same id, what'd happen in this case (translation from the first bundle would be taken or translation from the second bundle will override the first one).

packages/localize/src/tools/src/translate/translation_files/translation_loader.ts
```diff
@@ -31,46 +33,88 @@ export class TranslationLoader {
...
         }
-
-        return {locale, translations, diagnostics};
       }
-      throw new Error(
-          `There is no "TranslationParser" that can parse this translation file: ${filePath}.`);
-    });
+      Object.keys(nextBundle.translations).forEach(messageId => {
+        if (bundle.translations[messageId] !== undefined) {
+          this.diagnostics?.add(
```
AndrewKushnir, line 110: Based on this logic, if we have multiple bundles and same translation is present in multiple bundles, the first translation wins. Intuitively, this problem is similar to merging objects, for example `{...a, ...b}`, so `b` overrides keys also present in `a`, i.e. the last object has higher priority. Pete, WDYT?

AndrewKushnir (commented):
Thanks for the updates Pete! I've left a few comments, could you please have a look when you get a chance?

packages/localize/src/tools/src/translate/translation_files/translation_loader.ts
petebacondarwin: Yeah, this is how it was before this PR too. I'll see what I can do.

petebacondarwin: I added a `merge()` method but I would like to keep the `messages` property public since this is how one reads the messages out of the `Diagnostics` object. Adding a getter property or method to hide this does not provide much better encapsulation, unless we actually clone the `messages` array when accessing it, which seems over the top. WDYT?

packages/localize/src/tools/src/translate/translation_files/translation_loader.ts
petebacondarwin: First translation wins is the approach I am aiming for. The conceptual model being that you provide the application translation file first and then "fallback" library translation files after:

```
[ 'app.en.xlf', 'lib1.en.xlf', 'lib2.en.xlf]
```

I will improve the commit message and documentation comments with this.

packages/localize/src/tools/src/translate/main.ts
petebacondarwin: Will do!

mhevery (approved):
Integration tests LGTM

AndrewKushnir (approved):
LGTM, thanks Pete (also special thanks for the docs!) 👍

PR #36797, opened 2020-04-24
build: enable platform-browser tests on Saucelabs

crisbeto:
Enables some passing `platform-browser` tests on Saucelabs. The reason they were disabled was an error log which doesn't actually fail the test run and has been there for a long time.

**Note:** these tests usually don't run against PRs, but I've verified that they pass by:
1. Running them against Chrome and IE on my local machine.
2. Enabling them for PRs temporarily (see the first commit).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -879,6 +879,13 @@ workflows:
           requires:
             - test_ivy_aot
 
+      ##################################
+      # Temporary
+      ##################################
+      - saucelabs_ivy:
+          requires:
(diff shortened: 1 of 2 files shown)


gkalpak (approved):

mhevery (approved):

AndrewKushnir:
FYI, I re-labelled this PR to `master-only`, since the original change was not present on patch branch, so the merge failed with `nothing to commit, working tree clean` error.

PR #36798, opened 2020-04-24
Revert dev infra benchmarking

wagnermaciel:
Reverts #36434

## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] angular.io application / infrastructure changes
- [ ] Other... Please describe:

## What is the current behavior?
g3sync will break because `//dev-infra` is not synced into google3 while `//modules` does.

Issue Number: N/A

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

PR #36800, opened 2020-04-24
Redo dev infra benchmarking

wagnermaciel:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

## Context

These changes were already submitted once in PR #36721. They were reverted by PR #36798 when @AndrewKushnir noticed that the updated files in the modules folder refer to the dev-infra package that we do not sync in g3, which would likely have caused build failures. I am not sure why NG Bot indicated (in “g3 status”) that the first PR wouldn't affect g3, when in fact it would have.

The difference between this PR and the previous one is that it has been g3synced using CL #309492386's copy.bara.sky configuration. This change syncs `dev-infra/benchmark/driver-utilities/**` into g3.

The commit messages (while they are confusing) reflect the correct history of these changes. I am OK with changing them if we would prefer to have clarity over historical accuracy in this case.


tools/gulp-tasks/format.js
```diff
@@ -0,0 +1,161 @@
+/**
```
jelbourn, line 1: This file looks like a rebasing artifact

IgorMinar (changes requested):
Hey Wagner!

What's going on with this PR? Can you please provide a bit more context? Thanks

Side-note: the current commit messages just state that these are reverts of reverts which is a bit confusing. But let's sort out what you are trying to do here before we worry about the commit messages.

wagnermaciel:
Hi Igor,

For context, these changes were already submitted once in PR #36721. They were reverted by PR #36798 when @AndrewKushnir noticed that the updated files in the modules folder refer to the dev-infra package that we do not sync in g3, which would likely have caused build failures. I am not sure why NG Bot indicated (in “g3 status”) that the first PR wouldn't affect g3, when in fact it would have.

The difference between this PR and the previous one is that it has been g3synced using CL [#309492386](https://critique-ng.corp.google.com/cl/310379663)'s copy.bara.sky configuration.

The commit messages (while they are confusing) reflect the correct history of these changes. I am OK with changing them if we would prefer to have clarity over historical accuracy in this case.

tools/gulp-tasks/format.js
wagnermaciel: Fixed. Seems like that file no longer exists in angular/angular so I just deleted it

dev-infra/BUILD.bazel
```diff
@@ -36,10 +36,30 @@ genrule(
 
 pkg_npm(
     name = "npm_package",
+    srcs = [
+        "BUILD.bazel",
+        "//dev-infra/benchmark:files",
+    ],
+    substitutions = {
+        # angular/angular should not consume it's own packages, so we use
+        # substitutions to replace these in the deployed version of dev-infra.
```
josephperrott, line 45: s/deployed version/published version

dev-infra/benchmark/driver-utilities/BUILD.bazel
```diff
@@ -1,15 +1,16 @@
-load("//tools:defaults.bzl", "ts_library")
-
 package(default_visibility = ["//visibility:public"])
 
+load("@npm_bazel_typescript//:index.bzl", "ts_library")
+
 ts_library(
```
josephperrott, line 5: Should this include `testonly = True`?

dev-infra/tmpl-package.json
```diff
@@ -8,19 +8,47 @@
...
     "shelljs": "<from-root>",
     "typescript": "<from-root>",
     "yaml": "<from-root>",
     "yargs": "<from-root>",
-    "tslib": "<from-root>"
+    "tslib": "<from-root>",
+    "zone.js": "^0.10.3"
+  },
+  "bazelWorkspaces": {
+    "npm_dev_infra": {
```
josephperrott, line 49: This workspace name should probably note that it is angular and private?  maybe `npm_angular_dev_infra_private`?  I know its a lot, but I we need to make sure we note its origin and usage when its read in sources where its used.

dev-infra/benchmark/driver-utilities/BUILD.bazel
wagnermaciel: Making `//dev-infra/benchmark/driver-utilities:driver-utilities` testonly breaks because dev-infra's `pkg_npm` rule `//dev-infra:npm_package` is a non-test target and non-test targets can't depend on testonly targets.

dev-infra/BUILD.bazel
```diff
@@ -36,10 +36,30 @@ genrule(
...
+    srcs = [
+        "BUILD.bazel",
+        "//dev-infra/benchmark:files",
+    ],
+    substitutions = {
+        # angular/angular should not consume it's own packages, so we use
+        # substitutions to replace these in the published version of dev-infra.
+        "//dev-infra/": "@npm_angular_dev_infra_private//",
+        "//packages/benchpress": "@npm//@angular/benchpress",
+        "//packages/bazel/src:esm5.bzl": "@npm_angular_bazel//src:esm5.bzl",
```
devversion, line 48: Should we make this more generic so that generally importing from `//packages/bazel` works?

<sup>Motivation for this is that it's easy to forget about these replacements</sup>

dev-infra/benchmark/component_benchmark/component_benchmark.bzl
```diff
@@ -1,17 +1,19 @@
...
-load("//modules/benchmarks:benchmark_test.bzl", "benchmark_test")
+load("//dev-infra/benchmark/ng_rollup_bundle:ng_rollup_bundle.bzl", "ng_rollup_bundle")
+load("//tools:defaults.bzl", "ng_module")
+load("@npm_bazel_typescript//:index.bzl", "ts_devserver", "ts_library")
+load(":benchmark_test.bzl", "benchmark_test")
 
 def copy_default_file(origin, destination):
     """
-    Copies a file from tools/components/defaults to the destination.
+    Copies a file from /defaults to the destination.
```
devversion, line 8: make this a relative path? right now it's unclear what `/` refers to.

dev-infra/benchmark/driver-utilities/BUILD.bazel
```diff
@@ -1,15 +1,16 @@
...
-    name = "e2e_util",
-    testonly = 1,
+    name = "driver-utilities",
     srcs = glob(["*.ts"]),
+    module_name = "@angular/dev-infra/benchmark/driver-utilities",
```
devversion, line 8: Should this be `@angular/dev-infra-private`? as that is the name in the NPM package?

modules/benchmarks/tsconfig.json
```diff
@@ -14,7 +14,7 @@
       "rxjs/*": ["../../node_modules/rxjs/*"],
       "@angular/*": ["../../packages/*"],
       "zone.js/*": ["../../packages/zone.js/*"],
-      "e2e_util/*": ["../e2e_util/*"]
+      "dev_infra/*": ["../../dev-infra/*"]
```
devversion, line 17: do you recall why we do this? Unless I miss something, we import things from `dev-infra` through relative paths anyway?

tools/defaults.bzl
```diff
@@ -245,7 +245,7 @@ def karma_web_test_suite(name, **kwargs):
         runtime_deps = runtime_deps,
         bootstrap = bootstrap,
         deps = deps,
-        browsers = ["//tools/browsers:chromium"],
+        browsers = ["//dev-infra/benchmark/browsers:chromium"],
```
devversion, line 248: I think this actually unveils that the browser build target should not be in the `benchmark` folder. Should we put that into a dedicated folder like `dev-infra/browsers`? cc. @josephperrott for thoughts on naming.

devversion (approved):
LGTM. A few minor comments from my side.

modules/benchmarks/tsconfig.json
wagnermaciel: No clue. Should I remove it or try using just `"dev_infra/"` instead of the relative path?

devversion: I'd propose removing.

kyliau (approved):
LGTM for language service

dev-infra/BUILD.bazel
josephperrott: Agreed, I think that makes the most sense for this case.

dev-infra/tmpl-package.json
josephperrott: Just wanted to highlight this one again in case it fell through the cracks?

tools/defaults.bzl
josephperrott: Agreed, I think it should be "top level" of `//dev-infra`. I think that `//dev-infra/browsers` makes sense unless we are worried that namespace would be needed for something else.

mhevery (approved):
LGTM for zone.js

IgorMinar (approved):
LGTM!

But OMG, I can't believe that this change is still not in. I hope that this time around it will stick.

wagnermaciel:
This has a passing presubmit
https://tap.corp.google.com/ui#id=OCL:312733720:BASE:314181500:1591040024284:2b3c7ff4

The status isn't updating due to a rate limit error. This should be good to merge.

wagnermaciel:
Currently rebasing to try and get around GitHub's status limit. Code should still be safe to merge, I am just trying to update the status here.

josephperrott (approved):
LGTM

PR #36802, opened 2020-04-24
build: create macro for transpiling javascript file to es5

josephperrott:
For testing on IE, shims must be served in es5.  Because the shims
served in these tests come from node_modules and are not part of
the angular/angular source code, they must be transpiled to es5 if
they are published as es6. This macro allows for a uniform method
for running this transpilation during the build process.

### DIFF
--- a/tools/utils.bzl
+++ b/tools/utils.bzl
@@ -1,0 +1,47 @@
+"""Simple utility bazel macros for convenience usage."""
+
+load("@npm//typescript:index.bzl", "tsc")
+
+def transpile_js_to_es5(name, js_file):
+    """Transpiles a provided javascript target to es5.
+
+    For testing on IE, shims must be served in es5, this macro can be used to
+    transpile es2015 JS shims to es5 for usage in IE testing.
+
+    Example usage:
+
+    transpile_js_to_es5(
+        name = "my-file",
+        js_file = "@npm//some_package/shim_files/es6_shim_file.js",
+    )
+
+    filegroup(
+        name = "some_shims_for_tests",
+        testonly = True,
+        srcs = [
+            ":my-file",
+            ...
+        ]
+    )
+    """
+    tsc(
+        name = name,
+        outs = [
+            "%s.js" % name,
+        ],
+        args = [
+            # Allow JS files to be used for transpiling
+            "--allowJs",
+            # Skip lib check as pure local javascript transpiling should be done
+            "--skipLibCheck",
+            # Transpile to ES5
+            "--target ES5",
+            # Output the transpiled file to the location provided by the name
+            "--outFile $(execpath :%s.js)" % name,
+            # Transpile the provided js_file
+            "$(execpath %s)" % js_file,
+        ],
+        data = [
+            js_file,
+        ],
+    )


crisbeto (approved):
LGTM

tools/defaults.bzl
```diff
@@ -476,3 +477,22 @@ def ts_api_guardian_test_npm_package(**kwargs):
         ],
         **kwargs
     )
+
+def transpile_js_to_es5(name, js_file):
```
devversion, line 481: Add a comment on where/how this is intended to be used?

package.json
```diff
@@ -45,8 +45,10 @@
     "@angular-devkit/build-optimizer": "0.900.3",
     "@angular-devkit/core": "9.0.3",
     "@angular-devkit/schematics": "9.0.3",
+    "@babel/cli": "^7.8.4",
```
devversion, line 48: Did you consider just running `tsc` instead of babel? We've been using that for downleveling in components too (with the old build system)

josephperrott: Changed to using `tsc`. When talking with @crisbeto about it we talked about `babel` while we were figuring it out so I went with that, but `tsc` makes more sense.

tools/defaults.bzl
devversion: Would it make sense to move this into a separate file since it's not really a default (similar to what we mentioned in the components repo). 

<sup>I think it's okay that this needs to be imported separately. defaults.bzl doesn't seem right and rather complicates the file</sup>

devversion (approved):
LGTM. Added a minor comment. Didn't notice that before.

tools/defaults.bzl
josephperrott: Is this something that makes sense to actually make a little more extensible by allowing the target JS version to be chosen and then have it in `//dev-infra`? 

```
def transpile_js_to_es5(name, js_file, target = "ES5"):
...
```

I might be overthinking it, or trying to create something no one else will ever need.

devversion: Definitely sounds useful, but might be best to have in `@bazel/typescript` IMO. 

I can't think of any other common use cases in Bazel where this is needed. Most of the time, the devmode output (which is easily retrievable for ts_library targets) should be ES5. if it's not, then it's not surprise that apps don't work in IE11.

josephperrott: @gregmagolan @alexeagle Thoughts? Does this sounds like something that could be made available by `@bazel/typescript`?

gregmagolan: This could be useful for other @bazel/typescript users if it was made generic so that it could transpile to a specified language level and/or a specified module format. One thing that doesn't port over to generic case is the load statement as the macro assumes `tsc` is loaded from the `@npm` repository. We are, however, planning to make loading of `ts_library` possible from `load("@npm//typescript:index.bzl", "ts_library")` as opposed to loading it from the less ergonomic `@npm_bazel_typescript` repository. Once we have this it would be more ergonomic to add it to `@bazel/typescript` as it could be loaded from `load("@npm//typescript:index.bzl", "transpile_js")` or something of that nature.

gregmagolan: I think for now landing in angular/angular is probably simplest as the `@bazel/typescript` version would have to take `tsc` as a parameter so it doesn't assume where it is loaded from

josephperrott: Okay, with that being the case, I will move it out of `defaults.bzl` but not into `//dev-infra`

Thanks @gregmagolan

devversion (approved):

PR #36805, opened 2020-04-25
refactor(forms): remove unnecessary `!` operators from validators

danranVm:
When we added the strict null checks, the lexer had some `!` operators added to prevent the compilation from failing.

See angular#24571

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/forms/src/directives/validators.ts
+++ b/packages/forms/src/directives/validators.ts
@@ -158,10 +158,8 @@ export const CHECKBOX_REQUIRED_VALIDATOR: StaticProvider = {
   host: {'[attr.required]': 'required ? "" : null'}
 })
 export class RequiredValidator implements Validator {
-  // TODO(issue/24571): remove '!'.
-  private _required!: boolean;
-  // TODO(issue/24571): remove '!'.
-  private _onChange!: () => void;
+  private _required = false;
+  private _onChange?: () => void;
 
   /**
    * @description
@@ -274,10 +272,8 @@ export const EMAIL_VALIDATOR: any = {
   providers: [EMAIL_VALIDATOR]
 })
 export class EmailValidator implements Validator {
-  // TODO(issue/24571): remove '!'.
-  private _enabled!: boolean;
-  // TODO(issue/24571): remove '!'.
-  private _onChange!: () => void;
+  private _enabled = false;
+  private _onChange?: () => void;
 
   /**
    * @description
@@ -368,17 +364,15 @@ export const MIN_LENGTH_VALIDATOR: any = {
   host: {'[attr.minlength]': 'minlength ? minlength : null'}
 })
 export class MinLengthValidator implements Validator, OnChanges {
-  // TODO(issue/24571): remove '!'.
-  private _validator!: ValidatorFn;
-  // TODO(issue/24571): remove '!'.
-  private _onChange!: () => void;
+  private _validator: ValidatorFn = Validators.nullValidator;
+  private _onChange?: () => void;
 
   /**
    * @description
    * Tracks changes to the the minimum length bound to this directive.
    */
-  // TODO(issue/24571): remove '!'.
-  @Input() minlength!: string|number;
+  @Input()
+  minlength!: string|number;  // This input is always defined, since the name matches selector.
 
   /**
    * @description
@@ -456,17 +450,15 @@ export const MAX_LENGTH_VALIDATOR: any = {
   host: {'[attr.maxlength]': 'maxlength ? maxlength : null'}
 })
 export class MaxLengthValidator implements Validator, OnChanges {
-  // TODO(issue/24571): remove '!'.
-  private _validator!: ValidatorFn;
-  // TODO(issue/24571): remove '!'.
-  private _onChange!: () => void;
+  private _validator: ValidatorFn = Validators.nullValidator;
+  private _onChange?: () => void;
 
   /**
    * @description
    * Tracks changes to the the maximum length bound to this directive.
    */
-  // TODO(issue/24571): remove '!'.
-  @Input() maxlength!: string|number;
+  @Input()
+  maxlength!: string|number;  // This input is always defined, since the name matches selector.
 
   /**
    * @description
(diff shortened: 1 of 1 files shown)


packages/forms/src/directives/validators.ts
```diff
@@ -547,16 +537,13 @@ export const PATTERN_VALIDATOR: any = {
...
   private _validator!: ValidatorFn;
-  // TODO(issue/24571): remove '!'.
-  private _onChange!: () => void;
+  private _onChange?: () => void;
 
   /**
    * @description
    * Tracks changes to the pattern bound to this directive.
    */
-  // TODO(issue/24571): remove '!'.
```
petebacondarwin, line 559: I don't think you should remove this one without first adding a comment as to why this property can have the `!` operator on it.

petebacondarwin (changes requested):
Please and you add more text to the commit message so that it passes our lint check.

packages/forms/src/directives/validators.ts
SchnWalter: The same applies to all other directives that have a property with the same name as the selector.

packages/forms/src/directives/validators.ts
```diff
@@ -547,16 +537,13 @@ export const PATTERN_VALIDATOR: any = {
   host: {'[attr.pattern]': 'pattern ? pattern : null'}
 })
 export class PatternValidator implements Validator, OnChanges {
-  // TODO(issue/24571): remove '!'.
```
petebacondarwin, line 550: I am not convinced that we should leave this `!` in place either.

packages/forms/src/directives/validators.ts
danranVm: I thought about it for a while, and then decided to give it a default value to avoid this situation.

```html
<input ngModel="value" pattern />
```

packages/forms/src/directives/validators.ts
danranVm: I set a default value for it and then removed `!`.

packages/forms/src/directives/validators.ts
```diff
@@ -593,6 +584,8 @@ export class PatternValidator implements Validator, OnChanges {
   }
 
   private _createValidator(): void {
-    this._validator = Validators.pattern(this.pattern);
+    if (this.pattern) {
```
petebacondarwin, line 587: Can `""` be a valid pattern? Perhaps you should be doing this instead here?

```suggestion
    if (this.pattern !== undefined) {
```

danranVm: test is failing.
it seems that I need to add a test for this situation ?

packages/forms/src/directives/validators.ts
```diff
@@ -502,16 +494,18 @@ export class MaxLengthValidator implements Validator, OnChanges {
   }
 
   private _createValidator(): void {
-    this._validator = Validators.maxLength(
-        typeof this.maxlength === 'number' ? this.maxlength : parseInt(this.maxlength, 10));
+    if (this.maxlength != null) {
```
petebacondarwin, line 497: This typing: `maxlength?: string|number;` means that `maxlength` cannot be `null`. So I would go for `this.maxlength !== undefined` to be clear.

packages/forms/src/directives/validators.ts
```diff
@@ -115,7 +115,7 @@ export interface AsyncValidator extends Validator {
  * @description
  * Provider which adds `RequiredValidator` to the `NG_VALIDATORS` multi-provider list.
  */
-export const REQUIRED_VALIDATOR: StaticProvider = {
+export const REQUIRED_VALIDATOR: ExistingProvider = {
```
AndrewKushnir, line 118: This change looks reasonable, but it's a breaking one and it can not be included into a patch branch. Even though these providers are not included into public API, they can be referenced through a deep import (which is not the best practice, but still a possibility). So we'd need to include this change into major version only. I'd propose to move this change to a separate PR to keep the changes in this PR backwards-compatible (as the "refactor" scope assumes).

AndrewKushnir (changes requested):

danranVm:
@AndrewKushnir modified according to your suggestion.

packages/forms/src/directives/validators.ts
```diff
@@ -368,16 +364,14 @@ export const MIN_LENGTH_VALIDATOR: any = {
...
-  private _validator!: ValidatorFn;
-  // TODO(issue/24571): remove '!'.
-  private _onChange!: () => void;
+  private _validator: ValidatorFn = Validators.nullValidator;
+  private _onChange?: () => void;
 
   /**
    * @description
    * Tracks changes to the the minimum length bound to this directive.
+   * See selectors field in the Directive decorator, a property with the same name as the selector must exist.
```
AndrewKushnir, line 373: Thanks for adding the comment! I think we should move it below to make sure it doesn't appear on the angular.io, [see here](https://angular.io/api/forms/PatternValidator) (since it's an internal implementation note). For example:
```
  @Input() minlength!: string|number;  // This input is always defined, since the name matches directive selector.
```
(the same applies to other cases below)

AndrewKushnir (changes requested):
Thanks for additional changes @danranVm 👍 

I've left a note on the comment that you added (I think we should move it). I've also updated PR title and description to reflect the latest changes (that you removed the providers type changes), used the title and description from the commit message.

It looks like the `lint` CI job is failing, could you please run the command [printed in the CI job output](https://app.circleci.com/pipelines/github/angular/angular/16502/workflows/754fc1d9-4620-47df-ae78-67494ef1595e/jobs/730424/steps)?

FYI, I've created a separate ticket https://github.com/angular/angular/issues/37493 to look into directive logic unification in the future (noticed that while doing code review).

Thank you.

AndrewKushnir (approved):
Thanks for additional updates @danranVm, the changes LGTM 👍

petebacondarwin (approved):

petebacondarwin:
@AndrewKushnir - can you deal with the presubmit, please?

AndrewKushnir:
[Presubmit](http://test/OCL:321174597:BASE:321174620:1594744589420:52e143d2)

PR #36807, opened 2020-04-25
fix(common): canceled JSONP requests won't throw error with missing callback function

### LINKED ISSUE #34818: Canceling an in-progress JSONP request causes "callback undefined" console error
# 🐞 bug report

### Affected Package
@angular/common/http

### Is this a regression?
No

### Description

It looks like HttpClientJsonpModule deletes the JSONP callback function for a request immediately upon cancellation. If the request was still in progress, an error is logged in the console when it completes because the function doesn't exist.

This issue occurs when the time between requests is usually lower than the time a request takes to complete.

## 🔬 Minimal Reproduction
https://angular-c6oeqi.stackblitz.io
https://stackblitz.com/edit/angular-c6oeqi

## 🔥 Exception or Error
<pre><code>
Uncaught ReferenceError: ng_jsonp_callback_16 is not defined
</code></pre>

## 🌍  Your Environment
_**Can be reproduced with Angular 8 (see repro)**_ ‼️

<pre><code>
Angular CLI: 6.2.9
Node: 8.17.0
OS: darwin x64
Angular: 6.1.10
... http, platform-browser, platform-browser-dynamic
... platform-server, router

@angular-devkit/architect         0.8.9
@angular-devkit/build-angular     0.8.9
@angular-devkit/build-optimizer   0.8.9
@angular-devkit/build-webpack     0.8.9
@angular-devkit/core              0.8.9
@angular-devkit/schematics        0.8.9
@angular/cdk                      6.4.7
@angular/cli                      6.2.9
@angular/language-service         7.2.15
@ngtools/webpack                  6.2.9
@schematics/angular               0.8.9
@schematics/update                0.8.9
rxjs                              6.2.2
typescript                        2.9.2
webpack                           4.16.4
</code></pre>

**Anything else relevant?**
No

martinsik:

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Fixes #34818

Unsubscribing from a JSONP request will throw console errors such as `Uncaught ReferenceError: ng_jsonp_callback_41 is not defined`. This is happening because the callback function is removed right on unsubscription but this doesn't stop the JSONP script from loading.

https://stackblitz.com/edit/angular-4zmfjr?file=src/app/app.component.ts

## What is the new behavior?

Unsubscribing from JSONP request won't remove callback handler. Handler is then removed by the handler itself:

https://github.com/angular/angular/blob/2365bb89d7a37089f596705db19b1ee5d0366b42/packages/common/http/src/jsonp.ts#L107

- [ ] Yes
- [x] No

### DIFF
--- a/packages/common/http/src/jsonp.ts
+++ b/packages/common/http/src/jsonp.ts
@@ -21,6 +21,12 @@ import {HttpErrorResponse, HttpEvent, HttpEventType, HttpResponse, HttpStatusCod
 // is shared among all applications on the page.
 let nextRequestId: number = 0;
 
+/**
+ * When a pending <script> is unsubscribed we'll move it to this document, so it won't be
+ * executed.
+ */
+let foreignDocument: Document|undefined;
+
 // Error text given when a JSONP script is injected, but doesn't invoke the callback
 // passed in its URL.
 export const JSONP_ERR_NO_CALLBACK = 'JSONP injected script did not invoke callback.';
@@ -101,21 +107,12 @@ export class JsonpClientBackend implements HttpBackend {
       // Whether the response callback has been called.
       let finished: boolean = false;
 
-      // Whether the request has been cancelled (and thus any other callbacks)
-      // should be ignored.
-      let cancelled: boolean = false;
-
       // Set the response callback in this.callbackMap (which will be the window
       // object in the browser. The script being loaded via the <script> tag will
       // eventually call this callback.
       this.callbackMap[callback] = (data?: any) => {
         // Data has been received from the JSONP script. Firstly, delete this callback.
         delete this.callbackMap[callback];
-
-        // Next, make sure the request wasn't cancelled in the meantime.
-        if (cancelled) {
-          return;
-        }
 
         // Set state to indicate data was received.
         body = data;
@@ -125,17 +122,15 @@ export class JsonpClientBackend implements HttpBackend {
       // cleanup() is a utility closure that removes the <script> from the page and
       // the response callback from the window. This logic is used in both the
       // success, error, and cancellation paths, so it's extracted out for convenience.
-      const cleanup = (removeCallback = true) => {
+      const cleanup = () => {
         // Remove the <script> tag if it's still on the page.
         if (node.parentNode) {
           node.parentNode.removeChild(node);
         }
 
-        if (removeCallback) {
-          // Remove the response callback from the callbackMap (window object in the
-          // browser).
-          delete this.callbackMap[callback];
-        }
+        // Remove the response callback from the callbackMap (window object in the
+        // browser).
+        delete this.callbackMap[callback];
       };
 
       // onLoad() is the success callback which runs after the response callback
@@ -143,11 +138,6 @@ export class JsonpClientBackend implements HttpBackend {
       // If something went wrong, onLoad() may run without the response callback
       // having been invoked.
       const onLoad = (event: Event) => {
-        // Do nothing if the request has been cancelled.
-        if (cancelled) {
-          return;
-        }
-
         // We wrap it in an extra Promise, to ensure the microtask
         // is scheduled after the loaded endpoint has executed any potential microtask itself,
         // which is not guaranteed in Internet Explorer and EdgeHTML. See issue #39496
@@ -186,10 +176,6 @@ export class JsonpClientBackend implements HttpBackend {
       // a Javascript error. It emits the error via the Observable error channel as
       // a HttpErrorResponse.
       const onError: any = (error: Error) => {
-        // If the request was already cancelled, no need to emit anything.
-        if (cancelled) {
-          return;
-        }
         cleanup();
 
         // Wrap the error in a HttpErrorResponse.
@@ -212,19 +198,24 @@ export class JsonpClientBackend implements HttpBackend {
 
       // Cancellation handler.
       return () => {
-        // Track the cancellation so event listeners won't do anything even if already scheduled.
-        cancelled = true;
-
-        // Remove the event listeners so they won't run if the events later fire.
-        node.removeEventListener('load', onLoad);
-        node.removeEventListener('error', onError);
+        if (!finished) {
+          this.removeListeners(node);
+        }
 
         // And finally, clean up the page.
-        // Unsubscription won't remove the callback handler because there's no way to stop
-        // loading <script> once it has been added to DOM.
-        cleanup(false);
+        cleanup();
       };
     });
+  }
+
+  private removeListeners(script: HTMLScriptElement): void {
+    // Issue #34818
+    // Changing <script>'s ownerDocument will prevent it from execution.
+    // https://html.spec.whatwg.org/multipage/scripting.html#execute-the-script-block
+    if (!foreignDocument) {
+      foreignDocument = (this.document.implementation as DOMImplementation).createHTMLDocument();
+    }
+    foreignDocument.adoptNode(script);
   }
 }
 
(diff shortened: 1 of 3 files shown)


packages/common/http/test/jsonp_spec.ts
```diff
@@ -18,7 +18,6 @@ function runOnlyCallback(home: any, data: Object) {
   const keys = Object.keys(home);
   expect(keys.length).toBe(1);
   const callback = home[keys[0]];
-  delete home[keys[0]];
```
martinsik, line 21: I removed this line because the handler should be removed by the handler itself:

https://github.com/angular/angular/blob/2365bb89d7a37089f596705db19b1ee5d0366b42/packages/common/http/src/jsonp.ts#L107

JoostK (commented):
It appears to be possible to prevent the script from executing by adopting the script element into another document, as described in https://html.spec.whatwg.org/multipage/scripting.html#execute-the-script-block. So instead of removing the script element as we do now, it could instead be adopted into a dummy document. I think that doing so is beneficial in terms of resources, as it avoids the need to parse and execute the script.

petebacondarwin:
@martinsik - would you have time to address the comments from @JoostK?

martinsik:
It's been so long I've completelly forgot I ever made this PR :).

@JoostK I made a small demo to test if changing <script>'s document will prevent it from exeuction and it does work https://stackblitz.com/edit/js-gs6nph?file=index.js  
I refactored the code a little bit. Seems like one state variable (`cancelled`) could be removed completely and I shouldn't even need to use `removeEventListener` because the events won't be triggered after changing the owner document.

I'm not sure why `ci/circleci: legacy-unit-tests-saucelab`  CICD process is failing but it looks like the same problem is in other PRs as well so it doesn't look like I broke it.

packages/common/http/test/jsonp_spec.ts
```diff
@@ -63,6 +62,11 @@ const SAMPLE_REQ = new HttpRequest<never>('JSONP', '/test');
       });
       document.mockError(error);
     });
+    it('allows the callback to be invoked when the request is cancelled', () => {
+      backend.handle(SAMPLE_REQ).subscribe().unsubscribe();
+      expect(Object.keys(home).length).toBe(0);
+      expect(document.mock!.ownerDocument).not.toEqual(document);
```
JoostK, line 68: Please add a comment here to indicate why this assertion is relevant. Something like:

```ts
// The script element should have been transferred to a different document to prevent it from executing.
```

packages/common/http/test/jsonp_spec.ts
```diff
@@ -63,6 +62,11 @@ const SAMPLE_REQ = new HttpRequest<never>('JSONP', '/test');
       });
       document.mockError(error);
     });
+    it('allows the callback to be invoked when the request is cancelled', () => {
+      backend.handle(SAMPLE_REQ).subscribe().unsubscribe();
+      expect(Object.keys(home).length).toBe(0);
```
JoostK, line 67: Is this sufficient to verify that the callback handler wasn't executed? Isn't this check only indicating that cleanup happened?

packages/common/http/test/jsonp_spec.ts
```diff
@@ -63,6 +62,11 @@ const SAMPLE_REQ = new HttpRequest<never>('JSONP', '/test');
       });
       document.mockError(error);
     });
+    it('allows the callback to be invoked when the request is cancelled', () => {
```
JoostK, line 65: I remember proposing this spec description and looking at it now it no longer makes sense to me (sorry!), as there is no sign of a callback anywhere in this test setup, and I think the callback should not at all be invoked, right?

So here's attempt two for a description:

```ts
    it('prevents the script from executing when the request is cancelled', () => {
```

would that be better?

JoostK (commented):
Thanks for your continued effort on this PR. It looks good to me; I'll request a review from @alxhub as code owner of fw-http so we can get this landed.

JoostK:
Btw don't worry about the Saucelabs failures; they have been _very_ flaky over the last couple of weeks and #44281 should help make it more stable.

packages/common/http/test/jsonp_spec.ts
martinsik: I updated the test to also check that the callback wasn't invoked.

martinsik:
@JoostK I updated this PR and implemented your comments.

packages/common/http/src/jsonp.ts
```diff
@@ -56,6 +56,8 @@ export class JsonpClientBackend implements HttpBackend {
    */
   private readonly resolvedPromise = Promise.resolve();
 
+  private readonly dummyDocument = this.document.implementation.createHTMLDocument();
+
```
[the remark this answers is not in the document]
martinsik: @AndrewKushnir I moved it into a separate method. I'm not sure if `removeListeners` isn't confusing because it's not really removing any listeners.

martinsik:
@AndrewKushnir I rebased the code and implemented your suggestions.

AndrewKushnir (approved):
@martinsik the change looks good 👍 

I'd like to ask @JoostK to take another look. After that we can run the necessary tests in Google's codebase and proceed with next steps if everything looks good.

Thank you.

packages/common/http/src/jsonp.ts
```diff
@@ -56,6 +56,12 @@ export class JsonpClientBackend implements HttpBackend {
    */
   private readonly resolvedPromise = Promise.resolve();
 
+  /**
+   * When a pending <script> is unsubscribed we'll move it to this document, so it won't be
+   * executed.
+   */
+  private foreignDocument!: Document;
```
JoostK, line 63: nit: I'd personally move this to become a top-level variable, as this can be safely shared between multiple instances (although unlikely to be relevant except for tests) and it minimizes better.

I'd also request not to use `!` here, as we can easily include `|undefined` in the type to make reads of this field null-safe. Or use `Document|null = null;` which produces more code but has the benefit that reads from this class stay monomorphic. But that wouldn't be a concern if it's a top-level var (nor is this performance sensitive code anyway, I'm mentioning it as a general takeaway).

tldr; I'd declare `let foreignDocument: Document|undefined;` as top-level declaration.

JoostK (approved):

packages/common/http/src/jsonp.ts
martinsik: @JoostK I moved `foreignDocument` outside of the class, but it seems like I'll need to be using `!` in `foreignDocument!.adoptNode(script);` because `this.document` is of type `any` so TypeScript can't know what type is returned from `createHTMLDocument()` .
I could use typecasting on this `(this.document.implementation as DOMImplementation).createHTMLDocument()` but that's probably not correct because `document` could use a different implementation.

JoostK: Ah, interesting. If we're relying on `createHTMLDocument` to be available, then we might as well do the cast. If we don't want to rely on that then we'd need to account for the situation where `implementaiton` or `createHTMLDocument` is not available as we'd expect, but I don't think there's a reasonable alternative we could use in that case so that seems not viable to me.

martinsik: Ok, I added typecast to `DOMImplementation`.

JoostK:
@martinsik thanks, LGTM. Could you please rebase on latest master to get CI to run?

jessicajaniuk:
This PR was merged into the repository by commit 909b21aa29c1fe8094ce3f7fb70ddf24668ba155.

PR #36825, opened 2020-04-27
ci: fix bad reference to `head` property in rebase-pr script

josephperrott:
Update rebase-pr script to properly reference a property on
the refs object using `target` rather than the previously
named `head`.


devversion (approved):

devversion (approved):

devversion (approved):

josephperrott:
Merge Assistance: It looks like the angular-robot seems to believe there is a phantom pending review.

PR #36826, opened 2020-04-28
feat(compiler): add name spans for property reads and method calls

ayazhafiz:
ASTs for property read and method calls contain information about
the entire span of the expression, including its receiver. Use cases
like a language service and compile error messages may be more
interested in the span of the direct identifier for which the
expression is constructed (i.e. an accessed property). To support this,
this commit adds a `nameSpan` property on

- `PropertyRead`s
- `SafePropertyRead`s
- `PropertyWrite`s
- `MethodCall`s
- `SafeMethodCall`s

The `nameSpan` property already existed for `BindingPipe`s.

This commit also updates usages of these expressions' `sourceSpan`s in
Ngtsc and the langauge service to use `nameSpan`s where appropriate.

## PR Checklist

## PR Type

- [x] Feature

- [ ] Yes
- [x] No

### DIFF
--- a/packages/compiler-cli/src/ngtsc/typecheck/src/expression.ts
+++ b/packages/compiler-cli/src/ngtsc/typecheck/src/expression.ts
@@ -10,7 +10,7 @@ import {AST, AstVisitor, ASTWithSource, Binary, BindingPipe, Chain, Conditional,
 import * as ts from 'typescript';
 
 import {TypeCheckingConfig} from './api';
-import {addParseSpanInfo, ignoreDiagnostics, wrapForDiagnostics} from './diagnostics';
+import {addParseSpanInfo, annotateSpan, wrapForDiagnostics} from './diagnostics';
 import {tsCastToAny} from './ts_util';
 
 export const NULL_AS_ANY =
@@ -179,9 +179,10 @@ class AstTranslator implements AstVisitor {
   visitMethodCall(ast: MethodCall): ts.Expression {
     const receiver = wrapForDiagnostics(this.translate(ast.receiver));
     const method = ts.createPropertyAccess(receiver, ast.name);
+    annotateSpan(method, ast.nameSpan);
     const args = ast.args.map(expr => this.translate(expr));
     const node = ts.createCall(method, undefined, args);
-    addParseSpanInfo(node, ast.nameSpan);
+    addParseSpanInfo(node, ast.sourceSpan);
     return node;
   }
 
@@ -215,11 +216,10 @@ class AstTranslator implements AstVisitor {
   visitPropertyWrite(ast: PropertyWrite): ts.Expression {
     const receiver = wrapForDiagnostics(this.translate(ast.receiver));
     const left = ts.createPropertyAccess(receiver, ast.name);
-    // TODO(joost): annotate `left` with the span of the property access, which is not currently
-    //  available on `ast`.
+    annotateSpan(left, ast.nameSpan);
     const right = this.translate(ast.value);
     const node = wrapForDiagnostics(ts.createBinary(left, ts.SyntaxKind.EqualsToken, right));
-    addParseSpanInfo(node, ast.nameSpan);
+    addParseSpanInfo(node, ast.sourceSpan);
     return node;
   }
 
@@ -235,18 +235,21 @@ class AstTranslator implements AstVisitor {
     if (this.config.strictSafeNavigationTypes) {
       // "a?.method(...)" becomes (null as any ? a!.method(...) : undefined)
       const method = ts.createPropertyAccess(ts.createNonNullExpression(receiver), ast.name);
+      annotateSpan(method, ast.nameSpan);
       const call = ts.createCall(method, undefined, args);
       node = ts.createParen(ts.createConditional(NULL_AS_ANY, call, UNDEFINED));
     } else if (VeSafeLhsInferenceBugDetector.veWillInferAnyFor(ast)) {
       // "a?.method(...)" becomes (a as any).method(...)
       const method = ts.createPropertyAccess(tsCastToAny(receiver), ast.name);
+      annotateSpan(method, ast.nameSpan);
       node = ts.createCall(method, undefined, args);
     } else {
       // "a?.method(...)" becomes (a!.method(...) as any)
       const method = ts.createPropertyAccess(ts.createNonNullExpression(receiver), ast.name);
+      annotateSpan(method, ast.nameSpan);
       node = tsCastToAny(ts.createCall(method, undefined, args));
     }
-    addParseSpanInfo(node, ast.nameSpan);
+    addParseSpanInfo(node, ast.sourceSpan);
     return node;
   }
 
--- a/packages/compiler-cli/src/ngtsc/typecheck/src/expression.ts
+++ b/packages/compiler-cli/src/ngtsc/typecheck/src/expression.ts
@@ -218,15 +218,18 @@ class AstTranslator implements AstVisitor {
   visitPropertyWrite(ast: PropertyWrite): ts.Expression {
     const receiver = wrapForDiagnostics(this.translate(ast.receiver));
     const left = ts.createPropertyAccess(receiver, ast.name);
+    addParseSpanInfo(left, ast.nameSpan);
     // TypeScript reports assignment errors on the entire lvalue expression. Annotate the lvalue of
     // the assignment with the sourceSpan, which includes receivers, rather than nameSpan for
     // consistency of the diagnostic location.
     // a.b.c = 1
     // ^^^^^^^^^ sourceSpan
     //     ^     nameSpan
-    addParseSpanInfo(left, ast.sourceSpan);
+    const leftWithPath = wrapForDiagnostics(left);
+    addParseSpanInfo(leftWithPath, ast.sourceSpan);
     const right = this.translate(ast.value);
-    const node = wrapForDiagnostics(ts.createBinary(left, ts.SyntaxKind.EqualsToken, right));
+    const node =
+        wrapForDiagnostics(ts.createBinary(leftWithPath, ts.SyntaxKind.EqualsToken, right));
     addParseSpanInfo(node, ast.sourceSpan);
     return node;
   }
--- a/packages/compiler-cli/src/ngtsc/typecheck/src/expression.ts
+++ b/packages/compiler-cli/src/ngtsc/typecheck/src/expression.ts
@@ -208,15 +208,23 @@ class AstTranslator implements AstVisitor {
     // This is a normal property read - convert the receiver to an expression and emit the correct
     // TypeScript expression to read the property.
     const receiver = wrapForDiagnostics(this.translate(ast.receiver));
-    const node = ts.createPropertyAccess(receiver, ast.name);
-    addParseSpanInfo(node, ast.nameSpan);
+    const name = ts.createPropertyAccess(receiver, ast.name);
+    addParseSpanInfo(name, ast.nameSpan);
+    const node = wrapForDiagnostics(name);
+    addParseSpanInfo(node, ast.sourceSpan);
     return node;
   }
 
   visitPropertyWrite(ast: PropertyWrite): ts.Expression {
     const receiver = wrapForDiagnostics(this.translate(ast.receiver));
     const left = ts.createPropertyAccess(receiver, ast.name);
-    addParseSpanInfo(left, ast.nameSpan);
+    // TypeScript reports assignment errors on the entire lvalue expression. Annotate the lvalue of
+    // the assignment with the sourceSpan, which includes receivers, rather than nameSpan for
+    // consistency of the diagnostic location.
+    // a.b.c = 1
+    // ^^^^^^^^^ sourceSpan
+    //     ^     nameSpan
+    addParseSpanInfo(left, ast.sourceSpan);
     const right = this.translate(ast.value);
     const node = wrapForDiagnostics(ts.createBinary(left, ts.SyntaxKind.EqualsToken, right));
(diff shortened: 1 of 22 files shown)


packages/compiler/src/expression_parser/parser.ts
```diff
@@ -305,9 +305,27 @@ export class _ParseAST {
     return this.peek(0);
   }
 
+  get atEOF(): boolean {
```
kyliau, line 308: please add brief documentation

packages/compiler/test/expression_parser/utils/unparser.ts
```diff
@@ -197,3 +197,33 @@ export function unparse(
...
+
+export function unparseWithSpan(
+    ast: ASTWithSource,
+    interpolationConfig: InterpolationConfig = DEFAULT_INTERPOLATION_CONFIG): UnparsedWithSpan[] {
+  const unparsed: UnparsedWithSpan[] = [];
+  const source = ast.source!;
+  const recursiveSpanUnparser = new class extends RecursiveAstVisitor {
+    private recordUnparsed(ast: any, spanKey: string, unparsedList: UnparsedWithSpan[]) {
+      const span = ast[spanKey];
+      const prefix = spanKey == 'span' ? '' : `[${spanKey}] `;
```
kyliau, line 211: prefer strict equality

packages/compiler/test/expression_parser/utils/unparser.ts
```diff
@@ -197,3 +197,33 @@ export function unparse(
...
+      const src = source.substring(span.start, span.end);
+      unparsedList.push([
+        unparse(ast, interpolationConfig),
+        prefix + src,
+      ]);
+    }
+
+    visit(ast: AST, unparsedList: UnparsedWithSpan[]) {
+      this.recordUnparsed(ast, 'span', unparsedList);
+      if ('nameSpan' in ast) {
```
kyliau, line 221: prefer `hasOwnProperty` over `in`

packages/compiler/test/render3/r3_ast_absolute_span_spec.ts
```diff
@@ -245,10 +245,18 @@ describe('expression AST absolute source spans', () => {
...
+  describe('property read', () => {
+    it('should provide absolute offsets of a property read', () => {
+      expect(humanizeExpressionSource(parse('<div>{{prop}}<div>').nodes)).toContain([
+        'prop?.safe', new AbsoluteSourceSpan(7, 11)
+      ]);
+    });
+
+    it('should provide absolute offsets of expressions in a property read', () => {
+      expect(humanizeExpressionSource(parse('<div>{{prop.safe}}<div>').nodes)).toContain([
+        'prop', new AbsoluteSourceSpan(7, 10)
```
kyliau, line 257: what does this absolute source span refer to?

packages/language-service/src/completions.ts
```diff
@@ -457,8 +457,13 @@ class ExpressionVisitor extends NullTemplateVisitor {
       const absValueOffset = ast.sourceSpan.start.offset;
       const {templateBindings} = this.info.expressionParser.parseTemplateBindings(
           templateKey, templateValue, templateUrl, absKeyOffset, absValueOffset);
-      // Find the template binding that contains the position.
-      const templateBinding = templateBindings.find(b => inSpan(this.position, b.sourceSpan));
+      // Find the nearest template binding to the position.
+      let lastBindingEnd = templateBindings.length > 0 &&
```
kyliau, line 461: prefer const over let

packages/language-service/src/expression_type.ts
```diff
@@ -444,3 +448,14 @@ export class AstType implements AstVisitor {
...
 }
+
+function refinedSpan(ast: AST): ng.Span {
+  // nameSpan is an absolute span, but the spans returned by the expression visitor are expected to
+  // be relative to the start of the expression.
+  // TODO: migrate to only using absolute spans
+  const absoluteOffset = ast.sourceSpan.start - ast.span.start;
+  if (ast instanceof PropertyRead || ast instanceof SafePropertyRead || ast instanceof MethodCall ||
+      ast instanceof SafeMethodCall || ast instanceof PropertyWrite || ast instanceof BindingPipe)
+    return offsetSpan(ast.nameSpan, -absoluteOffset);
```
kyliau, line 459: please wrap if body in curly braces

packages/language-service/src/expressions.ts
```diff
@@ -120,6 +120,17 @@ export function getExpressionSymbol(
     return new AstType(scope, templateInfo.query, {}, templateInfo.source).getType(ast);
   }
 
+  function spanFromName(ast: AST&{nameSpan: Span}): Span {
```
kyliau, line 123: isn't nameSpan of type `AbsoluteSourceSpan`?

packages/language-service/test/diagnostics_spec.ts
```diff
@@ -1004,40 +1002,40 @@ describe('diagnostics', () => {
             `Consider using the safe navigation operator (optional?.length) ` +
             `or non-null assertion operator (optional!.length).`);
     expect(category).toBe(ts.DiagnosticCategory.Suggestion);
-    expect(content.substring(start!, start! + length!)).toBe('optional.length');
+    expect(content.substring(start!, start! + length!)).toBe('length');
   });
 
   it('should report error if method is not found on non-nullable receiver', () => {
     const expressions = [
```
kyliau, line 1009: could you please rename this to `receivers` instead?

packages/language-service/test/diagnostics_spec.ts
```diff
@@ -1004,40 +1002,40 @@ describe('diagnostics', () => {
...
       expect(messageText)
           .toBe(`Identifier 'someMethod' is not defined. 'string' does not contain such a member`);
       expect(category).toBe(ts.DiagnosticCategory.Error);
-      expect(content.substring(start!, start! + length!)).toBe(expression);
+      expect(content.substring(start!, start! + length!)).toBe('someMethod()');
     }
   });
 
   it('should report error if property is not found on non-nullable receiver', () => {
     const expressions = [
```
kyliau, line 1026: ditto

kyliau (approved):

packages/compiler-cli/src/ngtsc/typecheck/src/type_check_block.ts
```diff
@@ -81,6 +81,13 @@ export function generateTypeCheckBlock(
   return fnDecl;
 }
 
+function refinedSpan(ast: AST): AbsoluteSourceSpan {
```
JoostK, line 84: I'm not fond of this method. The name span is not necessarily a "better" option to report diagnostics on, that all depends on context. Therefore, I don't think it's appropriate to define what a refined span is without that context.

packages/compiler/test/expression_parser/utils/unparser.ts
```diff
@@ -197,3 +197,33 @@ export function unparse(
     ast: AST, interpolationConfig: InterpolationConfig = DEFAULT_INTERPOLATION_CONFIG): string {
   return sharedUnparser.unparse(ast, interpolationConfig);
 }
+
+type UnparsedWithSpan = [string, string];
```
JoostK, line 201: Please add a comment here what the tuple indices represent, that's not obvious to me.

JoostK (changes requested):
Thanks for working on this. I do have some comments on the chosen source spans in the template type checker.

In general, I think our AST nodes could be cleaned up a lot if we were to have something like `Identifier`, similar to the TypeScript AST. That seems like a large project by itself though, so it's not necessary for this PR. It may also be problematic to push that forward, as tools that depend on the compiler for parsing ASTs would have to be updated. Although the compiler is not technically considered public API, such breakages are still problematic to push through.

packages/compiler-cli/src/ngtsc/typecheck/src/type_check_block.ts
```diff
@@ -1087,7 +1094,7 @@ class TcbExpressionTranslator {
 
       const expr = this.translate(ast.value);
       const result = ts.createParen(ts.createBinary(target, ts.SyntaxKind.EqualsToken, expr));
-      addParseSpanInfo(result, ast.sourceSpan);
+      addParseSpanInfo(result, ast.nameSpan);
```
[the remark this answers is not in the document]
ayazhafiz: Thanks for reviewing this, I really just kind of blew by this and did not even look at what result was. I'll pay more attention next time.

ayazhafiz:
> In general, I think our AST nodes could be cleaned up a lot if we were to have something like `Identifier`, similar to the TypeScript AST.

> such breakages are still problematic to push through.

I agree that this would be difficult to push through. IMO the benefit of this isn't worth the churn -- the expression grammar is fairly small, so exposing a common interface for identifier-like ASTs is probably enough.

packages/compiler-cli/src/ngtsc/typecheck/src/expression.ts
```diff
@@ -275,7 +278,7 @@ class AstTranslator implements AstVisitor {
       const expr = ts.createPropertyAccess(ts.createNonNullExpression(receiver), ast.name);
       node = tsCastToAny(expr);
     }
-    addParseSpanInfo(node, ast.sourceSpan);
+    addParseSpanInfo(node, ast.nameSpan);
```
JoostK, line 281:
```suggestion
    addParseSpanInfo(node, ast.sourceSpan);
```

packages/compiler-cli/src/ngtsc/typecheck/src/type_check_block.ts
```diff
@@ -1125,6 +1125,7 @@ class TcbExpressionTranslator {
       return result;
     } else if (ast instanceof MethodCall && ast.receiver instanceof ImplicitReceiver) {
       // Resolve the special `$any(expr)` syntax to insert a cast of the argument to type `any`.
+      // `$any(expr) -> expr as any`
```
JoostK, line 1128: If you want to keep this comment I would introduce two more backticks:

```suggestion
      // `$any(expr)` -> `expr as any`
```

packages/compiler-cli/src/ngtsc/typecheck/test/diagnostics_spec.ts
```diff
@@ -190,7 +190,7 @@ runInEachFileSystem(() => {
           [{type: 'pipe', name: 'Pipe', pipeName: 'pipe'}]);
 
       expect(messages).toEqual([
-        `synthetic.html(1, 28): Argument of type 'number' is not assignable to parameter of type 'string'.`,
+        `synthetic.html(1, 35): Argument of type 'number' is not assignable to parameter of type 'string'.`,
```
JoostK, line 193: This should have stayed column 28, as it's about the full argument. I'm not sure what the comment structure currently looks like, but my guess is that this can be resolved using some parenthesis. See `wrapForDiagnostics` function and its usages that adds those parens.

packages/compiler-cli/src/ngtsc/typecheck/test/diagnostics_spec.ts
```diff
@@ -329,7 +329,7 @@ runInEachFileSystem(() => {
           };
         }`);
 
-        expect(messages).toEqual([`synthetic.html(1, 26): Object is possibly 'undefined'.`]);
+        expect(messages).toEqual([`synthetic.html(1, 41): Object is possibly 'undefined'.`]);
```
JoostK, line 332: Much better :+1:

packages/compiler-cli/src/ngtsc/typecheck/test/diagnostics_spec.ts
```diff
@@ -367,6 +367,85 @@ class TestComponent {
...
+    it('reports unassignable value on property write span', () => {
+      const messages = diagnose(`<div (click)="person.name = 2"></div>`, `
+        export class TestComponent {
+          person: {
+            name: string;
+          };
+        }`);
+
+      expect(messages).toEqual(
+          [`synthetic.html(1, 22): Type '2' is not assignable to type 'string'.`]);
```
JoostK, line 446: TypeScript would report these on the whole assignment path, i.e. `person.name`, which I agree would be a better choice.

packages/compiler-cli/src/ngtsc/typecheck/src/expression.ts
ayazhafiz: 👍

packages/compiler-cli/src/ngtsc/typecheck/test/diagnostics_spec.ts
ayazhafiz: Actually I think it should be at column 9, since that's the entire expression being assigned. What do you think?

ayazhafiz: fixed

packages/compiler-cli/src/ngtsc/typecheck/test/diagnostics_spec.ts
ayazhafiz: fixed

ayazhafiz:
@JoostK there are still some minor things that could be improved, for example wrapping method calls for diagnostics and annotating that with the span including any receivers (like is done for property reads). I think this can be done in another PR though, since this one is very large as is.

JoostK (approved):
LGTM for compiler changes.

ayazhafiz:
Caretaker: could you run g3sync? (I think there may be some failures with Codelyzer)

alxhub:
[Presubmit](http://test/OCL:309997242:BASE:309997279:1588706596131:81a944da)

Issue #36796, opened 2020-04-24
Parallel build fails: Another process, with id XXXXX, is currently running ngcc


arthurspa:
# 🐞 bug report

### Affected Package

 The issue is caused by package @angular/compiler-cli/ngcc

### Is this a regression?

 The previous version in which this bug was not present was: angular 6 (I upgrade from angular 6 to 9 (6->7, 7->8, 8->9) following the steps in the https://update.angular.io/

### Description

 I have a shared node_modules/ in a Linux environment. It is used by CI and others users to build their local angular 9 app. After upgrade to angular 9 many builds fail. Re-triggering them work sometimes.

## 🔬 Minimal Reproduction

 Create a repo and issue many (I don't know, maybe 50) build commands at the same time.

## 🔥 Exception or Error
<pre><code>
Another process, with id 372274, is currently running ngcc.
Waiting up to 25s for it to finish.
Error: Timed out waiting 25s for another ngcc process, with id 372274, to complete.
(If you are sure no ngcc process is running then you should delete the lock-file at /proj/pt/tools/nodejs_modules/nodegui_dep_3_0_0/node_modules/@angular/compiler-cli/ngcc/__ngcc_lock_file__.)
    at AsyncLocker.<anonymous> (/proj/pt/tools/nodejs_modules/nodegui_dep_3_0_0/node_modules/@angular/compiler-cli/ngcc/src/locking/async_locker.js:91:31)
    at step (/proj/pt/tools/nodejs_modules/nodegui_dep_3_0_0/node_modules/tslib/tslib.js:139:27)
    at Object.next (/proj/pt/tools/nodejs_modules/nodegui_dep_3_0_0/node_modules/tslib/tslib.js:120:57)
    at fulfilled (/proj/pt/tools/nodejs_modules/nodegui_dep_3_0_0/node_modules/tslib/tslib.js:110:62)
An unhandled exception occurred: NGCC failed.
See "/tmp/ng-USvOMI/angular-errors.log" for further details.

</code></pre>

## 🌍  Your Environment

**Angular Version:**
<pre><code>
./node_modules/.bin/ng version

     _                      _                 ____ _     ___
    / \   _ __   __ _ _   _| | __ _ _ __     / ___| |   |_ _|
   / △ \ | '_ \ / _` | | | | |/ _` | '__|   | |   | |    | |
  / ___ \| | | | (_| | |_| | | (_| | |      | |___| |___ | |
[...]

PR #36838, opened 2020-04-28
ngcc: configure and update timeouts

petebacondarwin:
Discussions on https://github.com/angular/angular/issues/36796 highlighted that we ought to make the ngcc async locker timeouts longer (and more configurable). This should sort that out.

--- 
Note that there is a bit of tidying up of the codebase along the way. Some of these refactoring/fix commits should be cherry-picked to the patch branch...

### DIFF
--- a/packages/compiler-cli/ngcc/main-ngcc.ts
+++ b/packages/compiler-cli/ngcc/main-ngcc.ts
@@ -24,7 +24,7 @@ if (require.main === module) {
       process.exitCode = 0;
     } catch (e) {
       console.error(e.stack || e.message);
-      process.exit(1);
+      process.exit(typeof e.code === 'number' ? e.code : 1);
     }
   })();
 }
--- a/packages/compiler-cli/ngcc/src/packages/configuration.ts
+++ b/packages/compiler-cli/ngcc/src/packages/configuration.ts
@@ -20,7 +20,27 @@ export interface NgccProjectConfig<T = NgccPackageConfig> {
   /**
    * The packages that are configured by this project config.
    */
-  packages: {[packagePath: string]: T};
+  packages?: {[packagePath: string]: T};
+  /**
+   * Options that control how locking the process is handled.
+   */
+  locking?: ProcessLockingConfiguration;
+}
+
+/**
+ * Options that control how locking the process is handled.
+ */
+export interface ProcessLockingConfiguration {
+  /**
+   * The number of times the AsyncLocker will attempt to lock the process before failing.
+   * Defaults to 50.
+   */
+  retryAttempts?: number;
+  /**
+   * The number of milliseconds between attempts to lock the process.
+   * Defaults to 500ms.
+   * */
+  retryDelay?: number;
 }
 
 /**
@@ -126,11 +146,17 @@ export const DEFAULT_NGCC_CONFIG: NgccProjectConfig = {
       },
     },
   },
+  locking: {
+    retryDelay: 500,
+    retryAttempts: 50,
+  }
 };
 
 interface VersionedPackageConfig extends NgccPackageConfig {
   versionRange: string;
 }
+
+type ProcessedConfig = Required<NgccProjectConfig<VersionedPackageConfig[]>>;
 
 const NGCC_CONFIG_FILENAME = 'ngcc.config.js';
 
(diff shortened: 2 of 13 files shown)


packages/compiler-cli/ngcc/src/packages/configuration.ts
```diff
@@ -20,7 +20,27 @@ export interface NgccProjectConfig<T = NgccPackageConfig> {
   /**
    * The packages that are configured by this project config.
    */
-  packages: {[packagePath: string]: T};
+  packages?: {[packagePath: string]: T};
+  /**
+   * Options that control how the locking the process is handled.
```
JoostK, line 25:
```suggestion
   * Options that control how locking the process is handled.
```

packages/compiler-cli/ngcc/src/packages/configuration.ts
```diff
@@ -20,7 +20,27 @@ export interface NgccProjectConfig<T = NgccPackageConfig> {
...
+   */
+  locking?: ProcessLockingConfiguration;
+}
+
+/**
+ * Options that control how the locking the process is handled.
```
JoostK, line 31:
```suggestion
 * Options that control how locking the process is handled.
```

packages/compiler-cli/ngcc/src/packages/configuration.ts
```diff
@@ -126,12 +146,18 @@ export const DEFAULT_NGCC_CONFIG: NgccProjectConfig = {
       },
     },
   },
+  locking: {
+    retryDelay: 500,
```
JoostK, line 150: This is 500ms, not 50ms.

JoostK (changes requested):
There is something off in the ms calculations in the latest commit. 50x50ms is not 25s, but just 2.5s if my math is not mistaken. It may currently work out well enough because the interval is actually set to 500ms, not 50ms.

packages/compiler-cli/ngcc/src/packages/configuration.ts
petebacondarwin: Yes in the earlier commits I mistakenly swapped  the `retryAttempts` with the `retryDelay` values. Should be resolved in the fixup commits.

JoostK (approved):
LGTM, except for the commit message of "refactor(ngcc): change async locker timeout to 250 secs" as its numbers are incorrect.

petebacondarwin:
Thanks @JoostK - reworded the commit message.

packages/compiler-cli/ngcc/main-ngcc.ts
```diff
@@ -24,7 +24,7 @@ if (require.main === module) {
       process.exitCode = 0;
     } catch (e) {
       console.error(e.stack || e.message);
-      process.exit(1);
+      process.exit(e.code || 1);
```
gkalpak, line 27: [process.exit()][1] must be called with an integer, but `e.code` can be string for [errors thrown in Node.js][2].

[1]: https://nodejs.org/api/process.html#process_process_exit_code
[2]: https://nodejs.org/api/errors.html#errors_error_code

packages/compiler-cli/ngcc/src/packages/configuration.ts
```diff
@@ -170,21 +196,33 @@ export class NgccConfiguration {
...
   }
 
+  /**
+   * Get the configuration options for locking the ngcc process.
+   */
+  getLockingConfig(): Required<ProcessLockingConfiguration> {
+    const retryAttempts =
+        this.projectConfig.locking.retryAttempts || this.defaultConfig.locking.retryAttempts!;
+    const retryDelay =
+        this.projectConfig.locking.retryDelay || this.defaultConfig.locking.retryDelay!;
```
gkalpak, line 206: What is someone wants to set `retryAttempts` or `retryDelay` to 0? :grin:

gkalpak (changes requested):
Nice 😄

packages/compiler-cli/ngcc/main-ngcc.ts
petebacondarwin: Good point. I'll add an "is numeric" check.

packages/compiler-cli/ngcc/src/packages/configuration.ts
petebacondarwin: OK OK

gkalpak (approved):

IgorMinar (approved):
LGTM! Thanks!

AndrewKushnir:
Hi @petebacondarwin, could you please rebase this PR and resolve conflicts? Thank you.

PR #36841, opened 2020-04-28
fix(core): should run a fake top event task to prevent drainMicroTaskQueue too early.

### LINKED ISSUE #36839: zone.js does not handle promise.then as async microTask in some case
- `zone.js` version: 0.10.3
- reproduce repo: https://github.com/JiaLiPassion/zone-promise-seq

The code will look like this.
```
<html>
  <script src="./node_modules/zone.js/dist/zone-evergreen.js"></script>
  <script>
    const promise = Promise.resolve(1);
    function doSomethingWithoutZoneTask() {
      window[Zone.__symbol__('requestAnimationFrame')](() => {
        console.log('handler begin');
        promise.then(() => {
          console.log('promise then called');
        });
        div.removeChild(btn);
        console.log('handler end');
      });
    }
    window.onload = function() {
      const btn = document.getElementById('btn');
      const div = document.getElementById('div');
      btn.addEventListener('blur', () => {
        console.log('blur');
      });
      btn.addEventListener('click', () => {
        doSomethingWithoutZoneTask();
      });
    }
  </script>
  <body>
    <div id="div">
      <button id="btn">Test</button>
    </div>
  </body>
</html>
```

And when user click button `Test`, the output in the console will be

```
handler begin
blur
promise then called
handler end
```

So the `promise.then` callback is being executed synchronize which is not correct.

The reason is here https://github.com/angular/angular/blob/master/packages/zone.js/lib/zone.ts#L1283, and it seems this issue is a known issue, since there is a test case here, https://github.com/angular/angular/blob/master/packages/zone.js/test/browser/element.spec.ts#L73

so in some cases, the `drainMicroTaskQueue` will be invoked too early since `zone.js` wrongly consider. the current task is the top task, we should always use `native promsie` to schedule a microTask to run `drainMicroTaskQueue`.

JiaLiPassion:
Close #36839

This is a known issue of zone.js,

```
(window as any)[(Zone as any).__symbol__('setTimeout')](() => {
  let log = '';
  button.addEventListener('click', () => {
    Zone.current.scheduleMicroTask('test', () => log += 'microtask;');
    log += 'click;';
  });
  button.click();
  expect(log).toEqual('click;microtask;');
  done();
});
```

Since in this case, we use native `setTimeout` which is not a ZoneTask,
so zone.js consider the button click handler as the top Task then drain the
microTaskQueue after the click at once, which is not correct(too early).

This case was an edge case and not reported by the users, until we have the
new option ngZoneEventCoalescing, since the event coalescing will happen
in native requestAnimationFrame, so it will not be a ZoneTask, and zone.js will
consider any Task happen in the change detection stage as the top task, and if
there are any microTasks(such as Promise.then) happen in the process, it may be
drained earlier than it should be, so to prevent this situation, we need to schedule
a fake event task, so the Task happen in the change detection stage will not be
considered as top ZoneTask.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

### DIFF
--- a/packages/core/src/zone/ng_zone.ts
+++ b/packages/core/src/zone/ng_zone.ts
@@ -243,6 +243,10 @@ interface NgZonePrivate extends NgZone {
   isStable: boolean;
   shouldCoalesceEventChangeDetection: boolean;
   nativeRequestAnimationFrame: (callback: FrameRequestCallback) => number;
+
+  // Cache of  "fake" top eventTask. This is done so that we don't need to schedule a new task every
+  // time we want to run a `checkStable`.
+  fakeTopEventTask: Task;
 }
 
 function checkStable(zone: NgZonePrivate) {
@@ -268,9 +272,23 @@ function delayChangeDetectionForEvents(zone: NgZonePrivate) {
     return;
   }
   zone.lastRequestAnimationFrameId = zone.nativeRequestAnimationFrame.call(global, () => {
-    zone.lastRequestAnimationFrameId = -1;
-    updateMicroTaskStatus(zone);
-    checkStable(zone);
+    // This is a work around for https://github.com/angular/angular/issues/36839.
+    // The core issue is that when event coalescing is enabled it is possible for microtasks
+    // to get flushed too early (As is the case with `Promise.then`) between the
+    // coalescing eventTasks.
+    //
+    // To workaround this we schedule a "fake" eventTask before we process the
+    // coalescing eventTasks. The benefit of this is that the "fake" container eventTask
+    //  will prevent the microtasks queue from getting drained in between the coalescing
+    // eventTask execution.
+    if (!zone.fakeTopEventTask) {
+      zone.fakeTopEventTask = Zone.root.scheduleEventTask('fakeTopEventTask', () => {
+        zone.lastRequestAnimationFrameId = -1;
+        updateMicroTaskStatus(zone);
+        checkStable(zone);
+      }, undefined, () => {}, () => {});
+    }
+    zone.fakeTopEventTask.invoke();
   });
   updateMicroTaskStatus(zone);
 }
--- a/packages/platform-browser/test/dom/events/event_manager_spec.ts
+++ b/packages/platform-browser/test/dom/events/event_manager_spec.ts
@@ -370,6 +370,52 @@ describe('EventManager', () => {
       done();
     });
   });
+
+  it('should not drain micro tasks queue too early with shouldCoalesceEventChangeDetection=true',
+     (done: DoneFn) => {
+       doc = getDOM().supportsDOMEvents() ? document : getDOM().createHtmlDocument();
+       zone = new NgZone({shouldCoalesceEventChangeDetection: true});
+       domEventPlugin = new DomEventsPlugin(doc);
+       const element = el('<div></div>');
+       const child = el('<div></div>');
+       doc.body.appendChild(element);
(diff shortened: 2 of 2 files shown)


packages/core/src/zone/ng_zone.ts
```diff
@@ -243,6 +243,10 @@ interface NgZonePrivate extends NgZone {
   isStable: boolean;
   shouldCoalesceEventChangeDetection: boolean;
   nativeRequestAnimationFrame: (callback: FrameRequestCallback) => number;
+
+  // this fake top eventTask will be cached, so we don't need to schedule a new one every time
+  // we want to run a checkStable
```
mhevery, line 248:
```suggestion
  // Cache of  "fake" top eventTask. This is done so that we don't need to schedule a new task every time
  // we want to run a `checkStable`.
```

packages/core/src/zone/ng_zone.ts
```diff
@@ -268,9 +272,25 @@ function delayChangeDetectionForEvents(zone: NgZonePrivate) {
...
+    // This is a known issue of zone.js, since the event coalescing will happen
+    // in native requestAnimationFrame, so it will not be a ZoneTask, and zone.js will
+    // consider any Task happen in the change detection stage as the top task, and if
+    // there are any microTasks(such as Promise.then) happen in the process, it may be
+    // drained earlier than it should be, so to prevent this situation, we need to schedule
+    // a fake event task, so the Task happen in the change detection stage will not be
+    // considered as top ZoneTask. And we schedule this task in root Zone so it will
+    // not trigger any additional hooks of angular ZoneSpec.
+    // And we don't need to schedule this event task every time, it is just a container,
+    // we can reuse it.
```
mhevery, line 285:
```suggestion
    // This is a work around for https://github.com/angular/angular/issues/36839.
    // The core issue is that when event coalescing is enabled it is possible for microtasks 
    // to get flushed too early (As is the case with `Promise.then`) between the
    // coalescing macrotasks.
    // 
    // To workaround this we schedule a "fake" macrotasks before we process the 
    // coalescing macrotasks. The benefit of this is that the "fake" container makrotask
    //  will prevent the microtaks queue from getting drained in between the coalescing 
    // macrotasks execution.
```

mhevery (approved):

packages/core/src/zone/ng_zone.ts
JiaLiPassion: Got it, thanks. updated.

packages/core/src/zone/ng_zone.ts
JiaLiPassion: Got it, thanks. updated.

mhevery:
[presubmit](http://test/OCL:310160547:BASE:310160591:1588780489309:d2ddf3b6)

IgorMinar:
http://test/OCL:312544132:BASE:312544170:1590006540854:953540eb

JiaLiPassion:
@IgorMinar, it seems the presubmit failed, is there any error information? Thanks.

AndrewKushnir:
Hi @JiaLiPassion, FYI I've started a new presubmit and will keep you updated.

Could you please also rebase this PR on top of the latest master (so that we also rerun CI)?

Thank you.

JiaLiPassion:
@AndrewKushnir, sure, I just rebased the PR, thank you.

mhevery:
Failures in `google3` seem unrelated.

PR #36842, opened 2020-04-28
Format/Lint bazel via ng-dev format

josephperrott:

___

Beginning with #36842 the previous `yarn bazel:*` commands are no longer used for formatting bazel related files across the repo. Instead this task is now performed using the ng-dev command. The formatting enforced and performed is identical and does not result in any change to the source code of the repository.

This is being done to streamline the commands and usages for running formatting and linting in the repository.

As we no longer rely on the yarn commands, if they are called a deprecation message will be returned.

Usage changes:
`yarn bazel:format` -> `yarn ng-dev format all`
`yarn bazel:lint-fix` -> `yarn ng-dev format all`
`yarn bazel:lint` -> `yarn ng-dev format all --check`


josephperrott:
Caretaker Note: Landing this PR will require coordination as this change will need to be announced to contributors.

dev-infra/format/formatters.ts
```diff
@@ -0,0 +1,81 @@
...
+    bazel: ['**/*.bzl', '**/BUILD.bazel', '**/WORKSPACE', '**/BUILD'],
+    jsTs: ['**/*.{t,j}s'],
+  },
+};
+
+// Active config used for execution, created from a merge of the default
+// config with the loaded config.
+const config: FormatConfig = {
+  matchers: {},
+  ...getAngularDevConfig<'format', FormatConfig>().format
```
devversion, line 25: This runs whenever the file is loaded. Does that mean that any `ng-dev` command will fail if no dev-infra or format configuratio is created? It might make sense to only compute this if needed.

dev-infra/format/cli.ts
```diff
@@ -34,9 +34,36 @@ export function buildFormatParser(localYargs: yargs.Argv) {
...
+            break;
+          default:
+            console.warn(`Error: Unrecognized previous command.`);
+        }
+        console.warn();
+        console.warn(`You can find more usage information by running:`);
+        console.warn(`  yarn ng-dev format --help`);
+        console.warn();
+        console.warn(`For more on the rationale and effects of this deprecation visit:`);
+        // TODO(josephperrott): Update this PR to the correct URL.
```
devversion, line 65: can this TODO be removed now? The URL seems to be correct already?

dev-infra/format/format.ts
```diff
@@ -57,26 +61,33 @@ export async function formatFiles(unfilteredFiles: string[]) {
...
-    if (code !== 0) {
-      failures.push(file);
-    }
-  });
+  const formatterResults = await Promise.allSettled([
+    await runFormatterInParallel(
+        files, BAZEL_FORMATTER, 'check',
+        (file, code, stdout) => {
+          // Add any files failing format checks to the list.
+          if (code !== 0 || !JSON.parse(stdout)['success']) {
```
devversion, line 73: Optional idea: Right now this code snippet requires understanding of what command is run for the Bazel formatter. The command is separated though. 

Should we move the logic for determining success to the `BAZEL_FORMATTER` too? potentially we can integrate the success detection logic into `runFormatterInParallel` too.

dev-infra/format/format.ts
```diff
@@ -8,42 +8,46 @@
...
-  await runInParallel(files, `${getFormatterBinary()} -i -style=file`, (file, code, _, stderr) => {
-    if (code !== 0) {
-      formatFailed = true;
-      console.error(`Error running clang-format on: ${file}`);
-      console.error(stderr);
-      console.error();
-    }
-  });
+  const formatterResults = await Promise.allSettled([
+    await runFormatterInParallel(
```
devversion, line 23: Does this now acquire twice of the available CPU threads?

dev-infra/format/format.ts
```diff
@@ -8,42 +8,46 @@
...
-  // time from 276 seconds to 39 seconds for the same 2700 files
-  await runInParallel(files, `${getFormatterBinary()} -i -style=file`, (file, code, _, stderr) => {
-    if (code !== 0) {
-      formatFailed = true;
-      console.error(`Error running clang-format on: ${file}`);
-      console.error(stderr);
-      console.error();
-    }
-  });
+  const formatterResults = await Promise.allSettled([
```
devversion, line 22: In what cases does `runFormatterInParallel` reject? If one fails, then why do we want to wait for the others to finish too? Can you add a comment on why `allSettled` is used?

dev-infra/format/formatters.ts
josephperrott: Ahh yes, I suppsed the `ng-dev`  command would fail if there is no`.dev-infra.json` file, it does correctly handle no `format` key since `{...undefined}` is properly handled as a noop by the spread operator.

I will need to rethink how we should do this config loading based on that.

dev-infra/format/format.ts
josephperrott: I thought about doing that, but then was worried I was over engineering things.  Glad to hear that it came to mind for someone else as well.

dev-infra/format/format.ts
josephperrott: It still runs serially because we are awaiting the `runFormatterInParallel` call in the array being used as a parameter to `Promise.allSettled`.

dev-infra/format/format.ts
josephperrott: It actually doesn't have any rejection cases, since it is mostly just in a promise to allow us to parallelize things.

dev-infra/format/format.ts
devversion: Oh right! Why do we need to use the `Promise.allSettled` at all then? If things are awaited already, then this shouldn't be needed, right?

josephperrott: I supposed thats true. Done.

dev-infra/format/format.ts
```diff
@@ -7,46 +7,21 @@
...
-      console.error();
-    }
-  });
+  if (failures === false) {
+    process.exit(0);
+  }
 
   // The process should exit as a failure if any of the files failed to format.
-  if (formatFailed) {
+  if (failures.length !== 0) {
```
devversion, line 24: It's a bit unfortunate that we need to mix where messages are printed. e.g. sometimes it's here (`checkFiles`), or in `runFormatterInParallel` or sometimes in the action callback.

I'm not sure if we should spend too much time on it though. Maybe a comment is useful enough and can say that the formatter callback prints these unexpected errors?

Related to that, it might be worth considering to moving the `No files matched for formatting.` message into both `checkFiles` and `formatFiles` for a more canonical place where non-debug messages are printed.

devversion (approved):
LGTM. Just added a few minor cleanup comments. Feel free to remove the cleanup label.

Also a quick heads up that the lint job fails in this PR.

dev-infra/format/formatters/index.ts
```diff
@@ -0,0 +1,28 @@
...
+import {ClangFormat} from './clang-format';
+
+/**
+ * Get all defined formatters which are active based on the current loaded config.
+ */
+export function getActiveFormatters() {
+  let config = {};
+  try {
+    config = getAngularDevConfig<'format', FormatConfig>().format || {};
+  } catch {}
```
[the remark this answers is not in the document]
josephperrott: The try catch is actually currently in place because `getAngularDevConfig` throws if the file does not exist.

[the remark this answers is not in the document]
devversion: Yeah, but isn't that intended now that we only call this when the formatting command runs?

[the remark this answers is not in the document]
josephperrott: Yes, but we still need to handle the error being thrown if `ng-dev format [command]` is run without the config file existing.

Or at least that is my thought in having this in place.

dev-infra/format/format.ts
josephperrott: I moved out the comments about no files matching into the `checkFiles` and `formatFiles` functions.

I will do a followup PR to update where messages are printed out of to make it a bit more consistent.  I think its not worth adding even more change into what's happened so far.

dev-infra/format/formatters/index.ts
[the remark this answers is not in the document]
devversion: I was thinking that that error is already specific enough and actually exists specifically for such cases. Otherwise `getAngularDevConfig` could just return null.

Either way, we can follow-up on this too.

PR #36844, opened 2020-04-28
Use ng-dev for bazel build stamping

josephperrott:

___

This change should be transparent to users as the stamping process is automatically called by bazel during the release.


josephperrott:
Caretaker Note: Merging this PR will need to be coordinated with an announcement so that caretakers are made aware of the under the hood change.

AndrewKushnir:
@josephperrott, I'm wondering if we can add tests for this change (if we have infrastructure for this)?

dev-infra/release/env-stamp.ts
```diff
@@ -0,0 +1,64 @@
...
+}
+
+/** Whether the repo has local changes. */
+function hasLocalChanges() {
+  return !execSync(`git status --untracked-files=no --porcelain`);
+}
+
+/** Get the version based on the most recent semver tag. */
+function getSCMVersion() {
+  const version = execSync(`git describe --match [0-9]*.[0-9]*.[0-9]* --abbrev=7 --tags HEAD`);
```
devversion, line 43: In the components repo we just rely on the version from the `package.json`. Do you know why we need to do find it through Git? I know that framework did this in the past, but I'm wondering if it's time to consider the other more predictable approach.

dev-infra/cli.ts
```diff
@@ -11,14 +11,16 @@ import {tsCircularDependenciesBuilder} from './ts-circular-dependencies/index';
...
 import {buildCommitMessageParser} from './commit-message/cli';
 import {buildFormatParser} from './format/cli';
+import {buildReleaseParser} from './release/cli';
 
 yargs.scriptName('ng-dev')
     .demandCommand()
     .recommendCommands()
-    .command('ts-circular-deps <command>', '', tsCircularDependenciesBuilder)
-    .command('pullapprove <command>', '', buildPullapproveParser)
+    .command('release <command>', '', buildReleaseParser)
```
gkalpak, line 19: No alphabetic ordering? 🤔

.bazelrc
```diff
@@ -42,7 +42,7 @@ test --incompatible_strict_action_env
 # Releases should always be stamped with version control info
 # This command assumes node on the path and is a workaround for
 # https://github.com/bazelbuild/bazel/issues/4802
-build:release --workspace_status_command="node ./tools/bazel_stamp_vars.js"
+build:release --workspace_status_command="yarn -s ts-node dev-infra/release/cli.ts build-env-stamp"
```
gkalpak, line 45: OOC, why not use this through the main cli (with `yarn -s ng-dev release build-env-stamp`)?

dev-infra/release/env-stamp.ts
```diff
@@ -0,0 +1,64 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {execSync as _execSync} from 'child_process';
```
gkalpak, line 9: OOC, why not use `shelljs.exec()` (as in other files)?

dev-infra/release/env-stamp.ts
```diff
@@ -0,0 +1,64 @@
...
+}
+
+/** Run the execSync command and return the stdout as a trimmed string. */
+function execSync(cmd: string) {
+  return _execSync(cmd).toString().trim();
+}
+
+/** Whether the repo has local changes. */
+function hasLocalChanges() {
+  return !execSync(`git status --untracked-files=no --porcelain`);
```
gkalpak, line 38:
```suggestion
  return !!execSync(`git status --untracked-files=no --porcelain`);
```

gkalpak (approved):

dev-infra/release/env-stamp.ts
josephperrott: It mostly was just a hold over from the script it originated from using `execSync` from `child_process`

dev-infra/release/env-stamp.ts
josephperrott: I believe this is done within `angular/angular` still to allow for the version to be checked to verify that the build being created with a stamp, is actually the tagged release.  If the current repository state is not the exact sha that is the tag point, then it does not have the version of the sha.

Additionally, this allows for us to have stamping within our build artifacts created during a run on CI without it being tagged in our `package.json`s that are created as the actually released version.


Its definitely something we should step back and consider for what is our best option.  I think this is something that definitely makes sense to unify on, to allow for better tooling both via `ng-dev` as well as our `bazel` macros/rules

devversion: That makes sense. The thing I'm slightly worried about is that the stamping relies on Git to determine the _current_ version. With the `package.json` it is super easy and predictable to tell what the release will be. That also matches what's usually the process when building and publishing through NPM. i.e. Updating the version in the `package.json`, building and publishing.

It's totally valid to create the tag after the release packages have been built (build could theoretically fail; and then a tag is not desired). 

The benefit for CI snapshot builds that you mentioned is definitely a good one. We could chat about this as a follow-up. I just want to make clear that components repo currently uses a different process and that I don't know whether this is too unpredictable through Git (see how it silently used the wrong version when we started for v10.x.x)

dev-infra/release/env-stamp.ts
```diff
@@ -0,0 +1,64 @@
...
+  console.info(`BUILD_SCM_HASH ${getCurrentSha()}`);
+  console.info(`BUILD_SCM_LOCAL_CHANGES ${hasLocalChanges()}`);
+  console.info(`BUILD_SCM_USER ${getCurrentGitUser()}`);
+  console.info(`BUILD_SCM_VERSION ${getSCMVersion()}`);
+  process.exit(0);
+}
+
+/** Run the exec command and return the stdout as a trimmed string. */
+function exec(cmd: string) {
+  return _exec(cmd, {silent: true}).toString().trim();
```
gkalpak, line 33: Nit: `.toString()` is not needed with `shelljs.exec()`.

gkalpak (approved):

dev-infra/release/env-stamp.ts
josephperrott: How do you feel about landing this as is for now, and then we can determine what flow/option is best to standardize on long term in follow up work?

devversion: Yeah, 100%. I also proposed above that we can chat about this in a follow-up 😄

devversion (approved):

josephperrott:
@AndrewKushnir I don't believe there is really anything to test in the unit test methodology as there is not logic per se, but rather we are just logging the output of specific git commands.

And for a more integration style test, we would essentially just be checking that the output of the command matches the output of manually running the git command since every time the test runs it would run in a different git context.

Issue #36539, opened 2020-04-09
fix(service-worker): handle eagerly-cached assets unexpectedly missing from the cache


gkalpak:
## Context

It has been observed that occasionally angular.io loads an `index.html` file that contains (transitive) references to some JS chunks (e.g. lazy-loaded custom elements, such as the ToC module/component) which no longer exist neither on the server nor on the SW cache. When this happens, the page cannot be loaded and user sees the app-shell with a blank content area.

The frequency of these issues (which have been reported for other apps using `@angular/service-worker`) seem to be declining recently (see https://github.com/angular/angular/issues/28114#issuecomment-581989614), possibly due to fixes such as #32525. They still happen, though.

This is puzzling because these missing chunks are part of the same asset cache group (`app-shell`) and should be cached together with the `index.html` and `main.js`/`runtime.js` bundles that reference them. #28114 has much more info and in-depth discussions.

## A Theory

While we haven't found the root cause of this, here is a theory:

The browser deletes some entries from the `app-shell` cache. E.g. it removes the ToC module/component chunk but retains the `index.html` and `main.js`/`runtime.js` that reference it.

Browsers are free to decide when/what to evict from the cache, so different browsers can have different behavior.

## Possible Solution

One possible solution is be to:
1. Detect this situation in the SW (i.e. when a hashed, eagerly cached asset cannot be found in the cache and also does not exist on the server).
2. Purge the corresponding cache.
3. Communicate from the SW to the page that the SW and the app are in an unrecoverable state and the page needs to be reloaded.
4. Have the page listen for the message and handle it accordingly. For example it could show a notification to the user prompting them to reload) or automatically reload the page. (For angular.io, the latter would make more sense, since there is basically no app state to be lost.)

sonukapoor:
@gkalpak I will try to take a look at this.

IgorMinar:
I'm marking this as freq:low (it does not happen very often) + severity: regression (when it does however happen, it really sucks because an app that worked no longer works).

IgorMinar:
thanks @sonukapoor for taking this on!

sonukapoor:
@gkalpak please see the below inline questions:

> One possible solution is be to:
> 
> 1. Detect this situation in the SW (i.e. when a hashed, eagerly cached asset cannot be found in the cache and also does not exist on the server).
> 2. Purge the corresponding cache.

If the asset doesn't exist in the cache anymore, then nothing needs to be purged. Is that my correct understanding?

[...]

PR #36847, opened 2020-04-29
fix(service-worker): add `UNRECOVERABLE_STATE` handler 

sonukapoor:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #36539

- [ ] Yes
- [X] No

### DIFF
--- a/aio/content/examples/service-worker-getting-started/src/app/unrecoverable-state.service.ts
+++ b/aio/content/examples/service-worker-getting-started/src/app/unrecoverable-state.service.ts
@@ -1,13 +1,18 @@
 import { Injectable } from '@angular/core';
 import { SwUpdate } from '@angular/service-worker';
 
+function alertUser(message): boolean {
+  return true;
+}
+
 // #docregion sw-unrecoverable-state
 @Injectable()
-export class UnrecoverableStateService {
-
+export class HandleUnrecoverableStateService {
   constructor(updates: SwUpdate) {
     updates.unrecoverable.subscribe(event => {
-        document.location.reload();
+      alertUser(
+        `An error occurred that we cannot recover from:\n${event.reason}\n\n` +
+        'Please reload the page.');
     });
   }
 }
--- a/aio/content/examples/service-worker-getting-started/src/app/unrecoverable-state.service.ts
+++ b/aio/content/examples/service-worker-getting-started/src/app/unrecoverable-state.service.ts
@@ -1,0 +1,14 @@
+import { Injectable } from '@angular/core';
+import { SwUpdate } from '@angular/service-worker';
+
+// #docregion sw-unrecoverable-state
+@Injectable()
+export class UnrecoverableStateService {
+
+  constructor(updates: SwUpdate) {
+    updates.unrecoverable.subscribe(event => {
+        document.location.reload();
+    });
+  }
+}
+// #enddocregion sw-unrecoverable-state
--- a/aio/content/guide/service-worker-communications.md
+++ b/aio/content/guide/service-worker-communications.md
@@ -67,6 +67,13 @@ Therefore, it is recommended to reload the page once the promise returned by `ac
 
 </div>
 
+### Handling an unrecoverable state
+
+This can happen when the browser evicts eagerly cached assets from the cache, which can also not be found on the server anymore. This can lead to broken state where only parts of the application will load and others will fail. In such cases, the best way to fix this is to reload the browser.
+
+<code-example path="service-worker-getting-started/src/app/unrecoverable-state.service.ts" header="unrecoverable-state.service.ts" region="sw-unrecoverable-state"></code-example>
+
+
 ## More on Angular service workers
 
 You may also be interested in the following:
--- a/aio/content/guide/service-worker-communications.md
+++ b/aio/content/guide/service-worker-communications.md
@@ -72,17 +72,26 @@ Therefore, it is recommended to reload the page once the promise returned by `ac
 In some cases, the version of the app used by the service worker to serve a client might be in a broken state that cannot be recovered from without a full page reload.
 
 For example, imagine the following scenario:
-- A user opens the app for the first time and the service worker caches the latest version of the app. Let's assume the app's cached assets include `index.html`, `main.<main-hash-1>.js` and `lazy-chunk.<lazy-hash-1>.js`.
+- A user opens the app for the first time and the service worker caches the latest version of the app.
+Let's assume the app's cached assets include `index.html`, `main.<main-hash-1>.js` and `lazy-chunk.<lazy-hash-1>.js`.
 - The user closes the app and does not open it for a while.
-- After some time, a new version of the app is deployed to the server. This newer version includes the files `index.html`, `main.<main-hash-2>.js` and `lazy-chunk.<lazy-hash-2>.js` (note that the hashes are different now, because the content of the files has changed). The old version is no longer available on the server.
-- In the meantime, the user's browser decides to evict `lazy-chunk.<lazy-hash-1>.js` from its cache. Browsers may decide to evict specific (or all) resources from a cache in order to reclaim disk space.
-- The user opens the app again. The service worker serves the latest version known to it at this point, namely the old version (`index.html` and `main.<main-hash-1>.js`).
+- After some time, a new version of the app is deployed to the server.
+This newer version includes the files `index.html`, `main.<main-hash-2>.js` and `lazy-chunk.<lazy-hash-2>.js` (note that the hashes are different now, because the content of the files has changed).
+The old version is no longer available on the server.
+- In the meantime, the user's browser decides to evict `lazy-chunk.<lazy-hash-1>.js` from its cache.
+Browsers may decide to evict specific (or all) resources from a cache in order to reclaim disk space.
+- The user opens the app again.
+The service worker serves the latest version known to it at this point, namely the old version (`index.html` and `main.<main-hash-1>.js`).
 - At some later point, the app requests the lazy bundle, `lazy-chunk.<lazy-hash-1>.js`.
-- The service worker is unable to find the asset in the cache (remember that the browser evicted it). Nor is it able to retrieve it from the server (since the server now only has `lazy-chunk.<lazy-hash-2>.js` from the newer version).
+- The service worker is unable to find the asset in the cache (remember that the browser evicted it).
+Nor is it able to retrieve it from the server (since the server now only has `lazy-chunk.<lazy-hash-2>.js` from the newer version).
 
-In the above scenario, the service worker is not able to serve an asset that would normally be cached. That particular app version is broken and there is no way to fix it without reloading the page. In such cases, the service worker notifies the client by sending a `UnrecoverableStateEvent` event. You can subscribe to `SwUpdate#unrecoverable` to be notified and handle these errors.
+In the above scenario, the service worker is not able to serve an asset that would normally be cached.
+That particular app version is broken and there is no way to fix the state of the client without reloading the page.
+In such cases, the service worker notifies the client by sending a `UnrecoverableStateEvent` event.
+You can subscribe to `SwUpdate#unrecoverable` to be notified and handle these errors.
 
-<code-example path="service-worker-getting-started/src/app/unrecoverable-state.service.ts" header="unrecoverable-state.service.ts" region="sw-unrecoverable-state"></code-example>
+<code-example path="service-worker-getting-started/src/app/handle-unrecoverable-state.service.ts" header="handle-unrecoverable-state.service.ts" region="sw-unrecoverable-state"></code-example>
 
 
 ## More on Angular service workers
--- a/aio/content/guide/service-worker-communications.md
+++ b/aio/content/guide/service-worker-communications.md
@@ -73,22 +73,22 @@ In some cases, the version of the app used by the service worker to serve a clie
 
 For example, imagine the following scenario:
 - A user opens the app for the first time and the service worker caches the latest version of the app.
-Let's assume the app's cached assets include `index.html`, `main.<main-hash-1>.js` and `lazy-chunk.<lazy-hash-1>.js`.
+  Let's assume the app's cached assets include `index.html`, `main.<main-hash-1>.js` and `lazy-chunk.<lazy-hash-1>.js`.
 - The user closes the app and does not open it for a while.
 - After some time, a new version of the app is deployed to the server.
-This newer version includes the files `index.html`, `main.<main-hash-2>.js` and `lazy-chunk.<lazy-hash-2>.js` (note that the hashes are different now, because the content of the files has changed).
-The old version is no longer available on the server.
+  This newer version includes the files `index.html`, `main.<main-hash-2>.js` and `lazy-chunk.<lazy-hash-2>.js` (note that the hashes are different now, because the content of the files has changed).
+  The old version is no longer available on the server.
 - In the meantime, the user's browser decides to evict `lazy-chunk.<lazy-hash-1>.js` from its cache.
-Browsers may decide to evict specific (or all) resources from a cache in order to reclaim disk space.
+  Browsers may decide to evict specific (or all) resources from a cache in order to reclaim disk space.
 - The user opens the app again.
-The service worker serves the latest version known to it at this point, namely the old version (`index.html` and `main.<main-hash-1>.js`).
+  The service worker serves the latest version known to it at this point, namely the old version (`index.html` and `main.<main-hash-1>.js`).
 - At some later point, the app requests the lazy bundle, `lazy-chunk.<lazy-hash-1>.js`.
 - The service worker is unable to find the asset in the cache (remember that the browser evicted it).
-Nor is it able to retrieve it from the server (since the server now only has `lazy-chunk.<lazy-hash-2>.js` from the newer version).
+  Nor is it able to retrieve it from the server (since the server now only has `lazy-chunk.<lazy-hash-2>.js` from the newer version).
 
 In the above scenario, the service worker is not able to serve an asset that would normally be cached.
 That particular app version is broken and there is no way to fix the state of the client without reloading the page.
-In such cases, the service worker notifies the client by sending a `UnrecoverableStateEvent` event.
+In such cases, the service worker notifies the client by sending an `UnrecoverableStateEvent` event.
 You can subscribe to `SwUpdate#unrecoverable` to be notified and handle these errors.
 
 <code-example path="service-worker-getting-started/src/app/handle-unrecoverable-state.service.ts" header="handle-unrecoverable-state.service.ts" region="sw-unrecoverable-state"></code-example>
--- a/aio/content/guide/service-worker-communications.md
+++ b/aio/content/guide/service-worker-communications.md
@@ -69,7 +69,18 @@ Therefore, it is recommended to reload the page once the promise returned by `ac
 
 ### Handling an unrecoverable state
 
-This can happen when the browser evicts eagerly cached assets from the cache, which can also not be found on the server anymore. This can lead to broken state where only parts of the application will load and others will fail. In such cases, the best way to fix this is to reload the browser.
+In some cases, the version of the app used by the service worker to serve a client might be in a broken state that cannot be recovered from without a full page reload.
+
+For example, imagine the following scenario:
+- A user opens the app for the first time and the service worker caches the latest version of the app. Let's assume the app's cached assets include `index.html`, `main.<main-hash-1>.js` and `lazy-chunk.<lazy-hash-1>.js`.
+- The user closes the app and does not open it for a while.
+- After some time, a new version of the app is deployed to the server. This newer version includes the files `index.html`, `main.<main-hash-2>.js` and `lazy-chunk.<lazy-hash-2>.js` (note that the hashes are different now, because the content of the files has changed). The old version is no longer available on the server.
+- In the meantime, the user's browser decides to evict `lazy-chunk.<lazy-hash-1>.js` from its cache. Browsers may decide to evict specific (or all) resources from a cache in order to reclaim disk space.
+- The user opens the app again. The service worker serves the latest version known to it at this point, namely the old version (`index.html` and `main.<main-hash-1>.js`).
+- At some later point, the app requests the lazy bundle, `lazy-chunk.<lazy-hash-1>.js`.
+- The service worker is unable to find the asset in the cache (remember that the browser evicted it). Nor is it able to retrieve it from the server (since the server now only has `lazy-chunk.<lazy-hash-2>.js` from the newer version).
+
+In the above scenario, the service worker is not able to serve an asset that would normally be cached. That particular app version is broken and there is no way to fix it without reloading the page. In such cases, the service worker notifies the client by sending a `UnrecoverableStateEvent` event. You can subscribe to `SwUpdate#unrecoverable` to be notified and handle these errors.
 
 <code-example path="service-worker-getting-started/src/app/unrecoverable-state.service.ts" header="unrecoverable-state.service.ts" region="sw-unrecoverable-state"></code-example>
 
--- a/goldens/public-api/service-worker/service-worker.d.ts
+++ b/goldens/public-api/service-worker/service-worker.d.ts
@@ -29,9 +29,15 @@ export declare class SwUpdate {
     readonly activated: Observable<UpdateActivatedEvent>;
     readonly available: Observable<UpdateAvailableEvent>;
     get isEnabled(): boolean;
+    readonly unrecoverable: Observable<UnrecoverableStateEvent>;
     constructor(sw: ɵangular_packages_service_worker_service_worker_a);
     activateUpdate(): Promise<void>;
     checkForUpdate(): Promise<void>;
+}
+
+export declare interface UnrecoverableStateEvent {
+    reason: string;
+    type: 'UNRECOVERABLE_STATE';
 }
 
 export declare interface UpdateActivatedEvent {
--- a/packages/service-worker/src/low_level.ts
+++ b/packages/service-worker/src/low_level.ts
@@ -14,6 +14,8 @@ export const ERR_SW_NOT_SUPPORTED = 'Service workers are disabled or not support
 /**
  * An event emitted when a new version of the app is available.
  *
+ * @see {@link guide/service-worker-communications Service worker communication guide}
+ *
  * @publicApi
  */
 export interface UpdateAvailableEvent {
@@ -24,6 +26,8 @@ export interface UpdateAvailableEvent {
 
 /**
  * An event emitted when a new version of the app has been downloaded and activated.
+ *
+ * @see {@link guide/service-worker-communications Service worker communication guide}
  *
  * @publicApi
  */
@@ -41,6 +45,8 @@ export interface UpdateActivatedEvent {
  * cache nor from the server. This could happen if a new version is deployed to the server and the
  * service worker cache has been partially cleaned by the browser, removing some files of a previous
  * app version but not all.
+ *
+ * @see {@link guide/service-worker-communications Service worker communication guide}
  *
  * @publicApi
  */
--- a/packages/service-worker/src/low_level.ts
+++ b/packages/service-worker/src/low_level.ts
@@ -34,6 +34,22 @@ export interface UpdateActivatedEvent {
 }
 
 /**
+ * An event emitted when the version of the app used by the service worker to serve this client is
+ * in a broken state that cannot be recovered from and a full page reload is required.
+ *
+ * For example, the ServiceWorker may not be able to retrieve a required resource, neither from the
+ * cache nor from the server. This could happen if a new version is deployed to the server and the
+ * ServiceWorker cache has been partially cleaned by the browser, removing some files of a previous
+ * app version but not all.
+ *
+ * @publicApi
+ */
+export interface UnrecoverableStateEvent {
+  type: 'UNRECOVERABLE_STATE';
+  reason: string;
+}
+
+/**
  * An event emitted when a `PushEvent` is received by the service worker.
  */
 export interface PushEvent {
@@ -41,7 +57,7 @@ export interface PushEvent {
   data: any;
 }
 
-export type IncomingEvent = UpdateAvailableEvent|UpdateActivatedEvent;
+export type IncomingEvent = UpdateAvailableEvent|UpdateActivatedEvent|UnrecoverableStateEvent;
 
 export interface TypedEvent {
   type: string;
--- a/packages/service-worker/src/low_level.ts
+++ b/packages/service-worker/src/low_level.ts
@@ -37,9 +37,9 @@ export interface UpdateActivatedEvent {
  * An event emitted when the version of the app used by the service worker to serve this client is
  * in a broken state that cannot be recovered from and a full page reload is required.
  *
- * For example, the ServiceWorker may not be able to retrieve a required resource, neither from the
+ * For example, the service worker may not be able to retrieve a required resource, neither from the
  * cache nor from the server. This could happen if a new version is deployed to the server and the
- * ServiceWorker cache has been partially cleaned by the browser, removing some files of a previous
+ * service worker cache has been partially cleaned by the browser, removing some files of a previous
  * app version but not all.
  *
  * @publicApi
--- a/packages/service-worker/src/update.ts
+++ b/packages/service-worker/src/update.ts
@@ -17,6 +17,8 @@ import {ERR_SW_NOT_SUPPORTED, NgswCommChannel, UnrecoverableStateEvent, UpdateAc
  * Subscribe to update notifications from the Service Worker, trigger update
  * checks, and forcibly activate updates.
  *
+ * @see {@link guide/service-worker-communications Service worker communication guide}
+ *
  * @publicApi
  */
 @Injectable()
--- a/packages/service-worker/src/update.ts
+++ b/packages/service-worker/src/update.ts
@@ -9,7 +9,7 @@
 import {Injectable} from '@angular/core';
 import {NEVER, Observable} from 'rxjs';
 
-import {ERR_SW_NOT_SUPPORTED, NgswCommChannel, UpdateActivatedEvent, UpdateAvailableEvent} from './low_level';
+import {ERR_SW_NOT_SUPPORTED, NgswCommChannel, UnrecoverableStateEvent, UpdateActivatedEvent, UpdateAvailableEvent} from './low_level';
 
 
 
@@ -32,6 +32,13 @@ export class SwUpdate {
   readonly activated: Observable<UpdateActivatedEvent>;
 
   /**
+   * Emits an `UnrecoverableStateEvent` event whenever the version of the app used by the service
+   * worker to serve this client is in a broken state that cannot be recovered from without a full
+   * page reload.
+   */
+  readonly unrecoverable: Observable<UnrecoverableStateEvent>;
+
+  /**
    * True if the Service Worker is enabled (supported by the browser and enabled via
    * `ServiceWorkerModule`).
    */
@@ -43,10 +50,12 @@ export class SwUpdate {
     if (!sw.isEnabled) {
       this.available = NEVER;
       this.activated = NEVER;
+      this.unrecoverable = NEVER;
       return;
     }
     this.available = this.sw.eventsOfType<UpdateAvailableEvent>('UPDATE_AVAILABLE');
     this.activated = this.sw.eventsOfType<UpdateActivatedEvent>('UPDATE_ACTIVATED');
+    this.unrecoverable = this.sw.eventsOfType<UnrecoverableStateEvent>('UNRECOVERABLE_STATE');
   }
 
   checkForUpdate(): Promise<void> {
--- a/packages/service-worker/test/comm_spec.ts
+++ b/packages/service-worker/test/comm_spec.ts
@@ -435,7 +435,7 @@ import {MockPushManager, MockPushSubscription, MockServiceWorkerContainer, MockS
           },
         });
       });
-      it('processes resource removed notifications when sent', done => {
+      it('processes unrecoverable notifications when sent', done => {
         update.unrecoverable.subscribe(event => {
           expect(event.reason).toEqual('Invalid Resource');
           expect(event.type).toEqual('UNRECOVERABLE_STATE');
--- a/packages/service-worker/test/comm_spec.ts
+++ b/packages/service-worker/test/comm_spec.ts
@@ -435,6 +435,14 @@ import {MockPushManager, MockPushSubscription, MockServiceWorkerContainer, MockS
           },
         });
       });
+      it('processes resource removed notifications when sent', done => {
+        update.unrecoverable.subscribe(event => {
+          expect(event.reason).toEqual('Invalid Resource');
+          expect(event.type).toEqual('UNRECOVERABLE_STATE');
+          done();
+        });
+        mock.sendMessage({type: 'UNRECOVERABLE_STATE', reason: 'Invalid Resource'});
+      });
       it('processes update activation notifications when sent', done => {
         update.activated.subscribe(event => {
           expect(event.previous).toEqual({hash: 'A'});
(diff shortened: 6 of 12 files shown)


IgorMinar (commented):
I like where this is heading, but I suggest that we use terminology different from `unrecovered` because it gets awkward and confusing to use it. How about using `defunkt`, `terminalError`, or something of that sort?

packages/service-worker/src/low_level.ts
```diff
@@ -33,6 +33,16 @@ export interface UpdateActivatedEvent {
   current: {hash: string, appData?: Object};
 }
 
+/**
+ * An event emitted when a a resource can not be found anymore in the cache or on the server.
```
gkalpak, line 37: I would rather make this more generic, because in the future there might be other scenarios that constitute an unrecoverable state. How about something like:

```suggestion
 * An event emitted when the version of the app used by the service worker to serve this client is
 * in a broken state that cannot be recovered from and a full page reload is required.
 *
 * For example, this can happen when a required resource cannot be found neither in the cache nor on
 * the server.
```

gkalpak:
@IgorMinar, I was the one that proposed "unrecoverable"/"unrecoverableState" (I assume "unrecovered" is a typo). I find "unrecoverable" more descriptive than "defunkt" or "terminalError" (but not a hill I'm going to die on :grin:).

gkalpak (changes requested):
Thx, @sonukapoor! I've suggested some changes. LMK what you think.

Note to self: We need to document this in the appropriate guides (with code examples).

packages/service-worker/worker/src/assets.ts
```diff
@@ -147,10 +147,9 @@ export abstract class AssetGroup {
       // may specify things like credential inclusion, but for assets these are not honored in order
       // to avoid issues with opaque responses. The SW requests the data itself.
       const res = await this.fetchAndCacheOnce(this.adapter.newRequest(req.url));
-
```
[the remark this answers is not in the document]
sonukapoor: Empty lines are haunting me 👻

packages/service-worker/worker/src/assets.ts
```diff
@@ -314,8 +313,12 @@ export abstract class AssetGroup {
...
       // should never be cached as this can completely break applications.
       if (!res.ok) {
-        throw new Error(`Response not Ok (fetchAndCacheOnce): request for ${
-            req.url} returned response ${res.status} ${res.statusText}`);
+        const clients = await this.scope.clients.matchAll();
+        clients.forEach(client => {
+          client.postMessage({type: 'UNRECOVERABLE_STATE', url: req.url});
+        });
+
+        return res;
```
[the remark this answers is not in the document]
sonukapoor: @gkalpak Any suggestions on where a better place would be to put that code?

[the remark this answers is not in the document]
gkalpak: Off the top of my head, the `AssetGroup`'s `handleFetch()` method. It knows whether the URL was in `config.urls` (not in `patterns`), it knows whether it could be found in the cache or not, it knows that the server request failed (with an error code).

packages/service-worker/src/update.ts
```diff
@@ -31,6 +31,11 @@ export class SwUpdate {
    */
   readonly activated: Observable<UpdateActivatedEvent>;
 
+  /**
+   * Emits an `UnrecoverableStateEvent` event whenever the resource can not longer be found in the cache or server.
```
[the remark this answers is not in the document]
sonukapoor: You mean [this](https://github.com/angular/angular/pull/36847#discussion_r440904549)?

packages/service-worker/worker/src/assets.ts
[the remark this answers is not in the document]
sonukapoor: I assume you mean [here](https://github.com/angular/angular/blob/6e134b55af40d8d552b416577a496f305e890847/packages/service-worker/worker/src/assets.ts#L151): If I add the same code into that section, how do you identify which client is broken? I do not see anything on the `Client` that could help with that.

packages/service-worker/src/low_level.ts
```diff
@@ -33,6 +33,16 @@ export interface UpdateActivatedEvent {
...
+ *
+ * @publicApi
+ */
+export interface UnrecoverableStateEvent {
+  type: 'UNRECOVERABLE_STATE';
+  url: string;
```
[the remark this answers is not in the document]
sonukapoor: I assume you mean as an additional property on the interface?

sonukapoor:
@gkalpak Please review again. I have made some of the suggested changes. I will address unit tests once we are OK with the changes.

packages/service-worker/src/low_level.ts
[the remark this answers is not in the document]
gkalpak: I mean replacing the `url` property of the interface with a `reason` property.

packages/service-worker/worker/src/assets.ts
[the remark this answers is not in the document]
gkalpak: Yes, I meant there. You are right that `AssetGroup` is not the right place to know which clients to notify. `AssetGroup` _is_ the right place to know whether we are in an unrecoverable state. So, it has to pass that information to the `Driver` (which knows which clients to notify.

Maybe, we could introduce a new `UnrecoverableStateError` error, which `AssetGroup` will throw in its `handleFetch()` method (if the conditions are met). Such an error will propagate through `AppVersion`'s [handleFetch()][1] method to `Driver`'s [handleFetch()][2] method.

The `Driver` can then check if it's an `UnrecoverableStateError` and handle it appropriately. It can determine the affected client similar to how we do it [here][3] and notify those clients only.

[1]: https://github.com/angular/angular/blob/8154bbd5381ace543941ac974adbf7f3f5ada717/packages/service-worker/worker/src/app-version.ts#L154
[2]: https://github.com/angular/angular/blob/8154bbd5381ace543941ac974adbf7f3f5ada717/packages/service-worker/worker/src/driver.ts#L436
[3]: https://github.com/angular/angular/blob/8154bbd5381ace543941ac974adbf7f3f5ada717/packages/service-worker/worker/src/driver.ts#L762-L764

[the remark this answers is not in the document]
sonukapoor: That makes sense @gkalpak - I had added an additional commit to address - take a look and let me know what you think.

packages/service-worker/worker/src/assets.ts
```diff
@@ -150,6 +150,10 @@ export abstract class AssetGroup {
 
       // If this is successful, the response needs to be cloned as it might be used to respond to
       // multiple fetch operations at the same time.
```
gkalpak, line 152: This comment should be just above the closing operation.

packages/service-worker/worker/src/assets.ts
```diff
@@ -150,6 +150,10 @@ export abstract class AssetGroup {
...
+      if (!res.ok) {
```
gkalpak, line 153: AFAICT, `fetchAndCacheOnce()` will not return a not OK response. It will [throw an error][1] instead.

[1]: https://github.com/angular/angular/blob/5fa748b196d388f34aff2808bd8e4a7ab6b3af79/packages/service-worker/worker/src/assets.ts#L320-L323

gkalpak, line 153: Also, as mentioned in https://github.com/angular/angular/pull/36847#discussion_r440821788 (and the subsequent discussion), the requirements for throwing an `UnrecoverableStateError` are:

- The request is matched by `config.urls` (not the `config.patterns`).
- It isn't found in the cache.
- It fails to be fetched by the server. (For now, we can detect this by checking whether `fetchAndCacheOnce()` failed with an non-critical error.)
- TBD: The `AssetGroup`'s `updateMode` should be `prefetch`, not `lazy`.

packages/service-worker/worker/src/assets.ts
```diff
@@ -150,6 +150,10 @@ export abstract class AssetGroup {
...
+        throw new UnrecoverableStateError(`Failed to retrieve asset from server`);
```
gkalpak, line 154: Please include the name of the `AssetGroup` and the resource URL in the reason.

packages/service-worker/worker/src/driver.ts
```diff
@@ -1003,6 +1006,25 @@ export class Driver implements Debuggable, UpdateSource {
...
+    }
+
+    const brokenHash = broken[0];
+    const affectedClients = Array.from(this.clientVersionMap.entries())
+                                .filter(([clientId, hash]) => hash === brokenHash)
+                                .map(([clientId]) => clientId);
+
+    affectedClients.forEach(async clientId => {
+      const client = await this.scope.clients.get(clientId);
+      client.postMessage({type: 'UNRECOVERABLE_STATE', reason: 'Resource not found.'});
```
gkalpak, line 1024: The reason should be passed through from the `UnrecoverableStateError`.

gkalpak (changes requested):
Thx for making the changes, @sonukapoor. Here is another round of review :smiley:

packages/service-worker/worker/src/error.ts
```diff
@@ -17,3 +17,8 @@ export function errorToString(error: any): string {
     return `${error}`;
   }
 }
+
+export class UnrecoverableStateError extends Error {
```
gkalpak, line 21: It just occurred to me that `SwUnrecoverableStateError` would be more consistent 😇

packages/service-worker/worker/src/assets.ts
```diff
@@ -143,14 +143,27 @@ export abstract class AssetGroup {
...
+
+        // If this is successful, the response needs to be cloned as it might be used to respond to
+        // multiple fetch operations at the same time.
+        return res.clone();
+      } catch(err) {
+        const isCriticalError = err instanceof SwCriticalError;
+        const isInConfig = this.config.urls.indexOf(url) !== -1;
+        const isPrefetch = this.config.installMode === 'prefetch';
+
+        if (isCriticalError && isInConfig && isPrefetch) {
```
gkalpak, line 161: Actually, thinking about it some more, I don't think any error should trigger an `UnrecoverableStateError`. For example, the client of the server might be offline. We need to specifically detect 404 errors. Which means that we somehow need access to the raw `Response` object here.

Not sure what is the best way to achieve this. I have to investigate this further 😞

gkalpak: So, here is a new plan:

- Update `cacheBustedFetchFromNetwrok()` to throw an `UnrecoverableStateError` if `cacheBustedResult.status` is 404:

    https://github.com/angular/angular/blob/1550663b9e5bcd7870380d870ef84aa5e9211957/packages/service-worker/worker/src/assets.ts#L415-L421

    Something like:
    ```ts
    if (!cacheBustedResult.ok) {
      if (cacheBustedResult.status === 404) {
        throw new UnrecoverabvleStateError(...);
      } else {
        throw new SwCriticalError(...);
      }
    } 
    ```

- Let the `Error` pass through in `handleFetch()`. At this point, if we get an `UnrecoverableStateError` we know that (1) the resource is missing from the cache, (2) the resource is missing from the server (404) and (3) the resource is hashed (which makes the `config.urls.indexOf()` check kind of redundant).

- Make `UnrecoverableStateError` extend `SwCriticalError`, because we still want to call [versionFailed()][1].

[1]: https://github.com/angular/angular/blob/ff8e66ca227017b9f97b67d8a7d17667a3b3d82b/packages/service-worker/worker/src/driver.ts#L754

gkalpak (changes requested):
Thx, @sonukapoor! Here are some more comments for you :grin:
(Some of my previous comments have not been addressed, so I left them unresolved.)

BTW, there have been some SW updates recently. Make sure you rebase this PR on latest master :wink:

goldens/public-api/service-worker/service-worker.d.ts
```diff
@@ -29,6 +29,7 @@ export declare class SwUpdate {
     readonly activated: Observable<UpdateActivatedEvent>;
     readonly available: Observable<UpdateAvailableEvent>;
     get isEnabled(): boolean;
+    readonly unrecovered: Observable<UnrecoverableStateEvent>;
```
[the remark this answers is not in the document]
sonukapoor: The script that generates the `*.d.ts` file automatically changes the order of the properties.

aikidave:
Anyone know why this PR was assigned to me? :)

gkalpak:
Maybe because of [this comment][1]? 🤷‍♂️ 

[1]: https://github.com/angular/angular/pull/36847#pullrequestreview-430247130

packages/service-worker/worker/src/assets.ts
```diff
@@ -422,10 +423,15 @@ export abstract class AssetGroup {
 
         // If the response was unsuccessful, there's nothing more that can be done.
         if (!cacheBustedResult.ok) {
-          throw new SwCriticalError(
-              `Response not Ok (cacheBustedFetchFromNetwork): cache busted request for ${
-                  req.url} returned response ${cacheBustedResult.status} ${
-                  cacheBustedResult.statusText}`);
+          if (cacheBustedResult.status === 404) {
+            throw new SwUnrecoverableStateError(
+                `Failed to retrieve asset from server: (AssetGroup): ${this.config.name}`);
```
gkalpak, line 428: I think it would be usefule to also capture the resource that failed to be retrieved. How about:

```suggestion
                `Failed to retrieve hashed resource from the server. (AssetGroup: ${this.config.name} | URL: ${url})`);
```

gkalpak (changes requested):
Looking much better (and simpler). I think we are almost there.
Now we need some tests and documentation :grin:

packages/service-worker/worker/test/happy_spec.ts
```diff
@@ -1738,6 +1738,101 @@ describe('Driver', () => {
...
+          installMode: 'prefetch',
+          updateMode: 'prefetch',
+          urls: updatedFiles.list(),
+          patterns: [],
+          cacheQueryOptions: {ignoreVary: true},
+        }],
+        dataGroups: [],
+        navigationUrls: processNavigationUrls(''),
+        hashTable: tmpHashTableForFs(updatedFiles),
+      };
```
gkalpak, line 1783: Let's create a helper function that accepts a `MockFileSystem` instance and generates a `Manifest`.
Or actually a helper that generates a `MockServerState` from a `MockFileSystem`.

packages/service-worker/worker/test/happy_spec.ts
```diff
@@ -1738,6 +1738,101 @@ describe('Driver', () => {
...
+      // Remove `foo.hash.js` from the cache to emulate the browser evicting files from the cache.
+      await clearAllCaches(scope.caches);
+
+      // Try to retrieve `foo.hash.js`, which is neither in the cache nor on the server.
+      // This should put the SW in an unrecoverable state and notify clients.
+      expect(await makeRequest(scope, '/foo.hash.js')).toBeNull();
+      updatedServer.assertSawRequestFor('/foo.hash.js');
+
+      const mockClient = scope.clients.getMock('default')!;
+      expect(mockClient.messages).toEqual([
```
gkalpak, line 1827: For the sake of completeness, let's assert that the client didn't have any messages before the last request for `/foo.hash.js`.

packages/service-worker/worker/test/happy_spec.ts
```diff
@@ -1738,6 +1738,101 @@ describe('Driver', () => {
...
+        {
+          type: 'UNRECOVERABLE_STATE',
+          reason:
+              'Failed to retrieve hashed resource from the server. (AssetGroup: assets | URL: /foo.hash.js)'
+        },
+      ]);
+    });
```
gkalpak, line 1834: Can you also assert (potentially in a new, dedicated test) that the driver has triggered the `versionFailed()` sequence (e.g. entered degraded mode)?

gkalpak (changes requested):
Woohoo! Almost there :tada:

BTW, it occurred to me that this will not help much if the app has not loaded yet (i.e. if the error happens with one of the main JS bundles), but it _will_ help with lazy-loaded bundles (where the app has loaded and is capable of handling these messages).

We can introduce this first, see how it goes and later investigate more aggressive options (such as auto-reloading the client).

packages/service-worker/worker/test/happy_spec.ts
```diff
@@ -1762,6 +1875,27 @@ describe('Driver', () => {
 });
 })();
 
+function generateMockServerState(fileSystem: MockFileSystem) {
```
gkalpak, line 1878: This helper is only used in two tests. I suggest wrapping these two tests in a `describe()` function and move the helper inside the `describe()` block.

E.g.:


```ts
describe('when in unrecoverable state', () => {
  const generateMockServerState = ...;

  it('does one thing', ...);
  
  it('does another thing', ...);
});
```

packages/service-worker/worker/test/happy_spec.ts
```diff
@@ -1738,6 +1738,119 @@ describe('Driver', () => {
       expect(requestUrls2).toContain(httpsRequestUrl);
     });
 
+    it('broadcasts notifications about unrecoverable state', async () => {
```
gkalpak, line 1741: Broadcasts is not accurate. It only notifies affected clients.

Now that I mentioned this, we should also have a test verifying that non-affected clients are not notified. For example:
- Create the server state with an original version.
- Get three different SW clients to the original version.
- Update the server to a new version (let's call it `version-2`).
- Update the first two clients to `version-2`. (Keep the third client to the original version.)
- Update the server to a newer version (let's call it `version-3`) and break the cache for `version-2` by removing an asset.
- Try to get the non-existent asset from the first client.
- Verify that the first two clients (which were on the broken version) were notified and that the third client (which was on a different version) was not notified.

packages/service-worker/worker/test/happy_spec.ts
```diff
@@ -1738,6 +1738,119 @@ describe('Driver', () => {
...
+
+      // The SW is still able to serve `foo.hash.js` from the cache.
+      expect(await makeRequest(scope, '/foo.hash.js')).toBe('console.log("FOO");');
+      updatedServer.assertNoRequestFor('/foo.hash.js');
+
+      // Remove `foo.hash.js` from the cache to emulate the browser evicting files from the cache.
+      await clearAllCaches(scope.caches);
+
+      const mockClient = scope.clients.getMock('default')!;
+      expect(mockClient.messages).toEqual([]);
```
gkalpak, line 1843: These are irreelvant to this test. Please remove.

packages/service-worker/worker/test/happy_spec.ts
```diff
@@ -1738,6 +1738,119 @@ describe('Driver', () => {
...
+
+      // Try to retrieve `foo.hash.js`, which is neither in the cache nor on the server.
+      // This should put the SW in an unrecoverable state and notify clients.
+      expect(await makeRequest(scope, '/foo.hash.js')).toBeNull();
+      updatedServer.assertSawRequestFor('/foo.hash.js');
+
+      // This should also enter the `SW` into degrade mode.
```
gkalpak, line 1850:
```suggestion
      // This should also enter the `SW` into degraded mode, because the broken version was the
      // latest one.
```

packages/service-worker/worker/test/happy_spec.ts
gkalpak: Actually, it probably makes more sense to update the current test to have three clients (as described above) instead of adding another test.

gkalpak (changes requested):
Note to self: We still need to document this (with code snippets).

packages/service-worker/worker/test/happy_spec.ts
```diff
@@ -1738,117 +1738,161 @@ describe('Driver', () => {
...
+            new MockFileSystemBuilder()
+                .addFile('/index.html', '<script src="baz.hash.js"></script>')
+                .addFile('/baz.hash.js', 'console.log("BAZ");')
+                .build());
+
+        // Update the server to the latest version
+        scope.updateServerState(latestServer);
+        expect(await driver.checkForUpdate()).toEqual(true);
+
+        // Remove `foo.hash.js` from the cache to emulate the browser evicting files from the cache.
```
gkalpak, line 1816: As discussed offline, this is not ideal (removing all caches instead of just removing `foo.hash.js` from the cache), but is good enough for our purposes here. Please add a comment explaining that. For example:

```suggestion
        // Remove `foo.hash.js` from the cache to emulate the browser evicting files from the cache.
        // NOTE: For simplicity, we are clearing all caches instead of just removing `foo.hash.js`.
        //       This is not ideal, but OK for our purposes here.
```

(Same for similar lines in other tests in this `describe()` block.)

packages/service-worker/worker/test/happy_spec.ts
```diff
@@ -1738,117 +1738,161 @@ describe('Driver', () => {
...
+        await clearAllCaches(scope.caches);
+
+        // Get first client and verify its messages
+        const mockClient1 = scope.clients.getMock('client1')!;
+        // Clear all messages for this client
```
gkalpak, line 1821: What messages do we expect this client to have?

packages/service-worker/worker/test/happy_spec.ts
```diff
@@ -1738,117 +1738,161 @@ describe('Driver', () => {
...
+        };
+
+        return new MockServerStateBuilder()
+            .withManifest(manifest)
+            .withStaticFiles(fileSystem)
+            .build();
+      };
+
+      it('notify affected clients about unrecoverable state', async () => {
+        const originalServer = generateMockServerState(
```
gkalpak, line 1767: IMO, it would be clearer to use numbers for server version names. I.e. instead of `originalServer`, `updatedServer`, `latestServer` use `serverState1`, `serverState2`, `serverState3`.

packages/service-worker/worker/test/happy_spec.ts
```diff
@@ -1738,117 +1738,161 @@ describe('Driver', () => {
...
+
+        expect(mockClient1.messages).toEqual([
+          {
+            type: 'UNRECOVERABLE_STATE',
+            reason:
+                'Failed to retrieve hashed resource from the server. (AssetGroup: assets | URL: /bar.hash.js)'
+          },
+        ]);
+
+        // Because `client` failed, `client2` has been moved to the latest version.
```
gkalpak, line 1837:
```suggestion
        // Because `client1` failed, `client2` has been moved to the latest version.
```

packages/service-worker/worker/test/happy_spec.ts
```diff
@@ -1738,117 +1738,161 @@ describe('Driver', () => {
...
+        expect(await makeRequest(scope, '/bar.hash.js', 'client1')).toBeNull();
+        updatedServer.assertSawRequestFor('/bar.hash.js');
+
+        expect(mockClient1.messages).toEqual([
+          {
+            type: 'UNRECOVERABLE_STATE',
+            reason:
+                'Failed to retrieve hashed resource from the server. (AssetGroup: assets | URL: /bar.hash.js)'
+          },
+        ]);
```
gkalpak, line 1835: We should verify the same for `client2` and also verify that `client3` did not receive any messages.

packages/service-worker/worker/test/happy_spec.ts
```diff
@@ -1738,117 +1738,161 @@ describe('Driver', () => {
...
+
+        // Because `client` failed, `client2` has been moved to the latest version.
+        // Verify that by retrieving `baz.hash.js`.
+        expect(await makeRequest(scope, '/baz.hash.js', 'client2')).toBe('console.log("BAZ");');
```
gkalpak, line 1839: We should verify the same for `client1`.

packages/service-worker/worker/test/happy_spec.ts
```diff
@@ -1738,117 +1738,161 @@ describe('Driver', () => {
...
+        updatedServer.assertNoRequestFor('/baz.hash.js');
+      });
```
gkalpak, line 1841: We should also verify that `client3` remains on the first version (and can successfully request `foo.hash.js`).

packages/service-worker/worker/test/happy_spec.ts
```diff
@@ -1738,117 +1738,161 @@ describe('Driver', () => {
...
+        updatedServer.assertNoRequestFor('/foo.hash.js');
 
-      // Remove `foo.hash.js` from the cache to emulate the browser evicting files from the cache.
-      await clearAllCaches(scope.caches);
+        // Remove `foo.hash.js` from the cache to emulate the browser evicting files from the cache.
+        await clearAllCaches(scope.caches);
 
-      const mockClient = scope.clients.getMock('default')!;
-      expect(mockClient.messages).toEqual([]);
+        const mockClient = scope.clients.getMock('default')!;
```
gkalpak, line 1885: Please remove this too.

packages/service-worker/worker/test/happy_spec.ts
```diff
@@ -1738,117 +1738,161 @@ describe('Driver', () => {
...
+          hashTable: tmpHashTableForFs(fileSystem),
+        };
+
+        return new MockServerStateBuilder()
+            .withManifest(manifest)
+            .withStaticFiles(fileSystem)
+            .build();
+      };
+
+      it('notify affected clients about unrecoverable state', async () => {
```
gkalpak, line 1766:
```suggestion
      it('notifies affected clients', async () => {
```

packages/service-worker/worker/test/happy_spec.ts
```diff
@@ -1738,117 +1738,161 @@ describe('Driver', () => {
...
+
+        // Because `client` failed, `client2` has been moved to the latest version.
+        // Verify that by retrieving `baz.hash.js`.
+        expect(await makeRequest(scope, '/baz.hash.js', 'client2')).toBe('console.log("BAZ");');
+        updatedServer.assertNoRequestFor('/baz.hash.js');
+      });
 
-      const originalServer = generateMockServerState(originalFiles);
-      const updatedServer = generateMockServerState(updatedFiles);
+      it('enters degraded mode when unrecoverable state', async () => {
```
gkalpak, line 1843:
```suggestion
      it('enters degraded mode', async () => {
```

gkalpak (changes requested):

packages/service-worker/worker/test/happy_spec.ts
sonukapoor: Because we are updating the server state, the following messages in that array: `UPDATE_AVAILABLE`, `UPDATE_ACTIVATED`, `UPDATE_AVAILABLE`. I am clearing those first because I want to validate the `UNRECOVERABLE_STATE`.

gkalpak: :+1:
Can you update the comment to mention that. Something like:

```suggestion
        // Clear messages related to previous updates for this client
```

packages/service-worker/worker/test/happy_spec.ts
```diff
@@ -1738,6 +1738,170 @@ describe('Driver', () => {
...
+          navigationUrls: processNavigationUrls(''),
+          hashTable: tmpHashTableForFs(fileSystem),
+        };
+
+        return {
+          serverState: new MockServerStateBuilder()
+                           .withManifest(manifest)
+                           .withStaticFiles(fileSystem)
+                           .build(),
+          manifest
```
gkalpak, line 1765:
```suggestion
          manifest,
```

packages/service-worker/worker/test/happy_spec.ts
```diff
@@ -1738,6 +1738,170 @@ describe('Driver', () => {
...
+                .addFile('/index.html', '<script src="foo.hash.js"></script>')
+                .addFile('/foo.hash.js', 'console.log("FOO");')
+                .build());
+
+        const {serverState: serverState2, manifest: manifest2} = generateMockServerState(
+            new MockFileSystemBuilder()
+                .addFile('/index.html', '<script src="bar.hash.js"></script>')
+                .addFile('/bar.hash.js', 'console.log("BAR");')
+                .build());
+
```
gkalpak, line 1781: Let's be consistent with the creation of server states: Either create all server states (including `serverState3`) at the top or create each state right before it is used.

packages/service-worker/worker/test/happy_spec.ts
```diff
@@ -1738,6 +1738,170 @@ describe('Driver', () => {
...
+        expect(await makeRequest(scope, '/foo.hash.js', 'client3')).toBe('console.log("FOO");');
+
+        await driver.initialized;
+        serverState1.clearRequests();
+
+        // Verify that the `foo.hash.js` file is cached.
+        expect(await makeRequest(scope, '/foo.hash.js')).toBe('console.log("FOO");');
+        serverState1.assertNoRequestFor('/foo.hash.js');
+
+        // Update the server to the updated version.
```
gkalpak, line 1798:
```suggestion
        // Update the ServiceWorker to the second version.
```

packages/service-worker/worker/test/happy_spec.ts
```diff
@@ -1738,6 +1738,170 @@ describe('Driver', () => {
...
+        expect(await makeRequest(scope, '/bar.hash.js', 'client1')).toBeNull();
+        serverState2.assertSawRequestFor('/bar.hash.js');
+
+        expect(mockClient1.messages).toEqual([
+          {
+            type: 'UNRECOVERABLE_STATE',
+            reason:
+                'Failed to retrieve hashed resource from the server. (AssetGroup: assets | URL: /bar.hash.js)'
+          },
+        ]);
```
gkalpak, line 1838: We should verify the same for `client2` and also verify that `client3` did not receive any messages.

packages/service-worker/worker/test/happy_spec.ts
```diff
@@ -1738,6 +1738,170 @@ describe('Driver', () => {
...
+
+        // Update the server state to emulate deploying a new version (where `foo.hash.js` does not
+        // exist any more). Keep the cache though.
+        scope = new SwTestHarnessBuilder()
+                    .withCacheState(scope.caches.dehydrate())
+                    .withServerState(updatedServer)
+                    .build();
+        driver = new Driver(scope, scope, new CacheDatabase(scope, scope));
+
+
```
gkalpak, line 1886:
```suggestion
```

gkalpak (changes requested):
Final(?) round of review comments :smiley:

Please squash the existing fixup commits before making the new changes. Also, the commit that adds the test helper should have a type of `test` (`test(service-worker): ...`). BTW, the test helper cannot be used in a commit before the commit that adds it :wink:

The final bit is documenting this in the [ServiceWorker communications][1] guide.

[1]: https://angular.io/guide/service-worker-communications

sonukapoor:
> ...BTW, the test helper cannot be used in a commit before the commit that adds it 😉

😁 Yes, of course.

gkalpak (changes requested):
This is almost ready :tada:
Great job, @sonukapoor 🏆 

FYI, I tried it with angular.io (by manually removing one of the lazy bundles from both the cache and the server and everything seems to work fine 💯).

aio/content/examples/service-worker-getting-started/src/app/unrecoverable-state.service.ts
```diff
@@ -0,0 +1,14 @@
+import { Injectable } from '@angular/core';
+import { SwUpdate } from '@angular/service-worker';
+
+// #docregion sw-unrecoverable-state
+@Injectable()
+export class UnrecoverableStateService {
+
+  constructor(updates: SwUpdate) {
+    updates.unrecoverable.subscribe(event => {
+        document.location.reload();
```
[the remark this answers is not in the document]
gkalpak: Yes, sorry 😇

aio/content/guide/service-worker-communications.md
```diff
@@ -67,6 +67,24 @@ Therefore, it is recommended to reload the page once the promise returned by `ac
...
+- In the meantime, the user's browser decides to evict `lazy-chunk.<lazy-hash-1>.js` from its cache. Browsers may decide to evict specific (or all) resources from a cache in order to reclaim disk space.
+- The user opens the app again. The service worker serves the latest version known to it at this point, namely the old version (`index.html` and `main.<main-hash-1>.js`).
+- At some later point, the app requests the lazy bundle, `lazy-chunk.<lazy-hash-1>.js`.
+- The service worker is unable to find the asset in the cache (remember that the browser evicted it). Nor is it able to retrieve it from the server (since the server now only has `lazy-chunk.<lazy-hash-2>.js` from the newer version).
+
+In the above scenario, the service worker is not able to serve an asset that would normally be cached. That particular app version is broken and there is no way to fix it without reloading the page. In such cases, the service worker notifies the client by sending a `UnrecoverableStateEvent` event. You can subscribe to `SwUpdate#unrecoverable` to be notified and handle these errors.
```
gkalpak, line 83:
```suggestion
In the above scenario, the service worker is not able to serve an asset that would normally be cached. That particular app version is broken and there is no way to fix it without reloading the page. In such cases, the service worker notifies the client by sending an `UnrecoverableStateEvent` event. You can subscribe to `SwUpdate#unrecoverable` to be notified and handle these errors.
```

Also, reading this again I think it would be a little more clear to say something like:

```suggestion
In the above scenario, the service worker is not able to serve an asset that would normally be cached. That particular app version is broken and there is no way to fix the state of the client without reloading the page. In such cases, the service worker notifies the client by sending a `UnrecoverableStateEvent` event. You can subscribe to `SwUpdate#unrecoverable` to be notified and handle these errors.
```

aio/content/guide/service-worker-communications.md
```diff
@@ -67,6 +67,24 @@ Therefore, it is recommended to reload the page once the promise returned by `ac
 
 </div>
 
+### Handling an unrecoverable state
+
+In some cases, the version of the app used by the service worker to serve a client might be in a broken state that cannot be recovered from without a full page reload.
+
+For example, imagine the following scenario:
+- A user opens the app for the first time and the service worker caches the latest version of the app. Let's assume the app's cached assets include `index.html`, `main.<main-hash-1>.js` and `lazy-chunk.<lazy-hash-1>.js`.
```
gkalpak, line 75: I just remembered we are trying to standardize on having one sentence per line in Markdown files. Can you please ensure new sentences are placed on a new line in this section?

```suggestion
- A user opens the app for the first time and the service worker caches the latest version of the app.
  Let's assume the app's cached assets include `index.html`, `main.<main-hash-1>.js` and `lazy-chunk.<lazy-hash-1>.js`.
```

aio/content/examples/service-worker-getting-started/src/app/unrecoverable-state.service.ts
```diff
@@ -0,0 +1,20 @@
+import { Injectable } from '@angular/core';
+import { SwUpdate } from '@angular/service-worker';
+
+function alertUser(message): boolean {
+  return true;
+}
+
+// #docregion sw-unrecoverable-state
+@Injectable()
+export class HandleUnrecoverableStateService {
```
gkalpak, line 10: We should also rename the file to match the service name:

`unrecoverable-state.service.ts` --> `handle-unrecoverable-state.service.ts`

gkalpak (changes requested):
Some minor comments. Otherwise lgtm :tada:

aio/content/guide/service-worker-communications.md
```diff
@@ -67,6 +67,33 @@ Therefore, it is recommended to reload the page once the promise returned by `ac
...
+Browsers may decide to evict specific (or all) resources from a cache in order to reclaim disk space.
+- The user opens the app again.
+The service worker serves the latest version known to it at this point, namely the old version (`index.html` and `main.<main-hash-1>.js`).
+- At some later point, the app requests the lazy bundle, `lazy-chunk.<lazy-hash-1>.js`.
+- The service worker is unable to find the asset in the cache (remember that the browser evicted it).
+Nor is it able to retrieve it from the server (since the server now only has `lazy-chunk.<lazy-hash-2>.js` from the newer version).
+
+In the above scenario, the service worker is not able to serve an asset that would normally be cached.
+That particular app version is broken and there is no way to fix the state of the client without reloading the page.
+In such cases, the service worker notifies the client by sending a `UnrecoverableStateEvent` event.
```
gkalpak, line 91:
```suggestion
In such cases, the service worker notifies the client by sending an `UnrecoverableStateEvent` event.
```

aio/content/guide/service-worker-communications.md
```diff
@@ -67,6 +67,33 @@ Therefore, it is recommended to reload the page once the promise returned by `ac
 
 </div>
 
+### Handling an unrecoverable state
+
+In some cases, the version of the app used by the service worker to serve a client might be in a broken state that cannot be recovered from without a full page reload.
+
+For example, imagine the following scenario:
+- A user opens the app for the first time and the service worker caches the latest version of the app.
+Let's assume the app's cached assets include `index.html`, `main.<main-hash-1>.js` and `lazy-chunk.<lazy-hash-1>.js`.
```
gkalpak, line 76:
```suggestion
  Let's assume the app's cached assets include `index.html`, `main.<main-hash-1>.js` and `lazy-chunk.<lazy-hash-1>.js`.
```

(Below too.)

gkalpak (changes requested):

packages/service-worker/src/low_level.ts
```diff
@@ -33,6 +33,22 @@ export interface UpdateActivatedEvent {
...
 
+/**
+ * An event emitted when the version of the app used by the service worker to serve this client is
+ * in a broken state that cannot be recovered from and a full page reload is required.
+ *
+ * For example, the ServiceWorker may not be able to retrieve a required resource, neither from the
+ * cache nor from the server. This could happen if a new version is deployed to the server and the
+ * ServiceWorker cache has been partially cleaned by the browser, removing some files of a previous
+ * app version but not all.
+ *
```
gkalpak, line 44: I think it is worth adding a link to the [Service worker communication][1] guide (here and in the other events too). Something like:

```suggestion
 *
 * @see {@link guide/service-worker-communications Service worker communication guide}
 *
```

And it's also a good idea to add a similar annotation in the `SwUpdate` class jsdoc.
(Adding these links is a separate change, so feel free to make them on a new commit (either on this PR or a new one).)

[1]: https://angular.io/guide/service-worker-communications

gkalpak (changes requested):

gkalpak (approved):
LGTM :tada:

Great job, @sonukapoor 🏆
I know this wasn't a walk in the park and it took a few iterations, but the end result is quite clean and well worth it. Thank you for seeing it through :smiley:

BTW, we also need a public API review (~~not sure why PullApprove didn't request one already~~ _it did just now_ 🤷‍♂️).

sonukapoor:
> LGTM 🎉
> 
> Great job, @sonukapoor 🏆
> I know this wasn't a walk in the park and it took a few iterations, but the end result is quite clean and well worth it. Thank you for seeing it through 😃

I definitely agree with that. It felt more like a hike in the Himalayas... 😄 
Thank you though for guiding me through this. 
I would have not been complete this without you. 
I have learned a lot from this PR.

atscott (approved):
I haven't looked at the whole progression of the conversation here, but this final state looks really well documented and clear!

LGTM
Reviewed-for: public-api

> BTW, we also need a public API review (~not sure why PullApprove didn't request one already~ _it did just now_ 🤷‍♂️).

PullApprove now waits until initial approvals are there before requesting public API review to reduce noise to that group :)

packages/service-worker/test/comm_spec.ts
```diff
@@ -435,6 +435,14 @@ import {MockPushManager, MockPushSubscription, MockServiceWorkerContainer, MockS
           },
         });
       });
+      it('processes resource removed notifications when sent', done => {
```
petebacondarwin, line 438:
```suggestion
      it('processes unrecoverable notifications when sent', done => {
```

petebacondarwin (approved):
Reviewed-for: public-api

AndrewKushnir (approved):
Reviewed-for: public-api

govi2010:
How can I get this fix? 
I am following this PR which is closed. 
I am using PWA with angular 9. 
How can I get this Fix ?

gkalpak:
@govi2010, this feature was introduced in v11.0.0-next.0. The first stable version that will have it is 11.0.0.

PR #36849, opened 2020-04-29
I18n refactor cherry picks

petebacondarwin:
A set of patch-worthy commits cherry-picked from master.

These commits landed as part of PRs that could only go on master due to new features or breaking changes in the PR. But these specific commits are valid for patch.


petebacondarwin:
Now we are in RC there is little point in cherry picking these commits to 9.x

PR #36850, opened 2020-04-29
feat(common): added generic for HttpErrorResponse

KostyaTretyak:
## PR Checklist

- [ ] Tests for the changes have been added. - **No runtime or API breaking change**
- [ ] Docs have been added / updated. - **No runtime or API breaking change**

## PR Type
- [x] Feature

`HttpErrorResponse` do not have the generic for the error type.

## What is the new behavior?
`HttpErrorResponse` have the generic for the error type.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/common/http/src/response.ts
+++ b/packages/common/http/src/response.ts
@@ -316,10 +316,10 @@ export class HttpResponse<T> extends HttpResponseBase {
  *
  * @publicApi
  */
-export class HttpErrorResponse extends HttpResponseBase implements Error {
+export class HttpErrorResponse<E = any> extends HttpResponseBase implements Error {
   readonly name = 'HttpErrorResponse';
   readonly message: string;
-  readonly error: any|null;
+  readonly error: E|ProgressEvent|ErrorEvent|Error|null;
 
   /**
    * Errors are never okay, even when the status code is in the 2xx success range.
@@ -327,7 +327,7 @@ export class HttpErrorResponse extends HttpResponseBase implements Error {
   readonly ok = false;
 
   constructor(init: {
-    error?: any;
+    error?: E|ProgressEvent|ErrorEvent|Error;
     headers?: HttpHeaders;
     status?: number;
     statusText?: string;
--- a/goldens/public-api/common/http/http.d.ts
+++ b/goldens/public-api/common/http/http.d.ts
@@ -1498,13 +1498,13 @@ export declare interface HttpDownloadProgressEvent extends HttpProgressEvent {
     type: HttpEventType.DownloadProgress;
 }
 
-export declare class HttpErrorResponse extends HttpResponseBase implements Error {
-    readonly error: any | null;
+export declare class HttpErrorResponse<E = any> extends HttpResponseBase implements Error {
+    readonly error: E | ProgressEvent | ErrorEvent | Error | null;
     readonly message: string;
     readonly name = "HttpErrorResponse";
     readonly ok = false;
     constructor(init: {
-        error?: any;
+        error?: E | ProgressEvent | ErrorEvent | Error;
         headers?: HttpHeaders;
         status?: number;
         statusText?: string;


KostyaTretyak:
I just fixed all the CI errors. I'm waiting for the merge.

jelbourn:
@KostyaTretyak is there further context on this PR? Perhaps discussion on an issue?

KostyaTretyak:
@jelbourn, I had my own issue with `HttpErrorResponse`. In my project, the HTTP response from the backend comes with this interface:

```ts
interface ApiResponse<T, M = ObjectAny, E = { message: string }> {
  data?: T[];
  /**
   * Metadata.
   */
  meta?: M;
  error?: E;
}
```

So, in context Angular application, I need to handle the errors as follows:

```ts
const msg = '...';
// ...
this.apiPreAccountService.login('login', this.formGroup.value).subscribe(
  () => {
    // ...
  },
  (err: HttpErrorResponse) => {
    if (err.error) {
      this.message = [err.error.error.message];
      return;
    }
    this.message = [msg];
  }
);
```

As you can see, `err.error` is not yet an object of the standard class `Error`, so I need to pass a data type to make it easier to work with such objects:

```ts
// ...
  (err: HttpErrorResponse<{ message: string }>) => {
    if (err.error) {
      this.message = [err.error.error.message];
      return;
    }
    this.message = [msg];
  }
);

**Comment by jelbourn** [comment]
@gkalpak this seems like a reasonable change to me, but I suspect I may be missing something. Thoughts?

File: packages/common/http/src/response.ts [hunk 1]
```diff
@@ -316,23 +316,19 @@ export class HttpResponse<T> extends HttpResponseBase {
  *
  * @publicApi
  */
-export class HttpErrorResponse extends HttpResponseBase implements Error {
+export class HttpErrorResponse<T = any> extends HttpResponseBase implements Error {
```
gkalpak, line 319: I know @IgorMinar would love to have this default to `unknown`, but that would unfortunately be a breaking change :grin:

gkalpak (approved):
I am not super familiar with `common/http`, but this looks reasonable to me too.
(BTW, I noticed the default value of the generic parameter does not show up correctly in the API docs. This is tracked in #37694.)

jelbourn (approved):
LGTM, though I think it would be good for @alxhub to also take a look

Reviewed-for: public-api

KostyaTretyak:
So, soon we will probably have the release of Angular v11. Could you please merge this PR?

JoostK:
@KostyaTretyak I think there is an issue with having a generic type for this purpose, as an `HttpErrorResponse` is also created for any transport-layer error.

For regular XHR transport it's a `ProgressEvent`:
https://github.com/angular/angular/blob/8a745081307c16644796c333531edcb4591ce235/packages/common/http/src/xhr.ts#L240-L249

The JSONP backend has two different errors, the first one being a generic `Error`:

https://github.com/angular/angular/blob/d1ea1f4c7f3358b730b0d94e65b00bc28cae279c/packages/common/http/src/jsonp.ts#L147-L157

and an actual transport error in the form of `ErrorEvent`:

https://github.com/angular/angular/blob/d1ea1f4c7f3358b730b0d94e65b00bc28cae279c/packages/common/http/src/jsonp.ts#L175-L189

If the server fails to response, or the client's request fails to be sent, or something like that, then you'll end up in this state where `error` is not according to what you typed it to be.

I realise that the generic type arguments in e.g. `HttpClient.get` are generally unsound, as there is no way to verify that the server response is according to what you say it is; but this is totally in control by the developer and the server contract. In the case of errors, I feel this is not the case: if `HttpResponseError.error` were to be of a generic type, the framework would incorrectly document that that is _always_ what's in there, ignoring the fact that it could be a transport-level error of any kind. This is misleading to users, and would ultimately be a bug in the framework.

JoostK (changes requested):
@KostyaTretyak I think there is an issue with having a generic type for this purpose, as an `HttpErrorResponse` is also created for any transport-layer error.

For regular XHR transport it's a `ProgressEvent`:
https://github.com/angular/angular/blob/8a745081307c16644796c333531edcb4591ce235/packages/common/http/src/xhr.ts#L240-L249

The JSONP backend has two different errors, the first one being a generic `Error`:

https://github.com/angular/angular/blob/d1ea1f4c7f3358b730b0d94e65b00bc28cae279c/packages/common/http/src/jsonp.ts#L147-L157

and an actual transport error in the form of `ErrorEvent`:

https://github.com/angular/angular/blob/d1ea1f4c7f3358b730b0d94e65b00bc28cae279c/packages/common/http/src/jsonp.ts#L175-L189

If the server fails to respond, or the client's request fails to be sent, or something like that, then you'll end up in this state where `error` is not according to what you typed it to be.

I realise that the generic type arguments in e.g. `HttpClient.get` are generally unsound, as there is no way to verify that the server response is according to what you say it is; but this is totally in control by the developer and the server contract. In the case of errors, I feel this is not the case: if `HttpResponseError.error` were to be of a generic type, the framework would incorrectly document that that is _always_ what's in there, ignoring the fact that it could be a transport-level error of any kind. This is misleading to users, and would ultimately be a bug in the framework.

KostyaTretyak:
> HttpClient.get are generally unsound, as there is no way to verify that the server response is according to what you say it is; but this is totally in control by the developer and the server contract. In the case of errors...

But we can write it like this:

```ts
HttpErrorResponse<T | ProgressEvent | ErrorEvent>
```

JoostK:
> > HttpClient.get are generally unsound, as there is no way to verify that the server response is according to what you say it is; but this is totally in control by the developer and the server contract. In the case of errors...
> 
> But we can write it like this:
> 
> ```ts
> HttpErrorResponse<CustomError | Error | ProgressEvent | ErrorEvent>
> ```

Not really, no. If you were to do something like that, it should not be generic type constraint but rather part of the union type in the declaration of `HttpErrorResponse.error`. However, the various backends could all have varying error types, so enumerating all types that are possible does not work for other backend implementations that may have different error behavior. To properly support that, there could be a `TransportError` or something like that:

```ts
export class HttpErrorResponse<E> {
  error: E | TransportError | null;
}

export class TransportError {
  error: unknown;
}
```

which forces consumers to account for the `TransportError` instead of incorrectly assuming it's always `E`. This would however be quite a breaking change.

KostyaTretyak:
Ok, I've made some changes in `HttpErrorResponse` type declaration:

```ts
readonly error: E | ProgressEvent | ErrorEvent | Error;
```

And now we can do this:
```ts
type TransportError = { someProperty: unknown }; // Here a list of desired error properties.
type CustomError = { error: { message: string } };

let err: HttpErrorResponse<CustomError | TransportError>;
```

> This would however be quite a breaking change.

The tests passed without errors, and there is a default value `any` for the generic. Why will there be a breaking change?

JoostK:
> Ok, I've made some changes in `HttpErrorResponse` type declaration:
> 
> ```ts
> readonly error: E | ProgressEvent | ErrorEvent | Error;
> ```
> 
> And now we can do this:
> 
> ```ts
> type TransportError = { someProperty: unknown }; // Here a list of desired error properties.
> type CustomError = { error: { message: string } };
> 
> let err: HttpErrorResponse<CustomError | TransportError>;
> ```
> 
> > This would however be quite a breaking change.
> 
> The tests passed without errors, and there is a default value `any` for the generic. Why will there be a breaking change?

That is not quite what I was suggesting. The `TransportError` would be a wrapper object that captures the `unknown` error (as its shape really isn't known, not even that it would be constrained to `ProgressEvent | ErrorEvent | Error` as that is entirely dependent on the backend) and then let `HttpErrorResponse.errors` be of type `E | TransportError | null`. This is breaking, as transport errors would then no longer be `response.error`, but rather `response.error.error` (after having tested that `response.error` is indeed a `TransportError`).

Note also that the generic type does not affect this `TransportError` type, it would always be there regardless of the generic type provided by a user. So you'd still be required to test for `TransportError` _or_ cast it in an unsafe manner as you would today.

petebacondarwin:
I think that the issues raised by @JoostK are enough to avoid making this change. In reality, since the class is not being created by the application developer, there seems to be little benefit in making it generic just to narrow the type when it appears in a callback.  This can be done just as easily by creating your own sub-type:

```ts
http.get("blah blah").subscribe(
  () => {},
  (err: MyErrorResponse<{ message: string }>) => {
    console.log(err.error.message);
  }
);
...
interface MyErrorResponse<E> extends HttpErrorResponse {
  error: E;
}
```

See https://stackblitz.com/edit/angular-ivy-tdwma9

KostyaTretyak:
Ok, thank you.

PR #34533, opened 2019-12-22
fix: should also allow subclass Promise without Symbol.species

JiaLiPassion:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] angular.io application / infrastructure changes
- [ ] Other... Please describe:

## What is the current behavior?

Issue Number: #34162

This is a regression issue introduced by #34162, before that PR, we call `this.constructor` in `Promise.then()` to create a new Promise.
In the #34162 PR, ZoneAwarePromise support to subclass Promise by Symbol.species.
But we still need to support subclass without Symbol.species implemented.

## What is the new behavior?
If the subclass doesn't have Symbol.species, we should use `this.constructor` instead of hardcoded `ZoneAwarePromise`.

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

PR #36851, opened 2020-04-29
fix(zone.js): remove unused Promise overwritten logic

JiaLiPassion:
In the early Zone.js versions (< 0.10.3), `ZoneAwarePromise` did not support `Symbol.species`,
so when user used a 3rd party `Promise` such as `es6-promise`, and try to load the promise library after import of `zone.js`, the loading promise library will overwrite the patched `Promise` from `zone.js` and will break `Promise` semantics with respect to `zone.js`.

Starting with `zone.js` 0.10.3, `Symbol.species` is supported therefore this will not longer be an issue. (https://github.com/angular/angular/pull/34533)

Before 0.10.3, the logic in zone.js tried to handle the case in the wrong way. It did so by overriding the descriptor of `global.Promise`, to allow the 3rd party libraries to override native `Promise`
instead of `ZoneAwarePromise`. This is not the correct solution, and since the `Promise.species` is now supported, the 3rd party solution of overriding `global.Promise` is no longer needed.

PR removes the wrong work around logic. (This will improve the bundle size.)

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/goldens/size-tracking/integration-payloads.json
+++ b/goldens/size-tracking/integration-payloads.json
@@ -3,8 +3,8 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 1485,
-        "main-es2015": 141394,
-        "polyfills-es2015": 36963
+        "main-es2015": 141151,
+        "polyfills-es2015": 36571
       }
     }
   },
@@ -22,7 +22,7 @@
       "uncompressed": {
         "runtime-es2015": 1485,
         "main-es2015": 147314,
-        "polyfills-es2015": 36963
+        "polyfills-es2015": 36571
       }
     }
   },
--- a/packages/zone.js/lib/common/promise.ts
+++ b/packages/zone.js/lib/common/promise.ts
@@ -459,44 +459,6 @@ Zone.__load_patch('ZoneAwarePromise', (global: any, Zone: ZoneType, api: _ZonePr
   ZoneAwarePromise['all'] = ZoneAwarePromise.all;
 
   const NativePromise = global[symbolPromise] = global['Promise'];
-  const ZONE_AWARE_PROMISE = Zone.__symbol__('ZoneAwarePromise');
-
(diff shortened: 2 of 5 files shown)


mhevery (approved):
Please update the commit message to:
```
In the early Zone.js versions (< 0.10.3), `ZoneAwarePromise` did not support `Symbol.species`,
so when user used a 3rd party `Promise` such as `es6-promise`, and try to load the promise library after import of `zone.js`, the loading promise library will overwrite the patched `Promise` from `zone.js` and will break `Promise` semantics with respect to `zone.js`.

Starting with `zone.js` 0.10.3, `Symbol.species` is supported therefore this will not longer be an issue. (https://github.com/angular/angular/pull/34533)

Before 0.10.3, the logic in zone.js tried to handle the case in the wrong way. It did so by overriding the descriptor of `global.Promise`, to allow the 3rd party libraries to override native `Promise`
instead of `ZoneAwarePromise`. This is not the correct solution, and since the `Promise.species` is now supported, the 3rd party solution of overriding `global.Promise` is no longer needed.

PR removes the wrong work around logic. (This will improve the bundle size.)
```

JiaLiPassion:
@mhevery , got it, thank you, updated the commit messages.

AndrewKushnir:
Hi @JiaLiPassion, FYI I've started a [presubmit](http://test/OCL:315992030:BASE:315992084:1591915066621:3cc94673).

It looks like the [`test_ivy_aot` CI job is failing](https://app.circleci.com/pipelines/github/angular/angular/16898/workflows/d0d5dcef-ec84-412f-8159-02bc62394d65/jobs/736766/steps) due to the polyfills size limit change (actually size decrease, yay). Could you please update the payload limits to reflect that?

Thank you.

mhevery:
Failures seem unrelated.

PR #36855, opened 2020-04-29
build: prepare for TypeScript 3.9

alan-agius4:
- Fix several compilation errors
- Update @microsoft/api-extractor to be compatible with TypeScript 3.9


packages/service-worker/BUILD.bazel
```diff
@@ -59,12 +59,14 @@ ng_package(
...
-ts_api_guardian_test_npm_package(
-    name = "service-worker_api",
-    actualDir = "angular/packages/service-worker/npm_package",
-    data = [
-        ":npm_package",
-        "//goldens:public-api",
-    ],
-    goldenDir = "angular/goldens/public-api/service-worker",
-)
+# Currently this fails with 'Error: Symbol "Generator_2" was aliased as "Generator". Aliases are not supported."' from ts-api-guardian
```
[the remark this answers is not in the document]
alan-agius4: I am also seeing the same problem for other API tests.

packages/compiler-cli/src/ngtsc/file_system/testing/src/test_helper.ts
```diff
@@ -75,7 +75,7 @@ export function initMockFileSystem(os: string, cwd?: AbsoluteFsPath): void {
 function createMockFileSystem(os: string, cwd?: AbsoluteFsPath): MockFileSystem {
   switch (os) {
     case 'OS/X':
-      return new MockFileSystemPosix(/* isCaseSensitive */ false, cwd);
+      return new MockFileSystemPosix(/* isCaseSensitive */ true, cwd);
```
petebacondarwin, line 78: This is not correct. OS/X by default is case-insensitive, although it doesn't have to be.
In any case, changing this to true, makes it identical to the `Unix` FS below, which doesn't add any value.

goldens/public-api/core/core.d.ts
```diff
@@ -313,7 +313,7 @@ export declare class ErrorHandler {
     handleError(error: any): void;
 }
 
-export declare class EventEmitter<T extends any> extends Subject<T> {
+export declare class EventEmitter<T = any> extends Subject<T> {
```
alan-agius4, line 316: This is done, so that we don't have a breaking change. Because in TS 3.9 `T extends any` behaviour changed

```
function foo<T extends any>(arg: T) {
    arg.spfjgerijghoied; // no error!
}

function foo<T extends any>(arg: T) {
    arg.spfjgerijghoied;
    //  ~~~~~~~~~~~~~~~
    // Property 'spfjgerijghoied' does not exist on type 'T'.
}
```

`T extends any` is more or less the same as `T extends unknown`

packages/compiler-cli/ngcc/index.ts
```diff
@@ -8,15 +8,15 @@
...
 export {ConsoleLogger} from './src/logging/console_logger';
 export {Logger, LogLevel} from './src/logging/logger';
 export {AsyncNgccOptions, NgccOptions, PathMappings, SyncNgccOptions} from './src/ngcc_options';
 
-export function process(options: AsyncNgccOptions): Promise<void>;
-export function process(options: SyncNgccOptions): void;
-export function process(options: NgccOptions): void|Promise<void> {
+export function process<T extends AsyncNgccOptions|SyncNgccOptions>(options: T):
+    T extends AsyncNgccOptions ? Promise<void>: void;
+export function process(options: AsyncNgccOptions|SyncNgccOptions): void|Promise<void> {
```
gkalpak, line 19: IMO, this would be more readable as:

```suggestion
export function process<T extends NgccOptions>(options: T):
    T extends AsyncNgccOptions ? Promise<void>: void {
```

Same for `mainNgcc()`. Did I miss something?

alan-agius4: There are not the same.

https://www.typescriptlang.org/v2/en/play?ts=3.9.0-beta#code/GYVwdgxgLglg9mABAUwB7IiKyAUBKALkQDc4YATAHwAUAnOAWxgGdkAeUigPkQG8AofgF9BaAA5xaURDDDZawAIYRkiAMoBPSADkA5hAgB5MbATM+-RFcSLmWiAH4iSgDatriYaNQSpMucgKyqoAgnY6+kYm8GDmAh629kRQtCCqVl78APRZiADqkgDWzPziktKgkKZIYvQqzMxsACooqNhg5OZh9noGxtXMlJoRfdFmXDhwY7FETYSILWjtnYjdI1EDiA6IdIws7JzkXESHANylPuWIldAxiLVw9cyT08xEaxC9GzGDw5+R-R+80ONHoTFYHDIRwsHloyCgIFoSDQGCwuDw5xE2VyAFFaPRaBdfBVwLcEPc6sgGgB9ZqtZZdcL-UYDIY9AGvCZTAazeaLNrIDqM9ksn5bHZg-aQ7gnKEw6xwhFI1qo7D4TGCIA

gkalpak: Hm...it would work for `process()` is `mainNgcc()` was also updated to the same signature (as I suggested above). But I now realize that we would need to make changes in `mainNgcc()` internally to satisfy the return new type.

So, let's keep it as is :grin:

alan-agius4:
There seems to be 2 main pending failure points.

**NGCC**
- @petebacondarwin, offered his help to find the root cause and solution

**LS**
- Found the root cause and will laise with @kyliau

alan-agius4:
Closing in favour of https://github.com/angular/angular/pull/36989

PR #36088, opened 2020-03-16
Bazelisk window ci [DO NOT MERGE]

gregmagolan:
Windows CI for https://github.com/angular/angular/pull/36078

PR #36859, opened 2020-04-29
fix(compiler-cli): ensure FS case-sensitivity is correct

petebacondarwin:
The `NodeJSFileSystem.isCaseSensitive()` method was computing
completely the opposite value for its result.

Also the `NgtscCompilerHost.getCanonicalFileName()` method was
not actually calling `useCaseSensitiveFileNames()` so it was always
assuming that the filenames were case sensitive and not actually
canonicalizing them. Instead just returning the original.

Once these two bugs were fixed, a number of issues with the rest of the
handling of case-insensitive files in compiler-cli and tests were uncovered.

This commit also fixes these issues. Here are some further notes:

The TS compiler canonicalizes filenames internally for lookups etc. So
if we are also doing file lookups we need to use the same canonicalized
forms. This is achieved by injecting the `ts.CompilerHost` where needed
and calling `getCanonicalFileName()`. For example, when computing the
`rootDirs` in the `LogicalFileSystem`.

Some tests were checking that particular source files were being exposed
by comparing the `ts.SourceFile.fileName` property with some non-canonical
filename. In some cases this would fail. Instead these comparisons are
achieved by directly comparing the SourceFile objects, if known, or getting
the SourceFile object from a non-canonical filename from the program, by
calling `program.getSourceFile(nonCanonicalFileName)`.

The `MockFileSystemWindows` class was not considering lower case drive names
when computing if a path is rooted (i.e. absolute).

The `NodeJSFileSystem.isCaseSensitive()` method was relying upon its `exists()`
method. But this itself relies upon the `isCaseSensitive()` to determine
if a file exists. So there was an infinite recursion. This is solved by
actually using the underlying real `fs.existsSync` to compute whether the
file system is case sensitive.

---

This was tested to work on Windows via #36873. See https://app.circleci.com/pipelines/github/angular/angular/14310/workflows/b347d457-f0ce-4d54-afbe-413a1c481003

### DIFF
--- a/packages/compiler-cli/src/ngtsc/file_system/src/logical.ts
+++ b/packages/compiler-cli/src/ngtsc/file_system/src/logical.ts
@@ -53,10 +53,13 @@ export class LogicalFileSystem {
    */
   private cache: Map<AbsoluteFsPath, LogicalProjectPath|null> = new Map();
 
-  constructor(rootDirs: AbsoluteFsPath[]) {
+  constructor(rootDirs: AbsoluteFsPath[], private compilerHost: ts.CompilerHost) {
     // Make a copy and sort it by length in reverse order (longest first). This speeds up lookups,
     // since there's no need to keep going through the array once a match is found.
-    this.rootDirs = rootDirs.concat([]).sort((a, b) => b.length - a.length);
+    this.rootDirs =
+        rootDirs.map(dir => this.compilerHost.getCanonicalFileName(dir) as AbsoluteFsPath)
+            .concat([])
+            .sort((a, b) => b.length - a.length);
   }
 
   /**
@@ -76,11 +79,13 @@ export class LogicalFileSystem {
    * of the TS project's root directories.
    */
   logicalPathOfFile(physicalFile: AbsoluteFsPath): LogicalProjectPath|null {
-    if (!this.cache.has(physicalFile)) {
+    const canonicalFilePath =
+        this.compilerHost.getCanonicalFileName(physicalFile) as AbsoluteFsPath;
+    if (!this.cache.has(canonicalFilePath)) {
       let logicalFile: LogicalProjectPath|null = null;
       for (const rootDir of this.rootDirs) {
-        if (physicalFile.startsWith(rootDir)) {
-          logicalFile = this.createLogicalProjectPath(physicalFile, rootDir);
+        if (isWithinBasePath(rootDir, canonicalFilePath)) {
+          logicalFile = this.createLogicalProjectPath(canonicalFilePath, rootDir);
           // The logical project does not include any special "node_modules" nested directories.
           if (logicalFile.indexOf('/node_modules/') !== -1) {
             logicalFile = null;
@@ -89,9 +94,9 @@ export class LogicalFileSystem {
           }
         }
       }
-      this.cache.set(physicalFile, logicalFile);
+      this.cache.set(canonicalFilePath, logicalFile);
     }
-    return this.cache.get(physicalFile)!;
+    return this.cache.get(canonicalFilePath)!;
   }
 
   private createLogicalProjectPath(file: AbsoluteFsPath, rootDir: AbsoluteFsPath):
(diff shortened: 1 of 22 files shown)


alxhub (approved):

AndrewKushnir (approved):
LGTM 👍

(CI errors look legit, so I've added the "cleanup" label)

petebacondarwin:
Indeed this fix has revealed a bunch of issues with our Windows support!!

AndrewKushnir:
@petebacondarwin just to be safe (and make sure everything works as expected on Windows) it might make sense to run `test_win` and `test_ivy_aot_win` CircleCI jobs. You can enable them in a separate commit in this PR (and revert it later) or create a separate PR (this is what Greg usually does for Bazel-relaated changes), and enable the mentioned CircleCI jobs [for example as shown here](https://github.com/angular/angular/pull/36088/files?file-filters%5B%5D=.yml#diff-1d37e48f9ceff6d8030570cd36286a61L876).

petebacondarwin:
Yes definitely

packages/compiler-cli/src/ngtsc/file_system/src/logical.ts
```diff
@@ -76,11 +79,13 @@ export class LogicalFileSystem {
...
   logicalPathOfFile(physicalFile: AbsoluteFsPath): LogicalProjectPath|null {
-    if (!this.cache.has(physicalFile)) {
+    const canonicalFilePath =
+        this.compilerHost.getCanonicalFileName(physicalFile) as AbsoluteFsPath;
+    if (!this.cache.has(canonicalFilePath)) {
       let logicalFile: LogicalProjectPath|null = null;
       for (const rootDir of this.rootDirs) {
-        if (physicalFile.startsWith(rootDir)) {
-          logicalFile = this.createLogicalProjectPath(physicalFile, rootDir);
+        if (!relative(rootDir, canonicalFilePath).startsWith('..')) {
```
AndrewKushnir, line 87: It looks like there is a similar check in this file that uses `startsWith('../')` condition. I'd propose extracting it into a separate utility function, which would also serve as extra documentation.

packages/compiler-cli/src/ngtsc/file_system/src/node_js_file_system.ts
```diff
@@ -68,7 +68,7 @@ export class NodeJSFileSystem implements FileSystem {
   }
   isCaseSensitive(): boolean {
     if (this._caseSensitive === undefined) {
-      this._caseSensitive = this.exists(togglePathCase(__filename));
+      this._caseSensitive = !fs.existsSync(togglePathCase(__filename));
```
AndrewKushnir, line 71: There is a note on this change in PR description, but it might make sense to put it here as well (as an inline comment), since using `this.exists` seems like a natural thing to do, but switching it back will result in errors.

packages/compiler-cli/src/ngtsc/file_system/testing/src/mock_file_system_windows.ts
```diff
@@ -34,14 +34,14 @@ export class MockFileSystemWindows extends MockFileSystem {
   }
 
   isRooted(path: string): boolean {
-    return /^([A-Z]:)?([\\\/]|$)/i.test(path);
+    return /^([A-Za-z]:)?([\\\/]|$)/i.test(path);
```
AndrewKushnir, line 37: Just curious if this change is needed, since the RegExp has the `i` flag?

AndrewKushnir (approved):
LGTM, just left a couple nits.

packages/compiler-cli/src/ngtsc/file_system/testing/src/mock_file_system_windows.ts
petebacondarwin: Doh! Of course.

packages/compiler-cli/src/ngtsc/file_system/src/logical.ts
petebacondarwin: Actually the two uses of `startsWith('..')` cannot easily be combined into a single function. I will change this usage to call a function.

IgorMinar:
@petebacondarwin can you please rebase this? thanks!

petebacondarwin:
Rebased...

alxhub:
@petebacondarwin this didn't merge cleanly to patch :( please open a separate PR if you feel like it should be merged there.

PR #36862, opened 2020-04-29
build: use sh instead of exec for release scripts

josephperrott:
Previously the exec command was used, however the exec command would
exit the original calling script regardless of the whether exit was
called.  This caused the release script to always exit after the
pre-check phase.


AndrewKushnir (approved):
LGTM, thanks for the fix @josephperrott!

AndrewKushnir:
We decided to proceed with this PR, since it's a minor change which blocks the release.

IgorMinar:
Post submit lgtm :)

PR #36864, opened 2020-04-29
build: Add entry point for Ivy language service

kyliau:
This commit adds a new entry point for the Ivy version of language
service. The entry point is just a shell for now, implementation will be
added in subsequent PRs.

The Ivy version of language service could be loaded from the NPM package
via `require(@angular/language-service/bundles/ivy.umd.js)`

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/language-service/bundles/BUILD.bazel
+++ b/packages/language-service/bundles/BUILD.bazel
@@ -18,6 +18,21 @@ ls_rollup_bundle(
     ],
 )
 
+ls_rollup_bundle(
+    name = "ivy",
+    entry_point = "//packages/language-service/ivy:ts_plugin.ts",
+    globals = {
+        "fs": "fs",
+        "path": "path",
+        "typescript/lib/tsserverlibrary": "ts",
+    },
+    license_banner = ":banner",
+    visibility = ["//packages/language-service:__pkg__"],
+    deps = [
+        "//packages/language-service/ivy",
+    ],
+)
+
 genrule(
     name = "banner",
     srcs = ["banner.js"],
--- a/packages/language-service/BUILD.bazel
+++ b/packages/language-service/BUILD.bazel
@@ -35,5 +35,6 @@ pkg_npm(
         ":language-service",
         # min bundle is not used at the moment; omit from package to speed up build
         "//packages/language-service/bundles:language-service.umd.js",
+        "//packages/language-service/bundles:ivy.umd.js",
     ],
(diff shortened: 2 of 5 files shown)


ayazhafiz (approved):
🎉

packages/language-service/bundles/BUILD.bazel
```diff
@@ -18,6 +18,21 @@ ls_rollup_bundle(
     ],
 )
 
+ls_rollup_bundle(
+    name = "ivy",
+    entry_point = "//packages/language-service/ivy:ts_plugin.ts",
+    globals = {
+        "fs": "fs",
+        "path": "path",
+        "typescript/lib/tsserverlibrary": "ts",
```
devversion, line 27: Do you know why this is `ts` and the other non-ivy bundle uses `tss`?

devversion (approved):

packages/language-service/bundles/BUILD.bazel
ayazhafiz: In the other bundle we use `ts` to refer to the `typescript` package and `tss` to refer to the `typescript/lib/tsserverlibrary` package.

devversion: Yeah, but shouldn't this be `tss` then? (so that it is consistent; `ts` would better suit for `typescript`)

ayazhafiz: @kyliau

kyliau: In our current implementation (non-ivy bundle) some code imports `typescript` whereas some imports `typescript/lib/tsserverlibrary`. We need to support and disambiguate both, hence the use of `tss` for the latter.
In the new implementation (ivy bundle), I'd like to converge on just tsserverlibrary, so `ts` is sufficient. tsserverlibrary is a superset of typescript, it contain additional stuff in the `ts.server` namespace.

devversion: That makes sense. To be clear: I don't think it's a big deal what the non-public facing namespace is called in the AMD import, so I'm fine either way 😄 

At the same time though, it would be consistent and matching with the actual _module name_ if we'd use `tss`.

PR #26428, opened 2018-10-13
docs(docs-infra): refactor use of media queries

carlosasj:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

```
[ ] Bugfix
[ ] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Documentation content changes
[x] angular.io application / infrastructure changes
[ ] Other... Please describe:
```

## What is the current behavior?
Screen widths close to breakpoints sometimes presented the layout with a mixture of styles from 2 different "screen sizes" (small and medium, or tiny and small...).
This occur because there is many media queries with inline values.

Also, for screen widths smaller than 900px there was a horizontal scroll on the home page.

## What is the new behavior?
This pull request aims to provide a consistent and semantic way to use media queries to avoid strange behaviors close to breakpoints.

Also, changes some values to remove the horizontal scroll on the home page for small screens.

## Does this PR introduce a breaking change?
```
[ ] Yes
[x] No
```

## Other information
There is 2 commits in this PR. The first one do the "obvious" work: it creates the mixins and apply them to media queries with values close to a breakpoint. The second one apply the mixins to media queries that are not so close to a breakpoint value, so I had to look for which screen size would be a better fit to the value inline.

PR #36869, opened 2020-04-30
feat(docs-infra): refactor use of media queries

ajitsinghkaler:
Building on the work done by @carlosasj. #26428

Change the use of media queries from inline values to a mixin with constants.
The motivation was to add consistency between media queries and avoid
gaps of 1px where the layout presented have mixed styles from two breakpoints.
Also, remove redundant CSS attributes.

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Media queries inlined using constants

Issue Number: N/A

## What is the new behavior?
Use mixins for media queries

- [ ] Yes
- [x] No

### DIFF
--- a/aio/src/styles/_constants.scss
+++ b/aio/src/styles/_constants.scss
@@ -123,6 +123,22 @@ $api-symbols: (
 );
 
 // OTHER
+
+// |0px                             |320px           |480px       |600px   |650px          |800px              |993px
+// |--------------------------------|----------------|------------|--------|---------------|-------------------|------- -  -   -
+// |             xxsmall            |     xsmall     |    teeny   |  tiny  |     small     |       medium      |   big
+// |________________________________|________________|____________|________|_______________|___________________|_______ _  _   _
+//
+// pattern:
+// $breakpoint-<size>: <minimum value that is inside that size>;
+$breakpoint-xxsmall: 0;
+$breakpoint-xsmall: 320px;
+$breakpoint-teeny: 480px;
+$breakpoint-tiny: 600px;
+$breakpoint-small: 650px;
+$breakpoint-medium: 800px;
+$breakpoint-big: 993px;
+
 $small-breakpoint-width: 840px;
-$phone-breakpoint: 480px;
-$tablet-breakpoint: 800px;
+$phone-breakpoint: $breakpoint-teeny;
+$tablet-breakpoint: $breakpoint-medium;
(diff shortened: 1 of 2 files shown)


ajitsinghkaler:
At the moment only added the mixins to the mixins file after that I can create a new commit for 4 new SCSS files every time. So that it is easy for @petebacondarwin to review too.

Or should I change all breakpoints at once

@petebacondarwin thoughts

ajitsinghkaler:
@petebacondarwin can you please have a look

petebacondarwin:
This change looks OK to me but this is not an area that I have much knowledge over. So I have requested reviews from @gkalpak and @sjtrimble, who know more than me.

sjtrimble:
My question is ... do we actually need all of these breakpoints? We really only differentiate between phone, tablet, desktop (and maybe extra large desktop). Can or should we simplify it to be more tailored to our current needs?

Otherwise makes sense so long as people know to utilize the `mixin` instead of the standard media query. Might need @gkalpak help in making sure it all makes sense :)

gkalpak:
Here are my random thoughts:

- :+1: for having less breakpoints. At the very least, we should start by converting our current media queries to use the mixin (so have as many break points as we use in media queries) and then try to consolidate to 4 breakpoints (if we have more than that currently).

- BTW, we already have a breakpoint mixin. Why not build on top of that? https://github.com/angular/angular/blob/4f2d1570dc3291c9cdca0fb30aff79ba96644b17/aio/src/styles/_mixins.scss#L130-L157

- Another thing we need to decide is whether to take a "mobile first" or "desktop  first" approach. I.e. have the default styles be for mobile or desktop and that have breakpoints for styling on bigger/smaller screens. IMO, since the bulk of our traffic comes from desktop, it makes sense to make larger screens our primary target and adjust styles for smaller screens (but it might turn out to be more convenient to do it the other way around - not sure).

- Finally, I don't know of a way to enforce the use of the mixin for media queries (other than via reviews) 😞
  We could probably put together a crude script that goes through `.scss` files and checks, but maybe there is some more sophisticated tool for linting Sass :grin:

ajitsinghkaler:
@gkalpak @petebacondarwin okay I'll start changing all breakpoints to mixins with 4 files at a time so that it is easy to review  too

petebacondarwin:
@ajitsinghkaler - are you going to make the changes to this PR or should we close it?

ajitsinghkaler:
@petebacondarwin I am working on this

ajitsinghkaler:
Actually its just a very large effort

PR #30397, opened 2019-05-10
fix(common): fix immutability for locale_data_api

artem-galas:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] angular.io application / infrastructure changes
- [ ] Other... Please describe:

## What is the current behavior?
Since no new array instance was returned (on `getLocaleMonthNames` / `getLocaleDayNames` methods), after manipulation (e.q `splice`, `push`) on array it's not possible to get full list of localized days/month.

Issue Number: #27003

## What is the new behavior?
Add `Array.from` to  return value from these methods. 
Add general tests for both methods (they didn't have any before), and also add tests for testing bugfix  

## Does this PR introduce a breaking change?

- [x] Yes
- [ ] No

PR #36875, opened 2020-04-30
fix(common): returnvalues should be readonly

Goobles:
## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

getLocaleDayNames returns a mutable array which can be modified, when getLocaleDayNames is called again, the mutated array is returned.

A common mistake is this: 
getLocaleDayNames returns an array starting with the American standard, with sunday first.
To change this to our locale standard I would do this:
```
const weekdays = getLocaleDayNames('en', FormStyle.Format, TranslationWidth.Narrow)
weekdays.push(weekdays.shift()); // push Sunday to the back of the array
```
this mutates the array unexpectedly and as a result the weekdays would cycle every time my code was called.

## What is the new behavior?
getLocaleDayNames returns a readonly array.

- [x] Yes
- [ ] No

Mutations to the returnvalue of getLocaleDayNames / getLocaleMonthNames should first be cloned before mutating.

### DIFF
--- a/packages/common/src/i18n/locale_data_api.ts
+++ b/packages/common/src/i18n/locale_data_api.ts
@@ -255,7 +255,7 @@ export function getLocaleDayPeriods(
  * @publicApi
  */
 export function getLocaleDayNames(
-    locale: string, formStyle: FormStyle, width: TranslationWidth): string[] {
+    locale: string, formStyle: FormStyle, width: TranslationWidth): ReadonlyArray<string> {
   const data = ɵfindLocaleData(locale);
   const daysData =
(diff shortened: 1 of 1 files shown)


petebacondarwin:
This is a change to our public API so we need to update the public API golden files. Please run:

```
yarn bazel run //packages/common:common_api.accept
```

And then commit the changed files.

See https://app.circleci.com/pipelines/github/angular/angular/14432/workflows/15f374a4-acdb-4cd6-9ddb-ebaa98b27e76/jobs/698673/steps

petebacondarwin:
This was actually fixed more comprehensively in https://github.com/angular/angular/pull/30397
Thanks for the PR @Goobles.

PR #36891, opened 2020-05-01
docs: update deprecation guide

kara:


IgorMinar (commented):
don't we need to run these migrations one last time in v10 to ensure that any regressions added since the last time the migration ran are updated?

kara:
Yes, we do. Not sure how that relates to this change though?

IgorMinar:
I'm not sure if I'm missing something, but here is my thinking:

If we plan on running this migration in v10, then we need to keep the docs for the migration in v10 because `ng update` will try to link to the docs on angular.io. If you delete the docs the then links from `ng update` will be broken.

Additionally if we are running this migration in v10 then users will likely want to know about this migration when reading the release notes that will state that these APIs have been removed and this migrations were used to migrate people away from them. If we delete the migration docs, we won't have anything to link to except for that archived v9.angular.io version which is odd as the migration is still present and active in v10.

Is there something else at play that I'm missing?

In short, I think we should keep the migration docs and update them with info about these APIs being removed in v10.

kara:
Ah, I think I see the misunderstanding. This PR isn't removing any migration guides. It just removes the blurbs in the deprecation guide, which duplicate the existing migration guides. You can see in the removal table that the guides are still linked.

IgorMinar (approved):
ah.. I see. thanks for the info @kara, I must have misinterpreted the commit message you wrote.

In that case this change LGTM. thanks!

PR #36897, opened 2020-05-02
refactor(localize): use the new workspaces API for ng-add schematic

SchnWalter:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Updates the `ng add @angular/localize` to use the new Workspaces API.

Issue Number: N/A

Related to PR https://github.com/angular/angular-cli/pull/17597 and https://github.com/angular/components/pull/19229

## What is the new behavior?

As the title describes, this PR updates the schematic to use the new workspaces API.

- [ ] Yes
- [x] No

## Other information

This PR is a followup for https://github.com/angular/angular-cli/pull/17597 where I've discovered that there are some unused methods in `@schematics/angular` and removed them.

cc: @alan-agius4

### DIFF
--- a/packages/localize/schematics/ng-add/index.ts
+++ b/packages/localize/schematics/ng-add/index.ts
@@ -8,55 +8,72 @@
  * @fileoverview Schematics for ng-new project that builds with Bazel.
  */
 
-import {virtualFs} from '@angular-devkit/core';
-import {chain, Rule, Tree} from '@angular-devkit/schematics';
-import {getWorkspace} from '@schematics/angular/utility/config';
-import {getProjectTargets} from '@schematics/angular/utility/project-targets';
-import {validateProjectName} from '@schematics/angular/utility/validation';
-import {BrowserBuilderTarget, Builders, ServeBuilderTarget} from '@schematics/angular/utility/workspace-models';
+import {virtualFs, workspaces} from '@angular-devkit/core';
+import {chain, Rule, SchematicsException, Tree} from '@angular-devkit/schematics';
+import {getWorkspace} from '@schematics/angular/utility/workspace';
+import {Builders} from '@schematics/angular/utility/workspace-models';
 
 import {Schema} from './schema';
 
 
 export const localizePolyfill = `import '@angular/localize/init';`;
 
-function getAllOptionValues<T>(
-    host: Tree, projectName: string, builderName: string, optionName: string) {
-  const targets = getProjectTargets(host, projectName);
+function getRelevantTargetDefinitions(
+    project: workspaces.ProjectDefinition, builderName: Builders): workspaces.TargetDefinition[] {
+  const definitions: workspaces.TargetDefinition[] = [];
+  project.targets.forEach((target: workspaces.TargetDefinition): void => {
+    if (target.builder === builderName) {
+      definitions.push(target);
+    }
+  });
+  return definitions;
+}
 
-  // Find all targets of a specific build in a project.
-  const builderTargets: (BrowserBuilderTarget|ServeBuilderTarget)[] = Object.values(targets).filter(
-      (target: BrowserBuilderTarget|ServeBuilderTarget) => target.builder === builderName);
-
-  // Get all options contained in target configuration partials.
-  const configurationOptions = builderTargets.filter(t => t.configurations)
-                                   .map(t => Object.values(t.configurations!))
-                                   .reduce((acc, cur) => acc.concat(...cur), []);
-
-  // Now we have all option sets. We can use it to find all references to a given property.
-  const allOptions = [
-    ...builderTargets.map(t => t.options),
-    ...configurationOptions,
-  ];
-
-  // Get all values for the option name and dedupe them.
-  // Deduping will only work for primitives, but the keys we want here are strings so it's ok.
-  const optionValues: T[] =
-      allOptions.filter(o => o[optionName])
-          .map(o => o[optionName])
-          .reduce((acc, cur) => !acc.includes(cur) ? acc.concat(cur) : acc, []);
-
+function getOptionValuesForTargetDefinition(
+    definition: workspaces.TargetDefinition, optionName: string): string[] {
+  const optionValues: string[] = [];
+  if (definition.options && optionName in definition.options) {
+    let optionValue: unknown = definition.options[optionName];
+    if (typeof optionValue === 'string') {
+      optionValues.push(optionValue);
+    }
+  }
+  if (!definition.configurations) {
+    return optionValues;
+  }
+  Object.values(definition.configurations)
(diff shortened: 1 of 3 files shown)


packages/localize/schematics/ng-add/index.ts
```diff
@@ -8,53 +8,67 @@
...
 
-import {virtualFs} from '@angular-devkit/core';
-import {chain, Rule, Tree} from '@angular-devkit/schematics';
-import {getWorkspace} from '@schematics/angular/utility/config';
-import {getProjectTargets} from '@schematics/angular/utility/project-targets';
-import {validateProjectName} from '@schematics/angular/utility/validation';
-import {BrowserBuilderTarget, Builders, ServeBuilderTarget} from '@schematics/angular/utility/workspace-models';
+import {virtualFs, workspaces} from '@angular-devkit/core';
+import {chain, Rule, SchematicsException, Tree} from '@angular-devkit/schematics';
+import {getWorkspace} from '@schematics/angular/utility/workspace';
```
SchnWalter, line 13: I tried replacing this `getWorkspace()` with something from `@angular-devkit/core`, but the only API I found was `workspaces.readWorkspace()`, which also requires a file path and doesn't work on the available `host: Tree`, see:

https://github.com/angular/angular-cli/tree/v10.0.0-next.3/packages/angular_devkit/core#usage-example

packages/localize/schematics/ng-add/index_spec.ts
```diff
@@ -155,6 +156,7 @@ export { renderModule, renderModuleFactory } from '@angular/platform-server';`;
 
   it('should not break when there are no polyfills', async () => {
     host.overwrite('angular.json', JSON.stringify({
+      version: 1,
```
SchnWalter, line 159: Looks like the new Workspaces API requires this version flag, otherwise it throws:
```
throw new Error('Unsupported workspace format.');
```

But there is no mention of this in the https://angular.io/guide/deprecations and the codebase seems to not "require" this flag:
```ts
// NOTE: future additions could also perform content analysis to determine format/version
const workspaceFiles: Record<string, WorkspaceFormat> = {
  'angular.json': WorkspaceFormat.JSON,
  '.angular.json': WorkspaceFormat.JSON,
};
```

If this version flag will be added in the future, the new API should first display a warning and not throw an Error, it should go through a deprecation phase.

petebacondarwin: @alan-agius4 ^^^

packages/localize/schematics/ng-add/index.ts
```diff
@@ -8,53 +8,70 @@
...
+}
 
-function prendendToTargetOptionFile(
-    projectName: string, builderName: string, optionName: string, str: string) {
+function prependToTargetFiles(
+    project: workspaces.ProjectDefinition, builderName: Builders, optionName: string, str: string) {
   return (host: Tree) => {
-    // Get all known polyfills for browser builders on this project.
-    const optionValues = getAllOptionValues<string>(host, projectName, builderName, optionName);
+    const optionValues = getRelevantFilePathsForProject(project, builderName, optionName);
```
petebacondarwin, line 74: NIT: Given all the other renames in this file, I would expect this to be something other than `optionValues`. Perhaps `filePaths`?

petebacondarwin: Or `fileList`?

petebacondarwin (approved):
LGTM - a few nits.

packages/localize/schematics/ng-add/index.ts
SchnWalter: I forced pushed a new version with 2 changes:
* renamed `getRelevantFilePathsForProject` to `getFileListForRelevantTargetDefinitions`
* renamed `optionValues` to `fileList`

packages/localize/schematics/ng-add/index_spec.ts
alan-agius4: angular.json must always have a version so this is LGTM

packages/localize/schematics/ng-add/index.ts
```diff
@@ -8,55 +8,72 @@
...
-  // Get all values for the option name and dedupe them.
-  // Deduping will only work for primitives, but the keys we want here are strings so it's ok.
-  const optionValues: T[] =
-      allOptions.filter(o => o[optionName])
-          .map(o => o[optionName])
-          .reduce((acc, cur) => !acc.includes(cur) ? acc.concat(cur) : acc, []);
+function getRelevantTargetDefinitions(
+    project: workspaces.ProjectDefinition, builderName: Builders): workspaces.TargetDefinition[] {
+  const definitions: workspaces.TargetDefinition[] = [];
+  for (const target of project.targets.values()) {
```
alan-agius4, line 24: Not: you shouldn’t be needed `.values` As target is an iterable.

SchnWalter: Changed to `for (const [, target] of project.targets) {...}` and rebased onto master.

packages/localize/schematics/ng-add/index_spec.ts
SchnWalter: Thank you.

I've now updated the `Does this PR introduce a breaking change?` for this MR.

packages/localize/schematics/ng-add/index.ts
petebacondarwin: I'm not sure that this is better :-) but happy to go with it.

SchnWalter: I'm not sure exactly how you would like to change this, could you please give an example.

If the `values()` method is not used, the iterator returns an `IterableIterator<[string,TargetDefinition]>` and to access the `TargetDefinition`, you either need to access it via an array index or by using a destructuring assignment, and I looked over the Angular CLI codebase and saw that there are 10 destructuring assignments that ignore some of the returned values and there were none where the destructuring assignment kept all values but didn't use some of them.

petebacondarwin: I'm not proposing a change. Just pointing out that although `targets` is iterable, it is an iterable of "entries", which results in the ugly `[, target]` destructuring.
Since `targets` is a `Map` (internally) then `targets.value()` is not costly to call, since it doesn't actually create a new array.
If I was pushed to make a suggestion, then I would do something like:

```ts
project.targets.forEach(target => {
  if (target.builder === builderName) {
    definitions.push(target);
  }
});
```

SchnWalter: Interesting, I didn't know that `forEach()` will provide `TargetDefinition, string` and not `[string,TargetDefinition], number`.

Thanks. I've changed the code to match your suggestion.

PR #36901, opened 2020-05-03
fix(core): handle pluralize functions that expect a number

### LINKED ISSUE #36888: [i18n] Different ICU and NgPlural behaviour when using sourceLocale
# 🐞 bug report

### Description

I found out that using a custom `sourceLocale` configuration ("es" in my case), the result from this code is unexpected:

```
<div *ngFor="let case of [0, 1, 2]">
  {{ case }}
  ICU: {case, plural, zero {zero} one {one} other {other}}

  NgPlural:
  <ng-container [ngPlural]="case">
    <ng-template ngPluralCase="zero">zero</ng-template>
    <ng-template ngPluralCase="one">one</ng-template>
    <ng-template ngPluralCase="other">other</ng-template>
  </ng-container>
</div>
```

### Expected:

0 ICU: other NgPlural: other
1 ICU: one NgPlural: one
2 ICU: other NgPlural: other

### Received:

0 ICU: other NgPlural: other
1 ICU: **other** NgPlural: one
2 ICU: other NgPlural: other

## 🔬 Minimal Reproduction

I haven't been able to reproduce it in Stackblitz, it doesn't recognize `sourceLocale` 
 configuration in `angular.json` https://stackblitz.com/edit/angular-g47pyh

Using CLI:
ng new project
Enable `localize` and set `sourceLocale` to "es" in angular.json
Copy the above code and run the project

## 🌍  Your Environment

**Angular Version:**
<pre><code>
@angular-devkit/architect         0.901.4
@angular-devkit/build-angular     0.901.4
@angular-devkit/build-optimizer   0.901.4
@angular-devkit/build-webpack     0.901.4
@angular-devkit/core              9.1.4
@angular-devkit/schematics        9.1.4
@ngtools/webpack                  9.1.4
@schematics/angular               9.1.4
@schematics/update                0.901.4
</code></pre>

**Anything else relevant?**
I'm using Windows

petebacondarwin:
Previously we were passing a string form of the value to pluralize
to the `getLocalePluralCase()` function that is extracted from the
locale data. But some locales have functions that rely upon this
value being a number not a string.

Now we convert the value to a number before passing it to the
locale data function.

Fixes #36888

### DIFF
--- a/packages/core/src/i18n/localization.ts
+++ b/packages/core/src/i18n/localization.ts
@@ -8,26 +8,15 @@
 
 import {getLocalePluralCase} from './locale_data_api';
 
+const pluralMapping = ['zero', 'one', 'two', 'few', 'many'];
+
 /**
  * Returns the plural case based on the locale
  */
-export function getPluralCase(value: any, locale: string): string {
-  const plural = getLocalePluralCase(locale)(value);
-
-  switch (plural) {
-    case 0:
-      return 'zero';
-    case 1:
-      return 'one';
-    case 2:
-      return 'two';
-    case 3:
-      return 'few';
-    case 4:
-      return 'many';
-    default:
-      return 'other';
-  }
+export function getPluralCase(value: string, locale: string): string {
+  const plural = getLocalePluralCase(locale)(parseInt(value, 10));
+  const result = pluralMapping[plural];
+  return (result !== undefined) ? result : 'other';
 }
 
 /**
--- a/packages/core/test/acceptance/i18n_spec.ts
+++ b/packages/core/test/acceptance/i18n_spec.ts
@@ -10,6 +10,7 @@
 import '@angular/localize/init';
 
 import {CommonModule, registerLocaleData} from '@angular/common';
+import localeEs from '@angular/common/locales/es';
 import localeRo from '@angular/common/locales/ro';
 import {computeMsgId} from '@angular/compiler';
 import {Component, ContentChild, ContentChildren, Directive, ElementRef, HostBinding, Input, LOCALE_ID, NO_ERRORS_SCHEMA, Pipe, PipeTransform, QueryList, TemplateRef, Type, ViewChild, ViewContainerRef} from '@angular/core';
@@ -770,7 +771,7 @@ onlyInIvy('Ivy i18n logic').describe('runtime i18n', () => {
       expect(fixture.nativeElement.innerHTML).toEqual(`<div>autre - 4<!--ICU 5--></div>`);
     });
 
-    it('should return the correct plural form for ICU expressions when using a specific locale', () => {
+    it('should return the correct plural form for ICU expressions when using "ro" locale', () => {
       loadTranslations({
         [computeMsgId(
             '{VAR_PLURAL, plural, =0 {no email} =one {one email} =few {a few emails} =other {lots of emails}}')]:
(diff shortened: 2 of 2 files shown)


packages/core/src/i18n/localization.ts
```diff
@@ -8,26 +8,16 @@
...
-    case 2:
-      return 'two';
-    case 3:
-      return 'few';
-    case 4:
-      return 'many';
-    default:
-      return 'other';
-  }
+export function getPluralCase(value: string, locale: string): string {
```
AndrewKushnir, line 16: It looks like we need to update `value` argument type, since we handle other types below as well (i.e. `typeof value === 'string' ? parseInt(value, 10) : value`). I'd also propose adding an inline comment on why we do that (as it's explained in PR description).

AndrewKushnir (changes requested):
Thanks for the fix @petebacondarwin 👍

I guess you also added a test case for this change, but it was hard to find it, since the `i18n_spec.ts` was re-formatted. I've seen undesirable re-formatting to happen because of long lines that were added (for example, for translation strings). If that's the case, you can transform a string into a concatenation of a smaller strings - it should help avoid this re-formatting.

packages/core/src/i18n/localization.ts
petebacondarwin: As it turns out this is always a string... So perhaps we should just avoid the `typeof value === 'string'` check?

packages/core/test/acceptance/i18n_spec.ts
```diff
@@ -814,6 +815,48 @@ onlyInIvy('Ivy i18n logic').describe('runtime i18n', () => {
       expect(fixture.nativeElement.innerHTML).toEqual('no email<!--ICU 2-->');
     });
 
+    it(`should return the correct plural form for ICU expressions when using "es" locale`, () => {
+      const icuMessage = '{VAR_PLURAL, plural, =0 {no email} =one ' +
+          '{one email} =few {a few emails} =other {lots of emails}}';
+      loadTranslations({[computeMsgId(icuMessage)]: icuMessage});
+      registerLocaleData(localeEs);
+      TestBed.configureTestingModule({providers: [{provide: LOCALE_ID, useValue: 'es'}]});
```
AndrewKushnir, line 823: Could you plz add a quick comment above this line on why "es" locale was selected for thus test (and may be also a comment on the previous test that uses "ro" locale)? Currently it might be a bit confusing that we have 2 tests with the same title (just different locale), when in fact we do these checks for a reason :)

As a possible option, we could also avoid adding a second test and reuse existing one:
```
// ... short description on why we test with these locales ...
['es', 'ro'].forEach(locale => {
  it(`should return the correct plural form for ICU expressions when using "${locale}" locale`, () => { ... });
});
```

AndrewKushnir (approved):
Thanks for the update @petebacondarwin 👍 

I've left 1 comment regarding a test, otherwise LGTM!

packages/core/test/acceptance/i18n_spec.ts
petebacondarwin: Originally I did do that. But the resulting expectations are different so it just becomes a confusion of conditionals.

I'll update the test descriptions.

AndrewKushnir:
[Presubmit](http://test/OCL:309822483:BASE:309822526:1588629610502:ca66d50)

AndrewKushnir:
@petebacondarwin FYI, presubmit was successful, please add "merge" label if PR is ready to be merged.

PR #36914, opened 2020-05-04
build: do not build runfile trees unnecessarily

devversion:
Disables Bazel runfile tree creation. Only if a given
execution strategy relies on runfile tree creation, the
runfile tree is created lazily. This helps as currently
Bazel spends unnecessary time on CI building runfile trees
for tests which are cached and aren't re-run.

The goal is to spend less time on CI for cached test/build
targets. We can't measure impact yet as there are targets
for the integration tests that hide the potential benefits.
on the components repo a 80% time reduction could be observed.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -33,6 +33,11 @@ build --incompatible_strict_action_env
 run --incompatible_strict_action_env
 test --incompatible_strict_action_env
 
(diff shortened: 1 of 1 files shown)


josephperrott (approved):
LGTM

PR #36918, opened 2020-05-04
update ng-dev to rely on a javascript file for configuration

josephperrott:
see individual commits


dev-infra/utils/config.ts
```diff
@@ -25,19 +27,18 @@ export function getRepoBaseDir() {
...
-        `Unable to find config file at:\n` +
-        `  ${configPath}`);
+    return require(configPath)() as DevInfraConfig<K, T>;
+  } catch (err) {
+    if (!supressError) {
+      throw Error(`Unable to load config file at:\n  ${configPath}`);
+    }
   }
-  return parse(rawConfig);
+  return {} as DevInfraConfig<K, T>;
```
devversion, line 41: For a follow-up, do you think it would make sense updating `DevInfraConfig` so that `K` is optional? It could be `undefined` and I think the type should reflect that.

devversion (approved):
LGTM. I need that for the merge script too!

dev-infra/utils/config.ts
```diff
@@ -25,19 +27,18 @@ export function getRepoBaseDir() {
 }
 
 /**
- * Retrieve the configuration from the .dev-infra.json file.
+ * Retrieve the configuration from the .dev-infra-config.js file.
```
gkalpak, line 30:
```suggestion
 * Retrieve the configuration from the .ng-dev-config.js file.
```

.ng-dev-config.js
```diff
@@ -0,0 +1,78 @@
...
+      '!**/dist/**',
+      '!**/built/**',
+      '!shims_for_IE.js',
+    ]
+  },
+  'buildifier': true
+};
+
+// Export function to build ng-dev configuration object.
+module.exports = () => {
```
gkalpak, line 76: Why export a function (and not an object)?

josephperrott: I did it initially just to match what @devversion already was doing in the components repo.

Though, I suppose its the same thing since we immediately invoke the method.

I will update it to just be an object.

PR #36853, opened 2020-04-29
fix(localize): ensure `getLocation()` works

petebacondarwin:
The `getLocation()` method was not working as there were typos in the
properties it was reading. This was not picked up because there were
neither typings for these properties nor unit tests to check it worked.

PR #36920, opened 2020-05-04
fix(localize): ensure `getLocation()` works

petebacondarwin:
This is a patch backport of #36853

The `getLocation()` method was not working as there were typos in the
properties it was reading. This was not picked up because there were
neither typings for these properties nor unit tests to check it worked.


petebacondarwin:
Merge-assistance: this is just a rebase of #36853 onto patch to avoid the merge conflict.

PR #36921, opened 2020-05-04
feat(compiler-cli): report error if undecorated class with Angular features is discovered

devversion:
* See individual commits.


devversion:
Note: The components repo unit test job currently fails as a new compiler diagnostic has been introduced in this PR and errors would need to be solved in the components repo. 

Regardless of that failing job though, it would be good to receive feedback on the PR.

atscott (approved):
reviewed for: fw-router

packages/compiler-cli/src/ngtsc/annotations/src/diagnostics.ts
```diff
@@ -78,6 +78,14 @@ export function getDirectiveDiagnostics(
   return diagnostics;
 }
 
+export function getUndecoratedClassWithAngularFeaturesDiagnostic(node: ClassDeclaration):
+    ts.Diagnostic {
+  return makeDiagnostic(
+      ErrorCode.UNDECORATED_CLASS_USING_ANGULAR_FEATURES, node.name,
+      `Class is using Angular features but is not decorated. Please add an explicit ` +
```
alxhub, line 85: This is good for now, but we want to make this error message better before final.

1) the error should go on the feature in question, not the class identifier. So if you have an `@Input() field: ...` on the class, the error should go on the `@Input` decorator itself. Or the lifecycle method name, etc.

2) the error should include an extra pointer to the class name (a `relatedInfo` node) that suggests adding a decorator to the class itself. For `@Input` or any of the directive/component specific annotations, we should suggest adding `@Directive()`. For lifecycle hooks or DI annotations in the constructor, we should suggest adding any decorator.

packages/compiler-cli/test/ngtsc/ngtsc_spec.ts
```diff
@@ -905,6 +907,23 @@ runInEachFileSystem(os => {
               `static ɵdir: i0.ɵɵDirectiveDefWithMeta<TestBase, never, never, { "input": "input"; }, {}, never>;`);
     });
 
+    it('should error if undecorated class with Angular features has been discovered', () => {
```
alxhub, line 910: Add a test for the lifecycle case as well.

alxhub (approved):
LGTM with a few comments, for the compiler piece only.

packages/compiler-cli/src/ngtsc/annotations/src/diagnostics.ts
devversion: (1) 👍 
(2) 👍 I did something like that initially but left that out as it felt out-of-scope. I can re-use that. https://gist.github.com/devversion/d412695321ca2a37fc59d84b6255f733 (patch is reversed)

devversion: @alxhub Regarding (2), you mention DI annotations. This is not something that is handled yet. It sounds like we should though? I assume we can do that as a follow-up too.

AndrewKushnir (approved):
Reviewed for: fw-forms

packages/compiler-cli/src/ngtsc/annotations/src/diagnostics.ts
kara: +1 to what Alex said. I'm good with that being a follow-up.

kara (changes requested):
Main things look good. Typos and tests

petebacondarwin (approved):
Reviewed for: fw-upgrade, docs-infra

packages/compiler-cli/test/ngtsc/ngtsc_spec.ts
```diff
@@ -905,6 +907,44 @@ runInEachFileSystem(os => {
...
+
+        const errors = env.driveDiagnostics();
+        expect(errors.length).toBe(1);
+        expect(trim(errors[0].messageText as string))
+            .toContain(
+                'Class is using Angular features but is not decorated. Please add an explicit ' +
+                'Angular decorator.');
+      });
+
+      it('should error if directive lifecye hook has been discovered', () => {
```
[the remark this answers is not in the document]
devversion: lol. Not sure how that one got cut off. Good catch.

kara (approved):
LGTM

IgorMinar (approved):
LGTM! Thank you!

PR #36922, opened 2020-05-04
feat(language-service): [ivy] Parse Angular compiler options

kyliau:
Parse Angular compiler options in Angular language service.

In View Engine, only TypeScript compiler options are read, Angular
compiler options are not. With Ivy, there could be different modes of
compilation, most notably how strict the templates should be checked.
This commit makes the behavior of language service consistent with the
Ivy compiler.

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/language-service/ivy/language_service.ts
+++ b/packages/language-service/ivy/language_service.ts
@@ -6,12 +6,50 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
+import {CompilerOptions, createNgCompilerOptions} from '@angular/compiler-cli';
 import * as ts from 'typescript/lib/tsserverlibrary';
 
 export class LanguageService {
-  constructor(private readonly tsLS: ts.LanguageService) {}
+  private options: CompilerOptions;
+
+  constructor(project: ts.server.Project, private readonly tsLS: ts.LanguageService) {
+    this.options = parseNgCompilerOptions(project);
+    this.watchConfigFile(project);
+  }
 
   getSemanticDiagnostics(fileName: string): ts.Diagnostic[] {
     return [];
   }
+
+  private watchConfigFile(project: ts.server.Project) {
+    // TODO: Check the case when the project is disposed. An InferredProject
+    // could be disposed when a tsconfig.json is added to the workspace,
+    // in which case it becomes a ConfiguredProject (or vice-versa).
+    // We need to make sure that the FileWatcher is closed.
+    if (!(project instanceof ts.server.ConfiguredProject)) {
+      return;
+    }
+    const {host} = project.projectService;
+    host.watchFile(
+        project.getConfigFilePath(), (fileName: string, eventKind: ts.FileWatcherEventKind) => {
+          project.log(`Config file changed: ${fileName}`);
+          if (eventKind === ts.FileWatcherEventKind.Changed) {
+            this.options = parseNgCompilerOptions(project);
+          }
+        });
+  }
+}
+
+export function parseNgCompilerOptions(project: ts.server.Project): CompilerOptions {
+  let config = {};
+  if (project instanceof ts.server.ConfiguredProject) {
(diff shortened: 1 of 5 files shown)


packages/language-service/ivy/language_service.ts
```diff
@@ -6,12 +6,46 @@
...
+  }
 
   getSemanticDiagnostics(fileName: string): ts.Diagnostic[] {
     return [];
   }
+
+  private watchConfigFile(project: ts.server.Project) {
+    if (!(project instanceof ts.server.ConfiguredProject)) {
+      return;
+    }
```
ayazhafiz, line 27: What happens if someone starts the LS in a project with no tsconfig, then adds a tsconfig later on? Not saying we should worry about it, but we won't detect the config change right?

packages/language-service/ivy/language_service.ts
```diff
@@ -6,12 +6,46 @@
...
+          }
+        });
+  }
+}
+
+export function parseNgCompilerOptions(project: ts.server.Project): CompilerOptions {
+  let config = {};
+  if (project instanceof ts.server.ConfiguredProject) {
+    const configPath = project.getConfigFilePath();
+    const result = ts.readConfigFile(configPath, (path: string) => project.readFile(path));
```
ayazhafiz, line 43:
```suggestion
    const result = ts.readConfigFile(configPath, path => project.readFile(path));
```

ayazhafiz (approved):

packages/language-service/ivy/language_service.ts
SchnWalter: Our application is using a prepare script to generate the tsconfig and angular configuration files because the files were getting huge, so we are now "building" the configuration files.

It would be nice to have this working with Ivy.

ayazhafiz: @SchnWalter when is the config file built? Today TypeScript only supports static analysis of tsconfig.json files. The language service relies only on tsconfig, not on other config files like angular.json.

SchnWalter: We generate our `tsconfig.json` file on `npm postinstall` for CI pipelines and on localhost we also use some some IDE launch commands that generate new tsconfig files, and this allows us to test various combinations of libraries. And it would be nice to not have to manually reload the TypeScript services or restart the IDEs whenever we regenerate the tsconfig files.

In the mono-repository we have quite a few libraries and several projects that consume them and most packages also support multiple "feature flags" which are also using various tsconfig path aliases.

ayazhafiz: I think this kind of request is better targeted at TypeScript. Angular only consumes TypeScript APIs. Sounds like you are looking for a conditional build system that tsconfig does not support. There is a long-open proposal for [conditional compilation directives](https://github.com/microsoft/TypeScript/issues/3538), which could be somewhat related.

filipesilva (approved):

packages/language-service/ivy/language_service.ts
kyliau: > What happens if someone starts the LS in a project with no tsconfig, then adds a tsconfig later on? Not saying we should worry about it, but we won't detect the config change right?

Good question! tsserver automatically takes care of situation like this. Without a `tsconfig`, an `InferredProject` is created to host the scripts. Upon creation of a `tsconfig` in the workspace, the scripts will be transferred to a `ConfiguredProject`, the old project will be disposed along with the language service, and a new language service will be created.
I added a TODO here to make sure we clean up the FileWatcher when this happens.

ayazhafiz: nice!

PR #36923, opened 2020-05-04
test(language-service): Add mock service to overwrite files

kyliau:
Add a mechanism to replace file contents for a specific file. This
allows us to write custom test scenarios in code without modifying the
test project.

Since we are no longer mocking the language service host, the file
overwrite needs to happen via the project service.
Project service manages a set of script infos, and overwriting the files
is a matter of updating the relevant script infos.

Note that the actual project service is wrapped inside a Mock Service.
Tests should not have direct access to the project service. All
manipulations should take place via the Mock Service.

The MockService provides a `reset()` method to undo temporary overwrites
after each test.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/language-service/ivy/test/mock_host.ts
+++ b/packages/language-service/ivy/test/mock_host.ts
@@ -36,6 +36,10 @@ export const APP_COMPONENT = join(PROJECT_DIR, 'app', 'app.component.ts');
 export const APP_MAIN = join(PROJECT_DIR, 'app', 'main.ts');
 export const PARSING_CASES = join(PROJECT_DIR, 'app', 'parsing-cases.ts');
 
+const NOOP_FILE_WATCHER: ts.FileWatcher = {
+  close() {}
+};
+
 export const host: ts.server.ServerHost = {
   ...ts.sys,
   readFile(absPath: string, encoding?: string): string |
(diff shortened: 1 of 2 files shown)


packages/language-service/ivy/test/mock_host_spec.ts
```diff
@@ -35,13 +40,27 @@ describe('mock host', () => {
     ]));
   });
 
+  it('can overwrite test file', () => {
+    service.overwrite(APP_MAIN, `const x: string = 0`);
+    const diags = tsLS.getSemanticDiagnostics(APP_MAIN);
+    expect(diags.length).toBe(1);
+    expect(diags[0].messageText).toBe(`Type '0' is not assignable to type 'string'.`);
+  });
```
ayazhafiz, line 48: IMO this test should check the contents of `APP_MAIN` rather than calling the language service. Could we also verify that the cursor symbol is stripped from the contents of the file, either in this test or another?

packages/language-service/ivy/test/mock_host_spec.ts
```diff
@@ -35,13 +40,27 @@ describe('mock host', () => {
...
     expect(diags).toEqual([]);
   });
+
+  it('can find the cursor', () => {
+    const content = service.overwrite(APP_MAIN, `const fo¦o = 'hello world';`);
+    const cursor = content.indexOf('¦');
+    expect(cursor).toBe(8);
+    const quickInfo = tsLS.getQuickInfoAtPosition(APP_MAIN, cursor);
+    expect(quickInfo?.displayParts?.map(p => p.text).join('')).toBe('const foo: "hello world"');
+  });
```
ayazhafiz, line 65: Same as above, I don't think we need this test currently and that it would be better just to check that the cursor symbol isn't present.

ayazhafiz (changes requested):

packages/language-service/ivy/test/mock_host_spec.ts
kyliau: Good call! I actually wanted to test that the project is up to date after the script is changed, but now that I think more about it, it should not be in the scope of this file.

packages/language-service/ivy/test/mock_host_spec.ts
kyliau: done

packages/language-service/ivy/test/mock_host.ts
```diff
@@ -73,5 +92,60 @@ export function setup() {
...
+      const reloaded = scriptInfo.reloadFromFile();
+      if (!reloaded) {
+        throw new Error(`Failed to reload ${scriptInfo.fileName}`);
+      }
+    }
+    const sameProgram = this.project.updateGraph();
+    if (sameProgram) {
+      throw new Error('Project should have updated program after reset');
+    }
+    this.overwrites.clear();
```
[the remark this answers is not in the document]
kyliau: I'm entirely not sure if this helps with readability. IMO having to reason through `this.overwrites.size > 0 && sameProgram` is more mental overhead than the early termination statement which short circuits the rest of the method.
It's also not very clear why we need to call `this.overwrites.clear()` even though the Set is empty.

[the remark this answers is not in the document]
ayazhafiz: Okay, sounds good :)

ayazhafiz (approved):

PR #33128, opened 2019-10-13
feat(router): add and remove `initialNavigation` options

CaerusKaru:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] angular.io application / infrastructure changes
- [ ] Other... Please describe:

## What is the current behavior?

Issue Number: https://github.com/angular/angular/issues/32887

## What is the new behavior?

Added two new options, removed four old options, as described in above issue.

## Does this PR introduce a breaking change?

- [x] Yes
- [ ] No

Added migration for removed options

PR #36926, opened 2020-05-05
feat(core): add initialNavigation schematic

CaerusKaru:
Add a schematic to update users to the new v11 `initialNavigation`
options for `RouterModule`. This replaces the deprecated/removed
`true`, `false`, `legacy_disabled`, and `legacy_enabled` options
with the newer `enabledBlocking` and `enabledNonBlocking` options.

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

## Other information

Follow-up to https://github.com/angular/angular/pull/33128

### DIFF
--- a/packages/core/schematics/migrations.json
+++ b/packages/core/schematics/migrations.json
@@ -74,6 +74,11 @@
       "version": "11.0.0-beta",
       "description": "NavigationExtras.preserveQueryParams has been removed as of Angular version 11.  This migration replaces any usages with the appropriate assignment of the queryParamsHandler key.",
       "factory": "./migrations/router-preserve-query-params/index"
+    },
+    "migration-v11-router-initial-navigation-options": {
+      "version": "11.0.0-beta",
+      "description": "Updates the `initialNavigation` property for `RouterModule.forRoot`.",
+      "factory": "./migrations/initial-navigation/index"
     }
   }
 }
--- a/packages/core/schematics/migrations/google3/initialNavigationRule.ts
+++ b/packages/core/schematics/migrations/google3/initialNavigationRule.ts
@@ -1,0 +1,55 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {RuleFailure, Rules} from 'tslint';
+import * as ts from 'typescript';
+
+import {InitialNavigationCollector} from '../initial-navigation/collector';
+import {TslintUpdateRecorder} from '../initial-navigation/google3/tslint_update_recorder';
+import {InitialNavigationTransform} from '../initial-navigation/transform';
+
+
+
+/**
+ * TSLint rule that updates RouterModule `forRoot` options to be in line with v10 updates.
+ */
+export class Rule extends Rules.TypedRule {
+  applyWithProgram(sourceFile: ts.SourceFile, program: ts.Program): RuleFailure[] {
+    const ruleName = this.ruleName;
+    const typeChecker = program.getTypeChecker();
+    const sourceFiles = program.getSourceFiles().filter(
+        s => !s.isDeclarationFile && !program.isSourceFileFromExternalLibrary(s));
+    const initialNavigationCollector = new InitialNavigationCollector();
+    const failures: RuleFailure[] = [];
+
+    // Analyze source files by detecting all ExtraOptions#InitialNavigation assignments
+    sourceFiles.forEach(sourceFile => initialNavigationCollector.visitNode(sourceFile));
+
+    const {assignments} = initialNavigationCollector;
+    const transformer = new InitialNavigationTransform(typeChecker, getUpdateRecorder);
+    const updateRecorders = new Map<ts.SourceFile, TslintUpdateRecorder>();
+
+    transformer.migrateInitialNavigationAssignments(assignments);
+
+    if (updateRecorders.has(sourceFile)) {
+      failures.push(...updateRecorders.get(sourceFile)!.failures);
+    }
+
+    return failures;
+
+    /** Gets the update recorder for the specified source file. */
+    function getUpdateRecorder(sourceFile: ts.SourceFile): TslintUpdateRecorder {
+      if (updateRecorders.has(sourceFile)) {
+        return updateRecorders.get(sourceFile)!;
+      }
+      const recorder = new TslintUpdateRecorder(ruleName, sourceFile);
+      updateRecorders.set(sourceFile, recorder);
+      return recorder;
+    }
+  }
+}
--- a/packages/core/schematics/migrations/google3/initialNavigationRule.ts
+++ b/packages/core/schematics/migrations/google3/initialNavigationRule.ts
@@ -24,7 +24,7 @@ export class Rule extends Rules.TypedRule {
     const typeChecker = program.getTypeChecker();
     const sourceFiles = program.getSourceFiles().filter(
         s => !s.isDeclarationFile && !program.isSourceFileFromExternalLibrary(s));
-    const initialNavigationCollector = new InitialNavigationCollector();
+    const initialNavigationCollector = new InitialNavigationCollector(typeChecker);
     const failures: RuleFailure[] = [];
 
     // Analyze source files by detecting all ExtraOptions#InitialNavigation assignments
@@ -34,7 +34,7 @@ export class Rule extends Rules.TypedRule {
     const transformer = new InitialNavigationTransform(typeChecker, getUpdateRecorder);
     const updateRecorders = new Map<ts.SourceFile, TslintUpdateRecorder>();
 
-    transformer.migrateInitialNavigationAssignments(assignments);
+    transformer.migrateInitialNavigationAssignments(Array.from(assignments));
 
     if (updateRecorders.has(sourceFile)) {
       failures.push(...updateRecorders.get(sourceFile)!.failures);
--- a/packages/core/schematics/migrations/initial-navigation/BUILD.bazel
+++ b/packages/core/schematics/migrations/initial-navigation/BUILD.bazel
@@ -1,0 +1,19 @@
+load("//tools:defaults.bzl", "ts_library")
+
+ts_library(
+    name = "initial-navigation",
+    srcs = glob(["**/*.ts"]),
+    tsconfig = "//packages/core/schematics:tsconfig.json",
+    visibility = [
+        "//packages/core/schematics:__pkg__",
+        "//packages/core/schematics/migrations/google3:__pkg__",
+        "//packages/core/schematics/migrations/initial-navigation/google3:__pkg__",
+        "//packages/core/schematics/test:__pkg__",
+    ],
+    deps = [
+        "//packages/core/schematics/utils",
+        "@npm//@angular-devkit/schematics",
+        "@npm//@types/node",
+        "@npm//typescript",
+    ],
+)
--- a/packages/core/schematics/migrations/initial-navigation/README.md
+++ b/packages/core/schematics/migrations/initial-navigation/README.md
@@ -1,0 +1,43 @@
+## initialNavigation migration
+
+Automatically migrates the `initialNavigation` property of the `RouterModule` to the newly
+available options: `enabledNonBlocking` (default), `enabledBlocking`, and `disabled`.
+
+#### Before
+```ts
+import { NgModule } from '@angular/core';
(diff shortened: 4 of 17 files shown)


packages/core/schematics/migrations/initial-navigation/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+### Disclaimer
+
+The migration only covers the most common patterns where developers set the `ExtraOptions#InitialNavigation` 
+option to an outdated value. Therefore, if a user declares the option using a number of other methods, 
+e.g. shorthand notation, variable declaration, or some other crafty method, they will have to migrate 
+those options by hand. Otherwise, the compiler will error if the types are sufficiently enforced.
+
+The basic migration strategy is as follows:
+* `legacy_disabled` || `false` => `disabled`
```
atscott, line 42: I'm not sure if we can have this change as part of the migration in 10.1 since `legacy_disabled` and `false` work slightly differently than `disabled`. We might only be able to make migrations that are exactly 1 to 1 now and then update this in v11.

devversion: Yeah, I'd agree with this.

packages/core/schematics/migrations/initial-navigation/collector.ts
```diff
@@ -0,0 +1,39 @@
...
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import * as ts from 'typescript';
+
+
+const INITIAL_NAVIGATION = 'initialNavigation';
```
devversion, line 11: Add a comment what this constant represents. i.e. the property name for the options that need to be migrated.

packages/core/schematics/migrations/initial-navigation/collector.ts
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Visitor that walks through specified TypeScript nodes and collects all
+ * found ExtraOptions#InitialNavigation assignments.
+ */
+export class InitialNavigationCollector {
+  public assignments: ts.PropertyAssignment[] = [];
+
+  constructor() {}
```
devversion, line 20: Nit: can be removed

packages/core/schematics/migrations/initial-navigation/collector.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  visitNode(node: ts.Node) {
+    if (ts.isPropertyAssignment(node) &&
+        (ts.isIdentifier(node.name) || ts.isStringLiteral(node.name) ||
+         ts.isNumericLiteral(node.name))) {
```
devversion, line 25: Out of curiosity: Is it contradictory that we handle numeric literal here but check for text below? We should probably remove this. Also we should use `ts.isStringLiteralLike` above to handle backtick property names (low-effort to add; but increases coverage)

packages/core/schematics/migrations/initial-navigation/collector.ts
```diff
@@ -0,0 +1,39 @@
...
+      if (node.name.text === INITIAL_NAVIGATION && isOldValue(node)) {
+        this.assignments.push(node);
+      }
+    }
+
+    ts.forEachChild(node, n => this.visitNode(n));
+  }
+}
+
+function isOldValue(node: ts.PropertyAssignment): boolean {
```
devversion, line 35: Nit: Add function description and potentially rename to: `isValidInitialNavigationValue`?

packages/core/schematics/migrations/initial-navigation/google3/tslint_update_recorder.ts
```diff
@@ -0,0 +1,25 @@
...
+import {UpdateRecorder} from '../update_recorder';
+
+export class TslintUpdateRecorder implements UpdateRecorder {
+  failures: RuleFailure[] = [];
+
+  constructor(private ruleName: string, private sourceFile: ts.SourceFile) {}
+
+  updateNode(node: ts.Node, newText: string): void {
+    this.failures.push(new RuleFailure(
+        this.sourceFile, node.getStart(), node.getEnd(), `Node needs to be updated to: ${newText}`,
```
devversion, line 20: I think in other schematics we are more precise with why nodes need to be changed. Here it's super generic and we just request a node to be updated. This is _totally_ fine if the lint rule just runs to migrate internal usages.

On the other hand though, if it's set up as lint rule in g3 until we removed the deprecated behavior, this might be not ideal. Do we have more info on how this is used?

packages/core/schematics/test/initial_navigation_migration_spec.ts
```diff
@@ -0,0 +1,212 @@
...
+            RouterModule.forRoot([], {...options, initialNavigation: 'legacy_disabled'}),
+          ]
+        })
+        export class AppModule {
+        }
+      `);
+
+    await runMigration();
+    expect(tree.readContent('/index.ts')).toContain(`enabledNonBlocking`);
+    expect(tree.readContent('/index.ts')).toContain(`disabled`);
```
devversion, line 180: Nit: should these assertions (and in other tests) be a little more precise? Right now options could theoretically swapped in wrong order. Might just make this a small regex checking for surrounding context.

devversion (approved):
LGTM. A few minor nits. I'm glad we decided on the more straightforward non-static-interpreter approach as that seems to capture the most common cases quite reliably.

packages/core/schematics/migrations/initial-navigation/README.md
atscott: @CaerusKaru now that `master` is targeting v11, I think we can submit this change. Can you do a quick pass over this migration and the follow-up PR to remove the old options?

atscott (approved):
LGTM with a couple small comments

packages/core/schematics/migrations.json
```diff
@@ -63,7 +63,11 @@
     "migration-v11-native-view-encapsulation": {
       "version": "11.0.0-beta",
       "description": "ViewEncapsulation.Native has been removed as of Angular version 11. This migration replaces any usages with ViewEncapsulation.ShadowDom.",
-      "factory": "./migrations/native-view-encapsulation/index"
+      "factory": "./migrations/native-view-encapsulation/index",
```
gkalpak, line 66:
```suggestion
      "factory": "./migrations/native-view-encapsulation/index"
    },
```

packages/core/schematics/test/google3/initial_navigation_rule_spec.ts
```diff
@@ -0,0 +1,211 @@
+/**
```
gkalpak, line 1: Nit: The `_rule` part in the file name feels confusing. I think it would be more consistent without the `_rule` part as well (although some other spec files have it, most don't).

packages/core/schematics/migrations/module-with-providers/transform.ts
```diff
@@ -20,7 +20,7 @@ export interface AnalysisFailure {
   message: string;
 }
 
-const TODO_COMMENT = 'TODO: The following node requires a generic type for `ModuleWithProviders';
+const TODO_COMMENT = 'TODO: The following node requires a generic type for `ModuleWithProviders`';
```
gkalpak, line 23: Nit: Since this change is unrelated, consider moving it to a separate commit.

packages/core/schematics/migrations/initial-navigation/BUILD.bazel
```diff
@@ -0,0 +1,23 @@
...
+        "//packages/core/schematics:__pkg__",
+        "//packages/core/schematics/migrations/google3:__pkg__",
+        "//packages/core/schematics/migrations/initial-navigation/google3:__pkg__",
+        "//packages/core/schematics/test:__pkg__",
+    ],
+    deps = [
+        "//packages/compiler-cli/src/ngtsc/annotations",
+        "//packages/compiler-cli/src/ngtsc/imports",
+        "//packages/compiler-cli/src/ngtsc/partial_evaluator",
+        "//packages/compiler-cli/src/ngtsc/reflection",
```
gkalpak, line 17: Are these needed? (If not, please remove them.)

gkalpak (changes requested):
Reviewed-for: dev-infra

packages/core/schematics/migrations/initial-navigation/collector.ts
```diff
@@ -0,0 +1,43 @@
...
+ * found ExtraOptions#InitialNavigation assignments.
+ */
+export class InitialNavigationCollector {
+  public assignments: ts.PropertyAssignment[] = [];
+
+  visitNode(node: ts.Node) {
+    if (ts.isPropertyAssignment(node) &&
+        (ts.isIdentifier(node.name) || ts.isStringLiteral(node.name) ||
+         ts.isStringLiteralLike(node.name))) {
+      if (node.name.text === INITIAL_NAVIGATION && isValidInitialNavigationValue(node)) {
```
crisbeto, line 25: This will capture property assignments that are called `initialNavigation`, but aren't used in a `RouterModule.forRoot` call (e.g. `const myRandomObject = {initialNavigation: 'foo', otherStuff: false}`).

A more robust way of doing it would be to identify the `RouterModule.forRoot` calls and trace back the literal to its declaration. We've done this in the `navigation-extras-omissions` migration although you may be able to simplify it a bit for this migration since detecting `RouterModule.forRoot` is simpler than detecting whether something is an instance of a `Router`.

I believe @atscott did something similar for the `relative-link-resolution` migration as well.

Note that if you end up going down this route, your tests might start failing because you'll have to create a fake router module for your test in order for type checking to pick it up. You can check the tests of the aforementioned migrations to see how it can be done.

packages/core/schematics/migrations/initial-navigation/collector.ts
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * Visitor that walks through specified TypeScript nodes and collects all
+ * found ExtraOptions#InitialNavigation assignments.
+ */
+export class InitialNavigationCollector {
+  public assignments: ts.PropertyAssignment[] = [];
+
+  visitNode(node: ts.Node) {
+    if (ts.isPropertyAssignment(node) &&
+        (ts.isIdentifier(node.name) || ts.isStringLiteral(node.name) ||
```
crisbeto, line 23: I think that `isStringLiteral` is covered under the `isStringLiteralLike` call right after it.

packages/core/schematics/migrations/initial-navigation/collector.ts
```diff
@@ -0,0 +1,43 @@
...
+
+/** The property name for the options that need to be migrated */
+const INITIAL_NAVIGATION = 'initialNavigation';
+
+/**
+ * Visitor that walks through specified TypeScript nodes and collects all
+ * found ExtraOptions#InitialNavigation assignments.
+ */
+export class InitialNavigationCollector {
+  public assignments: ts.PropertyAssignment[] = [];
```
crisbeto, line 19: Should this be a `Set` so we don't migrate the same node multiple times?

packages/core/schematics/migrations/google3/initialNavigationRule.ts
```diff
@@ -0,0 +1,55 @@
...
+    transformer.migrateInitialNavigationAssignments(assignments);
+
+    if (updateRecorders.has(sourceFile)) {
+      failures.push(...updateRecorders.get(sourceFile)!.failures);
+    }
+
+    return failures;
+
+    /** Gets the update recorder for the specified source file. */
+    function getUpdateRecorder(sourceFile: ts.SourceFile): TslintUpdateRecorder {
```
crisbeto, line 46: Consider moving it into a private method or outside of `applyWithProgram` so the function doesn't get re-declared on each invocation.

packages/core/schematics/test/initial_navigation_migration_spec.ts
```diff
@@ -0,0 +1,217 @@
...
+    // Switch into the temporary directory path. This allows us to run
+    // the schematic against our custom unit test tree.
+    shx.cd(tmpDirPath);
+  });
+
+  afterEach(() => {
+    shx.cd(previousWorkingDir);
+    shx.rm('-r', tmpDirPath);
+  });
+
```
crisbeto, line 48: We should have a test for the case where an object with a `initialNavigation` property assignment **isn't** passed into `forRoot`.

crisbeto (approved):

gkalpak (approved):

PR #36929, opened 2020-05-05
build: fix and re-enable elements tests on saucelabs

crisbeto:
The `elements` tests were disabled on Saucelabs, because they were failing on IE10. The problem was that we were loading an es2015 file from npm directly, causing a syntax error. These changes transpile the file to es5.


josephperrott (approved):
LGTM

andrewseguin (approved):
LGTM

devversion (approved):

PR #36930, opened 2020-05-05
feat(language-service): [ivy] wrap ngtsc to handle typecheck files

kyliau:
This commit adds a Compiler interface that wraps the actual ngtsc
compiler. The language-service specific compiler manages multiple
typecheck files using the Project interface, creating and adding
ScriptInfos as necessary.

This commit also adds `overrideInlineTemplate()` method to the mock
service so that we could test the Compiler diagnostics feature.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/language-service/ivy/BUILD.bazel
+++ b/packages/language-service/ivy/BUILD.bazel
@@ -1,12 +1,13 @@
 load("//tools:defaults.bzl", "ts_library")
 
-package(default_visibility = ["//visibility:public"])
+package(default_visibility = ["//packages/language-service:__subpackages__"])
 
 ts_library(
     name = "ivy",
     srcs = glob(["*.ts"]),
     deps = [
         "//packages/compiler-cli",
+        "//packages/language-service/ivy/compiler",
         "@npm//typescript",
     ],
 )
--- a/packages/language-service/ivy/compiler/compiler.ts
+++ b/packages/language-service/ivy/compiler/compiler.ts
@@ -1,0 +1,120 @@
+
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {CompilerOptions} from '@angular/compiler-cli';
+import {NgCompiler, NgCompilerHost} from '@angular/compiler-cli/src/ngtsc/core';
+import {AbsoluteFsPath} from '@angular/compiler-cli/src/ngtsc/file_system';
+import {TypeCheckingProgramStrategy, UpdateMode} from '@angular/compiler-cli/src/ngtsc/typecheck';
+import * as ts from 'typescript/lib/tsserverlibrary';
+
+import {makeCompilerHostFromProject} from './compiler_host';
+
+export class Compiler {
+  private tsCompilerHost: ts.CompilerHost;
+  private lastKnownProgram: ts.Program;
+  private compiler: NgCompiler;
+  private readonly strategy: TypeCheckingProgramStrategy;
+
+  constructor(private readonly project: ts.server.Project, private options: CompilerOptions) {
+    this.tsCompilerHost = makeCompilerHostFromProject(project);
+    const ngCompilerHost = NgCompilerHost.wrap(
+        this.tsCompilerHost,
+        project.getRootFiles(),  // input files
+        options,
+        null,  // old program
+    );
+    this.strategy = createTypeCheckingProgramStrategy(project);
+    this.lastKnownProgram = this.strategy.getProgram();
+    this.compiler = new NgCompiler(ngCompilerHost, options, this.lastKnownProgram, this.strategy);
+  }
+
+  setCompilerOptions(options: CompilerOptions) {
+    this.options = options;
+  }
+
+  analyze(): ts.Program|undefined {
+    const inputFiles = this.project.getRootFiles();
+    const ngCompilerHost =
+        NgCompilerHost.wrap(this.tsCompilerHost, inputFiles, this.options, this.lastKnownProgram);
+    const program = this.strategy.getProgram();
+    this.compiler =
+        new NgCompiler(ngCompilerHost, this.options, program, this.strategy, this.lastKnownProgram);
+    try {
+      // This is the only way to force the compiler to update the typecheck file
+      // in the program. We have to do try-catch because the compiler immediately
+      // throws if it fails to parse any template in the entire program!
+      const d = this.compiler.getDiagnostics();
+      if (d.length) {
+        // There could be global compilation errors. It's useful to print them
+        // out in development.
+        console.error(d.map(d => ts.flattenDiagnosticMessageText(d.messageText, '\n')));
+      }
+    } catch (e) {
+      console.error('Failed to analyze program', e.message);
+      return;
+    }
+    this.lastKnownProgram = this.compiler.getNextProgram();
+    return this.lastKnownProgram;
+  }
+
+  getDiagnostics(sourceFile: ts.SourceFile): ts.Diagnostic[] {
+    return this.compiler.getDiagnostics(sourceFile);
+  }
+}
+
+function createTypeCheckingProgramStrategy(project: ts.server.Project):
+    TypeCheckingProgramStrategy {
+  return {
+    getProgram(): ts.Program {
+      const program = project.getLanguageService().getProgram();
+      if (!program) {
+        throw new Error('Language service does not have a program!');
+      }
+      return program;
+    },
+    updateFiles(contents: Map<AbsoluteFsPath, string>, updateMode: UpdateMode) {
+      if (updateMode !== UpdateMode.Complete) {
+        throw new Error(`Incremental update mode is currently not supported`);
+      }
+      for (const [fileName, newText] of contents) {
+        const scriptInfo = getOrCreateTypeCheckScriptInfo(project, fileName);
+        const snapshot = scriptInfo.getSnapshot();
+        const length = snapshot.getLength();
+        scriptInfo.editContent(0, length, newText);
+      }
+    },
+  };
+}
+
+function getOrCreateTypeCheckScriptInfo(
+    project: ts.server.Project, tcf: string): ts.server.ScriptInfo {
+  // First check if there is already a ScriptInfo for the tcf
+  const {projectService} = project;
+  let scriptInfo = projectService.getScriptInfo(tcf);
+  if (!scriptInfo) {
+    // ScriptInfo needs to be opened by client to be able to set its user-defined
+    // content. We must also provide file content, otherwise the service will
+    // attempt to fetch the content from disk and fail.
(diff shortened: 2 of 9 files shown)


packages/language-service/ivy/compiler/compiler_host.ts
```diff
@@ -0,0 +1,103 @@
...
+          const path = project.projectService.toPath(fileName);
+          return project.getSourceFile(path);
+        },
+    getSourceFileByPath(
+        fileName: string, path: ts.Path, languageVersion: ts.ScriptTarget,
+        onError?: (message: string) => void, shouldCreateNewSourceFile?: boolean): ts.SourceFile |
+        undefined {
+          return project.getSourceFile(path);
+        },
+    getCancellationToken(): ts.CancellationToken {
```
ayazhafiz, line 42: Is this a copy of the implementation in the project? If so, maybe `return project.getCancellationToken.bind(project)` would be better.

packages/language-service/ivy/language_service.ts
```diff
@@ -6,12 +6,31 @@
...
+  }
 
   getSemanticDiagnostics(fileName: string): ts.Diagnostic[] {
-    return [];
+    const program = this.tsLS.getProgram()!;
+    const sourceFile = program.getSourceFile(fileName);
+    if (!sourceFile) {
+      return [];
+    }
+    const newProgram = this.compiler.analyze(program);
```
ayazhafiz, line 30: Do you anticipate using the program outside the compiler, in the language service? I don’t think we should have to, at least not directly. If not I think it would be better to return a flag of whether analyze succeeded.

packages/language-service/ivy/test/diagnostic_spec.ts
```diff
@@ -0,0 +1,21 @@
...
+
+import {APP_COMPONENT, setup} from './mock_host';
+
+describe('diagnostic', () => {
+  const {project, tsLS} = setup();
+  const ngLS = new LanguageService(project, tsLS);
+
+  it('should not produce error for AppComponent', () => {
+    const diags = ngLS.getSemanticDiagnostics(APP_COMPONENT);
+    expect(diags).toEqual([]);
```
ayazhafiz, line 19: 🎉

packages/language-service/ivy/test/mock_host.ts
```diff
@@ -35,14 +35,20 @@ export const TSCONFIG = join(PROJECT_DIR, 'tsconfig.json');
...
 export const host: ts.server.ServerHost = {
   ...ts.sys,
   readFile(absPath: string, encoding?: string): string |
       undefined {
-        // TODO: Need to remove all annotations in templates like we do in
-        // MockTypescriptHost
-        return ts.sys.readFile(absPath, encoding);
+        const content = ts.sys.readFile(absPath, encoding);
+        if (content === undefined) {
+          return content;
```
ayazhafiz, line 46: nit

```suggestion
          return undefined;
```

packages/language-service/ivy/language_service.ts
ayazhafiz: This is pedantic, I think we can leave it for now, but just something to think about

packages/language-service/ivy/test/mock_host.ts
```diff
@@ -75,3 +81,14 @@ export function setup() {
   const tsLS = project.getLanguageService();
   return {projectService, project, tsLS};
 }
+
+const REF_MARKER = /«(((\w|\-)+)|([^ᐱ]*ᐱ(\w+)ᐱ.[^»]*))»/g;
+const LOC_MARKER = /\~\{(\w+(-\w+)*)\}/g;
+
+function removeReferenceMarkers(value: string): string {
+  return value.replace(REF_MARKER, (_, text) => text.replace(/ᐱ/g, ''));
```
ayazhafiz, line 89: why is the callback needed?

packages/language-service/ivy/compiler/compiler.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  constructor(private readonly project: ts.server.Project, private options: CompilerOptions) {
+    this.tsCompilerHost = makeCompilerHostFromProject(project);
+    const ngCompilerHost =
+        NgCompilerHost.wrap(this.tsCompilerHost, project.getRootFiles(), options);
+    this.program = injectTypeCheckFile(project, ngCompilerHost.typeCheckFile);
+    this.compiler = new NgCompiler(ngCompilerHost, options, this.program);
+  }
+
+  setCompilerOptions(options: CompilerOptions) {
```
ayazhafiz, line 29: This looks unused. Are you trying to fill an interface?

ayazhafiz (changes requested):

packages/language-service/ivy/compiler/compiler_host.ts
kyliau: `project.getCancellationToken()` returns `HostCancellationToken` which is  of the shape
```
    interface HostCancellationToken {
        isCancellationRequested(): boolean;
    }
```
that only partially implements `ts.CancellationToken`

packages/language-service/ivy/compiler/compiler.ts
```diff
@@ -0,0 +1,82 @@
...
+    this.options = options;
+  }
+
+  analyze(program: ts.Program): ts.Program|undefined {
+    const inputFiles = this.project.getRootFiles();
+    const ngCompilerHost = NgCompilerHost.wrap(this.tsCompilerHost, inputFiles, this.options);
```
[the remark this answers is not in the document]
kyliau: It is somewhat cheap to recreate, but this is just a temporary state.
In the final form of the compiler there's no need to create a host every time `analyze` is called.

> Does the type check file in this host == the file in the old host? If so, does that mean the file path is always the same?

Yes, the path is hard-coded in the host. But this too will be going away soon after incremental tcf lands.

packages/language-service/ivy/compiler/compiler.ts
kyliau: This should be called by language service when the config file is changed. I've added TODO here.

packages/language-service/ivy/compiler/compiler.ts
```diff
@@ -0,0 +1,84 @@
...
+    const ngCompilerHost =
+        NgCompilerHost.wrap(this.tsCompilerHost, project.getRootFiles(), options);
+    this.program = injectTypeCheckFile(project, ngCompilerHost.typeCheckFile);
+    this.compiler = new NgCompiler(ngCompilerHost, options, this.program);
+  }
+
+  setCompilerOptions(options: CompilerOptions) {
+    // TODO: Language service must call this method when it detects that the
+    // config file has changed.
+    this.options = options;
```
ayazhafiz, line 32: I think this can be added now since #36922 landed

packages/language-service/ivy/compiler/compiler.ts
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  analyze(program: ts.Program): ts.Program|undefined {
+    const inputFiles = this.project.getRootFiles();
+    const ngCompilerHost = NgCompilerHost.wrap(this.tsCompilerHost, inputFiles, this.options);
+    const oldProgram = this.program;
```
ayazhafiz, line 38: I think it would be better to name the program stored in this class `oldProgram` or `lastKnownProgram`

kyliau: I think the latter is better, I've changed it to `lastKnownProgram`

ayazhafiz (approved):

packages/language-service/ivy/BUILD.bazel
```diff
@@ -6,7 +6,18 @@ ts_library(
     name = "ivy",
     srcs = glob(["*.ts"]),
     deps = [
+        ":compiler",
         "//packages/compiler-cli",
         "@npm//typescript",
     ],
 )
+
+ts_library(
```
josephperrott, line 15: Can you move this `ts_library` into the `compiler/` subdirectory?  Long term we are looking to have small buildable units and it will be easiest to manage if we have the BUILD files as close to the sources as we can.

kyliau: done :)

josephperrott (approved):
LGTM, thanks!

packages/language-service/ivy/compiler/compiler.ts
```diff
@@ -0,0 +1,119 @@
...
+    TypeCheckingProgramStrategy {
+  return {
+    getProgram(): ts.Program {
+      const program = project.getLanguageService().getProgram();
+      if (!program) {
+        throw new Error('Language service does not have a program!');
+      }
+      return program;
+    },
+    updateFiles(contents: Map<AbsoluteFsPath, string>, updateMode: UpdateMode) {
```
ayazhafiz, line 81: Ah, so the compiler passes back to us the TCF updates. Nice!

packages/language-service/ivy/compiler/compiler.ts
```diff
@@ -0,0 +1,119 @@
...
+        ts.server.toNormalizedPath(tcf),
+        true,              // openedByClient
+        '',                // fileContent
+        ts.ScriptKind.TS,  // scriptKind
+    );
+    if (!scriptInfo) {
+      throw new Error(`Failed to create script info for ${tcf}`);
+    }
+  }
+  // Now that we have the ScriptInfo, check if it's part of the project
```
ayazhafiz, line 114: Why does the tcf script need to be registered to the project here? Is this so that it is picked up in the program on a later analysis request?

I think this comment could be a bit better, maybe like "Add ScriptInfo to project if it's missing".

packages/language-service/ivy/test/diagnostic_spec.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  it('should report member does not exist', () => {
+    const content = service.overwriteInlineTemplate(APP_COMPONENT, '{{ nope }}');
+    const diags = ngLS.getSemanticDiagnostics(APP_COMPONENT);
+    expect(diags.length).toBe(1);
+    const {category, file, start, length, messageText} = diags[0];
+    expect(category).toBe(ts.DiagnosticCategory.Error);
+    expect(file?.fileName).toBe(APP_COMPONENT);
+    expect(content.substring(start!, start! + length!)).toBe('nope');
+    expect(messageText).toBe(`Property 'nope' does not exist on type 'AppComponent'.`);
```
ayazhafiz, line 36: wow!!

packages/language-service/ivy/compiler/compiler.ts
kyliau: There are two reasons:
1. You're right, we attach it to the project so that it'll be part of the program.
2. `ScriptInfo` that is not part of any project is known as "orphan" and will be closed by the service.

ayazhafiz (approved):

PR #36936, opened 2020-05-05
ci: removing CI environment variable caching setup

josephperrott:
A caching mechanism was put in place to prevent repeated calls to
the Github API.  As the CI setup no longer relies on calls to the
Github API, this caching is no longer necessary.

It was discovered that this caching was causing a contention issue
for saucelabs testing as the same tunnel was being reused for
multiple jobs simultaneously.  With this caching mechanism removed
the jobs will once again run via separate tunnels.


devversion (approved):

gregmagolan (approved):
🌮

PR #36939, opened 2020-05-05
feat(core):deprecate NgIterable type

### LINKED ISSUE #36854: Add support for ReadonlyArray in NgIterable
# 🚀 feature request

### Relevant Package

 This feature request is for @angular/core

### Description

```
type NgIterable<T> = Array<T> | Iterable<T>;
```

Please change it into

```
type NgIterable<T> = ReadonlyArray<T> | Iterable<T>;
```

ajitsinghkaler:
typescript contains Array of type ReadonlyArray earlier we only used to support Array type added support for ReadonlyArray too in the core.js file

Fixes #36854

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Added readonlyArray type not in iterable

Issue Number: #36854

## What is the new behavior?
Added readonlyArray type in iterable

- [ ] Yes
- [x] No

### DIFF
--- a/aio/content/guide/deprecations.md
+++ b/aio/content/guide/deprecations.md
@@ -61,8 +61,9 @@ v9 - v12
 | `@angular/router`             | [`ActivatedRoute` params and `queryParams` properties](#activatedroute-props) | unspecified |
 | template syntax               | [`/deep/`, `>>>`, and `::ng-deep`](#deep-component-style-selector)            | <!--v7--> unspecified |
 | browser support               | [`IE 9 and 10, IE mobile`](#ie-9-10-and-mobile)                               | <!--v10--> v11 |
-
+| `@angular/core`               | [`NgIterable`](api/core/NgIterable)                        | unspecified |
 For information about Angular CDK and Angular Material deprecations, see the [changelog](https://github.com/angular/components/blob/master/CHANGELOG.md).
+
 
 ## Deprecated APIs
 
@@ -97,6 +98,7 @@ Tip: In the [API reference section](api) of this doc site, deprecated APIs are i
 | [`entryComponents`](api/core/NgModule#entryComponents) | none | v9 | See [`entryComponents`](#entryComponents) |
 | [`ANALYZE_FOR_ENTRY_COMPONENTS`](api/core/ANALYZE_FOR_ENTRY_COMPONENTS) | none | v9 | See [`ANALYZE_FOR_ENTRY_COMPONENTS`](#entryComponents) |
 | [`WrappedValue`](api/core/WrappedValue) | none | v10 | See [removing `WrappedValue`](#wrapped-value) |
+| [`NgIterable`](api/core/NgIterable) | Iterabe | unspecified | Use the default Iterable type instead of NgIterable |
 
 
 
--- a/packages/compiler-cli/src/ngtsc/typecheck/src/type_emitter.ts
+++ b/packages/compiler-cli/src/ngtsc/typecheck/src/type_emitter.ts
@@ -67,26 +67,24 @@ export function canEmitType(type: ts.TypeNode, resolver: TypeReferenceResolver):
  * For example, consider the following code:
  *
  * ```
- * import {NgIterable} from '@angular/core';
- *
- * class NgForOf<T, U extends NgIterable<T>> {}
+ * class NgForOf<T, U extends Iterable<T>> {}
  * ```
  *
  * Here, the generic type parameters `T` and `U` can be emitted into a different context, as the
- * type reference to `NgIterable` originates from an absolute module import so that it can be
+ * type reference to `Iterable` originates from an absolute module import so that it can be
  * emitted anywhere, using that same module import. The process of emitting translates the
- * `NgIterable` type reference to a type reference that is valid in the context in which it is
+ * `Iterable` type reference to a type reference that is valid in the context in which it is
  * emitted, for example:
  *
  * ```
  * import * as i0 from '@angular/core';
  * import * as i1 from '@angular/common';
  *
- * const _ctor1: <T, U extends i0.NgIterable<T>>(o: Pick<i1.NgForOf<T, U>, 'ngForOf'>):
+ * const _ctor1: <T, U extends Iterable<T>>(o: Pick<i1.NgForOf<T, U>, 'ngForOf'>):
  * i1.NgForOf<T, U>;
  * ```
  *
- * Notice how the type reference for `NgIterable` has been translated into a qualified name,
+ * Notice how the type reference for `Iterable` has been translated into a qualified name,
  * referring to the namespace import that was created.
  */
 export class TypeEmitter {
--- a/packages/core/src/change_detection/differs/iterable_differs.ts
+++ b/packages/core/src/change_detection/differs/iterable_differs.ts
@@ -15,10 +15,11 @@ import {DefaultIterableDifferFactory} from '../differs/default_iterable_differ';
 
 /**
  * A type describing supported iterable types.
- *
+ * @deprecated Use `Iterable` from the standard TypeScript library instead of `NgIterable`. The two
+ *     are equivalent.
  * @publicApi
  */
-export type NgIterable<T> = Array<T>|Iterable<T>;
+export type NgIterable<T> = Iterable<T>;
 
 /**
  * A strategy for tracking changes over time to an iterable. Used by {@link NgForOf} to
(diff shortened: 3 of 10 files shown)


ajitsinghkaler:
Can someone please explain what are golden files which cause the ci set up to fail

gkalpak:
@ajitsinghkaler, the failure message should contain a link to some docs and the command you need to run to update the golden files: [docs/PUBLIC_API.md#golden-files][1]

That said, this reminds me that the infrastructure has changed slightly (with PRs such as #35768 and #36630) and the docs have not be updated (https://github.com/angular/angular/pull/35768#issuecomment-597719247 :grin:). Feel free to send a PR to update the docs too, to match the new error output and fix commands (and the location where the golden files are kept). 

[1]: https://github.com/angular/angular/blob/b6bc0aeb611c6d3577fb9d90cfaa8363ecf5a8aa/docs/PUBLIC_API.md#golden-files

packages/compiler-cli/src/ngtsc/typecheck/test/test_utils.ts
```diff
@@ -93,7 +93,7 @@ export function angularCoreDts(): TestFile {
       subscribe(generatorOrNext?: any, error?: any, complete?: any): unknown;
     }
     
-    export declare type NgIterable<T> = Array<T> | Iterable<T>;
+    export declare type NgIterable<T> = Array<T> | Iterable<T> | ReadonlyArray<T>;
```
pkozlowski-opensource, line 96: Do we need _both_ `ReadonlyArray` _and_ `Array`? More importantly, wouldn't `Iterable<T>` be enough? Aren't all Array  being ReadonlyArray and all ReadonlyArray being Iterable?

ajitsinghkaler: @pkozlowski-opensource yes they are but earlier they were added separately. S, I thought this was because of some implementation detail that I don't know. If its not I'll remove this

pkozlowski-opensource: @ajitsinghkaler yeh, I must say that I don't have the full historical context either and _maybe_ there are reasons that I don't understand. Having said this it seems like tests are passing with your change, which is a good sign :-) 

Now, looking at the code after your change, I wonder why do we need `NgIterable`  at all???

ajitsinghkaler:
@IgorMinar @pkozlowski-opensource @alxhub is this change needed

packages/compiler-cli/src/ngtsc/typecheck/test/test_utils.ts
jelbourn: Looking at this, I agree that there doesn't seem to be a reason for `NgIterable` to exist based on the current TypeScript landscape. My best guess is that some earlier version of TypeScript had some subtle difference between `Array` and `Iterable` that made this necessary. I think the right move would be to deprecate `NgIterable` and remove it in a future version.

I did a quick look, and there are ~17 uses of `NgIterable` around Google that would need to be cleaned up.

ajitsinghkaler:
@IgorMinar @alxhub @pkozlowski-opensource @jelbourn why is component repo unit tests failing and how to make them right there is nothing about this in developer.md in the repo

pkozlowski-opensource:
@ajitsinghkaler see: https://github.com/angular/components/blob/master/src/cdk/scrolling/virtual-for-of.ts#L25 - looks like CDK is using `NgIterable` - which tells me that this is a public API (at least in practice) and we can't / should not remove it right now? I guess we should still make all the changes but keep the `NgIterable` and deprecate it.

packages/compiler-cli/src/ngtsc/typecheck/test/type_parameter_emitter_spec.ts
```diff
@@ -73,36 +73,6 @@ runInEachFileSystem(() => {
           .toEqual('<T extends string[]>');
     });
 
-    it('can emit references into external modules', () => {
```
IgorMinar, line 76: we can't remove these tests - we have to modify them to test for a presence of another symbol similar to NgIterable.

packages/core/src/change_detection/differs/iterable_differs.ts
```diff
@@ -12,14 +12,6 @@ import {Optional, SkipSelf} from '../../di/metadata';
 import {DefaultIterableDifferFactory} from '../differs/default_iterable_differ';
 
 
-
-/**
- * A type describing supported iterable types.
- *
- * @publicApi
- */
-export type NgIterable<T> = Array<T>|Iterable<T>;
```
IgorMinar, line 21: we can't remove this type as that would be a breaking change, but we can deprecate it. Please mark it as `@deprecated` like this:

```
/**	
 * A type describing supported iterable types.	
 *	
 * @deprecated Use `Iterable` from the standard TypeScript library instead of `NgIterable`. The two are equivalent.
 * @publicApi	
 */	
export type NgIterable<T> = Array<T>|Iterable<T>;
```

IgorMinar (changes requested):
overal LGTM, except that we can't remove this API we do have to deprecate it as this would be a breaking change.

In addition to marking the api as `@deprecated` please add it to `deprecations.md`. Thanks

Reviewed-for: public-api

IgorMinar (changes requested):
Reviewed-for: public-api

JoostK:
As this touches the generic type constraint of `NgForOf`, we'd have to verify that it does not cause a de-opt in the type checker that was achieved in #34021. I.e. changing `U extends NgIterable` to `U extends Iterable` may de-opt as `Iterable` is a global ambient type.

packages/compiler-cli/src/ngtsc/typecheck/test/type_parameter_emitter_spec.ts
ajitsinghkaler: @IgorMinar as far as I can search we have no symbol like NgIterable in public api

aio/content/guide/deprecations.md
```diff
@@ -97,6 +98,7 @@ Tip: In the [API reference section](api) of this doc site, deprecated APIs are i
 | [`entryComponents`](api/core/NgModule#entryComponents) | none | v9 | See [`entryComponents`](#entryComponents) |
 | [`ANALYZE_FOR_ENTRY_COMPONENTS`](api/core/ANALYZE_FOR_ENTRY_COMPONENTS) | none | v9 | See [`ANALYZE_FOR_ENTRY_COMPONENTS`](#entryComponents) |
 | [`WrappedValue`](api/core/WrappedValue) | none | v10 | See [removing `WrappedValue`](#wrapped-value) |
+| [`NgIterable`](api/core/NgIterable) | Iterabe | unspecified | Use the default Iterable type instead of NgIterable |
```
rafaelss95, line 101:
```suggestion
| [`NgIterable`](api/core/NgIterable) | Iterable | unspecified | Use the default Iterable type instead of NgIterable |
```

JoostK:
I just tested how we deal with global ambient types in the type emitter and they are not supported at the moment. Because of this, the changed generic type constraint on `NgForOf` would have a performance impact on template type checking. The type constructor for `NgForOf` will no longer be declared in each generated template type check file, but instead the actual declaration in the `@angular/common` declaration file has to be updated, resulting in a more expensive rebuild.

I will mark this as blocked per the above, I'll see if I can extend the type emitter logic in the coming week.

packages/compiler-cli/src/ngtsc/typecheck/src/type_emitter.ts
```diff
@@ -67,26 +67,24 @@ export function canEmitType(type: ts.TypeNode, resolver: TypeReferenceResolver):
  * For example, consider the following code:
  *
  * ```
- * import {NgIterable} from '@angular/core';
- *
- * class NgForOf<T, U extends NgIterable<T>> {}
+ * class NgForOf<T, U extends Iterable<T>> {}
```
JoostK, line 70: This comment should not be updated like this, as it invalidates the situation for which this comment explains how the type emitter works. Instead, I think we should just introduce a different type that needs to be imported from a namespace.

JoostK (changes requested):
I just tested how we deal with global ambient types in the type emitter and they are not supported at the moment. Because of this, the changed generic type constraint on `NgForOf` would have a performance impact on template type checking. The type constructor for `NgForOf` will no longer be declared in each generated template type check file, but instead the actual declaration in the `@angular/common` declaration file has to be updated, resulting in a more expensive rebuild.

I will mark this as blocked per the above, I'll see if I can extend the type emitter logic in the coming week.

packages/compiler-cli/src/ngtsc/typecheck/src/type_emitter.ts
ajitsinghkaler: I did not find any other type like NgIterabe the nearest is HttpEvent but that is in @angular/common So what to use instead of NgIterable

Issue #36941, opened 2020-05-05
i18n within foreignObject doesn't work with nested tags


Goodwine:
# 🐞 bug report

### Affected Package

@angular/compiler
Angular i18n

### Is this a regression?

No

### Description
An `svg` element with `foreignObject` can contain any kind of HTML tags like `<div>` or `<span>`.
However If a `span` is nested inside a `div` that contains `i18n` the build breaks.

## 🔬 Minimal Reproduction

https://stackblitz.com/edit/angular-va9pj8?file=src%2Fapp%2Fapp.component.html
See the comment about removing `i18n`

## 🔥 Exception or Error
```
Error in /turbo_modules/@angular/compiler@9.1.0/bundles/compiler.umd.js (21449:23)

Unexpected closing tag "xhtml:span". It may happen when the tag has already been closed by another tag. For more info see https://www.w3.org/TR/html5/syntax.html#closing-elements-that-have-implied-end-tags ("Lorem ipsum dolor sit amet, <:xhtml:span class="foo">rutrum[ERROR ->]</:xhtml:span> imperdiet eros."): ng:///AppModule/AppComponent.html@0:59
```

## 🌍  Your Environment

**Angular Version:**
9.x

**Anything else relevant?**
Existing workaround is to move the HTML from within `<foreignObject>` to an `<ng-template>`
https://stackblitz.com/edit/angular-hfqm3v?file=src/app/app.component.html
```diff
<svg:g>
  ...
  <foreignObject>
    <xhtml:div>
      ...
-      <div i18n="asd">x<span>{{foo.bar.baz //i18n(ph="name")}}</span></div>
+      <ng-container *ngTemplateOutlet="myTemplate; context: {name:foo.bar.baz}">
+      </ng-container>
    </xhtml:div>
  </foreignObject>
</svg:g>
+<ng-template #myTemplate let-name="name">
+  <div i18n="asd">x<span>{{name //i18n(ph="name")}}</span></div>
+</ng-template>
```

AndrewKushnir:
Hi @Goodwine,

Thanks for reporting the issue. I've performed additional investigation and found out that this problem affects View Engine only and the same use-case works correctly in Ivy. I've created https://github.com/angular/angular/pull/36943 with additional tests to confirm that.

Thank you.

petebacondarwin:
Since ViewEngine is deprecated and this is fixed in Ivy, I am going to close this.

PR #36943, opened 2020-05-06
test(core): verify that Ivy i18n works correctly with HTML namespaces

AndrewKushnir:
This commit adds several tests to verify that i18n logic in Ivy handles elements with HTML namespaces correctly.

Related to #36941.

## PR Type

- [x] Other... Please describe: test coverage improvement

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/test/acceptance/i18n_spec.ts
+++ b/packages/core/test/acceptance/i18n_spec.ts
@@ -9,15 +9,15 @@
 // below. This would normally be done inside the application `polyfills.ts` file.
 import '@angular/localize/init';
 
-import {CommonModule, registerLocaleData} from '@angular/common';
+import {CommonModule, DOCUMENT, registerLocaleData} from '@angular/common';
 import localeEs from '@angular/common/locales/es';
 import localeRo from '@angular/common/locales/ro';
 import {computeMsgId} from '@angular/compiler';
-import {Component, ContentChild, ContentChildren, Directive, ElementRef, HostBinding, Input, LOCALE_ID, NO_ERRORS_SCHEMA, Pipe, PipeTransform, QueryList, TemplateRef, Type, ViewChild, ViewContainerRef} from '@angular/core';
+import {Component, ContentChild, ContentChildren, Directive, ElementRef, HostBinding, Input, LOCALE_ID, NO_ERRORS_SCHEMA, Pipe, PipeTransform, QueryList, RendererFactory2, TemplateRef, Type, ViewChild, ViewContainerRef, ɵsetDocument} from '@angular/core';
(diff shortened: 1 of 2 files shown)


packages/core/test/acceptance/i18n_spec.ts
```diff
@@ -530,6 +530,75 @@ onlyInIvy('Ivy i18n logic').describe('runtime i18n', () => {
...
+        ɵsetDocument(document);
+        return document;
+      }
+
+      TestBed.configureTestingModule({
+        providers: [
+          {provide: DOCUMENT, useFactory: _document, deps: []},
+          // TODO(FW-811): switch back to default server renderer (i.e. remove the line below)
+          // once it starts to support Ivy namespace format (URIs) correctly. For now, use
+          // `DomRenderer` that supports Ivy namespace format.
```
petebacondarwin, line 546: So this is saying that we don't support namespaces on tags when doing SSR?
If so, I agree that this is outside the scope of this PR.

packages/core/test/acceptance/i18n_spec.ts
```diff
@@ -530,6 +530,75 @@ onlyInIvy('Ivy i18n logic').describe('runtime i18n', () => {
...
+            </xhtml:div>
+          </foreignObject>
+        </svg>
+      `);
+
+      const element = fixture.nativeElement;
+      expect(element.textContent.trim()).toBe('Bonjour monde');
+      expect(element.querySelector('svg').namespaceURI).toBe('http://www.w3.org/2000/svg');
+      expect(element.querySelector('div').namespaceURI).toBe('http://www.w3.org/1999/xhtml');
+      expect(element.querySelector('span').namespaceURI).toBe('http://www.w3.org/1999/xhtml');
```
petebacondarwin, line 575: Interesting... so the `span` inherits its namespace from its containing `div`. I didn't know that. But now it seems obvious.

petebacondarwin (approved):

packages/core/test/acceptance/i18n_spec.ts
AndrewKushnir: Hi @petebacondarwin. Yes, this is a pre-existing problem with default server renderer, [see comment here](https://github.com/angular/angular/blob/master/packages/platform-server/src/server_renderer.ts#L76-L78).

atscott (approved):

AndrewKushnir:
Merge assistance: the only missing approval is `fw-compiler` that I'm a member of. This change is not risky (adding new tests only) and Andrew and Pete LGTM'ed it, so we should be OK to move forward with this change.

PR #36892, opened 2020-05-01
refactor(core): make generic mandatory for ModuleWithProviders

kara:
In v9, we deprecated the use of ModuleWithProviders
without a generic. In v10, we will be requiring the
generic when using ModuleWithProviders. You can read
more about the reasoning behind this change in the
migration guide:
http://v9.angular.io/guide/migration-module-with-providers

PR #36948, opened 2020-05-06
ngcc: fix ModuleWithProviders function parsing

petebacondarwin:
This unblocks removal of the default type (`any`) on the `ModuleWithProviders` type. (#36892)


packages/compiler-cli/ngcc/src/analysis/module_with_providers_analyzer.ts
```diff
@@ -108,14 +207,13 @@ export class ModuleWithProvidersAnalyzer {
...
-    if (!ts.isClassDeclaration(dtsNgModule) || !hasNameIdentifier(dtsNgModule)) {
+    if (!isNamedClassDeclaration(dtsNgModule)) {
       throw new Error(`The referenced NgModule in ${
           fn.declaration
               .getText()} is not a named class declaration in the typings program; instead we get ${
           dtsNgModule.getText()}`);
     }
-
-    return {node: dtsNgModule, known: null, viaModule: null, identity: null};
+    return new Reference(dtsNgModule, ngModule.bestGuessOwningModule);
```
alxhub, line 216:
```suggestion
    return new Reference(dtsNgModule, /* owningModule */ null);
```

This is always going to be `null` I believe, because of the early return above.

alxhub (approved):

packages/compiler-cli/ngcc/src/analysis/module_with_providers_analyzer.ts
```diff
@@ -122,47 +136,37 @@ export class ModuleWithProvidersAnalyzer {
...
-    const returnExpression =
-        lastStatement && ts.isReturnStatement(lastStatement) && lastStatement.expression || null;
-    const ngModuleProperty = returnExpression && ts.isObjectLiteralExpression(returnExpression) &&
-            returnExpression.properties.find(
-                prop =>
-                    !!prop.name && ts.isIdentifier(prop.name) && prop.name.text === 'ngModule') ||
-        null;
-
-    if (!ngModuleProperty || !ts.isPropertyAssignment(ngModuleProperty)) {
+    if (lastStatement === null || !ts.isReturnStatement(lastStatement) ||
```
JoostK, line 143: `lastStatement` could potentially be `undefined` for empty bodies:

```suggestion
    if (lastStatement == null || !ts.isReturnStatement(lastStatement) ||
```

packages/compiler-cli/ngcc/src/analysis/module_with_providers_analyzer.ts
```diff
@@ -122,47 +136,37 @@ export class ModuleWithProvidersAnalyzer {
...
+        lastStatement.expression == undefined) {
```
JoostK, line 144: but this one can't, so this could be

```suggestion
        lastStatement.expression === undefined) {
```

JoostK (approved):
LGTM. Could you please clarify the commit message of the first commit to mention that the moved code is still tested using existing tests for the `ModuleWithProvidersAnalyzer`.

packages/compiler-cli/ngcc/src/analysis/module_with_providers_analyzer.ts
```diff
@@ -68,6 +68,100 @@ export class ModuleWithProvidersAnalyzer {
...
+
+    // The ngModuleValue could be of the form `SomeModule` or `namespace_1.SomeModule`
+    let ngModuleValue = ngModuleProperty.initializer;
+    if (ts.isPropertyAccessExpression(ngModuleValue)) {
+      ngModuleValue = ngModuleValue.expression;
+    }
+
+    if (!ts.isIdentifier(ngModuleValue)) {
+      return null;
+    }
```
gkalpak: This part of the function seems to have changed. My understanding is that this commit should only move things around. Was this change intentionally included in this commit?

kara:
@petebacondarwin  @joostK @alxhub As discussed, let's do cleanup in a follow-up PR

packages/compiler-cli/ngcc/src/host/ngcc_host.ts
```diff
@@ -19,30 +19,6 @@ export function isSwitchableVariableDeclaration(node: ts.Node):
       ts.isIdentifier(node.initializer) && node.initializer.text.endsWith(PRE_R3_MARKER);
 }
 
-/**
- * A structure returned from `getModuleWithProviderInfo` that describes functions
- * that return ModuleWithProviders objects.
- */
-export interface ModuleWithProvidersFunction {
```
gkalpak, line 26: This is still imported in `module_with_providers_analyzer.ts`: https://github.com/angular/angular/blob/6d6f281c9e5a4a31604aab01a45eb9dcc383c31f/packages/compiler-cli/ngcc/src/analysis/module_with_providers_analyzer.ts#L13

Did you mean to remove it in a subsequent commit?

gkalpak (changes requested):

packages/compiler-cli/ngcc/src/analysis/module_with_providers_analyzer.ts
petebacondarwin: I think that was a rebase mistake on my part. I'll revert that here.

petebacondarwin: Actually, no it was on purpose. The original code used an internal helper `getDeclarationOfExpression()`, which is not available outside.

Also it turns out that `getDeclarationOfExpression()` just calls `getDeclarationOfIdentifier()` if the expression is an identifier, which in this case we already know is true.

So rather than exposing an internal helper into `NgccReflectionHost` I just called the function directly.

packages/compiler-cli/ngcc/src/host/ngcc_host.ts
petebacondarwin: It should be moved to the analyzer in this commit. Thanks.

packages/compiler-cli/ngcc/src/analysis/module_with_providers_analyzer.ts
petebacondarwin: Going for:

```
    const body = definition.body;
    if (body === null || body.length === 0) {
      return null;
    }
```

packages/compiler-cli/ngcc/src/analysis/module_with_providers_analyzer.ts
```diff
@@ -137,11 +137,15 @@ export class ModuleWithProvidersAnalyzer {
...
     }
 
-    // Get hold of the return statement expression for the function
     const body = definition.body;
+    if (body === null || body.length === 0) {
+      return null;
+    }
+
+    // Get hold of the return statement expression for the function
     const lastStatement = body && body[body.length - 1];
```
gkalpak, line 146: Couldn't this now be:

```suggestion
    const lastStatement = body[body.length - 1];
```

packages/compiler-cli/ngcc/src/analysis/module_with_providers_analyzer.ts
```diff
@@ -137,11 +137,15 @@ export class ModuleWithProvidersAnalyzer {
...
     if (lastStatement === null || !ts.isReturnStatement(lastStatement) ||
```
gkalpak, line 147: Can `lastStatement` be `null` now? (I think not.)

gkalpak (approved):
LGTM ✨

PR #35621, opened 2020-02-21
refactor(ivy): remove style sanitization

matsko:
This patch is the first of many commits to disable sanitization for
[stlye.prop] and [style] bindings in Angular.

Historically, style-based sanitization has only been required for old
IE browsers (IE6 and IE7). Since Angular does not support these old
browsers at all, there is no reason for the framework to support
style-based sanitization.

PR #36965, opened 2020-05-06
refactor(core): remove style sanitization code for `[style]`/`[style.prop]` bindings

matsko:
In 420b9be1c14ec6900692b641a3d89b99eb9c6e5b all style-based sanitization code was disabled because modern
browsers no longer allow for javascript expressions within CSS. This
patch is a follow-up patch which removes all traces of style
sanitization code (both instructions and runtime logic) for the
`[style]` and `[style.prop]` bindings.


packages/core/test/acceptance/styling_spec.ts
```diff
@@ -1914,7 +1914,7 @@ describe('styling', () => {
       });
 
   onlyInIvy('only ivy has [style] support')
-      .it('should sanitize style values before writing them', () => {
+      .it('should not sanitize style values before writing them', () => {
```
IgorMinar, line 1917: It's odd that you are changing the description of the spec without changing the spec itself. Is that right?

IgorMinar (approved):
Lgtm

Reviewed-for: public-api

IgorMinar:
BTW awesome cleanup!

packages/core/test/acceptance/styling_spec.ts
matsko: This is here because I forgot to comment about in the [previous PR](https://github.com/angular/angular/pull/35621)

matsko:
https://test.corp.google.com/ui#id=OCL:310984540:BASE:310984610:1589230032483:42d8f9f6

matsko:
presubmit = https://test.corp.google.com/ui#id=OCL:310984540:BASE:310984610:1589230032483:42d8f9f6

mhevery (approved):

alxhub (approved):

packages/core/src/sanitization/bypass.ts
```diff
@@ -143,8 +143,8 @@ export function bypassSanitizationTrustHtml(trustedHtml: string): SafeHtml {
 /**
  * Mark `style` string as trusted.
  *
- * This function wraps the trusted string in `String` and brands it in a way which makes it
- * recognizable to {@link styleSanitizer} to be trusted implicitly.
+ * Note that this function is no longer necessary since Angular doesn't perform
+ * style sanitization anymore.
  *
```
kara, line 148: Remind me what the timeline is for deprecating this?  It seems like we're almost doing it with this statement.

kara (approved):
LGTM

packages/core/src/sanitization/bypass.ts
matsko: https://hackmd.io/_sF2TzXLSiuuYbY88MpdQA#Step-2-Post-Angular-v10

kara: Hmm, that doesn't really answer my question. Are you intending to deprecate this with this statement or no? If not, I think we should remove the text until we can properly depreate it.

PR #36987, opened 2020-05-07
fix(compiler-cli): don't try to tag non-ts files as shims

alxhub:
Some projects include .js source files (via the TypeScript allowJs option).
Previously, the compiler would attempt to tag these files for shims, which
caused errors as the regex used to create shim filenames assumes a .ts file.
This commit fixes the bug by filtering out non-ts files during tagging.

### DIFF
--- a/packages/compiler-cli/test/ngtsc/env.ts
+++ b/packages/compiler-cli/test/ngtsc/env.ts
@@ -59,6 +59,7 @@ export class NgtscTestEnvironment {
         "outDir": "built",
         "rootDir": ".",
         "baseUrl": ".",
+        "allowJs": true,
         "declaration": true,
         "target": "es5",
         "newLine": "lf",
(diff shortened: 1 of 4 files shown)


petebacondarwin (approved):

kara (approved):
LGTM

packages/compiler-cli/test/ngtsc/env.ts
```diff
@@ -59,6 +59,7 @@ export class NgtscTestEnvironment {
         "outDir": "built",
         "rootDir": ".",
         "baseUrl": ".",
+        "allowJs": true,
```
JoostK, line 62: I would move this local to the test, as the test continues to pass without this flag (so anyone removing this in the future won't be able to see the effect of it, but it would render the test useless.)

JoostK (approved):

PR #37003, opened 2020-05-08
Ngtsc revert filesystem stuff

petebacondarwin:


JoostK (approved):

IgorMinar (commented):
Do we have a reproduction that confirms that this is the culprit?

Also, can you please update the commit messages with the rollback reason? Thanks!

petebacondarwin:
Commit messages updated.

JoostK:
I can reproduce the issue on Windows and both this PR as well as #37008 solve it for me.

AndrewKushnir (approved):

AndrewKushnir:
@petebacondarwin I just noticed that the "PR target" label is missing: I believe it should be "master&patch"?

petebacondarwin:
Thanks for spotting that @AndrewKushnir - no this is patch only.

AndrewKushnir:
@petebacondarwin ok, got it. For the "10.0.0-next.X" release the plan is to fix forward with #37008?

petebacondarwin:
Yes.

AndrewKushnir:
FYI, g3 presubmit should not be needed for this PR, since it's a patch-only (i.e. we keep these changes in master) and g3 sync happens from the master branch.

petebacondarwin:
Landed on 9.1.x

PR #37015, opened 2020-05-08
fix(docs-infra): better detect failed attempts to update the preview server Docker container

gkalpak:
In order to avoid unnecessary operations, the `update-preview-server.sh` script, that is used to update the PR preview server Docker container, will only try to update the Docker container if either any files in the `aio/aio-builds-setup/` directory have changed since the last update or if a previous update failed. A failed previous update is detected by checking whether the temporary `aio-builds:provisional` Docker image still exists. This temporary image is created during the update operation and is renamed to `aio-builds:latest` if the update goes well.

Previously, the update script was not able to detect a previous failed attempt if the operation failed before creating the `aio-builds:provisional` Docker image, such as if the `create-image.sh` script failed. This could lead to a situation where the preview server Docker container would not be updated after a failed attempt.

This commit improves the logic for detecting failed attempts to account for this type of failures. It does this by not removing an older `aio-builds:provisional` Docker image until a new one is successfully created.

NOTE: While this is not full-proof, it is an improvement as it eliminates a certain kind of failures.

##
This PR also includes some docs improvements. See the rspective commit for more details.

### DIFF
--- a/aio/aio-builds-setup/docs/vm-setup--create-docker-image.md
+++ b/aio/aio-builds-setup/docs/vm-setup--create-docker-image.md
@@ -4,9 +4,8 @@
 ## Install git, Node.js and yarn
 - `sudo apt-get update`
 - `sudo apt-get install -y git`
-- Install [nvm](https://github.com/nvm-sh/nvm#installing-and-updating).
-- Install Node.js: `nvm install 12`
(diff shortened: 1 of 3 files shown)


IgorMinar (approved):
thanks @gkalpak!

PR #37017, opened 2020-05-08
Migrate to typescript file for ng-dev config

josephperrott:
See individual commits.


devversion (approved):

gkalpak (commented):
Doesn't this change mean that `ng-dev` will only work correctly if invoked via `ts-node` if the config file is in TS? How is it invoked in other repos (that install it via npm)?

josephperrott:
Correct, it only works to use a typescript file if its invoked via `ts-node`, but since we don't include the file extension, it will still import via javascript as well.

PR #13380, opened 2016-12-11
fix(router): routerLink support of null/undefined

DzmitryShylovich:
Fixes https://github.com/angular/angular/issues/6971

PR #37018, opened 2020-05-08
fix(router): update type for routerLink to include null and undefined

### LINKED ISSUE #36544: routerLink type doesn't support null | undefined
# 🐞 bug report

### Affected Package
@angular/router

### Is this a regression?

Yes, this code works with strictTemplates disabled.

### Description

The router link directive test the value with != null, and handles null & undefined as an empty relative route.  We use this when we want to change the fragment of the URL and stay on the same page.

```<a [routerLink]="" fragment="foo">A Link</a>```

This can be worked around by using:
```<a [routerLink]="[]" fragment="foo">A Link</a>```

But the code does support null and undefined, it's just the type that doesn't.

## 🔬 Minimal Reproduction
https://github.com/james-schwartzkopf/ng9-issues/tree/routerLink-undefined

## 🔥 Exception or Error
```
    ERROR in src/app/app.component.html:1:4 - error TS2322: Type 'undefined' is not assignable
to type 'string | any[]'.
    
    1 <a [routerLink]="" fragment="foo">A Link</a>
         ~~~~~~~~~~~~~~~
    
      src/app/app.component.ts:5:16
                         ~~~~~~~~~~~~~~~~~~~~~~
 ```

## 🌍  Your Environment

**Angular Version:**
<pre><code>
$ ng --version

                |___/
    

Angular CLI: 9.1.0
Node: 12.16.1
OS: win32 x64

Angular: 9.1.0
... router
Ivy Workspace: Yes

@angular-devkit/architect          0.901.0
@angular-devkit/build-angular      0.901.0
@angular-devkit/build-ng-packagr   0.901.0
@angular-devkit/build-optimizer    0.901.0
@angular-devkit/build-webpack      0.901.0
@angular-devkit/core               9.1.0
@angular-devkit/schematics         9.1.0
@ngtools/webpack                   9.1.0
@schematics/angular                9.1.0
@schematics/update                 0.901.0

</code></pre>

atscott:
PR #13380 added support for `null` and `undefined` but the type on the parameter was not updated. This would result in a compilation error if `fullTemplateTypeCheck` is enabled.

Fixes #36544

### DIFF
--- a/packages/router/src/directives/router_link.ts
+++ b/packages/router/src/directives/router_link.ts
@@ -139,7 +139,7 @@ export class RouterLink {
   }
 
   @Input()
-  set routerLink(commands: any[]|string) {
+  set routerLink(commands: any[]|string|null|undefined) {
     if (commands != null) {
       this.commands = Array.isArray(commands) ? commands : [commands];
     } else {
@@ -229,7 +229,7 @@ export class RouterLinkWithHref implements OnChanges, OnDestroy {
   }
 
   @Input()
-  set routerLink(commands: any[]|string) {
+  set routerLink(commands: any[]|string|null|undefined) {
     if (commands != null) {
       this.commands = Array.isArray(commands) ? commands : [commands];
     } else {
--- a/packages/router/src/directives/router_link.ts
+++ b/packages/router/src/directives/router_link.ts
@@ -115,19 +115,49 @@ import {UrlTree} from '../url_tree';
 export class RouterLink {
   // TODO(issue/24571): remove '!'.
   @Input() queryParams!: {[k: string]: any};
+  /**
+   * Passed to `Router.createUrlTree` as part of the `NavigationExtras`.
+   * @see [NavigationExtras](https://angular.io/api/router/NavigationExtras#navigationextras)
+   * @see [Router.createUrlTree](https://angular.io/api/router/Router#createurltree)
+   */
   // TODO(issue/24571): remove '!'.
   @Input() fragment!: string;
+  /**
+   * Passed to `Router.createUrlTree` as part of the `NavigationExtras`.
+   * @see [QueryParamsHandling](https://angular.io/api/router/QueryParamsHandling#queryparamshandling)
+   * @see [NavigationExtras](https://angular.io/api/router/NavigationExtras#navigationextras)
+   * @see [Router.createUrlTree](https://angular.io/api/router/Router#createurltree)
+   */
   // TODO(issue/24571): remove '!'.
   @Input() queryParamsHandling!: QueryParamsHandling;
+  /**
+   * Passed to `Router.createUrlTree` as part of the `NavigationExtras`.
+   * @see [NavigationExtras](https://angular.io/api/router/NavigationExtras#navigationextras)
+   * @see [Router.createUrlTree](https://angular.io/api/router/Router#createurltree)
+   */
   // TODO(issue/24571): remove '!'.
   @Input() preserveFragment!: boolean;
+  /**
+   * Passed to `Router.createUrlTree` as part of the `NavigationExtras`.
+   * @see [NavigationExtras](https://angular.io/api/router/NavigationExtras#navigationextras)
+   * @see [Router.createUrlTree](https://angular.io/api/router/Router#createurltree)
+   */
   // TODO(issue/24571): remove '!'.
   @Input() skipLocationChange!: boolean;
+  /**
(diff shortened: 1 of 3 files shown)


AndrewKushnir (approved):
LGTM 👍

atscott:
[presubmit](https://test.corp.google.com/ui#id=OCL:310590576:BASE:310590602:1588961276540:a823eae6)

packages/router/src/directives/router_link.ts
```diff
@@ -229,7 +229,7 @@ export class RouterLinkWithHref implements OnChanges, OnDestroy {
   }
 
   @Input()
-  set routerLink(commands: any[]|string) {
+  set routerLink(commands: any[]|string|null|undefined) {
```
IgorMinar, line 232: can you please add docs and document what will be the resulting behavior of passing `null` and `undefined` into `[router-link]`? thanks!

IgorMinar (changes requested):

IgorMinar (changes requested):
can you do a light past to add the docs? Please don’t spend too much time on it, even a single sentence per property is better than nothing.

https://next.angular.io/api/router/RouterLink
https://next.angular.io/api/router/RouterLinkWithHref

Thanks!

packages/router/src/directives/router_link.ts
```diff
@@ -115,19 +115,49 @@ import {UrlTree} from '../url_tree';
 export class RouterLink {
   // TODO(issue/24571): remove '!'.
   @Input() queryParams!: {[k: string]: any};
+  /**
+   * Passed to `Router.createUrlTree` as part of the `NavigationExtras`.
```
IgorMinar, line 119: please use `#` instead of `.` in this case. The community convention is that`#` is used for instance methods and `.` is for static methods.
```suggestion
   * Passed to `Router#createUrlTree` as part of the `NavigationExtras`.
```

packages/router/src/directives/router_link.ts
```diff
@@ -115,19 +115,49 @@ import {UrlTree} from '../url_tree';
...
+   * @see [NavigationExtras](https://angular.io/api/router/NavigationExtras#navigationextras)
+   * @see [Router.createUrlTree](https://angular.io/api/router/Router#createurltree)
```
IgorMinar, line 121: Please don't use absolute links anywhere in angular.io as this will break navigation flow when a user is visiting a versioned deployment of the docs (e.g. v9.angular.io). In fact, all that you should need here is `Router#createUrlTree`.

See an existing example: https://github.com/angular/angular/blob/22d80a6780c235b2e8c64b23ebe7dd49c49f9aae/packages/core/src/interface/lifecycle_hooks.ts#L13-L18

packages/router/src/directives/router_link.ts
```diff
@@ -115,19 +115,49 @@ import {UrlTree} from '../url_tree';
...
+   */
```
IgorMinar, line 122: btw I'm not sure if you know, but if you add `aio: preview` label to this PR you get a comment posted on the PR by @mary-poppins - the comment then contains a link to the preview of the docs at each SHA in this PR.

You can then navigate to that URL and see what your change will look like on angular.io  without you having to build it locally.

Example link for this particular change: https://pr37018-01c5d4f.ngbuilds.io/api/router/RouterLink#properties

packages/router/src/directives/router_link.ts
```diff
@@ -228,15 +300,25 @@ export class RouterLinkWithHref implements OnChanges, OnDestroy {
...
+  set routerLink(commands: any[]|string|null|undefined) {
     if (commands != null) {
       this.commands = Array.isArray(commands) ? commands : [commands];
     } else {
       this.commands = [];
     }
   }
 
+  /**
+   * @deprecated 4.0.0 use `queryParamsHandling` instead.
```
IgorMinar, line 320:
```suggestion
   * @deprecated As of Angular v4.0 use `queryParamsHandling` instead.
```

IgorMinar (approved):
LGTM after the docs are updated. Thank you!

Reviewed-for: public-api

packages/router/src/directives/router_link.ts
atscott: I learned this after see you add the label to the PR. Very useful, thanks for pointing that out.

atscott:
merge assistance: I'm the owner of the router. LGTMs from Igor and Andrew Kushnir should cover required approvals.

PR #37027, opened 2020-05-09
docs(router): filter the event to subscribe

santoshyadavdev:
The current code will not work as the `e` will be an event, 
If we try to access e.id and e.url it will throw an exception, the correct way is to filter down to the event we want to subscribe.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/router/src/events.ts
+++ b/packages/router/src/events.ts
@@ -30,8 +30,8 @@ export type NavigationTrigger = 'imperative'|'popstate'|'hashchange';
  * class MyService {
  *   constructor(public router: Router, logger: Logger) {
  *     router.events.pipe(
- *       filter(e => e instanceof RouterEvent)
- *     ).subscribe(e => {
+ *        filter((e: Event): e is RouterEvent => e instanceof RouterEvent)
+ *     ).subscribe((e: RouterEvent) => {
  *       logger.log(e.id, e.url);
  *     });
  *   }


jnizet:
That fix won't make the code more correct. What's missing is a type assertion. Something like

```
    router.events.pipe(
      filter(e => e instanceof RouterEvent),
      map(e => e as RouterEvent)
    )
```

santoshyadav198613:
> That fix won't make the code more correct. What's missing is a type assertion. Something like
> 
> ```
>     router.events.pipe(
>       filter(e => e instanceof RouterEvent),
>       map(e => e as RouterEvent)
>     )
> ```

Yeah, thanks for pointing this out this will also work.

atscott:
How about `filter((e: Event): e is RouterEvent => e instanceof RouterEvent)` instead of adding `map`?

santoshyadav198613:
> How about `filter((e: Event): e is RouterEvent => e instanceof RouterEvent)` instead of adding `map`?

The code throws 
Property 'subscribe' does not exist on type 'MonoTypeOperatorFunction<Event>'

atscott:
@santoshyadav198613 - Did you write it like this?

image
https://stackblitz.com/edit/angular-ivy-ryq3mv

santoshyadav198613:
Ahh my bad, let me try it sometime.

santoshyadav198613:
Thanks, @atscott  it works changed and committed the code.

atscott (approved):
LGTM - thanks!

santoshyadav198613:
> LGTM - thanks!

Do I need to rebase, looks like some test is breaking.

atscott:
@santoshyadav198613 - I just triggered a rerun of the failed test in case it was a flake. If it fails again, you can try rebasing.

santoshyadav198613:
> @santoshyadav198613 - I just triggered a rerun of the failed test in case it was a flake. If it fails again, you can try rebasing.

Thanks :)

PR #22571, opened 2018-03-03
docs(aio): Reorganize style guide sections on component/directive custom prefixes

shaungrady:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

```
[ ] Bugfix
[ ] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[x] Documentation content changes
[ ] angular.io application / infrastructure changes
[ ] Other... Please describe:
```

## What is the current behavior?

Issue Number: Closes #22081

## What is the new behavior?

In the Style Guide...

- Move the `Custom prefix for directives` subsection to immediately follow the `Directive selectors` subsection, and rename it to `Directive custom prefix` to visually connect with the  `Directive selectors` subsection preceding it.

- Move `Component selector names` subsection from the `Components` section to the `Naming` section.

- Rename the `Custom prefix for components` to `Component custom prefix` to be consistent with `Directive custom prefix`.

- Rule numbers have not been changed.

## Does this PR introduce a breaking change?
```
[ ] Yes
[x] No
```

Issue #22081, opened 2018-02-08
docs: Style Guide's "Directive selectors" doesn't mention prefixing selectors


shaungrady:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Feature request
[x] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead see https://github.com/angular/angular/blob/master/CONTRIBUTING.md#question
</code></pre>

## Current behavior
In the [Directive selectors](https://angular.io/guide/styleguide#directive-selectors) section of the **Style Guide**, nothing is mentioned about prefixing selectors for name spacing.

_However_, in the **Guide** for [Write a structural directive](https://angular.io/guide/structural-directives#write-a-structural-directive), in the second to last paragraph, you can find the following:

> The directive _attribute name_ should be spelled in _lowerCamelCase_ and begin with a prefix. Don't use `ng`. That prefix belongs to Angular. Pick something short that fits you or your company. In this example, the prefix is `app`.

This is not mentioned in the style guide. Worse, Angular's directives seem to _only be prefixed_ with `ng` when it's a "single word" directive—`ngForm`—but they are not prefixed with they are multiple words—`formControlName`.

## Expected behavior

The Style Guide should make a recommendation about prefixing or not prefixing directive attribute names. I would be more than happy to submit a PR to clarify the issue in the Style Guide once I receive direction. 👍

kapunahelewong:
@shaungrady thanks for this (and your awesome tweet!). Let me know when you're ready and I'll help you through the contributing process if needed. So excited!

shaungrady:
I'm ready! So, the first question is, what's the recommendation the style guide should make? Should all directives be prefixed the same way that components are?

Yerkon:
`Should all directives be prefixed the same way that components are?`

Right, components/directives should be prefixed

Yerkon:
> In the Directive selectors section of the Style Guide, nothing is mentioned about prefixing selectors for name spacing.

https://angular.io/guide/styleguide#custom-prefix-for-components
https://angular.io/guide/styleguide#custom-prefix-for-directives

shaungrady:
[...]

PR #37033, opened 2020-05-10
docs: fix style guide order and the corresponding examples

### LINKED ISSUE #37030: Angular style guide documentation page is not showing sections in sequence.

### Description

In Angular style guide(https://angular.io/guide/styleguide) I found one problem the topics are not in sequence example after section 2-5 it shows section 5-2 and then 2-7 and then 2-6. 

when looking at a broader perspective both 5-2 and 2-7 seem to be talking about component selector name. then this different section makes me confuse especially when I am a new user.

https://angular.io/guide/styleguide

### Reproduction Steps**
open the style guide URL mentioned, go to section 2-5 (Bootstrapping) after this it shows section 05-02 instead of 2-6. 

when I open the style guide URL mentioned, go to section 2-5 (Bootstrapping) after this it should show section 2-6 instead of 5-2.  

## 📷Screenshot
Screenshot 2020-05-09 at 11 48 14 PM

## 🔥 Exception or Error

</code></pre>

### Anything else relevant?
If this is done intentionally would like to know the reason.

sonukapoor:
This commit fixes the number ordering of the style guide. It contained
missing sections.

Closes #37030

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: 37030

- [ ] Yes
- [X] No


sonukapoor:
@IgorMinar I tried to bring the style guide into the right sequence and order. I assume some of the sections were removed, which caused the sequence to break off at some point. To bring the guide into the correct order, it would make sense to fix the accompanying code examples for each section as well. I did both of these in the same PR, however, it could have also be broken down into multiple PRs/Commits. Let me know what you think and if you want to keep it in the current format or if you want to change it to multiple commits.

darpankumar:
@sonukapoor why this thing requires this much big change. multiple files are deleted

gkalpak (changes requested):
I am afraid this is not a desirable way to address #37030. Although I am not sure what it is, I suspect there is some reasoning behind naming the "rules" a certain way. Having sequential rules should not be a requirement (nor a goal).

More importantly, people (or even tolling such as `codelyzer`) might have links to these rules and renaming them like this will break things and cause great confusion.

##
If you want to fix #37030, you could try looking at the git history (e.g. via `git blame`) and try to figure out how/why `05-02` ended up between `02-05` and `02-07`.
- Was it always there? Was it moved there is a subsequent PR? Why?
- Did there used to be a `02-06` rule? Was it removed? Why?

Note that part of the git history might leave in the old (now archived) repo where angular.io source code lived before being moved into `angular/angular`: https://github.com/angular/angular.io

sonukapoor:
@gkalpak Thanks for the information. 

I have looked into the original changes using `git blame` and found out that it was changed about two years ago. The fix in the [PR](https://github.com/angular/angular/pull/22571) was to group the `Component Selectors` section into the [`Naming`](https://angular.io/guide/styleguide#naming) section. In the same PR, there was a discussion about the ordering and it is unfortunately not clear why that format was used ([here](https://github.com/angular/angular/issues/22081#issuecomment-367758634), and [here](https://github.com/angular/angular/issues/22081#issuecomment-368086474)).  You are also right about the usage in `Codelyzer` - this has been mentioned there as well. To be safe, it was decided to leave the existing ordering as is and add the new section with a new number.

gkalpak:
Thx for looking into it, @sonukapoor (and sorry you had to do all that work 😞).

sonukapoor:
> Thx for looking into it, @sonukapoor (and sorry you had to do all that work 😞).

No worries. Part of the learning process.

PR #37040, opened 2020-05-10
fix(ngcc): correctly get config for packages in nested `node_modules/`

gkalpak:
Previously, ngcc would only be able to match an ngcc configuration to packages that were located inside the project's top-level `node_modules/`. However, if there are multiple versions of a package in a project (e.g. as a transitive dependency of other packages), multiple copies of a package (at different versions) may exist in nested `node_modules/` directories. For example, one at `<project-root>/node_modules/some-package/` and one at `<project-root>/node_modules/other-package/node_modules/some-package/`. In such cases, ngcc was only able to detect the config for the first copy but not for the second.

This commit fixes this by returning a new instance of `ProcessedNgccPackageConfig` for each different package path (even if they refer to the same package name). In these `ProcessedNgccPackageConfig`, the `entryPoints` paths have been processed to take the package path into account.

### DIFF
--- a/packages/compiler-cli/ngcc/src/entry_point_finder/directory_walker_entry_point_finder.ts
+++ b/packages/compiler-cli/ngcc/src/entry_point_finder/directory_walker_entry_point_finder.ts
@@ -10,7 +10,7 @@ import {EntryPointWithDependencies} from '../dependencies/dependency_host';
 import {DependencyResolver, SortedEntryPointsInfo} from '../dependencies/dependency_resolver';
 import {Logger} from '../logging/logger';
 import {NgccConfiguration} from '../packages/configuration';
-import {getEntryPointInfo, INCOMPATIBLE_ENTRY_POINT, NO_ENTRY_POINT} from '../packages/entry_point';
+import {getEntryPointInfo, IGNORED_ENTRY_POINT, INCOMPATIBLE_ENTRY_POINT, isEntryPoint, NO_ENTRY_POINT} from '../packages/entry_point';
 import {EntryPointManifest} from '../packages/entry_point_manifest';
 import {PathMappings} from '../path_mappings';
 import {NGCC_DIRECTORY} from '../writing/new_entry_point_file_writer';
@@ -79,7 +79,9 @@ export class DirectoryWalkerEntryPointFinder implements EntryPointFinder {
 
     const entryPoints: EntryPointWithDependencies[] = [];
     if (primaryEntryPoint !== NO_ENTRY_POINT) {
-      entryPoints.push(this.resolver.getEntryPointWithDependencies(primaryEntryPoint));
+      if (primaryEntryPoint !== IGNORED_ENTRY_POINT) {
+        entryPoints.push(this.resolver.getEntryPointWithDependencies(primaryEntryPoint));
+      }
       this.collectSecondaryEntryPoints(
           entryPoints, sourceDirectory, sourceDirectory, this.fs.readdir(sourceDirectory));
 
@@ -148,14 +150,12 @@ export class DirectoryWalkerEntryPointFinder implements EntryPointFinder {
       }
 
       // If the path is a JS file then strip its extension and see if we can match an
-      // entry-point.
+      // entry-point (even if it is an ignored one).
       const possibleEntryPointPath = isDirectory ? absolutePath : stripJsExtension(absolutePath);
-      let isEntryPoint = false;
       const subEntryPoint =
           getEntryPointInfo(this.fs, this.config, this.logger, packagePath, possibleEntryPointPath);
-      if (subEntryPoint !== NO_ENTRY_POINT && subEntryPoint !== INCOMPATIBLE_ENTRY_POINT) {
+      if (isEntryPoint(subEntryPoint)) {
         entryPoints.push(this.resolver.getEntryPointWithDependencies(subEntryPoint));
-        isEntryPoint = true;
       }
 
       if (!isDirectory) {
@@ -163,9 +163,11 @@ export class DirectoryWalkerEntryPointFinder implements EntryPointFinder {
         continue;
       }
 
-      // This directory may contain entry-points of its own.
+      // If not an entry-point itself, this directory may contain entry-points of its own.
+      const canContainEntryPoints =
+          subEntryPoint === NO_ENTRY_POINT || subEntryPoint === INCOMPATIBLE_ENTRY_POINT;
       const childPaths = this.fs.readdir(absolutePath);
-      if (!isEntryPoint &&
+      if (canContainEntryPoints &&
           childPaths.some(
               childPath => childPath.endsWith('.js') &&
                   this.fs.stat(this.fs.resolve(absolutePath, childPath)).isFile())) {
--- a/packages/compiler-cli/ngcc/src/packages/configuration.ts
+++ b/packages/compiler-cli/ngcc/src/packages/configuration.ts
@@ -16,11 +16,11 @@ import {PackageJsonFormatPropertiesMap} from './entry_point';
 /**
  * The format of a project level configuration file.
  */
-export interface NgccProjectConfig<T = NgccPackageConfig> {
+export interface NgccProjectConfig<T = RawNgccPackageConfig> {
   /**
    * The packages that are configured by this project config.
    */
-  packages?: {[packagePath: string]: T};
+  packages?: {[packagePath: string]: T|undefined};
   /**
    * Options that control how locking the process is handled.
    */
@@ -44,17 +44,16 @@ export interface ProcessLockingConfiguration {
 }
 
 /**
- * The format of a package level configuration file.
+ * The raw format of a package level configuration (as it appears in configuration files).
  */
-export interface NgccPackageConfig {
+export interface RawNgccPackageConfig {
   /**
    * The entry-points to configure for this package.
    *
-   * In the config file the keys can be paths relative to the package path;
-   * but when being read back from the `NgccConfiguration` service, these paths
-   * will be absolute.
+   * In the config file the keys are paths relative to the package path.
    */
-  entryPoints: {[entryPointPath: string]: NgccEntryPointConfig;};
+  entryPoints?: {[entryPointPath: string]: NgccEntryPointConfig};
+
   /**
    * A collection of regexes that match deep imports to ignore, for this package, rather than
    * displaying a warning.
@@ -71,6 +70,7 @@ export interface NgccPackageConfig {
 export interface NgccEntryPointConfig {
   /** Do not process (or even acknowledge the existence of) this entry-point, if true. */
   ignore?: boolean;
+
   /**
    * This property, if provided, holds values that will override equivalent properties in an
    * entry-point's package.json file.
@@ -91,6 +91,12 @@ export interface NgccEntryPointConfig {
    */
   generateDeepReexports?: boolean;
 }
+
+interface VersionedPackageConfig extends RawNgccPackageConfig {
+  versionRange: string;
+}
+
+type PartiallyProcessedConfig = Required<NgccProjectConfig<VersionedPackageConfig[]>>;
 
 /**
  * The default configuration for ngcc.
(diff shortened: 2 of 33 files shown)


devversion (approved):
LGTM for dev-infra.

packages/compiler-cli/ngcc/src/packages/configuration.ts
```diff
@@ -44,17 +44,16 @@ export interface ProcessLockingConfiguration {
...
   /**
    * The entry-points to configure for this package.
    *
-   * In the config file the keys can be paths relative to the package path;
-   * but when being read back from the `NgccConfiguration` service, these paths
-   * will be absolute.
+   * In the config file the keys can be paths relative to the package path.
    */
-  entryPoints: {[entryPointPath: string]: NgccEntryPointConfig;};
+  entryPoints?: {[entryPointPath: string]: NgccEntryPointConfig|undefined};
```
petebacondarwin, line 55:
```suggestion
  entryPoints?: {[entryPointPath: string]: NgccEntryPointConfig};
```

packages/compiler-cli/ngcc/src/packages/configuration.ts
```diff
@@ -152,13 +158,48 @@ export const DEFAULT_NGCC_CONFIG: NgccProjectConfig = {
...
+  ignorableDeepImportMatchers: RegExp[];
+
+  constructor(packagePath: AbsoluteFsPath, {
+    entryPoints = {},
+    ignorableDeepImportMatchers = [],
+  }: RawNgccPackageConfig) {
+    const absolutePathEntries: [AbsoluteFsPath, NgccEntryPointConfig][] =
+        Object.entries(entryPoints).map(([
+                                          relativePath, config
+                                        ]) => [resolve(packagePath, relativePath), config!]);
```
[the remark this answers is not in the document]
petebacondarwin: If this is the case, then it is better to remove the `undefined` from the definition of:

```
entryPoints?: {[entryPointPath: string]: NgccEntryPointConfig|undefined};
```

petebacondarwin (changes requested):
Can you remove the `undefined` from the `RawNgccPackageConfig.exports` declaration.
Otherwise LGTM (with a nit or two).

packages/compiler-cli/ngcc/src/packages/configuration.ts
gkalpak: The reason I added `undefined` is that it will better reflect the fact that `.entryPoints.someEntryPoint` might or might not be defined. Previously, TS would infer it as `NgccEntryPointConfig` although that was not necessarily the case.

This was more of an issue when this type was returned by `getPackageConfig()`, so happy to remove it now that `RawNgccPackageConfig` is only used internally.

packages/compiler-cli/ngcc/src/packages/entry_point.ts
```diff
@@ -118,26 +118,28 @@ export type GetEntryPointResult = EntryPoint|typeof INCOMPATIBLE_ENTRY_POINT|typ
 export function getEntryPointInfo(
     fs: FileSystem, config: NgccConfiguration, logger: Logger, packagePath: AbsoluteFsPath,
     entryPointPath: AbsoluteFsPath): GetEntryPointResult {
+  const packageName = packagePath.replace(/^.*\/node_modules\//i, '');
```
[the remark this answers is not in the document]
gkalpak: I thought about it too, but I felt it would require `getPackageConfig()` to assume too much. It is an implementation detail that `getEntryPointInfo()` only considers packages inside `node_modules/` (which might change in the future) and I think it is the caller's responsibility to determine the package name.

The dependency resolver, for example, does it differently: https://github.com/angular/angular/blob/ac92b5ac6c79f0a7cc2900b48bf8f6a3f5a85aaf/packages/compiler-cli/ngcc/src/dependencies/dependency_resolver.ts#L228-L229

WDYT?

packages/compiler-cli/ngcc/src/packages/configuration.ts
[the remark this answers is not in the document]
gkalpak: The main purpose of `undefined` was to guard against the fact that accessing arbitrary properties on `entryPoints` is not guaranteed to return an `NgccEntryPointConfig` (see https://github.com/angular/angular/pull/37040#discussion_r424670933). Here we are accessing properties that we know exist on `entryPoints`, so it is safe(r) to assume they are not `undefined`. (Not that it is not possible for people to set `entryPoints.foo` to undefined in their configs, but addressing that would be outside the scope of this PR :grin:)

Removing the `undefined` as mentioned in https://github.com/angular/angular/pull/37040#discussion_r424670933...

gkalpak:
Addressed feedback (or way or another 😛). @petebacondarwin, ptal.

gkalpak:
This turned out to be a little more tricky than I expected. I finally have something I am happy with (I think 😅).

@petebacondarwin, ptal.

packages/compiler-cli/ngcc/src/packages/entry_point.ts
```diff
@@ -294,16 +301,41 @@ function guessTypingsFromPackageJson(
...
+    const lastSegment = fs.basename(packagePath);
+    const secondLastSegment = fs.basename(fs.dirname(packagePath));
+
+    packageName =
+        secondLastSegment.startsWith('@') ? `${secondLastSegment}/${lastSegment}` : lastSegment;
+  }
+
+  return {
+    packageName,
+    packageVersion: packageJson?.version ?? null,
```
petebacondarwin, line 340: If the entry-point package.json does not contain a `version` property, which is possible, then we could just load up the `package.json` of the main package (found at `packagePath`)?

packages/compiler-cli/ngcc/test/packages/entry_point_spec.ts
```diff
@@ -166,6 +169,136 @@ runInEachFileSystem(() => {
...
+                    override: {typings: './index.d.ts'},
+                  },
+                },
+              });
+            }
+
+            const entryPoint = getEntryPointInfo(fs, config, logger, packagePath, entryPointPath);
+
+            // Ensure we did not get `NO_ENTRY_POINT` or `INCOMPATIBLE_ENTRY_POINT`.
+            expect(entryPoint).toBeInstanceOf(Object);
```
petebacondarwin, line 213: Another opportunity to use the `isEntryPoint()` helper suggested earlier.

packages/compiler-cli/ngcc/test/packages/entry_point_spec.ts
```diff
@@ -166,6 +169,136 @@ runInEachFileSystem(() => {
...
+                },
+              ]);
+            } else {
+              // Ensure the entry-point `package.json` does not exist (e.g. from a previous test).
+              if (fs.exists(entryPointPackageJsonPath)) {
+                fs.removeFile(entryPointPackageJsonPath);
+              }
+
+              // Ensure there is an ngcc config for the entry-point.
+              getPackageConfigSpy.withArgs(packagePath, null).and.returnValue({
```
petebacondarwin: Could we avoid a spy and just generate an `ngcc.config.js` file here?

packages/compiler-cli/ngcc/test/packages/entry_point_spec.ts
```diff
@@ -166,6 +169,136 @@ runInEachFileSystem(() => {
...
+            if (entryPointName !== undefined) {
+              // Ensure a `package.json` exists for the entry-point (containing `entryPointName`).
+              loadTestFiles([
+                {
+                  name: entryPointPackageJsonPath,
+                  contents: JSON.stringify({name: entryPointName, typings: './index.d.ts'}),
+                },
+              ]);
+            } else {
+              // Ensure the entry-point `package.json` does not exist (e.g. from a previous test).
```
petebacondarwin, line 192: Do you mean from a previous `expect()`? Since each `it()` clause should get its own new `MockFileSystem`.
Perhaps it would be safer to move each of these `expect()` calls into their own `it()` clause if we are concerned about bleeding data between them?

petebacondarwin: This would also allow us to have a description for each of them, which would be good documentation of what the tests are actually doing.

petebacondarwin (changes requested):
I am not sure how it passes integration tests but I am pretty sure we cannot rely upon secondary entry-points containing a `version` property.

packages/compiler-cli/ngcc/src/entry_point_finder/directory_walker_entry_point_finder.ts
```diff
@@ -148,14 +150,15 @@ export class DirectoryWalkerEntryPointFinder implements EntryPointFinder {
...
+      // entry-point (even if it is an ignored one).
       const possibleEntryPointPath = isDirectory ? absolutePath : stripJsExtension(absolutePath);
-      let isEntryPoint = false;
       const subEntryPoint =
           getEntryPointInfo(this.fs, this.config, this.logger, packagePath, possibleEntryPointPath);
-      if (subEntryPoint !== NO_ENTRY_POINT && subEntryPoint !== INCOMPATIBLE_ENTRY_POINT) {
+      const isEntryPoint =
+          subEntryPoint !== NO_ENTRY_POINT && subEntryPoint !== INCOMPATIBLE_ENTRY_POINT;
+      if (subEntryPoint !== NO_ENTRY_POINT && subEntryPoint !== INCOMPATIBLE_ENTRY_POINT &&
+          subEntryPoint !== IGNORED_ENTRY_POINT) {
```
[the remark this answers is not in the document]
gkalpak: It felt less readable. How about re-using `isEntryPoint`:

```ts
const isEntryPoint =
    subEntryPoint !== NO_ENTRY_POINT && subEntryPoint !== INCOMPATIBLE_ENTRY_POINT;
if (isEntryPoint && subEntryPoint !== IGNORED_ENTRY_POINT) {
```

packages/compiler-cli/ngcc/test/packages/entry_point_spec.ts
gkalpak: Yes, from a previous `expect()`. I thought about splitting up into different `it()` calls, but I felt it wouldn't make the tests easier to follow/read.

gkalpak: Happy to split it up if you think it would be better.

packages/compiler-cli/ngcc/test/packages/entry_point_spec.ts
gkalpak: I guess we could, but what would be the benefit?

packages/compiler-cli/ngcc/src/packages/entry_point.ts
```diff
@@ -294,16 +301,41 @@ function guessTypingsFromPackageJson(
...
+ * Find or infer the name and version of a package.
  *
- * The version is read off of the `version` property of `packageJson`. It is assumed that even if
- * `packageJson` corresponds to a secondary entry-point (e.g. `@angular/common/http`) it will still
- * contain the same version as the main package (e.g. `@angular/common`).
+ * - The name is computed based on the `name` property of `packageJson` (if available) or inferred
+ *   from the package's path.
+ * - The version is read off of the `version` property of `packageJson`. It is assumed that even if
+ *   `packageJson` corresponds to a secondary entry-point (e.g. `@angular/common/http`) it will
+ *   still contain the same version as the main package (e.g. `@angular/common`).
```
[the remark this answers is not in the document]
gkalpak: 😱

packages/compiler-cli/ngcc/src/packages/entry_point.ts
```diff
@@ -299,16 +296,14 @@ function guessTypingsFromPackageJson(
 /**
  * Find the version of the package at `packageJsonPath`.
  *
- * @returns the version string or `null` if the package.json does not exist or is invalid.
+ * The version is read off of the `version` property of `packageJson`. It is assumed that even if
+ * `packageJson` corresponds to a secondary entry-point (e.g. `@angular/common/http`) it will still
+ * contain the same version as the main package (e.g. `@angular/common`).
```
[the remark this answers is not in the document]
gkalpak: 😱

packages/compiler-cli/ngcc/src/packages/entry_point.ts
```diff
@@ -124,31 +125,32 @@ export function getEntryPointInfo(
...
   const entryPointPackageJson = hasConfig ?
       mergeConfigAndPackageJson(
           loadedEntryPointPackageJson, entryPointConfig, packagePath, entryPointPath) :
-      loadedEntryPointPackageJson;
-
-  if (entryPointPackageJson === null) {
-    // package.json exists but could not be parsed and there was no redeeming config
-    return INCOMPATIBLE_ENTRY_POINT;
-  }
+      loadedEntryPointPackageJson!;
```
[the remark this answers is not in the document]
gkalpak: Agreed this is not great. I have moved things around to get rid of the non-null assertions.

gkalpak:
@petebacondarwin, I have addressed your comments in fixup commits (all fixup commits are new) plus this new regular commit:
- **fix(ngcc): correctly retrieve a package's version from it's `package.json`** (https://github.com/angular/angular/pull/37040/commits/8d353e36fd0db65bf8830a8418291c46afffa8ae)

petebacondarwin (approved):
LGTM - Just the one question about the order of checking for entry-point version that I think is worth considering.

petebacondarwin (approved):
LGTM (again)

JoostK (approved):
Nice cleanup overall!

PR #37049, opened 2020-05-11
feat(dev-infra): introduce validators for ng-dev config loading

josephperrott:
Introduces infrastructure to validate configuration of the ng-dev
command at run time.  Allowing for errors to be returned to the
user running the command.

### DIFF
--- a/dev-infra/commit-message/config.ts
+++ b/dev-infra/commit-message/config.ts
@@ -6,7 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {NgDevConfig} from '../utils/config';
+import {getConfig, NgDevConfig, processErrors} from '../utils/config';
 
 export interface CommitMessageConfig {
   maxLineLength: number;
@@ -15,11 +15,17 @@ export interface CommitMessageConfig {
   scopes: string[];
 }
 
-/** Validate the configuration correctly provides commitMessage information. */
-export function isCommitMessageConfig(
-    config: any, errors: string[]): config is NgDevConfig<{commitMessage: CommitMessageConfig}> {
+/** Retrieve and validate the config as FormatConfig. */
+export function getCommitMessageConfig() {
+  // List of errors encountered validating the config.
+  const errors: string[] = [];
+  // The unvalidated config object.
+  const config: Partial<NgDevConfig<{commitMessage: CommitMessageConfig}>> = getConfig();
+
   if (config.commitMessage === undefined) {
     errors.push(`No configuration defined for "commitMessage"`);
   }
-  return true;
+
+  processErrors(errors);
+  return config as Required<typeof config>;
 }
--- a/dev-infra/commit-message/config.ts
+++ b/dev-infra/commit-message/config.ts
@@ -25,5 +25,5 @@ export function isCommitMessageConfig(
   if (config.commitMessage === undefined) {
     errors.push(`No configuration defined for "commitMessage"`);
   }
-  return;
+  return true;
 }
--- a/dev-infra/commit-message/config.ts
+++ b/dev-infra/commit-message/config.ts
@@ -6,7 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {getConfig, NgDevConfig, processErrors} from '../utils/config';
+import {assertNoErrors, getConfig, NgDevConfig} from '../utils/config';
 
 export interface CommitMessageConfig {
   maxLineLength: number;
@@ -15,7 +15,7 @@ export interface CommitMessageConfig {
   scopes: string[];
 }
 
-/** Retrieve and validate the config as FormatConfig. */
+/** Retrieve and validate the config as `CommitMessageConfig`. */
 export function getCommitMessageConfig() {
   // List of errors encountered validating the config.
   const errors: string[] = [];
@@ -26,6 +26,6 @@ export function getCommitMessageConfig() {
     errors.push(`No configuration defined for "commitMessage"`);
   }
 
-  processErrors(errors);
+  assertNoErrors(errors);
   return config as Required<typeof config>;
 }
--- a/dev-infra/commit-message/config.ts
+++ b/dev-infra/commit-message/config.ts
@@ -16,12 +16,14 @@ export interface CommitMessageConfig {
 }
 
 export const COMMIT_MESSAGE = {
-  configKey: 'commitMessage',
-  validator,
+  validator: isCommitMessageConfig,
 };
 
 /** Validate the configuration correctly provides commitMessage information. */
-export function validator(
+export function isCommitMessageConfig(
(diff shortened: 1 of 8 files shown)


dev-infra/commit-message/config.ts
```diff
@@ -5,9 +5,23 @@
...
 }
+
+export const COMMIT_MESSAGE = {
+  configKey: 'commitMessage',
+  validator,
+};
+
+/** Validate the configuration correctly provides commitMessage information. */
+export function validator(
+    config: any, errors: string[]): config is NgDevConfig<{commitMessage: CommitMessageConfig}> {
```
devversion, line 25: I'd prefer getting rid of the `configKey` logic as it rather seems to complicate the `getConfig` logic.

Ideally we'd validate the config just here since that matches what the type constraint suggests here. I know it's slightly more overhead, but will also give us more flexibility in what error we print if not configured.

dev-infra/format/config.ts
```diff
@@ -6,8 +6,43 @@
...
+  [keyof: string]: boolean|Formatter;
+}
+
+export const FORMAT = {
+  configKey: 'format',
+  validator,
+};
+
+/** Validate the configuration correctly provides format information. */
+export function validator(config: any, errors: string[]): config is NgDevConfig<{format: Format}> {
```
devversion, line 25: If we go with what I proposed above (i.e. getting rid of `configKey`), then I think we should consistently call these methods `validateConfig` or `isFormatConfig`.

dev-infra/utils/config.ts
```diff
@@ -9,13 +9,103 @@
 import {join} from 'path';
 import {exec} from 'shelljs';
 
+/**
+ * The common configuration for ng-dev.  This type signature is included whenever
```
devversion, line 13: There seems to be an additional whitespace here (unless intentional?). Nit: I think it might not be needed to mention specifics of type signature here. Might be better to have that where the intersection happens?

dev-infra/format/config.ts
```diff
@@ -6,8 +6,43 @@
...
+  const formatConfig: Partial<Format> = config.format;
+  const localErrors: string[] = [];
```
devversion, line 27: Why do we need another errors array here? We could just push to the `errors` array I think. 

Related to that, the return value of the type constraint is noop anyway. Maybe that means we should do something else?

```ts
export const formatConfigValidator: ConfigValidator<FormatConfig> = (config, errors) => {
  // no return needed
};
```

dev-infra/utils/config.ts
```diff
@@ -9,13 +9,103 @@
...
+}
+
+/**
+ * Load the configuration from the file system, returning the already loaded copy if it
+ * is defined.
+ */
+function loadConfig() {
+  // If the global config is defined, return it rather than reloading from file system.
+  if (CONFIG) {
+    return {...CONFIG};
```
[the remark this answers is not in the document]
josephperrott: My thinking was that, it should always return the config object that was loaded from the config file.  This way if something were to modify the config file, but then request it again, it would be "fresh"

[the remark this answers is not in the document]
devversion: I was asking because we were not cloning it when we return it from the initial read. We should add a comment on this and fix it for the initial load then?

dev-infra/format/config.ts
josephperrott: Removed the `localErrors`  array.

I think its worth continuing to require that the `validator` function provided meets:
`(config: any, errors?: string[]) => config is T`

I think its worth continuing to rely on the type narrowing because it makes it obvious what the function is providing.

devversion (approved):
Generally LGTM. Just a few comments.

dev-infra/utils/config.ts
```diff
@@ -9,13 +9,90 @@
...
+ * the configuration.
+ */
+export type ConfigValidator<T> = (config: any, errors: string[]) => config is T;
+
 // The filename expected for creating the ng-dev config, without the file
 // extension to allow either a typescript or javascript file to be used.
 const CONFIG_FILE_NAME = '.ng-dev-config';
 
+/** The configuration for ng-dev. */
+let CONFIG: unknown|null = null;
```
gkalpak, line 34: OOC, why not `{}|null`?

gkalpak (approved):

dev-infra/utils/config.ts
devversion: I initially proposed `unknown` because it could be anything that the config returns, but given that we cast this later down the road to `{}`, we could probably avoid these casts by using `{}|null` here. Ideally we'd validate that the loaded config is an object first though.

dev-infra/format/config.ts
```diff
@@ -6,8 +6,46 @@
...
+  if (config.format === undefined) {
+    errors.push(`No configuration defined for "format"`);
+  }
+
+  for (const [key, value] of Object.entries(config.format!)) {
+    switch (typeof value) {
+      case 'boolean':
+        break;
+      case 'object':
+        isFormatterConfig(key, value, errors);
```
devversion, line 35: Optional nit: might want to rename that function to `checkFormatterConfig`.

dev-infra/commit-message/config.ts
```diff
@@ -5,9 +5,27 @@
...
+import {getConfig, NgDevConfig, processErrors} from '../utils/config';
+
 export interface CommitMessageConfig {
   maxLineLength: number;
   minBodyLength: number;
   types: string[];
   scopes: string[];
 }
+
+/** Retrieve and validate the config as FormatConfig. */
```
devversion, line 18:
```suggestion
/** Retrieve and validate the config as `CommitMessageConfig`. */
```

dev-infra/utils/config.ts
```diff
@@ -9,13 +9,67 @@
...
+ * The configuration for the specific ng-dev command, providing both the common
+ * ng-dev config as well as the specific config of a subcommand.
+ */
+export type NgDevConfig<T = {}> = CommonConfig&T;
+
+/**
+ * Configuration loader object, providing information needed to load and validate
+ * the configuration.
+ */
+export type ConfigValidator<T> = (config: any, errors: string[]) => config is T;
```
devversion, line 27: This is unused now, right?

dev-infra/utils/config.ts
```diff
@@ -9,13 +9,67 @@
...
+}
+
+/** Validate the common configuration has been met for the ng-dev command. */
+function validateCommonConfig(config: NgDevConfig<CommonConfig>) {
+  // TODO: add validation for the common configuration
+
+  return config;
+}
+
+export function processErrors(errors: string[]) {
```
devversion, line 61: Add a method description? Might also consider making it more obvious that this function will exit the process if the are errors. How about `assertNoErrors()` (or something similar).

devversion (approved):

PR #37051, opened 2020-05-11
feat(dev-infra): create tool to determine conflicts created by a PR

josephperrott:
Creates a tool in ng-dev to determine the PRs which become conflicted
by merging a specified PR.  Often the question is brought up of how
many PRs require a rebase as a result of a change.  This script allows
to determine this impact.

### DIFF
--- a/dev-infra/pr/discover-new-conflicts.ts
+++ b/dev-infra/pr/discover-new-conflicts.ts
@@ -1,0 +1,158 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {Bar} from 'cli-progress';
+import {types as graphQLTypes} from 'typed-graphqlify';
+
+import {getConfig, NgDevConfig} from '../utils/config';
+import {getCurrentBranch, hasLocalChanges} from '../utils/git';
+import {getPendingPrs} from '../utils/github';
+import {exec} from '../utils/shelljs';
+
+
+/* GraphQL schema for the response body for each pending PR. */
+const PR_SCHEMA = {
+  headRef: {
+    name: graphQLTypes.string,
+    repository: {
+      url: graphQLTypes.string,
+      nameWithOwner: graphQLTypes.string,
+    },
+  },
+  baseRef: {
+    name: graphQLTypes.string,
+    repository: {
+      url: graphQLTypes.string,
+      nameWithOwner: graphQLTypes.string,
+    },
+  },
+  updatedAt: graphQLTypes.string,
+  number: graphQLTypes.number,
+  mergeable: graphQLTypes.string,
+  title: graphQLTypes.string,
+};
+
+/* Pull Request response from Github GraphQL query */
+type RawPullRequest = typeof PR_SCHEMA;
+
+/** Convert raw Pull Request response from Github to usable Pull Request object. */
+function processPr(pr: RawPullRequest) {
+  return {...pr, updatedAt: (new Date(pr.updatedAt)).getTime()};
+}
+
+/* Pull Request object after processing, derived from the return type of the processPr function. */
+type PullRequest = ReturnType<typeof processPr>;
+
+/** Name of a temporary local branch that is used for checking conflicts. **/
+const tempWorkingBranch = '__NgDevRepoBaseAfterChange__';
+
+/** Checks if the provided PR will cause new conflicts in other pending PRs. */
+export async function discoverNewConflictsForPr(
+    newPrNumber: number, updatedAfter: number, config: Pick<NgDevConfig, 'github'> = getConfig()) {
+  // If there are any local changes in the current repository state, the
+  // check cannot run as it needs to move between branches.
+  if (hasLocalChanges()) {
+    console.error('Cannot run with local changes. Please make sure there are no local changes.');
+    process.exit(1);
+  }
+
+  /** The active github branch when the run began. */
+  const originalBranch = getCurrentBranch();
+  /* Progress bar to indicate progress. */
+  const progressBar = new Bar({format: `[{bar}] ETA: {eta}s | {value}/{total}`});
+  /* PRs which were found to be conflicting. */
+  const conflicts: Array<PullRequest> = [];
+  /* String version of the updatedAfter value, for logging. */
+  const updatedAfterString = new Date(updatedAfter).toLocaleDateString();
+
+  console.info(`Requesting pending PRs from Github`);
+  /** List of PRs from github currently known as mergable. */
+  const allPendingPRs = (await getPendingPrs(PR_SCHEMA, config.github)).map(processPr);
+  /** The PR which is being checked against. */
+  const requestedPr = allPendingPRs.find(pr => pr.number === newPrNumber);
+  if (requestedPr === undefined) {
+    console.error(
+        `The request PR, #${newPrNumber} was not found as a pending PR on github, please confirm`);
+    console.error(`the PR number is correct and is an open PR`);
+    process.exit(1);
+  }
+
+  const pendingPrs = allPendingPRs.filter(pr => {
+    return (
+        // PRs being merged into the same target branch as the requested PR
+        pr.baseRef.name === requestedPr.baseRef.name &&
+        // PRs which either have not been processed or are determined as mergable by Github
+        pr.mergeable !== 'CONFLICTING' &&
+        // PRs updated after the provided date
+        pr.updatedAt >= updatedAfter);
+  });
+  console.info(`Retrieved ${allPendingPRs.length} total pending PRs`);
+  console.info(`Checking ${pendingPrs.length} PRs for conflicts after a merge of #${newPrNumber}`);
+
+  // Fetch and checkout the PR being checked.
+  exec(`git fetch ${requestedPr.headRef.repository.url} ${requestedPr.headRef.name}`);
+  exec(`git checkout -B ${tempWorkingBranch} FETCH_HEAD`);
+
+  // Rebase the PR against the PRs target branch.
+  exec(`git fetch ${requestedPr.baseRef.repository.url} ${requestedPr.baseRef.name}`);
+  const result = exec(`git rebase FETCH_HEAD`);
+  if (result.code) {
+    console.error('The requested PR currently has conflicts');
+    cleanUpGitState(originalBranch);
+    process.exit(1);
+  }
+
+  // Start the progress bar
+  progressBar.start(pendingPrs.length, 0);
+
+  // Check each PR to determine if it can merge cleanly into the repo after the target PR.
+  for (const pr of pendingPrs) {
+    // Fetch and checkout the next PR
+    exec(`git fetch ${pr.headRef.repository.url} ${pr.headRef.name}`);
+    exec(`git checkout --detach FETCH_HEAD`);
+    // Check if the PR cleanly rebases into the repo after the target PR.
+    const result = exec(`git rebase ${tempWorkingBranch}`);
+    if (result.code !== 0) {
+      conflicts.push(pr);
+    }
+    // Abort any outstanding rebase attempt.
+    exec(`git rebase --abort`);
+
+    progressBar.increment(1);
+  }
+  // End the progress bar as all PRs have been processed.
+  progressBar.stop();
+  console.info(`\nResult:`);
+
+  cleanUpGitState(originalBranch);
+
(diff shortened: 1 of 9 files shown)


sonukapoor (commented):
Are there any unit tests for these additions?

dev-infra/utils/github.ts
```diff
@@ -0,0 +1,88 @@
...
+/** Get all pending PRs from github  */
+export async function getPendingPrs<RawPullRequest, PullRequest>(
+    prNode: RawPullRequest, processor: (pr: RawPullRequest) => PullRequest) {
+  // The GraphQL query object to get a page of pending PRs
+  const PRS_QUERY = params(
+      {
+        $first: 'Int',      // How many entries to get with each request
+        $after: 'String',   // The cursor to start the page at
+        $owner: 'String!',  // The organization to query for
+        $name: 'String!',   // The organization to query for
```
gkalpak, line 34:
```suggestion
        $name: 'String!',   // The organization to query forDid you mean "repository" or something?
```

dev-infra/tmpl-package.json
```diff
@@ -9,13 +9,16 @@
     "ts-circular-deps": "./ts-circular-dependencies/index.js"
   },
   "dependencies": {
+    "@bazel/buildifier": "<from-root>",
```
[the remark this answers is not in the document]
josephperrott: Bad rebase artifact, its now an item in `peerDependencies`

dev-infra/pr/discover-new-conflicts.ts
```diff
@@ -0,0 +1,164 @@
...
+  const result = exec(`git rebase upstream/master`);
+  if (result.code) {
+    exitGracefullyWithError('the base PR being checked doesnt work', originalBranch);
+  }
+
+  console.info(`Requesting pending PRs from Github.`);
+  // Retrieve all of the pending PRs which are currently know to be mergable.
+  const allPendingPRs = (await getPendingPrs(PR_SCHEMA, processPr));
+  const mergablePrs = allPendingPRs.filter(pr => pr.mergeable === 'MERGEABLE');
+  const pendingPrs = mergablePrs.filter(pr => pr.updatedAt >= updatedAfter);
```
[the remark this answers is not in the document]
josephperrott: Unfortunately, we cannot add them into the request because the GraphQL API doesn't allow filtering by these values in the request.

dev-infra/pr/discover-new-conflicts.ts
```diff
@@ -0,0 +1,164 @@
...
+    // Abort any outstanding rebase attempt.
+    exec(`git rebase --abort`);
+
+    progressBar.increment(1);
+  }
+  // End the progress bar as all PRs have been processed.
+  progressBar.stop();
+  console.info(`\nResult:`);
+
+  resetGitEnvironment(originalBranch);
```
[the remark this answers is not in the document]
josephperrott: I removed `resetGitEnvironment`, but I we actually need `progresBar.stop()` to be here because we can't write to `stdout` while the `progressBar` is active, as it will rewrite the line.

dev-infra/utils/github.ts
```diff
@@ -0,0 +1,88 @@
...
+   * Gets the query and queryParams for a specific page of entries.
+   */
+  const queryBuilder = (count: number, cursor?: string) => {
+    return {
+      query,
+      params: {
+        after: cursor || null,
+        first: count,
+        owner: 'angular',
+        name: 'angular',
```
devversion, line 65: This should be based on the dev-infra config right? Can you add a TODO since we don't have the base config yet.

dev-infra/utils/github.ts
```diff
@@ -0,0 +1,88 @@
...
+  headers: {
+    // TODO(josephperrott): Remove reference to TOKEN environment variable as part of larger
+    // effort to migrate to expecting tokens via GITHUB_ACCESS_TOKEN environment variables.
+    authorization: `token ${process.env.TOKEN || process.env.GITHUB_ACCESS_TOKEN}`,
+  }
+});
+
+/** Get all pending PRs from github  */
+export async function getPendingPrs<RawPullRequest, PullRequest>(
+    prNode: RawPullRequest, processor: (pr: RawPullRequest) => PullRequest) {
```
devversion, line 27: Should the transformation be just happening later? I think this function should be self-contained and should just accept the pull request schema.

`prNode` could probably be named better by mentioning `schema` for example. Also might want to document the schema parameter in the JSdoc.

dev-infra/pr/discover-new-conflicts.ts
```diff
@@ -0,0 +1,163 @@
...
+}
+
+// Pull Request object after processing, derived from the
+// return type of the processPr function.
+type PullRequest = ReturnType<typeof processPr>;
+
+// Progress bar to indicate progress.
+const progressBar = new Bar({format: `[{bar}] ETA: {eta}s | {value}/{total}`});
+
+// Branch name used
```
devversion, line 46:
```suggestion
/** Name of a temporary local branch that is used for checking conflicts. **/
```

dev-infra/pr/discover-new-conflicts.ts
```diff
@@ -0,0 +1,163 @@
...
+function processPr(pr: RawPullRequest) {
+  return {...pr, updatedAt: (new Date(pr.updatedAt)).getTime()};
+}
+
+// Pull Request object after processing, derived from the
+// return type of the processPr function.
+type PullRequest = ReturnType<typeof processPr>;
+
+// Progress bar to indicate progress.
+const progressBar = new Bar({format: `[{bar}] ETA: {eta}s | {value}/{total}`});
```
devversion, line 44: Should we initialize these things only when needed? Currently, all of this runs when any arbitrary `ng-dev` command runs (even though it might not do much; it's unnecessary overhead).

I'm generally advocating for having files without top-level expressions that run (actual literal constants are fine)

dev-infra/pr/discover-new-conflicts.ts
```diff
@@ -0,0 +1,163 @@
...
+const newBaseRef = 'ng-dev-repo-base-after-pr';
+
+/**
+ * Checks if the provided PR will cause new conflicts in other pending PRs.
+ */
+export async function discoverNewConflictsForPr(newPrNumber: number, updatedAfter: number) {
+  // If there are any local changes in the current repository state, the
+  // check cannot run as it needs to move between branches.
+  if (hasLocalChanges()) {
+    console.error('Cant run with local changes');
```
devversion, line 56:
```suggestion
    console.error('Cannot run with local changes. Please make sure there are no local changes.');
```

dev-infra/pr/discover-new-conflicts.ts
```diff
@@ -0,0 +1,163 @@
...
+  // The branch checked out when the run started
+  const originalBranch = getCurrentBranch();
+  // PRs which were found to be conflicting.
+  const conflicts: Array<PullRequest> = [];
+  // String version of the updatedAfter value, for logging.
+  const updatedAfterString = new Date(updatedAfter).toLocaleDateString();
+
+  // Fetch and checkout the PR being checked.
+  exec(`git fetch upstream pull/${newPrNumber}/head`);
+  exec(`git checkout -b ${newBaseRef} FETCH_HEAD`);
```
devversion, line 69: Do you want to use `checkout -B` so that it will proceed if the temporary branch has not been cleaned up properly?  (which could happen commonly when the script is exited early, right?)

devversion: Also, did you consider using the actual remote URL instead of assuming `upstream`. That's what we do in the current merge scripts too? Add a TODO when we have the common config?

dev-infra/pr/discover-new-conflicts.ts
```diff
@@ -0,0 +1,163 @@
...
+    exitGracefullyWithError('The requested PR currently has conflicts', originalBranch);
+  }
+
+  console.info(`Requesting pending PRs from Github`);
+  // Retrieve all of the pending PRs which are currently known to be mergable.
+  const allPendingPRs = (await getPendingPrs(PR_SCHEMA, processPr));
+  const mergablePrs = allPendingPRs.filter(pr => pr.mergeable === 'MERGEABLE');
+  const pendingPrs = mergablePrs.filter(pr => pr.updatedAt >= updatedAfter);
+  console.info(`Retrieved ${allPendingPRs.length} pending PRs`);
+  console.info(`${mergablePrs.length} of these PRs have no known conflicts`);
```
devversion, line 82: No known conflicts to what? These are just known to not conflict with their `base` branch, but not with the checked PR, right? Should we clarify this? Don't we need to check `mergeablePrs` too?

dev-infra/pr/discover-new-conflicts.ts
```diff
@@ -0,0 +1,163 @@
...
+ */
+function exitGracefully(code: number, originalBranch: string) {
+  progressBar.stop();
+  resetGitEnvironment(originalBranch);
+  process.exit(code);
+}
+/**
+ * Exit the process by throwing an error after cleaning the git environment.
+ */
+function exitGracefullyWithError(errMessage: string, originalBranch: string) {
```
devversion, line 135: I might be wrong here, but is this contradictory? i.e. `graceful` and then throwing an error? 

<sup>Also nit: Missing line between the function above</sup>

dev-infra/pr/discover-new-conflicts.ts
```diff
@@ -0,0 +1,161 @@
...
+    name: graphQLTypes.string,
+    repository: {
+      sshUrl: graphQLTypes.string,
+      nameWithOwner: graphQLTypes.string,
+    },
+  },
+  baseRef: {
+    name: graphQLTypes.string,
+    repository: {
+      sshUrl: graphQLTypes.string,
```
devversion, line 34: Out of curiosity: Didn't we talk about using http for these things? The merge script only uses http with the provided access token. The reason for this is that SSH usually requires a password to be entered (unless the ssh agent runs), and this is not ideal for a tool that establishes multiple fetches?

dev-infra/pr/discover-new-conflicts.ts
```diff
@@ -0,0 +1,161 @@
...
+/** Convert raw Pull Request response from Github to usable Pull Request object. */
+function processPr(pr: RawPullRequest) {
+  return {...pr, updatedAt: (new Date(pr.updatedAt)).getTime()};
+}
+
+/* Pull Request object after processing, derived from the return type of the processPr function. */
+type PullRequest = ReturnType<typeof processPr>;
+
+/** Name of a temporary local branch that is used for checking conflicts. **/
+const newBaseRef = '__NgDevRepoBaseAfterChange__';
```
devversion, line 56: Nice branch name 😄 Nit: can we make it more obvious in the variable name that this is the local temporary working branch?

dev-infra/utils/git.ts
```diff
@@ -0,0 +1,25 @@
...
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {exec as _exec} from 'shelljs';
+
+/* Run all exec commands as silent. */
+function exec(cmd: string) {
```
devversion, line 12: Did you consider exposing this function somewhere? I see that we use a similar function quite often (also in the actual subtool here)

dev-infra/pr/discover-new-conflicts.ts
```diff
@@ -0,0 +1,161 @@
...
+    process.exit(1);
+  }
+
+  const sameTargetPrs = allPendingPRs.filter(pr => pr.baseRef.name === requestedPr.baseRef.name);
+  const mergablePrs = sameTargetPrs.filter(pr => pr.mergeable === 'MERGEABLE');
+  const pendingPrs = mergablePrs.filter(pr => pr.updatedAt >= updatedAfter);
+  console.info(`Retrieved ${allPendingPRs.length} pending PRs`);
+  console.info(
+      `${sameTargetPrs.length} of these PRs target the same branch as #${requestedPr.number}`);
+  console.info(`${mergablePrs.length} of these PRs are currently mergable`);
```
devversion, line 95: I think this is nice information to provide. I'm still slightly unsure though if this information is helpful though given that the `mergeable` state is not always guaranteed to be up-to-date by Github, right?

devversion (approved):
LGTM. Just a few minor comments/questions.

dev-infra/pr/discover-new-conflicts.ts
josephperrott: Fixed, I was unable to find the http url in the GraphQL schema initially.

PR #37052, opened 2020-05-11
docs(platform-webworker): remove mention of version 10

kara:
The deprecation notice for platform-webworker
APIs is too prescriptive and notes that we will
remove the package in version 10. Since we are
not planning to do this for version 10, this
commit updates the notice to read "a future
version of Angular".


pkozlowski-opensource (approved):
LGTM

PR #37053, opened 2020-05-11
Update API reference doc for core

jbogarthyde:
Add introductions to usage examples and edit descriptions to be more complete and consistent with current API reference styles

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Many examples have not descriptions or introduction. Language is not always clear or consistent.

Issue Number: N/A

## What is the new behavior?

Updates API doc in the core package for consistency of style. Mostly adds introduction to examples, 

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/change_detection/change_detector_ref.ts
+++ b/packages/core/src/change_detection/change_detector_ref.ts
@@ -9,10 +9,13 @@
 import {injectChangeDetectorRef as render3InjectChangeDetectorRef} from '../render3/view_engine_compatibility';
 
 /**
- * Base class for Angular Views, provides change detection functionality.
+ * Base class for `ViewRef` that provides change detection functionality.
  * A change-detection tree collects all views that are to be checked for changes.
  * Use the methods to add and remove views from the tree, initiate change-detection,
- * and explicitly mark views as _dirty_, meaning that they have changed and need to be rerendered.
+ * and explicitly mark views as _dirty_, meaning that they have changed and need to be re-rendered.
+ *
+ * @see [Using change detection hooks](guide/lifecycle-hooks#using-change-detection-hooks)
+ * @see [Defining custom change detection](guide/lifecycle-hooks#defining-custom-change-detection)
  *
  * @usageNotes
  *
--- a/packages/core/src/change_detection/change_detector_ref.ts
+++ b/packages/core/src/change_detection/change_detector_ref.ts
@@ -9,7 +9,7 @@
 import {injectChangeDetectorRef as render3InjectChangeDetectorRef} from '../render3/view_engine_compatibility';
 
 /**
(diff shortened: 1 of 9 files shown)


packages/core/src/change_detection/change_detector_ref.ts
```diff
@@ -9,10 +9,13 @@
 import {injectChangeDetectorRef as render3InjectChangeDetectorRef} from '../render3/view_engine_compatibility';
 
 /**
- * Base class for Angular Views, provides change detection functionality.
+ * Base class for `ViewRef` that provides change detection functionality.
```
pkozlowski-opensource, line 12: The fact that `ViewRef` extends `ChangeDetectorRef` is the internal implementation detail and I don't think that we need to expose this information to the end users. How about "Class that provides change detection functionality"?

packages/core/src/linker/view_ref.ts
```diff
@@ -10,12 +10,12 @@ import {ApplicationRef} from '../application_ref';
...
 
 /**
- * Represents an Angular [view](guide/glossary#view),
- * specifically the [host view](guide/glossary#view-tree) that is defined by a component.
- * Also serves as the base class
- * that adds destroy methods for [embedded views](guide/glossary#view-tree).
+ * Represents an Angular [view](guide/glossary#view "Definition"),
+ * specifically the [host view](guide/glossary#view-tree "Definition") that is defined by a
+ * component. Also serves as the base class that adds destroy methods for [embedded
+ * views](guide/glossary#view-tree "Definition").
```
pkozlowski-opensource, line 16: I know that this is a pre-existing text but I find it incorrect. In reality a `ViewRef` can represent any type of view (a component host view or an embedded view) while the text makes it sounds like it is mostly component view. 

How about simply saying "Represents an Angular view", link to the glossary and skip the rest?

packages/core/src/metadata/directives.ts
```diff
@@ -314,7 +314,7 @@ export interface ComponentDecorator {
    * An Angular app contains a tree of Angular components.
    *
    * Angular components are a subset of directives, always associated with a template.
-   * Unlike other directives, only one component can be instantiated per an element in a template.
+   * Unlike other directives, only one component can be instantiated for one element in a template.
```
pkozlowski-opensource, line 317: nit: my personal preference would go to the oryginal phrasing as for me it was better conveying the fact that there can be only one component for a given element. Maybe let's use "only one component can be instantiated for a given element in a template" if we want to get rid of "per"?

pkozlowski-opensource (approved):
LGTM with few comments that I would like to see addressed before merging (hence cleanup label)

jbogarthyde:
Docs only, should not need G3

PR #37071, opened 2020-05-12
feat(platform-server): use absolute URLs from Location for HTTP

CaerusKaru:
Currently, requests from the server that do not use absolute URLs
fail because the server does not have the same fallback mechanism
that browser XHR does. This adds that mechanism by pulling the
full URL out of the document.location object, if available.

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/packages/platform-server/src/http.ts
+++ b/packages/platform-server/src/http.ts
@@ -10,10 +10,11 @@
 const xhr2: any = require('xhr2');
 
 import {Injectable, Injector, Provider} from '@angular/core';
-
+import {DOCUMENT} from '@angular/common';
 import {HttpEvent, HttpRequest, HttpHandler, HttpBackend, XhrFactory, ɵHttpInterceptingHandler as HttpInterceptingHandler} from '@angular/common/http';
-
 import {Observable, Observer, Subscription} from 'rxjs';
+
+const isAbsoluteUrl = /^[a-zA-Z\-\+.]+:\/\//;
 
 @Injectable()
 export class ServerXhr implements XhrFactory {
@@ -102,11 +103,14 @@ export abstract class ZoneMacroTaskWrapper<S, R> {
 
 export class ZoneClientBackend extends
     ZoneMacroTaskWrapper<HttpRequest<any>, HttpEvent<any>> implements HttpBackend {
-  constructor(private backend: HttpBackend) {
+  constructor(private backend: HttpBackend, private doc: Document) {
     super();
   }
 
   handle(request: HttpRequest<any>): Observable<HttpEvent<any>> {
+    if (!isAbsoluteUrl.test(request.url) && this.doc.location.href) {
+      return this.wrap(request.clone({url: this.doc.location.href + request.url}));
+    }
     return this.wrap(request);
   }
 
@@ -115,12 +119,16 @@ export class ZoneClientBackend extends
   }
 }
 
-export function zoneWrappedInterceptingHandler(backend: HttpBackend, injector: Injector) {
+export function zoneWrappedInterceptingHandler(
+    backend: HttpBackend, injector: Injector, doc: Document) {
   const realBackend: HttpBackend = new HttpInterceptingHandler(backend, injector);
-  return new ZoneClientBackend(realBackend);
+  return new ZoneClientBackend(realBackend, doc);
 }
 
 export const SERVER_HTTP_PROVIDERS: Provider[] = [
-  {provide: XhrFactory, useClass: ServerXhr},
-  {provide: HttpHandler, useFactory: zoneWrappedInterceptingHandler, deps: [HttpBackend, Injector]}
+  {provide: XhrFactory, useClass: ServerXhr}, {
+    provide: HttpHandler,
+    useFactory: zoneWrappedInterceptingHandler,
+    deps: [HttpBackend, Injector, DOCUMENT]
+  }
 ];
--- a/packages/platform-server/src/http.ts
+++ b/packages/platform-server/src/http.ts
@@ -14,7 +14,9 @@ import {DOCUMENT} from '@angular/common';
 import {HttpEvent, HttpRequest, HttpHandler, HttpBackend, XhrFactory, ɵHttpInterceptingHandler as HttpInterceptingHandler} from '@angular/common/http';
 import {Observable, Observer, Subscription} from 'rxjs';
 
+// @see https://www.w3.org/Protocols/HTTP/1.1/draft-ietf-http-v11-spec-01#URI-syntax
 const isAbsoluteUrl = /^[a-zA-Z\-\+.]+:\/\//;
+const FORWARD_SLASH = '/';
 
 @Injectable()
 export class ServerXhr implements XhrFactory {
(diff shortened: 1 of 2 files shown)


CaerusKaru:
cc @alan-agius4

packages/platform-server/src/http.ts
```diff
@@ -10,11 +10,12 @@
 const xhr2: any = require('xhr2');
 
 import {Injectable, Injector, Provider} from '@angular/core';
-
+import {DOCUMENT} from '@angular/common';
 import {HttpEvent, HttpRequest, HttpHandler, HttpBackend, XhrFactory, ɵHttpInterceptingHandler as HttpInterceptingHandler} from '@angular/common/http';
-
 import {Observable, Observer, Subscription} from 'rxjs';
 
+const isAbsoluteUrl = /^[a-zA-Z\-\+.]+:\/\//;
```
kyliau, line 17: Is there a spec for the scheme? If so, could you please add a comment for it?
I'm surprised that characters like `+` are allowed.

packages/platform-server/test/integration_spec.ts
```diff
@@ -793,6 +793,23 @@ describe('platform-server integration', () => {
          });
        }));
 
+    it('can make relative HttpClient requests', async(() => {
+         const platform = platformDynamicServer([
+           {provide: INITIAL_CONFIG, useValue: {document: '<app></app>', url: 'http://localhost'}}
+         ]);
+         platform.bootstrapModule(HttpClientExampleModule).then(ref => {
```
kyliau, line 800: Consider using await instead of then, since this is an `async` function.

packages/platform-server/test/integration_spec.ts
```diff
@@ -793,6 +793,23 @@ describe('platform-server integration', () => {
...
+           const mock = ref.injector.get(HttpTestingController) as HttpTestingController;
+           const http = ref.injector.get(HttpClient);
+           ref.injector.get<NgZone>(NgZone).run(() => {
```
kyliau, line 803: Is the explicit cast for `NgZone` here necessary?

packages/platform-server/test/integration_spec.ts
```diff
@@ -793,6 +793,23 @@ describe('platform-server integration', () => {
...
+             http.get<string>('/testing').subscribe((body: string) => {
```
kyliau, line 804: Likewise, is the explicit cast for `string` necessary?

packages/platform-server/src/http.ts
```diff
@@ -102,11 +103,14 @@ export abstract class ZoneMacroTaskWrapper<S, R> {
...
 export class ZoneClientBackend extends
     ZoneMacroTaskWrapper<HttpRequest<any>, HttpEvent<any>> implements HttpBackend {
-  constructor(private backend: HttpBackend) {
+  constructor(private backend: HttpBackend, private doc: Document) {
     super();
   }
 
   handle(request: HttpRequest<any>): Observable<HttpEvent<any>> {
+    if (!isAbsoluteUrl.test(request.url) && this.doc.location.href) {
+      return this.wrap(request.clone({url: this.doc.location.href + request.url}));
```
kyliau, line 112: Do we have to consider trailing slash in `this.doc.location.href` or leading slash in `request.url`?
i.e.  what if the slash is missing or is present in both?

kyliau (changes requested):

kyliau:
g3 presubmit: http://test/OCL:311212750:BASE:311212803:1589321997009:95b55e6c

packages/platform-server/test/integration_spec.ts
CaerusKaru: `string` is necessary, otherwise `body` is not typed correctly.

kyliau (approved):

naveedahmed1:
Should this change create any side effects? I upgraded my app from Angular `10.0.0-next.6` to `10.0.0-rc.0` and suddenly it broke the SSR for my app. I downgraded `"@angular/platform-server": "10.0.0-next.6",` and everything started working fine.

alan-agius4:
@naveedahmed1, sorry to hear that this change caused a regression.

Please fill an issue together with a minimal reproduction.

Thanks.

naveedahmed1:
Thanks @alan-agius4, the reproduction would be a little difficult, since in my case its a .net core app.

Let me share the situation, if this can be helpful: 

1. In my case I am already converting the request url from relative to absolute in my `http interceptor`'s `intercept` method.
2. When on server, when I receive a response of the http request I add it to `this.transferState.set(CACHE_KEY, event as any);` and on client, I check if we already have the cached response, its served from cache otherwise request is sent to the server.

Does the change in this pull request honor this change I make to request url?

CaerusKaru:
This PR should leave absolute URL’s alone. It only injects the host name at the tail end of the request, eg right before the request is made. So if you provide an absolute URL, everything should work as it has in the past.

As @alan-agius4 said, if this is an issue for you, please open an issue and provide a minimal reproduction. Without this, we can’t reasonably help further.

naveedahmed1:
Thank you @CaerusKaru for your response, I am just reviewing my code and investigating it on my end. If I don't find any solution I will definitely open an a new issue.

If you could please just comment on this:

I have a url `http://localhost:53189/jobs/job/204182`, when I open it in my browser the request goes to server which fetch data from `https://www.mustakbil.com/ws/jobs/job/204182`, render it and returns final response to client. In my component I make `http` request to `/ws/jobs/job/204182` and then in intercept method of my interceptor I convert it to `https://www.mustakbil.com/ws/jobs/job/204182`.

After this pull request, if I `console log(req.url)` in `intercept` method of my interceptor, before any modification to the request it shows `/jobs/job/204182/ws/jobs/job/204182`, whereas I expect it to be `/ws/jobs/job/204182` because that's what I passed from my component i.e. `this.http.get("/ws/jobs/job/204182")`. Any idea what's happening here?

One more question, does this pull request modifies the url after user defined interceptors have been executed?

CaerusKaru:
As I said in my previous comment, the new logic is executed at the tail end of the request, after all interceptors complete. That’s why you won’t see any changes in your code.

naveedahmed1:
Can I add you to a private angular repo, and if you could please review the interceptor?

alan-agius4:
@naveedahmed1, can you please open an issue? Ideally the reproduction provide contain the least possible amount of code needed to demonstrate the issue. 

Unfortunately, if the reproduction is complex or large it will take a lot of effort to debug and get to the bottom of the issue.

Also, PR’s is not the correct place to discuss issues.

Thanks for your understanding.

naveedahmed1:
Thanks @alan-agius4 , I have created a new issue here https://github.com/angular/angular/issues/37314

packages/platform-server/src/http.ts
```diff
@@ -102,11 +105,21 @@ export abstract class ZoneMacroTaskWrapper<S, R> {
 
 export class ZoneClientBackend extends
     ZoneMacroTaskWrapper<HttpRequest<any>, HttpEvent<any>> implements HttpBackend {
-  constructor(private backend: HttpBackend) {
+  constructor(private backend: HttpBackend, private doc: Document) {
     super();
   }
 
   handle(request: HttpRequest<any>): Observable<HttpEvent<any>> {
+    const href = this.doc.location.href;
```
naveedahmed1, line 113: Shouldn't we use `window.location.hostname` or `location.location.hostname` ? Since `Location href` Property returns the entire URL (of the current page) and we just need the `hostname` to prepend to convert the relative url to absolute url.

jedjebari:
I have the same issue than you @naveedahmed1 .

If we bootstrap the app on root route (http://myapp/), then everything is ok.
But if we bootstrap on a feature root (http://myapp/myfeature), then it breaks, because the http request is done on http://myapp/myfeature/api/blablabla instead of http://myapp/api/blablabla.

So that's not so complex to reproduce. Can someone else try on its side ?

PR #37073, opened 2020-05-12
refactor(core): remove ActiveIndexFlag from LContainer

atscott:
The `ActiveIndexFlag` is no longer needed because we no longer have "inline embedded views".
There is only one type of embedded view so we do not need complex tracking for
inline embedded views.

`HAS_TRANSPLANTED_VIEWS` now takes the place of the `ACTIVE_INDEX` slot as a
simple boolean rather than being a shifted flag inside the `ACTIVE_INDEX` bits.

This is a follow-up of https://github.com/angular/angular/pull/34715

### DIFF
--- a/packages/core/src/render3/interfaces/container.ts
+++ b/packages/core/src/render3/interfaces/container.ts
@@ -26,7 +26,16 @@ export const TYPE = 1;
  * without having to remember the specific indices.
  * Uglify will inline these when minifying so there shouldn't be a cost.
  */
-export const ACTIVE_INDEX = 2;
+
+/**
+ * Flag to signify that this `LContainer` may have transplanted views which need to be change
+ * detected. (see: `LView[DECLARATION_COMPONENT_VIEW])`.
+ *
+ * This flag, once set, is never unset for the `LContainer`. This means that when unset we can skip
+ * a lot of work in `refreshDynamicEmbeddedViews`. But when set we still need to verify
+ * that the `MOVED_VIEWS` are transplanted and on-push.
+ */
+export const HAS_TRANSPLANTED_VIEWS = 2;
 
 // PARENT, NEXT, TRANSPLANTED_VIEWS_TO_REFRESH are indices 3, 4, and 5
 // As we already have these constants in LView, we don't need to re-create them.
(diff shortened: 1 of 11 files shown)


atscott:
[ivy presubmit](https://test.corp.google.com/ui#id=OCL:311163039:BASE:311163095:1589306959659:20490f68)

packages/core/src/render3/util/view_traversal_utils.ts
```diff
@@ -53,3 +55,26 @@ export function getRootContext(viewOrComponent: LView|{}): RootContext {
       assertDefined(rootView[CONTEXT], 'RootView has no context. Perhaps it is disconnected?');
   return rootView[CONTEXT] as RootContext;
 }
+
+
+/**
+ * Gets the first `LContainer` in the LView or `null` if none exists.
+ */
+export function getFirstLContainer(lView: LView): LContainer|null {
+  let viewOrContainer = lView[CHILD_HEAD];
```
AndrewKushnir, line 64: @atscott just a quick comment: it looks like `getFirstLContainer` and `getNextLContainer` functions are similar and the only delta is what the initial value is. I was wondering if it makes sense to create a shared function to encapsulate common logic (if there is no perf implications on calling extra fn)?

packages/core/src/render3/util/view_traversal_utils.ts
```diff
@@ -53,3 +55,25 @@ export function getRootContext(viewOrComponent: LView|{}): RootContext {
...
+}
+
+/**
+ * Gets the next `LContainer` that is a sibling of the given container.
+ */
+export function getNextLContainer(container: LContainer): LContainer|null {
+  return getNearestLContainer(container[NEXT]);
+}
+
+function getNearestLContainer(viewOrContainer: LContainer|LView|null) {
```
kara, line 74: Nice cleanup 👍

packages/core/src/render3/interfaces/container.ts
```diff
@@ -26,7 +26,16 @@ export const TYPE = 1;
  * without having to remember the specific indices.
  * Uglify will inline these when minifying so there shouldn't be a cost.
  */
-export const ACTIVE_INDEX = 2;
```
kara, line 29: Looks like we're missing one more reference to `ACTIVE_INDEX` in `interfaces/view.ts`. Remove?

kara (approved):
LGTM

atscott:
[ivy presubmit](https://test.corp.google.com/ui#id=OCL:311163039:BASE:311380640:1589397913043:e4b352c5)

PR #37075, opened 2020-05-12
ngcc: implement a new program-based entry-point finder

petebacondarwin:
Add a new `EntryPointFinder` that can be seeded from the imports in the program specified by a tsconfig.json file. This should be much faster than the `DirectoryWalkerEntryPointFinder` when the active program only imports a small proportion of the installed entry-points.

On the AIO project only half the entry-points would be processed by ngcc if the CLI async integration with ngcc made use of the new flag.

### DIFF
--- a/packages/compiler-cli/ngcc/src/command_line_options.ts
+++ b/packages/compiler-cli/ngcc/src/command_line_options.ts
@@ -36,7 +36,14 @@ export function parseCommandLineOptions(args: string[]): NgccOptions {
             alias: 'target',
             describe:
                 'A relative path (from the `source` path) to a single entry-point to process (plus its dependencies).\n' +
-                'If this property is provided then `error-on-failed-entry-point` is forced to true',
+                'If this property is provided then `error-on-failed-entry-point` is forced to true.\n' +
+                'This option overrides the `--use-program-dependencies` option.',
+          })
+          .option('use-program-dependencies', {
+            type: 'boolean',
+            describe:
+                'If this property is provided then the entry-points to process are parsed from the program defined by the loaded tsconfig.json. See `--tsconfig`.\n' +
+                'This option is overridden by the `--target` option.',
           })
           .option('first-only', {
             describe:
@@ -116,6 +123,7 @@ export function parseCommandLineOptions(args: string[]): NgccOptions {
   const enableI18nLegacyMessageIdFormat = options['legacy-message-ids'];
   const invalidateEntryPointManifest = options['invalidate-entry-point-manifest'];
   const errorOnFailedEntryPoint = options['error-on-failed-entry-point'];
+  const findEntryPointsFromTsConfigProgram = options['use-tsconfig'];
   // yargs is not so great at mixed string+boolean types, so we have to test tsconfig against a
   // string "false" to capture the `tsconfig=false` option.
   // And we have to convert the option to a string to handle `no-tsconfig`, which will be `false`.
@@ -134,6 +142,7 @@ export function parseCommandLineOptions(args: string[]): NgccOptions {
     async: options['async'],
     invalidateEntryPointManifest,
     errorOnFailedEntryPoint,
-    tsConfigPath
+    tsConfigPath,
+    findEntryPointsFromTsConfigProgram
   };
 }
--- a/packages/compiler-cli/ngcc/src/command_line_options.ts
+++ b/packages/compiler-cli/ngcc/src/command_line_options.ts
@@ -123,7 +123,7 @@ export function parseCommandLineOptions(args: string[]): NgccOptions {
   const enableI18nLegacyMessageIdFormat = options['legacy-message-ids'];
   const invalidateEntryPointManifest = options['invalidate-entry-point-manifest'];
   const errorOnFailedEntryPoint = options['error-on-failed-entry-point'];
-  const findEntryPointsFromTsConfigProgram = options['use-tsconfig'];
+  const findEntryPointsFromTsConfigProgram = options['use-program-dependencies'];
   // yargs is not so great at mixed string+boolean types, so we have to test tsconfig against a
   // string "false" to capture the `tsconfig=false` option.
   // And we have to convert the option to a string to handle `no-tsconfig`, which will be `false`.
@@ -143,6 +143,6 @@ export function parseCommandLineOptions(args: string[]): NgccOptions {
     invalidateEntryPointManifest,
     errorOnFailedEntryPoint,
     tsConfigPath,
-    findEntryPointsFromTsConfigProgram
+    findEntryPointsFromTsConfigProgram,
   };
 }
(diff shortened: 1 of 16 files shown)


petebacondarwin:
The approach here, which involves parsing each source file, could be slow for a large app.
We should investigate using the TS scanner directly or even a simple regular expression, since we only care about capturing the import paths and can be quite tolerant of false positives.

packages/compiler-cli/ngcc/test/entry_point_finder/program_based_entry_point_finder_spec.ts
```diff
@@ -0,0 +1,147 @@
...
+      logger = new MockLogger();
+      const srcHost = new EsmDependencyHost(fs, new ModuleResolver(fs));
+      const dtsHost = new DtsDependencyHost(fs);
+      config = new NgccConfiguration(fs, projectPath);
+      resolver = new DependencyResolver(fs, logger, config, {esm2015: srcHost}, dtsHost);
+      manifest = new EntryPointManifest(fs, config, logger);
+    });
+
+    describe('findEntryPoints()', () => {
+      it('should find entry-points imported into the program', () => {
```
JoostK, line 50: I would like to see a test for the scenario where path mappings into TS source files are used, as that is done in some projects (e.g. [this article](https://blog.angulartraining.com/path-mapping-with-typescript-cac16d12c416) and [another one](https://dev.to/cooperdev/restructure-with-ease-thanks-to-typescript-path-mappings-4b0e)).

As these are absolute imports, ngcc's module resolver will probably recognize them as such and not traverse into the source files, causing parts of the program not to be scanned. I haven't verified this assumption, so could be wrong here.

JoostK (commented):
Typo in "capture dynamic import expressions as well as declarations" commit:

- "and creates lesss" -> "and creates less"

packages/compiler-cli/ngcc/src/command_line_options.ts
```diff
@@ -116,6 +123,7 @@ export function parseCommandLineOptions(args: string[]): NgccOptions {
   const enableI18nLegacyMessageIdFormat = options['legacy-message-ids'];
   const invalidateEntryPointManifest = options['invalidate-entry-point-manifest'];
   const errorOnFailedEntryPoint = options['error-on-failed-entry-point'];
+  const findEntryPointsFromTsConfigProgram = options['use-tsconfig'];
```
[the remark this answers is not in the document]
petebacondarwin: 😱

packages/compiler-cli/ngcc/src/dependencies/esm_dependency_host.ts
```diff
@@ -13,20 +13,200 @@ import {DependencyHostBase} from './dependency_host';
...
+      token = this.skipExplicitClause();
+    }
+    // Expect a `from` clause, if not bail out
+    if (token !== ts.SyntaxKind.FromKeyword) {
+      return null;
+    }
+    return this.tryStringLiteral();
+  }
+
+  protected skipStarClause(): ts.SyntaxKind|null {
```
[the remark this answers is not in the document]
petebacondarwin: Strictly they have two names depending upon whether they are in an `import` or an `export` statement, which is why I didn't go with that name. But given that it was weird enough for you to comment, I will change it.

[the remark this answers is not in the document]
petebacondarwin: I'm going with `skipNamespacedClause()` and `skipNamedClause()`.

packages/compiler-cli/ngcc/test/entry_point_finder/program_based_entry_point_finder_spec.ts
petebacondarwin: Actually ngcc's module resolver understands pathMappings and so would indeed resolve these. I will add to the test...

petebacondarwin: Added

packages/compiler-cli/ngcc/src/dependencies/esm_dependency_host.ts
```diff
@@ -13,20 +13,200 @@ import {DependencyHostBase} from './dependency_host';
...
+        case ts.SyntaxKind.ExportKeyword:
+          const reexportPath = this.extractReexportPath();
+          if (reexportPath !== null) {
+            imports.add(reexportPath);
+          }
+          break;
+      }
+      lastToken = currentToken;
+    }
+    return imports;
```
JoostK, line 76: Just thought of this: consider resetting the scanner's text to the empty string to avoid holding on to the most recent file content.

petebacondarwin: Sounds reasonable.

petebacondarwin: Done here https://github.com/angular/angular/pull/37075/commits/ffc717a68be19616f77676b207c3196cdb21e95a

JoostK (approved):

gkalpak (approved):
Reviewed-for: dev-infra

packages/compiler-cli/ngcc/src/dependencies/esm_dependency_host.ts
```diff
@@ -93,7 +225,7 @@ export class EsmDependencyHost extends DependencyHostBase {
  * in this file, true otherwise.
  */
 export function hasImportOrReexportStatements(source: string): boolean {
-  return /(import|export)\s.+from/.test(source);
+  return /(?:import|export)[\s\S]+?(["'])(?:(?:\\\1|.)*?)\1/.test(source);
```
gkalpak, line 228: Nice :+1:

There is one unneeded set of parenthesis (afaict), but not really important:
```suggestion
  return /(?:import|export)[\s\S]+?(["'])(?:\\\1|.)+?\1/.test(source);
```

packages/compiler-cli/ngcc/src/entry_point_finder/tracing_entry_point_finder.ts
```diff
@@ -0,0 +1,206 @@
...
+import {getBasePaths} from './utils';
+
+/**
+ * An EntryPointFinder that starts from a set of initial files and only returns entry-points that
+ * are dependencies of these files.
+ *
+ * This is faster than searching the entire file-system for all the entry-points,
+ * and is used primarily by the CLI integration.
+ *
+ * There are two concrete implementation of this class.
```
gkalpak, line 27:
```suggestion
 * There are two concrete implementations of this class.
```

gkalpak (commented):
Post-merge review :grin::
**tl;dr** LGTM :+1:

packages/compiler-cli/ngcc/src/dependencies/esm_dependency_host.ts
petebacondarwin: I'm a bit frightened to touch that again :-)

petebacondarwin:
Thanks for the extra review @gkalpak - here is a fixup #37461

PR #37080, opened 2020-05-13
fix(zone.js): should call isIE as function.

JiaLiPassion:
zone.js ignores patching `window.onerror` property in IE browser, so zone.js need to check
whether the current browser is IE, and before this PR, zone.js does not check `isIE` correctly,
since `isIE` is not a flag but a function.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

### DIFF
--- a/packages/compiler/src/i18n/digest.ts
+++ b/packages/compiler/src/i18n/digest.ts
@@ -6,7 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {newArray, utf8Encode} from '../util';
+import {Byte, newArray, utf8Encode} from '../util';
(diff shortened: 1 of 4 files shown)


gkalpak:
I think this is fixed in #36989.

alan-agius4:
@gkalpak, master would be yes, but not patch. So probably we should change the target label to patch only in here.

JiaLiPassion:
@gkalpak, @alan-agius4 , thank you for the review, if the PR https://github.com/angular/angular/pull/36989 is not for `patch`, I will change this PR's target to `patch-only`.

mhevery (approved):

PR #37096, opened 2020-05-13
ci: set up dedicated workflow for running windows tests

devversion:


devversion:
Need to chat with the CircleCI team about this. It looks like `PR action: rerun CI at HEAD` does not work anymore, and it's not possible to trigger pipelines for PRs through the CircleCI rest API.

JoostK:
@devversion I think this has become obsolete now that Windows jobs are always run for PRs?

devversion:
@JoostK Seems like it; I'm unsure how much of a problem windows CI currently is in terms of speed, but we could consider running windows only on push builds; and getting the label working.

PR #37097, opened 2020-05-13
build: add github config for ng-dev configuration

josephperrott:
Adds the gitub configuration to the ng-dev configuration. This github
configuration provides information needed for making API requests to
github.  Upcoming tooling related PRs will require these API requests
being possible.

### DIFF
--- a/dev-infra/utils/config.ts
+++ b/dev-infra/utils/config.ts
@@ -12,8 +12,8 @@ import {exec} from 'shelljs';
 /** The common configuration for ng-dev. */
 type CommonConfig = {
   github: {
-    owner: string;
-    name: string;
+    owner: string,
+    name: string,
   }
 };
 
--- a/dev-infra/utils/config.ts
+++ b/dev-infra/utils/config.ts
@@ -11,7 +11,7 @@ import {exec} from 'shelljs';
 
 /** The common configuration for ng-dev. */
 type CommonConfig = {
-  // Github repository configuration used for API Requests, determining upstream remote, etc.
+  /* Github repository configuration used for API Requests, determining upstream remote, etc. */
   github: {
     owner: string,
     name: string,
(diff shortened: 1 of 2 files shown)


dev-infra/utils/config.ts
```diff
@@ -46,9 +49,21 @@ export function getConfig(): NgDevConfig {
...
 
 /** Validate the common configuration has been met for the ng-dev command. */
-function validateCommonConfig(config: NgDevConfig<CommonConfig>) {
-  // TODO: add validation for the common configuration
-  return config;
+function validateCommonConfig(config: Partial<NgDevConfig<CommonConfig>>) {
+  const errors: string[] = [];
+  // Validate the github configuration.
+  if (config.github === undefined) {
+    errors.push(`No configuration defined for "format"`);
```
devversion, line 56:
```suggestion
    errors.push(`Github repository not configured. Set the "github" option.`);
```

dev-infra/utils/config.ts
```diff
@@ -46,9 +49,21 @@ export function getConfig(): NgDevConfig {
...
+  } else {
+    if (config.github.name === undefined) {
+      errors.push(`"github.name" is not defined`);
+    }
+    if (config.github.owner === undefined) {
+      errors.push(`"github.org"  is not defined`);
```
devversion, line 62: Extra whitespace and the option seems to be called `owner` instead.

dev-infra/utils/config.ts
```diff
@@ -46,9 +49,21 @@ export function getConfig(): NgDevConfig {
...
+    }
+  }
+
+  return config as NgDevConfig<CommonConfig>;
```
devversion, line 66: Can this just be `NgDevConfig` as that already will include the common config?

dev-infra/utils/config.ts
```diff
@@ -11,7 +11,10 @@ import {exec} from 'shelljs';
 
 /** The common configuration for ng-dev. */
 type CommonConfig = {
-    // TODO: add common configuration
+  github: {
+    owner: string;
+    name: string;
+  }
```
devversion, line 17: It would be great to have a description for these properties I think. so that it is obvious what this configuration is needed for. i.e. API requests, determining upstream remote; pushing upstream etc.

josephperrott:
@devversion Ready for another look

dev-infra/utils/config.ts
```diff
@@ -11,7 +11,11 @@ import {exec} from 'shelljs';
 
 /** The common configuration for ng-dev. */
 type CommonConfig = {
-    // TODO: add common configuration
+  // Github repository configuration used for API Requests, determining upstream remote, etc.
```
devversion, line 14: Optional nit: Use multi-line comment so the description shows up in IDE's

devversion (approved):

PR #37100, opened 2020-05-13
docs(router): Update router guide to use UrlTree for guard redirects

### LINKED ISSUE #36187: Navigating to lazy loaded module from route guard causes the URL path to briefly be blank
# 🐞 bug report

### Affected Package

### Is this a regression?
Not sure

### Description
If you have a page that is protected by a route guard which triggers a redirect (thus cancelling the current navigation) to a page which is lazy-loaded, the URL will briefly read just the host name (`/`) while the module loads. (An example of this would be having a dashboard that's protected by a user login guard that redirects to the login page (which in my case is lazy-loaded).)

This is further an issue because it changes the history so that if the user hits the back button, they will be taken to `/` rather than the page they originally tried to navigate to.

Also, if there is a network issue retrieving the module, the user will be abandoned on the home page. If they refresh to try to fix the issue, they will no longer be where they started.

## 🔬 Minimal Reproduction
[angular-lazy-load-guard-test.zip](https://github.com/angular/angular/files/4363701/angular-lazy-load-guard-test.zip)

Navigate to `/page1` and you should get redirected to `/page2`. However, notice that the URL briefly reads `/`. If this is not apparent, try adding network throttling (a Firefox feature) to slow down the retrieval of the lazy module.

I also have a second redirect that emulates the behavior of the guard in the Page1Component. If you disable the card, the page1 component will still redirect. *However*, note that the URL never reads blank at this point...it always shows something. 

## 🔥 Exception or Error
N/A

## 🌍  Your Environment

**Angular Version:**
<pre><code>
Angular CLI: 9.0.7
Node: 12.3.1
OS: linux x64

Angular: 9.0.7
... router
Ivy Workspace: Yes

@angular-devkit/architect         0.900.7
@angular-devkit/build-angular     0.900.7
@angular-devkit/build-optimizer   0.900.7
@angular-devkit/build-webpack     0.900.7
@angular-devkit/core              9.0.7
@angular-devkit/schematics        9.0.7
@ngtools/webpack                  9.0.7
@schematics/angular               9.0.7
@schematics/update                0.900.7

</code></pre>

atscott:
The current implementation for redirecting users inside guards was in place
before the feature was added to allow `CanActivate` and `CanActivateChild` guards
to return `UrlTree` for redirecting users.

Returning `UrlTree` should be the default method, as it provides a more desirable
redirecting experience. When using `router.navigate` followed by `return false`,
the `Router` calls `resetUrlToCurrentUrlTree` (in the `finalize` operator) before
processing the navigation to the new route.  This can result in an undesirable
history if the navigation was the first navigation in the application - that is,
the route will briefly be reset to just `/` (see #36187).

Fixes #36187

Note: This PR targets master only because returning `UrlTree` is a feature that is only available to `CanActivateChild` guards in v10 (#36610).

### DIFF
--- a/aio/content/examples/router/src/app/auth/auth.guard.2.ts
+++ b/aio/content/examples/router/src/app/auth/auth.guard.2.ts
@@ -1,6 +1,6 @@
 // #docregion
 import { Injectable } from '@angular/core';
-import { CanActivate, ActivatedRouteSnapshot, RouterStateSnapshot, Router } from '@angular/router';
+import { CanActivate, ActivatedRouteSnapshot, RouterStateSnapshot, Router, UrlTree } from '@angular/router';
 
 import { AuthService }      from './auth.service';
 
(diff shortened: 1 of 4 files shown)


kapunahelewong (approved):
LGTM! Thank you, @atscott!!

PR #37108, opened 2020-05-14
fix(language-service): Recover from error in analyzing Ng Modules

### LINKED ISSUE #777: Server crashes on invalid entry component in version 0.901.7
**Describe the bug**

After updating to version 0.901.7 the language service crashes when an invalid entry component is encountered

**To Reproduce**
```
@NgModule({
  entryComponents: [
    CommonModule  // Any module here will do. Also something that is not a component works
  ]
})
export class SomeModule { }
```

**Expected behavior**

While the error in this case is correct, the language service should log the error, as the previous version, but not crash entirely.

**Logs**
```
c:\Users\cfede\.vscode\extensions\angular.ng-template-0.901.7\server\node_modules\@angular\language-service\bundles\language-service.umd.js:22314
                throw syntaxError(dirType.name + " cannot be used as an entry component.");
                ^

Error: CommonModule cannot be used as an entry component.
    at syntaxError (c:\Users\cfede\.vscode\extensions\angular.ng-template-0.901.7\server\node_modules\@angular\language-service\bundles\language-service.umd.js:2679:21)
    at CompileMetadataResolver._getEntryComponentMetadata (c:\Users\cfede\.vscode\extensions\angular.ng-template-0.901.7\server\node_modules\@angular\language-service\bundles\language-service.umd.js:22314:23)
    at c:\Users\cfede\.vscode\extensions\angular.ng-template-0.901.7\server\node_modules\@angular\language-service\bundles\language-service.umd.js:21950:57
    at Array.map (<anonymous>)
    at CompileMetadataResolver.getNgModuleMetadata (c:\Users\cfede\.vscode\extensions\angular.ng-template-0.901.7\server\node_modules\@angular\language-service\bundles\language-service.umd.js:21950:22)
    at c:\Users\cfede\.vscode\extensions\angular.ng-template-0.901.7\server\node_modules\@angular\language-service\bundles\language-service.umd.js:23828:57
    at Array.forEach (<anonymous>)
    at analyzeFile (c:\Users\cfede\.vscode\extensions\angular.ng-template-0.901.7\server\node_modules\@angular\language-service\bundles\language-service.umd.js:23798:57)
    at visitFile (c:\Users\cfede\.vscode\extensions\angular.ng-template-0.901.7\server\node_modules\@angular\language-service\bundles\language-service.umd.js:23774:32)
    at c:\Users\cfede\.vscode\extensions\angular.ng-template-0.901.7\server\node_modules\@angular\language-service\bundles\language-service.umd.js:23780:56 {
  ngSyntaxError: true
}
[Error - 15:51:29] Connection to server got closed. Server will not be restarted.
```

ayazhafiz:
Closes https://github.com/angular/vscode-ng-language-service/issues/777

## PR Checklist

## PR Type

- [x] Bugfix

- [ ] Yes
- [x] No

### DIFF
--- a/packages/language-service/src/typescript_host.ts
+++ b/packages/language-service/src/typescript_host.ts
@@ -175,8 +175,15 @@ export class TypeScriptServiceHost implements LanguageServiceHost {
       }
     };
     const programFiles = this.program.getSourceFiles().map(sf => sf.fileName);
-    this.analyzedModules =
-        analyzeNgModules(programFiles, analyzeHost, this.staticSymbolResolver, this.resolver);
+
+    try {
+      this.analyzedModules =
+          analyzeNgModules(programFiles, analyzeHost, this.staticSymbolResolver, this.resolver);
+    } catch (e) {
+      // Analyzing modules may throw; in that case, reuse the old modules.
+      this.error(`Analyzing NgModules failed. ${e}`);
+      return this.analyzedModules;
+    }
 
     // update template references and fileToComponent
     const urlResolver = createOfflineCompileUrlResolver();
--- a/packages/language-service/test/typescript_host_spec.ts
+++ b/packages/language-service/test/typescript_host_spec.ts
@@ -6,6 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
+import * as ngc from '@angular/compiler';
 import * as ts from 'typescript';
 
 import {TypeScriptServiceHost} from '../src/typescript_host';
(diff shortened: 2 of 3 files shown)


packages/language-service/src/typescript_host.ts
```diff
@@ -175,8 +177,13 @@ export class TypeScriptServiceHost implements LanguageServiceHost {
...
     const programFiles = this.program.getSourceFiles().map(sf => sf.fileName);
-    this.analyzedModules =
-        analyzeNgModules(programFiles, analyzeHost, this.staticSymbolResolver, this.resolver);
+
+    try {
+      this.analyzedModules = this.analyzeNgModules(
+          programFiles, analyzeHost, this.staticSymbolResolver, this.resolver);
+    } catch {
+      // Analyzing modules may throw; in that case, reuse the old modules.
+    }
```
ayazhafiz, line 186: We should propagate the error to an end user, maybe by attaching a logger to the host.

kyliau: The logger method is right in this class

https://github.com/angular/angular/blob/7acd33007d9bd30288a76fa2a3b6fe5948143d9f/packages/language-service/src/typescript_host.ts#L557-L566

packages/language-service/test/typescript_host_spec.ts
```diff
@@ -215,4 +216,29 @@ describe('TypeScriptServiceHost', () => {
     // But the content should be exactly the same
     expect(newModules).toEqual(oldModules);
   });
+
+  it('should recover from error in analyzing ng modules', () => {
+    // First create a TypescriptHost with empty script names
```
kyliau, line 221: Please try to override `/app/main.ts` with an invalid `entryComponent` in the `NgModule`.

kyliau (changes requested):

packages/language-service/src/typescript_host.ts
```diff
@@ -65,6 +65,8 @@ export class TypeScriptServiceHost implements LanguageServiceHost {
   private readonly collectedErrors = new Map<string, any[]>();
   private readonly fileVersions = new Map<string, string>();
 
+  analyzeNgModules = analyzeNgModules;
```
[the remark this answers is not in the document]
ayazhafiz: I changed the PR to use a spy. I would rather use a spy because I think we should test that we handle errors from analysis generally, not committing to a specific analysis error. Lmk what you think.

packages/language-service/test/typescript_host_spec.ts
ayazhafiz: See comment above

ayazhafiz:
ptal @kyliau. The error messages we get from resolution failure would make great diagnostics, so I think we should look into propagating it from the compiler.

packages/language-service/src/typescript_host.ts
```diff
@@ -71,6 +71,8 @@ export class TypeScriptServiceHost implements LanguageServiceHost {
     ngModuleByPipeOrDirective: new Map(),
     ngModules: [],
   };
+  private lastAnalysisSucceeded = true;
```
kyliau, line 74: Is there any reason why this flag is needed? I think the code would work without it.

ayazhafiz: yeah sorry, an artifact of when i was doing the spy thing and didn't update source files

kyliau (approved):
Thank you for fixing this!

PR #37110, opened 2020-05-14
fix(dev-infra): do not require a commit body for release commits

josephperrott:
Release commits do not require a commit body as the context, usually
provided in commit body, is already available in the process of
releasing.  No additional value is gained from adding a body message
on these commits.

### DIFF
--- a/dev-infra/commit-message/validate.ts
+++ b/dev-infra/commit-message/validate.ts
@@ -136,6 +136,11 @@ export function validateCommitMessage(
     return false;
   }
 
+  // Commits with the type of `release` do not require a commit body.
+  if (commit.type === 'release') {
+    return true;
+  }
+
   //////////////////////////
   // Checking commit body //
   //////////////////////////


devversion (approved):

dev-infra/commit-message/validate.ts
```diff
@@ -136,6 +136,11 @@ export function validateCommitMessage(
     return false;
   }
 
+  // Commits with the type of `release` do not require a commit body.
+  if (commit.type === 'release') {
```
devversion, line 140: @josephperrott Should this be configurable in the config? `release` as type is not guaranteed to exist right?

josephperrott: I think in a follow-up we should sort out how to do this configuration, hut for the short term, opting out for `release` fixes frameworks issue at hand.

devversion: Sounds good.

dev-infra/commit-message/validate.ts
```diff
@@ -136,6 +136,11 @@ export function validateCommitMessage(
...
+    return true;
```
gkalpak, line 141: IMO, the fact that `release` commits do not require a commit message body doesn't mean that we should not validate it if it exists. By returning here, we skip both the `minBodyLength` check (desirable) but also the `maxLineLength` check (undesirable imo). Also, if we add more checks below in the future, they will be skipped as well for `release` commits.

Ideally, we should only skip the `minBodyLength` check and continue with the rest imo.

PR #37111, opened 2020-05-14
refactor(core): assert TNode is not a container when setting attribute on element

### LINKED ISSUE #35994: Ivy regression: TypeError: el.setAttribute is not a function
# 🐞 bug report

### Is this a regression?
Yes, it worked well on Angular 8, and still works well on Angular 9 when disabling Ivy.

### Description
In a routed component, when applying a specific directive (`kendoGridSelectAllCheckbox`) on `ng-template`, an error is thrown: `TypeError: el.setAttribute is not a function`.
Again, this occurs only in a _routed_ component: if the same template is used directly in the AppComponent (or any other component that explicitly belongs to the root component tree), the issue is gone.

## 🔬 Minimal Reproduction
An extremely simplified reproduction is in [this GitHub repository](https://github.com/ItielBeeri/ng9-kendo-grid-problem).
1. Clone that repository
2. Run `yarn`
3. Run `ng s` and navigate to `/list` route

### Actual behavior
The above error is thrown and nothing rendered.

### Expected behavior
No errors. Since the grid isn't fed with data, the string `No records available.` should be rendered.
You can see this expected behavior just by disabling Ivy.

## 🔥 Exception or Error
<pre><code>
ERROR TypeError: el.setAttribute is not a function
    at DefaultDomRenderer2.setAttribute (platform-browser.js:1151)
    at elementAttributeInternal (core.js:12989)
    at Module.ɵɵattribute (core.js:19843)
    at SelectAllCheckboxDirective_HostBindings (index.js:17657)
    at setHostBindingsByExecutingExpandoInstructions (core.js:11453)
    at refreshView (core.js:11820)
    at refreshComponent (core.js:13213)
    at refreshChildComponents (core.js:11504)
    at refreshView (core.js:11825)
    at refreshDynamicEmbeddedViews (core.js:13138)
</code></pre>

## 🌍  Your Environment

**Angular Version:**
<pre><code>
Angular CLI: 9.0.5
Node: 12.14.0
OS: win32 x64

Angular: 9.0.5
Ivy Workspace: Yes

@angular-devkit/architect         0.900.5
@angular-devkit/build-angular     0.900.5
@angular-devkit/build-optimizer   0.900.5
@angular-devkit/build-webpack     0.900.5
@angular-devkit/core              9.0.5
@angular-devkit/schematics        9.0.5
@ngtools/webpack                  9.0.5
@schematics/angular               9.0.5
@schematics/update                0.900.5
</code></pre>

atscott:
This PR provides a more helpful error than the one currently present:
`el.setAttribute is not a function`. It is not valid to have directives with host bindings
on `ng-template` or `ng-container` nodes. VE would silently ignore this, while Ivy
attempts to set the attribute and throws an error because these are comment nodes
and do not have `setAttribute` functionality.

It is better to throw a helpful error than to silently ignore this because
putting a directive with host binding on an `ng-template` or `ng-container` is most often a mistake.
Developers should be made aware that the host binding will have no effect in these cases.

Note that an error is already thrown in Ivy, as mentioned above, so this
is not a breaking change and can be merged to both master and patch.

Resolves #35994

### DIFF
--- a/packages/core/src/render3/node_assert.ts
+++ b/packages/core/src/render3/node_assert.ts
@@ -34,6 +34,15 @@ export function assertNodeOfPossibleTypes(tNode: TNode|null, ...types: TNodeType
       `Should be one of ${types.map(typeName).join(', ')} but got ${typeName(tNode.type)}`);
 }
 
+export function assertNodeNotOfTypes(tNode: TNode, types: TNodeType[], message?: string): void {
+  assertDefined(tNode, 'should be called with a TNode');
+  const found = types.some(type => tNode.type === type);
+  assertEqual(
+      found, false,
+      message ??
+          `Should not be one of ${types.map(typeName).join(', ')} but got ${typeName(tNode.type)}`);
+}
+
 function typeName(type: TNodeType): string {
   if (type == TNodeType.Projection) return 'Projection';
   if (type == TNodeType.Container) return 'Container';
--- a/packages/core/test/acceptance/host_binding_spec.ts
+++ b/packages/core/test/acceptance/host_binding_spec.ts
@@ -1427,4 +1427,53 @@ describe('host bindings', () => {
         '<img src="unsafe:javascript:alert(3)">', bypassSanitizationTrustHtml,
         /* isAttribute */ false);
   });
+
+  onlyInIvy('VE would silently ignore this').describe('host binding on containers', () => {
+    @Directive({selector: '[staticHostAtt]', host: {'static': 'attr'}})
+    class StaticHostAttr {
+      constructor() {}
+    }
+
+    @Directive({selector: '[dynamicHostAtt]', host: {'[attr.dynamic]': '"dynamic"'}})
+    class DynamicHostAttr {
+      constructor() {}
+    }
+
+    it('should fail with expected error with ng-container', () => {
+      @Component({
+        selector: 'my-app',
+        template: `
+          <ng-template #ref></ng-template>
+          <ng-container [ngTemplateOutlet]="ref" staticHostAtt dynamicHostAtt></ng-container>
+        `
+      })
+      class App {
+      }
+
+      const comp =
+          TestBed.configureTestingModule({declarations: [App, StaticHostAttr, DynamicHostAttr]})
+              .createComponent(App);
+      // TODO(FW-2202): binding static attrs won't throw an error. We should be more consistent.
+      expect(() => comp.detectChanges())
+          .toThrowError(
+              /Attempted to set attribute `dynamic` on a container node. Host bindings are not valid on ng-container or ng-template./);
+    });
+
+    it('should fail with expected error with ng-template', () => {
+      @Component({
+        selector: 'my-app',
+        template: ` <ng-template staticHostAtt dynamicHostAtt></ng-template> `
+      })
+      class App {
+      }
+
+      const comp =
+          TestBed.configureTestingModule({declarations: [App, StaticHostAttr, DynamicHostAttr]})
+              .createComponent(App);
+      // TODO(FW-2202): binding static attrs won't throw an error. We should be more consistent.
+      expect(() => comp.detectChanges())
+          .toThrowError(
+              /Attempted to set attribute `dynamic` on a container node. Host bindings are not valid on ng-container or ng-template./);
+    });
+  });
 });
--- a/goldens/size-tracking/integration-payloads.json
+++ b/goldens/size-tracking/integration-payloads.json
@@ -62,7 +62,7 @@
         "bundle": "TODO(i): we should define ngDevMode to false in Closure, but --define only works in the global scope.",
         "bundle": "TODO(i): (FW-2164) TS 3.9 new class shape seems to have broken Closure in big ways. The size went from 169991 to 252338",
         "bundle": "TODO(i): after removal of tsickle from ngc-wrapped / ng_package, we had to switch to SIMPLE optimizations which increased the size from 252338 to 1197869, see PR#37221 and PR#37317 for more info",
-        "bundle": 1197869
+        "bundle": 1198831
       }
     }
   }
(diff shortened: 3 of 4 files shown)


pkozlowski-opensource:
@atscott thnx for taking a stab at it! I must say that I was on a fence regarding throwing vs. silently ignoring (as VE does). I think that your arguments in the PR description convince me so I'm fine with taking the direction you are proposing here.

Now, the change in this PR looks fine to me, after a short scan, but I think that we should cover more similar cases here (otherwise we will loose context over time). What I'm mostly talking about here are:
* host listeners
* property bindings (those would just nothing I guess)

WDYT?

atscott:
Hey @pkozlowski-opensource - That's a good point and I do think we should also take a look at host listeners and property bindings. One thing that we need to consider with those is that if there isn't already an error being thrown, it would be a breaking change to throw a new error. If that's the case, I would want this to be merged separately and we can put something in the backlog for a v11 follow-up.

pkozlowski-opensource:
> One thing that we need to consider with those is that if there isn't already an error being thrown, it would be a breaking change to throw a new error.

Totally! I _think_ that the error will be thrown in case of listeners (as you need to do `comment. addEventListener()` which I believe will fail. For a property binding I think it will be silent (no error) as the operation is `comment[prop] = val`. 

How about adding tests for the 2 above cases and seeing what happens?

pkozlowski-opensource (changes requested):
Marking as "Request changes" based on https://github.com/angular/angular/pull/37111#issuecomment-628757168

packages/core/src/render3/node_assert.ts
```diff
@@ -34,6 +34,16 @@ export function assertNodeOfPossibleTypes(tNode: TNode, ...types: TNodeType[]):
       `Should be one of ${types.map(typeName).join(', ')} but got ${typeName(tNode.type)}`);
 }
 
+export function assertNodeNotOfTypes(tNode: TNode, types: TNodeType[], message?: string): void {
+  assertDefined(tNode, 'should be called with a TNode');
+  const found = types.some(type => tNode.type === type);
+  assertEqual(
+      found, false,
+      message ?
```
AndrewKushnir, line 42: We can probably use nullish coalescing operator here?
```
assertEqual(
      found, false,
      message ?? `Should not be one of ${types.map(typeName).join(', ')} but got ${typeName(tNode.type)}`);
```

atscott:
[Ivy presubmit](https://test.corp.google.com/ui#id=OCL:311550660:BASE:311550681:1589474935895:3369f186)

With regards to property bindings, I think this is already an error of the form `Can't bind to 'X' since it isn't a known property of 'ng-template'`.

For host listeners, the runtime doesn't currently differentiate between `HostListener` and a listener added in the template (or even synthetic listeners, for that matter). All listeners share the `listenerInternal` implementation, which allows `TNodeType` `Element`, `Container`, and `ElementContainer`
https://github.com/angular/angular/blob/b2a332525821709d187d63a20e3055e9f2718718/packages/core/src/render3/instructions/listener.ts#L129-L131
It is valid to put a listener on an `ng-template` or `ng-container` for the purposes of listening to directive outputs https://github.com/angular/angular/blob/b2a332525821709d187d63a20e3055e9f2718718/packages/core/test/acceptance/directive_spec.ts#L518-L558
It may be useful to warn (or error) when the tNode type is `Container|ElementContainer` and`tNode.outputs === null || tNode.outputs[eventName] === undefined`. This would mean there was a listener put on a container and there is no matching directive output. Maybe this should be a  follow up PR?

AndrewKushnir (approved):
LGTM 👍 +1 to take care of other use-cases as the next step (and a separate PR) as Andrew suggested in https://github.com/angular/angular/pull/37111#issuecomment-628804910.

packages/core/test/acceptance/host_binding_spec.ts
```diff
@@ -1427,4 +1427,39 @@ describe('host bindings', () => {
...
+          <ng-container [ngTemplateOutlet]="ref" hostAtt></ng-container>
+        `
+      })
+      class App {
+      }
+
+      const comp =
+          TestBed.configureTestingModule({declarations: [App, HostAttr]}).createComponent(App);
+      // Ensure that we get an expected error, i.e. that we had an `assert` that failed.
+      expect(() => comp.detectChanges()).toThrowError(/ASSERTION ERROR.*/);
```
pkozlowski-opensource, line 1451: nit: I would love to see an assert on the actual error message here.

packages/core/test/acceptance/host_binding_spec.ts
```diff
@@ -1427,4 +1427,39 @@ describe('host bindings', () => {
...
+
+    it('should fail with expected error with ng-template', () => {
+      @Component({selector: 'my-app', template: ` <ng-template hostAtt></ng-template> `})
+      class App {
+      }
+
+      const comp =
+          TestBed.configureTestingModule({declarations: [App, HostAttr]}).createComponent(App);
+      // Ensure that we get an expected error, i.e. that we had an `assert` that failed.
+      expect(() => comp.detectChanges()).toThrowError(/ASSERTION ERROR.*/);
```
pkozlowski-opensource, line 1462: same as above, would love to see assert on the error message

packages/core/test/acceptance/host_binding_spec.ts
```diff
@@ -1427,4 +1427,39 @@ describe('host bindings', () => {
         '<img src="unsafe:javascript:alert(3)">', bypassSanitizationTrustHtml,
         /* isAttribute */ false);
   });
+
+  onlyInIvy('VE would silently ignore this').describe('host binding on containers', () => {
+    @Directive({selector: '[hostAtt]', host: {'static': 'attr', '[attr.dynamic]': '"dynamic"'}})
```
pkozlowski-opensource, line 1432: Could we have 2 separate directives (one with a static and another with a bound attribute)? As things stand today it is not clear if the failure occurs because of the static attr or a bound attr (especially that we don't assert on error messages - see my other comment).

pkozlowski-opensource (commented):
@atscott I did another pass on this PR today and left a couple of nit comments (see inline). 

I'm fine if we take next steps in a separate PR as long as we've got a GitHub / JIRA issue to track the remaining work. As things stand today we are very inconsistent between attrs / props and ivy / VE. Here is a stackblitz illustrating what happens with properties: https://stackblitz.com/edit/angular-wmck8s?file=src/app/app.module.ts

2 things to notice:
* `<ng-template>` and `<ng-container>` behave differently (!)
* the behaviour is different from VE (it systematically throws).

Personally I believe that we should unify what happens for attars and props as otherwise things are super - confusing for our users and I believe that it would be better to do all in this PR as we would have all the context / use-cases in front of us.

WDYT?

pkozlowski-opensource (approved):
The change LGTM but CI is read, so it needs cleanup

IgorMinar (approved):
LGTM 

Reviewed-for: size-tracking

jelbourn (approved):
LgTM

atscott:
merge assistance: pullapprove seems to be all kinds of messed up

Prashanth-P-Prabhu:
Is this change available in release 9.1.11?

atscott:
Hi @Prashanth-P-Prabhu - this change was merged to the `10.0.x` branch and but not `9.1.x`. Because it's only an improvement on the error being reported and it's not a bug fix, we didn't think it necessary to also cherry pick it to `9.1`.

PR #37117, opened 2020-05-14
refactor(core): rename refreshDynamicEmbeddedViews to refreshEmbedded…

atscott:
…Views

Dynamic embedded views were conceptually different from inline embedded views, but we have since
removed the inline embedded views so we now only have "embedded views".
See related refactoring work to remove inline embedded views in #34715
and #37073.


pkozlowski-opensource (approved):
LGTM, good move!

AndrewKushnir (approved):
LGTM, thanks for the refactoring @atscott 👍

PR #37121, opened 2020-05-14
docs(changelog): add breaking change note for ModuleWithProviders

kara:
When I made the generic mandatory for ModuleWithProviders
in 20cc3ab, I forgot to add a formal "breaking changes" note
for the CHANGELOG. This commit adds the missing breaking
changes information.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,16 @@
 * **language-service:** [ivy] wrap ngtsc to handle typecheck files ([#36930](https://github.com/angular/angular/issues/36930)) ([1142c37](https://github.com/angular/angular/commit/1142c37))
 * **core** make generic mandatory for ModuleWithProviders ([#36892](https://github.com/angular/angular/issues/36892)) ([20cc3ab](https://github.com/angular/angular/commit/20cc3ab))
 
+### BREAKING CHANGES
+
+* **core:** make generic mandatory for ModuleWithProviders 
+
+A generic has always been required for the `ModuleWithProviders` pattern to work with Ivy, but prior to this commit, View Engine allowed the generic to be omitted (though support was officially deprecated).
+If you're using `ModuleWithProviders` without a generic in your application code, a v10 migration will update your code for you. 
+
+However, you may run into issues if you're using View Engine and also depending on a library that omits the generic. 
+In this case, ngcc won't help you (because it's Ivy-only) and the migration only covers application code.
+As a workaround, we suggest setting `skipLibChecks` to false in your tsconfig or updating your app to use Ivy.
 
 
 <a name="9.1.7"></a>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,11 +35,18 @@
 
 * **core:** make generic mandatory for ModuleWithProviders 
 
-A generic has always been required for the `ModuleWithProviders` pattern to work with Ivy, but prior to this commit, View Engine allowed the generic to be omitted (though support was officially deprecated).
-If you're using `ModuleWithProviders` without a generic in your application code, a v10 migration will update your code for you. 
+A generic type parameter has always been required for the `ModuleWithProviders` pattern to work with Ivy, but prior to this commit, View Engine allowed the generic type to be omitted (though support was officially deprecated).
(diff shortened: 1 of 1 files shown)


filipesilva (approved):

CHANGELOG.md
```diff
@@ -31,6 +31,16 @@
 * **language-service:** [ivy] wrap ngtsc to handle typecheck files ([#36930](https://github.com/angular/angular/issues/36930)) ([1142c37](https://github.com/angular/angular/commit/1142c37))
 * **core** make generic mandatory for ModuleWithProviders ([#36892](https://github.com/angular/angular/issues/36892)) ([20cc3ab](https://github.com/angular/angular/commit/20cc3ab))
 
+### BREAKING CHANGES
+
+* **core:** make generic mandatory for ModuleWithProviders
```
petebacondarwin, line 36: The word generic in this title and the message below doesn't seem right to me. The class itself is the generic and we are not making the class mandatory in this change.

I worry that developers will not pick up on what it is implying so easily. Could we use the phrase `type parameter` or `type argument` instead?

```suggestion
* **core:** make type parameter mandatory for ModuleWithProviders 
```

CHANGELOG.md
```diff
@@ -31,6 +31,16 @@
...
 * **core** make generic mandatory for ModuleWithProviders ([#36892](https://github.com/angular/angular/issues/36892)) ([20cc3ab](https://github.com/angular/angular/commit/20cc3ab))
 
+### BREAKING CHANGES
+
+* **core:** make generic mandatory for ModuleWithProviders 
+
+A generic has always been required for the `ModuleWithProviders` pattern to work with Ivy, but prior to this commit, View Engine allowed the generic to be omitted (though support was officially deprecated).
+If you're using `ModuleWithProviders` without a generic in your application code, a v10 migration will update your code for you. 
+
+However, you may run into issues if you're using View Engine and also depending on a library that omits the generic.
```
petebacondarwin, line 41: I think it would be valuable to spell out the error message that people will see so that can quickly identify that this BC is the cause and fix it.

```suggestion
However, if you are using View Engine and also depending on a library that omits the generic, you will now get a build time error similar to:

`error TS2314: Generic type 'ModuleWithProviders<T>' requires 1 type argument(s).`
```

CHANGELOG.md
```diff
@@ -31,6 +31,16 @@
...
+### BREAKING CHANGES
+
+* **core:** make generic mandatory for ModuleWithProviders 
+
+A generic has always been required for the `ModuleWithProviders` pattern to work with Ivy, but prior to this commit, View Engine allowed the generic to be omitted (though support was officially deprecated).
+If you're using `ModuleWithProviders` without a generic in your application code, a v10 migration will update your code for you. 
+
+However, you may run into issues if you're using View Engine and also depending on a library that omits the generic. 
+In this case, ngcc won't help you (because it's Ivy-only) and the migration only covers application code.
+As a workaround, we suggest setting `skipLibChecks` to false in your tsconfig or updating your app to use Ivy.
```
petebacondarwin, line 43:
```suggestion
You should contact the library author to fix their library to provide a type parameter when they use this class.

As a workaround, we suggest setting `"skipLibChecks": false` in your tsconfig to instruct TypeScript to ignore this error, or updating your application to use Ivy, in which case ngcc will add the type parameter as necessary.
```

petebacondarwin (changes requested):
I think we could make it even easier for developers to match this breaking change to their errors and understand the consequences of the proposed workarounds.

CHANGELOG.md
kara: I think it might be confusing that the commit message no longer matches the "breaking change" headline. We can elaborate in the explanation, but not in the headline.

CHANGELOG.md
kara: Ah, this is much better. Good call!

kara:
@petebacondarwin Thanks for the review! Made your requested changes here:

https://github.com/angular/angular/pull/37121/commits/33c290b4dfca8b33657a75856f3d155c05b09e20

petebacondarwin (approved):
Perfect!

Issue #36415, opened 2020-04-03
Configurable module creation with IVY and AOT on


vdshb:
# 🐞 bug report

### Affected Package
The issue is caused by package @angular/core

### Is this a regression?
No

### Description
I want to create configurable Angular 9 module with IVY and AOT on.
In the simplest scenario it must provide a single stateful service configurable by name:
```typescript
@Injectable()
export class CounterService {
  private counter = 0;
  shot() {
    return this.counter++;
  }
}

@NgModule()
export class CounterModule {
  static withConfig(name: string): ModuleWithProviders {
    return {
      ngModule: CounterModule,
      providers: [{
        provide: name,
        useClass: CounterService
      }]
    };
  }
}
```
Everything works fine at this point. But if I want to add any logic into `CounterModule.withConfig`:
```typescript
@NgModule()
export class CounterModule {
  static withConfig(name?: string): ModuleWithProviders {
    const counterProviderToken = name ? name : CounterService;
    return {
      ngModule: CounterModule,
      providers: [{
        provide: counterProviderToken,
        useClass: CounterService
      }]
    };
  }
}
```
I'm getting a compilation error NG1010.

## 🔬 Minimal Reproduction
https://github.com/vdshb/ng-modules-problem

Please, check out the last commit. It adds code to get compilation exception.

## 🔥 Exception or Error
<pre><code>
error NG1010: Value at position 1 in the NgModule.imports of AppModule is not a reference: [object Object]
</code></pre>

## 🌍  Your Environment

**Angular Version:**
<pre><code>

Angular CLI: 9.1.0
Node: 13.12.0
OS: linux x64

Angular: 9.1.0
... animations, cli, common, compiler, compiler-cli, core, forms
... language-service, platform-browser, platform-browser-dynamic
... router
Ivy Workspace: Yes

Package                           Version
-----------------------------------------------------------
@angular-devkit/architect         0.901.0
@angular-devkit/build-angular     0.901.0
@angular-devkit/build-optimizer   0.901.0
@angular-devkit/build-webpack     0.901.0
@angular-devkit/core              9.1.0
@angular-devkit/schematics        9.1.0
@ngtools/webpack                  9.1.0
@schematics/angular               9.1.0
@schematics/update                0.901.0
rxjs                              6.5.4
typescript                        3.7.5
webpack                           4.42.0
</code></pre>

**Anything else relevant?**
I'm not sure if it's a bug or we have any other ways to make a configurable module. 

[...]

Issue #14410, opened 2017-02-10
ngc with AOT: support deprecations


ThomasBurleson:
Consider the situation where I want to deprecate the static `.forRoot()` call in a library module.
I want to warn the developers that the call will soon be deprecated.

```js
/**
 *
 */
@NgModule({
  declarations: ALL_DIRECTIVES,
  imports: [MediaQueriesModule],
  exports: [MediaQueriesModule, ...ALL_DIRECTIVES],
  providers: [ MediaMonitor ]
})
export class FlexLayoutModule {
  /** @deprecated */
  static forRoot(): ModuleWithProviders {
    console.warn('FlexLayoutModule.forRoot() has been deprecated and is no longer needed.');
    return {
      ngModule: FlexLayoutModule
    };
  }
}
```

Application developers using the `FlexLayoutModule.forRoot()` calls  and compiling with **ngc** and AOT,  will see the following error:

```console
Error encountered resolving symbol values statically. 
Calling function 'FlexLayoutModule', function calls are not supported. 
Consider replacing the function or lambda with a reference to an exported function, 
resolving symbol AppModule in /Users/thomasburleson/Desktop/ng2-aot-demo/builds/release/app.module.ts, 
resolving symbol AppModule in /Users/thomasburleson/Desktop/ng2-aot-demo/builds/release/app.module.ts
```

---- 

Discussions with @vicb:

The 'forRoot()' fn is never evaluated in AoT, ngc only analyzes the content of the fn to get the module and ngc does not know what to do with the first statetement (console.warn).

One possible solution would be to introduce a `@deprecated` annotation that would be understood by ngc

ThomasBurleson:
Ping @LucasSloan

trotyl:
> The 'forRoot()' fn is never evaluated in AoT, ngc only analyzes the content of the fn to get the module and ngc does not know what to do with the first statetement (console.warn).

Fixed by Ivy, any provider expression will be evaluated at runtime. So https://github.com/angular/angular/issues/11996 would be enough.

JoostK:
This restriction has been lifted with Ivy now that #37126 has been merged. As there's a more general feature request for dealing with deprecations in #11996, I'll go ahead and close this.

PR #37126, opened 2020-05-15
fix(compiler-cli): use ModuleWithProviders type if static eval fails

alxhub:
When the compiler encounters a function call within an NgModule imports
section, it attempts to resolve it to an NgModule-annotated class by
looking at the function body and evaluating the statements there. This
evaluation can only understand simple functions which have a single
return statement as their body. If the function the user writes is more
complex than that, the compiler won't be able to understand it and
previously the PartialEvaluator would return a "DynamicValue" for
that import.

With this change, in the event the function body resolution fails the
PartialEvaluator will now attempt to use its foreign function resolvers to
determine the correct result from the function's type signtaure instead. If
the function is annotated with a correct ModuleWithProviders type, the
compiler will be able to understand the import without static analysis of
the function body.

### DIFF
--- a/packages/compiler-cli/test/ngtsc/ngtsc_spec.ts
+++ b/packages/compiler-cli/test/ngtsc/ngtsc_spec.ts
@@ -2390,6 +2390,45 @@ runInEachFileSystem(os => {
     });
 
     describe('unwrapping ModuleWithProviders functions', () => {
+      it('should use a local ModuleWithProviders-annotated return type if a function is not statically analyzable',
+         () => {
+           env.write(`module.ts`, `
+            import {NgModule, ModuleWithProviders} from '@angular/core';
+
+            export function notStaticallyAnalyzable(): ModuleWithProviders<SomeModule> {
+              console.log('this interferes with static analysis');
+              return {
+                ngModule: SomeModule,
+                providers: [],
+              };
+            }
+
+            @NgModule()
+            export class SomeModule {}
+          `);
+
+           env.write('test.ts', `
+            import {NgModule} from '@angular/core';
+            import {notStaticallyAnalyzable} from './module';
+
+            @NgModule({
+              imports: [notStaticallyAnalyzable()]
+            })
+            export class TestModule {}
+          `);
+
+           env.driveMain();
+
+           const jsContents = env.getContents('test.js');
+           expect(jsContents).toContain('imports: [notStaticallyAnalyzable()]');
+
+           const dtsContents = env.getContents('test.d.ts');
+           expect(dtsContents).toContain(`import * as i1 from "./module";`);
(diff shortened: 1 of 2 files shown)


packages/compiler-cli/src/ngtsc/annotations/src/ng_module.ts
```diff
@@ -584,7 +584,7 @@ export class NgModuleDecoratorHandler implements
         throw new FatalDiagnosticError(
             ErrorCode.VALUE_HAS_WRONG_TYPE, expr,
             `Value at position ${idx} in the NgModule.${arrayName} of ${
-                className} is not a reference: ${entry}`);
+                className} is not a reference: ${entry!.constructor.name}`);
```
JoostK, line 587: This seems to me like a dangerous change to make, given that `entry` could legitimately be `null`/`undefined`.

packages/compiler-cli/test/ngtsc/ngtsc_spec.ts
```diff
@@ -2390,6 +2390,45 @@ runInEachFileSystem(os => {
     });
 
     describe('unwrapping ModuleWithProviders functions', () => {
+      it('should use a local ModuleWithProviders-annotated return type if a function is not statically analyzable',
+         () => {
+           env.write(`module.ts`, `
+            import {NgModule, ModuleWithProviders} from '@angular/core';
+
+            export function notStaticallyAnalyzable(): ModuleWithProviders<SomeModule> {
```
JoostK, line 2398: How about adding a test where the return type is missing. We can then assert on the insightful diagnostic that should be reported by the FFR (i.e. [here](https://github.com/angular/angular/blob/b07b7c9f3e653408fdedba258778fd4cc8003535/packages/compiler-cli/src/ngtsc/annotations/src/ng_module.ts#L507-L511)) which would address the poor UX that we had until this change.

JoostK (changes requested):
I believe this would address #36415 (FW-2048) and make #14410 obsolete.

packages/compiler-cli/test/ngtsc/ngtsc_spec.ts
alxhub: Oh, that's an interesting idea. Will do!

JoostK: It can still occur when someone leaves off the return type altogether.

JoostK (approved):

alxhub:
[Presubmit](http://test/OCL:312572515:BASE:312572747:1590015639100:10da24d1)

alxhub:
[Presubmit](http://test/OCL:312572515:BASE:314214782:1591050283518:7c59689f)

PR #37129, opened 2020-05-15
refactor: remove support for TypeScript 3.8

alan-agius4:
With this change we drop support for TypeScript 3.8 and remove all related tests.

BREAKING CHANGE:

TypeScript 3.8 is no longer supported, please update to TypeScript 3.9


alan-agius4:
Blocking until a release that supports both 3.8 and 3.9 is cut, since this is required to move CLI of TS 3.8.

aio/src/app/layout/doc-viewer/doc-viewer.component.ts
```diff
@@ -189,14 +189,14 @@ export class DocViewerComponent implements OnDestroy {
       const seconds = Number(cssValue.replace(/s$/, ''));
       return 1000 * seconds;
     };
+
+    type StringValueCSSStyleDeclaration
```
petebacondarwin, line 193: I think this would benefit from a comment explaining that some properties of `CSSStyleDeclaration` are not assignable and so we cannot animate them.

aio/src/app/layout/doc-viewer/doc-viewer.component.spec.ts
```diff
@@ -45,7 +45,7 @@ describe('DocViewerComponent', () => {
     let renderSpy: jasmine.Spy;
 
     const setCurrentDoc = (newDoc: TestParentComponent['currentDoc']) => {
-      parentComponent.currentDoc = newDoc && {id: 'fizz/buzz', ...newDoc};
+      parentComponent.currentDoc = newDoc;
```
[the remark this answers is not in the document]
petebacondarwin: I think what is needed here is

```suggestion
      parentComponent.currentDoc = newDoc && {...newDoc, id: 'fizz/buzz'};
```

petebacondarwin (changes requested):
Couple of suggestions.

aio/src/app/layout/doc-viewer/doc-viewer.component.spec.ts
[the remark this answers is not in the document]
alan-agius4: @petebacondarwin Seeing the usage, an `id` is always provided and so is a new object.

Example:
```ts
setCurrentDoc({contents: 'foo', id: 'bar'});
```

The suggested with alter the test because `fizz/buzz` will always override the id provided.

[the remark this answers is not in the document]
alan-agius4: I am not seeing any reason why the object should be cloned.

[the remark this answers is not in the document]
petebacondarwin: OK. Sorry you are correct.

petebacondarwin (approved):

IgorMinar (approved):
LGTM! Thank you! @alan-agius4 you rock!

kara:
@alan-agius4 Can you look at the CI failure?

.circleci/config.yml
```diff
@@ -32,8 +32,8 @@ var_4_win: &cache_key_win_fallback v7-angular-win-node-12-{{ checksum ".bazelver
 
 # Cache key for the `components-repo-unit-tests` job. **Note** when updating the SHA in the
 # cache keys also update the SHA for the "COMPONENTS_REPO_COMMIT" environment variable.
-var_5: &components_repo_unit_tests_cache_key v7-angular-components-448523bffffecd2b53a3d2854c3051b6b7a3934f
-var_6: &components_repo_unit_tests_cache_key_fallback v7-angular-components-
+var_5: &components_repo_unit_tests_cache_key v10-angular-components-448523bffffecd2b53a3d2854c3051b6b7a3934f
+var_6: &components_repo_unit_tests_cache_key_fallback v10-angular-components-
```
devversion, line 36: Can you provide context on why the cache key suffix has been increased? It looks like the cache key no longer matches with the SHA that you set the `COMPONENTS_REPO_COMMIT` variable to. That might be why you actually needed to bump the version.

IgorMinar: I think it was a mistake due to misunderstanding of how this worked - it's not clear how this key should be updated.

I'll push a fix with my attempt. Which also adds docs with my interpretation of how this works.

devversion: There might be a different reason why the cache version has been increased. I think the comment we already have is sufficient (but that might be just me). Having a concrete example definitely does not hurt though and it's better over-communicating 😄 

My suspicion though is that this won't fix the components-repo-unit-tests job as we still actually use TypeScript 3.8 in `master`. We could consider merging https://github.com/angular/components/pull/19336 in now, and then this PR can be updated.  We could do this to unblock this PR. 

Usually we'd have a cyclic dependency between the FW and components repo (e.g. components cannot update because it waits on FW; while FW cannot update until components updates). Fortunately though `disableTypeScriptVersionCheck` helps here and we could move forward with the PR in components without needing this change.

.circleci/config.yml
```diff
@@ -32,6 +32,11 @@ var_4_win: &cache_key_win_fallback v7-angular-win-node-12-{{ checksum ".bazelver
 
 # Cache key for the `components-repo-unit-tests` job. **Note** when updating the SHA in the
 # cache keys also update the SHA for the "COMPONENTS_REPO_COMMIT" environment variable.
+# The cache key should follow this format: v<X>-angular-components-<COMPONENTS_REPO_COMMIT>
+# where:
+#   - <X> is cache version number used in all of the cache keys above (example value: 7)
```
devversion, line 37: Nit for the the comment: The cache version does not necessarily need to match with the cache version from the cache keys above. The version can be bumped per cache (including its fallback) when a new/empty cache is desired (i.e. when the cache is poisoned or invalid for whatever reason)

IgorMinar: Hmmm... Are there scenarios under which we would want to rev X only for certain caches and not all of them at once?

How can we tell when this is desirable vs required? I definitely can't tell for sure. Do other people reving X know when to do it and which caches to invalidate?

Considering that we rev X very infrequently, isn't it simply safer to rev X across all caches "just to be safe" since correct cache invalidation is hard? I genuinely don't know how to rev X safely and correctly, so if I'm thinking about this incorrectly, can you please correct me and document how to do it? (the current docs in the config.yml are very light)

IgorMinar:
I rebased this PR and tried to use the latest green master SHA from the components repo. If that doesn't help, then we are likely dealing with some real incompatibility which will need to be investigated.

alan-agius4:
The failure is related to the emit helpers typescript bug https://github.com/microsoft/TypeScript/issues/38501

In components TS 3.9 PR, they get around this by using `importHelpers`. So we either need to wait for that PR to land, or cheer-pick the importHelpers change.

.circleci/config.yml
alan-agius4: Indeed, the failure is not related to the cache.

See: https://github.com/angular/angular/pull/37129#issuecomment-629595640

alan-agius4: And indeed I updated the cache key version because it was not clear. Thanks @IgorMinar the added docs rectify this 🙂

IgorMinar:
@alan-agius4  what would the cherry-pick look like? what is this `importHelpers` change? And also does this mean that we'll see similar issues in the Angular ecosystem?

alan-agius4:
The importHelpers change is this one: https://github.com/angular/components/pull/19336#discussion_r424391843

The emitted helpers causes a runtime error in UMDs, I don’t see any reason why this could impact the Angular eco system for a couple of reasons.

1. Applications should be using ES2015+ as a module and not UMD.
2. In Angular CLI we always configure workspaces to use TypeScript helpers (tslib) vis `importHelpers` compiler options, as opposed to inlined/emitted helpers to reduce the final bundle size. This is because otherwise typescript will inline the required helpers in each file that requires it instead of importing from tslib, which would result in multiple identical copies of the same helper in the final bundle.
3. When building the library in UMD, we do embed helpers, this is however done via Rollup and not TypeScript, for the same reason as above.

Side note: this issue will be solved in TS 3.9.3 via https://github.com/microsoft/TypeScript/pull/38502

.circleci/config.yml
devversion: In the past these cache versions distinguished all the time, but now they just seem to match by coincidence with other cache keys. Sometimes cache needs to be invalidated for the components repo job for example, as noted in https://github.com/angular/angular/commit/0e52d9297fc86f970321e7ecb9a800f16729c0df. For those cases, it's surely acceptable to just invalidate all other caches too, but I don't see why we'd want to do that. Invalidating the dedicated components repo cache is sufficient.

I don't think people often change the cache keys often (mostly it's done by dev-infra I'd guess), so it shouldn't be necessary unifying the version for all cache keys.

Either way though, I'm definitely also fine unifying this. Just mentioned this because the versions deviated in the past and the comment doesn't match this. If we want to unify this, then it might be good to have this documented for the other cache keys too.

IgorMinar:
Awesome. Thank you @alan-agius4!

However I'm still confused about why is the CI job failing and what do we need to do to make it green?

Waiting for 3.9.3 is one option, but is there another option that would allow us to merge this PR sooner?

.circleci/config.yml
IgorMinar: What I struggle with is that we don't really know when to invalidate these keys. Usually we need to do it because our cache got poisoned which is a really really bad thing to ever happen and in such case it might be prudent to just reset all the caches. Am I thinking about it incorrectly?

Do we have a hope of getting to a state where we have our deps under control and don't need to worry about cache poisoning at all?

Would completing the Bazel migration get us there?

alan-agius4:
@IgorMinar, the CI is failing because we are cloning the material repo and run a number of tests. But their testing setup in master doesn’t yet work with TS3.9 and exposes the TS helper bug, which was worked-around by using `importHelpers` https://github.com/angular/components/pull/19336/files#diff-58385c517f2bdbfc43246daac79ba239R7.  The previous SHA was using TypeScript 3.8, but we can now this can longer use it, because the FW doesn't support TS 3.8 with this PR.

The components team seem to be holding off their TS 3.9 PR (https://github.com/angular/components/pull/19336) because there is no Angular versions that supports TS 3.9 on NPM yet. Following a convo with @kara, she mentioned that a release that supports both TS3.8 and TS 3.9 will be cut on Monday.

Fortunately, there are a number of options that we can go with to merged without a TS release.

Option 1:
- Monday FW releases a version that supports 3.9
- Update https://github.com/angular/components/pull/19336 with the latest FW next version.
- When merged update SHA which should make this PR ready to get merged.
- Try to do all of the above before Wednesday morning PST, so this PR is in Wednesday release.

Option 2:
- Merge the https://github.com/angular/components/pull/19336 with `disableTypeScriptVersionCheck`.
- Update tests SHA
- Merge this PR after Monday’s release/Tuesday morning.
- Follow up with a PR in the components repo to remove `disableTypeScriptVersionCheck`

Option 3
- Create a PR that changes the `importHelpers` option in the material repo. (I am assuming that the tests will past after this change and there are further changes required in the material repo)
- Once merged update tests SHA
- Merge this PR after Monday’s release/Tuesday morning.

Option 4
- Worst case as suggested by @devversion, temporary disable these tests.

Option 5
- run the tests Material tests against https://github.com/angular/components/pull/19336 Instead of master
- Follow up with another PR to test against master when merged.

**Note**: whichever option we go for we always need this to land after the release on Monday, because to finish TS 3.9 support in the CLI we need a release a version that supports both because the current version of NGCC on NPM doesn't work with TS 3.9, thus the CLI TS 3.9 CI is red.

In a nutshell the problem here is, FW release is blocking CLI and Components from updating to TS 3.9. And FW is blocked from removing TS 3.8 support because the components repo doesn't yet work on TS 3.9. Moving everything into one repo is the ultimate goal, and would resolve issues like this.

aio/src/app/layout/doc-viewer/doc-viewer.component.spec.ts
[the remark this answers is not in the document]
gkalpak: In case anyone is wondering, here is where it changed to the current implementation: https://github.com/angular/angular/commit/f45c43188fa48d3655fb95405286b114e426b51d#diff-c28d3a9f68016393b299ddc959f60d41R48
(Closing the object is indeed not necessary now.)

aio/src/app/layout/doc-viewer/doc-viewer.component.ts
```diff
@@ -189,14 +189,15 @@ export class DocViewerComponent implements OnDestroy {
       const seconds = Number(cssValue.replace(/s$/, ''));
       return 1000 * seconds;
     };
+
+    // Some properties are not assignable and thus cannot be animated.
```
gkalpak, line 193: Maybe add an example or two 🙏 😁
For reference, the types that we want to exclude are:
- `number`, `"length"`, `"parentRule"` (readonly and CSS properties anyways)
- `"getPropertyPriority"`, `"getPropertyValue"`, `"item"`, `"removeProperty"`, `"setProperty"` (methods)

aio/src/app/layout/doc-viewer/doc-viewer.component.ts
```diff
@@ -189,14 +189,15 @@ export class DocViewerComponent implements OnDestroy {
...
+    type StringValueCSSStyleDeclaration
+      = { [K in keyof CSSStyleDeclaration]: CSSStyleDeclaration[K] extends string ? K : never }[keyof CSSStyleDeclaration];
```
gkalpak, line 195: Nice! You could also exclude `number`:

```suggestion
      = Exclude<{ [K in keyof CSSStyleDeclaration]: CSSStyleDeclaration[K] extends string ? K : never }[keyof CSSStyleDeclaration], number>;
```

gkalpak (approved):

IgorMinar:
Aha! I see. I made in incorrect assumption that components repo was already on 3.9 because I saw them merge the PR but I see that that PR didn't work so we are fixing that but now that is blocked on FW.

I suggest we:
- disable the test + release FW
- merge the Components PR + release
- reenable the FW test
- make all the changes we need in the CLI repo.

.circleci/config.yml
devversion: I think we haven't really had any cache poisoning cases in the past, but it's possible. The only thing that surfaced a couple of times is that we restore incompatible caches due to the fallback cache. This happened separately for individual caches. e.g.

* (For the standard CI cache): A restored cache contained old packages that caused module resolution problems because Yarn did not check all restored files. I created DEV-51 for this in the past.
* (For the components repo CI cache): A restored cache is based on a separate branch, or contains invalid retained postinstall patches. In those cases, the job will fail on postinstall, and the cache version would need to be incremented. This happened in https://github.com/angular/angular/commit/0e52d9297fc86f970321e7ecb9a800f16729c0df IIRC.

These were just a few examples. In my experience, we can certainly tell what cache is affected. In such cases, it would be unnecessary wiping all other caches upstream and on forks as we are clear on what cache is affected. As said, we could surely wipe all other caches, but I don't think it's worth in favor of a simpler model here. We are not wiping the caches without understanding the problem (i hope) and in those cases, we should have understanding which caches are affected.

We could potentially solve the concrete issues mentioned above, but there might be more and this just highlighted that in the common case, only individual caches are affected (judging on what happened in the past)

PR #37132, opened 2020-05-15
build(docs-infra): remove `{@searchKeywords}` from content when rende…

petebacondarwin:
…ring

Previously this inline-tag-def was returning the `doc` which is rendered
to the output as `[Object Object]` - obviously not what is intended.

Now it returns `''` which effectively strips the tag handler from the
rendered output.


gkalpak (approved):
Good catch! 😅

PR #37138, opened 2020-05-15
feat(dev-infra): add script for merging pull requests

devversion:
Introduces a new merge script that can be shared across all Angular repositories.
The components repo has been using this for a while now, and we plan on
using this for framework too.

The merge script helps with:

* cherry-picking into the appropriate branches (based on target labels)
* ensuring CI checks are all green
* ensuring CLA has been signed
* ensuring PR is rebased on a given commit (e.g. after code owner approval)
* ensuring PR is actually merge ready
* allows commit message modifications if needed

Another big goal is that the script recovers properly if merging fails.
Based on feedback, and inspection of the current merge script in framework,
it's easy to run into a state where the local Git repo needs to be cleaned up with
various commands (like `cherry-pick --abort). The script is flexible enough so that we
can add new features if desired. It can be also eventually used for merge automation.


.ng-dev-config.ts
```diff
@@ -72,8 +83,64 @@ const format = {
...
+    const matches = branchName.match(/refs\/heads\/([0-9]+\.[0-9]+\.x)/);
+    if (matches !== null) {
+      return matches[1];
+    }
+  }
+
+  throw Error('Could not determine patch branch name.');
+};
+
+const merge = () => {
```
josephperrott, line 109: Can you add a doc string here expressing what the merge config is used for?

.ng-dev-config.ts
```diff
@@ -1,3 +1,14 @@
+import {exec} from 'shelljs';
+
+import {MergeConfig} from './dev-infra/pr/merge/config';
```
josephperrott, line 3: I know that this is how we have to do this, but I really don't like that we are doing a relative path import for this.

dev-infra/pr/merge/config.ts
```diff
@@ -0,0 +1,114 @@
...
+
+import {getConfig, NgDevConfig} from '../../utils/config';
+
+import {GithubApiMergeStrategyConfig} from './strategies/api-merge';
+
+/**
+ * Possible merge methods supported by the Github API.
+ * https://developer.github.com/v3/pulls/#merge-a-pull-request-merge-button.
+ */
+export type GithubApiMergeMethod = 'merge'|'squash'|'rebase';
```
josephperrott, line 17: Can we not include `merge` here since we, as a "policy", don't allow merge commits in our repos?

dev-infra/pr/merge/config.ts
```diff
@@ -0,0 +1,114 @@
...
+   * Can also be wrapped in a function that accepts the target branch specified in the
+   * Github Web UI. This is useful for supporting labels like `target: development-branch`.
+   */
+  branches: string[]|((githubTargetBranch: string) => string[]);
+}
+
+/** Configuration for the merge script. */
+export interface MergeConfig {
+  /** Configuration for the upstream repository. */
+  repository: {user: string; name: string; useSsh?: boolean};
```
josephperrott, line 37: Do we still want to support `useSsh`?

dev-infra/pr/merge/console.ts
```diff
@@ -0,0 +1,20 @@
+/**
```
josephperrott, line 1: Should this be moved into `utils/`?  It seems like a common task to prompt for confirmation like this.  I know that we could have immediate usage in the `rebase-pr` tool

dev-infra/pr/merge/git.ts
```diff
@@ -0,0 +1,115 @@
...
+    return this.run(['rev-parse', '--abbrev-ref', 'HEAD']).stdout.trim();
+  }
+
+  /** Gets whether the current Git repository has uncommitted changes. */
+  hasUncommittedChanges(): boolean {
+    return this.runGraceful(['diff-index', '--quiet', 'HEAD']).status !== 0;
+  }
+
+  /** Sanitizes a given message by omitting the provided Github token if present. */
+  omitGithubTokenFromMessage(value: string): string {
```
josephperrott, line 112: should this be `private`?

dev-infra/pr/merge/config.ts
devversion: I'd prefer keeping as I'd like to tool to reflect configuration options available in Github, but we can always re-add this if needed.

dev-infra/pr/merge/console.ts
devversion: yeah, I didn't think of that. Makes sense to me.

dev-infra/pr/merge/git.ts
devversion: It's intentionally made public, so that we can omit the token in other places when needed. Not really used yet, but part conceptually should be public as part of the merge script architecture.

Happy to change if you think it would make sense.

.ng-dev-config.ts
devversion: Yeah, looks like there is no better way for now. It's not too bad given that we are in the repo where the dev-infra sources live. In the components repo that would be the NPM import.

dev-infra/pr/merge/config.ts
devversion: I'd prefer keeping as it's low-effort to keep and it's unclear what authentication teams prefer. I think we both are set on the personal access token that is required anyway for the API requests.

dev-infra/pr/merge/config.ts
josephperrott: okay, thats fine with me

dev-infra/pr/merge/config.ts
```diff
@@ -0,0 +1,136 @@
...
+    return {errors};
+  }
+
+  if (mergeConfig.remote) {
+    mergeConfig.remote = {...config.github, ...mergeConfig.remote};
+  } else {
+    mergeConfig.remote = config.github;
+  }
+
+  // We ensures that the `remote` is set, so we can cast it to a `MergeConfigWithRemote`.
```
josephperrott, line 114:
```suggestion
  // Ensure that the `remote` is set, so we can cast it to a `MergeConfigWithRemote`.
```

josephperrott (approved):
LGTM

kara:
@devversion Ironically, this commit did not apply cleanly to the patch branch. Can you create a separate PR for patch?

devversion:
@kara I created https://github.com/angular/angular/pull/37184 for the patch-only version.

PR #37143, opened 2020-05-15
build: move CHANGELOG to public-api approval group

kara:
Previously, the CHANGELOG file was caught under the general
"*" glob, which meant that it fell under the dev-infra
approval group. This does not really make sense since it
has more to do with public-facing changes. As such, this
commit moves the CHANGELOG file specifically to the
public-api approval group.

### DIFF
--- a/.pullapprove.yml
+++ b/.pullapprove.yml
@@ -1030,6 +1030,7 @@ groups:
     reviewers:
       users:
         - IgorMinar
+        - kara
 
 
   # ================================================
--- a/.pullapprove.yml
+++ b/.pullapprove.yml
@@ -948,7 +948,7 @@ groups:
     conditions:
       - *can-be-global-approved
       - >
-        contains_any_globs(files, [
+        contains_any_globs(files.exclude("CHANGELOG.md"), [
           '*',
           '.circleci/**',
           '.devcontainer/**',
--- a/.pullapprove.yml
+++ b/.pullapprove.yml
@@ -1020,6 +1020,7 @@ groups:
       - >
         contains_any_globs(files, [
           'goldens/public-api/**',
+          'CHANGELOG.md',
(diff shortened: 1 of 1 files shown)


.pullapprove.yml
```diff
@@ -1020,6 +1020,7 @@ groups:
       - >
         contains_any_globs(files, [
           'goldens/public-api/**',
+          'CHANGELOG.md',
```
IgorMinar, line 1023: I believe that this change alone is not sufficient. we also need to exclude this file from the `dev-infra` group. otherwise we'll need both the `public-api` group and the `dev-infra` groups to approve any changes to the CHANGELOG.md file.

IgorMinar (changes requested):
What if we removed the `'*'` rule from `dev-infra` group and explicitly specified which of the top level files belong to `dev-infra` and which belong elsewhere?

It seems to me that top level `'*'` is a recipe for more confusion in the future. Being explicit will prevent that.

IgorMinar:
I can help make the change and push to this branch if you'd like.

.pullapprove.yml
kara: ohh good point.

IgorMinar:
I also think that we should add you to the `public-api` group even though we don't have staged reviews yet. That way you can help with approving changelog.md and other changes that are straightforward and don't need much discussion.

sidenote: I'd love for us to have a public-api "club" where we discuss public api changes and mentor more people on how to evolve the public api surface in a way that is backwards compatible, ergonomic, consistent, and enables incremental evolution. Would you be interested in that?

IgorMinar (approved):
lgtm, now. We can either make the remaining changes here or in a follow up PR. You pick.

IgorMinar (approved):
LGTM!

PR #37147, opened 2020-05-15
build: complete removal of gulp format

josephperrott:
Remove gulp format deprecation message to complete the removal of
formatting Typescript and Javascript files via gulp command.


IgorMinar (commented):
YAY!!! Finally!!!

Just like with bazel format PR - could we error with a pointer to `ng-dev` instead of removing these tasks. It will make for a much better DX when you get a descriptive error that says "you did X, now you need to do Y instead". Thanks!

filipesilva (approved):

josephperrott:
@IgorMinar My intention here is to remove the remaining references to when we performed this task using `gulp` as it has been ~4 weeks since we made the deprecation.

PR #37148, opened 2020-05-15
build: complete removal of bazel format yarn commands

josephperrott:
Remove bazel yarn format deprecation message to complete the removal of
formatting bazel related files via yarn command command.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,11 @@
     "url": "https://github.com/angular/angular.git"
   },
   "scripts": {
+    "/": "",
+    "// 1": "Many developer of our checks/scripts/tools have moved to our ng-dev tool",
+    "// 2": "Find the usage you are looking for with:",
+    "// 3": "yarn ng-dev --help",
+    "/ ": "",
     "preinstall": "node tools/yarn/check-yarn.js",
     "postinstall": "node scripts/webdriver-manager-update.js && node --preserve-symlinks --preserve-symlinks-main ./tools/postinstall-patches.js",
     "check-env": "gulp check-env",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -16,9 +16,6 @@
     "url": "https://github.com/angular/angular.git"
   },
   "scripts": {
-    "bazel:format": "yarn -s ng-dev format deprecation-warning bazel:format",
-    "bazel:lint": "yarn -s ng-dev format deprecation-warning bazel:lint",
-    "bazel:lint-fix": "yarn -s ng-dev format deprecation-warning bazel:lint-fix",
```
IgorMinar, line 21: How about we keep these in here for a bit longer, but instead of displaying a warning, we error?

TBH I didn't catch that we should be now using `ng-dev format` instead of `yarn bazel:format` and if this got merged without me randomly opening this PR I wouldn't know what to do next time I need to reformat some files.

josephperrott: Currently, we already fail to perform the format and instead instruct the user to run the new command.

IgorMinar: you are right, I must have been lucky and not have a need to run the script because my IDE formatter does the job and I rarely need to reformat from the command line, but when I do I'd prefer to at least see a comment in package.json under scripts that reminds me to use `yarn ng-dev --help` instead.

josephperrott: Added general note in `"scripts"` about our increased migration to using `ng-dev `

IgorMinar: thank you!

IgorMinar (approved):
LGTM!

PR #37163, opened 2020-05-17
fix(router): remove parentheses for primary outlet segment

### LINKED ISSUE #24656: Parentheses remains in URL when child auxiliary route outlet path is set to null
## I'm submitting a...

<pre><code>
[x] Bug report  
</code></pre>

## Current behavior
From the following URL:
`/some/(primary/route//auxOutlet:someAuxRoute)`

When routed like this:
` this.router.navigate( ['/some', { outlets: { auxOutlet: null } }]);`

Parentheses remains in URL string:
`https://angular-kasves.stackblitz.io/some/(primary/route)`

## Expected behavior
URL string should be trimmed of parentheses like that:
`https://angular-kasves.stackblitz.io/some/primary/route`

## Demo
https://stackblitz.com/edit/angular-kasves
https://angular-kasves.stackblitz.io/some/primary/route
First, click "Aux On" button. Then click "Aux Off" button.

## Reproduction Steps
* Route to some child auxiliary route:
`this.router.navigate( ['/some', { outlets: { auxOutlet: ['someAuxRoute'] } }]);`
<br>URL will be something similar:
`https://angular-kasves.stackblitz.io/some/(primary/route//auxOutlet:someAuxRoute)`

* After that, set auxiliary router outlet path to null:
 `this.router.navigate( ['/some', { outlets: { auxOutlet: null } }]);`
<br>URL will still contain parentheses:
`https://angular-kasves.stackblitz.io/some/(primary/route)`

ggradnig:
For URLs that use auxiliary route outlets in the second or following path elements,
when removing the auxiliary route segment, parentheses remain for the primary outlet segment.
This causes the following error when trying to reload the URL: "Cannot match any route".
The commit adds a check for this scenario, serializing the URL as "a/b" instead of "a/(b)".

PR Close #24656

## PR Checklist

## PR Type

- [x] Bugfix

For the following URL:
`a/(b//right:c)`

When removing the auxiliary outlet with:
`this.router.navigate( ['a', { outlets: { right: null } }]);`

Parentheses remain in the URL:
`a/(b)`

Issue Number: #24656

## What is the new behavior?
Parentheses are not added if the only child is a primary outlet segment:
`a/b`

- [x] No

### DIFF
--- a/packages/router/src/url_tree.ts
+++ b/packages/router/src/url_tree.ts
@@ -341,6 +341,11 @@ function serializeSegment(segment: UrlSegmentGroup, root: boolean): string {
       return [`${k}:${serializeSegment(v, false)}`];
     });
 
+    // use no parenthesis if the only child is a primary outlet route
+    if (Object.keys(segment.children).length === 1 && segment.children[PRIMARY_OUTLET] != null) {
+      return `${serializePaths(segment)}/${children[0]}`;
+    }
+
     return `${serializePaths(segment)}/(${children.join('//')})`;
   }
 }
--- a/packages/router/test/create_url_tree.spec.ts
+++ b/packages/router/test/create_url_tree.spec.ts
@@ -128,6 +128,12 @@ describe('createUrlTree', () => {
     expect(serializer.serialize(t)).toEqual('/a');
   });
 
+  it('should support removing parenthesis for primary segment on second path element', () => {
+    const p = serializer.parse('/a/(b//right:c)');
+    const t = createRoot(p, ['a', {outlets: {right: null}}]);
+    expect(serializer.serialize(t)).toEqual('/a/b');
+  });
+
   it('should update matrix parameters', () => {
     const p = serializer.parse('/a;pp=11');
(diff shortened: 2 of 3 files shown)


atscott:
Hmm, I just ran a [global presubmit](https://test.corp.google.com/ui#id=OCL:316552241:BASE:316671091:1592343672246:19a6199d) (every test affected by this change in Google) and there was only one failure similar to the one in `packages/router/test/integration.spec.ts` and it has since been updated. Maybe we really _can_ submit this as a non-breaking bug fix.

atscott (approved):

atscott:
TL;DR I've marked this for merge into the master branch, targeting the 10.1 release.
I think we can consider this a more or less non-breaking based on the results of the Google tests. We _just_ missed the cutoff for the 10.0 release. Because there's still a tiny chance it can cause some tests to fail (such as the one that needed to change in this PR), I think it would be a good idea to only merge it to `10.1` rather than `10.0.x`.

atscott (approved):

ggradnig:
Sounds great :) Thank you for reviewing.

PR #37167, opened 2020-05-18
fix(core): properly move embedded views of dynamic component's projec…

### LINKED ISSUE #37120: In Ivy: ViewContainerRef.move causes ASSERTION ERROR
# 🐞 bug report

### Affected Package
    "@angular/animations": "~10.0.0-next.7",
    "@angular/common": "~10.0.0-next.7",
    "@angular/compiler": "~10.0.0-next.7",
    "@angular/core": "~10.0.0-next.7",
    "@angular/forms": "~10.0.0-next.7",
    "@angular/platform-browser": "~10.0.0-next.7",
    "@angular/platform-browser-dynamic": "~10.0.0-next.7",
    "@angular/router": "~10.0.0-next.7",

### Description

I have this  dynamic component called `HostViewAccessor`:
```ts
@Component({ 
  selector: 'host-view-accessor',
  template: `<ng-template #portal><ng-content></ng-content></ng-template>`
})
export class HostViewAccessor {
  @ViewChild('portal', { static: true, read: TemplateRef }) templateRef: TemplateRef<void>;
}
```

And I use HostViewAccessor component above in this `CellComponent`:
```ts
@Component({
  selector: 'cell',
  template: `<ng-content></ng-content>`
})
export class CellComponent  {
  view: EmbeddedViewRef<void>;  

  constructor(private el: ElementRef, private cfr: ComponentFactoryResolver) {}

  ngOnInit() {
    const dynamicHostViewAccessorFactory = this.cfr.resolveComponentFactory(HostViewAccessor);
    const dynamicHostViewAccessorCompRef = dynamicHostViewAccessorFactory.create(Injector.NULL, [[this.el.nativeElement]]);
    this.view = dynamicHostViewAccessorCompRef.instance.templateRef.createEmbeddedView();
  }
}
```
You can see that I am creating the embedded view of the `CellComponent` itself through `HostViewAccessor`.

And I project cell components inside this `RowComponent`. Then, I access and insert the views of the cell components into the adjacent `ng-container` with the template id of `#cellsContainer`: 
```ts
@Component({
  selector: 'row',
  template: `<ng-content></ng-content><ng-container #cellsContainer></ng-container>`
})
export class RowComponent  {
  @ViewChild('cellsContainer', { static: false, read: ViewContainerRef }) cellsContainer: ViewContainerRef;

  @ContentChildren(CellComponent) cells: QueryList<CellComponent>;

  ngAfterViewInit() {
    this.cells.forEach(cell => this.cellsContainer.insert(cell.view));
    setTimeout(() => {
      // swap the first and second cell views.
      this.cellsContainer.move(this.cellsContainer.get(1), 0);
    }, 3000);
  }
}
```

Also, you can see that after inserting the cell views, I try to swap the first and second cell views. However, that exact line causes the following error:

```
ERROR Error: ASSERTION ERROR: Should be one of Element, Container, ElementContainer, IcuContainer, Projection but got <unknown> [Expected=> true == false <=Actual]
```

## 🔬 Minimal Reproduction

Here is the minimal reproduction of the issue in a new app:
https://github.com/Shijir/ng10-testing

An interesting thing I've noticed here is that I couldn't reproduce the error on NG-RUN even though I've used the exact same code here:
https://ng-run.com/edit/MIwT9pp8UwanQonc8Qt3?open=app%2Fapp.component.ts&layout=1
However, this NG-RUN example is running on Typescript@3.7.5. So I suspect also environmental discrepancies might be causing the issues. 

## 🌍  Your Environment

**Angular Version:**
<pre><code>
~10.0.0-next.7
"typescript": "^3.8.3"

</code></pre>

**Anything else relevant?**

Shijir:
…table nodes

This commit fixes the issue of the ASSERTION ERROR issue when
a projected node(RNode) inside an array is checked against the types
of TNodeType.Element, TNodeType.Container, TNodeType.ElementContainer,
TNodeType.IcuContainer, TNodeType.Projection. As it's inside an array,
it doesn't fall into any of those types, as a result, it throws
the ASSERTION ERROR.

Fixes #37120

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Moving embedded views created out of dynamic component's projectable nodes causes ASSERTION ERROR.

Issue Number: https://github.com/angular/angular/issues/37120

## What is the new behavior?

Properly moves embedded views created out of dynamic component's projectable nodes causes ASSERTION ERROR.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/render3/node_manipulation.ts
+++ b/packages/core/src/render3/node_manipulation.ts
@@ -12,7 +12,8 @@ import {RendererStyleFlags2} from '../render/api_flags';
 import {addToArray, removeFromArray} from '../util/array_utils';
 import {assertDefined, assertDomNode, assertEqual, assertFunction, assertString} from '../util/assert';
 import {escapeCommentText} from '../util/dom';
-import {assertLContainer, assertLView, assertTNodeForLView} from './assert';
+
+import {assertLContainer, assertLView, assertParentView, assertProjectionSlots, assertTNodeForLView} from './assert';
 import {attachPatchData} from './context_discovery';
 import {icuContainerIterate} from './i18n/i18n_tree_shaking';
 import {CONTAINER_HEADER_OFFSET, HAS_TRANSPLANTED_VIEWS, LContainer, MOVED_VIEWS, NATIVE, unusedValueExportToPlacateAjd as unused1} from './interfaces/container';
@@ -772,20 +773,31 @@ function getFirstNativeNode(lView: LView, tNode: TNode|null): RNode|null {
       // If the ICU container has no nodes, than we use the ICU anchor as the node.
       return rNode || unwrapRNode(lView[tNode.index]);
     } else {
-      const componentView = lView[DECLARATION_COMPONENT_VIEW];
-      const componentHost = componentView[T_HOST] as TElementNode;
-      const parentView = getLViewParent(componentView);
-      const firstProjectedTNode: TNode|null =
-          (componentHost.projection as (TNode | null)[])[tNode.projection as number];
-
-      if (firstProjectedTNode != null) {
-        return getFirstNativeNode(parentView!, firstProjectedTNode);
+      const projectionNodes = getProjectionNodes(lView, tNode);
+      if (projectionNodes !== null) {
+        if (Array.isArray(projectionNodes)) {
+          return projectionNodes[0];
+        }
+        const parentView = getLViewParent(lView[DECLARATION_COMPONENT_VIEW]);
+        ngDevMode && assertParentView(parentView);
+        return getFirstNativeNode(parentView!, projectionNodes);
       } else {
         return getFirstNativeNode(lView, tNode.next);
       }
     }
   }
 
+  return null;
+}
+
+export function getProjectionNodes(lView: LView, tNode: TNode|null): TNode|RNode[]|null {
+  if (tNode !== null) {
+    const componentView = lView[DECLARATION_COMPONENT_VIEW];
+    const componentHost = componentView[T_HOST] as TElementNode;
+    const slotIdx = tNode.projection as number;
+    ngDevMode && assertProjectionSlots(lView);
+    return componentHost.projection![slotIdx];
+  }
   return null;
 }
 
--- a/packages/core/test/linker/integration_spec.ts
+++ b/packages/core/test/linker/integration_spec.ts
@@ -7,7 +7,7 @@
  */
 
 import {CommonModule, DOCUMENT, ɵgetDOM as getDOM} from '@angular/common';
-import {Compiler, ComponentFactory, ComponentRef, ErrorHandler, EventEmitter, Host, Inject, Injectable, InjectionToken, Injector, NgModule, NgModuleRef, NO_ERRORS_SCHEMA, OnDestroy, SkipSelf, ViewRef, ɵivyEnabled as ivyEnabled} from '@angular/core';
+import {Compiler, ComponentFactory, ComponentRef, ErrorHandler, EventEmitter, Host, Inject, Injectable, InjectionToken, Injector, NgModule, NgModuleRef, NO_ERRORS_SCHEMA, OnDestroy, SkipSelf, ViewChild, ViewRef, ɵivyEnabled as ivyEnabled} from '@angular/core';
 import {ChangeDetectionStrategy, ChangeDetectorRef, PipeTransform} from '@angular/core/src/change_detection/change_detection';
 import {getDebugContext} from '@angular/core/src/errors';
 import {ComponentFactoryResolver} from '@angular/core/src/linker/component_factory_resolver';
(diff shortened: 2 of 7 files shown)


packages/core/src/render3/node_manipulation.ts
```diff
@@ -724,11 +724,14 @@ function getFirstNativeNode(lView: LView, tNode: TNode|null): RNode|null {
       const componentView = lView[DECLARATION_COMPONENT_VIEW];
       const componentHost = componentView[T_HOST] as TElementNode;
       const parentView = getLViewParent(componentView);
-      const firstProjectedTNode: TNode|null =
+      const firstProjectedTNode: TNode|RNode[]|null =
           (componentHost.projection as (TNode | null)[])[tNode.projection as number];
```
pkozlowski-opensource, line 728: we should adjust typing here as well, no?

pkozlowski-opensource, line 728: We should adjust typing here as well, no? Didn't look too deeply but it should be something like `(TNode|RNode[])[]` if we assume that there are nodes. But it also can be null, no? So I believe that the correct type case should be `(TNode|RNode[])[]|null` - and we need to account for `null` (test).

In short: it looks like the current `as` type override is incorrect and this is what allowed the bug to slip in the first place!

packages/core/src/render3/node_manipulation.ts
```diff
@@ -724,11 +724,14 @@ function getFirstNativeNode(lView: LView, tNode: TNode|null): RNode|null {
...
-
       if (firstProjectedTNode != null) {
-        return getFirstNativeNode(parentView!, firstProjectedTNode);
+        if (Array.isArray(firstProjectedTNode)) {
+          return firstProjectedTNode[0];
```
pkozlowski-opensource, line 731: What would happen in the array here is empty? Or has an undefined value? We need tests for those cases.

pkozlowski-opensource (dismissed):
@Shijir thnx so much for looking into this fix and comprehensive tests! This PR goes in the right direction but I think that we need to tighten types and corner cases checks given that so many bugs are hiding in those areas! I've left a couple of in-line comments but more importantly - we start to have the same logic in multiple places (see my recent change in https://github.com/angular/angular/commit/e30e1325f31144396bd012dc9a38dbb4832712bf#diff-6214bef312f4279eb8c0a3f40aed7511R362-R377) and I think that maybe we should extract the logic of getting `TNode|RNode[]|null` into a common util function. 

WDYT?

packages/core/src/render3/node_manipulation.ts
Shijir: You are right.  That type override should have been `TNode|RNode[]|null`. As you've suggested, I've fixed that issue by creating a util function that incorporates the logic of getting `TNode|RNode[]|null`. Also, I've created a couple of assert functions for checking projection slots and parent veiws.

packages/core/src/render3/node_manipulation.ts
Shijir: Definitely! I've added a test for that case.
Looks like it does treat it just like an empty `ng-content` slot.

gnomeontherun:
@Shijir Can you update and rebase to latest? We need to keep this up to date to help them review it.

packages/core/src/render3/node_manipulation.ts
```diff
@@ -711,14 +711,15 @@ function getFirstNativeNode(lView: LView, tNode: TNode|null): RNode|null {
...
+      if (projectionNodes !== null) {
+        if (Array.isArray(projectionNodes)) { 
+          return projectionNodes[0];
+        } else {
+          const parentView = getLViewParent(lView[DECLARATION_COMPONENT_VIEW]);
+          ngDevMode && assertParentView(parentView)
+          return getFirstNativeNode(parentView!, projectionNodes);
+        }
       } else {
         return getFirstNativeNode(lView, tNode.next);
```
jessicajaniuk, line 724: You can eliminate some `else` statements here.
```suggestion
        }
        const parentView = getLViewParent(lView[DECLARATION_COMPONENT_VIEW]);
        ngDevMode && assertParentView(parentView)
        return getFirstNativeNode(parentView!, projectionNodes);
      }
      return getFirstNativeNode(lView, tNode.next);
```

packages/core/test/linker/integration_spec.ts
```diff
@@ -1641,6 +1641,162 @@ function declareTests(config?: {useJit: boolean}) {
...
+
+           fixture.detectChanges();
+
+           menuItemsContainer.move(menuItemsContainer.get(0)!, 1);
+           expect(fixture.nativeElement.textContent).toBe('B1bA0aC2c');
+           menuItemsContainer.move(menuItemsContainer.get(2)!, 1);
+           expect(fixture.nativeElement.textContent).toBe('B1bC2cA0a');
+         });
+
+         it('should support moving embedded views with empty projectable slots', () => {
```
jessicajaniuk, line 1771: I think this indentation is off.

jessicajaniuk (approved):
Just some minor formatting suggestions. You'll want to rebase again, but otherwise LGTM.

packages/core/test/linker/integration_spec.ts
Shijir: Fixed it.

packages/core/src/render3/node_manipulation.ts
Shijir: Changed as requested.

Shijir:
Hi @pkozlowski-opensource, could you take another look at the PR?

mhevery (approved):
reviewed-for: global-approvers

mhevery:
[presubmit](http://test/OCL:356601612:BASE:356601642:1612912105547:e18ee398)

PR #37186, opened 2020-05-18
Update API ref doc for platform browser package

jbogarthyde:
Edit descriptions, usage examples, and add links to be complete and consistent with API reference doc style

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

API reference doc for this package largely missing or incomplete.

Issue Number: N/A

## What is the new behavior?
Add and edit reference doc.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/application_ref.ts
+++ b/packages/core/src/application_ref.ts
@@ -156,7 +156,8 @@ export function createPlatform(injector: Injector): PlatformRef {
  * Creates a factory for a platform. Can be used to provide or override `Providers` specific to
  * your applciation's runtime needs, such as `PLATFORM_INITIALIZER` and `PLATFORM_ID`.
  * @param parentPlatformFactory Another platform factory to modify. Allows you to compose factories
- * to build up configurations that might be required by different libraries or parts of the application.
+ * to build up configurations that might be required by different libraries or parts of the
+ * application.
  * @param name Identifies the new platform factory.
  * @param providers A set of dependency providers for platforms created with the new factory.
  *
--- a/packages/core/src/application_ref.ts
+++ b/packages/core/src/application_ref.ts
@@ -135,7 +135,7 @@ export class NgProbeToken {
 
 /**
  * Creates a platform.
- * Platforms have to be eagerly created via this function.
+ * Platforms must be created on launch using this function.
  *
  * @publicApi
  */
@@ -153,7 +153,7 @@ export function createPlatform(injector: Injector): PlatformRef {
 }
 
 /**
- * Creates a factory for a platform
+ * Creates a factory for a platform.
  *
  * @publicApi
  */
@@ -182,7 +182,7 @@ export function createPlatformFactory(
 }
 
 /**
- * Checks that there currently is a platform which contains the given token as a provider.
+ * Checks that there is currently a platform that contains the given token as a provider.
  *
  * @publicApi
  */
@@ -202,7 +202,8 @@ export function assertPlatform(requiredToken: any): PlatformRef {
 }
 
 /**
- * Destroy the existing platform.
+ * Destroys the current Angular platform and all Angular applications on the page.
+ * Destroys all modules and listeners registered with the platform.
  *
  * @publicApi
  */
@@ -259,12 +260,11 @@ export interface BootstrapOptions {
 }
 
 /**
- * The Angular platform is the entry point for Angular on a web page. Each page
- * has exactly one platform, and services (such as reflection) which are common
+ * The Angular platform is the entry point for Angular on a web page.
+ * Each page has exactly one platform. Services (such as reflection) which are common
  * to every Angular application running on the page are bound in its scope.
- *
- * A page's platform is initialized implicitly when a platform is created via a platform factory
- * (e.g. {@link platformBrowser}), or explicitly by calling the {@link createPlatform} function.
+ * A page's platform is initialized implicitly when a platform is created using a platform
+ * factory such as `PlatformBrowser`, or explicitly by calling the `createPlatform()` function.
  *
  * @publicApi
  */
@@ -278,11 +278,11 @@ export class PlatformRef {
   constructor(private _injector: Injector) {}
 
   /**
-   * Creates an instance of an `@NgModule` for the given platform
-   * for offline compilation.
+   * Creates an instance of an `@NgModule` for the given platform for offline compilation.
    *
    * @usageNotes
-   * ### Simple Example
+   *
+   * The following example creates the NgModule for a browser platform.
    *
    * ```typescript
    * my_module.ts:
@@ -384,14 +384,14 @@ export class PlatformRef {
   }
 
   /**
-   * Register a listener to be called when the platform is disposed.
+   * Registers a listener to be called when the platform is destroyed.
    */
   onDestroy(callback: () => void): void {
     this._destroyListeners.push(callback);
   }
 
   /**
-   * Retrieve the platform {@link Injector}, which is the parent injector for
+   * Retrieves the platform {@link Injector}, which is the parent injector for
    * every Angular application on the page and provides singleton providers.
    */
   get injector(): Injector {
@@ -399,7 +399,8 @@ export class PlatformRef {
   }
 
   /**
-   * Destroy the Angular platform and all Angular applications on the page.
+   * Destroys the current Angular platform and all Angular applications on the page.
+   * Destroys all modules and listeners registered with the platform.
    */
   destroy() {
     if (this._destroyed) {
--- a/packages/core/src/application_ref.ts
+++ b/packages/core/src/application_ref.ts
@@ -153,7 +153,12 @@ export function createPlatform(injector: Injector): PlatformRef {
 }
 
 /**
- * Creates a factory for a platform.
+ * Creates a factory for a platform. Can be used to provide or override `Providers` specific to
+ * your applciation's runtime needs, such as `PLATFORM_INITIALIZER` and `PLATFORM_ID`.
+ * @param parentPlatformFactory Another platform factory to modify. Allows you to compose factories
+ * to build up configurations that might be required by different libraries or parts of the application.
+ * @param name Identifies the new platform factory.
+ * @param providers A set of dependency providers for platforms created with the new factory.
  *
  * @publicApi
  */
--- a/packages/platform-browser/src/browser/meta.ts
+++ b/packages/platform-browser/src/browser/meta.ts
@@ -10,7 +10,11 @@ import {DOCUMENT, ɵDomAdapter as DomAdapter, ɵgetDOM as getDOM} from '@angular
 import {Inject, Injectable, ɵɵinject} from '@angular/core';
 
 /**
- * Represents a meta element.
+ * Represents the attributes of an HTML `<meta>` element. The element itself is
+ * represented by the internal `HTMLMetaElement`.
+ *
+ * @see [HTML meta tag](https://developer.mozilla.org/docs/Web/HTML/Element/meta)
+ * @see `Meta`
  *
  * @publicApi
  */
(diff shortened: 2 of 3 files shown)


packages/platform-browser/src/browser/meta.ts
```diff
@@ -30,14 +30,16 @@ export type MetaDefinition = {
...
- * Factory to create Meta service.
+ * Factory to create a `Meta` service instance for the current DOM document.
  */
 export function createMeta() {
   return new Meta(ɵɵinject(DOCUMENT));
 }
 
 /**
- * A service that can be used to get and add meta tags.
+ * A service for managing HTML <`meta`> tags.
```
atscott, line 40:
```suggestion
 * A service for managing HTML `<meta>` tags.
```

atscott (approved):

packages/platform-browser/src/browser/meta.ts
```diff
@@ -30,14 +32,27 @@ export type MetaDefinition = {
...
 }
 
 /**
- * A service that can be used to get and add meta tags.
+ * A service for managing HTML `<meta>` tags.
+ *
+ * @usageNotes
+ * Properties of the `MetaDefinition` object match those of the
+ * HTML `<meta>` tag, which is used to provide structured metadata
+ * for a web page. Tags are helpful in Search Engine Optimization (SEO).
```
Splaktar, line 47:
```suggestion
 * for a web page. `<meta>` tags define document metadata that is important for
 * things like configuring a Content Security Policy, 
 * defining browser compatibility and security settings, setting HTTP Headers,
 * defining rich content for social sharing, and Search Engine Optimization (SEO).
```

packages/platform-browser/src/browser/meta.ts
```diff
@@ -30,14 +32,27 @@ export type MetaDefinition = {
...
+ *
+ * To identify specific `<meta>` tags in a document, use an attribute selection
+ * string in the format `"tag_attribute=`value string`"`.
```
Splaktar, line 50:
```suggestion
 * string in the format `"tag_attribute='value string'"`.
```

packages/platform-browser/src/browser/meta.ts
```diff
@@ -30,14 +32,27 @@ export type MetaDefinition = {
...
+ * For example, an `attrSelector` value of `"name='description'"` matches a tag
+ * whose `name` attribute has the value `"description"`.
+ *
+ * @see [HTML meta tag](https://www.w3schools.com/tags/tag_meta.asp)
```
Splaktar, line 54:
```suggestion
 * @see [HTML meta tag](https://developer.mozilla.org/docs/Web/HTML/Element/meta)
```

packages/platform-browser/src/browser/meta.ts
```diff
@@ -47,12 +62,27 @@ export class Meta {
...
     this._dom = getDOM();
   }
-
+  /**
+   * Retrieves or creates a specific `<meta>` tag element in the current HTML document.
+   * In searching for an existing tag, Angular attempts to match the `name` or `property` values
+   * in the provided tag definition.
+   * @param tag The definition of a `<meta>` element to match or create.
+   * @param forceCreation True to create a new element without checking whether one already exists.
+   * @returns The matching element if found, or the new element.
```
Splaktar, line 71:
```suggestion
   * @returns The existing element with the same attributes and values if found, the new element, or `null` if the tag parameter is not defined. An existing element with the same attributes and values is equivalent, therefore no changes to the element's attributes or values occur.
```

packages/platform-browser/src/browser/meta.ts
```diff
@@ -47,12 +62,27 @@ export class Meta {
   constructor(@Inject(DOCUMENT) private _doc: any) {
     this._dom = getDOM();
   }
-
+  /**
+   * Retrieves or creates a specific `<meta>` tag element in the current HTML document.
+   * In searching for an existing tag, Angular attempts to match the `name` or `property` values
```
Splaktar, line 67:
```suggestion
   * In searching for an existing tag, Angular attempts to match the `name` or `property` attribute values
```

packages/platform-browser/src/browser/meta.ts
```diff
@@ -47,12 +62,27 @@ export class Meta {
...
+   * @returns The matching element if found, or the new element.
+   */
   addTag(tag: MetaDefinition, forceCreation: boolean = false): HTMLMetaElement|null {
     if (!tag) return null;
     return this._getOrCreateElement(tag, forceCreation);
   }
 
+  /**
+   * Retrieves or creates a set of `<meta>` tag elements in the current HTML document.
+   * In searching for an existing tag, Angular attempts to match the `name` or `property` values
```
Splaktar, line 80:
```suggestion
   * In searching for an existing tag, Angular attempts to match the `name` or `property` attribute values
```

packages/platform-browser/src/browser/meta.ts
```diff
@@ -63,17 +93,37 @@ export class Meta {
     }, []);
   }
 
+  /**
+   * Retrieves a `<meta>` tag element in the current HTML document.
+   * @param attrSelector The tag attribute and value to match against, in the format
+   * `"tag_attribute=`value string`"`.
```
Splaktar, line 99:
```suggestion
   * `"tag_attribute='value string'"`.
```
I don't think that the nested backticks is right (here and a number of other places), unless you are trying to do something with ES6 string templates?

packages/platform-browser/src/browser/meta.ts
```diff
@@ -63,17 +93,37 @@ export class Meta {
...
     const list /*NodeList*/ = this._doc.querySelectorAll(`meta[${attrSelector}]`);
     return list ? [].slice.call(list) : [];
   }
 
+  /**
+   * Modifies an existing `<meta>` tag element in the current HTML document.
+   * @param tag The tag description with which to replace the existing tag content.
+   * @param selector A tag attribute and value to match against, to identify
+   * an existing tag. A string in the format `"tag_attribute=`value string`"`.
+   * If not supplied, matches a tag with the same `name` or `property` value as the replacement tag.
```
Splaktar, line 124:
```suggestion
   * If not supplied, matches a tag with the same `name` or `property` attribute value as the replacement tag.
```

packages/platform-browser/src/browser/meta.ts
```diff
@@ -47,12 +62,27 @@ export class Meta {
...
+   */
   addTag(tag: MetaDefinition, forceCreation: boolean = false): HTMLMetaElement|null {
     if (!tag) return null;
     return this._getOrCreateElement(tag, forceCreation);
   }
 
+  /**
+   * Retrieves or creates a set of `<meta>` tag elements in the current HTML document.
+   * In searching for an existing tag, Angular attempts to match the `name` or `property` values
+   * in the provided tag definition.
```
Splaktar, line 81:
```suggestion
   * in the provided tag definition in addition to verifying that all other attribute values are equal.
```

packages/platform-browser/src/browser/meta.ts
```diff
@@ -47,12 +62,27 @@ export class Meta {
   constructor(@Inject(DOCUMENT) private _doc: any) {
     this._dom = getDOM();
   }
-
+  /**
+   * Retrieves or creates a specific `<meta>` tag element in the current HTML document.
+   * In searching for an existing tag, Angular attempts to match the `name` or `property` values
+   * in the provided tag definition.
```
Splaktar, line 68:
```suggestion
   * in the provided tag definition in addition to verifying that all other attribute values are equal.
```

packages/platform-browser/src/browser/meta.ts
```diff
@@ -84,10 +134,19 @@ export class Meta {
     return this._getOrCreateElement(tag, true);
   }
 
+  /**
+   * Removes an existing `<meta>` tag element in the current HTML document.
```
Splaktar, line 138:
```suggestion
   * Removes an existing `<meta>` tag element from the current HTML document.
```

Splaktar (changes requested):
Thank you so much for taking a look at improving these API docs! These updates will make a huge difference and save people using the `Meta` service a lot of time!

packages/platform-browser/src/browser/meta.ts
Splaktar: Oh I see in the preview, the backticks are removed, but I still think that it is confusing.

Perhaps it would be clearer to just explain how these selectors are used with [`querySelector`](https://developer.mozilla.org/docs/Web/API/Document/querySelector) in the following format: `meta[${attrSelector}]`. It may be worth adding a "See Also" link for this (https://developer.mozilla.org/docs/Web/API/Document/querySelector) as well?

Perhaps we can just explain the `attrSelector` formatting a single time in the class documentation and then refer to it from the various functions that use `attrSelector`?

packages/core/src/application_ref.ts
```diff
@@ -201,7 +201,7 @@ export function assertPlatform(requiredToken: any): PlatformRef {
 }
 
 /**
- * Destroy the existing platform.
+ * Destroys the existing platform.
```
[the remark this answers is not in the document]
jbogarthyde: Can you be more specific about why and when needed? Does this function invoke destroy and cleanup functions in or accessed through the `PlatformRef`?

packages/core/src/application_ref.ts
```diff
@@ -152,7 +152,7 @@ export function createPlatform(injector: Injector): PlatformRef {
 }
 
 /**
- * Creates a factory for a platform
+ * Creates a factory for a platform.
```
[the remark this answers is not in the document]
jbogarthyde: It would. Why do you?

packages/core/src/application_ref.ts
```diff
@@ -258,12 +258,11 @@ export interface BootstrapOptions {
...
 /**
- * The Angular platform is the entry point for Angular on a web page. Each page
- * has exactly one platform, and services (such as reflection) which are common
+ * The Angular platform is the entry point for Angular on a web page.
+ * Each page has exactly one platform. Services (such as reflection) which are common
  * to every Angular application running on the page are bound in its scope.
- *
- * A page's platform is initialized implicitly when a platform is created via a platform factory
- * (e.g. {@link platformBrowser}), or explicitly by calling the {@link createPlatform} function.
+ * A page's platform is initialized implicitly when a `platformBrowser` is created using a platform
```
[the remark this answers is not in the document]
jbogarthyde: The links are not removed - they are autolinked using the backtick syntax. We only use the `@link` syntax now when autolinks fail.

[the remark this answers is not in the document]
jbogarthyde: Didn't remove the links, just updated the syntax to use backtick autolinking.

packages/core/src/application_ref.ts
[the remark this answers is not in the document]
Splaktar: @atscott do you have any input on this? I figured out the part about "what effect it has", but I'm not sure the right text to add for "why".

packages/core/src/application_ref.ts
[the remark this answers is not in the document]
Splaktar: @atscott any input on this?

packages/core/src/application_ref.ts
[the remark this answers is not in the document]
Splaktar: Oh that's cool! Good to know, thank you.

Splaktar (approved):
LGTM. This is **So Much** better than before! 👏

packages/core/src/application_ref.ts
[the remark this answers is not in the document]
atscott: I think you'd generally use it to provide or override `Providers` specific to your application's runtime needs, including `PLATFORM_INITIALIZER` and `PLATFORM_ID`. These might be useful if you have some common components/services that act differently in different "platforms". Also, because the first parameter accepts another `PlatformFactory`, you can compose factories to build up configurations that might be required by different libraries/parts of the application. 
As an example, there's a `platformBrowserTesting` that sets up `DomAdapter` and `BrowserDetection`. `TestBed.initTestEnvironment` uses a custom platform that overrides the `ResourceLoader`. 
I've personally never used this API, but that's how I understand things from digging into it a bit.

packages/core/src/application_ref.ts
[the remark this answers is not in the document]
atscott: For the most part, this is used in tests (https://github.com/angular/angular/blob/master/packages/core/test/acceptance/bootstrap_spec.ts). I've only found one instance where it's used in practice. One project has some interesting architecture around deploying parts of the application separately from others (because it's so big) and serving them from iframes. I imagine when the iframe is destroyed, i.e. navigating away from that app-within-an-app, they platform needs to be destroyed to not leak memory(?).

Splaktar:
Can this be marked "Ready for review" and be moved out of Draft PR status?

PR #37191, opened 2020-05-18
refactor(core): remove `looseIdentical` in favor of built-in `Object.is`

atscott:
Remove `looseIdentical` implementation and instead use the ES2015 `Object.is` in its place.
They behave exactly the same way except for `+0`/`-0`.
`looseIdentical(+0, -0)` => `true`
`Object.is(+0, -0)` => `false`

Other than the difference noted above, this is not be a breaking change because:
1. `looseIdentical` is a private API
2. ES2015 is listed as a mandatory polyfill in the [browser support
guide](https://angular.io/guide/browser-support#mandatory-polyfills)
3. Also note that `Ivy` already uses `Object.is` in `bindingUpdated`.

### DIFF
--- a/packages/core/src/change_detection/change_detection_util.ts
+++ b/packages/core/src/change_detection/change_detection_util.ts
@@ -6,7 +6,6 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {looseIdentical} from '../util/comparison';
 import {getSymbolIterator} from '../util/symbol';
 
 export function devModeEqual(a: any, b: any): boolean {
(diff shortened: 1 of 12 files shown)


atscott:
[global presubmit](https://test.corp.google.com/ui#id=OCL:312166426:BASE:312275994:1589924257359:9c34abb2)

pkozlowski-opensource (approved):
LGTM for fwk / runtime. Good catch, always good to shave off few bytes!

AndrewKushnir (approved):
LGTM, thanks @atscott 👍 

Reviewed-for: fw-forms

gkalpak (approved):

PR #37192, opened 2020-05-18
Create logging util system in ng-dev

josephperrott:
See individual commits.

### DIFF
--- a/dev-infra/utils/console.ts
+++ b/dev-infra/utils/console.ts
@@ -6,7 +6,14 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
+import chalk from 'chalk';
 import {prompt} from 'inquirer';
+
+
+/** Reexport of chalk colors for convenient access. */
+export const red: typeof chalk = chalk.red;
+export const green: typeof chalk = chalk.green;
+export const yellow: typeof chalk = chalk.yellow;
 
 /** Prompts the user with a confirmation question and a specified message. */
 export async function promptConfirm(message: string, defaultValue = false): Promise<boolean> {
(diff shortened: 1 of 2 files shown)


dev-infra/cli.ts
```diff
@@ -14,9 +14,27 @@ import {buildFormatParser} from './format/cli';
...
+    // when using logging functions from ./utils/console.
+    .option('log-level', {
+      type: 'string',
+      demandOption: false,
+      coerce: (level: string) => {
+        const logLevel = LOG_LEVELS[level.toUpperCase()];
+        if (logLevel === undefined) {
+          throw Error(
+              `Invalid --log-level value provided: "${level}"\n` +
+              `  Available options are: SILENT, ERROR, WARN, LOG, INFO, DEBUG`);
```
devversion, line 32: We might want to use yargs' builtin choices as we then can avoid manual coercion. e.g. https://github.com/angular/angular/blob/c98a4d6ddde28986e6e2241dbba8886957c0b126/packages/compiler-cli/ngcc/src/command_line_options.ts#L74

dev-infra/cli.ts
```diff
@@ -14,9 +14,27 @@ import {buildFormatParser} from './format/cli';
...
+        }
+        process.env['LOG_LEVEL'] = logLevel;
```
devversion, line 34: Hmm. When we initially chatted about this, I was thinking there is a way to have an handler for a single option. Going through `coerce` definitely feels not ideal 😄 How do other tools built with yargs handle this? Do they handle the log-level in every command handler?

josephperrott: As discussed offline, we will need to sort out the versioning issue we have with types before we can sort out what we want to do.

dev-infra/utils/console.ts
```diff
@@ -18,3 +19,90 @@ export async function promptConfirm(message: string, defaultValue = false): Prom
...
+export function warn(color: ChalkColor|string, ...entries: string[]) {
+  runConsoleCommand(console.warn, LOG_LEVELS.WARN, color, ...entries);
+}
+
+/**
+ * Run the console command provided, if the environments logging level greater than the
+ * provided logging level.
+ */
+function runConsoleCommand(
+    command: Function, logLevel: LOG_LEVELS, color: ChalkColor|string, ...entries: string[]) {
```
devversion, line 92: Optional nit: Might want to use `literals` or just `text`.

devversion (commented):
When you get a chance, could you update the commit message? It seems to still mention the initial design where the `--log-level` flag was created.

dev-infra/utils/console.ts
```diff
@@ -18,3 +25,88 @@ export async function promptConfirm(message: string, defaultValue = false): Prom
...
+export function warn(color: typeof chalk|string, ...entries: string[]) {
+  runConsoleCommand(console.warn, LOG_LEVELS.WARN, color, ...entries);
+}
+
+/**
+ * Run the console command provided, if the environments logging level greater than the
+ * provided logging level.
+ */
+function runConsoleCommand(
+    command: Function, logLevel: LOG_LEVELS, color: typeof chalk|string, ...text: string[]) {
```
devversion, line 87: I was hoping we'd make the `...text` change to all other overloads too, but that can be in a follow-up too.

devversion (approved):

dev-infra/utils/console.ts
josephperrott: I did a rename of the symbol, and didn't think about how that wouldn't cover the other functions.

PR #37196, opened 2020-05-19
feat(dev-infra): register ts-node when reading configuration

devversion:
`ts-node` is now an optional peer dependency of the shared dev-infra
package. Whenever a `ng-dev` command runs, and a TypeScript-based
configuration file exists, `ts-node` is set up if available.

That allows consumers of the package (as the components repo) to more
conveniently use a TypeScript-based configuration for dev-infra. Currently,
commands would need to be proxied through `ts-node` which rather
complicates the setup:

```bash
NG_DEV_COMMAND="ts-node ./node_modules/@angular/dev-infra-private/cli.js"
```

I'm thinking that it should be best-practice to use TypeScript for writing
the configuration files. Given that the tool is used primarily in Angular
projects (for which most sources are TypeScript), this should be acceptable.

Additional changes in this PR:

* Fixes common config errors not reported
* Small code change to avoid unnecessary config clone.

### DIFF
--- a/dev-infra/tmpl-package.json
+++ b/dev-infra/tmpl-package.json
@@ -24,7 +24,13 @@
   "peerDependencies": {
     "@bazel/buildifier": "<from-root>",
     "clang-format": "<from-root>",
+    "ts-node": "<from-root>",
     "tslib": "<from-root>",
     "typescript": "<from-root>"
+  },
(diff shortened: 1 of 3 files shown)


josephperrott (approved):
LGTM

josephperrott (approved):
LGTM

Issue #27296, opened 2018-11-27
Empty spaces in Extra Options router properties docs


vvasc:
# 📚 Docs or angular.io bug report

### Description

Empty spaces at https://angular.io/api/router/ExtraOptions between properties

### Expected vs Actual Behavior**
Do not show empty spaces between properties

gkalpak:
@vvasc, are you referring to the fact that the first column (`Property`) is too wide?
This is because there is one particularly long property signature 😞

cc @petebacondarwin

vvasc:
oh, my mistake

ericmartinezr:
My 5c here. In small screens it looks like there's no 2nd column, and you have to scroll all the way down to see the horizontal scrollbar.

![image](https://user-images.githubusercontent.com/13039652/49101744-b6558500-f255-11e8-98b9-ac16b388f9fc.png)

In wider screens you can see part of the 2nd column

![image](https://user-images.githubusercontent.com/13039652/49101791-d422ea00-f255-11e8-94e7-47093c4ffe37.png)

gkalpak:
I agree this is a little problematic. If anything, there should be a max cell width and the horizontal scrolling should happen on the cell-level.

petebacondarwin:
I believe we have an issue somewhere about the many issues of tables and their layout in the docs... @jenniferfell ?

s-gbz:
Faced the same issue on Firefox 66.0.2

![angular-too-wide-router-extraOption](https://user-images.githubusercontent.com/23424538/55705691-a5750700-59df-11e9-9824-71a63aae0c43.png)

sjtrimble:
Is there a reason we have the `white-space: pre; ` styling on these `code` elements? If we remove this, it helps. Still need to fix how the table columns behave.

petebacondarwin:
Do you suggest `pre-wrap` or `pre-line` or something instead? The trouble is that occasionally these code elements are multi-line (I think) and they may contain indentation. E.g.

```
{
  someProp: 1;
}
```

gkalpak:
The situation has been considerably improved with #31013.

![router-extra-options](https://user-images.githubusercontent.com/8604205/71327914-57754480-2517-11ea-8c6b-6c9013a171e6.png)

Should this be closed or are there more improvements we want to do?

ericmartinezr:
Can I be a little bit annoying? :P 

This happens now with wider screens (2560 x 1440) (only in API docs, the other docs are fine)

[...]

Issue #24707, opened 2018-06-29
set max line length for content area of documentation


jenniferfell:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead see https://github.com/angular/angular/blob/master/CONTRIBUTING.md#question
[ X] Other... Please describe: AIO styles
</code></pre>

## Current behavior

Currently, if I'm viewing the Angular docs (angular.io/docs) from a wide device, each line of text in the main center frame will be as long as absolutely possible. 

## Expected behavior

Do not allow text in the main center frame to go beyond 75 characters. If the screen is wider, then add white space left rather than continuing to expand text area. 

## Minimal reproduction of the problem with instructions

See above.

## What is the motivation / use case for changing the behavior?

Reason: Optimal line length for reading is in the 60-75 character range, depending on which research you use. All research agrees that lines over 75 characters are hard to read. The eye simply cannot effectively scan and read longer lines, so long lines limit comprehension.

## Environment

<pre><code>
Angular version: X.Y.Z

Browser:
- [ ] Chrome (desktop) version XX
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX
 
For Tooling issues:
- Node version: XX  
- Platform:  

Others:

</code></pre>

gkalpak:
I agree on having a max width, but 75 chars is way to few imo. Unless we have a two-column layout, which fills a little weird for the docs :/

Splaktar:
Added screenshots that demonstrate this issue to https://github.com/angular/angular/issues/22246#issuecomment-575447837.

jessicajaniuk:
Thanks for reporting this issue. This issue is now obsolete due to changes in the recent releases. Please update to the most recent Angular version.

If the problem still exists in your application, please [open a new issue](https://github.com/angular/angular/issues/new/choose) and follow the instructions in the issue template.

PR #37197, opened 2020-05-19
docs(api): remove api-body width for wide screens

abbeboa:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #27296

## What is the new behavior?
Removed hard coded width for api-body. Can now use full width of wide screens.

- [ ] Yes
- [x] No

### DIFF
--- a/aio/tools/transforms/templates/api/base.template.html
+++ b/aio/tools/transforms/templates/api/base.template.html
@@ -30,7 +30,7 @@
   {% endblock %}
   <aio-toc class="embedded"></aio-toc>
 
-  <div class="api-body">
+  <div class="content">
     {% block body %}{% endblock %}
   </div>
 </article>


abbeboa:
@googlebot I signed it!

aio/tools/transforms/templates/api/base.template.html
```diff
@@ -30,7 +30,7 @@ <h1>{$ doc.name $}</h1>
   {% endblock %}
   <aio-toc class="embedded"></aio-toc>
 
-  <div class="api-body">
+  <div class="content">
```
gkalpak, line 33: FWIW, I am not sure this is something we want to do (for example there have been discussions about doing the opposite: #24707). @sjtrimble has to chime in.

In any case, if we decided to do it, remiving the `api-body` class would _not_ be the way to go (since other styles depend on it). We would change the `.api-body` styles instead: [aio/src/styles/2-modules/_api-pages.scss][1]

[1]: https://github.com/angular/angular/blob/cf7dd410a159fbce0e5733c4cefb1f9a63b1347d/aio/src/styles/2-modules/_api-pages.scss#L1-L2

gkalpak (changes requested):
Thx for submitting a pull request, @abbeboa!
I do not think this is the right fix and more importantly I am not sure whether we want to fix this at all (see my comment below). More discussion is needed.

cc @sjtrimble

AndrewKushnir:
Closing this PR due to the lack of response and based on George's comment above. Please reopen if needed. Thank you.

PR #37198, opened 2020-05-19
refactor: update to tslib 2.0 and move to direct dependencies

### LINKED ISSUE #37188: Update to tslib v2
# 🐞 bug report

### Affected Package
All Angular v10 packages

### Is this a regression?
No.

### Description
According to the upgrade to TypeScript 3.9 in last v10 beta, Angular should upgrade to tslib v2 ([releases notes](https://github.com/microsoft/tslib/releases)) and requires v2 as minimum peerDependency.

alan-agius4:
Tslib version is bound to the TypeScript version used to compile the library. Thus, we shouldn't list `tslib` as a  `peerDependencies`. This is because, a user can install libraries which have been compiled with older versions of TypeScript and thus require multiple `tslib` versions to be installed.

Reference: TOOL-1374 and TOOL-1375

Closes: #37188

### DIFF
--- a/integration/bazel/src/BUILD.bazel
+++ b/integration/bazel/src/BUILD.bazel
@@ -22,6 +22,7 @@ ng_module(
         "@npm//@angular/router",
         "@npm//@types",
         "@npm//rxjs",
+        "@npm//tslib",
     ],
 )
 
--- a/integration/bazel/src/hello-world/BUILD.bazel
+++ b/integration/bazel/src/hello-world/BUILD.bazel
@@ -21,6 +21,7 @@ ng_module(
     deps = [
         "@npm//@angular/core",
         "@npm//@types",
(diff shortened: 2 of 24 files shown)


IgorMinar (approved):
LGTM! Thank you!

lacolaco:
@alan-agius4 @IgorMinar  I think this change is worth to be mentioned in CHANGELOG because the opposite `direct => peer` change is contained as a 9.0.0 feature.
https://github.com/angular/angular/blob/master/CHANGELOG.md#900-2020-02-06

Developers cannot trace the `tslib` change in CHANGELOG. 
Could I add it as a feature of `10.0.0-next.9`?

IgorMinar:
@lacolaco sounds good. Could you please send a PR to update the changelog? Thanks

PR #37208, opened 2020-05-19
build: fix deprecation notice in scripts/github/merge-pr script

josephperrott:
The deprecation notice for the merge-pr script errantly referenced
'ng dev' rather than 'ng-dev', this PR corrects this.


devversion (approved):
LGTM (i should have catched this too)

PR #37209, opened 2020-05-19
fix(dev-infra): make commit-message-filter script executable for merge

josephperrott:
Marks the commit-mesage-filter.js file as executable, allowing it
to be run by 'git filter-branch' during the merge script process.


devversion (approved):
LGTM.

PR #37210, opened 2020-05-20
fix(dev-infra): ignore `dev-infra` scope while checking for features in patch branch

AndrewKushnir:
This commit updates the script that checks master and patch branches to ignore features with `dev-infra` scope while verifying that there are no feature commits in patch branch. It's ok and in fact desirable for dev-infra features to be on the patch branch.

## PR Type

- [x] Bugfix

- [ ] Yes
- [x] No


IgorMinar (approved):
LGTM. Thank you!

PR #37217, opened 2020-05-20
feat(dev-infra): expose script for determining merge branches 

devversion:
* See individual commits

### DIFF
--- a/dev-infra/pr/merge/determine-merge-branches.ts
+++ b/dev-infra/pr/merge/determine-merge-branches.ts
@@ -1,0 +1,77 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import * as semver from 'semver';
+import {GithubConfig} from '../../utils/config';
+import {exec} from '../../utils/shelljs';
+
+/** Regular expression that matches remote head refs. */
+const UPSTREAM_HEAD_REF = /refs\/heads\/(.*)$/;
+
+/**
+ * Helper function that can be used to determine merge branches based on a given
+ * project version. The function determines merge branches primarily through the
+ * specified version, but falls back to consulting the Git remote when needed.
+ *
+ * Consulting the upstream Git remote may slow down merging, so whenever possible,
+ * the branches are determined based on the current version.
+ */
+export function determineMergeBranches(
+    currentVersion: string, github: GithubConfig): {minor: string, patch: string} {
+  const projectVersion = semver.parse(currentVersion);
+  if (projectVersion === null) {
+    throw Error('Cannot parse version set in project "package.json" file.');
+  }
+  const {major, minor, patch, prerelease} = projectVersion;
+  const isMajor = minor === 0 && patch === 0;
+  const isMinor = minor !== 0 && patch === 0;
+
+  // If there is no prerelease, then we compute patch and minor branches based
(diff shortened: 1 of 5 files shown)


dev-infra/pr/merge/determine-merge-branches.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  // If we are set to a major prerelease, we cannot statically determine the stable
+  // patch branch (as the latest minor segment is unknown). We determine it by looking
+  // for existing patch branches upstream.
+  const upstreamRefs =
+      exec(`git ls-remote --heads https://github.com/${github.owner}/${github.name}.git`)
+          .trim()
+          .split('\n');
+
+  // Iterate over retrieved upstream refs in reverse. Git sorts them lexicographically ascending.
```
josephperrott, line 57: Since it orders them lexicographically, are we gonna have issues with order since `10.x` will be after `9.x`?

From my quick test (I just pushed the same commit to different branches for testing):
```
$ git ls-remote --heads https://github.com/josephperrott/angular
1465372a0e6c8ef0f32fa4172c8c2dd5ee0a1194        refs/heads/10.1.x
1465372a0e6c8ef0f32fa4172c8c2dd5ee0a1194        refs/heads/2.x
1465372a0e6c8ef0f32fa4172c8c2dd5ee0a1194        refs/heads/9.1.x
```


I think we might have to filter the list to just things that match semver and then sort it manually.

josephperrott (changes requested):

dev-infra/pr/merge/determine-merge-branches.ts
devversion: That is a good concern, and I think this would have ultimately caused issues with the old merge script. See: https://github.com/angular/angular/blob/master/scripts/github/merge-pr#L141

With this more strict determination logic, we would _not_ pick up a wrong branch because we ensure the branch name matches a pre-determined major digit. Though it might fail for multi-digit minors. 

I think our best bet is to just consult NPM for the latest patch version. This has been proposed by @jelbourn and I think it would be safest. Not ideal is that we'd need to rely on a single package for the version, compared to consulting the actual monorepo state, but that might be just a trade-off. What do you think?

devversion: @josephperrott Updated. Should be good for another look. I think that is quite robust now

josephperrott: I think its worth thinking about what we want to potentially be determining the version.

Its odd that in our merge script we use NPM, but for when creating releases we rely on git tags.

Should be just rely on the same logic for both?  Maybe we can leverage `getSCMVersion()` from `release/env-stamp.ts`?

devversion: I think I already hinted for having such a discussion in the past (i.e. when the stamping landed), but we said that we'll handle this as a follow-up.

I still personally feel strongly about using the actual version in `package.json` for release stamping, but the merge script is concerned slightly different as it is not necessarily interested in the latest tag (that would include prereleases), but rather in the latest stable version that has been published. Most reliable seems using the NPM registry for that.

I'd be still happy to chat about this, but we can also defer this another time. We aren't set for the stamping script in components either yet.

josephperrott: I see what you are saying about them being separate concerns, I am fine with doing it this way.  Worst case is that we discover we need to take a different approach, but odds are you are right about how we should be going about it.

josephperrott (approved):
LGTM

PR #37222, opened 2020-05-20
docs: add entry for new `tsconfig.json` and updated `tsconfig.base.json` files

aikithoughts:
In version 10, there is a new `tsconfig.json` file, which contains
the paths to all other `tsconfig` files used in a workspace. The
previous `tsconfig.json` file still exists, but has been renamed to
`tsconfig.base.json`.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

n/a

Issue Number: N/A

## What is the new behavior?
n/a

- [ ] Yes
- [X] No

### DIFF
--- a/aio/content/guide/file-structure.md
+++ b/aio/content/guide/file-structure.md
@@ -40,7 +40,8 @@ The top level of the workspace contains workspace-wide configuration files, conf
 | `package-lock.json`     | Provides version information for all packages installed into `node_modules` by the npm client. See [npm documentation](https://docs.npmjs.com/files/package-lock.json) for details. If you use the yarn client, this file will be [yarn.lock](https://yarnpkg.com/lang/en/docs/yarn-lock/) instead. |
 | `src/`                  | Source files for the root-level application project. |
 | `node_modules/`         | Provides [npm packages](guide/npm-packages) to the entire workspace. Workspace-wide `node_modules` dependencies are visible to all projects. |
-| `tsconfig.json`         | Default [TypeScript](https://www.typescriptlang.org/) configuration for projects in the workspace. |
+| `tsconfig.json`         | The `tsconfig.json` file is a ["Solution Style"](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-9.html#support-for-solution-style-tsconfigjson-files) TypeScript configuration file. Code editors and TypeScript’s language server use this file to improve development experience. Compilers do not use this file. |
+| `tsconfig.base.json`    | The base [TypeScript](https://www.typescriptlang.org/) configuration for projects in the workspace. All other configuration files inherit from this base file. For more information, see the [Configuration inheritance with extends](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html#configuration-inheritance-with-extends) section of the TypeScript documentation.|
 | `tslint.json`           | Default [TSLint](https://palantir.github.io/tslint/) configuration for projects in the workspace. |
 
 
@@ -95,7 +96,7 @@ Angular components, templates, and styles go here.
 The application-specific configuration files for the root application reside at the workspace root level.
 For a multi-project workspace, project-specific configuration files are in the project root, under `projects/project-name/`.
 
-Project-specific [TypeScript](https://www.typescriptlang.org/) configuration files inherit from the workspace-wide `tsconfig.json`, and project-specific [TSLint](https://palantir.github.io/tslint/) configuration files inherit from the workspace-wide `tslint.json`.
+Project-specific [TypeScript](https://www.typescriptlang.org/) configuration files inherit from the workspace-wide `tsconfig.base.json`, and project-specific [TSLint](https://palantir.github.io/tslint/) configuration files inherit from the workspace-wide `tslint.json`.
 
 | APPLICATION-SPECIFIC CONFIG FILES    | PURPOSE |
 | :--------------------- | :------------------------------------------|
--- a/aio/content/guide/typescript-configuration.md
+++ b/aio/content/guide/typescript-configuration.md
@@ -15,22 +15,49 @@ that are important to Angular developers, including details about the following 
 
 {@a tsconfig}
 
-## TypeScript configuration
+## Configuration files
 
-A TypeScript configuration file called `tsconfig.json` guides the compiler as it generates JavaScript files for a project.
-This file contains options and flags that are essential for Angular applications.
-Typically, the file is found at the [root level of the workspace](guide/file-structure).
+A given Angular workspace contains several TypeScript configuration files.
+At the root level, there are two main TypeScript configuration files: a `tsconfig.json` file and a `tsconfig.base.json` file.
+
+The `tsconfig.json` file is a ["Solution Style"](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-9.html#support-for-solution-style-tsconfigjson-files) TypeScript configuration file.
+Code editors and TypeScript’s language server use this file to improve development experience.
+Compilers do not use this file.
+
+The `tsconfig.json` file contains a list of paths to the other TypeScript configuration files used in the workspace.
+
+<code-example lang="json" header="tsconfig.json" linenums="false">
+{
+ "files": [],
+ "references": [
+   {
+     "path": "./tsconfig.app.json"
+   },
+   {
+     "path": "./tsconfig.spec.json"
+   },
+   {
+     "path": "./projects/my-lib/tsconfig.lib.json"
+   }
+ ]
+}
+</code-example>
+
+The `tsconfig.base.json` file specifies the base TypeScript and Angular compiler options that all projects in the workspace inherit.
+
+The TypeScript and Angular have a wide range of options which can be used to configure type-checking features and generated output.
+For more information, see the [Configuration inheritance with extends](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html#configuration-inheritance-with-extends) section of the TypeScript documentation.
 
 <div class="alert is-helpful">
 
-For details about `tsconfig.json`, see the official
-[TypeScript wiki](http://www.typescriptlang.org/docs/handbook/tsconfig-json.html).
+For more information TypeScript configuration files, see the official [TypeScript wiki](http://www.typescriptlang.org/docs/handbook/tsconfig-json.html)
+For details about configuration inheritance, see the [Configuration inheritance with extends](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html#configuration-inheritance-with-extends) section.
 
 </div>
 
-The initial `tsconfig.json` for an Angular app typically looks like the following example.
+The initial `tsconfig.base.json` for an Angular workspace typically looks like the following example.
 
-<code-example lang="json" header="tsconfig.json" linenums="false">
+<code-example lang="json" header="tsconfig.base.json" linenums="false">
 {
   "compileOnSave": false,
   "compilerOptions": {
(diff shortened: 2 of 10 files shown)


aio/content/guide/file-structure.md
```diff
@@ -40,7 +40,8 @@ The top level of the workspace contains workspace-wide configuration files, conf
 | `package-lock.json`     | Provides version information for all packages installed into `node_modules` by the npm client. See [npm documentation](https://docs.npmjs.com/files/package-lock.json) for details. If you use the yarn client, this file will be [yarn.lock](https://yarnpkg.com/lang/en/docs/yarn-lock/) instead. |
 | `src/`                  | Source files for the root-level application project. |
 | `node_modules/`         | Provides [npm packages](guide/npm-packages) to the entire workspace. Workspace-wide `node_modules` dependencies are visible to all projects. |
-| `tsconfig.json`         | Default [TypeScript](https://www.typescriptlang.org/) configuration for projects in the workspace. |
+| `tsconfig.json`         | Contains the paths to all other `tsconfig.json` files used in the workspace. |
+| `tsconfig.base.json`    | Default [TypeScript](https://www.typescriptlang.org/) configuration for projects in the workspace. |
```
alan-agius4, line 44: Maybe instead of default it’s better to go with base.

IgorMinar: +1 - all the other tsconfig.json files inherit from this one using the tsconfig inheritance mechanism: https://www.typescriptlang.org/docs/handbook/tsconfig-json.html#configuration-inheritance-with-extends

@aikidave it might be worth explicitly documenting this and linking to the typescript docs.

IgorMinar (commented):
This looks like a change in the right direction, but we should also re-review and and update if needed all docs that mention tsconfig.json or TypeScript configuration. Examples:

- https://next.angular.io/guide/typescript-configuration
- https://next.angular.io/guide/language-service
- https://next.angular.io/guide/angular-compiler-options
- https://next.angular.io/guide/aot-compiler
- https://next.angular.io/guide/docs-style-guide
- https://next.angular.io/guide/elements
- https://next.angular.io/guide/dependency-injection
- https://next.angular.io/guide/deployment
- https://next.angular.io/guide/ivy-compatibility
- https://next.angular.io/guide/template-typecheck
- https://next.angular.io/guide/universal
- https://next.angular.io/guide/router

aikidave:
@alan-agius4 Can you let me know if we need to update the files that @IgorMinar listed? I could make an educated guess, but I'm not sure how accurate I would be.

alan-agius4:
Going through the above links mentioned by @IgorMinar, there are some pages that we need to update, but also, I find that in many case "tsconfig.json" is used more of a term rather than the actual file name.

Some comments below;

## I don't think that tsconfig.json in the below should be changed
- https://next.angular.io/guide/angular-compiler-options
> When you use AOT compilation, you can control how your application is compiled by specifying template compiler options in the tsconfig.json TypeScript configuration file.

To being with, you don't control only how your application is compiled, as this also applies to libraries, in the above case, it also seems to be inclusive for all tsconfig.json files.

> Like the TypeScript compiler, The Angular AOT compiler also supports extends in the angularCompilerOptions section of the TypeScript configuration file, tsconfig.json

I feel like the above section, we shouldn't change the names, because it's explaining the compiler options in a generic way.

- https://next.angular.io/guide/elements
> target: "es2015" property in their project's tsconfig.json. 
Again, to me this looks like a generic reference to a typescript config file, not to mention that in the CLI we don't have a story for elements yet.

- https://next.angular.io/guide/docs-style-guide
This is using `tsconfig.json` file to explain what is a tree, it can be any file really ex: `tslint.json`.
 
- https://next.angular.io/guide/aot-compiler
> Enable this phase explicitly by adding the compiler option "fullTemplateTypeCheck" in the "angularCompilerOptions" of the project's tsconfig.json (see Angular Compiler Options).
> You can provide options in the tsconfig.json TypeScript configuration file that control the compilation process. See Angular compiler options for a complete list of available options.

This seems to be as a reference again, because a project tsconfig can be named `tsconfig.app.json` or `tsconfig.lib.json`.

- https://next.angular.io/guide/template-typecheck
> You can disable strict checks entirely by setting strictTemplates: false in the application's TypeScript configuration file, tsconfig.json.

Seems to be the same, as an application tsconfig file is named `tsconfig.app.json`.

- https://next.angular.io/guide/deployment
> Differential loading is supported by default with version 8 and later of the Angular CLI. For each application project in your workspace, you can configure how builds are produced based on the browserslist and tsconfig.json configuration files in your application project.

This seems to be also a reference here again, and also the contents of the tsconfig.json file their are out of date. Plus, why is differential loading in the deployment section instead of a build section (https://next.angular.io/guide/build)? I find it confusing and misleading, if I had to look up DL, it wouldn't cross my mind to look for it in the deployment section.

## Need to change

- https://next.angular.io/guide/typescript-configuration
We also need to update the initial TypeScript configuration file as it's out of date

Here's the latest content
```json
{
  "compileOnSave": false,
  "compilerOptions": {
    "baseUrl": "./",
    "outDir": "./dist/out-tsc",
    "sourceMap": true,
    "declaration": false,
    "downlevelIteration": true,
    "experimentalDecorators": true,
    "moduleResolution": "node",
    "importHelpers": true,
    "target": "es2015",
    "module": "es2020",
    "lib": [
      "es2018",
      "dom"
    ]
  }
}
```
Also, in states that section stats the following;
> The initial tsconfig.json for an Angular app typically looks like the following example.
That is not totally correct, because the file described there is the base tsconfig.json which is shared amongst all workspace projects.

https://next.angular.io/guide/ivy-compatibility
https://next.angular.io/guide/universal

aio/content/guide/file-structure.md
aikidave: Done. Thank you!

aio/content/guide/typescript-configuration.md
```diff
@@ -15,22 +15,45 @@ that are important to Angular developers, including details about the following
...
 {@a tsconfig}
 
-## TypeScript configuration
+## Configuration files
 
-A TypeScript configuration file called `tsconfig.json` guides the compiler as it generates JavaScript files for a project.
+A given Angular workspace contains several TypeScript configuration files.
+At the root level, there are two: a `tsconfig.json` file and a `tsconfig.base.json` file.
+
+The `tsconfig.json` file contains a list of paths to the other TypeScript configuration files used in the workspace.
```
alan-agius4, line 23: Can we please explain why this config is used?

Something like;

This is a "Solution Style" tsconfig.json file, and is used by editors and TypeScript’s language server to improve development experience and is not intended to be used to perform a compilation.

aio/content/guide/file-structure.md
```diff
@@ -40,7 +40,8 @@ The top level of the workspace contains workspace-wide configuration files, conf
 | `package-lock.json`     | Provides version information for all packages installed into `node_modules` by the npm client. See [npm documentation](https://docs.npmjs.com/files/package-lock.json) for details. If you use the yarn client, this file will be [yarn.lock](https://yarnpkg.com/lang/en/docs/yarn-lock/) instead. |
 | `src/`                  | Source files for the root-level application project. |
 | `node_modules/`         | Provides [npm packages](guide/npm-packages) to the entire workspace. Workspace-wide `node_modules` dependencies are visible to all projects. |
-| `tsconfig.json`         | Default [TypeScript](https://www.typescriptlang.org/) configuration for projects in the workspace. |
+| `tsconfig.json`         | Contains the paths to all other `tsconfig.json` files used in the workspace. |
```
alan-agius4, line 43: Same as https://github.com/angular/angular/pull/37222/files#r429170749

aio/content/guide/typescript-configuration.md
```diff
@@ -15,22 +15,45 @@ that are important to Angular developers, including details about the following
...
+     "path": "./tsconfig.spec.json"
+   },
+   {
+     "path": "./projects/my-lib/tsconfig.lib.json"
+   }
+ ]
+}
+</code-example>
+
+The`tsconfig.base.json` guides the compiler as it generates JavaScript files for a project.
```
alan-agius4, line 42: Missing space after `the`

aio/content/guide/typescript-configuration.md
```diff
@@ -15,22 +15,45 @@ that are important to Angular developers, including details about the following
...
 This file contains options and flags that are essential for Angular applications.
```
alan-agius4, line 43: - Angular applications -> Angular applications and libraries

or 

Angular applications

aio/content/guide/universal.md
```diff
@@ -49,6 +49,7 @@ src/
     app.server.module.ts     <i>* server-side application module</i>
 server.ts                    <i>* express web server</i>
 tsconfig.json                <i>TypeScript client configuration</i>
+tsconfig.base.json           <i>TypeScript client configuration</i>
```
alan-agius4, line 52: tsconfig.json                <i>TypeScript solution style configuration</i>
tsconfig.base.json       <i>TypeScript base configuration</i>

aio/content/guide/typescript-configuration.md
```diff
@@ -15,22 +15,45 @@ that are important to Angular developers, including details about the following
...
 
-For details about `tsconfig.json`, see the official
-[TypeScript wiki](http://www.typescriptlang.org/docs/handbook/tsconfig-json.html).
+For more information TypeScript configuration files, see the official [TypeScript wiki](http://www.typescriptlang.org/docs/handbook/tsconfig-json.html)
+For details about configuration inheritance, see the [Configuration inheritance with extends](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html#configuration-inheritance-with-extends) section.
 
 </div>
 
-The initial `tsconfig.json` for an Angular app typically looks like the following example.
+The initial `tsconfig.base.json` for an Angular app typically looks like the following example.
```
alan-agius4, line 54: `The initial tsconfig.base.json for an Angular app typically looks like the following example.`

to 

`The initial tsconfig.base.json for an Angular workspace typically looks like the following example.`

aio/content/guide/typescript-configuration.md
```diff
@@ -15,22 +15,45 @@ that are important to Angular developers, including details about the following
 
 {@a tsconfig}
 
-## TypeScript configuration
+## Configuration files
 
-A TypeScript configuration file called `tsconfig.json` guides the compiler as it generates JavaScript files for a project.
+A given Angular workspace contains several TypeScript configuration files.
+At the root level, there are two: a `tsconfig.json` file and a `tsconfig.base.json` file.
```
alan-agius4, line 21: At root level, there can be more, but those two are the `main` once used by all projects within the workspace.

So, maybe re-phrase a bit, 

`At the root level, there are two main once: a `tsconfig.json` file and a `tsconfig.base.json` file.`

aio/content/guide/typescript-configuration.md
alan-agius4, line 42: Maybe we can provide more context here.
The `tsconfig.base.json` file specifies the base TypeScript and Angular compiler options which are inherited by all projects in the workspace.

The TypeScript and Angular have a wide range of options which can be used to configure type-checking features and generated output.

IgorMinar (approved):
This change now looks good to me. We might want to also update the https://github.com/angular/angular/pull/37512 PR with `TypeScript config file` terminology.

aio/content/guide/universal.md
```diff
@@ -48,11 +48,11 @@ src/
...
     app.server.module.ts     <i>* server-side application module</i>
 server.ts                    <i>* express web server</i>
-tsconfig.json                <i>TypeScript client configuration</i>
-tsconfig.app.json            <i>TypeScript client configuration</i>
-tsconfig.server.json         <i>* TypeScript server configuration</i>
-tsconfig.spec.json           <i>TypeScript spec configuration</i>
-package.json                 <i>npm configuration</i>
+tsconfig.json                <i>TypeScript solution style configuration</i>
+tsconfig.base.json           <i>TypeScript base configuration</i>
+tsconfig.base.json           <i>TypeScript browser application configuration</i>
```
alan-agius4, line 53: tsconfig.base.json -> tsconfig.app.json

alan-agius4 (approved):
LGTM, just 1 comment.

Thanks @aikidave

PR #37226, opened 2020-05-20
[patch] fix(elements): a couple of small fixes

gkalpak:
This is the same as #36114 but for the 9.1.x branch. It includes a small change to remove a TS 3.9 specific fix that is not necessary for TS <3.9 and in fact breaks TS 3.8: b178b97c61bf91a2294cdc5175eb6a70a81a66c3

See [here][1] for an example failure.

[1]: https://circleci.com/workflow-run/1b50c214-a01c-4f01-9674-305638ad90d5

### DIFF
--- a/packages/elements/src/component-factory-strategy.ts
+++ b/packages/elements/src/component-factory-strategy.ts
@@ -26,7 +26,7 @@ const DESTROY_DELAY = 10;
 export class ComponentNgElementStrategyFactory implements NgElementStrategyFactory {
   componentFactory: ComponentFactory<any>;
 
-  constructor(private component: Type<any>, private injector: Injector) {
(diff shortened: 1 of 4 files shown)


gkalpak:
BTW, I couldn't find a way (not that I tried too hard 😛) for make it work on both TS 3.8 and 3.9 (so we could have the same code on 9.1.x and master). So, master (aka 10.0.x) has code that only works with TS >=3.9 and 9.1.x will have code that only works with TS <=3.8.

I assume that is OK, right?

kara (approved):
LGTM. (This was already approved for patch by code owner, I'm only approving the 1 line fix that makes this PR different)

kara:
Closed in 1465372a0e6c8ef0f32fa4172c8c2dd5ee0a1194

PR #37232, opened 2020-05-20
Migrate ng-dev tools to use new logging system

josephperrott:
See individual commits.

### DIFF
--- a/dev-infra/commit-message/validate.ts
+++ b/dev-infra/commit-message/validate.ts
@@ -5,6 +5,8 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
+import {error} from '../utils/console';
+
 import {getCommitMessageConfig} from './config';
 
 /** Options for commit message validation. */
@@ -62,8 +64,8 @@ export function parseCommitMessage(commitMsg: string) {
 /** Validate a commit message against using the local repo's config. */
 export function validateCommitMessage(
     commitMsg: string, options: ValidateCommitMessageOptions = {}) {
-  function error(errorMessage: string) {
-    console.error(
+  function printError(errorMessage: string) {
+    error(
         `INVALID COMMIT MSG: \n` +
         `${'─'.repeat(40)}\n` +
         `${commitMsg}\n` +
@@ -91,7 +93,7 @@ export function validateCommitMessage(
   // the git history anyway, unless the options provided to not allow squash commits.
   if (commit.isSquash) {
     if (options.disallowSquash) {
-      error('The commit must be manually squashed into the target commit');
+      printError('The commit must be manually squashed into the target commit');
       return false;
     }
     return true;
@@ -104,7 +106,7 @@ export function validateCommitMessage(
   // check.
   if (commit.isFixup) {
     if (options.nonFixupCommitHeaders && !options.nonFixupCommitHeaders.includes(commit.header)) {
-      error(
+      printError(
           'Unable to find match for fixup commit among prior commits: ' +
           (options.nonFixupCommitHeaders.map(x => `\n      ${x}`).join('') || '-'));
       return false;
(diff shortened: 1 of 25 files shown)


dev-infra/commit-message/validate.spec.ts
```diff
@@ -43,7 +45,7 @@ describe('validate-commit-message.js', () => {
   beforeEach(() => {
     lastError = '';
 
-    spyOn(console, 'error').and.callFake((msg: string) => lastError = msg);
+    spyOn(console, 'error').and.callFake(((...msg: string[]) => lastError = msg[0]) as any);
```
devversion, line 48: Can't we just continue spying on `console.error` (as this is also seems simpler)

My stance is that we would want to actually spy on `console.error`. The test should not be concerned with how messages are passed to our logging utils (if we ever have transformation), but rather just assert the actual output that is printed.

dev-infra/commit-message/validate.ts
```diff
@@ -62,8 +64,8 @@ export function parseCommitMessage(commitMsg: string) {
 /** Validate a commit message against using the local repo's config. */
 export function validateCommitMessage(
     commitMsg: string, options: ValidateCommitMessageOptions = {}) {
-  function error(errorMessage: string) {
-    console.error(
+  function buildError(errorMessage: string) {
```
devversion, line 67: This is not just building an error. It's also printing. Can we disambiguate the function name?

dev-infra/ts-circular-dependencies/index.ts
```diff
@@ -84,43 +84,41 @@ export function main(
...
-      value.sort().forEach(specifier => console.info(`      ${specifier}`));
+      info(`  • ${getRelativePath(baseDir, key)}`);
+      value.sort().forEach(specifier => info(`      ${specifier}`));
     });
   } else {
-    console.info(chalk.yellow(`⚠  ${warningsCount} imports could not be resolved.`));
-    console.info(chalk.yellow(`   Please rerun with "--warnings" to inspect unresolved imports.`));
+    info(
+        yellow(`⚠  ${warningsCount} imports could not be resolved.`),
+        yellow(`   Please rerun with "--warnings" to inspect unresolved imports.`));
```
devversion, line 96: I saw this in multiple places now. Doesn't this change the message formatting? Unless I miss something, now the literals will show in the same line (also with unexpected whitespace).

Can we please keep messages as is, and in separate calls? I think this is a more clear model (each call -> new line). This will help with consistent messaging, better formatting in the code, and allow us to easily infer how output shows up in the terminal.

dev-infra/commit-message/validate.spec.ts
josephperrott: I agree, however it seems that it doesn't properly mock it out.  I will have to do more investigation to sort out what the issue is.

josephperrott: I sorted out the cause.  Because our logger loads the `console.*` symbols in a closure, its already in the closure before jasmine `spyOn` is called.

josephperrott: Updated `console.ts` functions to use loader functions instead.

dev-infra/pr/merge/git.ts
```diff
@@ -74,7 +77,8 @@ export class GitClient {
     // To improve the debugging experience in case something fails, we print all executed
     // Git commands. Note that we do not want to print the token if is contained in the
     // command. It's common to share errors with others if the tool failed.
-    console.info('Executing: git', this.omitGithubTokenFromMessage(args.join(' ')));
+    info('Executing: git');
+    info(this.omitGithubTokenFromMessage(args.join(' ')));
```
devversion, line 81: Looks like this got split into separate calls by accident.

dev-infra/release/env-stamp.ts
```diff
@@ -19,12 +21,12 @@ import {exec as _exec} from 'shelljs';
...
-  console.info(`BUILD_SCM_HASH ${getCurrentSha()}`);
-  console.info(`BUILD_SCM_LOCAL_CHANGES ${hasLocalChanges()}`);
-  console.info(`BUILD_SCM_USER ${getCurrentGitUser()}`);
-  console.info(`BUILD_SCM_VERSION ${getSCMVersion()}`);
+  info(`BUILD_SCM_BRANCH ${getCurrentBranch()}`);
+  info(`BUILD_SCM_COMMIT_SHA ${getCurrentSha()}`);
+  info(`BUILD_SCM_HASH ${getCurrentSha()}`);
+  info(`BUILD_SCM_LOCAL_CHANGES ${hasLocalChanges()}`);
+  info(`BUILD_SCM_USER ${getCurrentGitUser()}`);
+  info(`BUILD_SCM_VERSION ${getSCMVersion()}`);
```
devversion, line 29: Should we use the native `console` calls here? I think it would be a no-go not printing them if the log-level is too low? (also it will break in the future if we have default colors)

dev-infra/utils/console.ts
```diff
@@ -44,56 +44,49 @@ export enum LOG_LEVELS {
...
 
 /** Write to the console for at WARN logging level */
-export function warn(...text: string[]): void;
-export function warn(color: typeof chalk, ...text: string[]): void;
-export function warn(color: typeof chalk|string, ...text: string[]) {
-  runConsoleCommand(console.warn, LOG_LEVELS.WARN, color, ...text);
+export const warn = buildLogLevelFunction(() => console.warn, LOG_LEVELS.WARN);
+
+/** Build an instance of a logging function for the provided level. */
+function buildLogLevelFunction(loadCommand: () => Function, level: LOG_LEVELS) {
```
devversion, line 63: Can we add a super small comment on why command is retrieved lazily? i.e. Needed so that jasmine spies still work on NodeJS builtin console methods.

devversion (approved):
LGTM. A few minor nits, but that looks great so far!

It might be cool in the future to add colors for info/warn/error etc. by default. We can chat about this.

dev-infra/release/env-stamp.ts
josephperrott: Sure, that works.

PR #37233, opened 2020-05-20
Add dev-infra/ directory to tslint enforced files

josephperrott:
See individual commits.

### DIFF
--- a/dev-infra/pr/merge/failures.ts
+++ b/dev-infra/pr/merge/failures.ts
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright Google LLC All Rights Reserved.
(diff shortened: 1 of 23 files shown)


dev-infra/pr/merge/failures.ts
```diff
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright Google LLC All Rights Reserved.
+ * Copyright Google Inc. All Rights Reserved.
```
devversion, line 3: Do I miss something, or shouldn't that be `LLC`?

josephperrott: Done, I had created this before the tslint update.

devversion (approved):

PR #37245, opened 2020-05-21
build: use static patch value for targetting branches in merge config

josephperrott:
Due to the desired patch branch (10.0.x) being on a semver version
that is unreleased as stable (there is no 10.0.0 on latest, it is on
next) our logic for determining target patch branches does not work.

This change is a workaround to unblock merging in the repo while a
longer term answer is discovered.


devversion (approved):
LGTM. For later reference: The old merge script would have run into the same issue as `10.0.x` lexicographically is smaller than `9.1.x` (which is currently the patch branch).

PR #37250, opened 2020-05-22
perf(ngcc): allow immediately reporting a stale lock file

terencehonles:
Currently, if an ngcc process is killed in a manner that it doesn't clean up its lock file (or is killed too quickly) the compiler reports that it is waiting on the PID of a process that doesn't exist, and that it will wait up to a maximum of N seconds. This PR updates the locking code to additionally check if the process exists, and if it does not it will immediately bail out, and print the location of the lock file so a user may clean it up.

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

`ngcc` will wait on a lock file that will never be resolved until the maximum timeout is reached, and only then will it tell the user the path to delete the lock file.

Issue Number: N/A

## What is the new behavior?

`ngcc` will exit immediately, and report the path to the lock file, and let the user decide if it should be deleted.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/compiler-cli/ngcc/src/locking/async_locker.ts
+++ b/packages/compiler-cli/ngcc/src/locking/async_locker.ts
@@ -56,22 +56,68 @@ export class AsyncLocker {
           pid = newPid;
         }
         if (attempts === 0) {
-          this.logger.info(
+          // Check to see if the process identified by the PID is still running. Because the
+          // process *should* clean up after itself, we only check for a stale lock file when the
+          // PID changes and only once. This may mean you have to wait if the process is killed
+          // after the first check and isn't given the chance to clean up after itself.
+          if (!this.isProcessRunning(pid)) {
+            // try to re-lock one last time in case there was a race condition checking the process.
+            try {
+              return this.lockFile.write();
+            } catch (e2) {
+              if (e2.code !== 'EEXIST') {
+                throw e2;
+              }
+            }
+
+            // finally check that the lock was held by the same process this whole time.
+            const finalPid = this.lockFile.read();
+            if (finalPid === pid) {
+              throw new TimeoutError(this.lockFileMessage(
+                  `Lock found, but no process with PID ${pid} seems to be running.`));
+            } else {
+              // attempts is still 0, but adjust the PID so the message below is correct.
+              pid = finalPid;
+            }
+          }
+
+          this.logger.info(this.lockFileMessage(
               `Another process, with id ${pid}, is currently running ngcc.\n` +
-              `Waiting up to ${this.retryDelay * this.retryAttempts / 1000}s for it to finish.\n` +
-              `(If you are sure no ngcc process is running then you should delete the lock-file at ${
-                  this.lockFile.path}.)`);
+              `Waiting up to ${this.retryDelay * this.retryAttempts / 1000}s for it to finish.`));
         }
         // The file is still locked by another process so wait for a bit and retry
         await new Promise(resolve => setTimeout(resolve, this.retryDelay));
       }
     }
     // If we fall out of the loop then we ran out of rety attempts
-    throw new TimeoutError(
-        `Timed out waiting ${
-            this.retryAttempts * this.retryDelay /
-            1000}s for another ngcc process, with id ${pid}, to complete.\n` +
-        `(If you are sure no ngcc process is running then you should delete the lock-file at ${
-            this.lockFile.path}.)`);
+    throw new TimeoutError(this.lockFileMessage(`Timed out waiting ${
+        this.retryAttempts * this.retryDelay /
+        1000}s for another ngcc process, with id ${pid}, to complete.`));
+  }
+
+  protected isProcessRunning(pid: string): boolean {
+    // let the normal logic run if this is not called with a valid PID
+    if (isNaN(+pid)) {
+      this.logger.debug(`Cannot check if invalid PID "${pid}" is running, a number is expected.`);
+      return true;
+    }
+
+    try {
+      process.kill(+pid, 0);
+      return true;
+    } catch (e) {
+      // If the process doesn't exist ESRCH will be thrown, if the error is not that, throw it.
+      if (e.code !== 'ESRCH') {
+        throw e;
+      }
+
+      return false;
+    }
+  }
+
+  private lockFileMessage(message: string): string {
+    return message +
+        `\n(If you are sure no ngcc process is running then you should delete the lock-file at ${
+               this.lockFile.path}.)`;
   }
 }
--- a/packages/compiler-cli/ngcc/test/locking/async_locker_spec.ts
+++ b/packages/compiler-cli/ngcc/test/locking/async_locker_spec.ts
@@ -71,6 +71,7 @@ runInEachFileSystem(() => {
           }
           return lockFileContents;
         });
+        spyOn(process, 'kill').and.returnValue();
 
         const promise = locker.lock(async () => log.push('fn()'));
         // The lock is now waiting on the lock-file becoming free, so no `fn()` in the log.
@@ -80,12 +81,54 @@ runInEachFileSystem(() => {
           `(If you are sure no ngcc process is running then you should delete the lock-file at ${
               lockFile.path}.)`
         ]]);
+        expect(process.kill).toHaveBeenCalledWith(188, 0);
 
         lockFileContents = null;
         // The lock-file has been removed, so we can create our own lock-file, call `fn()` and then
         // remove the lock-file.
         await promise;
         expect(log).toEqual(['write()', 'read() => 188', 'write()', 'fn()', 'remove()']);
+      });
+
+      it('should fail fast when waiting on a dead process', async () => {
+        const fs = getFileSystem();
+        const log: string[] = [];
+        const lockFile = new MockLockFile(fs, log);
+        const logger = new MockLogger();
+        const locker = new AsyncLocker(lockFile, logger, 100, 10);
+
+        let lockFileContents: string|null = '188';
+        spyOn(lockFile, 'write').and.callFake(() => {
+          log.push('write()');
+          if (lockFileContents) {
+            throw {code: 'EEXIST'};
+          }
+        });
+        spyOn(lockFile, 'read').and.callFake(() => {
+          log.push('read() => ' + lockFileContents);
+          if (lockFileContents === null) {
+            throw {code: 'ENOENT'};
+          }
+          return lockFileContents;
+        });
+        spyOn(process, 'kill').and.callFake(() => {
+          throw {code: 'ESRCH'};
+        });
+
+        const promise = locker.lock(async () => log.push('fn()'));
+        // The lock has already failed so no `fn()` in the log.
+        expect(log).toEqual(['write()', 'read() => 188', 'write()', 'read() => 188']);
+        expect(logger.logs.info).toEqual([]);
+        expect(process.kill).toHaveBeenCalledWith(188, 0);
+        // Check that a missing process errors out.
+        let error: Error;
+        await promise.catch(e => error = e);
+        expect(log).toEqual(['write()', 'read() => 188', 'write()', 'read() => 188']);
+        expect(error!.message)
+            .toEqual(
+                `Lock found, but no process with PID 188 seems to be running.\n` +
+                `(If you are sure no ngcc process is running then you should delete the lock-file at ${
+                    lockFile.path}.)`);
       });
 
       it('should extend the retry timeout if the other process locking the file changes', async () => {
@@ -109,6 +152,7 @@ runInEachFileSystem(() => {
           }
           return lockFileContents;
         });
+        spyOn(process, 'kill').and.returnValue();
 
         const promise = locker.lock(async () => log.push('fn()'));
         // The lock is now waiting on the lock-file becoming free, so no `fn()` in the log.
@@ -118,6 +162,7 @@ runInEachFileSystem(() => {
           `(If you are sure no ngcc process is running then you should delete the lock-file at ${
               lockFile.path}.)`
         ]]);
+        expect(process.kill).toHaveBeenCalledWith(188, 0);
 
         lockFileContents = '444';
         // The lock-file has been taken over by another process - wait for the next attempt
@@ -131,6 +176,7 @@ runInEachFileSystem(() => {
            `(If you are sure no ngcc process is running then you should delete the lock-file at ${
                lockFile.path}.)`]
         ]);
+        expect(process.kill).toHaveBeenCalledWith(444, 0);
 
         lockFileContents = null;
         // The lock-file has been removed, so we can create our own lock-file, call `fn()` and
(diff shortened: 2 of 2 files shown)


packages/compiler-cli/ngcc/src/locking/lock_file_with_child_process/index.ts
```diff
@@ -49,9 +49,49 @@ export class LockFileWithChildProcess implements LockFile {
...
+
+      // try to re-lock one last time in case there was a race condition checking the process.
+      try {
+        this.fs.writeFile(this.path, process.pid.toString(), /* exclusive */ true);
+      } catch (e2) {
+        if (e2.code !== 'EEXIST') {
+          throw e2;
+        }
+
+        throw new Error(
```
terencehonles, line 90: Because this doesn't have the `.code === 'EEXIST'` the async and sync lockers will both throw this error to their calling code. Only the async locker subclasses `Error` and adds the code `NGCC_TIMED_OUT_EXIT_CODE`. I'm not sure if I should be adding that code here too, or if this is sufficient.

petebacondarwin (changes requested):
Hi @terencehonles - thanks for sending in this PR. It is a great idea.
Given the `SyncLocker` already bails out immediately when the lockfile is not free then this is actually only a concern of the `AsyncLocker`. So I think we should move the logic there, where we are already doing a catch of the error and checking for `EEXIST`.

Also I would be keen to make the check for whether the referenced process is still running to a separate method of `AsyncLocker`:

```ts
private isRunning(pid: string): boolean {
  try {
    return process.kill(+pid, 0);
  } catch (e) {
    // if there is a problem checking if the process is running throw the original error.
    if (e.code !== 'ESRCH') {
      throw e;
    }
  }
  return false;
}
```

terencehonles:
@petebacondarwin sounds good! I was contemplating breaking it out to a function, but was avoiding changing the surface of the `LockFile` interface. I had noticed the sync locker didn't retry, but I figured that might be coming from somewhere else. I guess not, and I realize now, that I ran into this because I was building an angular project "experimentally" instead of running `ngcc` like our normal build tooling does. 

The other reason I put it in the `LockFile` implementation is that implementation is the one reading/writing the lock file and describing how the lock is actually implemented including its contents. I was thinking there could be a case where the lock is being requested over the network or in some space that the PID wouldn't be found anymore, but still be valid. I'm probably overthinking things, and the lockers both do know enough about the internal implementation that I can see that the better process boundary would be at the locker's `.lock(...)` method. I still think the actual contents of the lock file is more "owned" by the `LockFile` itself, but I don't mind moving the code.

petebacondarwin:
I agree that the underlying implementation of the lockfile are leaking into the lockers already (e.g. the `EEXIST` error.

The main reason for abstracting it out was to allow the two different lockers, rather than having multiple different lock files. If we decided to go that way then I think we would need to move more into the lockfile and out of the lockers anyway.

terencehonles:
@petebacondarwin yup, anyways I refactored the PR. I noticed a lint error from before and figured out how to run the formatter, and I also noticed the build failed because `process.kill` is not typed to return `boolean`. I believe the cast I did and the changes should be satisfactory, but I'll poke at the tests when they run and hopefully they just pass :slightly_smiling_face:

gkalpak:
Probably stating the obvious (but just in case): We need to ensure that this works on Linux, macOS and Windows.

terencehonles:
> Probably stating the obvious (but just in case): We need to ensure that this works on Linux, macOS and Windows.

Yup, I'm on Linux and macOS should be the same. From what I can tell it does seem like node emulates `kill(pid, 0)` for Windows, but I don't have a machine to easily test against.

terencehonles:
Ok, all the tests should be passing at this point.

terencehonles:
> Ok, all the tests should be passing at this point.

GitHub seems to be having issues (I believe some of the tests failed to run, not the tests failed). Trying to shake things free.

packages/compiler-cli/ngcc/src/locking/async_locker.ts
```diff
@@ -56,22 +56,60 @@ export class AsyncLocker {
...
+    try {
+      return process.kill(+pid, 0) as unknown as true;
+    } catch (e) {
+      // If the process doesn't exist ESRCH will be thrown, if the error is not that, throw it.
+      if (e.code !== 'ESRCH') {
+        throw e;
+      }
+    }
+
+    return false;
```
petebacondarwin, line 107: NIT: this could go inside the `catch` block, which would avoid it looking like there is a possible branch where the processing exits the `try` block.

packages/compiler-cli/ngcc/src/locking/async_locker.ts
```diff
@@ -56,22 +56,60 @@ export class AsyncLocker {
...
+  }
+
+  private lockFileMessage(message: string): string {
+    return message +
+        `\n(If you are sure no ngcc process is running then you should delete the lock-file at ${
+               this.lockFile.path}.)`;
```
petebacondarwin, line 113: Nice idea to put this in a method.

packages/compiler-cli/ngcc/test/locking/async_locker_spec.ts
```diff
@@ -71,6 +71,8 @@ runInEachFileSystem(() => {
           }
           return lockFileContents;
         });
+        // @ts-ignore: process.kill is typed to return void
+        spyOn(process, 'kill').and.returnValue(true);
```
petebacondarwin, line 75: If you change the code above not to care about the return value of `kill` then the `@ts-ignore` is not needed.

packages/compiler-cli/ngcc/test/locking/async_locker_spec.ts
```diff
@@ -88,6 +91,47 @@ runInEachFileSystem(() => {
...
+          }
+        });
+        spyOn(lockFile, 'read').and.callFake(() => {
+          log.push('read() => ' + lockFileContents);
+          if (lockFileContents === null) {
+            throw {code: 'ENOENT'};
+          }
+          return lockFileContents;
+        });
+        spyOn(process, 'kill').and.callFake(() => {
```
petebacondarwin, line 115: Jasmine now has a `.andThrowError()` method. (https://jasmine.github.io/2.0/introduction.html#section-Spies:_%3Ccode%3Eand.throwError%3C/code%3E)

petebacondarwin (approved):
A couple of NITs - but this looks good. We need to check that it works on Windows...

petebacondarwin:
I tried this out on my Windows box and it appears to work! 🎉

packages/compiler-cli/ngcc/src/locking/async_locker.ts
```diff
@@ -56,22 +56,60 @@ export class AsyncLocker {
...
+
+  protected isProcessRunning(pid: string): boolean {
+    // let the normal logic run if this is not called with a valid PID
+    if (isNaN(+pid)) {
+      this.logger.debug(`Cannot check if invalid PID "${pid}" is running, a number is expected.`);
+      return true;
+    }
+
+    try {
+      return process.kill(+pid, 0) as unknown as true;
```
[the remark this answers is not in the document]
terencehonles: Sounds fine with me

packages/compiler-cli/ngcc/test/locking/async_locker_spec.ts
terencehonles: Yeah, that's what I used initially, but it wasted a bunch of my time :unamused: the `.andThrowError` function takes a string or Error object, and since `{code: 'ESRCH'}` is neither it will cast it to a string (`'[object Object]'` :unamused: ) which then will easily confuse you on why the test is failing, but it really should be working.

...and yes, I ended up having to looking for the source code for Jasmine, which annoyingly their documentation doesn't actually seem to point out the signatures for their test functions.

If you would like me to change the code I can, but it would probably need to be something like `.andThrowError(Object.assign(new Error(), {code: 'ESRCH'}))`

Sadly JavaScript still has some way to go till it is as nice as writing in a more mature language.

petebacondarwin: Ugh that is a pain... Happy to leave this as-is unless you wanted to make a helper function:

```
function createNodeError(code: string, message?: string) {
  const e = new Error(message);
  e.code = code;
  return e;
}
```

And then do something like:

```
.andThrowError(createNodeError('ESRCH'));
```

(I haven't checked whether that works!)

Happy either way.

JoostK (approved):
Being pedantic: this is marked as "feat" but targeting patch (which is 10.RC at the moment), so perhaps "perf" would be more appropriate.

packages/compiler-cli/ngcc/src/locking/async_locker.ts
```diff
@@ -56,22 +56,61 @@ export class AsyncLocker {
...
-        `(If you are sure no ngcc process is running then you should delete the lock-file at ${
-            this.lockFile.path}.)`);
+    throw new TimeoutError(this.lockFileMessage(`Timed out waiting ${
+        this.retryAttempts * this.retryDelay /
+        1000}s for another ngcc process, with id ${pid}, to complete.`));
+  }
+
+  protected isProcessRunning(pid: string): boolean {
+    // let the normal logic run if this is not called with a valid PID
+    if (isNaN(+pid)) {
```
gkalpak, line 93: Nit: Converting to number for `isNaN()` is redundant. You can just use `isNaN(pid)` (or `Number.isNaN(+pid)`).

packages/compiler-cli/ngcc/src/locking/async_locker.ts
```diff
@@ -56,22 +56,61 @@ export class AsyncLocker {
...
+            try {
+              return this.lockFile.write();
+            } catch (e2) {
+              if (e2.code !== 'EEXIST') {
+                throw e2;
+              }
+            }
+
+            throw new TimeoutError(this.lockFileMessage(
+                `Lock found, but no process with PID ${pid} seems to be running.`));
```
gkalpak, line 74: Theoretically, it is still possible to have a race condition resulting in an inaccurate message:
- Say `pid: 1` is running so `lockFile.write()` fails at the top of the `for` loop.
- We read the PID from the lock file and set the `pid` variable to `1`.
- Since `attempts` is 0, we call `isProcessRunning(1)`.
- Now, let's say that the process has ended between the last time we read the lock-file to get the PID and us calling `isProcessRunning(1)`.
- In this case, it is (theoretically) possible that another process has started and created a lock-file. Thus, the `lockFile.write()` call right after `isProcessRunning(1)` will fail (but not because the original PID (1) is still in the lock-file, as the error message will imply).

I admit that the likelihood of this race condition happening is too low (and could potentially be ignored), but we could avoid it by doing an extra check that `pid === this.lockFile.read()`.

gkalpak (approved):
A couple of optional nits. LGTM otherwise ✨

packages/compiler-cli/ngcc/src/locking/async_locker.ts
petebacondarwin: We can't even do `pid === this.lockFile.read()`, since this could also fall foul of a race condition. (Even more unlikely).
Perhaps the best option is to change the error message to accept that it might be that this is the case and still bail.

petebacondarwin:
> Being pedantic: this is marked as "feat" but targeting patch (which is 10.RC at the moment), so perhaps "perf" would be more appropriate.

This makes sense. @terencehonles can you change this?

packages/compiler-cli/ngcc/src/locking/async_locker.ts
terencehonles: I was getting type errors w/o converting to a Number. It seems like TypeScript still defines `isNaN` as taking a Number only. I can change it to `Number.isNaN` if you'd prefer that @gkalpak otherwise not sure what you'd suggest.

packages/compiler-cli/ngcc/src/locking/async_locker.ts
terencehonles: I contemplated doing another read for a PID change, but I figured it was rather unlikely, and the point of this PR was more to bail out and give a helpful message as to what might be happening. I can do the last ditch read if that might be useful, but just to be clear the complete message is the following:

> Lock found, but no process with PID ${pid} seems to be running.
> (If you are sure no ngcc process is running then you should delete the lock-file at ${this.lockFile.path}.)

and I felt that it was probably clear enough that there is a chance this was not correct (and why I didn't have `ngcc` automatically try to clean up the lock file).

I'm fine either way, but I'd probably need a suggestion on what to change the message to if that's the direction we want to take things.

packages/compiler-cli/ngcc/src/locking/async_locker.ts
terencehonles: Looks like this will also probably never change https://github.com/microsoft/TypeScript/issues/4002#issuecomment-124201510

terencehonles:
> > Being pedantic: this is marked as "feat" but targeting patch (which is 10.RC at the moment), so perhaps "perf" would be more appropriate.
> 
> This makes sense. @terencehonles can you change this?

Done, I'll let you or @gkalpak chime in on the remaining comments.

petebacondarwin:
If @gkalpak doesn't have any more suggestions by tomorrow night, I'll mark this for merge.

packages/compiler-cli/ngcc/src/locking/async_locker.ts
gkalpak: Fair enough. Let's leave it as is. Thx for looking into it.

packages/compiler-cli/ngcc/src/locking/async_locker.ts
gkalpak: > We can't even do `pid === this.lockFile.read()`, since this could also fall foul of a race condition. (Even more unlikely).

I would still prefer the less likely race condition :grin:
But I don't feel too strongly about it, given the likelyhood is quite low anyway. Let's leave it as is (I doubt anyone will run into it often enough to be bothered 😛).

petebacondarwin:
@terencehonles - the CI failure appears to be related to a built files size regression in one of integration tests. I don't think this PR should affect that. Can you try rebasing on master?

packages/compiler-cli/ngcc/src/locking/async_locker.ts
terencehonles: Since I was rebasing the change, I added a final `.read()` to check the PID remained the same.

packages/compiler-cli/ngcc/test/locking/async_locker_spec.ts
terencehonles: I slightly doubt the helper function would be reused :confused: so I think it may be better to just leave it as is.

I may look at submitting a patch to Jasmine so that if it's not strictly a string it doesn't try to construct an Error around the value passed in. That way this code could be changed in the future, or other people who are trying to do the same thing aren't bitten by the same problem.

terencehonles:
@petebacondarwin done, and almost forgot to push my change for https://github.com/angular/angular/pull/37250/files#r434037351 :sweat_smile:

packages/compiler-cli/ngcc/test/locking/async_locker_spec.ts
terencehonles: I created PR https://github.com/jasmine/jasmine/pull/1822 so hopefully they like the suggestion.

terencehonles:
Looks like I probably needed 1b55da10b1 to fix the size limits

petebacondarwin:
Good work @terencehonles - CI is green! We are ready to merge.

Issue #23370, opened 2018-04-13
Explain the newEvent helper function used in the live example of Testing docs


keilla:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Performance issue
[ ] Feature request
[X] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead see https://github.com/angular/angular/blob/master/CONTRIBUTING.md#question
[ ] Other... Please describe:
</code></pre>

## Current behavior

In [Testing Section](angular.io/guide/testing#component-dom-testing), there is a typo. 
The example in _Change an input value with dispatchEvent()_ section, in the test is showing 

 _nameInput.dispatchEvent(newEvent('input'));_

## Expected behavior

 nameInput.dispatchEvent(new Event('input'));

## Environment

<pre><code>
Angular version: X.Y.Z

Browser:
- [ ] Chrome (desktop) version XX
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX
 
For Tooling issues:
- Node version: XX  
- Platform:  

Others:

</code></pre>

marzann:
Is this issue open for contribution?

jenniferfell:
docsarea=testing
filename=guide/testing.md
URL=https://angular.io/guide/testing

@brandonroberts When you get a chance, would you look into this. The question is about the example code. Thanks!

brandonroberts:
Hi @keilla! The code used there is not a typo. There is a `newEvent` helper function that's in the live example, but not explained in the documentation. I think it would be valuable to show it though to avoid confusion.

brandonroberts:
@marzann this issue is open for contribution. Would you like to submit a PR?

1083:
So far haven't been able to find any information on this random, basically un-duck-duck-go-able `newEvent` function that's mentioned, but never imported or explained, in the docs.

1083:
@brandonroberts Definitely looks to be a typo: shouldn't `newEvent(...)` just be `new Event(...)`?

brandonroberts:
@1083 its not a typo, its not just exposed in the guide. The source for the example is here https://github.com/angular/angular/blob/master/aio/content/examples/testing/src/testing/index.ts#L23

[...]

PR #36967, opened 2020-05-07
Update hero-detail.component.spec.ts

andrewasche:
Trying to resolve the following issue: . . . 

Explain the newEvent helper function used in the live example of Testing docs #23370 . . . . https://github.com/angular/angular/issues/23370

## PR Checklist
Please check if your PR fulfills the following requirements:

- [ ] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] angular.io application / infrastructure changes
- [ ] Other... Please describe:

## What is the current behavior?

Issue Number: N/A

## Does this PR introduce a breaking change?

- [ ] Yes
- [ ] No

PR #32001, opened 2019-08-05
docs: add comment about newEvent utility function

jakpaw:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [x] Documentation content changes
- [ ] angular.io application / infrastructure changes
- [ ] Other... Please describe:

## What is the current behavior?

Issue Number: #23370

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

PR #37251, opened 2020-05-22
docs: drop `newEvent()` compatibility function

PowerKiKi:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Users are confused by the use of `newEvent` that looks like a typo, and whose origin is unclear.

Issue Number: #23370

## What is the new behavior?

Use the straightforward `new Event()` syntax.

Because PhantomJS has been deprecated since March 2018, and `newEvent`
is very confusing for newcomers that read the testing documentation,
we remove it entirely, and instead assume most, if not all, newcomers
will run tests in Chrome as it is the default.

- [ ] Yes
- [x] No

### DIFF
--- a/aio/content/examples/testing/src/app/demo/demo.testbed.spec.ts
+++ b/aio/content/examples/testing/src/app/demo/demo.testbed.spec.ts
@@ -28,7 +28,7 @@ import {
   ComponentFixture, fakeAsync, inject, TestBed, tick, waitForAsync
 } from '@angular/core/testing';
 
-import { addMatchers, newEvent, click } from '../../testing';
+import { addMatchers, click } from '../../testing';
 
 export class NotProvided extends ValueService { /* example below */ }
 beforeEach(addMatchers);
@@ -274,9 +274,11 @@ describe('demo (with TestBed):', () => {
         expect(comp.name).toBe(expectedOrigName,
           `comp.name should still be ${expectedOrigName} after value change, before binding happens`);
 
-        // dispatch a DOM event so that Angular learns of input value change.
+        // Dispatch a DOM event so that Angular learns of input value change.
         // then wait while ngModel pushes input.box value to comp.name
-        input.dispatchEvent(newEvent('input'));
+        // In older browsers, such as IE, you might need a CustomEvent instead. See
+        // https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent#Polyfill
+        input.dispatchEvent(new Event('input'));
         return fixture.whenStable();
       })
         .then(() => {
@@ -312,9 +314,11 @@ describe('demo (with TestBed):', () => {
       expect(comp.name).toBe(expectedOrigName,
         `comp.name should still be ${expectedOrigName} after value change, before binding happens`);
 
-      // dispatch a DOM event so that Angular learns of input value change.
+      // Dispatch a DOM event so that Angular learns of input value change.
       // then wait a tick while ngModel pushes input.box value to comp.name
-      input.dispatchEvent(newEvent('input'));
+      // In older browsers, such as IE, you might need a CustomEvent instead. See
+      // https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent#Polyfill
+      input.dispatchEvent(new Event('input'));
       tick();
       expect(comp.name).toBe(expectedNewName,
         `After ngModel updates the model, comp.name should be ${expectedNewName} `);
@@ -335,10 +339,12 @@ describe('demo (with TestBed):', () => {
       // simulate user entering new name in input
       input.value = inputText;
 
-      // dispatch a DOM event so that Angular learns of input value change.
+      // Dispatch a DOM event so that Angular learns of input value change.
       // then wait a tick while ngModel pushes input.box value to comp.text
       // and Angular updates the output span
-      input.dispatchEvent(newEvent('input'));
+      // In older browsers, such as IE, you might need a CustomEvent instead. See
+      // https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent#Polyfill
+      input.dispatchEvent(new Event('input'));
       tick();
       fixture.detectChanges();
       expect(span.textContent).toBe(expectedText, 'output span');
--- a/aio/content/examples/testing/src/app/hero/hero-detail.component.spec.ts
+++ b/aio/content/examples/testing/src/app/hero/hero-detail.component.spec.ts
@@ -3,7 +3,7 @@ import { ComponentFixture, fakeAsync, inject, TestBed, tick, waitForAsync } from
 import { Router } from '@angular/router';
 
 import {
-  ActivatedRoute, ActivatedRouteStub, asyncData, click, newEvent
+  ActivatedRoute, ActivatedRouteStub, asyncData, click
 } from '../../testing';
 
 import { Hero } from '../model/hero';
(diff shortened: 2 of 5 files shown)


aio/content/examples/testing/src/app/demo/demo.testbed.spec.ts
```diff
@@ -275,7 +275,7 @@ describe('demo (with TestBed):', () => {
 
         // dispatch a DOM event so that Angular learns of input value change.
         // then wait while ngModel pushes input.box value to comp.name
-        input.dispatchEvent(newEvent('input'));
+        input.dispatchEvent(new Event('input'));
```
gkalpak, line 278: Does this work on all supported browsers (including IE 9-11)?
If not we should either keep (and document) the helper or mention the limitation and the workaround for older browsers.

gkalpak (dismissed):
Thx, @PowerKiKi! I've left a comment (i.e. I don't think we are ready to just drop the helper).

aio/content/examples/testing/src/app/demo/demo.testbed.spec.ts
PowerKiKi: No, it does not work in any IE versions [according to MDN](https://developer.mozilla.org/en-US/docs/Web/API/Event/Event#Browser_compatibility).

However since all the code updated here is strictly a _sample_ for _testing_, and thus will never reach production, I assumed it was OK to drop support for those browser. Edge has now been available for 4 years. And I doubt many people actually run tests on IE out there. And for the few people who actually do I am sure they are very well aware of all the limitations and edge cases that comes with IE.

I don't think (this section of) the documentation is the right place to address issues for legacy browsers that concern a very small minority of users. Especially since this is only testing code, not production code.

gkalpak: I don't necessarily agree (personally I would expect the docs examples to work on all supported browsers), but I'll leave to the docs team to decide.

pkozlowski-opensource: I would side with @gkalpak - I don't think that we are gaining much by removing this `newEvent` helper. We can do this as we stop supporting browsers that need it.

PowerKiKi: By removing the helper, you would gain in clarity and reduce the noise in the issue tracker, see #23370, #36967 and #32001.

But feel free to close this PR if it does not align with the project goals.

daicoden: It's very weird to have documentation and a comment `//not provided by angular`, with no pointers to what/why/where it comes from, followed the rabbit hole and ended up here.

2 cents from a random user who couldn't figure out what to do from the docs - examples in documentation should work, and the ones from the testing document don't currently.

If you're saying that angular supports I.E 9-11, then it would be great if the angular testing framework provided support for I.E 9-11, and the docs use whatever the first class API is.

It's a small function, so another solution would be to put a note in the doc with code to copy into your project but right now it's confusing when reading the examples.

gkalpak: I generally agree. That's why in my original comment (https://github.com/angular/angular/pull/37251#discussion_r429110025) I said _"we should either keep **(and document)** the helper or **mention the limitation and the workaround** for older browsers"_.

PowerKiKi: do you mean add a single line comment above each `new Event` call ? something like:

```
// For compatibility with older browsers, such a IE11, see https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent#Polyfill
input.dispatchEvent(new Event('input'));
```

gkalpak: Yeah, something like that. I am not sure if it makes sense to mention it in the guide in addition/instead of as a comment in the source code. (Something for the docs team to decide.)

PowerKiKi: Then I'll wait for the official decision from the doc team, before changing anything

aikidave: Hi @PowerKiKi! Thanks for this PR! This does seem to be a tricky issue. On the one hand, we want to keep code as simple as possible. On the other hand, we don't want to confuse anyone.

I like your idea of adding a single line comment above each use of `new Event().` It's a bit of extra work to add it, but I think it will help add greater clarity for other developers.

Happy to discuss further if you want!

PowerKiKi: I added a single line comment everywhere as discussed

aio/content/examples/testing/src/app/demo/demo.testbed.spec.ts
```diff
@@ -273,9 +273,10 @@ describe('demo (with TestBed):', () => {
         expect(comp.name).toBe(expectedOrigName,
           `comp.name should still be ${expectedOrigName} after value change, before binding happens`);
 
-        // dispatch a DOM event so that Angular learns of input value change.
+        // Dispatch a DOM event so that Angular learns of input value change.
         // then wait while ngModel pushes input.box value to comp.name
-        input.dispatchEvent(newEvent('input'));
+        // For compatibility with older browsers, such a IE11, see https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent#Polyfill
```
gkalpak, line 278: I find this a little confusing. We send users to a page that shows them how to polyfill `CustomEvent`, but in the code below we are not using `CustomEvent`. Mayse say something like:

> In older browsers, such as IE, you might need to a `CustomEvent` instead.
> See https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent#Polyfill.

PowerKiKi: Updated accordingly

gkalpak:
@PowerKiki, can you please rebase this on master and squash all commits into one 🙏

aio/content/examples/testing/src/app/demo/demo.testbed.spec.ts
```diff
@@ -273,9 +273,10 @@ describe('demo (with TestBed):', () => {
         expect(comp.name).toBe(expectedOrigName,
           `comp.name should still be ${expectedOrigName} after value change, before binding happens`);
 
-        // dispatch a DOM event so that Angular learns of input value change.
+        // Dispatch a DOM event so that Angular learns of input value change.
         // then wait while ngModel pushes input.box value to comp.name
-        input.dispatchEvent(newEvent('input'));
+        // In older browsers, such as IE, you might need a CustomEvent instead, see https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent#Polyfill
```
gkalpak, line 278: This long line will not look good in the code-snippets: 

new-event-wrapped

Please change it to something like (here and in the other files):

```suggestion
        // In older browsers, such as IE, you might need a CustomEvent instead. See
        // https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent#Polyfill
```

gkalpak (changes requested):
@PowerKiki, can you please rebase this on master and squash all commits into one (that should make CI happy) 🙏
Also, please change the commit message to `docs: ...` (instead of `test(docs-infra): ...`).

aio/content/examples/testing/src/app/demo/demo.testbed.spec.ts
PowerKiKi: done

PowerKiKi:
@gkalpak, I rebased and squashed, feel free to review again.

aikithoughts (approved):
Changes look good! Thank you, @PowerKiKi for your perseverance! 

kapunahelewong:
@PowerKiKi, could CircleCI be failing because you've edited the spec files?

PowerKiKi:
I couldn't tell, because I cannot access CircleCI logs. And I am not willing to give write access to all my unrelated public repos to see logs.

gkalpak (approved):
LGTM :+1: Thx, @PowerKiKi!

FYI: I rebased it on latest master to make CI happy.

gkalpak:
Somehow this slipped through the cracks (sorry for that, @PowerKiKi).
I rebased on master, resolved conflicts and marked it for merging (as soon as CI is happy) 🚀

Issue #36128, opened 2020-03-18
Router should push to history if the state differs from the previous when the url is the same


ayeletdn:
# 🐞 bug report

### Affected Package

 The issue is caused by package @angular/router

### Is this a regression?

 I'm not sure

### Description
 When using a resolver to load a resource and the resource fails, trying to redirect the user to the same route they are currently on does not set the router's `state`. 
```
this.router.navigate(['route1'], { state: { resolveFailed: path } });
                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

### Use case: 
A table shows the user a list of items. The user requests to load a given item using the resolver, but the item no longer exists on the list. The API will return an error which requires preventing the user from navigating to the resource's page, but still showing an error message. Setting the `state` on the navigation extras can be used exactly for this purpose.

## 🔬 Minimal Reproduction

 

[Reproduced issue on Stackblitz](https://stackblitz.com/edit/angular-resolve-router-state-failure?file=src/app/route3/route3-resolver.service.ts)

### Steps:
1. click on route1 button, text should say `Route 1 State: {"fromPage": "main", "navigationId": ...}`
2. click on route2 button, text should say `Route 2 State: {"fromPage": "main", "navigationId": ...}`
3. click again on route1 button
4. click on route3 button, text should say `Route 2, State: { "resolveFailed": "https://api.example.com", "navigationId": ... }`
5. click on route1 button
6. click on route2 button, text should say `Route 2 State: {"fromPage": "main", "navigationId": ...}`
7. click on route 3 button:

Text **should** say `Route 2 State: {"resolveFailed": "https://api.example.com", "navigationId": ...}`
**Actual** text displayed: `Route 2 State: {"fromPage": "main", "navigationId": ...}`

### Explanation:
Route 3 has a resolver which connects to a resource that doesn't exist, that in turn throws an error, caught by `catchError` rxjs operator. The catchError redirects the user back to `route2`. 
- If the user asked to navigate to the route from route1, then the state will be set successfully, as seen in step 4. 
- If the request was made from `route2`, then the user doesn't effectively change route and the state doesn't get set.

## 🌍  Your Environment

**Angular Version:**
<pre><code>

8.2.14
</code></pre>

atscott:
I believe this is working as expected, at least in the example you've provided. If you enable tracing, you'll see that the last navigation was cancelled, so the router state should not be changed. Can you explain why you would expect the state to change even if the navigation gets cancelled?

[...]

PR #37267, opened 2020-05-24
fix(router):Router should push to history if the state differs and …

eaglet:
…url is the same (#36128)

Originally, the router will cancel the navigation even if it has the same URL as previously successfully processed navigation but with a different state. This is incorrect as A navigation action was clearly requested, even if it is for the same resource. If the navigation is a valid request, therefor the state change should be valid as well. This PR creates isStateChanged variable to detect if the state is different from the last successful navigation and put it into consideration when process a new navigation request.

## PR Checklist

- [] Docs have been added / updated (for bug fixes / features)

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Router will cancel the navigation even if the state differs from the previous when the URL is the same.

Issue Number: 36128

## What is the new behavior?
Router will now process the navigation if the state differs from the previous when the URL is the same.

- [x] Yes
- [] No

BREAKING CHANGE: The router will now process the navigation if it has the same URL as previously successfully processed navigation but with a different states.


eaglet:
@googlebot I fixed

atscott (changes requested):
Thanks for taking a stab at this. In addition to the comments on the code, please address the following:

* The commit message header should start with `fix(router): ` rather than `fix(packages): `
* This PR needs tests. Please add some to `integration.spec.ts`
* Please update the commit message to include "BREAKING CHANGE: " in the footer and explain that this is a breaking change because it changes how navigation is handled.
* Please update the commit message body to include some defense for why this change is necessary (what the current behavior is, what the new behavior is, and why this is a good idea)

packages/router/src/router.ts
```diff
@@ -477,8 +477,10 @@ export class Router {
                        const processCurrentUrl =
                            (this.onSameUrlNavigation === 'reload' ? true : urlTransition) &&
                            this.urlHandlingStrategy.shouldProcessUrl(t.rawUrl);
+                       const isStateChanged = this.lastSuccessfulNavigation ?
+                           t.extras !== this.lastSuccessfulNavigation.extras : false;
```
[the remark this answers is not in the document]
eaglet: Hi @atscott, thanks for the review and raised the two points. It's reasonable and I have verified it. Will make the change in next few days.

eaglet:
@atscott I have done the change as requested. Please kindly review it again. Thanks!

packages/router/test/integration.spec.ts
```diff
@@ -159,6 +159,30 @@ describe('Integration', () => {
          expect(navigation.extras.state).toEqual({foo: 'bar'});
        })));
 
+    it('should set history.state if navigation url remain the same but navigation state changed',
```
atscott, line 162:
```suggestion
    it('should trigger navigation if url remains the same but state changed',
```

eaglet:
Hi @atscott, thanks for raising this issue. It turns out that the code I submitted did have an issue. But it's not about the `navigationId` being added to the state.

Originally, the code:

`t.extras.state !== this.lastSuccessfulNavigation.extras.state`

will always be true as typescript is comparing their memory reference.

I have submitted the code below instead to fix this issue:

`JSON.stringify(t.extras.state) !== JSON.stringify(this.lastSuccessfulNavigation.extras.state)

The point you mentioned (`navigationId`  being added to the state) will not affect here. This is because the adding operation will only happen when it is going to add in the history. But the lastSuccessfulNavigation which we use for comparison is the variable that records the navigation that hasn't been added to history.

I have done some test in my local environment and now it works as expected. I think the TODO can be added to another task?

Please let me know your thoughts!

`
> Hi @eaglet, thanks for updating that. This looks pretty close to something that could work. That said, I think this may not be possible to submit at this time.
> 
> Object equality will always fail because the `navigationId` is added to the state:
> https://github.com/angular/angular/blob/044a19966dad01221296d796c4052d05a992b26f/packages/router/src/router.ts#L157-L160
> 
> 
> I think you should add another test that checks to make sure navigation events don't occur if you call `router.navigate` with the same state. I believe this test will fail. Your change will likely cause navigation events to _always_ fire because the state is always different.
> There's a note about relying on the `ng` namespace rather than adding `navigationId` to the `state`
> https://github.com/angular/angular/blob/044a19966dad01221296d796c4052d05a992b26f/packages/router/src/router.ts#L1196-L1197
> 
> 
> This was before my time so I would need to investigate what this would entail.
> I'm adding the `blocked` label to this PR. As is, I don't think it can be submitted. There are two things that could move this forward:
> 
> * Do a shallow comparison on the state object and ignore `navigationId`. That is, compare each property of the `state` for equality except for `navigationId`
> * Investigate what it means to "rely on `ng` namespace" rather than addition `navigationId` to the state

atscott:
> Hi @atscott, thanks for raising this issue. It turns out that the code I submitted did have an issue. But it's not about the `navigationId` being added to the state.
> 
> Originally, the code:
> 
> `t.extras.state !== this.lastSuccessfulNavigation.extras.state`
> 
> will always be true as typescript is comparing their memory reference.
> 
> I have submitted the code below instead to fix this issue:
> 
> `JSON.stringify(t.extras.state) !== JSON.stringify(this.lastSuccessfulNavigation.extras.state)`
> 
> The point you mentioned (`navigationId` being added to the state) will not affect here. This is because the adding operation will only happen when it is going to add in the history. But the lastSuccessfulNavigation which we use for comparison is the variable that records the navigation that hasn't been added to history.

Hi @eaglet thanks for staying engaged in this ongoing discussion. 

Good catch on using `lastSuccessfulNavigation` - I think you're right and that should work. 

With respect to `JSON.stringify` - I would prefer that you do use object reference comparison `Object.is` over `JSON.stringify`. This is the same as what's used in the Angular change detection. `ngStyle`, `ngClass`, and `ngFor` are the only exceptions there and we're looking to align those with the rest of Angular in the future.

Please add another test to verify that the navigation events are *not* called when the `state` is the same.

eaglet:
> > Hi @atscott, thanks for raising this issue. It turns out that the code I submitted did have an issue. But it's not about the `navigationId` being added to the state.
> > Originally, the code:
> > `t.extras.state !== this.lastSuccessfulNavigation.extras.state`
> > will always be true as typescript is comparing their memory reference.
> > I have submitted the code below instead to fix this issue:
> > `JSON.stringify(t.extras.state) !== JSON.stringify(this.lastSuccessfulNavigation.extras.state)`
> > The point you mentioned (`navigationId` being added to the state) will not affect here. This is because the adding operation will only happen when it is going to add in the history. But the lastSuccessfulNavigation which we use for comparison is the variable that records the navigation that hasn't been added to history.
> 
> Hi @eaglet thanks for staying engaged in this ongoing discussion.
> 
> Good catch on using `lastSuccessfulNavigation` - I think you're right and that should work.
> 
> With respect to `JSON.stringify` - I would prefer that you do use object reference comparison `Object.is` over `JSON.stringify`. This is the same as what's used in the Angular change detection. `ngStyle`, `ngClass`, and `ngFor` are the only exceptions there and we're looking to align those with the rest of Angular in the future.
> 
> Please add another test to verify that the navigation events are _not_ called when the `state` is the same.

Hi @atscott, thanks for the instant reply and keeping watching this issue with me!

Unfortunately, Object.is() is actually the same as the operator '===' which will return false if the two object references are not the same ([Reference](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is)). I think it's not going to provide the expectation that we want when we are in a typical situation to compare the two objects.

But you are right that JSON.stringify might not be the best solution as it's actually comparing the string not object. I am thinking to use Lodash or just create a new function for comparing objects.

So there are three solutions now:
1. Use JSON.stringift
2. Use _.isEqual() 
3. Create new function

Please let me know your preference!

atscott:
Hi @eaglet, `Object.is` is what we want here. As I said before, this is in line with the rest of Angular. Here's an example https://github.com/angular/angular/blob/55979fe0ae91a2e5306050d008b0173a491e53b9/packages/core/src/render3/bindings.ts#L46-L53
Bindings are only considered "updated" if the reference has changed.

eaglet:
@atscott, thanks for the reply. I try to use `Object.is()` but it failed.
In the router.ts, the `lastSuccessfulNavigation` is equal to the `currentNavigation` at the end of processing and the `currentNavigation` is created by the line:
```
this.currentNavigation = {
                         id: t.id,
                         initialUrl: t.currentRawUrl,
                         extractedUrl: t.extractedUrl,
                         trigger: t.source,
                         extras: t.extras,
                         previousNavigation: this.lastSuccessfulNavigation ?
                             {...this.lastSuccessfulNavigation, previousNavigation: null} :
                             null
                       };
```
Which means it actually always creates a new memory space when a new navigation request coming. Thus, to use `Object.is()` here will always return true if we compare `Object.is(t.extras.state, this.lastSuccessfulNavigation.extras.state)`

So I think we might need another solution other than using `Object.is()`. 
I hope this makes sense.

eaglet:
Hi @atscott, I might use the wrong code for reference. Sorry for making the confusion.
I think this [example](https://www.typescriptlang.org/play/index.html#code/MYewdgzgLgBNCGUCmMC8MDeN4C4YEYYBfAbgChRJYFkAmNTbPQ0i8aGABwCckA3AJYgArhABy8PgyxIAHlG7wIeLDRRFi5ShzBIA7hKnoZ8xcsZr6G1vAA2SblAAUPfkNGGAdHIVLPatFR0XQNJb1M-SwBKcjsHZwB5ACMAKyRgKE8BCBdeQRFxMJ8zfyhEJAAaGBCvYsiyuiiYsjjHJwApAGUEsVLuATAAcwEAMwBPXLcC2oiIUvKowPQunr6B4fGnGqLZ+cbmoA) better explains the situation we got:
So here we are comparing: 1. t.extras.state 2. this.lastSuccessfulNavigation.extras.state
Since t is current navigation and lastSuccessfulNavigation is the t from last round of the navigation, their state might be a different object.

Basically, I am using this [stackblitz](https://stackblitz.com/edit/angular-resolve-router-state-failure-4uzs9o?file=src/app/route1/route1.component.ts) code which provided by the person who raises this issue to do the test and I try to use both router.navigate with the same state object and creating a new object with the same values. But Object.is() will return false in both cases.

And the reason why my current branch is working as expected is that right now it's still using `JSON.stringify`.

Please let me know if there is a more efficient communication channel with you! I am happy to do the demo.

I 'll look into the imperative case! Thanks for letting me know!

And thanks again for your time here.

eaglet:
Hi @atscott, 
As discussed, I have added the deep comparison function and the test case for the situation when state and URL are both the same. Please let me know the test result!

Thanks,
Henry

packages/router/src/shared.ts
```diff
@@ -156,3 +156,23 @@ export function defaultUrlMatcher(
 
   return {consumed: segments.slice(0, parts.length), posParams};
 }
+
+/**
+ * Takes two values and returns true only if they are the same value or are
+ * objects with the same properties whose values are also equal
+ * @param params two objects or string to deep compare.
+ * @returns true or false
```
atscott, line 164:
```suggestion
```

packages/router/src/shared.ts
```diff
@@ -156,3 +156,23 @@ export function defaultUrlMatcher(
...
+ */
+export function deepEqual(a: any, b: any) {
```
atscott, line 166:
```suggestion
export function shallowEqual(a: any, b: any) {
```

This is a shallow equality check, not deep.

packages/router/src/router.ts
```diff
@@ -472,8 +472,12 @@ export class Router {
                        };
                      }),
                      switchMap(t => {
+                       const isStateChanged = this.lastSuccessfulNavigation ?
+                           !deepEqual(t.extras.state, this.lastSuccessfulNavigation.extras.state) :
+                           false;
```
atscott, line 477:
```suggestion
                       const isStateChanged = 
                           !deepEqual(t?.extras?.state, this.lastSuccessfulNavigation?.extras?.state);
```
If there is no `lasSuccessfulNavigation`, then the state is changed. `this.navigated` will also be `false` so there should really be no difference.

`extras` might actually be `undefined` even though the type doesn't say so. The issue comes from these two places in the code:
https://github.com/angular/angular/blob/cc49a91de7e228c6bdff8ca58936f05abe0ceb9f/packages/router/src/router.ts#L756
https://github.com/angular/angular/blob/cc49a91de7e228c6bdff8ca58936f05abe0ceb9f/packages/router/src/router.ts#L506-L509

atscott (changes requested):
A couple more comments. I'll run another presubmit once they're addressed.

packages/router/src/shared.ts
eaglet: Hi @atscott,
I am a little bit confused here. Below is the definition of two types of checks:
Shallow equality checking (or reference equality) is checking that two different variables reference the same object; in contrast, deep equality checking (or value equality) is checking every value of two objects' properties.

And Isn't our code below checking every value in the object? 
```
aKeys.every(k => b.hasOwnProperty(k) && Object.is(a[k], b[k]));
```

atscott: I guess this is one of those things that isn't exactly agreed upon by everyone. I think the argument would be that this is actually neither, but a [depth-1 equality check](https://books.google.com/books?id=CFtuCQAAQBAJ&pg=PA235&lpg=PA235&dq=shallow+equality+check+definition&source=bl&ots=9ip6ZiRSYE&sig=ACfU3U2p3Ef1jTj9WgFAISaQTXXhKkRvCw&hl=en&sa=X&ved=2ahUKEwjrmr_s0obqAhUKo54KHXndDToQ6AEwD3oECAgQAQ#v=onepage&q=shallow%20equality%20check%20definition&f=false). It's certainly not deep, which would imply that we recursively check the properties. But I can see that shallow equals might actually refer to reference equality.

eaglet: Agree with the depth-1 equality check! Should I name it depthOneEqual? or keep it as shallowEqual?

atscott: depthOneEqual sounds good to me 👍

eaglet:
@atscott , I just fixed the comments! Thanks for the review!

atscott:
Hi @eaglet, I still have to look at the results of the presubmit, but they are likely irrelevant because of another snag in the implementation. When handling a navigation, the state will actually be replaced rather than pushed, which is what the feature request was:
https://github.com/angular/angular/blob/1197965e69d658c92422d49b1f2a1ba90498af67/packages/router/src/router.ts#L1204-L1206

Additionally, the component will be re-used and the lifecycle hooks won't be rerun as might be expected due to the `DefaultRouteReuseStrategy` implementation here:
https://github.com/angular/angular/blob/1197965e69d658c92422d49b1f2a1ba90498af67/packages/router/src/route_reuse_strategy.ts#L77

This is getting a lot more involved than anticipated and I'm not sure we'll be able to solve this without _yet another_ option on the router to enable this. I'm feeling a bit more confident that the actual solution would be to recommend using the current options available in the router to achieve this specific use-case rather than building the functionality into the router itself. WDYT?

eaglet:
Hi @atscott, thanks for pointing out the lifecycle hook won't re-run. I was trying to search for it for couple of days!

I think I still preferred to implement this into the router as it should be there originally. But I also understand your concern about how this breaking change could affect the existing applications. 

Can we implement this as another option on the router for Angular 9 and set it as default in Angular 10? Will this minimize the impact to other applications?

atscott:
> Hi @atscott, thanks for pointing out the lifecycle hook won't re-run. I was trying to search for it for couple of days!
> 
> I think I still preferred to implement this into the router as it should be there originally. But I also understand your concern about how this breaking change could affect the existing applications.
> 
> Can we implement this as another option on the router for Angular 9 and set it as default in Angular 10? Will this minimize the impact to other applications?

There are at least two things that would need to be addressed in this PR:
* The logic that prevents duplicate navigations needs to be adjusted: https://github.com/angular/angular/blob/35502c739c70427bfcee239882d91ba0f4285cf9/packages/router/src/router.ts#L1133-L1136
* The logic for setting the browser url via `replaceState` or `pushState` needs adjusting: https://github.com/angular/angular/blob/1197965e69d658c92422d49b1f2a1ba90498af67/packages/router/src/router.ts#L1204-L1206
These are the two changes that are most likely to be breaking. They may not, but it's something to think about.

In addition, we need more tests to verify the new behavior. None of the current tests caught that a new item isn't pushed to history - they are only checking the navigation events.

Lastly, we need to consider the tradeoffs for changes in terms of bundle size. The additional code to support this case is small, but will be in every bundle regardless of whether applications need it. So a few questions are:
* Is the current behavior clearly incorrect?
* Is this something that a lot of people need?
* Are there acceptable workarounds using the current API?
* Is the change sufficiently small or is it tree-shakeable (can the code be removed if unused - here, the answer is no)?

Side note: could you please revert the changes to `browser-providers.conf.js`

atscott:
Hi @eaglet - Thanks for giving this one a shot. I'm going to close this PR due to the aforementioned concerns. This can be revisited in the future if it becomes clear that we can justify the tradeoffs.

PR #37280, opened 2020-05-25
build(docs-infra): add deployment mode for release-candidate versions

gkalpak:
This PR adds the necessary build configuration/deployment mode (`rc`) for deploying release-candidate versions to angular.io. It is part of the work needed to prepare angular.io for [Versioning - Merging, Branching & Deployment](https://docs.google.com/document/d/197kVillDwx-RZtSVOBtPb4BBIAw0E9RT3q3v6DZkykU).

See individual commits for more info.


gkalpak:
This is blocked on renaming/creating the necessary Firebase projects.

aio/package.json
```diff
@@ -175,4 +175,4 @@
     "xregexp": "^4.0.0",
     "yargs": "^7.0.2"
   }
-}
\ No newline at end of file
+}
```
petebacondarwin, line 178: @mgol  would be pleased :-)

aio/scripts/deploy-to-firebase.js
```diff
@@ -0,0 +1,151 @@
...
+    // List the branches that start with the major version.
+    exec(`git ls-remote origin refs/heads/${currentBranchMajorVersion}.*.x`).split('\n')
+        // Extract the version number.
+        .map(line => line.split('/')[2])
+        // Sort by the minor version.
+        .sort((a, b) => a.split('.')[1] - b.split('.')[1])
+        // Get the highest version.
+        .pop();
+
+  // Do not deploy as it is not the latest branch for the given major version.
```
petebacondarwin:
```suggestion
  // Do not deploy if it is not the latest branch for the given major version.
```

aio/scripts/deploy-to-firebase.js
```diff
@@ -0,0 +1,151 @@
...
+function computeMajorVersion(branchName) {
+  return +branchName.split('.', 1)[0];
+}
+
+function exec(cmd, opts) {
+  return _exec(cmd, {silent: true, ...opts}).trim();
+}
+
+function yarn(cmd) {
+  return exec(`yarn ${cmd}`);
```
petebacondarwin, line 191: Does this output error messages to the console if the command fails?

aio/scripts/deploy-to-firebase.js
```diff
@@ -0,0 +1,151 @@
...
+
+// Set `deployedUrl` as parameter in the opensearch description.
+// `deployedUrl` must end with `/`.
+yarn(`set-opensearch-url ${deployedUrl}`);
+
+// Check payload size.
+yarn('payload-size');
+
+// Deploy to Firebase.
+yarn(`firebase use "${projectId}" --token "${CI_SECRET_AIO_DEPLOY_FIREBASE_TOKEN}"`);
```
petebacondarwin: Does the `yarn()` helper display the secret token to the console?

aio/scripts/deploy-to-firebase.js
```diff
@@ -89,15 +85,24 @@ if (CI_BRANCH === 'master') {
...
 
-  deployInfo = deployInfoPerTarget.archive;
+  deployInfo = (currentBranchMajorVersion < stableBranchMajorVersion) ?
+      // This is the latest minor version for a major that is less than the stable major version:
+      // Deploy as `archive`.
+      deployInfoPerTarget.archive :
+      // This is the latest minor version for a major that is equal or greater than the stable major
+      // version, but not the stable version itself:
+      // Deploy as `rc`.
+      deployInfoPerTarget.rc;
```
petebacondarwin: 👍  nice bit of logic here

aio/scripts/deploy-to-firebase.js
```diff
@@ -7,150 +7,190 @@ const {cd, cp, exec: _exec, set} = require('shelljs');
...
 
-const {deployEnv, projectId, deployedUrl} = deployInfo;
-console.log(
-    `Git branch        : ${CI_BRANCH}\n` +
-    `Build/deploy mode : ${deployEnv}\n` +
-    `Firebase project  : ${projectId}\n` +
-    `Deployment URL    : ${deployedUrl}\n`);
+function computeMajorVersion(branchName) {
+  return +branchName.split('.', 1)[0];
+}
```
petebacondarwin, line 117: OOC why move this above `computeInputVars()`?

petebacondarwin (approved):

petebacondarwin:
Oh and the first commit has a typo: `switch` -> `switches`.

aio/scripts/deploy-to-firebase.js
gkalpak: These comments were just moved over from `deploy-to-firebase.sh` and I didn't want to change them. This is removed in a subsequent commit, so we're good :smiley:

aio/scripts/deploy-to-firebase.js
gkalpak: It does output a generic error message (something like `exec: internal error`), but not the actual error message.

Good catch! I'll update this.

aio/scripts/deploy-to-firebase.js
gkalpak: Without the `--silent/-s` flag, yarn does indeed echo the command (including expanded vars). So, it does indeed print the token to the output (even with out current setup). However, CircleCI masks secret env vars with `*****`. See https://circleci.com/gh/angular/angular/726807.

In this new implementation, we pass `silent: true` to ShellJS `exec()`, so the output (including the token) will not be printed to the output unless we explicitly `console.log()` the value returned by `yarn(...)`. (And even if we accidentally printed it, CircleCI would hopefully mask it for us.)

I can pass the `--silent` flag to make this even more safe.

aio/scripts/deploy-to-firebase.js
gkalpak: No idea. I usually order my function alphabetically, so I assume I originally named `computeInputVars()` something else that was lexicographically greater than `computeMajorVersion()` 🤷‍♂️

Moved below. Thx for catching 😇

aio/scripts/deploy-to-firebase.js
gkalpak: Actually, it seems that this is fixed in the latest version of ShellJS, so I'll update ShellJS from v0.7.7 to v0.8.4 instead.

gkalpak:
Thx for the review, @petebacondarwin. I added two more commits addressing your feedback:
- 6a53ab96e17b5fc50a461e04fc79c26eba137b2f
- 40bf8c31323531cf3c2f4e88f5938c3e827d068b

PTAL

petebacondarwin (approved):

IgorMinar (approved):
This looks great!! But I don't think that we can land this now because rc-angular-io is not accessible to the aio-deploy@angular.io account for the reasons we discussed in the docs-infra meeting.

Should we still land this but change the project target for rc and next deployments to be the current  project? (I believe it's the aio-staging)

gkalpak:
It depends on what we want to do with RCs. The `aio-staging` project is indeed what will become `next-angular-io`, but currently we do not deploy RCs anywhere. If we want to keep it this way, then we could partially land this PR, but I don't think there is much benefit in doing so.

gkalpak:
Superceded by #39470.

Issue #36998, opened 2020-05-08
Reactive form async validation fires on change when updateOn: 'submit'


popec:
# 🐞 bug report

### Affected Package

 The issue is caused by package @angular/forms@9.1.5

### Is this a regression?

 Don't know

### Description

 Async validation for FormGroup fires on every input change event when updateOn: 'submit' is set.

## 🔬 Minimal Reproduction

Grab the example form validation from docs (https://angular.io/guide/form-validation) and change Reactive form declaration like this
```
this.heroForm = new FormGroup({
  'name': new FormControl(this.hero.name, {
    validators: [Validators.required, Validators.minLength(4), forbiddenNameValidator(/bob/i)],
    updateOn: 'change'
  }),
  'alterEgo': new FormControl(this.hero.alterEgo, {
    validators: [Validators.required],
    updateOn: 'change'
  }),
  'power': new FormControl(this.hero.power, {
    validators: [Validators.required],
    updateOn: 'change'
  })
}, {
  validators: identityRevealedValidator,
  asyncValidators:  [this.alterEgoValidator.validate.bind(this.alterEgoValidator)],
  updateOn: 'submit'
});
```

## 🌍  Your Environment

**Angular Version:**
<pre><code>

Angular CLI: 9.1.4
Node: 12.13.0
OS: win32 x64

Angular: 9.1.5
... animations, common, compiler, compiler-cli, core, forms
... platform-browser, platform-browser-dynamic, router, upgrade
Ivy Workspace: Yes

Package                           Version
-----------------------------------------------------------
@angular-devkit/architect         0.901.4
@angular-devkit/build-angular     0.901.4
@angular-devkit/build-optimizer   0.901.4
@angular-devkit/build-webpack     0.901.4
@angular-devkit/core              9.1.4
@angular-devkit/schematics        9.1.4
@angular/cli                      9.1.4
@ngtools/webpack                  9.1.4
@schematics/angular               9.1.4
@schematics/update                0.901.4
rxjs                              6.5.5
typescript                        3.8.3
webpack                           4.42.0

</code></pre>

**Anything else relevant?**

AndrewKushnir:
Hi @popec, thanks for reporting the issue. Could you please also provide a minimal repro using [Stackblitz](https://stackblitz.com/), so that we can look into this? Thank you.

popec:
https://stackblitz.com/edit/angular-ivy-euzbxu

cosminv6u:
[...]

PR #37286, opened 2020-05-25
fix(forms): do not update group with "submit" update strategy on child controls input

dimakuba:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Previously, group was updating on child controls input, even if updateOn was set to "submit".

Issue Number: https://github.com/angular/angular/issues/36998

## What is the new behavior?
Do not update group with "submit" update strategy on child controls input.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/forms/src/model.ts
+++ b/packages/forms/src/model.ts
@@ -660,7 +660,8 @@ export abstract class AbstractControl {
       (this.statusChanges as EventEmitter<string>).emit(this.status);
     }
 
-    if (this._parent && !opts.onlySelf) {
+    if (this._parent && !opts.onlySelf &&
+        (this._parent.updateOn === 'change' || this._parent.updateOn === this.updateOn)) {
       this._parent.updateValueAndValidity(opts);
     }
   }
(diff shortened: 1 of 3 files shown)


dylhunn:
@AndrewKushnir What do you think about this change, from a philosophical perspective?

On the one hand, this would allow you to have a control update on `change`, and its parent update on `submit`, and both settings would be respected. On the other hand, it could be confusing -- the inner control might be `invalid`, while the parent group could be `valid`.

@dimakuba, does this affect any other case besides explicitly setting the parent to `submit` and the child to `change`?

dylhunn:
Consider the following nested structure:

```
FormGroup A (updateOn: change)
  FormGroup B (updateOn: submit)
    FormControl (updateOn: change)
```

The value update and validity propagation runs from bottom to top, i.e. `FormControl` -> `FormGroup B` -> `FormGroup A`.
Imagine that `FormControl` gets an update from the UI (thus `updateOn` is in `change` mode), and the update should propagate to the top.

Currently all levels would be notified, but with the proposed fix, the process will be stopped at `FormGroup B`, because its `updateOn` is not `change`. However, `FormGroup A` has `updateOn` defined as `change`, so it expects its value/validity to be updated.

Let's say we change the fix to keep propagation, but prevent value/validity changes on individual levels (thus all items receive updates, but `FormGroup B` will just ignore that) - now it might be tricky to say that the `FormGroup A` is valid since the `FormGroup B` is unknown in this case (i.e. underlying control's value has changed, but validator never ran). 

This is a whole can of worms, and it's probably best to air on the side of always running the value/validation update -- it would be much worse to never run them at all in the above cases.

Issue #37061, opened 2020-05-12
Missing locale


Mejans:
# 🐞 bug report

### Affected Package
All

### Description
Hello,
in https://github.com/angular/angular/tree/master/packages/common/locales there is no locale for Occitan, usually oc_FR because it is in the «seed» folder of CLDR I think.
But all the information that are extracted from CLDR are already in the CLDR file.
Would it be possible to integrated it?
Some developers don't know how to make an application with this locale when this file is missing.
See here: https://github.com/Chocobozzz/PeerTube/issues/2612

Best regards

## 🔬 Minimal Reproduction

 https://stackblitz.com/...

## 🔥 Exception or Error
<pre><code>

</code></pre>

## 🌍  Your Environment

**Angular Version:**
<pre><code>

</code></pre>

**Anything else relevant?**

Mejans:
And I ping @Chocobozzz so he can better explain what problem he is facing :)
Many thanks!

petebacondarwin:
Not sure how to bring in locales that are in seed still. Anyone got any ideas?
Meanwhile, it is definitely "feasible" to handcraft your own locale data file if you are desperate.

Mejans:
Yes this is what I suggested here:
https://github.com/Chocobozzz/PeerTube/issues/2612
It's even really easy as the data, day and month day, are really easily available across the Internet :)

petebacondarwin:
Looks like it was worked around in https://github.com/Chocobozzz/PeerTube/commit/2a39506c7da9ef79671d4c21539313b6d49b1884

Closing as I don't think we need to do anything further in the framework to support this. If the OC locale makes it into the main CLDR repository then we will pick it up automatically.

PR #37315, opened 2020-05-27
[i18n] Adds Occitan language

Mejans:
In its general form, oc_FR

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #37061

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/common/locales/oc.ts
+++ b/packages/common/locales/oc.ts
@@ -1,0 +1,103 @@
+// This code is not generated
+// See angular/tools/gulp-tasks/cldr/extract.js
+
+const u = undefined;
+
+function plural(n: number): number {
+  let i = Math.floor(Math.abs(n));
+  if (i === 0 || i === 1) return 1;
+  return 5;
+}
+
+export default [
+  'oc',
+  [['a. m.', 'p. m.'], u, u],
+  u,
+  [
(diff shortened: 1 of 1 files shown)


Mejans:
@googlebot I signed it!

petebacondarwin (changes requested):
Hand coding a locale data file is not a viable solution here, I'm afraid.
It would not be possible to keep it updated automatically.
The only option I could foresee us accepting would be if we could change the extraction tooling (i.e. angular/tools/gulp-tasks/cldr/extract.js) to be able to bring in this locale from the CLDR data.

Mejans:
I totally agree but at the moment it's the only way =/
Or maybe consider adding locales that are in seeds and that have the required data?

petebacondarwin:
Is it not possible to include this file in your own application and then load it manually using the `registerLocaleData()` method?

Chocobozzz:
> Is it not possible to include this file in your own application and then load it manually using the registerLocaleData() method?

I can try, but Occitan could be used in other Angular projects :)

petebacondarwin:
> Occitan could be used in other Angular projects :)

That is possible, but you are the first so far, and once it becomes common enough it will leave the CLDR seed and enter the main repository, in which case we will get it here for free.

petebacondarwin:
I'm going to close this PR, since we won't land it as-is. The workaround is to provide a custom locale in your own project.

PR #37317, opened 2020-05-27
fix(core): reenable decorator downleveling for Angular npm packages

IgorMinar:
In #37221 we disabled tsickle passes from transforming the tsc output that is used to publish all
Angular framework and components packages (@angular/*).

This change however revealed a bug in the ngc that caused __decorate and __metadata calls to still
be emitted in the JS code even though we don't depend on them.

Additionally it was these calls that caused code in @angular/material packages to fail at runtime
due to circular dependency in the emitted decorator code documeted as
https://github.com/microsoft/TypeScript/issues/27519.

This change partially rolls back #37221 by reenabling the decorator to static fields (static
properties) downleveling.

This is just a temporary workaround while we are also fixing root cause in `ngc` - tracked as
FW-2199.

Resolves FW-2198.
Related to FW-2196

### DIFF
--- a/packages/bazel/src/ngc-wrapped/index.ts
+++ b/packages/bazel/src/ngc-wrapped/index.ts
@@ -191,13 +191,12 @@ export function compile({
     fileLoader = new UncachedFileLoader();
   }
 
+  compilerOpts.annotationsAs = 'static fields';
   if (!bazelOpts.es5Mode) {
     if (bazelOpts.workspaceName === 'google3') {
       compilerOpts.annotateForClosureCompiler = true;
-      compilerOpts.annotationsAs = 'static fields';
     } else {
       compilerOpts.annotateForClosureCompiler = false;
-      compilerOpts.annotationsAs = 'decorators';
     }
   }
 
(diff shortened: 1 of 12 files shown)


kyliau (approved):

kara (approved):
LGTM 

packages/bazel/src/ngc-wrapped/index.ts
```diff
@@ -191,13 +191,12 @@ export function compile({
     fileLoader = new UncachedFileLoader();
   }
 
+  compilerOpts.annotationsAs = 'static fields';
```
alxhub, line 194: I think previously this was only done for ES2015 code, and ES5 used decorators. Is this an intentional change from the pre-v10 behavior?

packages/bazel/src/ngc-wrapped/index.ts
```diff
@@ -281,6 +280,8 @@ export function compile({
   // Prevent tsickle adding any types at all if we don't want closure compiler annotations.
   if (compilerOpts.annotateForClosureCompiler) {
     bazelHost.transformTypesToClosure = true;
+  }
+  if (compilerOpts.annotationsAs === 'static fields') {
```
alxhub, line 284:
```suggestion
  if (compilerOpts.annotateForClosureCompiler || compilerOpts.annotationsAs === 'static fields') {
```
To ensure we preserve the previous behavior.

alxhub (approved):

packages/bazel/src/ngc-wrapped/index.ts
IgorMinar: yes. we no longer ship es5 to npm so the only thing affected by this is our dev mode within angular/angular and @devversion is looking into switching that over to es2015 as well, so none of this will matter soon.

packages/bazel/src/ngc-wrapped/index.ts
IgorMinar: I see. I can do that.

IgorMinar: done

PR #37318, opened 2020-05-27
refactor(dev-infra): move GitClient to common util

josephperrott:
Moves GitClient from merge script into common utils for unified
method of performing git actions.


dev-infra/pr/merge/config.ts
```diff
@@ -31,30 +31,20 @@ export interface TargetLabel {
...
-  useSsh?: boolean
-}
-
 /**
  * Configuration for the merge script with all remote options specified. The
  * default `MergeConfig` has does not require any of these options as defaults
  * are provided by the common dev-infra github configuration.
  */
-export type MergeConfigWithRemote = MergeConfig&{remote: MergeRemote};
+export type MergeConfigWithRemote = MergeConfig&{remote: GithubConfig};
```
devversion, line 39: I think the idea was that `MergeRemote` is a subset of the `GithubConfig` (if we ever add more stuff).  Can we have a more fine-grained type for the `GitClient` like `GithubRemote`? That type can then be incorporated into the ng-dev `GithubConfig` type.

dev-infra/utils/git.ts
```diff
@@ -6,15 +6,128 @@
...
+  /** Instance of the authenticated Github octokit API. */
+  api: Octokit;
+  /** The file path of project's root directory. */
+  private _projectRoot = getRepoBaseDir();
+
+  /** Regular expression that matches the provided Github token. */
+  private _tokenRegex = new RegExp(this._githubToken, 'g');
+
+  constructor(
+      private _githubToken = '', private _config: Pick<NgDevConfig, 'github'> = getConfig()) {
```
devversion, line 59: Can you add add a comment to `GitClient` that explains that a token is optional? Also it would be good mentioning inside there that by default the ng-dev config configured github remote is used.

devversion (approved):
LGTM. Just two minor comments (also as a note: lint seems to fail)

devversion (approved):
LGTM. Re-approving since the Github check shows me as pending again

atscott:
I think this is also blocked by: https://github.com/angular/angular/pull/37422. Merging to the patch branch failed.

PR #37325, opened 2020-05-28
fix(compiler-cli): attach incremental driver only in watch mode

ayazhafiz:
An incremental driver may be very large, and a provider of the program
passed to the Angular compiler may not wish for a driver to be attached
to the provided program. This commit attaches the compiler's incremental
driver to a `ts.Program` iff the compiler is told to run in a watch-like
mode. A watch-like mode is either an Angular watch compilation, invoked
by the `--watch` CLI flag, or a `"watch": true` configuration in the
Angular compiler options. The latter configuration does not start a
watch compilation, but instructs the compiler to optimize the
compilation as if it were in watch mode (i.e. attach an incremental
driver).

## PR Checklist

## PR Type

- [x] Bugfix

- [x] Yes

Previously all compilations would attach an incremental driver by default. Now, only compilations in watch mode attach an incremental driver by default. In practice I don't think this will be impactful since the incremental driver is not useful in 1-pass compilations (i.e. outside watch mode).

### DIFF
--- a/packages/compiler-cli/test/ngtsc/incremental_error_spec.ts
+++ b/packages/compiler-cli/test/ngtsc/incremental_error_spec.ts
@@ -21,7 +21,7 @@ runInEachFileSystem(() => {
     beforeEach(() => {
       env = NgtscTestEnvironment.setup(testFiles);
       env.enableMultipleCompilations();
-      env.tsconfig();
+      env.tsconfig({watch: true});
 
       // This file is part of the program, but not referenced by anything else. It can be used by
       // each test to verify that it isn't re-emitted after incremental builds.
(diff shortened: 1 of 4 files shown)


kyliau (approved):

packages/compiler-cli/test/ngtsc/incremental_error_spec.ts
```diff
@@ -21,7 +21,7 @@ runInEachFileSystem(() => {
     beforeEach(() => {
       env = NgtscTestEnvironment.setup(testFiles);
       env.enableMultipleCompilations();
-      env.tsconfig();
+      env.tsconfig({watch: true});
```
alxhub, line 24: `enableMultipleCompilations` should turn this on, it should not be necessary (or possible) to hack the tsconfig to do this.

ayazhafiz: let's drop this pr in favor of yours

Issue #34197, opened 2019-12-02
UrlTree created by Router.createUrlTree might incorrectly have fragment === undefined


Ferdy89:
# 🐞 bug report

### Affected Package

The issue is caused by package `@angular/router`, particularly by the method `Router.createUrlTree`.

### Is this a regression?

To my knowledge, no.

### Description

When creating a [`UrlTree`](https://github.com/angular/angular/blob/0cd843169884a57d92c1ec114f7e778720b5ef24/packages/router/src/url_tree.ts#L106-L129) using [`Router.createUrlTree`](https://github.com/angular/angular/blob/0cd843169884a57d92c1ec114f7e778720b5ef24/packages/router/src/router.ts#L925-L952) without passing any fragments, the resulting `UrlTree` contains a `fragment === undefined`.  This does not conform to [its type signature](https://github.com/angular/angular/blob/0cd843169884a57d92c1ec114f7e778720b5ef24/packages/router/src/url_tree.ts#L118).

## 🔬 Minimal Reproduction

https://stackblitz.com/edit/angular-az3elw

## 🌍  Your Environment

**Angular Version:** 8.0.0

Issue #29683, opened 2019-04-03
Directive `routerLink` dont correctly work with empty `fragment` option


Dok11:
# 🐞 bug report

### Description

When `fragment` of `routerLink` are empty the anchor save `#` in `href`.
But expected will remove `#` character for use clean url.

So, for fix this issue I need use two diffrent temaplates of one link: with fragment and without.
I think this is not pretty decision.

And I think the solution located something here: https://github.com/angular/angular/blob/7.2.x/packages/router/src/url_tree.ts#L293

Now:
```ts
const fragment =
        typeof tree.fragment === `string` ? `#${encodeUriFragment(tree.fragment !)}` : '';
```

But it should be something like this:
```ts
const fragment = (typeof tree.fragment === `string` && tree.fragment)
    ? `#${encodeUriFragment(tree.fragment !)}` : '';
```

## 🔬 Minimal Reproduction
```ts
<!-- correct link with fragment -->
<a routerLink="/abc" fragment="myFragment">link 1 with fragment</a>

<!-- bug with empty fragment -->
<a routerLink="/def" fragment="">link 2 with blank fragment</a>
```

https://stackblitz.com/edit/angular-puzmfb?file=src%2Fapp%2Fapp.component.html

![image](https://user-images.githubusercontent.com/2697890/55480773-b0b7e380-5629-11e9-9b6d-dee24c9a2d7e.png)

alan-agius4:
Hi, this seems like a framework issue and such bugs should be logged here: https://github.com/angular/angular

Thanks.

Dok11:
@alan-agius4 I dont feel difference. You provide link to this same repo

alan-agius4:
Opps sorry @Dok11 my mistake. I got confused.

jasonaden:
Thanks for the issue. Looks valid. Would you be interested in creating a pull request to fix this?

[Developer docs](https://github.com/angular/angular/blob/master/docs/DEVELOPER.md) can be found here that explain how to set up your environment.

lbassin:
Is anyone doing a PR for this issue ? Otherwise i have some free time and can do it

atscott:
Closing this as working as intended. Empty string fragments are valid and act the same as writing `<a href="#">navigate to empty fragment</a>`. Additionally, `[fragment]="null"` works as I would expect by removing the fragment completely.

PR #37334, opened 2020-05-28
fix(router): fragment must be a string

### LINKED ISSUE #23894: Type declaration for ActivatedRoute.fragment doesn't declare possible null values
## I'm submitting a...

<pre><code>
[x] Bug report  
[ ] Performance issue
[ ] Feature request
[x] Documentation issue or request
[ ] Other... Please describe:
</code></pre>

## Current behavior

The type declaration for `ActivatedRoute.fragment` declares that it will emit string instances and nothing else.

```ts
interface ActivatedRoute {
  fragment: Observable<string>
}
```

To my surprise, it emits `null` values for routes that don't have a fragment.

## Expected behavior

I see three valid options here (in no particular order):
1. emit `''` instead of `null` for missing fragments
2. don't emit at all for missing fragments
3. just change the type declaration to `Observable<string | null>`

https://angular-nullable-fragment.stackblitz.io

image

## Environment

<pre><code>
Angular version: 5.2.9

Browser:
- [x] Chrome (desktop) version 65
- [x] Firefox version 59

Others:

</code></pre>

MatthiasKunnen:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

`ActivatedRoute.fragment` is typed as `Observable<string>` but can emit both `null` and `undefined` due to incorrect non-null assertion. 

Issue Number: Fixes #23894

## What is the new behavior?
The non-null assertions have been removed and fragment will now be an empty string instead of null. This behavior is consistent with [window.location.hash](https://html.spec.whatwg.org/multipage/history.html#dom-location-hash).

- [x] Yes
- [ ] No

ActivatedRoute.fragment will no longer be null when the URL does not have a fragment. Difference for fragment values:
    
|              | Old  | New |
|--------------|------|-----|
| https://e.g  | `null` | `''`  |
| https://e.g# | `''`   | `''`  |

Applications should check for an empty string instead of `null` in order to determine that no fragment is set.

### DIFF
--- a/packages/router/src/apply_redirects.ts
+++ b/packages/router/src/apply_redirects.ts
@@ -79,7 +79,7 @@ class ApplyRedirects {
         this.expandSegmentGroup(this.ngModule, this.config, this.urlTree.root, PRIMARY_OUTLET);
     const urlTrees$ = expanded$.pipe(
         map((rootSegmentGroup: UrlSegmentGroup) => this.createUrlTree(
-                rootSegmentGroup, this.urlTree.queryParams, this.urlTree.fragment!)));
+                rootSegmentGroup, this.urlTree.queryParams, this.urlTree.fragment)));
     return urlTrees$.pipe(catchError((e: any) => {
       if (e instanceof AbsoluteRedirect) {
         // after an absolute redirect we do not apply any more redirects!
@@ -101,7 +101,7 @@ class ApplyRedirects {
         this.expandSegmentGroup(this.ngModule, this.config, tree.root, PRIMARY_OUTLET);
     const mapped$ = expanded$.pipe(
         map((rootSegmentGroup: UrlSegmentGroup) =>
-                this.createUrlTree(rootSegmentGroup, tree.queryParams, tree.fragment!)));
+                this.createUrlTree(rootSegmentGroup, tree.queryParams, tree.fragment)));
     return mapped$.pipe(catchError((e: any): Observable<UrlTree> => {
       if (e instanceof NoMatch) {
         throw this.noMatchError(e);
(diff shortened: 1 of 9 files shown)


atscott:
Hi @MatthiasKunnen, as I mentioned in the issue report, I think the correct fix is to update the type rather than convert null to empty string. https://github.com/angular/angular/issues/23894#issuecomment-635549846

MatthiasKunnen:
I was already making this change and missed your comment.

Two questions about changing the type:
- Should `undefined` be converted to `null`?
-Wwhat value should fragment have in this case: `https://e.g/#`? `null` or empty string?

MatthiasKunnen:
Might be interesting, according to the [spec](https://html.spec.whatwg.org/multipage/history.html#dom-location-hash) internally, the URL fragment is `string | null` as is `UrlTree`'s. However, publicly, it is an empty string in case of `null`. Something might be said about consistency with the HTML spec. The decision is yours of course.

atscott:
Undefined should be converted to null. You can also mark your or as fixing this: https://github.com/angular/angular/issues/34197

The answer to your second question I think is a bit harder to answer. I personally think that should be empty string( see  https://github.com/angular/angular/issues/29683#issuecomment-635499263 ).  What’s the current behavior?

MatthiasKunnen:
Current behavior is an empty string

MatthiasKunnen:
I propose I close this PR and make a new one since this PR's intentions have changed and it might get confusing. Agreed?

atscott:
> I propose I close this PR and make a new one since this PR's intentions have changed and it might get confusing. Agreed?

Yea, that sounds good. Thanks!

PR #37339, opened 2020-05-28
refactor(compiler-cli): make IncrementalBuild strategy configurable

alxhub:
Commit 24b2f1da2b introduced an `NgCompiler` which operates on a
`ts.Program` independently of the `NgtscProgram`. The NgCompiler got its
`IncrementalDriver` (for incremental reuse of Angular compilation results)
by looking at a monkey-patched property on the `ts.Program`.

This monkey-patching operation causes problems with the Angular indexer
(specifically, it seems to cause the indexer to retain too much of prior
programs, resulting in OOM issues). To work around this, `IncrementalDriver`
reuse is now handled by a dedicated `IncrementalBuildStrategy`. One
implementation of this interface is used by the `NgtscProgram` to perform
the old-style reuse, relying on the previous instance of `NgtscProgram`
instead of monkey-patching. Only for `NgTscPlugin` is the monkey-patching
strategy used, as the plugin sits behind an interface which only provides
access to the `ts.Program`, not a prior instance of the plugin.

### DIFF
--- a/packages/compiler-cli/BUILD.bazel
+++ b/packages/compiler-cli/BUILD.bazel
@@ -27,6 +27,7 @@ ts_library(
         "//packages/compiler-cli/src/ngtsc/core:api",
         "//packages/compiler-cli/src/ngtsc/diagnostics",
         "//packages/compiler-cli/src/ngtsc/file_system",
+        "//packages/compiler-cli/src/ngtsc/incremental",
         "//packages/compiler-cli/src/ngtsc/indexer",
         "//packages/compiler-cli/src/ngtsc/perf",
         "//packages/compiler-cli/src/ngtsc/typecheck",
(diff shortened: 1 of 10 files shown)


kyliau (approved):
LGTM, will have to update the constructor call sites in language service as well
(line 34 and 47 below):

https://github.com/angular/angular/blob/df082307b9fe8e76c05730155df16b09c5ac32ac/packages/language-service/ivy/compiler/compiler.ts#L34-L47

alxhub:
[Presubmit](http://test/OCL:314215486:BASE:314215545:1591050533090:1dcda7a9)

atscott:
Hi @alxhub, it looks like there are several failures on CI. I'm removing the merge label and adding cleanup. Please take a look and add back the merge label when ready.

IgorMinar (dismissed):
can you please get the test to pass? thanks

alxhub:
[Presubmit](http://test/OCL:314215486:BASE:315960282:1591905637006:5dc15a3d)

AndrewKushnir (approved):

alxhub:
Caretaker: this change should not require further approval.

PR #37341, opened 2020-05-29
fix(platform-server): correctly handle absolute relative URLs

### LINKED ISSUE #37314: platform-server - Changes introduced in 10.0.0-next.8 to convert relative URLs to absolute URLs breaks SSR
# 🐞 bug report

### Affected Package
The issue is caused by package @angular/platform-server

### Is this a regression?

Yes, the previous version in which this bug was not present was: 10.0.0-next.6

### Description
I recently updated my app from Angular `10.0.0-next.6` to `10.0.0-rc.0`, after which I noticed strange behavior in SSR requests. After the update it broke the SSR for my app. I downgraded `@angular/platform-server` to  `10.0.0-next.6` and everything started working fine.

I have an http interceptor, which:

1. Converts the request url from relative to absolute urls in `intercept` method of  my `http interceptor`.
2. When on server, when I receive a response of the http request I add it to `this.transferState.set(CACHE_KEY, event as any);` and on client, I check if we already have the cached response, its served from cache otherwise request is sent to the server.

After the update if I open `http://localhost:53189` in browser, which pulls data from `https://www.mustakbil.com/ws/home/index`, request seems to be working fine.

But if I open any other url e.g. `http://localhost:53189/jobs/job/204182`, when I open it in my browser the request goes to server which should fetch data from `https://www.mustakbil.com/ws/jobs/job/204182`, render it and return final response to client. 
But this doesn't happen, and the SSR is broken.

In my component I make `http` request to `/ws/jobs/job/204182` and then in intercept method of my interceptor I convert it to `https://www.mustakbil.com/ws/jobs/job/204182`.

After https://github.com/angular/angular/pull/37071 pull request , if I `console log(req.url)` in `intercept` method of my interceptor, before any modification to the request it shows `/jobs/job/204182/ws/jobs/job/204182`, whereas I expect it to show `/ws/jobs/job/204182` because that's what I passed from my component using this code `this.http.get("/ws/jobs/job/204182")`.
 
## 🔬 Minimal Reproduction
I can add anyone from Angular team to a private repo.

## 🔥 Exception or Error
<pre><code>
SSR is broken since the server is sending request to wrong url.
</code></pre>

## 🌍  Your Environment

**Angular Version:**
<pre><code>
Angular CLI: 10.0.0-rc.0
Node: 12.16.2
OS: win32 x64

Angular: 10.0.0-rc.0
... platform-server, router, service-worker
Ivy Workspace: <error>

Package                                    Version
--------------------------------------------------------------------
@angular-devkit/architect                  0.901.7
@angular-devkit/build-angular              0.1000.0-rc.0
@angular-devkit/build-optimizer            0.901.7
@angular-devkit/build-webpack              0.901.7
@angular-devkit/core                       10.0.0-rc.0
@angular-devkit/schematics                 10.0.0-rc.0
@angular/cdk                               10.0.0-next.1
@angular/fire                              6.0.0
@angular/google-maps                       10.0.0-next.1
@angular/material                          10.0.0-next.1
@angular/pwa                               0.1000.0-rc.0
@ngtools/webpack                           9.1.7
@nguniversal/module-map-ngfactory-loader   9.0.0-next.9
@schematics/angular                        10.0.0-rc.0
@schematics/update                         0.1000.0-rc.0
rxjs                                       6.5.5
typescript                                 3.9.3
webpack                                    4.41.2
</code></pre>

**Anything else relevant?**

I am using `aspnet-prerendering` package to create `createServerRenderer`, should it cause issue?

CaerusKaru:
Previously, we would simply prepend any relative URL with the HREF
for the current route (pulled from document.location). However,
this does not correctly account for the leading slash URLs that
would otherwise be parsed correctly in the browser, or the
presence of a base HREF in the DOM.

Therefore, we use the built-in URL implementation for NodeJS,
which implements the WHATWG standard that's used in the browser.
We also pull the base HREF from the DOM, falling back on the full
HREF as the browser would, to form the correct request URL.

Fixes #37314

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: 37314

## What is the new behavior?

Add support for base HREF and absolute relative URLs on SSR HTTP requests.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/platform-server/src/http.ts
+++ b/packages/platform-server/src/http.ts
@@ -10,13 +10,12 @@
 const xhr2: any = require('xhr2');
 
 import {Injectable, Injector, Provider} from '@angular/core';
-import {DOCUMENT} from '@angular/common';
+import {PlatformLocation} from '@angular/common';
 import {HttpEvent, HttpRequest, HttpHandler, HttpBackend, XhrFactory, ɵHttpInterceptingHandler as HttpInterceptingHandler} from '@angular/common/http';
 import {Observable, Observer, Subscription} from 'rxjs';
 
 // @see https://www.w3.org/Protocols/HTTP/1.1/draft-ietf-http-v11-spec-01#URI-syntax
 const isAbsoluteUrl = /^[a-zA-Z\-\+.]+:\/\//;
-const FORWARD_SLASH = '/';
 
 @Injectable()
 export class ServerXhr implements XhrFactory {
@@ -105,20 +104,18 @@ export abstract class ZoneMacroTaskWrapper<S, R> {
 
 export class ZoneClientBackend extends
     ZoneMacroTaskWrapper<HttpRequest<any>, HttpEvent<any>> implements HttpBackend {
-  constructor(private backend: HttpBackend, private doc: Document) {
+  constructor(private backend: HttpBackend, private platformLocation: PlatformLocation) {
     super();
   }
 
   handle(request: HttpRequest<any>): Observable<HttpEvent<any>> {
-    const href = this.doc.location.href;
-    if (!isAbsoluteUrl.test(request.url) && href) {
-      const urlParts = Array.from(request.url);
-      if (request.url[0] === FORWARD_SLASH && href[href.length - 1] === FORWARD_SLASH) {
-        urlParts.shift();
-      } else if (request.url[0] !== FORWARD_SLASH && href[href.length - 1] !== FORWARD_SLASH) {
-        urlParts.splice(0, 0, FORWARD_SLASH);
-      }
-      return this.wrap(request.clone({url: href + urlParts.join('')}));
+    const {href, protocol, hostname} = this.platformLocation;
+    if (!isAbsoluteUrl.test(request.url) && href !== '/') {
+      const baseHref = this.platformLocation.getBaseHrefFromDOM() || href;
+      const urlPrefix = `${protocol}//${hostname}`;
+      const baseUrl = new URL(baseHref, urlPrefix);
+      const url = new URL(request.url, baseUrl);
+      return this.wrap(request.clone({url: url.toString()}));
     }
     return this.wrap(request);
   }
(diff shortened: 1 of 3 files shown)


packages/platform-server/src/http.ts
```diff
@@ -105,20 +104,20 @@ export abstract class ZoneMacroTaskWrapper<S, R> {
...
-    const href = this.doc.location.href;
-    if (!isAbsoluteUrl.test(request.url) && href) {
-      const urlParts = Array.from(request.url);
-      if (request.url[0] === FORWARD_SLASH && href[href.length - 1] === FORWARD_SLASH) {
-        urlParts.shift();
-      } else if (request.url[0] !== FORWARD_SLASH && href[href.length - 1] !== FORWARD_SLASH) {
-        urlParts.splice(0, 0, FORWARD_SLASH);
-      }
-      return this.wrap(request.clone({url: href + urlParts.join('')}));
+    const href = this.platformLocation.href;
```
alan-agius4, line 112: I find it cleaner to use spread  in here Ex
```ts
    const { href, hostname, protocol } = this.platformLocation;
    if (!isAbsoluteUrl.test(request.url) && href !== '/') {
      const baseHref = this.platformLocation.getBaseHrefFromDOM() || href;
      const urlPrefix = `${protocol}//${hostname}`;
      const baseUrl = new URL(baseHref, urlPrefix);
      const url = new URL(request.url, baseUrl);
      return this.wrap(request.clone({ url: url.toString() }));
    }
    return this.wrap(request);
  }
```

packages/platform-server/src/http.ts
```diff
@@ -105,20 +104,20 @@ export abstract class ZoneMacroTaskWrapper<S, R> {
...
+    if (!isAbsoluteUrl.test(request.url) && href !== '/') {
+      const hostname = this.platformLocation.hostname;
+      const protocol = this.platformLocation.protocol;
+      const baseHref = this.platformLocation.getBaseHrefFromDOM() || href;
```
alan-agius4, line 116: getBaseHrefFromDOM` can return an empty string, in which case it will be falsy in which case we should respect it.

alan-agius4 (approved):
LGTM, one small nit.

packages/platform-server/src/http.ts
```diff
@@ -10,13 +10,12 @@
 const xhr2: any = require('xhr2');
 
 import {Injectable, Injector, Provider} from '@angular/core';
-import {DOCUMENT} from '@angular/common';
+import {PlatformLocation} from '@angular/common';
 import {HttpEvent, HttpRequest, HttpHandler, HttpBackend, XhrFactory, ɵHttpInterceptingHandler as HttpInterceptingHandler} from '@angular/common/http';
 import {Observable, Observer, Subscription} from 'rxjs';
 
 // @see https://www.w3.org/Protocols/HTTP/1.1/draft-ietf-http-v11-spec-01#URI-syntax
 const isAbsoluteUrl = /^[a-zA-Z\-\+.]+:\/\//;
```
alxhub, line 18: Note that this regex also doesn't cover `//foo.com/bar` URLs, which are absolute paths. I suspect they might work by accident, as they'll be treated as "rooted" by the first `/`.

alxhub (approved):

alxhub:
[Presubmit](http://test/OCL:314157781:BASE:314157830:1591033268095:7a16508d)

packages/platform-server/src/http.ts
alan-agius4: @CaerusKaru can you do a followup PR for this please?

PR #37346, opened 2020-05-29
feat(dev-infra): ensure that merge tooling always loads latest config

josephperrott:
Ensures that when the merge script runs, the commit being loaded is
the same commit as upstream/master.  If it is not the same commit,
it checks out the latest from upstream/master and reloads the config.

### DIFF
--- a/dev-infra/pr/merge/config.ts
+++ b/dev-infra/pr/merge/config.ts
@@ -7,6 +7,7 @@
  */
 
 import {getConfig, NgDevConfig} from '../../utils/config';
+import {GitClient} from './git';
 
 import {GithubApiMergeStrategyConfig} from './strategies/api-merge';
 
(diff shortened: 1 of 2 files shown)


dev-infra/pr/merge/config.ts
```diff
@@ -111,6 +112,21 @@ export function loadAndValidateConfig(): {config?: MergeConfigWithRemote, errors
...
   }
 
+  // Ensure that the currently checked out branch is the latest from upstream/master.
+  const git = new GitClient('', '', mergeConfig as MergeConfigWithRemote);
+  const githubUrl = `https://github.com/${config.github.owner}/${config.github.name}.git`
+  // Fetch the latest SHA from github and compare it to the current SHA of HEAD.
+  git.runGraceful(['fetch', githubUrl]);
+  const upstreamMasterSha = git.runGraceful(['rev-parse', 'FETCH_HEAD']).stdout.trim();
+  const currentSha = git.runGraceful(['rev-parse', 'HEAD']).stdout.trim();
+  if (currentSha !== upstreamMasterSha) {
```
devversion, line 122: Hmm. I'm not a fan of this detached head as it _could_ conflict with actual merge strategies. I think just failing (as initially discussed) is not good either as it will be inconvenient when a PR is merged. The merge script does not refresh the local `master` branch after merge. I still don't think we should refresh local branches afterwards as it could conflict with local changes.

I'm not sure what's best. We probably should think/chat more about it.

PR #37361, opened 2020-05-30
fix(compiler): Expression converter typo

### LINKED ISSUE #21233: Expression converter typo
## I'm submitting a...

<pre><code>
[x] Bug report  
[ ] Feature request
</code></pre>

## Current behavior
There is a typo in expression_converter.ts
[link](https://github.com/angular/angular/blob/bc7a6d7b0074b0c845035b99d05a831d286b60e4/packages/compiler/src/compiler_util/expression_converter.ts#L340)

```
    // For literal values of null, undefined, true, or false allow type inteference
    // to infer the type.
    const type =
        ast.value === null || ast.value === undefined || ast.value === true || ast.value === true ?
        o.INFERRED_TYPE :
        undefined;
```
```
ast.value === true || ast.value === true
```
## Expected behavior

```
ast.value === true || ast.value === false
```

Just a typo

## Environment
Any
<pre><code>
Angular version: 5.1.2

Browser:
- [ ] Chrome (desktop) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX
 
For Tooling issues:
- Node version: XX  
- Platform:  

Others:

</code></pre>

ajitsinghkaler:
In expression converter for literal values of null, undefined, true, or false type interference was to be done but it was not done for value false. Added false also to the check.

Fixes #21233

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

a typo
Issue Number: #21233

## What is the new behavior?
typo removed

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/compiler/src/compiler_util/expression_converter.ts
+++ b/packages/compiler/src/compiler_util/expression_converter.ts
@@ -467,7 +467,7 @@ class _AstToIrVisitor implements cdAst.AstVisitor {
     // For literal values of null, undefined, true, or false allow type interference
     // to infer the type.
     const type =
-        ast.value === null || ast.value === undefined || ast.value === true || ast.value === true ?
+        ast.value === null || ast.value === undefined || ast.value === true || ast.value === false ?
         o.INFERRED_TYPE :
         undefined;
     return convertToStatementIfNeeded(


pkozlowski-opensource:
We _definitively_ need an additional test (one that would fail without the code change and starts passing with the code change)

ajitsinghkaler:
I don't see any unit tests added for the method visitLiteralPrimitive() do I write an integration test

petebacondarwin:
Ideally there should be a test of some kind to demonstrate this.

AndrewKushnir (commented):
@ajitsinghkaler, thanks for creating this PR. As Pawel and Pete mentioned, it'd be great to have a test to verify this change. The code was originally introduced in 
https://github.com/angular/angular/commit/82bcd8356600b118b33b681bb7ef9d20ba128bf4 and there are some tests there, so may be you can use one of them as a base for the new test.

I'm also adding @alxhub as a reviewer as he may have additional context on this.

Thank you.

ajitsinghkaler:
@AndrewKushnir I tried to write the test but could not understand how to write a test for this can you please help me.

AndrewKushnir:
Hi @ajitsinghkaler, I tried to reuse one of the tests from https://github.com/angular/angular/commit/82bcd8356600b118b33b681bb7ef9d20ba128bf4 and was able to trigger the codepath updated in this PR, but I was unable to come up with a case when a test would fail with current state of the master branch and will pass after the change in this PR (which is needed to validate that everything works as needed). May be @petebacondarwin or @alxhub have additional thoughts/ideas. Thank you.

petebacondarwin:
I think the reason we can't make a failing test is that our typing checks tend not to distinguish between `null`, `undefined` and `false` (or even `0`!!).

See https://github.com/angular/angular/blob/master/packages/compiler/src/output/output_ast.ts#L149

JoostK:
@petebacondarwin but the type would be `o.INFERRED_TYPE`. I _think_ a test that might currently fail would be the following scenario:

```html
<button (click)="doSomething(false)"></button>
```
```ts
export class MyCmp {
  doSomething(param: number) {}
}
```

One would expect this not to typecheck, als literal `false` should be inferred to `boolean` instead of just `any`.

Note that since Ivy, any type information on the output AST is not used at all.

petebacondarwin:
I created a template that looks like this:

```
<div *myIf="person.display != false"> {{person.name}} </div>
```

where `person.display` is of type `false`.

The rendered type check block is:

```
function _View_MainComp_1_0():void {
  var _any:any = (null as any);
  const currVal_0:any = (_decl0_0.person.display != false);
  currVal_0;
}
```

I believe that because `false` has been assigned `null` for its type, it gets cast to a "dynamic" type, which then causes the result of the comparison to also become "dynamic", hence the `currVal_0:any`.

petebacondarwin:
I'll have a play with a function call as in your suggestion @JoostK .

petebacondarwin:
So a template like:

```html
<button [id]="computeId(false)"></button>
```

doesn't trigger the `visitLiteralPrimitive()` method here at all!

JoostK:
After looking into it some more, I can't think of a way to reproduce a bug here.

The type is only relevant if a temporary has to be introduced, which only happens for the LHS of safe navigation operators. Therefore, the literal itself is never assigned into a temporary. In the TS emitter, the type is only used when the literal is `null`:

https://github.com/angular/angular/blob/f150a304e8509f731836da4d6d28db93ad39bb54/packages/compiler/src/output/ts_emitter.ts#L96-L103

For the boolean case, the literal will be emitted as is. Therefore, the literal type is retained.

Alas, I don't think the code with the bug has a real purpose.

ajitsinghkaler:
@JoostK so should we remove this code

petebacondarwin:
I think that despite no test (and possibly despite this code being used!!) we should land this change since clearly the current code is not correct.

jelbourn:
Closing this since it's been several months- feel free to open a new PR that either cleans up the unnecessary code or adds tests for a behavior change.

PR #37363, opened 2020-05-30
fix(ngcc): do not inline source-maps for non-inline typings source-maps

### LINKED ISSUE #37324: ngcc generate inline sourcemaps in `.d.ts` files and break completely DX
# 🐞 bug report

### Affected Package
The issue is caused by package @angular/ALL

### Is this a regression?
I think yes, the previous version in which this bug was not present was: I don't know, maybe `8.x.x`.

### Description

When `ngcc` run it appends a massive source-map at the end of definition files (`.d.ts`) in every angular package, and it gets stuck my code editor (`vscode`). I have to restart `vscode` to kill the `tsserver`, and avoid to enter to any `angular`'s file definition.

## 🔬 Minimal Reproduction
`ng new ...`, `npm run start` (to run `ngcc`) and follow, for example, `Component` definition to its `.d.ts` file at `@angular/core`. VSCode and `ng-language-service` take a lot of time to process the file and consume a lot of CPU.

## 🔥 Exception or Error
image

## 🌍  Your Environment

**Angular Version:**
<pre><code>
Angular CLI: 9.1.6
Node: 13.7.0
OS: linux x64

Angular: 9.1.9
... localize, platform-browser, platform-browser-dynamic, router
Ivy Workspace: Yes

@angular-devkit/architect         0.901.1
@angular-devkit/build-angular     0.901.6
@angular-devkit/build-optimizer   0.901.6
@angular-devkit/build-webpack     0.901.6
@angular-devkit/core              8.3.25
@angular-devkit/schematics        8.3.25
@angular/cdk                      9.2.4
@angular/cli                      9.1.6
@ngtools/webpack                  9.1.6
@schematics/angular               9.1.6
@schematics/update                0.901.6
    
Done in 0.58s.

</code></pre>

**Anything else relevant?**
image

petebacondarwin:
Inline source-maps in typings files can impact IDE performance
so ngcc should only add such maps if the original typings file
contains inline source-maps.

Fixes #37324

### DIFF
--- a/packages/compiler-cli/ngcc/src/rendering/source_maps.ts
+++ b/packages/compiler-cli/ngcc/src/rendering/source_maps.ts
@@ -42,8 +42,11 @@ export function renderSourceAndMap(
 
     const rawMergedMap: RawSourceMap = generatedFile.renderFlattenedSourceMap();
     const mergedMap = fromObject(rawMergedMap);
-    if (generatedFile.sources[0]?.inline) {
-      // The input source-map was inline so make the output one inline too.
+    const firstSource = generatedFile.sources[0];
+    if (firstSource && firstSource.inline &&
+        (firstSource.rawMap !== null || !sourceFile.isDeclarationFile)) {
+      // We render an inline source map if either the input source-map was inline or there was no
+      // input source map and this is not a typings file.
       return [
         {path: generatedPath, contents: `${generatedFile.contents}\n${mergedMap.toComment()}`}
       ];
--- a/packages/compiler-cli/ngcc/src/rendering/source_maps.ts
+++ b/packages/compiler-cli/ngcc/src/rendering/source_maps.ts
@@ -43,10 +43,15 @@ export function renderSourceAndMap(
     const rawMergedMap: RawSourceMap = generatedFile.renderFlattenedSourceMap();
     const mergedMap = fromObject(rawMergedMap);
     const firstSource = generatedFile.sources[0];
-    if (firstSource && firstSource.inline &&
-        (firstSource.rawMap !== null || !sourceFile.isDeclarationFile)) {
-      // We render an inline source map if either the input source-map was inline or there was no
-      // input source map and this is not a typings file.
+    if (firstSource && (firstSource.rawMap !== null || !sourceFile.isDeclarationFile) &&
+        firstSource.inline) {
+      // We render an inline source map if one of:
+      // * there was no input source map and this is not a typings file;
+      // * the input source map exists and was inline.
+      //
+      // We do not generate inline source maps for typings files unless there explicitly was one in
+      // the input file because these inline source maps can be very large and it impacts on the
+      // performance of IDEs that need to read them to provide intellisense etc.
       return [
         {path: generatedPath, contents: `${generatedFile.contents}\n${mergedMap.toComment()}`}
       ];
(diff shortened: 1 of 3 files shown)


packages/compiler-cli/ngcc/test/rendering/renderer_spec.ts
```diff
@@ -640,6 +640,25 @@ UndecoratedBase.ɵdir = ɵngcc0.ɵɵdefineDirective({ type: UndecoratedBase, vie
              expect(mapFile.path).toEqual(_('/node_modules/test-package/src/file.js.map'));
              expect(JSON.parse(mapFile.contents)).toEqual(MERGED_OUTPUT_PROGRAM_MAP.toObject());
            });
+
+
+        it('should render an internal source map for files whose original file does not have a source map',
+           () => {
+             const sourceFiles: TestFile[] =
+                 [{name: JS_CONTENT.name, contents: JS_CONTENT.contents}];
```
gkalpak, line 648: Why not `[JS_CONTENT]` or `[{...JS_CONTENT}]`?

packages/compiler-cli/ngcc/src/rendering/source_maps.ts
```diff
@@ -42,8 +42,11 @@ export function renderSourceAndMap(
 
     const rawMergedMap: RawSourceMap = generatedFile.renderFlattenedSourceMap();
     const mergedMap = fromObject(rawMergedMap);
-    if (generatedFile.sources[0]?.inline) {
-      // The input source-map was inline so make the output one inline too.
+    const firstSource = generatedFile.sources[0];
+    if (firstSource && firstSource.inline &&
+        (firstSource.rawMap !== null || !sourceFile.isDeclarationFile)) {
```
gkalpak, line 47: I am a little confused: How can `inline` be `true` if `rawMap` is `null`?

packages/compiler-cli/ngcc/src/rendering/source_maps.ts
```diff
@@ -42,8 +42,11 @@ export function renderSourceAndMap(
...
+      // We render an inline source map if either the input source-map was inline or there was no
+      // input source map and this is not a typings file.
```
gkalpak, line 49: We should expand a bit on why typinfs files are treated differently.

gkalpak (changes requested):

packages/compiler-cli/ngcc/test/rendering/renderer_spec.ts
petebacondarwin: Oh yes!

packages/compiler-cli/ngcc/src/rendering/source_maps.ts
petebacondarwin: I guess this is a limitation of the sourcemaps library, I built. Normally the `inline` value will be decided by how the source file's source-map was brought in. In the case that there is no source-map in the original file, I had to assign a `true` or `false` value and opted for the default of `true` if there was no source-map.

Do you have a suggestion of how this could be better?

packages/compiler-cli/ngcc/src/rendering/source_maps.ts
petebacondarwin: Will do.

packages/compiler-cli/ngcc/src/rendering/source_maps.ts
petebacondarwin: I will re-order the condition and enhance the comment.

gkalpak: I see. If there are three states (inline, external, and none), an enum would work better imo, but that is out of scope for this PR.
An enhanced comment works for me :+1:

gkalpak (approved):
LGTM :+1:

PR #29083, opened 2019-03-03
fix(ivy): platform module bootstrap does not resolve resources 

devversion:
* See individual commits.

PR #37372, opened 2020-06-01
perf(core): avoid pulling in jit-specific code in aot bundles

crisbeto:
In #29083 a call to `getCompilerFacade` was added to `ApplicationRef` which pulls in a bit of JIT-specific code. Since the code path that calls the function can't be hit for an AOT-compiled app, these changes add an `ngJitMode` guard which will allow for dead code elimination to drop it completely. Testing it out against a new CLI project showed a difference of ~1.2kb.

### DIFF
--- a/goldens/size-tracking/integration-payloads.json
+++ b/goldens/size-tracking/integration-payloads.json
@@ -3,7 +3,7 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 1485,
-        "main-es2015": 142590,
+        "main-es2015": 141394,
         "polyfills-es2015": 36963
       }
     }
@@ -21,7 +21,7 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 1485,
-        "main-es2015": 148510,
+        "main-es2015": 147314,
         "polyfills-es2015": 36963
       }
     }
@@ -30,7 +30,7 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 1485,
-        "main-es2015": 138196,
+        "main-es2015": 136873,
         "polyfills-es2015": 37640
       }
     }
@@ -39,7 +39,7 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 2289,
-        "main-es2015": 247942,
+        "main-es2015": 246781,
         "polyfills-es2015": 36657,
         "5-es2015": 751
       }
@@ -49,7 +49,7 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 2289,
-        "main-es2015": 223603,
+        "main-es2015": 221961,
         "polyfills-es2015": 36657,
         "5-es2015": 779
       }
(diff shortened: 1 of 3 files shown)


packages/core/src/application_ref.ts
```diff
@@ -86,7 +86,7 @@ export function compileNgModuleFactory__POST_R3__<M>(
   // there won't be any resource loader. This can happen with Ivy, because AOT compiled
   // modules can be still passed through "bootstrapModule". In that case we shouldn't
   // unnecessarily require the JIT compiler.
-  if (compilerProviders.length === 0) {
+  if (compilerProviders.length === 0 || (typeof ngJitMode !== 'undefined' && !ngJitMode)) {
```
devversion, line 89: Out of curiosity: Did you consider doing this before `isComponentResourceResolutionQueueEmpty`? That might avoid bringing in the `isComponentResourceResolutionQueueEmpty` function?

JoostK: Yep, I think we can do better than this. The `compilerOptions`/`compilerProviders` is not used in AOT mode, so the JIT guard could span more code.

crisbeto: Nice 👍 . Changed.

JoostK (approved):
💯

devversion (approved):

goldens/size-tracking/integration-payloads.json
```diff
@@ -61,8 +61,8 @@
         "bundle": "TODO(i): temporarily increase the payload size limit from 105779 - this is due to a closure issue related to ESM reexports that still needs to be investigated",
         "bundle": "TODO(i): we should define ngDevMode to false in Closure, but --define only works in the global scope.",
         "bundle": "TODO(i): (FW-2164) TS 3.9 new class shape seems to have broken Closure in big ways. The size went from 169991 to 252338",
-        "bundle": "TODO(i): after removal of tsickle from ngc-wrapped / ng_package, we had to switch to SIMPLE optimizations which increased the size from 252338 to 1197869, see PR#37221 and PR#37317 for more info",
-        "bundle": 1197869
+        "bundle": "TODO(i): after removal of tsickle from ngc-wrapped / ng_package, we had to switch to SIMPLE optimizations which increased the size from 252338 to 1197896, see PR#37221 and PR#37317 for more info",
+        "bundle": 1197896
```
IgorMinar, line 65: isn't it odd that this one went up? not a blocker, just weird.

IgorMinar (approved):
LGTM! Thank you!

goldens/size-tracking/integration-payloads.json
crisbeto: This one was also smaller the first time around, but I had to resolve a merge conflict.

mhevery (approved):

atscott:
@crisbeto As discussed, switching this to master only as there are conflicts with the patch branch. Please create a follow-up for patch only

PR #37378, opened 2020-06-01
build: remove wombot proxy registry from package.jsons for release

josephperrott:
Due to an outage with the proxy we rely on for publishing, we need
to temporarily directly publish to NPM using our own angular
credentials again.


matsko (approved):

IgorMinar (approved):
LGTM! Thank you!

PR #37382, opened 2020-06-01
fix(compiler-cli): downlevel angular decorators to static properties

### LINKED ISSUE #30106: forwardRef breaks at runtime with ES2015
# 🐞 bug report

### Affected Package

### Is this a regression?

 No.

### Description

Using `forwardRef` as described in https://angular.io/api/core/forwardRef while targeting ES2015 results on a `Uncaught ReferenceError: Cannot access 'Lock' before initialization` runtime error.

## 🔬 Minimal Reproduction
- make a new project `ng new forward-ref-project && cd forward-ref-project`
- ensure `tsconfig.json` contains `"target": "es2015",`
- replace the contents of `src/main.ts` with:
```
import { Inject, forwardRef, ReflectiveInjector } from '@angular/core';
class Door {
  lock: Lock;

  // Door attempts to inject Lock, despite it not being defined yet.
  // forwardRef makes this possible.
  constructor(@Inject(forwardRef(() => Lock)) lock: Lock) { this.lock = lock; }
}

// Only at this point Lock is defined.
class Lock { }

const injector = ReflectiveInjector.resolveAndCreate([Door, Lock]);
const door = injector.get(Door);
console.log(door instanceof Door);
console.log(door.lock instanceof Lock);
```
- `ng serve -o`

## 🔥 Exception or Error
```
Uncaught ReferenceError: Cannot access 'Lock' before initialization
    at Module../src/main.ts (main.ts:21)
    at __webpack_require__ (bootstrap:78)
    at Object.2 (main.ts:30)
    at __webpack_require__ (bootstrap:78)
    at checkDeferredModules (bootstrap:45)
    at Array.webpackJsonpCallback [as push] (bootstrap:32)
    at main.js:1
```

## 🌍  Your Environment

**Angular Version:**
<pre><code>
Angular CLI: 8.0.0-beta.18
Node: 10.10.0
OS: win32 x64
Angular: 8.0.0-beta.14
... router

@angular-devkit/architect         0.800.0-beta.18
@angular-devkit/build-angular     0.800.0-beta.18
@angular-devkit/build-optimizer   0.800.0-beta.18
@angular-devkit/build-webpack     0.800.0-beta.18
@angular-devkit/core              8.0.0-beta.18
@angular-devkit/schematics        8.0.0-beta.18
@angular/cli                      8.0.0-beta.18
@ngtools/webpack                  8.0.0-beta.18
@schematics/angular               8.0.0-beta.18
@schematics/update                0.800.0-beta.18
typescript                        3.4.5
webpack                           4.30.0
</code></pre>

**Anything else relevant?**

`forwardRef` is provided specifically for the purpose of referencing something that isn't defined. This is useful in breaking [circular dependencies](https://angular.io/guide/dependency-injection-in-action#break-circularities-with-a-forward-class-reference-forwardref), and when declaring both services and components in the [same file](https://angular.io/guide/dependency-injection#create-and-register-an-injectable-service).

`forwardRef` works because it delays the resolution of the reference to a time at which it is already declared through the callback indirection. In the [API example](https://angular.io/api/core/forwardRef), the symbol we want to delay resolution is `Lock`:
```
class Door {
  lock: Lock;

  // Door attempts to inject Lock, despite it not being defined yet.
  // forwardRef makes this possible.
  constructor(@Inject(forwardRef(() => Lock)) lock: Lock) { this.lock = lock; }
}

// Only at this point Lock is defined.
class Lock { }
```

But `Lock` is actually being referenced in more places than just inside `forwardRef`. It is also being used as a TS type in the class property, and in the constructor parameter.

Types don't usually have a runtime representation so that shouldn't be a problem. But constructor types are an exception and actually do have a runtime representation. We can see this by looking at the transpiled code:

```
import * as tslib_1 from "tslib";
import { Inject, forwardRef, ReflectiveInjector } from '@angular/core';
let Door = class Door {
    // Door attempts to inject Lock, despite it not being defined yet.
    // forwardRef makes this possible.
    constructor(lock) { this.lock = lock; }
};
Door = tslib_1.__decorate([
    tslib_1.__param(0, Inject(forwardRef(() => Lock))),
    tslib_1.__metadata("design:paramtypes", [Lock])
], Door);
// Only at this point Lock is defined.
class Lock {
}
const injector = ReflectiveInjector.resolveAndCreate([Door, Lock]);
const door = injector.get(Door);
console.log(door instanceof Door);
console.log(door.lock instanceof Lock);
```

The `Lock` type in the for the constructor parameter was transpiled into `tslib_1.__metadata("design:paramtypes", [Lock])`. This reference does not have a delayed resolution like the injected `forwardRef` and is instead immediately resolved, resulting in `Uncaught ReferenceError: Cannot access 'Lock' before initialization`.

This error isn't observed when targetting ES5 however. We can understand why by looking at the code when transpiled to ES5 :
```
import * as tslib_1 from "tslib";
import { Inject, forwardRef, ReflectiveInjector } from '@angular/core';
var Door = /** @class */ (function () {
    // Door attempts to inject Lock, despite it not being defined yet.
    // forwardRef makes this possible.
    function Door(lock) {
        this.lock = lock;
    }
    Door = tslib_1.__decorate([
        tslib_1.__param(0, Inject(forwardRef(function () { return Lock; }))),
        tslib_1.__metadata("design:paramtypes", [Lock])
    ], Door);
    return Door;
}());
// Only at this point Lock is defined.
var Lock = /** @class */ (function () {
    function Lock() {
    }
    return Lock;
}());
var injector = ReflectiveInjector.resolveAndCreate([Door, Lock]);
var door = injector.get(Door);
console.log(door instanceof Door);
console.log(door.lock instanceof Lock);
```

In ES5 there are no `class` declarations, so TS instead uses a `var`. One important different between `var` and `class/let/const` is that the latter are all subject to the [Temporal Dead Zone](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let#Temporal_dead_zone). 

In practical terms the TDZ means that using a `var` before it is declared resolves to `undefined`, but using a `class/let/const` instead throws a `ReferenceError`. This is the error we are seeing here.

A possible workaround is to **not** declare the type in the constructor:
```
// Instead of adding the type in the parameter
constructor(@Inject(forwardRef(() => Lock)) lock: Lock) { 
  this.lock = lock;
}

// Add it as a cast in the constructor body
constructor(@Inject(forwardRef(() => Lock)) lock) {
  this.lock = lock as Lock;
}
```

This will change the transpiled code and remove the reference, avoiding the `ReferenceError`:
```
Door = tslib_1.__decorate([
    tslib_1.__param(0, Inject(forwardRef(() => Lock))),
    tslib_1.__metadata("design:paramtypes", [Object])
                                             ^^^^^^ was Lock before
], Door);
```

One important note is that the `ReferenceError` does not come up on Angular CLI projects compiled with AOT. This is because there we actually transform transpiled TS code and remove Angular decorators, so the metadata reference (`tslib_1.__metadata("design:paramtypes", [Lock])`) never reached the browser and thus there is no `ReferenceError`.

### LINKED ISSUE #30141: property decorators break at runtime with ES2015 and circular deps
# 🐞 bug report

### Affected Package

### Is this a regression?

 No

### Description

 Using the `@Input()` property decorator (or any other property decorators I imagine) targetting ES2015 with a circular dependency on a type will cause a `Uncaught ReferenceError` at runtime.

## 🔬 Minimal Reproduction
```
git clone https://github.com/filipesilva/istiti-circular-deps-es2015
cd istiti-circular-deps-es2015
ng serve
# src/app/autocomplete.directive.ts contains the problematic decorator and circular dep
```

## 🔥 Exception or Error
<pre><code>
main.js:9545 Uncaught ReferenceError: Cannot access 'AutocompleteAbstract' before initialization
    at Module.AutocompleteAbstract (main.js:9545)
    at Module../src/app/autocomplete.directive.ts (main.js:9580)
    at __webpack_require__ (runtime.js:79)
    at Module../src/app/autocomplete.abstract.ts (main.js:9546)
    at __webpack_require__ (runtime.js:79)
    at Module../src/app/parent.component.ts (main.js:9605)
    at __webpack_require__ (runtime.js:79)
    at Module../src/app/app.module.ts (main.js:9507)
    at __webpack_require__ (runtime.js:79)
    at Module../src/main.ts (main.js:9666)
</code></pre>

## 🌍  Your Environment

**Angular Version:**
<pre><code>
Angular CLI: 8.0.0-beta.16
Node: 10.10.0
OS: win32 x64
Angular: 8.0.0-beta.13
... router

@angular-devkit/architect         0.800.0-beta.16
@angular-devkit/build-angular     0.800.0-beta.16
@angular-devkit/build-optimizer   0.800.0-beta.16
@angular-devkit/build-webpack     0.800.0-beta.16
@angular-devkit/core              8.0.0-beta.16
@angular-devkit/schematics        8.0.0-beta.16
@angular/cli                      8.0.0-beta.16
@ngtools/webpack                  8.0.0-beta.16
@schematics/angular               8.0.0-beta.16
@schematics/update                0.800.0-beta.16
typescript                        3.4.4
webpack                           4.30.0
</code></pre>

**Anything else relevant?**

cc @istiti

This is similar to https://github.com/angular/angular/issues/30106. In this case it is not a constructor parameter type, but rather a class property type.

A class property by itself doesn't need metadata with a type reference:
```
// transpiled src/app/autocomplete.directive.ts without Input() decorator
import * as tslib_1 from "tslib";
import { Directive } from '@angular/core';
let AutocompleteDirective = class AutocompleteDirective {
    constructor() { }
};
// this is used in autocomplete.abstract.ts
AutocompleteDirective.EMPTYLIST_ELEMENT = "_EMPTYLIST_";
AutocompleteDirective = tslib_1.__decorate([
    Directive({
        selector: '[appAutocomplete]'
    }),
    tslib_1.__metadata("design:paramtypes", [])
], AutocompleteDirective);
export { AutocompleteDirective };
```

But when the property is decorated, the metadata will contain a reference to the type:
```
// transpiled src/app/autocomplete.directive.ts with Input() decorator
import * as tslib_1 from "tslib";
import { Directive, Input } from '@angular/core';
import { AutocompleteAbstract } from './autocomplete.abstract';
let AutocompleteDirective = class AutocompleteDirective {
    constructor() { }
};
// this is used in autocomplete.abstract.ts
AutocompleteDirective.EMPTYLIST_ELEMENT = "_EMPTYLIST_";
tslib_1.__decorate([
    Input('appAutocomplete'),
    tslib_1.__metadata("design:type", AutocompleteAbstract)
], AutocompleteDirective.prototype, "autocomplete", void 0);
AutocompleteDirective = tslib_1.__decorate([
    Directive({
        selector: '[appAutocomplete]'
    }),
    tslib_1.__metadata("design:paramtypes", [])
], AutocompleteDirective);
export { AutocompleteDirective };
```

In ES5 there are no `class` declarations, so TS instead uses a `var`. One important different between `var` and `class/let/const` is that the latter are all subject to the [Temporal Dead Zone](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let#Temporal_dead_zone). 

Without the `Input()` decorator this isn't a problem because there is no circular dependency on the type.

### LINKED ISSUE #30586: Libraries sometimes depend on tsickle decorator downleveling
# 🐞 bug report

### Affected Package

`@angular/compiler-cli` / `@angular/bazel`

### Is this a regression?

No

### Description

Angular applications which are built with NGC but aren't setting up `tsickle` for decorator downleveling (e.g. don't set the `annotationsAs: "static fields"`), can't be rendered on the server if DOM-specific types are referenced on class members which are decorated.

e.g.

```ts
@Directive(...)
export class MyDirective {
  @Input() triggerElement: HTMLElement;
}
```

This results in the following ES5 JavaScript output if `tsickle` is **not enabled**

```js
    __decorate([
        core_1.Input(),
        __metadata("design:type", HTMLElement)
    ], MyDirective.prototype, "triggerElement");
```
Notice how the `__decorate` call tries to access `HTMLELement` from a global context. This causes an exception when the following file is executed in NodeJS since DOM APIs are not available. We mock most of the DOM APIs with `domino` but providing these mocks globally so that such failures can be avoided is not common right now.

Building the project **with tsickle** fixes the problem automatically as decorators are downleveled into a static field while `tsickle` also discards the type reference to `HTMLElement` (as it is not relevant for Angular).

```js
MyDirective.propDecorators = {
  triggerElement: [{ type: core_1.Input }]
};
```

TL;DR: 

1) Everything currently works as expected within the CLI since it provides optimizations that removes the decorators anyway when using AOT mode or build-optimizer.
2) It becomes an issue if a project is not built with the CLI / and does not use build-optimizer. Previously tsickle was part of the NGC workflow **by default**.. so it now causes unexpected failures. Needs documentation? See: https://github.com/angular/angular/commit/29761ea5f8e4bae68442236f4a0abc1666487a35

**How is Angular Bazel affected?**

With Bazel it is not possible to opt-in for tsickle's decorator downleveling. Tsickle can be enabled by passing the `bazelOptions: {tsickle: true}`, but that comes with more tsickle transformations which aren't desired. e.g. type imports are rewritten to `goog.requireType`.

## 🔬 Minimal Reproduction

https://github.com/devversion/ngc-without-tsickle

## 🔥 Exception or Error
<pre><code>__metadata("design:type", HTMLElement)
                                  ^
ReferenceError: HTMLElement is not defined
</code></pre>

## 🌍  Your Environment

**Angular Version:**
<pre><code>Latest: 8.0.0-rc.4</code></pre>

**Anything else relevant?**

devversion:
fix(compiler-cli): downlevel angular decorators to static properties

In v7 of Angular we removed `tsickle` from the default `ngc` pipeline.
This had the negative potential of breaking ES2015 output and SSR due
to a limitation in TypeScript.

TypeScript by default preserves type information for decorated constructor
parameters when `emitDecoratorMetadata` is enabled. For example,
consider this snippet below:

```ts
@Directive()
export class MyDirective {
  constructor(button: MyButton) {}
}

export class MyButton {}
```

TypeScript would generate metadata for the `MyDirective` class it has
a decorator applied. This metadata would be needed in JIT mode, or
for libraries that provide `MyDirective` through NPM. The metadata would
look as followed:

```js
let MyDirective = class MyDir {}

MyDirective = __decorate([
  Directive(),
  __metadata("design:paramtypes", [MyButton]),
], MyDirective);

let MyButton = class MyButton {}
```

Notice that TypeScript generated calls to `__decorate` and
`__metadata`. These calls are needed so that the Angular compiler
is able to determine whether `MyDirective` is actually an directive,
and what types are needed for dependency injection.

The limitation surfaces in this concrete example because `MyButton`
is declared after the `__metadata(..)` call, while `__metadata`
actually directly references `MyButton`. This is illegal though because
`MyButton` has not been declared at this point. This is due to the
so-called temporal dead zone in JavaScript. Errors like followed will
be reported at runtime when such file/code evaluates:

```js
Uncaught ReferenceError: Cannot access 'MyButton' before initialization
```

As noted, this is a TypeScript limitation because ideally TypeScript
shouldn't evaluate `__metadata`/reference `MyButton` immediately.
Instead, it should defer the reference until `MyButton` is actually
declared. This limitation will not be fixed by the TypeScript team
though because it's a limitation as per current design and they will
only revisit this once the tc39 decorator proposal is finalized
(currently stage-2 at time of writing).

Given this wontfix on the TypeScript side, and our heavy reliance on
this metadata in libraries (and for JIT mode), we intend to fix this
from within the Angular compiler by downleveling decorators to static
properties that don't need to evaluate directly. For example:

```js
MyDirective.ctorParameters = () => [MyButton];
```

With this snippet above, `MyButton` is not referenced directly. Only
lazily when the Angular runtime needs it. This mitigates the temporal
dead zone issue caused by a limitation in TypeScript's decorator
metadata output. See: https://github.com/microsoft/TypeScript/issues/27519.

In the past (as noted; before version 7), the Angular compiler by
default used tsickle that already performed this transformation. We
moved the transformation to the CLI for JIT and `ng-packager`, but now
we realize that we can move this all to a single place in the compiler
so that standalone ngc consumers can benefit too, and that we can
disable tsickle in our Bazel `ngc-wrapped` pipeline (that currently
still relies on tsickle to perform this decorator processing).

This transformation also has another positive side-effect of making
Angular application/library code more compatible with server-side
rendering. In principle, TypeScript would also preserve type information
for decorated class members (similar to how it did that for constructor
parameters) at runtime. This becomes an issue when your application
relies on native DOM globals for decorated class member types. e.g.

```ts
@Input() panelElement: HTMLElement;
```

Your application code would then reference `HTMLElement` directly
whenever the source file is loaded in NodeJS for SSR. `HTMLElement`
does not exist on the server though, so that will become an invalid
reference. One could work around this by providing global mocks for
these DOM symbols, but that doesn't match up with other places where
dependency injection is used for mocking DOM/browser specific symbols.

More context in this issue: #30586. The TL;DR here is that the Angular
compiler does not care about types for these class members, so it won't
ever reference `HTMLElement` at runtime.

Fixes #30106. Fixes #30586. Fixes #30141.
Resolves FW-2196. Resolves FW-2199.


goldens/size-tracking/aio-payloads.json
```diff
@@ -21,8 +21,8 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 3097,
-        "main-es2015": 429710,
-        "polyfills-es2015": 52195
+        "main-es2015": 431366,
+        "polyfills-es2015": 52655
```
devversion, line 25: * `main-es2015` previously was at `429841`, so this is an increase of `1.525kb`.
* `polyfills-es2015` remained as is. No change. Just updated it to reflect `master`.

I need to look into it and do some comparisons.

devversion: Resolved by not using a TS trick to retain ctor parameter imports as done in CLI. The trick has the negative side-effect of preserving full imports (with potentially unused specifiers).

devversion:
@alxhub @IgorMinar I marked this as `master-only` because I think it's not critical to go into V10 given that we have re-enabled tsickle in `ng_package`. We surely could get it into v10 as a fix if desired 😄

alan-agius4:
I think it’s best that this goes in v10, because otherwise we’ll be unable to remove this logic from ng-packagr and CLI until v11, that’s because ng-packagr and CLI are typically compatible for a particular major version of the framework.

Note: the transformer needs to be added as part of the public api, because in the CLI this transformer is only used when using JIT mode, which in CLI happens via TSC.

devversion:
@alan-agius4 Yeah, we chatted about that being exposed. I wonder how we can do that as compiler code never is public API as per `PUBLIC_API.md`. How has it been done in the past for similar things?

alan-agius4:
Since the compiler code is not bundled we can opt to use deep imports in this case.

alan-agius4 (approved):
LGTM.

IgorMinar (changes requested):
Overall this looks really really good @devversion!

I left a few comments and I'm going to rely on @alxhub's review of the transformer code itself(I skipped that part during this review pass).

The biggest sticking point for me is `emitDecoratorMetadata`- see my longer comment about that.

With regards to your comment about "master-only" vs "master&patch". I think that we should get this in v10 to enable simplification of the downstream tooling as soon as possible. So please relabel the PR.

Lastly, and I'm very hesitant to even suggest it this late in the RC, but I do have to bring it up because otherwise we'll have to go through more churn and migrations: how about we rename `.decorators` and `ctorParams` to `ɵdec` and `ɵctor` (or something similar prefixed with `ɵ`). The goal of this would be to ensure that these props are clearly marked as private APIs and that we follow the same conventions for these apis as we do elsewhere.

The property name change needs more discussion, so please do not act on it, but think about it and we can discuss it during the FW sync meeting. 

(one known downstream change that would need to happen as a result of this renaming would be to update the CLI/build-optimizer passes to look for both the old and new property names, @alan-agius4 could weigh in on how difficult that would be and if there is more to it than an extra property lookup in AST)

alan-agius4:
Renaming the props is not a big change from the CLI POV, I think the framework will be impacted more, because the compiler & ngcc will need to be aware of these new annotations for decorators.

Currently the compiler/cli/ngcc can process/reflect decorators that are either emitted either by tsickle or TypeScript.

packages/compiler-cli/integrationtest/tsconfig-build.json
```diff
@@ -12,6 +12,7 @@
     "baseUrl": ".",
     "declaration": true,
     "experimentalDecorators": true,
+    "emitDecoratorMetadata": true,
```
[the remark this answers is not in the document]
devversion: Let's discuss this below in the overall `emitDecoratorMetadata` comment.

packages/compiler-cli/test/ngc_spec.ts
```diff
@@ -654,9 +646,42 @@ describe('ngc transformer command-line', () => {
...
         expect(mymoduleSource).toContain('args: [{ declarations: [] },] }');
         expect(mymoduleSource).not.toContain(`__metadata`);
-        expect(mymoduleSource).toContain(`import { AClass } from './aclass';`);
-        expect(mymoduleSource).toContain(`{ type: AClass }`);
+        expect(mymoduleSource).not.toContain(`import { AClass } from './aclass';`);
+        expect(mymoduleSource).not.toContain(`{ type: AClass }`);
       });
+
+      it('should add metadata as static fields with type information if ' +
+             '`emitDecoratorMetadata` is enabled',
```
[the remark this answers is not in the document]
devversion: That is a good question. I've been thinking about this the last days a lot, and there are various reasons why this ended up like that:

TypeScript has a limitation in transformers that doesn't allow us to preserve imports without the `emitDecoratorMetadata` option. We need to preserve some imports that previously were type-only, but now are used as values in `ctorParameters`. TypeScript does refresh it's type checking state after transformers, so TypeScript will start eliding these imports incorrectly. e.g.

```ts
import {MatButton} from '@angular/material/button';

@Directive()
export class Test {
  constructor(button: MatButton) {}
}
```

In the example above: `MatButton` is type-only as per TypeScript's analysis. That is correct.. but not when we run our transformer that accesses `MatButton` as a value in `Test.ctorParameters = <..>`.

There are tricks that we can use to ensure that these imports are not elided. Though these have downsides. e.g. it could cause other unused imports to be preserved in the output too. In fact I had such trick applied in the initial version of this PR, but these retained imports can cause significant size increase as applications are not by default marked as `sideEffects: false`, and Webpack would bring in unnecessary chunks that are unused when build optimizer removes `ctorParameters`.

The CLI runs into a similar limitation when it removes decorators. It has a custom transformer for cleaning up unused imports later.. though it seems contradictory to implement such one in ngc too as it would basically mean that we re-implement TypeScript's logic for determining/eliding unused imports. We could certainly do that in NGC though.

An alternative workaround I came up with (one that will not prevent TS from eliding unused imports). This one would be the best if it wouldn't rely on some TS internals. It basically tells TypeScript that the given type symbols are now referenced as values, and their imports need to be retained. This matches the exact logic `emitDecoratorMetadata` has in TypeScript. That is actually why conceptually `emitDecoratorMetadata` needs to be set for type information.

This is aligned with how TypeScript emits decorators/metadata. e.g. by default, decorators are just downleveled, but if `emitDecoratorMetadata` is set, we also include type information.

The thinking here is that applications by default run in AOT mode and don't have `emitDecoratorMetadata` set (as done in CLI's).  In those cases, they wouldn't actually need the type information, nor the downleveled decorators. The CLI has a transformer for deleting decorators. Libraries on the other side want to include type information and would set `emitDecoratorMetadata`. That makes their library compatible for consumption in AOT and JIT.

Angular w/ tsickle worked around this in the past by just always setting `emitDecoratorMetadata = true` if decorators needed to be downleveled. These are the options from what I can tell:

1. Use the TS trick that keeps imports, but prevents eliding of other unused ones. We would then need another transform for deleting unused imports.. Basically re-implementing TypeScript's logic..
  The transformer would then most likely diverge a lot from the original one in tsickle.
2. Use the workaround that tells TypeScript about imported symbols that shouldn't be elided. No need to re-implement logic for deleting unused imports. Could be prone to internal changes in TypeScript as it relies on non-public API (TS doesn't export all the good things needed for good AST transformation!)
3. Always set `emitDecoratorMetadata = true` in NGC. This would match with what [we did before v7](https://github.com/angular/angular/blob/6.0.0/packages/compiler-cli/src/main.ts#L52) when tsickle still ran by default. I'm not sure if that has any impacts these days.
4. Require libraries or JIT consumers to explicitly set `emitDecoratorMetadata` when they want type information for DI included. That matches with how TypeScript emits metadata/decorator. 

I initially used (2) and it worked like a charm, but it was rather more like an experiment to see if we can do better than what has been done in the past of Angular with tsickle. Generally I'd definitely also prefer that we always include type information as that would be the most predictable way.

packages/bazel/src/ngc-wrapped/index.ts
```diff
@@ -277,12 +259,23 @@ export function compile({
...
 
+  // Always disable tsickle decorator transforming in the tsickle compiler host. The
+  // Angular compilers have their own logic for decorator processing and we wouldn't
+  // want tsickle to interfere with that.
+  bazelHost.transformDecorators = false;
+
+  // By default, annotations for closure compiler are disabled in prodmode,
+  // unless the workspace is known to be `google3`.
+  if (!bazelOpts.es5Mode && compilerOpts.annotateForClosureCompiler === undefined) {
+    compilerOpts.annotateForClosureCompiler = bazelOpts.workspaceName === 'google3';
```
alxhub, line 270: I prefer the former `if` statement that explicitly set `= true` or `= false` depending on the workspace name. This is trickier to follow.

packages/compiler-cli/test/ngc_spec.ts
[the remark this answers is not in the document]
alxhub: I think `ngc` (View Engine) should continue to assume `emitDecoratorMetadata` under the same circumstances that it did before. If `annotationsAs !== 'decorators'`, we turn on both our transform and `emitDecoratorMetadata`. I believe the absence of this logic is what's breaking g3 presubmit, since ngc_wrapped sets `annotationsAs = 'static fields'`.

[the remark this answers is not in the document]
clydin: > Yeah, that sounds like option (3). Definitely not an ideal solution, but one with the fewest tricks required. One issue here would be that projects end up with type metadata for custom decorators, even though they had emitDecoratorMetadata = false. The CLI and build optimizer will only strip Angular decorators.

I think this is could turn out to be particularly problematic for libraries with custom decorators.  By leaving the metadata, the underlying TS issue could be triggered causing these libraries to no longer function.  I think `emitDecoratorMetadata` should be avoided completely if possible since its design is fundamentally flawed in ES2015+ code.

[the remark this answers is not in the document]
devversion: @clydin Agreed. That's what I was referring too. Please feel free to also weigh in on the various other options being discussed here. I think your feedback would be appreciated!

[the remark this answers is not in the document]
alxhub: I think the right approach is a modification of option 3:

`emitDecoratorMetadata` is automatically set by `ngc` unless the user has explicitly set it to `false`.

If `emitDecoratorMetadata` is set to `false` explicitly, no transformation happens - we emit the decorators as TS would have done.

If `emitDecoratorMetadata` is set (or coerced) to `true`, we add our transform and generate the full suite of metadata that `tsickle` used to.

As far as Bazel goes, its behavior should then be no different than plain `ngc`. We set `annotateForClosureCompiler` in `ngc_wrapped`, which causes `ngc` to emit via `tsickle`. Decorators are transformed internally (not in `tsickle`) because `emitDecoratorMetadata` will be assumed as above.

Does that sound reasonable? This way if applications/consumers decide they don't want decorator metadata, they can disable it explicitly.

packages/compiler-cli/src/transformers/program.ts
```diff
@@ -99,8 +93,7 @@ class AngularCompilerProgram implements Program {
   private _structuralDiagnostics: Diagnostic[]|undefined;
   private _programWithStubs: ts.Program|undefined;
   private _optionsDiagnostics: Diagnostic[] = [];
-  // TODO(issue/24571): remove '!'.
-  private _reifiedDecorators!: Set<StaticSymbol>;
+  private _transformTsDiagnostics: ts.Diagnostic[] = [];
```
alxhub, line 96: This is cleanly done, but sadly I think it's futile. This array only gets populated during `emit`, long after any diagnostics have been retrieved.

If you look at how this works with diagnostics coming from tsickle, the compiler drops them on the floor (it pays no attention to the return value of `emitWithTsickle`). We can just do the same.

packages/compiler-cli/test/ngc_spec.ts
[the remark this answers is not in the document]
clydin: > If emitDecoratorMetadata is set to false explicitly, no transformation happens - we emit the decorators as TS would have done.

Wouldn't that break JIT mode for libraries that cannot enable the option?

packages/bazel/src/ngc-wrapped/index.ts
IgorMinar: funny. I actually thought the same. Then I tried to rewrite the code and ended up with just suggesting that the comment above is clarified. I'm fine with making this an explicit `true` / `false` assignment to improve the readability of the code.

devversion: I don't have anything opposed to changing it back to the nested `if`. Seems more verbose for something super simple, but I'd agree that it's slightly more readable.

packages/compiler-cli/test/ngc_spec.ts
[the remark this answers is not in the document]
devversion: I had a long chat with Alex yesterday about this. We acknowledged that (3) and (4) are not good options as they would break libraries that use custom decorators. e.g.

```ts
@CustomDecorator()
export class SomeClass {
  constructor(button: Button) {}
}

export class Button {}
```

This currently could be shipped as part of libraries with ng-packagr, but when we'd require `ng-packagr` to enable `emitDecoratorMetadata`, it will break because the constructor type would be captured in the vanilla TS metadata and the [TypeScript temporal dead zone](https://github.com/microsoft/TypeScript/issues/27519) would surface.

Options remaining are (1) and (2). We are going with a mix of (1) and (2) where we properly tell TypeScript to preserve imports, but do not need a custom post-transform (as in the CLI) that cleans up unused imports. We can fully rely on TypeScript cleaning up unused/type-only imports. And the best thing is that we do not even need to rely on an internal TS API (though on some internal understanding of how the compiler works)

I will push these changes today. We should be able to get another review on this then.

packages/compiler-cli/src/transformers/program.ts
devversion: That's a fair point. I think we could incorporate these diagnostics into the emit result diagnostics, but I'm not sure if we even would want to report diagnostics (given that tsickle dropped them on the floor too).

If we'd now start printing diagnostics that could signify a breaking change. Changing it to ignore these diagnostics. Let me know if you think we should handle them as part of the emit result.

packages/compiler-cli/src/transformers/downlevel_decorators_transform.ts
```diff
@@ -0,0 +1,660 @@
...
+    }
+
+    return (sf: ts.SourceFile) => {
+      // Ensure that referenced type symbols are not elided by TypeScript. Imports for
+      // such parameter type symbols previously could be type-only, but now might be also
+      // used in the `ctorParameters` static property as a value. We want to make sure
+      // that TypeScript does not elide imports for such type references. Read more
+      // about this in the description for `patchAliasReferenceResolution`.
+      if (patchAliasReferenceResolution(
+              context as TransformationContextWithResolver, referencedParameterTypes) === false) {
```
clydin, line 651: I think `referencedParameterTypes` should be cleared per source file invocation.
The transformer can technically be reused to process more than one file.  TS doesn't appear to currently when using a `ts.Program` but it is setup to do so.
https://github.com/microsoft/TypeScript/blob/093e08396d6175520aad303dbacc32d7970e9fcd/src/compiler/transformer.ts#L137-L147

packages/compiler-cli/src/transformers/downlevel_decorators_transform.ts
```diff
@@ -0,0 +1,660 @@
...
+/**
+ * Gets a transformer for downleveling Angular decorators.
+ * @param typeChecker Reference to the program's type checker.
+ * @param diagnostics List which will be populated with diagnostics if any.
+ * @param host Reflection host that is used for determining decorators.
+ * @param isCore Whether the current TypeScript program is for the `@angular/core` package.
+ * @param isClosureCompilerEnabled Whether closure annotations need to be added where needed.
+ */
+export function getDownlevelDecoratorsTransform(
+    typeChecker: ts.TypeChecker, diagnostics: ts.Diagnostic[], host: ReflectionHost,
```
clydin, line 421: Use of `ReflectionHost` may be problematic for CLI integration.  This transformer will only be used in JIT there and AOT compiler functionality is not present in that mode.

packages/compiler-cli/src/transformers/downlevel_decorators_transform.ts
```diff
@@ -0,0 +1,660 @@
...
+ * resolver exposed. There are requests upstream in TypeScript to expose
+ * that as public API: https://github.com/microsoft/TypeScript/issues/17516..
+ */
+type TransformationContextWithResolver = ts.TransformationContext&{
+  getEmitResolver?: () => EmitResolver;
+};
+
+/** Describes a subset of the TypeScript internal emit resolver. */
+type EmitResolver = {
+  isReferencedAliasDeclaration(node: ts.Node, checkChildren?: boolean): boolean;
```
alxhub, line 31: This should be an optional method as well - we should be able to detect if it's not present.

packages/compiler-cli/src/transformers/downlevel_decorators_transform.ts
```diff
@@ -0,0 +1,660 @@
...
+ * matches conceptually with the logic that runs internally in TypeScript when the
+ * `emitDecoratorMetadata` flag is enabled. TypeScript basically surfaces the same problem and
+ * solves it conceptually the same way, but obviously doesn't need to access an `@internal` API.
+ *
+ * See:
+ * https://sourcegraph.com/github.com/microsoft/TypeScript@3eaa7c65f6f076a08a5f7f1946fd0df7c7430259/-/blob/src/compiler/checker.ts#L31219-31257
+ *
+ * @returns A boolean indicating whether the alias reference solution could be patched properly.
+ */
+function patchAliasReferenceResolution(
```
alxhub, line 77: 1) Move this function into a separate file, so it's not lost in the rest of the transform code.
2) Instead of returning `boolean`, let's have it throw the error itself. We can rename it to `patchAliasReferenceResolutionOrDie`.
3) The error should be user-friendly and should explain that the version of TS is incompatible with the Angular compiler and they should downgrade (presumably, they're seeing it for the first time after updating TS).

packages/compiler-cli/src/transformers/downlevel_decorators_transform.ts
```diff
@@ -0,0 +1,660 @@
...
+ *   2. https://github.com/microsoft/TypeScript/issues/17516.
+ *   3. https://github.com/angular/tsickle/issues/635.
+ *
+ * The patch we apply to tell TypeScript about actual referenced aliases (i.e. imported symbols),
+ * matches conceptually with the logic that runs internally in TypeScript when the
+ * `emitDecoratorMetadata` flag is enabled. TypeScript basically surfaces the same problem and
+ * solves it conceptually the same way, but obviously doesn't need to access an `@internal` API.
+ *
+ * See:
+ * https://sourcegraph.com/github.com/microsoft/TypeScript@3eaa7c65f6f076a08a5f7f1946fd0df7c7430259/-/blob/src/compiler/checker.ts#L31219-31257
```
alxhub, line 73: Link directly to Github and not to sourcegraph.

packages/compiler-cli/src/transformers/downlevel_decorators_transform.ts
```diff
@@ -0,0 +1,660 @@
...
+ *
+ * @returns A boolean indicating whether the alias reference solution could be patched properly.
+ */
+function patchAliasReferenceResolution(
+    context: TransformationContextWithResolver, referencedAliases: Set<ts.Declaration>): boolean {
```
alxhub, line 78: The public API of this function should accept a `ts.TransformationContext` and cast internally.

alxhub (changes requested):

clydin (changes requested):
Nice job integrating this.

Couple additional comments:
* I think the property name changes are a good idea but agree with @IgorMinar that they can wait.
* Along with the renaming, I think that static methods should be used instead of static properties where appropriate (e.g., `ctorParameters` which is really a static method in disguise).  All static properties need to be down-leveled so their use is giving more work to the compiler during each build/rebuild.
* How will the integration into the CLI for JIT support look?

packages/compiler-cli/src/transformers/downlevel_decorators_transform.ts
devversion: The reflection host should just rely on the type-checker and we just use it for determining the origin of decorators (so we don't need to re-implement import resolution again). I don't think that should be an issue.

packages/compiler-cli/src/transformers/downlevel_decorators_transform.ts
```diff
@@ -0,0 +1,660 @@
...
+     */
+    function decoratorDownlevelVisitor(node: ts.Node): ts.Node {
+      switch (node.kind) {
+        case ts.SyntaxKind.ClassDeclaration: {
+          return transformClassDeclaration(node as ts.ClassDeclaration);
+        }
+        default:
+          return ts.visitEachChild(node, decoratorDownlevelVisitor, context);
+      }
+    }
```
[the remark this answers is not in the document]
devversion: Sure. I'll go through an clean up all the original tsickle code, but note that as per filecomment, this is the _original_ tsickle code with small modifications (no clean ups)

packages/compiler-cli/src/transformers/downlevel_decorators_transform.ts
```diff
@@ -0,0 +1,660 @@
...
+      // that TypeScript does not elide imports for such type references. Read more
+      // about this in the description for `patchAliasReferenceResolution`.
+      if (patchAliasReferenceResolution(
+              context as TransformationContextWithResolver, referencedParameterTypes) === false) {
+        throw Error('Unable to downlevel Angular decorators. Emit resolver could not be patched.');
+      }
+      // Downlevel decorators and constructor parameter types. We will keep track of all
+      // referenced constructor parameter types so that we can instruct TypeScript to
+      // not elide their imports if they previously were only type-only.
+      return decoratorDownlevelVisitor(sf) as ts.SourceFile;
```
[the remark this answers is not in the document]
devversion: Same as below. I just ported this from the tsickle transform w/ small modifications (no cleanup). I'll clean this up though.

packages/compiler-cli/src/transformers/downlevel_decorators_transform.ts
```diff
@@ -0,0 +1,660 @@
...
+        const decoratorNode = decorator.node! as ts.Decorator;
+        if (!isAngularDecorator(decorator, isCore)) {
+          decoratorsToKeep.push(decoratorNode);
+          continue;
+        }
+        toLower.push(decoratorNode);
+      }
+      if (!toLower.length) return [undefined, element, []];
+
+      if (!element.name || element.name.kind !== ts.SyntaxKind.Identifier) {
```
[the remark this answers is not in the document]
devversion: Same as below. I intentionally didn't clean things up, as the goal was to keep modifications minimal in the tsickle port. I'll clean this up based on discussion with Alex though.

packages/compiler-cli/src/transformers/downlevel_decorators_transform.ts
```diff
@@ -0,0 +1,660 @@
...
+
+import * as ts from 'typescript';
+import {Decorator, ReflectionHost} from '../ngtsc/reflection/src/host';
+
+/**
+ * Whether a given decorator should be treated as an Angular decorator.
+ * Either it's used in @angular/core, or it's imported from there.
+ */
+function isAngularDecorator(decorator: Decorator, isCore: boolean): boolean {
+  return isCore || (decorator.import !== null && decorator.import.from === '@angular/core');
```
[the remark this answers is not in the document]
devversion: Yeah, I think this is a general issue also in ngtsc. We should definitely consider this there.

packages/compiler-cli/src/transformers/downlevel_decorators_transform.ts
devversion: haha! good one 😛 I would have done that if the file wasn't ~2.4mb and could be displayed in Github. I figured adding a comment on why Github vs. Sourcegraph isn't necessary.

packages/compiler-cli/src/transformers/downlevel_decorators_transform.ts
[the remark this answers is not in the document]
clydin: Understood.  There's probably a decent about of additional refactoring that would be good long-term especially with newer TS capabilities (most of this was written against TS 2.x).
But I noticed a could small changes that could remove casting (which can be very error prone) so figured I'd point them out.  Feel free to ignore them for now if preferred.

packages/compiler-cli/src/transformers/downlevel_decorators_transform.ts
devversion: We're detecting it already, but just not having it typed here. I'll make it optional.

packages/compiler-cli/src/transformers/downlevel_decorators_transform.ts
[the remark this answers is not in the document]
devversion: Yeah, that makes sense. Definitely appreciated. Thanks! I figured I'd just make it clear that initial goal was to not clean this up, but I agree that there are a lot of possibilities 😄

packages/compiler-cli/src/transformers/downlevel_decorators_transform.ts
```diff
@@ -0,0 +1,660 @@
...
+  // we abort gracefully as most likely TS changed the internal structure of the emit resolver.
+  if (originalReferenceResolution === undefined) {
+    return false;
+  }
+  emitResolver.isReferencedAliasDeclaration = function(node, checkChildren) {
+    if (isAliasImportDeclaration(node) && referencedAliases.has(node)) {
+      return true;
+    }
+    return originalReferenceResolution.call(emitResolver, node, checkChildren);
+  };
```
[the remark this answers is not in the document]
devversion: I like it. 👍  thanks!

packages/compiler-cli/src/transformers/downlevel_decorators_transform.ts
```diff
@@ -0,0 +1,660 @@
...
+ * https://sourcegraph.com/github.com/microsoft/TypeScript@3eaa7c65f6f076a08a5f7f1946fd0df7c7430259/-/blob/src/compiler/checker.ts#L31219-31257
+ *
+ * @returns A boolean indicating whether the alias reference solution could be patched properly.
+ */
+function patchAliasReferenceResolution(
+    context: TransformationContextWithResolver, referencedAliases: Set<ts.Declaration>): boolean {
+  // If the `getEmitResolver` method is not available, TS most likely changed the
+  // internal structure of the transformation context. We will abort gracefully.
+  if (context.getEmitResolver === undefined) {
+    return false;
```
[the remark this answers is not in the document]
devversion: FWIW: We are throwing below. This has been moved into the method now for clarity though

packages/compiler-cli/src/transformers/downlevel_decorators_transform.ts
```diff
@@ -0,0 +1,660 @@
...
+}
+
+/**
+ * Gets whether a given node corresponds to an import alias declaration. Alias
+ * declarations can be import specifiers, namespace imports or import clauses
+ * as these do not declare an actual symbol but just point to a target declaration.
+ */
+function isAliasImportDeclaration(node: ts.Node): node is ts.ImportSpecifier|ts.NamespaceImport|
+    ts.ImportClause {
+  return ts.isImportSpecifier(node) || ts.isNamespaceImport(node) || ts.isImportClause(node);
```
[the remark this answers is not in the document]
devversion: Yeah, that is correct. Let's keep them in the list of referenced alias declarations for safety and completeness though. Shouldn't hurt.

packages/compiler-cli/src/transformers/downlevel_decorators_transform.ts
devversion: The problem I'd see here though is that it could happen that we clear the list before TypeScript even ran its import eliding. We could solve that though by just re-declaring a new set. I assume that's what you meant?

devversion: I think we should keep it as is because otherwise we'll have N sets whereas `N` is the amount of source files. In the worst case, a separate emit resolver will have a list of alias import declarations that it doesn't even check reference for (unless separate emit resolvers; but same AST nodes with parse tree; that would be seem contradictory though). What do you think?

packages/compiler-cli/src/transformers/downlevel_decorators_transform.ts
devversion: @clydin Given that the reflection host should be fine with any arbitrary standard TS program, this should be good. We'll have a better factory entry-point for the CLI in `compiler-cli/tooling` where it only takes a TypeScript program and nothing more. Does that sound reasonable to you?

The plan would be then that this transform is used in the CLI, so that we have a canonical place for decorator/metadata downleveling. I'm not sure though how exactly it will work in the CLI. Could the CLI start depending on the compiler-cli exposed transformer. What happens if one uses an older compiler-cli version, but the latest CLI?

packages/compiler-cli/src/transformers/program.ts
```diff
@@ -591,18 +519,26 @@ class AngularCompilerProgram implements Program {
       flatModuleMetadataTransforms.push(transformer);
     }
 
-    if (stripDecorators) {
```
clydin, line 594: Is Ivy not stripping decorators anymore with this change?

devversion: This flag was only `true` for `program.emitRender3`. Though that is no longer the real entry-point for Ivy as it uses its own program. I removed it while being at it.

packages/compiler-cli/src/transformers/downlevel_decorators_transform.ts
clydin: Both the program instance and the type checker instance would be available when creating the transformer.
The FW and CLI are version locked by the major version.  So in this case CLI v10 can safely use anything that will be in FW v10.  However, if something was only available in FW v10.1 then the CLI could use it but would need a fallback if FW v10.0 was used.

devversion: Got it. Sounds like we would be in good time to use this in the CLI. Thanks for the details!

clydin (approved):

packages/compiler-cli/src/transformers/patch_alias_reference_resolution.ts
```diff
@@ -0,0 +1,119 @@
...
+ * Describes a TypeScript transformation context with the internal emit
+ * resolver exposed. There are requests upstream in TypeScript to expose
+ * that as public API: https://github.com/microsoft/TypeScript/issues/17516..
+ */
+type TransformationContextWithResolver = ts.TransformationContext&{
+  getEmitResolver: () => EmitResolver;
+};
+
+/** Describes a subset of the TypeScript internal emit resolver. */
+type EmitResolver = {isReferencedAliasDeclaration ? (node: ts.Node, checkChildren?: boolean) :
```
clydin, line 21: nit: maybe an interface for this to improve the formatting.

alxhub: I agree with this, for both types.

packages/compiler-cli/src/transformers/program.ts
```diff
@@ -7,26 +7,28 @@
...
+import {AotCompiler, AotCompilerOptions, core, createAotCompiler, FormattedMessageChain, GeneratedFile, getParseErrors, isFormattedError, isSyntaxError, MessageBundle, NgAnalyzedFileWithInjectables, NgAnalyzedModules, ParseSourceSpan, PartialModule, Serializer, Xliff, Xliff2, Xmb} from '@angular/compiler';
 import * as fs from 'fs';
 import * as path from 'path';
 import * as ts from 'typescript';
 
-import {translateDiagnostics, TypeCheckHost} from '../diagnostics/translate_diagnostics';
-import {createBundleIndexHost, MetadataCollector, ModuleMetadata} from '../metadata';
+import {translateDiagnostics} from '../diagnostics/translate_diagnostics';
+import {createBundleIndexHost, MetadataCollector} from '../metadata';
+import {isAngularCorePackage} from '../ngtsc/core/src/compiler';
```
alxhub, line 17: ngtsc tries to ensure that all imports go through the barrel `index.ts` files - so this should be:
```suggestion
import {isAngularCorePackage} from '../ngtsc/core';
```
You might need to add a re-export to make that work.

alxhub (approved):

alxhub:
[Presubmit](http://test/OCL:314993213:BASE:314993260:1591391321435:7367c7d1)

IgorMinar (approved):
LGTM now. Thank you! Excellent work @devversion! ❤️

IgorMinar:
[presubmit](http://test/OCL:314731511:BASE:315028346:1591404948218:a3455f1c)

IgorMinar (approved):
LGTM! Thank you!

packages/bazel/test/ng_package/example_package.golden
```diff
@@ -611,7 +611,7 @@ let MyService = /** @class */ (() => {
...
...ata:application/json;base64,eyJ2ZXJzaW9uIjozLCJmaWxlIjoicHVibGljLWFwaS5qcyIsInNvdXJjZVJvb3QiOiIiLCJzb3VyY2VzIjpbIi4uLy4uLy4uLy4uLy4uLy4uLy4uLy4uLy4uL3BhY2thZ2VzL2JhemVsL3Rlc3QvbmdfcGFja2FnZS9leGFtcGxlL2ltcG9ydHMvcHVibGljLWFwaS50cyJdLCJuYW1lcyI6W10sIm1hcHBpbmdzIjoiQUFBQTs7Ozs7O0dBTUc7QUFFSCxPQUFPLEVBQUMsVUFBVSxFQUFDLE1BQU0sZUFBZSxDQUFDO0FBQ3pDLE9BQU8sRUFBQyxlQUFlLEVBQUMsTUFBTSxVQUFVLENBQUM7OztBQUV6QztJQUFBLE1BQ2EsU0FBUztRQUNwQixZQUFtQixhQUE4QjtZQUE5QixrQkFBYSxHQUFiLGFBQWEsQ0FBaUI7UUFBRyxDQUFDOzs7O2dCQUZ0RCxVQUFVLFNBQUMsRUFBQyxVQUFVLEVBQUUsTUFBTSxFQUFDOzs7Z0JBRnhCLGVBQWU7O29CQVR2QjtLQWNDO1NBRlksU0FBUyIsInNvdXJjZXNDb250ZW50IjpbIi8qKlxuICogQGxpY2Vuc2VcbiAqIENvcHlyaWdodCBHb29nbGUgTExDIEFsbCBSaWdodHMgUmVzZXJ2ZWQuXG4gKlxuICogVXNlIG9mIHRoaXMgc291cmNlIGNvZGUgaXMgZ292ZXJuZWQgYnkgYW4gTUlULXN0eWxlIGxpY2Vuc2UgdGhhdCBjYW4gYmVcbiAqIGZvdW5kIGluIHRoZSBMSUNFTlNFIGZpbGUgYXQgaHR0cHM6Ly9hbmd1bGFyLmlvL2xpY2Vuc2VcbiAqL1xuXG5pbXBvcnQge0luamVjdGFibGV9IGZyb20gJ0Bhbmd1bGFyL2NvcmUnO1xuaW1wb3J0IHtNeVNlY29uZFNlcnZpY2V9IGZyb20gJy4vc2Vjb25kJztcblxuQEluamVjdGFibGUoe3Byb3ZpZGVkSW46ICdyb290J30pXG5leHBvcnQgY2xhc3MgTXlTZXJ2aWNlIHtcbiAgY29uc3RydWN0b3IocHVibGljIHNlY29uZFNlcnZpY2U6IE15U2Vjb25kU2VydmljZSkge31cbn1cbiJdfQ==
```
devversion, line 614: Note for reviewer: You might wonder why this changed the source map. This is because before this fix, the actual parameter type identifier was used for `ctorParameters`. Now since we use the import alias though for `ctorParameters`, the source map will change back to what it was with tsickle too.

alxhub (approved):
LGTM with the fix for aliased constructor parameter names.

alxhub:
[Presubmit](http://test/OCL:315509583:BASE:315509629:1591723487449:ab74d3d)

alxhub:
[And again](http://test/OCL:315535390:BASE:315535448:1591730591781:f0118361)

alxhub:
[Here we go again](http://test/OCL:315568666:BASE:315568709:1591740426265:5efd5946)

PR #37383, opened 2020-06-01
docs(core): remove v10 mention from @Injectable warning

kara:
In v9, we started showing a console warning when
instantiating a token that inherited its @Injectable
decorator rather than providing its own. This warning
said that the pattern would become an error in v10.

However, we have decided to wait until at least v11
to throw in this case, so this commit updates the
warning to be less prescriptive about the exact
version when the pattern will no longer be supported.

### DIFF
--- a/packages/core/src/di/jit/injectable.ts
+++ b/packages/core/src/di/jit/injectable.ts
@@ -46,7 +46,7 @@ export function compileInjectable(type: Type<any>, srcMeta?: Injectable): void {
     // undecorated parent class. We work around it by defining a method which should be used
     // as a fallback. This should only be a problem in JIT mode, because in AOT TypeScript
     // seems to have a workaround for static properties. When inheriting from an undecorated
-    // parent is no longer supported in v10, this can safely be removed.
+    // parent is no longer supported (v11 or later), this can safely be removed.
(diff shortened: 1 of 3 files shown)


packages/core/src/di/jit/injectable.ts
```diff
@@ -46,7 +46,7 @@ export function compileInjectable(type: Type<any>, srcMeta?: Injectable): void {
     // undecorated parent class. We work around it by defining a method which should be used
     // as a fallback. This should only be a problem in JIT mode, because in AOT TypeScript
     // seems to have a workaround for static properties. When inheriting from an undecorated
-    // parent is no longer supported in v10, this can safely be removed.
+    // parent is no longer supported in v11 or so, this can safely be removed.
```
atscott, line 49:
```suggestion
    // parent is no longer supported (v11 or later), this can safely be removed.
```
🤷

atscott (approved):

packages/core/src/di/jit/injectable.ts
kara: Good call, fixing....

kara:
merge-assistance: I'm not able to run presubmits today due to some technical issues with my security key. Caretaker, can you help?

Issue #11058, opened 2016-08-25
Use of '+' character in URLSearchParams()


gnujeremie:
**I'm submitting a ...**

```
[x ] bug report => search github for a similar issue or PR before submitting
[ ] feature request
[ ] support request => Please do not submit support request here, instead see https://github.com/angular/angular/blob/master/CONTRIBUTING.md#question
```

**Current behavior**
Code exemple

```
let body = new URLSearchParams();

for (let i in values) {
    body.set(i, values[i]);
}

let options = RequestOptions();
options.headers.append('Content-Type', 'application/x-www-form-urlencoded')
return this.http.post(uri, body, options).then();
[...]
```

If `values[i]` has a '+' character in it, then it is replaced by spaces. 

**Expected/desired behavior**
'+' should be sent as any other character.

**Please tell us about your environment:**
- **Angular version:** 2.0.0-rc.5
- **Browser:** [all | Chrome XX | Firefox XX | IE XX ] 
- **Language:** [Typescript]

vicb:
Have you checked for pending issues / PRs ?

gnujeremie:
I checked for the pending issues, found nothing, but now looking for PRs, it may be related to this one https://github.com/angular/angular/pull/9651 . But it is merged, so I should not have the problem right ? Or do I do something wrong ? >.<

gnujeremie:
After digging deeper, it seems like the value posted is ok, (and the '+' appears in the string), but it's on the server side, it is replaced by an empty space, even when I simply get it from $_POST in PHP.

mhevery:
Your code example does not actually show the issue. Could you create a better example for reproducing the problem?

gnujeremie:
Sure : 
```javascript
let body = new URLSearchParams();

body.set('label', 'here is a label with + in it');

let options = RequestOptions();
options.headers.append('Content-Type', 'application/x-www-form-urlencoded')
return this.http.post('http://localhost/admin/save/', body, options).then();
[...]
```

Then, in ```admin/save``` :
```php
$label = $_POST['label'];
echo $label;
```
It displays : 
```
here is a label with    in it
```

When I check the browser dev tools, I can see that the value sent is ```here is a label with + in it```, so I am not sure if this is an angular bug or if it is the way PHP handles it. For now, I fixed it by using a custom QueryEncoder with ```URLSearchParams('', new CustomQueryEncoder())```, with encodeKey() and encodeValue() just doing this : 
```javascript
[...]

PR #32598, opened 2019-09-11
feat(common): make `HttpParameterCodec` injectable

johnny-mh:
You can inject custom `HttpParameterCodec` to not use default parameter encoder.

## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

```
[ ] Bugfix
[x] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Documentation content changes
[ ] angular.io application / infrastructure changes
[ ] Other... Please describe:
```

## What is the current behavior?

Specific API server can't accept character. '+' so we need to encode that.

https://github.com/angular/angular/blob/5298b2bda34a8766b28c8425e447f94598b23901/packages/common/http/src/params.ts#L64

But angular's standard encoder decode that.

So developer needs to add custom encoder every http method invokes. and it's very annoying.

Related Issue Numbers: #19710, #23157 #18261, #18719, #11058

## What is the new behavior?

User can provide default custom http parameter codec.

## Does this PR introduce a breaking change?
```
[ ] Yes
[x] No
```

PR #37385, opened 2020-06-01
fix(http): provide alternative to broken HttpUrlEncodingCodec

alxhub:
Currently in Angular, by default `HttpParam` instances are encoded with the
`HttpUrlEncodingCodec`. This is an implementation of `HttpParameterCodec`
which uses `encodeURIComponent` plus a number of manual post-processing
substitutions to encode keys and values for x-www-form-urlencoded contexts.

Unfortunately, the logic of `HttpUrlEncodingCodec` is fatally flawed. Among
other problems, the character '+' is left unencoded. In such contexts, '+'
is interpeted as a space by any correct decoder, meaning that this
implementation actually changes values. This logic is indefensible, and
appears to be the result of a copy/paste bug from angular.js encoding meant
for a different context.

Fixing this bug directly is challenging since existing backends may be
relying on Angular's broken behavior, so changing the default is a breaking
change and will need to be done in an Angular major version. This commit
begins that process by introducing a new codec, `HttpUrlComponentCodec`,
which implements a corrected encoding algorithm. In a future commit for v11,
this corrected codec can be made the default.

In the meantime, this commit introduces a static method which can change the
built-in default codec, for convenience.

```typescript
HttpParams.setDefaultParameterCodec(HttpUrlComponentCodec);
```

See #11058 for more details.

### DIFF
--- a/packages/common/http/src/module.ts
+++ b/packages/common/http/src/module.ts
@@ -6,17 +6,21 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {Injectable, Injector, ModuleWithProviders, NgModule} from '@angular/core';
+import {Inject, Injectable, InjectionToken, Injector, ModuleWithProviders, NgModule, Optional} from '@angular/core';
 import {Observable} from 'rxjs';
 
 import {HttpBackend, HttpHandler} from './backend';
 import {HttpClient} from './client';
 import {HTTP_INTERCEPTORS, HttpInterceptor, HttpInterceptorHandler, NoopInterceptor} from './interceptor';
 import {JsonpCallbackContext, JsonpClientBackend, JsonpInterceptor} from './jsonp';
+import {HttpParams, HttpUrlComponentCodec as HttpCanonicalParamCodec} from './params';
 import {HttpRequest} from './request';
 import {HttpEvent} from './response';
 import {BrowserXhr, HttpXhrBackend, XhrFactory} from './xhr';
 import {HttpXsrfCookieExtractor, HttpXsrfInterceptor, HttpXsrfTokenExtractor, XSRF_COOKIE_NAME, XSRF_HEADER_NAME} from './xsrf';
+
+export const USE_CANONICAL_PARAM_ENCODING =
+    new InjectionToken<boolean>('USE_CANONICAL_PARAM_ENCODING');
 
 /**
  * An injectable `HttpHandler` that applies multiple interceptors
@@ -164,6 +168,25 @@ export class HttpClientXsrfModule {
   ],
 })
 export class HttpClientModule {
+  constructor(@Inject(USE_CANONICAL_PARAM_ENCODING) @Optional() useCanonicalParamEncoding: boolean|
+              null) {
+    if (useCanonicalParamEncoding === true) {
+      HttpParams.setDefaultParameterCodec(HttpCanonicalParamCodec);
+    }
+  }
+
+  static withOptions(options: {useCanonicalParamEncoding?: boolean}):
+      ModuleWithProviders<HttpClientModule> {
+    return {
+      ngModule: HttpClientModule,
+      providers: [
+        {
+          provide: USE_CANONICAL_PARAM_ENCODING,
+          useValue: options.useCanonicalParamEncoding || false,
+        },
+      ],
+    };
+  }
 }
 
 /**
--- a/packages/common/http/src/params.ts
+++ b/packages/common/http/src/params.ts
@@ -14,10 +14,32 @@
  * @publicApi
  **/
 export interface HttpParameterCodec {
+  /**
+   * Encodes a key name for a URL parameter or query-string.
+   * @param key The key name.
+   * @returns The encoded key name.
+   */
   encodeKey(key: string): string;
+
+  /**
+   * Encodes the value of a URL parameter or query-string.
+   * @param value The value.
+   * @returns The encoded value.
+   */
   encodeValue(value: string): string;
 
+  /**
+   * Decodes an encoded URL parameter or query-string key.
+   * @param key The encoded key name.
+   * @returns The decoded key name.
+   */
   decodeKey(key: string): string;
+
+  /**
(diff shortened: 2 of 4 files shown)


packages/common/http/src/params.ts
```diff
@@ -251,6 +292,10 @@ export class HttpParams {
         .join('&');
   }
 
+  static setDefaultParameterCodec(codec: {new(): HttpParameterCodec}): void {
```
JoostK, line 295: IMHO, we should not expose the static default as public API. Together with `useCanonicalParamEncoding` it opens up a couple scenarios where setting the option overrides an encoder that was previously set using this setter.

This being a static field in the first place is unlike other Angular designs, where behavior is typically customize-able using DI tokens. There's actually existing PR #32598 that implements a DI approach.

packages/common/http/src/module.ts
```diff
@@ -164,6 +168,25 @@ export class HttpClientXsrfModule {
...
 })
 export class HttpClientModule {
+  constructor(@Inject(USE_CANONICAL_PARAM_ENCODING) @Optional() useCanonicalParamEncoding: boolean|
+              null) {
+    if (useCanonicalParamEncoding === true) {
+      HttpParams.setDefaultParameterCodec(HttpCanonicalParamCodec);
+    }
+  }
+
+  static withOptions(options: {useCanonicalParamEncoding?: boolean}):
```
JoostK, line 178: I'm contemplating whether we should go with this option, or just expose a DI token that people can provide. The option is nice from a migration-path perspective, but it does pull in both encoder implementations as they are always both referenced. This may be acceptable during the migration period, as the overhead can be removed once the deprecated behavior is deleted. WDYT?

If we stick with the options, the option and `withOptions` in total should be deprecated from the start IMO. Also, if the option exists, perhaps we shouldn't expose the `HttpUrlComponentCodec` implementation as public API (although it probably doesn't hurt either way)

packages/common/http/src/params.ts
```diff
@@ -30,45 +52,59 @@ export interface HttpParameterCodec {
...
   decodeValue(value: string) {
     return decodeURIComponent(value);
   }
 }
 
+/**
+ * Provides a canonical encoding and decoding of URL parameter keys and values, using the native
+ * `encodeURIComponent` and `decodeURIComponent` functions within the browser.
+ *
+ * Replaces `HttpUrlEncodingCodec` which
```
JoostK, line 79: This is unfinished.

packages/common/http/src/params.ts
```diff
@@ -30,45 +52,59 @@ export interface HttpParameterCodec {
...
+  }
+
+  decodeValue(value: string): string {
+    return decodeURIComponent(value);
+  }
+}
+
+/**
+ * Encodes using `encodeURIComponent`, but replaces '%20' which encodes a space character with
+ * '+'.
```
JoostK, line 103: Please include that the URL encoded space is translated into `+` because the key/values appear in query parameters, which are `application/x-www-form-urlencoded` encoded.

packages/common/http/test/params_spec.ts
```diff
@@ -91,5 +91,30 @@ import {HttpParams} from '@angular/common/http/src/params';
...
+    describe('HttpUrlEncodingCodec', () => {
+      beforeEach(() => {
+        HttpParams.setDefaultParameterCodec(HttpUrlComponentCodec);
+      });
+
+      afterEach(() => {
+        HttpParams.setDefaultParameterCodec(HttpUrlEncodingCodec);
+      });
+
+      it('should encode + characters correctly', () => {
```
JoostK, line 104: I am missing tests with spaces, especially multiple spaces to verify that the replacement happens for all occurrences.

packages/common/http/src/params.ts
alxhub: I agree that the static method should probably be `@internal`.

I considered the design in #32598 before implementing this. It's definitely more elegant, but I find it weird/confusing that `HttpParams.toString()` will return different serializations depending on whether it's gone through the `HttpClient` or not. That has the potential to be especially problematic in tests, when you might want to test that the request you're generating is correct without actually dispatching it.

In this PR's model, every `HttpParams` instance gets its codec set on construction, and it doesn't change throughout the lifetime of the object. Thus `new HttpParams()` will automatically serialize correctly.

It's potentially problematic for things like multiple apps on the page that each want their own defaults, but as we've talked about before that's kind of an edge case on an edge case.

packages/common/http/src/module.ts
alxhub: A private DI token would work. I don't think it should be a public token, as I don't want this hacky way of setting the default codec to stick around.

`HttpUrlComponentCodec` is public by design, as it's expected that users might subclass it to implement their own codecs.

packages/common/http/src/params.ts
JoostK: Ah, I had not considered the scenario where `HttpParams` is constructed manually, that would indeed be cumbersome if it depends only on DI.

If this is `@internal` then the static field can be removed when the migration flag is removed.

alxhub:
(Closing until v12)

petebacondarwin:
Re-opened this, since closing it appears to have allowed it to be forgotten.

I am also uncomfortable with using a combination of static on `HttpParams` and DI in `HttpClient`.

The primary argument for the static is that it makes it easier to create `HttpParams` objects without having to know what the default encoder is.

I would like to propose a different API approach, which involves adding helpers to `HttpClient`:

```
class HttpClient {
  constructor(..., private defaultParamEncoder: HttpParameterCodec) { ... }

  ...

  createHttpParams(options: HttpParamsOptions = {}): HttpParams {
    return new HttpParams({encode: this.defaultParamEncoder, ...options});
  }
}
```

Then in tests one just needs to have access to the `HttpClient` and replace all `new HttpParams(...)` with `http.createHttpParams(...)` calls.

This avoids adding static state to the `HttpParams` class and is more in keeping with how Angular packages generally rely upon DI.

alxhub:
Closing in favor of #45111

PR #32353, opened 2019-08-27
Add trusted types in client side Angular 

Siegrift:
## Trusted Types
[Trusted Types](https://github.com/WICG/trusted-types) ([spec](https://wicg.github.io/trusted-types/dist/spec/), [introductory article](https://developers.google.com/web/updates/2019/02/trusted-types)) is a new experimental DOM API implemented within the WICG , with a working [Chrome implementation](https://www.chromestatus.com/feature/5650088592408576).

The API creates a few new objects available on the global object in the browser, like most other web APIs ([impl in TS](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/trusted-types/index.d.ts) and in [Closure compiler](https://github.com/WICG/trusted-types/blob/master/externs/externs.js)).

Under certain conditions, controlled by a HTTP header (analogous to Content-Security-Policy behavior), the API can enable the enforcement - then it changes the signature of several DOM API functions and property setters, such that they accept specific object types, and reject strings. Colloquially, DOM API becomes strongly typed.

For example, with Trusted Types Element.innerHTML property setter accepts a TrustedHTML object.

Trusted Type objects stringify to their inner value. This API shape is a deliberate choice that enables existing web applications and libraries to gradually migrate from strings to Trusted Types without breaking functionality. In our example, it makes it possible to write the following:

```javascript
const policy = TrustedTypes.createPolicy('foo', { 
  createHTML: (s) => { /* some validation*/; return s} 
});

const trustedHTML = policy.createHTML('bar');
anElement.innerHTML = trustedHTML

anElement.innerHTML === 'bar'
```

The above code works regardless if the Trusted Types enforcement is enabled or not.

Reading from the DOM is unaffected, so Element.innerHTML getter returns a string. That's for practical reasons -- web applications read from DOM more often than they write to it, and only writing exposes the application to DOM XSS risks. Typing only the setters allows us to secure web applications with minimal code changes.

## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit _(Although, it would be nice to squash them)_
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**This is only a draft PR for now, there are basically zero tests written and the PR is only tested locally on my sample application. In next few days I will add more tests, and I can squash the commits if you want want.**

## PR Type
What kind of change does this PR introduce?

[...]

PR #37397, opened 2020-06-02
refactor(animations): do not assign to innerHTML

taymonbeal:
This should not change behavior, but it prevents false-positive warnings in various static analysis
tools, including tools used internally at Google.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [X] No

### DIFF
--- a/packages/animations/browser/src/render/css_keyframes/css_keyframes_driver.ts
+++ b/packages/animations/browser/src/render/css_keyframes/css_keyframes_driver.ts
@@ -72,7 +72,7 @@ export class CssKeyframesDriver implements AnimationDriver {
     keyframeStr += `}\n`;
 
     const kfElm = document.createElement('style');
-    kfElm.innerHTML = keyframeStr;
+    kfElm.textContent = keyframeStr;
     return kfElm;
   }
 


jelbourn (approved):
LGTM

Reviewed-for: fw-animations

jelbourn (approved):
LGTM, using global approvers since Igor and Pete are on vacation

packages/animations/browser/src/render/css_keyframes/css_keyframes_driver.ts
```diff
@@ -72,7 +72,7 @@ export class CssKeyframesDriver implements AnimationDriver {
     keyframeStr += `}\n`;
 
     const kfElm = document.createElement('style');
-    kfElm.innerHTML = keyframeStr;
+    kfElm.textContent = keyframeStr;
```
AndrewKushnir, line 75: Quick comment: since it's tricky to have a test for this change, could you please add a comment explaining that the `textContent` is used here intentionally (and why) and mention that it should not be changed to `innerHTML`. That should help make sure this part of the code is only changed when taking into account additional context (provided in a comment). Thank you.

taymonbeal: I actually have discovered a behavior difference of sorts: `innerHTML` doesn't work if Trusted Types are enabled. Is there any way we could test based on that?

(Also, I don't think there was much of a rationale for using `innerHTML` in the first place; conceptually, the thing you want here is CSS text, not HTML.)

AndrewKushnir: There is an open PR #32353 to add Trusted Types support, so my guess is that the mentioned PR should land first to simplify the testing. As an option, you can look at the tests in that PR to see if you can use some of the code to repro the issue. If adding a test without the changes from PR #32353 would be difficult, we can just add a comment to the code to indicate that `textContent` is used intentionally (instead of `innerHTML`). This should help prevent changing that code in the future without taking into account the described use-case/scenario.

atscott:
Hi @taymonbeal, can you take a look at the failures in the [presubmit](https://test.corp.google.com/ui#id=OCL:325905615:BASE:325905670:1597099984982:aa23ec2f)? This change doesn't build and produces this error:  "Do not assign non-constant values to HTMLStyleElement#textContent or HTMLStyleElement#innerText, as this can open side channels for sensitive information to leak"

taymonbeal:
This is going to involve some configuration changes in google3. Working on it.

bjarkler:
This has now been unblocked on the google3 side: [new presubmit](http://tap/336290993).

This should be a completely safe change, and will be needed to unblock Trusted Types support in applications that use the animations package as well as Angular's tests. @atscott, can you take another look?

packages/animations/browser/src/render/css_keyframes/css_keyframes_driver.ts
bjarkler: Testing this change specifically for Trusted Types compliance is not necessary, since regular unit tests for this feature will fail in a testing environment that enforces Trusted Types. I've verified that locally and changing this to `textContent` indeed fixes that. (Hence fixing this is actually needed before we can enable Trusted Types enforcement in unit tests.)

AndrewKushnir: @bjarkler thanks for testing this change. Since the changes to support Trusted Types are in progress, we can probably keep things as is and there is no need for additional comment in the code if there are unit tests that verify this behavior and prevent regressions.

AndrewKushnir:
@bjarkler thanks for running a presubmit 👍 I think it'd be great to run a global presubmit as well, to make sure this change is safe for all targets in google3. Please let us know if you need help with that.

packages/animations/browser/src/render/css_keyframes/css_keyframes_driver.ts
bjarkler: Yes, definitely. We'll have precautions in place (in the form of unit and integration tests that enforce Trusted Types) to prevent regressions, so there's no need to worry about this at individual locations in the codebase.

bjarkler:
Sounds good @AndrewKushnir, I'll start the global presubmit.

IgorMinar (approved):
This change looks good to me.

I think it was an oversight to use `innerHTML` here in the first place, but since `innerText` and `innerHTML` have the same behavior for style elements this was never a security issue. It's good to make the code more explicit though.

bjarkler:
[global presubmit](http://tap/336290993)

PR #37404, opened 2020-06-02
fix(common): prevent duplicate URL change notifications

LayZeeDK:
Prevent duplicate notifications from being emitted when multiple URL change listeners are registered using `Location#onUrlChange`.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

When two URL change listeners are registered, the first one will receive duplicate URL change notifications.

Issue Number: N/A

## What is the new behavior?
When two URL change listeners are registered, both of them will only receive a single notification when the URL changes.

- [ ] Yes
- [x] No

## Other information
Cc @aaronfrost

`Location#onUrlChange` is only used by NgUpgrade's `$location` shim for AngularJS.

### DIFF
--- a/packages/common/src/location/location.ts
+++ b/packages/common/src/location/location.ts
@@ -64,6 +64,7 @@ export class Location {
   _platformLocation: PlatformLocation;
   /** @internal */
   _urlChangeListeners: ((url: string, state: unknown) => void)[] = [];
+  private _urlChangeSubscription?: SubscriptionLike;
 
   constructor(platformStrategy: LocationStrategy, platformLocation: PlatformLocation) {
     this._platformStrategy = platformStrategy;
@@ -194,9 +195,12 @@ export class Location {
    */
   onUrlChange(fn: (url: string, state: unknown) => void) {
     this._urlChangeListeners.push(fn);
-    this.subscribe(v => {
-      this._notifyUrlChangeListeners(v.url, v.state);
-    });
+
+    if (!this._urlChangeSubscription) {
+      this._urlChangeSubscription = this.subscribe(v => {
+        this._notifyUrlChangeListeners(v.url, v.state);
+      });
+    }
   }
 
   /** @internal */
--- a/packages/common/test/location/location_spec.ts
+++ b/packages/common/test/location/location_spec.ts
@@ -7,7 +7,7 @@
  */
 
 import {CommonModule, Location, LocationStrategy, PathLocationStrategy, PlatformLocation} from '@angular/common';
-import {MockPlatformLocation} from '@angular/common/testing';
+import {MockLocationStrategy, MockPlatformLocation} from '@angular/common/testing';
 import {inject, TestBed} from '@angular/core/testing';
 
 const baseUrl = '/base';
@@ -84,7 +84,7 @@ describe('Location Class', () => {
       TestBed.configureTestingModule({
         imports: [CommonModule],
         providers: [
-          {provide: LocationStrategy, useClass: PathLocationStrategy},
+          {provide: LocationStrategy, useClass: MockLocationStrategy},
           {
             provide: PlatformLocation,
             useFactory: () => {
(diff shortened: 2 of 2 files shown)


LayZeeDK:
Does it make sense to add an `ngOnDestroy` lifecycle hook to unsubscribe from the `_onUrlChangeSubscription` in the `Location` service? Would this make a difference to multiple applications on the same page? Is there a use case for applications being shut down after bootstrapping them? Hot Module Replacement maybe?

JoostK:
> Does it make sense to add an `ngOnDestroy` lifecycle hook to unsubscribe from the `_onUrlChangeSubscription` in the `Location` service? Would this make a difference to multiple applications on the same page? Is there a use case for applications being shut down after bootstrapping them? Hot Module Replacement maybe?

It looks like it won't be necessary to teardown the subscription manually, as the source is an `EventEmitter` that is local to the instance, and all referenced memory is also local to the instance. As soon as the instance is no longer referenced from other places, it will become eligible for the GC to collect.

packages/common/test/location/location_spec.ts
```diff
@@ -113,5 +113,28 @@ describe('Location Class', () => {
...
+               notificationCount += 1;
+
+               return undefined;
+             }
+
+             function noopChangeListener(url: string, state: unknown) {
+               return undefined;
+             }
+
+             let notificationCount = 0;
```
JoostK, line 131: I would move this to before `incrementChangeListener` as it's referenced in that callback function.

packages/common/test/location/location_spec.ts
```diff
@@ -113,5 +113,28 @@ describe('Location Class', () => {
...
+             location.onUrlChange(incrementChangeListener);
+             location.onUrlChange(noopChangeListener);
+
+             locationStrategy.simulatePopState('/test');
+
+             expect(notificationCount).toBe(1);
```
JoostK, line 137: Consider adding this assertion to above `simulatePopState` with an expected count of 0, making it more explicit when the counter is expected to be updated.

packages/common/test/location/location_spec.ts
LayZeeDK: I like to make my declarations follow hoisting, but I can change the functions to lambda expressions to make us both happy 🙂

packages/common/test/location/location_spec.ts
LayZeeDK: In the same test case, right?

packages/common/test/location/location_spec.ts
JoostK: I have the opposite preference, apparently :-) I had looked into using an arrow function as well, but noticed other tests are also written with a named function so the current state looks good to me.

packages/common/src/location/location.ts
```diff
@@ -64,6 +64,8 @@ export class Location {
   _platformLocation: PlatformLocation;
   /** @internal */
   _urlChangeListeners: ((url: string, state: unknown) => void)[] = [];
+  /** @internal */
+  _urlChangeSubscription?: SubscriptionLike;
```
atscott, line 68: Is there any reason this can't be `private` instead of `internal`?

packages/common/test/location/location_spec.ts
```diff
@@ -113,5 +113,31 @@ describe('Location Class', () => {
          expect((location as any)._urlChangeListeners.length).toBe(1);
          expect((location as any)._urlChangeListeners[0]).toEqual(changeListener);
        }));
+
+    it('should only notify listeners once when multiple listeners are registered',
+       inject(
+           [Location, LocationStrategy],
+           (location: Location, locationStrategy: MockLocationStrategy) => {
+             let notificationCount = 0;
```
atscott, line 121:
```suggestion
       () => {
       const location = TestBed.get(Location);
       const locationStrategy = TestBed.get(MockLocationStrategy);
             let notificationCount = 0;
```
nit: can we update this to use the newer `TestBed` APIs?

atscott (changes requested):
👍 Thanks for this fix! Only some small nits in the comments

atscott:
[presubmit](https://test.corp.google.com/ui#id=OCL:314602448:BASE:314602915:1591219093378:851c605b)

packages/common/src/location/location.ts
LayZeeDK: I'm trying to follow the code style of this class. If we want to change the access modifiers to `private`, shouldn't we do them all?

packages/common/test/location/location_spec.ts
LayZeeDK: I'm trying to follow the code style of this test suite. If we want to use `TestBed`, shouldn't we change all the test cases? Also, we should use `TestBed.inject`.

packages/common/src/location/location.ts
atscott: Would you mind still making this change for the new item and contributing a follow-up PR to update the others? I don't think that we should continue bad/old conventions just for the sake of consistency.

packages/common/test/location/location_spec.ts
atscott: Right, use `TestBed.inject` instead. Also, same comment as above - please still update this one to the newer style and then follow up with a PR to update refactor the others. Thanks!

packages/common/src/location/location.ts
LayZeeDK: Got it. We still want the underscore prefix though, right?

atscott: Yep, that convention hasn't changed AFAIK

LayZeeDK:
Good to go.

I'll open another PR afterwards to:
- Prefer `private` over `@internal`
- Prefer `TestBed.inject` over `inject`

in these files.

atscott (approved):

LayZeeDK:
I think you have to remove the `PR action: review` label, @atscott 🙂

PR #37408, opened 2020-06-02
fix(router): fix navigation ignoring logic to compare to the browser url

atscott:
This PR changes the logic for determining when to skip route processing from
using the URL of the last attempted navigation to the actual resulting URL after
that transition.

Because guards may prevent navigation and reset the browser URL, the raw
URL of the previous transition may not match the actual URL of the
browser at the end of the navigation process. For that reason, we need to use
`urlAfterRedirects` instead.

Other notes:
* These checks in `scheduleNavigation` were added in https://github.com/angular/angular/commit/eb2ceff4ba19a09a5f400f81e371ec176878cd37
The test still passes and, more surprisingly, passes if the checks are removed
completely. There have likely been changes to the navigation handling that
handle the test in a different way. That said, it still appears to be important
to keep the checks there in some capacity because it does affect how many
navigation events occur. 
* This addresses an issue that came up in #16710: https://github.com/angular/angular/issues/16710#issuecomment-634869739
* This also partially addresses #13586 in fixing history for imperative
navigations that are cancelled by guards.

### DIFF
--- a/packages/router/src/router.ts
+++ b/packages/router/src/router.ts
@@ -1128,10 +1128,13 @@ export class Router {
       priorPromise?: {resolve: any, reject: any, promise: Promise<boolean>}): Promise<boolean> {
     const lastNavigation = this.getTransition();
     // If the user triggers a navigation imperatively (e.g., by using navigateByUrl),
-    // and that navigation results in 'replaceState' that leads to the same URL,
-    // we should skip those.
+    // and that navigation results in another navigation that leads to the same URL,
+    // we should skip those. Additionally, we compare to `urlAfterRedirects` because
+    // that is the final result of the imperative navigation. If an imperative navigation is
+    // cancelled by a guard and followed by a back/forward button or manual URL change, we should
+    // not skip those navigations so we need to check against `urlAfterRedirects`.
     if (lastNavigation && source !== 'imperative' && lastNavigation.source === 'imperative' &&
-        lastNavigation.rawUrl.toString() === rawUrl.toString()) {
+        lastNavigation.urlAfterRedirects.toString() === rawUrl.toString()) {
       return Promise.resolve(true);  // return value is not used
     }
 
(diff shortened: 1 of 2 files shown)


atscott:
[presubmit](https://test.corp.google.com/ui#id=OCL:314434132:BASE:314586999:1591242294598:c4d93732) - 
one failing test that appears to be verifying incorrect behavior. It's a little complicated, but the TL;DR is that this is actually working as intended. A second navigation is triggered before flushing the second one that results from the location change. As a result, the ordering gets a little messed up when looking at `lastNavigation`. This PR actually demonstrates correct behavior for that test, even without the timeout flushing because `lastNavigation.urlAfterRedirects` is able to check what the browser URL is at the moment, not where the inserted nav is trying to go.

atscott:
[another global presubmit](https://test.corp.google.com/ui#id=OCL:314434132:BASE:314916222:1591394096069:14d49258)

packages/router/src/router.ts
```diff
@@ -1128,10 +1128,13 @@ export class Router {
       priorPromise?: {resolve: any, reject: any, promise: Promise<boolean>}): Promise<boolean> {
     const lastNavigation = this.getTransition();
     // If the user triggers a navigation imperatively (e.g., by using navigateByUrl),
-    // and that navigation results in 'replaceState' that leads to the same URL,
-    // we should skip those.
+    // and that navigation results in another navigation that leads to the same URL,
```
AndrewKushnir, line 1131: Thanks for adding more docs 👍

Minor proposal: can we update it a bit and may be present it as a list (and possibly sub-lists) of items, like a decision tree, for ex:
- ConditionA
  - if yes - we do this
  - if no - do smth else
- Condition B
  - ...

AndrewKushnir (approved):
Thanks for adding extra docs and tests @atscott 👍 LGTM, I've just left a minor comment.

PR #37421, opened 2020-06-03
feat(dev-infra): Add oauth scope check to ensure necessary permissions for merge tooling

josephperrott:
Adds an assertion that the provided TOKEN has OAuth scope permissions for `repo`
as this is required for all merge attempts.

On failure, provides detailed error message with remediation steps for the user.

Addresses DEV-108

### DIFF
--- a/dev-infra/pr/merge/git.ts
+++ b/dev-infra/pr/merge/git.ts
@@ -49,7 +49,7 @@ export class GitClient {
   api: Octokit;
 
   /** The OAuth scopes available for the provided Github token. */
-  private _oauthScopes: Promise<string[]>;
+  private _oauthScopes = Promise.resolve<string[]>([]);
   /** Regular expression that matches the provided Github token. */
   private _tokenRegex = new RegExp(this._githubToken, 'g');
 
--- a/dev-infra/pr/merge/git.ts
+++ b/dev-infra/pr/merge/git.ts
@@ -13,6 +13,11 @@ import {info, yellow} from '../../utils/console';
 
 import {MergeConfigWithRemote} from './config';
 
+/** Github response type extended to include the `x-oauth-scopes` headers presence. */
+type RateLimitResponseWithOAuthScopeHeader = Octokit.Response<Octokit.RateLimitGetResponse>&{
+  headers: {'x-oauth-scopes': string};
+};
+
 /** Error for failed Github API requests. */
 export class GithubApiRequestError extends Error {
   constructor(public status: number, message: string) {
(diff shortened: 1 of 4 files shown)


dev-infra/pr/merge/task.ts
```diff
@@ -48,6 +51,9 @@ export class PullRequestMergeTask {
    * @param force Whether non-critical pull request failures should be ignored.
    */
   async merge(prNumber: number, force = false): Promise<MergeResult> {
+    // Assert the authenticated GitClient has access on the required scopes.
+    await this.git.assertOauthScopes(...REQUIRED_SCOPES);
```
devversion, line 55: Can we only assert this, and perform the API request when the API merge strategy is use?

Ideally we'd just perform the request and assert the permissions as part of the strategy. We can have it inside `.prepare` where `prepare` returns a boolean?

josephperrott: We can move the assertion into the strategy definitions, however both autosquash and merge strategies require this scope since we define the remote as:
`https://<TOKEN>@github.com/<org>/<name>.git`

Alternatively we can require that the autosquash must use `ssh`, though personally I would want to move towards just requiring everyone to use `http` for this toolset.

devversion: Oh right. I forgot about the autosquash strategy also using the token. We should definitely keep that as is. My comment was based on the assumption that we only need the permission for the API merge, so we should be good having this here.

devversion: One suggestion though: Right now this is the only unhandled error in the merge script. We should just return a boolean/list of missing scopes and return a failure that is handled by the merge script CLI (as done with other errors).

dev-infra/pr/merge/git.ts
```diff
@@ -55,6 +57,13 @@ export class GitClient {
       // expect Github API errors better and in a non-ambiguous way.
       throw new GithubApiRequestError(error.status, error.message);
     });
+
+    // OAuth scopes are loaded via the /rate_limit endpoint to prevent
+    // usage of a request against that rate_limit for this lookup.
+    this._oauthScopes = this.api.request('/rate_limit').then(response => {
+      const scopes: string = response.headers['x-oauth-scopes'] || '';
+      return scopes.split(',').map(scope => scope.trim());
+    });
```
devversion, line 66: Should we do this lazily? The git client will be a general util and I don't think we'd need this request always.

devversion (approved):
LGTM! 🎉 Just two minor comments

josephperrott:
@devversion PTAL, as I ended up doing better failure case handling which led to a wider changeset.

devversion (approved):
Nice! I like it 😄 

Adding `cleanup` label as CI is red. Most likely we need to explicitly type `_oauthScopes` as nullable.

PR #37425, opened 2020-06-03
docs: update file header to be correct

atscott:
The file header should be Google LLC rather than Google Inc. because it is now an LLC after Alphabet Holdings was formed.


josephperrott (approved):
LGTM

PR #37430, opened 2020-06-04
fix(language-service): show element selectors only in element autocomplete

kyliau:
This commit fixes a bug whereby element in CSS selectors like `form` in
the selector of the `NgForm` directive

```ts
@Directive({
  selector: 'form:not([ngNoForm]):not([formGroup]),ng-form,[ngForm]',
  // ...
})
export class NgForm extends ControlContainer implements Form, AfterViewInit {
```

are inadvertently included in the autocomplete results for a HTML
element. Only component selectors and element selectors in directives
should be included in the results.

In addition, this commit assigns a selector to the `TemplateReference`
class. Without an explicit `selector`, Angular assigns the generic
name `ng-component`.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/language-service/src/completions.ts
+++ b/packages/language-service/src/completions.ts
@@ -15,7 +15,7 @@ import {getExpressionCompletions} from './expressions';
 import {attributeNames, elementNames, eventNames, propertyNames} from './html_info';
 import {InlineTemplate} from './template';
 import * as ng from './types';
-import {diagnosticInfoFromTemplateInfo, findTemplateAstAt, getPathToNodeAtPosition, getSelectors, inSpan, isStructuralDirective, spanOf} from './utils';
+import {diagnosticInfoFromTemplateInfo, findTemplateAstAt, getPathToNodeAtPosition, getSelectorMap, inSpan, isStructuralDirective, spanOf} from './utils';
 
 const HIDDEN_HTML_ELEMENTS: ReadonlySet<string> =
     new Set(['html', 'script', 'noscript', 'base', 'body', 'title', 'head', 'link']);
@@ -344,13 +344,14 @@ function elementCompletions(info: ng.AstResult): ng.CompletionEntry[] {
 
   // Collect the elements referenced by the selectors
   const components = new Set<string>();
-  for (const selector of getSelectors(info).selectors) {
+  const selectorMap = getSelectorMap(info.directives, true /* isElementSelector */);
+  for (const [selector, {isComponent}] of selectorMap) {
     const name = selector.element;
     if (name && !components.has(name)) {
       components.add(name);
       results.push({
         name,
-        kind: ng.CompletionKind.COMPONENT,
+        kind: isComponent ? ng.CompletionKind.COMPONENT : ng.CompletionKind.DIRECTIVE,
         sortText: name,
       });
     }
@@ -538,8 +539,8 @@ class ExpressionVisitor extends NullTemplateVisitor {
     const key = attr.name.substring(1);  // remove leading asterisk
 
     // Find the selector - eg ngFor, ngIf, etc
-    const selectorInfo = getSelectors(this.info);
-    const selector = selectorInfo.selectors.find(s => {
+    const selectorMap = getSelectorMap(this.info.directives, false /* isElementSelector */);
+    const selector = Array.from(selectorMap.keys()).find(s => {
       // attributes are listed in (attribute, value) pairs
       for (let i = 0; i < s.attrs.length; i += 2) {
         if (s.attrs[i] === key) {
@@ -562,7 +563,7 @@ class ExpressionVisitor extends NullTemplateVisitor {
       if (equalLocation > 0 && valueRelativePosition > equalLocation) {
         // We are after the '=' in a let clause. The valid values here are the members of the
         // template reference's type parameter.
-        const directiveMetadata = selectorInfo.map.get(selector);
+        const directiveMetadata = selectorMap.get(selector);
         if (directiveMetadata) {
           const contextTable =
               this.info.template.query.getTemplateContext(directiveMetadata.type.reference);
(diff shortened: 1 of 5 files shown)


JoostK:
Drive by comment: if you have a directive with selector `'app-directive'`, why wouldn't that classify as valid autocomplete suggestion? AFAIK there isn't much of a difference between directives and components, apart from the fact that components render their view into the host element. Or am I misunderstanding the proposed change here?

packages/language-service/src/completions.ts
```diff
@@ -344,7 +344,8 @@ function elementCompletions(info: ng.AstResult): ng.CompletionEntry[] {
 
   // Collect the elements referenced by the selectors
   const components = new Set<string>();
-  for (const selector of getSelectors(info).selectors) {
+  const {selectors} = getSelectors(info.directives.filter(isComponent));
+  for (const selector of selectors) {
```
ayazhafiz, line 348: I agree with Joost's comment. In 95% of cases if a directive has an element name selector, the selector also corresponds to an HTML element or a component. But then there are "special" cases like [`ng-form`](https://angular.io/api/forms/NgForm) which is neither a native HTML element or component, but still a valid Angular element name.

Here, could we conditionally set the `kind` of the added result based on whether it is a directive or component?

ayazhafiz (changes requested):

kyliau:
@ayazhafiz @JoostK My apologies, I didn't do a good job at explaining the bug, and the test actually adds confusion to the discussion rather than pinpointing the issue.

The problem I'm trying to solve is that, when autocompletion is requested at a position where a HTML element is expected, this is currently the list of items returned:

name | kind
------|------
ng-container | angular element
ng-content | angular element
ng-template | angular element
form | component
option | component
input | component
textarea | component
select | component
ng-form | component
my-app | component
my-comp | component
test-comp | component

We can see that `form`, `option`, `input`, `textarea` and `select` are clearly not components, but they get included because they are part of the selectors in various Directives in `@angular/forms`.

For example, `form` and `ng-form` are extracted from the selector in `NgForm` directive:
https://github.com/angular/angular/blob/17996bf7196c4ad05b34bb752fc9ac0a9ee11f4e/packages/forms/src/directives/ng_form.ts#L92-L99

Note that these are completions for an **external** template, and language service do not provide autocomplete for native HTML elements by design (that's the responsibility of the HTML extension). Therefore, I'd like to remove them from the results. However, since the extraction logic deals solely with the selectors, there is not an easy way to distinguish html elements vs css selectors.

The solution I proposed here is to eliminate directives entirely, which solves the problem, but like both of you pointed out, could also lead to more removals than necessary. What do you think is the best approach here?

ayazhafiz:
> However, since the extraction logic deals solely with the selectors, there is not an easy way to distinguish html elements vs css selectors.

I think you could filter out selectors that are not only element selectors (with `selector.isElementSelector`) and then check if the the element is an HTML element with the [DomElementSchemaRegistry](https://github.com/angular/angular/blob/master/packages/compiler/src/schema/dom_element_schema_registry.ts). The TS host already creates one of these so it could be shared.

kyliau:
> I think you could filter out selectors that are not only element selectors (with `selector.isElementSelector`) and then check if the the element is an HTML element with the [DomElementSchemaRegistry](https://github.com/angular/angular/blob/master/packages/compiler/src/schema/dom_element_schema_registry.ts). The TS host already creates one of these so it could be shared.

Yes, I also found that after I commented about this! Thank you for the pointer :)
However, there're some tricky situations like
https://github.com/angular/angular/blob/aaa20093b23882f828ac1a96766bd4367d74aa14/packages/forms/src/directives/select_control_value_accessor.ts#L206-L207

In this case, I think it's ok to include the selector and not dedupe it against the DomRegistry.
I've added a new `CompletionKind.Directive` to denote that such selectors come from Directives. PTAL at the update code and commit message, thanks!

packages/language-service/src/utils.ts
```diff
@@ -64,17 +64,23 @@ export function isStructuralDirective(type: CompileTypeMetadata): boolean {
...
+export function getSelectorMap(directives: CompileDirectiveSummary[], isElementSelector: boolean) {
+  const selectorMap = new Map<CssSelector, CompileDirectiveSummary>();
+  for (const directive of directives) {
     const selectors: CssSelector[] = CssSelector.parse(directive.selector!);
     for (const selector of selectors) {
-      results.push(selector);
-      map.set(selector, directive);
+      if (selector.isElementSelector() !== isElementSelector) {
+        continue;
+      }
```
ayazhafiz, line 79: this will skip element selectors if `isElementSelector` is false :)

ayazhafiz (commented):
otherwise lgtm

packages/language-service/src/utils.ts
kyliau: that's what I intended.
if `isElementSelectors` is true, return only element selectors
if `isElementSelectors` is false, return anything that's not element selectors  (css selectors)

ayazhafiz: why would we want to exclude element sectors? I guess looking at the other usages this makes sense, but can you document it on the parameter JSdoc?

ayazhafiz (approved):

packages/language-service/src/utils.ts
kyliau: The other use cases are attribute and microsyntax autocomplete, which also call this function.
I don't think element selectors make sense in cases where an attribute is expected.
Like `<div |>`, I don't think we should suggest `ng-form` at the cursor location.
Perhaps @JoostK can chime in and correct me if I'm wrong?

ayazhafiz: yep, makes sense

packages/language-service/test/project/app/parsing-cases.ts
```diff
@@ -82,6 +82,7 @@ export class TestComponent {
 } /*EndTestComponent*/
 
 @Component({
+  selector: 'my-comp',
```
JoostK, line 85: Instead of assigning a selector to the component, I'd actually keep this without one as that is a valid scenario to have. If we don't want to suggest `ng-component`, the LS should filter it out. The default `ng-component` comes from `ElementSchemaRegistry.getDefaultComponentElementName` btw.

JoostK (changes requested):
The terminology used is misleading to me, however that is due to existing code being like this. All selectors are "element selectors" to me, as selectors only target elements in the first place. An element's name is typically referred to as its "tag", which would be a better name in place of "element".

I do not suggest to change it now as it would be a large change which also changes `@angular/compiler`, but it's something to keep in mind.

---

In the commit message, you still mention "component selectors and element selectors" which is no longer accurate I believe.

kyliau:
Closing this, since current effort is focusing on improving Ivy LS.

Issue #34675, opened 2020-01-08
Angular uses old URL params when calling router.navigate from a resolver


dzinxed:
# 🐞 Angular uses old URL params when calling router.navigate from a resolver

### Affected Package
Router. Specifically, [this line](https://github.com/angular/angular/blob/18d89c9c891844440a919b835c0f198a2b7395a6/packages/router/src/router.ts#L983)

### Is this a regression?
Probably not, the inconsitency is caused by urlUpdateStrategy, which is a recent thing.

### Description
When you call router.navigate from a resolver (e.g. when a resolver determines that the user does not have required permissions and you want to redirect to an error page), the query parameters of the URL are not taken from the URL. Rather, they are taken from rawUrlTree. This can cause a problem because the parameters might have changed with urlUpdateStrategy set to 'eager'.

## 🔬 Minimal Reproduction
Here's a stackblitz repro based on heroes app from the router tutorial:
https://stackblitz.com/edit/angular-router-redirect

The example is changed from [the original](https://angular.io/guide/router#wrap-up-and-final-app) in the following ways:
 - it keeps the user ID in a query param instead of the path param,
 - initially, it loads /superheroes?id=11,
 - for /superhero path, it adds a RedirectingResolver that, when id=12, redirects to /crisis-center,
 - it has urlUpdateStrategy: 'eager'.

*To trigger the error, follow these steps*:
1. Open https://stackblitz.com/edit/angular-router-redirect
1. You should start on the list of heroes (url is /superheroes?id=11), with hero 11 selected.
1. Click "Narco" (hero 12) (its URL is /superhero?id=12)
1. See that you get redirected to /crisis-center?id=11.

However, if you look at RedirectingResolver in heroes-routing.module.ts, you'll see that:
 - it checks for query parameter and determines it's equal to 12
 - it redirects to /crisis-center with merged query params

With urlUpdateStragegy: 'eager' in app-routing.module.ts, I would expect to land at /crisis-center?id=12.

## 🌍  Your Environment

**Angular Version:**
Angular 8.2.14 on StackBlitz.

## Fix proposal

This issue could be fixed by changing [this line](https://github.com/angular/angular/blob/18d89c9c891844440a919b835c0f198a2b7395a6/packages/router/src/router.ts#L983):

<code>const mergedTree = this.urlHandlingStrategy.merge(urlTree, this.rawUrlTree);</code>

to:

[...]

PR #37435, opened 2020-06-04
fix(router): merge query params from previous url after navigation from a resolver

dimakuba:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Router uses old query params after navigation from a resolver.

Issue Number: https://github.com/angular/angular/issues/34675

## What is the new behavior?
Merge query params from previous url after navigation from a resolver.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/router/src/router.ts
+++ b/packages/router/src/router.ts
@@ -577,6 +577,7 @@ export class Router {
                          rawUrl: rawUrlTree,
                          extras: {skipLocationChange, replaceUrl}
                        } = t;
+                       this.currentUrlTree = t.urlAfterRedirects;
                        return this.hooks.beforePreactivation(targetSnapshot!, {
                          navigationId,
                          appliedUrlTree,
@@ -699,7 +700,6 @@ export class Router {
                         URL and the RouterState, as well as updated the browser URL. All this should
                         happen *before* activating. */
                      tap((t: NavigationTransition) => {
-                       this.currentUrlTree = t.urlAfterRedirects;
                        this.rawUrlTree =
                            this.urlHandlingStrategy.merge(this.currentUrlTree, t.rawUrl);
 
--- a/packages/router/test/integration.spec.ts
+++ b/packages/router/test/integration.spec.ts
@@ -1556,15 +1556,28 @@ describe('Integration', () => {
       }
     }
 
+    @Injectable()
+    class ResolveEmptyAndRedirect implements Resolve<never> {
+      constructor(private router: Router) {}
+
+      resolve(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Observable<never> {
+        this.router.navigate(['/one'], {queryParamsHandling: 'merge'});
+
+        return EMPTY;
+      }
+    }
+
     beforeEach(() => {
       TestBed.configureTestingModule({
         providers: [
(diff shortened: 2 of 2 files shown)


atscott:
At first glance, this seems like it could break current behavior pretty significantly. `this.currentUrlTree` represents the url of the activated route and is currently set right before activating the route synchronously. The proposed solution in the issue report mentioned using `browserUrlTree` rather than `rawUrlTree`, which would respect the `eager`/`deferred` values for `urlUpdateStrategy` and seems reasonable. Did you happen to try that, and if so, did it not work as expected or break something?

dmkuba:
@atscott Hi, `browserUrlTree` updates accordingly to `urlUpdateStrategy`, so if this strategy will be set to `deferred` the problem is still present with the proposed solution, because `currentUrlTree` isn’t stored before preactivation to determine query params for the next navigation

atscott:
> @atscott Hi, `browserUrlTree` updates accordingly to `urlUpdateStrategy`, so if this strategy will be set to `deferred` the problem is still present with the proposed solution, because `currentUrlTree` isn’t stored before preactivation to determine query params for the next navigation

Right, so I guess my question is why would you consider that to not be working as intended?

dmkuba:
> > @atscott Hi, `browserUrlTree` updates accordingly to `urlUpdateStrategy`, so if this strategy will be set to `deferred` the problem is still present with the proposed solution, because `currentUrlTree` isn’t stored before preactivation to determine query params for the next navigation
> 
> Right, so I guess my question is why would you consider that to not be working as intended?

query params are computes within `createUrlTree` function:
https://github.com/angular/angular/blob/21d3d7eedcd17c2814948f2f7c3d875d7e1369f2/packages/router/src/router.ts#L988

which uses `currentUrlTree` to merge query params from previous url with a new one, so we need to store `currentUrlTree` before any guards/resolvers check to make sure we have relevant url 
at the point when `router.navigate(...)` requests from guards/resolvers

atscott:
I understand why it might be desirable to have this as the behavior, but I think it can be easily argued the other way. A user might not want guards and resolver navigations to merge with a navigation that hasn't completed or changed the url yet.

The biggest difficulty with this PR is how it changes what `currentUrlTree`  means. While it's technically private, it can be indirectly accessed through `isActive` and `url` so changing the time you update the value is definitely a breaking change that needs a lot of consideration. 

We would also need to look at how `currentUrlTree` is or is not reset on cancelled navigations. Note that a second call to `router.navigate` will result in the first one being cancelled (because the navigation pipe is a `switchMap`) but it will not go through the `catchError` recovery path for when `NavigationCancel` is thrown. This means that `currentUrlTree` won't be restored correctly in those situations.

I'm just trying to point out that almost every change to the router, unless it's _purely_ a bug fix for an obviously broken behavior (like generating urls that don't make sense and don't work), is a breaking change because it modifies how routing is handled. If we could convince ourselves that this is the right direction, the best we could do here would be to add a new configuration to the router to allow people to opt-in to the change.

dimakuba:
Thanks a lot for the review. You’re right: technically it could be correct solution, but these changes might break already bound projects to public api that are related to `currentUrlTree` in scope of guards & resolvers.

PR #37437, opened 2020-06-04
feat(language-service): TS references from template items

ayazhafiz:
Keen and I were talking about what it would take to support getting
references at a position in the current language service, since it's
unclear when more investment in the Ivy LS will be available. Getting TS
references from a template is trivial -- we simply need to get the
definition of a symbol, which is already handled by the language
service, and ask the TS language service to give us the references for
that definition.

This doesn't handle references in templates, but that could be done in a
subsequent pass.

Part of https://github.com/angular/vscode-ng-language-service/issues/29

## PR Checklist

## PR Type

- [x] Feature

- [ ] Yes
- [x] No

### DIFF
--- a/packages/language-service/src/typescript_host.ts
+++ b/packages/language-service/src/typescript_host.ts
@@ -359,6 +359,10 @@ export class TypeScriptServiceHost implements LanguageServiceHost {
     return program;
   }
 
+  getReferencesAtPosition(fileName: string, position: number): tss.ReferenceEntry[]|undefined {
+    return this.tsLS.getReferencesAtPosition(fileName, position);
+  }
+
   /**
    * Return the TemplateSource if `node` is a template node.
    *
--- a/packages/language-service/src/typescript_host.ts
+++ b/packages/language-service/src/typescript_host.ts
@@ -72,8 +72,7 @@ export class TypeScriptServiceHost implements LanguageServiceHost {
     ngModules: [],
   };
 
-  constructor(
-      readonly tsLsHost: tss.LanguageServiceHost, private readonly tsLS: tss.LanguageService) {
+  constructor(readonly tsLsHost: tss.LanguageServiceHost, readonly tsLS: tss.LanguageService) {
     this.summaryResolver = new AotSummaryResolver(
         {
           loadSummary(_filePath: string) {
@@ -357,10 +356,6 @@ export class TypeScriptServiceHost implements LanguageServiceHost {
       throw new Error('No program in language service!');
     }
     return program;
-  }
-
-  getReferencesAtPosition(fileName: string, position: number): tss.ReferenceEntry[]|undefined {
-    return this.tsLS.getReferencesAtPosition(fileName, position);
   }
 
   /**
--- a/packages/language-service/test/references_spec.ts
+++ b/packages/language-service/test/references_spec.ts
@@ -1,0 +1,57 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import * as ts from 'typescript';
+
+import {createLanguageService} from '../src/language_service';
+import {TypeScriptServiceHost} from '../src/typescript_host';
+
+import {MockTypescriptHost} from './test_utils';
+
+describe('references', () => {
+  const mockHost = new MockTypescriptHost(['/app/main.ts']);
+  const tsLS = ts.createLanguageService(mockHost);
+  const ngHost = new TypeScriptServiceHost(mockHost, tsLS);
+  const ngLS = createLanguageService(ngHost);
+
+  beforeEach(() => {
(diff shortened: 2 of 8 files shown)


packages/language-service/src/typescript_host.ts
```diff
@@ -359,6 +359,10 @@ export class TypeScriptServiceHost implements LanguageServiceHost {
     return program;
   }
 
+  getReferencesAtPosition(fileName: string, position: number): tss.ReferenceEntry[]|undefined {
+    return this.tsLS.getReferencesAtPosition(fileName, position);
```
kyliau, line 363: Let's just make `tsLS` public, since `tsLSHost` is already public.

packages/language-service/test/references_spec.ts
```diff
@@ -0,0 +1,57 @@
...
+  const ngHost = new TypeScriptServiceHost(mockHost, tsLS);
+  const ngLS = createLanguageService(ngHost);
+
+  beforeEach(() => {
+    mockHost.reset();
+  });
+
+  describe('component members', () => {
+    it('should get TS references for a member in template', () => {
+      const fileName = mockHost.addCode(`
```
kyliau, line 28: Could you please use `override` instead?

We could use the `content.substring()` method to verify the span instead of putting markers in the file.

packages/language-service/test/references_spec.ts
```diff
@@ -0,0 +1,57 @@
...
+  const mockHost = new MockTypescriptHost(['/app/main.ts']);
+  const tsLS = ts.createLanguageService(mockHost);
+  const ngHost = new TypeScriptServiceHost(mockHost, tsLS);
+  const ngLS = createLanguageService(ngHost);
+
+  beforeEach(() => {
+    mockHost.reset();
+  });
+
+  describe('component members', () => {
```
kyliau, line 26: Could you please also add a case for inline template via `overrideInlineTemplate`?

kyliau (changes requested):
Did you mean to not expose this for now?

We'll need to make `getReferencesAtPosition` part of the plugin for it to be externally available.
https://github.com/angular/angular/blob/c8f7fc22c77882ae4c737c76fda7d474bf5e87b2/packages/language-service/src/ts_plugin.ts#L90-L101

devversion (approved):
LGTM for bazel build change! (as always 😄 )

ayazhafiz:
> Did you mean to not expose this for now?
> 
> We'll need to make `getReferencesAtPosition` part of the plugin for it to be externally available.

I think it's okay to expose it in this package, it doesn't need to be known by the VSCode extension until it's in a more useful state IMO.

packages/language-service/test/references_spec.ts
```diff
@@ -0,0 +1,63 @@
...
+        mockHost.overrideInlineTemplate(APP_COMPONENT, template);
+        return APP_COMPONENT;
+      } else {
+        mockHost.override(TEST_TEMPLATE, template);
+        return TEST_TEMPLATE;
+      }
+    }
+  }
+
+  function getSource({fileName, textSpan}: ts.ReferenceEntry): string {
```
kyliau, line 59: Just curious, is `ts.ReferenceEntry.contextSpan` available? 
It'd be helpful to also assert it if it's available, since it shows the relevant declaration.

packages/language-service/test/references_spec.ts
```diff
@@ -0,0 +1,63 @@
...
+        it('should get TS references for a member in template', () => {
+          const fileName = overrideTemplate('{{«title»}}');
+          const marker = mockHost.getReferenceMarkerFor(fileName, 'title');
+          const references = ngLS.getReferencesAtPosition(fileName, marker.start)!;
+
+          expect(references).toBeDefined();
+          expect(references.length).toBe(2);
+
+          for (const reference of references) {
+            expect(getSource(reference)).toBe('title');
```
kyliau, line 41: Could you please add a brief comment where the two references are?

kyliau (approved):

packages/language-service/test/references_spec.ts
```diff
@@ -0,0 +1,63 @@
...
+          }
+        });
+      });
+    });
+
+    // TODO: override parsing-cases#TemplateReference for inline templates.
```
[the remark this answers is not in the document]
ayazhafiz: yeah but we can just overwrite and replace `templateUrl`

packages/language-service/test/references_spec.ts
```diff
@@ -0,0 +1,63 @@
...
+    const overrideTemplate = (template: string): string => {
```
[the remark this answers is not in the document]
ayazhafiz: I prefer this too but apparently this is not allowed when targeting ES5: https://github.com/microsoft/TypeScript/blob/master/lib/diagnosticMessages.generated.json#L209

packages/language-service/test/references_spec.ts
ayazhafiz: done

kyliau (approved):

PR #37439, opened 2020-06-04
fix(dev-infra): properly determine oauth scopes for git client token

devversion:
We recently added a better reporting mechanism for oauth tokens with https://github.com/angular/angular/commit/c025357fb86562ba2105ec2c9f55adb45cdb8327
in the dev-infra git util. Unfortunately the logic broke as part of addressing PR
review feedback. Right now, always the empty promise from `oauthScopes` will be
used as `getAuthScopes` considers it as the already-requested API value. This is not
the case as the default promise is also truthy. We should just fix this by making the
property nullable.


josephperrott (approved):
LGTM

PR #37444, opened 2020-06-04
refactor(dev-infra): use the `exec()` helper from `utils/shelljs` whenever possible

gkalpak:
There is an `exec()` helper provided by `utils/shelljs.ts`, which is a wrapper around ShellJS' `exec()` with some default options (currently `silent: true`). The intention is to avoid having to pass these options to every invocation of the `exec()` function.

This commit updates all code inside `dev-infra/` to use this helper whenever possible).

NOTE: For simplicity, the `utils/shelljs` helper does not support some of the less common call signatures of the original `exec()` helper, so in some cases we still need to use the original.

### DIFF
--- a/dev-infra/utils/shelljs.ts
+++ b/dev-infra/utils/shelljs.ts
@@ -6,9 +6,9 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {exec as _exec, ShellString} from 'shelljs';
+import {exec as _exec, ExecOptions, ShellString} from 'shelljs';
 
 /* Run an exec command as silent. */
-export function exec(cmd: string): ShellString {
-  return _exec(cmd, {silent: true});
+export function exec(cmd: string, opts?: ExecOptions&{async?: false}): ShellString {
+  return _exec(cmd, {silent: true, ...opts});
 }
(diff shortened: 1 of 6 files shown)


josephperrott (approved):
LGTM

dev-infra/utils/shelljs.ts
```diff
@@ -6,9 +6,9 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {exec as _exec, ShellString} from 'shelljs';
+import {exec as _exec, ExecOptions, ShellString} from 'shelljs';
 
 /* Run an exec command as silent. */
-export function exec(cmd: string): ShellString {
-  return _exec(cmd, {silent: true});
+export function exec(cmd: string, opts?: ExecOptions&{async?: false}): ShellString {
```
devversion, line 12: We could `Omit` the async option (and optionally also override it to false). That would hide it nicely I guess.

devversion (approved):

dev-infra/utils/shelljs.ts
gkalpak: I am not sure what you mean (and it's too late to make any changes to the PR now :grin:)

PR #26243, opened 2018-10-04
fix(router): fix some empty path related navigation bugs

miginmrs:
fix some empty path related navigation bugs

PR Close #26224, #10726

## PR Checklist
Please check if your PR fulfills the following requirements:

- [X] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [X] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

```
[X] Bugfix
[ ] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Documentation content changes
[ ] angular.io application / infrastructure changes
[ ] Other... Please describe:
```

## What is the current behavior?

Issue Number: #26224 , #10726

## What is the new behavior?

* The next path of a pathless activated route is no longer added to the url and the commands passed to the createUrlTree are no longer added as children of an undefined outlet of this path but as children of the actual url.
* The navigation to aux routes with empty-path parent route is now possible.

## Does this PR introduce a breaking change?
```
[ ] Yes
[X] No
```

PR #37446, opened 2020-06-04
fix(router): Fix relative link generation from empty path components

### LINKED ISSUE #13011: RouterLink: incorrect relative link if defined in component having empty path
```
[ ] feature request
```

**Current behavior**
In the case of such router config with ContainerComponent consuming empty path: 
```
const routes = [
  { 
    path: '', 
    component: ContainerComponent,
    children: [
      { path: 'a', component: AComponent },
      { path: 'b', component: BComponent },
    ]
  }
];
```
and ContainerComponent having relative links to children in template:

```
<a [routerLink]="['a']">Link to A</a>     (or  <a [routerLink]="['./a']">Link to A</a>)
<a [routerLink]="['b']">Link to B</a>
```

RouterLink directive generates incorrect links when user navigated to one of the child component. F.e. if user went to `/a` route, links would be like that: `/a/(a)` and `/a/(b)`.

**Expected behavior**

Relative links should be generated properly: `/a` and `/b`.

http://plnkr.co/edit/Lf2Vfe?p=preview
Click on any link. Then you could see that links rendered in AppComponent are correct, but not the ones in ContainerComponent (they both look like `/a/(a)` and `/a/(b)`).

* **Angular version:** 2.2.1

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

 
* **Language:** [all | TypeScript X.X | ES6/7 | ES5]

* **Node (for AoT issues):** 4.2.1

### LINKED ISSUE #35687: relative routing to a route with empty path causes weird urls and breaks the routing eventually
# 🐞 bug report

### Affected Package

### Is this a regression?

I dont know

### Description

relative navigation to a route with empty path causes the url to contain additional /() on the next navigation. In some constelations undefined can also appear in the url. When the user then does some more navigating, at some point the route won't recognize the url anymore and the app breaks.

## 🔬 Minimal Reproduction

https://stackblitz.com/edit/angular-lazy-route-relative-bug
I couldn't create the stackblitz with a base path (which I have in my app), so the routing imediatelly breaks, because the routerLink will generate the url as "//(lazy)".

## 🔥 Exception or Error

I don't get an error, just wrong urls, which then cause the app to break a few navigations later or on page reload.

## 🌍  Your Environment

**Angular Version:**
<pre><code>
Angular CLI: 8.3.0
Node: 12.9.0
OS: win32 x64
Angular: 8.2.3
... platform-server, router

@angular-devkit/architect          0.803.0
@angular-devkit/build-angular      0.803.0
@angular-devkit/build-ng-packagr   0.803.0
@angular-devkit/build-optimizer    0.803.0
@angular-devkit/build-webpack      0.803.0
@angular-devkit/core               8.3.0
@angular-devkit/schematics         8.3.0
@angular/cdk                       8.1.3
@angular/cli                       8.3.0
@angular/material                  8.1.3
@angular/material-moment-adapter   8.1.3
@ngtools/webpack                   8.3.0
@schematics/angular                8.3.0
@schematics/update                 0.803.0
ng-packagr                         5.5.0
rxjs                               6.5.2
typescript                         3.5.3
webpack                            4.39.2

</code></pre>

**Anything else relevant?**

I dont think that it is browser/os specific, but I had the problem with Chrome and Firefox on Windows 10

atscott:
Partial resubmit of #26243
Fixes incorrect url tree generation for empty path components with children.
Adds a test to demonstrate the failure of createUrlTree for those routes.
Fixes #13011
Fixes #35687

### DIFF
--- a/packages/router/src/create_url_tree.ts
+++ b/packages/router/src/create_url_tree.ts
@@ -148,7 +148,12 @@ function findStartingPosition(nav: Navigation, tree: UrlTree, route: ActivatedRo
   }
 
   if (route.snapshot._lastPathIndex === -1) {
-    return new Position(route.snapshot._urlSegment, true, 0);
+    const segmentGroup = route.snapshot._urlSegment;
+    // Pathless ActivatedRoute has _lastPathIndex === -1 but should not process children
+    // see issue #26224, #13011, #35687
+    // However, if the ActivatedRoute is the root we should process children like above.
+    const processChildren = segmentGroup === tree.root;
+    return new Position(segmentGroup, processChildren, 0);
   }
 
   const modifier = isMatrixParams(nav.commands[0]) ? 0 : 1;
--- a/packages/router/test/create_url_tree.spec.ts
+++ b/packages/router/test/create_url_tree.spec.ts
@@ -11,7 +11,7 @@ import {BehaviorSubject} from 'rxjs';
 import {createUrlTree} from '../src/create_url_tree';
 import {ActivatedRoute, ActivatedRouteSnapshot, advanceActivatedRoute} from '../src/router_state';
 import {Params, PRIMARY_OUTLET} from '../src/shared';
-import {DefaultUrlSerializer, UrlSegmentGroup, UrlTree} from '../src/url_tree';
+import {DefaultUrlSerializer, UrlSegment, UrlSegmentGroup, UrlTree} from '../src/url_tree';
 
 describe('createUrlTree', () => {
   const serializer = new DefaultUrlSerializer();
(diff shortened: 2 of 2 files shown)


villelahdenvuo (approved):

atscott:
There are two reproductions (https://stackblitz.com/edit/angular-lazy-route-relative-bug, https://stackblitz.com/edit/relative-link-eager-loading) in #35687 that aren't fixed by the change as it is now. However, both of those would work if the fix were changed to 
`return new Position(segmentGroup, false, 0);`
instead of
`return new Position(segmentGroup, segmentGroup.segments.length === 0, 0);`

Adding cleanup, blocked, and presubmit label to indicate that I need to investigate if this fix can be extended to include #35687, adding a unit test to cover that situation, and running presubmits to ensure it doesn't break anything.

atscott:
[global presubmit](https://test.corp.google.com/ui#id=OCL:314828528:BASE:314916222:1591394097001:ed7957e2)

atscott:
[global presubmit](https://test.corp.google.com/ui#id=OCL:314962601:BASE:315182571:1591565450001:fa1b9fed) with additional change (`processChildren = segmentGroup === tree.root` instead of `segmentGroup.segments.length === 0`) to address #35687.

Some things to note for review:
* All three added tests *fail* without the change
* The generated links without the change are completely wrong (`/a/(b)` and `//(lazy)`) meaning this shouldn't be behavior anyone could have been relying on and this is purely a bug fix
* Global presubmit shows no failures

AndrewKushnir (approved):
LGTM 👍 Thanks for the fix @atscott!

atscott:
merge assistance: I'm the router owner so pullapprove won't be happy without global approval

PR #37311, opened 2020-05-27
feat(common): add ReadonlyMap in place of Map in keyValuePipe

ajitsinghkaler:
ReadonlyMap is a superset of Map, in keyValuePipe we do not change the value of the object so ReadonlyPipe Works right in this case and we can accomodate more types. To accomodate more types added ReadonlyMap in Key Value pipe.

Fixes #37308

## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [ ] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] angular.io application / infrastructure changes
- [ ] Other... Please describe:

## What is the current behavior?

Only map support in keyValue pipe

Issue Number: #37308 

## What is the new behavior?
ReadonlyMap support in keyValue pipe

## Does this PR introduce a breaking change?

- [ ] Yes
- [ ] No

PR #37447, opened 2020-06-04
Improve typing of common pipes

### LINKED ISSUE #36404: add Promise<T> | Observable<T> async pipe overload
# 🚀 feature request

### Relevant Package
`@angular/common`

### Description
Currently the async pipe can take a Promise or Observable, but only as seperate obverloads, so with `strictTemplates` you get an error if you pass in a `Promise<T> | Observable<T>`

Add an overload which looks like `transform<T>(obj: Promise<T> | Observable<T> | null | undefined): T | null;`

converting the promise into a observable, but this isn't ergonomic

Happy to submit a patch

ranma42:
## PR Checklist

## PR Type

- [x] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #36259
Several pipes have incomplete or incorrect typing.

## What is the new behavior?
The common pipes should now all have a stricter typing.

- [X] Yes
- [x] No

It depends on what is considered a breaking change: some code that would compile will now trigger a compilation error. On the other hand, the code that is made invalid by this change, such as passing an object to a numeric pipe, was incorrect in the first place (it caused errors at runtime instead of at compile time).

Fixing the fallout should be trivial.

## Other information
Originally posted as https://github.com/angular/angular/pull/36277

### DIFF
--- a/goldens/public-api/common/common.d.ts
+++ b/goldens/public-api/common/common.d.ts
@@ -194,6 +194,8 @@ export declare abstract class LocationStrategy {
 
 export declare class LowerCasePipe implements PipeTransform {
     transform(value: string): string;
+    transform(value: null | undefined): null;
+    transform(value: string | null | undefined): string | null;
 }
 
 export declare class NgClass implements DoCheck {
@@ -392,6 +394,8 @@ export declare type Time = {
 
 export declare class TitleCasePipe implements PipeTransform {
     transform(value: string): string;
+    transform(value: null | undefined): null;
+    transform(value: string | null | undefined): string | null;
 }
 
 export declare enum TranslationWidth {
@@ -403,6 +407,8 @@ export declare enum TranslationWidth {
 
 export declare class UpperCasePipe implements PipeTransform {
     transform(value: string): string;
+    transform(value: null | undefined): null;
+    transform(value: string | null | undefined): string | null;
 }
 
 export declare const VERSION: Version;
--- a/goldens/public-api/common/common.d.ts
+++ b/goldens/public-api/common/common.d.ts
@@ -381,9 +381,10 @@ export declare function registerLocaleData(data: any, localeId?: string | any, e
 
 export declare class SlicePipe implements PipeTransform {
     transform<T>(value: ReadonlyArray<T>, start: number, end?: number): Array<T>;
+    transform(value: null | undefined, start: number, end?: number): null;
+    transform<T>(value: ReadonlyArray<T> | null | undefined, start: number, end?: number): Array<T> | null;
     transform(value: string, start: number, end?: number): string;
-    transform(value: null, start: number, end?: number): null;
-    transform(value: undefined, start: number, end?: number): undefined;
+    transform(value: string | null | undefined, start: number, end?: number): string | null;
 }
 
 export declare type Time = {
--- a/goldens/public-api/common/common.d.ts
+++ b/goldens/public-api/common/common.d.ts
@@ -13,7 +13,9 @@ export declare class CommonModule {
 
 export declare class CurrencyPipe implements PipeTransform {
     constructor(_locale: string, _defaultCurrencyCode?: string);
-    transform(value: any, currencyCode?: string, display?: 'code' | 'symbol' | 'symbol-narrow' | string | boolean, digitsInfo?: string, locale?: string): string | null;
+    transform(value: number | string, currencyCode?: string, display?: 'code' | 'symbol' | 'symbol-narrow' | string | boolean, digitsInfo?: string, locale?: string): string | null;
+    transform(value: null | undefined, currencyCode?: string, display?: 'code' | 'symbol' | 'symbol-narrow' | string | boolean, digitsInfo?: string, locale?: string): null;
+    transform(value: number | string | null | undefined, currencyCode?: string, display?: 'code' | 'symbol' | 'symbol-narrow' | string | boolean, digitsInfo?: string, locale?: string): string | null;
 }
 
 export declare class DatePipe implements PipeTransform {
@@ -25,7 +27,9 @@ export declare class DatePipe implements PipeTransform {
 
 export declare class DecimalPipe implements PipeTransform {
     constructor(_locale: string);
-    transform(value: any, digitsInfo?: string, locale?: string): string | null;
+    transform(value: number | string, digitsInfo?: string, locale?: string): string | null;
+    transform(value: null | undefined, digitsInfo?: string, locale?: string): null;
+    transform(value: number | string | null | undefined, digitsInfo?: string, locale?: string): string | null;
 }
 
 export declare const DOCUMENT: InjectionToken<Document>;
@@ -342,7 +346,9 @@ export declare class PathLocationStrategy extends LocationStrategy {
 
 export declare class PercentPipe implements PipeTransform {
     constructor(_locale: string);
-    transform(value: any, digitsInfo?: string, locale?: string): string | null;
+    transform(value: number | string, digitsInfo?: string, locale?: string): string | null;
+    transform(value: null | undefined, digitsInfo?: string, locale?: string): null;
+    transform(value: number | string | null | undefined, digitsInfo?: string, locale?: string): string | null;
 }
 
 export declare abstract class PlatformLocation {
--- a/goldens/public-api/common/common.d.ts
+++ b/goldens/public-api/common/common.d.ts
@@ -140,21 +140,13 @@ export declare interface KeyValue<K, V> {
 
 export declare class KeyValuePipe implements PipeTransform {
     constructor(differs: KeyValueDiffers);
-    transform<K, V>(input: null, compareFn?: (a: KeyValue<K, V>, b: KeyValue<K, V>) => number): null;
-    transform<V>(input: {
-        [key: string]: V;
-    } | ReadonlyMap<string, V>, compareFn?: (a: KeyValue<string, V>, b: KeyValue<string, V>) => number): Array<KeyValue<string, V>>;
-    transform<V>(input: {
-        [key: string]: V;
-    } | ReadonlyMap<string, V> | null, compareFn?: (a: KeyValue<string, V>, b: KeyValue<string, V>) => number): Array<KeyValue<string, V>> | null;
-    transform<V>(input: {
-        [key: number]: V;
-    } | ReadonlyMap<number, V>, compareFn?: (a: KeyValue<number, V>, b: KeyValue<number, V>) => number): Array<KeyValue<number, V>>;
-    transform<V>(input: {
-        [key: number]: V;
-    } | ReadonlyMap<number, V> | null, compareFn?: (a: KeyValue<number, V>, b: KeyValue<number, V>) => number): Array<KeyValue<number, V>> | null;
     transform<K, V>(input: ReadonlyMap<K, V>, compareFn?: (a: KeyValue<K, V>, b: KeyValue<K, V>) => number): Array<KeyValue<K, V>>;
-    transform<K, V>(input: ReadonlyMap<K, V> | null, compareFn?: (a: KeyValue<K, V>, b: KeyValue<K, V>) => number): Array<KeyValue<K, V>> | null;
+    transform<K extends number, V>(input: Record<K, V>, compareFn?: (a: KeyValue<string, V>, b: KeyValue<string, V>) => number): Array<KeyValue<string, V>>;
+    transform<K extends string, V>(input: Record<K, V> | ReadonlyMap<K, V>, compareFn?: (a: KeyValue<K, V>, b: KeyValue<K, V>) => number): Array<KeyValue<K, V>>;
+    transform(input: null | undefined, compareFn?: (a: KeyValue<unknown, unknown>, b: KeyValue<unknown, unknown>) => number): null;
+    transform<K, V>(input: ReadonlyMap<K, V> | null | undefined, compareFn?: (a: KeyValue<K, V>, b: KeyValue<K, V>) => number): Array<KeyValue<K, V>> | null;
+    transform<K extends number, V>(input: Record<K, V> | null | undefined, compareFn?: (a: KeyValue<string, V>, b: KeyValue<string, V>) => number): Array<KeyValue<string, V>> | null;
+    transform<K extends string, V>(input: Record<K, V> | ReadonlyMap<K, V> | null | undefined, compareFn?: (a: KeyValue<K, V>, b: KeyValue<K, V>) => number): Array<KeyValue<K, V>> | null;
 }
 
 export declare class Location {
--- a/goldens/public-api/common/common.d.ts
+++ b/goldens/public-api/common/common.d.ts
@@ -18,7 +18,9 @@ export declare class CurrencyPipe implements PipeTransform {
 
 export declare class DatePipe implements PipeTransform {
     constructor(locale: string);
-    transform(value: any, format?: string, timezone?: string, locale?: string): string | null;
+    transform(value: Date | string | number, format?: string, timezone?: string, locale?: string): string | null;
+    transform(value: null | undefined, format?: string, timezone?: string, locale?: string): null;
+    transform(value: Date | string | number | null | undefined, format?: string, timezone?: string, locale?: string): string | null;
 }
 
 export declare class DecimalPipe implements PipeTransform {
--- a/goldens/public-api/common/common.d.ts
+++ b/goldens/public-api/common/common.d.ts
@@ -110,7 +110,7 @@ export declare class HashLocationStrategy extends LocationStrategy {
 
 export declare class I18nPluralPipe implements PipeTransform {
     constructor(_localization: NgLocalization);
-    transform(value: number, pluralMap: {
+    transform(value: number | null | undefined, pluralMap: {
         [count: string]: string;
     }, locale?: string): string;
 }
--- a/goldens/public-api/common/common.d.ts
+++ b/goldens/public-api/common/common.d.ts
@@ -3,10 +3,9 @@ export declare const APP_BASE_HREF: InjectionToken<string>;
 export declare class AsyncPipe implements OnDestroy, PipeTransform {
     constructor(_ref: ChangeDetectorRef);
     ngOnDestroy(): void;
-    transform<T>(obj: null): null;
-    transform<T>(obj: undefined): undefined;
-    transform<T>(obj: Observable<T> | null | undefined): T | null;
-    transform<T>(obj: Promise<T> | null | undefined): T | null;
+    transform<T>(obj: Observable<T> | Promise<T>): T | null;
+    transform<T>(obj: null | undefined): null;
+    transform<T>(obj: Observable<T> | Promise<T> | null | undefined): T | null;
 }
 
 export declare class CommonModule {
--- a/packages/common/src/pipes/async_pipe.ts
+++ b/packages/common/src/pipes/async_pipe.ts
@@ -94,11 +94,10 @@ export class AsyncPipe implements OnDestroy, PipeTransform {
     }
   }
 
-  transform<T>(obj: null): null;
-  transform<T>(obj: undefined): undefined;
-  transform<T>(obj: Observable<T>|null|undefined): T|null;
-  transform<T>(obj: Promise<T>|null|undefined): T|null;
-  transform(obj: Observable<any>|Promise<any>|null|undefined): any {
+  transform<T>(obj: Observable<T>|Promise<T>): T|null;
+  transform<T>(obj: null|undefined): null;
+  transform<T>(obj: Observable<T>|Promise<T>|null|undefined): T|null;
+  transform<T>(obj: Observable<T>|Promise<T>|null|undefined): T|null {
     if (!this._obj) {
       if (obj) {
         this._subscribe(obj);
@@ -108,7 +107,7 @@ export class AsyncPipe implements OnDestroy, PipeTransform {
 
     if (obj !== this._obj) {
       this._dispose();
-      return this.transform(obj as any);
+      return this.transform(obj);
     }
 
     return this._latestValue;
--- a/packages/common/src/pipes/case_conversion_pipes.ts
+++ b/packages/common/src/pipes/case_conversion_pipes.ts
@@ -29,8 +29,11 @@ export class LowerCasePipe implements PipeTransform {
   /**
    * @param value The string to transform to lower case.
    */
-  transform(value: string): string {
-    if (!value) return value;
+  transform(value: string): string;
+  transform(value: null|undefined): null;
+  transform(value: string|null|undefined): string|null;
+  transform(value: string|null|undefined): string|null {
+    if (value == null) return null;
     if (typeof value !== 'string') {
       throw invalidPipeArgumentError(LowerCasePipe, value);
     }
@@ -72,8 +75,11 @@ export class TitleCasePipe implements PipeTransform {
   /**
    * @param value The string to transform to title case.
    */
-  transform(value: string): string {
-    if (!value) return value;
+  transform(value: string): string;
+  transform(value: null|undefined): null;
+  transform(value: string|null|undefined): string|null;
+  transform(value: string|null|undefined): string|null {
+    if (value == null) return null;
     if (typeof value !== 'string') {
       throw invalidPipeArgumentError(TitleCasePipe, value);
     }
@@ -96,8 +102,11 @@ export class UpperCasePipe implements PipeTransform {
   /**
    * @param value The string to transform to upper case.
    */
-  transform(value: string): string {
-    if (!value) return value;
+  transform(value: string): string;
+  transform(value: null|undefined): null;
+  transform(value: string|null|undefined): string|null;
+  transform(value: string|null|undefined): string|null {
+    if (value == null) return null;
     if (typeof value !== 'string') {
       throw invalidPipeArgumentError(UpperCasePipe, value);
     }
--- a/packages/common/src/pipes/keyvalue_pipe.ts
+++ b/packages/common/src/pipes/keyvalue_pipe.ts
@@ -50,31 +50,35 @@ export class KeyValuePipe implements PipeTransform {
   private differ!: KeyValueDiffer<any, any>;
   private keyValues: Array<KeyValue<any, any>> = [];
 
-  transform<K, V>(input: null, compareFn?: (a: KeyValue<K, V>, b: KeyValue<K, V>) => number): null;
-  transform<V>(
-      input: {[key: string]: V}|ReadonlyMap<string, V>,
-      compareFn?: (a: KeyValue<string, V>, b: KeyValue<string, V>) => number):
-      Array<KeyValue<string, V>>;
-  transform<V>(
-      input: {[key: string]: V}|ReadonlyMap<string, V>|null,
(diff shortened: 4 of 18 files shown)


kyliau:
Thank you for reopening the PR @ranma42!

For context, this is a continuation of https://github.com/angular/angular/pull/36277. It was accidentally closed by me due to my silly mistake in `git push`!

packages/common/src/pipes/number_pipe.ts
```diff
@@ -246,8 +248,8 @@ export class CurrencyPipe implements PipeTransform {
   }
 }
 
-function isEmpty(value: any): boolean {
-  return value == null || value === '' || value !== value;
+function isValue(value: number|string|null|undefined): value is number|string {
+  return !(value == null || value === '' || value !== value);
```
ayazhafiz, line 252: nit: I think this is fine to keep as is, just change the return type to be `value is null|undefined`. Alternatively, you could use the [nullish coalescing](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#nullish-coalescing) operator to get rid of this method entirely.

packages/common/src/pipes/slice_pipe.ts
```diff
@@ -61,12 +61,13 @@ export class SlicePipe implements PipeTransform {
...
   transform<T>(value: ReadonlyArray<T>, start: number, end?: number): Array<T>;
+  transform(value: string|null|undefined, start: number, end?: number): string|null;
   transform(value: string, start: number, end?: number): string;
-  transform(value: null, start: number, end?: number): null;
-  transform(value: undefined, start: number, end?: number): undefined;
-  transform(value: any, start: number, end?: number): any {
-    if (value == null) return value;
+  transform<T>(value: ReadonlyArray<T>|string|null|undefined, start: number, end?: number):
+      Array<T>|string|null {
+    if (value == null) return null;
```
ayazhafiz, line 70: Prefer strict equality

```suggestion
    if (value === null || value === undefined) return null;
```

it's a nit but it makes the intention clearer IMO

packages/common/test/pipes/keyvalue_pipe_spec.ts
```diff
@@ -17,7 +17,7 @@ describe('KeyValuePipe', () => {
   });
   it('should return null when given undefined', () => {
     const pipe = new KeyValuePipe(defaultKeyValueDiffers);
-    expect(pipe.transform(undefined as any)).toEqual(null);
+    expect(pipe.transform(undefined)).toEqual(null);
```
ayazhafiz, line 20: :+1: love that the cast can be dropped here

packages/language-service/src/typescript_symbols.ts
```diff
@@ -229,6 +229,24 @@ function selectSignature(type: ts.Type, context: TypeContext, _types: Symbol[]):
   // Consider creating a TypeCheckBlock host in the language service that may also act as a
   // scratchpad for type comparisons.
   const signatures = type.getCallSignatures();
+  // Try to select a matching signature in which all parameter types match.
+  const passedInTypes: Array<ts.Type|undefined> = _types.map(type => {
```
ayazhafiz, line 233: Can we rename `_types` to `types`? I think I added this to make TS stop complaining about an unused variable..

packages/language-service/src/typescript_symbols.ts
```diff
@@ -229,6 +229,24 @@ function selectSignature(type: ts.Type, context: TypeContext, _types: Symbol[]):
...
+      return type.tsType;
+    }
+  });
+  function allParametersMatch(signature: ts.Signature) {
+    const tc = context.checker;
+    return signature.getParameters().every((parameter: ts.Symbol, i: number) => {
+      const type = tc.getTypeOfSymbolAtLocation(parameter, parameter.valueDeclaration);
+      return type === passedInTypes[i];
+    });
+  }
```
ayazhafiz, line 244: Can you add a comment that this is still kind of a hack, because we're checking for strict type equality rather than compatibility. For example, if we the signature contains a `ReadonlyArray<int>` and the passed parameter type is an `Array<int>`, this will fail.

ayazhafiz (changes requested):

packages/language-service/src/typescript_symbols.ts
kyliau: The initial comment about the most appropriate approach to take still stands:
https://github.com/angular/angular/blob/3569fdf45195c36713e43a91a4feba4fa3c9facf/packages/language-service/src/typescript_symbols.ts#L227-L230
I've added additional comments below to make it clear that this is just a best-effort approach over what we currently have.

kyliau:
@ranma42 I've fixed up the language service part, should be good to go now.
@petebacondarwin Could you please take another look? Sorry I messed up the earlier PR and accidentally closed it.

packages/language-service/src/typescript_symbols.ts
ayazhafiz: Forgot that was there, thanks!

kyliau (approved):
LGTM for language service

packages/common/src/pipes/number_pipe.ts
ranma42: I would rather avoid the `value is null|undefined` route, as the signature would be very misleading: value could also be `''` (which could be added to the signature) or `NaN` (which cannot be expressed as a literal type).

It is not clear to me how to use the `??` to get rid of this function... wouldn't you still need to check for `''` and `NaN` at every location?

packages/common/src/pipes/slice_pipe.ts
ranma42: Right now all of the (common) pipes that work on scalar values check for `== null`. Should I change it also in the other ones?

ayazhafiz: didn’t realize it was a convention, disregard

ayazhafiz (approved):

packages/common/src/pipes/case_conversion_pipes.ts
```diff
@@ -29,8 +29,10 @@ export class LowerCasePipe implements PipeTransform {
   /**
    * @param value The string to transform to lower case.
    */
-  transform(value: string): string {
-    if (!value) return value;
+  transform(value: string|null|undefined): string|null;
```
petebacondarwin, line 32: Is it worth adding yet another overload of

```ts
transform(value: null|undefined): null;
```

petebacondarwin: And similarly elsewhere.

(Disclaimer: I have not checked whether this would break the type system...)

petebacondarwin (approved):
Apart from the additional suggested overloads, which IMO are optional for this PR, LGTM

Reviewed-for: fw-i18n

packages/common/src/pipes/case_conversion_pipes.ts
ranma42: I tried locally and it looks like it should be fine, apart from a couple of language-service tests that would need to be updated (they check that there is exactly one type for a pipe, while this change would make the resolver find two types).
Should I add it to this branch or is it better to do it in another PR?
In the first case, should I squash the changes or add them on top? (what is best for reviews/the convention for this repo?)

ranma42: > Is it worth adding yet another overload of
> 
> ```ts
> transform(value: null|undefined): null;
> ```

I had omitted them assuming that it is a niche use case, but it certainly makes typing more accurate and I guess it might help spotting issues in some buggy code.

petebacondarwin: Not a blocker for this PR from my point of view. I just mention it because, this PR involves a public API change, and a second PR would also do so, which is extra work checking the API each time.

I would rather get @IgorMinar another public API approver to comment before you make any further changes.

jelbourn:
It looks like the changes from `any` to more strict types are breaking API changes- am I interpreting that right?

IgorMinar (commented):
Thank you @ranma42 for the PR but this is definitely a breaking change and because we are already in the v10 RC period we will have to wait for the master branch to open for v11 changes before we can entertain this change.

I skimmed through the diff and at the high level this looks like a good change to me. Let's revisit this once we are working on v11.

ranma42:
Yup, as mentioned in the checklist the typing change is a breaking change.
@IgorMinar should I do something to make the merge happen in time for v11?
(this branch sat idle for about 2 months and only lately it got some movement)

ranma42:
Is now a good time to start working on this again?
Should I rebase and/or add the `transform(value: null|undefined): null;` specializations?

petebacondarwin:
Now that 10.0.0 is out, the `master` branch is now open for new PRs so this PR is unblocked.
@ranma42 - please do rebase and let's see what we can do to move it forward.

ranma42:
Nice! Rebased :)

jelbourn:
@petebacondarwin since this is a breaking change, I think this wouldn't be able to move closer until we're on the v11 release train, no?

petebacondarwin:
Oops! Sorry. I forgot we are on minors for a bit now.

ranma42:
It looks like the branch does not merge cleanly anymore; should I rebase it now or wait for master to be open to breaking changes?

Also, would it make sense to split the breaking changes and the compatible ones in two separate branches, so that the latter can proceed "immediately" while the breaking ones wait for 11?
(This would obviously require twice as many reviews, but each would be on a smaller change)

petebacondarwin:
Whether to rebase now or later is up to you. No doubt it will need another rebase closer to its merge date. But it may be that rebasing regularly will make each conflict resolution simpler? Or maybe it is just adding more work...

Given that we are entering the summer when there are naturally fewer resources to handle reviewing and merging and v11 might be open for merging early Autumn, I would just sit on this whole PR until then, unless you desperately need some of the non-breaking changes in the next release.

ranma42:
I would hate to miss again the merge window, but I am not in a rush to get this merged (the projects I work on that would benefit from this are mostly fine with some additional casts).

petebacondarwin (approved):
Reviewed-for: public-api

atscott:
Also fixes #36404

mhevery (approved):

mhevery (approved):
reviewed-for: global-approvers
reviewed-for: fw-core

packages/common/src/pipes/keyvalue_pipe.ts
```diff
@@ -50,30 +50,29 @@ export class KeyValuePipe implements PipeTransform {
   private differ!: KeyValueDiffer<any, any>;
   private keyValues: Array<KeyValue<any, any>> = [];
 
-  transform<K, V>(input: null, compareFn?: (a: KeyValue<K, V>, b: KeyValue<K, V>) => number): null;
+  transform<K, V>(
```
dev054, line 53: Could we also support `ReadonlyMap` here?

ranma42: Yes :)
That has already been changed in https://github.com/angular/angular/pull/37311 and will stay like that when these changes are rebased

ranma42:
rebased

ranma42:
@petebacondarwin this PR was assigned new tags, so I rebased it.

Note that I only fixed the conflicts; I did not address [this comment](https://github.com/angular/angular/pull/37447#discussion_r436912138), but I can certainly do it if it it is desired 😄

petebacondarwin:
Thanks for keeping this PR up to date @ranma42 - it just needs to pass the "presubmit" which is where we run it against a subset of the Google projects that use Angular internally.

packages/common/src/pipes/async_pipe.ts
```diff
@@ -94,10 +94,7 @@ export class AsyncPipe implements OnDestroy, PipeTransform {
     }
   }
 
-  transform<T>(obj: null): null;
-  transform<T>(obj: undefined): undefined;
-  transform<T>(obj: Observable<T>|null|undefined): T|null;
-  transform<T>(obj: Promise<T>|null|undefined): T|null;
+  transform<T>(obj: Observable<T>|Promise<T>|null|undefined): T|null;
   transform(obj: Observable<any>|Promise<any>|null|undefined): any {
```
ranma42, line 98: since the object wrapping has been removed, we might avoid even just use the "strict" signature, if we like it better:

```suggestion
  transform<T>(obj: Observable<T>|Promise<T>|null|undefined): T|null {
```

note that this would not change anything for the API consumers

AndrewKushnir:
[Presubmit](http://test/OCL:332024221:BASE:332024251:1600274148829:43188e36).

AndrewKushnir:
Hi @ranma42, quick update: we ran tests in Google's codebase and found out that there are some part of code that require updates in order to land this change. We'll be look into these cases and keep this thread updated on the progress.

This PR introduces a breaking change, so we need to include the `BREAKING CHANGE` section into one of the commit message to describe what the change is and what developers should do to resolve the problem is they come across it. You can find an example of the breaking changes note in https://github.com/angular/angular/commit/6acea54f62fe7e62e9b36458b30d98501277236f.

Thank you.

ranma42:
> Hi @ranma42, quick update: we ran tests in Google's codebase and found out that there are some part of code that require updates in order to land this change. We'll be look into these cases and keep this thread updated on the progress.
> 
> This PR introduces a breaking change, so we need to include the `BREAKING CHANGE` section into one of the commit message to describe what the change is and what developers should do to resolve the problem is they come across it.

Several commits introduce some kind of breaking change.
They are somewhat related, as they are all typing changes on the pipes, but each is independent and possibly different.
It is better to add the `BREAKING CHANGE` section to each of them (explaining what that specific commit is breaking) or should I put it in the first/last commit of the branch?

petebacondarwin:
If it makes sense for the changes to be in separate commits then they should each get their own BREAKING CHANGE (BC) block that is specific to that change. This has two benefits:

* the change log will have seperate breaking changes items for each one, which makes it easier for the reader to find BCs specific to their needs
* if. for some reason, a commit needs to be reverted, the other commits (and their BCs) are not affected.

If you feel that this is overkill then perhaps the commits should be merged?

AndrewKushnir:
Hi @ranma42, I performed further testing and came across an issue with the `TitleCasePipe` (and it might be applicable to some other pipes as well).

Here is a quick example of a pipe, similar to what we have for the `TitleCasePipe` one (from types perspective) in this PR, but a simplified version ([TypeScript playground page](https://www.typescriptlang.org/play?#code/MYGwhgzhAECyCeAFAlgBwKbQN4ChrQBcAnMAOwgDMB7IgWwAoA3MEAV3QC5oJjlSBzAD6lWIEINakAJugp90UgJRceRPkJFiA3HkIly1OkxbsVvAcu7n+O-MTKUaDZm05W1A4aPGSZc0gqWqupeYti6+MgU0Mau0AC88dCaIIrQROgErESkyd62+OmZ2bku7AUAvjhVOMBU5ATQfDxkwJhJAQDucEho6PSKOnUNRRCiBACCEADK1glNDa3oAHT2Bk70AORgAEbAm4PQAPRH0OgAHhjABApcGWMgkzNzABaQ0AAGwQIfhPAYABpoGBrqwWFwvnNBHkxB9avUeKNxlMAHLeebNAhLVb6RxGFKHE5nS7oa63JGPVHot4wD4pX4Ef7oIEgrLgz7ffjQaH0oA)):
```
class SimplePipe {
  transform(value: string|null|undefined): string|null;
  transform(value: string): string;
  transform(value: string|null|undefined): string|null {
    if (value == null) return null;
    return value;
  }
}

const pipe = new SimplePipe();
const result = pipe.transform('abc');
```

What happens here is that the `result` has the `string | null` type, when the expectation is that it should have just `string` (based on the second overload). I believe we should avoid this overlap (so that 2 overloads match the value type) to resolve the issue, so the code of the pipe might look like this ([TypeScript playground page](https://www.typescriptlang.org/play?#code/MYGwhgzhAECyCeAFAlgBwKbQN4ChrQBcAnMAOwgDMB7IgWwAoA3MEAV3QC5oJjlSBzAJRceRPvwDceQiXLU6TFuy6lWIEAB9WpACboKfdDuHRV6qfmJlKNBszaduvARrObteg6SMnR412og2NL4yBTQig7QALzRpoGC0EToBKxEpPHmIUkpaRn27BbQAL44pTjAVOQE0Hw8ZMCYcd4A7nBIaOj0glKV1TkQagQAghAAys78MbXVDegAdFZytvQA5GAARsCrPdAA9HvQ6AAeGMAERtNg56wsXMmDICPjk9AAFpDQAAZ+Al+E8AwFSqPAGQ1GADlAtM6gQ5otZDYFG5dgcjqd0OdLnFrqk7mCnpDoR8YF83P8CID0EA)):

```
class SimplePipe {
  transform(value: string): string;
  transform(value: null|undefined): null;
  transform(value: string|null|undefined): string|null {
    if (value == null) return null;
    return value;
  }
}

const pipe = new SimplePipe();
const result = pipe.transform('abc');
```

What do you think?

I'm adding the "blocked" label for now so that we resolve this question before adding this PR to the merge queue.

// cc @petebacondarwin

JoostK:
@AndrewKushnir I think you'll want to just swap the first and second signatures, such that the `transform(value: string): string` becomes the preferred signature. In your proposal, someone would get type errors for inputs of type `string|null`, as none of the signatures allow for that combination.

AndrewKushnir:
Thanks @JoostK, that makes sense 👍 

@ranma42, it looks like the following pipes are affected by this problem:

- KeyValuePipe
- LowerCasePipe
- TitleCasePipe
- UpperCasePipe
- SlicePipe

So we need to move more specific overloads for these pipes higher on the overload list.

ranma42:
you are right, the overloads should be listed from the most specific to the most generic

ranma42:
I tried to fix all of the issues you noticed, namely:
 - bad overload ordering (TypeScript resolves on a first-matching basis, so the specific overloads must appear before the generic ones)
 - incomplete typing around `null | undefined`
 - missing notices about `BREAKING CHANGES`

In doing so I also noticed that the typing of `KeyValuePipe` was still incorrect: if it is passed a value which is typed as `Record<number, V>` it actually returns the keys as strings.
The typing has been updated accordingly (the tests already check for this).

AndrewKushnir:
Hi @ranma42, thanks for updating the changes 👍 

FYI, I've rebased this PR on top of the latest master (the CI didn't start because the previous base commit was too old). The rebase was successful, no additional changes were made.

Since there were some changes, I'd like to ask @petebacondarwin and @JoostK to have a quick look at the changes once again to make sure everything is ok.

Thank you.

AndrewKushnir:
I've also started [tests in Google's codebase](http://test/OCL:332024221:BASE:332754939:1600646445846:9358fba5) and will keep this thread updated.

ranma42:
I just noticed that in one of the commit messages I skipped a word :(

"The typing of the `keyvalue` has been fixed..."
instead of
"The typing of the `keyvalue` pipi has been fixed..."

packages/common/src/pipes/async_pipe.ts
```diff
@@ -108,7 +107,7 @@ export class AsyncPipe implements OnDestroy, PipeTransform {
 
     if (obj !== this._obj) {
       this._dispose();
-      return this.transform(obj as any);
+      return this.transform(obj);
```
petebacondarwin, line 110: 🚀

packages/common/src/pipes/number_pipe.ts
```diff
@@ -246,8 +266,8 @@ export class CurrencyPipe implements PipeTransform {
...
```
petebacondarwin, line 270: Do we not prefer use of `isNaN`?

```suggestion
  return !(value == null || value === '' || isNaN(value));
```

goldens/public-api/common/common.d.ts
```diff
@@ -110,7 +110,7 @@ export declare class HashLocationStrategy extends LocationStrategy {
 
 export declare class I18nPluralPipe implements PipeTransform {
     constructor(_localization: NgLocalization);
-    transform(value: number, pluralMap: {
+    transform(value: number | null | undefined, pluralMap: {
```
petebacondarwin, line 113: We could go crazy here and do something like:

```suggestion
    transform(value: number, ...): string;
    transform(value: null | undefined, ...): '';
    transform(value: number | null | undefined, ...): string;
```

petebacondarwin (approved):
Reviewed-for: public-api

Nice work @ranma42  - thanks for your patience.

A typo in commit 1:

> instead ~~reflcts~~reflects the behaviour on `null` and `undefined`.

And a few nits and suggestions.

packages/common/src/pipes/keyvalue_pipe.ts
```diff
@@ -50,31 +50,30 @@ export class KeyValuePipe implements PipeTransform {
...
-      Array<KeyValue<number, V>>;
-  transform<V>(
-      input: {[key: number]: V}|ReadonlyMap<number, V>|null,
-      compareFn?: (a: KeyValue<number, V>, b: KeyValue<number, V>) => number):
-      Array<KeyValue<number, V>>|null;
   transform<K, V>(
       input: ReadonlyMap<K, V>,
       compareFn?: (a: KeyValue<K, V>, b: KeyValue<K, V>) => number): Array<KeyValue<K, V>>;
+  transform<K extends number, V>(
+      input: Record<K, V>, compareFn?: (a: KeyValue<string, V>, b: KeyValue<string, V>) => number):
```
JoostK, line 57: Shouldn't this be
```suggestion
      input: Record<K, V>, compareFn?: (a: KeyValue<number, V>, b: KeyValue<number, V>) => number):
```

Which makes it equivalent to the next overload with `K extends string|number`.

packages/common/src/pipes/keyvalue_pipe.ts
```diff
@@ -50,31 +50,30 @@ export class KeyValuePipe implements PipeTransform {
...
+  transform(
+      input: null|undefined,
+      compareFn?: (a: KeyValue<unknown, unknown>, b: KeyValue<unknown, unknown>) => number): null;
   transform<K, V>(
-      input: ReadonlyMap<K, V>|null,
+      input: ReadonlyMap<K, V>|null|undefined,
+      compareFn?: (a: KeyValue<K, V>, b: KeyValue<K, V>) => number): Array<KeyValue<K, V>>|null;
+  transform<K extends number, V>(
+      input: Record<K, V>|null|undefined,
+      compareFn?: (a: KeyValue<string, V>, b: KeyValue<string, V>) => number):
```
JoostK, line 70: Same here

packages/common/src/pipes/number_pipe.ts
```diff
@@ -67,8 +67,12 @@ export class DecimalPipe implements PipeTransform {
    * When not supplied, uses the value of `LOCALE_ID`, which is `en-US` by default.
    * See [Setting your app locale](guide/i18n#setting-up-the-locale-of-your-app).
    */
-  transform(value: any, digitsInfo?: string, locale?: string): string|null {
-    if (isEmpty(value)) return null;
+  transform(value: number|string, digitsInfo?: string, locale?: string): string|null;
```
JoostK, line 70: I think this signature is not technically required, as its return type is `string|null` just like the third signature and the ordering doesn't matter in this case (i.e. this signature can be moved one down without changing the matching behavior, and can then be merged with the 3rd signature). I don't particularly mind, and from a docs perspective it may seem weird that the primary signature only takes `null|undefined`, so I am fine with keeping as is.

This comment also applies to some other pipes below.

packages/common/test/pipes/number_pipe_spec.ts
```diff
@@ -125,8 +149,20 @@ import {beforeEach, describe, expect, it} from '@angular/core/testing/src/testin
...
+          expect(pipe.transform(null)).toEqual(null);
+        });
+
+        it('should return null for undefined', () => {
+          expect(pipe.transform(undefined)).toEqual(null);
+        });
+
         it('should not support other objects', () => {
-          expect(() => pipe.transform({}))
+          expect(() => pipe.transform(<any>{}))
```
JoostK, line 165: nit: we typically use `as` casts.

```suggestion
          expect(() => pipe.transform({} as any))
```

packages/common/src/pipes/number_pipe.ts
petebacondarwin: I agree it is not really needed but it makes the overload options clearer to the user of the code.

goldens/public-api/common/common.d.ts
ranma42: sure! :) should I go forward with that?

packages/common/src/pipes/keyvalue_pipe.ts
ranma42: no, the `keyvalue` pipe passes the `number` keys as strings in the case of a `Record<number, V>` object :(

I tried to explain it in the commit message, but maybe it is not explicit enough (and maybe we should also add a comment in the code to explain it, so it does not get lost in the git history)

packages/common/src/pipes/keyvalue_pipe.ts
ranma42: as above

packages/common/src/pipes/number_pipe.ts
ranma42: Removing it also makes the pipe very weird when applied to `any`, as in that case TypeScript uses the first overload (since any overload matches `any`) instead of the most generic one.

This is very inconvenient when the first overload handles "special" cases (`null | undefined`) as it causes the result to be typed `null` as well

packages/common/test/pipes/number_pipe_spec.ts
ranma42: oops! this was based on the `should not support other objects` from `case_conversion_pipes_spec.ts`, I did not know about the new convention.

I will fix the casts am I adding/modifying.
Should I also update the existing `<xyz>` casts in the files I an touching?

ranma42: I tried enabling the `no-angle-bracket-type-assertion` lint, but it looks like the codebase is still using them in several places.
I guess that kind of "global" refactor is better handled in another (non-BC) pull request.

In the meantime I will fix them in the code I am touching, so that PR will be a little bit simpler ;)

JoostK: Yes that sounds like the correct approach, thanks!

packages/common/src/pipes/number_pipe.ts
ranma42: TypeScript rejects it because `value` is typed as `number|string` at that point.

packages/common/test/pipes/number_pipe_spec.ts
JoostK: Oh by the way, these could be improved even further in the future using `// @ts-expect-error` instead of the `any` cast.

goldens/public-api/common/common.d.ts
ranma42: In a way, it looks like a good idea, as it provides a type-level "documentation" of the behavior in these special cases.
OTOH, it looks like a strange commitment from the perspective of the API

JoostK (approved):
LGTM, thanks for putting this together!

AndrewKushnir:
[Presubmit](http://test/OCL:332024221:BASE:333330678:1600884669816:b2811fb9).

AndrewKushnir:
Hi @ranma42, just want to let you know that this PR was merged into master branch and the change will be included into upcoming v11 release. Thank you for contributing to Angular!

PR #37462, opened 2020-06-05
fix(dev-infra): properly determine oauth scopes for git client token

devversion:
Resubmit of b2bd38699b0595d0a35b501f20251f8715a9fd1c since
85b6c94cc6d4dd1cfae68fc14575d0efa57574b3 accidentally reverted
the fix due to rebasing most likely.


josephperrott (approved):
LGTM

PR #37470, opened 2020-06-06
fix(router): stricter typing for Route.pathMatch

### LINKED ISSUE #37469: router: stricter typing for Route.pathMatch
# 🚀 feature request

### Relevant Package

 This feature request is for @angular/router

### Description

`Route.pathMatch` field is now typed as `string | undefined` but actually it can accept `prefix` or `full` as a valid value.
I want to restrict invalid values statically at compilation step with a string literal types.

Typing `pathMatch` as `pathMatch?: 'prefix' | 'full'`

This is a breaking change because it may cause build error by narrowing the acceptable value (even if those already have caused runtime error).

lacolaco:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #37469

## What is the new behavior?

```
interface Route {
  pathMatch?: 'prefix' | 'full';
}
```

- [x] Yes
- [ ] No

### DIFF
--- a/goldens/public-api/router/router.d.ts
+++ b/goldens/public-api/router/router.d.ts
@@ -302,7 +302,7 @@ export declare interface Route {
     matcher?: UrlMatcher;
     outlet?: string;
     path?: string;
-    pathMatch?: string;
+    pathMatch?: 'prefix' | 'full';
     redirectTo?: string;
     resolve?: ResolveData;
     runGuardsAndResolvers?: RunGuardsAndResolvers;
--- a/modules/playground/src/routing/app/inbox-app.ts
+++ b/modules/playground/src/routing/app/inbox-app.ts
@@ -8,7 +8,7 @@
 
 
 import {Component, Injectable} from '@angular/core';
-import {ActivatedRoute, Router} from '@angular/router';
+import {ActivatedRoute, Router, Routes} from '@angular/router';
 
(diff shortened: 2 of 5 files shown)


atscott:
Hi @lacolaco, thanks for the PR to fix this. I've added this to our v11 candidates tracking. This is the right change, but would be breaking so we'll have to wait until the master branch is accepting changes for v11 to merge it.

IgorMinar (approved):
LGTM for v10.0 unless google3 reveals that this change breaks the world. If we don't manage to land this in time for v10, this will have to wait for v11 because if we use the strictest interpretation this is a breaking change. However it should be so rare and if it does break apps, it should only reveal pre-existing fatal bugs - for this reason I'm ok with this landing in v10.0.

Reviewed-for: public-api

IgorMinar:
I'm labeling this as master&patch becase we don't yet have `target: rc` label.

atscott (approved):
Adding my LGTM as this is the right change, but it's still blocked since it's a breaking change and needs to target v11. We would also need to determine if we can get a migration for this or if g3 needs a patch to still accept `string`.

Airblader:
Duplicate of #35950?

atscott:
Duplicate of #35950

PR #37475, opened 2020-06-07
test(language-service): Remove all markers from test project

kyliau:
This commit removes all markers from the inline template in
`AppComponent` and external template in `TemplateReference`.

Test scenarios should be colocated with the test cases themselves.
Besides, many existing cases are not setup correctly.
For example, if we want to test autocomplete for HTML element, the existing test case is like:
```
<~{cursor} h1>
```
This doesn't make much sense, because the language service already sees
the `h1` tag in the template. The more representative test case should be:
```
<~{cursor
```
IMO, this reflects the real-world use case better.

This commit also uncovers a bug in the way HTML entities autocompletion
is done. There's an off-by-one error in which a cursor that is placed immediately after the ampersand character fails to trigger autocomplete.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/language-service/src/completions.ts
+++ b/packages/language-service/src/completions.ts
@@ -361,12 +361,18 @@ function elementCompletions(info: ng.AstResult): ng.CompletionEntry[] {
 
 function entityCompletions(value: string, position: number): ng.CompletionEntry[] {
   // Look for entity completions
+  // TODO(kyliau): revisit the usefulness of this feature. It provides
+  // autocompletion for HTML entities, which IMO is outside the core functionality
+  // of Angular language service. Besides, we do not have a complete list.
+  // See https://dev.w3.org/html5/html-author/charref
   const re = /&[A-Za-z]*;?(?!\d)/g;
   let found: RegExpExecArray|null;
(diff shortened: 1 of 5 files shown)


packages/language-service/src/completions.ts
```diff
@@ -361,12 +361,18 @@ function elementCompletions(info: ng.AstResult): ng.CompletionEntry[] {
 
 function entityCompletions(value: string, position: number): ng.CompletionEntry[] {
   // Look for entity completions
+  // TODO(kyliau): revisit the usefulness of this feature. It provides
```
ayazhafiz, line 364: I think it’s fine to drop this, IMO entities are the kind of thing where either you know them or you don’t so without also providing information about what the entity is rendered as this not very helpful.

ayazhafiz (approved):

packages/language-service/src/completions.ts
kyliau: Sounds good, I'll do this in a separate PR.

PR #37480, opened 2020-06-07
feat(router): add new initialNavigation options to replace legacy

CaerusKaru:
As of Angular v4, four of the options for
`ExtraOptions#initialNavigation` have been deprecated. We intend
to remove them in v11. The final state for these options is:
`enabledBlocking`, `enabledNonBlocking`, and `disabled`. We plan
to remove and deprecate the remaining option in the next two
major releases.

New options:
- `enabledNonBlocking`: same as legacy_enabled
- `enabledBlocking`: same as enabled

BREAKING CHANGE:

* The `initialNavigation` property for the options in
  `RouterModule.forRoot` no longer supports `legacy_disabled`,
  `legacy_enabled`, `true`, or `false` as valid values.
  `legacy_enabled` (the old default) is instead `enabledNonBlocking`
* `enabled` is deprecated as a valid value for the
  `RouterModule.forRoot` `initialNavigation` option. `enabledBlocking`
  has been introduced to replace it

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [x] Yes
- [ ] No

### DIFF
--- a/goldens/public-api/router/router.d.ts
+++ b/goldens/public-api/router/router.d.ts
@@ -156,7 +156,7 @@ export declare class GuardsCheckStart extends RouterEvent {
     toString(): string;
 }
 
-export declare type InitialNavigation = true | false | 'enabled' | 'disabled' | 'legacy_enabled' | 'legacy_disabled';
+export declare type InitialNavigation = 'disabled' | 'enabled' | 'enabledBlocking' | 'enabledNonBlocking';
 
 export declare type LoadChildren = LoadChildrenCallback | DeprecatedLoadChildren;
 
--- a/packages/router/src/router_module.ts
+++ b/packages/router/src/router_module.ts
@@ -222,7 +222,10 @@ export function provideRoutes(routes: Routes): any {
  * Allowed values in an `ExtraOptions` object that configure
  * when the router performs the initial navigation operation.
  *
- * * 'enabled' - The initial navigation starts before the root component is created.
+ * * 'enabledNonBlocking' - (default) The initial navigation starts after the
+ * root component has been created. The bootstrap is not blocked on the completion of the initial
+ * navigation.
+ * * 'enabledBlocking' - The initial navigation starts before the root component is created.
  * The bootstrap is blocked until the initial navigation is complete. This value is required
  * for [server-side rendering](guide/universal) to work.
  * * 'disabled' - The initial navigation is not performed. The location listener is set up before
@@ -230,24 +233,16 @@ export function provideRoutes(routes: Routes): any {
  * more control over when the router starts its initial navigation due to some complex
  * initialization logic.
  *
- * The following values have been [deprecated](guide/releases#deprecation-practices) since v4,
+ * The following values have been [deprecated](guide/releases#deprecation-practices) since v11,
  * and should not be used for new applications.
  *
- * * 'legacy_enabled'- (Default, for compatibility.) The initial navigation starts after the root
- * component has been created. The bootstrap is not blocked until the initial navigation is
- * complete.
- * * 'legacy_disabled'- The initial navigation is not performed. The location listener is set up
- * after the root component gets created.
- * * `true` - same as 'legacy_enabled'.
- * * `false` - same as 'legacy_disabled'.
- *
- * The 'legacy_enabled' and 'legacy_disabled' should not be used for new applications.
+ * * 'enabled' - This option is 1:1 replaceable with `enabledNonBlocking`.
  *
  * @see `forRoot()`
  *
  * @publicApi
  */
-export type InitialNavigation = true|false|'enabled'|'disabled'|'legacy_enabled'|'legacy_disabled';
+export type InitialNavigation = 'disabled'|'enabled'|'enabledBlocking'|'enabledNonBlocking';
 
 /**
  * A set of configuration options for a router module, provided in the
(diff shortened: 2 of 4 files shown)


atscott (approved):

jelbourn (commented):
If there any deprecation documentation or similar we can link to in the commit message here?

packages/router/src/router_module.ts
```diff
@@ -236,6 +236,10 @@ export function provideRoutes(routes: Routes): any {
  * the root component gets created. Use if there is a reason to have
  * more control over when the router starts its initial navigation due to some complex
  * initialization logic.
+ * * 'enabledNonBlocking' - (Will be default as of v11) The initial navigation starts after the
+ * root component has been created. The bootstrap is not blocked until the initial navigation is
+ * complete.
+ * * 'enabledBlocking' - Same as `enabled`
```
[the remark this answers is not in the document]
atscott: Hey Jeremy, this change is an attempt to clean up all the weirdly named options. The goal is to have the final options be:
`'enabledBlocking'|'enabledNonBlocking'|'disabled'`
Unfortunately we can't get there in one swoop because of our deprecation policy. I wonder if we should have an exception to the policy when we have a 1:1 replacement with a migration - i.e. we just want to rename something in the public API.

packages/router/src/router_module.ts
```diff
@@ -512,13 +517,15 @@ export class RouterInitializer {
       const opts = this.injector.get(ROUTER_CONFIGURATION);
 
       if (this.isLegacyDisabled(opts) || this.isLegacyEnabled(opts)) {
+        opts.initialNavigation !== undefined &&
```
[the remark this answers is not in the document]
jelbourn: Follow-up to off-thread conversation: I think in this particular case, I would use an `if` statement. This is based on the guarded statement being especially long, vs e.g. `ngDevMode` which is always followed by a relatively short assert statement

jelbourn:
I have learned since the last time I looked at this that there's a deprecation guide that outlines deprecations and what people should do about them. Should the changes here be included in there?

CaerusKaru:
@jelbourn Since this PR does not actually deprecate/remove (and is outside a major version), I don’t think now is the time to add it. These options can be used independent of a deprecation plan. The deprecation guide update is captured in #33128, which was bumped from v10 to v11 and will land when we open up the v11 release train.

jelbourn (approved):
LGTM

Reviewed-for: public-api

packages/router/src/router_module.ts
```diff
@@ -250,7 +254,8 @@ export function provideRoutes(routes: Routes): any {
  *
  * @publicApi
  */
-export type InitialNavigation = true|false|'enabled'|'disabled'|'legacy_enabled'|'legacy_disabled';
+export type InitialNavigation = true|false|'enabled'|'enabledBlocking'|'enabledNonBlocking'|
```
atscott, line 257: I'm curious what the effect of the following would be:
```
export type InitialNavigation = 'enabledBlocking'|'enabledNonBlocking'|'disabled'|DeprecatedInitialNavigation;

/** @deprecated */
export type DeprecatedInitialNavigation = true | false | 'enabled' | 'legacy_enabled' | 'legacy_disabled';
```
This might make it more clear what our desired final options are.
I think we can also mark 'enabled' as deprecated in 10.1 -- the final removal would still be n+2 from 11.

goldens/public-api/router/router.d.ts
```diff
@@ -148,7 +148,8 @@ export declare class GuardsCheckStart extends RouterEvent {
     toString(): string;
 }
 
-export declare type InitialNavigation = true | false | 'enabled' | 'disabled' | 'legacy_enabled' | 'legacy_disabled';
+/** @deprecated */
```
atscott, line 151: Hmm, I wonder if we should also export `DeprecatedInitialNavigation` through the `index` so we can have documentation for it and it shows up on aio. `DeprecatedLoadChildren` could be used a precedent.

Other than that, I really like how this looks. It's very clear what the preferred options are and which ones are deprecated.

packages/router/src/router_module.ts
```diff
@@ -218,13 +218,20 @@ export function provideRoutes(routes: Routes): any {
...
+export type DeprecatedInitialNavigation = true|false|'enabled'|'legacy_enabled'|'legacy_disabled';
+
 /**
  * Allowed values in an `ExtraOptions` object that configure
  * when the router performs the initial navigation operation.
  *
  * * 'enabled' - The initial navigation starts before the root component is created.
  * The bootstrap is blocked until the initial navigation is complete. This value is required
- * for [server-side rendering](guide/universal) to work.
+ * for [server-side rendering](guide/universal) to work. @deprecated since v10.1
```
atscott, line 230: What do you think of moving this down with the rest of the deprecated options and just change the language from "The following values have been deprecated since v4" to just "The following values have been deprecated"

atscott (changes requested):
Oh, one last thing: can you please add a note to the `aio/content/guide/deprecations.md` about this (https://angular.io/guide/deprecations)?

petebacondarwin (dismissed):
Reviewed-for: public-api

I agree that we should also export `DeprecatedInitialNavigation` from the barrel so that it appears in the docs. Currently the name `DeprecatedInitialNavigation` appears there but it is not clickable and so it is not possible to tell what those options are. Having it in the docs would also make it easier for people to search for the values that have been deprecated.

Screenshot 2020-07-24 at 12 42 04

atscott (approved):
Reviewed-for: fw-router
Reviewed-for: public-api

packages/router/src/router_module.ts
```diff
@@ -272,24 +267,15 @@ export interface ExtraOptions {
...
-   * [server-side rendering](guide/universal) to work.
-   * When set to `disabled`, the initial navigation is not performed.
-   * The location listener is set up before the root component gets created.
-   * Use if there is a reason to have more control over when the router
+   * One of `enabled`, `enabledBlocking`, `enabledNonBlocking` or `disabled`.
+   * When set to `enabled` or `enabledBlocking`, the initial navigation starts before the root
+   * component is created. The bootstrap is blocked until the initial navigation is complete. This
+   * value is required for [server-side rendering](guide/universal) to work. When set to
+   * `enabledNonBlocking`, the initial navigation starts after the root component has been created.
+   * The bootstrap is not blocked until the initial navigation is complete. When set to `disabled`,
```
petebacondarwin, line 275: I think this could be a bit less ambiguous...

```suggestion
   * The bootstrap is not blocked waiting for the initial navigation is complete. When set to `disabled`,
```

petebacondarwin: Or

```suggestion
   * The bootstrap is not blocked on the completion of the initial navigation. When set to `disabled`,
```

petebacondarwin (approved):
Reviewed-for: public-api

Minor docs suggestions.

AndrewKushnir (approved):
Thanks for the changes @CaerusKaru 👍 

I've just left a couple comments, but all of them are minor.

Reviewed-for: public-api

AndrewKushnir (approved):
Adding a review from size tracking group...

Reviewed-for: size-tracking

atscott (approved):

atscott (approved):
reviewed-for: size-tracking

atscott:
[presubmit](https://test.corp.google.com/ui#id=OCL:336894045:BASE:336894983:1602607428681:dda85d02) (this will require a g3 patch, see cl/336894045)

PR #37483, opened 2020-06-08
perf ngClass batch add/remove classes operations

alan-agius4:
A 3P partner/client using Angular version 9 and Ivy reported that while they browser runtime became faster their server side runtime code is slower. Their tracked it down to `ngClass` with multiple classes.

Ex: `<div [ngClass]="{'class1 class2 class3’ : true}”></div>`

From there, myself and their engineers took 2 started to investigate the root cause, I did this on a toy app, while they did this on their real world app.

We both came to the same conclusion that the main bottleneck is that classes are added and removed individually, they did some patches to try to test what's the impact of added/removing classes all at once, and it resulted in 10% reduction of rendering time on the server.

Here are some matrix on my toy app for `<div [ngClass]="{'class1 class2 class3’ : true}”></div>` we get the following matrix on the server:

Current
```
_toggleClass 1.3ms
_applyKeyValueChanges 1.4ms
```

With this change
```
_toggleClass 0.9ms
_applyKeyValueChanges 1.0ms
```


packages/common/src/directives/ng_class.ts
```diff
@@ -137,24 +165,38 @@ export class NgClass implements DoCheck {
...
-          this._renderer.addClass(this._ngEl.nativeElement, klass);
-        } else {
-          this._renderer.removeClass(this._ngEl.nativeElement, klass);
-        }
-      });
+  private _toggleClass(klass: string|string[], enabled: boolean): void {
+    let classes = typeof klass === 'string' ? klass : klass.join(' ');
+    classes = classes.trim();
+    if (classes) {
+      const splitClasses = classes.split(/\s+/g);
```
kyliau, line 185: Could you please clarify how using the regex is different from the default behavior of `String.prototype.split()`?

atscott (changes requested):
There should be some added tests for the batching behavior in `ng_class_spec`, right?

alan-agius4:
@atscott, doesn't the existing tests already cover that multiple classes are added/removed to an element?

https://github.com/angular/angular/blob/650974e1b37fc1db6af93222a8eee25138e6492f/packages/common/test/directives/ng_class_spec.ts#L187-L195

packages/common/src/directives/ng_class.ts
alan-agius4: This was there, but my guess if because of multiple spaces.

Ex: `'foo  bar'.split(/\s+/g);`

atscott:
> @atscott, doesn't the existing tests already cover that multiple classes are added/removed to an element?
> 
> https://github.com/angular/angular/blob/650974e1b37fc1db6af93222a8eee25138e6492f/packages/common/test/directives/ng_class_spec.ts#L187-L195

Hmm, it sure does! I wonder if it would be useful to spy on the `Render2` `addClass`/`removeClass` and assert that it's only called once. 🤷

JoostK:
Note also that this is a breaking change, so I'm afraid it can't go in atm.

alan-agius4:
@JoostK what's the breaking change?

JoostK:
@alan-agius4  The `Renderer2`  interface is public API and it's expected to be implemented by third-parties, e.g. NativeScript. The changed method signature means that those implementors would need to be adjusted to deal with arrays, making this change breaking.

alan-agius4:
@JoostK right 😩

IgorMinar (approved):
LGTM for public api, but we should get @mhevery and @matsko review this as well since their styling refactor has resulted in NgClass (and NgStyle) to become mostly obsolete and we plan to deprecate them soon. This change still seems to be useful even with that deprecation in mind.

I wonder if the `[class]` binding could be modified in this way - @matsko and @mhevery know better if this would make sense.

And as pointed out by @JoostK this is a breaking change and can't be made in the v10.x branch any more, so this will have to wait for v11. In v11 we can make this change as long as we give heads up to known implementers of the `Render2` interface that they now have to support this new method signature.

In an API that others extend, we don't always have to follow the N+2 deprecation policy. And the rule of thumb we use to figure out what to do is:
- if an api is used by application developers and most library developers then we must create a new interface, deprecate the old one using N+2, and get everyone to use the new interface before we remove the old once once the deprecation period is over.
- if the api is used by only a handful of partners, we can extend the interface without the N+2 period as long as we work with the known parties to update their implementation of the interface to support the new signature.

In the case of `Renderer2`, NativeScript is the biggest implementer of this interface. Most of the other usages of `Renderer2` are applications using the API rather than extending it so they will not be affected by this change. So in short based on the information available to me, I believe that this change is a go for v11.

Reviewed-for: public-api

IgorMinar (approved):
Reviewed-for: size-tracking
Reviewed-for: dev-infra

alan-agius4:
Maybe we can use classList from the element directly in `ngClass` instead of using the renderer to avoid the breaking change?

Example from
```ts
this._renderer.removeClass(this._ngEl.nativeElement, klass);
```
to
```ts
this._ngEl.nativeElement.classList.remove(klass)
```

pkozlowski-opensource:
> Maybe we can use classList from the element directly in ngClass instead of using the renderer to avoid the breaking change?

@alan-agius4 this would be a breaking change as well since the operations that were previously sent to a renderer all of the sudden wouldn't, no?

alan-agius4:
@pkozlowski-opensource, indeed it can be if in the `addClass` you have some extra logic.

matsko (approved):
LGTM

packages/common/src/directives/ng_class.ts
```diff
@@ -137,24 +165,38 @@ export class NgClass implements DoCheck {
...
-    klass = klass.trim();
-    if (klass) {
-      klass.split(/\s+/g).forEach(klass => {
-        if (enabled) {
-          this._renderer.addClass(this._ngEl.nativeElement, klass);
-        } else {
-          this._renderer.removeClass(this._ngEl.nativeElement, klass);
-        }
-      });
+  private _toggleClass(klass: string|string[], enabled: boolean): void {
```
jelbourn, line 181:
```suggestion
  private _toggleClass(classNames: string|string[], enabled: boolean): void {
```

jelbourn, line 181: Could you add a JsDoc description to this method? This would be a good place to explain the goal of using a single `classList.toggle` operation

packages/common/src/directives/ng_class.ts
jelbourn, line 185: It seems strange to `join` and then immediately `split`. Would it work to do something like this instead?
```ts
let normalizedClassNames = typeof classNames === 'string' ? 
    classNames.trim() : 
    classNames.map(c => c.trim());
```

packages/platform-browser/src/dom/dom_renderer.ts
```diff
@@ -215,12 +215,20 @@ class DefaultDomRenderer2 implements Renderer2 {
     }
   }
 
-  addClass(el: any, name: string): void {
-    el.classList.add(name);
+  addClass(el: any, name: string|string[]): void {
+    if (typeof name === 'string') {
+      el.classList.add(name);
+    } else {
+      el.classList.add(...name);
```
jelbourn, line 222: Can you add a comment here (and the remove) that explicitly calls out that this requires a polyfill for IE11?

jelbourn (commented):
FYI @alan-agius4 it's hard to get the full idea for a PR from the commit messages when there are several so it would be great to still include a brief summary in the PR description. 

Can you point to anything that documents the magnitude improvement of using the classList APIs this way?

packages/platform-browser/src/dom/dom_renderer.ts
alan-agius4: Sure.

packages/common/src/directives/ng_class.ts
alan-agius4: The reason for that is because a `string` can contain a series of classes in this form `"klass1 klass2 klass3"`. Thus, we  always need to split `split`, even when we have a string.

alan-agius4:
@jelbourn, thanks for the feedback.

With regards to documentation about the improvements there isn't any. Mostly these improvements are for server side rendering were the DOM API is patched and add/remove operator is not optimised for repetitive operations. See: fgnass/domino#166. (While this helps this is not enough)

A 3P partner/client using Angular version 9 and Ivy reported that while they browser runtime became faster their server side runtime code is slower. Their tracked it down to `ngClass` with multiple classes.

Ex: `<div [ngClass]="{'class1 class2 class3’ : true}”></div>`

From there, myself and their engineers took 2 started to investigate the root cause, I did this on a toy app, while they did this on their real world app.

We both came to the same conclusion that the main bottleneck is that classes are added and removed individually, they did some patches to try to test what's the impact of added/removing classes all at once, and it resulted in 10% reduction of rendering time on the server.

Here are some matrix on my toy app for `<div [ngClass]="{'class1 class2 class3’ : true}”></div>` we get the following matrix on the server:

Current
```
_toggleClass 1.3ms
_applyKeyValueChanges 1.4ms
```

With this change
```
_toggleClass 0.9ms
_applyKeyValueChanges 1.0ms
```

packages/common/src/directives/ng_class.ts
alan-agius4: We don't use `classList.toggle`, but I think I got the point 😁

jelbourn:
@alan-agius4 Understood! Could you add a comment to the code that links to that domino issue as well as a very brief summary of that explanation about the client report? Everything else LGTM

alan-agius4:
@jelbourn, inline comment added.

packages/common/src/directives/ng_class.ts
```diff
@@ -137,24 +170,62 @@ export class NgClass implements DoCheck {
...
+      if (enabled) {
+        this._renderer.addClass(this._ngEl.nativeElement, splitClasses);
+      } else {
+        this._renderer.removeClass(this._ngEl.nativeElement, splitClasses);
+      }
+    }
+  }
+
+  /**
+   * This methode is use to batch and update/remove classes with a single operation.
```
jelbourn, line 208:
```suggestion
   * This method is use to batch and update/remove classes with a single operation.
```

jelbourn (approved):
LGTM

Reviewed-for: public-api
Reviewed-for: size-tracking

packages/common/test/directives/ng_class_spec.ts
```diff
@@ -390,6 +390,48 @@ import {async, ComponentFixture, TestBed} from '@angular/core/testing';
...
+
+        TestBed.configureTestingModule({declarations: [Cmp]});
+        const fixture = TestBed.createComponent(Cmp);
+        const renderer = fixture.componentRef.injector.get(Renderer2);
+        const removeClassSpy = spyOn(renderer, 'removeClass');
+        const element = fixture.nativeElement.querySelector('div');
+
+        fixture.detectChanges();
+        expect(removeClassSpy).toHaveBeenCalledWith(element, ['class1', 'class2', 'class3']);
+        expect(removeClassSpy).toHaveBeenCalledTimes(1);
```
pkozlowski-opensource, line 432: Could you please expand this test? I would like to see:
* toggle between true and false;
* verification that the desired classes were removed / added to the element

pkozlowski-opensource (approved):
LGTM overall but needs cleanup (rebase + stricter test I've requested).

TBH it would be great if we haven't had to do this change at all as it is a breaking change (so it can make it only for v11) and adds a bit more code / complexity (especially to NgClass). On top of this we plan to deprecate  NgClass / NgStyle so part of those changes will be null & void in the future.

I don't think I've got enough context here but maybe an alternative (better?) approach would be to try to improve things upstream (is https://github.com/fgnass/domino/pull/166 aiming at doing this?).

I'm approving this PR "with cleanup" as I don't want to block progress but yeh, I'm not excited about the breaking change, additional size and code complexity... Are there any alternatives that we could consider (ex. fixing upstream in domino)?

Reviewed-for: public-api
Reviewed-for: size-tracking

jelbourn:
Based on @pkozlowski-opensource's comments, I'm also wondering whether it's worth doing this now. How hard would it be to fix on the domino side?

alan-agius4:
@pkozlowski-opensource & @jelbourn, thanks for the feedback. 

There is already a PR for domino (https://github.com/fgnass/domino/pull/166) to improve the performance related to repeated operations. Despite of that, since the framework is invoking `addClass`, `removeClass` on the same node multiple times (One for each class on a given Node) there is still some overhead over calling addClass/removeClass with an array.  I will actually gather some more timings and numbers around the difference between this change and Dominos change. Likely though this will have to wait some time, as currently I am mostly focused on v10 issues, and seeing that this is for v11, there is no immediate urgency in tackling this right away.

That being said, in general I do think that irrespective of this fix/change, I think it is good to allow the renderer to add/remove multiple classes at once since this is part of the standard DOM API.

With regards to @pkozlowski-opensource comment about the deprecation and removal of NgClass/NgStyle, I am aware of that, but to completely remove those will take several major versions, and I am hoping that the alternate approach to these 2 directives will take into more into consideration also performance with mocked DOM APIs.

Side note: recently we also had chats around the possibility in replacing Domino with something which is more maintained and supported. However, from some initial benchmarking it still looks like Domino is one of the faster Node.JS DOM APIs.

@mgechev, can you kindly try to reach out to Domino and get https://github.com/fgnass/domino/pull/166 merged?

alan-agius4:
Closing as the performance fix from Domino has landed, which makes the timings acceptable.

alfaproject:
Wouldn't this still be a decent addition to the API?

PR #37486, opened 2020-06-08
docs: Add content for new `strict` mode for Angular CLI

aikithoughts:
In v10, the Angular CLI supports a strict mode, which turns
on additional flags for the TypeScript and Angular compilers.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

No documentation for strict mode.

Issue Number: N/A

## What is the new behavior?
Now we have documentation!

- [ ] Yes
- [X] No

### DIFF
--- a/aio/content/guide/file-structure.md
+++ b/aio/content/guide/file-structure.md
@@ -78,6 +78,12 @@ Files at the top level of `src/` support testing and running your application. S
 | `styles.sass`          | Lists CSS files that supply styles for a project. The extension reflects the style preprocessor you have configured for the project. |
 | `test.ts`              | The main entry point for your unit tests, with some Angular-specific configuration. You don't typically need to edit this file. |
 
+<div class="alert is-helpful">
+
+If you create an application using Angular's strict mode, you will also have an additional `package.json` file in the `src/app` directory. For more information, see [Strict mode](/guide/strict-mode).
+
+</div>
+
 {@a app-src}
 
 Inside the `src/` folder, the `app/` folder contains your project's logic and data.
@@ -90,7 +96,7 @@ Angular components, templates, and styles go here.
 | `app/app.component.css`     | Defines the base CSS stylesheet for the root `AppComponent`. |
 | `app/app.component.spec.ts` | Defines a unit test for the root `AppComponent`. |
 | `app/app.module.ts`         | Defines the root module, named `AppModule`, that tells Angular how to assemble the application. Initially declares only the `AppComponent`. As you add more components to the app, they must be declared here. |
-| `app/package.json`              | This file is generated only in applications created using `--strict` mode. This file is not used by package managers. It is used to tell the tools and bundlers whether the code under this directory is free of non-local [side-effects](https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free). |
+| `app/package.json`              | This file is generated only in applications created using `--strict` mode. This file is not used by package managers. It is used to tell the tools and bundlers whether the code under this directory is free of non-local [side-effects](guide/strict-mode#side-effect). |
 
 ### Application configuration files
 
--- a/aio/content/guide/setup-local.md
+++ b/aio/content/guide/setup-local.md
@@ -25,49 +25,48 @@ To use the Angular framework, you should be familiar with the following:
 
 Knowledge of [TypeScript](https://www.typescriptlang.org/) is helpful, but not required.
 
+To install Angular on your local system, you need the following:
+
 {@a nodejs}
-### Node.js
 
-Make sure your development environment includes `Node.js®` and an npm package manager.
+* **Node.js**
+  
+  Angular requires a [current, active LTS, or maintenance LTS](https://nodejs.org/about/releases) version of Node.js.
 
-Angular requires a [current, active LTS, or maintenance LTS](https://nodejs.org/about/releases/) version of `Node.js`. See the `engines` key for the specific version requirements in our [package.json](https://unpkg.com/@angular/cli/package.json).
+  <div class="alert is-helpful">
 
-* To check your version, run `node -v` in a terminal/console window.
+  For information about specific version requirements, see the `engines` key in the [package.json](https://unpkg.com/@angular/cli/package.json) file.
 
-* To get `Node.js`, go to [nodejs.org](https://nodejs.org "Nodejs.org").
+  </div>
+
+  For more information on installing Node.js, see [nodejs.org](http://nodejs.org "Nodejs.org").
+  If you are unsure what version of Node.js runs on your system, run `node -v` in a terminal window.
 
 {@a npm}
-### npm package manager
 
-Angular, the Angular CLI, and Angular apps depend on features and functionality provided by libraries that are available as [npm packages](https://docs.npmjs.com/getting-started/what-is-npm). To download and install npm packages, you must have an npm package manager.
+* **npm package manager**
 
-This setup guide uses the [npm client](https://docs.npmjs.com/cli/install) command line interface, which is installed with `Node.js` by default.
-
-To check that you have the npm client installed, run `npm -v` in a terminal/console window.
+  Angular, the Angular CLI, and Angular applications depend on [npm packages](https://docs.npmjs.com/getting-started/what-is-npm) for many features and functions.
+  To download and install npm packages, you need an npm package manager.
+  This guide uses the [npm client](https://docs.npmjs.com/cli/install) command line interface, which is installed with `Node.js` by default.
+  To check that you have the npm client installed, run `npm -v` in a terminal window.
 
 
 {@a install-cli}
 
-## Step 1: Install the Angular CLI
+## Install the Angular CLI
 
-You use the Angular CLI
-to create projects, generate application and library code, and perform a variety of ongoing development tasks such as testing, bundling, and deployment.
+You use the Angular CLI to create projects, generate application and library code, and perform a variety of ongoing development tasks such as testing, bundling, and deployment.
 
-Install the Angular CLI globally.
-
-To install the CLI using `npm`, open a terminal/console window and enter the following command:
-
+To install the Angular CLI, open a terminal window and run the following command:
 
 <code-example language="sh" class="code-shell">
   npm install -g @angular/cli
-
 </code-example>
-
-
 
 {@a create-proj}
 
-## Step 2: Create a workspace and initial application
+## Create a workspace and initial application
 
 You develop apps in the context of an Angular [**workspace**](guide/glossary#workspace).
 
@@ -86,16 +85,22 @@ The Angular CLI installs the necessary Angular npm packages and other dependenci
 
 The CLI creates a new workspace and a simple Welcome app, ready to run.
 
+<div class="alert is-helpful">
+
+You also have the option to use Angular's strict mode, which can help you write better, more maintainable code.
+For more information, see [Strict mode](/guide/strict-mode).
+
+</div>
 
 {@a serve}
 
-## Step 3: Run the application
+## Run the application
 
-The Angular CLI includes a server, so that you can easily build and serve your app locally.
+The Angular CLI includes a server, so that you can build and serve your app locally.
 
-1. Go to the workspace folder (`my-app`).
+1. Navigate to the workspace folder, such as `my-app`.
 
-1. Launch the server by using the CLI command `ng serve`, with the `--open` option.
+1. Run the following command:
 
 <code-example language="sh" class="code-shell">
   cd my-app
@@ -108,7 +113,7 @@ and rebuilds the app as you make changes to those files.
 The `--open` (or just `-o`) option automatically opens your browser
 to `http://localhost:4200/`.
 
-You will see:
+If your installation and setup was successful, you should see a page similar to the following.
 
 
 <div class="lightbox">
--- a/aio/content/guide/strict-mode.md
+++ b/aio/content/guide/strict-mode.md
@@ -1,0 +1,46 @@
+# Strict mode
+
+When you create a new workspace or a project you have an option to create them in a strict mode using the `--strict` flag.
+
+Enabling this flag initializes your new workspace or project with a few new settings that improve maintainability, help you catch bugs ahead of time, and allow the CLI to perform advanced optimizations on your application.
+Additionally, applications that use these stricter settings are easier to statically analyze, which can help the `ng update` command refactor code more safely and precisely when you are updating to future versions of Angular.
+
+Specifically, the `strict` flag does the following:
+
+* Enables [`strict` mode in TypeScript](https://www.staging-typescript.org/tsconfig#strict), as well as other strictness flags recommended by the TypeScript team. Specifically, `forceConsistentCasingInFileNames`, `noImplicitReturns`,  `noFallthroughCasesInSwitch`.
+* Turns on strict Angular compiler flags [`strictTemplates`](guide/angular-compiler-options#stricttemplates) and [`strictInjectionParameters`](guide/angular-compiler-options#strictinjectionparameters)
+* [Bundle size budgets](guide/build#configuring-size-budgets) have been reduced by ~75%
+* Turns on [`no-any` tslint rule](https://palantir.github.io/tslint/rules/no-any/) to prevent declarations of type `any`
+* [Marks your application as side-effect free](https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free) to enable more advanced tree-shaking
+
+You can apply these settings at the workspace and project level.
+
+To create a new workspace and application using the strict mode, run the following command:
+
+<code-example language="sh" class="code-shell">
+
+ng new [project-name] --strict
+
+</code-example>
+
+To create a new application in the strict mode within an existing non-strict workspace, run the following command:
+
+<code-example language="sh" class="code-shell">
+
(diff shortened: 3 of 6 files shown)


alan-agius4 (commented):
I think this is not the right place for this piece.

The main reason is that while we do enable stricter TypeScript options when using the `--strict` options, we also enable other stricter options.  Such as bundle budgets, tslint rules and code optimizations rules.

In addition to that, some `strict` settings are workspace-wide like TS options and TSLint options, but some others are at application project level like code optimization rules and bundle budgets. 

Creates an empty strict workspace
```
ng new [project-name] --no-create-application --strict
```

Creates a strict workspace and an application with strict project level settings.
```
ng new [project-name] --strict
```

Creates a new application with strict project level settings in the current workspace.
```
ng generate application [project-name] --strict
```

I tried to look for a section where this is more suitable, the only places I can think of are
https://angular.io/cli/new
https://angular.io/cli/generate#application-command

aikidave:
@IgorMinar Can you share your thoughts on this?

IgorMinar (changes requested):
as @alan-agius4 said the --strict mode is a workspace level setting. It's a super set of `tsc --strict` mode. 

I looked through docs we currently have and realized there isn't a single one best place where these docs should live.

The docs that currently document topics relevant to the strict mode are:
- https://angular.io/cli/new - this doc should absolutely contain a link to another place where this mode is documented at depth (or the docs should live here? seems wrong since this is an api doc).

Other docs are:
- https://angular.io/guide/workspace-config - this one seems the most relevant to this change
- aio/content/guide/typescript-configuration.md - this doc should only contain a reference that the typescript configuration is different based on if the project was created in the strict mode or not.

Misc docs:
- https://angular.io/tutorial/toh-pt0 - this is the only guide that covers project creation, but since it's a tutorial, it seems wrong to discuss the strict mode here.


When it comes to content, I believe that we have good coverage, with the exception of the omission of `<project_root>/src/app/package.json` which enables more aggressive optimizations.

IgorMinar:
I reread @alan-agius4's comment and can confirm that I'm aligned with his suggestions.

IgorMinar:
Oh, one more thing: it would be useful to prefix the description with a preamble that is is our intention to turn this mode on by default in a future major version of Angular

aio/content/guide/typescript-configuration.md
```diff
@@ -59,6 +59,67 @@ The initial `tsconfig.json` for an Angular app typically looks like the followin
 }
 </code-example>
 
+### Strict mode
+
+When you create a new Angular project, you can configure it to automatically use TypeScript's _strict mode_.
```
IgorMinar, line 64: As @alan-agius4 mentioned this description downplays the full scope of the Angular strict mode which is: `tsc --strict` + `ngc --strictTemplates --strictInjectionParams` + `src/app/package.json` + stricter is and css budgets + more, @alan-agius4 can provide the full list in case you don't have it yet.

alan-agius4: @aikidave, ping me if when you need more details.

alan-agius4:
@IgorMinar, I am not sure that https://angular.io/guide/workspace-config is the correct place either. Because this page describes the workspace config file (IE: the `angular.json`).

Under strict mode we change 4 files, `tslint.json`, `angular.json`, `tsconfig.base.json` and `<project_root>/src/app/package.json `

aio/content/guide/typescript-configuration.md
IgorMinar: @alan-agius4 can you please provide a full list of changes that the `--strict` mode causes? I think it's very difficult for a non-implementer to figure out the full diff. Thank you!

aikidave:
After reviewing everyone's comments, I agree that we don't have the best location to talk about this feature. I've reviewed the documentation. In my opinion, we should at least mention how to use strict mode in the Setting up the Local Environment topic. This topic is one of the only places where we talk about running `ng new` to create a new workspace and/or application. I've updated this topic to have a section about using the strict mode.

If we want to provide developers with a clear understanding about what strict mode does, I need to have that information. (I thought I had the complete story, but I'm getting the impression that strict mode does more than my source material suggests.) I don't think we need to explain every file change unless that is the only way to explain the benefits of using strict mode to the community. 

When a user turns on strict mode, what happens? I suggest providing more than just "this flag is set to true." What is that flag for? We can include that information in the `ng new` documentation, if having it in the Typescript Configuration file (which is what we are doing now) is not appropriate.

Thanks!

alan-agius4:
When using strict mode, a user opts in stricter type checking and optimization effective coding rules.

**ng-new schematic**
- We add stricter TypeScript and Angular type-checking options which effect all the projects in a workspace. (These are the settings that you mentioned in this PR).
- Add stricter tslint rules ex: `no-any` which effect all the projects in a workspace.
- When `createApplication` is true, which is the default, the generated root application is also created with strict settings (described below).

**application schematic**
- A `package.json` is added in the `app` folder, to tell the bundlers whether the application is free from side-effect code. When `strict` is `true`. the `sideEffects` will be set `false`.
- Stricter application bundle budgets. 

We should also mentioned that we are targeting to make `strict` mode the default in the future.

aio/content/guide/typescript-configuration.md
```diff
@@ -79,6 +79,67 @@ The initial `tsconfig.base.json` for an Angular workspace typically looks like t
 }
 </code-example>
 
+### Strict mode
```
IgorMinar, line 82: remove most of this and just keep the heading + one paragraph that points people to the new strict mode guide?

aio/content/guide/workspace-config.md
```diff
@@ -41,6 +41,11 @@ When you create a library project with `ng generate library`, the library projec
 
 </div>
 
+## Strict mode
+
+When you create new workspaces and projects, you have the option to use Angular's strict mode, which can help you write better, more maintainable code.
+For more information, see [Employing strict mode when building applications](/guide/strict-mode).
```
IgorMinar, line 47: Update the heading? (if we do end up changing the title of the guide)

aio/content/navigation.json
```diff
@@ -675,6 +675,11 @@
           "url": "guide/browser-support",
           "title": "Browser Support",
           "tooltip": "Browser support and polyfills guide."
+        },
+        {
+          "url": "guide/strict-mode",
+          "title": "Employ strict mode",
```
IgorMinar, line 681: Update the heading? (if we do end up changing the title of the guide)

aio/content/guide/file-structure.md
```diff
@@ -78,6 +78,12 @@ Files at the top level of `src/` support testing and running your application. S
 | `styles.sass`          | Lists CSS files that supply styles for a project. The extension reflects the style preprocessor you have configured for the project. |
 | `test.ts`              | The main entry point for your unit tests, with some Angular-specific configuration. You don't typically need to edit this file. |
 
+<div class="alert is-helpful">
+
+If you create an application using Angular's strict mode, you will also have an additional `package.json` file in the `src/app` directory. For more information, see [Employing strict mode when building applications](/guide/strict-mode).
```
IgorMinar, line 83: Update the heading? (if we do end up changing the title of the guide)

aio/content/guide/setup-local.md
```diff
@@ -25,49 +25,44 @@ To use the Angular framework, you should be familiar with the following:
 
 Knowledge of [TypeScript](https://www.typescriptlang.org/) is helpful, but not required.
 
-{@a nodejs}
```
IgorMinar, line 28: this anchor might be used by various other docs even outside of angular.io. Do we need to remove it? (in general we should try to treat links including anchors as public api and break them rarely and carefully)

aio/content/guide/setup-local.md
```diff
@@ -25,49 +25,44 @@ To use the Angular framework, you should be familiar with the following:
...
-Make sure your development environment includes `Node.js®` and an npm package manager.
+* **Node.js**
+  
+  Angular requires a [current, active LTS, or maintenance LTS](https://nodejs.org/about/releases) version of Node.js.
 
-Angular requires a [current, active LTS, or maintenance LTS](https://nodejs.org/about/releases/) version of `Node.js`. See the `engines` key for the specific version requirements in our [package.json](https://unpkg.com/@angular/cli/package.json).
+  <div class="alert is-helpful">
 
-* To check your version, run `node -v` in a terminal/console window.
+  For information about specific version requierments, see the `engines` key in the [package.json](https://unpkg.com/@angular/cli/package.json) file.
```
IgorMinar, line 36: I believe that this one **is** *just a typo*.
```suggestion
  For information about specific version requirements, see the `engines` key in the [package.json](https://unpkg.com/@angular/cli/package.json) file.
```

aio/content/guide/setup-local.md
```diff
@@ -86,16 +81,40 @@ The Angular CLI installs the necessary Angular npm packages and other dependenci
 
 The CLI creates a new workspace and a simple Welcome app, ready to run.
 
+### Create a workspace using the `--strict` flag
```
IgorMinar: I think we should remove this section and replace it with a "did you know?" highlight promoting `--strict` mode as a good practice.

aio/content/guide/setup-local.md
```diff
@@ -25,49 +25,44 @@ To use the Angular framework, you should be familiar with the following:
...
-Angular requires a [current, active LTS, or maintenance LTS](https://nodejs.org/about/releases/) version of `Node.js`. See the `engines` key for the specific version requirements in our [package.json](https://unpkg.com/@angular/cli/package.json).
+  <div class="alert is-helpful">
 
-* To check your version, run `node -v` in a terminal/console window.
+  For information about specific version requierments, see the `engines` key in the [package.json](https://unpkg.com/@angular/cli/package.json) file.
 
-* To get `Node.js`, go to [nodejs.org](https://nodejs.org "Nodejs.org").
+  </div>
 
-{@a npm}
```
IgorMinar, line 39: please don't remove this one.

IgorMinar (approved):
This LGTM once my suggestions are considered. Thanks @aikidave for putting this together!

aio/content/guide/strict-mode.md
```diff
@@ -0,0 +1,52 @@
+# Employing strict mode when building applications
+
+When you create a new workspace or application, you have the option to use a more strict project setup when you create a new workspace with `ng new`.
+
+```
+ng new --strict
+```
+
+Enabling this flag initializes your new project with a few new settings that improve maintainability, help you catch bugs ahead of time, and allow the CLI to perform advanced optimizations on your app.
```
alan-agius4, line 9: Can we avoid abbreviates. We should use application instead of app

aio/content/guide/strict-mode.md
```diff
@@ -0,0 +1,52 @@
...
+
+To create a new workspace and application using the strict mode, run the following command:
+
+<code-example language="sh" class="code-shell">
+
+ng new [project-name] --strict
+
+</code-example>
+
+To use strict mode for a new project in an existing application, run the following command:
```
alan-agius4, line 30: To use strict mode for a new application project in an existing workspace

aio/content/guide/strict-mode.md
```diff
@@ -0,0 +1,46 @@
...
+
+To create a new workspace and application using the strict mode, run the following command:
+
+<code-example language="sh" class="code-shell">
+
+ng new [project-name] --strict
+
+</code-example>
+
+To use strict mode for a new application project in an existing workspace, run the following command:
```
IgorMinar, line 26:
```suggestion
To create a new project in the strict mode within an existing non-strict workspace workspace, run the following command:
```

aio/content/guide/strict-mode.md
```diff
@@ -0,0 +1,46 @@
...
+
+<code-example language="sh" class="code-shell">
+
+ng generate application [project-name] --strict
+
+</code-example>
+
+{@a side-effect}
+
+### Determine if an application has non-local side effects
```
IgorMinar, line 36:
```suggestion
### Configuring if an application has non-local side effects
```

IgorMinar (approved):

aio/content/guide/strict-mode.md
```diff
@@ -0,0 +1,46 @@
...
+
+To create a new workspace and application using the strict mode, run the following command:
+
+<code-example language="sh" class="code-shell">
+
+ng new [project-name] --strict
+
+</code-example>
+
+To create a new project in the strict mode within an existing non-strict workspace, run the following command:
```
alan-agius4, line 26:
```suggestion
To create a new application in the strict mode within an existing non-strict workspace, run the following command:
```

`project` is a generic term which can mean a number of project types. Ex: Application and Library

aio/content/guide/strict-mode.md
```diff
@@ -0,0 +1,46 @@
+# Strict mode
+
+When you create a new workspace or a project you have an option to create them in a strict mode using the `--strict` flag.
+
+Enabling this flag initializes your new project with a few new settings that improve maintainability, help you catch bugs ahead of time, and allow the CLI to perform advanced optimizations on your app.
```
alan-agius4, line 5:
```suggestion
Enabling this flag initializes your new workspace with a few new settings that improve maintainability, help you catch bugs ahead of time, and allow the CLI to perform advanced optimizations on your application.
```

aio/content/guide/strict-mode.md
```diff
@@ -0,0 +1,46 @@
...
+
+Specifically, the `strict` flag does the following:
+
+* Enables [`strict` mode in TypeScript](https://www.staging-typescript.org/tsconfig#strict), as well as other strictness flags recommended by the TypeScript team. Specifically, `forceConsistentCasingInFileNames`, `noImplicitReturns`,  `noFallthroughCasesInSwitch`.
+* Turns on strict Angular compiler flags [`strictTemplates`](guide/angular-compiler-options#stricttemplates) and [`strictInjectionParameters`](guide/angular-compiler-options#strictinjectionparameters)
+* [Bundle size budgets](guide/build#configuring-size-budgets) have been reduced by ~75%
+* Turns on [`no-any` tslint rule](https://palantir.github.io/tslint/rules/no-any/) to prevent declarations of type `any`
+* [Marks your application as side-effect free](https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free) to enable more advanced tree-shaking
+
+You can apply these settings at the workspace and project level.
```
alan-agius4, line 16: Umm.. not quite right, there are strict mode settings which are workspace wide and strict mode settings which are at project level. Some settings are workspace level and some are at application project level.

I think it would also be good to clarify this in the docs. As using `ng generate application [project-name] --strict` in a non strict workspace is different from `ng generate application [project-name] --strict` in a strict workspace.

See 
https://github.com/angular/angular/pull/37486#issuecomment-644600427 for more info.

aio/content/guide/file-structure.md
```diff
@@ -78,6 +78,12 @@ Files at the top level of `src/` support testing and running your application. S
...
 
+<div class="alert is-helpful">
+
+If you create an application using Angular's strict mode, you will also have an additional `package.json` file in the `src/app` directory. For more information, see [Strict mode](/guide/strict-mode).
+
+</div>
+
 {@a app-src}
 
 Inside the `src/` folder, the `app/` folder contains your project's logic and data.
```
alan-agius4, line 89: Down below there is a reference for `src/app/package.json`, can we add a link to the section `Non-local side effects in applications` you created in this PR?

aio/content/guide/strict-mode.md
IgorMinar: @alan-agius4 I've just tested `ng generate application [project-name] --strict` in a non-strict workspace and it doesn't seem to generate a strict project.

In general the diff between strict application in a non-strict workspace and an application within a strict workspace should be minimal or none. Anything else would likely be perceived as bug. Can you please look into why this doesn't seem to work as expected? thanks

IgorMinar: For the time being, let's resolve this conversation and keep the documentation as is until we have more information about why this must work differently.

alan-agius4: ## The Why
It works differently because `strict` on `ng new` is a combination of 2 schematics. the workspace and the application schematics. Both of which set different strict settings. 

The `ng new` schematic is just a "wrapper" which combines both of these into a single schematic which generates a workspace and an application. This is equivalent in running `ng g workspace myapp --strict/--no-strict` and `ng g app myapp --strict/--no-strict`. 
 
But when you use `ng g app --strict` you will only get the strict settings which are available/set only at application level, such as bundle budgets, side effects package.json etc.., you will not get the workspace strict settings such as stricter TSC compiler options, stricter tslint rules etc...

## The What
### ng-new/workspace schematic

- We add stricter TypeScript and Angular type-checking options which effect all the projects in a workspace. (These are the settings that you mentioned in this PR).
- Add stricter tslint rules ex: no-any which effect all the projects in a workspace.
- When createApplication is true, which is the default, the generated root application is also created with strict settings (described below).

### application schematic

- A package.json is added in the app folder, to tell the bundlers whether the application is free from side-effect code. When strict is true. the sideEffects will be set false.
- Stricter application bundle budgets.

## Epilogue
Thinking about this, maybe we should hide the `strict` option from the application schematic, and only generate strict mode applications in strict workspaces.

**Note**: future applications are always generated in strict mode in strict workspaces even when not using `--strict`.

PR #37489, opened 2020-06-08
fix(dev-infra): rebase pr script and token sanitization not working

devversion:
* See individual commits.

### DIFF
--- a/dev-infra/utils/git.ts
+++ b/dev-infra/utils/git.ts
@@ -142,7 +142,7 @@ export class GitClient {
 
   /** Whether the repo has any local changes. */
   hasLocalChanges(): boolean {
-    return !!this.runGraceful(['git', 'status', '--porcelain']).stdout.trim();
+    return this.runGraceful(['diff-index', '--quiet', 'HEAD']).status !== 0;
   }
 
   /** Sanitizes a given message by omitting the provided Github token if present. */
--- a/dev-infra/utils/git.ts
+++ b/dev-infra/utils/git.ts
@@ -73,7 +73,7 @@ export class GitClient {
     // also create a regular expression that can be used for sanitizing Git command output
     // so that it does not print the token accidentally.
     if (_githubToken != null) {
-      this._githubTokenRegex = new RegExp(_githubToken, 'g')
+      this._githubTokenRegex = new RegExp(_githubToken, 'g');
     }
 
     this.api = new Octokit({auth: _githubToken});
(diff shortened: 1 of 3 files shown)


dev-infra/utils/git.ts
```diff
@@ -61,11 +61,21 @@ export class GitClient {
...
+   */
+  private _githubTokenRegex: RegExp|null = null;
 
   constructor(
-      private _githubToken = '', private _config: Pick<NgDevConfig, 'github'> = getConfig()) {
+      private _githubToken?: string, private _config: Pick<NgDevConfig, 'github'> = getConfig()) {
+    // If a token has been specified, pass it to the Octokit API and also create
+    // a regular expression that can be used for sanitizing Git command output
+    // so that it does not print the token accidentally.
+    if (this._githubToken != null) {
```
josephperrott, line 75: Should this check be `typeof this._githubToken === 'string'`?

dev-infra/utils/git.ts
```diff
@@ -61,11 +61,21 @@ export class GitClient {
...
-  private _tokenRegex = new RegExp(this._githubToken, 'g');
+  /**
+   * Regular expression that matches the provided Github token. Used for
+   * sanitizing the token from Git child process output.
+   */
+  private _githubTokenRegex: RegExp|null = null;
 
   constructor(
-      private _githubToken = '', private _config: Pick<NgDevConfig, 'github'> = getConfig()) {
+      private _githubToken?: string, private _config: Pick<NgDevConfig, 'github'> = getConfig()) {
```
josephperrott, line 71: Does it make sense to _not_ save `_githubToken` as a member?  Since we shouldn't have to be using it anywhere else after initialization.

josephperrott (approved):
LGTM

dev-infra/utils/git.ts
devversion: We only allow `string` or `undefined`, so I think the typeof check wouldn't be needed. The goal is to not proceed if the token is empty or if the parameter has not been passed.

Happy to clarify this in a comment if you think it's helpful.

josephperrott: I suppose thats true, I was thinking of what would be loaded from the config file, but the token is loaded by either an environment variable or from a cli flag, so its always going to be a string or undefined.

dev-infra/utils/git.ts
devversion: We need it above when constructing the Git URL.

dev-infra/utils/git.ts
devversion: Updated the comment slightly so that this is more clear now. Let me know if that works!

josephperrott: LGTM

PR #37490, opened 2020-06-08
feat(router): allow redirect from guards by returning URL string

dimakuba:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Guards allow returning either boolean or UrlTree.

## What is the new behavior?
Allow returning either boolean, UrlTree or URL string.

- [ ] Yes
- [x] No

### DIFF
--- a/goldens/public-api/router/router.d.ts
+++ b/goldens/public-api/router/router.d.ts
@@ -52,19 +52,19 @@ export declare class ActivationStart {
 }
 
 export declare interface CanActivate {
-    canActivate(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Observable<boolean | UrlTree> | Promise<boolean | UrlTree> | boolean | UrlTree;
+    canActivate(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Observable<boolean | UrlTree | string> | Promise<boolean | UrlTree | string> | boolean | UrlTree | string;
 }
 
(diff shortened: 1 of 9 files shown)


atscott:
Hi @dimakuba, thanks for the suggestion. Can you describe what prompted you to make this change? Maybe there's something I'm missing or we can improve the documentation to ensure developers are aware of the available options and how to most easily use them. After a quick glance, I'm trying to determine why simply returning a `UrlTree` by using `Router#parseUrl` (https://angular.io/api/router/Router#parseurl) wouldn't cover this use-case. 

We want to be sure that we are balancing code complexity, bundle size, and API surface with convenience. More options can be more confusing and also requires additional maintenance. It appears that a straightforward API already exists here and that this change may be unnecessary.

dmkuba:
Hi @atscott, it makes so simple to redirect in a declarative manner for non-complex urls, such as `/login`

atscott:
> Hi @atscott, it makes so simple to redirect in a declarative manner for non-complex urls, such as `/login`

In cases like this where a relatively straightforward API already exists (`router.parseUrl('/login')` or `router.createUrlTree(['/login'])`), the balance shifts in favor of avoiding unnecessary bloat to the code size and API surface. Consequently, we’ve decided not to add these API changes and will be closing the issue. But keep the suggestions coming.

PR #36144, opened 2020-03-19
Add lightweight token doc to library section

jbogarthyde:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [ ] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [x] Documentation content changes
- [ ] angular.io application / infrastructure changes
- [ ] Other... Please describe:

## What is the current behavior?
Misko wrote up a design doc for the new lightweight-token design pattern that is recommended for library developers.

Issue Number: N/A

## What is the new behavior?
 The design doc content is transformed to  user doc aimed at library developers and added to AIO as a new page in the Libraries section, with a link from 
https://angular.io/guide/creating-libraries#refactoring-parts-of-an-app-into-a-library

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

PR #37506, opened 2020-06-09
feat(core): support injection token as predicate in queries

### LINKED ISSUE #21152: BUG: InjectionToken is not assignable to ViewChildren or ContentChildren
## I'm submitting a...

<pre><code>
[X] Bug report  
[ ] Feature request
</code></pre>

## Current behavior
If i have `InjectionToken` i cant assign it to `ViewChildren` or `ContentChildren` as mentioned in https://github.com/angular/angular/issues/12758 Maybe due to changes of Typescript.

## Expected behavior
This code is working if you type it to `<any>` but without it cant be compiled using typescript.

This will not compile :).

``` typescript
export const SomeClassToken = new InjectionToken<SomeClass>('SomeClassToken');

export class TestClass
{
    @ViewChildren(SomeClassToken)
    public children: QueryList<SomeClass>;
}
```

It should compile, because it works if typed to `<any>`.

## Environment

<pre><code>
Angular version: 5.1.2

Browser:
During compilation
- [ ] Chrome (desktop) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX
 
For Tooling issues:
- Typescript version: 2.5.3
- Node version: 8.1.2  
- Platform:  Windows 

Others:

</code></pre>

devversion:
Currently Angular internally already handles `InjectionToken` as
predicates for queries. This commit exposes this as public API as
developers already rely on this functionality but currently use
workarounds to satisfy the type constraints (e.g. `as any`).

We intend to make this public as it's low-effort to support, and
it's a significant key part for the use of light-weight tokens as
described in the upcoming guide: #36144.

In concrete, applications might use injection tokens over classes
for both optional DI and queries, because otherwise such references
cause classes to be always retained. This was also an issue in View
Engine, but now with Ivy, this pattern became worse, as factories are
directly attached to retained classes (ultimately ending up in the
production bundle, while being unused).

More details in the light-weight token guide and in: angular/angular-cli#16866.

Closes #21152. Related to #36144.

### DIFF
--- a/goldens/public-api/core/core.d.ts
+++ b/goldens/public-api/core/core.d.ts
@@ -163,11 +163,11 @@ export declare interface ConstructorSansProvider {
 export declare type ContentChild = Query;
 
 export declare interface ContentChildDecorator {
-    (selector: Type<any> | Function | string, opts?: {
+    (selector: Type<any> | InjectionToken<unknown> | Function | string, opts?: {
         read?: any;
         static?: boolean;
     }): any;
-    new (selector: Type<any> | Function | string, opts?: {
+    new (selector: Type<any> | InjectionToken<unknown> | Function | string, opts?: {
         read?: any;
         static?: boolean;
     }): ContentChild;
@@ -176,11 +176,11 @@ export declare interface ContentChildDecorator {
 export declare type ContentChildren = Query;
 
 export declare interface ContentChildrenDecorator {
-    (selector: Type<any> | Function | string, opts?: {
+    (selector: Type<any> | InjectionToken<unknown> | Function | string, opts?: {
         descendants?: boolean;
         read?: any;
     }): any;
-    new (selector: Type<any> | Function | string, opts?: {
+    new (selector: Type<any> | InjectionToken<unknown> | Function | string, opts?: {
         descendants?: boolean;
         read?: any;
     }): Query;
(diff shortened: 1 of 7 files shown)


jelbourn (approved):
LGTM

goldens/public-api/core/core.d.ts
```diff
@@ -163,11 +163,11 @@ export declare interface ConstructorSansProvider {
 export declare type ContentChild = Query;
 
 export declare interface ContentChildDecorator {
-    (selector: Type<any> | Function | string, opts?: {
+    (selector: Type<any> | InjectionToken<any> | Function | string, opts?: {
```
IgorMinar, line 166: Any reason not to use `unknown` instead of `any`? Historically we used to use `any` in our public api surface whenever the type was unknown, but ever since TypeScript introduced the `unknown` type we started to to use that one instead. Sometimes `unknown` is not suitable, but in this case it might be. Can you please try it?


```suggestion
    (selector: Type<any> | InjectionToken<unknown> | Function | string, opts?: {
```

IgorMinar (changes requested):
otherwise LGTM

Reviewed-for: public-api

goldens/public-api/core/core.d.ts
devversion: good idea! no strong reason. I just did it that way to align with existing generics for query decorators.  Updated it to use `unknown`!

I assume we can clean up the other generics too. I can try sending a follow-up, or incorporate this here, if possible.

IgorMinar (approved):
LGTM now! Thanks

(changing existing apis to use `unknown` could be a breaking change, so we can do so only in a major version and only with proper discussion/review. So I suggest we punt that discussion until master is open for v11)

[presubmit](http://test/OCL:315949641:BASE:315949698:1591902482391:17d3956c)

PR #37509, opened 2020-06-09
docs(dev-infra): add comment about what the requiredBaseCommit is

atscott:
Add a comment to describe what the commit was for the given SHA so that we don't have to look it up.


josephperrott (approved):
LGTM

PR #37514, opened 2020-06-09
perf(core): avoid pulling in jit-specific code in aot bundles (#37372)

crisbeto:
**Note:** this is the patch-only version of #37372.

In #29083 a call to `getCompilerFacade` was added to `ApplicationRef` which pulls in a bit of JIT-specific code. Since the code path that calls the function can't be hit for an AOT-compiled app, these changes add an `ngJitMode` guard which will allow for dead code elimination to drop it completely. Testing it out against a new CLI project showed a difference of ~1.2kb.

PR Close #37372

### DIFF
--- a/goldens/size-tracking/aio-payloads.json
+++ b/goldens/size-tracking/aio-payloads.json
@@ -12,7 +12,7 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 2987,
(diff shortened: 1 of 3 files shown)


atscott (approved):

IgorMinar (approved):
LGTM! Thank you!

PR #37521, opened 2020-06-10
Docs 2 small changes

alan-agius4:
See each commit.

### DIFF
--- a/aio/content/guide/file-structure.md
+++ b/aio/content/guide/file-structure.md
@@ -90,7 +90,7 @@ Angular components, templates, and styles go here.
 | `app/app.component.css`     | Defines the base CSS stylesheet for the root `AppComponent`. |
 | `app/app.component.spec.ts` | Defines a unit test for the root `AppComponent`. |
 | `app/app.module.ts`         | Defines the root module, named `AppModule`, that tells Angular how to assemble the application. Initially declares only the `AppComponent`. As you add more components to the app, they must be declared here. |
-| `package.json`              | This file is not used by package managers. It is used to tell the tools and bundlers whether the code under this directory is free of non-local [side-effects](https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free). |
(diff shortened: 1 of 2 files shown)


alan-agius4:
FYI: @kyliau

kyliau (changes requested):

kyliau (approved):

aio/content/guide/file-structure.md
```diff
@@ -89,6 +89,7 @@ Angular components, templates, and styles go here.
 | `app/app.component.css`     | Defines the base CSS stylesheet for the root `AppComponent`. |
 | `app/app.component.spec.ts` | Defines a unit test for the root `AppComponent`. |
 | `app/app.module.ts`         | Defines the root module, named `AppModule`, that tells Angular how to assemble the application. Initially declares only the `AppComponent`. As you add more components to the app, they must be declared here. |
+| `package.json`              | This file is not used by package managers. It is used to tell the tools and bundlers whether the code under this directory is free of non-local [side-effects](https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free). |
```
IgorMinar, line 92:
```suggestion
| `app/package.json`              | This file is generated only in projects created in the `ng new --strict` mode. This file is not used by package managers. It is used to tell the tools and bundlers whether the code under this directory is free of non-local [side-effects](https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free). |
```

IgorMinar (approved):
lgtm otherwise

aio/content/guide/file-structure.md
alan-agius4: Actually that is not correct, the file is always created the value is changed based on if the project is in strict mode or not.

alan-agius4: Let’s me know if we should update the schematic to only generate this file when using strict mode.

alan-agius4: Following a convo with @kyliau and @mgechev we decided to drop this file when not in strict mode.

alan-agius4: Updated the description a bit to: `This file is generated only in applications created using `--strict` mode.`. because this also effects `ng g application`.

PR #37523, opened 2020-06-10
perf(router): apply prioritizedGuardValue operator to optimize CanLoad guards

dimakuba:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

CanLoad guards are processed in the order they appear in the config.

## What is the new behavior?
CanLoad guards are processed in asynchronous manner with the following rules:
* If all guards return `true`, operator returns `true`;
* `false` and `UrlTree` values wait for higher priority guards to resolve;
* Highest priority `false` or `UrlTree` value will be returned.

### DIFF
--- a/packages/router/test/integration.spec.ts
+++ b/packages/router/test/integration.spec.ts
@@ -14,7 +14,7 @@ import {By} from '@angular/platform-browser/src/dom/debug/by';
 import {expect} from '@angular/platform-browser/testing/src/matchers';
 import {ActivatedRoute, ActivatedRouteSnapshot, ActivationEnd, ActivationStart, CanActivate, CanDeactivate, ChildActivationEnd, ChildActivationStart, DefaultUrlSerializer, DetachedRouteHandle, Event, GuardsCheckEnd, GuardsCheckStart, Navigation, NavigationCancel, NavigationEnd, NavigationError, NavigationStart, ParamMap, Params, PreloadAllModules, PreloadingStrategy, PRIMARY_OUTLET, Resolve, ResolveEnd, ResolveStart, RouteConfigLoadEnd, RouteConfigLoadStart, Router, RouteReuseStrategy, RouterEvent, RouterModule, RouterPreloader, RouterStateSnapshot, RoutesRecognized, RunGuardsAndResolvers, UrlHandlingStrategy, UrlSegmentGroup, UrlSerializer, UrlTree} from '@angular/router';
 import {EMPTY, Observable, Observer, of, Subscription} from 'rxjs';
-import {filter, first, map, tap} from 'rxjs/operators';
+import {delay, filter, first, map, mapTo, tap} from 'rxjs/operators';
 
 import {forEach} from '../src/utils/collection';
 import {RouterTestingModule, SpyNgModuleFactoryLoader} from '../testing';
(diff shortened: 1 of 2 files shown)


atscott (changes requested):
Hi @dimakuba, could you please add at least one test for this?

dmkuba:
Hi @atscott, i think [this](https://github.com/angular/angular/blob/master/packages/router/test/operators/prioritized_guard_value.spec.ts) test suite covers all scenarios of `prioritizedGuardValue` operator

atscott:
> Hi @atscott, i think [this](https://github.com/angular/angular/blob/master/packages/router/test/operators/prioritized_guard_value.spec.ts) test suite covers all scenarios of `prioritizedGuardValue` operator

It does test the `prioritizedGuardValue` but this PR changes the expectations for how `canLoad` guards are executed. Your new test(s) likely should go in [integration.spec.ts](https://github.com/angular/angular/blob/master/packages/router/test/integration.spec.ts) and should verify that the new expectations for how the `canLoad` guards execute

dimakuba:
Hi @atscott, i've added integration tests.

packages/router/test/integration.spec.ts
```diff
@@ -3946,6 +3939,115 @@ describe('Integration', () => {
              })));
     });
 
+    describe('should run CanLoad guards concurrently', () => {
```
atscott, line 3942: All of these tests still pass if I revert the changes to `apply_redirects.ts`. Is the change truly necessary? Part of the reason I wanted to have these tests written is that I couldn't really determine how `prioritizedGuardValue` operates differently from `concatAll`. Since `concatAll` emits values in order, wouldn't that already satisfy the priority constraint? I honestly haven't really looked much at the implementation of `prioritizedGuardValue` so I'm happy to be proven wrong about this. Maybe it's doing something else.

dimakuba: > I honestly haven't really looked much at the implementation of prioritizedGuardValue so I'm happy to be proven wrong about this. Maybe it's doing something else.

`prioritizedGuardValue` uses `combineLatest` which in order subscribes to each Observable immediately (not waiting when previous one completes that `concatAll` do). So it makes some advantages in order to run them concurrently. Respectively, a time to resolve all guards will be reduced.

> All of these tests still pass if I revert the changes to `apply_redirects.ts`

Ah this is because [delayPromise](https://github.com/angular/angular/blob/4766f268782bb8af1b0cf394b6edd8d1754f3386/packages/router/test/integration.spec.ts#L5704-L5709) is not deferred and starts execution at the point of computing guards. Would be better to wrap the logic of computing guards with `defer`.

```typescript
const canLoadObservables = canLoad.map((injectionToken: any) => {
  return defer(() => {
    const guard = moduleInjector.get(injectionToken);
    let guardVal;
    if (isCanLoad(guard)) {
      guardVal = guard.canLoad(route, segments);
    } else if (isFunction<CanLoadFn>(guard)) {
      guardVal = guard(route, segments);
    } else {
      throw new Error('Invalid CanLoad guard');
    }
    return wrapIntoObservable(guardVal);
  });
});
```

instead of:

https://github.com/angular/angular/blob/567e88579d059e97f82815e8006aa99f25df8694/packages/router/src/apply_redirects.ts#L325-L336

atscott: > prioritizedGuardValue uses combineLatest which in order subscribes to each Observable immediately (not waiting when previous one completes that concatAll do). 

👍 

> Ah this is because delayPromise is not deferred and starts execution at the point of computing guards. Would be better to wrap the logic of computing guards with `defer`.

👍  I see that's what `check_guards` does as well. I think also changing the test to use a `delayObservable` helper would work too but either is fine with me.
```
function delayObservable(delayMs: number): Observable<boolean> {
  return of(delayMs).pipe(delay(delayMs), mapTo(true));
}
```

Could you make one of those changes so we're verifying the new behavior with `prioritizedGuardValue`?

dimakuba: thanks for suggestion, but i did [try](https://github.com/angular/angular/pull/37523/commits/06737d2325093c247426d3d15bc9e6d908453943#diff-6a4970197ae55b072e7e9ab05c02b26fR3963) to write this test suite with observables and got in [failure](https://app.circleci.com/pipelines/github/angular/angular/17182/workflows/df199617-fd5a-4d6b-b06b-3bcf816d5a54/jobs/741080/steps) somehow when `--config=ivy` was checking

atscott: > thanks for suggestion, but i did [try](https://github.com/angular/angular/pull/37523/commits/06737d2325093c247426d3d15bc9e6d908453943#diff-6a4970197ae55b072e7e9ab05c02b26fR3963) to write this test suite with observables and got in [failure](https://app.circleci.com/pipelines/github/angular/angular/17182/workflows/df199617-fd5a-4d6b-b06b-3bcf816d5a54/jobs/741080/steps) somehow when `--config=ivy` was checking

😬 I think this might actually be a bug in the Ivy TestBed. The tests pass if each is run individually with `fit(...)`. Using `defer` like you noted in the `canLoadObservables` should be fine to unblock progress here.

atscott:
[Google internal test results](http://test/OCL:317207313:BASE:317207421:1592523734175:942e290e)

atscott:
@dmkuba All the tests are green! One final thing before marking this as merge-ready: Could you squash your commits into one and amend the message to include your comments about what `prioritizedGuardValue` does compared to the current behavior:

>`prioritizedGuardValue` uses `combineLatest` which in order subscribes to each Observable immediately (not waiting when previous one completes that `concatAll` do). So it makes some advantages in order to run them concurrently. Respectively, a time to resolve all guards will be reduced.

dmkuba:
> @dmkuba All the tests are green! One final thing before marking this as merge-ready: Could you squash your commits into one and amend the message to include your comments about what `prioritizedGuardValue` does compared to the current behavior:
> 
> > `prioritizedGuardValue` uses `combineLatest` which in order subscribes to each Observable immediately (not waiting when previous one completes that `concatAll` do). So it makes some advantages in order to run them concurrently. Respectively, a time to resolve all guards will be reduced.

done

JiaLiPassion:
The tests failed because an issue of `fakeAsync` integration with `rxjs`, I will create an issue and fix it asap.

dmkuba:
> The tests failed because an issue of `fakeAsync` integration with `rxjs`, I will create an issue and fix it asap.

thanks

atscott:
@dmkuba since Jia was able to quickly find the fix for `fakeAsync`, let's wait for that to merge so we can have the proper tests here (ones that would fail without the change).

atscott:
@dmkuba - Jia's fixes were just merged. Could you update the tests to use the `delay` again? Sorry for the back and forth changes here. I really do think the `delay` tests are what we want since they more clearly demonstrate how this changes the current behavior (i.e. the tests would fail without the change). I think that you can also remove the `defer` call from the `canLoad.map` since it was only added to make the tests work as expected.

dimakuba:
> @dmkuba - Jia's fixes were just merged. Could you update the tests to use the `delay` again? Sorry for the back and forth changes here. I really do think the `delay` tests are what we want since they more clearly demonstrate how this changes the current behavior (i.e. the tests would fail without the change). I think that you can also remove the `defer` call from the `canLoad.map` since it was only added to make the tests work as expected.

Hi @atscott, `defer` makes sure guards call will be controlled by the `prioritizedGuardValue` operator that brings more stability and control (by using both: rxjs & promises), should i really remove `defer`?

dimakuba:
tests were updated & passed by using `delay` again.

atscott:
@dimakuba The reason I feel that `defer` isn't needed is that we still get the correct behavior without it. I know there truly is a _slight_ difference in how it actually works with and without `defer` but I think the end result is the same: we execute all the `canLoad` guards at once rather than waiting for each one to finish in-order before starting the next. 

Does that seem fair? `check_guards.ts` is the only other place that uses `defer` so maybe we can remove it there as well if you agree (in a different PR) and save a few bytes by not including that operator in the bundle.

alfaproject:
@atscott I think the issue is if a `canLoad` guard returns a promise. In that case, it will be executed eagerly outside the observable chain. Not sure if a test exists for that scenario or if it's an issue at all but I think that's what @dimakuba is referring to.

dimakuba:
The main purpose of `prioritizedGuardValue` operator is to run `canLoad` guards concurrently (doesn’t matter who starts first), so no need to use `defer` in this case. In a case if `canLoad` guards return promise, emitted values are managed by the  `prioritizedGuardValue` operator, so hight priority guards will be handled properly.

atscott (approved):
👍 Thanks @dimakuba! I just have to run a subset of the Google tests as a final sanity check and then we should be able to get this merged some time this week

[presubmit](https://test.corp.google.com/ui#cl=317207313) (only unrelated failures)

PR #37530, opened 2020-06-10
docs(core): fix typo "filter" should be "first"

move-zig:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/application_ref.ts
+++ b/packages/core/src/application_ref.ts
@@ -481,7 +481,7 @@ function optionsReducer<T extends Object>(dst: any, objs: T|T[]): T {
  * ```
  * constructor(appRef: ApplicationRef) {
  *   appRef.isStable.pipe(
- *      filter(stable => stable)
+ *      first(stable => stable)
  *   ).subscribe(() => console.log('App is stable now');
  *   interval(1000).subscribe(counter => console.log(counter));
  * }


destus90:
The existed code looks correct to me. How is it incorrect for you?

move-zig:
It seems like the point of the examples is to illustrate how the application will never be stable if you start a recurrent asynchronous task, such as `interval` when the application starts.

We have the "bad" code

 ```
constructor(appRef: ApplicationRef) {
  appRef.isStable.pipe(
    filter(stable => stable)
  ).subscribe(() => console.log('App is stable now');
  interval(1000).subscribe(counter => console.log(counter));
}
```

And then the "good" code:

```
 constructor(appRef: ApplicationRef) {
   appRef.isStable.pipe(
     first(stable => stable),
     tap(stable => console.log('App is stable now')),
     switchMap(() => interval(1000))
   ).subscribe(counter => console.log(counter));
 }
```

The first example uses `filter` but the second example uses `first`. If I'm not mistaken, that is not the salient difference. I think both examples should use the same operator.

petebacondarwin (changes requested):

I agree that this is not a particularly useful change.

The point of the "bad" example is to show that even if you continue to subscribe to all emissions of the `isStable` observable you will never get a value of `true`. Changing it to `first` is not correct because it precludes the chance of this observable emitting `false` followed by `true`, which would demonstrate that the app can finally stabliise despite the `interval`.

One could argue that the change to `first` in the "good" example is unnecessary, since it is using `switchMap()` to trigger the `interval` observable, so although you may receive multiple `true` values from the `isStable` observable, the current `interval` will be stopped before the new one is started, resulting in only one `interval` observable being active at any one time. Using `first()` is an optimisation to avoid having to stop and start the `interval` observables every time a `true` is emitted.

So I don't think that this PR is right. If anything it would be more helpful to explain the benefit that is provided by using `first()` but I would say that this would just add noise to the key point of this section of the doc.

Either this PR should be updated to make a very small explanation of the `first()` optimisation or it should be closed.

move-zig:
@petebacondarwin I fixated on the difference in the two operators at first and thought it might have been a typo. But now it's clear that they are intentional. I agree with closing

petebacondarwin:
Thanks

PR #37533, opened 2020-06-11
docs: add redirects for cli generated config files

alan-agius4:
With this change we add redirects for config files generated by the Angular CLI. These links form part of a comment in the generated files, thus it is important that they valid for the many years to come.

Blocking: https://github.com/angular/angular-cli/pull/17906


IgorMinar (approved):
LGTM. thanks!

PR #37539, opened 2020-06-11
feat(platform-server): add option for absolute URL HTTP support

CaerusKaru:
In version 10.0.0-next.8, we introduced absolute URL support for
server-based HTTP requests, so long as the fully-resolved URL was
provided in the initial config. However, doing so represents a
breaking change for users who already have their own interceptors
to model this functionality, since our logic executes before all
interceptors fire on a request.

Therefore, we introduce a flag to make this change consistent with
v9 behavior, allowing users to opt in to this new behavior. This
commit also fixes two issues with the previous implementation:
1. if the server was initiated with a relative URL, the absolute
URL construction would fail because needed components were empty
2. if the user's absolute URL was on a port, the port would not
be included

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/packages/platform-server/src/tokens.ts
+++ b/packages/platform-server/src/tokens.ts
@@ -14,8 +14,25 @@ import {InjectionToken} from '@angular/core';
  * @publicApi
  */
 export interface PlatformConfig {
+  /**
+   * The initial DOM to use to bootstrap the server application.
+   * @default create a new DOM using Domino
+   */
   document?: string;
+  /**
+   * The URL for the current application state. This is
+   * used for initializing the platform's location and
+   * for setting absolute URL resolution for HTTP requests.
+   * @default none
+   */
   url?: string;
+  /**
+   * Whether to append the absolute URL to any relative HTTP
+   * requests. If set to true, this logic executes prior to
+   * any HTTP interceptors that may run later on in the request.
+   * @default false
+   */
+  useAbsoluteUrl?: boolean;
 }
 
 /**
--- a/packages/platform-server/test/integration_spec.ts
+++ b/packages/platform-server/test/integration_spec.ts
@@ -793,104 +793,149 @@ describe('platform-server integration', () => {
          });
        }));
 
-    it('can make relative HttpClient requests', async () => {
-      const platform = platformDynamicServer([
-        {provide: INITIAL_CONFIG, useValue: {document: '<app></app>', url: 'http://localhost'}}
-      ]);
-      const ref = await platform.bootstrapModule(HttpClientExampleModule);
-      const mock = ref.injector.get(HttpTestingController) as HttpTestingController;
-      const http = ref.injector.get(HttpClient);
-      ref.injector.get(NgZone).run(() => {
-        http.get<string>('/testing').subscribe((body: string) => {
-          NgZone.assertInAngularZone();
-          expect(body).toEqual('success!');
+    describe('relative requests', () => {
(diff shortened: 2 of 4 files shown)


packages/platform-server/src/tokens.ts
```diff
@@ -16,6 +16,7 @@ import {InjectionToken} from '@angular/core';
 export interface PlatformConfig {
   document?: string;
   url?: string;
+  useAbsoluteUrl?: boolean;
```
alan-agius4, line 19: Nit: can we add a comment here, on what this option does?

I do understand what it does, but users will probably not, unless they look at thIs PR.

CaerusKaru: I think we should have a docs followup to make this option known in the guide. The docs for this object in general are pretty poor.

packages/platform-server/src/tokens.ts
```diff
@@ -14,8 +14,25 @@ import {InjectionToken} from '@angular/core';
  * @publicApi
  */
 export interface PlatformConfig {
+  /**
+   * The initial DOM to use to bootstrap the server app.
```
alan-agius4, line 18: Nit: application instead of app

alan-agius4 (approved):
LGTM

One small nit

kyliau:
Note: We've decided to rollback this feature in v10.0.
I've set the target to `master-only` for this to go out in `v10.1`

IgorMinar (approved):
LGTM! Thanks!

Reviewed-for: public-api

packages/platform-server/test/integration_spec.ts
```diff
@@ -793,103 +793,147 @@ describe('platform-server integration', () => {
...
-      ]);
-      const ref = await platform.bootstrapModule(HttpClientExampleModule);
-      const mock = ref.injector.get(HttpTestingController) as HttpTestingController;
-      const http = ref.injector.get(HttpClient);
-      ref.injector.get(NgZone).run(() => {
-        http.get<string>('/testing').subscribe((body: string) => {
-          NgZone.assertInAngularZone();
-          expect(body).toEqual('success!');
+    describe('relative requests', () => {
+      it('base config', async () => {
```
jelbourn, line 797: Could you reword these test names to read a sentences, similar to how they were before? I find that this structure helps me understand what the test intends to assert more quickly

jelbourn (approved):
LGTM

Reviewed-for: public-api

pkozlowski-opensource (approved):
LGTM

Reviewed-for: public-api

kyliau:
presubmit: http://test/OCL:318067014:BASE:318067048:1593012031031:2c21e0f0

kyliau (approved):

chamikasandamal:
This is not working for me in `10.0.4`. Is there any working example for this?

CaerusKaru:
It's only a part 10.1. It was reverted out of 10.0.

chamikasandamal:
But in the documentation it says,

If you are using one of the @nguniversal/*-engine packages (such as @nguniversal/express-engine), this is taken care for you automatically. You don't need to do anything to make relative URLs work on the server.

So I have to keep using the universal interceptor until 10.1?

CaerusKaru:
Yes

PR #37545, opened 2020-06-11
refactor(compiler-cli): skip class decorators in tooling constructor parameters transform

devversion:
We recently added a transformer to NGC that is responsible for downleveling Angular
decorators and constructor parameter types. The primary goal was to mitigate a
TypeScript limitation/issue that surfaces in Angular projects due to the heavy
reliance on type metadata being captured for DI. Additionally this is a
pre-requisite of making `tsickle` optional in the Angular bazel toolchain.

See: 401ef71ae5b01be95d124184a0b6936fc453a5d4 for more context on this.

Another (less important) goal was to make sure that the CLI can re-use
this transformer for its JIT mode compilation. The CLI (as outlined in
the commit mentioned above), already has a transformer for downleveling
constructor parameters. We want to avoid this duplication and exported
the transform through the tooling-private compiler entry-point.

Early experiments in using this transformer over the current one, highlighted
that in JIT, class decorators cannot be downleveled. Angular relies on those
to be invoked immediately for JIT so that factories etc. are generated upon loading
due to side-effects of those decorators.

Before downleveling decorators, upon file loading, the decorator function is invoked
immediately. Resulting in the e.g. factories being generated. Now with the downleveled
ones though, the decorator function is not invoked immediately. This results in missing
factories that are _required_ for JIT compilation.

The transformer we exposed, always downleveles such class decorators
though, so that would break CLI's JIT mode. We can address the CLI's
needs by adding another flag to skip class decorators. This will allow
us to continue with the goal of de-duplication.

### DIFF
--- a/packages/compiler-cli/src/transformers/downlevel_decorators_transform.ts
+++ b/packages/compiler-cli/src/transformers/downlevel_decorators_transform.ts
@@ -327,10 +327,17 @@ interface ParameterDecorationInfo {
  * @param diagnostics List which will be populated with diagnostics if any.
  * @param isCore Whether the current TypeScript program is for the `@angular/core` package.
  * @param isClosureCompilerEnabled Whether closure annotations need to be added where needed.
+ * @param skipClassDecorators Whether class decorators should be skipped from downleveling.
+ *   This is useful for JIT mode where class decorators should be preserved as they could rely
+ *   on immediate execution. e.g. downleveling `@Injectable` means that the injectable factory
+ *   is not created, and injecting the token will not work. If this decorator would not be
+ *   downleveled, the `Injectable` decorator will execute immediately on file load, and
+ *   Angular will generate the corresponding injectable factory.
  */
 export function getDownlevelDecoratorsTransform(
     typeChecker: ts.TypeChecker, host: ReflectionHost, diagnostics: ts.Diagnostic[],
-    isCore: boolean, isClosureCompilerEnabled: boolean): ts.TransformerFactory<ts.SourceFile> {
+    isCore: boolean, isClosureCompilerEnabled: boolean,
+    skipClassDecorators: boolean): ts.TransformerFactory<ts.SourceFile> {
   return (context: ts.TransformationContext) => {
     let referencedParameterTypes = new Set<ts.Declaration>();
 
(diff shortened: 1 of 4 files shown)


alxhub (approved):

packages/compiler-cli/src/transformers/downlevel_decorators_transform.ts
```diff
@@ -327,10 +327,15 @@ interface ParameterDecorationInfo {
...
+ * @param skipClassDecorators Whether class decorators should be skipped from downleveling.
+ *   This is useful for JIT mode where class decorators should be preserved as they can rely
+ *   on immediate execution. e.g. downleveling `@Injectable` means that the injectable factory
+ *   is not created. Angular does not generate it on-demand when it comes across the class.
  */
 export function getDownlevelDecoratorsTransform(
     typeChecker: ts.TypeChecker, host: ReflectionHost, diagnostics: ts.Diagnostic[],
-    isCore: boolean, isClosureCompilerEnabled: boolean): ts.TransformerFactory<ts.SourceFile> {
+    isCore: boolean, isClosureCompilerEnabled: boolean,
+    skipClassDecorators: boolean): ts.TransformerFactory<ts.SourceFile> {
```
alxhub, line 338: Instead of `skipClassDecorators`, let's go with `emitClassDecorators` actually. Positive effect flags are easier to understand than negative effect flags.

clydin (approved):

clydin (approved):

clydin (approved):
Reviewed-For: tooling-cli-shared-api

packages/compiler-cli/src/transformers/downlevel_decorators_transform.ts
devversion: I'd actually prefer it that way because the transform primarily is concerned with all decorators (hence it's name), so this is rather opt-out.

I don't feel too strongly about it, so happy to change if you feel strongly about it.

packages/compiler-cli/src/transformers/downlevel_decorators_transform.ts
```diff
@@ -327,10 +327,15 @@ interface ParameterDecorationInfo {
  * @param diagnostics List which will be populated with diagnostics if any.
  * @param isCore Whether the current TypeScript program is for the `@angular/core` package.
  * @param isClosureCompilerEnabled Whether closure annotations need to be added where needed.
+ * @param skipClassDecorators Whether class decorators should be skipped from downleveling.
+ *   This is useful for JIT mode where class decorators should be preserved as they can rely
+ *   on immediate execution. e.g. downleveling `@Injectable` means that the injectable factory
+ *   is not created. Angular does not generate it on-demand when it comes across the class.
```
kyliau, line 333: > Angular does not generate it on-demand when it comes across the class.

Instead of saying what Angular doesn't do in this case, perhaps it's clearer to explain what it does.
The "doesn't do" case is already explained in previous sentence.

Consider something like
> Instead, Angular will retain the default `__decorator` method from `tslib`.

kyliau (approved):

kyliau (approved):
Reviewed for: tooling-cli-shared-api

kyliau (approved):
Reviewed-for: tooling-cli-shared-api

packages/compiler-cli/src/transformers/downlevel_decorators_transform.ts
devversion: Updated. Should be better now! thx

IgorMinar:
[presubmit](https://test.corp.google.com/ui#id=OCL:316507354:BASE:316507389:1592245611935:6e9e3022)

PR #37558, opened 2020-06-12
ci(compiler-cli): exempt compiler-cli .bazel files from dev-infra approval

alxhub:
Previously, dev-infra approval (via PullApprove) was required for all
.bazel files in the monorepo, including those in packages/compiler-cli.

The compiler-cli package is a little special in this sense:
 * it's not shipped to NPM in the APF
 * it uses lots of internal subpackages to organize and test its code

As a result:
 * changes to compiler-cli BUILD.bazel files are not user visible and
   don't have larger implications for the packages published to NPM,
   unlike changes to other BUILD.bazel files in the repo
 * the requirement for dev-infra approval for BUILD.bazel changes is
   overly burdensome, because compiler-cli build files change more
   rapidly than those of other packages.

This commit exempts the compiler-cli's build files from the requirement
for dev-infra approval. It will be sufficient for such files to be
approved by the normal compiler reviewers.

### DIFF
--- a/.pullapprove.yml
+++ b/.pullapprove.yml
@@ -1051,7 +1051,7 @@ groups:
     conditions:
       - *can-be-global-approved
       - >
-        contains_any_globs(files.exclude("CHANGELOG.md"), [
+        contains_any_globs(files.exclude("CHANGELOG.md").exclude("packages/compiler-cli/**/BUILD.bazel"), [
           '*',
           '.circleci/**',
           '.devcontainer/**',


thecoder8890 (approved):

IgorMinar:
@josephperrott, I believe that this is the change that was discussed in the dev-infra meeting two weeks ago and the conclusion was that `.bzl` files were a no-no, but `Build.bazel` files were ok. Can you please confirm?

@alxhub can you please adjust the change to exclude only the build files? thanks

josephperrott:
@IgorMinar correct, we want to still ensure that `.bzl` files are managed by dev-infra as we are wanting to limit the number of macros/custom rules created unnecessarily.

But for the `.build` files themselves I think we can try out changing the review/ownership model.

IgorMinar:
and I adjusted the labels to include the patch branch, we don't want the master and patch to diverge in something as common as code ownership.

josephperrott (approved):
LGTM

IgorMinar (approved):
LGTM. Thanks!

Reviewed-for: code-ownership

alxhub:
Caretaker: the AIO failures are unrelated/flaky. This PR only affects the pullapprove config.

PR #37560, opened 2020-06-12
ci: upload build results to ResultStore for CI linux bazel executions

josephperrott:
Bazel invocations will upload to ResultStore to allow for us to have better viewing
of execution/build logs.  This is only done on CI as the BES API requires credentials
from service accounts, rather than end user accounts.

Example of what the ResultStore entry looks like: https://source.cloud.google.com/results/invocations/410a5790-54eb-43e8-96eb-c3cfa0f8960d

### DIFF
--- a/.circleci/bazel.linux.rc
+++ b/.circleci/bazel.linux.rc
@@ -19,4 +19,12 @@ build --local_ram_resources=14336
 
 # All build executed remotely should be done using our RBE configuration.
 build:remote --google_default_credentials
+
+# Upload to GCP's Build Status viewer to allow for us to have better viewing of execution/build
+# logs. This is only done on CI as the BES (GCP's Build Status viewer) API requires credentials
+# from service accounts, rather than end user accounts.
+build:remote --bes_backend=buildeventservice.googleapis.com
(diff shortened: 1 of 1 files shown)


.circleci/bazel.linux.rc
```diff
@@ -19,4 +19,12 @@ build --local_ram_resources=14336
 
 # All build executed remotely should be done using our RBE configuration.
 build:remote --google_default_credentials
+
+# Upload to ResultStore to allow for us to have better viewing of execution/build logs.  This
```
devversion, line 23:
```suggestion
# Upload to ResultStore to allow for us to have better viewing of execution/build logs. This
```

Do we have something where we can link to? I've been looking for `ResultStore` but cannot find anything 😄

devversion (approved):

.circleci/bazel.linux.rc
josephperrott: I pushed again with updated comment text.  I went with something more descriptive rather than the project name since its an Alpha project that doesn't really have any public facing documentation.

PR #37575, opened 2020-06-14
feat(router): add support for canLoadChild

dimakuba:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

It's not possible to define an array of DI tokens in order to determine if the current user is allowed to load a child of the component.

## What is the new behavior?
Allows to define an array of DI tokens in order to determine if the current user is allowed to load a child of the component.

- [ ] Yes
- [x] No

### DIFF
--- a/goldens/public-api/router/router.d.ts
+++ b/goldens/public-api/router/router.d.ts
@@ -295,6 +295,7 @@ export declare interface Route {
     canActivateChild?: any[];
     canDeactivate?: any[];
     canLoad?: any[];
+    canLoadChild?: any[];
     children?: Routes;
     component?: Type<any>;
     data?: Data;
--- a/goldens/size-tracking/integration-payloads.json
+++ b/goldens/size-tracking/integration-payloads.json
@@ -39,7 +39,7 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 2289,
-        "main-es2015": 246085,
+        "main-es2015": 246667,
         "polyfills-es2015": 36938,
         "5-es2015": 751
       }
@@ -49,7 +49,7 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 2289,
-        "main-es2015": 221268,
+        "main-es2015": 221813,
         "polyfills-es2015": 36938,
         "5-es2015": 779
       }
--- a/packages/router/src/apply_redirects.ts
+++ b/packages/router/src/apply_redirects.ts
@@ -11,12 +11,13 @@ import {EmptyError, from, Observable, Observer, of} from 'rxjs';
 import {catchError, concatAll, every, first, map, mergeMap, tap} from 'rxjs/operators';
 
 import {LoadedRouterConfig, Route, Routes} from './config';
-import {CanLoadFn} from './interfaces';
+import {CanLoadChildFn, CanLoadFn} from './interfaces';
+import {prioritizedGuardValue} from './operators/prioritized_guard_value';
 import {RouterConfigLoader} from './router_config_loader';
 import {defaultUrlMatcher, navigationCancelingError, Params, PRIMARY_OUTLET} from './shared';
 import {UrlSegment, UrlSegmentGroup, UrlSerializer, UrlTree} from './url_tree';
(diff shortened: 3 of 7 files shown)


atscott:
Hi @dimakuba, thanks for the suggestion. This functionality is already covered in the `canLoad` guard, an "Interface that a class can implement to be a guard deciding if children can be loaded." https://angular.io/api/router/CanLoad#canload

If there's a case that you can't solve with `CanLoad`, please add a test for that and open a new PR. I can get your current tests to pass by using `canLoad` instead of `canLoadChild`

```
it('should not load when any canLoadChild guards return false',
...
      const config = [{
        path: '',
        canLoad: ['guard1', 'guard2'],
        children: [
          {path: 'a', canLoad: ['guard1', 'guard2'], component: ComponentA, loadChildren: 'children'}
        ]
      }];
```

dimakuba:
Hi @atscott, by using `CanLoad` we need to duplicate guards definitions for each descendants. With the proposed solution we can just define guards list on a parent route config to apply for all descendants:

Without `CanLoadChild`:
```
[
  {
    path: '',
    children: [
      {path: 'a', canLoad: ['checkPermissions'], component: ComponentA, loadChildren: 'children'},
      {path: 'b', canLoad: ['checkPermissions'], component: ComponentB, loadChildren: 'children'},
      {path: 'c', canLoad: ['checkPermissions'], component: ComponentC, loadChildren: 'children'},
      {path: 'd', canLoad: ['checkPermissions'], component: ComponentD, loadChildren: 'children'},
      ...
    ]
  }
]
```

With `CanLoadChild`:
```
[
  {
    path: '',
    canLoadChild: ['checkPermissions'],
    children: [
      {path: 'a', component: ComponentA, loadChildren: 'children'},
      {path: 'b', component: ComponentB, loadChildren: 'children'},
      {path: 'c', component: ComponentC, loadChildren: 'children'},
      {path: 'd', component: ComponentD, loadChildren: 'children'},
      ...
    ]
  }
]
```

atscott:
Hi @dimakuba, thanks for the clarification and I can get behind the goal of reducing repetition. I think this could instead be resolved by extracting the child routes to an intermediate module:

```
[
  {
    path: '',
    canLoad: ['checkPermissions'],
    loadChildren: () => import('root-children-routes.module').then(m => m.RootChildrenRoutes)
  }
]

===

@NgModule({
  RouterModule.forChild([{
    path: '',
    children: [
      {path: 'a', component: ComponentA, loadChildren: 'children'},
      {path: 'b', component: ComponentB, loadChildren: 'children'},
      {path: 'c', component: ComponentC, loadChildren: 'children'},
      {path: 'd', component: ComponentD, loadChildren: 'children'},
      ...
    ]
  }
export class RootChildrenRoutes {}

})
```
example: https://stackblitz.com/angular/vrbbdnnlxam?file=src%2Fapp%2Froot-children-routing.module.ts

I understand that this is a bit more work, but from the framework perspective, we need to ensure that we're balancing API surface, code size, and code complexity with features that make the API easier to use. Additionally, we need to consider these changes in respect to how often it would be used. When there's a relatively straightforward workaround, we lean towards avoiding unnecessary bloat.

dimakuba:
This is why we have `CanActivateChild` to reduce boilerplates of code duplication by using shared guards. My thought was that it would be useful for `CanLoad` guards as well.

atscott:
Hi @dimakuba, totally fair points. Thanks for being patient and arguing in favor of making this change! I'm re-opening this but can't make any promises on when it will be fully reviewed -- there's quite a backlog of PRs and issues that need triaging.

atscott:
Hi @dimakuba, I created a [feature request](https://github.com/angular/angular/issues/37628) that references this PR. For the reasons I mentioned before, I'm going to close this again, at least for now. We may revisit this idea in the future.
For now we'd like to focus on addressing bugs and highly requested features. 
It's important to note that, while relatively small, this change increases the bundle size for all users of the Router, even if they don't use `canLoadChild`.

Issue #37577, opened 2020-06-14
Unable to distinguish between different NgZones used on the same page


KevinCathcart:
# 🚀 feature request

This is really about an API design flaw/limitation, so it falls somewhere between a bug and a feature request. However there is a specific situation in the upgrade package that could arguably be considered a bug.

### Affected Package
The issue is caused by package @angular/core.

### Description

`NgZone.isInAngularZone()`, is a static method, which means that it can only answer if we are currently running in any Angular Zone, rather than the one for the "current" application.

Angular has always been designed to support multiple applications running on a single page. While the CLI is not particularly conducive to this situation, that does not change this design intent. 

However, by default, each application will have its own NgZone, and underlying zone. Thus if you have two apps called App1 and App2, they will have different zones, unless you go out of your way to avoid that.

The problem is that you cannot tell which one you are running in. `NgZone.isInAngularZone()` will return true regardless of which of App1 or App2's zone you are running in. Now this is not always a particularly huge deal. That API is only used twice in angular outside the tests or assertions. The first is merely to print a warning message if navigation is triggered, so confusing the zones only means the warning is not printed in a case where it should. Not really a huge deal, as nothing will actually break.

However, if used in a pattern like this: `NgZone.isInAngularZone() ? cb() : ngZone.run(cb)`, that is a problem, since the zone we are currently running in could be the one for App2, when the callback needs to run in the zone for App1. If that code seems at all unlikely, it shouldn't, since it came from this repository. It only runs when using `downgradeModule()`, but it is still arguably buggy:

https://github.com/angular/angular/blob/d1ea1f4c7f3358b730b0d94e65b00bc28cae279c/packages/upgrade/src/common/src/downgrade_component.ts#L90

I expect a very similar pattern occurs from time to time in third party libraries, or even in application code. 

### Describe the solution you'd like
Ideally there would be an instance method on NgZone that would return true if you are running in that specific zone.

Also, ideally the existing static methods would be deprecated, in favor of new versions named something like `IsInAnyAngularZone`, to indicate that it returns true if you are in any such zone. This would encourage existing users of this api to decide if they they are concerned with running in the zone for a specific app, or if they care about running in any angular zone. For example, some code may simply want to make sure it is running outside of any angular zone for performance or correctness reasons. 

### Describe alternatives you've considered
A different possibility would be to move the zone to the platform level, and force all apps to use the same zone. That would be a big breaking change, with implications on several other apis. It also eliminates the change-detection speedup that comes from each app having their own zone.

PR #37581, opened 2020-06-15
fix(core): unify the signature between ngZone and noopZone

JiaLiPassion:
Now we have two implementations of Zone in Angular, one is NgZone, the other is NoopZone.
They should have the same signatures, includes
1. properties
2. static methods
3. methods

In this PR, unify the signatures of the two implementations, and remove the unnecessary cast.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/zone/ng_zone.ts
+++ b/packages/core/src/zone/ng_zone.ts
@@ -177,7 +177,7 @@ export class NgZone {
    * If a synchronous error happens it will be rethrown and not reported via `onError`.
    */
   run<T>(fn: (...args: any[]) => T, applyThis?: any, applyArgs?: any[]): T {
-    return (this as any as NgZonePrivate)._inner.run(fn, applyThis, applyArgs) as T;
+    return (this as any as NgZonePrivate)._inner.run(fn, applyThis, applyArgs);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


gkalpak (approved):

packages/core/src/zone/ng_zone.ts
```diff
@@ -388,19 +388,30 @@ export class NoopNgZone implements NgZone {
...
+  static isInAngularZone(): boolean {
+    return false;
+  }
+
+  static assertInAngularZone(): void {
+    throw new Error('Expected to be in Angular Zone, but it is not!');
+  }
+
+  static assertNotInAngularZone(): void {}
+
```
mhevery, line 400: Why do you think that the `static` methods should be part of this API? My understanding is that the existing `static` methods on `NgZone` have the correct implementation and correctly return if the code is in `NgZone`. The new methods here, have the code hard wired, and do not actually check. (It is possible that we have two apps on single page and some run `NgZone` and some run `NoopNgZone` I would just remove these `static` methods.

mhevery (changes requested):

packages/core/src/zone/ng_zone.ts
JiaLiPassion: @mhevery , yeah, you are right, in fact there is an issue about this one (mixing two angular app in the same page),https://github.com/angular/angular/issues/37577,
I will remove all those static methods.

mhevery (approved):

mhevery:
[presubmit](http://test/OCL:320657636:BASE:320657672:1594410747140:94dbe14c)

PR #37582, opened 2020-06-15
fix(zone.js): defineProperty patch should not swallow error

### LINKED ISSUE #37432: ZoneJS: legacy Object.defineProperty patch breaks configurable descriptor attribute
# bug report

### Affected Package

zone.js

### Is this a regression?

Yes, but super back in the past! It broke with v0.6.24.

### Description

ZoneJS has this super old legacy `Object.defineProperty` patch that basically partially re-implements
the `Object.defineProperty` functionality in order to resolve some custom elements issue. I'm confident that these issues are no longer surfacing and patching `Object.defineProperty` is not the right solution these days.

The issue is that this overwritten `Object.defineProperty` function behaves different to the native implementation. i.e. by default defined properties are not configurable. In ZoneJS, with the patched version, properties are _always_ configurable. This means that:

1. The default `configurable` property descriptor attribute is not respected.
2. Defined properties are by default always `configurable: true`. This doesn't match the native implementation.

Due to these inconsistencies, ZoneJS could potentially hide app issues that would usually cause errors in production. e.g. when using differential loading, legacy patches might be used in tests, but not always in production. So, calling `defineProperty` multiple times in your app might work with the legacy patches in tests and legacy browsers, but otherwise cause a `Uncaught TypeError: Cannot redefine property: X` error in evergreen browsers.

We saw this issue in the Angular Material test harnesses: https://github.com/angular/components/issues/19440

Here is the commit that introduced this logic back in the old repository:
https://github.com/angular/zone.js/commit/383b47905d0a55c0fd87983fa6effd0278cf70f1. And here is the commit that caused the incorrect `configurable` default, and the flag to be not respected at all: https://github.com/angular/zone.js/commit/7b7258b9bf0b4ffd944b2ae015bfb5cb045e3c05.

## Minimal Reproduction

```ts
const x = {}
Object.defineProperty(x, 'defaultPrevented', { get: () => true, configurable: false });
console.error(Object.getOwnPropertyDescriptor(x, 'defaultPrevented'));
```

This prints `configurable: false` for the `defaultPrevented` property without ZoneJS. If the ZoneJS legacy patches are loaded, this is not the case though and `configurable` is always `true`.

## Exception or Error

`Object.defineProperty` behaves different depending on whether ZoneJS legacy patches are loaded or not. ZoneJS could **hide app failures** in tests as `Object.defineProperty` never sets `configurable: false` while in reality, defining a non-configurable property again, causes a `Cannot redeclare property X` runtime error in browsers.

##  Your Environment

**Angular Version:**

Angular v10.0.0-rc.2.
zonejs: 0.10.

**Anything else relevant?**

We should audit if we still need this legacy patch. I'm confident, we wouldn't. I'm hoping we can remove this as it's prone to inconsistencies with the native `Object.defineProperty` behavior.

JiaLiPassion:
Close #37432

zone.js monkey patches the `Object.defineProperty` API long time ago
https://github.com/angular/zone.js/commit/383b47905d0a55c0fd87983fa6effd0278cf70f1
to resolve issues in very old version of Chrome web which override the
property of `CustomElements`, and this is not an issue any longer, so
we may remove this monkey patch, since it may swallow the errors when the user
want to define property on unconfigurable or frozen object properties.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [x] Yes
- [ ] No

Since `zone.js` doesn't patch `Object.defineProperty` and not swallow the error when trying to define property on unconfigurable or `undefined` object, some application currently not throw error will begin to break.

For example, in our `saucelab` test for IE and old firefox/chrome, before this PR, the code/spec are compiled to `commonjs`, and the js code will look like this.

```
Object.defineProperty(exports, "__esModule", { value: true })
...
```

And the `exports` is `undefined`, It should throw error that trying to defineProperty on `undefined`, since `zone.js` monkey patch the `Object.defineProperty` and will catch the error and do nothing, so the test cases will continue to work. 
After this PR, `zone.js` will not do that any more, so the test cases will break.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -653,8 +653,10 @@ jobs:
           name: Starting Saucelabs tunnel service
           command: ./tools/saucelabs/sauce-service.sh run
           background: true
-      - run: yarn tsc -p packages
-      - run: yarn tsc -p modules
+      # add module umd tsc compile option so the test can work
+      # properly in the legacy browsers
+      - run: yarn tsc -p packages --module UMD
+      - run: yarn tsc -p modules --module UMD
       - run: yarn bazel build //packages/zone.js:npm_package
       # Build test fixtures for a test that rely on Bazel-generated fixtures. Note that disabling
       # specific tests which are reliant on such generated fixtures is not an option as SystemJS
--- a/packages/zone.js/lib/browser/define-property.ts
+++ b/packages/zone.js/lib/browser/define-property.ts
@@ -16,8 +16,6 @@ let _defineProperty: any;
 let _getOwnPropertyDescriptor: any;
 let _create: any;
 let unconfigurablesKey: any;
-const registerElementsCallbacks =
-    ['createdCallback', 'attachedCallback', 'detachedCallback', 'attributeChangedCallback'];
 
 export function propertyPatch() {
   zoneSymbol = Zone.__symbol__;
@@ -105,9 +103,11 @@ function _tryDefineProperty(obj: any, prop: string, desc: any, originalConfigura
         return _defineProperty(obj, prop, desc);
       } catch (error) {
         let swallowError = false;
-        if (typeof document !== 'undefined' && obj === document &&
-            registerElementsCallbacks.find(c => c === prop)) {
+        if (prop === 'createdCallback' || prop === 'attachedCallback' ||
+            prop === 'detachedCallback' || prop === 'attributeChangedCallback') {
           // We only swallow the error in registerElement patch
+          // this is the work around since some applications
+          // fail if we throw the error
           swallowError = true;
         }
         if (!swallowError) {
--- a/packages/zone.js/lib/browser/define-property.ts
+++ b/packages/zone.js/lib/browser/define-property.ts
@@ -16,6 +16,8 @@ let _defineProperty: any;
 let _getOwnPropertyDescriptor: any;
 let _create: any;
 let unconfigurablesKey: any;
+const registerElementsCallbacks =
+    ['createdCallback', 'attachedCallback', 'detachedCallback', 'attributeChangedCallback'];
 
 export function propertyPatch() {
   zoneSymbol = Zone.__symbol__;
@@ -102,6 +104,18 @@ function _tryDefineProperty(obj: any, prop: string, desc: any, originalConfigura
       try {
         return _defineProperty(obj, prop, desc);
       } catch (error) {
+        let swallowError = false;
+        if (typeof document !== 'undefined' && obj === document &&
+            registerElementsCallbacks.find(c => c === prop)) {
+          // We only swallow the error in registerElement patch
+          swallowError = true;
+        }
+        if (!swallowError) {
+          throw error;
+        }
+        // TODO: @JiaLiPassion, Some application such as `registerElement` patch
+        // still need to swallow the error, in the future after these applications
+        // are updated, the following logic can be removed.
         let descJson: string|null = null;
         try {
           descJson = JSON.stringify(desc);


modules/tsconfig-umd.json
```diff
@@ -0,0 +1,6 @@
+{
+  "extends": "./tsconfig",
+  "compilerOptions": {
+    "module": "umd"
```
devversion, line 4: Did you consider setting `module: umd` just for the `packages/tsconfig.json` file? I think that _could_ be fine, given that Bazel uses a completely different tsconfig file.

devversion (approved):
Thanks @JiaLiPassion. It would be great if  we can land this PR! It's quite confusing that zone.js currently has the capacity of hiding `Object.defineProperty` errors that could surface in production.

modules/tsconfig-umd.json
JiaLiPassion: Thanks for the review, got it, I will check who are using the `packages/tsconfig.json`, if only for test, I will directly update this file.

JiaLiPassion: @devversion, I am using `yarn tsc -p packages/tsconfig.json --module UMD` to compile the code to `UMD` and still use the other settings from `tsconfig.json`, so the new `tsconfig-umd.json` is no longer needed.

devversion: Awesome! that's a good solution. Could you please add a small comment above (where you call `--module UMD`) on why we prefer UMD in the Saucelabs job? Other than that, looks great!

JiaLiPassion: Sure, I just added the comments, thanks for the review.

mhevery (approved):

mhevery:
[presubmit](http://test/OCL:320657938:BASE:320658017:1594410841335:1a52f446)

atscott:
FYI I'm removing the merge label and adding back the presubmit label, as it seems there is one test that is consistently failing and doesn't appear to be a flake. @mhevery - Please add back the merge label once you've finished investigating.

mhevery:
There is a CustomElement tests in g3 running on ie11-win7 which fails to bootstrap with this change. Seems like it can't be landed in the current form?

Perhaps we can conditionally add the `defineProperty` only if we detect that it is needed?

JiaLiPassion:
@mhevery , got it, I will make `defineProperty` an optional module, and is that possible the `g3 ie11-win7` test cases be updated? I mean after the `defineProperty` become an optional module, the current `g3 ie11-win7` test need to load this module explicitly.

JiaLiPassion:
@mhevery, @atscott , I just added a new commit to make `define-property` patch a new package called `zone-patch-define-property`, by default, it will not be loaded, to enable it, import it after loading zone.js

```
import 'zone.js/dist/zone';
import 'zone.js/dist/zone-patch-define-property';
```

AndrewKushnir:
FYI, I've started [global presubmit](https://test.corp.google.com/OCL:323489151:BASE:323489378:1595901991473:394b7dd) to check behavior in Google's codebase.

josephperrott:
@JiaLiPassion 

After working through the errors in our presubmit it looks like we might need to go about this change a little differently.

From how I understand it, it seems like we are wanting to now have different capabilities/effects from zone.js depending on the browser (with this change it seems IE vs everything else).  This is not something we have done in the past, in the past we have split out features which could be chosen to be shipped or not shipped to the browser, but we always sent the same thing to every browser.

After this change it seems that we now would need to create different zone.js bundles depending on the browser making the request.  This type of feature from the server is something that requires quite a bit of change from our current requirements to be able to serve and use zone.js.  

It seems that a better option at this point in time might be to determine at runtime in the browser if the `defineProperty` monkey patch should be used or not.

JiaLiPassion:
@josephperrott , thank you for the information.

> It seems that a better option at this point in time might be to determine at runtime in the browser if the defineProperty monkey patch should be used or not

Sure, I will update the PR like this.

And in fact this PR is a bug fix, if my understanding is correct, current `g3` app need this `defineProperty` monkey patch in `IE`. Which is not correct, I mean the `app` should not depend on this monkey patch, since it swallow the error and cause some other potential errors. So it need to be fixed in the future eventually.

JiaLiPassion:
@josephperrott, @AndrewKushnir  I have updated the PR, could you please create another presubmit, thanks!

AndrewKushnir:
[Presubmit](http://test/OCL:323489151:BASE:325920118:1597104755933:89a7f39d)

AndrewKushnir:
Hi @JiaLiPassion, thanks for updating the PR!

We ran a presubmit and found out that there is a number of targets that are failing now with this change. One pattern that we've observed is:

```
const mockObj = {};
Object.defineProperty(mockObj, 'getData', {get: () => null});
// ...
spyOnProperty(mock, 'getData', 'get').and.returnValue([]);
```

The error message thrown is:
```
Error: Uncaught (in promise): Error: <spyOnProperty> : getData is not declared configurable
Usage: spyOnProperty(<object>, <propName>, [accessType])
```

The high-level pattern is that by default the `Object.defineProperty` sets the property as non-configurable, but when ZoneJS is present, [it becomes configurable](https://github.com/angular/angular/blob/master/packages/zone.js/lib/browser/define-property.ts#L76). You've described that in the PR description too, so it's an instance of the proposed breaking change.

We can probably do some cleanup in tests, but this change might actually break more apps in some subtle ways, so I was wondering if we can come up with some strategy on how to handle this (externally as well as internally in Google's codebase). We should also consider if changing this behavior now worth introducing a breaking change.

cc @mhevery @josephperrott

JiaLiPassion:
@AndrewKushnir , thank you very much for the information, I just updated the PR to only remove the logic that `swallow the error`, so the `configurable=true` logic is still there since currently the bigger issue is the error not been thrown correctly.

Could you run another `presubmit` to check whether there are any failing cases? Thanks!

AndrewKushnir:
Hi @JiaLiPassion, thanks for the update. I've started a new [global presubmit](https://test.corp.google.com/ui#id=OCL:326166812:BASE:326167039:1597208052228:4d792240) and will let you know the results.

AndrewKushnir:
Hi @JiaLiPassion, I ran a new presubmit and found a few targets that require attention, but it should be manageable to update them. I've re-requested review from Misko and Paul (since this PR changed a lot since the latest review) and added the "breaking changes" flag (since the code may start throwing errors which were swallowed before). Thank you.

JiaLiPassion:
@AndrewKushnir , got it, thank you!

mhevery (approved):
Re-approved per request.

AndrewKushnir:
Just a quick update: after running tests in Google's codebase and additional investigation, it turned out that the changes in this PR trigger errors in IE 11 (related to `createdCallback`). @JiaLiPassion has additional information and will look into possible solutions. Thank you.

JiaLiPassion:
@AndrewKushnir , I just updated the PR.
1. `regsiterElement` will still use the old logic, which means `defineProperty` will change the `configurable` to true and also catch error.
2. The other parts will use the new logic, only change the `configurable` to true and not swallow the error.

Could you re-run the presubmit? Thanks!

AndrewKushnir:
Thanks for the update @JiaLiPassion 👍 I've [started a new global presubmit](https://test.corp.google.com/OCL:329248096:BASE:329248204:1598856555636:87795b49) and will keep you updated.

.circleci/config.yml
```diff
@@ -653,8 +653,10 @@ jobs:
           name: Starting Saucelabs tunnel service
           command: ./tools/saucelabs/sauce-service.sh run
           background: true
-      - run: yarn tsc -p packages
-      - run: yarn tsc -p modules
+      # add module umd tsc compile option so the test can work
+      # properly in the legacy browsers
+      - run: yarn tsc -p packages --module UMD
+      - run: yarn tsc -p modules --module UMD
```
AndrewKushnir, line 659: I'd propose to extract this change to a separate commit (which would also explain why this change is needed) within this PR (if this change is needed in a context of these changes).

AndrewKushnir:
Quick update after running tests in Google's codebase: the Windows-based target is still failing with the new set of changes. It looks like the problem was caused by the polyfill that tries to patch the `createdCallback` (actually adding missing functionality in IE) and as a result, the error is being thrown. Probably the check (on which property we try to patch and whether an error should be handled) should be moved to lower level code (such as `_redefineProperty`). @JiaLiPassion has some additional information that I shared with him directly. Thank you.

.circleci/config.yml
JiaLiPassion: Got it, thanks, I just split the commits.

AndrewKushnir:
Thanks for additional updates @JiaLiPassion. I've started a [new presubmit](http://test/OCL:329412159:BASE:329412181:1598920300679:d97d5bcb) and will let you know the results once I have them.

devversion:
> We can probably do some cleanup in tests, but this change might actually break more apps in some subtle ways, so I was wondering if we can come up with some strategy on how to handle this (externally as well as internally in Google's codebase). We should also consider if changing this behavior now worth introducing a breaking change.

It's a good question whether this is worth introducing a breaking change. Personally, I'd definitely say yes. Currently ZoneJS modifies native browser behavior. This wouldn't be a big deal at first glance, but it becomes a quite impacting issue IMO if the legacy browser ZoneJS bundle is conditionally loaded (as done in the CLI IIRC). It can then happen that your tests always pass, but once the application runs in production with a non-legacy browser, the application could suddenly break with runtime errors.

devversion (approved):
LGTM. Sounds reasonable to make an exception for the `registerElement` patch.

AndrewKushnir:
FYI, I've started a [new presubmit](http://test/OCL:329536399:BASE:329536433:1598982163581:7a62a6c0) and will keep this thread updated.

AndrewKushnir:
Hi @JiaLiPassion, I'm still in a process of running tests in Google's codebase, but everything looks good so far. Since this is a breaking change, could you please update commit message to include the `BREAKING CHANGES` section (for ex. as it's done in this commit: https://github.com/angular/angular/pull/38354/commits/9d59b7de0e276f208cd0a7b1775f8aa6f0852b72)?

I've re-requested review from @mhevery once again, since implementation has changed since his last review. Misko, could you please have a look again?

Thank you.

AndrewKushnir:
FYI, global presubmit went well, @JiaLiPassion plz add the "action: merge" label once commit message is updated and there are no additional changes planned on your side. Thank you.

JiaLiPassion:
@AndrewKushnir , thank you so much for the great help!

packages/zone.js/lib/browser/define-property.ts
```diff
@@ -102,6 +102,20 @@ function _tryDefineProperty(obj: any, prop: string, desc: any, originalConfigura
       try {
         return _defineProperty(obj, prop, desc);
       } catch (error) {
+        let swallowError = false;
+        if (prop === 'createdCallback' || prop === 'attachedCallback' ||
+            prop === 'detachedCallback' || prop === 'attributeChangedCallback') {
+          // We only swallow the error in registerElement patch
+          // this is the walk around since some applications
```
mhevery, line 109:
```suggestion
          // this is the work around since some applications
```

mhevery (approved):

packages/zone.js/lib/browser/define-property.ts
JiaLiPassion: Got it, thanks

PR #37583, opened 2020-06-15
feat(core): rename async to waitForAsync to avoid confusing

JiaLiPassion:
@angular/core/testing provide `async` test utility, but the name `async` is
confusing with the javascript keyword `async`. And in some test case, if you
want to use both the `async` from `@angular/core/testing` and `async/await`,
you may have to write the code like this.

```typescript
function asyncFunc(): Promise<{value: number}> {
    return new Promise(res => setTimeout(() => { res({value: 1}); }));
  }

  function timeout(obj) {
    setTimeout(() => {
      obj.value += 1;
    });
  }

  it('test async', async () => {
    const r: {value: number} = await asyncFunc();
    timeout(r);
    setTimeout(() => {
      expect(r.value).toBe(2);
    });
   });
```

So in this PR, the `async` is renamed to `asyncTest`.

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [x] Yes
- [ ] No

### DIFF
--- a/packages/core/testing/src/async.ts
+++ b/packages/core/testing/src/async.ts
@@ -16,7 +16,7 @@ import {asyncFallback} from './async_fallback';
  * Example:
  *
  * ```
- * it('...', async(inject([AClass], (object) => {
+ * it('...', waitForAsync(inject([AClass], (object) => {
  *   object.doSomething.then(() => {
  *     expect(...);
  *   })
@@ -25,12 +25,12 @@ import {asyncFallback} from './async_fallback';
  *
  * @publicApi
  */
-export function async(fn: Function): (done: any) => any {
+export function waitForAsync(fn: Function): (done: any) => any {
   const _Zone: any = typeof Zone !== 'undefined' ? Zone : null;
   if (!_Zone) {
     return function() {
       return Promise.reject(
-          'Zone is needed for the async() test helper but could not be found. ' +
+          'Zone is needed for the waitForAsync() test helper but could not be found. ' +
           'Please make sure that your environment includes zone.js/dist/zone.js');
     };
   }
@@ -42,4 +42,13 @@ export function async(fn: Function): (done: any) => any {
   // TODO @JiaLiPassion, remove this after all library updated to
   // newest version of zone.js(0.8.25)
   return asyncFallback(fn);
+}
+
+/**
+ * @deprecated use `waitForAsync()`, (expected removal in v12)
+ * @see {@link waitForAsync}
+ * @publicApi
+ * */
+export function async(fn: Function): (done: any) => any {
+  return waitForAsync(fn);
 }
--- a/aio/content/examples/forms-overview/src/app/app.component.spec.ts
+++ b/aio/content/examples/forms-overview/src/app/app.component.spec.ts
@@ -1,31 +1,31 @@
-import { TestBed, async } from '@angular/core/testing';
+import { TestBed, waitForAsync } from '@angular/core/testing';
+
 import { AppComponent } from './app.component';
-import { TemplateModule } from './template/template.module';
 import { ReactiveModule } from './reactive/reactive.module';
+import { TemplateModule } from './template/template.module';
 
 describe('AppComponent', () => {
-
-  beforeEach(async(() => {
-    TestBed.configureTestingModule({
-      imports: [ReactiveModule, TemplateModule],
-      declarations: [
-        AppComponent
-      ],
-    }).compileComponents();
+  beforeEach(waitForAsync(() => {
+    TestBed
+        .configureTestingModule({
+          imports: [ReactiveModule, TemplateModule],
+          declarations: [AppComponent],
+        })
+        .compileComponents();
   }));
 
-  it('should create the app', async(() => {
-    const fixture = TestBed.createComponent(AppComponent);
-    const app = fixture.componentInstance;
+  it('should create the app', waitForAsync(() => {
+       const fixture = TestBed.createComponent(AppComponent);
+       const app = fixture.componentInstance;
 
-    expect(app).toBeTruthy();
-  }));
+       expect(app).toBeTruthy();
+     }));
 
-  it('should render title', async(() => {
-    const fixture = TestBed.createComponent(AppComponent);
-    fixture.detectChanges();
+  it('should render title', waitForAsync(() => {
+       const fixture = TestBed.createComponent(AppComponent);
+       fixture.detectChanges();
 
-    const compiled = fixture.debugElement.nativeElement;
-    expect(compiled.querySelector('h1').textContent).toContain('Forms Overview');
-  }));
+       const compiled = fixture.debugElement.nativeElement;
+       expect(compiled.querySelector('h1').textContent).toContain('Forms Overview');
+     }));
 });
--- a/aio/content/examples/forms-overview/src/app/reactive/favorite-color/favorite-color.component.spec.ts
+++ b/aio/content/examples/forms-overview/src/app/reactive/favorite-color/favorite-color.component.spec.ts
@@ -1,19 +1,18 @@
-import { async, ComponentFixture, TestBed } from '@angular/core/testing';
+import { ComponentFixture, TestBed, waitForAsync } from '@angular/core/testing';
 import { ReactiveFormsModule } from '@angular/forms';
 
-import { FavoriteColorComponent } from './favorite-color.component';
 import { createNewEvent } from '../../shared/utils';
+import { FavoriteColorComponent } from './favorite-color.component';
 
 describe('Favorite Color Component', () => {
   let component: FavoriteColorComponent;
   let fixture: ComponentFixture<FavoriteColorComponent>;
 
-  beforeEach(async(() => {
-    TestBed.configureTestingModule({
-      imports: [ ReactiveFormsModule ],
-      declarations: [ FavoriteColorComponent ]
-    })
-    .compileComponents();
+  beforeEach(waitForAsync(() => {
+    TestBed
+        .configureTestingModule(
+            {imports: [ReactiveFormsModule], declarations: [FavoriteColorComponent]})
+        .compileComponents();
   }));
 
   beforeEach(() => {
--- a/aio/content/examples/forms-overview/src/app/template/favorite-color/favorite-color.component.spec.ts
+++ b/aio/content/examples/forms-overview/src/app/template/favorite-color/favorite-color.component.spec.ts
@@ -1,19 +1,16 @@
-import { async, ComponentFixture, TestBed, tick, fakeAsync } from '@angular/core/testing';
+import { ComponentFixture, fakeAsync, TestBed, tick, waitForAsync } from '@angular/core/testing';
 import { FormsModule } from '@angular/forms';
 
-import { FavoriteColorComponent } from './favorite-color.component';
 import { createNewEvent } from '../../shared/utils';
+import { FavoriteColorComponent } from './favorite-color.component';
 
 describe('FavoriteColorComponent', () => {
   let component: FavoriteColorComponent;
   let fixture: ComponentFixture<FavoriteColorComponent>;
 
-  beforeEach(async(() => {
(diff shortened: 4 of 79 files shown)


petebacondarwin (changes requested):
Reviewed-for: fw-upgrade

Don't forget to update: https://github.com/angular/angular/blob/efdf347f9c04e19c8c4fa8b8c05f4310f42728d5/aio/content/guide/upgrade-setup.md

Otherwise LGTM

And I believe you are also going to rename `fakeAsync()` too.

clydin:
Can you provide some details on when a developer would need to use this:
```
it('test async operations', async(async() => {
  const result = await asyncMethod();
  expect(result).toEqual('expected');
}));
```
instead of this:
```
it('test async operations', async () => {
  const result = await asyncMethod();
  expect(result).toEqual('expected');
});
```

jelbourn:
@JiaLiPassion I'm personally missing some context on this; is this something that's been previously proposed/discussed with the larger team?

JiaLiPassion:
@jelbourn, there is no discussion about this one, I just want to make a proposal by making this PR so it may easier to review and discuss. Thanks.

petebacondarwin (approved):
Reviewed-for: fw-upgrade

atscott (approved):
reviewed for: fw-router

This would need to land in a major release and should have a migration with it, right?

atscott (approved):
reviewed-for: fw-router

pkozlowski-opensource (approved):
The change itself LGTM and I see why we would like to rename those functions. If I understand correctly for now we are just deprecating the `async` / `fakeAsync` functions _and_ removing the usage from the main repo? It would have been slightly more clear with 2 separate commits (one adding a new function + depreciation and another one with usage changes).

Having said all this I guess we will have to plan the actual removal very carefully and be sure to have an automatic migration - otherwise it would be too much of the breaking change (as we can see in the main repo).

Reviewed-for: public-api 
Reviewed-for: fw-core 
Reviewed-for: fw-testing

pkozlowski-opensource:
Also @JiaLiPassion for some reason CI didn't run all the jobs so I'm adding the cleanup label.

JiaLiPassion:
@pkozlowski-opensource , thank you for the review, I have rebased and also split the PR into two commits.

> Having said all this I guess we will have to plan the actual removal very carefully and be sure to have an automatic migration - otherwise it would be too much of the breaking change (as we can see in the main repo).

Yeah, I agree, will try to figure out how to make an automatic migration tool later.

clydin:
I believe this change has broader implications than just the FW changes.  For example, tooling will also be affected (both Angular and third-party).  Instructional material such as courseware, videos, blogs/articles, etc. will need to be updated by their providers as well (this can be a time consuming and expensive endeavor).   As mentioned, a migration would also be of importance as developers would receive potentially hundreds of deprecation warnings after upgrading.

I think it may also be useful to have a larger discussion on whether the `async` helper function should continue to exist moving forward.  And if it should not then a renaming may not be necessary (`fakeAsync` is fairly distinctly named in relation to the `async` keyword).

jelbourn:
@StephenFluin and @mgechev I'd like to get your input on this change and the naming 

TL;DR: having a function named `async` is problematic when there's a native `async` keyword, so this change proposes deprecating the names `async` and `fakeAsync` in favor of `asyncTest` and `fakeAsyncTest`

alfaproject:
I'm with @clydin here. We haven't used `async` ever since we could use native `async/await`. Can't it just be straight up deprecated?

shairez:
I get why `async` needs to change (or to get deprecated).

From my experience - I really didn't see a lot of usage for it as well.

But `fakeAsync` is the more commonly used function and I agree with @clydin that this change will have a huge negative impact in multiple aspects.. while not super necessary since there aren't any collisions with native functions. 

So... I say - do what ever you want with `async`, just [leave `fakeAsync` alone](https://www.youtube.com/watch?v=0wG_h3U568w)  😄

yjaaidi:
+1 to get rid of `async()` somehow :)
The confusing thing with `asyncTest` is that it sounds weird if used with `beforeEach` or `afterEach` (e.g. `beforeEach(asyncTest(...))` ) as we are not talking about a test here.
Maybe `asyncRun` or `asyncWait`? or the opposite `runAsync` or `waitAsync`?

yjaaidi:
By the way, would you agree to replace the component's test schematic:
```typescript
beforeEach(async(() => {
  TestBed.configureTestingModule({})
  .compileComponents();
}));
```
by:
```typescript
beforeEach(async () => {
  await TestBed.configureTestingModule({ })
  .compileComponents();
});
```

this makes tests zoneless friendly by default.


_We might also think about the option below but I am not in favor of it as developers might replace it with a block and forget to add `async / await`._
```typescript
beforeEach(() => TestBed.configureTestingModule({}).compileComponents());
```

cc. @BioPhoton @LayZeeDK

petebacondarwin:
Actually I really like this combo: `runAsync` and `fakeAsync` that would resolve the concerns raised by @clydin , @shairez and @yjaaidi.

LayZeeDK:
I think both `async` and `asyncTest` are equally poor function names for what this one does.

Like many other Zone.js-related things in Angular, this one is pretty difficult to explain and reason about. If I recall correctly, `async` intercepts all monkey-patched APIs through a zone and flushes them after callback execution.

A more descriptive name might be something like `interceptAndFlushAsync` or `waitForAsync`.

@yjaaidi
Regarding the spec generation schematics, I remember reading somewhere that it's not even necessary to run `TestBed#compileComponents` asynchronously if we're using the Angular CLI.

That it's called the same as a keyword has been an issue since Angular version 2.0 since TypeScript has had `async-await` since version 1.7. I'm with @clydin in that this will deprecate a lot of material including Angular.io's testing guide, Angular's codebase and the codebases of a ton of lot of libraries in the ecosystem.

I would love to get a more descriptive name though.

One other aspect that @yjaaidi brings up is related to the flaky future of Zone.js. Angular appications can't target `es2018` as TypeScript output as long as they're using Zone.js unless we force downleveling of `async-await` into the Angular compiler or add an explicit NgZone trigger after each `async-await` statement (also using the compiler). This would need to affect all code imported from `node_modules` as well.

If removing Zone.js from Angular is within scope of +3 major versions (I don't know the roadmap for Angular+Zone.js), would making a big change like this, preceded by a deprecation period, be worth it, seeing that this is only related to NgZone-enabled applications?

JiaLiPassion:
@LayZeeDK, I agree with your points that `async` and `asyncTest` may not be the good names.

> If removing Zone.js from Angular is within scope of +3 major versions (I don't know the roadmap for Angular+Zone.js), would making a big change like this, preceded by a deprecation period, be worth it, seeing that this is only related to NgZone-enabled applications?

`async` and `fakeAsync` are just test helper functions, they have nothing to do with `ngZone`, `async` will call `done` automatically when all `macroTasks and microTasks` finished, and `fakeAsync` just like another version of `jasmine clock and mockDate` to make async test to sync. those functions are using `zone.js` but even Angular doesn't depend on `zone.js` to trigger change detection any longer, the test helper functions still work as expected.

packages/core/testing/src/async.ts
```diff
@@ -43,3 +43,12 @@ export function async(fn: Function): (done: any) => any {
   // newest version of zone.js(0.8.25)
   return asyncFallback(fn);
 }
+
+/**
+ *
```
mhevery, line 48: Needs docs of how to use. (Even if it says "see ...")

packages/core/testing/src/async.ts
```diff
@@ -43,3 +43,12 @@ export function async(fn: Function): (done: any) => any {
...
+ * @publicApi
+ * @deprecated use asyncTest()
```
mhevery, line 50:
```suggestion
 * @deprecated use `asyncTest()` (expected removal in v12)
```

packages/core/testing/src/async.ts
```diff
@@ -25,7 +25,7 @@ import {asyncFallback} from './async_fallback';
  *
  * @publicApi
  */
-export function async(fn: Function): (done: any) => any {
+export function asyncTest(fn: Function): (done: any) => any {
```
mhevery, line 28: `<bikeshedding>`
- `waitForAsync` Describes what the function does. 
- `waitForAsyncToComplete` Describes what the function does in a more complete sentence. 

`</bikeshedding>`

packages/core/testing/src/fake_async.ts
```diff
@@ -43,14 +43,22 @@ export function resetFakeAsyncZone(): void {
...
     return fakeAsyncTestModule.fakeAsync(fn);
   } else {
     return fakeAsyncFallback(fn);
   }
 }
 
+/**
+ *
+ * @publicApi
+ * @deprecated use fakeAsyncTest()
```
mhevery, line 57:
```suggestion
 * @deprecated use `fakeAsyncTest()` (expected removal in v12)
```

packages/core/testing/src/fake_async.ts
```diff
@@ -43,14 +43,22 @@ export function resetFakeAsyncZone(): void {
...
+export function fakeAsyncTest(fn: Function): (...args: any[]) => any {
   if (fakeAsyncTestModule) {
     return fakeAsyncTestModule.fakeAsync(fn);
   } else {
     return fakeAsyncFallback(fn);
   }
 }
 
+/**
+ *
```
mhevery, line 55: Needs JSDOC

packages/core/testing/src/fake_async.ts
```diff
@@ -43,14 +43,22 @@ export function resetFakeAsyncZone(): void {
  *
  * @publicApi
  */
-export function fakeAsync(fn: Function): (...args: any[]) => any {
+export function fakeAsyncTest(fn: Function): (...args: any[]) => any {
```
mhevery, line 46: Not really convinced that renaming `fakeAsync` to `fakeAsyncTest` improves readability.

mhevery (changes requested):

mhevery:
- [ ] The backwards compatible rename can land in patch release. 
- [ ] The automatic migration tool should land before the major release. 
- [ ] Need to discuss with leads if this kind of rename is what we want.

packages/core/testing/src/async.ts
JiaLiPassion: Got it, I will rename it to `waitForAsync`.

petebacondarwin:
I removed the review label and added cleanup because a) it has conflicts and needs a rebase; b) we have agreed (I believe) to use `waitForAsync ` and keep `fakeAsync`.

JiaLiPassion:
@petebacondarwin , I have rebased and rename `async` to `waitForAsync`, also revert `fakeAsync`. Please review, thanks.

petebacondarwin:
I'm afraid the AIO test code needs reformatting

packages/core/testing/src/async.ts
JiaLiPassion: Got it, updated, tahnks.

packages/core/testing/src/async.ts
JiaLiPassion: Added JSDoc `@see {@link waitForAsync}`

JiaLiPassion:
@petebacondarwin , yeah, I updated the format and commit again, please review, thanks.

petebacondarwin (approved):
Reviewed-for: fw-upgrade

mhevery:
[presubmit](http://test/OCL:324084689:BASE:324084717:1596144868193:45b9734f)

mhevery (approved):
LGTM
- [ ] Sorry to be a pain, but could you break the PR into two. One for the change, and one to move `async` to `waitForAsync` We had it broken up but somehow it gote merged.

Reviewed-for: fw-core

mhevery:
[presubmit](http://test/OCL:324084689:BASE:324276416:1596228236860:99c16274)

mhevery:
merge-assistance: global-approval

PR #37587, opened 2020-06-15
feat(compiler-cli): explain why an expression cannot be used in AOT compilations

JoostK:
During AOT compilation, the value of some expressions need to be known at
compile time. The compiler has to ability to statically evaluate expressions
the best it can, but there can be occurrences when an expression cannot be
evaluated statically. For instance, the evaluation could depend on a dynamic
value or syntax is used that the compiler does not understand. Alternatively,
it is possible that an expression could be statically evaluated but the
resulting value would be of an incorrect type.

In these situations, it would be helpful if the compiler could explain why it
is unable to evaluate an expression. To this extend, the static interpreter
in Ivy keeps track of a trail of `DynamicValue`s which follow the path of nodes
that were considered all the way to the node that causes an expression to be
considered dynamic. Up until this commit, this rich trail of information was
not surfaced to a developer so the compiler was of little help to explain
why static evaluation failed, resulting in situations that are hard to debug
and resolve.

This commit adds much more insight to the diagnostic that is produced for static
evaluation errors. For dynamic values, the trail of `DynamicValue` instances
is presented to the user in a meaningful way. If a value is available but not
of the correct type, the type of the resolved value is shown.

Resolves FW-2155

---

Screenshot 2020-06-15 at 13 07 14
Screenshot 2020-06-15 at 13 11 56
Screenshot 2020-06-15 at 13 15 46

### DIFF
--- a/packages/compiler-cli/src/ngtsc/partial_evaluator/src/diagnostics.ts
+++ b/packages/compiler-cli/src/ngtsc/partial_evaluator/src/diagnostics.ts
@@ -10,6 +10,7 @@ import * as ts from 'typescript';
 
 import {makeRelatedInformation} from '../../diagnostics';
 import {Reference} from '../../imports';
+import {FunctionDefinition} from '../../reflection';
 import {DynamicValue, DynamicValueVisitor} from './dynamic';
 import {EnumValue, KnownFn, ResolvedModule, ResolvedValue} from './result';
 
@@ -102,6 +103,16 @@ class TraceDynamicValueVisitor implements DynamicValueVisitor<ts.DiagnosticRelat
         value.node,
         `A value for ${
             description} cannot be determined statically, as it is an external declaration.`)];
+  }
+
+  visitComplexFunctionCall(value: DynamicValue<FunctionDefinition>):
+      ts.DiagnosticRelatedInformation[] {
+    return [
+      makeRelatedInformation(
+          value.node,
+          'Unable to evaluate function call of complex function. A function must have exactly one return statement.'),
+      makeRelatedInformation(value.reason.node, 'Function is declared here.')
+    ];
   }
 
   visitInvalidExpressionType(value: DynamicValue): ts.DiagnosticRelatedInformation[] {
--- a/packages/compiler-cli/src/ngtsc/partial_evaluator/src/diagnostics.ts
+++ b/packages/compiler-cli/src/ngtsc/partial_evaluator/src/diagnostics.ts
@@ -1,0 +1,176 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import * as ts from 'typescript';
+
+import {makeRelatedInformation} from '../../diagnostics';
+import {Reference} from '../../imports';
+import {DynamicValue, DynamicValueVisitor} from './dynamic';
+import {EnumValue, KnownFn, ResolvedModule, ResolvedValue} from './result';
+
+/**
+ * Derives a type representation from a resolved value to be reported in a diagnostic.
+ *
+ * @param value The resolved value for which a type representation should be derived.
+ * @param maxDepth The maximum nesting depth of objects and arrays, defaults to 1 level.
+ */
+export function describeResolvedType(value: ResolvedValue, maxDepth: number = 1): string {
+  if (value === null) {
+    return 'null';
+  } else if (value === undefined) {
+    return 'undefined';
+  } else if (typeof value === 'number' || typeof value === 'boolean' || typeof value === 'string') {
+    return typeof value;
+  } else if (value instanceof Map) {
+    if (maxDepth === 0) {
+      return 'object';
+    }
+    const entries = Array.from(value.entries()).map(([key, v]) => {
+      return `${quoteKey(key)}: ${describeResolvedType(v, maxDepth - 1)}`;
+    });
+    return entries.length > 0 ? `{ ${entries.join('; ')} }` : '{}';
+  } else if (value instanceof ResolvedModule) {
+    return '(module)';
+  } else if (value instanceof EnumValue) {
+    return value.enumRef.debugName ?? '(anonymous)';
+  } else if (value instanceof Reference) {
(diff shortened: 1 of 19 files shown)


packages/compiler-cli/src/ngtsc/partial_evaluator/src/diagnostics.ts
```diff
@@ -0,0 +1,187 @@
...
+    return 'module';
+  } else if (value instanceof EnumValue) {
+    return value.enumRef.debugName ?? 'anonymous';
+  } else if (value instanceof Reference) {
+    return value.debugName ?? 'anonymous';
+  } else if (Array.isArray(value)) {
+    if (maxDepth === 0) {
+      return 'Array';
+    }
+    return `[${value.map(v => describeResolvedType(v, maxDepth - 1)).join(', ')}]`;
```
JoostK, line 48: I choose to display arrays as a tuple representation (e.g. `[string, string, number]`), but we may as well choose to show them as array with a union type (e.g. `(string | number)[]`). Alternatively, we could limit the number of entries to include.

packages/compiler-cli/src/ngtsc/partial_evaluator/src/diagnostics.ts
```diff
@@ -0,0 +1,176 @@
...
+import {DynamicValue, DynamicValueVisitor} from './dynamic';
+import {EnumValue, KnownFn, ResolvedModule, ResolvedValue} from './result';
+
+/**
+ * Derives a type representation from a resolved value to be reported in a diagnostic.
+ *
+ * @param value The resolved value for which a type representation should be derived.
+ * @param maxDepth The maximum nesting depth of objects and arrays, defaults to 1 level.
+ */
+export function describeResolvedType(value: ResolvedValue, maxDepth: number = 1): string {
```
alxhub, line 23: This function works great for now.

Longer term, we should think about a change in architecture for the partial evaluator that will allow leveraging TypeScript to describe types for us. If we could somehow retain the `ts.Node` of values we produce, we can get a `ts.Type` from it, and I believe TS has ways of stringifying the "effective" type of a node.

packages/compiler-cli/src/ngtsc/partial_evaluator/src/diagnostics.ts
```diff
@@ -0,0 +1,176 @@
...
+      return 'object';
+    }
+    const entries = Array.from(value.entries()).map(([key, v]) => {
+      return `${quoteKey(key)}: ${describeResolvedType(v, maxDepth - 1)}`;
+    });
+    return entries.length > 0 ? `{ ${entries.join('; ')} }` : '{}';
+  } else if (value instanceof ResolvedModule) {
+    return 'module';
+  } else if (value instanceof EnumValue) {
+    return value.enumRef.debugName ?? 'anonymous';
```
alxhub, line 41:
```suggestion
    return value.enumRef.debugName ?? '(anonymous)';
```

packages/compiler-cli/src/ngtsc/partial_evaluator/src/diagnostics.ts
```diff
@@ -0,0 +1,176 @@
...
+
+  visitUnsupportedSyntax(value: DynamicValue): ts.DiagnosticRelatedInformation[] {
+    return [makeRelatedInformation(value.node, 'This syntax is not supported.')];
+  }
+
+  /**
+   * Determines whether the dynamic value reported for the node should be traced, i.e. if it is not
+   * part of the container for which the most recent trace was created.
+   */
+  private shouldTrace(node: ts.Node): boolean {
```
alxhub, line 138: This is so clever!

alxhub (approved):
👍 really clean implementation - this is going to be awesome!

gkalpak (approved):
A couple of optional nits and a couple of commit message typos. Otherwise looks great 💯 
- 2nd commit: `has to ability to statically evaluate` --> `has the ability to statically evaluate` (?)
- 3rd commit: `a function calls` --> `a function call`

packages/compiler-cli/src/ngtsc/partial_evaluator/src/diagnostics.ts
```diff
@@ -0,0 +1,176 @@
...
+      return 'object';
+    }
+    const entries = Array.from(value.entries()).map(([key, v]) => {
+      return `${quoteKey(key)}: ${describeResolvedType(v, maxDepth - 1)}`;
+    });
+    return entries.length > 0 ? `{ ${entries.join('; ')} }` : '{}';
+  } else if (value instanceof ResolvedModule) {
+    return 'module';
+  } else if (value instanceof EnumValue) {
+    return value.enumRef.debugName ?? '(anonymous)';
```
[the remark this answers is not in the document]
JoostK: Enum values are essentially just `Reference`s, but refer to a particular member of that enum. As we're just printing the type here, only the information from the reference is used. As such, it's displayed as other `Reference`s would be. We _could_ include the kind of declaration for any `Reference`, e.g. show `class MyCmp` and `enum ViewEncapsulation`, but I kept the notation as close as TS's native representation of types.

What might make sense to do is to extend `createValueHasWrongTypeError` such that it creates a `ts.DiagnosticRelationInformation` pointing to the declaration of the enum, as it does for `Reference`s.

packages/compiler-cli/src/ngtsc/partial_evaluator/src/diagnostics.ts
```diff
@@ -0,0 +1,176 @@
...
+  } else if (value instanceof Reference) {
+    return value.debugName ?? '(anonymous)';
+  } else if (Array.isArray(value)) {
+    if (maxDepth === 0) {
+      return 'Array';
+    }
+    return `[${value.map(v => describeResolvedType(v, maxDepth - 1)).join(', ')}]`;
+  } else if (value instanceof DynamicValue) {
+    return 'dynamic';
```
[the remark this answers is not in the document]
JoostK: Good point, changed as proposed. I also went over all others and added parenthesis to some to indicate that these are not actual types.

alxhub:
[Presubmit](http://test/OCL:316927731:BASE:316927763:1592418273314:5ec71141)

alxhub:
[Take 2](http://test/OCL:316927731:BASE:317213212:1592525737968:e9aa6e04)

PR #37593, opened 2020-06-15
refactor(dev-infra): migrate github api in GitClient to rely on GithubClient

josephperrott:
GitClient now uses GithubClient for github API interactions.  GithubClient is
a class which extends Octokit and provides a member which allows for GraphQL
requests against the Github GraphQL api, as well as providing convenience methods
for common/repeated Github API requests.

### DIFF
--- a/dev-infra/utils/git.ts
+++ b/dev-infra/utils/git.ts
@@ -6,13 +6,21 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
+import {graphql} from '@octokit/graphql';
 import * as Octokit from '@octokit/rest';
 import {spawnSync, SpawnSyncOptions, SpawnSyncReturns} from 'child_process';
+import {query, types} from 'typed-graphqlify';
 
 import {getConfig, getRepoBaseDir, NgDevConfig} from './config';
 import {info, yellow} from './console';
 
-
+/**
+ * An object representation of a GraphQL Query to be used as a response type and to generate
+ * a GraphQL query string.
+ */
+type GraphQLQueryObject = Parameters<typeof query>[1];
+/** Parameters for a GraphQL request. */
+type RequestParameters = Parameters<typeof graphql>[1];
 /** Github response type extended to include the `x-oauth-scopes` headers presence. */
 type RateLimitResponseWithOAuthScopeHeader = Octokit.Response<Octokit.RateLimitGetResponse>&{
   headers: {'x-oauth-scopes': string};
@@ -55,12 +63,10 @@ export class GitClient {
       `https://${this._githubToken}@github.com/${this.remoteConfig.owner}/${
           this.remoteConfig.name}.git`;
   /** Instance of the authenticated Github octokit API. */
-  api: Octokit;
+  github: GithubClient;
 
   /** The file path of project's root directory. */
   private _projectRoot = getRepoBaseDir();
-  /** The OAuth scopes available for the provided Github token. */
-  private _oauthScopes: Promise<string[]>|null = null;
   /**
    * Regular expression that matches the provided Github token. Used for
    * sanitizing the token from Git child process output.
@@ -76,12 +82,7 @@ export class GitClient {
       this._githubTokenRegex = new RegExp(_githubToken, 'g');
     }
 
-    this.api = new Octokit({auth: _githubToken});
-    this.api.hook.error('request', error => {
-      // Wrap API errors in a known error class. This allows us to
-      // expect Github API errors better and in a non-ambiguous way.
-      throw new GithubApiRequestError(error.status, error.message);
-    });
+    this.github = new GithubClient(_githubToken);
   }
 
   /** Executes the given git command. Throws if the command fails. */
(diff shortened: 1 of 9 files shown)


dev-infra/utils/git.ts
```diff
@@ -186,22 +187,138 @@ export class GitClient {
...
 }
+
+
+/**
+ * A Github client for interacting with the Github APIs.
+ *
+ * Additionally, provides convienience methods for actions which require multiple requests, or
+ * would provide value from memoized style responses.
+ **/
+class GithubClient extends Octokit {
```
devversion, line 212: Can we have these in separate files? We should probably have `utils/git/`?

dev-infra/utils/git.ts
```diff
@@ -186,22 +187,138 @@ export class GitClient {
...
+  async query<T extends GraphQLQueryObject>(queryObject: T, params: RequestParameters = {}) {
+    const queryString = query(queryObject);
+    return (await this.graqhql(queryString, params)) as typeof queryObject;
+  }
+
+  /**
+   * Perform multiple queries using Github's GraphQL API to retrieve multiple pages of queried
+   * information.
+   **/
+  async multiPageQuery<T extends GraphQLQueryObject, O extends {} = T>(
```
devversion, line 264: Nit: can we have descriptions for `T` and `O`? might just use longer generic names for more context?

dev-infra/utils/git.ts
```diff
@@ -186,22 +187,138 @@ export class GitClient {
...
+      queryObject: T, params: RequestParameters,
+      processPage:
+          (result: T,
+           currentResult: O) => {nextCursor: string, mergedResult: O, checkNextPage: boolean}) {
+    /**
+     * The object to be returned after queries are completed, initially empty, to be merged into on
+     * each iteration through pages.
+     */
+    let output: O = <O>{};
```
devversion, line 273: Should we call this `previousResult` here and in the `processPage` function above? I think this also be `null` initially. I think assuming an empty literal is not always correct?

dev-infra/utils/git.ts
```diff
@@ -186,22 +187,138 @@ export class GitClient {
...
+    return (await this.graqhql(queryString, params)) as typeof queryObject;
+  }
+
+  /**
+   * Perform multiple queries using Github's GraphQL API to retrieve multiple pages of queried
+   * information.
+   **/
+  async multiPageQuery<T extends GraphQLQueryObject, O extends {} = T>(
+      queryObject: T, params: RequestParameters,
+      processPage:
```
devversion, line 266: Could you elaborate in what cases we'd need this? Can we make the API more strict and simple? i.e. just returning the new result? I'm thinking if someone does more advanced stuff with cursors and `hasNextPage`, then he is better implementing it on his own. I'd like:

```ts
// Note that this matches quite with the actual array reduce signature
graphql.paginateQuery(query, params, (current, previous) => [...current, previous], []);
```

dev-infra/utils/git.ts
```diff
@@ -186,22 +187,138 @@ export class GitClient {
...
+
+/**
+ * Simple memoize decorator to returning the cached result when available on each call of decorated
+ * methods.
+ *
+ * This is intentionally not exported as it is written for extremely simple use cases within the
+ * scope the GithubClient and GitClient classes.  It is limited to only memoizing methods with no
+ * parameters and makes assumptions related to this limitation.
+ */
+function memoize() {
```
devversion, line 300: I think in the current state, this is more overhead than benefit. If we plan to use it more often (right now we do just twice; where a memo property is fine too), then we should move it into a common util IMO?

dev-infra/utils/git.ts
```diff
@@ -186,22 +187,138 @@ export class GitClient {
...
+          `to be returned by the method on subsequent calls.`);
+    }
+    /** The symbol being used to store the result in between calls. */
+    let result: any;
+    /**
+     * The new method to be called, in place of the original, calling the original if no result
+     * value is defined.
+     */
+    descriptor.value = function() {
+      return result = result || originalMethod.apply(this, arguments);
```
devversion, line 320: This should most likely not use truthyness, but rather `undefined`, or ideally an unique `Symbol` so that `undefined` and falsy values are cached properly.

devversion (commented):
Yeah, I think this is the right way. I left a few comments.

dev-infra/utils/git.ts
josephperrott: Yeah, as I continued to test I found that we actually will run into issues with trying to use decorators because we just use the same tsconfig as the rest of the repo.

We will have to reconsider a decorator strategy if we want to got that route long term.

dev-infra/utils/git.ts
josephperrott: Removed until we have a need for it directly.

I think there is value in providing a utility for managing pagination in graphql, but we can address it when it comes up.

devversion: 👍

dev-infra/utils/git/_github.ts
```diff
@@ -0,0 +1,99 @@
...
+  }
+}
+
+/**
+ * An object representation of a GraphQL Query to be used as a response type and to generate
+ * a GraphQL query string.
+ */
+type GraphQLQueryObject = Parameters<typeof query>[1];
+/** Parameters for a GraphQL request. */
+type RequestParameters = Parameters<typeof graphql>[1];
```
devversion, line 75: Optional I think you could import explicit types from `@octokit/types`. Unfortunate though that it's not directly exported. The types of octokit are not very intuitive..

dev-infra/utils/git/_github.ts
```diff
@@ -0,0 +1,99 @@
...
+export class _GithubClient extends Octokit {
+  /** The Github GraphQL (v4) API. */
+  graqhql: GithubGraphqlClient;
+
+  /** The current user based on checking against the Github API. */
+  private _currentUser: string|null = null;
+
+  constructor(token?: string) {
+    // Pass in authentication token to base Octokit class.
+    super({token});
```
devversion, line 42: Previously this was passed to Octkit through `auth`? how does that work?

dev-infra/utils/git/index.ts
```diff
@@ -55,7 +51,7 @@ export class GitClient {
       `https://${this._githubToken}@github.com/${this.remoteConfig.owner}/${
           this.remoteConfig.name}.git`;
   /** Instance of the authenticated Github octokit API. */
-  api: Octokit;
+  github: _GithubClient;
```
devversion, line 54: We can directly create it with `github = new _GithubClient(this._githubToken);`

devversion (approved):

dev-infra/utils/git/_github.ts
josephperrott: Whoops

PR #37594, opened 2020-06-15
refactor(dev-infra): cleanup shared package dependencies

devversion:
* See individual commits

### DIFF
--- a/dev-infra/tmpl-package.json
+++ b/dev-infra/tmpl-package.json
@@ -9,7 +9,9 @@
     "ts-circular-deps": "./ts-circular-dependencies/index.js"
   },
   "dependencies": {
+    "@angular/benchpress": "0.2.0",
     "@octokit/graphql": "<from-root>",
+    "brotli": "<from-root>",
     "chalk": "<from-root>",
     "cli-progress": "<from-root>",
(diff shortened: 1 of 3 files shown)


dev-infra/tmpl-package.json
```diff
@@ -21,34 +22,30 @@
...
-    "rollup-plugin-sourcemaps": "<from-root>",
     "selenium-webdriver": "<from-root>",
     "ts-node": "<from-root>",
-    "tslib": "<from-root>",
-    "typescript": "<from-root>",
-    "zone.js": "^0.10.3"
+    "typescript": "<from-root>"
   },
   "peerDependenciesMeta": {
+    "@bazel/buildifier": {
```
josephperrott, line 37: @bazel/buildifier has been removed from `peerDependencies`? It should still be present correct?

josephperrott (changes requested):

dev-infra/tmpl-package.json
devversion: yeah, that was by mistake. Just wiped all `@bazel/*` dependencies.

josephperrott (approved):
LGTM

PR #37595, opened 2020-06-15
feat(dev-infra): support for caretaker note label in merge script

devversion:
Adds support for a caretaker note label to the merge script.
Whenever a configured label is applied, the merge script will
not merge automatically, but instead prompt first in order
to ensure that the caretaker paid attention to the manual
caretaker note on the PR. This helps if a PR needs special
attention.

### DIFF
--- a/dev-infra/pr/merge/index.ts
+++ b/dev-infra/pr/merge/index.ts
@@ -114,8 +114,8 @@ export async function mergePullRequest(
         error(red('An error related to interacting with Github has been discovered.'));
         error(failure!.message);
         return false;
-      case MergeStatus.ABORTED:
-        info(`Merge of pull request aborted: #${prNumber}`);
+      case MergeStatus.USER_ABORTED:
+        info(`Merge of pull request has been aborted manually: #${prNumber}`);
         return true;
       case MergeStatus.FAILED:
         error(yellow(`Could not merge the specified pull request.`));
--- a/dev-infra/pr/merge/index.ts
+++ b/dev-infra/pr/merge/index.ts
@@ -90,7 +90,7 @@ export async function mergePullRequest(
   /**
    * Handles the merge result by printing console messages, exiting the process
    * based on the result, or by restarting the merge if force mode has been enabled.
-   * @returns Whether the merge was successful or not.
+   * @returns Whether the merge completed without errors or not.
    */
   async function handleMergeResult(result: MergeResult, disableForceMergePrompt = false) {
     const {failure, status} = result;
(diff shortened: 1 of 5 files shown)


dev-infra/pr/merge/task.ts
```diff
@@ -72,6 +75,14 @@ export class PullRequestMergeTask {
       return {status: MergeStatus.FAILED, failure: pullRequest};
     }
 
+    // If the pull request has a caretaker note applied, raise awareness by prompting
+    // the caretaker. The caretaker can then decide to proceed or abort the merge.
+    if (pullRequest.hasCaretakerNote &&
```
devversion, line 80: Note: I considered extracting/printing the caretaker note automatically, but that is most likely something larger we need to investigate/consider. Created DEV-119 for that. Should be low-priority P5.

dev-infra/pr/merge/pull-request.ts
```diff
@@ -77,13 +81,17 @@ export async function loadAndValidatePullRequest(
       config.requiredBaseCommits && config.requiredBaseCommits[githubTargetBranch];
   const needsCommitMessageFixup = !!config.commitMessageFixupLabel &&
       labels.some(name => matchesPattern(name, config.commitMessageFixupLabel));
+  const hasCaretakerNote = !!config.caretakerNoteLabel &&
+      labels.some(name => matchesPattern(name, config.caretakerNoteLabel!));
```
josephperrott, line 85: Do we need to have the non-null assertion(`!`) since the `!!config.caretakerNoteLabel` should already check non-null?

dev-infra/pr/merge/task.ts
josephperrott: Agreed.

dev-infra/pr/merge/index.ts
```diff
@@ -114,6 +114,9 @@ export async function mergePullRequest(
         error(red('An error related to interacting with Github has been discovered.'));
         error(failure!.message);
         return false;
+      case MergeStatus.ABORTED:
```
josephperrott, line 117: Is the intention that this status is used for when the user causes the merge to be aborted?  If so I think it might be helpful to note that in the logged message as well.

josephperrott (approved):
LGMT, just a couple nits

dev-infra/pr/merge/pull-request.ts
devversion: Yeah, it's required (was surprised too). This is because we use use it inside of a callback.

dev-infra/pr/merge/index.ts
devversion: Yeah, nice suggestion. How about `USER_ABORTED`?

josephperrott: Yep, that works for me.

dev-infra/pr/merge/pull-request.ts
devversion: In case you are wondering: It works above with commit message fixup as that one is a required option and not optional in the merge config interface.

dev-infra/pr/merge/index.ts
devversion: done!

PR #37599, opened 2020-06-15
fix(compiler-cli): NgTscPlugin should clean up after program creation

alxhub:
`NgTscPlugin` wraps an incoming `ts.CompilerHost` in an `NgCompilerHost`
which patches all `ts.SourceFile`s in the program with referenced shims. In
order to not emit triple-slash `<reference>` directives that refer to shim
files, these patches have to be cleaned up by calling
`NgCompilerHost.postProgramCreationCleanup()` before emit. Previously, the
`NgTscPlugin` was not doing this cleanup.

This commit fixes the bug (which is a one-line change), plus adds a
mechanism in the `NgtscTestEnvironment` to run compilation via the plugin,
allowing for a test to be written.

### DIFF
--- a/packages/compiler-cli/test/ngtsc/env.ts
+++ b/packages/compiler-cli/test/ngtsc/env.ts
@@ -6,19 +6,20 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {CustomTransformers, defaultGatherDiagnostics, Program} from '@angular/compiler-cli';
+import {CustomTransformers, defaultGatherDiagnostics, NgTscPlugin, Program, readConfiguration} from '@angular/compiler-cli';
 import * as api from '@angular/compiler-cli/src/transformers/api';
 import * as ts from 'typescript';
 
 import {createCompilerHost, createProgram} from '../../index';
 import {main, mainDiagnosticsForTest, readNgcCommandLineAndConfiguration} from '../../src/main';
-import {absoluteFrom, AbsoluteFsPath, FileSystem, getFileSystem, NgtscCompilerHost} from '../../src/ngtsc/file_system';
+import {absoluteFrom, AbsoluteFsPath, FileSystem, getFileSystem, NgtscCompilerHost, setFileSystem} from '../../src/ngtsc/file_system';
 import {Folder, MockFileSystem} from '../../src/ngtsc/file_system/testing';
 import {IndexedComponent} from '../../src/ngtsc/indexer';
 import {NgtscProgram} from '../../src/ngtsc/program';
 import {LazyRoute} from '../../src/ngtsc/routing';
 import {setWrapHostForTest} from '../../src/transformers/compiler_host';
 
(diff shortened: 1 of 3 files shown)


alexeagle (approved):
Awesome, thanks a bunch for fitting this into the v10 release! It's nice to see the fix can be made without anything Bazel-specific like changing rules_typescript.

gregmagolan (approved):
Nice! Thanks @alxhub

alexeagle:
FYI I patched that line into rules_nodejs to confirm that it's green downstream https://github.com/bazelbuild/rules_nodejs/pull/1932

packages/compiler-cli/test/ngtsc/plugin_spec.ts
```diff
@@ -0,0 +1,43 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {getFileSystem} from '@angular/compiler-cli/src/ngtsc/file_system';
```
JoostK, line 9:
```suggestion
import {getFileSystem} from '../../src/ngtsc/file_system';
```

packages/compiler-cli/test/ngtsc/env.ts
```diff
@@ -6,19 +6,21 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {CustomTransformers, defaultGatherDiagnostics, Program} from '@angular/compiler-cli';
+import {CustomTransformers, defaultGatherDiagnostics, Program, readConfiguration} from '@angular/compiler-cli';
+import {NgTscPlugin} from '@angular/compiler-cli/src/ngtsc/tsc_plugin';
```
JoostK, line 10:
```suggestion
import {NgTscPlugin} from '../../src/ngtsc/tsc_plugin';
```

and needs to be moved down probably.

JoostK (approved):
LGTM with comments

alxhub:
[Presubmit](https://test.corp.google.com/ui#id=OCL:316732356:BASE:316732389:1592335052259:4a1a5a15)

JoostK (approved):

alexeagle:
Agh, this wasn't included in Angular 10? So anyone using Bazel will have problems upgrading. I wonder if any such issues were already reported

alexeagle:
oh it looks like maybe the fix went in https://github.com/angular/angular/pull/37690 already so this PR just adds a specific test case for the regression we saw in ts_library

JoostK:
> oh it looks like maybe the fix went in #37690 already so this PR just adds a specific test case for the regression we saw in ts_library

@alexeagle Yes, that's correct. 10.0.0 was still broken, but 10.0.1 fixed the issue in #37690.

alexeagle:
Cool we upgraded the Bazel example and removed the patch, all good on our end.

Issue #10010, opened 2016-07-12
bug(forms) Initial select option set to null not working on IE/Edge


patrickracicot:
**I'm submitting a ...**

```
[x ] bug report
[ ] feature request
[ ] support request => Please do not submit support request here, instead see https://github.com/angular/angular/blob/master/CONTRIBUTING.md#question
```

**Current behavior**
When setting the initial value of a `<select>` control to either null or undefined, the rendered `<select>` element displays the first `<option>`.

**Expected/desired behavior**
If the value of the Control is set to null, undefined or to a value not found in the list of options, the rendered element should displays an empty select.

**Reproduction of the problem**
Here is a plnkr: 
[http://plnkr.co/edit/7aHyWPt41YdvQH09MH8x?p=preview](http://plnkr.co/edit/7aHyWPt41YdvQH09MH8x?p=preview)

**What is the expected behavior?**
The current behavior found on Chrome and Firefox are the expected behavior.

**What is the motivation / use case for changing the behavior?**
The `<select>` element that is rendered should not display a selected option if the model backing it is not set to the value of that option.

**Please tell us about your environment:**
- **Angular version:** 2.0.0-rc.4
- **Browser:** IE 11 & Edge 25
- **Language:** TypeScript 1.8.10

gstgusts:
We have the same issue in IE
![ie11_problem](https://cloud.githubusercontent.com/assets/19387973/17104673/90a31004-528c-11e6-85f6-9afd9720b9d5.png)

MrHOY:
I have same issue too... I set ngModel is null and it work good on chrome
![chrome](https://cloud.githubusercontent.com/assets/20653706/17124560/3551cfbc-5316-11e6-92d7-462839d2800f.jpg)
but on IE/Edge always selected the first value so I can't handling event ngModelChange if user not change new value. 
![ie](https://cloud.githubusercontent.com/assets/20653706/17124623/ab5d631a-5316-11e6-8e55-5cd08e634bb7.jpg)

gnujeremie:
I have the same issue. The current fix woud be to add an empty option as the first options list of the select tag.

patrickracicot:
@gnujeremie Unfortunately this option is not always viable.  In a case where you don't want to permit a user the selection of an empty value once he has actually chosen one in the first place.

gnujeremie:
@patrickracicot then let's hope it will be fixed in RC6 !

MartinMa:
[...]

PR #23784, opened 2018-05-09
fix(forms): improve select performance and ensure that selected option matches the ngModel value

theodorejb:
Fixes #41330, fixes #14505, fixes #18430.

## PR Checklist
- [x] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?
- [x] Bugfix

## What is the current behavior?
Currently Angular calls the `writeValue()` over and over for every individual option element that is added or removed from the DOM. This results in exponentially more `_compareValue` calls than the number of option elements (issue #41330).

Furthermore, calling `writeValue()` when rendering individual option elements causes an issue in Safari and IE 11 where the first option element fails to be deselected when no option matches the select ngModel. This is because Angular sets the select element's value property before appending the option's child text node to the DOM (issue #14505).

Finally, the current behavior causes an issue with delayed element removal when using the animations module (in all browsers). When a selected option is removed (so no option matches the ngModel anymore), Angular changes the select element value before actually removing the option from the DOM. Then when the option is finally removed from the DOM, the browser changes the select value to that of the first option, even though it doesn't match the ngModel (issue #18430).

## What is the new behavior?
Rather than calling `writeValue()` over and over for individual option elements that are added/removed from the DOM, call it once after all the options have finished rendering via the `afterNextRender` function. This dramatically improves performance and results in correct behavior across all supported browsers.

## Does this PR introduce a breaking change?
- [ ] Yes
- [x] No

PR #37614, opened 2020-06-16
refactor(forms): remove dead code

theodorejb:
This functionally dead code was originally introduced via pull request #13903, apparently in a blind attempt to fix issue #10010. But no tests were added to verify the fix, and the many comments on that issue after it was closed indicate that it wasn't actually resolved.

In fact, setting `selectedIndex` does absolutely nothing here, since the selected index is immediately overridden by setting the `value` property. A working fix (with tests) for the Safari/IE behavior is in pull request #23784. Originally this dead code was removed as part of that PR, but @AndrewKushnir recommended creating a separate PR for the cleanup.

## PR Type

- [x] No

### DIFF
--- a/packages/forms/src/directives/select_control_value_accessor.ts
+++ b/packages/forms/src/directives/select_control_value_accessor.ts
@@ -115,16 +115,12 @@ export class SelectControlValueAccessor extends BuiltInControlValueAccessor impl
   private _compareWith: (o1: any, o2: any) => boolean = Object.is;
 
   /**
-   * Sets the "value" property on the input element. The "selectedIndex"
-   * property is also set if an ID is provided on the option element.
+   * Sets the "value" property on the select element.
    * @nodoc
    */
   writeValue(value: any): void {
     this.value = value;
     const id: string|null = this._getOptionId(value);
-    if (id == null) {
-      this.setProperty('selectedIndex', -1);
-    }
     const valueString = _buildValueString(id, value);
     this.setProperty('value', valueString);
   }


AndrewKushnir (commented):
Thanks for creating this PR @theodorejb 👍 

I think we should look into that as the next step after #23784 lands (I will put the "blocked" label for now to indicate that).

Thank you.

theodorejb:
@AndrewKushnir It would be a lot easier if this could land before the other PR, since then it wouldn't be necessary to add a separate method there for updating the DOM. This change should be easy to land in master and patch, since it's quite impossible for it to break anything.

AndrewKushnir:
Hi @theodorejb, I've received your reply via email notification, but it doesn't show up here and I'd like to provide some additional thoughts/comments.

The refactoring that you proposed looks reasonable, but I think we should land more important change (bug fix) in #23784 first to avoid any risks associated with specific/unusual browser-related behavior. It would also be great to find when that code was originally introduced and what was the goal of adding it (for ex. if it was added to handle some use-cases or to accommodate for some browser-specific behavior).

Thank you.

theodorejb:
@AndrewKushnir I deleted my reply because after additional investigation in the other PR, I realized that a separate method for updating the DOM doesn't make sense anyway, so it doesn't matter to me if this change lands before or after.

The code in question was originally introduced via PR #13903 in an apparent blind attempt to fix the same issue that my PR fixes. But no tests were added to verify the fix, and the many comments in issue #10010 after it was closed indicate that it didn't have the desired effect (as I said before, it's impossible for this code to work, since the selected index is immediately overridden).

Since the code was added in attempt to fix the same issue as PR #23784, maybe it makes sense to combine this change with that PR as it was originally?

Thanks!

AndrewKushnir:
@theodorejb, thanks a lot for the investigation and identifying when the code was added and why 👍

I think we should still keep this PR separate and just update the commit message (and PR description) with the information that you mentioned in the comment.

Once #23784 lands, let's rebase this PR and I will run tests in Google's codebase.

Thank you.

AndrewKushnir (approved):
LGTM, thanks @theodorejb. I've added "cleanup" label to indicate that the commit message should be updated to include additional information. Thank you.

theodorejb:
@AndrewKushnir I updated the commit message and PR description now.

theodorejb:
@AndrewKushnir Is there any reason this can't be merged? This is dead code that just makes it harder to understand the behavior of select_control_value_accessor.

AndrewKushnir:
@theodorejb I plan to revisit this change once there is a clear path forward for #23784. Removing this code should be relatively safe, but I'd prefer to avoid any potential browser quirks. We can probably add a quick comment into the code for now (and refer to PR #23784) to improve readability.

dylhunn:
Now that IE is gone, the original purpose for introducing this (broken) code no longer applies, so I think we should just go ahead and merge this?

dylhunn (approved):
reviewed-for: fw-forms

dylhunn:
Running a presubmit: [link](http://test/OCL:424693222:BASE:424693265:1643318790097:d40b7e31)

dylhunn:
Presubmit is green. Since we no longer support IE at all, we don't have to worry about the browser compat issues Andrew mentioned last year, and we can merge this. Hooray for free bundle size improvements 🎉

caretaker: CI is rejecting the commit message erroneously, but the format appears to be correct.

AndrewKushnir:
FYI I've rebased this PR to re-run CI since last rebase was 8 months ago.

dylhunn:
> FYI I've rebased this PR to re-run CI since last rebase was 8 months ago.

Thanks Andrew! It still merged cleanly, hence why I didn't rebase before running the presubmit.

jessicajaniuk:
This PR was merged into the repository by commit f1a51457f9193209db9e43be0782dc95128f4836.

PR #37620, opened 2020-06-17
fix(forms): correct usage of `selectedOptions`

### LINKED ISSUE #37433: select_multiple_control_value_accessor uses IE degradation 100% of the time
# 🐞 bug report

### Affected Package
https://github.com/angular/angular/blob/master/packages/forms/src/directives/select_multiple_control_value_accessor.ts

### Is this a regression?

No, it's always been broken

### Description

Minor. Functionality works fine, just a mistake causing unoptimized JS

## 🔬 Minimal Reproduction

Any `<select>` will reproduce it

## 🔥 Exception or Error
```
  registerOnChange(fn: (value: any) => any): void {
    this.onChange = (_: any) => {
      const selected: Array<any> = [];
      if (_.hasOwnProperty('selectedOptions')) {
        const options: HTMLCollection = _.selectedOptions;
        for (let i = 0; i < options.length; i++) {
          const opt: any = options.item(i);
          const val: any = this._getOptionValue(opt.value);
          selected.push(val);
        }
      }
      // Degrade on IE
      else {
        const options: HTMLCollection = <HTMLCollection>_.options;
        for (let i = 0; i < options.length; i++) {
          const opt: HTMLOption = options.item(i);
          if (opt.selected) {
            const val: any = this._getOptionValue(opt.value);
            selected.push(val);
          }
        }
      }
      this.value = selected;
      fn(selected);
    };
  }
```

Notice the check for newer browers uses:
```
_.hasOwnProperty('selectedOptions')
```
This doesn't work, because it's an inherited property.
Alternatives are:
```
_.constructor.prototype.hasOwnProperty('selectedOptions')
_.selectedOptions !== undefined
'selectedOptions' in _
```

## 🌍  Your Environment

**Angular Version:**
<pre><code>
Angular CLI: 9.1.4
Node: 12.13.1
OS: win32 x64

Angular: 9.1.4
... animations, cli, common, compiler, compiler-cli, 
core, forms
Ivy Workspace: Yes

@angular-devkit/architect         0.901.4
@angular-devkit/build-angular     0.901.4
@angular-devkit/build-optimizer   0.901.4
@angular-devkit/build-webpack     0.901.4
@angular-devkit/core              9.1.4
@angular-devkit/schematics        9.1.4
@ngtools/webpack                  9.1.4
@schematics/angular               9.1.4
@schematics/update                0.901.4
</code></pre>

**Anything else relevant?**

sonukapoor:
Previously, `registerOnChange` used `hasOwnProperty` to identify if the
property is supported. However, this does not work as the `selectedOptions`
property is an inherited property. This commit fixes this by verifying
the property on the prototype instead.

Closes #37433

## PR Checklist

## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #37433

- [ ] Yes
- [X] No

### DIFF
--- a/packages/forms/src/directives/select_multiple_control_value_accessor.ts
+++ b/packages/forms/src/directives/select_multiple_control_value_accessor.ts
@@ -92,8 +92,6 @@ export class SelectMultipleControlValueAccessor implements ControlValueAccessor 
   _optionMap: Map<string, ɵNgSelectMultipleOption> = new Map<string, ɵNgSelectMultipleOption>();
   /** @internal */
   _idCounter: number = 0;
-  /** @internal */
-  _hasNativeSelectedOptionsSupport = HTMLSelectElement.prototype.hasOwnProperty('selectedOptions');
 
   /**
    * @description
@@ -158,7 +156,7 @@ export class SelectMultipleControlValueAccessor implements ControlValueAccessor 
   registerOnChange(fn: (value: any) => any): void {
     this.onChange = (_: any) => {
       const selected: Array<any> = [];
-      if (this._hasNativeSelectedOptionsSupport) {
+      if (_.selectedOptions !== undefined) {
         const options: HTMLCollection = _.selectedOptions;
         for (let i = 0; i < options.length; i++) {
           const opt: any = options.item(i);
--- a/packages/forms/src/directives/select_multiple_control_value_accessor.ts
+++ b/packages/forms/src/directives/select_multiple_control_value_accessor.ts
@@ -92,6 +92,8 @@ export class SelectMultipleControlValueAccessor implements ControlValueAccessor 
   _optionMap: Map<string, ɵNgSelectMultipleOption> = new Map<string, ɵNgSelectMultipleOption>();
   /** @internal */
   _idCounter: number = 0;
+  /** @internal */
+  _hasNativeSelectedOptionsSupport = HTMLSelectElement.prototype.hasOwnProperty('selectedOptions');
 
   /**
    * @description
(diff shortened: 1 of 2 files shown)


packages/forms/src/directives/select_multiple_control_value_accessor.ts
```diff
@@ -92,6 +92,8 @@ export class SelectMultipleControlValueAccessor implements ControlValueAccessor
   _optionMap: Map<string, ɵNgSelectMultipleOption> = new Map<string, ɵNgSelectMultipleOption>();
   /** @internal */
   _idCounter: number = 0;
+  /** @internal */
+  _hasNativeSelectedOptionsSupport = HTMLSelectElement.prototype.hasOwnProperty('selectedOptions');
```
alfaproject, line 96: Is this `node.js`/SSR friendly?

CaerusKaru: An SSR-friendly alternative is to inject `DOCUMENT` into the class, and then use the following:
`this.DOCUMENT.defaultView.HTMLSelectElement.prototype.hasOwnProperty('selectedOptions');`

sonukapoor: @CaerusKaru Would something like this work from the SSR perspective?

```
const isBrowser = Node !== 'undefined' && Node !== null;
const hasNativeSelectedOptionsSupport = isBrowser && HTMLSelectElement.prototype.hasOwnProperty('selectedOptions');
```

CaerusKaru: I mean it would work in a sense, but I wouldn't advise it. You never know when a user has patched the global scope, which would make you think the global is there, but it's not.

sonukapoor: Got it. I will go with a safer approach then. Thanks.

AndrewKushnir (approved):
LGTM, thanks for fixing this @sonukapoor 👍 

Also huge thanks to @alfaproject and @CaerusKaru for providing very useful feedback!

AndrewKushnir:
[Presubmit](http://test/OCL:316992493:BASE:316992554:1592437759135:b8f74900)

AndrewKushnir:
FYI, presubmit went well, marking this PR as ready for merge. Thank you.

PR #37623, opened 2020-06-17
refactor: cleanup esm5 output flavor in bazel and fix rollup optimization pipeline

devversion:
* See individual commits

In follow-ups we can clean up things further by:

1. Using `@bazel/rollup` in `ng_package`. Currently uses the same legacy/non-idiomatic construct as with `ng_rollup_bundle` previously.
2. Converge zone.js `rollup_bundle` proxy with the one from `@angular/bazel`. Once done, consider merging that rollup macro with the one from `dev-infra`. That would require a flexible API, but might be worth long-term.
3. `angular/components` could use the rollup bundle macro for their integration size-checks.


packages/language-service/bundles/index.bzl
```diff
@@ -0,0 +1,25 @@
...
+        A variant of ng_rollup_bundle for the language-service bundle that
+        outputs in AMD format.
+    """
+    visibility = kwargs.pop("visibility", None)
+
+    # Note: the output file is called "umd.js" because of historical reasons.
+    # The format is actually AMD and not UMD, but we are afraid to rename
+    # the file because that would likely break the IDE and other integrations that
+    # have the path hardcoded in them.
+    ng_rollup_bundle(
```
devversion, line 14: FWIW: This also now means that the language service bundle is using ES2015. This shouldn't be a problem as our minimum supported NodeJS version for Angular, supports ES2015. That could also speed up the language service.

The language-service package does also not follow APF and this is even an UMD file technically. I think we should be good with that. There is also a task of switching devmode output to ES2015. If that happens, the compiler-cli would also be es2015. It's also not best that the language service relies on dev-infra/benchmark rule for building bundles. This isn't new in this PR though, so I'll keep this in a follow-up as mentioned in the PR description.

josephperrott (approved):
Looks good to me from a general prospective, but definitely needs to have review from a language service perspective.

ayazhafiz (commented):
Keen is more familiar with this than I

packages/language-service/bundles/index.bzl
kyliau: We should just go ahead and rename this to amd. This package is only consumed by `@angular/language-server`. Other IDEs consumer the latter package, so they won't see this change.
Switching to ES2015 is totally fine. In fact, it's great, because it'll improve performance, like you mentioned.
I'm happy to do this in a follow-up PR to reduce the diff in this PR.

kyliau (approved):

packages/language-service/bundles/index.bzl
devversion: Sounds good. Thanks for the details @kyliau. I'll create a JIRA for following up on this

devversion: Created: FW-2240.

alxhub (approved):
LGTM for changes to language service and `ng_module.bzl`.

IgorMinar:
We'll need to update ` packages/bazel/src/esm5.bzl` patch in google3 (that file has a local mod). Can @alxhub or @kyliau help creating the patch? Hopefully the google3 usage won't affect our ability to simplify these rules on GitHub.

packages/bazel/test/ng_package/example_package.golden
```diff
@@ -180,11 +180,11 @@ Hello
     var A11yModule = /** @class */ (function () {
         function A11yModule() {
         }
-        A11yModule.decorators = [
-            { type: core.NgModule, args: [{},] }
-        ];
         return A11yModule;
     }());
+    A11yModule.decorators = [
```
IgorMinar, line 185: This diff is surprising. Do we know what caused it?

packages/bazel/test/ng_package/example_package.golden
```diff
@@ -180,11 +180,11 @@ Hello
     var A11yModule = /** @class */ (function () {
```
IgorMinar, line 180: This emit shape (`class` wrapped in IFFE) seems stale TS 3.9.5 no longer emits that shape. Is this PR rebased on master?

packages/bazel/src/ng_package/rollup.config.js
```diff
@@ -144,6 +146,27 @@ if (bannerFile) {
...
 }
 
+const downlevelToEs5Plugin = {
+  name: 'downlevel-to-es5',
+  transform: (code, filePath) => {
+    const compilerOptions = {
+      target: ts.ScriptTarget.ES5,
+      module: ts.ModuleKind.ES2015,
+      allowJs: true,
+      sourceMap: true,
```
IgorMinar, line 158: I don't believe that we can do it in this way and keep sourcemap working. As far as I know, typescript doesn't support *input* source maps, which means that while it emits source maps, it can't accept previously generated sourcemaps and use those as the input for generating the output source maps.

This means that typescript will generate incomplete/broken source maps that link to the input javascript code and not to the original/upstream typescript source code.

For this reason, I recommend using babel here instead of typescript for es2015 to es5 down leveling - just as Angular CLI does.

packages/core/test/bundling/cyclic_import/bundle.golden_symbols.json
```diff
@@ -1,34 +1,10 @@
 [
-  {
```
IgorMinar, line 2: nice!!!

IgorMinar (approved):
Overall, this is a great cleanup!! Thank you @devversion.

Five cleanup requests:
- [ ] ~please switch to using babel for es5 downleveling~
- [x] please measure the build time savings due to this change (use `blaze clean` before invoking `./scripts/build/build-packages-dist.js` both before and after this change)
- [x] rebrand this as a dev-infra change and highlight it in the dev-infra changelog (with the build latency decrease as measured above)
- [x] let's rebase this PR on master and confirm that we know why the ng_package golden files are changing (if they still are after the rebase). I currently don't see a reason for a change there.
- [x] lastly please work with a googler to update the google3 patch.

IgorMinar:
@devversion as a followup PR, could you please look at https://github.com/angular/angular/pull/36425 and resurrect that test (as a bazel test) and verify that the source maps are properly pointing all the way to the TS sources for all of our output formats (esm2015, fesm2015, umd) for one or two representative entrypoints (e.g. `@angular/core` and optionally `@angular/common/http`). Thanks.

packages/bazel/src/ng_package/rollup.config.js
devversion: That makes sense. We had the same issue in the past with the legacy package tools in Angular Material, but I thought this was solved as the approach currently used, matches with what `ng-packagr` does. I agree that using babel for input source map processing makes sense. We probably should do the same for `ng-packagr` too.

packages/bazel/test/ng_package/example_package.golden
devversion: Yeah, it's definitely rebased. This is actually the ES5 UMD output, so the IIFE is expected here.

packages/bazel/test/ng_package/example_package.golden
devversion: Yeah, I wondered too. The reason is that we no longer directly compile to ES5, but downlevel ES2015 code. Since in ES2015 output, the IIFE is not used, the static properties are not moved back into the IIFE.

If we'd directly compile to ES5 (like we did before), the static properties would be wrapped in the IIFE by TypeScript. Do you think that could cause any issues? Tests seem to show that it works for far.

packages/bazel/src/ng_package/rollup.config.js
devversion: Looked more into it, and experimented with Babel. It looks like `rollup` does not need transformers to respect input source maps as it composes a collapsed one at the end of all transformations. i.e. the source maps are correct when `tsc` is used for downleveling.

Same would work for babel, but then we'd need to opt out of `inputSourceMaps` as rollup expects source maps to capture the isolated transformation (es2015 -> es5), not TS to es5.

Also Babel is less convenient as it requires adding a new dependency, plus we'd have another _large_ set of helpers in the UMD files (besides the tslib ones). I think we should be good using TSC, similar to ng-packagr. https://github.com/rollup/rollup/pull/175

devversion:
@IgorMinar Yup. Sounds like a good follow-up to have. I created a JIRA for it, and will be looking into it. Thanks for the review so far!

devversion:
Added an entry to the dev-infra changed as requested. Also added information about the measured build time reduction to the corresponding commit.

josephperrott:
Caretaker Note:  This change will need to be landed alongside cl/317674053.  This changelist, accounts for the changes in this PR in our internal patch files run by copybara.

**Order of operations:**
- Merge cl/317674053
  _g3 presubmit will fail until next step completed_
- Merge pr#37623
  _Other g3 presubmits can resume again_

packages/bazel/test/ng_package/example_package.golden
IgorMinar: ah.. you are right! it's downleveled to a function, so that's ok. great. thanks

packages/bazel/src/ng_package/rollup.config.js
IgorMinar: Interesting! Good to know! thank you for looking into this and sharing the analysis.

IgorMinar:
based on g.co/ng/dev-infra-changelog this change improved build time for npm packages by ~27%. that's awesome!! thanks @devversion!

PR #37624, opened 2020-06-17
build: move shims_for_IE to third_party directory

### LINKED ISSUE #37615: Move code in shims_for_IE.js to third_party directory
Any vendor third_party code should all live in `third_party`. The shims_for_IE.js file contains vendor code that predates the `third_party` directory. This file is currently used for internal karma testing setup.

josephperrott:
The shims_for_IE.js file contains vendor code that predates the third_party
directory. This file is currently used for internal karma testing setup. This
change corrects this by moving the shims_for_IE file to //third_part/

Closes #37615


atscott (approved):

AndrewKushnir:
Hi @josephperrott, I just did a quick search in Google's codebase and found a few references. It looks like this file was excluded from the sync and since the location of the file has changed, it looks like it will be included. Could you please have a quick look at g3 and let me know if there is a patch CL we should land alongside with this change? I'm adding a "cleanup" label for now. Thank you.

josephperrott:
@AndrewKushnir confirmed we are good to land as is.  The references found in google3 are actually to the `shims_for_IE.js` file from AngularJS.  This file was not synced into google previously and its new location in `/third_party` also will not be synced in.

The `karma.conf.js` file that is synced into google that changed is unreferenced.

PR #37627, opened 2020-06-17
WIP: perf diagnostics

alxhub:


packages/compiler-cli/src/ngtsc/annotations/src/component.ts
```diff
@@ -953,3 +965,18 @@ export interface ParsedTemplate {
...
+
+function countTemplateNodes(nodes: TmplAstNode[]): number {
+  let count = nodes.length;
+  for (const node of nodes) {
+    if (node instanceof TmplAstElement || node instanceof TmplAstTemplate) {
+      count += countTemplateNodes(node.children);
+      count += node.attributes.length + node.inputs.length + node.outputs.length +
+          node.references.length;
+      if (node instanceof TmplAstTemplate) {
+        count + node.templateAttrs.length + node.variables.length;
```
diicar, line 977: it shouldn't be 
`count += node.templateAttrs.length + node.variables.length;`

PR #29909, opened 2019-04-15
docs: add example for multi provider

santoshyadavdev:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [x] Documentation content changes
- [x] angular.io application / infrastructure changes
- [ ] Other... Please describe:

## What is the current behavior?

Issue Number: #29860

## What is the new behavior?

Added example for multi provider.

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

PR #37645, opened 2020-06-19
docs(core): add note about not mutating multi provider arrays

### LINKED ISSUE #37481: Angular 9 - ngOnDestroy called on not OnDestroy instance when in multi provision
# 🐞 bug report

### Affected Package
probably `core`, 9.1.9

### Is this a regression?
since 8 - yes, since 9.0 - maybe.
Connected issue - https://github.com/angular/angular/issues/35231
As least I just migrated from 8.2.14

### Description
So, as in mentioned ticket 35231 the problem occurs when `ngOnDestroy` is called.
Recently the context was changed, now it seems to be called on every instance in `multi` provision. Even when some of them does not implement `OnDestroy` (at least it looks so, but as you later will see - it does not fit to logic, context is changed again).
Screenshot from chrome debugger when destroy was called (file `core.js`).
image
The problem is - `RouteResultExternalLinksManager` does not implement `OnDestroy`, and because of that exception occurs:
image
I would even say that `RouteResultExternalLinksManager` is not an instance of `Subscriber` as it looks like in debugger:
image
image

## 🔥 Exception or Error
<pre><code>
TypeError: this.clearSubscriptions is not a function
    at RouteResultExternalLinksManager.push.../../libs/imap/shared/flow-utils/subscriber.ts.Subscriber.ngOnDestroy (subscriber.ts:9)
    at executeOnDestroys (core.js:9522)
    at cleanUpView (core.js:9448)
    at destroyViewTree (core.js:9248)
    at destroyLView (core.js:9401)
    at RootViewRef.push.../../node_modules/@angular/core/__ivy_ngcc__/fesm5/core.js.ViewRef.destroy (core.js:10143)
    at ComponentRef.push.../../node_modules/@angular/core/__ivy_ngcc__/fesm5/core.js.ComponentRef.destroy (core.js:23229)
    at RouterOutlet.push.../../node_modules/@angular/router/__ivy_ngcc__/fesm5/router.js.RouterOutlet.deactivate (router.js:5355)
    at ActivateRoutes.push.../../node_modules/@angular/router/__ivy_ngcc__/fesm5/router.js.ActivateRoutes.deactivateRouteAndOutlet (router.js:2328)
    at ActivateRoutes.push.../../node_modules/@angular/router/__ivy_ngcc__/fesm5/router.js.ActivateRoutes.deactivateRouteAndItsChildren (router.js:2308)

</code></pre>

## 🌍  Your Environment

**Angular Version:**
<pre><code>

                |___/
    

Angular CLI: 9.1.7
Node: 12.13.1
OS: win32 x64

Angular: 9.1.9
Ivy Workspace: Yes

@angular-devkit/architect          0.901.7
@angular-devkit/build-angular      0.901.7
@angular-devkit/build-ng-packagr   0.901.7
@angular-devkit/build-optimizer    0.901.7
@angular-devkit/build-webpack      0.901.7
@angular-devkit/core               8.3.23
@angular-devkit/schematics         8.3.23
@angular/cdk                       9.2.4
@angular/cli                       9.1.7
@angular/material                  9.2.4
@ngtools/webpack                   9.1.7
@schematics/angular                8.3.23
@schematics/update                 0.901.7
ng-packagr                         9.1.5
</code></pre>

**Anything else relevant?**
I am using chrome 83.0.4103.97 64 bit if it helps

crisbeto:
Adds a note to the provider docs that users shouldn't mutate an array that 
is returned from a `multi` provider, because it can cause unforeseen 
consequences in other parts of the app.

Closes #37481.

### DIFF
--- a/aio/content/guide/dependency-injection-providers.md
+++ b/aio/content/guide/dependency-injection-providers.md
@@ -262,7 +262,7 @@ For example, when bootstrapping an application, you can register many initialize
 
 ```
 export const APP_TOKENS = [
- { provide: PLATFORM_INITIALIZER, useFactory: platformInitialized, multi: true    },
+ { provide: PLATFORM_INITIALIZER, useFactory: platformInitialized, multi: true },
  { provide: APP_INITIALIZER, useFactory: delayBootstrapping, multi: true },
  { provide: APP_BOOTSTRAP_LISTENER, useFactory: appBootstrapped, multi: true },
 ];
@@ -283,6 +283,23 @@ the [ROUTES](api/router/ROUTES) token combines all the different provided sets o
 Search for [Constants in API documentation](api?type=const) to find more built-in tokens.
 
 </div>
+
+<div class="alert is-helpful">
+
+Note that the reference to the array returned for a `multi` provider is shared between all the
+places where the token is injected. We recommend avoiding mutations of the array (especially for
+predefined tokens) as it may lead to unexpected behavior in other parts of the app that inject
+the same token. You can prevent the value from being mutated by setting its type to `ReadonlyArray`.
+
+</div>
+
+You can use `ReadonlyArray` to type your `multi` provider, so TypeScript triggers an error in case
+of unwanted array mutations:
+
+```
+constructor(@Inject(MULTI_PROVIDER) multiProvider: ReadonlyArray<MultiProvider>) {
+}
+```
 
 {@a tree-shakable-provider}
 {@a tree-shakable-providers}


aio/content/guide/dependency-injection-providers.md
```diff
@@ -278,6 +278,11 @@ When you provide multiple sets of routes using [RouterModule.forRoot](api/router
 and [RouterModule.forChild](api/router/RouterModule#forchild) in a single module,
 the [ROUTES](api/router/ROUTES) token combines all the different provided sets of routes into a single value.
 
+Note that the array that is returned for a `multi` provider is immutable in development mode, which
```
AndrewKushnir, line 281: I think we should mention that it's in Ivy only + may be add a note to Ivy compatibility guide?

AndrewKushnir (commented):
Thanks for this update @crisbeto 👍 

I believe this might be a risky change to merge into a patch branch. May be we should target "master-only" to have it in 10.1.0?

I also added @pkozlowski-opensource and @mhevery as reviewers to provide additional feedback.

Thank you.

aio/content/guide/dependency-injection-providers.md
crisbeto: My concern with mentioning Ivy here is that there's nothing that'll alert us to remove it once Ivy is the only rendering engine option.

crisbeto:
I think what we discussed was to run a presubmit for these changes to get a better idea of how breaking it would be.

AndrewKushnir:
Hi @crisbeto, right, I was thinking that it'd be breaking anyways (so we can apply labels, etc), we can just estimate the impact after running presubmit in Google's codebase. I've started it a few mins ago ([link](http://test/OCL:317733969:BASE:317734045:1592860792801:e02b7f91)) and will let you know the result. Thank you.

pkozlowski-opensource (approved):
LGTM from the code change point of view and based on what we've discussed during the fwk sync. 
A bit concerned that we do this freezing on the one off basis (there are probably other parts of the framework that we should review) but this is outside of the scope of this PR.

Marking as blocked since we need to wait for the presubmit results to see how breaky it is.

crisbeto:
@AndrewKushnir I've turned this into a docs change as discussed a while ago. Can you take another look?

AndrewKushnir:
Thanks for the update @crisbeto. I have a couple thoughts/ideas:

- should we highlight this section (using `<div class="alert is-helpful">..</div>`)? It feels like it's quite important piece of information.
- would it make sense to add an example on how/where to set the `Readonly` type?

Not related to this PR: it looks like we should consider splitting the `Predefined tokens and multiple providers` into 2 sections: one that talks about multi providers in general (how to use them) and the second section that describes predefined (multi) tokens. Currently it reads like multi providers can only be used to extend the built-in tokens, when in fact multi providers can be used with other (user-defined) tokens. @aikidave can we add that to the docs team backlog?

crisbeto:
Good points @AndrewKushnir, I've incorporated both of them. My only concern is that now we have two `<div class="alert is-helpful">` in a row.

aio/content/guide/dependency-injection-providers.md
```diff
@@ -284,6 +284,22 @@ Search for [Constants in API documentation](api?type=const) to find more built-i
...
+<div class="alert is-helpful">
+
+Note that the array that is returned for a `multi` provider is shared between all the places where
+the token is injected. We recommend that you don't mutate this array, because it can lead to unforeseen
+consequences in a different part of your app. You can prevent the value from being mutated by
+setting its type to `ReadonlyArray`.
+
+</div>
+
+For example, this is how you can type your `multi` provider to avoid unwanted array mutations:
```
AndrewKushnir, line 296:
```suggestion
You can use `ReadonlyArray` to type your `multi` provider, so TypeScript triggers an error in case of unwanted array mutations:
```

AndrewKushnir (commented):
Thanks @crisbeto! I've added a couple more comments - feel free to use (or ignore) them as you see fit.

AndrewKushnir:
> Not related to this PR: it looks like we should consider splitting the Predefined tokens and multiple providers section (see here) into 2 sections...

FYI, I also found a PR #29909 that introduces the multi provider section (but it looks like it needs some cleanup). cc @aikidave

crisbeto:
I've incorporated the latest set of feedback.

AndrewKushnir (approved):
Thanks for the changes @crisbeto 👍

pkozlowski-opensource (approved):
LGTM

PR #37650, opened 2020-06-19
Revert "fix(router): fix navigation ignoring logic to compare to the browser url (#37408)"

AndrewKushnir:
This reverts commit d3a817549b82c3f8c866296af4e952d1af5ac087.

The reason for the revert is the problem reported in g3 which requires additional investigation.

## PR Type

- [x] Bugfix

- [ ] Yes
- [x] No


AndrewKushnir:
@atscott I've marked it as "master-only", so that this PR can be merged asap. However the original PR was merged to master and patch, so the changes will remain in patch branch and will be included in v10 (which might be undesirable?). Feel free to set it to "master & patch" if it's more appropriate. Thank you.

AndrewKushnir:
[Presubmit](http://test/OCL:317359931:BASE:317359972:1592595669970:5098e7ac)

atscott (approved):

jelbourn:
Note for posterity: we're deciding to cherry-pick this revert into the 10.0.0 release because:
* The original change appears to have broken an application in Google
* The change is very small and the revert has been deemed low-risk

PR #37665, opened 2020-06-22
perf(ngcc): use `EntryPointManifest` to speed up noop `ProgramBaseEntryPointFinder`

petebacondarwin:
Previously the `ProgramBasedEntryPointFinder` was parsing all the
entry-points referenced by the program for dependencies even if all the
entry-points had been processed already.

Now this entry-point finder will re-use the `EntryPointManifest` to load
the entry-point dependencies when possible which avoids having to parse
them all again, on every invocation of ngcc.

Previously the `EntryPointManifest` was only used in the
`DirectoryWalkerEntryPointFinder`, which also contained the logic for
computing the contents of the manifest. This logic has been factored out
into an `EntryPointCollector` class. Both the `ProgramBasedEntryPointFinder`
and `DirectoryWalkerEntryPointFinder` now use the `EntryPointManifest` and
the `EntryPointCollector`.

The result of this change is that there is a small cost on the first run of
ngcc to compute and store the manifest - the processing takes 102% of the
processing time before this PR. But on subsequent runs there is a
significant benefit on subsequent runs - the processing takes around 50%
of the processing time before this PR.

---

Using the `super-productivity` as a test environment, we get the following
times for the "hot" (arguably noop) run of `ngcc` using the following command:

```bash
time node node_modules/.bin/ngcc --source node_modules --properties es2015 --first-only --create-ivy-entry-points --use-program-dependencies --async=false --tsconfig src/tsconfig.app.json
```

**Current master:**

(Last row is the average)

User time | System time | CPU usage | Total time
-- | -- | -- | --
2.93 | 0.75 | 106% | 3.465
2.74 | 0.61 | 119% | 2.811
2.8 | 0.62 | 120% | 2.835
2.76 | 0.61 | 119% | 2.818
2.74 | 0.61 | 120% | 2.777
**2.794** | **0.64** | **117%** | **2.9412**

**This PR:**

(Last row is the average)

User time | System time | CPU usage | Total time
-- | -- | -- | --
1.21 | 0.28 | 108% | 1.378
1.24 | 0.28 | 108% | 1.394
1.26 | 0.29 | 110% | 1.398
1.19 | 0.28 | 104% | 1.393
1.17 | 0.28 | 107% | 1.341
1.25 | 0.28 | 101% | 1.511
**1.22** | **0.28** | **106%** | **1.4025**

From these tables it can be seen that the noop (hot) run of ngcc drops from 2.94 to 1.4s (half the time).

### DIFF
--- a/packages/compiler-cli/ngcc/src/entry_point_finder/program_based_entry_point_finder.ts
+++ b/packages/compiler-cli/ngcc/src/entry_point_finder/program_based_entry_point_finder.ts
@@ -8,15 +8,17 @@
 import {AbsoluteFsPath, FileSystem} from '../../../src/ngtsc/file_system';
 import {Logger} from '../../../src/ngtsc/logging';
 import {ParsedConfiguration} from '../../../src/perform_compile';
-
-import {createDependencyInfo} from '../dependencies/dependency_host';
+import {createDependencyInfo, EntryPointWithDependencies} from '../dependencies/dependency_host';
 import {DependencyResolver} from '../dependencies/dependency_resolver';
 import {EsmDependencyHost} from '../dependencies/esm_dependency_host';
 import {ModuleResolver} from '../dependencies/module_resolver';
 import {NgccConfiguration} from '../packages/configuration';
+import {EntryPointManifest} from '../packages/entry_point_manifest';
 import {getPathMappingsFromTsConfig} from '../path_mappings';
 
+import {EntryPointCollector} from './entry_point_collector';
 import {TracingEntryPointFinder} from './tracing_entry_point_finder';
+import {trackDuration} from './utils';
 
 /**
  * An EntryPointFinder that starts from the files in the program defined by the given tsconfig.json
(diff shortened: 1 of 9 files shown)


packages/compiler-cli/ngcc/src/entry_point_finder/program_based_entry_point_finder.ts
```diff
@@ -7,16 +7,18 @@
  */
 import {AbsoluteFsPath, FileSystem} from '../../../src/ngtsc/file_system';
 import {ParsedConfiguration} from '../../../src/perform_compile';
-
```
gkalpak, line 10: 🤔

gkalpak (approved):

packages/compiler-cli/ngcc/src/entry_point_finder/tracing_entry_point_finder.ts
```diff
@@ -21,184 +20,79 @@ import {getBasePaths} from './utils';
...
+ * by the CLI integration.
  *
  * There are two concrete implementations of this class.
  *
- * * `TargetEntryPointFinder` - is given a single entry-point as the initial entry-point
- * * `ProgramBasedEntryPointFinder` - computes the initial entry-points from program files given by
- * a `tsconfig.json` file.
+ * * `TargetEntryPointFinder` - is given a single entry-point as the initial entry-point. This can
+ *   be used in the synchronous CLI integration where the build tool has identified an external
+ * import to one of the source files being built.
```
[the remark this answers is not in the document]
petebacondarwin: Clang formatter doesn't like manual indentation...

packages/compiler-cli/ngcc/src/entry_point_finder/program_based_entry_point_finder.ts
petebacondarwin: /shurg

JoostK (approved):

PR #12160, opened 2016-10-07
feat(router): add support for ng1/ng2 migration

vsavkin:

PR #37674, opened 2020-06-22
fix(router): ensure duplicate popstate/hashchange events are handled …

### LINKED ISSUE #16710: Navigation events are called twice when using HashLocationStrategy
```
[ ] feature request
```

**Current behavior**
When using `HashLocationStrategy` and change url manually to the one, which is protected by CanActivate guard (this returns false and fire redirect to the other page), navigation events are called twice.

**Expected behavior**
All navigation event are called once only.

Use the demo in http://plnkr.co/edit/4xlYD9wmetG5G0Dvjp0E?p=preview
Open window mode http://run.plnkr.co/plunks/4xlYD9wmetG5G0Dvjp0E/#/component-one
Click `Guarded Component Two` link - only one `NavigationStart `and `NavigationCancel `events logged in console.
Manually change url from `/#/component-one` to `/#/component-two` and hit enter - two pair of `NavigationStart `and `NavigationCancel` events logged in console.

As the duplication of navigation event causes guards logic to be executed twice (and this may contain API calls) - it is a performance issue. Potentially, if guard resolving depends on some conditions which are true after first failed attempt - the second one will be successful, however it is not desired behavior.

* **Angular version:** 4.1.0

* **Browser:** Chrome 57
 
* **Language:** TypeScript

atscott:
…correctly

The current method of handling duplicate navigations caused by 'hashchange' and 'popstate' events for the same url change does not correctly handle cancelled navigations. Because `scheduleNavigation` is called in a `setTimeout` in the location change subscription, the duplicate navigations are not flushed at the same time. This means that if the initial navigation hits a guard that schedules a new navigation, the navigation for the duplicate event will not compare to the correct transition (because we inserted another navigation between the duplicates). See https://github.com/angular/angular/issues/16710#issuecomment-646919529

Fixes #16710

Change summary: 
* Move the 'hashchange' / 'popstate' duplicate nav checks from `scheduleNavigation` to the location subscription. This is where the events are initially handled / detected 
* Use `pairwise` to compare with the previously detected 'hashchange' / 'popstate'
* Save the current router transition id at the time of the location event to aid in comparison

### DIFF
--- a/packages/router/src/router.ts
+++ b/packages/router/src/router.ts
@@ -6,9 +6,9 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {Location} from '@angular/common';
+import {Location, PopStateEvent} from '@angular/common';
 import {Compiler, Injectable, Injector, isDevMode, NgModuleFactoryLoader, NgModuleRef, NgZone, Type, ɵConsole as Console} from '@angular/core';
-import {BehaviorSubject, EMPTY, Observable, of, Subject, Subscription} from 'rxjs';
+import {BehaviorSubject, EMPTY, Observable, of, Subject, SubscriptionLike} from 'rxjs';
 import {catchError, filter, finalize, map, switchMap, tap} from 'rxjs/operators';
 
 import {QueryParamsHandling, Route, Routes, standardizeConfig, validateConfig} from './config';
@@ -277,6 +277,16 @@ function defaultRouterHook(snapshot: RouterStateSnapshot, runExtras: {
 }
 
 /**
+ * Information related to a location change, necessary for scheduling follow-up Router navigations.
+ */
+type LocationChangeInfo = {
+  source: 'popstate'|'hashchange',
+  urlTree: UrlTree,
+  state: RestoredState|null,
+  transitionId: number
+};
+
+/**
  * @description
  *
  * A service that provides navigation and URL manipulation capabilities.
@@ -298,8 +308,12 @@ export class Router {
   private lastSuccessfulNavigation: Navigation|null = null;
   private currentNavigation: Navigation|null = null;
 
-  // TODO(issue/24571): remove '!'.
-  private locationSubscription!: Subscription;
+  private locationSubscription?: SubscriptionLike;
+  /**
+   * Tracks the previously seen location change from the location subscription so we can compare
+   * the two latest to see if they are duplicates. See setUpLocationChangeListener.
+   */
+  private lastLocationChangeInfo: LocationChangeInfo|null = null;
   private navigationId: number = 0;
   private configLoader: RouterConfigLoader;
   private ngModule: NgModuleRef<any>;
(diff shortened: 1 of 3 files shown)


atscott:
[presubmit](http://test/OCL:317682448:BASE:317696352:1592850205107:6d76faf0)

atscott:
[global presubmit](https://test.corp.google.com/ui#id=OCL:317682448:BASE:317860733:1592922368361:185632ec)

packages/router/src/router.ts
```diff
@@ -858,16 +858,51 @@ export class Router {
...
-        let rawUrlTree = this.parseUrl(change['url']);
-        const source: NavigationTrigger = change['type'] === 'popstate' ? 'popstate' : 'hashchange';
-        // Navigations coming from Angular router have a navigationId state property. When this
-        // exists, restore the state.
-        const state = change.state && change.state.navigationId ? change.state : null;
-        setTimeout(() => {
-          this.scheduleNavigation(rawUrlTree, source, state, {replaceUrl: true});
-        }, 0);
-      });
+      this.locationSubscription = new Subscription();
```
jelbourn, line 861: Rather than manually tracking subscriptions, could this use `takeUntil` with a `destroyed` subject?
(https://medium.com/@benlesh/rxjs-dont-unsubscribe-6753ed4fda87)

packages/router/src/router.ts
```diff
@@ -858,16 +858,51 @@ export class Router {
...
+          this._locationSubject
+              .pipe(
+                  map(change => ({
+                        source: change['type'] === 'popstate' ? 'popstate' : 'hashchange',
+                        urlTree: this.parseUrl(change['url']!),
+                        // Navigations coming from Angular router have a navigationId state
+                        // property. When this exists, restore the state.
+                        state: change.state && change.state.navigationId ? change.state : null,
+                        transitionId: this.getTransition().id
+                      } as const)),
```
jelbourn, line 873: Could this mapping be split into a separate function? Something like
```ts
map(getSomethingForRouteChange)
```
(I'm not sure how to describe the type being constructed here)

packages/router/src/router.ts
```diff
@@ -858,16 +858,51 @@ export class Router {
...
+                  startWith(null),
+                  pairwise(),
+                  )
+              .subscribe(([previous, current]) => {
+                const {source, state, urlTree} = current!;
+
+                function shouldScheduleNavigation(
```
jelbourn, line 880: I would move this function definition out of the subscribe block to a module-level function

packages/router/src/router.ts
```diff
@@ -858,16 +858,51 @@ export class Router {
...
+                        urlTree: this.parseUrl(change['url']!),
+                        // Navigations coming from Angular router have a navigationId state
+                        // property. When this exists, restore the state.
+                        state: change.state && change.state.navigationId ? change.state : null,
+                        transitionId: this.getTransition().id
+                      } as const)),
+                  startWith(null),
+                  pairwise(),
+                  )
+              .subscribe(([previous, current]) => {
```
jelbourn, line 877: I think this would be more readable if you split this up form being one large statement, something roughly like
```ts
// Compose a stream of ___ that emits both the previous and current route state
// so that we can compare the previous to the current state and determine ___.
const pairwiseRouteChanges = this._locationSubject.pipe(
  map(getSomethingForRouteChange),
  startWith(null),
  pairwise(),
  takeUntil(this.destroyed),
);

pairwiseRouteChanges.subscribe(([previous, current]) => {
  ...
});
```

packages/router/src/router.ts
```diff
@@ -858,16 +858,51 @@ export class Router {
...
+                const {source, state, urlTree} = current!;
+
+                function shouldScheduleNavigation(
+                    previous: {source: string, urlTree: UrlTree, transitionId: number}|null,
+                    current: {source: string, urlTree: UrlTree, transitionId: number}): boolean {
```
jelbourn, line 882: Introduce a local type alias for this?

packages/router/src/router.ts
```diff
@@ -858,16 +858,51 @@ export class Router {
...
+                  if ((current.source === 'hashchange' && previous.source === 'popstate') ||
+                      (current.source === 'popstate' && previous.source === 'hashchange')) {
+                    return false;
+                  }
+
+                  return true;
+                }
+
+                if (shouldScheduleNavigation(previous, current!)) {
+                  setTimeout(() => {
```
jelbourn, line 901: I know it was already like this, but optionally it would be nice to add a comment here that explains why this timeout is used.

packages/router/src/router.ts
```diff
@@ -298,8 +298,8 @@ export class Router {
   private lastSuccessfulNavigation: Navigation|null = null;
   private currentNavigation: Navigation|null = null;
 
-  // TODO(issue/24571): remove '!'.
-  private locationSubscription!: Subscription;
+  private readonly _locationSubject = new Subject<PopStateEvent>();
```
jelbourn, line 301: (can't comment on unchanged lines)
Optional: even though it was already like this, it would be great to expand the JsDoc description on `setUpLocationChangeListener` to expand on why there's a location change listener and what it's doing in response to those changes.

jelbourn (commented):
Is this a do-over of the previous router change that was rolled back? If so, I would reference that original change and its revert in the commit message for reference.

atscott:
> Is this a do-over of the previous router change that was rolled back? If so, I would reference that original change and its revert in the commit message for reference.

This change is unrelated to the one that was rolled back.

packages/router/src/router.ts
atscott: `setUpLocationChangeListener` is public so we need to track the subscription somehow since we don't have control over how many times it's called. 
As far as `takeUntil` and a `destroyed` subject, the difference is pretty minimal. They're both methods of tracking subscriptions that need to be unsubscribed on destroy. I think this is usually just a stylistic choice; since we have to track the existence of the subscription already, this approach meant I didn't have to create another `Subject`.

packages/router/src/router.ts
atscott: The `setTimeout` was added in [feat(router): add support for ng1/ng2 migration](https://github.com/angular/angular/pull/12160). My best guess is that this is somehow related to hybrid apps, but I really don't know the answer to why.
I can add a comment to that effect.

jelbourn (approved):
LGTM

jelbourn (approved):
LGTM

Reviewed-for: size-tracking

packages/router/src/router.ts
```diff
@@ -851,26 +861,70 @@ export class Router {
...
   }
 
+  /** Extracts router-related information from a `PopStateEvent`. */
+  private extractLocationChangeInfoFromEvent(change: PopStateEvent): LocationChangeInfo {
+    return {
+      source: change['type'] === 'popstate' ? 'popstate' : 'hashchange',
+      urlTree: this.parseUrl(change['url']!),
+      // Navigations coming from Angular router have a navigationId state
+      // property. When this exists, restore the state.
+      state: change.state && change.state.navigationId ? change.state : null,
```
JoostK, line 901: While we're here:

```suggestion
      state: change.state?.navigationId ? change.state : null,
```

packages/router/src/router.ts
```diff
@@ -851,26 +861,70 @@ export class Router {
...
+
+  /**
+   * Determines whether two events triggered by the Location subscription are due to the same
+   * navigation. The location subscription can fire two events (popstate and hashchange) for a
+   * single navigation. The second one should be ignored, that is, we should not schedule another
+   * navigation in the Router.
+   */
+  private shouldScheduleNavigation(previous: LocationChangeInfo|null, current: LocationChangeInfo):
+      boolean {
+    const sameDestination = current.urlTree.toString() !== previous?.urlTree?.toString();
```
JoostK, line 914: `sameDestination` sounds inverted from the what the comparison represents?

packages/router/src/router.ts
```diff
@@ -851,26 +861,70 @@ export class Router {
...
+    const eventsOccurredAtSameTime = current.transitionId === previous?.transitionId;
+    if (!previous || !eventsOccurredAtSameTime || !sameDestination) {
```
JoostK, line 916: nit: looks like this bails out if `previous` is `null`, so I would move this outside of this fn, allowing the `previous` param to drop the `null` type.

JoostK (approved):
This is generally a great cleanup! I left some suggestions in the code. Otherwise LGTM.

IgorMinar (dismissed):
Could we fix this bug without the ~800b penalty? @JoostK's suggestion looks reasonable to me. I know that it sucks, but 800b is quite a bit especially if it can be avoided.

Reviewed-for: size-tracking

atscott:
[updated presubmit](http://test/OCL:317682448:BASE:318834589:1593449683795:34728148)
May want to do another global as well

pkozlowski-opensource (approved):
LGTM

Reviewed-for: size-tracking

pkozlowski-opensource (approved):
LGTM

Reviewed-for: size-tracking

alxhub (approved):
Reviewed-for: size-tracking

Issue #37653, opened 2020-06-20
`processNewMacroTasksSynchronously: false` does not work well with rxjs' timer and interval functions


artaommahe:
# 🐞 bug report

### Affected Package
The issue is caused by package @angular/angular/core/testing

### Is this a regression?

No

### Description

Using `tick(0, { processNewMacroTasksSynchronously: false })` with rxjs' `timer` and `interval` does not worked as expected.

## 🔬 Minimal Reproduction

Example with timer, callback never been called

```ts
it('with timer', fakeAsync(() => {
  const callback = jasmine.createSpy('callback');

  timer(0, 0).pipe(skip(1), take(1)).subscribe(callback);

  expect(callback).not.toHaveBeenCalled();

  tick(undefined, { processNewMacroTasksSynchronously: false });

  expect(callback).not.toHaveBeenCalled();

  tick(undefined, { processNewMacroTasksSynchronously: false });

  expect(callback).toHaveBeenCalledTimes(1);
}));
```

Same with interval, code in test file.
Repo with this test

- `git clone git@github.com:artaommahe/tick-with-timer-and-interval.git`
- `yarn install`
- `yarn test`

[test file](https://github.com/artaommahe/tick-with-timer-and-interval/blob/master/src/app/tickWithTimerAndInterval.spec.ts)

Even using `tick()` does not work here.

## 🔥 Exception or Error
<pre><code>
Chrome 83.0.4103.106 (Windows 10) tickWithTimerAndInterval with timer FAILED
        Error: Expected spy callback to have been called once. It was called 0 times.
            at <Jasmine>
            at UserContext.<anonymous> (http://localhost:9876/_karma_webpack_/src/app/tickWithTimerAndInterval.spec.ts:24:22)
</code></pre>

## 🌍  Your Environment

**Angular Version:**
<pre><code>
Angular CLI: 10.0.0-rc.5
Node: 12.18.1
OS: win32 x64

Angular: 10.0.0-rc.6
... animations, common, compiler, compiler-cli, core, forms
... platform-browser, platform-browser-dynamic, router
Ivy Workspace: Yes

Package                           Version
-----------------------------------------------------------
@angular-devkit/architect         0.1000.0-rc.5
@angular-devkit/build-angular     0.1000.0-rc.5
@angular-devkit/build-optimizer   0.1000.0-rc.5
@angular-devkit/build-webpack     0.1000.0-rc.5
@angular-devkit/core              10.0.0-rc.5
[...]

PR #37680, opened 2020-06-23
fix(core): fake_async_fallback should have the same logic with fake-async

JiaLiPassion:
PR #37523 failed when trying to use `rxjs delay` operator
inside `fakeAsync`, and the reasons are:

1. we need to import `rxjs-fake-async` patch to make the integration work.
2. since in `angular` repo, the bazel target `/tools/testing:node` not using `zone-testing` bundle,
instead it load `zone-spec` packages seperately, so it causes one issue which is the `zone.js/testing/fake-async`
package is not loaded, we do have a fallback logic under `packages/core/testing` calles `fake_async_fallback`,
but the logic is out of date with `fake-async` under `zone.js` package.
https://github.com/angular/angular/blob/master/packages/zone.js/lib/testing/fake-async.ts

Now the `fake_async_fallback` didn't handle `patch Date.now` correctly, and this issue has been fixed in `fake-async` under `zone.js` package.

So this PR, I updated the content of `fake_async_fallback` to make it consistent with
`fake-async` to handle `Date.now` patch.

And I will make another PR to try to remove the `fallback` logic.

And the `Angular` application will not have this issue since they all import `zone.js/dist/zone-testing`, only `angular repo` testing may have this issue.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: 37523

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/testing/src/fake_async_fallback.ts
+++ b/packages/core/testing/src/fake_async_fallback.ts
@@ -27,6 +27,9 @@ let _fakeAsyncTestZoneSpec: any = null;
  * @publicApi
  */
 export function resetFakeAsyncZoneFallback() {
+  if (_fakeAsyncTestZoneSpec) {
+    _fakeAsyncTestZoneSpec.unlockDatePatch();
+  }
   _fakeAsyncTestZoneSpec = null;
   // in node.js testing we may not have ProxyZoneSpec in which case there is nothing to reset.
   ProxyZoneSpec && ProxyZoneSpec.assertPresent().resetDelegate();
@@ -73,6 +76,7 @@ export function fakeAsyncFallback(fn: Function): (...args: any[]) => any {
       let res: any;
       const lastProxyZoneSpec = proxyZoneSpec.getDelegate();
       proxyZoneSpec.setDelegate(_fakeAsyncTestZoneSpec);
+      _fakeAsyncTestZoneSpec.lockDatePatch();
       try {
         res = fn.apply(this, args);
         flushMicrotasksFallback();


JoostK:
Would this maybe also address #37653?

atscott:
[presubmit](http://test/OCL:317852724:BASE:317852737:1592918729076:f69f53b7)

packages/core/testing/src/fake_async_fallback.ts
```diff
@@ -27,6 +27,9 @@ let _fakeAsyncTestZoneSpec: any = null;
  * @publicApi
  */
 export function resetFakeAsyncZoneFallback() {
+  if (_fakeAsyncTestZoneSpec) {
```
atscott, line 30: Hi Jia, should this be extracted to a shared method so we don't have to maintain two identical implementations here and in `fake-async.ts`? I noticed that these two methods are exactly the same and `fakeAsync` / `fakeAsyncFallback` are _almost_ the same so maybe there are some common blocks that can be extracted in those as well.

JiaLiPassion: @atscott , yes, you are right, the code here is for some history reason, since the old `zone.js` doesn't have this logic, I will create another PR to remove all those `fallback` logic.

atscott (approved):

mhevery (changes requested):
Could we have a test demonstrating the correct behavior?

JiaLiPassion:
@mhevery, there is test for the `patch Date` in `zone.js`, and the `fallback` logic doesn't have any test, since it should not be called at all, so this PR is a quick fix to make the current test work, I will create a follow up PR to remove all the `fallback` logic.

mhevery (approved):

AndrewKushnir (approved):

PR #37686, opened 2020-06-23
refactor(core): delegate `EventEmitter` subscription signatures processing to `Subscriber`

dimakuba:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

For now `EventEmitter` takes care of the processing subscription signatures.

## What is the new behavior?
Would be better to delegate this processing to `Subscriber` which knows all subscription signatures and how to handle them.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/event_emitter.ts
+++ b/packages/core/src/event_emitter.ts
@@ -104,64 +104,20 @@ class EventEmitter_ extends Subject<any> {
     super.next(value);
   }
 
-  subscribe(generatorOrNext?: any, error?: any, complete?: any): Subscription {
-    let schedulerFn: (t: any) => any;
-    let errorFn = (err: any): any => null;
-    let completeFn = (): any => null;
-
-    if (generatorOrNext && typeof generatorOrNext === 'object') {
-      schedulerFn = this.__isAsync ? (value: any) => {
-        setTimeout(() => generatorOrNext.next(value));
-      } : (value: any) => {
-        generatorOrNext.next(value);
-      };
-
-      if (generatorOrNext.error) {
-        errorFn = this.__isAsync ? (err) => {
-          setTimeout(() => generatorOrNext.error(err));
-        } : (err) => {
-          generatorOrNext.error(err);
-        };
-      }
-
-      if (generatorOrNext.complete) {
-        completeFn = this.__isAsync ? () => {
-          setTimeout(() => generatorOrNext.complete());
-        } : () => {
-          generatorOrNext.complete();
-        };
-      }
-    } else {
-      schedulerFn = this.__isAsync ? (value: any) => {
-        setTimeout(() => generatorOrNext(value));
-      } : (value: any) => {
-        generatorOrNext(value);
-      };
-
-      if (error) {
-        errorFn = this.__isAsync ? (err) => {
-          setTimeout(() => error(err));
-        } : (err) => {
-          error(err);
-        };
-      }
-
-      if (complete) {
-        completeFn = this.__isAsync ? () => {
-          setTimeout(() => complete());
-        } : () => {
-          complete();
-        };
-      }
+  /**
+   * @internal
+   */
+  _subscribe(subscriber: any): Subscription {
+    if (this.__isAsync) {
+      subscriber.next = (value: any) =>
+          setTimeout(subscriber.constructor.prototype.next.bind(subscriber, value));
+      subscriber.error = (err: any) =>
+          setTimeout(subscriber.constructor.prototype.error.bind(subscriber, err));
+      subscriber.complete = () =>
+          setTimeout(subscriber.constructor.prototype.complete.bind(subscriber));
     }
 
-    const sink = super.subscribe(schedulerFn, errorFn, completeFn);
-
-    if (generatorOrNext instanceof Subscription) {
-      generatorOrNext.add(sink);
-    }
-
-    return sink;
+    return super._subscribe(subscriber);
   }
 }
 


goldens/size-tracking/aio-payloads.json
```diff
@@ -12,7 +12,7 @@
     "master": {
       "uncompressed": {
         "runtime-es2015": 2987,
-        "main-es2015": 451406,
+        "main-es2015": 475138,
```
jelbourn, line 15: This appears to be increasing the size of angular.io by 23 kilobytes?

gkalpak: No, this change is not valid. The 475K sometimes appears on CI (which I assume is where @dimakuba got it), but it is wrong (we are trying to figure out why it happens).

dimakuba: Now it indicates correct payload size increasing by 43B.

dimakuba: After my change to correct size, CI again sent wrong number 😕

packages/core/src/event_emitter.ts
```diff
@@ -105,63 +105,44 @@ class EventEmitter_ extends Subject<any> {
...
+}
 
-      if (generatorOrNext.error) {
-        errorFn = this.__isAsync ? (err) => {
-          setTimeout(() => generatorOrNext.error(err));
-        } : (err) => {
-          generatorOrNext.error(err);
-        };
-      }
+class AsyncSubscriber<T> extends Subscriber<T> {
```
jelbourn, line 112: Could you add a JsDoc description to this class that explains what it does differently from `Subscriber`?

packages/core/src/event_emitter.ts
```diff
@@ -105,63 +105,44 @@ class EventEmitter_ extends Subject<any> {
...
+  constructor(private isAsync: boolean, generatorOrNext?: any, error?: any, complete?: any) {
```
jelbourn, line 113: Since this is a new API, could you replace the `any` use in this constructor with more specific types?

packages/core/src/event_emitter.ts
```diff
@@ -105,63 +105,44 @@ class EventEmitter_ extends Subject<any> {
...
+      notification.observe(this.destination);
     }
+  }
 
-    const sink = super.subscribe(schedulerFn, errorFn, completeFn);
+  /**
+   * @internal
+   */
+  protected _next(value: T): void {
+    this.schedule(Notification.createNext(value));
```
jelbourn, line 129: I'm pretty dumb when it comes to RxJS- would it make sense to add a comment for `_next`, `_error`, and `_complete` in this class that explains why they don't call their corresponding `super` method?

IgorMinar (dismissed):
Hi @dimakuba, thanks for trying to clean up this code, but unfortunately our approach brings in a lot of stuff from RxJS into the Angular Core which increases the baseline payload size for Angular applications.

Is it possible to make this change while not importing more code from RxJS?

Also, it sounds like your motivation here is for the `EventEmitter` to support new signatures. If that's the case then can you please explain what signatures you are trying to support?

I'm asking because your change doesn't actually change the type surface of the EventEmitter interface, so your change would have only runtime impact and would likely require explicit cast for the code to compile. This is undesirable and we won't accept a change which intentionally widens the runtime signature compared to the compile-time signature.

If you could provide more info about your motivations then we could assist you better with this change. Thanks

Reviewed-for: size-tracking

dimakuba:
> Hi @dimakuba, thanks for trying to clean up this code, but unfortunately our approach brings in a lot of stuff from RxJS into the Angular Core which increases the baseline payload size for Angular applications.
> 
> Is it possible to make this change while not importing more code from RxJS?
> 
> Also, it sounds like your motivation here is for the `EventEmitter` to support new signatures. If that's the case then can you please explain what signatures you are trying to support?
> 
> I'm asking because your change doesn't actually change the type surface of the EventEmitter interface, so your change would have only runtime impact and would likely require explicit cast for the code to compile. This is undesirable and we won't accept a change which intentionally widens the runtime signature compared to the compile-time signature.
> 
> If you could provide more info about your motivations then we could assist you better with this change. Thanks
> 
> Reviewed-for: size-tracking

Hi @IgorMinar, thanks for the review. My motivations is to encapsulate subscribe signatures processing to `RxJS` by avoiding code duplications and always be in sync with `RxJS`. I'll provide another solution without payload size increasing asap.

dimakuba:
Hi @IgorMinar, could you please take a look for the solution?

jelbourn (approved):
LGTM, this seems to clean up the code quite nicely. The original code is pretty hard to follow, but as far as I can tell it looks like this has the same result. 

@pkozlowski-opensource is the only reviewer still needed for this now

AndrewKushnir:
FYI, just came across this PR while reviewing #41450. If I understand correctly, this change relies on the `_subscribe` function of the `Subject` class defined in RxJS. The naming of the `_subscribe` method hints at it being private and likely unexpected for external use. I believe relying on this method might be risky long-term and we'd have to rework this code (and potentially change the overall approach) in case `_subscribe` changes in one of the next RxJS versions (and if it's a private function, there is a chance that the change would not be announced in release notes, etc).

Current implementation of the `EventEmitter` duplicates some code and #41450 should address that. The implementation in this PR is still a bit more compact, but that comes with the risk of a potential breaking change, so I'm closing this PR in favor of PR #41450. Please reopen if needed.

dimakuba:
Provided another approach earlier in this PR with no ricks (because of using only public api), but unfortunately, it increases the budget because of extra imports from 'rxjs'

PR #37690, opened 2020-06-23
perf(compiler-cli): fix regressions in incremental program reuse

alxhub:
Version of #37641 for the patch branch.

Review not required; this change has already been approved for `master`.


AndrewKushnir:
This PR is now merged into the patch (10.0.x) branch.

PR #37700, opened 2020-06-24
docs: create coding standards doc

jelbourn:
Create initial document for Angular framework coding standards. This
document will evolve over time. This version contains all
non-controversial rules as discussed in a recent team meeting. Some text
and examples were copied from angular/components.

### DIFF
--- a/docs/CODING_STANDARDS.md
+++ b/docs/CODING_STANDARDS.md
@@ -1,0 +1,249 @@
+# Angular Framework Coding Standards
+
+## Code style
+
+The [Google JavaScript Style Guide](https://google.github.io/styleguide/jsguide.html) is the
+basis for Angular's coding style, with additional guidance here pertaining to TypeScript. The team
+uses `clang-format` to automatically format code; automatic formatting is enforced by CI.
+
+## Code practices
+
+#### Write useful comments
+
+Comments that explain what some block of code does are nice; they can tell you something in less
+time than it would take to follow through the code itself.
+
+Comments that explain why some block of code exists at all, or does something the way it does,
+are _invaluable_. The "why" is difficult, or sometimes impossible, to track down without seeking out
+the original author. When collaborators are in the same room, this hurts productivity.
+When collaborators are in different timezones, this can be devastating to productivity.
+
+For example, this is a not-very-useful comment:
+```typescript
+// Set default tabindex.
+if (!attributes['tabindex']) {
+  element.setAttribute('tabindex', '-1');
+}
+```
+
+While this is much more useful:
+```typescript
+// Unless the user specifies otherwise, the calendar should not be a tab stop.
+// This prevents ngAria from overzealously adding a tabindex to anything with an ng-model.
+if (!attributes['tabindex']) {
+  element.setAttribute('tabindex', '-1');
+}
+```
+
+In TypeScript code, use JsDoc-style comments for descriptions (on classes, members, etc.) and
+use `//` style comments for everything else (explanations, background info, etc.).
+
+### API Design
+
+#### Boolean arguments
+
+Avoid adding boolean arguments to a method in cases where that argument means "do something extra".
+In these cases, prefer breaking the behavior up into different functions.
+
+```typescript
+// AVOID
+function getTargetElement(createIfNotFound = false) {
+  // ...
+}
+```
+
+```typescript
+// PREFER
+function getExistingTargetElement() {
+  // ...
+}
+
+function createTargetElement() {
+ // ...
+}
+```
+
+#### Optional arguments
+
+Use optional function arguments only when such an argument makes sense for an API. Don't use
+optional arguments merely for convenience in implementation.
+
+### TypeScript
+
+#### Typing
+
+Avoid `any` where possible. If you find yourself using `any`, consider whether a generic or
+`unknown` may be appropriate in your case.
+
+#### Getters and Setters
+
+* Only use getters and setters for `@Input` properties or when otherwise required for API
+compatibility.
+* Avoid long or complex getters and setters. If the logic of an accessor would take more than
+three lines, introduce a new method to contain the logic.
+* A getter should immediately precede its corresponding setter.
+* Decorators such as `@Input` should be applied to the getter and not the setter.
+* Always use a `readonly` property instead of a getter (with no setter) when possible.
+
+  ```typescript
+  /** YES */
+  readonly active: boolean;
+
+  /** NO */
+  get active(): boolean {
+    // Using a getter solely to make the property read-only.
+    return this._active;
+  }
+  ```
+
+#### Iteration
+
+Prefer `for of` to `Array.prototype.forEach`. The `for of` construct makes debugging easier and may
+reduce overhead from unnecessary function invocations (though browsers do generally optimize this
+well).
+
+#### JsDoc comments
+
+All public APIs must have user-facing comments. These are extracted for API documentation and shown
+in IDEs.
+
+Private and internal APIs should have JsDoc when they are not obvious. Ultimately it is the purview
+of the code reviewer as to what is "obvious", but the rule of thumb is that *most* classes,
+properties, and methods should have a JsDoc description.
+
+Properties should have a concise description of what the property means:
+```typescript
+  /** The label position relative to the checkbox. Defaults to 'after' */
+  @Input() labelPosition: 'before' | 'after' = 'after';
+```
+
+Methods blocks should describe what the function does and provide a description for each parameter
+and the return value:
+```typescript
+  /**
+   * Opens a modal dialog containing the given component.
+   * @param component Type of the component to load into the dialog.
+   * @param config Dialog configuration options.
+   * @returns Reference to the newly-opened dialog.
+   */
+  open<T>(component: ComponentType<T>, config?: MatDialogConfig): MatDialogRef<T> { ... }
+```
+
+Boolean properties and return values should use "Whether..." as opposed to "True if...":
+```ts
+  /** Whether the button is disabled. */
+  disabled: boolean = false;
+```
+
+#### Try-Catch
+
+Avoid `try-catch` blocks, instead preferring to prevent an error from being thrown in the first
+place. When impossible to avoid, the `try-catch` block **must** include a comment that explains the
+specific error being caught and why it cannot be prevented.
+
+#### Private contructors
+
+Consider private constructors for all new code. This prevents extension outside of the framework,
+which is generally discouraged unless otherwise documented.
+
+##### Variable declarations
+
+Prefer `const` wherever possible. Avoid `var` unless absolutely necessary.
+
+##### `readonly`
+
+Use `readonly` members wherever possible.
+
+#### Naming
+
+##### General
+
+* Prefer writing out words instead of using abbreviations.
+* Prefer *exact* names over short names (within reason). E.g., `labelPosition` is better than
+`align` because the former much more exactly communicates what the property means.
+* Except for `@Input` properties, use `is` and `has` prefixes for boolean properties / methods.
+
+##### Observables
+
+Don't suffix observables with `$`.
+
+##### Classes
+
+Name classes based on their responsibility. Names should capture what the code *does*,
+not how it is used:
+```typescript
+/** NO: */
+class DefaultRouteReuseStrategy { }
+
+/** YES: */
+class NonStoringRouteReuseStrategy { }
+``` 
+
+##### Methods
+
+The name of a method should capture the action performed *by* that method rather than
+describing when the method will be called. For example:
+
+```typescript
+/** AVOID: does not describe what the function does. */
+handleClick() {
+  // ...
+}
+
+/** PREFER: describes the action performed by the function. */
+activateRipple() {
+  // ...
+}
+```
+
+##### Test classes and examples
+
+Give test classes and examples meaningful, descriptive names.
+
+```ts
+/** PREFER: describes the scenario under test. */
+class FormGroupWithCheckboxAndRadios { /* ... */ }
+class InputWithNgModel { /* ... */ }
+
+
+/** AVOID: does not fully describe the scenario under test. */
+class Comp { /* ... */ }
+class InputComp { /* ... */ }
+```
+
+
+#### RxJS
+
+When importing the `of` function to create an `Observable` from a value, alias the imported
+function as `observableOf`.
+
+```typescript
+import {of as observableOf} from 'rxjs';
+```
+
+#### Testing
+
+##### Test names
+
+Use descriptive names for jasmine tests. Ideally, test names should read as a sentence, often of
+the form "it should...".
+
+```typescript
+/** PREFER: describes the scenario under test. */
+describe('Router', () => {
+  describe('with the default route reuse strategy', () => {
+    it('should not reuse routes upon location change', () => {
+      // ...
+    });
+  })
+});
+
+/** AVOID: does not fully describe the scenario under test. */
+describe('Router', () => {
+  describe('default strategy', () => {
+    it('should work', () => {
+      // ...
+    });
+  })
+});
+```
--- a/docs/CODING_STANDARDS.md
+++ b/docs/CODING_STANDARDS.md
@@ -1,5 +1,8 @@
 # Angular Framework Coding Standards
 
+The coding practices in this doc apply only to development on Angular itself, not applications
+built _with_ Angular. (Though you can follow them too if you really want).
+
 ## Code style
 
 The [Google JavaScript Style Guide](https://google.github.io/styleguide/jsguide.html) is the
@@ -8,7 +11,7 @@ uses `clang-format` to automatically format code; automatic formatting is enforc
 
 ## Code practices
 
-#### Write useful comments
+### Write useful comments
 
 Comments that explain what some block of code does are nice; they can tell you something in less
 time than it would take to follow through the code itself.
@@ -42,8 +45,8 @@ use `//` style comments for everything else (explanations, background info, etc.
 
 #### Boolean arguments
 
-Avoid adding boolean arguments to a method in cases where that argument means "do something extra".
-In these cases, prefer breaking the behavior up into different functions.
+Generally avoid adding boolean arguments to a method in cases where that argument means
+"do something extra". In these cases, prefer breaking the behavior up into different functions.
 
 ```typescript
 // AVOID
@@ -59,14 +62,16 @@ function getExistingTargetElement() {
 }
 
 function createTargetElement() {
- // ...
+  // ...
 }
 ```
 
+You can ignore this guidance when necessary for performance reasons in framework code.
+
 #### Optional arguments
 
-Use optional function arguments only when such an argument makes sense for an API. Don't use
-optional arguments merely for convenience in implementation.
+Use optional function arguments only when such an argument makes sense for an API or when required
+for performance. Don't use optional arguments merely for convenience in implementation.
 
 ### TypeScript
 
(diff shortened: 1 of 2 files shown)


docs/CODING_STANDARDS.md
```diff
@@ -0,0 +1,249 @@
...
+##### General
+
+* Prefer writing out words instead of using abbreviations.
+* Prefer *exact* names over short names (within reason). E.g., `labelPosition` is better than
+`align` because the former much more exactly communicates what the property means.
+* Except for `@Input` properties, use `is` and `has` prefixes for boolean properties / methods.
+
+##### Observables
+
+Don't suffix observables with `$`.
```
CaerusKaru, line 168: Proponents of Polish notation everywhere weep

docs/CODING_STANDARDS.md
```diff
@@ -0,0 +1,249 @@
...
+    // Using a getter solely to make the property read-only.
+    return this._active;
+  }
+  ```
+
+#### Iteration
+
+Prefer `for of` to `Array.prototype.forEach`. The `for of` construct makes debugging easier and may
+reduce overhead from unnecessary function invocations (though browsers do generally optimize this
+well).
```
JoostK, line 103: I would add an exception for hot code paths here, e.g. in the Ivy runtime only `for (const i = 0; i < a.length; i++)` is used when iterating arrays.

In general, maybe it would be a good idea to include a chapter for performance sensitive code, where some of these standards may not be desirable.

docs/CODING_STANDARDS.md
```diff
@@ -0,0 +1,249 @@
...
+#### Try-Catch
+
+Avoid `try-catch` blocks, instead preferring to prevent an error from being thrown in the first
+place. When impossible to avoid, the `try-catch` block **must** include a comment that explains the
+specific error being caught and why it cannot be prevented.
+
+#### Private contructors
+
+Consider private constructors for all new code. This prevents extension outside of the framework,
+which is generally discouraged unless otherwise documented.
```
JoostK, line 147: This could use some samples as to what a private constructor is and how such classes have to be instantiated instead using named constructors in the form of static members.

docs/CODING_STANDARDS.md
```diff
@@ -0,0 +1,249 @@
...
+
+##### Variable declarations
+
+Prefer `const` wherever possible. Avoid `var` unless absolutely necessary.
```
JoostK, line 151: I would mention `let` as the alternative for `var`, as currently this may be interpreted as `var` being the one alternative to `const`.

docs/CODING_STANDARDS.md
```diff
@@ -0,0 +1,249 @@
...
+
+Give test classes and examples meaningful, descriptive names.
+
+```ts
+/** PREFER: describes the scenario under test. */
+class FormGroupWithCheckboxAndRadios { /* ... */ }
+class InputWithNgModel { /* ... */ }
+
+
+/** AVOID: does not fully describe the scenario under test. */
```
JoostK, line 209: You don't want to look at the compiler tests, then :-)

docs/CODING_STANDARDS.md
```diff
@@ -0,0 +1,249 @@
...
+
+In TypeScript code, use JsDoc-style comments for descriptions (on classes, members, etc.) and
+use `//` style comments for everything else (explanations, background info, etc.).
+
+### API Design
+
+#### Boolean arguments
+
+Avoid adding boolean arguments to a method in cases where that argument means "do something extra".
+In these cases, prefer breaking the behavior up into different functions.
```
JoostK, line 46: An alternative alternative could be to have an enum to replace the boolean.

docs/CODING_STANDARDS.md
```diff
@@ -0,0 +1,249 @@
...
+```
+
+```typescript
+// PREFER
+function getExistingTargetElement() {
+  // ...
+}
+
+function createTargetElement() {
+ // ...
```
gkalpak, line 62:
```suggestion
  // ...
```

docs/CODING_STANDARDS.md
```diff
@@ -0,0 +1,249 @@
...
+// This prevents ngAria from overzealously adding a tabindex to anything with an ng-model.
+if (!attributes['tabindex']) {
+  element.setAttribute('tabindex', '-1');
+}
+```
+
+In TypeScript code, use JsDoc-style comments for descriptions (on classes, members, etc.) and
+use `//` style comments for everything else (explanations, background info, etc.).
+
+### API Design
```
gkalpak, line 41: Does this apply to Angular public API (i.e. stuff on our docs), public API of a class/module (i.e. public methods, exported functions, etc.) or any method/function?

docs/CODING_STANDARDS.md
gkalpak: :+1:
I would add another paragraph discussing `for ... of` vs "plain" `for`.

docs/CODING_STANDARDS.md
```diff
@@ -0,0 +1,249 @@
...
+All public APIs must have user-facing comments. These are extracted for API documentation and shown
+in IDEs.
+
+Private and internal APIs should have JsDoc when they are not obvious. Ultimately it is the purview
+of the code reviewer as to what is "obvious", but the rule of thumb is that *most* classes,
+properties, and methods should have a JsDoc description.
+
+Properties should have a concise description of what the property means:
+```typescript
+  /** The label position relative to the checkbox. Defaults to 'after' */
```
gkalpak, line 116: Is the `Defaults to` part needed when the default value is obvious in TS?

docs/CODING_STANDARDS.md
```diff
@@ -0,0 +1,249 @@
...
+```
+
+Methods blocks should describe what the function does and provide a description for each parameter
+and the return value:
+```typescript
+  /**
+   * Opens a modal dialog containing the given component.
+   * @param component Type of the component to load into the dialog.
+   * @param config Dialog configuration options.
+   * @returns Reference to the newly-opened dialog.
```
gkalpak, line 127: We need a rule that says it must be `@returns` and not `@return`. At all const 😛

docs/CODING_STANDARDS.md
gkalpak: Also, where would `var` be absolutely necessary? (I can't think of any case.)

docs/CODING_STANDARDS.md
```diff
@@ -0,0 +1,249 @@
...
+##### `readonly`
+
+Use `readonly` members wherever possible.
+
+#### Naming
+
+##### General
+
+* Prefer writing out words instead of using abbreviations.
+* Prefer *exact* names over short names (within reason). E.g., `labelPosition` is better than
```
gkalpak, line 162:
```ts
const myCoolComponentFactoryResolvedComponentFactory = componentFactoryResolver.resolveComponentFactory(myCoolComponentFactoryResolvedComponent);
```

😛

docs/CODING_STANDARDS.md
gkalpak, line 168: FWIW, I don't like this rule :cry:

docs/CODING_STANDARDS.md
```diff
@@ -0,0 +1,249 @@
...
+class FormGroupWithCheckboxAndRadios { /* ... */ }
+class InputWithNgModel { /* ... */ }
+
+
+/** AVOID: does not fully describe the scenario under test. */
+class Comp { /* ... */ }
+class InputComp { /* ... */ }
+```
+
+
```
gkalpak, line 214:
```suggestion
```

docs/CODING_STANDARDS.md
```diff
@@ -0,0 +1,249 @@
...
+```typescript
+import {of as observableOf} from 'rxjs';
+```
+
+#### Testing
+
+##### Test names
+
+Use descriptive names for jasmine tests. Ideally, test names should read as a sentence, often of
+the form "it should...".
```
gkalpak, line 229: :100:

docs/CODING_STANDARDS.md
spock123: Very surprised seeing this as well.
I was under impression that was a good practise. 
I'm definitely fond of it.

docs/CODING_STANDARDS.md
mhevery: Could we add a caveat. I think the above is very true for public API, but for private(internal) API we often do boolean arguments for perf reasons and or size. We should make that distinction here as well.

docs/CODING_STANDARDS.md
```diff
@@ -0,0 +1,249 @@
...
+
+function createTargetElement() {
+ // ...
+}
+```
+
+#### Optional arguments
+
+Use optional function arguments only when such an argument makes sense for an API. Don't use
+optional arguments merely for convenience in implementation.
```
mhevery, line 69: NOTE: under some circumstances (When the method can't be inlined, the VM needs to create a trampoline with matching argument sets), option arguments have negative perf impact.

docs/CODING_STANDARDS.md
```diff
@@ -0,0 +1,249 @@
...
+
+#### Typing
+
+Avoid `any` where possible. If you find yourself using `any`, consider whether a generic or
+`unknown` may be appropriate in your case.
+
+#### Getters and Setters
+
+* Only use getters and setters for `@Input` properties or when otherwise required for API
+compatibility.
```
mhevery, line 81:
```suggestion
compatibility. (Using getters/setters implies side-effect, and creates a higher cognitive load to the reader. Use only when it is necessary.)
```

docs/CODING_STANDARDS.md
mhevery: Maybe a link to https://github.com/angular/angular/blob/master/packages/core/src/render3/PERF_NOTES.md

docs/CODING_STANDARDS.md
```diff
@@ -0,0 +1,249 @@
...
+
+Boolean properties and return values should use "Whether..." as opposed to "True if...":
+```ts
+  /** Whether the button is disabled. */
+  disabled: boolean = false;
+```
+
+#### Try-Catch
+
+Avoid `try-catch` blocks, instead preferring to prevent an error from being thrown in the first
```
mhevery, line 140: Not sure I agree, perhaps justification as to what you are trying to avoid would be helpfull. Why is try-catch bad?

mhevery (dismissed):

docs/CODING_STANDARDS.md
dgp1130, line 140: Not to start a bikeshedding war on code style, but the one advantage of throwing errors is that they can contain string messages about what went wrong which can be easily rethrown and composed to maintain useful error messages.

For example, consider the error-less implementation:

```typescript
function computeAThing(): number|null {
  const firstResult = computeFirstPiece();
  if (!firstResult) return null;

  const secondResult = computeFirstPiece();
  if (!secondResult) return null;

  return firstResult + secondResult;
}

console.log(computeAThing()); // If this logs `null`, why?
```

There is no way to know which of the two pieces went wrong. Alternatively consider:

```typescript
function computeAThing(): number {
  let firstResult: number;
  try {
    firstResult = computeFirstPiece();
  } catch (err) {
    throw new Error(`Failed to compute the first piece: ${err.message}`);
  }

  let secondResult: number;
  try {
    secondResult = computeSecondPiece();
  } catch (err) { 
    throw new Error(`Failed to compute the second piece: ${err.message}`);
  }

  return firstResult + secondResult;
}

try {
  console.log(computeAThing());
} catch (err) {
  console.log('Failed to compute a thing: ${err.message}');
}
```

The terrible `try-catch` syntax aside, this has the advantage of more accurately communicating exactly what went wrong at any given point in a computation with composable error messages giving full context about the operation. It also allows types to be much clearer; because `computeAThing()` always returns a `number` there is no ambiguity as to what `null` represents. Errors naturally propagate as well, so omitting a `try-catch` at one level of abstraction will still be caught at a higher level of abstraction.

I do agree that `Error` in JavaScript is not ideal. You can't wrap an `Error` like in other languages, you just have to throw a new `Error` with a concatenated message. `try-catch` syntax is awkward due to scoping and often requires a mutable `let`. It's also a bit of a pain to disambiguate different kinds of errors (though whether that is a good idea is a separate discussion).

My main point here is just that I don't feel we should outright ban them when we lack effective alternative options to fill that need. A contextual error message can be incredibly helpful for debugging problems.

A couple notable alternatives:

"Return an abstract data type (ADT)": As much as I like this pattern in general, I don't think it can be used in public-facing APIs because anyone using them in TypeScript can use an exhaustive switch statement, which means that adding a new failure state is actually a breaking change. Exhaustive switch is a useful feature, but most developers don't see this caveat and tend to misuse it as a result.

"Use a `Result` monad" (a la `Result<T, Err>` in Rust or `Maybe` is Haskell): Personally I think this an ideal solution but it just isn't well supported in the JavaScript ecosystem. We'd really need a standardized form of the type to be able to use it consistently and effectively.

docs/CODING_STANDARDS.md
dgp1130: `var` is useful if you want to declare the value on the global scope:

```javascript
// At root scope:
var foo = 1;
let bar = 2;
console.log(window.foo); // 1
console.log(window.bar); // undefined
```

I agree that this behavior is usually undesired in the first place and is very rarely actually useful. If this is ever needed there should be a sizeable comment next to it explaining exactly why it is necessary.

docs/CODING_STANDARDS.md
```diff
@@ -0,0 +1,249 @@
...
+class InputWithNgModel { /* ... */ }
+
+
+/** AVOID: does not fully describe the scenario under test. */
+class Comp { /* ... */ }
+class InputComp { /* ... */ }
+```
+
+
+#### RxJS
```
dgp1130, line 215: This may not belong in style guide, but I'll put it here just in case:

One rule I always use is prefer `defer()` over `from()` when translating a `Promise` into an `Observable`. Consider:

```typescript
declare function doAsyncThing(): Promise<string>;

function doObservableThing(): Observable<string> {
  return from(doAsyncThing()); // doAsyncThing() is immediately invoked.
}
```

vs:

```
declare function doAsyncThing(): Promise<string>;

function doObservableThing(): Observable<string> {
  return defer(() => from(doAsyncThing())); // doAsyncThing() is invoked on `subscribe()`
}
```

You almost always want the latter option, so I'd love to see a little more awareness about this particular foot-gun. Though a style guide may not be the best place to put it.

docs/CODING_STANDARDS.md
```diff
@@ -0,0 +1,249 @@
...
+
+When importing the `of` function to create an `Observable` from a value, alias the imported
+function as `observableOf`.
+
+```typescript
+import {of as observableOf} from 'rxjs';
+```
+
+#### Testing
+
```
AndrewKushnir, line 225: Just as an idea/proposal: include human-language description of the test in case it has some non-trivial setup or scenario (in a trivial cases it's typically excessive). For example:
```
it('should calculate transitive scopes when overrides are present', () => {
  // Set up an NgModule hierarchy with two modules, A and B, each with their own component.
  // Module B additionally re-exports module A. Also declare two mock components which can be
  // used in tests to verify that overrides within this hierarchy are working correctly.
  TestBed.configureTestingModule({ ... });
  ...
});
```

This is actually @alxhub's idea that he mentioned during the code review for one of my PRs in
https://github.com/angular/angular/pull/36649#discussion_r411570865.

docs/CODING_STANDARDS.md
```diff
@@ -0,0 +1,249 @@
...
+## Code style
+
+The [Google JavaScript Style Guide](https://google.github.io/styleguide/jsguide.html) is the
+basis for Angular's coding style, with additional guidance here pertaining to TypeScript. The team
+uses `clang-format` to automatically format code; automatic formatting is enforced by CI.
+
+## Code practices
+
+#### Write useful comments
```
kapunahelewong, line 11: Did you mean `###` here?

docs/CODING_STANDARDS.md
kapunahelewong, line 162: To avoid Latin abbreviations (and confusion), we've been changing them to their English equivalents: e.g. --> for example

docs/CODING_STANDARDS.md
```diff
@@ -0,0 +1,249 @@
...
+Use `readonly` members wherever possible.
+
+#### Naming
+
+##### General
+
+* Prefer writing out words instead of using abbreviations.
+* Prefer *exact* names over short names (within reason). E.g., `labelPosition` is better than
+`align` because the former much more exactly communicates what the property means.
+* Except for `@Input` properties, use `is` and `has` prefixes for boolean properties / methods.
```
kapunahelewong, line 164:
```suggestion
* Except for `@Input()` properties, use `is` and `has` prefixes for boolean properties / methods.
```

kapunahelewong (commented):
This is great! So excited to see how it comes together. I peeked out of curiosity and just noticed a few little things. I'm happy to do a more thorough docs review down the road too, if you'd like.

docs/CODING_STANDARDS.md
jelbourn: Changed this to be less opinionated. 

Related: I know modern VMs do optimize try-catch blocks, but I'm guessing that `try` does still incur _some_ overhead?

jelbourn: Changed this to be less opinionated.

I was really just trying to get across "Don't use `try` to avoid expected error conditions like array-index-out-of-bounds"

docs/CODING_STANDARDS.md
jelbourn: I removed this bit.

docs/CODING_STANDARDS.md
jelbourn: Updated to include a mention of `let`.

docs/CODING_STANDARDS.md
jelbourn: If I had written `ComponentFactoryResolver`, I would have just called the method `resolve`, since what else is a `ComponentFactoryResolver` resolving if not a component factory? :p

jelbourn: Done

docs/CODING_STANDARDS.md
jelbourn: Done.

docs/CODING_STANDARDS.md
jelbourn: This is here because consistency beats everything else when it comes to coding style. If all of the observables in the framework / router / http client used a `$` suffix from the start, that would have been fine, but it would be bad to go to a state where only _some_ observables have a `$`.

docs/CODING_STANDARDS.md
jelbourn: Done.

docs/CODING_STANDARDS.md
jelbourn: I think this belong here since this doc aims to be "coding practices" and not purely a style guide. I will, however, leave it to you to add in a follow-up PR so you can get the credit 😛

docs/CODING_STANDARDS.md
jelbourn: SGTM, but I will leave it to you to add in a follow-up PR so you can get the credit 😛

jelbourn (commented):
Comments addressed

docs/CODING_STANDARDS.md
gkalpak: Basically, I am fine with either, but we need a rule that says we must be consistent 😛

jelbourn: Agreed (I prefer `@returns` also). I'll leave this for a follow-up PR.

mhevery (approved):

Issue #18030, opened 2017-07-10
aio: configure dgeni to break build if there are "dangling" images


gkalpak:
Example from current output:

```
warn: Unable to load src in image tag `<img src="generated/images/bios/shield-bio-placeholder.png">` - doc "news" (content)  - from file "news.html" - starting at line 1
warn: Unable to load src in image tag `<img src="generated/images/bios/angular-gde-bio-placeholder.png">` - doc "news" (content)  - from file "news.html" - starting at line 1
```

<sub>(These specific errors don't matter, because we don't use the `news` doc any more.)</sub>

petebacondarwin:
Fixed via #18034

PR #37707, opened 2020-06-24
feat(ngcc): support passing a list of entry-points to process

petebacondarwin:
Similar to the targeted entry-point mode, where we can pass in a
single entry-point path as the starting point for processing, this
commit adds support for passing in a list of such target entry-points.
This is achieved by pointing ngcc at a JSON file that holds the list
of paths to the entry-points. These entry-points, and their dependencies,
will then be processed by ngcc.

The path to the file containing the entry-point paths is configured
programmatically by setting the `entryPointListPath` property or via
the command line by the `--entry-points-file` option.


petebacondarwin:
This was a proof-of-concept for pre-parsing the application source-code for entry-points that need to be processed by ngcc.

By doing this inside the CLI, the module-resolution and source-file parsing can be cached (in the CompilerHost) to be used in the actual compilation that would happen later in the build.

We feel that this would have a limited benefit especially if we can avoid running ngcc at all if we believe that the entry-points have already been processed (e.g. #18030).

PR #37708, opened 2020-06-24
refactor(compiler): Update %COMP% string to %C% and updated regex to not search globally.

ngwattcos:
Changed the %COMP% component placeholder in the temporary attribute to %C% to shorten the attribute.

Changed COMPONENT_REGEX in dom_renderer.ts to not search the string globally, since %C% will only appear once in an attribute selector.

Shortening the attribute allows a smaller bundle size and a smaller download before the platform-server shis the component attribute with component ID.

All spec tests involving %COMP% have been updated to use %C%.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ x ] Other... Please describe:
The API is slightly changed because the outputted %COMP% string in angular attributes is now %C%. Tests have been updated and re-run to verify this works.

The component placeholder variable in the angular attribute currently outputs %COMP%.

Issue Number: N/A

## What is the new behavior?
outputted %COMP% string is replaced with %C%.

- [ ] Yes
- [ x ] No


packages/compiler/src/style_compiler.ts
```diff
@@ -6,90 +6,135 @@
...
+  identifierName,
+} from "./compile_metadata";
+import { ViewEncapsulation } from "./core";
+import * as o from "./output/output_ast";
+import { ShadowCss } from "./shadow_css";
+import { UrlResolver } from "./url_resolver";
+import { OutputContext } from "./util";
 
-const COMPONENT_VARIABLE = '%COMP%';
+const COMPONENT_VARIABLE = "%C%";
```
kseamon, line 22: Use single-quote '

(more or less, try to follow the google style guide or at least just match nearby code)
https://google.github.io/styleguide/jsguide.html

packages/platform-browser/src/dom/dom_renderer.ts
```diff
@@ -6,23 +6,32 @@
...
+  svg: "http://www.w3.org/2000/svg",
+  xhtml: "http://www.w3.org/1999/xhtml",
+  xlink: "http://www.w3.org/1999/xlink",
+  xml: "http://www.w3.org/XML/1998/namespace",
+  xmlns: "http://www.w3.org/2000/xmlns/",
 };
 
-const COMPONENT_REGEX = /%COMP%/g;
-const NG_DEV_MODE = typeof ngDevMode === 'undefined' || !!ngDevMode;
+const COMPONENT_REGEX = /%C%/;
```
kseamon, line 31: As I noted in chat, please verify the runtime performance impact of removing /g as well as seeing what difference $ would make (probably with jsperf)

JoostK: I don't see why the `g` flag can be dropped. For a selector such as `one two`, a shimmed selector `one[_ngcontent-%COMP%] two[_ngcontent-%COMP%]` is generated which requires both occurrences of `%COMP%` to be replaced.

For the `HOST_ATTR` and `CONTENT_ATTR` case, we _can_ however update `shimHostAttribute` and `shimContentAttribute` to simply append the component id to the `_nghost-` and `_ngcontent-` strings.

ngwattcos: Right, I misunderstood the regex expression. I'll also update to simply use the component id.

packages/compiler/src/style_compiler.ts
ngwattcos: Yikes, every time I try to save, it auto-formats to a single quote.

PR #37711, opened 2020-06-24
refactor(router): prevent duplicate navigations that can result in infinite loops

atscott:


atscott:
[presubmit](http://test/OCL:318120922:BASE:318120947:1593027894042:444fd90d)

PR #37710, opened 2020-06-24
test(core): introduce microbenchmarks for ngFor

matsko:

PR #37713, opened 2020-06-24
perf(core): add ability to provide a test setup function to benchmarks

atscott:
This commit adds the ability to provide a test setup function to a micro benchmark. With this,
benchmarks can now handle expensive setup operations that may mutate state which needs
to be reset on each iteration. These setups will not be included in profile timings.

I modified this a little from what I had done in #37710 so that the time to run the test setup is tracked separately. If you think modifying the timestamp directly is more understandable (see changes in #37710), I can change it back.

### DIFF
--- a/packages/core/test/render3/perf/micro_bench.ts
+++ b/packages/core/test/render3/perf/micro_bench.ts
@@ -28,17 +28,26 @@ export interface Profile {
   noImprovementCount: number;
 }
 
-export function createBenchmark(benchmarkName: string): Benchmark {
+/**
+ * @param benchmarkName The name for this benchmark, used in the console.log for identification.
+ * @param beforeEachIteration A function that can be used to do the necessary set up for each
+ *     iteration. Because tests may rely on mutating some state, there may be a need to reset that
+ *     state in each iteration. In addition, the time to run the test setup is excluded from the
+ *     performance timing.
+ */
+export function createBenchmark(
+    benchmarkName: string, beforeEachIteration?: () => void): Benchmark {
   const profiles: Profile[] = [];
 
   const benchmark = function Benchmark(profileName: string): Profile {
-    let iterationCounter: number = 0;
-    let timestamp: number = 0;
+    let iterationCounter = 0;
+    let sampleStartTime = 0;
+    let timeToRunTestSetup = 0;
     const profile: Profile = function Profile() {
       if (iterationCounter === 0) {
         let runAgain = false;
         // if we reached the end of the iteration count than we should decide what to do next.
-        if (timestamp === 0) {
+        if (sampleStartTime === 0) {
           // this is the first time we are executing
           iterationCounter = profile.iterationCount;
           runAgain = true;
@@ -47,7 +56,7 @@ export function createBenchmark(benchmarkName: string): Benchmark {
         } else {
           profile.sampleCount++;
           // we came to an end of a sample, compute the time.
-          const duration_ms = performance.now() - timestamp;
+          const duration_ms = (performance.now() - timeToRunTestSetup) - sampleStartTime;
           const iterationTime_ms = duration_ms / profile.iterationCount;
           if (profile.bestTime > iterationTime_ms) {
             profile.bestTime = iterationTime_ms;
@@ -69,9 +78,20 @@ export function createBenchmark(benchmarkName: string): Benchmark {
           }
         }
         iterationCounter = profile.iterationCount;
-        timestamp = performance.now();
+        // If a test setup function was provided, run it before starting the performance timing
+        // (performance.now).
+        if (beforeEachIteration) beforeEachIteration();
+        sampleStartTime = performance.now();
+        timeToRunTestSetup = 0;
         return runAgain;
       } else {
+        // If there is a test setup function provided, we need to run it before starting the next
+        // iteration.
+        if (beforeEachIteration) {
+          const start = performance.now();
+          beforeEachIteration();
+          timeToRunTestSetup += performance.now() - start;
+        }
         // this is the common path and it needs te be quick!
         iterationCounter--;
         return true;


packages/core/test/render3/perf/micro_bench.ts
```diff
@@ -28,17 +28,25 @@ export interface Profile {
...
+ *     Because tests may rely on mutating some state, there may be a need to reset that state in
+ *     each iteration. In addition, the time to run the test setup is excluded from the performance
+ *     timing.
+ */
+export function createBenchmark(benchmarkName: string, testSetup?: () => void): Benchmark {
   const profiles: Profile[] = [];
 
   const benchmark = function Benchmark(profileName: string): Profile {
-    let iterationCounter: number = 0;
-    let timestamp: number = 0;
```
pkozlowski-opensource, line 36: Thnx for removing `number` type 👍

packages/core/test/render3/perf/micro_bench.ts
```diff
@@ -28,17 +28,25 @@ export interface Profile {
...
 
-export function createBenchmark(benchmarkName: string): Benchmark {
+/**
+ * @param benchmarkName The name for this benchmark, used in the console.log for identification.
+ * @param testSetup A function that can be used to do the necessary set up for each iteration.
+ *     Because tests may rely on mutating some state, there may be a need to reset that state in
+ *     each iteration. In addition, the time to run the test setup is excluded from the performance
+ *     timing.
+ */
+export function createBenchmark(benchmarkName: string, testSetup?: () => void): Benchmark {
```
pkozlowski-opensource, line 38: nit: maybe rename `testSetup` => `beforeEach` as this plays the same role as Jasmine's  `beforeEach`, right?

Generally speaking we start to have a bit of confusion here since there are 2 levels (benchmark, profile) and only one function so it is not clear from the name description if this function runs once before all scenarios (profile) or once for a scenario (that is, if the number of runs is noOfIterations or noOfIterations*numberOfProfiles.

Not sure if this is a documentation or naming issue - what do you think?

pkozlowski-opensource (commented):
@atscott I left a couple of comments but more importantly I would love to chat with you about this bench overall - I must say that I find it harder and harder to follow which isn't great since I wanted to keep it dead-simple (so it is "obvious" that nothing funky is going on - TBH I've seen way too much broken perf harness libraries...).

Let's chat offline to see if we can improve things (and if there is anything to improve here in the first place...).

pkozlowski-opensource (approved):
LGTM, thnx for doing the renames.

As discussed yesterday I think that we should simplify the micro-benchmark as it starts to be hard to follow. We can do so in a separate PR, though.

CI is failing atm so it will need cleanup.

packages/core/test/render3/perf/micro_bench.ts
```diff
@@ -69,9 +78,20 @@ export function createBenchmark(benchmarkName: string): Benchmark {
...
+        if (beforeEachIteration) beforeEachIteration();
+        sampleStartTime = performance.now();
+        timeToRunTestSetup = 0;
         return runAgain;
       } else {
+        // If there is a test setup function provided, we need to run it before starting the next
+        // iteration.
+        if (beforeEachIteration) {
+          const start = performance.now();
+          beforeEachIteration();
```
mhevery, line 92: Imagine we have 
```
function underProfile1() {
  return 1 + 2;
}
function underProfile2() {
  return 2 + 1;
}
```

We could pass `underProfile1` to `createBenchmark` something like this:
```
const profile1 = createBenchmark(underProfile, ...);
profile1();
const profile2 = createBenchmark(underProfile, ...);
profile2();

```
But if we did the above the VM would inline first `underProfile1` and run it fast. Then it would get to `underProfile2` and VM would realize that the inlining was wrong and it would deopt. Going forward it would call `underProfile2` rather than inline. This would make it seems that `underProfile2` is slower than `underProfile1` when in reality they are the same.

For the above reason we do this instead:
```
const profile1 = createBenchmark(...);
while(profile1()) {
  1+2;
}
const profile2 = createBenchmark(...);
while(profile1()) {
  2+1;
}
```
Now instead of inlining `underProfile1` into `createBanchmark` the reverse is happening. (`createBanchmark` is inlined into `underProfile1` test) Because of this no deopt will happen and both tests will run same speed.

----

Your change does exactly the above. It causes the `beforeEach` to be inlined into the `createBanchmark` causing a deopt on second invocation. This is a problem.


----

But there is a second problem here. To subtract out cost of `beforeEachIteration` you use `performance.now()` which will only work if `beforeEachIteration` is significantly more expensive than resolution of `performance.now()`. In most cases it will not be which will introduce noise.

This algorithm specifically is designed to be able to measure things which are smaller than resolution of `performance.now()` by iterating many times over the work until we have work larger than `performance.now()` resolution and then dividing by number of iterations. Trying to subtract out `beforeEachIteration` will be futile in that case.

mhevery (changes requested):
Sorry this approach introduces two issues described in more detail in comments:
1. The invocation of `beforeEach` will cause a deopt leading to wrong measurements.
2. The `performance.now()` timer resolution is not good enough to subtract out `beforeEach` 

I don't really know of a good way to solve the two issues.

atscott:
To summarize after our discussion: 

I think we both agreed that the problematic part isn't the deopt, since the deopt happens with the `beforeEachIteration` and the change attempts to subtract the time it takes to run that from the benchmark (so it doesn't matter if it's fast/slow/different on each iteration). However, the real problem is the `performance.now()` resolution - we can't accurately subtract the `beforeEachIteration` runtime because (1) the time to resolve `performance.now` may be significant if the code being benchmarked is faster, and (2) `performance.now` intentionally has some "fuzziness" built in so again, if the benchmark code is very quick, this fuzziness could significantly affect results.

Closing. As a follow up, I need to try rewriting the `ngFor`/`watchCollection` benchmarks to not use the testSetup

PR #37718, opened 2020-06-24
fix(dev-infra): merge script should not always require full repo permissions

devversion:
We recently added OAuth scope checking to the dev-infra Git client
and started leveraging it for the merge script. We set the `repo` scope
as required for running the merge script. We can loosen this requirement
as in the Angular org where the script is consumed, only pull requests on
public repositories are merged through the script.

This should help with reducing the risk with compromised tokens as no
access had to be granted on `repo:invite`, `repo_deployment` etc.

### DIFF
--- a/dev-infra/utils/config.ts
+++ b/dev-infra/utils/config.ts
@@ -21,6 +21,8 @@ export interface GitClientConfig {
   name: string;
   /** If SSH protocol should be used for git interactions. */
   useSsh?: boolean;
+  /** Whether the specified repository is private. */
+  private?: boolean;
 }
 
 /**
--- a/dev-infra/pr/merge/task.ts
+++ b/dev-infra/pr/merge/task.ts
@@ -16,9 +16,6 @@ import {isPullRequest, loadAndValidatePullRequest,} from './pull-request';
 import {GithubApiMergeStrategy} from './strategies/api-merge';
 import {AutosquashMergeStrategy} from './strategies/autosquash-merge';
 
-/** Github OAuth scopes required for the merge task. */
-const REQUIRED_SCOPES = ['repo'];
-
 /** Describes the status of a pull request merge. */
 export const enum MergeStatus {
   UNKNOWN_GIT_ERROR,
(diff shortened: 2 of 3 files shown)


jelbourn (approved):
LGTM

dev-infra/utils/config.ts
```diff
@@ -21,6 +21,8 @@ export interface GitClientConfig {
   name: string;
   /** If SSH protocol should be used for git interactions. */
   useSsh?: boolean;
+  /** Whether the specified repository is private. */
+  private?: boolean;
```
josephperrott, line 25: Are we going further than we need to with private repo support?  Since private repos aren't really something we do.

josephperrott (approved):
LGTM, just one more high level question I wanted to make sure we think through.

dev-infra/utils/config.ts
devversion: That's a good question. The merge script has been designed in a way that no assumptions are made, providing full flexibility and a standalone API so that the script could potentially used in other infrastructure projects. Making the assumption that we only run the tool in public repositories (is very likely) but goes against that principle.

I don't think there are any good arguments to make here (as we don't have any use-cases yet), but conceptually I'd feel unhappy always assuming `public_repo`. The rest of dev-infra also seems generic enough to be used in arbitrary projects (thanks for the configuration files)

devversion: We should probably evaluate how much effort it is to support this option. I can't think of a lot complexity, except for tools relying on Github, where it should be straightforward to distinguish between private and public auth scopes.

josephperrott: Okay, makes sense to me.

Especially for this instance there is not a huge amount of overhead to support both.  I just want to make sure that we don't end up creating a large amount of effort to support something we will never use.

As I said though, in this case its a trivial amount to have to support both cases.

josephperrott:
@devversion Removed the cleanup and merge labels, so that you can mark for merge once you are ready.

AndrewKushnir:
Hi @devversion, this PR had some conflicts with the patch branch, so I merged it to master only. Could you please create a new PR that targets a patch branch? Thank you.

PR #37721, opened 2020-06-25
docs-infra: update deploy-to-firebase.sh script to support v9 multisi…

IgorMinar:
…te setup

v9.angular.io was used to pilot the firebase hosting multisites setup for angular.io.

The deployments so far have been done manually to control the deployment process.

This change, automates the deployment for v9.angular.io so that future deployments can be made from
the CI.

See https://angular-team.atlassian.net/browse/DEV-125 for more info.

In the process of updating the scripts I rediscovered a bug in the deploy-to-firebase.sh script that incorrect compared two numbers as strings. This previously worked correctly because we were comparing single digit numbers. With the release of v10, we now compare 9 > 10 which behaves differently for strings and numbers. The bug was fixed by switching to an arithmetic comparison of the two variables.
    
This bug has been fixed on the master branch but not on the 9.1.x branch. I realized this during the rebase, but found my version to be a bit cleaner, so I kept it.


IgorMinar:
Once this PR is approved, it will need to be cherry-picked into the v9 branch, otherwise it won't autodeploy the next time we push to it (which should be rare, but can still happen).

aio/firebase.json
```diff
@@ -1,6 +1,7 @@
 {
   "hosting": {
     "public": "dist",
+    "target": "aio",
```
gkalpak, line 4: Nit:

```suggestion
    "target": "aio",
    "public": "dist",
```

gkalpak (approved):
:tada:

aio/firebase.json
```diff
@@ -1,7 +1,7 @@
 {
   "hosting": {
-    "public": "dist",
```
IgorMinar, line 3: :) thank you @gkalpak!

PR #37727, opened 2020-06-25
build: upgrade angular build, integration/bazel and @angular/bazel package to rule_nodejs 2.2.0

gregmagolan:
Updates to rules_nodejs 2.2.0. This is the first major release in 7 months and includes a number of features as well
as breaking changes.

Release notes:
https://github.com/bazelbuild/rules_nodejs/releases/tag/2.0.0
https://github.com/bazelbuild/rules_nodejs/releases/tag/2.0.1
https://github.com/bazelbuild/rules_nodejs/releases/tag/2.0.2
https://github.com/bazelbuild/rules_nodejs/releases/tag/2.2.0

Features of note for angular/angular:

* stdout/stderr/exit code capture; this could be potentially be useful

* TypeScript (ts_project); a simpler tsc rule that ts_library that can be used in the repo where ts_library is too
  heavy weight

Breaking changes of note for angular/angular:

* loading custom rules from npm packages: `ts_library` is no longer loaded from `@npm_bazel_typescript//:index.bzl`
  (which no longer exists) but is now loaded from `@npm//@bazel/typescript:index.bzl`

* with the loading changes above, `load("@npm//:install_bazel_dependencies.bzl", "install_bazel_dependencies")` is
  no longer needed in the WORKSPACE which also means that yarn_install does not need to run unless building/testing
  a target that depends on @npm. In angular/angular this is a minor improvement as almost everything depends on @npm.

* @angular/bazel package is also updated in this PR to support the new load location; Angular + Bazel users that
  require it for ng_package (ng_module is no longer needed in OSS with Angular 10) will need to load from
  `@npm//@angular/bazel:index.bzl`. I investigated if it was possible to maintain backward compatability for the old
  load location `@npm_angular_bazel` but it is not since the package itself needs to be updated to load from
  `@npm//@bazel/typescript:index.bzl` instead of `@npm_bazel_typescript//:index.bzl` as it depends on ts_library
  internals for ng_module.

* runfiles.resolve will now throw instead of returning undefined to match behavior of node require

Other changes in angular/angular:

* integration/bazel has been updated to to use both ng_module and ts_libary with use_angular_plugin=true.
  This latter is the recommended way for rules_nodejs users to compile Angular 10 with Ivy. Bazel + Angular ViewEngine is
  supported with @angular/bazel <= 9.0.5 and Angular <= 8. There is still Angular ViewEngine example on rules_nodejs
  https://github.com/bazelbuild/rules_nodejs/tree/stable/examples/angular_view_engine on these older versions but users
  that want to update to Angular 10 and are on Bazel must switch to Ivy and at that point ts_library with
  use_angular_plugin=true is more performant that ng_module. Angular example in rules_nodejs is configured this way
  as well: https://github.com/bazelbuild/rules_nodejs/tree/stable/examples/angular. As an aside, we also have an
  example of building Angular 10 with architect() rule directly instead of using ts_library with angular plugin:
  https://github.com/bazelbuild/rules_nodejs/tree/stable/examples/angular_bazel_architect.

NB: ng_module is still required for angular/angular repository as it still builds ViewEngine & @angular/bazel
also provides the ng_package rule. ng_module can be removed in the future if ViewEngine is no longer needed in
angular repo.

* JSModuleInfo provider added to ng_module. this is for forward compat for future rules_nodejs versions.
  @josephperrott, this touches `packages/bazel/src/external.bzl` which will make the sync to g3 non-trivial.


gregmagolan:
Two fixes to rules_nodejs 2.0.0-rc.0 so far due to bugs that angular repo exposed:
1. https://github.com/bazelbuild/rules_nodejs/pull/1987; linker bug; landed and added test case
2. https://github.com/bazelbuild/rules_nodejs/pull/1991; ts_library + linker bug; pending

jbedard:
@angular/bazel/src/external.bzl also needs a minor change:

```
--- a/node_modules/@angular/bazel/src/external.bzl
+++ b/node_modules/@angular/bazel/src/external.bzl
@@ -2,11 +2,11 @@
 """
 
 load(
-    "@npm_bazel_typescript//internal:build_defs.bzl",
+    "@@npm//@bazel/typescript//internal:build_defs.bzl",
     _tsc_wrapped_tsconfig = "tsc_wrapped_tsconfig",
 )
 load(
-    "@npm_bazel_typescript//internal:common/compilation.bzl",
+    "@@npm//@bazel/typescript//internal:common/compilation.bzl",
     _COMMON_ATTRIBUTES = "COMMON_ATTRIBUTES",
     _COMMON_OUTPUTS = "COMMON_OUTPUTS",
     _DEPS_ASPECTS = "DEPS_ASPECTS",
@@ -14,7 +14,7 @@ load(
     _ts_providers_dict_to_struct = "ts_providers_dict_to_struct",
 )
 load(
-    "@npm_bazel_typescript//internal:ts_config.bzl",
+    "@@npm//@bazel/typescript//internal:ts_config.bzl",
     _TsConfigInfo = "TsConfigInfo",
 )
 load(

```

alexeagle:
🦸‍♀️

petebacondarwin:
@gregmagolan - could you make this PR "draft" so that it doesn't trigger review request from lots of the team?

packages/localize/src/tools/BUILD.bazel
```diff
@@ -1,7 +1,7 @@
 package(default_visibility = ["//visibility:public"])
 
 load("//tools:defaults.bzl", "pkg_npm", "ts_library")
-load("@npm_bazel_typescript//:index.bzl", "ts_config")
+load("@npm//@bazel/typescript:index.bzl", "ts_config")
```
petebacondarwin, line 4: Perhaps we should have a macro for `ts_config` and then we would not have needed to update all these references to `@npm_bazel_typescript`, since the change would have been isolated to the `tools:defaults.bzl` file?

tools/npm_integration_test/test_runner.js
```diff
@@ -188,8 +188,12 @@ class TestRunner {
...
             binary = `../${binary.substring('external/'.length)}`;
           }
-          const runfilesBinary = runfiles.resolveWorkspaceRelative(binary);
-          binary = fs.existsSync(runfilesBinary) ? runfilesBinary : binary;
+          let runfilesBinary;
+          try {
+            runfilesBinary = runfiles.resolveWorkspaceRelative(binary);
+          } catch (e) {
+          }
+          binary = (runfilesBinary && fs.existsSync(runfilesBinary)) ? runfilesBinary : binary;
```
petebacondarwin, line 196: NIT: This could be a bit simpler:

```suggestion
          try {
            const runfilesBinary = runfiles.resolveWorkspaceRelative(binary);
            binary = (runfilesBinary && fs.existsSync(runfilesBinary)) ? runfilesBinary : binary;
          } catch (e) {
          }
```

petebacondarwin (approved):
Reviewed-for: fw-i18n

packages/localize/src/tools/BUILD.bazel
devversion: yeah, I'd like that too. That is one of the primary reasons why we proxy a lot of these things through `defaults.bzl` in the components repo. We should probably do it here too, but maybe as a follow-up.

packages/bazel/BUILD.bazel
```diff
@@ -17,9 +18,10 @@ pkg_npm(
     ],
     substitutions = {
         "(#|\/\/)\\s+BEGIN-DEV-ONLY[\\w\W]+?(#|\/\/)\\s+END-DEV-ONLY": "",
-        "//packages/bazel/src/ngc-wrapped": "@npm//@angular/bazel/bin:ngc-wrapped",
-        "//packages/bazel/": "//",
-        "angular/packages/bazel/": "npm_angular_bazel/",
+        "//packages/bazel/src/ng_package:ng_package.bzl": "//:node_modules/@angular/bazel/src/ng_package/ng_package.bzl",
+        "//packages/bazel/src:ng_module.bzl": "//:node_modules/@angular/bazel/src/ng_module.bzl",
+        "//packages/bazel/src:external.bzl": "//:node_modules/@angular/bazel/src/external.bzl",
+        "//packages/bazel/src/ng_package:collect-type-definitions.bzl": "//:node_modules/@angular/bazel/src/ng_package/collect-type-definitions.bzl",
```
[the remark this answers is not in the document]
devversion: Given the comment below (on `_DEFAULT_ROLLUP` etc.) I think it would be better to always use `@npm//@angular/bazel`. Especially since binaries like `api-extractor` are accessed from there too (consistency might make debugging significantly easier here)

tools/npm_integration_test/test_runner.js
```diff
@@ -361,6 +365,15 @@ log_verbose(`env: ${JSON.stringify(process.env, null, 2)}`);
...
 log(`running in ${process.cwd()}`);
 
+// remove --preserve-symlinks-main from node wrapper as it breaks node_modules/.bin files
+// run by yarn which must be resolved to their module location to work
+const isWindows = process.platform === 'win32';
+const nodePath = runfiles.resolve(
+    `build_bazel_rules_nodejs/internal/node/_node_bin/${isWindows ? 'node.bat' : 'node'}`);
+const nodeContents =
+    fs.readFileSync(nodePath, {encoding: 'utf-8'}).replace(' --preserve-symlinks-main ', ' ');
+fs.writeFileSync(nodePath, nodeContents, 'utf8');
```
devversion, line 375: Does this have the side-effect in non-sandbox environments (i.e. windows) to change the actual `build_bazel_rules_nodejs` repository? If so, that _could_ potentially cause a lot of confusion/debugging resources.

devversion (commented):
A few minor comments, but otherwise this looks great to me!

integration/bazel/tools/angular_ts_library.bzl
```diff
@@ -0,0 +1,11 @@
+"Shows how to enable both worker mode and use_angular_plugin to make a drop-in replacement for ng_module"
+
+load("@npm//@bazel/typescript:index.bzl", "ts_library")
+
+def ng_ts_library(**kwargs):
```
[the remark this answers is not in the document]
josephperrott: I agree that we should continue to use `ng_module` until we transition away from it in our actual builds as well.

But as @devversion mentioned, we should probably have tests set up for both of these.

josephperrott (commented):
Have we confirmed this updates effect on RBE?

integration/bazel/tools/angular_ts_library.bzl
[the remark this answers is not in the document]
gregmagolan: Oh. I thought I switched it in the example. Agreed that we should keep coverage for ng_module but also add coverage for ts_library with use_angular_plugin. The two should be compatible so I'll update the integration test can be to exercise both.

tools/npm_integration_test/test_runner.js
gregmagolan: That is a very good point. With no sandbox this will modify the actual build_bazel_rules_nodejs. I need to think about how to get around this problem some more.

tools/npm_integration_test/test_runner.js
gregmagolan: Nice catch. Thanks.

integration/bazel/tools/angular_ts_library.bzl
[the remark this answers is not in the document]
gregmagolan: Done. integraiton/bazel now tests both ng_module & ts_library with use_angular_plugin. Ivy only.

packages/bazel/src/ng_package/ng_package.bzl
```diff
@@ -29,9 +29,9 @@ def _debug(vars, *args):
         print("[ng_package.bzl]", args)
 
 _DEFAULT_NG_PACKAGER = "@npm//@angular/bazel/bin:packager"
-_DEFAULT_ROLLUP_CONFIG_TMPL = "@npm_angular_bazel//src/ng_package:rollup.config.js"
-_DEFALUT_TERSER_CONFIG_FILE = "@npm_angular_bazel//src/ng_package:terser_config.default.json"
-_DEFAULT_ROLLUP = "@npm_angular_bazel//src/ng_package:rollup_for_ng_package"
+_DEFAULT_ROLLUP_CONFIG_TMPL = "@npm//@angular/bazel/src/ng_package:rollup.config.js"
+_DEFALUT_TERSER_CONFIG_FILE = "@npm//@angular/bazel/src/ng_package:terser_config.default.json"
+_DEFAULT_ROLLUP = "@npm//@angular/bazel/src/ng_package:rollup_for_ng_package"
```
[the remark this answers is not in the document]
gregmagolan: That is correct. We want to remove the hard-coded @npm here. Thanks.

packages/bazel/BUILD.bazel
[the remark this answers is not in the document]
gregmagolan: Yes. We can simplify to one substitution per package given the different location of `:`. Could potentially be done in a single regex with groups but doesn't seem worth the effort.
To your point of original vs. copied .bzl files. The copied one is the one that is typically loaded in user space so I'll change to,
```
        "//packages/bazel/src/ng_package:": "//:@angular/bazel/src/ng_package/",
        "//packages/bazel/src:": "//:@angular/bazel/src/",
```

gregmagolan:
> Have we confirmed this updates effect on RBE?

Doesn't fix OSX cross-platform after the Bazel 3.0 breakage. I need to dig into that at some point and fix the toolchains in Bazel to work with OSX. I presume linux is covered by CI?

tools/npm_integration_test/test_runner.js
gregmagolan: There are likely other issues with npm_integration_test and Windows. I don't think it works on Windows yet anyway so we may just have to detect if OS is WIndows and fail gracefully. However, the issue at hand here is non-sandbox envs which could be OSX & Linux as well if strategy is forced to local. Still thinking about this.

gregmagolan: Looks like that last issue with this upgrade.

JiaLiPassion (approved):
reviewed for @angular/fw-zones , LGTM!

tools/npm_integration_test/test_runner.js
devversion: yeah, that is a good point. I agree with Windows being a good exception here. Also good point that other platforms could run in non-sandbox too.

One option might be to restore the binary file after test completion.. though that still has the risk of never cleaning up I'd guess.

packages/bazel/src/ng_package/ng_package.bzl
```diff
@@ -28,10 +28,10 @@ def _debug(vars, *args):
     if "VERBOSE_LOGS" in vars.keys():
         print("[ng_package.bzl]", args)
 
-_DEFAULT_NG_PACKAGER = "@npm//@angular/bazel/bin:packager"
-_DEFAULT_ROLLUP_CONFIG_TMPL = "@npm_angular_bazel//src/ng_package:rollup.config.js"
-_DEFALUT_TERSER_CONFIG_FILE = "@npm_angular_bazel//src/ng_package:terser_config.default.json"
-_DEFAULT_ROLLUP = "@npm_angular_bazel//src/ng_package:rollup_for_ng_package"
+_DEFAULT_NG_PACKAGER = "//@angular/bazel/bin:packager"
+_DEFAULT_ROLLUP_CONFIG_TMPL = "//:node_modules/@angular/bazel/src/ng_package/rollup.config.js"
+_DEFALUT_TERSER_CONFIG_FILE = "//:node_modules/@angular/bazel/src/ng_package/terser_config.default.json"
```
devversion, line 33: Just coming back to this: Do you think it might make sense to just use `@npm//@angular/bazel` here too? That will make all labels to the Angular Bazel code consistent and the source of truth would be `external/npm/@angular/bazel` instead of a mix?

gregmagolan:
Sorted out the `--preserve-symlinks-main` issue upstream. When this lands https://github.com/bazelbuild/rules_nodejs/pull/2176 and we cut a 2.1.1 then this PR can finally be wrapped up.

tools/npm_integration_test/test_runner.js
gregmagolan: https://github.com/bazelbuild/rules_nodejs/pull/2176  will sort this out upstream

gregmagolan: updating to 2.2.0 which just got released makes this hack no longer necessary :)

packages/localize/src/tools/BUILD.bazel
gregmagolan: good idea. there were only a handful of spots so added that change into this PR. thanks for the suggestion.

packages/bazel/src/ng_package/ng_package.bzl
gregmagolan: Not entirely sure what you mean

gregmagolan: The default ```_DEFAULT_NG_PACKAGER = "//@angular/bazel/bin:packager"``` would be to the `@npm` workspace in user-land. That default is overridden in `/tools/defaults.bzl` to be ```_INTERNAL_NG_PACKAGE_PACKAGER = "//packages/bazel/src/ng_package:packager"``` since `@npm//@angular/bazel` does not exist during the angular build.

devversion: I might be missing something here but my understanding is that there will be two versions of `@angular/bazel`. 

* One in `external/npm/@angular/bazel`
* One in `external/npm/node_modules/@angular/bazel

We rely on the files from the first path in the majority of cases IIRC, but currently the rollup config and terser config are explicitly going with the latter. I'm proposing to use (1) here for consistency. If we have a mix of (1) and (2) then it will be hard to debug things (based on my experience in the past with how e.g. `npm_angular_bazel` has existed as separate repository)

josephperrott (commented):
I think this looks overall good.  I am also curious about Paul's outstanding comment regarding loading via `//@npm` vs `//:node_modules`

tools/npm_integration_test/test_runner.js
```diff
@@ -188,8 +188,11 @@ class TestRunner {
           if (binary.startsWith('external/')) {
             binary = `../${binary.substring('external/'.length)}`;
           }
-          const runfilesBinary = runfiles.resolveWorkspaceRelative(binary);
-          binary = fs.existsSync(runfilesBinary) ? runfilesBinary : binary;
+          try {
+            const runfilesBinary = runfiles.resolveWorkspaceRelative(binary);
+            binary = (runfilesBinary && fs.existsSync(runfilesBinary)) ? runfilesBinary : binary;
+          } catch (e) {
```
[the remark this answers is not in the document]
gregmagolan: The intention is that if `runfiles.resolveWorkspaceRelative(binary);` throws then the binary specified by the user is just a built-in command such as `mkdir` or `rm`.
For example,
```
commands = [
        "rm -rf ./.yarn_local_cache",
        "mkdir .yarn_local_cache",
        "$(rootpath @nodejs//:yarn_bin) install --cache-folder ./.yarn_local_cache",
        "$(rootpath @nodejs//:yarn_bin) test",
]
```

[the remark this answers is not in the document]
gregmagolan: resolveWorkspaceRelative would resolve for something like `$(rootpath @nodejs//:yarn_bin)`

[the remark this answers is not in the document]
gregmagolan: I'll add a comment so this is clearer

josephperrott (approved):
LGTM

josephperrott (changes requested):

josephperrott:
Caretaker Note: The local changes required `JSModuleInfo` in `ng_module` for g3 have already landed in google.

packages/bazel/src/ng_module.bzl
```diff
@@ -6,7 +6,7 @@
 """
 
 load(
-    ":external.bzl",
+    "//packages/bazel/src:external.bzl",
```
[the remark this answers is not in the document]
gregmagolan: that change doesn't look necessary; put it back to `":external.bzl"`

gregmagolan:
Rebased this and resolved a few inflight build failures with bazel load statements. I'll keep an eye on it and rebase it periodically as needed until it lands.

packages/bazel/src/ng_package/ng_package.bzl
josephperrott: Can you confirm our understanding of this?

AndrewKushnir (approved):
Reviewed-for: fw-testing

packages/bazel/src/ng_package/ng_package.bzl
gregmagolan: Okay. I think I understand what @devversion was getting at. It is a good question.

I suspect the better pattern is to use .js and other files required for rules from their installed node_modules location in @npm. We've discussed changing rules_nodejs to _only_ copy over .bzl and BUILD.bazel from `external/npm/node_modules/@scope/package` to `external/npm/@scope/package` instead of copying _all_files as it currently does. Not having duplicate copies would remove the ambiguity for debugging @devversion alluded to. Keeping this as

```
_DEFAULT_ROLLUP_CONFIG_TMPL = "//:node_modules/@angular/bazel/src/ng_package/rollup.config.js"
_DEFALUT_TERSER_CONFIG_FILE = "//:node_modules/@angular/bazel/src/ng_package/terser_config.default.json"
```

as it currently is would be forward compatible with that change.

Actually, to be completely portable, there can be no references to `@npm` in the published package. The pattern for @angular/bazel should be

```
            # BEGIN-INTERNAL
            "@npm" +
            # END-INTERNAL
            "//label:path"
```

otherwise it assumes @npm is the name of the external repository for the end-user.

I will add that change right now. Its a quick one.

josephperrott (approved):
LGTM

PR #37732, opened 2020-06-25
fix(migrations): do not incorrectly add todo for @Injectable or @Pipe

### LINKED ISSUE #37726: @angular/core update schematic v9 to v10 adds odd todo comments in a correct-looking file
# 🐞 bug report

### Affected Package

### Description

`ng update @angular/core` from v9 to v10 adds multiple comments that tell me I need to change something which I did not expect

```
> Undecorated classes with Angular features migration.
  In version 10, classes that use Angular features and do not have an Angular decorator are no longer supported.
  Read more about this here: https://v10.angular.io/guide/migration-undecorated-classes
    Could not migrate all undecorated classes that use Angular features.
    Please manually fix the following failures:
    ⮑   projects\client-core\src\lib\i18n\model-i18n\translateField.pipe.ts@6:1: Class uses Angular features but cannot be migrated automatically. Please add an appropriate Angular decorator.

```

The code looks like this

```typescript
import { AccessPathLike } from '@aeb/graphql-accessor';
import { ChangeDetectorRef, OnDestroy, Pipe, PipeTransform } from '@angular/core';
import { Subscription } from 'rxjs';
import { ModelLocalizationService } from './model-localization.service';

// TODO: Add Angular decorator.
// TODO: Add Angular decorator.
// TODO: Add Angular decorator.
// TODO: Add Angular decorator.
// TODO: Add Angular decorator.
// TODO: Add Angular decorator.
@Pipe({
    name: 'translateField',
    pure: false
})
export class TranslateFieldPipe implements PipeTransform, OnDestroy {
    private subscription: Subscription | undefined = undefined;
    // don't initialize with undefined because then the pipe might break consumers expecting string
    private currentValue: string = '';
    private currentPath: AccessPathLike | undefined = undefined;

    constructor(
        private readonly modelLocalizationService: ModelLocalizationService,
        private readonly changeDetectorRef: ChangeDetectorRef
    ) {
    }

    ngOnDestroy() {
        if (this.subscription) {
            this.subscription.unsubscribe();
            this.subscription = undefined;
        }
    }

    transform(value: AccessPathLike|undefined): string {
        if (value === this.currentPath) {
            return this.currentValue;
        }
        this.currentPath = value;
        if (this.subscription) {
            this.subscription.unsubscribe();
            this.subscription = undefined;
        }
        if (!value) {
            return '';
        }
        this.subscription = this.modelLocalizationService.watchFieldLabel(value).subscribe(label => this.updateValue(label));
        return this.currentValue;
    }

    private updateValue(value: string) {
        this.currentValue = value;
        this.changeDetectorRef.markForCheck();
    }
}
```

It already has a decorator. I assume this is a bug and I can't do anything to satisfy ng update @angular/core ?

## 🔬 Minimal Reproduction
I'm not sure if the above example alone is enough for a reproduction on your side. Let me know if you need more!

## 🌍  Your Environment

```
                |___/
    

Angular CLI: 10.0.0
Node: 10.19.0
OS: win32 x64

Angular: 10.0.0
Ivy Workspace: No

@angular-devkit/architect          0.900.7
@angular-devkit/build-angular      0.1000.0
@angular-devkit/build-ng-packagr   0.1000.0
@angular-devkit/build-optimizer    0.1000.0
@angular-devkit/build-webpack      0.1000.0
@angular-devkit/core               10.0.0
@angular-devkit/schematics         10.0.0
@angular/flex-layout               10.0.0-beta.32
@ngtools/webpack                   10.0.0
@schematics/angular                10.0.0
@schematics/update                 0.1000.0
ng-packagr                         10.0.0
typescript                         3.9.5
webpack                            4.43.0
```

devversion:
As of v10, the `undecorated-classes-with-decorated-fields` migration
generally deals with undecorated classes using Angular features. We
intended to run this migation as part of v10 again as undecorated
classes with Angular features are no longer supported in planned v11.

The migration currently behaves incorrectly in some cases where an
`@Injectable` or `@Pipe` decorated classes uses the `ngOnDestroy`
lifecycle hook. We incorrectly add a TODO for those classes. This
commit fixes that.

Additionally, this change makes the migration more robust to
not migrate a class if it inherits from a component, pipe
injectable or non-abstract directive. We previously did not
need this as the undecorated-classes-with-di migration ran
before, but this is no longer the case.

Last, this commit fixes an issue where multiple TODO's could be
added. This happens when multiple Angular CLI build targets have
an overlap in source files. Multiple programs then capture the
same source file, causing the migration to detect an undecorated
class multiple times (i.e. adding a TODO twice).

Fixes #37726.

### DIFF
--- a/packages/core/schematics/test/helpers.ts
+++ b/packages/core/schematics/test/helpers.ts
@@ -6,9 +6,13 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
+/** Regular expression that matches empty line whitespace. */
+const omitEmptyLineWhitespaceRegex = /^[ \t]$/gm;
+
 /**
  * Template string function that can be used to dedent the resulting
  * string literal. The smallest common indentation will be omitted.
+ * Additionally, whitespace in empty lines is removed.
  */
 export function dedent(strings: TemplateStringsArray, ...values: any[]) {
   let joinedString = '';
@@ -24,5 +28,6 @@ export function dedent(strings: TemplateStringsArray, ...values: any[]) {
 
   const minLineIndent = Math.min(...matches.map(el => el.length));
   const omitMinIndentRegex = new RegExp(`^[ \\t]{${minLineIndent}}`, 'gm');
-  return minLineIndent > 0 ? joinedString.replace(omitMinIndentRegex, '') : joinedString;
+  const result = minLineIndent > 0 ? joinedString.replace(omitMinIndentRegex, '') : joinedString;
+  return result.replace(omitEmptyLineWhitespaceRegex, '');
 }
--- a/packages/core/schematics/test/helpers.ts
+++ b/packages/core/schematics/test/helpers.ts
@@ -6,9 +6,6 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-/** Regular expression that matches empty line whitespace. */
-const omitEmptyLineWhitespaceRegex = /^[ \t]$/gm;
-
 /**
  * Template string function that can be used to dedent the resulting
  * string literal. The smallest common indentation will be omitted.
@@ -28,6 +25,7 @@ export function dedent(strings: TemplateStringsArray, ...values: any[]) {
 
   const minLineIndent = Math.min(...matches.map(el => el.length));
   const omitMinIndentRegex = new RegExp(`^[ \\t]{${minLineIndent}}`, 'gm');
+  const omitEmptyLineWhitespaceRegex = /^[ \t]+$/gm;
   const result = minLineIndent > 0 ? joinedString.replace(omitMinIndentRegex, '') : joinedString;
   return result.replace(omitEmptyLineWhitespaceRegex, '');
 }
--- a/integration/ng_update_migrations/src/app/migration-tests/undecorated-classes-with-fields.ts
+++ b/integration/ng_update_migrations/src/app/migration-tests/undecorated-classes-with-fields.ts
@@ -4,8 +4,10 @@ import {
   ElementRef,
   HostBinding,
   HostListener,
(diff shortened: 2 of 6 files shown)


packages/core/schematics/test/helpers.ts
```diff
@@ -6,9 +6,13 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
+/** Regular expression that matches empty line whitespace. */
+const omitEmptyLineWhitespaceRegex = /^[ \t]$/gm;
```
crisbeto, line 10: Having this in a global variable is dangerous, because global regexes maintain state between usages. Should it be moved inside the function where it's used?

crisbeto (approved):
LGTM, aside from one comment.

packages/core/schematics/test/helpers.ts
devversion: good catch. looks like I messed that one up when I copied it from the regex in the function. It should also be accept multiple spaces and tabs.

mhevery (commented):
LGTM for `integration/**`
CLENUP:
- [x] Please add comment explaining the tests. Not obvious that no comment is what we are asserting.

mhevery (approved):
Reviewed-for: integration-test
CLEANUP:
- [x] Please add comment explaining the tests. Not obvious that no comment is what we are asserting.

mhevery (approved):

mhevery:
[presubmit](http://test/OCL:318321842:BASE:318321892:1593112482712:e14329c0)

AndrewKushnir:
FYI, presubmit looks good (only unrelated failures), so I'm forcing "green" g3 status. Thank you.

PR #37737, opened 2020-06-25
fix(dev-infra): support running scripts from within a detached head

devversion:
Scripts provided in the `ng-dev` command might use local `git`
commands. For such scripts, we keep track of the branch that
has been checked out before the command has been invoked.

We do this so that we can later (upon command completion)
restore back to the original branch. We do not want to
leave the Git repository in a dirty state.

It looks like this logic currently only deals with branches
but does not work properly when a command is invoked from
a detached head. We can make it work by just checking out
the previous revision (if no branch is checked out).

### DIFF
--- a/dev-infra/utils/git/index.ts
+++ b/dev-infra/utils/git/index.ts
@@ -119,9 +119,16 @@ export class GitClient {
     return this.run(['branch', branchName, '--contains', sha]).stdout !== '';
   }
 
-  /** Gets the currently checked out branch. */
-  getCurrentBranch(): string {
-    return this.run(['rev-parse', '--abbrev-ref', 'HEAD']).stdout.trim();
+  /** Gets the currently checked out branch or revision. */
+  getCurrentBranchOrRevision(): string {
+    const branchName = this.run(['rev-parse', '--abbrev-ref', 'HEAD']).stdout.trim();
+    // If no branch name could be resolved. i.e. `HEAD` has been returned, then Git
+    // is currently in a detached state. In those cases, we just want to return the
+    // currently checked out revision/SHA.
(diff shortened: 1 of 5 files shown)


dev-infra/utils/git/index.ts
```diff
@@ -119,9 +119,16 @@ export class GitClient {
...
-  /** Gets the currently checked out branch. */
-  getCurrentBranch(): string {
-    return this.run(['rev-parse', '--abbrev-ref', 'HEAD']).stdout.trim();
+  /** Gets the currently checked out branch or revision. */
+  getCurrentBranchOrRevision(): string {
+    const branchName = this.run(['rev-parse', '--abbrev-ref', 'HEAD']).stdout.trim();
+    // If no branch name could be resolved. i.e. `HEAD` has been returned, then Git
+    // is currently in a detached state. In those cases, we just want to return the
+    // currently checked out revision/SHA.
+    if (branchName === 'HEAD') {
```
josephperrott, line 128: Should we also have a case for `FETCH_HEAD`?

devversion: I don't think there is any special handling needed. When `FETCH_HEAD` is checked out, `rev-parse --abbrev-ref HEAD` still returns `HEAD`. The logic then correctly determines the SHA then that can later be used to checkout the fetched Git revision (as long as it did not expire; which we cannot control; or guarantee)

josephperrott: Okay, just wanted to confirm.  I know there is differences in what is being referenced, but as you noted that once `FETCH_HEAD` is checked out, it becomes `HEAD` anyway.

josephperrott (approved):
LGTM

PR #37739, opened 2020-06-25
fix(ngcc): ensure lockfile is removed when analyzeFn fails

petebacondarwin:
Previously an error thrown in the `analyzeFn` would cause
the ngcc process to exit immediately without removing the
lockfile, and potentially before the unlocker process had been
successfully spawned resulting in the lockfile being orphaned
and left behind.

Now we catch these errors and remove the lockfile as needed.

### DIFF
--- a/packages/compiler-cli/ngcc/src/execution/cluster/executor.ts
+++ b/packages/compiler-cli/ngcc/src/execution/cluster/executor.ts
@@ -28,13 +28,13 @@ export class ClusterExecutor implements Executor {
 
   async execute(analyzeEntryPoints: AnalyzeEntryPointsFn, _createCompileFn: CreateCompileFn):
       Promise<void> {
-    return this.lockFile.lock(() => {
+    return this.lockFile.lock(async () => {
       this.logger.debug(
           `Running ngcc on ${this.constructor.name} (using ${this.workerCount} worker processes).`);
       const master = new ClusterMaster(
           this.workerCount, this.fileSystem, this.logger, this.fileWriter, this.pkgJsonUpdater,
           analyzeEntryPoints, this.createTaskCompletedCallback);
-      return master.run();
+      return await master.run();
     });
   }
 }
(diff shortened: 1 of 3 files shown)


packages/compiler-cli/ngcc/src/execution/cluster/executor.ts
```diff
@@ -28,13 +28,13 @@ export class ClusterExecutor implements Executor {
...
       Promise<void> {
-    return this.lockFile.lock(() => {
+    return this.lockFile.lock(async () => {
       this.logger.debug(
           `Running ngcc on ${this.constructor.name} (using ${this.workerCount} worker processes).`);
       const master = new ClusterMaster(
           this.workerCount, this.fileSystem, this.logger, this.fileWriter, this.pkgJsonUpdater,
           analyzeEntryPoints, this.createTaskCompletedCallback);
-      return master.run();
+      return await master.run();
```
gkalpak, line 37: Why does this make a difference? 😕

gkalpak (approved):

packages/compiler-cli/ngcc/src/execution/cluster/executor.ts
petebacondarwin: It is the `async` at the top of the function that makes the difference. Because of that, when the `new ClusterMaster()` throws an error it gets caught and turned into a rejected promise.
We need the `await` on the `return` statement to ensure that the resolved value returned from `master.run()` gets added to the promise returned from this function (otherwise it will be `undefined`).

gkalpak: Nevermind. I realized it's the addition of `async` that fixes the issue. The `await` shouldn't be necessary (with the current implementation of `master.run()`), but doesn't hurt and it's safer anyway :+1:

JoostK (approved):

PR #37747, opened 2020-06-25
fix(compiler-cli): use `null` in type checking of safe navigation operators

### LINKED ISSUE #37622: optional chaining in template uses 'null', but strictTemplates treats it as 'undefined'
# 🐞 bug report

### Affected Package

 The issue is caused by package @angular/compiler or @angular/language-service?

### Is this a regression?

no

### Description

Using optional chaining (or safe navigation operator) in templates uses `null`, but the language-service and compiler (with strictTemplates) treat it as `undefined`.

This is not a request to change `null` to `undefined` (that's proposed by #34385). Instead I want the static types to be correct so I'm not surprised by the different semantics of Angular's safe navigation operator compared to TypeScript/ECMAScript.

## 🔬 Minimal Reproduction
https://stackblitz.com/edit/angular-ivy-hh5mwz?file=src%2Fapp%2Fapp.component.ts

## 🔥 Exception or Error
<pre><code>

</code></pre>

## 🌍  Your Environment

**Angular Version:**
<pre><code>
9.1.11

</code></pre>

**Anything else relevant?**

JoostK:
Safe navigation operators in Angular templates will evaluate to `null` if the LHS
operand is `null` or `undefined`, however the type checker would use `undefined`
instead. This results in inaccurate type checking when `strictNullChecks` is enabled.

BREAKING CHANGE:

Using the safe navigation operator in templates will now infer the type of the
expressions to be a union type including `null`, instead of `undefined` as it
previously was. If TypeScript's `strictNullChecks` is enabled together with Angular's
`strictTemplates` or `strictSafeNavigationTypes` you may start to see new type errors.
The previous behavior of including `undefined` in the type does not correspond with
the behavior at runtime, where the `null` value is used instead.

Fixes #37622

### DIFF
--- a/packages/compiler-cli/src/ngtsc/typecheck/src/expression.ts
+++ b/packages/compiler-cli/src/ngtsc/typecheck/src/expression.ts
@@ -248,7 +248,7 @@ class AstTranslator implements AstVisitor {
       const method = ts.createPropertyAccess(ts.createNonNullExpression(receiver), ast.name);
       addParseSpanInfo(method, ast.nameSpan);
       const call = ts.createCall(method, undefined, args);
-      node = ts.createParen(ts.createConditional(NULL_AS_ANY, call, UNDEFINED));
+      node = ts.createParen(ts.createConditional(NULL_AS_ANY, call, ts.createNull()));
     } else if (VeSafeLhsInferenceBugDetector.veWillInferAnyFor(ast)) {
       // "a?.method(...)" becomes (a as any).method(...)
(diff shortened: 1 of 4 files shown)


alfaproject:
Shouldn't TypeScript handle this operator now? As in, shouldn't this have the same semantics as the JS version?

JoostK:
We have decided not to land this fix, as it brings us further away from ES2020 optional chaining semantics, which would require the opposite breaking change in the future. No timeline is currently know for this effort.

PR #37754, opened 2020-06-25
fix(docs-infra): fix typo in the deploy-to-firebase.sh script

IgorMinar:
This typo caused the script to fail on Linux (interestingly it works fine on Mac).

This is a painful reminder that we should not write any more Bash scripts EVER. shelljs FTW! :-)


IgorMinar:
@gkalpak fyi :-) (no more bash scripts, promise?)

AndrewKushnir (approved):
LGTM 👍 Thanks for the fix @IgorMinar!

PR #37763, opened 2020-06-25
test(compiler-cli): disable DynamicValue diagnostic tests on Windows

alxhub:
This commit disables all diagnostic tests for DynamicValue diagnostics which
make assertions about the diagnostic filename while running tests on Windows.

Such assertions are currently suffering from a case sensitivity issue.


AndrewKushnir (approved):
LGTM

PR #37764, opened 2020-06-26
dev-infra: exclude docs from minBodyLength requirement

IgorMinar:
see individual commits

This change was discussed and agreed upon by the docs stakeholders.


josephperrott (approved):
LGTM

Nit: in one of the commit messages it says `mibBodyLength` rather than `minBodyLength`

aikithoughts (approved):
reviewed-for: global-docs-approvers

I think this change is excellent. For lo! While 100 characters may seem like but a trifle, for some the cacophony of characters causes great distress!

PR #37767, opened 2020-06-26
fix(http): Add support for requesting body as URLSearchParams type

sera-wi:
There is no support for URLSearchParams type request in the current version. This commit adds serialize body function for URLSearchParams type.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: 36317

There is no support for URLSearchParams type request in the current version. When a user passes the body parameter with this type, Angular considers it as a random object. It ends up being omitted. 

## What is the new behavior?
This commit adds serialization for URLSearchParams type. It will handle URLSearchParams type request, it will not be omitted.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/common/http/src/request.ts
+++ b/packages/common/http/src/request.ts
@@ -70,7 +70,7 @@ function isFormData(value: any): value is FormData {
  *
  * In some execution environments URLSearchParams is not defined.
  */
-function isURLSearchParams(value: any): value is URLSearchParams {
+function isURLSearchParams(value: unknown): value is URLSearchParams {
   return typeof URLSearchParams !== 'undefined' && value instanceof URLSearchParams;
 }
 
--- a/packages/common/http/src/request.ts
+++ b/packages/common/http/src/request.ts
@@ -66,6 +66,15 @@ function isFormData(value: any): value is FormData {
 }
 
 /**
+ * Safely assert whether the given value is a URLSearchParams.
+ *
+ * In some execution environments URLSearchParams is not defined.
+ */
+function isURLSearchParams(value: any): value is URLSearchParams {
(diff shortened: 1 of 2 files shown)


packages/common/http/src/request.ts
```diff
@@ -65,6 +65,15 @@ function isFormData(value: any): value is FormData {
   return typeof FormData !== 'undefined' && value instanceof FormData;
 }
 
+/**
+ * Safely assert whether the given value is a URLSearchParams.
+ *
+ * In some execution environments URLSearchParams is not defined.
+ */
+function isURLSearchParams(value: any): value is URLSearchParams {
```
alfaproject, line 73:
```suggestion
function isURLSearchParams(value: unknown): value is URLSearchParams {
```

petebacondarwin (changes requested):
Hi @sera0731 - thanks for sending in this PR.

I note that @ajitsinghkaler also created a PR at #37852. But since yours came first I would like to help you to get this merged if you are still interested.

I think that serializing the value to a string is not the correct approach. Instead I think we should do similar to the other PR where we treat it like a special type that just gets passed through - like Blob and FormData.  The gives the most flexibility to the caller to then use the `URLSearchParams` API to extract what they want.

Can you update your code to do this and change the test accordingly?

alxhub:
👍 to @petebacondarwin's requests.

Additionally, this PR is currently 4 separate commits - it should probably be squashed into a single commit with the full implementation, and a commit message describing the changes being made and why in more detail.

petebacondarwin:
Reverting to #37852 due to a lack of response.

PR #37773, opened 2020-06-26
fix(core): incorrectly validating properties on ng-content and ng-container

crisbeto:
Fixes the following issues related to how we validate properties during JIT:
- The invalid property warning was printing `null` as the node name for `ng-content`. The problem is that when generating a template from `ng-content` we weren't capturing the node name.
- We weren't running property validation on `ng-container` at all. This used to be supported on ViewEngine and seems like an oversight.

In the process of making these changes, I found and cleaned up a few places where we were passing in `LView` unnecessarily.

### DIFF
--- a/packages/compiler-cli/test/compliance/r3_compiler_compliance_spec.ts
+++ b/packages/compiler-cli/test/compliance/r3_compiler_compliance_spec.ts
@@ -1551,6 +1551,43 @@ describe('compiler compliance', () => {
         const result = compile(files, angularFiles);
         expectEmit(result.source, SimpleComponentDefinition, 'Incorrect MyApp definition');
       });
+
+      it('should capture the node name of ng-content with a structural directive', () => {
+        const files = {
+          app: {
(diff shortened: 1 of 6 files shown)


alxhub (approved):

alxhub:
[Presubmit](http://test/OCL:320042899:BASE:320042983:1594151851978:683097a6)

AndrewKushnir:
FYI, I've started a [new presubmit](http://test/OCL:321176222:BASE:321176269:1594745172119:7705f4fb) (the previous one had some flakes).

@crisbeto could you please also rebase this PR (to re-run CI with the most recent version of changes in master branch)?

crisbeto:
Rebased.

AndrewKushnir:
@crisbeto presubmit looks good, but the `lint` CI job started to fail after rebasing. Could you please update the commit message?

crisbeto:
Fixed the lint issue.

PR #37782, opened 2020-06-26
test(compiler-cli): re-enable dynamic value diagnostic tests on Windows CI

JoostK:
Essentially reverts #37763 and fixes the test + emulation to allow tests to pass on Windows.

### DIFF
--- a/packages/compiler-cli/src/ngtsc/annotations/test/diagnostics_spec.ts
+++ b/packages/compiler-cli/src/ngtsc/annotations/test/diagnostics_spec.ts
@@ -6,99 +6,92 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {platform} from 'os';
 import * as ts from 'typescript';
 
-import {FatalDiagnosticError} from '../../diagnostics';
-import {absoluteFrom as _} from '../../file_system';
+import {absoluteFrom as _, getSourceFileOrError} from '../../file_system';
 import {runInEachFileSystem, TestFile} from '../../file_system/testing';
 import {PartialEvaluator} from '../../partial_evaluator';
 import {TypeScriptReflectionHost} from '../../reflection';
 import {getDeclaration, makeProgram} from '../../testing';
 import {createValueHasWrongTypeError} from '../src/diagnostics';
 
-runInEachFileSystem(os => {
+runInEachFileSystem(() => {
   describe('ngtsc annotation diagnostics', () => {
-    // These tests are currently disabled when running in Windows mode as the assertions involving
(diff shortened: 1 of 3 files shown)


alxhub (approved):

alxhub:
[Presubmit](http://test/OCL:320637399:BASE:320637564:1594404783539:77f665ff)

alxhub:
I actually suspect we might not have caught the full root cause here.

I've seen failures that indicate `absoluteFrom()` does not always return the same path for different inputs that map to the same file. In particular, I suspect that on Windows, the native implementations of `absoluteFrom("C:\file.txt")` and `absoluteFrom("c:\file.txt")` will effectively result in these being identity operations, which end up returning two different `AbsoluteFsPath`s for the same file. This is probably okay in production executions, but can break tests which assert the equality of paths.

We should fix this bug in the `NodeJsFileSystem` by normalizing the case of the drive letter on Windows as part of `normalize()`.

alxhub:
[Presubmit](https://test.corp.google.com/ui#id=OCL:320637399:BASE:324698270:1596494334799:edec2b49)

JoostK:
Actually, marking as blocked on #38249 because I know these two conflict.

AndrewKushnir:
[Presubmit](http://test/OCL:332863193:BASE:332863204:1600704659908:9ca8da18).

AndrewKushnir:
FYI, there were no targets affected by this change in google3. The `packages/compiler-cli/src/ngtsc/file_system/testing/src/mock_file_system_windows.ts` is the only file that was synced. I'm wondering if we actually should just exclude the `packages/compiler-cli/src/ngtsc/file_system/testing` folder (and may be other `testing` folders) from g3 sync as these files are most likely unused...

JoostK:
@AndrewKushnir I was wondering the same thing. But are you saying that the `//packages/compiler-cli/src/ngtsc/filesystem/testing` target is not synced at all?! How do the tests run in g3 then? The testing target is marked as `//visibility:public` so I assumed it could have been used from within g3 (but maybe I am misunderstanding how this all works).

AndrewKushnir:
@JoostK we do sync the `testing` folder, but there are no targets in g3 that rely on it (so running a presubmit was effectively a noop). It looks like the visibility was set to `public` so that it can be shared between our packages (`ngcc` uses it), but I'm not sure these utility functions/classes should be used outside of the `ngtsc` and `ngcc` packages (for ex. in app testing code).

This is not a blocker/issue for this PR (we can take care of this later), I just wanted to mention that in case we can stop syncing unnecessary files in g3 and avoid the need to run presubmits for them (to land some PRs faster).

PR #36578, opened 2020-04-11
fix(core): do not trigger CSP alert/report in Firefox and Chrome

peruukki:
Default to using DOMParser if it is available and fall back to createDocument if needed. This is the approach used by DOMPurify and suggested in the related Angular.js pull request angular/angular.js#17013. It also safely avoids using an inline style tag that causes CSP violation errors if inline CSS is prohibited.

The related unit tests in `html_sanitizer_spec.ts`, "should not allow JavaScript execution when creating inert document" and "should not allow JavaScript hidden in badly formed HTML to get through sanitization (Firefox bug)", are left untouched to assert that the behavior hasn't changed in those scenarios.

Fixes #25214.

## PR Checklist
Please check if your PR fulfills the following requirements:

- [X] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [X] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] angular.io application / infrastructure changes
- [ ] Other... Please describe:

## What is the current behavior?

If [innerHTML] is used in a component and the Content-Security-Policy does not allow inline styles, a CSP violation will be reported by at least Firefox and Chrome.

Issue Number: #25214 

## What is the new behavior?

No CSP violation is reported.

## Does this PR introduce a breaking change?

- [ ] Yes
- [X] No

## Other information

Like mentioned above, there is an existing unit test related to this change that tests that badly-formatted HTML is sanitized properly, and I didn't change the test because the behavior shouldn't have changed on that regard. I don't think there is a convenient way to test CSP violations, so I didn't write any new tests.

This is my first contribution attempt for Angular. 😃

PR #37784, opened 2020-06-26
ci: decrease payload size limit for integration tests

AndrewKushnir:
This commit updates the payload size limit for the `hello_world` test app built using Closure. This is likely an effect of the changes in https://github.com/angular/angular/pull/36578 (that reduces the bundle size for most of the apps) and additional changes in subsequent commits.

## PR Type

- [x] CI related changes

- [ ] Yes
- [x] No


jelbourn (approved):
LGTM

Issue #25355, opened 2018-08-07
ViewEncapsulation.Emulated and validator.w3.org html validity results


fx772:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Performance issue
[x] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead see https://github.com/angular/angular/blob/master/CONTRIBUTING.md#question
[ ] Other... Please describe:
</code></pre>

## Current behavior

Without specifying ViewEncapsulation.None in my components, html elements will be rendered with _ng-content-c* tags. The company I work for relies on https://validator.w3.org/ for validity of the webpage. When inserting the source of my html5 angular based webpage, a lot of errors show up.

For example:

`Error: Attribute _ngcontent-c1 not allowed on element header at this point.
(...) <header _ngcontent-c1=""><div  (...)`

I really like the ViewEncapsulation.Emulated(default) functionality, but can't seem to use it to adhere to the W3 rules.

## Expected behavior
Output valid html.
Im not sure if this can be fixed, tags starting with "data-" are allowed and could help with html validity. However, I might be missing a key point or design strategy of the angular devs that made this decision, in that case please let me know.

## Minimal reproduction of the problem with instructions

Compile an Angular (v6.0.7) app and inspect the generated html.

## What is the motivation / use case for changing the behavior?

Output valid html.

## Environment

<pre><code>
Angular version: 6.0.7

Browser:
- [x] Chrome (desktop) version 67.0.3396.99
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX
 
For Tooling issues:
- Node version: 8.11.3
- Platform:  Windows

Others:

</code></pre>

trotyl:
Duplicate of https://github.com/angular/angular/issues/15891

mlc-mlapis:
@fx772 ... how do you validate it? Just manual copy / paste from run-time maintained DOM via JS? What is the sense of it?

fx772:
My bad.. its a [duplicate question](https://github.com/angular/angular/issues/15891). However, the answers on the OP's question weren't really enlightening enough for me unfortunately.

[...]

PR #37786, opened 2020-06-27
refactor(compiler): shorten _nghost- and _ngcontent- prefixes to _h- and _c-

ngwattcos:
This replaces all occurrences of _nghost- to _h- and _ngcontent to _c-, including in docs and tests.

Shortening _nghost and _ngcontent will significantly reduce the overhead from adding additional attributes to HTML elements.

This allows for smaller bundles, which can improve download speeds and reduce parsing time.

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/aio/content/guide/component-styles.md
+++ b/aio/content/guide/component-styles.md
@@ -318,10 +318,10 @@ encapsulation enabled, each DOM element has some extra attributes
 attached to it:
 
 <code-example format="">
-  &lt;hero-details _nghost-pmm-5>
-    &lt;h2 _ngcontent-pmm-5>Mister Fantastic&lt;/h2>
-    &lt;hero-team _ngcontent-pmm-5 _nghost-pmm-6>
-      &lt;h3 _ngcontent-pmm-6>Team&lt;/h3>
+  &lt;hero-details _h-pmm-5>
+    &lt;h2 _c-pmm-5>Mister Fantastic&lt;/h2>
+    &lt;hero-team _c-pmm-5 _h-pmm-6>
+      &lt;h3 _c-pmm-6>Team&lt;/h3>
     &lt;/hero-team>
   &lt;/hero-detail>
 
@@ -339,12 +339,12 @@ generated and you never refer to them in application code. But they are targeted
 by the generated component styles, which are in the `<head>` section of the DOM:
 
 <code-example format="">
-  [_nghost-pmm-5] {
+  [_h-pmm-5] {
     display: block;
     border: 1px solid black;
   }
 
-  h3[_ngcontent-pmm-6] {
+  h3[_c-pmm-6] {
     background-color: white;
     border: 1px solid #777;
   }
--- a/aio/content/marketing/test.html
+++ b/aio/content/marketing/test.html
@@ -48,7 +48,7 @@
 &lt;hero-details <em>nghost-pmm-5&gt;
   &lt;h2 </em>ngcontent-pmm-5&gt;Bah Dah Bing&lt;/h2&gt;
   &lt;hero-team <em>ngcontent-pmm-5 </em>nghost-pmm-6&gt;
-    &lt;h3 _ngcontent-pmm-6&gt;Headless Team&lt;/h3&gt;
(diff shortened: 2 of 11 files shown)


jelbourn:
@ngwattcos looks like there are still some outstanding test failures:
* `integration-payloads.json` looks like it needs to be updated
* On the angular/components repo, we have a test for `MatBadge` in `badge.spec.ts` that tests for the presence of `ngcontent` to verify that ViewEncapsulation is applied; that test will need to be updated before this PR could go in
* There seems be be a test in `legacy-unit-tests-saucelabs` that is looking at the attribute names

ngwattcos:
@jelbourn Made a pull request to badge.spec.ts. However, I'm not clear on what needs to be changed in ```integration-payloads.json```. Also, could you point me to the test on ```legacy-unit-tests-saucelabs```? Thanks.

jelbourn:
@ngwattcos you should be able to find the specific test failures from the CircleCI result pages. 

@pkozlowski-opensource @alxhub FYI this change is from a Google summer intern and was proposed by Igor. What are your thoughts on changing these encapsulation attributes? Shortening the string seems like a good idea to me, but I'm not totally sure whether we would want to do this in a minor or a major. In angular/components, we had one unit tests that was looking at this attribute. Inside Google, there are 11 places where `_ngcontent` is referenced and 4 for `_nghost`, mostly for people doing an exact HTML comparison.

pkozlowski-opensource:
tl;dr; I like the direction of the change but IMO it is breaking enough so that we can do it only in major.

> FYI this change is from a Google summer intern and was proposed by Igor. What are your thoughts on changing these encapsulation attributes? Shortening the string seems like a good idea to me, but I'm not totally sure whether we would want to do this in a minor or a major. In angular/components, we had one unit tests that was looking at this attribute. Inside Google, there are 11 places where _ngcontent is referenced and 4 for _nghost, mostly for people doing an exact HTML comparison.

Just to make sure that I've got proper context here - my understanding is that we are trying to reduce size of the generated code for styles, right? As such the idea sounds really good (still would be interesting to see real-life savings data as those common prefixes should compress quite well?).

The change in the code looks good but it reveals how many tests depend on those prefixes in the repo - I would expect this to be the case in the wild as well. In fact we can see that even some high-profile projects use  _ngcontent / _nghost in their tests: https://github.com/search?l=TypeScript&q=_nghost&type=Code

Apart from tests I'm also worried that people do hacks in their code, mostly to work around various issues - example from a random GitHub search: https://github.com/alexurquhart/ng-tippy/blob/6b4109aae5d4b569971b8f67d638339e18a864d0/src/app/tooltip.directive.ts#L41-L49

Given all this I would say that this is a braking change (in the sense that it will break people projects, even if this isn't formally our public API). As such I would do it only in major and consider mitigation strategies (ex. automatic migration or  configuration option). 

I will do more in-depth review of this PR when it gets green on CI.

jelbourn:
@pkozlowski-opensource that's some good evidence, I'm now convinced that this should only happen in a major version

@ngwattcos @kseamon this would mean that realistically we wouldn't be able to merge this change until we're on the 11.0.0 release train, which is at least a few months away (I don't know the exact timeframe off the top of my head)

petebacondarwin:
@ngwattcos - could you rebase this PR? It has conflicts. Unfortunately we missed the 11.0.0 release, so now this PR will need to target the 12.0.0 release in the spring.

jelbourn:
For context: the author of this PR was a Google summer intern. Since their internship is long over, this is something we'd have to pick up ourselves if we want to move forward with it.

petebacondarwin:
In the context of https://github.com/angular/angular/issues/25355 I would like to consider taking on this PR but changing it so that we add `data-` to the front. I know that this will undue some of the benefit from shortening the names. But if we are going to go to the hassle of updating these then we might as well make the attributes W3C valid.

cc @jelbourn

petebacondarwin:
Closing this for similar reasons to that given for closing https://github.com/angular/angular/pull/40595#issuecomment-820561403.

PR #37799, opened 2020-06-27
fix(elements): detect matchesSelector prototype without IIFE

### LINKED ISSUE #24551: bug(elements): importing `createCustomElement` breaks SSR
## I'm submitting a...

<pre><code>
[x] Bug report  
</code></pre>

## Current behavior

Importing `createCustomElement` throws an error during angular universal server side rendering.

Note: it appears that simply importing the function throws the error. Currently, even the following code throws an error during SSR:

```
import {createCustomElement} from '@angular/elements';
. . .
export class AppModule {
  constructor() {
    if (false as any) {
      createCustomElement
    }
  }
}
```
#### Error:
```
var elProto = Element.prototype;
              ^

ReferenceError: Element is not defined
    at Object.<anonymous> (/user/apps/tests/universal-starter/dist/server.js:126911:15)
    at __webpack_require__ (/user/apps/tests/universal-starter/dist/server.js:20:30)
    at Object.@angular/elements (/user/apps/tests/universal-starter/dist/server.js:126796:18)
    at __webpack_require__ (/user/apps/tests/universal-starter/dist/server.js:126326:30)
    at Object../src/app/app.module.ts (/user/apps/tests/universal-starter/dist/server.js:126453:18)
    at __webpack_require__ (/user/apps/tests/universal-starter/dist/server.js:126326:30)
    at Object../src/app/app.server.module.ngfactory.js (/user/apps/tests/universal-starter/dist/server.js:126504:11)
    at __webpack_require__ (/user/apps/tests/universal-starter/dist/server.js:126326:30)
    at Object../src/main.server.ts (/user/apps/tests/universal-starter/dist/server.js:126712:37)
    at __webpack_require__ (/user/apps/tests/universal-starter/dist/server.js:126326:30)
```

## Expected behavior

Ideally, custom elements could be rendered via SSR normally, but, at minimum, it'd expect the ability to import `createCustomElement` and then *conditionally call it* using `isPlatformBrowser()` so that the custom element was only defined + registered on the client side.

A reproduction of the issue can be found here: https://github.com/thefliik/universal-starter/tree/elements-ssr-bug. Simply clone the repo and run `yarn build:ssr && yarn serve:ssr`.

The repo was made by cloning the `@angular/universal-starter` repo, adding `@angular/elements` to `project.json`, and updating `app.module.ts` with:

```
import {createCustomElement} from '@angular/elements';

...

export class AppModule {
  constructor() {
    /**
     * # Note
     * 
     * Remove this statement and no errors are thrown. I *think* this is because
     * tree shaking will remove the `createCustomElement` import statement without
     * this line.
     */
    if (false as any) {
      createCustomElement
    }
  }
}
```

To allow custom elements built with `@angular/elements` to be used in a SSR app.

My specific use case for angular elements is as a work around for issue #14324 (& #17168).

## Environment

<pre><code>
Angular CLI: 6.0.0
Node: 8.9.1
OS: darwin x64
Angular: 6.0.5
... animations, common, compiler, core, elements, forms, http
... platform-browser, platform-browser-dynamic, platform-server
... router

@angular-devkit/architect         0.6.0
@angular-devkit/build-angular     0.6.0
@angular-devkit/build-optimizer   0.6.0
@angular-devkit/core              0.6.0
@angular-devkit/schematics        0.6.0
@angular/cli                      6.0.0
@angular/compiler-cli             6.0.0
@angular/language-service         6.0.0
@ngtools/webpack                  6.0.0
@schematics/angular               0.6.0
@schematics/update                0.6.0
rxjs                              6.0.0
typescript                        2.7.2
webpack                           4.6.0
</code></pre>

CaerusKaru:
Although in SSR we patch the global prototypes with DOM globals
like Element and Node, this patch does not occur before the
matches function is called in Angular Elements. This is similar
to the behavior in @angular/upgrade.

Fixes #24551

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: 24551

- [ ] Yes
- [x] No

### DIFF
--- a/packages/elements/src/utils.ts
+++ b/packages/elements/src/utils.ts
@@ -7,12 +7,6 @@
  */
 import {ComponentFactoryResolver, Injector, Type} from '@angular/core';
 
-const matches = (() => {
-  const elProto = Element.prototype as any;
-  return elProto.matches || elProto.matchesSelector || elProto.mozMatchesSelector ||
-      elProto.msMatchesSelector || elProto.oMatchesSelector || elProto.webkitMatchesSelector;
-})();
-
 /**
  * Provide methods for scheduling the execution of a callback.
  */
@@ -96,11 +90,20 @@ export function kebabToCamelCase(input: string): string {
   return input.replace(/-([a-z\d])/g, (_, char) => char.toUpperCase());
 }
 
+let _matches: (this: any, selector: string) => boolean;
+
 /**
  * Check whether an `Element` matches a CSS selector.
+ * NOTE: this is duplicated from @angular/upgrade, and can
+ * be consolidated in the future
  */
-export function matchesSelector(element: Element, selector: string): boolean {
-  return matches.call(element, selector);
+export function matchesSelector(el: any, selector: string): boolean {
+  if (!_matches) {
+    const elProto = <any>Element.prototype;
+    _matches = elProto.matches || elProto.matchesSelector || elProto.mozMatchesSelector ||
+        elProto.msMatchesSelector || elProto.oMatchesSelector || elProto.webkitMatchesSelector;
+  }
+  return el.nodeType === Node.ELEMENT_NODE ? _matches.call(el, selector) : false;
 }
 
 /**


gkalpak (approved):
I left a couple of comments. Otherwise lgtm.

##
Not directly related to this PR:

It seems we have similar (but not 100% the same) code in `@angular/animations`' [shared.ts][1] and `@angular/upgrade`'s [downgrade_component_adapter.ts][2]. I wonder if it would make sense to have a shared (internal) utility package for such helpers.

Also, these helpers are not 100% the same. It is hard to tell whether the differences account for slightly different needs of each package and how the helpers are used or whether we should align all implementations 😕 

[1]: https://github.com/angular/angular/blob/eba9dd364e51a6e857a94fd5f8980fec4350c856/packages/animations/browser/src/render/shared.ts#L168-L181
[2]: https://github.com/angular/angular/blob/eba9dd364e51a6e857a94fd5f8980fec4350c856/packages/upgrade/src/common/src/downgrade_component_adapter.ts#L291-L300

packages/elements/src/utils.ts
```diff
@@ -8,9 +8,18 @@
 import {ComponentFactoryResolver, Injector, Type} from '@angular/core';
 
 const matches = (() => {
-  const elProto = Element.prototype as any;
-  return elProto.matches || elProto.matchesSelector || elProto.mozMatchesSelector ||
-      elProto.msMatchesSelector || elProto.oMatchesSelector || elProto.webkitMatchesSelector;
+  if (typeof Element === 'undefined' || Element.prototype.matches) {
+    return (element: any, selector: string) => element.matches(selector);
```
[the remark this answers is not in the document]
CaerusKaru: Since this is borrowed from animations shared code, I'm not entirely sure of the typing decisions. But I think until we consolidate, we should make sure the code is identical to avoid people trying to figure out which one is "better" later.

packages/elements/src/utils.ts
```diff
@@ -7,11 +7,36 @@
...
-  return elProto.matches || elProto.matchesSelector || elProto.mozMatchesSelector ||
-      elProto.msMatchesSelector || elProto.oMatchesSelector || elProto.webkitMatchesSelector;
-})();
+export function isNode(): boolean {
+  // Checking only for `process` isn't enough to identify whether or not we're in a Node
+  // environment, because Webpack by default will polyfill the `process`. While we can discern
+  // that Webpack polyfilled it by looking at `process.browser`, it's very Webpack-specific and
+  // might not be future-proof. Instead we look at the stringified version of `process` which
+  // is `[object process]` in Node and `[object Object]` when polyfilled.
+  return typeof process !== 'undefined' && {}.toString.call(process) === '[object process]';
```
petebacondarwin, line 16: I guess you are using `{}.toString.call()` because `process` could feasibly be `null`?

packages/elements/src/utils.ts
```diff
@@ -7,11 +7,36 @@
...
+      const fn = proto.matchesSelector || proto.mozMatchesSelector || proto.msMatchesSelector ||
+          proto.oMatchesSelector || proto.webkitMatchesSelector;
+      if (fn) {
+        return (element: any, selector: string) => fn.apply(element, [selector]);
+      } else {
+        return _matches;
+      }
+    }
+  })();
+}
```
petebacondarwin, line 38: This logic seems super complex. Is it to try to keep the code-size down or something else?

petebacondarwin: I guess it is just a copy from the animations library.

petebacondarwin (approved):
Reviewed-for: size-tracking

Not going to block the merge. I guess 600 bytes is not a killer.

But I think it would be good to consider either (or both) of the following:

1. Make the `_matches` code easier to understand
2. Make the `_matches` code smaller

It is likely that these two are in opposition to each other. But I feel that as it stands we have the worst of both scenarios. I feel that it could be smaller or simpler...

Another option, as @gkalpak suggested, is to move this to a shared function, since I imagine that `elements` is rarely used without `animations`, in which case we are getting hit twice for this logic. Both of these have `core` as peer dependency, so it could just be moved there an exported "internally".

jelbourn (commented):
Would it be possible to add this shim exclusively to server environments rather than always baking it into the client-side lib?

packages/elements/src/utils.ts
```diff
@@ -7,11 +7,36 @@
  */
 import {ComponentFactoryResolver, Injector, Type} from '@angular/core';
 
-const matches = (() => {
-  const elProto = Element.prototype as any;
-  return elProto.matches || elProto.matchesSelector || elProto.mozMatchesSelector ||
-      elProto.msMatchesSelector || elProto.oMatchesSelector || elProto.webkitMatchesSelector;
-})();
+export function isNode(): boolean {
```
alan-agius4, line 10: I don’t see any reason why this should be exported?

packages/elements/src/utils.ts
alan-agius4: I also find this to be super complex.

alan-agius4 (dismissed):
Can we change this implementation to something like: https://github.com/angular/angular/blob/d1ea1f4c7f3358b730b0d94e65b00bc28cae279c/packages/upgrade/src/common/src/downgrade_component_adapter.ts#L291-L300

it's much simpler, and should solve the issue since it's not using an IIFE.

PatrickJS:
if we use `Object.getPrototypeOf` we can keep the code the same e.g.
https://github.com/angular/angular/pull/38388/files#diff-97b1abb49de0e54f94eba3f03982ee1cR103

gkalpak:
@PatrickJS: I don't think we can (see https://github.com/angular/angular/pull/38388/files#r473900039).

@CaerusKaru: Are you still interested/available to move this forward? (If not, I'd be happy to take it over.)

packages/elements/src/utils.ts
```diff
@@ -96,11 +90,20 @@ export function kebabToCamelCase(input: string): string {
   return input.replace(/-([a-z\d])/g, (_, char) => char.toUpperCase());
 }
 
+let _matches: (this: any, selector: string) => boolean;
+
 /**
  * Check whether an `Element` matches a CSS selector.
+ * NOTE: this is duplicated from @angular/upgrade, and can
+ * be consolidated in the future
```
gkalpak, line 98: :100:

gkalpak (approved):
Still lgtm :+1:
Thx, @CaerusKaru!

atscott:
[presubmit](http://test/OCL:336295809:BASE:336295836:1602256688031:8b350869)

vinothbabu:
Does this issue fix the `createCustomElement` issue in SSR?

gkalpak:
What is the `createCustomElement()` issue? 😕

vinothbabu:
> What is the `createCustomElement()` issue? 😕

Angular Element is not available in SSR. (Below is the issue ticket). 
https://github.com/angular/angular/issues/24551

gkalpak:
Yes, this PR should have fixed #24551.

vinothbabu:
> Yes, this PR should have fixed #24551.

Wow, that's a great news. Would that mean we can use custom-elements in SSR and which release we are making it?

gkalpak:
It will be in the next 10.1.x release (10.1.6 - assuming there are going to be more v10 releases before v11) and the next 11.x release (11.0.0-next.6).

PR #37809, opened 2020-06-28
fix(compiler): allow banana-in-a-box bindings to end with non-null assertion

### LINKED ISSUE #36551: Two-way binding not work with non-null assertion operator
# 🐞 bug report

### Affected Package

@angular/compiler-cli or @angular/compiler ?

### Is this a regression?

No

### Description

I enabled strictTemplates in my project and follow this [guide](https://angular.io/guide/template-typecheck#strict-null-checks) to fix type errors.

But when I add a non-null assertion operator to the two-way binding value, the compilation result does not meet expectations.

## 🔬 Minimal Reproduction

```ts
import { Component } from '@angular/core';

@Component({
  selector: 'app-root',
  template: `
    <input type="checkbox" [(ngModel)]="obj.value!">
    <!--                                ^^^^^^^^^ -->
  `,
  styleUrls: ['./app.component.less']
})
export class AppComponent {
  obj: {
    value?: boolean
  } = {}
}
```

#### Expected

```js
function AppComponent_Template(rf, ctx) {
    if (rf & 1) {
        i0.ɵɵelementStart(0, "input", 0);
        i0.ɵɵlistener("ngModelChange", function AppComponent_Template_input_ngModelChange_0_listener($event) {
            return ctx.obj.value = $event;
            //                     ^^^^^^^
        });
        i0.ɵɵelementEnd();
    }
    if (rf & 2) {
        i0.ɵɵproperty("ngModel", ctx.obj.value);
    }
}
```

#### Actually

```js
function AppComponent_Template(rf, ctx) {
    if (rf & 1) {
        i0.ɵɵelementStart(0, "input", 0);
        i0.ɵɵlistener("ngModelChange", function AppComponent_Template_input_ngModelChange_0_listener($event) {
            return ctx.obj.value != $event;
            //                   ^^^^^^^^^
        });
        i0.ɵɵelementEnd();
    }
    if (rf & 2) {
        i0.ɵɵproperty("ngModel", ctx.obj.value);
    }
}
```

## 🌍  Your Environment

**Angular Version:**
<pre><code>

Angular CLI: 9.1.1
Node: 12.16.1
OS: darwin x64

Angular: 9.1.1
... router
Ivy Workspace: Yes

@angular-devkit/architect          0.901.1
@angular-devkit/build-angular      0.901.1
@angular-devkit/build-ng-packagr   0.901.1
@angular-devkit/build-optimizer    0.901.1
@angular-devkit/build-webpack      0.901.1
@angular-devkit/core               9.1.1
@angular-devkit/schematics         9.1.1
@ngtools/webpack                   9.1.1
@schematics/angular                9.1.1
@schematics/update                 0.901.1

</code></pre>

**Anything else relevant?**

#### Current solution

```html
<input type="checkbox" [(ngModel)]="!!obj.value">
```
```js
function AppComponent_Template(rf, ctx) {
    if (rf & 1) {
        i0.ɵɵelementStart(0, "input", 0);
        i0.ɵɵlistener("ngModelChange", function AppComponent_Template_input_ngModelChange_0_listener($event) {
            return !!(ctx.obj.value = $event);
        });
        i0.ɵɵelementEnd();
    }
    if (rf & 2) {
        i0.ɵɵproperty("ngModel", !!ctx.obj.value);
    }
}
```

or

```html
<input type="checkbox" [(ngModel)]="$any(obj).value">
```

or

```html
<input type="checkbox" [ngModel]="obj.value!" (ngModelChange)="obj.value = $event">
```

JoostK:
For two-way-bindings that use the banana-in-a-box syntax, the compiler
synthesizes an event assignment expression from the primary expression.
It is valid for the primary expression to be terminated by the non-null
operator, however naive string substitution is used for the synthesized
expression, such that the `!` would immediately precede the `=` token,
resulting in the valid `!=` operator token. The expression would still
parse correctly but it doesn't implement the proper semantics, resulting
in incorrect runtime behavior.

Changing the expression substitution to force a space between the
primary expression and the assignment avoids this mistake, but it
uncovers a new issue. The grammar does not allow for the LHS of an
assignment to be the non-null operator, so the synthesized expression
would fail to parse. To alleviate this, the synthesized expression is
parsed with a special parser flag to allow for this syntax.

Fixes #36551

---

Implementation note: it is up for debate whether extending the grammar to accept more expressions as LHS of an assignment would be desirable for regular output bindings. I choose not to change the grammar at this moment, as it seems like a change that needs to be introduced in at least a minor. Additionally, I feel like the whole expression substitution is a bit of a hack in the first place, so an alternative way of handling this case may be desirable in the future. To allow for this cleanup, I feel like it's best to wait for VE removal as the parser infrastructure is currently quite convoluted.

### DIFF
--- a/packages/compiler/src/expression_parser/parser.ts
+++ b/packages/compiler/src/expression_parser/parser.ts
@@ -29,19 +29,41 @@ export class TemplateBindingParseResult {
       public errors: ParserError[]) {}
 }
 
+/**
+ * Represents the possible parse modes to be used as a bitmask.
+ */
+export const enum ParseFlags {
+  None = 0,
+
+  /**
+   * Whether an output binding is being parsed.
+   */
+  Action = 1 << 0,
+
+  /**
+   * Whether an assignment event is being parsed, i.e. an expression originating from
+   * two-way-binding aka banana-in-a-box syntax.
+   */
+  AssignmentEvent = 1 << 1,
+}
+
 export class Parser {
   private errors: ParserError[] = [];
 
   constructor(private _lexer: Lexer) {}
 
   parseAction(
-      input: string, location: string, absoluteOffset: number,
+      input: string, isAssignmentEvent: boolean, location: string, absoluteOffset: number,
       interpolationConfig: InterpolationConfig = DEFAULT_INTERPOLATION_CONFIG): ASTWithSource {
     this._checkNoInterpolation(input, location, interpolationConfig);
     const sourceToLex = this._stripComments(input);
     const tokens = this._lexer.tokenize(sourceToLex);
+    let flags = ParseFlags.Action;
+    if (isAssignmentEvent) {
+      flags |= ParseFlags.AssignmentEvent;
+    }
     const ast =
-        new _ParseAST(input, location, absoluteOffset, tokens, true, this.errors, 0).parseChain();
+        new _ParseAST(input, location, absoluteOffset, tokens, flags, this.errors, 0).parseChain();
     return new ASTWithSource(ast, input, location, absoluteOffset, this.errors);
   }
 
@@ -80,7 +102,7 @@ export class Parser {
     this._checkNoInterpolation(input, location, interpolationConfig);
     const sourceToLex = this._stripComments(input);
     const tokens = this._lexer.tokenize(sourceToLex);
-    return new _ParseAST(input, location, absoluteOffset, tokens, false, this.errors, 0)
+    return new _ParseAST(input, location, absoluteOffset, tokens, ParseFlags.None, this.errors, 0)
         .parseChain();
   }
 
@@ -115,8 +137,8 @@ export class Parser {
       absoluteValueOffset: number): TemplateBindingParseResult {
     const tokens = this._lexer.tokenize(templateValue);
     const parser = new _ParseAST(
-        templateValue, templateUrl, absoluteValueOffset, tokens, false /* parseAction */,
-        this.errors, 0 /* relative offset */);
+        templateValue, templateUrl, absoluteValueOffset, tokens, ParseFlags.None, this.errors,
+        0 /* relative offset */);
     return parser.parseTemplateBindings({
       source: templateKey,
       span: new AbsoluteSourceSpan(absoluteKeyOffset, absoluteKeyOffset + templateKey.length),
@@ -137,7 +159,8 @@ export class Parser {
       const sourceToLex = this._stripComments(expressionText);
       const tokens = this._lexer.tokenize(sourceToLex);
       const ast =
-          new _ParseAST(input, location, absoluteOffset, tokens, false, this.errors, offsets[i])
+          new _ParseAST(
+              input, location, absoluteOffset, tokens, ParseFlags.None, this.errors, offsets[i])
               .parseChain();
       expressionNodes.push(ast);
     }
@@ -155,10 +178,9 @@ export class Parser {
       ASTWithSource {
     const sourceToLex = this._stripComments(expression);
     const tokens = this._lexer.tokenize(sourceToLex);
-    const ast = new _ParseAST(
-                    expression, location, absoluteOffset, tokens,
-                    /* parseAction */ false, this.errors, 0)
-                    .parseChain();
+    const ast =
+        new _ParseAST(expression, location, absoluteOffset, tokens, ParseFlags.None, this.errors, 0)
+            .parseChain();
     const strings = ['', ''];  // The prefix and suffix strings are both empty
     return this.createInterpolationAst(strings, [ast], expression, location, absoluteOffset);
   }
@@ -368,7 +390,7 @@ export class _ParseAST {
 
   constructor(
       public input: string, public location: string, public absoluteOffset: number,
-      public tokens: Token[], public parseAction: boolean, private errors: ParserError[],
+      public tokens: Token[], public parseFlags: ParseFlags, private errors: ParserError[],
       private offset: number) {}
 
   peek(offset: number): Token {
@@ -551,7 +573,7 @@ export class _ParseAST {
       exprs.push(expr);
 
       if (this.consumeOptionalCharacter(chars.$SEMICOLON)) {
-        if (!this.parseAction) {
+        if (!(this.parseFlags & ParseFlags.Action)) {
           this.error('Binding expression cannot contain chained expression');
         }
         while (this.consumeOptionalCharacter(chars.$SEMICOLON)) {
@@ -576,7 +598,7 @@ export class _ParseAST {
     const start = this.inputIndex;
     let result = this.parseExpression();
     if (this.consumeOptionalOperator('|')) {
-      if (this.parseAction) {
+      if (this.parseFlags & ParseFlags.Action) {
         this.error('Cannot have a pipe in an action expression');
       }
 
@@ -932,7 +954,7 @@ export class _ParseAST {
     let receiver: AST;
 
     if (isSafe) {
-      if (this.consumeOptionalOperator('=')) {
+      if (this.consumeOptionalAssignment()) {
         this.error('The \'?.\' operator cannot be used in the assignment');
         receiver = new EmptyExpr(this.span(start), this.sourceSpan(start));
       } else {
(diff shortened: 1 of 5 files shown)


packages/compiler/src/expression_parser/parser.ts
```diff
@@ -150,7 +169,7 @@ export class Parser {
     const tokens = this._lexer.tokenize(templateValue);
     const parser = new _ParseAST(
         templateValue, templateUrl, absoluteValueOffset, tokens, templateValue.length,
-        false /* parseAction */, this.errors, 0 /* relative offset */);
+        ParseFlags.None /* parseFlags */, this.errors, 0 /* relative offset */);
```
ayazhafiz, line 172: I think the comment is unnecessary now
```suggestion
        ParseFlags.None, this.errors, 0 /* relative offset */);
```

packages/compiler/test/template_parser/template_parser_spec.ts
```diff
@@ -902,6 +902,19 @@ Animation trigger is missing ("<div [ERROR ->]@></div>"): TestComp@0:5`);
...
+        expect(humanizeTplAst(parse('<div [(prop)]="v!">', []))).toEqual([
+          [ElementAst, 'div'],
+          [BoundElementPropertyAst, PropertyBindingType.Property, 'prop', 'v!', null],
+          [BoundEventAst, 'propChange', null, 'v = $event']
+        ]);
+      });
+
+      it('should report an error for assignments into non-null asserted expressions', () => {
+        expect(() => parse('<div (prop)="v! = $event"></div>', []))
+            .toThrowError(/Unexpected token '=' at column 4/);
```
ayazhafiz, line 915: This error message isn't incorrect, but I think I better message would mention that the non-null assertion operator cannot appear on the LHS of an assignment expression. If this sounds reasonable, could you add a todo here?

ayazhafiz (approved):
Nice fix!

packages/compiler/test/template_parser/template_parser_spec.ts
JoostK: Agreed, I added a note here. It's really not that hard to have a specialized error message, however I'd rather not touch this at the moment (especially if we do want to support this syntax)

ayazhafiz: yep, makes sense

alxhub (approved):

alxhub:
[Presubmit](http://test/OCL:320651042:BASE:320651076:1594408711779:3d56f76)

alxhub:
[And again](http://test/OCL:320651042:BASE:324706478:1596496911578:dce3612f)

alxhub:
The g3 failure here is legitimate, as this does break `codelyzer` internally, due to the extra parameter added to `parseAction`. Let's sync up on this.

alxhub:
@JoostK we can fix this internally just by patching `codelyzer`. Can you test this change with `angular-eslint` (the spiritual successor) and verify that no external fixes are needed there?

jessicajaniuk:
@JoostK Looks like this PR stalled. Any idea what the status is here and whether we still plan on landing it?

JoostK:
Picking this one up again. This was rebased on latest master and I tested the build artefacts in `angular-eslint` (on its master branch after updating to v12 RC releases) and there were no issues there (which is to be expected as they don't touch lower level parse API, unlike Codelyzer).

JoostK:
Blocked on a g3 sync issue as this change is incompatible with Codelyzer.

packages/compiler/src/expression_parser/parser.ts
```diff
@@ -399,7 +418,7 @@ export class _ParseAST {
 
   constructor(
       public input: string, public location: string, public absoluteOffset: number,
-      public tokens: Token[], public inputLength: number, public parseAction: boolean,
+      public tokens: Token[], public inputLength: number, public parseFlags: number,
```
petebacondarwin, line 421:
```suggestion
      public tokens: Token[], public inputLength: number, public parseFlags: ParseFlags,
```

??

JoostK: Thanks, good catch!

packages/compiler/src/expression_parser/parser.ts
```diff
@@ -29,6 +29,21 @@ export class TemplateBindingParseResult {
       public errors: ParserError[]) {}
 }
 
+export const enum ParseFlags {
+  None = 0,
+
+  /**
+   * Whether an output binding is parsed.
```
petebacondarwin, line 36:
```suggestion
   * Whether an output binding is being parsed.
```

packages/compiler/src/expression_parser/parser.ts
```diff
@@ -29,6 +29,21 @@ export class TemplateBindingParseResult {
...
+   */
+  Action = 1 << 0,
+
+  /**
+   * Whether an assignment event is parsed, i.e. an expression originating from two-way-binding
```
petebacondarwin, line 41:
```suggestion
   * Whether an assignment event is being parsed, i.e. an expression originating from two-way-binding
```

AndrewKushnir:
[Presubmit](http://test/OCL:407637974:BASE:407638034:1636052694155:9eea53b9).

packages/compiler/src/expression_parser/parser.ts
```diff
@@ -29,6 +29,21 @@ export class TemplateBindingParseResult {
       public errors: ParserError[]) {}
 }
 
+export const enum ParseFlags {
```
AndrewKushnir, line 32: Could you please add a quick comment above this enum on what it represents?

packages/compiler/src/expression_parser/parser.ts
```diff
@@ -1007,6 +1026,22 @@ export class _ParseAST {
...
+  private consumeOptionalAssignment(): boolean {
+    // When parsing assignment events (originating from two-way-binding aka banana-in-a-box syntax),
+    // it is valid for the primary expression to be terminated by the non-null operator. This
+    // primary expression is substituted as LHS of the assignment operator to achieve
+    // two-way-binding, such that the LHS could be the non-null operator. The grammar doesn't
+    // naturally allow for this syntax, so assignment events are parsed specially.
+    if ((this.parseFlags & ParseFlags.AssignmentEvent) && this.next.isOperator('!') &&
+        this.peek(1).isOperator('=')) {
+      this.advance();
+      this.advance();
```
AndrewKushnir, line 1038: Could you plz add a quick comment here to provide more context on why we have 2 advance calls?

dram2040:
```
- 1037

```
**# 1038**

dram2040: [B:B$C3](~~`https://help.audiostretch.com/hc/en-us`~~)

AndrewKushnir:
[Global Presubmit](https://test.corp.google.com/ui#id=OCL:407637974:BASE:407735154:1636084468867:21301389).

AndrewKushnir:
@JoostK FYI the most recent [TGP is fully "green"](https://test.corp.google.com/ui#id=OCL:407637974:BASE:407828381:1636126234966:62703070) (internal-only link) with this change: there are no issues with Codelyzer. I don't have an access to Alex's test runs (the results are no longer available), but I believe TGP should've triggered Codelyzer-related targets.

@alxhub I've also looked at the Codelyzer folder and didn't notice any updates that make it compatible with this change. So I'm wondering if we should run more tests or we can proceed with caution (merge and sync the change as a separate commit, be ready to rollback). What do you think?

jessicajaniuk:
@JoostK Can you rebase this to resolve the conflicts and ping me when you do? We'll run another TGP and see how it goes. Hopefully we can land this.

JoostK:
Rebased and added to the merge queue :+1:

AndrewKushnir:
[Started a new TGP](https://test.corp.google.com/OCL:407637974:BASE:426791709:1644194387100:f6e10afa).

AndrewKushnir:
TGP is "green" :) Adding this PR to the merge queue...

dylhunn:
This PR was merged into the repository by commit db6cf7e7c168f9a25550d7fa53840c08b50b889c.

PR #37811, opened 2020-06-28
fix(compiler): properly compile DI factories when coverage reporting is enabled

### LINKED ISSUE #31337: Dependency injection and class inheritance doesn't work in Angular 8 library test
# 🐞 bug report

### Affected Package

The issue is caused by package @angular/core@^8

### Is this a regression?
Yes, the previous version in which this bug was not present was: @angular/core@7

### Description

The problem occurs when:

1. The base class has a constructor with an injected service. (`Injector` from the example)

```ts
@Injectable()
export class BaseService {
  // "Injector" is just for the sake of example. Any injectable would have the same problem.
  constructor(public injector: Injector) {
  }
}
```
2. Child class has no constructor (i.e., it relies on base class constructor) AND at least one property with the default value.

```ts
@Injectable()
export class MyLibService extends BaseService {

  public foo = 'bar';

}
```

3. (_dont ask me how I have found this, but:_) In `angular.json` `[lib name].test.options.codeCoverage` should be true. 

### PROBLEM:   

`injector` is undefined in the instance of `MyLibService`. 

--- 

🚩 Removing property from the child class _(2)_ AND/OR opting out `codeCoverage` config in angular.json _(3)_ doesn't brake test.

As a result, I can't have a class inheritance and code coverage at the same time.

---

I could put a constructor in the child class and call `super` by passing all required arguments manually, but it's the last option since the real-life scenario is much more complicated than the example provided here.

## 🔬 Minimal Reproduction
git repo: https://github.com/ddramone/ng-8-library-inheritance-di-test-bug

reproduce steps:

1. Install angular 8 cli
2. Clone `git clone https://github.com/ddramone/ng-8-library-inheritance-di-test-bug.git `
3. `npm i`
4. run Test `ng test my-lib`

 ### Expected behavior

Test should pass = Injector service should be available in the service instance.

JoostK:
When running tests with code coverage using Istanbul, the code is
instrumented with coverage reporting statements. These statements are
also inserted into synthesized constructors, preventing Angular from
properly recognizing them as synthesized constructor.

This commit extends the logic to detect synthesized constructors by
first canonicalizing the constructor's code, which strips out Istanbul's
instrumentation logic.

The tests have been extended with an input file that is being
instrumented using the `babel-plugin-istanbul` for both ES2015 and ES5
targets, in order to verify that the approach works for real-world
usages.

Fixes #31337


sviat9440:
Any news?

packages/core/src/reflection/reflection_capabilities.ts
```diff
@@ -58,6 +58,32 @@ export const ES2015_INHERITED_CLASS_WITH_CTOR =
...
+/**
+ * Registers a function which computes the canonical form of a constructor function's source code,
+ * to conform with the expected structures as outlined in the above regular expressions.
+ *
+ * This is used in testing environments, where code coverage instrumentation statements have been
+ * injected which need to be removed.
+ *
+ * @param fn The function that implements the canonicalization behavior.
+ */
+export function setCtorCanonicalizer(fn: (ctor: string) => string): void {
```
JoostK, line 76: Note: this avoids pulling in the Istanbul related canonicalization logic into production logic, with the downside that testing environments need to configure it correctly. This PR does that for `TestBed`, but other situations are not covered. Should we just pull in the canonicalization logic here instead? That seems a bit simpler, but I'd love some feedback here.

klemenoslaj:
Any news on this?

JoostK:
@klemenoslaj I haven't had cycles to work on this due to v12 priorities, but I was planning to pick up on this sometime soon.

klemenoslaj:
> @klemenoslaj I haven't had cycles to work on this due to v12 priorities, but I was planning to pick up on this sometime soon.

Thanks for a super fast reply @JoostK.
I am now setting target to ES5, but just in the `tsconfig.spec.json`. Is that a valid workaround?

JoostK:
> > @klemenoslaj I haven't had cycles to work on this due to v12 priorities, but I was planning to pick up on this sometime soon.
> 
> Thanks for a super fast reply @JoostK.
> I am now setting target to ES5, but just in the `tsconfig.spec.json`. Is that a valid workaround?

AFAIK the problem would also occur in ES5, as the code coverage statements cause Angular to consider a constructor as manually written instead of being just a delegated constructor to the base type.

A workaround may be to always write the constructor, including all parameters and a `super` call in any of your own classes. This doesn't help for libraries but that shouldn't be a problem if they are AOT compiled by ngcc (and not recompiled by `TestBed` using `override*` API). This does have some bundle size impact thought.

klemenoslaj:
> > > @klemenoslaj I haven't had cycles to work on this due to v12 priorities, but I was planning to pick up on this sometime soon.
> > 
> > 
> > Thanks for a super fast reply @JoostK.
> > I am now setting target to ES5, but just in the `tsconfig.spec.json`. Is that a valid workaround?
> 
> AFAIK the problem would also occur in ES5, as the code coverage statements cause Angular to consider a constructor as manually written instead of being just a delegated constructor to the base type.
> 
> A workaround may be to always write the constructor, including all parameters and a `super` call in any of your own classes. This doesn't help for libraries but that shouldn't be a problem if they are AOT compiled by ngcc (and not recompiled by `TestBed` using `override*` API). This does have some bundle size impact thought.

Interestingly, I just tested with and without es5 setup. Without it fails, with it works fine. 🤔

JoostK:
> > > > @klemenoslaj I haven't had cycles to work on this due to v12 priorities, but I was planning to pick up on this sometime soon.
> > > 
> > > 
> > > Thanks for a super fast reply @JoostK.
> > > I am now setting target to ES5, but just in the `tsconfig.spec.json`. Is that a valid workaround?
> > 
> > 
> > AFAIK the problem would also occur in ES5, as the code coverage statements cause Angular to consider a constructor as manually written instead of being just a delegated constructor to the base type.
> > A workaround may be to always write the constructor, including all parameters and a `super` call in any of your own classes. This doesn't help for libraries but that shouldn't be a problem if they are AOT compiled by ngcc (and not recompiled by `TestBed` using `override*` API). This does have some bundle size impact thought.
> 
> Interestingly, I just tested with and without es5 setup. Without it fails, with it works fine. 🤔

I have revised the PR a bit and replaced the integration test using the CLI with manually instrumenting code using `babel-plugin-istanbul`, which made it far easier to exercise the logic for both ES2015 and ES5. Surprisingly to me, the ES5 test passes even without the changes in this PR; so that confirms your findings for ES5. The ES5's regex is probably more lenient, allowing it to work as desired even with instrumentation instructions.

packages/core/src/reflection/istanbul_canonicalizer.ts
```diff
@@ -0,0 +1,52 @@
...
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+/**
+ * When running tests with code coverage using Istanbul, the code is instrumented with coverage
+ * reporting statements. These statements are also inserted into synthesized constructors,
+ * preventing Angular from properly recognizing them as synthesized constructor.
```
petebacondarwin, line 12:
```suggestion
 * preventing Angular from properly recognizing them as synthesized constructors.
```

packages/core/test/BUILD.bazel
```diff
@@ -32,6 +34,64 @@ genrule(
...
+tsc(
+    name = "istanbul_inheritance_fixture_es2015",
+    testonly = True,
+    outs = ["reflection/istanbul_inheritance_fixture_es2015.js"],
+    args = [
+        "--target es2015",
+        "--module umd",
+        "--outDir $(@D)",
+        "--skipLibCheck",
+        "$(execpath reflection/istanbul_inheritance_fixture_es2015.ts)",
```
petebacondarwin, line 46: Is there any need to have two separate fixture files? Can we not have two rules that both process the same fixture (producing the ES2015 and ES5 outputs)?

petebacondarwin (approved):
LGTM - a couple of suggestions.

petebacondarwin:
BTW - I assume that `core/reflections` code does not normally appear in AOT built production ivy code?

packages/core/test/BUILD.bazel
JoostK: The input file specifies the AMD module name using a `/// <amd-module name="..">` comment, which is needed such that the generated UMD module has the correct module name. There's no TypeScript compiler option to achieve this so if there were only a single input file, I think we'd need an intermediate Bazel target using `genrule` to replace the module name ad-hoc, which I felt wasn't much better than just duplicating the file.

JoostK:
> BTW - I assume that `core/reflections` code does not normally appear in AOT built production ivy code?

Indeed, it shouldn't. The `ReflectionCapabilities` should only be pulled in when using the JIT compiler.

packages/core/test/BUILD.bazel
petebacondarwin: Oh! I didn't spot that difference between the files. Perhaps you could make that more clear in the comment in each?

JoostK: Added a note to emphasize this.

JoostK:
Converting back to draft as there may be an alternative fix that would be superior, if it works.

JoostK:
Closing as superseded by #44732.

Issue #24181, opened 2018-05-29
@angular/elements component events are called from an inconsistent zone


divdavem:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead see https://github.com/angular/angular/blob/master/CONTRIBUTING.md#question
[ ] Other... Please describe:
</code></pre>

## Current behavior

Event handlers in an angular element (registered as a custom element with `@angular/elements`), are called either from the angular zone or from the root zone depending on how the element is instantiated, leading to different behaviors:
- if the element is in the initial markup of the page (before the custom element is registered), then event handlers are called from the angular zone (as expected), and the change detection and refresh of the template are done correctly
- if the element is created from the angular zone (after the custom element is registered), then event handlers are called either from the angular zone (in Chrome, leading to a correct behavior) or from the root zone (in other browsers, leading to an incorrect behavior: the change detection is not done and the template is not refreshed)
- if the element is created from the root zone (after the custom element is registered), then event handlers are called from the root zone (both in Chrome and other browsers), leading to an incorrect behavior: the change detection is not done and the template is not refreshed.

## Expected behavior

The angular element should behave correctly no matter how it is instantiated.
In the minimal reproduction sample, clicking on "Toggle display" should work no matter the line.

## Minimal reproduction of the problem with instructions

https://stackblitz.com/edit/angular-dxu5td?file=src%2Fapp%2Fhello.component.ts
Depending on the line and the browser, clicking on the "Toggle display" works or has no effect.
In Chrome, only the first two lines work correctly. In other browsers, only the first line works correctly.

## What is the motivation / use case for changing the behavior?

The current behavior is inconsistent. Event handlers should be called from the angular zone no matter how the component is instantiated, which would allow those components to be created from any other framework even if that framework does not run in the angular zone.

## Environment

<pre><code>
Angular version: 6.0.3

Browser:
- [x] Chrome (desktop) version 66.0.3359.181
- [x] Firefox version 60.0.1
- [x] IE version XX
- [x] Edge version XX

</code></pre>

divdavem:
@remackgeek Thank you for your quick response with a PR!

robwormald:
[...]

PR #24185, opened 2018-05-29
fix(elements): initialization should run in ngZone (#24181)

remackgeek:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x ] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [x ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

```
[x ] Bugfix
[ ] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Documentation content changes
[ ] angular.io application / infrastructure changes
[ ] Other... Please describe:
```

## What is the current behavior?

Issue Number: #24181 
When an angular element is created outside the angular zone, such as in setTimeout(), automatic change detection no longer works

## What is the new behavior?
during connect(), initializeComponent is always called in the angular zone

## Does this PR introduce a breaking change?
```
[ ] Yes
[ x] No
```

## Other information
@JiaLiPassion this is based/extends on your PR 23885, please review

PR #37814, opened 2020-06-29
fix(elements): elements should run in correct zone (#24181)

remackgeek:
## PR Checklist

- [x ] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [x ] Tests for the changes have been added (for bug fixes / features)

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

When an angular element is created outside the angular zone, such as in setTimeout(), automatic change detection no longer works

Issue Number:  [#24181](https://github.com/angular/angular/issues/24181)

## What is the new behavior?
ComponentNgElementStrategy captures the zone it was created in and insures all methods are run in that zone.

- [ ] Yes
- [x ] No

### DIFF
--- a/packages/elements/src/component-factory-strategy.ts
+++ b/packages/elements/src/component-factory-strategy.ts
@@ -6,7 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {ApplicationRef, ComponentFactory, ComponentFactoryResolver, ComponentRef, EventEmitter, Injector, OnChanges, SimpleChange, SimpleChanges, Type} from '@angular/core';
+import {ApplicationRef, ComponentFactory, ComponentFactoryResolver, ComponentRef, EventEmitter, Injector, NgZone, OnChanges, SimpleChange, SimpleChanges, Type} from '@angular/core';
 import {merge, Observable, ReplaySubject} from 'rxjs';
 import {map, switchMap} from 'rxjs/operators';
 
@@ -73,6 +73,13 @@ export class ComponentNgElementStrategy implements NgElementStrategy {
    */
   private readonly unchangedInputs = new Set<string>();
 
+  /** Service for setting zone context. */
+  private readonly ngZone = this.injector.get<NgZone>(NgZone);
+
+  /** The zone the element was created in or `null` if Zone.js is not loaded. */
+  private readonly elementZone =
+      (typeof Zone === 'undefined') ? null : this.ngZone.run(() => Zone.current);
+
   constructor(private componentFactory: ComponentFactory<any>, private injector: Injector) {}
 
   /**
@@ -80,16 +87,19 @@ export class ComponentNgElementStrategy implements NgElementStrategy {
    * destruction.
    */
   connect(element: HTMLElement) {
-    // If the element is marked to be destroyed, cancel the task since the component was reconnected
-    if (this.scheduledDestroyFn !== null) {
-      this.scheduledDestroyFn();
-      this.scheduledDestroyFn = null;
-      return;
-    }
+    this.runInZone(() => {
+      // If the element is marked to be destroyed, cancel the task since the component was
+      // reconnected
+      if (this.scheduledDestroyFn !== null) {
+        this.scheduledDestroyFn();
+        this.scheduledDestroyFn = null;
+        return;
+      }
 
-    if (this.componentRef === null) {
-      this.initializeComponent(element);
-    }
+      if (this.componentRef === null) {
+        this.initializeComponent(element);
+      }
+    });
   }
 
   /**
@@ -97,19 +107,21 @@ export class ComponentNgElementStrategy implements NgElementStrategy {
    * being moved across the DOM.
    */
   disconnect() {
-    // Return if there is no componentRef or the component is already scheduled for destruction
-    if (this.componentRef === null || this.scheduledDestroyFn !== null) {
-      return;
-    }
-
-    // Schedule the component to be destroyed after a small timeout in case it is being
-    // moved elsewhere in the DOM
-    this.scheduledDestroyFn = scheduler.schedule(() => {
-      if (this.componentRef !== null) {
-        this.componentRef.destroy();
-        this.componentRef = null;
+    this.runInZone(() => {
+      // Return if there is no componentRef or the component is already scheduled for destruction
+      if (this.componentRef === null || this.scheduledDestroyFn !== null) {
+        return;
       }
-    }, DESTROY_DELAY);
+
+      // Schedule the component to be destroyed after a small timeout in case it is being
+      // moved elsewhere in the DOM
+      this.scheduledDestroyFn = scheduler.schedule(() => {
+        if (this.componentRef !== null) {
+          this.componentRef.destroy();
+          this.componentRef = null;
+        }
+      }, DESTROY_DELAY);
+    });
   }
 
   /**
@@ -117,11 +129,13 @@ export class ComponentNgElementStrategy implements NgElementStrategy {
    * retrieved from the cached initialization values.
    */
   getInputValue(property: string): any {
-    if (this.componentRef === null) {
-      return this.initialInputValues.get(property);
-    }
+    return this.runInZone(() => {
+      if (this.componentRef === null) {
+        return this.initialInputValues.get(property);
+      }
 
-    return this.componentRef.instance[property];
+      return this.componentRef.instance[property];
+    });
   }
 
   /**
@@ -129,22 +143,24 @@ export class ComponentNgElementStrategy implements NgElementStrategy {
    * cached and set when the component is created.
    */
   setInputValue(property: string, value: any): void {
-    if (this.componentRef === null) {
-      this.initialInputValues.set(property, value);
-      return;
-    }
+    this.runInZone(() => {
+      if (this.componentRef === null) {
+        this.initialInputValues.set(property, value);
+        return;
+      }
 
-    // Ignore the value if it is strictly equal to the current value, except if it is `undefined`
-    // and this is the first change to the value (because an explicit `undefined` _is_ strictly
-    // equal to not having a value set at all, but we still need to record this as a change).
-    if (strictEquals(value, this.getInputValue(property)) &&
-        !((value === undefined) && this.unchangedInputs.has(property))) {
-      return;
-    }
+      // Ignore the value if it is strictly equal to the current value, except if it is `undefined`
+      // and this is the first change to the value (because an explicit `undefined` _is_ strictly
+      // equal to not having a value set at all, but we still need to record this as a change).
+      if (strictEquals(value, this.getInputValue(property)) &&
+          !((value === undefined) && this.unchangedInputs.has(property))) {
+        return;
+      }
 
-    this.recordInputChange(property, value);
-    this.componentRef.instance[property] = value;
-    this.scheduleDetectChanges();
+      this.recordInputChange(property, value);
+      this.componentRef.instance[property] = value;
+      this.scheduleDetectChanges();
+    });
   }
 
   /**
@@ -263,5 +279,10 @@ export class ComponentNgElementStrategy implements NgElementStrategy {
 
     this.callNgOnChanges(this.componentRef);
     this.componentRef.changeDetectorRef.detectChanges();
+  }
+
+  /** Runs in the angular zone, if present. */
+  private runInZone(fn: () => unknown) {
+    return (this.elementZone && Zone.current !== this.elementZone) ? this.ngZone.run(fn) : fn();
   }
 }
--- a/packages/elements/test/component-factory-strategy_spec.ts
+++ b/packages/elements/test/component-factory-strategy_spec.ts
@@ -6,7 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {ComponentFactory, ComponentRef, Injector, NgModuleRef, SimpleChange, SimpleChanges, Type} from '@angular/core';
+import {ApplicationRef, ComponentFactory, ComponentFactoryResolver, ComponentRef, Injector, NgModuleRef, NgZone, SimpleChange, SimpleChanges, Type} from '@angular/core';
 import {fakeAsync, tick} from '@angular/core/testing';
 import {Subject} from 'rxjs';
 
@@ -20,22 +20,40 @@ describe('ComponentFactoryNgElementStrategy', () => {
   let injector: any;
   let componentRef: any;
   let applicationRef: any;
+  let ngZone: any;
+
+  let injectables: Map<unknown, unknown>;
 
   beforeEach(() => {
     factory = new FakeComponentFactory();
     componentRef = factory.componentRef;
 
     applicationRef = jasmine.createSpyObj('applicationRef', ['attachView']);
+
+    ngZone = jasmine.createSpyObj('ngZone', ['run']);
+    ngZone.run.and.callFake((fn: () => unknown) => fn());
+
     injector = jasmine.createSpyObj('injector', ['get']);
-    injector.get.and.returnValue(applicationRef);
+    injector.get.and.callFake((token: unknown) => {
+      if (!injectables.has(token)) {
+        throw new Error(`Failed to get injectable from mock injector: ${token}`);
+      }
+      return injectables.get(token);
+    });
+
+    injectables = new Map<unknown, unknown>([
+      [ApplicationRef, applicationRef],
+      [NgZone, ngZone],
+    ]);
 
     strategy = new ComponentNgElementStrategy(factory, injector);
+    ngZone.run.calls.reset();
   });
 
   it('should create a new strategy from the factory', () => {
     const factoryResolver = jasmine.createSpyObj('factoryResolver', ['resolveComponentFactory']);
     factoryResolver.resolveComponentFactory.and.returnValue(factory);
-    injector.get.and.returnValue(factoryResolver);
+    injectables.set(ComponentFactoryResolver, factoryResolver);
 
     const strategyFactory = new ComponentNgElementStrategyFactory(FakeComponent, injector);
     expect(strategyFactory.create(injector)).toBeTruthy();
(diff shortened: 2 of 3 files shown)


kseamon:
As I had commented way back, what this is still missing is exiting the NgZone in outputs - we don't want to leak our app's NgZone to the parent app (which would cause unrelated activities in the parent app to trigger change detection within the wrapped child application).

gkalpak:
For reference: There is some relevant discussion in #24185.

gkalpak:
@kseamon, can you give an example of the problem (I don't quite understand how we can "leak" our Zone).
Also, is this issue something that is introduced by this PR? Or is it possible that this "Zone leaking" can happen today (without the PR)?

kseamon:
It would look something like this:

// Parent app
ngElementNode.addEventListener('customEvent', () => {
  setTimeout(() => {
    doSomethingUnrelated();
    // Whoops we just triggered CD within the Angular Element.
  });
});

gkalpak:
Since this change is unrelated to this PR, I consider it outside of scope for this PR. Feel free to create an issue about so we can discuss there.

##
On a quick note:

I wouldn't expect Zones to be leaked the way you described (neither with nor without the changes in this PR). If I am not mistaken, what matters is the Zone in which you subscribe to the observable and this happens in the CustomElement's [connectedCallback()][1] and is not affected by the strategy's `runInZone()`.

If anything, we might need to enter the Zone in the event listener callback, if the element is created in a way that causes its `connectedCallback()` method to be called outside of the Angular Zone 😛

<sub>I only took a quick look, so I may have missed something. We should take a closer look and look at concrete examples (both with and without CustomElement polyfills).</sub>

[1]: https://github.com/angular/angular/blob/71956250ddc7faee2a813c293f4ebf91a1509afa/packages/elements/src/create-custom-element.ts#L189-L214

packages/elements/src/component-factory-strategy.ts
```diff
@@ -71,78 +71,108 @@ export class ComponentNgElementStrategy implements NgElementStrategy {
    */
   private readonly unchangedInputs = new Set<string>();
 
-  constructor(private componentFactory: ComponentFactory<any>, private injector: Injector) {}
+  /** check for zone.js loaded */
+  private readonly zoneless: boolean;
```
gkalpak, line 75: Let's move the initialization here. I would also consider renaming this to `isZoneless`:

```suggestion
  private readonly isZoneless = typeof Zone === 'undefined';
```

packages/elements/src/component-factory-strategy.ts
```diff
@@ -71,78 +71,108 @@ export class ComponentNgElementStrategy implements NgElementStrategy {
...
+      this.recordInputChange(property, value);
+      this.componentRef.instance[property] = value;
+      this.scheduleDetectChanges();
+    });
+  }
 
-    this.recordInputChange(property, value);
-    this.componentRef.instance[property] = value;
-    this.scheduleDetectChanges();
+  private runInZone(fn: () => any) {
```
gkalpak, line 174: Please move private methods below protected ones. (Although I have no idea why they are protected instead of private 🤷)

gkalpak, line 174: Please use `unknown` instead of `any` whenever possible.

packages/elements/src/component-factory-strategy.ts
```diff
@@ -71,78 +71,108 @@ export class ComponentNgElementStrategy implements NgElementStrategy {
    */
   private readonly unchangedInputs = new Set<string>();
 
-  constructor(private componentFactory: ComponentFactory<any>, private injector: Injector) {}
+  /** check for zone.js loaded */
+  private readonly zoneless: boolean;
+
+  /** service for setting zone context */
```
gkalpak, line 77:
```suggestion
  /** Service for setting zone context. */
```

(for consistency 😇)

packages/elements/src/component-factory-strategy.ts
```diff
@@ -71,78 +71,108 @@ export class ComponentNgElementStrategy implements NgElementStrategy {
...
+  private readonly ngZone: NgZone;
```
gkalpak, line 78: Let's move the initialization here:

```suggestion
  private readonly ngZone = this.injector.get<NgZone>(NgZone);
```

packages/elements/src/component-factory-strategy.ts
```diff
@@ -71,78 +71,108 @@ export class ComponentNgElementStrategy implements NgElementStrategy {
...
+
+  /** the zone the element was created in */
```
gkalpak, line 80:
```suggestion
  /** The zone the element was created in or `null` if Zone.js is not loaded. */
```

packages/elements/src/component-factory-strategy.ts
```diff
@@ -71,78 +71,108 @@ export class ComponentNgElementStrategy implements NgElementStrategy {
...
+  private readonly elementZone: Zone|undefined;
```
gkalpak, line 81: Can we move the initialization here?

```suggestion
  private readonly elementZone = (typeof Zone === 'undefined') ? null : this.ngZone.run(() => Zone.current);
```

packages/elements/src/component-factory-strategy.ts
gkalpak, line 75: Actually, do we need this extra property? Couldn't we use `elementZone` being `undefined` (or `null`) instead?

packages/elements/test/component-factory-strategy_spec.ts
```diff
@@ -239,6 +255,39 @@ describe('ComponentFactoryNgElementStrategy', () => {
          expect(componentRef.destroy).toHaveBeenCalledTimes(1);
        }));
   });
+
+  describe('runInZone', () => {
+    it('should have zones (baseline)', () => {
+      expect(strategy['zoneless']).toEqual(false);
+      expect(strategy['ngZone']).toBeTruthy();
+      expect(strategy['elementZone']).toBeTruthy();
+    });
```
gkalpak, line 264: This doesn't seem to add much value (just asserting internal implementation details). Let's remove it.

packages/elements/test/component-factory-strategy_spec.ts
```diff
@@ -239,6 +255,39 @@ describe('ComponentFactoryNgElementStrategy', () => {
...
+
+    it('should run already in original zone', () => {
+      let called = false;
+
+      ngZone.run.calls.reset();
```
gkalpak, line 269: Is this necessary?

packages/elements/test/component-factory-strategy_spec.ts
```diff
@@ -239,6 +255,39 @@ describe('ComponentFactoryNgElementStrategy', () => {
...
   });
+
+  describe('runInZone', () => {
+    it('should have zones (baseline)', () => {
+      expect(strategy['zoneless']).toEqual(false);
+      expect(strategy['ngZone']).toBeTruthy();
+      expect(strategy['elementZone']).toBeTruthy();
+    });
+
+    it('should run already in original zone', () => {
```
gkalpak, line 266: Consider something more descriptive, such as:

```suggestion
    it('should run the callback directly when invoked in element\'s zone', () => {
```

packages/elements/test/component-factory-strategy_spec.ts
```diff
@@ -239,6 +255,39 @@ describe('ComponentFactoryNgElementStrategy', () => {
...
+
+      strategy['runInZone'](() => {
+        called = true;
+      });
+
+      expect(called).toEqual(true);
+      expect(ngZone.run).not.toHaveBeenCalled();
+    });
+
+    it('should run placed in original zone', () => {
```
gkalpak, line 279:
```suggestion
    it('should run the callback inside the element\'s zone when invoked in a different zone', () => {
```

packages/elements/test/component-factory-strategy_spec.ts
```diff
@@ -239,6 +255,39 @@ describe('ComponentFactoryNgElementStrategy', () => {
...
+      let called = false;
+      ngZone.run.calls.reset();
```
gkalpak, line 281: Is this necessary?

packages/elements/test/component-factory-strategy_spec.ts
```diff
@@ -239,6 +255,39 @@ describe('ComponentFactoryNgElementStrategy', () => {
...
+
+      Zone.root.run(() => (strategy['runInZone'](() => {
+                      called = true;
+                    })));
+
+      expect(called).toEqual(true);
+      expect(ngZone.run).toHaveBeenCalled();
```
gkalpak, line 288: Can you also verify that it was called with the correct callback?

packages/elements/test/component-factory-strategy_spec.ts
```diff
@@ -239,6 +255,39 @@ describe('ComponentFactoryNgElementStrategy', () => {
...
+    });
+  });
```
gkalpak, line 290: Please add some tests emulating a zoneless setup.

packages/elements/test/component-factory-strategy_spec.ts
```diff
@@ -20,22 +20,38 @@ describe('ComponentFactoryNgElementStrategy', () => {
...
   beforeEach(() => {
     factory = new FakeComponentFactory();
     componentRef = factory.componentRef;
 
     applicationRef = jasmine.createSpyObj('applicationRef', ['attachView']);
+
+    ngZone = jasmine.createSpyObj('ngZone', ['run']);
+    ngZone.run.and.callFake((fn: any) => {
+      return fn();
+    });
```
gkalpak, line 35:
```suggestion
    ngZone.run.and.callFake((fn: () => unknown) => fn());
```

packages/elements/test/component-factory-strategy_spec.ts
```diff
@@ -20,22 +20,38 @@ describe('ComponentFactoryNgElementStrategy', () => {
...
+    injector.get.and.callFake(function(identify: any) {
+      const name = identify && identify.name;
+      if (name === 'ApplicationRef') {
+        return applicationRef;
+      } else if (name === 'NgZone') {
+        return ngZone;
+      } else if (name === 'ComponentFactoryResolver') {
+        return factoryResolver;
+      }
+    });
```
gkalpak, line 47: This implementation does not look very robust or easy to work with. How about something like:

```ts
describe('ComponentFactoryNgElementStrategy', () => {
  ...
  let injectables: Map<Type, unknown>;

  beforeEach(() => {
    ...

    injector = jasmine.createSpyObj('injector', ['get']);
    injector.get.and.callFake((token: Type) => {
      if (!injectables.has(token)) {
        throw new Error(`Failed to get injectable from mock injector: ${token}`);
      }

      injectables.get(token);
    });

    ...

    injectables = new Map([
      [ApplicationRef, applicationRef],
      [NgZone, ngZone],
    ]);
  });

  ...
});
```

Then, if a test needs to inject somethine else, it can do `injectables.set(<token>, <value>). E.g.:

```ts
injectables.set(ComponentFactoryResolver, factoryResolver);
```

gkalpak (changes requested):
Thx for working on this, @remackgeek!

I think this is heading into the right direction. I have left a few minor comments.
Can you also please rebase on latest master and squash the commits into one?

##
BTW, it would be great to have a more realistic test via an actual example app (either in [packages/examples/][1] or in [integration/][2] (similar to or as part of [integration/ng_elements/][3])), but that might be too much to ask for this PR :grin:

[1]: https://github.com/angular/angular/tree/71956250ddc7faee2a813c293f4ebf91a1509afa/packages/examples
[2]: https://github.com/angular/angular/blob/71956250ddc7faee2a813c293f4ebf91a1509afa/integration
[3]: https://github.com/angular/angular/blob/71956250ddc7faee2a813c293f4ebf91a1509afa/integration/ng_elements

packages/elements/test/component-factory-strategy_spec.ts
remackgeek: Yes, ngZone.run() gets called once during construction of the strategy (setting elementZone).

packages/elements/test/component-factory-strategy_spec.ts
remackgeek: I'm not sure what you're asking for here. Don't those 2 expectations show that the callback was called and in the correct way?

packages/elements/test/component-factory-strategy_spec.ts
remackgeek: Yes, ngZone.run() gets called once during construction of the strategy (setting elementZone).

packages/elements/test/component-factory-strategy_spec.ts
gkalpak: The current assertions prove that the callback was called and that `ngZone.run()` was called. But they don't prove that the callback was called by `ngZone.run()`. (For all we know, we could be doing something like `callback(); ngZone.run(() => {});` and the test would still pass :grin:)

packages/elements/test/component-factory-strategy_spec.ts
gkalpak: Right. In that case, can you move the `ngZone.run.calls.reset()` call right below the strategy construction?

packages/elements/test/component-factory-strategy_spec.ts
```diff
@@ -20,22 +20,41 @@ describe('ComponentFactoryNgElementStrategy', () => {
   let injector: any;
   let componentRef: any;
   let applicationRef: any;
+  let ngZone: any;
+  let factoryResolver: any;
```
gkalpak, line 24: This is only used in one tests. There is no benefit in having it as a `describe`-wide variable here.

packages/elements/test/component-factory-strategy_spec.ts
```diff
@@ -20,22 +20,41 @@ describe('ComponentFactoryNgElementStrategy', () => {
...
+      [NgZone, ngZone],
+    ]);
 
     strategy = new ComponentNgElementStrategy(factory, injector);
+    ngZone.run.calls.reset();
   });
 
   it('should create a new strategy from the factory', () => {
-    const factoryResolver = jasmine.createSpyObj('factoryResolver', ['resolveComponentFactory']);
+    factoryResolver = jasmine.createSpyObj('factoryResolver', ['resolveComponentFactory']);
```
gkalpak, line 55:
```suggestion
    const factoryResolver = jasmine.createSpyObj('factoryResolver', ['resolveComponentFactory']);
```

packages/elements/test/component-factory-strategy_spec.ts
```diff
@@ -266,6 +285,31 @@ describe('ComponentFactoryNgElementStrategy', () => {
...
+
+    it('should run the callback directly when invoked in element\'s zone', () => {
+      expect(strategy['runInZone'](fn)).toEqual('foofoo');
+      expect(ngZone.run).not.toHaveBeenCalled();
+    });
+
+    it('should run the callback inside the element\'s zone when invoked in a different zone',
+       () => {
+         expect(Zone.root.run(() => (strategy['runInZone'](fn)))).toEqual('foofoo');
+         expect(ngZone.run).toHaveBeenCalled();
```
gkalpak, line 301: This is now redundant (since it is covered by the `toHaveBeenCalledWith()` below):

```suggestion
```

packages/elements/test/component-factory-strategy_spec.ts
```diff
@@ -266,6 +285,31 @@ describe('ComponentFactoryNgElementStrategy', () => {
...
+         expect(ngZone.run).toHaveBeenCalledWith(fn);
+       });
+
+    it('should run the callback directly when called without without zone.js loaded', () => {
```
gkalpak, line 305:
```suggestion
    it('should run the callback directly when called without zone.js loaded', () => {
```

packages/elements/test/component-factory-strategy_spec.ts
```diff
@@ -20,22 +20,41 @@ describe('ComponentFactoryNgElementStrategy', () => {
   let injector: any;
   let componentRef: any;
   let applicationRef: any;
+  let ngZone: any;
+  let factoryResolver: any;
+
+  let injectables: Map<unknown, unknown>;
```
gkalpak, line 26: Couldn't this be `Map<Type, unknown>` (or `Map<Type<unknown>, unknown>`)?

gkalpak (changes requested):

packages/elements/test/component-factory-strategy_spec.ts
remackgeek: Those don't work because `ComponentFactoryResolver` is an abstract class. You get this:
**Argument of type 'typeof ComponentFactoryResolver' is not assignable to parameter of type 'Type<unknown>'.
  Cannot assign an abstract constructor type to a non-abstract constructor type.ts(2345)**

Using `unknown` makes this inconsistent:
` injector.get.and.callFake((token: Type<unknown>) => {`
but typescript didn't catch it and it works at runtime. Should I change it to `unknown` or leave it because it's more descriptive of what's going on?

gkalpak: Yeah, in that case, `unknown` sounds good.

gkalpak (approved):
LGTM - Thx, @remackgeek ✨
Can you please update [goldens/size-tracking/aio-payloads.json#L15][1] to the new size?

For reference, this PR increases the angular.io main bundle by 295B (451100B --> 451395B). I inspected the bundle and confirmed that the only difference is the updated `ComponentFactoryNgStrategy` class.

[1]: https://github.com/angular/angular/blob/196bfa8fae6b7da9df2601a400164e35e45cefd6/goldens/size-tracking/aio-payloads.json#L15

jelbourn (approved):
LGTM

Reviewed-for: size-tracking

petebacondarwin:
Can you rebase and fix the size-tracking conflict so that we can approve for `size-tracking`?

remackgeek:
Rebased, broken size tracking changed to match actuals (not sure what else to put there).

petebacondarwin (approved):
Reviewed-for: size-tracking

petebacondarwin:
Awesome thanks!

kyliau:
presubmit: http://test/OCL:321780152:BASE:321780169:1594998439408:2f37a981

AndrewKushnir:
[Presubmit #2](http://test/OCL:322471863:BASE:322471888:1595375776801:19f72b8f) (will also start global presubmit in a couple hours).

Issue #37697, opened 2020-06-23
Forms: consider using `ngDevMode` to tree-shake error messages text in prod mode


AndrewKushnir:
In Ivy runtime code, the `ngDevMode` const is used to guard asserts required in dev mode only. We should consider doing the same in Forms package to tree-shake error messages text (which cannot be minified) and some dev mode specific checks. That should help reduce the size of the `forms` package in production mode.

sonukapoor:
@AndrewKushnir I can assist with that.

PR #37821, opened 2020-06-29
perf(forms): use `ngDevMode` to tree-shake error messages

sonukapoor:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #37697

- [ ] Yes
- [X] No

### DIFF
--- a/packages/forms/src/directives/ng_control.ts
+++ b/packages/forms/src/directives/ng_control.ts
@@ -13,7 +13,9 @@ import {ControlValueAccessor} from './control_value_accessor';
 import {AsyncValidator, AsyncValidatorFn, Validator, ValidatorFn} from './validators';
 
 function unimplemented(): any {
-  throw new Error('unimplemented');
+  if (typeof ngDevMode === 'undefined' || ngDevMode) {
+    throw new Error('unimplemented');
+  }
 }
 
 /**
--- a/packages/forms/src/directives/ng_model.ts
+++ b/packages/forms/src/directives/ng_model.ts
@@ -317,18 +317,20 @@ export class NgModel extends NgControl implements OnChanges, OnDestroy {
   }
 
   private _checkParentType(): void {
-    if (!(this._parent instanceof NgModelGroup) &&
-        this._parent instanceof AbstractFormGroupDirective) {
-      TemplateDrivenErrors.formGroupNameException();
-    } else if (!(this._parent instanceof NgModelGroup) && !(this._parent instanceof NgForm)) {
-      TemplateDrivenErrors.modelParentException();
+    if (typeof ngDevMode === 'undefined' || ngDevMode) {
+      if (!(this._parent instanceof NgModelGroup) &&
+          this._parent instanceof AbstractFormGroupDirective) {
+        TemplateDrivenErrors.formGroupNameException();
+      } else if (!(this._parent instanceof NgModelGroup) && !(this._parent instanceof NgForm)) {
+        TemplateDrivenErrors.modelParentException();
+      }
     }
   }
 
   private _checkName(): void {
     if (this.options && this.options.name) this.name = this.options.name;
 
-    if (!this._isStandalone() && !this.name) {
+    if (!this._isStandalone() && !this.name && (typeof ngDevMode === 'undefined' || ngDevMode)) {
       TemplateDrivenErrors.missingNameException();
     }
   }
--- a/packages/forms/src/directives/radio_control_value_accessor.ts
+++ b/packages/forms/src/directives/radio_control_value_accessor.ts
@@ -17,6 +17,13 @@ export const RADIO_VALUE_ACCESSOR: any = {
   multi: true
 };
 
+function throwNameError() {
+  throw new Error(`
+      If you define both a name and a formControlName attribute on your radio button, their values
+      must match. Ex: <input type="radio" formControlName="food" name="food">
+    `);
+}
+
 /**
  * @description
  * Class used by Angular to track radio buttons. For internal use only.
@@ -213,16 +220,10 @@ export class RadioControlValueAccessor implements ControlValueAccessor, OnDestro
   }
 
   private _checkName(): void {
-    if (this.name && this.formControlName && this.name !== this.formControlName) {
-      this._throwNameError();
+    if (this.name && this.formControlName && this.name !== this.formControlName &&
+        (typeof ngDevMode === 'undefined' || ngDevMode)) {
+      throwNameError();
     }
     if (!this.name && this.formControlName) this.name = this.formControlName;
-  }
-
-  private _throwNameError(): void {
-    throw new Error(`
-      If you define both a name and a formControlName attribute on your radio button, their values
-      must match. Ex: <input type="radio" formControlName="food" name="food">
-    `);
   }
 }
--- a/packages/forms/src/directives/reactive_directives/form_control_directive.ts
+++ b/packages/forms/src/directives/reactive_directives/form_control_directive.ts
@@ -68,11 +68,13 @@ export class FormControlDirective extends NgControl implements OnChanges {
 
   /**
    * @description
-   * Triggers a warning that this input should not be used with reactive forms.
+   * Triggers a warning in dev mode that this input should not be used with reactive forms.
    */
   @Input('disabled')
   set isDisabled(isDisabled: boolean) {
-    ReactiveErrors.disabledAttrWarning();
+    if (typeof ngDevMode === 'undefined' || ngDevMode) {
+      ReactiveErrors.disabledAttrWarning();
+    }
   }
 
   // TODO(kara): remove next 4 properties once deprecation period is over
--- a/packages/forms/src/directives/shared.ts
+++ b/packages/forms/src/directives/shared.ts
@@ -33,8 +33,10 @@ export function controlPath(name: string|null, parent: ControlContainer): string
 }
 
 export function setUpControl(control: FormControl, dir: NgControl): void {
-  if (!control) _throwError(dir, 'Cannot find control with');
-  if (!dir.valueAccessor) _throwError(dir, 'No value accessor for form control with');
+  if (typeof ngDevMode === 'undefined' || ngDevMode) {
+    if (!control) _throwError(dir, 'Cannot find control with');
+    if (!dir.valueAccessor) _throwError(dir, 'No value accessor for form control with');
+  }
 
   control.validator = Validators.compose([control.validator!, dir.validator]);
   control.asyncValidator = Validators.composeAsync([control.asyncValidator!, dir.asyncValidator]);
@@ -64,8 +66,14 @@ export function setUpControl(control: FormControl, dir: NgControl): void {
 }
 
 export function cleanUpControl(control: FormControl, dir: NgControl) {
-  dir.valueAccessor!.registerOnChange(() => _noControlError(dir));
-  dir.valueAccessor!.registerOnTouched(() => _noControlError(dir));
+  const noop = () => {
+    if (typeof ngDevMode === 'undefined' || ngDevMode) {
+      _noControlError(dir);
+    }
+  };
+
+  dir.valueAccessor!.registerOnChange(noop);
+  dir.valueAccessor!.registerOnTouched(noop);
 
   dir._rawValidators.forEach((validator: any) => {
     if (validator.registerOnValidatorChange) {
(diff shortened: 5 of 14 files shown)


packages/forms/src/directives/ng_control.ts
```diff
@@ -13,7 +13,9 @@ import {ControlValueAccessor} from './control_value_accessor';
 import {AsyncValidator, AsyncValidatorFn, Validator, ValidatorFn} from './validators';
 
 function unimplemented(): any {
-  throw new Error('unimplemented');
+  if (ngDevMode) {
+    throw new Error('unimplemented');
```
alfaproject, line 17: Might as well not even call this method so that we don't bundle a NOOP method

AndrewKushnir:
FYI, I just did a quick test with a demo app and ReactiveForms module. The `@angular/forms` package got reduced from 41kb to 34kb (which is around 20% decrease) 🎉

We should consider creating a test app with all Forms features (Reactive, Template-driven, validators, etc) in the `integration` folder and have a size check setup there as well, so that we can monitor the effect of the changes performed in different PRs in the `@angular/forms` package size.

packages/forms/src/directives/shared.ts
```diff
@@ -120,13 +120,16 @@ function setUpModelChangePipeline(control: FormControl, dir: NgControl): void {
...
     control: FormGroup|FormArray, dir: AbstractFormGroupDirective|FormArrayName) {
-  if (control == null) _throwError(dir, 'Cannot find control with');
+  if (control == null && ngDevMode) _throwError(dir, 'Cannot find control with');
   control.validator = Validators.compose([control.validator, dir.validator]);
   control.asyncValidator = Validators.composeAsync([control.asyncValidator, dir.asyncValidator]);
 }
 
 function _noControlError(dir: NgControl) {
-  return _throwError(dir, 'There is no FormControl instance attached to form control element with');
+  if (ngDevMode) {
```
AndrewKushnir, line 129: I think in this case we can keep the `_noControlError` as is and add the `ngDevMode` check before calling `_noControlError`, so we don't include a noop function.

packages/forms/src/directives/radio_control_value_accessor.ts
```diff
@@ -10,6 +10,7 @@ import {Directive, ElementRef, forwardRef, Injectable, Injector, Input, OnDestro
 
 import {ControlValueAccessor, NG_VALUE_ACCESSOR} from './control_value_accessor';
 import {NgControl} from './ng_control';
+import {RadioControlValueAccessorErrors} from './radio_control_value_accessor_errors';
```
AndrewKushnir, line 13: We should probably just keep the `throwNameError` function in this file and make it a top level one (i.e. extract from the class).

packages/forms/src/directives/reactive_directives/form_control_directive.ts
```diff
@@ -68,11 +68,13 @@ export class FormControlDirective extends NgControl implements OnChanges {
 
   /**
    * @description
-   * Triggers a warning that this input should not be used with reactive forms.
+   * Triggers a warning in dev mode that this input should not be used with reactive forms.
    */
   @Input('disabled')
   set isDisabled(isDisabled: boolean) {
-    ReactiveErrors.disabledAttrWarning();
+    if (ngDevMode) {
```
AndrewKushnir, line 75: This is an interesting case: the input is used for dev-mode-only checks, so we should check if we can also tree-shake it away with `ngDevMode` (otherwise we would include a noop function).

packages/forms/src/directives/ng_model.ts
```diff
@@ -317,18 +317,20 @@ export class NgModel extends NgControl implements OnChanges, OnDestroy {
...
+      if (!(this._parent instanceof NgModelGroup) &&
+          this._parent instanceof AbstractFormGroupDirective) {
+        TemplateDrivenErrors.formGroupNameException();
+      } else if (!(this._parent instanceof NgModelGroup) && !(this._parent instanceof NgForm)) {
+        TemplateDrivenErrors.modelParentException();
+      }
     }
   }
 
   private _checkName(): void {
```
AndrewKushnir, line 330: It looks like we can make additional refactoring here: this function can be extracted from the class and we can put the `ngDevMode` check before calling it. The following line:
```
  if (this.options && this.options.name) this.name = this.options.name;
```
Can be moved into the main code (before calling the `_checkName` function).

AndrewKushnir (commented):
Thanks for these changes @sonukapoor, as I mentioned in my previous comment in the thread, it helps save around 7kb (or ~20%).

I've left a few comments, mainly around additional refactoring that can be made to avoid having noop functions (when the whole body of the function is guarded with `ngDevMode` check).

Another thing that we should consider is to make sure this works correctly with View Engine (since `ngDevMode` was mostly used in Ivy runtime code).

Thank you.

packages/forms/src/directives/shared.ts
```diff
@@ -33,8 +33,10 @@ export function controlPath(name: string|null, parent: ControlContainer): string
 }
 
 export function setUpControl(control: FormControl, dir: NgControl): void {
-  if (!control) _throwError(dir, 'Cannot find control with');
-  if (!dir.valueAccessor) _throwError(dir, 'No value accessor for form control with');
+  if (!control && (typeof ngDevMode === 'undefined' || ngDevMode))
```
AndrewKushnir, line 36: For readability, I think we can just wrap 2 `if` statements like this:
```
  if (typeof ngDevMode === 'undefined' || ngDevMode) {
    if (!control) _throwError(dir, 'Cannot find control with');
    if (!dir.valueAccessor) _throwError(dir, 'No value accessor for form control with');
  }
```

packages/forms/src/directives/shared.ts
```diff
@@ -64,8 +66,10 @@ export function setUpControl(control: FormControl, dir: NgControl): void {
 }
 
 export function cleanUpControl(control: FormControl, dir: NgControl) {
-  dir.valueAccessor!.registerOnChange(() => _noControlError(dir));
-  dir.valueAccessor!.registerOnTouched(() => _noControlError(dir));
+  if (typeof ngDevMode === 'undefined' || ngDevMode) {
+    dir.valueAccessor!.registerOnChange(() => _noControlError(dir));
```
AndrewKushnir, line 70: In this particular case we can not wrap the logic into the `ngDevMode` check - the `dir.valueAccessor!.registerOnChange` and `dir.valueAccessor!.registerOnTouched` calls should still happen (in both dev and prod modes), since it resets the subscriptions. We can do something like this:
```
  const noop = () => {
    if (typeof ngDevMode === 'undefined' || ngDevMode) {
      _noControlError(dir);
    }
  };
  dir.valueAccessor!.registerOnChange(noop);
  dir.valueAccessor!.registerOnTouched(noop);
```

packages/forms/src/directives/ng_model.ts
AndrewKushnir: I looked at this code again and I think we can make this refactor later and have only `ngDevMode`-only changes in this PR.

packages/forms/src/directives/reactive_directives/form_control_directive.ts
```diff
@@ -68,11 +68,14 @@ export class FormControlDirective extends NgControl implements OnChanges {
 
   /**
    * @description
-   * Triggers a warning that this input should not be used with reactive forms.
+   * Triggers a warning in dev mode that this input should not be used with reactive forms.
    */
+
```
AndrewKushnir, line 73: Let's remove this extra line.

packages/forms/src/directives/ng_control.ts
```diff
@@ -67,7 +67,11 @@ export abstract class NgControl extends AbstractControlDirective {
    * @throws An exception that this method is not implemented
    */
   get validator(): ValidatorFn|null {
-    return <ValidatorFn>unimplemented();
+    if (typeof ngDevMode === 'undefined' || ngDevMode) {
+      return <ValidatorFn>unimplemented();
+    } else {
+      return null;
```
AndrewKushnir, line 73: I think your initial change was actually good (i.e. guard the `throw` with `ngDevMode`). I didn't realize that we actually return from this function and having a return value dependent on `ngDevMode` doesn't look good correct. Could you please revert this change (i.e. move `ngDevMode` back to `unimplemented` function)?

AndrewKushnir:
Hi @sonukapoor, thanks for performing further updates! That's really great to see the dev-mode-only symbols go from the golden file 👍 I just left a quick comment on `unimplemented` function and also [started tests in Google's codebase](http://test/OCL:323096665:BASE:323096698:1595634088430:80a3e8e0). Will keep you updated. Thank you.

packages/forms/src/directives/ng_control.ts
sonukapoor: Done.

sonukapoor:
> Hi @sonukapoor, thanks for performing further updates! That's really great to see the dev-mode-only symbols go from the golden file 👍 I just left a quick comment on `unimplemented` function and also [started tests in Google's codebase](http://test/OCL:323096665:BASE:323096698:1595634088430:80a3e8e0). Will keep you updated. Thank you.

No worries. I have added the changes to the same commit.

muhammedMoussa:
> FYI, I just did a quick test with a demo app and ReactiveForms module. The `@angular/forms` package got reduced from 41kb to 34kb (which is around 20% decrease) 🎉
> 
> We should consider creating a test app with all Forms features (Reactive, Template-driven, validators, etc) in the `integration` folder and have a size check setup there as well, so that we can monitor the effect of the changes performed in different PRs in the `@angular/forms` package size.


Is `ngDevMode` mechanism used at other packages likes was in forms, maybe need to be updated as @sonukapoor suggestions, to reduce packages too

sonukapoor:
> Is `ngDevMode` mechanism used at other packages likes was in forms, maybe need to be updated as @sonukapoor suggestions, to reduce packages too

We can definitely look into that as well. @AndrewKushnir might already know more about this.

AndrewKushnir:
Hi @muhammedMoussa, the `ngDevMode` flag is actively used in Ivy runtime code to tree-shake away all dev-mode-only code in production builds. In this PR we are trying out this approach with Forms package. If everything goes well and there are no major blockers, we would consider doing the same in other packages too. Thank you.

packages/forms/src/directives/ng_control.ts
```diff
@@ -13,7 +13,9 @@ import {ControlValueAccessor} from './control_value_accessor';
 import {AsyncValidator, AsyncValidatorFn, Validator, ValidatorFn} from './validators';
 
 function unimplemented(): any {
-  throw new Error('unimplemented');
+  if (typeof ngDevMode === 'undefined' || ngDevMode) {
```
mhevery, line 16: I don't believe that calling `typeof ngDevMode === 'undefined'` is necessary. As long as angular code runs first it should be properly set up. The only place this would be an issue is if you used `ngDevMode` before call to angular (ie from the top level file context) So. I don't think it is necessary here.

mhevery (approved):
LGTM, 

but see my comment about `typeof ngDevMode === 'undefined'` (I don't think it is needed)

packages/forms/src/directives/ng_control.ts
AndrewKushnir: Thanks for the review Misko! 👍 

This code needs to run with both Ivy and View Engine and if I understand correctly, the `ngDevMode` would not be initialized in VE (thus there is a risk of having `ngDevMode` undeclared here)?

mhevery:
I see, yes I think you are right there.

jelbourn (approved):
LGTM

Reviewed-for: public-api

clydin (approved):

AndrewKushnir (approved):

clydin:
By using the example forms code found [here](https://github.com/angular/angular/blob/763023472b15f82a7551c9c4f8f43cda1d259dd5/packages/core/test/bundling/forms/index.ts) in combination with a newly generated CLI application, a savings of ~7kb was observed for both View Engine and Ivy renderers.

One area that might be beneficial to investigate is to potentially change the View Engine startup process to guarantee that `ngDevMode` is defined.  That should allow the removal of the `undefined` checks throughout this PR.

AndrewKushnir:
[Presubmit #2](http://test/OCL:326349299:BASE:326349329:1597278786661:ab80e26a).

AndrewKushnir:
FYI, tests in Google's codebase (global presubmit) went well.

Also adding a link to similar changes in Components repo for visibility: https://github.com/angular/components/pull/20146.

PR #37841, opened 2020-06-30
docs: add missing single quote

alan-agius4:
The current code is missing a single quote at the end of the import.

### DIFF
--- a/aio/content/guide/i18n.md
+++ b/aio/content/guide/i18n.md
@@ -834,7 +834,7 @@ Global variants of the locale data are available in [`@angular/common/locales/gl
 The following example imports the global variants for French (`fr`):
 
 <code-example language="javascript" header="app.module.ts">
-import '@angular/common/locales/global/fr;
+import '@angular/common/locales/global/fr';
 </code-example>
 
 {@a custom-id}


petebacondarwin (approved):

AndrewKushnir:
Hi @alan-agius4, it looks like this change was made from a branch that belongs to the `angular/angular` repo, not from your fork of the repo, thus the `deploy_aio` is failing and the merge script may not work. Could you please re-create this PR using a branch from your fork of Angular? Thank you.

alan-agius4:
@AndrewKushnir, I can do that, but there seems to be something wrong in the process because I clicked the edit button from the AIO page.

If this is not allowed/correct workflow, we should remove the edit button from the aio pages.

petebacondarwin:
I imagine that this happened because @alan-agius4 is a member of the angular/angular repo, so it allows him to create branches there. I see that when I click the edit button, I also have this issue, and there doesn't appear to be a way to create the branch on a different fork.

Screenshot 2020-06-30 at 18 55 45

Unfortunately there is no way for us to guess the person's fork, I think.

alan-agius4:
New pr git push --set-upstream origin intl-typo

petebacondarwin:
That being said, @AndrewKushnir, I wonder why the merge script should prevent such a local branch from being merged?

alan-agius4:
New pr https://github.com/angular/angular/pull/37854

AndrewKushnir:
Thanks for re-creating this PR @alan-agius4 👍 

@petebacondarwin, I'm not sure if this use-case is expected in the merge script and I've decided to be a bit more cautious. @devversion could you please let us know if this is a valid use-case for a merge script to work with PRs created based on a branch created directly in the `angular/angular` repo?

petebacondarwin:
It might be worth also checking with @gkalpak  to see if we could tweak the `deploy_aio` job to work with local branches?

PR #37866, opened 2020-07-01
fix(core): handle spaces after `select` and `plural` ICU keywords

AndrewKushnir:
Currently when the `plural` or `select` keywords in an ICU contain trailing spaces (e.g. `{count, select , ...}`), these spaces are also included into the key names in ICU vars (e.g. "VAR_SELECT "). These trailing spaces are not desirable, since they will later be converted into `_` symbols while normalizing placeholder names, thus causing mismatches at runtime (i.e. placeholder will not be replaced with the correct value). This commit updates the code to trim these spaces while generating an object with placeholders, to make sure the runtime logic can replace these placeholders with the right values.

Another approach that I was considering is to trim these spaces at the Lexer level, but that would be a big breaking change affecting a portion of apps due to the fact that i18n message ids would be changed in this case.

## PR Type

- [x] Bugfix

- [ ] Yes
- [x] No

### DIFF
--- a/packages/compiler-cli/test/compliance/r3_view_compiler_i18n_spec.ts
+++ b/packages/compiler-cli/test/compliance/r3_view_compiler_i18n_spec.ts
@@ -3692,7 +3692,7 @@ describe('i18n support in the template compiler', () => {
             $I18N_3$ = i0.ɵɵi18nPostprocess($I18N_3$, { "VAR_PLURAL": "\uFFFD1\uFFFD" });
           `;
 
-         verify(input, output, {verbose: true});
+         verify(input, output);
        });
   });
 
--- a/packages/compiler-cli/test/compliance/r3_view_compiler_i18n_spec.ts
+++ b/packages/compiler-cli/test/compliance/r3_view_compiler_i18n_spec.ts
@@ -3661,6 +3661,39 @@ describe('i18n support in the template compiler', () => {
 
       verify(input, output);
     });
+
+    it('should produce proper messages when `select` or `plural` keywords have spaces after them',
+       () => {
+         const input = `
+            <div i18n>
+              {count, select , 1 {one} other {more than one}}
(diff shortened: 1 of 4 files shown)


petebacondarwin (changes requested):
Hi @AndrewKushnir - I think we can fix this a bit further up the call stack without breaking the current message ids, but I might be wrong so please check this idea out and see what you think...

```diff
diff --git a/packages/compiler/src/render3/r3_template_transform.ts b/packages/compiler/src/render3/r3_template_transform.ts
index e91502af63..137fb7bff8 100644
--- a/packages/compiler/src/render3/r3_template_transform.ts
+++ b/packages/compiler/src/render3/r3_template_transform.ts
@@ -276,7 +276,8 @@ class HtmlAstToIvyAst implements html.Visitor {
         // ICU expression is a plain string, not wrapped into start
         // and end tags, so we wrap it before passing to binding parser
         const wrapped = `${config.start}${value}${config.end}`;
-        vars[key] = this._visitTextWithInterpolation(wrapped, expansion.sourceSpan) as t.BoundText;
+        vars[key.trim()] =
+            this._visitTextWithInterpolation(wrapped, expansion.sourceSpan) as t.BoundText;
       } else {
         placeholders[key] = this._visitTextWithInterpolation(value, expansion.sourceSpan);
       }
```

This is much simpler and doesn't require touching `i18nBindProps` and callers at all.

AndrewKushnir:
Thanks for the review @petebacondarwin 👍 Your proposal makes sense, I've updated the code. Could you please have another look when you get a chance? Thank you.

petebacondarwin (approved):
Works for me!

AndrewKushnir:
[Presubmit](http://test/OCL:319279695:BASE:319279732:1593633132713:b6e33fb1)

PR #37881, opened 2020-07-02
feat(forms): AbstractControl to store raw validators in addition to combined validators function

AndrewKushnir:
This commit refactors the way we store validators in AbstractControl-based classes: in addition to the combined validators function that we have, we also store the original list of validators. This is needed to have an ability to clean then up later at destroy time (currently it's problematic since they are combined in a single function).

The change preserves backwards compatibility by making sure public APIs stay the same. The only public API update is the change to the `AbstractControl` class constructor to extend the set of possible types that is can accept and process (which should not be breaking).

## PR Type

- [x] Feature

- [ ] Yes
- [x] No

### DIFF
--- a/packages/forms/src/model.ts
+++ b/packages/forms/src/model.ts
@@ -69,17 +69,24 @@ function _find(control: AbstractControl, path: Array<string|number>|string, deli
   return controlToFind;
 }
 
+/**
+ * Gets validators from either an options object or given validators.
+ */
 function pickValidators(validatorOrOpts?: ValidatorFn|ValidatorFn[]|AbstractControlOptions|
                         null): ValidatorFn|ValidatorFn[]|null {
   return (isOptionsObj(validatorOrOpts) ? validatorOrOpts.validators : validatorOrOpts) || null;
 }
 
-function coerceToValidator(validatorOrOpts?: ValidatorFn|ValidatorFn[]|AbstractControlOptions|
-                           null): ValidatorFn|null {
-  const validator = pickValidators(validatorOrOpts);
+/**
+ * Creates validator function by combining provided validators.
+ */
+function coerceToValidator(validator: ValidatorFn|ValidatorFn[]|null): ValidatorFn|null {
   return Array.isArray(validator) ? composeValidators(validator) : validator || null;
 }
 
+/**
+ * Gets async validators from either an options object or given validators.
+ */
 function pickAsyncValidators(
     asyncValidator?: AsyncValidatorFn|AsyncValidatorFn[]|null,
     validatorOrOpts?: ValidatorFn|ValidatorFn[]|AbstractControlOptions|null): AsyncValidatorFn|
@@ -87,13 +94,13 @@ function pickAsyncValidators(
   return (isOptionsObj(validatorOrOpts) ? validatorOrOpts.asyncValidators : asyncValidator) || null;
 }
 
-function coerceToAsyncValidator(
-    asyncValidator?: AsyncValidatorFn|AsyncValidatorFn[]|null,
-    validatorOrOpts?: ValidatorFn|ValidatorFn[]|AbstractControlOptions|null): AsyncValidatorFn|
-    null {
-  const origAsyncValidator = pickAsyncValidators(asyncValidator, validatorOrOpts);
-  return Array.isArray(origAsyncValidator) ? composeAsyncValidators(origAsyncValidator) :
-                                             origAsyncValidator || null;
+/**
+ * Creates async validator function by combining provided async validators.
+ */
+function coerceToAsyncValidator(asyncValidator?: AsyncValidatorFn|AsyncValidatorFn[]|
+                                null): AsyncValidatorFn|null {
+  return Array.isArray(asyncValidator) ? composeAsyncValidators(asyncValidator) :
+                                         asyncValidator || null;
 }
 
 export type FormHooks = 'change'|'blur'|'submit';
@@ -176,7 +183,7 @@ export abstract class AbstractControl {
    *
    * @internal
    */
-  private _coercedValidator: ValidatorFn|null;
+  private _composedValidatorFn: ValidatorFn|null;
 
   /**
    * Contains the result of merging asynchronous validators into a single validator function
@@ -184,7 +191,7 @@ export abstract class AbstractControl {
    *
    * @internal
    */
-  private _coercedAsyncValidator: AsyncValidatorFn|null;
+  private _composedAsyncValidatorFn: AsyncValidatorFn|null;
 
   /**
    * Synchronous validators as they were provided:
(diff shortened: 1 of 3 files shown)


AndrewKushnir:
[Presubmit](https://test.corp.google.com/ui#id=OCL:319333129:BASE:319333144:1593653273374:4142474e)

crisbeto:
If we end up going with this, would it make sense to expose the current validators on the `NgControl`? We've had a few long-standing issue reports on the Components repo where we can't detect and style something as required that uses `new FormControl(null, [Validators.required])`.

AndrewKushnir:
Hi @crisbeto, thanks for the comment. If everything goes well and we end up merging this change, it should help provide an API (in the future) to retrieve the list of currently applied validators (something like `control.getValidators()`). Could you please share a link to the relevant issue in Components repo, so I can have a look to get more context? Thank you.

crisbeto:
Here's the issue: https://github.com/angular/components/issues/2574. For context, in the Material form field we show a small asterisk next to the text if it's required. Usually we can infer it from the `required` attribute, but that doesn't work in reactive forms.

albyrock87:
I agree with @crisbeto on `control.getValidators()` and also some `addValidators(...)`, `removeValidators(...)` would be helpful to the developer to manage validators after the control has been created

These methods can be used by directives to add `NG_VALIDATORS`.

packages/forms/src/model.ts
```diff
@@ -175,11 +199,42 @@ export abstract class AbstractControl {
...
+      asyncValidators: AsyncValidatorFn|AsyncValidatorFn[]|null) {
+    this._rawValidators = validators;
+    this._rawAsyncValidators = asyncValidators;
+  }
+
+  /**
+   * Gets/sets coerced validator function.
+   */
+  get validator(): ValidatorFn|null {
+    if (!this._coercedValidator) {
```
albyrock87, line 217: We should check for `undefined` here to avoid calling `coerceToValidators` every time a control with no validators is being validated.

packages/forms/src/model.ts
```diff
@@ -159,6 +173,16 @@ export abstract class AbstractControl {
   private _parent!: FormGroup|FormArray;
   private _asyncValidationSubscription: any;
 
+  /** @internal */
+  private _coercedValidator!: ValidatorFn|null;
```
albyrock87, line 177:
```suggestion
  private _coercedValidator!: ValidatorFn|null|undefined;
```

packages/forms/src/model.ts
```diff
@@ -361,7 +419,10 @@ export abstract class AbstractControl {
    *
    */
   setAsyncValidators(newValidator: AsyncValidatorFn|AsyncValidatorFn[]|null): void {
-    this.asyncValidator = coerceToAsyncValidator(newValidator);
+    this._rawAsyncValidators = newValidator;
+    // Reset coerced async validator state, so that it can be re-calculated
+    // when needed in a `asyncValidator` getter function.
+    this._coercedAsyncValidator = null;
```
albyrock87, line 425:
```suggestion
    this._coercedAsyncValidator = undefined;
```

packages/forms/src/model.ts
AndrewKushnir: Good catch, I've updated the logic to differentiate between `null` and `undefined` (and also documented it).

AndrewKushnir:
Thanks for the feedback @albyrock87, I've applied the changes based on your comments.

> I agree with @crisbeto on control.getValidators() and also some addValidators(...), removeValidators(...) would be helpful to the developer to manage validators after the control has been created

Yes, I agree. We need to land this change first and that would open up the possibility to implement `getValidators` function (which would be just exposing the `_rawValidators` field value). I believe having `setValidators` and `getValidators` API methods should provide enough logic to manage validators and we can look into `addValidators` function later if needed.

petebacondarwin:
Quick note: there are two typos in the commit message:

* `clean then up` -> `clean them up`
* `that is` -> `that it`

packages/forms/src/model.ts
```diff
@@ -159,6 +173,54 @@ export abstract class AbstractControl {
...
+   *  - `undefined` indicates that the merge didn't happen yet, so once the merged validator
+   * function is requested (by accessing `control.validator`), the `Validators.compose` should
+   * be invoked before returning the result.
+   *  - `null` is set when empty set of validators was provided and the merge operation was
+   * completed. In this case, it's a final value and `Validators.compose` should *not* be called
+   * again.
+   *
+   * @internal
+   */
+  private _coercedValidator: ValidatorFn|null|undefined;
```
petebacondarwin, line 190: What is the benefit in lazy initialising these properties? Why not just initialise them in the constructor and save the extra complexity in the getter functions?

```
  constructor(
      validators: ValidatorFn|ValidatorFn[]|null,
      asyncValidators: AsyncValidatorFn|AsyncValidatorFn[]|null) {
    this._rawValidators = validators;
    this._rawAsyncValidators = asyncValidators;
    this._coercedValidator = coerceToValidator(this._rawValidators);
    this._coercedAsyncValidator = coerceToAsyncValidator(this._rawAsyncValidators);
  }

  ...

  get validator(): ValidatorFn|null {
    return this._coercedValidator;
  }
  set validator(validatorFn: ValidatorFn|null) {
    this._rawValidators = this._coercedValidator = validatorFn;
  }

  ...

  get asyncValidator(): AsyncValidatorFn|null {
    return this._coercedAsyncValidator;
  }
  set asyncValidator(asyncValidatorFn: AsyncValidatorFn|null) {
    this._rawAsyncValidators = this._coercedAsyncValidator = asyncValidatorFn;
  }

  ...

  setValidators(newValidator: ValidatorFn|ValidatorFn[]|null): void {
    this._rawValidators = newValidator;
    this._coercedValidator = coerceToValidator(newValidator);
  }

  ...

  setAsyncValidators(newValidator: AsyncValidatorFn|AsyncValidatorFn[]|null): void {
    this._rawAsyncValidators = newValidator;
    this._coercedAsyncValidator = coerceToAsyncValidator(newValidator);
  }
```

packages/forms/src/model.ts
```diff
@@ -159,6 +173,54 @@ export abstract class AbstractControl {
...
+
+  /**
+   * Synchronous validators as they were provided:
+   *  - in `AbstractControl` constructor
+   *  - as an argument while calling `setValidators` function
+   *  - while calling the setter on the `validator` field (e.g. `control.validator = validatorFn`)
+   *
+   * @internal
+   */
+  private _rawValidators!: ValidatorFn|ValidatorFn[]|null;
```
petebacondarwin, line 211: The `!` operator is needed here (nor on `_rawAsyncValidators`) since they are initialised in the constructor.

packages/forms/src/model.ts
```diff
@@ -69,18 +69,32 @@ function _find(control: AbstractControl, path: Array<string|number>|string, deli
   return controlToFind;
 }
 
+function pickValidators(validatorOrOpts?: ValidatorFn|ValidatorFn[]|AbstractControlOptions|
+                        null): ValidatorFn|ValidatorFn[]|null {
+  const _validator = isOptionsObj(validatorOrOpts) ? validatorOrOpts.validators : validatorOrOpts;
+  return _validator || null;
```
petebacondarwin, line 75: NIT: I don't know if the minifier would remove it for us but the extra `_validator` variable seems unnecessary:
Same for `pickAsyncValidators()`.

```suggestion
  return (isOptionsObj(validatorOrOpts) ? validatorOrOpts.validators : validatorOrOpts) || null;
```

petebacondarwin (changes requested):

AndrewKushnir:
Thanks for the review @petebacondarwin 👍
I really like your proposal and updated the code accordingly (+ fixed typos in commit message).
Could you please have another look when you get a chance?

petebacondarwin (approved):
Great! LGTM

petebacondarwin (approved):
Reviewed-for: public-api

AndrewKushnir:
[Presubmit #2](http://test/OCL:321957292:BASE:321957295:1595097577409:8cbe3143)

alxhub (approved):
Reviewed-For: public-api

packages/forms/src/model.ts
```diff
@@ -69,18 +69,29 @@ function _find(control: AbstractControl, path: Array<string|number>|string, deli
   return controlToFind;
 }
 
+function pickValidators(validatorOrOpts?: ValidatorFn|ValidatorFn[]|AbstractControlOptions|
```
jelbourn, line 72: I would add a JsDoc description 
```ts
/** Gets validators from either an options object or given validators. */
```
(similar for async validators)

packages/forms/src/model.ts
```diff
@@ -159,6 +170,43 @@ export abstract class AbstractControl {
   private _parent!: FormGroup|FormArray;
   private _asyncValidationSubscription: any;
 
+  /**
+   * Contains the result of merging synchronous validators into a single validator function
+   * (combined using `Validators.compose`).
+   *
+   * @internal
+   */
+  private _coercedValidator: ValidatorFn|null;
```
jelbourn, line 179: I know the term was already in this file, but "coerced" seems odd to me compared to something like `_composedValidator`

jelbourn (approved):
LGTM

Reviewed-for: public-api

AndrewKushnir:
[Presubmit #3](http://test/OCL:322246941:BASE:322246981:1595286733651:d8b3a42d) (after additional updates/cleanup).

petebacondarwin (approved):
SLGTM :-)

AndrewKushnir:
FYI, [Global Presubmit](https://test.corp.google.com/ui#id=OCL:322246941:BASE:322296399:1595309922471:71efce3e) went well.

PR #37889, opened 2020-07-02
feat(dev-infra): merge script should link to original commit when cherry-picking with API strategy

devversion:
The merge script uses `git cherry-pick` for both the API merge strategy
and the autosquash strategy. It uses cherry-pick to push commits to
different target branches (e.g. into the `10.0.x` branch).

Those commits never point to the commits that landed in the primary
Github branch though. For the autosquash strategy the pull request number
is always included, so there is a way to go back to the source. On the other
hand though, for commits cherry-picked in the API merge strategy, the
pull request number might not always be included (due to Github's
implementation of the rebase merge method).

e.g. https://github.com/angular/components/commit/27f52711c0618a9ae4eab4c888c8ab3245638e77

For those cases we'd want to link the cherry-picked commits to the
original commits so that the corresponding PR is easier to track
down. This is not needed for the autosquash strategy (as outlined
before), but it would have been good for consistency. Unfortunately
though this would rather complicate the strategy as the autosquash
strategy cherry-picks directly from the PR head, so the SHAs that
are used in the primary branch are not known.

### DIFF
--- a/dev-infra/pr/merge/strategies/api-merge.ts
+++ b/dev-infra/pr/merge/strategies/api-merge.ts
@@ -135,7 +135,13 @@ export class GithubApiMergeStrategy extends MergeStrategy {
 
     // Cherry pick the merged commits into the remaining target branches.
     const failedBranches = await this.cherryPickIntoTargetBranches(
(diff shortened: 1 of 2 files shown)


josephperrott (approved):
LGTM

PR #37891, opened 2020-07-02
fix(dev-infra): include `bazel` utility files in npm package

devversion:
We recently added a new folder for common bazel utilties
to `dev-infra`. The `ng_rollup_bundle` rule relies on an
utility that is provided by this `bazel/` folder.

Unfortunately though it looks like this folder is currently
not included in the NPM package, so that the `ng_rollup_bundle`
rule does not work as expected. This commit fixes that by
including the bazel utilties in the NPM package.

### DIFF
--- a/dev-infra/BUILD.bazel
+++ b/dev-infra/BUILD.bazel
@@ -39,6 +39,7 @@ pkg_npm(
     name = "npm_package",
     srcs = [
         "BUILD.bazel",
+        "//dev-infra/bazel:files",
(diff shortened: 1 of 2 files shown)


josephperrott (approved):
LGTM

wagnermaciel (approved):

atscott:
@devversion somehow this conflicts with the patch branch. Can you advise how to move forward here? My understanding is that this shouldn't happen for dev-infra changes so maybe something has gone wrong here.

PR #37894, opened 2020-07-02
fix(common): add body to delete request

### LINKED ISSUE #19438: HttpClient.delete() cannot handle a body in its request
## I'm submitting a...
<pre><code>
[ ] Bug report  
[ x ] Feature request
[ x ] Documentation issue or request
</code></pre>

## Current behavior
The [`HttpClient.delete()`](https://angular.io/api/common/http/HttpClient#delete) does not accept a body parameter. 

## Expected behavior
The [`HttpClient.delete()`](https://angular.io/api/common/http/HttpClient#delete) needs to accept a body or clear guidance on why it has this limitation and how to get around it. It was possible with the original [`Http.delete()`](https://angular.io/api/http/Http#delete) via the [`RequestOptionsArgs`](https://angular.io/api/http/RequestOptionsArgs#body)

I work with APIs that have always handled the body of DELETE requests and this is an unexpected change in REST api interaction.

## Environment
<pre><code>
Angular version: 4.3.2

Browser:
- [ x ] Chrome (desktop) version 61.0.3163.79
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX

Others:

</code></pre>

ajitsinghkaler:
delete request supports a body but it was left out as many dont support the http body. As it is supported by default so angular should not take any decision weather to add it or not the decision should be given to the developer

Fixes #19438

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Delete request cannot send a body

Issue Number: #19438

## What is the new behavior?
Delete request can send a body

- [x] Yes
- [ ] No

### DIFF
--- a/goldens/public-api/common/http/http.d.ts
+++ b/goldens/public-api/common/http/http.d.ts
@@ -6,7 +6,7 @@ export declare abstract class HttpBackend implements HttpHandler {
 
 export declare class HttpClient {
     constructor(handler: HttpHandler);
-    delete(url: string, options: {
+    delete(url: string, body: any | null, options: {
         headers?: HttpHeaders | {
             [header: string]: string | string[];
         };
@@ -18,7 +18,7 @@ export declare class HttpClient {
         responseType: 'arraybuffer';
         withCredentials?: boolean;
     }): Observable<ArrayBuffer>;
-    delete(url: string, options: {
+    delete(url: string, body: any | null, options: {
         headers?: HttpHeaders | {
             [header: string]: string | string[];
         };
@@ -30,7 +30,7 @@ export declare class HttpClient {
         responseType: 'blob';
         withCredentials?: boolean;
     }): Observable<Blob>;
-    delete(url: string, options: {
+    delete(url: string, body: any | null, options: {
         headers?: HttpHeaders | {
             [header: string]: string | string[];
         };
@@ -42,7 +42,7 @@ export declare class HttpClient {
         responseType: 'text';
         withCredentials?: boolean;
     }): Observable<string>;
-    delete(url: string, options: {
+    delete(url: string, body: any | null, options: {
         headers?: HttpHeaders | {
             [header: string]: string | string[];
         };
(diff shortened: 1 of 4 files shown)


jelbourn:
This change, as is, looks like it's a pretty big breaking change. I'm wondering if we can make it a non-breaking change by adding more method overloads, and then having the concrete implementation handle the difference in behavior for two vs. three arguments. Otherwise, it would take a large effort to land this change, which probably isn't the most valuable use of resources (compared to other higher priority issues/features).

ajitsinghkaler:
@jelbourn but overloads will create confusion that sometimes the second parameter of the function is the options and sometimes the third.

We can create the body as the third parameter in that case but the main problem is that it will not conform with the style of post and put request.

jelbourn:
Yeah, I don't see adding the body as a third argument as a viable option. Unfortunately, just making this breaking change is also not a viable option. Changes like this are genuinely painful for both our customers and the team ourselves, because we're responsible for updating all usage inside Google (which has to happen as part of landing the change because [all of Google is one massive repo where everything is at head all the time](https://cacm.acm.org/magazines/2016/7/204032-why-google-stores-billions-of-lines-of-code-in-a-single-repository/fulltext)).

ajitsinghkaler:
@jelbourn so what do you think we should do it's a highly requested feature 66 likes or should I drop this pr

petebacondarwin (changes requested):
I don't think we should land this PR. The large breaking change or inconsistent API is not offset by the minimal benefit to developers.

There is a perfectly reasonable workaround, which is to use the `client.request()` method: https://github.com/angular/angular/issues/19438#issuecomment-346596988

jelbourn:
If we could make it work in an ergonomic way with overloads it could be doable, but I don't see a practical way to move forward on this in any near timeframe.

Issue #37901, opened 2020-07-03
xi18n configuration stopped working after upgrading to Angular 10


SanderBreivik:
# 🐞 bug report

### Affected Package

 The issue is caused by package @angular/localize

### Is this a regression?

 Yes, the previous version in which this bug was not present was: 9.x

### Description

 I upgraded from Angular 9.x to 10.x using ng update @angular/core @angular/cli. After doing some more programming and adding more translation to my solution, I then tried to run my english configuration using 'ng serve --configuration=en' and I got the error shown later in this issue. This configuration has worked flawlessly before upgrading to 10.0.2.

## 🔬 Minimal Reproduction

Hard to reproduce since I cannot extract i18n from stackblitz, nor run a different configuration
 https://stackblitz.com/edit/angular-ivy-ju9lvy

## 🔥 Exception or Error
<pre><code>

[error] Error: Unsupported translation file format.
    at C:\Developement\Microservices\Statistics\app\node_modules\@angular-devkit\build-angular\src\utils\load-translations.js:24:15
    at Object.configureI18nBuild (C:\Developement\Microservices\Statistics\app\node_modules\@angular-devkit\build-angular\src\utils\i18n-options.js:163:28)
    at async Object.generateI18nBrowserWebpackConfigFromContext (C:\Developement\Microservices\Statistics\app\node_modules\@angular-devkit\build-angular\src\utils\webpack-browser-config.js:86:36)
    at async setup (C:\Developement\Microservices\Statistics\app\node_modules\@angular-devkit\build-angular\src\dev-server\index.js:76:47)

</code> </pre>

## 🌍  Your Environment

**Angular Version:**
<pre><code>

     _                      _                 ____ _     ___
    / \   _ __   __ _ _   _| | __ _ _ __     / ___| |   |_ _|
   / △ \ | '_ \ / _` | | | | |/ _` | '__|   | |   | |    | |
  / ___ \| | | | (_| | |_| | | (_| | |      | |___| |___ | |
 /_/   \_\_| |_|\__, |\__,_|_|\__,_|_|       \____|_____|___|
                |___/
    

Angular CLI: 10.0.1
Node: 12.18.1
OS: win32 x64

Angular: 10.0.2
... animations, common, compiler, compiler-cli, core, forms
... localize, platform-browser, platform-browser-dynamic, router
Ivy Workspace: Yes

Package                           Version
-----------------------------------------------------------
@angular-devkit/architect         0.1000.1
[...]

PR #37909, opened 2020-07-03
feat(localize): expose `canParse()` diagnostics

petebacondarwin:
When loading a translation file we ask each `TranslationParser`
whether it can parse the file. Occasionally, this check can find
errors in the file that would be useful to the developer. For example
if the file has invalid XML.

This commit allows the caller to pass a diagnostics object to the
`canParse()` call which will be filled with any messages that can
be used to diagnose problems with the format of the file.

Closes #37901

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/localize/src/tools/src/translate/translation_files/translation_loader.ts
+++ b/packages/localize/src/tools/src/translate/translation_files/translation_loader.ts
@@ -57,8 +57,9 @@ export class TranslationLoader {
   private loadBundle(filePath: AbsoluteFsPath, providedLocale: string|undefined):
       TranslationBundle {
     const fileContents = this.fs.readFile(filePath);
+    const canParseDiagnostics = new Diagnostics();
     for (const translationParser of this.translationParsers) {
-      const result = translationParser.canParse(filePath, fileContents);
+      const result = translationParser.canParse(filePath, fileContents, canParseDiagnostics);
       if (!result) {
         continue;
       }
@@ -90,8 +91,9 @@ export class TranslationLoader {
 
       return {locale, translations, diagnostics};
     }
-    throw new Error(
-        `There is no "TranslationParser" that can parse this translation file: ${filePath}.`);
+
+    throw new Error(canParseDiagnostics.formatDiagnostics(
+        `There is no "TranslationParser" that can parse this translation file: ${filePath}.`));
   }
 
   /**
--- a/packages/localize/src/tools/src/translate/translation_files/translation_loader.ts
+++ b/packages/localize/src/tools/src/translate/translation_files/translation_loader.ts
@@ -57,11 +57,11 @@ export class TranslationLoader {
   private loadBundle(filePath: AbsoluteFsPath, providedLocale: string|undefined):
       TranslationBundle {
     const fileContents = this.fs.readFile(filePath);
-    const canParseResults = new Map<TranslationParser<any>, ParseAnalysis<any>>();
+    const unusedParsers = new Map<TranslationParser<any>, ParseAnalysis<any>>();
     for (const translationParser of this.translationParsers) {
       const result = translationParser.analyze(filePath, fileContents);
-      canParseResults.set(translationParser, result);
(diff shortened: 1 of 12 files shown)


packages/localize/src/tools/src/translate/translation_files/translation_parsers/translation_utils.ts
```diff
@@ -80,25 +80,38 @@ export interface XmlTranslationParserHint {
...
+      xml.errors.forEach(e => addParseError(diagnostics, e));
+    }
     return false;
   }
 
   const rootElements = xml.rootNodes.filter(isNamedElement(rootNodeName));
   const rootElement = rootElements[0];
   if (rootElement === undefined) {
+    if (diagnostics !== undefined) {
+      diagnostics.warn(`The XML file does not contain a <${rootNodeName}> root node.`);
```
AndrewKushnir, line 100: Can we call `addParseDiagnostic` with `ParseErrorLevel.WARNING` level?

AndrewKushnir:
Hi @petebacondarwin, thanks for the PR 👍 

I was thinking is we can rename `canParse` method to something like `analyze` and return an object that would contain information extracted during this initial phase, for ex.:

```
{
  canParse: boolean,
  parsedContent: any,
  diagnostics: Diagnostics
}
```

In this case we don't need to supply the `Diagnostics` instance into the `canParse` function call, which looks like a way to extend the `canParse` function to return more than just a boolean. What do you think Pete?

Thank you.

petebacondarwin:
Thanks for the "weekend!" review @AndrewKushnir. The problem with have with `canParse` is that it is effectively public API because it is used by the CLI directly. So I have been trying to avoid breaking the signature that they use.
We could agree with them to change to the new API that you suggest, which I like.

AndrewKushnir:
Thanks for the reply @petebacondarwin! We can probably keep the `canParse` function for now (as is), mark is as deprecated and introduce a new `analyze` function that would return more info. The `canParse` can call `analyze` internally, so having both functions available for some period of time would allow avoiding a breaking change (and provide a transition period).

petebacondarwin:
Yes. That sounds like a good plan.  I'll do that today.  Thanks

packages/localize/src/tools/src/translate/translation_files/translation_parsers/translation_utils.ts
petebacondarwin: `addParseDiagnostic()` requires a `sourceSpan` but I don't think there is one available in this case, since there were no root elements found.

petebacondarwin:
Updated

packages/localize/src/tools/src/translate/translation_files/translation_loader.ts
```diff
@@ -57,14 +57,16 @@ export class TranslationLoader {
   private loadBundle(filePath: AbsoluteFsPath, providedLocale: string|undefined):
       TranslationBundle {
     const fileContents = this.fs.readFile(filePath);
+    const canParseResults = new Map<TranslationParser<any>, ParseAnalysis<any>>();
```
AndrewKushnir, line 60: nit: may be rename this const to mention that it's a result of the `analyze` operation, not the `canParse` call?

AndrewKushnir (approved):
LGTM (just one nit) 👍 Thanks for additional changes Pete!

packages/localize/src/tools/src/translate/translation_files/translation_loader.ts
petebacondarwin: I propose `unusedParsers`, since it will be a map of parsers that cannot parse the file.

AndrewKushnir: Thanks Pete 👍

PR #37911, opened 2020-07-03
fix(compiler): evaluate safe navigation expressions in correct bindin…

### LINKED ISSUE #37194: angular 9 ivy input, OnChanges bug
# 🐞 bug report

### Affected Package
may be @angular/core

### Is this a regression?

angular 8 woking ok but angular 9 problem

#### expected  
only one input value change
#### issue
but other input value change

## 🔬 Minimal Reproduction

https://stackblitz.com/edit/angular-ivy-input-change-issue

## 🌍  Your Environment

stackbliz  dependencies

JoostK:
…g order

When using the safe navigation operator in a binding expression, a temporary
variable may be used for storing the result of a side-effectful call.
For example, the following template uses a pipe and a safe property access:

```html
<app-person-view [enabled]="enabled" [firstName]="(person$ | async)?.name"></app-person-view>
```

The result of the pipe evaluation is stored in a temporary to be able to check
whether it is present. The temporary variable needs to be declared in a separate
statement and this would also cause the full expression itself to be pulled out
into a separate statement. This would compile into the following
pseudo-code instructions:

```js
var temp = null;
var firstName = (temp = pipe('async', ctx.person$)) == null ? null : temp.name;
property('enabled', ctx.enabled)('firstName', firstName);
```

Notice that the pipe evaluation happens before evaluating the `enabled` binding,
such that the runtime's internal binding index would correspond with `enabled`,
not `firstName`. This introduces a problem when the pipe uses `WrappedValue` to
force a change to be detected, as the runtime would then mark the binding slot
corresponding with `enabled` as dirty, instead of `firstName`. This results
in the `enabled` binding to be updated, triggering setters and affecting how
`OnChanges` is called.

In the pseudo-code above, the intermediate `firstName` variable is not strictly
necessary---it only improved readability a bit---and emitting it inline with
the binding itself avoids the out-of-order execution of the pipe:

```js
var temp = null;
property('enabled', ctx.enabled)
  ('firstName', (temp = pipe('async', ctx.person$)) == null ? null : temp.name);
```

This commit introduces a new `BindingForm` that results in the above code to be
generated and adds compiler and acceptance tests to verify the proper behavior.

Fixes #37194

### DIFF
--- a/packages/core/test/acceptance/pipe_spec.ts
+++ b/packages/core/test/acceptance/pipe_spec.ts
@@ -6,9 +6,10 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {ChangeDetectionStrategy, ChangeDetectorRef, Component, Directive, Inject, Injectable, InjectionToken, Input, NgModule, OnDestroy, Pipe, PipeTransform, ViewChild} from '@angular/core';
+import {ChangeDetectionStrategy, ChangeDetectorRef, Component, Directive, Inject, Injectable, InjectionToken, Input, NgModule, OnChanges, OnDestroy, Pipe, PipeTransform, SimpleChanges, ViewChild, WrappedValue} from '@angular/core';
 import {TestBed} from '@angular/core/testing';
 import {expect} from '@angular/platform-browser/testing/src/matchers';
+import {ivyEnabled} from '@angular/private/testing';
 
 describe('pipe', () => {
   @Pipe({name: 'countingPipe'})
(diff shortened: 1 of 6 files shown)


pkozlowski-opensource (commented):
LGTM, overall, but as discussed offline I would like to see one more test where a component uses `ngOnChanges` as reported in #37194. Also the CI size check is failing so it will need cleanup (hence the appropriate label).

Thnx so much for this PR, must have been fun to track down :-) Oh, and the PR description / commit message is absolutely excellent 👍 


Reviewed-for: fw-core

alxhub:
[Presubmit](http://test/OCL:324706798:BASE:324706850:1596497026341:2646bda4)

packages/core/test/acceptance/pipe_spec.ts
```diff
@@ -285,6 +286,133 @@ describe('pipe', () => {
...
+        'pipe',
+        'set value2=3',
+      ]);
+    });
+
+    it('should not include unrelated inputs in ngOnChanges', () => {
+      // https://github.com/angular/angular/issues/37194
+      // https://github.com/angular/angular/issues/37591
+      // Using a pipe in the LHS of safe navigation operators would clobber unrelated bindings
+      // iff the pipe returns WrappedValue, incorrectly causing the unrelated binding
```
pkozlowski-opensource, line 358: "iff" typo?

pkozlowski-opensource (approved):
LGTM, thnx for adding all the runtime tests!

packages/core/test/acceptance/pipe_spec.ts
JoostK: it's a deliberate ["if-and-only-if"](https://en.wikipedia.org/wiki/If_and_only_if)

PR #37912, opened 2020-07-03
fix(compiler-cli): avoid creating value expressions for symbols from type-only imports

### LINKED ISSUE #37900: Injector does not respect TypeScript type-only imports
# 🐞 bug report

### Description

TypeScript introduced [type-only imports in version 3.8](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-exports) which contains the following text in the documentation:

> `import type` only imports declarations to be used for type annotations and declarations. **It always gets fully erased, so there’s no remnant of it at runtime**.

Now consider a new Angular CLI project with the following relevant files:

**foo.ts**:
```ts
console.log('I do not expect this statement to print if the file is imported as type-only.')

export class Foo { }
```

**foo.token.ts**:
```ts
import { InjectionToken } from '@angular/core';

export const FOO = new InjectionToken('Foo', {
    factory: () => { };
});
```

**app.component.ts**:
```ts
import { Component, Inject } from '@angular/core';
import { FOO } from './foo.token';
import type { Foo } from './foo';

@Component({ ... })
export class AppComponent {


  constructor(@Inject(FOO) foo: Foo) {}
}
```
The key things to note is that the `Foo` class is being imported as a type only. It is also not being imported anywhere else in the project. As the `console.log` message in the `foo.ts` file states, I do not expect that line to print at run time. I do not expect that line to print because TypeScript says that type-only imports "always gets fully erased, so there’s no remnant of it at runtime". The actual behavior does not match my expected behavior because the `console.log` message is being printed at runtime.

## 🔬 Minimal Reproduction
[StackBlitz](https://stackblitz.com/edit/angular-ivy-54e8de)
or
[GitHub repo](https://github.com/samherrmann/angular-sandbox/commit/9c77f6a57b2c7d2bc036700d0df278232ef0a314)

Observe the console.

</code></pre>

## 🌍  Your Environment

**Angular Version:**
<pre><code>

Angular CLI: 10.0.0
Node: 12.15.0
OS: linux x64

Angular: 10.0.2
... router
Ivy Workspace: Yes

@angular-devkit/architect         0.1000.0
@angular-devkit/build-angular     0.1000.0
@angular-devkit/build-optimizer   0.1000.0
@angular-devkit/build-webpack     0.1000.0
@angular-devkit/core              10.0.0
@angular-devkit/schematics        10.0.0
@angular/cli                      10.0.0
@ngtools/webpack                  10.0.0
@schematics/angular               10.0.0
@schematics/update                0.1000.0
typescript                        3.9.5
webpack                           4.43.0

</code></pre>

JoostK:
In TypeScript 3.8 support was added for type-only imports, which only brings in
the symbol as a type, not their value. The Angular compiler did not yet take
the type-only keyword into account when representing symbols in type positions
as value expressions. The class metadata that the compiler emits would include
the value expression for its parameter types, generating actual imports as
necessary. For type-only imports this should not be done, as it introduces an
actual import of the module that was originally just a type-only import.

This commit lets the compiler deal with type-only imports specially, preventing
a value expression from being created.

Fixes #37900

---

image

### DIFF
--- a/packages/compiler-cli/test/ngtsc/ngtsc_spec.ts
+++ b/packages/compiler-cli/test/ngtsc/ngtsc_spec.ts
@@ -4202,6 +4202,43 @@ runInEachFileSystem(os => {
          expect(jsContents).toMatch(setClassMetadataRegExp('type: undefined'));
        });
 
+    it('should use `undefined` in setClassMetadata if types originate from type-only imports',
+       () => {
+         env.write(`types.ts`, `
+           export default class {}
+           export class TypeOnly {}
+         `);
+         env.write(`test.ts`, `
+           import {Component, Inject, Injectable} from '@angular/core';
+           import type DefaultImport from './types';
+           import type {TypeOnly} from './types';
+           import type * as types from './types';
+
+           @Component({
+             selector: 'some-comp',
+             template: '...',
+           })
+           export class SomeComp {
+             constructor(
+               @Inject('token') namedImport: TypeOnly,
+               @Inject('token') defaultImport: DefaultImport,
+               @Inject('token') namespacedImport: types.TypeOnly,
+             ) {}
+           }
+        `);
+
+         env.driveMain();
+         const jsContents = trim(env.getContents('test.js'));
+         // Module specifier for type-only import should not be emitted
+         expect(jsContents).not.toContain('./types');
+         // Default type-only import should not be emitted
+         expect(jsContents).not.toContain('DefaultImport');
+         // Named type-only import should not be emitted
+         expect(jsContents).not.toContain('TypeOnly');
+         // The parameter type in class metadata should be undefined
+         expect(jsContents).toMatch(setClassMetadataRegExp('type: undefined'));
+       });
+
     it('should not throw in case whitespaces and HTML comments are present inside <ng-content>',
        () => {
          env.write('test.ts', `
--- a/packages/compiler-cli/test/ngtsc/ngtsc_spec.ts
+++ b/packages/compiler-cli/test/ngtsc/ngtsc_spec.ts
@@ -2079,7 +2079,13 @@ runInEachFileSystem(os => {
 
              const errors = env.driveDiagnostics();
              expect(errors.length).toBe(1);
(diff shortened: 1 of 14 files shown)


packages/compiler-cli/test/ngtsc/ngtsc_spec.ts
```diff
@@ -4238,6 +4238,43 @@ runInEachFileSystem(os => {
          expect(jsContents).toMatch(setClassMetadataRegExp('type: undefined'));
        });
 
+    it('should use `undefined` in setClassMetadata if types originate from typ-only imports',
```
samherrmann, line 4241: `typ-only` => `type-only`

JoostK: This is my signature typo, honestly :-)

samherrmann: well don't feel bad, I mean you spelled it correctly 7/8 times in this PR 😂

alxhub:
I agree with the spirit of this PR :D 

I think we need to introduce an error, though, where the user does:

```typescript
import type {Foo} from './foo';

@Component({...})
export class SomeCmp {
  constructor(private foo: Foo) {...}
}
```

That is, they use the type-only symbol in a DI position _without_ providing an `@Inject` value token. In this case, I think we should offer a diagnostic highlighting the type symbol and its related import, and explaining that DI requires the symbol to be a valid value.

JoostK:
I like that idea, updated the PR accordingly. See screenshot in issue description for a preview, looks rather nice 😃 This includes suggestive messages for all other situations where a type is unsuitable as injection token.

alxhub:
[Presubmit](http://test/OCL:324707124:BASE:324707148:1596497122846:7f426e9d)

alxhub (approved):
👍 this looks awesome!

AndrewKushnir:
Hi @JoostK, thanks for the fix!

Merging this PR to the patch branch failed due to some conflicts, so I merged it to master only for now. Could you please create a version of this PR for the patch branch?

PR #37918, opened 2020-07-03
feat(compiler): support unary operators for more accurate type checking

### LINKED ISSUE #20845: Using a negative literal produces (0 - <literal>) which might type check
## I'm submitting a...

<pre><code>
[x] Bug report  
</code></pre>

## Current behavior

A template with `-1` is emitted as `0 - 1` which is equivalent at runtime (and is converted back to -1 by uglify) but it does match a type such as `1 | -1`.

## Expected behavior

It should emit -1.

Pending.

Correct type checking.

### LINKED ISSUE #36178: fullTemplateCheck not recognizing unary (+) operator in template
# 🐞 bug report

### Description

I'm finally getting around to using ```angularCompilerOptions``` and have started to use ```fullTemplateCheck```. It seems as if while checking the templates, there is a failure to see the ```+``` unary operator in front of the ```hours``` variable (```hours``` is a string). The ```+``` turns the string into a number so I believe this is an incorrect error, and therefore a bug?

## 🔥 Exception or Error
<pre>
error TS2362: The left-hand side of an arithmetic operation must be of type 'any', 'number', 'bigint' or an enum type.
&ltdiv (click)="formatHours(<b>+hours</b> + 1)"&gt
</pre>

## Environment

**Angular Version:**
<pre><code>
Angular CLI: 9.0.4
Node: 10.16.1
OS: linux x64

Angular: 9.0.4
... router
Ivy Workspace: Yes

@angular-devkit/architect         0.900.4
@angular-devkit/build-angular     0.900.4
@angular-devkit/build-optimizer   0.900.4
@angular-devkit/build-webpack     0.900.4
@angular-devkit/core              9.0.4
@angular-devkit/schematics        9.0.4
@angular/cdk                      9.1.0
@angular/material                 9.1.0
@ngtools/webpack                  9.0.4
@schematics/angular               9.0.4
@schematics/update                0.900.4
</code></pre>

JoostK:
Prior to this change, the unary + and - operators would be parsed as `x - 0`
and `0 - x` respectively. The runtime semantics of these expressions are
equivalent, however they may introduce inaccurate template type checking
errors as the literal type is lost, for example:

```ts
@Component({
  template: `<button [disabled]="isAdjacent(-1)"></button>`
})
export class Example {
  isAdjacent(direction: -1 | 1): boolean { return false; }
}
```

would incorrectly report a type-check error:

> error TS2345: Argument of type 'number' is not assignable to parameter
  of type '-1 | 1'.

Additionally, the translated expression for the unary + operator would be
considered as arithmetic expression with an incompatible left-hand side:

> error TS2362: The left-hand side of an arithmetic operation must be of
  type 'any', 'number', 'bigint' or an enum type.

To resolve this issues, the implicit transformation should be avoided.
This commit adds a new unary AST node to represent these expressions,
allowing for more accurate type-checking.

Fixes #20845
Fixes #36178

### DIFF
--- a/packages/compiler/src/expression_parser/ast.ts
+++ b/packages/compiler/src/expression_parser/ast.ts
@@ -216,17 +216,6 @@ export class Interpolation extends AST {
   }
 }
 
-export class Unary extends AST {
-  constructor(
-      span: ParseSpan, sourceSpan: AbsoluteSourceSpan, public operation: string,
-      public right: AST) {
-    super(span, sourceSpan);
-  }
-  visit(visitor: AstVisitor, context: any = null): any {
-    return visitor.visitUnary(this, context);
-  }
-}
-
 export class Binary extends AST {
   constructor(
       span: ParseSpan, sourceSpan: AbsoluteSourceSpan, public operation: string, public left: AST,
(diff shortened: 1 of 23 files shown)


packages/compiler/src/expression_parser/ast.ts
```diff
@@ -361,6 +372,7 @@ export interface TemplateBindingIdentifier {
 }
 
 export interface AstVisitor {
+  visitUnary(ast: Unary, context: any): any;
```
JoostK, line 375: I'm curious how breaking this will be, I can imagine e.g. Codelyzer choking on this new method.

alxhub:
[Presubmit](http://test/OCL:324709838:BASE:324709871:1596497929929:aa68ac4)

alxhub (approved):

alxhub:
Funnily enough, codelyzer is fine with this one 👍

kyliau (approved):

kyliau:
presubmit: 	http://test/OCL:327567923:BASE:327567938:1597899560264:b7598ce2

PR #37566, opened 2020-06-12
fix(forms): clean up ControlValueAccessor connection when FormControlDirective is destroyed

AndrewKushnir:
NOTE: this PR is still in the "work in progress" state.

Prior to this commit, removing `FormControlDirective` instance didn't clear the callbacks previously registered on FromControl instance. As a result, these callbacks were executed even after `FormControlDirective` instance was destroyed.

This PR was inspired by #29335.

## PR Type
What kind of change does this PR introduce?

- [x] Bugfix

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

PR #37924, opened 2020-07-04
fix(forms): FormControlName directive not cleaning up on destroy

### LINKED ISSUE #37431: Using ngIf, DOM nodes cannot be garbage collected because of form control
# 🐞 bug report

### Is this a regression?
No

### Description
Not sure if I am using this in the wrong way. I get a leak when I switch a container (using ngIf) which has reference to form control inside it.

I provided a stackblitz for reproducing the issue.

I also provided a stackblitz where I remove the control from the form group when I toggle the container. That resolves the issue but is that what I am expected to do or how do I go about this?

## 🔬 Minimal Reproduction
To reproduce:
https://stackblitz.com/edit/angular-9-ngif-leak

Example where I remove the form control on every toggle:
https://stackblitz.com/edit/angular-9-ngif-remove-control

## 🔥 Exception or Error
The full container with children will be left in detached mode on each switch. I think the DefaultValueAccessor is the retainer.

## 🌍  Your Environment

**Angular Version:**
9.0.1

crisbeto:
Fixes a memory leak caused by the `FormControlName` directive which wasn't cleaning up after itself on destroy. It ended up retaining a reference to the detached DOM node that the directive is applied to.

Fixes #37431.

### DIFF
--- a/packages/forms/src/directives/reactive_directives/form_control_name.ts
+++ b/packages/forms/src/directives/reactive_directives/form_control_name.ts
@@ -15,7 +15,7 @@ import {ControlContainer} from '../control_container';
 import {ControlValueAccessor, NG_VALUE_ACCESSOR} from '../control_value_accessor';
 import {NgControl} from '../ng_control';
 import {ReactiveErrors} from '../reactive_errors';
-import {_ngModelWarning, composeAsyncValidators, composeValidators, controlPath, isPropertyUpdated, selectValueAccessor} from '../shared';
+import {_ngModelWarning, cleanUpControl, composeAsyncValidators, composeValidators, controlPath, isPropertyUpdated, selectValueAccessor} from '../shared';
 import {AsyncValidator, AsyncValidatorFn, Validator, ValidatorFn} from '../validators';
 
 import {NG_MODEL_WITH_FORM_CONTROL_WARNING} from './form_control_directive';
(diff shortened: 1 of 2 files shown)


AndrewKushnir:
Hi @crisbeto, thanks for creating this PR.

I was working on that problem as a part of #37566 (it's in a Draft mode, since I was checking how things work with Google apps). [This part of the fix](https://github.com/angular/angular/pull/37566/files#diff-1cd69b47f0a5d7294368b9d435c57c12R161) from #37566 is very similar to what you implemented (which confirms the overall direction), but the idea that I had was to keep the `setUpControl` and `cleanUpControl` (which performs reverse operation from `setUpControl`) function calls on the same logical level (i.e. within the same class). Also it turned out that the  `cleanUpControl` function itself is not working correctly in all cases (see additional changes in #37566), for example it doesn't clear validators attached to the FormControl instances - I plan to fix that too after landing #37881 (which should allow more granular operations with the list of validators). The delta between our fixes is that you also clear the `control` field and I should look into that, so let's keep this PR open and I will share more info after additional investigation.

Thank you.

crisbeto:
Closing in favor of https://github.com/angular/angular/pull/37566.

rppig42:
Hi there, would this patch be merged into Angular@9.x?

And thank you for your work

crisbeto:
It might be better to ask on #37566 since this PR didn't end up getting merged.

PR #37938, opened 2020-07-06
fix(docs-infra): prevent search-bar from overlapping nav-items

### LINKED ISSUE #37667: > [angular/aio/src/polyfills.ts](https://github.com/angular/angular/blob/0822dc70f2580a3d0e5c936ef07707dd09e7a878/aio/src/polyfills.ts#L38)
> [angular/aio/src/polyfills.ts](https://github.com/angular/angular/blob/0822dc70f2580a3d0e5c936ef07707dd09e7a878/aio/src/polyfills.ts#L38)
> 
> Line 38 in [0822dc7](/angular/angular/commit/0822dc70f2580a3d0e5c936ef07707dd09e7a878)
> 
>  import 'zone.js/dist/zone';  // Included with Angular CLI. 
> Getting this error: ERROR in ./src/polyfills.ts
> Module not found: Error: Can't resolve 'zone.js/dist/zone' in 'C:\Projects\myProject\src'

I have the same error: 

Try this.
> **"npm i zone.js"**
above command worked in my case.

this is also worked for me

_Originally posted by @assadshehbaz in https://github.com/angular/angular/issues/25288#issuecomment-468644655_

gkalpak:
As part of angular.io's responsive layout, the menu shown in the top-bar is collapsed into the sidenav on narrow screens at the point where the search-bar (on the right side of the top-bar) would overlap with the menu's nav-items.

Previously, the value used as break-point would work on marketing pages, where the hamburger button is not shown on wide screens. However, on docs pages (where the hamburger button is always shown, pushing the menu further to the right), the search-bar would still overlap the menu nav-items on some resolutions.

This commit fixes it by raising the screen width threshold at a value that ensures there is no overlap even on pages where the hamburger button is visible alongside the top-bar menu.

Fixes #37667.

##
This PR also includes some refactorings in preparation of the fix. See individual commits for details.

### DIFF
--- a/aio/src/app/app.component.spec.ts
+++ b/aio/src/app/app.component.spec.ts
@@ -193,8 +193,8 @@ describe('AppComponent', () => {
         fixture.detectChanges();
       };
 
-      describe('when side-by-side (wide)', () => {
-        beforeEach(() => resizeTo(dockSideNavWidth + 1));  // side-by-side
+      describe('when view is wide', () => {
+        beforeEach(() => resizeTo(dockSideNavWidth + 1));  // wide view
 
         it('should open when navigating to a guide page (guide/pipes)', () => {
           navigateTo('guide/pipes');
@@ -240,8 +240,8 @@ describe('AppComponent', () => {
         });
       });
 
-      describe('when NOT side-by-side (narrow)', () => {
-        beforeEach(() => resizeTo(dockSideNavWidth - 1)); // NOT side-by-side
+      describe('when view is narrow', () => {
+        beforeEach(() => resizeTo(dockSideNavWidth - 1)); // narrow view
 
         it('should be closed when navigating to a guide page (guide/pipes)', () => {
           navigateTo('guide/pipes');
@@ -294,7 +294,7 @@ describe('AppComponent', () => {
         });
       });
 
-      describe('when changing side-by-side (narrow --> wide)', () => {
+      describe('when changing from narrow to wide view', () => {
         const sidenavDocs = ['api/a/b/c/d', 'guide/pipes'];
         const nonSidenavDocs = ['features', 'about'];
 
@@ -351,7 +351,7 @@ describe('AppComponent', () => {
         });
       });
 
-      describe('when changing side-by-side (wide --> narrow)', () => {
+      describe('when changing from wide to narrow view', () => {
         const sidenavDocs = ['api/a/b/c/d', 'guide/pipes'];
         const nonSidenavDocs = ['features', 'about'];
 
@@ -384,7 +384,7 @@ describe('AppComponent', () => {
       async function setupSelectorForTesting(mode?: string) {
         createTestingModule('a/b', mode);
         await initializeTest();
-        component.onResize(dockSideNavWidth + 1); // side-by-side
+        component.onResize(dockSideNavWidth + 1); // wide view
         selectElement = fixture.debugElement.query(By.directive(SelectComponent));
         selectComponent = selectElement.componentInstance;
       }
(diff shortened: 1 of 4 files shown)


aio/src/app/app.component.spec.ts
```diff
@@ -186,7 +194,7 @@ describe('AppComponent', () => {
       };
 
       describe('when side-by-side (wide)', () => {
```
petebacondarwin, line 196:
```suggestion
      describe('when view is wide', () => {
```

aio/src/app/app.component.spec.ts
```diff
@@ -186,7 +194,7 @@ describe('AppComponent', () => {
...
-        beforeEach(() => resizeTo(sideBySideBreakPoint + 1));  // side-by-side
+        beforeEach(() => resizeTo(dockSideNavWidth + 1));  // side-by-side
```
petebacondarwin, line 197:
```suggestion
        beforeEach(() => resizeTo(dockSideNavWidth + 1));  // wide view
```

aio/src/app/app.component.spec.ts
```diff
@@ -233,7 +241,7 @@ describe('AppComponent', () => {
       });
 
       describe('when NOT side-by-side (narrow)', () => {
```
petebacondarwin, line 243:
```suggestion
      describe('when view is narrow', () => {
```

aio/src/app/app.component.spec.ts
```diff
@@ -233,7 +241,7 @@ describe('AppComponent', () => {
...
-        beforeEach(() => resizeTo(sideBySideBreakPoint - 1)); // NOT side-by-side
+        beforeEach(() => resizeTo(dockSideNavWidth - 1)); // NOT side-by-side
```
petebacondarwin, line 244:
```suggestion
        beforeEach(() => resizeTo(dockSideNavWidth - 1)); // narrow view
```

aio/src/app/app.component.spec.ts
```diff
@@ -292,51 +300,51 @@ describe('AppComponent', () => {
```
petebacondarwin, line 300: This describe block could be:

```
describe('when changing from narrow to wide view', () => {
```

aio/src/styles/1-layouts/_top-menu.scss
```diff
@@ -1,5 +1,5 @@
 // VARIABLES
-$hamburgerShownMaxWidth: 991px;
+$hamburgerShownMaxWidth: 1047px;
 $hamburgerHiddenMinWidth: $hamburgerShownMaxWidth + 1;
```
petebacondarwin, line 3: Ideally this should have the same name (or as close as possible to the constant in `app.component.ts` (`showTopMenuWidth`) and I would argue that it would be better for the previous line to be computed from this line, rather than the other way around so that the same constant value (i.e. `1048`) is used in both.

The reasoning is that there is no clear indication that these need to be kept in sync. One could imagine searching for the equivalent name or value and not finding anything.

petebacondarwin (approved):
A few minor suggestions... I never liked that `side-by-side` thing so thanks for getting rid of that.

aio/src/app/app.component.spec.ts
petebacondarwin: there is probably an equivalent `narrow` describe block in this file too.

aio/src/styles/1-layouts/_top-menu.scss
gkalpak: Good point :+1:

petebacondarwin (approved):

PR #37940, opened 2020-07-06
refactor(core): remove duplicated WrappedValue class

pkozlowski-opensource:
Before this refactoring we had the WrappedValue class in
2 separate places:
- packages/core/src/change_detection/change_detection_util.ts
- packages/core/src/util/WrappedValue.ts

This commit removes the duplicate, leaving the class that has
the deprecation notice.


pkozlowski-opensource:
@atscott would you mind reviewing / running pre-submit on this one, as you were poking around WrappedValue recently?

atscott:
[presubmit](http://test/OCL:320006978:BASE:320007030:1594141811507:d067fb35)

atscott (approved):
LGTM - I don't see any references to this anywhere so I can't imagine there will be any issue with removing it.

atscott:
caretaker note: marking as master and patch because this specific instance isn't used anywhere. It's marked as publicApi but not exported (the other one is).

Issue #34538, opened 2019-12-23
Directives don't clear provided NG_VALIDATORS when destroyed


thefliik:
# 🐞 bug report

### Affected Package
@angular/forms

### Is this a regression?

No

### Description

If a directive adds a custom validator to a form control via the `NG_VALIDATORS` provider, and that directive is destroyed (e.g. because an `ngIf` hides the directive/component), it is expected that the custom validator is removed from the form control. In reality, the custom validator is not removed from the form control.

I *think* the underlying cause of this is the same as #20007. However, the problem is not just one of a memory leak:

1. If a directive defines a custom validator and that directive is destroyed, the form control should no longer undergo the validation associated with the directive (but it does).
2. If a directive's validation is, itself variable (i.e. the directive might require a control to be `true` or `false`, depending on the *directive's* inputs), then a form control can end up with multiple versions of the directive's validators. One version requiring the control to be `true`, the other version requiring the control to be `false`. In this case, the control will always be invalid. 

## 🔬 Minimal Reproduction
https://stackblitz.com/edit/forms-issue-reproduction?file=src%2Fapp%2Fapp.component.html

## 🔥 Exception or Error
None

## 🌍  Your Environment

**Angular Version:**
<pre><code>
Angular CLI: 8.3.2
Node: 10.16.0
OS: darwin x64
Angular: 
... 

Package                      Version
------------------------------------------------------
@angular-devkit/architect    0.803.2
@angular-devkit/core         8.3.2
@angular-devkit/schematics   8.3.2
@schematics/angular          8.3.2
@schematics/update           0.803.2
rxjs                         6.4.0
</code></pre>

**Anything else relevant?**

- Related: https://github.com/angular/angular/issues/20007

AndrewKushnir:
Closing this issue since it has the same root cause as #20007 and should be resolved by #39235. Thank you.

Issue #26792, opened 2018-10-26
Angular Reactive forms custom form controls inconsistency in validations after conditionally recreated by angular


sankarswaroop:
## I'm submitting a...

<pre><code>
[x] Bug report 
</code></pre>

## Current behavior
Reactive-Forms:
Custom Form Control validations stick to initial state when recreating this control conditionally by angular

## Expected behavior
The validations should work the same every time this control is recreated

## Minimal reproduction of the problem with instructions
https://stackblitz.com/edit/angular-4ansfk
or
https://angular-4ansfk.stackblitz.io
1. Load the above url. 
2. After the screen loaded, click 'toggle' first and you can see 'Custom Name' field.
3. Type in text input and the error goes off.
4. Clear text and you can see error is back
5. Now click 'toggle' and you can see 'First Name' field
6. Click 'toggle' and you can see 'Custom Name' field with error message
7. Now type in the textinput and you can still see the error shows. The error message should go off as per the validations i written.

## What is the motivation / use case for changing the behavior?
Consistent validations with custom form controls while recreating by angular conditionally

## Environment

<pre><code>
Angular version: 5.2.10
I cannot upgrade to latest version and I have observed the same in 4.6.4 too

Browser:
- [x] Chrome (desktop) version 70.0 or earlier too
- [x] IE version 11
 
For Tooling issues:
- Node version: 8.2.0
- Platform:  Windows
</code></pre>

lloydst:
you forgot to update the dom, add a update rule:
```
constructor(private fb: FormBuilder) {
    this.formGroup = this.fb.group({
      firstName: new FormControl(),
      customName: new FormControl('', [Validators.required]),
    }, {updateOn:blur});
    this.toggle();
```

sankarswaroop:
@lloydst Thanks for your replay. This is more about on what action the validations should run. This will execute validations on blur of each control. This doesn't solved the actual problem

AndrewKushnir:
Hi @sankarswaroop, thanks for reporting the issue. It looks related to #20007 and #34538. I'll keep this ticket open, so this use-case can be verified once the issue is fixed (no ETA at this moment). Thank you.

AndrewKushnir:
Closing this issue since it has the same root cause as #20007 and should be resolved by #39235.

Issue #20007, opened 2017-10-29
Memory leak when FormControlName created/destroyed few times


IliaVolk:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead see https://github.com/angular/angular/blob/master/CONTRIBUTING.md#question
</code></pre>

## Current behavior

1. FormControlName, objects provided by NG_VALIDATORS (may be some other) can not be collected by gc until there are refs for FormGroup.
This is caused by collecting closures created in https://github.com/angular/angular/blob/4.4.6/packages/forms/src/directives/shared.ts#L33  in https://github.com/angular/angular/blob/4.4.6/packages/forms/src/model.ts#L744 .
2. All provided validators (via NG_VALIDATORS token) are called as many times as form is recreated.

## Expected behavior
1. After destroying FormControlName it should be collected by gc even if there are refs for FormGroup.
2. Provided validators should be removed from FormControl after destroying FormControlName.

## Minimal reproduction of the problem with instructions
https://plnkr.co/edit/eCtf2naoxLe8bWTJ8IkV?p=preview (used example from material demo as starting point)

#### Steps to reproduce
1. show/hide form clicking checkbox.
2. displayed below number of listeners is increasing.
3. there are more and more messages 'Validate called' appears in console when editing input (In my project I had issue when validators set outdated errors, had to use workaround).

There are 3 inputs to show that they are not synchronized (Is there a significant reason to store all closures, not only last?).

## What is the motivation/use case for changing the behavior?

1. Allow to have permanent FormControl/FormGroup without memory leaks.
2. Allow to dynamically create FormControlName with custom control, which knows how to validate itself (providing NG_VALIDATORS), without calling validators of destroyed controls.
## Environment

<pre><code>
Angular version: 4.4.6

Browser:
- [x] Chrome (desktop) version 62.0.3202.75
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX
 
Others:

</code></pre>

aalbericio:
We've hit this bug.

[...]

PR #37943, opened 2020-07-06
fix(forms): separate reactive-driven validators from teamplte-driven ones to better support mixed mode

albyrock87:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

When using ReactiveForms approach mixed with template driven validators (see [Angular Material's Date Picker](https://github.com/angular/components/blob/master/src/material/datepicker/datepicker-input.ts#L43), FormControl's validators are combined on directive's initialization.
If the developer sets its own validators via `setValidators` in a future moment, the FormControl will loose `NG_VALIDATORS` provided by the directive.

Another problem appears when the control is being recreated (by an `*ngIf` for example) and attached to the same FormControl instance: the new combined validator will contain the old validator and the new.
This is both a memory leak and a functional bug.

Issue Number: #20007, #26792, #34538

## What is the new behavior?
I've separated validators (reactive driven) from view validators (template driven) to achieve:
- Cleanup view validators on destroy to avoid strange behaviors and memory leaks
- Separate validators for model and view, this way I can use for example a DatePicker that validates formally the input (viewValidator) and a model validator to check if date respects an interval (validator)

FormControl (FormGroup, FormArray) will combine validators internally before using them to validate the value.

- [x] Yes
- [ ] No

AbstractControl.setValidators and AbstractControl.setAsyncValidators wont clear validators setted by template directives anymore.

If someone is using setValidators to voluntary remove NG_VALIDATORS setted in the view, he will need to use `formControl.viewValidators = null`.

## Other information
I searched for issues about `validators` and found issues mentioned above, but probabily there are similar issues solved by this patch.

### DIFF
--- a/packages/forms/src/directives/ng_model.ts
+++ b/packages/forms/src/directives/ng_model.ts
@@ -306,7 +306,6 @@ export class NgModel extends NgControl implements OnChanges, OnDestroy {
 
   private _setUpStandalone(): void {
     setUpControl(this.control, this);
-    this.control.updateValueAndValidity({emitEvent: false});
   }
 
   private _checkForErrors(): void {
--- a/packages/forms/src/directives/reactive_directives/form_control_directive.ts
+++ b/packages/forms/src/directives/reactive_directives/form_control_directive.ts
@@ -6,14 +6,14 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {Directive, EventEmitter, forwardRef, Inject, InjectionToken, Input, OnChanges, Optional, Output, Self, SimpleChanges} from '@angular/core';
+import {Directive, EventEmitter, forwardRef, Inject, InjectionToken, Input, OnChanges, OnDestroy, Optional, Output, Self, SimpleChange, SimpleChanges} from '@angular/core';
 
 import {FormControl} from '../../model';
 import {NG_ASYNC_VALIDATORS, NG_VALIDATORS} from '../../validators';
 import {ControlValueAccessor, NG_VALUE_ACCESSOR} from '../control_value_accessor';
 import {NgControl} from '../ng_control';
 import {ReactiveErrors} from '../reactive_errors';
-import {_ngModelWarning, composeAsyncValidators, composeValidators, isPropertyUpdated, selectValueAccessor, setUpControl} from '../shared';
+import {_ngModelWarning, cleanUpControl, composeAsyncValidators, composeValidators, isPropertyUpdated, selectValueAccessor, setUpControl} from '../shared';
 import {AsyncValidator, AsyncValidatorFn, Validator, ValidatorFn} from '../validators';
 
 
@@ -52,7 +52,7 @@ export const formControlBinding: any = {
  */
 @Directive({selector: '[formControl]', providers: [formControlBinding], exportAs: 'ngForm'})
 
-export class FormControlDirective extends NgControl implements OnChanges {
+export class FormControlDirective extends NgControl implements OnChanges, OnDestroy {
   /**
    * @description
    * Internal reference to the view model value.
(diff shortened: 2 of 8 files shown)


albyrock87:
I just saw that @AndrewKushnir ( #37881 ) and @crisbeto ( #37924 ) are working on the same problem but with a different approach.

AndrewKushnir:
Hi @albyrock87, thanks for creating this PR.

As you mentioned in your latest message, we are working on a couple PRs to fix the issue caused by not clearing validators and not invoking cleanup functions when components/directives are destroyed. We are trying to find a solution that would be backwards compatible and based on initial tests performed in Google's codebase, there is a high chance that we should be able to achieve that.

I'd like to share the current plan:

- I've created a draft PR #37881 (I need to do some cleanup and add more tests) that updates `AbstractControl` to store raw validators in addition to coerced validator function. That should provide the necessary flexibility to add/remove particular validators.
- As the next step we would expose an API (private or public) to retrieve the list of validators. In conjunction with the `setValidator` API, we should be able to get/set validators in FormControl directives.
- I also have a PR #37566 with additional logic to invoke cleanup functions and it will include validators cleanup logic once previous 2 steps are completed.

We will definitely take your PR into account while working on #37566 and I'd be curious to have your feedback on #37881 and #37566 PRs.

Thank you.

packages/forms/src/directives/ng_model.ts
```diff
@@ -306,7 +306,6 @@ export class NgModel extends NgControl implements OnChanges, OnDestroy {
 
   private _setUpStandalone(): void {
     setUpControl(this.control, this);
-    this.control.updateValueAndValidity({emitEvent: false});
```
AndrewKushnir, line 309: Just curious why `updateValueAndValidity` call was removed?

packages/forms/src/directives/reactive_directives/form_control_directive.ts
```diff
@@ -123,19 +123,34 @@ export class FormControlDirective extends NgControl implements OnChanges {
...
+      const previousControl = (changes.form as SimpleChange).previousValue;
+      if (previousControl) {
+        cleanUpControl(previousControl, this);
+      }
+
       setUpControl(this.form, this);
       if (this.control.disabled && this.valueAccessor!.setDisabledState) {
         this.valueAccessor!.setDisabledState!(true);
       }
-      this.form.updateValueAndValidity({emitEvent: false});
```
AndrewKushnir, line 130: Just curious why `updateValueAndValidity` call was removed here as well?

AndrewKushnir: Sorry, I see it now - it was moved to `setUpControl` function.

albyrock87:
Hi @AndrewKushnir ,

thanks for reviewing my PR.

I moved `updateValueAndValidity` call from directives into `setupControl` and `cleanupControl` trying to have a single code location where all directives setup are managed: maybe you can take it into account when working on #37566 (look in example at `cleanUpFormContainer`).

If you want, I can contribute to #37566 by making a PR to your branch (I don't know if this is possible/allowed - in case, let me know how to do it). 

I'm closing this PR because #37881 will offer more features (i.e. detect `required` validator) than `viewValidators` approach.

AndrewKushnir:
Hi @albyrock87, thanks for additional information.

> If you want, I can contribute to #37566 by making a PR to your branch (I don't know if this is possible/allowed - in case, let me know how to do it).

It'd be great, thank you. I think it'd be very helpful to make sure we cover all possible cases where cleanup should happen and have sufficient test coverage to verify that. I will ping you once #37881 (validators change) lands, which would unblock further work on #37566 (actual cleanup). Thank you.

Issue #37849, opened 2020-06-30
i18n example errors while extracting


Theraggon:
# 📚 Docs or angular.io bug report

### Description

 After downloading i18n example from https://angular.io/generated/zips/i18n/i18n.zip and  extracting it using 7-zip it shows warnings and some extracted files are broken like missing half a file with gibberish at the end.

Same with winRAR.

### What's the affected URL?**
 
https://angular.io/generated/zips/i18n/i18n.zip 

### Reproduction Steps**

download https://angular.io/generated/zips/i18n/i18n.zip
7-zip Test archive or extract

## 🔥 Exception or Error

image with 7-Zip errors:
https://imgur.com/5wI7WNe

End of angular.json file:
<pre>
<code>
            "fr": {
              "localize": [
                "fr"
              ]
            }
          }
        },
        "
  dm         "output     }
  tt
a vl
d     hent    "oulzer": true,
              m    r          ],
        "maximumWa e,
            ftMsiowser",
  e,
       T        "max.tob  "namedChunks"lor"
      "oulzer": true,
 
</code>
</pre>

## 🌍  Your Environment

Windows 10
7-Zip 19.00

petebacondarwin:
I just tried downloading the zip on OS/X and it cannot be expanded there either. So it appears this is no OS specific and that we are generating a corrupt zip file.

petebacondarwin:
It also fails on `next.angular.io` but is OK on `v8.angular.io`...

rockument69:
On macOS Mojave (10.14.6) I get `Unable to expand "i18n.zip" into "Downloads". (Error 2 - No such file or directory.)`. The live example shows up in stackblitz OK.

petebacondarwin:
It's definitely something to do with our zip generation. I will take a quick look later today.

petebacondarwin:
Should be fixed by #37947

PR #37947, opened 2020-07-06
build(docs-infra): remove unnecessary zipper.json file

petebacondarwin:
The `zipper.json` file is only needed if the example does not
have a `stackblitz.json` file, which this (i18n) example does.

Moreover, it appears that having both can cause the generated
zip file to be corrupted and not unzippable.

Fixes #37849


gkalpak (changes requested):
We can still have both `stackblitz.json` and `zipper.json` if we want to include different files in each. I am not sure whether that is the case here.

I understand little of the i18n example, but I noticed the following (which I am not sure whether intentional):

- [doc-files/][1] are included in both the StackBlitz and the zip. Is that intentional? Are these files useful to the user?
- [src/locale/][2] is not included in neither the StackBlitz nor the zip. Is that intentional?

[1]: https://github.com/angular/angular/tree/79850983e6fcedb82377f2784e6d59c62727b71e/aio/content/examples/i18n/doc-files
[2]: https://github.com/angular/angular/tree/79850983e6fcedb82377f2784e6d59c62727b71e/aio/content/examples/i18n/src/locale

petebacondarwin:
> We can still have both stackblitz.json and zipper.json if we want to include different files in each. I am not sure whether that is the case here.

I am not sure that this is correct. From my reading of https://github.com/angular/angular/blob/1c385a1c2200ffc90b2848b6a07c19c0293cae82/aio/tools/example-zipper/exampleZipper.js#L24 having both files will result in the zipper attempting to produce two zip files, right?

petebacondarwin:
I agree that the StackBlitz and zip file should have not include docs-files and include locale files. So we should probably add these lines to the `stackblitz.json`. 

```
    "!doc-files/**/*",	
    "**/*.xlf"
```

petebacondarwin:
Updated the stackblitz.json...

gkalpak (approved):
:100:

petebacondarwin:
Merge-assistance: Github seems to think there are two outstanding code-reviews, but this is not correct. The PR has been reviewed and approved as required.

Splaktar (approved):
LGTM

PR #37950, opened 2020-07-07
docs: Generalize the data transformation of pipes

KingMario:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

The current description of pipes is too specific about data formating which may cause some misunderstanding of the purpose and usage of pipes.

Issue Number: N/A

## What is the new behavior?

Make it more generalized about the transformation that the pipes do. Give another example of transformation.

- [ ] Yes
- [x] No

### DIFF
--- a/aio/content/guide/pipes.md
+++ b/aio/content/guide/pipes.md
@@ -1,7 +1,7 @@
 # Transforming Data Using Pipes
 
-Use [pipes](guide/glossary#pipe "Definition of a pipe") to transform and format strings, currency amounts, dates, and other display data.
-Pipes are simple functions you can use in [template expressions](/guide/glossary#template-expression "Definition of template expression") to accept an input value and return a transformed value.
+Use [pipes](guide/glossary#pipe "Definition of a pipe") to transform strings, currency amounts, dates, and other data for display.
+Pipes are simple functions you can use in [template expressions](/guide/glossary#template-expression "Definition of template expression") to accept an input value and return a transformed value. Pipes are useful because you can use them throughout your application, while only declaring each pipe once.
 For example, you would use a pipe to show a date as **April 15, 1988** rather than the raw string format.
 
 <div class="alert is-helpful">
@@ -63,7 +63,7 @@ function.
 
 {@a parameterizing-a-pipe}
 
-## Formatting data with parameters and chained pipes
+## Transforming data with parameters and chained pipes
 
 Use optional parameters to fine-tune a pipe's output.
 For example, you can use the [`CurrencyPipe`](api/common/CurrencyPipe "API reference") with a country code such as EUR as a parameter.


KingMario:
PS: in a code review for a pipe of a form control checking, my peer argued that a pipe should only format the data as described in this doc.

```
transform(control: FormControl): boolean {
    return !!control && control.touched && control.invalid;
}
```

I think, as described in the doc of Version 8, the essential of a pipe is `You may notice that you desire many of the same transformations repeatedly, both within and across many applications.` https://github.com/angular/angular.io/blob/master/public/docs/ts/latest/guide/pipes.jade#L14

That's why I make this PR and want it to be merged.

Correct me if I am wrong.

Thanks.

aio/content/guide/pipes.md
```diff
@@ -1,8 +1,9 @@
 # Transforming Data Using Pipes
 
-Use [pipes](guide/glossary#pipe "Definition of a pipe") to transform and format strings, currency amounts, dates, and other display data.
-Pipes are simple functions you can use in [template expressions](/guide/glossary#template-expression "Definition of template expression") to accept an input value and return a transformed value.
-For example, you would use a pipe to show a date as **April 15, 1988** rather than the raw string format.
+Use [pipes](guide/glossary#pipe "Definition of a pipe") to transform strings, currency amounts, dates, and other data for display (of contents, structure and styles etc.).
+A pipe is a simple function you can use repeatedly in [template expressions](/guide/glossary#template-expression "Definition of template expression") to accept an input value and return a transformed value, without declaring it in multiple places.
+For example, you would use a pipe to show a date as **April 15, 1988** rather than the raw string format. 
+Another example is a pipe that you can use to check if a form control is touched and invalid. If so, you can modify the display style of the associated input element to highlight it.
```
AndrewKushnir, line 6: I think this example is definitely valid, but feels be too specific for the guide that talks about Pipes in general.

AndrewKushnir (commented):
Thanks for the updates @KingMario. I reviewed it and left 1 comment.

I'm also adding @kapunahelewong as a reviewer to this PR to get additional feedback from the docs team perspective.

Thank you.

aio/content/guide/pipes.md
KingMario: Thank you @AndrewKushnir.

Do you think I should remove this sample pipe, as for now, the idea that a pipe is not just for formatting has been clarified in the first paragraph?

Or may I shorten it like `You would also use a pipe to determine whether to highlight an input element by checking the status of the associated form control.`?

Thanks.

AndrewKushnir: @KingMario I'd propose removing the sentence that starts with `Another example ...`. I think the date-based example is enough for this part of the document (more examples and information is available below).

Regarding other changes in the guide: I feel like current version is pretty accurate and I'd probably keep it as is for now. We could potentially improve it by re-phrasing/clarifying what "display data" is and which "raw string format" of the date we transform to **April 15, 1988**.

May be @kapunahelewong can provide additional input on this and whether we should change this guide.

KingMario: @AndrewKushnir I've removed the sentence that starts with `Another example...`.

Please review it again. Thanks.

aio/content/guide/pipes.md
```diff
@@ -1,7 +1,7 @@
 # Transforming Data Using Pipes
 
-Use [pipes](guide/glossary#pipe "Definition of a pipe") to transform and format strings, currency amounts, dates, and other display data.
-Pipes are simple functions you can use in [template expressions](/guide/glossary#template-expression "Definition of template expression") to accept an input value and return a transformed value.
+Use [pipes](guide/glossary#pipe "Definition of a pipe") to transform strings, currency amounts, dates, and other data for display (of contents, structure and styles etc.).
```
kapunahelewong, line 3: We're trying to avoid the use of parentheses as well as Latin abbreviations––for speakers of other languages with less Latin influence. I'd omit the parenthetical like this:
```suggestion
Use [pipes](guide/glossary#pipe "Definition of a pipe") to transform strings, currency amounts, dates, and other data for display.
```

aio/content/guide/pipes.md
```diff
@@ -1,7 +1,7 @@
...
+A pipe is a simple function you can use repeatedly in [template expressions](/guide/glossary#template-expression "Definition of template expression") to accept an input value and return a transformed value, without declaring it in multiple places.
```
kapunahelewong, line 4: I think you could leave out the word repeatedly because that is often you how you'd use a function. That is, ideally you wouldn't have many single-use functions. What about this?
```suggestion
A pipe is a function you can use in [template expressions](/guide/glossary#template-expression "Definition of template expression") to accept an input value and return a transformed value.
```
Notice I've removed "simple".

I'm not sure on the last phrase since this is the introductory paragraph on the topic. Getting into specific details like that might be better for later on somewhere else in the doc. @rockument69 might have some insights for us on why he chose the wording he did.

aio/content/guide/pipes.md
```diff
@@ -63,7 +63,7 @@ function.
 
 {@a parameterizing-a-pipe}
 
-## Formatting data with parameters and chained pipes
+## Transforming data with parameters and chained pipes
```
kapunahelewong, line 66: Is there a technical difference between formatting and transforming a string? I searched around and mostly found "formatting". Usually I found transforming in the case of CSS. It seems to me that there is probably a subtle difference, but I'm not sure what it is.

kapunahelewong (commented):
Thanks for these updates, @KingMario! I've shared a few thoughts, but I'd also like to ask another documentation team member to take a look. @rockument69 Recently worked a lot in this file, so he might have some helpful feedback too.

aio/content/guide/pipes.md
KingMario: Thanks @kapunahelewong 

I change this because, in my company, in a code review for a pipe, we had an argument that should a pipe only change the format of a data, or can it be used to extract the property or properties from a complex object.

The date format pipe sample and the title of this section given in this guide might make the readers think that pipes should only return formatted strings.

This is the pipe that raised the argument.

```typescript
@Pipe({
  name: 'touchedAndInvalid',
  pure: false,
})
export class TouchedAndInvalidPipe implements PipeTransform {
  transform(control: FormControl): boolean {
    return !!control && control.touched && control.invalid;
  }
}
```

Some of my peers think this should not be declared as a pipe because it doesn't return a formatted string, among them is one who had studied in the United States and has a good knowledge of English.

While according to @AndrewKushnir , it is definitely valid. 

A thought comes to me that this guide should also describe when to declare and use pipes.

I noticed, back in version 8 of this guide, there's a sentence `You may notice that you desire many of the same transformations repeatedly, both within and across many applications.`. I think it tells the essential of pipes, that is, you want to use it **repeatedly**, **directly** and **effortlessly** in template expressions without declaring it in every component.

As you have mentioned, `repeatedly` is often how you'd use a function. Then do you think I could add `directly` and `effortlessly` to the sentence like this:

`A pipe is a function you can use directly and effortlessly in [template expressions](/guide/glossary#template-expression "Definition of template expression") to accept an input value and return a transformed value.`

Does that make sense?

Thanks.

aio/content/guide/pipes.md
rockument69: I think the introductory sentence should stay as it is now (recently updated): " Pipes are simple functions you can use in template expressions to accept an input value and return a transformed value." 

As for the initial suggestion, "and other data for display (of contents, structure and styles etc.)" I would argue that "transformation" covers all of this, including changing any property, structure, style, or content. It also includes translation, for example (which deals with content). 

Also, this part of the page is all about the built-in pipes. To do more than the built-in pipes do, you must create a custom pipe, and that section is later in the doc.

aikidave: I think there's room for both ideas here.

The current introductory statement does a good job explaining what a pipe does. I think what @KingMario is trying to point out is that we could add more information about **why** you might want to use a pipe. Perhaps the following text might work?

```suggestion
Pipes are functions you can use in [template expressions](/guide/glossary#template-expression "Definition of template expression") to accept an input value and return a transformed value. Pipes are useful because you can use them throughout your application, while only declaring it once.
```

aio/content/guide/pipes.md
aikidave: In this case, the use of the word `transform` is correct. The reason we're using the term is because pipes can do more than just format values; consider the `async` pipe, for example. We view formatting as a subset of transforming. 

I'm glad you brought this up, @KingMario! These discussions are really helpful for clarifying terminology.

kapunahelewong: Thank you both. I did not know the difference and now I do!

kapunahelewong: @KingMario, I think `repeatedly` might have been better in that it is more succinct. We definitely want to make sure that the sentence doesn't get too long.

KingMario: @kapunahelewong , thanks for the help from @aikidave , I think it's OK without `repeatedly` or `directly`, `effortlessly` now. The sentence added by @aikidave, `Pipes are useful because you can use them throughout your application, while only declaring it once.`, straightforwardly answered the question `when to declare and use pipes`.

And if you agree, I'll resolve this conversation regarding the term `transform`.

Thanks.

aio/content/guide/pipes.md
KingMario: Thanks @aikidave

Do you think the last sentence should be updated to `Pipes are useful because you can use them throughout your application, while only declaring each pipe once.`?

aikidave: Excellent point! The word `it` can be very ambiguous. I agree with your suggestion!

aikithoughts (approved):
Looks good to me!

kapunahelewong (approved):
This is great, @KingMario! Now we just need to make sure it passes the tests to be eligible for merge. Could you do these things?

- [Squash](https://gist.github.com/kapunahelewong/567c30f7b353fd1c39bbf6d8343e2dca#how-to-squash) so you only have one commit.
- [Rebase](https://gist.github.com/kapunahelewong/567c30f7b353fd1c39bbf6d8343e2dca#how-to-rebase) so everything is up-to-date.
- [Amend your commit message](https://gist.github.com/kapunahelewong/567c30f7b353fd1c39bbf6d8343e2dca#how-to-amend-a-commit) so it has a body of at least 100 characters (but fewer than 120 per line)

For your commit message, you could use this if you like:

```
docs: clarify the description of pipes

This commit clarifies some of the language regarding pipes in the pipes guide.
This commit also specifies the term transforming rather than formatting.

```

Notice that there is a blank line before and after the body. Those links are my own personal notes so if you need help with these steps, please tag me. I'm happy to help you with the process. Thank you for your very thoughtful work!

AndrewKushnir (approved):

KingMario:
Thanks @kapunahelewong 

The commit is rebased and squashed. And thanks for the commit message. I'm using it.

Thanks.

AndrewKushnir:
Hi @KingMario, thanks for contributing to Angular! Your PR is now merged and the changes are propagated to angular.io website. You can see the updated page at https://angular.io/guide/pipes (if you don't see the change, please clear the cache of the browser). Thank you.

PR #37956, opened 2020-07-07
build: filter out duplicate cherry-picked commits in changelog

devversion:
Often changelogs are generated from the patch branch and then
cherry-picked into the `CHANGELOG.md` file in `master` for
better access and readability. This is problematic though as
`conventional-changelog` (the tool we use for generating the
changelog), will duplicate commits when a future changelog
is generated from `master` then (i.e. for a new minor release).

This happens because conventional-changelog always generates the
changelog from the latest tag in a given branch to `HEAD`. The
tag in the patch branch does not correspond to any SHA in `master`
so the intersection of commits is not automatically omitted.

We work around this naively (until we have a better tool provided
by dev-infra), by deduping commits that are already part of the
changelog. This has proven to work as expected in the components
repo.

### DIFF
--- a/tools/gulp-tasks/changelog.js
+++ b/tools/gulp-tasks/changelog.js
@@ -64,7 +64,7 @@ function createDedupeWriterOptions() {
 
         // Filter out commit groups which don't have any commits. Commit groups will become
         // empty if we filter out all duplicated commits.
-        return group.commits.length;
+        return group.commits.length !== 0;
       });
 
       return context;
--- a/tools/gulp-tasks/changelog.js
+++ b/tools/gulp-tasks/changelog.js
@@ -6,6 +6,9 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
+const {readFileSync} = require('fs');
+const {bold, yellow} = require('chalk');
+
 module.exports = (gulp) => () => {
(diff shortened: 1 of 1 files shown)


tools/gulp-tasks/changelog.js
```diff
@@ -16,11 +19,55 @@ module.exports = (gulp) => () => {
...
+          if (existingChangelogContent.includes(commit.subject)) {
+            console.info(yellow(`  ↺   Skipping duplicate: "${bold(commit.header)}"`));
+            return false;
+          }
+          return true;
+        });
+
+        // Filter out commit groups which don't have any commits. Commit groups will become
+        // empty if we filter out all duplicated commits.
+        return group.commits.length;
```
josephperrott, line 67: Should we be explicit as we have begun doing in the `//dev-infra` code and have `group.commits.length !== 0`

josephperrott (approved):
LGTM, one nit to take a look at if you want.

jelbourn (approved):
LGTM

tools/gulp-tasks/changelog.js
devversion: Definitely. I was just copying this from Angular Components, but generally, being explicit is better 👍

PR #37968, opened 2020-07-08
refactor(dev-infra): allow for consumption with rules_nodejs v2.0.0

devversion:
With `rules_nodejs` v2.0.0 being in RC phase currently, we should
make sure that the package is compatible so that we can use it
in the components repo in combination with rules_nodejs v2.0.0.

In v2.0.0 of the NodeJS rules, Bazel workspaces are recommended
to no longer be symlinked under a separate repository. Instead,
bazel rules and targets should be accessed directly from the
user-selected NPM repository. Usually `@npm`, so that the import
changes to `@npm//@angular/dev-infra-private/<..>`.

### DIFF
--- a/dev-infra/BUILD.bazel
+++ b/dev-infra/BUILD.bazel
@@ -38,7 +38,7 @@ genrule(
 pkg_npm(
     name = "npm_package",
     srcs = [
-        "BUILD.bazel",
+        "index.bzl",
         "//dev-infra/bazel:files",
         "//dev-infra/benchmark:files",
     ],
--- a/dev-infra/index.bzl
+++ b/dev-infra/index.bzl
@@ -1,0 +1,9 @@
+# Copyright Google LLC All Rights Reserved.
+#
+# Use of this source code is governed by an MIT-style license that can be
+# found in the LICENSE file at https://angular.io/license
+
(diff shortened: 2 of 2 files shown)


josephperrott:
cc @gregmagolan as Greg has been looking at the migration to rules_nodejs 2.0 from the rules_nodejs perspective.

josephperrott (approved):
LGTM, but lets have Greg take a look as well.

gregmagolan:
You can also remove the `bazelWorkspaces` section from `dev-infra/tmpl-package.json`:

```
  "bazelWorkspaces": {
    "npm_angular_dev_infra_private": {
      "version": "0.0.0-PLACEHOLDER",
      "rootPath": "."
    }
  }
```

gregmagolan:
Cutting rules_nodejs 2.0.0-rc.2 today likely and then 2.0.0 final next week. Lots of little bug fixes coming in during RC period which is great. A few from angular/angular.

devversion:
@gregmagolan Unfortunately I cannot remove that. If we'd remove it, the components repo unit test job would start failing, so we need to support both patterns until components repo has updated to RC.

gregmagolan (approved):

PR #37977, opened 2020-07-08
build(compiler-cli): fix bazel deps rules for ngtsc testing packages

ocombe:
## PR Type

- [x] Build related changes

2 testing packages are missing from the bazel deps rules, which means that they are not included in the releases

## What is the new behavior?
They should be released with the next version of compiler-cli

- [x] No

### DIFF
--- a/packages/compiler-cli/BUILD.bazel
+++ b/packages/compiler-cli/BUILD.bazel
@@ -61,6 +61,8 @@ pkg_npm(
     deps = [
         ":compiler-cli",
         "//packages/compiler-cli/ngcc",
+        "//packages/compiler-cli/src/ngtsc/file_system/testing",
+        "//packages/compiler-cli/src/ngtsc/logging/testing",
     ],
 )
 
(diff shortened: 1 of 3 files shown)


petebacondarwin (approved):

devversion (approved):
LGTM for dev-infra. Should the commit message capture the motivation for having these testing packages published? My understanding is that those primarily exist for our internal usage and the compiler is not public API. Though apparently there are use cases, so for later reference, it would be good to capture the motivation.

ocombe:
@devversion those packages are used by `@angular/localize` in private APIs that are actually not so private since they are used by `@angular/cli` (and by my library Locl, but I'll take responsibility if it breaks 😀).
Exporting those would help writing better tests (without needing to copy all of the code), according to @petebacondarwin they were meant to be available

devversion:
@ocombe Thanks for providing context. Makes sense! Good that this is now captured in the PR :smile:

devversion (approved):

devversion:
@petebacondarwin I've changed label to also `patch` branch. Any reason why you marked it as `master only`?

petebacondarwin:
The stuff that @ocombe wants to use is only on `master`. Also the `logging` package was only moved over in `master` too.

alxhub (approved):

PR #37982, opened 2020-07-09
docs: correct flag default values in `--strict`

dgp1130:
Docs state that `strictInjectionParameters` is true by default in `ng new`, however this is not the case in `10.0.1`. It is only set when `--strict` is provided. Clarified that the `--strict` flag is required.

`strictTemplates` does not mention anything about `--strict`, so I included a similar point that it is `true` when a new project is generated with `--strict`.

## PR Checklist

- [N/A] Tests for the changes have been added (for bug fixes / features)

## PR Type

- [X] No

### DIFF
--- a/aio/content/guide/angular-compiler-options.md
+++ b/aio/content/guide/angular-compiler-options.md
@@ -186,7 +186,7 @@ For library projects generated with the CLI, the dev configuration default is `t
 
 When `true` (recommended), reports an error for a supplied parameter whose injection type cannot be determined. When `false` (currently the default), constructor parameters of classes marked with `@Injectable` whose type cannot be resolved produce a warning.
 
-When you use the CLI command `ng new`, it is set to `true` by default in the generated project's configuration.
+When you use the CLI command `ng new --strict`, it is set to `true` in the generated project's configuration.
 
 ### `strictTemplates`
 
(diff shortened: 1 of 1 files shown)


aio/content/guide/angular-compiler-options.md
```diff
@@ -186,14 +186,15 @@ For library projects generated with the CLI, the dev configuration default is `t
 
 When `true` (recommended), reports an error for a supplied parameter whose injection type cannot be determined. When `false` (currently the default), constructor parameters of classes marked with `@Injectable` whose type cannot be resolved produce a warning.
 
-When you use the CLI command `ng new`, it is set to `true` by default in the generated project's configuration.
+When you use the CLI command `ng new --strict`, it is set to `true` by default in the generated project's configuration.
```
jbogarthyde, line 189: That sounds like the flag actually overrides the default, so probably leave out the "by default" here. 
In this case, the default is how it is set when you do NOT provide the flag.

jbogarthyde (changes requested):

aio/content/guide/angular-compiler-options.md
dgp1130: Good point, removed `by default` since this option is just overriding that default value in the generated configuration.

jbogarthyde (approved):
LGTM

dgp1130:
@AndrewKushnir, do you need to take a look at this? PullApprove added you, but I'm not sure if that's expected or not?

AndrewKushnir (approved):
LGTM 👍

Sorry for the delay in reviewing this PR.

dgp1130:
Change seems safe, but the test deploy is failing due to my username? Do I need some permission to get this green or something?

AndrewKushnir:
The `deploy_aio` fails because this PR is based on the branch created directly in `angular/angular` repo. The change is safe and we can merge it (feel free to add the `merge` label back).

dgp1130:
Ah, because I created this directly in the `angular/angular` repo rather than my personal fork. Thanks @AndrewKushnir!

PR #37994, opened 2020-07-09
ci: add more owners for some categories

jelbourn:
* Add petebacondarwin to public-api, size-tracking, and
-circular-dependencies
* Add josephperrott and jelbourn to code-ownership

### DIFF
--- a/.pullapprove.yml
+++ b/.pullapprove.yml
@@ -1141,9 +1141,10 @@ groups:
           ])
     reviewers:
       users:
-        - alxhub
         - IgorMinar
+        - alxhub
         - jelbourn
+        - petebacondarwin
         - pkozlowski-opensource
     reviews:
       request: -1   # request reviews from everyone
@@ -1164,9 +1165,10 @@ groups:
           ])
     reviewers:
       users:
-        - alxhub
         - IgorMinar
+        - alxhub
         - jelbourn
+        - petebacondarwin
         - pkozlowski-opensource
     reviews:
       request: -1   # request reviews from everyone
@@ -1190,6 +1192,7 @@ groups:
         - IgorMinar
         - jelbourn
         - josephperrott
+        - petebacondarwin
         - pkozlowski-opensource
 
 
@@ -1211,6 +1214,9 @@ groups:
     reviewers:
       users:
         - IgorMinar
+        - jelbourn
+        - josephperrott
+        - mhevery
 
 
   # ====================================================


petebacondarwin (approved):
Lucky me!

.pullapprove.yml
```diff
@@ -1209,6 +1212,9 @@ groups:
     reviewers:
       users:
         - IgorMinar
+        - jelbourn
+        - josephperrott
+        - mhevery
```
IgorMinar, line 1217: Let's add also add:

```
reviews:
      request: -1   # request reviews from everyone
      required: 2   # require at least 2 approvals
```

This group is the root group in the repo, and it would make little sense if the rule for approving this group were more lenient than the requirements for the public-api or size-tracking groups.

.pullapprove.yml
```diff
@@ -1166,6 +1167,7 @@ groups:
         - IgorMinar
         - jelbourn
         - pkozlowski-opensource
+        - petebacondarwin
```
IgorMinar, line 1170: please alpha-sort

.pullapprove.yml
```diff
@@ -1189,6 +1191,7 @@ groups:
         - jelbourn
         - josephperrott
         - pkozlowski-opensource
+        - petebacondarwin
```
IgorMinar, line 1194: please alpha-sort

.pullapprove.yml
```diff
@@ -1143,6 +1143,7 @@ groups:
         - IgorMinar
         - jelbourn
         - pkozlowski-opensource
+        - petebacondarwin
```
IgorMinar, line 1146: please alpha-sort

IgorMinar (changes requested):

.pullapprove.yml
jelbourn: That seems a bit like overkill to me. The reason public-api approvers requires more people is because API decisions have a level of judgement involved in each change and warrant some discussion. This, by contrast, just seems like an accounting of who's responsible for this time which shouldn't be controversial.

jenhsieh: Let's keep to the current policy and not add the additional requirement of 2 approvals. I trust our team not to add themselves or each other to code owners irresponsibly.

jelbourn:
Caretaker: this PR needs assistance because Igor is currently the only one who can approve changes to this code owners, which this PR fixes.

PR #37999, opened 2020-07-10
docs: Testing for real response with TestBed

OmarioHassan:
I think it will worth adding an example for testing on real response
I know this can be written in a more perfect style but I was asking if this really worth to mention or not

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [X] No

### DIFF
--- a/aio/content/guide/http.md
+++ b/aio/content/guide/http.md
@@ -1137,3 +1137,49 @@ Alternatively, you can call `request.error()` with an `ErrorEvent`.
   path="http/src/testing/http-client.spec.ts"
   region="network-error">
 </code-example>
+
+### Testing for real response
+
+Now you can write a test that get a real response and make your expect for it.
+
+Imgagine we need to get a `post` from `https://jsonplaceholder.typicode.com`
+Then we need to have an interface for every post `post.ts` like this:
+
+`export interface Post {
+  userId: number;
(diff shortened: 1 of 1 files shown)


OmarioHasan:
@AndrewKushnir i this didn't really worth mentioning?

kapunahelewong:
Thanks for this, @OmarioHasan! I can only speak from a documentation perspective, and we will definitely need engineering feedback on this before we even get to docs style. However, there are a couple of notes about documentation style to have in mind if you do move ahead with this. We are trying to avoid the use of the first person - so no "we", "let's", etc. And we're working to make as many code snippets as possible be part of working examples. So, if you do end up using those chunks of code, you'll want to add that functionality to the `http` example app if possible. Let me know if you have any questions. Thanks!

AndrewKushnir:
Hi @OmarioHasan, thanks for the proposed improvement. This guide mentions that the HTTP service should be mocked for testing purposes. That should avoid having flaky tests that depend on the network and the status of the remote service. Developers may still prefer to use live remote services in some cases, but this is outside of the context of the mentioned guide, so I think we should not include this section. Thank you.

PR #38003, opened 2020-07-10
fix(ngcc): report a warning if ngcc tries to use a solution-style tsconfig

### LINKED ISSUE #36386: Angular 9 (Ivy) App build with local library (mono-repo) failing
# 🐞 Bug report

- [ ] new
- [x] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?
Haven't tested in older version, so i'm unsure, but considering that this is Ivy related, i would imagine it is regression.

### Description

I have an Angular 9 app that i'm starting to move some of the components into a dedicated library project, which uses secondary entry points (trying to mimic Angular Material's approach).  Everything builds fine if i build both library and app with the Ivy compiler, but if i build both with the --prod flag (which causes the library to be built without Ivy), the app build fails with the following error:
```
Compiling @fabric/components/breadcrumbs : es2015 as esm2015

ERROR in node_modules/@fabric/components/breadcrumbs/fab-breadcrumbs.module.d.ts:1:22 - error NG6002: Appears in the NgModule.imports of AppModule, but could not be resolved to an NgModule class.

This likely means that the library (@fabric/components/breadcrumbs) which declares FabBreadcrumbsModule has not been processed correctly by ngcc, or is not compatible with Angular Ivy. Check if a newer version of the library is available, and update if s
o. Also consider checking with the library's authors to see if the library is expected to be compatible with Ivy.

1 export declare class FabBreadcrumbsModule {
                       ~~~~~~~~~~~~~~~~~~~~
```
I tried updating to the latest version (9.0.2 -> 9.1.0), but it's still doing the same thing.  I tried recreating this with a brand new super simple mono-repo project, but in that simple case it works fine.  I compared the tsconfig and angular.json config files between my app and the simple working app, but they're fairly identical.  As a last resort, i added `"postinstall": "ngcc"` to my root package.json, wiped out my library's folder in my node_modules dir, and re-ran npm install.  This forced ngcc to run on everything including my library.  After that, the app build worked as it should.  So it would seem that the app build isn't recognizing that the library build was done with viewEngine and kicking off the ngcc itself? 

## 🔥 Exception or Error
<pre><code>
Compiling @fabric/components/breadcrumbs : es2015 as esm2015

ERROR in node_modules/@fabric/components/breadcrumbs/fab-breadcrumbs.module.d.ts:1:22 - error NG6002: Appears in the NgModule.imports of AppModule, but could not be resolved to an NgModule class.

This likely means that the library (@fabric/components/breadcrumbs) which declares FabBreadcrumbsModule has not been processed correctly by ngcc, or is not compatible with Angular Ivy. Check if a newer version of the library is available, and update if s
o. Also consider checking with the library's authors to see if the library is expected to be compatible with Ivy.

1 export declare class FabBreadcrumbsModule {
                       ~~~~~~~~~~~~~~~~~~~~
</code></pre>


## 🌍 Your Environment
<pre><code>
<!-- run `ng version` and paste output below -->
Angular CLI: 9.1.0
Node: 12.13.0
OS: win32 x64

Angular: 9.1.0
Ivy Workspace: Yes

@angular-devkit/architect          0.901.0
@angular-devkit/build-angular      0.901.0
@angular-devkit/build-ng-packagr   0.901.0
@angular-devkit/build-optimizer    0.901.0
@angular-devkit/build-webpack      0.901.0
@angular-devkit/core               9.1.0
@angular-devkit/schematics         9.1.0
@angular/material                  9.2.0
@ngtools/webpack                   9.1.0
@schematics/angular                9.1.0
@schematics/update                 0.901.0
</code></pre>

**Anything else relevant?**
<!-- ✍️Is this a browser specific issue? If so, please specify the browser and version. -->

<!-- ✍️Do any of these matter: operating system, IDE, package manager, HTTP server, ...? If so, please mention it below. -->

petebacondarwin:
In CLI v10 there was a move to use the new solution-style tsconfig
which became available in TS 3.9. The result of this is that the standard
tsconfig.json, which ngcc (and ngc and tsc) infer if not given a explicit
tsconfig file path no longer contains important information such as
"paths" mappings, which ngcc might need to correctly compute
dependencies.

This commit logs a warning in this case to inform the developer
that they might not have meant to load this tsconfig and offer
alternative options.

Fixes #36386

### DIFF
--- a/packages/compiler-cli/ngcc/src/ngcc_options.ts
+++ b/packages/compiler-cli/ngcc/src/ngcc_options.ts
@@ -193,12 +193,12 @@ export function getSharedSetup(options: NgccOptions): SharedSetup&RequiredNgccOp
         `The inferred tsconfig file "${tsConfig.project}" appears to be "solution-style" ` +
         `since it contains no root files but does contain project references.\n` +
         `This is probably not wanted, since ngcc is unable to infer settings like "paths" mappings from such a file.\n` +
-        `Perhaps you should have explicitly specified one of the referenced projects using the --tsconfig option. For example:\n` +
-        tsConfig.projectReferences.map(ref => `ngcc ... --tsconfig "${ref.originalPath}"\n`)
+        `Perhaps you should have explicitly specified one of the referenced projects using the --tsconfig option. For example:\n\n` +
+        tsConfig.projectReferences.map(ref => `  ngcc ... --tsconfig "${ref.originalPath}"\n`)
             .join('') +
-        `Find out more about solution-style tsconfig at https://devblogs.microsoft.com/typescript/announcing-typescript-3-9/#solution-style-tsconfig.\n` +
-        `If you did intend to use this file, then you can hide this warning by providing it explicitly:\n` +
-        `ngcc ... --tsconfig ${fileSystem.relative(projectPath, tsConfig.project)}`);
+        `\nFind out more about solution-style tsconfig at https://devblogs.microsoft.com/typescript/announcing-typescript-3-9/#solution-style-tsconfig.\n` +
+        `If you did intend to use this file, then you can hide this warning by providing it explicitly:\n\n` +
+        `  ngcc ... --tsconfig "${fileSystem.relative(projectPath, tsConfig.project)}"`);
   }
 
   return {
--- a/packages/compiler-cli/ngcc/src/ngcc_options.ts
+++ b/packages/compiler-cli/ngcc/src/ngcc_options.ts
@@ -186,20 +186,7 @@ export function getSharedSetup(options: NgccOptions): SharedSetup&RequiredNgccOp
     errorOnFailedEntryPoint = true;
   }
 
-  if (options.tsConfigPath !== null && !options.tsConfigPath && tsConfig !== null &&
-      tsConfig.rootNames.length === 0 && tsConfig.projectReferences !== undefined &&
-      tsConfig.projectReferences.length > 0) {
-    logger.warn(
-        `The inferred tsconfig file "${tsConfig.project}" appears to be "solution-style" ` +
-        `since it contains no root files but does contain project references.\n` +
-        `This is probably not wanted, since ngcc is unable to infer settings like "paths" mappings from such a file.\n` +
-        `Perhaps you should have explicitly specified one of the referenced projects using the --tsconfig option. For example:\n\n` +
-        tsConfig.projectReferences.map(ref => `  ngcc ... --tsconfig "${ref.originalPath}"\n`)
-            .join('') +
-        `\nFind out more about solution-style tsconfig at https://devblogs.microsoft.com/typescript/announcing-typescript-3-9/#solution-style-tsconfig.\n` +
-        `If you did intend to use this file, then you can hide this warning by providing it explicitly:\n\n` +
-        `  ngcc ... --tsconfig "${fileSystem.relative(projectPath, tsConfig.project)}"`);
-  }
+  checkForSolutionStyleTsConfig(fileSystem, logger, projectPath, options.tsConfigPath, tsConfig);
 
   return {
     basePath,
@@ -247,4 +234,23 @@ function getTsConfig(tsConfigPath: string): ParsedConfiguration|null {
 export function clearTsConfigCache() {
   tsConfigPathCache = null;
   tsConfigCache = null;
+}
+
+function checkForSolutionStyleTsConfig(
+    fileSystem: FileSystem, logger: Logger, projectPath: AbsoluteFsPath,
+    tsConfigPath: string|null|undefined, tsConfig: ParsedConfiguration|null): void {
+  if (tsConfigPath !== null && !tsConfigPath && tsConfig !== null &&
+      tsConfig.rootNames.length === 0 && tsConfig.projectReferences !== undefined &&
+      tsConfig.projectReferences.length > 0) {
+    logger.warn(
+        `The inferred tsconfig file "${tsConfig.project}" appears to be "solution-style" ` +
+        `since it contains no root files but does contain project references.\n` +
+        `This is probably not wanted, since ngcc is unable to infer settings like "paths" mappings from such a file.\n` +
+        `Perhaps you should have explicitly specified one of the referenced projects using the --tsconfig option. For example:\n\n` +
+        tsConfig.projectReferences.map(ref => `  ngcc ... --tsconfig "${ref.originalPath}"\n`)
+            .join('') +
+        `\nFind out more about solution-style tsconfig at https://devblogs.microsoft.com/typescript/announcing-typescript-3-9/#solution-style-tsconfig.\n` +
+        `If you did intend to use this file, then you can hide this warning by providing it explicitly:\n\n` +
+        `  ngcc ... --tsconfig "${fileSystem.relative(projectPath, tsConfig.project)}"`);
+  }
 }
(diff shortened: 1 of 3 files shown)


packages/compiler-cli/ngcc/src/ngcc_options.ts
```diff
@@ -186,6 +186,21 @@ export function getSharedSetup(options: NgccOptions): SharedSetup&RequiredNgccOp
...
+    logger.warn(
+        `The inferred tsconfig file "${tsConfig.project}" appears to be "solution-style" ` +
+        `since it contains no root files but does contain project references.\n` +
+        `This is probably not wanted, since ngcc is unable to infer settings like "paths" mappings from such a file.\n` +
+        `Perhaps you should have explicitly specified one of the referenced projects using the --tsconfig option. For example:\n` +
+        tsConfig.projectReferences.map(ref => `ngcc ... --tsconfig "${ref.originalPath}"\n`)
+            .join('') +
+        `Find out more about solution-style tsconfig at https://devblogs.microsoft.com/typescript/announcing-typescript-3-9/#solution-style-tsconfig.\n` +
+        `If you did intend to use this file, then you can hide this warning by providing it explicitly:\n` +
+        `ngcc ... --tsconfig ${fileSystem.relative(projectPath, tsConfig.project)}`);
```
JoostK, line 201: For consistency:
```suggestion
        `ngcc ... --tsconfig "${fileSystem.relative(projectPath, tsConfig.project)}"`);
```

packages/compiler-cli/ngcc/src/ngcc_options.ts
```diff
@@ -186,6 +186,21 @@ export function getSharedSetup(options: NgccOptions): SharedSetup&RequiredNgccOp
...
 
+  if (options.tsConfigPath !== null && !options.tsConfigPath && tsConfig !== null &&
+      tsConfig.rootNames.length === 0 && tsConfig.projectReferences !== undefined &&
+      tsConfig.projectReferences.length > 0) {
+    logger.warn(
+        `The inferred tsconfig file "${tsConfig.project}" appears to be "solution-style" ` +
+        `since it contains no root files but does contain project references.\n` +
+        `This is probably not wanted, since ngcc is unable to infer settings like "paths" mappings from such a file.\n` +
+        `Perhaps you should have explicitly specified one of the referenced projects using the --tsconfig option. For example:\n` +
+        tsConfig.projectReferences.map(ref => `ngcc ... --tsconfig "${ref.originalPath}"\n`)
```
JoostK, line 197: I think this will be more legible with some additional whitespace, e.g. have surrounding newlines and a small indent for the commands:

image

It currently looks like this, which I find harder to read:

image

packages/compiler-cli/ngcc/src/ngcc_options.ts
```diff
@@ -186,6 +186,21 @@ export function getSharedSetup(options: NgccOptions): SharedSetup&RequiredNgccOp
     errorOnFailedEntryPoint = true;
   }
 
+  if (options.tsConfigPath !== null && !options.tsConfigPath && tsConfig !== null &&
```
JoostK, line 189: nit: I would move this while block to immediately follow the declaration of `tsConfig`

JoostK (approved):
Type in commit message: "a explicit" -> "an explicit"

packages/compiler-cli/ngcc/src/ngcc_options.ts
petebacondarwin: Unfortunately that would be before we have a `logger` object.

petebacondarwin: And I thought, that it would be cleaner to have it directly before the `return` rather than between the call to `setup()` and the `if (!!targetEntryPointPath)` block.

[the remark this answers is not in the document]
JoostK: Ha, that was exactly my thought process! I figured the guard was quite complex, especially the `tsConfigPath !== null && !tsConfigPath` bit, but after reading the message it became clear. I was about to suggest introducing something like:

```ts
const isTsConfigInferred = options.tsConfigPath !== null && !options.tsConfigPath && tsConfig !== null;
```

but it would not allow `tsConfig` to be narrowed to non-null. Excluding that check from `isTsConfigInferred` back into the if-guard could still be an option, but at that point it doesn't add too much clarity.

JoostK: Oh, I see. Yes, agreed then. One possibility would be to pull it all out into a separate function.

petebacondarwin: That is a very good suggestion!!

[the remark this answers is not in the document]
petebacondarwin: Moving it into another function also helps with documenting it.

alxhub:
[Presubmit](http://test/OCL:320646003:BASE:320646031:1594407245150:16d176ea)

alan-agius4 (approved):
LGTM

rhysowen9:
I think it would be beneficial to update the Angular Ivy guide to explain _which_ tsconfig file to specify.
The current docs explain that ngcc is beneficial, but as soon as I add it, I start getting this ⬆️ new warning.

petebacondarwin:
@rhysowen9 - how are you running ngcc? What is the content of the warning that you get? It should be listing the paths to the possible tsconfigs that you can choose. It would be normal to run ngcc against the "app" config (possibly `tsconfig.app.json`) but exactly which file would depend upon your set up. If you are using Angular CLI then you should not need to run ngcc yourself, since it takes care of that (including passing the appropriate tsconfig path).

rhysowen9:
@petebacondarwin ,

I believe that `ngcc` was added to our `package.json` file when upgrading from Angular 8 to Angular 9 back in February (would have been done using `ng update`.

It is a `postinstall` script which specifies `--properties es2015 browser module main --first-only --create-ivy-entry-points`.

Now that we have upgraded to Angular 10, and updated to the latest patch, we are getting the warning introduced by this PR.

We've stripped `--create-ivy-entry-points` and set `--tsconfig ./src/tsconfig.app.json` which seems to be working fine, but I can't say I fully understand this side of Angular, so this is nothing more than a best guess as the docs don't seem to offer much clarity.

petebacondarwin:
Please feel free to remove this postinstall hook, as long as you are using CLI for building your app.

maxisam:
> Please feel free to remove this postinstall hook, as long as you are using CLI for building your app.

Could you explain why we can remove that postinstall hook? It is still in the document

https://update.angular.io/#9.0:10.0l3

If you depend on many Angular libraries you may consider speeding up your build by invoking the ngcc (Angular Compatibility Compiler) in an npm postinstall script via small change to your package.json.

bhanna1693:
> Please feel free to remove this postinstall hook, as long as you are using CLI for building your app.

I agree with @maxisam. The postinstall script is "recommended" in both the angular documentation and the angular update guides. Can we get clarification?

petebacondarwin:
Currently our docs do not recommend running ngcc as a postinstall hook, except for a particular scenario related to Angular Universal: https://angular.io/guide/ivy#ivy-and-universalapp-shell.
I agree that we should fix the update site. I have already created an issue there for it: https://github.com/StephenFluin/angular-update-guide/issues/76

bhanna1693:
@petebacondarwin ah you're right, I misread that page i guess. Thank you for the help!

petebacondarwin:
No problem. Sorry for the confusion

PR #38014, opened 2020-07-11
fix(core): error due to integer overflow when there are too many host bindings

### LINKED ISSUE #37876: Hostbinding fails when there are several thousand context matches
# 🐞 bug report

### Affected Package

### Is this a regression?

### Description

On a page with a large number of [routerLink] directives the following Error is thrown.  The number of router links appears to be somewhere > 3000.

## 🔬 Minimal Reproduction

 https://stackblitz.com/...

A repro repo can be found here:  https://github.com/langfors/manyRouterlinks

## 🔥 Exception or Error
<pre><code>

core.js:6228 ERROR TypeError: Cannot read property 'href' of undefined
    at RouterLinkWithHref_HostBindings (router.js:8437)
    at setHostBindingsByExecutingExpandoInstructions (core.js:11665)
    at refreshView (core.js:12046)
    at refreshComponent (core.js:13445)
    at refreshChildComponents (core.js:11716)
    at refreshView (core.js:12051)
    at refreshEmbeddedViews (core.js:13391)
    at refreshView (core.js:12022)
    at refreshComponent (core.js:13445)
    at refreshChildComponents (core.js:11716)
</code></pre>

## 🌍  Your Environment

**Angular Version:**
<pre><code>

</code>9.1.11</pre>

**Anything else relevant?**

crisbeto:
We currently use 16 bits to store information about nodes in a view. The 16 bits give us 65536 entries in the array, but the problem is that while the number is large, it can be reached by ~4300 directive instances with host bindings which could realistically happen in a very large view, as seen in #37876. Once we hit the limit, we end up overflowing which eventually leads to a runtime error.

These changes bump to using 20 bits which gives us around 1048576 entries in the array or 16 times more than the current amount which could still technically be reached, but is much less likely and the user may start hitting browser limitations by that point.

I picked the 20 bit number since it gives us enough buffer over the 16 bit one, while not being as massive as a 24 bit or 32 bit.

I've also added a dev mode assertion so it's easier to track down if it happens again in the future.

Fixes #37876.

### DIFF
--- a/packages/core/src/render3/instructions/shared.ts
+++ b/packages/core/src/render3/instructions/shared.ts
@@ -11,7 +11,7 @@ import {CUSTOM_ELEMENTS_SCHEMA, NO_ERRORS_SCHEMA, SchemaMetadata} from '../../me
 import {ViewEncapsulation} from '../../metadata/view';
 import {validateAgainstEventAttributes, validateAgainstEventProperties} from '../../sanitization/sanitization';
 import {Sanitizer} from '../../sanitization/sanitizer';
-import {assertDataInRange, assertDefined, assertDomNode, assertEqual, assertGreaterThan, assertNotEqual, assertNotSame, assertSame} from '../../util/assert';
+import {assertDataInRange, assertDefined, assertDomNode, assertEqual, assertGreaterThan, assertLessThan, assertNotEqual, assertNotSame, assertSame} from '../../util/assert';
 import {createNamedArrayType} from '../../util/named_array_type';
 import {initNgDevMode} from '../../util/ng_dev_mode';
 import {normalizeDebugBindingName, normalizeDebugBindingValue} from '../../util/ng_reflect';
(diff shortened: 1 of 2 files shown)


pkozlowski-opensource (commented):
The change looks good and this is corner case but still it would be good to have a test for this particular scenario. Do you think it would make sense to generate a directive with a large number of host bindings (256+) and a template with 256+ usages of such directive? Or did you assume that it is impractical / not worth having a test for this case?

crisbeto:
I thought that it would be impractical and it wouldn't provide much value. The basic logic is already covered by other test cases and all we'd be testing is higher and higher numbers.

pkozlowski-opensource (approved):
LGTM

AndrewKushnir:
[Presubmit](http://test/OCL:321639499:BASE:321639544:1594932997506:a5c39114)

PR #38020, opened 2020-07-12
refactor(forms): refactor common validators used in unit tests

### LINKED ISSUE #37831: Forms: refactor common validators used in unit tests and extract them to util file
# 🐞 bug report

### Affected Package

 The issue is caused by package @angular/forms

### Is this a regression?

No, this is a refactor of the existing unit tests

### Description
Following https://github.com/angular/angular/pull/22575 , further refactoring around async validators in `forms` unit tests were suggested by @AndrewKushnir 
https://github.com/angular/angular/pull/22575#issuecomment-650664473

> I was thinking if we can extract common validators code into a new `packages/forms/test/util.ts` file (we can store more functions there later). Some validators that are used across many tests (may be you can find more):
> 
> * function asyncValidator
> * function asyncValidatorReturningObservable
> * may be `simpleAsyncValidator` added in this test can use some common code from `asyncValidator`
> * I also see a value of moving the `currentStateOf` function (added in this PR) into the `util.ts` file, so we can reuse it for other validators-related tests in other files

## 🔬 Minimal Reproduction
N/A

## 🔥 Exception or Error
N/A

## 🌍  Your Environment

**Angular Version:**
`master` branch

**Anything else relevant?**
No

benbraou:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

`asyncValidator` is duplicated across multiple form test files.

Issue Number: N/A
 #37831

## What is the new behavior?
A util file is added to forms test package:
- it exposes `simpleAsyncValidator`, `asyncValidator` and `asyncValidatorReturningObservable` validators
- it refactors `simpleAsyncValidator` and `asyncValidator` to use common promise creation code
- it exposes `currentStateOf` allowing to get the validation state of a list of `AbstractControl`

- [ ] Yes
- [x] No

### DIFF
--- a/packages/forms/test/util.ts
+++ b/packages/forms/test/util.ts
@@ -1,0 +1,87 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {EventEmitter} from '@angular/core';
+import {AbstractControl, AsyncValidatorFn, ValidationErrors} from '@angular/forms';
+import {of} from 'rxjs';
+
+function createValidationPromise(
+    result: ValidationErrors|null, timeout: number): Promise<ValidationErrors|null> {
+  return new Promise(resolve => {
+    if (timeout == 0) {
+      resolve(result);
+    } else {
+      setTimeout(() => {
+        resolve(result);
+      }, timeout);
+    }
+  });
+}
+
+/**
+ * Returns a promise-based async validator that emits, after a delay, either:
+ *  - an error `{async: true}` if the control value does not match the expected value
+ *  - or null, otherwise
+ *  The delay is either:
+ *  - defined in `timeouts` parameter, as the association to the control value
+ *  - or 0ms otherwise
+ *
+ * @param expected The expected control value
+ * @param timeouts A dictionary associating a control value to when the validation will trigger for
+ *     that value
+ */
+export function asyncValidator(expected: string, timeouts = {}): AsyncValidatorFn {
+  return (control: AbstractControl) => {
+    const timeout = (timeouts as any)[control.value] ?? 0;
+    const result = control.value != expected ? {async: true} : null;
+    return createValidationPromise(result, timeout);
+  };
+}
+
+/**
+ * Returns an async validator that emits null or a custom error after a specified delay.
+ * If the delay is set to 0ms, the validator emits synchronously.
+ *
+ * @param timeout Indicates when the validator will emit
+ * @param shouldFail When true, a validation error is emitted, otherwise null is emitted
+ * @param customError When supplied, overrides the default error `{async: true}`
+ */
+export function simpleAsyncValidator({
+  timeout = 0,
+  shouldFail,
+  customError =
+  {
+    async: true
+  }
+}: {timeout?: number, shouldFail: boolean, customError?: any}): AsyncValidatorFn {
+  const result = shouldFail ? customError : null;
+  return (c: AbstractControl) =>
+             timeout === 0 ? of(result) : createValidationPromise(result, timeout);
+}
+
+/**
+ * Returns the asynchronous validation state of each provided control
+ * @param controls A collection of controls
+ */
+export function currentStateOf(controls: AbstractControl[]):
+    {errors: any; pending: boolean; status: string;}[] {
+  return controls.map(c => ({errors: c.errors, pending: c.pending, status: c.status}));
+}
+
+/**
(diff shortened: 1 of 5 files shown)


packages/forms/test/util.ts
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {EventEmitter} from '@angular/core';
+import {AbstractControl, AsyncValidatorFn, ValidationErrors} from '@angular/forms';
+import {of} from 'rxjs';
+
+function createValidationPromise(res: any, timeout: number) {
```
AndrewKushnir, line 13: It looks like this function can be refactored/simplified a bit more (I also renamed `res` => `result` and added type to the `result` argument and return value of the function):

```
function createValidationPromise(result: ValidationErrors|null, timeout: number): Promise<ValidationErrors|null> {
  return new Promise(resolve => {
    if (timeout == 0) {
      resolve(result);
    } else {
      setTimeout(() => {
        resolve(result);
      }, timeout);
    }
  });
}
```

What do you think?

packages/forms/test/util.ts
```diff
@@ -0,0 +1,67 @@
...
+  } else {
+    setTimeout(() => {
+      resolve(res);
+    }, timeout);
+  }
+
+  return promise;
+}
+
+export function asyncValidator(expected: string, timeouts = {}) {
```
AndrewKushnir, line 30: I'd propose to refactor this function a bit:
- add return value type
- use full const names

So the code may look like this:
```
export function asyncValidator(expected: string, timeouts = {}): AsyncValidatorFn {
  return (control: AbstractControl) => {
    const timeout = (timeouts as any)[control.value] ?? 0;
    const result = control.value != expected ? {async: true} : null;
    return createValidationPromise(result, timeout);
  };
}
```

packages/forms/test/util.ts
```diff
@@ -0,0 +1,67 @@
...
+
+export function simpleAsyncValidator({
+  timeout = 0,
+  shouldFail,
+  customError =
+  {
+    async: true
+  }
+}: {timeout?: number, shouldFail: boolean, customError?: any}): AsyncValidatorFn {
+  const res = shouldFail ? customError : null;
```
AndrewKushnir, line 47: Let's use full name here as well (i.e. `result`).

packages/forms/test/util.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  if (timeout === 0) {
+    return (c: AbstractControl) => of(res);
+  }
+
+  return (c: AbstractControl) => createValidationPromise(res, timeout);
```
AndrewKushnir, line 53: I think we can also refactor this a bit more (move the `if` logic inside the validator fn):
```
  return (c: AbstractControl) => timeout === 0 ? of(result) : createValidationPromise(result, timeout);
```

packages/forms/test/util.ts
```diff
@@ -0,0 +1,67 @@
...
+}
+
+export function currentStateOf(controls: AbstractControl[]):
+    {errors: any; pending: boolean; status: string;}[] {
+  return controls.map(c => ({errors: c.errors, pending: c.pending, status: c.status}));
+}
+
+export function asyncValidatorReturningObservable(c: AbstractControl) {
```
AndrewKushnir, line 61: Let's add return value type here as well.

AndrewKushnir (changes requested):
Thanks for performing this refactoring @benbraou 👍 

I've left a few comments, please have a look when you get a chance.

I think it'd also be great to add some light-weight docs (describing what the function does and what arguments are) for all exported functions in the `util.ts` file.

Thank you.

packages/forms/test/util.ts
benbraou: Sounds good. Thank you for the review, and sorry for the delay to update this PR. I had limited access in the past weeks

packages/forms/test/util.ts
benbraou: updated

packages/forms/test/util.ts
benbraou: ok

packages/forms/test/util.ts
benbraou: sounds good

packages/forms/test/util.ts
benbraou: ok

benbraou:
Thank you for the review @AndrewKushnir 🙏. I updated the PR accordingly

AndrewKushnir (approved):
Thanks for the updates @benbraou, the changes look great 👍

PR #38026, opened 2020-07-13
Fix dev-infra zone.js substitution

wagnermaciel:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] angular.io application / infrastructure changes
- [ ] Other... Please describe:

## What is the current behavior?

Issue Number: N/A

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

PR #38029, opened 2020-07-13
feat(dev-infra): add bazel firefox browser with RBE compatibility

devversion:
Adds Firefox as browser to `dev-infra/browsers` with RBE
compatibility. The default Firefox browser is not compatible similar to
the default Chromium version exposed by `rules_webtesting`.

The Angular Components repository will use this browser target as
it enables RBE support. Also it gives us more flexibility about
the Firefox version we test against. The version provided by
`rules_webtesting` is very old and most likely not frequently
updated (based on past experience).

Additionally, for a more structured `dev-infra/browsers` folder, chromium
and firefox specifics are stored in dedicated folders. This results in
unexpected diff for chromium, but there is no change other than it
being moved into a subfolder. 

Note for reviewer that the webtesting rules require us to explicitly specify
`dev-infra/browsers/chromium:chromium` when consuming the browsers
as it picks up the browser name from the target name after the colon.


wagnermaciel:
@devversion It seems like something is broken but I don't think it was a change you made. I'm getting this error

```
ERROR: error loading package '': Label '@npm_angular_dev_infra_private//browsers:browser_repositories.bzl' is invalid because 'browsers' is not a package; perhaps you meant to put the colon here: '@npm_angular_dev_infra_private//:browsers/browser_repositories.bzl'?
```

wagnermaciel:
I've noticed one small issue with the new bazel package that I think we can resolve here. Lets discuss tomorrow :+1:

wagnermaciel (approved):
LGTM. As per our discussion I will handle the BUILD file issue in PR #38026

JiaLiPassion (approved):
LGTM for the `zone.js` part.

PR #38044, opened 2020-07-14
build(forms): create sample forms app

sonukapoor:
This commit creates a sample forms test application to introduce the symbol tests. It serves as a guard to ensure that any future work on the forms package does not unintentionally increase the payload size.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [X] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [X] No

### DIFF
--- a/packages/core/test/bundling/forms/BUILD.bazel
+++ b/packages/core/test/bundling/forms/BUILD.bazel
@@ -1,6 +1,6 @@
 package(default_visibility = ["//visibility:public"])
 
-load("//tools:defaults.bzl", "jasmine_node_test", "ng_module", "ng_rollup_bundle", "ts_devserver", "ts_library")
+load("//tools:defaults.bzl", "jasmine_node_test", "ng_module", "ng_rollup_bundle", "ts_library")
 load("//tools/symbol-extractor:index.bzl", "js_expected_symbol_test")
 load("@npm//http-server:index.bzl", "http_server")
 
@@ -12,7 +12,6 @@ ng_module(
     ],
     deps = [
         "//packages/core",
-        "//packages/core/test/bundling/util:reflect_metadata",
         "//packages/forms",
         "//packages/platform-browser",
     ],
@@ -27,7 +26,6 @@ ng_rollup_bundle(
     deps = [
         ":forms",
         "//packages/core",
-        "//packages/core/test/bundling/util:reflect_metadata",
         "//packages/forms",
         "//packages/platform-browser",
         "@npm//rxjs",
@@ -72,39 +70,6 @@ js_expected_symbol_test(
         "ivy-aot",
         "ivy-only",
     ],
-)
-
-genrule(
-    name = "tslib",
-    srcs = [
-        "@npm//:node_modules/tslib/tslib.js",
-    ],
-    outs = [
-        "tslib.js",
-    ],
-    cmd = "cp $< $@",
-    tags = [
-        "ivy-only",
-    ],
-)
-
-ts_devserver(
-    name = "devserver",
-    entry_module = "@angular/core/test/bundling/forms",
-    scripts = [
-        "//tools/rxjs:rxjs_umd_modules",
-    ],
-    # Use a serving_path that matches the common static index.html used
-    # in both devserver & prodserver
-    serving_path = "/bundle.min.js",
-    static_files = [
-        "index.html",
-        ":tslib",
-    ],
-    tags = [
-        "ivy-only",
-    ],
-    deps = [":forms"],
 )
 
 http_server(
--- a/packages/core/test/bundling/forms/BUILD.bazel
+++ b/packages/core/test/bundling/forms/BUILD.bazel
@@ -1,0 +1,120 @@
+package(default_visibility = ["//visibility:public"])
+
+load("//tools:defaults.bzl", "jasmine_node_test", "ng_module", "ng_rollup_bundle", "ts_devserver", "ts_library")
+load("//tools/symbol-extractor:index.bzl", "js_expected_symbol_test")
+load("@npm//http-server:index.bzl", "http_server")
+
+ng_module(
+    name = "forms",
+    srcs = ["index.ts"],
+    tags = [
+        "ivy-only",
+    ],
+    deps = [
+        "//packages/core",
+        "//packages/core/test/bundling/util:reflect_metadata",
+        "//packages/forms",
+        "//packages/platform-browser",
+    ],
+)
+
+ng_rollup_bundle(
+    name = "bundle",
+    entry_point = ":index.ts",
+    tags = [
+        "ivy-only",
+    ],
+    deps = [
+        ":forms",
+        "//packages/core",
+        "//packages/core/test/bundling/util:reflect_metadata",
+        "//packages/forms",
+        "//packages/platform-browser",
+        "@npm//rxjs",
+    ],
+)
+
+ts_library(
+    name = "test_lib",
+    testonly = True,
+    srcs = glob(["*_spec.ts"]),
+    tags = [
+        "ivy-only",
+    ],
+    deps = [
+        "//packages:types",
+        "//packages/compiler",
+        "//packages/core",
+        "//packages/core/testing",
+        "//packages/private/testing",
+    ],
+)
+
+jasmine_node_test(
+    name = "test",
+    data = [
+        ":bundle.js",
+        ":bundle.min.js",
+        ":bundle.min.js.br",
+        ":bundle.min_debug.js",
+    ],
+    tags = [
+        "ivy-only",
+    ],
+    deps = [":test_lib"],
+)
+
+js_expected_symbol_test(
+    name = "symbol_test",
+    src = ":bundle.min_debug.js",
+    golden = ":bundle.golden_symbols.json",
+    tags = [
+        "ivy-aot",
+        "ivy-only",
+    ],
+)
+
+genrule(
+    name = "tslib",
+    srcs = [
+        "@npm//:node_modules/tslib/tslib.js",
+    ],
+    outs = [
+        "tslib.js",
+    ],
+    cmd = "cp $< $@",
+    tags = [
+        "ivy-only",
+    ],
+)
+
+ts_devserver(
+    name = "devserver",
+    entry_module = "@angular/core/test/bundling/forms",
+    scripts = [
+        "//tools/rxjs:rxjs_umd_modules",
+    ],
+    # Use a serving_path that matches the common static index.html used
+    # in both devserver & prodserver
+    serving_path = "/bundle.min.js",
+    static_files = [
+        "index.html",
+        ":tslib",
+    ],
+    tags = [
+        "ivy-only",
+    ],
+    deps = [":forms"],
+)
+
+http_server(
+    name = "prodserver",
+    data = [
+        "index.html",
+        ":bundle.min.js",
+        ":bundle.min_debug.js",
+    ],
+    tags = [
+        "ivy-only",
+    ],
+)
--- a/packages/core/test/bundling/forms/forms_e2e_spec.ts
+++ b/packages/core/test/bundling/forms/forms_e2e_spec.ts
@@ -1,0 +1,67 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import '@angular/compiler';
+import {ɵwhenRendered as whenRendered} from '@angular/core';
+import {withBody} from '@angular/private/testing';
+import * as path from 'path';
+
+const PACKAGE = 'angular/packages/core/test/bundling/forms';
+const BUNDLES = ['bundle.js', 'bundle.min_debug.js', 'bundle.min.js'];
+
+describe('functional test for forms', () => {
+  BUNDLES.forEach((bundle) => {
+    describe(`using ${bundle} bundle`, () => {
+      it('should render template form', withBody('<app-root></app-root>', async () => {
+           require(path.join(PACKAGE, bundle));
+           await (window as any).waitForApp;
+
+           // Template forms
+           const templateFormsComponent = (window as any).templateFormsComponent;
+           await whenRendered(templateFormsComponent);
+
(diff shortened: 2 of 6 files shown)


devversion:
@googlebot I consent.

packages/core/test/bundling/forms/index.ts
```diff
@@ -0,0 +1,120 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import {Component, NgModule, ɵrenderComponent as renderComponent} from '@angular/core';
+import {FormArray, FormBuilder, FormControl, FormGroup, FormsModule, ReactiveFormsModule, Validators,} from '@angular/forms';
```
AndrewKushnir, line 9:
```suggestion
import {FormArray, FormBuilder, FormControl, FormGroup, FormsModule, ReactiveFormsModule, Validators} from '@angular/forms';
```

packages/core/test/bundling/forms/treeshaking_spec.ts
```diff
@@ -0,0 +1,33 @@
...
+import '@angular/compiler';
+import * as fs from 'fs';
+import * as path from 'path';
+
+const UTF8 = {
+  encoding: 'utf-8'
+};
+const PACKAGE = 'angular/packages/core/test/bundling/forms';
+
+describe('treeshaking with uglify', () => {
```
AndrewKushnir, line 18: I think we should have some simple e2e tests here with basic checks that Angular was loaded, 2 forms present on the page and may be 1-2 actions (like clicking a button/input field). See `packages/core/test/bundling/todo/todo_e2e_spec.ts` as an example.

AndrewKushnir (changes requested):

packages/core/test/bundling/forms/forms_e2e_spec.ts
```diff
@@ -0,0 +1,72 @@
...
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import '@angular/compiler';
+import {ɵwhenRendered as whenRendered} from '@angular/core';
+import {withBody} from '@angular/private/testing';
+import * as path from 'path';
+
+const UTF8 = {
```
AndrewKushnir, line 14: It looks like this is unused in this file (probably copy-n-paste'ed from a different one).

packages/core/test/bundling/forms/forms_e2e_spec.ts
```diff
@@ -0,0 +1,72 @@
...
+  encoding: 'utf-8'
+};
+const PACKAGE = 'angular/packages/core/test/bundling/forms';
+const BUNDLES = ['bundle.js', 'bundle.min_debug.js', 'bundle.min.js'];
+
+describe('functional test for forms', () => {
+  BUNDLES.forEach(bundle => {
+    describe(bundle, () => {
```
AndrewKushnir, line 22:
```suggestion
    describe(`using ${bundle} bundle`, () => {
```

packages/core/test/bundling/forms/forms_e2e_spec.ts
```diff
@@ -0,0 +1,72 @@
...
+      require(path.join(PACKAGE, bundle));
```
AndrewKushnir, line 23: Should it be in `beforeAll` call like this:
```
beforeAll(() => {
  require(path.join(PACKAGE, bundle));
});
```

packages/core/test/bundling/forms/forms_e2e_spec.ts
```diff
@@ -0,0 +1,72 @@
...
+           expect(buttons.length).toBe(1);
+           expect(buttons[0]).toBeDefined();
+
+           // Make sure button click works
+           const templateFormSpy = jasmine.createSpy(templateFormsComponent, 'addItem' as any);
+           buttons[0].click();
+           expect(templateFormSpy).toHaveBeenCalled();
+         }));
+
+      it('should render reactove form', withBody('<app-root></app-root>', async () => {
```
AndrewKushnir, line 47:
```suggestion
      it('should render reactive form', withBody('<app-root></app-root>', async () => {
```

sonukapoor:
@AndrewKushnir ci failed for some reason. I have added the `PR action: rerun CI at HEAD`

packages/core/test/bundling/forms/index.ts
```diff
@@ -0,0 +1,130 @@
...
+    </div>
+  </form>`,
+})
+class TemplateFormsComponent {
+  name = {first: 'Nancy', last: 'Drew', subscribed: true};
+  constructor() {
+    (window as any).templateFormsComponent = this;
+  }
+
+  addItem(form: NgForm) {}
```
AndrewKushnir, line 41: It's a bit unexpected that this function is empty. Could we add some basic logic to mimic what we have in Reactive Forms component below?

packages/core/test/bundling/forms/index.ts
```diff
@@ -0,0 +1,130 @@
...
+      Subscribe:
+      <input type='checkbox' formControlName='subscribed' />
+    </div>
+
+    <div>Disabled: <input formControlName='disabledInput' /></div>
+    <div
+      formArrayName='addresses'
+      *ngFor='let item of itemControls; let i = index'
+    >
+      <div [formGroupName]='i'>
```
AndrewKushnir, line 66: It looks like `*ngFor` should be placed on this `<div>`, but not the container one?

packages/core/test/bundling/forms/index.ts
```diff
@@ -0,0 +1,130 @@
...
+        <div>City: <input formControlName='city' /></div>
+      </div>
+    </div>
+    <button (click)='addItem()'>Add</button>
```
AndrewKushnir, line 70:
```suggestion
    <button (click)='addCity()'>Add City</button>
```

AndrewKushnir (changes requested):
Hi @sonukapoor, thanks for additional updates. I've left a couple comments regarding specific places that we can improve. Also some additional comments:

- I think we should match Reactive and Template-driven Form examples closer (basically make them equal, just implement using different techniques)
- Let's merge all commits into one
- In commit message (and in PR description) it'd be great to mention that this test app was created to introduce the symbol test and this app is also a foundation for additional tests/checks in the future including payload size tests (i.e. let's describe why we add this test app to a repo).

Thank you.

packages/core/test/bundling/forms/forms_e2e_spec.ts
```diff
@@ -0,0 +1,67 @@
...
+           require(path.join(PACKAGE, bundle));
+           await (window as any).waitForApp;
+
+           // Template forms
+           const templateFormsComponent = (window as any).templateFormsComponent;
+           await whenRendered(templateFormsComponent);
+
+           const templateForm = document.querySelector('app-template-forms')!;
+
+           // check for inputs
```
AndrewKushnir, line 30: nit: the comments start with lowercase letters in some cases and uppercase in others, let's use them consistently.

packages/core/test/bundling/forms/index.ts
```diff
@@ -0,0 +1,137 @@
...
+ */
+import {Component, NgModule, ɵNgModuleFactory as NgModuleFactory} from '@angular/core';
+import {FormArray, FormBuilder, FormControl, FormGroup, FormsModule, NgForm, ReactiveFormsModule, Validators} from '@angular/forms';
+import {BrowserModule, platformBrowser} from '@angular/platform-browser';
+
+@Component({
+  selector: 'app-template-forms',
+  template: `
+  <form #f='ngForm' novalidate>
+  <div ngModelGroup='profileForm' #nameCtrl='ngModelGroup'>
```
AndrewKushnir, line 16: I don't see how `#nameCtrl` and `#f` local refs are used in the template. Do we need them?

packages/core/test/bundling/forms/index.ts
```diff
@@ -0,0 +1,137 @@
...
+    <div>
+      First Name:
+      <input name='first' ngModel required #first='ngModel' />
```
AndrewKushnir, line 19: This local ref seems to be unused as well.

packages/core/test/bundling/forms/index.ts
```diff
@@ -0,0 +1,137 @@
...
+    </div>
+    <div>
+      Last Name:
+      <input name='last' ngModel />
```
AndrewKushnir, line 23: It looks like we use double quotes in templates of other test apps, may be use them for these components too?

packages/core/test/bundling/forms/index.ts
```diff
@@ -0,0 +1,137 @@
...
+
+    <button (click)='addCity()'>Add City</button>
+  </div>
+</form>`,
+})
+class TemplateFormsComponent {
+  name = {first: 'Nancy', last: 'Drew', subscribed: true};
+  addresses = [{city: 'Toronto'}];
+  constructor() {
+    (window as any).templateFormsComponent = this;
```
AndrewKushnir, line 44: I'd propose to put a quick comment here saying that we do that to have a reference to a component instance in tests.

packages/core/test/bundling/forms/index.ts
```diff
@@ -0,0 +1,137 @@
...
+
+    <div>
+      Subscribe:
+      <input type='checkbox' formControlName='subscribed' />
+    </div>
+
+    <div>Disabled: <input formControlName='disabledInput' /></div>
+    <div
+      formArrayName='addresses'
+    >
```
AndrewKushnir, line 72:
```suggestion
    <div formArrayName="addresses">
```

packages/core/test/bundling/forms/index.ts
```diff
@@ -0,0 +1,137 @@
...
+ * found in the LICENSE file at https://angular.io/license
+ */
+import {Component, NgModule, ɵNgModuleFactory as NgModuleFactory} from '@angular/core';
+import {FormArray, FormBuilder, FormControl, FormGroup, FormsModule, NgForm, ReactiveFormsModule, Validators} from '@angular/forms';
+import {BrowserModule, platformBrowser} from '@angular/platform-browser';
+
+@Component({
+  selector: 'app-template-forms',
+  template: `
+  <form novalidate>
```
AndrewKushnir, line 15: Indentation is a bit incorrect in this template (around open and close `<form>` tags)

AndrewKushnir (commented):
LGTM, just one final comment (indentation in a template). Thanks Sonu!

You'll also need to rebase, since there are new symbols that should be included into the golden file.

packages/core/test/bundling/forms/index.ts
sonukapoor: Please check now. I have checked the other files and the template usually starts in a new line. I did the same with this file.

AndrewKushnir: Thanks for the comment. Sorry for the confusion, I wanted to point out that the `<form>` and `<div>` are on the same level when `<div>` is actually a child of a `<form>`, thus it needs to be moved a couple spaces to the right (as well as the rest of the template). The `ReactiveFormsComponent` component has it right.

AndrewKushnir (approved):
LGTM, just one minor comment on template formatting.

AndrewKushnir:
FYI, I also changed the PR target to master-only, since the golden file would be slightly different between master and patch (thus this PR will cause CI errors on patch branch). I think having it only in master is OK and it will become a part of the patch branch after `10.1.0` release. Thank you.

packages/core/test/bundling/forms/index.ts
sonukapoor: Ah. My bad - It should be fixed now

sonukapoor:
@AndrewKushnir My bad about the indentation. It should be fixed now.

packages/core/test/bundling/forms/treeshaking_spec.ts
```diff
@@ -0,0 +1,33 @@
...
+  let content: string;
+  const contentPath = require.resolve(path.join(PACKAGE, 'bundle.min_debug.js'));
```
devversion, line 20: Can you add a comment here that we use the debug bundle as otherwise symbols/identifiers would be mangled (and the test would always pass).

packages/core/test/bundling/forms/BUILD.bazel
```diff
@@ -0,0 +1,120 @@
...
+    tags = [
+        "ivy-aot",
+        "ivy-only",
+    ],
+)
+
+genrule(
+    name = "tslib",
+    srcs = [
+        "@npm//:node_modules/tslib/tslib.js",
```
devversion, line 80: Is this genrule actually needed here? I don't see tslib being imported anywhere and I'd assume it's included in the bundle already.

packages/core/test/bundling/forms/BUILD.bazel
```diff
@@ -0,0 +1,120 @@
...
+ng_rollup_bundle(
+    name = "bundle",
+    entry_point = ":index.ts",
+    tags = [
+        "ivy-only",
+    ],
+    deps = [
+        ":forms",
+        "//packages/core",
+        "//packages/core/test/bundling/util:reflect_metadata",
```
devversion, line 30: Is reflect metadata here and above needed? I think since we use `ng_module` for compilation, there are no instances where the `Reflect` API is needed and this just complicates/slows-down our build graph.

packages/core/test/bundling/forms/BUILD.bazel
```diff
@@ -0,0 +1,120 @@
...
+    entry_module = "@angular/core/test/bundling/forms",
+    scripts = [
+        "//tools/rxjs:rxjs_umd_modules",
+    ],
+    # Use a serving_path that matches the common static index.html used
+    # in both devserver & prodserver
+    serving_path = "/bundle.min.js",
+    static_files = [
+        "index.html",
+        ":tslib",
```
devversion, line 102: Do we actually need the prodserver and the devserver? I don't see any protractor e2e tests here, so we could probably just remove those targets. We could keep the prodserver for testing the app in a browser.

devversion (approved):
LGTM. Just a few minor nits for the BUILD configuration

devversion (approved):
LGTM.

Issue #37847, opened 2020-06-30
Angular page - twitter and github icons are not showing in header


vesivagu:
![image](https://user-images.githubusercontent.com/25737946/86128192-ac18fe00-bafe-11ea-95a4-58badbd59aa9.png)

IE 11
windows 10

saifabusaleh:
Hi, 
Can I take this issue?

petebacondarwin:
@saifabusaleh - sure, please do have a go. Thanks

PR #38046, opened 2020-07-14
fix(docs-infra): correctly display SVG icons in IE11

saifabusaleh:
## PR Checklist

## PR Type

- [x ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: https://github.com/angular/angular/issues/37847

- [ ] Yes
- [x ] No

## Other information

This PR fixes the issue of hidden icons in IE11

The issue was that there was not enough spacing for the icons-container in IE.

Added right container to the header toolbar to contain the search bar + icons container
 added `flew-grow:1` to search input to take the empty space.

This also fixes that in mobile the icons is truncated in all browsers when width is `320px` (small screen) 
image

There is still an issue that the Github svg is not shown in IE, its because the svg of gitHub itself somehow can't be rendered by IE.

This is how it looks like now in IE:
image
 There is space for gitHub svg but its not rendered
I would be glad if someone can provide updated Github svg

### DIFF
--- a/aio/src/app/shared/custom-icon-registry.ts
+++ b/aio/src/app/shared/custom-icon-registry.ts
@@ -57,10 +57,12 @@ export class CustomIconRegistry extends MatIconRegistry {
   }
 
   private loadSvgElements(svgIcons: SvgIconInfo[]) {
-    const div = document.createElement('DIV');
     svgIcons.forEach(icon => {
       const ns = icon.namespace || DEFAULT_NS;
       const nsIconMap = this.preloadedSvgElements[ns] || (this.preloadedSvgElements[ns] = {});
+
+      // Creating a new `<div>` per icon is necessary for the SVGs to work correctly in IE11.
+      const div = document.createElement('DIV');
 
       // SECURITY: the source for the SVG icons is provided in code by trusted developers
       div.innerHTML = icon.svgSource;
--- a/aio/src/styles/1-layouts/_top-menu.scss
+++ b/aio/src/styles/1-layouts/_top-menu.scss
@@ -244,6 +244,7 @@ aio-search-box.search-container {
   .toolbar-external-icons-container {
     display: flex;
     flex-direction: row;
+    flex-shrink: 0; // This is required for the icons to be displayed correctly in IE11.
     height: 100%;
 
     a {


saifabusaleh:
Hi @gkalpak, please let me know if there something missing to push this PR forward

Thanks

gkalpak (changes requested):
Thx for starting work on this, @saifabusaleh :+1:

I am afraid this is a little more complicated than we initially expected. AFAICT, there are two issues:

1. No SVG icon other than the Twitter-logo seems to be displayed on IE 11. If you check the source, you will see that the other SVGs (such as the GitHub logo and the right arrows in the sidenav) appear empty (i.e. `<svg></svg>`). No idea why this happens.

2. The Twitter and GitHub logo `<mat-icon>` elements next to the search bar are very small (they do not respect the `width: 24px; height: 24px;` CSS styles). Again no idea why 😕

We need to figure these two issues out before we can proceed with an appropriate fix. I started looking into it, but no luck so far 😞 
If I find something I will post an update here.

##
BTW, can you please add `Fixes #37847` at the bottom of the commit message (to more closely follow our [commit message guidelines](https://github.com/angular/angular/blob/1d9e56eba06958e37ef8062ae7b65afa85f90632/CONTRIBUTING.md#commit)).

gkalpak:
I think adding `flex-shrink: 0;` to the styles for [.toolbar-external-icons-container][1] solves the second issue. Can you confirm that works for you too?
AFAICT, we can replace the changes that you currently have in this PR with that change. Did I miss something?

Now we need to figure out what's going on with the first issue 🤔 

[1]: https://github.com/angular/angular/blob/79850983e6fcedb82377f2784e6d59c62727b71e/aio/src/styles/1-layouts/_top-menu.scss#L244

gkalpak:
Regarding the second issue, it turns out that IE 11 messes up the SVGs if we use the same `<div>` elements to create them. If we move the `const div = document.createElement('DIV');` inside the `forEach()` callback, everything seems to work fine 🤷‍♂️

https://github.com/angular/angular/blob/019a696a6a5f3ef4f53253558559b713e3e596dd/aio/src/app/shared/custom-icon-registry.ts#L59-L70

##
@saifabusaleh, can you apply these two changes, verify that they work for you too (and I didn't miss something) and update the PR. Also, please add some comments above the updated lines to explain that the changes are needed for IE 11 to show the SVG correctly.

saifabusaleh:
@gkalpak great job :) I tried to apply what you posted for both of the issues and it seems to fix everything 💯 
I didn't know that icons working with `CustomIconRegistry`

should I revert commit what I did so far and do another commit with the fixes?

gkalpak:
Awesome! So, yes, if you find that my proposed changes fix the issues, I suggest you revert your original changes and apply mine (since they seem to be simpler/less disruptive). WDYT?

saifabusaleh:
@gkalpak  Looks fine for me 👍 Doing revert and pushing new changes now

saifabusaleh:
Tested on IE and looks fine, also verified that it didn't break behavior in chrome/safari 

@gkalpak I think its better also if you double check everything, and if everything is fine its possible to merge
Thanks for helping me out with this issue and pointing me to the root cause

aio/src/styles/1-layouts/_top-menu.scss
```diff
@@ -244,6 +244,7 @@ aio-search-box.search-container {
   .toolbar-external-icons-container {
     display: flex;
     flex-direction: row;
+    flex-shrink: 0;
```
gkalpak, line 247: Could you add a comment here to avoid someone accidentally removing it in the future (thinking it is not needed):

```suggestion
    flex-shrink: 0;  // This is required for the icons to be displayed correctly in IE11.
```

aio/src/app/shared/custom-icon-registry.ts
```diff
@@ -57,11 +57,13 @@ export class CustomIconRegistry extends MatIconRegistry {
   }
 
   private loadSvgElements(svgIcons: SvgIconInfo[]) {
-    const div = document.createElement('DIV');
     svgIcons.forEach(icon => {
       const ns = icon.namespace || DEFAULT_NS;
       const nsIconMap = this.preloadedSvgElements[ns] || (this.preloadedSvgElements[ns] = {});
 
+      // this change is fix for IE11 to show the SVG correctly
```
gkalpak, line 64: It is not clear what the change is by looking at the comment. How about making more clear (without having to look at the code histiry) with something like:

```suggestion
      // Creating a new `<div>` per icon is necessary for the SVGs to work correctly in IE11.
```

gkalpak (changes requested):
Thx, @saifabusaleh! It seems to work fine for me too.

I have left a couple of minor suggestions. OTher than that lgtm :+1:

Can you also squash all commits into one and update the commit message to mention both fixes. For example:

> fix(docs-infra): correctly display SVG icons in IE11
>
> Fix two issues that affected displaying of SVG icons in IE11:
> 1. All SVG icons except for one appeared empty. This was related how the
>    `CustomIconRegistry` re-used the same `<div>` element to create all
>    SVG elements.
> 2. The GitHub and Twitter buttons next to the search bar were not sized
>    properly. 
>
> Fixes #37847

saifabusaleh:
@gkalpak Done :)

Thanks for all the feedback

aio/src/app/shared/custom-icon-registry.ts
```diff
@@ -57,11 +57,13 @@ export class CustomIconRegistry extends MatIconRegistry {
   }
 
   private loadSvgElements(svgIcons: SvgIconInfo[]) {
-    const div = document.createElement('DIV');
     svgIcons.forEach(icon => {
```
gkalpak, line 60: 💡 Idea: We can lazy-create SVGs when requested, instead of creating them up-front.
Something you can do in a future PR, if interested :wink:

gkalpak (approved):
LGTM ✨

One last thing: Please update the commit message header as suggested in https://github.com/angular/angular/pull/38046#pullrequestreview-450500631 :pray:

aio/src/app/shared/custom-icon-registry.ts
saifabusaleh: Not sure how it can be done, also I don't think it will change much since there there is only 6 SVG's for now

+ Updated the commit message header

gkalpak: Yes it wouldn't have a dramatic impact indeed. At the same time there is no downside in doing it and it would save some unnecessary work.

Thx for updating the header :tada:

PR #38053, opened 2020-07-14
docs: add routing terms to glossary

jbogarthyde:
Update glossary to add term definitions for routing; componentless route, link parameters array, router outlet.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Terms used in Routing & Navigation guide missing from glossary

Issue Number: N/A

## What is the new behavior?
Adds important routing terms to glossary.

- [ ] Yes
- [x] No

### DIFF
--- a/aio/content/guide/glossary.md
+++ b/aio/content/guide/glossary.md
@@ -210,6 +210,7 @@ An Angular component class is responsible for exposing data and handling most of
 
 Read more about component classes, templates, and views in [Introduction to Angular concepts](guide/architecture).
 
+
 ## configuration
 
 See  [workspace configuration](#cli-config)
@@ -578,7 +579,6 @@ Angular calls these hook methods in the following order:
 
 To learn more, see [Lifecycle Hooks](guide/lifecycle-hooks).
 
-
 {@a M}
 
 {@a module}
@@ -738,6 +738,26 @@ When using reactive forms:
 * The associated Angular directives are prefixed with `form`, such as `formControl`, `formGroup`, and `formControlName`.
 
 The alternative is a template-driven form. For an introduction and comparison of both forms approaches, see [Introduction to Angular Forms](guide/forms-overview).
+
+{@a resolver}
+
+## resolver
+
+A class that implements the [Resolve](api/router/Resolve "API reference") interface (or a function with the same signature as the [resolve() method](api/router/Resolve#resolve "API reference")) that you use to produce or retrieve data that is needed before navigation to a requested route can be completed.
+
+Resolvers run after all [route guards](#route-guard "Definition") for a route tree have been executed and have succeeded.
+
+See an example of using a [resolve guard](guide/router-tutorial-toh#resolve-guard "Routing techniques tutorial") to retrieve dynamic data.
+
+{@a route-guard}
+
+## route guard
+
+A method that controls navigation to a requested route in a routing application.
+Guard functions that use this method signature can determine whether a route can be activated or deactivated, and whether a lazy-loaded module can be loaded.
+
+Learn more in the [Routing and Navigation](guide/router#preventing-unauthorized-access "Examples") guide.
+
 
 {@a router}
 {@a router-module}
--- a/aio/content/guide/glossary.md
+++ b/aio/content/guide/glossary.md
@@ -754,7 +754,7 @@ See an example of using a [resolve guard](guide/router-tutorial-toh#resolve-guar
 ## route guard
 
 A method that controls navigation to a requested route in a routing application.
-Guard functions that use this method signature can determine whether a route can be activated or deactivated, and whether a lazy-loaded module can be loaded.
+Guards determine whether a route can be activated or deactivated, and whether a lazy-loaded module can be loaded.
 
 Learn more in the [Routing and Navigation](guide/router#preventing-unauthorized-access "Examples") guide.
 


aio/content/guide/glossary.md
```diff
@@ -766,6 +790,12 @@ An Angular [component](#component) with a `RouterOutlet` directive in its templa
 
 For more information, see [Routing and Navigation](guide/router).
 
+{@a routing-module}
+
+## routing module
+
+An NgModule that contains an application's routing configuration. A dedicated routing module does not declare components, but serves to separate routing concerns from other application concerns. It provides a well-known location for routing service providers such as guards and resolvers. A routing module is also easy to replace or remove when testing the application.
+
```
jelbourn, line 798: I'm not sure we should add routing module because we plan to eliminate this concept

aio/content/guide/glossary.md
```diff
@@ -210,6 +210,14 @@ An Angular component class is responsible for exposing data and handling most of
...
 Read more about component classes, templates, and views in [Introduction to Angular concepts](guide/architecture).
 
+{@a componentless-route}
+
+## componentless route
+
+In configuring the [Angular router](#router), a container `Route` for a set of related child routes that is not itself associated with a component.
+
+Read more about nested routes in the [Routing and Navigation](guide/router) guide.
+
```
jelbourn, line 220: I'm not sure this term is widely used enough to warrant being in the big glossary.

aio/content/guide/glossary.md
```diff
@@ -578,6 +586,14 @@ Angular calls these hook methods in the following order:
...
 To learn more, see [Lifecycle Hooks](guide/lifecycle-hooks).
 
+{@a link-parameters-array}
+
+## link parameters array
+
+In the Angular [router service](#router), an array which contains an object that associates a route path string with one or more required or optional query parameters. The router service resolves that array into a complete URL.
+
+To learn more, see the [Routing and Navigation](guide/router#link-parameters-array) guide.
+
```
jelbourn, line 596: I'm not sure this term is widely used enough to warrant being in the big glossary.

jelbourn (commented):
@mgechev I would be interested in getting your feedback here

aio/content/guide/glossary.md
jbogarthyde: I'll move this definition to the Router Guide, where the term is used a lot.

aio/content/guide/glossary.md
jbogarthyde: Agreed - it can be defined where used in the Router guide.

aio/content/guide/glossary.md
jbogarthyde: This is a little problematic - currently this app structure is strongly recommended in all existing tutorials and guides, and a routing module is created by the CLI when you generate a  new app with `ng new my-app --routing` (which the command prompts for).

aio/content/guide/glossary.md
```diff
@@ -739,6 +739,14 @@ When using reactive forms:
 
 The alternative is a template-driven form. For an introduction and comparison of both forms approaches, see [Introduction to Angular Forms](guide/forms-overview).
 
+{@a route-guard}
+
+## route guard
+
+A function that controls access to a view in a routing application.
```
atscott, line 746: Guards aren't _necessarily_ for views. They guard against activation of a `Route` which may or may not contain a component. For example
```
{
  path: 'something'
  children: [
     ...
   ],
   canActivate: [MyGuard],
}
```

Additionally, I think `canLoad` is also referred to as a "guard" which guards against the loading of a lazy module (replace `canActivate` with `canLoad` in the above example).

Lastly, it might be more accurate to say "to or from" because of `canDeactivate`.

Maybe "A function that controls access to or from a particular `Route` config in a routing application."? This is hard...guards do a lot of things and there's some nuance to them.

aio/content/guide/glossary.md
atscott: I don't know that we have to specifically coin the term. In fact, it sort of only makes sense in context because on its own it's a little confusing since we have `RouterModule` as public API as well.

jbogarthyde: I'll remove this from the glossary and try to talk about it in a more general way in the Router guide.

aio/content/guide/glossary.md
```diff
@@ -739,6 +739,23 @@ When using reactive forms:
 
 The alternative is a template-driven form. For an introduction and comparison of both forms approaches, see [Introduction to Angular Forms](guide/forms-overview).
 
+{@a resolver}
+
+## resolver
+
+In the `Router` service, a function that implements the [resolve() method](api/router/Resolve "API reference") to resolve data needed before navigation to a requested route can be completed.
```
mgechev, line 746: I'd say that a resolver is a service which implements the `api/router/Resolve` interface, to resolve data needed before navigation to a requested route to be completed.

The first part seem redundant.

aio/content/guide/glossary.md
```diff
@@ -739,6 +739,23 @@ When using reactive forms:
...
+Resolvers run after all [route guards](#route-guard "Definition") for a route tree have been executed and have succeeded.
+
+{@a route-guard}
+
+## route guard
+
+A function that controls navigation to a requested route in a routing application.
```
mgechev, line 753: The guard is a class.

mgechev (changes requested):
Left few comments.

aio/content/guide/glossary.md
petebacondarwin, line 746:
```suggestion
In the context of the `Router` service, a function that implements the [resolve() method](api/router/Resolve "API reference") to resolve data needed before navigation to a requested route can be completed.
```

petebacondarwin (approved):
A small suggestion to clarify, since as it stands it seems like the `resolve()` function appears on the `Router` service.

aio/content/guide/glossary.md
petebacondarwin: Strictly a resolver can be a free standing function that has the same signature as the `resolve()` method of `Resolve`.
I also wondered about changing the wording. But since this is a glossary and it links to the `Resolve` interface that does a more detailed job of describing it, with examples, I opted for leaving it as-is.

aio/content/guide/glossary.md
petebacondarwin: Strictly, it is a method on a class - since each of the guard interfaces only contain a single method which is to be implemented. I also considered asking for this change, but on further thought, for the glossary it seems about right to call it a function that is called. Perhaps a method would be more accurate?

petebacondarwin (approved):

atscott (approved):

petebacondarwin (approved):
Reviewed-for: public-api

aio/content/guide/glossary.md
mgechev: >Strictly a resolver can be a free standing function that has the same signature as the resolve() method of Resolve.

Maybe we can change it to "A class which implements the `Resolve` interface or a function with the same signature as the [resolve() method](api/router/Resolve "API reference") used to get data needed before navigation to a requested route can be completed.

To me it seems inaccurate saying that a *function* implements a *method*.

My second comment here was related to saying "In the `Router` service". This brings some extra context that I like, but it makes assumptions about the implementation. I'd think saying: "in a route declaration", would bring better clarity.

aio/content/guide/glossary.md
mgechev: Yes, method seems more accurate.

jbogarthyde: How about this?
 "A method that controls navigation to a requested route in a routing application.
Guard functions that use this method signature can determine whether a route can be activated or deactivated, and whether a lazy-loaded module can be loaded."

jbogarthyde: Or second sentence: "You can implement a guard function with the appropriate signature to determine..."

aio/content/guide/glossary.md
```diff
@@ -739,6 +739,26 @@ When using reactive forms:
...
+Resolvers run after all [route guards](#route-guard "Definition") for a route tree have been executed and have succeeded.
+
+See an example of using a [resolve guard](guide/router-tutorial-toh#resolve-guard "Routing techniques tutorial") to retrieve dynamic data.
+
+{@a route-guard}
+
+## route guard
+
+A method that controls navigation to a requested route in a routing application.
+Guard functions that use this method signature can determine whether a route can be activated or deactivated, and whether a lazy-loaded module can be loaded.
```
mgechev, line 757: Can we simplify the second part to:

>Guards determine whether a round can be activated or deactivated, and whether a lazy-loaded module can be loaded.

mgechev (approved):

AndrewKushnir:
Hi @jbogarthyde, it looks like this PR is ready (it has the "merge" label), but the updated file belongs to the `public-api` reviewers group and requires 3 approvals (currently it's just Pete's approval - Minko, Andrew and myself are not in that group). Given the fact that it's a docs-only change, it should be safe to proceed, but I think it'd be great if @jelbourn gives its final approval and I can merge it after that. Thank you.

jelbourn (approved):
LGTM

Reviewed-for: public-api

PR #38081, opened 2020-07-15
refactor(dev-infra): allow use-case and confusing types to be marked as triaged

atscott:
Some issue reports don't really fall into any of the current buckets that count
towards triage level 2: bug/fix, feature, or refactor. Some reports are:
* working as intended but confusing - the labels might be 'type: confusing', 'comp: docs', 'comp: router'
* generally working as originally designed but a use-case could be argued for a different implementation.
 This type of report is a little hard to triage; it may be neither a bug, nor feature but more of a
 'type: use-case'. These may eventually turn into a bug/fix or feature, but can't necessarily be
 put in those buckets immediately.

### DIFF
--- a/.github/angular-robot.yml
+++ b/.github/angular-robot.yml
@@ -156,11 +156,9 @@ triage:
       - "comp: *"
     -
       - "type: confusing"
-      - "freq*"
       - "comp: *"
     -
       - "type: use-case"
-      - "freq*"
       - "comp: *"
 
 # options for the triage PR plugin
--- a/.github/angular-robot.yml
+++ b/.github/angular-robot.yml
@@ -154,6 +154,14 @@ triage:
     -
       - "type: RFC / Discussion / question"
       - "comp: *"
+    -
+      - "type: confusing"
+      - "freq*"
+      - "comp: *"
+    -
+      - "type: use-case"
+      - "freq*"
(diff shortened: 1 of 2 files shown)


.github/angular-robot.yml
```diff
@@ -154,6 +154,14 @@ triage:
     -
       - "type: RFC / Discussion / question"
       - "comp: *"
+    -
+      - "type: confusing"
+      - "freq*"
```
AndrewKushnir, line 159: I'd propose to avoid treating the `freq` label as required (similar to `refactor` or `feature`).

AndrewKushnir (approved):

pkozlowski-opensource (approved):
Yay! LGTM, thnx for sending this PR (I should have done it before).
Since we are at it - could you please also update https://github.com/angular/angular/blob/master/docs/TRIAGE_AND_LABELS.md to reflect those new types (hence adding the cleanup label)

atscott:
> Since we are at it - could you please also update https://github.com/angular/angular/blob/master/docs/TRIAGE_AND_LABELS.md to reflect those new types (hence adding the cleanup label)

👍  Done - good addition. I didn't even know that doc existed 😄

josephperrott (approved):
LGTM

PR #38082, opened 2020-07-15
i18n: check duplicate messages

### LINKED ISSUE #38077: add an option for xi18n tool to detect Inconsistency for ID content
# 🚀 Feature request

- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [x ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Description
With angular i18n, If the same message ID is used in several different places in the source code , but contains different messages, the extracted.xlf contains only one occurence of the ID, one source message (the first one found in sources) and the different locations where the message have been found.

But sometimes during developement process, for different reasons (copy/paste, modification of the message in one source and not in others, etc ....) the same ID, does not contain the same message, and it generates bugs because the messsage will be replace by the only one translation.

In one angular application the same i18n ID used in different places MUST refer the same message. If not, it means there is a problem to be fixed.

example:
`<button mat-menu-itemi18n="@@MY_TEXT_LABEL"(click)="openAuthenticationConfigurationModal()">
  The first Text
</button>
`
 
and somewhere else in the code
`<app-readonly-field i18n-label="@@MY_TEXT_LABEL" label="Another Text">{{getLabelFor()}}</app-readonly-field>`

To avoid this problem, it would be nice to have an option for xi18n tool to detect if a message ID , used several times, refers the same message or not.

something like :
--checkIdContentConsistency
--checkIdSourceConsistency

The xi18n tool has all information to do this check easily

I've thought to develop some scripts to do, but no easy to implement a angular code parser to extract all messages from code

petebacondarwin:
Previously, the i18n message extractor just quietly ignored messages that
it extracted that had the same id. It can be helpful to identify these
to track down messages that have the same id but different message text.

Now the messages are checked for duplicate ids with different message text.
Any that are found can be reported based on the new `--duplicateMessage`
command line option (or `duplicateMessage` API options property).

* "ignore" - no action is taken
* "warning" - a diagnostic warning is written to the logger
* "error" - the extractor throws an error and exits

Fixes #38077

### DIFF
--- a/packages/localize/src/tools/src/extract/duplicates.ts
+++ b/packages/localize/src/tools/src/extract/duplicates.ts
@@ -1,0 +1,58 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import {AbsoluteFsPath, FileSystem} from '@angular/compiler-cli/src/ngtsc/file_system';
+import {ɵMessageId, ɵParsedMessage} from '@angular/localize';
+
+import {DiagnosticHandlingStrategy, Diagnostics} from '../diagnostics';
+import {serializeLocationPosition} from '../source_file_utils';
+
+/**
+ * Check each of the given `messages` to find those that have the same id but different message
+ * text. Add diagnostics messages for each of these duplicate messages to the given `diagnostics`
+ * object (as necessary).
+ */
+export function checkDuplicateMessages(
+    fs: FileSystem, messages: ɵParsedMessage[],
+    duplicateMessageHandling: DiagnosticHandlingStrategy, basePath: AbsoluteFsPath): Diagnostics {
+  const diagnostics = new Diagnostics();
+  if (duplicateMessageHandling === 'ignore') return diagnostics;
+
+  const messageMap = new Map<ɵMessageId, ɵParsedMessage[]>();
+  for (const message of messages) {
+    if (messageMap.has(message.id)) {
+      messageMap.get(message.id)!.push(message);
+    } else {
+      messageMap.set(message.id, [message]);
+    }
+  }
+
+  for (const duplicates of messageMap.values()) {
+    if (duplicates.length <= 1) continue;
+    if (duplicates.every(message => message.text === duplicates[0].text)) continue;
+
+    const diagnosticMessage = `Duplicate messages with id "${duplicates[0].id}":\n` +
+        duplicates.map(message => serializeMessage(fs, basePath, message)).join('\n');
+    diagnostics.add(duplicateMessageHandling, diagnosticMessage);
+  }
+
+  return diagnostics;
+}
+
+/**
+ * Serialize the given `message` object into a string.
+ */
+function serializeMessage(
+    fs: FileSystem, basePath: AbsoluteFsPath, message: ɵParsedMessage): string {
+  if (message.location === undefined) {
+    return `   - "${message.text}"`;
+  } else {
+    const locationFile = fs.relative(basePath, message.location.file);
+    const locationPosition = serializeLocationPosition(message.location);
+    return `   - "${message.text}" : ${locationFile}:${locationPosition}`;
+  }
+}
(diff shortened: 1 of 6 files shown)


packages/localize/src/tools/src/extract/main.ts
```diff
@@ -11,6 +11,10 @@ import {ConsoleLogger, Logger, LogLevel} from '@angular/compiler-cli/src/ngtsc/l
 import {ɵParsedMessage} from '@angular/localize';
 import * as glob from 'glob';
 import * as yargs from 'yargs';
+
+import {DiagnosticHandlingStrategy, Diagnostics} from '../diagnostics';
+
+import {checkDuplicateMessages} from './duplicates';
```
AndrewKushnir, line 17: @petebacondarwin it looks like the `duplicates.ts` file is missing...

petebacondarwin: Doh!

petebacondarwin: Should be there now :-)

packages/localize/src/tools/src/extract/duplicates.ts
```diff
@@ -0,0 +1,58 @@
...
+                    `   - "${message.text}" ${serializeLocation(fs, basePath, message.location)}`)
+            .join('\n');
+    diagnostics.add(duplicateHandling, diagnosticMessage);
+  }
+}
+
+/**
+ * Serialize the given `location` object into a string.
+ */
+function serializeLocation(
```
AndrewKushnir, line 49: nit: may be we can move this function to some util file (it looks quite generic), so it can be reused later?

packages/localize/src/tools/src/extract/main.ts
```diff
@@ -90,6 +101,7 @@ if (require.main === module) {
     logger,
     useSourceMaps: options['useSourceMaps'],
     useLegacyIds: options['useLegacyIds'],
+    duplicateMessage,
```
AndrewKushnir, line 104: nit: I'd propose renaming it to something like `duplicateMessageHandlingStrategy` (or `duplicateHandling` how it's called in `checkDuplicateMessages` function) to provide more context on what this parameter is responsible for.

packages/localize/src/tools/src/extract/main.ts
```diff
@@ -139,22 +155,34 @@ export function extractTranslations({
...
+  const basePath = fs.resolve(rootPath);
+  const extractor = new MessageExtractor(fs, logger, {basePath, useSourceMaps});
 
   const messages: ɵParsedMessage[] = [];
   for (const file of sourceFilePaths) {
     messages.push(...extractor.extractMessages(file));
   }
 
+  const diagnostics = new Diagnostics();
+  checkDuplicateMessages(fs, messages, diagnostics, duplicateMessage, basePath);
```
AndrewKushnir, line 171: nit: would it make sense for the `checkDuplicateMessages` to return `Diagnostics` instance instead of passing it as an argument?

AndrewKushnir (approved):
LGTM, thanks for adding these checks @petebacondarwin 👍 

I've left a few minor comments, please have a look when you get a chance.

Thank you.

packages/localize/src/tools/src/extract/duplicates.ts
petebacondarwin: Part of it is generic - the bit that computes the position string. This is used in the translation file serializers. But each of them serializes the overall location slightly differently. I'll extract out the position serialization.

packages/localize/src/tools/src/extract/main.ts
petebacondarwin: Yeah, I was thinking about that. The reason I went with this is that we have basically the same option in the translator: https://github.com/angular/angular/blob/master/packages/localize/src/tools/src/translate/main.ts#L113

Happy to change it given that you brought it up

packages/localize/src/tools/src/extract/main.ts
petebacondarwin: I did that initially. But then I thought that in the future we might like to add diagnostics messages from things like the translation file serializers (or the message extractors), in which case passing round a single diagnostics object to share seemed like the best option. This approach is also used in the translator. What do you think?

petebacondarwin: But I guess that `Diagnostics` does have a `merge()` method.

petebacondarwin:
Thanks for the review (and approval) @AndrewKushnir 
I resolved all your NITs, and since the commit message needed changing anyway, I just amended rather than fixing up a new commit.

AndrewKushnir:
Thanks for the updates @petebacondarwin 👍

PR #38093, opened 2020-07-16
build: fix symbol extractor not dealing with ES2015 classes

devversion:
We recently reworked our `ng_rollup_bundle` rule to no longer output
ESM5 and to optimize applications properly (previously applications were
not optimized properly due to incorrect build optimizer setup).

This change meant that a lot of symbols have been removed from the
golden correctly. See: fd65958b887f6ea8dd5235e6de1d533e4c578602 (due to the app now being optimized properly)

Unfortunately though, a few symbols have been accidentally removed
because they are now part of the bundle as ES2015 classes which the
symbol extractor does not pick up. This commit fixes the symbol
extractor to capture ES2015 classes. We also update the golden to
reflect this change.

### DIFF
--- a/packages/core/test/bundling/cyclic_import/bundle.golden_symbols.json
+++ b/packages/core/test/bundling/cyclic_import/bundle.golden_symbols.json
@@ -36,6 +36,12 @@
     "name": "NO_CHANGE"
   },
   {
+    "name": "NodeInjectorFactory"
+  },
+  {
+    "name": "SimpleChange"
+  },
+  {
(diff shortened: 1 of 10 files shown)


sonukapoor (approved):

AndrewKushnir (approved):
LGTM 👍 Thanks for the fix @devversion!

devversion:
Caretaker Note: Adding merge assistance because this PR has all required reviews, but due to a bug in Github we cannot remove the superfluous rrview request.

PR #38115, opened 2020-07-17
perf(router): don't preload already loaded route configs

dimakuba:
For now preloading process is running every `NavigationEnd` event,
even route configs are loaded, that increases scripting time.
Would be better to preload only route configs that are not loaded yet.
These changes allow to preload router configs either:
first `NavigationEnd` event & every router config changes via `Router#resetConfig`.

### DIFF
--- a/packages/router/src/router_preloader.ts
+++ b/packages/router/src/router_preloader.ts
@@ -8,7 +8,7 @@
 
 import {Compiler, Injectable, Injector, NgModuleFactoryLoader, NgModuleRef, OnDestroy} from '@angular/core';
 import {from, Observable, of, Subscription} from 'rxjs';
-import {catchError, concatMap, filter, map, mergeAll, mergeMap} from 'rxjs/operators';
+import {catchError, concatMap, filter, map, mergeAll, mergeMap, scan} from 'rxjs/operators';
 
 import {LoadedRouterConfig, Route, Routes} from './config';
 import {Event, NavigationEnd, RouteConfigLoadEnd, RouteConfigLoadStart} from './events';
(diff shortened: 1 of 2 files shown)


atscott:
Hi @dimakuba, this is a good thought for optimization. I'm wondering what might happen in some of the following cases though:
* The preload fails. With the current implementation, it would be retried on next `NavigationEnd` but with this new implementation, it would never be retried
* A custom preload strategy that conditionally preloads modules based on some state in the app. This change would break those preloading strategies.

I think this may be better to handle in a custom preloading strategy rather than in the router itself. For example
```
export class MyPreloadStrategy implements PreloadingStrategy {
  preloaded: false;
  preload(route: Route, fn: () => Observable<any>): Observable<any> {
    if (this.preloaded) return of(null);

    return fn().pipe(
        tap(() => this.preloaded = true), 
        catchError(() => of(null))
    );
  }
}
```

There is a separate issue to not load modules that are already loaded here #22842. The associated PR #36760 may be a more appropriate way to address this.

franklinpious:
Hi @atscott , In the solution you provided, the preloaded variable will be true on custom prefetching a lazy loaded module and doesnt prefetch other lazy loaded modules. Right ?

franklinpious:
Hi @atscott  , In the solution you provided, the preloaded variable will be true on custom prefetching a lazy loaded module and doesnt prefetch other lazy loaded modules. Right ?

atscott:
> Hi @atscott , In the solution you provided, the preloaded variable will be true on custom prefetching a lazy loaded module and doesnt prefetch other lazy loaded modules. Right ?

I don't completely understand the question, but what I mentioned was not meant to be a complete solution, just an example of how to handle not preloading already loaded configs inside a `PreloadingStrategy`.

PR #38120, opened 2020-07-17
fix(language-service): [ivy] do not retrieve ts.Program at startup

kyliau:
This commit removes compiler instantiation at startup.
This is because the constructor is invoked during the plugin loading phase,
in which the project has not been completely loaded.

Retrieving `ts.Program` at startup will trigger an `updateGraph` operation,
which could only be called after the Project has loaded completely.
Without this change, the Ivy LS will throw when the tsserver proxy is created.

Note that the whole `Compiler` class is temporary, so changes made there are
only for development. Once we have proper integration with ngtsc the
`Compiler` class would be removed.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/language-service/ivy/compiler/compiler.ts
+++ b/packages/language-service/ivy/compiler/compiler.ts
@@ -17,44 +17,40 @@ import * as ts from 'typescript/lib/tsserverlibrary';
 
 import {makeCompilerHostFromProject} from './compiler_host';
 
+interface AnalysisResult {
+  compiler: NgCompiler;
+  program: ts.Program;
+}
+
 export class Compiler {
   private tsCompilerHost: ts.CompilerHost;
-  private lastKnownProgram: ts.Program;
-  private compiler: NgCompiler;
+  private lastKnownProgram: ts.Program|null = null;
   private readonly strategy: TypeCheckingProgramStrategy;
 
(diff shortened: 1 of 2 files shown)


ayazhafiz:
@kyliau should I wait to review until after this is no longer a draft?

kyliau:
@ayazhafiz this is ready for review, thank you very much!

packages/language-service/ivy/compiler/compiler.ts
```diff
@@ -16,10 +16,14 @@ import * as ts from 'typescript/lib/tsserverlibrary';
...
+}
+
 export class Compiler {
   private tsCompilerHost: ts.CompilerHost;
-  private lastKnownProgram: ts.Program;
-  private compiler: NgCompiler;
+  private lastKnownProgram: ts.Program|null = null;
   private readonly strategy: TypeCheckingProgramStrategy;
 
   constructor(private readonly project: ts.server.Project, private options: CompilerOptions) {
```
ayazhafiz, line 29: Can you call out here that the project is not completely loaded, and that is why we don't create a compiler? It's very unexpected since we're accepting a project parameter...

ayazhafiz (approved):
+me after comment is resolved

PR #38085, opened 2020-07-15
Edit and organize router guide

jbogarthyde:
Adds conceptual material from tutorial into guide, edits content for structure, style, and clarity, removing redundancy.

## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [ ] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [x] Documentation content changes
- [ ] angular.io application / infrastructure changes
- [ ] Other... Please describe:

## What is the current behavior?
Router guide does not conform to current standards of style and terminology, and does not present concepts logically.

Issue Number: N/A

## What is the new behavior?
Edited for style and structure, concept sections added for topics that were introduced and discussed only in the tutorial.

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

Issue #25323, opened 2018-08-06
BrowserPlatformLocation pushState and replaceState use incorrect origin


CarstenLeue:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead see https://github.com/angular/angular/blob/master/CONTRIBUTING.md#question
[ ] Other... Please describe:
</code></pre>

## Current behavior
Consider a setup in which the base URL of the index.html is different from the location  that the document is accessed from (see example below). 

Angular calls the pushState and replaceState during routing with a server relative URL (i.e. starting with a '/' but without host or domain (i.e. no origin)).
In this case the browser's implementation of history.pushState and history.replaceState will construct an absolute URL using the domain of the base tag. In this usecase  this does not match the window's domain, so the operation fails.

## Expected behavior

Angular should use absolute URLs for pushState and replaceState and should use the window origin, not the document origin. This is because the navigation happens on the window URL, not the document URL.

## Minimal reproduction of the problem with instructions

Configure your index.html like this:

```html
<!doctype html>
<html lang="en">

<head>
  <!-- my CDN --> 
   <base href="http://my.cdn.org/myapp/">
<body>
...
</body>
</html>
```

Expose the 'APP_BASE_HREF' like this:

```typescript
@NgModule({
  declarations: [AppComponent, TestComponent],
  imports: [BrowserModule, AppRoutingModule],
  providers: [
    {
      provide: APP_BASE_HREF,
      useValue: '/my-context-root/'
    }
  ],
  bootstrap: [AppComponent]
})
export class AppModule {}
```

Access the application like this:

```
https://my-domain.com/my-context-root/
```

## What is the motivation / use case for changing the behavior?

Motivation:
- suppose you have a custom domain 'https://my-domain.com' but no efficient CDN in front of your domain
- I would like to host the "heavy" application binaries on a CDN (which does not have my desired hostname) but use my own domain for navigation purposes

- so I'd like to point the <base> url to my CDN's entry point
- expose the context root of my app via the APP_BASE_URL token
- Angular should just preserve the origin that I used to navigate to the app

## Environment

<pre><code>
Angular version: 6.1.1

Browser:
- [x] Chrome (desktop) version XX
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
[...]

PR #38123, opened 2020-07-17
docs(router): clarify how base href is used to construct targets

atscott:
The documentation is not clear on how the base href and APP_BASE_HREF are used. This commit
should help clarify more complicated use-cases beyond the most common one of just a '/'

### DIFF
--- a/aio/content/guide/router.md
+++ b/aio/content/guide/router.md
@@ -557,6 +557,7 @@ Those developers may still use HTML5 URLs by taking the following two steps:
 1. Provide the router with an appropriate `APP_BASE_HREF` value.
 1. Use root URLs for all web resources: CSS, images, scripts, and template HTML files.
 
+A URL can have the following parts:
 ```
 foo://example.com:8042/over/there?name=ferret#nose
 \_/   \______________/\_________/ \_________/ \__/
@@ -565,8 +566,9 @@ scheme    authority      path        query   fragment
 ```
 
 * The `<base href>` path should end with a "/", as browsers ignore characters in the path that follow the right-most "/".
-* The query part in the `<base href>` is only used if the path in the page URI is empty and also has not query. This means the query from the base will only be included when using `HashLocationStrategy`.
-* If a URI has an `authority` part, the authority, path, query, and fragment are taken from the URI. In this way, an `APP_BASE_HREF` with an authority can override the `<base href>` value.
+* If the `<base href>` path includes a query part, that part is only used if the page-URL path is empty and has no query.
+This means that a query in the base is only included when using `HashLocationStrategy`.
+* If a URL has an `authority` part, the authority, path, query, and fragment are taken from the URL. In this way, an `APP_BASE_HREF` with an authority can override the `<base href>` value.
 * A fragment in the `<base href>` is _never_ persisted.
 
 For more complete information on how `<base href>` is used to construct target URIs, see the [RFC](https://tools.ietf.org/html/rfc3986#section-5.2.2) section on transforming references.
--- a/aio/content/guide/router.md
+++ b/aio/content/guide/router.md
@@ -567,7 +567,7 @@ scheme    authority      path        query   fragment
 
 * The `<base href>` path should end with a "/", as browsers ignore characters in the path that follow the right-most "/".
 * If the `<base href>` path includes a query part, that part is only used if the page-URL path is empty and has no query.
-This means that a query in the base is only included when using `HashLocationStrategy`.
+This means that a query in the `<base href>` is only included when using `HashLocationStrategy`.
 * If a URL has an `authority` part, the authority, path, query, and fragment are taken from the URL. In this way, an `APP_BASE_HREF` with an authority can override the `<base href>` value.
 * A fragment in the `<base href>` is _never_ persisted.
 
--- a/aio/content/guide/router.md
+++ b/aio/content/guide/router.md
@@ -554,8 +554,22 @@ Some developers may not be able to add the `<base>` element, perhaps because the
 
 Those developers may still use HTML5 URLs by taking the following two steps:
 
-1. Provide the router with an appropriate [APP_BASE_HREF][] value.
+1. Provide the router with an appropriate `APP_BASE_HREF` value.
 1. Use root URLs for all web resources: CSS, images, scripts, and template HTML files.
+
+```
+foo://example.com:8042/over/there?name=ferret#nose
+\_/   \______________/\_________/ \_________/ \__/
+ |           |            |            |        |
+scheme    authority      path        query   fragment
+```
+
+* The `<base href>` path should end with a "/", as browsers ignore characters in the path that follow the right-most "/".
+* The query part in the `<base href>` is only used if the path in the page URI is empty and also has not query. This means the query from the base will only be included when using `HashLocationStrategy`.
+* If a URI has an `authority` part, the authority, path, query, and fragment are taken from the URI. In this way, an `APP_BASE_HREF` with an authority can override the `<base href>` value.
+* A fragment in the `<base href>` is _never_ persisted.
+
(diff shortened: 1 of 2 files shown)


aio/content/guide/router.md
```diff
@@ -554,9 +554,23 @@ Some developers may not be able to add the `<base>` element, perhaps because the
...
+```
+foo://example.com:8042/over/there?name=ferret#nose
+\_/   \______________/\_________/ \_________/ \__/
+ |           |            |            |        |
+scheme    authority      path        query   fragment
+```
+
+* The `<base href>` path should end with a "/", as browsers ignore characters in the path that follow the right-most "/".
+* The query part in the `<base href>` is only used if the path in the page URI is empty and also has not query. This means the query from the base will only be included when using `HashLocationStrategy`.
+* If a URI has an `authority` part, the authority, path, query, and fragment are taken from the URI. In this way, an `APP_BASE_HREF` with an authority can override the `<base href>` value.
```
jbogarthyde, line 569: Is a distinction being made between "URI" and "URL"? If not, we should standardize on URI, as used in the RFC.

jbogarthyde (changes requested):
Please hold this until https://github.com/angular/angular/pull/38085 lands -- unless you would like me to add this to that PR.

aio/content/guide/router.md
atscott: URL is a more specific URI. I've changed these all to URL since it seems that's what we mostly use in the docs.
https://danielmiessler.com/study/difference-between-uri-url/#:~:text=Here%20are%20the%20differences.,as%20HTTPs%20%2C%20FTP%20%2C%20etc.

jbogarthyde (approved):
LGTM.

packages/common/src/location/location_strategy.ts
```diff
@@ -81,17 +81,21 @@ export const APP_BASE_HREF = new InjectionToken<string>('appBaseHref');
...
- * or add a base element to the document. This URL prefix that will be preserved
- * when generating and recognizing URLs.
+ * or add a base element to the document.
  *
- * For instance, if you provide an `APP_BASE_HREF` of `'/my/app'` and call
+ * For instance, if you provide an `APP_BASE_HREF` of `'/my/app/'` and call
  * `location.go('/foo')`, the browser's URL will become
- * `example.com/my/app/foo`.
+ * `example.com/my/app/foo`. To ensure all relative URIs resolve correctly,
+ * the base href should end with a `/`.
```
petebacondarwin, line 89: Should this be?

```suggestion
 * the `APP_BASE_HREF` should end with a `/`.
```

packages/common/src/location/location_strategy.ts
```diff
@@ -81,17 +81,21 @@ export const APP_BASE_HREF = new InjectionToken<string>('appBaseHref');
...
  *
- * Similarly, if you add `<base href='/my/app'/>` to the document and call
+ * Similarly, if you add `<base href='/my/app/'/>` to the document and call
  * `location.go('/foo')`, the browser's URL will become
  * `example.com/my/app/foo`.
  *
+ * Note that when using `PathLocationStrategy`, neither the query nor
+ * the fragment in the base href will be preserved, as outlined
```
petebacondarwin, line 96:
```suggestion
 * the fragment in the `<base href>` will be preserved, as outlined
```

or does this cover both `<base href>` and `APP_BASE_HREF`?

petebacondarwin (changes requested):

aio/content/guide/router.md
```diff
@@ -554,9 +554,25 @@ Some developers may not be able to add the `<base>` element, perhaps because the
 
 Those developers may still use HTML5 URLs by taking the following two steps:
 
-1. Provide the router with an appropriate [APP_BASE_HREF][] value.
+1. Provide the router with an appropriate `APP_BASE_HREF` value.
 1. Use root URLs for all web resources: CSS, images, scripts, and template HTML files.
```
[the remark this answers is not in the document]
atscott: Not sure... Maybe it means absolute URLs?

aio/content/guide/router.md
```diff
@@ -554,9 +554,25 @@ Some developers may not be able to add the `<base>` element, perhaps because the
...
 
+A URL can have the following parts:
```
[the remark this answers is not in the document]
atscott: This section is simply here to define the parts of a URL so I can reference them below. This URL is not meant to be interpreted as a `<base href>` but just _any_ URL.

aio/content/guide/router.md
```diff
@@ -554,9 +554,25 @@ Some developers may not be able to add the `<base>` element, perhaps because the
...
+foo://example.com:8042/over/there?name=ferret#nose
+\_/   \______________/\_________/ \_________/ \__/
+ |           |            |            |        |
+scheme    authority      path        query   fragment
+```
+
+* The `<base href>` path should end with a "/", as browsers ignore characters in the path that follow the right-most "/".
+* If the `<base href>` path includes a query part, that part is only used if the page-URL path is empty and has no query.
+This means that a query in the base is only included when using `HashLocationStrategy`.
+* If a URL has an `authority` part, the authority, path, query, and fragment are taken from the URL. In this way, an `APP_BASE_HREF` with an authority can override the `<base href>` value.
```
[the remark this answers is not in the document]
atscott: hmmm, kind of...The way you've written it is confusing as well. What this means is that if an `APP_BASE_HREF` is something like "http://localhost:4200", i.e. it has an `authority`, then the `<base href>` is ignored. Angular always constructs link URLs using the `APP_BASE_HREF`. After that, the browser takes over and decides how to use the `<base href>` with those links on the page. If those links have an `authority` part because the `APP_BASE_HREF` did, then `<base href>` is not used. This is specifically to address the confusion in #25323

packages/common/src/location/location_strategy.ts
atscott: Yes and no. `APP_BASE_HREF` technically doesn't _need_ to end with `/` because we do `joinWithSlash(BASE_HREF, URL)` in the code. The one that _really_ needs a trailing slash is the `<base href>` element. I would really encourage a trailing slash in both though.

packages/common/src/location/location_strategy.ts
atscott: Only covers `<base href>`. Using a hash and/or fragment in the `APP_BASE_HREF` with `PathLocationStrategy` is incorrect and won't work at all.

aio/content/guide/router.md
```diff
@@ -554,8 +563,16 @@ Some developers may not be able to add the `<base>` element, perhaps because the
...
 
-1. Provide the router with an appropriate [APP_BASE_HREF][] value.
-1. Use root URLs for all web resources: CSS, images, scripts, and template HTML files.
+1. Provide the router with an appropriate `APP_BASE_HREF` value.
+1. Use root URLs (URLs with an `authority`) for all web resources: CSS, images, scripts, and template HTML files.
+
+* The `<base href>` `path` should end with a "/", as browsers ignore characters in the `path` that follow the right-most "/".
+* If the `<base href>` includes a `query` part, the `query` is only used if the `path` of a link in the page is empty and has no `query`.
+This means that a `query` in the `<base href>` is only included when using `HashLocationStrategy`.
+* If a link in the page has an `authority` part, the `<base href>` is not used. In this way, an `APP_BASE_HREF` with an authority will cause all links created by Angular to ignore the `<base href>` value.
```
petebacondarwin, line 572: Tentative suggestion. Feel free to ignore:

```suggestion
* If a link in the page is a root URL (has an `authority`), the `<base href>` is not used. In this way, an `APP_BASE_HREF` with an authority will cause all links created by Angular to ignore the `<base href>` value.
```

petebacondarwin (approved):
LGTM!

atscott:
blocked by #38085

atscott:
Removing blocked label. These changes should be incorporated into #38085 rather than blocking this small change by a large refactor that's not ready to submit.

PR #38124, opened 2020-07-17
build: cleanup .bazelrc file to no longer set unused flags

alexeagle:
This option is no longer needed in Bazel and will be an error in the future.


devversion (approved):
LGTM. Looks like the commit message needs to be updated (to match our guidelines). Let me know, and I can push to this PR to green it up. Thanks Alex!

alexeagle:
@devversion why didn't we run commitlint as a husky pre-push hook? It would be a lot nicer to know that before running through all the CI

josephperrott:
@alexeagle It [should have run](https://github.com/angular/angular/blob/master/package.json#L209), did it fail to run for you?

josephperrott (approved):
LGTM

PR #38126, opened 2020-07-17
fix(compiler): properly associate source spans for implicitly closed elements

### LINKED ISSUE #36118: HtmlParser incorrectly reports end tags (@angular/compiler)
# 🐞 bug report

### Affected Package

The issue is caused by package: `@angular/compiler`

### Is this a regression?

 Not sure

### Description

The Angular HTML parser incorrectly reports end tags for elements. 

I'm not sure what this could affect internally (other than incorrect reporting of html source code locations for users), but seems like something that should be fixed to make sure downstream functionality works correctly.

It's also affecting the codemod tooling that I'm building on top of Angular's HTML parser (which is how I discovered the issue).

## 🔬 Minimal Reproduction

 

```
import { HtmlParser } from '@angular/compiler';

const parser = new HtmlParser();
const ast = parser.parse(`<div><input></div>`, 'tmp.html');

console.log(ast);
```

StackBlitz: https://stackblitz.com/edit/angular-vfni6r?file=src%2Fapp%2Fapp.component.ts

The AST shows the `<input>` element having an end tag of `</div>` (via its `endSourceSpan`), and the `<div>` element has an `endSourceSpan` of `null` (signifying that it has no end tag)

## 🌍  Your Environment

**Angular Version:** StackBlitz / Angular v9.0.6

JoostK:
HTML is very lenient when it comes to closing elements, so Angular's parser has
rules that specify which elements are implicitly closed when closing a tag.
The parser keeps track of the nesting of tag names using a stack and parsing
a closing tag will pop as many elements off the stack as possible, provided
that the elements can be implicitly closed.

For example, consider the following templates:

- `<div><br></div>`, the `<br>` is implicitly closed when parsing `</div>`,
  because `<br>` is a void element.
- `<div><p></div>`, the `<p>` is implicitly closed when parsing `</div>`,
  as `<p>` is allowed to be closed by the closing of its parent element.
- `<ul><li>A <li>B</ul>`, the first `<li>` is implicitly closed when parsing
  the second `<li>`, whereas the second `<li>` would be implicitly closed when
  parsing the `</ul>`.

In all the cases above the parsed structure would be correct, however the source
span of the closing `</div>` would incorrectly be assigned to the element that
is implicitly closed. The problem was that closing an element would associate
the source span with the element at the top of the stack, however this may not
be the element that is actually being closed if some elements would be
implicitly closed.

This commit fixes the issue by assigning the end source span with the element
on the stack that is actually being closed. Any implicitly closed elements that
are popped off the stack will not be assigned an end source span, as the
implicit closing implies that no ending element is present.

Note that there is a difference between self-closed elements such as `<input/>`
and implicitly closed elements such as `<input>`. The former does have an end
source span (identical to its start source span) whereas the latter does not.

Fixes #36118
Resolves FW-2004

### DIFF
--- a/packages/compiler/src/ml_parser/parser.ts
+++ b/packages/compiler/src/ml_parser/parser.ts
@@ -337,56 +337,12 @@ class _TreeBuilder {
     return this._elementStack.length > 0 ? this._elementStack[this._elementStack.length - 1] : null;
   }
 
-  /**
-   * Returns the parent in the DOM and the container.
-   *
-   * `<ng-container>` elements are skipped as they are not rendered as DOM element.
-   */
-  private _getParentElementSkippingContainers():
-      {parent: html.Element|null, container: html.Element|null} {
-    let container: html.Element|null = null;
-
-    for (let i = this._elementStack.length - 1; i >= 0; i--) {
-      if (!isNgContainer(this._elementStack[i].name)) {
-        return {parent: this._elementStack[i], container};
-      }
-      container = this._elementStack[i];
-    }
-
-    return {parent: null, container};
-  }
-
   private _addToParent(node: html.Node) {
     const parent = this._getParentElement();
     if (parent != null) {
       parent.children.push(node);
     } else {
       this.rootNodes.push(node);
-    }
-  }
-
-  /**
-   * Insert a node between the parent and the container.
-   * When no container is given, the node is appended as a child of the parent.
-   * Also updates the element stack accordingly.
-   *
-   * @internal
-   */
-  private _insertBeforeContainer(
-      parent: html.Element, container: html.Element|null, node: html.Element) {
-    if (!container) {
-      this._addToParent(node);
-      this._elementStack.push(node);
-    } else {
-      if (parent) {
-        // replace the container with the new node in the children
-        const index = parent.children.indexOf(container);
-        parent.children[index] = node;
-      } else {
-        this.rootNodes.push(node);
-      }
-      node.children.push(container);
-      this._elementStack.splice(this._elementStack.indexOf(container), 0, node);
     }
   }
 
--- a/packages/compiler/src/ml_parser/parser.ts
+++ b/packages/compiler/src/ml_parser/parser.ts
@@ -261,8 +261,9 @@ class _TreeBuilder {
     const el = new html.Element(fullName, attrs, [], span, span, undefined);
     this._pushElement(el);
     if (selfClosing) {
-      this._popElement(fullName);
-      el.endSourceSpan = span;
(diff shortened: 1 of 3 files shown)


packages/compiler/src/ml_parser/parser.ts
```diff
@@ -281,25 +282,26 @@ class _TreeBuilder {
...
-    } else if (!this._popElement(fullName)) {
+    } else if (!this._popElement(fullName, endTagToken.sourceSpan)) {
       const errMsg = `Unexpected closing tag "${
           fullName}". It may happen when the tag has already been closed by another tag. For more info see https://www.w3.org/TR/html5/syntax.html#closing-elements-that-have-implied-end-tags`;
       this.errors.push(TreeError.create(fullName, endTagToken.sourceSpan, errMsg));
     }
   }
 
-  private _popElement(fullName: string): boolean {
+  private _popElement(fullName: string, sourceSpan: ParseSourceSpan): boolean {
```
petebacondarwin, line 296: NIT: rename this parameter to `endSourceSpan` or `closingSourceSpan`

packages/compiler/src/ml_parser/parser.ts
```diff
@@ -281,25 +282,26 @@ class _TreeBuilder {
...
     for (let stackIndex = this._elementStack.length - 1; stackIndex >= 0; stackIndex--) {
       const el = this._elementStack[stackIndex];
       if (el.name == fullName) {
+        // Record the parse span with the element that is being closed. Any element that is removed
```
petebacondarwin, line 300:
```suggestion
        // Record the parse span with the element that is being closed. Any elements that are removed
```

packages/compiler/test/ml_parser/html_parser_spec.ts
```diff
@@ -602,12 +605,53 @@ import {humanizeDom, humanizeDomSourceSpans, humanizeLineColumn} from './ast_spe
           expect(node.endSourceSpan!.end.offset).toEqual(12);
         });
 
+        it('should not set the end source span for self-closed elements', () => {
```
petebacondarwin, line 608: NIT:
```suggestion
        it('should not set the end source span for self-closing elements', () => {
```

petebacondarwin (approved):
Really nice and clean. Great commit message! And very nice way of testing.
A couple of nits, 'tis all! Thanks for taking this one.

petebacondarwin (approved):

AndrewKushnir:
[Presubmit](http://test/OCL:321957107:BASE:321957114:1595097371399:8b0b0029)

sonukapoor:
> Really nice and clean. Great commit message! And very nice way of testing.

Very much agreed 💯

PR #38130, opened 2020-07-18
fix(core): ICU expression previous value not removed inside template

### LINKED ISSUE #38073: Duplicate text when using ICU expression in list without 'other'
# 🐞 bug report

### Affected Package

### Is this a regression?

Pretty sure this wasn't an issue in Angular 7. Noticed the issue after updating to Angular 9.

### Description

When using an ICU select expression inside a list, if you do not include the "other" placeholder, then subsequent changes to an item in the list causes the replacement text to display twice.

Issue #34018 suggests that the "other" case should actually be required but this is now no longer enforced. It might be worth reinstating the warning if this is deemed not to be a bug.

## 🔬 Minimal Reproduction

https://stackblitz.com/edit/angular-ivy-cuoymq?file=src%2Fapp%2Fapp.component.html

## 🌍  Your Environment

**Angular Version:**
<pre><code>
Angular CLI: 9.1.10
Node: 12.18.2
OS: win32 x64

Angular: 9.1.11
Ivy Workspace: Yes

@angular-devkit/architect         0.901.10
@angular-devkit/build-angular     0.901.10
@angular-devkit/build-optimizer   0.901.10
@angular-devkit/build-webpack     0.901.10
@angular-devkit/core              9.1.10
@angular-devkit/schematics        9.1.10
@angular/cli                      9.1.10
@ngtools/webpack                  9.1.10
@schematics/angular               9.1.10
@schematics/update                0.901.10
rxjs                              6.6.0

</code></pre>

crisbeto:
Fixes an issue where the previous value of an ICU expression wasn't being
removed, if it didn't have an `other` parameter and it's the last value inside
a template. The issue is due to the `createDynamicNodeAtIndex` function
which sets a new node at a particular index inside the `LView`, but it doesn't
account for any node that may have existed there beforehand. Since we've
overwritten the reference, we don't have a way of removing the node from
the DOM anymore.

Also cleans up an optional parameter in one of the internal functions and
adds stronger typing to the i18n unit tests.

Fixes #38073.

### DIFF
--- a/packages/core/src/render3/i18n.ts
+++ b/packages/core/src/render3/i18n.ts
@@ -756,8 +756,10 @@ function createDynamicNodeAtIndex(
     tView: TView, lView: LView, index: number, type: TNodeType, native: RElement|RText|null,
     name: string|null): TElementNode|TIcuContainerNode {
   const previousOrParentTNode = getPreviousOrParentTNode();
-  ngDevMode && assertDataInRange(lView, index + HEADER_OFFSET);
-  lView[index + HEADER_OFFSET] = native;
+  const nodeIndex = index + HEADER_OFFSET;
+  ngDevMode && assertDataInRange(lView, nodeIndex);
+  lView[nodeIndex] !== null && removeNode(tView, lView, index, false);
+  lView[nodeIndex] = native;
   // FIXME(misko): Why does this create A TNode??? I would not expect this to be here.
   const tNode = getOrCreateTNode(tView, lView[T_HOST], index, type as any, name, null);
 
--- a/packages/core/test/acceptance/i18n_spec.ts
+++ b/packages/core/test/acceptance/i18n_spec.ts
@@ -1435,6 +1435,45 @@ onlyInIvy('Ivy i18n logic').describe('runtime i18n', () => {
       // checking the second ICU case
       expect(fixture.nativeElement.textContent.trim()).toBe('deux articles');
     });
+
+    it('should handle select expressions without an `other` parameter inside a template', () => {
+      const fixture = initWithTemplate(AppComp, `
+        <p *ngFor="let item of items">{item.value, select, 0 {A} 1 {B} 2 {C}}</p>
+      `);
+      fixture.componentInstance.items = [{value: 0}, {value: 1}, {value: 1337}];
+      fixture.detectChanges();
+      expect(fixture.nativeElement.textContent.trim()).toBe('AB');
+
+      fixture.componentInstance.items[0].value = 2;
+      fixture.detectChanges();
+      expect(fixture.nativeElement.textContent.trim()).toBe('CB');
+    });
+
+    it('should render an element whose case did not match initially', () => {
+      const fixture = initWithTemplate(AppComp, `
+        <p *ngFor="let item of items">{item.value, select, 0 {A} 1 {B} 2 {C}}</p>
+      `);
+      fixture.componentInstance.items = [{value: 0}, {value: 1}, {value: 1337}];
+      fixture.detectChanges();
+      expect(fixture.nativeElement.textContent.trim()).toBe('AB');
+
+      fixture.componentInstance.items[2].value = 2;
+      fixture.detectChanges();
+      expect(fixture.nativeElement.textContent.trim()).toBe('ABC');
+    });
+
+    it('should remove an element whose case matched initially, but does not anymore', () => {
+      const fixture = initWithTemplate(AppComp, `
+        <p *ngFor="let item of items">{item.value, select, 0 {A} 1 {B} 2 {C}}</p>
+      `);
+      fixture.componentInstance.items = [{value: 0}, {value: 1}];
+      fixture.detectChanges();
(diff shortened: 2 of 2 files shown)


petebacondarwin (commented):
You absolute legend @crisbeto !!

packages/core/test/acceptance/i18n_spec.ts
```diff
@@ -1435,6 +1435,19 @@ onlyInIvy('Ivy i18n logic').describe('runtime i18n', () => {
...
+
+    it('should handle select expressions without an `other` parameter inside a template', () => {
+      const fixture = initWithTemplate(AppComp, `
+        <p *ngFor="let item of items">{item.value, select, 0 {A} 1 {B} 2 {C}}</p>
+      `);
+      fixture.componentInstance.items = [{value: 0}, {value: 1}, {value: 1337}];
+      fixture.detectChanges();
+      expect(fixture.nativeElement.textContent.trim()).toBe('AB');
+
+      fixture.componentInstance.items[0].value = 2;
```
AndrewKushnir, line 1447: I think it'd be great to add a couple more tests/checks that would:
- update the value of an element that was not rendered originally (i.e. `items[2].value` in this case) and set the value to one of the conditions present in ICU
- change the value of an element that correspond to one of the cases in ICU (e.g. '1') to a value that would not be matched by ICU conditions (e.g. '5')

May be also copy all these tests for a standalone ICU (i.e. outside of an `ngFor`).

AndrewKushnir (commented):
Thanks for the fix @crisbeto 👍 

I just left a quick comment (with the proposal to check more cases) and will have closer look at the fix on Monday.

packages/core/test/acceptance/i18n_spec.ts
crisbeto: I've added the extra test cases. As for copying them to be without an `ngFor`, I didn't do it because I wasn't able to reproduce the issue without it. I think it has more to do with the number of nodes than the template definition itself.

packages/core/src/render3/i18n.ts
```diff
@@ -874,7 +874,7 @@ function readCreateOpCodes(
 
 function readUpdateOpCodes(
     updateOpCodes: I18nUpdateOpCodes, icus: TIcu[]|null, bindingsStartIndex: number,
-    changeMask: number, tView: TView, lView: LView, bypassCheckBit = false) {
+    changeMask: number, tView: TView, lView: LView, bypassCheckBit: boolean) {
```
petebacondarwin, line 877: This change is just a refactoring, yes?

petebacondarwin (commented):
I have to admit I don't really understand the sequence of calls to `readUpdateOpCode()` but it seems strange that we have to leave the `activeCaseIndex` as non-null (at line 944), even though (from my reading of the code) it should have been cleared up earlier on in the function (lines 916-941).

packages/core/src/render3/i18n.ts
petebacondarwin: I read in the commit message that it is. (NIT: personally I would have made these refactoring type changes in another commit to make it really clear that it is not a required part of the fix).

packages/core/test/acceptance/i18n_spec.ts
```diff
@@ -1435,6 +1435,45 @@ onlyInIvy('Ivy i18n logic').describe('runtime i18n', () => {
...
+      fixture.detectChanges();
+      expect(fixture.nativeElement.textContent.trim()).toBe('CB');
+    });
+
+    it('should render an element whose case did no match initially', () => {
```
petebacondarwin, line 1452:
```suggestion
    it('should render an element whose case did not match initially', () => {
```

packages/core/src/render3/i18n.ts
crisbeto: I figured that it was small enough that it wouldn't influence the review too much.

crisbeto:
My understanding of the entire process is somewhat high level, but I think the way it's set up is that it'll remove the old node next time it runs through the entire routine. If we reset it back to null, next time around it won't be able to remove the node since it doesn't know the index from last time.

petebacondarwin:
From what I can see, all the iterations of the `ngFor` loop use the same `icuTNode`, which means that on each loop the following happens:

1. remove the previous iteration's active case
2. create the current iteration's active case
3. update the current iteration's active case

This seems wrong to me... if the `icuTNode` is shared across all the ngFor iterations, then I would think it is not the right place to store the `activeCaseIndex`, right?

And I think it is this weird sharing that is causing the first iteration to be updated incorrectly, since the `activeCaseIndex`, from the last iteration, in the previous loop, is `null` it doesn't remove its previous active case before adding the new active case.

crisbeto:
I see what you mean now. Setting this back to draft and I'll take a better look later.

AndrewKushnir:
Hi @crisbeto, it looks like https://github.com/angular/angular/commit/6d8c73a4d62380bd7cfcaeaa23d5db802467d485 supersedes the changes in this PR? If that's the case, can we close this PR? Thank you.

PR #38131, opened 2020-07-18
fix(core): `Attribute` decorator `attributeName` is mandatory

### LINKED ISSUE #32658: Attribute decorator should allow attribute name as optional
# 🐞 bug report

### Affected Package
@angular/core
@angular/compiler

### Is this a regression?
Not sure, but I think not.

### Description
Attribute decorator is defined in `@angular/core` with `attributeName` as optional, but compiler throws an error when attributeName is missing.

I consider this a bug cause compiler doesn't meet the core specification.

## 🔬 Minimal Reproduction

This works

```typescript
constructor(@Attribute('customAttr') customAttr: string){}
```

This doesn't work

```typescript
constructor(@Attribute() customAttr: string) {}
```

## 🔥 Exception or Error

3428681797-image

## 🌍  Your Environment

**Angular Version:**
Angular 7.2.8

ajitsinghkaler:
`Attribute` decorator has defined `attributeName` as optional but actually its
 mandatory and compiler throwa an error if `attributeName` is undefined. Made
`attributeName` mandatory in the `Attribute` decorator to reflect this functionality

Fixes #32658

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #32658

- [ ] Yes
- [x] No

### DIFF
--- a/goldens/public-api/core/core.d.ts
+++ b/goldens/public-api/core/core.d.ts
@@ -53,7 +53,7 @@ export declare function asNativeElements(debugEls: DebugElement[]): any;
 export declare function assertPlatform(requiredToken: any): PlatformRef;
 
 export declare interface Attribute {
-    attributeName?: string;
+    attributeName: string;
 }
 
 export declare const Attribute: AttributeDecorator;
--- a/packages/compiler/src/core.ts
+++ b/packages/compiler/src/core.ts
@@ -22,10 +22,10 @@ export const createInjectionToken = makeMetadataFactory<object>(
     'InjectionToken', (desc: string) => ({_desc: desc, ɵprov: undefined}));
 
 export interface Attribute {
-  attributeName?: string;
+  attributeName: string;
 }
 export const createAttribute =
-    makeMetadataFactory<Attribute>('Attribute', (attributeName?: string) => ({attributeName}));
+    makeMetadataFactory<Attribute>('Attribute', (attributeName: string) => ({attributeName}));
 
 export interface Query {
   descendants: boolean;
--- a/packages/core/src/di/metadata.ts
+++ b/packages/core/src/di/metadata.ts
@@ -271,7 +271,7 @@ export interface Attribute {
   /**
    * The name of the attribute whose value can be injected.
    */
-  attributeName?: string;
+  attributeName: string;
 }
 
 /**


sonukapoor:
@ajitsinghkaler Are there any tests that cover this change?

ajitsinghkaler:
I don't think it is a breaking change because if the attributeName is left empty it will give you an error already. So eveybody should already be following it's just that we have not enforced it.

ajitsinghkaler:
> @ajitsinghkaler Are there any tests that cover this change?

@sonukapoor Actually I'm a bit confused if a test is actually needed for this because the attribute was already optional

ajitsinghkaler:
@sonukapoor Actually I'm a bit confused if a test is actually needed for this because the attribute was already optional

I actually don't know even what actually to test for example should we test if the attributeName is empty it should throw but that was already happening. A bit confused by this

sonukapoor:
> @sonukapoor Actually I'm a bit confused if a test is actually needed for this because the attribute was already optional
> 
> I actually don't know even what actually to test for example should we test if the attributeName is empty it should throw but that was already happening. A bit confused by this

I assume there is already a unit test for this, that checks if the `attributeName` is not passed, then an error is thrown. Can you just double-check for that?

crisbeto:
I don't know if that's actually the reason, but something to consider: maybe the idea with keeping the `attributeName` optional was that, if it omitted, the name would be picked up from the property that was decorated with it?

ajitsinghkaler:
I don't think so even in docs it is not optional.

jelbourn (approved):
LGTM, if this throws an error already, I don't see this as a breaking change

Reviewed-for: public-api

IgorMinar (approved):
LGTM!

Considering that there is no way to inject an attribute without also specifying the attribute name, this change is not a breaking change and therefor it should be fine to land it without much trouble.

If google3 presubmits pass, then we can go as far as landing this in a patch version.

Reviewed-for: public-api

petebacondarwin (approved):
Reviewed-for: public-api

jelbourn:
@AndrewKushnir PTAL

AndrewKushnir (approved):

ajitsinghkaler:
Should we add it to patch version

PR #38140, opened 2020-07-19
fix(forms): ValueAccessor.writeValue(...) sometimes called with null

### LINKED ISSUE #14988: ValueAccessor.writeValue is being called twice, first time with a phantom null value
```
[ ] feature request
```

**Current behavior**
The ValueAccessor.writeValue method is called twice on a custom value accessor during component initialization, when a control is bound using [(ngModel)], first time with a phantom null value.

Note that this is not an issue in form builder scenarios when control is bound using [formControlName].

**Expected behavior**
ValueAccessor.writeValue should be called only once, with an actual value of the property bound to [(ngModel)].

Run the following plunker:
[http://plnkr.co/edit/g5kjqQ9bQ9DF2cs4DbEH?p=preview](http://plnkr.co/edit/g5kjqQ9bQ9DF2cs4DbEH?p=preview)

MyValueAccessor is applied to the MyCmp component, the latter is bound using [(ngModel)] to the property initialized to 'Hello'. MyValueAccessor.writeValue writes the passed value to the console, the output looks as follows:
```
MyValueAccessor.writeValue: null
MyValueAccessor.writeValue: Hello
```

* **Angular version:** RC.4 - 2.4.9

* **Browser:** Chrome, IE11 

 
* **Language:** TypeScript

* **Node (for AoT issues):** `node --version` =

N2D4:
Fixes #14988
Fixes [@ckelley7's marriage](https://github.com/angular/angular/issues/14988#issuecomment-396342060)
## PR Checklist

(technically no, but no doc changes were needed)

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

`ValueAccessor.writeValue(...)` is called twice on the very first `ngModel` change if the form has not yet been set up; once by `setUpControl` and once when the value is actually updated. However, because `setUpControl` is called before the value is set, the first call passes `null`.

The bug is most common for standalone form controls which are never set up before the first `ngModel` change but may also happen if any other form control is not added to its parent first by calling `NgForm.addControl(...)`.

Issue Number: #14988

## What is the new behavior?
Optionally skip the `writeValue(...)` call in `setUpControl` if the value is a special sentinel value indicating it has not yet been initialized.

- [x] Yes
(see [below](https://github.com/angular/angular/pull/38140#issuecomment-661516210))
- [ ] No

### DIFF
--- a/packages/forms/src/directives/shared.ts
+++ b/packages/forms/src/directives/shared.ts
@@ -27,12 +27,20 @@ import {SelectControlValueAccessor} from './select_control_value_accessor';
 import {SelectMultipleControlValueAccessor} from './select_multiple_control_value_accessor';
 import {AsyncValidator, AsyncValidatorFn, Validator, ValidatorFn} from './validators';
 
+/**
+ * Sentinel to detect when a form control value has not yet been set. If it hasn't,
+ * writeValue(...) won't be called on the accessor.
+ *
+ * @internal
+ */
+export const INITIAL_CONTROL_VALUE = {};
 
 export function controlPath(name: string|null, parent: ControlContainer): string[] {
   return [...parent.path!, name!];
 }
 
-export function setUpControl(control: FormControl, dir: NgControl): void {
+export function setUpControl(
+    control: FormControl, dir: NgControl, opts?: {skipValueInitialization?: boolean}): void {
   if (typeof ngDevMode === 'undefined' || ngDevMode) {
     if (!control) _throwError(dir, 'Cannot find control with');
     if (!dir.valueAccessor) _throwError(dir, 'No value accessor for form control with');
@@ -40,7 +48,7 @@ export function setUpControl(control: FormControl, dir: NgControl): void {
 
   control.validator = Validators.compose([control.validator!, dir.validator]);
   control.asyncValidator = Validators.composeAsync([control.asyncValidator!, dir.asyncValidator]);
-  dir.valueAccessor!.writeValue(control.value);
+  if (!opts?.skipValueInitialization) dir.valueAccessor!.writeValue(control.value);
 
   setUpViewChangePipeline(control, dir);
   setUpModelChangePipeline(control, dir);
@@ -138,7 +146,7 @@ function _noControlError(dir: NgControl) {
   return _throwError(dir, 'There is no FormControl instance attached to form control element with');
 }
 
-function _throwError(dir: AbstractControlDirective, message: string): void {
+function _throwError(dir: AbstractControlDirective, message: string): never {
   let messageEnd: string;
   if (dir.path!.length > 1) {
     messageEnd = `path: '${dir.path!.join(' -> ')}'`;
--- a/packages/forms/src/directives/ng_form.ts
+++ b/packages/forms/src/directives/ng_form.ts
@@ -185,13 +185,27 @@ export class NgForm extends ControlContainer implements Form, AfterViewInit {
    *
    * @param dir The `NgModel` directive instance.
    */
-  addControl(dir: NgModel): void {
+  addControl(dir: NgModel): void;
+  /**
+   * @description
+   * Method that sets up the control directive in this group, re-calculates its value
+   * and validity, and adds the instance to the internal list of directives.
+   *
+   * @param dir The `NgModel` directive instance.
+   * @param opts An optional set of options. If `skipValueInitialization` is true, the control's
+   *     validity and value will not be updated and value accessors will not be written to. Use this
+   *     flag if the control's value is updated immediately after this call.
+   * @internal
+   */
+  addControl(dir: NgModel, opts: {skipValueInitialization?: boolean}): void;
+  /** @internal */
+  addControl(dir: NgModel, opts?: {skipValueInitialization?: boolean}) {
     resolvedPromise.then(() => {
       const container = this._findContainer(dir.path);
       (dir as {control: FormControl}).control =
           <FormControl>container.registerControl(dir.name, dir.control);
-      setUpControl(dir.control, dir);
-      dir.control.updateValueAndValidity({emitEvent: false});
+      setUpControl(dir.control, dir, opts);
+      if (!opts?.skipValueInitialization) dir.control.updateValueAndValidity({emitEvent: false});
       this._directives.push(dir);
     });
   }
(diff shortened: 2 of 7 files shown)


AndrewKushnir:
Hi @N2D4, thanks for creating this PR 👍 

I noticed that it's in Draft mode (i.e. it's probably not fully ready for review), but I wanted to share an idea. If I understand correctly the first "writeValue" with the `undefined` value comes as a result of `FormControl` instance creation in `ng_model.ts` [here](https://github.com/angular/angular/blob/master/packages/forms/src/directives/ng_model.ts#L137) (with no value). I was thinking if we can instantiate `FormControl` there with some "initial value" and check for it before calling the `writeValue` method in the `setUpControl` function, i.e. something like this:

```
// in packages/forms/src/directives/ng_model.ts:

export const INITIAL_CONTROL_VALUE = {};

...

public readonly control: FormControl = new FormControl(INITIAL_CONTROL_VALUE);


// in packages/forms/src/directives/shared.ts:

if (control.value !== INITIAL_CONTROL_VALUE) {
  dir.valueAccessor.writeValue(control.value);
}
```
Would that work? (If that works as needed, it should avoid passing configuration params though a few function calls)

Thank you.

N2D4:
> Hi @N2D4, thanks for creating this PR 👍
> 
> [...]
> 
> Would that work? (If that works as needed, it should avoid passing configuration params though a few function calls)
> 
> Thank you.

Sure, implemented. I also added an integration test (/modified an existing one), so this is now ready for review 👍

packages/forms/src/directives/shared.ts
```diff
@@ -27,6 +27,15 @@ import {SelectControlValueAccessor} from './select_control_value_accessor';
...
 import {AsyncValidator, AsyncValidatorFn, Validator, ValidatorFn} from './validators';
 
+/**
+ * Sentinel to detect when a form control value has not yet been set. If it hasn't,
+ * writeValue(...) won't be called on the accessor.
+ *
+ * @internal
+ */
+export const INITIAL_CONTROL_VALUE = {
+  _sentinel: ['shared/INITIAL_CONTROL_VALUE']
```
alfaproject, line 37: There's no need to increase bundle size with this extra attribute. The object reference is enough and you can just add a comment here instead if you want

N2D4: The logic is that if anyone messes around with the code elsewhere and doesn't see where the value is coming from they will have an easier time finding it, but I'll change it regardless

AndrewKushnir:
Hi @N2D4,

I performed initial review and added the "breaking changes" flag to this PR, since the new behavior is different from the old one (i.e. there us no additional `writeValue` call) and developers may rely on that, thus we need to highlight this change in the Breaking Changes section of Release Notes.

I've also ran a set of tests in Google's codebase ([internal-only link](https://test.corp.google.com/ui#id=OCL:322189886:BASE:322189965:1595270425941:46fe1c85)) and it looks like the problem is a bit more tricky. I found some additional cases were this change would be breaking. Here is a couple snippets that illustrate the problem:

Component's template:
```
<input
  myValidator
  [(ngModel)]="someField"
/>
```

Validator (simplified version):
```
@Directive({
  selector: '[myValidator]',
  providers: [{provide: NG_VALIDATORS, useExisting: MyValidator, multi: true}]
})
export class MyValidator implements Validator {
  validate(control: AbstractControl) {
    console.log(control.value);
    return null;
  }
}
```

The problem is that the `control.value` value in the `validate` function would be `INITIAL_CONTROL_VALUE`, i.e. just an empty object, thus causing some code to fail (the code that may not expect to see the object there at all). Also from API perspective having the `{}` here is quite unexpected. Also the question is whether a validator (sync/async) should actually be invoked in that case as well (I believe it should not)?

Given the fact that the `INITIAL_CONTROL_VALUE` may leak in some other places (possibly in `ngOnChanges` hook if you use `ngModel` as an input on a custom element, like `<my-comp [(ngModel)]="someField">`), I'm wondering if we should go back to the flags-based approach.

In any case, I think it'd be great to add the above mentioned example as a test case (it should be useful to count the number of times sync/async validator was called) and evaluate whether `INITIAL_CONTROL_VALUE`-based approach may work or other solution should be explored.

Please let me know your opinion.

Thank you.

N2D4:
Hi @AndrewKushnir,

Sorry for the delay, since this is scheduled for v11 I ended up re-prioritizing things but I'm back to work on this.

As an open-source intern I don't have access to the internal link, but I could manage to reproduce the error you mentioned. I agree that the validator shouldn't be called there, and added a unit test to make sure it's not. Yes, I believe in the `ngOnChanges` example you mentioned the value would leak as well, so I reverted back to the flag approach. Thanks for the feedback.

Now that we also skip validators when the form control is created, I took a closer look and decided to rename the flag to `skipValueInitialization` as that fits it better semantically. It should be passed to `addControl` (and the other methods) if and only if the value will be set/updated/initialized immediately after the call (as compared to already having been initialized). I've been thinking that we could also pass an initialization value instead - if given, `addControl` itself would initialize the control with that value first. It would complicate some of the logic in `setupControl` and `addControl` by a bit, but removes that weird invariant of "if you pass this flag, you MUST update the value immediately". Let me know if you find one better than the other.

There was one test that relied on the old behavior of calling validators synchronously during `ngOnChanges` when the component is first created. The call with argument `null` was already asynchronous for non-standalone, but synchronous for standalone components. Since the `null` call no longer happens, no such synchronous validator call will happen anymore. I don't think this should cause a lot of breakage in consumer unit tests, but I still thought it's worth pointing out.

JounQin:
Then, any news about merging?

packages/forms/src/directives/shared.ts
```diff
@@ -27,20 +27,28 @@ import {SelectControlValueAccessor} from './select_control_value_accessor';
 import {SelectMultipleControlValueAccessor} from './select_multiple_control_value_accessor';
 import {AsyncValidator, AsyncValidatorFn, Validator, ValidatorFn} from './validators';
 
+/**
+ * Sentinel to detect when a form control value has not yet been set. If it hasn't,
+ * writeValue(...) won't be called on the accessor.
+ *
+ * @internal
+ */
+export const INITIAL_CONTROL_VALUE = {};
```
JounQin, line 36: It is not used anywhere anymore.

alxhub:
This fix requires some additional design work / research as the current approach is too breaking to land as-is.

PR #38143, opened 2020-07-20
refactor(docs-infra): update docs examples `tslint.json` to match CLI and fix linting failures

gkalpak:
This PR updates the `tslint.json` configuration file, that is used to lint the docs examples, to match the one generated for new Angular CLI apps. There are some minimal differences (marked with `TODO` comments) for things, such as component selector prefix, that would require extensive and/or difficult to validate changes in guides.

This PR also includes the necessary changes to make the docs examples code compatible with the new tslint rules (broken up into multiple commits for easier reviewing).


aio/content/examples/testing/src/app/banner/banner-initial.component.spec.ts
```diff
@@ -75,45 +75,45 @@ describe('BannerComponent (with beforeEach)', () => {
...
+  // #docregion v4-test-4
+  it('should find the <p> with fixture.debugElement.nativeElement)', () => {
+    // #docregion debugElement-nativeElement
+    const bannerDe: DebugElement = fixture.debugElement;
+    const bannerEl: HTMLElement = bannerDe.nativeElement;
+    // #enddocregion debugElement-nativeElement
+    const p = bannerEl.querySelector('p');
+    expect(p.textContent).toEqual('banner works!');
+  });
+  // #enddocregion v4-test-4
```
petebacondarwin, line 107: Will this indentation affect the way that the region is displayed in the example?

petebacondarwin: OK, I just checked and it doesn't affect it!

gkalpak: No, beacuse we are left-aligning all docregions: https://github.com/angular/angular/blob/136acdfab62767f106c233f86bb47a7d2682ff00/aio/tools/transforms/examples-package/services/region-parser.js#L110

aio/content/examples/upgrade-module/src/app/hero-detail.directive.ts
```diff
@@ -12,7 +12,7 @@ export function heroDetailDirective() {
       <div><label>id: </label>{{$ctrl.hero.id}}</div>
       <button ng-click="$ctrl.onDelete()">Delete</button>
     `,
-    controller: function() {
+    controller: function HeroDetailController() {
```
petebacondarwin, line 15: Doesn't AngularJS support the `controller() { ... }` method format for these component definition objects?

petebacondarwin: Is it because this function is injected? And we don't support arrow function type stuff for injectables?

gkalpak: Controller's are `new`'d, so they need to either be constructor functions or classes.

aio/content/examples/dynamic-component-loader/src/app/ad-banner.component.ts
```diff
@@ -43,8 +43,8 @@ export class AdBannerComponent implements OnInit, OnDestroy {
     const viewContainerRef = this.adHost.viewContainerRef;
     viewContainerRef.clear();
 
-    const componentRef = viewContainerRef.createComponent(componentFactory);
-    (<AdComponent>componentRef.instance).data = adItem.data;
+    const componentRef = viewContainerRef.createComponent<AdComponent>(componentFactory);
```
petebacondarwin, line 46: 👍

aio/content/examples/observables/src/multicasting.ts
```diff
@@ -9,18 +9,18 @@ function sequenceSubscriber(observer) {
 
   // Will run through an array of numbers, emitting one value
   // per second until it gets to the end of the array.
-  function doSequence(arr, idx) {
+  function doInSequence(arr, idx) {
```
petebacondarwin, line 12: Is this change because of tslint or personal choice?

aio/content/examples/template-reference-variables/e2e/src/app.e2e-spec.ts
```diff
@@ -2,21 +2,16 @@ import { browser, element, by } from 'protractor';
...
+  beforeEach(() => browser.get(''));
 
   // helper function used to test what's logged to the console
   async function logChecker(button, contents) {
     const logs = await browser
       .manage()
       .logs()
       .get(logging.Type.BROWSER);
-    const message = logs.filter(({ message }) =>
-      message.indexOf(contents) !== -1 ? true : false
```
petebacondarwin, line 19: LOL!

aio/content/examples/animations/src/app/hero-list-page.component.ts
```diff
@@ -52,13 +52,11 @@ export class HeroListPageComponent implements OnInit {
...
 
-  _heroes = [];
 // #docregion filter-animations
   heroTotal = -1;
 // #enddocregion filter-animations
-  get heroes() {
-    return this._heroes;
-  }
+  get heroes() { return this._heroes; }
+  private _heroes = [];  // tslint:disable-line: variable-name
```
petebacondarwin, line 59: Unfortunately this comment gets rendered in the example, which I think is unwanted.
Perhaps better to disable the rule for the block?

aio/content/examples/architecture/e2e/src/app.e2e-spec.ts
```diff
@@ -86,13 +86,11 @@ function getPageElts() {
 
 async function heroFromDetail(detail: ElementFinder): Promise<Hero> {
   // Get hero id from the first <div>
-  // let _id = await detail.all(by.css('div')).first().getText();
-  let _id = await detail.all(by.css('div')).first().getText();
+  let id = await detail.all(by.css('div')).first().getText();
   // Get name from the h2
-  // let _name = await detail.element(by.css('h4')).getText();
```
petebacondarwin, line 94: strange - why the duplicated lines?

aio/content/examples/component-interaction/src/app/name-child.component.ts
```diff
@@ -6,13 +6,11 @@ import { Component, Input } from '@angular/core';
...
-  private _name = '';
-
   @Input()
+  get name(): string { return this._name; }
   set name(name: string) {
     this._name = (name && name.trim()) || '<no name set>';
   }
-
-  get name(): string { return this._name; }
+  private _name = '';  // tslint:disable-line: variable-name
```
petebacondarwin, line 14: Similar issue here, with the tslint comment being rendered in the example

aio/content/examples/inputs-outputs/src/app/aliasing/aliasing.component.ts
```diff
@@ -2,6 +2,7 @@
 /* tslint:disable:use-output-property-decorator */
 
 /* tslint:disable:no-input-rename */
+/* tslint:disable:no-output-rename */
```
petebacondarwin, line 5: Any reason not to combine this and the lines above into a single tslint disable directive?

aio/content/examples/built-in-directives/src/app/app.component.ts
```diff
@@ -76,11 +76,7 @@ export class AppComponent implements OnInit {
   }
 
   giveNullCustomerValue() {
-    !(this.nullCustomer = null) ? (this.nullCustomer = 'Kelly') : (this.nullCustomer = null);
```
petebacondarwin, line 73: I don't even understand what this code was trying to achieve... perhaps there is a typo in the condition (`=` should be `==`) ?

petebacondarwin (approved):
Great stuff. I made a bunch of small comments. Nothing to block merging this as it a definite improvement to the maintainability and health of all this code.

My two suggestions would be to check the docs themselves to ensure that changes to the examples (e.g. variable renames and `let`->`const`) match the prose of the docs too.  I have definitely seen some use of `let` in markdown guides.

aio/content/examples/observables/src/multicasting.ts
gkalpak: There is another `doSequence()` function in scope and tslint was complaining about shadowing it. So, I renamed this one.

aio/content/examples/animations/src/app/hero-list-page.component.ts
gkalpak: Yes, I was on the fence with this one. On one hand, showing the comment might be a little confusing for beginners. On the other hand, not showing it might be confusing when they will be getting a tslint error in case they try to use the same code in an app.

aio/content/examples/architecture/e2e/src/app.e2e-spec.ts
gkalpak: 🤷‍♂️ (Wasn't bothered to investigate the why :grin:)

aio/content/examples/inputs-outputs/src/app/aliasing/aliasing.component.ts
gkalpak: Good question 🤔 I'll fix it.

aio/content/examples/built-in-directives/src/app/app.component.ts
gkalpak: Yeah, it sounds like this function was trying to toggle the `nullCustomer` value. But it is not what the name suggests and toggling isn't needed for the example afaict, so I opted for changing the method to do what it says on the tin.

charanseeram:
Please unsubscribe me, looks unsubscribing over git is not working

gkalpak:
@charanseeram, I am not sure what you mean by "unsubscribing over git". Getting notifications for new issues/PRs is a GitHub feature. I assume you are "watching" this repository, so you can "unwatch" it to stop receiving these notifications. See the [GitHub docs on unwatching a repository][1] for more info.

[1]: https://docs.github.com/en/enterprise/2.17/user/github/receiving-notifications-about-activity-on-github/watching-and-unwatching-repositories#unwatching-a-single-repository

aio/content/examples/dependency-injection-in-action/e2e/src/app.e2e-spec.ts
```diff
@@ -1,76 +1,74 @@
-'use strict'; // necessary for es6 output in node
```
alan-agius4, line 1: Why not enable this via the TS compiler option `alwaysStrict`, instead of having it part of the code?

alan-agius4:
Can you actually update these to use the strict mode tslint.json?
```
"no-any": true
```

Also in v10 a new rule was added
```
"typedef": [true, "call-signature"],
```

aio/content/examples/dependency-injection-in-action/e2e/src/app.e2e-spec.ts
gkalpak: Not sure what you mean :confused:

gkalpak:
Thx for taking a look, @petebacondarwin and @alan-agius4!
@alan-agius4, I think I'll leave `"no-any": true` and `"typedef": [true, "call-signature"]` for another PR (together with updating docs examples to v10). This is one is pretty big as it is 😅

aio/content/examples/dependency-injection-in-action/e2e/src/app.e2e-spec.ts
alan-agius4: Why not let the compiler emit `'use strict';`, by setting `alwaysStrict` compiler options.

alan-agius4: Sorry I misunderstood that you removed  `use strict `

petebacondarwin (approved):
Reviewed-for: public-api

petebacondarwin (approved):
Reviewed-for: public-api

jelbourn:
@gkalpak in the future, it would be much better to split changes like this up into smaller PRs. In this case, probably making one type of style change at a time or targeting only a subset of the codebase at a time, then updating the lint rules after the code had been updated. It's rough when I see 400+ files changed.

aio/tools/examples/shared/boilerplate/systemjs/tslint.json
```diff
@@ -34,59 +64,86 @@
...
+        "check-branch",
+        "check-decl",
+        "check-operator",
+        "check-separator",
+        "check-type",
+        "check-typecast"
+      ]
+    },
+    "no-conflicting-lifecycle": true,
+    "no-host-metadata-property": true,
```
jelbourn, line 134: I personally strongly disagree with this rule, maybe even going so far as to call it an anti-pattern. It pains me that this is the recommendation and I hate to see it proliferating. 

@mgechev @StephenFluin would you have any objections to just removing this rule from the default Angular project tslint config?

aio/content/examples/animations/src/app/hero-list-page.component.ts
```diff
@@ -52,13 +54,11 @@ export class HeroListPageComponent implements OnInit {
...
 
-  _heroes = [];
 // #docregion filter-animations
   heroTotal = -1;
 // #enddocregion filter-animations
-  get heroes() {
-    return this._heroes;
-  }
+  get heroes() { return this._heroes; }
+  private _heroes = [];
```
jelbourn, line 61: Instead of adding tslint:disable, this looks like it could be fixed by just doing
```suggestion
  readonly heroes = [];
```

aio/tools/examples/shared/boilerplate/systemjs/tslint.json
```diff
@@ -34,59 +64,86 @@
...
+    "no-input-rename": true,
```
jelbourn, line 135: I also disagree with this rule, at a very slightly weaker strength than `no-host-metadata-property` and would like to get input on removing from the CLI's default config instead

aio/tools/examples/shared/boilerplate/systemjs/tslint.json
gkalpak: (If we decide to change this, we might also want to update the [style guide][1].)

[1]: https://angular.io/guide/styleguide#hostlistenerhostbinding-decorators-versus-host-metadata

aio/content/examples/animations/src/app/hero-list-page.component.ts
gkalpak: The problem with that is that we would then have to make changes to the rest of the component to be able to update the `heroes` property from inside the class (i.e. [here][1] and [there][2]), which felt outside the scope of this PR.

While it is possible to refactor the component to avoid these problems, it is probably outside the scope of this PR.

[1]: https://github.com/angular/angular/blob/9308804ae024529fddbe5c9bfc1883ef4dd4af93/aio/content/examples/animations/src/app/hero-list-page.component.ts#L64
[2]: https://github.com/angular/angular/blob/9308804ae024529fddbe5c9bfc1883ef4dd4af93/aio/content/examples/animations/src/app/hero-list-page.component.ts#L70

gkalpak:
> @gkalpak in the future, it would be much better to split changes like this up into smaller PRs. In this case, probably making one type of style change at a time or targeting only a subset of the codebase at a time, then updating the lint rules after the code had been updated. It's rough when I see 400+ files changed.

@jelbourn, while I generally agree on smaller, focused PRs, I am not sure it is a good approach for the types of changes in this PR. I feel that splitting this type of PRs into multiple smaller ones would involve either a huge overhead for the author (for little convenience for the reviewer) or more overhead and context switching for both the reviewer and the author. (I admit that this is my view and your mileage may vary :grin:)

I sometimes ask the primary reviewer how they prefer the work split up to facilitate reviewing. In this case I didn't do it, because the way it is currently split up seemed "objectively" better to me 😛
Made a mental note to avoid making assumptions on how others prefer a larger piece of work split up for easier reviewing 🧠 🎵

aio/tools/examples/shared/boilerplate/systemjs/tslint.json
alfaproject: If it is an anti-pattern, why do we have the decorator? I'm wondering what I should suggest my team after reading this and more importantly, why.

jelbourn: For this rule in particular, my rationale is:
*   HostBinding and HostListener cause problems with SSR because decorated methods preserve their type information at run-time, which can include browser globals that don't exist in node.js environments.
*   For complex components, the `host` property is much easier to quickly parse. See [MatSlider](https://github.com/angular/components/blob/master/src/material/slider/slider.ts#L120-L150) as an example- the CSS classes alone are _much_ easier to reason about when they're all defined in one place, and the seeing the ARIA attributes all together helps sanity check that the right attributes are applied.
*   HostListener doesn't allow _expressions_, which encourages developers to create unnecessary one-line methods (I've seen this in action a number of times)
*   HostBinding doesn't allow for static values (e.g. `role` and `tabindex`), leading users to create superfluous properties and bindings

The history here is that the style guide was written before anybody had any substantive experience authoring Angular apps, and it's now tricky to _change_ because we need to decide if we're going to do anything to update apps that were following the existing advice.

jelbourn:
@gkalpak yeah, I totally understand the rationale. I was chatting about this w/ Pete earlier, here's my thinking about why I'd have done this one in smaller parts:
* I wouldn't want to approve a PR without actually reading every file. When the PR is 400+ files with heterogeneous changes, I can mentally handle like 40 - 50 at a time, so it will take me longer than I'd like to get through the whole thing. I know I can look through commit-by-commit, but I personally have a harder time keeping track of what I've reviewed already that way (compared to using the "Reviewed" checkbox in the whole-PR view). 
* Large PRs can send a signal that other large PRs are acceptable, which we're on the same page about discouraging 
* Two of the lints here I think would be potentially better as disabled in the CLI defaults, but I also feel bad about holding up all of the changes on just a that small subset.

aio/content/examples/http/src/testing/http-client.spec.ts
```diff
@@ -73,7 +73,7 @@ describe('HttpClient testing', () => {
 
     // Make an HTTP GET request with specific header
     httpClient.get<Data>(testUrl, {
-        headers: new HttpHeaders({'Authorization': 'my-auth-token'})
+        headers: new HttpHeaders({Authorization: 'my-auth-token'})
```
jelbourn, line 76: This one is interesting because the difference here would be meaningful for Closure Compiler. With advanced optimizations (which are used inside Google), removing the quotes on this one would cause `Authorization` to be renamed/minified, which would affect the http request. It doesn't actually matter in this case, but makes wonder if this rule shouldn't really affect places where the type is (effectively) `Record<string, whatever>`

jelbourn (commented):
Okay, halfway through!

aio/content/examples/styleguide/e2e/src/app.e2e-spec.ts
```diff
@@ -59,20 +59,6 @@ describe('Style Guide', function () {
     expect(buttons.get(1).getText()).toBe('Hide toast');
   });
 
-  // temporarily disabled because of a weird issue when used with rxjs v6 with rxjs-compat
```
atscott, line 62: So 03-06 is deleted because the test is `xit`? I can't find where any of the `03-*` styles are documented. Are they all unused?

aio/tools/examples/shared/boilerplate/systemjs/tslint.json
mgechev: As we discussed offline, we can defer reconsidering this practice after we remove view engine.

jelbourn (approved):
LGTM aside from Andrew's comment

Reviewed-for: public-api

AndrewKushnir (approved):
LGTM 👍 

Reviewed-for: public-api

aio/content/examples/styleguide/e2e/src/app.e2e-spec.ts
gkalpak: Good catch! I thought only the `03-06` rule was removed, but it turns out all `03-*` rules were removed (in be0bc799f3fa99582238f29cb575e0705ce0c10f).

So, yes, all `03-*` examples are currently unused (i.e. not referenced anywhere in the docs), so they can all be removed.

gkalpak: Added a fixuo commit to removethe code corresponding to the remaining `03-*` rules: https://github.com/angular/angular/pull/38143/commits/93a02a036fdc85116e31582e5a960de94c1c2ab3

PTAL

atscott (approved):
Reviewed-for: public-api

gkalpak:
Regarding https://github.com/angular/angular/pull/38143#issuecomment-661248648, I have updated the docs examples tslint config in #38993, but left `no-any` and `typedef` commented out (as TODOs), because there were many failures.
I'll hopefully get to it some time in the future (or better yet, someone might beat me to it 😛).

PR #38146, opened 2020-07-20
build: update pullapprove to remove @matsko from reviewer lists

josephperrott:
With @matsko leaving the Angular team, we need to update the pullapprove
configuration to reflect his no longer being a reviewer for file groups
throughout the repository.

### DIFF
--- a/.pullapprove.yml
+++ b/.pullapprove.yml
@@ -175,7 +175,6 @@ groups:
          - kara                    # Kara Erickson
          - kyliau                  # Keen Yee Liau
          - manughub                # Manu Murthy
-         - matsko                  # Matias Niemela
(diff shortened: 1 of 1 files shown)


AndrewKushnir (approved):

.pullapprove.yml
```diff
@@ -206,7 +205,9 @@ groups:
           ])
     reviewers:
       users:
-        - matsko
+        - IgorMinar
+        - crisbeto
+        - jelbourn
```
AndrewKushnir, line 210: nit: it looks like we have these list in alphabet order in other places in this file, may be change it to alphabet order here too?

jelbourn (approved):
LGTM

PR #38147, opened 2020-07-20
fix(compiler): mark `NgModuleFactory` construction as not side effectful

dgp1130:
This allows Closure compiler to tree shake unused constructor calls to `NgModuleFactory`, which is otherwise considered side-effectful. The Angular compiler generates factory objects which are exported but typically not used, as they are only needed for compatibility with View Engine. This results in top-level constructor calls, such as:

```typescript
export const FooNgFactory = new NgModuleFactory(Foo);
```

`NgModuleFactory` has a side-effecting constructor, so this statement cannot be tree shaken, even if `FooNgFactory` is never imported. The `NgModuleFactory` continues to reference its associated `NgModule` and prevents the module and all its unused dependencies from being tree shaken. This effectively prevents all components from being tree shaken, making Closure builds significantly larger than they should be.

The fix here is to wrap `NgModuleFactory` constructor with `noSideEffects(() => /* ... */)`, which tricks the Closure compiler into assuming that the invoked function has no side effects. This allows it to tree-shake unused `NgModuleFactory()` constructors when they aren't imported. Since the factory can be removed, the module can also be removed (if nothing else references it), thus tree shaking unused components as expected.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: http://b/161548609

Unused components are not tree shaken in Closure compiler.

## What is the new behavior?

Unused components are now tree shaken in Closure compiler.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/compiler-cli/test/ngtsc/fake_core/index.ts
+++ b/packages/compiler-cli/test/ngtsc/fake_core/index.ts
@@ -103,3 +103,5 @@ export interface QueryList<T>/* implements Iterable<T> */ {
 export type NgIterable<T> = Array<T>|Iterable<T>;
 
 export class NgZone {}
+
+export declare function ɵnoSideEffects<T>(fn: () => T): T;
--- a/packages/compiler-cli/test/ngtsc/ngtsc_spec.ts
+++ b/packages/compiler-cli/test/ngtsc/ngtsc_spec.ts
@@ -3538,7 +3538,9 @@ runInEachFileSystem(os => {
       expect(factoryContents).toContain(`import * as i0 from '@angular/core';`);
       expect(factoryContents).toContain(`import { NotAModule, TestModule } from './test';`);
       expect(factoryContents)
-          .toContain(`export var TestModuleNgFactory = new i0.\u0275NgModuleFactory(TestModule);`);
+          .toContain(
+              'export var TestModuleNgFactory = ' +
+              'i0.ɵnoSideEffects(function () { return new i0.\u0275NgModuleFactory(TestModule); });');
       expect(factoryContents).not.toContain(`NotAModuleNgFactory`);
       expect(factoryContents).not.toContain('\u0275NonEmptyModule');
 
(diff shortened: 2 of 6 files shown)


dgp1130:
http://test/OCL:322443590:BASE:322443701:1595367128330:be18d491

dgp1130:
@mhevery, @alxhub, can you please take a look at this PR?

mhevery (approved):

dgp1130:
Status checks wanted me to rerun g3 presubmit, so here's the new results:
http://test/OCL:322443590:BASE:323406497:1595875409766:177c87aa

Just a number of flakes/preexisting failures. I don't see anything that seems related to this PR.

packages/compiler-cli/test/ngtsc/fake_core/index.ts
```diff
@@ -103,3 +103,7 @@ export interface QueryList<T>/* implements Iterable<T> */ {
 export type NgIterable<T> = Array<T>|Iterable<T>;
 
 export class NgZone {}
+
+export function ɵnoSideEffects<T>(fn: () => T): T {
+  return fn();
```
alxhub, line 108: This is fine, but there's no actual reason to have an implementation here - `fake_core` is just used for its type.

packages/compiler-cli/test/ngtsc/ngtsc_spec.ts
```diff
@@ -41,17 +41,19 @@ function getDiagnosticSourceCode(diag: ts.Diagnostic): string {
...
 
-runInEachFileSystem(os => {
-  describe('ngtsc behavioral tests', () => {
-    let env!: NgtscTestEnvironment;
-
-    beforeEach(() => {
-      env = NgtscTestEnvironment.setup(testFiles);
-      env.tsconfig();
-    });
+runInEachFileSystem(
```
alxhub, line 44: There are some weird formatting issues in this file - we should be able to avoid this.

alxhub (changes requested):
This LGTM, save for the formatting of `ngtsc_spec`.

packages/compiler-cli/test/ngtsc/fake_core/index.ts
dgp1130: Updated to an `export declare function` with no body.

packages/compiler-cli/test/ngtsc/ngtsc_spec.ts
dgp1130: Sorry, the formatter got a little over-zealous. Reverted the changes and updated my code so the formatter doesn't flip out from it.

alxhub (approved):

PR #38149, opened 2020-07-20
Update developer test documentation

dgp1130:
A couple times now I've run `yarn bazel test //packages/...` as the documentation suggests only to find some tests failing. This is because it runs all tests with View Engine even though not all actually support View Engine. I've updated the docs to use `yarn test-ivy-aot` and `yarn test-non-ivy` which filter out tests that do not support each renderer.


petebacondarwin (approved):

josephperrott (approved):
LGTM

PR #38151, opened 2020-07-21
docs: Remove incorrect RxJS-related instruction

benlesh:
It appears that this was changed code from the pre-pipeable operator days.

I'm making this PR to draw it to your attention more than anything. I'm not sure how you want to word this, but now users would just pull in whatever operators they need, if they need to, from `rxjs/operators` directly in the files they use them in. What's here looks like it's a remnant from when we had to modify `Observable.prototype` with things like `import "rxjs/add/operator/map"` etc.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/aio/content/guide/http.md
+++ b/aio/content/guide/http.md
@@ -38,14 +38,6 @@ You can then inject the `HttpClient` service as a dependency of an application c
   header="app/config/config.service.ts (excerpt)">
 </code-example>
 
-The `HttpClient` service makes use of [observables](guide/glossary#observable "Observable definition") for all transactions. You must import the RxJS observable and operator symbols that appear in the example snippets. These `ConfigService` imports are typical.
-
-<code-example
-  path="http/src/app/config/config.service.ts"
-  region="rxjs-imports"
-  header="app/config/config.service.ts (RxJS imports)">
-</code-example>
-
 <div class="alert is-helpful">
 
 You can run the <live-example></live-example> that accompanies this guide.


petebacondarwin:
I don't understand this change @benlesh. The example you delete in this guide does use pipeable operators, and the imports are valid. Can you explain what is misleading or wrong about the current doc?

kapunahelewong:
Do we need to update the example too? Right now the snippet is this:

```
import { Observable, throwError } from 'rxjs';
import { catchError, retry } from 'rxjs/operators';
```

But this seems like what you're describing as the current way, right @benlesh? 
I could definitely be missing something here. 🙃 😅

benlesh:
I think this is a remnant from when we used to have operator patching back in rxjs 5. The config service would have a bunch of imports in it like `import 'rxjs/add/operator/map'`. Since rxjs 5.5 , it has been recommended to import "pipeable" operators directly into the file where you use it, and use them directly with `pipe`.

kapunahelewong:
Thanks, @benlesh. 😊

For the documentation team: we will need to pair with an engineer to make sure that we update the example correctly. I think this would be a quick refactor.

petebacondarwin:
Closing this. The example already uses pipeable operators. So it is already following the correct RxJS practices.

PR #38164, opened 2020-07-21
refactor(dev-infra): Add support for groups in the conditions evaluator

atscott:
Conditions can refer to the groups array that is a list of the preceding groups.
This commit adds support to the verification for those conditions.

This commit also adds some tests to the parsing and condition matching
to ensure everything works as expected.

This PR is needed to unblock #37798

### DIFF
--- a/dev-infra/pullapprove/BUILD.bazel
+++ b/dev-infra/pullapprove/BUILD.bazel
@@ -40,7 +40,6 @@ ts_library(
         ":pullapprove",
         "@npm//@types/jasmine",
         "@npm//@types/node",
-        "@npm//inquirer",
         "@npm//typescript",
     ],
 )
--- a/dev-infra/pullapprove/BUILD.bazel
+++ b/dev-infra/pullapprove/BUILD.bazel
@@ -1,4 +1,5 @@
 load("@npm_bazel_typescript//:index.bzl", "ts_library")
+load("//tools:defaults.bzl", "jasmine_node_test")
 
 ts_library(
     name = "pullapprove",
@@ -8,6 +9,8 @@ ts_library(
         "group.ts",
         "logging.ts",
         "parse-yaml.ts",
+        "pullapprove_arrays.ts",
+        "utils.ts",
         "verify.ts",
     ],
     module_name = "@angular/dev-infra-private/pullapprove",
@@ -24,4 +27,26 @@ ts_library(
         "@npm//yaml",
         "@npm//yargs",
     ],
+)
+
+ts_library(
+    name = "pullapprove_test_lib",
+    testonly = True,
+    srcs = glob(
+        ["*.spec.ts"],
+    ),
+    visibility = ["//visibility:private"],
+    deps = [
+        ":pullapprove",
+        "@npm//@types/jasmine",
+        "@npm//@types/node",
+        "@npm//inquirer",
+        "@npm//typescript",
+    ],
+)
+
+jasmine_node_test(
+    name = "pullapprove_test",
+    srcs = [":pullapprove_test_lib"],
+    visibility = ["//visibility:private"],
 )
--- a/dev-infra/pullapprove/group.ts
+++ b/dev-infra/pullapprove/group.ts
@@ -117,4 +117,8 @@ export class PullApproveGroup {
       groupName: this.groupName,
     };
   }
+
+  toString() {
+    return this.groupName;
+  }
 }
--- a/dev-infra/pullapprove/group.ts
+++ b/dev-infra/pullapprove/group.ts
@@ -6,14 +6,15 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {error} from '../utils/console';
+import {error, warn} from '../utils/console';
 import {convertConditionToFunction} from './condition_evaluator';
 import {PullApproveGroupConfig} from './parse-yaml';
+import {PullApproveGroupStateDependencyError} from './pullapprove_arrays';
 
 /** A condition for a group. */
 interface GroupCondition {
   expression: string;
-  checkFn: (files: string[]) => boolean;
+  checkFn: (files: string[], groups: PullApproveGroup[]) => boolean;
   matchedFiles: Set<string>;
 }
 
@@ -39,7 +40,9 @@ export class PullApproveGroup {
   /** List of conditions for the group. */
   conditions: GroupCondition[] = [];
 
-  constructor(public groupName: string, config: PullApproveGroupConfig) {
+  constructor(
+      public groupName: string, config: PullApproveGroupConfig,
+      readonly precedingGroups: PullApproveGroup[] = []) {
     this._captureConditions(config);
   }
 
@@ -78,18 +81,26 @@ export class PullApproveGroup {
   testFile(filePath: string): boolean {
     return this.conditions.every(({matchedFiles, checkFn, expression}) => {
       try {
-        const matchesFile = checkFn([filePath]);
+        const matchesFile = checkFn([filePath], this.precedingGroups);
         if (matchesFile) {
           matchedFiles.add(filePath);
         }
         return matchesFile;
       } catch (e) {
-        const errMessage = `Condition could not be evaluated: \n\n` +
-            `From the [${this.groupName}] group:\n` +
-            ` - ${expression}` +
-            `\n\n${e.message} ${e.stack}\n\n`;
-        error(errMessage);
-        process.exit(1);
+        if (e instanceof PullApproveGroupStateDependencyError) {
+          const errMessage = `Condition could not be evaluated: \n` +
+              `${e.message}\n` +
+              `From the [${this.groupName}] group:\n` +
+              ` - ${expression}`;
+          warn(errMessage);
+        } else {
+          const errMessage = `Condition could not be evaluated: \n\n` +
+              `From the [${this.groupName}] group:\n` +
+              ` - ${expression}` +
+              `\n\n${e.message} ${e.stack}\n\n`;
+          error(errMessage);
+          process.exit(1);
+        }
       }
     });
   }
--- a/dev-infra/pullapprove/group.ts
+++ b/dev-infra/pullapprove/group.ts
@@ -87,6 +87,10 @@ export class PullApproveGroup {
         }
         return matchesFile;
       } catch (e) {
+        // In the case of a condition that depends on the state of groups we want to just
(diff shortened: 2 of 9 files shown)


dev-infra/pullapprove/group.ts
```diff
@@ -107,3 +109,70 @@ export class PullApproveGroup {
     };
   }
 }
+
+/**
+ * Superset of a native array. The superset provides methods which mimic the
+ * list data structure used in PullApprove for files in conditions.
+ */
+export class PullApproveGroupArray extends Array<PullApproveGroup> {
```
devversion, line 117: Should we move this and the other class for the `files` array into a dedicated file? Also did you consider this extending the more general `PullApproveArray` class?

dev-infra/pullapprove/group.ts
```diff
@@ -107,3 +109,70 @@ export class PullApproveGroup {
...
+   */
+  get inactive() {
+    return this;
+  }
+
+  /**
+   * Returns the current list.
+   * Note that this does not currently support accurate pending/active/inactive/rejected filtering verification.
+   * The implementations are here only to prevent errors when evaluating conditions which use these properties.
+   * More robust implementations could be built out in the future, but are not available at this time.
```
devversion, line 161: I was actually thinking that we throw an unique error in such stateful methods? that error is then caught and results in the condition being ignored (as it's reliant on actual approved/rejected/inactive/active groups). I think that might make sense long-term too.

devversion (commented):
Nice work so far 💯 (and nice to have tests!)

dev-infra/pullapprove/verify.ts
```diff
@@ -100,3 +95,12 @@ export function verify() {
   // Provide correct exit code based on verification success.
   process.exit(verificationSucceeded ? 0 : 1);
 }
+
+/** Parses all of the groups defined in the pullapprove yaml. */
+export function getGroupsFromYaml(pullApproveYamlRaw: string): PullApproveGroup[] {
```
josephperrott, line 100: Could we move this function into the `parse-yaml` file?  I think it makes sense to group all of the functions that handle the raw yaml together.

josephperrott (changes requested):
Love the tests!  Thanks!

dev-infra/pullapprove/group.ts
atscott: Moved both to `pullapprove_arrays.ts`. I did consider extending the PullApproveArray, but the implementations don't quite match enough because `include` and `exclude` require the members to be a `string`. The best I could do would be an abstract class that has abstract `include` and `exclude` methods.

dev-infra/pullapprove/group.ts
atscott: 👍

dev-infra/pullapprove/verify.ts
atscott: Moved

dev-infra/pullapprove/BUILD.bazel
```diff
@@ -25,3 +26,26 @@ ts_library(
...
+    srcs = glob(
+        ["*.spec.ts"],
+    ),
+    visibility = ["//visibility:private"],
+    deps = [
+        ":pullapprove",
+        "@npm//@types/jasmine",
+        "@npm//@types/node",
+        "@npm//typescript",
+        "@npm//inquirer",
```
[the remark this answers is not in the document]
atscott: It's somehow required...The tests wouldn't run without it. Do you want to pull down this PR and see if you can figure out if it can be removed?

dev-infra/pullapprove/group.ts
devversion: Thanks for moving the files. The reason I was proposing extending `PullApproveArray` is that currently `include/exclude` do not implement patterns correctly as they only support a RegExp and not a glob. Sounds like a discrepancy we should fix?

atscott: Hmmmm, yea you're right. I'm not actually sure what the `include` and `exclude` methods on `Groups` is supposed to support. The documentation looks like it is just a copy-paste from the files https://docs.pullapprove.com/context/#groups. I don't know what matching against a glob would really mean for groups.

devversion: I'm thinking it's the same base logic as for he `files` context (hence my initial thought of sharing that class). Should we change it?

dev-infra/pullapprove/BUILD.bazel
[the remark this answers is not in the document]
josephperrott: So it looks like its actually a missing dependency from `//dev-infra/utils/`

`//dev-infra/utils:utils` needs to add `"@npm//inquirer"` as a dependency.

Can we go ahead and add it there in this PR?

atscott:
Comments addressed. PTAL

dev-infra/pullapprove/group.ts
```diff
@@ -78,18 +81,26 @@ export class PullApproveGroup {
...
         }
         return matchesFile;
       } catch (e) {
-        const errMessage = `Condition could not be evaluated: \n\n` +
-            `From the [${this.groupName}] group:\n` +
-            ` - ${expression}` +
-            `\n\n${e.message} ${e.stack}\n\n`;
-        error(errMessage);
-        process.exit(1);
+        if (e instanceof PullApproveGroupStateDependencyError) {
```
josephperrott, line 90: Can you add commenting in here that explains why there is one path that is just a warning and one that is an error and exit?

dev-infra/pullapprove/group.ts
```diff
@@ -106,4 +117,8 @@ export class PullApproveGroup {
       groupName: this.groupName,
     };
   }
+
+  toString() {
```
josephperrott, line 121: Is there a reason we have a `toString` method? I don't see it referenced anywhere.

dev-infra/pullapprove/pullapprove_arrays.ts
```diff
@@ -0,0 +1,106 @@
...
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import {PullApproveGroup} from './group';
+import {getOrCreateGlob} from './utils';
+
+export class PullApproveGroupStateDependencyError extends Error {
+  constructor(message?: string) {
+    super(message);
+    // see: typescriptlang.org/docs/handbook/release-notes/typescript-2-2.html
```
josephperrott, line 14: It looks like due to how the DNS records are set up for typescriptlang.org, this URL needs to contain www

dev-infra/pullapprove/pullapprove_arrays.ts
```diff
@@ -0,0 +1,106 @@
...
+    Object.setPrototypeOf(this, new.target.prototype);      // restore prototype chain
+    this.name = PullApproveGroupStateDependencyError.name;  // stack traces display correctly now
```
josephperrott, line 16: Can you expand the comment on what is being done with this `name` value?

dev-infra/pullapprove/verify.ts
```diff
@@ -23,12 +21,7 @@ export function verify() {
   const REPO_FILES = allFiles();
   /** The pull approve config file. */
   const pullApproveYamlRaw = readFileSync(PULL_APPROVE_YAML_PATH, 'utf8');
-  /** JSON representation of the pullapprove yaml file. */
-  const pullApprove = parsePullApproveYaml(pullApproveYamlRaw);
-  /** All of the groups defined in the pullapprove yaml. */
-  const groups = Object.entries(pullApprove.groups).map(([groupName, group]) => {
-    return new PullApproveGroup(groupName, group);
-  });
+  const groups = getGroupsFromYaml(pullApproveYamlRaw);
```
josephperrott, line 24: Can you add a comment for `groups` symbol?

josephperrott (changes requested):

dev-infra/pullapprove/group.ts
devversion: I assume it's a leftover from testing. I agree that this looks unused.

dev-infra/pullapprove/pullapprove_arrays.ts
```diff
@@ -0,0 +1,106 @@
...
+  protected getFilterKey(s: string) {
+    return s;
+  }
+}
+
+/**
+ * Superset of a native array. The superset provides methods which mimic the
+ * list data structure used in PullApprove for groups in conditions.
+ */
+export class PullApproveGroupArray extends PullApproveArray<PullApproveGroup> {
```
devversion, line 73: After reading through this. do you know in what cases one would rely on `groups.include/exclude´? Just out of curiosity.

dev-infra/pullapprove/pullapprove_arrays.ts
devversion: Yeah, should probably link to https://github.com/Microsoft/TypeScript/wiki/FAQ#why-doesnt-extending-built-ins-like-error-array-and-map-work for this. Could also just the actual class name for more clarity.

devversion (approved):
LGTM, except for Joey's comments.

dev-infra/pullapprove/group.ts
```diff
@@ -78,18 +81,31 @@ export class PullApproveGroup {
...
-            ` - ${expression}` +
-            `\n\n${e.message} ${e.stack}\n\n`;
-        error(errMessage);
-        process.exit(1);
+        // In the case of a condition that depends on the state of groups we want to just
+        // warn that the verification can't accurately evaluate the condition and then
+        // continue processing. Other types of errors fail the verification, as conditions
+        // should otherwise be able to execute without throwing.
+        if (e instanceof PullApproveGroupStateDependencyError) {
+          debugger;
```
josephperrott, line 95: Looks like a `debugger` got left in place

josephperrott (changes requested):
Looking good, just one `debugger` got left in place.

josephperrott (approved):
LGTM

atscott:
@devversion I had to revert the shared base class of `PullApproveArray<T>` because I couldn't get it to work as expected. I've re-requested your review for approval.

devversion (approved):
LGTM. @atscott makes sense. I assume there would be ways to get the base class work, but it's not worth the effort. Thanks for trying it in the first place. This looks great to me!

devversion:
@googlebot Ping.

PR #38027, opened 2020-07-13
docs: refactor observables overview

jbogarthyde:
Refactor existing documentation to follow current style standards and use task-based organization

## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [ ] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [x] Documentation content changes
- [ ] angular.io application / infrastructure changes
- [ ] Other... Please describe:

## What is the current behavior?

Style: uses "we" construction
Language is sometimes awkward
Refers to things "discussed later" without explicit reference

Issue Number: N/A

## What is the new behavior?

Correct construction and language for style and clarity
Add ref to RxJS Subject doc for multicasting (from contributor suggestion).
Move error handling from RxJS page to this page (replacing forward reference)

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

PR #38165, opened 2020-07-21
Refactor observables doc for style, clarity, and structure

jbogarthyde:
Edit to conform to current doc standards, combine overview and RxJS pages into one, introduce relationship to RxJS

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Document uses older styles, language not always clear

Issue Number: This replaces earlier PR https://github.com/angular/angular/pull/38027

## What is the new behavior?
-- Correct construction and language for style and clarity
-- Add ref to RxJS Subject doc for multicasting, with example  (from contributor suggestion).
-- Combine short RxJS page w/Overview page, clarify relation to RxJS in introduction

- [ ] Yes
- [x] No


aio/content/guide/observables.md
```diff
@@ -116,4 +127,105 @@ myObservable.subscribe({
...
+
+<code-example path="rx-library/src/simple-creation.ts" region="interval" header="Create an observable from a counter"></code-example>
+
+<code-example path="rx-library/src/simple-creation.ts" region="event" header="Create an observable from an event"></code-example>
+
+<code-example path="rx-library/src/simple-creation.ts" region="ajax" header="Create an observable that creates an AJAX request"></code-example>
+
+### Operators
+
+Operators are functions that build on the observables foundation to enable sophisticated manipulation of collections. For example, RxJS defines operators such as `map()`, `filter()`, `concat()`, and `flatMap()`.
```
aikidave, line 184:
```suggestion
Operators are functions that allow you to manipulate the values observables return. For example, RxJS defines operators such as `map()`, `filter()`, `concat()`, and `flatMap()`.
```

aio/content/guide/observables.md
```diff
@@ -116,4 +127,105 @@ myObservable.subscribe({
...
+
+Operators take configuration options, and they return a function that takes a source observable. When executing this returned function, the operator observes the source observable’s emitted values, transforms them, and returns a new observable of those transformed values. Here is a basic example:
```
aikidave, line 186:
```suggestion
Operators take configuration options, and they return a function that takes a source observable. When executing this returned function, the operator observes the source observable’s emitted values, transforms them, and returns a new observable of those transformed values. For example:
```

aio/content/guide/observables.md
```diff
@@ -116,4 +127,105 @@ myObservable.subscribe({
...
+
+<code-example path="rx-library/src/operators.ts" header="Map operator"></code-example>
+
+You can use _pipes_ to link operators together. Pipes let you combine multiple functions into a single function. The `pipe()` function takes as its arguments the functions you want to combine, and returns a new function that, when executed, runs the composed functions in sequence.
+
+<div class="alert is-helpful">
+
+  Note that, for Angular apps, combining operators with pipes is preferred to _chaining_, which is used in many RxJS examples.
```
aikidave, line 194:
```suggestion
 For Angular applications, it is recommended that you combine operators with pipes.
```

aio/content/guide/observables.md
```diff
@@ -116,4 +127,105 @@ myObservable.subscribe({
...
+
+</div>
+
+A set of operators applied to an observable is a recipe&mdash;that is, a set of instructions for producing the values you’re interested in. By itself, the recipe doesn’t do anything. You need to call `subscribe()` to produce a result through the recipe.
```
aikidave, line 198: Is recipe in our glossary?

aio/content/guide/observables.md
```diff
@@ -116,4 +127,105 @@ myObservable.subscribe({
...
+
+Here’s an example:
+
+<code-example path="rx-library/src/operators.1.ts" header="Standalone pipe function"></code-example>
+
+The `pipe()` function is also a method on the RxJS `Observable`, so you can use this shorter form to define the same operation:
```
aikidave, line 204:
```suggestion
The `pipe()` function is also a method on the RxJS `Observable`, which means you can use this shorter form to define the same operation:
```

aio/content/guide/observables.md
```diff
@@ -116,4 +127,105 @@ myObservable.subscribe({
...
+
+| Area | Operators |
+| :------------| :----------|
+| Creation |  `from`,`fromEvent`, `of` |
+| Combination | `combineLatest`, `concat`, `merge`, `startWith` , `withLatestFrom`, `zip` |
+| Filtering | `debounceTime`, `distinctUntilChanged`, `filter`, `take`, `takeUntil` |
+| Transformation | `bufferTime`, `concatMap`, `map`, `mergeMap`, `scan`, `switchMap` |
+| Utility | `tap` |
+| Multicasting | `share` |
+
```
aikidave, line 221:
```suggestion
RxJS provides many operators. The most common operators used in Angular applications are as follows: 

| Area | Operators |
| :------------| :----------|
| Creation |  `from`,`fromEvent`, `of` |
| Combination | `combineLatest`, `concat`, `merge`, `startWith` , `withLatestFrom`, `zip` |
| Filtering | `debounceTime`, `distinctUntilChanged`, `filter`, `take`, `takeUntil` |
| Transformation | `bufferTime`, `concatMap`, `map`, `mergeMap`, `scan`, `switchMap` |
| Utility | `tap` |
| Multicasting | `share` |

For a full list of operators and usage samples, visit the [RxJS API documentation](https://rxjs.dev/api).
```

aio/content/guide/observables.md
```diff
@@ -116,4 +127,105 @@ myObservable.subscribe({
...
+
+## Naming conventions for observables
+
+Because Angular applications are mostly written in TypeScript, you will typically know when a variable is an observable. Although the Angular framework does not enforce a naming convention for observables, you will often see observables named with a trailing “$” sign.
```
aikidave, line 225:
```suggestion
Angular applications are mostly written in TypeScript, which makes it easier to know when a variable is an observable. However, a common naming convention for observables is to add a `$` to the variable name. For example, `response$`. Remember that the Angular framework does not enforce this convention.
```

aio/content/guide/observables.md
```diff
@@ -116,4 +127,105 @@ myObservable.subscribe({
...
+
+This can be useful when scanning through code and looking for observable values. Also, if you want a property to store the most recent value from an observable, it can be convenient to simply use the same name with or without the “$”.
```
aikidave, line 227:
```suggestion
Using this naming convention can help you when you are scanning through code and looking for observable values. 
```
I recommend deleting the last sentence. First we tell people about a convention, then we tell them it's not enforced, then we tell them they may or may not use it? I think it's fine to explain that it's a useful, if unenforced, convention, but that's it.

aio/content/guide/observables.md
jbogarthyde: I think we should keep the note about pipes vs chaining, since we refer to RxJS doc for operators, and their examples use chaining.

aio/content/guide/observables.md
jbogarthyde: No, it's a general descriptive term that Jason came up with. I'll try to rephrase.

aio/content/guide/observables.md
jbogarthyde: Rephrased to describe usage in example.

jbogarthyde:
Thanks @aikidave

aio/content/guide/observables.md
```diff
@@ -1,15 +1,22 @@
...
 
-The observer pattern is a software design pattern in which an object, called the *subject*, maintains a list of its dependents, called *observers*, and notifies them automatically of state changes.
-This pattern is similar (but not identical) to the [publish/subscribe](https://en.wikipedia.org/wiki/Publish%E2%80%93subscribe_pattern) design pattern.
+Angular makes use of and extends Reactive Extensions for JavaScript (RxJS), a library for [reactive programming](https://en.wikipedia.org/wiki/Reactive_programming "Wikipedia").
+RxJS provides an implementation of the `Observable` type, which makes it easier to compose asynchronous or callback-based code in TypeScript applications. The library also provides many useful [operators and utility functions](#rx-library "Using RxJS functions and operators") for working with observables.
+
+## How observables work
+
+Observables get their name from the observer _pattern_. In this pattern, an object (sometimes called the *subject*) maintains a list of its dependents, called *observers*, and notifies them automatically of state changes.
```
niklas-wortmann, line 10: I honestly think this might confuse many readers or at least moves them in the wrong direction as Subjects and Observables are fundamentally different in RxJS.

aio/content/guide/observables.md
```diff
@@ -1,15 +1,22 @@
...
 
-An observable can deliver multiple values of any type&mdash;literals, messages, or events, depending on the context. The API for receiving values is the same whether the values are delivered synchronously or asynchronously. Because setup and teardown logic are both handled by the observable, your application code only needs to worry about subscribing to consume values, and when done, unsubscribing. Whether the stream was keystrokes, an HTTP response, or an interval timer, the interface for listening to values and stopping listening is the same.
+An observable can deliver multiple values of any type&mdash;literals, messages, or events, depending on the context.
+Consumers can receive observable values synchronously or asynchronously.
+You subscribe to an observable to consume values, and unsubscribe when the execution completes or you no longer need to listen for new values.
+The interface for listening to values and stopping listening is the same, regardless of whether the value stream reports keystrokes, an HTTP response, or the state of an interval timer.
```
niklas-wortmann, line 19: "The interface for listening to values and stopping listening is the same" I don't understand this?

aio/content/guide/observables.md
```diff
@@ -35,33 +42,34 @@ A handler for receiving observable notifications implements the `Observer` inter
...
   path="observables/src/subscribing.ts"
   region="observer"
   header="Subscribe using observer"></code-example>
 
-Alternatively, the `subscribe()` method can accept callback function definitions in line, for `next`, `error`, and `complete` handlers. For example, the following `subscribe()` call is the same as the one that specifies the predefined observer:
+As an alternative to passing in an observer object, the `subscribe()` method can accept inline callback function definitions for `next`, `error`, and `complete` handlers, in that order. For example, the following `subscribe()` call is the same as the one that specifies the predefined observer:
 
 <code-example path="observables/src/subscribing.ts" region="sub_fn" header="Subscribe with positional arguments"></code-example>
 
 In either case, a `next` handler is required. The `error` and `complete` handlers are optional.
```
niklas-wortmann, line 69: Not part of this pr, but this sentence is actually wrong. One can also just call `subscribe()` without any callback

aio/content/guide/observables.md
```diff
@@ -71,43 +79,51 @@ For example, to create an observable equivalent to the `of(1, 2, 3)` above, you
...
+## Multicasting  values to multiple subscribers
 
-Sometimes, instead of starting an independent execution for each subscriber, you want each subscription to get the same values&mdash;even if values have already started emitting. This might be the case with something like an observable of clicks on the document object.
+*Multicasting* is the practice of broadcasting emitted values from a single observable execution to a list of multiple subscribers.
 
-*Multicasting* is the practice of broadcasting to a list of multiple subscribers in a single execution. With a multicasting observable, you don't register multiple listeners on the document, but instead re-use the first listener and send values out to each subscriber.
+A typical observable creates a new, independent execution for each subscribed observer.
+When an observer subscribes, the observable wires up an event handler and delivers values to that observer.
```
niklas-wortmann, line 95: Event handler might be a little bit irritating. I think what you are referring to is a Producer instance?

aio/content/guide/observables.md
```diff
@@ -71,43 +79,51 @@ For example, to create an observable equivalent to the `of(1, 2, 3)` above, you
...
+The following code changes the observable to use multicasting instead:
 
 <code-example path="observables/src/multicasting.ts" region="multicast_sequence" header="Create a multicast subscriber"></code-example>
 
-<div class="alert is-helpful">
-   Multicasting observables take a bit more setup, but they can be useful for certain applications. Later we will look at tools that simplify the process of multicasting, allowing you to take any observable and make it multicasting.
-</div>
+### Multicast with the _Subject_ class
+
+RxJS provides the [`Subject` class](https://rxjs.dev/guide/subject "RxJS API reference") that simplifies the process of multicasting, allowing you to take any observable and make it multicasting.
```
niklas-wortmann, line 117: Maybe it's just the fact that I am not a native speaker but "allowing you to take any observable and make it multicasting" kinda sounds like an Observable is a constructor argument for a Subject.

aio/content/guide/observables.md
```diff
@@ -71,43 +79,51 @@ For example, to create an observable equivalent to the `of(1, 2, 3)` above, you
...
+
+<code-example path="observables/src/multicasting-subject.ts" region="multicast_subject" header="Use a Subject for multicasting"></code-example>
 
 ## Error handling
 
-Because observables produce values asynchronously, try/catch will not effectively catch errors. Instead, you handle errors by specifying an `error` callback on the observer. Producing an error also causes the observable to clean up subscriptions and stop producing values. An observable can either produce values (calling the `next` callback), or it can complete, calling either the `complete` or `error` callback.
+An observable can either produce values (calling the `next` callback), or it can complete, calling either the `complete` or `error` callback.
```
niklas-wortmann, line 123: technically an Observable can produce values and complete afterwards or error and it can also do none of those :D

aio/content/guide/observables.md
```diff
@@ -71,43 +79,51 @@ For example, to create an observable equivalent to the `of(1, 2, 3)` above, you
...
+
+Because observables produce values asynchronously, the try/catch construction does not effectively catch errors. Instead, you generally handle errors by specifying an `error` callback on the observer.
```
niklas-wortmann, line 125: Values are not produced asynchronously, at least not if it isn't necessary. But errors are always emitted asynchronously

aio/content/guide/observables.md
```diff
@@ -116,4 +132,114 @@ myObservable.subscribe({
...
+<code-example path="rx-library/src/simple-creation.ts" region="interval" header="Create an observable from a counter"></code-example>
+
+<code-example path="rx-library/src/simple-creation.ts" region="event" header="Create an observable from an event"></code-example>
+
+<code-example path="rx-library/src/simple-creation.ts" region="ajax" header="Create an observable that creates an AJAX request"></code-example>
+
+### Operators
+
+Operators are functions that allow you to manipulate the values that observables return.
+For example, RxJS defines operators such as `map()`, `filter()`, `concat()`, and `flatMap()`.
```
niklas-wortmann, line 190: flatMap is deprecated, please use mergeMap

aio/content/guide/observables.md
jbogarthyde: What happens to the published values if there is no next handler? Should this option be mentioned?

aio/content/guide/observables.md
jbogarthyde: Is this true:  "It can also do none of these things, producing only side effects." 

 Is this why the "next" handler is not required? How would this work? Example?

Should there be a section discussing this usage? (Presumably it wouldn't go in the "Error handling" section.)

niklas-wortmann: yes it is true, but the next handler is not required just from an API design perspective, maybe you are not interested in the next values but just error or complete or you really just want to perform the side effects and therefore you just want to subscribe to an observable. 

I am not really sure if this should be covered in more detail as this is very specific and not really help with the overall understanding of rxjs. I just wanted to clarify this information here.

aio/content/guide/observables.md
niklas-wortmann: so if you have no handler registered it will not be handled, but produced. So in angular, especially with the `async`pipe I often see people doing any kind of side effect in `tap` and actually not subscribing manually anymore therefore I would like this to be corrected. I am not sure what level of information should be part of this document, but probably @aikidave has an opinion on this

jbogarthyde: This does seem a bit out of scope, especially this early in the description. I took out the optional/required language.

kapunahelewong (approved):
LGTM. Thank you, Judy!!

aio/content/guide/observables.md
```diff
@@ -1,15 +1,21 @@
...
-The subscribed consumer then receives notifications until the function completes, or until they unsubscribe.
+The subscribed consumer then receives notifications until the function completes or errors out, or until they unsubscribe.
 
-An observable can deliver multiple values of any type&mdash;literals, messages, or events, depending on the context. The API for receiving values is the same whether the values are delivered synchronously or asynchronously. Because setup and teardown logic are both handled by the observable, your application code only needs to worry about subscribing to consume values, and when done, unsubscribing. Whether the stream was keystrokes, an HTTP response, or an interval timer, the interface for listening to values and stopping listening is the same.
+An observable can deliver multiple values of any type&mdash;literals, messages, or events, depending on the context.
+Consumers can receive observable values synchronously or asynchronously.
+You subscribe to an observable to consume values, and unsubscribe when the execution completes or you no longer need to listen for new values.
```
alxhub, line 18: I think it's worth expanding on the concepts here, so I've taken a stab at this:

You subscribe to an observable to begin receiving values from it, and can unsubscribe when you no longer need to listen for new values. Some observables (such as those from `HttpClient`) are finite, and will unsubscribe automatically upon completion.

aio/content/guide/observables.md
```diff
@@ -35,33 +41,32 @@ A handler for receiving observable notifications implements the `Observer` inter
 
 An observer object can define any combination of these handlers. If you don't supply a handler for a notification type, the observer ignores notifications of that type.
 
-## Subscribing
+## Subscribing to observables
 
-An `Observable` instance begins publishing values only when someone subscribes to it. You subscribe by calling the `subscribe()` method of the instance, passing an observer object to receive the notifications.
+An `Observable` instance begins publishing values only when you subscribe to it.
```
alxhub, line 46: It may be worth including here that subscribing may cause the observable to take some action to begin producing values - for example, subscribing to an `HttpClient` observable causes an HTTP request to be made.

aio/content/guide/observables.md
```diff
@@ -35,33 +41,32 @@ A handler for receiving observable notifications implements the `Observer` inter
...
 
-Here's an example of creating and subscribing to a simple observable, with an observer that logs the received message to the console:
-
 <code-example
   path="observables/src/subscribing.ts"
   region="observer"
   header="Subscribe using observer"></code-example>
 
-Alternatively, the `subscribe()` method can accept callback function definitions in line, for `next`, `error`, and `complete` handlers. For example, the following `subscribe()` call is the same as the one that specifies the predefined observer:
+As an alternative to passing in an observer object, the `subscribe()` method can accept inline callback function definitions for `next`, `error`, and `complete` handlers, in that order. For example, the following `subscribe()` call is the same as the one that specifies the predefined observer:
```
alxhub, line 64: Is it worth switching this example around? In my experience subscribing using an observer is fairly uncommon, and the majority of applications will use the inline form.

aio/content/guide/observables.md
```diff
@@ -71,43 +76,51 @@ For example, to create an observable equivalent to the `of(1, 2, 3)` above, you
...
 
 <code-example path="observables/src/creating.ts" region="fromevent" header="Create with custom fromEvent function"></code-example>
 
-Now you can use this function to create an observable that publishes keydown events:
+You can use this function to create an observable that publishes keydown events, as follows:
 
 <code-example path="observables/src/creating.ts" region="fromevent_use" header="Use custom fromEvent function"></code-example>
 
-## Multicasting
+## Multicasting  values to multiple subscribers
```
alxhub, line 87:
```suggestion
## Multicasting values to multiple subscribers
```

aio/content/guide/observables.md
```diff
@@ -71,43 +76,51 @@ For example, to create an observable equivalent to the `of(1, 2, 3)` above, you
...
 
-*Multicasting* is the practice of broadcasting to a list of multiple subscribers in a single execution. With a multicasting observable, you don't register multiple listeners on the document, but instead re-use the first listener and send values out to each subscriber.
+Sometimes, instead of starting an independent execution for each subscriber, you want each subscription to get the same values&mdash;even if values have already started emitting.
+For example, you might want to create an observable of clicks on a document object.
+With a multicasting observable, you don't register multiple listeners on the document, but instead re-use the first listener and send values out to each subscriber.
 
-When creating an observable you should determine how you want that observable to be used and whether or not you want to multicast its values.
-
-Let’s look at an example that counts from 1 to 3, with a one-second delay after each number emitted.
+Consider the following example that counts from 1 to 3, with a one-second delay after each number emitted.
```
alxhub, line 99: It might be valuable to use `HttpClient` observables instead, since this offers a real-world example of when multicasting might be valueable.

A single HttpClient observable, subscribed twice -> two separate HTTP requests
A single HttpClient observable, multicasted and subscribed twice -> a single HTTP request with two observers

aio/content/guide/observables.md
```diff
@@ -71,43 +76,51 @@ For example, to create an observable equivalent to the `of(1, 2, 3)` above, you
...
 
 <code-example path="observables/src/multicasting.ts" region="delay_sequence" header="Create a delayed sequence"></code-example>
 
-Notice that if you subscribe twice, there will be two separate streams, each emitting values every second. It looks something like this:
+When creating the observable, you should determine how you want to use that observable and whether you want to multicast its values.
```
alxhub, line 103: It's not actually when creating the observable, but when subscribing. You might want to multicast observables that you didn't create yourself (e.g. those from `HttpClient`).

aio/content/guide/observables.md
```diff
@@ -71,43 +76,51 @@ For example, to create an observable equivalent to the `of(1, 2, 3)` above, you
...
+If you subscribe twice, notice that there are two separate streams, each emitting values every second.
 
 <code-example path="observables/src/multicasting.ts" region="subscribe_twice" header="Two subscriptions"></code-example>
 
- Changing the observable to be multicasting could look something like this:
+The following code changes the observable to use multicasting instead:
```
alxhub, line 108: Looking through the examples here, I think unfortunately they all really miss the point of multicasting.

The example actually has nothing to do with multicasting. It defines a custom "hot" observable which treats subsequent subscribers differently than the first one. This is not how most applications would make use of multicasting.

Most applications will deal with multicasting by using operators to share a single (unicast?) observable subscription among multiple subscribers, typically with the `multicast` or `share` operators (with which one can choose a specific `Subject` implementation to fine tune the behavior).

So a better set of examples would be:

1. One showing the problem of duplicate requests when subscribing to an HTTP request observable in two places
2. Using the `share()` operator to turn the observable into a multicast observable, ensuring that only one request is made for both subscriptions
3. Using the `shareReplay()` operator to ensure that late subscribers still receive the response, even after the HTTP request is complete.

aio/content/guide/observables.md
```diff
@@ -116,4 +129,114 @@ myObservable.subscribe({
...
+
+For instance, suppose you have an observable that makes an API request and maps to the response from the server. If the server returns an error or the value doesn’t exist, an error is produced. If you catch this error and supply a default value, your stream continues to process values rather than erroring out.
+
+The following example demonstrates the `catchError` operator:
+
+<code-example path="rx-library/src/error-handling.ts" header="catchError operator"></code-example>
+
+### Retrying a failed observable
+
+Where the `catchError` operator provides a simple path of recovery, the [`retry` operator](https://rxjs.dev/api/operators/retry "RxJS API reference") lets you retry a failed request.
```
alxhub, line 142:
```suggestion
Whereas the `catchError` operator provides a simple path of recovery, the [`retry` operator](https://rxjs.dev/api/operators/retry "RxJS API reference") lets you retry a failed request.
```

aio/content/guide/observables.md
```diff
@@ -116,4 +129,114 @@ myObservable.subscribe({
 });
 </code-example>
 
-Error handling (and specifically recovering from an error) is covered in more detail in a later section.
+In addition to the `error()` handler that you provide on subscription, the RxJS library provides the [`catchError` operator](https://rxjs.dev/api/operators/catchError "RxJS API reference") that lets you handle known errors in the observable code.
+
+For instance, suppose you have an observable that makes an API request and maps to the response from the server. If the server returns an error or the value doesn’t exist, an error is produced. If you catch this error and supply a default value, your stream continues to process values rather than erroring out.
```
alxhub, line 134:
```suggestion
For instance, suppose you have an observable that makes an API request and maps to the response from the server. If the server returns an error or the value doesn’t exist, an error is produced. If you catch this error and supply a default value, your stream continues to process values rather than propagating the error.
```

aio/content/guide/observables.md
```diff
@@ -116,4 +129,114 @@ myObservable.subscribe({
...
+
+The following example demonstrates the `catchError` operator:
+
+<code-example path="rx-library/src/error-handling.ts" header="catchError operator"></code-example>
+
+### Retrying a failed observable
+
+Where the `catchError` operator provides a simple path of recovery, the [`retry` operator](https://rxjs.dev/api/operators/retry "RxJS API reference") lets you retry a failed request.
+
+Use the `retry` operator before the `catchError` operator. It resubscribes to the original source observable, which can then re-run the full sequence of actions that resulted in the error. If this includes an HTTP request, it will retry that HTTP request.
```
alxhub, line 144: It might be worth mentioning that retrying works by re-subscribing to the original observable, which e.g. for HTTP requests, will result in a new request being fired.

That's also why it's important to `retry()` before `catchError()` or any multicasting operations like `share()`.

aio/content/guide/observables.md
```diff
@@ -116,4 +129,114 @@ myObservable.subscribe({
...
+Operators are functions that allow you to manipulate the values that observables return.
+For example, RxJS defines operators such as `map()`, `filter()`, `concat()`, and `mergeMap()`.
+
+Operators take configuration options, and they return a function that takes a source observable.
+When executing this returned function, the operator observes the source observable’s emitted values, transforms them, and returns a new observable of those transformed values.
+For example:
+
+<code-example path="rx-library/src/operators.ts" header="Map operator"></code-example>
+
+You can use _pipes_ to link operators together. Pipes let you combine multiple functions into a single function.
```
alxhub, line 195: This is a little confusingly worded. There is no named concept of a "pipe" in Rx. `pipe()` is the name of the method we use to transform an observable via one or more operators.

aio/content/guide/observables.md
```diff
@@ -116,4 +129,114 @@ myObservable.subscribe({
...
+The `pipe()` function takes as its arguments the functions you want to combine, and returns a new function that, when executed, runs the composed functions in sequence.
+
+<div class="alert is-helpful">
+
+ For Angular applications, it is recommended that you combine operators with pipes (rather than _chaining_, which is used in many RxJS examples).
```
alxhub, line 200:
```suggestion
 For Angular applications, it is recommended that you combine operators with `pipe()` (rather than _chaining_, which was used in older versions of RxJS).
```

aio/content/guide/observables.md
```diff
@@ -116,4 +129,114 @@ myObservable.subscribe({
...
+
+</div>
+
+A set of operators applied to an observable is a set of instructions for producing the values you’re interested in.
+The operators require input from the execution of an observable.
+You need to call `subscribe()` to execute the observable and produce a result from the operators.
+
+Here’s an example:
+
+<code-example path="rx-library/src/operators.1.ts" header="Standalone pipe function"></code-example>
```
alxhub, line 210: Nobody uses the standalone `pipe` function - we should either switch these examples, or just not teach the standalone way at all.

aio/content/guide/observables.md
```diff
@@ -116,4 +129,114 @@ myObservable.subscribe({
...
+RxJS provides many operators. The following are commonly used in Angular applications:
+
+
+| Area | Operators |
+| :------------| :----------|
+| Creation |  `from`,`fromEvent`, `of` |
+| Combination | `combineLatest`, `concat`, `merge`, `startWith` , `withLatestFrom`, `zip` |
+| Filtering | `debounceTime`, `distinctUntilChanged`, `filter`, `take`, `takeUntil` |
+| Transformation | `bufferTime`, `concatMap`, `map`, `mergeMap`, `scan`, `switchMap` |
+| Utility | `tap` |
```
alxhub, line 227: It might actually be useful to have a section on debugging RxJS code, where `tap` can come in handy for logging or adding breakpoints during an RxJS sequence.

aio/content/guide/observables.md
```diff
@@ -116,4 +129,114 @@ myObservable.subscribe({
...
+For a full list of operators and usage samples, visit the [RxJS API documentation](https://rxjs.dev/api).
+
+
+## Naming conventions for observables
+
+Angular applications are mostly written in TypeScript, which makes it easier to know when a variable is an observable.
+However, a common naming convention for observables is to add a `$` to the variable name; for example, `response$`.
+Remember that the Angular framework does not enforce this convention.
+
+Using this naming convention can help you when you are scanning through code and looking for observable values, or when you want to associate an observable with a property that stores its returned values.
```
alxhub, line 239: I've also found it useful when you have both an Observable and its value in the same namespace. For example, within a template, the following might be common:
```html
<div *ngFor="let user of users$ | async as users">...</.div>
```
where `users$` is the Observable that's passed to the `async` pipe, and the unwrapped/emitted value is aliased as `users`. This saves you from having to get creative with naming for both the sync and async versions of a value.

alxhub (changes requested):

aio/content/guide/http.md
```diff
@@ -165,7 +165,7 @@ Next, specify that interface as the `HttpClient.get()` call's type parameter in
 
 <div class="alert is-helpful">
 
- When you pass an interface as a type parameter to the `HttpClient.get()` method, you can use the [RxJS `map` operator](guide/rx-library#operators) to transform the response data as needed by the UI. You can then pass the transformed data to the [async pipe](api/common/AsyncPipe).
+ When you pass an interface as a type parameter to the `HttpClient.get()` method, you can use the [RxJS `map` operator](guide/observables#operators) to transform the response data as needed by the UI. You can then pass the transformed data to the [async pipe](api/common/AsyncPipe).
```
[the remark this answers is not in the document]
jbogarthyde: massaged it a little -- not sure if it makes sense to shoehorn in the part about what to do with the response data - moved it out of the type discussion.

aio/content/guide/observables.md
jbogarthyde: I added the description of the more realistic case, but left in the old example that shows how to make the observable multicasting.  

Can you make the observable returned by an `HttpClient` call multicasting? 
Again, the example came from someone else, and I can't create an equivalent -- do you want to try?

aio/content/guide/observables.md
jbogarthyde: Can you or someone please supply these examples?

aio/content/guide/observables.md
jbogarthyde: Let me know if you can provide source for such a section.

vhnguyen28 (approved):

aikithoughts (changes requested):
Need to examine these changes more thoroughly. Thanks!

PR #38168, opened 2020-07-21
refactor(dev-infra): update commit body checks

### LINKED ISSUE #37865: Commit body is calculated incorrectly in some cases locally
While investigating the problem with `docs` scope exclusion from min body length checks (see https://github.com/angular/angular/pull/37862), I noticed that the commit body that we use for length check also includes the content that would be dropped after the commit is made. For example when I make some change, run the `git commit` command and presented with the screen to type commit message like this:

```
docs: some test

# Please enter the commit message for your changes. Lines starting
# with '#' will be ignored, and an empty message aborts the commit.
#
# On branch docs-test
# Changes to be committed:
#	modified:   aio/content/marketing/resources.json
#
# Changes not staged for commit:
#	modified:   .ng-dev/config.ts
#
```

Everything below the first 2 lines (all lines that start with `#`) are treated as commit body, thus passing the min length validation even though it would later fail on CI.

The proposal is to process commit body before performing the length check and remove all lines that start with `#`.

AndrewKushnir:
This commit updates the commit body checking logic:
* avoid checking the length of lines in commit body
* ignores lines that start with `#` char, since these lines are skipped by git
* adds `release` into the `minBodyLengthTypeExcludes` list in a config rather than having a hard-coded exclusion for this commit type in the code

Resolves #37865.

## PR Type

- [ ] Yes
- [x] No

### DIFF
--- a/dev-infra/commit-message/validate.ts
+++ b/dev-infra/commit-message/validate.ts
@@ -22,7 +22,6 @@ const REVERT_PREFIX_RE = /^revert:? /i;
 const TYPE_SCOPE_RE = /^(\w+)(?:\(([^)]+)\))?\:\s(.+)$/;
 const COMMIT_HEADER_RE = /^(.*)/i;
 const COMMIT_BODY_RE = /^.*\n\n([\s\S]*)$/;
-const COMMIT_BODY_URL_LINE_RE = /^https?:\/\/.*$/;
 
 /** Parse a full commit message into its composite parts. */
 export function parseCommitMessage(commitMsg: string) {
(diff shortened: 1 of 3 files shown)


jelbourn (approved):
lgtm

devversion (approved):
LGTM. Nice cleanup!

dev-infra/commit-message/validate.ts
```diff
@@ -140,33 +139,23 @@ export function validateCommitMessage(
...
-    printError(
-        `The commit message body contains lines greater than ${config.maxLineLength} characters`);
-    return false;
+  if (!config.minBodyLengthTypeExcludes?.includes(commit.type)) {
+    const bodyByLine = commit.bodyWithoutLinking.trim().split('\n');
+    const actualBodyLength =
+        bodyByLine
+            // Exclude lines that start with `#`, since these lines are ignored by git.
+            .filter(line => !line.startsWith('#'))
+            .join('')
```
gkalpak, line 152: Why not join on `'\n'` (since that is what you split on)?
Also, I would suggest `trim()`ing (as we did before).

gkalpak (commented):
FWIW, I am not in favor of removing the max line length limit, but otherwise lgtm :grin:

dev-infra/commit-message/validate.ts
AndrewKushnir: Thanks for the review @gkalpak.

> Why not join on '\n' (since that is what you split on)?

Joining on `\n` will add 2 chars to the body for each line, so 10+ empty lines would be treated as a valid commit message body.

> Also, I would suggest trim()ing (as we did before).

We didn't trim lines before, we just trimmed the whole body contents and this logic is present now as well.

josephperrott (changes requested):
A couple of things:

Can you split the commits into one that contains the changes to the `.ng-dev/commit-messages.ts` file and one that contains the rest.

The commit scope of the config file change should be `ci` while the others scope should be `dev-infra`


---

Additionally, I would like to push back on the line length removal?  I don't understand the historical context for why it was in place and I want to be sure that we aren't having unintended/misunderstood consequences.

Do we know why it was in place in the first place?

dev-infra/commit-message/validate.ts
gkalpak: > Joining on `\n` will add 2 chars to the body for each line, so 10+ empty lines would be treated as a valid commit message body.

While this is partially true (if you trim the body as suggested above, then one would need to have consecutive empty lines between actual text), if someone did this then it sounds intentional enough (and if they wanted to trick the validation script, they could have used twenty dots or something instead).

Also, not removing the nwlines is the behavior we had before, so if this is an intentional change/improvement, it should go on its own commit (with an explanation of the reasoning).

<sub>(For the record, `\n` adds one characters 🤓)</sub>

> We didn't trim lines before, we just trimmed the whole body contents and this logic is present now as well.

Yes, I am not suggesting trimming individual lines. I am suggesting trimming the body _after_ removing the comments. The intention was to remove extra space around the actual commit message body, so we should trim it after having remove the comments to make sure that any extra space between the actual body and the comments is also removed.

AndrewKushnir: Thanks for additional clarification @gkalpak 👍
I've applied the necessary changes (and added a test to verify the behavior).

gkalpak:
If you are curious, this is how long commit messages are displayed on GitHub's PR commit tab 😛:

long-commit-message

PR #38170, opened 2020-07-22
ci: add more owners for limited categories

jelbourn:
* Add alxhub, atscott, and AndrewKushnir to code owners
* Add atscott & AndrewKushnir to public-api and size-tracking
* Change public-api and size-tracking from requesting `-1` (all) to requesting four (still requiring three and two, respectively)

Follow-up to #37994

### DIFF
--- a/.pullapprove.yml
+++ b/.pullapprove.yml
@@ -1175,14 +1175,16 @@ groups:
           ])
     reviewers:
       users:
+        - AndrewKushnir
         - IgorMinar
         - alxhub
+        - atscott
         - jelbourn
         - petebacondarwin
         - pkozlowski-opensource
     reviews:
-      request: -1   # request reviews from everyone
-      required: 3   # require at least 3 approvals
+      request: 4 # Request reviews from four people
+      required: 3 # Require that three people approve
       reviewed_for: required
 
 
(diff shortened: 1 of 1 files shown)


.pullapprove.yml
```diff
@@ -1155,13 +1155,15 @@ groups:
...
+        - AndrewKushnir
         - IgorMinar
         - alxhub
+        - atscott
         - jelbourn
         - petebacondarwin
         - pkozlowski-opensource
     reviews:
-      request: -1   # request reviews from everyone
+      request: 4   # request reviews from everyone
```
atscott, line 1166: This comment is no longer accurate.

.pullapprove.yml
```diff
@@ -1155,13 +1155,15 @@ groups:
...
       required: 3   # require at least 3 approvals
```
atscott, line 1167: This comment seems a little redundant

.pullapprove.yml
```diff
@@ -1179,13 +1181,15 @@ groups:
...
         - IgorMinar
         - alxhub
+        - atscott
         - jelbourn
         - petebacondarwin
         - pkozlowski-opensource
     reviews:
-      request: -1   # request reviews from everyone
+      request: 4   # request reviews from everyone
       required: 2   # require at least 2 approvals
```
atscott, line 1193:
```suggestion
      request: 4
      required: 2
```

atscott (approved):

AndrewKushnir (approved):
LGTM, thanks @jelbourn 👍 

I'm wondering if we can do similar thing for `circular-dependencies` group as well (request 3-4 and require 2 approvals)?

.pullapprove.yml
josephperrott: I would prefer to over comment throughout this file as we have found historically that its hard to remember/reason about what exactly every setting does and how/ordering PullApprove parses the file.

jelbourn:
Comments addressed

.pullapprove.yml
jelbourn: I added the comments back, though I do agree they're a bit redundant when the values are positive numbers

josephperrott (approved):
LGTM

PR #38173, opened 2020-07-22
build(docs-infra): improve and simplify docs examples boilerplate tooling

gkalpak:
This commit removes some obsolete files from the docs examples boilerplate, ensures some boilerplate files are copied over correctly and simplifies the boilerplate tooling to not require explicitly listing boilerplate files in `example-boilerplate.js`.

See individual commits for more details.


petebacondarwin (approved):
We agreed offline to remove the first commit and fix a typo in the message of the 3rd.
Otherwise LGTM!

PR #38192, opened 2020-07-23
build(docs-infra): simplify `ExampleZipper` by removing `PackageJsonCustomizer`

gkalpak:
Previously, `ExampleZipper` (the tool used for creating ZIP archives from our docs examples) used the `PackageJsonCustomizer` to generate `package.json` files for each example type. This had the following drawbacks:
- The generated files had to be kept up-to-date with the corresponding boilerplate files in `aio/tools/examples/shared/boilerplate/` and there was no easy way to find out when the files got out-of-sync.
- The `PackageJsonCustomizer` logic was non-trivial and difficult to reason about.
- The same information was duplicated in the boilerplate files and the customizer configuration files.

This setup was useful when we used a single `package.json` file for all docs examples. Now, however, each example type can have its own boilerplate `package.json` file, including scripts and dependencies relevant to the example type. Therefore, it is no longer necessary to generate `package.json` files for ZIP archives.

This commit eliminates the drawbacks mentioned above and simplifies the `ExampleZipper` tool by removing `PackageJsonCustomizer` and re-using the boilerplate `package.json` files for ZIP archives.

The changes in this commit also fix some ZIP archives that were previously broken (for example due to missing dependencies).

##
**UPDATE:**

In order to make it easier to see how the changes in this PR affect the final ZIP archives, I pushed two temporary commits which contained the `package.json` files from all ZIP archives:
- Before this PR (i.e. with `PackageJsonCustomizer`): ea5cf9b6b25c321eaa48848c538c4d30b51484f2
- After this PR (i.e. without `PackageJsonCustomizer`): 1afa5b9fcda656a758a4e45a85e72040aa64d134

Basically, one can see the resulting changes by looking at the latter (1afa5b9fcda656a758a4e45a85e72040aa64d134).
I looked through all changes and (unless I missed something) all changed `package.json` files contain the same changes as for example [accessibility/package.json](https://github.com/angular/angular/pull/38192/commits/1afa5b9fcda656a758a4e45a85e72040aa64d134#diff-c7c7cf4e524feac0de7bc88683ee8b1e) with the exception of [universal/package.json](https://github.com/angular/angular/pull/38192/commits/1afa5b9fcda656a758a4e45a85e72040aa64d134#diff-1d9b6695754ee6b9dedcc09e349b629f) and [upgrade-lazy-load-ajs/package.json](https://github.com/angular/angular/pull/38192/commits/1afa5b9fcda656a758a4e45a85e72040aa64d134#diff-8f7e2699afc4f026e6d94fb8285da227), which are slightly different.


aio/tools/examples/shared/boilerplate/cli/package.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "angular.io-example",
   "version": "0.0.0",
+  "description": "Example project from an angular.io guide.",
```
gkalpak, line 4: This is taken from the removed [aio/tools/example-zipper/customizer/package-json/package.json][1] file, which was used as the basis for generated `package.json` files in ZIP archives.

[1]: https://github.com/angular/angular/blob/3c1866b5bb1ffc3aabef1646e55b2f888a928caf/aio/tools/example-zipper/customizer/package-json/package.json#L5

aio/tools/examples/shared/boilerplate/systemjs/package.json
```diff
@@ -1,8 +1,8 @@
 {
-  "name": "angular-examples",
-  "version": "1.0.0",
-  "private": true,
-  "description": "Example package.json, only contains needed scripts for examples. See _examples/package.json for master package.json.",
+  "name": "angular.io-example",
```
gkalpak, line 2: The changes in this file are to bring it closer to the format used for other project types (for example [aio/tools/examples/shared/boilerplate/cli/package.json][1]).

[1]: https://github.com/angular/angular/blob/cdf29edda1c06ad1adc9e8fc86b1f527f787fb9f/aio/tools/examples/shared/boilerplate/cli/package.json

kapunahelewong (approved):
LGTM!

PR #38205, opened 2020-07-23
feat(dev-infra): allow prefixed versions in git tags for build-env-stamp

josephperrott:
Adding support for prefixes of versions for git tags will allow for
angular/angular-cli to utilize the merge tooling provided by ng-dev.
This change allows for a configuration to allow for proper discovery
of the current version for determining target patch branches.

### DIFF
--- a/dev-infra/release/config.ts
+++ b/dev-infra/release/config.ts
@@ -1,0 +1,24 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {assertNoErrors, getConfig, NgDevConfig} from '../utils/config';
+
+export interface ReleaseConfig {
+  tagPrefix?: string;
+}
+
+/** Retrieve and validate the config as `ReleaseConfig`. */
+export function getReleaseConfig() {
+  // List of errors encountered validating the config.
+  const errors: string[] = [];
+  // The unvalidated config object.
+  const config: Partial<NgDevConfig<{release: ReleaseConfig}>> = getConfig();
+
+  assertNoErrors(errors);
+  return config as Required<typeof config>;
+}
--- a/dev-infra/release/env-stamp.ts
+++ b/dev-infra/release/env-stamp.ts
@@ -8,6 +8,8 @@
 
 import {exec as _exec} from '../utils/shelljs';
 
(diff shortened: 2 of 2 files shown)


devversion:
Is this actually a symptom of us using the Git branches for stamping. I'd still advocate for determining the release version from the project `package.json` as that enforces hermeticity and predictable stamping. Right now, stamping is reliant on external tags in the remote, while all the information we need is actually provided in the project `package.json` unless I miss something here.

e.g. look at how we do it within the components repo. I've brought this up a couple of times, but we decided to defer the discussion. Maybe it would be a good time thinking about this now? especially since we deal with branching/versioning as a larger project currently? https://github.com/angular/components/blob/master/tools/bazel-stamp-vars.js

dev-infra/release/env-stamp.ts
```diff
@@ -19,12 +21,15 @@ import {exec as _exec} from '../utils/shelljs';
...
+  /** The configuration for the release tooling of ng-dev. */
+  const config = getReleaseConfig();
+
   console.info(`BUILD_SCM_BRANCH ${getCurrentBranch()}`);
   console.info(`BUILD_SCM_COMMIT_SHA ${getCurrentSha()}`);
   console.info(`BUILD_SCM_HASH ${getCurrentSha()}`);
   console.info(`BUILD_SCM_LOCAL_CHANGES ${hasLocalChanges()}`);
   console.info(`BUILD_SCM_USER ${getCurrentGitUser()}`);
-  console.info(`BUILD_SCM_VERSION ${getSCMVersion()}`);
+  console.info(`BUILD_SCM_VERSION ${getSCMVersion(config)}`);
```
devversion, line 32: Nit: Since you read the config as a default parameter value, this could be avoided.

dev-infra/release/config.ts
```diff
@@ -0,0 +1,24 @@
...
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {assertNoErrors, getConfig, NgDevConfig} from '../utils/config';
+
+export interface ReleaseConfig {
+  tagPrefix?: string;
```
devversion, line 12: If we go with that approach, we should document this property and what it does.

josephperrott:
@devversion, I think you are right that now is the time we should decide on what we are doing to unify on our approach.

PR #38210, opened 2020-07-24
Detect and print cycles that cause `setComponentScope()` to be used

dgp1130:
This detects and prints any cycles which cause `setComponentScope()` to be used and deoptimize tree-shakeability.

This patch serves as a work-around for #38211.


JoostK:
I believe this PR has become obsolete since the changes in #40678. There's no dedicated configuration option for how to deal with cycles yet, but using `compilationMode: 'partial'` will error when cycles are present.

PR #38213, opened 2020-07-24
fix(compiler): share identical stylesheets between components in the same file

### LINKED ISSUE #38203: ConstantPool timing issue with multiple components in a file
`ConstantPool` deduplicates constants by returning a particular kind of `o.Expression` called a `FixupExpression`, to which it retains a reference. For the first usage, this `FixupExpression` transparently returns the original `o.Expression`, so nothing special happens.

Whenever the `ConstantPool` encounters a second usage of the same value, however, it creates a shared variable declaration for the value, and uses its reference to the `FixupExpression` to replace its contents with a reference to this variable instead. Thus, the original usage is updated to refer to the constant instead of the literal value.

This works as long as all potential constants pass through the `ConstantPool` _before_ any of the `o.Expression` outputs are processed. However, in ngtsc the final compilation of classes violates this condition.

This compilation happens in a TypeScript transform (in `ngtsc/transform/src/transform.ts`, within `IvyVisitor`'s `visitClassDeclaration`). Each class goes through a process of `o.Expression` generation (which includes `ConstantPool` wrapping), then reification of `o.Expression`s into `ts.Expression`s via `translateExpression`, and then updating of the `ts.ClassDeclaration` to add the new static properties. Because `o.Expression`s are immediately converted to `ts.Expression`s, any `FixupExpression`s produced by the `ConstantPool` are converted **as they are** immediately following processing of that single class. Any future deduplication based on usage of the same constants by other classes will not affect them.

In practice, this means that such shared constants are always duplicated for the first class that uses them, and all subsequent classes use the shared variable declaration. This only affects constant reuse _between_ classes - constant reuse within the same class works normally.

### LINKED ISSUE #38204: Share identical stylesheets between components in the same file
In rare cases, multiple components can share the same styles. For example, Material does this with `MatButton` and `MatAnchor`.

View Engine had a feature which converted `styleUrls` into `.ngstyle.js` classes with exported constants, and thus naturally deduplicated reuse of external stylesheets. Ivy does not perform this optimization by design - it was a design goal to remove all additional "shim" files.

A limited form of this optimization in Ivy is possible - within a single file, it should be possible to use the `ConstantPool` to deduplicate stylesheet text that is shared between multiple components. This would be useful both for external stylesheets as well as inline style text, as long as it's identical. This optimization is of course conditional on `ViewEncapsulation` - encapsulated and un-encapsulated styles will have different contents, and thus no reuse between them is possible.

This optimization would apply to the Material example above, since `MatButton` and `MatAnchor` are both declared in the same TS file.

AndrewKushnir:
This PR consists of 3 separate commits and updates the compiler to allow identical stylesheets to be reused between components in the same file. This is achieved by allowing strings to be handled in `ConstantPool` and actually using `ConstantPool` while generating `styles` array on component def. More information can be found in individual commit messages.

Resolves #38204.
Resolves #38203.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/compiler-cli/src/ngtsc/transform/src/transform.ts
+++ b/packages/compiler-cli/src/ngtsc/transform/src/transform.ts
@@ -86,13 +86,15 @@ class IvyTransformationVisitor extends Visitor {
       VisitListEntryResult<ts.Statement, ts.ClassDeclaration> {
     // If this class is not registered in the map, it means that it doesn't have Angular decorators,
     // thus no further processing is required.
-    if (!this.classCompilationMap.has(node)) return {node};
+    if (!this.classCompilationMap.has(node)) {
+      return {node};
+    }
 
     // There is at least one field to add.
     const statements: ts.Statement[] = [];
     const members = [...node.members];
 
-    this.classCompilationMap.get(node)!.forEach(field => {
+    for (const field of this.classCompilationMap.get(node)!) {
       // Translate the initializer for the field into TS nodes.
       const exprNode = translateExpression(
           field.initializer, this.importManager, this.defaultImportRecorder,
@@ -120,7 +122,7 @@ class IvyTransformationVisitor extends Visitor {
           .forEach(stmt => statements.push(stmt));
 
       members.push(property);
-    });
+    }
 
     // Replace the class declaration with an updated version.
     node = ts.updateClassDeclaration(
--- a/packages/compiler-cli/src/ngtsc/transform/src/transform.ts
+++ b/packages/compiler-cli/src/ngtsc/transform/src/transform.ts
@@ -14,6 +14,7 @@ import {Decorator, ReflectionHost} from '../../reflection';
 import {ImportManager, translateExpression, translateStatement} from '../../translator';
 import {visit, VisitListEntryResult, Visitor} from '../../util/src/visitor';
 
+import {CompileResult} from './api';
 import {TraitCompiler} from './compilation';
 import {addImports} from './utils';
 
@@ -43,12 +44,15 @@ export function ivyTransformFactory(
   };
 }
 
-class IvyVisitor extends Visitor {
-  constructor(
-      private compilation: TraitCompiler, private reflector: ReflectionHost,
-      private importManager: ImportManager, private defaultImportRecorder: DefaultImportRecorder,
-      private isClosureCompilerEnabled: boolean, private isCore: boolean,
-      private constantPool: ConstantPool) {
+/**
+ * Visits all classes, performs Ivy compilation where Angular decorators are present and collects
+ * result in a Map that associates a ts.ClassDeclaration with Ivy compilation results. This visitor
+ * does NOT perform any TS transformations.
+ */
+class IvyCompilationVisitor extends Visitor {
+  public classCompilationMap = new Map<ts.ClassDeclaration, CompileResult[]>();
+
+  constructor(private compilation: TraitCompiler, private constantPool: ConstantPool) {
     super();
   }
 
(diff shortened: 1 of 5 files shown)


packages/compiler-cli/src/ngtsc/transform/src/transform.ts
```diff
@@ -43,69 +44,94 @@ export function ivyTransformFactory(
...
+      private defaultImportRecorder: DefaultImportRecorder,
+      private isClosureCompilerEnabled: boolean, private isCore: boolean) {
+    super();
+  }
+
+  visitClassDeclaration(node: ts.ClassDeclaration):
+      VisitListEntryResult<ts.Statement, ts.ClassDeclaration> {
+    // If this class is not registered in the map, it means that it doesn't have Angular decorators,
+    // thus no further processing is required.
+    if (!this.classCompilationMap.has(node)) return {node};
```
alxhub, line 89: Style in the compiler is to always use braces:
```suggestion
    if (!this.classCompilationMap.has(node)) {
      return {node};
    }
```

packages/compiler-cli/src/ngtsc/transform/src/transform.ts
```diff
@@ -43,69 +44,94 @@ export function ivyTransformFactory(
...
+
+    // There is at least one field to add.
+    const statements: ts.Statement[] = [];
+    const members = [...node.members];
+
+    this.classCompilationMap.get(node)!.forEach(field => {
```
alxhub, line 95:
```suggestion
    for (const field of this.classCompilationMap.get(node)!) {
```

packages/compiler/src/constant_pool.ts
```diff
@@ -305,3 +313,8 @@ function invalid<T>(this: o.ExpressionVisitor, arg: o.Expression|o.Statement): n
 function isVariable(e: o.Expression): e is o.ReadVarExpr {
   return e instanceof o.ReadVarExpr;
 }
+
+function isLongStringExpr(expr: o.LiteralExpr): boolean {
+  return typeof expr.value === 'string' &&
+      expr.value.length > POOL_INCLUSION_LENGTH_THRESHOLD_FOR_STRINGS;
```
alxhub, line 319: Typically I would expect a "threshold" constraint to be satisfied on the boundary:

```suggestion
      expr.value.length >= POOL_INCLUSION_LENGTH_THRESHOLD_FOR_STRINGS;
```

alxhub (approved):
Thanks, this looks great - just a couple comments but otherwise approved!

AndrewKushnir:
Thanks for the review @alxhub, I've applied the necessary changes.

alxhub (approved):

jelbourn (approved):
LGTM

Reviewed-for: size-tracking

petebacondarwin (approved):
Reviewed-for: size-tracking

AndrewKushnir:
[Presubmit](http://test/OCL:323029519:BASE:323029547:1595613782401:38188e63)

packages/compiler-cli/src/ngtsc/transform/src/transform.ts
```diff
@@ -43,69 +44,96 @@ export function ivyTransformFactory(
...
-      private compilation: TraitCompiler, private reflector: ReflectionHost,
-      private importManager: ImportManager, private defaultImportRecorder: DefaultImportRecorder,
-      private isClosureCompilerEnabled: boolean, private isCore: boolean,
-      private constantPool: ConstantPool) {
+/**
+ * Visits all classes, performs Ivy compilation where Angular decorators are present and collects
+ * result in a Map that associates a ts.ClassDeclaration with Ivy compilation results. This visitor
+ * does NOT perform any TS transformations.
+ */
+class IvyCompilationVisitor extends Visitor {
```
JoostK, line 52: Using a `Visitor` seems a bit overkill here, given that we never transform the nodes. The `TraitCompiler` has all the compilation records that you can request using `this.compiler.recordsFor(sourceFile)`, that should be sufficient to drive the compilation without needing to walk the AST.

JoostK (changes requested):
Sorry for being so late, but I think we can avoid the first AST walk.

JoostK (approved):

AndrewKushnir:
Quick note after trying to avoid first AST pass: the approach proposed by Joost generally works great, but it requires some additional calls to `ts.getOriginalNode` to make sure that we always operate with the same instance of an AST node when Closure compiler annotations are enabled (thus Tsickle transforms some nodes). We agreed to keep the current solution for now to make sure we operate with the same state of AST nodes.

AndrewKushnir:
Merge-assistance: for some reasons the `ci/angular: merge status` indicates that 1 review is missing (from Alex), when [he did approve this change](https://github.com/angular/angular/pull/38213#pullrequestreview-455078561).

PR #38223, opened 2020-07-24
feat(dev-infra): provide organization-wide merge-tool label configuration

devversion:
**See individual commits.**. This PR implements changes as outlined in [the proposal for more consistent versioning/labeling/branching/merging](https://docs.google.com/document/d/197kVillDwx-RZtSVOBtPb4BBIAw0E9RT3q3v6DZkykU).


dev-infra/pr/merge/defaults/branches.ts
```diff
@@ -0,0 +1,210 @@
...
+}
+
+/**
+ * Asserts that the given branch corresponds to an active LTS version-branch that can receive
+ * backported fixes. Throws an error if LTS expired or an invalid branch is selected.
+ */
+export async function assertActiveLtsBranch(repo: GithubRepo, branchName: string) {
+  const version = await getVersionOfBranch(repo, branchName);
+  const {'dist-tags': distTags, time} =
+      await (await fetch(`https://registry.npmjs.org/${repo.npmPackageName}`)).json();
```
josephperrott, line 157: Do we need to cache this response so we can do multiple assertions?

dev-infra/pr/merge/defaults/branches.ts
```diff
@@ -0,0 +1,210 @@
...
+
+  // LTS versions should be tagged in NPM in the following format: `v{major}-lts`.
+  const ltsVersion = parse(distTags[`v${version.major}-lts`]);
+
+  // Ensure that there is a LTS version tagged for the given version-branch major.
+  // e.g. if the version branch is `9.2.x` then we want to make sure that there is
+  // a LTS version tagged in NPM for `v9`.
+  if (ltsVersion === null) {
```
josephperrott, line 165: I think we shouldn't allow a force merge in this case, we should ensure that we are only merging into marked LTS branches.

We could include instructions about how to properly tag it?

dev-infra/pr/merge/defaults/branches.ts
```diff
@@ -0,0 +1,210 @@
...
+        `Not using last-minor branch for v${version.major} LTS version. PR ` +
+        `should be updated to target: ${ltsVersion.major}.${ltsVersion.minor}.x`);
+  }
+
+  const today = new Date();
+  const releaseDate = new Date(time[`${version.major}.0.0`]);
+  const ltsEndDate = new Date(
+      releaseDate.getFullYear(),
+      releaseDate.getMonth() + majorActiveSupportDuration + majorActiveTermSupportDuration,
+      releaseDate.getDate(), releaseDate.getMinutes(), releaseDate.getSeconds(),
```
josephperrott, line 189: Are we missing hours?

dev-infra/pr/merge/defaults/branches.ts
```diff
@@ -0,0 +1,210 @@
...
+      releaseDate.getMilliseconds());
+
+  // Check if LTS has already expired for the targeted major version. If so, we do not
+  // allow the merge as per our LTS guarantees. Can be forcibly overridden if desired.
+  // See: https://angular.io/guide/releases#support-policy-and-schedule.
+  if (today > ltsEndDate) {
+    const ltsEndDateText = ltsEndDate.toLocaleDateString();
+    warn(
+        red(`Long-term support has already expired for v${version.major} on ` +
```
josephperrott, line 198: I think this line might read a little bit better as:

```
red(`Long-term support ended for v${version.major} on ${ltsEndDateText}`);
```

josephperrott (commented):
Overall looking pretty good, left some comments.  We additionally should have @petebacondarwin or @jelbourn take a look to make sure we are all on the same page of the execution of the Pete's proposal.

dev-infra/pr/merge/target-label.ts
```diff
@@ -21,8 +37,14 @@ export function getTargetLabelFromPullRequest(config: MergeConfig, labels: strin
...
 }
 
-/** Gets the branches from the specified target label. */
-export function getBranchesFromTargetLabel(
-    label: TargetLabel, githubTargetBranch: string): string[] {
-  return typeof label.branches === 'function' ? label.branches(githubTargetBranch) : label.branches;
+/**
+ * Gets the branches from the specified target label.
+ *
+ * @throws {InvalidTargetLabelError} Invalid label has been applied to pull request.
```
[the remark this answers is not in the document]
devversion: I think since we do not catch them in this function, we should have them here. That was at least the intention.

dev-infra/pr/merge/defaults/labels.ts
```diff
@@ -0,0 +1,116 @@
...
+const nextBranchName = 'master';
+
+/**
+ * Gets a label configuration for the merge tooling that reflects the default Angular
+ * organization wide labeling and branching semantics as outlined in the specification.
+ *
+ * https://docs.google.com/document/d/197kVillDwx-RZtSVOBtPb4BBIAw0E9RT3q3v6DZkykU
+ */
+export async function getDefaultTargetLabelConfiguration(
+    api: GithubClient, github: GithubConfig, npmPackageName: string): Promise<TargetLabel[]> {
```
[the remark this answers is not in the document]
devversion: The github client is not necessarily coupled to a specific repository. I considered it initially, but given that a client is not necessarily authenticated and can work with any arbitrary repo, I didn't go with it. 

Note that I've used something similar for combining the two. See `GithubRepo`. This is the interface to the dev-infra config so here the consumer should pass both things individually IMO.

dev-infra/pr/merge/defaults/branches.ts
```diff
@@ -0,0 +1,210 @@
...
+
+  // If the current `next` version is for an upcoming major version, we know that there
+  // is no patch branch or feature-freeze/release-candidate branch for this major digit.
+  // If the current `next` version is the first minor of a major version, we know that
+  // the feature-freeze/release-candidate branch can only be the actual major branch. The
+  // patch branch is based on that, either the actual major branch or the last minor from the
+  // preceding major version. In all other cases, the patch branch and feature-freeze or
+  // release-candidate branch are part of the same major version. Consider the following:
+  //
+  //  CASE 1. next: 11.0.0-next.0: patch and feature-freeze/release-candidate can only be
```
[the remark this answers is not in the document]
devversion: Total typo! good catch 😄

dev-infra/pr/merge/defaults/branches.ts
```diff
@@ -0,0 +1,210 @@
...
+    // Collect all version-branches that should be considered for the latest
+    // version-branch (i.e. patch branch) or the feature-freeze/release-candidate.
+    // Branches that represent a minor more recent than the one in `next` are ignored.
+    if (semver !== null && majorVersionsToConsider.includes(semver.major)) {
+      branches.push({name, semver});
+    }
+  }
+
+  // Sort captured version-branches in descending order.
+  branches.sort((a, b) => compare(a.semver, b.semver) * -1);
```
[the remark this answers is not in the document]
devversion: Love it! I didn't know this existed. nice find 😄

dev-infra/pr/merge/defaults/branches.ts
devversion: I think we only need this assertion at maximum once per merge execution, so it should be good. Are you thinking of something else?

dev-infra/pr/merge/defaults/branches.ts
devversion: Yeah, that was something I wasn't sure about so I added a prompt as usually folks then ask dev-infra if they cannot get around / have a prompt. Do we have a section in the release instructions for tagging LTS? would love to link to it.

dev-infra/pr/merge/defaults/branches.ts
devversion: No?! 😅 Good catch

dev-infra/pr/merge/defaults/branches.ts
devversion: 👍

dev-infra/pr/merge/target-label.ts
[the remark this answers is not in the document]
devversion: I'm adding it to both as that shows best that the error is not caught here.

dev-infra/pr/merge/defaults/branches.ts
devversion: Updated to just error and not prompt. Should be clear from the message that the version needs to be tagged if there is actually LTS support for it (should be fine raising awareness here as it's obviously wrong if a LTS version is not tagged).

devversion:
@josephperrott Thanks for reviewing. Please have another look. 

Yeah agreed. I added @jelbourn as reviewer. I think Pete is OOO as of today.

dev-infra/pr/merge/defaults/labels.ts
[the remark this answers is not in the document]
josephperrott: Okay, works for me

dev-infra/pr/merge/defaults/branches.ts
josephperrott: Okay, we can leave it for now, probably something worth keeping in mind for ourselves long term.

josephperrott (approved):
LGTM

dev-infra/pr/merge/defaults/branches.ts
```diff
@@ -0,0 +1,201 @@
...
+  // next version-branch as that one is supposed to be the latest active version-branch. If it
+  // is not, then an error will be thrown due to two FF/RC branches existing at the same time.
+  for (const {name} of branches) {
+    const version = await getVersionOfBranch(repo, name);
+    const isPrerelease = version.prerelease[0] === 'rc' || version.prerelease[0] === 'next';
+    if (isPrerelease) {
+      if (releaseCandidateBranch !== null) {
+        throw Error(
+            `Unable to determine latest release-train. Found two consecutive ` +
+            `branches in feature-freeze/release-candidate phase. Expected "${name}" to not ` +
```
jelbourn, line 115: I would print out both branch names here

dev-infra/pr/merge/defaults/branches.ts
```diff
@@ -0,0 +1,201 @@
...
+    throw Error(
+        `Unable to determine the latest release-train. The following branches ` +
+        `have been considered: [${branches.join(', ')}]`);
+  }
+
+  return {releaseCandidateBranch, latestVersionBranch};
+}
+
+/** Gets the version of a given branch by reading the `package.json` upstream. */
+export async function getVersionOfBranch(repo: GithubRepo, branchName: string): Promise<SemVer> {
```
jelbourn, line 139:
```suggestion
export async function getBranchVersion(repo: GithubRepo, branchName: string): Promise<SemVer> {
```
you could probably shorten this a tiny bit

dev-infra/pr/merge/defaults/labels.ts
```diff
@@ -0,0 +1,116 @@
...
+import {InvalidTargetBranchError, InvalidTargetLabelError} from '../target-label';
+
+import {assertActiveLtsBranch, fetchActiveReleaseTrainBranches, getVersionOfBranch, GithubRepo, releaseTrainBranchNameRegex} from './branches';
+
+/** Branch name for the `next` branch. */
+const nextBranchName = 'master';
+
+/**
+ * Gets a label configuration for the merge tooling that reflects the default Angular
+ * organization wide labeling and branching semantics as outlined in the specification.
```
jelbourn, line 21:
```suggestion
 * organization-wide labeling and branching semantics as outlined in the specification.
```

dev-infra/pr/merge/defaults/labels.ts
```diff
@@ -0,0 +1,116 @@
...
+          throw new InvalidTargetLabelError(
+              `Unable to merge pull request as "${nextBranchName}" branch will be ` +
+              `released as a minor version.`);
+        }
+        return [nextBranchName];
+      },
+    },
+    {
+      pattern: 'target: minor',
+      branches: [nextBranchName],
```
jelbourn, line 49: Maybe leave a note here that this would need to be updated if we ever want to cut a minor release while we're already on a major release train

(e.g. we're working on 12.0.0, and then realize we want to backport a feature to an 11.3.0 to make migration easier)

dev-infra/pr/merge/defaults/labels.ts
```diff
@@ -0,0 +1,116 @@
...
+        // into the release-candidate/feature-freeze branch, and a separate PR has been created.
+        if (githubTargetBranch === releaseCandidateBranch) {
+          return [releaseCandidateBranch];
+        }
+        // Otherwise, merge into the next and active release-candidate/feature-freeze branch.
+        return [nextBranchName, releaseCandidateBranch];
+      },
+    },
+    {
+      pattern: 'target: lts',
```
jelbourn, line 93: Maybe add a comment here along the lines
```ts
// LTS changes are rare enough that we won't worry about cherry-picking changes into LTS
// branches for PRs created against any other branch. Instead, PR authors need to manually
// create a separate PR for the LTS branch.
```

dev-infra/pr/merge/defaults/branches.ts
```diff
@@ -0,0 +1,201 @@
...
+        throw Error(
+            `Discovered unexpected old feature-freeze/release-candidate branch. Expected no ` +
+            `version-branch in feature-freeze/release-candidate mode for v${version.major}`);
+      }
+      releaseCandidateBranch = name;
+    } else {
+      latestVersionBranch = name;
+      break;
+    }
+  }
```
jelbourn, line 127: optional: I might try splitting this function up into two or three smaller functions

jelbourn:
I think everything here matches up with the doc. I do think though, that this tool has reached the point where I'd feel a lot more comfortable with some unit tests. I know that will be tricky due needing to fake external dependencies, but we should probably start adding unit tests for this in a follow-up.

dev-infra/pr/merge/defaults/labels.ts
josephperrott: Its probably worth noting additionally, that often differences between active branches and LTS branches can be enough that typically cherry-picking isn't even an option.

dev-infra/pr/merge/defaults/branches.ts
```diff
@@ -0,0 +1,201 @@
...
+/** Type describing a Github repository with corresponding API client. */
+export interface GithubRepo {
+  /** API client that can access the repository. */
+  api: GithubClient;
+  /** Owner login of the repository. */
+  owner: string;
+  /** Name of the repository. */
+  repo: string;
+  /** NPM package representing this repository. */
+  npmPackageName: string;
```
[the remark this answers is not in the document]
devversion: yeah, agreed.

dev-infra/pr/merge/defaults/branches.ts
```diff
@@ -0,0 +1,201 @@
...
+  } else if (nextVersion.minor === 1) {
+    expectedReleaseCandidateMajor = nextVersion.major;
+    majorVersionsToConsider.push(nextVersion.major, nextVersion.major - 1);
+  } else {
+    expectedReleaseCandidateMajor = nextVersion.major;
+    majorVersionsToConsider.push(nextVersion.major);
+  }
+
+  for (const {name} of branchData) {
+    if (!releaseTrainBranchNameRegex.test(name)) {
```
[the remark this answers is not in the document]
devversion: 👍

dev-infra/pr/merge/defaults/branches.ts
```diff
@@ -0,0 +1,201 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import fetch from 'node-fetch';
+import {parse, rcompare, SemVer} from 'semver';
```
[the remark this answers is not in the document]
devversion: Switched to a namespace import for `semver.parse` and `semver.rcompare`

dev-infra/pr/merge/defaults/branches.ts
```diff
@@ -0,0 +1,201 @@
...
+ * https://angular.io/guide/releases#support-policy-and-schedule.
+ */
+export const majorActiveTermSupportDuration = 12;
+
+/**
+ * Fetches the active release train and their branches for the specified major version. e.g.
+ * the latest active release-train branch name is resolved and an optional version-branch for
+ * a currently active feature-freeze/release-candidate release-train.
+ */
+export async function fetchActiveReleaseTrainBranches(repo: GithubRepo, nextVersion: SemVer) {
```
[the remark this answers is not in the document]
devversion: yeah, that's a good point. Added an explicit returns type with comments.

dev-infra/pr/merge/defaults/branches.ts
```diff
@@ -0,0 +1,201 @@
...
+export const majorActiveSupportDuration = 6;
+
+/**
+ * Number of months a major version has active long-term support. See:
+ * https://angular.io/guide/releases#support-policy-and-schedule.
+ */
+export const majorActiveTermSupportDuration = 12;
+
+/**
+ * Fetches the active release train and their branches for the specified major version. e.g.
```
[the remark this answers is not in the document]
devversion: 👍

dev-infra/pr/merge/defaults/branches.ts
devversion: I'd prefer leaving this as is unless you feel strongly about it. I think `branch-version` is less explicit than `version of branch`.

dev-infra/pr/merge/defaults/labels.ts
devversion: 👍

dev-infra/pr/merge/defaults/labels.ts
devversion: Yeah, added a comment and a TODO. I think there are ways to handle this automatically. I'll play with it an follow-up.

dev-infra/pr/merge/defaults/branches.ts
devversion: 👍

devversion:
@jelbourn Thanks for the feedback. Addressed the feedback and also went ahead with adding tests.

dev-infra/pr/merge/defaults/branches.ts
```diff
@@ -0,0 +1,201 @@
...
+  // If the current `next` version is for an upcoming major version, we know that there
+  // is no patch branch or feature-freeze/release-candidate branch for this major digit.
+  // If the current `next` version is the first minor of a major version, we know that
+  // the feature-freeze/release-candidate branch can only be the actual major branch. The
+  // patch branch is based on that, either the actual major branch or the last minor from the
+  // preceding major version. In all other cases, the patch branch and feature-freeze or
+  // release-candidate branch are part of the same major version. Consider the following:
+  //
+  //  CASE 1. next: 11.0.0-next.0: patch and feature-freeze/release-candidate can only be
+  //          most recent `10.<>.x` branches. The FF/RC branch can only be the last-minor of v10.
```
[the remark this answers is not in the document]
devversion: Would that be for the first case? I'm slightly confuse whether this belongs here or not. FWIW: I added a different comment in `labels.ts` for the exceptional minor releases. Maybe that is sufficient already

jelbourn (approved):
LGTM, great tests!

PR #38224, opened 2020-07-24
test: remove instructions from golden api files

IgorMinar:
see individual commits


IgorMinar:
@alxhub can you please confirm that the three instructions that I identified as public api are the only ones that are out there. thanks!

jelbourn (approved):
LGTM

Reviewed-for: public-api

jelbourn:
cc @devversion since I know you've worked w/ the api guard before

packages/core/src/di/interface/defs.ts
```diff
@@ -22,9 +22,9 @@ import {ClassProvider, ConstructorProvider, ExistingProvider, FactoryProvider, S
  * `InjectorDef`, `NgModule`, or a special scope (e.g. `'root'`). A value of `null` indicates
  * that the injectable does not belong to any scope.
  *
- * NOTE: This is a private type and should not be exported
- *
- * @publicApi
+ * @codeGenApi
+ * @publicApi This instruction has been emitted by ViewEngine for some time and is deployed to npm.
```
devversion, line 26: Nit: I don't think this is an instruction, but rather a definition type IIRC. Should we just say that this is emitted in `d.ts` files for injectables (also by ViewEngine) and not mention instruction here?

mhevery (approved):

devversion (approved):

packages/core/src/di/interface/defs.ts
IgorMinar: sgtm

tools/ts-api-guardian/index.bzl
```diff
@@ -109,8 +112,11 @@ def ts_api_guardian_test_npm_package(
     ]
 
     for i in strip_export_pattern:
-        # The below replacement is needed because under Windows '^' needs to be escaped twice
-        args += ["--stripExportPattern", i.replace("^", "^^^^")]
+        # Escape regexp before passing it via the command line.
+        # The below replacement is needed because under Windows '^' needs to be escaped twice,
+        # the remaining characters need a single escape on all platforms.
+        escaped_pattern = i.replace("^", "^^^^").replace("(", "\(").replace(")", "\)").replace("|", "\|")
+        args += ["--stripExportPattern", escaped_pattern]
```
devversion, line 119: I think this could be simplified/made more-robust by escaping the argument with a quote. e.g. 

```bzl
args += ["--stripExportPattern", "\"%s\"" % i]
```

It looks like the issue is that the argument is directly put in the `nodejs_binary` executable without any quotes (so this fixes it quite nicely IMO)

IgorMinar: sgtm, but does that mean that the `^` no longer needs to be escaped as well? I can't test this on windows, can you please confirm?

devversion: Seems to work for me. Looks like the `^` escape isn't even needed on my end at all. I think it might be best to keep this as is and I can send a follow-up to simplify this/test on Windows CI too (not only my local machine)

devversion: Ah, you already made the change. That works for me too 😄 I'll test the change on Windows CI too to make sure.

jelbourn (approved):
LGTM

Reviewed-for: fw-security

tools/ts-api-guardian/index.bzl
IgorMinar: cool!

devversion: passes on windows CI and locally tested with Powershell, CMD and Git Bash. Thanks for making the change!

petebacondarwin:
Typo in commit message: `test: update ts-api-guardian's strip_export_pattern to exclude Ivy instructions`

> We've since also ~~came~~come up with an alternative plan to compile libraries with the Ivy compiler
for npm deployment and this plan does not rely on making Ivy instructions public.

petebacondarwin (approved):
Reviewed-for: public-api

mhevery (approved):

mhevery (approved):

alxhub (approved):
(2 instructions, and one interface)

Yes, `defineInjectable` and `inject` are the instructions we currently ship to NPM. The `@publicApi` comments make sense for now.

However long term, the plan is to continue shipping these instructions to NPM with Ivy, so these will be intentionally public API, not just for historical reasons.

mhevery (approved):
Review-for: fw-security

mhevery (commented):
Reviewed-for: fw-security

mhevery (approved):
Reviewed-for: fw-security

mhevery (approved):

PR #38229, opened 2020-07-24
refactor(router): extract Router config utils to a separate file

### LINKED ISSUE #38212: `RouterModule` directives cannot be tree shaken under Ivy
# 🐞 bug report

### Affected Package

### Is this a regression?

 Yes, the previous version in which this bug was not present was: View Engine

### Description

Directives in `RouterModule` cannot be tree shaken in Ivy builds. This is because of a circular dependency which requires the use of `setComponentScope()` to be used. `setComponentScope()` is inherently side-effectful, and is the compiler's way of working around circular dependencies, where it can not import the required dependencies to define a template scope without causing a cycle in the import graph.

The solution here is to remove the cycle, and the compiler will be able to generate code that does **not** use `setComponentScope()`. This should allow compilers like Closure to tree shake unused components in `RouterModule`.

## 🔬 Minimal Reproduction

Import something from `RouterModule` (such as `RouterLink`). In the output bundle (at least for Closure, I didn't check Terser) it will include `RouterLinkActive`, despite the fact that it was not actually used in the application.

## 🌍  Your Environment

**Angular Version:** `master`

**Anything else relevant?**

Spin off from #38211. I was able to generate a circular dependency list by patching #38210, this is the dependency chain which needs to be broken:

```
/home/dparker/Source/ng/packages/router/src/components/empty_outlet.ts
/home/dparker/Source/ng/packages/router/src/directives/router_outlet.ts
/home/dparker/Source/ng/packages/router/src/config.ts
/home/dparker/Source/ng/packages/router/src/components/empty_outlet.ts
```

Also see http://b/162019906 for more investigative information.

AndrewKushnir:
This commit refactors Router package to move config utils to a separate file for better
organization and to resolve the problem with circular dependency issue.

Resolves #38212.

## PR Type

- [ ] Yes
- [x] No

### DIFF
--- a/packages/router/src/config.ts
+++ b/packages/router/src/config.ts
@@ -9,9 +9,7 @@
 import {NgModuleFactory, NgModuleRef, Type} from '@angular/core';
 import {Observable} from 'rxjs';
 
-import {EmptyOutletComponent} from './components/empty_outlet';
 import {ActivatedRouteSnapshot} from './router_state';
-import {PRIMARY_OUTLET} from './shared';
 import {UrlSegment, UrlSegmentGroup} from './url_tree';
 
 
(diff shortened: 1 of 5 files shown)


AndrewKushnir:
[Presubmit](http://test/OCL:323088984:BASE:323088999:1595631388578:91ef7e5c)

AndrewKushnir:
FYI, also tested in Google's codebase and observed the effect of this change: the `setComponentScope` call is no longer generated and `RouterLink` is tree-shaken away.

atscott (approved):
👍  Looks good; thanks for the refactor!

PR #38244, opened 2020-07-27
fix(docs-infra): preserve focus on copy (and prevent scrolling to bottom on IE11)

gkalpak:
The `CopierService` is used for copying text to the user's clipboard. It is, for example, used in `CodeComponent` to copy example code snippets. This is implemented by creating a temporary, hidden `<textarea>` elements, setting its value to the text that needs to be copied, executing the `copy` command and finally removing the element from the DOM.

Previously, as a result of `CopierService`'s implementation, the focused element would lose focus, while the temporary `<textarea>` element would implicitly gain focus when selecting its contents. This had an even worse side-effect on IE11, which seems to scroll to the bottom of the containing element (here `<body>`) when the focused element is removed.

This PR fixes these issues by keeping track of the previously focused element and restoring its focus after the copy operation. It also includes some minor refactorings. See individual commits for more details.

### DIFF
--- a/aio/src/app/shared/copier.service.ts
+++ b/aio/src/app/shared/copier.service.ts
@@ -59,14 +59,13 @@ export class CopierService {
     style.padding = '0';
     style.margin = '0';
 
-    // Move element out of screen horizontally.
-    style.position = 'absolute';
-    style[ isRTL ? 'right' : 'left' ] = '-9999px';
+    // Make the element invisible and move it out of screen horizontally.
+    style.opacity = '0';
+    style.position = 'fixed';
+    style.top = '0';
+    style[isRTL ? 'right' : 'left'] = '-999em';
 
-    // Move element to the same position vertically.
-    const yPosition = window.pageYOffset || docElem.scrollTop;
-    style.top = yPosition + 'px';
-
+    textArea.setAttribute('aria-hidden', 'true');
     textArea.setAttribute('readonly', '');
     textArea.value = text;
 
--- a/aio/src/app/shared/copier.service.ts
+++ b/aio/src/app/shared/copier.service.ts
@@ -5,6 +5,9 @@
  * - https://github.com/zenorocha/clipboard.js/
  *
  * Both released under MIT license - © Zeno Rocha
+ *
+ * It is also influenced by the Angular CDK `PendingCopy` class:
+ * https://github.com/angular/components/blob/master/src/cdk/clipboard/pending-copy.ts
  */
 
 
@@ -18,6 +21,8 @@ export class CopierService {
    * @return Whether the copy operation was successful.
    */
   private copyTextArea(textArea: HTMLTextAreaElement): boolean {
+    const currentFocus = document.activeElement as HTMLOrSVGElement | null;
+
     try {
       textArea.select();
       textArea.setSelectionRange(0, textArea.value.length);
@@ -25,6 +30,8 @@ export class CopierService {
       return document.execCommand('copy');
     } catch {
       return false;
+    } finally {
+      currentFocus?.focus();
     }
   }
 
--- a/aio/src/app/shared/copier.service.ts
+++ b/aio/src/app/shared/copier.service.ts
@@ -10,11 +10,11 @@
 
 export class CopierService {
   /**
-   * Copy the contents of a `<textarea>` element to clipboard.
+   * Copy the contents of a `<textarea>` element to the clipboard.
    *
    * NOTE: For this method to work, the elements must be already inserted into the DOM.
    *
-   * @param textArea The area containing the text to be copied to clipboard.
+   * @param textArea The area containing the text to be copied to the clipboard.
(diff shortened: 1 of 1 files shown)


aio/src/app/shared/copier.service.ts
```diff
@@ -5,63 +5,87 @@
...
-      // Prevent zooming on iOS
-      this.fakeElem.style.fontSize = '12pt';
+  /**
+   * Create a temporary, hidden `<textarea>` element, set its value to the specified text and insert
+   * it into the DOM.
+   *
+   * @param text The text to be inserted into the textarea.
+   * @return The temporary `<textarea>` element containing the specified text.
+   */
+  private createTextArea(text: string): HTMLTextAreaElement {
```
saifabusaleh, line 43: this function does two things, 
it creates text area
it adds it to DOM

Is this name is good enough?

saifabusaleh:
Hi @gkalpak ,

I wrote few comments

in MDN is saying that `select` is not always doing `focus`:
https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/select


but anyway returning the `focus` to previously focused element sounds like good fix 👍 

LGTM 

Thanks

aio/src/app/shared/copier.service.ts
```diff
@@ -5,63 +5,87 @@
...
+      textArea.select();
+      textArea.setSelectionRange(0, textArea.value.length);
 
-      // Create a fake element to hold the contents to copy
-      this.fakeElem = document.createElement('textarea');
+      return document.execCommand('copy');
+    } catch {
+      return false;
+    } finally {
+      currentFocus?.focus();
```
saifabusaleh, line 32: Maybe its needed to write comment why this is needed after the `select` ?

saifabusaleh:
I see also now that the white border is now persistent after clicking the icon to copy, while it wasn't like that before

image

aio/src/app/shared/copier.service.ts
gkalpak: Good catch. I will update the function to not add the testarea to the DOM. I'll move the operqtion to the `copyText()` method, which is also responsible for removing the element from the DOM (so making it responsible for also inserting it makes sense).

aio/src/app/shared/copier.service.ts
gkalpak: Good call. I'll add a comment.

gkalpak:
Thank you for the review, @saifabusaleh ❤️ I have made the changes.

> I see also now that the white border is now persistent after clicking the icon to copy, while it wasn't like that before

This is now happening because the `copy` button retains its focus (and what you see is its focus style). Previously, the focus would go to the hidden textarea, so the `copy` button would never be focused.

If we don't like the focus style, we can improve it in a separate PR (but such a change would be outside the scope of this PR).

saifabusaleh:
@gkalpak LGTM

Thanks 👍

aio/src/app/shared/copier.service.ts
```diff
@@ -5,63 +5,95 @@
...
+ *
+ * It is also influenced by the Angular CDK `PendingCopy` class:
+ * https://github.com/angular/components/blob/master/src/cdk/clipboard/pending-copy.ts
  */
 
 
 export class CopierService {
-    private fakeElem: HTMLTextAreaElement|null;
+  /**
+   * Copy the contents of a `<textarea>` element to clipboard.
```
kapunahelewong, line 16:
```suggestion
   * Copy the contents of a `<textarea>` element to the clipboard.
```

aio/src/app/shared/copier.service.ts
```diff
@@ -5,63 +5,95 @@
...
+   *
+   * NOTE: For this method to work, the elements must be already inserted into the DOM.
+   *
+   * @param textArea The area containing the text to be copied to clipboard.
```
kapunahelewong, line 20:
```suggestion
   * @param textArea The area containing the text to be copied to the clipboard.
```

aio/src/app/shared/copier.service.ts
```diff
@@ -5,63 +5,95 @@
...
+  }
 
-    removeFake() {
-      if (this.fakeElem) {
-        document.body.removeChild(this.fakeElem);
-        this.fakeElem = null;
-      }
-    }
+  /**
+   * Copy the specified text to clipboard.
```
kapunahelewong, line 76:
```suggestion
   * Copy the specified text to the clipboard.
```

aio/src/app/shared/copier.service.ts
```diff
@@ -5,63 +5,95 @@
...
+   *
+   * @param text The text to be copied to clipboard.
```
kapunahelewong, line 78:
```suggestion
   * @param text The text to be copied to the clipboard.
```

kapunahelewong (approved):
LGTM! (just a few minor notes)

Issue #30981, opened 2019-06-11
Type checking of templates that set private inputs


kolkov:
I try to build with Ivy and get an error:
```
WARNING in Failed to read entry point info from D:/Go/pickup/ui/node_modules/@angular/pwa/node_modules/@schematics/angular/workspace/files/pack
age.json with error SyntaxError: Unexpected token < in JSON at position 1121.
```

```
ERROR in src/__ng_typecheck__.ts(31,64): error TS2344: Type '"dataSource" | "trackBy" | "inputs" | "carouselLoad" | "onMove"' does not satisfy
the constraint '"ngOnInit" | "ngDoCheck" | "ngOnDestroy" | "ngAfterContentInit" | "dataSource" | "trackBy" | "_dataSubscription" | "_dataSource
" | "_dataDiffer" | "styleid" | "activePoint" | ... 42 more ... | "velocity"'.
  Type '"inputs"' is not assignable to type '"ngOnInit" | "ngDoCheck" | "ngOnDestroy" | "ngAfterContentInit" | "dataSource" | "trackBy" | "_dat
aSubscription" | "_dataSource" | "_dataDiffer" | "styleid" | "activePoint" | ... 42 more ... | "velocity"'.
```

![image](https://user-images.githubusercontent.com/3740898/59298846-02b55f00-8c94-11e9-8a02-740a449f14b8.png)

```
Angular CLI: 8.0.2
Node: 11.14.0
OS: win32 x64
Angular: 8.1.0-next.1
... animations, common, compiler, compiler-cli, core, forms
... language-service, platform-browser, platform-browser-dynamic
... router, service-worker

Package                           Version
-----------------------------------------------------------
@angular-devkit/architect         0.800.2
@angular-devkit/build-angular     0.800.2
@angular-devkit/build-optimizer   0.800.2
@angular-devkit/build-webpack     0.800.2
@angular-devkit/core              8.0.2
@angular-devkit/schematics        8.0.2
@angular/cdk                      8.0.1
@angular/cli                      8.0.2
@angular/material                 8.0.1
@angular/pwa                      0.12.4
@ngtools/webpack                  8.0.2
@schematics/angular               8.0.2
@schematics/update                0.800.2
rxjs                              6.5.2
typescript                        3.4.5
webpack                           4.30.0
```
```
"private": true,
  "dependencies": {
    "@angular/animations": "^8.1.0-next.1",
    "@angular/cdk": "^8.0.1",
[...]

PR #38249, opened 2020-07-27
refactor(compiler-cli): only use type constructors for directives with generic types

### LINKED ISSUE #38185: Ivy LS: refactor type constructors and input binding checking
Currently input bindings in templates are type-checked via a type constructor call for the directive or component in question. This is mainly for convenience, since type constructors were always called with the inputs present anyway.

This is undesirable for two reasons:
1) Type constructors are expensive, and often unnecessary. They only need to be called if the directive or component is actually generic, and only need those inputs passed which reference the generic type parameters.

2) For language service use cases, it is valuable to have a direct assignment operation for inputs in the TCB in order to extract LS information (e.g. `ts.TypeChecker.getSymbolAtLocation`).

Thus, we should refactor the way input checking works:
* only call type constructors when necessary, and only with those inputs that affect generic type inference
* always emit assignment operations for inputs
* for diagnostics, rely on the type ctor call for generic inputs, and on the assignment operations for non-generic inputs (adding the ignore flag to the others)

JoostK:
Prior to this change, the template type checker would always use a
type-constructor to instantiate a directive. This type-constructor call
serves two purposes:

1. Infer any generic types for the directive instance from the inputs
   that are passed in.
2. Type check the inputs that are passed into the directive's inputs.

The first purpose is only relevant when the directive actually has any
generic types and using a type-constructor for these cases inhibits
a type-check performance penalty, as a type-constructor's signature is
quite complex and needs to be generated for each directive.

This commit refactors the generated type-check blocks to only generate
a type-constructor call for directives that have generic types. Type
checking of inputs is achieved by generating individual statements for
all inputs, using assignments into the directive's fields.

Even if a type-constructor is used for type-inference of generic types
will the input checking also be achieved using the individual assignment
statements. This is done to support the rework of the language service,
which will start to extract symbol information from the type-check
blocks.

As a future optimization, it may be possible to reduce the number of
inputs passed into a type-constructor to only those inputs that
contribute the the type-inference of the generics. As this is not a
necessity at the moment this is left as follow-up work.

Closes #38185

---

On top of #38226 as it requires that fix for the tests to pass.

### DIFF
--- a/packages/compiler-cli/src/ngtsc/core/api/src/public_options.ts
+++ b/packages/compiler-cli/src/ngtsc/core/api/src/public_options.ts
@@ -151,7 +151,9 @@ export interface StrictTemplateOptions {
    * Whether to check if the input binding attempts to assign to a restricted field (readonly,
    * private, or protected) on the directive/component.
    *
-   * Defaults to `false`, even if "fullTemplateTypeCheck" and/or "strictInputTypes" is set.
+   * Defaults to `false`, even if "fullTemplateTypeCheck", "strictTemplates" and/or
+   * "strictInputTypes" is set. Note that if `strictInputTypes` is not set, or set to `false`, this
+   * flag has no effect.
    */
   strictInputAccessModifiers?: boolean;
 
--- a/packages/compiler-cli/src/ngtsc/core/api/src/public_options.ts
+++ b/packages/compiler-cli/src/ngtsc/core/api/src/public_options.ts
@@ -148,6 +148,14 @@ export interface StrictTemplateOptions {
   strictInputTypes?: boolean;
 
   /**
+   * Whether to check if the input binding attempts to assign to a restricted field (readonly,
+   * private, or protected) on the directive/component.
+   *
+   * Defaults to `false`, even if "fullTemplateTypeCheck" and/or "strictInputTypes" is set.
+   */
+  strictInputAccessModifiers?: boolean;
+
+  /**
    * Whether to use strict null types for input bindings for directives.
    *
    * If this is `true`, applications that are compiled with TypeScript's `strictNullChecks` enabled
--- a/packages/compiler-cli/src/ngtsc/core/api/src/public_options.ts
+++ b/packages/compiler-cli/src/ngtsc/core/api/src/public_options.ts
@@ -154,6 +154,8 @@ export interface StrictTemplateOptions {
    * Defaults to `false`, even if "fullTemplateTypeCheck", "strictTemplates" and/or
    * "strictInputTypes" is set. Note that if `strictInputTypes` is not set, or set to `false`, this
    * flag has no effect.
+   *
+   * Tracking issue for enabling this by default: https://github.com/angular/angular/issues/38400
    */
   strictInputAccessModifiers?: boolean;
 
--- a/packages/compiler-cli/src/ngtsc/typecheck/src/ts_util.ts
+++ b/packages/compiler-cli/src/ngtsc/typecheck/src/ts_util.ts
@@ -90,13 +90,13 @@ export function tsDeclareVariable(id: ts.Identifier, type: ts.TypeNode): ts.Vari
  * Create a `ts.VariableStatement` that initializes a variable with a given expression.
  *
  * Unlike with `tsDeclareVariable`, the type of the variable is inferred from the initializer
- * expression.
+ * expression unless an explicit type is provided.
  */
 export function tsCreateVariable(
-    id: ts.Identifier, initializer: ts.Expression): ts.VariableStatement {
+    id: ts.Identifier, initializer: ts.Expression, type?: ts.TypeNode): ts.VariableStatement {
   const decl = ts.createVariableDeclaration(
       /* name */ id,
-      /* type */ undefined,
+      /* type */ type,
       /* initializer */ initializer);
   return ts.createVariableStatement(
       /* modifiers */ undefined,
--- a/packages/compiler-cli/src/ngtsc/typecheck/src/ts_util.ts
+++ b/packages/compiler-cli/src/ngtsc/typecheck/src/ts_util.ts
@@ -87,6 +87,21 @@ export function tsDeclareVariable(id: ts.Identifier, type: ts.TypeNode): ts.Vari
 }
 
 /**
+ * Creates a `ts.TypeQueryNode` for a coerced input.
+ *
+ * For example: `typeof MatInput.ngAcceptInputType_value`, where MatInput is `typeName` and `value`
+ * is the `coercedInputName`.
+ *
+ * @param typeName The `EntityName` of the Directive where the static coerced input is defined.
+ * @param coercedInputName The field name of the coerced input
+ */
+export function tsCreateTypeQueryForCoercedInput(
+    typeName: ts.EntityName, coercedInputName: string): ts.TypeQueryNode {
+  return ts.createTypeQueryNode(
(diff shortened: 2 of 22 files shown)


atscott:
@googlebot I consent

JoostK:
Note for public API reviewers: this adds a new compiler option `"strictInputAccessModifers"` that is currently disabled by default. If enabled, the template type checker will report errors when assigning into private/protected/readonly input fields.

atscott (approved):

atscott (approved):
reviewed-for: public-api

AndrewKushnir (approved):
Reviewed-for: public-api

packages/compiler-cli/src/ngtsc/typecheck/src/ts_util.ts
```diff
@@ -90,13 +90,13 @@ export function tsDeclareVariable(id: ts.Identifier, type: ts.TypeNode): ts.Vari
  * Create a `ts.VariableStatement` that initializes a variable with a given expression.
  *
  * Unlike with `tsDeclareVariable`, the type of the variable is inferred from the initializer
- * expression.
+ * expression unless an explicit type is provided.
  */
 export function tsCreateVariable(
-    id: ts.Identifier, initializer: ts.Expression): ts.VariableStatement {
+    id: ts.Identifier, initializer: ts.Expression, type?: ts.TypeNode): ts.VariableStatement {
```
alxhub, line 111: `null` instead of the optional argument here.

I know this seems a bit silly, especially because of the way it's just passed directly to TypeScript, but I think this kind of thing is important. In the context of TCBs, `tsCreateVariable` is actually doing two jobs here - creating variables in a way that uses type inference, and creating variables with explicit types. Arguably it could be two functions - `tsCreateInferredVariable` and `tsCreateTypedVariable`. Rather than get that specific, I think it's reasonable to use an argument to control whether the type is explicitly set or inferred, but I also it's important not to lose that distinction at the call site. Having an explicit indicator that the type should be inferred makes code using these functions a lot more readable.

packages/compiler-cli/src/ngtsc/typecheck/src/type_check_block.ts
```diff
@@ -307,14 +308,36 @@ class TcbTextInterpolationOp extends TcbOp {
   }
 }
 
+/**
+ * TODO: docs, especially how it compares to `TcbDirectiveCtorOp`
```
alxhub: I'm guessing these are things that will be fixed before this is merged?

packages/compiler-cli/src/ngtsc/typecheck/src/type_check_block.ts
```diff
@@ -307,14 +308,36 @@ class TcbTextInterpolationOp extends TcbOp {
...
+ *
+ * Executing this operation returns a reference to the directive instance variable with its inferred
+ * type.
+ */
+class TcbDirectiveOp extends TcbOp {
```
alxhub, line 320:
```suggestion
class TcbDirectiveTypeOp extends TcbOp {
```

packages/compiler-cli/src/ngtsc/typecheck/src/type_check_block.ts
```diff
@@ -307,14 +308,36 @@ class TcbTextInterpolationOp extends TcbOp {
...
+
+    const type = this.tcb.env.referenceType(this.dir.ref);
+    this.scope.addStatement(tsDeclareVariable(id, type));
+    return id;
+  }
+}
+
 /**
  * A `TcbOp` which constructs an instance of a directive with types inferred from its inputs, which
  * also checks the bindings to the directive in the process.
```
alxhub, line 338: This needs to be updated - not all bindings are checked anymore, only ones to generic inputs (if I'm reading the code correctly).

packages/compiler-cli/src/ngtsc/typecheck/src/type_check_block.ts
```diff
@@ -323,37 +346,143 @@ class TcbDirectiveOp extends TcbOp {
...
+    // TODO: report duplicate properties
+
+    const inputs = bindInputs(this.dir, this.node, this.tcb);
+    for (const input of inputs) {
+      // For bound inputs, the property is assigned the binding expression.
+      let expr = translateInput(input.attribute, this.tcb, this.scope);
+      if (!this.tcb.env.config.checkTypeOfInputBindings) {
+        // If checking the type of bindings is disabled, cast the resulting expression to 'any'
+        // before the assignment.
+        // TODO: now that we're not forced to satisfy the type constructor parameter, we may also
```
alxhub: I don't think so, because we still want this expression for Language Service usage.

So maybe it needs to be behind an internal config flag?

packages/compiler-cli/src/ngtsc/typecheck/src/type_check_block.ts
```diff
@@ -1327,34 +1448,63 @@ function tcbGetDirectiveInputs(
...
-      type: 'binding',
-      field: field,
-      expression: expr,
-      sourceSpan: attr.sourceSpan,
-    });
   }
+  return propertyToFieldNames;
 }
 
+type TcbDirectiveInput = {
```
alxhub, line 1501: I think for readability, let's split this into `TcbDirectiveBoundInput` and `TcbDirectiveUnsetInput`.

alxhub (approved):

packages/compiler-cli/src/ngtsc/typecheck/src/ts_util.ts
JoostK: Looks like this has gone unused through several more refactorings, so I reverted the change altogether.

packages/compiler-cli/src/ngtsc/core/api/src/public_options.ts
```diff
@@ -147,6 +147,16 @@ export interface StrictTemplateOptions {
    */
   strictInputTypes?: boolean;
 
+  /**
+   * Whether to check if the input binding attempts to assign to a restricted field (readonly,
+   * private, or protected) on the directive/component.
+   *
+   * Defaults to `false`, even if "fullTemplateTypeCheck", "strictTemplates" and/or
+   * "strictInputTypes" is set. Note that if `strictInputTypes` is not set, or set to `false`, this
+   * flag has no effect.
```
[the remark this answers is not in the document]
JoostK: Done in #38400, included a link in this comment.

packages/compiler-cli/src/ngtsc/typecheck/src/type_check_block.ts
```diff
@@ -323,37 +352,160 @@ class TcbDirectiveOp extends TcbOp {
...
+
+          const id = this.tcb.allocateId();
+          const type = tsCreateTypeQueryForCoercedInput(dirTypeRef.typeName, fieldName);
+          this.scope.addStatement(tsDeclareVariable(id, type));
+
+          target = id;
+        } else if (this.dir.undeclaredInputFields.has(fieldName)) {
+          // If no coercion declaration is present nor is the field declared (i.e. the input is
+          // declared in a `@Directive` or `@Component` decorator's `inputs` property) there is no
+          // assignment target available, so this field is skipped.
```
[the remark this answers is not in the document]
JoostK: Additionally, we should report diagnostics for restricted inputs. That would address #30981.

alxhub:
[Presubmit](http://test/OCL:325869426:BASE:325869565:1597089642055:deaf8f7e)

jelbourn (approved):
LGTM

Reviewed-for: public-api

PR #38253, opened 2020-07-27
refactor(compiler): wrap large strings in function

dgp1130:
Large strings constants are now wrapped in a function which is called whenever used. This works around a unique limitation of Closure, where it will **always** inline string literals at **every** usage, regardless of how large the string literal is or how many times it is used.

The workaround is to use a function rather than a string literal. Closure has differently inlining semantics for functions, where it will check the length of the function and the number of times it is used before choosing to inline it. By using a function, `ngtsc` makes Closure more conservative about inlining large strings, and avoids blowing up the bundle size.

This optimization is only used if the constant is a large string. A wrapping function is not included for other use cases, since it would just increase the bundle size and add unnecessary runtime performance overhead.

## PR Checklist

## PR Type

Angular Material button component is almost twice as large in Ivy as View Engine. This appears to be because it shares styles across two components (`MatButton` and `MatAnchor`), which are deduplicated in View Engine but not in Ivy. This is partially because Closure compiler will **always** inline string constants at **every** usage. This causes strings to be duplicated, and large strings (such as CSS styling) get duplicated.

Issue Number: http://b/161849099

## What is the new behavior?

The Angular compiler will generate large string constants as being wrapped in a function and invoked where needed. This tricks Closure into **not** inlining the large strings and prevents the bundle size duplication.

- [X] No

## Other information

Also see http://b/162106151 for additional context and a response from the Closure compiler team.

I'm also not totally sure what implication this will have on Terser. I don't anticipate any unique changes, but maybe we should mark the function pure? I'm not sure how we go about that in the compiler.

### DIFF
--- a/packages/compiler/src/constant_pool.ts
+++ b/packages/compiler/src/constant_pool.ts
@@ -102,8 +102,10 @@ export class ConstantPool {
 
   private nextNameIndex = 0;
 
+  constructor(private readonly isClosureCompilerEnabled: boolean = false) {}
+
   getConstLiteral(literal: o.Expression, forceShared?: boolean): o.Expression {
-    if ((literal instanceof o.LiteralExpr && !isLongStringExpr(literal)) ||
+    if ((literal instanceof o.LiteralExpr && !isLongStringLiteral(literal)) ||
         literal instanceof FixupExpression) {
       // Do no put simple literals into the constant pool or try to produce a constant for a
       // reference to a constant.
@@ -121,9 +123,39 @@ export class ConstantPool {
     if ((!newValue && !fixup.shared) || (newValue && forceShared)) {
       // Replace the expression with a variable
       const name = this.freshName();
-      this.statements.push(
-          o.variable(name).set(literal).toDeclStmt(o.INFERRED_TYPE, [o.StmtModifier.Final]));
-      fixup.fixup(o.variable(name));
+      let definition: o.WriteVarExpr;
+      let usage: o.Expression;
+      if (this.isClosureCompilerEnabled && isLongStringLiteral(literal)) {
+        // For string literals, Closure will **always** inline the string at
+        // **all** usages, duplicating it each time. For large strings, this
+        // unnecessarily bloats bundle size. To work around this restriction, we
+        // wrap the string in a function, and call that function for each usage.
+        // This tricks Closure into using inline logic for functions instead of
+        // string literals. Function calls are only inlined if the body is small
+        // enough to be worth it. By doing this, very large strings will be
+        // shared across multiple usages, rather than duplicating the string at
+        // each usage site.
+        //
+        // const myStr = function() { return "very very very long string"; };
+        // const usage1 = myStr();
+        // const usage2 = myStr();
+        definition = o.variable(name).set(new o.FunctionExpr(
+            [],  // Params.
+            [
+              // Statements.
+              new o.ReturnStatement(literal),
+            ],
+            ));
+        usage = o.variable(name).callFn([]);
+      } else {
+        // Just declare and use the variable directly, without a function call
+        // indirection. This saves a few bytes and avoids an unncessary call.
+        definition = o.variable(name).set(literal);
+        usage = o.variable(name);
+      }
+
+      this.statements.push(definition.toDeclStmt(o.INFERRED_TYPE, [o.StmtModifier.Final]));
+      fixup.fixup(usage);
     }
 
     return fixup;
@@ -314,7 +346,7 @@ function isVariable(e: o.Expression): e is o.ReadVarExpr {
   return e instanceof o.ReadVarExpr;
 }
 
-function isLongStringExpr(expr: o.LiteralExpr): boolean {
-  return typeof expr.value === 'string' &&
+function isLongStringLiteral(expr: o.Expression): boolean {
+  return expr instanceof o.LiteralExpr && typeof expr.value === 'string' &&
       expr.value.length >= POOL_INCLUSION_LENGTH_THRESHOLD_FOR_STRINGS;
 }
--- a/packages/compiler-cli/src/ngtsc/transform/src/transform.ts
+++ b/packages/compiler-cli/src/ngtsc/transform/src/transform.ts
@@ -249,7 +249,7 @@ function transformIvySourceFile(
     importRewriter: ImportRewriter, file: ts.SourceFile, isCore: boolean,
     isClosureCompilerEnabled: boolean,
     defaultImportRecorder: DefaultImportRecorder): ts.SourceFile {
-  const constantPool = new ConstantPool();
+  const constantPool = new ConstantPool(isClosureCompilerEnabled);
   const importManager = new ImportManager(importRewriter);
 
   // The transformation process consists of 2 steps:
--- a/packages/compiler-cli/test/ngtsc/ngtsc_spec.ts
+++ b/packages/compiler-cli/test/ngtsc/ngtsc_spec.ts
@@ -6685,6 +6685,66 @@ export const Foo = Foo__PRE_R3__;
                 // remains a string in the `styles` array.
                 '"img[_ngcontent-%COMP%] { background: url(/b/some-very-very-long-path.png); }"]');
       });
+
+      it('large strings are wrapped in a function for Closure', () => {
+        env.tsconfig({
+          annotateForClosureCompiler: true,
+        });
+
+        env.write('test.ts', `
+          import {Component} from '@angular/core';
+
+          @Component({
+            selector: 'comp-a',
+            template: 'Comp A',
+            styles: [
+              'div { background: url(/a.png); }',
+              'div { background: url(/some-very-very-long-path.png); }',
+            ]
+          })
+          export class CompA {}
+
+          @Component({
+            selector: 'comp-b',
+            template: 'Comp B',
+            styles: [
+              'div { background: url(/b.png); }',
+              'div { background: url(/some-very-very-long-path.png); }',
+            ]
+          })
+          export class CompB {}
+        `);
+
+        env.driveMain();
+        const jsContents = env.getContents('test.js');
+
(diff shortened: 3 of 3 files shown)


alfaproject:
Is this trying to solve the same issue that was fixed here? https://github.com/angular/angular/pull/38213

dgp1130:
@alfaproject, it's a related change. That PR fixes the problem in Angular, so we use the same variable for duplicated styles. However, that alone isn't enough for Closure, which will duplicate the strings each time. This PR works around the Closure limitations.

packages/compiler/src/constant_pool.ts
```diff
@@ -121,9 +121,37 @@ export class ConstantPool {
...
+      if (isLongStringLiteral(literal)) {
+        // For string literals, Closure will **always** inline the string at
+        // **all** usages, duplicating it each time. For large strings, this
+        // unnecessarily bloats bundle size. To work around this restriction, we
+        // wrap the string in a function, and call that function for each usage.
+        // This tricks Closure into using inline logic for functions instead of
+        // string literals. Function calls are only inlined if the body is small
+        // enough to be worth it. By doing this, very large strings will be
+        // shared across multiple usages, rather than duplicating the string at
+        // each usage site.
```
AndrewKushnir, line 135: Thanks for adding a comment here 👍
I'm wondering if we could also add a small snippet on how this would look like in generated code?

packages/compiler/src/constant_pool.ts
```diff
@@ -121,9 +121,37 @@ export class ConstantPool {
...
+        definition = o.variable(name).set(new o.FunctionExpr(
+          [],  // Params.
+          [  // Statements.
+            new o.CommentStmt(
+                'Use a function to trick Closure to avoid duplicating this large string across mulitiple usages. See TODO: PR.'),
```
AndrewKushnir, line 140: It looks like this part: `See TODO: PR.` needs an update (i.e. reference to an actual issue/PR)?

packages/compiler/src/constant_pool.ts
```diff
@@ -121,9 +121,37 @@ export class ConstantPool {
...
+            new o.ReturnStatement(literal),
+          ],
+        )).toDeclStmt(o.INFERRED_TYPE, [o.StmtModifier.Final]);
+        usage = o.variable(name).callFn([]);
+      } else {
+        // Just declare and use the variable directly, without a function call
+        // indirection. This saves a few bytes and avoids an unncessary call.
+        definition = o.variable(name).set(literal)
```
AndrewKushnir, line 148: I think we can try to have this logic one level up, i.e have it somewhere here instead:  https://github.com/angular/angular/blob/master/packages/compiler/src/render3/view/compiler.ts#L234 and also do that inly when Closure annotations are enabled by passing the flag from here: https://github.com/angular/angular/blob/master/packages/compiler-cli/src/ngtsc/annotations/src/component.ts#L575:
```
      const res = compileComponentFromMetadata(meta, pool, makeBindingParser(), this.annotateForClosureCompiler);
```
In this case the current logic (which is a bit more performant at runtime) will be preserved for non-Closure case and the logic in the `ConstantPool` will remain a bit more simple.

WDYT?

AndrewKushnir (commented):
Hi @dgp1130, thanks for the PR 👍
I've left a few comments, please have a look when you get a chance.

packages/compiler/src/constant_pool.ts
dgp1130: Added a simplistic example.

packages/compiler/src/constant_pool.ts
dgp1130: Yes, thanks for catching that. Updated with this PR link.

packages/compiler/src/constant_pool.ts
dgp1130: That's an interesting suggestion Andrew. I agree there may be a performance hit for Terser (not sure how smart it is about inlining functions of this pattern).

I would rather not pull this logic out to be specific to styles if possible. While styles are the primary motivating use case, any sufficiently large string runs into the same problem. By doing this in `ConstantPool`, we leverage the optimization for any such string.

If the main concern is around potential regressions in Terser-compiled code, would it make sense to conditionalize the new logic in `ConstantPool` based on whether or not we're annotating for Closure? I'm not sure how hard it is to get that value when we construct the `ConstantPool`.

packages/compiler/src/constant_pool.ts
dgp1130: Thanks for catching that, updated with a reference to this PR.

dgp1130:
I tried to make the linter happy, however I can't find any change that will stop it from reformatting the entirety of `ngtsc_spec.ts`. I think the long regex confuses it and there's no good way to split that over multiple lines.

:shrug:

dgp1130:
http://test/OCL:323456881:BASE:323456908:1595889882918:b3702682

AndrewKushnir:
> I tried to make the linter happy, however I can't find any change that will stop it from reformatting the entirety of ngtsc_spec.ts. I think the long regex confuses it and there's no good way to split that over multiple lines.

I've seen this weird lint behavior in the past :) You can try to go back to the string-based check in ngtsc_spec.ts and split the long string into multiple lines using string concatenation.

packages/compiler/src/constant_pool.ts
```diff
@@ -121,9 +121,45 @@ export class ConstantPool {
...
+                .set(new o.FunctionExpr(
+                    [],  // Params.
+                    [
+                      // Statements.
+                      new o.CommentStmt(
+                          'Use a function to trick Closure to avoid duplicating this large string across mulitiple usages. See https://github.com/angular/angular/pull/38253.'),
+                      new o.ReturnStatement(literal),
+                    ],
+                    ))
+                .toDeclStmt(o.INFERRED_TYPE, [o.StmtModifier.Final]);
```
AndrewKushnir, line 151: nit: I think the `.toDeclStmt(o.INFERRED_TYPE, [o.StmtModifier.Final])` can be moved to a place where we push `definition` to the statements list below (since it's called for both branches of the `if` condition):

```
this.statements.push(definition.toDeclStmt(o.INFERRED_TYPE, [o.StmtModifier.Final]));
```

packages/compiler/src/constant_pool.ts
```diff
@@ -121,9 +121,45 @@ export class ConstantPool {
...
+        // const usage2 = myStr();
+        definition =
+            o.variable(name)
+                .set(new o.FunctionExpr(
+                    [],  // Params.
+                    [
+                      // Statements.
+                      new o.CommentStmt(
+                          'Use a function to trick Closure to avoid duplicating this large string across mulitiple usages. See https://github.com/angular/angular/pull/38253.'),
+                      new o.ReturnStatement(literal),
```
AndrewKushnir, line 148: nit: for readability purposes may be split it into multiple lines (with string concat)?
(note: there is a small typo in the 'multiple')

```suggestion
                      new o.CommentStmt(
                          'Use a function to trick Closure to avoid duplicating this large string across multiple usages. ' +
                          'See https://github.com/angular/angular/pull/38253.'),
                      new o.ReturnStatement(literal),
```

AndrewKushnir (commented):
Thanks for the update @dgp1130. FYI, added a couple nit comments.

packages/compiler/src/constant_pool.ts
AndrewKushnir: I think passing the flag to the `ConstantPool` should be straightforward, the `ConstantPool` for Ivy AOT compilation is being [created here](https://github.com/angular/angular/blob/master/packages/compiler-cli/src/ngtsc/transform/src/transform.ts#L252) (and I think we can just keep the old logic for VE).

FYI, the payload size check in the `test_aio_local` CircleCI job failed with 2KB delta, so may be this is not handled by Terser correctly (thus the isClosure flag would make sense)?

packages/compiler/src/constant_pool.ts
dgp1130: Done.

packages/compiler/src/constant_pool.ts
dgp1130: Split the comment up over multiple lines.

packages/compiler/src/constant_pool.ts
dgp1130: Added a check for Closure in `ConstantPool`. I only provided that value for the one usage you mentioned and made it optional for now. Longer term, we should probably change all `new ConstantPool()` calls to specify whether they are being used for Closure or not.

dgp1130:
> I've seen this weird lint behavior in the past :) You can try to go back to the string-based check in ngtsc_spec.ts and split the long string into multiple lines using string concatenation.

I switched it back to a string check. I was hoping to avoid that as the content of the comment and nuances of whitespace aren't important to this test, but I guess the formatter won't allow any other option.

AndrewKushnir (approved):
LGTM, thanks @dgp1130 👍 

@alxhub could you please also have a quick look when you get a chance?

Thank you.

dgp1130:
Updated g3 tests: http://test/OCL:323456881:BASE:323633473:1595965840647:fa8c97f5

packages/compiler/src/constant_pool.ts
```diff
@@ -121,9 +123,42 @@ export class ConstantPool {
...
+        // each usage site.
+        //
+        // const myStr = function() { return "very very very long string"; };
+        // const usage1 = myStr();
+        // const usage2 = myStr();
+        definition = o.variable(name).set(new o.FunctionExpr(
+            [],  // Params.
+            [
+              // Statements.
+              new o.CommentStmt(
```
alxhub, line 146: I would say let's not emit a comment here - we generally don't add such explainers to our generated code in other cases, and this output is very well explained by the comments on the generation code above.

alxhub (approved):

packages/compiler/src/constant_pool.ts
dgp1130: Removed.

PR #38257, opened 2020-07-27
fix(dev-infra): roll back phased review conditions

atscott:
It was determined that the list of 'pending' groups also included inactive groups.
That means that the 'no-groups-above-this-pending' would generally fail because
there's almost always some inactive group above it.

This commit removes the conditions for phased review while we
investigate if the inactive groups can be excluded.

### DIFF
--- a/.pullapprove.yml
+++ b/.pullapprove.yml
@@ -1160,8 +1160,6 @@ groups:
   public-api:
     <<: *defaults
     conditions:
-      - *no-groups-above-this-pending
-      - *no-groups-above-this-rejected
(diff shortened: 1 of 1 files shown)


jelbourn (approved):
LGTM

josephperrott (approved):
LGTM

josephperrott (changes requested):
This change commit should actually be type `ci`

```
ci: roll back phased review conditions
```

josephperrott (approved):
LGTM

PR #38259, opened 2020-07-28
ci: only check active groups for the `no-groups-above-this-*` checks

josephperrott:
Since PullApprove starts all inactive groups as a pending state, to properly
assess if any groups we care about are pending we must only check the active
groups.  We additionally do this with rejected because the intention of the
reusable checks is around checking active rules only.

Must go in after #38257

### DIFF
--- a/.pullapprove.yml
+++ b/.pullapprove.yml
@@ -1162,6 +1162,8 @@ groups:
   public-api:
     <<: *defaults
     conditions:
+      - *no-groups-above-this-pending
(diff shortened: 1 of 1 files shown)


atscott (approved):

atscott:
Just a note that this will also have to re-add the conditions being removed in #38257 once that goes in.

josephperrott:
Unblocked with the merge of #38257

PR #38263, opened 2020-07-28
fix(platform-browser): key event handler should run in the correct zone

### LINKED ISSUE #38260: KeyEventPlugin forces event handler in `ngZone`, the behavior is different with non key event handler.
Consider the following case.

app.component.html
```
<button #btnToggle>Toggle</button>
<div *ngIf="show">
  <button (click)="onClick()">Button</button>
  <input (keydown.enter)="onKeydown()"/>
</div>
```

app.component.ts
```
export class AppComponent implements AfterViewInit {
    @ViewChild('btnToggle') btnToggle: ElementRef;
    show = false;
    constructor(private zone: NgZone, private cd: ChangeDetectorRef) {}

    ngAfterViewInit() {
      this.zone.runOutsideAngular(() => {
        const el = this.btnToggle.nativeElement as HTMLElement;
        el.addEventListener('click', e => {
          this.show = true;
          this.cd.detectChanges();
        })
      })
   }

   onClick() {
      console.log('button click in ngZone? ', NgZone.isInAngularZone());
   }

   onKeydown() {
      console.log('keydown in ngZone? ', NgZone.isInAngularZone());
    }
}
```

The live sample is here https://stackblitz.com/edit/angular-hga7lk?file=src%2Fapp%2Fapp.component.ts

After click the `toggle` button, then click the inside `button` and press enter inside the `input`.
The output will be:

```
button click in ngZone?  false
keyDown in ngZone? true
```

This is not consistent

JiaLiPassion:
Close #38260 .

Currently the KeyEventPlugin handle key event such as `keydown.enter` and always run into `ngZone` no matter
which zone it is added. And this behavior is not consistent with the other event handlers.

So this commit fixes this issue.

Also it make sure the `event handler` in key event plugin call on the correct context object.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #38260

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/application_ref.ts
+++ b/packages/core/src/application_ref.ts
@@ -599,9 +599,7 @@ export class ApplicationRef {
       private _initStatus: ApplicationInitStatus) {
     this._onMicrotaskEmptySubscription = this._zone.onMicrotaskEmpty.subscribe({
       next: () => {
-        this._zone.run(() => {
-          this.tick();
-        });
+        this.tick();
       }
     });
 
@@ -831,6 +829,15 @@ export class ApplicationRef {
    * detection pass during which all change detection must complete.
    */
   tick(): void {
+    if (NgZone.isInAngularZone()) {
+      this._tick();
+    } else {
+      this._zone.run(() => this._tick());
+    }
+  }
+
+  /** @internal */
+  _tick(): void {
     if (this._runningTick) {
       const errorMessage = (typeof ngDevMode === 'undefined' || ngDevMode) ?
           'ApplicationRef.tick is called recursively' :
--- a/packages/platform-browser/src/dom/dom_renderer.ts
+++ b/packages/platform-browser/src/dom/dom_renderer.ts
@@ -51,19 +51,7 @@ export function flattenStyles(
 }
 
 function decoratePreventDefault(eventHandler: Function): Function {
-  // `DebugNode.triggerEventHandler` needs to know if the listener was created with
-  // decoratePreventDefault or is a listener added outside the Angular context so it can handle the
-  // two differently. In the first case, the special '__ngUnwrap__' token is passed to the unwrap
-  // the listener (see below).
-  return (event: any) => {
-    // Ivy uses '__ngUnwrap__' as a special token that allows us to unwrap the function
-    // so that it can be invoked programmatically by `DebugNode.triggerEventHandler`. The debug_node
-    // can inspect the listener toString contents for the existence of this special token. Because
-    // the token is a string literal, it is ensured to not be modified by compiled code.
-    if (event === '__ngUnwrap__') {
-      return eventHandler;
-    }
-
+  const listener: Function = (event: any) => {
     const allowDefaultBehavior = eventHandler(event);
     if (allowDefaultBehavior === false) {
       // TODO(tbosch): move preventDefault into event plugins...
@@ -73,6 +61,12 @@ function decoratePreventDefault(eventHandler: Function): Function {
 
     return undefined;
   };
+  // `DebugNode.triggerEventHandler` needs to know if the listener was created with
+  // decoratePreventDefault or is a listener added outside the Angular context so it can handle the
+  // two differently. In the first case, bind the original listener to the special '__ngUnwrap__'
+  // property of the wrapped listener.
+  (listener as any)['__ngUnwrap__'] = eventHandler;
+  return listener;
 }
 
 let hasLoggedNativeEncapsulationWarning = false;
--- a/packages/platform-browser/src/dom/events/event_manager.ts
+++ b/packages/platform-browser/src/dom/events/event_manager.ts
@@ -122,12 +122,12 @@ export abstract class EventManagerPlugin {
    */
   ensureAddListenerInZone(element: HTMLElement, eventName: string, handler: Function): Function {
     const zone = this.manager.getZone();
-    if (!NgZone.isInAngularZone()) {
-      return zone.run(() => {
-        return getDOM().onAndCancel(element, eventName, handler);
-      });
-    } else {
-      return getDOM().onAndCancel(element, eventName, handler);
-    }
+    return zone.runOutsideAngular(() => {
+      const handlerInZone = () => {
+        return zone.run(handler as any, element, [eventName]);
+      };
+      handlerInZone['__ngUnwrap__'] = (handler as any)['__ngUnwrap__'] || handler;
+      return getDOM().onAndCancel(element, eventName, handlerInZone);
+    });
   }
 }
--- a/packages/platform-browser/src/dom/events/event_manager.ts
+++ b/packages/platform-browser/src/dom/events/event_manager.ts
@@ -108,4 +108,26 @@ export abstract class EventManagerPlugin {
     }
     return this.addEventListener(target, eventName, handler);
   }
+
+  /**
+   * Ensure that the event listener is added inside NgZone so the listener will trigger
+   * change detection automatically. if currently the context is already the NgZone,
+   * we just call addEventListener directly, otherwise run inside NgZone.
+   *
+   * @param element The HTML element to receive event notifications.
+   * @param eventName The name of the event to listen for.
+   * @param handler A function to call when the notification occurs. Receives the
+   * event object as an argument.
+   * @returns  A callback function that can be used to remove the handler.
+   */
+  ensureAddListenerInZone(element: HTMLElement, eventName: string, handler: Function): Function {
+    const zone = this.manager.getZone();
+    if (!NgZone.isInAngularZone()) {
+      return zone.run(() => {
+        return getDOM().onAndCancel(element, eventName, handler);
+      });
+    } else {
+      return getDOM().onAndCancel(element, eventName, handler);
+    }
+  }
 }
--- a/packages/platform-browser/src/dom/events/key_events.ts
+++ b/packages/platform-browser/src/dom/events/key_events.ts
@@ -8,6 +8,7 @@
 
 import {DOCUMENT, ɵgetDOM as getDOM} from '@angular/common';
 import {Inject, Injectable, NgZone} from '@angular/core';
+
 import {EventManagerPlugin} from './event_manager';
 
 /**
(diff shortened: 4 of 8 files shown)


packages/platform-browser/test/dom/events/key_events_spec.ts
```diff
@@ -66,5 +71,80 @@ import {KeyEventsPlugin} from '@angular/platform-browser/src/dom/events/key_even
       expect(() => plugin.addGlobalEventListener('window', 'keyup.control.esc', () => {}))
           .not.toThrowError();
     });
+
+    describe('NgZone', () => {
+      const zone = new NgZone({});
+      const doc = getDOM().supportsDOMEvents() ? document : getDOM().createHtmlDocument();
+      const manager = new EventManager([new KeyEventsPlugin(doc)], zone);
+
+      it('key event handler should in the ngZone when added inside ngZone', () => {
```
mhevery, line 80:
```suggestion
      it('key event handler should be in the ngZone when added inside ngZone', () => {
```

packages/platform-browser/test/dom/events/key_events_spec.ts
```diff
@@ -66,5 +71,80 @@ import {KeyEventsPlugin} from '@angular/platform-browser/src/dom/events/key_even
...
+        };
+        zone.run(() => {
+          manager.addEventListener(element, 'keydown.enter', handler);
+        });
+        const evt = createKeyboardEvent('keydown', {bubbles: true, cancelable: true, key: 'enter'});
+        getDOM().dispatchEvent(element, evt);
+        expect(isInNgZone).toEqual(`key handler inside ngZone: true`);
+      });
+
+      it('global key event handler should in the ngZone when added inside ngZone', () => {
```
mhevery, line 94:
```suggestion
      it('global key event handler should be in the ngZone when added inside ngZone', () => {
```

packages/platform-browser/test/dom/events/key_events_spec.ts
```diff
@@ -66,5 +71,80 @@ import {KeyEventsPlugin} from '@angular/platform-browser/src/dom/events/key_even
...
+        };
+        zone.run(() => {
+          manager.addGlobalEventListener('document', 'keydown.enter', handler);
+        });
+        const evt = createKeyboardEvent('keydown', {bubbles: true, cancelable: true, key: 'enter'});
+        getDOM().dispatchEvent(doc, evt);
+        expect(isInNgZone).toEqual(`key handler inside ngZone: true`);
+      });
+
+      it('key event handler should outside of the ngZone when added outside of the ngZone', () => {
```
mhevery, line 107:
```suggestion
      it('key event handler should be outside of the ngZone when added outside of the ngZone', () => {
```

packages/platform-browser/test/dom/events/key_events_spec.ts
```diff
@@ -66,5 +71,80 @@ import {KeyEventsPlugin} from '@angular/platform-browser/src/dom/events/key_even
...
+        };
+        zone.runOutsideAngular(() => {
+          manager.addEventListener(element, 'keydown.enter', handler);
+        });
+        const evt = createKeyboardEvent('keydown', {bubbles: true, cancelable: true, key: 'enter'});
+        getDOM().dispatchEvent(element, evt);
+        expect(isInNgZone).toEqual(`key handler inside ngZone: false`);
+      });
+
+      it('global key event handler should outside of the ngZone when added outside of the ngZone',
```
mhevery, line 121:
```suggestion
      it('global key event handler should be outside of the ngZone when added outside of the ngZone',
```

packages/platform-browser/testing/src/browser_util.ts
```diff
@@ -116,6 +375,58 @@ export function createMouseEvent(eventType: string): MouseEvent {
   return evt;
 }
 
+export function getKeyCode(key?: string) {
+  if (!key) {
+    return -1;
+  }
+  for (let i = 0; i < keyCodeMapping.length; i++) {
```
mhevery, line 382: instead of having a big loop, would it not be more intuitive to store `keyCodeMapping` as a map?

packages/platform-browser/testing/src/browser_util.ts
```diff
@@ -116,6 +375,58 @@ export function createMouseEvent(eventType: string): MouseEvent {
...
+    if (keyCodeMapping[i] === key.toUpperCase()) {
+      return i;
+    }
+  }
+  return -1;
+}
+
+export function createKeyboardEvent(
```
mhevery, line 390: This seems like a public API. Can we have proper jsdocs for it?

packages/platform-browser/src/dom/events/key_events.ts
```diff
@@ -167,13 +164,12 @@ export class KeyEventsPlugin extends EventManagerPlugin {
    * Configures a handler callback for a key event.
    * @param fullKey The event name that combines all simultaneous keystrokes.
    * @param handler The function that responds to the key event.
-   * @param zone The zone in which the event occurred.
    * @returns A callback function.
    */
-  static eventCallback(fullKey: any, handler: Function, zone: NgZone): Function {
-    return (event: any /** TODO #9100 */) => {
+  static eventCallback(fullKey: any, handler: Function): Function {
+    return function(this: unknown, event: any /** TODO #9100 */) {
```
mhevery, line 170: could we resolve the todo?

mhevery (changes requested):

jessicajaniuk:
@JiaLiPassion Any updates on this PR? Is it still needed?

packages/platform-browser/src/dom/events/event_manager.ts
```diff
@@ -47,7 +47,16 @@ export class EventManager {
    */
   addEventListener(element: HTMLElement, eventName: string, handler: Function): Function {
     const plugin = this._findPluginFor(eventName);
-    return plugin.addEventListener(element, eventName, handler);
+    if (NgZone.isInAngularZone()) {
```
devversion, line 50: How helpful is it actually knowing that the event is "registered" in the proper zone? Couldn't the event plugins still somehow mess this up, depending on how the event is registered (it might not be patched by zonejs for e.g. hammerjs right?)

TL;DR: Would it make sense to always run the handler guarded in the zone?

devversion (dismissed):
LGTM conceptually. Nice 👍 

@JiaLiPassion should we also remove the `runGuarded` calls in other event plugins then?

dylhunn (approved):
reviewed-for: fw-core

packages/platform-browser/src/dom/events/event_manager.ts
JiaLiPassion: @devversion, I updated the PR to include several fixes.
And the reason I didn't use 
```
ngZone.runOutside(() => {
  addEventListener(element, eventName, () => {
     ngZone.run(() => {listener(eventName});
  });
});
```
to do this because there are several issues with this solution.

1. this solution will bypass all functionality of `zone.js`, zone.js will not be able to moniter the `eventTask`, so the `onScheduleTask`, `onInvokeTask` will not be triggered if some user want to moniter.
2. now `zone.js` support `UNPATCHED_EVENTS` feature, the user can define `UNPATCHED_EVENTS=['mouseover', 'scroll']` so the `zone.js` will use the `native addEventListener` for these events, and the listeners will not trigger change detection. So if we are using the solution above, this feature will be lost.

So to still use `zone.js` functionality, the behavior should be ensure the `addEventListener` happened inside of `NgZone`.

packages/core/src/application_ref.ts
```diff
@@ -831,6 +829,15 @@ export class ApplicationRef {
    * detection pass during which all change detection must complete.
    */
   tick(): void {
+    if (NgZone.isInAngularZone()) {
+      this._tick();
+    } else {
+      this._zone.run(() => this._tick());
```
devversion, line 835: This is related to my previous comment (which I think was a little confusingly phrased): Is there any benefit in checking with `NgZone.isInAngularZone()`, or could we just always run it in the zone? like:

```ts
function tick(): void {
  this._zone.run(() => this._tick());
}
```

packages/platform-browser/src/dom/dom_renderer.ts
```diff
@@ -55,7 +55,7 @@ function decoratePreventDefault(eventHandler: Function): Function {
   // decoratePreventDefault or is a listener added outside the Angular context so it can handle the
   // two differently. In the first case, the special '__ngUnwrap__' token is passed to the unwrap
   // the listener (see below).
-  return (event: any) => {
+  const listener: Function = (event: any) => {
     // Ivy uses '__ngUnwrap__' as a special token that allows us to unwrap the function
     // so that it can be invoked programmatically by `DebugNode.triggerEventHandler`. The debug_node
     // can inspect the listener toString contents for the existence of this special token. Because
```
devversion, line 61: Can this logic below `event === '__ngUnwrap__'` be removed now?

packages/platform-browser/src/dom/events/event_manager.ts
```diff
@@ -108,4 +108,15 @@ export abstract class EventManagerPlugin {
     }
     return this.addEventListener(target, eventName, handler);
   }
+
+  ensureAddListenerInZone(element: HTMLElement, eventName: string, handler: Function): Function {
```
devversion, line 112: If you would be okay with it: I think we should add a `JSDoc` description here explaining why this is desirable. It's surely obvious to us from the context of your PR but this would be helpful long-term I feel like

devversion (commented):
Seems very reasonable to me. Just added one question (similar to my previous) comment, wondering if we could simplify code a little more

packages/platform-browser/src/dom/events/event_manager.ts
devversion: @JiaLiPassion all of this makes sense! I think my comment was a little ambiguous/unclear. I was more asking why we would not always do the following:

```js
ngZone.run(() => 
  plugin.addEventListener(element, eventName, handlerEnsureInZone));
```

I added a new comment for it though. Thanks for the reply/and details.

dylhunn (dismissed):
reviewed-for: fw-core

packages/core/src/application_ref.ts
JiaLiPassion: @devversion , the reason not to use `this._zone.run()` directly is if we use `this.zone.run()` even we are already inside the `NgZone`, it will trigger another `ApplicationRef.tick()` and cause the `NG0101 nested ApplicationRef.tick()` error.

packages/platform-browser/src/dom/dom_renderer.ts
JiaLiPassion: I am just concerned about that some app or the dev tools are using `eventListener('__ngUnwrap__')` to retrive the original listener, I will remove it and if the `g3` is ok, I think it is safe.

packages/platform-browser/src/dom/events/event_manager.ts
JiaLiPassion: Sure, I just add the `JSDoc` to explain this function.

AndrewKushnir:
[Global Presubmit](https://test.corp.google.com/OCL:431828588:BASE:431828730:1646190511323:68bf6ddf).

AndrewKushnir:
FYI, based on a global presubmit, there are ~500 failing targets, so I'm adding the "blocked" label for now.

packages/platform-browser/src/dom/dom_renderer.ts
devversion: 👍

packages/core/src/application_ref.ts
devversion: interesting, good thing to have this noted here 👍

PR #38273, opened 2020-07-28
fix(compiler-cli): type-check inputs that include undefined when there's coercion members

JoostK:
For attribute bindings that target a directive's input, the template
type checker is able to verify that the type of the input expression is
compatible with the directive's declaration for said input. This
checking adheres to the `strictNullChecks` flag as configured in the
TypeScript compilation, such that errors are reported for expressions
that include `undefined` or `null` in their type if the input's
declaration does not include those types.

There was a bug with this level of type-checking for directives that
also declare coercion members, where binding an expression that includes
the `undefined` type to a directive's input that does not include the
`undefined` type would not be reported as error.

This commit fixes the bug by changing the type-constructor in type-check
code to use an intersection type of regular inputs and coerced inputs,
instead of a union type. The union type would inadvertently allow
`undefined` types to be assigned into the regular inputs, as that would
still satisfy the characteristics of a union type.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,8 +32,8 @@ var_4_win: &cache_key_win_fallback v7-angular-win-node-12-{{ checksum ".bazelver
 
 # Cache key for the `components-repo-unit-tests` job. **Note** when updating the SHA in the
 # cache keys also update the SHA for the "COMPONENTS_REPO_COMMIT" environment variable.
-var_5: &components_repo_unit_tests_cache_key v7-angular-components-f428c00465dfcf8a020237f22532480eedbd2cb6
-var_6: &components_repo_unit_tests_cache_key_fallback v7-angular-components-
+var_5: &components_repo_unit_tests_cache_key v9-angular-components-09e68db8ed5b1253f2fe38ff954ef0df019fc25a
+var_6: &components_repo_unit_tests_cache_key_fallback v9-angular-components-
 
 # Workspace initially persisted by the `setup` job, and then enhanced by `build-npm-packages` and
 # `build-ivy-npm-packages`.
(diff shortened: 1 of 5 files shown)


atscott:
[presubmit](https://test.corp.google.com/ui#id=OCL:323823951:BASE:323824019:1596045746186:3f0c8d92)

atscott (changes requested):
@JoostK - Can you update the commit message to include a small note about how to get around this, either by using a non-null assertion operator or by turning off `strictNullInputTypes` (?). In addition, maybe adding another test to demonstrate that `strictNullInputTypes: false` (again, not sure if this is the correct option) does not produce an error in this case.

atscott (approved):

devversion (approved):
LGTM for dev-infra.

@JoostK When you get a chance (for later reference): could you briefly explain why we go from v7 to v9 in the cache key?

JoostK:
D'oh! I meant to do that but then completely forgot about it.

I used the v8 key for testing against the required commit using my fork, to verify whether the components repo PR would actually build successfully. Upon switching back to the primary repo, the v8 fallback cache was poisoned with a checkout of my fork.

AndrewKushnir:
Hi @JoostK, thanks for fixing the issue! 👍 

Quick question: should this PR be also included into the patch branch, since this is a fix (it has a `master-only` PR target atm)?

Thank you.

atscott:
@AndrewKushnir We merged to master only out of an abundance of caution. It's technically a breaking change because it can produce new errors that weren't detected before. Those errors _should_ have already been appearing though, so this is a bug fix but also breaking.

atscott:
@JoostK Speaking of breaking, should we have added the `BREAKING CHANGE:` footnote?

AndrewKushnir:
Thanks for additional information @atscott, that makes sense.

JoostK:
@AndrewKushnir the change to update the components repo commit should also be in patch, so I opened #38378 to land that in patch.

@atscott I intentionally did _not_ include an explicit breaking changes note, as we considered this as a true bug fix that can land anytime, without constraining ourselves to targeting it for a major. I guess we don't really have a proper communication channel to give people a heads up about these kinds of changes; the callout in the changelog would still be useful, but currently that can only be done using a breaking change note (AFAIK).

AndrewKushnir:
Thanks @JoostK 👍

PR #38279, opened 2020-07-28
refactor(compiler-cli): only use type constructors for directives with generic types

### LINKED ISSUE #38185: Ivy LS: refactor type constructors and input binding checking

atscott:
Should wait on #38273, as this addresses the same issue when the Directive/Component is not generic (since non-generic Directive/Components will no longer use the type constructor)

### DIFF
--- a/packages/compiler-cli/src/ngtsc/metadata/src/util.ts
+++ b/packages/compiler-cli/src/ngtsc/metadata/src/util.ts
@@ -12,7 +12,7 @@ import {Reference} from '../../imports';
 import {ClassDeclaration, ClassMember, ClassMemberKind, isNamedClassDeclaration, ReflectionHost, reflectTypeEntityToDeclaration} from '../../reflection';
 import {nodeDebugInfo} from '../../util/src/typescript';
 
-import {DirectiveMeta, MetadataReader, NgModuleMeta, PipeMeta, TemplateGuardMeta} from './api';
+import {DirectiveMeta, DirectiveTypeCheckMeta, MetadataReader, NgModuleMeta, PipeMeta, TemplateGuardMeta} from './api';
 
 export function extractReferencesFromType(
     checker: ts.TypeChecker, def: ts.TypeNode, ngModuleImportedFrom: string|null,
@@ -78,13 +78,15 @@ export function readStringArrayType(type: ts.TypeNode): string[] {
   return res;
 }
 
-
-export function extractDirectiveGuards(node: ClassDeclaration, reflector: ReflectionHost): {
-  ngTemplateGuards: TemplateGuardMeta[],
-  hasNgTemplateContextGuard: boolean,
-  coercedInputFields: Set<string>,
-} {
-  const staticMembers = reflector.getMembersOfClass(node).filter(member => member.isStatic);
+/**
+ * TODO: rename to `extractDirectiveInfo` and rename the `guards` field in
+ * `DirectiveHandlerData`/`ComponentAnalysisData` to a more appropriate name.
+ */
+export function extractDirectiveGuards(
+    node: ClassDeclaration, inputs: {[fieldName: string]: string|[string, string]},
+    reflector: ReflectionHost): DirectiveTypeCheckMeta {
+  const members = reflector.getMembersOfClass(node);
+  const staticMembers = members.filter(member => member.isStatic);
   const ngTemplateGuards = staticMembers.map(extractTemplateGuard)
                                .filter((guard): guard is TemplateGuardMeta => guard !== null);
   const hasNgTemplateContextGuard = staticMembers.some(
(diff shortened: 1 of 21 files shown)


packages/compiler-cli/src/ngtsc/metadata/src/util.ts
```diff
@@ -107,8 +107,9 @@ export function extractDirectiveTypeCheckMeta(
       undeclaredInputFields.add(fieldName);
     } else if (isRestricted(field.node)) {
       restrictedInputFields.add(fieldName);
-    } else if (hasGenericType(field.node)) {
-      genericInputFields.add(fieldName);
+      if (field.nameNode !== null && ts.isStringLiteral(field.nameNode)) {
+        stringLiteralInputFields.add(fieldName);
+      }
```
JoostK, line 112: I'm surprised this currently works (?) as I suppose this should be outside of this `isRestricted` block?

packages/compiler-cli/src/ngtsc/typecheck/src/api.ts
```diff
@@ -90,6 +90,8 @@ export interface TypeCheckingConfig {
    */
   checkTypeOfInputBindings: boolean;
 
+  checkAccessModifiersForInputBindings: boolean;
```
JoostK, line 93: Needs docs

packages/compiler-cli/src/ngtsc/typecheck/src/type_check_block.ts
```diff
@@ -456,6 +451,31 @@ class TcbDirectiveInputsOp extends TcbOp {
...
+            const id = this.tcb.allocateId();
+            const dirTypeRef = this.tcb.env.referenceType(this.dir.ref);
+            if (!ts.isTypeReferenceNode(dirTypeRef)) {
+              throw new Error(
+                  `Expected TypeReferenceNode from reference to ${this.dir.ref.debugName}`);
+            }
+            const type = ts.createIndexedAccessTypeNode(
+                ts.createTypeQueryNode(dirId as ts.Identifier),
+                ts.createLiteralTypeNode(ts.createStringLiteral(fieldName)));
+            const temp = tsCreateVariable(id, ts.createNonNullExpression(ts.createNull()), type);
```
JoostK, line 473: This can probably be:

```suggestion
            const temp = tsDeclareVariable(id, type);
```

packages/compiler-cli/src/ngtsc/typecheck/src/type_check_block.ts
```diff
@@ -456,6 +451,31 @@ class TcbDirectiveInputsOp extends TcbOp {
...
+            addParseSpanInfo(temp, input.attribute.sourceSpan);
```
JoostK, line 474: I expect this parse span annotation to be irrelevant, as there shouldn't be any diagnostics produced for this statement.

packages/compiler-cli/src/ngtsc/metadata/src/util.ts
atscott: You're right that it should be pulled out of `isRestricted`. The reason it worked though is that there's only special handling for stringLiteral fields when the field is restricted. Only when a field is restricted do we try to write `t2.fieldName` to get the correct diagnostic for assigning to protected/private/readonly. Otherwise, we just use `t2["fieldName"]` so there doesn't need to be special handling.

atscott:
[presubmit](https://test.corp.google.com/ui#id=OCL:323678011:BASE:323681270:1595980204033:e5aa2f8e)

packages/compiler-cli/src/ngtsc/metadata/src/util.ts
```diff
@@ -93,7 +96,45 @@ export function extractDirectiveGuards(node: ClassDeclaration, reflector: Reflec
...
       new Set(staticMembers.map(extractCoercedInput)
                   .filter((inputName): inputName is string => inputName !== null));
-  return {hasNgTemplateContextGuard, ngTemplateGuards, coercedInputFields};
+
+  const restrictedInputFields = new Set<string>();
+  const stringLiteralInputFields = new Set<string>();
+  const undeclaredInputFields = new Set<string>();
+
+  for (const fieldName of Object.keys(inputs)) {
+    const field = members.find(member => member.name === fieldName);
```
JoostK, line 105: Note to self: filter out getters to avoid being sensitive to ordering, i.e. a public getter and private setter must be recorded as restricted, regardless of whether the setter or getter is ordered first.

```suggestion
    const field = members.find(member => member.name === fieldName && member.kind !== ClassMemberKind.Getter);
```

packages/compiler-cli/src/ngtsc/metadata/src/util.ts
```diff
@@ -93,7 +96,45 @@ export function extractDirectiveGuards(node: ClassDeclaration, reflector: Reflec
...
+    if (field === undefined || field.node === null) {
+      undeclaredInputFields.add(fieldName);
+    } else if (isRestricted(field.node)) {
+      restrictedInputFields.add(fieldName);
+    }
+    if (field && field.nameNode !== null && ts.isStringLiteral(field.nameNode)) {
+      stringLiteralInputFields.add(fieldName);
+    }
```
JoostK, line 113: nit:

```suggestion
    } else {
      if (isRestricted(field.node)) {
        restrictedInputFields.add(fieldName);
      }
      if (ts.isStringLiteral(field.nameNode)) {
        stringLiteralInputFields.add(fieldName);
      }
    }
```

(or maybe even insert a `continue` statement in the first if-block)

packages/compiler-cli/src/ngtsc/metadata/src/util.ts
atscott: Quick note that this has changed since the comment. Alex suggested that we instead generate `t2.fieldName` in most cases and only use `t2["fieldName"]` in the special case where the field is a string literal _or_ the user opts out of the option to honor access modifiers on the input.

packages/compiler-cli/src/ngtsc/metadata/src/util.ts
JoostK: Alright, so I tried adding a test for this scenario and found out that a getter/setter pair needs to agree on a visibility, so this point is moot.

PR #38280, opened 2020-07-29
refactor(forms): move common validators-related logic to the `AbstractControlDirective` class

AndrewKushnir:
This commit refactors validators-related logic that is common across most of the directives.

A couple notes on this refactoring:
* common logic was moved to the `AbstractControlDirective` class (including `validator` and
`asyncValidator` getters)
* sync/async validators are now composed in `AbstractControlDirective` class eagerly when validators
are set with `_setValidators` and `_setAsyncValidators` calls and the result is stored in directive
instance (thus getters return cached versions of validator fn). This is needed to make sure composed
validator function remains the same (retains its identity) for a given directive instance, so that
this function can be added and later removed from an instance of an AbstractControl-based class
(like `FormControl`). Preserving validator function is required to perform proper cleanup (in followup
PRs) of the AbstractControl-based classes when a directive is destroyed.

From **public API** perspective, this refactoring should be a backwards-compatible change, since `validator ` and `asyncValidator ` getters were just moved to the base class.

Regarding the **circular deps golden file** update: this change is an improvement removing 4 cycles, the rest is an update of existing cycles making them more straightforward to investigate and fix in the future.

## PR Type

- [ ] Yes
- [x] No

### DIFF
--- a/packages/forms/src/directives/abstract_control_directive.ts
+++ b/packages/forms/src/directives/abstract_control_directive.ts
@@ -7,8 +7,12 @@
  */
 
 import {Observable} from 'rxjs';
+
 import {AbstractControl} from '../model';
-import {ValidationErrors} from './validators';
+import {composeAsyncValidators, composeValidators} from '../validators';
+
+import {AsyncValidator, AsyncValidatorFn, ValidationErrors, Validator, ValidatorFn} from './validators';
+
 
 /**
  * @description
@@ -163,6 +167,67 @@ export abstract class AbstractControlDirective {
    */
   get path(): string[]|null {
     return null;
+  }
+
+  /**
+   * Contains the result of merging synchronous validators into a single validator function
+   * (combined using `Validators.compose`).
+   */
+  private _composedValidatorFn: ValidatorFn|null|undefined;
+
+  /**
+   * Contains the result of merging asynchronous validators into a single validator function
+   * (combined using `Validators.composeAsync`).
+   */
+  private _composedAsyncValidatorFn: AsyncValidatorFn|null|undefined;
+
+  /**
+   * Set of synchronous validators as they were provided while calling `setValidators` function.
+   * @internal
+   */
+  _rawValidators: Array<Validator|ValidatorFn> = [];
+
+  /**
+   * Set of asynchronous validators as they were provided while calling `setAsyncValidators`
+   * function.
+   * @internal
+   */
+  _rawAsyncValidators: Array<AsyncValidator|AsyncValidatorFn> = [];
+
+  /**
+   * Sets synchronous validators for this directive.
+   * @internal
+   */
+  _setValidators(validators: Array<Validator|ValidatorFn>|undefined): void {
+    this._rawValidators = validators || [];
+    this._composedValidatorFn = composeValidators(this._rawValidators);
+  }
+
+  /**
+   * Sets asynchronous validators for this directive.
+   * @internal
+   */
+  _setAsyncValidators(validators: Array<AsyncValidator|AsyncValidatorFn>|undefined): void {
+    this._rawAsyncValidators = validators || [];
+    this._composedAsyncValidatorFn = composeAsyncValidators(this._rawAsyncValidators);
+  }
+
+  /**
+   * @description
+   * Synchronous validator function composed of all the synchronous validators registered with this
+   * directive.
+   */
+  get validator(): ValidatorFn|null {
+    return this._composedValidatorFn || null;
+  }
+
+  /**
+   * @description
+   * Asynchronous validator function composed of all the asynchronous validators registered with
+   * this directive.
+   */
+  get asyncValidator(): AsyncValidatorFn|null {
+    return this._composedAsyncValidatorFn || null;
   }
 
   /**
--- a/goldens/circular-deps/packages.json
+++ b/goldens/circular-deps/packages.json
@@ -1003,37 +1003,13 @@
     "packages/core/testing/src/test_bed.ts"
   ],
   [
-    "packages/forms/src/directives/abstract_control_directive.ts",
-    "packages/forms/src/model.ts",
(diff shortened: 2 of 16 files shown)


AndrewKushnir:
[Global Presubmit](https://test.corp.google.com/OCL:323723039:BASE:323723137:1596001331788:33b8c967).

AndrewKushnir:
[Initial Presubmit](http://test/OCL:323723039:BASE:323887803:1596063791529:28fda448).

jelbourn (approved):
LGTM

Reviewed-for: public-api

atscott (approved):
Reviewed-for: public-api

Hey Andrew, you mentioned that there will be follow-up PRs for more cleanup. I was wondering if these changes (and the follow ups) have an issue report that you can link to. It would be great to get a little more info about the broader context of why these changes are being made.

sonukapoor (approved):

sonukapoor:
@AndrewKushnir Can you rebase and resolve conflicts?

petebacondarwin (commented):
Reviewed-for: public-api

From a backward compatibility point of view, this could actually be breaking if someone (outside the framework) has a class that derives from `AbstractControlDirective` but does not currently provide the `validator` and `asyncValidator` properties...  Is this likely?

At the least we should check CDK and perhaps some of the more popular forms libraries.

AndrewKushnir:
Hi @petebacondarwin, thanks for the comment.

You are right, the `AbstractControlDirective` abstract class can be used as a base class for other directives outside of the framework. However the `validator` and `asyncValidator` getters have implementations and are not marked as `abstract`, so child classes should not require these fields to be re-implemented. Directives in Forms package that extend the `AbstractControlDirective` abstract class don't have their specific implementations of `validator` and `asyncValidator` functions. Could you please let me know if I understood your comment correctly?

Thank you.

petebacondarwin (approved):
Cripes! For some reason I was thinking these were abstract methods. You are quite right. Now worries from me.

Reviewed-for: public-api

AndrewKushnir:
[Presubmit](http://test/OCL:332759195:BASE:332759193:1600649602569:3eac7c6d) + [Global Presubmit](https://test.corp.google.com/OCL:332759195:BASE:332759284:1600649674833:77e7931c).

AndrewKushnir:
[Presubmit](http://test/OCL:332759195:BASE:338883763:1603602178310:d23d4c8c) + [Global Presubmit](https://test.corp.google.com/OCL:332759195:BASE:338883894:1603602270521:696536cd).

packages/forms/test/validators_spec.ts
```diff
@@ -8,7 +8,8 @@
 
 import {fakeAsync, tick} from '@angular/core/testing';
 import {describe, expect, it} from '@angular/core/testing/src/testing_internal';
-import {AbstractControl, AsyncValidator, AsyncValidatorFn, FormArray, FormControl, ValidationErrors, ValidatorFn, Validators} from '@angular/forms';
+import {AbstractControl, AsyncValidatorFn, FormArray, FormControl, Validators} from '@angular/forms';
+import {AsyncValidator, ValidationErrors, ValidatorFn} from '@angular/forms/src/directives/validators';
```
petebacondarwin, line 12: These are exported from `@angular/forms` so there is no need to import them via this deep import

packages/forms/src/directives/abstract_control_directive.ts
```diff
@@ -165,6 +169,69 @@ export abstract class AbstractControlDirective {
     return null;
   }
 
+  /**
+   * Contains the result of merging synchronous validators into a single validator function
+   * (combined using `Validators.compose`).
+   * @internal
```
petebacondarwin, line 175: Since this is private, it doesn't need the `@internal` tag. Same for `_composedAsyncValidatorFn`.

packages/forms/src/directives/abstract_control_directive.ts
```diff
@@ -165,6 +169,69 @@ export abstract class AbstractControlDirective {
...
+   * (combined using `Validators.composeAsync`).
+   * @internal
+   */
+  private _composedAsyncValidatorFn: AsyncValidatorFn|null|undefined;
+
+  /**
+   * Set of synchronous validators as they were provided while calling `setValidators` function.
+   * @internal
+   */
+  _rawValidators: Array<Validator|ValidatorFn> = [];
```
petebacondarwin, line 190: I note that this and `_rawAsyncValidators` are internal purely to allow the `setUpControl()` shared function to register `OnValidatorChange` handlers.

I wonder if, instead, we should add an (internal?) method to `AbstractControlDirective` that will hide this internal representation:

```
registerValidatorChangeHandler(handler: () => void) {
  this._rawValidators.forEach(validator => {
    if (isValidator(validator)).registerOnValidatorChange)
      validator.registerOnValidatorChange!(handler);
  });

  dir._rawAsyncValidators.forEach(validator => {
    if (isValidator(validator)).registerOnValidatorChange)
      validator.registerOnValidatorChange!(handler);
  });
}
```

Then in the shared.ts file we just have:

```
dir.registerOnValidatorChange(() => control.updateValueAndVisibility());
```

petebacondarwin (approved):
A couple of nits but otherwise LGTM!

packages/forms/src/directives/abstract_control_directive.ts
AndrewKushnir: Thanks for the proposal @petebacondarwin 👍 
This PR preserves the original logic and I have a couple followup PRs where the logic of managing validators will be updated. I will look into a more straightforward API that you proposed as an alternative to having these validators exposed via `_rawValidators` and `_rawAsyncValidators` fields.

AndrewKushnir:
[Presubmit #2](http://test/OCL:332759195:BASE:339372625:1603848587628:43320de0).

PR #38294, opened 2020-07-30
test(compiler-cli): disable one TypeChecker test on Windows due to path sensitivity issue

AndrewKushnir:
This commit disables one TypeChecker test (added as a part of
https://github.com/angular/angular/pull/38105) which make assertions about the filename while
running on Windows.

Such assertions are currently suffering from a case sensitivity issue.

## PR Type

- [x] Other... Please describe: disabling currently failing tests on Windows

- [ ] Yes
- [x] No

### DIFF
--- a/packages/compiler-cli/src/ngtsc/typecheck/test/type_checker_spec.ts
+++ b/packages/compiler-cli/src/ngtsc/typecheck/test/type_checker_spec.ts
@@ -6,6 +6,8 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
+import {platform} from 'os';
+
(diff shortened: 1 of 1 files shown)


AndrewKushnir:
@alxhub FYI, I removed the "PR target" for now, since we may want to roll back the entire PR https://github.com/angular/angular/pull/38105 from patch branch, since some code there relies on the code that is present in master only. If we would revert PR https://github.com/angular/angular/pull/38105 from the patch branch, we'll need to land this change only in master.

alxhub (approved):

PR #38304, opened 2020-07-30
feat(dev-infra): save invalid commit message attempts to be restored on next commit attempt

josephperrott:
When a commit message fails validation, rather than throwing out the commit message entirely
the commit message is saved into a draft file and restored on the next commit attempt.

### DIFF
--- a/dev-infra/commit-message/cli.ts
+++ b/dev-infra/commit-message/cli.ts
@@ -18,9 +18,24 @@ export function buildCommitMessageParser(localYargs: yargs.Argv) {
   return localYargs.help()
       .strict()
       .command(
-          'restore-commit-message-draft [filePath] [source] [_]', false, {},
+          'restore-commit-message-draft', false, {
+            'file-env-variable': {
+              type: 'string',
+              conflicts: ['file'],
+              required: true,
+              description:
+                  'The key of the environment variable for the path of the commit message file.',
+              coerce: arg => {
+                const [file, source] = process.env[arg].split(' ');
+                if (!file) {
+                  throw new Error(`Provided environment variable "${arg}" was not found.`);
+                }
+                return [file, source];
+              },
+            }
+          },
           args => {
-            restoreCommitMessage(args.filePath, args.source);
+            restoreCommitMessage(args.fileEnvVariable[0], args.fileEnvVariable[1]);
           })
       .command(
           'pre-commit-validate', 'Validate the most recent commit message', {
--- a/dev-infra/commit-message/cli.ts
+++ b/dev-infra/commit-message/cli.ts
@@ -24,9 +24,11 @@ export function buildCommitMessageParser(localYargs: yargs.Argv) {
               conflicts: ['file'],
               required: true,
               description:
-                  'The key of the environment variable for the path of the commit message file.',
+                  'The key for the environment variable which holds the arguments for the ' +
+                  'prepare-commit-msg hook as described here: ' +
+                  'https://git-scm.com/docs/githooks#_prepare_commit_msg',
               coerce: arg => {
-                const [file, source] = process.env[arg].split(' ');
+                const [file, source] = (process.env[arg] || '').split(' ');
                 if (!file) {
                   throw new Error(`Provided environment variable "${arg}" was not found.`);
                 }
--- a/dev-infra/commit-message/cli.ts
+++ b/dev-infra/commit-message/cli.ts
@@ -9,6 +9,7 @@ import * as yargs from 'yargs';
 
 import {info} from '../utils/console';
 
+import {restoreCommitMessage} from './restore-commit-message';
 import {validateFile} from './validate-file';
 import {validateCommitRange} from './validate-range';
 
@@ -16,6 +17,11 @@ import {validateCommitRange} from './validate-range';
 export function buildCommitMessageParser(localYargs: yargs.Argv) {
   return localYargs.help()
       .strict()
+      .command(
+          'restore-commit-message-draft [filePath] [source] [_]', false, {},
+          args => {
+            restoreCommitMessage(args.filePath, args.source);
+          })
       .command(
           'pre-commit-validate', 'Validate the most recent commit message', {
             'file': {
--- a/dev-infra/commit-message/commit-message-draft.ts
+++ b/dev-infra/commit-message/commit-message-draft.ts
@@ -7,24 +7,24 @@
  */
 import {existsSync, readFileSync, unlinkSync, writeFileSync} from 'fs';
 
-const SAVED_COMMIT_MSG_FILE_PATH = '.git/COMMIT_EDITMSG.ngDevSave';
-
 /** Load the commit message draft from the file system if it exists. */
-export function loadCommitMessageDraft() {
-  if (existsSync(SAVED_COMMIT_MSG_FILE_PATH)) {
-    return readFileSync(SAVED_COMMIT_MSG_FILE_PATH).toString();
+export function loadCommitMessageDraft(basePath: string) {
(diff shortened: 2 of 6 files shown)


josephperrott:
Setting @alxhub as a reviewer as this feature was at his request.  Can you confirm this is what you were looking for?

dev-infra/commit-message/commit-message-draft.ts
```diff
@@ -0,0 +1,30 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import {existsSync, readFileSync, unlinkSync, writeFileSync} from 'fs';
+
+const SAVED_COMMIT_MSG_FILE_PATH = '.git/COMMIT_EDITMSG.ngDevSave';
```
gkalpak, line 10: This path will not necessarily work correctly in a [git worktree][1].

[1]: https://git-scm.com/docs/git-worktree

package.json
```diff
@@ -206,7 +206,8 @@
   "cldr-data-coverage": "full",
   "husky": {
     "hooks": {
-      "commit-msg": "yarn -s ng-dev commit-message pre-commit-validate --file-env-variable HUSKY_GIT_PARAMS"
+      "commit-msg": "yarn -s ng-dev commit-message pre-commit-validate --file-env-variable HUSKY_GIT_PARAMS",
+      "prepare-commit-msg": "yarn -s ng-dev commit-message restore-commit-message-draft ${HUSKY_GIT_PARAMS}"
```
gkalpak, line 210: I am confused on how this works.

If I understand correctly, the current implementation restores the previous (invalid) message to `.git/COMMIT_EDITMSG`. But from what I can tell, (1) having something in `.git/COMMIT_EDITMSG` does not cause git to automatically pre-populate the commit message with it and (2) git stores the last entered commit message in `.git/COMMIT_EDITMSG` anyway. (We even used to mention that in the previous validation tool: ab30b2ad010f37f65e1689700a20b506b21076a4)

Is this different from what people see?

##
BTW, if you are using the command line, you can pre-populate the commit message from a file with the [--template/-t][1] option (e.g. `git commit -t .git/COMMIT_EDITMSG`). In the previous validation tool, we 

[1]: https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---templateltfilegt

gkalpak (commented):
:tada:

package.json
josephperrott: As currently written, we will automatically restore the previous commit message only if it was a failure.

On failing commit messages, we save it to `{providedPath}.ngDevSave` (typically `.git/COMMIT_EDITMSG.ngDevSave`) and then restore it on the next attempt.

The intention here is for the message to be automatically resumed and by the time `prepare-commit-msg` is triggered, the old `COMMIT_EDITMSG` value is already gone unless provided by the user.

gkalpak: Yes, I get that. What I don't understand if why:
1. Isn't the message already in `.git/COMMIT_EDITMSG`?
2. Does git pre-populates the commit message in the editor with what is in `.git/COMMIT_EDITMSG` for you (because it doesn;t for me)?

josephperrott: 1. The message is in `.git/COMMIT_EDITMSG` until immediately before the `prepare-commit-msg` hook.  From the docs:
> [`prepare-commit-msg`] is invoked by git-commit[1] right after preparing the default log message, and before the editor is started.

So the message is no longer available at the time of the hook.

2. It does not, you can prepopulate it with `git commit -t .git/COMMIT_EDITMSG`, however this will just use the last commit message regardless of its previous success.

gkalpak: Oh, so it sounds like the `.git/COMMIT_EDITMSG` file is overwritten by git with the default message before the `prepare-commit-msg` hook is invoked. Makes sense now. Thx for clarifying.

package.json
```diff
@@ -206,7 +206,8 @@
   "cldr-data-coverage": "full",
   "husky": {
     "hooks": {
-      "commit-msg": "yarn -s ng-dev commit-message pre-commit-validate --file-env-variable HUSKY_GIT_PARAMS"
+      "commit-msg": "yarn -s ng-dev commit-message pre-commit-validate --file-env-variable HUSKY_GIT_PARAMS",
+      "prepare-commit-msg": "yarn -s ng-dev commit-message restore-commit-message-draft  --file-env-variable HUSKY_GIT_PARAMS"
```
devversion, line 210:
```suggestion
      "prepare-commit-msg": "yarn -s ng-dev commit-message restore-commit-message-draft --file-env-variable HUSKY_GIT_PARAMS"
```

dev-infra/commit-message/cli.ts
```diff
@@ -9,13 +9,34 @@ import * as yargs from 'yargs';
...
   return localYargs.help()
       .strict()
+      .command(
+          'restore-commit-message-draft', false, {
+            'file-env-variable': {
+              type: 'string',
+              conflicts: ['file'],
+              required: true,
+              description:
+                  'The key of the environment variable for the path of the commit message file.',
```
devversion, line 27: Should we link here to the description of what the arguments will be about (in the variable)? https://git-scm.com/docs/githooks#_prepare_commit_msg

dev-infra/commit-message/restore-commit-message.ts
```diff
@@ -0,0 +1,47 @@
...
+      info('A commit message was already provided via the command with a -m or -F flag');
+    }
+    if (source === 'template') {
+      info('A commit message was already provided via the -t flag or config.template setting');
+    }
+    if (source === 'squash') {
+      info('A commit message was already provided as a merge action or via .git/MERGE_MSG');
+    }
+    if (source === 'commit') {
+      info('A commit message was already provided via another commit via --fixup, -c, -C');
```
devversion, line 33: Nit suggestion (if it makes sense): Might sound little better if we avoid the double `via`:

```suggestion
      info('A commit message was already provided through a revision specified via --fixup, -c, -C');
```

dev-infra/commit-message/restore-commit-message.ts
```diff
@@ -0,0 +1,47 @@
...
+import {resolve} from 'path';
+
+import {getRepoBaseDir} from '../utils/config';
+
+import {loadCommitMessageDraft} from './commit-message-draft';
+
+/**
+ * Restore the commit message draft to the git to be used as the default commit message.
+ */
+export function restoreCommitMessage(filePath: string, source: string) {
```
devversion, line 20: Can we type `source` here? or at least document what it will be about in the JSDoc (ideally linking to the githooks documentation too)

dev-infra/commit-message/restore-commit-message.ts
```diff
@@ -0,0 +1,47 @@
...
+      info('or --amend flag');
+    }
+    process.exit(0);
+  }
+  /** A draft of a commit message. */
+  const commitMessage = loadCommitMessageDraft(filePath);
+
+  // If the commit message draft has content, restore it into the provided filepath.
+  if (commitMessage) {
+    writeFileSync(resolve(getRepoBaseDir(), filePath), commitMessage);
```
devversion, line 43: Do we need to resolve the path here? Inside `loadCommitMessageDraft` we use the path without `getRepoBaseDir` too? I think the git hook has the proper working directory, so we can omit the `resolve` here.

devversion (approved):
Nice! A few minor comments.

gkalpak (approved):

M-050:
في خميس، 13 أغسطس، 2020 في 2:34 م، كتب George Kalpakas <
notifications@github.com>:

PR #38305, opened 2020-07-30
fix(compiler-cli): ensure literal types are retained when `strictNullInputTypes` is disabled

JoostK:
Consider the `NgModel` directive which has the `ngModelOptions` input:

```ts
class NgModel {
  @Input() ngModelOptions: { updateOn: 'blur'|'change'|'submit' };
}
```

In a template this may be set using an object literal as follows:

```html
<input ngModel [ngModelOptions]="{updateOn: 'blur'}">
```

This assignment should be accepted, as the object's type aligns with the
`ngModelOptions` input in `NgModel`. However, if the `strictNullInputTypes`
option is disabled this assignment would inadvertently produce an error:

```
Type '{ updateOn: string; }' is not assignable to type '{ updateOn: "blur"|"change"|"submit"; }'.
  Types of property 'updateOn' are incompatible.
    Type 'string' is not assignable to type '"blur"|"change"|"submit"'
```

This is due to the `'blur'` value being inferred to be of type `string`
instead of retaining its literal type. The non-null assertion operator
that is automatically inserted for input binding assignments when
`strictNullInputTypes` is disabled inhibits TypeScript from inferring
the string value as its literal type.

This commit fixes the issue by omitting the insertion of the non-null
operator for object literals and array literals.

### DIFF
--- a/packages/compiler-cli/src/ngtsc/typecheck/src/type_check_block.ts
+++ b/packages/compiler-cli/src/ngtsc/typecheck/src/type_check_block.ts
@@ -701,16 +701,7 @@ class TcbDirectiveInputsOp extends TcbOp {
     const inputs = getBoundInputs(this.dir, this.node, this.tcb);
     for (const input of inputs) {
       // For bound inputs, the property is assigned the binding expression.
-      let expr = translateInput(input.attribute, this.tcb, this.scope);
-      if (!this.tcb.env.config.checkTypeOfInputBindings) {
-        // If checking the type of bindings is disabled, cast the resulting expression to 'any'
-        // before the assignment.
-        expr = tsCastToAny(expr);
-      } else if (!this.tcb.env.config.strictNullInputBindings) {
-        // If strict null checks are disabled, erase `null` and `undefined` from the type by
-        // wrapping the expression in a non-null assertion.
-        expr = ts.createNonNullExpression(expr);
-      }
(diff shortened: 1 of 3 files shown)


packages/compiler-cli/src/ngtsc/typecheck/src/expression.ts
```diff
@@ -170,7 +170,9 @@ class AstTranslator implements AstVisitor {
     } else if (ast.value === null) {
       node = ts.createNull();
     } else {
-      node = ts.createLiteral(ast.value);
+      // Emit a literal using an `as const` cast to force TypeScript to retain the literal type.
+      // This avoids implicit widening of the literal into its primitive type.
+      node = tsCastToConst(ts.createLiteral(ast.value));
```
JoostK, line 175: Implementation note: we could move this cast to object literals (`LiteralMap`) instead, which would also fix this issue. I'm not sure what the consequences of that would be for other scenarios, so I choose to cast all literal primitives for now.

petebacondarwin (approved):
Really nice commit. Both because it fixes the error, but also because it removes duplication and makes the code easier to understand.

AndrewKushnir:
[Presubmit](http://test/OCL:408491253:BASE:408491280:1636425040025:792cebf8).

atscott:
This PR was merged into the repository by commit 2a2744721b429cbce7fd9dabc2963c1c679b67bd.

PR #38308, opened 2020-07-31
fix(forms): remove validators while cleaning up a control

AndrewKushnir:
Prior to this commit, the `cleanUpControl` function (responsible for cleaning up control instance)
was not taking validators into account. As a result, these validators remain registered on a detached
form control instance, thus causing memory leaks. This commit updates the `cleanUpControl` function
logic to also run validators cleanup.

As a part of this change, the logic to setup and cleanup validators was refactored and moved to
separate functions, so that they can be reused in the future.

NOTE: this PR resolves memory leaks in **certain scenarios only**. Other scenarios will be handled in followup PRs.

## PR Type

- [x] Bugfix

- [ ] Yes
- [x] No

### DIFF
--- a/packages/forms/src/directives/shared.ts
+++ b/packages/forms/src/directives/shared.ts
@@ -9,7 +9,7 @@
 import {isDevMode} from '@angular/core';
 
 import {FormArray, FormControl, FormGroup} from '../model';
-import {normalizeValidators, Validators} from '../validators';
+import {Validators} from '../validators';
 
 import {AbstractControlDirective} from './abstract_control_directive';
 import {AbstractFormGroupDirective} from './abstract_form_group_directive';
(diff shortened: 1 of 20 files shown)


AndrewKushnir:
[Initial Presubmit](http://test/OCL:324127940:BASE:324127966:1596159807659:add8ce5e).

packages/forms/src/directives/shared.ts
```diff
@@ -50,36 +50,90 @@ export function setUpControl(control: FormControl, dir: NgControl): void {
...
+ * @param control Form control from where directive validators should be removed.
+ * @param dir Directive instance that contains validators to be removed.
+ * @param handleOnValidatorChange Flag that determines whether directive validators should be
+ *     cleaned up to stop handling validator input change.
+ */
+export function cleanUpValidators(
+    control: AbstractControl|null, dir: AbstractControlDirective,
+    handleOnValidatorChange: boolean): void {
+  if (control !== null) {
+    if (dir.validator !== null) {
```
sonukapoor, line 113: This `if` statement could be merged with the one line (112) above.

AndrewKushnir:
Closing this PR in favor of #39234.

PR #38310, opened 2020-07-31
refactor(language-service): [Ivy] remove temporary compiler

kyliau:
Now that Ivy compiler has a proper `TemplateTypeChecker` interface
(see https://github.com/angular/angular/pull/38105) we no longer need to
keep the temporary compiler implementation.

The temporary compiler was created to enable testing infrastructure to
be developed for the Ivy language service.

This commit removes the whole `ivy/compiler` directory and moves two
functions `createTypeCheckingProgramStrategy` and
`getOrCreateTypeCheckScriptInfo` to the `LanguageService` class.

Also re-enable the Ivy LS test since it's no longer blocking development.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No


packages/language-service/ivy/language_service.ts
```diff
@@ -66,3 +88,80 @@ export function parseNgCompilerOptions(project: ts.server.Project): CompilerOpti
...
+    getCurrentDirectory(): string {
+      return project.getCurrentDirectory();
+    },
+    getCanonicalFileName(fileName: string): string {
+      return project.projectService.toCanonicalFileName(fileName);
+    },
+  };
+}
+
+function createTypeCheckingProgramStrategy(project: ts.server.Project):
```
kyliau, line 117: this function is copied over from the deleted file `compiler.ts`

packages/language-service/ivy/language_service.ts
```diff
@@ -66,3 +88,80 @@ export function parseNgCompilerOptions(project: ts.server.Project): CompilerOpti
...
+        const scriptInfo = getOrCreateTypeCheckScriptInfo(project, fileName);
+        const snapshot = scriptInfo.getSnapshot();
+        const length = snapshot.getLength();
+        scriptInfo.editContent(0, length, newText);
+      }
+    },
+  };
+}
+
+function getOrCreateTypeCheckScriptInfo(
```
kyliau, line 142: this function is copied over from the deleted file `compiler.ts`

packages/language-service/ivy/language_service.ts
```diff
@@ -7,30 +7,53 @@
...
-    this.compiler = new Compiler(project, this.options);
   }
 
   getSemanticDiagnostics(fileName: string): ts.Diagnostic[] {
-    const result = this.compiler.analyze();
-    if (!result) {
-      return [];
+    const program = this.strategy.getProgram();
+    const compiler = this.createCompiler(program);
+    if (fileName.endsWith('.ts')) {
```
ayazhafiz, line 35: semantic diagnostics aren't applicable for .d.ts files, right? Is that something we want to check here?

packages/language-service/ivy/language_service.ts
```diff
@@ -7,30 +7,53 @@
...
+      this.lastKnownProgram = compiler.getNextProgram();
+      return diagnostics;
     }
-    const {compiler, program} = result;
-    const sourceFile = program.getSourceFile(fileName);
-    if (!sourceFile) {
-      return [];
-    }
-    return compiler.getDiagnostics(sourceFile);
+    throw new Error('Ivy LS currently does not support external template');
```
ayazhafiz, line 45: Is there a TODO to an issue # we can add here?

packages/language-service/ivy/language_service.ts
```diff
@@ -66,3 +88,80 @@ export function parseNgCompilerOptions(project: ts.server.Project): CompilerOpti
   const basePath = project.getCurrentDirectory();
   return createNgCompilerOptions(basePath, config, project.getCompilationSettings());
 }
+
+function createNgCompilerAdapter(project: ts.server.Project): NgCompilerAdapter {
+  return {
+    entryPoint: null,  // entry point is only needed if code is emitted
+    constructionDiagnostics: [],
+    ignoreForEmit: new Set(),
```
ayazhafiz, line 96: Shouldn't this be all project files?

packages/language-service/ivy/language_service.ts
```diff
@@ -66,3 +88,80 @@ export function parseNgCompilerOptions(project: ts.server.Project): CompilerOpti
...
+}
+
+function createTypeCheckingProgramStrategy(project: ts.server.Project):
+    TypeCheckingProgramStrategy {
+  return {
+    supportsInlineOperations: false,
+    shimPathForComponent(component: ts.ClassDeclaration): AbsoluteFsPath {
+      return TypeCheckShimGenerator.shimFor(absoluteFromSourceFile(component.getSourceFile()));
+    },
+    getProgram(): ts.Program {
```
ayazhafiz, line 124: I'm sure this is not the case, but because I am naive -- is there any worry that the program known to the compiler and the program retrieved by the compiler for typechecking differ? For example if someone starts editing while diagnostics are being generated, the program associated with the compiler and the one used for checking would be different. Could this lead to discrepancies in the diagnostics created for analysis of NgModule structure, for example, vs the diagnostics created from template TCBs?

ayazhafiz: Oh actually that's a totally moot question because the semantic diagnostics request should be cancelled and a fresh would be created if the user makes edit. Nevermind.

ayazhafiz:
LGTM, +me after comments are resolved

packages/language-service/ivy/language_service.ts
kyliau: `d.ts` files are not really a concern at this point, because the suffix check here is mainly to differentiate between typescript files and Angular templates. If we really want to make sure we do not process `d.ts` files the right place to perform the check should be in the extension, where this function gets called.
Currently we do the same check in the View Engine language service:
https://github.com/angular/angular/blob/71079ce47e974df4139fb7f93cca5cc4da237a9d/packages/language-service/src/typescript_host.ts#L465-L478

packages/language-service/ivy/language_service.ts
kyliau: Added issue number.

packages/language-service/ivy/language_service.ts
kyliau: It's here mainly to satisfy the input interface. Since language service never calls `emit()`, this set is not actually used.
https://github.com/angular/angular/blob/71079ce47e974df4139fb7f93cca5cc4da237a9d/packages/compiler-cli/src/ngtsc/program.ts#L203-L211

kyliau (commented):
@ayazhafiz could you please take another look? Sorry I was OOO last week.

ayazhafiz (approved):

gkalpak (approved):
There is a typo in the commit message: `[Ivy[` --> `[Ivy]`
Also, can you please add some info in the commit message (to give some context about the change).

Otherwise lgtm :+1:

Reviewed-for: dev-infra

PR #38319, opened 2020-07-31
fix(compiler): update unparsable character reference entity error messages

### LINKED ISSUE #26067: ml_parser - leading '&#' leads to a cryptic error
## I'm submitting a...

<pre><code>
[X] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Other... Please describe:
</code></pre>

## Current behavior
Any attribute value with an leading `&#` will throw:
```
Template parse errors:
Unexpected character "'" ("<hello name="{{ name }}" [icon]="'&#[ERROR ->]'"></hello>
<p>
Start editing to see some magic happen :)
"): ng:///AppModule/AppComponent.html@0:36
Unexpected closing tag "hello". It may happen when the tag has already been closed by another tag. For more info see https://www.w3.org/TR/html5/syntax.html#closing-elements-that-have-implied-end-tags ("<hello name="{{ name }}" [icon]="'&#'">[ERROR ->]</hello>
<p>
Start editing to see some magic happen :)
"): ng:///AppModule/AppComponent.html@0:39
Evaluating src/main.ts
Booting application
```

## Expected behavior

1. The parser should point to the exact issue that an entity was expected. The error message [here](https://github.com/angular/angular/blob/master/packages/compiler/src/ml_parser/lexer.ts#L330) should be more meaningful. **or**
2. The parser should proceed and handle it as text because parser can't know if it's an entity. **or**
3. **or** Handle e.g `&#333` as `ō` that's the behaviour of [parse5](https://astexplorer.net/#/gist/f538a3a7119c6c1aff26693455d52e7a/5fe7c65b6d20130bc8ef79002807f079cedef702)

But we can definitly improve the error message.

https://stackblitz.com/edit/angular-fjgsqz?file=src/app/app.component.html

User should get the correct error message

## Environment
Angular version: 6.0.0

josephperrott:
Within an angular template, when a character entity is unable to be parsed, previously a generic
unexpected character error was thrown.  This does not properly express the issue that was discovered
as the issue is actually caused by the discovered character making the whole of the entity unparsable.
The compiler will now instead inform via the error message what string was attempted to be parsed
and what it was attempted to be parsed as.

Example, for this template:
```html
<p>
  &#x123p
</p>
```
Before this change:
`Unexpected character "p"`

After this change:
`Unable to parse entity "&#x123p" - hexadecimal character reference entities must end with ";"`

Fixes #26067

### DIFF
--- a/packages/compiler/src/ml_parser/lexer.ts
+++ b/packages/compiler/src/ml_parser/lexer.ts
@@ -139,7 +139,8 @@ function _unknownEntityErrorMsg(entitySrc: string): string {
 }
 
 function _unparsableEntityErrorMsg(type: CharacterReferenceType, entityStr: string): string {
-  return `Unable to parse entity "${entityStr}" - ${type} character reference entities must end with ";"`;
+  return `Unable to parse entity "${entityStr}" - ${
+      type} character reference entities must end with ";"`;
 }
 
 enum CharacterReferenceType {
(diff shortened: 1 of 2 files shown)


packages/compiler/src/ml_parser/lexer.ts
```diff
@@ -400,8 +409,10 @@ class _Tokenizer {
       const codeStart = this._cursor.clone();
       this._attemptCharCodeUntilFn(isDigitEntityEnd);
       if (this._cursor.peek() != chars.$SEMICOLON) {
+        this._cursor.advance();
```
AndrewKushnir, line 412: May be add a quick comment here on why we need to advance the cursor?

AndrewKushnir (approved):
Thanks for the fix @josephperrott! LGTM (with one nit) 👍

AndrewKushnir:
FYI, adding a "cleanup" label, since lint is failing.

PR #38320, opened 2020-08-01
Mark eager `NgModuleFactory` construction as not side effectful

dgp1130:
Roll forward of #38147.

This allows Closure compiler to tree shake unused constructor calls to `NgModuleFactory`, which is otherwise considered
side-effectful. The Angular compiler generates factory objects which are exported but typically not used, as they are
only needed for compatibility with View Engine. This results in top-level constructor calls, such as:

```typescript
export const FooNgFactory = new NgModuleFactory(Foo);
```

`NgModuleFactory` has a side-effecting constructor, so this statement cannot be tree shaken, even if `FooNgFactory` is
never imported. The `NgModuleFactory` continues to reference its associated `NgModule` and prevents the module and all
its unused dependencies from being tree shaken, making Closure builds significantly larger than necessary.

The fix here is to wrap `NgModuleFactory` constructor with `noSideEffects(() => /* ... */)`, which tricks the Closure
compiler into assuming that the invoked function has no side effects. This allows it to tree-shake unused
`NgModuleFactory()` constructors when they aren't imported. Since the factory can be removed, the module can also be
removed (if nothing else references it), thus tree shaking unused dependencies as expected.

The one notable edge case is for lazy loaded modules. Internally, lazy loading is done as a side effect when the lazy
script is evaluated. For Angular, this side effect is registering the `NgModule`. In Ivy this is done by the
`NgModuleFactory` constructor, so lazy loaded modules **cannot** have their top-level `NgModuleFactory` constructor
call tree shaken. We handle this case by looking for the `id` field on `@NgModule` annotations. All lazy loaded modules
include an `id`. When this `id` is found, the `NgModuleFactory` is generated **without** with `noSideEffects()` call,
so Closure will not tree shake it and the module will lazy-load correctly.

## PR Checklist

## PR Type

- [X] Bugfix

Closure does not tree shake unused `NgModule`s from the bundle.

Issue Number: http://b/161548609

## What is the new behavior?
Closure will now tree shake unused `NgModule`s.

- [X] No

### DIFF
--- a/packages/compiler-cli/src/ngtsc/annotations/src/ng_module.ts
+++ b/packages/compiler-cli/src/ngtsc/annotations/src/ng_module.ts
@@ -307,7 +307,10 @@ export class NgModuleDecoratorHandler implements
     });
 
     if (this.factoryTracker !== null) {
-      this.factoryTracker.track(node.getSourceFile(), analysis.factorySymbolName);
+      this.factoryTracker.track(node.getSourceFile(), {
+        name: analysis.factorySymbolName,
+        hasId: !!analysis.id,
+      });
     }
 
     this.injectableRegistry.registerInjectable(node);
--- a/packages/compiler-cli/src/ngtsc/annotations/src/ng_module.ts
+++ b/packages/compiler-cli/src/ngtsc/annotations/src/ng_module.ts
@@ -309,7 +309,7 @@ export class NgModuleDecoratorHandler implements
     if (this.factoryTracker !== null) {
       this.factoryTracker.track(node.getSourceFile(), {
         name: analysis.factorySymbolName,
-        hasId: !!analysis.id,
+        hasId: analysis.id !== null,
       });
     }
 
--- a/packages/compiler-cli/src/ngtsc/shims/src/factory_generator.ts
+++ b/packages/compiler-cli/src/ngtsc/shims/src/factory_generator.ts
@@ -186,8 +186,24 @@ function transformFactorySourceFile(
 
         // Otherwise, check if this export is a factory for a known NgModule, and retain it if so.
         const match = STRIP_NG_FACTORY.exec(decl.name.text);
-        if (match !== null && moduleSymbols.has(match[1])) {
-          transformedStatements.push(stmt);
+        const module = match ? moduleSymbols.get(match[1]) : null;
+        if (module) {
+          // If the module can be tree shaken, then the factory should be wrapped in a
+          // `noSideEffects()` call which tells Closure to treat the expression as pure, allowing
+          // it to be removed if the result is not used.
+          //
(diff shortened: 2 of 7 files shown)


dgp1130:
http://test/OCL:324330100:BASE:324331503:1596244876644:437c49d7

Also running an global presubmit over the weekend since the last attempt caused some breakages.

mhevery (approved):
reviewed-for: fw-core

dgp1130:
Global presubmit: http://test/OCL:324330100:BASE:324338380:1596247508772:b8ff4372

Failures appear to be mostly flakes, I don't see anything that appears related to this PR (of course that's what I said last time, so feel free to double check).

packages/compiler-cli/src/ngtsc/shims/src/factory_generator.ts
```diff
@@ -263,3 +282,62 @@ function getFileoverviewComment(sourceFile: ts.SourceFile): string|null {
...
+      'ɵnoSideEffects',
+  );
+
+  return ts.createCall(
+      noSideEffects,
+      [],  // typeArguments
+      [
+        // arguments
+        ts.createFunctionExpression(
+            [],         // modifiers.
```
alxhub, line 306: Hmm. The pattern in the compiler has been to prefix arguments in `ts.create*` methods instead of postfix:

```typescript
ts.createFunctionExpression(
/* modifiers */ [],
/* asteriskToken */ undefined,
...
)
```

That said, this style is undeniably nicely formatted - at least for the single line arguments.

alxhub (approved):

alxhub:
Also, an additional note re: the commit messages: the scope for this part of the code is `compiler-cli`, not `compiler` (weird< I know).

packages/compiler-cli/src/ngtsc/shims/src/factory_generator.ts
dgp1130: Done.

packages/compiler-cli/src/ngtsc/annotations/src/ng_module.ts
```diff
@@ -307,7 +307,10 @@ export class NgModuleDecoratorHandler implements
     });
 
     if (this.factoryTracker !== null) {
-      this.factoryTracker.track(node.getSourceFile(), analysis.factorySymbolName);
+      this.factoryTracker.track(node.getSourceFile(), {
+        name: analysis.factorySymbolName,
+        hasId: !!analysis.id,
```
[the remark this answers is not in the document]
dgp1130: Done.

dgp1130:
@alxhub, updated the commit scope to `compiler-cli`.

PR #38344, opened 2020-08-04
fix(router): prevent calling unsubscribe on undefined subscription in RouterPreloader

josephperrott:
Previously, the `ngOnDestroy` method called `unsubscribe` regardless of if `subscription` had
been initialized.  This can lead to an error attempting to call `unsubscribe` of undefined.
This change prevents this error, and instead only attempts `unsubscribe` when the subscription
has been defined.

### DIFF
--- a/packages/router/src/router_preloader.ts
+++ b/packages/router/src/router_preloader.ts
@@ -98,11 +98,10 @@ export class RouterPreloader implements OnDestroy {
     return this.processRoutes(ngModule, this.router.config);
   }
 
-  // TODO(jasonaden): This class relies on code external to the class to call setUpPreloading. If
-  // this hasn't been done, ngOnDestroy will fail as this.subscription will be undefined. This
-  // should be refactored.
   ngOnDestroy(): void {
-    this.subscription.unsubscribe();
+    if (this.subscription) {
+      this.subscription.unsubscribe();
+    }
   }
 
   private processRoutes(ngModule: NgModuleRef<any>, routes: Routes): Observable<void> {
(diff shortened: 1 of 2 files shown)


packages/router/src/router_preloader.ts
```diff
@@ -98,11 +98,10 @@ export class RouterPreloader implements OnDestroy {
     return this.processRoutes(ngModule, this.router.config);
   }
 
-  // TODO(jasonaden): This class relies on code external to the class to call setUpPreloading. If
-  // this hasn't been done, ngOnDestroy will fail as this.subscription will be undefined. This
-  // should be refactored.
   ngOnDestroy(): void {
-    this.subscription.unsubscribe();
+    if (this.subscription) {
```
atscott, line 102: Could you also change the `private subscription!: Subscription` to `private subscription?: Subscription` above?

Also, it would be nice if you were able to create a test that replicates the scenario you encountered where subscription was `undefined`. If it's not straightforward, a I think a test is otherwise unnecessary given the change from `!` to `?`  (which more accurately reflects the type) *requires* this to be updated.

atscott (approved):

packages/router/src/router_preloader.ts
josephperrott: Done, added test as requested.

packages/router/test/router_preloader.spec.ts
```diff
@@ -19,6 +19,24 @@ describe('RouterPreloader', () => {
...
+        providers: [{provide: PreloadingStrategy, useExisting: PreloadAllModules}]
+      });
+    });
+
+    it('being destroyed before expected', () => {
+      const preloader: RouterPreloader = TestBed.get(RouterPreloader);
+      // Calling the RouterPreloader's ngOnDestroy method is done to simulate what would happen if
+      // the containing NgModule is destroyed.
+      preloader.ngOnDestroy();
+      // No assertion is made as this test is confirming no errors are thrown.
```
atscott, line 36:
```suggestion
      expect(() => preloader.ngOnDestroy()).not.toThrow();
```

PR #38154, opened 2020-07-21
refactor(core): add `debug` for i18n

mhevery:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [ ] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] angular.io application / infrastructure changes
- [ ] Other... Please describe:

## What is the current behavior?

Issue Number: N/A

## Does this PR introduce a breaking change?

- [ ] Yes
- [ ] No

PR #38345, opened 2020-08-04
fix(core): Store the currently selected ICU in `LView`

mhevery:
This PR builds on:
- [x] [refactor(core): add `debug` for i18n #38154](https://github.com/angular/angular/pull/38154) 
- [ ] [refactor(core): add debug ranges to `LViewDebug` #38359](https://github.com/angular/angular/pull/38359)

---

The currently selected ICU was incorrectly being stored it `TNode`
rather than in `LView`.

Remove: `TIcuContainerNode.activeCaseIndex`
Add: `LView[TIcu.currentCaseIndex]`

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/core/src/render3/i18n.ts
+++ b/packages/core/src/render3/i18n.ts
@@ -13,13 +13,13 @@ import {getTemplateContent, SRCSET_ATTRS, URI_ATTRS, VALID_ATTRS, VALID_ELEMENTS
 import {getInertBodyHelper} from '../sanitization/inert_body';
 import {_sanitizeUrl, sanitizeSrcset} from '../sanitization/url_sanitizer';
 import {addAllToArray} from '../util/array_utils';
-import {assertDataInRange, assertDefined, assertEqual} from '../util/assert';
+import {assertDefined, assertEqual, assertGreaterThan, assertIndexInRange} from '../util/assert';
 
 import {bindingUpdated} from './bindings';
 import {attachPatchData} from './context_discovery';
 import {i18nMutateOpCodesToString, i18nUpdateOpCodesToString} from './i18n_debug';
 import {setDelayProjection} from './instructions/all';
-import {allocExpando, elementAttributeInternal, elementPropertyInternal, getOrCreateTNode, setInputsForProperty, setNgReflectProperties, textBindingInternal} from './instructions/shared';
+import {allocExpando, elementAttributeInternal, elementPropertyInternal, getOrCreateTNode, setInputsForProperty, setNgReflectProperties, textBindingInternal as applyTextBinding} from './instructions/shared';
 import {LContainer, NATIVE} from './interfaces/container';
 import {getDocument} from './interfaces/document';
 import {COMMENT_MARKER, ELEMENT_MARKER, I18nMutateOpCode, I18nMutateOpCodes, I18nUpdateOpCode, I18nUpdateOpCodes, IcuType, TI18n, TIcu} from './interfaces/i18n';
@@ -726,7 +726,7 @@ function i18nEndFirstPass(tView: TView, lView: LView) {
   const lastCreatedNode = getPreviousOrParentTNode();
 
   // Read the instructions to insert/move/remove DOM elements
-  const visitedNodes = readCreateOpCodes(rootIndex, tI18n.create, tView, lView);
+  const visitedNodes = applyCreateOpCodes(tView, rootIndex, tI18n.create, lView);
 
   // Remove deleted nodes
   let index = rootIndex + 1;
@@ -756,7 +756,7 @@ function createDynamicNodeAtIndex(
     tView: TView, lView: LView, index: number, type: TNodeType, native: RElement|RText|null,
     name: string|null): TElementNode|TIcuContainerNode {
   const previousOrParentTNode = getPreviousOrParentTNode();
-  ngDevMode && assertDataInRange(lView, index + HEADER_OFFSET);
+  ngDevMode && assertIndexInRange(lView, index + HEADER_OFFSET);
   lView[index + HEADER_OFFSET] = native;
   // FIXME(misko): Why does this create A TNode??? I would not expect this to be here.
   const tNode = getOrCreateTNode(tView, lView[T_HOST], index, type as any, name, null);
@@ -770,8 +770,16 @@ function createDynamicNodeAtIndex(
   return tNode;
 }
 
-function readCreateOpCodes(
-    index: number, createOpCodes: I18nMutateOpCodes, tView: TView, lView: LView): number[] {
+/**
+ * Apply `I18nMutateOpCodes` OpCodes.
+ *
+ * @param tView Current `TView`
+ * @param rootIndex Pointer to the root (parent) tNode for the i18n.
+ * @param createOpCodes OpCodes to process
+ * @param lView Current `LView`
+ */
+function applyCreateOpCodes(
+    tView: TView, rootindex: number, createOpCodes: I18nMutateOpCodes, lView: LView): number[] {
   const renderer = lView[RENDERER];
   let currentTNode: TNode|null = null;
   let previousTNode: TNode|null = null;
@@ -792,7 +800,7 @@ function readCreateOpCodes(
         case I18nMutateOpCode.AppendChild:
           const destinationNodeIndex = opCode >>> I18nMutateOpCode.SHIFT_PARENT;
           let destinationTNode: TNode;
-          if (destinationNodeIndex === index) {
+          if (destinationNodeIndex === rootindex) {
             // If the destination node is `i18nStart`, we don't have a
             // top-level node and we should use the host node instead
             destinationTNode = lView[T_HOST]!;
@@ -852,7 +860,6 @@ function readCreateOpCodes(
               tView, lView, commentNodeIndex, TNodeType.IcuContainer, commentRNode, null);
           visitedNodes.push(commentNodeIndex);
           attachPatchData(commentRNode, lView);
-          (currentTNode as TIcuContainerNode).activeCaseIndex = null;
           // We will add the case nodes later, during the update phase
           setIsNotParent();
           break;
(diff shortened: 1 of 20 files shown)


packages/core/src/render3/i18n.ts
```diff
@@ -932,68 +950,96 @@ function readUpdateOpCodes(
...
+ * @param icuIndex Index into `tIcus` to process.
+ * @param bindingsStartIndex Location of the first `ɵɵi18nApply`
+ * @param lView Current `LView`
+ * @param changeMask Each bit corresponds to a `ɵɵi18nExp` (Counting backwards from
+ *     `bindingsStartIndex`)
+ */
+function applyIcuUpdateCase(
+    tView: TView, tIcus: TIcu[], icuIndex: number, bindingsStartIndex: number, lView: LView,
+    caseCreated: boolean) {
+  const tIcu = tIcus[icuIndex];
```
AndrewKushnir, line 967: nit: may be we can add an assert that `icuIndex` is is in the right range (`[0, tIcus.length - 1]`)?

packages/core/src/render3/i18n.ts
```diff
@@ -932,68 +950,96 @@ function readUpdateOpCodes(
...
+ *
+ * @param tView Current `TView`
+ * @param tIcus ICUs active at this location.
+ * @param icuIndex Index into `tIcus` to process.
+ * @param lView Current `LView`
+ * @returns true if a new case was created (needed so that the update executes regardless of the
+ *     bitmask)
+ */
+function applyIcuSwitchCaseRemove(tView: TView, tIcus: TIcu[], icuIndex: number, lView: LView) {
+  const tIcu = tIcus![icuIndex];
```
AndrewKushnir, line 1022:
```suggestion
  const tIcu = tIcus[icuIndex];
```
It should not be needed based on the `tIcus` type.

packages/core/src/render3/i18n.ts
```diff
@@ -932,68 +950,96 @@ function readUpdateOpCodes(
...
+function applyIcuSwitchCase(
+    tView: TView, tIcus: TIcu[], icuIndex: number, lView: LView, value: string): boolean {
+  applyIcuSwitchCaseRemove(tView, tIcus, icuIndex, lView);
+
+  // Rebuild a new case for this ICU
   let caseCreated = false;
-  // If there is an active case, delete the old nodes
-  if (icuTNode.activeCaseIndex !== null) {
-    const removeCodes = tIcu.remove[icuTNode.activeCaseIndex];
+  const tIcu = tIcus[icuIndex];
```
AndrewKushnir, line 996: It looks like we can calculate this at the very beginning of the function and pass `tIcu` to the `applyIcuSwitchCaseRemove` function call (instead of passing `tIcus` and `icuIndex` and calculating the current ICU there).

packages/core/src/render3/i18n.ts
```diff
@@ -1239,6 +1286,9 @@ function icuStart(
   const tIcu: TIcu = {
     type: icuExpression.type,
     vars,
+    currentCaseIndex: HEADER_OFFSET +
+        expandoStartIndex  // expandoStartIndex does not include the header so add it.
+        + 1,               // The first item stored is the `<!--ICU #-->` anchor so skip it.
```
AndrewKushnir, line 1291: Since this is a new slot in an `LView` that we are allocating, should we also increase the space in expando block (i.e. call `allocExpando` with the `i18nVarsCount + 1`)?

packages/core/src/render3/instructions/shared.ts
```diff
@@ -240,8 +240,7 @@ export function getOrCreateTNode(
     // For performance reasons it is important that the tNode retains the same shape during runtime.
     // (To make sure that all of the code is monomorphic.) For this reason we seal the object to
     // prevent class transitions.
-    // FIXME(misko): re-enable this once i18n code is compliant with this.
-    // Object.seal(tNode);
+    Object.seal(tNode);
```
AndrewKushnir, line 243: I feel like we may need to extract this change in it's own PR to minimize risks in this PR.

packages/core/src/render3/interfaces/i18n.ts
```diff
@@ -346,6 +346,14 @@ export interface TIcu {
    */
   vars: number[];
 
+  /**
+   * Currently selected ICU case pointer.
+   *
+   * `lView[currentCaseIndex]` stores the currently selected case. This is needed to know how to
+   * clean up the current case when transitioning no the new case.
+   */
+  currentCaseIndex: number;
```
AndrewKushnir, line 355: I was initially confused that we still keep the current case index in T-structure (as the name of this field suggests), but this is a pointer to an LView, so everything is great. I was just thinking if we can give this field slightly different name to avoid this confusion in the future. However finding a good name for it is tricky, since it's "an index to an index" :) May be something like `currentCaseLViewIndex`?

AndrewKushnir (commented):
Thanks for the fix @mhevery 👍  I've just left a few comments.

packages/core/src/render3/instructions/shared.ts
mhevery: Move the `seal` into `createTNodeAtIndex`

mhevery: Maybe add `seal` to other data structures as well.

packages/core/src/render3/i18n.ts
mhevery: added.

packages/core/src/render3/i18n.ts
mhevery: Yes, we could, and I originally had it that way, but it does not work for all of the apply methods, so I think for consistency I would prefer to pass in `tIcus: TIcu[], icuIndex: number` to all of these methods. Is that OK with you?

packages/core/src/render3/i18n.ts
mhevery: fixed

packages/core/src/render3/i18n.ts
mhevery: Yes, we need to do that. That is done later. See: 
```
vars: 1,  // allocate space for `TIcu.currentCaseIndex`
```

packages/core/src/render3/instructions/shared.ts
mhevery: fixed

packages/core/src/render3/interfaces/i18n.ts
mhevery: done.

mhevery:
[presubmit](http://test/OCL:325328186:BASE:325328202:1596754395777:fd5d67c9)

packages/core/src/render3/i18n.ts
```diff
@@ -932,68 +950,98 @@ function readUpdateOpCodes(
...
+ * @param tView Current `TView`
+ * @param tIcus If ICUs present than this contain them.
+ * @param icuIndex Index into `tIcus` to process.
+ * @param bindingsStartIndex Location of the first `ɵɵi18nApply`
+ * @param lView Current `LView`
+ * @param changeMask Each bit corresponds to a `ɵɵi18nExp` (Counting backwards from
+ *     `bindingsStartIndex`)
+ */
+function applyIcuUpdateCase(
+    tView: TView, tIcus: TIcu[], icuIndex: number, bindingsStartIndex: number, lView: LView,
```
AndrewKushnir, line 965: nit: I'd propose to rename `icuIndex` -> `tIcuIndex` to establish a relation with the `tIcus` argument.
```suggestion
    tView: TView, tIcus: TIcu[], tIcuIndex: number, bindingsStartIndex: number, lView: LView,
```

packages/core/src/render3/i18n.ts
```diff
@@ -932,68 +950,98 @@ function readUpdateOpCodes(
...
-  const icuTNode = getTNode(tView, nodeIndex) as TIcuContainerNode;
-  if (icuTNode.activeCaseIndex !== null) {
-    readUpdateOpCodes(
-        tIcu.update[icuTNode.activeCaseIndex], tIcus, bindingsStartIndex, changeMask, tView, lView,
-        caseCreated);
+/**
+ * Apply OpCodes associated with updating an existing ICU.
+ *
+ * @param tView Current `TView`
+ * @param tIcus If ICUs present than this contain them.
```
AndrewKushnir, line 957: I believe in case of `applyIcuUpdateCase`, ICUs are always present.
```suggestion
 * @param tIcus ICUs active at this location.
```

packages/core/src/render3/i18n.ts
```diff
@@ -932,68 +950,98 @@ function readUpdateOpCodes(
...
+ * @param tView Current `TView`
+ * @param tIcus ICUs active at this location.
+ * @param icuIndex Index into `tIcus` to process.
+ * @param lView Current `LView`
+ * @param value Value of the case to update to.
+ * @returns true if a new case was created (needed so that the update executes regardless of the
+ *     bitmask)
+ */
+function applyIcuSwitchCase(
+    tView: TView, tIcus: TIcu[], icuIndex: number, lView: LView, value: string): boolean {
```
AndrewKushnir, line 993:
```suggestion
    tView: TView, tIcus: TIcu[], tIcuIndex: number, lView: LView, value: string): boolean {
```

packages/core/src/render3/i18n.ts
AndrewKushnir: Sounds good 👍

packages/core/src/render3/i18n.ts
```diff
@@ -932,68 +950,98 @@ function readUpdateOpCodes(
...
+  applyIcuSwitchCaseRemove(tView, tIcus, icuIndex, lView);
+
+  // Rebuild a new case for this ICU
   let caseCreated = false;
-  // If there is an active case, delete the old nodes
-  if (icuTNode.activeCaseIndex !== null) {
-    const removeCodes = tIcu.remove[icuTNode.activeCaseIndex];
+  const tIcu = tIcus[icuIndex];
```
AndrewKushnir, line 998: May be add the checks that the index is in the range here as well?
```suggestion
  ngDevMode && assertIndexInRange(tIcus, icuIndex);
  const tIcu = tIcus[icuIndex];
```

packages/core/src/render3/i18n.ts
```diff
@@ -932,68 +950,98 @@ function readUpdateOpCodes(
...
+ *
+ * @param tView Current `TView`
+ * @param tIcus ICUs active at this location.
+ * @param icuIndex Index into `tIcus` to process.
+ * @param lView Current `LView`
+ * @returns true if a new case was created (needed so that the update executes regardless of the
+ *     bitmask)
+ */
+function applyIcuSwitchCaseRemove(tView: TView, tIcus: TIcu[], icuIndex: number, lView: LView) {
+  const tIcu = tIcus[icuIndex];
```
AndrewKushnir, line 1024: May be add the checks that the index is in the range here as well?
```suggestion
  ngDevMode && assertIndexInRange(tIcus, icuIndex);
  const tIcu = tIcus[icuIndex];
```

AndrewKushnir (approved):
Looks great (just left a few minor comments) 👍

packages/core/src/render3/i18n.ts
mhevery: fixed

packages/core/src/render3/i18n.ts
mhevery: done

packages/core/src/render3/i18n.ts
mhevery: done

packages/core/src/render3/i18n.ts
mhevery: done

packages/core/src/render3/i18n.ts
mhevery: done

atscott (approved):
reviewed-for: size-tracking

AndrewKushnir (approved):
Reviewed-for: size-tracking

PR #38349, opened 2020-08-05
fix(router): ensure routerLinkActive updates when associated routerLinks change

### LINKED ISSUE #18469: routerLinkActive not updating when routerLink changed
I have a routerLink URL that can change over time, but when the URL changes, routerLinkActive is not re-evaluated.

## I'm submitting a...

<pre><code>
[ x ] Bug report  
[ ] Feature request
</code></pre>

## Current behavior
Change a routerLink URL to match/not match the current URL. routerLinkActive class is not applied/removed.

## Expected behavior
routerLinkActive is re-evaluated when its corresponding routerLink is updated.

Click "link a" or "link b" to make them active, then click the switchLinks button to change the URLs. Nothing changes. 
http://plnkr.co/edit/DHn4my9uAKJSIQSTIm1g?p=preview

I have multiple departments managed by the same UI, with department name as a variable in the URL. When the user switches departments, and the on screen links or URL is changed, the UI doesn't update the routerLinkActive.

I think it's because ContentChildren.changes doesn't fire if the contents of a directive is changed, only when directives are added or removed.:
https://github.com/angular/angular/blob/master/packages/router/src/directives/router_link_active.ts#L106

Maybe it's too bad for performance to fix this?

## Environment

Angular version: 4.3
Mac

Browser:
- [x ] Chrome (desktop) version 59
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX

atscott:
This commit introduces a new subscription in the `routerLinkActive` directive which triggers an update
when any of its associated routerLinks have changes. `RouterLinkActive` not only needs to know when
links are added or removed, but it also needs to know about if a link it already knows about
changes in some way.

Quick note that `from...mergeAll` is used instead of just a simple
`merge` (or `scheduled...mergeAll`) to avoid introducing new rxjs
operators in order to keep bundle size down.

Fixes #18469

### DIFF
--- a/packages/router/src/directives/router_link.ts
+++ b/packages/router/src/directives/router_link.ts
@@ -7,8 +7,8 @@
  */
 
 import {LocationStrategy} from '@angular/common';
-import {Attribute, Directive, ElementRef, HostBinding, HostListener, Input, isDevMode, OnChanges, OnDestroy, Renderer2} from '@angular/core';
-import {Subscription} from 'rxjs';
+import {Attribute, Directive, ElementRef, HostBinding, HostListener, Input, isDevMode, OnChanges, OnDestroy, Renderer2, SimpleChanges} from '@angular/core';
+import {Subject, Subscription} from 'rxjs';
 
 import {QueryParamsHandling} from '../config';
 import {Event, NavigationEnd} from '../events';
@@ -115,7 +115,7 @@ import {UrlTree} from '../url_tree';
  * @publicApi
  */
 @Directive({selector: ':not(a):not(area)[routerLink]'})
-export class RouterLink {
+export class RouterLink implements OnChanges {
   /**
    * Passed to {@link Router#createUrlTree Router#createUrlTree} as part of the `NavigationExtras`.
    * @see {@link NavigationExtras#queryParams NavigationExtras#queryParams}
@@ -167,12 +167,19 @@ export class RouterLink {
   private commands: any[] = [];
   private preserve!: boolean;
 
+  /** @internal */
+  onChanges = new Subject<void>();
+
   constructor(
       private router: Router, private route: ActivatedRoute,
       @Attribute('tabindex') tabIndex: string, renderer: Renderer2, el: ElementRef) {
     if (tabIndex == null) {
       renderer.setAttribute(el.nativeElement, 'tabindex', '0');
     }
+  }
+
+  ngOnChanges(changes: SimpleChanges) {
+    this.onChanges.next();
   }
 
   /**
@@ -298,6 +305,9 @@ export class RouterLinkWithHref implements OnChanges, OnDestroy {
   // TODO(issue/24571): remove '!'.
   @HostBinding() href!: string;
 
+  /** @internal */
+  onChanges = new Subject<void>();
+
   constructor(
       private router: Router, private route: ActivatedRoute,
       private locationStrategy: LocationStrategy) {
@@ -336,8 +346,9 @@ export class RouterLinkWithHref implements OnChanges, OnDestroy {
   }
 
   /** @nodoc */
-  ngOnChanges(changes: {}): any {
+  ngOnChanges(changes: SimpleChanges): any {
     this.updateTargetUrlAndHref();
+    this.onChanges.next();
   }
   /** @nodoc */
   ngOnDestroy(): any {
--- a/packages/router/src/directives/router_link.ts
+++ b/packages/router/src/directives/router_link.ts
@@ -178,7 +178,10 @@ export class RouterLink implements OnChanges {
     }
   }
 
+  /** @nodoc */
   ngOnChanges(changes: SimpleChanges) {
+    // This is subscribed to by `RouterLinkActive` so that it knows to update when there are changes
+    // to the RouterLinks it's tracking.
     this.onChanges.next();
   }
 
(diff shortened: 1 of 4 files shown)


atscott:
[presubmit](https://test.corp.google.com/ui#id=OCL:325049401:BASE:325064353:1596652835642:658bbfd3)

packages/router/src/directives/router_link.ts
```diff
@@ -175,6 +178,10 @@ export class RouterLink {
     }
   }
 
+  ngOnChanges(changes: SimpleChanges) {
+    this.onChanges.next();
```
AndrewKushnir, line 182: I'd propose to add a comment here to mention why we emit here even though there is no additional logic in this lifecycle hook. I believe the reason is that the logic in corresponding getters is invoked, thus `ngOnChanges` was empty.

packages/router/src/directives/router_link_active.ts
```diff
@@ -104,9 +104,22 @@ export class RouterLinkActive implements OnChanges, OnDestroy, AfterContentInit
 
 
   ngAfterContentInit(): void {
-    this.links.changes.subscribe(_ => this.update());
-    this.linksWithHrefs.changes.subscribe(_ => this.update());
-    this.update();
+    from([this.links.changes, this.linksWithHrefs.changes, of(null)])
```
AndrewKushnir, line 107: Nice refactoring 👍 May be add a quick comment on why we have `of(null)` here?

packages/router/src/directives/router_link.ts
```diff
@@ -175,6 +178,10 @@ export class RouterLink {
     }
   }
 
+  ngOnChanges(changes: SimpleChanges) {
```
AndrewKushnir, line 181: FYI, this method will show up on https://angular.io/api/router/RouterLink, but this is similar to what [RouterLinkWithHref](https://angular.io/api/router/RouterLinkWithHref) has. I also noticed that on other Directives in other packages (like Forms), so I think we may need to come up with a strategy on how to represent these methods (and whether they are actually add any value in docs?) consistently for all directives, which is a topic for another PR(s) :)

AndrewKushnir (approved):
LGTM (just a couple nits) 👍 

Reviewed-for: public-api

packages/router/src/directives/router_link.ts
atscott: Maybe we can/should mark them with `/** @internal */`? As long as I'm making the changes in this PR, I might as well do that here if this seems like a good solution.

AndrewKushnir: This sounds good to me, thanks Andrew 👍

mmalerba (approved):
LGTM

jelbourn (approved):
LGTM

Reviewed-for: public-api

packages/router/src/directives/router_link_active.ts
```diff
@@ -79,14 +80,13 @@ import {RouterLink, RouterLinkWithHref} from './router_link';
   exportAs: 'routerLinkActive',
 })
 export class RouterLinkActive implements OnChanges, OnDestroy, AfterContentInit {
-  // TODO(issue/24571): remove '!'.
   @ContentChildren(RouterLink, {descendants: true}) links!: QueryList<RouterLink>;
-  // TODO(issue/24571): remove '!'.
```
petebacondarwin, line 84: Generally I am against removing these comments unless the `!` is removed (somehow) or replacing the comment with a reason why the `!` can be retained.

alxhub (approved):
Reviewed-For: public-api

packages/router/src/directives/router_link_active.ts
petebacondarwin: But I guess Angular developers are aware that queries are only available after the appropriate hook has been called.

packages/router/src/directives/router_link.ts
```diff
@@ -175,6 +178,13 @@ export class RouterLink {
     }
   }
 
+  /** @nodoc */
+  ngOnChanges(changes: SimpleChanges) {
+    // This is subscribed to by `RouterLinkActive` so that it knows to update when there are changes
+    // to the RouterLink's it's tracking.
```
petebacondarwin, line 184:
```suggestion
    // to the RouterLinks it's tracking.
```

petebacondarwin (approved):
Reviewed-for: public-api

packages/router/src/directives/router_link_active.ts
atscott: My thought process was that they could "technically" be undefined until `afterContentInit` but are reliably initialized after that. And that we could really never make this `?` instead so the comment isn't really actionable in that way. I did some more reading in the issue report and found these guidelines that support keeping `!` as well:

> the field is reliably initialized, but TS cannot infer that (initialization can occur outside the ctor, for example), so ! should be kept.

https://github.com/angular/angular/issues/24571#issue-333470063
I think `ContentChildren` and `ViewChildren` could be considered "reliably initialized" after the correct lifecycle hooks.

> If the field is annotated with @ViewChildren or @ContentChildren - Add back '!' - c).

https://github.com/angular/angular/issues/24571#issuecomment-404606595

PR #38350, opened 2020-08-05
TypeError: Right-hand side of 'instanceof' is not an object

rifler:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/packages/zone.js/lib/common/to-string.ts
+++ b/packages/zone.js/lib/common/to-string.ts
@@ -49,9 +49,10 @@ Zone.__load_patch('toString', (global: any) => {
   const originalObjectToString = Object.prototype.toString;
   const PROMISE_OBJECT_TO_STRING = '[object Promise]';
   Object.prototype.toString = function() {
-    if (this instanceof Promise) {
+    if (typeof Promise === 'function' && this instanceof Promise) {
       return PROMISE_OBJECT_TO_STRING;
     }
+
     return originalObjectToString.call(this);
   };
 });
--- a/packages/zone.js/test/common/toString.spec.ts
+++ b/packages/zone.js/test/common/toString.spec.ts
@@ -18,6 +18,18 @@ describe('global function patch', () => {
           .toEqual(Function.prototype.toString.call(g[zoneSymbol('setTimeout')]));
     });
 
+    it('should not throw error if Promise is not a function', () => {
+      const P = g.Promise;
+      try {
+        g.Promise = undefined;
+        expect(() => {
+          const a = {}.toString();
+        }).not.toThrow();
+      } finally {
+        g.Promise = P;
+      }
+    });
+
     it('MutationObserver toString should be the same with native version',
        ifEnvSupports('MutationObserver', () => {
          const nativeMutationObserver = g[zoneSymbol('MutationObserver')];


AndrewKushnir:
Hi @rifler, thanks for creating this PR.

It looks like it's missing some critical information to make this PR actionable, specifically:

- problem description, which problem this PR resolves
- links to open issues (if any)
- tests that would verify this change (ideally, the set of tests that would fail without the change)

I will close this PR for now, feel free to reopen or create a new one once you have the necessary information and tests.

Thank you.

rifler:
@AndrewKushnir `instanceof` is not safe operation, so if you monkeypatch such substantial method as `Object.prototype.toString`, you have to use try..catch

rifler:
@AndrewKushnir I create an issue - https://github.com/angular/angular/issues/38361

but cannot reopen pull-request

AndrewKushnir:
@rifler I've re-opened this PR based on https://github.com/angular/angular/issues/38361#issuecomment-671604212. Thank you.

packages/zone.js/lib/common/to-string.ts
```diff
@@ -49,9 +49,12 @@ Zone.__load_patch('toString', (global: any) => {
...
   Object.prototype.toString = function() {
-    if (this instanceof Promise) {
-      return PROMISE_OBJECT_TO_STRING;
-    }
+    try {
+      if (this instanceof Promise) {
+        return PROMISE_OBJECT_TO_STRING;
+      }      
+    } catch (_) {}
+
```
JiaLiPassion, line 57: instead of using `try/catch`, I think maybe 
```
if (typeof Promise !== 'undefined' && this instanceof Promise) {
  return PROMISE_OBJECT_TO_STRING;
}
```
will be better, although the code still fails if some app set Promise to some invalid value such as `window.Promise = 1`, but in that case, the app already fails in a lot of other places.

JiaLiPassion (changes requested):

JiaLiPassion:
@rifler, thanks and year `typeof Promise === 'function'` is better.
Could you also add a test case to test it?

mhevery (approved):

mhevery:
LGTM after @JiaLiPassion comments are resolved.

JiaLiPassion:
@rifler , I just added one test case.
And please do the follow cleanups.

`git rebase master -i`
1. then `squash` the current 4 commits to 1 commit.
2. And then modify the commit message to following the commit guide line here https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
something like

```
fix(zone.js): zone.js toString patch should check typeof Promise is function

Close #38361

zone.js monkey patch toString, and check the instance is `Promise` or not by using `instanceof Promise`,
sometimes when Promise is not available, the `instanceof` operation fails and throw `TypeError: Right-hand side of 'instanceof' is not an object`, this PR check `typeof Promise` equals to function or not to prevent the error.
```

rifler:
fixed

JiaLiPassion:
LGTM.

JiaLiPassion (approved):

AndrewKushnir:
[Presubmit](http://test/OCL:328242092:BASE:328242105:1598316758426:3159f9da)

PR #38354, opened 2020-08-05
fix(forms): ensure to emit `statusChanges` on subsequent value update…

### LINKED ISSUE #14542: FormGroup & FormControl statusChanges are not emitted on creation
**I'm submitting a ...**
```
[X] bug report
[ ] feature request
[ ] support request
```

**Current behavior**
FormGroup and FormControl don't emit statusChange after first creation (while untouched / pristine?) with asyncValidator.

1. Initial status of FormGroup and FormControl is "PENDING"
2. Both status changes to "VALID" or "INVALID"
3. statusChange is not emitted

**Expected behavior**
statusChange should always be emitted.

1. Create a FormGroup with the FormBuilder and add a FormControl with asyncValidator(e.g. with a TimeOut Promise of 2sec).
2. Subscribe to FormGroup.statusChange or FormGroup.controls.key.statusChange

**Angular version**
2.4.1

### LINKED ISSUE #20424: When replacing form control which has async validator the status from statusChange stays PENDING
## I'm submitting a...
<pre><code>[X] Bug report</code></pre>

## Current behavior
When replacing form control with a new one which has asynchronous validator only `PENDING` status is broadcasted from `statusChanges` observable.

```ts
const myForm = new FormGroup({
  myControl: new FormControl('simple input');
});
    
myForm.statusChanges
  .subscribe(status => console.log(status));

myForm.setControl('myControl', new FormControl('another input', null, asyncValidator));
```
Console will only print `PENDING`. Nothing more.

## Expected behavior
Final `VALID` or `INVALID` status should be broadcasted after async validator finishes his job.

https://stackblitz.com/edit/angular-qmdbkx

## Please tell us about your environment
<pre><code>Angular version: 7.0.1</code></pre>

sonukapoor:
… / validations

This commit ensures that the `updateValueAndValidity` method takes the
`asyncValidator` into consideration to emit on the `statusChanges` observables.
This is necessary so that any subsequent changes are emitted properly to any
subscribers.

BREAKING CHANGE:

Previously if FormControl, FormGroup and FormArray class instances had async validators
defined at initialization time, the status change event was not emitted once async validator
completed. After this change, the status event is emitted into the `statusChanges` observable.
If your code relies on the old behaviour, you can filter/ignore this additional status change
event.

Closes #20424
Closes #14542

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [x] Yes
- [ ] No

### DIFF
--- a/packages/forms/test/form_group_spec.ts
+++ b/packages/forms/test/form_group_spec.ts
@@ -1825,6 +1825,391 @@ describe('FormGroup', () => {
     });
   });
 
+  describe('emit `statusChanges` and `valueChanges` with/without async/sync validators', () => {
+    const attachEventsLogger = (control: AbstractControl, log: string[], controlName?: string) => {
+      const name = controlName ? ` (${controlName})` : '';
+      control.statusChanges.subscribe(status => log.push(`status${name}: ${status}`));
+      control.valueChanges.subscribe(value => log.push(`value${name}: ${JSON.stringify(value)}`));
+    };
+
+    describe('stand alone controls', () => {
+      it('should run the async validator on stand alone controls and set status to `INVALID`',
+         fakeAsync(() => {
+           const logs: string[] = [];
+           const c =
+               new FormControl('', null, simpleAsyncValidator({timeout: 0, shouldFail: true}));
+
+           attachEventsLogger(c, logs);
+
+           expect(logs.length).toBe(0);
+
+           tick(1);
+
+           c.setValue('new!', {emitEvent: true});
+
+           tick(1);
+
+           // Note that above `simpleAsyncValidator` is called with `timeout:0`.  When the timeout
+           // is set to `0`, the function returns `of(error)`, and the function behaves in a
+           // synchronous manner. Because of this there is no `PENDING` state as seen in the
+           // `logs`.
+           expect(logs).toEqual([
+             'status: INVALID',  // status change emitted as a result of initial async validator run
+             'value: "new!"',    // value change emitted by `setValue`
+             'status: INVALID'   // async validator run after `setValue` call
+           ]);
+         }));
+
+      it('should run the async validator on stand alone controls and set status to `VALID`',
+         fakeAsync(() => {
+           const logs: string[] = [];
+           const c = new FormControl('', null, asyncValidator('new!'));
+
+           attachEventsLogger(c, logs);
+
+           expect(logs.length).toBe(0);
+
+           tick(1);
+
+           c.setValue('new!', {emitEvent: true});
+
+           tick(1);
+
+           expect(logs).toEqual([
+             'status: INVALID',  // status change emitted as a result of initial async validator run
+             'value: "new!"',    // value change emitted by `setValue`
+             'status: PENDING',  // status change emitted by `setValue`
+             'status: VALID'     // async validator run after `setValue` call
+           ]);
+         }));
+
+      it('should run the async validator on stand alone controls, include `PENDING` and set status to `INVALID`',
+         fakeAsync(() => {
+           const logs: string[] = [];
+           const c =
+               new FormControl('', null, simpleAsyncValidator({timeout: 1, shouldFail: true}));
+
+           attachEventsLogger(c, logs);
+
+           expect(logs.length).toBe(0);
+
+           tick(1);
+
+           c.setValue('new!', {emitEvent: true});
+
+           tick(1);
+
+           expect(logs).toEqual([
+             'status: INVALID',  // status change emitted as a result of initial async validator run
+             'value: "new!"',    // value change emitted by `setValue`
+             'status: PENDING',  // status change emitted by `setValue`
+             'status: INVALID'   // async validator run after `setValue` call
+           ]);
+         }));
+
+      it('should run setValue before the initial async validator and set status to `VALID`',
+         fakeAsync(() => {
+           const logs: string[] = [];
+           const c = new FormControl('', null, asyncValidator('new!'));
+
+           attachEventsLogger(c, logs);
+
+           expect(logs.length).toBe(0);
+
+           c.setValue('new!', {emitEvent: true});
+
+           tick(1);
+
+           // The `setValue` call invoked synchronously cancels the initial run of the
+           // `asyncValidator` (which would cause the control status to be changed to `INVALID`), so
+           // the log contains only events after calling `setValue`.
+           expect(logs).toEqual([
+             'value: "new!"',    // value change emitted by `setValue`
+             'status: PENDING',  // status change emitted by `setValue`
+             'status: VALID'     // async validator run after `setValue` call
+           ]);
+         }));
+
+      it('should run setValue before the initial async validator and set status to `INVALID`',
+         fakeAsync(() => {
+           const logs: string[] = [];
+           const c =
+               new FormControl('', null, simpleAsyncValidator({timeout: 1, shouldFail: true}));
+
+           attachEventsLogger(c, logs);
+
+           expect(logs.length).toBe(0);
+
+           c.setValue('new!', {emitEvent: true});
+
+           tick(1);
+
+           // The `setValue` call invoked synchronously cancels the initial run of the
+           // `asyncValidator` (which would cause the control status to be changed to `INVALID`), so
+           // the log contains only events after calling `setValue`.
+           expect(logs).toEqual([
+             'value: "new!"',    // value change emitted by `setValue`
+             'status: PENDING',  // status change emitted by `setValue`
+             'status: INVALID'   // async validator run after `setValue` call
+           ]);
+         }));
+
+      it('should cancel initial run of the async validator and not emit anything', fakeAsync(() => {
+           const logger: string[] = [];
+           const c =
+               new FormControl('', null, simpleAsyncValidator({timeout: 1, shouldFail: true}));
+
+           attachEventsLogger(c, logger);
+
+           expect(logger.length).toBe(0);
+
+           c.setValue('new!', {emitEvent: false});
+
+           tick(1);
+
+           // Because we are calling `setValue` with `emitEvent: false`, nothing is emitted
+           // and our logger remains empty
+           expect(logger).toEqual([]);
+         }));
+
+      it('should run the sync validator on stand alone controls and set status to `INVALID`',
+         fakeAsync(() => {
+           const logs: string[] = [];
+           const c = new FormControl('new!', Validators.required);
+
+           attachEventsLogger(c, logs);
+
+           expect(logs.length).toBe(0);
+
+           tick(1);
+
+           c.setValue('', {emitEvent: true});
+
+           tick(1);
+
+           expect(logs).toEqual([
+             'value: ""',       // value change emitted by `setValue`
+             'status: INVALID'  // status change emitted by `setValue`
+           ]);
+         }));
+
+      it('should run the sync validator on stand alone controls and set status to `VALID`',
+         fakeAsync(() => {
+           const logs: string[] = [];
+           const c = new FormControl('', Validators.required);
+
+           attachEventsLogger(c, logs);
+
+           expect(logs.length).toBe(0);
+
+           tick(1);
+
+           c.setValue('new!', {emitEvent: true});
+
+           tick(1);
+
+           expect(logs).toEqual([
+             'value: "new!"',  // value change emitted by `setValue`
+             'status: VALID'   // status change emitted by `setValue`
+           ]);
+         }));
+    });
+
+    describe('combination of multiple form controls', () => {
+      it('should run the async validator on the FormControl added to the FormGroup and set status to `VALID`',
+         fakeAsync(() => {
+           const logs: string[] = [];
+           const c1 = new FormControl('one');
+           const g1 = new FormGroup({'one': c1});
+
+           // Initial state of the controls
+           expect(currentStateOf([c1, g1])).toEqual([
+             {errors: null, pending: false, status: 'VALID'},  // Control 1
+             {errors: null, pending: false, status: 'VALID'},  // Group
+           ]);
+
+           attachEventsLogger(g1, logs, 'g1');
+
+           const c2 = new FormControl('new!', null, asyncValidator('new!'));
+
+           attachEventsLogger(c2, logs, 'c2');
+
+           // Initial state of the new control
+           expect(currentStateOf([c2])).toEqual([
+             {errors: null, pending: true, status: 'PENDING'},  // Control 2
+           ]);
+
+           expect(logs.length).toBe(0);
+
+           g1.setControl('one', c2);
+
+           tick(1);
+
+           expect(logs).toEqual([
+             'value (g1): {"one":"new!"}',  // value change emitted by `setControl`
+             'status (g1): PENDING',        // value change emitted by `setControl`
+             'status (c2): VALID',          // async validator run after `setControl` call
+             'status (g1): VALID'           // status changed from the `setControl` call
+           ]);
+
+           // Final state of all controls
+           expect(currentStateOf([g1, c2])).toEqual([
+             {errors: null, pending: false, status: 'VALID'},  // Group
+             {errors: null, pending: false, status: 'VALID'},  // Control 2
+           ]);
+         }));
+
+      it('should run the async validator on the FormControl added to the FormGroup and set status to `INVALID`',
+         fakeAsync(() => {
+           const logs: string[] = [];
+           const c1 = new FormControl('one');
+           const g1 = new FormGroup({'one': c1});
+
+           // Initial state of the controls
+           expect(currentStateOf([c1, g1])).toEqual([
+             {errors: null, pending: false, status: 'VALID'},  // Control 1
+             {errors: null, pending: false, status: 'VALID'},  // Group
+           ]);
+
+           attachEventsLogger(g1, logs, 'g1');
+
+           const c2 =
+               new FormControl('new!', null, simpleAsyncValidator({timeout: 1, shouldFail: true}));
+
+           attachEventsLogger(c2, logs, 'c2');
+
+           // Initial state of the new control
+           expect(currentStateOf([c2])).toEqual([
+             {errors: null, pending: true, status: 'PENDING'},  // Control 2
+           ]);
+
+           expect(logs.length).toBe(0);
+
+           g1.setControl('one', c2);
+
+           tick(1);
+
+           expect(logs).toEqual([
+             'value (g1): {"one":"new!"}',
+             'status (g1): PENDING',  // g1 async validator is invoked after `g1.setControl` call
+             'status (c2): INVALID',  // c2 async validator trigger at c2 init, completed with the
+                                      // `INVALID` status
+             'status (g1): INVALID'   // g1 validator completed with the `INVALID` status
+           ]);
+
+           // Final state of all controls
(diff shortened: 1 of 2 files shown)


packages/forms/test/form_group_spec.ts
```diff
@@ -1825,6 +1825,225 @@ describe('FormGroup', () => {
...
+           c.valueChanges.subscribe((value) => logger.push(`value: ${value}`));
+           c.statusChanges.subscribe((status) => logger.push(`status: ${status}`));
+
+           tick(1);
+
+           c.setValue('new!', {emitEvent: true});
+
+           tick(1);
+
+           expect(logger).toEqual(['status: INVALID', 'value: new!', 'status: INVALID']);
```
AndrewKushnir, line 1844: I think it'd be great to add comments to describe some of the items in the `logger` array (in this test and in other tests as well), for example:
```
expect(logger).toEqual([
  'status: INVALID',    // status change emitted as a result of initial async validator run
  'value: new!',
  'status: INVALID'     // async validator run after `setValue` call
]);
```

Also, should we see a PENDING state here as well before the second `INVALID` state?

AndrewKushnir: We should also add a test for a case when `setValue` is called *before* initial async validator run is completed, for example:

```
  ...
  const logger: string[] = [];
  const c = new FormControl('', null, asyncValidator('new!')));
  c.valueChanges.subscribe((value) => logger.push(`value: ${value}`));
  c.statusChanges.subscribe((status) => logger.push(`status: ${status}`));

  // setValue is called synchronously before initial run of async validator is completed
  c.setValue('new!', {emitEvent: true});
  ...
```

We should check 2 cases here: one if when the initial value was valid and it was changed to invalid and vice versa.

packages/forms/test/form_group_spec.ts
```diff
@@ -1825,6 +1825,225 @@ describe('FormGroup', () => {
     });
   });
 
+  describe('emit `statusChanges` and `valueChanges` with/without async/sync validators', () => {
+    describe('in stand alone', () => {
```
AndrewKushnir, line 1829:
```suggestion
    describe('stand alone controls', () => {
```

packages/forms/test/form_group_spec.ts
```diff
@@ -1825,6 +1825,225 @@ describe('FormGroup', () => {
...
+           tick(1);
+
+           c.setValue('new!', {emitEvent: true});
+
+           tick(1);
+           expect(logger).toEqual(['value: new!', 'status: VALID']);
+         }));
+    });
+
+    describe('in combination', () => {
```
AndrewKushnir, line 1895:
```suggestion
    describe('combination of multiple form controls', () => {
```

packages/forms/test/form_group_spec.ts
```diff
@@ -1825,6 +1825,225 @@ describe('FormGroup', () => {
...
+           const c2 = new FormControl('new!', null, asyncValidator('new!'));
+           // Initial state of the new control
+           expect(currentStateOf([c2])).toEqual([
+             {errors: null, pending: true, status: 'PENDING'},  // Control 2
+           ]);
+
+           g1.setControl('one', c2);
+
+           tick(1);
+           expect(logger).toEqual(['value: {"one":"new!"}', 'status: PENDING', 'status: VALID']);
```
AndrewKushnir, line 1919: I think we should record status changes for the `c2` control too, so we'll see the full log of status events. In this case may be also include control name into the log like `status (g1): PENDING` or `status (c2): VALID`.

This is applicable to other tests too, so that we can see full history of state changes.

packages/forms/test/form_group_spec.ts
```diff
@@ -1825,6 +1825,225 @@ describe('FormGroup', () => {
...
+
+           c.setValue('new!', {emitEvent: true});
+
+           tick(1);
+           expect(logger).toEqual(['value: new!', 'status: VALID']);
+         }));
+    });
+
+    describe('in combination', () => {
+      it('should run the async validator and set status to `VALID`', fakeAsync(() => {
```
AndrewKushnir, line 1896: Let's also add a bit more info into the test description (for this test and other tests as well):
```suggestion
      it('should run the async validator on FormGroup and FormControl and set status to `VALID`', fakeAsync(() => {
```

AndrewKushnir (commented):
Thanks for adding tests @sonukapoor 👍 

I've left a few comments (most of them are applicable to many tests), please have a look when you get a chance. Once the feedback is addressed, I will have another look.

Thank you.

packages/forms/test/form_group_spec.ts
sonukapoor: > I think it'd be great to add comments to describe some of the items in the `logger` array (in this test and in other tests as well), for example:

That's a good idea. I added them. 

> Also, should we see a PENDING state here as well before the second `INVALID` state?

No, because the `error` property contains a value. When this happens, the `INVALID` status takes precedence over the `PENDING` state. You can see that [here](https://github.com/angular/angular/blob/445ac15a785121cd7de099a33a3fded0c8b9ab83/packages/forms/src/model.ts#L939).

packages/forms/test/form_group_spec.ts
sonukapoor: Makes sense. I have added them

sonukapoor:
@AndrewKushnir All changes are done. The PR is ready for the second review.

AndrewKushnir:
[Initial Presubmit](http://test/OCL:325933856:BASE:325933864:1597110699853:e5c0d7a3).

packages/forms/test/form_group_spec.ts
```diff
@@ -1825,6 +1825,340 @@ describe('FormGroup', () => {
...
+           tick(1);
+
+           c.setValue('new!', {emitEvent: true});
+
+           tick(1);
+
+           expect(logger).toEqual([
+             'status: INVALID',  // status change emitted as a result of initial async validator run
+             'value: new!',      // value change emitted by `setValue`
+             'status: INVALID'   // async validator run after `setValue` call
```
AndrewKushnir, line 1849: I was a bit surprised that there is no `PENDING` state present here and did some debugging. The thing is that the `simpleAsyncValidator` has timeout defined as `0`, so it returns `of(error)` and the code actually acts in a synchronous way (even though the validator is async). So I think we should add a comment above the `expect(logger).toEqual` line to describe that and also create a copy of this test, but with the `timeout` greater that 0, so the `PENDING` status shows up in the log.

packages/forms/test/form_group_spec.ts
```diff
@@ -1825,6 +1825,340 @@ describe('FormGroup', () => {
...
+           c.valueChanges.subscribe((value) => logger.push(`value: ${value}`));
+           c.statusChanges.subscribe((status) => logger.push(`status: ${status}`));
+
+           expect(logger.length).toBe(0);
+
+           c.setValue('new!', {emitEvent: true});
+
+           tick(1);
+
+           expect(logger).toEqual([
```
AndrewKushnir, line 1889: I'd propose to add a quick comment here:
```suggestion
           // The `setValue` call invoked synchronously cancels the initial run of the `asyncValidator` (which 
           // would cause the control status to be changed to `INVALID`), so the log contains only events after
           // calling `setValue`. 
           expect(logger).toEqual([
```

packages/forms/test/form_group_spec.ts
```diff
@@ -1825,6 +1825,340 @@ describe('FormGroup', () => {
...
```
AndrewKushnir, line 1910: Please add a similar comment as mentioned above, since it's the same use-case, just different final status.

packages/forms/test/form_group_spec.ts
```diff
@@ -1825,6 +1825,340 @@ describe('FormGroup', () => {
...
+             'value: new!',      // value change emitted by `setValue`
+             'status: PENDING',  // status change emitted by `setValue`
```
AndrewKushnir, line 1912: I think it'd be also great to add similar test where `setValue` is called with `{emitEvent: false}` to make sure we cancel initial run of the async validator.

packages/forms/test/form_group_spec.ts
```diff
@@ -1825,6 +1825,340 @@ describe('FormGroup', () => {
...
+           ]);
+
+           g1.statusChanges.subscribe((status) => logger.push(`status (g1): ${status}`));
+           g1.valueChanges.subscribe(
+               (value) => logger.push(`value (g1): ${JSON.stringify(value)}`));
+
+           const c2 = new FormControl('new!', null, asyncValidator('new!'));
+           c2.statusChanges.subscribe((status) => logger.push(`status (c2): ${status}`));
+           c2.valueChanges.subscribe(
+               (value) => logger.push(`value (c2): ${JSON.stringify(value)}`));
```
AndrewKushnir, line 1979: We can probably refactor tests a bit and introduce a function that attaches a logger (to avoid repeating code), something like this:

```
function attachEventsLogger(control: AbstractControl, log: string:[], controlName?: string) {
  const name = controlName ? `(${name})` : '';
  control.statusChanges.subscribe(status => log.push(`status ${name}: ${status}`));
  control.valueChanges.subscribe(value => log.push(`value ${name}: ${JSON.stringify(value)}`));
}

// ... in tests code ...
attachEventsLogger(c2, log, 'c2');
```

AndrewKushnir (commented):
Hi @sonukapoor, thanks for additional changes. I've left a few more comments, please have a look when you get a chance. Thank you.

sonukapoor:
@AndrewKushnir It's ready for another review. Thanks

packages/forms/test/form_group_spec.ts
```diff
@@ -1825,6 +1825,386 @@ describe('FormGroup', () => {
...
+           tick(1);
+
+           c.setValue('new!', {emitEvent: true});
+
+           tick(1);
+
+           // Note that above `simpleAsyncValidator` is called with `timeout:0`.  When the timeout
+           // is set to `0`, the function returns `of(error)`, and the function behaves in a
+           // synchronous manner. Because of this there is no `PENDING` state as seen in the
+           // `logger`.
```
AndrewKushnir, line 1855:
```suggestion
           // `logs` array.
```

packages/forms/test/form_group_spec.ts
```diff
@@ -1825,6 +1825,386 @@ describe('FormGroup', () => {
...
+           // `asyncValidator` (which would cause the control status to be changed to `INVALID`), so
+           // the log contains only events after calling `setValue`.
+           expect(logs).toEqual([
+             'value: "new!"',    // value change emitted by `setValue`
+             'status: PENDING',  // status change emitted by `setValue`
+             'status: INVALID'   // async validator run after `setValue` call
+           ]);
+         }));
+
+      it('should cancel initial run of the async validator and not emit anything', fakeAsync(() => {
```
AndrewKushnir, line 1957: That's a great test to have 👍

packages/forms/test/form_group_spec.ts
```diff
@@ -1825,6 +1825,386 @@ describe('FormGroup', () => {
...
+
+           expect(logs).toEqual([
+             'value: "new!"',  // value change emitted by `setValue`
+             'status: VALID'   // status change emitted by `setValue`
+           ]);
+         }));
+    });
+
+    describe('combination of multiple form controls', () => {
+      it('should run the async validator on FormGroup and FormControl and set status to `VALID`',
```
AndrewKushnir, line 2019: I don't think we run validators for the `FormGroup` in that test (please correct me if I'm wrong).
```suggestion
      it('should run the async validator on the FormControl  added to the FormGroup and set status to `VALID`',
```

packages/forms/test/form_group_spec.ts
```diff
@@ -1825,6 +1825,386 @@ describe('FormGroup', () => {
...
+
+           g1.setControl('one', c2);
+
+           tick(1);
+
+           expect(logs).toEqual([
+             'value (g1): {"one":"new!"}',  // value change emitted by `setControl`
+             'status (g1): PENDING',        // value change emitted by `setControl`
+             'status (c2): VALID',          // async validator run after `setControl` call
+             'status (g1): VALID'           // async validator run after `setControl` call
```
AndrewKushnir, line 2052: I think this comment is not correct, since the `FormGroup` has no validators.

packages/forms/test/form_group_spec.ts
```diff
@@ -1825,6 +1825,386 @@ describe('FormGroup', () => {
...
+           ]);
+
+           // Final state of all controls
+           expect(currentStateOf([c1, g1, c2])).toEqual([
```
AndrewKushnir, line 2056: Since we replace `c1` with `c2` in the test, we should not check its status here (it should not be relevant).

packages/forms/test/form_group_spec.ts
```diff
@@ -1825,6 +1825,386 @@ describe('FormGroup', () => {
...
+             {errors: null, pending: false, status: 'VALID'},  // Control 1
+             {errors: null, pending: false, status: 'VALID'},  // Group
+             {errors: null, pending: false, status: 'VALID'},  // Control 2
+           ]);
+         }));
+
+      it('should run the async validator and set status to `INVALID`', fakeAsync(() => {
```
AndrewKushnir, line 2063: The comments above are applicable to this test as well. We should also update test description to describe what we are testing (the description is too generic atm).

packages/forms/test/form_group_spec.ts
```diff
@@ -1825,6 +1825,386 @@ describe('FormGroup', () => {
...
+           g1.setControl('one', c2);
+
+           tick(1);
+
+           expect(logs).toEqual([
+             'value (g1): {"one":"new!"}',  // value change emitted by `setControl`
+             'status (g1): PENDING',        // value change emitted by `setControl`
+             'status (c2): INVALID',        // async validator run after `setControl` call
+             'status (g1): PENDING',        // async validator run after `setControl` call
+             'status (g1): INVALID'         // async validator run after `setControl` call
```
AndrewKushnir, line 2143: I'd propose updating comments a bit:
```suggestion
             'value (g1): {"one":"new!"}',
             'status (g1): PENDING',    // g1 async validator is invoked after `g1.setControl` call
             'status (c2): INVALID',      // c2 async validator trigger at c2 init, completed with the `INVALID` status
             'status (g1): PENDING',    // c2 update triggered g1 to re-run validation
             'status (g1): INVALID'       // g1 validator completed with the `INVALID` status
```

packages/forms/test/form_group_spec.ts
```diff
@@ -1825,6 +1825,386 @@ describe('FormGroup', () => {
...
+
+           expect(logs.length).toBe(0);
+
+           g1.setControl('one', c2);
+
+           tick(1);
+
+           expect(logs).toEqual([
+             'value (fa): [{"one":"new!"}]',  // value change emitted by `setControl`
+             'status (fa): PENDING',          // value change emitted by `setControl`
```
AndrewKushnir, line 2190: I think we need to update comments here as well (to describe why we have this sequence of events), see my comment above.

AndrewKushnir (commented):
Thanks for updating tests @sonukapoor, great work 👍 

I've a few more comments, they are mostly around comments/test descriptions, so I think we should move this PR to the "Ready for Review" state and perform a final round of review.

Thank you.

packages/forms/test/form_group_spec.ts
sonukapoor: 💯

packages/forms/test/form_group_spec.ts
sonukapoor: Good catch.

packages/forms/test/form_group_spec.ts
sonukapoor: Ah. Yes 👍

packages/forms/test/form_group_spec.ts
sonukapoor: In order to make the sequence of the different status from the `FormControl` -> `FormGroup` -> `FormArray`, I have subscribed to `statusChanges` of the `FormGroup` as well. I have documented how the status changes at each phase. I think this makes it a little bit easier to understand. LMKWYT

packages/forms/test/form_group_spec.ts
```diff
@@ -1825,6 +1825,391 @@ describe('FormGroup', () => {
...
+           expect(logs).toEqual([
+             'value (g1): {"one":"new!"}',    // g1's call to `setControl` triggered value update
+             'status (g1): PENDING',          // g1's call to `setControl` triggered status update
+             'value (fa): [{"one":"new!"}]',  // g1 update triggers the `FormArray` value update
+             'status (fa): PENDING',          // g1 update triggers the `FormArray` status update
+             'status (c2): INVALID',          // async validator run after `setControl` call
+             'status (g1): PENDING',          // async validator run after `setControl` call
+             'status (fa): PENDING',          // async validator run after `setControl` call
+             'status (g1): INVALID',          // g1 validator completed with the `INVALID` status
+             'status (fa): PENDING',          // fa validator still running
```
AndrewKushnir, line 2199: It looks like `fa` validator was restarted at this point, most likely after `g1` validator completion and status change (pending -> invalid). So FormArray (`fa`) and FormGroup (`g1`) validators were running in parallel at some point and these logs indicate that:

```
'status (g1): PENDING',          // async validator run after `setControl` call
'status (fa): PENDING',          // async validator run after `setControl` call
```

Could you please check if that's the case?

Let's may be also update the test (or timeouts?) to avoid this situation.

AndrewKushnir (commented):
Thanks for updating descriptions in tests, this is super helpful 👍 

I've just left a comment on the last test, everything else looks great!

Since this is a change in behavior, could you please include a draft of the breaking changes note into the commit description (we can review and update it)?

packages/forms/test/form_group_spec.ts
kitti66: เยี่ยม

Neck26638:
🔗

Neck26638:
🔗

packages/forms/test/form_group_spec.ts
sonukapoor: > Could you please check if that's the case?

Actually, `fa` async validator has not run yet, which is why it's in a `PENDING` state. When `g1` completes with `invalid`, it will call `_updateControlsErrors` on the parent of `g1`, which is `fa` and since `fa` has `_hasOwnPendingAsyncValidator` set to true, it will emit `PENDING` again. Once the async validator has run, the state of `fa` is set to `INVALID`. 

Do you still want to change this?

AndrewKushnir: Ok, sounds good, thanks for checking this Sonu!

AndrewKushnir (approved):
LGTM, thanks Sonu 👍

AndrewKushnir:
[Global Presubmit #2](https://test.corp.google.com/OCL:328667912:BASE:328668060:1598501837824:c4a2ea48).

AndrewKushnir:
Note to Caretaker: [the global presubmit went well](https://test.corp.google.com/ui#id=OCL:328667912:BASE:328836898:1598569937525:77ac01f4&include_skipped=true) (only pre-existing failures), however since it's technically a breaking change, it might make sense to merge it and sync as a separate change (in case a rollback is needed). Thank you.

PR #38355, opened 2020-08-05
refactor(compiler-cli): add getTemplateOfComponent to TemplateTypeChecker

### LINKED ISSUE #38352: Ivy LS: API to access the component template
For some of the `TemplateTypeChecker` APIs, it'll be necessary for the language service to use the "real" parsed template nodes, not just nodes independently parsed from the same template. That's because internally, the `TemplateTypeChecker` may need to look up information about the node in maps that are keyed on the node instance.

alxhub:
This commit adds a `getTemplateOfComponent` method to the
`TemplateTypeChecker` API, which retrieves the actual nodes parsed and used
by the compiler for template type-checking. This is advantageous for the
language service, which may need to query other APIs in
`TemplateTypeChecker` that require the same nodes used to bind the template
while generating the TCB.

Fixes #38352

### DIFF
--- a/packages/compiler-cli/src/ngtsc/typecheck/api/checker.ts
+++ b/packages/compiler-cli/src/ngtsc/typecheck/api/checker.ts
@@ -29,6 +29,14 @@ export interface TemplateTypeChecker {
   resetOverrides(): void;
 
   /**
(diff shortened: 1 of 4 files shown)


atscott (approved):
👍  LGTM

PR #38359, opened 2020-08-06
refactor(core): add debug ranges to `LViewDebug` with matchers

mhevery:
Review last SHA only

---

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/core/src/render3/interfaces/node.ts
+++ b/packages/core/src/render3/interfaces/node.ts
@@ -7,12 +7,9 @@
  */
 import {KeyValueArray} from '../../util/array_utils';
 import {TStylingRange} from '../interfaces/styling';
-
-import {DirectiveDef} from './definition';
 import {CssSelector} from './projection';
 import {RNode} from './renderer';
 import {LView, TView} from './view';
-
 
 
 /**
@@ -44,6 +41,20 @@ export const enum TNodeType {
    */
   IcuContainer = 5,
 }
+
+/**
+ * Converts `TNodeType` into human readable text.
+ * Make sure this matches with `TNodeType`
+ */
+export const TNodeTypeAsString = [
+  'Container',         // 0
+  'Projection',        // 1
+  'View',              // 2
+  'Element',           // 3
+  'ElementContainer',  // 4
+  'IcuContainer'       // 5
+] as const;
+
 
 /**
  * Corresponds to the TNode.flags property.
--- a/packages/core/src/render3/interfaces/view.ts
+++ b/packages/core/src/render3/interfaces/view.ts
@@ -15,10 +15,10 @@ import {Sanitizer} from '../../sanitization/sanitizer';
 import {LContainer} from './container';
 import {ComponentDef, ComponentTemplate, DirectiveDef, DirectiveDefList, HostBindingsFunction, PipeDef, PipeDefList, ViewQueriesFunction} from './definition';
 import {I18nUpdateOpCodes, TI18n} from './i18n';
-import {TConstants, TElementNode, TNode, TViewNode} from './node';
+import {TConstants, TElementNode, TNode, TNodeTypeAsString, TViewNode} from './node';
 import {PlayerHandler} from './player';
 import {LQueries, TQueries} from './query';
-import {RElement, Renderer3, RendererFactory3} from './renderer';
+import {RComment, RElement, Renderer3, RendererFactory3} from './renderer';
 import {TStylingKey, TStylingRange} from './styling';
 
 
@@ -69,6 +69,15 @@ export interface OpaqueViewState {
  * don't have to edit the data array based on which views are present.
  */
 export interface LView extends Array<any> {
+  /**
+   * Human readable representation of the `LView`.
+   *
+   * NOTE: This property only exists if `ngDevMode` is set to `true` and it is not present in
+   * production. Its presence is purely to help debug issue in development, and should not be relied
+   * on in production application.
+   */
+  debug?: LViewDebug;
+
   /**
    * The host node for this LView instance, if this is a component view.
    * If this is an embedded view, HOST will be null.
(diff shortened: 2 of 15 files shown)


packages/core/src/render3/interfaces/node.ts
```diff
@@ -45,6 +43,21 @@ export const enum TNodeType {
   IcuContainer = 5,
 }
 
+/**
+ * Converts `TNodeType` into human readable text.
+ * Make sure this matches with `TNodeType`
+ */
+export const TNodeTypeAsString:
+    ['Container', 'Projection', 'View', 'Element', 'ElementContainer', 'IcuContainer'] = [
```
AndrewKushnir, line 51: nit: I think it'd be great if we add a test that verifies that all elements in this array are set defined correctly, i.e. we can check that `TNodeTypeAsString[TNodeType.Container] === 'Container'` and similarly for other fields + also check the length of keys in the `TNodeType` enum and compare it with the `TNodeTypeAsString` length. This should help make sure that if `TNodeTypeAsString` and `TNodeType` are out of sync, it'd be caught by tests.

packages/core/src/render3/interfaces/view.ts
```diff
@@ -69,6 +69,8 @@ export interface OpaqueViewState {
  * don't have to edit the data array based on which views are present.
  */
 export interface LView extends Array<any> {
+  debug?: LViewDebug;
```
AndrewKushnir, line 72: I think adding a comment here would be helpful, to describe when the `debug` is defined and note that it's used for testing purposes only.

packages/core/src/render3/interfaces/view.ts
```diff
@@ -826,3 +828,190 @@ export type TData =
...
+  readonly next: LViewDebug|LContainerDebug|null;
+  readonly hasTransplantedViews: boolean;
+}
+
+
+
+/**
+ * `LView` is subdivided to ranges where the actual data is stored. Some of these ranges such as
+ * `decls` and `vars` are known at compile time. Other such as `i18n` and `expando` are runtime only
+ * concepts.
```
AndrewKushnir, line 946: Since the name of this data structure doesn't contain `Debug` in it, may be add a comment that it's a debug-only interface?

packages/core/src/render3/interfaces/view.ts
```diff
@@ -826,3 +828,190 @@ export type TData =
...
+   * The merged content of the range. `t` contains data from `TView.data` and `l` contains `LView`
+   * data at an index.
+   */
+  content: LViewRangeContent[];
+}
+
+/**
+ * For convenience the static and instance portions of `TView` and `LView` are merged into a single
+ * object in `LViewRange`.
+ */
```
AndrewKushnir, line 974: Similar to `LViewRange` - may be add a comment to mention that it's debug-only data structure?

packages/core/src/render3/interfaces/node.ts
AndrewKushnir: Do we need to define type explicitly here? We can probably use `as const` here instead:
```
export const TNodeTypeAsString = [ ... ] as const;
```

packages/core/src/render3/interfaces/view.ts
```diff
@@ -826,3 +828,190 @@ export type TData =
...
+export interface DebugNode {
+  /**
+   * HTML representation of the node.
+   */
+  html: string|null;
+
+  /**
+   * Human readable node type.
+   */
+  type: 'Container'|'Projection'|'View'|'Element'|'ElementContainer'|'IcuContainer';
```
AndrewKushnir, line 1006: I think this should work here:
```suggestion
  type: typeof TNodeTypeAsString[number];
```

AndrewKushnir (commented):
Thanks for improving debugging experience @mhevery 👍

packages/core/src/render3/interfaces/node.ts
mhevery: added test and `as const` (did not know about that trick, thank you.)

packages/core/src/render3/interfaces/node.ts
```diff
@@ -45,6 +43,21 @@ export const enum TNodeType {
...
+      'Container',         // 0
+      'Projection',        // 1
+      'View',              // 2
+      'Element',           // 3
+      'ElementContainer',  // 4
+      'IcuContainer'       // 4
```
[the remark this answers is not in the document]
mhevery: fixed

packages/core/src/render3/interfaces/view.ts
mhevery: added.

packages/core/src/render3/interfaces/view.ts
mhevery: done

packages/core/src/render3/interfaces/view.ts
mhevery: done

packages/core/src/render3/interfaces/view.ts
mhevery: Ohh, that is cool. Did not know that. Thanks for showing me a new trick.

mhevery:
[presubmit](http://test/OCL:325324045:BASE:325324075:1596753036989:24ebdf2)

AndrewKushnir (approved):
LGTM 👍

josephperrott (approved):
LGTM

Reviewed-For: dev-infra

AndrewKushnir (approved):
Reviewed-for: size-tracking

AndrewKushnir:
@josephperrott for some reasons the `circular-dependencies` group was not added to this PR even though the `goldens/circular-deps/packages.json` file was modified...

atscott:
> @josephperrott for some reasons the `circular-dependencies` group was not added to this PR by PullApprove even though the `goldens/circular-deps/packages.json` file was modified...

@AndrewKushnir this is a rare case where both circular-deps and size-tracking are required and is due to the phased review. circular-deps will likely be requested once size-tracking has approved.

atscott (approved):
Reviewed-for: size-tracking

josephperrott:
Correct, as @atscott mentioned its a weird edge case.  However, @AndrewKushnir has already provided the approval for `circular-dependencies` so that group is already approved once added.

AndrewKushnir:
Thanks for additional information @josephperrott and @atscott 👍

PR #38364, opened 2020-08-06
docs(service-worker): describe how asset-/data-group order affects request handling

### LINKED ISSUE #21189: [Feature Request] [Service Worker] DataGroups configurable with query parameters
## I'm submitting a...

<pre><code>
[ ] Bug report  
[x] Feature request
[x] Documentation issue or request
</code></pre>

## Current behavior

Right now, Service Worker will cache data in the cache using the entire URL as the key.  It would be nice to have data groups of the same base URL to have different expiration / performance policies for query parameters.

For example:
We have an API: `/documents`.  It provides a list of documents.  It also returns an ID called the `knowledge` parameter, which is a special ID that represents the current state of the user's document list.  I'd like to have this call use the `'performance'` strategy.

We can use the API `/documents?knowledge=KNOWLEDGE_ID` to fetch the delta of documents that have changed.  I'd like to have this call use the `'freshness'` strategy.

## Expected behavior

I am not sure if this is already supported, and in that case, it should be documented.

If it is not supported, I have been brainstorming of some ideas.

One is that the `dataGroups` entry in the `ngsw-config.json` could have a property: `queryParameters`, like this:

```json
"dataGroups": [{
    "name": "documents-fetch",
    "urls": [ "/documents" ],
    "strategy": "performance"
}, {
    "name": "documents-sync",
    "urls": [ "/documents" ],
    "queryParameters": [ "knowledge" ],
    "strategy": "freshness"
}]
```

Just brainstorming, but maybe the algorithm could select the _most specific_ query parameters for the same URLs if the URL matches in a group.

N/A

To have configurable SW API caching based on query parameters as well as URL.

## Others:

This article isn't exactly the same thing, but there might be some prior work within Google and `sw-precache`: https://developers.google.com/web/showcase/2015/service-workers-iowa#watch_out_for_extra_query_parameters

gkalpak:
The order of asset- and data-groups in `ngsw-config.json` affects how a request is handled by the ServiceWorker. Previously, this was not clearly documented.

This commit describes how the order of asset-/data-groups affects request handling.

Closes #21189.

### DIFF
--- a/aio/content/guide/service-worker-config.md
+++ b/aio/content/guide/service-worker-config.md
@@ -66,10 +66,10 @@ This field contains an array of asset groups, each of which defines a set of ass
 
 <div class="alert is-helpful">
 
-When the ServiceWorker is handling a request, it checks asset groups in the order in which they appear in `ngsw-config.json`.
+When the ServiceWorker handles a request, it checks asset groups in the order in which they appear in `ngsw-config.json`.
 The first asset group that matches the requested resource handles the request.
 
-Therefore, you should put the more specific asset groups higher in the list.
+It is recommended that you put the more specific asset groups higher in the list.
 For example, an asset group that matches `/foo.js` should appear before one that matches `*.js`.
 
 </div>
@@ -153,10 +153,10 @@ This field contains an array of data groups, each of which defines a set of data
 
 <div class="alert is-helpful">
 
-When the ServiceWorker is handling a request, it checks data groups in the order in which they appear in `ngsw-config.json`.
+When the ServiceWorker handles a request, it checks data groups in the order in which they appear in `ngsw-config.json`.
 The first data group that matches the requested resource handles the request.
 
-Therefore, you should put the more specific data groups higher in the list.
+It is recommended that you put the more specific data groups higher in the list.
 For example, a data group that matches `/api/foo.json` should appear before one that matches `/api/*.json`.
 
 </div>
--- a/aio/content/guide/service-worker-config.md
+++ b/aio/content/guide/service-worker-config.md
@@ -53,13 +53,26 @@ This field contains an array of asset groups, each of which defines a set of ass
 
 ```json
 {
-  "assetGroups": [{
-    ...
-  }, {
-    ...
-  }]
+  "assetGroups": [
+    {
+      ...
+    },
+    {
+      ...
+    }
+  ]
 }
 ```
+
+<div class="alert is-helpful">
+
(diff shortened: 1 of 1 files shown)


aio/content/guide/service-worker-config.md
```diff
@@ -53,14 +53,27 @@ This field contains an array of asset groups, each of which defines a set of ass
...
+    {
+      ...
+    }
+  ]
 }
 ```
 
+
+When the ServiceWorker is handling a request, it checks asset groups in the order in which they appear in `ngsw-config.json`.
```
aikidave, line 69:
```suggestion
When the ServiceWorker handles a request, it checks asset groups in the order in which they appear in `ngsw-config.json`.
```

aio/content/guide/service-worker-config.md
```diff
@@ -53,14 +53,27 @@ This field contains an array of asset groups, each of which defines a set of ass
...
+The first asset group that matches the requested resource handles the request.
+
+Therefore, you should put the more specific asset groups higher in the list.
```
aikidave, line 72:
```suggestion
It is recommended that you put more specific asset groups higher in the list.
```
"Therefore" is a bit archaic. This rewording makes the fact that we're recommending a course of action clearer.

aio/content/guide/service-worker-config.md
```diff
@@ -123,6 +136,31 @@ These options are used to modify the matching behavior of requests. They are pas
...
+    {
+      ...
+    }
+  ]
+}
+```
+
+<div class="alert is-helpful">
+
+When the ServiceWorker is handling a request, it checks data groups in the order in which they appear in `ngsw-config.json`.
```
aikidave, line 156:
```suggestion
When the ServiceWorker handles a request, it checks data groups in the order in which they appear in `ngsw-config.json`.
```

aio/content/guide/service-worker-config.md
```diff
@@ -123,6 +136,31 @@ These options are used to modify the matching behavior of requests. They are pas
...
+The first data group that matches the requested resource handles the request.
+
+Therefore, you should put the more specific data groups higher in the list.
```
aikidave, line 159:
```suggestion
It is recommended that you put more specific data groups higher in the list.
```
"Therefore" is a bit archaic. This rewording makes the fact that we're recommending a course of action clearer.

aikithoughts (approved):
Just a few minor changes to tighten up the language.

PR #38353, opened 2020-08-05
docs: Restructure table of contents

aikithoughts:
This PR attempts to address how overwhelmed new Angular developers feel when they read our documentation. It restructures our content to put more focus on critical content areas.

## PR Checklist
Please check if your PR fulfills the following requirements:

- [ ] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit
- [ ] Tests for the changes have been added (for bug fixes / features)
- [X] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [X] Documentation content changes
- [ ] angular.io application / infrastructure changes
- [ ] Other... Please describe:

## What is the current behavior?

Issue Number: N/A

## Does this PR introduce a breaking change?

- [ ] Yes
- [X] No

PR #38365, opened 2020-08-06
fix(docs-infra): correctly generate CLI commands docs when the overview page moves

gkalpak:
Previously, the [processCliCommands][1] dgeni processor, which is used to generate the docs pages for the CLI commands, expected the CLI commands overview page (with a URL of `cli`) to exist as a child of a top-level navigation section (`CLI Commands`). If one tried to move the `CLI Commands` section inside another section, `processCliCommnads` would fail to find it and thus fail to generate the CLI commands docs. This problem came up in #38353.

This commit updates the `processCliCommands` processor to be able to find it regardless of the position of the `CLI Commands` section inside the navigation doc.

[1]: https://github.com/angular/angular/blob/dca4443a8ed65d341356fe62f6df9934e6314dfe/aio/tools/transforms/cli-docs-package/processors/processCliCommands.js#L7-L9


aikithoughts (approved):
Looks good to me! Thank you so much!

PR #38367, opened 2020-08-06
fix(platform-server): remove styles added by ServerStylesHost on destruction

josephperrott:
When a `ServerStylesHost` instance is destroyed, all of the shared styles added to the DOM
head element by that instance should be removed.  Without this removal, over time a large
number of style rules will build up and cause extra memory pressure.  This brings the
`ServerStylesHost` in line with the `DomStylesHost` used by the platform browser, which
performs this same cleanup.


alxhub (approved):

josephperrott:
[Presubmit](https://test.corp.google.com/OCL:325312900:BASE:325864201:1597088191745:46ff0956)

PR #38371, opened 2020-08-07
refactor(forms): get rid of duplicate functions

AndrewKushnir:
This commit performs minor refactoring in Forms package to get rid of duplicated functions.
It looks like the functions were duplicated due to a slightly different type signatures, but
their logic is completely identical. The logic in retained functions remains the same and now
these function also accept a generic type to achieve the same level of type safety.

## PR Type

- [ ] Yes
- [x] No


atscott (approved):
Nice refactor!

AndrewKushnir:
[Presubmit](http://test/OCL:325451151:BASE:325451194:1596818561811:7ffcfd85)

PR #38375, opened 2020-08-07
fix(router): If users are using the Alt key when clicking the router links, prioritize browser’s default behavior

yuheiy:
When users click a link while holding the Alt key together, the browsers behave as follows.

Windows 10:

|Browser|Behavior|
|:--|:--|
|Chrome 84|Download the target resource|
|Firefox 79|Prevent navigation and therefore do nothing|
|Edge 84|Download the target resource|
|IE 11|No impact|

macOS Catalina:

|Browser|Behavior|
|:--|:--|
|Chrome 84|Download the target resource|
|Firefox 79|Prevent navigation and therefore do nothing|
|Safari 13|Download the target resource|

The Alt key has no impact on IE, but on most other browsers it does. As with other modifier keys, the router should stop the original navigation to avoid preventing the browser’s default behavior.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] Yes
- [x] No

This doesn't have an associate test for a few reasons:
* No tests exist for the meta keys already, so adding one for this would take more effort
* The test, if it were to be a unit test, wouldn't provide much value because it would have to pass in `true` specifically for the meta key and simply expect `true` back. This is akin to a change detector test
* The ideal test would be a protractor test that verifies the behavior as described above -- i.e. that the Angular router does not trigger a navigation by clicking the routerLink, but instead the browser handles the click natively. This may be a little overkill to write for this simple change.

### DIFF
--- a/goldens/public-api/router/router.d.ts
+++ b/goldens/public-api/router/router.d.ts
@@ -432,7 +432,7 @@ export declare class RouterLinkWithHref implements OnChanges, OnDestroy {
     constructor(router: Router, route: ActivatedRoute, locationStrategy: LocationStrategy);
     ngOnChanges(changes: SimpleChanges): any;
     ngOnDestroy(): any;
-    onClick(button: number, ctrlKey: boolean, metaKey: boolean, shiftKey: boolean): boolean;
+    onClick(button: number, ctrlKey: boolean, shiftKey: boolean, altKey: boolean, metaKey: boolean): boolean;
 }
 
 export declare class RouterModule {
--- a/packages/router/src/directives/router_link.ts
+++ b/packages/router/src/directives/router_link.ts
@@ -359,9 +359,12 @@ export class RouterLinkWithHref implements OnChanges, OnDestroy {
   }
 
   /** @nodoc */
-  @HostListener('click', ['$event.button', '$event.ctrlKey', '$event.metaKey', '$event.shiftKey'])
-  onClick(button: number, ctrlKey: boolean, metaKey: boolean, shiftKey: boolean): boolean {
-    if (button !== 0 || ctrlKey || metaKey || shiftKey) {
+  @HostListener(
+      'click',
+      ['$event.button', '$event.ctrlKey', '$event.shiftKey', '$event.altKey', '$event.metaKey'])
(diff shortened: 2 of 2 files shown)


atscott:
Thanks @yuheiy, this is a really great summary of the behavior in each browser! Would you mind copying that summary in the PR description into the commit message? We try to make all the information immediately available there rather than having to try to find the associated PR after merging.

Can you also run `yarn bazel run //packages/router:router_api.accept` and add the new golden to the commit so the CI test passes?

yuheiy:
I’ve updated the commit message and added the golden!

atscott (approved):

atscott (approved):
reviewed-for: public-api
reviewed-for: fw-router

AndrewKushnir (approved):
Reviewed-for: public-api

jelbourn:
@atscott @AndrewKushnir if we don't consider `onClick` to be public API here, could we mark it as `@internal`?

atscott:
> @atscott @AndrewKushnir if we don't consider `onClick` to be public API here, could we mark it as `@internal`?

Yea, we probably should mark it as internal.

atscott:
> @atscott @AndrewKushnir if we don't consider `onClick` to be public API here, could we mark it as `@internal`?

@jelbourn - I created a separate PR for this https://github.com/angular/angular/pull/38448

petebacondarwin (approved):
Reviewed-for: public-api

The addition of `/** @nodoc */` in a recent commit has caused a conflict. Can you please rebase?

yuheiy:
@petebacondarwin I rebased.

petebacondarwin:
Caretaker - merge-assistance: the broken test is unrelated to this PR.

petebacondarwin:
Sorry I was a bit to fast with the merge label. This change needs to go through presubmit against the internal Google repository.

atscott:
[presubmit](https://test.corp.google.com/ui#id=OCL:325441130:BASE:329725338:1599063341142:7088668c)

jelbourn (approved):
LGTM

Reviewed-for: public-api

PR #38378, opened 2020-08-07
test: update components repo to test against recent revision (#38273)

JoostK:
Backports 570d156ce469ef6ff5e9d2863f807a9dafd2da26 to patch. The original PR (#38273) was master-only as it contained another commit that was only supposed to go into master. On the patch branch, however, we do want to test against the same components repo commit in order to avoid discrepancies when backporting upcoming fixes.


JoostK:
merge-assistance: this has been approved for master already.

devversion (approved):

JoostK:
I probably should have known that@devversion would approve this before the merge-assistence label became of relevance 😄

AndrewKushnir:
Thanks for back porting this change to the patch branch 👍 This PR is now merged.

PR #38379, opened 2020-08-07
fix(router): support lazy loading for empty path named outlets

### LINKED ISSUE #12842: Lazy loaded module in named outlet throws error
[x] bug report 
[ ] feature request
[ ] support request 

**Current behavior**
Navigating to route with lazy loaded module in named outlet throws error: 

> Uncaught (in promise): TypeError: Cannot read property 'routes' of undefined

**Expected behavior**
Navigating to route with lazy loaded module in named outlet without errors.

When I have configured my routes with named outlet and want to have lazy loaded module in that outlet. 

```TypeScript
export const routes: Routes = [
 {  
   path: '',
   component: AppComponent,
   children: [
     {
       path: '',
       loadChildren: "app/home/home.module#HomeModule"
     },   {
       path: '',
       loadChildren: "app/about/about.module#AboutModule",
       outlet: "about"
     }
   ]
  }
];
```
Here is the plnkr to demonstrate the behavior: http://plnkr.co/edit/nK9kJYdakXILZU0UKFrP?p=preview 
(Anything is displayed because of error)

Using lazy loaded modules in named outlets.

Please tell us about your environment:

Angular version: 2.0.X

Browser: [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

Language: [all | TypeScript X.X | ES6/7 | ES5]

Node (for AoT issues): node --version =

atscott:
In general, the router only matches and loads a single Route config tree. However,
named outlets with empty paths are a special case where the router can
and should actually match two different `Route`s and ensure that the
modules are loaded for each match.

This change updates the "ApplyRedirects" stage to ensure that named
outlets with empty paths finish loading their configs before proceeding
to the next stage in the routing pipe. This is necessary because if the
named outlet has `loadChildren` but the associated lazy config is not loaded
before following stages attempt to match and activate relevant `Route`s,
an error will occur.

fixes #12842

[global presubmit](https://test.corp.google.com/ui#id=OCL:324899076:BASE:325249448:1596762883117:fc79df48) - Will want to do another rerun as I made a tiny change since then to pass the test "should work work with redirects when other outlet comes before the one being activated"

### DIFF
--- a/packages/router/src/apply_redirects.ts
+++ b/packages/router/src/apply_redirects.ts
@@ -8,7 +8,7 @@
 
 import {Injector, NgModuleRef} from '@angular/core';
 import {defer, EmptyError, Observable, Observer, of} from 'rxjs';
-import {catchError, concatAll, first, map, mergeMap, tap} from 'rxjs/operators';
+import {catchError, first, map, mergeMap, switchMap, tap} from 'rxjs/operators';
 
 import {LoadedRouterConfig, Route, Routes} from './config';
 import {CanLoadFn} from './interfaces';
@@ -148,28 +148,47 @@ class ApplyRedirects {
       ngModule: NgModuleRef<any>, segmentGroup: UrlSegmentGroup, routes: Route[],
       segments: UrlSegment[], outlet: string,
       allowRedirects: boolean): Observable<UrlSegmentGroup> {
-    return of(...routes).pipe(
-        map((r: any) => {
-          const expanded$ = this.expandSegmentAgainstRoute(
-              ngModule, segmentGroup, routes, r, segments, outlet, allowRedirects);
-          return expanded$.pipe(catchError((e: any) => {
-            if (e instanceof NoMatch) {
-              // TODO(i): this return type doesn't match the declared Observable<UrlSegmentGroup> -
-              // talk to Jason
-              return of(null) as any;
-            }
-            throw e;
-          }));
-        }),
-        concatAll(), first((s: any) => !!s), catchError((e: any, _: any) => {
-          if (e instanceof EmptyError || e.name === 'EmptyError') {
-            if (this.noLeftoversInUrl(segmentGroup, segments, outlet)) {
-              return of(new UrlSegmentGroup([], {}));
-            }
-            throw new NoMatch(segmentGroup);
-          }
-          throw e;
-        }));
+    type MatchedSegment = {segment: UrlSegmentGroup, outlet: string};
+    // This logic takes each route and switches to a new observable that depends on the result of
+    // the previous route expansion. In this way, we compose a list of results where each one can
+    // depend on and look at the previous to determine how to proceed with expansion of the
+    // current route.
+    return routes
+        .reduce(
+            (accumulatedResults: Observable<Array<MatchedSegment>>, r: Route) => {
+              return accumulatedResults.pipe(switchMap(resultsThusFar => {
+                // If we already matched a previous `Route` with the same outlet as the current,
+                // we should not process the current one.
+                if (resultsThusFar.some(result => result && result.outlet === getOutlet(r))) {
+                  return of(resultsThusFar);
+                }
+                const expanded$ = this.expandSegmentAgainstRoute(
+                    ngModule, segmentGroup, routes, r, segments, outlet, allowRedirects);
+                return expanded$.pipe(
+                    map((segment) => resultsThusFar.concat({segment, outlet: getOutlet(r)})),
+                    catchError((e: any) => {
+                      if (e instanceof NoMatch) {
+                        return of(resultsThusFar);
+                      }
+                      throw e;
+                    }));
+              }));
+            },
+            of([] as MatchedSegment[]))
+        .pipe(
+            // Find the matched segment whose outlet matches the one we're looking for.
+            map(results => results.find(s => s.outlet === outlet)?.segment),
+            first((s): s is UrlSegmentGroup => s !== undefined),
+            catchError((e: any, _: any) => {
+              if (e instanceof EmptyError || e.name === 'EmptyError') {
+                if (this.noLeftoversInUrl(segmentGroup, segments, outlet)) {
+                  return of(new UrlSegmentGroup([], {}));
+                }
+                throw new NoMatch(segmentGroup);
+              }
+              throw e;
+            }),
+        );
   }
 
   private noLeftoversInUrl(segmentGroup: UrlSegmentGroup, segments: UrlSegment[], outlet: string):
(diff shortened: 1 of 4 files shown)


packages/router/src/apply_redirects.ts
```diff
@@ -148,28 +148,47 @@ class ApplyRedirects {
...
-              return of(new UrlSegmentGroup([], {}));
-            }
-            throw new NoMatch(segmentGroup);
-          }
-          throw e;
-        }));
+    type MatchedSegment = {segment: UrlSegmentGroup, outlet: string};
+    // This logic takes each route and switches to a new observable that depends on the result of
+    // the previous route expansion. In this way, we compose a list of results where each one can
+    // depend on and look at the the previous to determine how to proceed with expansion of the
```
acohenOT, line 154: nit: `the the`

atscott:
[new global presubmit](https://test.corp.google.com/ui#id=OCL:325465574:BASE:325836076:1597107798934:3443e018)

packages/router/src/apply_redirects.ts
```diff
@@ -148,28 +148,47 @@ class ApplyRedirects {
...
+                    map((segment) => resultsThusFar.concat({segment, outlet: getOutlet(r)})),
+                    catchError((e: any) => {
+                      if (e instanceof NoMatch) {
+                        return of(resultsThusFar);
+                      }
+                      throw e;
+                    }));
+              }));
+            },
+            of([]) as Observable<Array<MatchedSegment>>)
```
alxhub, line 177: Either way is fine, but this feels a little simpler:
```suggestion
            of([] as Array<MatchedSegment>>))
```

alxhub (approved):

packages/router/test/apply_redirects.spec.ts
```diff
@@ -754,6 +837,47 @@ describe('applyRedirects', () => {
...
+             .subscribe(
+                 (tree: UrlTree) => {
+                   expect(tree.toString()).toEqual('/b');
+                 },
+                 () => {
+                   fail('should not be reached');
+                 });
+       });
+
+    it('should work work when entry point is named outlet', () => {
```
acohenOT, line 864:
```suggestion
    it('should work when entry point is named outlet', () => {
```

packages/router/test/apply_redirects.spec.ts
```diff
@@ -754,6 +837,47 @@ describe('applyRedirects', () => {
     });
   });
 
+  describe('multiple matches with empty path named outlets', () => {
+    it('should work work with redirects when other outlet comes before the one being activated',
```
acohenOT, line 841:
```suggestion
    it('should work with redirects when other outlet comes before the one being activated',
```

petebacondarwin (approved):
Reviewed-for: size-tracking

The size change only affects the lazy-load integration test, as should be expected.

jelbourn (approved):
LGTM

Reviewed-for: size-tracking

mhevery:
Reverted: [reason](https://sponge.corp.google.com/target?id=14096760-7e52-4ba4-9d4e-7224b4c19e84&target=//cloud/console/integration/tests/shell/navigation:redirect_hermetic_chrome-linux&searchFor=&show=FAILED&sortBy=STATUS)

asvishnyakov:
@mhevery Might you please post reasons in public? I saw links like this and references to internal google issues in the past, and I don't understand why public project like angular should revert changes and wait fixes in google services or adopt to them (and not vice versa), even though it obviously sponsored by Google and developed by Google developers.

atscott:
Hi @asvishnyakov, I understand how frustrating it is not to be able to get complete information about why something was reverted! This change should not have been a breaking change, so any broken test inside Google likely indicates that there's an issue with the implementation. The internal links, while not helpful to external developers, are required for us to reference so we can do more investigation. 
We immediately revert commits that have been found to break internal tests because we want to ensure the framework is stable and does not introduce unexpected breaking changes for _anyone_, internal or external to Google. This generally should happen _before_ a PR is merged, but sometimes failures are detected late in the game. 
Until we know the root cause for the failure, there's unfortunately not really much more we can offer. Once we find the cause, we do our best circle back on the issues and provide more information.

atscott:
@alxhub PTAL

The internal failure was due to a stack overflow caused by the `switchMap` approach (test added that failed in the previous version). The new approach is to 
1. separate the `Route`s by their `outlet` names
2.  run the exact same algorithm as before, but ignoring errors from expanding routes of outlets that differ from the one being activated (I wish the git diff was showing this better). 
3. Wait for each of the observables-by-outlet to complete and return only the one for the outlet being activated

This has the added benefit of loading all matched outlets concurrently. Additionally, the `concatAll, first()` is able to exit the processing as soon as a match is found rather than proceed with the `switchMaps` and produce no-ops for following routes of the same outlet like the old algorithm would.

[global presubmit](https://test.corp.google.com/ui#id=OCL:329520431:BASE:329494703:1598981400357:7356e111)

packages/router/src/apply_redirects.ts
```diff
@@ -148,28 +149,53 @@ class ApplyRedirects {
...
+    // We need to expand each outlet group independently to ensure that we not only load modules
+    // for routes matching the given `outlet`, but also those which will be activated because
+    // their path is empty string. This can result in multiple outlets being activated at once.
+    const routesByOutlet: Map<string, Route[]> = groupRoutesByOutlet(routes);
+    if (!routesByOutlet.has(outlet)) {
+      routesByOutlet.set(outlet, []);
+    }
+
+    const expandRoutes = (routes: Route[]) => {
+      return of(...routes).pipe(
```
alxhub, line 161: I don't think this is any different than:

```suggestion
      return from(routes).pipe(
```

packages/router/src/apply_redirects.ts
```diff
@@ -148,28 +149,53 @@ class ApplyRedirects {
...
+            const expanded$ = this.expandSegmentAgainstRoute(
+                ngModule, segmentGroup, routes, r, segments, outlet, allowRedirects);
+            return expanded$.pipe(catchError(e => {
+              if (e instanceof NoMatch) {
+                return of(null);
+              }
+              throw e;
+            }));
+          }),
+          concatAll(),
```
alxhub, line 172: Not that I mind the explicitness, but I think `map(), `concatAll()` is equivalent to `concatMap()`. I dunno if the router has any preference for one or the other.

alxhub (approved):

danylobilokha:
hi guys, any update on that? Stuck with this issue and need this PR

atscott:
@danylobilokha it’s available in the v11 release, currently 11.0.0-next.2

Issue #37837, opened 2020-06-30
core.js:6185 ERROR TypeError: Cannot read property 'pipe' of undefined     at MatSelect.ngAfterContentInit


jellypop:
🛑🛑🛑🛑🛑🛑🛑🛑🛑🛑🛑🛑🛑🛑🛑🛑🛑🛑🛑🛑🛑🛑🛑🛑🛑🛑🛑

Please file any Angular Components issues at: https://github.com/angular/components/issues/new

For the time being, we keep Angular Components issues in a separate repository.

🛑🛑🛑🛑🛑🛑🛑🛑🛑🛑🛑🛑🛑🛑🛑🛑🛑🛑🛑🛑🛑🛑🛑🛑🛑🛑🛑

pkozlowski-opensource:
Thanks for reporting this issue. However, you didn't provide sufficient information for us to understand and reproduce the problem. Please check out [our submission guidelines](https://github.com/angular/angular/blob/master/CONTRIBUTING.md#-submitting-an-issue) to understand why we can't act on issues that are lacking important information.

If the problem still exists in your application, please [open a new issue](https://github.com/angular/angular/issues/new/choose) and follow the instructions in the issue template.

PR #38380, opened 2020-08-07
refactor(forms): get rid of duplicate functions [patch-only version]

AndrewKushnir:
This is a patch-only version of the PR #37837, which didn't merge cleanly into the patch branch due to the fact that the `packages/core/test/bundling/forms` test app (which also keeps track of the golden symbols) is not present in the `10.0.x` branch. 

--------

This commit performs minor refactoring in Forms package to get rid of duplicate functions.
It looks like the functions were duplicated due to a slightly different type signatures, but
their logic is completely identical. The logic in retained functions remains the same and now
these function also accept a generic type to achieve the same level of type safety.

## PR Type

- [ ] Yes
- [x] No


AndrewKushnir:
This PR is now merged into the patch branch.

PR #38383, opened 2020-08-07
docs: update team contributors page

cindygk:
Removing Kara, Denny, Judy, Tony who are no longer actively working on the project


sonukapoor:
@cindygk I think we can also delete the pictures.

cindygk:
Thanks for the suggestion, Sonu.  The page has a large number of photos that need to be removed so I will take care of these at a later date.

PR #38385, opened 2020-08-08
docs: keep tables in order

ajitsinghkaler:
The things that are explained in first table are there in
second table while explaining the structural directive microsyntax
So the second table should preceed the first table

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/aio/content/guide/structural-directives.md
+++ b/aio/content/guide/structural-directives.md
@@ -318,6 +318,22 @@ When you write your own structural directives, use the following grammar:
 
 The following tables describe each portion of the microsyntax grammar.
 
+<!-- The items in this table seem different. Is there another name for how we should describe them? -->
+<table>
+  <tr>
+    <th></th>
+  </tr>
+  <tr>
+    <td colspan="3"><code>keyExp = :key ":"? :expression ("as" :local)? ";"? </code></td>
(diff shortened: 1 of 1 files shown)


kapunahelewong (commented):
Thank you for this, @ajitsinghkaler! I think this section needs more help, but this is a step in the right direction. I think I left those comments about what to put in the headers. It would be nice if we had some sort of explanation around those tables, but I don't understand the nuances enough to describe it. Do you know? I'd be happy to help you compose a couple of sentences from a docs perspective if you could share the expertise on how it works. If not, we can just move forward with this change as it is. What do you think?

ajitsinghkaler:
@kapunahelewong I won't have time for a month or so if this change seems appropriate we can move forward with the way it is.

mhevery (approved):

kapunahelewong:
Sure, @ajitsinghkaler. 🙂 what do you think about us merging it after the substantial rewrite in https://github.com/angular/angular/pull/40015? That way I won't accidentally overwrite this change with that PR.

PR #38387, opened 2020-08-09
fix(compiler-cli): preserve quotes in class member names

### LINKED ISSUE #38311: Using an Angular Decorator like @Output with a string literal class member causes a build error
# 🐞 bug report

### Affected Package

 The issue is caused by package @angular/compiler-cli@10.0.6

### Is this a regression?

 Yes, the previous version in which this bug was not present was: @angular/compiler-cli@10.0.5

### Description

I think this is related to the feature added in PR https://github.com/angular/angular/pull/38226

I had some code like this before:

```
@Output() 'did-finish-load' = new EventEmitter();
```

(`did-finish-load` is the name of an [Electron API](https://www.electronjs.org/docs/api/web-contents#event-did-finish-load) I was using, I was just trying to keep the same naming scheme).

It looks like before, output names like `'did-finish-load'` were just ignored completely.
In the newer Angular version, it causes a mysterious build error.

## 🔬 Minimal Reproduction

I don't think I can control the compiler-cli version on StackBlitz, so I'll submit a github repo.

https://github.com/patdx/reproduction-string-literal-angular

Steps:

1. npm install
2. ng build
3. You should get the same error as reported below

## 🔥 Exception or Error

I just tried creating a small test:
```
export class AppComponent {
  @Output() 'did-finish-load-one' = new EventEmitter();
  @Output() didFinishLoadTwo = new EventEmitter();
}
```

With `@angular/compiler-cli@10.0.5`, it produced the following code in dist/main.js:

```
class AppComponent {
  constructor() {
    this[
      "did-finish-load-one"
    ] = new _angular_core__WEBPACK_IMPORTED_MODULE_0__["EventEmitter"]();
    this.didFinishLoadTwo = new _angular_core__WEBPACK_IMPORTED_MODULE_0__[
      "EventEmitter"
    ]();
  }
}
AppComponent.ɵfac = function AppComponent_Factory(t) {
  return new (t || AppComponent)();
};
AppComponent.ɵcmp = _angular_core__WEBPACK_IMPORTED_MODULE_0__[
  "ɵɵdefineComponent"
]({
  type: AppComponent,
  selectors: [["app-root"]],
  outputs: { didFinishLoadTwo: "didFinishLoadTwo" },
  decls: 0,
  vars: 0,
  template: function AppComponent_Template(rf, ctx) {},
  encapsulation: 2,
});
/*@__PURE__*/ (function () {
  _angular_core__WEBPACK_IMPORTED_MODULE_0__["ɵsetClassMetadata"](
    AppComponent,
    [
      {
        type: _angular_core__WEBPACK_IMPORTED_MODULE_0__["Component"],
        args: [
          {
            selector: "app-root",
            template: "",
          },
        ],
      },
    ],
    null,
    {
      didFinishLoadTwo: [
        {
          type: _angular_core__WEBPACK_IMPORTED_MODULE_0__["Output"],
        },
      ],
    }
  );
})();

```

This code is running, but "did-finish-load-one" was totally ignored as an output. Oh! Actually, I didn't realize this before. 

With `@angular/compiler-cli@10.0.6`, I got a build error instead:

```bash
# ...
chunk {styles} styles.js, styles.js.map (styles) 12.4 kB [initial] [rendered]
chunk {vendor} vendor.js, vendor.js.map (vendor) 2.07 MB [initial] [rendered]
Date: 2020-07-31T04:45:05.022Z - Hash: ad1181a1bd231bc94cd9 - Time: 5662ms

ERROR in ./src/app/app.component.ts 17:19
Module parse failed: Unexpected token (17:19)
File was processed with these loaders:
 * ./node_modules/@ngtools/webpack/src/index.js
You may need an additional loader to handle the result of these loaders.
|                 template: '',
|             }]
>     }], null, { did-finish-load-one: [{
|             type: Output
|         }], didFinishLoadTwo: [{
```

It guess if the compiler supports string literals with decorators like Output, it needs to escape the value here, as in

```diff
                 template: '',
             }]
-     }], null, { did-finish-load-one: [{
+     }], null, { 'did-finish-load-one': [{
             type: Output
         }], didFinishLoadTwo: [{
```

## 🌍  Your Environment

**Angular Version:**
<pre><code>
Angular CLI: 10.0.4
Node: 12.3.1
OS: win32 x64

Angular: 10.0.7
... animations, common, compiler, core, forms, platform-browser
Ivy Workspace: Yes
</code></pre>

**Anything else relevant?**

As far as I can tell, all that matters is `@angular/compiler-cli@10.0.5` (ignored but no error) vs `@angular/compiler-cli@10.0.6` (build error).

crisbeto:
When we were outputting class members for `setClassMetadata` calls, we were using the string representation of the member name. This can lead to us generating invalid code when the name contains dashes and is quoted (e.g. `@Output() 'has-dashes' = new EventEmitter()`), because the quotes will be stripped for the string representation.

These changes fix the issue by using the original name AST node that was used for the declaration and which knows whether it's supposed to be quoted or not.

Fixes #38311.

### DIFF
--- a/packages/compiler-cli/src/ngtsc/annotations/src/metadata.ts
+++ b/packages/compiler-cli/src/ngtsc/annotations/src/metadata.ts
@@ -70,8 +70,8 @@ export function generateSetClassMetadataCall(
         `Duplicate decorated properties found on class '${clazz.name.text}': ` +
         duplicateDecoratedMemberNames.join(', '));
   }
-  const decoratedMembers =
-      classMembers.map(member => classMemberToMetadata(member.name, member.decorators!, isCore));
+  const decoratedMembers = classMembers.map(
+      member => classMemberToMetadata(member.nameNode || member.name, member.decorators!, isCore));
   if (decoratedMembers.length > 0) {
     metaPropDecorators = ts.createObjectLiteral(decoratedMembers);
   }
@@ -127,7 +127,7 @@ function ctorParameterToMetadata(
  * Convert a reflected class member to metadata.
  */
 function classMemberToMetadata(
-    name: string, decorators: Decorator[], isCore: boolean): ts.PropertyAssignment {
+    name: ts.PropertyName|string, decorators: Decorator[], isCore: boolean): ts.PropertyAssignment {
   const ngDecorators = decorators.filter(dec => isAngularDecorator(dec, isCore))
                            .map((decorator: Decorator) => decoratorToMetadata(decorator));
   const decoratorMeta = ts.createArrayLiteral(ngDecorators);
--- a/packages/compiler-cli/src/ngtsc/annotations/src/metadata.ts
+++ b/packages/compiler-cli/src/ngtsc/annotations/src/metadata.ts
@@ -71,7 +71,7 @@ export function generateSetClassMetadataCall(
         duplicateDecoratedMemberNames.join(', '));
   }
   const decoratedMembers = classMembers.map(
-      member => classMemberToMetadata(member.nameNode || member.name, member.decorators!, isCore));
+      member => classMemberToMetadata(member.nameNode ?? member.name, member.decorators!, isCore));
   if (decoratedMembers.length > 0) {
     metaPropDecorators = ts.createObjectLiteral(decoratedMembers);
   }
--- a/packages/compiler-cli/src/ngtsc/annotations/test/metadata_spec.ts
+++ b/packages/compiler-cli/src/ngtsc/annotations/test/metadata_spec.ts
@@ -90,6 +90,19 @@ runInEachFileSystem(() => {
     `);
       expect(res).toBe('');
     });
+
+    it('should preserve quotes around class member names', () => {
+      const res = compileAndPrint(`
+        import {Component, Input} from '@angular/core';
+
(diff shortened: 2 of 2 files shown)


packages/compiler-cli/src/ngtsc/annotations/src/metadata.ts
```diff
@@ -70,8 +70,8 @@ export function generateSetClassMetadataCall(
         `Duplicate decorated properties found on class '${clazz.name.text}': ` +
         duplicateDecoratedMemberNames.join(', '));
   }
-  const decoratedMembers =
-      classMembers.map(member => classMemberToMetadata(member.name, member.decorators!, isCore));
+  const decoratedMembers = classMembers.map(
+      member => classMemberToMetadata(member.nameNode || member.name, member.decorators!, isCore));
```
alxhub, line 74:
```suggestion
      member => classMemberToMetadata(member.nameNode ?? member.name, member.decorators!, isCore));
```

alxhub (approved):

AndrewKushnir:
[Presubmit](http://test/OCL:325878690:BASE:325878731:1597092210341:a21e42d0)

PR #38388, opened 2020-08-09
fix(elements): lazily check matches

### LINKED ISSUE #24551: bug(elements): importing `createCustomElement` breaks SSR

PatrickJS:
fixes #24551

same code as 
https://github.com/angular/angular/blob/d1ea1f4c7f3358b730b0d94e65b00bc28cae279c/packages/upgrade/src/common/src/downgrade_component_adapter.ts#L291-L299
without the global check

simpler version of another PR
https://github.com/angular/angular/pull/37799

### DIFF
--- a/packages/elements/src/utils.ts
+++ b/packages/elements/src/utils.ts
@@ -7,12 +7,6 @@
  */
 import {ComponentFactoryResolver, Injector, Type} from '@angular/core';
 
-const matches = (() => {
-  const elProto = Element.prototype as any;
-  return elProto.matches || elProto.matchesSelector || elProto.mozMatchesSelector ||
-      elProto.msMatchesSelector || elProto.oMatchesSelector || elProto.webkitMatchesSelector;
-})();
-
 /**
  * Provide methods for scheduling the execution of a callback.
  */
@@ -97,9 +91,19 @@ export function kebabToCamelCase(input: string): string {
 }
 
 /**
+ * Lazily check the correct `matches` method on `Element`.
+ */
+let matches: any;
+
+/**
  * Check whether an `Element` matches a CSS selector.
  */
 export function matchesSelector(element: Element, selector: string): boolean {
+  if (!matches) {
+    const elProto = Object.getPrototypeOf(element) as any;
+    matches = elProto.matches || elProto.matchesSelector || elProto.mozMatchesSelector ||
+        elProto.msMatchesSelector || elProto.oMatchesSelector || elProto.webkitMatchesSelector;
+  }
   return matches.call(element, selector);
 }
 


PatrickJS:
@googlebot I signed it!

packages/elements/src/utils.ts
```diff
@@ -96,10 +90,20 @@ export function kebabToCamelCase(input: string): string {
   return input.replace(/-([a-z\d])/g, (_, char) => char.toUpperCase());
 }
 
+/**
+ * Lazily check the correct `matches` method on `Element`.
+ */
+let _matches: (this: any, selector: string) => boolean;
```
alan-agius4, line 96: Nit: shouldn't `this` be of type `Element`?

packages/elements/src/utils.ts
```diff
@@ -96,10 +90,20 @@ export function kebabToCamelCase(input: string): string {
...
+ * Lazily check the correct `matches` method on `Element`.
+ */
+let matches: any;
+
 /**
  * Check whether an `Element` matches a CSS selector.
  */
 export function matchesSelector(element: Element, selector: string): boolean {
+  if (!matches) {
+    const elProto = Object.getPrototypeOf(element) as any;
```
gkalpak, line 103: OOC, why not use `Element.prototype`?

PatrickJS: Element is still not defined on the server globally. if we inject window and use the Element ref off that then yeah it would work. `getPrototypeOf` works in every environment though and this PR also requires the least amount of change/review

gkalpak: Why wouldn't `Element` be available on server? At the point when this is called, all window globals should be available as globals. Did I miss something?

alan-agius4: Domino is loaded at a later stage.

gkalpak: AFAICT, the `matchesSelector()` function is not called before the rendering phase.
(Note this is a different implementation that the IIFE we had before trying to access `Element` as soon as the file was imported.)

gkalpak:
The issue has been fixed by #37799. I believe this can be closed.

Issue #38393, opened 2020-08-10
Broken link in angular.io


ljharb:
# 📚 Docs or angular.io bug report

### Description

https://angular.io/guide/template-syntax#safe-navigation-operator / https://github.com/angular/angular.io/blob/281efb9ca0d278b36e2e7fa0850a807d7005e50b/public/docs/ts/latest/guide/template-syntax.jade#L1773 used to work.

Now, it's https://angular.io/guide/template-expression-operators#the-safe-navigation-operator----and-null-property-paths / https://github.com/angular/angular/blob/5b31a0a2942c50059cac4c7ccb92047a50473347/aio/content/guide/template-expression-operators.md#the-safe-navigation-operator----and-null-property-paths

## 🔬 Minimal Reproduction

Visit the old URL, see the 404.

### What's the affected URL?**
https://angular.io/guide/template-syntax#safe-navigation-operator

### Reproduction Steps**
Visit https://angular.io/guide/template-syntax#safe-navigation-operator.

### Expected vs Actual Behavior**
Expected behavior: [cool URLs don't change](https://www.w3.org/Provider/Style/URI), and any URL that has ever worked should work forever.

### Anything else relevant?
This prompted https://github.com/tc39/proposal-optional-chaining/pull/140, which will fix the issue for new visitors to the docs, but the old URL should 301 redirect indefinitely.

JoostK:
Thanks for reporting! [There are some redirect rules](https://github.com/angular/angular/blob/445ac15a785121cd7de099a33a3fded0c8b9ab83/aio/firebase.json), but this change has not been listed as a redirect.

JoostK:
After looking at #38399 (thank you @Nishikoh!) I don't think we can actually address this issue, see https://github.com/angular/angular/pull/38399#pullrequestreview-464475899 for details.

A safer reference to the docs would have been to use the versioned URL of `https://v10.angular.io/`, such that refactoring of docs content would have a smaller impact on the URLs. I realize this situation is exactly point 1 of https://www.w3.org/Provider/Style/URI; the versioned URL would reduce the chance of inaccurate links in the future.

ljharb:
@JoostK it seems like it could still be somewhat addressed (albeit not properly with a 301/302) with a clientside JS redirect, or, the old page statically just containing a link, so that users at least don't get a 404 and only have one extra click to get to the new content?

JoostK:
[...]

PR #38399, opened 2020-08-10
fix(aio): fix broken link "*#safe-navigation-operator"

Nishikoh:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

The link doesn't redirect.
Issue Number: #38393 

## What is the new behavior?
The link is redirected

- [ ] Yes
- [x] No


JoostK (changes requested):
Heya, looking at this more closely, I don't think we can actually fix this case. I interpreted "broken link" in #38393 as a link that yields a 404, but I find out now that is not actually the case here. The content has been split from one page to multiple pages, and the `/guide/template-syntax` page still exists as an index of all chapters. A URL fragment is _not_ sent to web servers, so the Firebase configuration cannot distinguish based on whether the fragment is present (it never is). Because of this, I don't think we can address this issue.

Nishikoh:
I got it. Close this PR.

JoostK:
@Nishikoh thank you for putting in the time and effort! ❤️

PR #38402, opened 2020-08-10
pre-commit formatting hook and formatting only staged files

josephperrott:
See individual commits.

### DIFF
--- a/dev-infra/utils/repo-files.ts
+++ b/dev-infra/utils/repo-files.ts
@@ -27,6 +27,18 @@ export function allChangedFilesSince(sha = 'HEAD') {
   return Array.from(new Set([...diffFiles, ...untrackedFiles]));
 }
 
+/**
+ * A list of all staged files which have been modified.
+ *
+ * Only added, created and modified files are listed as others (deleted, renamed, etc) aren't
+ * changed or available as content to act upon.
+ */
+export function allStagedFiles() {
+  return gitOutputAsArray(`git diff --staged --name-only --diff-filter=ACM`);
+}
+
+
+
 export function allFiles() {
   return gitOutputAsArray(`git ls-files`);
 }
--- a/dev-infra/format/cli.ts
+++ b/dev-infra/format/cli.ts
@@ -7,7 +7,7 @@
  */
 import * as yargs from 'yargs';
 
-import {allChangedFilesSince, allFiles} from '../utils/repo-files';
+import {allChangedFilesSince, allFiles, allStagedFiles} from '../utils/repo-files';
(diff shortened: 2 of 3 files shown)


JoostK:
How would this change behave for someone like me who never syncs the `master` ref? I currently run this command and explicitly pass `upstream/master` as ref. I think this will run the formatter on lots more files than necessary, making it somewhat slow?

Related, what happens for e.g. cherry-pick commits to older branches? Will this diff against master and therefore be similarly slow?

josephperrott:
@JoostK, good point!  I have updated the PR to include a change that allows for only running the formatter on staged files.

This should allow for you to run `ng-dev format staged` and not have to concern about a ref being used for diffing against.  Additionally, we will use this for our pre-commit hook formatting.

dev-infra/utils/repo-files.ts
```diff
@@ -27,6 +27,11 @@ export function allChangedFilesSince(sha = 'HEAD') {
   return Array.from(new Set([...diffFiles, ...untrackedFiles]));
 }
 
+/** A list of all staged files which have been modified. */
+export function allStagedFiles() {
+  return gitOutputAsArray(`git diff --cached --name-only --diff-filter=ACM`);
```
devversion, line 32: Optional nit: It might be nicer using `--staged`. I find the `--cached` option somewhat ambiguous and it would match with the general naming of the functions/logic. Also if you intend to address it, might want to add a small comment saying that we only are interested in modified, added or copied files (i.e. not renamed ones; permission changes ones etc.)

devversion (approved):
Nice! One minor optional nit.

<sup>Adding cleanup so you have a chance to reply/make changes</sup>

dev-infra/utils/repo-files.ts
```diff
@@ -27,6 +27,18 @@ export function allChangedFilesSince(sha = 'HEAD') {
...
 }
 
+/**
+ * A list of all staged files which have been modified.
+ *
+ * Only added, created and modified files are listed as others (deleted, renamed, etc) aren't
+ * changed or available as content to act upon.
+ */
+export function allStagedFiles() {
+  return gitOutputAsArray(`git diff --staged --name-only --diff-filter=ACM`);
```
gkalpak, line 37: A little late to the party, but:
Files that are both renamed and modified may appear as "renamed" (if the changes are small relative to the file's total length). So, by omitting renamed files you might still get formatting errors.

This might or might not be a desired trade-off (taking speed over accuracy for this particular command), but it might be worth documenting this in a comment here.

